PR #194, opened 2022-06-27
Update the current clusterrole

sharlns:
- Tetragon daemonset should be only allowed to get/list/watch TracingPolicy objects
- Tetragon daemonset should be only allowed to create/update the TracingPolicy CRD definition


sharlns:
I'm closing this. It will be implemented in https://github.com/cilium/tetragon/pull/195.

Issue #108, opened 2022-05-31
alignchecker for event_config


kkourt:
Once #53 is merged, add alignchecker (https://github.com/cilium/tetragon/tree/main/tools/alignchecker, https://github.com/cilium/tetragon/tree/main/pkg/alignchecker) support for `event_config`.

PR #200, opened 2022-06-28
chore: add event_config alignment check

Forsworns:
Take the issue #108 . Have tested locally.

--- 
BTW, the alignment-checker reminds me that 
```
C and Go structs alignment check failed: execvemap.ExecveValue(48) size does not match execve_map_value(112)
```

Is this expected? I found that they are defined as 
```C
// bpf/lib/process.h
struct execve_map_value {
	struct msg_execve_key key;
	struct msg_execve_key pkey;
	__u32 flags;
	__u32 nspid;
	__u32 binary;
	__u32 pad;
	struct msg_ns ns;
	struct msg_capabilities caps;
} __attribute__((packed)) __attribute__((aligned(8)));
```

and 

```go
// pkg/sensors/exec/execvemap/execve.go
type ExecveValue struct {
	Process processapi.MsgExecveKey
	Parent  processapi.MsgExecveKey
	Flags   uint32
	Nspid   uint32
	Buffer  uint64
}
```
Apparently, they are not the same...
And another related `ExecveValueL` struct in `pkg/sensors/exec/execvemap/execve.go` seems deprectated.

### DIFF
--- a/bpf/alignchecker/bpf_alignchecker.c
+++ b/bpf/alignchecker/bpf_alignchecker.c
@@ -3,6 +3,8 @@
 #include "include/vmlinux.h"
 #include "include/api.h"
 #include "lib/hubble_msg.h"
+#include "process/retprobe_map.h"
+#include "process/types/basic.h"
 
 /* DECLARE declares a unique usage of the union or struct 'x' on the stack.
  *
@@ -36,6 +38,7 @@ int main(void)
 	// from maps
 	DECLARE(struct, event, iter);
 	DECLARE(struct, execve_map_value, iter);
+	DECLARE(struct, event_config, iter);
 
 	return 0;
 }
--- a/pkg/alignchecker/alignchecker.go
+++ b/pkg/alignchecker/alignchecker.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/cilium/tetragon/pkg/api/processapi"
 	"github.com/cilium/tetragon/pkg/api/testapi"
+	"github.com/cilium/tetragon/pkg/api/tracingapi"
 	"github.com/cilium/tetragon/pkg/sensors/exec/execvemap"
 
 	check "github.com/cilium/cilium/pkg/alignchecker"
@@ -29,6 +30,7 @@ func CheckStructAlignments(path string) error {
 		"msg_exit":         {reflect.TypeOf(processapi.MsgExitEvent{})},
 		"msg_test":         {reflect.TypeOf(testapi.MsgTestEvent{})},
 		"execve_map_value": {reflect.TypeOf(execvemap.ExecveValue{})},
+		"event_config":     {reflect.TypeOf(tracingapi.EventConfig{})},
 	}
 	return check.CheckStructAlignments(path, toCheck, true)
 }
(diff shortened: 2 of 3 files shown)


bpf/alignchecker/bpf_alignchecker.c
```diff
@@ -3,6 +3,8 @@
 #include "include/vmlinux.h"
 #include "include/api.h"
 #include "lib/hubble_msg.h"
+#include "process/retprobe_map.h"
+#include "process/types/basic.h"
```
tpapagian, line 7: I believe that this include statement is not required here.

tpapagian (commented):
Thanks @Forsworns! Can you please rebase and drop the second commit (i.e. "Merge branch 'cilium:main' into alignchecker_event_config").

tpapagian:
Additionally, in `EventConfig struct` you may also add align directives similar to:https://github.com/cilium/tetragon/blob/77e32bf8f35931c84dda6f5eef1be9bb04a15ec0/pkg/api/processapi/processapi.go#L136-L140 

Using this, the alignchecker does not only check the size of the struct but also the fields inside.

tpapagian:
Regarding the check between `execvemap.ExecveValue` and `execve_map_value`, I believe that these may not be the same. `execve_map_value` is not shared with the user-space. So we can remove that check.

tpapagian:
> > these may not be the same. `execve_map_value` is not shared with the user-space.
> 
> In fact, they are shared. In `pkg/sensors/exec/procevents/proc_reader.go`, `writeExecveMap` will open write to the `execve_map` at the ebpf side. It collects the information all of the running processes when Tetragon is initialized.
> 
> See
> 
> https://github.com/cilium/tetragon/blob/77e32bf8f35931c84dda6f5eef1be9bb04a15ec0/pkg/sensors/exec/procevents/proc_reader.go#L245
> 
> and
> 
> https://github.com/cilium/tetragon/blob/77e32bf8f35931c84dda6f5eef1be9bb04a15ec0/pkg/sensors/exec/procevents/proc_reader.go#L179
> 
> Maybe it's better to append the missing items to `ExecveValue`.

I agree, maybe it is not enough to add new fields there, but we should also initialise them properly. I propose to keep that to a separate PR, so feel free to create a new one to solve this. Thanks!

tpapagian:
Created https://github.com/cilium/tetragon/issues/205 to keep track of `execve_map_value issue.

bpf/alignchecker/bpf_alignchecker.c
Forsworns: If not include them, the checker will output 

```
C and Go structs alignment check failed: tracingapi.EventConfig(80) size does not match event_config(-1)
```

I add a print before the 

https://github.com/cilium/tetragon/blob/77e32bf8f35931c84dda6f5eef1be9bb04a15ec0/vendor/github.com/cilium/cilium/pkg/alignchecker/alignchecker.go#L95

```
fmt.Printf("%v, %v\n", st.StructName, st.ByteSize)
```

And it turns out that the struct size is not read correctly.

```
event_config, -1
msg_exit, 40
msg_test, 48
```

tpapagian:
Regarding the CI failure maybe you should also add "Signed-off-by" on the PR message (except from the commits where you already did it).

tpapagian (approved):
Thanks

jrfastab:
Thanks @Forsworns nice set of work.

Issue #203, opened 2022-06-29
Panic with unexpected Execve events


Forsworns:
I run Tetragon on a server out of the k8s cluster via

`sudo LD_LIBRARY_PATH=$(realpath ./lib) ./tetragon --bpf-lib bpf/objs`

And then I login into the server via ssh, the Tetragon on the server run into the panic 

```
panic: runtime error: slice bounds out of range [:-1]

goroutine 72 [running]:
github.com/cilium/tetragon/pkg/sensors/exec.execParse(0x1)
        /root/tetragon/pkg/sensors/exec/exec.go:115 +0xae5
```
The corresponding code is 

https://github.com/cilium/tetragon/blob/a797a064874715de9126da1298a030e834c14d75/pkg/sensors/exec/exec.go#L114

I tried to add a check before slicing to avoid the panic (see the PR #207), but I don't know if this perf event is expected during login. Maybe we have to read the ebpf part to explore the reason.

Here is the perf event causing the panic, I add a breakpoint before it:

![s2](https://user-images.githubusercontent.com/24567080/176355595-90fc3665-3d54-472b-957d-d6de11b06de4.jpg)

PR #204, opened 2022-06-29
Remove legacy codes on perf events

Forsworns:
1. remove codes for perf map in `pkg/bpf`

2. remove legacy run_events and replace it with the new one

Signed-off-by: Peihao Yang <peihao.young@gmail.com>


Forsworns:
Maybe I should split off the commit for #203

tixxdz:
Hi @Forsworns thank you!

> Maybe I should split off the commit for #20

If it makes sense: yes please!  it is always good to split commits especially ones that removes or relocate codes from the one that changes functionality to easy the review process

Trung-DV:
Seems that this PR already solve by #381

PR #211, opened 2022-06-30
tetragon: Check the returned data length for exec arguments

olsajiri:
It can actually happen that there will be no arguments
or filename, in which case we need to check the data
length and api.EventErrorFilename flag.

### DIFF
--- a/pkg/sensors/exec/exec.go
+++ b/pkg/sensors/exec/exec.go
@@ -110,10 +110,12 @@ func execParse(reader *bytes.Reader) (processapi.MsgProcess, bool, error) {
 		}
 		proc.Filename = string(data[:])
 		args = args[unsafe.Sizeof(desc):]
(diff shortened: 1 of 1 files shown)


willfindlay (commented):
@olsajiri can we do similar bounds checking here? This could also panic if Index() returns -1 (which it can).

https://github.com/cilium/tetragon/blob/d6d4adeeccc51ee875487c9b163b03676a120589/pkg/sensors/exec/exec.go#L113-L116

jrfastab:
And unit test would be great.

kkourt (approved):

kkourt:
> maybe we could have a way to instrument bpf program to trigger the error for some binary name and test that.. but that'd mean more changes

:+1:

A simpler solution might be to add some unit tests on the go side for this.

PR #212, opened 2022-06-30
[docs] Correct name to Tetragon

lizrice:
CRD docs referred to FGS

Signed-off-by: Liz Rice <liz@lizrice.com>


willfindlay (approved):
Nice catch, let's ship it

PR #213, opened 2022-06-30
tetragon: improve protobuf usage to remove monolithic proto definitions

jrfastab:
Its easier to read and nicer to work on when the proto defs are not one monolithic proto with everything included in a single proto. Further, I'm looking at doing some testing/CI tooling around this so if we can just import the necessary proto files things look cleaner.

As part of the testing infra I wanted ability to test subsets of the base BPF codes. This is helpful to isolate the programs and test functionality in isolation. We add support for loading a defined set of base programs here which can easily be just a single program we want to test.

Finally, some of the codegen was not strictly necessarily so turn this into reflect.* bits to avoid doing unnecessary codegen.


cmd/protoc-gen-go-tetragon/common/common.go
```diff
@@ -16,6 +16,9 @@ import (
 // TetragonPackageName is the import path for the Tetragon package
 var TetragonPackageName = "github.com/cilium/tetragon"
 
+// EventcheckerPackageName is the import path for the Eventchecker package
+var EventcheckerPackageName = "github.com/cilium/"
```
willfindlay, line 20: This isn't the full path to the eventchecker package. Maybe the variable could have a different name?

cmd/protoc-gen-go-tetragon/common/common.go
```diff
@@ -235,10 +246,24 @@ func GetEvents(f *protogen.File) ([]*protogen.Message, error) {
...
+}
+
+func isReservedField(field string) bool {
+	for _, f := range reservedFields {
+		if field == f {
+			return true
+		}
+	}
+	return false
+}
```
willfindlay, line 261: Small nit, a map would be more idiomatic here I think.

cmd/protoc-gen-go-tetragon/eventchecker/eventchecker.go
```diff
@@ -11,54 +11,54 @@ import (
 )
 
 // Generate generates boilerplate code for the eventcheckers
-func Generate(gen *protogen.Plugin, f *protogen.File) error {
-	g := common.NewCodegenFile(gen, f, "eventchecker")
-	yaml := common.NewCodegenFile(gen, f, "eventchecker/yaml")
+func Generate(gen *protogen.Plugin, files []*protogen.File) error {
+	g := common.NewCodegenFile(gen, files[0], "eventchecker")
+	yaml := common.NewCodegenFile(gen, files[0], "eventchecker/yaml")
```
willfindlay, line 16: What's the logic behind `files[0]` here? How do we know that's the file we're looking for? A comment might be helpful.

cmd/protoc-gen-go-tetragon/helpers/helpers.go
```diff
@@ -11,43 +11,8 @@ import (
 	"google.golang.org/protobuf/compiler/protogen"
 )
 
-func generateEventTypeString(g *protogen.GeneratedFile, f *protogen.File) error {
```
willfindlay, line 14: Why are we removing this?

willfindlay (commented):
Thanks, just a few comments/nits from me.

pkg/observer/listener.go
```diff
@@ -7,14 +7,21 @@ import (
 	"encoding/gob"
 	"io"
 	"net"
+
+	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/cilium/tetragon/pkg/reader/notify"
 )
 
+type NotifyInterface interface {
```
kkourt, line 15: nit: maybe `MsgInterface` instead of `NotifyInterface`?

pkg/reader/notify/notify.go
```diff
@@ -0,0 +1,7 @@
+package notify
+
+import "github.com/cilium/tetragon/api/v1/tetragon"
+
+type NotifyInterface interface {
```
kkourt, line 5: This seems to be defined elsewhere also (pkg/observer/listener.go)

kkourt (approved):
Thanks! Just some nits from my side.

Assuming tests are ✅ and @willfindlay's comments are addressed, LGTM!

api/v1/tetragon/tetragon_ext.pb.go
```diff
@@ -0,0 +1,59 @@
+package tetragon
```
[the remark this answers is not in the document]
jrfastab: follow up PR to codegen would be nice to have.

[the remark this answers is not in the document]
jrfastab: follow up I guess.

cmd/protoc-gen-go-tetragon/common/common.go
jrfastab: I think its unused so can just be removed. It was a hold out from early version.

cmd/protoc-gen-go-tetragon/common/common.go
jrfastab: yeah its nicer that way.

cmd/protoc-gen-go-tetragon/helpers/helpers.go
jrfastab: I don't think we need them.

pkg/reader/notify/notify.go
jrfastab: ah yeah let me move it all over to notify.go it simplifies the cycle import problems.

pkg/observer/listener.go
jrfastab: hmm but its the Interface for the Notify()

cmd/protoc-gen-go-tetragon/eventchecker/eventchecker.go
jrfastab: Its only used to generate the name of the new file. We take the Prefix of the filename and then use it for our generated files. It doesn't actually matter because all gen files are in the tetragon pkg. So its arbitrary I just picked any file.

For example if we find foobar.proto as the first file then we will generate foobar.pb.go files.

pkg/observer/listener.go
kkourt: > hmm but its the Interface for the Notify()

I guess it's both the interface Notify() needs and what messages are required to implement (AFAIU). We can leave it as is. No strong feelings.

willfindlay (approved):
LGTM

PR #214, opened 2022-07-01
refactor: eliminate variables by BPF_CORE_READ_INTO

Forsworns:
The BPF_CORE_READ_INTO can help reduce unnecessary variables and simplify the logic

Signed-off-by: Peihao Yang <peihao.young@gmail.com>

### DIFF
--- a/bpf/process/bpf_process_event.h
+++ b/bpf/process/bpf_process_event.h
@@ -548,11 +548,8 @@ __event_get_task_info(struct msg_execve_event *msg, __u8 op, bool walker,
 	struct cgroup_subsys_state *subsys;
 	struct msg_process *curr;
 	struct task_struct *task;
-	struct nsproxy *nsproxy;
 	struct css_set *cgroups;
-	struct kernfs_node *kn;
 	struct cgroup *cgrp;
-	struct net *net_ns;
 	const char *name;
 
 	msg->common.op = op;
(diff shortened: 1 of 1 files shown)


bpf/lib/bpf_helpers.h
```diff
@@ -65,6 +65,21 @@ enum bpf_enum_value_kind {
...
+ *
+ * Generally, use **bpf_probe_read_user**\ () or
+ * **bpf_probe_read_kernel**\ () instead.
+ *
+ * Returns
+ * 0 on success, or a negative error in case of failure.
+ */
+static long (*bpf_probe_read)(void *dst, __u32 size,
+			      const void *unsafe_ptr) = (void *)4;
+
```
tpapagian, line 82: Why do we need that?

bpf/process/bpf_process_event.h
```diff
@@ -612,18 +603,12 @@ __event_get_task_info(struct msg_execve_event *msg, __u8 op, bool walker,
...
-					} else {
-						curr->flags |=
-							EVENT_DOCKER_NAME_ERR;
-					}
+				if (BPF_CORE_READ_INTO(&name, cgrp, kn, name) ==
+				    0) {
+					probe_read_str(msg->kube.docker_id,
+						       DOCKER_ID_LENGTH, name);
+				} else {
+					curr->flags |= EVENT_DOCKER_NAME_ERR;
```
kkourt, line 611: Thanks!

Would it make sense to use BPF_CORE_READ_STR_INTO for the final string?

Also, I wonder if it would make sense to also wrap the cgroups / subsys reads above into such a call. This would mean that we wouldn't not be able to distinguish the different errors (e.g., EVENT_DOCKER_SUBSYS_ERR) but I think that's fine.

Forsworns: yep, it's possible to wrap all of them. I found tetragon records those errors, so I didn't change them. 

I'll review related error handlers in golang side to check if it's safe to remove these error flags.

Forsworns: @kkourt In the golang part, these errors are defined at

https://github.com/cilium/tetragon/blob/b87b5d65ac1e92fc43cd6c3d85f6c37e4c1ae490/pkg/api/flags.go#L77

and decoded at 

https://github.com/cilium/tetragon/blob/b87b5d65ac1e92fc43cd6c3d85f6c37e4c1ae490/pkg/reader/exec/exec.go#L57

emmm, don't know if it is a good idea to remove them...

kkourt: Thanks for looking into this!

I'd be personally OK with keeping a single error for all different ways that getting the docker id would fail. @jrfastab what do you think?

kkourt: We can leave that as a followup, I guess.

kkourt (approved):

Forsworns:
Rebase since the definition of `bpf_probe_read` has been added before :)

tpapagian (approved):

PR #218, opened 2022-07-06
Add tests for pkg/grpc/exec

tpapagian:
An exec and exit event can be delivered out of order to the agent. For this reason, we have a cache that keeps the un populated events for a small period of time to wait for the exec event.

There are also other caches that deal with this situations.

This commit adds tests for pkg/grpc/exec. For now it only contains 2 different cases:
1. exec and exit events in order (TestGrpcExecInOrder)
2. exec and exit events out of order (TestGrpcExecOutOfOrder)

### DIFF
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -74,12 +74,6 @@ func handleExecEvent(event *cacheObj, nspid uint32) error {
 func handleEvent(event *cacheObj) error {
 	p := event.event.GetProcess()
 
-	endpoint := process.GetProcessEndpoint(p)
-	if p.Docker != "" && (endpoint == nil || p.Pod == nil) {
-		errormetrics.ErrorTotalInc(errormetrics.EventCacheEndpointRetryFailed)
-		return fmt.Errorf("failed to get process endpoint ifno")
-	}
-
 	// If the process wasn't found before the Add(), likely because
 	// the execve event was processed after this event, lets look it up
 	// now because it should be available. Otherwise we have a valid
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -50,15 +50,19 @@ type Cache struct {
 }
 
 func handleExecEvent(event *cacheObj, nspid uint32) error {
+	var podInfo *tetragon.Pod
+
 	p := event.event.GetProcess()
 	containerId := p.Docker
 	filename := p.Binary
 	args := p.Arguments
 
-	podInfo, _ := process.GetPodInfo(containerId, filename, args, nspid)
-	if podInfo == nil {
-		errormetrics.ErrorTotalInc(errormetrics.EventCachePodInfoRetryFailed)
-		return fmt.Errorf("failed to get pod info")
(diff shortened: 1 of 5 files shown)


pkg/eventcache/eventcache.go
```diff
@@ -25,7 +25,7 @@ const (
 
 // garbage collection run interval
 const (
-	eventRetryTimer = time.Second * 10
+	eventRetryTimer = time.Second * 1
```
kkourt, line 28: `1s` seems reasonable to me, but I wonder if the `10s` timeout was there for some practical reason based on things observed on the field. Is the motivation for this to execute the test faster? In that case, I think we can just have a var for `eventRetryTimer` and modify it for the unit test (and we can even set it lower than `1s`).

pkg/grpc/exec/exec_test.go
```diff
@@ -0,0 +1,264 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+// CGO_LDFLAGS=-L$(realpath ./lib) go test -gcflags="" -c ./pkg/grpc/exec/ -o go-tests/grpc-exec.test
+// sudo LD_LIBRARY_PATH=/home/apapag/tetragon/lib ./go-tests/grpc-exec.test  [ -test.run TestGrpcExec ]
```
kkourt, line 5: nit: use $(realpath ./lib) here as well?

pkg/grpc/exec/exec_test.go
```diff
@@ -0,0 +1,264 @@
...
+	lExecCache := execcache.New(lServer, lDns)
+
+	return New(lExecCache, lEventCache, false, false), cancel
+}
+
+func TestGrpcExecOutOfOrder(t *testing.T) {
+	var cancelWg sync.WaitGroup
+
+	AllEvents = nil
+	grpc, cancel := initEnv(t, &cancelWg)
```
kkourt, line 182: nit: maybe do 

```
defer func() {
    cancel()
   cancelWg.Wait()
}()
```

here?

pkg/grpc/exec/exec_test.go
```diff
@@ -0,0 +1,264 @@
...
+	if e1 != nil {
+		AllEvents = append(AllEvents, e1)
+	}
+
+	time.Sleep(2 * time.Second) // wait for cache to do it's work
+
+	assert.Equal(t, len(AllEvents), 2)
+
+	var ev1 *tetragon.GetEventsResponse
+	var ev2 *tetragon.GetEventsResponse
```
kkourt, line 246: nit: `var ev1, ev2 *tetragon.GetEventsResponse`

kkourt (commented):
Thanks!

Your last commit mentions:
> This commit tries to fix the issues with the previous tests.

Are those issues described somewhere? That is, where there bugs uncovered with the test? If so, I think it's worth describing them in the commit message. E.g., did the test fail? How?

Another concern I have is that this test (as is) means that we would need to duplicate it fully in OSS, but I don't see any (easy) way out of this so I don't have anything to suggest :(

jrfastab (approved):

Issue #220, opened 2022-07-07
bpf_execve_event_v53 compiled with clang-13 fails verifier on Linux 5.4 


willfindlay:
```
Verifying bpf/objs/bpf_execve_event_v53.o... Failed!
 R0=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R1_w=inv4294967295 R6_w=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R7=map_value(id=0,off=0,ks=4,vs=1848,imm=0) R8=map_value(id=0,off=248,ks=4,vs=1848,imm=0) R9=map_value(id=0,off=0,ks=4,vs=1848,imm=0) R10=fp0 fp-8=????mmmm fp-72=mmmmmmmm fp-80=mmmmmmmm fp-88=mmmmmmmm fp-112=mmmmmmmm fp-120=mmmmmmmm fp-128=mmmmmmmm fp-136=????mmmm fp-176=mmmm???? fp-184=ctx fp-192=mmmmmmmm fp-200=ctx fp-208=mmmmmmmm fp-216=mmmmmmmm fp-224=map_value fp-232=map_value fp-248=mmmmmmmm fp-256=mmmmmmmm
1587: (7b) *(u64 *)(r10 -272) = r8
1588: (79) r7 = *(u64 *)(r10 -248)
;
1589: (0f) r7 += r9
last_idx 1589 first_idx 1583
regs=80 stack=0 before 1588: (79) r7 = *(u64 *)(r10 -248)
regs=0 stack=40000000 before 1587: (7b) *(u64 *)(r10 -272) = r8
regs=0 stack=40000000 before 1586: (15) if r6 == 0x0 goto pc+177
regs=0 stack=40000000 before 1584: (18) r1 = 0xffffffff
regs=0 stack=40000000 before 1583: (bf) r6 = r0
[...]

PR #171, opened 2022-06-15
build: use clang-13 to build bpf programs

willfindlay:
This needs more testing on additional kernels before we merge. For example, while testing locally I noticed some verifier errors on 5.4 that we should fix first.

PR #221, opened 2022-07-08
Fix for verifier issue in 5.4 with clang-13

tpapagian:
FIXES: https://github.com/cilium/tetragon/issues/220

Tested over https://github.com/cilium/tetragon/pull/171 for clang-13 support.

### DIFF
--- a/bpf/process/data_event.h
+++ b/bpf/process/data_event.h
@@ -6,30 +6,26 @@
 static inline __attribute__((always_inline)) long
 __do_bytes(void *ctx, struct msg_data *msg, unsigned long uptr, size_t bytes)
 {
-	size_t max = sizeof(msg->arg) - 1;
-	size_t rd_bytes, size;
 	int err;
 
-	rd_bytes = bytes > max ? max : bytes;
-	msg->common.size = offsetof(struct msg_data, arg) + rd_bytes;
-
 	/* Code movement from clang forces us to inline bounds checks here */
-	asm volatile("%[rd_bytes] &= 0x7fff;\n"
-		     "if %[rd_bytes] < 32736 goto +1\n;"
-		     "%[rd_bytes] = 32736;\n"
-		     :
-		     : [rd_bytes] "+r"(rd_bytes)
-		     :);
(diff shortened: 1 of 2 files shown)


willfindlay:
Thanks for the fix! I'll have a go at testing this as well on Monday.

olsajiri (approved):

bpf/process/data_event.h
```diff
@@ -6,30 +6,26 @@
...
-
 	/* Code movement from clang forces us to inline bounds checks here */
-	asm volatile("%[rd_bytes] &= 0x7fff;\n"
-		     "if %[rd_bytes] < 32736 goto +1\n;"
-		     "%[rd_bytes] = 32736;\n"
-		     :
-		     : [rd_bytes] "+r"(rd_bytes)
-		     :);
-	err = probe_read(&msg->arg[0], rd_bytes, (char *)uptr);
+	asm volatile goto("%[bytes] &= 0x7fff;\n"
```
kkourt, line 12: I'm reading the comment in: https://github.com/cilium/tetragon/blob/38d3a6135cf9e23265c4319c3507bfffb3b42a5f/bpf/lib/data_msg.h#L18-L28

which describes the `size &= 0x7fff` part. Is this still needed after this change?

tpapagian: Yes, you are correct. It seems that we can remove the `"%[bytes] &= 0x7fff;\n"` command as well. I will update the PR.

kkourt: Thanks!

How about we also add a:

```
#define MESSAGE_DATA_ARG_LEN 32736
```

And use that both in `struct msg_data` and the assembly code? (I think there should be a way to pass constants as arguments).

tpapagian: Done and tested in 4.19, 5.4, 5.10, and 5.13.

kkourt (approved):

jrfastab:
ping @tpapagian would like to merge this if we can and get off clang fork. Can you take a look thanks!

PR #226, opened 2022-07-11
tests: introduce end-to-end test framework

willfindlay:
This PR introduces a new end-to-end test framework in `tests/e2e` which we can use to
write cross-provider end-to-end tests for Tetragon. The tests themselves live in
`tests/e2e/tests`. I have added a skeleton file with lots of comments that folks can use
as a starting point for writing new tests.

Updated the Makefile to support e2e-tests by adding a separate e2e-test target and making
sure that regular `make test` will not call into the e2e-tests by pointing it directly at
`./pkg/...`. Also added an EXTRA_TESTFLAGS variable to the Makefile that can be used to
pass flags directly to the test binary from either a `make test` or `make e2e-test`.

The plan is to use this new framework to start writing e2e tests for Tetragon features.


tests/e2e/checker/rpcchecker.go
```diff
@@ -0,0 +1,259 @@
...
+		logs:             new(bytes.Buffer),
+		checkerStartedWG: new(sync.WaitGroup),
+		lock:             new(sync.Mutex),
+	}
+	// Mark that the checker has not yet started.
+	rc.checkerStartedWG.Add(1)
+	return rc
+}
+
+// WithEventLimit sets the event limit for an RPCChecker. RPCChecker.Check sees more
```
kkourt, line 62: nit: `if RPCChecker.Check sees...`

tests/e2e/checker/rpcchecker.go
```diff
@@ -0,0 +1,259 @@
...
+		{
+			Namespace: namespaces,
+		},
+	}, []*tetragon.Filter{})
+}
+
+func (rc *RPCChecker) CheckWithFilters(connTimeout time.Duration, allowList, denyList []*tetragon.Filter) features.Func {
+	return func(ctx context.Context, t *testing.T, _ *envconf.Config) context.Context {
+		rc.lock.Lock()
+		defer rc.lock.Unlock()
```
kkourt, line 116: I'm wondering why is this lock needed? AFAICT, it is only used in this function. Can this function be called concurrently from different goroutines? If so, this seems weird to me because I assumed we had one checker that talks to the multiplexer.

tests/e2e/checker/rpcchecker.go
```diff
@@ -0,0 +1,259 @@
...
+			assert.Fail(t, "failed to find forwarded gRPC ports")
+			return ctx
+		}
+
+		var addrs []string
+		for _, port := range ports {
+			addrs = append(addrs, fmt.Sprintf("localhost:%d", port))
+		}
+
+		if rc.getEvents == nil {
```
kkourt, line 132: Could this be not nil? Can we enter the `CheckWithFilters` with .connect() already called? (AFAICT, this is the only place that getEvents is set)

tests/e2e/helpers/dumpinfo.go
```diff
@@ -197,3 +200,52 @@ func dumpMetrics(port string, podName string, exportDir string) {
 		klog.ErrorS(err, "failed to write to metrics file", "file", fname, "addr", metricsAddr)
 	}
 }
+
+// dumpBpftool dumps bpftool progs and maps for a pod
+func dumpBpftool(ctx context.Context, client klient.Client, exportDir, podNamespace, podName, containerName string) {
```
kkourt, line 205: Is there a reason why we are not just running `bugtool` here?

kkourt (approved):
Thanks William.

I have some minor questions/comments, but otherwise LGTM.

tests/e2e/checker/rpcchecker.go
willfindlay: If memory serves I was concerned about test authors using the same checker multiple times in a single testenv.TestInParallel.

tests/e2e/checker/multiplexer.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	return nil
+}
+
+// GetEventsWithFilters calls GetEvents for each client in the multiplexer and returns a channel that
+// multiplexes the GetEventsResponses. allowList and denyList can be used to filter what
+// events we care about.
+func (cm *ClientMultiplexer) GetEvents(ctx context.Context, allowList, denyList []*tetragon.Filter) (chan GetEventsResult, error) {
+	c := make(chan GetEventsResult)
```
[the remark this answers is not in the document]
willfindlay: Figuring out when writers are done is up to the caller, since we're queueing both the GetEventsResponse and the error. However I think we do need to close it when we handle ctx.Done() down below, so I'll do that here.

tests/e2e/checker/rpcchecker.go
tixxdz: So maybe a comment here? to prevent accidental changes

tests/e2e/checker/rpcchecker.go
```diff
@@ -0,0 +1,259 @@
...
+			return ctx
+		}
+	}
+}
+
+func (rc *RPCChecker) Check(connTimeout time.Duration) features.Func {
+	return rc.CheckWithFilters(connTimeout, []*tetragon.Filter{}, []*tetragon.Filter{})
+}
+
+func (rc *RPCChecker) CheckNamespace(connTimeout time.Duration, namespaces ...string) features.Func {
```
tixxdz, line 105: Maybe rename to CheckWithNamespaceFilters ?

tests/e2e/helpers/dumpinfo.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/cilium/tetragon/tests/e2e/state"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/e2e-framework/klient/k8s/resources"
+	"sigs.k8s.io/e2e-framework/pkg/envconf"
+)
+
+var (
+	TetragonContainerName = "tetragon"
+	TetragonJsonPathname  = "/var/run/cilium/tetragon/tetragon.log"
```
tixxdz, line 31: Hmm maybe /var/run/tetragon? is there a reason for having cilium there ?

tests/e2e/runners/runners.go
```diff
@@ -0,0 +1,169 @@
...
+		klog.Fatalf("Cannot call Runner.Setup() twice")
+	}
+	r.hasCalledSetup = true
+
+	cfg, err := envconf.NewFromFlags()
+	if err != nil {
+		klog.Fatalf("Failed to configure test environment: %w", err)
+	}
+	klog.Info("IMPORTANT: Tetragon e2e tests require parallel tests enabled. User preferences will be ignored.")
+	cfg = cfg.WithParallelTestEnabled()
```
tixxdz, line 125: Can we have the flags or the configurations dumped before tests output ? as we do it now with tetragon, otherwise in separate PR

tests/e2e/tests/skeleton/skeleton_test.go
```diff
@@ -0,0 +1,169 @@
...
+	// Any additional setup and cleanup can be performed here if you like.
+	// This would be done using testenv.Setup() and testenv.Finish() respectively.
+	//
+	// These take a list of functions to register at test start and test end that can be
+	// used to maniplate the test context and test environment.
+	//
+	// Here we ensure our test namespace doesn't already exist then create it.
+	testenv.Setup(func(ctx context.Context, c *envconf.Config) (context.Context, error) {
+		ctx, _ = helpers.DeleteNamespace(namespace, true)(ctx, c)
+
```
tixxdz, line 75: Just to get this right: overall we can have parallel tests and they won't clash as long as they run in different namespace? and it is developer responsibility to make sure that he is reusing the same namespace between DeleteNamespace() and tests running ?
Thinking if we should remove DeleteNamespace() here and if the test env exists and it fails we know maybe something is wrong on how they are setup... better than having tests stopped or something disappearing and trying to figure it out... does this makes sense ?

tests/e2e/helpers/dumpinfo.go
willfindlay: This is specific to our Helm chart. If we want to change it here we have to change it there first.

tests/e2e/checker/rpcchecker.go
willfindlay: How about CheckInNamespace? I don't want it to be too verbose.

tests/e2e/tests/skeleton/skeleton_test.go
willfindlay: > Just to get this right: overall we can have parallel tests...

In theory yes, but unfortunately due to the way the Cilium installation helpers are written on the Cilium side you can't run tests in parallel at the moment. Once we resolve those issues up stream then yes, it should be possible.

> ... and they won't clash as long as they run in different namespace?

Right. Also thinking about running the same test back to back here. For example you could imagine running with -count 3 or something.

> ...and it is developer responsibility to make sure that he is reusing the same namespace between DeleteNamespace() and tests running?

Another option is to just use whatever namespace e2e-framework gives you. There is definitely a trade-off here in terms of idempotency vs too much magic. I lean toward giving the test author control over what namespace(s) to use in the test though.

> Thinking if we should remove DeleteNamespace() here and if the test env exists and it fails we know maybe something is wrong on how they are setup

That's how I had it initially but it breaks the case where you want to run tests back to back e.g. with -count (see above).

Let me know if the above makes sense to you. Happy to have my mind changed on any of these points if folks feel strongly one way or the other.

tests/e2e/checker/rpcchecker.go
tixxdz: Perfect!

tests/e2e/tests/skeleton/skeleton_test.go
tixxdz: Sounds good thank you @willfindlay

tixxdz (approved):

tests/e2e/checker/doc.go
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+// This package provides a Tetragon gRPC client multiplexer and an RPCChecker that wraps
+// a MultiEventChecker and uses the gRPC multiplexer to get a stream of events from all
+// Tetragon pods.
```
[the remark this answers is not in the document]
willfindlay: :smile:

tests/e2e/checker/rpcchecker.go
kkourt: > I was concerned about test authors using the same checker multiple times in a single testenv.TestInParallel.

I see. Out of curiosity, how would one go about doing this?

tests/e2e/checker/rpcchecker.go
kkourt: > In the case where we want to use the same checker more than once we keep the original connection open. Useful for example when we want to run the same workload multiple times in a stress test.

Similar to my previous comment. How would one go about doing this?

kkourt:
@willfindlay since you seem to be still working on it, I'll move it to draft. Please mark it as ready when you feel it's ready to be merged.

willfindlay:
@kkourt I think it's basically ready tbh (needs a squash/rebase obviously), but we should wait until the bug causing the new e2e job to fail has been fixed. @jrfastab says he is working on it atm.

willfindlay:
@kkourt I think it's in a good state to ship it now. I'll leave it for you just in case you want to do another round of comments in the morning.

PR #231, opened 2022-07-13
tetragon: create api go module for tooling

jrfastab:
Working on tooling and event handlers, such as hubble, its a bit ugly
to pull in the entire go mod when the user simply wants the API
definitions. Further, its useful to version the API so we can all
agree on what version is in use. Allowing us to fix core code easily
and keep the API side consistent from a versioning perspective.

To facilitate this create go module out of API code.


willfindlay (approved):
LGTM. I just had one question and one suggestion.

Question: Will vendoring API inside of tetragon and tetragon inside of API ever create a dependency cycle in the future? I would suggest maybe having pkg/matchers be its own module or moving pkg/matchers into API to avoid this.

Suggestion: We should consider moving cmd/protoc-gen-go-tetragon into the API module to cleanly separate concerns between the two modules. It made sense to have it live in cmd/ when they were all one module but IMO not so much anymore.

I don't think either of the above points are worth blocking this on so I'll just ACK for now.

PR #232, opened 2022-07-13
Tetra: fix a typo in tracingpolicy command short description

mtardy:
It's as boring as it sounds but I spotted it when I discovered the project, so why not fix it!


tpapagian (approved):
Thanks!

tixxdz (approved):
Thank you @mtardy

PR #233, opened 2022-07-13
pkg:sensors: simplify GetRunningProcs() when we read running processes

tixxdz:
GetRunningProcs() that reads process entries from procfs during startup
takes two boolean arguments which we always set to true, then it passes
these arguments down to how we send data to listeners and to write to the
execve_map.

Let's simplify this and remove these arguments as this makes reading code
easier.

Signed-off-by: Djalal Harouni <djalal@isovalent.com>


kkourt (approved):
Thanks!

PR #240, opened 2022-07-15
eventchecker: fix type mismatch bug in argument checkers

willfindlay:
See commit messages


willfindlay:
Just beating up CI to see if we get any new checker failures from user error while writing tests. These would have been invisible before.

tpapagian (approved):
Thanks!

kkourt (approved):

PR #242, opened 2022-07-18
tetragon: Avoid passing dir options around and add them to option

jrfastab:
Dir variables are passed around and embedded in the observer even though
they are set once and fixed. Further they are not part of the observer
object at all. The fallout is then random pkgs reach into observer
just to lookup directories. This can then cause cycles and makes dev
work more difficult.

Lets skip passing them around and read them directly through the option
pkg.

### DIFF
--- a/pkg/observer/observer_test_helper.go
+++ b/pkg/observer/observer_test_helper.go
@@ -229,10 +229,10 @@ func newDefaultTestOptions(t *testing.T, opts ...TestOption) *TestOptions {
 }
 
 func newDefaultObserver(t *testing.T, oo *testObserverOptions) *Observer {
-	return NewObserver(observerTestDir,
-		observerTestDir,
-		"",
-		oo.config)
+	option.Config.BpfDir = observerTestDir
+	option.Config.MapDir = observerTestDir
+	option.Config.CiliumDir = ""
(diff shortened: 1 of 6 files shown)


pkg/observer/observer_test_helper.go
```diff
@@ -398,15 +398,15 @@ func loadExporter(t *testing.T, obs *Observer, opts *testExporterOptions, oo *te
 	return nil
 }
 
-func loadObserver(t *testing.T, obs *Observer, base *sensors.Sensor, sens []*sensors.Sensor, notestfail bool) error {
-	if err := base.Load(context.TODO(), obs.bpfDir, obs.mapDir, obs.ciliumDir); err != nil {
+func loadObserver(t *testing.T, base *sensors.Sensor, sens []*sensors.Sensor, notestfail bool) error {
```
willfindlay, line 401: Could we rename this to `loadSensors` since it no longer has anything to do with the observer?

willfindlay (approved):
LGMT, just left one minor comment.

pkg/observer/observer_test_helper.go
jrfastab: Yes but I'll do it in follow up.

PR #244, opened 2022-07-18
tetragon: Add bpf program loader tests

olsajiri:
Adding CheckSensorLoad function that takes array of programs
and maps and checks they are loaded and share specified maps.

Plus related changes to allow that.
    
### DIFF
--- a/pkg/sensors/program/coll.go
+++ b/pkg/sensors/program/coll.go
@@ -1,0 +1,170 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package program
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/tetragon/pkg/logger"
+)
+
+// The idea of LoadedCollection is to keep loaded programs and maps
+// from ebpf.Collection to be used later for 'TestLoad*' tests to
+// verify loaded programs and maps.
+//
+// We can't just iterate all existing programs and maps for 2 reasons:
+//   - we could race with others loading bpf programs
+//   - we get limited name (16 bytes), which we usually cross with
+//     our names for maps and programs
+//
+// The process of loading LoadedCollection is following:
+//
+//   - ebpf collection is loaded
+//
+//     coll := NewCollectionWithOptions
+//
+//   - copy all programs/maps from ebpf collection
+//
+//     copyLoadedCollection(lc, coll)
+//
+//   - collection is removed and only 'used' programs and maps stay loaded
+//     so we filter out unused programs/maps with (based on their IDs)
+//
+//     load.LC = filterLoadedCollection(lc)
+//
+// This way we have only programs/maps realted to our test and with
+// original long names.
+//
+// All this is happening only when program.KeepCollection is set true,
+// so it's enabled only for testing code.
+
+var (
+	KeepCollection bool
+)
+
+type LoadedMap struct {
+	ID ebpf.MapID
+}
+
+type LoadedProgram struct {
+	ID     ebpf.ProgramID
+	MapIDs []ebpf.MapID
+	Type   ebpf.ProgramType
+}
+
+type LoadedCollection struct {
+	Programs map[string]*LoadedProgram
+	Maps     map[string]*LoadedMap
+}
+
+func newLoadedCollection() *LoadedCollection {
+	lc := &LoadedCollection{}
+	lc.Maps = map[string]*LoadedMap{}
+	lc.Programs = map[string]*LoadedProgram{}
+	return lc
+}
+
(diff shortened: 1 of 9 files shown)


pkg/sensors/program/coll.go
```diff
@@ -0,0 +1,168 @@
...
+//     copyLoadedCollection(lc, coll)
+//
+//   - collection is removed and only 'used' programs and maps stay loaded
+//     so we filter out unused programs/maps with
+//
+//     load.LC = filterLoadedCollection(lc)
+//
+// This was we have only programs/maps realted to our test and with
+// original long names.
+//
```
kkourt, line 40: Thanks for writing this down!

One thing that got me thinking is the following. You say that:
> // We can't just iterate all load programs and maps for 2 reasons:
> //   - we could race with others loading bpf programs

Couldn't we race with other programs between copyLoadedCollection() and filterLoadedCollection()? If we only use this for testing, the race might be acceptable but I'm curious.

pkg/sensors/program/coll.go
```diff
@@ -0,0 +1,168 @@
...
+//   - copy all programs/maps from ebpf collection
+//
+//     copyLoadedCollection(lc, coll)
+//
+//   - collection is removed and only 'used' programs and maps stay loaded
+//     so we filter out unused programs/maps with
+//
+//     load.LC = filterLoadedCollection(lc)
+//
+// This was we have only programs/maps realted to our test and with
```
kkourt, line 38: Not sure I fully understand the above. How about:
> Now load.LC contains programs/maps that are related to our test, are actually loaded, and have their original long names.

pkg/sensors/program/coll.go
```diff
@@ -0,0 +1,168 @@
...
+	for name, lp := range lc.Programs {
+		logger.GetLogger().Infof("%d: %s - %v", lp.ID, name, lp.MapIDs)
+	}
+	logger.GetLogger().Infof("Maps (%s):", str)
+	for name, lm := range lc.Maps {
+		logger.GetLogger().Infof("%d: %s", lm.ID, name)
+	}
+}
+
+func copyLoadedCollection(lc *LoadedCollection, coll *ebpf.Collection) error {
```
kkourt, line 81: how about writing this as:
> copyLoadedCollection(coll *ebpf.Collection) (*LoadedCollection, error) 

I think it makes the interface clearer (?)

pkg/sensors/program/coll.go
```diff
@@ -0,0 +1,168 @@
...
+		lp.MapIDs = mapIDs
+		lp.Type = p.Type()
+		lc.Programs[name] = lp
+	}
+	return nil
+}
+
+// Gets all the programs/maps and removes any non existent ones
+// from passed collection
+func filterLoadedCollection(lc *LoadedCollection) *LoadedCollection {
```
kkourt, line 122: nit: how about writing this as:
> func (lc *LoadedCollection) filter() *LoadedCollection {
> ...
> }

We coud even do
> func (lc *LoadedCollection) filter() {
> ...
> }

And do the changes in-place.

pkg/observer/observer.go
```diff
@@ -252,6 +252,8 @@ type Observer struct {
 
 	/* YAML Configuration File */
 	configFile string
+
+	Sensors []*sensors.Sensor
```
kkourt, line 256: It fees a bit strange adding this here. The way that the observer is supposed to manage sensor is via the "sensor manager" which is a goroutine for managing sensors.

See: https://github.com/cilium/tetragon/blob/bb6f985d32a3a5777d9c922961c77ac5d0bdbfb9/pkg/sensors/sync.go#L24-L30

From what I can tell, this is for testing only, so maybe we can keep the list elsewhere if we need to?

kkourt (commented):
Thanks! LGTM! This is definitely needed.

I have some comments. None of them is a blocker, but might be worthwhile discussing them before merging.

pkg/sensors/program/coll.go
olsajiri: as we discussed later the filtering is based on program/map ID, so it won't race

pkg/sensors/program/coll.go
olsajiri: LC contains arrays of program and maps that are related to the object we loaded plus we filtered out unused program and maps that we unloaded.. all this with original long names for both programs and maps

pkg/sensors/program/coll.go
olsajiri: yes, I will put the LoadedCollection as return value, thanks

pkg/observer/observer.go
olsajiri: as discussed later I'll change this to have sensors are return value

kkourt (approved):

PR #235, opened 2022-07-14
kprobes: matchArgs improvements (for fd/file)

tpapagian:
In kprobes, we support `matchArgs` selector for `fd` and `file` types with `Equal`, `Prefix`, and `Postfix` operators.

This PR first introduces unit-tests for all of these cases. These tests show an issue in the case of `Postfix` operator and we also fix this issue.

Currently, `matchArgs` for `fd/file` types support up to 2 values. This PR also increases this number for newer kernels (5.3+) to 8. It keeps that to 2 for older kernels. 

Finally, we also update unit-tests to check these.

ps. Tested locally on 4.19, 5.4, 5.10, and 5.13.

Signed-off-by: Anastasios Papagiannis <tasos.papagiannnis@gmail.com>

PR #246, opened 2022-07-19
add support for bpf_attr type

sarahfujimori:
This patch allows us to collect name, program type, and instruction
count of eBPF programs when they are verified.


tpapagian (commented):
Thanks!

Some initial comments:
1. You should also commit the generated files (i.e. inside `api/` directory type `make` to generate grpc files and then in the top-level directory `make codegen && make generate`).
2. It seems that it lacks proper code formatting (i.e. `make format`).
3. In `crds/examples/` there are a couple of examples on how to use various types. It would be great if you can add an example there as well.
4. It would be also great if we can have a simple unit test for that (but this can be also done as a followup into a separate PR).

tpapagian (commented):
There is an error in `checkpatch.pl` which requires a header similar to [this](https://github.com/cilium/tetragon/blob/391dc9777bb8c65f80685972987c781177e04bbf/bpf/process/bpf_execve_event.c#L1-L2) in `bpf/process/types/bpfattr.h`.

Other than that, this looks good to me.

api/v1/tetragon/tetragon.proto
```diff
@@ -170,6 +176,7 @@ message KprobeArgument {
 	KprobeSock sock_arg = 9;
 	KprobeCred cred_arg = 10;
 	int64 long_arg = 11;
+    KprobeBpfAttr bpf_attr_arg = 12;
```
tpapagian, line 179: Please fix also this formatting issue. I think that our linters do not run on the `.proto` file.

willfindlay:
We can ignore the checkpatch job for now, that warning is a false positive which I am fixing here https://github.com/cilium/tetragon/pull/258

willfindlay:
@sarahfujimori Mind running `go mod tidy -compat=1.17 && go mod vendor` and rebasing? The test failure we are seeing here was also allegedly fixed in https://github.com/cilium/tetragon/pull/235 so a rebase on top of latest main should do the trick there too.

jrfastab:
needs a rebase as well..

sarahfujimori:
see #275

PR #248, opened 2022-07-19
tetragon: Add K8s API go.mod for tooling

jrfastab:
At the moment operators and other tooling pull in all of tetragon just to
get the CRD APIs. Lets make it a go.mod and then they can get just the API.
It also helps clearly define reviewers so that operator owners and other
tools relying on the API can easily target just a single go.mod.


michi-covalent (approved):
assuming that the CI passes

PR #253, opened 2022-07-20
fix checker codegen

willfindlay:
This patch fixes checker codegen so that it at least runs and compiles in the presence of protobuf maps. For now we aren't running any event checks for maps, that can come later. I need to think some more about the right approach first.


michi-covalent (approved):

willfindlay:
I'm closing this since @michi-covalent already merged it into main.

PR #254, opened 2022-07-21
Implement pod label filter

michi-covalent:
2 commits:

- add pod_labels field.
- implement a filter that matches on pod_labels field.


cmd/protoc-gen-go-tetragon/eventchecker/field.go
```diff
@@ -209,6 +209,10 @@ func doGetFieldFrom(field *Field, g *protogen.GeneratedFile, handleList, handleO
 		return doListFrom()
 	}
 
+	if handleList && field.Desc.IsMap() {
+		return "// TODO: implement fromMap", nil
```
sharlns, line 213: Is it enough the implement these checks later?

sharlns (approved):

cmd/protoc-gen-go-tetragon/eventchecker/field.go
willfindlay: @sharlns It's on my TODO list, but I don't think it has to block this PR. FWIW I think this is still WIP.

pkg/filters/labels.go
```diff
@@ -0,0 +1,57 @@
...
+		s, err := k8sLabels.Parse(selector)
+		if err != nil {
+			return nil, err
+		}
+		selectors = append(selectors, s)
+	}
+	return func(ev *v1.Event) bool {
+		process := GetProcess(ev)
+		if process == nil || process.Pod == nil {
+			return false
```
kkourt, line 30: I guess this means that if user specified 'key != label', which according to https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors is legit, an event with no pod information will not match. I think that's (probaby) fine, but maybe we should document it?

(see other comment)

api/v1/tetragon/events.proto
```diff
@@ -42,6 +42,9 @@ message Filter {
     // Filter by process.arguments field using RE2 regular expression syntax:
     // https://github.com/google/re2/wiki/Syntax
     repeated string arguments_regex = 7;
+    // Filter events by pod labels using Kubernetes label selector syntax:
+    // https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors
```
kkourt, line 46:
```
+ If  a labels is defined, events without Pod information will _always_ be filtered out.
```

(see other comment)

michi-covalent: yeah let me add a separate documentation for the current filter syntax

pkg/filters/labels.go
michi-covalent: yeah that's an interesting point, let me think about this a bit and i'll document it 💭

api/v1/tetragon/events.proto
michi-covalent: updated the comment. i'll follow up with a separate PR to document filter syntax once this PR gets merged.

kkourt (approved):

Issue #238, opened 2022-07-15
tracing tests: Mkdir failed: mkdir /tmp2: file exists


kkourt:
Using the new test option (`-disable-tetragon-logs`, https://github.com/cilium/tetragon/pull/237) I'm seeing:

```
--- PASS: TestKprobeObjectOpen (2.78s)  
=== RUN   TestKprobeObjectOpenMount     
    kprobe_test.go:583: Mkdir failed: mkdir /tmp2: file exists
    kprobe_test.go:584:                 
--- SKIP: TestKprobeObjectOpenMount (0.00s)
=== RUN   TestKprobeObjectMultiValueOpen
    jsonchecker.go:132: jsonTestCheck: opening: /tmp/tetragon.gotest.TestKprobeObjectMultiValueOpen.359677843.json
    filenames.go:60: deleting export file for TestKprobeObjectMultiValueOpen (/tmp/tetragon.gotest.TestKprobeObjectMultiValueOpen.359677843.json)
--- PASS: TestKprobeObjectMultiValueOpen (2.75s)
=== RUN   TestKprobeObjectMultiValueOpenMount
    kprobe_test.go:583: Mkdir failed: mkdir /tmp2: file exists
    kprobe_test.go:584:      
```

Which points to:
https://github.com/cilium/tetragon/blob/0c08c89c09b60043be5c0c6e8776619cb3d45e9a/pkg/sensors/tracing/kprobe_test.go#L578-L584

We should probably fix this since the test is skipped. I think the best option would be to create a temp directory (https://pkg.go.dev/os#MkdirTemp) for each test and remove it afterwards.

jrfastab:
fixed here. https://github.com/cilium/tetragon/pull/255/commits/4351d2c688169e7ed3f2d631ca1b02ed8f5d0077

PR #255, opened 2022-07-21
Create temp directories for all tests in kprobes

tpapagian:
Before that tests that fail to create a new directory (i.e. already exists) just skipped.

FIXES: https://github.com/cilium/tetragon/issues/238

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -35,8 +35,6 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-var mountPath = "/tmp2"
-
 const (
 	testConfigFile = "/tmp/tetragon.gotest.yaml"
 	kprobeTestDir  = "/sys/fs/bpf/testObserver/"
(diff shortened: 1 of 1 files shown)


pkg/sensors/tracing/kprobe_test.go
```diff
@@ -535,29 +533,14 @@ spec:
...
-func getOpenatMntChecker() ec.MultiEventChecker {
+func getOpenatChecker(dir string) ec.MultiEventChecker {
 	kpChecker := ec.NewProcessKprobeChecker().
 		WithFunctionName(sm.Full("__x64_sys_openat")).
 		WithArgs(ec.NewKprobeArgumentListMatcher().
 			WithOperator(lc.Ordered).
 			WithValues(
 				ec.NewKprobeArgumentChecker().WithIntArg(-100),
-				ec.NewKprobeArgumentChecker().WithStringArg(sm.Full(mountPath+"/testfile")),
+				ec.NewKprobeArgumentChecker().WithStringArg(sm.Full(filepath.Join(dir, "testfile"))),
```
willfindlay, line 543: This will fail if someone passes a non-absolute path here. I would suggested using a suffix matcher here, possibly stripping the leading periods if they exist.

willfindlay (approved):
Looks good, just one minor bit from me.

pkg/sensors/tracing/kprobe_test.go
tpapagian: Good catch! So this will fail in the case where we use `getOpenatChecker()` in a testcase where we don't use the correct path in `__x64_sys_openat`. I propose to just be aware of that (i.e. add a comment/issue) when we add new tests. 

I believe that we need more work to make that generic enough to handle all cases in `__x64_sys_openat`. WDYT?

willfindlay: Documenting the limitation is a fine approach IMO.

kkourt (approved):

jrfastab:
@tpapagian follow up with some documentation per review where appropriate.

PR #256, opened 2022-07-21
VM multi-kernel testing using little-vm-helper

kkourt:
This PR adds multi-kernel testing to tetragon.

We use https://github.com/cilium/little-vm-helper/ and https://github.com/cilium/little-vm-helper-images to implement this.
The PR builds functionality gradually. This both because this is how it was developed, but also to ease reviewing. 
  
TODO:
 * [x] remove --force-rebuild-image and always rebuid the image
 * [x] list results for each test (rather than each package)
 * [x] use `tests/vmtests/` instead of `testing/`
 * [x] add ignore file pattern
      * ignore md files (docs) for now 
  * [x] add 4.19 kernel: requires btf image

Followups:
  * [ ] try out MacOS GH actions runners
  * [ ] manually specify tests to be skipped (can be done in the split program)
  * [ ] getting e2e-framework running on lvh


tools/split-tetragon-gotests/main.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+var CiBlacklist = []vmtests.GoTest{
+	// pkg.exporter has a rate limit test, which is time-dependent. There
+	// was a previous attempt to fix the test, but failed. Ignore it for
+	// now.
+	{PackageProg: "pkg.exporter"},
+	// https://github.com/cilium/tetragon/issues/247
+	{PackageProg: "pkg.sensors.tracing", Test: "TestCopyFd"},
+}
```
willfindlay, line 54: :heart:

tools/split-tetragon-gotests/main.go
```diff
@@ -0,0 +1,91 @@
...
+		"Usage: %s [flags] <ngroups> -- split tests in <ngroups> groups. Output is %s/test-group-X.\n",
+		os.Args[0], outDir,
+	)
+	flag.PrintDefaults()
+}
+
+func main() {
+
+	var cirun bool
+	flag.BoolVar(&cirun, "ci-run", false, "This option enables CI blacklist for tests")
```
willfindlay, line 68: Could we add a flag here for manually specifying skipped tests too? I'm specifically thinking of splitting e2e-framework into a separate job when that lands for example.

pkg/alignchecker/alignchecker_test.go
```diff
@@ -7,12 +7,14 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	tus "github.com/cilium/tetragon/pkg/testutils/sensors"
```
willfindlay, line 11: Just a nit but I think I would prefer testutils as an import name here over tus.

cmd/protoc-gen-go-tetragon/eventchecker/multichecker.go
```diff
@@ -140,33 +140,38 @@ func generateUnorderedEventChecker(g *protogen.GeneratedFile) error {
             return true, nil
         }
 
+        totalMatched := checker.totalChecks - pending
         if logger != nil {
-            logger.Infof("UnorderedEventChecker: %d/%d matched", checker.totalChecks-pending, checker.totalChecks)
+            logger.Infof("UnorderedEventChecker: starting checks with %d/%d total matched", totalMatched, checker.totalChecks)
```
willfindlay, line 145: To me "starting checks" here would imply the checker is starting for the first time, not running checks on a specific event. Maybe consider rewording this somehow?

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -645,6 +646,8 @@ func testKprobeObjectFiltered(t *testing.T,
 	assert.NoError(t, err)
 	err = jsonchecker.JsonTestCheck(t, checker)
 	if expectFailure {
+		t.Logf("Expectinfg failure")
```
willfindlay, line 652: Typo

```suggestion
		t.Logf("Expecting failure")
```

testing/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+There are two go programs: one that runs inside the VM (tetragon-tester), and one that runs outside
+the VM (tetragon-vmtests-run). tetragon-vmtests-run, running outside, will prepare an image and then
+boot it using qemu. The image will be build based on the base image provided with --base.
+Some of the steps of preparing the image are:
+ * add tetragon-tester as a systemd service, so that it starts when the machine boots
+ * mount the teragon source directory inside the VM (currently, as a 9p filesystem)
+ * write a configuration file to be read by tetragon-tester
+
+Once the machine boots,
```
willfindlay, line 72: Missing extra text here?

willfindlay (commented):
Some comments/nits from me. Overall looks great! I'll circle back and approve once you have a chance to ack this.

willfindlay:
@kkourt On the Cilium side they opted to use a MacOS runner as an intermediary stopgap. What do you think about trying that rather than the split test approach? Maybe we can give it a try and compare timings?

willfindlay:
Also maybe we can rename the testing directory to something more specific like vmtests? From your commit messages, it sounds like it is specific to VM testing. WDYT?

.github/workflows/vmtests.yml
```diff
@@ -0,0 +1,132 @@
...
+        path: go/src/github.com/cilium/tetragon/testing/vmtests-results-*
+        retention-days: 5
+
+    - name: Upload test results on failure
+      if: failure()
+      uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8
+      with:
+        name: tetragon-vmtests-${{ matrix.kernel }}-${{ matrix.group }}-results
+        path: go/src/github.com/cilium/tetragon/testing/vmtests-results-*
+        retention-days: 5
```
willfindlay, line 124: Duplicated?

.github/workflows/vmtests.yml
```diff
@@ -0,0 +1,132 @@
...
+
+    - name: Upload test results on cancelation
+      if: cancelled()
+      uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8
+      with:
+        name: tetragon-vmtests-${{ matrix.kernel }}-${{ matrix.group }}-results
+        path: go/src/github.com/cilium/tetragon/testing/vmtests-results-*
+        retention-days: 5
```
willfindlay, line 132: This could be folded into the above with `failure() || cancelled()`

willfindlay (commented):
A couple more things

tools/split-tetragon-gotests/main.go
kkourt: I think that makes sense. How about we do it as a followup when it is needed?

pkg/alignchecker/alignchecker_test.go
kkourt: You mean:
> testutils "github.com/cilium/tetragon/pkg/testutils/sensors"

Isn't that somewhat confusing since there is also a `testutils` package one level up?
There might even be files that need to import both "testutils"  and "testutils/sensors"

cmd/protoc-gen-go-tetragon/eventchecker/multichecker.go
kkourt: How about "starting event check(s)"?

kkourt:
> @kkourt On the Cilium side they opted to use a MacOS runner as an intermediary stopgap. What do you think about trying that rather than the split test approach? Maybe we can give it a try and compare timings?

I think it's worth a try. 

> rather than the split test approach

Even if we have a more efficient virtualization platform, I think we might need both eventually as the number of tests increase. So I would opt to keep the splitting.

.github/workflows/vmtests.yml
kkourt: ah, yes! thanks!

pkg/alignchecker/alignchecker_test.go
willfindlay: @kkourt or maybe just leave it as sensors? or maybe sensorutils?

cmd/protoc-gen-go-tetragon/eventchecker/multichecker.go
willfindlay: @kkourt maybe "checking event"?

willfindlay:
@kkourt I think I remember hearing that you can only have one active OSX runner per PR at a time. So we might need to end up picking one or the other. But let's try it :)

testing/README.md
```diff
@@ -0,0 +1,72 @@
...
+Run tests on 5.4, without KVM acceleration. Doing so, makes the run closer to the GH action
+environment (which do not support nested virtualization) and it  also uncover other issues due to
+the unconventional timing.
+
+```
+$ ./testing/tetragon-vmtests-run \
+	--kernel testing/test-data/kernels/5.4/boot/vmlinuz-5.4.206 \
+	--base testing/test-data/images/base.qcow2 \
+	--qemu-disable-kvm
+```
```
olsajiri, line 51: I'm getting fail on fedora:
```
$ ./testing/tetragon-vmtests-run --kernel testing/test-data/kernels/5.4/boot/vmlinuz-5.4.206 --base testing/test-data/images/base.qcow2 --qemu-disable-kvm
Error: required cmd 'mmdebstrap' not found
```
mmdebstrap seems to be debian specific tool, is this limited to debian or perhaps ubuntu?

willfindlay: It is specific to debian based distros. One workaround could be to package a Docker container with mmdebstrap installed.

willfindlay:
@kkourt #226 introduces a `tests/e2e/` directory. I would propose we unify things and have a `testing/vmtest/` where this PR lives and move `tests/e2e/` to `testing/e2e/`. That way we won't have both a `tests/` and `testing/` directory with no clear indication of what belongs where.

testing/README.md
kkourt: Ah good catch, thanks @olsajiri! `mmdebstrap` is not needed here, so I'll just remove the check in:
https://github.com/cilium/little-vm-helper/blob/9e7539fbeba9ef971f98c33e84282db7de3743ac/pkg/images/env.go#L16-L21

pkg/alignchecker/alignchecker_test.go
kkourt: How about tusensors? :)

kkourt:
> @kkourt #226 introduced a `tests/e2e/` directory. I would propose we unify things and have a `testing/vmtest/` where this PR lives and move `tests/e2e/` to `testing/e2e/`. That way we won't have both a `tests/` and `testing/` directory with no clear indication of what belongs where.

Makes sense to me.

pkg/alignchecker/alignchecker_test.go
kkourt: Actually, I think I'm going to keep it as is in this PR for consistency, since it already exists in the tree: https://github.com/cilium/tetragon/blob/7cadc51fbafb022fee35d972d4b7a83faf9e9440/pkg/sensors/exec/exec_test.go#L23. We can rename it later.

testing/README.md
kkourt: yap, fixed. Thanks!

pkg/alignchecker/alignchecker_test.go
willfindlay: sounds good

testing/README.md
kkourt: @olsajiri can you try again with latest? I believe everything should work now.

willfindlay:
Would also like to add looking into getting e2e-framework running on lvh as another follow-up.

willfindlay (approved):
LGTM, just left one minor comment.

.github/workflows/vmtests.yml
```diff
@@ -0,0 +1,133 @@
...
+        sudo chmod go+rX -R /boot/
+
+    - name: set panic on RCU stall
+      run: |
+        sudo sh -c 'echo 1 > /proc/sys/kernel/panic_on_rcu_stall'
+
+    - name: download build data
+      uses: actions/download-artifact@v3
+      with:
+         name: tetragon-build
```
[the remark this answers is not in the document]
kkourt: sounds sensible to me, but AFAICT there is no action to delete artifacts: https://github.com/actions/upload-artifact/issues/5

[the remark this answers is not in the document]
kkourt: Ah wait, there seems to be a REST API: https://docs.github.com/en/rest/actions/artifacts#delete-an-artifact

kkourt:
Converted to draft while trying out artifact deletion

.github/workflows/vmtests.yml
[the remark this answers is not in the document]
willfindlay: Maybe it's not even worth the effort then. Perhaps we could just drop retention days to 1?

kkourt:
> Converted to draft while trying out artifact deletion

Marking ready again. Could not find a way to delete the build artifact. Instead, we truncate them. (see latest patch).

.github/workflows/vmtests.yml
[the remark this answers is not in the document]
kkourt: I've added a workaround where the file is truncated (instead of deleted). Seems to work in a test run I did. Trying it out now with the full test.

.github/workflows/vmtests.yml
```diff
@@ -131,3 +131,17 @@ jobs:
...
+    steps:
+    - name: Create truncated build file
+      run: |
+        touch /tmp/tetragon.tar
+    - name: Upload truncated file
+      uses: actions/upload-artifact@v3
+      with:
+         name: tetragon-build
+         path: /tmp/tetragon.tar
+         retention-days: 1
```
willfindlay, line 147: amazing workaround

PR #257, opened 2022-07-21
eventchecker: fix broken codegen for map checker

michi-covalent:
The logic for determining type names from protobuf maps was incorrect because we were
subbing in the FieldDescriptor rather than the underlying type. Fix this for now to get
codegen working. A subsequent commit will fix the eventchecker itself so arbitrary map
key/values will actually work.


willfindlay (approved):
Just ship it man

PR #260, opened 2022-07-25
eventchecker: add support for checking map[string]string

willfindlay:
This commit adds support to eventchecker for checking maps of string -> string. Note that
other kinds of maps are not supported for now as this would require a non-trivial refactor
to the existing code. But that should be fine for now. We can always revist this later if
needed.


willfindlay:
flaky test struggle

kkourt (approved):

willfindlay:
Let's ship it man :rocket:

PR #261, opened 2022-07-26
tetragon: Add bench application

olsajiri:
Adding benchmark application to measure tracing workloads. 

The tetragon-bench application starts tetragon (or not for -baseline option)
for given CRD file and runs workload.

At the end it displays some stats, like:

```
Benchmark summary
-----------------
Test started:      2022-09-08 15:04:28.04513503 +0200 CEST m=+0.016539427
Test ended:        2022-09-08 15:04:38.268072909 +0200 CEST m=+10.239477312
Workload start:    2022-09-08 15:04:38.268072909 +0200 CEST m=+10.239477312
Arguments:         
Total duration:    10.222937885s
Setup duration:    1.743823005s
Workload duration: 8.479094472s
Test duration:     10.222937885s
Export duration:   0s
Export stats:      bytes=0, writes=0
FGS cpu usage:     system=15.170223s, user=17.522276s, rss=230504, ctxsw=1231909
Ring buffer:       received=3958406, lost=41842, errors=0
BPF statistics:
  event_execve                   [tracepoint/212]   :   8.89µs (2)
  event_wake_up_new_task         [kprobe/211]       :   1.41µs (6)
  generic_kprobe_event           [kprobe/205]       :   1.54µs (4001580)
  restrict_filesystems           [lsm/58]           :   0.13µs (1569)
  event_exit                     [tracepoint/210]   :   0.49µs (32)
```

It's possible to run predefined workloads with builtin CRD files,
or run specific custom benchmark.

The predefined benchmarks are rw and open and can be run with:

```
# ./tetragon-bench -trace rw
# ./tetragon-bench -trace open
```
They do and monitor some simple read/write open operations and they are there
mainly to show the possibility of builtin benchmark if we think it's good idea ;-)

The custom benchmark is what I'm using most at the moment and
it allows to pass CRD file to start tetragon with and specify workload, like:

```
# ./tetragon-bench -crd crds/examples/names.yaml  -- perf bench sched pipe
```

I'm doing perf profiles to check on CRD stuff with:

```
# perf record -g -e cycles:k ./tetragon-bench -crd ./read_write.yaml -json-encode -- perf bench sched pipe
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,7 +25,7 @@ GOLANGCILINT_VERSION = $(shell golangci-lint version 2>/dev/null)
 
 
 .PHONY: all
-all: tetragon-bpf tetragon tetra tetragon-alignchecker test-compile tester-progs protoc-gen-go-tetragon
+all: tetragon-bpf tetragon tetra tetragon-alignchecker test-compile tester-progs protoc-gen-go-tetragon tetragon-bench
 
 -include Makefile.docker
 
@@ -75,12 +75,15 @@ tetragon-bpf-container:
 verify: tetragon-bpf
 	sudo contrib/verify/verify.sh bpf/objs
 
-.PHONY: tetragon tetra tetragon-operator tetragon-alignchecker
+.PHONY: tetragon tetra tetragon-operator tetragon-alignchecker tetragon-bench
 tetragon:
 	$(GO) build -gcflags=$(GO_GCFLAGS) -ldflags=$(GO_LDFLAGS) -mod=vendor ./cmd/tetragon/
 
 tetra:
 	$(GO) build -gcflags=$(GO_GCFLAGS) -ldflags=$(GO_LDFLAGS) -mod=vendor ./cmd/tetra/
+
+tetragon-bench:
+	$(GO) build -gcflags=$(GO_GCFLAGS) -ldflags=$(GO_LDFLAGS) -mod=vendor ./cmd/tetragon-bench/
 
 tetragon-operator:
 	$(GO) build -gcflags=$(GO_GCFLAGS) -ldflags=$(GO_LDFLAGS) -mod=vendor -o $@ ./operator
--- a/pkg/bench/bench.go
+++ b/pkg/bench/bench.go
@@ -41,6 +41,8 @@ import (
 
 type Arguments struct {
 	Trace       string
+	Crd         string
+	CmdArgs     []string
 	Debug       bool
 	PrintEvents bool
 	JSONEncode  bool
--- a/pkg/bench/bench.go
+++ b/pkg/bench/bench.go
@@ -1,0 +1,253 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package bench
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+	"os/signal"
+	"path"
+	"sync"
+	"sync/atomic"
+	"syscall"
+	"time"
+
+	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/cilium/tetragon/pkg/api/readyapi"
+	"github.com/cilium/tetragon/pkg/bpf"
+	"github.com/cilium/tetragon/pkg/btf"
+	"github.com/cilium/tetragon/pkg/cilium"
+	"github.com/cilium/tetragon/pkg/config"
+	"github.com/cilium/tetragon/pkg/defaults"
+	"github.com/cilium/tetragon/pkg/exporter"
+	"github.com/cilium/tetragon/pkg/grpc"
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/observer"
+	"github.com/cilium/tetragon/pkg/option"
+	"github.com/cilium/tetragon/pkg/process"
+	"github.com/cilium/tetragon/pkg/reader/notify"
+	"github.com/cilium/tetragon/pkg/sensors"
+	"github.com/cilium/tetragon/pkg/sensors/base"
+	"github.com/cilium/tetragon/pkg/watcher"
+
+	// Imported to allow sensors to be initialized inside init().
+	_ "github.com/cilium/tetragon/pkg/sensors/exec"
+	_ "github.com/cilium/tetragon/pkg/sensors/tracing"
+)
+
+type Arguments struct {
+	Trace       string
+	Debug       bool
+	PrintEvents bool
+	JSONEncode  bool
+	Baseline    bool
+}
+
+func readConfig(file string) (*config.GenericTracingConf, error) {
+	if file == "" {
+		return nil, nil
+	}
+
+	yamlData, err := os.ReadFile(file)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read yaml file %s: %w", file, err)
+	}
+	cnf, err := config.ReadConfigYaml(string(yamlData))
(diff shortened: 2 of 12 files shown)


pkg/bench/bench.go
```diff
@@ -0,0 +1,243 @@
...
+	} else {
+		exePath, err := os.Executable()
+		if err != nil {
+			log.Fatal(err)
+		}
+		option.Config.HubbleLib = path.Join(path.Dir(exePath), "bpf/objs")
+
+		if _, err := os.Stat(option.Config.HubbleLib); err != nil {
+			// Running outside the source tree, fall back to default location.
+			option.Config.HubbleLib = "/var/lib/hubble-fgs"
```
tixxdz, line 93: Maybe this one: https://github.com/cilium/tetragon/blob/main/pkg/defaults/defaults.go#L44 ?

pkg/bench/bpfstats.go
```diff
@@ -0,0 +1,78 @@
...
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+	"os/exec"
+	"time"
+)
+
+func EnableBpfStats() {
+	err := os.WriteFile("/proc/sys/kernel/bpf_stats_enabled", []byte("1"), 0666)
```
tixxdz, line 16: Is the perm mode ok? maybe restrict it a bit

pkg/bench/cpuusage.go
```diff
@@ -0,0 +1,149 @@
...
+	cpuUsage.SystemTime = timevalToDuration(rusage.Stime)
+	cpuUsage.MaxRss = rusage.Maxrss
+	cpuUsage.ContextSwitches = rusage.Nivcsw + rusage.Nvcsw
+	return
+}
+
+func CPUUsageFromCPUAcct(containerID string) CPUUsage {
+
+	rss := int64(0)
+	memStatFilename := fmt.Sprintf("/sys/fs/cgroup/memory/docker/%s/memory.stat", containerID)
```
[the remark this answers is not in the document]
tixxdz: btw where this is called?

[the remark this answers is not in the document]
tixxdz: The full cgroup path may not container docker directory at all or another container runtime, I'm missing context, but if you could read /proc/self/cgroups and adapt according to cgroup version and the controllers you are interested in to get right working paths without hardcoding it here. There is some logic in that cgroups.go package

pkg/bench/bench.go
olsajiri: ok, will use that one, thanks

pkg/bench/bpfstats.go
olsajiri: that permission will be used only if the file does not exist and needs to be created, so we are fine here

pkg/bench/cpuusage.go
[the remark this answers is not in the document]
olsajiri: right, I'll remove it for now, because it's not used and check on that more, thanks

pkg/bench/summary.go
```diff
@@ -0,0 +1,109 @@
...
+
+	"github.com/cilium/tetragon/pkg/metrics/ringbufmetrics"
+	"github.com/fatih/color"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
+)
+
+// Summary gathers benchmark results. Serializes to JSON.
+// This is updated from multiple places concurrently, but currently
+// there is no overlap on writes, so this isn't yet protected by a mutex.
```
kkourt, line 21: Are the concurrent writes all happen in BpfStats?

pkg/bench/summary.go
```diff
@@ -0,0 +1,109 @@
...
+	RunTime            time.Time
+	SetupDurationNanos time.Duration
+	TestDurationNanos  time.Duration
+
+	JSONEncodingDurationNanos time.Duration
+	ExportStats               CountingDiscardWriter
+
+	TetragonCPUUsage CPUUsage
+
+	BpfStats map[int64]*BpfProgStats
```
kkourt, line 38: Is CPU id the key here?

pkg/bench/bench.go
```diff
@@ -0,0 +1,241 @@
...
+	cnf, err := config.ReadConfigYaml(string(yamlData))
+	if err != nil {
+		return nil, err
+	}
+
+	return cnf, nil
+}
+
+func (args *Arguments) String() string {
+	return fmt.Sprintf("")
```
kkourt, line 70: Is this intentional?

Maybe something like `fmt.Sprintf("%+v", args)` would be better?

pkg/bench/cpuusage.go
```diff
@@ -0,0 +1,64 @@
...
+func (cu CPUUsage) Sub(cu2 CPUUsage) CPUUsage {
+	cu.UserTime -= cu2.UserTime
+	cu.SystemTime -= cu2.SystemTime
+	cu.ContextSwitches -= cu2.ContextSwitches
+	return cu
+}
+
+func (cu CPUUsage) Add(cu2 CPUUsage) CPUUsage {
+	cu.UserTime += cu2.UserTime
+	cu.SystemTime += cu2.SystemTime
```
kkourt, line 58: `Sub` has a `ContextSwitches` as well? maybe it's needed here as well?

pkg/bench/trace.go
```diff
@@ -0,0 +1,105 @@
...
+import (
+	"context"
+	"log"
+	"os"
+	"strings"
+	"time"
+)
+
+type TraceBench interface {
+	Crd(ctx context.Context, args *Arguments) string
```
kkourt, line 15: What is this method supposed to do? Maybe we can come up with a more informative name?

Based on the caller, it returns the configuration filename? Maybe `ConfigFilename()`?

Makefile
```diff
@@ -75,13 +75,16 @@ tetragon-bpf-container:
 verify: tetragon-bpf
 	sudo contrib/verify/verify.sh bpf/objs
 
-.PHONY: tetragon tetra tetragon-operator tetragon-alignchecker
+.PHONY: tetragon tetra tetragon-operator tetragon-alignchecker bench
```
kkourt, line 78: How about we name the binary `tetragon-bench`?

kkourt (changes requested):
Thanks! Some minor requests / suggestions.

Makefile
kkourt: Could we also add the binary to `.gitignore`?

olsajiri: yes, good idea, thanks

pkg/bench/trace.go
olsajiri: right, it gives back the CRD file path. will rename that

pkg/bench/cpuusage.go
olsajiri: good catch, it's a bug, will fix

pkg/bench/bench.go
olsajiri: leftover, will add, thanks

pkg/bench/summary.go
olsajiri: it's the ExportStats that are updated from tetragon process,
rest should be done just in bench process, I'll update the comment

pkg/bench/summary.go
kkourt: Is see. Could you please add a comment about what the key is?

olsajiri: yep, will do

kkourt (approved):
Thanks!

PR #263, opened 2022-07-26
chore: fix vendor target and run make vendor

willfindlay:
We just merged #260 but it seems that the CI job was out of sync with main and did not account for the new module split. So let's update the make vendor target to make this situation impossible. See individual commits.


michi-covalent (approved):

willfindlay:
@kkourt I'm going to merge this since it's fairly urgent and I got an ack from @michi-covalent

PR #264, opened 2022-07-26
exec: fix cache for pod info

willfindlay:
After 4dac2750a, exec.go was adding events to the wrong cache. This was causing missing
pod info for essentially all events. Fix this by using the execCache again rather than the
eventCache.


willfindlay:
Planning on unifying the two caches but I will do that as a follow-up. For now let's fix the regression.

michi-covalent (approved):

PR #265, opened 2022-07-26
tests/e2e: add an end-to-end tests for pod labels

willfindlay:
Add an end-to-end test for the new PodLabels field of Process.Pod which verifies that we
see at least one PROCESS_EXEC event with each of the expected labels per pod.


michi-covalent (approved):
magical

willfindlay:
Now using the new and improved demo app :eyes:

PR #266, opened 2022-07-27
tetragon: unify eventcache and execcache

willfindlay:
It wasn't necessary to split execcache and eventcache so let's fold it all into eventcache. See individual commits.


jrfastab:
I like the direction so I'll merge it I see at least a couple issues remaining but will merge this and then work out any fallout with @tpapagian in the next days.

PR #270, opened 2022-07-29
filters: implement pod regex filter + misc fixes

willfindlay:
This PR implements a new filter for filtering by pod name using a regex. Note that we
had to use number 7 here and shift everything else over by one. This is technically
a breakage but needs to be done for historical compatibility reasons.

Also rolled in some fixes for the eventchecker.


jrfastab:
Its not breaking anything we haven't release yet.

michi-covalent (approved):

PR #273, opened 2022-08-01
tetragon: Fixed missing grpc port expose via helm install

pxp928:
Currently the tetragon service does not expose the grpc port when it is enabled via values.yaml. This checks to see if the grpc is enabled and exposes the port via the service.

Signed-off-by: pxp928 <parth.psu@gmail.com>

### DIFF
--- a/install/kubernetes/templates/service.yaml
+++ b/install/kubernetes/templates/service.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.tetragon.prometheus.enabled -}}
+{{- if or (eq .Values.tetragon.prometheus.enabled true) (eq .Values.tetragon.grpc.enabled true) }}
 ---
 apiVersion: v1
 kind: Service
@@ -13,10 +13,18 @@ metadata:
   namespace: {{ .Release.Namespace }}
 spec:
   ports:
+{{- if .Values.tetragon.prometheus.enabled }}
     - name: metrics
       port: 2112
       protocol: TCP
       targetPort: {{ .Values.tetragon.prometheus.port }}
+{{- end }}
+{{- if .Values.tetragon.grpc.enabled }}
+    - name: grpc
+      port: {{ .Values.tetragon.grpc.port }}
+      protocol: TCP
+      targetPort: {{ .Values.tetragon.grpc.port }}
+{{- end }}
   selector:
     {{- with .Values.daemonSetLabelsOverride}}
     {{- toYaml . | nindent 4 }}


michi-covalent (approved):

willfindlay (changes requested):
I would strongly suggest putting this behind a separate configuration option with clear documentation on the benefits/risks if we are going to do this at all. The reason being it is quite scary to expose all the information Tetragon gathers on an open service port by default. This could for example be used to leak sensitive information. cc @tixxdz @sharlns

pxp928:
@willfindlay Correct me if I am wrong. New to tetragon. But you can limit which address the grpc is exposed via the values.yaml. So only the specified address would be able to connect and all other unauthorized addresses would not be able to connect.

michi-covalent:
i think @willfindlay 's concern is that the traffic is not encrypted, so if you expose the grpc endpoint as a service, anybody on the network can connect to it. that's why currently we only listen on the local interface by default. in cilium / hubble, we use mTLS to ensure that only hubble-relay can connect to cilium: https://docs.cilium.io/en/stable/concepts/observability/hubble-configuration/

stepping back a little bit, could you describe your use case? maybe we can come up with something that doesn't require exposing plaintext grpc endpoint.

pxp928:
@michi-covalent So I am creating an integration with another tool where I want to filter on a specific pod in a specific namespace and create a record of the set of execs, syscalls, etc. (whatever you add to your tracing). I agree with your point, any suggestions on how we can proceed? The other tool that I would be integrating with would be in the same cluster as tetragon so there should be no reason to expose outside of the cluster.

pxp928:
Would it make more sense to expose grpc via a unix socket instead?

willfindlay:
> Would it make more sense to expose grpc via a unix socket instead?

Yes, we've actually had offline discussions about this before but nobody has gotten around to it yet. Would be happy to accept a PR for this, although ideally it would be configurable with the Unix socket as a default imo.

sharlns (changes requested):
It sounds good if we make it configurable and default to unix socket, but also allow running with TCP socket.

jrfastab:
@pxp928 looks like there is some agreement here I'll wait for an update to (re)review and hopefully merge.

kkourt:
Since this PR has not seen updates for sometime, I'll move it into draft for now. @pxp928 feel free to mark it as ready for review whenever it is ready. Thanks!

mtardy:
This has been stale for 3 years, I'll close it for now.

PR #274, opened 2022-08-01
tetragon: eventcache Needed() is incomplete

jrfastab:
Resolve obvious cases for Needed() where we have a process pointer, but
are missing binary and other information. This happens through common
paths for example when exit events are received before their exec event.

To fix just extend the Needed() check to search for Binary name. I
introduced this while preparing initial release and doing some
refactoring.

Fixes: 931b70f2c9878 "Tetragon: Cilium developers happily contribute Tetragon"


jrfastab:
@tpapagian has extra tests here and infra to further refine this logic, but I'm likely to merge this assuming it passes CI to fix the obvious cases.

jrfastab:
just an illustration.

PR #276, opened 2022-08-02
eventcache: use correct unix epoch value for start time

willfindlay:
eventcache wasn't compiling because we were trying to access p.StartTime.Value which
doesn't exist. What we want here is the translation to Unix epoch, so let's add a helper
here that can do this.


willfindlay:
Looks like this was already covered in another PR, closing.

PR #277, opened 2022-08-02
hubble-fgs: event cache doesn't use endpoint for anything drop the query

willfindlay:
Collecting enpoint does not do anything here so lets drop it for now.
Presumably, we lost the code that used this in one of our refactors so
we can add it back if we have a need for it later.

We also need to update the skeleton e2e test since we are now receiving more events than
initially expected which would cause the test to cross the event limit. To fix this, bump
the limit up to a conservative 100 events.


tpapagian (approved):
Great, thanks!

willfindlay:
This was rolled into #218 so I'm closing it

PR #278, opened 2022-08-02
metrics/eventmetrics: include exec id in event count

willfindlay:
Add an extra label to event count metrics so that we can include exec id of a process.
There is a trade-off here in that this change will introduce extra noise into the metrics
(i.e. we would no longer aggregate multiple runs of a single binary under the same
counter). This can be useful for debugging as well as triaging any suspicious events in
production. Since this can be quite noisy, hide it behind a configuration flag that is
disabled by default.

### DIFF
--- a/cmd/tetragon/flags.go
+++ b/cmd/tetragon/flags.go
@@ -56,6 +56,8 @@ const (
 	keyExportDenylist  = "export-denylist"
 
 	keyNetnsDir = "netns-dir"
+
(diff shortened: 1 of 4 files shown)


willfindlay:
Marking this as draft for now, considering alternative approaches that might be less noisy.

willfindlay:
On second thought this really isn't that useful and is probably not an idiomatic way to use metrics. Closing for now.

PR #279, opened 2022-08-02
Remove 'packed' attribute where possible

kevsecurity:
Applying 'packed' attribute to structs causes clang to write to the
members byte-by-byte, as offsets may not be aligned. This is bad for
performance, instruction count and complexity, so don't apply this
attribute to structs where members are correctly aligned already
(e.g. by padding, layout).


kevsecurity:
Seeing some issues with testing. Needs further investigation before merging.

PR #281, opened 2022-08-02
Add test for tracing bpf_check

sarahfujimori:
This is a unit test for tracing bpf verifier checks during program loads. We do this by first loading a CRD that traces the function bpf_check, then loading a second CRD and checking for events that record the correct program name and type. Calling the function GetDefaultObserverWithFile twice causes errors since we don’t want to create the fake k8s watcher, load the exporter, etc. again. Instead, for the second CRD, we only get and load the sensors. This approach requires importing a few extra packages, so it may be useful to find a better way to do this.

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -19,14 +19,19 @@ import (
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
 	"github.com/cilium/tetragon/pkg/bpf"
+	yaml "github.com/cilium/tetragon/pkg/config"
 	"github.com/cilium/tetragon/pkg/jsonchecker"
 	"github.com/cilium/tetragon/pkg/kernels"
 	bc "github.com/cilium/tetragon/pkg/matchers/bytesmatcher"
 	lc "github.com/cilium/tetragon/pkg/matchers/listmatcher"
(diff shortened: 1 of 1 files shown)


willfindlay:
If you're concerned about bloat maybe we could just use cilium/ebpf to load a stub kprobe and attach it?

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -2300,3 +2305,90 @@ func TestKprobeMatchArgsFdPrefix(t *testing.T) {
...
+	if cnf != nil {
+		var err error
+		sens, err = sensors.GetSensorsFromParserPolicy(&cnf.Spec)
+		if err != nil {
+			return err
+		}
+	}
+
+	b := base.GetInitialSensor()
+	if err := b.Load(context.TODO(), option.Config.BpfDir, option.Config.MapDir, option.Config.CiliumDir); err != nil {
```
jrfastab, line 2331: Looks good to me though should we also unload the program here in the flow?

willfindlay: Would also be good to replace context.TODO()s here with a single context.Background() like we recently did for other tests.

sarahfujimori: the test logs show that the other program (sys_write kprobe) is already being unloaded at the end, so I think that part is fine..

willfindlay (approved):
Looks good to me. I think it's OK to use sensor infra for running the test workload here, at least for now. Can always change our minds later if we want.

PR #282, opened 2022-08-03
tetragon: Update eventcache on New() call and cleanup old state

jrfastab:
Currently, if eventcache.New() is called multiple times the old cache
will continue to be used. But, this is problematic if the caller is
trying to configure the cache with new parameters, such as duration
or exporter.

Specifically this is the case in the unit tests where each Test* is
creating a new process manager and exporter using a new file so that
we have specific logs for each test. But, the current behavior means
any events going through the eventcache (after the first test) are
then lossed because they attempt to export into the old filename.

To fix simply stop the old go() on New() and reconfigure it. When
I ran this locally forcing all events through the cache I am now
able to PASS multiple unit tests.

### DIFF
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -44,6 +44,7 @@ type cacheObj struct {
 
 type Cache struct {
 	objsChan chan cacheObj
+	done     chan bool
 	cache    []cacheObj
 	server   *server.Server
 	dur      time.Duration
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):
it's probably better to get rid of the package variable `cache` but it's not directly related to this pr. approved 🚢

jrfastab:
Yeah I guess that cache variable has become fairly useless... tbd.

Issue #247, opened 2022-07-19
ci flake: kprobe_copyfd_test missing all expected events


willfindlay:
Just got a CI flake in https://github.com/cilium/tetragon/runs/7410818480?check_suite_focus=true#step:7:10985 where we are not seeing any of the expected events in kprobe_copyfd_test. // cc @tpapagian 

Here is the Tetragon json log from the test: [tetragon-json.zip](https://github.com/cilium/tetragon/files/9142224/tetragon-json.zip)

tpapagian:
During the testing of https://github.com/cilium/tetragon/pull/235 I have seen that there are some issues with close system call and unfollowfd. I believe that this is also the case here. 

I am assigning it to myself.

willfindlay:
@tpapagian just adding some more info here. I think something recent may have made this flake way more common. Just saw it again only a few CI runs later.

tpapagian:
I believe https://github.com/cilium/tetragon/pull/235/commits/d6cdbaab549519f67e0d846c95dcfa04e496023c should fix this issue. So let's wait for https://github.com/cilium/tetragon/pull/235 to be merged.

willfindlay:
@tpapagian it's not fixed :( here's another failure https://github.com/cilium/tetragon/runs/7469979029?check_suite_focus=true

Forsworns:
I met this problem too... I guess the problem comes from: 

---

@willfindlay You add an event check in #240. So that the event checker would fail on misaligned events.

Take [this failure](https://github.com/cilium/tetragon/runs/7469979029?check_suite_focus=true#step:7:211) as an example.

The `pkg\sensors\tracing\kprobe_copyfd_test.go` initialize a `ProcessKprobeChecker` to trace krpobe events on `__x64_sys_read`. 

Sadly, Tetragon defines two tracepoint events `ProcessExec` and `ProcessExit` and traces them by default. Then the checker found captured events are `ProcessExec` (maybe from executable binaries like `contrib\tester-progs\dup-tester.c`). 

Maybe we should disable the default tracing during tests, or at least disable them in configuration file like `testdata\specs\copyfd-__fd_install.yaml.tmpl`?

---

I'm not sure since I didn't read the codes for event checkers in detail....  Just provide a possible reason for these failures

willfindlay:
@Forsworns I think it's unrelated unforunately. This test has been flaky for ages, although recently we started seeing these flakes more frequently.

[...]

PR #283, opened 2022-08-03
Add a unit test for missing exec event in the eventcache

tpapagian:
This unit-test emulates the case where we get an exit event but we miss the exec event. In that case, we should wait for the cache iterations to end and then produce an exit event with partial process info. In that case we only have the process start time and the process pid.

This PR also fixes https://github.com/cilium/tetragon/issues/285 and https://github.com/cilium/tetragon/issues/247. OOO events was related to these issues and thus I don't use a separate PR for them. To test this I have run 8 successful consecutive runs in the CI while before a kprobe flake appears every ~3 runs. 

### DIFF
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -19,14 +19,11 @@ import (
 	"github.com/cilium/tetragon/pkg/server"
 )
 
-// garbage collection states
 const (
-	threeStrikes = 3
-)
-
-// garbage collection run interval
-const (
-	eventRetryTimer = time.Second * 10
+	// garbage collection retries
+	cacheStrikes = 15
+	// garbage collection run interval
+	eventRetryTimer = time.Second * 2
 )
 
 var (
@@ -108,7 +105,7 @@ func (ec *Cache) handleEvents() {
 
 		if err != nil {
 			event.color++
-			if event.color < threeStrikes {
+			if event.color < cacheStrikes {
 				tmp = append(tmp, event)
 				continue
 			}
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -21,9 +21,9 @@ import (
 
 const (
 	// garbage collection retries
-	cacheStrikes = 15
+	CacheStrikes = 15
 	// garbage collection run interval
-	eventRetryTimer = time.Second * 2
+	EventRetryTimer = time.Second * 2
 )
 
 var (
@@ -105,7 +105,7 @@ func (ec *Cache) handleEvents() {
 
 		if err != nil {
 			event.color++
-			if event.color < cacheStrikes {
+			if event.color < CacheStrikes {
 				tmp = append(tmp, event)
 				continue
 			}
@@ -129,7 +129,7 @@ func (ec *Cache) loop() {
 	for {
 		select {
 		case <-ticker.C:
-			/* Every 'eventRetryTimer' walk the slice of events pending pod info. If
+			/* Every 'EventRetryTimer' walk the slice of events pending pod info. If
 			 * an event hasn't completed its podInfo after two iterations send the
 			 * event anyways.
 			 */
@@ -198,7 +198,7 @@ func NewWithTimer(s *server.Server, dur time.Duration) *Cache {
 }
 
 func New(s *server.Server) *Cache {
-	return NewWithTimer(s, eventRetryTimer)
+	return NewWithTimer(s, EventRetryTimer)
 }
 
 func Get() *Cache {
--- a/pkg/process/cache.go
+++ b/pkg/process/cache.go
@@ -52,7 +52,8 @@ func (pc *Cache) cacheGarbageCollector() {
 			select {
 			case <-pc.stopChan:
 				ticker.Stop()
-				pc.Purge()
+				pc.cache.Purge()
+				pc.pidMap.Purge()
 			case <-ticker.C:
 				newQueue = newQueue[:0]
 				for _, p := range deleteQueue {
--- a/pkg/grpc/exec/exec_test.go
+++ b/pkg/grpc/exec/exec_test.go
@@ -24,6 +24,10 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
+const (
+	cacheTimerMs = 100
+)
+
 var (
 	AllEvents []*tetragon.GetEventsResponse
 )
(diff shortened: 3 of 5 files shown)


pkg/api/readyapi/readyapi.go
```diff
@@ -9,3 +9,9 @@ type MsgTETRAGONReady struct{}
 func (msg *MsgTETRAGONReady) HandleMessage() *tetragon.GetEventsResponse {
 	return nil
 }
+
+func (msg *MsgTETRAGONReady) SetCached() {}
```
willfindlay, line 13: Maybe a comment here to explain that MessageReady.IsCached() should always be true and that's why we're leaving this blank?

willfindlay (approved):
Just a minor comment from me.

jrfastab:
I'm somehow missing how a msg in the eventcache gets 'readded' to the cache. The eventcache kicks the event out to the NotifyListeners these can do an Add() in the eventcache again?

jrfastab:
Aha so fairly sure the issue here is the dummyInterface is calling HandleMessage which does not reflect what the actual code does run from main does. The flow is `HandleMessage -> Exporter` with some constructors through processManager to add the exporter as a listener and then server.go used to construct the exporter correctly. To follow along the relevant files are:

 pkg/exporter/exporter.go: Implements the Send() method and Start() methods to satisfy the interface
 pkg/server/server.go: Embeds into exporter above and used by exporter to do aggregation and filtering
 pkg/grpc//process_manager.go: Has the Add logic to add multiple listeners. This lets us run both gRPC listeners, dbg listeners and json writers.

jrfastab (changes requested):
Fixing dummy notifier should resolve issue and then we don't need patch 3.

pkg/eventcache/eventcache.go
```diff
@@ -19,14 +19,11 @@ import (
...
 
-// garbage collection states
 const (
-	threeStrikes = 3
-)
-
-// garbage collection run interval
-const (
-	eventRetryTimer = time.Second * 10
+	// garbage collection states
```
kkourt, line 24: nit: is states a typo?

pkg/eventcache/eventcache.go
```diff
@@ -21,9 +21,9 @@ import (
```
kkourt, line 22: >  Make eventRetryTimer and cacheStrikes from eventcache public 

Maybe add something like:

> 
> We do this so that we can use the value in the json checker and in testing. See next two commit for details.

in the commit message?

kkourt (approved):
Thanks! 

Nice catch on the timing issue. I believe this will fix the linked issues.

tpapagian:
> Nice catch on the timing issue. I believe this will fix the linked issues.

@kkourt It seems that it still have some flakes related to these issues. This makes things better but I believe that it does not solve this entirely. Working on that (this is the reason that this PR is in draft).

pkg/eventcache/eventcache.go
```diff
@@ -99,20 +98,25 @@ func (ec *Cache) handleEvents() {
...
 			nspid := getNsPid.GetNsPid()
 			err = handleExecEvent(&event, nspid)
+			callFn = 1
 		} else {
 			err = handleEvent(&event)
+			callFn = 2
 		}
 
 		if err != nil {
+			logger.GetLogger().WithError(err).Warnf("handleEvents() - %d - %d", callFn, event.color)
```
willfindlay, line 112: I think this is going to spam our logs with a lot of messages, since we expect err to not be nil quite frequently. I would suggest moving it to Debugf

pkg/process/cache.go
```diff
@@ -112,6 +112,7 @@ func (pc *Cache) cacheGarbageCollector() {
 }
 
 func (pc *Cache) deletePending(process *ProcessInternal) {
+	logger.GetLogger().Warnf("DELETE EXEC - PID: %d - TIME: %d - EXECID: %s", process.process.Pid.Value, process.process.ExecId)
```
willfindlay, line 115: Do we expect this to happen frequently? If so let's move it to Debugf imo

pkg/process/cache.go
```diff
@@ -127,6 +128,7 @@ func (pc *Cache) refInc(p *ProcessInternal) {
 }
 
 func (pc *Cache) Purge() {
+	logger.GetLogger().Warnf("procCache Purge")
```
willfindlay, line 131: Same question as above

pkg/eventcache/eventcache.go
```diff
@@ -80,6 +78,7 @@ func handleEvent(event *cacheObj) error {
 	// now because it should be available. Otherwise we have a valid
 	// process and lets copy it across.
 	if event.internal == nil {
+		logger.GetLogger().Warnf("handleEvent() - PID: %d - TIME: %d - EXECID: %s", p.Pid.Value, event.timestamp, process.GetProcessID(p.Pid.Value, event.timestamp))
```
willfindlay, line 81: Do we expect this to happen frequently? If so let's move it to Debugf imo

pkg/eventcache/eventcache.go
tpapagian: Just trying to debug the issue here. I don't plan to keep them when the issue is resolved. (Same for the other log messages)

tpapagian:
> Fixing dummy notifier should resolve issue and then we don't need patch 3.

Fixed dummy notifier and addressed the other comments. Ready for review now.

pkg/eventcache/eventcache.go
```diff
@@ -19,14 +19,11 @@ import (
...
-	threeStrikes = 3
-)
-
-// garbage collection run interval
-const (
-	eventRetryTimer = time.Second * 10
+	// garbage collection retries
+	cacheStrikes = 15
+	// garbage collection run interval
+	eventRetryTimer = time.Second * 2
```
jrfastab: The 10seconds was about balancing GC performance cost vs getting events out. We should add a configMap option for this I think.

jrfastab:
This all looks good to me. With one question about making the time to retry cache configurable. For testing and watching results over grpc or cmd line its nice to have the system be more reactive, but when writing to logs for offsystem analysis I'm not sure it matters much if its 2 seconds or 2minutes.

jrfastab (approved):

pkg/eventcache/eventcache.go
tpapagian: Created https://github.com/cilium/tetragon/issues/296 to keep track of that.

PR #291, opened 2022-08-05
eventcachemetrics: fix eventcache and add error metrics

willfindlay:
Minor fixes to event cache handlers so that they deal with missing pod info separately
from missing process info.

Also update the event cache to increment pod and process info errors by event type when we
hit our cache strikes limit.

### DIFF
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -4,7 +4,7 @@
 package eventcache
 
 import (
-	"fmt"
+	"errors"
 	"time"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
@@ -47,6 +47,11 @@ type Cache struct {
 	dur      time.Duration
 }
 
+var (
+	ErrFailedToGetPodInfo     = errors.New("failed to get pod info")
+	ErrFailedToGetProcessInfo = errors.New("failed to get process info")
+)
+
 func handleExecEvent(event *cacheObj, nspid uint32) error {
 	var podInfo *tetragon.Pod
 
@@ -59,8 +64,13 @@ func handleExecEvent(event *cacheObj, nspid uint32) error {
 		podInfo, _ = process.GetPodInfo(containerId, filename, args, nspid)
 		if podInfo == nil {
 			errormetrics.ErrorTotalInc(errormetrics.EventCachePodInfoRetryFailed)
-			return fmt.Errorf("failed to get pod info")
+			return ErrFailedToGetPodInfo
 		}
+	}
+
+	if event.internal == nil {
+		errormetrics.ErrorTotalInc(errormetrics.EventCacheProcessInfoFailed)
+		return ErrFailedToGetProcessInfo
 	}
 
 	event.internal.AddPodInfo(podInfo)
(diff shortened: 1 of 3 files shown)


pkg/reader/notify/notify.go
```diff
@@ -11,3 +16,12 @@ type Event interface {
 	SetProcess(*tetragon.Process)
 	Encapsulate() tetragon.IsGetEventsResponse_Event
 }
+
+func EventTypeString(event Event) string {
+	// Get the concrete type of the event
+	ty := fmt.Sprintf("%T", event)
```
tpapagian, line 22: Maybe an alternative way to get the type of event is `reflect.TypeOf(event).String()`. I don't know if there are any reason to prefer one way over another but this seems to work fine.

tpapagian (approved):

pkg/reader/notify/notify.go
willfindlay: @tpapagian from my understanding they are basically equivalent. maybe one is more idiomatic than the other, who knows 😀

kkourt (approved):

pkg/reader/notify/notify.go
willfindlay: According to some Go experts I asked, the way I currently have it does indeed seem to be preferred

willfindlay:
@tpapagian just curious because I remember you mentioned you were going to look into it. Does this actually resolve the kprobes issue you were looking into are did we just get lucky with the CI?

tpapagian:
> @tpapagian just curious because I remember you mentioned you were going to look into it. Does this actually resolve the kprobes issue you were looking into or did we just get lucky with the CI?

@willfindlay Sadly, no. I rerun this PR and the same errors appeared. 

On the other hand, I think I have found the issue. I hope that https://github.com/cilium/tetragon/pull/283/commits/2557de2184fd2ced62752564f4c7fd4345058320 and https://github.com/cilium/tetragon/pull/283/commits/3ac2672915f7001c0d804926c561716858635f72 (part of https://github.com/cilium/tetragon/pull/283) should solve all kprobe flakes!

tpapagian:
I have also rebased this PR to the changes that I have done in https://github.com/cilium/tetragon/pull/283.

willfindlay:
Nice. Anyway let's get the metrics in regardless. I guess we should wait for an ack from @jrfastab but I don't think I have anything controversial here so maybe we can just merge this into your branch.

pkg/eventcache/eventcache.go
```diff
@@ -59,10 +64,15 @@ func handleExecEvent(event *cacheObj, nspid uint32) error {
 		podInfo, _ = process.GetPodInfo(containerId, filename, args, nspid)
 		if podInfo == nil {
 			errormetrics.ErrorTotalInc(errormetrics.EventCachePodInfoRetryFailed)
-			return fmt.Errorf("failed to get pod info")
+			return ErrFailedToGetPodInfo
 		}
 	}
 
+	if event.internal == nil {
```
jrfastab, line 71: This would just be Exit events correct? Execve events should have this?

willfindlay: @jrfastab This handler is only being run on exec events right now. Should it be run for exit events too?

michi-covalent (approved):
🚢

pkg/eventcache/eventcache.go
jrfastab: no its good as is.... :coffee:

willfindlay:
@jrfastab I'm gonna ship it

pkg/eventcache/eventcache.go
jrfastab: shouldn't we call process.GetParentProcessInternal(p.Pid.Value, event.timestamp) here? Where are we going to get out of this loop?

PR #306, opened 2022-08-10
devel README.md: add podman instructions

dmitris:
PR extends `docs/contributing/development/README.md` with instructions on using `podman` instead of `docker`.

### DIFF
--- a/docs/contributing/development/README.md
+++ b/docs/contributing/development/README.md
@@ -113,6 +113,13 @@ You can build Tetragon as follows:
 make
 ```
 
+If you want to use `podman` instead of `docker`, you can do the following (assuming you
+need to use `sudo` with `podman`):
+```
+CONTAINER_ENGINE='sudo podman' make
+```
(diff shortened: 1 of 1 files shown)


willfindlay (approved):
Thanks :)

docs/contributing/development/README.md
```diff
@@ -113,6 +113,13 @@ You can build Tetragon as follows:
 make
 ```
 
+If you want to use `podman` instead of `docker`, you can do the following (assuming you
+need to use `sudo` with `podman`):
+```
+CONTAINER_ENGINE='sudo podman' make
+```
+You can ignore `/bin/sh: docker: command not found` in the output.
```
jrfastab, line 121: would be nice if the docker command not found wasn't printed but I wont hold up the PR for this. Thanks.

willfindlay: I know why it's being printed. I can probably fix it up tomorrow if nobody else gets to it first.

PR #314, opened 2022-08-11
btf: improve logging around BTF files lookup

tixxdz:
Users and even developers may get confused on why Tetragon is not working or did it find the appropriate BTF files. Improve logging and reduce confusion


tixxdz:
thanks @olsajiri ! fixed tests as it seems btf file is under /boot/ and we have some env sets, all pass now.

PR #319, opened 2022-08-12
tetragon: before parsing environment vars replace "-" with "_"

tixxdz:
Before parsing environment variables we have to replace every "-" by "_"
in variable names.

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime/pprof"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -368,6 +369,8 @@ func execute() error {
 			}
 			log.WithField(keyConfigDir, configDir).Info("Loaded config from directory")
 		}
+		replacer := strings.NewReplacer("-", "_")
+		viper.SetEnvKeyReplacer(replacer)
 		viper.AutomaticEnv()
 	})
 


cmd/tetragon/main.go
```diff
@@ -368,6 +369,8 @@ func execute() error {
 			}
 			log.WithField(keyConfigDir, configDir).Info("Loaded config from directory")
 		}
+		replacer := strings.NewReplacer("-", "_")
+		viper.SetEnvKeyReplacer(replacer)
```
kkourt, line 373: I thought viper did the "-" -> "_" translation on its own, is this not the case?

willfindlay: @kkourt not by default according to their docs, but they have this in one of their examples.

willfindlay: Near the end of this section https://github.com/spf13/viper#working-with-environment-variables

tixxdz: @kkourt yep seems not the case! it doesn't handle parameters with "-"... by default but with that replacer it works, also I checked cilium and they do this manually, replace all "-" with "_" concat then env prefix then uppercase...

tixxdz: @kkourt found this: https://github.com/spf13/viper/blob/master/viper_test.go#L720

kkourt (approved):
Thanks!

(vmtests failures are due to known RCU stall issue, and this PR is unrelated anyway, merging)

PR #325, opened 2022-08-12
tetragon: handling clone events and exec events more correctly

jrfastab:
Currently, we leak clone events into the procCache. The procCache is an LRU so not free'ing events correctly means after some uptime we have a bunch of unused memory around and LRU will free things. Nothing breaks but its unnecessary memory usage.

To fix ensure clone and execve events work together correctly see patch1. This issue was introduced with the clone event relatively recently.

### DIFF
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -192,11 +192,20 @@ event_execve(struct sched_execve_args *ctx)
 			init_curr = 1;
 #endif
 		curr->key.pid = execve->pid;
-		curr->key.ktime = execve->ktime;
 		curr->nspid = execve->nspid;
 		curr->pkey = event->parent;
+		/* If this event was preceded by a clone event use the ktime
+		 * of when the clone happened. This allows us to build a single
+		 * execId for the clone+exec common path. The execId can then
+		 * be used for the procCache. If we are running exec without
+		 * a clone() then we need a new entry and user space will
+		 * learn that we smashed the current stack with exec.
+		 */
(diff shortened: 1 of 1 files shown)


kkourt:
> 2022-08-12T23:29:51.6448826Z --- FAIL: TestGrpcExecOutOfOrder (1.90s)

Failures seem legit.

jrfastab:
Dropped the 2nd patch as I ponder if ths is a good idea. The 2nd patch theory was in the case where we have a clone event followed by an exec event we currently check if the clone bit is set in the exec event and then inc the parent. If its not set we do a bit of a dance with an inc followed by a dec. But, receiving the clone event indicates we are in the clone path so we could do the inc there without any if/else logic and then exec could drop the inc/dec dance.

Although we hit the problem that the exec tests don't currently do clone events even though they set the clone bit. So we need a bit more work to make above correct from testing side. Basically we need to add clone events to the tests and then also test the other case where we get exec() calls without clones(). This happens in libs and such when they smash their own stack with a new binary. Often used by launcher codes to avoid creating more stacks.

So TLDR I think its still correct just didn't want to hold the first patch up to get all the tests in place.

jrfastab:
Further there are some interesting cases to consider from an ordering side. Some that come to mind is what happens if we get OOO exec and clone events? In this way having the bit set is nice because the exec event is mostly self contained. And what happens if the parent event is missing when we get the clone event. At the moment the clone event just folds and gives up. But current code will recover reasonably well because the exec event will go through the cache.

jrfastab:
@tpapagian some things to consider ^^^

PR #330, opened 2022-08-17
tetragon: Do not fail because of the empty map ids

olsajiri:
Add program even with no maps.


jrfastab (approved):

tixxdz:
lgtm! thank you @olsajiri

PR #332, opened 2022-08-17
Add UnloadHook to Sensors

kevsecurity:
Some sensors need to perform additional work when unloaded, in order to
leave the system in a clean state, as part of facilitating the
capability to load and unload sensors repeatedly. A clear example would
be if a sensor installed a timer (an implementation of a stopable
ticker) in an enable function.

This commit adds an UnloadHook that can optionally be set by a sensor to
point to a function to call, prior to the programs and maps being
unloaded, when the sensor is being unloaded.

### DIFF
--- a/pkg/sensors/sync.go
+++ b/pkg/sensors/sync.go
@@ -276,6 +276,12 @@ func UnloadSensor(ctx context.Context, bpfDir, mapDir string, sensor *Sensor) er
 		return fmt.Errorf("unload of sensor %s failed: sensor not loaded", sensor.Name)
 	}
 
+	if sensor.UnloadHook != nil {
+		if err := sensor.UnloadHook(); err != nil {
+			logger.GetLogger().Warnf("Sensor %s unload hook failed: %s", sensor.Name, err)
+		}
+	}
+
 	for _, p := range sensor.Progs {
 		RemoveProgram(bpfDir, p)
 	}
--- a/pkg/sensors/sensors.go
+++ b/pkg/sensors/sensors.go
@@ -43,6 +43,10 @@ type Sensor struct {
 	Loaded bool
 	// Ops contains an implementation to perform on this sensor.
 	Ops Operations
+	// UnloadHook can optionally contain a pointer to a function to be
+	// called during sensor unloading, prior to the programs and maps being
+	// unloaded.
+	UnloadHook SensorUnloadHook
 }
 
 // Operations is the interface to the underlying sensor implementations.
@@ -53,6 +57,10 @@ type Operations interface {
 	GetConfig(cfg string) (string, error)
 	SetConfig(cfg string, val string) error
 }
+
+// SensorUnloadHook is the function signature for an optional function
+// that can be called during sensor unloading.
+type SensorUnloadHook func() error
 
 func SensorCombine(name string, sensors ...*Sensor) *Sensor {
 	progs := []*program.Program{}


pkg/sensors/sync.go
```diff
@@ -276,6 +276,10 @@ func UnloadSensor(ctx context.Context, bpfDir, mapDir string, sensor *Sensor) er
 		return fmt.Errorf("unload of sensor %s failed: sensor not loaded", sensor.Name)
 	}
 
+	if sensor.UnloadHook != nil {
```
kaworu, line 279: Note that there is a potential typed `nil` issue here depending on how `sensor.UnloadHook` is setup.

kaworu (approved):
No objection, but please see my comment.

pkg/sensors/sync.go
kevsecurity: sensor.UnloadHook is defined in type Sensor struct in sensors.go, as a SensorUnloadHook interface containing the Unload() implementation. So, IIRC, whatever is assigned to sensor.UnloadHook has to include an implementation of Unload(), so if not nil then the implementation exists, otherwise is nil, and doesn't have a type. It seems to work correctly where some sensors define it and others don't. Can you describe the case where it would go wrong so I can tighten this up if necessary? Thanks!

kaworu: > […] so if not nil then the implementation exists, otherwise is nil, and doesn't have a type. […]

This is not completely accurate, an interface variable (such as `Sensor.UnloadHook`) is basically a fat pointer with both type and value. Here is a simple example showcasing the issue that can arise:

```go
package main

import "fmt"

type SensorUnloadHook interface {
	Unload() error
}

type Sensor struct {
	UnloadHook SensorUnloadHook
}

type HookImpl struct{}

// Unload implement SensorUnloadHook for HookImpl.
func (*HookImpl) Unload() error {
	fmt.Println("HookImpl.Unload")
	return nil
}

func NewSensor(onUnload SensorUnloadHook) *Sensor {
	return &Sensor{
		UnloadHook: onUnload,
	}
}

func main() {
	ok := NewSensor(nil)
	// prints true
	fmt.Println(ok.UnloadHook == nil)

	var h *HookImpl = nil
	boom := NewSensor(h)
	// prints false!
	fmt.Println(boom.UnloadHook == nil)
}

```

pkg/sensors/sync.go
```diff
@@ -276,6 +276,10 @@ func UnloadSensor(ctx context.Context, bpfDir, mapDir string, sensor *Sensor) er
...
+		sensor.UnloadHook.Unload()
```
kaworu, line 280: Here the returned error is never checked.

pkg/sensors/sync.go
kevsecurity: I see. If you explicitly assign a typed nil then that will of course fail. Is there a better way to build in an optional interface?

pkg/sensors/sync.go
kevsecurity: True. I'll add a log output.

pkg/sensors/sync.go
kaworu: AFAIK there are several potential  fix / workaround:

1. declare the unload hook as a function instead of an interface. This would sidestep the problem entirely by always using a concrete type.
2. put the burden on the interface implementations and require them to be `nil`-safe. In my example above, you can actually call `boom.UnloadHook.Unload()` because it doesn't use the `nil` receiver.
3. use `reflect` to dynamically find whether we have a typed `nil`,  e.g. `sensor.UnloadHook != nil && reflect.ValueOf(sensor.UnloadHook) != nil`

Pick you poison :sweat_smile:

kevsecurity: Thanks! Wasn't aware I could just have a function pointer and that it could be nil, so (1) above is actually the best solution to this. I've updated the commit.

PR #225, opened 2022-07-11
Tetragon: rework cgroups handling

tixxdz:
This improves Tetragon cgroups and process tracking.

Signed-off-by: Djalal Harouni [djalal@isovalent.com](mailto:djalal@isovalent.com)

PR #334, opened 2022-08-18
tetragon: Debug output fixes

olsajiri:
Adding -debug option to sensor tests and moving some of the loading messages to debug level.

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/cilium/tetragon/pkg/ratelimit"
 	"github.com/cilium/tetragon/pkg/sensors"
 	"github.com/cilium/tetragon/pkg/sensors/base"
+	"github.com/cilium/tetragon/pkg/sensors/program"
 	"github.com/cilium/tetragon/pkg/server"
 	"github.com/cilium/tetragon/pkg/version"
 	"github.com/cilium/tetragon/pkg/watcher"
@@ -113,6 +114,11 @@ func tetragonExecute() error {
 	// Logging should always be bootstrapped first. Do not add any code above this!
 	if err := logger.SetupLogging(option.Config.LogOpts, option.Config.Debug); err != nil {
 		log.Fatal(err)
+	}
+
+	// enable extra programs/maps loading debug output
+	if logger.DefaultLogger.IsLevelEnabled(logrus.DebugLevel) {
+		program.KeepCollection = true
 	}
 
 	log.WithField("version", version.Version).Info("Starting tetragon")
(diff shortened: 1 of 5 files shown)


cmd/tetragon/main.go
```diff
@@ -115,6 +116,9 @@ func tetragonExecute() error {
 		log.Fatal(err)
 	}
 
+	// enable extra programs/maps loading debug output
+	program.KeepCollection = option.Config.Debug
+
```
tixxdz, line 121: @olsajiri to get the right log level context, we need something like this:
https://github.com/cilium/tetragon/pull/225/commits/8930368b778f8287f6cc2bde9d92638ede095ce6

olsajiri: hm, but this one is just global bool to get the debug messages,
plus option.Config.Debug bool enables debug output in above statement

tixxdz: Ah yes, debug can also be set by log-level=debug where option.Config.Debug is just an alias to: log-level=debug... this was done to not break old tetragon helm charts and follow cilium IIRC

So since this touch main tetragon, please check here: https://github.com/cilium/tetragon/blob/main/cmd/tetragon/main.go#L114 we setup logging, so probably you should do something like that helper to get if loglevel is debug or above after we setup that logging... hope it makes sense.

I think we may add a comment to note that option.Config.Debug is just an alias and we should go GetLogLevel() ?

olsajiri: > Ah yes, debug can also be set by log-level=debug where option.Config.Debug is just an alias to: log-level=debug... this was done to not break old tetragon helm charts and follow cilium IIRC
> 
> So since this touch main tetragon, please check here: https://github.com/cilium/tetragon/blob/main/cmd/tetragon/main.go#L114 we setup logging, so probably you should do something like that helper to get if loglevel is debug or above after we setup that logging... hope it makes sense.

hum, not sure I follow.. the logger.SetupLogging sets debug level based
on the `option.Config.Debug` value, all this change does is to enable
additional debug output if `option.Config.Debug` value is true

maybe I can make it more explicit? with:

```
        // enable extra programs/maps loading debug output
        if option.Config.Debug {
                program.KeepCollection = true
        }
```

olsajiri: ah ok there's --log-level option, I see your point.. how about this then:

```
       // enable extra programs/maps loading debug output
       if logger.DefaultLogger.IsLevelEnabled(logrus.DebugLevel) {
               program.KeepCollection = true
       }
```

tixxdz: @olsajiri perfect ;-)  thank you!

kkourt (approved):

kkourt:
Treating @tixxdz's comment:
>  perfect ;-) thank you!

As an implicit ✅, merging.

PR #337, opened 2022-08-18
encoder: improve api to make it more generic

willfindlay:
See commits.


willfindlay:
Still messing around with this, marking as draft

michi-covalent (approved):
let's roll

Issue #33, opened 2022-05-17
Add first documentation for contribution and developement


tixxdz:
Add documentation guides for contribution and development.

This issue and the commit log that fixes it will be referenced in the doc as examples.

PR #339, opened 2022-08-18
Vagrant on MacOS fix information

dmitris:
PR adds a hint regarding vagrant/virtualbox fix - from https://stackoverflow.com/questions/18149546/macos-vagrant-up-failed-dev-vboxnetctl-no-such-file-or-directory.

See also
https://cilium.slack.com/archives/C03EV7KJPJ9/p1660806156966189?thread_ts=1660693909.096529&cid=C03EV7KJPJ9

PR is currently cumulative with (the tiny) #33, can rebase if/when #33 is merged.

/cc @kkourt

### DIFF
--- a/README.md
+++ b/README.md
@@ -622,7 +622,7 @@ we provide a standard VagrantFile with the required components enabled. Simply r
  $ vagrant ssh
  ```
 
-This should be sufficient to create a Kind cluster and run Tetragon.
+This should be sufficient to create a Kind cluster and run Tetragon. For more information on the vagrant builds, see the [Development Guide](docs/contributing/development/README.md#local-development-in-vagrant-box).
 
 # FAQ
 
--- a/docs/contributing/development/README.md
+++ b/docs/contributing/development/README.md
@@ -210,6 +210,8 @@ If you are on a Mac, use Vagrant to create a dev VM:
     vagrant ssh
     make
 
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -617,6 +617,8 @@ we provide a standard VagrantFile with the required components enabled. Simply r
 
 This should be sufficient to create a Kind cluster and run Tetragon.
 
+If you are getting an error, you can try to run `sudo launchctl load /Library/LaunchDaemons/org.virtualbox.startup.plist` (from https://stackoverflow.com/questions/18149546/macos-vagrant-up-failed-dev-vboxnetctl-no-such-file-or-directory).
+
```
kkourt, line 621: Instead of making this note twice, how about we add a link  to the vagrant section in development/README.md?

kkourt (changes requested):
Thanks!

The first patch we should definitely use. Not so sure about the second, since i) it is not related to tetragon, and ii) it is easy to find information, and iii) might become outdated. On the other hand, I don't use Vagrant so if people that do think it's useful, I'm all for keeping it.

README.md
dmitris: how about the current version? you can preview on https://github.com/dmitris/tetragon/blob/vagrant-fix/README.md#btf-requirement

kkourt: I think we could also remove the text above that, and say "To run with [vagrant](...), please see the instructions in the [Development Guilde](...). WDYT?

kkourt:
e2e test failure seems unrelated, retrying.

kkourt (approved):

PR #347, opened 2022-08-22
color: make ProcessInfo public

willfindlay:
Make ProcessInfo usable across packages. This should be useful in the future if we decide
to split color.go out of the encoder package. It also helps folks downstream who want to
write custom encoders.


michi-covalent (approved):
✅

PR #350, opened 2022-08-23
trace perf event alloc

sarahfujimori:
This patch allows us to print information about the first step of the kprobe attach process, which is when a perf event is created. We do this by collecting kprobe function name, type, config, and probe offset from the perf_event arg of the function security_perf_event_alloc.


tpapagian (approved):
Seems fine, thanks! I believe that this is a good idea to add a unit-test for that (maybe in a follow-up PR).

jrfastab (approved):

PR #361, opened 2022-08-25
tests/e2e: move cilium-e2e packages into Tetragon

willfindlay:
Upstream Cilium folks have decided to not use e2e framework for now. In order to prevent
us keeping an extra Cilium dependency around with the potential for conflicts, let's just
move all the cilium/cilium e2e functionality into Tetragon.


willfindlay:
Bad rebase, will investigate

kkourt (approved):

PR #362, opened 2022-08-25
tetragon: Allow retry logic for clone events

olsajiri:
The AddCloneEvent fails to add process the the cache if
the parent does not exist - there's no record of it. 

This can happen for example in TestFork test where fork events
can be delievered out of order and tetragon sees the child fork
event before the parent's one.

Adding retry logic for clone events, which will retry the fork
event processing and adding the hild process entry after the 
timeout. During the timeout the parent is updated and following
retry logic will add the child's entry properly.

### DIFF
--- a/pkg/api/readyapi/readyapi.go
+++ b/pkg/api/readyapi/readyapi.go
@@ -12,6 +12,10 @@ import (
 
 type MsgTetragonReady struct{}
 
+func (msg *MsgTetragonReady) Notify() bool {
+	return false
+}
+
 func (msg *MsgTetragonReady) RetryInternal(ev notify.Event, timestamp uint64) (*process.ProcessInternal, error) {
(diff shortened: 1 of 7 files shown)


tpapagian (approved):
This seems reasonable, thanks!

it would be great if we can test that this PR solves all `TestFork` flakes. @kkourt had a way to repro the flake in `TestFork` (I tried to reproduce that locally but didn't manage to achieve that).

olsajiri:
> This seems reasonable, thanks!
> 
> it would be great if we can test that this PR solves all `TestFork` flakes. @kkourt had a way to repro the flake in `TestFork` (I tried to reproduce that locally but didn't manage to achieve that).

TestFork was failing on 4.19 for me, with this I can no longer reproduce the fail, so I'm hopeful ;-)

jrfastab (approved):

PR #367, opened 2022-08-29
tetragon: Switch to BTF defined maps

olsajiri:
Switching to BTF defined maps. This allows to dump map values nicely, like:

```
# bpftool map dump pinned execve_map | head -10
[{
        "key": 431671,
        "value": {
            "key": {
                "pid": 431671,
                "pad": [0,0,0,0
                ],
                "ktime": 171231811545404
            },
            "pkey": {
...
```


tpapagian (approved):

kkourt (approved):
Checkpatch failures seem to be in bpf_helpers.h, so not much we can do about it.

PR #368, opened 2022-08-29
Delete entries from execve_map when pid != tgid

tpapagian:

### DIFF
--- a/bpf/process/bpf_exit.h
+++ b/bpf/process/bpf_exit.h
@@ -28,8 +28,10 @@ event_exit_send(struct sched_execve_args *ctx, __u64 current)
 	 * and we can skip the lookup/insert/delete cycle that
 	 * would otherwise occur.
 	 */
-	if (pid != tgid)
+	if (pid != tgid) {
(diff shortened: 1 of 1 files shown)


olsajiri (approved):

jrfastab:
hmm so how did we get an entry in the map though with a non thread group ID?

olsajiri:
> 

there are places like 
https://github.com/cilium/tetragon/blob/a34d1ffe6650510ed6f060c8b283a10b2bdd690a/bpf/lib/bpf_events.h#L198

and 
https://github.com/cilium/tetragon/blob/a34d1ffe6650510ed6f060c8b283a10b2bdd690a/bpf/lib/bpf_events.h#L152

where we go down through parents.. which I'm actually not sure they need to be group leaders, but we don't check.. also execve tracepoint does not check if the pid is group leader

jrfastab:
I think those should be execve_map_get_noinit()?

tpapagian:
Closing this as we are going to take the fix in https://github.com/cilium/tetragon/pull/370

PR #369, opened 2022-08-29
cgroups: select which cgroup hierarchy and subsystem state to use

tixxdz:
This was part of https://github.com/cilium/tetragon/pull/225 , it was cleaned in order to improve our logic how we operate on cgroup hierarchies.

bpf:cgroups: pass subsys index when operating on cgroup_subsys_state set
    
Select which cgroup controllers to use at runtime by analyzing current machine
cgroup configuration and adapt the bpf helpers to use the best option.
    
We have experienced events that did not have the proper container ID 'docker' nor the pod fields set. The reason is due to:
    
In Cgroupv1 mode usually systemd only sets up by default the 'cpu, cpuacct, memory, devices and pids' controllers, cpuset which in normal cases indexed at 0 is not installed. Since some container runtimes and different environments may use systemd as a cgroup driver this can cause problems where we won't operate on the right hierarchy. We should also note that these controllers are kernel compile CONFIG_* options, and Tetragon should only work on machines that have the CONFIG_CGROUP_PIDS or CONFIG_CGROUP_MEMORY controllers compiled. Most production machines these days have or must have these compiled in to properly work.
    
Let's be consistent with systemd, Kubernetes and Container runtimes and select by default either the 'memory' or 'pids' to be used as the tracking Cgroup hierarchy for all processes. Usually these two controllers are always present and set. We do this by selecting the right hierarchy ID and the cgroup subsystem state index.
    
In Cgroupv2 mode, systemd successfully sets the related controllers that are safe to be used by default. However we have experienced machines that did not have the cpuset controller. In order to avoid such errors we do same operation for Cgroupv1, we gather the Cgroup subsystem state index and pass it into tetragon_conf struct at startup. To get the Cgroup ID we first use the default Cgroupv2 BPF helpers, if they fail we fallback to the per subsystem index. Last, to get the Cgroup name we always query the subsystem index and read the kernfs node name.
    
This allows Tetragon to work on different environments, regardless of the Cgroup configuration and driver being used.
    
Further reference: https://github.com/systemd/systemd/blob/main/src/basic/cgroup-util.h#L20

### DIFF
--- a/bpf/include/vmlinux.h
+++ b/bpf/include/vmlinux.h
@@ -6434,6 +6434,21 @@ struct kernfs_node {
 	struct kernfs_iattrs *iattr;
 };
 
+/* Represent old kernfs node present in 5.4 kernels and older */
+union kernfs_node_id {
+	struct {
+		/*
+		 * blktrace will export this struct as a simplified 'struct
+		 * fid' (which is a big data struction), so userspace can use
+		 * it to find kernfs node. The layout must match the first two
+		 * fields of 'struct fid' exactly.
+		 */
+		u32 ino;
+		u32 generation;
+	};
+	u64 id;
+};
+
 struct kernfs_open_file;
 
 struct kernfs_ops {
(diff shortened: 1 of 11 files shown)


willfindlay (approved):
This fixes a nasty cgroupsv2 bug we've been tracking on 4.19 :tada:

kevsecurity (approved):

PR #371, opened 2022-08-29
tetragon: make gops address configurable

willfindlay:
It would be nice to make the gops address configurable so we can set it to a predictable
value in helm and expose it via the helm chart. That way we can easily get gops data
via a port forward.


michi-covalent (approved):
looks innocent enough

willfindlay:
😇

PR #381, opened 2022-09-02
sensor cleanups

kkourt:
Please see commits


cmd/tetragon/main.go
```diff
@@ -246,24 +239,32 @@ func tetragonExecute() error {
 		go crd.WatchTracePolicy(ctx, observer.SensorManager)
 	}
 
-	var startSensors []*sensors.Sensor
+	// load base sensor
+	if err := base.GetInitialSensor().Load(ctx, observerDir, observerDir, option.Config.CiliumDir); err != nil {
```
olsajiri, line 243: I'm scratching my head how this get compiled.. base.GetInitialSensor returns *sensors.Sensor and we return it as 'err' as error return value, what do I miss?

kkourt: `GetInitialSensor()` returns a `*sensors.Sensor`, which defines a `Load(...)` method that returns an error. So doing `GetInitialSensor().Load()` returns an error.

olsajiri: aaargh there's .Load .. sorry for noise ;-)

pkg/observer/observer.go
```diff
@@ -107,84 +89,7 @@ func (k *Observer) receiveEvent(data []byte, cpu int) {
...
-	}
-	defer e.CloseAll()
-
-	k.__loopEvents(stopCtx, e)
-	return nil
-}
-
-func (k *Observer) runEventsNew(stopCtx context.Context, ready func()) error {
-	/* Probe runtime configuration and do not fail on errors */
-	k.UpdateRuntimeConf()
```
kevsecurity, line 187: It looks like this line (added to support Cgroups on older kernels AFAIK) has been lost in the process of swapping over.

willfindlay: Yeah we will need to keep this in, IIRC this is part of what resolved that nasty bug on 4.19 with cgroupsv2.

willfindlay (approved):
LGTM once @kevsecurity's comment is addressed.

pkg/observer/observer.go
kkourt: ah, nice catch. I messed up the rebase. Let me fix it.

kkourt: should be OK now, PTAL.

tixxdz (approved):

kevsecurity (approved):

PR #382, opened 2022-09-04
tetragon: Use lru.Cache for data events storage

olsajiri:
There's a chance that by losing events (in perf ring buffer)
we would never use data stored by data events. Adding the data
into lru cache so we eventually remove unused entries to make
space for new ones and don't grow endlessly.


kkourt (approved):
Cool! Thanks!

jrfastab (approved):

kkourt:
@olsajiri maybe it's worth rebasing this one?

PR #384, opened 2022-09-06
docs: add pure container and docker deployment steps

tixxdz:
Users requested this.


tixxdz:
Before merging we need the latest tag on quay.io/cilium/tetragon

docs/contributing/development/README.md
```diff
@@ -154,6 +154,32 @@ Finally, should you wish to modify any of the resulting codegen files (ending in
 do not modify them directly. Instead, you can edit the files in
 `cmd/protoc-gen-go-tetragon` and then re-run `make codegen`.
 
+### Docker Container
```
kkourt, line 157: How a bout "Building and running a docker image"?

docs/deployment/docker/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Deployment
+
+To run Tetragon image:
+
+```
+docker run --name tetragon \
+   --rm -it -d --pid=host \
+   --cgroupns=host --privileged \
+   -v /sys/kernel/btf/vmlinux:/var/lib/tetragon/btf \
+   quay.io/cilium/tetragon:latest \
```
kkourt, line 27: I don't think we tag latest because it is used by default and we want to avoid that. @michi-covalent should know more.

What I would suggest, instead, is to direct users to the logs of the images we build (both for main builds but also for PRs) and ask them to add this tag.

docs/contributing/development/README.md
tixxdz: Done

docs/deployment/docker/README.md
tixxdz: Ok done, both to main and tetragon-ci that @michi-covalent did setup! thank you

docs/contributing/development/README.md
```diff
@@ -154,6 +154,32 @@ Finally, should you wish to modify any of the resulting codegen files (ending in
...
+make image
+```
+
+To run the image:
+```
+docker run --name tetragon \
+   --rm -it -d --pid=host \
+   --cgroupns=host --privileged \
+   -v /sys/kernel/btf/vmlinux:/var/lib/tetragon/btf \
+   cilium/tetragon:latest \
```
kkourt, line 173: Doesn't cilium/tetragon:latest have the same issue as the one we discussed before? It might make sense to use a single paragraph for both

kkourt (approved):

PR #387, opened 2022-09-06
Add Exec Test For Docker/Container ID

kevsecurity:
Add a test that checks that exec events contain the correct
Docker/container ID.

### DIFF
--- a/pkg/sensors/exec/exec_test.go
+++ b/pkg/sensors/exec/exec_test.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/cilium/ebpf"
 	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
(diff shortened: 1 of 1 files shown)


pkg/sensors/exec/exec_test.go
```diff
@@ -449,3 +450,53 @@ func TestLoadInitialSensor(t *testing.T) {
...
+
+	checker := ec.NewUnorderedEventChecker(
+		ec.NewProcessExecChecker().
+			WithProcess(selfChecker).
+			WithParent(ec.NewProcessChecker()),
+		ec.NewProcessExecChecker().
+			WithProcess(ncSrvChecker),
+	)
+
+	time.Sleep(1 * time.Second)
```
tpapagian, line 498: Why do we need this `Sleep()`? (maybe the same question also for the `Sleep()` in line 474)

kevsecurity: This one isn't needed, I'll remove it. The one on 474 is just to give the process chance to start up and be logged before we check the logs.

tpapagian (approved):

PR #389, opened 2022-09-06
build-images-ci: Remove comments from tags section

michi-covalent:
They don't handle comments.

https://github.com/cilium/tetragon/runs/8213728526?check_suite_focus=true


willfindlay (approved):

willfindlay:
the comment would work if it were dedented but that's ugly so yeah let's drop it

tixxdz (approved):

PR #393, opened 2022-09-07
tests/e2e: summary dump improvements

willfindlay:
See commits.


willfindlay:
marking dontmerge while I mess around with this

PR #395, opened 2022-09-09
tetragon: Add clang 14 

olsajiri:
use clang 14


olsajiri:
we can't use the workaround for clang 14 introduced in https://github.com/cilium/tetragon/pull/395/commits/754cdbed97d5ccdee7f620893d4ba8dd407e7685 ,

because it unrolls into ~16k instructions and it breaks validation for us later in other code that uses this function

olsajiri:
@willfindlay I guess you could unblock the CI? thnx

olsajiri:
using https://github.com/cilium/tetragon/pull/397 for this

PR #396, opened 2022-09-09
api: implement field filters for GetEventsRequest

willfindlay:
This commit introduces two new field filters for GetEventsRequest. These allow the user to optionally specify a set of paths that should be included/excluded from the GetEventsResponse where exclusion takes precedence over inclusion. Leaving the filters empty will keep the default behaviour of including all fields in the GetEventsResponse. Filters are specified using period-separated names as they appear in protobuf JSON (i.e. lower snake case).

This new filtering mechanism has several use cases, for example filtering out unwanted noisy fields from events in a customer environment or implementing functionality that transforms events for further processing (for example an event recorder that filters out noise).


willfindlay:
Need to reconsider the best way to handle oneofs here, marking as draft for now.

PR #401, opened 2022-09-13
tests:cgroups: tests for cgroups package

tixxdz:


pkg/cgroups/cgroups_test.go
```diff
@@ -4,26 +4,57 @@
...
+}
+
+func TestDetectCgroupModeInvalid(t *testing.T) {
 	mode, err := detectCgroupMode("invalid-cgroupfs-path")
 	assert.Error(t, err)
 	assert.Equal(t, CGROUP_UNDEF, mode)
 }
 
-func TestDetectCgoupModeSimple(t *testing.T) {
+func TestDetectCgroupModeDefault(t *testing.T) {
```
kevsecurity, line 52: Wasn't this added in the previous commit? If so, please edit the previous commit rather than adding the change in this commit; unless there's a good reason of course!

tixxdz: Indeed, I fixed this and added more comments to explain what the tests do including commit logs, much appreciated @kevsecurity

kevsecurity (approved):

kkourt (approved):

PR #402, opened 2022-09-13
refactor: move from io/ioutil to io and os packages

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://pkg.go.dev/io/ioutil). This PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

- `ioutil.Discard` => `io.Discard`
- `ioutil.NopCloser` => `io.NopCloser`
- `ioutil.ReadAll` => `io.ReadAll`
- `ioutil.ReadDir` => `os.ReadDir` (returns a slice of `os.DirEntry` rather than a slice of `fs.FileInfo`)
- `ioutil.ReadFile` => `os.ReadFile`
- `ioutil.TempDir` => `os.MkdirTemp`
- `ioutil.TempFile` => `os.CreateTemp`
- `ioutil.WriteFile` => `os.WriteFile`

### DIFF
--- a/tests/e2e/helpers/dumpinfo.go
+++ b/tests/e2e/helpers/dumpinfo.go
@@ -7,7 +7,6 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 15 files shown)


tests/e2e/helpers/dumpinfo.go
```diff
@@ -95,10 +94,7 @@ func CreateExportDir(ctx context.Context, t *testing.T) (context.Context, error)
 		return ctx, nil
 	}
 
-	dir, err = ioutil.TempDir("", fmt.Sprintf("tetragon.e2e.%s.*", t.Name()))
-	if err != nil {
-		return ctx, err
-	}
+	dir = t.TempDir()
```
willfindlay, line 97: Our CI jobs rely on this path being predictable to upload dumps when the job fails. So we will need a way to do the same here. And also a way to ensure the file is kept around when the test fails.

willfindlay (changes requested):

willfindlay (approved):
LGTM, thanks!

PR #405, opened 2022-09-14
Add a function to load (but not attach) an eBPF program

tpapagian:
This can be used to load tail call programs.

### DIFF
--- a/pkg/sensors/program/loader.go
+++ b/pkg/sensors/program/loader.go
@@ -125,6 +125,16 @@ func KprobeAttach(load *Program) AttachFunc {
 	}
 }
 
+func NoAttach(load *Program) AttachFunc {
+	return func(prog *ebpf.Program, spec *ebpf.ProgramSpec) (unloader.Unloader, error) {
+		return unloader.ChainUnloader{
+			unloader.PinUnloader{
+				Prog: prog,
+			},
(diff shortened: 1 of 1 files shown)


pkg/sensors/program/loader.go
```diff
@@ -139,6 +149,11 @@ func LoadKprobeProgram(bpfDir, mapDir string, load *Program, verbose int) error
 	return loadProgram(bpfDir, []string{mapDir}, load, KprobeAttach(load), ci, verbose)
 }
 
+func LoadTailCallProgram(bpfDir, mapDir string, load *Program, verbose int) error {
+	ci := &customInstall{fmt.Sprintf("%s-tail-call", load.PinPath), "kprobe"}
```
tixxdz, line 153: maybe %s-tail-calls with s?

tpapagian: sure, will update that now

tpapagian: maybe I can remove this line entirely and use nil in a similar way to `LoadRawTracepointProgram`. Checking that now.

tpapagian: removed that entirely

kkourt (approved):

Issue #408, opened 2022-09-15
/sys/fs/bpf hierarchy reflecting sensors/programs/etc.


kkourt:
![tetragon bpf fs hierarchy](https://user-images.githubusercontent.com/221481/190360018-f4b49969-5393-4f43-a05c-c52fc1fda2f5.png)

kkourt:
Done by https://github.com/cilium/tetragon/pull/2128 :tad

PR #409, opened 2022-09-15
sensor pinning fixes (and misc updates)

kkourt:
This PR aims to sanitize pinning and meant as an intermediate step towards https://github.com/cilium/tetragon/issues/408. It also includes some simple (first 3) patches.

One of the immediate effects of this PR is that generic tracepoints have unique map names and collisions that may  cause problems are avoided.

See patches for more information.

### DIFF
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -601,8 +601,17 @@ func parseSelector(
 	return nil
 }
 
-// array := [number][filter1][filter2][...][filtern]
-// filter := [length][matchPIDs][matchBinaries][matchArgs][matchNamespaces][matchCapabilities][matchNamespaceChanges][matchCapabilityChanges]
+// array :=
+// [number][filter1_offset][filter2_offset][...][filtern_offset][filter1][filter2][...][filtern]
+// filter := [length]
+//           [matchPIDs]
+//           [matchNamespaces]
+//           [matchCapabilities]
+//           [matchNamespaceChanges]
+//           [matchCapabilityChanges]
+//           [matchBinaries]
+//           [matchArgs]
+//           [matchActions]
 // matchPIDs := [num][PID1][PID2]...[PIDn]
 // matchBinaries := [num][op][Index]...[Index]
 // matchArgs := [num][ARGx][ARGy]...[ARGn]
--- a/pkg/sensors/tracing/generictracepoint.go
+++ b/pkg/sensors/tracing/generictracepoint.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"path"
 	"reflect"
+	"sync/atomic"
 
 	"github.com/cilium/tetragon/pkg/api/ops"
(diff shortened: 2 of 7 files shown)


pkg/selectors/kernel.go
```diff
@@ -601,8 +601,16 @@ func parseSelector(
 	return nil
 }
 
-// array := [number][filter1][filter2][...][filtern]
-// filter := [length][matchPIDs][matchBinaries][matchArgs][matchNamespaces][matchCapabilities][matchNamespaceChanges][matchCapabilityChanges]
+// array := [number][filter1_offset][filter2_offset][...][filtern_offset]
```
tpapagian, line 604: Assuming array is the return value of `InitKernelSelectors()` shouldn't we have:
`array := [number][filter1_offset][filter2_offset][...][filtern_offset][filter1][filter2][...][filtern]`
after the offsets don't we have the actual filters?

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -547,11 +552,18 @@ func (t *observerTracepointSensor) SpecHandler(raw interface{}) (*sensors.Sensor
 		spec = &s
 	}
 
+	var name string
+	if named, ok := raw.(interface{ Name() string }); ok {
+		name = named.Name()
```
tpapagian, line 557: Can this happen? Shouldn't we have `Name()` in https://github.com/cilium/tetragon/blob/c9a0eaeefb40beef5d78f6872404c6e7309e756e/pkg/sensors/sensors.go#L140-L143?

tpapagian (approved):
LGTM!

nit: typo "hiearachy" -> "hierarchy" in the description of https://github.com/cilium/tetragon/pull/409/commits/68fae5bf9d20a7c58979c6817f96202d164a3fa5

pkg/sensors/tracing/generictracepoint.go
kkourt: Hmm... I don't think so. Seems like a leftover. I'll remove it.

pkg/selectors/kernel.go
kkourt: Yap, good point.

kkourt: Changed it to:
```
// [number][filter1_offset][filter2_offset][...][filtern_offset][filter1][filter2][...][filtern]
```

pkg/sensors/tracing/generictracepoint.go
kkourt: removed

kkourt:
> LGTM!
> 
> nit: typo "hiearachy" -> "hierarchy" in the description of [68fae5b](https://github.com/cilium/tetragon/commit/68fae5bf9d20a7c58979c6817f96202d164a3fa5)

Thanks! Pushed a new version with above fixed, as well as the rest of the comments.

PR #411, opened 2022-09-15
Fix parent reference counting in RetryInternal

tpapagian:
This PR does some fixes in the eventcache logic. Please check individual commits.

### DIFF
--- a/contrib/tester-progs/.gitignore
+++ b/contrib/tester-progs/.gitignore
@@ -5,3 +5,4 @@ sigkill-tester
 dup-tester
 trigger-test-events
 nop
+sigkill-unprivileged-user-ns-tester
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -58,12 +58,24 @@ var (
 // the process info yet.
 func HandleGenericInternal(ev notify.Event, timestamp uint64) (*process.ProcessInternal, error) {
 	p := ev.GetProcess()
-	internal, _ := process.GetParentProcessInternal(p.Pid.Value, timestamp)
+	internal, parent := process.GetParentProcessInternal(p.Pid.Value, timestamp)
+	var err error
+
+	if parent != nil {
+		if ev.GetParent() == nil {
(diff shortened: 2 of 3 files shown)


pkg/reader/notify/notify.go
```diff
@@ -12,6 +12,7 @@ type Message interface {
 	HandleMessage() *tetragon.GetEventsResponse
 	RetryInternal(Event, uint64) (*process.ProcessInternal, error)
 	Retry(*process.ProcessInternal, Event) error
+	DoRefCnt(Event, *process.ProcessInternal, *process.ProcessInternal)
```
kkourt, line 15: I think it's hard to understand this code, and adding one more method to the interface does not help in that regard. Especially since `DoRefCnt` seems to be doing more than ref counting (it also seems to be setting a parent) I wonder if there is a cleaner way to implement the same functionality.

https://github.com/cilium/tetragon/blob/c2a3d198c83beeb97b0d73b23f2b8f5b3a4f45c4/pkg/grpc/exec/exec.go#L227-L248


> In the case where we call this multiple times, and parent is
> not nil but process is nil, we will call RefDec() more than
> once in the parent. This will lead to negative refcnt and
> it will never removed from the pidMap.

Can we just check whether the parent has been set already?

This way we don't have to execute:
>	ev.SetParent(parent.GetProcessCopy()) 
>      parent.RefDec() 

twice.

tpapagian: yes, I agree with that DoRefCnt is actually doing more that what it's name says. 

I think the most common case will be to set the parent and wait for the process. We set the process a bit later (after waiting for the pod info if this is missing) and that complicates that a bit.

For sure, I will remove parent handling from DoRefCnt. I will also check if there is a more "clean" way to do that.

tpapagian: @kkourt I moved to a more simple solution as the one you proposed. Actually only the RefDec for the parent can be called more than once. The current implementation handles this case for the process and ensures that RefDec will not be called more than once.

kkourt (approved):
Thanks!

PR #412, opened 2022-09-15
testutils: fix export file when test name contains path separator

willfindlay:
Subtests in Go have a path separator ("/") in their name. This breaks our export file creation since it is considered an error by os.CreateTemp(). Replace the path separator with a "-" here to correct this.

### DIFF
--- a/pkg/testutils/filenames.go
+++ b/pkg/testutils/filenames.go
@@ -6,6 +6,7 @@ package testutils
 import (
 	"fmt"
 	"os"
+	"strings"
 	"sync"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


kkourt (approved):

willfindlay:
@kkourt I just realized I have to fix up some of the other funcs that expect a test name as well. Marking as draft till then.

willfindlay:
There we go, should be fixed.

kkourt:
> @kkourt I just realized I have to fix up some of the other funcs that expect a test name as well. Marking as draft till then.

Why was 
image

needed? `tName` is a just a key in a map 🤔

PR #413, opened 2022-09-15
Pr/sarahfujimori/trace map create

sarahfujimori:


api/v1/tetragon/tetragon.proto
```diff
@@ -188,6 +196,7 @@ message KprobeArgument {
 	int64 long_arg = 11;
 	KprobeBpfAttr bpf_attr_arg = 12;
 	KprobePerfEvent perf_event_arg = 13;
+    KprobeBpfMap bpf_map_arg = 14;
```
kkourt, line 199: nit: indentation here seems a bit off?

kkourt (approved):
LGTM, thanks!

api/v1/tetragon/tetragon.proto
sarahfujimori: Thanks! will fix the indentation

bpf/process/types/bpfmap.h
```diff
@@ -0,0 +1,15 @@
...
+
+#ifndef __BPFMAP_H__
+#define __BPFMAP_H__
+
+struct bpf_map_info_type {
+	__u32 map_type;
+	__u32 key_size;
+	__u32 value_size;
+	__u32 max_entries;
+	char map_name[16U];
```
tixxdz, line 12: Could you please #define the 16U to something like BPF_MAP_NAME_LEN? and use it everywhere, same goes for go user space part?

tixxdz (approved):
Minor thing about using cont names instead of values, lgtm thank you!

PR #415, opened 2022-09-16
Improved bpf unit testing

kkourt:
This PR introduces two new features that aim to improve BPF unit testing.

First, it adds utilities for testing using the ringbuffer directly. This allows to directly check events generated by bpf  programs, without having to go via the observer. An example test is added for tracepoints, but the same approach could be followed for kprobes as well. One of the benefits of above utilities is now we now exactly when to terminate tests, and there is no need to retry when checking that all events are there.

Second, it adds functionality to reload a program:  unlink (detach) it without unloading it, and then reloading it. This allows to do tests with different configurations, without having to reload (and thus re-verify) the program. That is, we can unload the program, change the filter map, reload it, and run tests without having to verify the program again. Reload works only for generic tracepoints for now, but doing the same for generic kprobes should be relatively easy. The reload ability could be improved and used in use-cases other than testing but this is future work.


pkg/sensors/tracing/generictracepoint.go
```diff
@@ -430,6 +436,66 @@ func (tp *genericTracepoint) EventConfig() (api.EventConfig, error) {
 	return config, nil
 }
 
+// ReloadGenericTracepointSelectors will reload a tracepoint by unlinking it, generating new
+// selector data and updating filter_map, and then relinking the tracepoint.
+//
+// This is intentended for speeding up testing, so DO NOT USE elswhere without checking its
```
tpapagian, line 442: By "do not use elsewhere" you mean outside of tracepoints or outside of `TestTracepointSelectors` (that seems to be the only use case for the next commit). 

Maybe it would be good to elaborate more on the limitations (as you said the config map is not updated, but there are other known limitation?). From my understanding this will be the skeleton to use this functionality for other than the tracepoint events.

tpapagian (approved):
Thanks, this adds a new functionality (i.e. to check directly on the ringbuffer) that can be useful to fix more bugs and make more targeted testing. 

I have done a complete pass and seems fine from my side.

pkg/sensors/tracing/generictracepoint.go
kkourt: > By "do not use elsewhere" you mean outside of tracepoints or outside of TestTracepointSelectors (that seems to be the only use case for the next commit).

I meant outside of testing :)

> Maybe it would be good to elaborate more on the limitations (as you said the config map is not updated, but there are other known limitation?). From my understanding this will be the skeleton to use this functionality for other than the tracepoint events.

I'm not aware of other limitations, but I haven't thought it through for use-cases other than testing (e.g., reloading the sensors if the CR changes). And yes, the same approach could be use for kprobes. (Hoping to do that in a next PR.)

pkg/observer/observer.go
```diff
@@ -72,23 +72,62 @@ func (k *Observer) RemoveListener(listener Listener) {
...
+		err = &handlePerfHandlerErr{op: op, err: err}
+	}
+	return events, err
+}
+
+func (k *Observer) receiveEvent(data []byte, cpu int) {
+
+	k.recvCntr++
+	events, err := HandlePerfData(data)
+	switch e := err.(type) {
```
kkourt, line 118: There is an issue here: this `switch` has to be under the `if err != nil` below. Pushing a fix soon.

PR #417, opened 2022-09-16
sandbox: crds to organize and kill syscalls per groups

tixxdz:
This is intended to demo and lay the ground on how Tetragon could be used to construct a syscall based sandbox. 

For now we just put the raw syscalls directly, in future we can have declarative groups to make it easy for users to perform kill or trace without writing a long spec.

### DIFF
--- a/contrib/tester-progs/sigkill-module-syscalls-group-tester.c
+++ b/contrib/tester-progs/sigkill-module-syscalls-group-tester.c
@@ -1,0 +1,316 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+// Copyright Authors of Tetragon
+
+#define _GNU_SOURCE
+#include <errno.h>
+#include <fcntl.h>
+#include <linux/module.h>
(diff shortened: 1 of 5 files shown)


contrib/tester-progs/sigkill-module-syscalls-group-tester.c
```diff
@@ -0,0 +1,316 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+// Copyright Authors of Tetragon
+
```
kkourt, line 3: Isn't it possible to write this in go? Why do we need the C code?

kkourt:
@tixxdz Are you actively working on this? If not, let's move it to draft mode.

kkourt:
> @tixxdz Are you actively working on this? If not, let's move it to draft mode.

Moving it to draft mode. Feel free to mark it as ready when it is ready for reviews.

tixxdz:
This feature was merged in another PR, closing!

PR #421, opened 2022-09-20
add InMap operator for filtering arguments, and improve eBPF unit testing for kprobes

kkourt:
Mainly, this PR adds an InMap (and NotInMap) argument operator for generic tracepoints and kprobes. InMap acts as a filter based on whether the value of an argument exists on a map or not. It works similar to the Equal operator, but it does not have the constrains on the number of values of the latter.

Secondly, this PR improves upon #415, by adding support for testing different selectors without reloading (and re-verifying) the program to generic kprobes.

Thirdly, it improves the implementation for both generic kprobes and tracepoints. Please review commit-by-commit.

### DIFF
--- a/bpf/process/argfilter_maps.h
+++ b/bpf/process/argfilter_maps.h
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright Authors of Cilium */
+
+#ifndef ARGFILTER_MAPS_H__
+#define ARGFILTER_MAPS_H__
+
+#define ARGFILTER_MAPS_OUTER_MAX_ENTRIES 8
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, ARGFILTER_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, 1);
+			__type(key, __u64);
+			__type(value, __u8);
+		});
+} argfilter_maps SEC(".maps");
+
+#endif // ARGFILTER_MAPS_H__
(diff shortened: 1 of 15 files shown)


bpf/process/argfilter_maps.h
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright Authors of Cilium */
+
+#ifndef ARGFILTER_MAPS_H__
+#define ARGFILTER_MAPS_H__
+
+#define ARGFILTER_MAPS_OUTER_MAX_ENTRIES 8
```
tpapagian, line 7: Question: What means to have 8 entries in the outer map? I saw another comment that means 64 bits but not sure I follow that. Why we need 1 outer map with 8 entries and each inner map has 1 element? Why not use one BPF_MAP_TYPE_HASH with 8 entries?

Based on these, I believe that I didn't get the details on how `op_filter_inmap` works. I understand that it will remove the limit of 4(?) values that we have but not sure how.

I propose to just add some example yaml files in `crds/examples/` that use `op_filter_inmap` and with some comments there. This will make much easier to understand on the high level idea of this mechanism.

kkourt: > What means to have 8 entries in the outer map? 

 That we can have 8 tables, that is 8 different set of values to check arguments against.
 
 > Why we need 1 outer map with 8 entries and each inner map has 1 element?
 
 The 1 element max_entry will be overridden by the loader. The inner map will have a maximum size that can fit all the entries we need.
 
 See the `populateArgFilterMap` function:
 
 ```
 innerSpec := &ebpf.MapSpec{
		Name:       innerName,
		Type:       ebpf.Hash,
		KeySize:    8, // NB: hardcoded to 64 bits for now
		ValueSize:  uint32(1),
		MaxEntries: uint32(len(innerData)),
}
```

tpapagian (approved):
Thanks, LGTM!

PR #422, opened 2022-09-20
tetragon: Fix merged kprobe events

olsajiri:
Fixing merging of kprobe events (returnCopy case) by adding 'ktime' to the merged key.

At the moment we merge kprobe events by matching the 'thread_id',
which can lead to wrong events in case we have LOST events like in
following sequence of received events:

```
 -  read/kprobe
 -  read/kreptprobe (LOST)
 -  read/kprobe (LOST)
 -  read/kretprobe
```

having 'ktime' part of the key eliminates such cases

This patchset:
  - adds metrics for merged events
  - adds support to 'rw' benchmark to generate case above
  -  store events in tetragon-bench for post processing
  - adds ktime to the key in kprobe merge map

With the fix and some post processing of the benchmark output data,
we can see more better numbers for correctly merged events:

We generated 'random sys_read' events and verified in the generated
output.json that the sys_read size argument matches with the length
of the retrieved data (it's the 'changed false' case in the output below).

with the fix:

```
        Ring buffer:        received=1829323, lost=58838568, errors=0
        Merged events:      pushed=1252435, ok=230527, errors=1021908

        [root@krava tetragon]# cat out | grep  'changed false' | wc -l
        229583
        [root@krava tetragon]# cat out | grep  'changed true' | wc -l
        912
```

base:

```
        Ring buffer:        received=1876130, lost=58814126, errors=0
        Merged events:      pushed=763227, ok=247434, errors=515793

        [root@krava tetragon]# cat out | grep  'changed true' | wc -l
        125061
        [root@krava tetragon]# cat out | grep  'changed false' | wc -l
        122337
```

We can see much better stats for the fix. The extra '912' bad events
seems is the other 'sys_read' traffic, where sys_read syscall did not
return full amount of bytes, because the 'rw' benchmark is system wide.  

### DIFF
--- a/bpf/process/bpf_generic_retkprobe.c
+++ b/bpf/process/bpf_generic_retkprobe.c
@@ -55,10 +55,13 @@ BPF_KRETPROBE(generic_retkprobe_event, unsigned long ret)
 	if (!retprobe_map_get(e->thread_id, &info))
 		return 0;
 
+	*(unsigned long *)e->args = info.ktime_enter;
+	size += sizeof(info.ktime_enter);
+
 	ty_arg = config->argreturn;
 	do_copy = config->argreturncopy;
 	if (ty_arg)
-		size += read_call_arg(ctx, e, 0, ty_arg, 0, ret, 0, 0);
+		size += read_call_arg(ctx, e, 0, ty_arg, size, ret, 0, 0);
 
 	/*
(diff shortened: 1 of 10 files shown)


jrfastab:
Maybe we could garbage collect events on exec close to try and avoid the LRU map growing too large? Either way series looks good.

PR #425, opened 2022-09-20
Add an issue template for the release process

michi-covalent:

### DIFF
--- a/contrib/update-helm-chart.sh
+++ b/contrib/update-helm-chart.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -ex
+
+shopt -s expand_aliases
+alias yq='docker run --rm -v "${PWD}":/workdir --user "$(id -u):$(id -g)" mikefarah/yq:4.27.3'
+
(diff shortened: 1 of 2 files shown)


contrib/update-helm-chart.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -ex
+
+shopt -s expand_aliases
+alias yq='docker run --rm -v "${PWD}":/workdir --user "$(id -u):$(id -g)" mikefarah/yq:4.27.3'
+
```
willfindlay, line 7: How about a grep here to check for malformed or unset version number?

michi-covalent: sure i can add some basic check

michi-covalent: ok added this check
```
if [ -z "$1" ] || [[ ! $1 =~ ^v[0-9]+\.[0-9]+\.[0-9]+.*$ ]]; then
  echo "USAGE: ./contrib/update-helm-chart.sh vX.Y.Z"
  exit 1
fi
```
should be good enough although it's not a strict semver check.

willfindlay (approved):

PR #429, opened 2022-09-21
tetragon: Update execve_map_stat value from procs reader

olsajiri:
When tetragon starts it searches for running tasks and updates execve_map, but does not update stats table, which can later lead to negative numbers in stats when the scanned processes die.

Updating stats with found procs tasks count.


tixxdz (approved):
Thank you @olsajiri

PR #430, opened 2022-09-21
License: remove Isovalent add Apache

xmulligan:


kkourt:
Thanks @xmulligan! No need to run tests, since the changes are only in comments. Merging.

PR #433, opened 2022-09-21
Add GetUrl and DnsRequest actions to probes

kevsecurity:
These two commits add GetUrl and DnsRequest actions to probes. These can be used to trigger Thinkst canaries or interact with any number of other systems that can be triggered by either getting a URL or requesting the DNS entry for a specific FQDN.


pkg/sensors/tracing/generickprobe.go
```diff
@@ -512,6 +520,11 @@ func ReadArgBytes(r *bytes.Reader, index int) (*api.MsgGenericKprobeArgBytes, er
 
 }
 
+func getUrl(url string) {
+	// We fire and forget URLs, and we don't care if they hit or not.
+	http.Get(url)
```
jrfastab, line 532: hah this is great. All the BPF mechanisms to kick a single go line http.Get(url)

jrfastab (approved):

PR #441, opened 2022-09-23
observer: log opcode number instead of name

willfindlay:
There is a mismatch between event numbers in userspace and their opcode numbers from the BPF side, so we can't log opcode names like we've been doing. In any case, the number would be more informative here so let's just do that everywhere.


michi-covalent (approved):

willfindlay:
Yikes that's a lot of CI failures :(

willfindlay:
should be good to go now

PR #443, opened 2022-09-23
tetragon: Warn on sensor bug can flood the logs make it debug

jrfastab:
Flooding the logs with Warnings when a sensors have an error can be rather noisy. Instead these should be behind metrics that we can collected. Anyways lets make it Debug fow now and prode sensor owners to add proper error metrics for each type.

Signed-off-by: John Fastabend <john.fastabend@gmail.com

### DIFF
--- a/pkg/observer/observer.go
+++ b/pkg/observer/observer.go
@@ -118,11 +118,11 @@ func (k *Observer) receiveEvent(data []byte, cpu int) {
 	if err != nil {
 		switch e := err.(type) {
 		case handlePerfUnknownOp:
-			k.log.WithField("opcode", e.op).WithField("event_type", tetragon.EventType(e.op).String()).Warn("unknown opcode ignored")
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):
do we already track these as metrics?

jrfastab:
Not sure we might have related metrics. I think we need to do a audit of the code and check we have metrics in the right spots.

jrfastab:
Merging seeing it was green before and I simply fixed commit sign off.

pkg/observer/observer.go
```diff
@@ -118,11 +118,11 @@ func (k *Observer) receiveEvent(data []byte, cpu int) {
 	if err != nil {
 		switch e := err.(type) {
 		case handlePerfUnknownOp:
-			k.log.WithField("opcode", e.op).WithField("event_type", tetragon.EventType(e.op).String()).Warn("unknown opcode ignored")
+			k.log.WithField("opcode", e.op).WithField("event_type", tetragon.EventType(e.op).String()).Debug("unknown opcode ignored")
```
kkourt, line 121: What was the cause of this? This should not happen, I think.

PR #445, opened 2022-09-23
hubble-fgs: Fix verifier error on non-upstreeam, non-LTS kernels

jrfastab:
RHEL kernels did a backport (I guess.) where they both have the kernfs_inode_id union defined, but instead of using it in the kernfs_inode struct definition they choose to use an anon union.

This breaks our code because when we find kernfs_inode_id union we try to walk through the union with pointer chasing like so, inode->kernfs_indoe_union->id. But, on these kernels the relocation lookup fails so relocation code at load puts a garbage call there to block verivier. And we get this error on load,

 ; if (!bpf_core_type_exists(union kernfs_node_id)) {
 2150: (55) if r1 != 0x0 goto pc+44
 last_idx 2150 first_idx 2146
 regs=2 stack=0 before 2149: (b7) r1 = 1
 ; _(&vfsmnt->mnt_root));
 2195: (85) call unknown#195896080
 invalid func unknown#195896080

So to fix this we make the observation that we can pretend its not a union and just directly read the u64 here regardless of if its a union (older kernels), an anonymous union (RHEL), or a newer kernel with embedded u64. This works because unions only have 'id' value on all known kernels. Of course this is fragile but it appears to work by my reasoning.

### DIFF
--- a/bpf/lib/bpf_cgroup.h
+++ b/bpf/lib/bpf_cgroup.h
@@ -47,18 +47,22 @@ static inline __attribute__((always_inline)) __u64
 __get_cgroup_kn_id(const struct kernfs_node *kn)
 {
 	__u64 id = 0;
-	struct kernfs_node___old *old_kn;
 
 	if (!kn)
 		return id;
 
-	/* Kernels prior to 5.5 have the kernfs_node_id */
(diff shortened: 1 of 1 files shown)


jrfastab:
This passed all tests previously except for a spelling error on checkpatch so once that passes I'll merge this.

PR #449, opened 2022-09-27
bpf: get caps and namespaces in tracepoints

kkourt:
Add caps and namespaces info in tracepoints. This alows to make this information part of the output and, also, have the corresponding matches work.

Signed-off-by: Kornilios Kourtis <kornilios@gmail.com>


tpapagian (approved):
LGTM, thanks!

PR #453, opened 2022-09-28
metrics: add handler error metrics

willfindlay:
This commit adds error metrics for event handler errors so we can tell when specific handlers are failing without flooding the logs with useless log messages. When more info is needed, debug logging can be enabled.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,6 +18,7 @@ require (
 	github.com/jpillora/longestcommon v0.0.0-20161227235612-adb9d91ee629
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/mennanov/fmutils v0.2.0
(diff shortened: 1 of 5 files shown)


kkourt (changes requested):
One of the patches seems (?) to be the same with https://github.com/cilium/tetragon/pull/455. 

I also think that this will break our perfring testing: https://github.com/cilium/tetragon/blob/291c2762039c878003a0ff9fd206216d1a7d233f/pkg/testutils/perfring/perfring.go#L71.

Could you please rebase to ensure that everything is in order?

willfindlay:
@kkourt no need for a rebase, I know what happened. I made that branch's target this branch with the intention that they would be merged together as one depends on the other. So when @jrfastab pulled the trigger on that PR the commit just moved here instead of into main. tl;dr we can just ship it.

PR #456, opened 2022-09-29
tetragon: Separate execve sensor into taill called programs

olsajiri:
The execve is getting more complex and on 4.19 it's reaching
the maximum program size.

Splitting execve sensor into 2 (so far) tail called programs and
adjusting TestLoad* tests.

### DIFF
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -262,6 +262,10 @@ struct msg_execve_event {
 		struct msg_process process;
 		char buffer[PADDED_BUFFER];
 	};
+	/* below fields are not part of the event, serve just as
+	 * heap for execve programs
+	 */
+	__u32 binary;
 }; // All fields aligned so no 'packed' attribute.
 
 struct execve_map_value {
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -11,6 +11,13 @@
 
 char _license[] __attribute__((section("license"), used)) = "GPL";
 
+struct {
+	__uint(type, BPF_MAP_TYPE_PROG_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, __u32);
+	__type(value, __u32);
+} execve_calls SEC(".maps");
+
 #ifdef __LARGE_BPF_PROG
 #include "data_event.h"
 
@@ -157,17 +164,13 @@ event_execve(struct sched_execve_args *ctx)
 {
 	struct task_struct *task = (struct task_struct *)get_current_task();
 	struct msg_execve_event *event;
-	struct execve_map_value *curr, *parent;
+	struct execve_map_value *parent;
 	struct msg_process *execve;
 	uint32_t binary = 0;
 	bool walker = 0;
 	__u32 zero = 0;
-	uint64_t size;
 	__u32 pid;
 	unsigned short fileoff;
-#if defined(__NS_CHANGES_FILTER) || defined(__CAP_CHANGES_FILTER)
-	bool init_curr = 0;
-#endif
 
 	event = map_lookup_elem(&execve_msg_heap_map, &zero);
 	if (!event)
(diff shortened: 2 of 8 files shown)


pkg/testutils/sensors/load.go
```diff
@@ -55,6 +55,63 @@ func findProgram(cache []*prog, name string, typ ebpf.ProgramType, t *testing.T)
 	return nil
 }
 
+func mergeSensorMaps(maps1, maps2 []SensorMap, progs1, progs2 []SensorProg, t *testing.T) ([]SensorMap, []SensorProg) {
```
kkourt, line 59: I guess this function is only intended for  testing?

If so, let's `t *testing.T`  place as the first argument.
Otherwise, let's avoid `t.Fatalf()` and have it return an error.

pkg/testutils/sensors/load.go
```diff
@@ -112,8 +112,33 @@ func mergeSensorMaps(maps1, maps2 []SensorMap, progs1, progs2 []SensorProg, t *t
 	return mapsReturn, progsReturn
 }
 
+func mergeInBaseSensorMaps(sensorMaps []SensorMap, sensorProgs []SensorProg, t *testing.T) ([]SensorMap, []SensorProg) {
```
kkourt, line 116: same with `mergeSensorMaps`: let's add the t *testing.T arg first.
Ideally, all test functions should be in _test.go files, but not sure if that's possible.

pkg/testutils/sensors/load.go
```diff
@@ -132,6 +134,12 @@ func mergeInBaseSensorMaps(sensorMaps []SensorMap, sensorProgs []SensorProg, t *
 		SensorMap{Name: "execve_val", Progs: []uint{2}},
 	}
 
+	if kernels.EnableLargeProgs() {
+		baseMaps = append(baseMaps, SensorMap{Name: "tcpmon_map", Progs: []uint{0, 1, 2, 3}})
+	} else {
+		baseMaps = append(baseMaps, SensorMap{Name: "tcpmon_map", Progs: []uint{1, 2, 3}})
+	}
+
```
kkourt, line 142: Why is this needed?

My reading is:
Program 0 is event_execve, but I don't see how event_execve will ever need tcpmon_map since we seem to be alwats tail-calling into the program that sends the event (execve_send). Am I missing something?

pkg/testutils/sensors/load.go
olsajiri: yes, it's just for testing to merge tests maps, I'll move the 't' arg to the first position

pkg/testutils/sensors/load.go
olsajiri: yes, will move the arg

it's all under `pkg/testutils/sensors/load.go` so testing related

pkg/testutils/sensors/load.go
olsajiri: there's the data interface that sends events and is part of event_execve,
but it's only compiled in for large programs.. it's still not there for 4.19

there's PR for that alreay, but I wanted to split the execve event first

kkourt (approved):

bpf/lib/process.h
```diff
@@ -262,6 +262,10 @@ struct msg_execve_event {
 		struct msg_process process;
 		char buffer[PADDED_BUFFER];
 	};
+	/* below fields are not part of the event, serve just as
+	 * heap for execve programs
+	 */
+	__u32 binary;
```
tixxdz, line 268: Wouldn't be possible to create a new struct keep it local to bpf_execve_event.c file that has both msg_execve_event and the binary embedded inside?

Just to keep msg_execve_event clean and for its purpose and maybe it reduce errors or confusions, yeh you commented it ;-) but we may also start doing same in future and it can get ugly !

tixxdz (approved):
Thank you @olsajiri this solves my problems too for 4.19 and cgroups ;-)

Approved with one comment, I think we shouldn't touch the struct but your call!

bpf/lib/process.h
tixxdz: I mark this resolved since we already do this for kprobes.

bpf/process/bpf_execve_event.c
```diff
@@ -185,10 +188,37 @@ event_execve(struct sched_execve_args *ctx)
 	fileoff = ctx->filename & 0xFFFF;
 	binary = event_filename_builder(ctx, execve, pid, EVENT_EXECVE, binary,
 					(char *)ctx + fileoff);
+	event->binary = binary;
```
jrfastab, line 191: this reads a bit strange I'm not sure why

event->binary = event_filename_builder(...)

is not used.

olsajiri: ugh, will fix that in another PR, thanks

PR #459, opened 2022-10-03
Test

tpapagian:


bpf/process/generic_calls.h
```diff
@@ -130,19 +123,11 @@ generic_process_event1(void *ctx, struct bpf_map_def *heap_map,
...
 	unsigned long a1;
 	int zero = 0;
-	bool walker = 0;
 	long ty, total;
-	__u32 ppid;
-
-	/* Preamble to setup context */
-	enter = event_find_curr(&ppid, &walker);
-	if (!enter)
-		return 0;
```
kkourt, line 145: This is a place where the execve_map_value is used, no? If the value is nil we return. From what I can tell, the value is nill if we cannot find a value matching our ktime and our (or our ancestors') pid in the execve map.

That being said, I think it's fine to remove the code. We are checking the same thing in `generic_process_filter` and return `PFILTER_CURR_NOT_FOUND` which will lead to a PFILTER_REJECT from its caller, but let's update the commit message to reflect that.

pkg/process/process.go
```diff
@@ -286,8 +286,8 @@ func AddCloneEvent(event *tetragonAPI.MsgCloneEvent) error {
 		if pi.process.Pod != nil && pi.process.Pod.Container != nil {
 			pi.process.Pod.Container.Pid = &wrapperspb.UInt32Value{Value: event.NSPID}
 		}
+		procCache.Add(pi)
```
kkourt, line 289: Can you add a short commit message (or even comment) about why is this change needed?
Is `AddCloneEvent` called multiple times until `pi.process` is not nil?

kkourt (approved):
Code changes make sense to me, thanks!

I've added a few comments, but PR LGTM.
Would be useful to have a test for:
```
  clone()             id=1
    kprobe()          id=1
    cleapup_process() id=1
    exec()            id=2
    kprobe()          id=2
    cleapup_process() id=2
    exec()            id=3
  exit()              id=3
```

Now that we have the ringbuffer tests, it should be easy to write such a test. Can be a followup.

bpf/process/generic_calls.h
tpapagian: done!

pkg/process/process.go
tpapagian: done!

jrfastab:
OK this looks good lets try it. @tpapagian can you rebase and merge it thanks!

PR #465, opened 2022-10-06
Add README explaining source of these examples

lizrice:
There are a separate set of example TracingPolicies in this directory associated with Natalia & Jed’s book. This README explains why they are separate from the other examples in the repo. 

Signed-off-by: Liz Rice <liz@lizrice.com>


lizrice:
FYI @xmulligan

Issue #198, opened 2022-06-27
Not getting all outputs when using `kubectl logs` with `tetragon-cli`


ChaosInTheCRD:
As mentioned in the README, I was using `kubectl logs -n kube-system ds/tetragon -c export-stdout -f | tetragon observe` in order to observe the tetragon events. Unfortunately it seems that `kubectl logs` is failing to follow the logs (I think in the event that too many logs are being produced at once.

To combat this, I have used [`stern`](https://github.com/wercker/stern) instead and this seems to be far more reliable:

```
stern -l app.kubernetes.io/name=tetragon -E tetragon -n kube-system -o raw | tetragon observe
```

Maybe it would be better to enable the `tetragon-cli` to not rely on being piped the output of logs from another binary.

Forsworns:
I think that currently `tetragon-cli` only transforms json messages to readable one-line logs, which seems implemented in `pkg/encoder/encoder.go`.

The log is in fact stored in `/run/cilium/tetragon/tetragon.log`. You may verify it via
`kubectl cp kube-system/"$(kubectl get pods -n kube-system | grep tetragon | awk '{print $1}')":/run/cilium/tetragon/tetragon.log ./tetragon.log -c tetragon`.

ChaosInTheCRD:
Hi @Forsworns 😄 

Sure this is the case, but I am simply saying that from a usability perspective, there may be a more optimal way of providing the user with a way of observing the events than what is currently proposed [here](https://github.com/cilium/tetragon#tetragon-cli).

@willfindlay, are their any reasons why `tetragon observe` would not benefit from some extra logic so that it can do its own streaming of the tetragon logs? I know that the Argo Workflows project does this in their CLI (see [here](https://github.com/argoproj/argo-workflows/blob/5eab921eb0f537f1102bbdd6c38b4e52740a88a9/util/logs/workflow-logger.go#L44)), which could be an option.

Alternatively, if this lack of reliability is something that others have observed from using `kubectl logs -n`, could it be considered to suggest a different command to pipe into `tetragon observe`?

PR #469, opened 2022-10-07
make: Add cli-release target

michi-covalent:
Add cli-release Make target that generates CLI release artifacts. Currently it only generates linux/amd64 binary.

Ref: #198


michi-covalent:
note that the tetra CLI still does not support reading events from stdin. i'll open a separate PR for that.

sharlns (approved):

willfindlay (approved):

PR #471, opened 2022-10-10
cgroups: add basic cgroups tracking and make it part of the testing framework

tixxdz:
This PR introduce basic Cgroups tracking. It handles first task of issue: https://github.com/cilium/tetragon/issues/477

The main functionality is `cgroup_mkdir` and `cgroup_rmdir` where we register our bpf programs and expect events to be sent to user space. Note that these programs are not loaded into main Tetragon, we are babysitting them with tests for now, so no new functionality is introduced.

- The patches prefixed with bpf are bpf related parts.
- The following patches are user space handlers for the cgroup events.
- Then we have helpers to make tests easy, followed with tests including bpf unit tests.

### DIFF
--- a/bpf/cgroup/bpf_cgroup_events.h
+++ b/bpf/cgroup/bpf_cgroup_events.h
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright Authors of Tetragon */
+
+#ifndef _BPF_CGROUP_EVENTS__
+#define _BPF_CGROUP_EVENTS__
+
+#include "bpf_helpers.h"
+#include "bpf_events.h"
+#include "environ_conf.h"
+
+/* This function will send the cgroup events to the ring buffer */
+static inline __attribute__((always_inline)) int
+send_cgrp_event(struct bpf_raw_tracepoint_args *ctx,
+		struct cgroup_tracking_value *cgrp_track, __u64 cgrpid,
+		__u32 op)
+{
+	pid_t pid;
+	char *path;
+	int zero = 0;
+	uint64_t size;
+	struct execve_map_value *curr;
+	struct msg_cgroup_event *msg;
+
+	msg = map_lookup_elem(&tg_cgrps_msg_heap, &zero);
+	if (!msg)
+		return 0;
+
+	size = sizeof(struct msg_cgroup_event);
+	msg->common.op = MSG_OP_CGROUP;
+	msg->common.size = size;
+
+	path = (char *)ctx->args[1];
+	pid = (get_current_pid_tgid() >> 32);
+
+	curr = execve_map_get(pid);
+	if (curr) {
+		msg->common.ktime = curr->key.ktime;
+		msg->parent = curr->pkey;
+		msg->flags = curr->flags;
+		msg->ktime = curr->key.ktime;
+	}
+	msg->cgrp_op = op;
+	msg->pid = pid;
+	msg->nspid = get_task_pid_vnr();
+	msg->cgrpid = cgrpid;
+	/* It is same as we are not tracking nested cgroups */
+	msg->cgrpid_tracker = cgrpid;
+	msg->cgrp_data.state = cgrp_track->state;
+	msg->cgrp_data.level = cgrp_track->level;
+	msg->cgrp_data.hierarchy_id = cgrp_track->hierarchy_id;
+	memcpy(&msg->cgrp_data.name, &cgrp_track->name, KN_NAME_LENGTH);
+	probe_read_str(&msg->path, PATH_MAP_SIZE - 1, path);
+
+	perf_event_output(ctx, &tcpmon_map, BPF_F_CURRENT_CPU, msg, size);
+
+	return 0;
+}
+
+#endif
--- a/bpf/cgroup/bpf_cgroup_mkdir.c
+++ b/bpf/cgroup/bpf_cgroup_mkdir.c
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright Authors of Tetragon */
+
+#include "vmlinux.h"
+#include "api.h"
+
+#include "hubble_msg.h"
+#include "bpf_cgroup.h"
+#include "bpf_events.h"
+#include "bpf_cgroup_events.h"
+
+char _license[] __attribute__((section(("license")), used)) = "GPL";
+#ifdef VMLINUX_KERNEL_VERSION
+int _version __attribute__((section(("version")), used)) =
+	VMLINUX_KERNEL_VERSION;
+#endif
+
+__attribute__((section(("raw_tracepoint/cgroup_mkdir")), used)) int
+tg_tp_cgrp_mkdir(struct bpf_raw_tracepoint_args *ctx)
+{
+	int level, zero = 0;
+	uint64_t cgrpid;
+	struct cgroup *cgrp;
+	struct cgroup_tracking_value *cgrp_heap;
+	struct tetragon_conf *config;
+
+	config = map_lookup_elem(&tg_conf_map, &zero);
+	if (!config || config->tg_cgrp_level == 0)
+		return 0;
+
+	cgrp = (struct cgroup *)ctx->args[0];
+
+	level = get_cgroup_level(cgrp);
+	/* This should never happen as the cgroup hierarchy has already been
+	 * set (e.g., by systemd)
+	 */
+	if (level == 0)
+		return 0;
+
+	cgrpid = get_cgroup_id(cgrp);
+	/* This should never happen unless the bpf helper failed */
+	if (cgrpid == 0)
+		return 0;
+
+	/* We want to track all processes of a container system so that we can
+	 * provide proper identity to events. To do that, we use a certain cgroup
+	 * level. Any cgroups that are created under that level, we ignore.
+	 * That is, if we are monitoring level 5, we do not care about cgroup
+	 * events with level >5.
+	 */
+	if (level <= config->tg_cgrp_level) {
+		cgrp_heap = __init_cgrp_tracking_val_heap(cgrp, CGROUP_NEW);
+		if (!cgrp_heap)
+			return 0;
+
+		/* We track only for now cgroups that are at same or above tetragon
+		 * level (ancestors level)
+		 */
+		map_update_elem(&tg_cgrps_tracking_map, &cgrpid, cgrp_heap,
+				BPF_ANY);
+
+		/* We forward bpf events only under TraceLevel */
+		if (unlikely(config->loglevel == LOG_TRACE_LEVEL))
+			send_cgrp_event(ctx, cgrp_heap, cgrpid,
+					MSG_OP_CGROUP_MKDIR);
+	}
+
+	return 0;
+}
--- a/bpf/cgroup/bpf_cgroup_mkdir.c
+++ b/bpf/cgroup/bpf_cgroup_mkdir.c
@@ -18,8 +18,8 @@ int _version __attribute__((section(("version")), used)) =
 __attribute__((section(("raw_tracepoint/cgroup_mkdir")), used)) int
 tg_tp_cgrp_mkdir(struct bpf_raw_tracepoint_args *ctx)
 {
-	int level, zero = 0;
 	uint64_t cgrpid;
+	int level, hierarchy_id, zero = 0;
 	struct cgroup *cgrp;
 	struct cgroup_tracking_value *cgrp_heap;
 	struct tetragon_conf *config;
@@ -29,6 +29,15 @@ tg_tp_cgrp_mkdir(struct bpf_raw_tracepoint_args *ctx)
 		return 0;
 
 	cgrp = (struct cgroup *)ctx->args[0];
+
+	hierarchy_id = get_cgroup_hierarchy_id(cgrp);
+	/*
+	 * In a cgroupv1 setup, there can be multiple cgroup hierarchies but
+	 * we want to track only one If this is not the hierarchy we care
+	 * about, exit.
+	 */
+	if (config->tg_cgrp_hierarchy != hierarchy_id)
+		return 0;
 
 	level = get_cgroup_level(cgrp);
 	/* This should never happen as the cgroup hierarchy has already been
--- a/bpf/cgroup/bpf_cgroup_release.c
+++ b/bpf/cgroup/bpf_cgroup_release.c
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright Authors of Tetragon */
+
+#include "vmlinux.h"
+#include "api.h"
+
+#include "hubble_msg.h"
+#include "bpf_cgroup.h"
+#include "bpf_events.h"
+#include "bpf_cgroup_events.h"
+
+char _license[] __attribute__((section(("license")), used)) = "GPL";
+#ifdef VMLINUX_KERNEL_VERSION
+int _version __attribute__((section(("version")), used)) =
+	VMLINUX_KERNEL_VERSION;
+#endif
+
+/* Ensure to remove tracked cgroups from bpf map */
+__attribute__((section(("raw_tracepoint/cgroup_release")), used)) int
+tg_tp_cgrp_release(struct bpf_raw_tracepoint_args *ctx)
+{
+	int zero = 0;
+	uint64_t cgrpid;
+	struct cgroup *cgrp;
+	struct tetragon_conf *conf;
+	struct cgroup_tracking_value *cgrp_track;
+
+	cgrp = (struct cgroup *)ctx->args[0];
+	cgrpid = get_cgroup_id(cgrp);
+	/* This should never happen unless our helper failed */
+	if (cgrpid == 0)
+		return 0;
+
+	cgrp_track = map_lookup_elem(&tg_cgrps_tracking_map, &cgrpid);
+	/* TODO: check cgroup level if it is under our tracking level
+	 *   then we probably did miss it and should report this.
+	 *   Otherwise the cgroup was never tracked and let's exit.
+	 */
+	if (!cgrp_track)
+		return 0;
+
+	map_delete_elem(&tg_cgrps_tracking_map, &cgrpid);
+
+	conf = map_lookup_elem(&tg_conf_map, &zero);
+	if (!conf)
+		return 0;
+
+	/* We forward bpf events only under TraceLevel */
+	if (unlikely(conf->loglevel == LOG_TRACE_LEVEL))
+		send_cgrp_event(ctx, cgrp_track, cgrpid, MSG_OP_CGROUP_RELEASE);
+
+	return 0;
+}
--- a/bpf/lib/bpf_cgroup.h
+++ b/bpf/lib/bpf_cgroup.h
@@ -202,6 +202,21 @@ get_cgroup_name(const struct cgroup *cgrp)
 }
 
 /**
+ * get_cgroup_level() Returns the cgroup level
+ * @cgrp: target cgroup
+ *
+ * Returns the cgroup level, or 0 if it can not be retrieved.
+ */
+static inline __attribute__((always_inline)) __u32
+get_cgroup_level(const struct cgroup *cgrp)
+{
+	__u32 level = 0;
+
+	probe_read(&level, sizeof(level), _(&cgrp->level));
+	return level;
+}
+
+/**
  * get_cgroup_id() Returns cgroup id
  * @cgrp: target cgroup
  *
@@ -228,7 +243,7 @@ get_cgroup_id(const struct cgroup *cgrp)
  *
  * To get cgroup and kernfs node information we want to operate on the right
  * cgroup hierarchy which is setup by user space. However due to the
- * incompatiblity between cgroup v1 and v2; how user space initialize and
+ * incompatibility between cgroup v1 and v2; how user space initialize and
  * install cgroup controllers, etc, it can be difficult.
  *
  * Use this helper and pass the css index that you consider accurate and
@@ -279,6 +294,56 @@ get_task_cgroup(struct task_struct *task, __u32 subsys_idx)
 
 	probe_read(&cgrp, sizeof(cgrp), _(&subsys->cgroup));
 	return cgrp;
+}
+
+/**
+ * __get_cgrp_tracking_val_heap() Get a cgroup_tracking_val from the
+ * tg_cgrps_tracking_heap map while setting its fields.
+ */
+static inline __attribute__((always_inline)) struct cgroup_tracking_value *
+__get_cgrp_tracking_val_heap(cgroup_state state, __u32 hierarchy_id,
+			     __u32 level)
+{
+	int zero = 0;
+	struct cgroup_tracking_value *heap;
+
+	heap = map_lookup_elem(&tg_cgrps_tracking_heap, &zero);
+	if (!heap)
+		return heap;
+
+	memset(heap, 0, sizeof(struct cgroup_tracking_value));
+	heap->state = state;
+	heap->hierarchy_id = hierarchy_id;
+	heap->level = level;
+
+	return heap;
+}
+
+/**
+ * __init_cgrp_tracking_val_heap() Initialize a cgroup_tracking_val that is
+ * obtained with __get_cgrp_tracking_val_heap(). It will initialize and
+ * set the cgroup name too.
+ */
+static inline __attribute__((always_inline)) struct cgroup_tracking_value *
+__init_cgrp_tracking_val_heap(struct cgroup *cgrp, cgroup_state state)
+{
+	const char *name;
+	struct kernfs_node *kn;
+	__u32 level, hierarchy_id;
+	struct cgroup_tracking_value *heap;
+
+	hierarchy_id = get_cgroup_hierarchy_id(cgrp);
+	level = get_cgroup_level(cgrp);
+	heap = __get_cgrp_tracking_val_heap(state, hierarchy_id, level);
+	if (!heap)
+		return heap;
+
+	kn = __get_cgroup_kn(cgrp);
+	name = __get_cgroup_kn_name(kn);
+	if (name)
+		probe_read_str(&heap->name, KN_NAME_LENGTH - 1, name);
+
+	return heap;
 }
 
 #endif // __BPF_CGROUP_
--- a/bpf/lib/bpf_cgroup.h
+++ b/bpf/lib/bpf_cgroup.h
@@ -10,8 +10,29 @@
 
 #define NULL ((void *)0)
 
+#ifndef CGROUP_SUPER_MAGIC
+#define CGROUP_SUPER_MAGIC 0x27e0eb /* Cgroupv1 pseudo FS */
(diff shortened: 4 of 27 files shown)


bpf/lib/bpf_cgroup.h
```diff
@@ -10,16 +10,88 @@
...
+	__u64 cgrpid_tracker; /* Cgroup ID that is used as a tracker for the current cgroup */
+	__u64 cgrpid; /* Current cgroup ID */
+	struct cgroup_tracking_value cgrp_data; /* Current cgroup data */
+	char path[PATH_MAP_SIZE]; /* Current cgroup path */
+} __attribute__((packed));
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 32768);
+	__type(key, __u64);
```
kkourt, line 77: A comment with what we use as key here (cgroup id) would be great!

bpf/lib/bpf_cgroup.h
```diff
@@ -10,16 +10,91 @@
...
+	/* Unique id for the hierarchy this is mostly for cgroupv1 */
+	__u32 hierarchy_id;
+
+	/* The depth this cgroup is at */
+	__u32 level;
+
+	__u32 pad;
+
+	/* Cgroup kernfs_node name */
+	char name[KN_NAME_LENGTH];
```
jrfastab, line 57: I'm not sure keeping the string name is needed here. What would we use this for? I think ideally this would be in the event and then a cache in userspace would map the name to the id. Then BPF side mainly deals with just numeric ids and avoids strings altogether.

bpf/lib/bpf_cgroup.h
```diff
@@ -281,4 +296,101 @@ get_task_cgroup(struct task_struct *task, __u32 subsys_idx)
...
+{
+	__u32 level;
+	__u64 id = 0;
+
+	if (unlikely(ancestor_level == 0))
+		return id;
+
+#ifdef BPF_FUNC_get_current_ancestor_cgroup_id
+	if (cgrpfs_ver == CGROUP2_SUPER_MAGIC)
+		id = get_current_ancestor_cgroup_id(ancestor_level);
```
jrfastab, line 330: is there a case where get_current_ancestor_cgroup_id would fail? Could this be,

 return get_current_ancestor_cgroup_id(ancestor_level)

then drop the if (id > 0) below.

bpf/cgroup/bpf_cgroup_events.h
```diff
@@ -0,0 +1,80 @@
...
+	if (!msg)
+		return 0;
+
+	size = sizeof(struct msg_cgroup_event);
+	msg->common.op = MSG_OP_CGROUP;
+	msg->common.size = size;
+
+	path = (char *)ctx->args[1];
+	task = (struct task_struct *)get_current_task();
+	probe_read(&pid, sizeof(pid), _(&task->tgid));
```
jrfastab, line 54: why not just use get_pid helper here? Seems the task is not used.

bpf/cgroup/bpf_cgroup_events.h
```diff
@@ -0,0 +1,80 @@
...
+	struct tetragon_conf *config;
+
+	/* First check if this message must be sent or not */
+	if (likely(force == EVENT_NOFORCE_SEND)) {
+		config = map_lookup_elem(&tg_conf_map, &zero);
+		if (!config)
+			return 0;
+
+		/* We forward bpf events only under TraceLevel */
+		if (likely(config->loglevel != LOG_TRACE_LEVEL))
```
jrfastab, line 40: bit of nit, but perhaps you can detect this condition from user space and simply not load the program.

bpf/cgroup/bpf_cgroup_release.c
```diff
@@ -0,0 +1,43 @@
...
+{
+	uint64_t cgrpid;
+	struct cgroup *cgrp;
+	struct cgroup_tracking_value *cgrp_track;
+
+	cgrp = (struct cgroup *)ctx->args[0];
+	cgrpid = get_cgroup_id(cgrp);
+	/* This should never happen unless our helper failed */
+	if (cgrpid == 0)
+		return 0;
```
jrfastab, line 30: if its really an error condition add a metric here eventually. Not needed for this series but future work.

bpf/cgroup/bpf_cgroup_mkdir.c
```diff
@@ -0,0 +1,65 @@
...
+	struct task_struct *task;
+
+	config = map_lookup_elem(&tg_conf_map, &zero);
+	if (!config || config->tg_cgrp_level == 0)
+		return 0;
+
+	cgrp = (struct cgroup *)ctx->args[0];
+
+	task = (struct task_struct *)get_current_task();
+	probe_read(&pid, sizeof(pid), _(&task->tgid));
```
jrfastab: same comment just use pid helper?

pkg/sensors/config/confmap/confmap.go
```diff
@@ -120,11 +124,13 @@ func UpdateTgRuntimeConf(mapDir string, nspid int) error {
 
 	log.WithFields(logrus.Fields{
 		"confmap-update":                configMap.Name,
-		"deployment.mode":               cgroups.DeploymentCode(deployMode).String(),
-		"cgroup.fs.magic":               cgroups.CgroupFsMagicStr(cgroupFsMagic),
+		"deployment.mode":               cgroups.DeploymentCode(v.Mode).String(),
```
jrfastab, line 137: I've got this far but still don't really understand how to set this deployment.mode thing. :/

bpf/lib/bpf_cgroup.h
tixxdz: I tend to agree, we made it this way in beginning so we can project how this could work from the upper layers when doing filtering, will recheck again with @kkourt and probably move it to user space.

bpf/cgroup/bpf_cgroup_events.h
tixxdz: We definitely need the BPF programs to store cgroup info and cleanup the maps, but sending events mkdir rmdir was more for debugging logging.

Now since we may move the cgroup name to user space we have to send mkdir the event anyway

bpf/cgroup/bpf_cgroup_events.h
tixxdz: Will do probably a left over from all the iterations...

tixxdz: > why not just use get_pid helper here? Seems the task is not used.

Will do probably a left over from all the iterations...

bpf/cgroup/bpf_cgroup_release.c
tixxdz: Noted

bpf/lib/bpf_cgroup.h
tixxdz: The bpf_get_current_ancestor_cgroup_id() work on the default hierarchy which is cgroupv2 interface

pkg/sensors/config/confmap/confmap.go
tixxdz: > I've got this far but still don't really understand how to set this deployment.mode thing. :/

It is automatically set, you don't have to do anything, when we start the observer we auto detect all this:

https://github.com/cilium/tetragon/blob/main/pkg/cgroups/cgroups.go#L612

That will trigger findMigrationPath() if it was not run before: https://github.com/cilium/tetragon/blob/92a2503f7788e3707f61c99cd2751667f9b326cc/pkg/cgroups/cgroups.go#L496

That calls Detect Cgroup Mode: https://github.com/cilium/tetragon/blob/92a2503f7788e3707f61c99cd2751667f9b326cc/pkg/cgroups/cgroups.go#L507

Then parse own cgroup paths according to mode, we do separate the hybrid from unified cgroup mode, cause if in hybrid mode systemd will use the cgroupv2 hierarchy only for process tracking where k8s will use legacy cgroupv1 for proper resources tracking. So back we parse the results paths here:

cgroupv1: https://github.com/cilium/tetragon/blob/92a2503f7788e3707f61c99cd2751667f9b326cc/pkg/cgroups/cgroups.go#L324

cgroupv2: https://github.com/cilium/tetragon/blob/92a2503f7788e3707f61c99cd2751667f9b326cc/pkg/cgroups/cgroups.go#L424

Where both will try to detect if we are cgroup namespaced etc, and finally call:
setDeploymentMode() on the final constructed cgroup path of Tetragon https://github.com/cilium/tetragon/blob/92a2503f7788e3707f61c99cd2751667f9b326cc/pkg/cgroups/cgroups.go#L246 

Of course this part of the code can be improved but don't want to touch it until we advance with e2e tests then easy to improve...

bpf/cgroup/bpf_cgroup_mkdir.c
tixxdz: Will do

tixxdz:
> overall looks fine. I think you went a bit too far breaking this into patches. Putting all the BPF progs in one patch would have been fine I don't see much value in splitting all those apart like that. But, only real critique would be to pull the string out of the tracking logic so we don't store strings and then use an id->name from user space. But, consider it a suggestion.

Thank you for the review, for splitting things up yeh, we had already several iterations where mkdir changed depending on cgroup versions and it was improved a lot, that's why on its own (helped me bisect bugs during dev), will wait for @kkourt review then put them together in that case.

For id->name in userspace will go throught and see.

bpf/lib/bpf_cgroup.h
kkourt: After discussing it offline with Djalal, it seems clear that we need a mapping from cgroup ids to names.  There does not seem to be a clear argument for keeping it in bpf and/or user-space. If we keep it in user-space, then we will avoid the syscall. If we keep it in bpf, it would persist agent restarts. Given that subsequent patches depend on having the data in the bpf map, I'd suggest we keep it as is. We can always rethink it later.

bpf/cgroup/bpf_cgroup_events.h
tixxdz: Done.

bpf/cgroup/bpf_cgroup_release.c
```diff
@@ -0,0 +1,43 @@
...
+
+	cgrp_track = map_lookup_elem(&tg_cgrps_tracking_map, &cgrpid);
+	/* This was never tracked exit now */
+	if (!cgrp_track)
+		return 0;
```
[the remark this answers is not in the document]
tixxdz: Hmm we don't track deep nested cgroups, we stop at current tetragon cgroup level or at max 32 int (arbitrary limit), so this could mean two things:
1. The cgroup is deep nested below are limit , could be case but how systemd organize hierarchies 32 is fine, and if unprivileged users keep creating nested cgroupv2 hierarchies we don't care we should attach them all the ancestors that we care about.

2. Maybe race conditions when we read already running process and cgroups that started before tetragon.

In following patches I have one patch that will push those even if we miss them during startup, as long as they do a fork() and execve() we catch and save them in the tracking cgroup map, however if they don't do the above and we miss between reading /proc and cgrooup_mkdir tracepoint being really active then yeh!

So yeh solution here check the level of the cgroup if it is not over our tracked limit then we did probably miss it, and should report this. Note added.

bpf/cgroup/bpf_cgroup_mkdir.c
tixxdz: Done.

bpf/lib/bpf_cgroup.h
tixxdz: Thanks @kkourt @jrfastab

pkg/sensors/exec/cgroups_test.go
```diff
@@ -67,3 +129,99 @@ func TestTgRuntimeConf(t *testing.T) {
...
+		if err != nil {
+			fmt.Printf("disableSensor failed: %s\n", err)
+		}
+	}()
+
+	val, err := testutils.GetTgRuntimeConf()
+	assert.NoError(t, err)
+
+	// Set Log Level to trace so we receive BPF events
+	val.LogLevel = uint32(logrus.TraceLevel)
```
kkourt, line 190: If set set the trace level, why are not we expected to receive bpf events? Is it because they are happening in a level that we are not monitoring?

pkg/sensors/exec/cgroups_test.go
```diff
@@ -225,3 +232,151 @@ func TestCgroupNoEvents(t *testing.T) {
 		}
 	}
 }
+
+// Ensure that we get cgroup_{mkdir|rmdir} events
+func TestCgroupEventMkdirRmdir(t *testing.T) {
```
kkourt, line 237: What is the difference between the test in `TestCgroupEventMkdirRmdir` and `TestCgroupNoEvents`? They seem to share a lot of initialization. Is there a way to have a common function that tests both?

Potentially using t.Run to create a new subtest?

If not, then maybe we can have a common function for their initialization?

kkourt (commented):
Thanks!

Most of the things look good to me.

It was not an easy PR to review though because many patches introduced something that was either used or tested many patches later. Also, I had to frequently check #225 to figure out why something exists.

That being said, I think the code looks good. I have some comments, but most of them are about improving small aspects of the code (e.g., adding comments).

pkg/sensors/test/cgroups.go
```diff
@@ -0,0 +1,72 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package test
+
```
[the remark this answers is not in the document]
tixxdz: Indeed!

pkg/testutils/confmap.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/cilium/tetragon/pkg/sensors/base"
+	"github.com/cilium/tetragon/pkg/sensors/config/confmap"
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	log = logger.GetLogger()
+)
+
+func GetTgRuntimeConf() (*confmap.TetragonConfValue, error) {
```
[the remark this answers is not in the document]
tixxdz: The corresponding part is here: https://github.com/cilium/tetragon/blob/main/pkg/sensors/config/confmap/confmap.go and I explicilty made it in test so we don't mess up with the other package.

I will add comments that this is explicity for testing where we read Tetragon Runtime Config to assert it, in main application we don't have to read it as we are the ones that set it and we shouldn't mess it up; in production we read current deployment environment and set the config.

Is this fine?

pkg/testutils/confmap.go
```diff
@@ -0,0 +1,97 @@
...
+		LogLevel:        uint32(logger.GetLogLevel()),
+		TgCgrpHierarchy: cgroups.GetCgrpHierarchyID(),
+		TgCgrpSubsysIdx: cgroups.GetCgrpSubsystemIdx(),
+		NSPID:           uint32(nspid),
+		CgrpFsMagic:     cgroupFsMagic,
+	}, nil
+}
+
+// Test `tg_conf_map` BPF MAP with explicit values
+func UpdateTgRuntimeConf(mapDir string, v *confmap.TetragonConfValue) error {
```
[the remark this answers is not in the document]
tixxdz: Yes same as above, these are only for testing, I added a comment test map with explicit values!

In reality we are not going to change values of the map, we adapt to the deployment environment where Tetragon is running so need to have these setters in the main packages.

Thank you!

pkg/sensors/exec/cgroups_test.go
```diff
@@ -67,3 +129,99 @@ func TestTgRuntimeConf(t *testing.T) {
...
+
+	registerSensors(loadedSensors)
+
+	mapDir := bpf.MapPrefixPath()
+	smanager, err := sensors.StartSensorManager(mapDir, mapDir, "")
+	if err != nil {
+		t.Fatalf("startSensorController failed: %s", err)
+	}
+	observer.SensorManager = smanager
+	defer func() {
```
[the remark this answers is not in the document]
tixxdz: Yes will do thank you!

pkg/sensors/exec/cgroups_test.go
tixxdz: So yes we may not receive events if they happen at level that we do not care about, but for this case we do not expect cgroup events cause we do not probe tetragon to migrate itself and finish the cgroup discovery.

We only detect environment and half of the cgroup setup, but not all of it.

See below
`val.TgCgrpLevel = 3`

pkg/sensors/exec/cgroups_test.go
tixxdz: The difference is that  `TestCgroupEventMkdirRmdir`  has `val.TgCgrpLevel = 3` set which is in main tetragon is set when we complete cgroup discovery when probing tetragon cgroup migration.

So if that one is set we expect an event, if it is not set means no events.

pkg/sensors/exec/cgroups_test.go
```diff
@@ -15,11 +22,66 @@ import (
...
+	tetragonCgrpRoot = "tetragon-tests"
+)
+
+func registerSensors(loaded []*sensors.Sensor) {
+	for _, s := range loaded {
+		sensors.RegisterSensorAtInit(s)
+	}
+}
+
+func enableSensors(ctx context.Context, smanager *sensors.Manager, loaded []*sensors.Sensor) error {
```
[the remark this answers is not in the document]
tixxdz: Ok will check and see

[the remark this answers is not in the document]
tixxdz: Yes done, merged functions together and using ones from pkg/testuitls/sensors

pkg/sensors/exec/cgroups_test.go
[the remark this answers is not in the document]
tixxdz: Done

pkg/sensors/exec/cgroups_test.go
tixxdz: Improved this, tests are more readable now!

bpf/lib/bpf_cgroup.h
```diff
@@ -10,16 +10,91 @@
...
 /* Represent old kernfs node with the kernfs_node_id
  * union to read the id in 5.4 kernels and older
  */
 struct kernfs_node___old {
 	union kernfs_node_id id;
 };
 
+struct cgroup_tracking_value {
+	/* State of cgroup */
+	cgroup_state state;
```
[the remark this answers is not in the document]
tixxdz: Hmm yes my memory tricked me... enums can be small, however for our case and targeted platform as discussed it will always be 4 bytes and we don't do short enums anyway, also we have that align checker, so I think we are good.

bpf/lib/bpf_cgroup.h
```diff
@@ -10,16 +10,91 @@
...
+
+	/* Unique id for the hierarchy this is mostly for cgroupv1 */
+	__u32 hierarchy_id;
+
+	/* The depth this cgroup is at */
+	__u32 level;
+
+	__u32 pad;
```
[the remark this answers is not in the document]
tixxdz: For me I prefer padding so I can ensure that I initialize memory especially if we do alignment later and we couple this with memcpy + sizeof() , there are also some other use of it but Im not sure why... will mark it resolved.

pkg/cgroups/cgroups.go
```diff
@@ -661,3 +661,13 @@ func DetectCgroupFSMagic() (uint64, error) {
 
 	return cgroupFSMagic, nil
 }
+
+// CgroupNameFromCstr() Returns a Golang string from the passed C language format string.
+func CgroupNameFromCStr(cstr []byte) string {
```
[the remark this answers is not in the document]
tixxdz: done

pkg/sensors/exec/cgroups_test.go
tixxdz: Ok I improved the test now and explained why how we set the cgroup tracking level, marking it resolved.

pkg/testutils/confmap.go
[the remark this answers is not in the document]
tixxdz: This will be done in follow up PR, thanks!

bpf/lib/bpf_cgroup.h
```diff
@@ -281,4 +296,101 @@ get_task_cgroup(struct task_struct *task, __u32 subsys_idx)
...
+ * This helper works for both Cgroupv1 and Cgroupv2. The root
+ * cgroup is at ancestor_level zero and each step down the
+ * hierarchy increments the level.
+ *
+ * If ancestor_level == level of passed cgroup, then return value
+ * will be the same as that of get_cgroup_id().
+ */
+static inline __attribute__((always_inline)) __u64
+get_ancestor_cgroup_id(const struct cgroup *cgrp, __u64 cgrpfs_ver,
+		       __u32 ancestor_level)
```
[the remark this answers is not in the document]
tixxdz: Ok removed for now

bpf/lib/environ_conf.h
```diff
@@ -6,8 +6,16 @@
 
 /* Tetragon runtime configuration */
 struct tetragon_conf {
+	__u32 mode; /* Tetragon deployment mode */
```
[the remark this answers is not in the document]
tixxdz: Ok removed, will add it later, the deployment detection logic is already shipped anyway and it is under log warning and does not fail... ;-) so real testing :-X  !

bpf/cgroup/bpf_cgroup_events.h
```diff
@@ -0,0 +1,78 @@
...
+
+/* This function will send the cgroup events to the ring buffer
+ * if Tetragon is running with log level set to LOG_TRACE_LEVEL.
+ * If the parameter force is set to EVENT_FORCE_SEND, then the
+ * event will be forwarded reguardless of the log level.
+ */
+static inline __attribute__((always_inline)) int
+send_cgrp_event(struct bpf_raw_tracepoint_args *ctx,
+		struct cgroup_tracking_value *cgrp_track, __u64 cgrpid,
+		__u32 op, bool force)
```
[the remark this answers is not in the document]
tixxdz: done

bpf/cgroup/bpf_cgroup_mkdir.c
```diff
@@ -0,0 +1,65 @@
...
+
+	cgrp = (struct cgroup *)ctx->args[0];
+
+	hierarchy_id = get_cgroup_hierarchy_id(cgrp);
+	/* Are we operating on the corresponding hierarchy? if no exit */
+	if (config->tg_cgrp_hierarchy != hierarchy_id)
+		return 0;
+
+	level = get_cgroup_level(cgrp);
+	/* This should never happen as the cgroup hierarchy has already been set */
```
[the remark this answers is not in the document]
tixxdz: done, thank you!

bpf/lib/msg_types.h
```diff
@@ -26,6 +26,17 @@ enum msg_ops {
...
+	MSG_OP_CGROUP = 25,
+
 	MSG_OP_MAX,
 };
+
+enum msg_cgroup_ops {
+	MSG_OP_CGROUP_UNDEF = 0,
+	MSG_OP_CGROUP_MKDIR = 1, /* cgroup_mkdir tracepoint */
+	MSG_OP_CGROUP_RMDIR = 2, /* cgroup_rmdir tracepoint */
+	MSG_OP_CGROUP_RELEASE = 3, /* cgroup_release tracepoint */
```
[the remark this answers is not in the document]
tixxdz: done

bpf/cgroup/bpf_cgroup_mkdir.c
```diff
@@ -0,0 +1,65 @@
...
+		 * level (ancestors level)
+		 */
+		map_update_elem(&tg_cgrps_tracking_map, &cgrpid, cgrp_heap,
+				BPF_ANY);
+
+		/* Notify events only about cgroup being tracked */
+		send_cgrp_event(ctx, cgrp_heap, cgrpid, MSG_OP_CGROUP_MKDIR,
+				EVENT_NOFORCE_SEND);
+	}
+
```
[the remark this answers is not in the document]
tixxdz: Done, yes this way it is clear for all developers, thank you!

bpf/cgroup/bpf_cgroup_mkdir.c
```diff
@@ -0,0 +1,65 @@
...
+	struct tetragon_conf *config;
+
+	config = map_lookup_elem(&tg_conf_map, &zero);
+	if (!config || config->tg_cgrp_level == 0)
+		return 0;
+
+	cgrp = (struct cgroup *)ctx->args[0];
+
+	hierarchy_id = get_cgroup_hierarchy_id(cgrp);
+	/* Are we operating on the corresponding hierarchy? if no exit */
```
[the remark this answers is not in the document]
tixxdz: Done

tixxdz:
> Thanks!
> 
> Most of the things look good to me.
> 
> It was not an easy PR to review though because many patches introduced something that was either used or tested many patches later. Also, I had to frequently check #225 to figure out why something exists.
> 
> I have some comments, but most of them are about improving small aspects of the code (e.g., adding comments).

Thanks @kkourt for the review, all comments handled!

pkg/sensors/config/confmap/confmap.go
tixxdz: This was removed from this PR, and it will be re-introduced in follow up one.

bpf/cgroup/bpf_cgroup_mkdir.c
```diff
@@ -0,0 +1,69 @@
...
+	 * events with level >5.
+	 */
+	if (level <= config->tg_cgrp_level) {
+		cgrp_heap = __init_cgrp_tracking_val_heap(cgrp, CGROUP_NEW);
+		if (!cgrp_heap)
+			return 0;
+
+		/* We track only for now cgroups that are at same or above tetragon
+		 * level (ancestors level)
+		 */
```
kkourt, line 67: I think we could remove this comment now.

kkourt (approved):
Nice work! 🚢 🚀

bpf/cgroup/bpf_cgroup_mkdir.c
tixxdz: Yes we could but it doesn't hurt to have it as it is correct too, so marking resolved!

tixxdz: Thank you!

bpf/cgroup/bpf_cgroup_events.h
tixxdz: Just an update here: we send events for debugging but also to detect tetragon cgroup config and assert it at user space!

the bpf programs have to be loaded anyway as they keep track of cgroup info inside the cgroup tracking map, which we will need later to do proper filtering.

bpf/lib/bpf_cgroup.h
tixxdz: @jrfastab just an update here, I got confused we definitely need this at the bpf map since when we generate events (exec) we do get the tracking cgroup ID => lookup into the bpf map and find the tracking cgroup name potentially pod or container ID, and in next PR we will send this info part of the event: tracking cgroup ID + tracing cgroup Name

PR #473, opened 2022-10-11
tests:perfring: allow to cancel context and close perfring buffer

tixxdz:
When waiting for events from the ring buffer we do not check if the context was canceled this may cause some tests to continue to run blocked at perfReader.Read() indefinitly until the go unit test timeout is reached.

Fix it by checking if the context is done, close the ring buffer and clean up related goroutines.

### DIFF
--- a/pkg/testutils/perfring/perfring.go
+++ b/pkg/testutils/perfring/perfring.go
@@ -10,6 +10,7 @@ package perfring
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"testing"
 
@@ -25,8 +26,24 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-//  EventFn is the type of function called by ProcessEvents for each event.
+// EventFn is the type of function called by ProcessEvents for each event.
 type EventFn func(ev notify.Message) error
+
+func loopEvents(events []observer.Event, eventFn EventFn, complChecker *testsensor.CompletionChecker) error {
+	for _, ev := range events {
+		switch xev := ev.(type) {
+		case *testapi.MsgTestEventUnix:
+			cpu := xev.Arg0
+			complChecker.Update(cpu)
+		}
+
+		if err := eventFn(ev); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
 
 // ProcessEvents will open the perf ringbuffer and process events.
 //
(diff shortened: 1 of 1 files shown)


tixxdz:
So ProcessEvents() here: https://github.com/cilium/tetragon/blob/main/pkg/testutils/perfring/perfring.go#L33

// It will complete, whenever it sees a number of MsgTestEventUnix on _all_
// cpus or when the ctx is done.  Hence, callers need to load the test sensor
// before they call it or use the ctx to cancel its operation.

That's not 100% right as this function does not handle cancel ctx, so if we don't load the test sensor to trigger https://github.com/cilium/tetragon/blob/main/pkg/testutils/perfring/perfring.go#L107  `testsensor.TestCheckerMarkEnd(t)` we may get stuck here: https://github.com/cilium/tetragon/blob/main/pkg/testutils/perfring/perfring.go#L66

One innocent test that did not load the test sensor ;-)    https://github.com/cilium/tetragon/pull/471/commits/47bc33d9534497b2ad5c9c952033067d74c74b4e#diff-6bd55012117f495f1a6932e5976d718519597ecdc7010851b329e2a248408c22R104

pkg/testutils/perfring/perfring.go
```diff
@@ -55,37 +72,69 @@ func ProcessEvents(t *testing.T, ctx context.Context, eventFn EventFn, wgStarted
...
 			}
 		}
-		if complChecker.Done() {
-			break
+	}()
+
+	for {
+		select {
+		case err := <-errChan:
+			t.Fatal(err)
```
kkourt, line 127: Any reason why we are calling Fatal here?

pkg/testutils/perfring/perfring.go
```diff
@@ -25,19 +26,35 @@ import (
...
 // before they call it or use the ctx to cancel its operation.
 //
 // Example code to load the test sensor:
-//     import testsensor "github.com/cilium/tetragon/pkg/sensors/test"
-//     import tus "github.com/cilium/tetragon/pkg/testutils/sensors"
-//     tus.LoadSensor(ctx, t, testsensor.GetTestSensor())
+//
+//	import testsensor "github.com/cilium/tetragon/pkg/sensors/test"
+//	import tus "github.com/cilium/tetragon/pkg/testutils/sensors"
+//	tus.LoadSensor(ctx, t, testsensor.GetTestSensor())
```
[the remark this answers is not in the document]
tixxdz: I think this is some go fmt thingy...

pkg/testutils/perfring/perfring.go
tixxdz: Original code was calling fatal , didn't change semantics, just added the missing sync logic

kkourt (approved):

michi-covalent (approved):

PR #475, opened 2022-10-12
Simplify the interaction with the EventCache

tpapagian:
This PR does a cleanup on the way the events interact with the EventCache. It also defines new functions AddProcessParent and AddProcessParentRef{Inc, Dec} that allow events to annotate events with process and parent information and also add them in the eventcache in the case where we miss something.

It also provides ways to use a custom way to interact with the EventCache for special cases like exec and clone events.

The intention of this PR is not to change the functionality but to cleanup the codebase.

### DIFF
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -8,7 +8,7 @@ import (
 	"time"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
-	"github.com/cilium/tetragon/pkg/ktime"
+	kt "github.com/cilium/tetragon/pkg/ktime"
 	"github.com/cilium/tetragon/pkg/metrics/errormetrics"
 	"github.com/cilium/tetragon/pkg/metrics/eventcachemetrics"
 	"github.com/cilium/tetragon/pkg/metrics/mapmetrics"
@@ -17,6 +17,7 @@ import (
 	"github.com/cilium/tetragon/pkg/reader/node"
 	"github.com/cilium/tetragon/pkg/reader/notify"
 	"github.com/cilium/tetragon/pkg/server"
+	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
 const (
@@ -31,12 +32,18 @@ var (
 	nodeName string
 )
 
+type retryFn func(notify.Event, *notify.CacheActions, uint32, uint64) error
+type procFn func(*process.ProcessInternal)
+
 type CacheObj struct {
-	internal  *process.ProcessInternal
 	event     notify.Event
-	timestamp uint64
+	pid       uint32 // process pid
+	ktime     uint64 // process ktime
+	timestamp uint64 // event ktime
 	color     int
 	msg       notify.Message
+	actions   *notify.CacheActions
+	retry     retryFn
 }
 
 type Cache struct {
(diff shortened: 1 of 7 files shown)


pkg/eventcache/eventcache.go
```diff
@@ -53,81 +60,183 @@ var (
...
+		tetragonParent = parent.GetProcessCopy()
+		e.SetParent(tetragonParent)
+	}
+
+	act := &notify.CacheActions{
+		NeedProcess:    NeededProcess(tetragonProcess),
+		NeedProcessPod: NeededPod(tetragonProcess),
+		NeedParent:     pid > 1 && NeededProcess(tetragonParent),
+		NeedParentPod:  pid > 1 && NeededPod(tetragonParent),
 	}
```
kkourt, line 124: How about we call the pFn only if NeedProcess is not set instead of if process is not nill?
pFn is called in the retry if `NeedProcess` or `NeedParent` is set.

But:
```
func NeededProcess(proc *tetragon.Process) bool {
	if proc == nil {
		return true
	}
	if proc.Binary == "" {
		return true
	}
	return false
}
```

so if  p != nil but p.Binary == "", I think we would call pFn twice. Once here and once whenever things are setup in the cache.

The same thing holds for the parent.

pkg/reader/notify/notify.go
```diff
@@ -5,17 +5,29 @@ import (
...
 )
 
+type CacheActions struct {
+	NeedProcess    bool
+	NeedProcessPod bool
+	NeedParent     bool
+	NeedParentPod  bool
+}
+
 type Message interface {
```
kkourt, line 17:
```suggestion
// Message is what messages need to implement.
// Moreover, messages may opt to implement the CacheRetry and PostProcessing interfaces (see below).
type Message interface {
```

pkg/eventcache/eventcache.go
```diff
@@ -53,81 +60,183 @@ var (
...
-		// now because it should be available. Otherwise we have a valid
-		// process and lets copy it across.
-		if event.internal == nil {
-			event.internal, err = event.msg.RetryInternal(event.event, event.timestamp)
-		}
-		if err == nil {
-			err = event.msg.Retry(event.internal, event.event)
+
+		retryFn := event.retry                          // in that case we use the default retry function (i.e. GenericRetry)
+		if c, ok := event.msg.(notify.CacheRetry); ok { // if the event.msg implements notify.CacheRetry we will use that instead
```
kkourt, line 207:
```suggestion
		retryFn := event.retry                                         // by default we use GenericRetry
		if c, ok := event.msg.(notify.CacheRetry); ok { // if underlying message implements notify.CacheRetry use that instead
```

pkg/grpc/exec/exec.go
```diff
@@ -166,202 +175,148 @@ type MsgCloneEventUnix struct {
...
+var _ notify.CacheRetry = (*MsgCloneEventUnix)(nil)
+
 func (msg *MsgCloneEventUnix) Notify() bool {
 	return false
 }
 
-func (msg *MsgCloneEventUnix) RetryInternal(ev notify.Event, timestamp uint64) (*process.ProcessInternal, error) {
-	return nil, process.AddCloneEvent(&msg.MsgCloneEvent)
-}
+func (msg *MsgCloneEventUnix) Retry(ev notify.Event, ca *notify.CacheActions, pid uint32, timestamp uint64) error {
```
kkourt, line 184: Can we add a comment on why this needs its own retry function? The same for the other events with a custom function.

pkg/reader/notify/notify.go
```diff
@@ -5,17 +5,29 @@ import (
...
 	HandleMessage() *tetragon.GetEventsResponse
-	RetryInternal(Event, uint64) (*process.ProcessInternal, error)
-	Retry(*process.ProcessInternal, Event) error
 	Notify() bool
 	Cast(o interface{}) Message
 }
 
+type CacheRetry interface {
```
kkourt, line 23:
```suggestion
// CacheRetry may (optionaly) be implemented by messages that requiring custom cache retry functionality.
type CacheRetry interface {
```

kkourt (changes requested):
Thanks!

Looks good to me.

I have a few minor suggestions, and what I think is a double reference count increase. With the code cleanup a lot of the logic is expressed better so it makes it easier to discuss the inner workings of the cache(s), which is great!

pkg/eventcache/eventcache.go
tpapagian: I see what you are saying there. I don't believe that we can have an entry in the cache (i.e. `proc != nil`) but it will be not full (i.e. `proc.Binary == ""`). 

I will change that to what you propose just to be sure that I don't miss any corner cases.

kkourt: > I see what you are saying there. I don't believe that we can have an entry in the cache (i.e. proc != nil) but it will be not full (i.e. proc.Binary == "").

Not sure I understand.

Are you saying that there is no way to to have an entry in the cache with Binary == ""?

tpapagian: Yes, I believe that this cannot happen but I want to check that again.

tpapagian:
I have updated this PR based on the review.

kkourt:
@tpapagian apologies but it took me a while to get to this PR, and it seems that it needs again a rebase. Could you please rebase it again?

tpapagian:
Moving to draft as I want to reconsider some cases. Will rebase once I finish with that.

PR #482, opened 2022-10-17
tetragon: use raw tracepoint for execve

jrfastab:
Raw tracepoints are faster than tracepoints so lets use them for execve hook.


kkourt (commented):
Just a note that I think this will set our kernel requirement to 4.17 (which is fine by me).

PR #160, opened 2022-06-14
support raw_syscalls tracepoint

kkourt:
This PR adds support for the raw_syscalls/sys_enter tracepoint. We need two things for doing so:
* support for long arguments
* support for arrays

For the second, we add a constant buffer type, where its size is passed as the metadata argument. Please see individual commits.

PR #483, opened 2022-10-18
Define syscallNames for darwin

michi-covalent:
Define syscallNames so that the tetra CLI compiles on Intel and ARM Mac.


michi-covalent:
🥺

sharlns (approved):

olsajiri:
ah sry did not make it in time.. fwiw looks good ;-)

kkourt:
Also late to the party, but I'm not  sure if that's the proper thing to do for the CLI. https://github.com/cilium/tetragon/pull/160/commits/f56c0f5e35455208f8190285d70f4f1de8bb0595, added https://github.com/cilium/tetragon/blob/27facf9a04c32e3d6f391ed824f915616e65d10b/pkg/encoder/encoder.go#L287

Which translates system call ids to names, and it uses that to figure out the arguments as well. So (at least in the above case) we would need the syscallNames of Linux (rather than dwarvin). Not sure what the best solution for this. It would probaby be better if we did this translation (id->name) on the agent.

tklauser:
Late to the party as well. As Kornilios mentioned, these should be syscall names for Linux and the correct `$GOARCH` rather than macOS. I agree that doing the translation id->name in the agent is probably most straight forward as the `$GOARCH` is known. Ideally, we'd add something like `SyscallName` (akin to [`SignalName`](https://pkg.go.dev/golang.org/x/sys/unix#SignalName)) to the `golang.org/x/sys/unix` package which would do the translation.

kkourt:
added issue to track this: https://github.com/cilium/tetragon/issues/539

PR #497, opened 2022-10-20
Remove pidMap

tpapagian:
pidMap holds mappings from PID to the most recent exec ID for the PID. This is used to find the parent of exec events that come after another exec event (i.e. not a clone). This can be racy as events can come OOO to the user space.

This information is also available in the execve_map inside the kernel where all events happen in-order.

https://github.com/cilium/tetragon/commit/45745a0a93c65811495f82deb9314a1c849e1403 introduced a new field in msg_execve_event (cleanup_key) that keeps the parent of the execve event. In that case this was needed for cleanup.

We can use the same information for the parent and remove entirely the pidMap. If there is a case where we cannot find the entry in the execve_map, we use as a parent the view of Linux (i.e. the process that clones the current process and not the previous execve).

This PR also contains small fixes to other components in the execve path.

### DIFF
--- a/pkg/process/cache.go
+++ b/pkg/process/cache.go
@@ -169,7 +169,7 @@ func (pc *Cache) get(processID string) (*ProcessInternal, error) {
 		errormetrics.ErrorTotalInc(errormetrics.ProcessCacheMissOnGet)
 		return nil, fmt.Errorf("invalid entry for process ID: %s", processID)
 	}
-	process, _ := entry.(*ProcessInternal)
+	process, ok := entry.(*ProcessInternal)
 	if !ok {
 		logger.GetLogger().WithField("process entry", entry).Debug("invalid entry in process cache")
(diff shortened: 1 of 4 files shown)


pkg/process/cache.go
```diff
@@ -169,7 +169,7 @@ func (pc *Cache) get(processID string) (*ProcessInternal, error) {
 		errormetrics.ErrorTotalInc(errormetrics.ProcessCacheMissOnGet)
 		return nil, fmt.Errorf("invalid entry for process ID: %s", processID)
 	}
-	process, _ := entry.(*ProcessInternal)
+	process, ok := entry.(*ProcessInternal)
```
kkourt, line 160: nice catch!

kkourt (approved):
Awesome!
Love PRs that simplify the code!

michi-covalent (approved):

PR #499, opened 2022-10-20
tetragon: detect execve of anonymous binaries

tixxdz:
Detect execution of anonymous binaries, these are binaries that are not linked on a filesystem, where nlink is zero.

1. An anonymous shared memory file
       https://man7.org/linux/man-pages/man7/shm_overview.7.html.
2. An anonymous file obtained with memfd API
       https://man7.org/linux/man-pages/man2/memfd_create.2.html.
3. Or it was deleted from the file system.

```
"process_exec": {
   "process": {
      "binary_properties": {
        "file": {
           "inode": {
              "number": "182698",
              "links": 0
            }
         }
      }
   }
}
```

### DIFF
--- a/api/v1/tetragon/tetragon.proto
+++ b/api/v1/tetragon/tetragon.proto
@@ -136,6 +136,20 @@ message ProcessCredentials {
 	UserNamespace user_ns = 11;
 }
 
+message Inode {
+	// The inode number
+	uint64 number = 1;
+	// The inode links on the file system. If zero means the file is only in memory
+	google.protobuf.UInt32Value links = 2;
+}
+
+message FileProperties {
+	// Inode of the file
+	Inode inode = 1;
+	// Path of the file
+	string path = 2;
+}
+
 message BinaryProperties {
 	// If set then this is the set user ID used for execution
 	google.protobuf.UInt32Value setuid = 1;
@@ -145,6 +159,11 @@ message BinaryProperties {
 	// a binary with the set-user-ID to root or file capability sets.
 	// The final granted privileges can be listed inside the `process_credentials` or capabilities fields part of of the `process` object.
 	repeated ProcessPrivilegesChanged privileges_changed = 3;
+	// File properties in case the executed binary is:
+	// 1. An anonymous shared memory file https://man7.org/linux/man-pages/man7/shm_overview.7.html.
+	// 2. An anonymous file obtained with memfd API https://man7.org/linux/man-pages/man2/memfd_create.2.html.
+	// 3. Or it was deleted from the file system.
+	FileProperties file = 4;
 }
 
 message Process {
--- a/bpf/process/bpf_execve_bprm_commit_creds.c
+++ b/bpf/process/bpf_execve_bprm_commit_creds.c
@@ -49,13 +49,9 @@ BPF_KPROBE(tg_kp_bprm_committing_creds, struct linux_binprm *bprm)
 	struct execve_map_value *curr;
 	struct execve_heap *heap;
 	struct task_struct *task;
+	struct file *file;
 	__u32 pid, ruid, euid, uid, egid, gid, sec = 0, zero = 0;
 	__u64 tid, permitted, new_permitted, new_ambient = 0;
-
-	sec = BPF_CORE_READ(bprm, per_clear);
-	/* If no flags to clear then this is not a privileged execution */
-	if (!sec)
-		return;
 
 	tid = get_current_pid_tgid();
 	pid = (tid >> 32);
@@ -69,6 +65,20 @@ BPF_KPROBE(tg_kp_bprm_committing_creds, struct linux_binprm *bprm)
 		return;
 
 	memset(&heap->info, 0, sizeof(struct execve_info));
+
+	/* Read binary file information */
+	if (BPF_CORE_READ_INTO(&file, bprm, file) != 0)
+		return;
+
+	if (BPF_CORE_READ_INTO(&heap->info.i_nlink, file, f_inode, __i_nlink) != 0)
+		return;
+
+	if (BPF_CORE_READ_INTO(&heap->info.i_ino, file, f_inode, i_ino) != 0)
+		return;
+
+	/* If no flags to clear then this is not a privileged execution */
+	if (BPF_CORE_READ_INTO(&sec, bprm, per_clear) != 0 || sec == 0)
+		goto out;
 
 	/* Check if this is a setuid or setgid */
 	euid = BPF_CORE_READ(bprm, cred, euid.val);
@@ -123,7 +133,7 @@ BPF_KPROBE(tg_kp_bprm_committing_creds, struct linux_binprm *bprm)
 	 */
 	BPF_CORE_READ_INTO(&new_ambient, bprm, cred, cap_ambient);
 	if (new_ambient)
-		return;
+		goto out;
 
 	/* Did we gain new capabilities through execve?
 	 *
@@ -141,7 +151,11 @@ BPF_KPROBE(tg_kp_bprm_committing_creds, struct linux_binprm *bprm)
 		heap->info.secureexec |= EXEC_FILE_CAPS;
 	}
 
-	/* Do we cache the entry? */
-	if (heap->info.secureexec != 0)
+out:
+	/* We cache the entry if:
+	 * 1. Privileged execution so secureexec will be set.
+	 * 2. Execution of an unlinked binary
+	 */
+	if (heap->info.secureexec != 0 || (heap->info.i_nlink == 0 && heap->info.i_ino != 0))
 		execve_joined_info_map_set(tid, &heap->info);
 }
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -130,18 +130,23 @@ read_cwd(void *ctx, struct msg_process *p)
 	return getcwd(p, p->size, p->pid);
 }
 
-static inline __attribute__((always_inline)) __u32
-read_execve_shared_info(void *ctx, __u64 pid)
+static inline __attribute__((always_inline)) void
+read_execve_shared_info(void *ctx, struct msg_process *p, __u64 pid)
 {
-	__u32 secureexec = 0;
 	struct execve_info *info;
 
 	info = execve_joined_info_map_get(pid);
-	if (info) {
-		secureexec = info->secureexec;
-		execve_joined_info_map_clear(pid);
+	if (!info) {
+		p->secureexec = 0;
+		p->i_ino = 0;
+		p->i_nlink = 0;
+		return;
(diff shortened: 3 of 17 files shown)


bpf/process/bpf_execve_bprm_check.c
```diff
@@ -0,0 +1,39 @@
...
+	heap = map_lookup_elem(&execve_heap, &zero);
+	if (!heap)
+		return 0;
+
+	probe_read(&file, sizeof(file), _(&bprm->file));
+	links = BPF_CORE_READ(file, f_path.dentry, d_inode, __i_nlink);
+	heap->info.inode.i_nlink = links;
+	heap->info.inode.initialized = 1;
+
+	execve_info_map_set(pid, &heap->info);
```
jrfastab, line 36: should this be the full pid+tgid? Or does it matter. Wonder if you have multiple threads all doing execs does this get confused?

jrfastab:
Other than comment about pid for map key. LGTM.

bpf/process/bpf_execve_bprm_check.c
tixxdz: Good catch! sorry missed that somehow... So let's make it simple I made it full tid, and commented the fact that execve is complex that can fail at any point and we don't care about kernel locking or thread races between... all may execute our hooks but only the leader will send us the final execve tracepoint which we care for this case.

Leader will get cleaned properly, other threads entries will be just re-used...

tixxdz:
> Other than comment about pid for map key. LGTM.

Ok updated fixed all, but also removed that empty `"info": {}` , now entries that don't have:  "inode.initialized == 1 && inode.nlinks == 0" will not have any displayed info field.  

```
{
  "process_exec": {
    "process": {
      "exec_id": "OjYwMjQ1NTcxNTg2MDU6MTA5MzY2",
      "pid": 109366,
      "uid": 1000,
      "binary": "/proc/self/fd/3",
      "flags": "execve",
      "start_time": "2022-10-22T22:22:13.554073460Z",
      "auid": 1000,
      "parent_exec_id": "OjYwMjQ1NTYwNjA2ODI6MTA5MzY2",
      "info": {
        "inode": {
          "deleted": true
        }
      }
    },
    "parent": {
      "exec_id": "OjYwMjQ1NTYwNjA2ODI6MTA5MzY2",
      "pid": 109366,
      "uid": 1000,
      "binary": "/usr/bin/memfdloader",
      "arguments": "/bin/true",
      "flags": "execve clone",
      "start_time": "2022-10-22T22:22:13.552976581Z",
      "auid": 1000,
      "parent_exec_id": "OjExMjc0MDAwMDAwMDA6MTA0MTU=",
      "refcnt": 2
    }
  },
  "time": "2022-10-22T22:22:13.554073906Z"
}
```

pkg/sensors/exec/exec_test.go
```diff
@@ -556,3 +558,94 @@ func TestExecPerfring(t *testing.T) {
...
+
+	observer.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	// Execute from memory
+	if err := exec.Command(execPath, strId).Run(); err != nil {
+		t.Fatalf("command failed: %s", err)
+	}
+
+	time.Sleep(1 * time.Second)
```
kkourt, line 638: Why is the sleep needed?

In my experience needing a sleep means that there is a potential race. Having a sleep will typically not fix the race, but will mitigate it causing unexpected flakes. I'd try to avoid adding new sleep calls in testing functions unless there is a reason for them.

tixxdz: Ok yes this was when I was testing just to sync on the fs, but yeh let's remove it since sync in memory is nop

bpf/process/bpf_execve_event.c
```diff
@@ -252,6 +258,8 @@ execve_send(struct sched_execve_args *ctx)
 #endif
 	}
 
+	execve_joined_info_map_clear(tid);
```
kkourt, line 261: Do we really need this?  I don't see a code path where `execve_joined_info_map_get` is not called but this function is. I would suggest to to keep things simple and remove `execve_joined_info_map_clear`.

tixxdz: Done, thanks @kkourt

bpf/process/bpf_execve_event.c
```diff
@@ -194,6 +197,8 @@ event_execve(struct sched_execve_args *ctx)
 	compiler_barrier();
 	__event_get_task_info(event, MSG_OP_EXECVE, walker, true);
 
+	execve_joined_info_map_get(tid, &event->info, true);
```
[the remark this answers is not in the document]
tixxdz: Yes

tixxdz:
Don't merge yet, I will convert it to follow base execve v6.0 kernel versions

tixxdz:
Reviving this

bpf/process/bpf_execve_bprm_commit_creds.c
```diff
@@ -70,6 +66,16 @@ BPF_KPROBE(tg_kp_bprm_committing_creds, struct linux_binprm *bprm)
 
 	memset(&heap->info, 0, sizeof(struct execve_info));
 
+	/* Read binary file information */
+	probe_read(&file, sizeof(file), _(&bprm->file));
```
tpapagian, line 70: Why not use something like `file = BPF_CORE_READ(bprm, file);` to be consistent with the next lines?

pkg/sensors/exec/exec_test.go
```diff
@@ -1263,3 +1263,97 @@ func TestExecProcessCredentialsFileCapChanges(t *testing.T) {
...
+
+	checker := ec.NewUnorderedEventChecker(
+		ec.NewProcessExecChecker("exec").
+			WithProcess(ec.NewProcessChecker().
+				WithBinary(sm.Suffix(execPath)).
+				WithArguments(sm.Full(strId)).
+				WithBinaryProperties(ec.NewBinaryPropertiesChecker().
+					WithFile(ec.NewFilePropertiesChecker().
+						WithInode(ec.NewInodeChecker().
+							WithLinks(0),
```
tpapagian, line 1350: Inode would be != 0 here, right? Does it make sense to check that as well? As we have an fd to that, a variant of stat should do that. I haven't tested it but I believe that it should work.

pkg/sensors/exec/exec_test.go
```diff
@@ -1357,3 +1357,67 @@ func TestExecDeletedBinaryMemfd(t *testing.T) {
...
```
tpapagian, line 1414: Same here, can we check the inode number as well?

pkg/sensors/exec/exec_test.go
```diff
@@ -1357,3 +1357,67 @@ func TestExecDeletedBinaryMemfd(t *testing.T) {
...
+	}
+
+	testDir := t.TempDir()
+	// Copy /bin/true
+	truePath := "/bin/true"
+	testTruePath := fmt.Sprintf("%s/true", testDir)
+
+	if err := exec.Command("/usr/bin/cp", "-f", truePath, testTruePath).Run(); err != nil {
+		t.Fatalf("Failed to copy binary from '%s' to '%s': %s", truePath, testTruePath, err)
+	}
```
tpapagian, line 1380: Not a big deal but this could be done also in golang (i.e. using `io.Copy` function). This can make error handling easier and we should not rely on the existence of `/usr/bin/cp`.

api/v1/tetragon/tetragon.proto
```diff
@@ -136,6 +136,20 @@ message ProcessCredentials {
 	UserNamespace user_ns = 11;
 }
 
+message Inode {
+	// The inode number
+    uint64 number = 1;
```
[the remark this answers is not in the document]
tixxdz: sure

api/v1/tetragon/tetragon.proto
```diff
@@ -136,6 +136,20 @@ message ProcessCredentials {
...
+	// The inode links on the file system. If zero means the file is only in memory
+	google.protobuf.UInt32Value links = 2;
```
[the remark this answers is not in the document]
tixxdz: I think if it is an uint32 it won't be printed at all right? but with a pointer then it will be printed even if 0 unless the ptr is nil

bpf/process/bpf_execve_bprm_commit_creds.c
tixxdz: Alright, I converted it and added some checks to ensure if it did not fail, but I don't propagate errors back into maps.

I will do those in a follow PR as I have to check how we are doing this since your last update.

pkg/sensors/exec/exec_test.go
tixxdz: Indeed makes sense ;-)

pkg/sensors/exec/exec_test.go
tixxdz: Indeed you are right! I just copied old code , fixed to use CopyFile helper ;-)

pkg/sensors/exec/exec_test.go
tixxdz: Done too, thanks ;-)

tixxdz:
> Thanks! I have added some comments. Nothing very difficult to change, but good to have these as well before merging.

@tpapagian much appreciated for the review ;-) ! all comments handled , let's see if green

api/v1/tetragon/tetragon.proto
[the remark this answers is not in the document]
tpapagian: This makes sense, yes.

pkg/sensors/exec/exec_test.go
```diff
@@ -1263,3 +1263,180 @@ func TestExecProcessCredentialsFileCapChanges(t *testing.T) {
...
+		t.Fatalf("Failed to copy binary: %s", err)
+	}
+	t.Cleanup(func() {
+		// clean up on errors
+		os.Remove(testTruePath)
+	})
+
+	if err := exec.Command("/usr/bin/cp", "-f", truePath, testTruePath).Run(); err != nil {
+		t.Fatalf("Failed to copy binary from '%s' to '%s': %s", truePath, testTruePath, err)
+	}
```
tpapagian, line 1393: Do we need that? We do the copy with `testutils.CopyFile` here as well. So we have to remove that from here.

tpapagian (approved):
Other than a small comment, this LGTM. Thanks!

pkg/sensors/exec/exec_test.go
tixxdz: Agg sorry! I was sure I removed it ... wil do

tixxdz: Fixed thanks Anastasios

PR #510, opened 2022-10-26
Fix a deadlock in eventcache

tpapagian:
There is a case where we have an OOO exec event and this will also try to add again a new entry in the eventcache that will cause a deadlock.

This commit fixes that by creating a goroutine for that case.

### DIFF
--- a/pkg/grpc/exec/exec.go
+++ b/pkg/grpc/exec/exec.go
@@ -30,13 +30,13 @@ const (
 	ProcessRefCnt = 1
 )
 
-func execCleanup(msg *MsgExecveEventUnix) {
-	if msg.CleanupProcess.Ktime != 0 {
-		ev := &MsgProcessCleanupEventUnix{
-			PID:   msg.CleanupProcess.Pid,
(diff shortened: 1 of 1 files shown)


kkourt (approved):

olsajiri:
@tpapagian so do I get it right, that the channel works like pipe and it Add will block the send until the other side reads it.. which it can't because it's in the same thread?

tpapagian:
> @tpapagian so do I get it right, that the channel works like pipe and it Add will block the send until the other side reads it.. which it can't because it's in the same thread?

Yes thats true. An event can call `Add()` to insert an event into the EventCache. `eventcache.handleEvents()` traverse all EventCache entries and do the appropriate actions. Before that `eventcache.handleEvents()` can also call `Add()`. 

We use an unbuffered channel that blocks until the receiver gets the event. So we can have a deadlock. Now we simply create a new goroutine to do that for us.

olsajiri (approved):

michi-covalent (approved):

PR #514, opened 2022-10-27
tetragon: Add bpf_printk helper from libbpf

olsajiri:
Adding bpf_printk helper from bpf selftests that allows to use bpf_trace_vprintk helper and display more than 3 format arguments. This helper did not make it to 5.15 so it's available later on.

Up to 5.15 we can use bpf_printk with 3 format arguments, later versions can use up to 12 format arguments.

We need to use global data variable for format string for kernel versions < 5.2 (needs to be on stack), so I'm using __LARGE_BPF_PROG for that.


olsajiri:
the checkpatch complains about libbpf macro style, which I think we can ignore

willfindlay:
@olsajiri want me to disable that lint? Can you think of a situation where we _would_ care about it?

kkourt (approved):

PR #515, opened 2022-10-28
releases: make systemd portable service package

tixxdz:


tixxdz:
Closing this as we will provide a tarball that can be extracted and started as a systemd service

PR #521, opened 2022-11-01
tetragon: fixup binary names selectors

jrfastab:
Binary selector names were not working reliably. Couple issues with BPF side using stale data and selectors not propagating correctly when using multiple binary values.

This is a first pass to improve binarys. Another set of patches will be useful to support >4 binary values  and then also we need some help on delete path.

### DIFF
--- a/pkg/selectors/binaryentry.go
+++ b/pkg/selectors/binaryentry.go
@@ -1,12 +1,20 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of Tetragon
-package tracing
+package selectors
 
 import (
 	"fmt"
+	"path/filepath"
+	"time"
 	"unsafe"
 
 	"github.com/cilium/tetragon/pkg/bpf"
+	"github.com/cilium/tetragon/pkg/sensors/base"
+)
+
+const (
+	maxMapRetries = 4
+	mapRetryDelay = 1
 )
 
 type BinaryMapKey struct {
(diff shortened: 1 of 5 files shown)


pkg/selectors/binaryentry.go
```diff
@@ -27,7 +35,21 @@ func (v *BinaryMapValue) NewValue() bpf.MapValue         { return &BinaryMapValu
...
+func WriteBinaryMap(id uint32, path string) error {
+	mapDir := bpf.MapPrefixPath()
+	m, err := bpf.OpenMap(filepath.Join(mapDir, base.NamesMap.Name))
+	for i := 0; err != nil; i++ {
+		m, err = bpf.OpenMap(filepath.Join(mapDir, base.NamesMap.Name))
+		if err != nil {
+			time.Sleep(mapRetryDelay * time.Second)
+		}
+		if i > maxMapRetries {
+			panic(err)
```
kkourt, line 47: Let's not panic and crush the agent. We can return an error instead.

pkg/selectors/binaryentry.go
```diff
@@ -27,7 +35,21 @@ func (v *BinaryMapValue) NewValue() bpf.MapValue         { return &BinaryMapValu
 func (v *BinaryMapValue) GetValuePtr() unsafe.Pointer    { return unsafe.Pointer(v) }
 func (v *BinaryMapValue) DeepCopyMapValue() bpf.MapValue { return &BinaryMapValue{} }
 
-func writeBinaryMap(id int, path string, m *bpf.Map) error {
+func WriteBinaryMap(id uint32, path string) error {
+	mapDir := bpf.MapPrefixPath()
+	m, err := bpf.OpenMap(filepath.Join(mapDir, base.NamesMap.Name))
```
kkourt, line 40: I think using this path means that if two kprobes use the binary  names feature, one will override the map of the other, and things will break. Shouldn't we make this a per-hook map?

There is some code for doing that. Using `filter_map` as an example:

It's added here to the sensor:
https://github.com/cilium/tetragon/blob/3076ae7b9cd2bbd7320af3708af240caa0fc684a/pkg/sensors/tracing/generickprobe.go#L498-L499

Note that the pinPath is set to be hook-specific here:
https://github.com/cilium/tetragon/blob/3076ae7b9cd2bbd7320af3708af240caa0fc684a/pkg/sensors/tracing/generickprobe.go#L476
And it's loaded using `MapLoad` in:
https://github.com/cilium/tetragon/blob/3076ae7b9cd2bbd7320af3708af240caa0fc684a/pkg/sensors/tracing/selectors.go#L59-L67

So the general idea is to store the needed information in the `KernelSelectorState`, and then have the mapLoad callback to load the data when the sensor is loaded.

kkourt (changes requested):
Hey John,

Please see some comments below. Thanks!

pkg/selectors/binaryentry.go
jrfastab: for what its worth seems we have a couple other panics from similar errors.

kkourt:
@tpapagian did you already fix this? If so, @jrfastab should we close this?

tpapagian:
> @tpapagian did you already fix this in antoher PR? If so, @jrfastab should we close this?

Yes, this is fixed in https://github.com/cilium/tetragon/pull/686

PR #531, opened 2022-11-04
various fixes motivated by a failure of the raw syscall test

kkourt:
see commits

### DIFF
--- a/pkg/selectors/selectors.go
+++ b/pkg/selectors/selectors.go
@@ -35,7 +35,7 @@ func WriteSelectorUint32(k *KernelSelectorState, v uint32) {
 
 func WriteSelectorInt64(k *KernelSelectorState, v int64) {
 	binary.LittleEndian.PutUint64(k.e[k.off:], uint64(v))
-	k.off += 4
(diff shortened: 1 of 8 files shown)


pkg/selectors/selectors.go
```diff
@@ -35,7 +35,7 @@ func WriteSelectorUint32(k *KernelSelectorState, v uint32) {
 
 func WriteSelectorInt64(k *KernelSelectorState, v int64) {
 	binary.LittleEndian.PutUint64(k.e[k.off:], uint64(v))
-	k.off += 4
+	k.off += 8
```
tixxdz, line 38: Could this be covered by simple tests?

tixxdz (approved):
Thank you @kkourt

pkg/selectors/selectors.go
jrfastab: ouch :( guess we don't use int64 much.

kkourt: > Could this be covered by simple tests?

The test is in the PR :)

PR #532, opened 2022-11-04
Fix for execve events that come after clone

tpapagian:
https://github.com/cilium/tetragon/commit/ef0a4260f4be702ca31d5e1525e20c841647d6eb changes the behaviour of exec events that come after a clone event.

Before that commit we used to have:
```
process_clone with process pid 10 parent pid 9 // we cannot see that event as process_clone are not exported 
process_exec with process pid 10 parent pid 9  // will have as parent pid the clone's event parent pid 
process_exec with process pid 10 parent pid 10 // will have as parent the previous exec's pid
```

Now we have:
```
process_clone with process pid 10 parent pid 9 // we cannot see that event as process_clone are not exported 
process_exec with process pid 10 parent pid 10 // will have as parent pid the one from the clone event 
process_exec with process pid 10 parent pid 10 // will have as parent the previous exec's pid
```

This commit fixes that to behave the same as before. This is done by using the real parent for the first exec event after a clone event.


olsajiri (approved):
looks good, fixes the problem for me

PR #537, opened 2022-11-10
tetragon: Add v6.0 bpf objects and related fixes

olsajiri:
With clang 14 the prepend_path loop code does not pass verifier
on v6.x kernels, which can be fixed by using bpf_loop helper.

In order to use bpf_loop helper easily, I'm adding v6.0 version of
bpf objects, so we don't need to detect features and have specific
binary for each.

The v6.0 seems to be the latest stable, I'm not sure it will be longterm,
but with 5.19 EOL the other possibility is 5.15 which does not have
bpf_loop helper.. but I'm open to suggestions in here ;-)

Note I plan to move kprobe_multi objects into v60 in following changes.

### DIFF
--- a/bpf/process/bpf_process_event.h
+++ b/bpf/process/bpf_process_event.h
@@ -164,6 +164,7 @@ struct cwd_read_data {
 	const struct path *root;
 	char *bf;
 	struct dentry *dentry;
+	struct vfsmount *vfsmnt;
 };
 
 static inline __attribute__((always_inline)) int
@@ -174,7 +175,6 @@ prepend_path(const struct path *path, const struct path *root, char *bf,
 		.root = root,
 		.bf = bf,
 	};
-	struct vfsmount *vfsmnt;
 	struct mount *mnt;
 	struct qstr d_name;
 	int error = 0;
@@ -185,8 +185,8 @@ prepend_path(const struct path *path, const struct path *root, char *bf,
 	bptr = *buffer;
 	blen = *buflen;
 	probe_read(&data.dentry, sizeof(data.dentry), _(&path->dentry));
-	probe_read(&vfsmnt, sizeof(vfsmnt), _(&path->mnt));
-	mnt = real_mount(vfsmnt);
+	probe_read(&data.vfsmnt, sizeof(data.vfsmnt), _(&path->mnt));
+	mnt = real_mount(data.vfsmnt);
 
 #ifndef __LARGE_BPF_PROG
 #pragma unroll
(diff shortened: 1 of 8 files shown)


jrfastab (approved):

bpf/process/bpf_process_event.h
```diff
@@ -166,6 +166,7 @@ struct cwd_read_data {
 	struct dentry *dentry;
 	struct vfsmount *vfsmnt;
 	struct mount *mnt;
+	int error;
```
kkourt, line 169: nit: I believe error can be a local variable in the new function, since we seem to always check for ->resolved in the end and set the error to UNRESOLVED_PATH_COMPONENTS.

bpf/process/bpf_process_event.h
```diff
@@ -172,6 +172,69 @@ struct cwd_read_data {
...
+	struct dentry *vfsmnt_mnt_root;
+	struct vfsmount *root_mnt;
+	struct dentry *root_dentry;
+	const struct path *root = data->root;
+	struct dentry *dentry = data->dentry;
+	struct vfsmount *vfsmnt = data->vfsmnt;
+	struct mount *mnt = data->mnt;
+
+	probe_read(&root_dentry, sizeof(root_dentry), _(&root->dentry));
+	probe_read(&root_mnt, sizeof(root_mnt), _(&root->mnt));
```
kkourt, line 189: Do root->dentry and root->mnt change? Do we need to read them on every iteration or can we just keep them at `cwd_read_data` and read them once? 

(I understand that this patch just moves the existing code around, but maybe we can improve it while we are at it?)

kkourt (commented):
Thanks!

Just some small comments from my side.

It would be nice to use bpf_d_path here, but AFAICT we [cannot](https://elixir.bootlin.com/linux/v6.0.8/source/kernel/trace/bpf_trace.c#L931):

```

static bool bpf_d_path_allowed(const struct bpf_prog *prog)
{
	if (prog->type == BPF_PROG_TYPE_TRACING &&
	    prog->expected_attach_type == BPF_TRACE_ITER)
		return true;

	if (prog->type == BPF_PROG_TYPE_LSM)
		return bpf_lsm_is_sleepable_hook(prog->aux->attach_btf_id);

	return btf_id_set_contains(&btf_allowlist_d_path,
				   prog->aux->attach_btf_id);
}
```

bpf/process/bpf_process_event.h
olsajiri: sounds good, will check, thanks

bpf/process/bpf_process_event.h
olsajiri: true, I'll check and add commit on top of the move change

Issue #540, opened 2022-11-14
cgroups: fix the default cgroup bpf helper to get Cgroup IDs


tixxdz:
Until now we have not used the cgroup ID of the execve events, only the cgroup names that we convert to container IDs.

Using the cgroup ID will not work on cgroupv1, as the upstream bpf helper  `bpf_get_current_cgroup_id()` is only for the default cgroupv2 hierarchy. Using it on cgroupv1 may return zero or the wrong cgroup ID that could point to a root cgroup or another node that does not reflect the k8s cgroup hierarchy.

Cgroup names have already been fixed.

PR #541, opened 2022-11-14
cgroups: ensure that cgroup IDs correlate with execve events

tixxdz:
This PR contains:

- Fixes: #540
   we want to support both Cgroupv1 and Cgroupv2 transparently as we still have users that did not switch to pure Cgroupv2 unified hierarchy. The cgroup IDs were not used before (we are planning to use them in next PRs), only the cgroup names are being used and they work.

- Tests to match cgroup IDs of cgroup mkdir events with the cgroup ID and Name (container ID) of the execve events. This allows to emulate how k8s and container runtimes work:
  - Create cgroup A, migrate the child (container pid 1) to cgroup A, and then execve the entry point where all container processes should belong to cgroup A.
  - Tests cover both unified and hybrid (cgroupv1 and cgroupv2) hierarchies.


pkg/sensors/exec/cgroups_test.go
```diff
@@ -1068,3 +1097,247 @@ func TestCgroupv1K8sHierarchyInHybridBlkio(t *testing.T) {
 func TestCgroupv1K8sHierarchyInHybridInvalid(t *testing.T) {
 	testCgroupv1K8sHierarchyInHybrid(t, "invalid-cgroup-controller")
 }
+
+// This will execute a helper program that creates cgroups
+// then migrates to them and perform for each one exec in order
+// to assert collected cgroup id + name of the exec event match.
+// This works in Hybrid mode and is performed for all hierarchies.
+func testCgroupv1ExecK8sHierarchyInHybrid(t *testing.T, selectedController string) {
```
kkourt, line 1149: It seems like a lot of the code here has been copy-pasted from `testCgroupv1K8sHierarchyInHybrid`.
Please add helper functions to i) avoid code duplication and ii) make it clearer what the code actual does.

kkourt (changes requested):
Thanks! Overall approach LGTM!

bpf/process/bpf_process_event.h
```diff
@@ -629,6 +629,9 @@ __event_get_cgroup_info(struct msg_execve_event *msg,
 	int zero = 0, subsys_idx = 0;
 	struct tetragon_conf *conf;
 
+	/* Ensure that we clear up cgroup info at the beginning*/
+	memset(&msg->kube, 0, sizeof(struct msg_k8s));
```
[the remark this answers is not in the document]
tixxdz: As commented in the code, it needs to be in the beginning so if we return with an error all msg->kube fields are already cleared, otherwise we may return old unrelated cgroup info of another process..., msg is in the heap.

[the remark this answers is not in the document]
kkourt: Ah that makes sense. How about we move it at the top level function, in event_execve? This way it would be obvious that memset will be called before we return?

Could you also add some additional information to the comment? For example:
> /*  Clear cgroup info at the beginning, so if we return early we do not pass previous data */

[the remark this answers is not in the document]
tixxdz: > Ah that makes sense. How about we move it at the top level function, in event_execve? This way it would be obvious that memset will be called before we return?

hmm

The event_execve() should only return early if it can't allocate entry from maps, otherwise it _should always_ submit a perf event, even if it is not complete while at same time marking the errors in the ->flags for missing related info. What I mean is if we can't read namespaces, then we mark ->flags and we must continue to read the other related data.

Here the return is just in the cgroup context msg->kube , if we fail to read cgroup, we return making sure that it is cleared, it does not mean return and stop collecting data.

Moving this in the beginning will take it out of from its context, and even if we memset(event) the full object: is too big and it will fail to compile... having the memset close to its usage is best choice.

Sure I will add that comment to improve it, ok!

bpf/process/bpf_process_event.h
```diff
@@ -645,7 +633,12 @@ __event_get_cgroup_info(struct msg_execve_event *msg,
 	}
 
 	/* Collect event cgroup ID */
-	__event_get_current_cgroup_id(msg, cgrp);
+	msg->kube.cgrpid =
+		tg_get_current_cgroup_id(task, cgrpfs_magic, subsys_idx);
+	if (!msg->kube.cgrpid) {
+		process->flags |= EVENT_DOCKER_SUBSYSCGRP_ERR;
```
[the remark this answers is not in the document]
tixxdz: Ok will do

pkg/sensors/exec/cgroups_test.go
tixxdz: We can but for the time being it is not easy, those tests are different

- `testCgroupv1K8sHierarchyInHybrid` does test only mkdir / rmdir events probing other cgroup controllers to ensure that we catch only events that belong to the right controller/hierarchy we are interested in.

- `testCgroupv1ExecK8sHierarchyInHybrid` does the mkdir / execve / rmdir but only on the pids / memory controllers without probing other controllers as we should not see their events (this is already handled by the above test). This test handles only this case of the execve, but also there is a bug in our CI for old 4.9 kernels where it fails when execve migrating on the cpuset controller, where local tests on same kernel succeeded... don't know what's the cause... 

I think the code that could be shared is already shared, but I will try to see if I can share more, I think I will add more comments to explain why and trying to keep things clear even if we have some lines duplicated, changing one test won't affect the other.

contrib/tester-progs/cgroup-migrate.bash
```diff
@@ -0,0 +1,194 @@
+#!/bin/bash
+
+# SPDX-License-Identifier: Apache-2.0
+# Copyright Authors of Tetragon
+
+# This is part of testing and it assumes callers
```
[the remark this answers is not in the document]
tixxdz: Sure will do , thank you!

pkg/sensors/exec/cgroups_test.go
kkourt: Given that this is the diff between the two functions:

```diff

-func testCgroupv1K8sHierarchyInHybrid(t *testing.T, selectedController string) {
+func testCgroupv1ExecK8sHierarchyInHybrid(t *testing.T, selectedController string) {
 	testutils.CaptureLog(t, logger.GetLogger().(*logrus.Logger))
 	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
 	defer cancel()
@@ -69,9 +68,9 @@
 		}
 
 		/* If not found falls back to default selected controller by
-		 * the cgroups auto discovery.
+		 * the cgroups auto discovery. For now we only do memory and pids controllers
 		 */
-		if !found {
+		if !found || (usedController != "memory" && usedController != "pids") {
 			usedController = cgroups.GetCgrpControllerName()
 		}
 	}
@@ -82,6 +81,8 @@
 	controllers := append([]cgroupController(nil), cgroupv1Controllers...)
 	kubeCgroupHierarchiesMap := make(map[string][]cgroupHierarchy, len(cgroupv1Controllers))
 
+	// Register cleanups first
+
 	// Ensure we unmount all cgroup mount points
 	cleanupMounts := func() {
 		for _, controller := range controllers {
@@ -129,8 +130,12 @@
 				} else {
 					n.path = c.path
 				}
-				if controller.name == usedController && uint32(i) < trackingCgrpLevel {
-					n.tracking = true
+				if controller.name == usedController {
+					if uint32(i) < trackingCgrpLevel {
+						n.tracking = true
+					}
+					// We are tracking this controller expect docker fields
+					n.expectedDocker = c.expectedDocker
 				} else {
 					n.tracking = false // Controller was not used at all, hierarchy not to be tracked
 				}
@@ -147,45 +152,72 @@
 	err = logCgroupMountInfo(t)
 	assert.NoError(t, err)
 
-	triggerCgroupMkdir := func() {
-		for hierarchy, cgroupHierarchy := range kubeCgroupHierarchiesMap {
-			path := ""
-			for _, dir := range cgroupHierarchy {
-				path = filepath.Join(path, dir.path)
-				err = cgroupMkdir(t, cgroupRoot, hierarchy, path)
-				if err != nil {
-					t.Fatalf("Failed to create cgroup %s/%s/%s", cgroupRoot, hierarchy, dir.path)
-				}
+	testCgroupMigrate := testutils.ContribPath("tester-progs/cgroup-migrate.bash")
+	triggerCgroupExec := func() {
+		path := filepath.Join(cgroupRoot, usedController)
+		for i, dir := range kubeCgroupHierarchiesMap[usedController] {
+			path = filepath.Join(path, dir.path)
+			if dir.tracking == true {
+				kubeCgroupHierarchiesMap[usedController][i].matchExecCgrpID = true
+			}
+			/* We expect docker field here */
+			if dir.expectedDocker == "yes" {
+				docker, _ := procevents.LookupContainerId(dir.path, true, false)
+				require.NotEmpty(t, docker)
+				kubeCgroupHierarchiesMap[usedController][i].expectedDocker = docker
+			}
+			var outb, errb bytes.Buffer
+			cmd := exec.Command(testCgroupMigrate, "--mode", "cgroupv1", "--controller", usedController,
+				"--new", dir.path, path)
+			cmd.Stdout = &outb
+			cmd.Stderr = &errb
+			t.Logf("Executing command: %v", cmd.String())
+			err := cmd.Run()
+			stderr := errb.String()
+			if len(stderr) > 0 {
+				t.Logf("\nstderr:\n%v", stderr)
+			}
+			if err != nil {
+				t.Fatalf("Command failed: %s", err)
+			}
+			if len(outb.String()) > 0 {
+				t.Logf("\nstdout:\n%v\n", outb.String())
 			}
 		}
 	}
 
 	triggerCgroupRmdir := func() {
-		for hierarchy, cgroupHierarchy := range kubeCgroupHierarchiesMap {
-			path := "/"
-			for _, dir := range cgroupHierarchy {
-				path = filepath.Join(path, dir.path)
-			}
-
-			for i := 0; i < len(cgroupHierarchy); i++ {
-				err := cgroupRmdir(t, cgroupRoot, hierarchy, path)
-				if err != nil {
-					t.Fatalf("Failed to remove cgroup %s/%s/%s", cgroupRoot, hierarchy, path)
-				}
-				path = filepath.Dir(path)
+		path := "/"
+		for _, dir := range kubeCgroupHierarchiesMap[usedController] {
+			path = filepath.Join(path, dir.path)
+		}
+		for i := 0; i < len(kubeCgroupHierarchiesMap[usedController]); i++ {
+			err := cgroupRmdir(t, cgroupRoot, usedController, path)
+			if err != nil {
+				t.Fatalf("Failed to remove cgroup %s/%s/%s", cgroupRoot, usedController, path)
 			}
+			path = filepath.Dir(path)
 		}
 	}
 
 	triggers := []func(){
-		triggerCgroupMkdir, triggerCgroupRmdir,
+		triggerCgroupExec, triggerCgroupRmdir,
 	}
 
 	// Loop over all created cgroup hierarchies
 	testCgroupv1HierarchyInHybrid(ctx, t, cgroupRoot, usedController, kubeCgroupHierarchiesMap,
 		trackingCgrpLevel, triggers)
 
+	// dump final constructed values
+	for controller, cgroupHierarchy := range kubeCgroupHierarchiesMap {
+		t.Logf("\nkubeCgroupHierarchiesMap[%s]=%+v\n", controller, cgroupHierarchy)
+	}
+
 	for _, cgroupHierarchy := range kubeCgroupHierarchiesMap {
 		assertCgroupDirTracking(t, cgroupHierarchy)
 	}
+
+	for _, cgroupHierarchy := range kubeCgroupHierarchiesMap {
+		assertCgroupIDsTracking(t, cgroupHierarchy)
+	}
 }
```

it seems to me that the main difference is the triggers, and those can be passed as arguments.

> I think I will add more comments to explain why and trying to keep things clear even if we have some lines duplicated, changing one test won't affect the other.

I think that also makes sense. The code does not seem easy to follow. I'd rather have simpler tests, but if we cannot we should at least make them so as understandable as possible both for other people but also for our future selves.

pkg/sensors/exec/cgroups_test.go
```diff
@@ -1068,3 +1141,415 @@ func TestCgroupv1K8sHierarchyInHybridBlkio(t *testing.T) {
...
+	testutils.CaptureLog(t, logger.GetLogger().(*logrus.Logger))
+	ctx, cancel := context.WithTimeout(context.Background(), defaultTimeout)
+	defer cancel()
+
+	option.Config.HubbleLib = tus.Conf().TetragonLib
+	option.Config.Verbosity = 5
+
+	_, err := observer.GetDefaultObserver(t, ctx, tus.Conf().TetragonLib)
```
kkourt, line 1157: Why is the observer used here? We shouldn't need it since we use the perf ring for testing.

kkourt (commented):
Some additional comments after a second look.

Also note that this:
```
func init() {
	tus.RegisterSensorsAtInit(loadedSensors)
}
```

should not be used and is going away. See: https://github.com/cilium/tetragon/pull/581/commits/8e8412f5175e810c3c67fd85ad4ed32b29dff8d3

pkg/sensors/exec/cgroups_test.go
tixxdz: > it seems to me that the main difference is the triggers, and those can be passed as arguments.

You are missing the controllers..., please take a look again, this diff shows exactly why they differ, without exec is trying events on all controller but ensures that we only work on one controller or hierarchy mkdir/rmdir events and again without exec context at all. Where the new one with exec just works on the desired controller we don't use other controllers as there is no point to test them...

That test is written explicitly like this to show that the `testCgroupv1ExecK8sHierarchyInHybrid` has the exec context hence in the name of test and we test it only desired controllers/hierarchies due the nature of the execve event and how things work in production... that's why the test are written like this... mixing them or sharing code will just add confusions

pkg/sensors/exec/cgroups_test.go
tixxdz: hmmm! we are loading tg_conf_map , doing execve checks here execve events...

bpf/process/bpf_process_event.h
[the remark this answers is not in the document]
tixxdz: Done, thanks!

pkg/sensors/exec/cgroups_test.go
tixxdz: ok got it, will replace with tus.LoadSensor(ctx, t, base.GetInitialSensor()), thank you ;-)

pkg/sensors/exec/cgroups_test.go
tixxdz: Ok, added comments to tests to make improve readability.

tixxdz:
> Thanks! Overall approach LGTM!

Thank you for the review! updated.

pkg/sensors/exec/cgroups_test.go
kkourt: It's still seems to me that there is a lot of code to be shared. I'd suggest you either:
 * have a helper function add pass appropriate functions for the trigger and the setup.
 * setup the environment and then call T.Run (https://pkg.go.dev/testing#T.Run)

bpf/process/bpf_process_event.h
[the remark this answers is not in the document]
kkourt: Not sure I understand.

You seem to be zeroing msg->kube because it has invalid data (from a previous run, I presume). Is that correct?
My argument is that we should do this at the top-level function, so that it will always hold even if we add other code that returns before the memset is executed.

That is, if we do add code like, then I think we would still have the same issue. 
```
     if (some_new_condition_here)
            goto err;
     ...
    __event_get_task_info(...)
     ...
     
     err:
       tail_call(ctx, &execve_calls, 0);
```

Am I missing something?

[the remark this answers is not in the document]
tixxdz: But we should not add code like this, doesn't really make sense to me. In this case we should not do tail_call nor send perf event at all as data in heap is bogus anyway...; we should just return 0 instead of that goto err.

There are no many conditions that we should return , unless -ENOMEM , otherwise for all other errors continue gather all data: memset or initialize values , and on errors set ->flag error

pkg/sensors/exec/cgroups_test.go
tixxdz: Ok updated both cgroupv1 and cgroupv2 with and without Exec context into their own helper functions.

bpf/process/bpf_process_event.h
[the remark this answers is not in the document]
tixxdz: I add more code comments to note the appropriate behavior and avoid possible mistakes in the future. PTAL

bpf/process/bpf_execve_event.c
```diff
@@ -175,6 +175,16 @@ event_execve(struct sched_execve_args *ctx)
...
 	if (!event)
 		return 0;
+
+	/*
+	 * At this point on errors proceed with care:
+	 *  1. Ensure that event fields are zeroed or initialized
+	 *  2. On errors update event->process.flags to reflect those
+	 *     errors and proceed collecting _all_ execve event data.
+	 *
+	 * This prevents tail calling execve_send() on an invalid event.
```
kkourt, line 185: How about we make this more specific?

"
The event we allocate above includes uninitiated data, which are zeroed out in __event_get_cgroup_info called by __event_get_task_info. If code changes so that we return before __event_get_task_info is called, ensure that the necessary fields are zeroed out and that event->process.flags is set properly.
"

contrib/tester-progs/cgroup-migrate.bash
```diff
@@ -0,0 +1,236 @@
...
+#   cgroup-migrate.bash -m cgroup2 -c memory -n new_cgroup_3 /sys/fs/cgroup/new_cgroup_1/new_cgroup_2/new_cgroup_3/
+#
+#  Or
+#
+#   cgroup-migrate.bash -m cgroup1 -c memory -n new_cgroup_1 /sys/fs/cgroup/memory/new_cgroup_1
+#   cgroup-migrate.bash -m cgroup1 -c memory -n new_cgroup_2 /sys/fs/cgroup/memory/new_cgroup_1/new_cgroup_2/
+#   cgroup-migrate.bash -m cgroup1 -c memory -n new_cgroup_3 /sys/fs/cgroup/memory/new_cgroup_1/new_cgroup_2/new_cgroup_3/
+#
+# At the end of tests, the '/sys/fs/cgroup/new_cgroup_1/...' cgroups can be
+#   safely removed.
```
kkourt, line 35: 🚀

kkourt (approved):
Thanks!

Just a small comment for your consideration.

bpf/process/bpf_execve_event.c
tixxdz: Done, thanks!

PR #542, opened 2022-11-14
tetragon: Add ReleasedPinnedBPF option to remove any old progs/maps

jrfastab:
This adds a ReleasePinnedBPF option that will remove any and all pinned programs and maps in the tetragon directory. In normal use cases we should not require this Tetragon should clean up and reinit its state correctly.

However, some cases we hit recently require a bigger hammer. Specifically, when upgrading between versions and maps or programs are no longer in the new version. We would have to keep code to tear them down on the upgrade. We do this in some cases where we expect users will commonly upgrade through these paths. But, then there are always questions, such as how many versions do we need to maintain this code? For ever? What about downgrades with new maps that no longer exist? And so on.

To handle these cases the ReleasePinnedBPF option gives users that have no need to keep maps across restart a way to do hard resetes of the BPF state of Tetragon. Further, gives us an escape hatch if needed.

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -190,6 +190,11 @@ func tetragonExecute() error {
 	option.Config.BpfDir = observerDir
 	option.Config.MapDir = observerDir
 
+	// Check if option to remove old BPF and maps is enabled.
+	if option.Config.ReleasePinned {
+		os.RemoveAll(observerDir)
+	}
+
 	// Get observer from configFile
 	obs := observer.NewObserver(configFile)
 	defer func() {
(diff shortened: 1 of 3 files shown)


michi-covalent (approved):

michi-covalent:
just as a future reference, we discussed how cilium handles this in a slack thread here: https://isovalent.slack.com/archives/C01N6G0CHFV/p1668463996755629?thread_ts=1668459661.414389&cid=C01N6G0CHFV

olsajiri (approved):

cmd/tetragon/main.go
```diff
@@ -190,6 +190,11 @@ func tetragonExecute() error {
 	option.Config.BpfDir = observerDir
 	option.Config.MapDir = observerDir
 
+	// Check if option to remove old BPF and maps is enabled.
+	if option.Config.ReleasePinned {
+		os.RemoveAll(observerDir)
```
kkourt, line 195: Do we need an option? Is there a reason to not just always do this?

jrfastab: At some point enforced maps and progs need to stay loaded across pod restart.

PR #543, opened 2022-11-15
tetragon: Check and remove not compatible map pin paths on loading

olsajiri:
Following cilium/ebpf interface creates map from pin path
and checks if the pin is compatible with the map spec:

  ebpf.NewMapWithOptions(spec, ebpf.MapOptions{PinPath: path, FullPath: true})

use that in the sensors map loading code

TODO split the cilium/ebpf change and push it separately 

### DIFF
--- a/pkg/sensors/load.go
+++ b/pkg/sensors/load.go
@@ -181,6 +181,17 @@ func (s *Sensor) LoadMaps(stopCtx context.Context, mapDir string) error {
 			}
 		}
 
+		spec, err := ebpf.LoadCollectionSpec(m.Prog.Name)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


jrfastab:
yep needs to pass testing but certainly good to have.

olsajiri:
> yep needs to pass testing but certainly good to have.

sure, it needs the cilium/ebpf change first

olsajiri:
> > yep needs to pass testing but certainly good to have.
> 
> sure, it needs the cilium/ebpf change first [cilium/ebpf#854](https://github.com/cilium/ebpf/pull/854)

got merged, meanwhile I added tetragon local version of that

kkourt (approved):

PR #552, opened 2022-11-22
support for using unix socket for gRPC

kkourt:
This PR adds support for using a unix socket for the gRPC agent interface.

### DIFF
--- a/cmd/tetragon/addr.go
+++ b/cmd/tetragon/addr.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+)
+
+// splitListenAddr splits the user-provided address a to a proto and an address field to be used
+// with net.Listen.
+//
+// addresses can be:
+//  unix://absolute_path for unix sockets
+//  <host>:<port> for TCP (more specifically, an address that can be passed to net.Listen)
+//
+// Note that the client (tetra) uses https://github.com/grpc/grpc-go/blob/v1.51.0/clientconn.go#L135
+// With the syntax is documented in https://github.com/grpc/grpc/blob/master/doc/naming.md. The
+// server uses net.Listen. And so the two are not compatible because the client expects "ipv4" or
+// "ipv6" for tcp connections.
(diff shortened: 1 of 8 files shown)


tixxdz:
Thank you @kkourt !

Could you please add an explicit chmod fix to the socket 0700 after it is created, so we don't rely on any lib...

kkourt:
> Thank you @kkourt !
> 
> Could you please add an explicit chmod fix to the socket 0700 after it is created, so we don't rely on any lib...

I don't think it's a good idea to assume what permissions the users want. I was thinking users can just define a proper `umask` for what they want.

tixxdz:
> > Thank you @kkourt !
> > Could you please add an explicit chmod fix to the socket 0700 after it is created, so we don't rely on any lib...
> 
> I don't think it's a good idea to assume what permissions users want. I was thinking users can just define a proper `umask` for what they want.

We should be safe by default, so doing it the other way is better: have proper permission fix after it is created, then document users if they want to expose it : `chmod go+rwx` something like that

Also we should turn it on by default for standalone mode, as it seems you have values in helm charts, is the tetra look into current json value find unix socket and then use it, added here?

kkourt:
> > > Thank you @kkourt !
> > > Could you please add an explicit chmod fix to the socket 0700 after it is created, so we don't rely on any lib...
> > 
> > 
> > I don't think it's a good idea to assume what permissions users want. I was thinking users can just define a proper `umask` for what they want.
> 
> We should be safe by default, so doing it the other way is better: have proper permission fix after it is created, then document users if they want to expose it : `chmod go+rwx` something like that

I don't think we should expect users to `chmod` things manually. I've added a patch that makes it so that the socket file is created with `0660`. Long term, I think we need something similar to what the Cilium agent does: https://github.com/cilium/cilium/blob/54e70e2933102bdadfadb27dd352f35d0b35622f/api/v1/server/server.go#L189.

PTAL.

tixxdz (approved):
Thank you! unix socket files should behave like normal files..  maybe a test for the ListenWithRename to stat ensure that socket is created with the right mode... up to you!

Issue #553, opened 2022-11-23
Add destination pod context


anjmao:
Currently tetragon only captures source pod info. Adding destination pod info for networking related events could be really useful.

I split this into 3 phases.
* **Phase1** Find destination pod by destination address. This is quite simple to implement. We can find destination pod from k8s pod status. Will need to extend informers to sync all pods by ip address. Also, we need to reduce memory usage by applying transform func.
* **Phase2** Find destination pod by destination address when calling via ClusterIP service. This will most likely require to get conntrack entries to find actual pod destination IP. Then use Phase1 technique.
* **Phase3** Find destination pod by destination address when calling pods on host network. This could require cooperation from other tetragon instances since we need to know listening socket port and process on the host.

PR #554, opened 2022-11-23
tetragon: Add destination pod info

anjmao:
This is initial work for https://github.com/cilium/tetragon/issues/553

Example ouput (destination pod is mapped)

```
🔌 connect test/nginx /usr/bin/curl tcp 10.244.0.8:23235 -> test/nginx2 10.244.0.9:80
📤 sendmsg test/nginx /usr/bin/curl tcp 10.244.0.8:23235 -> test/nginx2 10.244.0.9:80 bytes 74
🧹 close   test/nginx /usr/bin/curl tcp 10.244.0.8:23235 -> test/nginx2 10.244.0.9:80
💥 exit    test/nginx /usr/bin/curl http://10.244.0.9 0
🚀 process test/nginx /usr/bin/curl http://10.244.0.10
🔌 connect test/nginx /usr/bin/curl tcp 10.244.0.8:18148 -> 10.244.0.10:80
💥 exit    test/nginx /usr/bin/curl http://10.244.0.10 SIGINT
🚀 process test/nginx /usr/bin/curl http://10.244.0.9/ab
🔌 connect test/nginx /usr/bin/curl tcp 10.244.0.8:23747 -> test/nginx2 10.244.0.9:80
📤 sendmsg test/nginx /usr/bin/curl tcp 10.244.0.8:23747 -> test/nginx2 10.244.0.9:80 bytes 76
🧹 close   test/nginx /usr/bin/curl tcp 10.244.0.8:23747 -> test/nginx2 10.244.0.9:80
💥 exit    test/nginx /usr/bin/curl http://10.244.0.9/ab 0
```


pkg/watcher/watcher.go
```diff
@@ -120,8 +133,29 @@ func NewK8sWatcher(k8sClient *kubernetes.Clientset, stateSyncIntervalSec time.Du
 	return &K8sWatcher{podInformer: podInformer}
 }
 
+// podTransformFunc removes unused fields from pod to reduce memory usage.
+func podTransformFunc(obj interface{}) (interface{}, error) {
+	switch t := obj.(type) {
+	case *corev1.Pod:
+		t.SetManagedFields(nil)
```
anjmao, line 140: Here we can remove everything what is not needed for proto tetragon.Pod

pkg/watcher/watcher.go
```diff
@@ -24,22 +23,27 @@ import (
...
 	containerIDLen = 15
 	containerIdx   = "containers-ids"
+	podAddrIdx     = "pods-addrs"
 )
 
 var (
 	errNoPod = errors.New("object is not a *corev1.Pod")
 )
 
+type PodFilter struct {
```
anjmao, line 38: I wanted to keep single method to find pod, but now this struct is not clear. Would it be fine if I add comment that you can search only by containerID or addr but not both?

pkg/watcher/watcher.go
```diff
@@ -97,17 +101,26 @@ func containerIndexFunc(obj interface{}) ([]string, error) {
...
+	switch t := obj.(type) {
+	case *corev1.Pod:
+		return []string{t.Status.PodIP}, nil
+	}
+	return nil, fmt.Errorf("%w - found %T", errNoPod, obj)
+}
+
 // NewK8sWatcher returns a pointer to an initialized K8sWatcher struct.
 func NewK8sWatcher(k8sClient *kubernetes.Clientset, stateSyncIntervalSec time.Duration) *K8sWatcher {
-	k8sInformerFactory := informers.NewSharedInformerFactoryWithOptions(k8sClient, stateSyncIntervalSec,
```
anjmao, line 102: I think it will make sense to enable this feature with feature flags (disabled by default). Also, later I would like to extend pod info with nodes zones to track multi-zone traffic. Such mapping can be also enabled by flag as it will require nodes indexer.

kkourt (changes requested):
Thanks for the PR (and the issue)!

I think the general approach makes sense.
I have some first comments that you can find below. 

I also have one request for the next iteration of this PR: could you please split the PR into multiple patches so that it is easier to review?

Not sure exactly what the best split would be, but here's a first suggestion:
* changes to watcher interface
* changes to .proto files
* changes to files for adding the relevant fields
* tests

The general idea is to ease reviewing by splitting the functionality introduced into separate changes that can stand on their own and be reviewed individually.

anjmao:
Thanks for comments. Will split into smaller PRs as suggested.

pkg/watcher/watcher.go
anjmao: I agree to have separate method.

pkg/watcher/watcher.go
anjmao: Ah, yes, I see there is `ciliumState.GetIPCache().GetIPIdentity()` which returns pod and namespace. If option.Config.EnableCilium is enabled we could use this one instead of adding new indexer.

pkg/watcher/watcher.go
kkourt: Sounds good! I merged #555 which will probably create some conflicts for you. Hopefully nothing too hard, but let me know if I can help with rebasing. Thanks!

kkourt:
@anjmao Marking this PR as draft. Feel free to mark it as ready when you feel people should review it. Thanks!

PR #558, opened 2022-11-25
tetragon: Switch exit tracepoint to __put_task_struct kprobe

olsajiri:
Switching exit tracepoint to __put_task_struct kprobe, so we are sure not other thread is running.

The reason for this is that you can have application that spwns a thread which continues
running while the group leader thread exits. Tetragon sends EXIT event for the leader
but the process continues.

I initially checked the task->thread_group list, but I couldn't figure out way how to solve
the race between 2 finishing processes. There are no spin lock support for tracing programs yet.

You can check that change with trivial check in here:
  https://github.com/olsajiri/tetragon/tree/exit_racy

I think it'd work most of the time, because hitting that race seems hard,
but in any case having the kprobe on __put_task_struct seems to serve
just as good without any race (that I know of ;-) ).

### DIFF
--- a/bpf/process/bpf_exit.c
+++ b/bpf/process/bpf_exit.c
@@ -1,17 +1,28 @@
 // SPDX-License-Identifier: GPL-2.0
 /* Copyright Authors of Cilium */
 
+#include "vmlinux.h"
 #include "bpf_exit.h"
+#include "bpf_tracing.h"
 
 char _license[] __attribute__((section("license"), used)) = "GPL";
 
-__attribute__((section("tracepoint/sys_exit"), used)) int
-event_exit(struct sched_execve_args *ctx)
+__attribute__((section("kprobe/__put_task_struct"), used)) int
+event_exit(struct pt_regs *ctx)
 {
-	__u64 pid_tgid;
+	struct task_struct *task =
+		(struct task_struct *)PT_REGS_PARM1_CORE(ctx);
+	__u32 pid, tgid;
 
-	pid_tgid = get_current_pid_tgid();
+	pid = BPF_CORE_READ(task, pid);
+	tgid = BPF_CORE_READ(task, tgid);
 
-	event_exit_send(ctx, pid_tgid);
+	/* We are only tracking group leaders so if tgid is not
+	 * the same as the pid then this is an untracked child
+	 * and we can skip the lookup/insert/delete cycle that
+	 * would otherwise occur.
+	 */
+	if (pid == tgid)
+		event_exit_send(ctx, tgid);
 	return 0;
 }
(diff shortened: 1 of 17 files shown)


bpf/process/bpf_exit.c
```diff
@@ -1,17 +1,28 @@
...
+	pid = BPF_CORE_READ(task, pid);
+	tgid = BPF_CORE_READ(task, tgid);
 
-	event_exit_send(ctx, pid_tgid);
+	/* We are only tracking group leaders so if tgid is not
+	 * the same as the pid then this is an untracked child
+	 * and we can skip the lookup/insert/delete cycle that
+	 * would otherwise occur.
+	 */
+	if (pid == tgid)
```
tpapagian, line 25: Do we need this check here? Based on [this](https://elixir.bootlin.com/linux/v5.10/source/include/linux/sched/task.h#L112) it seems that it calls `__put_task_struct` only once per `task_struct`. 

Could we miss that event in the case where `pid` is not equal to `tgid` and this holds the last reference to the `task_struct`?

olsajiri: there's task_struct for each thread and `__put_task_struct ` will be called for each

we want the exit event to be send just once and group leader is released when the
last child is released.. so having the (pid == tgid) check makes sure we send the
exit event just once for the leader when every other thread is dead

tpapagian: > there's task_struct for each thread

Ok, missed that. Thanks

kkourt (commented):
Great! Thanks!

How about we add a test for this as well?
I've worked on something which seems to work locally: https://gist.github.com/kkourt/b87a19ec348252ebf47c3e485d3d35f3. Can you please add it to this PR?

olsajiri:
> Great! Thanks!
> 
> How about we add a test for this as well? I've worked on something which seems to work locally: https://gist.github.com/kkourt/b87a19ec348252ebf47c3e485d3d35f3. Can you please add it to this PR?

thanks, I added some tests as well and added yours on top of that (and moved it to test_exit.go)

kkourt:
> > Great! Thanks!
> > How about we add a test for this as well? I've worked on something which seems to work locally: https://gist.github.com/kkourt/b87a19ec348252ebf47c3e485d3d35f3. Can you please add it to this PR?
> 
> thanks, I added some tests as well and added yours on top of that (and moved it to test_exit.go)

This seems to work: https://github.com/cilium/tetragon/pull/577/commits/364ef78a137ae7462e684d8f07896f1cf5e509fc

kkourt (approved):
🚀 

Thanks!

tixxdz (approved):

PR #560, opened 2022-11-28
Fix: avoid hang when err occure

zhiyu0729:
Premature WaitGroup waiting may cause deadlock.
If a code error occurs after ProcessManager is started (for example, an error occurs in loading a configuration file),
ProcessManager cannot exit and a deadlock occurs.

Signed-off-by: Zhiyu Wang <zhiyuwang.newbis@gmail.com>

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -131,7 +131,6 @@ func stopProfile() {
 
 func tetragonExecute() error {
 	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
 
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
@@ -246,6 +245,7 @@ func tetragonExecute() error {
 
 	var cancelWg sync.WaitGroup
 	defer cancelWg.Wait()
+	defer cancel()
 
 	pm, err := tetragonGrpc.NewProcessManager(
 		ctx,


tixxdz:
Thank you @zhiyu0729 , could you please point where it got stuck and maybe add commit log?

Also seems checkpath is complaining about type

zhiyu0729:
@tixxdz added

tixxdz:
Thank you @zhiyu0729 for updating this.

I guess you are running in k8s deployment mode? and adding tracing policies dynamically? quick test with:
`sudo ./tetragon --log-level=trace --bpf-lib bpf/objs/ --config-file crds/examples/crd-with-errors.yaml`

Returned:
```
time="2022-12-02T00:44:24+01:00" level=fatal msg="Failed to start tetragon" error="error converting YAML to JSON: yaml: line 7: did not find expected key"
```

did not hang...

We recently changed this code, however I think some code path there does not properly track context or hangs between, suspicious ones:
https://github.com/cilium/tetragon/blob/main/cmd/tetragon/main.go#L273
https://github.com/cilium/tetragon/blob/main/cmd/tetragon/main.go#L282

Thoughts? I will check it tomorrow

zhiyu0729:
@tixxdz try this command
```
sudo ./tetragon --bpf-lib bpf/objs/ --config-file 123  --export-filename /tmp/tetragon.log
```

kkourt:
Thanks for the patch!

While it solves the problem you describe, I think it might have other issues. For example, if we return due to an error before we `defer cancel()`,  `cancel` will not be called. And we do pass the context in some functions (e.g., cilium.InitCiliumState) and they might wait on this context.

How about something like: https://github.com/cilium/tetragon/pull/574?

zhiyu0729:
> Thanks for the patch!
> 
> While it solves the problem you describe, I think it might have other issues. For example, if we return due to an error before we `defer cancel()`, `cancel` will not be called. And we do pass the context in some functions (e.g., cilium.InitCiliumState) and they might wait on this context.
> 
> How about something like: #574?

LGTM

PR #144, opened 2022-06-09
contrib/verify: touch verify.sh

willfindlay:
This commit adds a helper script that enables us to easily confirm whether Tetragon BPF
programs will pass the verifier. Uses bpftool under the hood.

Co-authored-by: Jussi Maki <jussi@isovalent.com>
Signed-off-by: William Findlay <will@isovalent.com>

Issue #143, opened 2022-06-09
tetragon: update libbpf usage to cilium/ebpf or latest version of libbpf


jrfastab:
We currently run a fairly old libbpf with an extra patch or two to get some functionality we needed. This issue is to update to either cilium/bpf (ideally) or at least run latest libbpf. This will clean up the build infra.

sh7dm:
I'm currently trying to implement it using cilium/ebpf replacing pkg/bpf calls. Ran into a weird issue (on kernel 5.17): 
`LoadPinnedProgram` returns `permission denied` when loading tracing program `bpf/objs/bpf_exit.o`. Of course, running with root.

sh7dm:
Might be somehow related to #164 as I'm using upstream clang-14

sh7dm:
> Might be somehow related to #164 as I'm using upstream clang-14

Irrelevant, copied objs from docker image, that didn't help

sh7dm:
I am on the way porting everything to cilium/ebpf package

kkourt:
@sh7dm cool!

Could you please sync with @olsajiri? He is also working on this.

sh7dm:
Oh, looks like this issue was created after the PR. Will check out #141 .

kkourt:
https://github.com/cilium/tetragon/pull/199

Issue #142, opened 2022-06-09
tetragon: use released clang


jrfastab:
We currently use an older clang release with BPF backports. This costs us time we have to maintain the clang branch and is now blocking building ARM releases.

[ ] upgrade tooling to use latest released clang
[ ] fixup errors on BPF code paths generated by clang
[ ] tests to ensure doesn't break us again (kernel, nightly, ?)

The reason for the branch was originally because we carried some out of tree fixes/patches. These have been merged upstream for a long time now (1+year) but I never managed to get the code to build on recent release so did a simpler task and backported the BPF backend patches/fixes I needed.

Lets fix this all up as original reason for the backport (extra alu32 patches) is no longer needed.

willfindlay:
@jrfastab Do we still want a tools-install target that installs upstream clang or are we just relying on the user to have the right version of clang installed?

willfindlay:
I've got a partial draft up that I believe covers item (2) for now. https://github.com/cilium/tetragon/pull/145

PR #573, opened 2022-12-02
tetragon: loader sensor

olsajiri:
Adding loader sensor that monitors loaded binaries/libraries and sends
process_loader event for each.

Each process_loader event contains loaded path with its build id together
with process info:

```
  {
    "process": {
      "exec_id": "OjE4ODU2ODcyMTY4NzQxOjcxMzE1",
      "pid": 71315,
      "uid": 1000,
      "cwd": "/home/jolsa/tetragon",
      "binary": "/usr/bin/jq",
      "arguments": ".process_loader",
      "flags": "execve clone",
      "start_time": "2022-12-11T22:39:29.367095937Z",
      "auid": 1000,
      "parent_exec_id": "OjM2NzkwMDAwMDAwOjExMDU=",
      "refcnt": 1
    },  
    "path": "/usr/lib64/libonig.so.5.3.0",
    "buildid": "eKQFQtnLHnJC7sULLm18xKlBZ4Y="
  }   
```


olsajiri:
I think we can ignore static check fail about BIT_MACRO

pkg/encoder/encoder.go
```diff
@@ -138,6 +139,20 @@ func (p *CompactEncoder) EventToString(response *tetragon.GetEventsResponse) (st
 			status = p.Colorer.Red.Sprint(exit.Status)
 		}
 		return CapTrailorPrinter(fmt.Sprintf("%s %s %s %s", event, processInfo, args, status), caps), nil
+	case *tetragon.GetEventsResponse_ProcessLoader:
+		loader := response.GetProcessLoader()
+		if loader.Process == nil {
+			return "", ErrMissingProcessInfo
+		}
+		event := p.Colorer.Blue.Sprintf("💥 %-7s", "loader")
```
sharlns, line 147: Can we change the emoji to 🧬 ? The 💥 stands for the "process_exit" event.

api/v1/tetragon/tetragon.proto
```diff
@@ -277,3 +277,9 @@ message HealthStatus {
 message GetHealthStatusResponse {
 	repeated HealthStatus health_status = 1;
 }
+
```
kkourt, line 280: How about adding a short comment about what this event is exactly?

pkg/sensors/tracing/loader.go
```diff
@@ -0,0 +1,154 @@
...
+	// into 'ids_map' for the bpf program to recognize our event.
+
+	var ids []uint64
+
+	for cpu := 0; cpu < nCpus; cpu++ {
+		fd, err := unix.PerfEventOpen(attr, -1, cpu, -1, unix.PERF_FLAG_FD_CLOEXEC)
+		if err != nil {
+			return fmt.Errorf("can't create perf event: %w", err)
+		}
+		loaderFds = append(loaderFds, fd)
```
kkourt, line 108: What is the purpose of `loaderFds`? AFAICT it is not used somewhere else. Was the intention to close the files at some point?

bpf/process/bpf_loader.c
```diff
@@ -0,0 +1,140 @@
...
+
+	msg = map_lookup_elem(&loader_heap, &(__u32){ 0 });
+	if (!msg)
+		return 0;
+
+	pe = (struct perf_event *)PT_REGS_PARM1_CORE(ctx);
+
+	/* Make sure it's our event that triggered perf_event_mmap_output,
+	 * to have all the needed info.
+	 */
```
kkourt, line 90: OK so to make sure I understand correctly:
 * we setup the perf monitoring from user-space
 * we get the perf event id from user-space and write it in the bpf map
 * we hook into the perf event output, and check whether the id of the event is the one we had set up
 * if it is, we get what we need from the event and write it to the event output
 
 And from what I remember from an offline discussion, we need to hook into the event because it's not easy to read the event generated in our main loop.

kkourt (commented):
I have some comments/suggestions, but overall LGTM! Thanks!

api/v1/tetragon/tetragon.proto
olsajiri: yes, will add

pkg/sensors/tracing/loader.go
olsajiri: I need fds in loaderFds to stay alive after leaving the function,
but I might be too paranoid in here, I'll double check and most likely change this,
will add more comments

bpf/process/bpf_loader.c
olsajiri: > OK so to make sure I understand correctly:
> 
>     * we setup the perf monitoring from user-space
> 
>     * we get the perf event id from user-space and write it in the bpf map
> 
>     * we hook into the perf event output, and check whether the id of the event is the one we had set up
> 
>     * if it is, we get what we need from the event and write it to the event output
> 

exactly, I'll put this description into comments ;-) thanks

> And from what I remember from an offline discussion, we need to hook into the event because it's not easy to read the event generated in our main loop.

yes, we can't rely on perf events as is, because under kubernetes the perf subsystem
converts pid/tid values into perf event owner's namespace.. so we get zero pid/tid
for all processes executed outside tetragon namespace

pkg/bpf/detect.go
```diff
@@ -79,3 +83,31 @@ func HasKprobeMulti() bool {
...
+		unix.Close(fd)
+		return true
+	}
+	return false
+}
+
+func HasBuildId() bool {
+	if buildid.initialized {
+		return buildid.detected
+	}
```
kkourt, line 108: Should we use sync.Once here to be on the safe side?

kkourt (approved):
Thanks!

(Note: checkpatch failures can be ignored)

pkg/bpf/detect.go
olsajiri: good idea.. I converted the rest of the features as well, thanks

kkourt:
Seems that there are some conflicts in the auto-generated code. @olsajiri Could you please rebase?

kkourt (approved):

kkourt:
Checkpatch errors are:

```
 CHECK:BIT_MACRO: Prefer using the BIT_ULL macro
  #142: FILE: bpf/process/bpf_loader.c:62:
  +#define ATTR_BIT_MMAP	 (1ULL << 8)
  
  CHECK:BIT_MACRO: Prefer using the BIT_ULL macro
  #143: FILE: bpf/process/bpf_loader.c:63:
  +#define ATTR_BIT_MMAP2	 (1ULL << 23)
  
  CHECK:BIT_MACRO: Prefer using the BIT_ULL macro
  #144: FILE: bpf/process/bpf_loader.c:64:
  +#define ATTR_BIT_BUILDID (1ULL << 34)
  ```
  
  which I think are safe to ignore. Merging.

PR #574, opened 2022-12-02
tetragon: fix initialization deadlock

kkourt:
There is a potential for a deadlock in the initialization case. This might happen if, for example, we pass a wrong filename for the config file.

We fix this by generating a new context which we can cancel without causing deadlocks.

The patch also adds comments to explain why this is needed so more details can be found there.

Reported-by:  Zhiyu Wang <zhiyuwang.newbis@gmail.com>


tpapagian (approved):
LGTM, thanks!

PR #579, opened 2022-12-07
vendor: bump golang-lru to v2 (requires Go >= v1.18 support for generics)

rolinh:
The new major version of [github.com/hashicorp/golang-lru](https://github.com/hashicorp/golang-lru) uses Go generics and thus Go >= v1.18 is required (this is already enforced via the `go.mod` file). Switching to generics means breaking changes in the API as, when declaring a new `lru.Cache` or invoking `lru.New`, the type of keys and values must be specified. However, this means that we can know trust the types of the key/value pairs we have in the cache and don't need to do any type assertion.


kkourt (approved):
Awesome!  Many thanks @rolinh!

PR #580, opened 2022-12-08
ci: bump golangci-lint to v1.50.1

rolinh:
Also bump the golangci-lint action to v3.3.1. Use the new `skip-cache` directive to disable caching instead of the more specific options. Remove `only-new-issues: false` given that `false` is the default value. Finally, remove `skip-go-installation` as this is an unknown config option to the linter which gives a warning.

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -18,14 +18,11 @@ jobs:
         with:
           go-version: '1.18.3'
       - name: Run static checks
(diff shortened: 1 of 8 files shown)


willfindlay (approved):
Can you run `make format` to fix the static check errors? Or if you don't mind me pushing to your branch I can do it.

willfindlay:
More struggle, needs a `make vendor` now

kkourt (approved):
Thanks!

PR #581, opened 2022-12-08
sensor cleanups

kkourt:
This PR cleanups the sensor code as a preparation for subsequent patches.

It is split in small patches to ease reviewing.

The only user-visible change is that when a sensor in a tracing policy fails to load, we will unload all the sensors of said tracing policy.

Please review commit by commit.

### DIFF
--- a/pkg/sensors/load.go
+++ b/pkg/sensors/load.go
@@ -95,7 +95,7 @@ func (s *Sensor) Load(stopCtx context.Context, bpfDir, mapDir, ciliumDir string)
 		return fmt.Errorf("tetragon, aborting could not find BPF programs: %w", err)
 	}
 
-	if err := s.LoadMaps(stopCtx, mapDir); err != nil {
+	if err := s.loadMaps(stopCtx, mapDir); err != nil {
 		return fmt.Errorf("tetragon, aborting could not load sensor BPF maps: %w", err)
 	}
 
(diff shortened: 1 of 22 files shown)


pkg/sensors/sensors.go
```diff
@@ -79,44 +79,47 @@ func SensorBuilder(name string, p []*program.Program, m []*program.Map) *Sensor
 	}
 }
 
+type specHandler interface {
```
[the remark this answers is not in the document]
kkourt: makes sense!
Added: https://github.com/cilium/tetragon/pull/581/commits/251ff3c66cda9c40c32f76af23acb5d9e4437e00

olsajiri (approved):

pkg/sensors/load.go
```diff
@@ -340,7 +340,7 @@ func createDir(bpfDir, mapDir string) {
 
 func UnloadAll(bpfDir string) {
```
kevsecurity, line 386: It looks like UnloadAll() is mostly called from test programs, but is also called from RemovePrograms in observer.go. This function (IIRC) doesn't call the unloadHook on the sensors (because it unloads programs rather than sensors). Suggest we move this to tests only (if actually required by tests) and prevent use by the real program, in order to force all unloads through sensor unloading which calls the hook. The idea is that sensors have more going on than just loaded programs, and they need stopping properly on unload.

kevsecurity (approved):

pkg/sensors/load.go
kkourt: I think we should probably remove UnloadAll(). It was added as a way to remove programs before we had the unload call, IIRC. It should only be called when the agent exits, so not running the unloadHook is probably fine, but I don't think we need it for neither the tests nor the agent shutdown anymore.

PR #582, opened 2022-12-08
tetragon: logging improvements for non k8s deployment

tixxdz:

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -346,6 +346,13 @@ func startExporter(ctx context.Context, server *server.Server) error {
 		MaxBackups: exportFileMaxBackups,
 		Compress:   exportFileCompress,
 	}
+
+	// For non k8s deployments we explicitly want log files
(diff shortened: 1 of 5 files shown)


cmd/tetragon/main.go
```diff
@@ -321,9 +321,26 @@ func tetragonExecute() error {
 		}
 	}
 
+	// Explicitly log status every 1 hour
+	go logStatus(ctx, obs, 1*time.Hour)
```
kkourt, line 325: Do we need to make this configurable? Is so, let's split the two patches and do two different PRs.

willfindlay: +1 on making this configurable

tixxdz: Ok will split it in separate patch and think about making it configurable, I'm not a fan of adding too much options better try to do the right thing by default

kkourt (approved):

michi-covalent (approved):

PR #590, opened 2022-12-12
logging: allow users to know more about the overall status

tixxdz:

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -322,7 +322,43 @@ func tetragonExecute() error {
 		}
 	}
 
+	// Periodically log status
+	go logStatus(ctx, obs)
+
 	return obs.Start(ctx)
+}
+
+// Periodically log current status every 1 hour. For lost or error
+// events we ratelimit statistics to 1 message per every 5mins to
+// continuously infor users that events are being lost without
+// polluting logs.
+func logStatus(ctx context.Context, obs *observer.Observer) {
+	prevLost := uint64(0)
+	prevErrors := uint64(0)
+	lostTicker := time.NewTicker(5 * time.Minute)
+	defer lostTicker.Stop()
+	logTicker := time.NewTicker(1 * time.Hour)
+	defer logTicker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-logTicker.C:
+			// We always print stats
+			obs.PrintStats()
+			// Update lost and errors, to not print two consecutive lines at same time
+			prevLost = obs.ReadLostEvents()
+			prevErrors = obs.ReadErrorEvents()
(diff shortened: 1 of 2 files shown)


cmd/tetragon/main.go
```diff
@@ -353,23 +354,46 @@ func startExporter(ctx context.Context, server *server.Server) error {
...
+	}
+	logFile := filepath.Base(exportFilename)
+	logsDir, err := filepath.Abs(filepath.Dir(filepath.Clean(exportFilename)))
+	if err != nil {
+		log.WithError(err).Warnf("Failed to get absolute path of exported JSON logs '%s'", exportFilename)
+		// Do not fail; we let lumberjack handle this. We want to
+		// log the rotate logs operation.
+		logsDir = filepath.Dir(exportFilename)
+	}
+	if exportFileRotationInterval > 0 {
```
kkourt, line 406: What happens if `exportFileRotationInterval == 0`? There seems to be a check initially but it was removed.

nit:
```
I'd also suggest we use, instead of two branches.

```
if (exportFileRotationInterval < 0) {
  return ...
}
...
```

File: cmd/tetragon/main.go [hunk 2]
```diff
@@ -322,9 +322,45 @@ func tetragonExecute() error {
...
+	prevLost := uint64(0)
+	prevErrors := uint64(0)
+	lostTicker := time.NewTicker(5 * time.Minute)
+	defer lostTicker.Stop()
+	logTicker := time.NewTicker(1 * time.Hour)
+	defer logTicker.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
kkourt, line 345: Do we print the stats elsewhere when we the agent exits?

kkourt (commented):
LGTM, just a couple of questions and nit. Thanks!

cmd/tetragon/main.go
tixxdz: for exportFileRotationInterval == 0 I assume that also equals default value 0 which converts to 0seconds, and we should not rotate under this condition.

Ok will do that change now.

cmd/tetragon/main.go
tixxdz: obs.PrintStats() is inside the signal handling path.

cmd/tetragon/main.go
tixxdz: Ok updated! changed the order same at the end... btw the logs should also be rotated by size --export-file-max-size-mb default is 10mb. Not related here and minor improvement I think since we can export logs to real storage, we should make it smarter taking GB annotation too...

tixxdz:
> LGTM, just a couple of questions and nit. Thanks!

Updated, thank you!

kkourt (approved):

PR #593, opened 2022-12-13
tetragon: Add buffer between perf reader and events processing code

olsajiri:
Adding extra layer between perf ring buffer reader and event
processing by sensor code.

This way we keep the perf reader fast and we won't get lost events
if sensor processing code have a hiccup and takes longer.

### DIFF
--- a/cmd/tetragon/flags.go
+++ b/cmd/tetragon/flags.go
@@ -67,6 +67,7 @@ const (
 
 	keyRBSize      = "rb-size"
 	keyRBSizeTotal = "rb-size-total"
+	keyRBQueueSize = "rb-queue-size"
 
 	keyEventQueueSize = "event-queue-size"
 
@@ -102,6 +103,7 @@ func readAndSetFlags() {
 
 	option.Config.RBSize = viper.GetInt(keyRBSize)
 	option.Config.RBSizeTotal = viper.GetInt(keyRBSizeTotal)
+	option.Config.RBQueueSize = viper.GetInt(keyRBQueueSize)
 
 	option.Config.GopsAddr = viper.GetString(keyGopsAddr)
 
--- a/pkg/observer/observer.go
+++ b/pkg/observer/observer.go
@@ -126,7 +126,6 @@ func (k *Observer) receiveEvent(data []byte) {
 	if option.Config.EnableMsgHandlingLatency {
 		timer = time.Now()
 	}
-	atomic.AddUint64(&k.recvCntr, 1)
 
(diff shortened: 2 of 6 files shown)


olsajiri:
[events buffer stats - results.pdf](https://github.com/cilium/tetragon/files/12366762/events.buffer.stats.-.results.pdf)

pkg/option/config.go
```diff
@@ -67,6 +67,8 @@ type config struct {
 	MemProfile string
 	PprofAddr  string
 
+	EventsQueueSize int
```
kkourt, line 70: This seems too similar of a name to EventQueueSize, how about something like `PerfEventsQueueSize`?

pkg/observer/observer.go
```diff
@@ -226,7 +238,12 @@ func (k *Observer) RunEvents(stopCtx context.Context, ready func()) error {
 				}
 			} else {
 				if len(record.RawSample) > 0 {
-					k.receiveEvent(record.RawSample)
+					select {
+					case eventsQueue <- &record:
+					default:
+						// eventsQueue channel is full, drop the event
```
kkourt, line 245: ~I think we need to count these lost events somehow.~
Nevermind, it's in the second patch.

cmd/tetragon/flags.go
```diff
@@ -68,6 +68,8 @@ const (
 	keyRBSize      = "rb-size"
 	keyRBSizeTotal = "rb-size-total"
 
+	keyEventsQueueSize = "events-queue-size"
```
kkourt, line 71: Similar with previous comment, how about:
keyPerfEventsQueueSize = "perf-events-queue-size"

kkourt (commented):
Cool! LGTM!

The only concern I have is about naming. How about using `Perf` in the identifiers (gave a few examples) to make it more clear?

cmd/tetragon/flags.go
olsajiri: ugh, did not notice how close those 2 options are now.. will rename

olsajiri: so I changed it to follow --rb-* options, because it's closely related to ringbuffer, so now we have:

        keyRBSize      = "rb-size"
        keyRBSizeTotal = "rb-size-total"
       keyRBQueueSize = "rb-queue-size"

also I changed the new metrics names to follow the new name

kkourt: Sounds good, thanks!

kkourt (approved):

jrfastab:
OK lets try it.

PR #599, opened 2022-12-22
helm: use a specific conf.d directory for --config-dir

tixxdz:
Use a specific directory for tetragon configmap.

Right now we are using /etc/tetragon/ as the base for configmap, however in future we want to put CRD rules under that same path /etc/tetragon/tetragon.rules.d/ .

So change the mountPath where tetragon-config points from:
  /etc/tetragon/ => /etc/tetragon/tetragon.conf.d/

### DIFF
--- a/install/kubernetes/templates/_container_tetragon.tpl
+++ b/install/kubernetes/templates/_container_tetragon.tpl
@@ -11,7 +11,7 @@
     - /usr/bin/tetragon
 {{- end }}
   args:
-    - --config-dir=/etc/tetragon
+    - --config-dir=/etc/tetragon/tetragon.conf.d/
 {{- with .Values.tetragon.argsOverride }}
   {{- toYaml . | nindent 2 }}
 {{- else }}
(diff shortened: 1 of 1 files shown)


tixxdz:
Before merging this, we have to answer if it is possible to have multiple configmaps in helm chart?

Something like:
configMap:
  - /etc/tetragon/tetragon.conf.d/   (as in this PR)
  - /etc/tetragon/x.conf.d/
  - /etc/tetragon/y.conf.d/

Where they are explicitly treated by k8s as configMaps and not just mount points. I'm not sure about this

install/kubernetes/templates/_container_tetragon.tpl
```diff
@@ -11,7 +11,7 @@
     - /usr/bin/tetragon
 {{- end }}
   args:
-    - --config-dir=/etc/tetragon
+    - --config-dir=/etc/tetragon/tetragon.conf.d/
```
kkourt, line 14: Is this a convention? /etc/tetragon/conf.d seems better to me

tixxdz: > Is this a convention? /etc/tetragon/conf.d seems better to me

Yes, but having explicit separate names could be consistent in future in case we add some specific tool with its conf that could load CRDs or policies into tetragon main agent.

kkourt (approved):

install/kubernetes/templates/_container_tetragon.tpl
kkourt: OK.

PR #603, opened 2023-01-03
Uprobes

olsajiri:


dmitris:
one question regarding the `UprobeSpec`:
```
type UProbeSpec struct {
 	// Name of the traced binary
 	Binary string `json:"binary"`
 	// Name of the traced symbol
 	Symbol string `json:"symbol"`
 }
 ```
 Currently it has the `Binary` field - but couldn't you also attach uprobes to symbols in shared libraries?  For example, on rhel8 we can do:
 ```sh
$ sudo bpftrace -l 'uprobe:/usr/lib64/libssl.so.1.1.1k:SSL_write*'
uprobe:/usr/lib64/libssl.so.1.1.1k:SSL_write
uprobe:/usr/lib64/libssl.so.1.1.1k:SSL_write_early_data
uprobe:/usr/lib64/libssl.so.1.1.1k:SSL_write_ex
```
and a uprobe attached to that library+symbol combination fires even if different binaries are using those libraries with the dynamic linker or `dlopen()`.  Maybe instead of `Binary` it should be something like `ELFFile` or `TargetFile` or `TargetObject`, which can be either a binary or a shared object?

olsajiri:
> one question regarding the `UprobeSpec`:
> 
> ```
> type UProbeSpec struct {
>  	// Name of the traced binary
>  	Binary string `json:"binary"`
>  	// Name of the traced symbol
>  	Symbol string `json:"symbol"`
>  }
> ```
> 
> Currently it has the `Binary` field - but couldn't you also attach uprobes to symbols in shared libraries? For example, on rhel8 we can do:
> 
> ```shell
> $ sudo bpftrace -l 'uprobe:/usr/lib64/libssl.so.1.1.1k:SSL_write*'
> uprobe:/usr/lib64/libssl.so.1.1.1k:SSL_write
> uprobe:/usr/lib64/libssl.so.1.1.1k:SSL_write_early_data
> uprobe:/usr/lib64/libssl.so.1.1.1k:SSL_write_ex
> ```
> 
> and a uprobe attached to that library+symbol combination fires even if different binaries are using those libraries with the dynamic linker or `dlopen()`. Maybe instead of `Binary` it should be something like `ELFFile` or `TargetFile` or `TargetObject`, which can be either a binary or a shared object?

yes, will change that.. I was thinking also just 'Path' might be ok

thanks

olsajiri:
need to figure out the uprobe config in under k8s pods

pkg/sensors/program/program.go
```diff
@@ -84,6 +84,10 @@ type Program struct {
 
 	MultiSymbols []string
 	MultiCookies []uint64
+
+	// uprobe setup
+	Path   string
+	Symbol string
```
kkourt, line 90: Can we use `LoaderData` for these fields?

We can have a:
```
type UprobeData struct {
      Path string
      Symbol string
}
```

set it in `SetUprobe` and in load time do a run-time cast to get it.

pkg/sensors/program/program.go
```diff
@@ -96,6 +100,12 @@ func (p *Program) SetLoaderData(d interface{}) *Program {
 	return p
 }
 
+func (p *Program) SetUprobe(path, symbol string) *Program {
+	p.Path = path
+	p.Symbol = symbol
+	return p
```
kkourt, line 106:
```
p.LoaderData = &UprobeData{
       Path: path,
       Symbol: symbol
}
```

pkg/sensors/program/loader.go
```diff
@@ -126,6 +126,31 @@ func KprobeAttach(load *Program) AttachFunc {
 	}
 }
 
+func UprobeAttach(load *Program) AttachFunc {
+	return func(prog *ebpf.Program, spec *ebpf.ProgramSpec) (unloader.Unloader, error) {
+		var linkFn func() (link.Link, error)
+
```
kkourt, line 132:
```
uprobeData, ok := load.LoaderData.(*UprobeData)
if !ok {
      // return error
}

File: pkg/sensors/program/loader.go [hunk 4]
```diff
@@ -126,6 +126,31 @@ func KprobeAttach(load *Program) AttachFunc {
 	}
 }
 
+func UprobeAttach(load *Program) AttachFunc {
+	return func(prog *ebpf.Program, spec *ebpf.ProgramSpec) (unloader.Unloader, error) {
+		var linkFn func() (link.Link, error)
```
kkourt, line 131: nit: how about we omit this and do `linkFn := ` below

pkg/sensors/tracing/uprobe_test.go
```diff
@@ -0,0 +1,72 @@
...
+		tus.SensorMap{Name: "tcpmon_map", Progs: []uint{6, 7, 8, 9, 10, 12}},
+
+		// shared with base sensor
+		tus.SensorMap{Name: "execve_map", Progs: []uint{6, 7, 8, 9, 10, 11, 12}},
+	}
+
+	nopHook := `
+apiVersion: cilium.io/v1alpha1
+metadata:
+  name: "uprbe"
```
kkourt, line 57: nit: typo (uprobe)

kkourt (changes requested):
Thanks!  I like the  test coverage.

The only thing that I would request to change is not adding uprobe-specific fields in Program, and instead use the LoaderData field.

pkg/sensors/program/program.go
olsajiri: right, will check on that

pkg/sensors/program/program.go
olsajiri: ok

pkg/sensors/program/loader.go
olsajiri: ok, will check

pkg/sensors/tracing/uprobe_test.go
olsajiri: ok, thanks

pkg/sensors/program/program.go
olsajiri: hum, so we already use LoaderData for passing data to LoadProbe interface
and it needs to contain generic uprobe data like `ID` to get uprobeEntry

while the UprobeAttach function is in generic program loader code which just
needs path and symbol for uprobes.. not necessarily generic in future ;-) 

I don't like mixing the genericuprobe sensor data with data for generic loader

maybe we could add another `AttachData` data holder to Program ,  like:

```
        // LoaderData represents per-type specific fields.
        LoaderData interface{}
 
+       // AttachData represents specific data for attaching probe
+       AttachData interface{}
+
        MapLoad []*MapLoad
```

and we would use it for uprobe specific attach data there and also data for multi kprobe data
which are now placed in Program's:

```
        MultiSymbols []string
        MultiCookies []uint64
```

kkourt (approved):
LGTM, thanks!

pkg/sensors/program/program.go
kkourt: Makes sense to me, thanks!

PR #604, opened 2023-01-03
tetragon: Cleanup func_id/id mess in struct msg_generic_kprobe

olsajiri:
Currently we have 'id' and 'func_id' in struct msg_generic_kprobe keeping the same value. There's no need, we can use just the 'id' value for everything.

Also use __u32 as config_map key type.


kkourt (approved):
Thanks!

tixxdz (approved):

jrfastab:
presumably had some idea in mind or refactored something to get here but anyways lgtm.

PR #606, opened 2023-01-03
tests/e2e: remove GKE auth plugin, it's deprecated

willfindlay:
The GKE auth plugin has reached the end of its lifecycle and now fails tests with a deprecation message. The new correct way to authenticate with GKE clusters is documented here: https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke

In case we want to run tests in GKE in the future, we will need to use this method instead.


willfindlay:
Marking this as draft while I do some local testing

kkourt:
There seem to be a large number of new files added by this PR. Is this expected?

willfindlay:
@kkourt yeah I'm switching to the generic auth plugin which not only seems to fix this GKE issue but adds support for other platforms too.

PR #609, opened 2023-01-05
eventcache: update PodInfoError on pod error

kkourt:
There seems to be a typo where we update the ProcessInfoError metric even when the error ErrFailedToGetPodInfo.


tpapagian (approved):
LGTM, thanks!

PR #615, opened 2023-01-10
ci: use large github runner

willfindlay:
Move away from the regular GitHub runner to a large one, which supports KVM acceleration.
This should significantly speed up test performance, so let's drop down to one group per
test and add a 5.15 test while we're at it.


kevsecurity (approved):
LGTM

PR #616, opened 2023-01-10
DONTMERGE: remove other tests

willfindlay:


willfindlay:
just bashing CI, nothing to see here

PR #618, opened 2023-01-10
Fix SBOM image signing and update image siganture verification docs

sandipanpanda:
Fix CI build PR updates image digest output which was causing the `Sign SBOM Image` step introduced in 343e9f94c6eca025307cc5219638102c57c184c2 to fail.

Update cosign verify command flags used to verify cosign claims
and signing certificates on the Tetragon images

### DIFF
--- a/.github/workflows/build-images-ci.yml
+++ b/.github/workflows/build-images-ci.yml
@@ -161,7 +161,7 @@ jobs:
         env:
           COSIGN_EXPERIMENTAL: "true"
         run: |
-          docker_build_ci_pr_digest="${{ steps.docker_build_ci_main.outputs.digest }}"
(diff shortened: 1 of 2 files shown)


willfindlay (approved):
Nice. Let's wait to see if the job passes now, then we can ship it 🚀

willfindlay:
@sandipanpanda looks like we're still hitting the same issue :(

willfindlay:
Ah you're right of course. Let me try it on an internal branch.

willfindlay:
@sandipanpanda no dice https://github.com/cilium/tetragon/actions/runs/3886939668/jobs/6632626500#step:18:21

kkourt:
After offline discussion with @aanm, it seems that the PR is expected to fail. So let's merge it!

PR #620, opened 2023-01-11
Make size of event queue configurable

kevsecurity:
The internal event queue had a fixed size of 100 events. If the event queue became full, additional events would be dropped. This commit raises the default size to 10000 events and makes the size configurable through a --event-queue-size command line flag.


tpapagian (approved):
LGTM!

PR #630, opened 2023-01-16
tetragon: Use probe task instead of current in event_exit_send

olsajiri:
We can't use get_current_task in __put_task_struct, we need to use 
__put_task_struct task argument instead, otherwise we read wrong
task's exit_code.

Fixes: 1a37adf4bd40 ("tetragon: Switch exit tracepoint to __put_task_struct kprobe")


kkourt (approved):
Nice catch!

I guess we don't have a test for the exit value then?
Added: https://github.com/cilium/tetragon/issues/634

kkourt (approved):

PR #633, opened 2023-01-16
backports/v0.8: tests/e2e: remove GKE auth plugin, it's deprecated

willfindlay:
[upstream commit: 046c80e245acad470f801fb8ed9b4d7dd55b4950]

The GKE auth plugin has reached the end of its lifecycle and now fails tests with a deprecation message. The new correct way to authenticate with GKE clusters is documented here: https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke

In case we want to run tests in GKE in the future, we will need to use this method instead.


willfindlay:
Backport of [#606](https://github.com/cilium/tetragon/pull/606)

michi-covalent (approved):

PR #635, opened 2023-01-17
tetragon: better config handling mechanism

tixxdz:
This is preparation to have tarball and package deployments.

This patch allows to have a better and stable configuration handling
mechanism, by allowing tetragon to read its configuration the
following order where each next configuration will override previous
one if it was set.

Note: drop-in configuration snippets are files where each filename
is the controlling setting and the content the corresponding value.

1. Drop-in configuration snippets inside the following directories:

   `/usr/lib/tetragon/tetragon.conf.d/*`
   `/usr/local/lib/tetragon/tetragon.conf.d/*`

2. `/etc/tetragon/tetragon.yaml` a yaml configuration.

3. Drop-in configuration snippets same as 1. that could override 1.
   and 2., under:

   `/etc/tetragon/tetragon.conf.d/*`

4. `config-dir` - drop-in configuration snippets to override all
   previous settings.

This offers the following advantages:

1. Package managers or distros must use /usr/lib/tetragon/ to set the
   vendor config.

2. Tarballs will use /usr/local/lib/tetragon/ to set its
   vendor config and override previous configs of 1. if any.

3. Configurations in `/etc/tetragon/` are strictly reserved for the
   local administrator, who may use this logic to override package
   managers or the default installed configuration 1. and 2.

4. Administrators can restore defaults settings by simply deleting the
   configuration file inside /etc/tetragon/ and all drop-ins.

5. Ability to ship sane and secure configurations by default inside
   the drop-ins without touching the /etc/ directory.

6. Ability to handle updates without touching /etc/ , reducing errors
   or overriding users administrators configurations.

7. Ability to ship simple systemd units without too much configurations.

8. Ability for administrators to override the systemd unit by
   placing it /etc/systemd/system/tetragon.service


kkourt (approved):
LGTM, thanks!

PR #636, opened 2023-01-17
.github: set the right step ID for pull request builds

aanm:
The step ID used to derive the image digest was not correctly set which made the SBOM signature to fail. This commit sets the right step ID so the SBOM is signed for PRs.

Fixes: 343e9f94c6ec ("build: Generate SBOM during image release")
Signed-off-by: André Martins <andre@cilium.io>


willfindlay:
We can close this since it's a duplicate of #618, which got merged already.

PR #643, opened 2023-01-18
tetra: Add a `GetFilter` var in getevents, add documentation and tests

mtardy:
First, I was wondering if we should follow the `-` convention for `stdin`. Like using `cat events | tetra getevents --namespace default -` like `cat pod.json | kubectl apply -f -`. Because, we are using this to detect that something was piped (and it can make tests a bit tricky):
```go
fi, _ := os.Stdin.Stat()
if fi.Mode()&os.ModeNamedPipe != 0 {
    // read events from stdin
    getEvents(context.Background(), newIOReaderClient(os.Stdin, viper.GetBool("debug")))
    return
}
```

And second, we are not using the `cmd.OutOrStdout()` or `cmd.Print()` families (I discovered their existence by writing this PR), which makes testing a bit trickier because we need to manipulate `os.Stdin` and `os.Stdout`. I'm not sure it would be a super great idea to use them but just for thought!


willfindlay (changes requested):
Thanks for this and thanks for writing tests. I noticed that your test is not being run in CI, though. Would you mind updating the Makefile `test` and `test-compile` targets to look inside `./cmd` in addition to `./pkg`. That should ensure that the test is being run.

cmd/tetra/getevents/getevents.go
```diff
@@ -21,6 +22,16 @@ import (
...
 
+// DocLong documents the commands with some examples
+const DocLong = `This command prints and filter events by connecting to the server or via
+redirection of events to the stdin. Examples:
+
+  # Connect and print events
+  %[1]s getevents --process netserver
+
+  # Redirect events and filter from stdin
+  cat events.json | %[1]s getevents -o compact --namespace default`
```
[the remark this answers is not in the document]
mtardy: Cool!!! We could add you examples you mentioned in this commit in the message https://github.com/cilium/tetragon/commit/0ec17ea1ad8255899a1e4740c1208638da8ef5dd.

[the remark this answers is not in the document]
willfindlay: @mtardy yes that would be useful indeed

mtardy:
> Thanks for this and thanks for writing tests. I noticed that your test is not being run in CI, though. Would you mind updating the Makefile `test` and `test-compile` targets to look inside `./cmd` in addition to `./pkg`. That should ensure that the test is being run.

Sure! I think in `test-compile`, `go list ./...` lists everything, including the packages in cmd. I updated `test`, so I think it's good!

cmd/tetra/getevents/getevents.go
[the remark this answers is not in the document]
mtardy: https://github.com/cilium/tetragon/pull/643/commits/bade1ca978592dc319364f64caee303bba2ef28e#diff-0e872caf2a3ba820a196526ac19d51a42136f47b2b2da3946aad52bbdc4b0d19R35-R39 !

willfindlay:
@mtardy Great, thanks. I noticed another minor problem: if you use `test-compile` to compile and run the getevents test, it is no longer able to find `testdata/events.json` since the relative path has changed.

I would suggest the following:

1. Move `testdata/events.json` into the top-level testdata directory and out of the `cmd/` hierarchy.
2. Implement something similar to https://github.com/cilium/tetragon/blob/ed2cc045bf888e135be1786b2205ae0de32f42ef/pkg/testutils/contrib.go#L11-L15 for the testdata path and use that in the test.

mtardy:
> @mtardy Great, thanks. I noticed another minor problem: if you use `test-compile` to compile and run the getevents test, it is no longer able to find `testdata/events.json` since the relative path has changed.
> 
> I would suggest the following:
> 
> 1. Move `testdata/events.json` into the top-level testdata directory and out of the `cmd/` hierarchy.
> 2. Implement something similar to https://github.com/cilium/tetragon/blob/ed2cc045bf888e135be1786b2205ae0de32f42ef/pkg/testutils/contrib.go#L11-L15
>     for the testdata path and use that in the test.

I took the liberty to factorize this one into a more generic one here https://github.com/cilium/tetragon/pull/643/commits/9e8cd8be3fdde766935cfb8a454a44884f645e9d, we can always revert if it's not a good idea.

cmd/tetra/getevents/getevents.go
```diff
@@ -29,7 +29,13 @@ var GetEncoder = func(w io.Writer, colorMode encoder.ColorMode, timestamps bool,
 	return json.NewEncoder(w)
 }
 
-func getRequest(includeFields, excludeFields []string, namespaces []string, host bool, processes []string, pods []string) *tetragon.GetEventsRequest {
+// GetFilter returns a filter for an event stream based on configuration options.
+var GetFilter = func() *tetragon.Filter {
```
kkourt, line 50: why not:
```
func GetFilter() *tetragon.Filter {
```

?

Also, does it need to be global?

cmd/tetra/getevents/getevents.go
```diff
@@ -133,5 +139,6 @@ func New() *cobra.Command {
 	flags.Bool("host", false, "Get host events")
 	flags.Bool("timestamps", false, "Include timestamps in compact output")
 	viper.BindPFlags(flags)
+
```
kkourt, line 160: nit: white-space change.

cmd/tetra/getevents/getevents.go
```diff
@@ -21,6 +22,22 @@ import (
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
 )
 
+// DocLong documents the commands with some examples
+const DocLong = `This command prints and filter events by connecting to the server or via
```
kkourt, line 26: Does this have to be global?

pkg/testutils/cmd.go
```diff
@@ -125,3 +129,72 @@ func (cbp *CmdBufferedPipes) ParseAndLogCmdOutput(
...
+	go func(ch chan error) {
+		_, err := io.Copy(&buf, r)
+		defer r.Close()
+		if err != nil {
+			ch <- err
+		}
+		ch <- nil
+	}(done)
+
+	cmd.Execute()
```
kkourt, line 189: Not sure I fully understand this, but it seems that this depends on `MockPipedFile` having set os.Stdin? Wouldn't be better if we used https://pkg.go.dev/os/exec#Cmd.StdinPipe here rather than setting os.Stdin?

kkourt (commented):
Thanks!

Some questions from my side.

pkg/testutils/cmd.go
mtardy: Here it's not an `exec.Cmd` but a https://pkg.go.dev/github.com/spf13/cobra#Command, all this thing is just to replace stdin with a real pipe to trigger this custom redirection in the command execution. That's why I was suggesting we could use the `-` convention like `kubectl apply -f -` instead of trying to detect it via this 'trick'.

https://github.com/cilium/tetragon/blob/main/cmd/tetra/getevents/getevents.go#L114-L115

kkourt: ah, got it!

kkourt (approved):

cmd/tetra/getevents/getevents.go
mtardy: You would like this whitespace removed? Remove line 142?

willfindlay (approved):

PR #647, opened 2023-01-19
tetragon: tarball deployment

tixxdz:


Makefile
```diff
@@ -1,4 +1,5 @@
 GO ?= go
+export ARCH ?= amd64
```
olsajiri, line 2: is there a reason why we can't use x86 directly?

Makefile
```diff
@@ -13,6 +13,9 @@ TESTER_PROGS_DIR = "contrib/tester-progs"
 # Extra flags to pass to test binary
 EXTRA_TESTFLAGS ?=
 
+BUILD_PKG_DIR ?= $(shell pwd)/build/
+BUILD_PKG_DIR_ARCH=$(BUILD_PKG_DIR)$(ARCH)
+
```
olsajiri, line 18: do we need 2 variables for that? looks like BUILD_PKG_DIR could have the ARCH directly?

Dockerfile.tarball
```diff
@@ -0,0 +1,31 @@
...
+# COPY configurations
+COPY install/linux-tarball/usr /usr
+COPY --from=tetragon-release /etc/tetragon/ /etc/tetragon/
+
+# Our binaries
+COPY --from=tetragon-release /usr/bin/tetragon /usr/local/bin/
+COPY --from=tetragon-release /usr/bin/tetra /usr/local/bin/
+
+# Helper binaries
+COPY --from=tetragon-release /usr/bin/bpftool /usr/local/lib/tetragon/
```
olsajiri, line 19: do we need to install bpftool? I think it's part of each major distro

Dockerfile.tarball
```diff
@@ -0,0 +1,31 @@
+FROM cilium/tetragon:latest as tetragon-release
+
+RUN mkdir -p /etc/tetragon/tetragon.conf.d/
+
```
olsajiri, line 4: I saw all the config stuff is placed under /usr/local/lib/tetragon/tetragon.conf.d
do we need to keep both directories?

Makefile
tixxdz: Usually x86 refers more to i386 , and amd64 the way to refer to x86_64
https://github.com/golang/go/blob/master/src/go/build/syslist.go
https://wiki.debian.org/SupportedArchitectures 
https://clang.llvm.org/docs/CrossCompilation.html

And in other places too

Makefile
tixxdz: Ok will try to remove!

Dockerfile.tarball
tixxdz: Yes, @kkourt suggested to ship our bpftool since it is used part of bugtool logic! maybe it has more new features?

Dockerfile.tarball
tixxdz: The /etc/tetragon/ during install will be empty! hmm just to make it easy for users but no strong opinion on this

tixxdz: The logic is /usr/ for us and packages and /etc for administrators: https://github.com/cilium/tetragon/pull/635

Makefile
tixxdz: However yes the libbpf parts we have expect __TARGET_ARCH_x86 and __TARGET_ARCH_i386

Dockerfile.tarball
tixxdz: Yes I think we should ship our bpftool so we endup with same reports and logic with bugtool and bpttool for both k8s containers environments and standalone bare metal ones.

tixxdz: The systemd service contains our own env PATH first which I assume should work

olsajiri: ok, makes sense

docs/deployment/package/README.md
```diff
@@ -0,0 +1,152 @@
...
+   sudo rm -fr /usr/local/bin/tetragon
+   sudo rm -fr /usr/local/lib/tetragon/
+   ```
+
+4. Install new Tetragon version
+
+   ```bash
+   tar -xvf tetragon-v0.8.4-x86_64.tar.gz
+   sudo cp -vRf tetragon-v0.8.4-x86_64/* /
+   ```
```
olsajiri, line 92: I was going through the installation process and it works nicely,

just one nit.. do we need to pack the tetragon-* directory ?
could the tarball be based on '/' directly? like for go: 

`$ rm -rf /usr/local/go && tar -C /usr/local -xzf go1.19.5.linux-amd64.tar.gz`

kkourt (approved):
LGTM, assuming @olsajiri's feedback is addressed.

Thanks! (especially for the nice documentation!)

olsajiri (approved):

PR #652, opened 2023-01-24
CRD examples: Replace invalid TracingPolicy names

mtardy:
Many `metadata.name` of the `TracingPolicy` examples contained underscores in their name, and unfortunately, that makes it impossible to deploy them on Kubernetes. I assumed they were passed to Tetragon via flags using `--config-file` or similar. 

Maybe it would be interesting to implement the regex check in Tetragon to avoid this pitfall? 

Kubernetes objects names comply with lowercase RFC 1123 subdomain convention: they must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`)

### DIFF
--- a/crds/examples/fd_install_cap_changes.yaml
+++ b/crds/examples/fd_install_cap_changes.yaml
@@ -1,7 +1,7 @@
 apiVersion: cilium.io/v1alpha1
 kind: TracingPolicy
 metadata:
-  name: "fd_install"
+  name: "fd-install"
 spec:
   kprobes:
   - call: "fd_install"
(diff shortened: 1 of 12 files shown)


tixxdz (approved):
As long as they deploy now lgtm! I think we deployed the with `tetra` or yes with --config-file.

tixxdz:
> Maybe it would be interesting to implement the regex check in Tetragon to avoid this pitfall?

Yes makes sense! next task? ;-)

kkourt (approved):

PR #661, opened 2023-01-25
pkg/config: add k8s validation on `metadata.name`

mtardy:
This forces TracingPolicy to follow the same constraint on the metadata.name value when it is loaded directly as when it is added as a Kubernetes object.

I used the k8s package `validation` since it was already a dependency!

This is related to https://github.com/cilium/tetragon/pull/652.

### DIFF
--- a/pkg/bench/trace_bench_open.go
+++ b/pkg/bench/trace_bench_open.go
@@ -78,7 +78,7 @@ func (src traceBenchOpen) ConfigFilename(ctx context.Context, args *Arguments) s
 	tmpl := `
 apiVersion: cilium.io/v1alpha1
 metadata:
-  name: "sys_write_writev"
+  name: "sys-write-writev"
 spec:
   kprobes:
   - call: "__x64_sys_openat"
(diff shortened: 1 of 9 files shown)


willfindlay:
@mtardy I suspect that the test failures are related to tracingpolicies that we define in-line in some of our unit tests. Might be worth checking those to make sure they have a conformant name field.

mtardy:
> @mtardy I suspect that the test failures are related to tracingpolicies that we define in-line in some of our unit tests. Might be worth checking those to make sure they have a conformant name field.

Correct! I checked the project with `rg "kind: TracingPolicy" -A2 | rg -e "name: \"[^_]*_.*$"`

mtardy:
I should have realized that my regex was not selecting my intended invalid_name in my template... I used this one to select more `rg "metadata:" -A1 | rg -e "name: \"[^_]*_.*$"`

mtardy:
I renamed things like `__x86_something` into `x86-something`: https://github.com/cilium/tetragon/pull/661/files#diff-1623695c95cb31f804db6b5ceb8e6dad79810fd8038a36618e0c0e76bd5e0637R1019, it makes less sense but it's valid :D

mtardy:
I forgot that uppercase letters are not okay 😐 `rg "metadata:" -A1 | rg -e "name: \"([^_]*_.*|[a-z-]*[A-Z]+[a-z-a]*)"`, there was an issue with `testing-file-matchArgs` that was transformed into `testing-file-match-args` which makes less sense also but is compatible.

tpapagian (approved):
LGTM, thanks!

willfindlay (approved):

PR #662, opened 2023-01-25
eventchecker: implement checker names

willfindlay:
This commit implments eventchecker names for top-level checkers (i.e. checkers that check a process event). These names will show up in the error logs when a check fails. That should hopefully improve the readability of test logs (especially e2e tests) and help us diagnose why tests are failing.

Note that for most of the tests that have only one checker, I have opted to give them an empty name. This is because these are cases where disambiguation is (obviously) unhelpful. For some practical examples of why this construct is actually useful, please refer to the end-to-end tests.


kkourt (approved):
Thanks!

PR #667, opened 2023-01-27
Changing bpf prog load, map create event names to be more descriptive

sharlns:
Signed-off-by: Natalia Reka Ivanko <natalia@isovalent.com>

### DIFF
--- a/pkg/encoder/encoder.go
+++ b/pkg/encoder/encoder.go
@@ -268,7 +268,7 @@ func (p *CompactEncoder) EventToString(response *tetragon.GetEventsResponse) (st
 			}
 			return CapTrailorPrinter(fmt.Sprintf("%s %s %s bytes %d", event, processInfo, args, bytes), caps), nil
 		case "bpf_check":
-			event := p.Colorer.Blue.Sprintf("🐝 %-7s", "bpf_check")
+			event := p.Colorer.Blue.Sprintf("🐝 %-7s", "bpf_load")
 			attr := ""
 			if len(kprobe.Args) > 0 && kprobe.Args[0] != nil {
 				ba := kprobe.Args[0].GetBpfAttrArg()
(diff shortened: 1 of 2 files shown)


pkg/encoder/encoder.go
```diff
@@ -268,7 +268,7 @@ func (p *CompactEncoder) EventToString(response *tetragon.GetEventsResponse) (st
 			}
 			return CapTrailorPrinter(fmt.Sprintf("%s %s %s bytes %d", event, processInfo, args, bytes), caps), nil
 		case "bpf_check":
-			event := p.Colorer.Blue.Sprintf("🐝 %-7s", "bpf_check")
+			event := p.Colorer.Blue.Sprintf("🐝 %-7s", "bpf_load")
```
tixxdz, line 271: bpf_check is related to the bpf verifier so the description is correct, it could be even changed to bpf_verifier_check for users.

And that one could fail at the verifier so strictly speaking the load may not happen, I suggest to have another call hook after the bpf_check to sure that the load did happen.

How this is triggered do we have a CRD for it?

pkg/encoder/encoder.go
```diff
@@ -284,7 +284,7 @@ func (p *CompactEncoder) EventToString(response *tetragon.GetEventsResponse) (st
 			}
 			return CapTrailorPrinter(fmt.Sprintf("%s %s %s", event, processInfo, attr), caps), nil
 		case "security_bpf_map_alloc":
-			event := p.Colorer.Blue.Sprintf("🗺 %-7s", "bpf_map_alloc")
+			event := p.Colorer.Blue.Sprintf("🗺 %-7s", "bpf_map_create")
```
tixxdz, line 287: Could be the same but this one is not a big deal.

tixxdz (changes requested):
It could be ok , however if we have users relying on bpf check results that this can be confusing, the prog type is also available after the bpf_check so I guess we can be more specific. @olsajiri

pkg/encoder/encoder.go
jrfastab: The issue here is the use case is to have a hook when bpf programs are loaded not so much about the verifier. So for users bpf_load makes sense IMO and is more intuitive.

tixxdz: Alright, so let's improve it with another patch something like at `bpf_audit_prog(prog, audit_load || unload);` that could better assert things really happened, would this be ok?

tixxdz (approved):

pkg/encoder/encoder.go
tixxdz: nah seems it can be inlined, will see later.

PR #670, opened 2023-01-30
pkg/config: print name in validation error

mtardy:
I forgot to add that and it would have helped me a lot to debug errors...


willfindlay (approved):
Seems innocent enough

Issue #179, opened 2022-06-20
docs: describe generic kprobe/tracepoint operations


kkourt:
e.g., FollowFD, UnfollowFD

sunnoy:
i want to see it too

PR #673, opened 2023-01-31
Improve README and document tracing policies

mtardy:
Fixes https://github.com/cilium/tetragon/issues/179.

This PR:
- Add documentation on `TracingPolicy`
- Reorganize the main README
- Add a TOC and cleanup title ordering
- Add `tetra` homebrew installation instruction
- Add a link to last Tetragon KubeCon presentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,7 +220,12 @@ A second way is to pretty print the events using the
 [`tetra` CLI](https://github.com/cilium/tetragon/releases/latest).
 The tool also allows filtering by process, pod, and other fields.
 
-You can download and install it by the following command:
+If you are using `homebrew`, you can install the latest release with:
+```console
+brew install tetra
+```
+
+Or you can download and install the latest release with the following commands:
 
 ```
 GOOS=$(go env GOOS)
--- a/README.md
+++ b/README.md
@@ -22,9 +22,41 @@ can be configured in relation to individual workloads.
 
 ![Tetragon Overview Diagram](docs/images/smart_observability.png)
 
-## Functionality Overview #
+## Table Of Content
 
-### eBPF Real-Time ##
+* [Functionality Overview](#functionality-overview)
+    * [eBPF Real-Time](#ebpf-real-time)
+    * [eBPF Flexibility](#ebpf-flexibility)
+    * [eBPF Kernel Aware](#ebpf-kernel-aware)
+* [Local Development](#local-development)
+* [Docker Deployment](#docker-deployment)
+* [Kubernetes Quickstart Guide](#kubernetes-quickstart-guide)
+    * [Requirements](#requirements)
+    * [Create a cluster](#create-a-cluster)
+        * [Kind](#kind)
+        * [GKE](#gke)
+    * [Deploy Tetragon](#deploy-tetragon)
+    * [Deploy Demo Application](#deploy-demo-application)
+    * [Explore Security Observability Events](#explore-security-observability-events)
+        * [Raw JSON events](#raw-json-events)
+        * [`tetra` CLI](#tetra-cli)
+    * [Process Execution](#process-execution)
+    * [File Access](#file-access)
+    * [Network Observability](#network-observability)
+    * [Privileged Execution](#privileged-execution)
+* [BTF Requirement](#btf-requirement)
+* [FAQ](#faq)
+* [Additional Resources](#additional-resources)
+* [Conference Talks](#conference-talks)
+    * [Book](#book)
+    * [Blog posts](#blog-posts)
+    * [Hands-on lab](#hands-on-lab)
+* [Community](#community)
+    * [Slack](#slack)
+
+## Functionality Overview
+
+### eBPF Real-Time
 
 Tetragon is a runtime security enforcement and observability tool. What this means is Tetragon applies
 policy and filtering directly in eBPF in the kernel. It performs the filtering,
@@ -75,7 +107,7 @@ For getting started with local development, you can refer to the [Development Gu
 
 [dev-guide]: ./docs/contributing/development/README.md
 
-## Docker deployment
+## Docker Deployment
 
 For getting started without having to deploy on a Kubernetes cluster, please refer to the [Docker deployment guide][docker-deployment].
 
@@ -674,7 +706,7 @@ $ COSIGN_EXPERIMENTAL=1 cosign verify --certificate-github-workflow-repository c
 It can be validated that the SBOM image was signed using Github Actions in the Cilium
 repository from the `Issuer` and `Subject` fields of the output.
 
-# FAQ
+## FAQ
 
 **Q:** Can I install and use Tetragon in standalone mode (outside of k8s)?
 
@@ -696,9 +728,9 @@ to run Tetragon.
 **A:** You need to add a signed-off-by line to your commit messages. The easiest way to do
 this is with `git fetch origin/main && git rebase --signoff origin/main`. Then push your changes.
 
-# Additional Resources
+## Additional Resources
 
-### Conference talks
+## Conference Talks
 
 [Uncovering a Sophisticated Kubernetes Attack in Real-Time](https://www.youtube.com/watch?v=bohnofE_dvw) - Jed Salazar & Natália Réka Ivánkó, KubeCon EU, 2020
 
@@ -724,9 +756,9 @@ this is with `git fetch origin/main && git rebase --signoff origin/main`. Then p
 
 [Security Observability with eBPF and Tetragon](https://isovalent.com/labs/) - Natália Réka Ivánkó, Roland Wolters, Raphaël Pinson
 
-# Community
+## Community
 
-## Slack
+### Slack
 
 Join the Tetragon [Slack channel](https://cilium.herokuapp.com/) to chat with developers, maintainers, and other users. This
 is a good first stop to ask questions and share your experiences.
--- a/README.md
+++ b/README.md
@@ -537,9 +537,7 @@ cases without Kubernetes, the same YAML configuration can be passed via a flag
 to the Tetragon binary or via the `tetra` CLI to load the policies via gRPC.
 
 For more information on `TracingPolicy` and how to write them, see the
-[`TracingPolicy` Guide](doc/todo.md).
-
-add this link in the guide https://github.com/cilium/tetragon/blob/main/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml
+[`TracingPolicy` Guide](docs/tracingpolicy/README.md).
 
 #### Use case 1: File Access
 
@@ -806,6 +804,8 @@ this is with `git fetch origin/main && git rebase --signoff origin/main`. Then p
 [You and Your Security Profiles; Generating Security Policies with the Help of eBPF](https://www.youtube.com/watch?v=EhQI1qPVb0E) - John Fastabend & Natália Réka Ivánkó, eBPF Day North America, 2022
 
 [Container Security and Runtime Enforcement with Tetragon](https://www.youtube.com/watch?v=fw40ROmswbM) - Djalal Harouni, eBPF Summit, 2022
+
+[Securing the Superpowers: Who Loaded That eBPF Program?](https://www.youtube.com/watch?v=UBVTJ0LeXxc) - John Fastabend & Natália Réka Ivánkó, CloudNative SecurityCon NA, 2023
 
 ### Book
 
--- a/README.md
+++ b/README.md
@@ -22,38 +22,6 @@ can be configured in relation to individual workloads.
 
 ![Tetragon Overview Diagram](docs/images/smart_observability.png)
 
-## Table Of Content
-
-* [Functionality Overview](#functionality-overview)
-    * [eBPF Real-Time](#ebpf-real-time)
-    * [eBPF Flexibility](#ebpf-flexibility)
-    * [eBPF Kernel Aware](#ebpf-kernel-aware)
-* [Local Development](#local-development)
-* [Docker Deployment](#docker-deployment)
-* [Kubernetes Quickstart Guide](#kubernetes-quickstart-guide)
-    * [Requirements](#requirements)
-    * [Create a cluster](#create-a-cluster)
-        * [Kind](#kind)
-        * [GKE](#gke)
-    * [Deploy Tetragon](#deploy-tetragon)
-    * [Deploy Demo Application](#deploy-demo-application)
-    * [Explore Security Observability Events](#explore-security-observability-events)
-        * [Raw JSON events](#raw-json-events)
-        * [`tetra` CLI](#tetra-cli)
-    * [Process Execution](#process-execution)
-    * [File Access](#file-access)
-    * [Network Observability](#network-observability)
-    * [Privileged Execution](#privileged-execution)
-* [BTF Requirement](#btf-requirement)
-* [FAQ](#faq)
-* [Additional Resources](#additional-resources)
-* [Conference Talks](#conference-talks)
-    * [Book](#book)
-    * [Blog posts](#blog-posts)
-    * [Hands-on lab](#hands-on-lab)
-* [Community](#community)
-    * [Slack](#slack)
-
 ## Functionality Overview
 
 ### eBPF Real-Time
@@ -100,6 +68,37 @@ namespace and capabilities, sockets to processes, process file descriptor to
 filenames and so on. For example, when an application changes its privileges we
 can create a policy to trigger an alert or even kill the process before it has
 a chance to complete the syscall and potentially run additional syscalls.
+
+## Table Of Content
+
+* [Local Development](#local-development)
+* [Docker Deployment](#docker-deployment)
+* [Kubernetes Quickstart Guide](#kubernetes-quickstart-guide)
+    * [Requirements](#requirements)
+    * [Create a cluster](#create-a-cluster)
+        * [Kind](#kind)
+        * [GKE](#gke)
+    * [Deploy Tetragon](#deploy-tetragon)
+    * [Deploy Demo Application](#deploy-demo-application)
+    * [Explore Security Observability Events](#explore-security-observability-events)
+        * [Raw JSON events](#raw-json-events)
+        * [`tetra` CLI](#tetra-cli)
+* [Tetragon Events](#tetragon-events)
+    * [Process execution](#process-execution)
+        * [Use case 1: Monitoring Process Execution](#use-case-1-monitoring-process-execution)
+        * [Use case 2: Privileged Execution](#use-case-2-privileged-execution)
+    * [Generic tracing](#generic-tracing)
+        * [Use case 1: File Access](#use-case-1-file-access)
+        * [Use case 2: Network Observability](#use-case-2-network-observability)
+* [BTF Requirement](#btf-requirement)
+* [FAQ](#faq)
+* [Additional Resources](#additional-resources)
+* [Conference Talks](#conference-talks)
+    * [Book](#book)
+    * [Blog posts](#blog-posts)
+    * [Hands-on lab](#hands-on-lab)
+* [Community](#community)
+    * [Slack](#slack)
 
 ## Local Development
 
@@ -246,10 +245,48 @@ The `tetra` CLI is also available inside `tetragon` container.
 kubectl exec -it -n kube-system ds/tetragon -c tetragon -- tetra getevents -o compact
 ```
 
-Tetragon is able to observe several events, here we provide a few small
-samples that can be used as a starting point:
+## Tetragon Events
 
-### Process Execution ###
+Tetragon is able to observe critical hooks in the kernel through its sensors
+and generates enriched events from them. In the next sections we detail the
+available sensors and the events they produce:
+1. [Process execution](#process-execution): generating `process_exec` and
+   `process_exit` events.
+1. [Generic tracing](#tracepoints-and-kprobes): generating `process_kprobes`
+   and `process_tracepoint` events.
+
+Along, we present use cases on how they can be used as a starting point.
+
+### Process execution
+
+Tetragon observes process creation and termination with default configuration
+and generates `process_exec` and `process_exit` events:
+
+- The `process_exec` events include useful information about the execution of
+  binaries and related process information. This includes the binary image that
+  was executed, command-line arguments, the UID context the process was
+  executed with, the process parent information, the capabilities that a
+  process had while executed, the process start time, the Kubernetes Pod,
+  labels and more.
+- The `process_exit` events, as the `process_exec` event shows how and when a
+  process started, indicate how and when a process is removed. The information
+  in the event includes the binary image that was executed, command-line
+  arguments, the UID context the process was executed with, process parent
+  information, process start time, the status codes and signals on process
+  exit. Understanding why a process exited and with what status code helps
+  understand the specifics of that exit.
+
+Both these events include Linux-level metadata (UID, parents, capabilities,
+start time, etc.) but also Kubernetes-level metadata (Kubernetes namespace,
+labels, name, etc.). This data make the connection between node-level concepts,
+the processes, and Kubernetes or container environments.
+
+These events enable a full lifecycle view into a process that can aid an
+incident investigation, for example, we can determine if a suspicious process
+is still running in a particular environment. For concrete examples of such
+events, see the next use case on process execution.
+
+#### Use case 1: Monitoring Process Execution
 
 This first use case is monitoring process execution, which can be observed with
 the Tetragon `process_exec` and `process_exit` JSON events.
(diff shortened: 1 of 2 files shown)


kkourt:
:wave: Any reason this is a draft?

mtardy:
> :wave: Any reason this is a draft?

Sure! The main doc on TracingPolicy is not part of the PR yet and unfinished.

README.md
```diff
@@ -214,10 +254,43 @@ The `tetra` CLI is also available inside `tetragon` container.
...
+Along, we present use cases on how they can be used as a starting point.
 
-### Process Execution ###
+### Process execution
+
+Tetragon observes process creation and termination with default configuration
+and generates `process_exec` and `process_exit` events:
+
+- The `process_exec` events include useful information about the execution of
+  binaries and related process information. This includes the binary image that
```
sharlns, line 275: I'd call out that we can see the Kubernetes Identity Aware metadata (K8s namespace, pod labels, name) in both process_exec and process_exit. In K8s context, this would help us to tie the process back to a specific workload. In non K8s environment, I believe we can see the container info, which would help to tie the process to a specific container.

README.md
```diff
@@ -214,10 +254,43 @@ The `tetra` CLI is also available inside `tetragon` container.
...
+- The `process_exit` events, as the `process_exec` event shows how and when a
+  process started, indicate how and when a process is removed. The information
+  in the event includes the binary image that was executed, command-line
+  arguments, the UID context the process was executed with, process parent
+  information, process start time, the status codes and signals on process
+  exit. Understanding why a process exited and with what status code helps
+  understand the specifics of that exit.
+
+These events enable a full lifecycle view into a process that can aid an
+incident investigation, for example, if a suspicious process is still running
```
sharlns, line 289: ... for example, we can determine if a suspicious process is still running in an environment.

README.md
```diff
@@ -402,22 +475,64 @@ Example `process_exec` and `process_exit` events can be:
 </p>
 </details>
 
-For the rest of the use cases we will use the Tetragon CLI to give the output.
+#### Use case 2: Privileged Execution
+
+Tetragon also provides the ability to check process capabilities and kernel namespaces.
```
sharlns, line 480: "and kernel namespace access. This information would help us determine which process or Kubernetes pod has started with privileges or host namespace access that it should not have or gained privileges and host namespace access later in the lifecycle that it should not have. This would help us answer questions like "which Kubernetes pods are running with CAP_SYS_ADMIN in my cluster? which Kubernetes pods have host network or pid namespace access in my cluster?" 

Feel free to reword it, but it would be worth calling the exact use cases

README.md
```diff
@@ -402,22 +475,64 @@ Example `process_exec` and `process_exit` events can be:
...
+# then save and exit
+```
+Restart the Tetragon daemonset:
+```
+kubectl rollout restart -n kube-system ds/tetragon
+```
+
+As a second step, let's start monitoring the Security Observability events from the privileged `test-pod` workload:
+```bash
+kubectl logs -n kube-system -l app.kubernetes.io/name=tetragon -c export-stdout -f | tetra getevents --namespace default --pod test-pod
```
sharlns, line 497: Any chance we can use something else rather than test-pod? (e.g nginx?)

README.md
```diff
@@ -402,22 +475,64 @@ Example `process_exec` and `process_exit` events can be:
 </p>
 </details>
 
-For the rest of the use cases we will use the Tetragon CLI to give the output.
+#### Use case 2: Privileged Execution
+
+Tetragon also provides the ability to check process capabilities and kernel namespaces.
+
+As a first step let's enable visibility to capability and namespace changes via the configmap
```
sharlns, line 482: It would be nice to describe what these flags do exactly in like 2 sentence. I think with this we expand every JSON event with caps/ns and we extract these when a process starts from a kernel struct. But I'm not 100% sure, would be nice to double check.

README.md
```diff
@@ -402,22 +475,64 @@ Example `process_exec` and `process_exit` events can be:
...
+kubectl apply -f https://raw.githubusercontent.com/cilium/tetragon/main/testdata/specs/testpod.yaml
+```
+
+If you observe the output in the first terminal, you can see the container start with `CAP_SYS_ADMIN`:
+```bash
+🚀 process default/test-pod /bin/sleep 365d                🛑 CAP_SYS_ADMIN
+🚀 process default/test-pod /usr/bin/jq -r .bundle         🛑 CAP_SYS_ADMIN
+🚀 process default/test-pod /usr/bin/cp /kind/product_name /kind/product_uuid /run/containerd/io.containerd.runtime.v2.task/k8s.io/7c7e513cd4d506417bc9d97dd9af670d94d9e84161c8c8 fdc9fa3a678289a59/rootfs/ 🛑 CAP_SYS_ADMIN
+```
+
```
sharlns, line 511: Can we do another use case "Use Case 3: Namespace access"? An easy way would be to monitor whenever the sys_setns syscall is invoked and just kubectl exec into a K8s pod, and run nsenter into different kernel namespaces. E.g:  "kubectl exec -it test-pod -- nsenter -t 1 -m -u -n -i -p bash" (in case of test-pod). And see what happens.

CRD:
```
kind: TracingPolicy
metadata:
  name: "namespace-change"
spec:
  kprobes:
  - call: "__x64_sys_setns"
    syscall: true
    args:
    - index: 0
      type: "int"
```

README.md
```diff
@@ -402,22 +475,64 @@ Example `process_exec` and `process_exit` events can be:
...
+For more information on `TracingPolicy` and how to write them, see the
+[`TracingPolicy` Guide](docs/tracingpolicy/README.md).
 
-`TracingPolicy` is a user-configurable Kubernetes custom resource definition (CRD)
-that allows users to trace arbitrary events in the kernel and define actions
-to take on a match. For bare metal or VM use cases without Kubernetes a YAML
-configuration file can be used.
+#### Use case 1: File Access
+
+This use case is file access, which can be observed with the Tetragon
```
sharlns, line 531: The first use case is file access,

README.md
```diff
@@ -402,22 +475,64 @@ Example `process_exec` and `process_exit` events can be:
...
+
+For more advanced use cases, Tetragon can observe tracepoints and arbitrary
+kernel calls via kprobes. For that, Tetragon must be extended and configured
+with custom resources objects named `TracingPolicy`. It can then generates
+`process_tracepoint` and `process_kprobes` events.
 
-### File Access
+`TracingPolicy` is a user-configurable Kubernetes custom resource that allows
+users to trace arbitrary events in the kernel and optionally define actions to
+take on a match, for enforcement for example. For bare metal or VM use cases
```
sharlns, line 522: .. take on a match. For example, in case all the selectors are matching in the TracingPolicy, we can send a Sigkill signal to the process or override the return value of the system call.

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
+# Tracing Policy
+
+`TracingPolicy` is a user-configurable Kubernetes custom resource (CR) that
+allows users to trace arbitrary events in the kernel and optionally define
+actions to take on a match, for enforcement for example. Currently, two types
+of events are supported: kprobes and tracepoints, but others may be added in
+the future (e.g., uprobes) by following a similar approach.
+
+Note that `TracingPolicy` can be considered low-level since they require
+knowledge about the Linux kernel and containers to be written correctly.
```
sharlns, line 10: to be written correctly. In the future, we are considering to add a high-level RuntimeSecurityPolicy which would take this abstraction away.

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+For the complete custom resource definition (CRD) refer to the YAML file
+[`cilium.io_tracingpolicies.yaml`](https://github.com/cilium/tetragon/blob/main/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml).
+One practical way to explore the CRD is to use `kubectl explain` against a
+Kubernetes API server on which it is installed, for example `kubectl explain
+tracingpolicy.spec.kprobes` provides field-specific documentation and details
+on kprobe spec.
+
+For bare metal or VM use cases without Kubernetes, the same YAML configuration
+can be passed via a flag to the Tetragon binary or via the `tetra` CLI to load
+the policies via gRPC.
```
sharlns, line 21: Can we make an example how to load it as a config file? I assume this is just an argument of the "--config-file" flag?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+kernel function.
+
+```yaml
+spec:
+  kprobes:
+  - call: "fd_install"
+    syscall: false
+```
+
+The `fd_install` kernel function is called each time a file descriptor must be
```
sharlns, line 114: "a file descriptor is intalled into ..."

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+```yaml
+args:
+- index: 1
+  type: "char_buf"
+  returnCopy: true
+  sizeArgIndex: 3
+- index: 2
+  type: "size_t"
+```
+This tells the printer to skip printing the `int` arg because it's not useful.
```
sharlns, line 205: Can we include which arguments are supported atm? Or just put a pointer to the CRD?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+
+### Selectors
+
+A `TracingPolicy` can contain from 0 to 5 selectors. A selector is composed of
```
sharlns, line 209: 5 is a maximum?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+1 or more filters. The available filters are the following:
```
sharlns, line 210: Do we have any limit on filters for a selector?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+- matchArgs:
+  - index: 1
+  operator: "Prefix"
+  values:
+  - "/etc"
+```
+In this situation, an event will be created every time a process tries to
+access a file under `/etc`.
+
+Although it makes less sense, you can also match over the first argument, to
```
sharlns, line 260: What is the use case for this?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+
+**Further examples**
+
+In the previous example, we used the operator `Equal`, but we can also use the
+`Prefix` operator and match against all files under `/etc` with:
+```yaml
+selectors:
+- matchArgs:
+  - index: 1
+  operator: "Prefix"
```
sharlns, line 253: Can we put a pointer what type of operators are supported for matchArgs?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+provide filtering based on the value of the function return value. It allows
+you to filter on the return value, thus success, error or value returned by a
+kernel call.
+
+```yaml
+matchReturnArgs:
+- operator: "NotEqual"
+  values:
+  - 0
+```
```
sharlns, line 289: Can we put a use case here like in 2 sentence? E.g: detect all the failed "sys_setuid" system calls, e.g: when somebody tried to run "sudo" and failed?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+#### Return args filter
+
+Arguments filters can be specified under the `returnMatchArgs` field and
+provide filtering based on the value of the function return value. It allows
+you to filter on the return value, thus success, error or value returned by a
+kernel call.
+
+```yaml
+matchReturnArgs:
+- operator: "NotEqual"
```
sharlns, line 286: Can we include what operators are supported?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+based on the value of host pid of the process. For example, the following
+`matchPIDs` filter tells the BPF code that observe only hooks for which the
+host PID is equal to either `pid1` or `pid2` or `pid3`:
+
+```yaml
+- matchPIDs:
+  - operator: "In"
+    followforks: true
+    values:
+    - "pid1"
```
sharlns, line 303: What is the limit on values?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+#### PIDs filter
+
+PIDs filters can be specified under the `matchPIDs` field and provide filtering
+based on the value of host pid of the process. For example, the following
+`matchPIDs` filter tells the BPF code that observe only hooks for which the
+host PID is equal to either `pid1` or `pid2` or `pid3`:
+
+```yaml
+- matchPIDs:
+  - operator: "In"
```
sharlns, line 300: What operator types are supported?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+    values:
+    - "pid1"
+    - "pid2"
+    - "pid3"
+```
+
+**Further examples**
+
+Another example can be to collect all processes not associated with a
+container's init PID, which is equal to 1. In this way, we are able to detect if
```
sharlns, line 311: I think we need another filter to exclude host processes, no? So don't observe host processes + container init PID. May be:
```
- matchPIDs:
  - operator: NotIn
    followforks: false
    isNamespacePID: true
    values:
    - 1
    - 0
```
But would be great to double check.

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+#### Binaries filter
+
+Binary filters can be specified under the `matchBinaries` field and provide
+filtering based on the value of a certain binary name. For example, the
+following `matchBinaries` selector tells the BPF code to process only system
+calls and kernel functions that are coming from `cat` or `tail`.
+
+```yaml
+- matchBinaries:
+  - operator: "In"
```
sharlns, line 331: How many values can we have?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+    - "CAP_CHOWN"
+    - "CAP_NET_RAW"
+```
+
+This will match if: [`Effective` capabilities contain `CAP_CHOWN`] OR
+[`Effective` capabilities contain `CAP_NET_RAW`]
+
+- `type` can be: `Effective`, `Inheritable`, or `Permitted`.
+- `operator` can be `In` or `NotIn`
+- `values` can be any supported capability. A list of all supported
```
sharlns, line 599: How many values can I have?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+
+**Limitations**
+
+1. There is no limit in the number of capabilities listed under `values`.
+2. Only one `type` field can be specified under `matchCapabilities`.
+
+#### Namespace changes filter
+
+Namespace changes filter can be specified under the `matchNamespaceChanges`
+field and provide filtering based on calls that are changing Linux namespaces.
```
sharlns, line 610: How do we keep track of changes? I guess we already know what namespaces the process has access to and we match against a "delta" whenever certain system calls are invoked? That's my guess but would be nice to double check and put like 2 sentence about the logic.

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+
+The `unshare` command, or executing in the host namespace using `nsenter` can
+be used to test this feature. See a
+[demonstration example](https://github.com/cilium/tetragon/blob/c511bf632ee58d401aa0f0cf1df50b33760515e3/crds/examples/match_namespace_changes.yaml)
+of this feature.
+
+#### Capability changes filter
+
+Capability changes filter can be specified under the `matchCapabilityChanges`
+field and provide filtering based on calls that are changing Linux capabilities.
```
sharlns, line 630: The previous comment applies here as well. What's the logic for keeping track of cap changes? we know what capabilities the process has and we match against a "delta" when a certain system call is invoked?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+  - "CAP_SETUID"
+```
+
+See a [demonstration
+example](https://github.com/cilium/tetragon/blob/c511bf632ee58d401aa0f0cf1df50b33760515e3/crds/examples/match_capability_changes.yaml)
+of this feature.
+
+#### Actions filter
+
+Actions filters are a list of actions that execute when an appropriate selector
```
sharlns, line 648: Does the ordering matter here? I assume the "matchAction" selector has to be the last filter. It'd be nice to put a sentence about that here.

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+
+Note that `Sigkill`, `Override`, `FollowFD`, `UnfollowFD`, `CopyFD` and `Post`
+are executed directly in the kernel BPF code while `GetUrl` and `DnsLookup` are
+happening in userspace after the reception of events.
+
+##### Sigkill action
+
+`Sigkill` action terminates synchronously the process that made the call that
+matches the appropriate selectors from the kernel. In the example below, every
+`__x64_sys_write` system call with a PID not equal to 1 or 0 attempting to
```
sharlns, line 668: I guess these are the processes inside a container - so not host and not init and its children.

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+    matchActions:
+    - action: Sigkill
+```
+
+##### Override action
+
+`Override` action allows to modify the return value of call. While `Sigkill`
+will terminate the entire process responsible for making the call, `Override`
+will override the return value that was supposed to be returned with the value
+given in the `argError` field.
```
sharlns, line 704: And the parent process will remain running?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+  type: "int"
+```
+
+Note, that whenever we would like to follow a file descriptor with a `FollowFD`
+block, there should be a matching `UnfollowFD` block, otherwise the BPF map
+will be broken.
+
+##### CopyFD action
+
+The `CopyFD` action is specific to duplication of file descriptor use cases.
```
sharlns, line 820: What is the use case for this? Can we include 2 sentence?

docs/tracingpolicy/README.md
```diff
@@ -0,0 +1,1193 @@
...
+
+```yaml
+matchActions:
+- action: GetUrl
+  argUrl: http://ebpf.io
+```
+
+##### DnsLookup action
+
+The `DnsLookup` action can be used to perform a remote interaction such as
```
sharlns, line 874: What was the motivation behind DnsLookup and GetUrl? What are the use cases? I think for GetUrl we wanted to send notifications whenever a certain policy gives an event, but I don't know about DnsLookup. Also do we return with the IP address after the lookup or no?

sharlns (changes requested):
Looks amazing, great work! 🚀  I mainly left some comments to (1) add more examples or (2) specifications, more details when it was not clear. 

Otherwise good to go.

docs/tracingpolicy/README.md
mtardy: added

docs/tracingpolicy/README.md
mtardy: yep https://github.com/cilium/tetragon/blob/745f6d2bab2ae98f40ec3c80a405a51e81dec38a/cmd/tetragon/flags.go#L40

docs/tracingpolicy/README.md
mtardy: Yes I wanted to add details on the various args we support!

docs/tracingpolicy/README.md
mtardy: As I understand yes:
https://github.com/cilium/tetragon/blob/c561a38884320b6d1486df34b056515f2eb0caa7/bpf/process/types/basic.h#L1273-L1277

docs/tracingpolicy/README.md
mtardy: I would love to know the answer :D

docs/tracingpolicy/README.md
mtardy: I think the idea from this comment is that we could think of a standard way of documenting each filter for next iteration:
- brief introduction
- inputs, operators supported
- examples

docs/tracingpolicy/README.md
mtardy: I think there is no use case, explaining the "Although it makes less sense" part, it's mostly to demonstrate the capabilities of the filter.

docs/tracingpolicy/README.md
mtardy: This is a nice use case but I don't think you can detect failed sudo like that because of how setuid binary works. But I will add something with fs access, a bit more boring...

docs/tracingpolicy/README.md
mtardy: The number of pids you can put? I don't know and I think we should test this.

docs/tracingpolicy/README.md
mtardy: I'm not sure to understand why to add PID 0, I think PID 0 which is the scheduler iirc is not really exposed to userspace and is a kernel thing?

docs/tracingpolicy/README.md
mtardy: I think the limit was 4 but this will improve the situation https://github.com/cilium/tetragon/pull/686.

docs/tracingpolicy/README.md
mtardy: I don't think there is any order restriction but it's mostly a guess, we'll need to dive into the code for that.

docs/tracingpolicy/README.md
mtardy: here we have the PID 0 thing again... interesting! I wonder why.

comment added on `kubectl exec`!

docs/tracingpolicy/README.md
mtardy: It completely depends on how the process will handle the return value! I would say mostly yes while many programs might terminate after an unsuccessful syscall that was supposed to succeed.

But I see your point here and will add some text.

docs/tracingpolicy/README.md
mtardy: I think there is no particular use case for `CopyFD` apart from making the whole `FollowFD` story works. Like you need this to handle the `dup` syscall family but that's not particularly interesting in itself.

docs/tracingpolicy/README.md
mtardy: The idea is explained here with the Thinkst canaries example, triggering some remote alerting systems by DNS or GET queries https://github.com/cilium/tetragon/pull/433. I don't think we return with the IP address, the point is mostly triggering some stuff and not really about the request itself.

mtardy (commented):
Thanks a lot Natalia!!

docs/tracingpolicy/README.md
mtardy: Yes it seems it's the idea:
https://github.com/cilium/tetragon/blob/1a68bf2681e79ec57c37863b1e5b809e7be04413/bpf/process/pfilter.h#L206-L209

docs/tracingpolicy/README.md
mtardy: Similarly:
https://github.com/cilium/tetragon/blob/1a68bf2681e79ec57c37863b1e5b809e7be04413/bpf/process/pfilter.h#L284-L297

mtardy:
I think some of the key takeaways (that I have not fixed yet) from the review would be:
- [ ] Use more realistic pod examples in introduction README https://github.com/cilium/tetragon/pull/673#discussion_r1111310265
- [ ] Add a new use case on namespace access https://github.com/cilium/tetragon/pull/673#discussion_r1111313834
- [ ] Document the available type of arguments https://github.com/cilium/tetragon/pull/673#discussion_r1111317788
- [ ] Update and clarify the limitations (and test them) https://github.com/cilium/tetragon/pull/673#discussion_r1111317960, https://github.com/cilium/tetragon/pull/673#discussion_r1111318032, https://github.com/cilium/tetragon/pull/673#discussion_r1111318715
- [ ] Improve the way filters are all presented with a more structured, unified description (systematically showing a list of operators available, inputs, etc.) https://github.com/cilium/tetragon/pull/673#discussion_r1111318376
- [ ] Improve description of the `changes` filters https://github.com/cilium/tetragon/pull/673#discussion_r1111322065

I will try to fix some this afternoon, but maybe some of them might be turned into issues in order for them to be addressed later.

docs/tracingpolicy/README.md
mtardy: I added the sentences about the logic!

docs/tracingpolicy/README.md
mtardy: I added the sentences about the logic!

docs/tracingpolicy/README.md
mtardy: I just added a link to the CRD but we could do better!

mtardy:
For me it's ready to merge @sharlns, I think the items mentioned in the last items could be improvement for future PRs. I tried to address all your comments and summarize the rest in the checklist above. Tell me what you think!

README.md
sharlns: Let's do it. I'd just use a simple nginx pod spec for this.

docs/tracingpolicy/README.md
sharlns: Sounds good :)

docs/tracingpolicy/README.md
sharlns: Yeah, let's ask in the tetragon channel :)

docs/tracingpolicy/README.md
sharlns: I'd ask in the tetragon channel :)

docs/tracingpolicy/README.md
sharlns: I'd clarify this in the tetragon channel :)

docs/tracingpolicy/README.md
sharlns: What do you mean exactly under "making the whole FollowFD story works"?

mtardy: I mean that it's just part of the feature that allows you to follow the file descriptor and path association. `CopyFD` is just a particular thing to handle the `dup` syscalls but is not that interesting in itself. In order to have a proper following of file descriptor/path association, we should `FollowFD` on `install_fd`, try to `UnfollowFD` on all the code parts that are able to close file descriptors, and use `CopyFD` on all the parts that are duplicating file descriptor for the same path.

README.md
mtardy: See https://github.com/cilium/tetragon/issues/705

README.md
mtardy: See https://github.com/cilium/tetragon/issues/706

sharlns (approved):

PR #674, opened 2023-01-31
tests/e2e: add ability to reset checker limits and retry demo app installation

willfindlay:
This should resolve a rare CI failure caused by the demo app failing to install correctly.


jrfastab (approved):

jrfastab:
I think this is OK we likely wont had any bugs with retries here because we are retrying the install of other things, not tetragon?

willfindlay:
@jrfastab correct. Basically, the issue is that we sometimes got eventchecker failures that weren't our fault because the demo app wasn't installing correctly. This will only retry when that installation fails, so we won't be missing legitimate failures.

tixxdz (approved):

PR #676, opened 2023-02-01
data: silence annoying debug log messages

willfindlay:
The "Data message received" and "Data message used" log messages flood the Tetragon logs when debug logging is enabled, to the point that it obfuscates all other debug messages. Bump these messages up to the trace level so that they are less annoying.

### DIFF
--- a/pkg/observer/data.go
+++ b/pkg/observer/data.go
@@ -45,7 +45,7 @@ func add(r *bytes.Reader, m *dataapi.MsgData) error {
 		dataMap.Add(m.Id, data)
 	}
 
-	logger.GetLogger().Debugf("Data message received id %v, size %v, total %v", m.Id, size, len(data))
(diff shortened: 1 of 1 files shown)


tixxdz (approved):

pkg/observer/data.go
```diff
@@ -45,7 +45,7 @@ func add(r *bytes.Reader, m *dataapi.MsgData) error {
 		dataMap.Add(m.Id, data)
 	}
 
-	logger.GetLogger().Debugf("Data message received id %v, size %v, total %v", m.Id, size, len(data))
+	logger.GetLogger().WithFields(nil).Tracef("Data message received id %v, size %v, total %v", m.Id, size, len(data))
```
kkourt, line 48: I guess the `WithFields` can be removed? Or is the idea to encode the fields there?

PR #677, opened 2023-02-01
tetragon: introduce Cgroup ID Tracker

tixxdz:
Add cgrpid_tracker inside execve_map_value struct which will be used to track every process by its pinned Cgroup ID.

The cgrpid_tracker will be set to the cgroup ID of the tracking cgroup, could be a pod/container/systemd service, etc. Right now we set it to current cgroup ID. As we advance in next patches we will set it to the tracking cgroup ID.

Signed-off-by: Djalal Harouni <djalal@isovalent.com>

### DIFF
--- a/bpf/lib/bpf_cgroup.h
+++ b/bpf/lib/bpf_cgroup.h
@@ -7,7 +7,7 @@
 #include "hubble_msg.h"
 #include "bpf_helpers.h"
 #include "environ_conf.h"
-#include "bpf_helpers.h"
+#include "process.h"
 
 #define NULL ((void *)0)
 
@@ -389,6 +389,65 @@ __init_cgrp_tracking_val_heap(struct cgroup *cgrp, cgroup_state state)
 		probe_read_str(&heap->name, KN_NAME_LENGTH - 1, name);
 
 	return heap;
+}
+
+/**
+ * __set_task_cgrpid_tracker() Sets the cgrpid_tracker of a task.
+ * It checks tetragon_conf if not available then exit.
+ * If tetragon_conf is available then checks if the task
+ * execve_map_value->cgrpid_tracker is set, if so do nothing.
+ * If not, then set the task execve_map_value->cgrpid_tracker
+ * to the tracking cgroup ID.
+ */
+static inline __attribute__((always_inline)) int
+__set_task_cgrpid_tracker(struct tetragon_conf *conf, struct task_struct *task,
+			  struct execve_map_value *execve_val, __u32 *error_flags)
+{
+	u32 flags = 0;
+	struct cgroup *cgrp;
+	__u64 cgrpid_tracker = 0;
+
+	/* Ensure that execve_val is not null */
+	if (unlikely(!conf) || unlikely(!execve_val))
+		return 0;
+
+	probe_read(&flags, sizeof(flags), _(&task->flags));
+	/* Skip kernel threads */
+	if (flags & PF_KTHREAD)
+		return 0;
+
+	/* Set the tracking cgroup id only if it was not set,
+	 * this avoids cgroup thread granularity mess!
+	 */
+	if (execve_val->cgrpid_tracker != 0)
+		return 0;
+
+	cgrp = get_task_cgroup(task, conf->tg_cgrp_subsys_idx, error_flags);
+	if (!cgrp)
+		return 0;
+
+	/* TODO: get current cgroup ancestors and their levels,
+	 *   so we can use them as a proper reference for setting,
+	 *   cgroup tracking IDs.
+	 *
+	 * For now we just use the current cgroup ID as tracker.
+	 */
+	cgrpid_tracker = tg_get_current_cgroup_id(cgrp, conf->cgrp_fs_magic);
+	if (!cgrpid_tracker) {
+		/* Failed to get cgrpid_tracker do nothing. This should never happen */
+		*error_flags |= EVENT_ERROR_CGROUP_ID;
+		return 0;
+	}
+
+	/* Update the execve_val with the tracking cgroup ID */
+	execve_val->cgrpid_tracker = cgrpid_tracker;
+
+	/* TODO:
+	 * Lookup cgroup data from the cgroup bpf map, create an entry
+	 * if not found, otherwise update previous with the corresponding
+	 * new cgroup state.
+	 */
+	return 0;
 }
 
 #endif // __BPF_CGROUP_
--- a/bpf/process/bpf_process_event.h
+++ b/bpf/process/bpf_process_event.h
@@ -646,14 +646,14 @@ static inline __attribute__((always_inline)) void
 __event_get_task_info(struct msg_execve_event *msg, __u8 op, bool walker,
 		      bool cwd_always)
 {
-	struct msg_process *curr;
+	struct msg_process *process;
 	struct task_struct *task;
 
 	msg->common.op = op;
 	msg->common.ktime = ktime_get_ns();
-	curr = &msg->process;
+	process = &msg->process;
 
-	if (cwd_always || curr->flags & EVENT_NEEDS_CWD) {
+	if (cwd_always || process->flags & EVENT_NEEDS_CWD) {
 		__u32 offset;
 		int err;
 		bool prealloc = false;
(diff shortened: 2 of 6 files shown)


bpf/process/bpf_process_event.h
```diff
@@ -514,13 +514,65 @@ get_namespaces(struct msg_ns *msg, struct task_struct *task)
...
+	 * Ensure that execve_val is not null.
+	 */
+	if (execve_val && execve_val->cgrpid_tracker) {
+		/* Set the k8s cgrpid with the tracking ID. */
+		msg->kube.cgrpid = execve_val->cgrpid_tracker;
+		return;
+	}
+
+	/* Gather Cgroup information using current context */
+
```
jrfastab, line 543: nit unneeded whitespace.

tixxdz: fixed! thanks

bpf/lib/bpf_cgroup.h
```diff
@@ -391,4 +391,68 @@ __init_cgrp_tracking_val_heap(struct cgroup *cgrp, cgroup_state state)
...
+		return 0;
+
+	/* Set the tracking cgroup id only if it was not set,
+	 * this avoids cgroup thread granularity mess!
+	 */
+	if (execve_val->cgrpid_tracker != 0)
+		return 0;
+
+	if (conf->tg_cgrp_subsys_idx != 0)
+		subsys_idx = conf->tg_cgrp_subsys_idx;
```
kkourt, line 428: I don't see why we need the branch here. If conf->tg_cgrp_subsys_idx is 0, then `subsys_idx` will end up being zero.


Can't we just do?
```
 get_task_cgroup(task, conf->tg_cgrp_subsys_idx, error_flags);
```

kkourt (changes requested):
Thanks!

I think the approach described in the message of the first commit makes sense. Do we need it for this patchset however? Wouldn't be better if we add it to the patch that implements the change?

The reason I'm suggesting this is because we might have to change our approach (for whatever reason) and then we will have something in our history which will be not true and this confusing. So, do you need that the description is needed to explain this patch, or should we add it to the one that actually does the changes?

bpf/lib/bpf_cgroup.h
tixxdz: Ah yes left over that indicated this is a cgroupv1 for readability, removed thanks

tixxdz:
> Thanks!
> 
> I think the approach described in the message of the first commit makes sense. Do we need it for this patchset however? Wouldn't be better if we add it to the patch that implements the change?
> 
> The reason I'm suggesting this is because we might have to change our approach (for whatever reason) and then we will have something in our history which will be not true and this confusing. So, do you need that the description is needed to explain this patch, or should we add it to the one that actually does the changes?

Yeh could have left if as it was more for a reference, but remove it just to merge stuff.
 
> In either case, can you also please squash the first two patches? I think it's better to have the definition of a function in the same patch as the point where it's being used. Thanks!

Done. Thanks!

kkourt (changes requested):
I believe with your changes, we end up executing the following code multiple times
```
	msg->kube.cgrpid = tg_get_current_cgroup_id(cgrp, cgrpfs_magic);
```
First we do it in `__event_get_cgroup_info()`:
https://github.com/cilium/tetragon/blob/7b16f6f1fccb686347e0336106556fe41bf03986/bpf/process/bpf_process_event.h#L635 

The same function then calls `__event_get_current_cgroup_id`:
https://github.com/cilium/tetragon/blob/7b16f6f1fccb686347e0336106556fe41bf03986/bpf/process/bpf_process_event.h#L642

`__event_get_current_cgroup_id` also seems to be doing the same thing:
https://github.com/cilium/tetragon/blob/7b16f6f1fccb686347e0336106556fe41bf03986/bpf/process/bpf_process_event.h#L549

I don't think we need to call this function twice. Am I missing something?

tixxdz:
> I believe with your changes, we end up executing the following code multiple times
> 
> ```
> 	msg->kube.cgrpid = tg_get_current_cgroup_id(cgrp, cgrpfs_magic);
> ```
> 
> First we do it in `__event_get_cgroup_info()`:
> 
> https://github.com/cilium/tetragon/blob/7b16f6f1fccb686347e0336106556fe41bf03986/bpf/process/bpf_process_event.h#L635
> 
> The same function then calls `__event_get_current_cgroup_id`:
> 
> https://github.com/cilium/tetragon/blob/7b16f6f1fccb686347e0336106556fe41bf03986/bpf/process/bpf_process_event.h#L642
> 
> `__event_get_current_cgroup_id` also seems to be doing the same thing:
> 
> https://github.com/cilium/tetragon/blob/7b16f6f1fccb686347e0336106556fe41bf03986/bpf/process/bpf_process_event.h#L549
> 
> I don't think we need to call this function twice. Am I missing something?

Good catch @kkourt yes we don't! fixing it.

tixxdz:
@kkourt updated, PTAL!

bpf/lib/bpf_cgroup.h
```diff
@@ -391,4 +391,64 @@ __init_cgrp_tracking_val_heap(struct cgroup *cgrp, cgroup_state state)
 	return heap;
 }
 
+/**
+ * __set_task_cgrpid_tracker() Sets the cgrpid_tracker of a task.
+ * It reads tetragon_conf if not available then exit.
```
kkourt, line 396: This comment seems to be outdate, since tetragon_conf is not read but passed as an argument.

bpf/lib/bpf_cgroup.h
```diff
@@ -391,4 +391,64 @@ __init_cgrp_tracking_val_heap(struct cgroup *cgrp, cgroup_state state)
...
+ * If tetragon_conf is available then checks if the task
+ * execve_map_value->cgrpid_tracker is set, if so do nothing.
+ * If not set then fetch task cgroup and cgroup level, compare it against
```
kkourt, line 399: This part also seems outdated, since there is no comparison with the cgroup level.

bpf/lib/bpf_cgroup.h
```diff
@@ -7,7 +7,7 @@
 #include "hubble_msg.h"
 #include "bpf_helpers.h"
 #include "environ_conf.h"
-#include "bpf_helpers.h"
+#include "process.h"
```
kkourt, line 10: The commit message of the first patch, mentions:
> by its pinned Cgroup ID.

It is not clear what the "pinned Cgroup ID" is. Some of the parts of the commit message seem to also
be out-of-date.

I suggest to use the following commit message for the first patch:

> This patch introduces the cgroup id tracker in the execve_map value.
> The goal is to track the cgroup id of the container that every process belongs to.
>
> As a first step, this patch retrieves the value of the cgroup id whenever a new task is created (fork),
> and inserts it into the execve_map. A subsequent patch will change the exec hook to use this value.

kkourt (changes requested):
Thanks!

The code LGTM. I think that some of the comments and commit messages are out-of-date. I think we should fix them before merging. I added some, but may have missed some. Can you please take a look and ensure that everything is up-to-date?

bpf/lib/bpf_cgroup.h
```diff
@@ -391,4 +391,64 @@ __init_cgrp_tracking_val_heap(struct cgroup *cgrp, cgroup_state state)
...
+
+	probe_read(&flags, sizeof(flags), _(&task->flags));
+	/* Skip kernel threads */
+	if (flags & PF_KTHREAD)
+		return 0;
+
+	/* Set the tracking cgroup id only if it was not set,
+	 * this avoids cgroup thread granularity mess!
+	 */
+	if (execve_val->cgrpid_tracker != 0)
```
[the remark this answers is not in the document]
tixxdz: Having it in this order makes sense cause we ignore kernel threads, the other way around is confusing.

bpf/lib/bpf_cgroup.h
```diff
@@ -391,4 +391,64 @@ __init_cgrp_tracking_val_heap(struct cgroup *cgrp, cgroup_state state)
...
+static inline __attribute__((always_inline)) int
+__set_task_cgrpid_tracker(struct tetragon_conf *conf, struct task_struct *task,
+			  struct execve_map_value *execve_val, __u32 *error_flags)
+{
+	u32 flags = 0;
+	struct cgroup *cgrp;
+	__u64 cgrpid_tracker = 0;
+
+	/* Ensure that execve_val is not null */
+	if (unlikely(!conf) || unlikely(!execve_val))
```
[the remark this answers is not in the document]
tixxdz: This makes the function more robust and can be used in any context.

bpf/lib/bpf_cgroup.h
tixxdz: Pinned cgroup ID could be the ID of container part of k8s, a systemd service, or a container outside of k8s. So not necessarily a container.

Improve the commit message.

bpf/lib/bpf_cgroup.h
tixxdz: Yeh due to the split.

tixxdz:
> Thanks!
> 
> The code LGTM. I think that some of the comments and commit messages are out-of-date. I think we should fix them before merging. I added some, but may have missed some. Can you please take a look and ensure that everything is up-to-date?

Comment messages strictly speaking are left over of the patch split, and they will be added back next PR.

I checked again, all make sense. Thanks!

kkourt (approved):
OK, LGTM, thanks.

PR #683, opened 2023-02-05
tetragon: Get proper exex cwd in case of no arguments

olsajiri:
If there are no arguments (api.EventDataArgs is not set) we trash the CWD string which follows arguments.

This can be reproduced by forcing arguments not to be stored like with following change:

```
  --- a/bpf/process/bpf_execve_event.c
  +++ b/bpf/process/bpf_execve_event.c
  @@ -51,7 +51,7 @@ event_args_builder(void *ctx, struct msg_execve_event *event)
          /* We use flags in asm to indicate overflow */
          compiler_barrier();
          probe_read(&mm, sizeof(mm), _(&task->mm));
  -       if (mm) {
  +       if (0 && mm) {
```

I'm not sure at the moment how to make test for this.


jrfastab:
happen to know when we introduced this? Did CWD really never work without args? Lots of execs don't push args I suspect.

olsajiri:
I don't think it would trigger that, we read args with:

```
1) read task->mm
2) read mm->arg_start and mm->arg_end
3) read/skip binary from arg_start (first arg)
4) read rest of the args if there's stil space event
```
so kernel tasks without mm could reproduce that.. the rest would need some read error later in the code path
also hubble does not have this problem.. for some reason this hunk that removed the else path was not backported (thanks @tixxdz for spotting that)

the commit time of the change is Aug 19 2022, but I think it got merged later than that

kkourt (approved):
🚀 Thanks!

tixxdz (approved):
Lgtm, thanks!

PR #686, opened 2023-02-06
matchBinaries improvements

tpapagian:
This PR fixes several issues regarding `matchBinaries` selector.

The first patch fixes some issues related to `matchBinaries` selector, but keeps the limit of 4 values. The second patch removes the limit of 4 values per `matchBinaries` selector. More details on the patches. 

Additionally, this PR adds support for NotIn operator in `matchBinaries` selector.

### DIFF
--- a/bpf/lib/generic.h
+++ b/bpf/lib/generic.h
@@ -70,7 +70,7 @@ struct names_map_key {
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	__uint(max_entries, 64);
+	__uint(max_entries, 256); /* maximum number of binary names for all matchBinary selectors */
 	__type(key, struct names_map_key);
 	__type(value, __u32);
 } names_map SEC(".maps");
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -135,7 +135,8 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid,
 	if (!heap)
 		return bin;
 
-	probe_read_str(heap->pathname, 255, filename);
+	memset(heap->pathname, 0, 256);
+	probe_read_str(heap->pathname, size, filename);
 	value = map_lookup_elem(&names_map, heap->pathname);
 	if (value)
 		return *value;
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -130,6 +130,10 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid, __u3
 	curr->ktime = ktime_get_ns();
 	curr->size = size + offsetof(struct msg_process, args);
 
+	// skip binaries check for long (> 255) filenames for now
+	if (flags & EVENT_DATA_FILENAME)
+		return 0;
+
 	heap = map_lookup_elem(&execve_heap, &zero);
 	if (!heap)
 		return 0;
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -138,7 +138,7 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid, __u3
 	if (!heap)
 		return 0;
 
-	memset(heap->pathname, 0, 256);
+	memset(heap->pathname, 0, PATHNAME_SIZE);
 	probe_read_str(heap->pathname, size, filename);
 	value = map_lookup_elem(&names_map, heap->pathname);
 	if (value)
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -94,8 +94,7 @@ event_args_builder(void *ctx, struct msg_execve_event *event)
 }
 
 static inline __attribute__((always_inline)) uint32_t
-event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid,
-		       __u32 flags, __u32 bin, void *filename)
+event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid, __u32 flags, void *filename)
 {
 	struct execve_heap *heap;
 	int64_t size = 0;
@@ -133,14 +132,14 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid,
 
 	heap = map_lookup_elem(&execve_heap, &zero);
 	if (!heap)
-		return bin;
+		return 0;
 
 	memset(heap->pathname, 0, 256);
 	probe_read_str(heap->pathname, size, filename);
 	value = map_lookup_elem(&names_map, heap->pathname);
 	if (value)
 		return *value;
-	return bin;
+	return 0;
 }
 
 __attribute__((section("tracepoint/sys_execve"), used)) int
@@ -150,7 +149,6 @@ event_execve(struct sched_execve_args *ctx)
 	struct msg_execve_event *event;
 	struct execve_map_value *parent;
 	struct msg_process *execve;
-	uint32_t binary = 0;
 	bool walker = 0;
 	__u32 zero = 0;
 	__u32 pid;
@@ -173,16 +171,13 @@ event_execve(struct sched_execve_args *ctx)
 	parent = event_find_parent();
 	if (parent) {
 		event->parent = parent->key;
-		binary = parent->binary;
 	} else {
 		event_minimal_parent(event, task);
 	}
 
 	execve = &event->process;
 	fileoff = ctx->filename & 0xFFFF;
-	binary = event_filename_builder(ctx, execve, pid, EVENT_EXECVE, binary,
-					(char *)ctx + fileoff);
-	event->binary = binary;
+	event->binary = event_filename_builder(ctx, execve, pid, EVENT_EXECVE, (char *)ctx + fileoff);
 
 	event_args_builder(ctx, event);
 	compiler_barrier();
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -83,12 +83,6 @@ struct selector_action {
 	__u32 act[];
 };
 
-struct selector_binary_filter {
-	__u32 arglen;
-	__u32 op;
-	__u32 index[4];
-};
-
 struct selector_arg_filter {
 	__u32 arglen;
 	__u32 index;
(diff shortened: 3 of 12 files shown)


pkg/selectors/selectors.go
```diff
@@ -13,6 +21,27 @@ type KernelSelectorState struct {
...
+	if k.binVals == nil {
+		k.binVals = make(map[uint32]string)
+	}
+
+	binMu.Lock()
+	idx := binIdx
+	binIdx++
+	binMu.Unlock()
+
+	k.binVals[idx] = binary
```
kkourt: I think the map assignment needs to also be under the lock.
See: https://go.dev/doc/faq#atomic_maps.

kkourt (changes requested):

pkg/selectors/selectors.go
tpapagian: Thanks, I have fixed that.

mtardy:
We should update the documentation on `matchBinaries` limitations in this PR when the doc PR merges!

Is it possible to add the `NotIn` operator along the way :p?

tpapagian:
> We should update the documentation on `matchBinaries` limitations in this PR when the doc PR merges!

Yes, let's see which PR is merged first and I will update accordingly.

> Is it possible to add the `NotIn` operator along the way :p?

Sure, I will try to do that in this PR!

tpapagian:
Moving to draft in order to do further improvement in this PR.

tpapagian:
I have added more stuff to this PR and now it is ready for review.

bpf/process/bpf_execve_event.c
```diff
@@ -135,7 +135,8 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid,
 	if (!heap)
 		return bin;
 
-	probe_read_str(heap->pathname, 255, filename);
+	memset(heap->pathname, 0, 256);
+	probe_read_str(heap->pathname, size, filename);
```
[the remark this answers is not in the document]
tpapagian: Now we support filenames larger than 256 bytes, right? So maybe I have to reconsider the limits there.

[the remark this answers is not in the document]
tpapagian: Added that, thanks.

kkourt (changes requested):
Thanks! This looks like a great improvement!

It took me a while to understand what is happening, so I've added some comment suggestions.

The only requested change is that I think we should check for the maximum length of the binary names and return an error if a user specifies a binary name >255.


Also:
> As a follow-up we can also clear entries from the (shared) names_map
when we remove kprobes. For now we also increase the size of that map to
256 entries. This means that we can define up to 256 unique binary names
among all matchBinaries selectors.

Let's create an issue for this? Since the same name may be used by multiple kprobes, I think we would need reference counting to know when to remove an entry. Or we can just scan everything. Another limitation is that we only catch binaries that are exec-ed after the policy started, which might be worth adding to the issue as well.

bpf/process/types/basic.h
```diff
@@ -928,14 +922,21 @@ static inline __attribute__((always_inline)) size_t type_to_min_size(int type,
 
 #define INDEX_MASK 0x3ff
 
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 256);
+	__type(key, __u32);
+	__type(value, __u32);
+} sel_names_map SEC(".maps");
```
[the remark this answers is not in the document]
tpapagian: Your understanding is correct and I have added a comment there.

pkg/selectors/kernel.go
```diff
@@ -640,39 +639,27 @@ func parseMatchCapabilityChanges(k *KernelSelectorState, actions []v1alpha1.Capa
...
+	if op != selectorOpIn {
+		return fmt.Errorf("matchBinary error: Only In operator is supported")
 	}
+	k.SetBinaryOp(op)
 	for _, s := range b.Values {
-		WriteSelectorUint32(k, k.AddBinaryName(s))
-	}
-	for l := len(b.Values); l < 4; l++ {
-		WriteSelectorUint32(k, math.MaxUint32)
+		k.AddBinaryName(s)
```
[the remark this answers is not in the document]
tpapagian: Added a check for path lengths > 255 chars

pkg/selectors/selectors.go
```diff
@@ -13,6 +13,8 @@ import (
 var (
 	binMu  sync.Mutex
 	binIdx uint32 = 1
```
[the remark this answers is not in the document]
tpapagian: Correct

bpf/process/types/basic.h
```diff
@@ -969,10 +969,19 @@ selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx)
...
+		// we just check if the key-value pair exists
+		// we ignore the value (should be always 1 which
+		// is set by the user-space during map initialization).
+		if (*op == op_filter_in) {
+			if (!bin_val) // value does not exist with op In
+				return 0;
+		} else if (*op == op_filter_notin) {
+			if (execve->binary != 0) { // binary == 0 -> no entry in names_map -> no matchBinary defines that
+				if (bin_val) // value exists with op NotIn
+					return 0;
```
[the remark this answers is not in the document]
tpapagian: You are correct. Maybe I missed that during testing. I changed that to what you proposed.

bpf/process/bpf_execve_event.c
```diff
@@ -130,6 +130,10 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid, __u3
 	curr->ktime = ktime_get_ns();
 	curr->size = size + offsetof(struct msg_process, args);
 
+	// skip binaries check for long (> 255) filenames for now
+	if (flags & EVENT_DATA_FILENAME)
+		return 0;
```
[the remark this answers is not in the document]
tpapagian: Added a check to reject MatchBinaries that are >255 chars

bpf/process/types/basic.h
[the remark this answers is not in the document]
tpapagian: I removed the `else { return 0; } ` as in the case where we don't have a `matchBinaries` selector `op` will be 0 (`op_filter_none`) and we will reject all of these cases.  

I believe this is fine, as we check the `op` value in the user-space.

tpapagian:
@kkourt thanks for the review. I have addressed all comments. 

I have also created https://github.com/cilium/tetragon/issues/722 with follow-ups.

bpf/process/types/basic.h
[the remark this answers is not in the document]
kkourt: Yap, makes sense to me! Thanks!

bpf/process/bpf_execve_event.c
```diff
@@ -135,7 +135,8 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid,
 	if (!heap)
 		return bin;
 
-	probe_read_str(heap->pathname, 255, filename);
+	memset(heap->pathname, 0, 256);
```
tixxdz, line 141: Usually it is a way better to have const here rather than hardconding values and losing the track... can be updated in next PR

pkg/selectors/kernel.go
```diff
@@ -639,16 +640,21 @@ func parseMatchCapabilityChanges(k *KernelSelectorState, actions []v1alpha1.Capa
...
-	WriteSelectorUint32(k, index)
+	if len(b.Values) > 4 {
+		return fmt.Errorf("Only support up to 4 values in MatchBinary")
+	}
+	for _, s := range b.Values {
+		WriteSelectorUint32(k, k.AddBinaryName(s))
+	}
+	for l := len(b.Values); l < 4; l++ {
+		WriteSelectorUint32(k, math.MaxUint32)
+	}
```
tixxdz, line 656: aren't these loops the same or could be one loop?

pkg/selectors/selectors.go
```diff
@@ -5,6 +5,14 @@ package selectors
 
 import (
 	"encoding/binary"
+	"sync"
+)
+
+// as we use a single names_map for all kprobes, so we have to use
+// a global variable to assign values to binary names
+var (
+	binMu  sync.Mutex
```
tixxdz, line 14: Usually a mutex should be part of the context it locks (struct fields...), so we know the locking

tixxdz: ah it is locking binIdx below hmm ok

pkg/selectors/selectors.go
tixxdz: hmm this is confusing, am I missing something here? there should be two mutexes one for the global one, and one per kernelselectorstate struct so we don't mix the two?

bpf/lib/generic.h
```diff
@@ -68,7 +68,7 @@ struct names_map_key {
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
-	__uint(max_entries, 64);
+	__uint(max_entries, 256);
```
tixxdz, line 71: same as before const make it easy to track

tixxdz (commented):
I tried to quickly review, but can continue later since this fixes a bug, however it seems a bit rushed and also would have been better the bug fix and the NotIN in separate PR so we can easily merge the former.

If you are sure about it feel free to merge, but I think those comments need to be addressed later

pkg/selectors/kernel.go
tpapagian: Here, we should call `WriteSelectorUint32` 4 times exactly.  The number of iterations in the first loop is equal to the number of binary names that the user provides (always <= 4). In the case where the first loop does < 4 iterations, the second loop will fill the empty entries with `math.MaxUint32`.

pkg/selectors/selectors.go
tpapagian: Yes, this is for a global variable.

pkg/selectors/kernel.go
tixxdz: Alright got you! so you can improve later those hardcoded 4 with their corresponding consts to make it easy

pkg/selectors/selectors.go
tixxdz: ok , see my other comments about using the same lock in an other context

pkg/selectors/selectors.go
tpapagian: I tried to simplify this a bit. We serialize all calls to `AddBinaryName` even if they refer to different sensors. This is not a performance-critical operation, so I believe this is fine (and simple).

Other than that, I don't believe that there is any reason to have a lock per KernelSelectorState as from my understanding always 1 thread accesses these variables.

tixxdz: ok, what I was referring is the global lock binMu is also used to lock the map of KernelSelectorState k.binVals[] per @kkourt comment https://github.com/cilium/tetragon/pull/686/commits/063434c9a04ccc8f61869f3a63e4690141c6e93f#r1104335002  in this patch context and I find it confusing.

So if AddBinaryName() are serialized, then why not have a helper that does the binIdx and binVals update lock/unlock inside, return the appropriate result and hide everything from AddBinaryName()... if that could work.

As said since we want to fix this feel free to merge, and please improve later if it makes sense.

bpf/process/bpf_execve_event.c
tpapagian: I added a commit to make that a define, thanks!

bpf/lib/generic.h
tpapagian: As this constant is only used here, I believe that it is fine as it is. I added a comment there to describe what this constant means.

tixxdz: Oh you are right sorry, thank you!

tpapagian:
> I tried to quickly review, but can continue later since this fixes a bug, however it seems a bit rushed and also would have been better the bug fix and the NotIN in separate PR so we can easily merge the former.

I feel that it is fine to have different commits in a single PR that refer to a single (matchBinaries) sub-system. Maybe the PR title was misleading as adding NotIn is not a fix but an improvement. So, I changed the title of this PR to better describe that.

pkg/selectors/selectors.go
tpapagian: > So if AddBinaryName() are serialized

`AddBinaryName` calls are not serialized by design. We serialize them using the global mutex. 

Yes, we could create new functions that do `binIdx` and `binVals` update with their appropriate locking but to be honest I found that a bit overkill. 

We can refactor that in a followup PR to make that even better. Thanks!

tixxdz: > > So if AddBinaryName() are serialized
> 
> `AddBinaryName` calls are not serialized by design. We serialize them using the global mutex.
> 

Hence my confusion, why are we using **one** same global mutex to serialize multiple different contexts!

> Yes, we could create new functions that do `binIdx` and `binVals` update with their appropriate locking but to be honest I found that a bit overkill.

There is no way around it, if you grep the code we do it everywhere when updating maps...

> 
> We can refactor that in a followup PR to make that even better. Thanks!

Alright , thanks!

tixxdz (approved):

mtardy:
🎉 Thanks @tpapagian

Y-dc:
Have you tested `tracepoint`, I tried it and it seems that the filtering is not successful

PR #688, opened 2023-02-07
tests/e2e: handle procRoot correctly in KinD clusters

willfindlay:
Because KinD clusters run in nested containers, the Tetragon Helm Chart was not correctly mounting the host procfs into the Tetragon container. This was causing potential flakes in the end-to-end tests by making it impossible to correctly initialize data for already running processes on the host system. Since we're not interested in offering full support for KinD, let's just make the tests work by hacking them to mount the real host procfs into the Tetragon container.

To do this, we create the KinD cluster with an extra volume that mounts host procfs then link it up with Tetragon by adding an extra volume mount via the Helm chart and modifying Tetragon's startup arguments. Finally, we update the end-to-end CI job to make use of this by removing some logic to use a single shared cluster and instead letting e2e framework take care of the cluster creation for us. If optimization is necessary in the future, we can look into adding some logic to use a single shared kind cluster within e2e framework, but I don't see this as necessary at the moment.

### DIFF
--- a/.github/workflows/run-e2e-tests.yaml
+++ b/.github/workflows/run-e2e-tests.yaml
@@ -11,7 +11,7 @@ concurrency:
 
 env:
   clusterName: tetragon-ci
-  ciliumCliVersion: v0.11.11
+  ciliumCliVersion: v0.12.11
 
 jobs:
   e2e-test:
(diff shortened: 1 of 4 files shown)


kkourt (approved):
🚀

kevsecurity (approved):

tixxdz (approved):
Seems good to go! Thank you!

willfindlay:
Just need to do a couple checks on my side then I will mark this as ready and ship it.

tpapagian (approved):
LGTM. thanks!

willfindlay:
The coveted quadruple approval. I'm gonna ship it.

PR #694, opened 2023-02-10
support for namespaced policies

kkourt:
Finalize support for namespaced policies.

A description of this feature can be found in: https://github.com/cilium/tetragon/blob/pr/kkourt/namespaced-policies/docs/content/en/docs/features/k8s-filtering.md

Previous PRs:
* https://github.com/cilium/tetragon/pull/695
* https://github.com/cilium/tetragon/pull/700
* https://github.com/cilium/tetragon/pull/723
* https://github.com/cilium/tetragon/pull/748
* https://github.com/cilium/tetragon/pull/749


docs/content/en/docs/features/k8s-filtering.md
```diff
@@ -0,0 +1,167 @@
...
+$ make image image-operator
+$ minikube image load --daemon=true cilium/tetragon:latest cilium/tetragon-operator:latest
+$ minikube ssh -- sudo mount bpffs -t bpf /sys/fs/bpf
+$ helm install --namespace kube-system \
+	--set tetragonOperator.image.override=cilium/tetragon-operator:latest \
+	--set tetragon.image.override=cilium/tetragon:latest  \
+	--set tetragon.enablePolicyFilter="true" \
+	--set tetragon.grpc.address="unix:///var/run/cilium/tetragon/tetragon.sock" \
+	tetragon ./install/kubernetes
+```
```
mtardy, line 60: Thank you so much for documenting this! :) We will need to improve the TracingPolicy reference as well.

For this code block and the following, can you remove the `$` and `#` at the beginning of the line and add `shell` just after the three backticks, like:
````
```shell
minikube start ...
```
````
It modifies the aspect like that (top is not done correctly, down is):
image

And not having the prompt make it easier to copy and paste.

Can you also try to separate input from the output (not using the `shell` thing with the output), adding "The output is similar to this" or other typical sentences between the two. See https://kubernetes.io/docs/contribute/style/style-guide/#separate-commands-from-output

Comment (kkourt) [reply docs/content/en/docs/features/k8s-filtering.md]: Thanks for the comment.

I prefer to use the `$` to indicate the shell prompt so that I can keep the output into the same box and differentiate between the two, but we need to be consistent which means that we need a common style and the kubernetes guide is as good as any (I guess). So I modified the commands to not have the shell prompt.

But, for the python test, I cannot figure out a better way to do it so I kept the python prompt. Technically, it is not "command output" so one could argue that it is not incompatible with the style guide.

Comment (mtardy) [reply docs/content/en/docs/features/k8s-filtering.md]: Just merged it, thanks a lot! :)

File: contrib/tester-progs/go/lseek-pipe/lseek-pipe.go [hunk 2]
```diff
@@ -0,0 +1,69 @@
...
+	"os"
+	"os/exec"
+	"strconv"
+
+	"golang.org/x/sys/unix"
+)
+
+// This program is used to test tracing policies. It can be called in two ways:
+//  - with three arguments: it uses the arguments to execute an lseek syscall
+//  - with no arguments, it three arguments from stdin, and executes itself with them
```
tpapagian, line 17: typo `... arguments, it three arguments ...` -> `... arguments: it uses three arguments ...`

pkg/sensors/tracing/policyfilter_test.go
```diff
@@ -0,0 +1,327 @@
...
+		m, err := cgroupsv2.NewSystemd("/", dir, -1, &res)
+		require.NoError(t, err)
+		t.Cleanup(func() {
+			m.DeleteSystemd()
+		})
+		for _, pid := range pids {
+			err = m.AddProc(pid)
+			require.NoError(t, err)
+		}
+		//path, err = FindCgroupDir(dir)
```
tpapagian, line 82: nit: maybe we can remove that comment.

contrib/rthooks/tetragon-oci-hook/main.go
```diff
@@ -0,0 +1,143 @@
...
+		return
+	}
+
+	hookName := os.Args[1]
+	switch hookName {
+	case "createContainer":
+		createContainerHook(log)
+	case "createRuntime":
+		// do nothing
+	case "postop":
```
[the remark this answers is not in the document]
kkourt: nice catch, thanks!

contrib/rthooks/tetragon-oci-hook/main.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/opencontainers/runtime-spec/specs-go"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+const (
+	logfname = "/var/log/tetragon-oci-hook.log"
+	// agentAddress = "localhost:54321"
```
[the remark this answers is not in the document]
kkourt: Yap, exactly! Removed the comment.

contrib/rthooks/tetragon-oci-hook/main.go
```diff
@@ -0,0 +1,143 @@
+package main
```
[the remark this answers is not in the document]
kkourt: correct!

Added the following comment:
```
+// OCI hook for tetragon
+// See: https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks
+//
+// This is an implementation of an OCI hook that can be used to notify the tetragon agent about
+// events such as container creation. It can be used by any container runtime that implements the
+// OCI Runtime Specification.
```

contrib/tester-progs/go/lseek-pipe/lseek-pipe.go
kkourt: thanks, fixed!

pkg/sensors/tracing/policyfilter_test.go
kkourt: yap, and FindCgroupDir with it I think. Thanks!

olsajiri (approved):
lgtm

PR #695, opened 2023-02-10
runtime hooks support for tetragon

kkourt:
This PR introduces a gRPC interface that can be used by run-time hooks to inform the tetragon agent about events such as container creation. Sensors can add hooks into these events so that they can be notified about them.


pkg/rthooks/dummy.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/cilium/tetragon/api/v1/tetragon"
+)
+
+type DummyHookRunner struct {
+	*testing.T
```
tpapagian, line 14: Why do we need that here? Why not use an empty struct?

tpapagian (approved):
Thanks! Other than minor issues, LGTM.

api/v1/tetragon/sensors.proto
```diff
@@ -108,4 +108,6 @@ service FineGuidanceSensors {
     rpc GetStackTraceTree(GetStackTraceTreeRequest) returns (GetStackTraceTreeResponse) {}
 
     rpc GetVersion(GetVersionRequest) returns (GetVersionResponse) {}
+
+    rpc RuntimeHook(RuntimeHookRequest) returns (RuntimeHookResponse) {}
```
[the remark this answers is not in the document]
kkourt: Good catch, should be fixed in thew new version.

pkg/rthooks/dummy.go
kkourt: Good point. I guess it was some leftover code. Removed it.

api/v1/tetragon/tetragon.proto
```diff
@@ -285,3 +285,23 @@ message ProcessLoader {
...
+        CreateContainer createContainer = 1;
+    }
+}
+
+message RuntimeHookResponse {}
+
+// CreateContainer informs the agent that a container was created
+message CreateContainer {
+    // cgroupPath is the cgroup path for the container
+    string cgroupPath = 1;
```
tixxdz, line 301: Can you past an example of the received cgroup path here? the received info

api/v1/README.md
```diff
@@ -198,6 +202,39 @@
...
 
+<a name="tetragon-CreateContainer"></a>
+
+### CreateContainer
+CreateContainer informs the agent that a container was created
+
+
+| Field | Type | Label | Description |
+| ----- | ---- | ----- | ----------- |
+| cgroupPath | [string](#string) |  | cgroupPath is the cgroup path for the container |
```
tixxdz, line 213: Also here cgroupPath with rootDir example if it makes sense

tixxdz (commented):
Could we have a workflow doc from how the OCI runtime hooks are triggered and how it propagates and relates to k8s watcher?

Also let's add OCI word here, clear confusion that runtime hooks are OCI runtime hooks right?

api/v1/tetragon/tetragon.proto
kkourt:
```
createContainer:{cgroupPath:\"/kubepods/besteffort/podb5f5a97e-b11b-416f-8bb3-722b4aa961a1/c7f5120dcae962bfea17a8ad5d5302f567f2ae35103aedc943feb9b69c3a0ce9\" rootDir:\"/run/containerd/io.containerd.runtime.v2.task/k8s.io/c7f5120dcae962bfea17a8ad5d5302f567f2ae35103aedc943feb9b69c3a0ce9/rootfs\"
```

kkourt:
> Could we have a workflow doc from how the OCI runtime hooks are triggered and how it propagates and relates to k8s watcher?
> 
> Also let's add OCI word here, clear confusion that runtime hooks are OCI runtime hooks right?

Updated description. The intended use is for OCI hooks, but that does not mean that we cannot use it from elsewhere.

tixxdz (approved):

PR #697, opened 2023-02-10
sensors: add VersionCheck hook

willfindlay:
Add a VersionCheck hook that can be used to perform optional kernel version checks when loading sensors.


willfindlay:
Marking as draft after internal discussions because there might be a better way to do this.

willfindlay:
Let's keep doing version checks in SpecHandler.

PR #698, opened 2023-02-16
Add start time to event cache object

kevsecurity:
The event cache takes a timestamp via its Add method. This is used in two places: once in the RetryInternal method and again as the timestamp of the event. The RetryInternal method actually wants a process start time, and that is typically what functions give it. Using this as the event timestamp is therefore incorrect.

This commit adds a further parameter to the Add method and the event cache object: process start time. Functions will have to provide the event timestamp and the process start time in order to use the event cache, but this will solve the issue of the event being given the wrong timestamp.

### DIFF
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -35,6 +35,7 @@ type CacheObj struct {
 	internal  *process.ProcessInternal
 	event     notify.Event
 	timestamp uint64
+	startTime uint64
 	color     int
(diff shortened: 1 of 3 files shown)


willfindlay (approved):
LGTM, but let's get a second ack on this as well

kkourt:
@tpapagian Could you please have a look?

tpapagian (approved):
LGTM, thanks!

PR #715, opened 2023-02-22
use  go 1.19.6

kkourt:
add a script to update cilium builder, and perform an update so that we can build with a new version of go.


tpapagian (approved):
LGTM, thanks!

PR #716, opened 2023-02-22
tetragon: cleanup grpc unix socket

tixxdz:
Since we rename unix socket, then an explicit cleanup is necessary.

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -489,10 +489,15 @@ func Serve(ctx context.Context, listenAddr string, server *server.Server) error 
 			log.WithError(err).Error("Failed to close gRPC server")
 		}
 	}(proto, addr)
-	go func() {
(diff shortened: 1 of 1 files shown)


cmd/tetragon/main.go
```diff
@@ -489,10 +489,15 @@ func Serve(ctx context.Context, listenAddr string, server *server.Server) error
 			log.WithError(err).Error("Failed to close gRPC server")
 		}
 	}(proto, addr)
-	go func() {
+	go func(proto, addr string) {
```
tpapagian, line 492: nit: I believe that we don't need the arguments here. `proto` and `addr` are accessible inside this function.

tixxdz: Indeed you are right! I guess I copied the function on top, used the same signature which is easy to read. We can keep it!

tpapagian (approved):

PR #719, opened 2023-02-22
Handle multiple URL and DNS selectors correctly

kevsecurity:
Currently we have experimental support for URL and DNS actions, which could be used to trigger Thinkst canaries. This experimental support incorrectly handles multiple selectors – it simply collects all the URLs in the kprobe and all the FQDNs in the kprobe into lists, and when a URL or FQDN action fires on it, it triggers everything in the corresponding list. This is obviously wrong.

This commit fixes this as follows. Each kprobe stores a table of URLs and FQDNs that it references, each entry with its own index. These indices are provided in the config to the match actions in the selectors, and the BPF program reports the matching index. In user space, the URL or FQDN is retrieved from the table using this index.

Note that only one URL and/or FQDN action is permitted per selector. It will be possible to enable multiple if necessary, but it is deemed that a single trigger should be sufficient to trigger any further triggers.

### DIFF
--- a/bpf/lib/generic.h
+++ b/bpf/lib/generic.h
@@ -37,6 +37,8 @@ struct msg_generic_kprobe {
 	__u64 id;
 	__u64 thread_id;
 	__u64 action;
+	__u32 action_arg_id; // only one URL or FQDN action can be fired per match
+	__u32 pad;
 	/* anything above is shared with the userspace so it should match structs MsgGenericKprobe and MsgGenericTracepoint in Go */
 	char args[24000];
 	unsigned long a0, a1, a2, a3, a4;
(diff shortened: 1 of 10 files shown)


bpf/lib/generic.h
```diff
@@ -37,6 +37,8 @@ struct msg_generic_kprobe {
 	__u64 id;
 	__u64 thread_id;
 	__u64 action;
+	__u32 url_id; // only one URL action can be fired per match
+	__u32 fqdn_id; // only one DNS action can be fired per match
```
kkourt, line 41: From the bpf-side, there is no difference between the URL and FQDN. It's just something that needs to happen in userspace. I think it would be better if we just use a single id for all the user-space actions and let the agent figure things out.

This will also make it easier to add more use-space actions (e.g., freeze).

I think this would simplify  the user-space side as well: having a single id table that has all the registered actions.

kkourt (commented):
Thanks!

bpf/lib/generic.h
kevsecurity: True. I can do that.

tixxdz (approved):

Issue #462, opened 2022-10-05
bpf: review and fix 0xeff masks


kkourt:
https://github.com/cilium/tetragon/commit/867c6f09be293fce279e09e28e9ed6392f935c85 fixes a 0xeff mask to 0xfff. There are other instances of such masks on the code, which might lead to similar bugs. Changing all of them, however, leads to verification issues so they need to be reviewed on a case-by-case basis.

see: https://github.com/cilium/tetragon/pull/458#discussion_r985674185

tpapagian:
Now sure that this is related but we still have some `0xeff` masks:
```
~/tetragon/bpf$ grep 3ff. -rnI
process/types/basic.h:177:	asm volatile("%[off] &= 0x3fff;\n" ::[off] "+r"(off)
process/types/basic.h:1490:		size = argm & 0x03ff;
process/data_event.h:24:	asm volatile("%[bytes] &= 0x3fff;\n"
process/pfilter.h:131:		asm volatile("%[o] &= 0x3ff;\n" ::[o] "+r"(o)
process/pfilter.h:150:		asm volatile("%[o] &= 0x3ff;\n" ::[o] "+r"(o)
process/bpf_process_event.h:386:	asm volatile("%[offset] &= 0x3ff;\n" ::[offset] "+r"(offset)
```

PR #720, opened 2023-02-22
tetragon: Fix action offset masking

olsajiri:
Kevin reported issue with longer selectors sending kprobes with
override action.

The problem is that we are masking action offset with wrong mask
bin(0xeff) == 0b111011111111, which was mangling offsets bigger
than 255, so the action code worked on top of wrong data.

Selector total length is 0x1000, so next longest mask to use here
is 0x7ff.

Note we should test longer selectors and make sure verifier is
happy with them.

Reported-by: Kevin Sheldrake <kevin.sheldrake@isovalent.com>

Fixes: https://github.com/cilium/tetragon/issues/462

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1310,13 +1310,13 @@ filter_read_arg(void *ctx, int index, struct bpf_map_def *heap,
 		if (f) {
 			bool postit;
 
-			asm volatile("%[pass] &= 0xeff;\n"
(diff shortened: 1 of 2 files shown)


kkourt (approved):
Thanks!

I think this resolves https://github.com/cilium/tetragon/issues/462, since there are no more `0xeff` values in the bpf code AFAICT.

kevsecurity (approved):

PR #723, opened 2023-02-23
introduce policyfilter mechanism

kkourt:
This PR introduces the policyfilter mechanism, both from bpf- and user-space side. It is currently unused, but it will be used in subsequent PRs for implementing namespaced tracing policies.

* Depends on: #700
* Part of: #694


pkg/policyfilter/state.go
```diff
@@ -0,0 +1,427 @@
...
+//  - periodic (or via external command) rescans to ensure that everything is up-to-date
+//  - optimization: policy<->pod matching caching so that we can go from pod -> matching policies
+//  without having to check all policies. Note that this would have to account for new
+//  policies added. Since we are only matching on namespace, we iterate over all policies for now
+//  since the performance impact of above optimization would be small.
+
+const (
+	// polMapSize is the number of entries for the (inner) policy map. It
+	// should be large enough to accommodate the number of containers
+	// running in a systme.
```
tpapagian, line 57: typo: systme -> system

pkg/policyfilter/map.go
```diff
@@ -0,0 +1,223 @@
...
+
+	ret, err := ebpf.NewMap(policyMapSpec)
+	if err != nil {
+		return PfMap{}, err
+	}
+
+	mapDir := bpf.MapPrefixPath()
+	pinPath := filepath.Join(mapDir, MapName)
+	os.Remove(pinPath)
+	os.Mkdir(mapDir, os.ModeDir)
```
tpapagian, line 57: Here you just remove the map. Why do you need to call `mkdir` again?

pkg/cgroups/cgroups.go
```diff
@@ -699,3 +700,64 @@ func CgroupNameFromCStr(cstr []byte) string {
 	}
 	return string(cstr[:i])
 }
+
+func tryHostCgoup(path string) error {
```
tpapagian, line 704: typo: tryHostCgoup -> tryHostCgroup

tpapagian (commented):
Thanks! I have added some initial comments from my side.

bpf/process/policy_filter.h
```diff
@@ -0,0 +1,50 @@
...
+	curr = execve_map_get_noinit(pid);
+	if (!curr || curr->cgrpid_tracker == 0)
+		return false;
+
+	policy_map = map_lookup_elem(&policy_filter_maps, &policy_id);
+	if (!policy_map)
+		return false;
+
+	policy_val = map_lookup_elem(policy_map, &curr->cgrpid_tracker);
+	return policy_val;
```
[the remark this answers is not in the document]
kkourt: makes sense -- changed it.

pkg/policyfilter/map.go
```diff
@@ -0,0 +1,223 @@
...
+
+// PfMap is a simple wrapper for ebpf.Map so that we can write methods for it
+type PfMap struct {
+	*ebpf.Map
+}
+
+// newMap returns a new policy filter map.
+func newPfMap() (PfMap, error) {
+	// use the generic kprobe program, to find the policy filter map spec
+	objName, _ := kernels.GenericKprobeObjs()
```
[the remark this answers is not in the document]
kkourt: I don't see why it wouldn't. We just need the spec for the map, which should be the same on all the objects.

pkg/policyfilter/state.go
kkourt: fixed, thanks!

pkg/policyfilter/map.go
kkourt: There are cases where the map does not exist (os.Remove() will just return an error, which we do not check) and neither does the directory. We could check if the directory exists or not but not sure if it's worth it.

pkg/cgroups/cgroups.go
kkourt: thanks, fixed.

pkg/policyfilter/map.go
tpapagian: Sounds reasonable, thanks!

pkg/policyfilter/map.go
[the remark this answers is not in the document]
tpapagian: Ok, I see. I miss-read that part.

bpf/process/bpf_generic_kprobe.c
```diff
@@ -90,6 +91,10 @@ generic_kprobe_start_process_filter(void *ctx)
 	config = map_lookup_elem(&config_map, &msg->idx);
 	if (!config)
 		return 0;
+
+	if (!policy_filter_check(config->policy_id))
+		return 0;
```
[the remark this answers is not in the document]
kkourt: Makes sense. Fixed it!

tpapagian (approved):
LGTM, thanks!

willfindlay (approved):

PR #724, opened 2023-02-23
Fix multiple file match bugs

kevsecurity:
This commit fixes a number of file match bugs:

* __filter_file_buf() was passing a length value of one less than the string length to cmpbytes() and as a result Equals and Prefix matches were testing one fewer bytes than required.

* ASM_RCMP (used in rcmpbytes()) was decrementing the string indices before testing if they were <1. This resulted in one fewer bytes being tested than required.

* ASM_RCMP50 comprisd of 2x ASM_RCMP20 and 1x ASM_RCMP5, totalling 45 iterations of ASM_RCMP, instead of 50.

* __filter_file_buf() tested failed postfix matches with a forward string match. This resulted in files that started with the postfix, but didn't end with it, matching when they shouldn't.

* cmpbytes() continued to loop to full number of iterations, even when the string length had been exhausted (inefficient).

* Added descriptions to cmpbytes() and rcmpbytes() to aid in providing the correct parameters.

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -295,10 +295,10 @@ parse_iovec_array(long off, unsigned long arg, int i, unsigned long max,
 			     : "c2", "r0");               \
 		if (c1 != c2)                             \
 			goto failed;                      \
-		n1--;                                     \
-		n2--;                                     \
 		if (n1 < 1 || n2 < 1)                     \
 			goto accept;                      \
+		n1--;                                     \
+		n2--;                                     \
 	}
 
 #define ASM_RCMP5        \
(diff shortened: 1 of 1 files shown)


kkourt (approved):

kevsecurity:
> I vaguely remember these things being off by one the other way and something about the \n\r chars but don't recall the details. Anyways tests should be able to verify this is all correct. (Note I only read the pr msg so far)./

Rudimentary tests (i.e. manipulating CRD and testing) seems to confirm the fixes work as expected.

kkourt:
added an issue for adding tests to catch the issues here: https://github.com/cilium/tetragon/issues/728

Issue #487, opened 2022-10-19
ARM64 support


kkourt:
This would allow to run tetragon on ARM machines, including VMs running on M1 machines.

See:
 * https://github.com/cilium/tetragon/issues/190
 * https://github.com/cilium/tetragon/pull/451
 * https://github.com/cilium/tetragon/pull/472
 
Let's recap the recent updates:
- https://github.com/cilium/tetragon/pull/734
- https://github.com/cilium/tetragon/pull/743
- https://github.com/cilium/tetragon/pull/752
- https://github.com/cilium/tetragon/pull/734
- https://github.com/cilium/tetragon/pull/816
- https://github.com/cilium/tetragon/pull/856
- https://github.com/cilium/tetragon/pull/880
- https://github.com/cilium/tetragon/pull/886
- https://github.com/cilium/tetragon/pull/884
- https://github.com/cilium/tetragon/pull/899
- https://github.com/cilium/tetragon/pull/900

We still have a minor few issues:
- https://github.com/cilium/tetragon/issues/786 _(this might not be really an issue)_
- https://github.com/cilium/tetragon/issues/776 _(this is very limited, only one test remaining)_
- https://github.com/cilium/tetragon/issues/955

We will continue the work on those issues but we can close this umbrella "arm64 support" issue since Tetragon now runs on arm64.

mtardy:
For the update we merged the following:
- https://github.com/cilium/tetragon/pull/743
- https://github.com/cilium/tetragon/pull/752
- https://github.com/cilium/tetragon/pull/734

So now Tetragon should compile and run transparently (like it does on amd64):
```shell
make tetragon tetragon-bpf
sudo ./tetragon --bpf-lib bpf/objs
```

However, we identified a few limitations:
- https://github.com/cilium/tetragon/issues/786
- https://github.com/cilium/tetragon/issues/776
- The CI is still `amd64`-centric
- While you can build it yourself, we still don't build and distribute Tetragon container images for arm64.

mtardy:
To let people play with Tetragon on arm64, I tried to address the last point:
> While you can build it yourself, we still don't build and distribute Tetragon container images for arm64.

- https://github.com/cilium/tetragon/pull/816

errordeveloper:
Hey folks, are there multi-arch images could try already? Just very keep to play with Teragon on amd64+arm64 cluster :)

errordeveloper:
[...]

PR #50, opened 2022-05-20
Setup CI to build multi-arch images with arm64 support

chancez:
Setup CI to build images that support both amd64 and arm64

PR #734, opened 2023-02-27
Add support for arm64 for building, running and testing

mtardy:
Thanks to the work done previously (see https://github.com/cilium/tetragon/issues/487), it is pretty straightforward to build Tetragon on arm64 and it seems to mostly work out of the box. The exec probes seem to be working, and I tried to load some TracingPolicy with kprobe hooks that just worked.

### Building

~~The makefile assumes that we are building using the clang docker image by default, and this one is built only for amd64 at the moment. I saw there was previously a PR attempt https://github.com/cilium/tetragon/pull/50 that was unfortunately reverted https://github.com/cilium/tetragon/pull/55 because it might have been really too early to build every images for arm64 (including the Tetragon one). It would simplify the workflow to build a multi-arch image and use it in the Makefile.~~

- [x] Build the clang image for arm64 via emulation, push a multi-arch image and update the Makefile with the new multi-arch image
   https://github.com/cilium/tetragon/pull/743

This part is done, and we can now easily build on arm64 just by using `make tetragon-bpf`.

### Tests

~~I tried running the tests on arm64 and some stuff broke with obvious issues displaying messages like:~~
```text
failed: creating perf_kprobe PMU: token __arm64___x64_sys_dup2: not found: no such file or directory
```
~~So we might have some quick wins on the tests and maybe some other more difficult things to fix.~~

- [x] Some docker image used in the test could not run on arm64, bumping tag was enough: https://github.com/cilium/tetragon/pull/734/commits/0be8dae2e45bd037432b341afa75927bc318167c
- [x] Issue: https://github.com/cilium/tetragon/pull/751
   PR: https://github.com/cilium/tetragon/pull/752
   Adapt the templates and the check to the new prefix-less syscalls symbols, see https://github.com/cilium/tetragon/pull/734/commits/d8fda5cd17a7b6d28d9533b2c0ee6a2c3cefb5bc.
- [x] Adapt a test on Tracepoint format because `char` is unsigned on arm64, contrary to amd64, see [Tracepoint: adapt a format test on task_newtask to arm64](https://github.com/cilium/tetragon/pull/734/commits/4ce7bdf4c359f198c7bb93b2669d57fb7bdf0387).

This part is done! Now tests can run on amd64 and arm64! 🥳 

### Trying Tetragon on arm64

```shell
# synchronize your git repo to my branch
git fetch && git checkout pr/mtardy/arm64
# compile Tetragon and the BPF programs
make tetragon tetragon-bpf
# run with
sudo ./tetragon --bpf-lib bpf/objs
```

### DIFF
--- a/bpf/include/vmlinux.h
+++ b/bpf/include/vmlinux.h
@@ -1411,6 +1411,8 @@ struct range {
 	u64 end;
 };
 
+#if defined(__TARGET_ARCH_x86)
+
 struct pt_regs {
 	long unsigned int r15;
 	long unsigned int r14;
@@ -1434,6 +1436,40 @@ struct pt_regs {
 	long unsigned int sp;
 	long unsigned int ss;
 };
+
+#elif defined(__TARGET_ARCH_arm64)
+/* definitions for arm64 in this vmlinux.h file might be incomplete or wrong
+ * for more information, see: https://github.com/cilium/tetragon/issues/786
+ */
+
+struct user_pt_regs {
+	__u64 regs[31];
+	__u64 sp;
+	__u64 pc;
+	__u64 pstate;
+};
+
+struct pt_regs {
+	union {
+		struct user_pt_regs user_regs;
+		struct {
+			u64 regs[31];
+			u64 sp;
+			u64 pc;
+			u64 pstate;
+		};
+	};
+	u64 orig_x0;
+	s32 syscallno;
+	u32 unused2;
+	u64 sdei_ttbr1;
+	u64 pmr_save;
+	u64 stackframe[2];
+	u64 lockdep_hardirqs;
+	u64 exit_rcu;
+};
+
+#endif
 
 struct desc_ptr {
 	short unsigned int size;
--- a/pkg/arch/arch.go
+++ b/pkg/arch/arch.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"runtime"
 	"strings"
+	"testing"
 )
 
 var supportedArchPrefix = map[string]string{"amd64": "__x64_", "arm64": "__arm64_"}
@@ -41,9 +42,15 @@ func AddSyscallPrefix(symbol string) (string, error) {
 	return addSyscallPrefix(symbol, runtime.GOARCH)
 }
 
-// MustAddSyscallPrefix is like AddSyscallPrefix but panics if the arch is
-// unsupported or if the symbol has already a prefix that doesn't correspond to
-// the running arch.
+// AddSyscallPrefixTestHelper is like AddSyscallPrefix but calls t.Fatal if the
+// arch is unsupported or if the symbol has already a prefix that doesn't
+// correspond to the running arch.
 //
 // It's a helper supposed to be used only in testing code.
-func MustAddSyscallPrefix(symbol string) string
+func AddSyscallPrefixTestHelper(t *testing.T, symbol string) string {
+	syscallName, err := AddSyscallPrefix(symbol)
+	if err != nil {
+		t.Fatal(err)
+	}
+	return syscallName
+}
--- a/pkg/arch/arch.go
+++ b/pkg/arch/arch.go
@@ -1,0 +1,49 @@
+package arch
+
+import (
+	"fmt"
+	"runtime"
+	"strings"
+)
+
+var supportedArchPrefix = map[string]string{"amd64": "__x64_", "arm64": "__arm64_"}
+
+func addSyscallPrefix(symbol string, arch string) (string, error) {
+	for prefix_arch, prefix := range supportedArchPrefix {
+		if strings.HasPrefix(symbol, prefix) {
+			// check that the prefix found is the correct one
+			if prefix_arch != arch {
+				return "", fmt.Errorf("expecting %s and got %s", supportedArchPrefix[arch], prefix)
+			}
+			return symbol, nil
+		}
+	}
+	if prefix, found := supportedArchPrefix[arch]; found {
+		return prefix + symbol, nil
+	}
+	return "", fmt.Errorf("unsupported architecture %s", arch)
+}
+
+// AddSyscallPrefix detects if the symbol is already prefixed with arch specific
+// prefix in the form of "__x64_" or "__arm64_", and if not, adds it based on
+// provided arch.
+//
+// Note that cilium/ebpf, retries to attach to an arch specific version of the
+// symbol after a failure thus making possible to attach to "sys_listen" which
+// will fail and retry with the corrected "__arm64_sys_listen" if you are
+// running on arm64. See https://github.com/cilium/ebpf/pull/304.  However, this
+// causes Tetragon users to believe that two symbols exist (following the same
+// example) "__arm64_sys_listen" and "sys_listen", which results in differents
+// events.  Because we actually know if a call is a syscall, we can pre-format
+// the symbol directly here to allow the user to specify the "sys_listen"
+// version while only registering the real symbol "__arm64_sys_listen".
+func AddSyscallPrefix(symbol string) (string, error) {
+	return addSyscallPrefix(symbol, runtime.GOARCH)
+}
+
+// MustAddSyscallPrefix is like AddSyscallPrefix but panics if the arch is
+// unsupported or if the symbol has already a prefix that doesn't correspond to
+// the running arch.
+//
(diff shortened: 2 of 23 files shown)


bpf/include/vmlinux.h
```diff
@@ -1411,6 +1411,8 @@ struct range {
 	u64 end;
 };
 
+#if defined(__TARGET_ARCH_x86)
```
tixxdz, line 1414: I think the kernel doesn't have these right?

Probably the best way to do it is to have different vmlinux.h per architecture, then we compile the right one according to the TARGET_ARCH, something like bpf/include/{amd64|arm64}/vmlinux.h ?

tixxdz:
I think you can submit this commit https://github.com/cilium/tetragon/pull/734/commits/d8fda5cd17a7b6d28d9533b2c0ee6a2c3cefb5bc  and the one related to it in a separate PR maybe?

pkg/arch/arch.go
```diff
@@ -41,9 +42,15 @@ func AddSyscallPrefix(symbol string) (string, error) {
...
-// unsupported or if the symbol has already a prefix that doesn't correspond to
-// the running arch.
+// AddSyscallPrefixTestHelper is like AddSyscallPrefix but calls t.Fatal if the
+// arch is unsupported or if the symbol has already a prefix that doesn't
+// correspond to the running arch.
 //
 // It's a helper supposed to be used only in testing code.
-func MustAddSyscallPrefix(symbol string) string
+func AddSyscallPrefixTestHelper(t *testing.T, symbol string) string {
+	syscallName, err := AddSyscallPrefix(symbol)
```
tixxdz, line 51: but what if we pass a prefix + symbole that corresponds to the right arch?  IMO it should not fail

mtardy: And it will not, it's this case https://github.com/cilium/tetragon/pull/734/files#diff-a79dc57275508064bab63e4361c0a043cd2f4225f0fb8d833448788294cc2ae9R19 (test case here https://github.com/cilium/tetragon/pull/734/files#diff-e7e583e8b6ff00f3e08fabf370b33e954d6419291b7d94f378ce4a165d6663c5L20)

tixxdz: nice thanks!

bpf/include/vmlinux.h
mtardy: I'm not sure to understand your question, but this is not a new thing from this PR, we are using this in libbpf currently for reading the registers https://github.com/cilium/tetragon/blob/main/bpf/libbpf/bpf_tracing.h#L18 of calls.

I mostly reused this because it's just more simple.

mtardy:
> I think you can submit this commit [d8fda5c](https://github.com/cilium/tetragon/commit/d8fda5cd17a7b6d28d9533b2c0ee6a2c3cefb5bc) and the one related to it in a separate PR maybe?

Yes, I could because it was possible thanks to this separate PR https://github.com/cilium/tetragon/pull/752 but what would be the benefits of doing so? It's only done in order for Tetragon tests to pass on an arm64 host. If we can merge this directly without squashing what would be the advantages of doing a separate PR?

tixxdz:
> > I think you can submit this commit [d8fda5c](https://github.com/cilium/tetragon/commit/d8fda5cd17a7b6d28d9533b2c0ee6a2c3cefb5bc) and the one related to it in a separate PR maybe?
> 
> Yes, I could because it was possible thanks to this separate PR #752 but what would be the benefits of doing so? It's only done in order for Tetragon tests to pass on an arm64 host. If we can merge this directly without squashing what would be the advantages of doing a separate PR?

alright makes sense, thanks!

bpf/include/vmlinux.h
tixxdz: The vmlinux header is dumped from the current running kernel using something like:
`bpftool btf dump file /sys/kernel/btf/vmlinux format c > vmlinux.h` 

Those registers are not the only things that may differ...  IIRC as an example on ARM char is unsigned by default, and on x86 it is not defined (depends on implementation)... etc

libbpf uses later CORE to locate fields etc but that's another story. So we should not mix it here, for each arch we must generate a proper vmlinux file, let's ping @jrfastab on to know which kernel config or kernel stock image they used to generate the current one.

mtardy: Yes yes I see your point but again what's the benefit of using different files and not using the preprocessor macros directly? Tidiness?

mtardy: btw how do you know which structs are accessed using CORE and not? (this one is kinda obvious since it's completely arch-specific but is there a way?). Just to be clear it compiles like that at the moment, so it's a matter of putting things in the right place?

Should I put this part in a separate header file?
```c
#if defined(__TARGET_ARCH_x86)

struct pt_regs {
	long unsigned int r15;
	long unsigned int r14;
	long unsigned int r13;
	long unsigned int r12;
	long unsigned int bp;
	long unsigned int bx;
	long unsigned int r11;
	long unsigned int r10;
	long unsigned int r9;
	long unsigned int r8;
	long unsigned int ax;
	long unsigned int cx;
	long unsigned int dx;
	long unsigned int si;
	long unsigned int di;
	long unsigned int orig_ax;
	long unsigned int ip;
	long unsigned int cs;
	long unsigned int flags;
	long unsigned int sp;
	long unsigned int ss;
};

#elif defined(__TARGET_ARCH_arm64)

struct user_pt_regs {
	__u64 regs[31];
	__u64 sp;
	__u64 pc;
	__u64 pstate;
};

struct pt_regs {
	union {
		struct user_pt_regs user_regs;
		struct {
			u64 regs[31];
			u64 sp;
			u64 pc;
			u64 pstate;
		};
	};
	u64 orig_x0;
	s32 syscallno;
	u32 unused2;
	u64 sdei_ttbr1;
	u64 pmr_save;
	u64 stackframe[2];
	u64 lockdep_hardirqs;
	u64 exit_rcu;
};

#endif
```

jrfastab: CORE accesses in our BPF code use the semantic `_(&xyz)` so you can check the BPF code directly.

I think doing a ifdef like this in a specific header is fine. I don't have anything better off-hand.

bpf/include/vmlinux.h
```diff
@@ -1435,6 +1437,37 @@ struct pt_regs {
 	long unsigned int ss;
 };
 
+#elif defined(__TARGET_ARCH_arm64)
```
kkourt, line 1440:
```suggestion
#elif defined(__TARGET_ARCH_arm64)
// see: https://github.com/cilium/tetragon/issues/786
```

I've tried to summarize the discussion (both here and offline in a comment). Let's use this approach for now, but add a comment that points to the issue so that we can fix it later.

kkourt (changes requested):
Thanks!

bpf/include/vmlinux.h
mtardy: See this: https://github.com/cilium/tetragon/pull/734/commits/c996118abb3d40a73c827705e87bcf448897b973#diff-e7b63fd817e31c67015e2c425624987ad9ef27a734909c92d9cb218b23fc8904R1441-R1443

Thanks!! :)

pkg/sensors/exec/exec_test.go
```diff
@@ -452,7 +452,7 @@ func TestDocker(t *testing.T) {
 	observer.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
 
 	readyWG.Wait()
-	serverDockerID := observer.DockerRun(t, "--name", "fgs-test-server", "--entrypoint", "nc", "quay.io/cilium/alpine-curl:1.0", "-nvlp", "8081", "-s", "0.0.0.0")
+	serverDockerID := observer.DockerRun(t, "--name", "fgs-test-server", "--entrypoint", "nc", "quay.io/cilium/alpine-curl:v1.6.0", "-nvlp", "8081", "-s", "0.0.0.0")
```
kkourt, line 455: Just a question: is this a multi-arch image?

kkourt (approved):

pkg/sensors/exec/exec_test.go
mtardy: Yep that's exactly the reason of the patch, see commit message here https://github.com/cilium/tetragon/pull/734/commits/6cd5873885f24b8ca396ea8a380f4549d4fb933f

PR #738, opened 2023-02-28
Revert "This reverts commit 7f9ac8a."

willfindlay:
b7a3a12553fe543d71e08022014f4908672931fc broke our image builds despite a similar solution working in cilium/cilium. Until we better understand what's going on, let's revert so that we can get our image builds working again.

This reverts commit b7a3a12553fe543d71e08022014f4908672931fc.


mtardy (approved):

mtardy:
It makes me sad 🥲 but we need to

willfindlay:
No need to revert, we have a better solution.

PR #742, opened 2023-03-01
Make functions in pkg/selectors/kernel.go public

tpapagian:
This patch makes selector functions public in order to be used in other (than kprobes and tracepoints) places.

This will allow use to define a different selector with different actions.


kevsecurity (approved):
LGTM

PR #743, opened 2023-03-01
clang image: add support for arm64 and detect arch for BPF compilation

mtardy:
Add multi-arch build for clang image to support both amd64 and arm64,
see https://docs.docker.com/build/ci/github-actions/multi-platform/.

To test, you can try on a platform that has the QEMU emulation thing: `docker buildx build --platform linux/arm64,linux/amd64 --tag mytag/clang . --file=Dockerfile.clang`, it works.

Occasionally adds `.DS_Store` files to .gitignore.

### DIFF
--- a/.github/workflows/build-clang-image.yaml
+++ b/.github/workflows/build-clang-image.yaml
@@ -23,6 +23,13 @@ jobs:
     runs-on: ubuntu-20.04
 
     steps:
+      # https://github.com/docker/setup-qemu-action
+      - name: Set up QEMU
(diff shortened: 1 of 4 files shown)


kkourt (approved):

mtardy:
Apart from the current failure because of SBOM, this PR will fail if we merge because we need the env to get the quay credential.
- we can modify the env to remove the approval requirement
- we can pass the quay credential in the general env

olsajiri (approved):

willfindlay (approved):

willfindlay:
:shipit:

PR #749, opened 2023-03-02
namespaced tracing policies: core implementation

kkourt:
Implementation of namespaced tracing policies.
 * Add namespaced tracing policies (both  simple yaml files and k8s CRDs)
 * refactor things so that we can handle them
 * introduce PolicyHandler which does what SpecHandler does but in a way that allows us to support policy filtering
 * CLI changes (list tracing policies)
 * more refactoring / changes
 * add pod event handlers for keep the policyfilter up-to-date

The last patch is a flag to enable the whole policyfilter functionality. It's off by default for now. Please read individual patches.

Part of:
* https://github.com/cilium/tetragon/pull/694

Depends on:
* https://github.com/cilium/tetragon/pull/748


kkourt:
(rebased due to merging conflicts)

pkg/sensors/manager.go
```diff
@@ -214,6 +219,19 @@ func (h *Manager) DelTracingPolicy(ctx context.Context, name string) error {
 	return err
 }
 
+// ListTracingPolicies retruns a list of the active tracing policies
```
tpapagian, line 222: typo: returns

pkg/cgroups/fsscan/fsscan.go
```diff
@@ -0,0 +1,199 @@
...
+	containerDir := findContainerDirectory(podPath, containerID)
+	if containerDir == "" {
+		return "", fmt.Errorf("found pod dir=%s but failed to find container for id=%s", podPath, containerID)
+	}
+	return filepath.Join(podPath, containerDir), nil
+}
+
+func podDirMatcher(podID uuid.UUID) func(p string) bool {
+	s1 := podID.String()
+	s2 := strings.Replace(s1, "-", "_", 4)
```
tpapagian, line 120: Why do you use 4 here?

tpapagian (commented):
Generally, it seems good to me. I have some comments/questions but nothing very important. 

Just a high-level question on how to use that.

I understand one way is to use:
```
kind: TracingPolicyNamespaced
metadata:
  name: "example"
  namespace: "NAMESPACE_TO_FILTER"
spec:
```

Can we also use that as:
```
kind: TracingPolicyNamespaced
metadata:
  name: "example"
spec:
```
and call `kubectl apply -n NAMESPACE_TO_FILTER ...`?

pkg/kernels/kernels.go
```diff
@@ -151,3 +151,14 @@ func GenericKprobeObjs() (string, string) {
 		return "bpf_generic_kprobe.o", "bpf_generic_retkprobe.o"
 	}
 }
+
+// ExecObj returns the exec object based on the kernel version
+func ExecObj() string {
```
olsajiri, line 156: nice, I would probably keep this one in the base.go file, but feel free to ignore ;-)

pkg/sensors/manager.go
```diff
@@ -47,6 +48,8 @@ func StartSensorManager(bpfDir, mapDir, ciliumDir string) (*Manager, error) {
 				err = handler.addTracingPolicy(op)
 			case *tracingPolicyDel:
 				err = handler.delTracingPolicy(op)
+			case *tracingPolicyList:
+				err = handler.listTracingPolicies(op)
```
olsajiri, line 52: note `tetra tracingpolicy list` returns me just policies that I added with 'tetra tracingpolicy add ...'
not the one specified with --config-file option. Not sure it's intended

pkg/config/yaml.go
```diff
@@ -92,3 +94,13 @@ func (cnf *GenericTracingConfNamespaced) TpSpec() *v1alpha1.TracingPolicySpec {
...
 	return cnf.Metadata.Name
 }
+
+func NamespacedPolicyFromYaml(data string) (tracingpolicy.TracingPolicy, error) {
+	var k GenericTracingConfNamespaced
+
+	err := yaml.UnmarshalStrict([]byte(data), &k)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
willfindlay: Yes I think we do. Otherwise we can have a NamespacedPolicy with the wrong naming convention here, as we disregard any prior checks.

pkg/sensors/sensors.go
```diff
@@ -83,6 +85,12 @@ func SensorBuilder(name string, p []*program.Program, m []*program.Map) *Sensor
 	}
 }
 
+type policyHandler interface {
+	// PolicyHandler returns a Sensor for a given policy
+	PolicyHandler(policy tracingpolicy.TracingPolicy, filterID policyfilter.PolicyID) (*Sensor, error)
+}
+
+// NB: deprecated, please use policyHandler
```
willfindlay, line 93: Is there a reason we have to deprecate this instead of just getting rid of it?

pkg/sensors/sensors.go
```diff
@@ -92,13 +100,16 @@ type probeLoader interface {
...
+	// list of registered policy handlers, see RegisterPolicyHandlerAtInit()
+	registeredPolicyHandlers = map[string]policyHandler{}
+	// list of registered Tracing handlers, see RegisterSpecHandlerAtInit()
 	registeredSpecHandlers = map[string]specHandler{}
 	// list of registers loaders, see registerProbeType()
 	registeredProbeLoad = map[string]probeLoader{}
 )
 
 // RegisterSpecHandlerAtInit registers a handler for Tracing policy.
+// NB: This is deprecated, please use RegisterPolicyHandlerAtInit for new code.
```
willfindlay, line 112: Same question as above.

willfindlay (approved):
Core changes LGTM. Just left a couple questions/comments.

api/v1/tetragon/sensors.proto
```diff
@@ -38,6 +38,26 @@ message ListSensorsResponse {
...
+	// id is the id of the policy
+	uint64 id = 1;
+	// name is the name of the policy
+	string name = 2;
+	// namespace is the namespace of the policy (or empty of the policy is global)
+	string namespace = 3;
+	// info is additional information about the policy
+	string info = 4;
+	// sensors loaded in the scope of this policy
+	repeated string sensors = 5;
```
tixxdz, line 53: I guess we have a consistent way on how these sensors are named right?

api/v1/tetragon/sensors.proto
```diff
@@ -38,6 +38,26 @@ message ListSensorsResponse {
...
+
+}
+
+message ListTracingPoliciesResponse {
+	repeated TracingPolicyStatus policies = 1;
+}
```
tixxdz, line 59: I know we don't support this now but do you see a case where a tracing policy is paused (loaded but inactive)? and if so I guess this can easily be added in future without breaking API?

pkg/config/yaml.go
```diff
@@ -74,3 +74,31 @@ func FileConfigSpec(fileName string) (*v1alpha1.TracingPolicySpec, error) {
 func FileConfigYaml(fileName string) (*GenericTracingConf, error) {
 	return fileConfig(fileName)
 }
+
+type MetadataNamespaced struct {
+	Name      string `yaml:"name"`
+	Namespace string `yaml:"namespace"`
```
tixxdz, line 72: Is this namespace field has to be explicitly set in the namespaced tracing policy?  or are we able to populate it from something like `kubectl command -n namespace` ?

pkg/sensors/handler.go
```diff
@@ -6,33 +6,70 @@ package sensors
...
+		pfState:     pfState,
+		// NB: policy ids start with 1, so that they can be used for filtering. In policy
+		// filtering code, a policy id of 0 means no filtering.
+		nextPolicyID: 1,
+	}, nil
+}
+
+func (h *handler) allocPolicyID() uint64 {
+	ret := h.nextPolicyID
+	h.nextPolicyID++
```
tixxdz, line 44: What about you make it more explicit , in newHandler() you do:

```
return &handler{ nextPolicyID: ZERO, }
```

and in allocPolicyID() you increment then return the usual way... per this logic I suppose you have to call allocPolicyID() anyway.

Then you may even convert that to `atomic.AddUint64() ` to be robust

pkg/sensors/handler.go
```diff
@@ -6,33 +6,70 @@ package sensors
...
+
+	// This is a namespaced policy, so update policy filter state before loading the sensors
+	// NB: the filterID is set to a non-zero value only if we have a namespaced policy and need
+	// to apply filtering.
+	if tpNs, ok := op.tp.(tracingpolicy.TracingPolicyNamespaced); ok {
+		filterID = policyfilter.PolicyID(tpID)
+		if err := h.pfState.AddPolicy(filterID, tpNs.TpNamespace()); err != nil {
+			return err
+		}
+	}
```
tixxdz, line 79: nit: what if instead of presetting `filterID := policyfilter.PolicyID(0)` , you do usual if else path and set the `filterID := policyfilter.PolicyID(0)` only on the else?

And replace that 0 with a constant something like TracingPolicyNotNamespaced so easy to read and we are not lost with the special zeros?

tixxdz (commented):
Quick round, will continue tomorrow

pkg/sensors/manager.go
tixxdz: Also in this context addTracingPolicy() is incrementing a policy ID but what about --config-file policy ? that should also be accounted?

On another note just to ensure I'm not mixing things up, this is not the ID that should identify a policy in json events right? I remember there was talk about it... that one should be a non recyclable ID or uuid...

tpapagian (approved):

pkg/config/yaml.go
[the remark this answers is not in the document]
kkourt: Nice catch, thanks! fixed.

pkg/cgroups/fsscan/fsscan.go
kkourt: s1 is "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx"
so there are only 4 "-". I've added a comment.

pkg/kernels/kernels.go
kkourt: makes sense. moved it to base.go :)

pkg/sensors/manager.go
kkourt: So the problem here is that we don't use the sensor manager to load the policy provided by the config file. We should fix this, so I made an issue: https://github.com/cilium/tetragon/issues/784.

Note that the same is true for the "base" sensor, but I think that's fine.

pkg/sensors/sensors.go
kkourt: Just to reduce the number of patches. Plan is to move them all. Made an issue here: https://github.com/cilium/tetragon/issues/785

pkg/sensors/sensors.go
kkourt: same answer :)

api/v1/tetragon/sensors.proto
kkourt: Each sensor has its name, yes.

api/v1/tetragon/sensors.proto
kkourt: I don't see why we would need to break the API. We can just add a status field.

pkg/config/yaml.go
kkourt: Both.

pkg/sensors/handler.go
kkourt: What is the problem with the current approach?

pkg/sensors/handler.go
kkourt: > nit: what if instead of presetting filterID := policyfilter.PolicyID(0) , you do usual if else path and set the filterID := policyfilter.PolicyID(0) only on the else?

I don't think this makes a difference. There is no uninitiated memory in go.
If I write:
```
var filterID policyfilter.PolicyID
```

It will initialize this to the default value which is zero.

kkourt:
> Generally, it seems good to me. I have some comments/questions but nothing very important.
> 
> Just a high-level question on how to use that.
> 
> I understand one way is to use:
> 
> ```
> kind: TracingPolicyNamespaced
> metadata:
>   name: "example"
>   namespace: "NAMESPACE_TO_FILTER"
> spec:
> ```
> 
> Can we also use that as:
> 
> ```
> kind: TracingPolicyNamespaced
> metadata:
>   name: "example"
> spec:
> ```
> 
> and call `kubectl apply -n NAMESPACE_TO_FILTER ...`?

Yes. You can also omit both namespaces and the "default" namespace will be used. I also believe that if you use conflicting namespaces applying the policy will fail.

kkourt:
Thanks all! I've pushed a new version with all the actionable comments addressed! (I think)

cmd/tetragon/main.go
```diff
@@ -629,6 +615,10 @@ func execute() error {
 	// disable.
 	flags.Bool(keyReleasePinnedBPF, true, "Release all pinned BPF programs and maps in Tetragon BPF directory. Enabled by default. Set to false to disable")
 
+	// Provide option to enable policy filtering. Because the code is new,
+	// this is set to false by default.
+	flags.Bool(keyEnablePolicyFilter, false, "Enable policy (beta) filter code")
```
tixxdz, line 620: Can we just not expose this? mark it hidden or at least add words here to say this will be turned on by default in future?
Users may start to rely on it being false, then complain later...

pkg/option/config.go
```diff
@@ -43,6 +43,8 @@ type config struct {
 	EventQueueSize uint
 
 	ReleasePinned bool
+
+	EnablePolicyFilter bool
```
tixxdz, line 47: Maybe a comment here to... that this will be removed

pkg/policyfilter/state.go
```diff
@@ -105,28 +110,33 @@ func (pod *podInfo) delContainers(id string) []containerInfo {
...
 	for i := range pod.containers {
 		container := &pod.containers[i]
-		if container.id == containerID && container.cgID == cgID {
-			// container was already handled, return
+		if container.id != containerID {
+			continue
+		}
+
+		// we found a matching container id, if no cgroup id is given, return
+		if cgIDp == nil {
```
tixxdz, line 121: Just to understand under which conditions this could happen not having a cgIDp?

pkg/policyfilter/state.go
```diff
@@ -105,28 +110,33 @@ func (pod *podInfo) delContainers(id string) []containerInfo {
...
+		// otherwise, also check that the cgroup id matches
+		if container.id == containerID && container.cgID == *cgIDp {
+			// container was already handled, return
+			return true
 		}
+
+		// issue  a warning and continue
+		// NB: if this happens, we might end up with multiple cgroup
+		// ids for the same container. Since we have no way of knowing
+		// which one is the correct, we keep both.
```
tixxdz, line 134: Same here under which conditions?

In cgroupv2 it should never happen, in cgroupv1 with multiple hierarchies I think I can see it, but we can definitely track the right hierarchy to reduce such noise

tixxdz (changes requested):
Thank you for the great work!

I think we should ensure the things that are exposed to users are good enough so we don't break later, otherwise yes there are lot of details that we can improve later.

pkg/sensors/sensors.go
```diff
@@ -148,25 +148,11 @@ type LoadProbeArgs struct {
...
-		sensors = append(sensors, sensor)
-	}
-	return sensors, nil
-}
-
-func GetMergedSensorFromParserPolicy(name string, policy interface{}) (*Sensor, error) {
-	sensors, err := GetSensorsFromParserPolicy(policy)
+func GetMergedSensorFromParserPolicy(tp tracingpolicy.TracingPolicy) (*Sensor, error) {
+	// NB: use a filter id of 0, so no filtering will happen
+	sensors, err := SensorsFromPolicy(tp, policyfilter.PolicyID(0))
```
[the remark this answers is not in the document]
kkourt: OK, let's keep it as `0` then.

I actually prefer it than a constant, especially because we are also using this value in bpf side.

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -421,6 +424,7 @@ func (tp *genericTracepoint) EventConfig() (api.EventConfig, error) {
 	}
 
 	config := api.EventConfig{}
+	config.PolicyID = uint32(tp.policyID)
```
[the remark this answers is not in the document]
kkourt: yes, policyID is `policyfilter.PolicyID`.

cmd/tetragon/main.go
kkourt: No I'd like to expose it so that people can try this out.

pkg/option/config.go
kkourt: I'm not sure if we are going to remove it. First, we will default it to `true`.

pkg/policyfilter/state.go
kkourt: It's explained in the commit message that adds the pod event handlers:

> 
>     pkg/policyfilter: add pod event handlers
>     
>     This patch adds pod event handlers for policyfilter.
>     
>     A new function RegisterPodHandlers is exported, which allows callers to
>     register the needed hooks.
>     
>     The hooks monitor pod additions and deletions, and update the
>     policyfilter accordingly. These handlers may work i) in combination with
>     run-time hooks, or ii) in the absence of run-time hooks. Note in case of
>     ii) there is a race.
>     
>     Removing containers is easy, we just remove the state for a given
>     container id.
>     
>     Adding containers is more involved because we do not know the cgroup id
>     to add to the policyfilter bpf map from the k8s API information. If a run-time hook was already
>     executed, our state should be up-to-date, so there is no more work to do.
>     However, there are cases where this might not have happened. One example
>     of this is when the agent starts, and the existing pods are processed.
>     Another example is case ii) as described above.
>     
>     For these cases, we need to find the cgroup id from the container name.
>     To do so, we use cgroups.fsscan. Hence, the function AddPodContainer is
>     modified to take a pointer to a cgroup id. If that pointer is not-nil,
>     everything works as before. Otherwise, the code will try to determine

Hope this explains why we need it.

pkg/policyfilter/state.go
kkourt: > Same here under which conditions?

I'm not aware of any conditions that could cause this.

tixxdz (approved):
Thanks !  still minor comment about the default option so users won't assume anything and we can easily hide it later, up to you!

kkourt:
rebased to resovle conflicts.

Issue #751, opened 2023-03-02
Abstract the arch syscalls symbol prefix (e.g. `__x64` or `__arm64`) for kprobes


mtardy:
Because we use a `syscall: <bool>` field, we could detect when the symbol of a kprobe is a syscall and if it already contains the arch-specific prefix, like:
- `__arm64_sys_ptrace`
- `__x64_sys_ptrace`

On the non-detection of the typical `__x64` or `__arm64` prefixes, we could allow the user to use an alternative `sys_ptrace` directly and inject the correct prefix according to the machine architecture. It would be retro-compatible and would allow writing arch-independent TracingPolicy containing kprobes on syscalls.

PR #752, opened 2023-03-02
kprobe: detect and add missing syscall arch prefix

mtardy:
This fixes https://github.com/cilium/tetragon/issues/751.

If kprobe call is a syscall, we try to detect if the user specified a prefix in the form of "__x64" or "__arm64", and if not, transparently add it based on runtime.GOARCH.

```
// Note that cilium/ebpf, retries to attach to an arch specific version of the
// symbol after a failure thus making possible to attach to "sys_listen" which
// will fail and retry with the corrected "__arm64_sys_listen" if you are
// running on arm64. See https://github.com/cilium/ebpf/pull/304.  However, this
// causes Tetragon users to believe that two symbols exist (following the same
// example) "__arm64_sys_listen" and "sys_listen", which results in differents
// events.  Because we actually know if a call is a syscall, we can pre-format
// the symbol directly here to allow the user to specify the "sys_listen"
// version while having only registering the real symbol "__arm64_sys_listen".
```

~~I would like to add some tests and docs but it's already reviewable.~~

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -14,7 +14,9 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"runtime"
 	"strconv"
+	"strings"
 	"sync/atomic"
 
 	"github.com/cilium/ebpf"
@@ -62,6 +64,8 @@ const (
 
 	MaxKprobesMulti = 100 // MAX_ENTRIES_CONFIG in bpf code
 )
+
+var supportedArchPrefix = map[string]string{"amd64": "__x64_", "arm64": "__arm64_"}
 
 func kprobeCharBufErrorToString(e int32) string {
 	switch e {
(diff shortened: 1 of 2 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -284,6 +288,25 @@ func createGenericKprobeSensor(name string, kprobes []v1alpha1.KProbeSpec) (*sen
...
+					break
+				}
+			}
+			if !alreadyPrefixed {
+				if prefix, found := supportedArchPrefix[runtime.GOARCH]; found {
+					funcName = prefix + "_" + funcName
+				} else {
+					logger.GetLogger().Warnf("kprobe syscall prefix detection, unsupported architecture: %s", runtime.GOARCH)
+				}
+			}
```
kkourt, line 308: Can we add a warning here if the existing prefix does not match the arch?
Would be nice to move that code into a function and write a couple of simple unit tests. 

I think we also need to change btf validation code: https://github.com/cilium/tetragon/blob/main/pkg/btf/validation.go#L128.

kkourt (changes requested):

pkg/sensors/tracing/generickprobe.go
mtardy: - I moved this thing into a function and now it's way better
- For btf validation code, I change `f.Call` var directly so that prefix appending is transparent to this code.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -255,6 +259,35 @@ func createMultiKprobeSensor(sensorPath string, multiIDs, multiRetIDs []idtable.
...
+// events.  Because we actually know if a call is a syscall, we can pre-format
+// the symbol directly here to allow the user to specify the "sys_listen"
+// version while only registering the real symbol "__arm64_sys_listen".
+func addSyscallPrefix(symbol string, arch string) (string, error) {
+	for _, p := range supportedArchPrefix {
+		if strings.HasPrefix(symbol, p) {
+			// check that the prefix found is the correct one
+			if p != supportedArchPrefix[arch] {
+				return "", fmt.Errorf("expecting %s and got %s", supportedArchPrefix[arch], p)
+			}
```
kkourt, line 281: nit: we can get the arch for p from the iterator and check it against `arch`

```
for p_arch, p := range supportedArchPrefix {
     if  strings.HasPrefix(symbol, p) {
         if arch != p_arch {
                    // return error
         }
}
```

kkourt (approved):

pkg/sensors/tracing/generickprobe.go
kkourt: (just a suggestion, feel free to ignore)

mtardy: I see but it would not be the same code I think. `supportedArchPrefix[arch]` is fixed and is the arch of the machine passed by the second function argument. Maybe it's not super clear but on the check:
```go
if p != supportedArchPrefix[arch] {
    return "", fmt.Errorf("expecting %s and got %s", supportedArchPrefix[arch], p)
}
```
I check for the error case when you have for example a x64 prefix on and your static running arch is arm64.

mtardy: yes good idea it would simplify this if statement that was a bit confusing (that's why I added a comment) :)

PR #754, opened 2023-03-03
matchBinaries: Add NotIn in tracing policy

tpapagian:
In https://github.com/cilium/tetragon/pull/686/commits/84e71d2777f43b53fcb8b53caa01431d4640ffa6 we introduced the implementation of NotIn operator in matchBinaries selector. We missed to add this in the tracing policy definition.

This patch fixes that.


kkourt (approved):
Thanks!

mtardy:
It would be good to update the doc as well @tpapagian :p

mtardy:
let's put it in the list of the new website doc

PR #758, opened 2023-03-03
tetragon: retry get pod info from informer

Y-dc:
This patch fix that the pod information is empty in cached process, but dokerid is not empty. This happens when pod information cannot be queried from the informer cache (informer sync interval) while caching processes (via events such as /proc or evecve).


api/v1/tetragon/tetragon.proto
```diff
@@ -95,6 +95,7 @@ message Process {
     uint32 refcnt = 13;
     Capabilities cap = 14;
     Namespaces ns = 15;
+    google.protobuf.UInt32Value ns_pid = 16;
```
tpapagian, line 98: I am not really sure that we need to export `ns_pid` to the user events. Another alternative can be to add a new field in `ProcessInternal` that keeps the `NSPID` value and use this when you need to call `GetPodInfo`.

tpapagian (changes requested):
Thanks! I have a comment on the approach that you used.

Can you also please explain (in the PR message) what is the case that you hit this issue? I am asking because when we generate a new event, we check if we miss the pod info and then use the `eventcache` to wait for the pod info to be available. Additionally, does this happen for both the process and the parent in your test case?

api/v1/tetragon/tetragon.proto
Y-dc: yep, i think that's better

Y-dc:
> 

In my case I disabled eventcache because grpc server is not used

Y-dc:
Supplement: On the one hand, I disabled `eventcache` because I only need prometheus exporter. On the other hand, in my test, I found that eventcache will take up a lot of memory. When there are some processes on the host machine, these processes themselves will not have podinfo, but they will To retry 15 times (CacheStrikes = 15). I think CacheStrikes can be put in the configuration

pkg/process/process.go
```diff
@@ -46,6 +46,7 @@ type ProcessInternal struct {
 	// garbage collector metadata
 	color  int
 	refcnt uint32
+	nsPid  uint32
```
tpapagian, line 49: nit: Can you please move that after `namespaces`? This is more an "additional internal field" rather than "garbage collector metadata".

tpapagian (approved):
Ok, I see. I believe that disabling the `eventcache` will cause more issues other than just missing pod information. i.e. 
1. if you use kprobe/tracepoint events there can be cases where we will miss process (or parent) information entirely.
2. there can be exit events that might miss process information etc. 

On the other hand, if this helps with what you are doing, I am happy to have this merged.

pkg/process/process.go
Y-dc: ok

pkg/process/process.go
```diff
@@ -43,6 +43,7 @@ type ProcessInternal struct {
 	// additional internal fields below
 	capabilities *tetragon.Capabilities
 	namespaces   *tetragon.Namespaces
+	nsPid        uint32
```
tixxdz, line 46: Thanks for the patch, hmm seems we are also calling this nsPid in other places where it should be just Pid!  hmm  will check and get back to you

tixxdz: Ok I documented it here https://github.com/cilium/tetragon/pull/775, thanks!

kkourt:
@tpapagian @tixxdz are you OK with merging this?

tpapagian:
> @tpapagian @tixxdz are you OK with merging this?

Yes, I am fine to merge that.

tixxdz (approved):
Lgtm! sorry for the late reply

kkourt:
@Y-dc could you please rebase the PR so that we can merge it? Thanks!

Issue #571, opened 2022-12-01
go.mod: update to a new cilium package version


kkourt:
We have many depandabot attempts to try and update to a new cilium version, but they fail due to dependency failures.

Doing so manually is also not trivial. If, for example, we do:
```
$ go get github.com/cilium/cilium@1.12.4
```

To update to the latest cilium package, then github.com/cilium/hubble does not work any more. Attempting to update that as well:
```
$ go get github.com/cilium/hubble
```

Leads us to a weird situation where there are many failed dependencies due to a lot of code being removed from cilium/hubble in https://github.com/cilium/hubble/pull/220.

Some of the issues can be dealt by changing `cilium/hubble/pkg` -> `cilium/cilium/pkg/hubble`. E.g., "github.com/cilium/cilium/pkg/hubble/filter". For others, however, I do not see a clear solution.

For example, not clear where hubble's old [cilium.CiliumState](https://github.com/cilium/hubble/blob/3e44df066567d02c2f20ecd6681211bb0bab34db/pkg/cilium/state.go#L57) that is used in teragon's [pkg/cilium](https://github.com/cilium/tetragon/blob/3a3f8990503153b2ea63e15b3a62cbed6f8d1949/pkg/cilium/state.go#L52) can be found if anywhere.

PR #759, opened 2023-03-03
pkg/cilium: retry if socket does not exist

kkourt:
Currently, we call .Fatal if .Dial on the monitor socket fails. Let's just retry to avoid agent restarts in case Cilium is not there.

Edit: As @tixxdz suggested, this might lead to having frequent messages. So we implement simple backoff for those. Some of these messages, however, were in the old hubble code so while we are at it, let's pull it in. This offers a tempo solution for https://github.com/cilium/tetragon/issues/571, until we figure out something better.


tixxdz:
Did you test that? I suspect that this will keeps warn logging, when events are received and we try to resolve the info

kkourt:
> Did you test that? 

I did not test it, no. 

> I suspect that this will keeps warn logging, when events are received and we try to resolve the info

And why is this a problem?

kkourt:
> I suspect that this will keeps warn logging, when events are received and we try to resolve the info
> 
> And why is this a problem?

Figured it out on my own 😅. Seems that the timeout is too short and we end up with lots of messages. PR now implements a simple backoff for the timeout.

cmd/tetra/getevents/io_reader_client.go
```diff
@@ -10,10 +10,10 @@ import (
 	"io"
 	"os"
 
-	v1 "github.com/cilium/hubble/pkg/api/v1"
-	hubbleFilters "github.com/cilium/hubble/pkg/filters"
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/filters"
+	v1 "github.com/cilium/tetragon/pkg/oldhubble/api/v1"
```
tixxdz, line 15: Could we please export it as  hubbleV1 like in other places?

pkg/filters/arguments_regex.go
```diff
@@ -8,9 +8,9 @@ import (
 	"fmt"
 	"regexp"
 
-	v1 "github.com/cilium/hubble/pkg/api/v1"
-	hubbleFilters "github.com/cilium/hubble/pkg/filters"
 	"github.com/cilium/tetragon/api/v1/tetragon"
+	v1 "github.com/cilium/tetragon/pkg/oldhubble/api/v1"
```
tixxdz, line 12: hubbleV1?

pkg/oldhubble/cilium/dns.go
```diff
@@ -35,15 +34,20 @@ type FqdnCache interface {
...
 func (s *State) syncFQDNCache() {
+	t0 := 1 * time.Second
+	t := t0
 	for {
 		entries, err := s.ciliumClient.GetFqdnCache()
 		if err != nil {
 			s.log.WithError(err).Error("Unable to obtain fqdn cache from cilium")
-			time.Sleep(fqdnCacheRetryInterval)
+			time.Sleep(t)
+			t = 2 * t
```
tixxdz, line 47: I know you tested it, but are we sure that the `s or s.ciliumClient ` are not nil?  if so fine, otherwise at least a nil check against s.ciliumClient could be helpful in future...

Also yes this is way better than messing up with state as the time is 64bits + we sleep so safe from any wrapping...

tixxdz (commented):
LGTM, minor comments , feel free to change or merge, thank you ;-)

pkg/oldhubble/cilium/dns.go
kkourt: AFAICT it's always not nil in our current code. Please let me know if you think otherwise. In any case, the nill check seems unrelated to this PR.

cmd/tetra/getevents/io_reader_client.go
kkourt: This patch just replaces:
> 	v1 "github.com/cilium/hubble/pkg/api/v1"

from the previous version.

Added a patch to use hubbleV1.

kkourt:
@tixxdz Added a patch to replace `v1` with `hubbleV1`. PTAL.

pkg/oldhubble/cilium/dns.go
tixxdz: Alright

pkg/filters/eventType.go
```diff
@@ -6,9 +6,9 @@ package filters
 import (
 	"context"
 
-	v1 "github.com/cilium/hubble/pkg/api/v1"
-	hubbleFilters "github.com/cilium/hubble/pkg/filters"
 	"github.com/cilium/tetragon/api/v1/tetragon"
+	v1 "github.com/cilium/tetragon/pkg/oldhubble/api/v1"
```
tixxdz, line 10: seems missed this one.

pkg/filters/filters.go
```diff
@@ -20,10 +20,10 @@ import (
 	"io"
 	"strings"
 
-	v1 "github.com/cilium/hubble/pkg/api/v1"
-	hubbleFilters "github.com/cilium/hubble/pkg/filters"
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/api/v1/tetragon/codegen/helpers"
+	v1 "github.com/cilium/tetragon/pkg/oldhubble/api/v1"
```
tixxdz, line 25: also this one.

pkg/filters/health_check.go
```diff
@@ -18,9 +18,9 @@ import (
 	"context"
 	"path"
 
-	v1 "github.com/cilium/hubble/pkg/api/v1"
-	hubbleFilters "github.com/cilium/hubble/pkg/filters"
 	"github.com/cilium/tetragon/api/v1/tetragon"
+	v1 "github.com/cilium/tetragon/pkg/oldhubble/api/v1"
```
tixxdz, line 22: same

tixxdz (approved):
Seems you missed some but approved anyway (the v1 only does not reflect package...)

kkourt:
> Seems you missed some but approved anyway (the v1 only does not reflect package...)

Thanks, let's do those as followup if we need to. This part was just about moving the dependency inside tetragon.

PR #763, opened 2023-03-04
Export selector ops from pkg/selectors/kernel.go

tpapagian:
This patch makes selector ops public in order to be used in other modules.


tixxdz (approved):
Thanks!

PR #576, opened 2022-12-02
tetragon: char_buf matchArgs works not as expected

Y-dc:
fix #575 .  preferably based on #568 and #564 .
I found the source of the bug in ```filter_char_buf```([source code](https://github.com/cilium/tetragon/blob/9aae81d74cecb66bc4ca802396f0aa074a1e790f/bpf/process/types/basic.h#L552)). the comparison of strings and their lengths is wrong.
form ```__copy_char_buf```([source code](https://github.com/cilium/tetragon/blob/9aae81d74cecb66bc4ca802396f0aa074a1e790f/bpf/process/types/basic.h#L530)) , the target ```arg``` in ```args``` should be offset by 8.
 
image

Signed-off-by: dechengyuan <dechengyuan@tencent.com>

PR #770, opened 2023-03-07
fix char buff matchargs

kkourt:
Got the patch from @Y-dc (see https://github.com/cilium/tetragon/pull/576) and added a test for it.


olsajiri (approved):
nice, I guess the fix goes in through this PR (not https://github.com/cilium/tetragon/pull/576) , because there's test

kkourt:
> nice, I guess the fix goes in through this PR (not #576) , because there's test

I think that's easier, yes. I'll wait until the discussion on #576 concludes and then I'll merge this one.

Issue #750, opened 2023-03-02
Improve build image workflows to dry-run on PR


mtardy:
It consists of:
- Adding a `paths` trigger on the workflow file on the `pull_request` GA event to dry-run on workflow file modification.
- Using a conditional to push only the built image on `push` GA events. 
- Skip `quay.io` login on dry-run. 
- Modify the environment thing in order for PR to dry-run without approval. The env provides currently the approval mechanism and the credentials to log in for the real run.

PR #777, opened 2023-03-08
clang image: rework workflow to dry run on PR

mtardy:
Fixes https://github.com/cilium/tetragon/issues/750 since other workflows `build-image-ci.yml` and `build-images-releases.yml` are already acting differently in the case of push on main and PRs.

- Add a path addition to allow for the workflow to be tested when we modify it in a PR.
- Add a boolean condition in quay login step and at pushing step to only push on commit on main.
- Remove the environment since the workflow can run without approval now. We will still need the credential from the environment to login to quay thought.

**Warning**: I need external action on the repo for this PR to work when merged:
```yaml
          username: ${{ secrets.QUAY_CLANG_RELEASE_USERNAME }}
          password: ${{ secrets.QUAY_CLANG_RELEASE_PASSWORD }}
```
These secrets must be moved from the `environment: release-clang` to the global env if it makes sense.

### DIFF
--- a/.github/workflows/build-clang-image.yaml
+++ b/.github/workflows/build-clang-image.yaml
@@ -1,15 +1,20 @@
 name: Build Clang Image
 
+# we want this action to dry run in case it's triggered from a PR
+# - docker/login-action only logs in to quay on GA 'push' event
+# - docker/build-push-action only pushes to quay on GA 'push' event
 on:
   push:
     branches:
       - main
       - v*
     paths:
-    - 'Dockerfile.clang'
-  pull_request:
+      - 'Dockerfile.clang'
+      - '.github/workflows/build-clang-image.yaml'
+  pull_request_target:
     paths:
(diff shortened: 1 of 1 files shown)


mtardy:
I have an issue because I cannot locally load the multi-arch build. And `bom` need the image to exist to generate the sbom. So I can maybe disable the whole sbom thing for dry-run?

mtardy:
> I have an issue because I cannot locally load the multi-arch build. And `bom` need the image to exist to generate the sbom. So I can maybe disable the whole sbom thing for dry-run?

This is basically because of this long standing issue https://github.com/docker/buildx/issues/59.

willfindlay (approved):
We ended up deciding to keep the release-clang environment but still do a dry-run only on PRs. This has the unfortunate side effect of requiring approval on dry-runs but I think we can live with that.

PR #778, opened 2023-03-08
Add a Tetragon documentation website and rename base folder `crds` into `examples`

mtardy:
This website is not finished, the goal is to merge it "as-is", not advertising it for now, finish the home page and branding and then advertise it!

When merging this, I need some manual steps to be performed on the repository settings to configure the GitHub page.

The main pain point that need to be addressed is that _currently_, this PR breaks the soft links that were made to the existing documentation that moved. Including examples from @sharlns and Jed report.

- [ ] Note for myself, remove the useless 400kb cereal image even if it might be too late :/


xmulligan:
do we have any stats on how often the links are used?

willfindlay (approved):
CI changes LGTM. Do we want to add a protection rule for the `github-pages` environment?

mtardy:
Thanks for looking at the CI changes :)

> Do we want to add a protection rule for the `github-pages` environment?

It shouldn't be needed, just merging = approved = deployed on the website. wdyt?

mtardy:
> do we have any stats on how often the links are used?

I really don't know but the report is still being actively distributed and it would be as easy as replacing the link redirection to permalinks I guess so, it would be nice to do it before moving this stuff.

mtardy:
For now, the easiest decision is to not delete it. So I just left `docs/security-observability-with-ebpf` intact, even if it looks weird.

One downside I overlooked as well is that I am deleting @tixxdz example configuration folder: https://github.com/cilium/tetragon/tree/main/docs/configuration/tetragon.conf.d. Maybe I could move it under `docs/examples/configuration/tetragon.conf.d`?

willfindlay:
We already have `/examples`, can you just move the `tetragon.conf.d` magic into there?

mtardy:
> We already have `/examples`, can we just move the `tetragon.conf.d` magic into there?

@willfindlay do you mean `/crds/examples`? I don't think we have a root level `/examples` folder. Although it would make way more sense to rename `crds` to `examples` and put things in like `configuration`, `tracingpolicy`, etc...

willfindlay:
Yeah sorry that's what I meant.

> Although it would make way more sense to rename crds to examples and put things in like configuration, tracingpolicy, etc...

Yeah I like this as an idea. Especially because they're not really CRDs...

mtardy:
> Yeah sorry that's what I meant.
> 
> > Although it would make way more sense to rename crds to examples and put things in like configuration, tracingpolicy, etc...
> 
> Yeah I like this as an idea. Especially because they're not really CRDs...

+1000, it just confuses everybody and then we are all saying CRD all the time instead of CR or TracingPolicy!

mtardy:
So now this PR does a bit more than intended:
- Add the documentation website.
- Rename the `crds` top folder into `examples`.
- Keep `docs/security-observability-with-ebpf/` in place (to not break printed links in the printed report).

Please look at the individual commits to review, it will be easier, especially for the renaming step.

This PR will break the link to documentation like https://github.com/cilium/tetragon/blob/main/docs/tracingpolicy/README.md, this is intended because we will then redirect people to the properly built website documentation when this thing merges.

PR #795, opened 2023-03-11
tetragon: Add --force-large-progs option to force large bpf programs

olsajiri:
Adding --force-large-progs option to force loading of large bpf
programs.

It's usefull in situations where kernel contains backports and
its version does not reflect actual features.

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -140,6 +140,18 @@ func tetragonExecute() error {
 	log.WithField("version", version.Version).Info("Starting tetragon")
 	log.WithField("config", viper.AllSettings()).Info("config settings")
 
+	if option.Config.ForceLargeProgs && option.Config.ForceSmallProgs {
+		log.Fatalf("Can't specify --force-small-progs and --force-large-progs together")
+	}
+
+	if option.Config.ForceLargeProgs {
+		log.Info("Force loading large programs")
+	}
+
+	if option.Config.ForceSmallProgs {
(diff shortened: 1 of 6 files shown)


tixxdz (approved):
How disruptive is to try to load large programs by default even on old kernels and if it fails fallback to small programs?  or a bpf feature / probe test?  I'm asking in case we can avoid adding a new command line parameter , otherwise fine!

Btw do we have info log messages that can note:
1. Force loading small programs on kernels >= 5.3
2. Force loading large programs on old kernels

Approved but would be great if we can do those ;-)

kkourt (approved):

cmd/tetragon/main.go
```diff
@@ -140,6 +140,14 @@ func tetragonExecute() error {
...
 	log.WithField("config", viper.AllSettings()).Info("config settings")
 
+	if option.Config.ForceLargeProgs {
+		log.Info("Force loading large programs")
+	}
+
+	if option.Config.ForceSmallProgs {
+		log.Info("Force loading smallprograms")
+	}
+
```
kkourt, line 150: Should we maybe check that not both `ForceLargeProgs` and `ForceSmallProgs` is set and bail out if so?

kkourt (changes requested):

cmd/tetragon/main.go
olsajiri: right, good idea, added, thanks

kkourt (approved):
Thanks

PR #796, opened 2023-03-12
tetragon: Multi kprobe fixes

olsajiri:
multi kprobe fixes

### DIFF
--- a/cmd/tetra/yaml/yaml.go
+++ b/cmd/tetra/yaml/yaml.go
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package yaml
+
+import (
+	"fmt"
+
+	"github.com/cilium/tetragon/pkg/btf"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func New() *cobra.Command {
+	yamlCmd := &cobra.Command{
+		Use:   "yaml",
+		Short: "Generate yaml policies",
+	}
+
+	yamlAllSyscallsCmd := &cobra.Command{
+		Use:   "all-syscalls",
+		Short: "Generate all syscalls",
+		Run: func(cmd *cobra.Command, args []string) {
+			allSyscalls(args)
+		},
+	}
+
+	flags := yamlAllSyscallsCmd.Flags()
+	flags.Bool("not", false, "Filter out all syscalls")
+	viper.BindPFlags(flags)
+
+	yamlCmd.AddCommand(yamlAllSyscallsCmd)
+	return yamlCmd
+}
+
+func allSyscalls(args []string) {
+	not := viper.GetBool("not")
+	crd, err := btf.GetSyscallsYaml(not)
+	if err != nil {
+		fmt.Print(err)
+		return
+	}
+	fmt.Printf("%s\n", crd)
+}
--- a/pkg/btf/validation.go
+++ b/pkg/btf/validation.go
@@ -5,6 +5,7 @@ package btf
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/cilium/ebpf/btf"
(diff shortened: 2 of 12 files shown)


cmd/tetra/yaml/yaml.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+
+	"github.com/cilium/tetragon/pkg/btf"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+func New() *cobra.Command {
+	yamlCmd := &cobra.Command{
+		Use:   "yaml",
```
tixxdz, line 16: Hi Jiri, could we consider `tetra tracingpolicy generate $arg` or something similar?

Also @kkourt mentioned about generating tracing policies, so maybe wait a bit for more feedback on the UI and then we decide ;-)

pkg/sensors/load.go
```diff
@@ -203,6 +203,10 @@ func (s *Sensor) loadMaps(stopCtx context.Context, mapDir string) error {
 			return fmt.Errorf("map '%s' not found from '%s'", m.Name, m.Prog.Name)
 		}
 
+		if max, ok := m.Prog.MaxEntriesMap[mapSpec.Name]; ok {
```
tixxdz, line 206: Maybe add "The max entries values needs to be set when map is created and then when program is loaded, so ebpf loader does not complain." ?  from your commit log as code comment too if it is not obvious... in case we try to do same or another developer, would help to not fight cryptic bpf errors... up to you

cmd/tetra/yaml/yaml.go
olsajiri: sure, I don't mind changing the command name, let's see @kkourt opinion on this,
also I'll add doc, np

pkg/sensors/load.go
olsajiri: ok, will add some more comments, thanks

pkg/btf/validation.go
```diff
@@ -261,3 +262,54 @@ func validateSycall(kspec *v1alpha1.KProbeSpec, name string) error {
...
+  name: "syscalls"
+spec:
+  kprobes:`
+
+	filter := `
+    selectors:
+    - matchBinaries:
+      - operator: "In"
+        values:
+        - "/krava"`
```
kkourt, line 279: Can we add a flag for this as well?

E.g., 

```
tetra tracingpolicy generate all-syscalls --match-binaries /krava
```

pkg/btf/validation.go
```diff
@@ -261,3 +262,54 @@ func validateSycall(kspec *v1alpha1.KProbeSpec, name string) error {
...
+
+	bspec, err := btf.LoadSpec(btfFile)
+	if err != nil {
+		return "", fmt.Errorf("BTF load failed: %v", err)
+	}
+
+	for _, key := range syscallinfo.SyscallsNames() {
+		var fn *btf.Func
+
+		key = "__x64_" + key
```
kkourt, line 299: I think we don't need the `__x64` prefix anymore, and omitting it should make things more portable wrt to ARM. CC: @mtardy.

contrib/bench/build-kernel.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+pushd /home/jolsa/bpf-next-1/tools/perf/
```
kkourt, line 3: This feels too specific to your setup, I think :)
How about adding an argument or an env var for the kernel dir?

kkourt (changes requested):
This looks great, thanks!
I just have some minor changes requests.

pkg/btf/validation.go
olsajiri: ok, will add

pkg/btf/validation.go
olsajiri: right, will change that

contrib/bench/build-kernel.sh
olsajiri: so the idea of this script is to be provided by user and executed as workload with:
  # ./contrib/bench/bench.sh <workload>

I just wanted to have it in, so I don't need to save it elsewhere,
I can put to comment that it's user specific

pkg/btf/validation.go
mtardy: yep exactly, please don't use :) thanks kkourt

contrib/bench/build-kernel.sh
kkourt: OK, sounds good to me, thanks!
One suggestion would be to just replace /home/jolsa with $HOME to make it somewhat less user-specific.

kkourt (approved):
The most important bits have been addresses, so LGTM!

tixxdz (approved):

PR #805, opened 2023-03-15
Documentation: add a links checker to the CI

mtardy:
Please appreciate how painful it was to make that work like a charm:
```yaml
      # to avoid automated spam, try to find an existing open issue before opening a new one
      - name: Search for existing issue number
        id: search-issue
        run: |
          encoded_issue_name=$(echo "$ISSUE_NAME" | sed 's/ /%20/g')
          response=$(curl -s -X GET "https://api.github.com/search/issues?q=$encoded_issue_name+repo:cilium/tetragon+state:open+label:automated-issue&type=Issues")
          issue_number=$(echo "$response" | jq -r '.items[0].number // empty')
          echo "issue_number=$issue_number" >> $GITHUB_OUTPUT
          
      - name: Create or update issue with report
        if: env.lychee_exit_code != 0
        uses: peter-evans/create-issue-from-file@433e51abf769039ee20ba1293a088ca19d573b7f
        with:
          title: ${{ env.ISSUE_NAME }}
          content-filepath: ./lychee/out.md
          issue-number: ${{ steps.search-issue.outputs.issue_number }}
          labels: automated-issue
```

### DIFF
--- a/.github/workflows/check-links-cron.yaml
+++ b/.github/workflows/check-links-cron.yaml
@@ -1,0 +1,53 @@
+name: Periodically check docs links
+
+on:
+  schedule:
+    - cron: '0 10 * * 1-5'
+  workflow_dispatch:
+  pull_request:
+    paths:
+      - '.github/workflows/check-links-cron.yaml'
+
+jobs:
+  links-checker:
+    env:
+      ISSUE_NAME: 'Documentation: broken links automatic report'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
+        with:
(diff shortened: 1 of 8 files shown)


mtardy:
Please appreciate the complexity of the workflows!
- The PR workflow checks all the links from the raw `docs/content`, it needs the hugo dev server to start to be able to check for relative links that refer to internal documents. So the workflow starts that thing in the background, waiting for it to be ready in order to check for the links. Otherwise, we could not check if relative links within a new PR would be valid!
- The cron workflow verifies that an open issue exists before creating a new one and updates it in order to avoid spam!

willfindlay (approved):
Looks great, amazing value

mtardy:
I added a step that might close the issue if the periodic check no longer finds failures. (For example in our case with the cilium.herokuapp that could be externally fixed).

willfindlay (approved):

PR #812, opened 2023-03-16
helm: recreate daemonset pod when configmap changes

cjtim:
After the helm upgrade with new configmap changes. the daemonset pod is not recreated to retrieve the new configmap.

The readconfig function only read on pod start https://github.com/cilium/tetragon/blob/main/cmd/tetragon/main.go#L542

to make Tetragon restart and retrieve the new configuration. Kubernetes annotation helps determine whenever the pod needs to be recreated.

### DIFF
--- a/install/kubernetes/templates/daemonset.yaml
+++ b/install/kubernetes/templates/daemonset.yaml
@@ -23,8 +23,9 @@ spec:
     {{- end }}
   template:
     metadata:
-    {{- if .Values.podAnnotations }}
       annotations:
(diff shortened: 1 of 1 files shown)


kkourt (approved):
I think that makes sense. Would be nice if @michi-covalent had a look as well.

Looking into this, we could also set the config map as immutable but that's orthogonal to this patch.

Thanks!

jrfastab:
LGTM thanks.

PR #816, opened 2023-03-16
Cross-compile arm64 container images

mtardy:
This is part of https://github.com/cilium/tetragon/issues/487.

Very fun fact that I learned today, `tetra` requires CGO because it imports `tetragon/pkg/bpf`. I hope you are laughing now!

PR #819, opened 2023-03-17
Workflow: fix syntax of build image CI

mtardy:
I accidentally added a run line here:

https://github.com/cilium/tetragon/blob/540a7bd4d5bd2d3df306a1b0244406efbaf249d5/.github/workflows/build-images-ci.yml#L175-L184

And it broke the workflow (which not triggered an alert in the checks since it was a `pull_request_target`) https://github.com/cilium/tetragon/pull/816#issuecomment-1472545789 it was in plain sight here. Anyway, here's a patch.


mtardy:
E2E is failing because images could not be produced (requiring this fix)

PR #820, opened 2023-03-17
use binary search to optimize performance

Lan-ce-lot:
Use binary search to optimize performance in pkg/ksyms/ksyms.go

### DIFF
--- a/pkg/ksyms/ksyms_test.go
+++ b/pkg/ksyms/ksyms_test.go
@@ -1,0 +1,88 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+package ksyms
+
+import (
+	"testing"
+)
+
+func TestGetFnOffset(t *testing.T) {
+	ksyms := &Ksyms{
+		table: []ksym{
+			{addr: 0x100, name: "addr1", ty: "t"},
+			{addr: 0x200, name: "addr2", ty: "t"},
+			{addr: 0x300, name: "addr3", ty: "w"},
+			{addr: 0x400, name: "addr4", ty: "t"},
+			{addr: 0x500, name: "addr5", ty: "t"},
+			{addr: 0x600, name: "addr6", ty: "t"},
+			{addr: 0x700, name: "addr7", ty: "t"},
+			{addr: 0x800, name: "addr8", ty: "t"},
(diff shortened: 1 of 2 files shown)


kkourt (changes requested):
Thanks!

Since binary search can be tricky to implement in some cases (e.g., https://ai.googleblog.com/2006/06/extra-extra-read-all-about-it-nearly.html), can we please add some unit tests for this?

Lan-ce-lot:
Ok, I will add test.

Lan-ce-lot:
I added some tests, the previous code and the current code have passed. And I prevents overflow at the same time.
```
m := l + ((r - l + 1) >> 1)
```
PTAL, thanks! @kkourt

pkg/ksyms/ksyms_test.go
```diff
@@ -0,0 +1,83 @@
...
+		},
+		{
+			name: "addr 0x900",
+			addr: 0x900,
+			want: FnOffset{SymName: "addr8", Offset: 0x100},
+		},
+		{
+			name:    "address is before first symbol",
+			addr:    0x100,
+			wantErr: true,
```
kkourt, line 59: Here, I believe we need to return the  symbol with an offset of zero.

kkourt (changes requested):
Thanks!

I believe that the correct behavior if a user requests the address of a symbol is to return this symbol with an offsset of zero.

Could you also add a Signed-off-by with a real name? See: https://docs.cilium.io/en/v1.13/contributing/development/contributing_guide/#dev-coo

pkg/ksyms/ksyms_test.go
Lan-ce-lot: Ok, I have modified.

Lan-ce-lot:
> Thanks!
> 
> I believe that the correct behavior if a user requests the address of a symbol is to return this symbol with an offset of zero.
> 
> Could you also add a Signed-off-by with a real name? See: https://docs.cilium.io/en/v1.13/contributing/development/contributing_guide/#dev-coo

Sure!

Lan-ce-lot:
I added a Signed-off-by with a real name.
PTAL, Thanks! @kkourt

kkourt (approved):
LGTM, thanks!

Issue #705, opened 2023-02-21
Privileged execution use case: use nginx instead of test-pod image


mtardy:
In the documentation on the privileged execution use case:

https://github.com/cilium/tetragon/blob/f4d6ac470658375109bcb37cfbe679b5432cdf6d/README.md?plain=1#L594-L609

We could use a more realistic example, using an nginx image instead of the test-pod made-up image.

> Let's do it. I'd just use a simple nginx pod spec for this, e.g:
> ```
> apiVersion: v1
> kind: Pod
> metadata:
>   name: privileged-the-pod
> spec:
>   hostPID: true
>   hostNetwork: true
>   containers:
>   - name: privileged-the-pod
>     image: nginx:latest
>     ports:
>     - containerPort: 80
>     securityContext:
>       privileged: true
> ```
> 
_Originally posted by @sharlns in https://github.com/cilium/tetragon/pull/673#discussion_r1112219567_

janvi01:
Hey @mtardy I would like to work on this.

PR #822, opened 2023-03-19
use nginx instead of test-pod image

zhy76:
Fixs: #705

### DIFF
--- a/testdata/specs/testpod.yaml
+++ b/testdata/specs/testpod.yaml
@@ -7,7 +7,8 @@ spec:
   hostNetwork: true
   containers:
   - name: test-pod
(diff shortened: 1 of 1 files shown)


mtardy (commented):
Thanks for taking the time to do this, the goal of the issue is to adapt the documentation and the tutorial with the capabilities use case. We will need to update both things not just the manifest.

kkourt:
> > Thanks for taking the time to do this, the goal of the issue is to adapt the documentation and the tutorial with the capabilities use case. We will need to update both things not just the manifest.
> 
> Thanks for your reply :)! I'll update the documentation and the tutorial both.

Thanks! marking this as draft for now. Please undraft when you feel the PR is ready.

Issue #761, opened 2023-03-03
Logger is not displaying wrapped errors correctly


mtardy:
When using `%w` to wrap errors like this:
```go
logger.GetLogger().Warnf("kprobe syscall prefix: %w", err)
```
Logs look like this:
```text
time="2023-03-03T16:01:04Z" level=warning msg="kprobe syscall prefix: %!w(*errors.errorString=&{expecting __arm64 and got __x64})
```

mtardy:
I think it's been fixed with https://github.com/cilium/tetragon/pull/3814

PR #823, opened 2023-03-19
Logger is not displaying wrapped errors correctly

zhy76:
fix: Logger is not displaying wrapped errors correctly
Fixes: #761 
Signed-off-by: Haiyu Zuo <958474674@qq.com>

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -343,6 +343,17 @@ func createGenericKprobeSensor(name string, kprobes []v1alpha1.KProbeSpec, polic
 		config.PolicyID = uint32(policyID)
 
 		argRetprobe = nil // holds pointer to arg for return handler
+
+		// modifying f.Call directly instead of writing to funcName
+		// because of BTF validation later using the whole v1alpha1.KProbeSpec object
+		if f.Syscall {
+			prefixedName, err := arch.AddSyscallPrefix(f.Call)
+			if err != nil {
+				logger.GetLogger().Warnf("kprobe syscall prefix: %v", err)
+			} else {
+				f.Call = prefixedName
+			}
+		}
 		funcName := f.Call
 
 		// Parse Arguments


mtardy (changes requested):
Hey, thanks. Maybe it was not clear but the goal of the issue is to see how we can fix the logger library to make it work with wrapped errors instead of just not using them.

zhy76:
> Hey, thanks. Maybe it was not clear but the goal of the issue is to see how we can fix the logger library to make it work with wrapped errors instead of just not using them.

Thanks for your reply! 😃After my research, I found the issue is caused by logrus. The logger under tetragon uses logrus and the func of `Warnf` will finally use `Logf `which uses `fmt.Sprintf`:
image
But fmt.Sprintf does not support error-wrapping directive %w
image
Maybe we can switch to a logger library that supports the %w verb, or manually unwrap the error and log the error message.For example:
```go
if err != nil {
    if unwrappedErr := errors.Unwrap(err); unwrappedErr != nil {
        logger.GetLogger().Warnf("kprobe syscall prefix: %s", unwrappedErr.Error())
    } else {
        logger.GetLogger().Warnf("kprobe syscall prefix: %s", err.Error())
    }
}

kkourt:
fwiw, I think this patch makes sense, even if it does not fix the above issue. How about we remove the "Fix: " part from the PR/patch and then merge it?

zhy76:
> Thanks!
> 
> fwiw, I think this patch makes sense, even if it does not fix the above issue. How about we remove the "Fix: " part from the PR/patch and then merge it?

Do you mean only change %w to %s, or use manually unwrap the error？

kkourt:
> > Thanks!
> > fwiw, I think this patch makes sense, even if it does not fix the above issue. How about we remove the "Fix: " part from the PR/patch and then merge it?
> 
> Thanks for reply！Did you mean only change %w to %v, or use manually unwrap the error？

Chagne `%w` to `%v`.

zhy76:
OK，already fixed.

zhy76:
PTAL,Thanks @kkourt

mtardy (commented):
This piece of code has been moved recently 

https://github.com/cilium/tetragon/blame/ecb77f97885efebd45872177bdaf969110d02dcf/pkg/sensors/tracing/generickprobe.go#L272-L282

jrfastab:
@zhy76 please take a look and see if the improvement is needed on the new code. Thanks!

zhy76:
> What's the status of this PR? It looks like events have superseded it. For the future I think WithError(err) is probably the best way to include an error message in a logged event (unless anyone knows better?).

I agree with you !

kevsecurity:
Okay, do you want to close the PR?

zhy76:
> Okay, do you want to close the PR?

OK

PR #827, opened 2023-03-20
matchArgs improvements

tpapagian:
1. Support multiple filters in matchArgs
2. Fix path argument type


pkg/selectors/kernel.go
```diff
@@ -395,9 +395,17 @@ func ParseMatchArg(k *KernelSelectorState, arg *v1alpha1.ArgSelector, sig []v1al
...
 func ParseMatchArgs(k *KernelSelectorState, args []v1alpha1.ArgSelector, sig []v1alpha1.KProbeArg) error {
+	actionOffset := GetCurrentOffset(k)
 	loff := AdvanceSelectorLength(k)
-	for _, a := range args {
+	argOff := make([]uint32, 5)
+	for i := 0; i < 5; i++ {
+		argOff[i] = AdvanceSelectorLength(k)
+		WriteSelectorOffsetUint32(k, argOff[i], 0)
+	}
+	for i, a := range args {
```
kkourt, line 410: Can we return an error if user species more than 5 arguments?

kkourt (approved):
LGTM, but let's return an error if user specifies more than the supported number of arguments.

pkg/selectors/kernel.go
tpapagian: Done!

jrfastab:
new PR here https://github.com/cilium/tetragon/pull/901

PR #829, opened 2023-03-22
ignore .idea

zhy76:
ignore .idea which is for IntellJ files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ contrib/tester-progs/sigkill-unprivileged-user-ns-tester
 
 # project tool artifacts
 .vagrant/
+.idea
 .vscode
 cscope.files
 cscope.po.out


zhy76:
> Hi!
> 
> Thanks for the patch!
> 
> From a quick search it seems that the `.idea` folder is for IntelliJ specific files. Since we already have a `.vscode` file there, the change makes sense to me.
> 
> Could you please:
> 
> * add one sentence in the commit message saying that this is for IntellJ files
> * add a `Signed-off-by` with a real name? See: https://docs.cilium.io/en/v1.13/contributing/development/contributing_guide/#dev-coo
> 
> Thanks!

Sure!

zhy76:
PTAL, Thanks! @kkourt

kkourt (approved):
Thanks!

kkourt:
(this is a simple change, no need to run tests)

PR #833, opened 2023-03-24
tetragon: Move matchBinaries filter to be executed earlier

olsajiri:
The matchBinary filter des not have any selector record, so we can do that as a first thing in probe.

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -15,6 +15,7 @@
 #include "user_namespace.h"
 #include "capabilities.h"
 #include "../argfilter_maps.h"
+#include "common.h"
 
 /* Type IDs form API with user space generickprobe.go */
 enum {
@@ -481,8 +482,8 @@ copy_capability(char *args, unsigned long arg)
 	return sizeof(struct capability_info_type);
 }
 
-#define ARGM_INDEX_MASK	 ((1 << 4) - 1)
-#define ARGM_RETURN_COPY (1 << 4)
+#define ARGM_RETURN_COPY BIT(4)
+#define ARGM_INDEX_MASK	 ARGM_RETURN_COPY - 1
 
 static inline __attribute__((always_inline)) bool
 hasReturnCopy(unsigned long argm)
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -93,6 +93,8 @@ struct selector_arg_filter {
 	__u8 value;
 } __attribute__((packed));
 
+#define FLAGS_EARLY_FILTER BIT(0)
+
 struct event_config {
 	__u32 func_id;
 	__s32 arg0;
@@ -119,6 +121,7 @@ struct event_config {
 	 * that the hook always applies and no check will be performed.
 	 */
 	__u32 policy_id;
+	__u32 flags;
 } __attribute__((packed));
 
 #define MAX_ARGS_SIZE	 80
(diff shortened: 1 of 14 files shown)


olsajiri:
PR https://github.com/cilium/tetragon/pull/774 gets in the way, moving back to draft ;-)

kkourt:
> PR #774 gets in the way, moving back to draft ;-)

#774 was just merged :)

olsajiri:
checkpatch fails on BIT macro warning, it's not a problem

kkourt:
There seem to be some 4.19 failures:
❌	pkg.sensors.tracing.TestLoadKprobeSensor			10.127s	(7m16.004s)
...
❌	pkg.sensors.tracing.TestLoadTracepointSensor			10.949s	(8m17.18s)
❌	pkg.sensors.tracing.TestUprobeLoad				9.418s	(8m26.598s)

Are these caused by this patch?

olsajiri:
> There seem to be some 4.19 failures: x pkg.sensors.tracing.TestLoadKprobeSensor 10.127s (7m16.004s) ... x pkg.sensors.tracing.TestLoadTracepointSensor 10.949s (8m17.18s) x pkg.sensors.tracing.TestUprobeLoad 9.418s (8m26.598s)
> 
> Are these caused by this patch?

yea, I just had to rebase, will fix that

bpf/process/types/basic.h
```diff
@@ -1337,13 +1351,17 @@ filter_read_arg(void *ctx, int index, struct bpf_map_def *heap,
...
 	size_t total;
 
 	e = map_lookup_elem(heap, &zero);
 	if (!e)
 		return 0;
-	pass = filter_args(e, index, filter);
+	config = map_lookup_elem(config_map, &e->idx);
+	if (!config)
+		return 0;
+	pass = filter_args(e, index, filter, config->flags & 1);
```
tpapagian, line 1364: Maybe we can do `config->flags & FLAGS_EARLY_FILTER` instead of `config->flags & 1` here in order to make it easier to understand.

bpf/process/types/basic.h
```diff
@@ -1009,7 +1012,17 @@ static inline __attribute__((always_inline)) int match_binaries(void *sel_names,
 }
 
 static inline __attribute__((always_inline)) int
-selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx)
+generic_process_filter_binary(struct event_config *config)
+{
+	/* single flag bit at the moment (FLAGS_EARLY_FILTER) */
+	if (!config->flags)
```
tpapagian, line 1018: As I believe that we will need `flags` later as well (and add more `FLAGS_*`), why not `(!(config->flags & FLAGS_EARLY_FILTER))` instead of `(!config->flags)`? In that case we can also remove the comment as well.

pkg/selectors/kernel.go
```diff
@@ -769,3 +769,23 @@ func HasOverride(spec *v1alpha1.KProbeSpec) bool {
...
+		if len(s.MatchPIDs) > 0 ||
+			len(s.MatchNamespaces) > 0 ||
+			len(s.MatchCapabilities) > 0 ||
+			len(s.MatchNamespaceChanges) > 0 ||
+			len(s.MatchCapabilityChanges) > 0 ||
+			len(s.MatchArgs) > 0 {
+			return false
+		}
+	}
+	if len(selectors) == 1 && len(selectors[0].MatchBinaries) > 0 {
```
tpapagian, line 787: As I understand here, we only apply this optimization in the case where we have a single selector that contains MatchBinaries. Why we cannot use that in the case where we have more than one selector? Maybe add a comment here to explain that.

bpf/process/types/basic.h
olsajiri: yea, it was a shortcut because flags is now 0 or 1, but I'll change it to proper check

pkg/selectors/kernel.go
olsajiri: because I need the selector number to do that, so the bpf code will call:
`  match_binaries(&sel_names_map, /* selector index */0);`
I'd need to pass selector indexes somehow..

PR #839, opened 2023-03-27
Makefile: add targets to build the docs via Docker

mtardy:
Add a Dockerfile to package hugo+extended and a Makefile to run `hugo server` easily from the root folder and /docs folder.

Also vendor dependency to avoid fetching the theme each time we build with Docker and also make sure we have the dependencies safe.


willfindlay (approved):

mtardy:
I tested it on amd64 with a native Docker install and it did not work for obscure Docker builds and permission reasons, I added a fix so that it should work for everyone!

mtardy:
The remaining checks are unrelated to the changes of this PR, merging! 🚢

PR #850, opened 2023-03-28
btf: improve BTF types validation

tixxdz:
When we perform BTF types validation, for some types we return the type name instead of the complete definition as this goes through the BTF process.

Example:
To match "struct perf_event *" through BTF we can just do "perf_event"

This will silence the following warnings:
msg="Kprobe spec pre-validation failed, but will continue with loading" error="validate argument 0 failed: type (perf_event) does not match spec type (perf_event)\n"

### DIFF
--- a/pkg/btf/validation.go
+++ b/pkg/btf/validation.go
@@ -130,7 +130,7 @@ func ValidateKprobeSpec(bspec *btf.Spec, kspec *v1alpha1.KProbeSpec) error {
 			return &ValidationWarn{s: fmt.Sprintf("could not get the function prototype for %s: arguments will not be verified", kspec.Call)}
 		}
 		syscall := strings.TrimPrefix(kspec.Call, prefix)
-		return validateSycall(kspec, syscall)
+		return validateSyscall(kspec, syscall)
 	}
 
 	fnNArgs := uint32(len(proto.Params))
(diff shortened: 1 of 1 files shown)


pkg/btf/validation.go
```diff
@@ -173,72 +173,90 @@ func getKernelType(arg btf.Type) string {
...
 		case "union bpf_attr *":
-			return true
+		case "bpf_attr":
+			return nil
 		}
 	case "perf_event":
 		switch kernelTy {
 		case "struct perf_event *":
-			return true
+		case "perf_event":
```
kkourt, line 230: Would it make sense to check for both `struct perf_event *` and `perf_event` here?

kkourt (commented):
Nice! Thanks!

kkourt:
@tixxdz do you plan to work in this PR or should we close it?

tixxdz:
> @tixxdz do you plan to work in this PR or should we close it?

Will fix the comments and push, I just hit this over the cred object, need to find little time

kkourt:
> > @tixxdz do you plan to work in this PR or should we close it?
> 
> Will fix the comments and push, I just hit this over the cred object, need to find little time

Sounds good, thanks! Feel free to make PRs as draft if you do not plan to work on them for the foreseeable future.

PR #856, opened 2023-03-30
Helm chart: bump quay.io/cilium/hubble-export-stdout to v1.0.3

mtardy:
This is part of https://github.com/cilium/tetragon/issues/487.

This new tag supports multi-arch for amd64 and arm64. See: https://github.com/cilium/hubble-export-stdout/pull/2


michi-covalent (approved):
😊

PR #880, opened 2023-04-04
Makefile: set `GOARCH` to `TARGET_ARCH` only if `GOARCH` is unset 

mtardy:
This fixes a bug where we currently build an amd64 binary in the arm64 layer of the tetragon-operator image 😞 . See https://github.com/cilium/tetragon/issues/487#issuecomment-1495605010.

TARGET_ARCH in Makefile makes more sense for the Tetragon and BPF build because it regroups the target architecture for the Go program and the BPF code. However for pure Go programs like tetragon-operator, it makes less sense that you have to set TARGET_ARCH and that GOARCH will be overridden. This patch allows you to use both TARGET_ARCH to set everything or just GOARCH for certain pure Go program use cases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,10 @@ ifeq ($(UNAME_M),aarch64)
 endif
 TARGET_ARCH ?= amd64
 
+# Set GOARCH to TARGET_ARCH only if it's not set so that we can still use both
+# GOARCH and TARGET_ARCH (make sense for pure Go program like tetragon-operator)
+GOARCH ?= $(TARGET_ARCH)
+
 ifeq ($(TARGET_ARCH),amd64)
 	BPF_TARGET_ARCH ?= x86
 endif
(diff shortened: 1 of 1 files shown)


mtardy:
This should indeed fix it, see https://github.com/cilium/tetragon/actions/runs/4606288087/jobs/8139400747?pr=880#step:16:287.

errordeveloper (approved):

mtardy:
Honestly this patch would work as well:
```diff
diff --git a/Makefile b/Makefile
index 91d76797..256c5c98 100644
--- a/Makefile
+++ b/Makefile
@@ -141,7 +141,7 @@ tetragon-image:
 	CGO_ENABLED=1 GOOS=linux GOARCH=$(TARGET_ARCH) $(GO) build -tags netgo -mod=vendor -ldflags=$(GO_IMAGE_LDFLAGS) ./cmd/tetra/
 
 tetragon-operator-image:
-	CGO_ENABLED=0 GOOS=linux GOARCH=$(TARGET_ARCH) $(GO) build -ldflags=$(GO_OPERATOR_IMAGE_LDFLAGS) -mod=vendor -o tetragon-operator ./operator
+	CGO_ENABLED=0 $(GO) build -ldflags=$(GO_OPERATOR_IMAGE_LDFLAGS) -mod=vendor -o tetragon-operator ./operator
 
 .PHONY: install
 install:
```

And we could set GOARCH to TARGET_ARCH if it's not already set. So that we have both GOARCH and TARGET_ARCH that work.

mtardy:
It looks good https://github.com/cilium/tetragon/actions/runs/4606532947/jobs/8139967082?pr=880#step:16:284.

kkourt (approved):

PR #883, opened 2023-04-04
Tetra: split commands set between OS for portability

mtardy:
Instead of having the standalone and non-standalone build with full commands and just `getevents` and `version` in the latter, now build version dedicated to linux, darwin and windows so that build have the maximum of commands available that make sense for the platform.

This fixes the impossibility to build on darwin after this commit: a21de54923dfa99f058346053b41e512879258cd.

However, a split is still done using the standalone tag between Linux dynamically linked and Linux statically linked, because the dump command imports the bpf package that needs CGO.

This results in 4 possible versions of the Tetra CLI but should fix the macOS build and provide more commands to the users on macOS and Windows. Maybe we can simplify that by looking into the dump command and see if we can remove the bpf package dependency.


kkourt (approved):
Thanks!

willfindlay (approved):
Tragic that we have to split it but makes sense as a workaround

mtardy:
I just modified it a little bit in order to have a commom [`commands.go`](https://github.com/cilium/tetragon/pull/883/files#diff-d949040bfc4806f6695bb8f60716325965b27c097a0130472ec9d8cc201bb52b) file that will make easier to add commands in the future.

Issue #872, opened 2023-04-03
Adapt documentation for arm64 support


mtardy:
- [x] a single line for macOS folk mentioning that they can run Tetragon using a Linux VM (e.g., https://mac.getutm.app/, https://github.com/lima-vm/lima) and kind.
   -> We now have this FAQ available https://tetragon.cilium.io/docs/faq/#can-i-run-tetragon-on-mac-computers
   https://github.com/cilium/tetragon/pull/899
- [x] Document the issue with macOS Docker Desktop, and point to issues on why this does not work:
  https://github.com/docker/for-mac/issues/6800
  https://github.com/linuxkit/linuxkit/issues/3755
  -> solved by https://github.com/cilium/tetragon/pull/899 new FAQ and note in Try Tetragon on Linux.
- [x] Documentation is amd64 centric. Fix the TracingPolicy reference documentation to refer to multi-arch symbols instead of x86_64 specific ones.
  https://github.com/cilium/tetragon/pull/884

mtardy:
Done! 🎉

PR #884, opened 2023-04-04
Docs: update the TracingPolicy reference for arm64

mtardy:
Fix partially https://github.com/cilium/tetragon/issues/872.

Deploy preview [available here](https://deploy-preview-884--tetragon.netlify.app/docs/reference/tracing-policy/).

Mostly add details about kprobes and tracepoints, change all the architecture-specific syscall symbols to generic ones, add a few note and caution blocks and fix some typos.

This PR also addresses a small issue in the Netlify config since it's used by the docs Makefile and adds the note, caution and warning blocks with some styles taken from kubernetes/website.

### DIFF
--- a/docs/content/en/docs/reference/tracing-policy.md
+++ b/docs/content/en/docs/reference/tracing-policy.md
@@ -22,11 +22,13 @@ Kubernetes API server on which it is installed, for example `kubectl explain
 tracingpolicy.spec.kprobes` provides field-specific documentation and details
 on kprobe spec.
 
+{{< note >}}
 For bare metal or VM use cases without Kubernetes, the same YAML configuration
 can be passed via a flag to the Tetragon binary (using `--config-file`) or via
 the `tetra` CLI to load the policies via gRPC.
+{{< /note >}}
 
-### Getting started with Cilium Tracing Policy
+### Getting started with Tracing Policy
 
 To keep the Cilium Policy Definitions unified, the `TracingPolicy` follows the
 same CR logic and semantics that you might be familiar with from other Cilium
@@ -58,7 +60,7 @@ spec:
         - "/etc/shadow"
 ```
 
-**First required fields**
+#### Required fields
 
 The first part follows a common pattern among all Cilium Policies or more
 widely [Kubernetes object](https://kubernetes.io/docs/concepts/overview/working-with-objects/kubernetes-objects/).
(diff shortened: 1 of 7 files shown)


docs/content/en/docs/reference/tracing-policy.md
```diff
@@ -73,13 +75,52 @@ metadata:
...
+be `__x64_` instead of `__arm64_`.
+
+{{< caution >}}
+Kernel symbols contain an architecture specific prefix when they refer to
+syscall symbols. To write portable tracing policies, i.e. policies that can run
+both on `x86_64` and `arm64`, just use the symbol name without the prefix.
+
+For example, instead of writing `call: "__arm64_sys_write"` or `call:
+"__x64_sys_write"`, just write `call: "sys_write"`, Tetragon will adapt and add
+the correct prefix based on the architecture of the underlying machine.
```
kkourt, line 115:
```suggestion
the correct prefix based on the architecture of the underlying machine.
Note that the event generated as output currently includes the prefix.
```

docs/content/en/docs/reference/tracing-policy.md
```diff
@@ -73,13 +75,52 @@ metadata:
...
+{{< /caution >}}
+
 In our example, we will explore a `kprobe` hooking into the
 [`fd_install`](https://elixir.bootlin.com/linux/v6.1.9/source/fs/file.c#L602)
-kernel function.
+kernel function. The `fd_install` kernel function is called each time a file
+descriptor is installed into the file descriptor table of a process, typically
+referenced within system calls like `open` or `openat`. Hooking `fd_install`
+has its benefits and limitation, which are out of the scope of this document.
```
kkourt, line 123:
```suggestion
has its benefits and limitations, which are out of the scope of this document.
```

docs/content/en/docs/reference/tracing-policy.md
```diff
@@ -88,29 +129,85 @@ spec:
...
+{{< /note >}}
 
-Note the `syscall` field, specific to a `kprobe` spec, that indicates whether
-we will be hooking a syscall or just a regular kernel function since they use a
-[different ABI](https://gitlab.com/x86-psABIs/x86-64-ABI).
 
-Similarly to other Policies, events can be defined independently in different
-policies, or together in the same Policy. For example, we can define trace
-multiple kprobes under the same `CiliumTracingPolicy`:
+Like usually, kprobes calls can be defined independently in different policies,
```
kkourt, line 140:
```suggestion
As usual, kprobes calls can be defined independently in different policies,
```

docs/content/en/docs/reference/tracing-policy.md
```diff
@@ -88,29 +129,85 @@ spec:
...
+        field:unsigned char common_preempt_count;       offset:3;       size:1; signed:0;
+        field:int common_pid;   offset:4;       size:4; signed:1;
+
+        field:void * skbaddr;   offset:8;       size:8; signed:0;
+        field:unsigned int len; offset:16;      size:4; signed:0;
+        field:__data_loc char[] name;   offset:20;      size:4; signed:0;
+
+print fmt: "dev=%s skbaddr=%px len=%u", __get_str(name), REC->skbaddr, REC->len
+```
+
```
kkourt, line 210:
```suggestion
Similarly to kprobes, tracepoints can also hook into system calls. For more details, see the `raw_syscalls` and `syscalls` subysystems.
```

kkourt (commented):
Excellent updates, thanks!
Have some minor suggestions, but, otherwise, LGTM.

docs/content/en/docs/reference/tracing-policy.md
```diff
@@ -73,13 +75,53 @@ metadata:
...
+You can see that the exact name of the symbol for the write syscall on our
+kernel version is `__arm64_sys_write`. Note that on `x86_64`, the prefix should
+be `__x64_` instead of `__arm64_`.
+
+{{< caution >}}
+Kernel symbols contain an architecture specific prefix when they refer to
+syscall symbols. To write portable tracing policies, i.e. policies that can run
+both on `x86_64` and `arm64`, just use the symbol name without the prefix.
```
errordeveloper, line 111: Won't the same apply to other architectures?

mtardy: Theoretically yes but Tetragon only supports amd64 and arm64 at the moment. It will need some manual changes to support a new arch. Do you have a specific arch in mind?

errordeveloper: No, I'm just thinking that if sentence doesn't need to be so specific, e.g.

```suggestion
on any architecture, just use the symbol name without the architecture prefix.
```

docs/content/en/docs/reference/tracing-policy.md
```diff
@@ -73,13 +75,53 @@ metadata:
...
+ffffdeb15092a740 d _eil_addr___arm64_sys_write
+```
+
+You can see that the exact name of the symbol for the write syscall on our
+kernel version is `__arm64_sys_write`. Note that on `x86_64`, the prefix should
+be `__x64_` instead of `__arm64_`.
+
+{{< caution >}}
+Kernel symbols contain an architecture specific prefix when they refer to
+syscall symbols. To write portable tracing policies, i.e. policies that can run
```
errordeveloper, line 110:
```suggestion
syscall symbols. It's probably a good idea to write portable tracing policies that can run
```

docs/content/en/docs/reference/tracing-policy.md
mtardy: Yes but the reality is that we support only amd64 and arm64 so I'm not sure being implicit about supporting everything would help?

errordeveloper: I guess it's just a bit more pragmatic, so you don't have to worry about this nuanced documentation goes out of date. But also, as user I wonder if other architectures maybe supported in the future or this policies will only ever work for the two architectures, which is a side effect of  such explicit docs.

mtardy: Ok I changed that to more generic `to run on multiple architectures`

mtardy: Ok I changed that to more generic `to run on multiple architectures`

tixxdz (approved):
Lgtm, maybe address this https://github.com/cilium/tetragon/pull/884/commits/7f253beef2161ca43959950c120a12c6ea2f96fe#r1159640920

docs/content/en/docs/reference/tracing-policy.md
tixxdz: Should this be resolved @mtardy ?

mtardy: I don't agree with the suggestion personally but we can debate

tixxdz: both are fine, so let's just merge as it is, and we can improve anyway later! much appreciated @errordeveloper for the feedback

Marked resolved

Issue #730, opened 2023-02-24
Add a `NoPost` action replacing `Post` action


kkourt:
It seems that when the Post action was added when creating an event was not the default behavior for all TracingPolicy. Now an event is created every time any selector matches. It would then maybe be interesting to propose a NoPost action to do the contrary, suppress the emission of an event. People might be interested in that one for performance reasons.

tixxdz:
Here I see two use cases:
1. Real NoPost action in case we have tracingpolicies that construct context like the fd_install hooks and others in future.

2. This gives the impression we do not have enough flexible mechanisms to allow users to edit/override/nullify shipped tracing policies especially big ones... so now they are looking for the NoPost action. From performance perspectives we should not load such tracing policies in the first place, so a flexible load tracing policies from multiple files and allow users to nullify/override tracing policies by editing their /etc/tetragon/tetragon.policy.d/ files  then load everything is better than hack around NoPost...

PR #885, opened 2023-04-04
tetragon: Add new NoPost and Signal actions

olsajiri:
Adding new nopost and signal actions together with fixes to allow
3 actions for selector.

New Signal action sends configured signal (argSig) to the current task:
```
      matchActions:
      - action: Signal
        argSig: 10
```

New NoPost action prevents the event to be sent to user space:
```
      matchActions:
      - action: Override
        argError: -2
      - action: NoPost
```


mtardy:
Related to https://github.com/cilium/tetragon/issues/730

bpf/process/types/basic.h
```diff
@@ -73,6 +73,7 @@ enum {
 	ACTION_COPYFD = 5,
 	ACTION_GETURL = 6,
 	ACTION_DNSLOOKUP = 7,
+	ACTION_NOPOST = 8,
```
tixxdz, line 76: Hey @olsajiri so we won't get the override event?

pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml
```diff
@@ -192,6 +192,10 @@ spec:
                                     action
                                   format: int32
                                   type: integer
+                                argSig:
+                                  description: A signal number for signal action
+                                  format: int32
```
tixxdz, line 197: Can we make please a string here? so we can handle both "10" or "SIGUSR1" as an example; all of them start with "SIG*"

kkourt (approved):
Thanks, LGTM!

bpf/process/types/basic.h
olsajiri: hum, not sure what you mean, there's OVERRIDE enum earlier in the list

pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml
olsajiri: good idea, is that ok as follow up change?

kkourt: I also think suggestion make sense.
Either option (doing the in this PR or as a followup) is fine with me.
Also, found the the function we can use: https://pkg.go.dev/golang.org/x/sys/unix#SignalNum.

tixxdz (approved):
Thank you!

PR #886, opened 2023-04-05
TracingPolicy examples: use portable symbols for syscalls

mtardy:
This patch includes the change done by this command: 
```shell
find . -type f -name '*' -print0 | xargs -0 sed -i 's/call: "__x64_sys_/call: "sys_/g'
```

Apart from that, it highlights the issue that for the user, the symbol declared `sys_write` will be different than the symbol he can see from the logs `__x64_sys_write` or `__arm64_sys_write`. Should we do something about it? We could:
- retranslate `__x64_sys_write` to `sys_write` at the agent level so that all logs contain this symbol name;
- just provide a translation in `tetra`;
- nothing.

### DIFF
--- a/examples/tracingpolicy/fd_memfd_kill.yaml
+++ b/examples/tracingpolicy/fd_memfd_kill.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   kprobes:
 # int close(int fd);
-  - call: "__x64_sys_close"
+  - call: "sys_close"
     syscall: true
     args:
     - index: 0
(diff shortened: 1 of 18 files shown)


kkourt (approved):
Thanks!

With respect to the output, I'm not sure. I can see cases where the prefix is useful and others where it is not. Let's leave it as is for now and we can revisit it when a specific need arises.

errordeveloper:
A small follow up fix https://github.com/cilium/tetragon/pull/891

PR #887, opened 2023-04-05
Dockerfiles: use ENTRYPOINT instead of CMD

mtardy:
ENTRYPOINT should be used instead of CMD to declare the main program of the Dockerfile. CMD should be used for default parameters to the entrypoint binary if any.

It's still possible to modify the entrypoint with the `--entrypoint` flag, to execute a bash in the container for example and it makes the image "ready to consume", i.e. you can directly pass args and flags to the image.

See this example where we simulate the ENTRYPOINT this patch would add with the `--entrypoint` flag, to understand the convenience: 
```
docker run --rm --entrypoint /usr/bin/tetragon quay.io/cilium/tetragon:v0.9.0 --help
```
This will become:
```
docker run --rm quay.io/cilium/tetragon:vTDB --help
```


jrfastab (approved):

tixxdz (approved):
LGTM thanks! I guess we have to adapt ;-)

PR #890, opened 2023-04-06
fix: injecting Git version into tetragon-operator binary fails

### LINKED ISSUE #881: Injecting Git version into tetragon-operator binary fails
See https://github.com/cilium/tetragon/actions/runs/4606288087/jobs/8139400747?pr=880#step:16:278 for example.

It seems that the build environment has no access to the git repository files.

YTGhost:
Fixes #881

### DIFF
--- a/Dockerfile.operator
+++ b/Dockerfile.operator
@@ -17,10 +17,11 @@ ARG TARGETOS
 # TARGETARCH is an automatic platform ARG enabled by Docker BuildKit.
 ARG TARGETARCH
 ARG NOSTRIP
+ARG TETRAGON_VERSION
 
 WORKDIR /go/src/github.com/cilium/tetragon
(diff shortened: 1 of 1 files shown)


YTGhost:
@mtardy PTAL

mtardy (approved):
Thank you, it does fix the issue 😊! 

Although I realized the operator is not using the Version we pass because despite having:
```make
GO_OPERATOR_IMAGE_LDFLAGS="-X 'github.com/cilium/tetragon/pkg/version.Version=$(VERSION)' -s -w"
```
In the Makefile, it doesn't import the package...

lmb:
FYI you can get vcs information from debuginfo on recent go releases: https://pkg.go.dev/runtime/debug#BuildSetting with key `vcs.revision` as far as I can tell.

PR #891, opened 2023-04-06
TracingPolicy examples: use conventional syscall function names in comments

errordeveloper:


mtardy (commented):
Yep, I saw them and purposely excluded them because I wanted to keep the kernel function signature intact, do you think it helps to modify these comments? I was not really sure since these signatures do not exist in the end.

errordeveloper:
@mtardy  no worries, I thought you might have missed these. I am not very sure, to be honest, the example comments aren't very consistent at the moment, it'd be good if these could be automated somehow. Perhaps an IDE plugin would be better, so you can bring together policy manifests with docs from C headers or man pages.

kkourt:
Closing this since it has not seen updates from quite some time. @errordeveloper feel free to reopen it if you think it still makes sense. Thanks!

PR #895, opened 2023-04-06
use-cases: add monitor process credentials at syscalls 

tixxdz:


sharlns (approved):

mtardy:
don't hesitate to tag me @tixxdz when writing docs, your doc pages do not render as expected I think https://tetragon.cilium.io/docs/tetragon-events/linux-process-credentials/ and we could have fixed a few things.

tixxdz:
> don't hesitate to tag me @tixxdz when writing docs, your doc pages do not render as expected I think https://tetragon.cilium.io/docs/tetragon-events/linux-process-credentials/ and we could have fixed a few things.
> 
> I guess this was not intended: image

hey @mtardy thanks for the feedback, and that was the intent... I'm not sure if there is something to fix there...

PR #899, opened 2023-04-07
Docs: add a try Tetragon on Linux getting-started tutorial 

mtardy:
Preview [here](https://deploy-preview-899--tetragon.netlify.app/docs/getting-started/try-tetragon-linux/)

The goal is to make people discover Tetragon features quickly and easily. Avoiding getting too technical or into the details.

### DIFF
--- a/docs/content/en/docs/getting-started/try-tetragon-linux.md
+++ b/docs/content/en/docs/getting-started/try-tetragon-linux.md
@@ -1,0 +1,398 @@
+---
+title: "Try Tetragon on Linux"
+weight: 1
+description: "Discover and experiment with Tetragon on your local Linux host"
+---
+
+{{< caution >}}
+This guide is not a tutorial on how to deploy Tetragon standalone (i.e. without
+Kubernetes), you can see the [container deployment](/docs/getting-started/deployment/container)
+and [package deployment](/docs/getting-started/deployment/package) guides for
+that. This is just a walkthrough to try and experiment Tetragon for the first
+time.
+{{< /caution >}}
+
+{{< note >}}
+This guide has been tested on Ubuntu 22.04 and 22.10 with respectively kernel
+`5.15.0` and `5.19.0` on amd64 and arm64 but
+[any recent distribution](https://github.com/libbpf/libbpf#bpf-co-re-compile-once--run-everywhere)
+shipping with a relatively recent kernel should work.
+
+Note that you cannot run Tetragon using Docker Desktop on macOS because of a
+limitation of the Docker Desktop Linux virtual machine. Learn more about this issue
+and how to run Tetragon on a Mac computer in [this section of the FAQ page](/docs/faq/#can-i-run-tetragon-on-mac-computers).
+{{< /note >}}
+
+## Start Tetragon
+
+The easiest way to start experimenting with Tetragon is to run it via Docker
+using the released container images. If you prefer running directly the
+binaries without Docker, please refer to the
+[development setup](/docs/contribution-guide/development-setup) to see how to
+build the binaries.
+
+```shell
+docker run --name tetragon-container --rm --pull always \
+    --pid=host --cgroupns=host --privileged             \
+    -v /sys/kernel/btf/vmlinux:/var/lib/tetragon/btf    \
+    quay.io/cilium/tetragon-ci:latest
+```
+
+Let's break down the previous command:
+- `--name tetragon-container` names our container so that we can refer to it
+  later via a simple name.
+- `--rm` will make Docker delete the container once it exists.
+- `--pull always` will force Docker to download the latest existing image
+  corresponding to the `latest` mutable tag.
+- `--pid=host` is needed to read the procfs for gathering context on processes
+  started before Tetragon.
+- `--cgroupns=host` is needed to detect container feature.
+- `--privileged` is needed for Tetragon to load its BPF programs.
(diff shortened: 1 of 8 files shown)


tixxdz (approved):
lgtm after renaming container tetragon-container and thank you ;-)

docs/content/en/docs/getting-started/try-tetragon-linux.md
```diff
@@ -0,0 +1,406 @@
...
+💥 exit     /usr/bin/sh  0
+```
+
+You can see the start of `sh` first, then the execution of the multiple
+programs, `ls`, `uname` and `whoami`, and finally the exit of `sh`.
+
+{{< note >}}
+Since Tetragon is running directly on the host, you may observe a lot of
+unrelated from various process executions on your machine unrelated to what you
+typed during the shell session. Indeed, Tetragon is monitoring every process
```
kkourt, line 85:
```suggestion
Since Tetragon monitors all processes on the host  host, you may observe events
unrelated to what was
typed in the shell session. Indeed, Tetragon is monitoring every process
```

docs/content/en/docs/getting-started/try-tetragon-linux.md
```diff
@@ -0,0 +1,406 @@
...
+start and exit in your environment.
+{{< /note >}}
+
+## Write a TracingPolicy to extend Tetragon events
+
+Tetragon observes the process lifecycle by default, but for more advanced use
+cases, you can write your own custom TracingPolicy. A TracingPolicy is a
```
kkourt, line 92:
```suggestion
Tetragon by default observes only process lifecycles.  More advanced use-cases, 
require configuring Tetragon via TracingPolicies. A TracingPolicy is a
```

docs/content/en/docs/getting-started/try-tetragon-linux.md
```diff
@@ -0,0 +1,406 @@
...
+user-configurable Kubernetes custom resource (CR) that allows users to trace
+arbitrary events in the kernel and optionally define actions to take on a
+match, for enforcement for example.
```
kkourt, line 95:
```suggestion
user-configurable Kubernetes custom resource (CR) that allows users to access additional functionality,
such as tracing arbitrary events in the kernel and, optionally, define actions to take on a match.
These actions can be, for example, used to implement enforcement.
```

docs/content/en/docs/getting-started/try-tetragon-linux.md
```diff
@@ -0,0 +1,406 @@
...
+}
+```
+
+You can see that we collected information on the process itself, with for
+example, the uid, the binary name, and the start\_time. But we also collected
+information on the parent of this process, here the binary `/bin/bash`. This is
+the same kind of information we get from the base process lifecycle sensors.
+Most importantly, we can see that we hooked the function `__x64_sys_openat`
+(for an arm64 host it would be `__arm64_sys_openat`) and we also collected the
+argument's value.
```
kkourt, line 248:
```suggestion
You can see that Tetragon collected information on the process itself, with for
example, the uid, the binary name, and the start\_time, as well as on the process parent,
`/bin/bash` in this case. This is
the same kind of information we get from the base process lifecycle sensors.
Finally, we can see that we hooked the function `__x64_sys_openat`
(for an arm64 host it would be `__arm64_sys_openat`) and we also collected the
argument's value.
```

docs/content/en/docs/getting-started/try-tetragon-linux.md
```diff
@@ -0,0 +1,406 @@
...
+the executed program.
+
+### Enforcement with TracingPolicy
+
+Let's see how we can modify the previous TracingPolicy to do enforcement with
+Tetragon.
+
+{{< warning >}}
+This example is just an illustration and should not be used in a production
+environment, filtering an argument for security can be difficult, especially
```
kkourt, line 279:
```suggestion
environment.  There are many challenges in building a production-grade policy, that go beyond the scope
of this guide. For example, filtering an argument for security can be difficult, especially
```

kkourt (approved):
Have some suggestions, but generally this looks great to me! 🚀

mtardy:
> lgtm after renaming container tetragon-container and thank you ;-)

What do you think of `tetracon` :D! I'm searching for a contraction so that it's fast and easy to type and not confused with `tetragon` itself.

mtardy:
Plans to merge that with Tetragon `v0.9.1` release because of some Dockerfile and tetra fixes.

mtardy:
I pushed some changes ([see here](https://github.com/cilium/tetragon/compare/c50837efa3dc2c87804bdd0949bfbe235f073aee..22f04423531369fdbc80394e5ab68a819af3810e)):
- I simplified the Tutorial thanks to the new encoder on tetra that "understands" `openat` args (and updated the output).
- added a note about Docker Desktop on macOS and an FAQ on how to run Tetragon on Mac computers.

mtardy:
I finally fell back to using `tetragon-ci:latest` instead of a proper release because I will need Tetragon `0.10` release for this tutorial to work correctly. I prefer to edit it later when the next release is created. [See the comparison for more details](https://github.com/cilium/tetragon/compare/22f04423531369fdbc80394e5ab68a819af3810e..2abb11ad8535a59822f778ae05df93e3ce45204e).

Using a proper release will avoid some potential issues of people in the future but I prefer to merge this now because I think this is a nice addition.

mtardy:
This PR will solve the rest of the work of https://github.com/cilium/tetragon/issues/872.

PR #900, opened 2023-04-07
pkg/encoder: make customized syscalls printers portable and add open/openat

mtardy:
Also rename the output for fd_install to fd_install instead of open and change the default emoji for unrecognized symbols since previous emoji was shifting alignment.


pkg/encoder/encoder.go
```diff
@@ -184,13 +184,27 @@ func (p *CompactEncoder) EventToString(response *tetragon.GetEventsResponse) (st
 			}
 			return CapTrailorPrinter(fmt.Sprintf("%s %s %s %v", event, processInfo, file, bytes), caps), nil
 		case "fd_install":
-			event := p.Colorer.Blue.Sprintf("📬 %-7s", "open")
+			event := p.Colorer.Blue.Sprintf("📬 %-7s", "fd_install")
```
jrfastab, line 187: We dd this intentionally with the idea that fd_install is the correct hook to monitor for file "open" events.

jrfastab: put this in its own PR if we want to debate this. Then we can merge the rest of this more or less now.

PR #924, opened 2023-04-13
pks/metrics: add syscall metrics

kkourt:
This patch adds system call metrics. It looks for raw_syscall/sys_enter events and updates metrics.

It is expected to work with the examples/tracingpolicy/raw_syscalls.yaml custom resource. We can also update the code to work with kprobes, but this is left as future work.

The patch also updates the above yaml, so that it can be applied via kubectl.

### DIFF
--- a/pkg/metrics/syscallmetrics/syscallmetrics.go
+++ b/pkg/metrics/syscallmetrics/syscallmetrics.go
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package syscallmetrics
+
+import (
+	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/cilium/tetragon/pkg/metrics/consts"
+	"github.com/cilium/tetragon/pkg/syscallinfo"
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


jrfastab (approved):

pkg/metrics/syscallmetrics/syscallmetrics.go
```diff
@@ -0,0 +1,58 @@
...
+
+func Handle(event interface{}) {
+	ev, ok := event.(*tetragon.GetEventsResponse)
+	if !ok {
+		return
+	}
+
+	var syscall, namespace, pod, binary string
+	if tpEvent := ev.GetProcessTracepoint(); tpEvent != nil {
+		if tpEvent.Subsys == "raw_syscalls" && tpEvent.Event == "sys_enter" {
```
jrfastab, line 30: We can probably relax the raw_syscalls check here and just have events for all the rules? Or maybe we want those events to be names differently? Not sure.

kkourt: This metric is just for system calls. So we need to check to: select system calls but also map the first argument to a syscall name. There is a PR coming adding metrics for all events per-policy.

kkourt: PR: https://github.com/cilium/tetragon/pull/927

PR #938, opened 2023-04-19
tetragon: deal with duplicated tcpmon_map issue

kkourt:
We have some reports of sensors using the wrong tcpmon_map. Specifically, we observe sensor programs using a different tcpmon_map than the execve program.

Below is an analysis of what may cause this, and a description of the solution implemented in this patch.

The intialization order in cmd/tetragon/main.go:
 1. initialize sensor manager (InitSensorManager)
 2. create k8s watcher (getWatcher)
 3. start gRPC server (Serve)
 4. load base sensor (directly)

The intention behind the sensor manager is to serialize all concurrent requests for loading programs/creating maps/etc. Hence, all sensors are loaded via the sensor manager except the base sensor (this is partly for historic reasons, but there are also some assumptions in the code that the base sensor is loaded first).

There are two ways that the code tries to avoid duplicated maps:

First, the code in loadMaps (pkg/sensors/load.go) checks whether the map is already loaded.

	if m.PinState.IsLoaded() {
		l.WithFields(logrus.Fields{
			"sensor": s.Name,
			"map":    m.Name,
		}).Info("map is already loaded, incrementing reference count")
		m.PinState.RefInc()
		continue
	}

For above to work, the senors need to use the same pkg.sensors.program.Map instance. This not the case for tcpmon_map.

Second, the loader code checks whether the pinned path exists for the map, and if it does it just loads the map from the filesystem.

    if err := m.LoadOrCreatePinnedMap(pinPath, mapSpec); err != nil {
	  return fmt.Errorf("failed to load map '%s' for sensor '%s': %w", m.Name, s.Name, err)
    }

Both of these approaches, however, cannot handle concurrent operations. For example, the IsLoaded() check does not happen atomically. Hence, the best solution without reorganizing a lot of code to deal with above issue is to ensure that loading the base sensor is also serialized with respect to the sensor manager.

To this end, we add a wait channel as an argument to the function that starts the sensor manager. if the channel is not nil, the sensor manager goroutine will wait on this channel before it starts serving requests. We use this channel in the main function to start the sensor manager after the base sensor is loaded.

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -221,10 +221,22 @@ func tetragonExecute() error {
 		cancel()
 	}()
 
-	if err := obs.InitSensorManager(); err != nil {
+	// start sensor manager, and have it wait on sensorMgWait until we load
+	// the base sensor. note that this means that calling methods on the
+	// manager will block so they will have to either be executed in a
+	// goroutine or after we close the sensorMgWait channel to avoid
+	// deadlock.
+	sensorMgWait := make(chan struct{})
+	defer func() {
+		// if we fail before closing the channel, close it so that
+		// the sensor manager routine is unblocked.
+		if sensorMgWait != nil {
+			close(sensorMgWait)
+		}
+	}()
+	if err := obs.InitSensorManager(sensorMgWait); err != nil {
 		return err
 	}
-	observer.SensorManager.LogSensorsAndProbes(ctx)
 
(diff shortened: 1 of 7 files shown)


kevsecurity (approved):
Channels for synchronisation is not my forte but LGTM.

jrfastab:
I'm OK with this fix. One thought though is we should send the default sensors through the same channel a CRD would go through. But I'm not sure its worth the hassle or even that nice of an idea.

Issue #933, opened 2023-04-18
Add 'and' mask filter on argument


olsajiri:
When having filter on open/openat it'd be usefull to be able
to filter on flag value for specific mask like:

```
apiVersion: cilium.io/v1alpha1
metadata:
  name: "sys-openat"
spec:
  kprobes:
  - call: "sys_openat"
    return: false
    syscall: true
    args:
    - index: 0
      type: "int"
    - index: 1
      type: "string"
    - index: 2
      type: "int"
    returnArg:
      type: "int"
    selectors:
    - matchArgs:
      - index: 1
        operator: "Equal"
        values:
        - "/tmp/passwd\0"
      - index: 2
        operator: "And"
        values:
          - 1  # O_WRONLY
          - 2  # O_RDWR
```

The matchArgs 'index: 2' filter would AND values on 3rd argument value and pass the filter if any of that  results != 0.

PR #940, opened 2023-04-19
tetragon: Add support for and filter operation

olsajiri:
Adding support and operator for kernel selector operation.

It's now possible to use it in matchArgs selector like:

```
  selectors:
    - index: 2
      operator: "And"
      values:
        - 1  # O_WRONLY
        - 2  # O_RDWR
```

Rebased on top of  https://github.com/cilium/tetragon/pull/885
Fixes: #933

### DIFF
--- a/docs/content/en/docs/reference/tracing-policy.md
+++ b/docs/content/en/docs/reference/tracing-policy.md
@@ -356,6 +356,7 @@ The available operators for `matchArgs` are:
 - `NotEqual`
 - `Prefix`
 - `Postfix`
+- `Mask`
 
 **Further examples**
 
@@ -1258,6 +1259,7 @@ There are different types supported for each operator. In case of `matchArgs`:
 * NotEqual
 * Prefix
 * Postfix
+* Mask
 
 The operator types `Equal` and `NotEqual` are used to test whether the certain
 argument of a system call is equal to the defined value in the CR.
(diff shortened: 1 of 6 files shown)


docs/content/en/docs/reference/tracing-policy.md
```diff
@@ -1308,6 +1310,29 @@ The above would be executed in the kernel as
...
+  operator: "And"
+  values:
+  - 1
+  - 0x200
+```
+
+The above would be executed in the kernel as
+```yaml
+arg & 1 OR arg & 0x200
+```
```
kkourt, line 1331: Thanks adding the docs!
It's somewhat confusing that the different "AND" values are OR-ed, but it makes sense and it's great that you clarified in the docs.

kkourt (approved):

docs/content/en/docs/reference/tracing-policy.md
tixxdz: Thank you very much @olsajiri for making this. It is also a bit confusing. The doc makes it clear.

However right now we have the logical `OR` and `AND` implicitly being used in our selectors and other match fields in case we have a CRD with multiple selectors or a selector with multiple matches.

Example the `OR` and `AND` are implicitly implied as in `Equal or NotEqual` for `matchArgs`.

So how about we make it clear that this is bitwise AND with `&` instead of "And" ?   most people who may use this are technical enough to know the flags so using `&` seems fine

This also saves the `OR` and `AND` so they can be combined maybe later with the `&` , we can then change the default behaviour from `AND` to be `OR` as it is now. Not sure about this last point, but safe keeping `OR` and `And` can allow to do perform such things...

tixxdz (commented):
So post a comment about using bitwise & as a key word instead

docs/content/en/docs/reference/tracing-policy.md
olsajiri: ok, but I'm not sure using '&' as non-alpha char might bring other issues,
how about using 'Mask' instead? @tixxdz

tixxdz: > ok, but I'm not sure using '&' as non-alpha char might bring other issues, how about using 'Mask' instead? @tixxdz

I see yep, could be problematic in tracingpolicies, I don't have that much clue on it. For 'mask' hmm not sure if it can be interpreted for other bitwise operations...

olsajiri: I see 'Mask' as better fit than 'And' ... I think mask is always used like 'value & mask', if I'd think of that before I wouldn't use 'And' probably ;-)

kkourt: `BitwiseAND` would be another option.

tixxdz: 'BitwiseAND' is fine, but 'mask' is easy writing! what about document that by **default** 'mask' is 'value & mask', and if there are other user bitwise cases: like 'or', 'xor', etc we can add a "not required" property to 'MatchArgs', like: 'bitmask: ' to tune the mask and it is about bits?

tixxdz (approved):

PR #941, opened 2023-04-19
tetragon: handle process threads in clone and process_{exec|exit} events

tixxdz:
This PR adds the TID the thread ID to Process object and then it adds support to clone and process_exec|exit events.

The TID makes more sense in kprobe and tracepoints, however since we share the Process object with process_exec, let's reuse it here and ensure that the TID is not zero during process_{exec|exit}

For more details, see commits.

### DIFF
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -149,7 +149,7 @@
 #define DOCKER_ID_LENGTH 128
 
 struct msg_execve_key {
-	__u32 pid;
+	__u32 pid; // Process TGID
 	__u8 pad[4];
 	__u64 ktime;
 }; // All fields aligned so no 'packed' attribute.
@@ -176,7 +176,8 @@ struct msg_process {
 struct msg_clone_event {
 	struct msg_common common;
 	struct msg_execve_key parent;
-	__u32 pid;
+	__u32 tgid;
+	__u32 tid;
 	__u32 nspid;
 	__u32 flags;
 	__u64 ktime;
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -93,7 +93,7 @@
  */
 #define CWD_MAX	     256
 #define BUFFER	     1024
-#define SIZEOF_EVENT 32
+#define SIZEOF_EVENT 40
 #define PADDED_BUFFER \
 	(BUFFER + MAXARGLENGTH + SIZEOF_EVENT + SIZEOF_EVENT + CWD_MAX)
 /* This is the usable buffer size for args and filenames. It is calculated
@@ -161,8 +161,10 @@ struct msg_execve_key {
  */
 struct msg_process {
 	__u32 size;
-	__u32 pid;
+	__u32 pid; // Process TGID
+	__u32 tid; // Process thread
 	__u32 nspid;
+	__u32 pad;
 	__u32 uid;
 	__u32 auid;
 	__u32 flags;
@@ -276,6 +278,9 @@ struct msg_execve_event {
 	__u32 binary;
 }; // All fields aligned so no 'packed' attribute.
 
+// The execve_map_value is tracked by the TGID of the thread group
+// the msg_execve_key.pid. The thread IDs are recorded on the
+// fly and sent with every corresponding event.
 struct execve_map_value {
 	struct msg_execve_key key;
 	struct msg_execve_key pkey;
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -206,7 +206,7 @@ enum {
 
 struct exit_info {
 	__u32 code;
-	__u32 pad;
+	__u32 tid; // Thread ID
 };
 
 struct msg_exit {
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -94,7 +94,7 @@ event_args_builder(void *ctx, struct msg_execve_event *event)
 }
 
 static inline __attribute__((always_inline)) uint32_t
-event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid, __u32 flags, void *filename)
+event_filename_builder(void *ctx, struct msg_process *curr, __u64 curr_pid, __u32 flags, void *filename)
 {
 	struct execve_heap *heap;
 	int64_t size = 0;
@@ -124,8 +124,16 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid, __u3
 			flags |= EVENT_DATA_FILENAME;
 		}
 	}
+
 	curr->flags = flags;
-	curr->pid = curr_pid;
+	/* Send the TGID and TID, as during an execve all threads other
+	 * than the calling thread are destroyed, but since we hook late
+	 * during the execve then the calling thread at the hook time is
+	 * already the new thread group leader.
+	 */
+	curr->pid = curr_pid >> 32;
+	curr->tid = (__u32)curr_pid;
+	curr->pad = 0;
 	curr->nspid = get_task_pid_vnr();
 	curr->ktime = ktime_get_ns();
 	curr->size = size + offsetof(struct msg_process, args);
@@ -155,7 +163,7 @@ event_execve(struct sched_execve_args *ctx)
 	struct msg_process *execve;
 	bool walker = 0;
 	__u32 zero = 0;
-	__u32 pid;
+	__u64 pid;
 	unsigned short fileoff;
 
 	event = map_lookup_elem(&execve_msg_heap_map, &zero);
@@ -171,7 +179,7 @@ event_execve(struct sched_execve_args *ctx)
 	 *  2. Set the event->process.flags to properly reflect errors.
 	 */
 
-	pid = (get_current_pid_tgid() >> 32);
+	pid = get_current_pid_tgid();
 	parent = event_find_parent();
 	if (parent) {
 		event->parent = parent->key;
--- a/bpf/process/bpf_fork.c
+++ b/bpf/process/bpf_fork.c
@@ -23,14 +23,15 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
 	struct execve_map_value *curr, *parent;
 	struct task_struct *current_task;
 	struct tetragon_conf *config;
-	u32 pid = 0, ppid = 0, error_flags = 0;
+	u32 tgid = 0, ppid = 0, error_flags = 0;
 	int zero = 0;
 
 	if (!task)
 		return 0;
 
-	probe_read(&pid, sizeof(pid), _(&task->tgid));
-	curr = execve_map_get(pid);
+	tgid = BPF_CORE_READ(task, tgid);
+
+	curr = execve_map_get(tgid);
 	if (!curr)
 		return 0;
 
@@ -44,7 +45,7 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
 		current_task = (struct task_struct *)get_current_task();
 		probe_read(&ppid, sizeof(ppid), _(&current_task->tgid));
 		/* If they share same thread group nothing todo... */
-		if (pid != ppid) {
+		if (tgid != ppid) {
 			parent = execve_map_get(ppid);
 			if (parent)
 				__set_task_cgrpid_tracker(config, current_task,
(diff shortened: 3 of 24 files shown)


bpf/lib/process.h
```diff
@@ -177,6 +177,7 @@ struct msg_clone_event {
 	struct msg_common common;
 	struct msg_execve_key parent;
 	__u32 pid;
+	__u32 tid;
```
kkourt, line 182: I think here pid is always tid, right?
If so, let's just keep `pid`. There is no need to add a new field if it always contains the same information with an existing one.

pkg/process/process.go
```diff
@@ -257,6 +257,9 @@ func AddCloneEvent(event *tetragonAPI.MsgCloneEvent) error {
 		pi.process.ParentExecId = parentExecId
 		pi.process.ExecId = GetProcessID(event.PID, event.Ktime)
 		pi.process.Pid = &wrapperspb.UInt32Value{Value: event.PID}
+		// Add TID of the thread leader so it will be cached. Then on non execve events
+		// the TID of the triggered events will be fetched on the fly since we do not track it.
+		pi.process.Tid = &wrapperspb.UInt32Value{Value: event.TID}
```
kkourt, line 282: Will we use the cached thread id only for exec events? What happens if exec is not executed by the thread group leader?

That is process does:
```
ppid: fork()   -> pid1 is created (pid1 == tid1)
pid1: pthread_create -> spawns tid2
tid2: exec()
```

Will we get tid1 in the exec event then?

Since the tid is per-thread, while the cache we maintain is per process, it does not seem right to add the tid to the cache. I think we should always fill it from information we get from the bpf side.

kkourt: I believe subsequent  commit "bpf:exec: send the TID per exec event" https://github.com/cilium/tetragon/pull/941/commits/947fde34923bafb001d2cbf7ab1276a308a7b7e3 deals with above case.

Let's not add the tid to the process cache then?

pkg/grpc/exec/exec.go
```diff
@@ -273,14 +273,15 @@ func (msg *MsgCloneEventUnix) Cast(o interface{}) notify.Message {
...
-		tetragonProcess = process.UnsafeGetProcess()
+	proc, parent := process.GetParentProcessInternal(event.ProcessKey.Pid, event.ProcessKey.Ktime)
+	if proc != nil {
+		tetragonProcess = proc.UnsafeGetProcess()
 	} else {
 		tetragonProcess = &tetragon.Process{
 			Pid:       &wrapperspb.UInt32Value{Value: event.ProcessKey.Pid},
 			StartTime: ktime.ToProto(event.ProcessKey.Ktime),
 		}
+		process.UpdateEventProcessTid(tetragonProcess, event.Info.Tid)
```
kkourt, line 284: Shouldn't we call UpdateEventProcessTid for other paths as well?

contrib/tester-progs/threads-tester.c
```diff
@@ -0,0 +1,89 @@
+#define _GNU_SOURCE
+
+#define _GNU_SOURCE
```
kkourt, line 3: nit: This seems to be defined twice.

pkg/testutils/threadevents.go
```diff
@@ -0,0 +1,63 @@
...
+	Thread1Pid, Thread1Tid, ParentThread1Pid uint32
+}
+
+var (
+	ParentStartsRe = regexp.MustCompile(`parent:\t\t\(pid:(\d+), tid:(\d+), ppid:(\d+)\)\tstarts`)
+	Child1Re       = regexp.MustCompile(`Child 1:\t\(pid:(\d+), tid:(\d+), ppid:(\d+)\)\t`)
+	Thread1Re      = regexp.MustCompile(`Thread 1:\t\(pid:(\d+), tid:(\d+), ppid:(\d+)\)\t`)
+)
+
+func (tti *ThreadTesterInfo) ParseLine(l string) error {
```
kkourt, line 23: Is there a reason for adding this to testutils? Are there any other users than threads_test?

pkg/process/process.go
tixxdz: > Will we use the cached thread id only for exec events? What happens if exec is not executed by the thread group leader?

Yes since the cached Process must have the TID anyway as it is [API](https://github.com/cilium/tetragon/blob/main/pkg/process/process.go#L42) we can't ignore it.
 
> That is process does:
> 
> ```
> ppid: fork()   -> pid1 is created (pid1 == tid1)
> pid1: pthread_create -> spawns tid2
> tid2: exec()
> ```
> 
> Will we get tid1 in the exec event then?
> 
> Since the tid is per-thread, while the cache we maintain is per process, it does not seem right to add the tid to the cache. I think we should always fill it from information we get from the bpf side.

Good point @kkourt , and as you point out it is handled by the "bpf:exec: send the TID per exec event" https://github.com/cilium/tetragon/commit/947fde34923bafb001d2cbf7ab1276a308a7b7e3 , the exec event will have the correct TID that is same of TGID, cause our hook is late.

During the execve hook the tid tgid transfert has already happened between the old thread group leader and the new thread leader that triggered the execve https://elixir.bootlin.com/linux/v6.0/source/fs/exec.c#L1128  , in the end from our perspective the tid tgid should be the same, from the kernel perspective a task IDs exchange happened. Will add a test to the exec to assert on that

tixxdz: > I believe subsequent commit "bpf:exec: send the TID per exec event" [947fde3](https://github.com/cilium/tetragon/commit/947fde34923bafb001d2cbf7ab1276a308a7b7e3) deals with above case.
> 
> Let's not add the tid to the process cache then?

You mean part of the clone event keep it nil? what's the point of that?

The process key added to the cache is the execID that is `event.PID and event.Ktime`  , the TID has no inference there, am I missing something?

pkg/grpc/exec/exec.go
tixxdz: For the `GetProcessExit->GetParentProcessInternal()` it should return a process that has the TID set from the exec event or from the clone one. Otherwise if we miss the cache we use the one that BPF recorded. These are the only path here

pkg/testutils/threadevents.go
tixxdz: Ah yes, in the kprobe test of the follow up PR  https://github.com/cilium/tetragon/pull/946/commits/4bee439cee92c5447055a1c1390e02013e8875b7  , guess that's fine

pkg/process/process.go
tixxdz: > Let's not add the tid to the process cache then?

Got your point now @kkourt ;-) yes ! but I don't think it will help us to improve things, beside see this comment https://github.com/cilium/tetragon/pull/941#issuecomment-1523297606 for the next PR kprobes and tracepoints and for out of order events, reading back the process from the cache to construct kprobe and tracepoint will return a process.TID == 0 and we have to overwrite it anyway with the one recorded from BPF during kprobe...

So I'm thinking what's the value of not pushing it, let me see try another version maybe you have a point and it is easy to read that way?

tpapagian, line 282: I don't have a strong opinion on adding `tid` into the process cache or not. We have to override that whenever we are going to generate any user events. Based on that we don't need to set this here, but it also does not cause any issues. 

Maybe for debugging purposes, we can set this here to an invalid value (i.e. `UINT32_MAX`?) and we should also check that we don't generate any events with that (wrong) value.

tpapagian (commented):
Thanks for that PR! This will be really useful especially in the case of `kprobe`/`tracepoint` events as you do in https://github.com/cilium/tetragon/pull/946.

I believe that it will work as expected but I have some comments on making that a bit more clear to argue about that.

contrib/tester-progs/threads-tester.c
tixxdz: thanks, removed

vendor/github.com/cilium/tetragon/api/v1/tetragon/tetragon.proto
```diff
@@ -95,6 +95,7 @@ message Process {
     uint32 refcnt = 13;
     Capabilities cap = 14;
     Namespaces ns = 15;
+    google.protobuf.UInt32Value tid = 16;
```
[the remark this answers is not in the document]
tixxdz: Indeed, done!

bpf/process/bpf_fork.c
```diff
@@ -72,6 +75,10 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
 			.common.ktime = curr->key.ktime,
 			.parent = curr->pkey,
 			.pid = curr->key.pid,
+			/* Explicitly use the TGID for TID as we generate one
+			 * event per process thread group.
+			 */
```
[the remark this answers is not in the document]
tixxdz: Thank you! added the comment

pkg/process/process.go
tixxdz: Ok updated per your request to not add it into the cache, after more thinking I kept only from BPF side and will assert on tests that we do send the correct information but also added debug message in user space to ensure we log in case a mismatch.

tixxdz: > I don't have a strong opinion on adding `tid` into the process cache or not. We have to override that whenever we are going to generate any user events. Based on that we don't need to set this here, but it also does not cause any issues.
> 
> Maybe for debugging purposes, we can set this here to an invalid value (i.e. `UINT32_MAX`?) and we should also check that we don't generate any events with that (wrong) value.

That makes sense indeed, @kkourt had same suggestion to not put into the cache but have that value 0 , then later we can use it to know that we didn't add the execve event as an indication! so will use 0 as it is an invalid pid, and see how to add a final log message at the execve to catch/assert on this eventually!

For the clone event sending TID, we had a long discussion and I don't think it hurts having it, especially since I added BPF tests and debug messages to assert on it, and the updated version is clear about what we expect. So I guess I will keep it! thank you @tpapagian !

bpf/process/bpf_execve_event.c
```diff
@@ -124,8 +124,21 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid, __u3
...
+	 * During an execve all threads other than the calling thread
+	 * are destroyed, but since we hook late during the execve then
+	 * the calling thread at the hook time is already the new thread
+	 * group leader.
+	 *
+	 * Send explicitly PID and TID even if they match so we are
+	 * consistent with other BPF events.
+	 */
 	curr->pid = curr_pid;
+	curr->tid = curr_pid;
```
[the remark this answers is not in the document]
tixxdz: Indeed, bpf code changed and I'm able to load it successfully, I had two bpf errors during this series with kprobes.

So, removed that part, used one argument and shifted the related bytes accordingly. Thank you!

pkg/grpc/exec/exec.go
tixxdz: Ok @kkourt , updated and made it simpler that all paths use the `UpdateEventProcessTid()`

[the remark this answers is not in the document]
tixxdz: Thank you @tpapagian alright updated it

tixxdz:
> > Thanks!
> > The patch looks good to me, overall. Having this will certainly make our events more robust.
> > I have some nits, but the main thing I'd like to ensure we do right is _always_ using the thread id from the bpf side rather than the one from the cache. This is something you mention in the comments, and I fully agree but it's not 100% clear to me that it is reflected in the implementation.
> 
> Sure that's the aim! so for this PR that handles process_exec/exit we do use the one from BPF side, let's break it please:
> 
>     * process_exec is received:
>       `(msg *MsgExecveEventUnix) HandleMessage()->GetProcessExec(msg, true)->process.AddExecEvent()->GetProcess(process MsgProcess)-> Tid: &wrapperspb.UInt32Value{Value: process.TID},` , so  AddExecEvent() constructs it, add it to the cache and return it for the event , we should be good.
> 
>     * process_exit with a cache hit, then good we just read it from the cache and that is same value that was added by Clone or Execve one. If an execve happens by another thread other than the thread leader, then tgid and tid transfert should happen and the thread exec'in is the new thread leader with same old values , but we still record that from BPF and return it to update the cache entry.
> 
>     * process_exit without a cache hit: then we just use the TID returned from BPF side.
> 
> 
> > Specifically, we seem to be adding the thread id in the process cache. Is there a reason for it? Shouldn't we always use the thread id we get from the bpf side?
> 

So I updated the series and converted the process_exit to always use the bpf TID for both cache hit or miss paths.

Now all events use BPF value, the execve is same one, just the implementation happens to be like that, plus I added warning to ensure that we log in case TID is not set and events are being reported without it..

Much appreciated for your time @kkourt @tpapagian ;-)

bpf/lib/process.h
kkourt: OK, let's go with that.

tpapagian (approved):
Thanks for the updates based on the comments. I think now this PR LGTM.

bpf/lib/process.h
tixxdz: Thank you much appreciated for your time! so I made all the other requested changes

kkourt (approved):
🚀

PR #945, opened 2023-04-20
policyfilter: pod label filter support for tracing policies.

kkourt:
This PR  uses the policyfilter mechanism to add pod label filters in tracing policies.

Please see commits.


pkg/labels/coverage.out
```diff
@@ -0,0 +1,35 @@
+mode: set
```
kkourt, line 1: Committed by acceding: will remove.

pkg/policyfilter/state.go
```diff
@@ -168,15 +193,22 @@ type policy struct {
...
-	if pol.namespace == "" {
-		return true
+func (pol *policy) podMatches(podNs string, podLabels labels.Labels) bool {
+	if pol.namespace != "" && podNs != pol.namespace {
+		return false
 	}
-	return pol.namespace == pod.namespace
+
+	return pol.podSelector.Match(podLabels)
+}
```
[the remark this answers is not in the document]
mtardy: Well, maybe we need more than proposition if we want to do NamespacedPolicies + Labels at the same time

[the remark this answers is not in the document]
mtardy: EDIT: I understood the thing, sorry for the noise, it's just that this thing will match based on empty label if it's the correct namespace.

mtardy (approved):
This is cool! I don't know about the BPF implementation, that I guess was already ready thanks to the namespaced policies since the `policy ID <-> cgroup ID` mechanism is already in place. But I think I understand the architecture of the userspace. This feature is tested on multiple levels:
- bpf map
- agent state
- "real" Kubernetes events

Thanks for adding this feature!!

pkg/policyfilter/state.go
```diff
@@ -130,8 +131,7 @@ func (pod *podInfo) delContainers(id string) []containerInfo {
 }
 
 // containerExists checks returns true if a container exists in the pod
-// it uses log to log inconsistencies in the data.
-func (pod *podInfo) containerExists(log logrus.FieldLogger, containerID string, cgIDp *CgroupID) bool {
+func (m *state) containerExists(pod *podInfo, containerID string, cgIDp *CgroupID) bool {
```
tpapagian, line 157: nit: Maybe we can move this function a bit later (i.e. after line 217 in the same file) to be in the same place as other`state` methods.

tpapagian: It seems that you don't use `m` in this commit (nor in the subsequent events). What is the reason for this change?

pkg/policyfilter/state.go
```diff
@@ -235,7 +237,8 @@ func (m *state) updatePodHandler(pod *v1.Pod) {
 	}
 
 	namespace := pod.Namespace
-	err = m.UpdatePod(PodID(podID), namespace, containerIDs)
+	podLabels := labels.FromPod(pod)
```
tpapagian, line 270: nit: Do we need function `labels.FromPod`? Otherwise, it would be `podLabels := pod.Labels` which is similar to what we do for example with Namespace in the previous line.

tpapagian (approved):
Thanks! Generally LGTM! I have some minor comments.

docs/content/en/docs/features/k8s-filtering.md
```diff
@@ -1,47 +1,53 @@
...
+in user-space. For example, let us consider the case where we want to block an application from
+performing a system call. Performing the filtering in-kernel means that the application will never
+finish executing the system call, which is not possible if enforcement happens in user-space
+(after the fact).
+
+To ensure that namespaced tracing policies are always correctly applied, Tetragon needs to perform
+actions before containers start executing. Tetragon supports this via [OCI runtime
+hooks](https://github.com/opencontainers/runtime-spec/blob/main/config.md#posix-platform-hooks). If
+such hooks are not added, Tetragon will apply policies in a best-effort manner using information
+from the k8s API server.
```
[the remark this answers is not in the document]
kkourt: Correct.
One implementation detail is that the K8s api server does not hold the cgroup ids, but only the container ids. We use the latter to find the former in the filesystem.

pkg/policyfilter/state.go
kkourt: We are using it for the log message:
```
-               log.WithFields(logrus.Fields{
+               m.log.WithFields(logrus.Fields{
+                       "pod-id":        pod.id,
```

a bit below.

kkourt: Moved the function!

PR #946, opened 2023-04-20
tetragon: handle process threads in kprobes and tracepoints

tixxdz:
This PR adds the TID the thread ID to kprobe and tracepoint events

### DIFF
--- a/bpf/process/bpf_generic_kprobe.c
+++ b/bpf/process/bpf_generic_kprobe.c
@@ -90,7 +90,7 @@ generic_kprobe_start_process_filter(void *ctx)
 		return 0;
 	if (!policy_filter_check(config->policy_id))
 		return 0;
-	msg->id = config->func_id;
+	msg->func_id = config->func_id;
 
 	/* Initialize selector index to 0 */
 	msg->sel.curr = 0;
--- a/bpf/lib/generic.h
+++ b/bpf/lib/generic.h
@@ -34,8 +34,8 @@ struct msg_generic_kprobe {
 	struct msg_execve_key current;
 	struct msg_ns ns;
 	struct msg_capabilities caps;
-	__u64 id;
-	__u64 thread_id;
+	__u64 func_id;
+	__u64 retprobe_id;
 	__u64 action;
 	__u32 action_arg_id; // only one URL or FQDN action can be fired per match
 	__u32 pad;
--- a/bpf/lib/generic.h
+++ b/bpf/lib/generic.h
@@ -38,7 +38,7 @@ struct msg_generic_kprobe {
 	__u64 retprobe_id;
 	__u64 action;
 	__u32 action_arg_id; // only one URL or FQDN action can be fired per match
-	__u32 pad;
+	__u32 tid; // Thread ID that triggered the event
 	/* anything above is shared with the userspace so it should match structs MsgGenericKprobe and MsgGenericTracepoint in Go */
 	char args[24000];
 	unsigned long a0, a1, a2, a3, a4;
--- a/bpf/process/bpf_generic_retkprobe.c
+++ b/bpf/process/bpf_generic_retkprobe.c
@@ -70,10 +70,10 @@ BPF_KRETPROBE(generic_retkprobe_event, unsigned long ret)
 	if (!config)
 		return 0;
 
-	e->id = config->func_id;
-	e->thread_id = retprobe_map_get_key(ctx);
+	e->func_id = config->func_id;
+	e->retprobe_id = retprobe_map_get_key(ctx);
 
-	if (!retprobe_map_get(e->id, e->thread_id, &info))
+	if (!retprobe_map_get(e->func_id, e->retprobe_id, &info))
 		return 0;
 
 	*(unsigned long *)e->args = info.ktime_enter;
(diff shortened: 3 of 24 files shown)


bpf/process/bpf_fork.c
```diff
@@ -23,14 +23,15 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
 	struct execve_map_value *curr, *parent;
 	struct task_struct *current_task;
 	struct tetragon_conf *config;
-	u32 pid = 0, ppid = 0, error_flags = 0;
+	u32 tgid = 0, ctgid = 0, error_flags = 0;
```
olsajiri, line 26: hum, not sure ctgid is better name, ppid makes more sense to me

bpf/process/bpf_execve_event.c
```diff
@@ -124,8 +124,18 @@ event_filename_builder(void *ctx, struct msg_process *curr, __u32 curr_pid, __u3
...
 	curr->flags = flags;
+	/* Get curr_pid from argument to avoid BPF verifier errors and
+	 * use it for both pid and tid.
+	 * During an execve all threads other than the calling thread
+	 * are destroyed, but since we hook late during the execve then
+	 * the calling thread at the hook time is already the new thread
+	 * group leader.
+	 */
 	curr->pid = curr_pid;
+	curr->tid = curr_pid;
```
olsajiri, line 137: do we really need tid field in msg_process if it will be always pid?
same for clone and exit events.. IIUC we record tid for generic kprobes/uprobes/tracepoints
and then set it to instance of user space process  event.. but tid is never stored, right?

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -417,11 +420,14 @@ func GetRunningProcs() error {
...
-			auid:   auid,
-			pid:    uint32(pid), nspid: nspid, args: cmdsUTF,
+			pflags:               api.EventProcFS | api.EventNeedsCWD | api.EventNeedsAUID,
+			pktime:               pktime,
+			uid:                  euid, // use euid to be compatible with ps
+			auid:                 auid,
+			pid:                  uint32(pid),
+			tid:                  uint32(pid), // Read dir does not return threads and we only track tgid
+			nspid:                nspid,
+			args:                 cmdsUTF,
```
olsajiri, line 429: this seems to add more fields than just tid, and there's no changelog explanation for that

bpf/process/bpf_generic_kprobe.c
```diff
@@ -80,6 +80,8 @@ generic_kprobe_start_process_filter(void *ctx)
 		return 0;
 	msg->func_id = config->func_id;
 
+	msg->tid = (__u32)get_current_pid_tgid();
```
olsajiri, line 83: this might go to generic_process_init, it's called by all probe types on init

olsajiri (commented):
I don't think loader is affected by this, there are no threads at the time we get loader events, and you seemed to cover uprobes already

pkg/sensors/exec/procevents/proc_reader.go
tixxdz: Ah it just adds tid, the others were inlined and patch added new lines

bpf/process/bpf_generic_kprobe.c
tixxdz: Ok thanks will do

tixxdz:
> I don't think loader is affected by this, there are no threads at the time we get loader events, and you seemed to cover uprobes already

Alright, for uprobes I just added it to initialize the field , will export it to JSON then, thank you!

bpf/process/bpf_execve_event.c
tixxdz: > do we really need tid field in msg_process if it will be always pid? same for clone and exit events..
 
Ok so to break it:

- Clone we don't export events, only from ringbuffer to user space cache, so we can do `tid=pid` at the userspace, I thought of it and for the sake of asserting our current design, especially since: a. clone can create threads, and b. we are fixing stuff now since it was overlooked, I prefer to have it set in BPF as it is clear and covered with ring buffer tests.

- For exec and exit event, no strong opinion, since right now we hook late at the exec, so a. we can remove the tid from msg_process and set it at user space, or b. send it part of the BPF to be clear and consistent?
In future in case we do binary digest, we may hook early to catch interpreter (bash, etc), in that case if tid != pid then we may rethink this...  

> IIUC we record tid for generic kprobes/uprobes/tracepoints
and then set it to instance of user space process event.. but tid is never stored, right?

Yes tid is recorded on the fly at bpf for kprobes/uprobes/tracepoints and then we set it in user space, since the process in userspace in the cache that came from execve or clone should have `pid==tid`,  so `tid` is stored in the user space cache but it should always equal `pid`.

For why it is stored, no choice since it is part the [process api](https://github.com/cilium/tetragon/blob/main/pkg/process/process.go#L42) inside of `ProcessInternal`

bpf/process/bpf_fork.c
tixxdz: ok will change back, thanks

tixxdz:
@olsajiri all handled, thank you ;-)

kkourt (approved):
LGTM, thanks!

PR #947, opened 2023-04-20
tests: fail test when tracingpolicy parsing fails

willfindlay:
Our tests used to silently ignore parse errors when loading tracingpolicies. This would lead to confusing failures and make it difficult to debug exactly why a test was failing in the presence of a subtly invalid tracingpolicy. Fix this issue by correctly propagating the error in observer_test_helper.


michi-covalent (approved):
😇 never ignore errors ✅

willfindlay:
Wow, interesting... this actually uncovered a bunch of invalid assumptions in our tests! :o

tixxdz (approved):
Maybe t.Log() instead of raw printf

willfindlay:
@tixxdz I'm gonna drop that commit before we merge, it's just for testing

PR #951, opened 2023-04-26
tetragon: Loader fixes

olsajiri:
Making the loader more configurable through opts and moving override
code loading out of the loader core into kprobe opts callbacks,
plus other small fixes.

### DIFF
--- a/pkg/sensors/program/loader.go
+++ b/pkg/sensors/program/loader.go
@@ -22,7 +22,7 @@ var (
 
 // AttachFunc is the type for the various attachment functions. The function is
 // given the program and it's up to it to close it.
-type AttachFunc func(*ebpf.Program, *ebpf.ProgramSpec) (unloader.Unloader, error)
+type AttachFunc func(*ebpf.CollectionSpec, *ebpf.Program, *ebpf.ProgramSpec) (unloader.Unloader, error)
 
 type customInstall struct {
 	mapName   string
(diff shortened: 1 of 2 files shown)


pkg/sensors/program/loader.go
```diff
@@ -22,19 +22,29 @@ var (
...
+type OpenFunc func(*ebpf.CollectionSpec) error
 
 type customInstall struct {
 	mapName   string
 	secPrefix string
 }
 
+type loadOpts struct {
+	attach AttachFunc
+	open   OpenFunc
```
kkourt, line 36: Can we add a comment here to explain what this does?
Below is my best guess:

```suggestion
         // open allows modifying the ebpf.CollectionSpec before the collection is loaded into the kernel.
         // It can be used, for example, to disable loading of certain programs.
	open   OpenFunc
```

pkg/sensors/program/loader.go
```diff
@@ -121,28 +121,73 @@ func RawTracepointAttach(load *Program) AttachFunc {
 	}
 }
 
-func KprobeAttach(load *Program) AttachFunc {
+func KprobeOpen(load *Program) OpenFunc {
+	return func(coll *ebpf.CollectionSpec) error {
+		// Disable loading of override program if it's not needed
```
kkourt, line 126: Out of curiosity, why do we need to disable the program?

kkourt (approved):
Thanks!

pkg/sensors/program/loader.go
olsajiri: right, I'll add comment to loadProgram function and descirbe that in more details

pkg/sensors/program/loader.go
olsajiri: The generic_kprobe_override program is part of bpf_generic_kprobe.o object,
so we want to disable it if the override is not configured. Otherwise it gets
loaded and bpftool will show it. I'll add this comment.

PR #956, opened 2023-04-28
dependabot: fix a tag issue with alpine-curl image

mtardy:
Partially reverts commit a05a988523c68eacd4ea3bad4f541560b0441f4c and fix by ignoring the latest tag.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -33,14 +33,11 @@ updates:
     open-pull-requests-limit: 1
     rebase-strategy: disabled
     ignore:
-      # this dependency has tags beginning with `v` and some without, which
-      # means that for dependabot, 1.1 < v1.6.0. Let's ignore the first tags
-      # without the `v` letter. See more information here:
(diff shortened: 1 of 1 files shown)


aanm (approved):

jrfastab (approved):

willfindlay (approved):

mtardy:
why don't we get this **before** merging 


image

mtardy:
> why don't we get this **before** merging

So I just contacted GitHub and apparently:
> There isn't currently a great way to validate the config on a PR. There is an internal issue tracking this. But at this time, we aren't able to say if or when this will be implemented, but you can keep an eye on the [GitHub blog](https://github.blog/) as well as our [GitHub Public Roadmap](https://github.com/github/roadmap/issues) for information on when and if this makes it into the product.

I would have been fine with just "a way" but indeed this looks problematic: https://github.com/dependabot/dependabot-core/issues/4605

Issue #957, opened 2023-04-28
Add comments on proto structures to enrich generated documentation on api


mtardy:
Tetragon users would eventually need documentation on the events Tetragon exports (see this [Slack thread for example](https://cilium.slack.com/archives/C03EV7KJPJ9/p1682622098441029)). So we could add documentation straight from the proto definition so that we have a well generated documentation to propose.

We can also include this documentation into the website directly so that users don't have to switch to an external github readme when reading docs.

Issue #486, opened 2022-10-19
changes to tracing policy API (v1alpha1) and gRPC API (v1)


kkourt:
There are two external APIs for tetragon:
 * The CRD definition for tracing policies, https://github.com/cilium/tetragon/tree/main/pkg/k8s/apis/cilium.io/v1alpha1
 * The gRPC proto definitions: https://github.com/cilium/tetragon/tree/main/api/v1

This issue maintains a list for potential improvements to these APIs that would break compatibility and would require a new version.

* Unify KprobeSpec and TracepointSpec, as well as their gRPC events
* Remove Test event (it is no longer needed)
* revise sensors API. See for example:  https://github.com/cilium/tetragon/issues/478
* Have event flags beeing a list rather than a string. (see, e.g., https://github.com/cilium/tetragon/tree/main/api/v1#process)
* Can we improve ActionSelector so that arguments for specific actions are grouped together? https://github.com/cilium/tetragon/blob/cc009d4499216e6678079d182be29b01d4f711f4/pkg/k8s/apis/cilium.io/v1alpha1/types.go#L191

mtardy:
When updating the documentation on TracingPolicy and testing things out, I had some remarks about inconsistencies in the API:
- [ ] `Equal` and `In` are identical and could be merged (since `Equal` can take multiples values). I would argue in favor of `In` and `NotIn` for everything.
- [ ] Fix `sizeArgIndex` to be the index of the arg and not the index + 1 (starting at 1 vs starting at 0).
- [ ] I think naming camel case consistency was already done (with `matchargs` becoming `matchArgs`, `followFd` becoming `followFD` etc...) but we should finish with `argFd`, `argFqdn`, `GetUrl`, `dnsLookup`
- [ ] The `matchAction` name is confusing since it's not matching anything but just performing an action when the whole selector matches. Maybe having filtering stuff and actions stuff in selectors? Or I know there was discussion on moving the action part of out of the selectors maybe.
- [ ] What do we think about the `matchNamespaces` with plural namespaces and `matchNamespaceChanges` with singular namespace?
- [ ] On vocabulary, it seems that in some places filters (like `matchArgs`, `matchBinaries`) can be named selectors which confused me a lot. I would make it clear that selectors are a group of filters (the last name might evolve because of `matchActions` that does not really sound like a filter).

willfindlay:
Thanks @mtardy, a couple more thoughts from me:

[...]

PR #959, opened 2023-04-28
api: add comments on fields for reference docs

mtardy:
Fixes https://github.com/cilium/tetragon/issues/957.

This PR includes no functional changes. Only comments for autogenerated documentation.
I need your help with a few fields.


api/v1/tetragon/tetragon.proto
```diff
@@ -45,71 +50,128 @@ message Container {
...
     string flags = 7;
+    // Start time of the execution.
     google.protobuf.Timestamp start_time = 8;
+    // TODO: what is this identifier about? related to uid apparently.
     google.protobuf.UInt32Value auid = 9;
+    // Information about the the Kubernetes Pod where the event originated.
     Pod pod = 10;
+    // TODO: is this the 15 first digits of the docker ID? We should name it
+    // something else than docker since most k8s install no longer use Docker.
     string docker = 11;
```
mtardy, line 139: Need minor help here!

api/v1/tetragon/tetragon.proto
```diff
@@ -45,71 +50,128 @@ message Container {
...
+    // Exec ID of the parent process.
     string parent_exec_id = 12;
+    // TODO: is this a similar refcnt than from the BPF stats?
     uint32 refcnt = 13;
+    // Set of capabilities that define the permissions the process can execute with.
     Capabilities cap = 14;
+    // TODO: is this used?
     Namespaces ns = 15;
```
mtardy, line 147: Do you know when it's used?

api/v1/tetragon/tetragon.proto
```diff
@@ -45,71 +50,128 @@ message Container {
...
     string cwd = 4;
+    // Absolute path of the executed binary.
     string binary = 5;
+    // Arguments passed to the binary at execution.
     string arguments = 6;
+    // Information about which syscalls generated events, use of internal
+    // Tetragon buffers, errors and more.
+    // TODO: not sure how to documents all the existing flags without dumping
+    // 60 lines of text
     string flags = 7;
```
mtardy, line 130: What do you think?

api/v1/tetragon/tetragon.proto
```diff
@@ -45,71 +50,128 @@ message Container {
...
+    // Start time of the execution.
     google.protobuf.Timestamp start_time = 8;
+    // TODO: what is this identifier about? related to uid apparently.
     google.protobuf.UInt32Value auid = 9;
```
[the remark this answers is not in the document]
willfindlay: auid is audit UID

api/v1/tetragon/tetragon.proto
willfindlay: Yep this is correct. And yes it should probably be deprecated and renamed.

api/v1/tetragon/tetragon.proto
```diff
@@ -45,71 +50,128 @@ message Container {
...
+    // Information about the the Kubernetes Pod where the event originated.
     Pod pod = 10;
+    // TODO: is this the 15 first digits of the docker ID? We should name it
+    // something else than docker since most k8s install no longer use Docker.
     string docker = 11;
+    // Exec ID of the parent process.
     string parent_exec_id = 12;
+    // TODO: is this a similar refcnt than from the BPF stats?
     uint32 refcnt = 13;
```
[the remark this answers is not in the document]
willfindlay: I believe this is the processCache reference count. When it hits zero, the entry is evicted from the processCache. https://github.com/cilium/tetragon/blob/main/pkg/process/cache.go#L37
https://github.com/cilium/tetragon/blob/bc5f11a52c6bc19607dc6c15aa7fe44894e86a1b/pkg/process/process.go#L91

api/v1/tetragon/tetragon.proto
willfindlay: https://github.com/cilium/tetragon/blob/bc5f11a52c6bc19607dc6c15aa7fe44894e86a1b/pkg/process/process.go#L134-L136

api/v1/tetragon/tetragon.proto
willfindlay: It's too bad this is a string instead of a protobuf enum :/ probably too late to change it sadly.

mtardy:
Thanks @willfindlay, I added some of your API changes suggestion there https://github.com/cilium/tetragon/issues/486#issuecomment-1431019714.

api/v1/tetragon/tetragon.proto
mtardy: ah okay there is a `option.Config.EnableProcessNs`

sharlns (approved):
Amazing, LGTM. Thank you! :)

api/v1/tetragon/tetragon.proto
```diff
@@ -45,71 +50,178 @@ message Container {
...
+    // - `rootCWD` Indicates that CWD is the root directory. This is necessary
+    // to inform readers the CWD is not in the event buffer and is '/' instead.
+    // - `errorCWD` An error flag indicating an error occurred while reading
+    // the CWD of a process. If this is set it should be reported to Tetragon
+    // developers for debugging.
+    // - `clone` Indicates the process issued a clone before exec*. This is the
+    // general flow to exec* a new process, however its possible to replace the
+    // current process with a new process by doing an exec* without a clone. In
+    // this case the flag will be omitted and the same PID will be used by the
+    // kernel for both the old process and the newly exec'd process.
```
tixxdz, line 177: This field is somehow for debugging, it contains some values that may not change but others may...

Would be better to add something like:
The flags is for debugging purposes only and should not be relied upon to propagate comprehensive information.

mtardy: when you say "add" you mean append this at the end or at the beginning or replacing the whole flag documentation with this?

tixxdz: I'm fine adding it at the beginning so it is clear and we keep current doc that you added, the developers are on their own here... ;-)

mtardy: @tixxdz see https://github.com/cilium/tetragon/compare/f4a972a98319ec1f9608f835d04d8524b0aefeda..46156cca63c72f36073b79d8deddfee603b66d69.

mtardy:
Check failure is related to dependabot.yaml config format fails and I have an open PR to fix that. Let's merge this.

api/v1/tetragon/tetragon.proto
tixxdz: @willfindlay we just made a more of debugging flag, we can add a more reliable field later and get rid of this one if developers or other tools request this... so I guess we are fine ;-) !

PR #961, opened 2023-04-29
tetragon: we only need to find the BTF file skip creating it

jrfastab:
Creating BTF obj consumes 30MB of data we really only need to find the file to feed to cilium/ebpf.

### DIFF
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -17,7 +17,6 @@ import (
 )
 
 var (
-	btfObj  *btf.Spec
 	btfFile string
 )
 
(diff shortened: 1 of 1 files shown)


pkg/btf/btf.go
```diff
@@ -89,14 +88,13 @@ func InitCachedBTF(ctx context.Context, lib, btf string) error {
...
 	return err
 }
 
 func GetCachedBTFFile() string {
 	return btfFile
 }
 
-func GetCachedBTF() *btf.Spec {
-	return btfObj
+func GetCachedBTF() (*btf.Spec, error) {
```
kkourt, line 98: Could we just remove this function altogether and have callers use NewBTF, since we do not cache the object anymore?

kkourt (approved):

michi-covalent (approved):
😇

pkg/btf/btf.go
jrfastab: sure I was just minimizing changes.

jrfastab:
Updated to remove GetCachedBTF altogether.

jrfastab:
@kkourt merged this but feel free to add more comments I'll fix anything.

PR #964, opened 2023-05-02
hubble-fgs: Slightly more aggressive Spec flushing

jrfastab:
We can't know when if ever another configuration will be loaded so lets flush the spec considering current use cases the configuration is mostly static. Prior to this we only flush specs created directly from Tetragon this will also flush the Specs loading via cilium/ebpf creates.

The advantage is a default run of tetragon monitoring exec can fit in <10MB of data leaving lots of room for metrics, logs, kprobes and so on.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/cilium/cilium v1.13.1
-	github.com/cilium/ebpf v0.9.4-0.20221102092914-a9cf21df64c2
+	github.com/cilium/ebpf v0.10.0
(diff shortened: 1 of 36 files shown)


tixxdz (approved):
It needs a make vendor first?

Seems calling btf.FlushKernelSpec() is fine, the aim here is main programs, but could also be added at the server paths when doing dynamic load create sensors with tracing policies... especially if there are no side effects

LGTM after make vendor, cc @olsajiri

jrfastab:
some other solution landed

PR #968, opened 2023-05-03
policyfilter: only deal with running containers

kkourt:
As things stand now, we pass all containers to the policyfilter code. As a result, we get log warnings for containers that are not running because we are unable to find their cgroup ids.

This patch fixes this issue by only dealing with containers that have a running status.


tpapagian (approved):
LGTM!

tixxdz (approved):

PR #980, opened 2023-05-09
dependabot: remove ignore on tag name

mtardy:
This kinda revert 22c9d55afb899aa5fbdfdbebbd0a154b428f749a.

Dependabot cannot ignore tags on Docker images, it only works on semantic version numbers. It is not documented but the GitHub support provided us the information. We could not patch dependabot behavior via configuration so we bumped the image tag itself.


tixxdz (approved):

mtardy:
Hopefully this will not be very useful since we are supposed to switch to renovate

PR #976, opened 2023-05-05
Cleanup: update golangci-lint and fix new linters alerts

mtardy:
This PR should be just a cleanup with no intended functional changes.
It fixes the following 129 alerts. I split the unused-parameter fixes between the one with and without `context.ctx` removal.

Please review carefully the commit that removes unused parameters since some look very suspicious and could correspond to bugs (the author might have not used the param without realizing it).

```
pkg/stacktracetree/stacktracetree_test.go:11:17               revive  unused-parameter: parameter 't' seems to be unused, consider removing or renaming it as _
pkg/stacktracetree/stacktracetree.go:37:26                    revive  empty-block: this block is empty, you can remove it
pkg/stacktracetree/stacktracetree.go:39:33                    revive  empty-block: this block is empty, you can remove it
pkg/vmtests/list.go:38:38                                     revive  unused-parameter: parameter 'testDir' seems to be unused, consider removing or renaming it as _
cmd/tetragon/conf_test.go:1075:22                             revive  unused-parameter: parameter 't' seems to be unused, consider removing or renaming it as _
cmd/tetragon/conf_test.go:1145:20                             revive  unused-parameter: parameter 't' seems to be unused, consider removing or renaming it as _
cmd/protoc-gen-go-tetragon/helpers/helpers.go:192:2           revive  if-return: redundant if ...; err != nil check, just return error instead.
pkg/testutils/sensors/load.go:25:91                           revive  unused-parameter: parameter 't' seems to be unused, consider removing or renaming it as _
pkg/testutils/sensors/load.go:46:68                           revive  unused-parameter: parameter 't' seems to be unused, consider removing or renaming it as _
pkg/sensors/exec/cgroups_test.go:244:33                       revive  unused-parameter: parameter 't' seems to be unused, consider removing or renaming it as _
contrib/rthooks/tetragon-oci-hook/main.go:33:19               revive  unused-parameter: parameter 'log' seems to be unused, consider removing or renaming it as _
pkg/policyfilter/dummy.go:16:27                               revive  unused-parameter: parameter 'polID' seems to be unused, consider removing or renaming it as _
pkg/policyfilter/dummy.go:20:27                               revive  unused-parameter: parameter 'polID' seems to be unused, consider removing or renaming it as _
pkg/policyfilter/dummy.go:24:33                               revive  unused-parameter: parameter 'podID' seems to be unused, consider removing or renaming it as _
[...]

PR #983, opened 2023-05-09
cleanup: add new linters and fix bugs and alerts

mtardy:
This is a follow-up of https://github.com/cilium/tetragon/pull/976 to enable new linters to increase the code quality, remove bugs and improve readability in a sustainable way.

Add the following linters and fix their alerts:
- **gosimple**: Linter for Go source code that specializes in simplifying code
- **typecheck**: Like the front-end of a Go compiler, parses and type-checks Go code
- **asciicheck**: Simple linter to check that your code does not contain non-ASCII identifiers
- **bidichk**: Checks for dangerous unicode character sequences
- **makezero**: Finds slice declarations with non-zero initial length **(found bugs)**
- **loggercheck**: Checks key value pairs for common logger libraries (kitlog,klog,logr,zap). **(found bugs)**
- **exportloopref**: Checks for pointers to enclosing loop variables. **(found bugs)**
- **dupword**: Checks for duplicate words in the source code
- **gofmt**: Gofmt checks whether code was gofmt-ed. By default this tool runs with -s option to check for code simplification.
- **errname**: Checks that sentinel errors are prefixed with the Err and error types are suffixed with the Error.
- **gocritic**: Provides diagnostics that check for bugs, performance and style issues. Extensible without recompilation through dynamic rules. Dynamic rules are written declaratively with AST patterns, filters, report message and optional suggestion.
- **unparam**: Reports unused function parameters

I would love to enable these ones about errors, but those will be a followup cleanup PR since this is a lot of work:
- errcheck (this one is particularly important)
- nilnil
- nilerr
- errorlint

We still don't catch situations where we use the return value _before_ checking the error.

Added this issue for follow-up https://github.com/cilium/tetragon/issues/989.

Just for info the pattern used here, that defaults to error and rely on the function to return nil and override is a bit weird and led to an issue while removing useless "always nil" return value from functions:
https://github.com/cilium/tetragon/blob/557a6c93b1f71aed0af6da3c94323867426b2e97/pkg/sensors/manager.go#L60

### DIFF
--- a/pkg/btf/validation.go
+++ b/pkg/btf/validation.go
@@ -295,10 +295,6 @@ spec:
 	for _, key := range syscallinfo.SyscallsNames() {
 		var fn *btf.Func
 
-		if key == "" {
-			continue
-		}
-
 		sym, err := arch.AddSyscallPrefix(key)
 		if err != nil {
 			return "", err
--- a/pkg/btf/validation.go
+++ b/pkg/btf/validation.go
@@ -316,7 +316,7 @@ spec:
         values:
         - "` + binary + `"`
 
-			crd = crd + filter
+			crd += filter
 		}
 	}
 
--- a/pkg/btf/validation.go
+++ b/pkg/btf/validation.go
@@ -310,7 +310,7 @@ spec:
 		}
 
 		crd = crd + "\n" + fmt.Sprintf("  - call: \"%s\"", key)
-		crd = crd + "\n" + fmt.Sprintf("    syscall: true")
+		crd = crd + "\n" + "    syscall: true"
 
 		if binary != "" {
 			filter := `
--- a/pkg/btf/validation.go
+++ b/pkg/btf/validation.go
@@ -46,7 +46,7 @@ func validate(btf bpf.BTF, spec *v1alpha1.KProbeSpec) (bpf.BtfID, error) {
 
 	llCallTy, err := btf.TypeByID(llCallID)
 	if err != nil {
-		fmt.Errorf("failed to to find syscall type by id: %w", err)
+		fmt.Errorf("failed to find syscall type by id: %w", err)
 	}
 
 	return btf.UnderlyingType(llCallTy)
--- a/pkg/btf/validation.go
+++ b/pkg/btf/validation.go
@@ -15,24 +15,24 @@ import (
 	"github.com/cilium/tetragon/pkg/syscallinfo"
 )
 
-// ValidationWarn is used to mark that validation was not successful but it's not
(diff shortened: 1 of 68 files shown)


pkg/btf/validation.go
```diff
@@ -295,10 +295,6 @@ spec:
 	for _, key := range syscallinfo.SyscallsNames() {
 		var fn *btf.Func
 
-		if key == "" {
-			continue
-		}
-
```
willfindlay, line 301: Doesn't this change the behaviour of the loop?

pkg/sensors/tracing/selectors_test.go
```diff
@@ -223,7 +223,7 @@ func TestTracepointSelectors(t *testing.T) {
 					return nil
 				}
 
-				ret[whence] = ret[whence] + 1
+				ret[whence] += +1
```
willfindlay, line 226: += +1 seems weird.

willfindlay (approved):
LGTM just couple questions/nits.

willfindlay:
Btw I'm also not a fan of that "BUG in SensorCtl" thing. Maybe we can refactor that slightly in the future.

pkg/btf/validation.go
mtardy: It was actually a fix for a bug because I think the author did not understand why they were empty keys in the slice. See this:
https://github.com/cilium/tetragon/pull/983/commits/f021b5de1998c6af6a74e588b882e5780eae0502#diff-468002e48832ac8408184ef20e818fa41c0f35b2a7562d8bc20b902182860bfdL47-R47.

See this commit message https://github.com/cilium/tetragon/pull/983/commits/f021b5de1998c6af6a74e588b882e5780eae0502.

mtardy:
> LGTM just couple questions/nits.

waw thanks a lot for that fast and good review! :)

pkg/sensors/tracing/selectors_test.go
mtardy: here it was rightly fixed https://github.com/cilium/tetragon/pull/983/files/3c105466c9d6eb785f87070ef35704167a6e43bf#diff-6b20c8ee0a0503a91754d733ce9932baa6b15a784384595466f3f05fdc033d5fL353-R353.

mtardy: fixed! thanks

pkg/btf/validation.go
willfindlay: Gotcha.

kevsecurity: Can we change the 'continue' to 'return error' instead? (As per my comment on #1832?

PR #984, opened 2023-05-09
tetragon: Use execve_map_get_noinit cgroup related code

olsajiri:
hunt for execve_map leak showed few places that use 
execve_map_get instead of execve_map_get_noinit.

Even though it did not turn out to be the source of the leak,
we should use execve_map_get_noinit calls.

The execve_map_get is used only by fork sensor to create the 
stat record.


tpapagian (approved):
LGTM, thanks!

tixxdz (approved):
Thank you @olsajiri !

I wonder if we could somehow compute the entries from /proc at startup they are pushed to bpf, so use the all value of execve_map_stats then with execve event, then correlate this with user space process cache

Seems we don't decrement the execve_map_stats counter on process exit ?! anyway not related here

PR #987, opened 2023-05-10
Revert: tetragon: Switch exit tracepoint to __put_task_struct kprobe

tpapagian:
Previously, we moved the exit hook from a `tracepoint` to `kprobe/__put_task_struct` because of some issues (more details in https://github.com/cilium/tetragon/pull/558).

On the other hand, with the new kprobe we have seen cases where the kprobe is interrupted (and aborted) which means that we missed close events.

As a first step, we revert https://github.com/cilium/tetragon/commit/1a37adf4bd400733292900335bf246f4263622bf until we find a better hook point.

This commit also disables `TestExitLeader` and `TestExitZombie` due to failures from reverting the previous commit.

### DIFF
--- a/pkg/sensors/exec/exit_test.go
+++ b/pkg/sensors/exec/exit_test.go
@@ -6,7 +6,6 @@ import (
 	"os/exec"
 	"sync"
 	"testing"
-	"time"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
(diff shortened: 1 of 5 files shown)


tixxdz (approved):

pkg/sensors/exec/exit_test.go
```diff
@@ -53,73 +52,75 @@ func TestExit(t *testing.T) {
...
-	if err := jsonchecker.JsonTestCheck(t, checker); err != nil {
-		t.Logf("error: %s", err)
-		t.Fail()
-	}
-}
-
+// FIXME: Disable that for now as we revert https://github.com/cilium/tetragon/commit/1a37adf4bd400733292900335bf246f4263622bf
+// func TestExitLeader(t *testing.T) {
+// 	var doneWG, readyWG sync.WaitGroup
+// 	defer doneWG.Wait()
```
kkourt, line 58: Just as a note, I think t.Skip() would have been a better option here.

tpapagian: I have created https://github.com/cilium/tetragon/pull/990 to fix that.

PR #988, opened 2023-05-10
logcapture: use T.Log instead of T.Logf

willfindlay:
The inner logrus logger is already taking care of formating, meaning that when we have an actual % character inside the resulting string,
we end up with potential formatting issues by attempting to use that string as a format string once again. This was causing some strange bugs in tests, such as the following:

    logcapture.go:23: time="2023-05-11T10:57:30-04:00" level=info msg="BPF events statistics: 72 received, 0%!e(MISSING)vents loss"

To fix this, we can use the plain T.Log() function instead of T.Logf() when writing to the LogCapturer.

### DIFF
--- a/pkg/testutils/logcapture.go
+++ b/pkg/testutils/logcapture.go
@@ -5,6 +5,7 @@ package testutils
 
 import (
 	"io"
+	"strings"
 	"testing"
 
 	tus "github.com/cilium/tetragon/pkg/testutils/sensors"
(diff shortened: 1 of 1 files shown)


willfindlay:
@tixxdz ah my mistake, I assumed that %g was a typo but it's actually not... let me try with fmt.Sprintf inside the Infof call and see if that fixes it.

willfindlay:
No dice, marking as draft for now while I investigate more.

michi-covalent (approved):

tixxdz (approved):

PR #994, opened 2023-05-12
tests: assert thread exit without extra execve

tixxdz:
tests: assert thread exit without extra execve

on top of https://github.com/cilium/tetragon/pull/946


tixxdz:
Closing this since we have the tests now!

PR #995, opened 2023-05-12
tetragon: load tracingpolicies from directory

tixxdz:
Load tracing policies from /etc/tetragon/tetragon.tp.d/
    
Allow to use 1 depth directory to organize tracing policies, example: file access, network access, etc...
    
### DIFF
--- a/.github/workflows/packages-e2e-tests.yaml
+++ b/.github/workflows/packages-e2e-tests.yaml
@@ -1,0 +1,72 @@
+name: Packages e2e Tests
+
+on:
+  pull_request:
+    paths-ignore:
+    - "**.md"
+
+jobs:
+  standalone-tests:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-22.04
+            arch: amd64
+
+    steps:
+      # https://github.com/docker/setup-buildx-action
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@2a1a44ac4aa01993040736bd95bb470da1a38365 # v2.8.0
+
+      - name: Checkout Source Code
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+          submodules: true
+
+      - name: Getting version tag
+        id: tag
+        run: echo "tag=$(make version)" >> $GITHUB_OUTPUT
+
+      - name: Generate Tetragon Tarball
+        id: tetragon-tarball
+        run: make tarball
+
+      - name: Copy bpf.yaml tracing policy to /etc/tetragon/tetragon.tp.d/
+        run: |
+          sudo mkdir -p /etc/tetragon/tetragon.tp.d/
+          sudo cp examples/tracingpolicy/bpf.yaml /etc/tetragon/tetragon.tp.d/bpf.yaml
+
+      - name: Install Tetragon Tarball
+        run: |
+          tar zxvf tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz
+          sudo ./tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}/install.sh
+        working-directory: ./build/${{ matrix.arch }}/linux-tarball/
+
+      - name: Wait for Tetragon service
+        uses: nick-fields/retry@v2
+        with:
+          timeout_seconds: 30
+          max_attempts: 5
+          retry_wait_seconds: 5
+          retry_on: error
+          command: |
+            sudo systemctl is-active tetragon
+            sudo tetra status
+
+      - name: Check Tetragon startup logs
+        run: sudo journalctl -b -u tetragon --no-pager
+
+      - name: Test Tetragon
+        run: |
+          sudo tetra status
+          sudo grep "tetra" /var/log/tetragon/tetragon.log
+          sudo tetra tracingpolicy list | grep bpf -
+          sudo tetra bugtool
+
+      - name: Uninstall Tetragon Tarball
+        run: |
+          sudo ./tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}/uninstall.sh
+        working-directory: ./build/${{ matrix.arch }}/linux-tarball/
--- a/Dockerfile
+++ b/Dockerfile
@@ -79,6 +79,8 @@ RUN curl -L https://github.com/libbpf/bpftool/releases/download/${BPFTOOL_TAG}/b
 FROM docker.io/library/alpine:3.18.2@sha256:82d1e9d7ed48a7523bdebc18cf6290bdb97b82302a8a9c27d4fe885949ea94d1 as base-build
 RUN apk add iproute2
 RUN mkdir /var/lib/tetragon/ && \
+    mkdir -p /etc/tetragon/tetragon.conf.d/ && \
+    mkdir -p /etc/tetragon/tetragon.tp.d/ && \
     apk add --no-cache --update bash
 COPY --from=tetragon-builder /go/src/github.com/cilium/tetragon/tetragon /usr/bin/
 COPY --from=tetragon-builder /go/src/github.com/cilium/tetragon/tetra /usr/bin/
--- a/cmd/tetragon/flags.go
+++ b/cmd/tetragon/flags.go
@@ -39,6 +39,7 @@ const (
 	keyEnableProcessNs   = "enable-process-ns"
 	keyConfigFile        = "config-file"
 	keyTracingPolicy     = "tracing-policy"
+	keyTracingPolicyDir  = "tracing-policy-dir"
 
 	keyCpuProfile = "cpuprofile"
 	keyMemProfile = "memprofile"
@@ -136,6 +137,8 @@ func readAndSetFlags() {
 	option.Config.EnableMsgHandlingLatency = viper.GetBool(keyEnableMsgHandlingLatency)
 
 	option.Config.EnablePidSetFilter = viper.GetBool(keyEnablePidSetFilter)
+
+	option.Config.TracingPolicyDir = viper.GetString(keyTracingPolicyDir)
 
 	// deprecation timeline: deprecated -> 0.10.0, removed -> 0.11.0
 	// manually handle the deprecation of --config-file
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -654,7 +654,7 @@ func startGopsServer() error {
 func execute() error {
 	rootCmd := &cobra.Command{
 		Use:   "tetragon",
-		Short: "Run the tetragon agent",
+		Short: "Tetragon - eBPF-based Security Observability and Runtime Enforcement",
 		Run: func(cmd *cobra.Command, args []string) {
 			readAndSetFlags()
 
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -5,6 +5,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"io/fs"
 	"net"
 	"net/http"
 	pprofhttp "net/http/pprof"
@@ -13,6 +14,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
+	"strings"
 	"sync"
 	"syscall"
 	"time"
@@ -132,6 +134,11 @@ func tetragonExecute() error {
 	if err := logger.SetupLogging(option.Config.LogOpts, option.Config.Debug); err != nil {
 		log.Fatal(err)
 	}
+
+	if filepath.IsAbs(option.Config.TracingPolicyDir) == false {
+		log.Fatalf("Failed path specified by --tracing-policy-dir '%q' is not absolute", option.Config.TracingPolicyDir)
+	}
+	option.Config.TracingPolicyDir = filepath.Clean(option.Config.TracingPolicyDir)
 
 	if option.Config.RBSize != 0 && option.Config.RBSizeTotal != 0 {
 		log.Fatalf("Can't specify --rb-size and --rb-size-total together")
@@ -278,8 +285,7 @@ func tetragonExecute() error {
 	obs.RemovePrograms()
 	os.Mkdir(defaults.DefaultRunDir, os.ModeDir)
 
-	err := btf.InitCachedBTF(option.Config.HubbleLib, option.Config.BTF)
-	if err != nil {
+	if err := btf.InitCachedBTF(option.Config.HubbleLib, option.Config.BTF); err != nil {
 		return err
 	}
 
@@ -361,14 +367,14 @@ func tetragonExecute() error {
 	sensorMgWait = nil
 	observer.SensorManager.LogSensorsAndProbes(ctx)
 
+	err = loadTpFromDir(ctx, option.Config.TracingPolicyDir)
+	if err != nil {
+		return err
+	}
+
 	// load sensor from tracing policy file
 	if len(option.Config.TracingPolicy) > 0 {
-		tp, err := tracingpolicy.PolicyFromYAMLFilename(option.Config.TracingPolicy)
-		if err != nil {
-			return err
-		}
-
-		err = observer.SensorManager.AddTracingPolicy(ctx, tp)
+		err = addTracingPolicy(ctx, option.Config.TracingPolicy)
 		if err != nil {
 			return err
 		}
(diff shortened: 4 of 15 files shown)


pkg/tracingpolicy/tracingpolicy.go
```diff
@@ -18,6 +18,8 @@ type TracingPolicy interface {
 	TpSpec() *v1alpha1.TracingPolicySpec
 	// TpInfo returns a description of the policy
 	TpInfo() string
+	// TpAuthor  returns the author of the policy
+	TpAuthor() string
```
kkourt, line 22: Just a first (quick) comment :)

What's the motivation for adding an author at the interface? 
Generally speaking I'd prefer the to keep the interface simple.

tixxdz: > Just a first (quick) comment :)
> 
> What's the motivation for adding an author at the interface? Generally speaking I'd prefer the to keep the interface simple.

I agree, the aim is to print the author of the policy: give credit to users or identify policies, but maybe we can do it without adding a new function or return author part of info?

kkourt: We can identify policies by their name. I'd leave the author field out for now unless a user requests it.

tixxdz: > We can identify policies by their name. I'd leave the author field out for now unless a user requests it.

Sure but author also helps to identify it (who did write it) this PR ads possibility to load from multiple directories..., so what we can do is add an annotation method which is already part of the k8s spec anyway without detailing the fields.

The annotations[author] is optional as it is now, and users can use it if they want.

About leaving that as a request for users, the other way around is better, be proactive , put some examples as we are trying todo, document it, etc which should allow users and projects to take advantage of it...

tixxdz: made it generic k8s annotations logic , then inside our helpers and log messages we can print the author field if set.

kkourt: I still don't see a use-case for this. We cannot copy all the information from the tracing policy anyway. I'd rather not add more methods to the tracing policy interface.

pkg/tracingpolicy/tracingpolicy.go
```diff
@@ -18,6 +18,10 @@ type TracingPolicy interface {
 	TpSpec() *v1alpha1.TracingPolicySpec
 	// TpInfo returns a description of the policy
 	TpInfo() string
+
+	TpHasAnnotation(string) bool
+
+	TpAnnotations() map[string]string
```
kkourt, line 24: As I said in the other comment, I don't see a use-case for this. Whatever we want we can add into `TpInfo()`. I'd like to keep the `TracingPolicy` interface as simple as possible.

kkourt (changes requested):

sn99:
Hi, I am not wrong then currently there is no way to automatically load policies?

tixxdz:
> Hi, I am not wrong then currently there is no way to automatically load policies?

Hi, actually there is with `--config-file` that was converted to `--tracing-policy` to specify tracing policy file.

This series is to expend on it... just for curiosity what gave you this impression? or any feedback so we improve the docs?

sn99:
> > Hi, I am not wrong then currently there is no way to automatically load policies?
> 
> Hi, actually there is with `--config-file` that was converted to `--tracing-policy` to specify tracing policy file.
> 
> This series is to expend on it... just for curiosity what gave you this impression? or any feedback so we improve the docs?

Hi-

I think the docs mention to install v0.9.0 which is missing `--tracing-policy` and does not mention `--config-file`, I had to go by PRs to figure out. https://tetragon.cilium.io/docs/concepts/tracing-policy/ : would be nice if it also mentioned `--config-file` alongside it.

While I have you here I had another question, is there a way to pass and forget? Currently I am using something like ` sudo tetragon --config-file policy/tcp-connect.yaml` But this requires the terminal session to remain open, atleast on my machine it does not exit. At same time using the `tetra` client also requires `sudo tetragon` session to be open. There seems to be documentation regarding dropping configs but there is no mention of tracing polices. Ideally I could just run a script to install policies and that's it.

tixxdz:
> > > Hi, I am not wrong then currently there is no way to automatically load policies?
> > 
> > 
> > Hi, actually there is with `--config-file` that was converted to `--tracing-policy` to specify tracing policy file.
> > This series is to expend on it... just for curiosity what gave you this impression? or any feedback so we improve the docs?
> 
> Hi-
> 
> I think the docs mention to install v0.9.0 which is missing `--tracing-policy` and does not mention `--config-file`, I had to go by PRs to figure out. https://tetragon.cilium.io/docs/concepts/tracing-policy/ : would be nice if it also mentioned `--config-file` alongside it.

Alright we will try to improve it, https://github.com/cilium/tetragon/issues/1229  we also take patches  ;-)

> While I have you here I had another question, is there a way to pass and forget? Currently I am using something like ` sudo tetragon --config-file policy/tcp-connect.yaml` But this requires the terminal session to remain open, atleast on my machine it does not exit. At same time using the `tetra` client also requires `sudo tetragon` session to be open. There seems to be documentation regarding dropping configs but there is no mention of tracing polices. Ideally I could just run a script to install policies and that's it.

Indeed we have doc about config, and we will have soon same about tracing policies that applies to both k8s and non k8s environments.

For running tetragon you could use this: https://tetragon.cilium.io/docs/getting-started/deployment/package/  then just create the file `/etc/tetragon/tetragon.conf.d/config-file` with a content to a tracing policy , like you pass --config-file  and tetragon should load it and run it as a systemd service...

If you build from source you can also do:  `make tarball`  that will generate the latest development version, you will have a tarball that you can install, just follow same doc ;-)

If you can't install a systemd service, and just launch from directory, then use it with tmux or screen tools to handle the terminal session then combine it with  systemd lingering to auto start it... but the package deployment is the standard way.

If you encounter further problems please let us know in a new [issue](https://github.com/cilium/tetragon/issues), so we can keep comments here around the PR

much appreciated for your feedback ;-)

pkg/tracingpolicy/tracingpolicy.go
tixxdz: I have made the change to not display the author for now, so it is added as fields for now to allow loading only.

tixxdz:
@kkourt updated with a ci test to check if loading from /etc/tetragon/tetragon.tp.d/ works, thank you!

cmd/tetragon/main.go
```diff
@@ -372,6 +372,33 @@ func tetragonExecute() error {
 		if err != nil {
 			return err
 		}
+	} else {
```
kkourt, line 375: Please do not use an `else` here. Users should be able to load both from files and directories. (we should have the cli suporting multiple files and directories, but we can leave this as a followup). As I mentioned in another comment, please add an option for the directory.

cmd/tetragon/main.go
```diff
@@ -382,6 +409,31 @@ func tetragonExecute() error {
 	return obs.Start(ctx)
 }
 
+func addTracingPolicy(ctx context.Context, file string) error {
```
kkourt, line 412: since you are introducing a new helper, please make sure it's used for both files and directories.

kkourt (changes requested):

cmd/tetragon/conf.go
```diff
@@ -20,6 +20,8 @@ var (
 		"/usr/lib/tetragon/tetragon.conf.d/",
 		"/usr/local/lib/tetragon/tetragon.conf.d/",
 	}
+
+	admingTpConfDir = "/etc/tetragon/tetragon.tp.d/"
```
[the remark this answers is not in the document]
tixxdz: I was thinking about adding directory option as follow up, so how about this:

by default we read from /etc/tetragon/tetragon.tp.d/  (that's default will change admin prefix to default), then  if --tracing-policy-dir is specified we read instead from it?

Then we do this as a follow up.

cmd/tetragon/main.go
```diff
@@ -372,6 +372,33 @@ func tetragonExecute() error {
 		if err != nil {
 			return err
 		}
+	} else {
+		entries, err := os.ReadDir(admingTpConfDir)
+		if err == nil {
+			for _, entry := range entries {
+				files := make([]string, 0)
+				if entry.IsDir() {
```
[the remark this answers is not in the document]
tixxdz: not really, will do thanks ;-)

cmd/tetragon/main.go
tixxdz: Ok, I agree here too... but just didn't want to change lot of things or behaviour at once, please see my reply above.  I can make it now read from /etc/tetragon/tetragon.tp.d/*  and --tracing-policy  if that's fine ?

Users should just clear out /etc/tetragon/tetragon.tp.d/ to restore

cmd/tetragon/main.go
tixxdz: will do thank you

mtardy (commented):
Hey, for the doc change, I don't really understand why you moved the configuration back to its own file.

I moved these in the deployment guide, to explain how you can:
- Install
- Configure
- Upgrade
- Uninstall
Grouped for the package, container, and Kubernetes install.

So here you moved back only the configuration part about the package to the top sections and you redirect k8s users (I expect to be the majority of the users) to the Kubernetes configuration guide in the Kubernetes deployment?

cmd/tetragon/conf.go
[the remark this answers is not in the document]
kkourt: > by default we read from /etc/tetragon/tetragon.tp.d/ (that's default will change admin prefix to default), then if --tracing-policy-dir is specified we read instead from it?

Sure, let's just set the default value of `--tracing-policy-dir` to be `/etc/tetragon/tetragon.tp.d`.

> Then we do this as a follow up.

Please do this in this PR. It should be just a few lines of code.

cmd/tetragon/main.go
kkourt: I don't see why we would need a followup. This is a simple change. Let's just do it in this PR.

tixxdz:
> Hey, for the doc change, I don't really understand why you moved the configuration back to its own file.
> 
> I moved these in the deployment guide, to explain how you can:
> 
>     * Install
> 
>     * Configure
> 
>     * Upgrade
> 
>     * Uninstall
> 
> 
> Grouped for the package, container, and Kubernetes install.

Hmm I don't see it, the configuration is only in the package... ?  https://tetragon.cilium.io/docs/getting-started/deployment/package/#configuration ?  this change is fixing it unless there are some other links

> So here you moved back only the configuration part about the package to the top sections and you redirect k8s users (I expect to be the majority of the users) to the Kubernetes configuration guide in the Kubernetes deployment. Could you explain why such a change?

We first document configuration, then on how to update configuration it is straightforward, nothing to document except for k8s which is the only change here, and for k8s they can use helm and kubectl tools ? for others it is change config and restart so what am I missing here?

cmd/tetragon/conf.go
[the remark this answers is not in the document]
tixxdz: done

cmd/tetragon/main.go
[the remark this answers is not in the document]
tixxdz: changed it but I don't like it, I have explicitly to tell it to not walk the entire tree... and for performance this path doesn't really count

kkourt (changes requested):

docs/content/en/docs/reference/tetragon-configuration.md
```diff
@@ -0,0 +1,197 @@
...
+
+Tetragon default controlling settings are set during compilation, so configuration
+is only needed when it is necessary to deviate from those defaults. This
+document lists those controlling settings and how they can be set
+as a CLI arguments or as configuration options from [YAML](https://yaml.org) files.
+
+## Tetragon Controlling Settings
+
+Tetragon CLI arguments:
+```bash
```
mtardy, line 15:
```suggestion
```text
```

docs/content/en/docs/reference/tetragon-configuration.md
```diff
@@ -0,0 +1,197 @@
+---
+title: "Tetragon Daemon Configuration"
+linkTitle: "Tetragon Daemon Configuration"
```
mtardy, line 3:
```suggestion
title: "Agent configuration"
```
Maybe? most importantly we don't need to say Tetragon again here. And agent is maybe more common than daemon.

mtardy (commented):
So this PR includes a bunch of different stuff:
- a new package e2e tests
- loading tracing policies from a directory
- reorganization of documentation
- addition of author annotation on tracing policy

I think it's cool with the configuration in reference for now. Thanks, it's better than in getting started

Dockerfile
```diff
@@ -79,6 +79,8 @@ RUN curl -L https://github.com/libbpf/bpftool/releases/download/${BPFTOOL_TAG}/b
 FROM docker.io/library/alpine:3.18.2@sha256:82d1e9d7ed48a7523bdebc18cf6290bdb97b82302a8a9c27d4fe885949ea94d1 as base-build
 RUN apk add iproute2
 RUN mkdir /var/lib/tetragon/ && \
+    mkdir -p /etc/tetragon/tetragon.conf.d/ && \
+    mkdir -p /etc/tetragon/tetragon.tp.d/ && \
```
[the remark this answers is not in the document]
tixxdz: so we should ensure to create these tetragon paths during install package (already done) or other deployments to have same consistent code path in the agent and don't stat directories or guess current deployment, etc... , we already have /run , etc those are same... also this avoids creating them in the agent which will clash if a user deletes them. So consistency.

cmd/tetragon/flags.go
```diff
@@ -137,6 +138,8 @@ func readAndSetFlags() {
 
 	option.Config.EnablePidSetFilter = viper.GetBool(keyEnablePidSetFilter)
 
+	option.Config.TracingPolicyDir = viper.GetString(keyTracingPolicyDir)
```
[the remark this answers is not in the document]
tixxdz: I thought of it and opt not, cause abs and a like could fail , didn't want to change function signature, plus we have all the validate flags logic inside tetragonExecute() spreading it there and here sounds strange, so better keep this simple set only flags.

cmd/tetragon/main.go
```diff
@@ -133,6 +135,13 @@ func tetragonExecute() error {
 		log.Fatal(err)
 	}
 
+	option.Config.TracingPolicyDir = filepath.Join("/", option.Config.TracingPolicyDir)
```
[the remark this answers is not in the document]
tixxdz: The Abs if the path is not absolute it will be joined with current directory, and I opted for defensive. If a user by accident points to a relative path we explicitly join with / then force the Abs below, making it fail later during walkdir since those tracing policies are bpf programs that could impact the system... So to ensure that users are specifying their correct paths (including us developers ;-) )

cmd/tetragon/main.go
```diff
@@ -382,6 +390,77 @@ func tetragonExecute() error {
...
+		}
+	}
+
+	err := fs.WalkDir(tpFS, ".", func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if d.IsDir() {
+			if strings.Count(path, string(os.PathSeparator)) >= tpMaxDepth {
```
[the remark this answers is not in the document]
tixxdz: This could be common defensive practice to avoid walking nested trees , especially since this could be set as a configuration option that may point anywhere... recursing all directories without a limit could take time,  people may have volumes with symlinks, etc and we don't chase symlinks , so just making it 1 directory is simple for now... we can improve later if necessary.

cmd/tetragon/main.go
```diff
@@ -656,7 +656,7 @@ func startGopsServer() error {
 func execute() error {
 	rootCmd := &cobra.Command{
 		Use:   "tetragon",
-		Short: "Run the tetragon agent",
+		Short: "Tetragon - eBPF-based Security Observability and Runtime Enforcement",
```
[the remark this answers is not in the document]
tixxdz: Sure , yes I was lazy

docs/content/en/docs/getting-started/deployment/package.md
```diff
@@ -70,70 +70,9 @@ installed in `/usr/local/lib/tetragon/tetragon.conf.d/`. Local administrators
 can change the configuration by adding drop-ins inside
 `/etc/tetragon/tetragon.conf.d/` to override the default settings or use the
 command line flags. To restore default settings, remove any added configuration
-inside `/etc/tetragon/`. See the following section for details about the
-configuration precedence.
+inside `/etc/tetragon/tetragon.conf.d/`.
 
-### Configuration precedence
```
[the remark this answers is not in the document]
tixxdz: will do, thanks

.github/workflows/packages-e2e-tests.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Checkout Source Code
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+          submodules: true
+
+      - name: Getting version tag
+        id: tag
+        run: echo "tag=$(git describe --tags --always)" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
tixxdz: oh we have this, sure

cmd/tetragon/main.go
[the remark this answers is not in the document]
kkourt: Let's check if user provided an absolute path, (IsAbs) and then reject the file afterwards.

cmd/tetragon/main.go
```diff
@@ -133,6 +135,13 @@ func tetragonExecute() error {
 		log.Fatal(err)
 	}
 
+	option.Config.TracingPolicyDir = filepath.Join("/", option.Config.TracingPolicyDir)
+	tracingPolicyDir, err := filepath.Abs(option.Config.TracingPolicyDir)
```
kkourt, line 139: if you want to clean up the file (which we now is absolute), just call `.Clean()`

cmd/tetragon/main.go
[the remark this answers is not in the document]
tixxdz: done, thanks ;-)

docs/content/en/docs/reference/tetragon-configuration.md
tixxdz: done

docs/content/en/docs/reference/tetragon-configuration.md
tixxdz: Thanks, both seem fine.Tetragon to differentiate from tetra

cmd/tetragon/main.go
tixxdz: done

tixxdz:
@mtardy @kkourt fixed all the comments, thank you for the review

kkourt (approved):

kkourt:
Thanks!

willfindlay (approved):

mtardy (approved):

PR #997, opened 2023-05-15
tests: filter unit tests by PID + fix pidSet bugs

willfindlay:
Fix pidSet filter so that it properly accounts for children of children. Update observer_test_helper to support adding export filters on protojson and update existing tests to filter by PID with a pidSet filter.

### DIFF
--- a/pkg/filters/pidSet.go
+++ b/pkg/filters/pidSet.go
@@ -18,52 +18,85 @@ import (
 	"context"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/cilium/tetragon/pkg/logger"
 	v1 "github.com/cilium/tetragon/pkg/oldhubble/api/v1"
 	hubbleFilters "github.com/cilium/tetragon/pkg/oldhubble/filters"
 )
 
-/* pidSet is the set of pids we are interested in receiving events
- * for. This includes the pids listed in filter and any of their
- * children.
- */
-var pidSet map[uint32]bool
+// We could use an LRU here but we really don't want to evict old entries and risk failing
+// a test that uses this filter. Instead, we take the safer approach from the perspective
+// of testing and opt to grow the map indefinitely and log a warning if the size exceeeds
+// a pre-determined threshold. Since we have protections in place to prevent this filter
+// being used in production, this should be acceptable.
+type ChildCache = map[uint32]struct{}
 
-func filterByPidSet(pids []uint32) hubbleFilters.FilterFunc {
-	return func(ev *v1.Event) bool {
-		process := GetProcess(ev)
-		if process == nil {
-			return false
-		}
-		for _, pid := range pids {
-			if pid == process.Pid.GetValue() {
-				pidSet[pid] = true
-				return true
-			}
-		}
-		parent := GetParent(ev)
-		if parent == nil {
-			return false
-		}
-		if pidSet[parent.Pid.GetValue()] == true {
+func checkPidSetMembership(pid uint32, pidSet []uint32, childCache ChildCache) bool {
+	// Check the original pidSet. The reason for doing this separately is that we never
+	// want to drop the original pidSet from the cache. Keeping this separately in a slice
+	// is an easy way to achieve this.
+	for _, p := range pidSet {
+		if pid == p {
 			return true
 		}
-		for _, pid := range pids {
-			if pid == parent.Pid.GetValue() {
-				pidSet[pid] = true
-				return true
-			}
-		}
+	}
+	// Fall back to childCache to check children.
+	_, ok := childCache[pid]
+	return ok
+}
+
+func doFilterByPidSet(ev *v1.Event, pidSet []uint32, childCache ChildCache, childCacheWarning *int) bool {
+	process := GetProcess(ev)
+	if process == nil {
 		return false
+	}
+
+	// Check the process against our cache
+	pid := process.Pid.GetValue()
+	if checkPidSetMembership(pid, pidSet, childCache) {
+		return true
+	}
+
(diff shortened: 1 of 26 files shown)


pkg/filters/pidSet.go
```diff
@@ -20,50 +20,76 @@ import (
...
+
+func filterByPidSet(pidSet []uint32, childrenCache *ChildrenCache) hubbleFilters.FilterFunc {
+	return func(ev *v1.Event) bool {
+		return eventIsInPidSet(ev, pidSet, childrenCache)
+	}
 }
 
 type PidSetFilter struct{}
 
 func (f *PidSetFilter) OnBuildFilter(_ context.Context, ff *tetragon.Filter) ([]hubbleFilters.FilterFunc, error) {
```
kkourt, line 83: Do we need a flag for enabling the children cache? Do we need to track children in all cases?
I was not aware of this filter to be honest, but I'd be surprised if by default it filtered children as well. Especially since the children filtering might have some intricacies (see previous comment.). So let's document the behavior (when we figure it out) and add a flag if we need to.

kkourt (changes requested):
Thanks! I think that makes sense but I believe we need to figure out the details.

pkg/filters/pidSet.go
```diff
@@ -20,50 +20,76 @@ import (
...
 	hubbleFilters "github.com/cilium/tetragon/pkg/oldhubble/filters"
+	lru "github.com/hashicorp/golang-lru/v2"
 )
 
-/* pidSet is the set of pids we are interested in receiving events
- * for. This includes the pids listed in filter and any of their
- * children.
- */
-var pidSet map[uint32]bool
+type ChildrenCache = lru.Cache[uint32, struct{}]
```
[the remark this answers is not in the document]
willfindlay: Right that's a fair point. To be honest we don't have anybody currently running this filter and I'm not certain that there would even be a production use case where doing PID-based matching even makes sense. I think the intention here was to always use this in debugging/testing.

pkg/filters/pidSet.go
willfindlay: If you don't want to track children then you would simply use the pid filter rather than the pidSet filter. My understanding based on reading the existing implementation was that it was always the intention to have pidSet track children but the implementation was simply not working as intended.

If we do think people might use this in production, then yes I think it definitely makes sense to document the behaviour and potentially provide a flag for configuring cache size. But as I said I think the goal here is to have something useful for debugging/testing and not so much for a production environment.

pkg/filters/pidSet.go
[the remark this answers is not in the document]
kkourt: I fully agree that this should not be used in production! It's only useful for testing/debugging.

pkg/filters/pidSet.go
willfindlay: > So my question then is, is there a way for end-users to use this filter?

As of right now, no besides what we expose through GetEventsRequest API. This can change if needed.

kkourt: I think that's OK. Thanks for checking it out!

willfindlay: @kkourt I think I misread the question and thought you were asking about whether the cache size would be configurable. The filter could in theory be used through our configmap, same as the others. I can add a check to block this though.

pkg/filters/pidSet.go
```diff
@@ -22,48 +22,84 @@ import (
...
-		if pidSet[parent.Pid.GetValue()] == true {
+const (
+	// Maximum number of keys in a ChildCache before it fills up and we stop adding new
+	// entries.
+	maxChildCacheSize = 8192
+)
+
+func checkPidSetMembership(pid uint32, pidSet []uint32, childCache ChildCache) bool {
+	// Check the original pidSet. The reason for doing this separately is that we never
+	// want to drop the original pidSet from the LRU. Keeping this separately in a slice
```
kkourt, line 40: s/LRU/cache

pkg/filters/pidSet.go
```diff
@@ -22,48 +22,84 @@ import (
...
 		}
-		for _, pid := range pids {
-			if pid == parent.Pid.GetValue() {
-				pidSet[pid] = true
-				return true
-			}
-		}
+	}
+	// Fall back to childCache to check children.
+	_, ok := childCache[pid]
```
kkourt, line 48: how about:
```
// we reached maximum size, which means that we cannot insert any more pids in thet set, so we always return // true to account for the lost pids.
if len(childCache) == maxChildCacheSize {
              return true
}
```

pkg/filters/pidSet.go
willfindlay: Fixed in next draft

pkg/filters/pidSet.go
kkourt: OK, but then for an acceptList, as things stand now, we might loose events, which might lead to flakes. Or am I missing something?

Maybe pass a configuration to the filter so that we make it reliable in both cases?

willfindlay: Yeah that's fair. I actually think my old approach using an LRU was maybe slightly better for this, since you'd need 8192 actively running children to start dropping events.

Anyway it's not an issue to make this configurable, so I'll just do it.

willfindlay: @kkourt I added a configuration knob for the cache size. FWIW I think 8192 is probably gonna be acceptable in most cases, but now we can at least adjust as needed.

kkourt (approved):
Discussed offline.
Let's try doubling the map if we run out of space.
Worst case scenario, let's just add a loud warning that things might not work as expected if we hit the limit.

Issue #651, opened 2023-01-23
tracingpolicy: deprecate --config-file and replace it with something related to CRD policies


tixxdz:
The --config-file option is misleading as it refers to tracingpolicies and CRDs, should be deprecated and replaced.

PR #998, opened 2023-05-15
Deprecate the `--config-file` flag, replace with `--tracing-policy` and merge `pkg/config` with `pkg/tracingpolicy`

mtardy:
Fixes https://github.com/cilium/tetragon/issues/651.

`--config-file` will not appear in the help anymore but `--tracing-policy` will take its place. However, it will continue to work, just issuing a warning on startup when using it, recommending to switch to the new flag.


cmd/tetragon/flags.go
```diff
@@ -129,4 +129,14 @@ func readAndSetFlags() {
 	option.Config.ReleasePinned = viper.GetBool(keyReleasePinnedBPF)
 	option.Config.EnablePolicyFilter = viper.GetBool(keyEnablePolicyFilter)
 	option.Config.EnablePolicyFilterDebug = viper.GetBool(keyEnablePolicyFilterDebug)
+
+	// manually handle the deprecation of --config-file
```
kkourt, line 133: Should we add a date and a version here so that we know when to fully remove it?

kkourt (approved):
LGTM, thanks!

cmd/tetragon/flags.go
mtardy: Yep maybe we can say (as we discussed offline):
- deprecation for next minor release, maybe `0.10.0`
- removal for the next minor release after that, maybe `0.11.0`

mtardy: @kkourt I think it would be reasonable to have a list of things that we are planning to deprecate/remove for version N so that we can check when we make the release that we are on track. Would creating an umbrella issue for deprecation make sense?

PR #1006, opened 2023-05-17
docs: add an FAQ entry about BTF requirement

### LINKED ISSUE #911: What to do if `/sys/kernel/btf` doesn't exist?
The docs currently say that most distros have `/sys/kernel/btf`, but there is nothing that explain what to do if that is not the case. Also, it'd be helpful if docs did explain what BTF actually is, as there is no explanation of that either.

Having spent some time digging around the internet, I was able to find out that there is `CONFIG_DEBUG_INFO_BTF` kernel config option, which sounds like what I should be looking for... But I've found out about that option from some very random page, and I cannot be so sure if it's all I need, or perhaps there could be more options that need checking. I have seen `CONFIG_PAHOLE_HAS_SPLIT_BTF` is another related option and I definitely have no idea what that might be for.

mtardy:
This fixes #911.

cc @errordeveloper

### DIFF
--- a/docs/content/en/docs/faq/_index.md
+++ b/docs/content/en/docs/faq/_index.md
@@ -90,4 +90,44 @@ If you encounter the issue under Linux:
   persistently change the value of the CC go environment variable you can use
   `go env -w "CC=clang"`.
 
+#### Tetragon failed to start complaining about a missing BTF file
+
+You might have encountered the following issues:
+```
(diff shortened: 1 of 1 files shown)


tpapagian (commented):
LGTM, thanks!

tpapagian (approved):

PR #1007, opened 2023-05-17
docs: reorganize and simplify docs structure with new "concepts" section

mtardy:
Pull various categories under the "concepts" section to simplify the readability of the section. Also remove the current Tracing Policy documentation from the reference because it's not a reference doc.


kkourt (approved):
Thanks!

kkourt:
Ah, needs rebase :(

PR #1008, opened 2023-05-17
Add sock and skb matchArgs selectors

kevsecurity:
Tetragon can specify arguments of type sock and skb, but doesn't have the ability to match on these (complex) arguments. This series adds selector operators: SAddr, DAddr, SPort, DPort, and Protocol. These can be used to match on either sock or skb types. See commits for details.


jrfastab:
for the code LGTM.

jrfastab (approved):

jrfastab:
I wouldn't block this PR but MATCH_MAX_ARGS should probably be resolved at some point. One idea would be to use a read-only map and walk it with a proper for loop on newer kernels. Then you could have a map sized as large as needed to fit the args.

kevsecurity:
> I wouldn't block this PR but MATCH_MAX_ARGS should probably be resolved at some point. One idea would be to use a read-only map and walk it with a proper for loop on newer kernels. Then you could have a map sized as large as needed to fit the args.

I think the solution was to put the matchargs in a hash and then look up the arg to see if it's there. I'm going to convert all matchArgs over.

PR #1009, opened 2023-05-17
docs: add a script to export the generated API doc to references

mtardy:
So that the reference is integrated into the docs website and we don't have to put an external link there.

[Preview here](https://deploy-preview-1009--tetragon.netlify.app/docs/reference/grpc-api/).

### DIFF
--- a/api/Makefile
+++ b/api/Makefile
@@ -17,6 +17,7 @@ v1:
 		--user "$(shell id -u):$(shell id -g)" \
 		$(BUILDER_IMAGE) \
 		make -C /src -f Makefile.protoc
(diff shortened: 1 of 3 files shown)


tixxdz (approved):
Thanks @mtardy !

Would it be possible later say to replace one section with a better crafted one by hand? I see you have some sed calls there so I guess we can?

mtardy:
> Would it be possible later say to replace one section with a better crafted one by hand? I see you have some sed calls there so I guess we can?

Yeah sure, you can do whatever you'd like in the script, I just try to keep it portable with UNIX commands available everywhere like sed.

Thanks for the review!

PR #1010, opened 2023-05-18
tetragon: Enable parallel build for bpf objects

olsajiri:
To speed up the build, stolen from perf tool makefiles.

By default the bpf build is now parallel with -j<nr_cpus> option.

It's possible to specify -j argument with JOBS variable, which can be also used to switch it off with 'make JOBS=1'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,12 @@ GO_OPERATOR_IMAGE_LDFLAGS="-X 'github.com/cilium/tetragon/pkg/version.Version=$(
 GOLANGCILINT_WANT_VERSION = 1.52.2
 GOLANGCILINT_VERSION = $(shell golangci-lint version 2>/dev/null)
 
+# Do a parallel build with multiple jobs, based on the number of CPUs online
+# in this system: 'make -j8' on a 8-CPU system, etc.
+#
+# (To override it, run 'make JOBS=1' and similar.)
+#
+JOBS ?= $(shell nproc)
 
 .PHONY: all
 all: tetragon-bpf tetragon tetra tetragon-alignchecker test-compile tester-progs protoc-gen-go-tetragon tetragon-bench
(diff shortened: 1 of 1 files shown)


tpapagian (approved):
LGTM!

Makefile
```diff
@@ -49,6 +49,17 @@ GO_OPERATOR_IMAGE_LDFLAGS="-X 'github.com/cilium/tetragon/pkg/version.Version=$(
 GOLANGCILINT_WANT_VERSION = 1.52.2
 GOLANGCILINT_VERSION = $(shell golangci-lint version 2>/dev/null)
 
+# Do a parallel build with multiple jobs, based on the number of CPUs online
+# in this system: 'make -j8' on a 8-CPU system, etc.
+#
+# (To override it, run 'make JOBS=1' and similar.)
+#
+ifeq ($(JOBS),)
+  JOBS := $(shell (getconf _NPROCESSORS_ONLN || grep -E -c '^processor|^CPU[0-9]' /proc/cpuinfo) 2>/dev/null)
```
mtardy, line 58: What's the difference with nproc?

mtardy (approved):
That's nice!!!

Makefile
olsajiri: mostly that I had no idea nproc existed ;-) seems like better choice, I'll update the PR, thanks

mtardy: Oh that's nice then!! Cool :) thanks Jiri

mtardy: You can directly -j $(nproc) then 😺

olsajiri: I still want a way to switch it off with JOBS=1 which is handy when you debug build failure
or you want to spread the load differently

Makefile
```diff
@@ -49,6 +49,17 @@ GO_OPERATOR_IMAGE_LDFLAGS="-X 'github.com/cilium/tetragon/pkg/version.Version=$(
...
+# in this system: 'make -j8' on a 8-CPU system, etc.
+#
+# (To override it, run 'make JOBS=1' and similar.)
+#
+ifeq ($(JOBS),)
+  JOBS := $(shell nproc)
+  ifeq ($(JOBS),0)
+    JOBS := 1
+  endif
+endif
```
mtardy, line 62:
```suggestion
JOBS ?= $(shell nproc)
```
I think you can simplify this a bit

mtardy (approved):

Makefile
olsajiri: true ;-)

PR #1012, opened 2023-05-19
tetragon: Use do_task_dead probe for exit sensor

olsajiri:
Hooking on do_task_dead kernel function, which is the last one the 
task would execute after exiting. It's stable since v4.19, so it's
safe to hook for us. 

To find out if we are the last thread of execution in the task we
use current->signal->live counter (thanks Djalal! ;-) )

It's initialized for thread leader:

```
  clone {
    copy_process
      copy_signal
        atomic_set(&sig->live, 1); 
  }
```
Incremented for each new thread:

```
  clone {
    copy_process
      atomic_inc(&current->signal->live);
    ... 
    wake_up_new_task
  }
```

Decremented for each exiting thread:

```
  do_exit {
    atomic_dec_and_test(&tsk->signal->live);
    ... 
    do_task_dead
      __schedule
      BUG 
  }
```

If task->signal->live == 0 we are the last thread of execution and we
won't race with another clone, because there's no other thread to call
it (current thread is in do_exit).

### DIFF
--- a/bpf/process/bpf_exit.c
+++ b/bpf/process/bpf_exit.c
@@ -7,11 +7,57 @@
 
 char _license[] __attribute__((section("license"), used)) = "GPL";
 
-__attribute__((section("tracepoint/sys_exit"), used)) int
-event_exit(struct sched_execve_args *ctx)
+/*
+ * Hooking on do_task_dead kernel function, which is the last one the
+ * task would execute after exiting. It's stable since v4.19, so it's
+ * safe to hook for us.
+ *
+ * To find out if we are the last thread of execution in the task we
+ * use current->signal->live counter (thanks Djalal! ;-) )
+ *
+ * It's initialized for thread leader:
+ *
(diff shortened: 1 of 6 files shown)


tixxdz (approved):
Lgtm , thank you ;-)

bpf/process/bpf_exit.c
```diff
@@ -7,11 +7,57 @@
...
+ *     atomic_dec_and_test(&tsk->signal->live);
+ *     ...
+ *     do_task_dead
+ *       __schedule
+ *       BUG
+ *   }
+ *
+ * If task->signal->live == 0 we are the last thread of execution and we
+ * won't race with another clone, because there's no other thread to call
+ * it (current thread is in do_exit).
```
kkourt, line 47: Thanks for the nice comment :)

kkourt (approved):
Cool! Thanks!

PR #1013, opened 2023-05-19
Update cilium-builder image

ferozsalam:
Picks up newer golang dependencies.


mtardy (approved):
Test failure seems unrelated but the `sensors/tracing` test got stuck in a deadlock at some point which is worrying :/

kkourt (approved):

Issue #1014, opened 2023-05-22
test: `TestExitCode` is flaky


mtardy:
For example, see:
```
=== RUN   TestExitCode
    jsonchecker.go:179: jsonTestCheck: opening: /tmp/tetragon.gotest.TestExitCode.1359798033.json
    exit_test.go:242: exitCode 0
    jsonchecker.go:138: JsonCheck (retry=0) took 11.44135ms
    jsonchecker.go:179: jsonTestCheck: opening: /tmp/tetragon.gotest.TestExitCode.1359798033.json
    exit_test.go:242: exitCode 0
    jsonchecker.go:138: JsonCheck (retry=0) took 6.2431ms
    jsonchecker.go:162: JsonCheck (retry=1) failed: JsonEOF: failed to match after 17 events: err:tetragon returns the exit code of the process uncorrectly. Retrying after 3s
    exit_test.go:242: exitCode 0
    jsonchecker.go:138: JsonCheck (retry=1) took 7.72733ms
    jsonchecker.go:162: JsonCheck (retry=2) failed: JsonEOF: failed to match after 20 events: err:tetragon returns the exit code of the process uncorrectly. Retrying after 3s
    exit_test.go:242: exitCode 0
    jsonchecker.go:138: JsonCheck (retry=2) took 7.97711ms
    exit_test.go:256: error: JsonTestCheck failed in retry cnt=2 and there were no new events from previous try: JsonEOF: failed to match after 20 events: err:tetragon returns the exit code of the process uncorrectly
    jsonchecker.go:179: jsonTestCheck: opening: /tmp/tetragon.gotest.TestExitCode.1359798033.json
    exit_test.go:242: exitCode 0
    jsonchecker.go:138: JsonCheck (retry=0) took 7.29139ms
    jsonchecker.go:162: JsonCheck (retry=1) failed: JsonEOF: failed to match after 21 events: err:tetragon returns the exit code of the process uncorrectly. Retrying after 3s
    exit_test.go:242: exitCode 42
    jsonchecker.go:138: JsonCheck (retry=1) took 8.69339ms
    jsonchecker.go:179: jsonTestCheck: opening: /tmp/tetragon.gotest.TestExitCode.1359798033.json
    exit_test.go:242: exitCode 125
    jsonchecker.go:138: JsonCheck (retry=0) took 7.96349ms
    jsonchecker.go:179: jsonTestCheck: opening: /tmp/tetragon.gotest.TestExitCode.1359798033.json
[...]

PR #1015, opened 2023-05-22
tests: rewrite pkg/sensors/exec:TestExitCode

mtardy:
Might fix https://github.com/cilium/tetragon/issues/1014.

The test was calling JsonTestCheck multiple times resulting in the deletion of the logs if one of the tests before the last one failed. Now test also uses the eventchecker framework and is simpler.

The test was also flaky, not sure this will fix the issue but I could reproduce the failure after a hundred runs with the previous code while I could not at the moment with this one after 200+ runs. At least we will have the failure logs if this test continues to fail in the future.


willfindlay (approved):
Thanks!

Issue #902, opened 2023-04-07
process:exit: ensure that the "status" process exit code field is always set


tixxdz:
Seems if a process is started before tetragon, then when its exits we may not get the process_exit->status field the exit code

tixxdz:
@RickDeb2004 really sorry for late reply on this, I missed it.

So the original issue was:  before I start tetragon, I do `sleep 10`, start tetragon then do the cli `tetra getevents`, I get this:
```
  "process_exit": {
    "process": {
      "exec_id": "OjcwNzg3MDAwMDAwMDo5NjIx",
      "pid": 9621,
      "uid": 1000,
      "cwd": "/home/tixxdz/work/station/code/src/github.com/tixxdz/tetragon",
      "binary": "/usr/bin/sleep",
      "arguments": "10",
      "flags": "procFS auid",
      "start_time": "2023-11-07T09:20:43.396990069Z",
      "auid": 1000,
      "parent_exec_id": "OjYwOTYwMDAwMDAwMDo5Mzk2",
      "cap": {},
      "tid": 9621,
      "process_credentials": {
        "uid": 1000,
        "gid": 1000,
        "euid": 1000,
        "egid": 1000,
        "suid": 1000,
        "sgid": 1000,
        "fsuid": 1000,
        "fsgid": 1000
      }
    },
    "parent": {
      "exec_id": "OjYwOTYwMDAwMDAwMDo5Mzk2",
      "pid": 9396,
      "uid": 1000,
      "cwd": "/home/tixxdz/work/station/code/src/github.com/tixxdz/tetragon",
      "binary": "/usr/bin/bash",
      "flags": "procFS auid",
      "start_time": "2023-11-07T09:19:05.126990053Z",
      "auid": 1000,
      "parent_exec_id": "OjIwMjAwMDAwMDAwMDo0NTAx",
      "cap": {},
      "tid": 9396,
      "process_credentials": {
        "uid": 1000,
        "gid": 1000,
        "euid": 1000,
        "egid": 1000,
        "suid": 1000,
        "sgid": 1000,
        "fsuid": 1000,
        "fsgid": 1000
      }
    },
    "time": "2023-11-07T09:20:53.408386527Z"
  },
  "time": "2023-11-07T09:20:53.408384856Z"
}
```

I don't get the `process_exit->status` field, reference here: https://tetragon.io/docs/reference/grpc-api/#processexit

However, I just tested on processes that start after tetragon and it we also miss the `process_exit->status`

[...]

PR #1016, opened 2023-05-22
Updated exit_test.go

RickDeb2004:


pkg/sensors/exec/exit_test.go
```diff
@@ -163,8 +163,15 @@ func TestExitZombie(t *testing.T) {
 	logWG := testPipes.ParseAndLogCmdOutput(t, nil, nil)
 	logWG.Wait()
 
-	if err := testCmd.Wait(); err != nil {
-		t.Fatalf("command failed with %s. Context error: %v", err, ctx.Err())
+	 err := testCmd.Wait(); if err != nil {
+		if exitError, ok := err.(*exec.ExitError); ok {
+			// Process exited with a non-zero exit code
+			exitCode := exitError.ExitCode()
+			fmt.Printf("Process exited with code: %d\n", exitCode)
```
jrfastab, line 170: I'm not following code here at the moment. If it was a Fatal error before can you explain how just removing the Fatal is OK? Also please add the explanation into the commit so when we read the git log going forward we understand what we did.

Thanks for looking into this.

RickDeb2004: ok @jrfastab

RickDeb2004: In the case of the  mentioned regarding the process_exit->status field not being available, it appears that the code may be terminating abruptly due to the error, preventing the rest of the program from executing and potentially causing other functionality to be affected.

By removing the Fatal error, I can allow the program to continue execution even if the process_exit->status field is not available. This could be a conscious decision made in the context of  specific application's requirements and error handling strategy.

RickDeb2004: I have added the explanation in the commit message

pkg/sensors/exec/exit_test.go
```diff
@@ -163,8 +163,15 @@ func TestExitZombie(t *testing.T) {
...
+	}else{
+		fmt.Printf("Error waiting for process to exit: %v\n", err)
```
kkourt, line 172: But here we will still have to `Fatal`, I think.
Also, please use t.Logf (or similar) instead of fmt.Printf

kkourt (changes requested):
Thanks!

Can you also squash the two commits into one?
Also, please use your real name in Signed-off-by. See: https://docs.cilium.io/en/latest/contributing/development/contributing_guide/#dev-coo.

RickDeb2004:
> Thanks!
> 
> This is the code that tests the exit code, so I don't see how it fixes #902.

review pls @kkourt

tixxdz:
Hi @RickDeb2004 I updated the original issue https://github.com/cilium/tetragon/issues/902 

As noted by @kkourt this change is wrong as it does only touch the tests, where we want to investigate: 1. why we don't have the `process_exit->status` , 2. then fix what we have to fix, and 3. add tests to ensure we don't break it again.

If you want to do it, then please explicitly update here!

Much appreciated for your help!

kkourt:
Drafting for now. Thanks!

Issue #1025, opened 2023-05-24
TestKprobeCloneThreads is flaky


kkourt:
E.g.,:

https://github.com/cilium/tetragon/issues/1014
https://github.com/cilium/tetragon/issues/1022
So that we can investigate and try to fix this

Issue #1022, opened 2023-05-24
test: `TestUprobeCloneThreads` is flaky


mtardy:
It looks like `pkg.sensors.tracing.TestUprobeCloneThreads` is flaky. It fails sometimes on various kernel versions. See
- https://github.com/cilium/tetragon/actions/runs/5063158326/jobs/9089625484?pr=1020#step:6:1170
- https://github.com/cilium/tetragon/actions/runs/5063160908/jobs/9089664071?pr=1021#step:6:1173
- https://github.com/cilium/tetragon/actions/runs/5054887831/jobs/9070570239#step:7:1158

Related artifacts:
[tetragon-vmtests-4.19-0-results.zip](https://github.com/cilium/tetragon/files/11552799/tetragon-vmtests-4.19-0-results.zip)
[tetragon-vmtests-5.10-0-results (1).zip](https://github.com/cilium/tetragon/files/11552800/tetragon-vmtests-5.10-0-results.1.zip)
[tetragon-vmtests-5.10-0-results (2).zip](https://github.com/cilium/tetragon/files/11552801/tetragon-vmtests-5.10-0-results.2.zip)

tixxdz:
fixed by https://github.com/cilium/tetragon/pull/1027

PR #1027, opened 2023-05-24
tetragon: Adding missing tid setup to HandleGenericEvent

olsajiri:
We set tid for process part of the event to the tid value
we received from kprobe/uprobe/tracepoint event.

We missed it in HandleGenericEvent which is called from
Retry eventcache callback.


tixxdz (approved):
Yes this fixes https://github.com/cilium/tetragon/issues/1025 and https://github.com/cilium/tetragon/issues/1022  thanks @olsajiri ;-)

tixxdz:
That internal.GetProcessCopy() inside HandleGenericEvent() may endup resetting the tid that was set before hence the failures.

I'm fine merging this as it is now... so we don't change too much layouts, then we can improve later

kkourt (approved):

mtardy (approved):

PR #1032, opened 2023-05-26
tests/e2e: dump info from all pods in the cluster on failure

willfindlay:
Refactor the info dump logic slightly to dump info from all pods in the cluster when tests
fail, not just Tetragon pods. This includes the output of kubectl logs and kubectl
describe.


kkourt:
Could we just run a sysdump? We already depend on cilium cli which supports running tetragon bugtool.

willfindlay:
> Could we just run a sysdump?

Probably, yes. I'll see if I can get what I need from a sysdump.

PR #1036, opened 2023-05-30
deps: add renovate to replace dependabot

mtardy:
- run renovate self-hosted on GitHub action with a workflow in order to use PostUpgradeCommands to commit generated files in deps update PRs.
- install docker CLI plugin buildx in the renovate container in order for our PostUpgradeCommands to work.
- use 'cilium-renovate' GitHub application in order to generate a fine grain permissions token.
- group some deps update together in a single PR.
- add a workflow to validate configuration change in PRs.

I had to merge https://github.com/renovatebot/github-action/pull/749 to make our commands work in the renovate container.

This will provide us a lot more control over our deps update bot, and a lot more configuration settings (since we are running it ourselves on top of changing from dependabot to renovate).

When the renovate deployment works, we can disable Dependabot.

### DIFF
--- a/install/kubernetes/test.sh
+++ b/install/kubernetes/test.sh
@@ -1,9 +1,15 @@
 #!/bin/bash
 
 set -e
+set -o pipefail
 shopt -s expand_aliases
 
+# make this script executable from anywhere
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd $SCRIPT_DIR
+
 alias helm='docker run --rm -v $(pwd):/apps alpine/helm:3.3.4'
+alias kubeval='docker run --rm -i garethr/kubeval:0.15.0@sha256:6962d8ecbb7839637667f66e6703e6ebaae0c29dfe93a31d9968fba4324c7b8d'
 helm dependency update .
 helm lint . --with-subcharts
 helm template tetragon . | kubeval --strict --additional-schema-locations https://raw.githubusercontent.com/joshuaspence/kubernetes-json-schema/master
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -1,0 +1,201 @@
+{
+  // self-hosted configuration
+  "username": "cilium-renovate[bot]",
+  "repositories": ["cilium/tetragon"],
+  "allowedPostUpgradeCommands": [
+    "^/tmp/install-buildx$", 
+    "^make codegen$", 
+    "^make generate$", 
+    "^install/kubernetes/test.sh$"
+  ],
+  // repository configuration
+  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
+  "extends": [
+    "config:base",
+    ":gitSignOff",
+    "helpers:pinGitHubActionDigests"
+  ],
+  // this ensures that the gitAuthor and gitSignOff fields match
+  "gitAuthor": "renovate[bot] <bot@renovateapp.com>",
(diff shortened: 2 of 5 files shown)


michi-covalent (approved):
🤖 ➡️ 🤖

install/kubernetes/test.sh
```diff
@@ -1,9 +1,15 @@
...
 set -e
+set -o pipefail
 shopt -s expand_aliases
 
+# make this script executable from anywhere
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+cd $SCRIPT_DIR
+
 alias helm='docker run --rm -v $(pwd):/apps alpine/helm:3.3.4'
+alias kubeval='docker run --rm -i garethr/kubeval:0.15.0'
```
kkourt, line 12: I'm always wary of using unofficial docker images.
Could we use binaries from https://github.com/instrumenta/kubeval/releases instead?

kkourt (commented):
Thanks!

Will all the changes be in a single commit?

install/kubernetes/test.sh
mtardy: Ah! I got it from https://www.kubeval.com/installation/, I thought it was "the official".

mtardy:
> Will all the changes be in a single commit?

I could split by files since we have workflows and renovate config but since this is the initial config it feels a bit weird to split down the process of creating this configuration in multiple commits. If you have a specific idea in mind, we can do anything :)

install/kubernetes/test.sh
mtardy: this project is not maintained by the way :/. It was mostly so that I can run in the PostUpgradeCommand without having to install stuff without Docker but we can also install stuff.

kkourt:
> > Will all the changes be in a single commit?
> 
> I could split by files since we have workflows and renovate config but since this is the initial config it feels a bit weird to split down the process of creating this configuration in multiple commits. If you have a specific idea in mind, we can do anything :)

Sorry, I meant whether the generated PRs will have a single commit.

kkourt (approved):
LGTM, thanks! 🚀 🎉 

At what point we can disable dependabot? 👿

install/kubernetes/test.sh
mtardy: Since it's not super clear if it's official or not I should pin the sha256 so that everything stays in place for sure.

mtardy:
> > > Will all the changes be in a single commit?
> 
> > 
> 
> > I could split by files since we have workflows and renovate config but since this is the initial config it feels a bit weird to split down the process of creating this configuration in multiple commits. If you have a specific idea in mind, we can do anything :)
> 
> 
> 
> Sorry, I meant whether the generated PRs will have a single commit.

Ah 😅. Yes generated PR are mono commit, even for grouping ones. But we can choose which one should be grouped and I chose:
- all GitHub actions
- go deps from api/go.mod

mtardy:
> At what point we can disable dependabot? 👿

Once we merge this and see this working I guess 😄! Let's delay the merging by a day since I'm traveling today and won't be available if there's an issue.

install/kubernetes/test.sh
mtardy: see https://github.com/cilium/tetragon/pull/1036/files#diff-04ccead65ce7bcd231732b9157e93b204f96488a6812f347c9cdc5f070cb028bR12

willfindlay (approved):

PR #1047, opened 2023-06-02
bugtool: dump the policyfilter map

kkourt:


mtardy (approved):
Yep indeed, CLI building is only a static build so it makes sense that it fails to build this. CGO is hardcoded-disabled in the make rule. Unfortunately, we have to split a lot, but as you explained in your commit message, it makes sense.

PR #1049, opened 2023-06-02
Found unused functions

prateek041:
While searching for cilium dependencies in tetragon, I found some unused function in pkg/podinfo/podinfo.go and pkg/process/process.go, should we remove them ?

### DIFF
--- a/pkg/podinfo/podinfo.go
+++ b/pkg/podinfo/podinfo.go
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of Tetragon
+
 package podinfo
 
 import (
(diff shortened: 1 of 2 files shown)


mtardy (commented):
So we have an [outstanding cleanup PR lying around](https://github.com/cilium/tetragon/pull/983) and we recently [merged another one](https://github.com/cilium/tetragon/pull/976). I guess it's more complicated to say if those functions are truly unused since they are part of packages and exported so they that can be imported by other stuff. If it's part of an exported package, with an exported name, it can:
- be here to implement an interface
- just for people to use in different projects

I think the linters we have and the ones we are supposed to add in the future PR that is not merged are pretty good at finding the unused code that is unexported and everything.

prateek041:
oh, so you mean that they might not be in use right now, but be used in future or by someone else ?

PR #1051, opened 2023-06-05
Tracing: Add optional labels to argument

kevsecurity:
When using generic kprobes to observe kernel functions with multiple reported arguments, it can sometimes be tricky to identify what each argument actually represents. For example, on
__cgroup_bpf_run_filter_skb, the argument in index 2 is an int and is reported as {"int_arg":0} or {"int_arg":1} depending on context. It would be helpful if a user could specify a label for this argument so that it can be more easily understood and more easily automatically processed.

This commit adds an optional label parameter to arguments specified in the tracing policy, which will be output with the value. The output for the above example would now look like: {"int_arg":0,"Label":"send"} or {"int_arg":1,"Label":"send"}, making it obvious that the first is not sending (therefore receiving) and the second is sending.


jrfastab:
Lgtm also there was a request once to add a label to the entire event. So that something behind tetragon could some easier post processing

PR #1061, opened 2023-06-07
tetragon: Remove unused cgo code for reading perf events

olsajiri:
We switched to cilium/ebpf for reading perf ring buffer some
time ago [1], but we did not remove the old ring buffer code,
removing it now.

Keeping just PerfEventConfig type used by observer, attr objects
used by map_linux.go and few helper functions used by other code.

[1] 6dd4ee1022eb ("pkg/observer: remove old runEvents code")


mtardy:
we might need to remove https://tetragon.cilium.io/docs/faq/#go-compiler-is-complaining-about-missing-symbols-in-pkgbpf and simplify tetra compilation on Linux.

kevsecurity (approved):
If it passes all the tests, LGTM!

mtardy (approved):
Nice to remove that! Thanks!

> we might need to remove https://tetragon.cilium.io/docs/faq/#go-compiler-is-complaining-about-missing-symbols-in-pkgbpf, we can leave it here for now but remove it in the future, and also simplify tetra compilation on Linux.

If you can just open an issue when this merges so that we don't forget about these items. 😇

olsajiri:
> If you can just open an issue when this merges so that we don't forget about these items. innocent

sure ;-)  https://github.com/cilium/tetragon/issues/1077

kkourt (approved):
🚀

Issue #955, opened 2023-04-27
Add arm64 CI runners for Go tests and e2e tests


mtardy:
We are planning to use [buildjet arm64 runners](https://buildjet.com/for-github-actions). Waiting on the CNCF for progress at the moment. This will allow us to detect when we break arm64 compatibility in the CI.

PR #1062, opened 2023-06-07
Add arm64 runners for go tests

mtardy:
Fix https://github.com/cilium/tetragon/issues/955


mtardy:
I need to add e2e and it's good.

willfindlay (approved):

PR #1071, opened 2023-06-12
feat: Creating a Custom Resource Definition for a pod IP to metadata mapper.

prateek041:
The entire issue requires implementation of **Custom Resource** that maps **pod IP** to **metadata** and eventually remove **cilium endpoint custom resource**. The issue resolution is to be done in four parts, this is the first step where I Create the Custom resource definition.

Next up, writing the controller logic.


michi-covalent:
lgtm overall. a few comments

- i'm having a second thought regarding the resource format. maybe it makes sense to just mirror k8s pod resource format instead of coming up with our own. for example:
```
apiVersion: cilium.io/alphav1
kind: PodInfo
metadata:
  annotations:
    pod: annotation
  labels:
    pod: label
  name: pod-name
  namespace: pod-namespace
spec:
  hostNetwork: false
status:
  podIP: 1.2.3.4
  podIPs:
  - ip: 1.2.3.4
  startTime: "2023-05-28T12:09:02Z"
```
- still trying to come up with a better name than `PodInfo`. i don't like that plural is `podinfoes`. it doesn't sound right. not a blocker but we might end up changing the name later.
- remove `podinfo/go.mod`. we don't need to add a separate go module. also remove 	`github.com/onsi/ginkgo/v2` and `github.com/onsi/gomega` dependencies. tetragon uses `github.com/stretchr/testify`. we should avoid introducing additional test frameworks.

prateek041:
- [ ] Replicate the original k8s pod Resource format instead of creating one of our own. (needs discussion)
- [x] Remove go.mod file along with ginkgo and gomega.

michi-covalent:
> You should be able to define your own plural, e.g., podsInfo. That's what we did for TracingPolicyNamespaced's plural being TracingPoliciesNamespaced rather than TracingPolicyNamespaceds. See: https://github.com/cilium/tetragon/commit/2ebd385724d8e2fc3c9402e3aee71cc51c160fc0.

yeah that's also an option. we ended up going with `TetragonPod` as the name for now. open to suggestions though, we can always rename later 📝

kkourt:
> yeah that's also an option. we ended up going with `TetragonPod` as the name for now. 

`TetragonPod` sounds good to me.

prateek041:
I made the changes suggested, please have a look.

Should I change the name to TetragonPod right away or later ?

cc @michi-covalent

podinfo/config/crd/bases/cilium.io.tetragon.cilium.io_podinfoes.yaml
```diff
@@ -33,19 +33,30 @@ spec:
           metadata:
             type: object
           spec:
-            description: PodInfoSpec defines the desired state of PodInfo. It contains
-              the pod IP and metadata that will be populated by the operator.
+            properties:
+              hostNetwork:
+                description: Host networking requested for this pod. Use the host's
+                  network namespace. If this option is set, the ports that will be
+                  used must be specified.
```
tixxdz, line 40: Maybe a stupid comment: doc says "the ports that will be used must be specified" but doesn't specify where or how? maybe if it makes sense add it here too?

prateek041: Thank you for pointing that out. I will make the changes.

@tixxdz

prateek041:
Tasks:

[x] Rename the CRD to Tetragonpod.
[ ] Integrate deployment of CRD through the operator.

podinfo/config/crd/bases/cilium.io.tetragon.cilium.io_podinfoes.yaml
prateek041: I haven't mad the changes there yet, will do right before the PR is ready.

tetragonpod/api/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,36 @@
...
+package v1alpha1
+
+import (
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"sigs.k8s.io/controller-runtime/pkg/scheme"
+)
+
+var (
+	// GroupVersion is group version used to register these objects
+	GroupVersion = schema.GroupVersion{Group: "cilium.io.tetragon.cilium.io", Version: "v1alpha1"}
```
michi-covalent, line 29: group name should be https://github.com/cilium/tetragon/blob/main/pkg/k8s/apis/cilium.io/register.go#L7

tetragonpod/PROJECT
```diff
@@ -0,0 +1,20 @@
...
+projectName: tetragonpod
+repo: github.com/cilium/tetragon
+resources:
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: tetragon.cilium.io
+  group: cilium.io
+  kind: PodInfo
```
michi-covalent, line 17: rename PodInfo to TetragonPod

tetragonpod/api/v1alpha1/client/register_crd.go
```diff
@@ -0,0 +1,43 @@
...
+	apiextclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/yaml"
+)
+
+var scopedLogger = logger.GetLogger()
+
+func RegisterCRD(client *apiextclientset.Clientset) error {
+	crd := GetCRD()
+	_, err := client.ApiextensionsV1().CustomResourceDefinitions().Create(context.TODO(), crd, metav1.CreateOptions{})
```
michi-covalent, line 18: this will fail if the CRD already exists. we need to call Update() if the CRD already exists.

tetragonpod/api/v1alpha1/groupversion_info.go
prateek041: Kubebuilder is giving Group value from group and domain flags, together `group.domain`.

While I was setting up kubebuilder using `kubebuilder init` command, I provided domain to be `tetragon.cilium.io` as suggested and when created the API I using `kubebuilder create api` provided `--group cililum.io`, as a result of which it is generating the group to be `cilium.io.tetragon.cilium.io` (the issue above).

Now, when I am changing things in `api/v1alpha1/groupversion_info.go`, it is not reflecting the changes in CRD yaml when running `make manifests`, Should I change the things in [Project config](https://github.com/prateek041/tetragon/blob/f02a0ee440967b3151b675a5aceca39fb8e616e5/tetragonpod/PROJECT) of my project ? It clearly suggests me not to. [ref](https://book.kubebuilder.io/reference/project-config.html)

@michi-covalent

tetragonpod/api/v1alpha1/client/register_crd.go
prateek041: oh thanks for pointing that out, I totally missed that. Working on it.

tetragonpod/api/v1alpha1/groupversion_info.go
prateek041: I have fixed the main CRD yaml, by changing the comment at [here](https://github.com/prateek041/tetragon/blob/f02a0ee440967b3151b675a5aceca39fb8e616e5/tetragonpod/api/v1alpha1/groupversion_info.go#L19). This fixed the issue in the main CRD, I still see that [rbac yaml filees](https://github.com/prateek041/tetragon/tree/f02a0ee440967b3151b675a5aceca39fb8e616e5/tetragonpod/config/rbac) still use old group and Kind in many places. how do I fix that ?

michi-covalent: the worst case you can just search and replace.

tetragonpod/api/v1alpha1/client/register_crd.go
prateek041: I made changes, the current implementation is like this:

- The CRD is created only if it doesn't already exists.
- If the CRD already exists, no changes are made.

I am having a few questions regarding the update logic, like how would it know that CRD needs to be updated ? I went through the TracingPolicy CRD logic, how it is handling it, but have a few doubts. Let's have a little discussion on it in the meeting today.

tetragonpod/PROJECT
prateek041: done

tetragonpod/api/v1alpha1/groupversion_info.go
prateek041: fixed

prateek041:
- [x] Delete unused files
- [x] Resolve merge conflicts
- [x] Logic to enable tetragonpod through helm

cc @michi-covalent

michi-covalent:
@prateek041 just an fyi i pushed a copy of this branch to https://github.com/cilium/tetragon/pull/1281 so that we can test the CI image build 🧑‍🔬

Issue #1075, opened 2023-06-13
tetragon-oci-hook can only use /var/run/cilium/tetragon/tetragon.sock


kkourt:
tetragon-oci-hook has the grpc address of the agent hardcoded:
https://github.com/cilium/tetragon/blob/7017fc3db34e779f09330c0c7f61c86c71ee37d3/contrib/rthooks/tetragon-oci-hook/main.go#L30

Instead, tetragon-oci-hook should read a configuration file where the address is specified.

PR #1076, opened 2023-06-13
tetragon-oci-hook: improvements

kkourt:
This PR:
*  adds flags to tetragon-oci-hook, that allows to configure its behavior.
* adds initial support for cri-o
* adds support for configurable failure based on annotations

See commits.

Fixes: https://github.com/cilium/tetragon/issues/1075


jrfastab (approved):
Seems its been sitting for awhile. It looks ok to me though. I think PTO may be the reason its parked?

PR #1078, opened 2023-06-13
pids: clear process.parent_exec_id if process kernel is tetragon generated

tixxdz:
Tetragon generates processes with pid == 0 and pushes them into bpf execve_map and user space process cache to easy development, in case a kprobe is triggered without a process context... this makes it easy to support without hacking too much the code.

This situation made some tools show that pid == 0 is from <kernel> which is nice, however same tools may break since the generated event will have process.exec_id == process.parent_exec_id

Lets clear the process.parent_exec_id as suggested by William Findlay when returning the event, this allows to keep current logic in tetragon without modifying cache handling and offer a solution to other tools to easily identify that this is a tetragon kernel generated event that does not have a parent.

### DIFF
--- a/pkg/api/processapi/processapi.go
+++ b/pkg/api/processapi/processapi.go
@@ -29,6 +29,10 @@ const (
 
 	// MsgUnixSize of msg
 	MsgUnixSize uint32 = 640
+
+	// Special values to identify entries related to kernel in the process cache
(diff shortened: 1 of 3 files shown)


tixxdz:
This is a bit speedy fix, since didn't confirm what will be the side effect on kernel executing user mod helpers, coredump handlers etc...

Issue #1083, opened 2023-06-15
Null bytes in process binary names


willfindlay:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

We noticed some null bytes in process binary names where there shouldn't be any. Reproducible by letting Tetragon run for a while under a simple test workload and watching the metrics. You end up with binary names like the following when viewing the mtrics in vim:

```
tetragon_events_total{binary="^@^@^@^@^@^@^@^@^@^@^@^@^@",namespace="",pod="",type="PROCESS_EXEC"} 1
tetragon_events_total{binary="^@^@^@^@^@^@^@^@^@^@^@^@^@",namespace="",pod="",type="PROCESS_EXIT"} 1
tetragon_events_total{binary="^@^@^@^@^@^@^@^@^@^@^@^@^@^@",namespace="",pod="",type="PROCESS_EXIT"} 1
```

### Tetragon Version

Main at tag eb69bdc84405547733ccabfc69b355ae20d0eaa3

### Kernel Version

Linux 5.15.0-69-generic x86_64

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

Issue #193, opened 2022-06-24
NULLs in tetragon data


jpegleg:
When tetragon is installed via helm as instructed, some fields are populated by null strings (as seen below) instead of the expected data:

```
[...]

PR #1090, opened 2023-06-18
Fix memory corruption bug

kkourt:
This PR includes a bug that fixes a memory corruption bug, and it includes some other minor patches as well. Please review patch-by-patch.

Fixes: https://github.com/cilium/tetragon/issues/1083
Fixes: https://github.com/cilium/tetragon/issues/193

### DIFF
--- a/pkg/bpf/map_linux.go
+++ b/pkg/bpf/map_linux.go
@@ -158,14 +158,17 @@ func OpenMap(name string) (*Map, error) {
 
 	info, err := GetMapInfo(os.Getpid(), fd)
 	if err != nil {
+		unix.Close(fd)
 		return nil, err
 	}
 
 	if info.MapType == 0 {
+		unix.Close(fd)
 		return nil, fmt.Errorf("Unable to determine map type")
 	}
 
 	if info.KeySize == 0 {
+		unix.Close(fd)
 		return nil, fmt.Errorf("Unable to determine map key size")
 	}
 
--- a/pkg/observer/observer_stats.go
+++ b/pkg/observer/observer_stats.go
@@ -47,38 +47,43 @@ func (s *statValue) DeepCopyMapValue() bpf.MapValue {
 	return v
 }
 
+func updateMapMetric(name string) {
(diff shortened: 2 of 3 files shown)


pkg/observer/observer_stats.go
```diff
@@ -41,44 +41,49 @@ func (s *statValue) GetValuePtr() unsafe.Pointer {
...
-				continue
-			}
-			sum := int64(0)
-			for cpu := int(0); cpu < runtime.NumCPU(); cpu++ {
-				sum += v.Value[cpu]
-			}
-			mapmetrics.MapSizeSet(m.Name, int(mapLink.MapInfo.MaxEntries), float64(sum))
-			mapLink.Close()
-			mapLinkStats.Close()
+	v, ok := value.DeepCopyMapValue().(*statValue)
```
kevsecurity, line 71: Do we need the deep copy here? It looks like we just sum the values. Could we cast instead and save a copy?

kevsecurity (approved):
LGTM. One comment unrelated, so feel free to ignore for now.

pkg/observer/observer_stats.go
kkourt: I think you are right.
The deepcopy was in the original code though, and maybe this is worth doing as a followup.

pkg/bpf/map_linux.go
```diff
@@ -158,14 +158,17 @@ func OpenMap(name string) (*Map, error) {
...
 		return nil, err
 	}
 
 	if info.MapType == 0 {
+		unix.Close(fd)
 		return nil, fmt.Errorf("Unable to determine map type")
 	}
 
 	if info.KeySize == 0 {
+		unix.Close(fd)
```
mtardy, line 171: I was wondering if there was a reason why we can't use `defer(unix.Close(fd))` here after `ObjGet(name)`?

mtardy: aaaah yeah sorry indeed!

PR #1099, opened 2023-06-19
encoder: use protojson encoder instead of encoding/json

willfindlay:
The standard json encoder does not work well with protobuf messages. For example, we may
experience issues during encoding/decoding such as [0]. To fix this, we need to switch
over to protojson. See commits.
[0]: https://github.com/golang/protobuf/issues/1433


jrfastab (approved):

pkg/encoder/encoder_test.go
```diff
@@ -489,3 +494,44 @@ func TestCompactEncoder_EncodeWithTimestamp(t *testing.T) {
...
+		78776406,
+		56343416,
+		68876713,
+		51156281,
+		45544244,
+		4011756,
+	} {
+		f.Add(n)
+	}
+	f.Fuzz(func(t *testing.T, seed int64) {
```
mtardy, line 510: 😻

kkourt (approved):

PR #1104, opened 2023-06-20
tetragon: Switch to clang 15

olsajiri:
Using new clang 15 to compile bpf code.


mtardy:
Saw this, just wanted to mention that it should be good now thanks to https://github.com/cilium/tetragon/pull/2777 :)

PR #1107, opened 2023-06-22
tetragon: Force configured kernel version in observerFindBTF

olsajiri:
Nicholas reported that we don't take into account kernel option
when checking for BTF data, fixing that.


kkourt (approved):
Thanks!

PR #1113, opened 2023-06-23
tetragon: Allow to override tracepoint argument type

olsajiri:
The syscall enter/exit tracepoints exports all arguments as 8 bytes
values (unsigned long). This might lead to problems when interpreting
smaller integers like for unsigned int for 'fd' values.

Adding a way to force type for given argument with 'Type' field.

And adding support for [u]int32 types.


mtardy (approved):
Thanks Jiri for this fix. Indeed while testing unit tests on arm64, I discovered that the tracepoints on `sys_enter_lseek` returned the `-1` provided to the syscall as fd as 2^32-1 on x86 and 2^64-1 on arm64. This makes it possible for Tetragon to cover those inconsistencies in the kernel tracepoints.

It will allow me to fix and merge https://github.com/cilium/tetragon/pull/1062.

olsajiri:
adding support for short args and more tests

kkourt (approved):

PR #1119, opened 2023-06-26
Issue 1108/replace the use of viper.bind flags with variables binding

### LINKED ISSUE #1108: Replace the use of `viper.BindPFlags` with variables binding
Replace all use of `viper.BindPFlags` with flag variables like here https://github.com/cilium/tetragon/pull/1106/commits/da795ea4b2890f51f2f922a19e194ac3926a0429 and add a static check on `BindPFlags`, like grep or something like that to prevent the use of this thing. Because:

> When using same local flag for different subcommands, binding to the global configuration will erase previous values with the value of the last command's flag registered. See https://github.com/spf13/viper/issues/233.

And this has been breaking flags for some time now.

Jack-R-lantern:
That PR is the one that changes viper.BindPFlags.

Fixed #1108

### DIFF
--- a/cmd/tetra/common/client.go
+++ b/cmd/tetra/common/client.go
@@ -11,12 +11,10 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/sirupsen/logrus"
-
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/defaults"
 	"github.com/cilium/tetragon/pkg/logger"
-	"github.com/spf13/viper"
+	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
 )
@@ -25,7 +23,7 @@ type daemonInfo struct {
 	ServerAddr string `json:"server_address"`
 }
 
-func getActiveServAddr(fname string) (string, error) {
+func readActiveServerAddress(fname string) (string, error) {
 	f, err := os.Open(fname)
 	if err != nil {
 		return "", err
@@ -41,43 +39,28 @@ func getActiveServAddr(fname string) (string, error) {
 }
 
 func connect(ctx context.Context) (*grpc.ClientConn, string, error) {
-	connCtx, connCancel := context.WithTimeout(ctx, viper.GetDuration(KeyTimeout))
+	connCtx, connCancel := context.WithTimeout(ctx, Timeout)
 	defer connCancel()
 
-	var conn *grpc.ClientConn
-	var serverAddr string
-	var err error
-
-	// The client cli can run remotely so to support most cases transparently
-	// Check if the server address was set
-	//   - If yes: use it directly, users know better
-	//   - If no: then try the default tetragon-info.json file to find the best
-	//       address if possible (could be unix socket). This also covers the
-	//       case that default address is localhost so we are operating in localhost
-	//       context anyway.
-	//       If that address is set try it, if it fails for any reason then retry
-	//       last time with the server address.
-	if viper.IsSet(KeyServerAddress) == false {
-		// server-address was not set by user, try the tetragon-info.json file
-		serverAddr, err = getActiveServAddr(defaults.InitInfoFile)
-		if err == nil && serverAddr != "" {
-			conn, err = grpc.DialContext(connCtx, serverAddr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
-		}
-		// Handle both errors
+	// resolve ServerAdress: if flag set by user, use it, otherwise try to read
+	// it from tetragon-info.json, if it doesn't exist, just use default value
+	if ServerAddress == "" {
+		var err error
+		ServerAddress, err = readActiveServerAddress(defaults.InitInfoFile)
+		// if address could not be found in tetragon-info.json file, use default
 		if err != nil {
+			ServerAddress = defaultServerAddress
+			logger.GetLogger().WithField("ServerAddress", ServerAddress).Debug("connect to server using default value")
+		} else {
 			logger.GetLogger().WithFields(logrus.Fields{
-				"InitInfoFile":   defaults.InitInfoFile,
-				"server-address": serverAddr,
-			}).WithError(err).Debugf("Failed to connect to server")
+				"InitInfoFile":  defaults.InitInfoFile,
+				"ServerAddress": ServerAddress,
+			}).Debug("connect to server using address in info file")
 		}
 	}
-	if conn == nil {
-		// Try the server-address prameter
-		serverAddr = viper.GetString(KeyServerAddress)
-		conn, err = grpc.DialContext(connCtx, serverAddr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
-	}
 
-	return conn, serverAddr, err
+	conn, err := grpc.DialContext(connCtx, ServerAddress, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
+	return conn, ServerAddress, err
 }
 
 func CliRunErr(fn func(ctx context.Context, cli tetragon.FineGuidanceSensorsClient), fnErr func(err error)) {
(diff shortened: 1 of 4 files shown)


kkourt:
@mtardy could you please take a look? You created the issue so you seem to have context.

cmd/tetra/common/flags.go
```diff
@@ -12,3 +14,13 @@ const (
...
 )
+
+var (
+	Color         string
+	Debug         bool
+	Output        string
+	Tty           string
+	ServerAddress string
+	Timeout       time.Duration
+	Retries       int
```
mtardy, line 25: It looks like some of those vars are unused, could we keep only the var used?

mtardy (commented):
Thanks indeed it removes the use of viper.Bind which is cool, thanks. However I have a few comments:

mtardy:
Can you rebase your PR on top of main as well so that the tests on vmtests will pass, I merged a PR to fix this recently :)

cmd/tetra/common/flags.go
Jack-R-lantern: sure i will remove unused variable

Jack-R-lantern:
@mtardy
While checking the code, I noticed that for ServerAddress, you are using viper.IsSet.
For this, I changed it from the existing string type to a string pointer type.

cmd/tetra/common/client.go
```diff
@@ -73,7 +72,7 @@ func connect(ctx context.Context) (*grpc.ClientConn, string, error) {
 	}
 	if conn == nil {
 		// Try the server-address prameter
-		serverAddr = viper.GetString(KeyServerAddress)
+		serverAddr = ServerAddress
```
mtardy, line 75: just as a note we could remove this useless variable but I'm okay to merge this as is! :)

mtardy (dismissed):
Thanks it looks good! Thank you for adding the feedback :)

cmd/tetra/main.go
```diff
@@ -42,10 +41,9 @@ func New() *cobra.Command {
 
 	addCommands(rootCmd)
 	flags := rootCmd.PersistentFlags()
-	flags.BoolP(common.KeyDebug, "d", false, "Enable debug messages")
-	flags.String(common.KeyServerAddress, "localhost:54321", "gRPC server address")
-	flags.Duration(common.KeyTimeout, 10*time.Second, "Connection timeout")
-	flags.Int(common.KeyRetries, 0, "Connection retries with exponential backoff")
-	viper.BindPFlags(flags)
+	flags.BoolVarP(&common.Debug, common.KeyDebug, "d", false, "Enable debug messages")
+	flags.StringVar(&common.ServerAddress, common.KeyServerAddress, "", "gRPC server address")
```
mtardy, line 45: ah sorry we will need a last fix, I just noticed by reading carefully that the default value `localhost:54321` value was removed here, not sure this is intentional?

cmd/tetra/common/client.go
```diff
@@ -57,7 +56,7 @@ func connect(ctx context.Context) (*grpc.ClientConn, string, error) {
 	//       context anyway.
 	//       If that address is set try it, if it fails for any reason then retry
 	//       last time with the server address.
-	if viper.IsSet(KeyServerAddress) == false {
+	if ServerAddress == "" {
```
mtardy, line 59: aaaah maybe you removed the default value because of this. I think @tixxdz worked on this, what do you think? Should we remove the default value `localhost:54321` from the flag, see this https://github.com/cilium/tetragon/pull/1119/files#r1379864446.

mtardy: sorry to delay this PR again @Jack-R-lantern, will just need a look from Djalal because he worked on the default gRPC address to put so let him just ack this :)

mtardy: as far as I understand it would not be okay to remove the default since for most `tetra` users, tetragon might not run locally and the file (tetragon-info.json) it's trying to read will not be available. But with the default value, this part of the code will never be triggered. Maybe we could try to read the file containing the server address and fall back to the default `localhost:54321` value on file read error?

mtardy: this part of the code was never reached in practice because `IsSet` returns true even if it comes from the default value apparently. See https://github.com/spf13/viper/issues/323

mtardy: I haven't tested it but I think `viper.IsSet` returned false for the default value. So we would need to test this and to get the same behavior as before we should change some stuff here.

Jack-R-lantern: I'll do some more thinking about that code as well

mtardy: @tixxdz could you take a look at this one :)?

mtardy (changes requested):
Really sorry for the delays here, Djalal seems busy, so I took a look and I find the code too complicated for what it wants to achieve, not your fault but the old code + this change are hard to read, so I suggest something like that:

```diff
diff --git a/cmd/tetra/common/client.go b/cmd/tetra/common/client.go
index 8d6d032e..237734a0 100644
--- a/cmd/tetra/common/client.go
+++ b/cmd/tetra/common/client.go
@@ -24,7 +24,7 @@ type daemonInfo struct {
 	ServerAddr string `json:"server_address"`
 }
 
-func getActiveServAddr(fname string) (string, error) {
+func readActiveServerAddress(fname string) (string, error) {
 	f, err := os.Open(fname)
 	if err != nil {
 		return "", err
@@ -43,42 +43,25 @@ func connect(ctx context.Context) (*grpc.ClientConn, string, error) {
 	connCtx, connCancel := context.WithTimeout(ctx, Timeout)
 	defer connCancel()
 
-	var conn *grpc.ClientConn
-	var serverAddr string
-	var err error
-
-	// The client cli can run remotely so to support most cases transparently
-	// Check if the server address was set
-	//   - If yes: use it directly, users know better
-	//   - If no: then try the default tetragon-info.json file to find the best
-	//       address if possible (could be unix socket). This also covers the
-	//       case that default address is localhost so we are operating in localhost
-	//       context anyway.
-	//       If that address is set try it, if it fails for any reason then retry
-	//       last time with the server address.
-	if ServerAddress != "" {
-		serverAddr = ServerAddress
-		conn, err = grpc.DialContext(connCtx, serverAddr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
-	} else {
-		serverAddr, err = getActiveServAddr(defaults.InitInfoFile)
-		if err == nil {
-			if serverAddr == "" {
-				serverAddr = defaultServerAddress
-			}
-			conn, err = grpc.DialContext(connCtx, serverAddr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
-			if err != nil {
-				logger.GetLogger().WithFields(logrus.Fields{
-					"InitInfoFile":   defaults.InitInfoFile,
-					"server-address": serverAddr,
-				}).WithError(err).Debugf("Failed to connect to server")
-			}
-		}
-		if conn == nil {
-			serverAddr = defaultServerAddress
-			conn, err = grpc.DialContext(connCtx, serverAddr, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
+	// resolve ServerAdress: if flag set by user, use it, otherwise try to read
+	// it from tetragon-info.json, if it doesn't exist, just use default value
+	if ServerAddress == "" {
+		var err error
+		ServerAddress, err = readActiveServerAddress(defaults.InitInfoFile)
+		// if address could not be found in tetragon-info.json file, use default
+		if err != nil {
+			ServerAddress = defaultServerAddress
+			logger.GetLogger().WithField("ServerAddress", ServerAddress).Debug("connect to server using default value")
+		} else {
+			logger.GetLogger().WithFields(logrus.Fields{
+				"InitInfoFile":  defaults.InitInfoFile,
+				"ServerAddress": ServerAddress,
+			}).Debug("connect to server using address in info file")
 		}
 	}
-	return conn, serverAddr, err
+
+	conn, err := grpc.DialContext(connCtx, ServerAddress, grpc.WithTransportCredentials(insecure.NewCredentials()), grpc.WithBlock())
+	return conn, ServerAddress, err
 }
 
 func CliRunErr(fn func(ctx context.Context, cli tetragon.FineGuidanceSensorsClient), fnErr func(err error)) {
```

You can save this into a file and git apply it to your branch, it should work

Jack-R-lantern:
@mtardy 
I'm fine. You don't have to worry about it.
I'll change it

mtardy (approved):
Thanks a lot for your patience, let's merge this, Djalal can always revert if anything is not good but I tested it and it's great, thanks!

Jack-R-lantern:
When I thought of 'revert', 'squash' seems to be a must.
I'll reflect it in the PR, thanks for the review.

mtardy:
Thanks again for your patience and work :)

PR #1161, opened 2023-06-27
Revert sigstore/cosign-installer update and ignore it

mtardy:


mtardy:
btw tests will fail because the image build is broken (this patch is fixing it) and it is using pull_request_target as trigger.

willfindlay (approved):
Ship it

PR #1180, opened 2023-07-03
Fix connect and datagram tracingpolicy examples

kevsecurity:
The tcp-connect.yaml and datagram.yaml examples include selectors that match on components of the sock and skb data structures, notably DAddr. The current release does not yet support these selectors, unintentionally breaking their use with the latest release. They should, however, work with latest main, should that be compiled from source.

This commit moves the current examples to new file names and replaces the existing examples with versions that should work with the current release.


tpapagian (approved):
LGTM!

PR #1181, opened 2023-07-03
Make ReturnArgAction optional

kevsecurity:
The ReturnArgAction was added to the KProbeSpec to allow actions to be taken on the return values from hooked functions. It was specified as an enumeration in order to constrain its use to the supported actions, namely: Post, TrackSock and UntrackSock. This has the unintended consequence of making it non-optional through K8S validation.

This commit makes the ReturnArgAction optional.

### DIFF
--- a/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml
+++ b/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml
@@ -186,11 +186,8 @@ spec:
                       - type
                       type: object
                     returnArgAction:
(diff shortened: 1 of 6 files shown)


olsajiri (commented):
could you please add the 'fixes: commit' tag for reference?

kevsecurity:
> could you please add the 'fixes: commit' tag for reference?

I don't think there's an issue for this; just a Slack thread.

olsajiri:
> > could you please add the 'fixes: commit' tag for reference?
> 
> I don't think there's an issue for this; just a Slack thread.

I mean the commit number that broke it

olsajiri (approved):

PR #1185, opened 2023-07-04
contrib: add a Docker images digest checker

mtardy:
Add a shell script and its workflow to automatically detect when we use arch-restricted Docker images when we could use multi-arch ones.


willfindlay (approved):

mtardy:
My only concern is that this thing gets rate limited by Dockerhub when trying to fetch the manifest from there (our images are on quay). Let's see how it behaves, it should run rarely given the triggering rules. Let's 🚢 !

PR #1187, opened 2023-07-05
Bump e2e jobs-app charts and fix workflow logs collection step

### LINKED ISSUE #1174: e2e tests workflow on arm64 is flaky
we are observing quite frequent failures, many examples in the recent PRs, even on main recently:
https://github.com/cilium/tetragon/actions/runs/5423284292/jobs/9861010056

mtardy:
Tries to fix #1174.


mtardy:
So it works on the first run but I have no idea if this fixed the flake because I could not reproduce it locally and this is a flake...

mtardy:
I will try to rerun the job a few times

mtardy:
I had 5 successful run in a row (first was cancelled because I saw the test passing, just wanted to speedup the rerun)

image

willfindlay (approved):

Issue #1194, opened 2023-07-07
windows binaries are missing the .exe extension


bakito:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Windows release artifacts are missing the .exe suffix in the releases tar.gz

All windows binaries should have the .exe suffix.

### Tetragon Version

v0.9.0

### Kernel Version

-

### Kubernetes Version

-

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mtardy:
Hey thanks for opening the issue, indeed it was missing and we had everything in place to make the fix super easy: https://github.com/cilium/tetragon/pull/1196.

PR #1196, opened 2023-07-07
tetra: add .exe suffix for Windows binaries

mtardy:
Fixes https://github.com/cilium/tetragon/issues/1194.

### DIFF
--- a/Makefile.cli
+++ b/Makefile.cli
@@ -30,6 +30,7 @@ cli-local-release: cli-clean
 				;; \
 			windows) \
 				ARCHS='arm64 amd64'; \
+				EXT='.exe'; \
 				;; \
 		esac; \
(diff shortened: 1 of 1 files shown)


Makefile.cli
```diff
@@ -30,6 +30,7 @@ cli-local-release: cli-clean
 				;; \
 			windows) \
 				ARCHS='arm64 amd64'; \
+				EXT='.exe' \
```
kaworu, line 33: nit: I would
```suggestion
				EXT='.exe'; \
```
just in case we add another line after this one at some point.

kaworu (approved):

Makefile.cli
mtardy: Fixed! Thanks!

PR #1200, opened 2023-07-07
fix issue when deleting policy and policyfilter is disabled.

kkourt:
In #1183, @tpapagian reported:

> The issue is that in pkg/policyfilter/dummy.go (which is used when
> policyfilter is disabled) we return an error on AddPolicy and DelPolicy,
> which in turn return an error in deleteTracingPolicy()
> (pkg/watcher/crd/watcher.go) and makes the sensor removal to fail. The
> same applies when we update an existing tracing policy.

And suggested:

> To fix that we simply return nil in AddPolicy and DelPolicy
> (pkg/policyfilter/dummy.go) which is reasonable as we don't have to take
> any action in the case where policyfilter is disabled.

However, is we rely on this return value to reject accept namespaced
policies or policies with pod selectors if policyfilter is disabled.

Instead, modify DelPolicy to return an error only if a NoFilterPolicyID
is returned. This means that there will be no error under normal
conditions, but we can catch the bogus case where an actual id is
passed to DelPolicy when policy filter is disabled.

This PR fixes the reported issue and adds tests for it. Please see commits.


tpapagian (approved):
LGTM, thanks! (maybe you can fix the PR title before merging)

Issue #1193, opened 2023-07-06
Tetragon panics when misusing the `const_buf` type


mtardy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### What happened?

Tetragon panics at `/pkg/sensors/tracing/generictracepoint.go:684` when trying to handle a generic tracepoint message if the type is badly configured to `const_buf`.

It would be nice to add a test since it's easy to reproduce this error easily.

### Reproduction steps
1. create this tracingpolicy file in `panic.yaml`:
   ```yaml
   apiVersion: cilium.io/v1alpha1
   kind: TracingPolicy
   metadata:
     name: "const-buf-panic"
   spec:
     tracepoints:
     - subsystem: "sched"
       event: "sched_process_exec"
       args:
       - index: 4
         type: const_buf
   ```
2. Start Tetragon `sudo ./tetragon --bpf-lib bpf/objs/ --tracing-policy panic.yaml`
3. Trigger any exec on the machine (type `ls` on another bash, or just open another bash)

### Tetragon Version

commit `79aa12a01b97ae30315c352712ff1dec12aea047`.

### Kernel Version

5.15.0-76-generic #83-Ubuntu SMP Thu Jun 15 19:21:56 UTC 2023 aarch64 aarch64 aarch64 GNU/Linux

### Relevant log output

```shell
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x172a7f0]

goroutine 83 [running]:
github.com/cilium/tetragon/pkg/sensors/tracing.handleGenericTracepoint(0x19b1600?)
        /home/mahe.linux/tetragon/pkg/sensors/tracing/generictracepoint.go:684 +0x600
github.com/cilium/tetragon/pkg/observer.HandlePerfData({0x40006ae240, 0x84, 0x84})
        /home/mahe.linux/tetragon/pkg/observer/observer.go:117 +0xb8
github.com/cilium/tetragon/pkg/observer.(*Observer).receiveEvent(0x40007d47e0, {0x40006ae240?, 0x183eda0?, 0x1cbac?})
        /home/mahe.linux/tetragon/pkg/observer/observer.go:130 +0xc4
github.com/cilium/tetragon/pkg/observer.(*Observer).runEvents.func1()
        /home/mahe.linux/tetragon/pkg/observer/observer.go:228 +0xd4
[...]

PR #1202, opened 2023-07-07
tetragon: do not allow users to set const_buf

kkourt:
732087036a7180d1647ec5350c7f624e4ed9d36c introduced override tracepoint argument types by users.

The const_buf type, however, requires special information for its decoding that is only set during automatic type detection.

This results in the agent panicking when this happens. Fix this by disallowing const_buf being defined by users. In addition add a nil check before the dereference to avoid panics in the same code in the future if thigns change.

Fixes: https://github.com/cilium/tetragon/issues/1193


kkourt:
CI failure seems to be https://github.com/cilium/tetragon/issues/1174

PR #1205, opened 2023-07-10
tetragon: Add killer sensor

olsajiri:
adding support to have fast way (when multi kprobe link is not available)
of attaching killer/override program to multiple syscalls

details in: https://github.com/olsajiri/tetragon/blob/killer/docs/content/en/docs/concepts/tracing-policy/selectors.md#notify-killer-action


pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpoliciesnamespaced.yaml
```diff
@@ -576,6 +576,9 @@ spec:
                       - generated_syscalls
                       - generated_ftrace
                       type: string
+                    validated:
+                      description: List was validated
+                      type: boolean
```
kkourt, line 595: So from what I can tell, this option means that no validation will be performed.
Maybe `ignore-validation` or `force-no-validation` is a better name?

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -315,3 +315,8 @@ type PodInfoList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []PodInfo `json:"items"`
 }
+
+type KillerSpec struct {
+	// syscalls to kill
```
kkourt, line 320: How about `description: syscalls where killer is executed in`?

AFAIU, these are the syscalls where we will load the killer and do the check on whether to "kill" the process or not.

api/v1/tetragon/tetragon.proto
```diff
@@ -433,6 +433,8 @@ message ProcessTracepoint {
     // Arguments definition of the observed tracepoint.
     // TODO: once we implement all we want, rename KprobeArgument to GenericArgument
     repeated KprobeArgument args = 6;
+    // Action performed when the kprobe matched.
```
kkourt, line 442: s/kprobe/tracepoint/?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -897,6 +898,95 @@ broken.
 
 Socket tracking is only available on kernel >=5.3.
 
+### Notify Killer action
+
+The `NotifyKiller` action notifies the killer program to kill or override syscall.
```
kkourt, line 903:
```suggestion
The `NotifyKiller` action notifies the killer program to kill or override a syscall.
```

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -897,6 +898,95 @@ broken.
...
+
+The specs needs to have `killer` program definition, that instructs tetragon to load
+`killer` program and attach it to specified syscalls.
```
kkourt, line 906:
```suggestion
the `killer` program and attach it to specified syscalls.
```

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -897,6 +898,95 @@ broken.
...
+      - operator: "In"
+        values:
+        - "/usr/bin/bash"
+      matchActions:
+      - action: "NotifyKiller"
+        argSig: 9
+```
+
+Note the `NotifyKiller` with killer program is meant to be used only on kernel versions
+with no support for fast attach of multiple kprobes (`kprobe_multi` link).
```
mtardy, line 967: Would it be possible to move that to the very beginning, after the "The `NotifyKiller` action notifies the killer program to kill or override syscall." sentence? Also, could we have a little more details on the why of this feature? Why this is interesting and maybe useful for the reader?

mtardy (approved):
Thanks, only docs nits and agree with Kornilios points above!! The way your commits are done makes this really easy to review, would love to learn to do the same.

pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpoliciesnamespaced.yaml
olsajiri: it's there to ensure the validation goes through just once, so it gets validated,
sets the flag so another sensor won't run it again, I'll update changelog with that, sry

mtardy: is there a way we could avoid implementation things to be visible/editable from the UI (the yaml)?

this is typically that could go into the `status` of the resource instead of the `spec`.

docs/content/en/docs/concepts/tracing-policy/selectors.md
olsajiri: ok, I'll add some description at the beginning, thanks

PR #1219, opened 2023-07-12
bugtool: detect bpftool path

tixxdz:
This helps to detect bpftool from agent context, and also allow users to specify it as command line switch to tetra cli.


pkg/bugtool/bugtool.go
```diff
@@ -43,6 +43,8 @@ type InitInfo struct {
 	MetricsAddr string `json:"metrics_address"`
 	GopsAddr    string `json:"gops_address"`
 	MapDir      string `json:"map_dir"`
+	Path        string `json:"path"` // Environment variable $PATH
```
kkourt, line 46: I don't think this is needed (or used). Can we remove it?

kkourt (changes requested):

pkg/bugtool/bugtool.go
tixxdz: done

kkourt (approved):
Thanks!

PR #1220, opened 2023-07-12
metrics: Switch from gauges to counters

lambdanis:
The following metrics:
* tetragon_notify_overflowed_events
* tetragon_ringbuf_perf_event_received
* tetragon_ringbuf_perf_event_lost
* tetragon_ringbuf_perf_event_errors

are conceptually counters, not gauges, so I changed the metric type to reflect
that. Also added a `_total` suffix to all of them, for compatibility with the
OpenMetrics standard. This is a breaking change, but it shouldn't be painful - the semantics of the metrics doesn't change, only the names.

Admittedly, it's also not super necessary - using plain Prometheus you can scrape and query existing metrics just fine. Maybe there are/will be other systems that are pedantic about the standard, then this change will bring some benefits, but otherwise it's kinda compatibility for its own sake. Please shout if you think it's not worth it, but IMO this change is innocent enough to do it :)

```release-note
The following gauge metrics: tetragon_notify_overflowed_events, tetragon_ringbuf_perf_event_received, tetragon_ringbuf_perf_event_lost, tetragon_ringbuf_perf_event_errors are now counter and have a _total suffix.
```

### DIFF
--- a/pkg/bench/summary.go
+++ b/pkg/bench/summary.go
@@ -51,14 +51,7 @@ func (s *Summary) Dump() {
 	}
 }
 
-func getGaugeValue(gauge prometheus.Gauge) int {
-	// Yep, this does seem to be the only way to read it.
-	var d dto.Metric
-	gauge.Write(&d)
-	return int(*d.Gauge.Value)
-}
-
-func getMetricValue(counter prometheus.Counter) int {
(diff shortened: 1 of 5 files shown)


mtardy (approved):
this is great, thanks!!

PR #1221, opened 2023-07-13
Clang image Dockerfile and workflow improvements and renovate settings tuning

mtardy:
Check individual commits for messages and changes.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -251,8 +251,19 @@
       "enabled": false
     },
     {
+      // do not update sigstore/cosign-installer as it breaks CI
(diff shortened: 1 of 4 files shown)


mtardy:
The workflow change will show skipped jobs, I don't think it's bothering since it will be rarely triggered:
image

However we can avoid that by splitting the jobs between two workflows files.

olsajiri (approved):

olsajiri:
the dockerfile clang change looks good, can't really tell for the github actions but it looks ok

kkourt (approved):

willfindlay (approved):

PR #1222, opened 2023-07-13
Infinite ports and CIDRs

kevsecurity:
This series adds functionality to socket matching for sock and skb types. Previously, ports and CIDRs could be specified, but only a limited number would be accepted, due to complexity issues surrounding checking members of lists. This series uses an argfilter map to specify ports so that theoretically any number could be specified, and a new addr4filter map to specify IPv4 CIDRs, that are matched using a longest prefix match TRIE map.

In addition, ports can now be specified as ranges using the 'min:max' format, and new operators SPortPriv and DPortPriv were added to specify the range of privileged ports, from 0 to 1023 inclusive. On top of that, all port and address operators now can be prefixed with the Not keyword to create an inverse match, e.g. NotSAddr, and these can be combined with their positive equivalents to remove entires from a larger set.

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -751,10 +751,14 @@ filter_inet(struct selector_arg_filter *filter, char *args)
 		break;
 	case op_filter_sport:
 	case op_filter_notsport:
+	case op_filter_sportpriv:
+	case op_filter_notsportpriv:
 		port = tuple->sport;
 		break;
 	case op_filter_dport:
 	case op_filter_notdport:
+	case op_filter_dportpriv:
+	case op_filter_notdportpriv:
 		port = tuple->dport;
 		break;
 	case op_filter_protocol:
@@ -770,6 +774,12 @@ filter_inet(struct selector_arg_filter *filter, char *args)
 	case op_filter_notsport:
 	case op_filter_notdport:
 		return filter_32ty_map(filter, (char *)&port);
+	case op_filter_sportpriv:
+	case op_filter_dportpriv:
+		return port < 1024;
+	case op_filter_notsportpriv:
+	case op_filter_notdportpriv:
+		return port >= 1024;
 	}
 
 #pragma unroll
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -15,6 +15,7 @@
 #include "user_namespace.h"
 #include "capabilities.h"
 #include "../argfilter_maps.h"
+#include "../addr_lpm_maps.h"
 #include "common.h"
 #include "process/data_event.h"
 
@@ -714,6 +715,32 @@ struct ip_ver {
 	u8 version : 4;
 };
 
+// use the selector value to determine a LPM Trie map, and do a lookup to determine whether the argument
+// is in the defined set.
+static inline __attribute__((always_inline)) long
+filter_addr4_map(struct selector_arg_filter *filter, __u32 addr)
+{
+	void *addrmap;
+	__u32 map_idx = filter->value;
+	struct addr4_lpm_trie arg;
+
+	addrmap = map_lookup_elem(&addr4lpm_maps, &map_idx);
+	if (!addrmap)
+		return 0;
+
+	arg.prefix = 32;
+	arg.addr = addr;
+
+	__u8 *pass = map_lookup_elem(addrmap, &arg);
+
+	switch (filter->op) {
+	case op_filter_saddr:
+	case op_filter_daddr:
+		return !!pass;
+	}
+	return 0;
+}
+
 /* filter_inet: runs a comparison between the IPv4 addresses and ports in
  * the sock or skb in the args aginst the filter parameters.
  */
@@ -780,24 +807,14 @@ filter_inet(struct selector_arg_filter *filter, char *args)
 	case op_filter_notsportpriv:
 	case op_filter_notdportpriv:
 		return port >= 1024;
+	case op_filter_saddr:
+	case op_filter_daddr:
+		return filter_addr4_map(filter, addr);
 	}
 
 #pragma unroll
 	for (i = 0; i < MAX_MATCH_VALUES; i++) {
 		switch (filter->op) {
-		case op_filter_saddr:
-		case op_filter_daddr: {
-			__u32 cidraddr = v[i * 2];
-			__u32 cidrmask = v[i * 2 + 1];
-			__u32 maskedaddr = addr & cidrmask;
-
-			if (cidraddr == maskedaddr)
-				return 1;
-			// placed here to allow llvm unroll this loop
-			j += 8;
-			if (j + 8 >= filter->vallen)
-				return 0;
-		} break;
 		case op_filter_protocol:
 			if (v[i] == protocol)
 				return 1;
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -750,9 +750,11 @@ filter_inet(struct selector_arg_filter *filter, char *args)
 		addr = tuple->daddr;
 		break;
 	case op_filter_sport:
+	case op_filter_notsport:
 		port = tuple->sport;
 		break;
 	case op_filter_dport:
+	case op_filter_notdport:
 		port = tuple->dport;
 		break;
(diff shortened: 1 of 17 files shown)


pkg/selectors/kernel.go
```diff
@@ -489,6 +524,11 @@ func ParseMatchArg(k *KernelSelectorState, arg *v1alpha1.ArgSelector, sig []v1al
 		if err != nil {
 			return fmt.Errorf("writeMatchValuesInMap error: %w", err)
 		}
+	case SelectorOpSport, SelectorOpDport:
+		err := writeMatchRangesInMap(k, arg.Values, argTypeU64)
```
kkourt, line 568: shouldn't `argTypeU64` be `ty` as do above?

pkg/selectors/kernel.go
```diff
@@ -363,7 +363,48 @@ func writeMatchValuesInMap(k *KernelSelectorState, values []string, ty uint32) e
...
+}
 
+func writeMatchRangesInMap(k *KernelSelectorState, values []string, ty uint32) error {
+	mid, m := k.newValueMap()
+	for _, v := range values {
+		var rangeVal [2]uint64
+		rangeStr := strings.Split(v, ":")
+		if len(rangeStr) > 2 {
+			return fmt.Errorf("MatchArgs value %s invalid: range should be 'min:max'", v)
+		} else if len(rangeStr) == 1 {
```
kkourt, line 403: It took me a while to parse this code. Can we add a comment like:
// In case a user defines a single value `val`, i.e., not a  range,  we treat it as a range of `val:val`.

pkg/selectors/kernel.go
```diff
@@ -363,7 +363,48 @@ func writeMatchValuesInMap(k *KernelSelectorState, values []string, ty uint32) e
 			return fmt.Errorf("Unknown type: %d", ty)
 		}
 		m[val] = struct{}{}
+	}
+	// write the map id into the selector
+	WriteSelectorUint32(k, mid)
+	return nil
+}
 
+func writeMatchRangesInMap(k *KernelSelectorState, values []string, ty uint32) error {
```
kkourt, line 396: We seem to, now, have two functions doing very similar things: `writeMatchRangesInMap` and `writeMatchValuesInMap`. Is there a reason we we can't use just one for both cases? I think it would simplify the code a lot.

pkg/selectors/kernel.go
```diff
@@ -489,6 +524,11 @@ func ParseMatchArg(k *KernelSelectorState, arg *v1alpha1.ArgSelector, sig []v1al
 		if err != nil {
 			return fmt.Errorf("writeMatchValuesInMap error: %w", err)
 		}
+	case SelectorOpSport, SelectorOpDport:
```
kkourt: What are  the type arguments that these operations can be applied to?
Can we do some checks? The type of the argument can be find via the `sig` array I believe.
It would be nice to reject policies that do not make sense at this point.

kkourt (changes requested):
Thanks! 

I went over the first patch, and overall looks good to me!

Most of them are nits, but the one that I'd like to be addressed if possible is checking the type of arguments and reject invalid policies (assuming that applies here).

As also discussed offline, it would be great if we could add some more tests for this new functionality.

pkg/selectors/kernel.go
kevsecurity: No because ty would be sock or skb. As these are structs, we force the type based on the operator.

pkg/selectors/kernel.go
kevsecurity: I was leaving the choice, but yep I can remove writeMatchValuesInMap.

pkg/selectors/kernel.go
```diff
@@ -363,7 +363,48 @@ func writeMatchValuesInMap(k *KernelSelectorState, values []string, ty uint32) e
...
+				}
+				rangeVal[idx] = i
+			default:
+				return fmt.Errorf("Unknown type: %d", ty)
+			}
+		}
+		if rangeVal[0] > rangeVal[1] {
+			rangeVal[0], rangeVal[1] = rangeVal[1], rangeVal[0]
+		}
+		for val := rangeVal[0]; val <= rangeVal[1]; val++ {
```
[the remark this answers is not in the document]
kevsecurity: Yep.

pkg/selectors/kernel.go
kevsecurity: Yep

pkg/selectors/kernel.go
kevsecurity: Yep.

docs/content/en/docs/concepts/tracing-policy.md
```diff
@@ -1408,6 +1408,21 @@ There are different types supported for each operator. In case of `matchArgs`:
...
+* NotSPortPriv
+* DPort
+* NotDPort
+* DPortPriv
+* NotDPortPriv
+* SAddr
+* NotSAddr
+* DAddr
+* NotDAddr
+* Protocol
```
mtardy, line 1425: Do you think you could document that?

Since this is adding a lot of specialized operators only for one arg type, it would be nice if people can understand the shortcuts like `SPortPriv` without having to read the commit message.

It looks like we could document the operators more since they have different meanings depending on the context (i.e. equal on struct file means path equality, this kind of stuff), especially those that only apply to very specific types.

kevsecurity: Yep, I can do that now.

bpf/process/types/basic.h
```diff
@@ -770,6 +774,12 @@ filter_inet(struct selector_arg_filter *filter, char *args)
 	case op_filter_notsport:
 	case op_filter_notdport:
 		return filter_32ty_map(filter, (char *)&port);
+	case op_filter_sportpriv:
+	case op_filter_dportpriv:
+		return port < 1024;
```
kkourt, line 811: Would be nice to have an LT,GT,LEQ,GEQ operators for ports. For one, we would not need to hardcode the privileged port. Also, it would make it easier to handle port ranges without inserting all the values in the map. Just a thought, not a blocker -- we can revisit this if needed.

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -3494,8 +3498,637 @@ spec:
...
+	readyWG.Wait()
+
+	tcpReady := make(chan bool)
+	go miniTcpNopServer(tcpReady)
+	<-tcpReady
+	addr, err := net.ResolveTCPAddr("tcp", "127.0.0.1:9919")
+	assert.NoError(t, err)
+	_, err = net.DialTCP("tcp", nil, addr)
+	assert.NoError(t, err)
+	//conn.Close()
```
kkourt, line 3583: Is commenting out the `Close()` intentional?

kkourt (approved):
This looks great, thanks! 🚀

bpf/process/types/basic.h
kevsecurity: I couldn't think of a use case for LT,GT,LEQ,GEQ on their own other than privileged ports. Typically port ranges are more useful.  For ex, >60000 is not going to be used much, but 60000-60999 might be.
I guess my concern with combining LT,GT,LEQ,GEQ to create ranges is where you want to specify two or more ranges of ports: e.g. 0-200, 30000-30099, 50100-50110. This either needs three selectors or needs a loop to walk three entries in a single selector. I think this is better served by a port lookup map, even if that map gets large – it's just memory, and each encoded port is notionally four bytes. I'd rather reduce CPU hit and reduce the number of selectors at the expense of memory I think, given we're talking small amounts and not megabytes.

pkg/sensors/tracing/kprobe_test.go
kevsecurity: Hmm, that is a relic already in main that I copied and pasted. I'll remove it.

bpf/process/types/basic.h
kkourt: FYI, @aspsk has a nice proposal about doing this: https://lore.kernel.org/bpf/20220907080140.290413-1-aspsk@isovalent.com/

PR #1224, opened 2023-07-13
tetragon: Check final size for data event

olsajiri:
Adding size check on receiving side of data event to make
sure we won't use incomplete data.

Also adding stats for data events to keep track of what's
happening in there.

### DIFF
--- a/pkg/observer/data.go
+++ b/pkg/observer/data.go
@@ -54,14 +54,20 @@ func add(r *bytes.Reader, m *dataapi.MsgData) error {
 	return DataAdd(m.Id, msgData)
 }
 
-func DataGet(id dataapi.DataEventId) ([]byte, error) {
-	data, ok := dataMap.Get(id)
+func DataGet(desc dataapi.DataEventDesc) ([]byte, error) {
+	data, ok := dataMap.Get(desc.Id)
 	if !ok {
-		return nil, fmt.Errorf("failed to find data for id: %v", id)
+		return nil, fmt.Errorf("failed to find data for id: %v", desc.Id)
 	}
 
-	dataMap.Remove(id)
-	logger.GetLogger().WithFields(nil).Tracef("Data message used id %v, data len %v", id, len(data))
+	dataMap.Remove(desc.Id)
+
+	// make sure we did not loose anything on the way through ring buffer
+	if len(data) != int(desc.Size-desc.Leftover) {
+		return nil, fmt.Errorf("failed to get correct data for id: %v", desc.Id)
+	}
+
+	logger.GetLogger().WithFields(nil).Tracef("Data message used id %v, data len %v", desc.Id, len(data))
 	return data, nil
 }
 
--- a/pkg/observer/data.go
+++ b/pkg/observer/data.go
@@ -68,8 +68,11 @@ func DataGet(desc dataapi.DataEventDesc) ([]byte, error) {
 	// make sure we did not loose anything on the way through ring buffer
 	if len(data) != int(desc.Size-desc.Leftover) {
 		DataEventMetricInc(DataEventBad)
+		DataEventMetricSizeBad(desc.Size)
 		return nil, fmt.Errorf("failed to get correct data for id: %v", desc.Id)
 	}
+
+	DataEventMetricSizeOk(desc.Size)
 
 	logger.GetLogger().WithFields(nil).Tracef("Data message used id %v, data len %v", desc.Id, len(data))
 	DataEventMetricInc(DataEventMatched)
--- a/pkg/observer/data.go
+++ b/pkg/observer/data.go
@@ -32,9 +32,11 @@ func DataAdd(id dataapi.DataEventId, msgData []byte) error {
 	data, ok := dataMap.Get(id)
 	if !ok {
 		dataMap.Add(id, msgData)
+		DataEventMetricInc(DataEventAdded)
 	} else {
 		data = append(data, msgData...)
 		dataMap.Add(id, data)
+		DataEventMetricInc(DataEventAppended)
 	}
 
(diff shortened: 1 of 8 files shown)


pkg/observer/data_stats.go
```diff
@@ -13,6 +13,13 @@ var (
 		Help:        "Data event statistics. For internal use only.",
 		ConstLabels: nil,
 	}, []string{"event"})
+
+	DataEventSizeHist = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name:        consts.MetricNamePrefix + "data_event_size_histogram",
```
lambdanis, line 18: I switched all metrics to use a `Namespace` field instead of concatenating the prefix in #1228, can you change this one too?

cmd/tetragon/flags.go
```diff
@@ -77,6 +77,8 @@ const (
 	keyEnablePidSetFilter = "enable-pid-set-filter"
 
 	keyEnableMsgHandlingLatency = "enable-msg-handling-latency"
+
+	keyEnableDataEventsSizeMetric = "enable-data-events-size-metric"
```
lambdanis, line 81: Is this flag needed?

From what I see, currently from all metrics only `LatencyStats` has to be explicitly enabled with a flag. Having a way to enable/disable metrics is useful, but having a separate flag for each of them doesn't seem scalable. If it's not necessary right now, then I would rather develop a more generic way to enable/disable metrics in the near future.

pkg/observer/data_stats.go
```diff
@@ -0,0 +1,65 @@
+package observer
```
lambdanis, line 1: SPDX header is missing here (a CI check for that would be nice, but that's unrelated to this PR).

pkg/observer/data_stats.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/cilium/tetragon/pkg/option"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	// Define a counter metric for data event statistics
+	DataEventStats = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   consts.MetricsNamespace,
+		Name:        "data_event_stats_total",
```
lambdanis, line 14:
```suggestion
		Name:        "data_events_total",
```

lambdanis (commented):
Looks good, I left some comments/suggestions for the userspace part.

cmd/tetragon/flags.go
olsajiri: any idea how bad/slow is the histogram observe call? I got the impression it's better if it's disabled by default

pkg/observer/data_stats.go
olsajiri: will fix, thanks

pkg/observer/data_stats.go
olsajiri: ok

pkg/observer/data_stats.go
```diff
@@ -0,0 +1,65 @@
...
+		Help:        "Data event statistics. For internal use only.",
```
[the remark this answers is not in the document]
olsajiri: ok, thanks

cmd/tetragon/flags.go
jrfastab: I'm ok with this the main problem with some of the metrics was cardinality of the labels. This doesn't look bad from that side.

pkg/observer/data.go
```diff
@@ -54,14 +54,19 @@ func add(r *bytes.Reader, m *dataapi.MsgData) error {
...
+func DataGet(desc dataapi.DataEventDesc) ([]byte, error) {
+	data, ok := dataMap.Get(desc.Id)
 	if !ok {
-		return nil, fmt.Errorf("failed to find data for id: %v", id)
+		return nil, fmt.Errorf("failed to find data for id: %v", desc.Id)
+	}
+
+	// make sure we did not loose anything on the way through ring buffer
+	if len(data) != int(desc.Size-desc.Leftover) {
+		return nil, fmt.Errorf("failed to get correct data for id: %v", desc.Id)
```
kkourt, line 70: Should we remove the id from the dataMap before we return here?

kkourt (approved):
Thanks!

pkg/observer/data.go
olsajiri: yes, I moved it up in the new push, thanks

cmd/tetragon/flags.go
olsajiri: ok, I removed the option

pkg/option/config.go
```diff
@@ -75,6 +75,8 @@ type config struct {
 	EnablePidSetFilter bool
 
 	EnableMsgHandlingLatency bool
+
+	EnableDataEventsSizeMetric bool
```
lambdanis, line 79: This seems unused now

pkg/observer/data_stats.go
```diff
@@ -0,0 +1,61 @@
...
+	DataEventStats = promauto.NewCounterVec(prometheus.CounterOpts{
+		Namespace:   consts.MetricsNamespace,
+		Name:        "data_events_total",
+		Help:        "The number of data events by type. For internal use only.",
+		ConstLabels: nil,
+	}, []string{"event"})
+
+	DataEventSizeHist = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Namespace:   consts.MetricsNamespace,
+		Name:        "data_event_stats_size",
```
lambdanis, line 23: The last nit, promise - can it be just `data_event_size`? Having "stats" in the metric name might be a bit confusing about what is actually measured IMO :)
```suggestion
		Name:        "data_event_size",
```

olsajiri: sure, np ;-)

PR #1228, opened 2023-07-14
Pass Namespace in metrics options instead of prefixing metric names

lambdanis:
This is a non-functional change.


tpapagian (approved):
LGTM, thanks!

PR #1233, opened 2023-07-14
docs update

kkourt:


mtardy (approved):
Thanks, the `sed` command makes it so much simpler.

If you ever wondered, this version is injected on this page https://deploy-preview-1233--tetragon.netlify.app/docs/getting-started/deployment/container/ for example. But it can be injected anywhere we need the latest tag.

PR #1245, opened 2023-07-18
vmtests: use latest kernels

kkourt:


kkourt:
Let's do https://github.com/cilium/tetragon/pull/1247 instead.

PR #1246, opened 2023-07-18
selector fixes

kkourt:
When setting multiple inner maps in argfilter_maps of different sizes,
loading the data into the maps fails for 5.4 and 4.19 kernels.

We attribute this failure to these kernels not having
https://lore.kernel.org/bpf/20200828011800.1970018-1-kafai@fb.com/,
which allows the outer map to be updated with an inner map of different
sizes. This change was introduced in 5.9

This PR:

1. Introduces a MaxEntriesInnerMap filed in the programs, which allows
   us to override the max entries defined in the bpf code. We follow
   this approach since there is already a MaxEntries field for a similar
   purpose.
2. Sets this field using SetMaxEntries in both generic kprobes and
   tracepoints. For this to work, we had to move the parsing of the
   kernel selectors earlier. This happens only for
   kerrnels < 5.9
3. Modifies the loading code so that it checks this entry when we load a
   new map.
4. Modifies the code that creates the inner map to use a single
   max_entry for all inner maps for kernels < 5.9.
5. Adds a test case in the selectors tests. For the test case to work,
   we had to modify the test to reload the programs every time/

### DIFF
--- a/pkg/sensors/tracing/selectors.go
+++ b/pkg/sensors/tracing/selectors.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 
 	"github.com/cilium/ebpf"
+	"github.com/cilium/tetragon/pkg/kernels"
 	"github.com/cilium/tetragon/pkg/selectors"
 	"github.com/cilium/tetragon/pkg/sensors"
 	"github.com/cilium/tetragon/pkg/sensors/program"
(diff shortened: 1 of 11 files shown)


pkg/sensors/tracing/selectors.go
```diff
@@ -101,8 +102,24 @@ func populateArgFilterMaps(
...
 ) error {
+	maxEntries := uint32(0)
+	if !kernels.MinKernelVersion("5.9") {
+		for _, vm := range k.ValueMaps() {
+			nentries := uint32(len(vm))
+			if nentries > maxEntries {
+				maxEntries = nentries
+			}
+		}
+	}
```
kevsecurity, line 113: Can't we use:
`maxEntries := k.ValueMapMaxExtries()`
here?

kevsecurity (approved):
One comment and other than the static complaints, LGTM

pkg/sensors/tracing/selectors.go
kkourt: yap, thanks!

willfindlay (approved):

PR #1247, opened 2023-07-18
gh/vmtests: use -main for latest version and add bpf-next

kkourt:
Update the gh action for vmtetss to use the latest version for each kernel, which is designated as "-main".


tpapagian (approved):
LGTM, thanks!

willfindlay (approved):

PR #1252, opened 2023-07-19
fix replace directives for Cilium v1.13.4

rolinh:
When the `cilium/cilium` dependency is updated, it's `go.mod` replace directives must be replicated.


tpapagian (approved):
Thanks!

pkg/k8s/go.mod
```diff
@@ -68,16 +68,10 @@ require (
...
 	github.com/optiopay/kafka => github.com/cilium/kafka v0.0.0-20180809090225-01ce283b732b
-	github.com/vishvananda/netlink => github.com/jrfastab/netlink v1.1.2
-
-	// Use a fork of lumberjack with patches to ensure compressed logs are created atomically
-	gopkg.in/natefinch/lumberjack.v2 => github.com/chancez/lumberjack v0.0.0-20220314160755-2b78c6a5f7bc
-
-	// Using private fork of controller-tools. See commit msg for more context
-	// as to why we are using a private fork.
-	sigs.k8s.io/controller-tools => github.com/christarazi/controller-tools v0.3.1-0.20200911184030-7e668c1fb4c2
+	go.universe.tf/metallb => github.com/cilium/metallb v0.1.1-0.20220829170633-5d7dfb1129f7
```
kkourt, line 75: 🏓 
We removed metallb https://github.com/cilium/tetragon/pull/1152.
This PR brought it back.
And then  we removed it again: https://github.com/cilium/tetragon/pull/1269

@rolinh is this something that is needed? Should we ensure that it is not removed again?

CC: @mtardy

mtardy: I just did the same thing twice without even realizing it...

rolinh: > @rolinh is this something that is needed? 

`go mod`/`go get` doesn't follow replace directives. If we don't replicate the replace directives from Cilium's `go.mod`, we end up using different versions of deps that the ones intended for the Cilium release we vendor which may cause subtle bugs.

mtardy: I'm sorry @rolinh, does that mean we should put it back or not?

kkourt: >  we end up using different versions of deps that the ones intended for the Cilium release we vendor which may cause subtle bugs.

Is this true even if we completely remove a dependency?

rolinh: > I'm sorry @rolinh, does that mean we should put it back or not?

Yes.

> Is this true even if we completely remove a dependency?

What do you mean by "completely remove a dependency"? If we stop vendoring Cilium, then we can safely stop replicating Cilium's replace directives. Or do you mean for the case where Tetragon doesn't rely directly on the dependency for which there is a replace directive? The tricky part is that we may or may not be affected, it all depends on how the code from the dependency that is used in Tetragon relies on its own dependencies for which there are replace directives.

kkourt: > Or do you mean for the case where Tetragon doesn't rely directly on the dependency for which there is a replace directive? 

That's what I mean. Tetragon does not use the replaced module (metallb) in this case.

PR #1253, opened 2023-07-19
Implement `DeleteTracingPolicy` functionality

### LINKED ISSUE #1110: Add `DeleteTracingPolicy` method to API

Currently it's possible to manage TracingPolicies with CRDs. They can be added or deleted and Tetragon seems to correctly handle both scenarios. In our case, however, preferred way of adding TracingPolicy is using `AddTracingPolicy` method of its API. But there is no `RemoveTracingPolicy` which makes it not possible to handle cases when one policy should be replaced with another. There is `RemoveSensor`, but it has different purpose.

I would like to be able to delete added TracingPolicy with  `RemoveTracingPolicy` method.

Add required method to API.

### Code of Conduct

inliquid:
- Implement `DeleteTracingPolicy` for gRPC service
- Implement "delete" command for `tetra` client
- Rename `DelTracingPolicy` -> `DeleteTracingPolicy` everywhere for consistency

Fixes #1110


mtardy (commented):
Hey 👋  thanks a lot for taking the time to write this PR, that's really nice of you!

If I may have a nit, would it be possible to keep `delete` instead of `del` everywhere (even in tetra)? I'm okay with having the whole `delete` word and changing this will unfortunately introduce a breaking change to the gRPC API (minimal I admit because it was not really exposed through a RPC but still).

mtardy:
Taking the need for the change to also warn you that you will need to add your "sign-off" to the commit. Just a formality:

It looks like your commit is not signed-off, to improve tracking of who did what, the Tetragon project uses a "sign-off" procedure which you can learn more about in the [Cilium project documentation](https://docs.cilium.io/en/latest/contributing/development/contributing_guide/#developer-s-certificate-of-origin).

<p>

```shell
```

```shell
git commit -s
```

```shell
git commit --amend -s
```

```
-s, --signoff, --no-signoff

   option on the command line.
```

</p>
</details>

mtardy:
And lastly, you can squash your commit and prefix the message with something like `gRPC: implement DeleteTracingPolicy functionality`.

I hope all of this is not too overwhelming, please feel free to ask any questions if needed. Again grateful that you took the time to implement this! :)

inliquid:
Hey @mtardy, thank you for detailed explanations, I think I've fixed these points, is there anything else I can do?

mtardy:
> Hey @mtardy, thank you for detailed explanations, I think I've fixed these points, is there anything else I can do?

It's all good let's run the CI and maybe ask for a review by Kornilios with who you interacted in the issue!

jrfastab (approved):

cmd/tetra/tracingpolicy/tracingpolicy.go
```diff
@@ -97,6 +108,21 @@ func addTracingPolicy(ctx context.Context, client tetragon.FineGuidanceSensorsCl
...
 
+func deleteTracingPolicy(ctx context.Context, client tetragon.FineGuidanceSensorsClient, yamlFname string) {
+	yamlb, err := os.ReadFile(yamlFname)
+	if err != nil {
+		fmt.Printf("failed to read yaml file %s: %s\n", yamlFname, err)
+		return
+	}
+
+	_, err = client.DeleteTracingPolicy(ctx, &tetragon.DeleteTracingPolicyRequest{
+		Yaml: string(yamlb),
```
kkourt, line 119: It seems that the `DeleteTracingPolicyRequest` takes the full yaml as an argument (this is probably my fault). I think it would make more sense to take the name as the argument.

So, basically, redefine `DeleteTracingPolicyRequest` to be:
```
message DeleteTracingPolicyRequest {
	string name = 1;
}
```

What do you think?

kkourt (commented):
Thanks!

I have a comment, please see bellow.

cmd/tetra/tracingpolicy/tracingpolicy.go
inliquid: I think it's very good observation, makes sense. Just one little question is that currently there is some logic for checking whether policy is [namespaced](https://github.com/inliquid/tetragon/blob/pr/inliquid/remove-tracing-policy-method/pkg/server/server.go#L259), and logging namespace. Apart from logging it looks like namespace is not used anywhere. So if we go for deleting tracing policy by name this logging will be removed as well. Is this ok?

mtardy: I think this is perfectly fine, it's still interesting to log the request though, which will only contain the name. It's just a debug trace if we need to see the gRPC handling in action.

mtardy: Keep something like
```go
	logger.GetLogger().WithFields(logrus.Fields{
		"metadata.name":      tp.TpName(),
	}).Debug("Received a DeleteTracingPolicy request")
```

inliquid: Fixed. Please note that "delete <sensor_name>" help string for tracing policy "delete" comment in `tetra` which is in line with sensor's "rm" command which is "rm <sensor_name>" as they effectively use same name on deletion, despite that sensors names are usually something like `gkp-sensor-1`.

inliquid: @mtardy I also squashed commits together as you suggested earlier, not sure if this is necessary.

cmd/tetra/tracingpolicy/tracingpolicy.go
```diff
@@ -97,6 +108,15 @@ func addTracingPolicy(ctx context.Context, client tetragon.FineGuidanceSensorsCl
 	}
 }
 
+func deleteTracingPolicy(ctx context.Context, client tetragon.FineGuidanceSensorsClient, name string) {
+	_, err := client.DeleteTracingPolicy(ctx, &tetragon.DeleteTracingPolicyRequest{
+		Name: name,
+	})
+	if err != nil {
+		fmt.Printf("failed to delete tracing policy: %s\n", err)
```
mtardy, line 116: A nit, sorry for the back and forth I just noticed that, a non-blocker because we do the same for `addTracingPolicy`. 

This is out of scope of this PR because the whole cmd need rework but: we should use `common.CliRunErr` in the `cobra.Command` and return an error to the `Run` function so that the command returns a non-0 return code on errors.

The most simple way of getting a non-0 return code here would maybe to use:
```go
logger.GetLogger().WithFields(logrus.Fields{"name": name }).WithError(err).Fatal("failed to delete tracing policy")
```

mtardy (commented):
Otherwise, for me, it's ready, any opinion @kkourt?

cmd/tetra/tracingpolicy/tracingpolicy.go
kkourt: > I think it's very good observation, makes sense. Just one little question is that currently there is some logic for checking whether policy is [namespaced](https://github.com/inliquid/tetragon/blob/pr/inliquid/remove-tracing-policy-method/pkg/server/server.go#L259), and logging namespace. Apart from logging it looks like namespace is not used anywhere. So if we go for deleting tracing policy by name this logging will be removed as well. Is this ok?

Short version: yes that's fine.

Long version: yes, that's fine because currently, we cannot have different policies with the same name, regardless of whether they are namespaced or not.

See https://github.com/cilium/tetragon/blob/f0ae9f0f27f0bfee0164d8996bfb90967fea489b/pkg/sensors/handler.go#L92

This means that the name uniquely identifies the policy (if it exists) so using just the name is fine.

I suspect that, in the future, we might want to change that to allow policies with the same name on different namespaces. When this happens, we can extend the request to include the namespace as well. 

> s in line with sensors "rm" command which is "rm <sensor_name>" as they effectively use same name on deletion

As another note, the sensors are just old interfaces that we should remove. We should only use tracing policies from now on.

kkourt (approved):
LGTM, thanks! 🚀

cmd/tetra/tracingpolicy/tracingpolicy.go
kkourt: Yeah, I think `Fatal` is the right thing to do if we want an error code. I'd say, let's merge the PR as it is and we can deal with above elsewhere.

mtardy: yep because overall the whole thing need a bit of rework for other functions that do return 0 on success.

mtardy (approved):

cmd/tetra/tracingpolicy/tracingpolicy.go
mtardy: created https://github.com/cilium/tetragon/issues/1262

cmd/tetra/tracingpolicy/tracingpolicy.go
inliquid: @kkourt 
> As another note, the sensors are just old interfaces that we should remove. We should only use tracing policies from now on.

We're using sensors to enable/disable policies, which is very useful when you want to just turn on/off some specific channel of information such as monitoring of network activity, as it may generate a lot of load. Would it still be possible to switch tracing policies on/off as easily as with sensors? We would reimplement our "controller" component to consume another gRPC methods if sensors had gone, but can this feature remain please?

kkourt: That's a good point, thanks for mentioning it. We should definitely do that. Created the following issue: https://github.com/cilium/tetragon/issues/1272

PR #1256, opened 2023-07-20
tetragon: improve how we handle TIDs and GetProcessCopy()

tixxdz:
Goes on top of https://github.com/cilium/tetragon/pull/1254


jrfastab:
So the argument here is exit should always be associated with the same TID as the execve?

tixxdz:
> So the argument here is exit should always be associated with the same TID as the execve?

Yes this is the intention, so to sum up  at execve tid==pid at our current hook time (late) there is only one single thread, then:

We improved how to track last threads https://github.com/cilium/tetragon/blob/main/bpf/process/bpf_exit.c#L10  so technically we can report during exit TID != PID (TGID)

But we still report TID==PID: https://github.com/cilium/tetragon/blob/main/bpf/process/bpf_exit.h#L59 

So if you see some value in reporting TID != PID at exit, we can make it and drop the last patch of this PR?

tid != pid at exit has some value but maybe only to expert users, plus the tid field is new!  in the other hand if users start to correlate the exceve and exit with PID and TID better have them equal, and that's it (simple).


For the related GetProcessCopy(): the long term goal is during kprobe if TID != PID then internally we just copy the shared fields during GetProcessCopy(), but for all the rest we have to set it at real time from BPF... this should reduce data copied during GetProcessCopy() but may add more cpu cycles for decoding non shared fields which we have to do anyway... and we already have an answer by masking some fields with --enable-process-cred --enable-process-ns flags to not decode nor report everything...

So here I will start incrementally isolating fields and making it clear in the code what fields we share and what not and assert that execve==exit events have some equal fields, but for other events we do case by case... to improve process cache handling and make it easy for us to remember :-D ...

tixxdz:
@kkourt @olsajiri opinion on above?

jrfastab:
Its not clear to me how to create "shared" fields until we rip the protobuf structs out of the core codes. I think we want to do this regardless though.

pkg/grpc/exec/exec.go
```diff
@@ -288,6 +288,15 @@ func GetProcessExit(event *MsgExitEventUnix) *tetragon.ProcessExit {
 	code := event.Info.Code >> 8
 	signal := readerexec.Signal(event.Info.Code & 0xFF)
 
+	// Ensure that we get PID == TID
+	if event.ProcessKey.Pid != event.Info.Tid {
+		logger.GetLogger().WithFields(logrus.Fields{
```
jrfastab, line 293: I'm tempted to say just remove these blocks. So it should never happen AFAICT and if it did happen in some rare instance I doubt debugging is enabled :/

pkg/grpc/exec/exec.go
```diff
@@ -306,10 +315,9 @@ func GetProcessExit(event *MsgExitEventUnix) *tetragon.ProcessExit {
 		parent.RefDec()
 	}
 	if proc != nil {
+		// Copy process before returning
 		tetragonEvent.Process = proc.GetProcessCopy()
```
jrfastab, line 319: Do we still need the copy then if we drop the UpdateEventProcessTid? We should be OK to just use the cached version then.

pkg/grpc/exec/exec.go
```diff
@@ -339,13 +347,11 @@ func (msg *MsgExitEventUnix) RetryInternal(ev notify.Event, timestamp uint64) (*
 	}
 
 	if internal != nil {
-		proc := internal.GetProcessCopy()
+		ev.SetProcess(internal.GetProcessCopy())
```
jrfastab, line 350: Same here if we are not doing a ProcessTid we can just use the pointer?

jrfastab:
@kkourt ProcessInternal can be mutable but *process can not be changed while being handled by JSON/GRPC writers or else we risk corrupting the output.

jrfastab:
I merged the patch 1 from the other PR as we discuss this.

kkourt:
> @kkourt ProcessInternal can be mutable but *process can not be changed while being handled by JSON/GRPC writers or else we risk corrupting the output.
> 
> And (not sure we do this now) ProcessInternal really shouldn't be written/read without using some atomic, read_once, or locking primitive because multiple go routines might be working on it. Or if its a new object not pushed into a cache yet its safe because it has just a single go routine with access to it.

Makes sense. Might worth reviewing the code and verifying that things work as above. ProcessInternal does have a `mu sync.Mutex` so all accesses need to be done with the mutex held.

pkg/grpc/exec/exec.go
tixxdz: I would agree, but we have the same bit in exec path and it helped uncover some paths that were missing between in some versions (variants) of tetragon... beside the bpf and userspace bits ;-)  , and for exec path it was not a debug message we made it an info or warn then we turned it to debug, after fixing all those...

For debugging enabled, we have a way now to turn it on during runtime, if it doesn't hurt lets please keep it as it could help point to the problem quickly in such cases...

tixxdz: hmm seems it was still warn() not debug, and I moved it a bit early but kept it warn() for now, see last patch. In my mind it was: run with warn() then turn into debug() after some time and I thought we already did...  anyway all warnings have been fixed but lets see...

pkg/grpc/exec/exec.go
tixxdz: yes done

tixxdz:
> > @kkourt ProcessInternal can be mutable but *process can not be changed while being handled by JSON/GRPC writers or else we risk corrupting the output.
> > And (not sure we do this now) ProcessInternal really shouldn't be written/read without using some atomic, read_once, or locking primitive because multiple go routines might be working on it. Or if its a new object not pushed into a cache yet its safe because it has just a single go routine with access to it.
> 
> Makes sense. Might worth reviewing the code and verifying that things work as above. ProcessInternal does have a `mu sync.Mutex` so all accesses need to be done with the mutex held.

@jrfastab @kkourt so that mutex is used to serialize:
1. retry / update `processinternal.process.pod`
2. get copy of `processinternal.process` to stream it to gRPC
3. annotate `processinternal.process` with namespace and capabilities

So not for processinternal but for the embedded `protobuf process struct`.

We have to do several things here:
1. Add another simple process struct inside processinternal and use it as the real holder instead of the protobuf one https://github.com/cilium/tetragon/blob/main/pkg/process/process.go#L42 
2. Always collect process credentials and namespace during exec/clone and store directly in this new simple form (caps as int64...) inside process cache, get rid of of these protobuf capabilities and namespaces https://github.com/cilium/tetragon/blob/main/pkg/process/process.go#L44  too. When exporting events we chose what fields we export... --enable-process-creds etc or have better fields filters...
* Decoding events to protobuf only when exporting events may add cpu overhead as now we decode once and use for both execve / exit , kprobes , etc, this also may do unnecessary GetProcessCopy() on *all* fields since we have to update the copies anyway for non main threads... (this is still in progress). so I think the schema below could work:

We may end up with:
```
// ProcessInternal is the internal representation of a process.
// nolint:revive // This is an acceptable case of "stuttering" since the name "Internal"
// wouldn't make much sense by itself.
type ProcessInternal struct {
        // internal process simple representation holder of the possible non shared information
        // that is the per thread info
        mainThread processSimple

	// muproto protects the modifications to cached proto process.
	muproto sync.Mutex
        // decoded protobuf fields externally visible process struct, decode once and serve many
	process *tetragon.Process

	// garbage collector metadata
	color  int
        refcnt uint32
}
```

* The per thread fields are stored in the mainThread object. When we operate on the main thread, execve/exit, kprobe, tracepoints etc we compare the thread ID, if main thread use the cached protobuf fields. If not main thread then decode the rest (non shared) and return.

This could make processInternal somehow immutable as pointed by @kkourt until we start to hook in the sensors call sites that may change those fields... 

@jrfastab I see you say it can be mutable is there some other reasons?

tixxdz:
issue here: https://github.com/cilium/tetragon/issues/1290

tixxdz:
> @kkourt ProcessInternal can be mutable but *process can not be changed while being handled by JSON/GRPC writers or else we risk corrupting the output.
> 

@jrfastab what are the scenarios where processInternal can be mutable? just ensuring I didnt miss any. IIRC the matchCapabilitiesChange updates the bpf execve_map and that's part of kprobe not a main sensor.

I think if we want to reflect this we have first to hook by default the setuid() family and update the credentials on the bpf execve_map then reflect this on the user space process cache too... we do it step by step, so we don't get tricked by some exploits etc...

olsajiri:
I'm probably missing something, but I'm getting bit lost on how the TID works now ;-)

IIRC we have following events flow:

```
- fork      (TID == PID)
- exec      (TID == PID)
- [ku]probe or tracepoint (we record TID and create extra copy of Process in user space with this TID for report)
- exit      (TID == PID)
```

could we have comment like that somewhere explaining the current situation?

also maybe we could remove some of the tid fields from the kernel->user events,
that gets initialized as pids anyway, seems confusing

tixxdz:
> I'm probably missing something, but I'm getting bit lost on how the TID works now ;-)
> 
> IIRC we have following events flow:
> 
> ```
> - fork      (TID == PID)
> - exec      (TID == PID)
> - [ku]probe or tracepoint (we record TID and create extra copy of Process in user space with this TID for report)
> - exit      (TID == PID)
> ```
> 
> could we have comment like that somewhere explaining the current situation?
> 
> also maybe we could remove some of the tid fields from the kernel->user events, that gets initialized as pids anyway, seems confusing

indeed will do

jrfastab:
Its not clear to me what the issue created is about? In general IMO we rely to heavily on protobufs in lower edges of stack, but that is about optimizing performance and json encoder. 

The mutex in processInternal is there to serialize writers/readers of processInternal. Somehow the 3 bullets above were lost on me.

tixxdz:
> > I'm probably missing something, but I'm getting bit lost on how the TID works now ;-)
> > IIRC we have following events flow:
> > ```
> > - fork      (TID == PID)
> > - exec      (TID == PID)
> > - [ku]probe or tracepoint (we record TID and create extra copy of Process in user space with this TID for report)
> > - exit      (TID == PID)
> > ```
>>
> >   
> > could we have comment like that somewhere explaining the current situation?

@olsajiri I did extensively last patch with commit log, please have a look 

> > also maybe we could remove some of the tid fields from the kernel->user events, that gets initialized as pids anyway, seems confusing

I agree that we should remove it as some point, but for now it is useful as it helps debug things and ensure our model is right, and we are still not finished with handling all threads please take a look at the last patch commit log and to this comment https://github.com/cilium/tetragon/pull/1256#discussion_r1278296380

pkg/grpc/exec/exec.go
```diff
@@ -288,6 +288,15 @@ func GetProcessExit(event *MsgExitEventUnix) *tetragon.ProcessExit {
...
 	signal := readerexec.Signal(event.Info.Code & 0xFF)
 
+	// Ensure that we get PID == TID
+	if event.ProcessKey.Pid != event.Info.Tid {
+		logger.GetLogger().WithFields(logrus.Fields{
+			"event.name":        "Exit",
+			"event.process.pid": event.ProcessKey.Pid,
+			"event.process.tid": event.Info.Tid,
+		}).Debug("ExitEvent: process PID and TID mismatch")
+	}
```
olsajiri, line 300: the comment says you want to ensure PID == TID, but there's just warning,
we don't set TID = PID as we do for exec event

pkg/grpc/exec/exec.go
```diff
@@ -306,10 +315,9 @@ func GetProcessExit(event *MsgExitEventUnix) *tetragon.ProcessExit {
 		parent.RefDec()
 	}
 	if proc != nil {
-		tetragonEvent.Process = proc.GetProcessCopy()
+		// Use cached version of process
+		tetragonEvent.Process = proc.UnsafeGetProcess()
 		proc.RefDec()
```
olsajiri, line 322: so the changelog says you want to ensure PID == TID for exec/exit events,
but the change has also changes above that apparently do something else,
could we split that into separate commit with its own explanation?

pkg/grpc/exec/exec.go
tixxdz: Indeed @olsajiri , I added a comment why in the updates patches. But basically we re-use the TID of exec/clone event anyway. so even if we set it , it will be overwritten after.

This check is for debugging and I commented in the patch that we are still not finished with the complete per thread transition, as we still don't handle capabilities and namespaces, and remember we have only one entry per process threads that is the leader either in bpf execve_val map or in the user space process cache.

Later I will remove those extra TIDs that I'm sending from BPF in clone/execve and exit, and just send PID and re-use it in userspace as TID.

pkg/grpc/exec/exec.go
tixxdz: You are completely right! I did split them and re-wrote extensive commit logs for it.

Sorry for that.

tixxdz:
Ok @olsajiri updated based on all comments, and yes it turned more work with documenting all the stuff that it is still missing as you suggested. so basically we are not finished with all per thread transition hence some extra debug and warning messages... here and there.

pkg/grpc/exec/exec.go
```diff
@@ -336,10 +336,9 @@ func GetProcessExit(event *MsgExitEventUnix) *tetragon.ProcessExit {
 		parent.RefDec()
 	}
 	if proc != nil {
-		tetragonEvent.Process = proc.GetProcessCopy()
+		// Use cached version of process
+		tetragonEvent.Process = proc.UnsafeGetProcess()
```
olsajiri, line 344: is this really needed? the etragonEvent.Process is initialized with tetragonProcess earlier, which is basically proc.UnsafeGetProcess ... looks like we just need to call the proc.RefDec in here? like for the parent case

tixxdz: makes sense ;-) !  done

pkg/grpc/exec/exec.go
```diff
@@ -288,6 +288,36 @@ func GetProcessExit(event *MsgExitEventUnix) *tetragon.ProcessExit {
 	code := event.Info.Code >> 8
 	signal := readerexec.Signal(event.Info.Code & 0xFF)
 
+	// Now since we want to correlate the exit and exec events, then
+	// the final TID of the exit event _must matches_ the PID and TID of exec
+	// event. The tetragonEvent.Process of this exit event is constructed either:
+	//   1. When looking up the process by its PID from user space cache
+	//      then we get the process that was pushed during exec or clone.
+	//   2. Out of order exec event, and this is same as (1).
```
jrfastab, line 298: I don't understand the (2) bullet here. I think it can just be dropped? If exec/exit are received OOO we we have to hold the exit event until we process the exec event. But its not really different than 1 which I think is what this comment says. Either way I found 2. more confusing than helpful. Maybe other have opinion.

pkg/grpc/exec/exec.go
```diff
@@ -288,6 +288,36 @@ func GetProcessExit(event *MsgExitEventUnix) *tetragon.ProcessExit {
...
+	//
+	// For all these cases the TID of the exit will be _automatically set_ to the
+	// TID and PID of the exec/clone event.
+	//
+	// However from bpf side exit events on their own must send bpf PID == TID, if not
+	// let's print a warning that is useful to bisect things. This helps to
+	// assert the different variants of tetragon and that bpf part did not change
+	// on behalf of user space, also useful as we cache only the main thread in the
+	// process cache, and it can die before other threads...
```
jrfastab, line 307: I don't understand point of this comment "also useful as we cache only the main thread in the
	// process cache, and it can die before other threads."
	
IF main thread dies we still wont get an exit event until all threads are complete right?

pkg/grpc/exec/exec.go
```diff
@@ -288,6 +288,36 @@ func GetProcessExit(event *MsgExitEventUnix) *tetragon.ProcessExit {
...
+	// Check must be against event.Info.Tid so we cover all the cases of
+	// the tetragonProcess.Pid against BPF.
+	if tetragonProcess.Pid.GetValue() != event.Info.Tid {
+		logger.GetLogger().WithFields(logrus.Fields{
+			"event.name":           "Exit",
+			"event.process.pid":    event.ProcessKey.Pid,
+			"event.process.tid":    event.Info.Tid,
+			"event.process.binary": tetragonProcess.Binary,
+		}).Warn("ExitEvent: process PID and TID mismatch")
+		// Final tetragonEvent.Tid will have the TID and PID of the exec/clone
```
jrfastab, line 320: This comment is saying same thing we have above in text. I would remove this comment it doesn't add extra knowledge as best I can tell.

pkg/process/process.go
```diff
@@ -202,13 +202,30 @@ func initProcessInternalExec(
 	protoPod := GetPodInfo(containerID, process.Filename, args, process.NSPID)
 	caps := caps.GetMsgCapabilities(capabilities)
 	ns := namespace.GetMsgNamespaces(namespaces)
+	binary := path.GetBinaryAbsolutePath(process.Filename, cwd)
+
+	// Ensure that exported events have the TID set. For events from Kernel
+	// we usually use PID == 0, so instead of checking against 0, assert that
```
jrfastab, line 210: Can we make this a touch clearer what we mean by "For events from Kernel we ..."->"For events generated by kernel threads we..."

tixxdz: will do yes

pkg/grpc/exec/exec.go
tixxdz: Indeed just trying to say that (2) will end up same as (1) but it takes different code path, maybe you are too familiar with it ;-) , will drop it sure

pkg/grpc/exec/exec.go
tixxdz: Indeed, I was trying to be verbose here... will remove it

jrfastab:
I find this commit message to be confusing,
```
 We want to correlate the exit and exec events, for this the final TID
of the exit event _must matches_ the PID and TID of exec / clone events
since we handle all these by the main thread.

The tetragonEvent.Process of the exit event is constructed either:

  1. When looking up the process by its PID from user space cache
     then we get the process that was pushed during exec or clone.

  2. Out of order exec event, and this is same as (1).

For all these cases the TID of the exit should be _automatically set_ to
the TID and PID of the exec/clone event. To achieve this we ensure that
clone and exec events have their TID also set to PID.

* clone events must store the TID in the process cache directly and
  we ensure that it is the PID as we generate only one event that
  is the thread leader.

* exec event we ensure and automatically set TID to PID too and insert it
  into the process cache as it is the main thread doing the execve call.

However from bpf side:
We also ensure that exit events on their own must send bpf PID == TID,
if not let's print a warning that is useful to bisect things.

This helps to assert the different variants of tetragon and that
sensors or bpf part did not change on behalf of user space and cache
handling, as again we handle only one main thread in the bpf side and
also in process cache, and we still did not complete all the per-thread
fields transition.
```
My sort of understanding is something like,
```
Exit events should have TID==PID and the {TID,PID} in the exit event must match the {TID,PID} pair from the exec event. They must match because its how we link the exit event to the exec event. If TID!=PID or the Exit{TID,PID} pair does not match an Exec{TID,PID} pair then there is a bug.

This patch adds extra logic to WARN on conditions where TID!=PID to aid debugging and catch this unexpected case. Typically this indicates a bug either in BPF or userspace caching logic. When this condition is encountered we attempt to correct it by setting the TID=PID.
```

pkg/process/process.go
```diff
@@ -202,13 +202,30 @@ func initProcessInternalExec(
...
+	// TGID == TID
+	if process.PID != process.TID {
+		logger.GetLogger().WithFields(logrus.Fields{
+			"event.name":            "Execve",
+			"event.process.pid":     process.PID,
+			"event.process.tid":     process.TID,
+			"event.process.binary":  binary,
+			"event.process.exec_id": execID,
+			"event.parent.exec_id":  parentExecID,
+		}).Warn("ExecveEvent: process PID and TID mismatch")
```
jrfastab, line 222: These warnings need to be metric error counters so we can actually find them. No one will notice a rare and random warning in the logs.

pkg/grpc/exec/exec.go
```diff
@@ -369,13 +366,12 @@ func (msg *MsgExitEventUnix) RetryInternal(ev notify.Event, timestamp uint64) (*
 	}
 
 	if internal != nil {
+		// Use cached version of the process
+		ev.SetProcess(internal.UnsafeGetProcess())
```
jrfastab, line 374: Can you remind me why this is safe? We dec the reference after this, but create a reference to the pointer here? The point is that we are starting the cache removal here and hoping that the event leaves the system before the destroy happens?

jrfastab: I think its the same logic used on HandleMessage() so should be ok?

jrfastab: OK nevermind switching runtimes C <->Golang created some brain melt.

So basic idea here is Exit event is telling us the object can be removed assuming no child reference it with a refinc. After Exit we may get kprobe events if they are OOO but this is OK because we only kick of the removal process. There is going to be a delay betwen exit dec'ing the refcnt and its actual removal from the cache. Finally it also doesn't matter if its removed before the Exit event is pushed because golang runtime knows we have a reference to it here and wont free the actual memory.

bpf/process/bpf_execve_event.c
```diff
@@ -177,10 +177,14 @@ event_execve(struct sched_execve_args *ctx)
...
-	 * than the calling thread are destroyed, but since we hook late
-	 * during the execve then the calling thread at the hook time is
-	 * already the new thread group leader.
+	/**
+	 * Per thread tracking rules TID == PID :
+	 *  At exec all threads other than the calling one are destroyed, so
+	 *  current becomes the new thread leader since we hook late during
+	 *  execve.
+	 *
+	 *  TODO: for now we send the TID but later we should remove this
```
jrfastab, line 186: can y ou drop this TODO comment. Create a github issue if we want it. I don't think its really critical given 4B and don't like code littered with TODOs.

bpf/process/bpf_exit.h
```diff
@@ -56,7 +56,15 @@ static inline __attribute__((always_inline)) void event_exit_send(void *ctx, __u
...
 
-		/* We track and report only thread leader so here tgid == tid */
+		/**
+		 * Per thread tracking rules TID == PID :
+		 *  We want the exit event to match the exec one, and since during exec
+		 *  we report the thread group leader, do same here as we read the exec
+		 *  entry from the execve_map anyway and explicitly set it to the to tgid.
+		 *
+		 *  TODO: for now we send the TID but later we should remove this
+		 *    and set it directly inside user space.
```
jrfastab, line 66: same drop todo please.

bpf/process/bpf_fork.c
```diff
@@ -75,9 +75,14 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
...
-			/* Since we generate one event per thread group, then when
-			 * the task wakes up, there will be only one thread here:
-			 * the thread group leader. Pass its thread id to user-space.
+			/**
+			 * Per thread tracking rules TID == PID :
+			 *  Since we generate one event per thread group, then when this task
+			 *  wakes up it will be the only one in the thread group, and it is
+			 *  the leader. Ensure to pass TID to user space.
+			 *
+			 *  TODO: for now we send the TID but later we should remove this
```
jrfastab, line 84: same

jrfastab (approved):
My comments aside looks good to me. Few small cleanups and I think we can get this merged.

pkg/grpc/exec/exec.go
tixxdz: ok removed

pkg/grpc/exec/exec.go
tixxdz: Ok re-wrote text as suggested with minor changes, thanks

pkg/process/process.go
tixxdz: done thanks ;-)

pkg/grpc/exec/exec.go
tixxdz: > OK nevermind switching runtimes C <->Golang created some brain melt.
> 
> So basic idea here is Exit event is telling us the object can be removed assuming no child reference it with a refinc. After Exit we may get kprobe events if they are OOO but this is OK because we only kick of the removal process. There is going to be a delay betwen exit dec'ing the refcnt and its actual removal from the cache. Finally it also doesn't matter if its removed before the Exit event is pushed because golang runtime knows we have a reference to it here and wont free the actual memory.

Indeed, so with the RefDec() if it is zero we push it to the deletequeue that is the delay between exit dec'ing the refcnt and the actual removal from the cache, the GC then will trigger at first round at X time for this process to be deleted which will set its color to deleteReady, then a second trigger after 30sec will then remove the process from the cache, during this time if an out of order events comes kprobes after we RefDec the process at exit, then the event cache triggers every 2 secs for a total of 15 cache strikes which covers the same duration, so I guess we are ok , and if by chance the process is removed and the event cache retry returns nil on the process then we have the metrics on the ProcessInfo or PodInfo errors to indicate that...

Great work btw ;-) !  and for C<->Golang and golang runtime couldn't say it better haha :-D !

bpf/process/bpf_execve_event.c
tixxdz: yes already tracking it in an other issue, removing

tixxdz:
> I find this commit message to be confusing,
> 
> ```
>  We want to correlate the exit and exec events, for this the final TID
> of the exit event _must matches_ the PID and TID of exec / clone events
> since we handle all these by the main thread.
> 
> The tetragonEvent.Process of the exit event is constructed either:
> 
>   1. When looking up the process by its PID from user space cache
>      then we get the process that was pushed during exec or clone.
> 
>   2. Out of order exec event, and this is same as (1).
> 
> For all these cases the TID of the exit should be _automatically set_ to
> the TID and PID of the exec/clone event. To achieve this we ensure that
> clone and exec events have their TID also set to PID.
> 
> * clone events must store the TID in the process cache directly and
>   we ensure that it is the PID as we generate only one event that
>   is the thread leader.
> 
> * exec event we ensure and automatically set TID to PID too and insert it
>   into the process cache as it is the main thread doing the execve call.
> 
> However from bpf side:
> We also ensure that exit events on their own must send bpf PID == TID,
> if not let's print a warning that is useful to bisect things.
> 
> This helps to assert the different variants of tetragon and that
> sensors or bpf part did not change on behalf of user space and cache
> handling, as again we handle only one main thread in the bpf side and
> also in process cache, and we still did not complete all the per-thread
> fields transition.
> ```
> 
> My sort of understanding is something like,
> 
> ```
> Exit events should have TID==PID and the {TID,PID} in the exit event must match the {TID,PID} pair from the exec event. They must match because its how we link the exit event to the exec event. If TID!=PID or the Exit{TID,PID} pair does not match an Exec{TID,PID} pair then there is a bug.
> 
> This patch adds extra logic to WARN on conditions where TID!=PID to aid debugging and catch this unexpected case. Typically this indicates a bug either in BPF or userspace caching logic. When this condition is encountered we attempt to correct it by setting the TID=PID.
> ```

Thanks @jrfastab I changed the commit log to what you suggested and also inside the code ;-)

pkg/process/process.go
tixxdz: So all fixed except for this adding metrics, will do in separate PR, so I don't mess up this one ;-)

tixxdz:
> My comments aside looks good to me. Few small cleanups and I think we can get this merged.

@olsajiri I did all you changes, plus last requests from @jrfastab too! so I will merge this one if it's green ;-)

PR #1259, opened 2023-07-20
Let renovate update Go toolchain in a single PR

### LINKED ISSUE #1169: Let renovatebot update Go toolchain in a single PR
See:
- https://github.com/cilium/cilium/issues/24806
- https://github.com/cilium/cilium/pull/24895

This is awesome and we should do it in Tetragon, our Golang versions are inconsistent and this is the ultimate solution.

We have only one task that might be needed before doing that which is getting rid of the cilium-builder image we use in the main Dockerfiles. We use it in this step: https://github.com/cilium/tetragon/blob/be355a8062cab637e6a9a7ec95f8a6b926e73816/Dockerfile#L23C1-L33

And this is an internal image used by Cilium where they have access to the Go information but we don't since the image has not external tags https://github.com/cilium/cilium/blob/438ed82626f9bf0a82c507c2c9faec28509f2b31/images/builder/Dockerfile.

mtardy:
Fixes #1169.

It was way more complicated than initially thought and it took me something like 2 days asynchronously to debug/test on the fork. The result should be as fancy as this PR on the fork https://github.com/mtardy/tetragon/pull/63/files.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -2,6 +2,12 @@
   // self-hosted configuration
   "username": "cilium-renovate[bot]",
   "repositories": ["cilium/tetragon"],
+  // renovate first reads this configuration, then reads the repository
+  // configuration, since we don't split between the self-hosted and the
+  // repository configuration, this can lead to duplicate of some areas of the
+  // config, for example the regex. See:
+  // https://docs.renovatebot.com/self-hosted-configuration/#requireconfig
+  "requireConfig": "ignored",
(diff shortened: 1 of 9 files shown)


kkourt (approved):
Regexes seem tricky, but what can you do? :)

willfindlay (approved):

mtardy:
How will the CI god punish me this time? Time to merge!! 🚢

PR #1271, opened 2023-07-24
e2e tests: remove gexe dependency in cilium management

mtardy:


willfindlay:
@mtardy this is fine by me. I actually copied most of this Cilium installation code from the Cilium folks without looking at it too much.

willfindlay (approved):

PR #1273, opened 2023-07-25
tetragon: Skip loading of override prog for kprobe multi object

olsajiri:
In case the override helper is not supported we can get verifier failure when loading kprobe multi object.

Adding the code to skip the loading of override program the same way we do for normal kprobes.


jrfastab (approved):

olsajiri:
got fixed in https://github.com/cilium/tetragon/pull/1218

PR #1276, opened 2023-07-26
sensors: harden string parsing from BPF events

mtardy:
Initially found this issue by trying to fuzz parts of the handler. You can run the fuzzer with the following command:
`go test -fuzz=Fuzz_parseString -run=Fuzz_parseString ./pkg/sensors/tracing/` Also added some unit test cases under `Test_parseString`.

Previous parsing had flaws that could make Tetragon panic or be OOM killed in certain circumstances. In reality, the string size was already bound by BPF copy_string function to 1024 but it could have theoretically returned a size of -2 as an error code which would have made it panic.

Also remove some code duplication that was rewriting the string parsing instead of using the common function.

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -7,7 +7,9 @@ import (
 	"bytes"
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"path"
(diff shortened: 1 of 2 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -954,7 +974,19 @@ func handleGenericKprobe(r *bytes.Reader) ([]observer.Event, error) {
...
 			arg.Index = uint64(a.index)
-			arg.Value = handleGenericKprobeString(r)
+			arg.Value, err = parseString(r)
+			if err != nil {
+				if errors.Is(err, errParseStringSize) {
+					// If no size then path walk was not possible and file was
+					// either a mount point or not a "file" at all which can
+					// happen if running without any filters and kernel opens an
+					// anonymous inode. For this lets just report its on "/" all
+					// though pid filtering will mostly catch this.
```
mtardy, line 984: I'm not sure I get this comment so maybe we should not apply this for `gt.GenericFileType, gt.GenericFdType, gt.GenericKiocb:` and `case gt.GenericPathType:`. I just reestablished it to have exactly the same behavior.

jrfastab: It looks like the case I'm trying to handle is the anonymous file case where the value returned here is empty. I guess its fine to do for all the other cases as well. What we need to be sure is we report something so the logs at least have an entry.

mtardy: ok I see, I just have one last question, in the patched that introduced that you wrote:
```text
When using fd_install hook to monitor program open* if the open is against
the mount then our path walking logic will return the size=0. The user
side then interprets this as an error and prints an ugly warning.
```
And the thing I wonder is, if the size is indeed 0 and it's correctly writing a 4 bytes 0, decoding should be okay and not error:
https://github.com/cilium/tetragon/blob/b2e3858a05da4305bbf2b97d999869b87a2c83a1/bpf/process/types/basic.h#L412
But I guess you encountered the error? Because the only situation I see that could trigger this would be that BPF sends a message less than 4 bytes in total for that arg, thus we would get an EOF.

pkg/sensors/tracing/generickprobe_test.go
```diff
@@ -0,0 +1,52 @@
...
+			}
+		})
+	}
+	t.Run("remove trailing null byte", func(t *testing.T) {
+		out, err := parseString(bytes.NewReader([]byte{6, 0, 0, 0, 'p', 'i', 'z', 'z', 'a', 0}))
+		if err != nil {
+			t.Errorf("unexpected error %v", err)
+		}
+		if out != "pizza" {
+			t.Errorf("got %q, want %q", out, "pizza")
```
kkourt, line 49: 🍕!

kkourt (approved):
Thanks!

pkg/sensors/tracing/generickprobe.go
```diff
@@ -774,35 +776,43 @@ func loadGenericKprobeSensor(bpfDir, mapDir string, load *program.Program, verbo
...
+	}
+
+	if size < 0 {
+		return "", errors.New("string size is negative")
 	}
-	outputStr := make([]byte, b)
-	err = binary.Read(r, binary.LittleEndian, &outputStr)
+
+	// limit the size of the string to avoid huge memory allocation and OOM kill in case of issue
+	if size > maxStringSize {
```
jrfastab, line 801: This check is defensive right? We should not get such a string size from the BPF kernel program?

jrfastab:
I'm happy with this one question though just to be sure I understood this correctly.

pkg/sensors/tracing/generickprobe.go
mtardy: Yes see MAX_STRING here

https://github.com/cilium/tetragon/blob/3f0e05b17f90fc015b1d170018eb70fc3b486e54/bpf/process/types/basic.h#L447

pkg/sensors/tracing/generickprobe.go
mtardy: Ok maybe this is never triggered anymore

PR #1280, opened 2023-07-27
contrib: Support upgrade in install-tetragon.sh

lambdanis:
Using `helm upgrade --install` instead of `helm install` makes it possible to use the same script for both installing and upgrading Tetragon in a Kubernetes cluster.

Also added a timeout for `kubectl rollout status` command to prevent the script from handing forever.

### DIFF
--- a/contrib/localdev/bootstrap-kind-cluster.sh
+++ b/contrib/localdev/bootstrap-kind-cluster.sh
@@ -1,13 +1,13 @@
 #! /bin/bash
 
 error() {
-    echo $@ 1>&2
+    echo "$@" 1>&2
     exit 1
 }
 
 set -eu
 
-PROJECT_ROOT="$(realpath $(dirname "${BASH_SOURCE[0]}")/../..)"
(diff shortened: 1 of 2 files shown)


contrib/localdev/bootstrap-kind-cluster.sh
```diff
@@ -1,13 +1,13 @@
...
 error() {
-    echo $@ 1>&2
+    echo "$@" 1>&2
     exit 1
 }
 
 set -eu
 
-PROJECT_ROOT="$(realpath $(dirname "${BASH_SOURCE[0]}")/../..)"
+PROJECT_ROOT="$(realpath "$(dirname "${BASH_SOURCE[0]}")"/../..)"
```
mtardy, line 10: Unrelated to this PR but `git rev-parse --show-toplevel` would be more resilient.

mtardy (approved):
Thanks!!

contrib/localdev/bootstrap-kind-cluster.sh
lambdanis: I'll update it too

PR #1281, opened 2023-07-27
Pr/michi/prateek copy

michi-covalent:
to test it locally:

```
% cat my-values.yaml
podWatcher:
  enabled: true
tetragonOperator:
  image:
    override: quay.io/cilium/tetragon-operator-ci:c128c9113c0b4c4512e154f9d9d721c37446fdf6

% helm install -n kube-system tetragon ./install/kubernetes/ -f my-values.yaml
```


tetragonpod/controller/tetragonpod_controller.go
```diff
@@ -0,0 +1,194 @@
...
+			// create the tetragonPod with the same name and namespace as the pod.
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      pod.Name,
+				Namespace: pod.Namespace,
+			},
+			Status: ciliumiov1alpha1.TetragonPodStatus{
+				PodIP:  newIP.IP,
+				PodIPs: make([]ciliumiov1alpha1.PodIP, 0),
+			},
+		}
```
michi-covalent, line 159: extract this logic into a separate function so that you can unit test it.

tetragonpod/controller/tetragonpod_controller.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	// CASE 2.2: TetragonPod does not exist, create new.
+	l.Info("Pod Created: ", "Name", req.Name, "Namespace", req.Namespace)
+	createTetragonPod(ctx, pod, r, l)
+	return ctrl.Result{}, nil
+}
+
+// deleteTetragonPod deletes the corresponding tetragonPod resource.
+func deleteTetragonPod(ctx context.Context, r *TetragonPodReconciler, req ctrl.Request, l logr.Logger) (ctrl.Result, error) {
```
michi-covalent, line 83:
```suggestion
func (r *TetragonPodReconciler) deleteTetragonPod(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
then just call `log.FromContext(ctx)` to access the logger. same for other functions.

tetragonpod/controller/tetragonpod_controller.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return nil, false
+	}
+	return tetragonPod, true
+}
+
+// createTetragonPod creates a tetragonPod resource.
+func createTetragonPod(ctx context.Context, pod *corev1.Pod, r *TetragonPodReconciler, l logr.Logger) (ctrl.Result, error) {
+	// create the tetragon pod since the pod exists
+	if pod.Status.Phase == corev1.PodRunning {
```
michi-covalent, line 144: instead of waiting for the pod to start, we want to create `TetragonPod` as soon as we have all the necessary info.

tetragonpod/controller/tetragonpod_controller.go
```diff
@@ -0,0 +1,194 @@
...
+	// CASE 2: Pod is created or updated.
+	// wait until the pod is running for cleaner logs.
+	if pod.Status.Phase != corev1.PodRunning {
+		return ctrl.Result{}, nil
+	}
+
+	// Check if corresponsind tetragonPod already exists.
+	tetragonPod, exists := checkIfAlreadyExists(ctx, req, r, l)
+	if exists {
+		// CASE 2.1: TetragonPod already exists, update it.
```
michi-covalent, line 70: avoid updating if there is no change in the fields that we are interested in.

tetragonpod/controller/tetragonpod_controller.go
```diff
@@ -0,0 +1,194 @@
...
+// Reconcile is the main loop of the controller that something changes with the pod resources in the cluster.
+func (r *TetragonPodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	l := log.FromContext(ctx)
+
+	// Get the Pod.
+	pod := &corev1.Pod{}
+	if err := r.Get(ctx, req.NamespacedName, pod); err != nil {
+
+		// CASE 1: Pod is deleted, therefore delete the corresponding tetragonPod resource.
+		if client.IgnoreNotFound(err) == nil {
```
michi-covalent, line 54: we need to handle the other case where error is something other than `NotFound`

tetragonpod/controller/tetragonpod_controller.go
```diff
@@ -0,0 +1,194 @@
...
+		tetragonPod := &ciliumiov1alpha1.TetragonPod{
+
+			// create the tetragonPod with the same name and namespace as the pod.
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      pod.Name,
+				Namespace: pod.Namespace,
+			},
+			Status: ciliumiov1alpha1.TetragonPodStatus{
+				PodIP:  newIP.IP,
+				PodIPs: make([]ciliumiov1alpha1.PodIP, 0),
```
michi-covalent, line 157: copy `PodIPs` field too

tetragonpod/controller/tetragonpod_controller.go
prateek041: What to do in that case ? log the error and return `ctrl.Result{}, err` ? or something else

tetragonpod/controller/tetragonpod_controller.go
prateek041: I added that to reduce the logs being generated, because the controller is triggered for every state change too. I should use `log.Debug` instead of `log.Info` as you suggested. Should I instead for the `pod.Status.Phase == corev1.PodPending` ?

tetragonpod/controller/tetragonpod_controller.go
michi-covalent: yes just return `ctrl.Result{}, err`

tetragonpod/controller/tetragonpod_controller.go
prateek041: Just re-confirming. The controller will wait for all the necessary information it needs, once done, it will create the respective tetragonPods right ?

tetragonpod/controller/tetragonpod_controller.go
prateek041: sure

tetragonpod/controller/tetragonpod_controller.go
prateek041: done

tetragonpod/controller/tetragonpod_controller.go
prateek041: done

tetragonpod/controller/tetragonpod_controller.go
prateek041: done

tetragonpod/controller/tetragonpod_controller.go
prateek041: done

michi-covalent:
let's close this one. we are using https://github.com/cilium/tetragon/pull/1410 instead

PR #1283, opened 2023-07-28
Lists

olsajiri:
syscalls specified in list:

```
kind: TracingPolicy
metadata:
  name: "list-syscalls"
spec:
  lists:
  - name: "syscalls"
    type: "syscalls"
    values:
    - "sys_dup"
    - "sys_dup2"
  kprobes:
  - call: "list:syscalls"
```

all syscalls (generated):

```
kind: TracingPolicy
metadata:
  name: "all-syscalls-for-kill"
spec:
  lists:
  - name: "all-syscalls"
    type: "generated_syscalls"
  kprobes:
  - call: "list:all-syscalls"
    selectors:
    - matchBinaries:
      - operator: "In"
        values:
        - "/usr/bin/kill"
```

ftrace functions (generated) with filter:
```
kind: TracingPolicy
metadata:
  name: "all-ksys-for-kill"
spec:
  lists:
  - name: "ksys"
    type: "generated_ftrace"
    generate: "^ksys_*"
  kprobes:
  - call: "list:ksys"
    selectors:
    - matchBinaries:
      - operator: "In"
        values:
        - "/usr/bin/kill"
```

standard functions in the list:
```
kind: TracingPolicy
metadata:
  name: "ftrace"
spec:
  lists:
  - name: "ftrace"
    values:
    - "ksys_dup3"
    - "ksys_fadvise64_64"
    - "ksys_fallocate"
    - "ksys_fchown"
    - "ksys_ioperm"
    ...
    - "ksys_sync_file_range"
    - "ksys_sync_helper"
    - "ksys_unshare"
    - "ksys_write"
  kprobes:
    - call: "list:ftrace"
```


pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -331,3 +335,15 @@ type UProbeSpec struct {
...
+type ListSpec struct {
+	// +kubebuilder:validation:Optional
+	// Name of the list
+	Name string `json:"name"`
+	// +kubebuilder:validation:Optional
+	// Values of the list
+	Values []string `json:"values"`
+	// +kubebuilder:validation:Optional
+	// Indicates whether the symbols in list are syscalls.
+	Syscall bool `json:"syscall"`
```
kkourt, line 348: Should we have an enum type here so that it can be extended to other types if needed?

kkourt (commented):
Cool! Had a quick look and left a comment.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
olsajiri: hum, so at the moment we have the verification only for syscalls based on this flag,
do you have some other types in mind?

kkourt: IP addresses as well.

olsajiri: ah right, so have it as general purpose list.. I was stuck with functions ;-) sounds good, will change that

pkg/sensors/tracing/generickprobe.go
```diff
@@ -315,6 +325,17 @@ func preValidateKprobes(name string, kprobes []v1alpha1.KProbeSpec) error {
 			}
 		}
 
+		if strings.HasPrefix(f.Call, "list:") {
```
kkourt, line 328: could there be functions in the kernel that this prefix collides with?

Also, what is the meaning of the rest of the fields when the call has a `list:` prefix. 

Looking at the kprobe definition, besides `Call`, we have: `Return`,  `Syscall`, `Args`,  `ReturnArg`, `ReturnArgAction`, and `Selectors`. For `Return`, `Syscall` and `Selectors`, we can say that whatever is defined there applies to all calls.

Not so sure about `Args`, `ReturnArg`, and `ReturnArgAction`. I think we could leave it as is. The argument fields will be applied to all calls in the list, and we assume the user is knows what they are doing. In the future, we could also provide `list:` arguments to specify them separately.

So I _think_ it works.  We should definitively document this though. I guess we would want to do this as a followup, so I created an issue: https://github.com/cilium/tetragon/issues/1365.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -331,3 +335,18 @@ type UProbeSpec struct {
...
+type ListSpec struct {
+	// Name of the list
+	Name string `json:"name"`
+	// +kubebuilder:validation:Optional
+	// Values of the list
+	Values []string `json:"values"`
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=syscalls;generated_syscalls;generated_ftrace
+	// Indicates the type of the list values.
+	Type string `json:"type"`
```
kkourt, line 348: As far as I can tell from the code, `generate_syscalls` will generate all syscalls (and ignore the `generate` argument). Is there a plan to add pattern matching in the `generate` argument as well?

pkg/sensors/tracing/generickprobe.go
olsajiri: > could there be functions in the kernel that this prefix collides with?

there's the ':' and that can't be part of kernel symbol

might be problem for C++ symbols for uprobes I guess.. but we could force single ':' compared to C++ '::'

> 
> Also, what is the meaning of the rest of the fields when the call has a `list:` prefix.
> 
> Looking at the kprobe definition, besides `Call`, we have: `Return`, `Syscall`, `Args`, `ReturnArg`, `ReturnArgAction`, and `Selectors`. For `Return`, `Syscall` and `Selectors`, we can say that whatever is defined there applies to all calls.
> 
> Not so sure about `Args`, `ReturnArg`, and `ReturnArgAction`. I think we could leave it as is. The argument fields will be applied to all calls in the list, and we assume the user is knows what they are doing. In the future, we could also provide `list:` arguments to specify them separately.
> 

the list in 'call' field is just sort of values iterator and the rest of the fields from kprobe applies to each value,
as if it was separate kprobe definition with 'call: valueX' from the list

> So I _think_ it works. We should definitively document this though. I guess we would want to do this as a followup, so I created an issue: #1365.

yes, I wanted to get some agrement first before I'd touch docs ;-)

pkg/k8s/apis/cilium.io/v1alpha1/types.go
olsajiri: I saw the use case for that in your recent document, so yes that can be added,
also I'm thinking of renaming the 'generate' field to 'pattern'

pkg/sensors/tracing/generickprobe.go
kkourt: > might be problem for C++ symbols for uprobes I guess.. but we could force single ':' compared to C++ '::'

We are only applying it to kprobes, so I think that's fine.

> the list in 'call' field is just sort of values iterator and the rest of the fields from kprobe applies to each value,
as if it was separate kprobe definition with 'call: valueX' from the list

👍🏼

pkg/k8s/apis/cilium.io/v1alpha1/types.go
kkourt: I personally like `pattern` but naming is subjective so I'm fine with either option :)

kkourt (approved):
Overall, LGTM! Thanks!

mtardy:
This is awesome! 🎉

PR #1288, opened 2023-07-28
Copy Helm chart values doc to docs reference

### LINKED ISSUE #1287: Add a reference page generated from the helm chart values

mtardy:
Fixes #1287


docs/content/en/docs/reference/helm-chart.md
```diff
@@ -0,0 +1,104 @@
+---
+title: "Helm chart"
```
tixxdz, line 2: Please make this: Kubernetes Helm chart  , so it is clear for everyone.

tixxdz (approved):
one minor comment, thanks

docs/content/en/docs/reference/helm-chart.md
mtardy: Thanks for the review, however, I don't agree, what is a non-Kubernetes helm chart? I try to keep it short for the summary navbar on the left and don't want to introduce unnecessary stuff.

PR #1291, opened 2023-07-31
tetragon: kprobe fixes

olsajiri:
assorted kprobe related fixes

### DIFF
--- a/cmd/tetra/tracingpolicy/tracingpolicy.go
+++ b/cmd/tetra/tracingpolicy/tracingpolicy.go
@@ -25,7 +25,6 @@ import (
 	"github.com/cilium/tetragon/pkg/btf"
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
 )
 
 func New() *cobra.Command {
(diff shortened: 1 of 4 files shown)


cmd/tetra/tracingpolicy/tracingpolicy.go
```diff
@@ -77,7 +77,7 @@ func New() *cobra.Command {
 	tpListFlags.StringVarP(&tpListOutputFlag, common.KeyOutput, "o", "text", "Output format. text or json")
 
 	tpGenerateCmd := &cobra.Command{
-		Use:   "generate",
+		Use:   "generate <all-syscalls|empty>",
```
mtardy, line 80: we could use `cobra.Command.ValidArgs` that would provide more feedback to the user but it's a nit.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -290,9 +290,13 @@ func preValidateKprobes(name string, kprobes []v1alpha1.KProbeSpec) error {
...
-		hasOverride := selectors.HasOverride(f)
-		if hasOverride && !bpf.HasOverrideHelper() {
-			return fmt.Errorf("Error override action not supported, bpf_override_return helper not available")
+		if selectors.HasOverride(f) {
+			if !bpf.HasOverrideHelper() {
+				return fmt.Errorf("Error override action not supported, bpf_override_return helper not available")
+			}
+			if !f.Syscall {
+				return fmt.Errorf("Error override action can be used only with syscalls")
+			}
```
mtardy, line 304: just a question: why did we put this limitation? Because it's too easy to shoot yourself in the foot and crash the kernel with an override on any other kernel function?

mtardy (approved):
Thanks!! :)

pkg/sensors/tracing/generickprobe.go
olsajiri: the override is limited for syscalls and few other functions (marked with ALLOW_ERROR_INJECTION in kernel)
so we limited this for syscalls only, because I don't think we have a usecase for the rest

cmd/tetra/tracingpolicy/tracingpolicy.go
olsajiri: ok, I'm adding more sub commands in following PR, I'll see if I can use that, thanks

pkg/sensors/tracing/generickprobe.go
mtardy: ah I see thanks! :)

olsajiri:
I had to rebase and added one more commit

tixxdz (approved):
Great, thanks

kkourt (approved):
Thanks!

PR #1295, opened 2023-07-31
Makefile: Print messages only once

lambdanis:
A tiny patch, I'm just bothered that things get printed twice.


mtardy (approved):
I'm for Anna's appeasement 😮‍💨

PR #1296, opened 2023-07-31
tetragon:process_exec: display uids/gids credentials and detect privileged execution

tixxdz:
Print process_credentials and detect privileged setuid and setgid execution during process_exec events.

```json
    {
      "process_exec": {
        "process": {
          "exec_id": "OjU0NTE4NTE3MzEwMzg5OjQ0ODUwNw==",
          "pid": 448507,
          "uid": 1000,
          "binary": "/usr/bin/sudo",
          "arguments": "id",
          "flags": "execve clone",
          "start_time": "2023-09-12T22:19:37.574413120Z",
          "auid": 1000,
          "parent_exec_id": "OjMyMjU3OTgwMDAwMDAwOjQ4NjMy",
          "tid": 448507,
          "process_credentials": {
            "uid": 1000,
            "gid": 1000,
            "euid": 0,
            "egid": 1000,
            "suid": 0,
            "sgid": 1000,
            "fsuid": 0,
            "fsgid": 1000
          },
          "binary_properties": {
            "setuid": 0
          }
        }
    }
```

Still tasks todo:
- [x] apply the --enable-process-creds flag on it, if set do not display it.
- [x] Do we need to show this information in process_exit?
- [x] During other events process_kprobe, etc?
- [x] Mask parent.binary_properties by default during process_exec


kkourt:
Maybe something like:
```
 "binary_properties" : {
     "setuid": 0, /* if it's not setuid -> nil */
     "memfd": true,
 }
```

tixxdz:
> Maybe something like:
> 
> ```
>  "binary_properties" : {
>      "setuid": 0, /* if it's not setuid -> nil */
>      "memfd": true,
>  }
> ```

Great, so I'll add in this PR:
1. process_credentials and for now only the uids/gids without caps into the process
2. binaries_properties as you suggested so we correlate the setuid data.
3. for all the rest I will track it separately, better we do this incrementally in regards to the other threading work

bpf/process/bpf_process_event.h
```diff
@@ -453,6 +454,26 @@ get_current_subj_caps(struct msg_capabilities *msg, struct task_struct *task)
...
+
+	probe_read(&info->uid, sizeof(__u32), _(&cred->uid));
+	probe_read(&info->gid, sizeof(__u32), _(&cred->gid));
+	probe_read(&info->euid, sizeof(__u32), _(&cred->euid));
+	probe_read(&info->egid, sizeof(__u32), _(&cred->egid));
+	probe_read(&info->suid, sizeof(__u32), _(&cred->suid));
+	probe_read(&info->sgid, sizeof(__u32), _(&cred->sgid));
+	probe_read(&info->fsuid, sizeof(__u32), _(&cred->fsuid));
+	probe_read(&info->fsgid, sizeof(__u32), _(&cred->fsgid));
+	probe_read(&info->securebits, sizeof(__u32), _(&cred->securebits));
```
jrfastab, line 473: you could probably do this with a single probe_read using sizeof(offset_of()) but would make some assumptions about struct layout. Probably by its self might not matter but we might consider this type of things across all the progs we are starting to do lots of probe_reads.

tixxdz: yes let me see, however for generalizing that I'm not sure as some structs may have ifdef CONFIG_** inside

tixxdz: ah btw also forget to mention about struct layouts tagged to be randomized with gcc  ...

jrfastab: sounds good it was sort of a micro-optimization anyways.

bpf/lib/process.h
```diff
@@ -264,6 +246,8 @@ struct msg_execve_event {
 	struct msg_execve_key parent;
 	__u64 parent_flags;
 	struct msg_capabilities caps;
+	/* TODO: convert this tg_cred_minimal to tg_cred and include caps above inside */
+	struct tg_cred_minimal creds;
```
tpapagian, line 262: Why not keep the naming consistent here (i.e. have the `msg_` prefix)? For example `msg_cred_minimal`. It's better to be consistent here. 

Same for `tg_user_namespace` and `tg_cred`.

bpf/Makefile
```diff
@@ -24,11 +24,13 @@ BPFTESTDIR := test/
 
 ALIGNCHECKER = bpf_alignchecker.o
 PROCESS = bpf_execve_event.o bpf_execve_event_v53.o bpf_fork.o bpf_exit.o bpf_generic_kprobe.o \
+	  bpf_execve_bprm_commit_creds.o bpf_execve_bprm_commit_creds_v53.o \
```
tpapagian, line 27: Is there any reason to have 3 different versions of this program? I cannot find any differences related to the kernel version. Why not just have `bpf_execve_bprm_commit_creds.o` for all kernel versions?

bpf/process/bpf_execve_event.c
```diff
@@ -152,6 +152,16 @@ binary_filter(void *ctx, struct msg_execve_event *event, void *filename)
 	return value ? *value : 0;
 }
 
+static inline __attribute__((always_inline)) __u32
+read_execve_shared_info(void *ctx, __u64 pid, struct msg_process *p)
```
tpapagian, line 156: It doesn't seem that you use `struct msg_process *p` somewhere. Maybe remove that?

tpapagian (commented):
Thanks! Some initial comments, I will have another look shortly.

pkg/grpc/process_manager_test.go
```diff
@@ -171,6 +171,12 @@ func TestProcessManager_GetProcessExec(t *testing.T) {
 		Common: processapi.MsgCommon{
 			Ktime: 1234,
 		},
+		Creds: processapi.MsgGenericCredMinimal{
+			Uid:  0,
+			Gid:  0,
+			Euid: 0,
+			Egid: 0,
+		},
```
kkourt, line 179: 0 is the default value,  processapi.MsgGenericCredMinimal{} has the same effect but is more concise.

bpf/lib/process.h
tixxdz: Alright , done ;-)

bpf/lib/process.h
```diff
@@ -324,6 +337,40 @@ struct {
 	__type(value, struct execve_heap);
 } execve_heap SEC(".maps");
 
+/* the tg_execve_joined_info_map allows to join and combine
+ * exec info that is gathered during different hooks
+ * through the execve call.
```
kkourt, line 342: Please document the hooks and how information is gathered.

bpf/lib/process.h
```diff
@@ -409,6 +456,77 @@ static inline __attribute__((always_inline)) void execve_map_delete(__u32 pid)
...
+{
+	int one = MAP_STATS_ERROR;
+	__s64 *cntr;
+
+	cntr = map_lookup_elem(&tg_execve_joined_info_map_stats, &one);
+	if (cntr)
+		*cntr = *cntr + 1;
+}
+
+static inline __attribute__((always_inline)) int
```
kkourt, line 470: The return value of this function is not used. I'd suggest either use it, or modify the function to return void.

bpf/lib/process.h
```diff
@@ -155,6 +155,18 @@ struct msg_execve_key {
...
+#define EXEC_SETUID 0x01
+#define EXEC_SETGID 0x02
+
+/* This is the struct stored in bpf map to share info between
+ * different execve hooks.
+ */
+struct execve_info {
+	__u32 secureexec;
+	__u32 pad;
+};
```
kkourt, line 168: Why the padding here? Can't we just have a __u32 type value?
The name `secureexec` seems confusing to me. How about just `flags`?

bpf/lib/process.h
```diff
@@ -409,6 +456,77 @@ static inline __attribute__((always_inline)) void execve_map_delete(__u32 pid)
...
+		 * on the stats map this can be a good indication of
+		 * long running workloads and if we have to make the
+		 * map size bigger for such cases.
+		 */
+		execve_joined_info_map_error();
+		return err;
+	}
+
+	cntr = map_lookup_elem(&tg_execve_joined_info_map_stats, &zero);
+	if (cntr && *cntr < EXECVE_JOINED_MAP_SIZE)
```
kkourt, line 488: Why are we checking against EXECVE_JOINED_MAP_SIZE here? What are we counting?

bpf/lib/process.h
```diff
@@ -409,6 +456,77 @@ static inline __attribute__((always_inline)) void execve_map_delete(__u32 pid)
...
+		if (cntr)
+			*cntr = *cntr - 1;
+	}
+	/* We don't care here about -ENOENT as there is no guarantee entries
+	 * will be present anyway.
+	 */
+}
+
+static inline __attribute__((always_inline)) bool
+execve_joined_info_map_get(__u64 tid, struct execve_info *info, bool delete)
```
kkourt, line 513: Is the bool argument really needed here? There is only one caller.

bpf/process/bpf_execve_event.c
tixxdz: Yes, I was planning to cache errors on p->flags, but then we can't predict really errors and since these are multiple stages of hooking execve, the proper way to catch all those errors is to have a map per entier pid so we can transfert the errors happing at stage X of execve to the last execve_send() event per pid... which means we need a bigger map and more memory to reflect if lot of pids are doing execve at same time...

So I opted now with:
* tg_execve_joined_info_map with less entries and not really core information , just extra enhanced info and if the LRU map is full between all the execve then so be it, we won't have that info , but the metrics on the map will show it.
* stats tg_execve_joined_info_map_stats() map with 2 entries counter of entries for metrics and errors counter there... and we see in the future... @jrfastab in case for metrics on error maps.

bpf/process/bpf_execve_bprm_commit_creds.c
```diff
@@ -0,0 +1,62 @@
...
+	struct task_struct *task;
+	__u32 pid, euid, uid, egid, gid, sec = 0, zero = 0;
+	__u64 tid;
+
+	sec = BPF_CORE_READ(bprm, per_clear);
+	/* If no flags to clear then this is not a privileged execution */
+	if (!sec)
+		return;
+
+	tid = get_current_pid_tgid();
```
kkourt, line 31: Should we check for an error here?

bpf/Makefile
tixxdz: Sure, thanks

bpf/lib/process.h
```diff
@@ -324,6 +337,40 @@ struct {
...
 
+/* the tg_execve_joined_info_map allows to join and combine
+ * exec info that is gathered during different hooks
+ * through the execve call.
+ *
+ * Important: the information stored here is complementary
+ * information only, the core logic should not depend on entries
+ * of this map to be present.
+ * This will be exposed to userspace to enhance the process_exec
+ * event.
```
kkourt, line 348: Is the map read by userspace? If not, this comment is misleading.

kkourt (commented):
Thanks!

Overall, looks good to me, but I have some comments / suggestions.

bpf/lib/bpf_cred.h
```diff
@@ -0,0 +1,72 @@
...
+ * definition and is used in kprobes.
+ * However since we are also moving to use creds for
+ * exec events, so let's do it step by step, as we already
+ * have the capabilities in execve inside the execve_map and
+ * the user space cache, so we start with this minimal
+ * credential object that holds only uids/gids, then we follow
+ * up by moving the capabilities into it, and make this cred
+ * the new storage in execve_map and user space process cache.
+ */
+struct tg_cred_minimal {
```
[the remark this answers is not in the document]
tixxdz: Ah this will make it touch kprobe logic where I tried to be only about execve, plus next patches I will make them one struct anyway.

bpf/lib/process.h
tixxdz: yes so reader knows get() could also delete

bpf/process/bpf_process_event.h
```diff
@@ -453,6 +454,26 @@ get_current_subj_caps(struct msg_capabilities *msg, struct task_struct *task)
...
+	info->pad = 0;
```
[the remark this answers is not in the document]
tixxdz: Just a reflex for being defensive and not send non initialized data to userspace... in userspace we don't exports pad through grpc so could be fine to remove it...

bpf/lib/process.h
tixxdz: ah left over when I was testing to fill up the map... btw forget I still didn't add the map sum accounting for metrics didn't find an example in the code base, so I will do it in separate PR

bpf/lib/process.h
tixxdz: ok will fix it, thanks

bpf/lib/process.h
tixxdz: for the padding IIRC entries are aligned on 8bytes...

For the securexec let me see, in the kernel there is bprm->secureexec bit field to indicate this, I can do flag or better document along with how kernel do things

bpf/lib/process.h
tixxdz: Sure done

bpf/lib/process.h
tixxdz: Ok I left it secureexec but extensive documentation where it comes from, flags is too generic for this.

pkg/sensors/base/base.go
```diff
@@ -103,3 +117,14 @@ func ExecObj() string {
 		return "bpf_execve_event.o"
 	}
 }
+
+// ExecBprmCommitObj returns the exec object based on the kernel version
+func ExecBprmCommitObj() string {
```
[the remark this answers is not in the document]
tixxdz: hehe indeed ;-)

tixxdz:
thanks @kkourt @tpapagian @jrfastab for the review ;-), all fixed

tpapagian (approved):
Thanks!

bpf/lib/process.h
```diff
@@ -154,6 +155,36 @@ struct msg_execve_key {
...
+	 * or set-group-ID binary (so that its real and effective UIDs or GIDs differ
+	 * from one another), or that it gained capabilities by executing a binary file
+	 * that has capabilities (see capabilities(7)).
+	 * Alternatively, a nonzero value may be triggered by a Linux Security Module.
+	 * When this value is nonzero, the dynamic linker disables the use of certain
+	 * environment variables.
+	 *
+	 * The secureexec here can have the following bit flags:
+	 *   EXEC_SETUID or EXEC_SETGID
+	 */
```
kkourt, line 183: 🚀

bpf/lib/process.h
kkourt: Please use a different function name then. If the argument has always the same value, there is no need for it. `execve_joined_info_map_get_clear` for example.

pkg/grpc/exec/exec.go
```diff
@@ -232,6 +261,43 @@ func (msg *MsgExecveEventUnix) Cast(o interface{}) notify.Message {
...
+//     case new information could have been added, so let's do another
+//     ev.SetProcess() call to update the process of the event.
+func (msg *MsgExecveEventUnix) finalize(ev notify.Event, internal *process.ProcessInternal, cache int) error {
+	// This should never happen by this time
+	if ev.GetProcess() == nil {
+		return process.ErrProcessInfoMissing
+	}
+
+	update := false
+	proc, err := internal.UpdateExecOutsideCache(option.Config.EnableProcessCred, &update)
```
kkourt, line 285: Wouldn't be better to have:
```suggestion
	proc, update, err := internal.UpdateExecOutsideCache(option.Config.EnableProcessCred)
```

kkourt (commented):
Thanks! I think it's in great shape. Just have a couple minor comments.

pkg/grpc/exec/exec.go
tixxdz: done

tixxdz:
@kkourt fixed last bits, if it's green I will merge it thank you ;-)

pkg/reader/proc/proc.go
```diff
@@ -164,29 +175,50 @@ func GetProcPid(pid string) (uint64, error) {
...
+// Returns all parsed UIDs on success. If we fail for one value we do not
 // return the overflow ID, we return the invalid UID 4294967295
 // (-1 as an unsigned integer).
 // The overflow ID is returned when the kernel decides and pass it back,
 // as it can be a valid indication of UID mapping error.
-func GetUids(status *Status) (uint32, uint32, error) {
-	ruid, err := strconv.ParseUint(status.Uids[0], 10, 32)
-	if err != nil {
-		return InvalidUid, InvalidUid, err
+func (status *Status) GetUids() ([]uint32, error) {
```
mtardy, line 183: @tixxdz 
would have it been possible to use an array `[4]uint32` instead of a slice and indicate which UIDS correspond to what in the doc? Same for GetGids

mtardy: I think the best here would have been to have 5 named return parameters (so that we get doc for free), I think the slice/array part makes the function less nice to use.

tixxdz: AFAICT it can't return an empty, as all fields are explicitly initialized to invalid uids, so no need for a length check unless I'm missing something? plus we need to explicitly initialize this otherwise we may endup with uid 0, IIRC slices are generic so I don't think this is a real issue?

for doc https://github.com/cilium/tetragon/blob/main/pkg/reader/proc/proc.go#L17 , unless you are referring to function godoc? then there is too much work and issues to fix that I didn't think about the godoc to be honest, for me as long as the code is covered with e2e tests then I'm good ;-)

kkourt:
> @kkourt fixed last bits, if it's green I will merge it thank you ;-)

@tixxdz Please give people that have involved in the PR some time to ✅ before merging.

pkg/reader/proc/proc.go
mtardy: > AFAICT it can't return an empty, as all fields are explicitly initialized to invalid uids, so no need for a length check unless I'm missing something?

sure! So why use a slice (since you know the size in advance) instead of an array or just return the values flat? I don't say it's wrong but it's just a bit weird interface, I would say if you use such a function, you have to trust it to not crash your program (if you don't check the length of the slice) while if you return an array or a flat 5 values return, it will guarantee by design that it cannot crash your program.

No, exactly I was searching for this kind of doc, because if you just read the function you don't know what is inside the slice.

mtardy: but anyway, it's just a small function that's no big deal but I just wondered why you chose a slice here.

tixxdz: let's say due to background returning a reference to memory is perfectly fine! it is initialized and have its length set, so no issue here... in the other hand returning multiple vars even if ok with Go is weird interface, it means you are returning too much context in one single function and you probably have to split the logic there...

mtardy: Ok we are going in circles 😁 

> it is initialized and have its length set, so no issue here...

Let me show you what I mean, it's not that it doesn't work, something is not initialized or whatever, it's just using the benefits of the Go language: https://go.dev/play/p/ov0JGylWXup

image

> returning multiple vars even if ok with Go is weird interface, it means you are returning too much context in one single function and you probably have to split the logic there...

I strongly disagree, hiding it in an array doesn't change anything from returning directly flat. And bundling this tuple of uids here totally make sense.

PR #1301, opened 2023-08-01
bugfix: Fix Prefix operator with file type in matchArgs 

tpapagian:
The tracing policy provided in https://raw.githubusercontent.com/cilium/tetragon/main/examples/tracingpolicy/file_monitoring.yaml does not work as expected. 

The matchArgs filter with file type and Prefix operator results in events with filenames with different prefixes as what we expected. This means that expect from the events that we care about, we also get a bunch of unrelated events. 

More details on commit messages.

```release-note
Fix Prefix operator with file type in matchArgs.
```

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -649,17 +649,17 @@ filter_char_buf(struct selector_arg_filter *filter, char *args, int value_off)
 	return 0;
 }
 
-static inline __attribute__((always_inline)) long
-__filter_file_buf(char *value, char *args, __u32 op)
-{
-	int err;
-	__u64 v, a;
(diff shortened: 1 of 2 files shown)


bpf/process/types/basic.h
```diff
@@ -649,17 +649,17 @@ filter_char_buf(struct selector_arg_filter *filter, char *args, int value_off)
 	return 0;
 }
 
+struct string_buf {
+	__u32 len;
+	char buf[];
```
kkourt, line 654: Nice, I like that! It makes the format explicit.

kkourt (approved):
Thanks!

PR #1304, opened 2023-08-01
Create metrics registry explicitly instead of using promauto

lambdanis:
It's a minimal set of changes required to get rid of the promauto package.
There are no visible behaviour changes.

This refactoring is needed to implement some metrics control features, for
example granular user configuration or automatic deletion of stale metrics.

One things that might be not obvious is that observer test helpers now register
metrics and start metrics server on their own instead of using a function from
pkg/metrics. This is done to avoid an import cycle and might be worth
refactoring in the future.


willfindlay (approved):
Test changes look good to me.

jrfastab (approved):

lambdanis:
Sorry for the force pushes, I rebased with an older main to test something, but nothing changed in this PR.

PR #1308, opened 2023-08-02
tetragon: bpf map handlers updates

jrfastab:
Couple fixes for map walking code and then a new map walking API to allow map walk to abort after some number of entries. This handles the case where map is being updated while we are walking it.

### DIFF
--- a/pkg/bpf/map_linux.go
+++ b/pkg/bpf/map_linux.go
@@ -22,6 +22,7 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"syscall"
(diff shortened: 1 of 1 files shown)


pkg/bpf/map_linux.go
```diff
@@ -318,6 +319,9 @@ func (m *Map) DumpWithCallback(cb DumpCallback) error {
 	}
 
 	if err := GetFirstKey(m.fd, unsafe.Pointer(&nextKey[0])); err != nil {
+		if err == io.EOF {
+			return nil
+		}
 		return nil
```
kevsecurity, line 325: I'm surprised the linter didn't complain because the if seems unnecessary here.

willfindlay (approved):
LGTM

pkg/bpf/map_linux.go
willfindlay: Good catch

jrfastab: that does seem a bit pointless...  What I mean is EOF is an expected error so return nil. Otherwise we want to return the error because its something else.

kevsecurity: Line 325 should be 'return err' maybe?

PR #1311, opened 2023-08-03
tetragon: Add stats support for errors

jrfastab:
Maps have an associated _stats map that is used to keep total number of entries values. Lets use the stats maps to also store any errors that map updates/deletes may have encountered.

### DIFF
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -309,10 +309,15 @@ struct {
 
 struct {
 	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
-	__uint(max_entries, 1);
+	__uint(max_entries, 2);
 	__type(key, __s32);
 	__type(value, __s64);
 } execve_map_stats SEC(".maps");
+
+enum {
+	MAP_STATS_COUNT = 0,
+	MAP_STATS_ERROR = 1,
+};
 
 struct {
 	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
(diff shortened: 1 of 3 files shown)


bpf/lib/process.h
```diff
@@ -357,6 +357,17 @@ validate_msg_execve_size(int64_t size)
 	return size;
 }
 
+// execve_map_error() will increment the map error counter
+static inline __attribute__((always_inline)) void execve_map_error(void)
+{
+	int one = 1;
```
willfindlay, line 363: Maybe this is too pedantic but could we use a named constant here like MAP_STATS_ERROR for the error key and similarly MAP_STATS_COUNT for the size key?

I think it makes the purpose of the two keys more clear for those who want to use or extend it in the future.

willfindlay (approved):

bpf/lib/process.h
jrfastab: sure that makes sense to me.

pkg/metrics/mapmetrics/mapmetrics.go
```diff
@@ -25,13 +25,24 @@ var (
 		Help:        "The total number of entries dropped per LRU map.",
 		ConstLabels: nil,
 	}, []string{"map"})
+
+	MapErrors = promauto.NewGaugeVec(prometheus.GaugeOpts{
```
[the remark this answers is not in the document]
jrfastab: np I was waiting for 1304 to be merged

PR #1314, opened 2023-08-03
generic traceponts: add support for string type

kkourt:
See commits.

```release-note
generic tracepoints: add support for string type
```


jrfastab:
very cool thanks for doing it.

PR #1315, opened 2023-08-04
tetragon: Remove pkg/bpf/map module

olsajiri:
Remove pkg/bpf/map module and changing its current users to use ebpf.Map API.


tpapagian (approved):
LGTM, thanks for the cleanup!

PR #1316, opened 2023-08-04
btf: Add support to load symbols from kernel modules

tpapagian:
Adding a kprobe hook to a function that defined inside a module is not currently supported.

This commit adds support for that.

The user can use a command line argument --kmods to define the desired modules. Adding --kmods overlay,tls will search for btf files in /sys/kernel/btf/overlay and /sys/kernel/btf/tls.

### DIFF
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -4,9 +4,12 @@
 package btf
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
+	"strings"
 
 	"github.com/cilium/ebpf/btf"
 	"github.com/cilium/tetragon/pkg/defaults"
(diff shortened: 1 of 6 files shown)


pkg/btf/btf.go
```diff
@@ -87,6 +90,57 @@ func NewBTF() (*btf.Spec, error) {
...
+	iter := spec.Iterate()
+	for iter.Next() {
+		allTypes = append(allTypes, iter.Type)
+	}
+
+	for _, module := range kmods {
+		path := filepath.Join("/sys/kernel/btf", module)
+		f, err := os.Open(path)
+		if err != nil {
+			logger.GetLogger().Warnf("btf: Path %s does not exist", path)
```
kkourt, line 106: nit: It's more idiomatic to use structured logging: logger.GetLogger().WithError(err).WithField("path", path).Warnf(...)

pkg/btf/btf.go
```diff
@@ -87,6 +90,57 @@ func NewBTF() (*btf.Spec, error) {
...
+	}
+
+	raw, err := b.Marshal(nil, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to call b.Marshal: %w", err)
+	}
+
+	spec, err = btf.LoadSpecFromReader(bytes.NewReader(raw))
+	if err != nil {
+		return nil, fmt.Errorf("failed to call bbtf.LoadSpecFromReader: %w", err)
```
kkourt, line 138: nit: typo bbtf

kkourt (approved):
LGTM, just a couple of nits.
Thanks!

pkg/btf/btf.go
tpapagian: Fixed!

pkg/btf/btf.go
tpapagian: Thanks, fixed!

Issue #1310, opened 2023-08-02
Add the "NotPrefix" and "NotPostfix" Operators to suitable Selector Types


BonySmoke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

We would like to block all write requests to files under a particular directory with the exception of certain folders.
Following the existing [example](https://github.com/cilium/tetragon/blob/19fdc3a56629e5396477baa776c599d2bbc76299/examples/tracingpolicy/open_kill.yaml), we can kill all processes that try to write to any file under the `/home` folder using the `Prefix` operator:
```yaml
  kprobes:
...
  - call: "sys_write"
    syscall: true
    args:
    - index: 0
      type: "fd"
    - index: 1
      type: "char_buf"
      sizeArgIndex: 3
    - index: 2
      type: "size_t"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/home/"
      matchActions:
      - action: Sigkill
```
However, we have folders under `/home` where we still would like to allow write requests.

If I understand it correctly, currently, it's only possible to add exact files to exceptions using the `NotEqual` operator.
```yaml
...
    selectors:
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/home/"
      - index: 0
        operator: "NotEqual"
        values:
        - "/home/anotherfolder/file.txt"
      matchActions:
      - action: Sigkill
```
Unfortunately, we cannot predict what files/directories will be in the `/home/anotherfolder` folder.

### Describe the feature you would like

It would be great to add a `NotPrefix` and optionally `NotPostfix` operators to the suitable selector types to make filtering much more flexible.
Using the example above, we could end up with the following selector:
```yaml
...
    selectors:
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/home/"
      - index: 0
        operator: "NotPrefix"
        values:
        - "/home/anotherfolder/"
      matchActions:
      - action: Sigkill
```
And this would translate to `(filepath starts with "/home/" AND filepath NOT starts with "/home/anotherfolder/")`

### Describe your proposed solution

I saw this [commit](https://github.com/cilium/tetragon/commit/edcebe77b5f52cd601d007661d07c8bcbf622605) adding a negative filter. Maybe the implementation of this feature will be similar.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kkourt:
[...]

PR #1325, opened 2023-08-07
matchArgs: Add Not{Equal, Prefix, Postfix} in fd, file, path types

tpapagian:
FIXES: https://github.com/cilium/tetragon/issues/1310

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -654,34 +654,32 @@ struct string_buf {
 	char buf[];
 };
 
+#define FILTER_FILE_MATCH   0
+#define FILTER_FILE_NOMATCH 1
+
 static inline __attribute__((always_inline)) long
 __filter_file_buf(struct string_buf *value, struct string_buf *args, __u32 op)
 {
 	__u32 v = value->len, a = args->len;
 	int err;
 
-	/* There are cases where file pointer may not contain a path.
-	 * An example is using an unnamed pipe. This is not a match.
-	 */
-	if (a == 0)
-		goto skip_string;
-	if (op == op_filter_eq) {
+	if (op == op_filter_eq || op == op_filter_neq) {
 		if (v != a)
 			goto skip_string;
-	} else if (op == op_filter_str_prefix) {
+	} else if (op == op_filter_str_prefix || op == op_filter_str_notprefix) {
 		if (a < v)
 			goto skip_string;
-	} else if (op == op_filter_str_postfix) {
+	} else if (op == op_filter_str_postfix || op == op_filter_str_notpostfix) {
 		err = rcmpbytes(value->buf, args->buf, v - 1, a - 1);
 		if (!err)
-			return 0;
+			return FILTER_FILE_MATCH;
 		goto skip_string;
 	}
 	err = cmpbytes(value->buf, args->buf, v);
 	if (!err)
-		return 0;
+		return FILTER_FILE_MATCH;
 skip_string:
-	return v + sizeof(struct string_buf);
+	return FILTER_FILE_NOMATCH;
 }
 
 struct file_buf {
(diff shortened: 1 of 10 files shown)


jrfastab (approved):

bpf/process/types/basic.h
```diff
@@ -697,26 +695,41 @@ struct file_buf {
...
 	char *values = (char *)filter + sizeof(struct file_buf); // all values are after the header
-	__u32 remain = filter->total_sz - sizeof(__u32) - sizeof(__u32);
-	int i, next;
+	__u32 next, remain = filter->total_sz - sizeof(__u32) - sizeof(__u32);
+	long i, match;
+
+	/* There are cases where file pointer may not contain a path.
+	 * An example is using an unnamed pipe. This is not a match.
+	 */
+	if (args->len == 0)
```
kkourt, line 707: Haven't followed the code closely, but shouldn't we return  for the `not` operators? Maybe `goto: done_filter_file`?

bpf/process/types/basic.h
```diff
@@ -697,26 +695,41 @@ struct file_buf {
...
-		if (!next)
+		struct string_buf *value = (struct string_buf *)values;
+
+		match = __filter_file_buf(value, args, filter->op);
+		if (filter->op == op_filter_neq || filter->op == op_filter_str_notprefix || filter->op == op_filter_str_notpostfix) {
+			if (match != FILTER_FILE_NOMATCH)
+				return 0;
+		} else if (match == FILTER_FILE_MATCH) {
 			return 1;
+		}
```
kkourt, line 722: I'd personally write this as:

```
....
// the file matched, which means that we can now make a decision.
if (match == FILTER_FILE_MATCH) {
     return is_not_operator(filter->op) ? 0 : 1; 
}
```

Where:
```
bool is_not_operator(op) {
    return (filter->op == op_filter_neq || filter->op == op_filter_str_notprefix || filter->op == op_filter_str_notpostfix);
}
```

But that's just me :)

kkourt (commented):
Thanks! I have some comments.


> The behaviour changes a bit when having multiple values. For operators Equal, Prefix, and Postfix the
> previous example will match when:
> [(arg OP val1) OR (arg OP val2) OR (arg OP val3)]
> 
> For operators NotEqual, NotPrefix, and NotPostfix the previous example will match when:
> [(arg OP val1) AND (arg OP val2) AND (arg OP val3)]

I think the right way to think of Not* operators is:

Equal: (arg == val1) || (arg == val2) || (arg == val3)
NotEqual: NOT(Equal)
 => NOT( (arg == val1) || (arg == val2) || (arg == val3) )
 => (arg != val1) && (arg != val2) && (arg != val3)
 
 So the behavior doesn't really change.

bpf/process/types/basic.h
tpapagian: In that case, we get an event where the path is an empty string. I don't find much value in having events like this when trying to monitor access to files. This is the reason I kept the same behavior for not operators. 

If this does not make sense, I can change that.

bpf/process/types/basic.h
tpapagian: Simplifies the code a bit. Done!

bpf/process/types/basic.h
kkourt: OK, that makes sense to me. Thanks!

kkourt (approved):

tixxdz:
This broken main, how did it pass tests?

tixxdz:
ah the kprobe_test doesn't have the observer package since it was moved so the tests need to rebase and use the new observerhelper package

tpapagian:
> ah the kprobe_test doesn't have the observer package since it was moved so the tests need to rebase and use the new observerhelper package

Fix: https://github.com/cilium/tetragon/pull/1347

PR #1335, opened 2023-08-08
Move observer_test_helpers.go to a separate package

lambdanis:
I hit an import cycle in `observer_test_helpers.go` when working on #1304, so I re-implemented `metrics.EnableMetrics()` there instead of importing the metrics package. Then I hit the same import cycle when working on #1279, so here I moved `observer_test_helpers.go` to a separate package.

It's a pretty straightforward refactoring, see commits for details. Technically it is a breaking change, but only developer-facing.

### DIFF
--- a/pkg/bench/trace_bench_rw.go
+++ b/pkg/bench/trace_bench_rw.go
@@ -14,7 +14,7 @@ import (
 	"text/template"
 	"time"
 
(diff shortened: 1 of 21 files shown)


jrfastab (approved):
ship it

kkourt (approved):

Issue #794, opened 2023-03-11
Remove remaining Cilium dependency from the Tetragon project


sharlns:
Tetragon can run both with and without Cilium on the same node. Some functionality, however, still depends on the Cilium agent being present. Specifically, Tetragon uses Cilium to retrieve the pod information for destination IPs for pods which are not local to the node. The goal of this project is to introduce this functionality on Tetragon. One approach would be for the Tetragon agent to keep information about all pods in the cluster, but this approach does not scale well due to the k8s API server  needing to propagate all pod information to all nodes. Instead, the plan to introduce a new custom resource (CR) which is maintained by the Tetragon operator and provides a mapping from IPs to the small subset of pod information that Tetragon needs. The Tetragon operator will monitor pod information and update the resource as needed. Tetragon agents will watch this CR to provide pod information for destination IPs.

please feel free to contact michi@isovalent.com (github id: @michi-covalent) if you'd like to get some feedback for your draft proposal before the application deadline.

Lan-ce-lot:
Hi @sharlns , I am astudent from SEL laboratory of Zhejiang University, familiar with cloud native, kubernetes, docker and Go. I'm learning eBPF and I also participated in a eBPF project. Therefore, I think this project is quite suitable for me. I plan to apply for GSOC 2023 and apply for this project. Could you have more suggestions to help me get started?

xmulligan:
Hi @Lan-ce-lot Thanks for your interest in the project. If you want to get started working with the project, I would suggest checking out the [getting started guide](https://github.com/cilium/tetragon#kubernetes-quickstart-guide) and some of the [good first issues](https://github.com/cilium/tetragon/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22). This project will be worked on whoever is selected as the GSoC mentee

Lan-ce-lot:
Thanks for your advice @xmulligan.

prateek041:
I came across this issue a few days ago, i am interested in contributing to it. Due to this issue, I started learning about CRDs and operators. As I learn more, I'm understanding the issue and the code better. I am also working on my proposal simultaneously, @xmulligan it would be great help if you could provide some review to my proposal.

xmulligan:
Thanks for your interest @prateek041. Unfortunately, we cannot review your proposal beforehand because it would be unfair to the other applicants

prateek041:
[...]

PR #1336, opened 2023-08-08
refactor: Move readConfig{Dir,File} to the option package

michi-covalent:
Move readConfig{Dir,File} to the option package. I'd like to use these functions to read tetragon operator configurations.

Ref: #794

### DIFF
--- a/cmd/tetragon/conf.go
+++ b/cmd/tetragon/conf.go
@@ -3,9 +3,6 @@
 package main
 
 import (
-	"fmt"
-	"os"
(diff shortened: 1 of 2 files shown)


tixxdz (approved):
Great!  maybe have /etc/tetragon/operator.conf.d/ for its config ? if that works...

michi-covalent:
> Great! maybe have /etc/tetragon/operator.conf.d/ for its config ? if that works...

yeah that's the idea. opening a followup PR soon 🚀

PR #1337, opened 2023-08-08
Add tetragon-operator-config ConfigMap

michi-covalent:
- Add tetragon-operator-config ConfigMap.
- Add tetragonOperator.skipCRDCreation Helm value.
- Mount the ConfigMap to /etc/tetragon/operator.conf.d/ and load the
  config from the directory.
- Log the config at the startup.

Ref: #794

sample log output:
```
level=info msg="Loaded config from directory" config-dir=/etc/tetragon/operator.conf.d/ subsys=tetragon-operator
level=info msg="Starting Tetragon Operator" config="&{SkipCRDCreation:false KubeCfgPath: ConfigDir:/etc/tetragon/operator.conf.d/}" subsys=tetragon-operator version=v0.10.0-145-g8b2a7e8f
level=info msg="CRD (CustomResourceDefinition) is installed and up-to-date" name=TracingPolicy/v1alpha1 subsys=k8s
level=info msg="CRD (CustomResourceDefinition) is installed and up-to-date" name=TracingPolicyNamespaced/v1alpha1 subsys=k8s
level=info msg="Initialization complete" subsys=tetragon-operator
```

### DIFF
--- a/docs/content/en/docs/reference/helm-chart.md
+++ b/docs/content/en/docs/reference/helm-chart.md
@@ -99,6 +99,7 @@ To use [the values available](#values), with `helm install` or `helm upgrade`, u
 | tetragon.securityContext.privileged | bool | `true` |  |
 | tetragonOperator.enabled | bool | `true` | Enable the tetragon-operator component (required). |
 | tetragonOperator.image | object | `{"override":null,"repository":"quay.io/cilium/tetragon-operator","suffix":"","tag":"v0.10.0"}` | tetragon-operator image. |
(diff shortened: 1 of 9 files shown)


mtardy (approved):
Makes sense!!

~This is out of the scope of this PR and maybe that's already the project but I guess it would make sense that the tetragon-operator is not an init-container per daemon running in k8s but just a deployment with only 1 instance that is long-running right?~ I guess this is completely what you are trying to do atm.

PR #1349, opened 2023-08-10
bpf: fmodret override on security_ hooks is available from 5.7

tixxdz:
The override on security_ hook is available starting from 5.7, so check before loading the policy.

https://lore.kernel.org/all/20200304191853.1529-5-kpsingh@chromium.org/

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy.md
+++ b/docs/content/en/docs/concepts/tracing-policy.md
@@ -983,9 +983,10 @@ The difference is to use the signal action with `SIGKILL(9)` signal.
 
 `Override` action allows to modify the return value of call. While `Sigkill`
 will terminate the entire process responsible for making the call, `Override`
-will override the return value that was supposed to be returned with the value
-given in the `argError` field. It's then up to the process handling of the
-return value of the function to stop or continue the execution.
+will run in place of the original kprobed function and return the value
+specified in the `argError` field. It's then up to the code path or the user
+space process handling the returned value to whether stop or proceed with the
+execution.
 
 For example, you can create a `TracingPolicy` that intercepts `sys_symlinkat`
 and will make it return `-1` every time the first argument is equal to the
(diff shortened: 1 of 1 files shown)


olsajiri (approved):
looks good, could that be done with some feature check as well? I'd think Redhat kernel might have this backported even if the kernel version is 4.18.. but we can start with this for sure

kkourt:
> looks good, could that be done with some feature check as well? I'd think Redhat kernel might have this backported even if the kernel version is 4.18.. but we can start with this for sure

kkourt:
> looks good, could that be done with some feature check as well? I'd think Redhat kernel might have this backported even if the kernel version is 4.18.. but we can start with this for sure

Agreed.
What would that feature check look like? I guess we could try and loading a program and see if it works?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -310,8 +310,11 @@ func preValidateKprobes(name string, kprobes []v1alpha1.KProbeSpec) error {
 			if !bpf.HasOverrideHelper() {
 				return fmt.Errorf("Error override action not supported, bpf_override_return helper not available")
 			}
-			if !f.Syscall && strings.HasPrefix(f.Call, "security_") == false {
-				return fmt.Errorf("Error override action can be used only with syscalls and security_ hooks")
+			// If not a system call and not a security_ hook on minimum kernel 5.7 fail the override
+			// https://lore.kernel.org/all/20200304191853.1529-5-kpsingh@chromium.org/
+			if !f.Syscall &&
+				(strings.HasPrefix(f.Call, "security_") == false || !kernels.MinKernelVersion("5.7")) {
+				return fmt.Errorf("Error override action can be used only with syscalls or on security_ hooks with a minimum kernel version 5.7")
```
kkourt, line 317: Until we have a feature check, would it make sense to have a flag so that we can ignore the kernel check here? As @olsajiri said, some vendor kernels backport many patches and the kernel version does not necessarily reflect the feature set of the kernel, and it would be nice to have a way to override the kernel check.

We can leave this as a follow up though.

kkourt (approved):

tixxdz:
> > looks good, could that be done with some feature check as well? I'd think Redhat kernel might have this backported even if the kernel version is 4.18.. but we can start with this for sure
> 
> Agreed. What would that feature check look like? I guess we could try and loading a program and see if it works?

Oh I see redhat kernels, so in that case I will convert this patch to be: document where override applies syscalls, security_ hooks with the corresponding kernels version or patch, and let it fail for other cases. Thoughts?

tixxdz:
> > > > looks good, could that be done with some feature check as well? I'd think Redhat kernel might have this backported even if the kernel version is 4.18.. but we can start with this for sure
> > > 
> > > 
> > > Agreed. What would that feature check look like? I guess we could try and loading a program and see if it works?
> > 
> > 
> > Oh I see redhat kernels, so in that case I will convert this patch to be: document where override applies syscalls, security_ hooks with the corresponding kernels version or patch, and let it fail for other cases. Thoughts?
> 
> You mean not have the kernel check? Yes, I think that's fine from my side.

Yes, improved the doc. If no objection will merge it later

mtardy:
Thanks!

PR #1350, opened 2023-08-11
operator: Move the CRD registration logic to a separate package

michi-covalent:
We'll introduce a proper operator deployment in #794. Once the operator deployment is available, we can move the CRD registration logic there instead of calling it from an init container in the Tetragon daemonset. This commit moves the CRD registration logic to a separate package so that it can be called from outside the main package.


jrfastab (commented):
sure

kkourt (approved):

PR #1358, opened 2023-08-15
Sock/skb: socket improvements

kevsecurity:
See commits for details.

Switch protocol selector match checking from array of (up to 4) values that are checked in a loop to a map look up as per port look ups.

Add state and family matchers.

Improve socket tracking for situations where processes have exited before a socket event is reported (for example, __sk_free can be triggered after the owning process has exited). To support this we store the process information (not just PID/TGID) in the socket tracking map so it can be reported in an event without having to look it up. This means that a process can exit and its entry in the execve map can be deleted, but we can still report the correct information, and rely on the fact that user space will still have a record of it in its cache for long enough to handle our event.


pkg/selectors/kernel.go
```diff
@@ -392,6 +421,16 @@ func writeMatchRangesInMap(k *KernelSelectorState, values []string, ty uint32) e
...
 			rangeStr = append(rangeStr, rangeStr[0])
+
+			// Special actions for particular network ops
+			switch op {
+			case SelectorOpProtocol:
+				protocol, err := network.InetProtocolNumber(v)
+				if err == nil {
+					protocolStr := fmt.Sprintf("%d", protocol)
+					rangeStr = []string{protocolStr, protocolStr}
+				}
```
kkourt, line 441: Can we return an error here? We do fail silently in other places, but I don't think is a good practice.

pkg/selectors/kernel.go
```diff
@@ -430,6 +439,18 @@ func writeMatchRangesInMap(k *KernelSelectorState, values []string, ty uint32, o
 					protocolStr := fmt.Sprintf("%d", protocol)
 					rangeStr = []string{protocolStr, protocolStr}
 				}
+			case SelectorOpFamily:
+				family, err := network.InetFamilyNumber(v)
+				if err == nil {
+					familyStr := fmt.Sprintf("%d", family)
+					rangeStr = []string{familyStr, familyStr}
+				}
```
kkourt, line 447: ditto for failing silently.

pkg/selectors/kernel.go
```diff
@@ -430,6 +439,18 @@ func writeMatchRangesInMap(k *KernelSelectorState, values []string, ty uint32, o
...
+			case SelectorOpState:
+				state, err := network.TcpStateNumber(v)
+				if err == nil {
+					stateStr := fmt.Sprintf("%d", state)
+					rangeStr = []string{stateStr, stateStr}
+				}
```
kkourt, line 453: ditto for failing silently.

kkourt (changes requested):
Thanks!

Had a quick look and left some first comments.

pkg/selectors/kernel.go
kevsecurity: We don't need to as it gets picked up when trying to convert to a number. This part just tries to convert the text string as a IPPROTO_ string. If it works, great, but if it doesn't then the string falls through and is attempted to be converted to a number in the following loop. So the logic was to try the special case, and if it fails, do the usual case.

pkg/selectors/kernel.go
kevsecurity: Ditto

pkg/selectors/kernel.go
kevsecurity: Ditto

pkg/selectors/kernel.go
kkourt: I see. Thanks!

kkourt (approved):
Thanks!

PR #1361, opened 2023-08-16
sock/skb: Add IPv6 Support

kevsecurity:
The sock and skb types currently support IPv4 addresses and CIDRs, and in fact the whole subsystem assumes network operations are only IPv4 or AF_INET.

This commit adds IPv6 support by allowing IPv6 addresses and CIDRs to be specified for arg matching for sock and skb types; collects IPv6 addresses from sockets and datagrams where the family is AF_INET6; and reports said addresses correctly in user space.

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -771,20 +771,40 @@ struct ip_ver {
 // use the selector value to determine a LPM Trie map, and do a lookup to determine whether the argument
 // is in the defined set.
 static inline __attribute__((always_inline)) long
-filter_addr4_map(struct selector_arg_filter *filter, __u32 addr)
+filter_addr_map(struct selector_arg_filter *filter, __u64 *addr, __u16 family)
 {
 	void *addrmap;
-	__u32 map_idx = filter->value;
-	struct addr4_lpm_trie arg;
+	__u32 *map_idxs = (__u32 *)&filter->value;
+	__u32 map_idx;
+	struct addr4_lpm_trie arg4;
+	struct addr6_lpm_trie arg6;
+	void *arg;
 
-	addrmap = map_lookup_elem(&addr4lpm_maps, &map_idx);
-	if (!addrmap)
+	switch (family) {
+	case AF_INET:
+		map_idx = map_idxs[0];
+		addrmap = map_lookup_elem(&addr4lpm_maps, &map_idx);
+		if (!addrmap)
+			return 0;
+		arg4.prefix = 32;
+		arg4.addr = addr[0];
+		arg = &arg4;
+		break;
+	case AF_INET6:
+		map_idx = map_idxs[1];
+		addrmap = map_lookup_elem(&addr6lpm_maps, &map_idx);
+		if (!addrmap)
+			return 0;
+		arg6.prefix = 128;
+		// write the address in as 4 u32s due to alignment
+		write_ipv6_addr32(arg6.addr, (__u32 *)addr);
+		arg = &arg6;
+		break;
+	default:
 		return 0;
+	}
 
-	arg.prefix = 32;
-	arg.addr = addr;
-
-	__u8 *pass = map_lookup_elem(addrmap, &arg);
+	__u8 *pass = map_lookup_elem(addrmap, arg);
 
 	switch (filter->op) {
 	case op_filter_saddr:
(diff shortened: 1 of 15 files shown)


bpf/process/types/tuple.h
```diff
@@ -7,13 +7,39 @@
...
 	__u16 dport;
 	__u16 protocol;
 	__u16 family;
 };
 
+static inline __attribute__((always_inline)) void
+write_ipv6_addr_from_ipv4(u64 *dest, u32 src)
+{
+	dest[0] = src;
+	dest[1] = 0;
```
kkourt, line 26: AFAICT, we keep everything in host order, but I'm not sure if above works for big-endian hosts.
I guess it does not matter much in practice before even for ARM. I think it supports both modes, and most people will run it as little endian.

bpf/process/types/basic.h
```diff
@@ -771,20 +771,40 @@ struct ip_ver {
...
+		arg4.prefix = 32;
+		arg4.addr = addr[0];
+		arg = &arg4;
+		break;
+	case AF_INET6:
+		map_idx = map_idxs[1];
+		addrmap = map_lookup_elem(&addr6lpm_maps, &map_idx);
+		if (!addrmap)
+			return 0;
+		arg6.prefix = 128;
```
kkourt, line 798: Out of curiosity, what is the reason for using an lpm tree if we are not matching prefixes. Do we plan to support matching besides /32, or just to save space?

kkourt (approved):
LGTM, thanks.

bpf/process/types/basic.h
kevsecurity: We are matching prefixes! The user can specify CIDRs with /nn netmask lengths and these are entered into the map. We then observe an IP address and ask for the longest prefix that matches it, specifying how many bits we have observed, which is the whole of the IP address so 32 bits for IPv4 and 128 bits for IPv6.
For example, we could store 127.0.0.0/8 in the map, and then observe 127.0.0.1 as an IP address. This will return the 127.0.0.0/8 match.

bpf/process/types/skb.h
```diff
@@ -16,6 +16,104 @@ struct skb_type {
...
+
+	e = map_lookup_elem(&tg_ipv6_ext_heap, &zero);
+	if (!e)
+		return 0;
+
+	e->ip_off = network_header_off;
+	e->curr = 255;
+	e->len = 0;
+	if (probe_read(&e->next, sizeof(e->next), _(&ip->nexthdr)) < 0)
+		return 0;
```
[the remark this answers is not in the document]
kevsecurity: As you state, 0 is an IPv6 header extension rather than a protocol so we'd never return it as a valid response.

bpf/process/types/tuple.h
kevsecurity: We store everything in network order, but we don't have any code to support big endian currently. I don't think it's a big deal, but it would be relatively easy to add if it became important.

bpf/process/types/basic.h
```diff
@@ -771,20 +771,40 @@ struct ip_ver {
...
+		// write the address in as 4 u32s due to alignment
+		write_ipv6_addr32(arg6.addr, (__u32 *)addr);
+		arg = &arg6;
+		break;
+	default:
 		return 0;
```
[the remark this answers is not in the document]
kkourt: So if a user says "NotDAddr 127.0.0.1", then we will basically match on (ipv6 || ipv4) && (Daddr != 127.0.0.1") and we will not match, for example, ARP packets. OK!

bpf/process/types/basic.h
kkourt: Thanks!

bpf/process/types/tuple.h
kkourt: Do we? I guess I misread the code. Let me check...

I guess that's what I saw:
>	event->tuple.sport = bpf_ntohs(event->tuple.sport);
>	event->tuple.dport = bpf_ntohs(event->tuple.dport);

So eveht->tuple has addresses in network order and ports in host order? Or am I missing something again?

kevsecurity: No you're correct, I was only thinking about IP addresses. Ports are indeed in host order.

PR #1368, opened 2023-08-18
tetragon: Factor InMap operators

olsajiri:
We are going to use InMap operators for loading values from lists,
so moving the network port parsing code into separate function,
because not every possible map value is a port number ;-)

### DIFF
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -480,7 +480,7 @@ func writeMatchRangesInMap(k *KernelSelectorState, values []string, ty uint32, o
 			for val := sRangeVal[0]; val <= sRangeVal[1]; val++ {
 				var valByte [8]byte
 				binary.LittleEndian.PutUint64(valByte[:], uint64(val))
-				m[valByte] = struct{}{}
+				m.Data[valByte] = struct{}{}
 			}
 
 		case argTypeU64:
@@ -490,7 +490,7 @@ func writeMatchRangesInMap(k *KernelSelectorState, values []string, ty uint32, o
 			for val := uRangeVal[0]; val <= uRangeVal[1]; val++ {
 				var valByte [8]byte
 				binary.LittleEndian.PutUint64(valByte[:], val)
-				m[valByte] = struct{}{}
+				m.Data[valByte] = struct{}{}
 			}
 		}
 	}
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -507,11 +507,43 @@ func writeMatchRangesInMap(k *KernelSelectorState, values []string, ty uint32, o
 	return nil
 }
 
+func writeListValuesInMap(k *KernelSelectorState, v string, ty uint32, m *ValueMap) error {
+	if k.listReader == nil {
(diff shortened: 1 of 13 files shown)


pkg/selectors/kernel.go
```diff
@@ -411,88 +411,131 @@ func argSelectorType(arg *v1alpha1.ArgSelector, sig []v1alpha1.KProbeArg) (uint3
...
 
+func writePortRangeInMap(v string, ty uint32, op uint32, m *ValueMap) error {
+	// We store the start and end of the range as uint64s for unsigned values, and as int64s
+	// for signed values. This is to allow both a signed range from -5 to 5, and also an
+	// unsigned range where at least limit exceeds max(int64). If we only stored the range
+	// limits as uint64s, then the range from -5 to 5 would be interpretted as being from
+	// 5 to uint64(-5), which is the literal opposite of what was intended. If we only stored
+	// the range as int64s, then we couldn't correctly accommodate values that exceed max(int64),
+	// for similar reasons.
+	var uRangeVal [2]uint64
```
kevsecurity, line 422: If this function is only handling port ranges (and not ranges of anything else) then it could all be switched to unsigned values for simplicity.

pkg/selectors/kernel.go
```diff
@@ -411,88 +411,131 @@ func argSelectorType(arg *v1alpha1.ArgSelector, sig []v1alpha1.KProbeArg) (uint3
...
+	WriteSelectorUint32(k, mid)
+	return nil
+}
+
+func writeMatchValuesInMap(k *KernelSelectorState, values []string, ty uint32, op uint32) error {
+	mid, m := k.newValueMap()
+	for _, v := range values {
+		var val [8]byte
+		// most likely port range
+		if strings.Contains(v, ":") {
```
kevsecurity, line 515: Do we ever call this function with a port range, or does that go straight to writeMatchRangesInMap() instead? If not, then maybe remove this check and branch.

olsajiri: I assumed that much, it seems to happened in this commit: 92cd5e964de2 Permit any number of port numbers in matches

I can remove that if it's not the case, it looks like there are special operators for ports

pkg/selectors/kernel.go
```diff
@@ -411,88 +411,131 @@ func argSelectorType(arg *v1alpha1.ArgSelector, sig []v1alpha1.KProbeArg) (uint3
...
 		switch ty {
 		case argTypeS64, argTypeInt:
-			if sRangeVal[0] > sRangeVal[1] {
-				sRangeVal[0], sRangeVal[1] = sRangeVal[1], sRangeVal[0]
-			}
-			for val := sRangeVal[0]; val <= sRangeVal[1]; val++ {
-				var valByte [8]byte
-				binary.LittleEndian.PutUint64(valByte[:], uint64(val))
-				m.Data[valByte] = struct{}{}
+			i, err := strconv.ParseInt(v, 10, 64)
```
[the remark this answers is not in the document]
olsajiri: sure, my use case is to load values from list, so no need for that at the moment

kkourt (approved):
LGTM, thanks.

tpapagian (approved):

PR #1375, opened 2023-08-21
tetragon-oci-hook: fix issue for containerd

kkourt:
containerd seems to be using a different label key for the container namespace than cri-o, which results in the fail detection of tetragon-oci-hook to not work properly because the teragon container cannot start.

While this can be fixed via appropriate arguments to the hook, we also want to have it working by default. So this patch:
 - changes the cel expression to support multiple keys
 - adds the key used by containerd (io.kubernetes.cri.sandbox-namespace) to the one used by cri-o (io.kubernetes.cri.sandbox-namespace).

The patch also adds a --debug option, and cleans up some dead code.

This issue was reported on slack by @akshay196.


willfindlay (approved):
Not sure if I know enough about this component to give a proper review but everything looks solid to me.

PR #1380, opened 2023-08-21
README: simplify the README to refer to the documentation

mtardy:
A lot of information was moved to the documentation so we ended up with duplicate of stuff between the doc website and the README. This should simplify the situation with a smaller README redirecting to the more complete documentation.

[Preview the README](https://github.com/cilium/tetragon/blob/b72a6dc9a9dd99e13576c42e87354203bbe2e039/README.md)


mtardy:
I'm not sure of this README, I think there's maybe too much hyperlink. If you have suggestions that would be helpful. Let me ping @xmulligan as well maybe he has an opinion on READMEs? :)

kkourt (approved):
I think it's an improvement over the current one, so LGTM!

PR #1384, opened 2023-08-23
gha: add pr-validation GH action

kkourt:
This action validates PRs based on labels.
It blocks PRs that have a dont-merge/* label, and blocks PRs that do not have a release-note/* labe.

### DIFF
--- a/.github/workflows/pr-validation.yaml
+++ b/.github/workflows/pr-validation.yaml
@@ -1,0 +1,36 @@
+name: pr
+
+on:
+  pull_request:
+    branches:
+      - main
+      - v*
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - edited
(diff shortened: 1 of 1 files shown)


mtardy:
> and blocks PRs that do not have a release-note/* label.

Do you think all PRs should be labeled with some sort of release-note label?

kkourt:
> > and blocks PRs that do not have a release-note/* label.
> 
> Do you think all PRs should be labeled with some sort of release-note label?

yes.

michi-covalent (approved):
i love dont-merge labels 🥰

michi-covalent:
we should just configure renovate to add a label, maybe `release-note/dependency` or something. then we we can group these PRs in the release note, or even exclude them if we decide not to include them in the release note 🥰

PR #1385, opened 2023-08-24
sensors: modify unload hooks and remove GetConfig and SetConfig

mtardy:
See individual commits, the last one removes a deprecated and unused thing in sensors.

These patches were made in order for us (me) to be able to use `PostUnloadHook` later on.


pkg/sensors/load.go
```diff
@@ -139,6 +139,13 @@ func (s *Sensor) Unload() error {
 	}
 
 	s.Loaded = false
+
+	if s.PostUnloadHook != nil {
+		if err := s.PostUnloadHook(); err != nil {
+			logger.GetLogger().Warnf("Sensor %s post unload hook failed: %s", s.Name, err)
```
kkourt, line 145: nit: how about we use structured logging here?

kkourt (approved):
LGTM, but it would be great if you could add a test to ensure that removing the entries works as expected.

mtardy:
> LGTM, but it would be great if you could add a test to ensure that removing the entries works as expected.

Thanks will add a test in this PR!

PR #1386, opened 2023-08-24
improve tetra tracingpolicy generate

kkourt:
See patches.

### DIFF
--- a/pkg/tracingpolicy/generate/generate.go
+++ b/pkg/tracingpolicy/generate/generate.go
@@ -1,0 +1,32 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+package generate
+
+import (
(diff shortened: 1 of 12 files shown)


olsajiri (commented):
I'm getting fail with generated policy:

`time="2023-08-25T06:56:08Z" level=fatal msg="Failed to start tetragon" error="error unmarshaling JSON: while decoding JSON: json: unknown field \"creationTimestamp\""
`

also what's omitempty for?

olsajiri (approved):

pkg/tracingpolicy/generate/generate.go
```diff
@@ -0,0 +1,30 @@
+package generate
```
lambdanis, line 1: nit: SPDX header is missing here

kkourt: Thanks, fixed!

PR #1401, opened 2023-08-29
tetragon: Adding lists documentation

olsajiri:
Adding documentation for lists.

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -83,6 +83,187 @@ spec:
     # [...]
 ```
 
+### Lists
+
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -83,6 +83,187 @@ spec:
...
+  - call: "sys_dup"
+    syscall: true
+  - call: "sys_dup2"
+    syscall: true
+  - call: "sys_dup3"
+    syscall: true
+```
+
+In does not seem like saving much, but it starts to make sense when you put in place
+additional filters.
```
kkourt, line 120:
```suggestion
As shown in subsequent examples, its main benefit is allowing a single definition for calls that have the same filters.
```

kkourt (approved):
🚀

PR #1402, opened 2023-08-29
cmd/tetra: retry failed gRPC connections with exponential backoff

willfindlay:
This PR adds two new command line options:

- `--timeout` allows the user to configure the desired gRPC connection timeout, instead of hard coding it at 10 seconds
- `--retries` sets the number of attempts we should make to connect to the gRPC server with exponential backoff


michi-covalent (approved):
😇

Issue #1395, opened 2023-08-28
Update homepage footer


xmulligan:
### What happened?

Since we are a CNCF project, we should follow the [website guidelines](https://github.com/cncf/foundation/blob/main/website-guidelines.md). Specifically:

1. Remove the year from the authors
2. Add the trademark statement (see # 7 above)
3. credit Isovalent as the original author "Tetragon was originally created by Isovalent"

you can see the cilium website as an example https://cilium.io/

### Tetragon Version

N/A

### Kernel Version

N/A

xmulligan:
cc: @mtardy

PR #1404, opened 2023-08-30
docs: fix footer

kkourt:
Fixes: #1395


kkourt:
Thanks, added suggestions.

mtardy:
we need to fix the size the header and the footer takes on the docs pages 
image

on a "normal" MacBook screen it's like 30% taken by those which is a bit annoying, but this is a task for later :)

Thanks kornilios!

PR #1407, opened 2023-08-30
helm: update livenessProbe to retry 5 times before failing

willfindlay:
We noticed issues in testing where gRPC queries would sometimes spuriously time out, causing our livenessProbe to fail and the Tetragon pod to restart. 7f2edda4 and 4f4feb8c introduced new configuration values in our CLI to allow it to re-attempt the connection in case of failure. Update our livenessProbe in helm to use these new configuration values so that we can re-attempt the connection.

### DIFF
--- a/install/kubernetes/templates/_container_tetragon.tpl
+++ b/install/kubernetes/templates/_container_tetragon.tpl
@@ -68,12 +68,15 @@
 {{- end }}
 {{- if .Values.tetragon.grpc.enabled }}
   livenessProbe:
+     timeoutSeconds: 60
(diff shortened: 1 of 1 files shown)


jrfastab:
Lgtm

jrfastab:
What was the old timeout default value?

willfindlay:
@jrfastab it was 1 second with no retries. Now it's 60 with 5 retries (each attempt takes 10 seconds with some extra time for exponential backoff).

jrfastab (approved):

Issue #1278, opened 2023-07-26
Add the "Prefix" operator for the "matchBinaries" selector


Vampouille:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

I would like to forbid network usage for every binaries in a specific folder. For example, binaries in `~/bin-no-network/` should not be able to create socket. This can be done with the following TracingPolicy: 

```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "connect"
spec:
  kprobes:
  - call: "tcp_connect"
    syscall: false
    args:
    - index: 0
      type: "sock"
    selectors:
    - matchBinaries:
      - operator: "In"
        values:
          - "/home/jacroute/bin-no-network/curl"
          - "/home/jacroute/bin-no-network/jq"
      matchActions:
      - action: Sigkill
```

But we need to list every binary in this folder and keep this list up-to-date.

### Describe the feature you would like

I can be nice to use the "Prefix" operator for this use-case but only "In" and "NotIn" operators are currently [implemented](https://github.com/cilium/tetragon/pull/686/files#diff-0b6c66cf370068fe2e441466d6124582732e8b67dd0daddccf72c7eb2bef16d4R994-R1000)

With this new feature, the TracingPolicy should look like : 

```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "connect"
spec:
  kprobes:
  - call: "tcp_connect"
    syscall: false
    args:
    - index: 0
      type: "sock"
    selectors:
    - matchBinaries:
      - operator: "Prefix"
        values:
          - "/home/jacroute/bin-no-network/"
      matchActions:
      - action: Sigkill
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mtardy:
So I spent some time on this in the light of https://github.com/cilium/tetragon/pull/1408. I think this is doable against a bit of work. This PR by Anastasios explains how matchBinaries work in details https://github.com/cilium/tetragon/pull/774#issue-1615126127. 

But the tl;dr is that we need to retrieve the binary path at the exec stage and we need to transmit that information for later when the kprobe hits. As of now, we don't transmit the whole string of the binary from the execve stage to the kprobe matchBinaries but only an index because it's more efficient. This means that during the kprobe matchBinaries BPF code we only know yes or no if we matched but not really on what (since we don't have the string but only an ID).

[...]

PR #1408, opened 2023-08-30
Convert string and char_buf matches to hash look ups

kevsecurity:
Currently string, char buf, file, fd, and path matchers (equal, prefix, postfix, plus the 'not' equivalents for file, fd and path) are all written in BPF using loops to inspect (potentially) every character in the strings to be matched.

This series (see commits for details) converts all these matchers to hash map look ups, using a set of 6 hash maps of varying sizes for equal matches, and LPM TRIE maps for prefix and postfix look ups. In addition, the 'not' equivalents are implemented for string and char buf.

The default string lengths for the matches has been set at 128 bytes, but could be increased to an arbitrary size for equals, and up to 256 bytes for prefix and postfix. For small programs (<5.3), the postfix limit (which is bounded by a byte-by-byte reverse copy) has been raised from 40 chars to 96 chars. For large programs, the postfix limit is set at 128 chars (raised from 50 chars).

### DIFF
--- a/bpf/process/string_maps.h
+++ b/bpf/process/string_maps.h
@@ -171,7 +171,7 @@ struct {
 #ifdef __LARGE_BPF_PROG
 #define STRING_POSTFIX_MAX_MATCH_LENGTH STRING_POSTFIX_MAX_LENGTH
 #else
-#define STRING_POSTFIX_MAX_MATCH_LENGTH 40
+#define STRING_POSTFIX_MAX_MATCH_LENGTH 96
 #endif
 
 struct string_postfix_lpm_trie {
--- a/bpf/process/string_maps.h
+++ b/bpf/process/string_maps.h
@@ -166,4 +166,38 @@ struct {
 	__uint(value_size, sizeof(struct string_prefix_lpm_trie));
 } string_prefix_maps_heap SEC(".maps");
 
+#define STRING_POSTFIX_MAX_LENGTH 128
+#define STRING_POSTFIX_MAX_MASK	  (STRING_POSTFIX_MAX_LENGTH - 1)
+#ifdef __LARGE_BPF_PROG
+#define STRING_POSTFIX_MAX_MATCH_LENGTH STRING_POSTFIX_MAX_LENGTH
+#else
+#define STRING_POSTFIX_MAX_MATCH_LENGTH 40
+#endif
+
+struct string_postfix_lpm_trie {
+	__u32 prefixlen;
+	__u8 data[STRING_POSTFIX_MAX_LENGTH];
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, STRING_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_LPM_TRIE);
+			__uint(max_entries, 1);
+			__type(key, __u8[sizeof(struct string_postfix_lpm_trie)]); // Need to specify as byte array as wouldn't take struct as key type
+			__type(value, __u8);
+			__uint(map_flags, BPF_F_NO_PREALLOC);
+		});
+} string_postfix_maps SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(struct string_postfix_lpm_trie));
+} string_postfix_maps_heap SEC(".maps");
+
 #endif // STRING_MAPS_H__
--- a/bpf/process/string_maps.h
+++ b/bpf/process/string_maps.h
@@ -138,4 +138,32 @@ struct {
 	__uint(value_size, 512);
 } string_maps_ro_zero SEC(".maps");
 
+#define STRING_PREFIX_MAX_LENGTH 128
+
+struct string_prefix_lpm_trie {
+	__u32 prefixlen;
+	__u8 data[STRING_PREFIX_MAX_LENGTH];
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, STRING_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_LPM_TRIE);
+			__uint(max_entries, 1);
+			__type(key, __u8[sizeof(struct string_prefix_lpm_trie)]); // Need to specify as byte array as wouldn't take struct as key type
+			__type(value, __u8);
+			__uint(map_flags, BPF_F_NO_PREALLOC);
+		});
+} string_prefix_maps SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, sizeof(struct string_prefix_lpm_trie));
+} string_prefix_maps_heap SEC(".maps");
+
 #endif // STRING_MAPS_H__
--- a/bpf/process/string_maps.h
+++ b/bpf/process/string_maps.h
@@ -1,0 +1,141 @@
+// SPDX-License-Identifier: GPL-2.0
+/* Copyright Authors of Cilium */
+
+#ifndef STRING_MAPS_H__
+#define STRING_MAPS_H__
+
+#define STRING_MAPS_OUTER_MAX_ENTRIES 8
+
+/*
+ * To facilitate an arbitrary number of strings that can be matched on, string matching
+ * uses a hash look up. The problem with this is that the key to a hash has to be a fixed
+ * size, so if the maximum string length is 128 bytes, then all stored strings will be
+ * 128 bytes long (padded with 0s) and the string to be looked up also has to be padded
+ * with 0s to 128 bytes. This means that a short string will be hashed as if it is 128
+ * bytes long.
+ *
+ * The BPF hash maps use jhash for key hashing. See include/linux/jhash.h. This requires
+ * approximately 1 CPU cycle per byte, so in the example above, hashing every string,
+ * regardless of length, will take ~128 cycles, which is clearly inefficient. See
+ * https://fosdem.org/2023/schedule/event/bpf_hashing/ for details.
+ *
+ * jhash hashes in 12 byte blocks (3 x u32). For all lengths >12, a number of 12 byte
+ * blocks are hashed, and the remainder is hashed using a combination of single byte
+ * loads/shifts, followed by a final mix. It appears that the most efficient use of
+ * jhash is with lengths equal to 12k + 1, minimising the number of single byte loads/
+ * shifts.
+ *
+ * In order to reduce the amount of hashing of padded 0s, we opt to store string matches
+ * in multiple hashes, with increasing key sizes, where the key size is one more than a
+ * multiple of 12. Each string to be stored is placed in the hash that has the smallest
+ * key size that can accommodate it (and is padded to the key size). Strings to be looked
+ * up are equally padded to the smallest key size that can accommodate them, and then
+ * looked up in the related map.
+ *
+ * The chosen key sizes are 25, 49, 73, 97, 121, 145 (6 maps).
+ *
+ * In order to distinguish between character buffers that end in 0s and similar buffers
+ * that are padded with 0s, each string will be prefixed by its length stored in a
+ * single byte.
+ */
+#define STRING_MAPS_KEY_INC_SIZE 24
+#define STRING_MAPS_SIZE_0	 1 * STRING_MAPS_KEY_INC_SIZE + 1
+#define STRING_MAPS_SIZE_1	 2 * STRING_MAPS_KEY_INC_SIZE + 1
+#define STRING_MAPS_SIZE_2	 3 * STRING_MAPS_KEY_INC_SIZE + 1
+#define STRING_MAPS_SIZE_3	 4 * STRING_MAPS_KEY_INC_SIZE + 1
+#define STRING_MAPS_SIZE_4	 5 * STRING_MAPS_KEY_INC_SIZE + 1
+#define STRING_MAPS_SIZE_5	 6 * STRING_MAPS_KEY_INC_SIZE + 1
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, STRING_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, 1);
+			__type(key, __u8[STRING_MAPS_SIZE_0]);
+			__type(value, __u8);
+		});
+} string_maps_0 SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, STRING_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, 1);
+			__type(key, __u8[STRING_MAPS_SIZE_1]);
+			__type(value, __u8);
+		});
+} string_maps_1 SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, STRING_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, 1);
+			__type(key, __u8[STRING_MAPS_SIZE_2]);
+			__type(value, __u8);
+		});
+} string_maps_2 SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, STRING_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, 1);
+			__type(key, __u8[STRING_MAPS_SIZE_3]);
+			__type(value, __u8);
+		});
+} string_maps_3 SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, STRING_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, 1);
+			__type(key, __u8[STRING_MAPS_SIZE_4]);
+			__type(value, __u8);
+		});
+} string_maps_4 SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, STRING_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, 1);
+			__type(key, __u8[STRING_MAPS_SIZE_5]);
+			__type(value, __u8);
+		});
+} string_maps_5 SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, 512);
+} string_maps_heap SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__uint(key_size, sizeof(__u32));
+	__uint(value_size, 512);
+} string_maps_ro_zero SEC(".maps");
+
+#endif // STRING_MAPS_H__
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -867,15 +867,24 @@ static inline __attribute__((always_inline)) bool is_not_operator(__u32 op)
 static inline __attribute__((always_inline)) long
 filter_file_buf(struct file_buf *filter, struct string_buf *args)
 {
-	char *values = (char *)filter + sizeof(struct file_buf); // all values are after the header
-	__u32 next, remain = filter->total_sz - sizeof(__u32) - sizeof(__u32);
-	long i, match;
+	long match;
 
 	/* There are cases where file pointer may not contain a path.
 	 * An example is using an unnamed pipe. This is not a match.
 	 */
 	if (args->len == 0)
 		return 0;
+
+	switch (filter->op) {
+	case op_filter_eq:
+	case op_filter_neq:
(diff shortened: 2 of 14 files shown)


bpf/process/string_maps.h
```diff
@@ -0,0 +1,141 @@
...
+ * The BPF hash maps use jhash for key hashing. See include/linux/jhash.h. This requires
+ * approximately 1 CPU cycle per byte, so in the example above, hashing every string,
+ * regardless of length, will take ~128 cycles, which is clearly inefficient. See
+ * https://fosdem.org/2023/schedule/event/bpf_hashing/ for details.
+ *
+ * jhash hashes in 12 byte blocks (3 x u32). For all lengths >12, a number of 12 byte
+ * blocks are hashed, and the remainder is hashed using a combination of single byte
+ * loads/shifts, followed by a final mix. It appears that the most efficient use of
+ * jhash is with lengths equal to 12k + 1, minimising the number of single byte loads/
+ * shifts.
```
kkourt, line 26: Based in above rationale, wouldn't be isn't 12k+1 resulting in one single byte load? Wouldn't the most efficient use be 12k where the single byte loads are 0?

pkg/sensors/tracing/selectors.go
```diff
@@ -230,6 +266,362 @@ func populateAddr6FilterMap(
...
+}
+
+func populateStringFilterMap1(
+	pinPathPrefix string,
+	outerMap *ebpf.Map,
+	innerID uint32,
+	innerData map[[selectors.StringMapsSize1]byte]struct{},
+	maxEntries uint32,
+) error {
+	innerName := fmt.Sprintf("string_maps_1_%d", innerID)
```
kkourt, line 311: Do we need 6 copies of this function? Can we just use an additional argument and a single function?

pkg/selectors/selectors.go
```diff
@@ -61,6 +73,13 @@ type KernelSelectorState struct {
 	newBinVals    map[uint32]string              // these should be added in the names_map
 
 	listReader ValueReader
+	// stringMaps are used to populate string and char buf matches
+	stringMaps1 []map[[StringMapsSize1]byte]struct{}
+	stringMaps2 []map[[StringMapsSize2]byte]struct{}
+	stringMaps3 []map[[StringMapsSize3]byte]struct{}
+	stringMaps4 []map[[StringMapsSize4]byte]struct{}
+	stringMaps5 []map[[StringMapsSize5]byte]struct{}
+	stringMaps6 []map[[StringMapsSize6]byte]struct{}
```
kkourt, line 95: Do we need to reflect the exact structure of the bpf code? Can we just have string maps here and do the conversion when we load the bpf map? This means that we could also use a single map for everything, which I think would simplify the code.

bpf/process/types/basic.h
```diff
@@ -631,9 +632,101 @@ copy_char_buf(void *ctx, long off, unsigned long arg, int argm,
...
+	len = (__u8)orig_len;
+	// Calculate padded string length
+	padded_len = len;
+	if (len % STRING_MAPS_KEY_INC_SIZE != 0)
+		padded_len = ((len / STRING_MAPS_KEY_INC_SIZE) + 1) * STRING_MAPS_KEY_INC_SIZE;
+
+	// Check if we have entries for this padded length.
+	// Do this before we copy data for efficiency.
+	index = (padded_len / STRING_MAPS_KEY_INC_SIZE) - 1;
+	map_idx = map_ids[index & 0x7];
```
kkourt, line 659: What is the map_idx here? Is this how we distinguish between different matchArgs? It's not entirely clear to me how we do that.

kkourt (changes requested):
Hi,

Had a look at the first patch. The idea of using hash maps makes perfect sense to me, so thanks for taking this on.

One thing I'm not so sure about is whether it is worth the complexity of having multiple maps instead of one. As far as I understand the motivation here is to save space. Could you provide some examples of how much space we would save using this approach? I think it depends on the use-case but my feeling is that if we are talking about 10-20 string values, then the difference is not that big.

In any case, I would suggest that the keep the go structures simpler than what we do in bpf, and then deal with the exact structure of the bpf maps when we load the data.

pkg/selectors/selectors.go
```diff
@@ -236,3 +395,91 @@ func (k *KernelSelectorState) insertAddr6Map(addr6map map[KernelLpmTrie6]struct{
...
+		mapid = uint32(len(k.stringMaps1))
+		k.stringMaps1 = append(k.stringMaps1, stringMap1)
+	} else {
+		mapid = 0xffffffff
+	}
+	details[0] = mapid
+
+	if len(stringMap2) > 0 {
+		mapid = uint32(len(k.stringMaps2))
+		k.stringMaps2 = append(k.stringMaps2, stringMap2)
```
olsajiri, line 434: could please put in comment on how mapid (as size of the k.stringMapsX) correlates to the map_idx value logic in filter_char_buf_equal

pkg/selectors/kernel_test.go
```diff
@@ -225,13 +225,17 @@ func TestParseMatchArg(t *testing.T) {
 	expected1 := []byte{
 		0x01, 0x00, 0x00, 0x00, // Index == 1
 		0x03, 0x00, 0x00, 0x00, // operator == equal
-		18, 0x00, 0x00, 0x00, // length == 18
+		32, 0x00, 0x00, 0x00, // length == 32
```
olsajiri, line 228: would be great to have tests for the maximum of the matched strings

pkg/sensors/tracing/selectors.go
kevsecurity: I looked at this, but as the key to the golang maps and the BPF maps is different for each map, I came to the conclusion that the only way in golang would be through an interface and type reflection. On balance, I thought it was more understandable to enforce the type safety imposed by golang rather than bypass it with reflection. Keeping the golang type safety meant passing arguments of the right types, and all six maps are technically different types, even though they look similar. Happy to change if you can provide a construction that doesn't use reflection to ascertain the type, or a good rationale for why reflection is more understandable and sensible here.

pkg/selectors/selectors.go
kevsecurity: I thought a one-to-one mapping was easier to understand, and follows the style for (at least the majority, if not all) of the other maps.

bpf/process/string_maps.h
```diff
@@ -0,0 +1,202 @@
...
+#define STRING_MAPS_SIZE_6	 6 * STRING_MAPS_KEY_INC_SIZE + 1
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
+	__uint(max_entries, STRING_MAPS_OUTER_MAX_ENTRIES);
+	__uint(key_size, sizeof(__u32));
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, 1);
```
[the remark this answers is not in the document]
kevsecurity: The max_entries here is a placeholder (as per InMap construction). It eventually holds as many as required when it is statically populated when the maps are created.

examples/tracingpolicy/hardlink-observe.yaml
```diff
@@ -22,4 +22,4 @@ spec:
       - index: 1
         operator: "Equal"
         values:
-        - "/etc/passwd\0"
+        - "/etc/passwd"
```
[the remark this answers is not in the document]
kevsecurity: The \0 shouldn't really be specified for equal operators as the match is now over the whole string and not a potential substring of it. That said, maybe we should handle (chop off) a \0 in a string type in case people have these in their tracing policies.

pkg/selectors/selectors.go
kevsecurity: Yes I can do that.

pkg/selectors/kernel_test.go
kevsecurity: I'm not sure how we'd do that TBH.

olsajiri: perhaps some that will test the code with all the different string sizes and corner cases?

kevsecurity: Yep, I could make a load of tests for string matches with lengths at the boundaries.

kevsecurity:
> In any case, I would suggest that the keep the go structures simpler than what we do in bpf, and then deal with the exact structure of the bpf maps when we load the data.

The problem with keeping the go structures simpler is that we need to know the array indices of the hash maps at the point that we build the kernel selectors. The easiest way to know these is to make lists of hash maps that mirror those in BPF. That said, some level of harmonisation is possible, which I shall do.

pkg/selectors/selectors.go
kevsecurity: I've changed this to an array of maps as a good middleground.

bpf/process/types/basic.h
kevsecurity: I added a comment per @olsajiri's comment to make this more clear.

examples/tracingpolicy/hardlink-observe.yaml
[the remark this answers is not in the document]
kevsecurity: I've added code to silently remove trailing \0 characters so it now supports (with tests) strings with and without the trailing \0.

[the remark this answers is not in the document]
mtardy: we have multiple examples with `\0` as well in the documentation so that's cool that it works in any case.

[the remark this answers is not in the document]
kevsecurity: Suggest after this is merged that we remove the \0 characters. It doesn't look great and could be confusing.

kkourt:
> > > One thing I'm not so sure about is whether it is worth the complexity of having multiple maps instead of one. As far as I understand the motivation here is to save space.
> > 
> > , and it would be nice to justify the code complexity of having multiple maps with some numbers.
> > The motivation is to save CPU cycles in the look up code. If we only have one hash map with a key size of (say) 128 bytes, then every string we try to look up will require hashing 128 bytes, regardless if many of them are 0x00. Using several maps with different size keys means we only hash an amount of bytes that is closer to the string length we wish to look up.
> 
> During offline sync we discussed quickly why the different maps, and to me it is fine. Maybe it is complex from readability, but from the performance view even if we don't have numbers, but the fact that 1. these selectors where not working properly and now they are being fixed, and 2. file name and string matching is the most basic and easy use case. Users may watch different paths then having _better_ benchmark results on top is always good.

(1) and (2) make sense to me, but they could also be implemented with a single map for all sizes. Having different maps for different sizes is a performance optimization (for CPU cycles as you mention above), and, given the code complexity it introduces, I think it's reasonable to ask: "is it worth it?" How much time do we save compared to when we use a single map for all sizes.

kevsecurity:
> (1) and (2) make sense to me, but they could also be implemented with a single map for all sizes. Having different maps for different sizes is a performance optimization (for CPU cycles as you mention above), and, given the code complexity it introduces, I think it's reasonable to ask: is it worth it? How much time do we save compared to when we use a single map for all sizes?

Jhash cost is a little over a CPU cycle per length of data to hash in bytes, but we could say that is equal; i.e. hashing 128 bytes costs 128 CPU cycles (see https://fosdem.org/2023/schedule/event/bpf_hashing/ for details/confirmation). The existing byte-by-byte comparison has a cost of CPU cycles that similarly grows with the length of the data to compare.

While fixing (1) and (2) above, we also don't want to incur extra CPU cost if we can avoid it. In this situation, we want to give the users the ability to specify fairly long strings/filenames to match on – say ~128 bytes. If we take the simplistic approach of a single map, then all look ups will require hashing 128 bytes, with a cost of ~128 CPU cycles. By dividing the strings into different maps, we reduce this cost to approximately the size of the key. Therefore, assuming many look up strings are less than 24 byte in length, many look ups would cost 25 CPU cycles compared with 128 CPU cycles. We could assume that the next biggest category would be slightly longer strings, with these costing 49 CPU cycles compared to 128 CPU cycles.

Compared to these gains, the code isn't really that complex. Rather than putting all the strings in the same map, they are simply divided into six maps based on their length; and then at look up, the appropriate map is chosen based on the length of the look up string. An advantage of this approach is that we could increase the maximum string length to, say, 1024 bytes with no additional CPU cost – there may be a small memory cost, however.

Given that Tetragon hooks kernel functions and performs comparisons during many of the hooks, we always cost a system CPU time. Looking at Sysmon (Windows) as an example of a tool with millions of installs, many of the use cases feature string and filename matching. IMO it's not a case of "Is it worth it?" but more a case of "How could we justify not doing it?"

kevsecurity:
I've taken some measurements using 50,000,000 calls to a kprobe:
* The difference between no hash and the hash based on 6 maps is 60.2ns/kprobe.
* The difference between no hash and the hash based on a single map is 127.6ns/kprobe.
* The difference between the hash based on 6 maps and the hash based on a single map is 67.4ns/kprobe.
Objectively, the hash for a single map is more than twice the cost of the hash for 6 maps. This is admittedly very small compared to the total cost of the kprobe (2.92us), and represents an increase of 2.3%.

olsajiri:
I'm getting following error for spec below, which I assume should work now

`time="2023-09-06T13:17:54Z" level=fatal msg="Failed to start tetragon" error="sensor gkp-sensor-1 from collection sys-symlink-passwd failed to load: failed prog bpf/objs/bpf_multi_kprobe_v61.o kern_version 393730 loadInstance: failed to insert string_maps_0_8: update: key too big for map: argument list too long"
`

it seems there's double values iteration, first in writeMatchValues and then in writeMatchStrings,
but I still don't get fully the mapid/map_idx correlation.. so I might be off ;-)

```
kind: TracingPolicy
metadata:
  name: "sys-symlink-passwd"
spec:
  kprobes:
  - call: "sys_symlinkat"
    syscall: true
    args:
    - index: 0
      type: "string"
    - index: 1
      type: "int"
    - index: 2
      type: "string"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Equal"
        values:
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
        - "AAAAAAAAAAAA"
```

kevsecurity:
> I'm getting following error for spec below, which I assume should work now
> 
> `time="2023-09-06T13:17:54Z" level=fatal msg="Failed to start tetragon" error="sensor gkp-sensor-1 from collection sys-symlink-passwd failed to load: failed prog bpf/objs/bpf_multi_kprobe_v61.o kern_version 393730 loadInstance: failed to insert string_maps_0_8: update: key too big for map: argument list too long" `
> 
> it seems there's double values iteration, first in writeMatchValues and then in writeMatchStrings, but I still don't get fully the mapid/map_idx correlation.. so I might be off ;-)

That's odd. I'll take a look. Thanks for trying it out.

mtardy (commented):
So I think after slowly reviewing this I learned many things, thanks, that's a really cool use of `BPF_MAP_TYPE_LPM_TRIE`! I must admit the multiple maps + padding stuff makes everything a bit more complex and hard to read indeed. I wonder after [your benchmarks](https://github.com/cilium/tetragon/pull/1408#issuecomment-1708180508) if you want to remove that optimization or not.

Anyway, that would be helpful as well for https://github.com/cilium/tetragon/issues/1278, as of now we use sets as well for `In` and `NotIn` that we could replace with those already implemented operators. Paths will be similar to string or char_buf.

I made some annoying remarks because it was confusing/or just convention, feel free to ignore them, just sayin.

It should be good to do a pass on the documentation as well (after that PR) to remove the trailing `\0`, even if they are harmless (thanks to your fix), they are confusing:
```diff
-  - "/etc/passwd\0"
+  - "/etc/passwd"
```

kevsecurity:
> it seems there's double values iteration, first in writeMatchValues and then in writeMatchStrings, but I still don't get fully the mapid/map_idx correlation.. so I might be off ;-)

Yep that's what it was – I fixed it in the last push. When I simplified the code I moved writeMatchStrings into writeMatchValues as I thought it made more sense there, but this introduced the double iteration. All looks good now.


Regarding mapid/map_idx: consider just the smallest key'd map (key size is 25 bytes). A single kprobe could have more than one string Equal match section – maybe for different parameters or different actions. Therefore there could be 2 or more hash tables of strings, one for each match section. Each of these is placed inside an array_of_hash map at indices 0, 1, 2... These indices are then stored in the kernel selectors, so each match knows which hash table to use.

But of course there are six key sizes, so therefore six array_of_hash maps, and therefore the above process six times. In the selector we store the index for the relevant hash map in the first array_of_hash, then the index for the relevant hash map in the second array_of_hash, and so on for all six. In BPF, we take the input string, work out which key size it would fall into, get the index from the selector for that hash map, and look it up in the relevant array_of_hash. Then we can pad our string to that key size and look up the string in that hash map.

HTH

olsajiri (approved):
I'd like to see more tests given the size of new code (also for prefix/postfix),
but I guess that can be added later, anyway it looks good to me now

kevsecurity:
> I'd like to see more tests given the size of new code (also for prefix/postfix), but I guess that can be added later, anyway it looks good to me now

Thanks for the approval. My thought about tests was that this isn't new functionality, so while more tests might be a good idea, we were already okay with the testing of the existing implementation (which happened to have bugs that I fixed a little while back!).

mtardy:
> > I'd like to see more tests given the size of new code (also for prefix/postfix), but I guess that can be added later, anyway it looks good to me now
> 
> Thanks for the approval. My thought about tests was that this isn't new functionality, so while more tests might be a good idea, we were already okay with the testing of the existing implementation (which happened to have bugs that I fixed a little while back!).

yep understandable, it would be nice though to add some tests on the limits, like the one Jiri proposed https://github.com/cilium/tetragon/pull/1408#issuecomment-1708358266. But as he said, we can add them later.

kevsecurity:
> yep understandable, it would be nice though to add some tests on the limits, like the one Jiri proposed [#1408 (comment)](https://github.com/cilium/tetragon/pull/1408#issuecomment-1708358266). But as he said, we can add them later.

I added tests to the first commit for the 6 map string look up, all focused on the limits. Prefix and postfix could potentially benefit from tests at the max string length, but I don't feel they are urgent. The fact that tests already exist for them (from the previous implementation) satisfies me.

bpf/process/types/basic.h
```diff
@@ -631,9 +632,101 @@ copy_char_buf(void *ctx, long off, unsigned long arg, int argm,
...
+	int zero = 0;
+
+	if (orig_len >= STRING_MAPS_SIZE_5 || !orig_len)
+		return 0;
+
+	len = (__u8)orig_len;
+	// Calculate padded string length
+	padded_len = len;
+	if (len % STRING_MAPS_KEY_INC_SIZE != 0)
+		padded_len = ((len / STRING_MAPS_KEY_INC_SIZE) + 1) * STRING_MAPS_KEY_INC_SIZE;
```
kkourt, line 654: Instead of the branch and the divide, you could also do:
```
padded_len = len  + ((STRING_MAPS_KEY_INC_SIZE - (size % STRING_MAPS_KEY_INC_SIZE)) % STRING_MAPS_KEY_INC_SIZE)
```

Test:

```shell-interactive
$ python3
Python 3.11.5 (main, Aug 29 2023, 15:31:31) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> def align1(size, alignment):
...      return size + ((alignment - (size % alignment)) % alignment)
...
>>> def align2(size, alignment):
...      if size % alignment == 0:
...          return size
...      return ( (size  // alignment) + 1)  * alignment
...
>>> for i in range(10000):
...     assert align1(i, 24) == align2(i, 24)
...
>>> align1(24, 24)
24
>>> align1(25, 24)
48

```

bpf/process/types/basic.h
```diff
@@ -740,6 +740,49 @@ filter_char_buf_prefix(struct selector_arg_filter *filter, char *arg_str, uint a
 	return !!pass;
 }
 
+static inline __attribute__((always_inline)) void
+copy_reverse(__u8 *dest, uint len, __u8 *src)
+{
+	uint i;
+
+	len &= (STRING_POSTFIX_MAX_LENGTH - 1);
```
kkourt, line 748: nit: might be worth defining `STRING_POSTFIX_MAX_MASK = STRING_POSTFIX_MAX_LENGTH` since we seem to be using it a lot.

kkourt (changes requested):
Thanks! This looks good to me.

I have some suggestions, but I feel the PR is definitely an improvement from the current state of things, so I'm happy to merge it as is. The only thing I would like to clarify before merging is the numbers. First, thanks for measuring! The summary seems to be that this gives us a 2.3% improvement, which is small but an improvement nevertheless.

~I'm not sure I understand the numbers though.~

> I've taken some measurements using 50,000,000 calls to a kprobe:
>
>   The difference between no hash and the hash based on 6 maps is 60.2ns/kprobe.
>   The difference between no hash and the hash based on a single map is 127.6ns/kprobe.
>   The difference between the hash based on 6 maps and the hash based on a single map is 67.4ns/kprobe.

> Objectively, the hash for a single map is more than twice the cost of the hash for 6 maps. This is admittedly very small compared to the total cost of the kprobe (2.92us), and represents an increase of 2.3%.

For example:
> The difference between the hash based on 6 maps and the hash based on a single map is 67.4ns/kprobe.


So my reading of this is that we are saving 67.4ns for every invocation of the hook.
But after you say that:
> This is admittedly very small compared to the total cost of the kprobe (2.92us), and represents an increase of 2.3%.

~So I'm not sure where the 2.3% comes from. Am I missing something?~
I was:  67.4 / 2920 = 2.3%. 

Could you add the measurements (maybe the absolute values?)  for every run in the commit message so that it is available as part of git history?

Thanks!

pkg/selectors/selectors.go
```diff
@@ -209,6 +289,48 @@ func ArgSelectorValue(v string) ([]byte, uint32) {
 	return b, uint32(len(b))
 }
 
+func ArgStringSelectorValue(v string, removeNul bool) ([]byte, int, error) {
```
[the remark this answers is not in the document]
kevsecurity: Yep, this was left over from the version where this could return strings padded to different lengths.

pkg/selectors/selectors.go
```diff
@@ -209,6 +289,48 @@ func ArgSelectorValue(v string) ([]byte, uint32) {
...
+	}
+
+	// Pad returned string to max string
+	fullPadLen := MaxStringMapsSize - s
+	// Add real length to start and padding to end
+	sByte := []byte{byte(s)}
+	b = append(sByte, b...)
+	b = append(b, make([]byte, fullPadLen)...)
+	// Total length is padded string len + prefixed length byte.
+	return b, paddedLen + 1, nil
```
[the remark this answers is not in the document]
kevsecurity: Yep, this was left over from the version where this could return strings padded to different lengths.

bpf/process/types/basic.h
kevsecurity: I think golang will optimise accordingly, and I think the branch and divide is more readable TBH.

bpf/process/types/basic.h
```diff
@@ -740,6 +740,49 @@ filter_char_buf_prefix(struct selector_arg_filter *filter, char *arg_str, uint a
...
+#ifndef __LARGE_BPF_PROG
+#pragma unroll
+#endif
+	for (i = 0; i < (STRING_POSTFIX_MAX_MATCH_LENGTH - 1); i++) {
```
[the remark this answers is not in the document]
kevsecurity: No this is correct. It's related to the masking and indexing (to keep the verifier happy). The actual maximum is indeed 1 less than the stated max.

kkourt (approved):
Thanks! 🚀

PR #1417, opened 2023-09-01
check if procfs is the actual host procfs

kkourt:
See patches.

### DIFF
--- a/cmd/tetra/dump/dump.go
+++ b/cmd/tetra/dump/dump.go
@@ -9,9 +9,12 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/cilium/ebpf"
 	"github.com/cilium/tetragon/pkg/defaults"
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/policyfilter"
+	"github.com/cilium/tetragon/pkg/sensors/base"
+	"github.com/cilium/tetragon/pkg/sensors/exec/execvemap"
 	"github.com/spf13/cobra"
 )
 
@@ -24,8 +27,26 @@ func New() *cobra.Command {
 	}
 
 	ret.AddCommand(
+		execveMapCmd(),
 		policyfilterCmd(),
 	)
+
+	return ret
+}
+
+func execveMapCmd() *cobra.Command {
+	mapFname := filepath.Join(defaults.DefaultMapRoot, defaults.DefaultMapPrefix, base.ExecveMap.Name)
+	ret := &cobra.Command{
+		Use:   "execve",
+		Short: "dump execve map",
+		Args:  cobra.ExactArgs(0),
+		Run: func(cmd *cobra.Command, _ []string) {
+			dumpExecveMap(mapFname)
+		},
+	}
+
+	flags := ret.Flags()
+	flags.StringVar(&mapFname, "map-fname", mapFname, "execve map filename")
 
 	return ret
 }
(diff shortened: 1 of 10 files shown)


tixxdz:
Maybe compare /proc/1/ns/pid with https://elixir.bootlin.com/linux/latest/source/include/linux/proc_ns.h#L46 IIRC it is an idr allocation that has been constant for a very long time... and init pid is statically declared anyway

bpf/process/bpf_gethostpid.c
```diff
@@ -0,0 +1,61 @@
...
+
+	if (PT_REGS_PARM2_CORE_SYSCALL(ctx_) != f->off)
+		return 0;
+
+	if (PT_REGS_PARM3_CORE_SYSCALL(ctx_) != f->whence)
+		return 0;
+
+	ev.pid = get_current_pid_tgid() >> 32;
+	perf_event_output(ctx, &tg_hpid_event_map, BPF_F_CURRENT_CPU, &ev, sizeof(ev));
+	return 0;
```
olsajiri, line 60: seems like too much (using perf rb) for retrieving just one pid value.. could we just update map? or perhaps something like in `pkg/sensors/map_update.go ` maybe the ebpf program could return the pid as return value, but I might be too optimistic ;-)

kkourt: I used a perf rb to handle the case where we get >1 than one events. it's very unlikely to happen but it does not seem like a lot of trouble to handle it.

kkourt: > maybe the ebpf program could return the pid as return value, but I might be too optimistic ;-)

That's what I tried first using BPF_PROG_RUN, but I couldn't make it to work. Specifically, I could not find a program type that would allow me to get the pid. So, instead, I used a system call. Is there  a way to use `get_current_pid_tgid()` from a bpf program that we BPF_PROG_RUN?

pkg/bpf/hostpid/hostpid.go
```diff
@@ -0,0 +1,127 @@
...
+	lseekProg, ok := col.Programs["tg_hpid_hook"]
+	if !ok {
+		return 0, fmt.Errorf("gethostpid program not in collection")
+	}
+
+	zero := uint32(0)
+	filterVal := filter{
+		fd:     -1,
+		off:    4343,
+		whence: 5151,
```
kevsecurity, line 91: If you were to make these two values completely random, then you'd have a 1 in 2^64 chance of colliding with another lseek call that also set the FD to -1 (highly unlikely). In fact, you could set the FD to a random negative value too (maximum number of open files appears to be much less than max(int) so overflow into negative numbers appears unlikely) and increase this to 1 in 2^95 which is reaching cryptographic collision avoidance territory.
Doing that would mean you wouldn't need the ring buffer and a simple map would suffice.

kevsecurity (commented):
I tend to agree with not using the ring buffer and instead using a map to communicate the value to user space. Choosing random values for the magic numbers in the lseek call would avoid collisions with other lseek calls. Suggest this would simplify the code somewhat.

tpapagian (commented):
I believe that what Djalal said in a previous comment makes sense. i.e. to compare `/<option.Config.ProcFS>/1/ns/pid` with https://elixir.bootlin.com/linux/latest/source/include/linux/proc_ns.h#L46.

```
$ kind create cluster
[...]
$ docker ps
CONTAINER ID   IMAGE                  COMMAND                  CREATED          STATUS          PORTS                       NAMES
799c437d3f63   kindest/node:v1.27.3   "/usr/local/bin/entr…"   43 seconds ago   Up 36 seconds   127.0.0.1:41377->6443/tcp   kind-control-plane
$ docker exec -it kind-control-plane bash
root@kind-control-plane:/# ls -la /proc/1/ns/pid
lrwxrwxrwx 1 root root 0 Oct  3 09:10 /proc/1/ns/pid -> 'pid:[4026532359]'
root@kind-control-plane:/#
exit
$ sudo ls -la /proc/1/ns/pid
lrwxrwxrwx 1 root root 0 Oct  2 10:17 /proc/1/ns/pid -> 'pid:[4026531836]'
```

If we run on the host pid namespace the value of `/proc/1/ns/pid` should always be 4026531836 (0xEFFFFFFC) (or I miss something).

kkourt:
Pushed a new version using @tixxdz's suggestion. PTAL, thanks!

tixxdz (approved):
Thanks

PR #1363, opened 2023-08-16
api: Deprecate Pod.labels field

michi-covalent:
Deprecate Pod.labels field which contains Cilium identity labels, and document that it has been replaced by Pod.pod_labels field.

Ref: #794

PR #1420, opened 2023-09-02
Remove tetragon.enableCiliumAPI configuration option

michi-covalent:
Pod.Labels field got deprecated in v0.11 [^1], so we can stop setting the
field in main branch. This means that Tetragon no longer depends on
the Cilium state, so we don't have to synchronize it anymore. This patch
removes tetragon.enableCiliumAPI Helm value and deletes packages related
to Cilium state synchronization.

[^1]: https://github.com/cilium/tetragon/pull/1363


cmd/tetragon/flags.go
```diff
@@ -94,7 +94,6 @@ func readAndSetFlags() {
 
 	option.Config.EnableProcessCred = viper.GetBool(keyEnableProcessCred)
 	option.Config.EnableProcessNs = viper.GetBool(keyEnableProcessNs)
-	option.Config.EnableCilium = viper.GetBool(keyEnableCiliumAPI)
```
kkourt, line 97: Would it make sense to keep the option and only fail it is set to true? I _think_ this would enable us to run the `:latest` image with the old helm charts that define this option. We can remove it once we have released the next versin.

mtardy:
Hey, should we merge that (once it's rebased)? It's still in draft for now.

mtardy:
Should we close this one now that https://github.com/cilium/tetragon/pull/1541 was merged? maybe it was missing the:
> and deletes packages related to Cilium state synchronization.

michi-covalent:
yeah this one is a much bigger change since it actually deletes `pkd/oldhubble` and such. let's close it for now, i'll reopen it when we are ready to delete these cilium-related packages 💣

PR #1426, opened 2023-09-04
Remove cilium-bpf configuration parameter

michi-covalent:
There is a rumor that Tetragon actually doesn't need to load Cilium BPF programs.


tpapagian (approved):

michi-covalent:
requesting review from @jrfastab for extra confirmation

jrfastab:
this is a thing I did in the distant past where I looked into the the CIlium NAT tables and did NAT translations for folks so they could see wire traffic vs local traffic. I dropped the code at some point so its not so useful anymore.

PR #1429, opened 2023-09-05
Add kernel stack traces to kprobes events

mtardy:
See individual commits for info.

See [related documentation preview](https://deploy-preview-1429--tetragon.netlify.app/docs/concepts/tracing-policy/selectors/#stack-traces).

### Preview

Looks like this with the following tracing policy:
```yaml
kind: TracingPolicy
metadata:
  name: syscall
spec:
  kprobes:
    - call: kfree_skb_reason
      selectors:
      - matchActions:
        - action: post
          stackTrace: true
```
And doing
```shell-session
nc -vz spooky.com 443
```

With
```shell-session
./tetra getevents -o compact --processes nc
```

The output looks like this (with color):

image

```release-note
Add a new kernel stack traces alpha feature to kprobes events.
```


olsajiri:
> Looks like this with the following tracing policy:
> 
> ```yaml
> kind: TracingPolicy
> metadata:
>   name: syscall
> spec:
>   kprobes:
>     - call: kfree_skb_reason
>       selectors:
>       - matchActions:
>         - action: post
>           stackTrace: true
> ```

bit late to the discussion, so sorry if I missed something.. but seems like this kind of setup will trigger the filtering code even if it's not necessary, because the selector is empty and actions tail call does not need to be called

could we perhaps instead add flag/bool to config_map and call get_stackid in generic_output? with spec like:

```yaml
spec:
  kprobes:
    - call: kfree_skb_reason
      stackTrace: true
```

tixxdz:
> > Looks like this with the following tracing policy:
> > ```yaml
> > apiVersion: cilium.io/v1alpha1
> > kind: TracingPolicy
> > metadata:
> >   name: syscall
> > spec:
> >   kprobes:
> >     - call: kfree_skb_reason
> >       selectors:
> >       - matchActions:
> >         - action: post
> >           stackTrace: true
> > ```
> 
> bit late to the discussion, so sorry if I missed something.. but seems like this kind of setup will trigger the filtering code even if it's not necessary, because the selector is empty and actions tail call does not need to be called
> 
> could we perhaps instead add flag/bool to config_map and call get_stackid in generic_output? with spec like:
> 
> ```yaml
> spec:
>   kprobes:
>     - call: kfree_skb_reason
>       stackTrace: true
> ```

+1 for @olsajiri suggestion, as this should apply to all actions or triggered probes anyway

Another thing inside the doc please state clearly that this feature prints kernel addresses, so please ensure to use unix socket for the API when loading tracing policies....

The kernel has some protections to restrict unprivileged from printing kernel addresses... hmm we need to think about this

mtardy:
> Another thing inside the doc please state clearly that this feature prints kernel addresses, so please ensure to use unix socket for the API when loading tracing policies....
> 
> The kernel has some protections to restrict unprivileged from printing kernel addresses... hmm we need to think about this

This is in the doc, have you checked it? 😄 
image

mtardy:
About putting this thing out of selectors, indeed it makes sense. We had discussions offline about moving that thing from another action to a parameter to post with @kkourt and @tixxdz, but indeed moving it out of selectors makes sense. I think here we have a "UI" vs "how things work" clash a little bit.

pkg/grpc/tracing/tracing.go
```diff
@@ -239,13 +242,34 @@ func GetProcessKprobe(event *MsgGenericKprobeUnix) *tetragon.ProcessKprobe {
 		}
 	}
 
+	var stackTrace []*tetragon.StackTraceEntry
+	for _, addr := range event.StackTrace {
+		if addr != 0 {
+			kernelSymbols, err := ksyms.KernelSymbols()
```
kkourt, line 248: What does an `addr` of zero mean? Do we need to issue a warning?

nit/suggestion:

```
if addr == 0 {
    continue
}
/* ... rest of the code goes here */
```

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1080,6 +1112,37 @@ func handleGenericKprobe(r *bytes.Reader) ([]observer.Event, error) {
...
+				})
+				if err != nil {
+					logger.GetLogger().WithError(err).Warn("failed to load the stacktrace map")
+				}
+				// close this in cleanup postHook defer stackTraceMap.Close()
+			}
+
+			// this can't be an else statement in the previous block since it
+			// must execute as well when the reference is first initialized
+			if gk.stackTraceMapRef != nil {
```
kkourt, line 1136: But can it be that LoadPinnedMap returns a not-nil value, without the error?
In other words, do we need the `if` statement? Isn't it always true?

kkourt (approved):
Thanks, I have some minor comments but this LGTM.

kkourt:
> > Looks like this with the following tracing policy:
> > ```yaml
> > apiVersion: cilium.io/v1alpha1
> > kind: TracingPolicy
> > metadata:
> >   name: syscall
> > spec:
> >   kprobes:
> >     - call: kfree_skb_reason
> >       selectors:
> >       - matchActions:
> >         - action: post
> >           stackTrace: true
> > ```
> 
> bit late to the discussion, so sorry if I missed something.. but seems like this kind of setup will trigger the filtering code even if it's not necessary, because the selector is empty and actions tail call does not need to be called
> 
> could we perhaps instead add flag/bool to config_map and call get_stackid in generic_output? with spec like:
> 
> ```yaml
> spec:
>   kprobes:
>     - call: kfree_skb_reason
>       stackTrace: true
> ```

The downside of having it as a global option is that you don't get to have a different option for different selectors. I'd also expect the selector filtering code to be lightweight if there are nothing to match, but could be wrong on that. I'd prefer having it as a part of the selector. If performance is an issue, I think we can fix it by having a global option and pass it to bpf if all selectors have the stack trace enabled.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1080,6 +1112,37 @@ func handleGenericKprobe(r *bytes.Reader) ([]observer.Event, error) {
...
+				err = gk.stackTraceMapRef.Lookup(id, &unix.StackTrace)
+				if err != nil {
+					log.Fatal(err)
```
mtardy, line 1139: just noticed this by reading, need to fix this fatal

pkg/sensors/tracing/generickprobe.go
mtardy: with the actual code, we need to check since `err != nil` doesn't prevent us to continue to execute since we just log the error.

mtardy: with the actual code, we need to check since `err != nil` doesn't prevent us to continue to execute since we just log the error.

kkourt: I see. I'm fine with keeping as is. Another option might be to move the code into a separate function from which we can just `return` if there is an error to simplify the flow.

pkg/grpc/tracing/tracing.go
mtardy: It's mostly because the event here is from this struct and we store the stack trace into an array (not a slice) which is already to the size of MAX (127) so we might have a lot of zeros https://github.com/cilium/tetragon/pull/1429/files/086c82354e96092f49008c17a857259d2645c00a..9a5111467b295f2f4e88c7e4f573f0a8c14839fb#diff-09da9c4618bddc46b4f0bfefea390a0d75bf59158a6a0c6d09670e1cfef638a5R420

jrfastab:
That warning about the addrs I find a bit alarming. I don't actually think its as serious of an issue as the warning states however. How about just blinding the addresses for now by mask and hash?

pkg/grpc/tracing/tracing.go
mtardy: added a comment and use the if addr = 0 { continue } code structure.

mtardy:
> That warning about the addrs I find a bit alarming. I don't actually think its as serious of an issue as the warning states however. How about just blinding the addresses for now by mask and hash?

do we want to use the same hash as the kernel or we don't care? I think the kernel does something like:
```C
+/* Maps a pointer to a 32 bit unique identifier. */
+static char *ptr_to_id(char *buf, char *end, void *ptr, struct printf_spec spec)
+{
+	unsigned long hashval;
+	const int default_width = 2 * sizeof(void *);
+
+	if (unlikely(!have_filled_random_ptr_key)) {
+		spec.field_width = default_width;
+		/* string length must be less than default_width */
+		return string(buf, end, "(ptrval)", spec);
+	}
+
+#ifdef CONFIG_64BIT
+	hashval = (unsigned long)siphash_1u64((u64)ptr, &ptr_key);
+	/*
+	 * Mask off the first 32 bits, this makes explicit that we have
+	 * modified the address (and 32 bits is plenty for a unique ID).
+	 */
+	hashval = hashval & 0xffffffff;
+#else
+	hashval = (unsigned long)siphash_1u32((u32)ptr, &ptr_key);
+#endif
+
+	spec.flags |= SMALL;
+	if (spec.field_width == -1) {
+		spec.field_width = default_width;
+		spec.flags |= ZEROPAD;
+	}
+	spec.base = 16;
+
+	return number(buf, end, hashval, spec);
+}
```

jrfastab (approved):

mtardy:
I converted it to a draft because we need to fix something with the multi-kprobe support with @olsajiri first. But it's 99% ready! :)

Issue #1428, opened 2023-09-05
invalid memory address or nil pointer dereference


rgarcia89:
### What happened?

1. AKS
2. helm chart with the following values
```
USER-SUPPLIED VALUES:
nodeSelector:
  kubernetes.azure.com/mode: user
tetragon:
  exportAllowList: '{"event_set":["PROCESS_EXEC"]}'
  exportDenyList: |-
    {"health_check":true}
    {"namespace":["", "calico-system", "elastic-system", "kube-system", "kyverno", "monitoring", "tigera-operator"]}
  fieldFilters: '{"fields": "parent.cwd,parent.binary,parent.arguments,parent.flags,parent.pod.container,parent.pod.name,parent.pod.namespace",
    "action": "INCLUDE"}'
  grpc:
    enabled: false
  prometheus:
    address: ""
    enabled: false
    port: 2112
    serviceMonitor:
      enabled: true
      labelsOverride: {}
```

### Tetragon Version

v0.11.0

### Kernel Version

5.15.116.1-2.cm2

### Kubernetes Version

1.25.11

### Bugtool

/ # tetra bugtool
time="2023-09-05T09:01:00Z" level=info msg="saving init info"
time="2023-09-05T09:01:00Z" level=info msg="retrieving lib directory" libDir=/var/lib/tetragon/
time="2023-09-05T09:01:00Z" level=warning msg="not an object file, ignoring" path=/var/lib/tetragon/
time="2023-09-05T09:01:01Z" level=info msg="skipping metadata directory" path=/var/lib/tetragon/metadata
time="2023-09-05T09:01:01Z" level=warning msg="no btf filename in tetragon config, attempting to fall back to /sys/kernel/btf/vmlinux"
time="2023-09-05T09:01:02Z" level=info msg="btf file added" btfFname=/sys/kernel/btf/vmlinux
time="2023-09-05T09:01:02Z" level=info msg="tetragon log file added" exportFname=/var/run/cilium/tetragon/tetragon.log
time="2023-09-05T09:01:02Z" level=info msg="executed command" cmd=/bin/dmesg dstFname=dmesg.out ret=0
time="2023-09-05T09:01:02Z" level=info msg="executed command" cmd="/sbin/tc filter show dev lo ingress" dstFname=tc-info.lo.ingress ret=0
time="2023-09-05T09:01:02Z" level=info msg="executed command" cmd="/sbin/tc filter show dev lo egress" dstFname=tc-info.lo.egress ret=0
[...]

PR #1432, opened 2023-09-05
Use a message copy to apply fieldFilters in exec events

tpapagian:
As we keep them in the process cache, removing fields may affect the correctness of the process cache. In order to fix that, we create a copy of the event and apply fieldFilters only on the copy.

FIXES: https://github.com/cilium/tetragon/issues/1428


willfindlay (approved):
LGTM, thanks for catching this

PR #1437, opened 2023-09-05
observer_test_helper: make helpers generic over TB

willfindlay:
The testing package supports additional test types, such as fuzzing and bench testing. However, our existing observer test helpers were designed only to work with unit tests. To support additional test types, modify the observer test helpers to use the generic testing.TB interface instead of testing.T. This should facilitate writing fuzz tests and benchmarks for Tetragon in the future.


mtardy (approved):
looks very 😇

tpapagian (approved):

PR #1440, opened 2023-09-07
Modify Operator cluster role

prateek041:
this commit:
- changes the name of operator cluster role to match the name in cluster role binding.


michi-covalent (commented):
let's fix https://github.com/cilium/tetragon/blob/main/install/kubernetes/templates/operator_clusterrolebinding.yaml#L11 instead of changing the cluster role name.

the current cluster role name is consistent with tetragon daemonset cluste rrole:

- tetragon cluster role name: `tetragon`
- tetragon operator cluster role name: `tetragon-operator`

michi-covalent (approved):

Issue #1037, opened 2023-05-30
Make prometheus metric pod label optional


rayjanoka:
### Is there an existing issue for this?

No

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Our system has a lot of unique pod names starting and stopping all the time and we found that after a little time (a few days) the cardinality is very high for the `tetragon_events_total` metric due to the `pod` label. I'm seeing 300k unique labels on a single tetragon with a 3-5s scrape time across 70 nodes and crashing prometheus.

```bash
➜ curl localhost:2112/metrics | awk -F\{ '{print$1}' | sort | uniq -c
...
305435 tetragon_events_total
```

### Describe the feature you would like

If the pod label for the `tetragon_events_total` metric was optionally removed that would greatly reduce the cardinality. 

I tried to filter the label out with `droplabel` in the prometheus serviceMonitor `metricRelabelings` but this results in multiple entries with the same set of labels and the values cannot be re-aggregated (added) back together at scrape time.

### Describe your proposed solution

We may be able to set a flag to enable this and send an empty pod name string [here](https://github.com/cilium/tetragon/blob/main/pkg/metrics/eventmetrics/eventmetrics.go#L84), but not sure. I could put something together if it makes sense.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

lambdanis:
@rayjanoka Thanks for creating this issue. Having pod as a label is definitely problematic in many envs.

> We may be able to set a flag to enable this and send an empty pod name string [here](https://github.com/cilium/tetragon/blob/main/pkg/metrics/eventmetrics/eventmetrics.go#L84), but not sure. I could put something together if it makes sense.

Yes, a flag to control the metrics cardinality is a way to go IMO.

What would be nice to have in the future is an option to configure different granularity, e.g. Kubernetes workload (deployment/daemonset/whatev) or `app` label. So I think something like `metrics-granularity` flag with (initial) options `"pod"` or empty would be a good solution.

Feel free to put something together :)

PR #1444, opened 2023-09-10
metrics: Add metrics label filter configuration

nap32:
Currently, metrics are all-or-nothing.
Certain labels may cause cardinality issues.

This patch introduces a new configuration option - MetricsLabelFilter. It is an allow-list for configuring namespace, workload, pod, and binary. Labels that utilize these fields will only add them if configured for it.

Fixes: #1037

### DIFF
--- a/cmd/tetragon/flags.go
+++ b/cmd/tetragon/flags.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/metrics/config"
 	"github.com/cilium/tetragon/pkg/option"
 
 	"github.com/spf13/viper"
@@ -31,13 +32,14 @@ const (
 	keyEnableCiliumAPI        = "enable-cilium-api"
 	keyEnableProcessAncestors = "enable-process-ancestors"
 
-	keyMetricsServer     = "metrics-server"
-	keyServerAddress     = "server-address"
-	keyGopsAddr          = "gops-address"
-	keyEnableProcessCred = "enable-process-cred"
-	keyEnableProcessNs   = "enable-process-ns"
-	keyTracingPolicy     = "tracing-policy"
-	keyTracingPolicyDir  = "tracing-policy-dir"
+	keyMetricsServer      = "metrics-server"
+	keyMetricsLabelFilter = "metrics-label-filter"
+	keyServerAddress      = "server-address"
+	keyGopsAddr           = "gops-address"
+	keyEnableProcessCred  = "enable-process-cred"
+	keyEnableProcessNs    = "enable-process-ns"
+	keyTracingPolicy      = "tracing-policy"
+	keyTracingPolicyDir   = "tracing-policy-dir"
 
 	keyCpuProfile = "cpuprofile"
 	keyMemProfile = "memprofile"
@@ -113,6 +115,7 @@ func readAndSetFlags() {
 	option.Config.DataCacheSize = viper.GetInt(keyDataCacheSize)
 
 	option.Config.MetricsServer = viper.GetString(keyMetricsServer)
+	option.Config.MetricsLabelFilter = config.ParseMetricsLabelFilter(viper.GetString(keyMetricsLabelFilter))
 	option.Config.ServerAddress = viper.GetString(keyServerAddress)
 
 	option.Config.ExportFilename = viper.GetString(keyExportFilename)
--- a/install/kubernetes/templates/tetragon_configmap.yaml
+++ b/install/kubernetes/templates/tetragon_configmap.yaml
@@ -37,6 +37,9 @@ data:
 {{- else }}
   metrics-server: ""
 {{- end }}
+{{- if .Values.tetragon.prometheus.enabled }}
+  metrics-label-filter: {{ .Values.tetragon.prometheus.metricsLabelFilter }}
+{{- end }}
 {{- if .Values.tetragon.grpc.enabled }}
   server-address: {{ .Values.tetragon.grpc.address }}
 {{- else }}
--- a/pkg/metrics/consts/consts.go
+++ b/pkg/metrics/consts/consts.go
@@ -4,3 +4,4 @@
 package consts
 
 var MetricsNamespace = "tetragon"
+var KnownMetricLabelFilters = []string{"namespace", "workload", "pod", "binary"}
--- a/pkg/metrics/eventmetrics/eventmetrics.go
+++ b/pkg/metrics/eventmetrics/eventmetrics.go
@@ -20,12 +20,12 @@ import (
 )
 
 var (
-	EventsProcessed = metrics.NewCounterVecWithPod(prometheus.CounterOpts{
+	EventsProcessed = metrics.MustNewGranularCounter(prometheus.CounterOpts{
 		Namespace:   consts.MetricsNamespace,
 		Name:        "events_total",
 		Help:        "The total number of Tetragon events",
 		ConstLabels: nil,
-	}, []string{"type", "namespace", "workload", "pod", "binary"})
+	}, []string{"type"})
 	FlagCount = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Namespace:   consts.MetricsNamespace,
 		Name:        "flags_total",
@@ -39,19 +39,19 @@ var (
 		ConstLabels: nil,
 	})
 
-	policyStats = metrics.NewCounterVecWithPod(prometheus.CounterOpts{
+	policyStats = metrics.MustNewGranularCounter(prometheus.CounterOpts{
 		Namespace:   consts.MetricsNamespace,
 		Name:        "policy_events_total",
 		Help:        "Policy events calls observed.",
 		ConstLabels: nil,
-	}, []string{"policy", "hook", "namespace", "workload", "pod", "binary"})
+	}, []string{"policy", "hook"})
 )
 
 func InitMetrics(registry *prometheus.Registry) {
-	registry.MustRegister(EventsProcessed)
+	registry.MustRegister(EventsProcessed.ToProm())
 	registry.MustRegister(FlagCount)
 	registry.MustRegister(NotifyOverflowedEvents)
-	registry.MustRegister(policyStats)
+	registry.MustRegister(policyStats.ToProm())
 }
 
 func GetProcessInfo(process *tetragon.Process) (binary, pod, workload, namespace string) {
@@ -93,10 +93,10 @@ func handleProcessedEvent(pInfo *tracingpolicy.PolicyInfo, processedEvent interf
 	default:
 		eventType = "unknown"
 	}
-	EventsProcessed.WithLabelValues(eventType, namespace, workload, pod, binary).Inc()
+	EventsProcessed.ToProm().WithLabelValues(metrics.FilterMetricLabels(eventType, namespace, workload, pod, binary)...).Inc()
 	if pInfo != nil && pInfo.Name != "" {
-		policyStats.
-			WithLabelValues(pInfo.Name, pInfo.Hook, namespace, workload, pod, binary).
+		policyStats.ToProm().
+			WithLabelValues(metrics.FilterMetricLabels(pInfo.Name, pInfo.Hook, namespace, workload, pod, binary)...).
 			Inc()
 	}
 }
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -4,11 +4,16 @@
 package metrics
 
 import (
+	"fmt"
 	"net/http"
 	"sync"
 	"time"
 
+	"golang.org/x/exp/slices"
+
 	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/metrics/consts"
+	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/podhooks"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -26,6 +31,33 @@ var (
 	podQueueOnce        sync.Once
 	deleteDelay         = 1 * time.Minute
 )
+
+type GranularCounter struct {
+	counter     *prometheus.CounterVec
+	CounterOpts prometheus.CounterOpts
+	labels      []string
+	register    sync.Once
+}
+
+func MustNewGranularCounter(opts prometheus.CounterOpts, labels []string) *GranularCounter {
+	for _, label := range labels {
+		if slices.Contains(consts.KnownMetricLabelFilters, label) {
+			panic(fmt.Sprintf("labels passed to GranularCounter can't contain any of the following: %v. These labels are added by Tetragon.", consts.KnownMetricLabelFilters))
+		}
+	}
+	return &GranularCounter{
+		CounterOpts: opts,
+		labels:      append(labels, consts.KnownMetricLabelFilters...),
+	}
+}
+
+func (m *GranularCounter) ToProm() *prometheus.CounterVec {
+	m.register.Do(func() {
+		m.labels = FilterMetricLabels(m.labels...)
+		m.counter = NewCounterVecWithPod(m.CounterOpts, m.labels)
+	})
+	return m.counter
+}
 
 // NewCounterVecWithPod is a wrapper around prometheus.NewCounterVec that also registers the metric
 // to be cleaned up when a pod is deleted. It should be used only to register metrics that have
@@ -141,4 +173,21 @@ func EnableMetrics(address string) {
 	logger.GetLogger().WithField("addr", address).Info("Starting metrics server")
 	http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
 	http.ListenAndServe(address, nil)
+}
+
+// The FilterMetricLabels func takes in string arguments and returns a slice of those strings omitting the labels it is not configured for.
+// IMPORTANT! The filtered metric labels must be passed last and in the exact order of consts.KnownMetricLabelFilters.
+func FilterMetricLabels(labels ...string) []string {
+	offset := len(labels) - len(consts.KnownMetricLabelFilters)
+	if offset < 0 {
+		logger.GetLogger().WithField("labels", labels).Debug("Not enough labels provided to metrics.FilterMetricLabels.")
+		return labels
+	}
+	result := labels[:offset]
+	for i, label := range consts.KnownMetricLabelFilters {
+		if _, ok := option.Config.MetricsLabelFilter[label]; ok {
+			result = append(result, labels[offset+i])
+		}
+	}
+	return result
 }
--- a/pkg/option/config.go
+++ b/pkg/option/config.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/metrics/consts"
 	"github.com/spf13/viper"
 )
 
@@ -46,10 +47,11 @@ type config struct {
 	ProcessCacheSize int
 	DataCacheSize    int
 
-	MetricsServer    string
-	ServerAddress    string
-	TracingPolicy    string
-	TracingPolicyDir string
+	MetricsServer      string
+	MetricsLabelFilter map[string]interface{}
+	ServerAddress      string
+	TracingPolicy      string
+	TracingPolicyDir   string
 
 	ExportFilename             string
 	ExportFileMaxSizeMB        int
@@ -93,6 +95,15 @@ var (
 
 		// LogOpts contains logger parameters
 		LogOpts: make(map[string]string),
+
+		// Default to logging metrics with the greatest granularity.
+		MetricsLabelFilter: func() map[string]interface{} {
+			result := make(map[string]interface{})
+			for _, label := range consts.KnownMetricLabelFilters {
+				result[label] = nil
+			}
+			return result
+		}(),
 	}
 )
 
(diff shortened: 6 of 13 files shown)


pkg/metrics/metrics.go
```diff
@@ -142,3 +145,19 @@ func EnableMetrics(address string) {
...
+        // Uh-oh.
+        return labels
+    }
+    result := labels[:offset]
+    for i, label := range consts.KnownMetricLabelFilters {
+        if _, ok := option.Config.MetricsLabelFilter[label]; ok {
+            result = append(result, labels[i+offset])
+        }
+    }
+	return result
```
nap32, line 162: I'll fix this spacing.

pkg/metrics/metrics.go
```diff
@@ -18,13 +20,14 @@ import (
...
-	podQueue            workqueue.DelayingInterface
-	podQueueOnce        sync.Once
-	deleteDelay         = 1 * time.Minute
+	registry            	*prometheus.Registry
+	registryOnce        	sync.Once
+	metricsWithPod      	[]*prometheus.MetricVec
+	metricsWithPodMutex 	sync.RWMutex
+	podQueue            	workqueue.DelayingInterface
+	podQueueOnce        	sync.Once
+	deleteDelay         	= 1 * time.Minute
```
nap32, line 29: I'll restore this spacing -- at one point I had introduced a variable here but later removed it.

pkg/metrics/metrics.go
nap32: Ran go fmt.

pkg/metrics/metrics.go
nap32: Ran go fmt.

jrfastab:
This looks reasonable to me. @lambdanis any opinions?

pkg/metrics/metrics.go
```diff
@@ -142,3 +144,19 @@ func EnableMetrics(address string) {
 	http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
 	http.ListenAndServe(address, nil)
 }
+
+// The filtered metric labels must be passed last and in the exact order of consts.KnownMetricLabelFilters.
+func FilterMetricLabels(labels ...string) []string {
+	offset := len(labels) - len(consts.KnownMetricLabelFilters)
```
lambdanis, line 150: I think it would be a bit easier for developers if `FilterMetricLabels` took only values corresponding to `KnownMetricLabelFilters` as arguments instead of doing this offsetting.

lambdanis (commented):
Thanks for tackling this @nap32! The overall approach looks good to me. I left a few comments - mostly considering maintainability, I think we could make this feature a bit easier to use (when adding metrics) and extend.

Without analysing all possibilities in detail it's unclear to me what happens when a user passes a label that's not in `KnownMetricLabelFilters` or passes labels in different order. Ideally, we would have some validation for such cases to prevent unexpected behaviour. But if it's a big effort, then having at least test cases covering them would be good, so that the behaviour is somehow documented.

Regarding tests, it looks like the global config change of `MetricsLabelFilter` leaks into other tests breaking them?

pkg/metrics/eventmetrics/eventmetrics.go
```diff
@@ -25,7 +25,7 @@ var (
 		Name:        "events_total",
 		Help:        "The total number of Tetragon events",
 		ConstLabels: nil,
-	}, []string{"type", "namespace", "workload", "pod", "binary"})
+	}, metrics.FilterMetricLabels("type", "namespace", "workload", "pod", "binary"))
```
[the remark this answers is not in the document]
nap32: I had originally scoped metrics.FilterMetricLabels to only be used for filtering the globally configured labels.
The result was a statement that seems to fight the language:
`append([]string{"type"}, metrics.FilterMetricLabels(const.KnownMetricLabelFilters...)...`

We can't pass const.KnownMetricLabelFilters to the function because "type" is not a configuration label.
How do we want to handle this?

install/kubernetes/templates/tetragon_configmap.yaml
```diff
@@ -37,6 +37,10 @@ data:
 {{- else }}
   metrics-server: ""
 {{- end }}
+{{- if .Values.tetragon.prometheus.enabled }}
+  metrics-label-filter: {{.Values.tetragon.prometheus.metricsLabelFilter }}
```
lambdanis, line 41: Helm fails to render this, because the value has to be a string, not an array. You can convert the array to comma-separated strings:
```
{{.Values.tetragon.prometheus.metricsLabelFilter | join "," }}
```

pkg/metrics/eventmetrics/eventmetrics.go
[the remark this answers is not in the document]
lambdanis: I see. Can we use the global config here? Something like:
```
append([]string{"type"}, option.Config.MetricsLabelFilter...}
```
The config is set before the metrics are registered.

pkg/metrics/metrics.go
nap32: Happy to revise it in this direction, just some points for consideration -

If we were to change it to only accept and only filter the global configuration variables,
every time a developer would look to use the function,
they'd have to handle appending slices for the metric definition and incrementing the metric.

The way this is implemented, as long as your label filters are in order and come last you can just wrap your values in this function and the rest is handled (for both the metric definition and the statements incrementing the metric).

Do we still want to revise and drop the offset and move the slice operations outside of the function?
As an alternative, would it help if we expanded on the original comment instead highlight the expected usage?

lambdanis: I see your point. I think this solution is good too, but `append` shouldn't be needed when updating metric, it would be something like
```
WithLabelValues("some-type", metrics.FilterMetricLabels(namespace, workload, pod, binary)...).Inc()
```
Am I missing something here?

nap32:
> Thanks for tackling this @nap32! The overall approach looks good to me. I left a few comments - mostly considering maintainability, I think we could make this feature a bit easier to use (when adding metrics) and extend.

Awesome!
I had responses to two of the comments you left for feedback,
I just want to make sure we're aligned and aware of the trade-off for developer usage.
I originally had implemented the function as your suggestion describes,
and if we want to go that route I can change it back no problem!

> Without analysing all possibilities in detail it's unclear to me what happens when a user passes a label that's not in `KnownMetricLabelFilters` or passes labels in different order. Ideally, we would have some validation for such cases to prevent unexpected behaviour. But if it's a big effort, then having at least test cases covering them would be good, so that the behaviour is somehow documented.

The filtering function's major constraint is that the last strings passed need to correspond to the labels listed in consts.KnownMetricLabelFilters and remain in the order that the slice lists them in. I believe this is an unavoidable constraint.

As long as your last N arguments are the values listed in consts.KnownMetricLabelFilters, you're good to go.
If you pass the last N values out-of-order, they will be filtered incorrectly as if in the order they appear in the consts slice.

- This would be the same as if you passed the values out-of-order to the Prometheus `.WithLabelValues(string...)`.

- This could mean that a value that should be filtered out is left in and listed under a different metric label, or that a value that should be left in is filtered out by mistake and would cause issues with correctness.

- If you pass more than N values (where N is the `len(consts.KnownMetricLabelFilters)`), the first preceding values are returned in the result in an effort to avoid requiring slice operations.

It is not possible to check for this ordering, as the function is used for both defining metrics and incrementing them;
there would be no reasonable way to validate that a "namespace" label value isn't a "pod" label value by mistake.
Do you agree?
 
I'll add another comment clarifying the intent of the function and try to highlight the importance of the function's contract / constraints. It is important to clarify this, as even if we remove the offset and require the developer to manage slice operations the expectation for ordering will remain. 
 
> Regarding tests, it looks like the global config change of `MetricsLabelFilter` leaks into other tests breaking them?

I've corrected this and the other issues you've raised, thanks for pointing those out!

pkg/metrics/eventmetrics/eventmetrics.go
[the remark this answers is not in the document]
nap32: We can't use the config here because the config is a map;
but we can use the consts.KnownMetricLabelFilters as you've suggested if you'd prefer that.
By using consts.KnownMetricLabelFilters, we do not have to assume labels in the config are in a specific order.

pkg/option/config.go
```diff
@@ -93,6 +95,15 @@ var (
...
 		// LogOpts contains logger parameters
 		LogOpts: make(map[string]string),
+
+		// Default to logging metrics with the greatest granularity.
+		MetricsLabelFilter: func() map[string]interface{} {
+            result := make(map[string]interface{})
+            for _, label := range consts.KnownMetricLabelFilters {
+                result[label] = nil
+            }
+            return result
```
nap32, line 105: I will go fmt, sorry for the spacing issues.

pkg/metrics/metrics.go
nap32: When I was working through the implementation, I was expecting that we could handle it that way too.
I ran into some errors; I just double-checked and had the same issues.

With this approach:
```
EventsProcessed.WithLabelValues(eventType, metrics.FilterMetricLabels(namespace, workload, pod, binary)...).Inc()
```

Golang does not like it - even wrapping parenthesis around it does not fix it.
```
# github.com/cilium/tetragon/pkg/metrics/eventmetrics
pkg/metrics/eventmetrics/eventmetrics.go:96:48: too many arguments in call to EventsProcessed.WithLabelValues
        have (string, []string)
        want (...string)
make: *** [Makefile:163: tetragon] Error 1
```

I think that's how I arrived at the `append( )` approach before spending some time revising it.
I think the current usage as a wrapping function call makes it easier to read the code and use the function,
but that is just an opinion and of course I'm open to being wrong and/or going in another direction (like `append()`)!

install/kubernetes/templates/tetragon_configmap.yaml
nap32: Ran `helm template .` locally and verified the helm output included the expected values.
Thanks for the tip!

pkg/metrics/eventmetrics/eventmetrics.go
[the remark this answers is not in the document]
lambdanis: Ok, let's use `consts.KnownMetricLabelFilters` then, we can live with an append.

lambdanis:
Thank you for the updates and responses @nap32. I think we should reuse `consts.KnownMetricLabelFilters` in metrics definitions and then we'll be good to go.

Regarding edge cases - one thing I was considering is a user passing in Helm values unexpected labels, for example `["type", "namespace", "ip", "pod"]`. IIUC, this is a valid config, and will result in "namespace" and "binary" being included in metrics, and "workload" and "pod" excluded. The additional "type" and "ip" labels passed won't change anything, right?

nap32:
> Thank you for the updates and responses @nap32. I think we should reuse consts.KnownMetricLabelFilters in metrics definitions and then we'll be good to go.

Got it! I'll introduce the change shortly.

> Regarding edge cases - one thing I was considering is a user passing in Helm values unexpected labels, for example ["type", "namespace", "ip", "pod"]. IIUC, this is a valid config, and will result in "namespace" and "binary" being included in metrics, and "workload" and "pod" excluded. The additional "type" and "ip" labels passed won't change anything, right?

I think there is a typo in the example above, but the expectation is correct.
If `[ "namespace", "ip", "type", "binary" ]` is passed as a configuration:
- `"namespace"` and `"binary"` would be included in the metrics.
- `"ip"` and `"type"` would be ignored, since they aren't in the `consts.KnownMetricFilterLabels`.
- `"workload"` and `"pod"` would not be included in the metrics.

lambdanis:
> I think there is a typo in the example above, but the expectation is correct.

Yes, thanks!

nap32:
I ran the suggested command from the lint helm chart job (`cd install/kubernetes && ./test.sh`) and the tests passed.
Is there something else I should be doing with this to validate the test locally?

Do you have any suggestions on how I might troubleshoot the e2e tests?
Is there any documentation on running these locally w/ my own kind cluster, or can I copy-paste the command?
I'm sure I've made a mistake somewhere, but it is unclear to me why I've caused these networking errors:
- https://github.com/cilium/tetragon/actions/runs/6177694851/job/16886826129?pr=1444#step:7:101
- https://github.com/cilium/tetragon/actions/runs/6177694851/job/16886825841?pr=1444#step:7:101

pkg/metrics/eventmetrics/eventmetrics.go
```diff
@@ -25,7 +25,7 @@ var (
 		Name:        "events_total",
 		Help:        "The total number of Tetragon events",
 		ConstLabels: nil,
-	}, []string{"type", "namespace", "workload", "pod", "binary"})
+	}, metrics.FilterMetricLabels(append([]string{"type"}, consts.KnownMetricLabelFilters...)...))
```
lambdanis, line 28: `metrics.FilterMetricLabels` doesn't filter correctly in metrics definitions, because these definitions are evaluated before `option.Config.MetricsLabelFilter` is set.

To support configuring labels dynamically, we need to create Prometheus metrics after the config is evaluated, not as global variables. This can be done by defining an internal metric type wrapping the Prometheus metric type. The internal type would be used in global metrics definitions, but the Prometheus type would be initialized later, when the metric is registered.

I suggest defining such type in `pkg/metrics`:
```
// GranularCounter wraps *prometheus.CounterVec. It's intended to be used to
// define metrics with labels set at register time.
type GranularCounter struct {
	counter *prometheus.CounterVec
	prometheus.CounterOpts
	labels   []string
	register sync.Once
}

// MustNewGranularCounter creates a new GranularCounter. It adds labels defined in
// const.KnownMetricLabelFilters to the labels passed in an argument. The function panics if the
// labels passed contain any of the labels from const.KnownMetricLabelFilters.
func MustNewGranularCounter(opts prometheus.CounterOpts, labels []string) *GranularCounter {
	for _, label := range labels {
		if slices.Contains(consts.KnownMetricLabelFilters, label) {
			panic(fmt.Sprintf("labels passed to GranularCounter can't contain any of the following: %v. These labels are added by Tetragon.", consts.KnownMetricLabelFilters))
		}
	}
	return &GranularCounter{
		CounterOpts: opts,
		labels:      append(labels, consts.KnownMetricLabelFilters...),
	}
}

// ToProm converts GranularCounter to prometheus.CounterVec. The first time it's called, it
// evaluates the labels and creates the underlying value.
func (m *GranularCounter) ToProm() *prometheus.CounterVec {
	m.register.Do(func() {
		m.labels = FilterMetricLabels(m.labels...)
		m.counter = NewCounterVecWithPod(m.CounterOpts, m.labels)
	})
	return m.counter
}
```
the we can define this metric as:
```
EventsProcessed = metrics.MustNewGranularCounter(prometheus.CounterOpts{
    Namespace:   consts.MetricsNamespace,
    Name:        "events_total",
    Help:        "The total number of Tetragon events",
    ConstLabels: nil,
}, []string{"type"})
```
and convert it to the Prometheus type whenever it's used:
```
EventsProcessed.ToProm()
```

Let me know if there are any questions/concerns about this approach.

lambdanis (changes requested):
@nap32 I usually run Tetragon in a local kind cluster (see [docs](https://tetragon.cilium.io/docs/contribution-guide/development-setup/#running-tetragon-in-kind)). Then using kubectl you can check why the pod is crashing, read the logs, etc.

I've run your PR locally, there are two main issues that need fixing: reading the config value and defining/registering metrics. I suggested how to approach them in comments. Let me know if something is unclear.

Thanks a lot for working on that! Sorry for the delayed review, I hope it will go smoothly now.

pkg/metrics/metrics.go
```diff
@@ -142,3 +144,20 @@ func EnableMetrics(address string) {
 	http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
 	http.ListenAndServe(address, nil)
 }
+
+// The FilterMetricLabels func takes in string arguments and returns a slice of those strings omitting the labels it is not configured for.
+// IMPORTANT! The filtered metric labels must be passed last and in the exact order of consts.KnownMetricLabelFilters.
+func FilterMetricLabels(labels ...string) []string {
+	offset := len(labels) - len(consts.KnownMetricLabelFilters)
+	if offset < 0 {
+		// Uh-oh.
```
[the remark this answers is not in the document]
nap32: Absolutely, good call!

pkg/metrics/consts/consts.go
```diff
@@ -4,3 +4,4 @@
 package consts
 
 var MetricsNamespace = "tetragon"
+var KnownMetricLabelFilters = []string{"namespace", "workload", "pod", "binary"}
```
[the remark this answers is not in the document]
lambdanis: ...you can't, I keep forgetting that :) Sorry.

cmd/tetragon/flags.go
```diff
@@ -114,6 +115,7 @@ func readAndSetFlags() {
 	option.Config.DataCacheSize = viper.GetInt(keyDataCacheSize)
 
 	option.Config.MetricsServer = viper.GetString(keyMetricsServer)
+	option.Config.MetricsLabelFilter = viper.GetStringMap(keyMetricsLabelFilter)
```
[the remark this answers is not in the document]
nap32: Instead of a helper function, I think instead we would be able to introduce changes to the template:
```
{{- if .Values.tetragon.prometheus.enabled }}
  metrics-label-filter: -|
    {{- range .Values.tetragon.prometheus.metricsLabelFilter }}
    - {{ . | quote }}
    {{- end }}
{{- end }}
```
This allows us to parse correctly with `viper.GetStringMap(...)`.
Do you have any concerns with this?

pkg/metrics/eventmetrics/eventmetrics.go
nap32: I'll work through implementing this!
Thanks for the suggestion and help troubleshooting!

cmd/tetragon/flags.go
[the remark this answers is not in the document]
lambdanis: @nap32 Will it work with an empty list? Also, how would you pass the list as a command-line flag? Tetragon supports bare metal installations, so this option should be usable outside of the Helm chart too. As long as these two cases are good, changing the format is fine I think.

nap32:
Working on testing the metric changes end-to-end -- just sharing how you might stand up a local Prometheus server to test. Would it be helpful to add this to the documentation as a separate change/PR?

1. Run the local tetragon build in kind as described here: https://tetragon.cilium.io/docs/contribution-guide/development-setup/#running-tetragon-in-kind
2. Grab the ClusterIP for the metrics endpoint: `kubectl get services -A`
3. Set the ClusterIP and metrics port in the prometheus-config.yaml provided below.
4. You can't rely on a NodePort because kind's cluster is running in a container. Instead, `kubectl port-forward service/prometheus 9090:9090`.
5. Open the Prometheus UI in your host's browser to query metrics and review labels: `http://localhost:9090/`

prometheus-service.yaml
```
apiVersion: v1
kind: Service
metadata:
  name: prometheus
spec:
  type: NodePort
  ports:
    - port: 9090
      targetPort: 9090
  selector:
    app: prometheus
```

prometheus-deployment.yaml
```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: prometheus
spec:
  replicas: 1
  selector:
    matchLabels:
      app: prometheus
  template:
    metadata:
      labels:
        app: prometheus
    spec:
      containers:
        - name: prometheus
          image: prom/prometheus:latest
          args:
            - "--config.file=/etc/prometheus/prometheus.yml"
          ports:
            - containerPort: 8080
          volumeMounts:
            - name: prometheus-config-vol
              mountPath: /etc/prometheus/
      volumes:
        - name: prometheus-config-vol
          configMap:
            name: prometheus-config
```

prometheus-config.yaml
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: prometheus-config
data:
  prometheus.yml: |
    global:
      scrape_interval: 15s
    scrape_configs:
      - job_name: 'prometheus'
        static_configs:
          - targets: ['<TETRAGON CLUSTER IP>:2112']
```

cmd/tetragon/flags.go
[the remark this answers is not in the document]
nap32: I've tested and don't believe it would work with command line.
I've switched it to your suggestion, thanks!

nap32:
I'm not seeing the labels attached to metrics like I'd expect to,
I still need to work on this PR.

lambdanis:
@nap32 I see there are conflicts with the main branch, can you rebase your PR?

It looks good to me now. I tested your changes locally and I saw the labels I expected.

For a simple local test you can also access the metrics endpoint directly, without running Prometheus. Port-forward the service:
```
kubectl -n kube-system port-forward svc/tetragon 2112:2112
```
and then you can read the metrics in the text format at `localhost:2112/metrics`.

nap32:
> @nap32 I see there are conflicts with the main branch, can you rebase your PR?

Yes!

> It looks good to me now. I tested your changes locally and I saw the labels I expected.
> 
> For a simple local test you can also access the metrics endpoint directly, without running Prometheus. Port-forward the service:
> 
> ```
> kubectl -n kube-system port-forward svc/tetragon 2112:2112
> ```
> 
> and then you can read the metrics in the text format at `localhost:2112/metrics`.

Wow, this is a simple and straightforward way to test -- thanks.
I was over-complicating it. :+1:

michi-covalent (approved):

pkg/metrics/metrics.go
```diff
@@ -4,11 +4,15 @@
 package metrics
 
 import (
+	"fmt"
+	"golang.org/x/exp/slices"
 	"net/http"
 	"sync"
 	"time"
```
lambdanis, line 11:
```suggestion
	"net/http"
	"sync"
	"time"

    "golang.org/x/exp/slices"
```
The lint message:
```
File is not `goimports`-ed (goimports)
```
is not very clear, but what it means is that the imports are not in the expected order. `"golang.org/x/exp/slices"` should come after the standard library and before the other imports.

lambdanis (commented):
There is a single lint, but other than that looks good!

pkg/metrics/metrics.go
```diff
@@ -4,11 +4,16 @@
 package metrics
 
 import (
+	"fmt"
 	"net/http"
 	"sync"
 	"time"
 
+    "golang.org/x/exp/slices"
```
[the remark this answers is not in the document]
nap32: Sorry about that - I must have missed `go fmt`. :grimacing:
I'll use `make format` and `make check` instead moving forward.
Thanks!

lambdanis (approved):
Thanks for contributing this! :bowing_woman:

PR #1449, opened 2023-09-11
CODEOWNERS: multiple fixes and move file in `.github`

mtardy:
- Fix paths, indeed `path/` means that it will select `/path/` but also `/something/path/`. So I explicitely mentioned `/path/` for the entries.
- Exclude myself from some auto-generated docs files that are often modified.
- Add @cilium/tetragon owner to `/.github/workflows/` to ping William on workflow changes without blocking PRs.
- Move file into `/.github` since it's tied to GitHub and will reduce the amount of files in the root.

cc @willfindlay, tell me if you are satisfied with the changed I did to your line.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -2,14 +2,17 @@
 * @cilium/tetragon
 
 # Request a review from William when codegen is edited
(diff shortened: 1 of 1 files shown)


willfindlay (approved):
Changes look good to me

PR #1458, opened 2023-09-12
Add IPv6 support to BPF rate limit

kevsecurity:
See commits.

First commit fixes a couple of (pseudo?) bugs in the in-BPF rate limit code. It worked before, but it wasn't optimal.
Second commit adds support for IPv6 tuples so that rate limiting isn't just limited to IPv4.

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1767,10 +1767,6 @@ do_action_signal(int signal)
 /* The number of bytes per argument to include in the key
  * that we use to check for repeating data.
  * 16 is good for IPv4 data. Should increase for IPv6.
- * Be aware that we use (KEY_BYTES_PER_ARG - 1) as a bit
- * mask to limit indicies to 0 - (KEY_BYTES_PER_ARG - 1)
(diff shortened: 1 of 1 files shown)


bpf/process/types/basic.h
```diff
@@ -1789,24 +1789,35 @@ struct {
...
 } ratelimit_heap SEC(".maps");
 
+// This is zeroed memory that we NEVER write to, and use to copy over reusable heap in order
+// to zero it.
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, __u32);
+	__type(value, __u8[sizeof(struct ratelimit_key) + 128]);
+} ratelimit_ro_heap SEC(".maps");
```
jrfastab, line 1808: would be nice to preface with tg_ratelimit_ro_heap and above so that bpftool reports them with tetragon 'tg' map preface. Not sure we have hit all the maps yet with the naming convention but will be nice going forward.

jrfastab (approved):

PR #1459, opened 2023-09-12
Refactor files under operator directory

michi-covalent:
- Move cmdref to a separate package. This is a general-purpose package that's not tied to the operator logic.
- Move ConfigPopulate() to the option package.
- Move the command initialization logic to a separate cmd package.


mtardy (commented):
Sorry to add stuff here but wouldn't it be an opportunity to move `/operator` in `/cmd/operator`?

lambdanis:
> Sorry to add stuff here but wouldn't it be an opportunity to move /operator in /cmd/operator?

What's the reasoning behind it? Operator is a separate application, and having root-level `/operator` is consistent with Cilium. (Bur neither of these is a strong argument TBH).

lambdanis (approved):

michi-covalent:
feel free to do it in a separate pr 🚀 i kind of like the way it's currently structured, everything related to operator under `/operator` directory 🧘

PR #1467, opened 2023-09-13
Add support for file identity matching

cpaplham:
This commit adds SameFile and NotSameFile as operators for the file and path argument types. These operators match based on file identity, determined via inode number and device identifier. This allows files to be tracked even when accessed via a hard link or separate mount point.

Initial file identity is determined in userspace at Tetragon startup; specifying a nonexistent file will make the tracing policy invalid, and Tetragon will fail to launch.

NOTE: This is intended for monitoring of system files in a bare-metal or bare-VM environment, and unlikely to be very useful in Docker deployments.


pkg/selectors/kernel.go
```diff
@@ -265,6 +270,8 @@ var selectorOpStringTable = map[uint32]string{
 	SelectorOpNotPostfix:   "NotPostfix",
 	SelectorOpFamily:       "Family",
 	SelectorOpState:        "State",
+	SelectorOpSameFile:     "SameFile",
+	SelectorOpNotSameFile:  "SameFile",
```
olsajiri, line 274: "NotSameFile" ?

bpf/process/bpf_process_event.h
```diff
@@ -371,14 +376,16 @@ getcwd(struct msg_process *curr, __u32 offset, __u32 proc_pid)
...
+	__u32 dev = 0;
 
 	probe_read(&fs, sizeof(fs), _(&task->fs));
 	if (!fs) {
 		curr->flags |= EVENT_ERROR_CWD;
 		return 0;
 	}
 
-	buffer = d_path_local(_(&fs->pwd), &size, &flags);
+	buffer = d_path_local(_(&fs->pwd), &size, &flags, &ino, &dev);
```
olsajiri, line 388: just pass null, null in here and check ino/dev != null in path_with_deleted,
no need for extra assignment if we have no use for it

bpf/process/types/basic.h
```diff
@@ -1698,12 +1731,13 @@ installfd(struct msg_generic_kprobe *e, int fd, int name, bool follow)
 			     : [nameoff] "+r"(nameoff)
 			     :);
 
-		size = *(__u32 *)&e->args[nameoff];
+        // String size is 12 bytes after start of buffer
```
olsajiri, line 1734: wrong indentation

bpf/process/types/basic.h
```diff
@@ -879,37 +887,62 @@ filter_char_buf(struct selector_arg_filter *filter, char *args, int value_off)
 struct string_buf {
 	__u32 len;
 	char buf[];
-};
+} __attribute__((packed));
```
olsajiri, line 890: why not to put it under one struct file_buf, there's no other usage for string_buf AFAICS

olsajiri (commented):
please separate it into more logical changes and  add tests for this, otherwise looks good

bpf/process/types/basic.h
cpaplham: I incorrectly figured that string_buf might end up getting used for the more general string cases in the future.

cpaplham:
I've made the changes requested above and added some testing. Please let me know if anything else is required; I wasn't sure what you meant by "separate it into more logical changes".

cpaplham:
Ah, I see there's linter issues and that it's failing on 4.19. I'll fix the former and investigate the latter.

api/v1/README.md
```diff
@@ -458,6 +458,8 @@ https://github.com/opencontainers/runtime-spec/blob/main/config.md#createcontain
 | mount | [string](#string) |  |  |
 | path | [string](#string) |  |  |
 | flags | [string](#string) |  |  |
+| inode | [uint64](#uint64) |  |  |
```
tixxdz, line 461: Thank you for this feature, so I'm thinking if the inode needs more context to be really useful in other kprobe use cases, not only this one. The dev as an integer doesn't make that much sense to users, something like /proc/self/mountinfo is probably good. We have some context about the file itself, maybe we need same about the inode and its backing store, so users can grock it.

How about:
```
"inode": {
      "number": "1",
      "fs": {
            "name": "ext4",
            "dev": "1:1"
       }
}
```

So from this it should be something like:
```
message FileSystem {
    string name = 1;   // name of the filesystem
    string dev = 2;
}

message Inode {
    uint64 number = 1;
    FileSystem fs = 2;
}
```

The filesystem allows us to augment with the super block and fscontext easily in future... we can put this into its own file.proto maybe? @tpapagian 

Let's see what the others think about it

tixxdz (commented):
Have one comment for now about the external API, let's please see what others think otherwise thank you for the contribution ;-)

api/v1/README.md
tpapagian: I believe that having an inode number and device ID is enough for everything. We can introduce other fields like file system etc. but this can also happen as part of a next PR.

What I would suggest is that having a device number as it is now makes things a bit more difficult to interpret for users. Generally, the device is in the form of `<major:minor>`. Changing `uint32` to `string` and printing something like `"1:1"` would be great. Another approach would be to have separate fields (i.e. `device_major` and `device_minor` both as `uint32`) but I would say that I prefer the `string` approach. 

We can also have separate messages for that like what Djalal suggested. i.e.:
```
message FileDetails {
  uint64 number = 1;
  string device = 2;
};
```
and reuse that in `KprobePath` and `KprobeFile`. And we can extend that later if we need to do so.

PR #1470, opened 2023-09-14
lvh: allow renovate to update kernel images instead of using main

mtardy:
Adapt the regex used in the workflows to work with both use cases.

Along I migrate the renovate config because it was needed.

### DIFF
--- a/.github/workflows/vmtests.yml
+++ b/.github/workflows/vmtests.yml
@@ -65,11 +65,16 @@ jobs:
         fail-fast: false
         matrix:
            kernel:
-              - 'bpf-next-main'
-              - '5.15-main'
-              - '5.10-main'
-              - '5.4-main'
-              - '4.19-main'
+              # renovate: datasource=docker depName=quay.io/lvh-images/kernel-images
+              - 'bpf-next-20230912.113936'
+              # renovate: datasource=docker depName=quay.io/lvh-images/kernel-images
+              - '5.15-20230912.113936'
+              # renovate: datasource=docker depName=quay.io/lvh-images/kernel-images
+              - '5.10-20230912.113936'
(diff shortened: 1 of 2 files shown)


.github/workflows/vmtests.yml
```diff
@@ -65,11 +65,16 @@ jobs:
...
+              # renovate: datasource=docker depName=quay.io/lvh-images/kind
+              - 'bpf-next-20230912.113936'
+              # renovate: datasource=docker depName=quay.io/lvh-images/kind
+              - '5.15-20230912.113936'
+              # renovate: datasource=docker depName=quay.io/lvh-images/kind
+              - '5.10-20230912.113936'
+              # renovate: datasource=docker depName=quay.io/lvh-images/kind
+              - '5.4-20230912.113936'
+              # renovate: datasource=docker depName=quay.io/lvh-images/kind
+              - '4.19-20230912.113936'
```
kkourt, line 77: We are  different than cilium (for the moment) in that we do not use the kind image. I think we should change this and use the kind image everywhere, but until we do, this should be quay.io/lvh-images/kernel-images instead of /kind.

kkourt (changes requested):

.github/workflows/vmtests.yml
kkourt: see: https://github.com/cilium/tetragon/blob/main/tests/vmtests/fetch-data.sh for more details

mtardy: so I changed it, in the end there is no difference since they have the same update cadence and tag, and anyway the workflow is pulling the right thing, this comment is just for renovate. But let's be more correct.

kkourt (approved):

PR #1474, opened 2023-09-15
renovate: rename correctly the lvh image name in config

mtardy:
In the last patch, I changed the image name in the workflow comments but forgot to reflect that change in the config.


kkourt (approved):

mtardy:
Let's fast merged, remaining checks are useless.

PR #1479, opened 2023-09-15
k8s: Generate PodInfo client

michi-covalent:
I forgot to add "// +genclient" for PodInfo.


mtardy (approved):
I dont know this genclient but I'm sure it's crazy good

PR #1484, opened 2023-09-18
tetragon: build arm64 tarball

tixxdz:


willfindlay (approved):
Approving but needs a release note label

willfindlay (approved):

PR #1492, opened 2023-09-18
operator: Call crd.RegisterCRDs() from operator deployment

michi-covalent:
- Refactor the CRD registration logic to a separate common package so that both the root and serve commands can call it.
- Mount tetragon-operator-config configmap for the operator deployment.


michi-covalent:
the ultimate goal is to get rid of the init container from the tetragon daemonset, and make the tetragon operator solely responsible for registering CRDs. we need to teach tetragon to wait for CRDs to get registered before we can get rid of the init container.

jrfastab (approved):

PR #1493, opened 2023-09-19
podinfo: Populate workload info

michi-covalent:
- Add WorkloadType and WorkloadObject fields to PodInfo.
- Export workload.GetWorkloadMetaFromPod() function so that the operator
  can call it to set WorkloadType and WorkloadObject fields.
- Update equal() function to take these fields into account.

I opted for defining a new WorkloadObjectMeta type instead of using
metav1.ObjectMeta to avoid generating unnecessary "unknown field" log
messages. See [1] for additional context.

[1]: https://github.com/kubernetes-sigs/controller-tools/issues/448


pkg/process/workload.go
```diff
@@ -11,22 +11,26 @@ import (
...
 	// try to capture more useful namespace/name info for deployments, etc.
 	// TODO(dougreid): expand to enable lookup of OWNERs recursively a la kubernetesenv
-	deployMeta := pod.ObjectMeta
-	deployMeta.ManagedFields = nil
-	deployMeta.OwnerReferences = nil
+	deployMeta := v1alpha1.WorkloadObjectMeta{
+		Name:        pod.GetObjectMeta().GetName(),
+		Namespace:   pod.GetObjectMeta().GetNamespace(),
+		Labels:      pod.GetObjectMeta().GetLabels(),
+		Annotations: pod.GetObjectMeta().GetAnnotations(),
```
lambdanis, line 32: hmm, so Labels and Annotations are set to the pod metadata here, and kinda pretending to be the workload metadata if I understand correctly. It's actually no different from the current behaviour, but I'm wondering if these fields are needed at all?

lambdanis (commented):
looks good, I have one question.

pkg/process/workload.go
michi-covalent: yeah let's get rid of them, i don't need them. we can always add them back later if we want it 🥰

lambdanis (approved):
:shipit:

PR #1494, opened 2023-09-19
bpf: add DEBUG() macro

willfindlay:
The bpf_printk() helper is sometimes useful for debugging but we don't want to include it in production code since it makes programs incredibly slow. However, it's nice to have a consistent set of debugging messages we can lean on when it's time to investigate an issue on the BPF side.

This commit introduces a new DEBUG macro that is a NOP unless we set DEBUG=1 in the BPF Makefile, in which case it turns into a bpf_printk() call. This will enable us to over time build up and check in a suite of debugging statements for diagnosing problems during the development cycle without impeding performance in production contexts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,9 +51,11 @@ ifeq ($(TARGET_ARCH),arm64)
 endif
 BPF_TARGET_ARCH ?= x86
 
+__BPF_DEBUG_FLAGS :=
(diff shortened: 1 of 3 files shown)


kevsecurity (approved):
LGTM. Maybe we can include this in our unit tests?

olsajiri (approved):

PR #1510, opened 2023-09-26
[WIP] Define map metrics as custom collectors

lambdanis:
This PR changes the way how map metrics are collected. Instead of polling BPF maps at an interval, the metrics are collected on demand when they are scraped. The map metrics that were previously polled are now bundled into a custom `prometheus.Collector`.

It works, but the developer interface is quite error prone at the moment, so I'm making it a draft now and will try to rework it a little.


kevsecurity (approved):
LGTM

Issue #1318, opened 2023-08-04
Add a validation test to test load examples in `examples/tracingpolicy`


mtardy:
We recently received many reports that some of the TracingPolicy examples don't load because of validation failure. To avoid that, we could have a test that loads all the YAMLs in this folder (recursively) to see at least if they pass validation.

We should also make sure that validation in standalone Tetragon is the same as the validation when loading the TracingPolicy into the Kubernetes API.

PR #1521, opened 2023-09-28
Add complete k8s object validation and defaults on standalone

mtardy:
See individual commits.

This PR actually uncovered a few issues/bugs in type declaration and TracingPolicy tests. We will now have the same custom resource validation and defaults when deploying on k8s using CRD mechanisms and in standalone, this should unify the behavior.

Code addition is huge because we needed to make a binary an independent Go module because of a version mismatch in a library and I vendored the dependencies.

This might break user workflow if they relied on the fact that the validation was not properly done on standalone.


mtardy:
I linked this PR to close https://github.com/cilium/tetragon/issues/1318 as well because trying to decode the content of `examples/tracingpolicy` is already done here:

https://github.com/cilium/tetragon/blob/940528167ecba5466616a14af2f51114d7713f50/pkg/tracingpolicy/generictracingpolicy_test.go#L456-L488

But with this PR it will actually now pass a proper validation step while decoding `FromYAML`.

mtardy:
Followup would be:
- Improve the state of PreValidate for kprobes and add something more standard that we can share for advanced validation between kprobes/tracepoints/uprobes with specific validation for each of them, outputting format similar to what k8s validation does.
- Add advanced pre validation in smoke test

kevsecurity (approved):
If it's all green, then I think it's good to go.

PR #1525, opened 2023-09-29
tetragon: Several observer related cleanups

olsajiri:
Several observer related cleanups.

### DIFF
--- a/pkg/observer/observertesthelper/observer_test_helper.go
+++ b/pkg/observer/observertesthelper/observer_test_helper.go
@@ -255,7 +255,7 @@ func getDefaultObserverSensors(tb testing.TB, ctx context.Context, base *sensors
 		ret = append(ret, cnfSensor)
 	}
 
-	if err := loadObserver(tb, base, cnfSensor); err != nil {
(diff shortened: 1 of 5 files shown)


pkg/observer/observertesthelper/observer_test_helper.go
```diff
@@ -324,13 +324,57 @@ func GetDefaultObserverWithFile(tb testing.TB, ctx context.Context, file, lib st
...
+		tp, err = tracingpolicy.PolicyFromYAMLFilename(o.observer.config)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse tracingpolicy: %w", err)
+		}
+	}
+
+	var sensor *sensors.Sensor
+
+	if tp != nil {
+		sensor, err = sensors.GetMergedSensorFromParserPolicy(tp)
```
kkourt, line 356: One minor note: I'd like to remove the `GetMergedSensorsFromParserPolicy` and use everywhere the same code path: parse the policy, get a list of sensors, and we load the sensors. No need to change it now, but something to avoid in the fugure.

kkourt (approved):
Thanks!

PR #1530, opened 2023-09-29
Dockerfile.clang: Don't pin clang patch version

michi-covalent:
I boldly assert that it's sufficient to specify the clang version by
using clang-15 package. This way we can pick up the latest security
release automatically. Otherwise this Dockerfile keeps breaking when
there is a new security release and the old package gets deleted.

Ref: https://github.com/cilium/tetragon/actions/runs/6355442579/job/17263533977#step:6:239
Ref: https://packages.ubuntu.com/jammy/clang-15

### DIFF
--- a/Dockerfile.clang
+++ b/Dockerfile.clang
@@ -1,9 +1,8 @@
 FROM docker.io/library/ubuntu:22.04@sha256:0bced47fffa3361afa981854fcabcd4577cd43cebbb808cea2b1f33a3dd7f508
-ARG VERSION=1:15.0.7-0ubuntu0.22.04.2
-RUN apt-get update && apt-get install -y --no-install-recommends                 \
-    clang-15=$VERSION libclang-common-15-dev=$VERSION libclang-cpp15=$VERSION    \
-    libllvm15=$VERSION llvm-15-linker-tools=$VERSION libclang1-15=$VERSION       \
-    llvm-15=$VERSION llvm-15-runtime=$VERSION llvm-15-linker-tools=$VERSION make \
+RUN apt-get update && apt-get install -y --no-install-recommends  \
+    clang-15 libclang-common-15-dev libclang-cpp15    	          \
+    libllvm15 llvm-15-linker-tools libclang1-15                   \
+    llvm-15 llvm-15-runtime llvm-15-linker-tools make             \
     && apt-get clean && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


michi-covalent:
✅ success: https://github.com/cilium/tetragon/actions/runs/6355641458/job/17264006923?pr=1530

chancez:
This is probably because the base image is `ubuntu:22.04` and a recent update went from 22.04.2 to 22.04.3 is my guess. Is there a reason the apt install is specifying specific versions? You're basically getting the latest of the package anyways, and have to use the correct version of the package for the base image's Ubuntu patch release version anyways.

michi-covalent:
yeah not sure why we are pinning the version. @jrfastab might know 🤔

willfindlay:
Our BPF programs are very sensitive to compiler version. That's why we're trying to pin it.

michi-covalent:
> Our BPF programs are very sensitive to compiler version. That's why we're trying to pin it.

i see, but what can we do if the package keeps disappearing 💭

willfindlay:
@michi-covalent I'm not super familiar with apt but if there's a way to pin it to 1.15 without caring about patch version that would be ideal

michi-covalent:
yeah it's already pinned, they have different packages for claing, like `clang-15`, so it won't randomly get bumped up to `clang-16` for example 🥰 let me update this PR.

willfindlay:
```
ERROR: failed to solve: docker.io/library/golang:1.21.1@sha256:19600fdcae402165dcdab18cb9649540bde6be7274dedb5d205b2f84029fe909: failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/library/golang/manifests/sha256:13a7dc7b46068caeaf811cc8af56d394a89c97d9b194550f8be5a4de05ca5f99: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

:jarno-eclipse:

michi-covalent:
> :jarno-eclipse:

yeah dockerhub rate limits actions runners that are not hosted by github. opened https://github.com/cilium/tetragon/pull/1532 to make things slightly better.

jrfastab:
ok but wish we were using clang-16

jrfastab:
I'm ok to try this

jrfastab (approved):
ACK but if it gives us trouble we can revert.

PR #1531, opened 2023-09-29
Misc changes to the packages structure under pkg/metrics

lambdanis:
This PR contains a bunch of developer-facing breaking changes. See commits for details. I figured that it's worth doing them, as they will make future development easier and Tetragon is still before 1.0 release. But feel free to call me out if I shouldn't break package structure so freely :)

Initially I wanted also to merge the `pkg/metrics/consts` package into `pkg/metrics`, but I'm hitting import cycles. So I left this idea for now, I might revisit it when iterating on #1548 (although it doesn't seem that changes introduced there make it any simpler).

```release-note
Reorganize code under pkg/metrics. There are no functional changes, but it's a breaking change for developers.
```

### DIFF
--- a/cmd/tetragon/flags.go
+++ b/cmd/tetragon/flags.go
@@ -5,7 +5,7 @@ package main
 
 import (
 	"github.com/cilium/tetragon/pkg/logger"
-	"github.com/cilium/tetragon/pkg/metrics/config"
+	"github.com/cilium/tetragon/pkg/metrics/metricsconfig"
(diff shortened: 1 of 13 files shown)


lambdanis:
I'm not sure if I should label it as "breaking-changes" or not, but there are zero user-visible changes, so I opted for "misc".

michi-covalent (approved):
sure why not

jrfastab:
ACK go ahead and merge when ready.

PR #1532, opened 2023-09-29
packages-e2e-tests: Don't fail fast

michi-covalent:
Docker Hub keeps rate limiting buildjet runners. Set fail-fast to false so that ubuntu-22.04 runners can finish the job, and then we can re-trigger the failed buildjet job later.

Ref: https://github.com/cilium/tetragon/actions/runs/6355863531/job/17264805755?pr=1530
Ref: https://github.com/actions/runner-images/issues/1445#issuecomment-714832439

### DIFF
--- a/.github/workflows/packages-e2e-tests.yaml
+++ b/.github/workflows/packages-e2e-tests.yaml
@@ -10,6 +10,7 @@ jobs:
   standalone-tests:
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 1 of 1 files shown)


michi-covalent:
✅ as intended => https://github.com/cilium/tetragon/actions/runs/6356014280/attempts/1?pr=1532

jrfastab (approved):

PR #1541, opened 2023-10-02
Remove tetragon.enableCiliumAPI option

michi-covalent:
Tetragon no longer depends on Cilium.


mtardy (approved):
sweet 🍬

mtardy:
maybe we can add a release note to say that the flag is removed

PR #1544, opened 2023-10-03
helm: Make operator imagePullPolicy configurable

lambdanis:
imagePullPolicy for the operator deployment can be now set in tetragonOperator.image.pullPolicy value, and defaults to IfNotPresent. Previously it defaulted to Always and couldn't be changed, what was problematic for example in local development.

```release-note
imagePullPolicy for the operator deployment can be set in tetragonOperator.image.pullPolicy Helm value and defaults to IfNotPresent.
```


michi-covalent (approved):
😇

kkourt (approved):

PR #1548, opened 2023-10-03
Refactor and extend the interface for metrics with configurable labels

lambdanis:
See commits for details.

```release-note
Expose an interface for defining metrics with configurable labels.
```


jrfastab (approved):

pkg/metrics/granularmetric.go
```diff
@@ -0,0 +1,226 @@
...
+package metrics
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/cilium/tetragon/pkg/metrics/consts"
+	"github.com/cilium/tetragon/pkg/option"
+	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/exp/slices"
```
mtardy, line 13: hey @lambdanis I think you reintroduced `golang.org/x/exp/slices` here, I don't know if it's intended since you removed that https://github.com/cilium/tetragon/pull/1560.

mtardy: I just rebased one of my PR and noticed that go.mod needed to include the new dependency.

lambdanis: I think the tests passed on this PR a while ago, before some deps got updated?
Anyway, thanks for fixing this import.

PR #1549, opened 2023-10-03
Add option to tetra to filter on event types

darox:
This PR adds the ability to filter on the event types.


darox:
Go tests fail but locally they're passing:
```
ok      github.com/cilium/tetragon/cmd/tetra/getevents  0.025s
```

mtardy:
Hey!! Thanks! I have this in the test right now:

```
Error: invalid value for "event-types" flag: INEXISTENT. Supported are UNDEF, PROCESS_KPROBE, PROCESS_TRACEPOINT, PROCESS_UPROBE, TEST, RATE_LIMIT_INFO, PROCESS_EXEC, PROCESS_EXIT, PROCESS_LOADER
Usage:
  getevents [flags]

Flags:
      --color string             Colorize compact output. auto, always, or never (default "auto")
  -e, --event-types strings      Include only events of given types
  -F, --exclude-fields strings   Exclude fields from events
  -h, --help                     help for getevents
      --host                     Get host events
  -f, --include-fields strings   Include only fields in events
  -n, --namespaces strings       Get events by Kubernetes namespaces
  -o, --output string            Output format. json or compact (default "json")
      --pods strings             Get events by pods name regex
      --processes strings        Get events by processes name regex
      --timestamps               Include timestamps in compact output
  -t, --tty-encode string        Encode terminal data by file path (all other events will be ignored)

panic: write |1: broken pipe

goroutine 8 [running]:
github.com/cilium/tetragon/pkg/testutils.MockPipedFile.func1()
	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/testutils/cmd.go:153 +0xe5
created by github.com/cilium/tetragon/pkg/testutils.MockPipedFile in goroutine 7
	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/testutils/cmd.go:147 +0xf7
FAIL	github.com/cilium/tetragon/cmd/tetra/getevents	0.025s
```

So I guess there may be an issue on how the one supposed to fail/do nothing is written or return.

darox:
I did some more testing, and apparently, it's a flake. Because sometimes it works even with:

```
go test -p 1 -parallel 1  -gcflags= -timeout 40m -failfast -cover ./pkg/... ./cmd/... ./operator/...
```

Result
```
ok      github.com/cilium/tetragon/cmd/tetra/getevents  0.023s  coverage: 78.6% of statements
```

cmd/tetra/getevents/getevents.go
```diff
@@ -151,6 +161,19 @@ func New() *cobra.Command {
...
+			var supportedEventTypes string
+
+			for _, v := range tetragon.EventType_name {
+				supportedEventTypes += v + ", "
+			}
+			supportedEventTypes = strings.TrimSuffix(supportedEventTypes, ", ")
+
+			for _, v := range Options.EventTypes {
+				if _, ok := tetragon.EventType_value[v]; !ok {
+					return fmt.Errorf("invalid value for %q flag: %s. Supported are %s", "event-types", v, supportedEventTypes)
```
mtardy, line 173: you do return an error here (that we retrieve in the test output), which will make the command error, I would suggest just filtering dynamically with the provided eventtypes without checking if it exists. filtering by DOESNOTEXIST should just return nothing as you wrote in your test?

darox: The idea is that when a user passes an argument that is not any of:
```
PROCESS_KPROBE, PROCESS_LOADER, TEST, PROCESS_UPROBE, RATE_LIMIT_INFO, UNDEF, PROCESS_EXEC, PROCESS_EXIT, PROCESS_TRACEPOINT
```

it should fail. 

Now, the following test always returns `0`

```
	t.Run("FilterInexistent", func(t *testing.T) {
		testutils.MockPipedFile(t, testutils.RepoRootPath("testdata/events.json"))
		cmd := New()
		cmd.SetArgs([]string{"--event-types", "INEXISTENT"})
		output := testutils.RedirectStdoutExecuteCmd(t, cmd)
		assert.Equal(t, 0, bytes.Count(output, []byte("\n")))
	})
```

How can I then sometimes fail and sometimes pass? 

Two attempts:
image

mtardy: I think it's `RedirectStdoutExecuteCmd` that doesn't expect the command to fail (return something different than 0) and maybe there's a race somewhere in the piping explaining the broken pipe thingie. I admit I cannot reproduce on my machine even running dozens of times without caching :/

darox: Interesting. I wonder if we really need a test for the failing case or if we can just remove it.

mtardy: let me propose you something

mtardy: ah we are seeing the panic because it's `MockPipedFile` that doesn't support the cmd to fail, stdin is still piped to the entry of the command which exits before the end of the test because of the error.

cmd/tetra/getevents/getevents_test.go
```diff
@@ -31,6 +31,24 @@ func Test_GetEvents_Namespaces(t *testing.T) {
...
+		assert.Equal(t, 3, bytes.Count(output, []byte("\n")))
+	})
+
+	t.Run("FilterInexistent", func(t *testing.T) {
+		testutils.MockPipedFile(t, testutils.RepoRootPath("testdata/events.json"))
+		cmd := New()
+		cmd.SetArgs([]string{"--event-types", "INEXISTENT"})
+		output := testutils.RedirectStdoutExecuteCmd(t, cmd)
+		assert.Equal(t, 0, bytes.Count(output, []byte("\n")))
+	})
```
mtardy, line 49:
```suggestion
	t.Run("FilterInexistent", func(t *testing.T) {
		cmd := New()
		cmd.SetArgs([]string{"--event-types", "INEXISTENT"})

		// redirect stderr and stdout to not pollute test outputs. Here it
		// works, compared to other use case where we have to redirect manually,
		// because Cobra internal error mechanism calls internal Print methods.
		var devNull bytes.Buffer
		cmd.SetErr(&devNull)
		cmd.SetOut(&devNull)

		err := cmd.Execute()
		assert.Error(t, err)
	})
```

cmd/tetra/getevents/getevents.go
```diff
@@ -151,6 +161,19 @@ func New() *cobra.Command {
...
+				}
+			}
```
mtardy, line 175:
```suggestion
			for _, v := range Options.EventTypes {
				if _, found := tetragon.EventType_value[v]; !found {
					var supportedEventTypes string
					for _, v := range tetragon.EventType_name {
						supportedEventTypes += v + ", "
					}
					supportedEventTypes = strings.TrimSuffix(supportedEventTypes, ", ")
					return fmt.Errorf("invalid value for %q flag: %s. Supported are %s", "event-types", v, supportedEventTypes)
				}
			}
```

I would compute the error message only if needed:
```go
var supportedEventTypes string
for _, v := range tetragon.EventType_name {
	supportedEventTypes += v + ", "
}
supportedEventTypes = strings.TrimSuffix(supportedEventTypes, ", ")
```

mtardy (commented):
This should fix your test case failing! :) And a small comment.

cmd/tetra/getevents/getevents.go
darox: Makes sense. Thank you.

darox:
> Could you squash your commit into the first one? :) thanks!!
> 
> take this opportunity to also start the commit message with `cmd/tetra: ....`

done

mtardy (approved):
Thanks a lot for taking the time to implement that! :) 🎉 🌮

darox:
> Thanks a lot for taking the time to implement that! :) 🎉 🌮
> 
> I'll let @willfindlay approve/merge.

No problem. It was a pleasure to work on the first PR for tetragon.

PR #1551, opened 2023-10-04
tetragon: fix crash in kprobe validation

olsajiri:
fix crash in kprobe validation


kevsecurity (approved):
LGTM

mtardy:
btw this is the kind of stuff (interdependence between two fields) that we can manage directly with https://kubernetes.io/blog/2022/09/23/crd-validation-rules-beta/. It could be interesting to have more validation directly in spec!

PR #1554, opened 2023-10-05
proc_reader: avoid warning for pid=0

kkourt:
In some cases, we try to get the docker of pid=0, which fails and produces the following warning:
time="2023-10-04T10:27:14Z" level=warning msg="Procfs execve event pods/ identifier error" error="open /procRoot/0/cgroup: no such file or directory"

This warning is confusing, so avoid it by not calling the relevant code for pid=0.


tixxdz (approved):
I'm not sure if inside procsDockerId() check and return "" and nil is better.

Thanks!

PR #1555, opened 2023-10-05
fix: add `kind: TracingPolicy` on new kprobe validation test

mtardy:
A PR that ensures the validity of all TracingPolicy was merged just before a PR adding this test, thus breaking the validity of Tracing Policies.


mtardy:
Fix in https://github.com/cilium/tetragon/pull/1514

PR #1557, opened 2023-10-05
Renovate: Ignore digest updates for k8s dependencies

lambdanis:
Renovate is configured to ignore minor and major updated of k8s dependencies, because they need to be bumped together with Cilium. However, some k8s dependencies (`k8s.io/kube-openapi`) are not tagged with a version, but pinned by digest. Their updates are incompatible with other k8s deps, what's blocking updates of other Go dependencies, see #1372. This PR updates the Renovate config to ignore digest updates for k8s deps too.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -336,6 +336,7 @@
       "matchUpdateTypes": [
         "major",
         "minor",
+        "digest",
       ],
     }
   ],


mtardy (approved):
Oh thanks a lot! Indeed that should remove this update from the grouped thing! :)

mtardy:
We still have patch but those should hopefully work? At least it was the idea, it's annoying that some dependencies are not versioned together.

lambdanis:
> We still have patch but those should hopefully work? At least it was the idea, it's annoying that some dependencies are not versioned together.

Should work, although kube-openapi will be excluded from automatic updates and it seems that it's compatibility and versioning is being discussed (https://github.com/kubernetes/kube-openapi/issues/421), so we'll see.

PR #1553, opened 2023-10-04
vendor: Pick up github.com/cilium/cilium v1.15.0-pre.1

michi-covalent:
pkg/tracingpolicy/generictracingpolicy.go: https://github.com/cilium/tetragon/pull/1553/files#diff-a2a732fedac395fb24099a9ae8b54f09a65ae9d88a678709cc2ee4a1e503fa65

PR #1560, opened 2023-10-06
Use Go stdlib instead of golang.org/x/exp

lambdanis:
Tetragon uses slices and maps functions from `golang.org/x/exp`. In Go 1.21 they
are a part of the Go standard library, so let's import them from there instead.


mtardy (approved):
I saw some similar changes in https://github.com/cilium/tetragon/pull/1553 but yes thanks for that! One less direct dependency to upgrade :)

PR #1562, opened 2023-10-06
Fix a regression on tetra tracingpolicy enable/disable

### LINKED ISSUE #1489: Regression on `v0.11.0` when re-enabling TracingPolicy with sensors API
### What happened?

I got a regression with Tetragon `v0.10.0` -> `v0.11.0`, there is policy for tcp connect events:
```yaml
kind: TracingPolicy
metadata:
  name: connect
spec:
  kprobes:
  - args:
    - index: 0
      returnCopy: false
      type: sock
    call: tcp_connect
    return: false
    syscall: false
  - args:
    - index: 0
      returnCopy: false
      type: sock
    call: tcp_close
    return: false
    syscall: false
  - args:
    - index: 0
      returnCopy: false
      type: sock
    - index: 2
      returnCopy: false
      type: int
    call: tcp_sendmsg
    return: false
    syscall: false
```

And if I disable it with `tetra sensors disable connect`  and try to re-enable: `tetra sensors enable connect` it returns an error and stays disabled:
```
# tetra sensors enable connect
failed to enable sensor connect: rpc error: code = Unknown desc = sensor gkp-sensor-1 from collection connect failed to load: failed prog /var/lib/tetragon/bpf_generic_kprobe_v53.o kern_version 328939 loadInstance: getting entry from genericKprobeTable failed with: invalid id (ID=0/invalid entry)
```

Tetragon logs:
```
time="2023-09-18T18:40:00Z" level=warning msg="Server EnableSensor request failed" error="sensor gkp-sensor-1 from collection connect failed to load: failed prog /var/lib/tetragon/bpf_generic_kprobe_v53.o kern_version 328939 loadInstance: getting entry from genericKprobeTable failed with: invalid id (ID=0/invalid entry)" sensor.name=connect
```

This all worked well with Tetragon `v0.10.0`, never had an issue with re-enabling this policy.

### Tetragon Version
```
# tetra version
CLI version: v0.11.0
```

### Kernel Version
```
$ uname -a
Linux mint-vm 5.4.0-155-generic #172-Ubuntu SMP Fri Jul 7 16:10:02 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
```

### Kubernetes Version
```
$ kubectl version
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"27", GitVersion:"v1.27.4", GitCommit:"fa3d7990104d7c1f16943a67f11b154b71f6a132", GitTreeState:"clean", BuildDate:"2023-07-19T12:20:54Z", GoVersion:"go1.20.6", Compiler:"gc", Platform:"linux/amd64"}
Kustomize Version: v5.0.1
Server Version: version.Info{Major:"1", Minor:"26", GitVersion:"v1.26.3", GitCommit:"9e644106593f3f4aa98f8a84b23db5fa378900bd", GitTreeState:"clean", BuildDate:"2023-03-30T06:34:50Z", GoVersion:"go1.19.7", Compiler:"gc", Platform:"linux/amd64"}
```

### Bugtool

[tetragon-bugtool.tar.gz](https://github.com/cilium/tetragon/files/12653262/tetragon-bugtool.tar.gz)

### Relevant log output

_No response_

### Anything else?

_No response_

mtardy:
Fixes #1489.

See individual commits, this PR:
- Fixes a regression on enable/disable because of `PostUnloadHook`
- Adds new enable/disable on tracingpolicy cmd
- Reworks the tracingpolicy commands so that we get accurate exit code
- Deprecates the sensors command
- Extends the TracingPolicyStatus to have a better gRPC API to list TracingPolicies


pkg/sensors/handler.go
```diff
@@ -181,6 +182,46 @@ func (h *handler) listTracingPolicies(op *tracingPolicyList) error {
...
+	}
+
+	if err := col.load(h.bpfDir, h.mapDir); err != nil {
+		return err
+	}
+
+	col.enabled = true
+	h.collections[op.name] = col
+	return nil
+}
```
mtardy, line 225: Will the policyID part will gracefully work with enable/disable?

mtardy:
maybe it would be reasonable to add a test covering that regression.

mtardy:
I added a test in https://github.com/cilium/tetragon/pull/1562/commits/c738ea3de19c17f08b9d014a854ab17f5b660e57.

pkg/sensors/load.go
```diff
@@ -149,6 +153,25 @@ func (s *Sensor) Unload() error {
...
+		return err
+	}
+
+	if s.DestroyHook != nil {
+		err = s.DestroyHook()
+		if err != nil {
+			return err
+		}
+	}
+	s.Destroyed = true
```
kkourt, line 170: So what happens if either Unload or Destroy fails? do we have a way to recover?
I'm wondering whether it makes sense to keep the s.Destroyed state or we if it's better to just remove the sensor altogether if the unload or the hook fails.

pkg/sensors/sensors.go
```diff
@@ -41,19 +41,25 @@ type Sensor struct {
...
+	// being triggered and must be recreated.
+	DestroyHook SensorHook
 }
 
-// SensorUnloadHook is the function signature for an optional function
-// that can be called during sensor unloading.
-type SensorUnloadHook func() error
+// SensorHook is the function signature for an optional function
+// that can be called during sensor unloading and removing.
+type SensorHook func() error
```
kkourt, line 62: nit: I wonder if it makes sense to have a different type for the two hooks.

pkg/sensors/handler.go
```diff
@@ -181,6 +182,46 @@ func (h *handler) listTracingPolicies(op *tracingPolicyList) error {
...
+	}
+
+	err := col.unload()
+	if err != nil {
+		col.err = fmt.Errorf("failed to unload tracing policy: %w", err)
+		return err
+	}
+
+	col.enabled = false
+	h.collections[op.name] = col
```
kkourt, line 204: ah... that's tricky.
Maybe it would have been better to use a pointer in the map.

pkg/sensors/handler.go
kkourt: I don't see why not. The policyfilter map will not be affected and the policy id will not change when we unload things.

cmd/tetra/tracingpolicy/tracingpolicy.go
```diff
@@ -144,7 +144,33 @@ func New() *cobra.Command {
 					}
 
 					sensors := strings.Join(pol.Sensors, ",")
-					cmd.Printf("%d %s (%s) %s %s\n", pol.Id, pol.Name, pol.Info, namespace, sensors)
+
+					// Previously, enabled, filterID and error were bundled in a
+					// string. To have a retro-compatible tetra command, we scan
+					// the string. If the scan fails, it means something else
+					// might be in Info and we print it.
```
kkourt, line 151: Let's make a note of the version that required this (0.11) so that we can remove it eventually.

kkourt (commented):
Thanks! LGTM!

The only thing I'm not sure about is what to do when destroy fails. If we cannot recover from such a case, I'd suggest we just remove the sensor anyway and issue a warning. I'm happy to be convinced otherwise though.

All other comments are just nits.

pkg/sensors/handler.go
mtardy: indeed, but I think it's okay!

pkg/sensors/load.go
mtardy: we can keep the `s.Destroyed` state indeed I agree but what do you imply by removing the sensor altogether? What could we do more?

pkg/sensors/sensors.go
mtardy: idk probably if `SensorUnloadHook` didn't exist I would have put `func() error`, what do you think would be the advantage of adding types?

pkg/sensors/load.go
mtardy: we could `defer s.Destroyed = true` at the beginning basically. 

I'm not sure it would be very fair for situations where it just failed for casual reasons like Unload failed for "sensor not loaded".

What would be your idea? To run the destroy hook and affect the Destroyed state even if the Unload failed? And Destroyed state even if the hook failed?

kkourt: So let's say user delete tracing policy. so we are in `deleteTracingPolicy`.

If everything goes well, we will call `delete(h.collections, op.name)` and the user will have no way of accessing the sensor.

If `s.Destroy()` returns an error, then the collection will not be deleted and the tracingpolicy will remain the sensor manager in a defunc state. There is nothing the user can do at this point.

So my two quetions are:
* Why do we keep the s.Destroyed variable? What is a scenario that we will end up using the value? (we reading it in Load(), so can we ever call Load after we call Destroy()? This seems weird to me. If a sensor is "destroyed", why are we holding a reference to it?
* What happens if unload hook fails? Then, the user will not be able to load a new policy with the same name and they will also not be able to remove the old policy that is now defunct. It seems better to me to remove the collection even if the Destroy() function fails.

Thoughts?

mtardy: ah yes I understand, I didn't think of `delete(h.collections, op.name)` because this is what is done by the caller of this function!

> Why do we keep the s.Destroyed variable? What is a scenario that we will end up using the value? (we reading it in Load(), so can we ever call Load after we call Destroy()? This seems weird to me. If a sensor is "destroyed", why are we holding a reference to it?

Yes agree, the `destroyed` state is just to prevent stuff and should not be triggered in normal use.

> What happens if unload hook fails? Then, the user will not be able to load a new policy with the same name and they will also not be able to remove the old policy that is now defunct. It seems better to me to remove the collection even if the Destroy() function fails.

Sure! We could basically do something like this:

```diff
func (h *handler) deleteTracingPolicy(op *tracingPolicyDelete) error {
	col, exists := h.collections[op.name]
	if !exists {
		return fmt.Errorf("tracing policy %s does not exist", op.name)
	}
+	defer delete(h.collections, op.name)
	
	err := col.destroy()
	if err != nil {
		col.err = fmt.Errorf("failed to unload tracing policy: %w", err)
		return err
	}

	filterID := policyfilter.PolicyID(col.policyfilterID)
	err = h.pfState.DelPolicy(filterID)
	if err != nil {
		col.err = fmt.Errorf("failed to remove from policyfilter: %w", err)
		return err
	}

-	delete(h.collections, op.name)
	return nil
}
```

Only from the `s.Destroy()` point of view, what do you think of the Destroyed state, should it automatically set when you call that even if it fails somehow. I just noticed that in `s.Unload()` the only returned error is `unload of sensor %s failed: sensor not loaded`, the rest is just logged for the failure of the PreUnload and PostUnload hooks. So maybe we could do something like that:

```diff
// Destroy will unload the hook and call DestroyHook, this hook is usually used
// to clean up resources that were created during creation of the sensor.
func (s *Sensor) Destroy() error {
	err := s.Unload()
	if err != nil {
		return err // <- only error possible is sensor was not loaded, should destroy fail for that? maybe it's okay to destroy an unloaded sensor and thus we can move the `s.Destroyed = true` before this.
	}

	if s.DestroyHook != nil {
	        err = s.DestroyHook()
+               logger.GetLogger().WithError(err).WithField("sensor", s.Name).Warn("Destroy hook failed")
-		if err != nil {
-			return err
-		}
	}
	s.Destroyed = true

	return nil
}
```

What do you think?

mtardy: if we don't accept any error from `Destroy` like it always succeed as noted in the comment above we can do that then:

```diff
func (h *handler) deleteTracingPolicy(op *tracingPolicyDelete) error {
	col, exists := h.collections[op.name]
	if !exists {
		return fmt.Errorf("tracing policy %s does not exist", op.name)
	}
+	defer delete(h.collections, op.name)
	
	err := col.destroy()
-	if err != nil {
-		col.err = fmt.Errorf("failed to unload tracing policy: %w", err)
-		return err
-	}

	filterID := policyfilter.PolicyID(col.policyfilterID)
	err = h.pfState.DelPolicy(filterID)
	if err != nil {
		col.err = fmt.Errorf("failed to remove from policyfilter: %w", err)
		return err
	}

-	delete(h.collections, op.name)
	return nil
}
```

When you delete a policy, is it important to know that you could not unload it because it was never loaded in the first place? I guess no. The only thing is that we are making a lot of assumptions about the return case that we call and this can change in the future.

kkourt: Above sounds good to me. The only minor comment would be that we probably want something like:
```
if (err) {
     logger.GetLogger().WithError(err).WithField("sensor", s.Name).Warn("Destroy hook failed")
 }
```
 
 in case err = s.Destroy() hook is not nill

pkg/sensors/sensors.go
kkourt: More strict typing. Let's leave it as is though.

pkg/sensors/load.go
mtardy: > in case err = s.Destroy() hook is not nill
ah yes yes sure, I wrote too fast

mtardy: What's your stance on ignoring or not the error from `s.Unload` in `s.Destroy`? It feels like we should, does this sound good to you? It would result in something like this:

```diff
func (h *handler) deleteTracingPolicy(op *tracingPolicyDelete) error {
	col, exists := h.collections[op.name]
	if !exists {
		return fmt.Errorf("tracing policy %s does not exist", op.name)
	}
+	defer delete(h.collections, op.name)
	
	col.destroy() // <- doesn't return an error
-	if err != nil {
-		col.err = fmt.Errorf("failed to unload tracing policy: %w", err)
-		return err
-	}

	filterID := policyfilter.PolicyID(col.policyfilterID)
	err = h.pfState.DelPolicy(filterID)
	if err != nil {
		col.err = fmt.Errorf("failed to remove from policyfilter: %w", err)
		return err
	}

-	delete(h.collections, op.name)
	return nil
}
```

kkourt: I agree. Let's go with that, and we can always change it if the need arises. Thanks!

mtardy: See the simplification in the new push https://github.com/cilium/tetragon/pull/1562/commits/03f40739b741a087c992dc01cce861f7ff921ddb. :)

pkg/sensors/collection.go
```diff
@@ -80,3 +82,10 @@ func (c *collection) unload() error {
 	}
 	return nil
 }
+
+// destroy will attempt to destroy all the sensors in a collection
+func (c *collection) destroy() {
+	for _, s := range c.sensors {
+		s.Destroy()
+	}
+}
```
mtardy, line 91: better like that

kkourt (approved):

PR #1575, opened 2023-10-10
deployment: configure logs

tixxdz:


install/kubernetes/values.yaml
```diff
@@ -52,10 +52,15 @@ tetragon:
...
-  #     value: bar
+  extraEnv:
+    # Tetragon can be deployed in different forms, from a Kubernetes pod, container,
+    # or a systemd service, having this environment variable helps to properly
+    # detect the environment where Tetragon is running.
+    - name: TETRAGON_DEPLOYMENT_MODE
+      value: k8s
+    # Tetragon export logs file permissions as a string, default readable/writable by owner only.
+    - name: TETRAGON_LOGS_PERM
+      value: "0660"
```
michi-covalent, line 63: ah yeah sorry i shouldn't have suggested to use `extraEnv`, i thought you were just testing something out locally to pass some environment variables. i think these should be proper helm values.

there are already existing helm values related to export files, so i think we should add:
```
tetragon.exportFilePermission: "0660"
```
or some such setting instead of using extraEnv.

regarding `TETRAGON_DEPLOYMENT_MODE`, i'm kind of hoping tetragon can somehow autodetect it instead of asking users to specify it 💭

tixxdz: for the extraEnv, I thought we can get a way with it ;-) as didn't want to overcharge tetragon --flags, as by default they should be restricted, and those permission make sense only under k8s... anyway I made the change now as you suggested!

For TETRAGON_DEPLOYMENT_MODE I removed it , and yes we have some autologic detection, that first checks --enable-k8s-api and assumes kubernetes, otherwise it does some foo cgroup namespace detection.

Then the improvement later would be to differentiate between k8s and pure container (without k8s) deployments as the container tries to hide some information, but will improve it later to do detection from kernel side.

michi-covalent (approved):

PR #1579, opened 2023-10-11
renovate: fix renovate upgrade of Go toolchain

mtardy:
PR #1536 broke "renovate update Go toolchain in a single PR" from #1259 because in the actual state it could only bump the go directive from minor versions, not bumping the patch needed by the workflows github actions reading that version.

From cilium/cilium@2f7e2f33659776a1c2adcf80111f69d6debdd850: According to [1] as of Go 1.21 we either need to specify the full toolchain version in the `go` directive or add a `toolchain` directive with the concrete toolchain version. Opt for the former and make sure it's kept up to date by renovate bot.

[1] golang/go#62278 (comment)

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -187,13 +187,6 @@
         "Dockerfile"
       ],
     },
-    // Upgrade the Go module directive, see:
-    // https://docs.renovatebot.com/modules/versioning/#go-modules-directive-versioning
-    // depName is 'go' so it will be group with the 'Go' groupName
-    {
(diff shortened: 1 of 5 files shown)


michi-covalent (approved):
sure

willfindlay (approved):
If it's good enough for Cilium, it's good enough for me

mtardy:
Thanks that was fast, hopefully, this works! :) We'll see here https://github.com/cilium/tetragon/pull/1577

PR #1588, opened 2023-10-12
alignchecker: refactor into single gotest

willfindlay:
This series refactors the alignchecker into a single gotest to make editing the test case easier going forward.


willfindlay:
Marking as draft while I test something quickly

tpapagian (approved):

PR #1589, opened 2023-10-12
refactor: Move flags.go to pkg/option

michi-covalent:
Logically flags.go belongs to pkg/option, so let's just move it there. I also had to move ParseMetricsLabelFilter to pkg/option to break a cyclic dependency.


mtardy:
Cool, I wanted to do that one day and I didn't!

tixxdz (approved):
Much appreciated @michi-covalent

lambdanis (approved):
cool makes sense

PR #1591, opened 2023-10-13
pkg/metrics: remove unused import

kkourt:


kkourt:
Let's use https://github.com/cilium/tetragon/pull/1592

PR #1594, opened 2023-10-16
custom event handling

kkourt:
see patches

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -1148,6 +1148,12 @@ func handleGenericKprobe(r *bytes.Reader) ([]observer.Event, error) {
 		return nil, fmt.Errorf("Failed to match id")
 	}
 
+	return handleMsgGenericKprobe(&m, gk, r)
+}
+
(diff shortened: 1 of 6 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -1148,7 +1155,11 @@ func handleGenericKprobe(r *bytes.Reader) ([]observer.Event, error) {
 		return nil, fmt.Errorf("Failed to match id")
 	}
 
-	return handleMsgGenericKprobe(&m, gk, r)
+	ret, err := handleMsgGenericKprobe(&m, gk, r)
```
mtardy, line 1158: Here just making sure you don't want to return the error if something happens? It will be overridden by the next call if the customHandler is not nil.

mtardy (approved):
Looks pretty innocent, just a little comment on the flow.

tpapagian (approved):

pkg/sensors/tracing/generickprobe.go
kkourt: yap, it's intentional. The handler can decide what to do both with the event and the error.

PR #1603, opened 2023-10-17
Various fixes, mostly on form and organization

mtardy:
See commits


michi-covalent:
alright let's see what happens

PR #1608, opened 2023-10-18
tetragon: Use pull_request_target event for e2e CI tests

olsajiri:
As suggested by Mahe, so it can run properly on fork-ed repos.


mtardy (approved):
thanks, it should fix the issue

PR #1609, opened 2023-10-18
tetragon: Add support to override security_ functions

olsajiri:
Adding support to override 'security_' kernel functions with
fmod_ret override program.

The fmod_ret program is specific and differs to kprobe override
program in following ways:

  1) it needs to be loaded with BTF ID of function it's attached to
  2) it's invoked from BPF trampoline which is executed as direct call
     after the ftrace trampoline is done

The 1) prevents us to have single program for multiple functions we
want to override, hence for this moment we allow to fmod_ret override
only in non kprobe_multi mode (--disable-kprobe-multi option).

This could be changed in future, but might imply some performance hits.

The 2) adds extra code paths execution compared to invoking kprobe override
program, so we are using fmod_ret override only for 'security_' functions
and kprobe override for the rest.

### DIFF
--- a/bpf/process/bpf_generic_kprobe.c
+++ b/bpf/process/bpf_generic_kprobe.c
@@ -302,3 +302,20 @@ generic_kprobe_override(void *ctx)
 	map_delete_elem(&override_tasks, &id);
 	return 0;
 }
+
+/* Putting security_task_prctl in here to pass contrib/verify/verify.sh test,
+ * in normal run the function is set by tetragon dynamically.
+ */
+__attribute__((section("fmod_ret/security_task_prctl"), used)) long
+generic_fmodret_override(void *ctx)
+{
+	__u64 id = get_current_pid_tgid();
+	__s32 *error;
+
+	error = map_lookup_elem(&override_tasks, &id);
+	if (!error)
+		return 0;
+
+	map_delete_elem(&override_tasks, &id);
+	return (long)*error;
+}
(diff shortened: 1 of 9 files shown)


pkg/bpf/detect.go
```diff
@@ -143,3 +143,25 @@ func HasModifyReturn() bool {
...
+	if HasKprobeMulti() {
+		feats += "kprobe_multi "
+	}
+	if HasModifyReturn() {
+		feats += "fmodret "
+	}
+	// cut the final ' '
+	if len(feats) != 0 {
+		feats = feats[:len(feats)-1]
+	}
```
kkourt, line 165: How about printing something like: "override_return: true, buildid: false"? It's more verbose, but it also provides more info.

bpf/process/bpf_generic_kprobe.c
```diff
@@ -302,3 +302,17 @@ generic_kprobe_override(void *ctx)
 	map_delete_elem(&override_tasks, &id);
 	return 0;
 }
+
+__attribute__((section("fmod_ret/security_task_prctl"), used)) long
```
kkourt, line 306: I guess we do not always attach this to security_task_prctl, correct? Can we remove the section, or is it needed?

kkourt (approved):
Thanks, LGTM!

Just a couple of minor comments.

pkg/bpf/detect.go
olsajiri: ok, also the code will be easier ;-) will change

bpf/process/bpf_generic_kprobe.c
olsajiri: it's needed for the `contrib/verify/verify.sh` script to pass, so it needs to stay, but I forgot to add some comments about that, will add

PR #1612, opened 2023-10-18
vendor: fix cilium/tetragon replacement in api

willfindlay:
We had `../../tetragon` as the go mod replacement for the api module. But what happens when the directory name of the git repo isn't `tetragon`? We should handle this case, so change it to `..` directly.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -25,4 +25,4 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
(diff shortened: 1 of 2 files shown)


glibsm (approved):

mtardy (approved):
Good idea, I came upon this weird thing when writing the one for this https://github.com/cilium/tetragon/blob/577dfe69576cab29436f892a8383f2ef35210843/contrib/rthooks/tetragon-oci-hook/go.mod#L39.

PR #1622, opened 2023-10-19
Events: Support events from kernel and init

kevsecurity:
Tetragon supports events from user space processes. These processes are stored in the execve map and in the process cache. Events that arrive in the agent without all the information they need, or arrive before the execve event arrives, are placed temporarily in the event cache, such that the information is likely to be available shortly after, and this information can then be added to the event before it is processed.

The process cache has an entry for PID=1 (init) in order to provide parent process information for daemons and orphaned processes. If, in the future, we generate events for the init process or for the kernel itself, then these will be identified as missing information and will be placed in the event cache. Unfortunately, no extra information will become available for these processes (PID=0 and PID=1) so this just delays the events from being processed.

This commit shortcuts the event cache when PID=0 or PID=1, and handles a process start time of 0 (boot time) gracefully.

### DIFF
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -203,6 +203,12 @@ func (ec *Cache) Needed(proc *tetragon.Process) bool {
 	if proc == nil {
 		return true
 	}
+	// Pid=0 is the kernel; Pid=1 is the init process.
+	// For both, we will not receive any execve events or k8s watcher info
+	// to enrich the events, so the cache is not needed for these.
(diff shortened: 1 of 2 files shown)


willfindlay (approved):

jrfastab:
How come we don't have an entry for the init process?

kevsecurity:
Unnecessary if ktime=1 is set on pid=0 events.

kevsecurity:
> How come we don't have an entry for the init process?

We do, if ktime=1.

PR #1624, opened 2023-10-19
tetragon: docs, minor updates to metrics concepts

jrfastab:
Update Metrics section to clarrify PACKAGE is about local installs. I at least was confused.


mtardy (approved):
Thanks!

PR #1626, opened 2023-10-20
tetragon: Add support to pass options through spec

olsajiri:
Adding support to process options passed in spec for kprobe sensor.
    
At the moment the only supported option is to disable kprobe multi
(like with --disable-kprobe-multi option).


pkg/sensors/tracing/options.go
```diff
@@ -0,0 +1,56 @@
...
+			name: option.KeyDisableKprobeMulti,
+			set: func(str string) (err error) {
+				options.DisableKprobeMulti, err = strconv.ParseBool(str)
+				return err
+			},
+		},
+	}
+
+	// cross reference allowed with passed options
+	for i := range specs {
```
tpapagian, line 40: nit: You can use something like:
```
	for _, spec := range specs {
		for _, opt := range opts {
```
to remove the need to declare new variables.

tpapagian: You can also have a global variable `opts` instead of having to initialize that every time we call that function. Additionally, you can use a map (key: name, value: set) and remove the need for the nested loop. 

(feel free to ignore these comments as this is not a performance-critical code, but I believe that these changes will make the code a bit easier to read)

tpapagian (approved):
Thanks! I have some small comments but nothing critical.

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -95,8 +95,12 @@ type TracingPolicySpec struct {
 	Lists []ListSpec `json:"lists,omitempty"`
 
 	// +kubebuilder:validation:Optional
+
 	// A killer spec.
 	Killers []KillerSpec `json:"killers,omitempty"`
+
+	// A list of overloaded options
+	Options []OptionSpec `json:"options,omitempty"`
```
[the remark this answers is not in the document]
olsajiri: ah ok, will add

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -95,8 +95,12 @@ type TracingPolicySpec struct {
 	Lists []ListSpec `json:"lists,omitempty"`
 
 	// +kubebuilder:validation:Optional
+
```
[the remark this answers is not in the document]
olsajiri: yep, missed that, thanks

pkg/sensors/tracing/options.go
olsajiri: yes, it's better with the map, will change that, thanks

kkourt (approved):

PR #1627, opened 2023-10-20
tetragon: namespace checks

tixxdz:


pkg/reader/namespace/namespace.go
```diff
@@ -180,30 +207,49 @@ func GetMsgNamespaces(ns processapi.MsgNamespaces) *tetragon.Namespaces {
...
-			Cgroup:          createHostNs("cgroup"),
-			User:            createHostNs("user"),
+func InitHostNamespace() (*tetragon.Namespaces, error) {
+	if hostNamespaces != nil {
+		return hostNamespaces, nil
+	}
+
+	kernelVer, _, _ := kernels.GetKernelVersion(option.Config.KernelVersion, option.Config.ProcFS)
+	// time and time_for_children namespaces introduced in kernel 5.6
+	TimeNsSupport = (int64(kernelVer) >= kernels.KernelStringToNumeric("5.6.0"))
```
tpapagian, line 220: Can we check if the kernel supports time namespace by checking if `/proc/1/ns/time` exists? I believe that this is a good idea not to rely on the kernel version. For example, the 4.18 redhat kernel may have backported time namespace.

(I know that we did that before, maybe it is a good time to fix that.)

pkg/reader/namespace/namespace.go
```diff
@@ -180,30 +207,49 @@ func GetMsgNamespaces(ns processapi.MsgNamespaces) *tetragon.Namespaces {
...
-			Mnt:             createHostNs("mnt"),
-			Pid:             createHostNs("pid"),
-			PidForChildren:  createHostNs("pid_for_children"),
-			Net:             createHostNs("net"),
-			Time:            createHostNs("time"),
-			TimeForChildren: createHostNs("time_for_children"),
-			Cgroup:          createHostNs("cgroup"),
-			User:            createHostNs("user"),
+func InitHostNamespace() (*tetragon.Namespaces, error) {
+	if hostNamespaces != nil {
```
tpapagian, line 214: As we are at Golang 1.21, we can use https://pkg.go.dev/sync#OnceValue to be safe for any races.

pkg/reader/namespace/namespace.go
```diff
@@ -180,30 +207,49 @@ func GetMsgNamespaces(ns processapi.MsgNamespaces) *tetragon.Namespaces {
...
 		}
 	}
-	return hostNamespace
-}
 
-func createHostNs(ns string) *tetragon.Namespace {
-	return &tetragon.Namespace{
-		Inum:   GetPidNsInode(1, ns),
-		IsHost: true,
+	hostNamespaces = &tetragon.Namespaces{
```
tpapagian, line 242: Do we really need both `hostNamespaces` and `knownNamespaces`? Can we initialize `hostNamespaces` directly?

tpapagian (commented):
Thanks! I have some comments/questions!

tpapagian (approved):
Thanks for the fixes! Please fix the static checks before merging.

tixxdz:
> Thanks for the fixes! Please fix the static checks before merging.

Much appreciated @tpapagian

PR #1630, opened 2023-10-20
docs: update enforcement page

kkourt:


docs/content/en/docs/concepts/enforcement/_index.md
```diff
@@ -5,6 +5,27 @@ weight: 4
...
+sending a signal (e.g., `SIGKILL`) to the process.
+
+
+## Override return value
+
+Override the return value of a call means that the function will never be executed and, instead, a
+value (typically an error) will be returned to the caller. Generally speaking, only system calls and
+security check functions allow to change their return value in this manner. Details about how users
+can configure tracing policies to override the return value can be found in the [Override
+action]({{< ref "/docs/concepts/tracing-policy/selectors#override-action" >}}) documentation.
```
mtardy, line 22: Should we mention the kernel option? https://elixir.bootlin.com/linux/v5.13.18/source/kernel/trace/Kconfig#L601. Not really a review of this but maybe for another PR.

docs/content/en/docs/concepts/enforcement/_index.md
```diff
@@ -5,6 +5,27 @@ weight: 4
...
+
+## Signals
+
+Another type of enforcement is signals. For example, users can write a TracingPolicy (details can be
+found in the [Signal action]({{<ref "/docs/concepts/tracing-policy/selectors#signal-action" >}})
+documentation) that sends a `SIGKILL` to a process matching certain criteria and thus terminate it.
+In contrast with overriding the return value, sending a signal does not always stop the operation
+performed by the process from being executed.  For example, a `SIGKILL` might be send in a `write()`
+system call does not guarantee that the data will not be written to the file.
```
mtardy, line 31:
```suggestion
performed by the process from being executed.  For example, a `SIGKILL` sent in a `write()`
system call does not guarantee that the data will not be written to the file.
```

mtardy (approved):

mtardy:
that's weird, the netlify preview was not triggered! the only time we need this it's not triggered 🤔

docs/content/en/docs/concepts/enforcement/_index.md
jrfastab: Or we could have a Requirements section at some point would be another option.

docs/content/en/docs/concepts/enforcement/_index.md
jrfastab: Can we highlight what it does guarantee though. Maybe,
```
In contrast with overriding the return value, sending a signal does not always stop the operation being performed by the process that triggered the operation. However, it does ensure that the process is terminated synchronously (and any threads will be stopped). To ensure the operation is also stopped the hook must be placed in the kernel carefully to ensure the kernel performs a check for any signals before the operation is performed.

Any many cases it is sufficient to ensure the process is stopped and unable to return from the call. Further, the kernel often checks signals before doing I/O operations so these can be used as long as kernel code paths are audited.
```

docs/content/en/docs/concepts/enforcement/_index.md
kkourt: I think the option would be better suited in the Override action documentation (in the selectors) section.

docs/content/en/docs/concepts/enforcement/_index.md
kkourt: Thanks! As discussed offline, I pushed the following:

```
In contrast with overriding the return value, sending a `SIGKILL` signal does not always stop the
operation being performed by the process that triggered the operation. For example, a `SIGKILL` sent
in a `write()` system call does not guarantee that the data will not be written to the file.
However, it does ensure that the process is terminated synchronously (and any threads will be
stopped). In some cases it may be sufficient to ensure the process is stopped and the process does
not handle the return of the call. To ensure the operation is not completed, though, the `Signal`
action should be combined with the `Override` action.
```

kkourt:
Merged. Let's improve later if need be.

PR #1631, opened 2023-10-20
tetragon: for now drop benchmark section

jrfastab:
I don't think we need Benchmark section for next release and I think we should spend some time on how to do repeatable benchmarks vs simply push this into the pages.

So delete section now and stay tuned for a better implementation.


mtardy (dismissed):
yep! probably drop the docs/benchmarks folder as well?

mtardy (approved):

PR #1639, opened 2023-10-23
tetra policyfilter command

kkourt:
Add a new policyfilter tetra command.  We add two sub-commands:
dump (which is the same as tetra dump policyfilter) and add to add
entries to the map. This is strictly for development/debugging, so we
mark the command hidden.


tpapagian (approved):
LGTM!

PR #1652, opened 2023-10-25
matchBinaries selector rework and add prefix operator

mtardy:
Selector matchBinaries had this flaw: it could not match processes started before Tetragon. This PR tries to fix that and add a new Prefix operator. The Postfix operator will be a bit more complex and will come in a separate PR.

This PR is not finished, only the prefix operator was added. `In` and `NotIn` still do not match processes started before Tetragon.


mtardy:
This has been implemented in two times with:
- https://github.com/cilium/tetragon/pull/1731
- https://github.com/cilium/tetragon/pull/1732

Issue #1623, opened 2023-10-19
New Tetragon logo


lizrice:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature you would like

Tetragon really needs a logo that looks great alongside the Cilium logo, with similar styling and colours, but that gives it a symbol of its own (much like Hubble has its own logo)

### Describe your proposed solution

Here's the proposed design
![alpha-back-v](https://github.com/cilium/tetragon/assets/458616/1c836504-cac1-4d6c-9d51-6e720c342861)

* It has the same colour scheme, font, and outline style as the Cilium logo, so they look good together
* Continues the shield theme
* Choice of yellow for the main T colour is a link to eBPF

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

jrfastab:
I like it. I marked it a release blocker for next release. Thanks Liz.

mtardy:
I think all the images are under:

- Favicons: https://github.com/cilium/tetragon/tree/main/docs/static/favicons
- Various images and diagrams with logo: https://github.com/cilium/tetragon/tree/main/docs/static/images
- svgs (logo): https://github.com/cilium/tetragon/tree/main/docs/static/svgs
- logo in content: https://github.com/cilium/tetragon/blob/main/docs/content/en/docs/logo.png
- default Twitter card: https://github.com/cilium/tetragon/blob/main/docs/static/default-twitter-card-image.png

PR #1658, opened 2023-10-26
docs: update tetragon logos in docs where necessary

hacktivist123:
Fixes: #1623


mtardy (commented):
Thanks a lot for doing this 🙏 , I have some comments though:

image

It would be great to replace all the favicons without the "Tetragon" text. Just the logo.



The aliasing sometimes looks completely off on the main logo in the top left corner, I'm not sure we use a SVG here but that could explain, maybe it's better to have a bitmap file like a png for this use case. It's not really problematic we can still fix this later.
Pasted Graphic


Then nits: this one is not centered, and we would need to redo the main image or make it less big but that's maybe out of scope:

image
image

hacktivist123:
> Thanks a lot for doing this 🙏 , I have some comments though:
> 
> 
> 
> image
> 
> 
> 
> It would be great to replace all the favicons without the "Tetragon" text. Just the logo.
> 
> 
> 
> 
> 
> 
> 
> The aliasing sometimes looks completely off on the main logo in the top left corner, I'm not sure we use a SVG here but that could explain, maybe it's better to have a bitmap file like a png for this use case. It's not really problematic we can still fix this later.
> 
> Pasted Graphic
> 
> 
> 
> 
> 
> Then nits: this one is not centered, and we would need to redo the main image or make it less big but that's maybe out of scope:
> 
> 
> 
> image
> 
> image
> 
> 
> 
> 

Got it! I'll replace the favicons with the logo containing just the shield. For the logo on top, I'll switch it to a png and see if it changes. 

The hero logo needs to be reworked to fit the landing page and also the one on the getting started docs page too. I tried to adjust it with css but it just wouldn't centralize.

mtardy:
> Got it! I'll replace the favicons with the logo containing just the shield. For the logo on top, I'll switch it to a png and see if it changes.

Thanks that's nice!

> The hero logo needs to be reworked to fit the landing page and also the one on the getting started docs page too. I tried to adjust it with css but it just wouldn't centralize.

yeah don't spend time and worry about that we'll do that later :)

PR #1671, opened 2023-10-30
vmtests: bump timeout and start multiple job for builds

mtardy:
The action was timing out: taking around 11min to build, 2min to create the archive, upload was more than 6/7min thus triggering the 20min limit.

### DIFF
--- a/.github/workflows/vmtests.yml
+++ b/.github/workflows/vmtests.yml
@@ -13,7 +13,7 @@ jobs:
   build:
     name: Build tetragon
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 30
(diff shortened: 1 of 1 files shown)


.github/workflows/vmtests.yml
```diff
@@ -39,7 +39,7 @@ jobs:
         GOPATH: /home/runner/work/tetragon/tetragon/go
       run: |
         cd go/src/github.com/cilium/tetragon/
-        make tetragon-bpf tester-progs test-compile
+        make -j3 tetragon-bpf tester-progs test-compile
```
kkourt, line 42: Why the `-j3` here?

kkourt (approved):

PR #1674, opened 2023-10-30
Add a metric to provide missed events per type

tpapagian:
Example:
```
$ curl localhost:2112/metrics 2> /dev/null | grep 'sent_events_total\|missed_events_total\|ringbuf_perf_event_lost_total\|ringbuf_queue_lost_total\|msg_op_total\|ringbuf_queue_received_total'
# HELP tetragon_missed_events_total The total number of Tetragon events per type that are failed to sent from the kernel.
# TYPE tetragon_missed_events_total counter
tetragon_missed_events_total{msg_op="13"} 73300
tetragon_missed_events_total{msg_op="23"} 28
tetragon_missed_events_total{msg_op="24"} 606
tetragon_missed_events_total{msg_op="5"} 20
tetragon_missed_events_total{msg_op="7"} 22
# HELP tetragon_msg_op_total The total number of times we encounter a given message opcode. For internal use only.
# TYPE tetragon_msg_op_total counter
tetragon_msg_op_total{msg_op="13"} 4.268532e+06
tetragon_msg_op_total{msg_op="23"} 12444
tetragon_msg_op_total{msg_op="24"} 2110
tetragon_msg_op_total{msg_op="5"} 11908
tetragon_msg_op_total{msg_op="7"} 12447
# HELP tetragon_ringbuf_perf_event_lost_total The total number of Tetragon ringbuf perf events lost.
# TYPE tetragon_ringbuf_perf_event_lost_total counter
tetragon_ringbuf_perf_event_lost_total 73976
# HELP tetragon_ringbuf_queue_lost_total The total number of Tetragon events ring buffer queue lost.
# TYPE tetragon_ringbuf_queue_lost_total counter
tetragon_ringbuf_queue_lost_total 0
# HELP tetragon_ringbuf_queue_received_total The total number of Tetragon events ring buffer queue received.
# TYPE tetragon_ringbuf_queue_received_total counter
tetragon_ringbuf_queue_received_total 4.307441e+06
```

This PR adds an eBPF map collector for getting metrics directly from a map. This map contains information about the return values of all `perf_event_output` calls (i.e. if it fails). This provides us the ability to determine missed events per type. Metric `tetragon_missed_events_total` contains such information.

Using the previous example, we can see that we lost 73976 events from the user-space (`tetragon_ringbuf_perf_event_lost_total`). This is the same as the sum of all `tetragon_missed_events_total` metrics gathered from the kernel.

### DIFF
--- a/bpf/cgroup/bpf_cgroup_events.h
+++ b/bpf/cgroup/bpf_cgroup_events.h
@@ -51,7 +51,7 @@ send_cgrp_event(struct bpf_raw_tracepoint_args *ctx,
 	memcpy(&msg->cgrp_data.name, &cgrp_track->name, KN_NAME_LENGTH);
 	probe_read_str(&msg->path, PATH_MAP_SIZE - 1, path);
 
-	perf_event_output(ctx, &tcpmon_map, BPF_F_CURRENT_CPU, msg, size);
+	perf_event_output_metric(ctx, MSG_OP_CGROUP, &tcpmon_map, BPF_F_CURRENT_CPU, msg, size);
 
 	return 0;
 }
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -535,4 +535,30 @@ execve_joined_info_map_get(__u64 tid)
 
 _Static_assert(sizeof(struct execve_map_value) % 8 == 0,
 	       "struct execve_map_value should have size multiple of 8 bytes");
+
+struct kernel_stats {
+	__u64 sent_failed[256];
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__type(key, __u32);
+	__type(value, struct kernel_stats);
+	__uint(max_entries, 1);
+} tg_stats_map SEC(".maps");
+
+static inline __attribute__((always_inline)) void perf_event_output_metric(void *ctx, u8 metric, void *map, u64 flags, void *data, u64 size)
+{
+	struct kernel_stats *valp;
+	__u32 zero = 0;
+	long err;
+
+	err = perf_event_output(ctx, map, flags, data, size);
+	if (err < 0) {
+		valp = map_lookup_elem(&tg_stats_map, &zero);
+		if (valp)
+			__sync_fetch_and_add(&valp->sent_failed[metric], 1);
+	}
+}
+
 #endif //_PROCESS__
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -2183,7 +2183,7 @@ generic_actions(void *ctx, struct bpf_map_def *heap,
 }
 
 static inline __attribute__((always_inline)) long
-generic_output(void *ctx, struct bpf_map_def *heap)
+generic_output(void *ctx, struct bpf_map_def *heap, u8 op)
 {
 	struct msg_generic_kprobe *e;
 	int zero = 0;
@@ -2226,7 +2226,7 @@ generic_output(void *ctx, struct bpf_map_def *heap)
 		     :
 		     : [total] "+r"(total)
 		     :);
-	perf_event_output(ctx, &tcpmon_map, BPF_F_CURRENT_CPU, e, total);
+	perf_event_output_metric(ctx, op, &tcpmon_map, BPF_F_CURRENT_CPU, e, total);
 	return 1;
 }
 
--- a/pkg/metrics/eventmetrics/collector.go
+++ b/pkg/metrics/eventmetrics/collector.go
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package eventmetrics
+
+import (
+	"path/filepath"
+	"strconv"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/tetragon/pkg/api/processapi"
+	"github.com/cilium/tetragon/pkg/option"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// bpfCollector implements prometheus.Collector. It collects metrics directly from BPF maps.
+type bpfCollector struct{}
+
+func NewBPFCollector() prometheus.Collector {
+	return &bpfCollector{}
+}
+
+func (c *bpfCollector) Describe(ch chan<- *prometheus.Desc) {
+	ch <- MissedEvents.Desc()
+}
+
+func (c *bpfCollector) Collect(ch chan<- prometheus.Metric) {
+	mapHandle, err := ebpf.LoadPinnedMap(filepath.Join(option.Config.MapDir, "tg_stats_map"), nil)
+	if err != nil {
+		return
+	}
+	defer mapHandle.Close()
+
+	var zero uint32
+	var allCpuValue []processapi.KernelStats
+	if err := mapHandle.Lookup(zero, &allCpuValue); err != nil {
+		return
+	}
+
+	sum := processapi.KernelStats{}
+	for _, val := range allCpuValue {
+		for i, data := range val.SentFailed {
+			sum.SentFailed[i] += data
+		}
+	}
+
+	for i, data := range sum.SentFailed {
+		if data > 0 {
+			ch <- MissedEvents.MustMetric(float64(data), strconv.Itoa(i))
+		}
+	}
+}
--- a/pkg/metrics/metricsconfig/initmetrics.go
+++ b/pkg/metrics/metricsconfig/initmetrics.go
@@ -50,6 +50,7 @@ func InitAllMetrics(registry *prometheus.Registry) {
 		observer.NewBPFCollector(),
 		process.NewBPFCollector(),
 	))
+	registry.MustRegister(eventmetrics.NewBPFCollector())
 
 	// register common third-party collectors
 	registry.MustRegister(collectors.NewGoCollector())
--- a/bpf/alignchecker/bpf_alignchecker.c
+++ b/bpf/alignchecker/bpf_alignchecker.c
@@ -23,3 +23,4 @@ struct execve_map_value _execve_map_value;
 struct event_config _event_config;
 struct tetragon_conf _tetragon_conf;
 struct cgroup_tracking_value _cgroup_tracking_value;
+struct kernel_stats _kernel_stats;
--- a/bpf/lib/msg_types.h
+++ b/bpf/lib/msg_types.h
@@ -29,6 +29,8 @@ enum msg_ops {
 
 	MSG_OP_CGROUP = 25,
 
+	MSG_OP_LOADER = 26,
+
 	MSG_OP_MAX,
 };
 
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -296,6 +296,6 @@ execve_send(struct sched_execve_args *ctx)
 		sizeof(struct msg_capabilities) +
 		sizeof(struct msg_cred_minimal) + sizeof(struct msg_ns) +
 		sizeof(struct msg_execve_key) + p->size);
-	perf_event_output(ctx, &tcpmon_map, BPF_F_CURRENT_CPU, event, size);
+	perf_event_output_metric(ctx, MSG_OP_EXECVE, &tcpmon_map, BPF_F_CURRENT_CPU, event, size);
 	return 0;
 }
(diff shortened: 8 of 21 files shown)


pkg/metrics/metricsconfig/initmetrics.go
```diff
@@ -49,6 +50,7 @@ func InitAllMetrics(registry *prometheus.Registry) {
 		eventcache.NewBPFCollector(),
 		observer.NewBPFCollector(),
 		process.NewBPFCollector(),
+		bpfmetrics.NewBPFCollector(),
```
lambdanis, line 53: This collector can be registered separately, like that:
```
registry.MustRegister(bpfmetrics.NewBPFCollector())
```

The map metrics collectors must be wrapped because they share metric descriptors (see [this comment](https://github.com/cilium/tetragon/blob/18d1b4ec2498c68a6d704dba0c1abca0472ccbe7/pkg/metrics/mapmetrics/mapmetrics.go#L41-L44)). But this new collector doesn't share metric descriptors with any other one.

pkg/bpfmetrics/metrics.go
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+package bpfmetrics
```
lambdanis, line 3: I would make it a part of `pkg/metrics/eventmetrics`, is there a reason for having it as a separate package?

pkg/metrics/metricsconfig/initmetrics.go
tpapagian: Ok, thanks. I will register that separately.

pkg/bpfmetrics/metrics.go
tpapagian: No specific reason. I will move that to `pkg/metrics/eventmetrics`.

bpf/lib/process.h
```diff
@@ -535,4 +535,42 @@ execve_joined_info_map_get(__u64 tid)
...
+
+#define METRIC_EXEC	  0
+#define METRIC_CLONE	  1
+#define METRIC_EXIT	  2
+#define METRIC_DATA	  3
+#define METRIC_CGROUP	  4
+#define METRIC_LOADER	  5
+#define METRIC_TRACEPOINT 6
+#define METRIC_KPROBE	  7
+#define METRIC_UPORBE	  8
```
lambdanis, line 547:
```suggestion
#define METRIC_UPROBE	  8
```
typo

bpf/process/types/basic.h
```diff
@@ -2226,7 +2227,18 @@ generic_output(void *ctx, struct bpf_map_def *heap)
...
+	ret = perf_event_output(ctx, &tcpmon_map, BPF_F_CURRENT_CPU, e, total);
+	switch (e->common.op) {
+	case MSG_OP_GENERIC_TRACEPOINT:
+		PERF_METRICS(METRIC_TRACEPOINT, ret);
+		break;
+	case MSG_OP_GENERIC_KPROBE:
+		PERF_METRICS(METRIC_KPROBE, ret);
+		break;
+	case MSG_OP_GENERIC_UPROBE:
+		PERF_METRICS(METRIC_UPORBE, ret);
```
lambdanis, line 2239:
```suggestion
		PERF_METRICS(METRIC_UPROBE, ret);
```
typo

pkg/metrics/eventmetrics/collector.go
```diff
@@ -0,0 +1,83 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+package eventmetrics
```
lambdanis, line 3: nit: Add a newline above the package statement, otherwise the SPDX header becomes the package documentation.

pkg/metrics/eventmetrics/collector.go
```diff
@@ -0,0 +1,83 @@
...
+		ch <- MissedEvents.MustMetric(float64(totalSent), "total_sent")
+	}
+
+	var totalSentFailed uint64
+	for i := 0; i < processapi.MetricMaxValues; i++ {
+		totalSentFailed += sum.SentFailed[i]
+	}
+	if totalSentFailed > 0 {
+		ch <- MissedEvents.MustMetric(float64(totalSentFailed), "total_failed")
+	}
```
lambdanis, line 82: These sums can be easily calculated in PromQL:
```
sum(tetragon_missed_events_total{event=~".*_failed"})
```
so there is no need to expose them as separate metrics. Having these extra "total" timeseries is rather confusing.

pkg/metrics/eventmetrics/collector.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, val := range allCpuValue {
+		for i := 0; i < processapi.MetricMaxValues; i++ {
+			sum.Sent[i] += val.Sent[i]
+			sum.SentFailed[i] += val.SentFailed[i]
+		}
+	}
+
+	for i := 0; i < processapi.MetricMaxValues; i++ {
+		if sum.Sent[i] > 0 {
+			ch <- MissedEvents.MustMetric(float64(sum.Sent[i]), fmt.Sprintf("%s_sent", metricsLabels[i]))
```
lambdanis, line 61: If I understand correctly, "sent" events are not "missed", so it's confusing that they're reported by `MissedEvents`. I would rather define another metric, `SentEvents` (`tetragon_sent_events_total`) for those. Then we can label metrics by the event type, without `_sent` and `_failed` suffixes.

lambdanis (changes requested):
I left some comments :) The most important one being that it looks like we want to expose two different metric names here.

tpapagian:
> I left some comments :) The most important one being that it looks like we want to expose two different metric names here.

Thanks for the review! I have made the changes that you proposed.

lambdanis (approved):
looks good to me

bpf/cgroup/bpf_cgroup_events.h
```diff
@@ -51,7 +51,7 @@ send_cgrp_event(struct bpf_raw_tracepoint_args *ctx,
 	memcpy(&msg->cgrp_data.name, &cgrp_track->name, KN_NAME_LENGTH);
 	probe_read_str(&msg->path, PATH_MAP_SIZE - 1, path);
 
-	perf_event_output(ctx, &tcpmon_map, BPF_F_CURRENT_CPU, msg, size);
+	PERF_METRICS(METRIC_CGROUP, perf_event_output(ctx, &tcpmon_map, BPF_F_CURRENT_CPU, msg, size));
```
kkourt, line 54: Is there a reason to use a macro here? can't we just use a (inlined) function?

E.g.,
```
perf_event_output_metric(ctx, METRIC_CGROUP, &tcpmmon_map, ...)
```

bpf/lib/process.h
```diff
@@ -535,4 +535,42 @@ execve_joined_info_map_get(__u64 tid)
...
+
+#define PERF_METRICS(__x, __v)                                                      \
+	do {                                                                        \
+		long retval = (__v);                                                \
+		long index = (__x);                                                 \
+		__u32 key = 0;                                                      \
+		struct kernel_stats *valp = map_lookup_elem(&tg_stats_map, &key);   \
+		if (valp) {                                                         \
+			if (retval >= 0)                                            \
+				__sync_fetch_and_add(&valp->sent[index], 1);        \
```
kkourt, line 570: Hm... do we need this? It adds overhead in the good path and we count the events we receive by type in user-space.

bpf/process/types/basic.h
```diff
@@ -2226,7 +2227,18 @@ generic_output(void *ctx, struct bpf_map_def *heap)
...
+	case MSG_OP_GENERIC_TRACEPOINT:
+		PERF_METRICS(METRIC_TRACEPOINT, ret);
+		break;
+	case MSG_OP_GENERIC_KPROBE:
+		PERF_METRICS(METRIC_KPROBE, ret);
+		break;
+	case MSG_OP_GENERIC_UPROBE:
+		PERF_METRICS(METRIC_UPROBE, ret);
+		break;
+	}
```
kkourt, line 2241: maybe 

```
switch () {
     case TRACEPOINT:
       metric = METRIC_TRACEPOINT
     case KPROBE
       metric METRIC_KPROBE
}
// send event?
```

Reason is that I don't want to rely on compiler optimization to add a single call to the function.

kkourt (commented):
Nice! I think this will help with debugging.
I have some small comments, PTAL.

bpf/lib/process.h
tpapagian: I have seen cases where the sum of all `tetragon_missed_events_total` is not equal to `tetragon_ringbuf_perf_event_lost_total` under very heavy load. That meant to me that there were some events that we successfully sent from the kernel and somehow we missed getting into the user space. 

I have tried to reproduce that now with the latest version but I didn't manage to do so. So let's remove the `tetragon_sent_events_total` metric entirely for now and we can reconsider that in the future if we see again similar cases. 

Regarding the overhead, I don't believe that this is a big deal as we use a per-cpu array.

kkourt: > I have seen cases where the sum of all tetragon_missed_events_total is not equal to tetragon_ringbuf_perf_event_lost_total under very heavy load. That meant to me that there were some events that we successfully sent from the kernel and somehow we missed getting into the user space

OK, that's good enough reason to keep it for me.
If we use a per-cpu array, do we need `__sync_fetch_and_add`?

tpapagian: > OK, that's good enough reason to keep it for me.

I will try to do that again. If I don't manage to do that I will remove that for now (I am just thinking if there was any bugs in the previous version and thus I have this behaviour). If I don't manage to reproduce, I will remove that for now. 

> If we use a per-cpu array, do we need __sync_fetch_and_add?

Hm, I would say no but let's be in the safe side. I will try to experiment with that as well a bit.

bpf/cgroup/bpf_cgroup_events.h
tpapagian: Fixed

bpf/process/types/basic.h
tpapagian: Fixed

bpf/lib/process.h
```diff
@@ -535,4 +535,39 @@ execve_joined_info_map_get(__u64 tid)
...
+#define METRIC_CLONE	  1
+#define METRIC_EXIT	  2
+#define METRIC_DATA	  3
+#define METRIC_CGROUP	  4
+#define METRIC_LOADER	  5
+#define METRIC_TRACEPOINT 6
+#define METRIC_KPROBE	  7
+#define METRIC_UPROBE	  8
+#define METRIC_MAX_VALUES 9
+
```
olsajiri, line 549: just a thought.. in my initial attempt I used MSG_OP_* values to identify metrics,
I wonder it's better than add new cathegories? also  might be  easier then in
output when comparing to tetragon_msg_op_total values? I assume MSG_OP_*
link to sensors directly anyway

then we might need to use BPF_MAP_TYPE_PERCPU_HASH because the calues are
scattered

kkourt: fwiw, I think using MSG_OP types makes sense.

> then we might need to use BPF_MAP_TYPE_PERCPU_HASH because the values are
scattered

They are not that scattered so I think it's fine to use an array to avoid the hash lookup.

tpapagian: Make sense to me as well. The values that we care are:
```
  MSG_OP_EXECVE = 5,
  MSG_OP_EXIT = 7,
  MSG_OP_GENERIC_KPROBE = 13,
  MSG_OP_GENERIC_TRACEPOINT = 14,
  MSG_OP_GENERIC_UPROBE = 15,
  MSG_OP_CLONE = 23,
  MSG_OP_DATA = 24,
  MSG_OP_CGROUP = 25,
```
So an array of 26 elements seems to be enough.

bpf/lib/process.h
tpapagian: After some experiments:
1. Let's remove the metric for successful `perf_event_output` calls. Maybe I have seen a bug in a previous version. Even with a very loaded host, I cannot reproduce that. We can always add that again if we see any numbers that do not make sense.
2. I don't really believe that `__sync_fetch_and_add` is necessary, but I would say let's keep it. It is called only when `perf_event_output` fails which is a rare event. Thus any performance overhead is minimal here. 

Does this make sense?

kkourt: Sounds good to me, thanks!

bpf/lib/process.h
kkourt: Let's go crazy and add 256.

tpapagian: Updated the PR based on that comment.

olsajiri (approved):

kkourt (approved):

kkourt:
Thanks! I think this will be very useful to have, which is why I added a backport label for 1.0.

tpapagian:
Backport PR: https://github.com/cilium/tetragon/pull/1702

Issue #1667, opened 2023-10-29
Tetragon metrics-label-filter Flags can't set metric-label-filter as an argument


Jack-R-lantern:
### What happened?

```
tetragon --help
```

```
Tetragon - eBPF-based Security Observability and Runtime Enforcement

Usage:
  tetragon [flags]

Flags:
      --bpf-lib string                            Location of Tetragon libs (btf and bpf files) (default "/var/lib/tetragon/")
      --btf string                                Location of btf
      --config-dir string                         Configuration directory that contains a file for each option
      --data-cache-size int                       Size of the data events cache (default 1024)
  -d, --debug                                     Enable debug messages. Equivalent to '--log-level=debug'
      --disable-kprobe-multi                      Allow to disable kprobe multi interface
      --enable-export-aggregation                 Enable JSON export aggregation
      --enable-k8s-api                            Access Kubernetes API to associate Tetragon events with Kubernetes pods
      --enable-msg-handling-latency               Enable metrics for message handling latency
      --enable-pid-set-filter                     Enable pidSet export filters. Not recommended for production use
      --enable-pod-info                           Enable PodInfo custom resource
      --enable-policy-filter                      Enable policy filter code (beta)
      --enable-policy-filter-debug                Enable policy filter debug messages
      --enable-process-ancestors                  Include ancestors in process exec events (default true)
      --enable-process-cred                       Enable process_cred events
      --enable-process-ns                         Enable namespace information in process_exec and process_kprobe events
      --event-queue-size uint                     Set the size of the internal event queue. (default 10000)
      --export-aggregation-buffer-size uint       Aggregator channel buffer size (default 10000)
      --export-aggregation-window-size duration   JSON export aggregation time window (default 15s)
      --export-allowlist string                   JSON export allowlist
      --export-denylist string                    JSON export denylist
      --export-file-compress                      Compress rotated JSON export files
      --export-file-max-backups int               Number of rotated JSON export files to retain (default 5)
      --export-file-max-size-mb int               Size in MB for rotating JSON export files (default 10)
      --export-file-perm string                   Access permissions on JSON export files (default "600")
      --export-file-rotation-interval duration    Interval at which to rotate JSON export files in addition to rotating them by size
      --export-filename string                    Filename for JSON export. Disabled by default
      --export-rate-limit int                     Rate limit (per minute) for event export. Set to -1 to disable (default -1)
      --expose-kernel-addresses                   Expose real kernel addresses in events stack traces
[...]

PR #1678, opened 2023-10-30
pkg/option: add metrics-label-filter flags

Jack-R-lantern:
The commit is to add the missing metrics-label-filter flag.

fix: #1667

### DIFF
--- a/pkg/option/flags.go
+++ b/pkg/option/flags.go
@@ -205,6 +205,7 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.String(KeyK8sKubeConfigPath, "", "Absolute path of the kubernetes kubeconfig file")
 	flags.Bool(KeyEnableProcessAncestors, true, "Include ancestors in process exec events")
 	flags.String(KeyMetricsServer, "", "Metrics server address (e.g. ':2112'). Disabled by default")
+	flags.String(KeyMetricsLabelFilter, "", "Comma-separated list of enabled metric labels. (e.g. \"namespace,workload,pod,binary\") By default all labels are enabled.")
 	flags.String(KeyServerAddress, "localhost:54321", "gRPC server address (e.g. 'localhost:54321' or 'unix:///var/run/tetragon/tetragon.sock'")
 	flags.String(KeyGopsAddr, "", "gops server address (e.g. 'localhost:8118'). Disabled by default")
 	flags.Bool(KeyEnableProcessCred, false, "Enable process_cred events")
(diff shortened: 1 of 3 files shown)


pkg/option/flags.go
```diff
@@ -205,6 +205,7 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.String(KeyK8sKubeConfigPath, "", "Absolute path of the kubernetes kubeconfig file")
 	flags.Bool(KeyEnableProcessAncestors, true, "Include ancestors in process exec events")
 	flags.String(KeyMetricsServer, "", "Metrics server address (e.g. ':2112'). Disabled by default")
+	flags.String(KeyMetricsLabelFilter, "", "Metrics Label Filter")
```
lambdanis, line 208: It's unclear from the help text what the user should pass, could we be more descriptive here? For example:
```
Comma-separated list of enabled metric labels. By default all labels are enabled.
```

Jack-R-lantern: ok i will change it

Jack-R-lantern:
filter-search
@lambdanis 
Checking the documentation, I can't seem to find any information related to filter.
Will it be added later?

pkg/option/flags.go
```diff
@@ -205,6 +205,7 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.String(KeyK8sKubeConfigPath, "", "Absolute path of the kubernetes kubeconfig file")
 	flags.Bool(KeyEnableProcessAncestors, true, "Include ancestors in process exec events")
 	flags.String(KeyMetricsServer, "", "Metrics server address (e.g. ':2112'). Disabled by default")
+	flags.String(KeyMetricsLabelFilter, "", "Metrics Label Filter Comma-separated list of enabled metric labels. By default all labels are enabled.")
```
lambdanis, line 208:
```suggestion
	flags.String(KeyMetricsLabelFilter, "", "Comma-separated list of enabled metric labels. By default all labels are enabled.")
```
No need to repeat the flag name in the help text.

lambdanis (changes requested):

lambdanis:
> Checking the documentation, I can't seem to find any information related to filter.
> Will it be added later?

@Jack-R-lantern I'm afraid the search in the docs might not work properly. This option is documented in the Helm chart docs, but only there. Do you have suggestions on how to document it better?

Jack-R-lantern:
> > Checking the documentation, I can't seem to find any information related to filter.
> > Will it be added later?
> 
> @Jack-R-lantern I'm afraid the search in the docs might not work properly. This option is documented in the Helm chart docs, but only there. Do you have suggestions on how to document it better?

I think the values.yaml file in the helm chart is sufficient, too
Even for 'standalone', I think "--help" is sufficient.

lambdanis (approved):

mtardy (approved):
If Anna approved this let's merge!

PR #1682, opened 2023-10-31
ci: replace buildjet arm64 runners with actuated ones

mtardy:
The CNCF signed a contract with Actuated to provide free arm64 runners
to CNCF project: see https://actuated.dev/blog/arm-ci-cncf-ampere.

After some trial and error we managed to compile a kernel for the
runners using Firecracker that works with BPF and especially Tetragon.
Only limitation is that they are running 5.10 for now and there is no
nested virtualization. But good news is that we can get more CPU and
more RAM for free!

Note: I know that I didn't change the buildjet runner for `Packages e2e Tests / standalone-tests (buildjet-2vcpu-ubuntu-2204-arm, arm64, arm64)`, but this should be eventually removed as `make tarball` now work with cross-compilation and we should not need a native runner here.

### DIFF
--- a/.github/workflows/gotests.yml
+++ b/.github/workflows/gotests.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-20.04, buildjet-2vcpu-ubuntu-2204-arm ]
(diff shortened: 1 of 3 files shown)


alexellis:
Hi @mtardy could you retrigger the build please? BTF has been deployed into the Kernel

lambdanis (approved):

PR #1685, opened 2023-10-31
tetragon: factor kprobes

olsajiri:
Due to a popular demand, trying to make the kprobe sensor config code more structured.

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -482,10 +482,6 @@ type addKprobeIn struct {
 	customHandler eventhandler.Handler
 }
 
-type addKprobeOut struct {
-	id idtable.EntryID
-}
-
 func getKprobeSymbols(symbol string, syscall bool, lists []v1alpha1.ListSpec) ([]string, bool, error) {
 	if strings.HasPrefix(symbol, "list:") {
 		name := symbol[len("list:"):]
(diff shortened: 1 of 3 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -540,17 +540,17 @@ func createGenericKprobeSensor(
 		selMaps:       selMaps,
 	}
 
-	for i := range kprobes {
-		syms, syscall, err := getKprobeSymbols(kprobes[i].Call, kprobes[i].Syscall, lists)
+	for _, kprobe := range kprobes {
```
kkourt, line 543: Is this correct?

`kprobes` is `[]KprobeSpec` so not a pointer, so kprobe will be a copy and when we update it, the original array element will not change. Not sure if this is an issue since none of the tests are failing, but it is different from the original behavior where we use `kprobe[i]` (e.g., `kprobe[i].Syscall = syscall` actually modifies the array element).

Maybe do:
```
for i := range kprobes {
       kprobe = kprobes[i]
}
```

To be on the safe side?

olsajiri: yes, but that kprobe is used only in that loop, so I think it's ok.. unless we use that kprobe def elsewhere.. in which case it's safer to keep it as it was.. ;-) I'll remove that commit to be on safe side as you said

kkourt: Ah, OK.

BTW, I meant doing `kprobe = &kprobes[i]`, but does not matter :)

kkourt (approved):
Thanks!

PR #1689, opened 2023-10-31
gh: use cosign sign -y

kkourt:


kkourt:
This is expected to fail, but it seems that the test on the branch succeeded: https://github.com/cilium/tetragon/actions/runs/6707841804/job/18227300170.

PR #1700, opened 2023-11-01
server: fix another segfault related to field filters

willfindlay:
Commit 1 fixes the bug, commit 2 adds a nil check and a metric to catch if a similar situation happens in the future.

```release-note
Fix a segmentation fault related to filtering out pid information with field filters
```

### DIFF
--- a/pkg/filters/fields.go
+++ b/pkg/filters/fields.go
@@ -81,15 +81,48 @@ type FieldFilter struct {
 	fields         fmutils.NestedMask
 	action         tetragon.FieldFilterAction
 	invertEventSet bool
+	needsCopy      map[tetragon.EventType]struct{}
 }
 
 // NewFieldFilter constructs a new FieldFilter from a set of fields.
 func NewFieldFilter(eventSet []tetragon.EventType, fields []string, action tetragon.FieldFilterAction, invertEventSet bool) *FieldFilter {
+	// We only need to copy exec and exit events when they are explicitly filtering out
+	// the PID. This is because we need the PID to not be nil when accessing the event
+	// later on from the eventcache. See additional comments below for details.
+	var maybeNeedsCopy bool
+	if action == tetragon.FieldFilterAction_EXCLUDE {
+		for _, field := range fields {
+			if strings.HasPrefix(field, "process") {
+				maybeNeedsCopy = true
+				break
+			}
+		}
+	} else if action == tetragon.FieldFilterAction_INCLUDE {
+		// For inclusion, it's the opposite situation from the above. If the process.pid
+		// field is NOT present, it will be trimmed. So assume we need a copy unless we
+		// see process.pid.
+		maybeNeedsCopy = true
+		for _, field := range fields {
+			if field == "process.pid" {
+				maybeNeedsCopy = false
+				break
+			}
+		}
+	}
+
+	needsCopy := make(map[tetragon.EventType]struct{})
+	if maybeNeedsCopy {
+		for _, t := range eventSet {
+			needsCopy[t] = struct{}{}
+		}
+	}
+
 	return &FieldFilter{
 		eventSet:       eventSet,
 		fields:         fmutils.NestedMaskFromPaths(fields),
 		action:         action,
 		invertEventSet: invertEventSet,
+		needsCopy:      needsCopy,
 	}
 }
 
@@ -105,12 +138,10 @@ func NewExcludeFieldFilter(eventSet []tetragon.EventType, fields []string, inver
 
 // FieldFilterFromProto constructs a new FieldFilter from a Tetragon API field filter.
 func FieldFilterFromProto(filter *tetragon.FieldFilter) *FieldFilter {
-	var fields fmutils.NestedMask
+	var fields []string
 
 	if filter.Fields != nil {
-		fields = fmutils.NestedMaskFromPaths(filter.Fields.Paths)
-	} else {
-		fields = make(fmutils.NestedMask)
+		fields = filter.Fields.Paths
 	}
 
 	invert := false
@@ -118,12 +149,7 @@ func FieldFilterFromProto(filter *tetragon.FieldFilter) *FieldFilter {
 		invert = filter.InvertEventSet.Value
 	}
 
-	return &FieldFilter{
-		eventSet:       filter.EventSet,
-		fields:         fields,
-		action:         filter.Action,
-		invertEventSet: invert,
-	}
+	return NewFieldFilter(filter.EventSet, fields, filter.Action, invert)
 }
 
 // FieldFiltersFromGetEventsRequest returns a list of EventFieldFilter for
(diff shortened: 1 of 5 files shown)


pkg/server/server.go
```diff
@@ -166,14 +166,14 @@ func (s *Server) GetEventsWG(request *tetragon.GetEventsRequest, server tetragon
...
-			}
+
+			var copied bool
 			for _, filter := range filters {
-				// we need not to change res
-				// maybe only for exec events
+				// We need a copy of the exec or exit event as modifing the original message
+				// can cause issues in the process cache (we keep a copy of that message there).
+				if filter.NeedsCopy(filterEvent) && !copied {
+					filterEvent = proto.Clone(event).(*tetragon.GetEventsResponse)
```
kkourt, line 175: copied = `true`?

Or even better (IMO):
```
if filterEvent == event && filter.NeedsCopy(filterEvent) {
```

pkg/filters/fields.go
```diff
@@ -81,15 +81,51 @@ type FieldFilter struct {
...
 	invertEventSet bool
+	needsCopy      map[tetragon.EventType]struct{}
 }
 
 // NewFieldFilter constructs a new FieldFilter from a set of fields.
 func NewFieldFilter(eventSet []tetragon.EventType, fields []string, action tetragon.FieldFilterAction, invertEventSet bool) *FieldFilter {
+	var maybeNeedsCopy bool
+	if action == tetragon.FieldFilterAction_EXCLUDE {
+		for _, field := range fields {
+			if field == "process.pid" || field == "parent.pid" {
```
kkourt, line 92: Can you explain this check? Why is it only "process.pid" that we care about? Shouldn't be the same with all the fields under process or parent? I would expect something like strings.HasPrefix(field, "process.") || strings.HasPrefix(field,  "parent.").

pkg/filters/fields.go
```diff
@@ -81,15 +81,51 @@ type FieldFilter struct {
...
+				maybeNeedsCopy = true
+				break
+			}
+		}
+	} else if action == tetragon.FieldFilterAction_INCLUDE {
```
kkourt, line 97: Can you explain the logic behind the INCLUDE?

I understand the EXCLUDE as follows: We need to copy if we need to remove fields from (and thus modify)  process or parent. Why do we need a copy on include? Is the idea that fields would have been excluded at the top level so unless we have both, something will be deleted?

Again, I'm not sure why we are only checking for .pid (see above). Is this the only field that can be included/excluded from .process and .parent?

pkg/filters/fields.go
```diff
@@ -81,15 +81,51 @@ type FieldFilter struct {
...
+			}
+		}
+		if hasPid && hasPpid {
+			maybeNeedsCopy = false
+		}
+	}
+
+	needsCopy := make(map[tetragon.EventType]struct{})
+	for _, t := range eventSet {
+		if t == tetragon.EventType_PROCESS_EXEC && maybeNeedsCopy {
```
kkourt, line 115: Why only check for exec and exit events? Wouldn't similar filters on all events have the same effect?

kkourt (commented):
Thanks! 

pkg/server/server.go
willfindlay: Whoops :)

pkg/filters/fields.go
willfindlay: The bug is specifically related to missing pids causing segfaults in the eventcache. If these fields aren't filtered, we don't hit the bug. Maybe it is safer to do just process and parent in general though.

pkg/filters/fields.go
willfindlay: @kkourt think of include like the inverse of exclude. If the fields we care about are NOT included they will be excluded. We're checking for that here.

pkg/filters/fields.go
willfindlay: Nope, the segfault is in the grpc/exec package which only deals with exec and exit events.

willfindlay:
Thanks @kkourt I'll fix and simplify some of the things you pointed out soon

pkg/filters/fields.go
kkourt: OK, but there might other similar bugs. Is there an argument we can make that the same will not happen for other fields in process?

My understanding of the issue is that the fieldfilter will modify .process and .parent of the event which are shared among different processes. So a fieldfilter applied to one event, might modify .process in a way that affects another event. Based on above reasoning, if a fieldFilter, for example, deletes process.pod for one event, it will also affect other types of events. Maybe I'm not understanding the issue correctly, but I'd like us to have an argument that something similar cannot happen for other fields under .process (and similarly .parent).

Let me also CC @tpapagian in case he has any thoughts.

willfindlay:
Marking as draft again while I grab another test image.

willfindlay:
The above will be squashed once I verify the changes ^

pkg/filters/fields.go
willfindlay: Making a copy of every event will absolutely kill performance but yeah maybe it's the way to go. I think to be fully confident we'd need a full account of where exactly fields from events are used in the eventcache.

willfindlay: The other option is refactoring so that this doesn't happen, but that is likely multiple weeks of work.

kkourt: > Making a copy of every event will absolutely kill performance but yeah maybe it's the way to go. I think to be fully confident we'd need a full account of where exactly fields from events are used in the eventcache.

I would argue that correctness is more important than performance.
 Also, this overhead would only be induced when  field filters are used, which I think is a reasonable tradeoff.

pkg/filters/fields.go
willfindlay: Actually you're right. Managed to get a different bug with another fieldfilter config. I've updated accordingly.

pkg/filters/fields.go
```diff
@@ -81,15 +81,48 @@ type FieldFilter struct {
...
+				break
+			}
+		}
+	} else if action == tetragon.FieldFilterAction_INCLUDE {
+		// For inclusion, it's the opposite situation from the above. If the process.pid
+		// field is NOT present, it will be trimmed. So assume we need a copy unless we
+		// see process.pid.
+		maybeNeedsCopy = true
+		for _, field := range fields {
+			if field == "process.pid" {
```
kkourt, line 106: Does this also need HasPrefix()?

kkourt (approved):
Thanks!

pkg/filters/fields.go
willfindlay: It has to be an exact match so that it's more conservative. If we used HasPrefix and you did something like `process.binary` then we would not do a copy but PID would still be missing.

willfindlay: Hopefully that makes sense

PR #1708, opened 2023-11-03
Cleanup execve_map entry on clone failure 

tpapagian:
Clone events are handled in `event_wake_up_new_task`. In the case, if `__event_find_parent` fails, we do not send an event to the user. On the other hand, `execve_map_get` has already added a new entry in `execve_map`. This causes an inconsistency between `execve_map` and the user-space process cache.

In order to solve this, we have to delete the newly added entry in the `execve_map` before returning.


olsajiri (approved):
nice catch!

bpf/process/bpf_fork.c
```diff
@@ -21,13 +21,23 @@ __attribute__((section("kprobe/wake_up_new_task"), used)) int
...
 		return 0;
 
 	tgid = BPF_CORE_READ(task, tgid);
 
+	/* Do not try to create any msg or calling execve_map_get
+	 * (that will add a new process in the execve_map) if we
+	 * cannot find it's parent in the execve_map.
+	 */
+	parent = __event_find_parent(task);
+	if (!parent)
```
kkourt, line 43: Would be nice to have a counter for this! I'm curious if  it will ever happen, when the second patch is also applied. 
That being said,  since we are not sending an event if we don't find the parent, there is no point in adding an entry to the map. So this patch makes sense to me as is.

kkourt (approved):
Thanks!

bpf/process/bpf_fork.c
tpapagian: Yes, I have in mind to add a counter for that (and other cases in exec/exit events) to check if this happens in a follow-up PR.

PR #1730, opened 2023-11-08
Makefile: also use JOBS for test-compile

kkourt:


willfindlay (approved):

lmb (commented):
Thanks!

Makefile
```diff
@@ -257,7 +257,7 @@ test-compile:
 			continue; \
 		fi; \
 		echo -c ./$$localpkg -o go-tests/$$localtestfile; \
-	done | xargs -P $$(nproc) -L 1 $(GO) test -gcflags=$(GO_BUILD_GCFLAGS)
+	done | xargs -P $(JOBS) -L 1 $(GO) test -gcflags=$(GO_BUILD_GCFLAGS)
```
lmb, line 260: Actually I think that needs `go test -p 1` or something like that, since it parallelises by default. So right now it does nproc*nproc compilations.

kkourt: Good point, added a `GOMAXPROCS=1`. Thanks!

PR #1731, opened 2023-11-08
Rework the matchBinaries selector implementation

mtardy:


bpf/process/bpf_fork.c
```diff
@@ -44,7 +44,6 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
 		curr->key.pid = tgid;
 		curr->key.ktime = ktime_get_ns();
 		curr->nspid = get_task_pid_vnr();
-		curr->binary = parent->binary;
```
mtardy, line 47: Is this okay @tpapagian? Naively, I replaced that line with `curr->bin = parent->bin` and it broke my tests, I'm not exactly sure why. I guess maybe we should copy the complete struct with the binary path inside of it?

tpapagian: I believe that we need to setup `bin` here (maybe with a `memcpy`?). An example that I am thinking of is having a program that does a `fork` but not an `exec`. 

I have tried a program like [this](https://gist.github.com/tpapagian/19d708e9319020854d76289809928bc0) that does a `fork` (and not an `exec`) and the child process reads `/etc/passwd`. I used [this](https://gist.github.com/tpapagian/70f0b15665b00e446d0a3f37b32ddf98) tracing policy. I expect that adding a `matchBinaries` selector in that tracing policy not to work because the child will not have the proper binary name. 

My understanding is that `bin` is used only for `matchBinaries`, as without that I got a correct event similar to [this](https://gist.github.com/tpapagian/f1abf29cca711b38dce846cedf03c9b6).

I will do a full review to understand the whole approach. Sorry if I miss anything.

mtardy: For some obscure reasons, this breaks the feature and I can't understand (yet) why:

```diff
diff --git a/bpf/process/bpf_fork.c b/bpf/process/bpf_fork.c
index 2d36be4f..55ffe2d7 100644
--- a/bpf/process/bpf_fork.c
+++ b/bpf/process/bpf_fork.c
@@ -58,6 +58,7 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
        curr->key.pid = tgid;
        curr->key.ktime = ktime_get_ns();
        curr->nspid = get_task_pid_vnr();
+       memcpy(&curr->bin, &parent->bin, sizeof(curr->bin));
        curr->pkey = parent->key;
 
        /* Setup the msg_clone_event and sent to the user. */
```

tpapagian: Yes, initially this sounds a bit strange to me as well. After looking a bit more into that, it seems that you need to set the binary path to 0 before doing `curr->bin.path_length = probe_read_str(curr->bin.path, BINARY_PATH_MAX_LEN, &p->args)` in `execve_send`. 

Without the `memcpy` in clone event `curr->bin.path` is always zeroed. With the `memcpy`, you set an initial value to that buffer during the clone event. The `probe_read_str` call in `execve_send` only writes the first `N` bytes of the buffer where `N` is the length of the binary. If the binary length during `clone` is larger than the binary length during `exec` we still have some random bytes at the end of the buffer. In that case, we will never match in the map during `matchBinaries`. This seems to be the case in the tests. 

I believe a simple fix is to call something like `memset(&(curr->bin), 0, sizeof(struct binary))` before `curr->bin.path_length = probe_read_str(curr->bin.path, BINARY_PATH_MAX_LEN, &p->args)`.

pkg/sensors/exec/execvemap/execve.go
```diff
@@ -20,4 +20,5 @@ type ExecveValue struct {
 	Pad          uint32                     `align:"pad"`
 	Namespaces   processapi.MsgNamespaces   `align:"ns"`
 	Capabilities processapi.MsgCapabilities `align:"caps"`
+	BinaryPath   processapi.Binary          `align:"binary_path"`
```
tpapagian: This should be `align:"bin"` here. You fix that in a next commit but you can have it correct from here.

tpapagian (commented):
Thanks! This is a good improvement! I have done a first review and added some comments.

> This is not tied to this PR, but the binary path is just the arg passed to execve and thus can be a relative path (this explains many of the users issues). A future patch is needed to read the absolute path of the task_struct (as we do on userspace side with /proc to fill the initial state of the execve_map) to make this feature complete.

I agree that this is not tied to this PR but this would be also a great improvement in `matchBinaries` for a follow-up PR. I believe what you do here will help to have the full binary path ready from the kernel. Maybe we need to create an issue to keep track of this (if we do not already have one).

pkg/sensors/tracing/selectors.go
```diff
@@ -340,6 +346,18 @@ func populateStringFilterMap(
 	return nil
 }
 
+func populateMatchBinariesMaps(
+	ks *selectors.KernelSelectorState,
+	bpfMap *ebpf.Map,
+) error {
+	for innerID, sel := range ks.MatchBinaries() {
```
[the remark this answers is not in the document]
mtardy: yes!! will do indeed it's a leftover

pkg/selectors/selectors.go
```diff
@@ -46,11 +27,12 @@ type ValueReader interface {
...
-	StringMapsNumSubMaps   = 6
-	MaxStringMapsSize      = 6*stringMapsKeyIncSize + 1
-	StringPrefixMaxLength  = 128
-	StringPostfixMaxLength = 128
+	stringMapsKeyIncSize       = 24
+	StringMapsNumSubMaps       = 6
+	MaxStringMapsSize          = 6*stringMapsKeyIncSize + 1
+	StringPrefixMaxLength      = 128
+	StringPostfixMaxLength     = 128
+	MatchBinariesPathMaxLength = 256
```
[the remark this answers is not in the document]
mtardy: ouh yes nice catch!

bpf/lib/process.h
```diff
@@ -292,6 +292,23 @@ struct msg_execve_event {
...
 }; // All fields aligned so no 'packed' attribute.
 
+#define BINARY_PATH_MAX_LEN 256
+
+// This structure stores the binary path that was recorded on execve.
+// Technically PATH_MAX is 4096 but we limit the length we store since we have
+// limits on the length of the string to compare:
+// - Artificial limits for full string comparison.
+// - Technical limits for prefix and postfix, using LPM_TRIE that have a 256
+//   bytes size limit.
```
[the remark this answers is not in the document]
mtardy: Yes you are correct. This idea behind the first _artificial_ limit of 256 is that we align with LPM_TRIE limitation for what we can propose for the Prefix/Postfix. Also the length matters because the hash lookup will be orders of magnitude slower, the longer it is, because most of the time it will be filled with zeroes (so we would need some kind of optimization with variable size maps as we do with matchArgs and strings).

The idea is also that we need to store this path in execve_map, here we limit to 256 bytes so that for each process we have around 256 bytes more to store in maps. if we go to 4K, memory consumption will automatically max to 4K*number of processes.

Storing only the first 256 bytes is a bit annoying when we will need to do the Postfix, that's why in some cases (for process that have binary path len > 256) we will need to have 256 bytes for the beginning of the string and 256 bytes for the end of the string. For most processes this will be already overkill but this reduces the memory from 4096 to 512 for each process to do In/Prefix/Postfix.

I think it's already a high limit? In any case, if you need to match up to 256 you can use `Prefix`.

I checked on my host filesystem when I needed to choose between 128 and 256 (because hash lookup cost etc...) and I didn't find a lot of real-life files that had path > 128 chars. I guess if people suffer from this for their use case in the future we can always increase by switching the constants!

pkg/selectors/selectors.go
```diff
@@ -97,8 +101,9 @@ type KernelSelectorState struct {
 	// addr6Maps are used to populate IPv6 address LpmTrie maps for sock and skb operators
 	addr6Maps []map[KernelLPMTrie6]struct{}
 
-	matchBinaries map[int]*MatchBinariesMappings // matchBinaries mappings (one per selector)
-	newBinVals    map[uint32]string              // these should be added in the names_map
+	newBinVals       map[uint32]string              // these should be added in the names_map
+	oldMatchBinaries map[int]*MatchBinariesMappings // matchBinaries mappings (one per selector)
```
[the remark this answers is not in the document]
mtardy: yep this is just a temporary rename to not clash the names and this is removed in commit 9c511816c0e5aefb2598c3d156e491acf929b9b4.

pkg/sensors/exec/execvemap/execve.go
mtardy: this is fixed indeed, maybe I fixed it in the wrong commit and it's still present in some commits https://github.com/cilium/tetragon/pull/1731/files#diff-52bf8d0069ecc52a261cdf0cb40e9b51da31ec1c19cfda5b520729203c13618dR21. I can rebase if it's the case.

mtardy:
> Thanks! This is a good improvement! I have done a first review and added some comments.
> 
> > This is not tied to this PR, but the binary path is just the arg passed to execve and thus can be a relative path (this explains many of the users issues). A future patch is needed to read the absolute path of the task_struct (as we do on userspace side with /proc to fill the initial state of the execve_map) to make this feature complete.
> 
> I agree that this is not tied to this PR but this would be also a great improvement in `matchBinaries` for a follow-up PR. I believe what you do here will help to have the full binary path ready from the kernel. Maybe we need to create an issue to keep track of this (if we do not already have one).

Cool thanks for the careful review, I think everything was whether already fixed or I fixed it in [last push](https://github.com/cilium/tetragon/compare/fc29fa02d820d13215fa95c96aa9a5558741b97c..a80de98a06eb736d4db2def2c77fdfaf2368a504). Thanks for the insight on the clone thing 🙏 !

bpf/lib/process.h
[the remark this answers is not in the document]
tpapagian: Yes, I agree. I that we are fine with 256 bytes. My comment was mainly to check if I missed anything and just to keep in mind that if needed we can increase that further in the future. Thanks!

pkg/sensors/exec/execvemap/execve.go
tpapagian: Yes, the final result is (and was) fine. What I wanted to say here is that it is good to have separate commits as clean as possible because (1) it is easier to review and (2) it is easier to revert a specific commit rather than reverting a full PR. 

As an example for this specific thing you introduce 
```
BinaryPath processapi.Binary `align:"binary_path"`
```
in the third commit https://github.com/cilium/tetragon/pull/1731/commits/087cd9eabfafb1f568ad6dd72aae86236fc7435c. Then you change `binary_path` to `bin` in the seventh commit https://github.com/cilium/tetragon/pull/1731/commits/b4d1f789ffe3a66756267b8eca4680d52f90a8db. If you see the PR as a whole it is fine, if you see the PR as separate commits it is not. The same applies to other minor things in this PR as well.

I am fine with that in this PR. This is just a comment about future PRs.

tpapagian (approved):
Thanks for taking the time to apply my proposed fixes!

pkg/sensors/exec/execvemap/execve.go
mtardy: Ah yeah, it would be nice to fix that to have each commit to compile and run. I might have end up with this situation because I reorganized those commits many times and I might have been wrong in some rebase :/! Thanks

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -294,8 +294,12 @@ func writeExecveMap(procs []procs) {
 		v.Namespaces.TimeChildInum = p.time_for_children_ns
 		v.Namespaces.CgroupInum = p.cgroup_ns
 		v.Namespaces.UserInum = p.user_ns
+		v.BinaryPath.PathLength = int64(copy(v.BinaryPath.Path[:], p.exe))
```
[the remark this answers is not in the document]
kkourt: Had the same comment as Anastasios, but it seems that it was not addressed?

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -294,8 +294,12 @@ func writeExecveMap(procs []procs) {
...
 
-		m.Put(k, v)
+		err := m.Put(k, v)
+		if err != nil {
+			logger.GetLogger().WithError(err).Fatal("Could not put value in execve_map")
```
kkourt, line 302: Do we need to crash the agent here?
That's typically a bad idea (tm)

bpf/process/types/basic.h
```diff
@@ -1436,73 +1436,64 @@ struct {
...
+
+		switch (selector_options->op) {
+		case op_filter_in:
+		case op_filter_notin:
+			path_map = map_lookup_elem(&tg_mb_paths, &selidx);
+			if (!path_map)
+				return 0;
+			found_key = map_lookup_elem(path_map, current->bin.path);
+			match = !!found_key;
+			break;
```
kkourt, line 1490: The way that the code is now is confusing, because the `break` has no effect.

The other question that comes to mind when reading the code is can `->op` have a different value from op_filter_in and op_filter_notin. My guess is no, because we are setting the op from user-space. So _technically_ the switch statement is never used.

My guess is that we would add cases for prefix later.

My suggestion would be to add a 
```
default:
     /* should not happen */
     return 0;
```

In the switch statement to make the code less confusing.

kkourt (approved):
Thanks! LGTM!

Let's replace the `Fatal` with a warning, unless there is a reason not to do so.
Everything else is a nit so feel free to opt out.

bpf/process/types/basic.h
mtardy: Ah yes this is a leftover from this https://github.com/cilium/tetragon/pull/1732/commits/e4c545d0009ae3496147dcd1d13741e6d52218bd for the followup PR, I can adjust!

pkg/sensors/exec/procevents/proc_reader.go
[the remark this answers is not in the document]
mtardy: I fixed it in the wrong commit (at a later stage instead at when I introduced this) by mistake, now it's good!

pkg/selectors/kernel.go
```diff
@@ -1170,22 +1170,31 @@ func ParseMatchBinary(k *KernelSelectorState, b *v1alpha1.BinarySelector, selIdx
...
-			return fmt.Errorf("matchBinary error: Binary names > 255 chars do not supported")
+
+	// prepare the selector options
+	sel := MatchBinariesSelectorOptions{}
+	sel.Op = op
+
+	switch op {
+	case SelectorOpIn, SelectorOpNotIn:
+		for _, s := range b.Values {
+			if len(s) > 255 {
```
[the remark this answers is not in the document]
mtardy: I changed it at later stage because I don't change the implem, just add the new sel_opts map

pkg/sensors/exec/procevents/proc_reader.go
mtardy: yep done as a warning, I also added correct fields to the logging.

mtardy:
Thank you both, Anastasios I tried to fix all issues related to that https://github.com/cilium/tetragon/pull/1731#discussion_r1392066552, and Kornilios I changed the [switch and the Fatal log](https://github.com/cilium/tetragon/compare/a80de98a06eb736d4db2def2c77fdfaf2368a504..753507fa30b52635666c037b3406282e0cc73bd7).

PR #1732, opened 2023-11-09
Add `Prefix` and `NotPrefix` operators to matchBinaries

mtardy:
This PR goes on top of https://github.com/cilium/tetragon/pull/1731, it looks scary because it contains the commits of the previous PR but those will disappear once the first one is merged. 

It fixes https://github.com/cilium/tetragon/issues/1278.

The implementation is pretty trivial thanks to the work that was already done with prefix on `matchArgs` and the changes of the https://github.com/cilium/tetragon/pull/1731.


mtardy:
cc @tpapagian @kkourt (I pinged you since you reviewed the previous one, it might be easier to read this one)

This is not 100% done (still in draft) because I still need to make it work with 4.19 but if you have some time to take a look 👀 while I'm on PTO (basically end of this week and next week) it would be nice.

For 4.19 to work, maybe we can reduce the max size of the prefix `STRING_PREFIX_MAX_LENGTH 256`, which might fix the E2BIG, or just not propose the `Prefix` and `NotPrefix` feature completely for <5.4.

mtardy:
So I've disabled matchBianries Prefix for kernel < 5.3, https://github.com/cilium/tetragon/pull/1732/commits/d92ee0ccf71d30c758893bca783d39217b3628da. Now everything passes.

bpf/process/types/basic.h
```diff
@@ -1536,13 +1538,26 @@ static inline __attribute__((always_inline)) int match_binaries(__u32 selidx)
...
+			break;
+		case op_filter_prefix:
+		case op_filter_notprefix:
+			path_map = map_lookup_elem(&string_prefix_maps, &selector_options->map_id);
+			if (!path_map)
+				return 0;
+			// prepare the key on the stack to perform lookup in the LPM_TRIE
+			memset(&prefix_key, 0, sizeof(prefix_key));
+			prefix_key.prefixlen = current->bin.path_length * 8; // prefixlen is in bits
+			copy_length = probe_read_str(prefix_key.data, current->bin.path_length & (STRING_PREFIX_MAX_LENGTH - 1), current->bin.path);
```
kevsecurity, line 1553: Should this not be a probe_read instead of probe_read_str? The latter will include a \0 at the end of the string, included in the length.

kevsecurity: And actually, maybe a path can contain a \0 at which point the probe_read_str would stop copying?

mtardy: I'm not sure a unix path can contain a `\0` but indeed we don't mind using `probe_read` or `probe_read_str` here. `probe_read_str` will return the copied_length while `probe_read` will return 0 on success, anyway here we just check that the return value is `< 0`. In the end it can copy a max of `256 - 1` in a buffer of size `256` with all values set to 0 so it won't change much, it seems that `prefix_key.data[255]` will always be `0` anyway.

I guess it's a cheaper approach to use `probe_rea`d since we know the size in advance?

mtardy:
cc @kevsecurity 

I bumped the CRD version, used `probe_read` instead `probe_read_str` in the `prefix_key.data` construction, and removed commit 608a1abf3564d5f990e81a3e02a24dca6b557448 that was renaming `op_filter_str_prefix` to `op_filter_prefix`.

kevsecurity (approved):
LGTM

tpapagian (approved):
LGTM, thanks!

PR #1735, opened 2023-11-10
pkg/bench: disable BPF stats collection at the end of execution

lmb:
Turning on BPF stats collection causes overhead, so it makes sense to turn it off at the end of benchmarking. On newer kernels we can use ebpf.EnableStats as a safer interface.


lmb:
Not sure how to test this exactly, but the code compiles xD

kkourt (approved):
Makes sense, thanks!

PR #1743, opened 2023-11-13
tetragon: save tetragon pid into pidfile

tixxdz:


pkg/pidfile/pidfile.go
```diff
@@ -0,0 +1,75 @@
...
+		return pid, ErrPidIsStillAlive
+	}
+
+	if err != nil &&
+		errors.Is(err, ErrPidFileNotPresent) == false &&
+		errors.Is(err, ErrPidIsNotAlive) == false {
+		return 0, err
+	}
+
+	pid, err = proc.GetSelfPid(option.Config.ProcFS)
```
mtardy, line 65: I think you could use standard lib https://pkg.go.dev/os#Getpid here as well to simplify and remove the function from `proc` package.

pkg/defaults/defaults.go
```diff
@@ -42,6 +42,9 @@ const (
 
 	// Default secure export logs permissions
 	DefaultLogsPermission = "600"
+
+	// Pid file where to write tetragon main PID
+	DefaultPidFile = DefaultRunDir + "tetragon.pid"
```
mtardy, line 47: nit: I would use `path.join` in case the `/` gets removed someday or whatever.

pkg/pidfile/pidfile.go
```diff
@@ -0,0 +1,75 @@
...
+	ErrPidIsStillAlive   = errors.New("process is already running")
+)
+
+func readPidFile() (uint64, error) {
+	_, err := os.Stat(defaults.DefaultPidFile)
+	if err != nil {
+		return 0, ErrPidFileNotPresent
+	}
+
+	data, err := os.ReadFile(defaults.DefaultPidFile)
```
mtardy, line 29: you can check for returned error from `os.ReadFile` (https://pkg.go.dev/os#pkg-variables) and it can return: ErrNotExist   = [fs](https://pkg.go.dev/io/fs).[ErrNotExist](https://pkg.go.dev/io/fs#ErrNotExist). You can remove the `os.Stat` then.

mtardy (commented):
Looks good thanks! A few comments if you are interested, I think you might be able to remove code here like `proc.GetSelfPid()` :)

I know that you don't want to change PR for convention things but just want to mention, for Golang convention for future PRs: [avoid stutter in package name and function](https://go.dev/doc/effective_go#package-names), like:
- avoid `pidfile.DeletePidFile()` in favor of `pidfile.Delete()`,
- avoid `pidfile.CreatePidFile()` instead name `pidfile.Create()`,
- etc.

pkg/pidfile/pidfile.go
tixxdz: Alright done ;-)

pkg/defaults/defaults.go
tixxdz: agree, but don't want to import extra packages there, think it is fine, thank you ;-)

pkg/pidfile/pidfile.go
tixxdz: Since we export the procfs so I want to go with it, and avoid any pid namespace foo that could mess up the getpid since usually it works with namespaces ;-)

tixxdz:
> Looks good thanks! A few comments if you are interested, I think you might be able to remove code here like `proc.GetSelfPid()` :)
> 
> I know that you don't want to change PR for convention things but just want to mention, for Golang convention for future PRs: [avoid stutter in package name and function](https://go.dev/doc/effective_go#package-names), like:
> 
>     * avoid `pidfile.DeletePidFile()` in favor of `pidfile.Delete()`,
> 
>     * avoid `pidfile.CreatePidFile()` instead name `pidfile.Create()`,
> 
>     * etc.

Good point @mtardy done ;-)

pkg/pidfile/pidfile.go
mtardy: ah I see! completely out of subject but speaking about this stuff, I read `exe` in /proc in the matchBinaries rework on the userside as root, is there anything about namespace that I should be concerned about?

mtardy (approved):
cool! thanks :)

PR #1753, opened 2023-11-14
Update install-k8s.md

lizrice:
Fix typo in Kind instructions


mtardy (approved):
The fix is good but maybe it would be nice to get to the bottom of this before merging?

kkourt:
> The fix is good but maybe it would be nice to get to the bottom of this before merging?

This flag was introduced in 1.0, so my guess is the kind chart that produced the error was before 1.0.
I've also realized, however, that the kind config would only work on Linux systems. So I pushed an additional patch to make this distinction. Let me know what you think.

PR #1754, opened 2023-11-14
Added some additional resources

saintdle:
Looks like they were dropped with the move to 1.0, so adding them back


mtardy (approved):
Thanks

PR #1756, opened 2023-11-14
renovate: fix allowed version for Go upgrades in 1.0 branch

mtardy:
I finally understood that the allowedVersion was broken, restricting <1.22. Docker does not have version semantics because of tags so it was using a fallback to npm and thus filtering out any version with the -alpine suffix. Explaining the missing updates on stable 1.0.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -209,6 +209,22 @@
       }
     },
     {
+      "groupName": "Go",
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):
i trust you

mtardy:
> i trust you

wow thank you I don't even trust myself

PR #1760, opened 2023-11-15
chore: site description update

hacktivist123:
This PR updates the site description

### DIFF
--- a/docs/hugo.toml
+++ b/docs/hugo.toml
@@ -1,6 +1,6 @@
 baseURL = "/"
 title = "Tetragon - eBPF-based Security Observability and Runtime Enforcement"
-description = "eBPF-based Security Observability and Runtime Enforcement"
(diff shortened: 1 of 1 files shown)


docs/hugo.toml
```diff
@@ -1,6 +1,6 @@
 baseURL = "/"
 title = "Tetragon - eBPF-based Security Observability and Runtime Enforcement"
-description = "eBPF-based Security Observability and Runtime Enforcement"
+description = "Tetragon is a low overhead, flexible, and Kubernetes native security observability tool that applies policy and filtering directly in the kernel"
```
kkourt, line 3:
```suggestion
description = "Tetragon is a low overhead, flexible, and Kubernetes native security observability tool that applies policy and filtering directly in the kernel using eBPF"
```

Would it make sense to keep eBPF in the description?

kkourt (commented):
Thanks! Please find a minor comment below.

hacktivist123:
I’m not sure that’s necessary to be honest since the Title mentions it. It would be repeating and can cause SEO downranking.

PR #1762, opened 2023-11-16
fieldfilters: fix case conversion and improve performance

willfindlay:
Commit 1 splits field filters into their own package.
Commit 2 adds fieldfilter benchmarks and commit 3 adds a new convenience target to the Makefile to run benchmarks.
Commit 4 fixes a bug related to configuring multiple field filters with the wrong field case.
Commit 5 significantly improves fieldfilter performance to be near zero-overhead.

```release-note
Fix a few bugs related to field filter configuration and significantly improve performance of field filters.
```


pkg/fieldfilters/fields.go
```diff
@@ -189,24 +204,22 @@ func (f *FieldFilter) Filter(event *tetragon.GetEventsResponse) (*tetragon.GetEv
...
-		event := rft.Mutable(eventDesc).Message().Interface()
-		switch f.action {
-		case tetragon.FieldFilterAction_INCLUDE:
-			f.fields.Filter(event)
-		default:
-			f.fields.Prune(event)
-		}
+		event := src.Get(fd).Message().Interface()
+		dstEvent := dst.Mutable(fd).Message().Interface()
+		filterErr = fieldmask_utils.StructToStruct(f.fields, event, dstEvent)
```
kkourt, line 216: maybe erorrs.Join() is better here so that we don't miss errors? I believe this code will only keep the value from the last iteration.

kkourt (approved):
Thanks, looks good to me.

pkg/fieldfilters/fields.go
willfindlay: the intention was actually to return early (which I failed to do here) but I like your idea better

PR #1773, opened 2023-11-20
tetragon: Move return filter to kernel

olsajiri:
moving the return filter to bpf program and adding support for return actions


kkourt (commented):
Thanks! This looks good overall.

Just some minor comments.

Also: was this https://github.com/cilium/tetragon/pull/1781#issuecomment-1824253669 resolved? Would be nice to reject the tracing policy early if something is not supported.

olsajiri:
> Thanks! This looks good overall.
> 
> Just some minor comments.
> 
> Also: was this [#1781 (comment)](https://github.com/cilium/tetragon/pull/1781#issuecomment-1824253669) resolved? Would be nice to reject the tracing policy early if something is not supported.

ugh, forgot about this one.. sry, added PR https://github.com/cilium/tetragon/pull/1863

pkg/sensors/tracing/generickprobe.go
```diff
@@ -360,6 +376,11 @@ func createMultiKprobeSensor(sensorPath string, multiIDs []idtable.EntryID) ([]*
 		retConfigMap := program.MapBuilderPin("config_map", sensors.PathJoin(pinPath, "retprobe_config_map"), loadret)
 		maps = append(maps, retConfigMap)
 
+		retFilterMap := program.MapBuilderPin("filter_map", sensors.PathJoin(pinPath, "retprobe_filter_map"), loadret)
+		maps = append(maps, retFilterMap)
+
+		maps = append(maps, filterMaps(loadret, pinPath, nil)...)
```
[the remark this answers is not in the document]
olsajiri: right..good catch.. the state is needed for single kprobe where we can fix maximums for kernel < 5.9, so it's not needed for multi kprobe which is available for later kernels, so there's no need for this fixes.. I'll make the fix and add comment

pkg/bpf/detect.go
```diff
@@ -42,12 +42,10 @@ func detectOverrideHelper() bool {
 		},
 		License: "GPL",
 	})
-	prog.Close()
```
mtardy, line 45: that's the kind of situation where you can use `defer`:
```go
defer prog.Close()
```
so that's it is executed anyway.

pkg/bpf/detect.go
```diff
@@ -69,12 +67,10 @@ func detectSignalHelper() bool {
...
```
mtardy, line 72: same here with `defer`.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -79,6 +79,7 @@ func kprobeCharBufErrorToString(e int32) string {
 
 type kprobeSelectors struct {
 	entry *selectors.KernelSelectorState
+	retrn *selectors.KernelSelectorState
```
mtardy, line 81: typo here on `retrn`.

mtardy (commented):
this looks good! thanks!

pkg/bpf/detect.go
olsajiri: hum, but we need to be sure the program got loaded and then we just return, so does not seem defer is helpful in this case.. but I just found same problem in detectLargeProgramSize, will fix that

pkg/sensors/tracing/generickprobe.go
olsajiri: on purpose ;-) can't have return as variable name

mtardy: ahah I'm stupid

pkg/bpf/detect.go
mtardy: oh okay nevermind!!

olsajiri:
hum, I think the 32 bit change that got merged blocked this change, because we cross the instructions limit for filter program on small kernels.. we'll need to split the filtering logic and then merge this change

olsajiri:
> hum, I think the 32 bit change that got merged blocked this change, because we cross the instructions limit for filter program on small kernels.. we'll need to split the filtering logic and then merge this change

ok, managed to reorg the code a bit and we're still good ;-) but I'll start on the filter re-work anyway

kkourt (approved):
Thanks!

PR #1777, opened 2023-11-20
podinfo: Handle an expected error gracefully

michi-covalent:
Sometimes creating a podinfo fails with AlreadyExists error even though the previous call to get the podinfo returned NotFound because of a timing issue. Requeue without returning the error when this happens to prevent the controller from logging an unnecessary error.

### DIFF
--- a/operator/podinfo/podinfo_controller.go
+++ b/operator/podinfo/podinfo_controller.go
@@ -56,7 +56,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			return ctrl.Result{}, err
 		}
 		// Pod info does not exist. Create it.
-		return ctrl.Result{}, r.Create(ctx, generatePodInfo(pod))
+		err = r.Create(ctx, generatePodInfo(pod))
+		if errors.IsAlreadyExists(err) {
+			// Sometimes Create fails with "AlreadyExists" error even though the
+			// previous call to Get returned "NotFound" because of a timing issue.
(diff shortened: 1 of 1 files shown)


operator/podinfo/podinfo_controller.go
```diff
@@ -56,7 +56,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 		}
 		// Pod info does not exist. Create it.
-		return ctrl.Result{}, r.Create(ctx, generatePodInfo(pod))
+		err = r.Create(ctx, generatePodInfo(pod))
+		if errors.IsAlreadyExists(err) {
+			// Sometimes Create fails with "AlreadyExists" error even though the
+			// previous call to Get returned "NotFound" because of a timing issue.
+			// Requeue without returning the error when this happens, otherwise
+			// the controller logs an error.
+			return ctrl.Result{Requeue: true}, nil
```
sayboras, line 65: Do we need to requeue?

sayboras (approved):
Small comment as per below.

operator/podinfo/podinfo_controller.go
michi-covalent: yeah so the idea is to requeue so that the reconciler exercises the update path the next time around, just to make sure the podinfo has all the up-to-date information.

sayboras: sounds good :heavy_check_mark:

willfindlay (approved):

PR #1779, opened 2023-11-21
String: Increase prefix size from 128 to 256 chars

kevsecurity:
The file prefix limit was originally 256 characters. When the code was converted to use LMP TRIE maps, the limit was set to 128 characters.

This commit changes the prefix limit from 128 to 256 characters.

```release-note
Increases the character limit for prefix matches from 128 to 256.
```

### DIFF
--- a/bpf/process/string_maps.h
+++ b/bpf/process/string_maps.h
@@ -138,7 +138,7 @@ struct {
 	__uint(value_size, 512);
 } string_maps_ro_zero SEC(".maps");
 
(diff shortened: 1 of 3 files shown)


lambdanis (approved):
I believe it's correct

mtardy:
I got raced 🏎️ https://github.com/cilium/tetragon/pull/1732/commits/c02af3e1fee37f9c57ed7c3caa67e786af092a9d

PR #1781, opened 2023-11-21
tetragon: Assorted fixes

olsajiri:
Assorted fixes as preparation for return kprobe filtering support.
Sending self contained fixes first, so the final PR is not too big.

### DIFF
--- a/bpf/process/bpf_generic_retkprobe.c
+++ b/bpf/process/bpf_generic_retkprobe.c
@@ -3,8 +3,10 @@
 
 #include "vmlinux.h"
 #include "api.h"
-#include "bpf_tracing.h"
 
+#define GENERIC_KRETPROBE
+
+#include "bpf_tracing.h"
 #include "bpf_event.h"
 #include "bpf_task.h"
 #include "retprobe_map.h"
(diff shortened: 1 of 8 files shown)


tpapagian (approved):
LGTM!

kkourt (commented):
Thanks! Had a quick look and everything LGTM.

I do have one question based on the newly added operators.

> Unfortunately we are reaching bpf program verifier complexity limit
on 4.19 kernels, so making this change only for large programs.

What will happen if users specify the operators in 4.19?
Do we have code in place to reject these policies since they will not be applied in the datapath?

olsajiri:
> What will happen if users specify the operators in 4.19? Do we have code in place to reject these policies since they will not be applied in the kernel side?

hm, right.. so I added that code because follow up PR is removing user space filters
for return probe and use this code in kernel instead.. I haven't though about using
LT/GT from entry probe.. that probably needs some additional checking..
is it ok in follow up PR? @kkourt

kkourt (approved):

kkourt:
> > What will happen if users specify the operators in 4.19? Do we have code in place to reject these policies since they will not be applied in the kernel side?
> 
> hm, right.. so I added that code because follow up PR is removing user space filters for return probe and use this code in kernel instead.. I haven't though about using LT/GT from entry probe.. that probably needs some additional checking.. is it ok in follow up PR? @kkourt

Yap. Maybe it's worth creating an issue to track this until it's resolved. Thanks.

PR #1786, opened 2023-11-23
tetragon: detect if binary execution raised process privileges

tixxdz:
tetragon: detect binary execution that raise process privileges

From a higher level:
Populate `binary_properties.privileges_changed` with the
the reasons why this binary execution gained new capabilities or
elevated its privileges. Usually this happens due:
1. File capability sets on the binary.
2. set-user-ID to root bit is set on the binary.

For this we introduce a new `ProcessPrivilegesChanged` enum type with the following values:

1. PRIVILEGES_RAISED_EXEC_FILE_CAP - The File capability sets on binary:
  The kernel supports associating capability sets with an executable
  file using `setcap` command. The file capability sets are stored in an extended attribute
  named `security.capability`. The file capability sets, in conjunction
  with the capability sets of the process, determine the process
  capabilities and privileges after the `execve` system call.
  For further reference, please check sections
  `File capability extended attribute versioning` and
  `Namespaced file capabilities` of the capabilities man pages:
  https://man7.org/linux/man-pages/man7/capabilities.7.html.

2. PRIVILEGES_RAISED_EXEC_FILE_SETUID - set-user-ID to root bit is set
  on the binary:
  When a process with nonzero UIDs executes a binary with a set-user-ID
  to root also known as suid-root executable, then the kernel switches
  the effective user ID to 0 (root) which is a privilege elevation
  operation since it grants access to resources owned by the root user.
  The effective user ID is listed inside the `process_credentials` part
  of the `process` object.
  For further reading, section
  `Capabilities and execution of programs by root` of
  https://man7.org/linux/man-pages/man7/capabilities.7.html.
  Afterward the kernel recalculates the capability sets of the process
  and grants all capabilities in the permitted and effective capability
  sets, except those masked out by the capability bounding set.
  If the binary also have file capability sets then these bits are honored
  and the process gains just the capabilities granted by the file
  capability sets (i.e., not all capabilities, as it would occur
  when executing a set-user-ID to root binary that does not have any
  associated file capabilities).
  This is described in section
  `Set-user-ID-root programs that have file capabilities` of
  https://man7.org/linux/man-pages/man7/capabilities.7.html.
  The new granted capabilities can be listed inside the `process` object.
  There is one exception for the special treatments of set-user-ID to
  root execution receiving all capabilities, if the `SecBitNoRoot` bit of
  the Secure bits is set, then the kernel does not grant any capability.
  The secure bits are already exposed part of the `process_credentials`
  object. Please check section:
  `The securebits flags: establishing a capabilities-only environment`
  of the capabilities man pages:
  https://man7.org/linux/man-pages/man7/capabilities.7.html

3. PRIVILEGES_RAISED_EXEC_FILE_SETGID - set-group-ID to root
   This is the execution of a binary with set-group-ID to root. Strictly speaking
   the kernel does not treat this as a privileged execution, however in Tetragon
   context this is a privilege raise since it allows to run with effective group
   ID root which grants access to resources owned by group root.

The new privileges_changed contains the reasons, no need to encode
again the full capabilities there, let's just indicate to the user
and if they are interested they can inspect other fields that contain
the permitted and effective capabilities of the current execution.

Implementation:
The capabilities are re-calculated by the capability LSM during execve,
so let's avoid trying to be smart and calculate or guess them...
we just display the final results.

To display these fields, Tetragon must be run with enable-process-cred.

1. If the setuid bit is set we check if it is a privileged root and if
   it's real uid is not global root and effective uid is global root, if so
   then this a suid binary execution which is already enough for
   Tetragon to report as a privilege change execution.

2. Check if the setuid bit is not set and the new permitted capability
   sets is higher then the previous one then is is a file capability
   sets execution.

Limitations:
1. We do not handle the case of an unprivileged user id that is
  _mapped_ to user id root 0 inside its own namespace then performs a
  setuid to root execution. Current eBPF infrastructure does not support
  user id translation into target user namespace. Plus current Tetragon
  design deals with global uids, so it is not clear at this moment how to
  handle this corner case. For this the binary_properties.privileges_changed
  is an array that can hold extra values in the future.

2. We do not handle the case of a binary having both setuid to root bit
  set and file capability sets on the binary. Right now we report in the
  `privileges_changed` only `PRIVILEGES_RAISED_EXEC_FILE_SETUID` where
  we should also add `PRIVILEGES_RAISED_EXEC_FILE_CAP`. This is left for
  the future as we have to read the `security.capability` field of the
  extended attributes of the binary and parse the capabilities that are
  actually set to ensure that new capabilities received are not only due
  to the set-user-ID root exection.

Example output event:

File capabilities execution:
```
  "process_exec": {
    "process": {
      "exec_id": "OjIwMTAxOTMxNzAzMjc4OjE0MDUwOA==",
      "pid": 140508,
      "uid": 1000,
      "binary": "/usr/bin/ping",
      "arguments": "ebpf.io",
      "flags": "execve clone",
      "start_time": "2023-11-23T14:32:53.227623175Z",
      "auid": 1000,
      "parent_exec_id": "OjI5NjYzNzAwMDAwMDA6NDQ4NTk=",
      "refcnt": 1,
      "cap": {
        "permitted": [
          "CAP_NET_RAW"
        ],
        "effective": [
          "CAP_NET_RAW"
        ]
      },
      "tid": 140508,
      "process_credentials": {
        "uid": 1000,
        "gid": 1000,
        "euid": 1000,
        "egid": 1000,
        "suid": 1000,
        "sgid": 1000,
        "fsuid": 1000,
        "fsgid": 1000
      },
      "binary_properties": {
        "privileges_changed": [
          "PRIVILEGES_RAISED_EXEC_FILE_CAP"
        ]
      }
    }
```

Setuid suid execution
(from non uid 0 -> to global uid 0 in root user namespace):
```
  "process_exec": {
    "process": {
      "exec_id": "OjMyMTg1ODY1MDcxNTQ6Mjg5NTU=",
      "pid": 28955,
      "uid": 1000,
      "binary": "/usr/bin/su",
      "arguments": "--help",
      "flags": "execve clone",
      "start_time": "2023-12-11T14:59:24.735197753Z",
      "auid": 1000,
      "parent_exec_id": "OjE0MjI5NzAwMDAwMDA6MTI4NjE=",
      "refcnt": 1,
      "cap": {
        "permitted": [
          "CAP_CHOWN",
          "DAC_OVERRIDE",
          "CAP_DAC_READ_SEARCH",
          "CAP_FOWNER",
          "CAP_FSETID",
          "CAP_KILL",
          "CAP_SETGID",
          "CAP_SETUID",
          "CAP_SETPCAP",
          "CAP_LINUX_IMMUTABLE",
          "CAP_NET_BIND_SERVICE",
          "CAP_NET_BROADCAST",
          "CAP_NET_ADMIN",
          "CAP_NET_RAW",
          "CAP_IPC_LOCK",
          "CAP_IPC_OWNER",
          "CAP_SYS_MODULE",
          "CAP_SYS_RAWIO",
          "CAP_SYS_CHROOT",
          "CAP_SYS_PTRACE",
          "CAP_SYS_PACCT",
          "CAP_SYS_ADMIN",
          "CAP_SYS_BOOT",
          "CAP_SYS_NICE",
          "CAP_SYS_RESOURCE",
          "CAP_SYS_TIME",
          "CAP_SYS_TTY_CONFIG",
          "CAP_MKNOD",
          "CAP_LEASE",
          "CAP_AUDIT_WRITE",
          "CAP_AUDIT_CONTROL",
          "CAP_SETFCAP",
          "CAP_MAC_OVERRIDE",
          "CAP_MAC_ADMIN",
          "CAP_SYSLOG",
          "CAP_WAKE_ALARM",
          "CAP_BLOCK_SUSPEND",
          "CAP_AUDIT_READ",
          "CAP_PERFMON",
          "CAP_BPF",
          "CAP_CHECKPOINT_RESTORE"
        ],
        "effective": [
          "CAP_CHOWN",
          "DAC_OVERRIDE",
          "CAP_DAC_READ_SEARCH",
          "CAP_FOWNER",
          "CAP_FSETID",
          "CAP_KILL",
          "CAP_SETGID",
          "CAP_SETUID",
          "CAP_SETPCAP",
          "CAP_LINUX_IMMUTABLE",
          "CAP_NET_BIND_SERVICE",
          "CAP_NET_BROADCAST",
          "CAP_NET_ADMIN",
          "CAP_NET_RAW",
          "CAP_IPC_LOCK",
          "CAP_IPC_OWNER",
          "CAP_SYS_MODULE",
          "CAP_SYS_RAWIO",
          "CAP_SYS_CHROOT",
          "CAP_SYS_PTRACE",
          "CAP_SYS_PACCT",
          "CAP_SYS_ADMIN",
          "CAP_SYS_BOOT",
          "CAP_SYS_NICE",
          "CAP_SYS_RESOURCE",
          "CAP_SYS_TIME",
          "CAP_SYS_TTY_CONFIG",
          "CAP_MKNOD",
          "CAP_LEASE",
          "CAP_AUDIT_WRITE",
          "CAP_AUDIT_CONTROL",
          "CAP_SETFCAP",
          "CAP_MAC_OVERRIDE",
          "CAP_MAC_ADMIN",
          "CAP_SYSLOG",
          "CAP_WAKE_ALARM",
          "CAP_BLOCK_SUSPEND",
          "CAP_AUDIT_READ",
          "CAP_PERFMON",
          "CAP_BPF",
          "CAP_CHECKPOINT_RESTORE"
        ]
      },
      "tid": 28955,
      "process_credentials": {
        "uid": 1000,
        "gid": 1000,
        "euid": 0,
        "egid": 1000,
        "suid": 0,
        "sgid": 1000,
        "fsuid": 0,
        "fsgid": 1000
      },
      "binary_properties": {
        "setuid": 0,
        "privileges_changed": [
          "PRIVILEGES_RAISED_EXEC_FILE_SETUID"
        ]
      }
    }
```


kkourt:
Would it be possible to add a test for this?

pkg/sensors/exec/exec_test.go
```diff
@@ -945,23 +948,125 @@ func TestExecProcessCredentialsSuid(t *testing.T) {
...
+	if err := exec.Command(testSuid).Run(); err != nil {
+		t.Fatalf("Failed to execute '%s' binary: %s\n", testSuid, err)
+	}
+
+	if err = syscall.Setuid(0); err != nil {
+		t.Fatalf("Failed to restore uid to 0 :  %s\n", err)
+	}
+	if err = syscall.Setgid(0); err != nil {
+		t.Fatalf("Failed to restore gid to 0 :  %s\n", err)
+	}
```
kkourt, line 982: What is this restoring?
I'm not seeing any Setuid() calls or Setgid calls in this test. There are some in the test below, so maybe that's what it tries to fix? In any case, I think we should add a t.Cleanup() for resetting the set{u,g}ids wherever we change them so that t is clear (i) why the happen, (ii) that the set values will be restored.

pkg/sensors/exec/exec_test.go
```diff
@@ -945,23 +948,125 @@ func TestExecProcessCredentialsSuid(t *testing.T) {
...
+	})
+
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	/* Executing a setgid to root with current gid as normal not root */
 	gid := 1879048188
 	if err := syscall.Setgid(gid); err != nil {
-		t.Fatalf("setegid(%d) error: %s", gid, err)
+		t.Fatalf("setgid(%d) error: %s", gid, err)
```
kkourt, line 1031: Maybe this is the Setgid we are trying to restore? ☝🏼

pkg/sensors/exec/exec_test.go
```diff
@@ -1215,3 +1216,75 @@ func TestExecProcessCredentialsSetuidChanges(t *testing.T) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
+	obs, err := observertesthelper.GetDefaultObserver(t, ctx, tus.Conf().TetragonLib, observertesthelper.WithMyPid())
+	if err != nil {
+		t.Fatalf("Failed to run observer: %s", err)
+	}
+
+	// The drop-privileges is a helper binary that drops privileges so we do not
+	// drop it inside this test which will break the test framework.
```
kkourt, line 1234: The binary makes sense to me.
But don't we set the gid/uid elsewhere in the tests? Is that different?

kkourt (commented):
Thanks for the great explanation and the tests!

I have some comments regarding the tests. There are two approaches to test this: (i) via binaries, and (ii) via syscalls. It seems that we do both. The (i) is simple. For (ii), we need to reset whatever we changed back to its original value after the test.

Do we need (ii) if we have (i)? That is, can't we just test everything we want with external binaries without messing with the test process?

If we need (ii), let's use t.Cleanup() to make it clear what we are resting back to its original.

pkg/sensors/exec/exec_test.go
tixxdz: Alright moved the setuid and setgid restore logic into t.Cleanup() for the tests that need it, for rest remove it as it useless, thank you !

pkg/sensors/exec/exec_test.go
tixxdz: Ah the gid/uid for this case are set in the drop-privileges binaries to drop all privileges and caps, then it will execute ping binary to check for new gained caps

tixxdz:
> Thanks for the great explanation and the tests!
> 
> I have some comments regarding the tests. There are two approaches to test this: (i) via binaries, and (ii) via syscalls. It seems that we do both. The (i) is simple. For (ii), we need to reset whatever we changed back to its original value after the test.
> 
> Do we need (ii) if we have (i)? That is, can't we just test everything we want with external binaries without messing with the test process?
>

We surely can, let's summarize:
- For every drop or change gid values it is fine, as these don't affect the capabilities nor the uid, so it is perfectly fine to do that in test and we restore in case.
- For every drop or change uid values, then it gets complicated as the kernel may also drop capabilities, we surely can keep some privileges with KEEP_CAPS secure bit, but then the test framework will fail since the uid is not 0 or the one we start with, it needs some extra environment variables and tmpfs mounts maybe etc, this is why I prefered to do it in an external binary. So we change the uids in external binary then execute the privileged binaries and check results. I'm improved test to make it clear.
 
> If we need (ii), let's use t.Cleanup() to make it clear what we are resting back to its original.

So I would have preferred to not use external binary but it gets complicated. So for (i) binaries it is only when you change uids (as kernel drop caps) and it handles all cases now, and for (ii) via syscalls indeed I converted it to be inside t.Cleanup()

You may find some extra restore setuid(0) in t.Cleanup() calls but that's just a guard in case we mess up things, that call without appropriate privileges will fail and make it easy to debug tests as that point, instead of wondering why next tests are failing.

Thank you for the review @kkourt

tpapagian (approved):
LGTM, thanks for the great explanation!

bpf/process/bpf_execve_bprm_commit_creds.c
```diff
@@ -76,12 +79,60 @@ BPF_KPROBE(tg_kp_bprm_committing_creds, struct linux_binprm *bprm)
 	gid = BPF_CORE_READ(task, cred, gid.val);
 
 	/* Is setuid? */
-	if (euid != uid)
+	if (euid != uid) {
 		heap->info.secureexec |= EXEC_SETUID;
+		/* If euid is being changed to root? */
+		ruid = BPF_CORE_READ(bprm, cred, uid.val);
+		if (!__is_uid_global_root(ruid) && __is_uid_global_root(euid))
```
kkourt, line 86: What is `ruid` and why we check whether it's not zero? Is there a case where `uid != 0` but `ruid == 0`?
I guess it's whatever the kernel will call `commit_creds` with?

bpf/lib/bpf_cred.h
```diff
@@ -69,4 +69,37 @@ struct msg_cred_minimal {
 	__u32 pad;
 } __attribute__((packed));
 
+/* Execution and cred related flags shared with userspace */
+#define EXEC_SETUID    0x01 /* This is a set-user-id execution */
+#define EXEC_SETGID    0x02 /* This is a set-group-id execution */
+#define EXEC_FS_CAPS   0x04 /* This binary execution gained new capabilities through file capabilities execution */
+#define EXEC_FS_SETUID 0x08 /* This binary execution gained new privileges through setuid to root execution */
```
kkourt, line 76: nit: no need to change anything, just something to be aware of.

I find the naming confusing. Here's my understanding based of the comments:
* We use `EXEC_SETUID` for all setuid binaries
* We use `EXEC_FS_SETUID` for execution of setuid binaries that lead to privilege elevation

Does the `FS` stand for filesystem? The reason I find it confusing is that, if my understanding is correct, `EXEC_SETUID` is also related to filesystem properties.

bpf/process/bpf_execve_bprm_commit_creds.c
```diff
@@ -102,8 +102,16 @@ BPF_KPROBE(tg_kp_bprm_committing_creds, struct linux_binprm *bprm)
 			heap->info.secureexec |= EXEC_FS_SETUID;
 	}
 	/* Is setgid? */
-	if (egid != gid)
+	if (egid != gid) {
 		heap->info.secureexec |= EXEC_SETGID;
+		/* Is egid is being changed to real root? */
+		gid = BPF_CORE_READ(bprm, cred, gid.val);
```
kkourt, line 108: Same question as for uid, what is `bprm->cred.gid`? Can it be different than task->cred.gid?

pkg/sensors/exec/exec_test.go
```diff
@@ -945,23 +948,125 @@ func TestExecProcessCredentialsSuid(t *testing.T) {
...
+	// We should be able to create suid on local mount point
+	testSuid := testutils.RepoRootPath("contrib/tester-progs/suidnop")
+	if err := testutils.CopyFile(testSuid, testBin, 0754|os.ModeSetuid|os.ModeSetgid); err != nil {
+		t.Fatalf("Failed to copy binary: %s", err)
+	}
+
+	t.Cleanup(func() {
+		if err = syscall.Setuid(0); err != nil {
+			t.Fatalf("Failed to restore uid to 0 :  %s\n", err)
+		}
```
kkourt, line 1014: Why are we setting setuid to `0` here? Shouldn't we set it to whatever the previous value was?

pkg/sensors/exec/exec_test.go
```diff
@@ -945,23 +948,125 @@ func TestExecProcessCredentialsSuid(t *testing.T) {
...
 
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	/* Executing a setgid to root with current gid as normal not root */
 	gid := 1879048188
 	if err := syscall.Setgid(gid); err != nil {
-		t.Fatalf("setegid(%d) error: %s", gid, err)
+		t.Fatalf("setgid(%d) error: %s", gid, err)
 	}
```
kkourt, line 1033: Adding my previous  [comment](https://github.com/cilium/tetragon/pull/1786#discussion_r1431248693) for context:
> I'm not seeing any Setuid() calls or Setgid calls in this test. There are some in the test below, so maybe that's what it tries to fix? In any case, I think we should add a t.Cleanup() for resetting the set{u,g}ids wherever we change them so that t is clear (i) why the happen, (ii) that the set values will be restored.

What I meant was that the t.Cleanup() should be added wherever we change the Setuid.

For example:

```go
oldGid := syscall.Getgid()
if err := syscall.Setgid(gid); err != nil {
  ...
}
t.Cleanup(func() {
    		if err = syscall.Setgid(oldGid); err != nil {
			t.Logf("Failed to restore gid to %d :  %v\n", oldGid, err)
		}
}) 
```

Above makes it clear what the cleanup functions fixes.

We can even have helper functions for this:

```go
func TestSetGID(t *testing.T, gid int)  {

    oldGid := syscall.Getgid()
   if err := syscall.Setgid(gid); err != nil {
      t.Fail("failed to set gid to %d", gid) // can also be t.Skip if we want to
   }

   t.Cleanup(func() {
    		if err = syscall.Setgid(oldGid); err != nil {
			t.Logf("Failed to restore gid to %d :  %v\n", oldGid, err)
		}
   }) 
}
```

```

kkourt (changes requested):
Thanks! 

I'm requesting changes only for the tests. Everything else is just comments, so there is no need to change anything.

> So I would have preferred to not use external binary but it gets complicated. 

Not sure if it was clear, but my suggestion was to use external binaries everywhere instead of changing things in the test binary and then calling `t.Cleanup()`. I think this would make things easier.

bpf/lib/bpf_cred.h
tixxdz: yes makes sense, those were internal vars but I renamed to be EXEC_SETUID_ROOT

bpf/process/bpf_execve_bprm_commit_creds.c
tixxdz: 1. ruid is the real user id that started the command so we check if it was not global root (this won't change for setuid execution it will keep its value)
2. Yes you could start with ruid == 0 as real root and execute a setuid binary to 1000 uid as an example, but for this case we will only report the setuid=1000 but we won't report the privilege `PRIVILEGES_RAISED_EXEC_FILE_SETUID` as it is more of a privilege drop, and such case usually don't happen in practice. It is more privilege raise to root. So that check is to ensure this logic.
3. For the commit_creds yes it happens later, we are in middle after the binary was parsed and its inode owners were read by kernel, then the capability LSM did some magic fixup to ensure everything is right, then our program is triggered just before the final commit_creds on the task cred, so we operate on the bprm (binary prameters) creds , and there won't be any change anymore what we are parsing is what will be committed in commit_creds

bpf/process/bpf_execve_bprm_commit_creds.c
tixxdz: Yes it can be different at the time of our hook, see this comment https://github.com/cilium/tetragon/pull/1786#discussion_r1447233686 same thing point 3

pkg/sensors/exec/exec_test.go
tixxdz: ok got you! will fix it

tixxdz: I did remove it, however we have been running tests with root for a long time, so if we ever change this these tests and some others may fail...

kkourt (approved):

bpf/process/bpf_execve_bprm_commit_creds.c
kkourt: > Yes you could start with ruid == 0 as real root and execute a setuid binary to 1000 uid as an example, 

What  would be the `ruid` and `uid` values in this case?

PR #1796, opened 2023-11-24
make:image: support arm64 cross compile

tixxdz:
This should support cross compile container image.

It does not use docker multiplatfor buildx
It uses the compiled arm64 bpftool with llvm , Todo: fix it so we use it only on cross compile in case bpftool still fails

The make tarball is just make image, when make image works then make tarball should work transparently, but in case with multiplatfor docker buildx, then we just need to select the right image with arch as suffix


tixxdz:
This is handled by https://github.com/cilium/tetragon/pull/1805

Issue #1778, opened 2023-11-21
Helm: Expose operator metrics


lambdanis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature you would like

Tetragon Operator exposes some default Prometheus metrics (Go & process) that can be used to monitor its health: https://github.com/cilium/tetragon/blob/3326a78fbb120a9f5946b64733f8786d08be4b0e/operator/cmd/serve/serve.go#L50)

However, the Helm chart doesn't contain a Service exposing the operator metrics port, so if the user wants to scrape the metrics, they have to scrape the pod directly. Let's add support for operator metrics in the Helm chart:
- [x] port in the operator Deployment + a Service exposing it
- [x] Helm value configuring the operator metrics port
- [x] optional ServiceMonitor scraping operator metrics (similar to [this one](https://github.com/cilium/tetragon/blob/a37a56fbb41e094d59c3bd5664a851532ea96e28/install/kubernetes/templates/servicemonitor.yaml) for the agent)
- [ ] Don't start the operator metrics server if `tetragonOperator.prometheus.enabled = false`

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

hungran:
I would like to work on this issue

lambdanis:
The first two points were implemented in #1797, there are two other items left. (In the initial issue I wrote `ServiceAccount` instead of `ServiceMonitor`, sorry :woman_facepalming: that was a mistake.)

@hungran If you want to work on them then feel free to open another PR, otherwise they can wait.

hungran:
> The first two points were implemented in #1797, there are two other items left. (In the initial issue I wrote `ServiceAccount` instead of `ServiceMonitor`, sorry 🤦‍♀️ that was a mistake.)

no worries

> @hungran If you want to work on them then feel free to open another PR, otherwise they can wait.

why not :)

hungran:
> Don't start the operator metrics server if tetragonOperator.prometheus.enabled = false

not really get the point on this? this should be a new flag that we can disable at here `operator/cmd/serve/serve.go`

lambdanis:
> > Don't start the operator metrics server if tetragonOperator.prometheus.enabled = false
> 
> @lambdanis not really get the point on this? should be a new flag that we can disable at here `operator/cmd/serve/serve.go`

Good question.
[...]

PR #1797, opened 2023-11-26
helm: adding optional to expose tetragon-operator metrics

hungran:
hey team, I'm new here so please correct me if I'm missing somewhere

This will fix the issue https://github.com/cilium/tetragon/issues/1778

 - [x] port in the operator Deployment + a Service exposing it
 - [x] Helm value configuring the operator metrics port
 - [ ] optional ServiceAccount scraping operator metrics following the issue -> not sure the 3rd point as I've seen we already had SA for [TetragonOperator](https://github.com/cilium/tetragon/blob/main/install/kubernetes/templates/operator_serviceaccount.yaml)
 
```release-note
Add a Kubernetes service exposing Tetragon Operator metrics. Make the operator metrics port configurable via Helm values and change the default to 2113.
```

### DIFF
--- a/install/kubernetes/templates/operator_service.yaml
+++ b/install/kubernetes/templates/operator_service.yaml
@@ -1,0 +1,19 @@
+{{- if .Values.tetragonOperator.prometheus.enabled -}}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  namespace: {{ .Release.Namespace }}
+  name: {{ .Release.Name }}-operator-metrics
+  labels: 
+    {{- include "tetragon-operator.labels" . | nindent 4 }}
+spec:
+  ports:
+    - name: metrics
+      port: {{ .Values.tetragonOperator.prometheus.port }}
+      targetPort: {{ .Values.tetragonOperator.prometheus.port }}
+      protocol: TCP
+  selector:
+    {{- include "tetragon-operator.selectorLabels" . | nindent 4 }}
+  type: ClusterIP
+{{- end }}
--- a/install/kubernetes/test.sh
+++ b/install/kubernetes/test.sh
@@ -9,10 +9,10 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 cd $SCRIPT_DIR
 
 alias helm='docker run --rm -v $(pwd):/apps alpine/helm:3.3.4'
-alias kubeval='docker run --rm -i garethr/kubeval:0.15.0@sha256:6962d8ecbb7839637667f66e6703e6ebaae0c29dfe93a31d9968fba4324c7b8d'
+alias kubeconform='docker run --rm -i ghcr.io/yannh/kubeconform:v0.6.4-alpine@sha256:e68a0b638c6e9b76f1b7d58b4ec94340ef3b6601db25b2e40b29e3ac2d68e4bf'
 helm dependency update .
 helm lint . --with-subcharts
-helm template tetragon . | kubeval --strict --additional-schema-locations https://raw.githubusercontent.com/joshuaspence/kubernetes-json-schema/master
+helm template tetragon . | kubeconform --strict --schema-location default
 
 # Update README.md.
 docker run --rm -v "$(pwd):/helm-docs" -u "$(id -u)" jnorwood/helm-docs:v1.11.0@sha256:66c8f4164dec860fa5c1528239c4aa826a12485305b7b224594b1a73f7e6879a
--- a/install/kubernetes/values.yaml
+++ b/install/kubernetes/values.yaml
@@ -217,7 +217,13 @@ tetragonOperator:
     # -- Enables the PodInfo CRD and the controller that reconciles PodInfo
     # custom resources.
     enabled: false
-
+  # -- Enables the Tetragon Operator metrics.
+  prometheus:
+    enabled: true
+    # -- The address at which to expose Tetragon Operator metrics. Set it to "" to expose on all available interfaces.
+    address: ""
+    # -- The port at which to expose metrics.
+    port: 2113
 # -- Tetragon event settings
 export:
   # "stdout". "" to disable.
--- a/docs/content/en/docs/reference/helm-chart.md
+++ b/docs/content/en/docs/reference/helm-chart.md
@@ -90,7 +90,7 @@ To use [the values available](#values), with `helm install` or `helm upgrade`, u
 | tetragon.prometheus.serviceMonitor.scrapeInterval | string | `"10s"` | Interval at which metrics should be scraped. If not specified, Prometheus' global scrape interval is used. |
 | tetragon.resources | object | `{}` |  |
 | tetragon.securityContext.privileged | bool | `true` |  |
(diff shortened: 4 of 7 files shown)


install/kubernetes/values.yaml
```diff
@@ -171,6 +171,11 @@ tetragonOperator:
     # -- Enables the PodInfo CRD and the controller that reconciles PodInfo
     # custom resources.
     enabled: false
+  # -- Enables the Tetragon Operator metrics.
+  metrics:
```
lambdanis, line 175: Could you name this section `prometheus` instead? I agree `metrics` makes sense, however, we already have `tetragon.prometheus` section for the Tetragon agent metrics, and IMO it's wroth keeping them consistent.

install/kubernetes/values.yaml
```diff
@@ -171,6 +171,11 @@ tetragonOperator:
...
+    enabled: true
+    # -- The port at which to expose metrics.
+    port: 2113
```
lambdanis, line 178: At the moment the [metrics port in the operator](https://github.com/cilium/tetragon/blob/3326a78fbb120a9f5946b64733f8786d08be4b0e/operator/cmd/serve/serve.go#L93) is by default set to 8080. The Helm chart default should be same IMO, but `2113` seems to me a better choice for the default, so maybe let's change it in the operator.

lambdanis (commented):
Thank you for tackling this @hungran. I left some comments about the Helm values.

install/kubernetes/values.yaml
hungran: thanks, let's go with `2113`

hungran:
indent level was re-corrected, @lambdanis might have chance to re-run the CI?

install/kubernetes/values.yaml
```diff
@@ -171,6 +171,11 @@ tetragonOperator:
     # -- Enables the PodInfo CRD and the controller that reconciles PodInfo
     # custom resources.
     enabled: false
+  # -- Enables the Tetragon Operator metrics.
+  prometheus:
+    enabled: true
+    # -- The port at which to expose metrics.
+    port: 2113
```
lambdanis, line 178: The port must be also passed to the operator in the [operator configmap](https://github.com/cilium/tetragon/blob/e287ce8957d5ed0439aa8a9d79df30b42629bc9b/install/kubernetes/templates/operator_configmap.yaml), as `metrics-bind-address`. Otherwise if you don't use the default, the k8s service won't match the port actually used by the operator pod.

install/kubernetes/values.yaml
```diff
@@ -171,6 +171,11 @@ tetragonOperator:
     # -- Enables the PodInfo CRD and the controller that reconciles PodInfo
     # custom resources.
     enabled: false
+  # -- Enables the Tetragon Operator metrics.
+  prometheus:
+    enabled: true
```
lambdanis, line 176: I see this value doesn't affect whether the operator runs the metrics server or not, only whether the k8s service is created or not. It would be nice to propagate it into the operator code too, and don't run the metrics server if it's set to false. But I think this can be done as a follow-up, not necessarily in this PR.

lambdanis (changes requested):

lambdanis:
@hungran Thanks for the update, I requested one more change (to pass port to the operator via the configmap).

Also, please run `cd install/kubernetes && ./test.sh` to generate Helm docs. Otherwise the CI will fail like here: https://github.com/cilium/tetragon/actions/runs/7014110773/job/19099622794

Could you also squash the commits when you make changes? Thanks!

install/kubernetes/values.yaml
hungran: that's better though, however, we change the default port at `operator/cmd/serve/serve.go` believed that some change at there too that we can serve the port from config too? or am i missing something

install/kubernetes/values.yaml
hungran: um agreed, this be done as follow-up, might update the comment in values to be easier track at that level though

install/kubernetes/test.sh
```diff
@@ -9,10 +9,10 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 cd $SCRIPT_DIR
 
 alias helm='docker run --rm -v $(pwd):/apps alpine/helm:3.3.4'
-alias kubeval='docker run --rm -i garethr/kubeval:0.15.0@sha256:6962d8ecbb7839637667f66e6703e6ebaae0c29dfe93a31d9968fba4324c7b8d'
+alias kubeconform='docker run --rm -i ghcr.io/yannh/kubeconform:v0.6.4-alpine'
 helm dependency update .
 helm lint . --with-subcharts
-helm template tetragon . | kubeval --strict --additional-schema-locations https://raw.githubusercontent.com/joshuaspence/kubernetes-json-schema/master
+helm template tetragon . | kubeconform --strict --schema-location default
```
hungran, line 15: since kubeval no longer supported, would be better if the chart test use [kubeconform](https://github.com/yannh/kubeconform), an alternative solution

hungran (commented):
@lambdanis  not sure why must pass port to configmap as `metrics-bind-address` as we changed the default value to `2113` or here we can have a way that cmd.Flag() could have metrics-bind-address from config file (config map)? is this idea here?

install/kubernetes/test.sh
lambdanis: Nice, thanks!

install/kubernetes/test.sh
```diff
@@ -9,10 +9,10 @@ SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 cd $SCRIPT_DIR
 
 alias helm='docker run --rm -v $(pwd):/apps alpine/helm:3.3.4'
-alias kubeval='docker run --rm -i garethr/kubeval:0.15.0@sha256:6962d8ecbb7839637667f66e6703e6ebaae0c29dfe93a31d9968fba4324c7b8d'
+alias kubeconform='docker run --rm -i ghcr.io/yannh/kubeconform:v0.6.4-alpine'
```
lambdanis, line 12: I think we'll still want to pin the docker image sha256

hungran:
> > not sure why must pass port to configmap as metrics-bind-address as we changed the default value to 2113 or here we can have a way that cmd.Flag() could have metrics-bind-address from config file (config map)? is this idea here?
> 
> The operator configmap is mounted in the operator deployment, and the flags are read from there already. So to make it possible to use a non-default port for the metrics server, we need to pass a flag via the configmap.

Hi @lambdanis  Correct me if I'm wrong, the current TetragonOperator not yet read that flag though?
I've try to used image the v1.0.0 from the hub and also built on my local and tried to deploy by difference port along with the configmap `metrics-bind-address`. However, seem like the metrics server still bind to the default port by cmd Flags as hardcode under `operator/cmd/serve/serve.go` https://github.com/cilium/tetragon/blob/564b819ab6cbe39978c65591c84fdbdf82239d89/operator/cmd/serve/serve.go#L93

install/kubernetes/test.sh
hungran: done

install/kubernetes/templates/operator_service.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels: 
+    {{- include "tetragon-operator.labels" . | nindent 4 }}
+spec:
+  ports:
+    - name: metrics
+      port: {{ .Values.tetragonOperator.prometheus.port }}
+      targetPort: {{ .Values.tetragonOperator.prometheus.port }}
+      protocol: TCP
+  selector:
+    {{- include "tetragon-operator.selectorLabels" . | nindent 6 }}
```
lambdanis, line 17:
```suggestion
    {{- include "tetragon-operator.selectorLabels" . | nindent 4 }}
```
too much indent

lambdanis (changes requested):
One indent in the service seems too big, but other than that changes look good :)

I see that the commit message contains multiple sign-offs. Could you amend it to keep only one at the end?

install/kubernetes/templates/operator_service.yaml
hungran: solved, thanks

hungran:
> One indent in the service seems too big, but other than that changes look good :)
> 
> I see that the commit message contains multiple sign-offs. Could you amend it to keep only one at the end?

ah right, done

hungran:
> @hungran It looks like you need to run `install/kubernetes/test.sh` once again, sorry.

no worries, a new local environment not yet have `gnu-sed` so the `install/kubernetes/test.sh` was failed btw :) updated

lambdanis (approved):
Thank you @hungran!

PR #1805, opened 2023-11-27
Make tarball target work with arm64 cross-compilation and simplify Makefile/Dockerfile

mtardy:
I tested on my amd64 VM and it generated the correct binaries for each architecture with cross-compilation. Can you check as well that it works for you with:

```shell-session
# on an amd64 machine, first should generate the amd64 tarball and second arm64 tarball
make tarball 
make TARGET_ARCH=arm64 tarball
```

```release-note
Add arm64 support for the tarball via cross-compilation on amd64.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -300,7 +300,7 @@ image-clang:
 .PHONY: tarball tarball-release tarball-clean
 # Share same build environment as docker image
 tarball: tarball-clean image
-	$(CONTAINER_ENGINE) build --build-arg TETRAGON_VERSION=$(VERSION) --build-arg TARGET_ARCH=$(TARGET_ARCH) -f Dockerfile.tarball -t "cilium/tetragon-tarball:${DOCKER_IMAGE_TAG}" .
+	$(CONTAINER_ENGINE) build --build-arg TETRAGON_VERSION=$(VERSION) --build-arg TARGET_ARCH=$(TARGET_ARCH) -f Dockerfile.tarball -t "cilium/tetragon-tarball:${DOCKER_IMAGE_TAG}" --platform=linux/${TARGET_ARCH} .
 	$(QUIET)mkdir -p $(BUILD_PKG_DIR)
 	$(CONTAINER_ENGINE) save cilium/tetragon-tarball:$(DOCKER_IMAGE_TAG) -o $(BUILD_PKG_DIR)/tetragon-$(VERSION)-$(TARGET_ARCH).tmp.tar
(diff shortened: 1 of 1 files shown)


mtardy:
> Hi @mtardy thank you for handling this.
> 
> 
> 
> I tested it , make tarball then make TARGET_ARCH=arm64 tarbal, it fails with:
> 
> ```
> 
> => CANCELED [tetragon-builder 3/4] COPY . .                                                                                                                                                           8.6s
> 
> ------                                                                                                                                                                                                      
> 
>  > [base-build 2/7] RUN apk add iproute2:                                                                                                                                                                   
> 
> 1.437 exec /bin/sh: exec format error
> 
> ------
> 
> Dockerfile:72
> 
> --------------------
> 
>   70 |     # retrieves (cross-)compiled binaries from builders
> 
>   71 |     FROM docker.io/library/alpine:3.18.4@sha256:eece025e432126ce23f223450a0326fbebde39cdf496a85d8c016293fc851978 as base-build
> 
>   72 | >>> RUN apk add iproute2
> 
>   73 |     RUN mkdir /var/lib/tetragon/ && \
> 
>   74 |         mkdir -p /etc/tetragon/tetragon.conf.d/ && \
> 
> --------------------
> 
> ERROR: failed to solve: process "/bin/sh -c apk add iproute2" did not complete successfully: exit code: 1
> 
> make: *** [Makefile:275: image] Error 1
> 
> ```
> 
> 
> 
> So it needs arm emulation here, if there is no easy way and it is complicated, then would it be possible at the last stage of copying the binaries make targetplatform same as buildplatform? so we run in amd64 context and we copy binaries from other arm64 builds?
> 
> 
> 
> Also we can add an environment variable TARBALL_CROSS_COMPILE  , so if set we use the above where targetplatform becomes buildplatform as the aim is to just have the arm binaries copied into the final one layer image. We don't want to run that image... just use it to dump into tarball. If TARBALL_CROSS_COMPILE is not set then no change current solution is perfect.
> 
> 

You just need to install qemu-user-static on your amd64 machine.

tixxdz (approved):

mtardy:
> Thank you Mahé, yes it seems the arm emulation got better, it used to fail past years due to some complex instructions... anyway just tested tetra arm64 it works, didn't test daemon.
> 
> Could you please test the generated tarball? and as a later step I will add a ci job for it.
> 
> Feel free to merge ;-)

I tested it manually on my arm64 VM and it worked fine. Thanks let's merge this.

PR #1806, opened 2023-11-27
String matching: match long strings

kevsecurity:
The current implementation ignore prefix and postfix matching on strings that are longer than the prefix max or postfix max (accordingly).

These patches modify the logic to handle strings that are longer than these maximums by restricting the string to the maximum permitted by the LPM TRIE maps – from the beginning for prefix and from the end for postfix.

```release-note
Fixes prefix and postfix matching for strings longer than the prefix or postfix maximum length
```


kkourt (dismissed):
Thanks, the code changes look good to me, but can you also add some tests for these specific cases?

kkourt:
Also, I believe we need to backport this to 1.0. I've added a label.

tpapagian (approved):
LGTM, thanks!

Issue #1761, opened 2023-11-16
Add custom annotations and labels to tetragon operator


vladimirkus:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

I encountered a limitation with my ability configuring annotations and labels for the tetragon-operator workload. Currently helm-chart doesn't support custom annotations and labels, however I would like to have this ability, due to the fact that in my infrastructure, I can configure prometheus scraping and log shipment only by adding annotations and labels.

### Describe the feature you would like

I would like to have an ability to add custom annotations and labels to the tetragon-operator pods and deployment, the same as it already implemented for the daemonset pods:
https://github.com/cilium/tetragon/blob/v1.0/install/kubernetes/values.yaml#L8
https://github.com/cilium/tetragon/blob/v1.0/install/kubernetes/values.yaml#L21

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kkourt:
Thanks, above make sense to me.


I think the best approach for this would be to:

1. Apply the `podAnnotations` in both the daemonset and the operator deployment pods
2. Have specific annotations for tetragon daemonset set under `tetragon` in the chart (`tetragon.podAnnotations`)
3. Have specific annotations for the operator deployment set under `tetragonOperator` in  chart (`tetragonOperator.podAnnotations`)

Thoughts? Would doing just (1) solve your specific issue?

Also, Would you be interested in implementing this?

Thanks,
Kornilios.

PR #1817, opened 2023-11-30
Helm: Added various configuration options for the Tetragon Operator Deployment

PhilipSchmid:
I have added multiple configuration options for the Tetragon Operator Deployment, which are often crucial for enterprise-grade deployments that need to happen within certain security boundaries.

According to my current understanding, this should not break anything for existing users. The only chance I see a breakage happen is when somebody already configured `.Values.serviceAccount.name` in the past. The thing is, this value was never considered, and the SA was simply named `{{ .Release.Name }}` or `{{ .Release.Name }}-operator-service-account`. With this change, it would suddenly be considered for the actual SA naming.

- helm: Added extraVolumes support for the Tetragon Operator
- helm: Added extra labels for the Tetragon Operator Deployment and Pods
- helm: Added configurable annotations for the Tetragon Operator Deployment and Pods
- helm: Made the placement of Tetragon Operator Deployment Pods configurable
- helm: Made update strategy configurable for the Tetragon Operator Deployment 
- helm: Made resource configurable for the Tetragon Operator Deployment
- helm: Added pod/container securityContext to the Tetragon Operator Deployment
- helm: Added priorityClassName to the Tetragon Operator Deployment
- helm: Added imagePullSecret to the Tetragon Operator Deployment

Fixes: https://github.com/cilium/tetragon/issues/1761

### DIFF
--- a/install/kubernetes/values.yaml
+++ b/install/kubernetes/values.yaml
@@ -168,9 +168,10 @@ tetragonOperator:
   annotations: {}
   # -- Annotations for the Tetragon Operator Deployment Pods.
   podAnnotations: {}
-  annotations: {}
-  # -- Annotations for the Tetragon Operator Deployment Pods
-  podAnnotations: {}
+  # -- Extra labels to be added on the Tetragon Operator Deployment.
+  extraLabels: {}
+  # -- Extra labels to be added on the Tetragon Operator Deployment Pods.
+  extraPodLabels: {}
   # -- priorityClassName for the Tetragon Operator Deployment Pods.
   priorityClassName: ""
   # -- tetragon-operator service account.
--- a/install/kubernetes/values.yaml
+++ b/install/kubernetes/values.yaml
@@ -1,7 +1,10 @@
+# -- Global settings
 enabled: true
-imagePullPolicy: IfNotPresent
-priorityClassName: ""
 imagePullSecrets: []
+
+# -- Tetragon agent settings
+priorityClassName: ""
+imagePullPolicy: IfNotPresent
 serviceAccount:
   create: true
   annotations: {}
@@ -158,6 +161,8 @@ tetragon:
   # host, the path is /proc. Exceptions to this are environments like kind, where the runtime itself
   # does not run on the host.
   hostProcPath: "/proc"
+
+# -- Tetragon Operator settings
 tetragonOperator:
   # -- tetragon-operator service account.
   serviceAccount:
@@ -176,6 +181,8 @@ tetragonOperator:
     # -- Enables the PodInfo CRD and the controller that reconciles PodInfo
     # custom resources.
     enabled: false
+
+# -- Tetragon event settings
 export:
   # "stdout". "" to disable.
   mode: "stdout"
--- a/install/kubernetes/values.yaml
+++ b/install/kubernetes/values.yaml
@@ -164,6 +164,8 @@ tetragon:
 
 # -- Tetragon Operator settings
 tetragonOperator:
+  # -- priorityClassName for the Tetragon Operator Deployment Pods.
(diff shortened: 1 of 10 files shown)


install/kubernetes/values.yaml
```diff
@@ -158,19 +161,64 @@ tetragon:
...
+# -- Tetragon Operator settings
 tetragonOperator:
+  # -- Annotations for the Tetragon Operator Deployment.
+  annotations: {}
+  # -- Annotations for the Tetragon Operator Deployment Pods.
+  podAnnotations: {}
+  # -- Extra labels to be added on the Tetragon Operator Deployment.
+  extraLabels: {}
+  # -- Extra labels to be added on the Tetragon Operator Deployment Pods.
+  extraPodLabels: {}
```
lambdanis, line 174: The agent labels are configured with:
```
podLabels: {}
daemonSetLabelsOverride: {}
selectorLabelsOverride: {}
podLabelsOverride: {}
serviceLabelsOverride: {}
```
([L20-L24](https://github.com/cilium/tetragon/pull/1817/files#diff-be0f5319dfa71d552848e99593254e89869fdf280b90695ef6fb645544212538L20-L24)). `podLabels` is adding extra labels to pods and it seems there is no option to add extra labels to the daemonset, but there is `daemonSetLabelsOverride` to override all labels.

It bothers me a bit that these values are so inconsistent :( But I like yours more, they're more intuitive and I think more consistent with the whole ecosystem. So IMO it's fine to leave it like that, just wanted to point it out in case you think it's something worth consideration.

lambdanis (approved):
This is great, thank you.

I left one comment about labels configuration, but more like food for thought.

install/kubernetes/values.yaml
PhilipSchmid: Good point. I also saw those labels on the DS but used a different approach for the operator as it wasn't intuitive IMO. For example,
1. What's the reason why somebody would want to override selector labels within the DS (`selectorLabelsOverride`)?
2. One can configure `podLabels` and `podLabelsOverride` simultaneously, and the provided values will all be added. I mean, where's the override here? 😅
3. There's no way to add additional labels to the DS. You can either specify `daemonSetLabelsOverride` which leads to the situation that all (quite important) labels like `helm.sh/chart` and `app.kubernetes.io/*` won't be added or you can have only have those (`helm.sh/chart` and `app.kubernetes.io/*`).

I've got the feeling it would make sense to refactor the DS labels but most likely via separate PR and only with the next minor release as I'd consider this a (small) breaking change.

lambdanis:
@PhilipSchmid I see the CI is failing:
* The generated docs are not up to date - `./install/kubernetes/test.sh` should do the trick
* The commit messages are too long

Could you fix these?

PhilipSchmid:
@lambdanis I just shortened the commit messages and checked that the `.test.sh` script no longer generates new changes. Unfortunately, the CI is now even more unhappy with this PR 😆.

lambdanis:
Oh dear, it looks like some Quay hiccup maybe? The CI failures definitively don't looks related to the changes. I'll rerun.

lambdanis:
> checked that the `.test.sh` script no longer generates new changes

Ok, I run on your branch and it looks like the trailing space [here](https://github.com/cilium/tetragon/blob/6ab6d21ea7633747a69b17926f04172a3e2e4ed9/install/kubernetes/export-doc.sh#L42) is the problem. I suppose your editor is removing trailing spaces automatically? If you could remove this space from the script, then it should pass :) (a bit surprising that nobody hit it before tbh)

lambdanis:
After some flakes the CI finally passes, so I'm merging it, thank you @PhilipSchmid!

PR #1827, opened 2023-12-01
Add masOS compatibility for 'sed' usage in scripts

PhilipSchmid:


api/export-doc.sh
```diff
@@ -15,15 +15,18 @@ cp v1/README.md $TMP_FILE
...
-sed -i -e '/## Table of Contents/,/##/{//!d}' $TMP_FILE
-sed -i '/^## Table of Contents/d' $TMP_FILE
+sed -i'' -e '/## Table of Contents/,/##/{//!d}' $TMP_FILE
+sed -i'' -e '/^## Table of Contents/d' $TMP_FILE
 # cleanup unecessary consecutive whitelines
-sed -i '/^[[:space:]]*$/N;/^[[:space:]]*\n[[:space:]]*$/D' $TMP_FILE
+sed -i'' -e '/^[[:space:]]*$/N;/^[[:space:]]*\n[[:space:]]*$/D' $TMP_FILE
+
+# Cleanup backup file that might be created by macOS sed:
+rm ${TMP_FILE}-e || true
```
mtardy, line 29: So I did not experience the backup file that might be created, if that happens, are you sure about its name (the `-e` suffix)?

Anyway could you suppress the error from rm `rm: cannot remove '/tmp/tmp.V2HaNn1uox-e': No such file or directory` because it looks something wrong happened

```suggestion
rm ${TMP_FILE}-e 2>/dev/null || true
```

install/kubernetes/export-doc.sh
```diff
@@ -15,18 +15,20 @@ cp README.md $TMP_FILE
...
+sed -i'' -e '/^----------------------------------------------$/d' $TMP_FILE
 # remove autogenerated copyright
-sed -i '/^Autogenerated from chart metadata using \[helm-docs /d' $TMP_FILE
+sed -i'' -e '/^Autogenerated from chart metadata using \[helm-docs /d' $TMP_FILE
 # cleanup unecessary consecutive whitelines
-sed -i '/^[[:space:]]*$/N;/^[[:space:]]*\n[[:space:]]*$/D' $TMP_FILE
+sed -i'' -e '/^[[:space:]]*$/N;/^[[:space:]]*\n[[:space:]]*$/D' $TMP_FILE
 
+# Cleanup backup file that might be created by macOS sed:
+rm ${TMP_FILE}-e || true
```
mtardy, line 31: same here

api/export-doc.sh
PhilipSchmid: > So I did not experience the backup file that might be created

On macOS or Linux? I've tested on macOS and did not get this `rm: cannot remove` error. Whatever, I've added the `2>/dev/null` as I think it makes sense either way. Thanks!

mtardy:
empty line issue, just run the script locally and push!

mtardy (changes requested):
previous
```
# cleanup unecessary consecutive whitelines
sed -i '/^[[:space:]]*$/N;/^[[:space:]]*\n[[:space:]]*$/D' $TMP_FILE
```
was working on macOS and Linux for me.

It doesn't work for you? can you dump your sed version maybe?

PhilipSchmid:
```bash
# cleanup unecessary consecutive whitelines
sed -i'' -e '/^[[:space:]]*$/N;/^[[:space:]]*\n[[:space:]]*$/D' $TMP_FILE
```

It works and does what it should do (`cleanup unecessary consecutive whitelines`) but there's no empty line at the end of the `docs/content/en/docs/reference/helm-chart.md` file on macOS, while there is one Linux. Hence, I'm still searching for a solution that produces the same output for both.

mtardy (changes requested):
this doesn't run the same in the CI and on your platform since `git status` returns something. You can't reproduce?

PhilipSchmid:
Rebased to `main`.

@mtardy Exactly, I'm not able to reproduce this locally. `cd install/kubernetes && ./test.sh` doesn't produce any changes on my macOS or my Linux server 🤷‍♂️.

jrfastab:
going to give this a small nudge. Would be nice if MacOS works also.

mtardy (approved):
Thanks Philip, sorry for the length but those kinds of compatibility things are a bit annoying as you saw 😛!

I think the generated file test passes now, don't know what changed but it looks good! Let's merge this when the whole CI passes.

install/kubernetes/export-doc.sh
```diff
@@ -42,8 +46,8 @@ This page was generated with github.io/cilium/tetragon/install/kubernetes/tetrag
 please do not edit directly.
 {{< /comment >}}
 
-The Tetragon Helm chart source is available under 
-[github.io/cilium/tetragon/install/kubernetes/tetragon](https://github.com/cilium/tetragon/tree/main/install/kubernetes/tetragon)
+The Tetragon Helm chart source is available under
+[github.io/cilium/tetragon/install/kubernetes](https://github.com/cilium/tetragon/tree/main/install/kubernetes)
```
mtardy, line 50: Oh lookout for this before merging, this got upgraded and this is reversing the change

PhilipSchmid: Good catch! Updated, thanks!

mtardy:
Thanks!! 🎉

PR #1830, opened 2023-12-02
Generate Helm values.yaml from a template

ghost:


lambdanis:
You know what, I changed my mind, I'll close this PR and maybe open another one with parts of it. Initially I wanted to template the Helm chart in a similar way as the Cilium Helm chart, but then I realized it's not needed and had a different idea.

PR #1832, opened 2023-12-04
linters: add makezero and fix a related bug

mtardy:
Makezero is a linter to ensure that slices are not initialized with non-zero length. Since in Go, we most of the time use append on the slice, initializing it with non-zero length can tend to create a slice full of zero that will clobber the beginning of the slice.

In the bug fixed here, we created a list with 300+ zeros and then append after them, only to skip the zero in the calling method. Usually we want to prealloc the slice with a capacity of len but a length of 0, i.e. replace make([]int, len(nums)) with make([]int, 0, len(nums)).  See more info here https://github.com/ashanbrown/makezero.

### DIFF
--- a/pkg/btf/validation.go
+++ b/pkg/btf/validation.go
@@ -283,12 +283,12 @@ func AvailableSyscalls() ([]string, error) {
 	}
 
 	ret := []string{}
-	for _, key := range syscallinfo.SyscallsNames() {
-		if key == "" {
-			continue
+	for key, value := range syscallinfo.SyscallsNames() {
+		if value == "" {
+			return nil, fmt.Errorf("syscall name for %q is empty", key)
 		}
 
-		sym, err := arch.AddSyscallPrefix(key)
+		sym, err := arch.AddSyscallPrefix(value)
 		if err != nil {
 			return nil, err
 		}
@@ -298,7 +298,7 @@ func AvailableSyscalls() ([]string, error) {
 			continue
 		}
 
-		ret = append(ret, key)
+		ret = append(ret, value)
 	}
 
 	return ret, nil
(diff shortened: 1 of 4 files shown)


pkg/btf/validation.go
```diff
@@ -325,10 +325,6 @@ func GetSyscallsList() ([]string, error) {
 	for _, key := range syscallinfo.SyscallsNames() {
 		var fn *btf.Func
 
-		if key == "" {
```
kevsecurity, line 328: Why is this check being removed?

mtardy: In the bug fixed here, we created a list with 300+ zeros and then append after them, only to skip the zero in the calling method.

mtardy: this is a workaround for the bug introduced that produces a slice two times "too big" with plenty of zeros

kevsecurity: I think I'd prefer (especially as it was catching a bug previously) that we keep the check but return an error instead of 'continue'. That way, if it ever happens then we should see it fail, instead of upsetting the rest of the loop.

mtardy: So basically the previous code was doing that, and doing "continue" on zero: https://go.dev/play/p/Ek0lSbDg3UQ.

The fix does it correctly: https://go.dev/play/p/aNxMhmxSnRQ.

Did you take a look at the `SyscallsNames()` function? There is no way for it to return an entry with an empty string, it was just put because the author did not understand why there was empty string because of the error in the creation of the slice. The `syscallNames` map does not contain an empty key, it would be a bug in the list.

```go
func SyscallsNames() []string {
	ret := make([]string, 0, len(syscallNames))

	for _, v := range syscallNames {
		ret = append(ret, v)
	}
	return ret
}
```

kevsecurity: > There is no way

***Today*** There is no way...

mtardy: Those are not even keys, those are the values that are misnamed, the map is int[string]

kevsecurity (approved):
LGTM. Like the variable renaming too!

PR #1833, opened 2023-12-04
tetragon: Detect large program with feature check

olsajiri:
currently we detect large programs with kernel version check (>=5.3.0)
but some kernels (rhel) have many backports on top of older versions

adding large program feature check and using it to load large programs


kkourt (approved):
Detecting features instead of relying on kernel version is a much better approach, thanks!

PR #1838, opened 2023-12-04
tracingpolicy: add a stack traces example

mtardy:


examples/tracingpolicy/stack_traces.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "stack-traces-example"
+spec:
+  kprobes:
+    - call: kfree_skb_reason
+      syscall: false
+      selectors:
```
kevsecurity, line 9: Suggest adding arg 0 (skb) and arg 1 (int = reason) to provide some extra context for what the stack trace is related to. It's great to have a stack trace, but it can also help to link to actual packets.

mtardy: good idea

PR #1842, opened 2023-12-05
Tetragon oci hook setup

kkourt:
Add an init container for installing the tetragon OCI hook.
This, currently, only supports `cri-o`, but it can be extended to support containerd's NRI: https://github.com/containerd/containerd/blob/main/docs/NRI.md.

See for a developer demo: https://github.com/cilium/tetragon/blob/pr/kkourt/tetragon-oci-hook-setup/contrib/rthooks/tetragon-oci-hook/docs/demo.md.

```release-note:
helm: add init container to install tetragon-oci-hook in cri-o setups
```


lambdanis (approved):

install/kubernetes/templates/_container_oci_hooks_setup.tpl
```diff
@@ -0,0 +1,19 @@
+{{- define "container.tetragon-oci-hook-setup" -}}
+- name: oci-hook-setup
+  securityContext:
+    privileged: true
```
lambdanis, line 4: Could it reuse `{{- toYaml .Values.tetragon.securityContext | nindent 4 }}`? (it's `privileged: true` by default)

install/kubernetes/templates/_container_oci_hooks_setup.tpl
```diff
@@ -0,0 +1,19 @@
+{{- define "container.tetragon-oci-hook-setup" -}}
```
lambdanis, line 1: Could we also define `extraVolumeMounts` and `resources` for this container, configurable via Helm values? Similarly like for the tetragon container.

lambdanis (commented):
I left some comments re container template.

Also, I caused some conflicts with https://github.com/cilium/tetragon/pull/1845, sorry :) I essentially moved the Helm chart into tetragon directory, so hopefully they should be straightforward to resolve.

install/kubernetes/templates/_container_oci_hooks_setup.tpl
kkourt: I ended up defining a different context for the init container. Part of the reason we want the init container is that it needs certain permissions to install the hooks on the host. So I think it would be useful for it to have a different security context that the agent.

install/kubernetes/templates/_container_oci_hooks_setup.tpl
kkourt: Done, thanks!

lambdanis (approved):

Makefile
```diff
@@ -179,6 +180,12 @@ tetragon-bench:
 tetragon-operator:
 	$(GO_BUILD) -o $@ ./operator
 
+tetragon-oci-hook:
```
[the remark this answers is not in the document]
kkourt: Yap, it's that the option are different and we need the hook binary for the ARM builds as well.
The `contrib/rthooks/tetragon-oci-hook` Makefile is fairly simple (meant for development) and does not support cross-compiling.

PR #1845, opened 2023-12-05
Move Helm chart into tetragon directory

lambdanis:
See commits for details (and let me know if I should squash some of them). It's done mainly to make tooling (scripts) separate from the Helm chart, and also for consistency with other projects.


lambdanis:
The links-checker CI workflow fails, because https://github.com/cilium/tetragon/tree/main/install/kubernetes/tetragon will be a thing only after this PR is merged. Maybe this doc should point to a tag instead of main branch and get updated on release. @mtardy wdyt?

willfindlay (approved):

mtardy:
> The links-checker CI workflow fails, because https://github.com/cilium/tetragon/tree/main/install/kubernetes/tetragon will be a thing only after this PR is merged. Maybe this doc should point to a tag instead of main branch and get updated on release. @mtardy wdyt?

This kind of links (that will be created by the PR) will always fail surely. What I could do is to replace `https://github.com/cilium/tetragon/tree/main/` with `https://github.com/cilium/tetragon/tree/<branch for PR>/` in the link checker (and simplify the awful JSON output as well). Let me create an issue.

install/kubernetes/export-doc.sh
```diff
@@ -47,7 +47,7 @@ The Tetragon Helm chart source is available under
 and is distributed from the Cilium helm charts repository [helm.cilium.io](https://helm.cilium.io).
 
 To deploy Tetragon using this Helm chart you can run the following commands:
-```shell-session
+```shell
```
mtardy, line 50: Thanks! well spotted, the new theme uses `shell` everywhere, really well spotted it just changed.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+/install/kubernetes/tetragon/README.md linguist-generated
+/docs/content/en/docs/reference/helm-chart.md linguist-generated
```
mtardy, line 2: wow that's amazing, TIL! I explicitly excluded myself has a code owner from those files but I guess this is even better?

.github/ISSUE_TEMPLATE/release_template.md
```diff
@@ -18,9 +18,7 @@ assignees: ''
 
       git checkout -b pr/prepare-$RELEASE
       ./contrib/update-helm-chart.sh $RELEASE
-      pushd install/kubernetes/tetragon
-      ./test.sh
-      popd
+      ./install/kubernetes/test.sh
```
mtardy: Indeed this is no longer needed, you saw this https://github.com/cilium/tetragon/commit/357206b421615ad16a46da821582ac0dd00d21d0, you have 👀  everywhere.

mtardy (approved):
Impressed (see comments)! Thanks for the rename, hopefully nothing was forgotten and this is all smooth!

install/kubernetes/export-doc.sh
lambdanis: heh, `install/kubernetes/tetragon/test.sh` was actually failing without this change :)

lambdanis:
I'm merging it, the links-checker CI job should pass on main (see https://github.com/cilium/tetragon/issues/1858). https://github.com/cilium/charts/pull/88 needs to follow.

PR #1846, opened 2023-12-05
helm: Add extensions templates

lambdanis:
The goal is to make it easy to build custom distributions of Tetragon Helm chart.

### DIFF
--- a/install/kubernetes/tetragon/templates/_container_tetragon.tpl
+++ b/install/kubernetes/tetragon/templates/_container_tetragon.tpl
@@ -51,6 +51,7 @@
       mountPath: {{ .mountPath }}
       readOnly: {{ .readOnly }}
 {{- end }}
+    {{- include "tetragon.volumemounts.extra" . | nindent 4 }}
   env:
     - name: NODE_NAME
       valueFrom:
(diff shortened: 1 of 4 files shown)


install/kubernetes/tetragon/templates/_container_tetragon.tpl
```diff
@@ -51,6 +51,7 @@
       mountPath: {{ .mountPath }}
       readOnly: {{ .readOnly }}
 {{- end }}
+    {{- include "tetragon.volumemounts.extra" . | nindent 4 }}
```
chancez, line 54: Does it make sense to pass the output of `include` to `toYaml` before indenting? eg:
```
 {{- include "tetragon.volumemounts.extra" . | toYaml | nindent 4 }}
 ``

lambdanis: I would expect these extensions to be yaml already, but I guess it won't hurt

lambdanis: No, `toYaml` converts templates to string literals actually. I think it's less error-prone to define yaml correctly in the extension and avoid complicating it here.

kkourt (approved):

PR #1848, opened 2023-12-06
Remove pod.labels field

michi-covalent:
The pod.labels field got deprecated in v1.0. Let's remove it for v1.1.

Ref: #1363


jrfastab:
Yes agree it doesn't scale. @lambdanis  wdyt we talked about making it configurable why not just remove it? We can add it back I guess when we have mechanism to configure.

michi-covalent:
> I mean labels are still in the pod.pod_labels field right?

yes this is correct.

PR #1853, opened 2023-12-06
tetragon: Detect and use disassociate_ctty as exit probe if needed

olsajiri:
Adding support to use disassociate_ctty exit probe in case
acct_process is missing in kernel.

### DIFF
--- a/bpf/process/bpf_exit.c
+++ b/bpf/process/bpf_exit.c
@@ -43,10 +43,28 @@ char _license[] __attribute__((section("license"), used)) = "Dual BSD/GPL";
  * we are the last one of the thread group.
  */
 __attribute__((section("kprobe/acct_process"), used)) int
-event_exit(struct pt_regs *ctx)
+event_exit_acct_process(struct pt_regs *ctx)
 {
 	__u64 pid_tgid = get_current_pid_tgid();
 
 	event_exit_send(ctx, pid_tgid >> 32);
+	return 0;
+}
+
+/*
+ * Hooking on acct_process kernel function, which is called on the task's
+ * exit path once the task is the last one in the group. It's stable since
+ * v4.19, so it's safe to hook for us.
+ *
+ * It's called with on_exit argument != 0 when called from do_exit
+ * function with same conditions like for acct_process described above.
+ */
+__attribute__((section("kprobe/disassociate_ctty"), used)) int
+event_exit_disassociate_ctty(struct pt_regs *ctx)
+{
+	int on_exit = (int)PT_REGS_PARM1_CORE(ctx);
+
+	if (on_exit)
+		event_exit_send(ctx, get_current_pid_tgid() >> 32);
 	return 0;
 }
--- a/pkg/grpc/tracing/tracing.go
+++ b/pkg/grpc/tracing/tracing.go
@@ -278,6 +278,7 @@ func GetProcessKprobe(event *MsgGenericKprobeUnix) *tetragon.ProcessKprobe {
 		kernelSymbols, err := ksyms.KernelSymbols()
 		if err != nil {
 			logger.GetLogger().WithError(err).Warn("stacktrace: failed to read kernel symbols")
+			continue
 		}
 		fnOffset, err := kernelSymbols.GetFnOffset(addr)
 		if err != nil {
--- a/pkg/sensors/base/base.go
+++ b/pkg/sensors/base/base.go
@@ -4,7 +4,12 @@
 package base
 
 import (
+	"log"
+	"sync"
+
 	"github.com/cilium/tetragon/pkg/kernels"
+	"github.com/cilium/tetragon/pkg/ksyms"
+	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/sensors"
 	"github.com/cilium/tetragon/pkg/sensors/program"
 )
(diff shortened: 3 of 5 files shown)


pkg/grpc/tracing/tracing.go
```diff
@@ -278,6 +278,7 @@ func GetProcessKprobe(event *MsgGenericKprobeUnix) *tetragon.ProcessKprobe {
 		kernelSymbols, err := ksyms.KernelSymbols()
 		if err != nil {
 			logger.GetLogger().WithError(err).Warn("stacktrace: failed to read kernel symbols")
+			continue
```
tixxdz, line 281: I think we are using this in other versions... maybe worth a backport? @mtardy @kkourt

bpf/process/bpf_exit.c
```diff
@@ -43,10 +43,28 @@ char _license[] __attribute__((section("license"), used)) = "Dual BSD/GPL";
  * we are the last one of the thread group.
  */
 __attribute__((section("kprobe/acct_process"), used)) int
-event_exit(struct pt_regs *ctx)
+event_exit_acct_process(struct pt_regs *ctx)
```
tixxdz, line 46: While you are it Jiri would it be nice if you can rename it to tg_kp_exit_acct_process and the other one tg_kp_exit_disassociate_ctty ?  I will add later some detect kprobes load / disable policies so just in time for those ;-)  (I can do it later too...)

tixxdz (approved):
LGTM , thank you Jiri

pkg/grpc/tracing/tracing.go
kkourt: I don't think is worth backporting, but happy to defer to @mtardy

pkg/testutils/sensors/load.go
```diff
@@ -15,6 +16,7 @@ type SensorProg struct {
 	Name  string
 	Type  ebpf.ProgramType
 	NotIn bool
+	RegEx bool
```
kkourt, line 19: > At the moment regular expression is implemented via strings.Contains
> call, because it's enough for now.

I think calling it regex is confusing. 

How about have an enum?

```go
type ProgMatch = int
const (
      ProgMatchFull ProgMatch = iota // ==
      ProgMatchPartial // strings.Contains()
)
```

we can add a regex later if needed.

pkg/sensors/base/base.go
```diff
@@ -64,6 +68,27 @@ var (
...
+
+		/* Preffer acct_process over disassociate_ctty */
+		if has_acct_process {
+			Exit.Attach = "acct_process"
+			Exit.Label = "kprobe/acct_process"
+		} else if has_disassociate_ctty {
+			Exit.Attach = "disassociate_ctty"
+			Exit.Label = "kprobe/disassociate_ctty"
+		}
+		failed = !(has_acct_process || has_disassociate_ctty)
```
kkourt, line 86: Is there a need to check `has_acct_process` and `has_disassociate_ctty` again?

How about:
```
} else {
   failed = false
}
```

pkg/sensors/base/base.go
```diff
@@ -64,6 +68,27 @@ var (
...
+	}
+	logger.GetLogger().WithField("failed", failed).
+		Infof("Exit probe on %s", Exit.Attach)
```
kkourt, line 89: Is there a way to recover from this? If not, I think we should just add a `.Fatal()` here so we die early.

kkourt (changes requested):
Thanks!

I have some comments, but LGTM!

A followup might be a way to test both approaches, since our testing now will only test the `acct_process`. This seems like a `good-first-issue`.

bpf/process/bpf_exit.c
olsajiri: hum, so we want all programs with tg_kp_ prefixes?

pkg/testutils/sensors/load.go
olsajiri: yep, sounds good, thanks

pkg/sensors/base/base.go
olsajiri: yes, I'll change that

pkg/sensors/base/base.go
olsajiri: we would start without hooked exit probe.. so it's probably better to bail out then do some strange things

bpf/process/bpf_exit.c
kkourt: I thin `tg_` should suffice. The general idea is to have a (reasonable) way to have a common prefix for tetragon objects so that we can identify them in a system (e.g., when doing `bpftool prog list`).

kkourt (approved):
Thanks!

kkourt:
The failure is a strange one, but we can ignore it.

image

All tests are successful, but we fail to upload the artifact file (which is empty, because all tests were successful).

tixxdz:
> The failure is a strange one, but we can ignore it.
> 
> image
> 
> All tests are successful, but we fail to upload the artifact file (which is empty, because all tests were successful). @olsajiri feel free to merge.

issue is tracked here https://github.com/cilium/tetragon/issues/1895  cc @mtardy knows about it

bpf/process/bpf_exit.c
tixxdz: So yeh tg_ we do want this, for the extra kp_ up to you no strong opinion, we can go without it.
I'm working on tracing kprobes arm disarm code so being able to identify with tg_ prefix is a quicky way to say if tetragon probes are being disarmed

pkg/grpc/tracing/tracing.go
mtardy: just saw that! indeed, out of curiosity, how you detected that?

PR #1865, opened 2023-12-08
workflow: fix PR link checker to check local branch links

### LINKED ISSUE #1858: docs: replace link checker `github.com/cilium/tetragon/tree/main` links with actual PR branch
Inspired by this discussion: https://github.com/cilium/tetragon/pull/1845#issuecomment-1845901826.

GitHub repository links that will be created on main by merging a PR are always making the link checker fail, we can fix that by replacing `cilium/tetragon/tree/main` with the correct PR branch. If I remember correctly, the link checker has this kind of replacement option.

mtardy:
Tries to fix #1858

PR link checker was checking the GitHub links against main which might not exist when the PR has not been merged on main yet.

Unfortunately `lychee --remap` just searched and replaced and could not just replace some part of the URL so I did good old find/sed commands.

### DIFF
--- a/.github/workflows/check-links-pr.yaml
+++ b/.github/workflows/check-links-pr.yaml
@@ -22,6 +22,23 @@ jobs:
           hugo-version: '0.120.4'
           extended: true
 
+      - name: Replace GitHub main links with local branch
+        shell: bash
+        run: |
+          REPO=${{ github.event.pull_request.head.repo.full_name }}
+          BRANCH=${{ github.event.pull_request.head.ref }}
+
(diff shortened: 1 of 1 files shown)


lambdanis:
Hmm, I think this won't work with PRs from forks, will it?

mtardy:
> Hmm, I think this won't work with PRs from forks, will it?

Ah indeed didn't think about that thanks, I need to replace more than just the branch name! That should be a quick fix!

mtardy:
> Hmm, I think this won't work with PRs from forks, will it?

This should be good now @lambdanis

lambdanis (approved):
looks good :)

PR #1871, opened 2023-12-13
testutils: Fix RepoRootPath when Tetragon is vendored

tpapagian:
We just need to skip the path on the right of vendor.


mtardy (approved):
looks weird but this function is already a hack so!

PR #1882, opened 2023-12-14
fieldfilters: fix regression with missing top-level info and add test

willfindlay:
We had a regression after some fieldfilter bug fixes and performance optimizations that caused top-level information to be missing from events when field filters are applied. The fix here is to make sure that we are setting non-event fields in the destination struct. Apply the fix and add a unit test to check for future regressions.

```release-note
Fix a regression related to field filters that could cause top-level information to be missing from events.
```


willfindlay:
Sigh looks like some Ubuntu mirror issues in CI again

willfindlay:
I'll kick the job again after a few minutes

kkourt (approved):

PR #1891, opened 2023-12-15
tetragon: Small filter fixes

olsajiri:
Removing early binary filter and small tailcall index changes.

### DIFF
--- a/bpf/process/pfilter.h
+++ b/bpf/process/pfilter.h
@@ -396,6 +396,10 @@ selector_process_filter(__u32 *f, __u32 index, struct execve_map_value *enter,
 	__u32 len;
 	__u64 i;
 
+	/* Do binary filter first for selector index */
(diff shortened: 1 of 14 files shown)


kevsecurity (approved):
I'd prefer more descriptive (and stand-alone) commit messages but otherwise LGTM.

bpf/process/pfilter.h
```diff
@@ -396,6 +396,10 @@ selector_process_filter(__u32 *f, __u32 index, struct execve_map_value *enter,
 	__u32 len;
 	__u64 i;
 
+	/* Do binary filter first for selector index */
+	if (!match_binaries(index))
```
jrfastab, line 400: we might get more binary filters e.g. uid/gid so this might become something like process_filter_binary with multiple filters soon. But fine its ok for now.

Issue #1707, opened 2023-11-03
filename monitoring not recgonizing/printing out binary cmd correctly


Jiahao0:
### What happened?

1. env:
Embedded Linux
`Linux tegra-ubuntu 5.10.120-rt70-tegra #1 SMP PREEMPT RT Wed Oct 25 10:07:33 CST 2023 aarch64 aarch64 aarch64 GNU/Linux`

2. config:
Using example config filename_monitoring.yaml. I comment out 'security_path_truncate' part, because it's not in /proc/kallsyms. I put it on this path:
`/etc/tetragon/tetragon.tp.d/filename_monitoring.yaml`

3. run:
`$ tetragon &`
then:
`$ tetra getevents -o compact`
In another console:
`/etc/tetragon/tetragon.tp.d# cat filename_monitoring.yaml`

4. error:
The command **cat** cannot be recgonize/printed out correctly.
```
📚 read     /usr/bin/bash /etc/tetragon/tetragon.tp.d                     
📚 read     /usr/bin/bash /etc/tetragon/tetragon.tp.d                     
🚀 process  /etc/tetragon/tetragon.tp.d/�D]� filename_monitoring.yaml 
📝 mmap-r   /etc/tetragon/tetragon.tp.d/�D]� /etc/ld.so.cache        
📚 read     /etc/tetragon/tetragon.tp.d/�D]� /etc/tetragon/tetragon.tp.d/filename_monitoring.yaml 
📚 read     /etc/tetragon/tetragon.tp.d/�D]� /etc/tetragon/tetragon.tp.d/filename_monitoring.yaml 
📚 read     /etc/tetragon/tetragon.tp.d/�D]� /etc/tetragon/tetragon.tp.d/filename_monitoring.yaml 
📚 read     /etc/tetragon/tetragon.tp.d/�D]� /etc/tetragon/tetragon.tp.d/filename_monitoring.yaml 
💥 exit     /etc/tetragon/tetragon.tp.d/�D]� filename_monitoring.yaml 0 
💥 exit       0
```

Any idea what may be the problem?
Thanks.

### Tetragon Version

```
$ tetra version
CLI version: v0.12.0-pre.0-231-g02861f05
```

### Kernel Version

`Linux tegra-ubuntu 5.10.120-rt70-tegra #1 SMP PREEMPT RT Wed Oct 25 10:07:33 CST 2023 aarch64 aarch64 aarch64 GNU/Linux`

### Kubernetes Version

NA

### Bugtool

[...]

PR #1902, opened 2023-12-18
bpf: unit tests and fixes for `prepend_name` function

mtardy:
See commits!


bpf/process/bpf_process_event.h
```diff
@@ -108,42 +114,50 @@ d_unlinked(struct dentry *dentry)
...
 	// Needed to bound that for probe_read call.
-	if (buffer_offset > PATH_MAP_SIZE - 256)
+	if (buffer_offset >= MAX_BUF_LEN)
 		return -ENAMETOOLONG;
 
-	probe_read(bf + buffer_offset, sizeof(char), &slash);
+	if (write_slash)
+		probe_read(buf + buffer_offset, sizeof(char), &slash);
+
 	// This ensures that dlen is < 256, which is aligned with kernel's max dentry name length
```
willfindlay, line 152: dlen -> namelen to match the variable rename

bpf/process/bpf_process_event.h
```diff
@@ -108,42 +114,50 @@ d_unlinked(struct dentry *dentry)
...
 	// with the current buffer pointer. This will be at max 256 bytes (similar to the initial
 	// size).
 	// Needed to bound that for probe_read call.
-	if (buffer_offset > PATH_MAP_SIZE - 256)
+	if (buffer_offset >= MAX_BUF_LEN)
 		return -ENAMETOOLONG;
 
-	probe_read(bf + buffer_offset, sizeof(char), &slash);
+	if (write_slash)
+		probe_read(buf + buffer_offset, sizeof(char), &slash);
```
willfindlay, line 150: Hm, couldn't this be simplified to avoid the probe_read call?

bpf/tests/prepend_name_test.c
```diff
@@ -0,0 +1,56 @@
...
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, int);
+	__type(value, struct test_prepend_name_state_map_value);
+} test_prepend_name_state_map SEC(".maps");
+
+__attribute__((section("raw_tracepoint/test"), used)) int
```
willfindlay, line 34: This BPF unit testing is extremely valuable and honestly long overdue. Thanks for adding the first of hopefully many to come

willfindlay (approved):
Just a couple nits, but LGTM. Might be helpful to have someone else do a pass of the `prepend_name` changes, but I didn't find any obvious bugs.

bpf/process/bpf_process_event.h
mtardy: This works on `6.5.0`, let's see if it passes tests on older versions.

```diff
diff --git a/bpf/process/bpf_process_event.h b/bpf/process/bpf_process_event.h
index ff145373..b2b46845 100644
--- a/bpf/process/bpf_process_event.h
+++ b/bpf/process/bpf_process_event.h
@@ -116,7 +116,6 @@ d_unlinked(struct dentry *dentry)
 static inline __attribute__((always_inline)) int
 prepend_name(char *buf, char **bufptr, int *buflen, const char *name, u32 namelen)
 {
-       char slash = '/';
        // contains 1 if the buffer is large enough to contain the whole name and a slash prefix
        bool write_slash = 1;
 
@@ -147,7 +146,7 @@ prepend_name(char *buf, char **bufptr, int *buflen, const char *name, u32 namele
                return -ENAMETOOLONG;
 
        if (write_slash)
-               probe_read(buf + buffer_offset, sizeof(char), &slash);
+               buf[buffer_offset] = '/';
 
        // This ensures that namelen is < 256, which is aligned with kernel's max dentry name length
        // that is 255 (https://elixir.bootlin.com/linux/v5.10/source/include/uapi/linux/limits.h#L12).
```

mtardy:
Note that this has a chance of fixing https://github.com/cilium/tetragon/issues/1707, it can be related.

bpf/tests/prepend_name_test.c
```diff
@@ -0,0 +1,56 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* Copyright Authors of Cilium */
+
+//go:build ignore
```
tpapagian, line 4: Question: Why do we need that?

tpapagian (approved):
LGTM, thanks! Having a skeleton to build eBPF unit tests would be really useful!

bpf/tests/prepend_name_test.c
mtardy: So this is a bit weird but you can't have a `.c` file in the same directory as a go module unless you are using CGO 🤔  most Go commands will return `C source files not allowed when not using cgo`.

mtardy: I wanted to house the C file with the BPF test harness and the Go file in the same directory so it's needed!

kkourt:
🚀 

@mtardy I feel like we should backport this change to 1.0. WDYT?

mtardy:
> 🚀 
> 
> 
> 
> @mtardy I feel like we should backport this (or at least the buffix) to 1.0. WDYT?

Yes, it would make sense since it might fix a user facing bug. Let me do that!

PR #1904, opened 2023-12-19
tetragon: Use cilium/ebpf detection

olsajiri:
We can outsource some of our detection code to ebpf/cilium,
some we need to keep so far, because there's no support for that yet.


mtardy (approved):
Good idea!

PR #1914, opened 2023-12-23
tetragon: Add uprobe multi support

olsajiri:
cilium/ebpf changes https://github.com/cilium/ebpf/pull/1269

### DIFF
--- a/pkg/bpf/detect.go
+++ b/pkg/bpf/detect.go
@@ -7,6 +7,7 @@
 package bpf
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"unsafe"
@@ -27,6 +28,7 @@ type Feature struct {
 
 var (
 	kprobeMulti         Feature
+	uprobeMulti         Feature
 	buildid             Feature
 	modifyReturn        Feature
 	modifyReturnSyscall Feature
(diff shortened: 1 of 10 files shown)


pkg/bpf/detect.go
```diff
@@ -70,6 +72,47 @@ func HasKprobeMulti() bool {
...
+		return false
+	}
+	defer prog.Close()
+
+	ex, err := link.OpenExecutable("/proc/self/exe")
+	if err != nil {
+		return false
+	}
+
+	// need cilium/ebp fix, can't pass just addresses without symbol
```
kkourt, line 100: Is there an issue for this on cilium/ebpf? Would be nice to have a reference to check when this is fixed and update the code.

nit: typo -> cilium/ebp

kkourt (approved):
Looks good to me, thanks!

jrfastab (approved):

jrfastab:
Looks good fix small typo and merge imo.

pkg/bpf/detect.go
olsajiri: ah, the comment is no longer valid, it got fixed as preparation for uprobe multi support in ebpf/cilium lib
now it is:

```
func (ex *Executable) addresses(symbols []string, addresses, offsets []uint64) ([]uint64, error) {
        n := len(symbols)
        if n == 0 {
                n = len(addresses)
        }
```
will remove the comment and fix the typo, thanks

PR #1915, opened 2023-12-26
chore: rename TracingPolicy function

sadath-12:
renaming ```waitForTracingPolicy``` to ```ensureTracingPolicy``` because that is what it does and the waiting part is done by ```WaitForTracingPolicy``` so we could import ```ensureTracingPolicy``` to other packages in future without confusions

```release-note
Rename a function in e2e test helpers.
```


lambdanis (approved):

tixxdz:
Thanks @sadath-12 for the patch. Can you please copy the PR message into the commit message?

sadath-12:
Thank you @tixxdz

tixxdz (approved):
Thanks

PR #1917, opened 2023-12-27
fix: fill the missing type in tracingpolicy example

sadath-12:
fills the missing ```type``` value under the ```args``` for tracingpolicy example


lambdanis (commented):
I don't have full context on why this directory exists, but I'm wondering if these should be moved under the top level `examples/tracingpolicy`.

lambdanis:
> I don't have full context on why this directory exists, but I'm wondering if these should be moved under the top level `examples/tracingpolicy`.

ok, so both directories are used in tests:
* `pkg/tracingpolicy/examples/tracepoint-lseek-pid.yaml` is embedded [here](https://github.com/cilium/tetragon/blob/ef1c5f5db1c0f7721dd8e93829036e1f5162d943/pkg/tracingpolicy/generictracingpolicy_test.go#L397)
* `examples/tracingpolicy` is walked [here](https://github.com/cilium/tetragon/blob/main/pkg/tracingpolicy/generictracingpolicy_test.go#L470)

The former was initially added for tests, and only then the latter started being used in tests too. I think we should move `tracepoint-lseek-pid.yaml` to `examples/tracingpolicy` and update the test, but check if there is a similar TracingPolicy there first. Could you do it @sadath-12?

sadath-12:
Sure thanks @lambdanis

sadath-12:
@lambdanis  the ```//go:embed`` there I think uses ```go list``` to list files in the current dir and that might be the reason the yaml is there .

sadath-12:
> tracepoint-lseek-test.yaml

yes it can be removed since ```pkg/tracingpolicy/examples/tracepoint-lseek-pid.yaml``` kind of similar

pkg/tracingpolicy/examples/tracepoint-lseek-pid.yaml
```diff
@@ -8,7 +8,9 @@ spec:
     event: "sys_enter_lseek"
     args:
     - index: 7
+      type: "auto"
```
lambdanis, line 11: "auto" is the default, so this is technically not needed, but doesn't hurt either

lambdanis (approved):

pkg/tracingpolicy/examples/tracepoint-lseek-pid.yaml
sadath-12: got it 👍😊

lambdanis:
The CI failure is a known issue: #2010. It's unrelated, so I'm merging this PR.

PR #1919, opened 2023-12-28
feat: validate crd versions

### LINKED ISSUE #1412: CI: check that the CRD version is updated
For every change in the CRD, we need to update the patch version in: https://github.com/cilium/tetragon/blob/1b454b2640622ed0e9f0c42d0033449d8aee0882/pkg/k8s/apis/cilium.io/v1alpha1/register.go#L18.

Create a GH actions that checks whether there are changes in the CRD and complains if the version was not updated.

sadath-12:
### Issue

This pr Closes #1412  

### Description
 If there is a change in the crd yaml files and if ```CustomResourceDefinitionSchemaVersion``` value is not updated accordingly in ```pkg/k8s/apis/cilium.io/v1alpha1/version.go``` or vice-versa . The CI will fail

How to test locally ?

follow this https://github.com/cilium/tetragon/pull/1919#pullrequestreview-1811074405 

for the event.json thing set it as 

```{
  "pull_request": {
    "head": {
      "ref": "test-dev-2"
    },
    "base": {
      "ref": "test-dev",
      "sha":"58acf63b51238a93cb2d13cba90a80d6662d45bd"
    }
  }
        }
```

you can replace the sha field by the latest commit on your base branch

here I added CustomResourceDefinitionSchemaVersion to ```pkg/k8s/apis/cilium.io/v1alpha1/version.go``` in both branches

and changed yaml files ```pkg/k8s/apis/cilium.io/v1alpha1/version.go``` and  CustomResourceDefinitionSchemaVersion in HEAD branch (test-dev-2) 

change the branch

```git checkout test-dev-2``` (the branch that makes the pr request because we have HEAD in the yaml)

then you should see the success output as I saw

```
[Check CRD version update/check-version] ⭐ Run Main actions/checkout@v3
[Check CRD version update/check-version]   🐳  docker cp src=/home/sadath/go/src/tetragon/. dst=/home/sadath/go/src/tetragon
[Check CRD version update/check-version]   ✅  Success - Main actions/checkout@v3
[Check CRD version update/check-version] ⭐ Run Main Check for CRD changes and version update
[Check CRD version update/check-version]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/2] user= workdir=
| old_version=0.10.1
| new_version=0.10.4
| crd_changed=1
| version_changed=1
[Check CRD version update/check-version]   ✅  Success - Main Check for CRD changes and version update
[Check CRD version update/check-version] Cleaning up container for job check-version
[Check CRD version update/check-version] 🏁  Job succeeded
```

if you changed crd yaml files but not version you will see the failure output something as I see here 

```
[Check CRD version update/check-version]   ✅  Success - Main Install git
[Check CRD version update/check-version] ⭐ Run Main actions/checkout@v3
[Check CRD version update/check-version]   🐳  docker cp src=/home/sadath/go/src/tetragon/. dst=/home/sadath/go/src/tetragon
[Check CRD version update/check-version]   ✅  Success - Main actions/checkout@v3
[Check CRD version update/check-version] ⭐ Run Main Check for CRD changes and version update
[Check CRD version update/check-version]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/2] user= workdir=
| old_version=0.10.1
| new_version=0.10.1
| CRD changed but version not updated
[Check CRD version update/check-version]   ❌  Failure - Main Check for CRD changes and version update
[Check CRD version update/check-version] exitcode '1': failure
[Check CRD version update/check-version] 🏁  Job failed
Error: Job 'check-version' failed
```
And for vice-versa 
```
[Check CRD version update/check-version] ⭐ Run Main actions/checkout@v3
[Check CRD version update/check-version]   🐳  docker cp src=/home/sadath/go/src/tetragon/. dst=/home/sadath/go/src/tetragon
[Check CRD version update/check-version]   ✅  Success - Main actions/checkout@v3
[Check CRD version update/check-version] ⭐ Run Main Check for CRD changes and version update
[Check CRD version update/check-version]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/2] user= workdir=
| old_version=0.10.1
| new_version=0.10.7
| Version updated but CRD not changed
[Check CRD version update/check-version]   ❌  Failure - Main Check for CRD changes and version update
[Check CRD version update/check-version] exitcode '1': failure
[Check CRD version update/check-version] 🏁  Job failed
Error: Job 'check-version' failed
```

### DIFF
--- a/.github/workflows/validate-crd.yaml
+++ b/.github/workflows/validate-crd.yaml
@@ -1,0 +1,45 @@
+name: Check CRD version update
+
+on:
+  pull_request:
+    paths:
+      - 'pkg/k8s/apis/cilium.io/client/crds/v1alpha1/*.yaml'
+      - 'pkg/k8s/apis/cilium.io/v1alpha1/version.go'
+
+jobs:
+  check-version:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+      - name: Check for CRD changes and version update
+        run: |
+          crd_changed=0
+          version_changed=0
+
+          # Check for CRD changes
+          crd_changes=$(git diff --name-only ${{ github.base_ref }} ${{ github.head_ref }} -- pkg/k8s/apis/cilium.io/client/crds/v1alpha1/*.yaml)
+          if [ -n "$crd_changes" ]; then
+            crd_changed=1
+          fi
+
+          # Check for version variable changes
+
+          old_version=$(git show ${{ github.event.pull_request.base.sha }}:pkg/k8s/apis/cilium.io/v1alpha1/version.go | sed -n 's/^const CustomResourceDefinitionSchemaVersion = "\(.*\)".*/\1/p')
+          new_version=$(sed -n 's/^const CustomResourceDefinitionSchemaVersion = "\(.*\)".*/\1/p' pkg/k8s/apis/cilium.io/v1alpha1/version.go)
+
+          echo "old_version=$old_version"
+          echo "new_version=$new_version"
+
+          if [ "$old_version" != "$new_version" ]; then
+            version_changed=1
+          fi
+
+          if [ "$crd_changed" -eq 1 ] && [ "$version_changed" -eq 0 ]; then
+            echo "Changes to the files pkg/k8s/apis/cilium.io/client/crds/v1alpha1/*.yaml requires CustomResourceDefinitionSchemaVersion to be updated in pkg/k8s/apis/cilium.io/v1alpha1/version.go"
+            exit 1
+          fi
+
+          if [ "$crd_changed" -eq 0 ] && [ "$version_changed" -eq 1 ]; then
+            echo "CustomResourceDefinitionSchemaVersion in pkg/k8s/apis/cilium.io/v1alpha1/version.go to be updated only in case of modifying the files pkg/k8s/apis/cilium.io/client/crds/v1alpha1/*.yaml"
+            exit 1
+          fi

### CODE REVIEW

File: .github/workflows/validate-crd.yaml [hunk 1]
```diff
@@ -0,0 +1,45 @@
...
+    paths:
+      - 'pkg/k8s/apis/cilium.io/client/crds/v1alpha1/*.yaml'
+
+jobs:
+  check-version:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: ${{ github.event_name == 'pull_request' && 2 || 0 }}
```
lambdanis, line 13: This workflow is configured to run only on pull requests, so why check here if `github.event_name == 'pull_request'`?

.github/workflows/validate-crd.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+            if [ -z "$old_version" ]; then
+              echo "Error: Unable to retrieve old version from the base branch."
+              exit 1
+            fi
+
+            echo "old_version=$old_version"
+            echo "new_version=$new_version"
+
+            if [ "$old_version" == "$new_version" ]; then
```
lambdanis, line 39: We could also check exactly the the patch number is incremented by 1. But I don't consider it necessary, just an idea.

lambdanis (commented):
Thank you for tackling this @sadath-12.

Ideally I would imagine this check as a Makefile rule, so that we can run it in CI, but also when validating changes locally. Can you define such a rule in this PR?

Please also squash the commits, so that the git history is more readable :)

.github/workflows/validate-crd.yaml
sadath-12: Ya I like the idea tho. but would it catch up properly since the version number could be changed in anyway for example 10.1.20 -- 10.2.01 something

sadath-12:
> Thank you for tackling this @sadath-12.
> 
> Ideally I would imagine this check as a Makefile rule, so that we can run it in CI, but also when validating changes locally. Can you define such a rule in this PR?
> 
> Please also squash the commits, so that the git history is more readable :)

can we check the commit difference in makefile ?

.github/workflows/validate-crd.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  check-version:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Get changed files
+        id: changed-files
+        run: echo "changed_files=$(git diff --name-only -r HEAD^1 HEAD | xargs)" >> $GITHUB_ENV
```
lambdanis, line 16: Since it looks only at the last commit, it won't work for PRs with multiple commits where the last one isn't modifying CRDs, right?

lambdanis:
> can we check the commit difference in makefile ?

Good question - we could, but it would be likely confusing what we're actually checking :) Maybe it's better to leave it as is.

.github/workflows/validate-crd.yaml
sadath-12: Your right . Just confirmed about it . Any suggestions how to proceed further?

sadath-12: yes for our case I think since we squash It wun't cause the issues . I'm open to your ideas on this

.github/workflows/validate-crd.yaml
sadath-12: just checked . If we remove this It wun't detect the changes

.github/workflows/validate-crd.yaml
willfindlay: There are definitely ways around this, but it's probably easiest to just use this action for it: https://github.com/marketplace/actions/get-all-changed-files

sadath-12: sure , I'll try them

sadath-12: thank you @willfindlay . I have updated

sadath-12:
Updated with other action

.github/workflows/validate-crd.yaml
kkourt: That's not an adequate reason to keep it though :)

.github/workflows/validate-crd.yaml
```diff
@@ -0,0 +1,52 @@
...
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: ${{ github.event_name == 'pull_request' && 2 || 0 }}
+
+      - name: Get all changed files
+        id: changed-files
+        uses: jitterbit/get-changed-files@v1
+
+      - name: Check for CRD changes and version update
+        run: |
```
kkourt, line 21: This seems more complicated that it has to be. My suggestion would be to follow a simpler approach.
Also, this issue requires changes to the CI that tricky to test.

My suggestion would be to use `git diff` to check the changes, and `act` to test the changes locally.

kkourt (commented):
Thank you for taking on this issue!
I wanted to provide a starting point on how you can start solving this issue.

First, let's create a simple workflow. I'll start with a clean repo to make things easier.

```
$ git clone git@github.com:cilium/tetragon.git
$ cd tetragon
$ git switch -c crd-validation
```

Let's add a starting point for the action:
```
$ curl -s https://gist.githubusercontent.com/kkourt/f886b0a4dff79b472a179c9200205c16/raw/c4c3d1cf35c90a917e514e8ebd1b3078a0ca7bc0/validate-crd.yaml | tee .github/workflows/validate-crd.yaml
```

You can see the file here: https://gist.github.com/kkourt/f886b0a4dff79b472a179c9200205c16

We can now use acct to execute the action. First, we need to properly set the base and head refs (see https://nektosact.com/beginner/index.html#using-event-file-to-provide-complete-event-payload).

To do so we add and `events.json` file
```

  "pull_request": {
    "head": {
      "ref": "crd-validation"
    },
    "base": {
      "ref": "origin/main"
    }
  }
}
```

Now we can use `acct`:
```
$ gh act -W .github/workflows/validate-crd.yaml --eventpath event.json
```

Which will result in:
```
| origin/main
| Differences in yaml files:
| Differences in schema:
```

Since there are no differences. 
We can start simple by creating the differences. Here's a simple commit that modifies the yaml file:

```
commit 9a90ffbd8063a4f1ba8caa89fc315566ba232f3a (HEAD -> crd-validation)
Author: Kornilios Kourtis <kornilios@gmail.com>
Date:   Tue Jan 9 13:51:27 2024 +0100

    dummy
    
    Signed-off-by: Kornilios Kourtis <kornilios@gmail.com>

diff --git a/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml b/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml
index 31ad2495..b95412aa 100644
--- a/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml
+++ b/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.yaml
@@ -1390,6 +1390,7 @@ spec:
                                   - State
                                   - InMap
                                   - NotInMap
+                                  - IsPizza
                                   type: string
                                 values:
                                   description: Value to compare the argument against.
```

Now running `act`, we get:
```
| Differences in yaml files:
| diff --git a/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpol
| icies.yaml b/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolici
| es.yaml
| index b95412aa..31ad2495 100644
| --- a/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.y
| aml
| +++ b/pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpolicies.y
| aml
| @@ -1390,7 +1390,6 @@ spec:
|                                    - State
|                                    - InMap
|                                    - NotInMap
| -                                  - IsPizza
|                                    type: string
|                                  values:
|                                    description: Value to compare the argument ag
| ainst.
| Differences in schema:
```

So we can see that there are differences in the `yaml` files.

Please use above a starting point to develop a solution. Once finished, please provide the necessary steps for folk reviewing this PR to test the PR using `acct` for the two cases that we need covered: i) changes in the resources but no changes in the version, and ii) changes in both the resources and the versions. In (i) the action should fail, and in (ii) the action should succeed.

sadath-12:
Thank you @kkourt  ill work on it

.github/workflows/validate-crd.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+          echo "old_version=$old_version"
+          echo "new_version=$new_version"
+
+          if [ "$old_version" != "$new_version" ]; then
+            version_changed=1
+          fi
+
+          if [ "$crd_changed" -eq 1 ] && [ "$version_changed" -eq 0 ]; then
+            echo "CRD changed but version not updated"
```
lambdanis, line 38: Could you write in this message what exactly needs to be changed? For a user not super familiar with the CRDs code it might be unclear.

.github/workflows/validate-crd.yaml
```diff
@@ -0,0 +1,48 @@
...
+            exit 1
+          fi
+
+          if [ "$crd_changed" -eq 0 ] && [ "$version_changed" -eq 1 ]; then
+            echo "Version updated but CRD not changed"
```
lambdanis, line 43: Similar comment as above - could you write more precisely what's wrong, i.e. point the user to concrete files?

.github/workflows/validate-crd.yaml
```diff
@@ -0,0 +1,48 @@
...
+            exit 1
+          fi
+
+          echo "crd_changed=$crd_changed"
+          echo "version_changed=$version_changed"
```
lambdanis, line 48: Hmm, I'm not sure there is a value in printing these vars :)

lambdanis (commented):
Thanks for the updates, I left some comments regarding UX.

sadath-12:
Thank you @lambdanis  for improving

willfindlay (approved):
LGTM but we should make sure to test this locally using `act` before merging.

.github/workflows/validate-crd.yaml
```diff
@@ -0,0 +1,45 @@
...
+          # Check for CRD changes
+          crd_changes=$(git diff --name-only ${{ github.base_ref }} ${{ github.head_ref }} -- pkg/k8s/apis/cilium.io/client/crds/v1alpha1/*.yaml)
+          if [ -n "$crd_changes" ]; then
+            crd_changed=1
+          fi
+
+          # Check for version variable changes
+
+          old_version=$(git show ${{ github.event.pull_request.base.sha }}:pkg/k8s/apis/cilium.io/v1alpha1/register.go | grep 'CustomResourceDefinitionSchemaVersion' | awk -F'"' '{print $2}')
+          new_version=$(git show HEAD:pkg/k8s/apis/cilium.io/v1alpha1/register.go | grep 'CustomResourceDefinitionSchemaVersion' | awk -F'"' '{print $2}')
```
kkourt, line 28: Not sure how this would work, since `CustomResourceDefinitionSchemaVersion` lives `version.go`:
https://github.com/cilium/tetragon/blob/03beec7f1a0ca06675ff4bf7d89fec9307db8359/pkg/k8s/apis/cilium.io/v1alpha1/version.go#L10.

Not sure why my original example had `register.go` (maybe it was based on an older version), but please ensure that everything works properly in the current `main` branch.

Thanks!

kkourt (changes requested):

.github/workflows/validate-crd.yaml
sadath-12: Thanks I changed and tested I get the values 

```
[Check CRD version update/check-version]   🐳  docker exec cmd=[bash --noprofile --norc -e -o pipefail /var/run/act/workflow/2] user= workdir=
| old_version=1.1.1
| new_version=1.1.2
| Version updated but CRD not changed
```

kkourt (approved):
LGTM, thanks!

kkourt:
failures seem unrelated, so merging.

sadath-12:
@kkourt  issue closed but this pr not merged

PR #1920, opened 2023-12-28
Improve local k8s developer experience

### LINKED ISSUE #1849: Improve local k8s developer experience
The current state of the local k8s devexp is the following:
* [contrib/localdev](https://github.com/cilium/tetragon/blob/7c7cb636d23b7c35cdff30940e3f16f01a64d25b/contrib/localdev) directory contains scripts for running a local Kind cluster and installing Tetragon in it
* [Development setup](https://tetragon.io/docs/contribution-guide/development-setup/#running-tetragon-in-kind) docs page explains how to use these scripts and also has some instructions for Minikube

The commands in the docs seem a bit complicated and they don't cover anything apart from the basic installation (for example passing custom Helm values). I'm not sure if Minikube instructions are useful at all.

To make local k8s development easier I suggest to:
- [ ] Introduce Makefile rules for easy local development in Kind. Maybe the [Cilium setup](https://github.com/cilium/cilium/blob/main/Makefile) (and [docs](https://docs.cilium.io/en/latest/contributing/development/dev_setup/)) can be some inspiration.
- [ ] Update dev docs to use Makefile rules and cover a bit more advanced options. Remove Minikube instructions if nobody needs them.

Somewhat related: #1847

sadath-12:
Compile all commands listed for kind setup into one make command via ```make kind-setup``` and updated docs accordingly and removes minikube setup as there is neglible chances that someone will use it over kind and we have to make our docs clean

Closes #1849

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -304,6 +304,9 @@ image-clang:
 	$(QUIET)@echo "Push like this when ready:"
 	$(QUIET)@echo "${CONTAINER_ENGINE} push cilium/clang:$(DOCKER_IMAGE_TAG)"
 
+.PHONY: images
+images: image image-operator
+
 .PHONY: tarball tarball-release tarball-clean
 # Share same build environment as docker image
 tarball: tarball-clean image
@@ -401,3 +404,14 @@ version:
 .PHONY: docs
 docs:
 	$(MAKE) -C docs
+
+.PHONY: kind
+kind:
+	./contrib/localdev/bootstrap-kind-cluster.sh
+
+.PHONY: kind-install-tetragon 
+kind-install-tetragon:
+	./contrib/localdev/install-tetragon.sh --image cilium/tetragon:latest --operator cilium/tetragon-operator:latest
+
+.PHONY: kind-setup
+kind-setup: images kind kind-install-tetragon
--- a/docs/content/en/docs/contribution-guide/development-setup.md
+++ b/docs/content/en/docs/contribution-guide/development-setup.md
@@ -112,19 +112,11 @@ make tarball
 
 ### Running Tetragon in kind
 
-The scripts in contrib/localdev will help you run Tetragon locally in a kind
-cluster. First, ensure that docker, kind, kubectl, and helm are installed on
-your system. Then, run the following commands:
+This command will setup tetragon, kind cluster and install tetragon in it. Ensure docker, kind, kubectl, and helm are installed.
 
 ```shell
-# Build Tetragon agent and operator images
-make LOCAL_CLANG=0 image image-operator
-
-# Bootstrap the cluster
-contrib/localdev/bootstrap-kind-cluster.sh
-
-# Install Tetragon
-contrib/localdev/install-tetragon.sh --image cilium/tetragon:latest --operator cilium/tetragon-operator:latest
+# Setup tetragon on kind
+make kind-setup
 ```
 
 Verify that Tetragon is installed by running:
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -304,6 +316,9 @@ image-clang:
 	$(QUIET)@echo "Push like this when ready:"
 	$(QUIET)@echo "${CONTAINER_ENGINE} push cilium/clang:$(DOCKER_IMAGE_TAG)"
 
+.PHONY: build 
+build: image image-operator
```
lambdanis, line 320: Maybe `images` instead of `build`? It's more intuitive IMO.

Makefile
```diff
@@ -200,6 +204,14 @@ install:
 	$(INSTALL) -m 0755 -d $(DESTDIR)$(BINDIR)
 	$(INSTALL) -m 0755 ./tetragon $(DESTDIR)$(BINDIR)
 
+.PHONY: install-operator 
+install-operator:
```
lambdanis, line 208: This rule install both Tetragon agent and operator in a Kind cluster, so `kind-install-tetragon` would be better IMO. Also, it should depend on `kind` and `build`/`images`.

Makefile
```diff
@@ -200,6 +204,14 @@ install:
...
+	./contrib/localdev/install-tetragon.sh --image cilium/tetragon:latest --operator cilium/tetragon-operator:latest
+
+.PHONY: setup
+setup-kind: build kind install-operator
```
lambdanis, line 212: The rule name doesn't match `.PHONY` here, but I don't think this rule is necessary - if we add dependencies to the one above, we can just use that.

lambdanis (commented):
Thank you @sadath-12, I left some comments regarding rules names mostly.

Please also squash the commits for clean git history :)

lambdanis (approved):
looks good to me, thanks!

Makefile
```diff
@@ -87,6 +87,10 @@ endif
 GO_BUILD = CGO_ENABLED=0 GOARCH=$(GOARCH) $(GO) build $(GO_BUILD_FLAGS)
 GO_BUILD_HOOK = CGO_ENABLED=0 GOARCH=$(GOARCH) $(GO) -C contrib/rthooks/tetragon-oci-hook build $(GO_BUILD_FLAGS)
 
+.PHONY: kind
+kind:
+	./contrib/localdev/bootstrap-kind-cluster.sh
+
```
mtardy, line 93: I would not put this as the first rule since it will be triggered by running `make` instead of the `all` target.

Makefile
```diff
@@ -200,6 +204,14 @@ install:
...
 	$(INSTALL) -m 0755 ./tetragon $(DESTDIR)$(BINDIR)
 
+.PHONY: kind-install-tetragon 
+kind-install-tetragon:
+	./contrib/localdev/install-tetragon.sh --image cilium/tetragon:latest --operator cilium/tetragon-operator:latest
+
+.PHONY: kind-setup
+kind-setup: images kind kind-install-tetragon
+	
+
```
mtardy, line 214: nit whitespace

docs/content/en/docs/contribution-guide/development-setup.md
```diff
@@ -112,19 +112,12 @@ make tarball
 
 ### Running Tetragon in kind
 
-The scripts in contrib/localdev will help you run Tetragon locally in a kind
-cluster. First, ensure that docker, kind, kubectl, and helm are installed on
-your system. Then, run the following commands:
+The command will help you build tetragon, provision kind cluster and install tetragon in it. First, ensure that docker, kind, kubectl, and helm are installed on
+your system. Then, run the following command:
```
mtardy, line 116: nit: if you can wrap the text to 80 chars

mtardy (changes requested):
Thanks look good, I'm sorry I would just block this as it will change the default make target. See comments.

mtardy:
This is also a nit but checkpatch is detecting you have duplicate signature in your commit message. You can use `git commit --amend` to remove the unnecessary line. The commit should look like this (with empty lines):

```
<subsystem>: <title>

<body>

<signature>
```

sadath-12:
updated . thank you @mtardy

Makefile
```diff
@@ -90,6 +90,17 @@ GO_BUILD_HOOK = CGO_ENABLED=0 GOARCH=$(GOARCH) $(GO) -C contrib/rthooks/tetragon
 .PHONY: all
 all: tetragon-bpf tetragon tetra generate-flags test-compile tester-progs protoc-gen-go-tetragon tetragon-bench
 
+.PHONY: kind
```
mtardy, line 93: well this is a nit but I would have put that at the end of the Makefile instead of the beginning for clarity but this is okay...

mtardy (approved):
I was mostly commenting on the Makefile changes, if @lambdanis approves this I'm okay! :) Thanks for the work!

Makefile
sadath-12: ok let me do it anyways

lambdanis (approved):

PR #1923, opened 2023-12-29
handling metrics correctly

### LINKED ISSUE #1787: metrics: increment missing process/parent info errors after giving up instead of on every retry
We currently increment the `event_cache_process_info_failed` and `event_cache_parent_info_failed` error metrics on every retry in the event cache. The problem is that many of these retries are legitimate and do not strictly indicate an error, since the operation may indeed succeed on a subsequent attempt. It would be better to only increment the metrics when the `event.Color` value reaches the predetermined cutoff.

sadath-12:
This pr Closes #1787 .

## Description:
 Where I have considered the thershold to be ```CacheStrikes``` value to increment the metrics ```event_cache_process_info_failed``` && ```event_cache_parent_info_failed```  and not on all retries . On retries we will inc the values 
 ```event_cache_parent_info_retry``` and ```event_cache_process_info_retry```

### DIFF
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/ktime"
-	"github.com/cilium/tetragon/pkg/metrics/errormetrics"
 	"github.com/cilium/tetragon/pkg/metrics/eventcachemetrics"
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/process"
@@ -70,7 +69,7 @@ func HandleGenericInternal(ev notify.Event, pid uint32, tid *uint32, timestamp u
 	if parent != nil {
 		ev.SetParent(parent.UnsafeGetProcess())
 	} else {
-		errormetrics.ErrorTotalInc(errormetrics.EventCacheParentInfoFailed)
+		eventcachemetrics.EventCacheRetries(eventcachemetrics.ParentInfo).Inc()
 		err = ErrFailedToGetParentInfo
 	}
 
@@ -86,7 +85,7 @@ func HandleGenericInternal(ev notify.Event, pid uint32, tid *uint32, timestamp u
 		process.UpdateEventProcessTid(proc, tid)
 		ev.SetProcess(proc)
 	} else {
-		errormetrics.ErrorTotalInc(errormetrics.EventCacheProcessInfoFailed)
+		eventcachemetrics.EventCacheRetries(eventcachemetrics.ProcessInfo).Inc()
 		err = ErrFailedToGetProcessInfo
 	}
 
@@ -103,7 +102,7 @@ func HandleGenericInternal(ev notify.Event, pid uint32, tid *uint32, timestamp u
 func HandleGenericEvent(internal *process.ProcessInternal, ev notify.Event, tid *uint32) error {
 	p := internal.UnsafeGetProcess()
 	if option.Config.EnableK8s && p.Pod == nil {
-		errormetrics.ErrorTotalInc(errormetrics.EventCachePodInfoRetryFailed)
+		eventcachemetrics.EventCacheRetries(eventcachemetrics.PodInfo).Inc()
 		return ErrFailedToGetPodInfo
 	}
 
@@ -141,7 +140,9 @@ func (ec *Cache) handleEvents() {
 				tmp = append(tmp, event)
 				continue
 			}
-			if errors.Is(err, ErrFailedToGetProcessInfo) {
+			if errors.Is(err, ErrFailedToGetParentInfo) {
+				eventcachemetrics.ParentInfoError(notify.EventTypeString(event.event)).Inc()
+			} else if errors.Is(err, ErrFailedToGetProcessInfo) {
 				eventcachemetrics.ProcessInfoError(notify.EventTypeString(event.event)).Inc()
 			} else if errors.Is(err, ErrFailedToGetPodInfo) {
 				eventcachemetrics.PodInfoError(notify.EventTypeString(event.event)).Inc()
--- a/pkg/grpc/exec/exec.go
+++ b/pkg/grpc/exec/exec.go
@@ -196,7 +196,7 @@ func (msg *MsgExecveEventUnix) Retry(internal *process.ProcessInternal, ev notif
 	if option.Config.EnableK8s && containerId != "" {
 		podInfo = process.GetPodInfo(containerId, filename, args, nspid)
 		if podInfo == nil {
-			errormetrics.ErrorTotalInc(errormetrics.EventCachePodInfoRetryFailed)
+			eventcachemetrics.EventCacheRetries(eventcachemetrics.PodInfo).Inc()
 			return eventcache.ErrFailedToGetPodInfo
 		}
 	}
@@ -434,7 +434,7 @@ func (msg *MsgExitEventUnix) RetryInternal(ev notify.Event, timestamp uint64) (*
 			msg.RefCntDone[ParentRefCnt] = true
 		}
 	} else {
-		errormetrics.ErrorTotalInc(errormetrics.EventCacheParentInfoFailed)
+		eventcachemetrics.EventCacheRetries(eventcachemetrics.ParentInfo).Inc()
 		err = eventcache.ErrFailedToGetParentInfo
 	}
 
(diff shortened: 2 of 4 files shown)


lambdanis (commented):
Thank you for tackling this @sadath-12. Your change is going in a good direction overall, and I think it revealed more issues with how these metrics are defined :) I left a few comments. I'm not super familiar with this code, so if I'm missing something then let me know.

lambdanis:
One more note, you can ask Github to automatically close the issue after the PR is merged by adding e.g. `Fixes #1787` to the PR description (see https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue for details).

pkg/eventcache/eventcache.go
```diff
@@ -131,21 +129,29 @@ func (ec *Cache) handleEvents() {
...
 			event.color++
 			if event.color < CacheStrikes {
 				tmp = append(tmp, event)
 				continue
 			}
-			if errors.Is(err, ErrFailedToGetProcessInfo) {
-				eventcachemetrics.ProcessInfoError(notify.EventTypeString(event.event)).Inc()
-			} else if errors.Is(err, ErrFailedToGetPodInfo) {
-				eventcachemetrics.PodInfoError(notify.EventTypeString(event.event)).Inc()
-			}
```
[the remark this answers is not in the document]
sadath-12: I thought since ```err``` variable passes through 2 functions ```event.msg.RetryInternal(``` && ```event.msg.Retry(``` meanwhile we must consider it at each call

pkg/eventcache/eventcache.go
```diff
@@ -70,7 +70,6 @@ func HandleGenericInternal(ev notify.Event, pid uint32, tid *uint32, timestamp u
 	if parent != nil {
 		ev.SetParent(parent.UnsafeGetProcess())
 	} else {
-		errormetrics.ErrorTotalInc(errormetrics.EventCacheParentInfoFailed)
```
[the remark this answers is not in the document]
sadath-12: yes thanks had that in mind but was confused little

pkg/eventcache/eventcache.go
```diff
@@ -131,21 +129,29 @@ func (ec *Cache) handleEvents() {
...
 		if event.internal == nil {
 			event.internal, err = event.msg.RetryInternal(event.event, event.startTime)
+			if err != nil && event.color >= CacheStrikes {
+				if errors.Is(err, ErrFailedToGetParentInfo) {
+					errormetrics.ErrorTotalInc(errormetrics.EventCacheParentInfoFailed)
+				} else if errors.Is(err, ErrFailedToGetProcessInfo) {
+					errormetrics.ErrorTotalInc(errormetrics.EventCacheProcessInfoFailed)
+					eventcachemetrics.ProcessInfoError(notify.EventTypeString(event.event)).Inc()
+				}
+			}
```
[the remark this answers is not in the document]
sadath-12: by looking at the code they seem to be monitoring 2 different things and would cause no harm and increase the observability of 2 things ? what do you say

sadath-12:
Thanks for helping me to improve this @lambdanis . Please have  look at them

pkg/metrics/eventcachemetrics/eventcachemetrics.go
```diff
@@ -33,13 +33,19 @@ var (
 		Help:        "The total of errors encountered while fetching process exec information from the cache.",
 		ConstLabels: nil,
 	}, []string{"error"})
+	eventCacheRetriesTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: consts.MetricsNamespace,
+		Name:      "event_cache_retries_total",
+		Help:      "The total number of retries for event caching per event type.",
+	}, []string{"event_type"})
```
lambdanis, line 40: Could you give this label a different name? `event_type` typically corresponds to [EventType](https://github.com/cilium/tetragon/blob/f08c16f6fff107dba829b3d961c602f85158221b/api/v1/tetragon/events.proto#L18) from Tetragon API. Maybe `entry_type` would be more suitable here.

pkg/eventcache/eventcache.go
```diff
@@ -141,7 +140,10 @@ func (ec *Cache) handleEvents() {
 				tmp = append(tmp, event)
 				continue
 			}
-			if errors.Is(err, ErrFailedToGetProcessInfo) {
+			if errors.Is(err, ErrFailedToGetParentInfo) {
+				errormetrics.ErrorTotalInc(errormetrics.EventCacheParentInfoFailed)
+			} else if errors.Is(err, ErrFailedToGetProcessInfo) {
+				errormetrics.ErrorTotalInc(errormetrics.EventCacheProcessInfoFailed)
```
lambdanis, line 146: We can delete `EventCacheParentInfoFailed` and `EventCacheProcessInfoFailed` entirely from `errormetrics` I think, and define `ParentInfoErrors` in `eventcachemetrics`, similar to `ProcessInfoError`. The goal is to 1) avoid duplicate metrics (like it seems to be the case now for process info errors) 2) report different errors in a consistent way.

lambdanis (changes requested):
Thank you for the updates @sadath-12, I left a few more comments. When you make changes, please squash the commits too :)

lambdanis:
Also @willfindlay FYI, as you have context here you might have some more insights or suggestions.

sadath-12:
Thank you for holding on @lambdanis . I hope the new changes match the expectations

pkg/metrics/eventcachemetrics/eventcachemetrics.go
```diff
@@ -56,3 +75,13 @@ func PodInfoError(eventType string) prometheus.Counter {
 func EventCacheError(err string) prometheus.Counter {
 	return eventCacheErrorsTotal.WithLabelValues(err)
 }
+
+// Get a new handle on the eventCacheRetriesTotal metric for an eventType
+func EventCacheRetries(eventType string) prometheus.Counter {
+	return eventCacheRetriesTotal.WithLabelValues(eventType)
+}
```
lambdanis, line 82: nit: let's rename `eventType` to `entryType` here too

lambdanis (changes requested):
looks good to me now, thanks! I requested one more change, a bit of a nit.

sadath-12:
Thank you @lambdanis

lambdanis:
Thanks, can you also update the commit message to contain some description of the change and only one sign-off?

sadath-12:
Hope its ready to merge

lambdanis (approved):
looks good to me now, thanks!

pkg/eventcache/eventcache.go
```diff
@@ -218,7 +219,8 @@ func (ec *Cache) Add(internal *process.ProcessInternal,
 	e notify.Event,
 	t uint64,
 	s uint64,
-	msg notify.Message) {
+	msg notify.Message,
+) {
```
olsajiri, line 223: no need for this change

pkg/grpc/exec/exec.go
```diff
@@ -25,9 +25,7 @@ import (
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
-var (
-	nodeName = node.GetNodeNameForExport()
-)
+var nodeName = node.GetNodeNameForExport()
```
olsajiri, line 28: there's no need to change this, right?

olsajiri (commented):
please put in changelog how do new metrics look like and what's changed in the existing ones, probably best with output from metrics retrieval, thanks

sadath-12:
@olsajiri  actually that was done by gofumpt formatter . reverted those changes thanks 🙂

pkg/metrics/eventcachemetrics/eventcachemetrics.go
```diff
@@ -33,13 +39,26 @@ var (
 		Help:        "The total of errors encountered while fetching process exec information from the cache.",
 		ConstLabels: nil,
 	}, []string{"error"})
+	eventCacheRetriesTotal = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: consts.MetricsNamespace,
+		Name:      "event_cache_retries_total",
+		Help:      "The total number of retries for event caching per event type.",
```
willfindlay, line 45:
```suggestion
		Help:      "The total number of retries for event caching per entry type.",
```

willfindlay (changes requested):

willfindlay:
Thanks. Need to do a deeper review pass later but for now I left a nit for you.

sadath-12:
sure , thanks @willfindlay

willfindlay (approved):
LGTM

Issue #1741, opened 2023-11-13
Read absolute path for binary execution for matchBinaries


mtardy:
This is a followup of https://github.com/cilium/tetragon/pull/1731.

[For matchBinaries] the binary path is just the arg passed to execve and thus can be a relative path (this explains many of the users issues). A future patch is needed to read the absolute path of the task_struct (as we do on userspace side with /proc to fill the initial state of the execve_map) to make this feature complete.

> I agree that this is not tied to this PR but this would be also a great improvement in `matchBinaries` for a follow-up PR. I believe what you do here will help to have the full binary path ready from the kernel. Maybe we need to create an issue to keep track of this (if we do not already have one).
>
> _Originally posted by @tpapagian in https://github.com/cilium/tetragon/pull/1731#pullrequestreview-1726722554_

Might be useful: https://github.com/cilium/tetragon/pull/90.

PR #1926, opened 2024-01-03
bpf: read and copy proc exe at execve for matchBinaries

mtardy:
Fix for https://github.com/cilium/tetragon/issues/1741

This PR changes how we read the binary path at `execve` time. We now read `exe` from the `task_struct` instead of relying upon the tracepoint argument that was just the path given to `execve` (thus possibly relative). This was an issue since `matchBinaries` could be avoided using relative links instead of absolute ones at `execve`.

### For a future patch maybe:

In addition to this, we could also use this `exe` value for the binary path of the event. We also previously used the argument of `execve`, being possibly relative and corrected in userspace using `cwd`, but also possibly a symlink. This would thus introduce a user-facing change: tetragon events will now contain the canonical binary path while previously it could have been a symlink.

Without that user might try to "matchBinaries" a symlink file displayed in the Tetragon events and fail to make it work.

Preliminary work would be to make `prepend_name` work with 4096 (or `MAX_PATH`) buffer size.

### DIFF
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -270,6 +270,19 @@ struct msg_k8s {
 	char docker_id[DOCKER_ID_LENGTH];
 }; // All fields aligned so no 'packed' attribute.
 
+#define BINARY_PATH_MAX_LEN 256
+
+struct heap_exe {
+	// because of verifier limitations, this has to be 2 * 256 bytes while 256
+	// should be theoretically sufficient, and actually is, in unit tests.
+	char buf[BINARY_PATH_MAX_LEN * 2];
+	// offset points to the start of the path in the above buffer. Use offset to
+	// read the path in the buffer since it's written from the end.
+	char *off;
+	__u32 len;
+	__u32 error;
+}; // All fields aligned so no 'packed' attribute.
+
 struct msg_execve_event {
 	struct msg_common common;
 	struct msg_k8s kube;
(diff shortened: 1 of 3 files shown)


mtardy:
Unfortunately, it looks like it doesn't fix for 4.19, I'll have to put it out for small progs kernels :/

bpf/lib/process.h
```diff
@@ -270,6 +270,19 @@ struct msg_k8s {
 	char docker_id[DOCKER_ID_LENGTH];
 }; // All fields aligned so no 'packed' attribute.
 
+#define BINARY_PATH_MAX_LEN 256
+
+struct heap_exe {
+	// because of verifier limitations, this has to be 2 * 256 bytes while 256
```
jrfastab, line 276: Is this true with large programs? We might be able to widen this some if we feel its an actual issue.

bpf/lib/process.h
```diff
@@ -270,6 +270,19 @@ struct msg_k8s {
...
+	// should be theoretically sufficient, and actually is, in unit tests.
+	char buf[BINARY_PATH_MAX_LEN * 2];
```
jrfastab, line 278: If we were a bit smarter this can be improved 2x waste of space. Over 256B though I don't care to do the work.

jrfastab (approved):

jrfastab:
OK lets try this. Merging I want to land this in a test cluster and play with it though so I'll do that shortly.

tixxdz:
> OK lets try this. Merging I want to land this in a test cluster and play with it though so I'll do that shortly.

So I wanted to suggest for security reasons and be bullet proof go inside https://github.com/cilium/tetragon/blob/main/bpf/process/bpf_execve_bprm_commit_creds.c#L47  using bprm->file

As it is now on the tracepoint hook is **somehow** late the window race is so small and I don't think it can be triggered to be honest, but I prefer no surprise ;-)
The `allow_write_access(exec) ` https://elixir.bootlin.com/linux/latest/source/fs/exec.c#L1789 allows an fput on the file and even if there is some RCU lifetime on some fields of the file struct which are used to resolve the path, let's just ensure for this case that we do it properly under the right locks and avoid dereferencing pointers like that especially if this code is triggered by everyone, not to mention it is core logic of binary handling and we have safe easy alternative.   The other suggestion on bprm->file is totally safe, the bprm is locked.

PR #1931, opened 2024-01-04
tetragon: run Tetragon without access to CRD

### LINKED ISSUE #1880: Run Tetragon without access to CRD

In some scenarios we have a limited access to k8s cluster, so that when installing tetragon with `helm` , it fails with getting CRD "tracingpolicies.cilium.io":
изображение

At the same time we don't need CRD as in our solution all required policies are deployed via Tetragon API.

We would like to be able to run Tetragon without TracingPolicy CRD, because in our scenarios it's not needed at all: all policies are deployed via API.

I think one of the possible solution would be a special option both for tetragon and helm chart which will make it run in k8s cluster without access to CRD.

### Code of Conduct

aohoyd:
Fixes #1880

## Tetragon
- new flag `enable-tracing-policy-crd` was added and enabled by default

## Tetragon operator
- new flag `skip-tracing-policy-crd` was added and disabled by default

## Helm chart
- two flags above were added into configmaps and values file

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -72,6 +72,7 @@ import (
 	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	apiextensionsinformer "k8s.io/apiextensions-apiserver/pkg/client/informers/externalversions/apiextensions/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 )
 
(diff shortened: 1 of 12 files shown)


aohoyd:
I've synced with `main` and re-generated doc files

kkourt (commented):
Thanks, this looks good to me overall. Could you please split the patch into two patches: one for the agent changes, and one for the operator changes?

aohoyd:
> Thanks, this looks good to me. Could you please split the patch into two patches: one for the agent changes, and one for the operator changes?

Sure, done

mtardy:
> > Thanks, this looks good to me. Could you please split the patch into two patches: one for the agent changes, and one for the operator changes?
> 
> Sure, done (I've splitted commit)
> 
> @kkourt did you mean to open separated PR?

I can't speak for Kornilios but I would say separate commits, so looking all good!

kkourt:
> > Thanks, this looks good to me. Could you please split the patch into two patches: one for the agent changes, and one for the operator changes?
> 
> Sure, done (I've splitted commit)
> 
> @kkourt did you mean to open separated PR?

I mean splitting the commits. Thanks!

kkourt (approved):
LGTM, thanks!

aohoyd:
synced fork

lambdanis:
@aohoyd Can you rebase your branch with cilium:main instead of merging it? It makes the git history easier to read. At the moment the checkpath CI job is failing because of the merge commit (https://github.com/cilium/tetragon/actions/runs/7526335727/job/20488255210?pr=1931).

aohoyd:
I'm not sure, but failed test looks like spontaneous error. Can anyone restart it?

kkourt:
Tests look all good! ✅ 

@mtardy, @tpapagian, @lambdanis: do you plan to review?
Otherwise, I would suggest we merge.

tpapagian (approved):
Not super-familiar with this part of the codebase, but LGTM! Thanks

lambdanis (approved):
Small note: if all CRDs are disabled, then I believe Tetragon operator has nothing to do, but it still runs, and there is no way to disable it. Unless someone corrects me, we can update the Helm chart to allow disabling the whole operator in such cases. But this can be a follow-up, not blocking for this PR.

lambdanis:
> Operator should be disabled in next case:
> 
> * if `tetragonOperator.podInfo.enabled` is `false` and `tetragonOperator.tracingPolicy.enabled` is `false`

Yes. One option is to introduce `tetragonOperator.enabled` Helm value, then it'll be easier to maintain I think if new CRDs are introduced. cc @michi-covalent to confirm/deny/comment.

michi-covalent (approved):
adding `tetragonOperator.enabled` seems reasonable if we want to be able to disable the operator altogether. we can do that as a follow-up PR ✅

aohoyd:
I've added the new flag in #2004

mtardy:
Failure in the CI is unrelated https://github.com/cilium/tetragon/issues/2010 and we can merge this one.

PR #1940, opened 2024-01-08
api:events: add message field to inform users what is happening

tixxdz:
Right now all fields in Tetragon events are generated from code,
the only exception is the policy_name which does not inform users
what is going on from a high perspective.
    
This optional "message" field if present in Tracing Policies will be
copied and printed in the final generated event.
    
We want to inform users what is really happening, why this event was
generated? it is not obvious to non technical users.
    
An optional abstracted could be easy to read something like:
   "message": "Kernel module requested and is being loaded automatically"
   "message": "Kernel module is being loaded"
    
As we want to capture new users, the policy_name is not enough and
strictly speaking it has its own purpose, using abstracted messages
or logs appeals to more users.
    
The down side of this is our generated events may grow in size, thus
the new "message" field will be chopped to 256 characters only.
    
In future if the event size is really a problem, we can start shipping
default field filters, that hide these fields from the source before
even constructing the event.

This was split from https://github.com/cilium/tetragon/pull/1921 to get it merged


kevsecurity (approved):
LGTM once the static checks are all okay.

tixxdz:
> LGTM once the static checks are all okay.

Thanks @kevsecurity , I ended up adding code and properly escaping the field, PTAL again and if it is green I merge it

lambdanis:
> we can start shipping default field filters

Hmm, I think it will start to be necessary. I suppose the message field is useful for alert use-cases, when TracingPolicy is effectively an alert definition. Is this the reasoning behind it? For users who are treating TracingPolicies more like a logging/audit configuration, this field seems to be effectively duplicating the policy name (as both are provided by the user).

tixxdz:
> > we can start shipping default field filters
> 
> Hmm, I think it will start to be necessary. I suppose the message field is useful for alert use-cases, when TracingPolicy is 

The message field filters could be that, but I was referring in case users complain about events length then we can reduce it with fields filters. Now I'm not aware if a user already complained here about this. I'm seeing it from the other side, be user friendly first get them to understand what's happening use the policies first, then we can deal with rest.

> effectively an alert definition. Is this the reasoning behind it? For users who are treating TracingPolicies more like a logging/audit > configuration, this field seems to be effectively duplicating the policy name (as both are provided by the user).

Be more user friendly, there is only **one** tracing policy name by **multiple** hooks function names and both don't really tell users what's happening. Policy name is another thing, you can't express thing with it.

The PR has this example which are totally two different things from security perspective :
"message": "Kernel module requested and is being loaded automatically"
"message": "Kernel module is being loaded"

lambdanis:
> Be more user friendly, there is only one tracing policy name by multiple hooks function names and both don't really tell users what's happening. Policy name is another thing, you can't express thing with it.

This makes sense, thanks for the answer.

PR #1941, opened 2024-01-08
Add missed probes metrics

olsajiri:
wip needs https://github.com/cilium/ebpf/pull/1295

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -284,9 +284,6 @@ func tetragonExecute() error {
 		return fmt.Errorf("Failed to move old tetragon base directory: %w", err)
 	}
 
-	// we need file system mounts setup above before we detect features
-	log.Info("BPF detected features: ", bpf.LogFeatures())
-
 	if option.Config.PprofAddr != "" {
 		go func() {
 			if err := servePprof(option.Config.PprofAddr); err != nil {
@@ -392,6 +389,9 @@ func tetragonExecute() error {
 	if err := btf.InitCachedBTF(option.Config.HubbleLib, option.Config.BTF); err != nil {
 		return err
 	}
+
+	// needs BTF, so caling it after InitCachedBTF
+	log.Info("BPF detected features: ", bpf.LogFeatures())
 
 	if err := observer.InitDataCache(option.Config.DataCacheSize); err != nil {
 		return err
--- a/docs/content/en/docs/reference/metrics.md
+++ b/docs/content/en/docs/reference/metrics.md
@@ -181,6 +181,24 @@ The total number of Tetragon events per type that are failed to sent from the ke
 | ----- | ------ |
 | `msg_op` | `11, 13, 14, 15, 23, 24, 25, 26, 5, 7` |
 
+### `tetragon_missed_link_probes_total`
+
+The total number of Tetragon probe missed by link.
+
+| label | values |
+| ----- | ------ |
+| `attach` | `sys_panic` |
+| `policy` | `monitor_panic` |
+
+### `tetragon_missed_prog_probes_total`
+
+The total number of Tetragon probe missed by program.
+
+| label | values |
+| ----- | ------ |
+| `attach` | `sys_panic` |
+| `policy` | `monitor_panic` |
+
 ### `tetragon_msg_op_total`
 
 The total number of times we encounter a given message opcode. For internal use only.
--- a/pkg/bpf/detect.go
+++ b/pkg/bpf/detect.go
@@ -17,9 +17,11 @@ import (
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/asm"
+	ebtf "github.com/cilium/ebpf/btf"
 	"github.com/cilium/ebpf/features"
 	"github.com/cilium/ebpf/link"
 	"github.com/cilium/tetragon/pkg/arch"
+	"github.com/cilium/tetragon/pkg/btf"
 	"github.com/cilium/tetragon/pkg/logger"
 	"golang.org/x/sys/unix"
 )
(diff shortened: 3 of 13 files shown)


pkg/metrics/probemetrics/probemetrics.go
```diff
@@ -0,0 +1,54 @@
...
+
+var (
+	MissedProbes = metrics.NewBPFCounter(prometheus.NewDesc(
+		prometheus.BuildFQName(consts.MetricsNamespace, "", "missed_probes_total"),
+		"The total number of Tetragon probe missed per policy,probe.",
+		[]string{"policy", "attach"}, nil,
+	))
+
+	lock  sync.Mutex
+	stats = make(map[Key]*Stat)
```
lambdanis, line 32: I don't think this map is needed at all. What we can do instead is put the loop from `getKprobeStats` directly in the `Collect` method. Then the metrics collector will read metrics directly from ebpf/link structs instead of this map.

Unless you need this map for something else than metrics?

pkg/metrics/probemetrics/probemetrics.go
```diff
@@ -0,0 +1,54 @@
...
+	attach string
+}
+
+type Stat struct {
+	policy string
+	missed float64
+}
+
+var (
+	MissedProbes = metrics.NewBPFCounter(prometheus.NewDesc(
```
lambdanis, line 25: To have this metric documented in the metrics reference, we need to create a separate dummy collector that reports this metrics with example labels. See https://github.com/cilium/tetragon/commit/861489a02bac3971ce07374b3d64e3db6b2554f4 for an example, and [consts.go](https://github.com/cilium/tetragon/blob/main/pkg/metrics/consts/consts.go) for example labels.

This is not ideal tbh. I'm working on a better metrics library for Tetragon use cases, so hopefully soon it will be more straightforward.

pkg/metrics/probemetrics/probemetrics.go
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package probemetrics
```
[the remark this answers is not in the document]
olsajiri: perhaps it's good idea.. it's generic probe metric, but at the moment it's kprobe only, I'll put it in kprobemetrics

pkg/metrics/probemetrics/probemetrics.go
olsajiri: hum I'd need to move progs and the link_info logic into metrics code, which I don't like,
I'd like the kprobe specific code to stay in generickprobe.go ... it's also convenient
to manage the timer from there..

pkg/metrics/probemetrics/probemetrics.go
olsajiri: sure, will add that

pkg/metrics/probemetrics/probemetrics.go
```diff
@@ -0,0 +1,54 @@
...
+type Stat struct {
+	policy string
+	missed float64
+}
+
+var (
+	MissedProbes = metrics.NewBPFCounter(prometheus.NewDesc(
+		prometheus.BuildFQName(consts.MetricsNamespace, "", "missed_probes_total"),
+		"The total number of Tetragon probe missed per policy,probe.",
+		[]string{"policy", "attach"}, nil,
```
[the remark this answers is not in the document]
olsajiri: so attach is kernel function name for single kprobe and `kprobe_multi (3 functions)` for kprobe multi
similar for uprobes.. I'm not sure it's good idea to standardize this, we might want to change in future ;-)

lambdanis (approved):
Looks good :)

olsajiri:
> Thanks, LGTM.
> 
> One question what exactly is this missed metric? :) Is it when a link exists without a program? Or is it something else?

so it's incremented any time we don't execute the bpf program due to recursion in kernel,
like when you have 2 kprobe programs trying to run on top of each other on the same cpu

there was a case +- year ago where we were loosing exit events for this reason due to wrong
hook we used.. now if that happens again we will see that in the metrics

olsajiri:
@lambdanis could you please recheck, I changed few things on metrics side because the tetragon base changed

lambdanis (commented):
A few comments:
1. Ideally we should use `MustNewCustomCounter` and `NewCustomCollector` from `pkg/metrics` here. See how other collectors were defined in https://github.com/cilium/tetragon/pull/2714/commits/a05fbc71ba49225c045a6354737724810409e4c7 or https://github.com/cilium/tetragon/pull/2714/commits/7309c9ba22a2b07bae182e5ea8b10aa214be13fd. If it's a trouble for some reason, I'm ok with merging as-is, but then we should have a follow-up item to refactor it.
2. The new metrics are not documented. Using helpers from 1. should provide an easy way to include them in the generated reference.
3. I'm not sure `kprobe_multi (3 functions)` is a useful label value. Seeing such metric, how do I find out which functions were missed?

olsajiri:
> A few comments:
> 
>     1. Ideally we should use `MustNewCustomCounter` and `NewCustomCollector` from `pkg/metrics` here. See how other collectors were defined in [a05fbc7](https://github.com/cilium/tetragon/commit/a05fbc71ba49225c045a6354737724810409e4c7) or [7309c9b](https://github.com/cilium/tetragon/commit/7309c9ba22a2b07bae182e5ea8b10aa214be13fd). If it's a trouble for some reason, I'm ok with merging as-is, but then we should have a follow-up item to refactor it.
> 

ok, used that in new version

>     2. The new metrics are not documented. Using helpers from 1. should provide an easy way to include them in the generated reference.
> 

ok

>     3. I'm not sure `kprobe_multi (3 functions)` is a useful label value. Seeing such metric, how do I find out which functions were missed?

we don't have any more insight than that.. we get the counter per link/program that is attached to multiple functions,
and the missed counter spans over all attached functions.. we discussed some time ago to add a feature that would break down the counter to specific functions, but there was no need so far

lambdanis:
> ok, used that in new version

I see you use `MustNewCustomCounter` but the collector is defined as a separate type, not using `NewCustomCollector`. It would be nice to have the new metrics documented (in docs generated by `make metrics-docs`), and using `NewCustomCollector` should make it easy (you need to pass a collectForDocs function reporting fake metrics).

lambdanis (approved):
looks good :)

Issue #1894, opened 2023-12-15
ci: segmentation fault at the generate sbom ci job


tixxdz:
### What happened?

Ci fails from time to time at `Generate SBOM` https://github.com/cilium/tetragon/blob/main/.github/workflows/build-images-ci.yml#L131 with:

```
Run bom generate --format json -o sbom_ci_pr_tetragon_00a3f6a2d6f305dca9a09e9ce09a2818e02c3502.spdx \
level=info msg="bom v0.5.1: Generating SPDX Bill of Materials"
level=info msg="Processing directory ."
level=info msg="Loading license data from downloader"
level=info msg="Using embedded v3.20 license list"
level=info msg="Got 536 licenses from downloader"
level=info msg="Writing license data to /tmp/spdx/downloadCache"
level=info msg="Writing 536 SPDX licenses to /tmp/spdx/licenses"
level=info msg="Applying 29 ignore patterns to list of 16080 filenames"
level=info msg="Scanning 15752 files and adding them to the SPDX package"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x50 pc=0x894aeb]

goroutine 690 [running]:
sigs.k8s.io/bom/pkg/spdx.(*spdxDefaultImplementation).PackageFromDirectory.func1({0xc0002511c0, 0x31}, 0xc00c8fd1e0)
	/home/runner/go/pkg/mod/sigs.k8s.io/bom@v0.5.1/pkg/spdx/implementation.go:1032 +0x34b
created by sigs.k8s.io/bom/pkg/spdx.(*spdxDefaultImplementation).PackageFromDirectory in goroutine 1
	/home/runner/go/pkg/mod/sigs.k8s.io/bom@v0.5.1/pkg/spdx/implementation.go:1047 +0x92f
Error: Process completed with exit code 2.
```

We have multiple ci steps named: Generate SBOM.

An example of run: [example](https://github.com/cilium/tetragon/actions/runs/7446084163/job/20255633711?pr=1942).

### Tetragon Version

upstream

### Kernel Version

none

mtardy:
I took a look into that, one simple solution would be to stop generating SBOM for CI images but this would hide the issue and this can again appear when it's critical and we want to generate main CI images or release images.

It cannot be solved just by bumping the version of bom since we are using the latest release. I will open an issue against the [kubernetes-sigs/bom](https://github.com/kubernetes-sigs/bom) repo.

PR #1944, opened 2024-01-08
workflows: bump bom version for race condition fix

mtardy:
Fixes https://github.com/cilium/tetragon/issues/1894.

We experienced #1894, which made me create kubernetes-sigs/bom#385 only to fix the race condition and realize that it was already patched but no release existed.

### DIFF
--- a/.github/workflows/build-clang-image.yaml
+++ b/.github/workflows/build-clang-image.yaml
@@ -127,7 +127,7 @@ jobs:
         shell: bash
         env:
           # renovate: datasource=github-releases depName=kubernetes-sigs/bom
-          BOM_VERSION: v0.5.1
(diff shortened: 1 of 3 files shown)


jrfastab (approved):

jrfastab:
OK can we get v6 or whatever once it gets official though. Thanks.

mtardy:
> OK can we get v6 or whatever once it gets official though. Thanks.

Yep I expect this person to make a release https://github.com/kubernetes-sigs/bom/issues/385#issuecomment-1881137963 and maybe close my issue so that I'll get a notification. Anyway, renovate is already trying to change that to `v0.5.1` so we'll have a renovate update when >`v0.5.1` is out.

image

Issue #878, opened 2023-04-04
Add documentation for uprobes


mtardy:
Add documentation, in the TracingPolicy reference for example in the first time https://github.com/cilium/tetragon/blob/main/docs/content/en/docs/reference/tracing-policy.md and maybe present some use cases.

sadath-12:
assign me

PR #1945, opened 2024-01-08
doc: added uprobe info

sadath-12:
Closes: #878 

Added doc support for uprobe for user to get overview of what it is and how to use with tracingPolicy in the concepts section


docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -158,9 +158,24 @@ spec:
...
 ## Uprobes
 
-{{% pageinfo %}}
-This hook point method lacks documentation, see [issue #878](https://github.com/cilium/tetragon/issues/878).
-{{% /pageinfo %}}
+## Uprobes
+
+Uprobes are similar to kprobes, but they allow you to dynamically hook into any user-space function and execute BPF code. Uprobes are also tied to the binary version of the user-space program, so they may not be portable across different versions or architectures.
+
+To use uprobes, you need to specify the path to the executable or library file, and the symbol of the function you want to probe. You can use tools like objdump, nm, or readelf to find the symbol of a function in a binary file. Alternatively, you can use the perf tool to list the available uprobes for a running process.
```
mtardy, line 165: This is already nice, here are some thoughts to improve this, you don't need to absolutely do that in that PR but we can do a followup:

those are nice explanations: could we put the tools name in backtick like `objdump`, `nm` and maybe put an example of how you can find the `readline` symbol in `/bin/bash` with `nm` or another one? Like with a little step showing:

```bash
nm /bin/bash | grep readline
```
and showing what the output would look like?

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -158,9 +158,24 @@ spec:
...
+Here is an example of a policy that defines an uprobe for the readline function in the bash executable, and applies it to all processes that use the bash binary:
+
+```yaml
+spec:
+  uprobes:
+  - path: "/bin/bash"
+    symbol: "readline"
+```
+
+This example shows how to use uprobes to capture the user input command line string from all bash shells. The readline function is a library function that reads a line from the standard input and returns it as a string. By hooking into this function, we can monitor the commands entered by the users.
```
mtardy, line 178: proposition: this would be nice to have an example of output from `tetra getevents -o compact` when you apply this kind of policy.

mtardy (commented):
Thanks for the work! That's a nice improvement.

Could you wrap the lines around 80 chars? It makes it easier to propose feedback when those are splitted. Here are some propositions, don't feel obligated to apply everything I said, we can discuss it :)

docs/content/en/docs/concepts/tracing-policy/hooks.md
sadath-12: If I see through the docs ```tetra getevents -o compact``` is used for capturing events and tracingpolicy we are defining is used to take some action based on any event right so here it might not be necessary that this policy would print the events and we should show the output if I'm not wrong . Please do correct thanks 😊

mtardy: So my idea was to get a complete example of how to use uprobe from start to finish. You gave the example of the uprobe on readline in bash.

1. Search and make sure the symbol exists in your binary (the thing you did with the `nm` command)
2. Write the TracingPolicy (the example you showed)
3. See how to see the result in action, so basically show what it would look like if you started Tetragon with that policy, wrote a few commands in a bash shell and show the output from `tetra getevents -o compact`.

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -158,9 +158,55 @@ spec:
...
+0000000000097e40 T posix_readline_initialize
+00000000000d5690 T readline
+00000000000d52f0 T readline_internal_char
+00000000000d42d0 T readline_internal_setup
+00000000000d4530 T readline_internal_teardown
+00000000001555ec D rl_gnu_readline_p
+0000000000154528 D rl_readline_name
+0000000000155f18 B rl_readline_state
+00000000001555e8 D rl_readline_version
+```
```
mtardy, line 192: maybe redact a few entries of the output to make it look more simple.

````suggestion
The output should look similar to this, with a few lines redacted:

```
[...]
000000000009f2b0 T pcomp_set_readline_variables
0000000000097e40 T posix_readline_initialize
00000000000d5690 T readline
00000000000d52f0 T readline_internal_char
00000000000d42d0 T readline_internal_setup
[...]
```

You can see in the `nm` output: first the symbol value, then the symbol type,
for the `readline` symbol `T` meaning that this symbol is in the text (code)
section of the binary, and finally the symbol name. This confirms that the
`readline` symbol is present in the `/bin/bash` binary and might be a function
name that we can hook with a uprobe.
````

**Review by mtardy — COMMENTED** [review]
Thanks I think it's a great addition

File: docs/content/en/docs/concepts/tracing-policy/hooks.md (hunk 2 above)
Comment (mtardy) [reply docs/content/en/docs/concepts/tracing-policy/hooks.md:209]: I can suggest something if it's not very clear, tell me :)

Comment (sadath-12) [reply docs/content/en/docs/concepts/tracing-policy/hooks.md:209]: yes I wanted to see what matchActions you would configure

Comment (sadath-12) [reply docs/content/en/docs/concepts/tracing-policy/hooks.md:209]: yes please . In the context of k8s

**Comment by sadath-12** [comment]
Thank you @mtardy  😊

**Comment by jrfastab** [comment]
@mtardy mind giving this an ACK if your happy with the updates?

**Comment by mtardy** [comment]
> @mtardy mind giving this an ACK if your happy with the updates?

yes, please give me some time to understand something on our uprobes support to document it correctly and we'll merge this.

File: docs/content/en/docs/concepts/tracing-policy/hooks.md (hunk 2 above)
Comment (mtardy) [reply docs/content/en/docs/concepts/tracing-policy/hooks.md:207]:
```suggestion
This example shows how to use uprobes to hook into the readline function
running in all the bash shells.
```
Let's be succinct as we can't retrieve the args of readline as of now.

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -158,9 +158,53 @@ spec:
...
+Here is an example of a policy that defines an uprobe for the readline
+function in the bash executable, and applies it to all processes that use the
+bash binary:
+
+```yaml
+spec:
+  uprobes:
+  - path: "/bin/bash"
+    symbol: "readline"
+```
```
mtardy, line 205: Let's write a complete TracingPolicy instead of just the spec.
````suggestion
```yaml
kind: TracingPolicy
metadata:
  name: "example-uprobe"
spec:
  uprobes:
  - path: "/bin/bash"
    symbol: "readline"
```
````

mtardy (commented):
I'll propose the following change but I think this will be enough for now

sadath-12:
Thank you for improving this @mtardy  😊

mtardy (approved):
Thanks!

Issue #1775, opened 2023-11-20
bpf maps:metrics: fix bpf maps metrics


tixxdz:
### What happened?

Issues:

1. ~~Reading wrong bpf maps: we add the stats to the maps where it should already have it? this could lead to x_stats_stats~~ https://github.com/cilium/tetragon/blob/3326a78fbb120a9f5946b64733f8786d08be4b0e/pkg/observer/observer_stats.go#L33

2. ~~We fail or return here~~ https://github.com/cilium/tetragon/blob/3326a78fbb120a9f5946b64733f8786d08be4b0e/pkg/observer/observer_stats.go#L37  ~~and I guess we iterate over all maps~~

3. Rename all bpf map stats with "tg_$name_map_stats" schema

### Tetragon Version

master

### Kernel Version

None

tpapagian:
(1) and (2) are fixed by https://github.com/cilium/tetragon/pull/1949

lambdanis:
The bugs here are fixed, so closing the issue. Further improvements to how Tetragon monitors BPF maps (including monitoring more of them) should be done as part of https://github.com/cilium/tetragon/issues/2778

PR #1949, opened 2024-01-09
metrics: Do not return when we cannot find a _stats map

tpapagian:
In Tetragon, we have maps ending with `_stats` that contain metadata about the main maps. An example is that we have `execve_map` and `execve_map_stats` that contain errors and other information related to the `execve_map` map.

Now when we fail to open the `_stats` map, we just stop the loop and we can possibly miss to check for other maps. This patch changes that and tries for the next map.

Additionally, we do not try to collect stats for maps that already end up with `_stats` as this would result in `_stats_stats` map names.

### DIFF
--- a/pkg/observer/observer_stats.go
+++ b/pkg/observer/observer_stats.go
@@ -6,12 +6,15 @@ package observer
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/cilium/ebpf"
+	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/metrics/mapmetrics"
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/sensors"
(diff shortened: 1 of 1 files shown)


tixxdz (approved):
Thanks @tpapagian , maybe some log warn only once if the final loadPinnedMap() fails? or maybe doesn't make that much sense ;-)

BTW for issue https://github.com/cilium/tetragon/issues/1775 there is still that (3) rename map, should be moved to separate issue? so it is not lost.

tpapagian:
> Thanks @tpapagian , maybe some log warn only once if the final loadPinnedMap() fails? or maybe doesn't make that much sense ;-)

Thanks! Yes, it makes sense. If we see any of these warning we have to fix that. I added a warning there. 

> BTW for issue #1775 there is still that (3) rename map, should be moved to separate issue? so it is not lost.

OK, I will post a comment to that issue that (1, 2) are fixed by this PR. So I will not close that.

PR #1952, opened 2024-01-10
policyfilter: add a label that allows to filter based on namespaces

### LINKED ISSUE #1854: policyfilter: add a label that allows to filter based on namespaces
Allow policy pod-label filters to use a label depicting the namespace of the pod (e.g., `k8s:io.kubernetes.pod.namespace`).
This would allow more flexibility to filter policy application than namespaced policies (e.g., `k8s:io.kubernetes.pod.namespace NOT IN [noisy_namespaces]`)

sadath-12:
Closes #1854 

Allow policy pod-label filters to use a label depicting the namespace of the pod k8s:io.kubernetes.pod.namespace

```release-note
policy pod-label filters: allow for namespace filtering using k8s:io.kubernetes.pod.namespace
```

### DIFF
--- a/pkg/labels/labels.go
+++ b/pkg/labels/labels.go
@@ -20,6 +20,10 @@ const (
 	opNotIn
 )
 
+const (
+	K8sPodNamespace = "k8s:io.kubernetes.pod.namespace"
+)
+
 type selectorOp struct {
 	key      string
 	operator operator
--- a/pkg/labels/labels_test.go
+++ b/pkg/labels/labels_test.go
@@ -14,6 +14,7 @@ import (
 type testLabel struct {
 	labels      Labels
 	expectedRes bool
+	namespace   string
 }
 
 type testCase struct {
@@ -27,8 +28,8 @@ func TestLabels(t *testing.T) {
 			// empty label selector should match everything
 			labelSelector: &slimv1.LabelSelector{},
 			tests: []testLabel{
-				{map[string]string{"app": "tetragon"}, true},
-				{nil, true},
+				{map[string]string{"app": "tetragon"}, true, "default"},
+				{Labels{}, true, "default"},
 			},
 		}, {
 			labelSelector: &slimv1.LabelSelector{
@@ -37,8 +38,8 @@ func TestLabels(t *testing.T) {
 				},
 			},
 			tests: []testLabel{
-				{map[string]string{"app": "tetragon"}, true},
-				{map[string]string{"app": "cilium"}, false},
+				{map[string]string{"app": "tetragon"}, true, "default"},
+				{map[string]string{"app": "cilium"}, false, "default"},
 			},
 		}, {
 			labelSelector: &slimv1.LabelSelector{
@@ -49,9 +50,9 @@ func TestLabels(t *testing.T) {
 				}},
 			},
 			tests: []testLabel{
-				{map[string]string{"app": "tetragon"}, true},
-				{map[string]string{"app": "cilium"}, true},
-				{map[string]string{"app": "hubble"}, false},
+				{map[string]string{"app": "tetragon"}, true, "default"},
+				{map[string]string{"app": "cilium"}, true, "default"},
+				{map[string]string{"app": "hubble"}, false, "default"},
 			},
 		}, {
 			labelSelector: &slimv1.LabelSelector{
@@ -62,9 +63,9 @@ func TestLabels(t *testing.T) {
 				}},
 			},
 			tests: []testLabel{
-				{map[string]string{"app": "tetragon"}, false},
-				{map[string]string{"app": "cilium"}, false},
-				{map[string]string{"app": "hubble"}, true},
+				{map[string]string{"app": "tetragon"}, false, "default"},
+				{map[string]string{"app": "cilium"}, false, "default"},
+				{map[string]string{"app": "hubble"}, true, "default"},
 			},
 		}, {
 			labelSelector: &slimv1.LabelSelector{
@@ -74,9 +75,9 @@ func TestLabels(t *testing.T) {
 				}},
 			},
 			tests: []testLabel{
-				{map[string]string{"app": "tetragon"}, true},
-				{map[string]string{"application": "cilium"}, false},
-				{map[string]string{"app": "hubble"}, true},
+				{map[string]string{"app": "tetragon"}, true, "default"},
+				{map[string]string{"application": "cilium"}, false, "default"},
+				{map[string]string{"app": "hubble"}, true, "default"},
 			},
 		}, {
 			labelSelector: &slimv1.LabelSelector{
@@ -86,9 +87,9 @@ func TestLabels(t *testing.T) {
 				}},
 			},
 			tests: []testLabel{
-				{map[string]string{"app": "tetragon"}, false},
-				{map[string]string{"application": "cilium"}, true},
-				{map[string]string{"app": "hubble"}, false},
+				{map[string]string{"app": "tetragon"}, false, "default"},
+				{map[string]string{"application": "cilium"}, true, "default"},
+				{map[string]string{"app": "hubble"}, false, "default"},
 			},
 		}, {
 			labelSelector: &slimv1.LabelSelector{
@@ -101,10 +102,59 @@ func TestLabels(t *testing.T) {
 				},
 			},
 			tests: []testLabel{
-				{map[string]string{"app": "tetragon"}, true},
-				{map[string]string{"application": "tetragon"}, false},
-				{map[string]string{"app": "tetragon", "application": "tetragon"}, false},
-				{map[string]string{"app": "tetragon", "pizza": "yes"}, true},
+				{map[string]string{"app": "tetragon"}, true, "default"},
+				{map[string]string{"application": "tetragon"}, false, "default"},
+				{map[string]string{"app": "tetragon", "application": "tetragon"}, false, "default"},
+				{map[string]string{"app": "tetragon", "pizza": "yes"}, true, "default"},
+			},
+		}, {
+			labelSelector: &slimv1.LabelSelector{
+				MatchExpressions: []slimv1.LabelSelectorRequirement{{
+					Key:      K8sPodNamespace,
+					Operator: "In",
+					Values:   []string{"tetragon"},
+				}},
+			},
+			tests: []testLabel{
+				{map[string]string{K8sPodNamespace: "tetragon"}, true, "tetragon"},
+				{map[string]string{K8sPodNamespace: "test"}, false, "default"},
+			},
+		}, {
+			labelSelector: &slimv1.LabelSelector{
+				MatchExpressions: []slimv1.LabelSelectorRequirement{{
+					Key:      K8sPodNamespace,
+					Operator: "In",
+					Values:   []string{"cilium", "tetragon"},
+				}},
+			},
+			tests: []testLabel{
+				{map[string]string{"app": "tetragon"}, true, "cilium"},
+				{map[string]string{"app": "cilium"}, true, "tetragon"},
+				{map[string]string{"app": "hubble"}, false, "default"},
+			},
+		}, {
+			labelSelector: &slimv1.LabelSelector{
+				MatchExpressions: []slimv1.LabelSelectorRequirement{{
+					Key:      K8sPodNamespace,
+					Operator: "NotIn",
+					Values:   []string{"cilium", "tetragon"},
+				}},
+			},
+			tests: []testLabel{
+				{map[string]string{"app": "tetragon"}, false, "cilium"},
+				{map[string]string{"app": "cilium"}, false, "tetragon"},
+				{map[string]string{"app": "hubble"}, true, "default"},
+			},
+		}, {
+			labelSelector: &slimv1.LabelSelector{
+				MatchExpressions: []slimv1.LabelSelectorRequirement{{
+					Key:      K8sPodNamespace,
+					Operator: "Exists",
+				}},
+			},
+			tests: []testLabel{
+				{map[string]string{K8sPodNamespace: "tetragon"}, true, "tetragon"},
+				{map[string]string{}, true, ""},
 			},
 		},
 	}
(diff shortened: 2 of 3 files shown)


pkg/policyfilter/state.go
```diff
@@ -377,6 +380,7 @@ func (m *state) delPod(id PodID) *podInfo {
 }
 
 // AddPolicy adds a policy
+// marked
```
kkourt, line 383: What is the meaning of this comment?

pkg/policyfilter/state.go
```diff
@@ -426,7 +430,6 @@ func (m *state) AddPolicy(polID PolicyID, namespace string, podLabelSelector *sl
 
 // DelPolicy will destroly all information for the provided policy
 func (m *state) DelPolicy(polID PolicyID) error {
-
```
kkourt, line 429: Please don't include whitespace changes in the patches that change functionality. Either skip them or add them in a  seperate patch.

pkg/labels/labels_test.go
```diff
@@ -49,11 +56,12 @@ func TestLabels(t *testing.T) {
 				}},
 			},
 			tests: []testLabel{
-				{map[string]string{"app": "tetragon"}, true},
-				{map[string]string{"app": "cilium"}, true},
-				{map[string]string{"app": "hubble"}, false},
+				{map[string]string{"app": "tetragon"}, true, "default"},
+				{map[string]string{"app": "cilium"}, true, defaultNs},
```
kkourt, line 60: Why one test is using "default" and the other defaultNs?

pkg/labels/labels.go
```diff
@@ -20,6 +21,8 @@ const (
 	opNotIn
 )
 
+var k8sPodNamespace = "k8s:io.kubernetes.pod.namespace"
```
kkourt, line 24: Shouldn't this be a const?

pkg/labels/labels_test.go
```diff
@@ -14,33 +14,40 @@ import (
...
 	expectedRes bool
+	namespace   string
 }
 
 type testCase struct {
 	labelSelector *slimv1.LabelSelector
 	tests         []testLabel
 }
 
+var defaultNs = "default"
```
kkourt, line 25: What's the purpose of this var? Can't we just "default" in the tests?

kkourt (changes requested):
Thanks!

Overall, approach makes sense to me.
I'd like to have a second look, but please find some first comments.

sadath-12:
Thank you @kkourt  . Also it was my formatter I think which did those kind of jobs would be great If I can match my golang configuration of vscode to yours

sadath-12:
my settings.json 

```{
  "emmet.triggerExpansionOnTab": true,
  "emmet.includeLanguages": {
    "javascript": "javascriptreact"
  },

  "workbench.iconTheme": "material-icon-theme",

  "editor.fontLigatures": true,

  "liveServer.settings.CustomBrowser": "chrome",

  "security.workspace.trust.untrustedFiles": "open",
  "files.autoSave": "afterDelay",
  "[javascriptreact]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "settingsSync.ignoredSettings": [],
  "[javascript]": {
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  },
  "liveServer.settings.donotShowInfoMsg": true,
  "editor.defaultFormatter": "esbenp.prettier-vscode",
  "prettier.singleQuote": true,
  "prettier.semi": false,
  "thunder-client.codeSnippetLanguage": "cs-httpclient",
  "editor.fontFamily": "Cascadia Code",
  "cmake.configureSettings": {},
  "hardhat.telemetry": true,
  "[solidity]": {
    "editor.defaultFormatter": "NomicFoundation.hardhat-solidity"
  },
  "editor.formatOnSave": true,
  "editor.codeLensFontSize": 2,
  "editor.fontSize": 22,
  "workbench.editorAssociations": {
    "*.mp4": "default",
    "*.pkl": "default"
  },
  "workbench.productIconTheme": "icons-carbon",
  "rapidapi.terminalLink.enabled": false,
  "workbench.preferredLightColorTheme": "Default Dark+",
  "workbench.preferredHighContrastLightColorTheme": "Visual Studio Light",
  "[python]": {
    "editor.formatOnType": true
  },
  "editor.formatOnPaste": true,
  "vs-kubernetes": {
    "vscode-kubernetes.minikube-path.windows": "C:\\Users\\sadat\\.vs-kubernetes\\tools\\minikube\\windows-amd64\\minikube.exe"
  },
  "workbench.colorTheme": "GitHub Dark Default",
  "go.delveConfig": {},
  "[go]": {
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
      "source.organizeImports": true
    },
    "editor.defaultFormatter": "golang.go"
  },

  "[go.mod]": {
    "editor.formatOnSave": true,
    "editor.codeActionsOnSave": {
      "source.organizeImports": true
    }
  },
  "go.toolsManagement.autoUpdate": true,
  "go.vetOnSave": "package",

  "tabnine.experimentalAutoImports": true,
  "editor.codeActionsOnSave": {},
  "html.format.contentUnformatted": "",
  "editor.inlineSuggest.enabled": true,
  "github.copilot.enable": {
    "*": true,
    "plaintext": false,
    "markdown": true,
    "scminput": false
  },
  "go.useLanguageServer": true,
  "gopls": {
    "formatting.gofumpt": true
  },

  "go.formatTool": "gofumpt",

  "go.alternateTools": {
    "goformat": "gofumpt"
  },
  "eslint.codeActionsOnSave.rules": null,
  "workbench.settings.applyToAllProfiles": [
  
  ],
  "remote.SSH.remotePlatform": {
    "Ubuntu-Vm": "linux",
    "Ubuntu-rit": "linux"
  },
  "go.buildTags": "e2e,perf,conftests,unit,integration_test,certtests,subtlecrypto,all_components,allcomponents",
  "remote.SSH.path": "~/.ssh/id_rsa",
  "diffEditor.ignoreTrimWhitespace": false,
  "go.toolsEnvVars": {},
  "easycode.model": "gpt-4-1106-preview",
  "[go][go.mod]": {
    "editor.codeActionsOnSave": {
      "source.organizeImports": "explicit"
    }
  },
  "remote.autoForwardPortsSource": "hybrid",
  "cloudcode.duetAI.project": "fir-86adc",
  "javascript.updateImportsOnFileMove.enabled": "always"
}
```
also I use ```gofumpt -w``` to pass the CI for formatting

pkg/labels/labels.go
```diff
@@ -35,8 +40,29 @@ func (s selectorOp) hasValue(val string) bool {
...
+		case opExists:
+			return s.containsValue(ns)
+		case opNotIn:
+			return !s.hasValue(ns)
+		case opIn:
+			return s.hasValue(ns)
+		default:
+			return false
+		}
+	}
```
kkourt, line 65: Here, we duplicate code which I'd rather avoid.

pkg/policyfilter/state.go
```diff
@@ -205,7 +207,7 @@ func (pol *policy) podMatches(podNs string, podLabels labels.Labels) bool {
 		return false
 	}
 
-	return pol.podSelector.Match(podLabels)
+	return pol.podSelector.Match(podLabels, podNs)
```
kkourt, line 210: How about:

```
labels := podLabels
if _, ok  := labels[k8sPodNamespace]; !ok { // only define the namespace label if one does not exist
        labels[k8sPodNamespace] = podNs
}
```

Please add a test for the case where k8sPodNamspace is already a label in the pod.

pkg/policyfilter/state.go
```diff
@@ -72,9 +72,11 @@ const (
...
 )
 
-type PolicyID uint32
-type PodID uuid.UUID
-type CgroupID uint64
+type (
+	PolicyID uint32
+	PodID    uuid.UUID
+	CgroupID uint64
+)
```
kkourt, line 79: This seems like an unrelated change.

kkourt (changes requested):
Let's try to avoid code duplication.

Thanks!

pkg/labels/labels.go
sadath-12: thanks , I tried to merge the switch into 1 . Not sure how much more to optimize the code since there is separate handling for both cases

pkg/labels/labels.go
```diff
@@ -35,8 +40,29 @@ func (s selectorOp) hasValue(val string) bool {
...
+		case opExists:
+			return s.containsValue(labels[K8sPodNamespace])
+		case opNotIn:
+			return !s.hasValue(labels[K8sPodNamespace])
+		case opIn:
+			return s.hasValue(labels[K8sPodNamespace])
+		default:
+			return false
+		}
+	}
```
kkourt, line 65: Is this still needed?

kkourt: Not sure I understand. Can you provide an example?

kkourt: I'm not sure I understand still.

Aren't we matching namespaces using `k8s:io.kubernetes.pod.namespace` as a key, and the namespace name as the value?

So it's not clear to me how the `Exists` operator can be used in this case.

If the user wants to match pods having the `cilium` namespace, wouldn't they used the Equal operator? That is `k8s:io.kubernetes.pod.namespace <Equals> cilium`

sadath-12: suppose we have namespaces ```cilium``` ```ciliumXYZ``` ```tetragon``` my thought was 

```exists``` operator for value ```cilium``` should match ```cilium``` and ```ciliumXYZ```

```equal``` operator for value ```cilium``` will only match ```cilium``` .

If we leave that additional conditional in ```exists``` operator maybe it will check if ```k8s:io.kubernetes.pod.namespace``` is available in any pod which might not be the behaviour we want for this right ?

Comment (kkourt) [reply pkg/labels/labels.go]: > exists operator for value cilium should match cilium and ciliumXYZ
> equal operator for value cilium will only match cilium

Was the intention to use the exists operator as a substring match? I believe that for every
other label, the exists checks whether the key exists. I dont think it makes sense to change the semantics for this specific key.

> If we leave that additional conditional in exists operator maybe it will check if k8s:io.kubernetes.pod.namespace is available in any pod which might not be the behaviour we want for this right ?

Do you mean that the check for the exists operator would return true for every pod? I think that's fine.

Comment (sadath-12) [reply pkg/labels/labels.go]: ohk got it thanks

File: pkg/policyfilter/state.go [hunk 10]
```diff
@@ -205,7 +205,12 @@ func (pol *policy) podMatches(podNs string, podLabels labels.Labels) bool {
 		return false
 	}
 
-	return pol.podSelector.Match(podLabels)
+	podLabels1 := podLabels
```
kkourt, line 208: There is a test failure, which seems to happen in this code because we are calling it with a `nil` podLabels. Can you PTAL?

```
 --- FAIL: TestK8s (0.67s)
    --- FAIL: TestK8s/containers_change (0.00s)
panic: assignment to entry in nil map [recovered]
	panic: assignment to entry in nil map

goroutine 15 [running]:
testing.tRunner.func1.2({0x1886580, 0x1d630b0})
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1545 +0x238
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1548 +0x397
panic({0x1886580?, 0x1d630b0?})
	/opt/hostedtoolcache/go/1.21.6/x64/src/runtime/panic.go:914 +0x21f
github.com/cilium/tetragon/pkg/policyfilter.(*policy).podMatches(0xc0001999f8, {0x1aef0f5, 0x7}, 0x0?)
	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/policyfilter/state.go:210 +0xf1
github.com/cilium/tetragon/pkg/policyfilter.(*policy).podInfoMatches(...)
	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/policyfilter/state.go:217
github.com/cilium/tetragon/pkg/policyfilter.(*state).AddPolicy(0xc00011f5c0, 0x2, {0x0, 0x0}, 0xc000305040?)
	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/policyfilter/state.go:408 +0x494
github.com/cilium/tetragon/pkg/policyfilter.testContainersChange(0xc000305040?, 0xc00011f5c0, 0xc00011f560?)
	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/policyfilter/k8s_test.go:494 +0x176
github.com/cilium/tetragon/pkg/policyfilter.TestK8s.func6(0x0?)
	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/policyfilter/k8s_test.go:605 +0x1b
testing.tRunner(0xc000305040, 0xc000111248)
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1595 +0xff
created by testing.(*T).Run in goroutine 7
	/opt/hostedtoolcache/go/1.21.6/x64/src/testing/testing.go:1648 +0x3ad
FAIL	github.com/cilium/tetragon/pkg/policyfilter	0.684s
```

kkourt (changes requested):

pkg/policyfilter/state.go
kkourt: This comment was resolved, but there are still whitespaces in the patch.

kkourt (approved):
Thanks, LGTM.

I think we can ignore the CI failure.

pkg/policyfilter/state.go
sadath-12: umm . my naughtly formatter

kkourt (approved):
Great! 
Thanks @sadath-12 🎉

sadath-12:
hey @kkourt, was going back through policyfilter this week and something from this pr stuck with me. the namespace label matching - is it resolved in userspace before the bpf map lookup or does any of it happen in the prog itself? couldn't fully tell from state.go. also curious if you think arbitrary k8s labels could ever fit here or if that just blows up the map size

PR #1957, opened 2024-01-10
policylibrary: consolidate privileges raising operations into privileges-raise.yaml single policy

tixxdz:
This adds kernel capability types (uint64) and consolidates some operations into single privileges-raise.yaml Tracing Policy.

Produced kprobe event:
```
        ...
        "function_name": "security_capset",
        "args": [
          {
            "process_credentials_arg": {
          {
            "cap_effective_arg": "000001ffffffffff"
          },
          {
            "cap_inheritable_arg": "0000000000000000"
          },
          {
            "cap_permitted_arg": "000001ffffffffff"
          }
        ],
        "return": {
          "int_arg": 0
        },
        "action": "KPROBE_ACTION_POST",
        "policy_name": "privileges-raise",
        "return_action": "KPROBE_ACTION_POST",
        "message": "Process changed its capabilities using capset system call",
```

For further details, please see patches. Doc use case will follow later.


tixxdz:
as requested by @jrfastab tags are in their own PR https://github.com/cilium/tetragon/pull/2008

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1561,6 +1561,62 @@ func handleMsgGenericKprobe(m *api.MsgGenericKprobe, gk *genericKprobe, r *bytes
 			}
 			arg.Label = a.label
 			unix.Args = append(unix.Args, arg)
+		case gt.GenericKernelCap:
+			var output uint64
+			var arg api.MsgGenericKprobeArgKernelCapType
```
jrfastab, line 1566: the push 64-bit arg is a bit redundant it would be nice if we could extract that somehow to a helper. But I think the typing is a bit awkward in golang/here for that to work.

jrfastab (approved):

jrfastab:
needs a rebase but otherwise lgtm.

bpf/process/types/basic.h
```diff
@@ -1469,6 +1474,10 @@ static inline __attribute__((always_inline)) size_t type_to_min_size(int type,
 	case size_type:
 	case s64_ty:
 	case u64_ty:
+	case kernel_cap_ty:
+	case cap_inh_ty:
+	case cap_prm_ty:
+	case cap_eff_ty:
```
kkourt, line 1480: Why do we need a different type from the bpf side, can we just treat it as u64?

tixxdz:
ci failure about validate crd is tracked here: https://github.com/cilium/tetragon/issues/2063. The version is updated so merging.

Issue #1955, opened 2024-01-10
Dockerhub timeout in package-e2e-tests on buildjet runners (arm64)


mtardy:
Workflow `.github/workflows/packages-e2e-tests.yaml` produces timeout on non-GitHub runners. For example, [this run](https://github.com/cilium/tetragon/actions/runs/7452795618/job/20309734628?pr=1931).

```
ERROR: failed to solve: docker.io/library/alpine:3.19.0@sha256:51b67269f354137895d43f3b3d810bfacd3945438e94dc5ac55fdac340352f48: failed to copy: httpReadSeeker: failed open: unexpected status code https://registry-1.docker.io/v2/library/alpine/manifests/sha256:a70bcfbd89c9620d4085f6bc2a3e2eef32e8f3cdf5a90e35a1f95dcbd7f71548: 429 Too Many Requests - Server message: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

I see two things:
1. This is the last workflow using the buildjet runners, we migrated all the others arm64 runners to actuated.
2. Cross-compilation should work correctly now after the last patches.

PR #1958, opened 2024-01-10
ci: cross compile the arm tarball for tests

tixxdz:
This makes the ARM tarball ci test cross compile then run it on an ARM machine

close https://github.com/cilium/tetragon/issues/1955

### DIFF
--- a/.github/workflows/packages-e2e-tests.yaml
+++ b/.github/workflows/packages-e2e-tests.yaml
@@ -7,7 +7,7 @@ on:
     - 'docs/**'
 
 jobs:
-  standalone-tests:
+  standalone-tarball-builds:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
@@ -17,9 +17,13 @@ jobs:
           - os: ubuntu-22.04
             arch: amd64
             match_arch: x86-64
-          - os: buildjet-2vcpu-ubuntu-2204-arm
+            cross_compile: no
+            upload_path: upload/
+          - os: ubuntu-22.04
             arch: arm64
             match_arch: arm64
+            cross_compile: yes
+            upload_path: upload-cross-compile/
 
     steps:
       # https://github.com/docker/setup-buildx-action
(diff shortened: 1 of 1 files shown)


.github/workflows/packages-e2e-tests.yaml
```diff
@@ -37,19 +41,71 @@ jobs:
...
+
+      - name: Generate Cross Compiled Tetragon EE Tarball
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y qemu qemu-user-static binfmt-support
+          sudo update-binfmts --display
+          make TARGET_ARCH=${{ matrix.arch }} tarball
+          mkdir ${{ matrix.upload_path }}
+          mv ./build/${{ matrix.arch }}/linux-tarball/tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz ./${{ matrix.upload_path }}
+        if: ${{ matrix.cross_compile == 'yes' }}
```
mtardy, line 60: To avoid this another var you could also use to compare:
> `RUNNER_ARCH` The architecture of the runner executing the job. Possible values are X86, X64, ARM, or ARM64.

not sure it's easier or better tbh

.github/workflows/packages-e2e-tests.yaml
```diff
@@ -37,19 +41,71 @@ jobs:
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        include:
+          - os: ubuntu-22.04
+            arch: amd64
+            match_arch: x86-64
+            cross_compile: no
+            upload_path: upload/
+          - os: buildjet-2vcpu-ubuntu-2204-arm
```
mtardy, line 81: please remove buildjet and use actuated instead for arm runners (those are free, and we pay for buildjet). You can adjust the number of CPU and amount of RAM independently as needed, but here is a suggestion:
```suggestion
          - os: actuated-arm64-4cpu-8gb
```

.github/workflows/packages-e2e-tests.yaml
```diff
@@ -37,19 +41,71 @@ jobs:
...
+            cross_compile: yes
+            upload_path: upload-cross-compile/
+
+    steps:
+      - name: Checkout Source Code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+        with:
+          persist-credentials: false
+          fetch-depth: 0
+          submodules: true
```
mtardy, line 93: You can remove all of them, `fetch-depth: 0` will slow down stuff for nothing here and we don't have submodules here.
```suggestion
```

mtardy: Same on the other `actions/checkout` in the previous building job.

.github/workflows/packages-e2e-tests.yaml
```diff
@@ -37,19 +41,71 @@ jobs:
...
+          mv ./build/${{ matrix.arch }}/linux-tarball/tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz ./${{ matrix.upload_path }}
+        if: ${{ matrix.cross_compile == 'yes' }}
+
+      # Cache tarball releases for later
+      - name: Save tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz Tarball
+        uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+        with:
+          name: tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}
+          path: ${{ matrix.upload_path }}
+          retention-days: 1
```
mtardy, line 68: if you want you could increase the retention-days so that we have the thing available in case of failure and run a delete artifact on success so that we don't fill the storage for nothing (like we do in vmtests).

.github/workflows/packages-e2e-tests.yaml
```diff
@@ -37,19 +41,71 @@ jobs:
...
-          password: ${{ secrets.DOCKER_HUB_PASSWORD_CI }}
-
       - name: Generate Tetragon Tarball
         id: tetragon-tarball
-        run: make tarball
+        run: |
+          make tarball
+          mkdir ${{ matrix.upload_path }}
+          mv ./build/${{ matrix.arch }}/linux-tarball/tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz ./${{ matrix.upload_path }}
+        if: ${{ matrix.cross_compile == 'no' }}
```
mtardy, line 50: nit: put those `if` statements as the first fields in the YAML so that it's easier to read. like:

```yaml
      - name: Generate Tetragon Tarball
        if: ${{ matrix.cross_compile == 'no' }}
        id: tetragon-tarball
        run: |
          make tarball
          mkdir ${{ matrix.upload_path }}
          mv ./build/${{ matrix.arch }}/linux-tarball/tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz ./${{ matrix.upload_path }}
```

mtardy (commented):
Thanks for tackling this, here are a few suggestions/comments!

.github/workflows/packages-e2e-tests.yaml
```diff
@@ -37,19 +41,71 @@ jobs:
...
+
+      - name: Generate Cross Compiled Tetragon EE Tarball
```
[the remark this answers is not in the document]
tixxdz: :-D !

.github/workflows/packages-e2e-tests.yaml
tixxdz: Done, thank you!

.github/workflows/packages-e2e-tests.yaml
tixxdz: Indeed

.github/workflows/packages-e2e-tests.yaml
tixxdz: I didn't get the delete artifact (no idea how it work ;-) , but retention-days 1 is minimum last time I checked, and I don't think we want to keep the tarball, the build is logged in the previous job, so we are covered

.github/workflows/packages-e2e-tests.yaml
```diff
@@ -17,9 +17,13 @@ jobs:
...
             arch: amd64
             match_arch: x86-64
-          - os: buildjet-2vcpu-ubuntu-2204-arm
+            cross_compile: no
+            upload_path: upload/
+          - os: ubuntu-22.04
             arch: arm64
             match_arch: arm64
+            cross_compile: yes
+            upload_path: upload-cross-compile/
```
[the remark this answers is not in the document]
tixxdz: having upload_path there make it easy to grep for both arch and it doesn't hurt anyway.

tixxdz:
Thanks @mtardy for the review, I fixed the most of them , except for the upload_path as it's not really an issue and it is easy for me to grep for it

.github/workflows/packages-e2e-tests.yaml
mtardy: Okay!

Delete works like this:
https://github.com/cilium/tetragon/blob/1a84735f6b21a9c5c58c82c18baf8311664572f8/.github/workflows/vmtests.yml#L159-L163

PR #1962, opened 2024-01-11
Actions: add rateLimitScope

kevsecurity:
Post actions can have a rateLimit argument that specifies how often identical events from the same hook and thread are generated. There is a use case to rate limit per process or generally.

This commit introduces the rateLimitScope argument, to be used with rateLimit, to specify whether the rate limiting should be limited to the same thread, the same process, or globally, using values "thread" (default), "process", or "global".

```release-note
Add rateLimitScope to control what rateLimit applies to.
```

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1865,9 +1865,13 @@ do_action_signal(int signal)
  */
 #define KEY_BYTES_PER_ARG 40
 
+/* Rate limit scope. */
+#define ACTION_RATE_LIMIT_SCOPE_THREAD	0
+#define ACTION_RATE_LIMIT_SCOPE_PROCESS 1
+#define ACTION_RATE_LIMIT_SCOPE_GLOBAL	2
+
 struct ratelimit_key {
 	__u64 func_id;
-	__u64 retprobe_id;
 	__u64 action;
 	__u64 tid;
 	__u8 data[MAX_POSSIBLE_ARGS * KEY_BYTES_PER_ARG];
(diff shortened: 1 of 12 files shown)


bpf/process/types/basic.h
```diff
@@ -1926,9 +1930,20 @@ rate_limit(__u64 ratelimit_interval, struct msg_generic_kprobe *e)
 	ro_heap = map_lookup_elem(&ratelimit_ro_heap, &zero);
 
 	key->func_id = e->func_id;
-	key->retprobe_id = e->retprobe_id;
```
tixxdz, line 1929: I guess this was never used?

bpf/process/types/basic.h
```diff
@@ -1926,9 +1930,20 @@ rate_limit(__u64 ratelimit_interval, struct msg_generic_kprobe *e)
 	ro_heap = map_lookup_elem(&ratelimit_ro_heap, &zero);
```
tixxdz, line 1931: would it make sense to memset zero those returned heap entries? (didn't check all the logic)

tixxdz (approved):
LGTM after I guess updating this https://github.com/cilium/tetragon/blob/78262f08662cf23cf1045bdb79e4c6a5165778f6/pkg/k8s/apis/cilium.io/v1alpha1/version.go#L10 ?

Really nice feature was planning to add it. Also we may want to add:
rateLimitBy="function (default) | policy_name"

bpf/process/types/basic.h
kevsecurity: Yeah was overkill.

bpf/process/types/basic.h
kevsecurity: I overwrite the heap with a read only heap that is all zero, using a probe_read. It should cost about the same as a memset but doesn't cost me the instruction / complexity count.

kevsecurity:
> LGTM after I guess updating this
> 
> https://github.com/cilium/tetragon/blob/78262f08662cf23cf1045bdb79e4c6a5165778f6/pkg/k8s/apis/cilium.io/v1alpha1/version.go#L10
> 
> ?
> Really nice feature was planning to add it. Also we may want to add: rateLimitBy="function (default) | policy_name"

Yeah I forgot the version! I'll do that now.

PR #1867, opened 2023-12-11
Allow event filtering based on tracing policy

ioandr:
This PR extends the list of filters by implementing the policy name filter. This allows Tetragon to filter events based on the name of a tracing policy.

~It also improves Tetragon's docs~ (see https://github.com/cilium/tetragon/pull/1965) and extends test suites as needed.

Closes https://github.com/cilium/tetragon/issues/1855

```release-note
add filter event based on tracing policy name
```

PR #1965, opened 2024-01-12
Documentation improvements

ioandr:
This PR provides the following fixes for Tetragon's docs:

1. Ensure that the URL with the manifest of the Cilium SW demo application is rendered in `kubectl` commands
2. Ensure that the K8s installation guide works for both `zsh` and `bash` shells

Signed-off-by: Ioannis Androulidakis <androulidakis.ioannis@gmail.com>

### DIFF
--- a/docs/hugo.toml
+++ b/docs/hugo.toml
@@ -113,6 +113,9 @@ github_project_repo = "https://github.com/cilium/tetragon"
 # Specify a value here if your content directory is not in your repo's root directory
 github_subdir = "docs"
 
+# URL of the Star Wars demo app.
+demo_app_url = "https://raw.githubusercontent.com/cilium/cilium/v1.15.0-pre.1/examples/minikube/http-sw-app.yaml"
+
 [params.search.algolia]
 appId = "UI18HE156K"
 apiKey = "77f164b3638095772b770596db900fea"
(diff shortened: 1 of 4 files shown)


kkourt (approved):
Thanks!

docs/hugo.toml
```diff
@@ -113,6 +113,9 @@ github_project_repo = "https://github.com/cilium/tetragon"
 # Specify a value here if your content directory is not in your repo's root directory
 github_subdir = "docs"
 
+# URL of the Star Wars demo app.
+demo_app_url = "https://raw.githubusercontent.com/cilium/cilium/v1.15.0-pre.1/examples/minikube/http-sw-app.yaml"
+
```
mtardy, line 118: Thanks, I would like to merge that but why did you move this piece of TOML in the file with no change? Did you forget something here?

If you want to update that, can you let it at the lines where it was sitting?

mtardy (changes requested):
Thanks a lot for creating a separate PR! Just a small thing to understand before we merge this :)!

docs/hugo.toml
ioandr: > why did you move this piece of TOML in the file with no change
I explained this in https://github.com/cilium/tetragon/pull/1867#discussion_r1449573367, PTAL and let me know if anything does not make sense.

mtardy: Wow indeed those 2 params don't work and it's broken, I think it's my fault I didn't realize it! thanks, will fix the other one, let's merge this.

mtardy (approved):

PR #1975, opened 2024-01-14
tetragon: Change uprobe spec

olsajiri:
Currently we define uprobe with path/symbol path, which not
handy when you have more symbols from single path to probe.
    
Changing the spec so the uprobe is defined by path and array
of symbols, like:
    
     spec:
       uprobes:
       - path: /bin/bash
         symbols:
         - "_start"
         - "main"
           "builtin_help"


mtardy (commented):
Hey, what was the motivation behind this? I guess it's a bit similar to what we do in kprobe with list, but also a bit different.

Could you update the documentation on uprobe as well? :)

olsajiri:
> Hey, what was the motivation behind this? I guess it's a bit similar to what we do in kprobe with list, but also a bit different.

it matches the kernel interface where you register symbols for given path/binary
and also the common usecase is to register more symbols for binary, so the current
implementation is bit cumbersome because you always need to specify binary/path
for each symbol

> Could you update the documentation on uprobe as well? :)

yep.. I was checking on that, but we did not add any so far,
so there was nothing to fix ;-) I'll add some paragraph

mtardy:
> > Hey, what was the motivation behind this? I guess it's a bit similar to what we do in kprobe with list, but also a bit different.
> 
> it matches the kernel interface where you register symbols for given path/binary and also the common usecase is to register more symbols for binary, so the current implementation is bit cumbersome because you always need to specify binary/path for each symbol
> 
> > Could you update the documentation on uprobe as well? :)
> 
> yep.. I was checking on that, but we did not add any so far, so there was nothing to fix ;-) I'll add some paragraph

Cool thanks for the details. It was just added indeed, it's still very limited so it should be quick to update, https://tetragon.io/docs/concepts/tracing-policy/hooks/#uprobes.

jrfastab (approved):
well the generate stuff is fun.

lambdanis:
Hey, replacing a required field in the CRD is a breaking change. Technically this should mean moving to `v1alpha2` I believe. But at the very least it should be documented.

@olsajiri could you add a release note instructing users how to upgrade? It would be nice to update the examples too.

PR #1978, opened 2024-01-15
tetragon: Add support for uprobe arguments

olsajiri:
Adding support for argument values in uprobes and factor the code a bit so we could shre the arguments code with kprobes.


mtardy (approved):
Thanks! I think it would be nice to get another review as I mostly read pretty quickly the PR commit by commit but maybe missed stuff. But the tests and code look good.

But again, I wish I could write PR with commits clearly separated like that, I'm learning from this!

For a follow-up PR it would be good to improve the doc example by retrieving the bash `readline` argument to show that you can eavesdrop on someone's bash session now that it's possible.

bpf/process/types/basic.h
```diff
@@ -2451,6 +2456,10 @@ read_call_arg(void *ctx, struct msg_generic_kprobe *e, int index, int type,
 	case int_type:
 	case s32_ty:
 	case u32_ty:
+	case s16_ty:
+	case u16_ty:
+	case s8_ty:
+	case u8_ty:
```
kkourt, line 2462: Not sure of the semantics of `probe_read`, but if we read the last byte of a page where the next page is unmapped, I don't see how this wouldn't pagefault. Or am I missing something?

olsajiri: hum pagefaults are disabled, so I think it'll return -EFAULT.. but right that's corner case that will fail, I wanted to avoid adding more code in here, but perhaps it's not that horrible

kkourt: I see. I agree that it's a corner case, but part of the problem is that we don't track these errors so there is no way to tell that something went wrong.

jrfastab: for what its worth (obvious?) you'll also get EFAULT from reading user memory on a syscall. I don't think we have an error counter there either.

olsajiri: ok, I think we could have also some indication/flag that read fail for the event.. and some metric counter

kkourt: > ok, I think we could have also some indication/flag that read fail for the event.. and some metric counter

Planned as followup I presume?

kkourt (approved):
lgtm, thanks.

bpf/process/types/basic.h
olsajiri: yes, todo updated ;-)

PR #1979, opened 2024-01-15
crd: Fix type mismatch warning

tpapagian:
When we update a tracing policy (i.e. apply a policy with the same name as an existing policy), we check if the new tracing policy has the same type (i.e. TracingPolicy or TracingPolicyNamespaced) as the old one.

Now we print the warning about type mismatch even if the types are the same. This patch fixes that by returing right after the actual update.

```release-notes
Do not print a warning during tracing policy updates when both old and new tracing policies have the same type.
```

### DIFF
--- a/pkg/watcher/crd/watcher.go
+++ b/pkg/watcher/crd/watcher.go
@@ -125,10 +125,12 @@ func updateTracingPolicy(ctx context.Context, log logrus.FieldLogger, s *sensors
 		}
 	}
 
+	var err error
 	switch oldTp := oldObj.(type) {
 	case *v1alpha1.TracingPolicy:
 		newTp, ok := newObj.(*v1alpha1.TracingPolicy)
 		if !ok {
+			err = fmt.Errorf("type mismatch")
 			break
 		}
 		// FIXME: add proper DeepEquals. The resource might have different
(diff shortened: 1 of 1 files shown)


pkg/watcher/crd/watcher.go
```diff
@@ -143,6 +141,7 @@ func updateTracingPolicy(ctx context.Context, log logrus.FieldLogger, s *sensors
 			"new": newTp.TpName(),
 		}).Info("updating tracing policy")
 		update(oldTp, newTp)
+		return
```
jrfastab, line 144: hmm rather than putting more returns here maybe we could have update return an error and then if (err) ... print the warning?

pkg/watcher/crd/watcher.go
```diff
@@ -115,13 +115,11 @@ func updateTracingPolicy(ctx context.Context, log logrus.FieldLogger, s *sensors
 			log.WithError(err).WithField(
 				"old-name", oldTp.TpName(),
 			).Warnf("updateTracingPolicy: failed to remove old policy")
-			return
 		}
```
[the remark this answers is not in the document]
tpapagian: Yes, you are correct. I will put this return back.

pkg/watcher/crd/watcher.go
tpapagian: Yes, this may be a cleaner solution there. I will try to make that better.

tpapagian: Updated that a bit. The error is not due to the update but on the different types of tracing policies. I believe it is cleaner now.

jrfastab (approved):

pkg/watcher/crd/watcher.go
jrfastab: ah that is much clearer to me what the issue was. thanks.

PR #1982, opened 2024-01-16
bpf/exec: Move call to read_exe to tail call

tpapagian:
In execve events we have 2 tail calls. Now we call read_exe in the first one. The second seems to be more lightweight. So let's move it there to improve loading time.


tixxdz (approved):
Thanks Anastasios

mtardy (approved):
Thanks, I don't think it changes anything. It made more sense to be in the first tail call since we could one day replace the filename in the args with that but let's make this work :)

Issue #1983, opened 2024-01-16
Extract path from linux_binprm in security_bprm_check


dwindsor:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature you would like

With respect to [TOCTOU](https://cwe.mitre.org/data/definitions/367.html) issues, the most atomic way to block execution of a file is via the LSM hook `security_bprm_check`. This function is executed immediately before the kernel launches a new program, during its search for a handler for the format of the file being executed.

In order to write a `TracingPolicy` using the path provided to `security_bprm_check`, we need to add support to Tetragon for extracting the path from `struct linux_binprm`, the only argument to `security_bprm_check`.

# Testing

The following new `process_kprobe` was generated by running `sample-exec` with the following TracingPolicy:

```bash
dave@dev:~$ /usr/bin/sample-exec
* Before executing /usr/bin/id
/usr/bin/sample-exec: 4: /usr/bin/id: Operation not permitted
* After executing /usr/bin/id
```

## New `process_kprobe` for `security_bprm_check` events

```json
{
  "process_kprobe": {
    "process": {
      "exec_id": "OjI3ODc0MDQ4Mzc3NTczMjoxMDAxODc5",
      "pid": 1001879,
      "uid": 1000,
      "cwd": "/home/dave",
      "binary": "/usr/bin/sample-exec",
      "arguments": "/usr/bin/sample-exec",
      "flags": "execve",
      "start_time": "2024-01-11T23:27:18.871818530Z",
      "auid": 1000,
      "parent_exec_id": "OjI3ODc0MDQ4MzIwMjM0OToxMDAxODc4",
      "refcnt": 1,
      "tid": 1001879
    },
    "parent": {
      "exec_id": "OjI3ODc0MDQ4MzIwMjM0OToxMDAxODc4",
      "pid": 1001878,
      "uid": 1000,
      "cwd": "/home/dave",
      "binary": "/usr/bin/sample-exec",
      "arguments": "/usr/bin/sample-exec",
      "flags": "execve clone",
      "start_time": "2024-01-11T23:27:18.871244532Z",
      "auid": 1000,
      "parent_exec_id": "OjI3NTY3MDM5MDAwMDAwMDo5ODU1MzI=",
      "tid": 1001878
    },
    "function_name": "security_bprm_check",
    "args": [
      {
        "linux_binprm_arg": {
          "path": "/usr/bin/id"
        }
      }
    ],
[...]

PR #1986, opened 2024-01-16
tetragon: extract linux_binprm member using CO:RE 

dwindsor:
See #1983 .


tixxdz:
Thank you for adding it, so this is useful feature as we need the binprm.

One point using the path like that for binprm context will limit us in the future, so binprm handles the binary being executed, the interpreter, and executable IIRC how it is called if the execve could have happened from an fd passed file... I will check later to see again the details, but just to say we need one layer maybe of abstraction since strictly speaking path here could be one of those depending on how which stage you are executing the binary and its format. Or we could rename the path to something more explicit, maybe...

Anyway this good feature as right now we have this https://github.com/cilium/tetragon/blob/main/examples/tracingpolicy/process-exec/process-exec-elf-begin.yaml but it just catches the late binary elf or flat execution not other misc or shebang execution, so it was on our todo list for a while ;-)

We will review, just wait more for feedback before doing extra changes in cases ;-)

mtardy:
> Thank you for adding it, so this is useful feature as we need the binprm.
> 
> One point using the path like that for binprm context will limit us in the future, so binprm handles the binary being executed, **not** the interpreter, and executable IIRC how it is called if the execve could have happened from an fd passed file... I will check later to see again the details, but just to say we need one layer maybe of abstraction since strictly speaking path here could be one of those depending on how which stage you are executing the binary and its format. Or we could rename the path to something more explicit, maybe...
> 
> Anyway this good feature as right now we have this https://github.com/cilium/tetragon/blob/main/examples/tracingpolicy/process-exec/process-exec-elf-begin.yaml but it just catches the late binary elf or flat execution not other misc or shebang execution, so it was on our todo list for a while ;-)
> 
> We will review, just wait more for feedback before doing extra changes in cases ;-)

Hey @tixxdz can we maybe have the conversation on the issue as I started here and had similar argument as you have. See https://github.com/cilium/tetragon/issues/1983#issuecomment-1894227101.

dwindsor:
> Thank you for adding it, so this is useful feature as we need the binprm.
> 
> One point using the path like that for binprm context will limit us in the future, so binprm handles the binary being executed, **not** the interpreter, and executable IIRC how it is called if the execve could have happened from an fd passed file... I will check later to see again the details, but just to say we need one layer maybe of abstraction since strictly speaking path here could be one of those depending on how which stage you are executing the binary and its format. Or we could rename the path to something more explicit, maybe...
> 
> Anyway this good feature as right now we have this https://github.com/cilium/tetragon/blob/main/examples/tracingpolicy/process-exec/process-exec-elf-begin.yaml but it just catches the late binary elf or flat execution not other misc or shebang execution, so it was on our todo list for a while ;-)
> 
> We will review, just wait more for feedback before doing extra changes in cases ;-)

The only change I will maybe make is to fix the CI failure that for some reason did not appear in my personal branch =)

bpf/lib/bpf_linux_binprm.h
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* Copyright Authors of Tetragon */
+
```
jrfastab, line 3: bit of a nit, but why is this in ./bpf/lib can we put it with the other types ./bpf/process/types/bpf_linux_binprm.h?

bpf/process/types/basic.h
```diff
@@ -2378,6 +2383,12 @@ read_call_arg(void *ctx, struct msg_generic_kprobe *e, int index, int type,
 	case iov_iter_type:
 		size = copy_iov_iter(ctx, orig_off, arg, argm, e, data_heap);
 		break;
+	case linux_binprm_type: {
+		struct linux_binprm *bprm = (struct linux_binprm *)arg;
+		arg = (unsigned long)(&bprm->file);
+
+		// fallthrough to file_ty
```
tpapagian, line 2390: I don't believe that this is correct here. You expect to go through `linux_binprm_type`, `file_ty`, and `path_ty` but as it is now, my understanding is that it goes through `linux_binprm_type`, `kiocb_type`, `file_ty`, and `path_ty`. `kiocb_type` affects the value of arg which can cause issues.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -55,7 +55,7 @@ type KProbeArg struct {
 	// +kubebuilder:validation:Minimum=0
 	// Position of the argument.
 	Index uint32 `json:"index"`
-	// +kubebuilder:validation:Enum=auto;int;uint32;int32;uint64;int64;char_buf;char_iovec;size_t;skb;sock;string;fd;file;filename;path;nop;bpf_attr;perf_event;bpf_map;user_namespace;capability;kiocb;iov_iter;cred;load_info;module;syscall64;
+	// +kubebuilder:validation:Enum=auto;int;uint32;int32;uint64;int64;char_buf;char_iovec;size_t;skb;sock;string;fd;file;filename;path;nop;bpf_attr;perf_event;bpf_map;user_namespace;capability;kiocb;iov_iter;cred;load_info;module;syscall64;linux_binprm
```
tpapagian, line 58: As you are changing the CRD you have also to update the version of that in https://github.com/cilium/tetragon/blob/64e5baf632be3ab62b2335f1f21c75efce5a762f/pkg/k8s/apis/cilium.io/v1alpha1/version.go#L10 (i.e. from 1.1.3 to 1.1.4).

bpf/process/types/basic.h
```diff
@@ -2378,6 +2383,12 @@ read_call_arg(void *ctx, struct msg_generic_kprobe *e, int index, int type,
 	case iov_iter_type:
 		size = copy_iov_iter(ctx, orig_off, arg, argm, e, data_heap);
 		break;
+	case linux_binprm_type: {
+		struct linux_binprm *bprm = (struct linux_binprm *)arg;
+		arg = (unsigned long)(&bprm->file);
```
tpapagian, line 2388: I believe that you should use CORE here as the linux_binprm can have different layout in different kernels. Something like `arg = (unsigned long)_(&bprm->file);` in a similar way to the line 2396 would be fine.

tpapagian (commented):
Thanks! I have added some comments. There are also some errors in the CI, it would be great if you can fix them. Let me know if anything does not make sense. 

I would also like to have a test for that. You can use https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/kprobe_test.go as an inspiration. Something like defining a tracing policy to use that, execute a file, and then check that the event that you are getting have the correct fields.

mtardy:
> I would also like to have a test for that. You can use https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/kprobe_test.go as an inspiration. Something like defining a tracing policy to use that, execute a file, and then check that the event that you are getting have the correct fields.

I also think this is important and you can do perfring tests, it will be more efficient than most of the tests out there, as example, you could use:
https://github.com/cilium/tetragon/blob/56d978b4d4e31e8a4918189ba1eec97bd3ed5518/pkg/sensors/tracing/kprobe_test.go#L3764-L3841

I can help you if you don't get something in the test.

bpf/process/types/basic.h
dwindsor: Thanks!

I've never seen `_` used this way! Is this something particular to Tetragon? I expected to see `BPF_CORE_READ*` et al for CO:RE ops =)

dwindsor:
> > I would also like to have a test for that. You can use https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/kprobe_test.go as an inspiration. Something like defining a tracing policy to use that, execute a file, and then check that the event that you are getting have the correct fields.
> 
> I also think this is important and you can do perfring tests, it will be more efficient than most of the tests out there, as example, you could use:
> 
> https://github.com/cilium/tetragon/blob/56d978b4d4e31e8a4918189ba1eec97bd3ed5518/pkg/sensors/tracing/kprobe_test.go#L3764-L3841
> 
> I can help you if you don't get something in the test.

@mtardy do you want to convert this to a suggestion so we can use it directly?

bpf/process/types/basic.h
mtardy: This is a macro you can find in `bpf/lib/bpf_helpers.h`:
```C
#define _(P) (__builtin_preserve_access_index(P))
```

mtardy:
> @mtardy do you want to convert this to a suggestion so we can use it directly?

Not sure I understand, do you want me to write the tests?

Here is a more helpful hint maybe:

```Go
 func typeLinuxBinprmPerfringTest(t *testing.T, operator string, values []string) { 
 	testutils.CaptureLog(t, logger.GetLogger().(*logrus.Logger)) 
 	defer cancel() 
  
 	if err := observer.InitDataCache(1024); err != nil { 
 		t.Fatalf("observertesthelper.InitDataCache: %s", err) 
 	} 
  
 	option.Config.HubbleLib = tus.Conf().TetragonLib 
 	tus.LoadSensor(t, base.GetInitialSensor()) 
 	tus.LoadSensor(t, testsensor.GetTestSensor()) 
 	sm := tus.GetTestSensorManager(ctx, t) 
  
 	policy := tracingpolicy.GenericTracingPolicy{ 
 		Metadata: v1.ObjectMeta{ 
 			Name: "match-binaries", 
 		}, 
 		Spec: v1alpha1.TracingPolicySpec{ 
 			[...] // <-- spec using the type
 		}, 
 	} 
  
 	err := sm.Manager.AddTracingPolicy(ctx, &policy) 
 	assert.NoError(t, err) 
  
 	var tailPID, headPID int 
 	ops := func() { 
           // <-- trigger the hook
 	} 
 	events := perfring.RunTestEvents(t, ctx, ops) 
  
 	for _, ev := range events { 
 		if kprobe, ok := ev.(*tracing.MsgGenericKprobeUnix); ok { 
 			// <-- verify you get the type value/matching whatever correct
 		} 
 	} 
 } 
 ```

**Comment by dwindsor** [comment]
> > @mtardy do you want to convert this to a suggestion so we can use it directly?
> 
> Not sure I understand, do you want me to write the tests?
> 
> Here is a more helpful hint maybe:
> 
> ```go
>  func typeLinuxBinprmPerfringTest(t *testing.T, operator string, values []string) { 
>  	testutils.CaptureLog(t, logger.GetLogger().(*logrus.Logger)) 
>  	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime) 
>  	defer cancel() 
>   
>  	if err := observer.InitDataCache(1024); err != nil { 
>  		t.Fatalf("observertesthelper.InitDataCache: %s", err) 
>  	} 
>   
>  	option.Config.HubbleLib = tus.Conf().TetragonLib 
>  	tus.LoadSensor(t, base.GetInitialSensor()) 
>  	tus.LoadSensor(t, testsensor.GetTestSensor()) 
>  	sm := tus.GetTestSensorManager(ctx, t) 
>   
>  	policy := tracingpolicy.GenericTracingPolicy{ 
>  		Metadata: v1.ObjectMeta{ 
>  			Name: "match-binaries", 
>  		}, 
>  		Spec: v1alpha1.TracingPolicySpec{ 
>  			[...] // <-- spec using the type
>  		}, 
>  	} 
>   
>  	err := sm.Manager.AddTracingPolicy(ctx, &policy) 
>  	assert.NoError(t, err) 
>   
>  	var tailPID, headPID int 
>  	ops := func() { 
>            // <-- trigger the hook
>  	} 
>  	events := perfring.RunTestEvents(t, ctx, ops) 
>   
>  	for _, ev := range events { 
>  		if kprobe, ok := ev.(*tracing.MsgGenericKprobeUnix); ok { 
>  			// <-- verify you get the type value/matching whatever correct
>  		} 
>  	} 
>  } 
> ```

My fault, I thought you had written the tests, but only as a comment.

File: bpf/process/types/basic.h (hunk 2 above)
Comment (dwindsor) [reply bpf/process/types/basic.h]: Thanks! I've switched this around a bit to be (hopefully) correct now!

**Comment by mtardy** [comment]
> @mtardy tests added, but they're failing for an unknown reason

I see you are pushing a lot of commits, I'll prefer taking a look when it's stable, just tell me :)

**Comment by dwindsor** [comment]
> I see you are pushing a lot of commits, I'll prefer taking a look when it's stable, just tell me :)

I think it's okay to go now, sorry for the many commits! I now know that CI needs to be manually triggered, so I'll reach out in future before spamming the project with PRs.

I think [`d02c9f20`](https://github.com/cilium/tetragon/pull/1986/commits/d02c9f20dbaa756823e92fcc4550650cd2f99977) fixed the failing CI test! At least it does so on my system:

```bash
dave@dev:~/tetragon$ sudo go test -count=1 ./pkg/sensors/tracing -run TestLinuxBinprmExtractPath
[sudo] password for dave:
ok  	github.com/cilium/tetragon/pkg/sensors/tracing	1.000s
```

mtardy:
Hey, [checkpatch](https://github.com/cilium/tetragon/actions/runs/7629322014/job/20784195222?pr=1986) gives you details on what could be done for each commit to make the PR ready to be merged (we just rebase and merge)? 

For that could you maybe rebase on main and also rebase your branch to squash some commits. I guess you could keep a few, don't squash everything into one but we don't need commits like "whitespace" or "go fmt" when we merge. See the "logical commit" explanation here https://tetragon.io/docs/contribution-guide/making-changes/, for example here you can still separate bpf/api/tests changes maybe.

dwindsor:
> Hey, [checkpatch](https://github.com/cilium/tetragon/actions/runs/7629322014/job/20784195222?pr=1986) gives you details on what could be done for each commit to make the PR ready to be merged (we just rebase and merge)?
> 
> For that could you maybe rebase on main and also rebase your branch to squash some commits. I guess you could keep a few, don't squash everything into one but we don't need commits like "whitespace" or "go fmt" when we merge. See the "logical commit" explanation here https://tetragon.io/docs/contribution-guide/making-changes/, for example here you can still separate bpf/api/tests changes maybe.

Sure! Rather than rebasing/squashing individual commits in this 2 week-old branch, is it okay if I just recreate this PR on top of today's `main`? I've done so in #2028 , it might prove easier to follow in the end?.

bpf/process/types/basic.h
```diff
@@ -2398,6 +2392,12 @@ read_call_arg(void *ctx, struct msg_generic_kprobe *e, int index, int type,
 		arg = (unsigned long)file;
 	}
 		// fallthrough to file_ty
```
jrfastab, line 2394: but now this will rewrite the arg from kiocb_type that is expecting to also  fall through? I think my abuse of fallthroughs has reached its limit of usability? Its a bit tricky to get this good on older kernels. We don't want to inline the code in duplicate chunks repeating here for insns count reasons.

Maybe an inline would get optimized to a goto by compiler so it would be ok? Other option would be a goto to file handler. Anyways as is it needs a fix right?

dwindsor: Hmm, if there are no cases where `struct linux_binrpm` wraps a `struct kiocb`, we should be okay, no?

dwindsor: Bump @jrfastab. 

`needs-rebase` got added, fully 0 point in doing another rebase (it'll be the 4th since this PR was opened, I think) until I get feedback on this comment.

mtardy:
Hey looks like you rebased your PR, cool, however some of your commits have wrong names/wrong content:
- https://github.com/cilium/tetragon/pull/1986/commits/c62c8ccbe8cb38b2972de040249639dc78fc8590
- https://github.com/cilium/tetragon/pull/1986/commits/fb64858b94ec2c0674c21030d7e5ceb4074ac903

Like those two at least.

dwindsor:
> Hey looks like you rebased your PR, cool, however some of your commits have wrong names/wrong content:
> 
> * [c62c8cc](https://github.com/cilium/tetragon/commit/c62c8ccbe8cb38b2972de040249639dc78fc8590)
> * [fb64858](https://github.com/cilium/tetragon/commit/fb64858b94ec2c0674c21030d7e5ceb4074ac903)
> 
> Like those two at least.

Thanks, renamed/squashed some commits.

jrfastab (changes requested):

jrfastab:
@dwindsor Just to add a summary here because I like the PR and want to get it merged, but needs a couple updates.

1. Take a look at the comment about that fallthrough code it looks like it has potential to break things. My suggestion would be to just avoid the fallthrough altogether. I was doing falltrhough logic early on before compiler got smarter to save insns count. I don't think its actually needed now. I'm would lean more towards making compiler smarter if we need to.

2. If you can roll this patch "tetragon: extract linux_binprm member using CO:RE" back into where you add the code just so our git history stays as clean as possible. This way we avoid introducing a bug and then immediately fixing it.

Thanks! Feel free to at me here if you have any questions. I'm traveling next week but can likely find time to read a few PRs. Happy to help.

dwindsor:
> @dwindsor Just to add a summary here because I like the PR and want to get it merged, but needs a couple updates.
> 

Thanks for such a helpful summary @jrfastab! It's been a long thread =).

> 1. Take a look at the comment about that fallthrough code it looks like it has potential to break things. My suggestion would be to just avoid the fallthrough altogether. I was doing falltrhough logic early on before compiler got smarter to save insns count. I don't think its actually needed now. I'm would lean more towards making compiler smarter if we need to.

Done. I moved the `stuct linux_binprm` handling code in `read_call_arg` to after the `struct file`-related cases with fallthroughs.

> 2. If you can roll this patch "tetragon: extract linux_binprm member using CO:RE" back into where you add the code just so our git history stays as clean as possible. This way we avoid introducing a bug and then immediately fixing it.
> 

Done. I squashed all commits into this single commit, I hope this is okay!

> Thanks! Feel free to at me here if you have any questions. I'm traveling next week but can likely find time to read a few PRs. Happy to help.

dwindsor:
CI is failing for something unrelated: https://github.com/cilium/tetragon/actions/runs/7803935180/job/21306442590?pr=1986#step:6:1534

> ❌	pkg.sensors.tracing.TestProcessSetCap							(total:1 failed:1 skipped:0)	8.626799908s	12m51.114887622s

tixxdz:
> CI is failing for something unrelated: https://github.com/cilium/tetragon/actions/runs/7803935180/job/21306442590?pr=1986#step:6:1534
> 
> > ❌	pkg.sensors.tracing.TestProcessSetCap							(total:1 failed:1 skipped:0)	8.626799908s	12m51.114887622s

Hmm I tried to see what changed but didn't spot anything, thanks for the quick rebase, let's see if those new debug messages spot anything. The problem we are missing some events from that test. I guess it is not related to your patches. I will shortly run your branch locally and see.

Much appreciated for all the efforts ;-) !

pkg/grpc/tracing/tracing.go
```diff
@@ -248,6 +248,11 @@ func getKprobeArgument(arg tracingapi.MsgGenericKprobeArg) *tetragon.KprobeArgum
 		a.Label = e.Label
 	case api.MsgGenericKprobeArgCapEffective:
 		a.Arg = &tetragon.KprobeArgument_CapEffectiveArg{CapEffectiveArg: caps.GetCapabilitiesHex(e.Caps)}
```
tixxdz, line 250: So here you remove the a.Label of the CapEffective. This needs to be fixed.

pkg/sensors/tracing/args.go
```diff
@@ -493,6 +493,18 @@ func getArg(r *bytes.Reader, a argPrinter) tracingapi.MsgGenericKprobeArg {
 		}
 
 		arg.Index = uint64(a.index)
```
tixxdz, line 495: ok so the test failing TestProcessSetCap is due to this change here, the arg.Label is missing plus the return arg, which make it miss the effective caps argument

tixxdz (changes requested):
Ok , see comments why the test s failing, thank you

pkg/grpc/tracing/tracing.go
dwindsor: Fixed, thanks

pkg/sensors/tracing/args.go
dwindsor: Fixed, thanks

tixxdz (approved):
Much appreciated! lgtm assuming all tests pass.

There are still things that we can improve in the test as an example, but as long as we get the event personally I'm fine, so I would say let's merge it, and we may improve or change some minor stuff before releasing maybe... ;-)

dwindsor:
> Much appreciated! lgtm assuming all tests pass.
> 
> There are still things that we can improve in the test as an example, but as long as we get the event personally I'm fine, so I would say let's merge it, and we may improve or change some minor stuff before releasing maybe... ;-)

Kudos to you for the help along the way! Still one [CI failure](https://github.com/cilium/tetragon/actions/runs/7820507657/job/21340402633?pr=1986), but only for the kernel `4.19` case:

```
154/507 tests failed 😞 (took: 12m17.91s, skipped:74)
```

 This one is odd because the other tests in this suite succeeded...

tixxdz:
> > Much appreciated! lgtm assuming all tests pass.
> > There are still things that we can improve in the test as an example, but as long as we get the event personally I'm fine, so I would say let's merge it, and we may improve or change some minor stuff before releasing maybe... ;-)
> 
> Kudos to you for the help along the way! Still one [CI failure](https://github.com/cilium/tetragon/actions/runs/7820507657/job/21340402633?pr=1986), but only for the kernel `4.19` case:
> 
> ```
> 154/507 tests failed 😞 (took: 12m17.91s, skipped:74)
> Error: Process completed with exit code 1.
> ```
> 
> This one is odd because the other tests in this suite succeeded...

Ok this one is probably a kernel bpf limitation, triggered by bpf verifier and errors could be a bit generic... :-/ 

So usually how we reproduce and track those is documented here: https://github.com/cilium/tetragon/tree/main/tests/vmtests ,  you may need to run with sudo when starting the vm.

We have the verbose 3 flag option to dump verifier logs in Tetragon, but can't recall how to use that in tests. So how I usually do it is I compile tetragon on host, run the vm with old kernel. Check `tetragon-vmtests-run --help` it has a flag `--just-boot` where then you login as root and run tetragon with `--tracing-policy yournewpolicy.yaml --verbose 3` to get a dump of bpf errors.

Could be two things: you have a buffer with 1024 MAX_STRING what happens if you make it 128 or 256 , or those fallthrough we have are some bpf optimizations and I guess we just reached the limit and we may be forced to work this part of the code unfortunately... :-/  

Also can you run the test on 4.9 with a small buffer?

dwindsor:
> > > Much appreciated! lgtm assuming all tests pass.
> > > There are still things that we can improve in the test as an example, but as long as we get the event personally I'm fine, so I would say let's merge it, and we may improve or change some minor stuff before releasing maybe... ;-)
> > 
> > 
> > Kudos to you for the help along the way! Still one [CI failure](https://github.com/cilium/tetragon/actions/runs/7820507657/job/21340402633?pr=1986), but only for the kernel `4.19` case:
> > ```
> > 154/507 tests failed 😞 (took: 12m17.91s, skipped:74)
> > Error: Process completed with exit code 1.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This one is odd because the other tests in this suite succeeded...
> 
> Ok this one is probably a kernel bpf limitation, triggered by bpf verifier and errors could be a bit generic... :-/
> 
> So usually how we reproduce and track those is documented here: https://github.com/cilium/tetragon/tree/main/tests/vmtests , you may need to run with sudo when starting the vm.
> 
> We have the verbose 3 flag option to dump verifier logs in Tetragon, but can't recall how to use that in tests. So how I usually do it is I compile tetragon on host, run the vm with old kernel. Check `tetragon-vmtests-run --help` it has a flag `--just-boot` where then you login as root and run tetragon with `--tracing-policy yournewpolicy.yaml --verbose 3` to get a dump of bpf errors. You will also need to pass the `--btf tests/vmtests/test-data/kernels/4.19/boot/btf-4.19.262` to tetragon.
> 
> Could be two things: you have a buffer with 1024 MAX_STRING what happens if you make it 128 or 256 , or those fallthrough we have are some bpf optimizations and I guess we just reached the limit and we may be forced to work this part of the code unfortunately... :-/
> 
> Also can you run the test on 4.9 with a small buffer? just follow https://github.com/cilium/tetragon/tree/main/tests/vmtests for 4.19 to run your single test is a good start.

Thanks for the tips!

I'm trying to run the tests as you mentioned, by running `tetragon-vm-tests` with `--just-boot`:

```
dave@dev:~/src/github.com/dwindsor/tetragon$ sudo ./tests/vmtests/tetragon-vmtests-run --kernel tests/vmtests/test-data/kernels/4.19/boot/vmlinuz-4.19.262 --base tests/vmtests/test-data/images/base.qcow2 --qemu-disable-kvm --just-boot
```

Logging in to the VM as root, we can see that qemu has started the proper image:

```
root@tetragon:~# uname -a
Linux tetragon 4.19.262 #1 SMP Thu Oct 27 16:52:10 UTC 2022 x86_64 GNU/Linux
```

However, when I'm in the VM, I can't find a `tetragon` binary, only `tetragon-tester` which doesn't appear to take any command-line options:

```
root@tetragon:~# sudo /usr/sbin/tetragon-tester --help
Linux tetragon 4.19.262 #1 SMP Thu Oct 27 16:52:10 UTC 2022
```

Is this expected? 🤔

dwindsor:
Hey @tixxdz, I was actually able to run tetragon as you suggested, on 4.19 using vmtests/qemu.

The verifier error we're getting here is: `load program: argument list too long`

I tried changing buffer sizes to 512, 256, 64 and 8, but the same error occurs in each case. I've gone ahead and changed the length of the buffer back to `MAX_STRING`, because this doesn't seem to be the cause.


Tracing policy:
```yaml
kind: TracingPolicy
metadata:
  name: "sample-no-exec-id"
spec:
  kprobes:
  - call: "security_bprm_check"
    syscall: false
    args:
    - index: 0
      type: "linux_binprm"
    returnArg:
      index: 0
      type: "int"
    selectors:
      - matchArgs:
          - index: 0
            operator: "Equal"
            values:
              - "/usr/bin/sample-exec"
```
Testing log: [vmtests-4.19-binprm.txt](https://github.com/cilium/tetragon/files/14271533/vmtests-4.19-binprm.txt)

dwindsor:
The verifier for 4.19 returns [E2BIG](https://github.com/torvalds/linux/blob/84df9525b0c27f3ebc2ebb1864fa62a97fdedb7d/kernel/bpf/verifier.c#L4187) in only a handful of places...

Interestingly, the 1million insn limit was introduced in 5.2...

dwindsor:
Thanks for the tips! The `#ifdef` fixes the failing test for 4.19.

> also it'd be great to split the changes and have at least bpf and pkg changes in separate commits, plus could you please add some example under examples/tracingpolicy ?
> 

Would be happy to do this, but IIUC this seems to contradict with an earlier [request](https://github.com/cilium/tetragon/pull/1986#issuecomment-1924482348) by @jrfastab . 

> thanks

Thanks!

tixxdz:
> The verifier for 4.19 returns [E2BIG](https://github.com/torvalds/linux/blob/84df9525b0c27f3ebc2ebb1864fa62a97fdedb7d/kernel/bpf/verifier.c#L4187) in only a handful of places...
> 
> Interestingly, the 1million insn limit was introduced in 5.2, which makes sense given our results (tests are passing on kernels >= 5.14). Bummer we have to support < 5.2 =(.

Yes we have some users on old kernels

tixxdz:
@olsajiri we can surely merge this as it is now, then we split the read_call_arg() stuff later?  Also please let me know if you plan todo it, cause there is also another PR by a contributor that could be affected after this change, this way you save us the trouble ;-)

jrfastab:
@dwindsor Generally as a matter of style we try to break commits up between BPF and userspace when it makes sense. My comment was intended to be specific to the single patch that just added CO-RE annotations to the reads in BPF. I didn't want to have BPF patch without the CORE style lookups and then add it later because if we bisect for some reason at that point it could introduce a bug. The bug being without CO-RE you will do some fixed offset lookup that probably only works correctly on some subset of kernels.

jrfastab:
@dwindsor I'm OK to take this as a single patch. Just take it as a hint for next time.

dwindsor:
Does anyone know what this CI error means:

https://github.com/cilium/tetragon/actions/runs/7920381909/job/21631895710?pr=1986#step:7:413

dwindsor:
> Does anyone know what this CI error means:
> 
> https://github.com/cilium/tetragon/actions/runs/7920381905/job/21630651516?pr=1986

This is complaining about a broken link in README.md that's not even touched by this PR. The link is broken in README.md `HEAD`.

tixxdz:
One of the CI test failed, I tracked it here: https://github.com/cilium/tetragon/issues/2110 and scheduled another run, thanks

jrfastab:
Thanks a lot! Looks good.

jrfastab (approved):

PR #1996, opened 2024-01-18
metrics: Do not process the same map multiple times

tpapagian:
For the `tetragon_map_in_use_gauge` metric, we iterate through `sensors.AllMaps` to provide their size and errors.

Depending on the sensors that are loaded and the dependencies between them, `sensors.AllMaps` may have the same map name multiple times. This can cause errors like:
`collected metric "XXX" {...} was collected before with the same name and label values`

To avoid that we keep a map and process each map only once.

```release-notes
Process each map only once during the calculation of tetragon_map_in_use_gauge.
```


tixxdz (approved):
Thank you ;-)

lambdanis (approved):
> Depending on the sensors that are loaded and the dependencies between them, sensors.AllMaps may have the same map name multiple times.

Huh, I haven't realized that, thanks!

PR #2002, opened 2024-01-19
tetragon: Add support to use security_ functions in killer

olsajiri:
Adding support to attach killer to security_ functions and use NotifyKiller from kprobe. It's used to override specific security_* function in case we want to ensure we bypass syscall processing before sigkill is delivered.


pkg/sensors/tracing/generickprobe.go
```diff
@@ -355,6 +355,9 @@ func createMultiKprobeSensor(sensorPath string, multiIDs []idtable.EntryID) ([]*
 		maps = append(maps, socktrack)
 	}
 
+	killerDataMap := program.MapBuilderPin("killer_data", "killer_data", load)
```
kkourt, line 358: Let's use https://github.com/cilium/tetragon/blob/506677dc9b70f7d87fe23ac4cfcb2e45481e0ee1/pkg/sensors/tracing/killer.go#L24?

pkg/sensors/tracing/killer.go
```diff
@@ -169,18 +169,32 @@ func (kh *killerHandler) createKillerSensor(
...
-		pfxSym, err := arch.AddSyscallPrefix(sym)
-		if err != nil {
-			return nil, err
+		kh.syscallsSyms = append(kh.syscallsSyms, sym)
+	}
+
+	var err error
+
+	// fix syscalls
+	for idx, sym := range kh.syscallsSyms {
```
kkourt, line 190:
```
  killers:
  - syscalls:
    - "security_file_permission"
```

I think we should rename `syscalls` to `calls` (or similar) since it's not just syscalls anymore.

kkourt (commented):
Thanks :) 

Had a quick look and added some first comments.

dwindsor:
> killer

Just curious, how is this different from killing a process inline by overriding the return value of one of the LSM calls to be non-zero?

olsajiri:
> > killer
> 
> Just curious, how is this different from killing a process inline by overriding the return value of one of the LSM calls to be non-zero?

not too much different, this way we get invoked earlier directly from security_* function call, lsm hook would get invoked later on, but that would not really matter, it'd work the same I think.. another thing is we don't need to depend on lsm being enabled

pkg/sensors/tracing/generickprobe.go
olsajiri: nice, did not see that, thanks

pkg/sensors/tracing/killer.go
olsajiri: ok

jrfastab:
Also its non-obvious that all syscalls map to a security hook. Many will but not sure its 1:1 without auditing.

jrfastab (approved):

pkg/sensors/tracing/killer.go
```diff
@@ -125,7 +125,10 @@ func selectOverrideMethod(specOpts *specOptions) (OverrideMethod, error) {
 			return OverrideMethodInvalid, fmt.Errorf("option override return set, but it is not supported")
 		}
 	case OverrideMethodFmodRet:
-		if !bpf.HasModifyReturnSyscall() {
+		if hasSyscall && !bpf.HasModifyReturnSyscall() {
+			return OverrideMethodInvalid, fmt.Errorf("option fmod_ret on syscall set, but it is not supported")
+		}
+		if hasNormal && !bpf.HasModifyReturn() {
 			return OverrideMethodInvalid, fmt.Errorf("option fmod_ret set, but it is not supported")
```
kkourt, line 132: I don't think the `hasNormal` check is needed.

AFIACT, if `HadModifyReturnSyscall()` is true , `HasModifyReturn()` is (always)  true.
So I think the following check:

```
if !bpf.HasModifyReturn() || (hasSyscall && !bpf.HasModifyReturnSyscall()) {
     return err
}
``` 

catches all the cases we care about, without the extra argument.

kkourt (commented):
Thanks!

tixxdz:
I didn't follow it closely but I have one question: would it be possible to just fmod_ret the security_ hook without killing the process? is the old behavior still supported? cause there is valid software there that is happy with -EPERM and continues to work as expected, delivering sigkill for such cases is too much and could break.

Also extra note didn't check code but as John noted is this planning a 1:1 mapping? the security hooks don't really map to syscalls, there are syscalls->operations (different context)-> lead to same security hook, and other cases where one syscall is handled through multiple security hooks according to context. The security hooks are more about the context of the process, actually constructing the context if we check that in details, then acting when necessary, where in syscall you may act in advance without context at all... The mapping could work for some read/write fs operation but for other cases where example the kernel acts on behalf is different story

olsajiri:
> I didn't follow it closely but I have one question: would it be possible to just fmod_ret the security_ hook without killing the process? is the old behavior still supported? cause there is valid software there that is happy with -EPERM and continues to work as expected, delivering sigkill for such cases is too much and could break.

yes, this change just adds functionality, does not remove any, it adds the possibility to attach killer to security_* functions

> 
> Also extra note didn't check code but as John noted is this planning a 1:1 mapping? the security hooks don't really map to syscalls, there are syscalls->operations (different context)-> lead to same security hook, and other cases where one syscall is handled through multiple security hooks according to context. The security hooks are more about the context of the process, actually constructing the context if we check that in details, then acting when necessary, where in syscall you may act in advance without context at all... The mapping could work for some read/write fs operation but for other cases where example the kernel acts on behalf is different story

the idea behind this change is to be able to kill process that's about to execute a syscall and make sure the syscall is not executed by attaching extra killer probe to security function (because there could be kernel configs where you can't override syscall directly) ... it's up to the user to pick up which security function to hook for which syscall, there's nothing hardcoded

pkg/sensors/tracing/killer.go
olsajiri: right, I guess I wanted it to be more explicit.. I'll change that

mtardy (approved):
thanks looks good! :)

kkourt (approved):
Thanks!

PR #2007, opened 2024-01-22
bugtool: fetch gops pprof heap

tixxdz:


pkg/bugtool/bugtool.go
```diff
@@ -83,6 +84,14 @@ func doSaveInitInfo(fname string, info *InitInfo) error {
 		logger.GetLogger().WithField("bpftool", info.BpfToolPath).Info("Successfully detected bpftool path")
 	}
 
+	gops, err := exec.LookPath("gops")
```
kkourt, line 89: Why do we need to do this on the agent?

tixxdz: cause IIRC the agent has the path of installation in systemd , like bpftool so it finds the path for later

tixxdz: So, we get the path in the agent we save it into the tetragon json inside /run/tetragon/, and later tetra tries to auto read it allowing users to get dump without specifying the search path or the full gops path.

pkg/bugtool/bugtool.go
```diff
@@ -498,11 +511,26 @@ func (s *bugtoolInfo) addBpftoolInfo(tarWriter *tar.Writer) {
...
+	_, err := os.Stat(s.info.GopsPath)
+	if err != nil {
+		s.multiLog.WithField("gops-address", s.info.GopsAddr).WithError(err).Warn("Failed to locate gops, please install it.")
+		return
+	}
+
+	s.multiLog.WithField("gops-address", s.info.GopsAddr).WithField("gops-path", s.info.GopsPath).Info("Dumping gops information")
+
+	s.execCmd(tarWriter, "gops.stack", s.info.GopsPath, "stack", s.info.GopsAddr)
+	s.execCmd(tarWriter, "gpos.stats", s.info.GopsPath, "stats", s.info.GopsAddr)
```
kkourt, line 561: nit: could you also fix the typo here please? gpos -> gops

kkourt (approved):
LGTM, thanks!

pkg/bugtool/bugtool.go
tixxdz: good catch, will do ;-)

jrfastab:
gops looks relatively small <10MB so should be ok.

PR #2009, opened 2024-01-23
Perf: Remove extra copies of data from ring buffer

kevsecurity:
Most sensors send data from kernel to user space via the perf ring buffer. They typically copy it into a struct, and copy it from that struct to another "Unix" struct. This series removes the second copy by replacing the duplicate members in the Unix struct with a pointer to the original struct. This should improve perf.


lambdanis (approved):
Looks good to me, thanks! If you have some benchmarks comparing the resources usage before/after these changes, it would be interesting to see.

PR #2024, opened 2024-01-26
tetragon: Harden loader sensor

olsajiri:
Making the loader sensor to send data for non excutable mmaps as well
to increase the chance that we might be able to read build id for the
binary at least from one of them.
    
As we can't sleep at the current loader probed function, we depend on
the page with build id being swapped-in, which might not be always the
case. By allowing to send build id data for mmap non executable mmap
events we increase the chance of getting the build id data for binary.


willfindlay (approved):
Makes sense to me

kkourt (approved):

PR #2026, opened 2024-01-26
api:proto: rename Inode to InodeProperties

tixxdz:
Be consistent with FileProperties as this abstracts kernel 'file' and 'inode' objects.


mtardy (approved):
Make sense! Thanks!

PR #2031, opened 2024-01-27
validation: improve error message when failing to load a TracingPolicy

### LINKED ISSUE #2022: validation: improve error message when failing to load a TracingPolicy
Users are faced with this message if they load a TracingPolicy that fails to satisfy the basics of metadata:
```
level=fatal msg="Failed to start tetragon" error="could not find validator for: /, Kind="
```
Tetragon tries to find a validator for an empty `Kind` since it's not specified if the file is empty by mistake for example.

This could be improved by adding some context while returning the error to the callers to emphasize that the issue is a tracing policy file that is invalid.

### Ideas

I see mainly here:
https://github.com/cilium/tetragon/blob/6644bbd6f7165731c9829e718f867c48f17c54a9/pkg/tracingpolicy/generictracingpolicy.go#L285-L291

Where we could do something like this:
```diff
policy, err := os.ReadFile(path) 
if err != nil { 
	return nil, err 
} 
- FromYAML(string(policy))
+ tp, err := FromYAML(string(policy))
+ if err != nil {
+    return nil, fmt.Errorf("failed loading tracing policy file %q: %w", path, err)
+ }
+ return tp, nil
```

You could also add more context here, but I'm not sure if you need it (it would not have helped in the empty file case).
https://github.com/cilium/tetragon/blob/6644bbd6f7165731c9829e718f867c48f17c54a9/pkg/tracingpolicy/generictracingpolicy.go#L273-L276
```diff
validationResult, err := ValidateCRD(policy)
if err != nil {
- 	return nil, err
+ 	return nil, fmt.Errorf("failed to validate policy %q: %w", policy.GetMetadata().Name, err)
}
```

christian-2:
Callers of `tracingpolicy.FromFile()` no longer receive highly specific errors, but a generic error, which can also be presented to end users. Callers of `tracingpolicy.FromJSON()` still receive the specific errors, and any generic error wraps a specific one.

The implementation is along the lines suggested by @mtardy in #2022. (Thanks for that input.) Please indicate if my use of `Panicf()` in `pkg/tracingpolicy/generictracingpolicy_test.go` is unwelcome.

BTW, it seems to me that `drop-privileges` should be added to `contrib/tester-progs/.gitignore`. I could create another PR for that, if applicable (plus an issue, if necessary for the trivial change).

Fixes #2022

Signed-off-by: Christian Hörtnagl <christian2@univie.ac.at>

### DIFF
--- a/pkg/tracingpolicy/generictracingpolicy_test.go
+++ b/pkg/tracingpolicy/generictracingpolicy_test.go
@@ -6,6 +6,7 @@ package tracingpolicy
 import (
 	"bytes"
 	_ "embed"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -526,4 +527,71 @@ func TestYamlNamespaced(t *testing.T) {
 	require.NoError(t, err)
 	_, ok := tp.(TracingPolicyNamespaced)
 	require.True(t, ok)
+}
+
+func createTempFile(t *testing.T, data string) string {
+	file, err := os.CreateTemp(t.TempDir(), "tetragon-")
+	if err != nil {
+		t.Fatalf("cannot create temp. file: %v", err)
+	}
+
+	_, err = file.WriteString(data)
+	if err != nil {
+		t.Fatalf("cannot write to temp. file: %v", err)
+	}
+	err = file.Close()
+	if err != nil {
+		t.Fatalf("cannot close temp. file: %v", err)
+	}
+
+	return file.Name()
+}
+
+const failedLoadingTpFile = "failed loading tracing policy file"
+
+func TestEmptyTracingPolicy(t *testing.T) {
+	path := createTempFile(t, "")
+	_, err := FromFile(path)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), failedLoadingTpFile, "unexpected error")
+	assert.Contains(t, errors.Unwrap(err).Error(), "could not find validator for: /", "unexpected wrapped error")
+}
+
+func TestInvalidYAMLInTracingPolicy(t *testing.T) {
+	path := createTempFile(t, "<not-quite-yaml>")
+	_, err := FromFile(path)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), failedLoadingTpFile, "unexpected error")
+	assert.Contains(t, errors.Unwrap(err).Error(), "failed to unmarshall policy", "unexpected wrapped error")
+}
+
+const tpWithoutMetadata = `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata: {}
+`
+
+func TestTracingPolicyWithoutMetadata(t *testing.T) {
+	path := createTempFile(t, tpWithoutMetadata)
+	_, err := FromFile(path)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), failedLoadingTpFile, "unexpected error")
+	assert.Contains(t, errors.Unwrap(err).Error(), "validation failure list:\nmetadata.name", "unexpected wrapped error")
+}
+
+const tpNotCoveredBySpec = `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: not-covered-by-spec
+spec:
+  some_field: some_value
+`
+
+func TestTracingPolicyNotCoveredBySpec(t *testing.T) {
+	path := createTempFile(t, tpNotCoveredBySpec)
+	_, err := FromFile(path)
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), failedLoadingTpFile, "unexpected error")
+	assert.Contains(t, errors.Unwrap(err).Error(), "unknown field \"some_field\"", "unexpected wrapped error")
 }
--- a/pkg/tracingpolicy/generictracingpolicy.go
+++ b/pkg/tracingpolicy/generictracingpolicy.go
@@ -272,11 +272,11 @@ func FromYAML(data string) (TracingPolicy, error) {
 
 	validationResult, err := ValidateCRD(policy)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("validation failed %q: %w", policy.GetMetadata().Name, err)
 	}
 
 	if len(validationResult.Errors) > 0 {
-		return nil, fmt.Errorf("validation failed: %w", validationResult.AsError())
+		return nil, fmt.Errorf("validation failed: %q: %w", policy.GetMetadata().Name, validationResult.AsError())
 	}
 
 	return policy, nil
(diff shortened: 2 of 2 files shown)


pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -527,3 +529,82 @@ func TestYamlNamespaced(t *testing.T) {
 	_, ok := tp.(TracingPolicyNamespaced)
 	require.True(t, ok)
 }
+
+func createTempFile(data string) string {
+	file, err := ioutil.TempFile("", "*")
+	if err != nil {
+		logger.GetLogger().Panic("cannot create temporary file")
```
tixxdz, line 536: Here just return an error, then in the caller do t.Fatalf() in case

tixxdz (commented):
So lgtm, just a minor fix in the test so we fatal in the test getting the context.

Also for that .gitignore yes another PR is welcome ;-)

Thank you!

tixxdz:
Some CI code format are failing too. Thank you

mtardy:
> > BTW, it seems to me that drop-privileges should be added to contrib/tester-progs/.gitignore. I could create another PR for that, if applicable (plus an issue, if necessary for the trivial change).
>
> Also for that .gitignore yes another PR is welcome ;-)

Thanks a lot that's nice, an issue is not needed for a trivial change like that!

pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -527,3 +529,82 @@ func TestYamlNamespaced(t *testing.T) {
 	_, ok := tp.(TracingPolicyNamespaced)
 	require.True(t, ok)
 }
+
+func createTempFile(data string) string {
+	file, err := ioutil.TempFile("", "*")
```
mtardy, line 534: As the linter said, you can use https://pkg.go.dev/os#CreateTemp here, you can add a prefix or a suffix with `tetragon` in it btw it can be useful in case the temp file fails to cleanup, not necessary since it should be in `tmp` but still it's nice. Maybe you can find other example in the codebase, not sure!

pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -527,3 +529,82 @@ func TestYamlNamespaced(t *testing.T) {
...
+	}
+	path := file.Name()
+
+	_, err = file.WriteString(data)
+	if err != nil {
+		logger.GetLogger().Panicf("cannot write to temporary file %q", path)
+	}
+	err = file.Close()
+	if err != nil {
+		logger.GetLogger().Panicf("cannot close temporary file %q", path)
```
mtardy, line 546: maybe you can just `T.Fail` here or just warn since it's not a big deal if we fail to close that file, and should not happen that often.

mtardy (commented):
Welcome to contributing to Tetragon! 🤗 

That's very nice of you to take the time to add tests for invalid tracing policies. I have a few minor comments, hope it's helpful with Djalal's review as well.

pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -527,3 +528,78 @@ func TestYamlNamespaced(t *testing.T) {
...
+	if err != nil {
+		t.Error(err)
+	}
+
+	t.Cleanup(func() {
+		err := os.Remove(path)
+		if err != nil {
+			t.Error(err)
+		}
+	})
```
mtardy, line 553: since you are using `file, err := os.CreateTemp(t.TempDir(), "tetragon-")`, you can remove the cleanup part since the `TempDir` folder [is already automatically removed](https://pkg.go.dev/testing#B.TempDir):
> TempDir returns a temporary directory for the test to use. The directory is automatically removed by Cleanup when the test and all its subtests complete.

pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -527,3 +528,78 @@ func TestYamlNamespaced(t *testing.T) {
 	_, ok := tp.(TracingPolicyNamespaced)
 	require.True(t, ok)
 }
+
+func createTempFile(t *testing.T, data string) string {
+	file, err := os.CreateTemp(t.TempDir(), "tetragon-")
+	if err != nil {
+		t.Error(err)
```
mtardy, line 535: I think you should use `Fatal` here since you cannot recover from this and you should stop the test here.

pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -527,3 +528,78 @@ func TestYamlNamespaced(t *testing.T) {
...
+	}
+	path := file.Name()
+
+	_, err = file.WriteString(data)
+	if err != nil {
+		t.Error(err)
```
mtardy, line 541: Same I think you should use `Fatal` here.

pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -527,3 +528,78 @@ func TestYamlNamespaced(t *testing.T) {
...
+	}
+
+	t.Cleanup(func() {
+		err := os.Remove(path)
+		if err != nil {
+			t.Error(err)
+		}
+	})
+
+	return path
```
mtardy, line 555:
```suggestion
	return file.Name()
```

pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -527,3 +528,78 @@ func TestYamlNamespaced(t *testing.T) {
 	_, ok := tp.(TracingPolicyNamespaced)
 	require.True(t, ok)
 }
+
+func createTempFile(t *testing.T, data string) string {
+	file, err := os.CreateTemp(t.TempDir(), "tetragon-")
+	if err != nil {
+		t.Error(err)
+	}
+	path := file.Name()
```
mtardy, line 537:
```suggestion
```

christian-2:
@mtardy I've taken your suggestions from the second round of reviewing on board now. The only difference is that I opted for `testing.assert.NoError()` instead of `testing.Fatal()`; both should stop the test's execution.

mtardy:
> @mtardy I've taken your suggestions from the second round of reviewing on board now. The only difference is that I opted for `testing.assert.NoError()` instead of `testing.Fatal()`; both should stop the test's execution.

Yeah it's fine. 

One can argue that it should not be an `assert.NoErrro()` since it's not strictly speaking a test error, meaning the test found something that is wrong but that the test machinery failed. In other words the test did not uncover an issue but just failed in itself. But honestly, this is just a nit at this stage, let's merge this as-is, it's not a big deal!

christian-2:
@mtardy Yep. I'd propose I'll consider this if we shall need another round of changes. All your input on this issue has been very constructive and helpful so far, so I'd have absolutely no problem with that. Thanks again.

mtardy:
> @mtardy Yep. I'd propose I'll consider this if we shall need another round of changes. All your input on this issue has been very constructive and helpful so far, so I'd have absolutely no problem with that. Thanks again.

it's up to you, we can merge as-is if you want.

christian-2:
Then I'd rather revert to `testing.Fatal()`, it's the (slightly) better choice also IMO. Just a sec ... ;)

mtardy (approved):
Thanks again! 🎉

PR #2043, opened 2024-01-31
Revert "update module github.com/cilium/little-vm-helper to v0.0.15"

mtardy:
This reverts commit 0f807f0e9108ad7bde794891ed7fb29eba04a810.

LVH bumped their Go directive to 1.21.6 which bump our main directives. They reversed that in cilium/little-vm-helper@fc8f9a977a13baf1f4fdd0e0669743def536be8d. Let's wait for another release before upgrading that so that it doesn't break our current Go directive convention.

I think we might need to embrace the new Go versioning system eventually for the Go directive as this can happen with any dependency but this is another discussion.


kkourt (approved):
It's fine to revert. The new lvh version was meant for features that are needed in cilium, but nothing that is needed in tetragon.

api/go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/cilium/tetragon/api
 
 // renovate: datasource=golang-version depName=go
-go 1.21.6
+go 1.21.0
```
joestringer, line 4: Depending on the renovate configuration, it may end up trying to bump this again. In other projects we just configure renovate to only pick up updates for minor Go releases.

mtardy: we do that in Tetragon as well, here it was a deps updates that bumped it via `go mod tidy`

mtardy: https://github.com/cilium/tetragon/blob/5bf9880ea3672c2e95583ba91ce90d7d74fabdf9/.github/renovate.json5#L303-L326

PR #2046, opened 2024-02-01
Update github.com/opencontainers/runc to v1.1.12

tpapagian:
To fix https://github.com/cilium/tetragon/security/dependabot/44


mtardy:
Thanks! at the same time on the renovate side we have:
image

it should have the right to see those 🤔

mtardy (approved):

Issue #2030, opened 2024-01-27
Tutorial: Correlate additional metadata with events


christian-2:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

From a practical perspective it would be helpful if Tetragon events could bear human-friendly usernames in addition to `uid` and `auid` . However, it seems ill-advised to place the gathering of such usernames in Tetragon itself (including `tetra`). #2015 includes arguments to that effect. What could be helpful instead is a tutorial that sketches alternative approaches for achieving this practical goal.

### Describe the feature you would like

A new [tutorial](https://tetragon.io/docs/tutorials/) by the title *Correlate additional metadata with events*. It (its initial version) would lay two ways of adding human-friendly usernames to Tetragon's events: 

1. A simpler approach with more overhead consists of "piping" Tetragon's JSON events into a script that suitably combines calls to `jq` and `id` for achieving the correlation.
2. A second approach with less overhead that employs a custom binary program, which serves as gRPC client opposite Tetragon, links against `libc` (from where usernames can be "pulled"), and serves as JSON server. So the program acts as event proxy with the tasks of adding an additional field of metadata. Its usage could be `username [ --import-filename | --server-address ] [ --uid | --auid ]`.

The itle of the new tutorial is inspired by a sentence from the O'Reilly report [*Security Observability with eBPF*,](https://isovalent.com/books/ebpf-security/) which addresses roughly similar concerns:

> eBPF programs enable Kubernetes support by bundling API “watcher” programs that pull identity metadata from the Kubernetes API server and correlate that with container events in the kernel.

### Describe your proposed solution

We can implement this feature by:
* the tutorial in `docs/content/en/docs/tutorials/correlate-metadata.md`
* an implementation of `username` in `contrib/username` (perhaps in C)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

christian-2:
@mtardy what is the next step for enhancement proposals such as this one; for instance, is a further review required? I'd be willing to work on this as well (at as self-determined pace).

mtardy:
You can work on this and propose a PR I think this can be valuable.

As of this:
> implementation of the custom program in contrib/usernames/ (optional; perhaps in C)

I would suggest using existing utilities like `id` if you can.

christian-2:
[...]

PR #2023, opened 2024-01-25
operator: Add flag to forcefully update CRDs

aohoyd:
Currently tetragon-operator allows to update CRDs only if running version is smaller. But in some cases it is required to just replace CRDs.

This PR makes next changes:
- Adds `pkg/k8s/crdutil/CRDOption` struct and `pkg/k8s/crdutil/RegisterCRDsWithOptions` function to pass options into control flow
- Adds flag to the operator
- Adds flag to the helm chart

Fixes: #2014

```release-note
Allow for the operator to forcefully update CRDs regardless of the version
```

Issue #2015, opened 2024-01-24
Output username in event logs


ycaoT:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

As a security team, we want to know users' activities on a host. However in the current event logs, it only has the user ID, making it hard to trace down to the real user. 

### Describe the feature you would like

A straightforward way would be to map the user id with the `/etc/passwd`, and output the username. I've found falco has similar things too, would like to request similar thing. https://falco.org/docs/reference/rules/supported-fields/#field-class-user 

### Describe your proposed solution

Something like this from falco, https://github.com/falcosecurity/libs/blob/master/userspace/libsinsp/user.cpp#L60

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mtardy:
Hey, thanks for taking the time to open an issue for this. So there are some technical limitations to this, reading `/etc/passwd` is not enough quite often, when using NIS or LDAP for example:

The `os/user` Golang standard library package provides two implementations:
- one is relying on the libc and is using routines such as `getpwuid_r`, `getgrnam_r`, and `getgrouplist`. Since it needs the libc, it needs to be dynamically linked with CGO.
- the other one is a pure Go implementation that has limitations since it's only reading `/etc/passwd`.

As of now, Tetragon is compiled statically without CGO, making it possible to run in a distroless environment, we could not use the libc implementation without changing that for that specific feature.

On top of that, the Tetragon pod would need to have access to the host `/etc/passwd` file, so that's another host mount (a touchy one) to add for that feature. When running directly on the host, this problem does not exist.

So, unfortunately, that feature which might look simple at first sight could require some fundamental changes to Tetragon, especially when deploying on Kubernetes.

christian-2:
FWIW I'm doing the translation (from `uid`s to username) downstream from, i.e. outside, Tetragon: the Linux kernel does not know about usernames (only `uid`s), so I feel this is the right approach, for Tetragon (since it's based on eBPF) sits fairly close to (and partly inside) the kernel.

mtardy:
> FWIW I'm doing the translation (from uids to username) downstream from, i.e. outside, Tetragon: the Linux kernel does not know about usernames (only uids), so I feel this is the right approach, for Tetragon (since it's based on eBPF) sits fairly close to (and partly inside) the kernel.

[...]

PR #2050, opened 2024-02-01
Tutorial: Correlate additional metadata with events

christian-2:
Tutorial with initial section that addresses translating uids and auids into user names.

Fixes: 2030

### DIFF
--- a/docs/content/en/docs/tutorials/correlate-metadata.md
+++ b/docs/content/en/docs/tutorials/correlate-metadata.md
@@ -1,0 +1,78 @@
+---
+title: "Correlate additional metadata with events"
+weight: 1
+description: "Enrich Tetragon events from sources other than the kernel"
+---
+
+### Add user names to Tetragon events
+
+From a practical perspective, it can be helpful if Tetragon events ultimately 
+bear human-friendly user names in addition to `uid`s and `auid`s. Tetragon sits
+very close to (and partly inside) the kernel, which has no knowledge of user
+names over these ids.
+
+Hence the addition of user names is best accomplished in a pipeline where
+Tetragon comes first and a script dedicated to the task comes second, more
+peripheral to the kernel. If a script is not deemed performamt enough, the task
+could be carried instead by a binary that would act as gRPC client opposite
+Tetragon.
+
+The following example details the script-based approach; it assumes that
+Tetragon has been started with option `-export-filename
+/var/log/tetragon/tetragon.log`.
+
+```
+tail -f /var/log/tetragon/tetragon.log | sudo add-usernames.sh
+```
+
+`add-usernames.sh` is as follows:
+
+```
+#!/bin/bash
+set -e
+trap 'echo "error: $0:$LINENO"' ERR
+
+pattern_1='^\{"process_(exec|exit)":\{'
+pattern_2='\{"process":\{.*\}, "parent":\{.*"binary":"[^"]*/'$(basename $0)'".*'
+
+while read -r event; do
+	# skip Tetragon events other than process_exec, process_exit;
+	# for those, avoid "recursion" whereby invocation of an external
+	# binary (e.g. jq) would lead to further Tetragon events process_exec,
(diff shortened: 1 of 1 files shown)


christian-2:
@tardy @ ycaoT I have taken a first step at the new tutorial. Please review and test the script, if possible. @ycaot, does this address your needs from a user perspective?

ycaoT:
@christian-2 Thanks for the tutorial, this is definitely a good start. A bash script works, but I have some concerns for putting this into the real world, it may not be performant enough and may not able to deliver robust real time results. I am going to explore some middle layers like FluentBit or Vector.dev to further enrich the event outputs.

christian-2:
@yaoT There is also always the other option that I had mentioned in #2023: an e.g. Go implementation of a gRPC client for carrying out the script's task with high performance. Admittedly I am not planning to work on this myself right now. But perhaps there is a way to "upvote" such a feature such that I may receive more overall attention. Maybe @mtardy can advise. In any case, good luck for your efforts.

ycaoT:
@christian-2 Is that the right link? I don't see things related to the gRPC client.

christian-2:
@ycaoT My bad, there was a typo: it's #2030. I've mentioned gRPC there as well as in the description that accompanies the Bash script in the earlier commit.

mtardy (commented):
Hey, we've been discussing this offline and we've realized that this would make an interesting blog post (since we want to remove the "tutorials" section). 

Would you have a blog on which you would like to publish this? If not, we can publish on https://cilium.io/blog/. In any cases we can also guide you :) tell me what you think about it.

christian-2:
@mtardy Yes, I for one would be interested in participating in a blog post suitable for https://cilium.io/blog . It's an honor. I don't maintain a blog otherwise.

mtardy:
Could you help us with the next steps @xmulligan? I think it would mean opening a PR to add a new entry under https://github.com/cilium/cilium.io/tree/main/src/posts.

xmulligan:
Yes, the correct way would be to create a PR like this one https://github.com/cilium/cilium.io/pull/448

ycaoT:
As an industry experience, especially for security and observability perspectives; I mentioned this in my original request too https://github.com/cilium/tetragon/issues/2015. It will be very helpful if tetragon agent can provide a way to output username, the `userId` does not provide much help in the real world. Falco already has this (if linking with `/etc`, they read from `/etc/passwd`, outputs like https://falco.org/docs/reference/rules/supported-fields/#field-class-user), I am sure tetragon can implement this too

ycaoT:
thanks @christian-2 .  Totally agreed that 
`UPDATE On reflection, it's perhaps too strong an assumption (even for Linux alone) that /etc/passwd always "owns" usernames. Clearly e.g. PAM allows for different configurations.` 
But at least it's a good start, e.g: having a flag or mount to that file when the agent run, then it can use that to enrich the event outputs. 

Again, I want to emphasize the intention of why, the `uid` is not going to help anything from security perspective, later turns to the compliance impact. We have to trace down events to the specific user (username) per both security and compliance requirements. Similar products, like Sysdig, Falco, Crowdstrike's agents are both above to deliver username from the outputs

mtardy:
> I.e. let Tetragon parse `/etc/passwd` (or whatever the Windows equivalent may be) when it starts up and let it put the association between uids and usernames into a dedicated eBPF map. Let it also monitor changes to `/etc/passwd` and update that map accordingly. Then let tracing policies use either uids or usernames as values for arguments that bear (a new) type `uid`.
> 
> **UPDATE** On reflection, it's perhaps too strong an assumption (even for Linux alone) that `/etc/passwd` always "owns" usernames. Clearly e.g. PAM allows for different configurations.

So indeed, `/etc/passwd` is most of the time not enough, and I want to emphasize that this is mostly a userspace construction which limits a bit the comparison between file descriptors.

This is tricky, in the case of Tetragon we can't even use the same things used as in other tools (using the libc) because we want to build statically.

But this thing as been requested by multiple people and we see the interest.

Issue #530, opened 2022-11-04
deduplicate generic type arguments 


kkourt:
There are two versions of generic type arguments:

https://github.com/cilium/tetragon/blob/3076ae7b9cd2bbd7320af3708af240caa0fc684a/pkg/generictypes/generictypes.go#L6-L36

And

https://github.com/cilium/tetragon/blob/3076ae7b9cd2bbd7320af3708af240caa0fc684a/pkg/selectors/kernel.go#L140-L159

There seem to already be conflicts, so we should deduplicate them.

kkourt:
Hi @rajaSahil,

Go ahead if you want. I'm not aware of someone else working on it at the moment.
Thanks!

SumaiyaSafdar:
hey @kkourt can i work on this issue

kkourt:
@SumaiyaSafdar please sync with @rajaSahil.

mtardy:
For info there is potential progress here https://github.com/cilium/tetragon/pull/696 but no response from the author yet.

mtardy:
There was a first attempt here https://github.com/cilium/tetragon/pull/696 that was abandoned, I'll close the PR for now but this can be of inspiration for someone that wants to complete that :)

Trung-DV:
This issue already solve by PR https://github.com/cilium/tetragon/pull/2168

mtardy:
Indeed thanks!

PR #2060, opened 2024-02-02
pkg/selectors: deduplicate generic type arguments

h3x-eilidh:
There are two versions of the same generic type arguments currently in use.

This patch unifies them by bringing all arguments into the pkg/generictypes/generictypes.go class.

Fixes: #530


mtardy:
Thanks for taking the time to contribute to Tetragon! Could you take a look at the CI result and rebase your PR?

You mostly need to remove the merge commit and look at the golangci-lint result, you can use `make check` to run locally.

mtardy (commented):
Thanks for taking the time to contribute to Tetragon! Could you take a look at the CI result and rebase your PR?

You mostly need to remove the merge commit and look at the golangci-lint result, you can use make check to run locally.

PR #2062, opened 2024-02-05
docs: fix assertions about tracingpolicy args field 

h3x-eilidh:
The docs about the _tracing policies_ `args` field make wrong assertions about its use.
The `args` field is not mandatory and it is currently not supported on `uprobes`.

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -214,11 +214,9 @@ running in all the bash shells.
 
 ## Arguments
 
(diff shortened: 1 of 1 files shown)


mtardy (commented):
Hey! Thanks a lot for reading and fixing the docs, you are correct for v1.0 it was indeed a mistake but this one was fixed very recently, so it's now technically correct if you want to try the latest build, see https://github.com/cilium/tetragon/pull/1978.

h3x-eilidh:
Nice! I didn't know about it, thank you @mtardy. 
I'll close the PR

PR #2064, opened 2024-02-05
bpf: document how to use read_exe()

tixxdz:


mtardy (approved):
Thanks @tixxdz for the details, that's valuable to not misuse that one in the future.

PR #2069, opened 2024-02-06
String: Support longer exact match strings

kevsecurity:
String matching is limited to strings of length 144 due to the way the hash look up tables were implemented (6 key sizes, 24 byte increments).

This commit adds maps to support larger key/string sizes: 256, 512, 1024, 2048, and 4096. This will therefore permit exact string matches for strings of length up to 4096 (not including null terminator).

Includes tests at map limits.

```release-note
Extend string matching to strings of length 4096 characters (increased from 144 characters).
```


jrfastab (approved):

dwindsor:
> String matching is limited to strings of length 144 due to the way the hash look up tables were implemented (6 key sizes, 24 byte increments).
> 
> This commit adds maps to support larger key/string sizes: 256, 512, 1024, 2048, and 4096. This will therefore permit exact string matches for strings of length up to 4096 (not including null terminator).
> 
> Includes tests at map limits.
> 
> ```
> Extend string matching to strings of length 4096 characters (increased from 144 characters).
> ```

Hi! Thanks for this fix.

Would it be possible to not store paths in the strings table, rather in another table that's keyed by a `(device, inode)` tuple in order to remove the limit on older kernels?

PR #2075, opened 2024-02-08
tetragon: Harden TestMatchCloneThreadsIDs test

olsajiri:
It's possible we receive the origincal clone event out of order after the exec event, which will store wrong value to clonePID/cloneTID variables and mess up the test logic.

Making sure we skip the first clone event.

### DIFF
--- a/pkg/sensors/exec/threads_test.go
+++ b/pkg/sensors/exec/threads_test.go
@@ -123,6 +123,8 @@ func TestMatchCloneThreadsIDs(t *testing.T) {
 		}
 	}
 
+	cloneEvents := []*grpcexec.MsgCloneEventUnix{}
+
 	execPID, execTID := uint32(0), uint32(0)
 	clonePID, cloneTID := uint32(0), uint32(0)
 	events := perfring.RunTestEvents(t, ctx, ops)
@@ -134,12 +136,8 @@ func TestMatchCloneThreadsIDs(t *testing.T) {
 				execTID = ev.Unix.Process.TID
 			}
 		case *grpcexec.MsgCloneEventUnix:
-			// We found the exec parent then catch the single per thread
-			// clone event
-			if execPID != 0 && clonePID == 0 {
-				clonePID = ev.PID
-				cloneTID = ev.TID
-			}
+			// Store all received clone events so we parse them later
+			cloneEvents = append(cloneEvents, ev)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/sensors/exec/threads_test.go
```diff
@@ -135,8 +135,9 @@ func TestMatchCloneThreadsIDs(t *testing.T) {
 			}
 		case *grpcexec.MsgCloneEventUnix:
 			// We found the exec parent then catch the single per thread
-			// clone event
-			if execPID != 0 && clonePID == 0 {
+			// clone event. Make sure it's not the first clone event that
+			// got out of order and arrived after exec event.
+			if execPID != 0 && execPID != ev.PID && clonePID == 0 {
```
tixxdz, line 140: This could fix it, thanks for checking on it ;-) !

Was thinking maybe: store the events, then after parsing thread-tester output, filling up tti and `tti.AssertPidsTids(t)` call , we go again over the events with something like:

```
diff --git a/pkg/sensors/exec/threads_test.go b/pkg/sensors/exec/threads_test.go
index b34c40293..54aa747f2 100644
--- a/pkg/sensors/exec/threads_test.go
+++ b/pkg/sensors/exec/threads_test.go
@@ -123,6 +123,8 @@ func TestMatchCloneThreadsIDs(t *testing.T) {
 		}
 	}
 
+	cloneEvents := make([]*grpcexec.MsgCloneEventUnix, 0)
+
 	execPID, execTID := uint32(0), uint32(0)
 	clonePID, cloneTID := uint32(0), uint32(0)
 	events := perfring.RunTestEvents(t, ctx, ops)
@@ -134,12 +136,8 @@ func TestMatchCloneThreadsIDs(t *testing.T) {
 				execTID = ev.Unix.Process.TID
 			}
 		case *grpcexec.MsgCloneEventUnix:
-			// We found the exec parent then catch the single per thread
-			// clone event
-			if execPID != 0 && clonePID == 0 {
-				clonePID = ev.PID
-				cloneTID = ev.TID
-			}
+			// Store all received clone events so we parse them later
+			cloneEvents = append(cloneEvents, ev)
 		}
 	}
 
@@ -148,14 +146,23 @@ func TestMatchCloneThreadsIDs(t *testing.T) {
 	require.NotZero(t, execPID)
 	require.Equal(t, execPID, execTID)
 
-	// ensure exec and all follow clone with all threads match back to parent exec
+	// ensure exec events match
 	require.Equal(t, execPID, tti.ParentPid)
 	require.Equal(t, execPID, tti.ParentChild1Pid)
 	require.Equal(t, execPID, tti.ParentThread1Pid)
 
-	// ensure clone event matches
+	for _, ev := range cloneEvents {
+		// Get the clone event that orginates from the exec event
+		if execPID == ev.Parent.Pid {
+			clonePID = ev.PID
+			cloneTID = ev.TID
+		}
+	}
+
+	// ensure clone event match
+	require.NotZero(t, clonePID)
 	require.Equal(t, clonePID, cloneTID)
-	require.Equal(t, clonePID, tti.Child1Pid) // First clone
+	require.Equal(t, clonePID, tti.Child1Pid)
 
 	// ensure that threads match on the thread group leader
 	require.Equal(t, tti.Child1Pid, tti.Thread1Pid)  
```

This way we discard all unrelated clones, also the clone event should fail if it does not find the parent IIRC so we should be covered, up to you ;-)

tixxdz (approved):

pkg/sensors/exec/threads_test.go
olsajiri: yea, makes sense, we could race with other clones

olsajiri:
@tixxdz I changed that the way you proposed, plz check

PR #2084, opened 2024-02-11
Update enforcement.md

next-ramses:
Docker example with enforcement was fetching the yaml policy without enforcement, that's fixed with this cange.


mtardy:
Thanks for taking the time to fix this, @jrfastab could you take a look?

mtardy (approved):
Thanks for taking the time to fix this :)

PR #2085, opened 2024-02-11
tetragon: Factor uprobe sensor setup

olsajiri:
after talking to ebpf/cilium maintainers, new release might still take a while,
so pushing at least the generic refactoring change from https://github.com/cilium/tetragon/pull/1914 so it won't get old 

Separating  uprobe sensor setup into addUprobe function, that creates and sets up uprobeEntry object and createSingleUprobeSensor that creates the actual sensor progs/maps arrays.

This way we can easily add multi uprobe sensor create function in following changes.


kevsecurity (approved):
LGTM

PR #2089, opened 2024-02-12
Parse kthreads during /proc scanning

tpapagian:


pkg/grpc/exec/kthread_init.go
```diff
@@ -0,0 +1,46 @@
...
+	"log"
+
+	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/cilium/tetragon/pkg/api/processapi"
+	"github.com/cilium/tetragon/pkg/process"
+	"github.com/cilium/tetragon/pkg/reader/notify"
+)
+
+// This is used only when we add kernel threads during /proc scanning.
+// We use the same interface with all the other events to make the handling
```
tixxdz, line 16: But we also add kernel threads from the clone right? your PR does that?  I'm just thinking on how to improve that comment so it reads we handle /proc scanning here and clone part of the usuall clone.

contrib/tester-progs/.gitignore
```diff
@@ -21,3 +21,4 @@ killer-tester-32
 /getcpu
 drop-privileges
 change-capabilities
+direct-write-tester
```
tixxdz, line 24: seems we are missing the direct-write-tester

tixxdz: ah already committed :-D

pkg/grpc/exec/kthread_init.go
tpapagian: > But we also add kernel threads from the clone right? your PR does that?

Yes, the next commit reverts a commit that didn't do that. 

This is only used for kernel threads during proc scanning. This should not be used during normal operation. The reason that I added that is that this is somewhere in the middle compared to what the exec and clone handler do. 

In that case, we don't need to generate events compared to exec. In comparison to clones, we do not have a way to provide a process name as we always inherit that from the parent.  

But this is also much more simple. There is no need to do anything related to the eventcache. I preferred to create a new (simple) handler compared to changing the existing handlers (and making them more complex).

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -186,58 +190,95 @@ func pushExecveEvents(p procs) {
...
-		Uid: p.uids[0], Euid: p.uids[1], Suid: p.uids[2], FSuid: p.uids[3],
-		Gid: p.gids[0], Egid: p.gids[1], Sgid: p.gids[2], FSgid: p.gids[3],
-	}
-	m.Unix.Process.Flags = p.flags | flags
-	m.Unix.Process.Ktime = p.ktime
-	m.Unix.Msg.Common.Ktime = p.ktime
-	m.Unix.Process.Filename = filename
-	m.Unix.Process.Args = args
+	if p.kernel_thread {
+		m := exec.MsgKThreadInitUnix{}
```
olsajiri, line 201: why can't we use MsgExecveEventUnix for kernel threads as well?

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -500,13 +578,13 @@ func listRunningProcs(procPath string) ([]procs, error) {
 
 		execPath, err := os.Readlink(filepath.Join(procPath, d.Name(), "exe"))
 		if err != nil {
-			logger.GetLogger().WithError(err).WithField("process", d.Name()).Warnf("reading process exe error")
+			execPath = strings.TrimSuffix(string(cmdline), "\n")
```
olsajiri, line 581: hm, how this became non error case?

tpapagian: In the case of kernel threads, `/proc/<pid>/exe` does not make sense. So in that case, I get the `execPath` from the `kthread` name. But still, for the non-kernel threads, we should report this error. I will update it to something like:
```
if kernelThread {
    execPath = strings.TrimSuffix(string(cmdline), "\n")
} else {
    logger.GetLogger().WithError(err).WithField("process", d.Name()).Warnf("reading process exe error")
}
```

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -408,6 +463,20 @@ func listRunningProcs(procPath string) ([]procs, error) {
 			if err != nil {
 				logger.GetLogger().WithError(err).Warnf("Reading Loginuid of %s failed, falling back to loginuid: %d", pathName, uint32(auid))
 			}
+
+			// Check and add only thread leader processes.
+			if len(status.NSpid) > 0 && len(status.NStgid) > 0 {
```
olsajiri, line 468: this is unrelated change, right? I think it should be in separated commit or at least mentioned in the changelog

olsajiri (approved):

pkg/sensors/exec/procevents/proc_reader.go
tpapagian: Yes, I will do that in a separate commit.

tixxdz: I didn't get this change! IIRC scanning /proc/ only shows thread leaders, it should not show threads or maybe I'm getting old and making stuff...

tixxdz (approved):
Some stuff to fix, about the fake pid 0 that we make in case we have things triggered from interrupt context... hmm maybe just make it a kthread too? up to you.

pkg/sensors/exec/procevents/proc_reader.go
tpapagian: I believe that you are correct. I have checked the logs from running this several days with many restarts and no similar log messages were found. 

I initially did that as a sanity check but it seems that we can remove that. Thanks!

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -352,16 +392,30 @@ func listRunningProcs(procPath string) ([]procs, error) {
...
 
 		cmdline, err := os.ReadFile(filepath.Join(pathName, "cmdline"))
 		if err != nil {
 			continue
 		}
-		if string(cmdline) == "" {
+
+		// We read comm in the case where cmdling is empty (i.e. kernel thread).
+		comm, err := os.ReadFile(filepath.Join(pathName, "comm"))
+		if err != nil {
```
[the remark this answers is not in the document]
tpapagian: We check that this is a kernel thread by using `string(cmdline) == ""` in a similar way that we did that before. I use `comm` just to get a name for the kernel thread in order to provide a reasonable name for the binary field.

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -186,58 +189,95 @@ func pushExecveEvents(p procs) {
...
+
+		m.Unix.Msg.Capabilities = processapi.MsgCapabilities{}
+		m.Unix.Msg.Namespaces = processapi.MsgNamespaces{}
+
+		m.Unix.Process.Size = p.size
+		m.Unix.Process.PID = p.pid
+		m.Unix.Process.TID = p.pid
+		m.Unix.Process.NSPID = p.nspid
+		m.Unix.Process.UID = 0
+		m.Unix.Process.AUID = 0
```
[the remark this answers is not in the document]
tpapagian: Done!

tpapagian:
> Some stuff to fix, about the fake pid 0 that we make in case we have things triggered from interrupt context... hmm maybe just make it a kthread too? up to you.

I believe that we do that [here](https://github.com/cilium/tetragon/blob/18ecc577b2209a8605ad1c6acd2d86cbf5752e4c/pkg/sensors/exec/procevents/proc_reader.go#L370). This PR does not touch this part. We handle this as an exec message as we also print that in the json logs.

PR #2090, opened 2024-02-12
pkg/sensors: expose policy status through metrics and tetra

mtardy:
This changes the behavior of the agent relatively to "impossible to load" tracing polices.

First of all, we apply a first validation on the TracingPolicy custom resource, using YAML kubebuilder validation fields and defaults, those are applied whether Tetragon runs along the Kubernetes API or without, in standalone. In the case it runs along the Kubernetes API, it never sees the malformed custom resource as it is parsed by the Kubernetes API. In the case it runs without the Kubernetes API, we import the code that does the parsing and do the same at an early stage in our gRPC API.

The issue is, that the validation we can perform on the YAML is limited and there are more situations in which a syntactically valid TracingPolicy cannot load on any Tetragon agents or just some nodes (in case the underlying kernel is different and Tetragon cannot provide the same functionalities). This step is performed after we load the new policy and is deeply tied to the loading of the BPF collections.

Before this PR, the only way to realize an applied (as in `kubectl apply`) TracingPolicy failed to load was to watch for all the Tetragon agent logs and search for error messages. Now a user could aggregate the metrics and check for new gauges:
```shell
# HELP tetragon_tracingpolicy_loaded The number of loaded tracing policy by status.
# TYPE tetragon_tracingpolicy_loaded gauge
tetragon_tracingpolicy_loaded{status="disabled"} 0
tetragon_tracingpolicy_loaded{status="enabled"} 1
tetragon_tracingpolicy_loaded{status="error"} 0
tetragon_tracingpolicy_loaded{status="load_error"} 1
```

Or use `tetra tracingpolicy list` (or short `tetra tp list`), with text output revamped from this PR:
```text
ID   NAME           STATE        FILTERID   NAMESPACE   SENSORS
1    invalid        load_error   0          (global)
2    block-binary   enabled      0          (global)    gkp-sensor-1
```

And maybe use `tetra tp list -o json` for full error message:
```json
{"policies":[{"id":"1","name":"invalid","load_error":"policy handler 'tracing' failed loading policy 'invalid': tracing policies with both kprobes and tracepoints are not currently supported"},{"id":"2","name":"block-binary","sensors":["gkp-sensor-1"],"enabled":true}]}
```

So using `tetra` on concerned hosts, or checking the aggregated metrics and seeking for the error message in the logs should help to troubleshoot a bad TracingPolicy. The next step would be to standardize the error messages in the logs to make them easy to search.


pkg/metrics/policystatusmetrics/policystatusmetrics.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (c *policyStatusCollector) convertStatusToMetrics(policiesStatus []*tetragon.TracingPolicyStatus) []prometheus.Metric {
+	loadErrorCount, errorCount, disabledCount, enabledCount := 0, 0, 0, 0
+
+	// This for loop assumes that a tracing policy can only be in one of those
+	// states exclusively: load error, error, disabled or enabled.
+	//
+	// Meaning that a tracing policy having a load error cannot also have an
+	// error, and that anything with a load error or just an error cannot be
+	// enabled or disabled. See unit test.
```
lambdanis, line 68: Is there any sort of state machine for tracing policies? This comment suggests that there is an implicit but not an explicit one, but I'm not familiar with the policy manager code.

It took me a moment to understand what this comment means: on the struct level a policy is always either enabled or disabled, and can have both errors. But here we count each policy in exactly one metric - meaning, the order of the `if`s matters a lot, i.e. determines the state of the policy. Does it sound right? Are there any combinations of these fields that would mean a bug in the policy manager?

pkg/metrics/policystatusmetrics/policystatusmetrics.go
```diff
@@ -0,0 +1,91 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package policystatusmetrics
```
lambdanis, line 4: Is there a benefit in having the metrics code in a separate package rather than `pkg/sensors`? This is how other metrics are defined, but I don't love this pattern as it's not immediately obvious what code is instrumented.

lambdanis (commented):
Looks good to me overall, thanks :bowing_woman: I left some questions to clarify the details, not blocking.

pkg/metrics/policystatusmetrics/policystatusmetrics.go
mtardy: I cannot really answer for that one, I was not familiar with this and I just followed the pattern.

pkg/metrics/policystatusmetrics/policystatusmetrics.go
mtardy: I would say no, there is no state machine. This is expressed in the unit test, I'm not sure there is something to prevent that a policy is enabled and has an error had the same time, I didn't check. For load_error I know that this is an exclusive state, and enabled/disabled is by nature exclusive.

So I would say most of this should be correct, and I think it's better for the user if we assume a policy is in only one state at a time and give the error state over instead of a confusing "enabled/error" if that's even possible.

pkg/metrics/policystatusmetrics/policystatusmetrics.go
lambdanis: The main issue would be import cycles if the code is moved under `pkg/sensors` - but I don't think it's the case here?

pkg/metrics/policystatusmetrics/policystatusmetrics.go
lambdanis: Ok, thanks. Could you mention in the comment that the state is determined here by the order of `if`s? I.e. on the struct level a policy has between 1 and 3 "states" (it's always enabled or disabled and you can have 0-2 errors), but here it gets mapped into 1 state.

pkg/sensors/handler.go
```diff
@@ -89,11 +89,24 @@ func (h *handler) updatePolicyFilter(tp tracingpolicy.TracingPolicy, tpID uint64
...
 
+	col := &collection{
+		name:            op.name,
+		tracingpolicy:   op.tp,
+		tracingpolicyID: uint64(tpID),
+	}
+	// Register the collection early even if there is a potential loading error.
+	// col.loadError should be set before returning the error so that error
+	// state is saved and the policy can be later overridden
+	h.collections[op.name] = col
```
kkourt, line 108: There is an interim state here where the new collection is part of .collections but it's state is wrong (appears as if it's properly working). This is fine, because we serialize everything in a single goroutine so the state is not visible.

If we care about it, we could do:
```
defer func() {
    h.collections[op.name] = col
}()
```
An alternative would be to have a state in the collection: {LOADING, RUNNING, ERROR, ...}.

In any case, no need to change anything just sharing some thoughts as I'm reading the code in case we ever want to do something more complicated to allow for concurrent operations since we 've discussed this offline.

pkg/metrics/policystatusmetrics/policystatusmetrics.go
kkourt: > > Is there any sort of state machine for tracing policies? This comment suggests that there is an implicit but not an explicit one, but I'm not familiar with the policy manager code.

> I would say no, there is no state machine.

I'd say there is an implicit state machine for policies. It is embedded in the "collection" structure. For example, if a collection's state is LOADED_ERROR it means that it holds no resources so we can just replace it in the collections map.

kkourt (approved):
Changes look good to me.

Reading the PR and the comments, I do wonder if it would make sense to have a `.state` in the collection.

so if `.state == LOADING && error != nil` then error is a loading error.

I don't feel strongly about it though, and it's also something that we can address as a followup.

pkg/sensors/handler.go
```diff
@@ -15,7 +15,7 @@ import (
 
 type handler struct {
 	// map of sensor collections: name -> collection
-	collections    map[string]collection
+	collections    map[string]*collection
```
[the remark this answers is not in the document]
mtardy: So this particular thing is unfortunately not allowed in Golang https://stackoverflow.com/a/32496031/4561420, I'm not very satisfied as well with this change :/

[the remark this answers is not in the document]
mtardy: I finally dropped that commit and just kept the map of value and explicitly wrote the map value at each error statement. I think it's easier to read like this even if this is longer.

pkg/sensors/handler.go
mtardy: yep indeed, let's keep writing value instead of pointers at appropriate timing https://github.com/cilium/tetragon/pull/2090#discussion_r1492559364. Let's keep this simple.

pkg/metrics/policystatusmetrics/policystatusmetrics.go
mtardy: I'm not sure I didn't have those issues writing this but cannot really say sorry :(

mtardy:
> Quick review from the UI stuff first:
> 
> ```
> ID   NAME           ENABLED   FILTERID   NAMESPACE   SENSORS        LOADERROR   ERROR
> 1    invalid        false     0          (global)                   true        false
> 2    block-binary   true      0          (global)    gkp-sensor-1   false       false
> ```
> 
> The ID is ok maybe! why do you have to print it? is this useful or used somewhere?

I just kept the fields that were printed before, but indeed for now it's only used internally but I think it could be useful, it would be nice to keep a `map[int]string` to let the user use the name or the id indifferently. Most of the time it's just easier to type `1` instead the name and that would be a nice QoL feature for tetra. Not a huge deal, but that could be a good-first-issue.

Similarly, I'm not sure displaying the filterID is extremely useful, but it was there so not sure I should remove it. We can discuss that but this can be part of another PR I think as I just wanted to make the human output readable and add the status thingie (see next answer paragraph).

We should also document those fields in tetra, maybe replicating the doc from the gRPC.

> Also the LOADERROR and ERROR for users are the same, what's the difference? then also they are boolean, are we sure we may not have many errors of the same tracing policy 'block-binary'? after load unload during the uptime of Tetragon? I would just not use boolean here, and have it as a string at least in the doc, this way we can extend later! I would remove LOADERROR and ERROR and put something like ACTIVE.
> 
> So that ACTIVE then can be: `active, failed or what ever state in the future.` Then if you want more details about the tracing policy in question you have to do `tetra tp $command $tracingpolicy_name` in question.

I think I agree and I will just replicate what we have in the metrics with the `STATUS={enabled,disabled,load_error,error}`. This presentation is not great you are right!

> tetra must present a good summary that tells users everything and should not follow gRPC that is developer oriented.
> 
> Also keep in mind we will certainly have more fields in the future, like was this autoloaded or was it insert during runtime? etc, so would try to keep some space as mush as possible and just put what we think is really useful ;-)

I would say for this it's not a huge deal if we later break `-o text` by removing/adding fields as people should use `-o json` if they want something consistent for scripting but I understand the argument.

pkg/metrics/policystatusmetrics/policystatusmetrics.go
mtardy: Regarding this comment and Djalal output on `tetra` output, I put that state part in a separate function (that we could move somewhere else) that I can reuse (in tetra), and changed everything from `status` to `state` to avoid confusion (and I think this discussion highlighted that state makes more sense).

pkg/metrics/policystatemetrics/policystatemetrics.go
```diff
@@ -0,0 +1,107 @@
...
+		return LoadErrorState
+	}
+	if policyStatus.Error != "" {
+		return ErrorState
+	}
+	if policyStatus.Enabled {
+		return EnabledState
+	}
+	return DisabledState
+}
```
mtardy: See @tixxdz @kkourt @lambdanis I added this so that this logic can be technically moved at some point outside the metrics package and maybe be part of the TracingPolicy logic. I could also reuse this in `tetra` and simplify the text output by printing the same `state` that we have in the metrics. 

It actually simplified the code and made it easier to test so I think this is what I was searching for initially.

mtardy:
All right I completely rewrote this PR and lost myself in the rebases. I'll add some unit tests tomorrow, fix the linter issue and potential test break, and ship this. I now integrated the `state` field in the TracingPolicyStatus to have an explicit state machine. I noticed that in theory, the error state is impossible (only load error).

mtardy:
would love to get an ack before I rebase to avoid multiple rebases :)

kkourt (approved):
🚀

api/v1/tetragon/sensors.proto
```diff
@@ -29,6 +29,19 @@ message ListSensorsResponse {
 
 message ListTracingPoliciesRequest { }
 
+enum TracingPolicyState {
+	// invalid state
+	INVALID = 0;
```
lambdanis, line 34: I usually see `UNKNOWN` instead of `INVALID` being used as 0 in similar enums, but invalid seems ok to me too.

lambdanis (approved):
looks good to me, thanks!

api/v1/tetragon/sensors.proto
mtardy: do you think I should change this?

tixxdz (approved):
Ack'ed as long as you think the status is properly covered ;-)

api/v1/tetragon/sensors.proto
lambdanis: yeah, I would say so, I think it's only a change in 1-2 commits?

lambdanis (approved):

PR #2091, opened 2024-02-13
chore: Sync bpf_tracing.h

tpapagian:
With the latest from
https://github.com/torvalds/linux/blob/master/tools/lib/bpf/bpf_tracing.h.

This mainly allows us to use BPF_PROG and BPF_KPROBE macros with more than 5 arguments.

### DIFF
--- a/bpf/libbpf/bpf_tracing.h
+++ b/bpf/libbpf/bpf_tracing.h
@@ -41,6 +41,10 @@
 
 #if defined(bpf_target_x86)
 
+/*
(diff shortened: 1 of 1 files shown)


tpapagian:
Based on https://github.com/cilium/tetragon/blob/2ec3cfb3cba309759ca4be03a84b5a1d766d8351/Makefile#L8 it seems that we do not link `bpf_tracing.h`. So I would expect to be fine to ignore the static check errors.

tpapagian:
> lgtm, did you cherry pick the changes or just copied the file over?

No, I just copied parts of the file. It seems that we have made some changes there so a cherry-pick may not work as expected.

Issue #2086, opened 2024-02-12
docs: remove the tutorial section and move its content


mtardy:
We discussed that the "tutorials" section does not make sense at the moment and we would prefer move its content under other sections for now.

Issue #1974, opened 2024-01-13
doc: improve troubleshooting Tetragon


tixxdz:
Improve troubleshooting Tetragon

- [x] Unify https://tetragon.io/docs/tutorials/debugging-tetragon/ into https://tetragon.io/docs/troubleshooting/
- [x] Improve doc around tetra bugtool, demo how to run it for all different installation k8s, container or systemd host installation.
- [ ] Add example how to activate gops server for all installation k8s if not by default, but also container or systemd
- [ ] Add example how to activate metrics for all installation if it is not already there.

janvi01:
Hey @tixxdz I would like to work on this issue if there is no update from the assignee.

UtkarshUmre:
Hey @janvi01  I'm actively working on this matter. I initially began the task, but it got sidelined due to exams and other obligations. Nonetheless, I'm back on track and giving it my full attention. If you're genuinely interested in contributing to it, we can collaborate on this

janvi01:
@UtkarshUmre You can open your PR, then I can add my inputs

PR #2097, opened 2024-02-14
docs: remove tutorials section and move its content under appropriate sections

mtardy:
Fixes https://github.com/cilium/tetragon/issues/2086.
Also first tickbox of https://github.com/cilium/tetragon/issues/1974.

Compatibility is maintained thanks to page aliases: i.e. removed page will redirect to the appropriate content and not 404.


tixxdz:
Hi Mahé thanks for the change, but the troubleshooting is more for production ;-) !

mtardy:
> Hi Mahé thanks for the change, but the troubleshooting is more for production ;-) !

yeah we can keep bade7ad2daf388b152c48d97c362fc13bcb8cdd8 out of this PR. It was discussed during hive time that it could make sense to put it down there but I don't mind.

kkourt:
Agreed, let's keep troubleshooting as is.

mtardy:
Okay I removed the commit!

Issue #527, opened 2022-11-03
cleanup: merge Config.BpfDir and Config.MapDir


kkourt:
Configuration defines a BpfDir and a MapDir:
https://github.com/cilium/tetragon/blob/411aa54de1697a0ee84f67453a380543e89c801a/pkg/option/config.go#L34-L35

This is the base directory for the pinned bpf programs and maps.
In most cases, however, they are the same.
E.g., https://github.com/cilium/tetragon/blob/411aa54de1697a0ee84f67453a380543e89c801a/cmd/tetragon/main.go#L179-L180

We can simplify the code and merge them into one.

Lan-ce-lot:
Hi, I want to try to solve it.

jrfastab:
@Lan-ce-lot feel free to tackle this if you want. I think it is a nice improvement.

PR #2106, opened 2024-02-15
tetragon: Remove superfluous MapDir

olsajiri:
Removing MapDir as preparation for loader refactoring,
it's not used, BpfDir is enough.

Fixes: https://github.com/cilium/tetragon/issues/527


kevsecurity (approved):
LGTM

PR #2117, opened 2024-02-16
tetragon: Rename killer to enforcer

olsajiri:
rename all the killer stuff to enforcer, plus one simple sensors fix


pkg/sensors/handler.go
```diff
@@ -251,15 +250,11 @@ func (h *handler) addSensor(op *sensorAdd) error {
...
 }
 
-func removeAllSensors(h *handler) error {
-	var errs error
+func removeAllSensors(h *handler) {
 	for _, col := range h.collections {
-		if err := col.unload(); err != nil {
-			errs = errors.Join(errs, err)
-		}
+		col.destroy()
```
kkourt, line 255: CC: @mtardy

kkourt (approved):
Thanks!

pkg/sensors/handler.go
mtardy: yep indeed this is a miss on my part thanks. The destroy hook was added in contrast to the unload one that was used when `enabling/disabling` vs removing completely the sensor. Allowing us to clean up stuff only when it's really removed and not only disabled.

kkourt: Sounds good. Can you please ACK the PR if you agree with the change?

mtardy: Too late but yes I agreed with the change

lambdanis:
This PR contains a breaking change to a CRD, so I labeled it as such. @olsajiri could you write in the release note which fields changed?

PR #2119, opened 2024-02-16
exporter: fix hang when exporter fails to start

willfindlay:
The Exporter.Start() implementation uses a WaitGroup to wait until the exporter is ready after launching it in a goroutine. Unfortunately, this WaitGroup never gets set to Done when we encounter an error while starting the exporter. This ultimately causes the Tetragon process to hang indefinitely. Fix the issue by simply calling readyWG.Done() when we encounter an error here.

```release-note
Fix a hang when the event exporter fails to start.
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -134,14 +134,23 @@ func (s *Server) GetEventsWG(request *tetragon.GetEventsRequest, server tetragon
 	}).Debug("Received a GetEvents request")
 	allowList, err := filters.BuildFilterList(s.ctx, request.AllowList, filters.Filters)
 	if err != nil {
+		if readyWG != nil {
+			readyWG.Done()
+		}
 		return err
 	}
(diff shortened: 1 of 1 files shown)


pkg/server/server.go
```diff
@@ -134,14 +134,17 @@ func (s *Server) GetEventsWG(request *tetragon.GetEventsRequest, server tetragon
 	}).Debug("Received a GetEvents request")
 	allowList, err := filters.BuildFilterList(s.ctx, request.AllowList, filters.Filters)
 	if err != nil {
+		readyWG.Done()
```
michi-covalent, line 137: do we need to check if `readyWG` is nil? like here => https://github.com/cilium/tetragon/pull/2119/files#diff-104a11712684e520f0affa95e6a053faeb713306a10f69ee99b9729cc9dd1996R157-R159

willfindlay: Probably right, let me fix that

michi-covalent (approved):

Issue #2010, opened 2024-01-23
vmtests post-test delete-artifact fails on fork


mtardy:
See [an example of workflow here](https://github.com/cilium/tetragon/actions/runs/7593848324/job/20721370961?pr=1931):
```
Run geekyeggo/delete-artifact@65041433121f7239077fa20be14c0690f70569de
Error: HttpError: Resource not accessible by integration
```
It looks like a permissions error because this is running in the context of a fork, so we'll need to consider:
- Is it essential to delete the artifacts in case of success, Is saving 1GB of artifacts space necessary?
- If yes, is it okay to add the necessary permission for forks?

PR #2123, opened 2024-02-19
Set node_name to hostname if NODE_NAME envvar is not found 

lambdanis:
This enables node_name field in the standalone (non-k8s) mode.

```release-note
Set events node_name field to the hostname in the standalone (non-k8s) mode.
```

### DIFF
--- a/pkg/reader/node/node.go
+++ b/pkg/reader/node/node.go
@@ -3,7 +3,11 @@
 
 package node
 
-import "os"
+import (
+	"os"
+
+	"github.com/cilium/tetragon/pkg/logger"
+)
 
 // getNodeNameForExport returns node name string for JSON export. It uses NODE_NAME
 // env variable by default, which is also used by k8s watcher to watch for local pods:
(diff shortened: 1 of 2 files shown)


pkg/reader/node/node.go
```diff
@@ -13,8 +13,12 @@ import "os"
...
 func GetNodeNameForExport() string {
 	nodeName := os.Getenv("HUBBLE_NODE_NAME")
-	if nodeName != "" {
-		return nodeName
+	if nodeName == "" {
+		nodeName = os.Getenv("NODE_NAME")
 	}
-	return os.Getenv("NODE_NAME")
+	if nodeName == "" {
+		nodeName, _ = os.Hostname()
```
tixxdz, line 20: l would cache this using a read once just to avoid extra calls in a global environment to be consistent too with k8s. The hostname may change during runtime but I don't think we care. cc @jrfastab 

Also we need to report an error here if os.Hostname() fails for some reasons! and maybe use the old utsname but you have to convert then the output from int8 to byte then strings ... a bit complicated. the Hostname() seems to call explicitly /proc at this location where we may have it configured --procfs but this is an extreme case, so up to you, just warning on Hostname() errors should be enough.

Thank you @lambdanis

tixxdz (approved):
LGTM minor comments on caching, errors, fixing tests and maybe add test to ensure that it returns /proc/sys/kernel/hostname if no envs are set.

pkg/reader/node/node.go
lambdanis: Thanks! I added a warning log.

pkg/reader/node/node.go
```diff
@@ -12,9 +16,16 @@ import "os"
...
 	nodeName := os.Getenv("HUBBLE_NODE_NAME")
-	if nodeName != "" {
-		return nodeName
+	if nodeName == "" {
+		nodeName = os.Getenv("NODE_NAME")
+	}
+	if nodeName == "" {
+		nodeName, err = os.Hostname()
+		if err != nil {
+			logger.GetLogger().Warningf("failed to retrieve hostname: %v", err)
```
kkourt, line 27: not GetLogger.WithError(err).Warn("failed to retrieve hostname")?

lambdanis:
Ok, merging it, the only CI failure is the known https://github.com/cilium/tetragon/issues/2010

PR #2125, opened 2024-02-20
docs: Privileged execution - use nginx instead of test-pod image

janvi01:
Fixes: https://github.com/cilium/tetragon/issues/705
Signed-off-by: janvi01 [janvibajo1@gmail.com](mailto:janvibajo1@gmail.com)

### DIFF
--- a/docs/content/en/docs/use-cases/process-lifecycle/privileged-execution.md
+++ b/docs/content/en/docs/use-cases/process-lifecycle/privileged-execution.md
@@ -16,33 +16,66 @@ have. This would help us answer questions like:
 > Which Kubernetes pods have host network or pid namespace access in my
 > cluster?
 
-As a first step let's enable visibility to capability and namespace changes via
-the configmap by setting `enable-process-cred` and `enable-process-ns` from
-`false` to `true`:
-```bash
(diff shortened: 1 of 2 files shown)


tixxdz (approved):
lgtm much appreciated for your contributions!  let's wait for @mtardy review too ;-)

jrfastab:
docs build had some issue with network connections retry fixed it.

jrfastab:
Otherwise thanks for the patch lgtm.

docs/content/en/docs/use-cases/process-lifecycle/privileged-execution.md
```diff
@@ -16,33 +16,66 @@ have. This would help us answer questions like:
...
+
+  ```bash
+  kubectl edit cm -n kube-system tetragon-config
+  ```
+
+* Set the following flags from "false" to "true":
+
+  ```bash
+  # enable-process-cred: true
+  # enable-process-ns: true
```
mtardy, line 31: I think we wanted to remove the `#` here.

mtardy (commented):
Thanks! I was too late for the review, this is only nit, not a big deal, maybe the `#` in the config is something we would like to fix.

PR #2126, opened 2024-02-20
renovate: match all Makefiles

willfindlay:
Remove the restrictions on matching only some Makefiles in renovate, since it would be nice to auto update additional docker dependencies as we add them.


mtardy (approved):
splendid

mtardy:
https://github.com/cilium/tetragon/actions/runs/7977053986/job/21779047815?pr=2126 passed, we can merge! No need to wait for the rest.

PR #2131, opened 2024-02-21
makefile: containerize CRD generation and rename codegen and generate targets

willfindlay:
Containerize the CRD generation workflow to resolve a version consistency issue we encountered. Rename the generate and codegen targets to disambiguate them.


lambdanis (approved):

Makefile
```diff
@@ -291,10 +291,15 @@ image-test: image-clang
...
-image-codegen:
-	$(CONTAINER_ENGINE) build -f Dockerfile.codegen -t "cilium/tetragon-codegen:${DOCKER_IMAGE_TAG}" .
+image-protogen:
+	$(CONTAINER_ENGINE) build -f Dockerfile.protogen -t "cilium/tetragon-protogen:${DOCKER_IMAGE_TAG}" .
 	$(QUIET)@echo "Push like this when ready:"
-	$(QUIET)@echo "${CONTAINER_ENGINE} push cilium/tetragon-codegen:$(DOCKER_IMAGE_TAG)"
+	$(QUIET)@echo "${CONTAINER_ENGINE} push cilium/tetragon-protogen:$(DOCKER_IMAGE_TAG)"
+
+image-crdgen:
+	$(CONTAINER_ENGINE) build -f Dockerfile.crdgen -t "cilium/tetragon-crdgen:${DOCKER_IMAGE_TAG}" .
```
mtardy, line 300: I think your `Dockerfile.crdgen` is maybe not useful as you can just use the `docker.io/library/golang:1.22.0@sha256:7b297d9abee021bab9046e492506b3c2da8a3722cbf301653186545ecc1e00bb` image with a `docker run --env myenv`.

Makefile
```diff
@@ -330,8 +330,9 @@ tarball-clean:
 fetch-testdata:
 	wget -nc -P testdata/btf 'https://github.com/cilium/tetragon-testdata/raw/main/btf/vmlinux-5.4.104+'
 
-.PHONY: generate codegen protoc-gen-go-tetragon
-generate:
+.PHONY: crdgen generate codegen protoc-gen-go-tetragon
+generate: | crds
```
mtardy, line 329: genuinely curious, I just discovered that: what's the use for the `|`, I searched and saw it's for "Order-Only Prerequisites" but what's the interest here?

tools/protoc-gen-go-tetragon/main.go
mtardy, line 1: I love that `move`!

mtardy (approved):
Thanks that's a really great patch, I was always confused between `generate` and `codegen` and this removes useless stuff and put stuff in the right place! Thank you

Makefile
willfindlay: I'll be honest I'm not fully certain it's necessary since they're all PHONY targets but I've been burned in the past by not including this. Basically the point is to avoid `make` making (see what I did there) assumptions about the relationship between the two targets, when all I want to do is just run the damn thing.

PR #2135, opened 2024-02-22
Loading: Check if probe type is registered

kevsecurity:
The loadInstance() function loads a BPF program. It checks its type against the list of registered types and calls the appropriate handler if there is a match. Otherwise it tries to match against some known types. If that fails, it attempts to load the program as a kprobe.

If a new probe type is created but isn't registered (due to a typo, bug, or some other error), then the program will silently attempt to load as a kprobe, instead of using the supplied handler.

This commit makes a list of inbuilt probe types, and checks the program type against it (calling the appropriate handler if it is in the list), then checks the program type against the registered probe types (again, calling the appropriate handler if it exists). If both of these fail, it returns an error instead of silently attempting a kprobe load.

This will catch bugs where program types are accidentally specified with an invalid type, or where probes are registered with incorrect names.

```release-note
Add error checking to program loading with respect to probe type.
```

### DIFF
--- a/pkg/sensors/load.go
+++ b/pkg/sensors/load.go
@@ -16,7 +16,6 @@ import (
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/sensors/program"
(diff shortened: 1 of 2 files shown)


willfindlay (approved):
Makes sense to me.

PR #2143, opened 2024-02-23
add USERS.md file for collating tetragon users

paularah:


kevsecurity (approved):
Would be infinitely better with a commit message explaining why the change, what the change, etc, but otherwise LGTM.

kevsecurity:
Actually, the commit definitely needs to be signed. Use ```git commit -s``` (or similar) to add a signature and repush.

And, in addition, it needs a label added, e.g. ```realease-misc``` which you can do from the right hand column in github.com.

If you do both, then it should go all green hopefully. :D

paularah:
closing this and opening a fresh one on a clean slate :)

PR #2145, opened 2024-02-25
tetragon: Assorted sensor fixes

olsajiri:
Adding several fixes before the loader sysfs refactor takes place.

### DIFF
--- a/pkg/observer/observertesthelper/observer_test_helper.go
+++ b/pkg/observer/observertesthelper/observer_test_helper.go
@@ -25,6 +25,7 @@ import (
 	"github.com/cilium/tetragon/pkg/observer"
 	hubbleV1 "github.com/cilium/tetragon/pkg/oldhubble/api/v1"
 	hubbleCilium "github.com/cilium/tetragon/pkg/oldhubble/cilium"
+	"github.com/cilium/tetragon/pkg/policyfilter"
 	"github.com/cilium/tetragon/pkg/tracingpolicy"
 	"github.com/sirupsen/logrus"
 
@@ -339,10 +340,10 @@ func GetDefaultSensorsWithFile(tb testing.TB, file, lib string, opts ...TestOpti
 		}
 	}
 
-	var sensor *sensors.Sensor
+	var sens []*sensors.Sensor
 
 	if tp != nil {
-		sensor, err = sensors.GetMergedSensorFromParserPolicy(tp)
+		sens, err = sensors.SensorsFromPolicy(tp, policyfilter.NoFilterID)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


pkg/observer/observertesthelper/observer_test_helper.go
```diff
@@ -340,22 +341,23 @@ func GetDefaultSensorsWithFile(tb testing.TB, file, lib string, opts ...TestOpti
 		}
 	}
 
-	var sensor *sensors.Sensor
+	var sens []*sensors.Sensor
 
 	if tp != nil {
-		sensor, err = sensors.GetMergedSensorFromParserPolicy(tp)
+		sens, err = sensors.SensorsFromPolicy(tp, policyfilter.NoFilterID)
```
tixxdz, line 347: So I guess no side effects for now?

tixxdz (approved):

pkg/observer/observertesthelper/observer_test_helper.go
olsajiri: not as far as I can say ;-)

pkg/sensors/tracing/generickprobe.go
```diff
@@ -117,9 +121,8 @@ type genericKprobe struct {
 	// is there override defined for the kprobe
 	hasOverride bool
 
-	// reference to a stack trace map, must be closed when unloading the kprobe,
-	// this is done in the sensor PostUnloadHook
-	stackTraceMapRef *ebpf.Map
+	// global data for multi kprobe, single otherwise
+	data *genericKprobeData
```
mtardy, line 125: I don't get the `data` naming, what is the intent behind the name? Could we get something more explicit or not really?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -577,23 +585,6 @@ func createGenericKprobeSensor(
...
-				if gk.stackTraceMapRef != nil {
-					err = gk.stackTraceMapRef.Close()
-					if err != nil {
-						errs = errors.Join(errs, fmt.Errorf("failed to close map: %v", gk.stackTraceMapRef))
-					}
-					gk.stackTraceMapRef = nil
-				}
-			}
-			return errs
-		},
```
mtardy, line 596: so we don't need to worry about the ref to this map being closed it's already done by other code?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1116,25 +1110,13 @@ func handleMsgGenericKprobe(m *api.MsgGenericKprobe, gk *genericKprobe, r *bytes
...
-			if gk.stackTraceMapRef != nil {
-				err = gk.stackTraceMapRef.Lookup(id, &unix.StackTrace)
+			if gk.data.stackTraceMap.MapHandle != nil {
+				err = gk.data.stackTraceMap.MapHandle.Lookup(id, &unix.StackTrace)
 				if err != nil {
 					logger.GetLogger().WithError(err).Warn("failed to lookup the stacktrace map")
 				}
+			} else {
+				logger.GetLogger().WithError(err).Warn("failed to load the stacktrace map")
 			}
```
mtardy, line 1120: that certainly looks cleaner!

mtardy (approved):
I checked the commit for the stack traces. I have a few comments but I'm happy with it if the stack traces continue to work (and hopefully they do since there is a small `TestKprobeStackTrace` test)!

pkg/sensors/tracing/generickprobe.go
olsajiri: it's sensor specific data that needs to be present in kprobeEntry, because we need it when we process event,
so it's specific for each kprobeEntry when we use single kprobes and it's one global instance when we use kprobe multi.. I'll update the comment with this

pkg/sensors/tracing/generickprobe.go
olsajiri: yes, stackTraceMap goes away when the sensor is unloaded, so it's safe

Issue #360, opened 2022-08-25
tetragon: print current main tetragon security context from /proc/self/attr


tixxdz:
Early at startup read current security context and print it, this will help debug cryptic errors cases if SELinux, Apparmor are applied to tetragon, even if we are under a seccomp...

tixxdz:
Information should be obtained from /proc/self/status,  /proc/self/attr/current and related directories inside /proc/self/attr/

ShubhamPalriwala:
Hey @tixxdz, I have been reading up a lot on eBPF space recently and would like to get involved with Tetragon and feel like this will help me start my deep dive into it. Could you help me with what am I supposed to do to solve this issue? Below is my understanding till now:
- As soon as Tetragon starts, we are supposed to log this information to help us get other 3rd party errors/warnings beforehand itself
- Now, this information can be found in the directories you have mentioned

However, I'm not clear with the following things:
- What should and should not be logged?

Also, if you could point me to any resource(article, video, blog) that could help me get a clear understanding wrt this issue/related, I would really really be thankful to you.

Would like to apologize already if I asked any dumb doubt :))

tixxdz:
Hi @ShubhamPalriwala 

The aim here is to log security context of Tetragon when it starts. If there are some Linux security module that are active they may restrict Tetragon, this may cause errors and can be hard to debug as probably errors are in the audit subsystem if it is active or in other logs... which are usually not easy to access.

This is mainly help debug Tetragon issue task.

You can start reading about /proc/self/attr format and what's there, then split tasks to print security contexts based on the LSM being used if it's active or not, etc

We want mainly the security context of tetragon in SELinux, Apparmor or any other bit that can be useful to say hey maybe tetragon is restricted in some way...

Just read about https://man7.org/linux/man-pages/man5/proc.5.html , then try to get your way from that ;-)

Anyway useful information should be logged during startup at info() level

Thanks!

ShubhamPalriwala:
Thank you so much for the detailed information, I've went through a couple of talks:
- [Seccomp](https://www.youtube.com/watch?v=QfJtKOtXeiw)
- [AppArmor and SELinux](https://www.youtube.com/watch?v=JFjXvIwAeVI)

getting a clear idea of `seccomp`, `AppArmour`, and `SELinux`. 

And I just went through the `proc` man page to understand the whats and hows of it.

According to me, the things that we can log right now can be:
[...]

PR #2146, opened 2024-02-25
tetragon: log current security context if any at startup

tixxdz:
closes: #360


tixxdz:
closed in favor of https://github.com/cilium/tetragon/pull/2149 (wrong remote without updates)

PR #2148, opened 2024-02-26
tetragon:trivial: create run dir early to avoid errors

tixxdz:
Create the /run/tetragon runtime directory of Tetragon early, as we have
code that tries to lookup for that directory early and could report
errors. One example is when we are looking for pidfile of tetragon.

On host install we always keep the cgroup2 mount inside /run/tetragon
which pins the runtime directory on the host forever. We do not have
strict unmount logic as there is no harm in keeping another mirror
of cgroup2 inside /run/tetragon, so on Tetragon restarts the runtime
will always be there. But on first ever start the directory won't be
there and errors will be reported.

Fix this by having the mkdir as early as possible before anything else.

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -173,6 +173,9 @@ func tetragonExecute() error {
 	log.WithField("version", version.Version).Info("Starting tetragon")
 	log.WithField("config", viper.AllSettings()).Info("config settings")
 
+	// Create run dir early
+	os.MkdirAll(defaults.DefaultRunDir, 0755)
+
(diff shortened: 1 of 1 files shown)


kevsecurity (approved):
I think the commit needs a message explaining why the change is helpful.
Also, it needs a label, like ```release/misc```

kevsecurity:
Also, the Packages e2e Tests failure looks relevant: ```stat: cannot statx '/var/run/tetragon/tetragon.sock': Permission denied```

kevsecurity (changes requested):
Looks like it caused a CI failure?

kevsecurity (approved):

PR #2149, opened 2024-02-26
tetragon: log current security context if any at startup

tixxdz:
closes: #360

### DIFF
--- a/pkg/reader/proc/proc.go
+++ b/pkg/reader/proc/proc.go
@@ -10,6 +10,10 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/option"
+	"github.com/sirupsen/logrus"
 )
 
 // Status reflects fields of `/proc/[pid]/status` and other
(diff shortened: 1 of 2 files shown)


pkg/reader/proc/proc.go
```diff
@@ -249,3 +253,35 @@ func PrependPath(s string, b []byte) []byte {
...
+			logLSM = true
+		}
+	}
+
+	if logLSM {
+		logger.GetLogger().WithFields(logrus.Fields{
+			"SELinux":  lsms["selinux"],
+			"AppArmor": lsms["apparmor"],
+			"Smack":    lsms["smack"],
+		}).Info("Tetragon current Security context")
```
kkourt, line 285:
```suggestion
		}).Info("Tetragon current security context")
```

tpapagian (approved):

pkg/reader/proc/proc.go
tixxdz: fixed

tixxdz:
Added lockdown context on top, as Tetragon was failing for some users running in confidentiality mode that blocks kprobes.

olsajiri (approved):

pkg/reader/proc/proc.go
```diff
@@ -277,11 +277,24 @@ func LogCurrentSecurityContext() {
...
+	data, err := os.ReadFile("/sys/kernel/security/lockdown")
+	if err == nil && len(data) > 0 {
+		values := strings.TrimSpace(string(data))
+		i := strings.Index(values, "[")
+		j := strings.Index(values, "]")
+		if i >= 0 && j > i {
+			lockdown = values[i+1 : j]
+			logLSM = true
+		}
+	}
```
kkourt, line 290: How about emitting a warning if lockdown is == "confidentiality"?

Maybe something like
"lockdown mode seems to be set to 'confidentiality', which does not allow Tetragon to operate because it prohibits loading BPF programs. See https://github.com/iovisor/bcc/issues/2565 for more details".

tixxdz: Ok make sense, I added it thank you!

Also created a doc issue for it https://github.com/cilium/tetragon/issues/2265

mtardy (approved):
Thanks that's great!

PR #2151, opened 2024-02-26
Initialize metrics with labels

lambdanis:
So far metrics with labels were registered but not initialized at startup. This leads to unpredictable resources usage and limits certain registry operations, e.g. linting or autogenerating docs.

This PR initializes metrics with known labels. It required explicitly defining known labels, so there are lots of commits doing just that - defining types for label values. Let me know if you want me to squash some of these commits. I also used `InitMetrics` functions as a notebook and left some comments for the future about metrics naming - hope that's ok.

A few remarks about the impact and limitations of this approach:
* Initializing metrics at startup means increased initial resources usage. However, it also means that metrics volume is more predictable and can be planned for - that's good.
* Some metrics have labels that depend on the environment: process metadata (namespace, workload, pod, binary), policy name, hook point. These metrics are not initialized.
* Some metrics have known labels, but are registered as custom collectors. These are also not initialized.


lambdanis:
I'll re-open this PR from a branch in the repo not a fork, so that I can stack follow-up PRs.

PR #2154, opened 2024-02-26
ci: only check docs on PR to main

willfindlay:
It doesn't make sense to run the links checker on old release branches, since we only deploy docs from main.


mtardy (approved):
agreed!

PR #2162, opened 2024-02-28
Initialize metrics with labels

lambdanis:
So far metrics with labels were registered but not initialized at startup. This leads to unpredictable resources usage and limits certain registry operations, e.g. linting or autogenerating docs.

This PR initializes metrics with known labels. It required explicitly defining known labels, so there are lots of commits doing just that - defining types for label values. Let me know if you want me to squash some of these commits. I also used `InitMetrics` functions as a notebook and left some comments for the future about metrics naming - hope that's ok.

A few remarks about the impact and limitations of this approach:
* Initializing metrics at startup means increased initial resources usage. However, it also means that metrics volume is more predictable and can be planned for - that's good.
* I changed some (not many) label values to be more useful. These are technically breaking changes. Details are documented in relevant commit messages.
* Some metrics have labels that depend on the environment: process metadata (namespace, workload, pod, binary), policy name, hook point. These metrics are not initialized.
* Some metrics have known labels, but are registered as custom collectors. These are also not initialized.

```release-note
Metrics with known labels values are initialized to 0 on startup. This helps to ensure stable resources usage and metrics queries. This also involves changes in several metrics labels:
* error_type label on tetragon_handler_errors_total metric is either "unknown_opcode" or "event_handler_failed" instead of the Go type of the error
* event_type label on tetragon_event_cache*_errors_total metrics is one of the values defined in Tetragon API (tetragon.EventType) instead of the Go type of the event
* error label on tetragon_event_cache_errors_total metric is "nil_process_pid"
* error label is removed from tetragon_policyfilter_metrics_total metric
```


pkg/grpc/tracing/stats.go
```diff
@@ -19,6 +19,9 @@ var (
 
 func InitMetrics(registry *prometheus.Registry) {
 	registry.MustRegister(LoaderStats)
+
+	// NOTES:
+	// * Rename process_loader_stats metric (to e.g. process_loader_events_total) and count label (to e.g. event)?
```
willfindlay, line 29: Does it make sense to link GH issues here?

pkg/metrics/errormetrics/errormetrics.go
```diff
@@ -12,25 +12,39 @@ import (
...
+	ProcessCacheMissOnRemove:       "process_cache_miss_on_remove",
+	ProcessPidTidMismatch:          "process_pid_tid_mismatch",
+	EventMissingProcessInfo:        "event_missing_process_info",
+	HandlerError:                   "handler_error",
+	EventFinalizeProcessInfoFailed: "event_finalize_process_info_failed",
+}
+
+func (e ErrorType) String() string {
+	return errorTypeLabelValues[e]
+}
```
willfindlay, line 45: Should we have a fallback here for missing types to avoid panics?

pkg/api/ops/ops.go
```diff
@@ -68,18 +68,22 @@ const (
...
-		13:  "GenericKprobe",
-		14:  "GenericTracepoint",
-		23:  "Clone",
-		24:  "Data",
-		25:  "Cgroup",
-		254: "Test",
-	}[op]
+	s, ok := OpCodeStrings[op]
+	if !ok {
+		return ""
```
willfindlay, line 94: How about:
```go
return fmt.Sprintf("Unknown(%d)", op)
```

And perhaps log an error so we know there's a bug?

pkg/metrics/eventcachemetrics/eventcachemetrics.go
```diff
@@ -8,12 +8,24 @@ import (
...
 )
 
+var cacheEntryTypeLabelValues = map[CacheEntryType]string{
+	ProcessInfo: "process_info",
+	ParentInfo:  "parent_info",
+	PodInfo:     "pod_info",
+}
+
+func (t CacheEntryType) String() string {
+	return cacheEntryTypeLabelValues[t]
```
willfindlay, line 27: Same comment as in errormetrics.

pkg/metrics/kprobemetrics/kprobemetrics.go
```diff
@@ -8,6 +8,22 @@ import (
...
+	MergeErrorTypeExit
+)
+
+var mergeErrorTypeLabelValues = map[MergeErrorType]string{
+	MergeErrorTypeEnter: "enter",
+	MergeErrorTypeExit:  "exit",
+}
+
+func (t MergeErrorType) String() string {
+	return mergeErrorTypeLabelValues[t]
```
willfindlay, line 24: Same comment as errormetrics.

pkg/metrics/policyfiltermetrics/policyfiltermetrics.go
```diff
@@ -12,6 +12,22 @@ import (
...
+	PodHandlersSubsys
+)
+
+var subsysLabelValues = map[Subsys]string{
+	PodHandlersSubsys: "pod-handlers",
+	RTHooksSubsys:     "rthooks",
+}
+
+func (s Subsys) String() string {
+	return subsysLabelValues[s]
```
willfindlay, line 24: Same comment as errormetrics.

pkg/metrics/watchermetrics/watchermetrics.go
```diff
@@ -8,6 +8,22 @@ import (
...
+const (
+	K8sWatcher Watcher = iota
+)
+
+var watcherTypeLabelValues = map[Watcher]string{
+	K8sWatcher: "k8s",
+}
+
+func (w Watcher) String() string {
+	return watcherTypeLabelValues[w]
```
willfindlay, line 24: Same comment as errormetrics.

willfindlay (approved):

willfindlay:
LGTM, I just left a few nits/comments. Feel free to take them or leave them.

pkg/metrics/errormetrics/errormetrics.go
lambdanis: It won't panic, in case `e` is not in `errorTypeLabelValues` it will return an empty string (zero value).

pkg/api/ops/ops.go
lambdanis: done, I added an info log

pkg/grpc/tracing/stats.go
lambdanis: yeah, I think so, just I'm not 100% sure which of these points are worth implementing and which not. Many of these are technically breaking changes, but also many metrics are probably used only by Tetragon developers so :woman_shrugging: I need one more pass over the metrics and maybe another pair of eyes, then I'll open issues for improvements.

lambdanis:
I added a release note to the PR description, covering all changes in metrics & labels. I'm merging this so that #2164 is unblocked.

pkg/metrics/policyfiltermetrics/policyfiltermetrics.go
```diff
@@ -66,9 +62,8 @@ func InitMetrics(registry *prometheus.Registry) {
 	// * Rename policyfilter_metrics_total to get rid of _metrics?
 }
 
-func OpInc(subsys Subsys, op Operation, err error) {
+func OpInc(subsys Subsys, op Operation) {
 	PolicyFilterOpMetrics.WithLabelValues(
 		subsys.String(), op.String(),
-		strings.ReplaceAll(fmt.Sprintf("%T", errors.Cause(err)), "*", ""),
```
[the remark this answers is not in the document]
lambdanis: IIRC this label didn't add much information about the error (the Go type of the error was something generic in all cases), but I see now we can't even say whether there was an error or not. Somehow I missed that.

Let's add an `error` label then. To make the label bounded and also more informative, a typical approach would be to define an error type and constants, like in [errormetrics](https://github.com/cilium/tetragon/blob/22eb995b19207ac0ced2dd83950ec8e8aedd122d/pkg/metrics/errormetrics/errormetrics.go) for example.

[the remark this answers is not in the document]
kkourt: PTAL: https://github.com/cilium/tetragon/pull/2282

PR #2164, opened 2024-02-28
Generate metrics reference docs

### LINKED ISSUE #1856: docs: Add a metrics reference
Ideally autogenerated from the metrics definitions.

lambdanis:
Add a metrics reference documentation page, autogenerated with [github.com/isovalent/metricstool](https://github.com/isovalent/metricstool/tree/v0.1.0/pkg/metricsmd).

Fixes #1856


Makefile
```diff
@@ -411,4 +411,16 @@ kind-install-tetragon:
...
 
 .PHONY: kind-setup
-kind-setup: images kind kind-install-tetragon
\ No newline at end of file
+kind-setup: images kind kind-install-tetragon
+
+METRICS_DOCS_PATH := docs/content/en/docs/reference/metrics.md
+
+.PHONY: metrics-docs
+metrics-docs: tetra
```
willfindlay, line 419: Can we document this in the `help` target above?

willfindlay (approved):
This looks AMAZING :tada:

mtardy:
Where's the Netlify preview link when we need it?

[Preview here](https://deploy-preview-2124--tetragon.netlify.app/docs/reference/metrics/).

cmd/tetra/metrics/print.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+func New() *cobra.Command {
+	targets := map[string]string{
+		"health":    "Tetragon Health",
+		"resources": "Tetragon Resources",
+		"events":    "Tetragon Events",
+	}
+
+	overrides := []metricsmd.LabelOverrides{
```
mtardy, line 23: do you need to provide "false" values here or can you just pass an empty object? Can be a bit misleading

docs/content/en/docs/reference/metrics.md
```diff
@@ -0,0 +1,422 @@
+---
+title: "Metrics Reference"
```
mtardy, line 2: I would remove this as it's already under the Reference section, like for the other ones.
```suggestion
title: "Metrics"
```

docs/content/en/docs/reference/metrics.md
```diff
@@ -0,0 +1,422 @@
...
+description: >
+  This reference documents Prometheus metrics exposed by Tetragon.
```
mtardy, line 4: The other descriptions explain a bit how this doc is generated, with:
- This reference is generated from the Tetragon Helm chart values.
- This reference is generated from the protocol buffer specification and documents the gRPC API of Tetragon.

maybe we can add a hint as well that could interesting?

docs/content/en/docs/reference/metrics.md
```diff
@@ -0,0 +1,422 @@
...
+weight: 4
+---
+<!-- This file is autogenerated via https://github.com/isovalent/metricstool -->
```
mtardy, line 7:
```suggestion
{{< comment >}}
This page is autogenerated via https://github.com/isovalent/metricstool,
please do not edit directly.
{{< /comment >}}
```
Please use the `comment` shortcode so that Hugo removes this before generating the website, with HTML comments, they end up in the final file served by the HTTP server and are still here in prod.

docs/content/en/docs/reference/metrics.md
```diff
@@ -0,0 +1,422 @@
...
+
+# Tetragon Health Metrics
```
mtardy, line 9: Could you make it a level 2 title (with all the other level 1 titles), otherwise it breaks the hierarchy and the formatting of the page 
```suggestion
## Tetragon Health Metrics
```

For example later:
<details><summary>Click to see preview</summary>
<p>
image
</p>
</details> 


It does not expect a level 1 after the page level 1 that is already the title from the frontmatter.

docs/content/en/docs/reference/metrics.md
```diff
@@ -0,0 +1,422 @@
...
+
+## `tetragon_build_info`
```
mtardy, line 11: Similarly, can we make every metric name a level 3 title?
```suggestion
### `tetragon_build_info`
```

docs/content/en/docs/reference/metrics.md
mtardy, line 1: I made some comments on the output of your tool, I see two ways of tweaking stuff:
- maybe you want your output to be consistent with a previous behavior and then we need some `sed` commands to modify the output as I do with gRPC and Helm (because I don't control the tool).
- you can simply modify your tool output directly.

pkg/metrics/policystatemetrics/policystatemetrics.go
```diff
@@ -85,3 +87,47 @@ func (c *policyStateCollector) Collect(ch chan<- prometheus.Metric) {
 		strings.TrimPrefix(strings.ToLower(tetragon.TracingPolicyState_TP_STATE_ENABLED.String()), "tp_state_"),
 	)
 }
+
+// policyStateZeroCollector implements prometheus.Collector. It collects "zero"
+// metrics. It's intended to be used when the sensor manager doesn't exist, but
+// we still want Prometheus metrics to be exposed.
```
mtardy, line 93: Changes in this file looks good to me, it's a nice way to populate those custom collector metrics, thanks! Maybe a nit would have been to mention that it's used for the metrics docs in this comment, but it's pretty obvious with the `InitMetricsForDocs` above and the git commit message so it's already good.

mtardy (commented):
Thanks, that will be a very useful addition to the doc!! Sorry I reviewed that one first but I wanted to highlight some things on the output of the autogenerated page.

docs/content/en/docs/reference/metrics.md
mtardy: See
image

cmd/tetra/metrics/print.go
lambdanis: I could pass empty strings as label values, then the docs would still list the label names but not values.

I'm thinking having example values is helpful (users can see the format), but currently the reference doesn't distinguish between example and real values, so I see how it can be misleading. Maybe the next iteration of metricstool could support rendering example and real label values differently or something.

lambdanis: For the build info labels specifically, we can make them less confusing by updating them on a release too.

docs/content/en/docs/reference/metrics.md
lambdanis: I opted to drop this comment completely and document that the page is autogenerated in the frontmatter description instead. Do you think that's enough?

lambdanis: ok, done

lambdanis:
Ok, after a few rebases this should be good now.

Here is the final docs preview: https://deploy-preview-2164--tetragon.netlify.app/docs/reference/metrics/
@mtardy Please take another look if you want.

docs/content/en/docs/reference/metrics.md
```diff
@@ -0,0 +1,420 @@
+---
+title: "Metrics"
+description: >
+  This reference is autogenerated from the Tetragon Prometheus metrics registry
+  using github.com/isovalent/metricstool
```
mtardy, line 5: This looks a bit off since links are not handled in the description and that's referring to `isovalent` repo. Maybe we can just remove "using github.com/isovalent/metricstool", or not if you want to promote that one, but then maybe in the body?

It looks a bit weird in the section TOC as well.

image

docs/content/en/docs/reference/metrics.md
```diff
@@ -0,0 +1,420 @@
...
+### `tetragon_build_info`
+
+Build information about tetragon
+
+| label | values |
+| ----- | ------ |
+| `commit` | `931b70f2c9878ba985ba6b589827bea17da6ec33` |
+| `go_version` | `go1.22.0` |
+| `modified` | `false` |
+| `time ` | `2022-05-13T15:54:45Z` |
```
mtardy, line 22: I like that it's referencing the OSS commit 😻

mtardy (approved):
I think it looks good now and will be very useful, thanks a lot!

docs/content/en/docs/reference/metrics.md
lambdanis: Yeah, this looks weird, I'll just remove it.

PR #2166, opened 2024-02-29
tetragon: Move enforcer-tester to .gitignore

olsajiri:
Rename leftover.


mtardy (approved):
Thanks! As we discussed offline, would be great to autodetect those as it happens from time to time.

PR #2174, opened 2024-03-01
ci/static-checks: ensure no binary files are checked in

willfindlay:
We've run into several instaces where binary files are accidentally checked into source control. Add a step to static-checks that will watch for this.


olsajiri (approved):
nice, I like the check for the binary file

kevsecurity (approved):

PR #2175, opened 2024-03-01
Support user mode stacktraces in events

anfedotoff:
Hi :wave:! 
I was inspired by #1429 and decided to add the ability for collecting user mode stacktraces. User-mode stacktraces allow you to enrich information about events and understand why this event occurred. For example, this information will help you understand why the seccomp policy is triggered.
Please, have a look, I'm ready to make changes if needed. Thanks!

```release-note
Support user mode stacktraces in events.
```


mtardy:
Welcome! Thanks for taking the time to add this, I started the CI, and some checks might be worth a look to see what's wrong. I'll try to review your PR quickly!

The only thing that immediately came to my mind is that we would break compatibility if we changed the name of the previous stack trace thing to kernel stack trace, but that might make sense to do so. Let me take a look and we'll see.

anfedotoff:
I'm trying to fix vmtests. But I'm lack of logs. I think, the problem is that there is no gcc  on VM image and I couldn't compile test binary.  @mtardy, do you have any ideas how to handle this in the right way?

mtardy:
> I think, the problem is that there is no gcc on VM image and I couldn't compile test binary. @mtardy, do you have any ideas how to handle this in the right way?

Yes, so usually, the C programs we need for testing are located in `contrib/tester-progs`, you just need to add them in the Makefile `PROGS` variables and also add the output binary name in the `contrib/tester-progs/.gitignore` and they should be there for you when testing. Those are built with `make tester-progs` from the top level Makefile.

mtardy:
Thanks a lot, tests look good now so it's mostly on me! Again I'll do a proper review soon :)

bpf/process/types/basic.h
```diff
@@ -2253,18 +2253,25 @@ do_action(void *ctx, __u32 i, struct msg_generic_kprobe *e,
...
 			// Here we just signal that there was a collision returning -EEXIST.
-			e->stack_id = get_stackid(ctx, &stack_trace_map, 0);
+			e->kernel_stack_id = get_stackid(ctx, &stack_trace_map, 0);
+		}
+
+		__u32 user_stack_trace = actions->act[++i];
+
+		if (user_stack_trace) {
+			e->common.flags |= MSG_COMMON_FLAG_USER_STACKTRACE;
+			e->user_stack_id = get_stackid(ctx, &stack_trace_map, BPF_F_USER_STACK);
```
mtardy, line 2274: For some reason (I need to investigate), when tracing many events, using `kfree_skb_reason` for example, I get a lot of `failed to retrieve user stacktrace: id equal to errno -14` which is `EFAULT` Bad address. I need to investigate why it happens.

contrib/tester-progs/go/getcpu/getcpu.go
```diff
@@ -18,5 +19,6 @@ func main() {
 		uintptr(unsafe.Pointer(&node)),
 		0,
 	)
+	time.Sleep(1 * time.Second)
```
mtardy, line 22: may I ask why you needed this?

docs/content/en/docs/concepts/tracing-policy/selectors.md
mtardy, line 1: Thanks a lot for taking the time to update the documentation as well.

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.19.0
 	github.com/prometheus/client_model v0.6.0
+	github.com/prometheus/procfs v0.12.0
```
mtardy, line 35: I wondered if we could avoid adding another explicit dependency for Tetragon, maybe for the little usage in `procsyms.go` we can do it "by hand" as we do so far.

pkg/encoder/encoder.go
```diff
@@ -201,14 +201,23 @@ func PrintNS(ns int32) string {
...
 				fmt.Fprintf(out, "+")
 				colorer.Yellow.Fprintf(out, "0x%x\n", st.Offset)
 			}
 		}
+		if ev.ProcessKprobe.UserStackTrace != nil {
+			fmt.Fprintf(out, "User space:\n")
+			for _, st := range ev.ProcessKprobe.UserStackTrace {
+				colorer.Green.Fprintf(out, "   0x%x:", st.Address)
+				colorer.Blue.Fprintf(out, " %s ", st.Symbol)
+				colorer.Yellow.Fprintf(out, "(%s+0x%x)\n", st.Module, st.Offset)
```
mtardy, line 218: This is a nit but I would maybe use a slightly different print format to avoid double space `" %s "` in case we don't have the symbol. Maybe something like `symbol@module+offset` idk what the usage.

pkg/option/flags.go
```diff
@@ -89,7 +89,7 @@ const (
 	KeyEnablePodInfo          = "enable-pod-info"
 	KeyEnableTracingPolicyCRD = "enable-tracing-policy-crd"
 
-	KeyExposeKernelAddresses = "expose-kernel-addresses"
+	KeyExposeAddresses = "expose-addresses"
```
mtardy, line 92: As we split everything in kernel-stack and user-stack, might be useful to also split this here. Or use a name like `expose-stack-addresses`. The concern was mostly about kernel address in the first place so that would be better if the idea of exposing _kernel_ address is still around.

pkg/encoder/encoder.go
```diff
@@ -201,14 +201,23 @@ func PrintNS(ns int32) string {
...
+			fmt.Fprintf(out, "Kernel:\n")
+			for _, st := range ev.ProcessKprobe.KernelStackTrace {
 				colorer.Green.Fprintf(out, "   0x%x:", st.Address)
 				colorer.Blue.Fprintf(out, " %s", st.Symbol)
 				fmt.Fprintf(out, "+")
 				colorer.Yellow.Fprintf(out, "0x%x\n", st.Offset)
 			}
 		}
+		if ev.ProcessKprobe.UserStackTrace != nil {
+			fmt.Fprintf(out, "User space:\n")
```
mtardy, line 214: nit, but since you just wrote "Kernel" before.
```suggestion
			fmt.Fprintf(out, "User:\n")
```

pkg/procsyms/procsyms.go
mtardy, line 1: I might need to take a closer look and see what we do regarding the new dep. But this looks good.

mtardy (commented):
Thanks a lot for doing this. I tried it and took a look, there is just a small issue with the fact that the BPF helper might return a lot of errors and I just need to check why and what we should do with that. Also, another thank you for adding proper testing :)!

Most questions are about retro-compatibility and deps addition choices. I think it's a nice opportunity that tonight we have the first Tetragon community call so I would love to bring this to the discussion. If you can join that would be lovely, otherwise I'll write the conclusion of the discussion here anyway. See https://isogo.to/tetragon-meeting-notes.

Again thank you, this does look very cool on dynamically linked binaries, especially when symbols are not stripped, it gives awesome observability!

anfedotoff:
> Thanks a lot for doing this. I tried it and took a look, there is just a small issue with the fact that the BPF helper might return a lot of errors and I just need to check why and what we should do with that. Also, another thank you for adding proper testing :)!
> 
> Most questions are about retro-compatibility and deps addition choices. I think it's a nice opportunity that tonight we have the first Tetragon community call so I would love to bring this to the discussion. If you can join that would be lovely, otherwise I'll write the conclusion of the discussion here anyway. See https://isogo.to/tetragon-meeting-notes.
> 
> Again thank you, this does look very cool on dynamically linked binaries, especially when symbols are not stripped, it gives awesome observability!

Thank you for review and invitation! I'll try take a part in the meeting where we can discuss the PR details! I will answer on comments soon).

bpf/process/types/basic.h
anfedotoff: Hmm, do you use both kernel and user stack trace capturing? Maybe the problem is because the [map](https://github.com/cilium/tetragon/blob/497af988f265a8197c43a390fc5989e2a7b0aea8/bpf/process/types/basic.h#L2223) is overflowed? Is this case reproducible when only kernel stack traces are captured?

contrib/tester-progs/go/getcpu/getcpu.go
anfedotoff: This is because user stack trace capturing and addresses resolving are async. From bpf we get only linear address and we need some time to resolve them in user space. During this, target process may be finished and we arn't able to read proc/<pid>/maps. So, I put 1 sec sleep in the test because it  finishes too fast to be captured.

pkg/encoder/encoder.go
anfedotoff: I fixed double space. Maybe keep this output? [ASAN](https://clang.llvm.org/docs/AddressSanitizer.html) prints stack traces this way. I think there are many people familiar with ASAN output. Here is some [examples](https://www.osc.edu/resources/getting_started/howto/howto_use_address_sanitizer).

P.S. ASAN has double space problem :smile: . For me it's OK if you prefer another output format.

bpf/process/types/basic.h
anfedotoff: If you share some steps how to reproduce this, I also could try to investigate the reason.

pkg/encoder/encoder.go
mtardy: reproducing ASAN output is a very nice idea and a good argument!

mtardy:
Could you `git checkout main && git fetch && git pull origin main && git checkout user-stacktrace && git rebase main` from your branch, run `make protogen` and then push the rebased version with `git push origin user-stacktrace --force-with-lease`. Otherwise, with conflicts we can't run the CI.

bpf/process/types/basic.h
mtardy: yep yep indeed, I would love to understand why it happens sometimes but I guess it's what we will end up doing. Sorry for being slow on that review.

anfedotoff: It's OK, we are slowly but right moving forward!

anfedotoff:
Hi, @mtardy! 
Could you tell me, please, is there anything to do here on my side for now?

mtardy:
> Hi, @mtardy! Could you tell me, please, is there anything to do here on my side for now?

Hey, sorry last week was KubeCon week in Paris and I was busy there. Not really, I've been chatting about the EFAULT error and there's not much we can do, from time to time, we'll indeed get a page fault when retrieving the trace from the kernel side and we can't avoid it. The best thing we could do is to warn for this in the doc (with some `{{< caution >}}` for example) on why it can happen and then this is not to be relied upon for very sensitive operations!

I think your PR is ready for me to take a last look and merge, even if some stuff is missing I'll take the time to add them after (like the doc thing I mentioned), I don't want you to wait longer again.

mtardy:
One thing we could do is to keep the `--expose-address` flag (in addition to the new one) but as deprecated, you can maybe see there's a way to put that thing hidden so that people's config keeps working. Then we can remove it in some minor versions.

mtardy:
the check is broken currently on forks but could you bump the version here https://github.com/cilium/tetragon/blob/main/pkg/k8s/apis/cilium.io/v1alpha1/version.go.

mtardy:
it seems that you might need to generate properly some files with `make help` can help: https://github.com/cilium/tetragon/actions/runs/8434642759/job/23106133639?pr=2175

anfedotoff:
> One thing we could do is to keep the `--expose-address` flag (in addition to the new one) but as deprecated, you can maybe see there's a way to put that thing hidden so that people's config keeps working. Then we can remove it in some minor versions.

You mean  to keep `expose-kernel-addresses`  as alias to `expose-stack-addresses`?  `expose-kernel-addresses` is the original name used before.

anfedotoff:
> Those two comments are mostly nits, if your PR passes the tests, we can consider merging and fixing the flag and the docs after.

Yes, let's do this way!

anfedotoff:
> The best thing we could do is to warn for this in the doc (with some {{< caution >}} for example) on why it can happen and then this is not to be relied upon for very sensitive operations!

It's a good point, but I don't understand why EFAULT is happened sometimes. I only have some ideas, but they might be wrong. Anyway some `{{< caution >}}` message would be helpful for users.

anfedotoff:
I think, it is good to squash all commits in one and write some commit description. I can do when PR will be camera ready  or it could be done while merging.

mtardy:
> I think, it is good to squash all commits in one and write some commit description. I can do when PR will be camera ready or it could be done while merging.

I think you could squash some things but please keep the separation for, for example:
- code
- addition of test
- documentation
- make generate etc... + bump of crd.

mtardy:
> > One thing we could do is to keep the `--expose-address` flag (in addition to the new one) but as deprecated, you can maybe see there's a way to put that thing hidden so that people's config keeps working. Then we can remove it in some minor versions.
> 
> You mean to keep `expose-kernel-addresses` as alias to `expose-stack-addresses`? `expose-kernel-addresses` is the original name used before.

ah indeed. Maybe yes that would be a good idea to not break compatibility.

anfedotoff:
@mtardy It seems to me, that ARM tests doesn't fit the limit...

mtardy:
> @mtardy It seems to me, that ARM tests doesn't fit the limit...

yes it reminds me of this https://github.com/cilium/tetragon/issues/2210, what's weird is that usually if we rerun we should have a successful run.

anfedotoff:
> > @mtardy It seems to me, that ARM tests doesn't fit the limit...
> 
> yes it reminds me of this #2210, what's weird is that usually if we rerun we should have a successful run.

Do you have any ideas how to debug this? Maybe let's raise the limit and see are they successful at all?

mtardy:
> > > @mtardy It seems to me, that ARM tests doesn't fit the limit...
> > 
> > 
> > yes it reminds me of this #2210, what's weird is that usually if we rerun we should have a successful run.
> 
> Do you have any ideas how to debug this? Maybe let's raise the limit and see are they successful at all?

You can try to push a patch by bumping the two timeouts (one in the go tests and one in the gha workflow), as of now there's no way to tell unfortunately if:
- this is flaky anyway
- your PR makes this thing fail constantly
- your PR made it worse

I can try to run locally as I have an arm64 machine with your PR and see if this appears or not!

anfedotoff:
> > > > @mtardy It seems to me, that ARM tests doesn't fit the limit...
> > > 
> > > 
> > > yes it reminds me of this #2210, what's weird is that usually if we rerun we should have a successful run.
> > 
> > 
> > Do you have any ideas how to debug this? Maybe let's raise the limit and see are they successful at all?
> 
> You can try to push a patch by bumping the two timeouts (one in the go tests and one in the gha workflow), as of now there's no way to tell unfortunately if:
> 
> * this is flaky anyway
> * your PR makes this thing fail constantly
> * your PR made it worse
> 
> I can try to run locally as I have an arm64 machine with your PR and see if this appears or not!

I increased test duration for 2 times (80m), it doesn't help... If you have time to debug this problem on ARM machine, it will be great. I have no ARM machine, unfortunaty, but I'll try to find an ARM VM somewhere).

anfedotoff:
I've tested the test that I've added on ARM VM it works:

```
sudo go test -run TestKprobeUser
time="2024-03-28T16:13:29Z" level=info msg="BTF discovery: default kernel btf file found" btf-file=/sys/kernel/btf/vmlinux
time="2024-03-28T16:13:34Z" level=info msg="Exit probe on acct_process"
PASS
time="2024-03-28T16:13:53Z" level=info msg="map dir `/sys/fs/bpf/testSensorTracing` still exists after test. Removing it."
ok      github.com/cilium/tetragon/pkg/sensors/tracing  25.110s
```

I use qemu without kvm. Maybe the problem is because I added one more test?

anfedotoff:
@mtardy, maybe let's test only `./pkg/sensors/tracing...` for debug, what do you think?

mtardy:
I converted to draft since it seems your added test is the issue on arm64, it seemed you could not reproduce this on arm64 machine, I'll try as well on my side and see if I can get any idea.

anfedotoff:
> I converted to draft since it seems your added test is the issue on arm64, it seemed you could not reproduce this on arm64 machine, I'll try as well on my side and see if I can get any idea.
> 
> It was failing on actuated machines:
> 
> ```
> Linux c99d86049feb9df7b83f0f00632bb69a9d2e8edf 5.10.201 #1 SMP Thu Feb 15 11:00:52 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
> ```
> 
> We can ssh into them if really needed if we don't find any way to reproduce this since it seems to reliably fail there.

My ARM machine:

```
Linux aarch64-test 5.15.0-101-generic #111-Ubuntu SMP Wed Mar 6 18:01:01 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux
```

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -6142,6 +6142,60 @@ spec:
 	err = jsonchecker.JsonTestCheck(t, checker)
 	assert.NoError(t, err)
 }
+func TestKprobeUserStackTrace(t *testing.T) {
```
mtardy, line 6145: so you already added the test in your first commit and then you modified it in your "add the test commit", could you cleanup this and properly add this new test in a separate one?

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -6142,6 +6142,60 @@ spec:
...
+	readyWG.Wait()
+
+	if err := exec.Command(testGetCpu).Run(); err != nil {
+		t.Fatalf("failed to run %s: %s", testGetCpu, err)
+	}
+
+	stackTraceChecker := ec.NewProcessKprobeChecker("user-stack-trace").
+		WithProcess(ec.NewProcessChecker().WithBinary(sm.Full(testGetCpu))).
+		WithUserStackTrace(ec.NewStackTraceEntryListMatcher().WithValues(
+			ec.NewStackTraceEntryChecker().WithSymbol(sm.Suffix(("main.main"))),
```
mtardy, line 6187: especially this version of the test will not pass 5.10. that would be better that each commit are separated logically and that they compile/pass the tests.

pkg/procsyms/procsyms.go
```diff
@@ -0,0 +1,134 @@
...
+	Module string
+	Offset uint64
+}
+
+// ToString returns a string representation of FnSym
+func (fsym *FnSym) ToString() string {
+	return fmt.Sprintf("%s (%s+0x%x)", fsym.Name, fsym.Module, fsym.Offset)
+}
+
+// GetFnSymbol -- returns the FnSym for a given address and PID
```
florianl, line 36: It should be documented that this function is limited to native code only (C, C++, Go and Rust). The given approach will not work for interpreted languages (Java, Python, Perl, etc), as it can not resolve the internal state of the interpreter.

PR #2177, opened 2024-03-03
tetragon: Add cgroup rate support

olsajiri:
Adding support to monitor cgroup rate for exec, clone and exit events. The rate is configurable.


bpf/process/bpf_rate.h
```diff
@@ -31,19 +33,30 @@ struct {
 
 static inline __attribute__((always_inline)) bool
 cgroup_rate(struct cgroup_rate_key *key, __u64 ns,
-	    const struct cgroup_rate_settings *settings)
+	    const struct cgroup_rate_settings *settings,
+	    bool *throttle)
```
kkourt, line 64: So if I understand correctly, we:
* return `true` if the rate is within our settings
* return `false` if the rate is not within our settings
     * we set throttle to true, if a throttle message is needed?

kkourt (commented):
Cool! left some first comments :)

bpf/process/bpf_rate.h
```diff
@@ -0,0 +1,61 @@
...
+		new_value.lastns = ns;
+		new_value.tokens = settings->tokens - 1;
+		map_update_elem(&cgroup_rate_map, key, &new_value, BPF_ANY);
+		return true;
+	}
+
+	since_lastns = ns - value->lastns;
+	if (since_lastns > settings->interval_ns) {
+		value->tokens = settings->tokens;
+		value->lastns = ns;
```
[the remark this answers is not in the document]
olsajiri: hm right, but I think it needs to be global - per cgroup ... at least I'm not sure how the per-core would work

I have another way of doing this but that involves timers (since 5.15) but should be sync safe ... I'll try that one

bpf/process/bpf_rate.h
olsajiri: right

bpf/process/bpf_rate.h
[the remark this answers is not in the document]
kkourt: The idea would be that every core would have a map for each cgroup. Maybe I'm missing something, but it should be exactly the same code with the only difference being that the hashmap is pser-cpu.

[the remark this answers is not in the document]
olsajiri: but we need per cgroup rate.. this would spread it across cpus, I might be missing something ;-)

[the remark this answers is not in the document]
kkourt: Hmm... Maybe I'm missing something.

The current implementation seems to use a global rate across all CPUs of the system. So if rate is X, the global event rate of a given cgroup cannot exceed X.

I was thinking of the rates being per-core rather than global (system-wide). So if rate is X, this would mean that the event rate of a given cgroup cannot exceed X on any given CPU. Note that this bounds the system rate into X*<ncpus> system-wide.

On problem with the per-core idea, however, is the cleanup. Because when a cgroup is removed in one core, you need to remove the entries from all of them, which (I think) means that you would need to do the removal in user-space.

bpf/process/bpf_rate.h
```diff
@@ -0,0 +1,83 @@
...
+	};
+	struct cgroup_rate_value *value;
+	__u64 delta, interval;
+	__u32 zero = 0;
+
+	options = map_lookup_elem(&cgroup_rate_options_map, &zero);
+	if (!options)
+		return true;
+
+	interval = options->interval;
```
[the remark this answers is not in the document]
olsajiri: hm, so the logic ATM is that tetragon does:

```
rate := cgrouprate.NewCgroupRate(ctx, base.CgroupRateMap, &option.Config.CgroupRate)
if rate != nil {
  cgrouprate.Config(base.CgroupRateOptionsMap)
  rate.AddListener(pm)
}
```
and we should get `rate == nil` if the option is disabled, so we don't call `cgrouprate.Config`, but there's always an element in options map.. I see, I guess we need extra check in bpf code:

```
diff --git a/bpf/process/bpf_rate.h b/bpf/process/bpf_rate.h
index af19204a6b2c..041773a032ee 100644
--- a/bpf/process/bpf_rate.h
+++ b/bpf/process/bpf_rate.h
@@ -52,6 +52,8 @@ cgroup_rate(__u64 cgroupid, __u64 time)
                return true;
 
        interval = options->interval;
+       if (!interval)
+               return true;
 
        value = map_lookup_elem(&cgroup_rate_map, &key);
        if (!value) {

```

bpf/process/bpf_rate.h
```diff
@@ -0,0 +1,83 @@
...
+			.time = (time / interval) * interval,
+			.curr = 1,
+		};
+
+		map_update_elem(&cgroup_rate_map, &key, &new_value, 0);
+		return true;
+	}
+
+	delta = time - value->time;
+	if (delta > interval) {
```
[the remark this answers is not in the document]
olsajiri: I think we can rely on `time >= value->time`, because `value->time` is `time` value from previous update on the same cpu so it should be monotonic

pkg/cgrouprate/cgrouprate.go
```diff
@@ -23,24 +33,52 @@ type Listener interface {
...
+}
+
 type CgroupRate struct {
 	listeners map[Listener]struct{}
 	log       logrus.FieldLogger
+	ch        chan *cgroupRate
+	flag      map[processapi.CgroupRateKey]bool
+	flagLock  sync.Mutex
+	rates     []*cgroupRate
+	opts      *option.CgroupRate
```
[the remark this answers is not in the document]
olsajiri: ah right, I think I'll just move the config patch before the logic patch

pkg/cgrouprate/cgrouprate.go
```diff
@@ -65,10 +103,202 @@ func (r *CgroupRate) Notify(msg notify.Message) {
...
+func (r *CgroupRate) checkRate(rate *cgroupRate, last uint64) bool {
+	values := make([]processapi.CgroupRateValue, bpf.GetNumPossibleCPUs())
+
+	if r.hash.MapHandle == nil {
+		return true
+	}
+
+	hash := r.hash.MapHandle
+
+	if err := hash.Lookup(rate.key, &values); err != nil {
```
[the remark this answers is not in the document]
olsajiri: I used metrics instead of warnings, it's added later on

pkg/cgrouprate/cgrouprate.go
```diff
@@ -65,10 +103,202 @@ func (r *CgroupRate) Notify(msg notify.Message) {
...
+	//         Rate
+	//        |-----------------|
+	//                          |
+	//                          Time
+	//
+	// TimeInPrev = IntervalLen - (Time - IntervalStart)
+	// Partial    = TimeInPrev / IntervalLen
+	// Rate       = Prev x Partial + Curr
+
+	compute := func(v *processapi.CgroupRateValue) uint64 {
```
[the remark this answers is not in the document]
olsajiri: I was thinking on what should be the scope of the function tests and I ended up with  `TestCheckThrottle` that runs `checkRate` function which essentially checks the computed rate through posted throttle events, would you have in mind some other test?

pkg/cgrouprate/cgrouprate.go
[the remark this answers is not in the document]
kkourt: Ah, great!

bpf/process/bpf_rate.h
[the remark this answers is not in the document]
kkourt: Makes sense, thanks!

pkg/cgrouprate/cgrouprate.go
```diff
@@ -65,10 +103,202 @@ func (r *CgroupRate) Notify(msg notify.Message) {
...
+	// for more than 5 iterations (atm 5 seconds). The cgroup rate is
+	// then removed from the rates list and it's put back on when
+	// there's new event on it.
+	isAlive := func() bool {
+		if events == 0 {
+			// Wait for aliveCnt rate loops to make sure the
+			// cgroup is silent
+			if rate.alive == 0 {
+				r.delFlag(rate.key)
+				if err := hash.Delete(rate.key); err != nil {
```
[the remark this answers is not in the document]
olsajiri: yes, we can miss event that would update the `cgroup_rate_map` and we can falsely assume there's no traffic on the cgroup, but we care about excesive event traffic, so if there are more events then the next event will do the map update and we are back, IOW one event lost does not matter

pkg/option/flags.go
```diff
@@ -184,6 +188,50 @@ func ParseMetricsLabelFilter(labels string) map[string]interface{} {
 	return result
 }
 
+type CgroupRate struct {
+	Events   uint64
+	Interval uint64
+}
+
+func parseCgroupRate(rate string) CgroupRate {
```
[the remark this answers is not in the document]
olsajiri: ok, good idea

pkg/cgrouprate/cgrouprate.go
```diff
@@ -23,24 +33,52 @@ type Listener interface {
...
 }
 
-func newCgroupRate() *CgroupRate {
+func newCgroupRate(hash *program.Map,
+	opts *option.CgroupRate) *CgroupRate {
+
+	if opts.Events == 0 || opts.Interval == 0 {
+		logger.GetLogger().Infof("Cgroup rate disabled (opts %d/%d)",
+			opts.Events, opts.Interval)
+		return nil
```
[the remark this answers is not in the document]
olsajiri: yes, I think we need the ebpf side also to check the interval value > 0 as suggested in previous comment

api/v1/tetragon/events.proto
```diff
@@ -149,6 +150,19 @@ message RateLimitInfo {
     uint64 number_of_dropped_process_events = 1;
 }
 
+enum ThrottleType {
+    THROTTLE_UNKNOWN = 0;
+    THROTTLE_START = 1;
+    THROTTLE_STOP = 2;
+}
+
+message ProcessThrottle {
```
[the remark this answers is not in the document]
olsajiri: we could save the last pid that modified the cgroup_rate_map and 'sell it' as the culprit in the throttle events, I'll check on that

gonchik:
@olsajiri do you have plan to finalize it ? 
I will be happy to see that PR in the release :)

olsajiri:
> @olsajiri do you have plan to finalize it ? I will be happy to see that PR in the release :)

yep, yet another version just landed waiting for comments ;-)

bpf/process/bpf_rate.h
```diff
@@ -0,0 +1,119 @@
...
+	 *                       |
+	 *    val->prev          | val->curr
+	 *   |-------------------|-----------
+	 *         val->rate
+	 *        |-------------------|
+	 *                            time
+	 */
+
+	slide = interval - (time - val->time);
+	val->rate = (slide * val->prev) / interval + val->curr;
```
kevsecurity, line 145: This calculates the rate of events per window but it assumes that the events in the prev window are evenly distributed. I think a better calculation would be:
val->rate = (val->prev + val->curr) / ((time - val->time) + interval)
[ extra brackets added for clarity ].
This calculation sums the previous window and the partial current window and then divides by the amount of time that passed since the beginning of the previous window. It does mean that the size of window that events are averaged over changes depending on the time and varies between a fraction over 1 interval, up to 2 intervals, and averages 1.5 intervals. Therefore if you set opt->interval to 2/3 the interval specified, then the window that events are averaged over will vary between 2/3 of the window specified and 4/3 of the window specified, averaging on exactly the window specified.
This approach doesn't care whether the events are evenly distributed or not.

bpf/process/bpf_rate.h
```diff
@@ -0,0 +1,119 @@
...
+
+	if (!val->throttled && val->rate >= opt->events)
+		val->throttled = time;
+
```
kevsecurity, line 151: Maybe the throttle start time should be updated every time there is an event received while throttling? So could change the above to:
if (val->rate >= opt->events)
    val->throttled = time;

Also, once val->throttled has a time how does it get reset to 0? Surely it would be easy to check here
if ((time - val->throttled > throttle_duration) && (val->rate < opt->events))
    val->throttled = 0;
Or if you take the suggestion to reset the throttle start time each event (as val->throttled will be updated):
if (time - val->throttled > throttle_duration)
    val->throttled = 0;

bpf/process/bpf_rate.h
```diff
@@ -0,0 +1,119 @@
...
+		return true;
+
+	val = map_lookup_elem(&cgroup_rate_map, &key);
+	if (!val) {
+		struct cgroup_rate_value new_value = {
+			.time = (time / interval) * interval,
+			.curr = 1,
+		};
+
+		map_update_elem(&cgroup_rate_map, &key, &new_value, 0);
```
kevsecurity, line 103: Why is this a per-CPU hash? Is the throttling only per CPU?

pkg/cgrouprate/cgrouprate.go
```diff
@@ -65,10 +101,137 @@ func (r *CgroupRate) Notify(msg notify.Message) {
...
+			r.updateCgroups(cq)
+		case <-ticker.C:
+			r.processCgroups()
+		}
+	}
+}
+
+func (r *CgroupRate) updateCgroups(cq *cgroupQueue) {
+	if _, ok := r.cgroups[cq.id]; ok {
+		// the group is garantued to be checked on next timer
```
kevsecurity, line 132: nit: typo

bpf/process/bpf_rate.h
kevsecurity: Okay I guess you want to avoid the race condition that exists even with atomic increments. That makes sense. From a personal perspective, I would have preferred the user space ticker patch to be included in this patch as that answers some of my questions.

bpf/process/bpf_rate.h
kevsecurity: I see from a later patch that you are summing the rates for the cores, so some of the above can be ignored. That said, if val->throttled is not zero, why not set val->throttled to the current time and reset the throttle duration?

if (val->throttled || val->rate >= opt->events)
val->throttled = time;

pkg/option/flags.go
```diff
@@ -295,4 +343,5 @@ func AddFlags(flags *pflag.FlagSet) {
 
 	flags.Bool(KeyGenerateDocs, false, "Generate documentation in YAML format to stdout")
 
+	flags.String(KeyCgroupRate, "", "Base sensor events cgroup rate (events,interval)")
```
kkourt, line 346: I guess empty ("") means that no cgroup rate is applied? If so, let's document this in the option.

bpf/process/bpf_rate.h
olsajiri: hum, you're right the current approach does not care whether events are evenly distributed,
but you suggestion is also approximation as you noted above.. I'll need to think some more on which one is worse/better ;-) at the moment both seem similar to me

pkg/cgrouprate/cgrouprate.go
```diff
@@ -0,0 +1,74 @@
...
+
+var (
+	handle     *CgroupRate
+	handleLock sync.RWMutex
+)
+
+type Listener interface {
+	Notify(msg notify.Message) error
+	io.Closer
+}
```
kkourt, line 37: Is there a reason for defining a new type and not using oveserver.Listener https://github.com/cilium/tetragon/blob/25a0050cda943d8744b3f846d8c4a82219bce2da/pkg/observer/listener.go#L17?

pkg/cgrouprate/cgrouprate.go
```diff
@@ -0,0 +1,74 @@
...
+	defer handleLock.Unlock()
+
+	handle = newCgroupRate()
+	go handle.process(ctx)
+	return handle
+}
+
+func (r *CgroupRate) AddListener(listener Listener) {
+	r.listeners[listener] = struct{}{}
+}
```
kkourt, line 94: We seem to be using the same pattern as the observer, which is:

```
	rate := cgrouprate.NewCgroupRate(ctx, base.CgroupRateMap, &option.Config.CgroupRate)
	if rate != nil {
		cgrouprate.Config(base.CgroupRateOptionsMap)
		rate.AddListener(pm)
	}
```

This seems too complicated to me. Is there a reason for why we cannot have a single listener:

```
type CgroupRate struct {
	listener observer.Listener
	log      logrus.FieldLogger
	ch       chan *cgroupQueue
	opts     *option.CgroupRate
	hash     *program.Map
	cgroups  map[uint64]string
}
```

And do a single call where we initialize everything? I don't see any other use of AddListener but maybe I missed it?

bpf/process/bpf_rate.h
olsajiri: ugh, sorry about that, I should have included some docs explaining the whole logistics,
there were so many versions of this and I forgot to do that, there's some explanation in https://github.com/cilium/tetragon/pull/2177/commits/363cb62a30710b6f3a925249fbd86ef554df29d0 changelog

basically we compute/monitor the rate per cpu, because I did not find a good way to make atomically,
when cpu rate crosses the limit we send event to user space that triggers throttle start event and triggers user space to periodically check on that cgroup rate, when the rate goes down or the cpu got idle (and all other cpus are ok for the cgroup) we send throttle stop event

we monitor the cgroups per cpu but throttle events are sent for the whole cgroup system wide to prevent getting
many throttle start/stop events from all cpus

having throttle start event means that one or more cpus that execute cgroup code crossed the limit and are throttled

when throttle stop is send that means all previously throttled cpus are now below allowed limit rate

I'll add some documentation

bpf/process/bpf_rate.h
olsajiri: the logic of the user space code needs to know the time the throttling happened on the cpu, so we record the time just once

kkourt (changes requested):
Thanks! Everything looks good to me.

I'm "Requesting Changes" because I would like to simplify the cgroup rate code as I mention in my comment. It might not seem like a big deal, but as the project grows I believe it's important to keep our code as simple as possible.

api/v1/tetragon/events.proto
```diff
@@ -151,6 +152,19 @@ message RateLimitInfo {
...
```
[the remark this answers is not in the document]
olsajiri: yes, we can add that, I recall we discussed that for previous version that we can live without it, but the new version is more straightforward, so we can easily add it

pkg/option/flags.go
olsajiri: yes, will add the doc

pkg/cgrouprate/cgrouprate.go
olsajiri: hm I think I did not want the observer dependency, I will check

pkg/cgrouprate/cgrouprate.go
olsajiri: right, there's just one listener, so that can be simplified, will do that, thanks

bpf/process/bpf_rate.h
olsajiri: @kevsecurity I'm inclined in to keep the current computation for now, mainly due to:
  - both do approximation and are not accurate
  - I'm still not sure which one is more accurate, but we don't really need highly accurate rate, we aim to spot misbehaving cgroups that spike huge traffic that will cross the limit

I have already a follow up change to add task into the event, I'd try your way of rate computation as a follow up as well and see how it goes.. would that be ok?

kevsecurity: My suggestion is a better approximation because it doesn't assume an even distribution. Therefore I think it will always be more useful. Appreciate it might be unnecessary but I think it's worth a try, especially if you can compare the two approaches.

kkourt: A good property of the original approach is that it uses a precise count for the current period, so it can react better to spikes.

For example, let's say that our interval is 10, current time is 15, and that there were 10  events between 0 and 5, 10 events between 5 and 10, and 100 events between 10 and 15. The actual rate is 110, since there are 110 events between [5,15].

For the current approach is:
```
slide = 10 - (15-10) = 5
rate = 110
```

The suggested approach is:
```
(100 + 20) / (15 - 10 + 10) = 8
```

Which is 8*10 = 80 events on the interval.

Given the above, the first approach seems preferable to me.

kevsecurity: Okay, there are a few things that jump out.
1. The rates aren't directly comparable because (as explained in my comment) the window size needs adjusting to accommodate the different approach.
2. If you want to detect spikes smaller than the window, then you could make the window smaller and it would detect them just fine. By definition, it's the rate over a window size with all the events in the window being averaged over the length of the window. If you're interested in spikes smaller than this then a smaller window would be more appropriate as then the events will then be averaged over a shorter duration, which is more likely to detect a spike.
3. If the events were front-loaded, e.g. 100 between 0-5 and 0 between 5-15 then the rate between 5-15 should be 0, but the original would calculate it as rate = (5 * 100) / 10 + 0 = 50. So if we're just choosing specific examples, then we can easily choose numbers that favour one approach over the other. The bigger question is which wins out over a large number of simulations. I'll write a Monte Carlo simulation to test this, with the deciding factor being which approach is closest to the actual rate in each simulation, reported as a proportion of the total runs.

kkourt: Not sure I follow all the above. In my mind, it's not about how long the spike is but rather how fast you can detect it.

> So if we're just choosing specific examples, then we can easily choose numbers that favour one approach over the other

For this specific use-case, detecting spikes is more important than detecting sinks.

FWIW, I think that's an interesting discussion but I wouldn't want to block the PR on it. I would suggest we go with the existing approach and we can always change it if we find evidence that a better approach exists.

kevsecurity: I don't think there's evidence that one approach detects spikes quicker than the other, except for the singular example given. What we need to know is if that is representative or not. If the aim is to detect quicker, then make the window smaller and the threshold lower.

In the example I gave we would report a false spike when there is no spike. That's not about detecting dips, but about not detecting spikes when they aren't there (I assume we'd prefer to not throttle when we don't meet the requirements).

kkourt: > I don't think there's evidence that one approach detects spikes quicker than the other, except for the singular example given. What we need to know is if that is representative or not. If the aim is to detect quicker, then make the window smaller and the threshold lower.

I agree that it's not conclusive that one approach detects spikes quicker than the other. The example I used was just to show my intuition that this is the case because the rate of the current period is not smoothed. It would be interesting to see whether a counter example exists whether the spike is detected quicker with what you propose.

> In the example I gave we would report a false spike when there is no spike. That's not about detecting dips, but about not detecting spikes when they aren't there (I assume we'd prefer to not throttle when we don't meet the requirements).

This is an approximation so we can't be perfect no matter what we choose. For this use-case it would be preferable to throttle even if we don't reach the rate (false positive) than the other way round (true negative).

olsajiri: sry for delay, github shows me replies with delay again.. 

I agree it's worth to try, still I think current way  has enough precision for what we need,
so I suggest I try your suggestion as a follow up and compare

bpf/process/bpf_rate.h
olsajiri: it's the duty of the user space to monitor the rate once it's throttled, the initial time is there just to make sure we don't get flooded with throttling events in case there are just rate spikes

bpf/process/bpf_rate.h
kevsecurity: I ran a Monte Carlo simulation (I'll share code asynchronously so you can run it yourself and check the code) and my proposed way was more accurate in ~68% of cases.

pkg/option/flags.go
```diff
@@ -182,9 +186,54 @@ func ReadAndSetFlags() error {
...
+func ParseCgroupRate(rate string) CgroupRate {
+	empty := CgroupRate{}
+
+	if rate == "" {
+		return empty
+	}
+
+	s := strings.Split(rate, ",")
+	if len(s) != 2 {
+		logger.GetLogger().Warnf("failed to parse cgroup rate '%s'", rate)
```
kkourt, line 207: if we fail to parse the cgroup string, we return an empty cgroup. From what I remember, this means that the feature is disabled (?). If so, let's add this to the message. E.g.,:
> logger.GetLogger().Warnf("failed to parse cgroup rate '%s: disabling cgroup rate throttling'", rate)

pkg/encoder/encoder.go
```diff
@@ -250,6 +250,17 @@ func (p *CompactEncoder) EventToString(response *tetragon.GetEventsResponse) (st
 			status = p.Colorer.Red.Sprint(exit.Status)
 		}
 		return CapTrailorPrinter(fmt.Sprintf("%s %s %s %s", event, processInfo, args, status), caps), nil
+	case *tetragon.GetEventsResponse_ProcessThrottle:
+		throttle := response.GetProcessThrottle()
+		event := p.Colorer.Red.Sprintf("🧬 %-7s", "throttle")
```
kkourt, line 255: 🧬 is an interesting choice :)

bpf/process/bpf_rate.h
```diff
@@ -0,0 +1,119 @@
...
+} cgroup_rate_map SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, __u32);
+	__type(value, struct cgroup_rate_options);
+} cgroup_rate_options_map SEC(".maps");
+
+#define NS (1000ULL * 1000ULL * 1000UL)
```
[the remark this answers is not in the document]
olsajiri: ok

pkg/option/flags.go
olsajiri: so there's this message that we faild to parse the option and there's later another one saying that cgroup rate is disabled:

```
time="2024-04-29T07:48:30Z" level=warning msg="failed to parse cgroup rate '100,bad' : %!w(*errors.errorString=&{time: invalid duration \"bad\"})"
...
time="2024-04-29T07:48:14Z" level=info msg="Cgroup rate disabled (0/0s)"
```

and if the option is valid we display:

```
time="2024-04-29T07:49:53Z" level=info msg="Cgroup rate started (100/1s)"
```

kkourt: Sounds good, thanks!

Trung-DV:
I think we could config the merge commit from PR to always squash for a better look :D 
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github#squashing-your-merge-commits

image

kkourt:
@Trung-DV We intentionally split changes into multiple patches. This makes things easier to understand when looking at git history.

PR #2191, opened 2024-03-06
docs: add FAQ entry about kernel config and tetra probe

mtardy:


docs/content/en/docs/faq/_index.md
```diff
@@ -14,54 +14,73 @@ for up to date information. Not all Tetragon features work with older kernel
...
+
+```
+override_return: true
+buildid: true
+kprobe_multi: false
+fmodret: true
+fmodret_syscall: true
+signal: true
+large: true
+```
```
kkourt, line 83: How about adding the `kubectl` one-liner as well?

For example:

> You can also run this command directly from the tetragon container. For example:
> 
> ```shell
> kubectl run bpf-probe --image=quay.io/cilium/tetragon-ci:latest --privileged --restart=Never -it --rm --command -- tetra probe
> ```

kkourt (approved):

PR #2196, opened 2024-03-06
Tracing: add (minimal) net_device type

kevsecurity:
Packet-level kprobes often have a struct net_device as a parameter, which describes the network device over which the function acts. We currently cannot report or match on this type. This commit adds minimal support for struct net_device, treating it's name (dev->name) as a simple string that can be matched upon and reported. Simply specify the arg type as 'net_device' and assume it is the string of its name.

The commit adds some of the structure required to make a more complete version of the support, to make it easier for someone to add that in the future. For example, we add skeleton types in user space that could be expanded to handle more parameters than simply device name.

```release-note
Provide minimal support for struct net_device
```


api/v1/tetragon/tetragon.proto
```diff
@@ -404,6 +408,7 @@ message KprobeArgument {
 	string cap_permitted_arg = 24; // Capabilities that are currently permitted in hexadecimal format.
 	string cap_effective_arg = 25; // Capabilities that are actually used in hexadecimal format.
 	KprobeLinuxBinprm linux_binprm_arg = 26;
+    KprobeNetDev net_dev_arg = 27;
```
tixxdz, line 411: maybe indention here?

tixxdz (approved):

api/v1/tetragon/tetragon.proto
kevsecurity: Ah ```make format``` doesn't fix that automatically and because its a .proto file, VSCode doesn't really know what to do with it. I will fix and push.

PR #2206, opened 2024-03-10
tetragon:api: support kprobes object

tixxdz:


tixxdz:
> Is the intent of this feature to provide observability into kprobe loads by non-Tetragon users?

The feature is for our users! both load and unload of kprobes including bpf via kprobes.

tixxdz:
> Is the intent of this feature to provide observability into kprobe loads by non-Tetragon users?

ah got you: yes a catch all ;-)

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -6580,3 +6581,249 @@ spec:
 	err = jsonchecker.JsonTestCheck(t, checker)
 	assert.NoError(t, err)
 }
+
+func getArmKprobeSymb(kSymbols *ksyms.Ksyms) string {
+	if kSymbols.IsAvailable("arm_kprobe") == true {
```
mtardy, line 6586: a nit but you don't need an extra comparison here, as with all the `if` and `else if` with `bool == true`.
```suggestion
	if kSymbols.IsAvailable("arm_kprobe") {
```

mtardy (commented):
that's cool kprobes monitoring with kprobes

bpf/process/types/kprobes.h
```diff
@@ -0,0 +1,18 @@
...
+/* Copyright Authors of Tetragon */
+
+#ifndef __LINUX_KPROBES_H__
+#define __LINUX_KPROBES_H__
+
+#ifndef KSYM_NAME_LEN
+#define KSYM_NAME_LEN 128U
+#endif
+
+struct msg_kprobe {
```
dwindsor, line 11: We also may want to capture `flags` here, as it contains some useful [state info](https://github.com/torvalds/linux/blob/e8f897f4afef0031fe618a8e94127a0934896aba/include/linux/kprobes.h#L96):

```c
/* Kprobe status flags */
#define KPROBE_FLAG_GONE	1 /* breakpoint has already gone */
#define KPROBE_FLAG_DISABLED	2 /* probe is temporarily disabled */
#define KPROBE_FLAG_OPTIMIZED	4 /*
				   * probe is really optimized.
				   * NOTE:
				   * this flag is only for optimized_kprobe.
				   */
#define KPROBE_FLAG_FTRACE	8 /* probe is using ftrace */
#define KPROBE_FLAG_ON_FUNC_ENTRY	16 /* probe is on the function entry */
```

bpf/process/types/kprobes.h
```diff
@@ -0,0 +1,18 @@
...
+	u64 addr;
+	u32 offset;
+	u32 pad;
+	char symbol[KSYM_NAME_LEN];
```
olsajiri, line 15: is it necessary to send the symbol? I'd think that the ksyms.KernelSymbols/kernelSymbols.GetFnOffset on user side that you already use should be fast enough

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -6580,3 +6581,249 @@ spec:
...
+	tracingPolicy := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "kprobes-tracing"
+  annotations:
+    description: "Detects kprobes operations"
+spec:
+  kprobes:
+  - call: ` + arm_kprobe + `
```
olsajiri, line 6646: hum, so this should not work with kprobe multi,  I'd think you need to add option like:

```
options:
- name: "disable-kprobe-multi"
  value: "1"
```

olsajiri:
not too much docs ;-) if the objective is to monitor kprobes/bpf-progs I think we should hook fprobe as well for kprobe multi probes

tpapagian:
It has been a while without any updates on that PR. I am closing that for now and we can re-open if we want to revive that. Thanks!

PR #2213, opened 2024-03-11
Memory: Swap probe_read to kernel or user version

kevsecurity:
We should always use the probe_read_kernel or probe_read_user helpers over the probe_read helper (ditto for _str versions).

This commit changes all probe_read to either probe_read_kernel or probe_read_user (ditto for _str versions).


jrfastab:
In general this seems the right direction. Does BTF report if its a user ptr or not? The kernel annotates these so we could learn it from the BTF type I think. Might need some BTF extensions to do I'm not sure off-hand.

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -5783,7 +5783,7 @@ spec:
 
 func TestKprobeListSyscallDupsRange(t *testing.T) {
 	if !kernels.MinKernelVersion("5.3.0") {
-		t.Skip("TestCopyFd requires at least 5.3.0 version")
+		t.Skip("TestKprobeListSyscallDupsRange requires at least 5.3.0 version")
```
kkourt, line 5787: Can you please split these changes into another patch?

bpf/process/bpf_generic_retkprobe.c
```diff
@@ -121,19 +113,57 @@ BPF_KRETPROBE(generic_retkprobe_event, unsigned long ret)
...
+	e->common.size = size;
+	e->common.ktime = ktime_get_ns();
+
+	tail_call(ctx, &retkprobe_calls, TAIL_CALL_FILTER);
+	return 1;
+}
+
+__attribute__((section("kprobe/2"), used)) int
+BPF_KRETPROBE(generic_retkprobe_copy_arg)
+{
```
kkourt, line 126: why is this added?

kkourt (changes requested):
The general idea makes sense to me but there are lot of changes in this PR and it's difficult to review it. Can you please split it into multiple patches and add commit messages on why each change is needed?

I would also like to avoid introducing yet another variant of our object files if we can avoid it.

bpf/Makefile
```diff
@@ -73,6 +78,7 @@ endef
 # Generic build targets for each sub-dir
 
 $(eval $(call DEFINE_VARIANT,v53))
+$(eval $(call DEFINE_VARIANT,v54))
```
kkourt, line 81: Why are we adding yet another variant?

bpf/process/bpf_generic_retkprobe.c
kevsecurity: Because the code additions pushed the individual programs over the 4096 instruction limit and therefore wouldn't load on a 4.19 kernel. This change simply follows the existing patterns by using a tail call (or rather splitting a program into two with the first tail-calling the second) to get under the instruction count limit.

pkg/sensors/tracing/kprobe_test.go
kevsecurity: I was just fixing a copy and paste error. I didn't think it would be an issue, but I can remove these fixes.

kevsecurity: Changes moved to a separate commit.

bpf/process/bpf_generic_retkprobe.c
kkourt: OK, but this needs to be marked and documented in a commit message in the PR so that it can be reviewed properly.

The patch that this change in is has the following diffstat:
```
 bpf/cgroup/bpf_cgroup_events.h                |   2 +-
 bpf/include/api.h                             |   3 +
 bpf/lib/bpf_cgroup.h                          |  20 ++--
 bpf/lib/bpf_helpers.h                         |  11 +-
 bpf/lib/bpf_task.h                            |  27 ++---
 bpf/lib/generic.h                             |   1 +
 bpf/lib/process.h                             |   2 +-
 bpf/libbpf/bpf_core_read.h                    | 382 +++++++++++++++++++++++++++++++++++++++++++++++++++-----------
 bpf/libbpf/bpf_tracing.h                      |   6 +-
 bpf/process/bpf_execve_bprm_commit_creds.c    |   1 +
 bpf/process/bpf_execve_event.c                |  22 ++--
 bpf/process/bpf_exit.c                        |   1 +
 bpf/process/bpf_exit.h                        |   4 +-
 bpf/process/bpf_generic_retkprobe.c           |  63 ++++++++---
 bpf/process/bpf_generic_tracepoint.c          |  16 +--
 bpf/process/bpf_loader.c                      |   7 +-
 bpf/process/bpf_process_event.h               | 140 +++++++++++------------
 bpf/process/data_event.h                      |  55 +++++----
 bpf/process/generic_calls.h                   |   9 +-
 bpf/process/retprobe_map.h                    |  11 +-
 bpf/process/types/basic.h                     | 327 ++++++++++++++++++++++++++++-------------------------
 bpf/process/types/probe_read_kernel_or_user.h | 150 ++++++++++++++++++++++++
 bpf/process/types/skb.h                       |  65 +++++------
 bpf/process/types/sock.h                      |  35 +++---
 pkg/api/tracingapi/client_kprobe.go           |   2 +
 pkg/sensors/tracing/args.go                   |   9 +-
 pkg/sensors/tracing/generickprobe.go          |  12 +-
 pkg/sensors/tracing/generictracepoint.go      |   5 +-
 pkg/sensors/tracing/genericuprobe.go          |   2 +-
 pkg/sensors/tracing/kprobe_test.go            |  23 ++--
 pkg/sensors/tracing/tracepoint_amd64_test.go  |   8 +-
 pkg/sensors/tracing/tracepoint_test.go        |   4 +-
 32 files changed, 956 insertions(+), 469 deletions(-)

```

Which makes it difficult to review. Please split the changes into smaller independent patches so that the changes needed for this are clearly marked and can be reviewed one-by-one.

pkg/sensors/tracing/tracepoint_test.go
```diff
@@ -824,11 +824,11 @@ spec:
         values:
         - ` + pidStr + `
       matchArgs:
-      - index: 0
+      - index: 4
         operator: "InMap"
         values:
         - "list:test"
-      - index: 1
+      - index: 5
```
kkourt, line 831: This change seems to indicate that the semantics of the `tracepoints:` changes which breaks existing policies.

kkourt (changes requested):

pkg/sensors/tracing/tracepoint_test.go
kevsecurity: This doesn't actually break existing policies, but the way it was originally written was wrong IMO. In other tracepoints (not raw_syscalls) the arg index in the selectors matches the arg index in the args section (and there is code here: https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/generictracepoint.go#L535 that rewrites the selector arg index (e.g. 5) to the list index of the arg that it applies to (e.g. 1). But it appears that you can either do that, or can use a 0-based index into the list of args in the args section directly (which is what the code is doing), adn this isn't documented AFAIK.

This change in the test makes no functional change to the code – so what worked before still works  – but it exercises the referenced code that rewrites the selector index number.

IMO there might be a bug here. If an args section references args with indices 1 and 2 (so ignoring the first argument to the function entirely), and the match arg selector specifies index 1, which argument does it apply to? The code will rewrite it to index 0 having matched it against the arg with index 1 (the first in the 0-indexed list), but the implication of the test (before the change) was that an arg selector with index 1 is specifying the second arg (0-indexed list of args), and therefore the arg with index 2.

I haven't addressed this directly in this PR because I think it needs discussion and a PR of its own. I did, however, modify the test so that it matches what the code expects. The test only works unmodified because an arg with index 1 doesn't exist in the args section (otherwise the index would have been rewritten).

kevsecurity: I will however break this change out into its own commit as I break up the other changes.

bpf/process/bpf_generic_retkprobe.c
kevsecurity: You returned to the review a bit too soon 😢 . I had pushed some changes but hadn't yet divided up the commit into many commits.

kkourt: Ah, apologies for that.
Saw the push and assumed the PR was ready.

bpf/process/types/basic.h
```diff
@@ -652,6 +653,12 @@ has_max_data(unsigned long argm)
 	return (argm & ARGM_MAX_DATA) != 0;
 }
 
+static inline __attribute__((always_inline)) bool
+is_userspace_data(unsigned long argm)
+{
+	return (argm & ARGM_USERSPACE_DATA) != 0;
```
kkourt, line 676: nit, just use `argm & ARGM_USERSPACE_DATA`. No need for `!= 0`

pkg/sensors/tracing/args.go
```diff
@@ -41,37 +43,39 @@ func argReturnCopy(meta int) bool {
...
+//	16-31 : size for const_buf
+func getMetaValue(arg *v1alpha1.KProbeArg, userspaceDataDefault bool) (int, error) {
+	meta := 0
 
 	if arg.SizeArgIndex > 0 {
 		if arg.SizeArgIndex > 15 {
 			return 0, fmt.Errorf("invalid SizeArgIndex value (>15): %v", arg.SizeArgIndex)
 		}
-		meta = int(arg.SizeArgIndex)
+		meta = meta | int(arg.SizeArgIndex)
```
kkourt, line 61: nit: `meta |= ...`

pkg/sensors/tracing/args.go
```diff
@@ -41,37 +43,39 @@ func argReturnCopy(meta int) bool {
...
-func getTracepointMetaValue(arg *v1alpha1.KProbeArg) int {
-	if arg.SizeArgIndex > 0 {
-		return int(arg.SizeArgIndex)
-	}
-	if arg.ReturnCopy {
-		return -1
+	if arg.IsUserspaceData == nil {
+		// If not set in policy, use the default.
+		if userspaceDataDefault {
+			meta = meta | argUserspaceDataBit
```
kkourt, line 72: nit: `meta |= ...`

pkg/sensors/tracing/args.go
```diff
@@ -41,37 +43,39 @@ func argReturnCopy(meta int) bool {
...
+		}
+	} else {
+		// Otherwise, use the provided value.
+		if *arg.IsUserspaceData {
+			meta = meta | argUserspaceDataBit
```
kkourt, line 77: nit: `meta |= ...`

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -222,10 +222,10 @@ func (out *genericTracepointArg) getGenericTypeId() (int, error) {
 		if err != nil {
 			return gt.GenericInvalidType, fmt.Errorf("failed to get size of array type %w", err)
 		}
-		if out.MetaArg == 0 {
-			// set MetaArg equal to the number of bytes we need to copy
-			out.MetaArg = nbytes
-		}
+		// set MetaArg's upper half-word equal to the number of bytes we need to copy
+		out.MetaArg = out.MetaArg & 0xffff
+		out.MetaArg = out.MetaArg | (nbytes << 16)
```
kkourt, line 227: Let's add a check and return an error if nbytes is cannot fit in the number of bits we have.

bpf/process/types/basic.h
```diff
@@ -2631,26 +2627,26 @@ read_call_arg(void *ctx, struct msg_generic_kprobe *e, int index, int type,
 	case size_type:
 	case s64_ty:
 	case u64_ty:
-		probe_read(args, sizeof(__u64), &arg);
+		*(__u64 *)args = arg;
```
kkourt, line 2654: This does not seem safe to me. We do not do nothing to ensure that the u64 values are properly aligned in the args array, and accessing unaligned values can cause issues. I think for x86_64 is fine, but I'm not sure about ARM. To be on the safe side, I'd rather we just use `probe_read` as we did before.

bpf/process/types/probe_read_kernel_or_user.h
```diff
@@ -0,0 +1,150 @@
...
+{
+	if (userspace) {
+		asm volatile("%[size] &= %1;\n"
+			     : [size] "+r"(size)
+			     : "i"(size_mask));
+		return probe_read_user(dst, size, src);
+	}
+	asm volatile("%[size] &= %1;\n"
+		     : [size] "+r"(size)
+		     : "i"(size_mask));
```
kkourt, line 32: Is there a reason for doing the assembly part twice? It's just a single instruction, but wouldn't it be better to do it once and then do the branch?

E.g.,
```
	asm volatile("%[size] &= %1;\n"
			     : [size] "+r"(size)
			     : "i"(size_mask));
	return userspace ? probe_read_user(dst, size, src) : probe_read_kernel(dst, size, src);
```

bpf/process/types/probe_read_kernel_or_user.h
```diff
@@ -0,0 +1,150 @@
...
+#define BPF_CORE_READ_KERNEL_OR_USER_STR_INTO(userspace, dst, src, a, ...)                                       \
+	({                                                                                                       \
+		typeof(dst) dst_x = dst;                                                                         \
+		typeof(src) src_x = src;                                                                         \
+		typeof(a) a_x = a;                                                                               \
+		if (userspace)                                                                                   \
+			___core_read(bpf_core_read_user_str, dst_x, src_x,                                       \
+				     a_x, ##__VA_ARGS__) else ___core_read(bpf_core_read_str, dst_x, src_x, a_x, \
+									   ##__VA_ARGS__)                        \
+	})
```
kkourt, line 116: What happens if `userspace` is false?

bpf/process/types/probe_read_kernel_or_user.h
```diff
@@ -0,0 +1,150 @@
...
+ * 4. return __t->g;
+ *
+ * Equivalence is logical, because there is a heavy type casting/preservation
+ * involved, as well as all the reads are happening through bpf_probe_read()
+ * calls using __builtin_preserve_access_index() to emit CO-RE relocations.
+ *
+ * N.B. Only up to 9 "field accessors" are supported, which should be more
+ * than enough for any practical purpose.
+ */
+#define BPF_CORE_READ_KERNEL_OR_USER(userspace, src, a, ...)              \
```
kkourt, line 140: Is this macro used somewhere?

kkourt (changes requested):
Thanks for splitting the PR. FWIW, I still find it hard to navigate, but it's definitely better.
I'm adding some first comments. I did not go over the full PR and plan to do another pass soon.

bpf/process/types/basic.h
kevsecurity: Unaligned is safe on ARM. If it isn't then this should fail tests on ARM. The changes I made are more expressive of what arg actually refers to, and avoids a call to a helper.

bpf/process/types/probe_read_kernel_or_user.h
kevsecurity: Yep, because otherwise the registers spill during the branch and then the masking is forgotten (on older kernels) when the registers are restored from the stack.

bpf/process/types/probe_read_kernel_or_user.h
kevsecurity: There is an else clause that calls bpf_core_read_str(). I guess 'make format' reorganised this so that it was less obvious.

bpf/process/types/probe_read_kernel_or_user.h
kevsecurity: This is to support cases where people would otherwise use BPF_CORE_READ() or BPF_CORE_READ_USER() but where they only know programmatically which one they want. The idea of adding them at this point was so that they are consistent with the existing macros and encourage people to use them over alternatives.

bpf/process/types/basic.h
kkourt: > Unaligned is safe on ARM. If it isn't then this should fail tests on ARM. 

Can you provide a reference for that? The fact that the tests are not failing is not an indication that unaligned access is OK.

kevsecurity: https://medium.com/@iLevex/the-curious-case-of-unaligned-access-on-arm-5dd0ebe24965 claims support.
Really, we should just add unaligned memory tests. Also, are we trusting probe_read to do the unaligned access, so we should probably consider trusting the compiler to do the same.
I can add unaligned memory tests to the patch.

kevsecurity: https://stackoverflow.com/questions/38535738/does-aarch64-support-unaligned-access is possibly a more useful reference.

kkourt: Yap, seems to be the case in https://developer.arm.com/documentation/102376/0100/Alignment-and-endianness as well:
> When the address is not a multiple of the element size, the access is unaligned. Unaligned accesses are allowed to addresses marked as Normal, but not to Device regions. An unaligned access to a Device region will trigger an exception (alignment fault).

kkourt: OK, sounds good, let's keep it!

kkourt (changes requested):
Did another pass of the patches I had not reviewed before. Thanks!

pkg/sensors/tracing/args.go
```diff
@@ -34,6 +34,7 @@ const (
 	argReturnCopyBit    = 1 << 4
 	argMaxDataBit       = 1 << 5
 	argUserspaceDataBit = 1 << 6
+	argRawSyscallsBit   = 1 << 7
```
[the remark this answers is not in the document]
kevsecurity: The idea was to prepare for being able to filter on arguments in raw_syscalls, and therefore arrays of arguments which could potentially be pointers. I thought it was a bit much to include that in this PR, but I was already thinking ahead for a following PR. Can do as you suggest if you prefer though.

bpf/lib/generic.h
```diff
@@ -49,6 +49,7 @@ struct msg_generic_kprobe {
 	/* anything above is shared with the userspace so it should match structs MsgGenericKprobe and MsgGenericTracepoint in Go */
 	char args[24000];
 	unsigned long a0, a1, a2, a3, a4;
+	unsigned long ret;
```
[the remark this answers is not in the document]
kevsecurity: It's used in the next patch. I can move it to that patch if you prefer.

pkg/sensors/tracing/args.go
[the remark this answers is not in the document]
kkourt: > Can do as you suggest if you prefer though.

Thanks! I would prefer to keep the code as simple as possible, even if we end up extending this for filtering.

I've also thought about filtering `raw_syscalls`.

One way would be to do as you suggest, and use the arguments but this seems tricky because we need to pass types for everything.

I think a better approach would be to do two-level filtering where we in  `raw_syscall` event we  to filter the type of the syscall and then we enable hooks on syscalls/ for filtering on the arguments. This seems simpler to me and it re-uses the arg filtering we already have. We would need to have a way to pass information from the `raw_syscall` hook to the specific syscall hook, but we already have something similar Jiri's notify enforcer action.

In any case, that's for another PR.

bpf/lib/generic.h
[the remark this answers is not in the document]
kkourt: It's a minor thing, but I like clean history so if you don't mind moving it I'd prefer it.

PR #2225, opened 2024-03-15
e2e tests minor patches

kkourt:


Makefile
```diff
@@ -122,6 +122,10 @@ help:
 	@echo '    Documentation:'
 	@echo '        docs              - preview documentation website'
 	@echo '        metrics-docs      - generate metrics reference documentation page'
+	@echo 'End-to-end tests: '
+	@echo '    e2e-test                                        - run e2e tests'
+	@echo '    e2e-test E2E_BUILD_IMAGES=0                     - run e2e tests without (re-)building images'
+	@echo '    e2e-test E2E_TESTS=./tests/e2e/tests/skeleton   - run a specific e2e test'
```
lambdanis, line 128: I was just thinking we should do something like https://www.thapaliya.com/en/writings/well-documented-makefiles/, because many targets are missing from this help. But I'm not sure how to best document multiple options for one target then. (I mean, it's very helpful to have options documented, I'm just thinking out loud here)

tests/e2e/flags/flags.go
```diff
@@ -26,7 +26,7 @@ var Opts = Flags{
 	},
 	KeepExportData: false,
 	InstallCilium:  true,
-	CiliumVersion:  "v1.12.6",
+	CiliumVersion:  "v1.15.1",
```
lambdanis, line 29: #renovateit We can at least add a note to [Renovate's draft PRs](https://github.com/cilium/tetragon/blob/31117d0596337b6219360ae813ae3c9393991a52/.github/renovate.json5#L393) to make sure it's updated I guess?

mtardy (approved):
thanks! Agree with anna, we could renovate it but it's already cool we bump this like Manually.

kevsecurity (approved):
LGTM

Makefile
kkourt: I guess we could something like
```
e2e-tests: ## run e2e tests, use E2E_BUILD_IMAGES=0 to avoid running images and E2E_TESTS=./tests/e2e/tests/skeleton ro run only the skeleton test
```

kkourt:
Thanks!

@lambdanis I think your comments make sense, but I'll merge this since I think it's an improvement and we can follow up with additional enhancements.

PR #2229, opened 2024-03-15
e2e: check policy state

kkourt:
When we check whether a policy is there, having it appear in the list is not enough. We need to check that its state is enabled and that there are no errors.


tpapagian (approved):

mtardy:
nice use of that new state thingie!

Issue #1879, opened 2023-12-13
Apply Policy Only to Specific Containers in a Pod


BonySmoke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Currently, we can apply a policy to a specific set of pods using the `podSelector` block.
We want to apply a policy only to certain container(s) inside a pod/pods.

In our use case, we have a main and a sidecar container inside a pod. We want to block write requests to files under a particular directory in the main container - let's say `/home/user`.
The problem is that the sidecar container may write to files in the same path as the main container and we want to allow it.

We tried to understand if we can differentiate containers based on binaries that perform write requests or PIDs. Unfortunately, either container can use the same binary and we couldn't understand how to differentiate containers based on the PID.

Please let us know if we are missing something.

### Describe the feature you would like

It would be great to be able to apply a policy only to specific containers inside a pod/pods.
As I can see, the information about the container is already present in JSON events.

### Describe your proposed solution

First of all, I want to say that I'm not familiar with the source code well enough to propose exact solutions, but I was thinking about the following 2 approaches:

1. Add a separate `containerSelector` block. It can be used independently to apply the policy to all containers in the cluster with specified names, as well as with `podSelector` limiting containers to a specific pod/pods.
```yaml
spec:
  containerSelector:
    operator: In
    values:
    - main
    - secondary
  # If podSelector is present, the policy applies only to containers specified above in this pod
  podSelector:
    matchLabels:
      app: "my-awesome-app"
```
2. If the approach above is not optimal or not feasible, maybe it's possible to add a `matchContainers` selector (similar to `matchBinaries`, etc.)?
```yaml
- matchContainers:
  - operator: "In"
    values:
    - "main"
    - "secondary"
```

Please let me know if none of the proposed solutions make sense and thank you in advance! :smiley: 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mtardy:
Thanks for the detailed request. I think the first proposition would make sense as the mechanism to filter things by namespace/labels could be adapted to containers in pods.

I'll let @kkourt take a look at this as he implemented the namespace/label filtering.

kkourt:
Thanks! I think that's a reasonable feature to have.

Not sure what the proper interface for the policy would be. Would need to think a bit more about it.

In terms of implementation, it's certainly something doable. We already track containers for pod-label filters so we would just need to do an additional filtering stage.

See:

[...]

PR #2231, opened 2024-03-17
policyfilter: Apply Policy Only to Specific Containers in a Pod

BonySmoke:
This PR proposes a solution for this feature request: https://github.com/cilium/tetragon/issues/1879

In general, it performs the following changes:
- Add `containerSelector` to the tracing policy CRD. This section will accept the same arguments as `podSelector` and reuse the same label filtering logic. We will construct a map of container fields with the corresponding values and they will be the same as labels. For now, the only field we include in this map is `name` but this can be extended.
- Add e2e and unit tests to cover the new changes.
- Add documentation to describe how this feature can be used.
- Minor changes to the `Makefile` and code for spinning up the cluster. Maybe this should be performed in a different PR, but I bumped the default image version for e2e tests to `v1.29.2` because I couldn't spin up a cluster with the current image version and kind `v0.22.0`.

This PR also includes files generated with `make crds`.
Documentation has been tested with `make docs`.

However, while running e2e tests and unit tests I faced a couple of issues (likely not related to these changes):
- e2e tests couldn't always create a cluster. For example, for `tests/e2e/tests/labels`, it looks like the repository is not available and returns 403. I didn't investigate these issues but targeted only `tests/e2e/tests/policyfilter` since this is what I changed.
```
I0317 13:32:30.031871  189507 helm.go:241] "Running Helm Operation" command="helm repo  add isovalent https://helm.isovalent.com --kubeconfig /tmp/kind-cluser-tetragon-ci-1988-kubecfg1161878701"
```
- unit tests were taking too long in my case. However, I tested all the changes in `pkg/policyfilter` and `pkg/sensors`.

```release-note
policyfilter: add a containerSelector that allows filtering policies by container name
```


lambdanis (changes requested):
Thank you for the PR @BonySmoke. I didn't only a quick pass, but looks good overall.

The CI failures pointed out some changes required:
* When updating CRDs, the version [here](https://github.com/cilium/tetragon/blob/2f58c9630d18a58950a26ba2fd01ad4afb805639/pkg/k8s/apis/cilium.io/v1alpha1/version.go#L10) has to be bumped too
* It seems `(*podInfo).cgroupIDs` is unused

Two other failures are known CI issues: #2010 and #2233, so ignore them.

BonySmoke:
@lambdanis thank you for this quick review!
I would like to clarify something before pushing the changes:
1. Can I bump the version of CRDs to `1.2.0`? I see previously people were bumping mostly the `patch` version even when it looked like it wasn't a bug fix.
2. You are totally right, after my changes, it's no longer used; however, I don't want to delete this code because it can be useful in the future:)
Do you think I need to delete this function or maybe I need to somehow tell the linter to ignore it?

lambdanis:
> 1. Can I bump the version of CRDs to `1.2.0`? I see previously people were bumping mostly the `patch` version even when it looked like it wasn't a bug fix.

So the idea is that the CRD minor version matches the Tetragon minor version. So yeah, it looks a bit weird that only the patch version is bumped on new features - but these version bumps are mostly for development/testing purposes. Users will update CRDs when upgrading Tetragon, so will always upgrade to the next minor version anyway.

> 2. You are totally right, after my changes, it's no longer used; however, I don't want to delete this code because it can be useful in the future:)
>    Do you think I need to delete this function or maybe I need to somehow tell the linter to ignore it?

Do you have an example of what it can be useful for? In general we're trying to avoid unused code, because hypothesizing about what will be useful in the future often ends on hypothesizing, not actual usage. One option is to add the code in one commit and delete in another - then the code stays only in the git history for the future reference. If the code can be really useful in other packages, then you can just export it - linter won't complain about exported symbols. But again, we can't to predict what will be useful in the future, so I'd avoid doing that.

BonySmoke:
@lambdanis Hello! Thank you for clarifying these points!
1. I bumped the CRD version to `1.1.9`.
2. Your explanation makes perfect sense. I don't have a good example where this unused code could be useful; therefore, I deleted it.

Makefile
```diff
@@ -415,7 +415,7 @@ docs:
 kind:
 	./contrib/localdev/bootstrap-kind-cluster.sh
 
-.PHONY: kind-install-tetragon 
+.PHONY: kind-install-tetragon
```
kkourt, line 418: Can you please split the patch into multiple patches so that there is one patch per change?
Having one patch per each change in the 1. to 4. list would make it much easier to review and manage the patches.

pkg/policyfilter/k8s_test.go
```diff
@@ -302,10 +302,46 @@ func (ts *testState) podsCgroupIDs(t *testing.T, podNames ...string) []uint64 {
...
+		var p *testPod
+		for _, pod := range ts.pods {
+			if pod.name == podName {
+				p = &pod
+				break
+			}
+		}
+
+		if p == nil {
+			t.Fatalf("unknown pod name: %s", podName)
```
kkourt, line 319: Let's not have `Fatal` here, unless we really want to crash the agent which I don't think we do.

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -90,6 +90,10 @@ type TracingPolicySpec struct {
 	// PodSelector selects pods that this policy applies to
 	PodSelector *slimv1.LabelSelector `json:"podSelector,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// ContainerSelector selects containers that this policy applies to
+	ContainerSelector *slimv1.LabelSelector `json:"containerSelector,omitempty"`
```
kkourt, line 95: I think it's worth documenting how the label selector is used in the container case.
As far as I understand based on the code, we use the `name` as a label `key` to match the container's name.
Here seems like a good place to document how this works.

pkg/policyfilter/state.go
```diff
@@ -222,6 +226,25 @@ func (pol *policy) podInfoMatches(pod *podInfo) bool {
 	return pol.podMatches(pod.namespace, pod.labels)
 }
 
+func (pol *policy) containerMatches(container *containerInfo) bool {
+	containerFilterFields := labels.Labels{
+		"name": container.name,
+	}
+	return pol.containerSelector.Match(containerFilterFields)
```
kkourt, line 225: Maybe I'm missing something but isn't the case that `containerSelector` might be nil? What would happen then?

kkourt (commented):
Thanks!

Had a quick look and overall approach makes sense to me. Please find some comments below.

pkg/policyfilter/state.go
BonySmoke: You are right, it can be `nil` when we read it from the tracing policy.
However, [before](https://github.com/cilium/tetragon/pull/2231/commits/7aebeacc6e6ef7473b5b8e2423eea72f071892f7#diff-980e5be14cfb8e087d0d09a0f42b7219a9510ad91dd199221abdc98d1767bf34R428) adding it to a new `policy`, we construct the selector from label selector and in case the `containerSelector` is `nil`, this will be an empty selector `[]selectorOp{}`.
[Here](https://github.com/cilium/tetragon/blob/main/pkg/labels/labels.go#L71-L72)'s the corresponding code. Also, the [same logic](https://github.com/cilium/tetragon/blob/main/pkg/policyfilter/state.go#L398) is used for `podSelector`.

Makefile
BonySmoke: Sure! I think I can also include the fix for the CRD version in the commit where I update CRD files.
I will update you once done:)

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
BonySmoke: You are right, it's worth documenting it inside the actual CRD.
I added some documentation. Hopefully, it looks clear :smiley:

Makefile
BonySmoke: I did this. I needed to push with force to rewrite the history, but hopefully, not it looks more readable :pray:

kkourt: Thanks!

When you want to re-request a review, please notify the reviewers by clicking on the github button "Re-request review".

pkg/policyfilter/k8s_test.go
kkourt: Ah you are right!
I missed that it was a `t.` `fatal` in my hasty review :)

pkg/policyfilter/state.go
kkourt: Thanks!
Am I correct to assume that there is a test for the above path (i.e, being `nil` in the tracing policy and having an empty selector when we call this function)? I would think yes, since AFAIR we do have tests where the container selector is nil.

kkourt:
Thanks! started the CI!

kkourt:
The Check CRD version test seems to not properly working for external branches:
> fatal: bad revision 'origin/pr/BonySmoke/add-container-selector'

We need to fix this, but it shouldn't block this PR.

pkg/policyfilter/state.go
BonySmoke: Yes, you are right, I tried to test the behavior when `containerSelector` is `nil` in multiple places!
- For example, [here](https://github.com/cilium/tetragon/pull/2231/commits/4764f1bb530d10166fc16c20e582a49d71b1d848#) I test that the policy with the `nil` pod and container selectors matches all containers.
- I also added the nil `containerSelector` to existing tests to make sure that the policy with the `podSelector` but without `containerSelector` still produces the same results. For example, [here](https://github.com/cilium/tetragon/pull/2231/commits/4764f1bb530d10166fc16c20e582a49d71b1d848#diff-0dfcd52d313187c0d7956a105e745763c25de1309923b1f8db9d3450af33363fR413).
- Also, just in case, [here](https://github.com/cilium/tetragon/pull/2231/commits/4764f1bb530d10166fc16c20e582a49d71b1d848#diff-0dfcd52d313187c0d7956a105e745763c25de1309923b1f8db9d3450af33363fR497) I tested that the policy with the `containerSelector` but without `podSelector` also matches all specified containers.

I tried to test other scenarios too, for example, when pod labels/containers change and we re-add containers. Hopefully, I covered the majority of cases :pray:

Makefile
BonySmoke: Oh, that's great, thank you! I will use this button from now on:)

BonySmoke:
@kkourt Hello! Thank you for this update!
I would like to ask you something as a part of the review. I probably should've included it in the PR description but I didn't, sorry for that :disappointed: 
There is this `pkg/policyfilter/rthooks/rthooks.go` file that registers new containers. You added it in [this commit](https://github.com/cilium/tetragon/commit/da6c5c8855097b4cd27347241df67fb473b9062d).

> The purpose of the callback is to set-up the policyfilter state before the container starts.

While adding container filtering, I needed to get container information too (to extract the name). Therefore, after finding the pod, I added the [logic for finding the container](https://github.com/cilium/tetragon/pull/2231/commits/fc0f13cb6f5ee0257512f972ed3c59b9a491b508#diff-0d06ff6b079b83e5948096c3ddab5693cdb637f590b5d6361a4ef84f2b244d6bR98-R101).
I wanted to confirm with you that this is the correct approach. There was no existing test for this function; therefore, I tried to test different scenarios in the live cluster and everything looked fine but still, I want to confirm this is the correct approach (mainly because I don't understand the concept of rthooks :smiling_face_with_tear: ).

kkourt:
> @kkourt Hello! Thank you for this update! I would like to ask you something as a part of the review. I probably should've included it in the PR description but I didn't, sorry for that 😞 There is this `pkg/policyfilter/rthooks/rthooks.go` file that registers new containers. You added it in [this commit](https://github.com/cilium/tetragon/commit/da6c5c8855097b4cd27347241df67fb473b9062d).
> 
> > The purpose of the callback is to set-up the policyfilter state before the container starts.
> 
> While adding container filtering, I needed to get container information too (to extract the name). Therefore, after finding the pod, I added the [logic for finding the container](https://github.com/cilium/tetragon/pull/2231/commits/fc0f13cb6f5ee0257512f972ed3c59b9a491b508#diff-0d06ff6b079b83e5948096c3ddab5693cdb637f590b5d6361a4ef84f2b244d6bR98-R101). I wanted to confirm with you that this is the correct approach. There was no existing test for this function; therefore, I tried to test different scenarios in the live cluster and everything looked fine but still, I want to confirm this is the correct approach (mainly because I don't understand the concept of rthooks 🥲 ).

Thanks for the head's up!
I'll take a look.

In the meantime, you might want to have a look at https://tetragon.io/docs/concepts/tracing-policy/k8s-filtering/. You can also use cri-o (it's actually a bit easier): https://github.com/cilium/tetragon/blob/main/contrib/rthooks/tetragon-oci-hook/docs/demo.md.

pkg/policyfilter/rthooks/rthooks.go
```diff
@@ -92,15 +92,25 @@ func createContainerHook(_ context.Context, arg *rthooks.CreateContainerArg) err
...
 
+	var containerFound bool
+	var container *corev1.ContainerStatus
 	namespace := pod.ObjectMeta.Namespace
+	pod, container, containerFound = arg.Watcher.FindContainer(containerID)
+	if !containerFound {
+		log.WithError(err).Warnf("failed to find container information %s, aborting hook.", containerID)
+	}
+
+	containerName := container.Name
```
kkourt, line 103: Thanks for pointing me to this code.

I had a look. There are ways to get the container name from the hook itself, which would be my preferred approach. That being said:
 i) this goes beyond the scope of this PR
 ii) the container name can be found in annotations which are not standard, which means that the hook may fail to figure out the container.
 
My suggestion would be to pass an empty containerName ("") for now. And then in the code that does the filtering based on the container name, we try  `arg.Watcher.FindContainer(containerID)` if we don't have the container name. The benefit of that is we only call FindContainer if the we have a container selector?

In a subsequent PR, we can extend the hook to also pass the container name.

What do you think?

BonySmoke: I think I got the idea; however, I would like to understand how to implement it correctly.

If I pass an empty container name to [pfState.AddPodContainer](https://github.com/cilium/tetragon/pull/2231/files#diff-0d06ff6b079b83e5948096c3ddab5693cdb637f590b5d6361a4ef84f2b244d6bR113), how can I use `arg.Watcher.FindContainer(containerID)` from [containerMatches](https://github.com/cilium/tetragon/pull/2231/files#diff-980e5be14cfb8e087d0d09a0f42b7219a9510ad91dd199221abdc98d1767bf34R221) (where the filtering logic is happening)? If I understand correctly, I also need to pass the `arg` struct from [createContainerHook](https://github.com/cilium/tetragon/blob/main/pkg/policyfilter/rthooks/rthooks.go#L34) to [pfState.AddPodContainer](https://github.com/cilium/tetragon/pull/2231/files#diff-0d06ff6b079b83e5948096c3ddab5693cdb637f590b5d6361a4ef84f2b244d6bR113) and probably store it somewhere (in `containerInfo`?) to use it when filtering because it holds the hook information for this container?
Maybe, an alternative would be to check if the container name is empty during filtering and if it is, try to find the pod (`corev1.Pod`) and the container in it by container ID? However, this sounds pretty complex and inefficient.
Sorry if I overcomplicate it, I just don't have a clear picture in my head at the moment :disappointed: 

Also, I realized I have a bug in rthooks [here](https://github.com/cilium/tetragon/pull/2231/commits/fc0f13cb6f5ee0257512f972ed3c59b9a491b508#diff-0d06ff6b079b83e5948096c3ddab5693cdb637f590b5d6361a4ef84f2b244d6bR100) because I don't return anything if the container was not found, I will fix it when making changes in this file.

-----------
And one more thing, I'm trying to set up a local environment to see OCI hooks in action and I followed the demo with [cri-o](https://github.com/cilium/tetragon/blob/main/contrib/rthooks/tetragon-oci-hook/docs/demo.md). Unfortunately, I get `connecting to agent (context deadline exceeded) failed: unix:///var/run/cilium/tetragon/tetragon.sock` logs in `/opt/tetragon/tetragon-oci-hook.log`. Just wanted to mention it because it affects my ability to see how these hooks work, I'm still figuring out what's the issue.

kkourt: > If I pass an empty container name to [pfState.AddPodContainer](https://github.com/cilium/tetragon/pull/2231/files#diff-0d06ff6b079b83e5948096c3ddab5693cdb637f590b5d6361a4ef84f2b244d6bR113), how can I use arg.Watcher.FindContainer(containerID) from [containerMatches](https://github.com/cilium/tetragon/pull/2231/files#diff-980e5be14cfb8e087d0d09a0f42b7219a9510ad91dd199221abdc98d1767bf34R221) (where the filtering logic is happening)? If I understand correctly, I also need to pass the arg struct from [createContainerHook](https://github.com/cilium/tetragon/blob/main/pkg/policyfilter/rthooks/rthooks.go#L34) to [pfState.AddPodContainer](https://github.com/cilium/tetragon/pull/2231/files#diff-0d06ff6b079b83e5948096c3ddab5693cdb637f590b5d6361a4ef84f2b244d6bR113) and probably store it somewhere (in containerInfo?) to use it when filtering because it holds the hook information for this container?
> Maybe, an alternative would be to check if the container name is empty during filtering and if it is, try to find the pod (corev1.Pod) and the container in it by container ID? However, this sounds pretty complex and inefficient.
> Sorry if I overcomplicate it, I just don't have a clear picture in my head at the moment 😞

Yap, you are right. Didn't think it through. Let's keep it simple then and leave it as you have it now.

> Also, I realized I have a bug in rthooks [here](https://github.com/cilium/tetragon/pull/2231/commits/fc0f13cb6f5ee0257512f972ed3c59b9a491b508#diff-0d06ff6b079b83e5948096c3ddab5693cdb637f590b5d6361a4ef84f2b244d6bR100) because I don't return anything if the container was not found, I will fix it when making changes in this file.

Thanks! Yeah, let's fail for now if we cannot find the container name.

> And one more thing, I'm trying to set up a local environment to see OCI hooks in action and I followed the demo with [cri-o](https://github.com/cilium/tetragon/blob/main/contrib/rthooks/tetragon-oci-hook/docs/demo.md). Unfortunately, I get connecting to agent (context deadline exceeded) failed: unix:///var/run/cilium/tetragon/tetragon.sock logs in /opt/tetragon/tetragon-oci-hook.log. Just wanted to mention it because it affects my ability to see how these hooks work, I'm still figuring out what's the issue.

Yap, it would be great to test it with the OCI hook, so let's fix it!

Are you using the following helm option as indicated in the doc:
>         --set tetragon.grpc.address="unix:///var/run/cilium/tetragon/tetragon.sock" \

the OCI hooks only work with a gRPC server listening on the unix socket.


Many thanks for helping out with this! I think it got more complicated that we originally expected, but we will figure it out :)

pkg/policyfilter/state.go
```diff
@@ -222,6 +218,25 @@ func (pol *policy) podInfoMatches(pod *podInfo) bool {
...
+}
+
+// get a slice of container cgroupIDs that match the policy
+func (pol *policy) matchingContainersCgroupIDs(containers []containerInfo) []CgroupID {
```
kkourt, line 229: Can you add some explanation in the commit message about how the implementation of the filtering?
How about something like:

> The actual filtering happens in the matchingContainersCgroupIDs() function. Whenever a list of cgroup ids is needed (typically to update policyfilter map), we use this function that returns cgroup ids after applying any containerSelector filtering.


Feel free to change above if you feel is wrong, it's just my understanding from reading the code.

pkg/policyfilter/rthooks/rthooks.go
kkourt: Here's a PR to update the hooks side to support the container name: https://github.com/cilium/tetragon/pull/2248. It still needs some work (e.g, docs) but most of the things should be there.

pkg/policyfilter/state.go
BonySmoke: Thank you for pointing this out! I added explanations for `containerMatches` and `matchingContainersCgroupIDs`.
Hopefully, not it makes more sense:)

pkg/policyfilter/rthooks/rthooks.go
BonySmoke: It took me some time but in the end, I managed to get a local environment with OCI hook setup! :partying_face: 
Initially, it didn't work because of the bug I mentioned before (no return statement when container is not found).

I still get `context deadline exceeded` (I used the options as indicated in the doc) and after that, I see this log `failCheck determined that we should not fail this container, even if there was an error`. However, it looks like this log is not related to adding containers because when that's the case I saw this log `failCheck determined that we should fail`.

The great thing is that I also see logs like `hook request to agent succeeded` when a new container is added. :tada: 
It turned out, we cannot use `arg.Watcher.FindContainer()` because there is no container ID inside the pod at the moment of checking the hook.
Thanks to your comment about annotations and after inspecting the request, I also realized we can get the name from there.

--------
Your PR looks great (but also pretty big :smile: ). I wonder what I should do in my PR, wait for yours to be merged and then use the name you added?
For now, I implemented a very similar logic as in [your commit](https://github.com/cilium/tetragon/pull/2248/commits/f8c5e73db291f16852e46b163fc72fbe6386412e) (I wanted to do it yesterday, but you were quicker:) ). I did it so everything could work from my PR too and it actually works!
```
time="2024-03-23T10:15:05Z" level=trace msg="policyfilter: add pod container" cgroup-id=6403 container-id=crio-6dfc31c8cc5405abaab3851c095fb0b3f590d64c3726144fd24a4815b9a6c846 container-name=ubuntu hook=create-container namespace=default pod-id=58397ab1-2fad-416e-908c-bfefc91c88d7 sub-system=rthook.policyfilter
```

--------
Also, I have a silly question: how do you usually clean up disk space used by tetragon images after multiple builds? Every time I build an image with `make image`, it takes about 2GB of disk space and I'm already afraid of building new images locally :sweat_smile:

pkg/policyfilter/state.go
BonySmoke: Also, just wanted to say that I updated docs in the same commit (I updated it) because I noticed a place where I forgot to mention the new filtering mechanism.

pkg/policyfilter/rthooks/rthooks.go
kkourt: Thanks for the update.

> I still get context deadline exceeded (I used the options as indicated in the doc) and after that, I see this log failCheck determined that we should not fail this container, even if there was an error. However, it looks like this log is not related to adding containers because when that's the case I saw this log failCheck determined that we should fail.

This happens when the hook cannot contact the agent. The `failCheck` is just a check on whether the hook should fail the creation of the container or not. By default, the hook will _not_ fail containers in the `kube-system` namespace so that we can start tetragon  if the hook is running but the tetragon agent is not. 

> Your PR looks great (but also pretty big 😄 ). I wonder what I should do in my PR, wait for yours to be merged and then use the name you added?

That would me my preference. Would that be OK with you? You can rebase your PR on top of my branch until the PR is merged.

> I did it so everything could work from my PR too and it actually works!

Nice! :)

> Also, I have a silly question: how do you usually clean up disk space used by tetragon images after multiple builds? Every time I build an image with make image, it takes about 2GB of disk space and I'm already afraid of building new images locally 

Have you tried using `docker rmi` to remove the images?

BonySmoke: > That would me my preference. Would that be OK with you? You can rebase your PR on top of my branch until the PR is merged.
I'm absolutely fine with this! I can wait for it to be merged and then rebase and push the changes to avoid multiple rebasing:) (or you would like me to test if it works with my changes before merging?)

> Have you tried using docker rmi to remove the images?
I use `docker rmi $(docker images -f "dangling=true" -q)` to remove dangling images upon each build but it doesn't remove much:(
Even if I remove `cilium/tetragon:latest` locally, it doesn't clean up a lot of space :disappointed:

BonySmoke: > I just merged https://github.com/cilium/tetragon/pull/2248 so you can just rebase to main.

That's awesome, thank you!
I rebased to `main`, and now use `arg.Req.ContainerName` to get the container name. I tested it, it worked, and I pushed the changes:)

> We do use some images for the build, but I would expect those images to be matched by "dangling=true" as well. So not sure if there is a better way.

It was `buildx` cache! It was my first time using buildx in this project and it was caching layers of every image I built. I pruned it and now I'm good :relieved:

kkourt: Cool, thanks!

Please, "Re-request review" when you it's ready for review again.

pkg/policyfilter/rthooks/rthooks.go
```diff
@@ -92,15 +93,14 @@ func createContainerHook(_ context.Context, arg *rthooks.CreateContainerArg) err
...
-	if !containerFound {
-		log.WithError(err).Warnf("failed to find container information %s, aborting hook.", containerID)
-	}
 
-	containerName := container.Name
+	containerName := arg.Req.ContainerName
+	if containerName == "" {
+		err := fmt.Errorf("failed to find container information %s, aborting hook", containerID)
+		log.Warn(err)
+		return err
```
kkourt, line 102: Not sure if failing the is the right thing to do here. 
If we do not get a containerName, we can still do policies that do not depend on container name.

 I think it's better to i) issue a warning and ii) increase a metric if the container name is not there. This would require adding a new policyfilter metric (see policyfiltermetrics), but this should be relatively simple.
 
 WDYT?

kkourt (changes requested):

pkg/policyfilter/rthooks/rthooks.go
BonySmoke: Hello!
i) You are totally right, it would be bad to break other filtering functionality if we don't get the container name, I will only leave the warning, thank you!
ii) I also totally agree, but I have a question about a separate metric. I see we have the `error_type` label in `policyfilter_metrics_total` and currently it's not used. What if we set it to `nil` by default as it is now, but in case of an error, we could specify an error string type, for example, `container-name-not-found` or maybe something else?:)
If you think we need to fully separate these two metrics, how do you see the new metric?
For example, I'm thinking about a generic `policyfilter_metrics_error_total` and it would contain `op` and `error_type`.
What would be your preferred solution?:) In the meantime, I will test adding an error message to `error_type` just to understand if it's feasible:)

kkourt: > i) You are totally right, it would be bad to break other filtering functionality if we don't get the container name, I will only leave the warning, thank you!

Great, thanks!

> i) I also totally agree, but I have a question about a separate metric. I see we have the error_type label in policyfilter_metrics_total and currently it's not used. What if we set it to nil by default as it is now, but in case of an error, we could specify an error string type, for example, container-name-not-found or maybe something else?:)

I guess the error label was removed recently: https://github.com/cilium/tetragon/commit/425f91e2be6b83c4b654062d64ec49ec5e072d4f. I'll sync with @lambdanis and figure out what's the best way moving forward (https://github.com/cilium/tetragon/pull/2162#discussion_r1546398091), but I don't think we need to block this PR. We can do the metric in a followup PR.

Thanks!

BonySmoke: From the logical point of view, it's likely better to have a separate metric as you suggested because if we mix these error types that are not errors in code, this will be very confusing :smiley:

BonySmoke: Looks like I wrote my latest message right after yours, so it may sound strange if you read it in chronological order, sorry for that:)

> I guess the error label was removed recently: https://github.com/cilium/tetragon/commit/425f91e2be6b83c4b654062d64ec49ec5e072d4f. I'll sync with @lambdanis and figure out what's the best way moving forward (https://github.com/cilium/tetragon/pull/2162#discussion_r1546398091), but I don't think we need to block this PR. We can do the metric in a followup PR.

Thank you! For now, I pushed the change not to abort the hook when we don't find the name (squashed with the previous commit), I also updated the commit message to reflect it there too:)

BonySmoke: Thank you!
One concern I have in this case is let's say we set the `error` label which should indicate that the operation failed; however, in reality, it's only a warning and we don't abort the hook, so it's a kind of success :sweat_smile: (I hope not to confuse anyone)

lambdanis: > let's say we set the error label which should indicate that the operation failed; however, in reality, it's only a warning and we don't abort the hook, so it's a kind of success 

This would be confusing indeed, thanks for pointing this out. I'd expect the metric semantics to be:
`sum(tetragon_policyfilter_operations_total)` - number of attempted operations
`sum(tetragon_policyfilter_operations_total{error=""})` - number of successful operations
`sum(tetragon_policyfilter_operations_total{error!=""})` - number of unsuccessful operations

To count missed container names (which still results in a successful operation), a separate metric seems like a better solution.

kkourt: A separate metric only for error makes sense to me.
An alternative approach would be to have the operation be "get_container_name" but then we would have to count the successes as well.
Either approach is fine by me.

BonySmoke: Thank you for your thoughts!
I spent some time thinking about an optimal approach and at some point, I realized, that adding new metrics may not be an optimal solution, let me explain:
- Adding a new label specific to this case sounds simple but I think at some point there will be more of such exceptions and adding a new label for each could be too much (but I could be wrong, please let me know)
- Adding a separate label for errors/warnings also sounds great but we need to attach the error/warning to some operations, e.g. add container, delete pod, etc. This means we will end up with the same labels (subsystem, operation + new labels); therefore, I think adding a new generic label could be suboptimal (again, I can be wrong, but I wonder what would be the good design).

Therefore, I decided to add a new `warning` label to `policyfilter_metrics_total`. I implemented it in the latest patch. It's worth noting that there can be multiple warnings that we want to report from the same function, and also if there is a warning, we cannot simply say the operation was successful. I handled it using conditions, so if we report a warning, we don't report success but at the same time, we can report multiple warnings if needed.

Also, after adding the `warning` label, I found a missing return statement [here](https://github.com/cilium/tetragon/blob/main/pkg/policyfilter/rthooks/rthooks.go#L104) and realized there will be no metric at all because we return an error. Therefore, I decided to add the `error` label for such cases. @lambdanis had a concern about high cardinality; therefore, I decided to make errors expected, i.e. not use the error message but have an enum just like with operations and warnings.
In the end, I ended up with these labels
```
tetragon_policyfilter_metrics_total{error="",op="add",subsys="pod-handlers",warning=""} 14
tetragon_policyfilter_metrics_total{error="",op="add-container",subsys="rthooks",warning=""} 4
tetragon_policyfilter_metrics_total{error="",op="add-container",subsys="rthooks",warning="container-name-not-found"} 0
tetragon_policyfilter_metrics_total{error="",op="delete",subsys="pod-handlers",warning=""} 2
tetragon_policyfilter_metrics_total{error="",op="update",subsys="pod-handlers",warning=""} 57
tetragon_policyfilter_metrics_total{error="add-pod-container-failed",op="add-container",subsys="rthooks",warning=""} 0
tetragon_policyfilter_metrics_total{error="add-pod-error",op="add",subsys="pod-handlers",warning=""} 0
tetragon_policyfilter_metrics_total{error="delete-pod-error",op="delete",subsys="pod-handlers",warning=""} 0
tetragon_policyfilter_metrics_total{error="update-pod-error",op="update",subsys="pod-handlers",warning=""} 0
```
These changes can be reviewed in this commit https://github.com/cilium/tetragon/pull/2231/commits/9a695c336fbe2b8919f47788414deb2bfbca26f7

WDYT? :smiley: 

----
P.S. There was a failed CI action related to a duplicate signature, I fixed it.

kkourt: @BonySmoke I guess it's the second time we try and solve the same thing: https://github.com/cilium/tetragon/pull/2282 😅 

Not sure how I feel about the idea of adding an "error" and "warning". I feel it's somewhat confusing.

I would still go with the idea of either having a separate metric for errors or adding a new operation for `get-container-name`. I think a separate metric is probably the simplest approach.


> Also, after adding the warning label, I found a missing return statement [here](https://github.com/cilium/tetragon/blob/main/pkg/policyfilter/rthooks/rthooks.go#L104) and realized there will be no metric at all because we return an error.

Nice catch! From a quick look, this seems to be indeed a bug.

lambdanis: My worry with the "warning" label is that it makes it unclear what the metric is actually measuring.

`sum(tetragon_policyfilter_metrics_total)` should be meaningful, and my understanding is that it's the number of attempted operations (there can be bugs, but this would be the end goal at least). So it should be incremented exactly once for each operation - either with empty or specific error. Now, if we have multiple "warnings" for one operations and increment the metric for each of these warnings, then `sum(tetragon_policyfilter_metrics_total)` becomes meaningless. So to count warnings (meaning something went wrong, but the operation completed in the end) a separate metric is a way to go IMO.

Side note: I'm reasoning here about how it *should* look like in my opinion, not how it looks like now. There might be more bugs in how this metric is used, if that's the case they should be fixed :) But it's out of scope for this PR. Thanks for digging into it!

BonySmoke: I added a very simple metric in https://github.com/cilium/tetragon/pull/2231/commits/85b8663f7f2e9ee7c4e24ae564a5c192eb86b684, hopefully, it's intuitive enough:)
```
# HELP tetragon_policyfilter_hook_container_name_missing_total The total number of operations when the container name was missing in the OCI hook
# TYPE tetragon_policyfilter_hook_container_name_missing_total counter
tetragon_policyfilter_hook_container_name_missing_total 0
```
Wdyt?:)
For now, I didn't add any documentation, just want to know your opinion. Also, I didn't add any labels to keep it as simple as possible but I can add any labels if needed.

pkg/metrics/policyfiltermetrics/policyfiltermetrics.go
```diff
@@ -53,15 +53,26 @@ var (
 	}, []string{"subsys", "op"})
 )
 
+var (
+	PolicyFilterHookContainerNameMissingMetrics = prometheus.NewCounterVec(prometheus.CounterOpts{
```
lambdanis, line 57: Since this metric doesn't have any labels you can use `NewCounter` instead of `NewCounterVec`. Then you can also delete L74 initializing this metric (it will get initialized automatically).

lambdanis: Also please run `make metrics-docs` when adding a new metric to have it autodocumented. Thanks!

BonySmoke: Thank you for the review! Rookie mistakes, sorry :sweat_smile: 
Initially, I only saw your first message; therefore, I didn't run `make metrics-docs` before pushing the changes but now I pushed the changes to docs too, thank you (I was afraid I would need to document them manually:) )!

lambdanis: Can you rephrase the commit message to have a shorter subject? It's a detail, I know, but otherwise CI fails with:
```
Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 78)
```
https://github.com/cilium/tetragon/actions/runs/8553067127/job/23447481238?pr=2231

BonySmoke: sure, I agree that long commit messages are hard to read, I removed the word `metric` from the subject as it should be (hopefully) self-explanatory from `policyfiltermetrics`

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -90,6 +90,13 @@ type TracingPolicySpec struct {
 	// PodSelector selects pods that this policy applies to
 	PodSelector *slimv1.LabelSelector `json:"podSelector,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// ContainerSelector selects containers that this policy applies to.
+	// A map of container fields will be constructed in the same way as a map of labels.
+	// The name of the field represents the label "key", and the value of the field - label "value".
+	// Currently, only the "name" field is supported.
+	ContainerSelector *slimv1.LabelSelector `json:"containerSelector,omitempty"`
```
lambdanis, line 98: I was confused at first by `LabelSelector`, as there are no labels involved, but I guess by doing so we get `MatchExpressions` in addition to exact name match essentially for free? This seems like a good enough reason to reuse `LabelSelector`, just checking my understanding.

lambdanis (commented):
Looks good to me now, thanks!

I see there is a conflict in rthooks, hopefully not painful. Could you rebase and resolve it?

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
BonySmoke: Thank you for the review!
> I see there is a conflict in rthooks, hopefully not painful. Could you rebase and resolve it?

Sure, it was caused by the addition of the error label to the `policyfilter_metrics_total` label. I resolved it:)

------
> I was confused at first by LabelSelector, as there are no labels involved, but I guess by reusing it we get MatchExpressions in addition to exact name match essentially for free?

You are totally right! I imagined container fields the same as labels (a map of key-value pairs) and therefore decided to reuse the existing filtering logic. As you said, essentially for free :sweat_smile:

lambdanis (approved):
Looks good to me, thanks for the PR and all the fixes!

pkg/policyfilter/rthooks/rthooks.go
```diff
@@ -92,15 +92,25 @@ func createContainerHook(_ context.Context, arg *rthooks.CreateContainerArg) err
 		return err
 	}
 
+	var containerFound bool
+	var container *corev1.ContainerStatus
 	namespace := pod.ObjectMeta.Namespace
+	pod, container, containerFound = arg.Watcher.FindContainer(containerID)
+	if !containerFound {
+		log.WithError(err).Warnf("failed to find container information %s, aborting hook.", containerID)
```
kkourt: I think "aborting hook" might be misleading, how about "failed to find container information %s, continuing with hook".

Edit: this is fixed in a subsequent patch, so it's OK.

kkourt (approved):
This looks great to me!

Thanks for all your work, this is a great PR!

Issue #2241, opened 2024-03-18
Filters: Introduce Redaction Field Filter


willfindlay:
Tetragon events can leak sensitive data such as secrets stored environment variables, passed as command line arguments, or accessed by traced functions/system calls. In production environments, it may be desirable to redact sensitive information so that it does not appear in the event logs. While field filters are sometimes an option, it is not always desirable to completely drop a field, as we may still wish to access some relevant subset of the information therein. Therefore, we need to achieve some middle ground here.

To do this, we can introduce a new filter type, the redaction filter. A redaction filter works like a field filter, except instead of completely dropping the field, it redacts a subset of information, which could be selected by a regular expression for example. In order to enable precise selection of information to be redacted, I would propose using regex capture groups to indicate which parts of the matched string should be redacted.

I imagine the filter syntax looking something like this:

```json
{
	"match": {
		"event_set": ["PROCESS_EXEC"]
	},
	"redact": [
		{"field": "args", "regex": "--password\\s+(\\S+)"}
	]
}
```

The above filter would match substrings in the `args` field of `PROCESS_EXEC` events that look like `--password foobar` and replace the capture groups of the regular expression with `*****`, effectively redacting the sensitive information.

PR #2243, opened 2024-03-20
implement redaction filters

willfindlay:
Implement a new field filter type, the redaction filter. Redaction filters use regular expressions to suppress sensitive information in string fields in Tetragon events. When a regular expression in a redaction filter matches a string, everything inside of its capture groups is replaced with `*****`, effectively censoring the output. For example, the regular expression `(?:--password|-p)(?:\s+|=)(\S*)` will convert the string "--password=foo" into "--password=*****".

In some cases, it is not desirable to apply a redaction filter to all events. For this use case, redaction filters also include an event filter which can be used to select events to redact. This event filter is configured with the same syntax as an export filter. As a more concrete example:

    {"match": {"binary_regex": ["^foo$"]}, "redact": ["\W(qux)\W"]}

The above filter would redact any occurrences of the word "qux" in events with the binary name "foo".

Due to the sensitive nature of redaction, these filters are applied as configured in the
agent, regardless of whether an event is exported via gRPC or the JSON exporter. In other
words, redaction filter configuration always happens at the agent config level, not in the
gRPC client CLI.

```release-note
Introduce redaction filters for censoring sensitive string data in process events.
```

Fixes: #2241


willfindlay:
marking as draft while I fix an oversight with empty redaction filters

cmd/tetragon/main.go
```diff
@@ -401,11 +406,18 @@ func tetragonExecute() error {
 
 	hookRunner := rthooks.GlobalRunner().WithWatcher(k8sWatcher)
 
+	redactionFilters, err := getRedactionFilters()
+	if err != nil {
+		return err
+	}
+	log.WithFields(logrus.Fields{"redactionFilters": redactionFilters}).Debug("Configured redaction filters")
```
kkourt, line 413: IMO, I think `Info` is better  here.

install/kubernetes/tetragon/values.yaml
```diff
@@ -111,7 +111,33 @@ tetragon:
...
+  # Now, an event which contains the string "--password=foo" would have that
+  # string replaced with "--password=*****".
+  # 
+  # Suppose we also see some passwords passed via the -p shorthand for a specific binary, foo.
+  # We can also redact these as follows:
+  #
+  #   {"match": [{"binary_regex": "(?:^|/)foo$"}], "redact": ["-p(?:\s+|=)(\S*)"]}
+  #
+  # With both of the above redaction filters in place, we are now redacting all
+  # password arguments.
```
kkourt, line 138: ❤️

kkourt (approved):
LGTM, thanks!

cmd/tetragon/main.go
willfindlay: Sure why not. For the record we're also dumping other filter types in debug logs. Maybe worth changing those too

willfindlay: Updated

jrfastab (approved):

Issue #2257, opened 2024-03-25
Local built image does not work as expected on MacOS arm64


f1ko:
### What happened?

Built a clean local image with `make images` on commit `505ce7f861efd13dbe49e48a730857b7b6736679`.
When running that image on a kind cluster (orbstack, arm64) the tetragon container shuts down with an error.

### Tetragon Version

Local build:
```
# tetra version
CLI version: 505ce7f8
```

### Kernel Version

From the node Tetragon is running on:
```
# uname -a
Linux kind-control-plane 6.4.16-orbstack-00105-g14094bfeec09 #1 SMP Mon Sep 18 21:45:38 UTC 2023 aarch64 GNU/Linux
```

### Kubernetes Version

```
$ kubectl version
Client Version: v1.29.1
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.27.3
```

### Relevant log output

```shell
$ kubectl logs -f tetragon-nhrxh -c tetragon
time="2024-03-25T13:37:51Z" level=info msg="Loaded config from directory" config-dir=/etc/tetragon/tetragon.conf.d/
time="2024-03-25T13:37:51Z" level=info msg="Starting gops server" addr="localhost:8118"
time="2024-03-25T13:37:51Z" level=info msg="Starting tetragon" version=505ce7f8
[...]

PR #2258, opened 2024-03-25
Makefile: add native build support for arm64 MacOS

f1ko:
On MacOS `uname -m` returns `arm64` whereas on Linux it returns `aarch64`.

Fixes https://github.com/cilium/tetragon/issues/2257

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -32,6 +32,9 @@ endif
 ifeq ($(UNAME_M),aarch64)
 	TARGET_ARCH ?= arm64
 endif
+ifeq ($(UNAME_M),arm64)
+	TARGET_ARCH ?= arm64
+endif
 TARGET_ARCH ?= amd64
 
 # Set GOARCH to TARGET_ARCH only if it's not set so that we can still use both


mtardy:
hey, I see that but I don't think you can build tetragon on macOS (because some pkg don't build on darwin) and we want to do that?

f1ko:
Essentially what is happening is that the Makefile is being evaluated on MacOS but the build itself can then happen elsewhere.
That is, for example, the case when using the `make images` target.
A Linux machine is required for using docker, so the actual build is happening on Linux anyways.
However, the target platform is being evaluated beforehand already while being on MacOS as `make images` is "natively" executed on the host.
By not being able to evaluate the target correctly on MacOS it falls back to creating an amd64 image which is later on not usable in kind running on the same arm64 machine.

Now an alternative may be to do everything on Linux immediately so that the target is rightfully evaluated when executing `make images`, however, it is a pain to do this setup for every repository that comes along.
For people working with the same small number of repositories all the time this is a one-time configuration that doesn't take too much effort to setup and doesn't need a lot of maintenance.
However, for someone that is constantly working with new repositories this is a pain to setup and to maintain in case of full disaster recovery automation.
One has to have enough space on the Linux VM, need to synchronise files back and forth between MacOS or maintain mounts, etc.

Therefore a common pattern on MacOS is to have a local lightweight Linux VM and exposing the docker socket to MacOS host so that one can "natively" use docker commands on MacOS.
Doing so results in a one-time setup that scales well regardless of projects that are being worked on and is therefore easier to automate and backup as well.

The proposed change does not break anything for existing users.
All it does is add support for a common workflow for MacOS development.

mtardy (approved):
yep okay why not! I'm thinking by the way that it would be a good change to write it with if/else statements:

```makefile
ifeq ($(UNAME_M),x86_64)
	TARGET_ARCH ?= amd64
else ifeq ($(UNAME_M),aarch64)
	TARGET_ARCH ?= arm64
else ifeq ($(UNAME_M),arm64)
	TARGET_ARCH ?= arm64
else
	TARGET_ARCH ?= amd64
endif
```

mtardy:
If that can help you, let's merge this!

Issue #2261, opened 2024-03-27
Filter out bash kubernetes healthchecks


alexeysofin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

When kubernetest pod is given exec probe healthcheck, e.g.

```yaml
readinessProbe:
  exec:
    command:
    - /health/ping_readiness_local.sh
    - "1"
```
taken from bitnami redis chart

after linux execs this command, a real execve contains script name in arguments list because linux appends it there, so tetragon seens every command inside a script as follows:

```json
"parent": {
    "binary": "/health/ping_readiness_local.sh",
    "arguments": "/health/ping_readiness_local.sh 1",
```

therefore, 

```yaml
exportDenyList: |-
  {"health_check":true}
```

does not filter such events.

### Describe the feature you would like

Filter bash healthchecks.

### Describe your proposed solution

Add ability to filter bash healchecks because almost all k8s healthchecks are done with sh scripts. It's complicated by the fact that many public charts have `sh -c script.sh` so tetragon will not even see a probe as a parent, but that's a feasible tradeoff.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2262, opened 2024-03-27
Filter out bash kubernetes healthchecks

alexeysofin:
This patch adds this by

1. filtering *.sh binaries in healthcheck filter
2. removing first argument in exec to correctly compare it to k8s pod's exec probe

Fixes: #2261

### DIFF
--- a/pkg/filters/health_check.go
+++ b/pkg/filters/health_check.go
@@ -22,10 +22,23 @@ func MaybeExecProbe(binary string, args string, execProbe []string) bool {
 	if err != nil {
 		return false
 	}
+
+	// Exec will append a script name to argument list if a sh/bash script is executed with a shebang,
+	// so we need to remove the first argument so that we can compare it to execProbe.
+	// e.g.
+	// "binary": "/health/ping_liveness_local.sh",
+	// "arguments": "/health/ping_liveness_local.sh 5"
+	// concatenated together will be ["/health/ping_liveness_local.sh", "/health/ping_liveness_local.sh", "5"],
+	// but execProbe will have only ["/health/ping_liveness_local.sh", "5"].
+	if execProbe[0] == binary && len(argList) > 0 && argList[0] == binary {
+		argList = argList[1:]
+	}
+
 	processCommand := append([]string{binary}, argList...)
 	if len(execProbe) != len(processCommand) {
 		return false
 	}
+
 	if path.IsAbs(execProbe[0]) {
 		// exec probe path is absolute. Compare the full paths.
 		if processCommand[0] != execProbe[0] {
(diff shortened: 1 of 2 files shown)


pkg/filters/health_check.go
```diff
@@ -22,10 +23,23 @@ func MaybeExecProbe(binary string, args string, execProbe []string) bool {
...
+	// Exec will append a script name to argument list if a sh/bash script is executed with a shebang,
+	// so we need to remove the first argument so that we can compare it to execProbe.
+	// e.g.
+	// "binary": "/health/ping_liveness_local.sh",
+	// "arguments": "/health/ping_liveness_local.sh 5"
+	// concatenated together will be ["/health/ping_liveness_local.sh", "/health/ping_liveness_local.sh", "5"],
+	// but execProbe will have only ["/health/ping_liveness_local.sh", "5"].
+	if binary == execProbe[0] && strings.HasSuffix(binary, ".sh") && len(argList) > 0 && binary == argList[0] {
+		argList = argList[1:]
+	}
```
tixxdz, line 36: Hi thanks for the contributions.

the ".sh" suffix test is weak, as it is not a standard thing. Yes IIRC shebang execution will always end up with binary == argList[0].

So if the `execProbe []string` argument here is: `["/health/ping_liveness_local.sh", "5"]` , then you could write if test with something like:
```
if len(argList) > 0 && execProbe[0] == binary && argList[0] == binary && reflect.DeepEqual(argList[1:], execProbe) {
   
}
```

What do you think?

tixxdz (changes requested):

tixxdz:
Let's cc @willfindlay for more input

pkg/filters/health_check.go
alexeysofin: Hi, thanks for comments, I agree that sh prefix is a bit redundant, that was just a double-protection from accidentally filtering scripts that are for some reason run with their name as the first argument.

In any case I added your proposal, except I don't do `reflect.DeepEqual(argList[1:], execProbe)` because comparison is done below.

What do you think?

tixxdz: Sounds good alright!

tixxdz (commented):
LGTM overall, thanks!

Still two things please: can you make it in one patch, and investigate the failing CI test as it could be related to your patch?

tixxdz (approved):
Alright much appreciated ;-) !

alexeysofin:
@tixxdz thank you very much! Let's see how CI passes

alexeysofin:
@tixxdz well, it failed, now with some transient error in `hostedtoolcache` and a huge stack trace.

These tests seem rather flaky, maybe try restarting the job?

PR #2267, opened 2024-03-28
backports/willfindlay/1.0/redaction filters

willfindlay:
[upstream commit 69a47435fe6f052a3f97be7599237616da8e0fb0]

Implement a new field filter type, the refaction filter. Redaction filters use regular expressions to suppress sensitive information in string fields in Tetragon events. When a regular expression in a redcation filter matches a string, everything inside of its capture groups is replaced with `*****`, effectively censoring the output. For example, the regular expression `(?:--password|-p)(?:\s+|=)(\S*)` will convert the string "--password=foo" into "--password=*****".

In some cases, it is not desirable to apply a redaction filter to all events. For this use case, redaction filters also include an event filter which can be used to select events to redact. This event filter is configured with the same syntax as an export filter. As a more concrete example:

    {"match": {"binary_regex": ["^foo$"]}, "redact": ["\W(qux)\W"]}

The above filter would redact any occurrences of the word "qux" in events with the binary name "foo".

Due to the sensitive nature of redaction, these filters are applied as configured in the agent, regardless of whether an event is exported via gRPC or the JSON exporter. In other words, redaction filter configuration always happens at the agent config level, not in the gRPC client CLI.

```release-note
Implement redaction filters to redact sensitive information from process events.
```


mtardy (approved):
Depending of the Go tests results but this should be good, thanks! Maybe we should disable some of the tests for v1.0 since they might fail anyway (like the links or validated of generated files)

willfindlay:
> Maybe we should disable some of the tests for v1.0 since they might fail anyway (like the links or validated of generated files)

For the generated files, it was actually user error. I forgot to generate the docs for CLI flags. 

I think it makes sense to disable links checker here though. For now let's ignore it.

willfindlay:
Shipping it

PR #2275, opened 2024-03-30
tetragon: store caps during fork

tixxdz:


kkourt (approved):
Cool, thanks!

I have on question. You mention:
> so we can check later if capabilities changed during the execve,

Where is this check happening?

tixxdz:
> Cool, thanks!
> 
> I have on question. You mention:
> 
> > so we can check later if capabilities changed during the execve,
> 
> Where is this check happening?

On the exec path on current, just after commit_cred() and before our exec hook, this is our window.

PR #2287, opened 2024-04-02
metrics: Remove pod from the queue after deleting metrics

lambdanis:
This is done to prevent a memory leak in environments with a high pod churn or very long-running Tetragon agents.

Credits to Tam Mach for fixing a similar issue in cilium/cilium#31714.


mtardy (approved):
Thanks!!

PR #2289, opened 2024-04-03
jsonchecker: reset state of the export file

kkourt:
Some tests run JsonTestCheckExpect multiple times. Add a KeepExportFile
function to keep the export file and call it at the start of JsonTestCheckExpect.


mtardy (approved):
lgtm

PR #2291, opened 2024-04-03
Improve metric for map and cache sizes

lambdanis:
See commits for details.

```release-note
Metrics for map and cache sizes are improved:
* tetragon_map_in_use_gauge metric is renamed to tetragon_map_entries and doesn't have total label anymore
* New tetragon_map_capacity metric exposes the BPF maps capacity
* New tetragon_event_cache_entries metric measures the event cache size
* New tetragon_process_cache_size metric measures the process cache size
* New tetragon_process_cache_capacity metric exposes the process cache capacity
```


mtardy (approved):
Thanks, that is a nice package of fixes!

PR #2314, opened 2024-04-08
backports/1.0: exporter: fix hang when exporter fails to start

willfindlay:
[upstream commit: 44bd44c]

The Exporter.Start() implementation uses a WaitGroup to wait until the exporter is ready after launching it in a goroutine. Unfortunately, this WaitGroup never gets set to Done when we encounter an error while starting the exporter. This ultimately causes the Tetragon process to hang indefinitely. Fix the issue by simply calling readyWG.Done() when we encounter an error here.

```release-note
Fix a hang when the exporter fails to start.
```


mtardy (approved):
Thanks for the backport, let's merge that and release v1.0.3!

PR #2318, opened 2024-04-08
Update events.md

zazathomas:
Added description for policy_names and capabilities filtering to the official events docs.


kkourt (approved):
Thanks! Would be nice if we could generate this from https://github.com/cilium/tetragon/tree/main/api/v1#filter, but in the meantime the patch LGTM.

Issue #2255, opened 2024-03-25
ci: CRD version check fails on PRs from forks


lambdanis:
The [CRD version check](https://github.com/cilium/tetragon/blob/main/.github/workflows/validate-crd.yaml) fails on PRs from forks that modify CRDs:
```
fatal: bad revision 'origin/<branch in the fork>'
Error: Process completed with exit code 128.
```
Examples: #2231, #2065

mtardy:
I have this here https://github.com/cilium/tetragon/issues/2010 as well.

kkourt:
Checked t hat the issue was fixed in https://github.com/cilium/tetragon/pull/2320

PR #2319, opened 2024-04-09
validate-crd: use SHAs for diff

kkourt:
The previous version did not work with forks.
Use pull_request.{base,head}.sha for the diff of files to fix above issue.

Fixes: https://github.com/cilium/tetragon/issues/2255


mtardy (approved):
Thanks!

PR #2322, opened 2024-04-10
filters/redaction: fix crash related to redaction filters and improve performance

willfindlay:
The previous redaction filters implementation was unsound, as it made modifications to the
process object within an event message, but this object was already being shared in the
event cache. This caused a data race that in turn resulted in a frequent crash when
applying redaction filters on a modest workload. Moreover, the prior implementation was
not very performant, since it relied on iterating over all fields in an event message to
redact strings. In practice, we only really care about arguments in the majority of use
cases.

To address the above issues, we make a small breaking change in how the redaction filters
work to focus ONLY on arguments and apply them much earlier in the pipeline. This both
fixes the crash and significantly reduces performance impact, while solving the primary
use case.

```release-note
Change the behaviour of redaction filters to fix a critical bug and improve performance.
```


api/v1/tetragon/events.proto
```diff
@@ -84,10 +84,12 @@ message CapFilterSet {
 }
 
 message RedactionFilter {
-    // Match events that the redaction filter will apply to.
-    repeated Filter match = 1;
+    // Deprecated, do not use.
+    repeated Filter match = 1 [deprecated=true];
```
jrfastab, line 88: by the way I think it makes sense to have both the specific optimized redactions for binary_regex and also the general ones even if they are slower. wdyt? (note not requesting a change for this PR but maybe something to consider.)

cmd/tetragon/main.go
```diff
@@ -109,9 +109,16 @@ func getFieldFilters() ([]*tetragon.FieldFilter, error) {
 	return filters, nil
 }
 
-func getRedactionFilters() (fieldfilters.RedactionFilterList, error) {
+func setRedactionFilters() error {
+	var err error
 	redactionFilters := viper.GetString(option.KeyRedactionFilters)
```
jrfastab, line 114: I think/believe finding a way to push these through a CRD would be more dynamic/easier to manage. But would be curious what experts managing clusters today think. For this PR lets just continue to use config like done here.

jrfastab (approved):

api/v1/tetragon/events.proto
willfindlay: I think we can definitely consider this moving forward.

PR #2324, opened 2024-04-11
backports/v1.0: redaction: fix crash related to redaction filters + improve performance

willfindlay:
[upstream commit: 91012a8a4f280c4c8a6012a479091dece87e2ea3]

The previous redaction filters implementation was unsound, as it made modifications to the process object within an event message, but this object was already being shared in the event cache. This caused a data race that in turn resulted in a frequent crash when applying redaction filters on a modest workload. Moreover, the prior implementation was not very performant, since it relied on iterating over all fields in an event message to redact strings. In practice, we only really care about arguments in the majority of use cases.

To address the above issues, we make a small breaking change in how the redaction filters work to focus ONLY on arguments and apply them much earlier in the pipeline. This both fixes the crash and significantly reduces performance impact, while solving the primary use case.

```release-note
Change the behaviour of redaction filters to fix a critical bug and improve performance.
```


jrfastab (approved):
..

PR #2332, opened 2024-04-12
api: revert breaking change in proto

willfindlay:
Commit b6d35839c1b4e2f0e5c848ecdf0a59763a716e67 introduced a breaking change in our protobuf API by renaming the existing stack_trace field in the kprobe message to kernel_stack_trace when introducing user stack traces. While a change of this nature is not ordinarily problematic in protocol buffer APIs, it is an issue for Tetragon specifically since we rely on protojson for event exports. This effectively means that renaming a field breaks compatibility for downstream consumers of event exports. Revert the rename to prevent breakages.

```release-note
Fix an accidental renaming of the stack_traces field in the kprobe gRCP message which would have broken our JSON API.
```


michi-covalent (approved):

mtardy (approved):
> While a change of this nature is not ordinarily problematic in protocol buffer APIs, it is an issue for Tetragon specifically since we rely on protojson for event exports.

indeed

willfindlay:
After some discussion offline, we determined that in this case the breakage is probably ok because of the nature of the feature. If this is not the case for someone, I'd be happy to hear an argument to reopen this. Closing for now.

PR #2333, opened 2024-04-12
vmtests: deduplicating code using LVH library for arm64 support

mtardy:
Deduplicate some codes shared between vmtests and LVH:
- inject LVH runner.RunConf into vmtests RunConf and reuse existing LVH configuration fields.
- use LVH runner.BuildQemuArgs along with custom vmtests logic.
- removes the support for HVF, we would need to put it in LVH to get this again but I don't think this is used nowadays.

This is mainly so that we benefit from the arm64 support as LVH was recently updated to support it.

### DIFF
--- a/cmd/tetragon-vmtests-run/conf.go
+++ b/cmd/tetragon-vmtests-run/conf.go
@@ -4,37 +4,35 @@
 package main
 
 import (
+	"fmt"
 	"path/filepath"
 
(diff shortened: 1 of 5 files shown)


mtardy:
I wouldn't mind a review @kkourt so that I can rebase afterward and make sure I don't rebase twice :)

kkourt (approved):
LGTM, thanks :)

Issue #2299, opened 2024-04-05
Creating a TracingPolicyNamespaced with the same name for a different namespace does not get applied. 


joshuajorel:
### What happened?

1. In our K8s environment, we deployed the same policy to two different namespaces. However, only the first policy gets applied. This was confirmed by running the `tetra tp list` command on the tetra pods. We tested this behavior with the fd-install `TracingPolicyNamespaced` config in two different namespaces (default and test):

```
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "fd-install"
spec:
  kprobes:
  - call: "fd_install"
    syscall: false
    args:
    - index: 0
      type: "int"
    - index: 1
      type: "file"
    selectors:
    - matchArgs:
      - index: 1
        operator: "Equal"
        values:
        - "/tmp/tetragon"
      matchActions:
      - action: Sigkill
```

The following is the output of `tetra tp list`:
```
[5] fd-install enabled:true filterID:5 namespace:default sensors:gkp-sensor-5
```
Only one policy is applied and the policy for test is not even though the k8s resource exists.

### Tetragon Version

CLI version: v1.0.1
Server version: v1.0.2 (installed via Helm)

### Kernel Version

Linux ubuntu-noble 6.8.0-11-generic #11-Ubuntu SMP PREEMPT_DYNAMIC Wed Feb 14 00:29:05 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux

### Kubernetes Version

Client Version: v1.29.2
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.28.3

### Bugtool

time="2024-04-03T23:20:06Z" level=info msg="saving init info"
time="2024-04-03T23:20:06Z" level=info msg="retrieving lib directory" libDir=/var/lib/tetragon/
time="2024-04-03T23:20:06Z" level=warning msg="not an object file, ignoring" path=/var/lib/tetragon/
time="2024-04-03T23:20:10Z" level=info msg="skipping metadata directory" path=/var/lib/tetragon/metadata
time="2024-04-03T23:20:10Z" level=warning msg="no btf filename in tetragon config, attempting to fall back to /sys/kernel/btf/vmlinux"
time="2024-04-03T23:20:11Z" level=info msg="btf file added" btfFname=/sys/kernel/btf/vmlinux
[...]

Issue #1272, opened 2023-07-25
deprecate sensors gRPC interface


kkourt:
The sensors gRPC interface is outdated. We should deprecate it and use tracingpolicies instead. Before deprecating, however, we need to provide the functionality to disable a tracingpolicy as we, currently, do with sensors.

- [x] implement disable tracing policies gRPC call: https://github.com/cilium/tetragon/pull/1562
- [ ] deprecate sensors

kkourt:
In an offline discussion, @mtardy mentioned that:
> disable [a policy] without modifying the resource does not sounds very "declarative"/"k8s" way. I think you can end up in a very confusing situation where you don't understand why you perfectly fine tracing policy does not apply. Having a field in the policy [that allows you to disable it] would make more sense indeed

inliquid:
@kkourt could you please clarify a little bit, does that mean that in order to "disable" policy I will have to delete one and then apply it with some additional field `disabled` set to `true`?

As a side note, I think when it comes to working with policies over gRPC instead of CRD, it's already some sort of not very "k8s way". But at the same it's the reason why we want tot use it, as it's much more flexible.

kkourt:
> @kkourt could you please clarify a little bit, does that mean that in order to "disable" policy I will have to delete one and then apply it with some additional field `disabled` set to `true`?

No. We want to keep a distinction between removing and disabling. The idea would be that you will have a filed in the policy to disable it.

> As a side note, I think when it comes to working with policies over gRPC instead of CRD, it's already some sort of not very "k8s way". But at the same it's the reason why we want to use it, as it's much more flexible.

kkourt:
> > The idea would be that you will have a field in the policy to disable the policy's functionality and then you would just update the k8s resource (without deleting).
> 
> Thanks, the point here is that we are managing `tetragon` through gRPC, and there is no API method to update the policy, only `AddTracingPolicy`, `DeleteTracingPolicy` and `ListTracingPolicies`, so how do I disable a policy it in this case?

I think we would need to add a new gRPC method for that. I haven't thought it through, but I can see two options: (i) having an UpdateTracingPolicy method that you can use to disable a policy or (ii) have a DisableTracingPolicy method.

inliquid:
> (ii) have a DisableTracingPolicy method

That's what I thought of initially.

inliquid:
Yep that would be totally fine.

PR #2337, opened 2024-04-16
feat: Add collection key to differentiate namespaced policies

joshuajorel:
Currently, `TracingPolicy` and `TracingPolicyNamespaced` encounter conflicts whenever a policy is applied with the same name. A policy with the same name, even if it is in a different namespace does not get applied. This behavior should be allowed given the k8s context e.g. a policy in one namespace is a different policy from one in a different namespace even with the same name. 

Fixes: #2299


pkg/sensors/collection.go
```diff
@@ -36,6 +36,12 @@ func (s TracingPolicyState) ToTetragonState() tetragon.TracingPolicyState {
 	}
 }
 
+// collection_key is the unique key for sensors
+// this enables policies with the same name for different namespaces
+type collection_key struct {
```
lambdanis, line 41:
```suggestion
type collectionKey struct {
```
Go convention is to use camelCase for all symbols. Following conventions consistently makes the code easier to read, so let's follow it here.

pkg/sensors/handler.go
```diff
@@ -94,10 +94,19 @@ func (h *handler) updatePolicyFilter(tp tracingpolicy.TracingPolicy, tpID uint64
 }
 
 func (h *handler) addTracingPolicy(op *tracingPolicyAdd) error {
+
+	var namespace string
+	if tpNs, ok := op.tp.(tracingpolicy.TracingPolicyNamespaced); ok {
+		namespace = tpNs.TpNamespace()
+	} else {
+		namespace = ""
+	}
```
lambdanis, line 103:
```suggestion
	}
```
nit: `else` is not needed here

pkg/sensors/handler.go
```diff
@@ -94,10 +94,19 @@ func (h *handler) updatePolicyFilter(tp tracingpolicy.TracingPolicy, tpID uint64
...
+
+	collectionKey := collection_key{op.name, namespace}
 	// allow overriding existing policy collection that resulted in an error
 	// during the loading state
-	if col, exists := h.collections[op.name]; exists && col.state != LoadErrorState {
-		return fmt.Errorf("failed to add tracing policy %s, a sensor collection with the name already exists", op.name)
+	if col, exists := h.collections[collectionKey]; exists && col.state != LoadErrorState {
+		return fmt.Errorf("failed to add tracing policy %s, a sensor collection with the name already exists", collectionKey)
```
lambdanis, line 109: For pretty printing, you can define `String()` method on `collectionKey` that returns "namespace/name" for namespaced policies and just "name" for clusterwide.

lambdanis (commented):
Thank you for tackling this @joshuajorel! I see it's still WIP, but I left a few comments.

pkg/sensors/manager.go
```diff
@@ -349,9 +349,10 @@ type tracingPolicyAdd struct {
 }
 
 type tracingPolicyDelete struct {
-	ctx     context.Context
-	name    string
-	retChan chan error
+	ctx       context.Context
+	name      string
+	namespace string
```
kkourt, line 354: Please use a `collectionKey` type here. This will force us to update all the callers.

You would need for example to update https://github.com/cilium/tetragon/blob/25a0050cda943d8744b3f846d8c4a82219bce2da/pkg/sensors/manager.go#L204-L208

pkg/sensors/manager.go
```diff
@@ -361,15 +362,17 @@ type tracingPolicyList struct {
 }
 
 type tracingPolicyDisable struct {
-	ctx     context.Context
-	name    string
-	retChan chan error
+	ctx       context.Context
+	name      string
+	namespace string
```
kkourt, line 367: same as `tracingPolicyDelete`, please use a `collectionKey` type  here and update its callers.

pkg/sensors/manager.go
```diff
@@ -380,32 +383,36 @@ type sensorOp interface {
 
 // sensorAdd adds a sensor
 type sensorAdd struct {
-	ctx     context.Context
-	name    string
-	sensor  *Sensor
-	retChan chan error
+	ctx       context.Context
+	name      string
+	namespace string
```
kkourt, line 388: For `sensorAdd` there is no namespace support, so let's not add a `namespace`.

pkg/sensors/manager.go
```diff
@@ -380,32 +383,36 @@ type sensorOp interface {
...
 type sensorRemove struct {
-	ctx     context.Context
-	name    string
-	all     bool
-	retChan chan error
+	ctx       context.Context
+	name      string
+	namespace string
+	all       bool
+	retChan   chan error
```
kkourt, line 399: Similar to `sensorAdd`,  let's not add a namespace.

pkg/sensors/manager.go
```diff
@@ -380,32 +383,36 @@ type sensorOp interface {
...
 }
 
 // sensorEnable enables a sensor
 type sensorEnable struct {
-	ctx     context.Context
-	name    string
-	retChan chan error
+	ctx       context.Context
+	name      string
+	namespace string
```
kkourt, line 406: Similar to `sensorAdd`, let's not add a namespace.

pkg/sensors/manager.go
```diff
@@ -380,32 +383,36 @@ type sensorOp interface {
...
 }
 
 // sensorDisable disables a sensor
 type sensorDisable struct {
-	ctx     context.Context
-	name    string
-	retChan chan error
+	ctx       context.Context
+	name      string
+	namespace string
```
kkourt, line 414: Similar to `sensorAdd`, let's not add a namespace.

pkg/sensors/manager.go
```diff
@@ -361,15 +362,17 @@ type tracingPolicyList struct {
...
+	retChan   chan error
 }
 
 type tracingPolicyEnable struct {
-	ctx     context.Context
-	name    string
-	retChan chan error
+	ctx       context.Context
+	name      string
+	namespace string
```
kkourt, line 374: same as `tracingPolicyDelete`, please use a `collectionKey` type  here and update its callers.

kkourt (changes requested):
Thanks!

pkg/sensors/manager.go
joshuajorel: Just a clarification, do you mean to add a `collectionKey` type or replacing the value of `name` altogether? Example:

```
type tracingPolicyAdd struct {
	ctx     context.Context
	name string
	ck      collectionKey
	tp      tracingpolicy.TracingPolicy
	retChan chan error
}
```

or 

```
type tracingPolicyAdd struct {
	ctx     context.Context
	ck      collectionKey
	tp      tracingpolicy.TracingPolicy
	retChan chan error
}
```

pkg/sensors/collection.go
joshuajorel: Updated. Thanks!

pkg/sensors/handler.go
joshuajorel: Added this in the latest commit as well. Thanks!

pkg/sensors/handler.go
joshuajorel: Updated this as well.

joshuajorel:
@kkourt @lambdanis - what is the usecase for enabling and disabling a `Sensor`? Since every interaction with Tetragon (as far I can see) goes through a `TracingPolicy`, where is this functionality used?

kkourt:
> @kkourt @lambdanis - what is the usecase for enabling and disabling a `Sensor`? Since every interaction with Tetragon (as far I can see) goes through a `TracingPolicy`, where is this functionality used?

Hi,

The sensor interface (and infrastructure) is indeed deprecated. There is an issue to remove it https://github.com/cilium/tetragon/issues/1272, but we first need to provide means to disable a tracing policy as we do with sensors.

pkg/sensors/manager.go
joshuajorel: Done with the latest commit.

lambdanis:
Hey @joshuajorel thank you for the updates.

Could you rewrite the commits so that changes are clearly grouped and documented? I think one commit introducing `collectionKey`, one introducing API changes, and one bringing everything together would be good. Or a single commit with all changes would be okay too. Please also keep the commit message titles no longer that 75 characters - otherwise the CI job fails.

joshuajorel:
> Looks good to me, thank you @joshuajorel. Let's wait for @kkourt to confirm nothing is missing.

Thank you for taking the time to review @lambdanis ! Will wait for @kkourt 's review

pkg/sensors/manager.go
```diff
@@ -344,13 +356,15 @@ type Manager struct {
 type tracingPolicyAdd struct {
 	ctx     context.Context
 	name    string
+	ck      collectionKey
```
kkourt, line 359: Do we still need the name here? Isn't it `ck.name`?

pkg/sensors/manager.go
```diff
@@ -363,12 +377,14 @@ type tracingPolicyList struct {
...
 	retChan chan error
 }
 
 type tracingPolicyEnable struct {
 	ctx     context.Context
 	name    string
+	ck      collectionKey
```
kkourt, line 387: Same as the previous comment.

pkg/sensors/manager.go
```diff
@@ -363,12 +377,14 @@ type tracingPolicyList struct {
 type tracingPolicyDisable struct {
 	ctx     context.Context
 	name    string
+	ck      collectionKey
```
kkourt, line 380: Same as the previous comment.

kkourt (changes requested):
Thanks, changes LGTM. I left one comment about removing the unneeded `name` field.

One of the things we also need to do is update the CLI: https://github.com/cilium/tetragon/blob/main/cmd/tetra/tracingpolicy/tracingpolicy.go. Could you also do that?

joshuajorel:
> Thanks, changes LGTM. I left one comment about removing the unneeded `name` field.
> 
> 
> 
> One of the things we also need to do is update the CLI: https://github.com/cilium/tetragon/blob/main/cmd/tetra/tracingpolicy/tracingpolicy.go. Could you also do that?

Sure. Let me do the changes and get back to you. Thank you very much!

kkourt:
> > Thanks, changes LGTM. I left one comment about removing the unneeded `name` field.
> > One of the things we also need to do is update the CLI: https://github.com/cilium/tetragon/blob/main/cmd/tetra/tracingpolicy/tracingpolicy.go. Could you also do that?
> 
> Sure. Let me do the changes and get back to you. Thank you very much!

Thanks! I would suggest doing the CLI updates in a different patch to ease reviewing.

joshuajorel:
@kkourt - after checking https://github.com/cilium/tetragon/blob/main/cmd/tetra/tracingpolicy/tracingpolicy.go, is there value in changing the behavior of the CLI? Seems like the CLI was meant for Linux based deployments, and you don't use `TracingPolicyNamespaced` for that. Otherwise, would adding a namespace argument to the `delete`, `enable`, and `disable` command suffice?

pkg/sensors/handler.go
```diff
@@ -96,13 +96,13 @@ func (h *handler) updatePolicyFilter(tp tracingpolicy.TracingPolicy, tpID uint64
 func (h *handler) addTracingPolicy(op *tracingPolicyAdd) error {
 	// allow overriding existing policy collection that resulted in an error
 	// during the loading state
-	if col, exists := h.collections[op.name]; exists && col.state != LoadErrorState {
-		return fmt.Errorf("failed to add tracing policy %s, a sensor collection with the name already exists", op.name)
+	if col, exists := h.collections[op.ck]; exists && col.state != LoadErrorState {
+		return fmt.Errorf("failed to add tracing policy %s, a sensor collection with the name already exists", op.ck)
```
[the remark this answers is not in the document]
joshuajorel: Change wording to reference key instead of name

pkg/sensors/manager.go
joshuajorel: Deleted `name` in the struct

pkg/sensors/manager.go
joshuajorel: Deleted `name` in the struct

pkg/sensors/manager.go
joshuajorel: Deleted `name` in the struct

cmd/tetra/tracingpolicy/tracingpolicy.go
```diff
@@ -48,15 +48,21 @@ func New() *cobra.Command {
 	}
 
 	tpDelCmd := &cobra.Command{
-		Use:   "delete <name>",
+		Use:   "delete <name> <namespace>",
```
mtardy, line 51: UX nit: it feels like we could use the common convention using a flag `-n <namespace>` or `--namespace <namespace>` instead of this. Like with `kubectl delete` or any `kubectl` cmd.

The slight difference is that the _implied_ namespace usually is the namespace of the current context while here it would be "global namespace" which means the cluster-wide resource.

kkourt: I don't have a strong opinion, but I'd agree with @mtardy that `--n|--namespace` is more idiomatic.

kkourt (approved):
Thanks, LGTM!

PR #2339, opened 2024-04-17
k8s: Fix CRDs Go code generation

lambdanis:
Commit a9c509f1f7c738bbeb1565f83eb9da70941a22ba replaced execution of deprecated scripts with
k8s.io/code-generator/kube_codegen.sh. However, kube_codegen.sh doesn't generate
code on its own, but rather defines helper functions. To migrate to
kube_codegen.sh we'll need to call these functions from another script.
Additionally, it'll change with version 1.30 of k8s libraries - that's why I
opted to simply revert the change for now. When k8s libraries are updated to
1.30, the code generation scripts should be updated too.

### DIFF
--- a/pkg/k8s/Makefile
+++ b/pkg/k8s/Makefile
@@ -21,10 +21,10 @@ generate:
 .PHONY: __do_generate
 __do_generate:
 	./tools/controller-gen crd paths=./apis/... output:dir=apis/cilium.io/client/crds/v1alpha1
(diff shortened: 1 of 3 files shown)


mtardy (approved):
Thanks Anna!

mtardy:
I can create an issue so that we don't forget to update, to note that it's marked as deprecated if you want :)

lambdanis:
@mtardy sure, feel free to open an issue. K8s dependencies are generally updated together with cilium/cilium, right?

PR #2340, opened 2024-04-17
Add export metrics

chancez:
This is useful to figure out how much data is being exported and if the log pipeline is behind in processing the incoming events.

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -713,7 +713,9 @@ func startExporter(ctx context.Context, server *server.Server) error {
 		}()
 	}
 
-	encoder := encoder.NewProtojsonEncoder(writer)
+	// Track how many bytes are written to the event export location
+	encoderWriter := exporter.NewExportedBytesTotalWriter(writer)
+	encoder := encoder.NewProtojsonEncoder(encoderWriter)
 	var rateLimiter *ratelimit.RateLimiter
 	if option.Config.ExportRateLimit >= 0 {
 		rateLimiter = ratelimit.NewRateLimiter(ctx, 1*time.Minute, option.Config.ExportRateLimit, encoder)
--- a/pkg/exporter/exporter.go
+++ b/pkg/exporter/exporter.go
@@ -58,9 +58,12 @@ func (e *Exporter) Send(event *tetragon.GetEventsResponse) error {
 		e.rateLimiter.Drop()
 		return nil
 	}
+
 	if err := e.encoder.Encode(event); err != nil {
 		logger.GetLogger().WithError(err).Warning("Failed to JSON encode")
 	}
+	eventsExportedTotal.Inc()
+	eventsExportTimestamp.Set(float64(event.GetTime().GetSeconds()))
 	return nil
 }
 
--- a/pkg/exporter/metrics.go
+++ b/pkg/exporter/metrics.go
@@ -1,0 +1,56 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package exporter
+
+import (
+	"io"
+
+	"github.com/cilium/tetragon/pkg/metrics/consts"
+	"github.com/prometheus/client_golang/prometheus"
+)
+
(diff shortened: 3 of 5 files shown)


cmd/tetragon/main.go
```diff
@@ -713,7 +713,9 @@ func startExporter(ctx context.Context, server *server.Server) error {
 		}()
 	}
 
-	encoder := encoder.NewProtojsonEncoder(writer)
+	// Track how many bytes are written to the event export location
+	encoderWriter := exporter.NewExportdBytesTotalWriter(writer)
```
chancez, line 717: I would have rather handled this in the exporter, but without the ability to wrap the writer before it goes into the encoder, I can't. I didn't want to do a big refactor either, so this is probably the simplest approach.

pkg/exporter/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+		Name:      "events_exported_bytes_total",
+		Help:      "Number of bytes exported for events",
+	})
+
+	eventsExportTimestamp = prometheus.NewGauge(prometheus.GaugeOpts{
+		Namespace: consts.MetricsNamespace,
+		Name:      "events_last_exported_timestamp",
+		Help:      "Timestamp of the most recent event to be exported",
+	})
+)
```
olsajiri, line 31: what's the use case for this one? need to check if the export is lagging?

pkg/exporter/exporter.go
```diff
@@ -58,9 +58,12 @@ func (e *Exporter) Send(event *tetragon.GetEventsResponse) error {
 		e.rateLimiter.Drop()
 		return nil
 	}
+
 	if err := e.encoder.Encode(event); err != nil {
 		logger.GetLogger().WithError(err).Warning("Failed to JSON encode")
 	}
+	eventsExportedTotal.Inc()
```
olsajiri, line 65: AFAICS Write is called for event, so this could be called from the byteCounterWriter.Write?

pkg/exporter/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+	registry.MustRegister(eventsExportTimestamp)
+}
+
+func newExportedBytesCounterWriter(w io.Writer, c prometheus.Counter) io.Writer {
+	return byteCounterWriter{Writer: w, bytesWritten: c}
+}
+
+type byteCounterWriter struct {
+	io.Writer
+	bytesWritten prometheus.Counter
```
olsajiri, line 45: do we need to pass the metrics like that? can't we just increment the eventsExportedBytesTotal directly?

pkg/exporter/metrics.go
chancez: Exactly.

pkg/exporter/exporter.go
chancez: Depending on the underlying encoder and writer and the event being encoded, it may take multiple writes for a single event, so that would be inaccurate.

pkg/exporter/metrics.go
chancez: If we wanted this to be a global singleton implementation, perhaps. However that would likely lead to mistakes later if the same type was used somewhere else.

pkg/exporter/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+	registry.MustRegister(eventsExportedTotal)
+	registry.MustRegister(eventsExportedBytesTotal)
+	registry.MustRegister(eventsExportTimestamp)
+}
+
+func newExportedBytesCounterWriter(w io.Writer, c prometheus.Counter) io.Writer {
+	return byteCounterWriter{Writer: w, bytesWritten: c}
+}
+
+type byteCounterWriter struct {
```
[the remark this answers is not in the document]
chancez: Yeah; that sounds like it would count the writes; but we're counting the number of bytes in the writes, not the writes themselves.

lambdanis (approved):

olsajiri (approved):

PR #2341, opened 2024-04-17
Makefile: Do not use $(MAKE) when calling make with `docker run`

chancez:
Using `$(MAKE)` will expand to the absolute path of `make` on the host, which can fail in docker if that path does not exist in the image.

Instead, just call `docker run ... make` without the $(MAKE) variable.


willfindlay (approved):

mtardy (approved):
Just learned about this today https://www.gnu.org/software/make/manual/make.html#MAKE-Variable, it seems that it makes sense in this case to just call `make`.

Issue #1976, opened 2024-01-15
Migrate e2e labels test from `isovalent/jobs-app`


mtardy:
The e2e labels test is suffering from flakes because the jobs-app sometimes fails to deploy, even though we provide enough resources, see https://github.com/cilium/tetragon/issues/1954. The idea would be to migrate from these deployments to something more lightweight and maintained to make this faster and more reliable (thus fix https://github.com/cilium/tetragon/issues/1954).

Our initial idea is to use ~~Google's microservices demo "Online Boutique" https://github.com/GoogleCloudPlatform/microservices-demo (this one doesn't deploy on arm64)~~ OpenTelemetry demo https://github.com/open-telemetry/opentelemetry-demo with the minimal footprint possible. 

The whole e2e labels test (https://github.com/cilium/tetragon/blob/main/tests/e2e/tests/labels/labels_test.go) will need to be adapted first by changing the Helm chart to be deployed:
https://github.com/cilium/tetragon/blob/004d9cfe1982614fa4e83a2e55edea2a20f37b56/tests/e2e/tests/labels/labels_test.go#L45-L49

And especially, the checkers, to suit the new deployments' labels:
https://github.com/cilium/tetragon/blob/004d9cfe1982614fa4e83a2e55edea2a20f37b56/tests/e2e/tests/labels/labels_test.go#L121-L184

prateek041:
Hey @mtardy what is the estimated deadline to resolve this issue ?

mtardy:
Hey, there is no deadline to resolve this issue.

prateek041:
Ok then, I'm gonna give this a shot. Assigning it to myself.

mtardy:
This is the detailed issue @jrfastab

rajesh-sampathrajan:
Hi  @mtardy , looks like this issue is still open?. If yes, may I pick up this issue?, please let me know

rajesamp:
working on resolving `undefined: unix.CBitFieldMaskBit34` -> even after upgrading sys/unix package to .18 and mod vendor, issue still exists.

PR #2176, opened 2024-03-02
<WIP> For issue#1976 - jobs-app is replaced with Google's online boutique app

rajesamp:
Hi 👋! 

This PR is ". 

Title : to address "TestLabelsDemoApp failures 

This patch fixes this by adding:
I have updated code reference of app - jobs-app to Google's Online Boutique in helm chart & labelsEventChecker which is lightweight and expected to run without failure.

Fixes: #1976

PR #2345, opened 2024-04-19
TestLabelsDemoApp: Replace isovalent/jobs-app by Opentelemetry demo app

Trung-DV:
Migrate e2e labels test from `isovalent/jobs-app` to Opentelemetry demo app
- Update the labels checker
- remove `-p 1 -parallel 1` flags when running tests

Fixes: #1976, #2176

### DIFF
--- a/tests/e2e/tests/labels/labels_test.go
+++ b/tests/e2e/tests/labels/labels_test.go
@@ -32,7 +32,7 @@ const (
 func installDemoApp(labelsChecker *checker.RPCChecker) features.Func {
 	return func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
 		manager := helm.New(c.KubeconfigFile())
-		if err := manager.RunRepo(helm.WithArgs("add", "isovalent", "https://helm.isovalent.com")); err != nil {
+		if err := manager.RunRepo(helm.WithArgs("add", "open-telemetry", "https://open-telemetry.github.io/opentelemetry-helm-charts")); err != nil {
 			t.Fatalf("failed to add helm repo: %s", err)
 		}
 
@@ -42,15 +42,17 @@ func installDemoApp(labelsChecker *checker.RPCChecker) features.Func {
 
 		for i := 0; i < demoAppRetry; i++ {
 			if err := manager.RunInstall(
-				helm.WithName("jobs-app"),
-				helm.WithChart("isovalent/jobs-app"),
-				helm.WithVersion("v0.7.1"),
+				helm.WithName("otel-demo"),
+				helm.WithChart("open-telemetry/opentelemetry-demo"),
+				helm.WithVersion("0.30.3"),
 				helm.WithNamespace(namespace),
-				helm.WithArgs("--create-namespace", "--wait"),
+				helm.WithArgs("--create-namespace"),
+				helm.WithWait(),
 			); err != nil {
 				labelsChecker.ResetTimeout()
 				t.Logf("failed to install demo app. run with `-args -v=4` for more context from helm: %s", err)
 			} else {
+				t.Log("demo app install successfully")
 				return ctx
 			}
 		}
@@ -64,7 +66,7 @@ func uninstallDemoApp() features.Func {
 	return func(ctx context.Context, t *testing.T, c *envconf.Config) context.Context {
 		manager := helm.New(c.KubeconfigFile())
 		if err := manager.RunUninstall(
-			helm.WithName("jobs-app"),
+			helm.WithName("otel-demo"),
 			helm.WithNamespace(namespace),
 		); err != nil {
 			t.Fatalf("failed to uninstall demo app. run with `-args -v=4` for more context from helm: %s", err)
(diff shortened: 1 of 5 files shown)


mtardy (commented):
Hey thanks a lot for doing this work, I think the patch works, failures are unrelated. Could you maybe squash your commits into one? 

I need to take a deeper look to see if we can make the label a bit more explicit maybe (add more of them if they are any) and test your patch locally in the meantime.

mtardy:
> make the label a bit more explicit maybe (add more of them if they are any)

Those are the only ones set by the default helm chart install, so let's keep this it's plenty enough and it simplifies how they are checked!

tests/e2e/tests/labels/labels_test.go
```diff
@@ -32,21 +33,13 @@ const (
...
-				helm.WithName("jobs-app"),
-				helm.WithChart("isovalent/jobs-app"),
-				helm.WithVersion("v0.7.1"),
-				helm.WithNamespace(namespace),
-				helm.WithArgs("--create-namespace", "--wait"),
+			if err := manager.RunUpgrade(
+				helm.WithArgs("onlineboutique"),
+				helm.WithArgs("oci://us-docker.pkg.dev/online-boutique-ci/charts/onlineboutique"),
+				helm.WithArgs("--install"),
+				helm.WithArgs("--create-namespace", "-n", namespace),
```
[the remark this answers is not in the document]
Trung-DV: Thank you @mtardy :D

Trung-DV:
> > make the label a bit more explicit maybe (add more of them if they are any)
> 
> Those are the only ones set by the default helm chart install, so let's keep this it's plenty enough and it simplifies how they are checked!

As mentioned in https://github.com/cilium/tetragon/pull/2176#issuecomment-2063497524, I replaced labels to be checked by looked into https://github.com/GoogleCloudPlatform/microservices-demo/tree/90b81cca772eb61b8695a638e44b9b2cde70853f/helm-chart/templates and saw only the app label :D 

So just let it simple like this, right? @mtardy :D

tests/e2e/tests/labels/labels_test.go
```diff
@@ -11,13 +11,14 @@ import (
...
+	"sigs.k8s.io/e2e-framework/pkg/features"
+	"sigs.k8s.io/e2e-framework/third_party/helm"
+
 	sm "github.com/cilium/tetragon/pkg/matchers/stringmatcher"
 	"github.com/cilium/tetragon/tests/e2e/checker"
 	"github.com/cilium/tetragon/tests/e2e/helpers"
 	"github.com/cilium/tetragon/tests/e2e/runners"
-	"sigs.k8s.io/e2e-framework/pkg/envconf"
-	"sigs.k8s.io/e2e-framework/pkg/features"
-	"sigs.k8s.io/e2e-framework/third_party/helm"
```
mtardy, line 20: and I'm sorry this is just a nit but could you remove this change? I don't think that's a linter that did that and it looks like we could avoid this little thing.

mtardy: And please suqashh the commits so that's there only one in the end :)

mtardy:
> As mentioned in [#2176 (comment)](https://github.com/cilium/tetragon/pull/2176#issuecomment-2063497524), I replaced labels to be checked by looked into https://github.com/GoogleCloudPlatform/microservices-demo/tree/90b81cca772eb61b8695a638e44b9b2cde70853f/helm-chart/templates and saw only the app label :D
> 
> So just let it simple like this, right? @mtardy :D

This is perfect like this, thanks a lot!

tests/e2e/tests/labels/labels_test.go
mtardy: aah okay I thought `goimports` was what most editors were using.

mtardy:
So we have a failure on e2e tests but I think it's unrelated https://github.com/cilium/tetragon/actions/runs/8752369371/job/24019870334?pr=2345#step:7:1061. Since all the checkers have matched, I don't think it's related to your changes. Let me just investigate a bit before merging this.

mtardy:
By the way, it seems the deployment time and checker time is a ~50% (or a bit more) gain from before, that's great for CI speed.

Trung-DV:
> By the way, it seems the deployment time and checker time is a ~50% (or a bit more) gain from before, that's great for CI speed.

Relate to the CI speed, I have a concern why we need `-p 1 -parallel 1` flags here https://github.com/cilium/tetragon/blob/main/Makefile#L257? 
As I understand, `-p 1` means only 1 build command (for the test package) can run at a time, `-parallel 1` means only 1  execution of test functions that call `t.Parallel` running.

An with my experience, the `-parallel 1` seems redundant with `-p 1`. In case we want to make sure the tests do not run in parallel, we can set the `-parallel 1` and let the compiler build the test packages in parallel. That could reduce the waiting time when running testing (https://github.com/cilium/tetragon/actions/runs/8752369371/job/24024402667?pr=2345#step:7:10 to https://github.com/cilium/tetragon/actions/runs/8752369371/job/24024402667?pr=2345#step:7:12)

mtardy:
> So we have a failure on e2e tests but I think it's unrelated https://github.com/cilium/tetragon/actions/runs/8752369371/job/24019870334?pr=2345#step:7:1061. Since all the checkers have matched, I don't think it's related to your changes. Let me just investigate a bit before merging this.

So in the end, even if the output is very confusing, it was related to this. Unfortunately the [microservices-demo](https://github.com/GoogleCloudPlatform/microservices-demo) does not ship arm64 images for all the microservices and it seems now it builds okay, they just don't want to publish the image for some reason https://github.com/GoogleCloudPlatform/microservices-demo/issues/622.

It worked fine locally on my computer since I use a rosetta share between my OS and the Linux VM I use and didn't realize the arch issue...

mtardy:
So yeah again I'm sorry, you did a really good job and helped us a lot on something we have been postponing forever and you just bumped into this not being the ideal solution in the end.

At least now we know it, and we can see what actions should be taken:
- we can see if Google moves quickly on publishing the arm64 images;
- or if we should find an alternative real-life deployment;
- or just use something made-up that is simpler.

mtardy (changes requested):
So yeah again I'm sorry, you did a really good job and helped us a lot on something we have been postponing forever and you just bumped into this not being the ideal solution in the end.

At least now we know it, and we can see what actions should be taken:
- we can see if Google moves quickly on publishing the arm64 images;
- or if we should find an alternative real-life deployment;
- or just use something made-up that is simpler.

Let me chat about that with the rest of the team and see what we can do! Thanks again for your PR :)

Trung-DV:
Thank @mtardy for supporting me :D. While waiting for support from Google microservices demo, let me check that may we reduce the `go test` time 💪

mtardy:
Hey, @lambdanis suggested I use this demo instead https://github.com/open-telemetry/opentelemetry-demo, apparently, it's a fork of the Google microservice demo and I tested it on arm64, and it works fine! You can find the doc here https://opentelemetry.io/docs/demo/kubernetes-deployment/.

If you still have the motivation to switch the e2e tests to use this one it would be amazing, otherwise, I can update the issue with the new desired deployment and someone can pick this up.

Trung-DV:
> Hey, @lambdanis suggested we use this demo instead https://github.com/open-telemetry/opentelemetry-demo, apparently, it's a fork of the Google microservice demo and I tested it on arm64, and it works fine! You can find the doc here https://opentelemetry.io/docs/demo/kubernetes-deployment/.
> 
> 
> 
> If you still have the motivation to switch the e2e tests to use this one it would be amazing, otherwise, I can update the issue with the new desired deployment and someone can pick this up.

Thank you for your suggestion. Let me change the Helm charts, that would be a great start for me 💪

Trung-DV:
Hi @mtardy , I have update the `labelsEventChecker` to match the otel-demo.
Since we have split e2e-tests to run separately, I suggest removing `-p 1 -parallel 1` so that the go's compiler can compile the test package in parallel (which is limited by -p 1)

jrfastab (approved):

jrfastab:
LGTM we can follow up with the -p * if we need/want to.

mtardy:
Lovely thanks!

PR #2348, opened 2024-04-19
tetra: no longer show singular getevents cmd flags as deprecated 

mtardy:
For some reason that I'm not sure I remember, I deprecated the singular
'--namespace', '--process' and '--pod' in https://github.com/cilium/tetragon/commit/23983ab2170650372558c2f73d78c80f77f37b3b. It's a bit annoying
to get those deprecation message when running the docs examples. Also, I
don't think that makes sense, especially when you are writing things
like:

	tetra getevents --namespace value1 --namspace value2

We still keep both singular and plural working but advertize the
singular so that it's easier to type.


olsajiri (approved):
lgtm

PR #2354, opened 2024-04-21
Tetragon e2e Test: Run e2e tests in parallel on multiple runners

Trung-DV:
[E2e tests](https://github.com/cilium/tetragon/tree/main/tests/e2e/tests) currently have 3 packages that can run independently with each other

By split the test into 3 runners, we can reduce about ~50% CI time: https://github.com/Trung-DV/tetragon/actions/runs/8771741177

### DIFF
--- a/.github/workflows/run-e2e-tests.yaml
+++ b/.github/workflows/run-e2e-tests.yaml
@@ -18,13 +18,35 @@ env:
   ciliumCliVersion: v0.15.0
 
 jobs:
-  e2e-test:
+  list-e2e-pkg:
+    runs-on: ubuntu-22.04
+    outputs:
+      packages: ${{ steps.set-packages.outputs.packages }}
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+    - name: Install Go
+      uses: actions/setup-go@0c52d547c9bc32b1aa3301fd7a9cb496313a4491 # v5.0.0
+      with:
+        go-version-file: go.mod
+
+    - id: set-packages
+      name: Listing e2e tests packages
+      shell: bash
+      run: |
+        echo PACKAGES=$(make ls-e2e-test | jq -Rnc '[inputs | {"s": split("/")[-1], "f":.}]') | tee -a $GITHUB_STEP_SUMMARY | tee -a $GITHUB_OUTPUT
+
+  run-e2e-test:
+    needs: list-e2e-pkg
     runs-on: ${{ matrix.os }}
     timeout-minutes: 40
+    name: ${{matrix.os}} / ${{ matrix.package.s }}
     strategy:
       fail-fast: false
       matrix:
         os: [ ubuntu-22.04, actuated-arm64-4cpu-16gb ]
+        package: ${{fromJson(needs.list-e2e-pkg.outputs.packages)}}
     steps:
     - name: Checkout Code
       uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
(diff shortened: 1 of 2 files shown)


kkourt:
Hi,

Thanks for the PR, but it's not clear to me that this will work as expected.

Each e2e test requires its own Tetragon instance, and having multiple Tetragon instances at the same time can be problematic. This is why we generally running the tests sequentially.

Trung-DV:
> Hi,
> 
> Thanks for the PR, but it's not clear to me that this will work as expected.
> 
> Each e2e test requires its own Tetragon instance, and having multiple Tetragon instances at the same time can be problematic. This is why we generally running the tests sequentially.

Hi @kkourt 
As I understand, currently, we run e2e test sequentially, we will init a Tetragon instance for each test package at a time.
So I just split out the runners, init a Tetragon instance for each runner, and each runner will a single test package.
All are independent environments (runners). 
Is my approach reasonable?

mtardy (commented):
That's a good idea thanks! The only downside I could see is that your workflow might require lots of runners and sometimes it's a bottleneck when GitHub or actuated doesn't provide us with enough runners. But that should be fine.

Is there a way we could make this easier to see which is which?
image

Perhaps you could use `jobs.<job_id>.name` with the last folder name of the matrix path, or the entire path.

Like:
```yaml
run-e2e-test:
  name: ${{ <your path> }}
```

Trung-DV:
I have updated
now it looks like https://github.com/Trung-DV/tetragon/actions/runs/8799449115

image

.github/workflows/run-e2e-tests.yaml
```diff
@@ -18,13 +18,36 @@ env:
...
+      with:
+        go-version-file: go.mod
+
+    - id: set-packages
+      name: Listing e2e tests packages
+      shell: bash
+      run: |
+        echo PACKAGES=$(make ls-e2e-test | jq -Rnc '[inputs | {"s": split("/")[-1], "f":.}]') | tee -a $GITHUB_STEP_SUMMARY | tee -a $GITHUB_OUTPUT
+
+
```
mtardy, line 40: nit, double empty line here

mtardy (approved):
Thanks I like it, I think it's a great improvement!

mtardy:
In addition to my previous comments, if you can squash your commits maybe! Thanks!

.github/workflows/run-e2e-tests.yaml
```diff
@@ -18,13 +18,36 @@ env:
...
+  run-e2e-test:
+    needs: list-e2e-pkg
     runs-on: ${{ matrix.os }}
     timeout-minutes: 40
+    name: '${{matrix.os}}-${{ matrix.package.s }}'
```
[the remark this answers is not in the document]
Trung-DV: I have updated

.github/workflows/run-e2e-tests.yaml
Trung-DV: I have updated

mtardy:
Please @willfindlay take a look and feel free to merge this!

.github/workflows/run-e2e-tests.yaml
```diff
@@ -18,13 +18,35 @@ env:
...
+      name: Listing e2e tests packages
+      shell: bash
+      run: |
+        echo PACKAGES=$(make ls-e2e-test | jq -Rnc '[inputs | {"s": split("/")[-1], "f":.}]') | tee -a $GITHUB_STEP_SUMMARY | tee -a $GITHUB_OUTPUT
+
+  run-e2e-test:
+    needs: list-e2e-pkg
     runs-on: ${{ matrix.os }}
     timeout-minutes: 40
+    name: '${{matrix.os}}/${{ matrix.package.s }}'
```
willfindlay, line 44:
```suggestion
    name: '${{matrix.os}} / ${{ matrix.package.s }}'
```

To match the way it's formatted ion our other workflows: https://github.com/cilium/tetragon/actions/runs/8801173979/workflow?pr=2354#L91

willfindlay (approved):
Looks amazing, I just have one nitpick.

.github/workflows/run-e2e-tests.yaml
Trung-DV: I have updated 💯

mtardy:
Thanks!!

alexellis:
How many more jobs would you expect this to generate @mtardy?

Anything that could generate dozens or hundreds of jobs will probably be misusing ["the commons"](https://en.wikipedia.org/wiki/Tragedy_of_the_commons) resource, if it's a couple more that is unlikely to make much difference.

mtardy:
> How many more jobs would you expect this to generate @mtardy?
> 
> Anything that could generate dozens or hundreds of jobs will probably be misusing ["the commons"](https://en.wikipedia.org/wiki/Tragedy_of_the_commons) resource, if it's a couple more that is unlikely to make much difference.

You are everywhere 👀 , it just transforms a 20min job into three jobs that are running significantly faster (something around 10min each) so we consume a bit more generally but also faster in a way. It should be reasonable as it asks for 4 CPUs and 16GB. We have been seeing bottlenecks on some occasions on the providing on actuated side, we hope our consumption is reasonable otherwise, we can find solutions.

alexellis:
We've had a couple of users starting matrix builds with 250+ jobs .. this of course causes some issues for the shared service.

What you're doing sounds fine.

You can right-size VMs by running vmmeter - https://actuated.dev/blog/right-sizing-vms-github-actions

PR #2355, opened 2024-04-22
Tetragon Go Test: Reduce CI time, run test packages in parallel

Trung-DV:
Currently, we run tests sequentially, each package by each process. We can increase the parallelism of the test by removing the `-p 1` flag, so that we can run multiple test packages at a time. 
Furthermore, since there is no `t.Parallel()` call in tests, we can now remove `-parallel 1` flags also.

By moving the order of test packages, the [slowest test](https://github.com/cilium/tetragon/actions/runs/8757182496/job/24035287586#step:8:167) to run first, I hope this change can reduce `Tetragon Go Test` from around [~20min](https://github.com/cilium/tetragon/actions/runs/8757182496) to [~15min](https://github.com/Trung-DV/tetragon/actions/runs/8773573293)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -225,7 +225,7 @@ clean: cli-clean tarball-clean
 
 .PHONY: test
 test: tester-progs tetragon-bpf
-	$(GO) test -exec "$(SUDO)" -p 1 -parallel 1 $(GOFLAGS) -gcflags=$(GO_BUILD_GCFLAGS) -timeout $(GO_TEST_TIMEOUT) -failfast -cover ./pkg/... ./cmd/... ./operator/... ${EXTRA_TESTFLAGS}
+	$(GO) test -exec "$(SUDO)" $(GOFLAGS) -gcflags=$(GO_BUILD_GCFLAGS) -timeout $(GO_TEST_TIMEOUT) -failfast -cover ./pkg/sensors/tracing/... ./pkg/... ./cmd/... ./operator/... ${EXTRA_TESTFLAGS}
 
 .PHONY: bench
 bench:


Trung-DV:
Hi @mtardy, could you take a look in case this PR can reduce CI time?

mtardy (commented):
So running our tests in parallel could produce false positives maybe. In many tests we assume only one Tetragon agent is running on the kernel and only the events from the tests exist which might not be correct if we start parallelize.

We could split the tests that actually load tetragon and run them in parallel to the other tests that don't (that may be what's done here) but we would need to make sure it's the case.

Trung-DV:
> So running our tests in parallel could produce false positives maybe. In many tests we assume only one Tetragon agent is running on the kernel and only the events from the tests exist which might not be correct if we start parallelize.
> 
> We could split the tests that actually load tetragon and run them in parallel to the other tests that don't (that may be what's done here) but we would need to make sure it's the case.

Hmm understand. The main point of this PR is to remove `-p 1`, this allows us to run multiple test packages at the same time. 
Since we don't use `t.Parallel()` in tests, I think we can remove the flags `-parallel 1`. This is optional :D

mtardy:
> The main point of this PR is to remove `-p 1`, this allows us to run multiple test packages at the same time.

yep I get it, I would just need to check if the only package that starts a tetragon instance is the tracing package. Otherwise, we might get false results.

Trung-DV:
> yep I get it, I would just need to check if the only package that starts a tetragon instance is the tracing package. Otherwise, we might get false results.

Ah right, :D In this case, I think we can move `tracing` into e2e test or sth like that

kevsecurity:
> In this case, I think we can move `tracing` into e2e test or sth like that

We aren't moving unit tests into e2e tests.

The tests may well involve running up Tetragon – only one instance can run at once due to the way we pin programs and maps (plus probably other assumptions of singular use). If we make this change we prevent every other package test from running Tetragon which seems unacceptable.

I vote for rejecting this PR.

kkourt:
We could potentially split the unit tests (indeed, we have done it in the past so some [support](https://github.com/cilium/tetragon/commit/476ab862502c7a6ac92aebb1587e7c5a0b95c564) for this already exists) and have them run in different GH jobs as was done in https://github.com/cilium/tetragon/pull/2354.

That being said, I agree with @kevsecurity that this PR is not the right approach to do this.

Trung-DV:
Thank you for your comment. After some consideration and evaluation, I think the benefits of this PR are too small to change. So I would like to close this PR :D

PR #2377, opened 2024-04-25
docs: explicitly state which regex flavour is used for redaction

f1ko:


mtardy (approved):
Sure thanks! I just requested William's review just in case.

PR #2385, opened 2024-04-26
Dockerfile: Remove syntax annotation

lambdanis:
It doesn't seem helpful, but it's being updated by Renovate, what causes extra PR and CI noise.


mtardy (approved):
thanks

PR #2393, opened 2024-04-29
Starting 1.2 development.

kkourt:


kkourt:
The Check CRD version can be ignored:
> CustomResourceDefinitionSchemaVersion in pkg/k8s/apis/cilium.io/v1alpha1/version.go to be updated only in case of modifying the files pkg/k8s/apis/cilium.io/client/crds/v1alpha1/*.yaml

We are updating the number because we are starting 1.2.

lambdanis (approved):

PR #2399, opened 2024-04-29
bpf: use CORE for execve hook

kkourt:
Users reported an issue on RHEL 9 with binary names being wrong.

Looking at the sched/sched_process_exec tracepoint we hook into:

```
  $ cat /sys/kernel/debug/tracing/events/sched/sched_process_exec/format
  name: sched_process_exec
  ID: 310
  format:
          field:unsigned short common_type;       offset:0;       size:2; signed:0;
          field:unsigned char common_flags;       offset:2;       size:1; signed:0;
          field:unsigned char common_preempt_count;       offset:3;       size:1; signed:0;
          field:int common_pid;   offset:4;       size:4; signed:1;
          field:unsigned char common_preempt_lazy_count;  offset:8;       size:1; signed:0;

          field:__data_loc char[] filename;       offset:12;      size:4; signed:1;
          field:pid_t pid;        offset:16;      size:4; signed:1;
          field:pid_t old_pid;    offset:20;      size:4; signed:1;
 ```

There is an additional argument: common_preempt_lazy_count, which means that the struct we use (sched_execve_args) is no longer valid.

This patch removes the above struct, and instead, uses CORE and the trace_event_raw_sched_process_exec struct.

Reproduced and patch tested locally on a Rocky Linux 5.3 with a 5.14.0-362.24.1.el9_3.0.1.x86_64  kernel.

Fixes: https://github.com/cilium/tetragon/issues/1707

### DIFF
--- a/bpf/lib/generic.h
+++ b/bpf/lib/generic.h
@@ -62,15 +62,4 @@ static inline __attribute__((always_inline)) size_t generic_kprobe_common_size()
 	return offsetof(struct msg_generic_kprobe, args);
 }
 
-/* tracepoint args */
-struct sched_execve_args {
-	unsigned short common_type;
-	unsigned char common_flags;
-	unsigned char common_preempt_count;
-	int common_pid;
-	int filename;
-	int pid;
-	int old_pid;
-};
-
 #endif // _GENERIC__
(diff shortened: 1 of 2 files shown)


olsajiri (approved):
nice catch!

kkourt:
Marked it for backporting

mtardy:
oh wow, nice catch indeed that could be the issue, and finally fixing this long-standing issue.

inliquid:
Is it in 1.1 already?

kkourt:
> Is it in 1.1 already?

Not yet. Will do the backport soon.

kkourt:
backoprt: https://github.com/cilium/tetragon/pull/2468

Issue #2403, opened 2024-05-02
Enable users to add namespace exceptions in case OCI hook fails


f1ko:
When using Tetragon with the OCI hook feature enabled (`ociHookSetup.enabled: true`) all Pod creations except for those in `kube-system` are prevented when the agent is down.

This can lead to an outage for business-critical applications that are deployed in other namespaces.

Users shall be able to pass a list of namespaces that shall be excluded in a similar fashion as is the case for `kube-system`.

Issue #2402, opened 2024-05-02
Tetragon with ociHookSetup.enabled cannot start in a namespace other than kube-system


f1ko:
## Description

When using the OCI hook feature, an exception for all Pods in the `kube-system` namespace are added so that Tetragon itself can start as the hook has a dependency on Tetragon.
However, this results in a deadlock scenario if Tetragon is being deployed in any other namespace.

---

## Reproduction

Install Tetragon with the oci-hook feature enabled inside another namespace (i.e. not in `kube-system`):
```
kubectl create ns tetragon
helm install --namespace tetragon \
        --set tetragonOperator.image.override=localhost/cilium/tetragon-operator:latest \
        --set tetragon.image.override=localhost/cilium/tetragon:latest  \
        --set tetragon.grpc.address="unix:///var/run/cilium/tetragon/tetragon.sock" \
        --set tetragon.ociHookSetup.enabled=true \
        tetragon ./install/kubernetes/tetragon
```

The init container starts as expected and configures the oci-hook.
However, this leads to the agent never being able to start as the oci-hook cannot reach the agent and the only exception being Pods in the `kube-system` namespace:
```
$ kubectl describe pod -n tetragon tetragon-tctms
[...]
Events:
  Type     Reason     Age                From               Message
  ----     ------     ----               ----               -------
  Normal   Scheduled  54s                default-scheduler  Successfully assigned tetragon/tetragon-tctms to minikube
  Normal   Pulling    53s                kubelet            Pulling image "localhost/cilium/tetragon:latest"
  Normal   Pulled     53s                kubelet            Successfully pulled image "localhost/cilium/tetragon:latest" in 11ms (11ms including waiting). Image size: 215976744 bytes.
  Normal   Created    53s                kubelet            Created container oci-hook-setup
  Normal   Started    53s                kubelet            Started container oci-hook-setup
  Warning  Failed     43s                kubelet            Error: container create failed: time="2024-04-29T11:36:18Z" level=error msg="runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: "
  Warning  Failed     33s                kubelet            Error: container create failed: time="2024-04-29T11:36:28Z" level=error msg="runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: "
[...]

PR #2404, opened 2024-05-03
oci-hook: allow users to set a list of namespace exceptions and definee default

f1ko:
Prior to this commit `kube-system` was always added as an exception in order for Pods in that namespace to be created in the event that the OCI hook was not able to reach the Tetragon agent. This leads to a deadlock scenario when Tetragon itself is not installed in `kube-system`.

Instead of always adding `kube-syste` this change will always add the namespace Tetragon is deployed in. In addition to that a user can now define additional namespaces as further exception. For example to ensure Pods in business-critical namespaces can still be created even if the OCI hook fails to reach the Tetragon agent.

---

Install Tetragon in a namespace other than `kube-system`:
```
helm install --namespace tetragon \
        --set tetragonOperator.image.override=localhost/cilium/tetragon-operator:latest \
        --set tetragon.image.override=localhost/cilium/tetragon:latest  \
        --set tetragon.grpc.address="unix:///var/run/cilium/tetragon/tetragon.sock" \
        --set tetragon.ociHookSetup.enabled=true \
        tetragon ./install/kubernetes/tetragon
```

```
$ kubectl get pods
NAME                                 READY   STATUS    RESTARTS   AGE
tetragon-46lr7                       2/2     Running   0          13s
tetragon-operator-7c8cc5964f-jzxdx   1/1     Running   0          13s
```

Fixes https://github.com/cilium/tetragon/issues/2402

---

Install helm with the new `failAllowNamespaces` values:
```
helm install --wait --namespace tetragon \
        --set tetragonOperator.image.override=localhost/cilium/tetragon-operator:latest \
        --set tetragon.image.override=localhost/cilium/tetragon:latest  \
        --set tetragon.grpc.address='unix:///var/run/cilium/tetragon/tetragon.sock' \
        --set tetragon.ociHookSetup.enabled=true \
        --set tetragon.ociHookSetup.failAllowNamespaces='business-critical-a\,business-critical-b' \
        tetragon ./install/kubernetes/tetragon
```

Uninstall Tetragon to trigger OCI hook error:
```
helm uninstall --namespace tetragon tetragon
```

Create various namespaces:
```
kubectl create ns business-critical-a
kubectl create ns business-critical-b
kubectl create ns business-critical-c
```

Create deployments in namespaces:
```
kubectl create deploy -n business-critical-a nginx --image docker.io/nginx
kubectl create deploy -n business-critical-b nginx --image docker.io/nginx
kubectl create deploy -n business-critical-c nginx --image docker.io/nginx
```

As can be seen Pods in `business-critical-a` and `business-critical-b` were created and are Running whereas `business-critical-c` wasn't part of the list and therefore continues to fail:
```
$ kubectl get pods -A -l app=nginx
NAMESPACE             NAME                     READY   STATUS                 RESTARTS   AGE
business-critical-a   nginx-68fd89949f-8h4xg   1/1     Running                0          12s
business-critical-b   nginx-68fd89949f-q7l2j   1/1     Running                0          12s
business-critical-c   nginx-68fd89949f-kjflf   0/1     CreateContainerError   0          12s
```

Fixes https://github.com/cilium/tetragon/issues/2403


f1ko:
I think it makes sense to add @kkourt as a reviewer as he knows the most about the OCI hook implementation at this point.

install/kubernetes/tetragon/values.yaml
```diff
@@ -200,6 +200,8 @@ tetragon:
     # "oci-hooks" (https://github.com/containers/common/blob/main/pkg/hooks/docs/oci-hooks.5.md).
     interface: "oci-hooks"
     installDir: "/opt/tetragon"
+    # -- Comma-separated list of namespaces to allow Pod creation for, in case tetragon-oci-hook fails to reach Tetragon agent.
+    failAllowNamespaces: ""
```
kkourt, line 204: Can we document that if this is left empty, the namespace where tetragon is installed will be used by default?

kkourt (approved):
Thanks, LGTM.
I do have a small suggestion, but I'm pre-approving since it's a trivial change.

install/kubernetes/tetragon/templates/_container_oci_hooks_setup.tpl
```diff
@@ -11,6 +11,7 @@
     - --local-install-dir={{  include "container.tetragonOCIHookSetup.installPath" . }}
     - --host-install-dir={{ .Values.tetragon.ociHookSetup.installDir }}
     - --oci-hooks.local-dir={{ include "container.tetragonOCIHookSetup.hooksPath" . }}
+    - --oci-hooks.fail-allow-namespaces={{ if .Values.tetragon.ociHookSetup.failAllowNamespaces }}{{ printf "%s,%s" .Release.Namespace .Values.tetragon.ociHookSetup.failAllowNamespaces }}{{ else }}{{ .Release.Namespace }}{{ end }}
```
kkourt, line 14: Actually, one more comment: Can we always add Tetragon's namespace here? If it's not there, it is certainly an issue.

kkourt (changes requested):
I'm changing my review to "Request changes", because of my later comment. It would be nice to have a way to always include the tetragon namespace in the failed namespaces list because otherwise Tetragon will not be able to start.

install/kubernetes/tetragon/values.yaml
f1ko: Solved via https://github.com/cilium/tetragon/pull/2404/commits/2424f81efbc2ef85b5301390c08eb9afb87048db

install/kubernetes/tetragon/templates/_container_oci_hooks_setup.tpl
f1ko: That's already the case in the current implementation:

If the user uses sets `failAllowNamespaces` to `my-namespace,your-namespace` then the line will read as `    - --oci-hooks.fail-allow-namespaces=my-namespace,your-namespace,tetragon` (assuming Tetragon is deployed in a namespaces called `tetragon`).

Otherwise if `failAllowNamespaces` is empty (as is the default), then the line will read as `    - --oci-hooks.fail-allow-namespaces=tetragon` (assuming Tetragon is deployed in a namespaces called `tetragon`).

So Tetragon's namespace is always added.

f1ko:
Thanks for the review!
Added a small hint to clarify that Tetragon's namespace is always added (even if the user leaves `failAllowNamespaces` empty).

As per your last comment this is already accounted for, see [this comment](https://github.com/cilium/tetragon/pull/2404#discussion_r1598504168).

install/kubernetes/tetragon/templates/_container_oci_hooks_setup.tpl
kkourt: Ah, right. Missed that part. Awesome, thanks!

kkourt (approved):
Thanks!

kkourt:
One final request: Can you squash the changes from my feedback into a single commit? (`git rebase --interactive` using the `squash` and `fixup` actions should help).

f1ko:
Done 👍

PR #2413, opened 2024-05-03
dev: Update vagrant configuration

sayboras:
## Description 

Pleases refer to individual commit for more details. 

## Testing

Testing was done locally with `make test` after VM provision as per below

```
$ make test
...
go test -exec "sudo" -p 1 -parallel 1  -gcflags= -timeout 20m -failfast -cover ./pkg/... ./cmd/... ./operator/... 
...
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,10 +1,51 @@
 $VM_MEMORY = (ENV['VM_MEMORY'] || 8912)
 $VM_CPUS = (ENV['VM_CPUS'] || 2)
+$GO_VERSION = (ENV['GO_VERSION'] || "1.22.0")
+
+$go_install = <<-'SCRIPT'
+# Install golang
+GO_VERSION=$1
+curl -O https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar.gz && \
+    rm -rf /usr/local/go \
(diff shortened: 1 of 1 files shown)


Vagrantfile
```diff
@@ -1,10 +1,51 @@
 $VM_MEMORY = (ENV['VM_MEMORY'] || 8912)
 $VM_CPUS = (ENV['VM_CPUS'] || 2)
+$GO_VERSION = (ENV['GO_VERSION'] || "1.22.0")
+
+$go_install = <<-'SCRIPT'
+# Install golang
+GO_VERSION=$1
+curl -O https://storage.googleapis.com/golang/go$GO_VERSION.linux-amd64.tar.gz && \
+    rm -rf /usr/local/go \
```
jrfastab, line 9: I don't think we  want to delete existing go on users. This is on the host?

Could we check if go version is correct first and then give a readable warning with instructions on how to do this? And/or prompt to replace?

mtardy (commented):
hey, just commenting since I think we don't use Vagrant for any of the CI tests. So it's a development thing I don't think one many people use. 

However, I have no objection to merge this since it might be well outdated but it's not well tested and thus not maintained so maybe we could discuss removing it?

sayboras:
Ah thanks both :+1:

PR #2431, opened 2024-05-14
Don't create PodInfo if the pod is being deleted

michi-covalent:
Check if the pod has DeletionTimestamp field set before reconciling to avoid the situation where pod deletion gets stuck because Tetragon operator recreates PodInfo during pod deletion.


jrfastab (approved):

kkourt (approved):

michi-covalent:
filed https://github.com/cilium/tetragon/issues/2434 for the go test failure. i'm shipping this.

michi-covalent:
backports:

- v1.1: https://github.com/cilium/tetragon/pull/2435
- v1.0: https://github.com/cilium/tetragon/pull/2436

Issue #2434, opened 2024-05-15
 Tetragon Go Test failing on actuated-arm64-4cpu-8gb: github.com/cilium/tetragon/pkg/sensors/tracing: TestTraceKernelModule


michi-covalent:
github: https://github.com/cilium/tetragon/actions/workflows/gotests.yml?query=event%3Apush
slack: https://cilium.slack.com/archives/C03EV7KJPJ9/p1715729652070269

kkourt:
@tixxdz you already seem to be working on this (https://github.com/cilium/tetragon/pull/2433) so I'm assigning to you.

PR #2435, opened 2024-05-15
Don't create PodInfo if the pod is being deleted

michi-covalent:
[ upstream commit be19aebaccbfc46217fc372e823a7c9ed925f92e ]

Check if the pod has DeletionTimestamp field set before reconciling to avoid the situation where pod deletion gets stuck because Tetragon operator recreates PodInfo during pod deletion.


michi-covalent:
`Tetragon Go Test / build (actuated-arm64-4cpu-8gb)` seems to be hitting #2434 https://github.com/cilium/tetragon/actions/runs/9097956603/job/25007044509?pr=2435

PR #2448, opened 2024-05-17
tetragon: cache username lookups

tixxdz:


pkg/reader/userdb/userdb.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"errors"
+	"os/user"
+	"strconv"
+	"sync"
+)
+
+// Max system users ID is 999, usually maintained by distributions
+// Reference: https://github.com/systemd/systemd/blob/main/docs/UIDS-GIDS.md#summary
+const MAX_SYSTEM_USERID = 999
```
kkourt, line 15: nit: let's use go syntax? MaxSystemUserID?

pkg/sensors/exec/exec.go
```diff
@@ -77,9 +76,8 @@ func msgToExecveAccountUnix(m *exec.MsgExecveEventUnix) {
 	if option.Config.UsernameMetadata == int(option.USERNAME_METADATA_UNIX) {
 		if ns, err := namespace.GetMsgNamespaces(m.Unix.Msg.Namespaces); err == nil {
 			if ns.Mnt.IsHost && ns.User.IsHost {
-				// use Golang user.LookupId() as we want to only parse /etc/passwd for now
-				if userInfo, err := user.LookupId(strconv.FormatUint(uint64(m.Unix.Process.UID), 10)); err == nil {
-					m.Unix.Process.User.Name = userInfo.Name
+				if username, err := userdb.UsersCache.LookupUser(m.Unix.Process.UID); err == nil {
+					m.Unix.Process.User.Name = username
```
kkourt, line 80: would be nice to have a metric here for errors.

kkourt (approved):
Thanks, overall looks good to me. I've left some minor comments.

olsajiri:
@tixxdz lookgs good, few questions

could you please provide more details in changelogs on why we want to do that, I guess it's faster, can we measure that?
also why do we cache just part of the user ids?

thanks

pkg/reader/userdb/userdb.go
tixxdz: fixed, thanks

pkg/sensors/exec/exec.go
tixxdz: Will do it separately, thanks

tixxdz:
> @tixxdz lookgs good, few questions
> 
> could you please provide more details in changelogs on why we want to do that, I guess it's faster, can we measure that? also why do we cache just part of the user ids?
> 
> thanks

I added the following commit logs:
```
    Cache UIDs/Usernames so we reduce:
    - I/O on /etc/passwd for each execution especially for root user and
      other system user IDs as by default lot of software runs under those.
    - Some users may watch /etc/passwd access, so caching those entries
      will reduce Tetragon generated events and improve performance overall.
    
    We only cache UIDs from 0..999 as these are considered standard
    system users, that are set by most distros packages. Distros maintainers
    usually do great job to not clash such users. For anything above > 999
    we do not cache it, we can that if users want such support later.
    
    Reference:
    https://github.com/systemd/systemd/blob/main/docs/UIDS-GIDS.md#summary
```

Thank you

PR #2452, opened 2024-05-20
tetragon: Add inline function macro

olsajiri:
we have all functions inline at the moment, unify the inline macro
as a preparation for using static programs in following changes

plus few other cleanups


bpf/alignchecker/bpf_alignchecker.c
```diff
@@ -2,6 +2,7 @@
 
 #include "include/vmlinux.h"
 #include "include/api.h"
+#include "func.h"
```
tixxdz, line 5: maybe rename this to tg-compiler.h or compiler.h if it passes?

tixxdz (approved):
lgtm, thanks

kkourt (approved):

PR #2457, opened 2024-05-22
release_template: also tag API module

kkourt:
Releasing multiple modules from the same repo, requires tags for each. Add a tag for the api/ modules in the release process.


mtardy (approved):
lgtm! 

aside, should we remove the `-s`? Is someone using that, should we start using it?

kkourt:
> lgtm!
> 
> aside, should we remove the `-s`? Is someone using that, should we start using it?

I think we should use it if we can, so let's keep it.

PR #2458, opened 2024-05-22
btf: print original error returned by ebpf btf.TypeByName()

tixxdz:
btf.TypeByName() could return ErrMultipleMatches if btf includes multiple matches.


olsajiri (approved):
looks good, could you please put example of the new log message into the changelog? thanks

tixxdz:
> looks good, could you please put example of the new log message into the changelog? thanks

Thanks just did and pushed

tixxdz:
Merging as CI error is not related

PR #2471, opened 2024-05-25
helm: Add tetragon.livenessProbe value

michi-covalent:
[ upstream commit 1871fe85f1fd5bf62846a861b9f995265cb214c9 ]

Add tetragon.livenessProbe Helm value that overrides the default liveness probe for the tetragon container. For example, to use grpc probe, you can specify tetragon.livenessProbe Helm value like this:

    tetragon:
      livenessProbe:
        grpc:
          port: 54321


mtardy (approved):
thanks again!

PR #2487, opened 2024-05-30
Document release notes process & introduce upgrade notes

lambdanis:
Documenting changes introduced in 1.1 release was not a trivial task, so here I'm attempting to improve the process around release & upgrade notes. There are two different aspects:
1. I documented the existing practices for `release-note` blurbs and labels in PRs (see `docs/content/en/docs/contribution-guide/release-notes.md`).
2. I want to introduce a practice of writing separate upgrade notes for changes that require attention when upgrading Tetragon. Initially I wanted to introduce a dedicated docs page, similar to the [one from Cilium](https://docs.cilium.io/en/latest/operations/upgrade/#version-specific-notes). However, I think keeping upgrade notes together with the regular release notes is a better solution, as many users at least skim through the release notes and Tetragon upgrade generally should be simple enough to not require a dedicated docs page. So instead I added an upgrade notes template to the `contrib` dir and updated the release process.

The first commit should be backported to the stable branches: v1.0 and v1.1.

Also I'm suggesting to stop using the `release-note/breaking-changes` label (it's not documented here) and highlight any potential breaking changes in the upgrade notes instead (with understanding that breaking changes should be avoided in the first place).

Pinging @cilium/tetragon on this one. All of this makes sense only if Tetragon core developers follow these practices, so please review and comment if you disagree with something.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_template.md
+++ b/.github/ISSUE_TEMPLATE/release_template.md
@@ -14,15 +14,22 @@ assignees: ''
 
       export RELEASE=v0.8.1
 
-- [ ] Open a pull request to update the Helm chart and docs version:
+- [ ] Open a pull request to update the Helm chart and docs:
 
       git checkout -b pr/prepare-$RELEASE
+
+      # update Helm chart
       ./contrib/update-helm-chart.sh $RELEASE
       make -C install/kubernetes
       git add install/kubernetes/tetragon/
-      # update hugo version
+
+      # update version in docs (Hugo config)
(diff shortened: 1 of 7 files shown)


jrfastab:
One nit from PR description. I'm all for not having breaking changes however we may continue to deprecate flags, fields, and metrics as we go. Protobuf definitions and JSON I would consider user facing e.g. in a database somewhere so we probably need to keep those going for a very long time before deprecating.

jrfastab (approved):

lambdanis:
I split script and release template updates into separate commits because the script should be backported into stable branches too.

Since there are no concerns I'll merge after the CI passes.

PR #2488, opened 2024-05-30
btf: take first entry on multiple btf validation

tixxdz:


pkg/btf/validation.go
```diff
@@ -53,6 +55,15 @@ func ValidateKprobeSpec(bspec *btf.Spec, call string, kspec *v1alpha1.KProbeSpec
 		if err == nil {
 			err = bspec.TypeByName(call, &fn)
 		}
+		if errors.Is(err, btf.ErrMultipleMatches) {
+			var types []btf.Type
```
olsajiri, line 59: could you please add some warning that it's happening?

tixxdz: Yes makes sense, will add it and will also make it work on non syscall, since I have some kprobes hook tests here, will update soon, thank you!

tixxdz: done ;-)

olsajiri (approved):

tixxdz:
Thanks Jiri, I added an if guard in case and tried to silence some go lint warning maybe. If it is green I merge.

PR #2494, opened 2024-05-31
renovate: switch to get Go version from toolchain directive

mtardy:
Since we now use the toolchain as the version we should use because of #2425, renovate need to pick up the toolchain version for its Go install fix.

Trying to fix the issue in https://github.com/cilium/tetragon/pull/2490.


zaru0099 (commented):
コメ

michi-covalent (approved):
so many slashes it must be correct

PR #2499, opened 2024-05-31
[v1.1] Introduce upgrade notes

lambdanis:
Backporting two commits from #2487 so that we can have upgrade notes for v1.1.x releases.


kevsecurity (approved):
LGTM

PR #2505, opened 2024-06-04
fix: path to og:image

yasell:
This PR fixes the path to `og:image` – open graph meta tags [preview](https://www.opengraph.xyz/url/https%3A%2F%2Fdeploy-preview-2505--tetragon.netlify.app%2F)


mtardy (approved):
thanks!

Issue #2429, opened 2024-05-14
sensor manager: allow concurrent listing with loading.


kkourt:
https://github.com/cilium/tetragon/pull/2285/commits/5ce0f715dd1a1b70a488277a9f333a4556595efc introduced a 0.9sec timeout to list the tracing policies from the sensor manager. The intention was to not block the metrics update for a long time. 

However, users seem to be hitting the following warning in production, probably due to the sensor manager loading sensors that take a long time.

> fix "error listing tracing policies to collect policies state" error="context deadline exceeded" warning messages

A solution would be to allow the listing of tracing policies to happen concurrently with loading. This can either be done via introducing a rw-lock or a new goroutine just for the loading.

Having statistics about how long a policy did take to load would also be useful.

PR #2506, opened 2024-06-04
allow for concurrent listing of tracing policies

kkourt:
Fixes: https://github.com/cilium/tetragon/issues/2429

This PR modifies the sensor manager to allow listing of tracing policies when sensors are loaded/unloaded.

Some BPF programs (and their corresponding sensors) take a long time to load. In the existing implementation, listing policies is serialized with loading/unloading so listing policies has to wait for any loading/unloading operations to complete.

Because tracing policies are listed as part of the metrics, https://github.com/cilium/tetragon/commit/5ce0f715dd1a1b70a488277a9f333a4556595efc introduced a timeout for listing tracing policies, which leads to this timeout being hit: https://github.com/cilium/tetragon/issues/2429.

This PR addresses this issue by introducing new states for policies (loading/unloading), and modifying the sensor manager to allow for concurrent listing of policies with loading/unloading. To do so, we add a RW lock and execute listing outside of the manager goroutine.

The PR also introduces a sensor interface that allows for better testing. The PR uses this interface to test the newly added functionality.

Please see patches for more details.

```release-note
sensors: allow reporting policy status when loading/unloading sensors
```


pkg/sensors/handler.go
```diff
@@ -129,29 +132,32 @@ func (h *handler) addTracingPolicy(op *tracingPolicyAdd) error {
 	if err != nil {
 		col.err = err
 		col.state = LoadErrorState
-		h.collections[op.ck] = &col
+		collections[op.ck] = &col
```
mtardy: you could simplify here and there now that you we use a pointer, you could insert the collection at its creation and then mutate it as we do in other methods (where you changed it in previous commits), so that we don't risk not to assign the pointer at a return point.

pkg/sensors/manager.go
```diff
@@ -335,10 +325,16 @@ func (h *Manager) LogSensorsAndProbes(ctx context.Context) {
...
+// policyLister allows read-only access to the collections map
+type policyLister interface {
+	listPolicies() []*tetragon.TracingPolicyStatus
+}
+
 // Manager handles dynamic sensor management, such as adding / removing sensors
 // at runtime.
 type Manager struct {
 	sensorCtl sensorCtlHandle
+	policyLister
```
mtardy, line 337: nit: as it might look a bit weird, maybe we can add a comment here to explain (but we have git history anyway).

pkg/sensors/handler.go
mtardy: ah nevermind, you do it https://github.com/cilium/tetragon/pull/2506/commits/6355329ee257ca5fbf487b926db4b48fb25de4ff.

pkg/sensors/handler.go
```diff
@@ -180,8 +180,8 @@ func (h *handler) disableTracingPolicy(op *tracingPolicyDisable) error {
 		return fmt.Errorf("tracing policy %s does not exist", op.ck)
 	}
 
-	if col.state == DisabledState {
-		return fmt.Errorf("tracing policy %s is already disabled", op.ck)
+	if col.state != EnabledState {
```
mtardy, line 187: well spotted

mtardy (approved):
Thanks, looks good to me, really like the testing with the fake delayed sensor.

As we discussed offline previously, this makes the code significantly more complex indeed, especially with the manual locking parts with the loading and unloading parts. Touching this code would require people to read the git history to understand, maybe we can add more immediate context by commenting a bit more. But the presence of explicit locks is already explicit comments in some ways.

pkg/sensors/manager.go
kkourt: pushed a new version with a comment. thanks!

PR #2509, opened 2024-06-05
renovate: update Go version properly for v1.1

mtardy:
Stable branch v1.1 was put in the rule to update <1.21 while it was already using 1.22. This fix should allow proper patch updates on v1.1 for the go version.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -295,7 +295,7 @@
       ]
     },
     {
-      // update go version until next minor for stable branches
+      // update go patch for 1.21 for stable branches
       "enabled": true,
       "matchPackageNames": [
         "go",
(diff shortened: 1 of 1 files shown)


kkourt (approved):

.github/renovate.json5
```diff
@@ -304,6 +304,17 @@
       "allowedVersions": "/^1\\.21\\.[0-9]+-?(alpine)?$/",
       "matchBaseBranches": [
         "v1.0",
+      ]
+    },
+    {
+      // update go version <1.22 for stable branches
```
kkourt, line 310: I'm reading this as <1.22 so <=1.21.x, but I guess the intention is to update to latest 1.22.x, correct?

PR #2516, opened 2024-06-05
helm: Template CRDs directly to get standard k8s labels

lambdanis:
So far CRDs were included in the Helm chart like this:
1. Copy into crds-yaml directory
2. Concatenate in a template using .Files.Get

With such approach, Helm doesn't add standard k8s "app.kubernetes.io/*" labels (e.g. version, managed-by) to CRDs like it does for other resources. Having these labels is useful from the operational perspective, so this commit changes how CRDs are included in the Helm chart. It now works like this:
1. Copy inside the templates directory
2. Wrap each CRD in an if block to check if it should be included

### DIFF
--- a/docs/content/en/docs/contribution-guide/making-changes.md
+++ b/docs/content/en/docs/contribution-guide/making-changes.md
@@ -75,7 +75,7 @@ these files (`pkg/k8s/`), you need to run code generation:
 
 ```shell
 make crds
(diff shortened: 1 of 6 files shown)


mtardy (approved):
looks like helm 🧙!

lambdanis:
I'll close this PR and open another one. The approach here won't work for my use cases I'm afraid.

PR #2520, opened 2024-06-06
Use the upstream Hubble packages

michi-covalent:
Hubble APIs have been very stable, so let's use the upstream version instead of maintaining our own copy.


mtardy (approved):
image

looks cool

PR #2522, opened 2024-06-06
renovate: group all go updates together and fix a rule

mtardy:
Let's group the Go updates together regardless of the branch so that's it's applied to everything. And fix a mistake commited in f03102ca24c1172da539b31548d1ccba556a0d9e putting stables branches instead of main branch in a package rule.


michi-covalent (approved):
🤖

PR #2538, opened 2024-06-11
Makefile: exclude api tags from version

kkourt:
cf0a49dcbb6ddc6ab0617aef9e16a492d7804c5c introduced a tag for the api module. This, however, breaks the release process because the tarball image generation uses the api version and fails. Exclude the api tags from the version.


michi-covalent (approved):
fascinating

PR #2540, opened 2024-06-11
tetragon: Add debug interface to track cgroups to workload/ns mappings

jrfastab:
Debugging BPF and some kernel functions I want to understand cgroup to namespace mappings at event side. This patch maintains a stable mapping between cgroups and human readable namespaces. The end goal is to filter out noisy namespaces from execs which will be follow up series. This is minimally useful as is.

To support this just extend the use of namespace filters from kprobe and tracepoints into a more general space where we can hook selectors.

### DIFF
--- a/bpf/process/policy_filter.h
+++ b/bpf/process/policy_filter.h
@@ -6,7 +6,15 @@
 
 #include "bpf_tracing.h"
 
-#define POLICY_FILTER_MAX_POLICIES 128
+#define POLICY_FILTER_MAX_POLICIES   128
+#define POLICY_FILTER_MAX_NAMESPACES 1024
+
+struct {
+	__uint(type, BPF_MAP_TYPE_LRU_HASH);
+	__uint(max_entries, POLICY_FILTER_MAX_NAMESPACES);
+	__uint(key_size, sizeof(u64));
+	__uint(value_size, sizeof(u64));
+} tg_cgroup_namespace_map SEC(".maps");
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH_OF_MAPS);
--- a/cmd/tetra/policyfilter/policyfilter.go
+++ b/cmd/tetra/policyfilter/policyfilter.go
@@ -28,8 +28,25 @@ func New() *cobra.Command {
 		dumpCmd(),
 		addCommand(),
 		cgroupGetIDCommand(),
+		dumpDebugCmd(),
 	)
 
+	return ret
+}
+
+func dumpDebugCmd() *cobra.Command {
+	mapFname := filepath.Join(defaults.DefaultMapRoot, defaults.DefaultMapPrefix, policyfilter.CgrpNsMapName)
+	ret := &cobra.Command{
+		Use:   "dumpcgrp",
+		Short: "dump cgroup ID to namespace state",
+		Args:  cobra.ExactArgs(0),
+		Run: func(_ *cobra.Command, _ []string) {
+			dump.NamespaceState(mapFname)
+		},
+	}
+
+	flags := ret.Flags()
+	flags.StringVar(&mapFname, "map-fname", mapFname, "policyfilter map filename")
 	return ret
 }
 
--- a/pkg/policyfilter/namespace.go
+++ b/pkg/policyfilter/namespace.go
@@ -1,0 +1,83 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package policyfilter
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/tetragon/pkg/bpf"
+	"github.com/cilium/tetragon/pkg/option"
+	"github.com/cilium/tetragon/pkg/sensors/exec/config"
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+const (
+	CgrpNsMapName      = "tg_cgroup_namespace_map"
+	namespaceCacheSize = 1024
+)
+
+type NSID struct {
+	Namespace string
+	Workload  string
+	Kind      string
+}
+
+// NamespaceMap is a simple wrapper for ebpf.Map so that we can write methods for it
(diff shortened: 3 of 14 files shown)


pkg/policyfilter/namespace.go
```diff
@@ -0,0 +1,137 @@
...
+			continue
+		}
+		logger.GetLogger().WithField("cgrp", c).WithField("pod", pod).WithField("id", m.id).Warn("update cgroupid map")
+
+		// If this is a new namespace we create a new map entry and bind it to a stable id.
+		if err := m.cgroupIdMap.Update(&c.cgID, m.id, ebpf.UpdateAny); err != nil {
+			logger.GetLogger().WithError(err).WithField("cgid", c.cgID).WithField("id", m.id).WithField("ns", c.name).Warn("Unable to insert cgroup id map")
+			continue
+		}
+		logger.GetLogger().Warn("nsIdMap.Add(%d %s\n", m.id, pod.namespace)
```
willfindlay, line 126: Should this maybe be an `Info` or `Debug` instead of `Warn`?

willfindlay (approved):
LGTM, just one nit.

willfindlay:
Also needs to be run though `goimports`

bpf/process/policy_filter.h
```diff
@@ -6,7 +6,15 @@
...
-#define POLICY_FILTER_MAX_POLICIES 128
+#define POLICY_FILTER_MAX_POLICIES   128
+#define POLICY_FILTER_MAX_NAMESPACES 1024
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, POLICY_FILTER_MAX_NAMESPACES);
+	__uint(key_size, sizeof(u64));
+	__uint(value_size, sizeof(u64));
+} cgroup_namespace_map SEC(".maps");
```
tixxdz, line 17: prefix with map name with tg_ while you are it?

pkg/policyfilter/namespace.go
```diff
@@ -0,0 +1,125 @@
...
+
+const (
+	CgrpNsMapName      = "cgroup_namespace_map"
+	namespaceCacheSize = 1024
+)
+
+// NamespaceMap is a simple wrapper for ebpf.Map so that we can write methods for it
+type NamespaceMap struct {
+	cgroupIdMap *ebpf.Map
+	nsIdMap     *lru.Cache[uint64, string]
```
tixxdz: the cgroup IDs could clash across nodes

cmd/tetra/policyfilter/policyfilter.go
```diff
@@ -28,11 +28,28 @@ func New() *cobra.Command {
...
 	)
 
 	return ret
 }
 
+func dumpDebugCmd() *cobra.Command {
+	mapFname := filepath.Join(defaults.DefaultMapRoot, defaults.DefaultMapPrefix, policyfilter.CgrpNsMapName)
+	ret := &cobra.Command{
+		Use:   "dumpcgrp",
+		Short: "dump cgroup ID to namespace state",
```
tixxdz, line 41: Cgroup IDs can clash, maybe add node here to give more context as we may not need to store it in map I think

tixxdz (commented):
So really quick review, but if you intent to raw print cgroup IDs or have them in logs somewhere better add more context if that makes sense.

pkg/policyfilter/namespace.go
jrfastab: I'm going to remove it. I'll add a follow up tetra command to produce some debug tables that are more useful.

pkg/policyfilter/namespace.go
jrfastab: sure that is fine.

cmd/tetra/policyfilter/policyfilter.go
jrfastab: Because this is a tetra command I think its OK to omit the node name. When I add the gRPC command that will have the node.

kkourt (changes requested):
Thanks, please find some comments below.

I think the title might be outdated because the final patch changes things to track workloads rather than namespaces.

pkg/policyfilter/namespace.go
```diff
@@ -0,0 +1,124 @@
...
+			logger.GetLogger().WithError(err).WithField("cgid", c.cgID).WithField("id", m.id).WithField("ns", c.name).Warn("Unable to insert cgroup id map")
+			continue
+		}
+		if ok := m.nsIdMap.Add(m.id, pod.namespace); ok {
+			logger.GetLogger().WithField("cgid", c.cgID).WithField("id", m.id).WithField("ns", c.name).Debug("Id to namespace map caused eviction")
+		}
+		if ok := m.nsNameMap.Add(pod.namespace, m.id); ok {
+			logger.GetLogger().WithField("cgid", c.cgID).WithField("id", m.id).WithField("ns", c.name).Debug("Namespace to Id map caused eviction")
+		}
+		m.id++
```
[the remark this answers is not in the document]
jrfastab: Collapsed a set of patches so now we just have a single patch building the end state we want. Currently, this still has the max limit of 1k workloads. I think what we really want is to have this configurable eventually. Its an LRU now though so oldest mappings will fall off the map and having 1k+ active workloads seems like an OK upper bound. I propose follow up PR to make this flexible.

pkg/policyfilter/namespace.go
```diff
@@ -0,0 +1,124 @@
...
+	if global == nil {
+		return "", false
+	}
+
+	return global.nsIdMap.Get(id)
+}
+
+// addCgroupIDs add cgroups ids to the policy map
+// todo: use batch operations when supported
+func addCgroupIDs(cinfo []containerInfo, pod *podInfo) error {
```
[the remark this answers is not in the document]
jrfastab: I made it a method of state. I think this is OK.

pkg/policyfilter/namespace.go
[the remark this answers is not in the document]
kkourt: Agreed, 1k active workloads seems like a good upper bound.
Shouldn't the BPF map be an LRU map as well though?

PR #2549, opened 2024-06-13
[backport v1.0] pkg/sensors: reduce stack trace map memory footprint

mtardy:
[ upstream commit 22510d98a16594b19b6ef8c68d3db5da5cb0a8cc ] 
This is a slightly adapted version because #2145 was merged after v1.0

We stopped on a stack trace map that has a max_entries of 32768, which is 64 bits pointers * PERF_MAX_STACK_DEPTH (which is fixed at 127 for now), so 127*64/8=1016 bytes per entry + it's key_size of 32 bits (4 bytes) so 1020 bytes per entry. So 1020 * 32768 = 33,423,360 bytes. From bpftool, this map has a total bytes_memlock of 34,079,040 bytes. So for each stack trace map we load, we had 34MB of kernel memory, and it happened to be loaded many times when we were loading any tracing policy.

Since the map is used by the generic program, the loader will allocate the memory needed for the map even if we don't create a reference from the agent side and create an anonymous map. So we end up allocating a small map of max_entries 1 by default and resize it when the tracing policy actually specifies a matchAction containing a kernelStackTrace or userStackTrace to true. This should drastically reduce the memory footprint of this feature when it's unused.

```release-note
Reduce the kernel memory footprint (accounted by the cgroup memory controller) of the stack trace feature when unused.
```

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -65,6 +65,8 @@ const (
 	CharBufErrorPageFault   = -2
 	CharBufErrorTooLarge    = -3
 	CharBufSavedForRetprobe = -4
+
+	stackTraceMapMaxEntries = 32768 // this value could be fine tuned
 )
 
 func kprobeCharBufErrorToString(e int32) string {
@@ -132,6 +134,9 @@ type genericKprobe struct {
 	// reference to a stack trace map, must be closed when unloading the kprobe,
 	// this is done in the sensor PostUnloadHook
 	stackTraceMapRef *ebpf.Map
+
+	// is there stacktrace defined in the kprobe
+	hasStackTrace bool
 
 	customHandler eventhandler.Handler
 }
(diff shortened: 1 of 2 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -211,6 +216,15 @@ func createMultiKprobeSensor(sensorPath string, multiIDs, multiRetIDs []idtable.
 	var progs []*program.Program
 	var maps []*program.Map
 
+	oneKprobeHasStackTrace := false
+	for _, id := range multiIDs {
+		gk, err := genericKprobeTableGet(id)
+		if err != nil {
+			continue
```
mtardy, line 223: I had to write this because we haven't https://github.com/cilium/tetragon/pull/2145, not sure it's okay to ignore that error like that, we could end up with a kprobe with stacktrace that we fail to retrieve from the kprobe table and then we don't resize the map properly. Might be difficult to debug that the issue is that the map is of size 1. wdyt @olsajiri ?

olsajiri: I think we should fail like we do in upstream

```
        for _, id := range multiIDs {
                gk, err := genericKprobeTableGet(id)
                if err != nil { 
                        return nil, nil, err
                }
```

mtardy: yes I would love to as well but the signature of the function in v1.0:
```go
func createMultiKprobeSensor(sensorPath string, multiIDs, multiRetIDs []idtable.EntryID) ([]*program.Program, []*program.Map)
```
Should I backport those changes as well then? There's no error returned in this version

pkg/sensors/tracing/generickprobe.go
```diff
@@ -770,6 +787,13 @@ func addKprobe(funcName string, f *v1alpha1.KProbeSpec, in *addKprobeIn, selMaps
 		config.Flags |= flagsEarlyFilter
 	}
 
+	hasStackTrace := false
+	for _, selector := range f.Selectors {
+		for _, matchAction := range selector.MatchActions {
+			hasStackTrace = matchAction.StackTrace
```
mtardy, line 793: fix this

kkourt (approved):
LGTM, thanks!

olsajiri (approved):

PR #2551, opened 2024-06-13
pkg/sensors: reduce ratelimit map memory footprint

mtardy:
For every ratelimit map loaded, we add ~10MB of kernel memory, and each kprobe added was adding a ratelimit map. We now only load that map if the user used the rateLimit field in a matchActions to reduce the memory footprint of this feature when unused.

```release-note
Reduce the kernel memory footprint (accounted by the cgroup memory controller) of the ratelimit feature when unused (around ~10MB per kprobe).
```

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1891,6 +1891,7 @@ FUNC_INLINE void do_action_signal(int signal)
  */
 #define KEY_BYTES_PER_ARG 40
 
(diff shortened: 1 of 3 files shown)


mtardy:
It's ready for review but it would be nice to add a test for `ratelimit` so that we are sure it wasn't broken here or will be.

kevsecurity (approved):
LGTM

PR #2563, opened 2024-06-14
pkg/sensors: reduce memory footprint of unused fdinstall maps and various fixes

mtardy:
~~The code is not great but it fixes the issue, maybe we can do it more elegantly idk~~

```release-note
Reduce the kernel memory footprint (accounted by the cgroup memory controller) of the ratelimit feature when unused (around ~11MB per kprobe).
```

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -62,6 +62,7 @@ const (
 	// much kernel memory when enabled.
 	stackTraceMapMaxEntries = 32768
 	ratelimitMapMaxEntries  = 32768
+	fdInstallMapMaxEntries  = 32000
 )
 
 func kprobeCharBufErrorToString(e int32) string {
@@ -271,7 +272,7 @@ func filterMaps(load *program.Program, pinPath string, kprobeEntry *genericKprob
 	return maps
 }
 
-func createMultiKprobeSensor(sensorPath, policyName string, multiIDs []idtable.EntryID) ([]*program.Program, []*program.Map, error) {
+func createMultiKprobeSensor(sensorPath, policyName string, multiIDs []idtable.EntryID, enableFDInstall bool) ([]*program.Program, []*program.Map, error) {
 	var multiRetIDs []idtable.EntryID
 	var progs []*program.Program
 	var maps []*program.Map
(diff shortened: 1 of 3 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -290,6 +297,10 @@ func createMultiKprobeSensor(sensorPath, policyName string, multiIDs []idtable.E
 		}
 		oneKprobeHasStackTrace = oneKprobeHasStackTrace || gk.hasStackTrace
 		oneKprobeHasRatelimit = oneKprobeHasRatelimit || gk.hasRatelimit
+		// it should be overkill (we could just get the value of the first)
+		// because the caller should have already uniformized hasFDInstall
+		// between all the kprobes
+		resizeFDInstall = resizeFDInstall || gk.hasFDInstall
```
olsajiri, line 303: could we just pass the detected oneKprobeHasFDInstall in createMultiKprobeSensor directly?
maybe we don't need to store it for kprobeEntry, but just pass it to createKprobeSensorFromEntry and createMultiKprobeSensor as single bool ?

mtardy: yep yep we can, it's another way of passing that stuff. I thought it would clobber the params of the functions but if that's simpler I'll try.

mtardy: this is much more reasonable and readable indeed https://github.com/cilium/tetragon/compare/1b4c684aa4b475b380a433dfa7cde7217b179b0c..f0189d37149a202690251a3d84e1b95fc4f87657

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -351,6 +354,7 @@ func createGenericTracepoint(
 		customHandler: customHandler,
 		message:       msgField,
 		tags:          tagsField,
+		hasFDInstall:  selectorsHaveFDInstall(conf.Selectors),
```
kkourt, line 357: Do we really need the extra field? Can we just run the check when we check the field?

kkourt (approved):
LGTM, just a minor not for your consideration.

pkg/sensors/tracing/generictracepoint.go
mtardy: ah yes we can do that, I mimicked the kprobe stuff but this is not needed here indeed.

mtardy: because we no longer have access to the spec in the kprobe side, so we need that, but here it's simpler.

kkourt: I see, thanks!

mtardy: https://github.com/cilium/tetragon/compare/a44a1fa6995cdad90fd43ece109646df66c56cbd..2631f9b91d87247770531dfc1582665ebf19935c

Issue #2409, opened 2024-05-03
Tetragon based file integrity monitoring (FIM)


anfedotoff:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature you would like

We could use Tetragon for file integrity monitoring: collect hashes of executed binaries and opened files and put this information in events. Hashes are calculated using IMA-measurement  Linux integrity subsystem.

### Describe your proposed solution

We already talked about  [FIM](https://docs.google.com/document/d/1BFMJLdtisiCSLfMct0GHof_ioL-5QVNLEaeMSlk_7Eo/edit#heading=h.fkeiyof823bv). I found some technical issues during my research, so I decided to provide a [CFP](https://docs.google.com/document/d/1id3tAsBkOMe6P5eop31cBu1pLMTleITm-44S94Z87vE/edit#heading=h.5mwz58y9syps) before PR.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

anfedotoff:
Hi :wave: , @kkourt!
If you have time, please, have a look. I'll be happy to have some discussion on implementation details.

xmulligan:
Once it is ready, please also add the CfP to the repo https://github.com/cilium/design-cfps

anfedotoff:
LGTM! We still able to filter by file path, before collecting a hash in your approach, right? In other words I mean not to call  ima bpf-helpers if filtering is not passed.

As far as I concerned, IMA bpf-helpers just retrieve the hash from IMA-measurement list.  Difference between `bpf_ima_inode_hash` (5.15) and `bpf_ima_file_hash` (5.18): if there is no hash in IMA-measurement list `bpf_ima_file_hash` will calculate the hash, update IMA-measurement list and return it to the caller.

> operator: "ima_file_hash"

Here you mean to call appropriate bpf-helper according to kernel version? Or user specifies the helper it prefers? I think, the first way is better.

kkourt:
> LGTM! We still able to filter by file path, before collecting a hash in your approach, right? In other words I mean not to call ima bpf-helpers if filtering is not passed.

I think it should be possible to collect the hash _after_ the filtering, but it's more tricky. In that case, collecting the hash in the action makes more sense to me, but we will need to maintain the necessary arguments to call the helpers.

> 
[...]

PR #2128, opened 2024-02-21
Pr/olsajiri/loader

olsajiri:
wip

PR #2566, opened 2024-06-16
LSM sensor

anfedotoff:
LSM sensor support allows to use LSM BPF programs the way we use BPF programs for kprobes/tracepoints/uprobes.

TracingPolicy example:

```yaml
kind: TracingPolicy
metadata:
  name: "lsm"
spec:
  lsmhooks:
  - hook: "file_open"
    args:
      - index: 0
        type: "file"
    selectors:
    - matchBinaries:
      - operator: "In"
        values:
        - "/usr/bin/cat"
```

Event example:

```
{                                                                                                     
  "process_lsm": {                                                                                    
    "process": {                                                                                      
      "exec_id": "dXNlci1uaXg6MTk5MzE1NDk0NTk3MzM6MzIwMTA4",                                  
      "pid": 320108,                                                                                  
      "uid": 1000,                                                                                    
      "cwd": "/home/user/go/src/github.com/cilium/tetragon",                                    
      "binary": "/usr/bin/cat",                                                                       
      "arguments": "/etc/passwd",                                                                     
      "flags": "execve clone",                                                                        
      "start_time": "2024-06-15T18:03:29.742161520Z",                                                                                                                                                       
      "auid": 1000,                                                                                   
      "parent_exec_id": "dXNlci1uaXg6ODg1NjMwMDAwMDAwMDoxNDc4MTI=",                                                                                                                                 
      "refcnt": 1,                                                                                    
      "tid": 320108,                                                                                  
      "user": {                                                                                       
        "name": "user"                                                                          
      }                                                                                               
    },                                                                                                
    "parent": {                                                                                       
      "exec_id": "dXNlci1uaXg6ODg1NjMwMDAwMDAwMDoxNDc4MTI=",                                  
      "pid": 147812,                                                                                  
      "uid": 1000,                                                                                    
      "cwd": "/home/user/go/src/github.com/cilium/tetragon",                                                                                                                                          
      "binary": "/usr/bin/zsh",                                                                                                                                                                             
      "flags": "procFS auid",                                                                         
      "start_time": "2024-06-15T14:37:33.597296165Z",
      "auid": 1000,
      "parent_exec_id": "dXNlci1uaXg6MTM1ODA0MDAwMDAwMDozMTQ2",
      "tid": 147812
    },
    "function_name": "file_open",
    "policy_name": "lsm",
    "args": [
      {
        "file_arg": {
          "path":"/etc/passwd",
          "permission":"-rw-r--r--"
        }
      }
    ],
  },
  "node_name": "user-nix",
  "time": "2024-06-15T18:03:29.743030933Z"
}
```
This is  also necessary for #2409.


anfedotoff:
There are things to do:

- Fix tests
- Add lsm test
- Decide minimal kernel version for LSM sensor support (LSM BPF is needed 5.7 kernel).
- Fix some bugs in code and make it more neat.

I managed to load LSM BPF programs, tail calls also work for LSM programs! The most terrifying problems are solved, I think. I managed to catch some Lsm events).

anfedotoff:
For now I managed to fix problem with args resolving. LSM BPF programs get args the way raw tracepoint programs do. So, LSM events now are looking good. It would be nice to start code review.

Some open questions:
* ./verify/verify.sh . I made a hack for loading generic LSM programs. But I need to install llvm-objcopy in CI
* Tests. I think we need some tests. Maybe more than one. I'll work on it. Maybe we discuss what kind of tests it's good to have?
* Enforce mode. If LSM BPF returns the value that is not qual to zero, than operation is not permitted. Maybe we should add an action for that? For example, if policy is violated just return -EPERM, instead of sending SIGKILL.

pkg/sensors/tracing/genericlsm.go
```diff
@@ -0,0 +1,465 @@
...
+) (*sensors.Sensor, error) {
+	var progs []*program.Program
+	var maps []*program.Map
+	var ids []idtable.EntryID
+	var selMaps *selectors.KernelSelectorMaps
+	var err error
+
+	if !kernels.MinKernelVersion("5.7") {
+		return nil, fmt.Errorf("to load LSM sensor kernel version must be greater or equal 5.7")
+	}
```
kkourt, line 280: Generally, it is a good idea to avoid doing kernel checks because in some cases features have been backported to older kernels. We have been doing it in the past, but I think we should avoid it for new code and try to migrate the code to just use feature detection.

How about we just check just for HasLSMPrograms, and we can change the error to:
> fmt.Errorf("Does you kernel support the bpf LSM? You can enable LSM BPF by modifying the GRUB configuration /etc/default/grub with GRUB_CMDLINE_LINUX=\"lsm=bpf\"")

pkg/sensors/tracing/genericlsm.go
```diff
@@ -0,0 +1,465 @@
...
+		path.Join(option.Config.HubbleLib, loadProgName),
+		lsmEntry.hook,
+		"lsm/generic_lsm",
+		pinProg,
+		"generic_lsm").
+		SetLoaderData(lsmEntry.tableId)
+	progs = append(progs, load)
+
+	fdinstall := program.MapBuilderPin("fdinstall_map", sensors.PathJoin(sensorPath, "fdinstall_map"), load)
+	maps = append(maps, fdinstall)
```
kkourt, line 355: It would have been nice to use the new interface, but it's not clear when it's going to be ready: https://github.com/cilium/tetragon/pull/2128.

CC: @olsajiri

pkg/sensors/tracing/genericlsm.go
```diff
@@ -0,0 +1,465 @@
...
+
+	if !kernels.MinKernelVersion("5.7") {
+		return nil, fmt.Errorf("to load LSM sensor kernel version must be greater or equal 5.7")
+	}
+
+	if !bpf.HasLSMPrograms() {
+		return nil, fmt.Errorf("enable LSM BPF by modifying the GRUB configuration /etc/default/grub with GRUB_CMDLINE_LINUX=\"lsm=bpf\"")
+	}
+
+	lsmHooks := spec.LsmHooks
```
kkourt, line 286: At this point, I think we should check the actions and reject policies with actions that are not properly supported in the LSM sensor.

One example is Post actions with stack trace, since in ` handleGenericLsm` we do not seem to do what is done in kprobes: https://github.com/cilium/tetragon/blob/e44f7dfc0e8ee90ddcef1d915111d420917b9ae0/pkg/sensors/tracing/generickprobe.go#L1217

Another example is the GetURL, DnsLookup actions: https://github.com/cilium/tetragon/blob/e44f7dfc0e8ee90ddcef1d915111d420917b9ae0/pkg/sensors/tracing/generickprobe.go#L1177

Unfortunately, there is a lot of code duplication among the kprobe/tracepoint/uprobe and now the LSM sensor. It would be nice if we could reduce it, but that's a different issue alltogether.

pkg/sensors/tracing/lsm_test.go
```diff
@@ -0,0 +1,117 @@
...
+    selectors:
+    - matchBinaries:
+      - operator: "In"
+        values:
+        - "/usr/bin/cat"
+      matchArgs:
+      - index: 0
+        operator: "Equal"
+        values:
+        - "/etc/passwd"
```
kkourt, line 87: nit: can we use either a different binary or a different file for that test? Something that can be randomized so that the policy does not affect anything else running on the machine (e.g., we can create a file under `/tmp` and check that.

kkourt (commented):
Thanks!

Overall, this looks good to me! I've left some minor comments.

>    ./verify/verify.sh . I made a hack for loading generic LSM programs. But I need to install llvm-objcopy in CI

Looks good to me. Yap, let's install it. I'm guessing it should be part of our clang container so we can just use that.

>    Tests. I think we need some tests. Maybe more than one. I'll work on it. Maybe we discuss what kind of tests it's good to have?

That's a good question. I would say start simple, and have one or two tests for the basic functionality (observability and enforcement)

>    Enforce mode. If LSM BPF returns the value that is not qual to zero, than operation > is not permitted. Maybe we should add an action for that? For example, if policy is > violated just return -EPERM, instead of sending SIGKILL.

Yap, I think that makes sense. We also have an override action that does something similar, but having a separate action for lsm (as you do in your PR) makes sense to me.

kkourt:
@olsajiri I added you to the reviewers. It would be great if you could have a look when you get a chance. Thanks!

bpf/Makefile
```diff
@@ -146,6 +149,12 @@ objs/%_v511.ll:
 $(DEPSDIR)%_v511.d:
 	$(CLANG) $(CLANG_FLAGS) -D__LARGE_BPF_PROG -D__LARGE_MAP_KEYS -MM -MP -MT $(patsubst $(DEPSDIR)%.d, $(OBJSDIR)%.ll, $@)   $< > $@
 
+objs/bpf_generic_lsm_v57.ll: process/bpf_generic_lsm.c
+	$(CLANG) $(CLANG_FLAGS) -D__LARGE_BPF_PROG -c $< -o $@
```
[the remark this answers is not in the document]
anfedotoff: Yes, it makes sense.

pkg/sensors/tracing/genericlsm.go
anfedotoff: I agree, also I found out that rhel8 in vmtests has 4.18 kernel, but LSM BPF is enabled on this VM!

pkg/sensors/tracing/genericlsm.go
anfedotoff: Ok, I'll carefully look what actions aren't available from LSM hooks and check for them.

Yes, I noticed that there are some code duplicates among kprobes/uprobes/tracepoints/lsm. Also data structures from kprobe sensor are used in another types of programs.  I agree, it is a good idea to do some refactoring in that case in future.

pkg/sensors/tracing/lsm_test.go
anfedotoff: I fixed tests and I also added test for LsmBlock action.

anfedotoff:
Some problems with tests:
- Verification of `bpf_generic_lsm_v61.o` fails. I need to investigate why (help is appreciated).
- Tetragon Go Test / build (ubuntu-20.04): LsmFileOpen test is failing, but locally it works fine.
- rhle8 vmtests for LSM are failing, but locally they are passing. Command to run:

```
 sudo ./tests/vmtests/tetragon-vmtests-run \
        --kernel tests/vmtests/test-data/kernels/rhel8/boot/vmlinuz-4.18.0 --base tests/vmtests/test-data/images/base.qcow2 --testsfile tests/vmtests/test-lsm --fail-fast
```

test-lsm:

```
pkg.sensors.tracing:TestLSMObjectLoad
pkg.sensors.tracing:TestLSMOpenFile
pkg.sensors.tracing:TestLSMBlockAction
```

tixxdz (changes requested):

bpf/process/bpf_generic_lsm.c
```diff
@@ -0,0 +1,195 @@
...
+	bool *block;
+
+	generic_output(ctx, (struct bpf_map_def *)&process_call_heap, MSG_OP_GENERIC_LSM);
+
+	block = map_lookup_elem(&lsm_block_map, &id);
+	if (!block)
+		return 0;
+
+	map_delete_elem(&lsm_block_map, &id);
+	return -1;
```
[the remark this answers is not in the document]
anfedotoff: Do you mean that there are different types of non-zero return values for lsm bpf hooks? And these values affect differently on further execution?
Than, it makes sense to have override action with user-provided return value. We can keep LSMblock action for -EPERM, the way we have SigKill and Signal action.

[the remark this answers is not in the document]
kkourt: Currently, we have an `Override` action that uses `ArgError` to override a return value.
Could we rese the existing `Override` action in LSM hook?

This would mean that we interpret the `Override` action differently if we are in an kprobe or in lsm: in the former we do bpf_override_return() and in the latter we just return. See: https://github.com/cilium/tetragon/blob/a5c666eee73a0a4412203395f9b0f1c348fde8ca/bpf/process/bpf_generic_kprobe.c#L224.

Maybe we can even re-use the `override_tasks` map that is used in kprobes instead of adding a new one: https://github.com/anfedotoff/tetragon/blob/lsm-sensor/bpf/process/types/basic.h#L126-L133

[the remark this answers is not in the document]
anfedotoff: Yes, sure, we could. Do we want to keep an LsmBlock action as an alias for Override action with ArgError = -1?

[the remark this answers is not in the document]
kkourt: > Do we want to keep an LsmBlock action as an alias for Override action with ArgError = -1?

No strong opinion on that.

kkourt:
> Verification of bpf_generic_lsm_v61.o fails. I need to investigate why (help is appreciated).

What is the verification error?

> Tetragon Go Test / build (ubuntu-20.04): LsmFileOpen test is failing, but locally it works fine.
> --- FAIL: TestLSMOpenFile (6.85s)

Had a quick look, but couldn't figure out what the issue is. I can have a closer look next week. Here are some first notes:

Looking at the GH CI logs:
>     logcapture.go:25: time="2024-07-05T10:53:43Z" level=info msg="Loaded generic LSM program: /home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/bpf/objs/bpf_generic_lsm_v511.o -> file_open"
>     logcapture.go:25: time="2024-07-05T10:53:43Z" level=info msg="BPF detected features: override_return: true, buildid: true, kprobe_multi: false, uprobe_multi false, fmodret: true, fmodret_syscall: true, signal: true, large: true, lsm: true"

So everything seems to work OK so far.

Logs also say:
>     jsonchecker.go:183: jsonTestCheck: opening: /tmp/tetragon.gotest.TestLSMOpenFile.3243682509.json

Downloading the artifacts (https://github.com/cilium/tetragon/actions/runs/9807021872?pr=2566), we can find the file there. Looking at the file, there are no lsm events there:

```
$ jq '. | del(.time) | del(.node_name) | to_entries[] | .key' < tetragon.gotest.TestLSMOpenFile.3243682509.json
"process_exec"
"process_exec"
"process_exit"
```

bpf/process/bpf_generic_lsm.c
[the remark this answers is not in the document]
anfedotoff: OK, let's just have an override action! We can add a new action in the future  if it is needed.

anfedotoff:
> What is the verification error?
You can find it in ARM tests. I fixed it locally, but arm ci verification step still failing...

> Had a quick look, but couldn't figure out what the issue is.
Yes, thank you for helping me). It is a little bit weird because locally, it is in logs. Vmtests are working fine also.

anfedotoff:
@kkourt I found out why verification is failed.

https://github.com/cilium/tetragon/blob/c7799feacf315d18d91064d3256c0c28efccb8c8/bpf/process/bpf_process_event.h#L251-L259.

So, the code with loop(...) fails the verification. If I force use the code part with `for`, everything works fine...  May be the problem somehow related with  error `missing btf func_info` and

https://github.com/cilium/tetragon/blob/c7799feacf315d18d91064d3256c0c28efccb8c8/bpf/process/bpf_process_event.h#L227-L232

olsajiri:
> @olsajiri I added you to the reviewers. It would be great if you could have a look when you get a chance. Thanks!

ugh, sry missed this one, will check

olsajiri (changes requested):
hi,
IMO this needs to be split into multiple patches, it's hard to review 6000+ lines diff

from quick look this can be separated to smaller logical changes, like:
  api changes
  pkg/sensors changes
  pkg/bpf changes
  .github changes
  bpf changes
  install changes
  ... 
  plz split anything that makes sense

anfedotoff:
> hi,
> IMO this needs to be split into multiple patches, it's hard to review 6000+ lines diff
> 
> from quick look this can be separated to smaller logical changes, like:
>   api changes
>   pkg/sensors changes
>   pkg/bpf changes
>   .github changes
>   bpf changes
>   install changes
>   ... 
>   plz split anything that makes sense

Yes, sure

pkg/sensors/tracing/lsm_test.go
```diff
@@ -94,24 +100,93 @@ spec:
...
+
+	err = jsonchecker.JsonTestCheck(t, ec.NewUnorderedEventChecker(lsmChecker))
+	assert.NoError(t, err)
+}
 
-	if err := catCmd.Run(); err != nil {
-		t.Fatalf("failed to run %s: %s", catCmd, err)
+func TestLSMBOverrideAction(t *testing.T) {
+	if !bpf.HasLSMPrograms() || !kernels.EnableLargeProgs() {
+		t.Skip()
```
olsajiri, line 127: should we somehow also check bpf is configured on running kernel in CONFIG_LSM ?

pkg/sensors/tracing/lsm_test.go
```diff
@@ -94,24 +100,93 @@ spec:
...
+	obs, err := observertesthelper.GetDefaultObserverWithFile(t, ctx, testConfigFile, tus.Conf().TetragonLib, observertesthelper.WithMyPid())
+	if err != nil {
+		t.Fatalf("GetDefaultObserverWithFile error: %s", err)
+	}
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	testCmd := exec.Command(testBin)
 
+	testCmd.Run()
```
olsajiri, line 189: please also check the overriden return value

olsajiri (changes requested):
lgtm, but plz split it into multiple changes

pkg/sensors/tracing/genericlsm.go
```diff
@@ -0,0 +1,465 @@
...
+		if len(s.MatchReturnArgs) > 0 {
+			return fmt.Errorf("MatchReturnArgs selector is not supported")
+		}
+		if len(s.MatchActions) > 0 {
+			for _, a := range s.MatchActions {
+				switch strings.ToLower(a.Action) {
+				case "sigkill":
+				case "signal":
+				case "nopost":
+					return nil
```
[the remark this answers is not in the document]
anfedotoff: Nice catch, thanks.

anfedotoff:
@olsajiri, I splitted code to logical commits and fix other comments, please, have a look.

anfedotoff:
What about tests:
 I tested on Ubuntu 20.04

`Linux lsm-test 5.15.0-1067-azure #76~20.04.1-Ubuntu SMP Thu Jun 13 18:00:23 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`

Tests are passing. For now I don't have a clue why CI is failing. As we found out, there is no LSM event in CI test. The kernel in CI is a bit different from which I test locally.
CI kernel:

```
Linux fv-az1496-117 5.15.0-1067-azure #76~20.04.1-Ubuntu SMP Thu Jun 13 18:00:23 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
```

I can try to change policy and look if we have LSM events without any filtering. @kkourt  do you have any ideas about this problem?

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -212,6 +212,64 @@ spec:
...
+CONFIG_BPF_LSM=y
+```
+
+If the output contains `CONFIG_BPF_LSM=y`, BPF LSM is supported.
+Then, if provided above conditions are met, use this command to check if BPF LSM is enabled:
+
+```
+$ cat /sys/kernel/security/lsm
+bpf,lockdown,integrity,apparmor
+```
```
olsajiri, line 236: should this be part of the lsm detection? I might be wrong here but IIUC the whole thing does not work of 'bpf' is not present in here, right?

olsajiri (commented):
looks good, left few comments

and thanks a lot for the split, it's so much easier to do the review

docs/content/en/docs/concepts/tracing-policy/hooks.md
anfedotoff: Yes, exactly you are right! `CONFIG_LSM=y`. Tells us that kernel is compiled with LSM bpf support.  `bpf` in  `/sys/kernel/security/lsm` tells us that LSM is enabled right here, right now.

`bpf.HasLSMPrograms()` tries to load LSM  BPF program. It seems to me, that to load `CONFIG_LSM=y`  is enough. That's  why only TestLSMObjectLoad working in gotests.

Thanks so much! I'll add this check to `bpf.HasLSMPrograms()`!

contrib/verify/verify.sh
```diff
@@ -78,13 +78,31 @@ for obj in "$TETRAGONDIR"/*.o; do
...
+			--rename-section lsm/0=lsm/file_open \
+			--rename-section lsm/1=lsm/file_open \
+			--rename-section lsm/2=lsm/file_open \
+			--rename-section lsm/3=lsm/file_open \
+			--rename-section lsm/4=lsm/file_open \
+			--rename-section lsm/5=lsm/file_open \
+			"$obj" "${obj}.fixup"
+		obj="${obj}.fixup"
+	fi
+
```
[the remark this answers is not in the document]
anfedotoff: It is not trivial to load generic LSM bpf object using bpftool. I just skip load check for LSM, the way we do for tracepoints. We have TestLSMObjectLoad I think it is enough. We already talk about this with @kkourt ,  at the last Tetragon Community Meeting.

.github/workflows/gotests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get update
-        sudo apt-get -y install libelf-dev netcat-traditional libcap-dev gcc
+        sudo apt-get -y install libelf-dev netcat-traditional libcap-dev gcc llvm
```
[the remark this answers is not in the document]
anfedotoff: This commit is no needed any more. I'll write some description to other commits.

bpf/process/bpf_generic_kprobe.c
```diff
@@ -158,7 +111,7 @@ generic_kprobe_start_process_filter(void *ctx)
 __attribute__((section((MAIN)), used)) int
 generic_kprobe_event(struct pt_regs *ctx)
 {
-	return generic_kprobe_start_process_filter(ctx);
+	return generic_start_process_filter(ctx, &maps, (struct bpf_map_def *)&config_map);
 }
```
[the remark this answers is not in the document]
anfedotoff: Yes, I see, I missed this somehow). I'll fix that.

kkourt:
> What about tests: I tested on Ubuntu 20.04
> 
> `Linux lsm-test 5.15.0-1067-azure #76~20.04.1-Ubuntu SMP Thu Jun 13 18:00:23 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`
> 
> Tests are passing. For now I don't have a clue why CI is failing. As we found out, there is no LSM event in CI test. I can try to change policy and look if we have LSM events without any filtering. @kkourt do you have any ideas about this problem?

The failure in https://github.com/cilium/tetragon/actions/runs/9894669687/job/27332734013?pr=2566 says:
>     observer_test_helper.go:447: LoadConfig error: failed prog /home/runner/actions-runner/_work/tetragon/tetragon/go/src/github.com/cilium/tetragon/bpf/objs/bpf_generic_lsm_v61.o kern_version 393562 loadInstance: attaching 'generic_lsm_event' failed: create raw tracepoint: not supported

Which might explain what's wrong? I'd suggest we probe for above functionality and skip the test if functionality is missing.

anfedotoff:
> > What about tests: I tested on Ubuntu 20.04
> > 
> > `Linux lsm-test 5.15.0-1067-azure #76~20.04.1-Ubuntu SMP Thu Jun 13 18:00:23 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux`
> > 
> > Tests are passing. For now I don't have a clue why CI is failing. As we found out, there is no LSM event in CI test. I can try to change policy and look if we have LSM events without any filtering. @kkourt do you have any ideas about this problem?
> 
> The failure in https://github.com/cilium/tetragon/actions/runs/9894669687/job/27332734013?pr=2566 says:
> >     observer_test_helper.go:447: LoadConfig error: failed prog /home/runner/actions-runner/_work/tetragon/tetragon/go/src/github.com/cilium/tetragon/bpf/objs/bpf_generic_lsm_v61.o kern_version 393562 loadInstance: attaching 'generic_lsm_event' failed: create raw tracepoint: not supported
> 
> Which might explain what's wrong? I'd suggest we probe for above functionality and skip the test if functionality is missing.

Yes,  you are right. It is another problem on ARM machines. The problem I mentioned earlier, we have solved already.
I'll have a look at how to check if raw tracepoints are available, and put this check in `bpf.HasLSMPrograms()`.

anfedotoff:
It seems to me, that test are passing and fixed all comments. @olsajiri @kkourt , please, could you have a look one more time:)?

olsajiri (approved):
lgtm, nice work! thanks

kkourt (approved):
LGTM, thanks!

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -212,6 +212,64 @@ spec:
...
+List of LSM hooks which can be instrumented can be found in `security/security.c`.
+
+To verify if BPF LSM is available use the following command:
+
+```
+$ cat /boot/config-$(uname -r) | grep BPF_LSM
+CONFIG_BPF_LSM=y
+```
+
+If the output contains `CONFIG_BPF_LSM=y`, BPF LSM is supported.
```
mtardy, line 230: This is a nit only on doc, could you [separate the cmd from its output](https://kubernetes.io/docs/contribute/style/style-guide/#separate-commands-from-output)?

````suggestion
```shell
cat /boot/config-$(uname -r) | grep BPF_LSM
```

The output should be similar to this if BPF LSM is supported:
```
CONFIG_BPF_LSM=y
```
````

File: docs/content/en/docs/concepts/tracing-policy/hooks.md (hunk 10 above)
Comment (mtardy) [inline docs/content/en/docs/concepts/tracing-policy/hooks.md:236]: same as previous on cmd/output separation (without $ or #)

**Review by mtardy — COMMENTED** [review]
This PR is lovely and nice to review! Thanks!

File: docs/content/en/docs/concepts/tracing-policy/hooks.md (hunk 14 above)
Comment (mtardy) [reply docs/content/en/docs/concepts/tracing-policy/hooks.md:230]: if you want to merge this as it's green and approved, and fix the doc later it's completely okay

Comment (anfedotoff) [reply docs/content/en/docs/concepts/tracing-policy/hooks.md]: No problem, I'll fix it right now!

File: bpf/process/bpf_generic_lsm.c [hunk 15]
```diff
@@ -0,0 +1,162 @@
...
+
+__attribute__((section("lsm/5"), used)) int
+generic_lsm_output(void *ctx)
+{
+	__u64 id = get_current_pid_tgid();
+	__s32 *error;
+
+	generic_output(ctx, (struct bpf_map_def *)&process_call_heap, MSG_OP_GENERIC_LSM);
+
+	error = map_lookup_elem(&override_tasks, &id);
```
tixxdz, line 156: Hmm, what am I missing here, in the previous tail call you could delete this entry? so making this check here will fail which will lead to return 0 ?  (I didn't check this part of code lately, so just dropping a comment...)

anfedotoff: The case you mentioned is only works when we set `NoPost` Action.

https://github.com/cilium/tetragon/blob/2a31728a71b689554fc1fca4573945908cae20e6/bpf/process/types/basic.h#L2329-L2332

olsajiri: I think the reason is that the override is done either in generic_lsm_actions or in generic_lsm_output depending on whether there's output or not.. but I think it'd good to have it in separate function like (untested):

```
diff --git a/bpf/process/bpf_generic_lsm.c b/bpf/process/bpf_generic_lsm.c
index 4a8736c180a3..1e80afcfa307 100644
--- a/bpf/process/bpf_generic_lsm.c
+++ b/bpf/process/bpf_generic_lsm.c
@@ -128,11 +128,8 @@ generic_lsm_filter_arg(void *ctx)
                               true);
 }
 
-__attribute__((section("lsm/4"), used)) int
-generic_lsm_actions(void *ctx)
+FUNC_INLINE int try_override(void *ctx)
 {
-       generic_actions(ctx, &maps);
-
        // If NoPost action is set, check for Override action here
        __u64 id = get_current_pid_tgid();
        __s32 *error;
@@ -145,18 +142,18 @@ generic_lsm_actions(void *ctx)
        return (long)*error;
 }
 
+__attribute__((section("lsm/4"), used)) int
+generic_lsm_actions(void *ctx)
+{
+       generic_actions(ctx, &maps);
+
+       // If NoPost action is set, check for Override action here
+       return try_override(ctx);
+}
+
 __attribute__((section("lsm/5"), used)) int
 generic_lsm_output(void *ctx)
 {
-       __u64 id = get_current_pid_tgid();
-       __s32 *error;
-
        generic_output(ctx, (struct bpf_map_def *)&process_call_heap, MSG_OP_GENERIC_LSM);
-
-       error = map_lookup_elem(&override_tasks, &id);
-       if (!error)
-               return 0;
-
-       map_delete_elem(&override_tasks, &id);
-       return (long)*error;
+       return try_override(ctx);
 }

```

anfedotoff: Yes, it's a good idea!

tixxdz (approved):
Didn't have time to review all, thank you for the nice contributions ;-)

kkourt:
Merging this, thanks! Great work!

PR #2575, opened 2024-06-17
process:bpf: report euid as the process.uid

tixxdz:
Instead of reporting the real uid of the task that is the owner of the task, let's report the effective uid that is used to calculate the privileges of the current task when acting upon other objects. This allows to be compatible with 'ps' too.

See commits for further details.


kkourt (approved):
LGTM!
Thanks for the detailed commit message and the doc updates.

PR #2580, opened 2024-06-18
Remove pkg/{cilium,oldhubble}

michi-covalent:
The option to enable Cilium APIs got removed in #1541. Now it's safe to remove pkg/{cilium,oldhubble} since Tetragon no longer accesses Cilium state.


kkourt (approved):
+12 −48,827

❤️

PR #2583, opened 2024-06-19
pkg/sensors: fix for ratelimit_map wrong path pinning

mtardy:
Commit 38ab0122a8d69368833457aad6d3d68858c0b561 pinned the ratelimit_map to the fs but used sensorPath instead pinPath since this is a per kprobe map and not a per sensor map.


olsajiri (approved):
lgtm

kkourt (approved):

PR #2587, opened 2024-06-19
Reorg to factor mac entries setup and add a max entries test

olsajiri:
centralizing map entries configuration and adding tests

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -272,14 +272,12 @@ func filterMaps(load *program.Program, pinPath string, kprobeEntry *genericKprob
 	return maps
 }
 
-func createMultiKprobeSensor(sensorPath, policyName string, multiIDs []idtable.EntryID, enableFDInstall bool) ([]*program.Program, []*program.Map, error) {
+func createMultiKprobeSensor(sensorPath, policyName string, multiIDs []idtable.EntryID, has hasMaps) ([]*program.Program, []*program.Map, error) {
 	var multiRetIDs []idtable.EntryID
 	var progs []*program.Program
 	var maps []*program.Map
 
 	data := &genericKprobeData{}
-	oneKprobeHasStackTrace := false
-	oneKprobeHasRatelimit := false
 
 	for _, id := range multiIDs {
 		gk, err := genericKprobeTableGet(id)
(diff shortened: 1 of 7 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -272,14 +272,12 @@ func filterMaps(load *program.Program, pinPath string, kprobeEntry *genericKprob
 	return maps
 }
 
-func createMultiKprobeSensor(sensorPath, policyName string, multiIDs []idtable.EntryID, enableFDInstall bool) ([]*program.Program, []*program.Map, error) {
+func createMultiKprobeSensor(sensorPath, policyName string, multiIDs []idtable.EntryID, has *hasMaps) ([]*program.Program, []*program.Map, error) {
```
mtardy, line 275: nit: maybe I would pass this by value for clarity that it's read only

pkg/sensors/tracing/generickprobe.go
```diff
@@ -552,6 +551,23 @@ func getKprobeSymbols(symbol string, syscall bool, lists []v1alpha1.ListSpec) ([
...
+	fdInstall  bool
+}
+
+func hasMapsSetup(spec *v1alpha1.TracingPolicySpec) *hasMaps {
+	has := &hasMaps{}
+	for _, kprobe := range spec.KProbes {
+		has.fdInstall = has.fdInstall || selectorsHaveFDInstall(kprobe.Selectors)
+		if has.fdInstall {
+			break
+		}
```
mtardy, line 566: you could simplify it, we can always modify if this gets more complicated later
```suggestion
		if selectorsHaveFDInstall(kprobe.Selectors) {
			has.fdInstall = true
			break
		}
```
or
```suggestion
		if selectorsHaveFDInstall(kprobe.Selectors) {
			return &hasMap{
				fdInstall: true,
			}
		}
```

mtardy (commented):
thanks! Yeah the hasMaps make sense, especially for fdinstall. In the singleKprobe part it looks a bit weird that we copy the info in this when it's available on the genericKprobe directly but overall with the multi_kprobe, it's better than this lonely fdinstall bool.

Thanks a lot for the test, I wanted to write something similar to check that those memory patch aren't broken! That's really cool :)

maybe we could document in this hasMap struct which are per kprobe and which are per sensor?

mtardy (approved):
Thanks! lgtm

mtardy (approved):
yep still like it!

PR #2588, opened 2024-06-20
tetragon: resolve uid to username for exec events from /proc fs

tixxdz:
```release-note
tetragon: resolve uid to username for exec events from /proc fs
```

### DIFF
--- a/pkg/sensors/exec/userinfo/userinfo.go
+++ b/pkg/sensors/exec/userinfo/userinfo.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/cilium/tetragon/pkg/api/processapi"
 	"github.com/cilium/tetragon/pkg/grpc/exec"
+	"github.com/cilium/tetragon/pkg/metrics/errormetrics"
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/reader/namespace"
 	"github.com/cilium/tetragon/pkg/reader/userdb"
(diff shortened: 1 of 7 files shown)


pkg/sensors/exec/userinfo/userinfo.go
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
```
kkourt, line 2: The first commit is named  "tetragon: resolve uid->username for processes from /proc", but it seems that it just refactors the code for subsequent patches. Please rename the commit. Also, all of the patches have a single tile. Please add some context what each patch does and why.

pkg/sensors/exec/userinfo/userinfo.go
```diff
@@ -37,6 +38,7 @@ func MsgToExecveAccountUnix(m *exec.MsgExecveEventUnix) error {
 	if option.Config.UsernameMetadata == int(option.USERNAME_METADATA_UNIX) {
 		username, err := getAccountUnix(m.Unix.Process.UID, &m.Unix.Msg.Namespaces)
 		if err != nil {
+			errormetrics.ErrorTotalInc(errormetrics.ProcessMetadataUsernameFailed)
```
kkourt, line 41: Shouldn't we check here if the error is ErrNotInHostNs?

kkourt (changes requested):
Thanks, please find some comments below.

pkg/sensors/exec/userinfo/userinfo.go
tixxdz: Actually it is used in this patch, but improved git log.

pkg/sensors/exec/userinfo/userinfo.go
tixxdz: Makes sense, I made it 'process_metadata_username_ignored_not_in_host_namespaces' a bit too long but it has to be descriptive anyway.

tixxdz:
> Thanks, please find some comments below.

Done, thank you, PTAL!

tixxdz:
backport is here: https://github.com/cilium/tetragon/pull/2603 , when merged will

kkourt (approved):

PR #2600, opened 2024-06-22
tetragon: Allow persistent enforcement during tetragon restart

olsajiri:
needs https://github.com/cilium/ebpf/pull/1496 (merged, waiting for cilium/ebpf release)

Adding support to configure tetragon to keep existing policy in place - create pins for all maps/programs/links and do not remove them when tetragon exits.

This allows the policy to be in place when tetragon process is gone.

At the moment the configuration is by global flag. I have changes for configuring this through the spec 'options' flag, but first I need first to find a good way how to configure base sensor with same option. I think it's better to have this in place now, the option flag can come later.


cmd/tetragon/main.go
```diff
@@ -157,6 +157,34 @@ func stopProfile() {
...
+		log.Info("Found bpf leftover instance, removing: %s", old)
+	}
+	if err := os.Rename(path, old); err != nil {
+		return "", err
+	}
+	log.Infof("Found bpf instance: %s, moved to: %s", path, old)
+	return old, nil
+}
+
+func deleteOlBpfDir(path string) {
```
tpapagian, line 177: Typo: deleteOlBpfDir -> deleteOldBpfDir

cmd/tetragon/main.go
```diff
@@ -157,6 +157,34 @@ func stopProfile() {
...
+	if path == "" {
+		return
+	}
+	if err := os.RemoveAll(path); err != nil {
+		log.Errorf("Failed to remove old bpf isntance '%s': %s\n", path, err)
```
tpapagian, line 182: Typo: isntance

tpapagian (approved):
LGTM, thanks!

Just a quick question: Is `link.Pin` also supported in the other types of programs that Tetragon supports (i.e. LSM, Tracing, Uprobes, RawTracepoints)? If yes, I believe that it is a good idea to update those as well. In that case, we also need to have a feature detection function using the same type of program instead of kprobes?

olsajiri:
> LGTM, thanks!
> 
> Just a quick question: Is `link.Pin` also supported in the other types of programs that Tetragon supports (i.e. LSM, Tracing, Uprobes, RawTracepoints)? If yes, I believe that it is a good idea to update those as well. In that case, we also need to have a feature detection function using the same type of program instead of kprobes?

yes, those should be supported as wee, I'll do follow up with that support, thanks

kkourt (approved):
LGTM, thanks!

PR #2604, opened 2024-06-24
Define health metrics group

lambdanis:
Depends on #2606

There are two things happening here:
1. Refactored custom collectors to use new helpers from `pkg/metrics` (commits 1-3). The main win here is simplifying how custom metrics are included in reference docs - we don't need to define a separate type and register it separately just for generating docs.
2. Defined a health metrics group (last commit). The idea is that in the future this group will have constrained cardinality and will be enabled by default (in contrast to another group with potentially high cardinality "debug" metrics). This commit only refactors the existing metrics initialization code to use the new framework. The health metrics group contains all metrics that were documented in the "health metrics" section, but in the future some of them will likely be moved to another group.


pkg/metrics/labels.go
```diff
@@ -38,3 +43,23 @@ func promContainsLabel(labels prometheus.ConstrainedLabels, label string) bool {
 	}
 	return false
 }
+
+// TODO: Standardize labels used by different metrics: op, msg_op, opcode.
```
tixxdz, line 47: Maybe open issues for these?

pkg/metrics/errormetrics/errormetrics.go
```diff
@@ -84,10 +85,17 @@ var (
...
+func RegisterMetrics(group metrics.Group) {
+	group.MustRegister(ErrorTotal)
+	group.MustRegister(HandlerErrors)
+}
 
+// TODO:
+//  1. Define metrics using functions from pkg/metrics
+//  2. Move initialization code to metrics definitions if needed or remove it
+//     if not needed
+//  3. Use label values defined as metrics.ConstrainedLabel
```
tixxdz, line 97: Do these still make sense? and if so maybe put them in issues?

tixxdz (approved):
Thank you @lambdanis

pkg/metrics/errormetrics/errormetrics.go
lambdanis: Yes, definitely, I'll create issues :+1:

PR #2610, opened 2024-06-26
Move metrics registry and initialization to pkg/metricsconfig

lambdanis:
Before, metrics registry was defined in `pkg/metrics` and initialization functions in `pkg/metrics/metricsconfig`. It make sense to keep them together and outside of `pkg/metrics`, which contains a library for defining metrics. Now all code related to running the metrics server sits in `pkg/metricsconfig`.

I took this out of #2604 to make it easier to stack PRs.


kevsecurity (approved):
Seems 😇

PR #2612, opened 2024-06-27
tetragon-metrics-docs: Move New function to a separate package

lambdanis:
Introduce metricsmd package containing New function from cmd/tetragon-metrics-docs. This is done so that this function can be imported and reused by different commands.


kkourt (approved):

michi-covalent (approved):
✨

PR #2619, opened 2024-06-28
[backport/v0.10] Fix libbpf submodule checkout in bpftool

tpapagian:
[upstream commit 0235ed57252b9e03c68ea1dbc6ad0b1b6b5e6298]

Now in the Dockerfile we do:
git clone --recurse-submodules https://github.com/libbpf/bpftool.git . && git checkout ${BPFTOOL_REV} which means that we fetch the libbpf submodule version of the main branch of bpftool and then we move to the appropriate commit that we need. After the checkout we also need to fetch the correct version of the submodules. This patch fixes that.


kkourt (approved):

tpapagian:
It seems that more fixes needed in that branch.

PR #2623, opened 2024-07-02
bpf: fix missed linux_binprm_type in selector_arg_offset function

anfedotoff:
It's seems to me, that `case linux_binprm_type:` in [selector_arg_offset](https://github.com/cilium/tetragon/blob/main/bpf/process/types/basic.h#L1617) function is missing. [security_bprm_check.yaml](https://github.com/cilium/tetragon/blob/main/examples/tracingpolicy/security_bprm_check.yaml) from examples doesn't work.

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1686,6 +1686,9 @@ selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx,
 			args += 4;
 		case file_ty:
 		case path_ty:
+#ifdef __LARGE_BPF_PROG
+		case linux_binprm_type:
+#endif
 			pass &= filter_file_buf(filter, (struct string_buf *)args);
 			break;
 		case string_type:
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -5989,6 +5989,9 @@ spec:
 }
 
 func TestLinuxBinprmExtractPath(t *testing.T) {
+	if !kernels.EnableLargeProgs() {
+		t.Skip("Older kernels do not support matchArgs with linux_binprm")
+	}
 	testutils.CaptureLog(t, logger.GetLogger().(*logrus.Logger))
 	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
 	defer cancel()
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -6028,7 +6028,7 @@ func TestLinuxBinprmExtractPath(t *testing.T) {
 						{
 							MatchArgs: []v1alpha1.ArgSelector{
 								{
-									Operator: "In",
+									Operator: "Equal",
 									Index:    0,
 									Values:   []string{"/usr/bin/id"},
 								},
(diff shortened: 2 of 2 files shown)


bpf/process/types/basic.h
```diff
@@ -1686,6 +1686,7 @@ selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx,
 			args += 4;
 		case file_ty:
 		case path_ty:
+		case linux_binprm_type:
```
kevsecurity, line 1689: This is great, but to match up with the reading of the args, please can you wrap this single line in:
```
#ifdef __LARGE_BPF_PROG

#endif
```
So we only try to match if we've actually read the arg. Otherwise, thanks!

kevsecurity (changes requested):

kevsecurity (approved):
LGTM

kkourt:
There is a test failure, which seems related:
> FAIL: TestLinuxBinprmExtractPath (1.12s)

kevsecurity (changes requested):
Need to work out why the ARM test is failing.

anfedotoff:
Yes, I'll look at failing tests.

kevsecurity:
> Yes, I'll look at failing tests.

It might be that the tests themselves are broken.

mtardy:
Yep it seems test itself is broken:
https://github.com/cilium/tetragon/blob/e44f7dfc0e8ee90ddcef1d915111d420917b9ae0/pkg/sensors/tracing/kprobe_test.go#L6033
It should be "Equal", "In" does not exist for `matchArgs`.

It bypasses validation since it doesn't pass YAML unmarshalling.

I just tried changing the exec command and the test passes all the time, it needs a fix!

mtardy:
@anfedotoff I have a patch for the test that actually fails without your patch and succeeds with your patch (which should have been what happened previously), is it okay if I push to your branch?

mtardy:
> @anfedotoff I have a patch for the test that actually fails without your patch and succeeds with your patch (which should have been what happened previously), is it okay if I push to your branch?

I created https://github.com/cilium/tetragon/pull/2624 so that I can run the tests from there. We can merge like this with your commit authored to you. Or I can merge the fix of the test and then we fix your patch, however you prefer.

anfedotoff:
> @anfedotoff I have a patch for the test that actually fails without your patch and succeeds with your patch (which should have been what happened previously), is it okay if I push to your branch?

Thanks, yes, sure!

anfedotoff:
> > @anfedotoff I have a patch for the test that actually fails without your patch and succeeds with your patch (which should have been what happened previously), is it okay if I push to your branch?
> 
> I created https://github.com/cilium/tetragon/pull/2624 so that I can run the tests from there. We can merge like this with your commit authored to you. Or I can merge the fix of the test and then we fix your patch, however you prefer.

I think, the easiest way is to merge your PR with my commit  and your test fix.

mtardy:
> > > @anfedotoff I have a patch for the test that actually fails without your patch and succeeds with your patch (which should have been what happened previously), is it okay if I push to your branch?
> > 
> > I created https://github.com/cilium/tetragon/pull/2624 so that I can run the tests from there. We can merge like this with your commit authored to you. Or I can merge the fix of the test and then we fix your patch, however you prefer.
> 
> I think, the easiest way is to merge your PR with my commit  and your test fix.

Ok we can do that, then given your patch, my test failed on 4.19 since it doesn't have large progs. If it's expected we should just skip this test!

anfedotoff:
> > > > @anfedotoff I have a patch for the test that actually fails without your patch and succeeds with your patch (which should have been what happened previously), is it okay if I push to your branch?
> > > 
> > > I created https://github.com/cilium/tetragon/pull/2624 so that I can run the tests from there. We can merge like this with your commit authored to you. Or I can merge the fix of the test and then we fix your patch, however you prefer.
> > 
> > I think, the easiest way is to merge your PR with my commit  and your test fix.
> 
> Ok we can do that, then given your patch, my test failed on 4.19 since it doesn't have large progs. If it's expected we should just skip this test!

Yes, I think it is expected. We use large progs to copy args of this type,  if I'm not mistaken.

kevsecurity:
Hmm, why's it failing some tests on bpf-next?

anfedotoff:
> Hmm, why's it failing some tests on bpf-next?

It seems to me, that changes it the PR don't influence on this test. I rebased this branch on current main, let's look at the tests.

mtardy (approved):
Thanks! I think the fix is complete now :)

kevsecurity (approved):

PR #2625, opened 2024-07-02
Improve Helm values docs

lambdanis:
I wanted to improve doc comment for `exportFilePerm` value, but noticed other issues in Helm values docs, so fixed them too. See commits.


install/kubernetes/tetragon/values.yaml
```diff
@@ -30,8 +30,6 @@ serviceLabelsOverride: {}
 dnsPolicy: Default
 # -- exportDirectory specifies directory to put Tetragon JSON export files.
 exportDirectory: "/var/run/cilium/tetragon"
-# -- exportFileRotationInterval specifies file creation interval for hubble-export-s3.
-exportFileCreationInterval: "120s"
```
tixxdz, line 43: How it is not used when it is defined? what am I missing?

tixxdz: it is rotation not creation...

michi-covalent (approved):
✨

PR #2630, opened 2024-07-04
deprecate sensors gRPC API

kkourt:
The gRPC sensors interface is the initial interface for manging collections of bpf programs at run-time. It has been replaced by tracing policies. This PR deprecates the sensor management gRPC calls.

Fixes: https://github.com/cilium/tetragon/issues/1272

CC: @inliquid


mtardy (approved):
I think we are okay to deprecate since the TracingPolicy can do everything one should need now. Note that the linter doesn't like that we use (our own) deprecated API, we should add a "ignore".

kkourt:
> I think we are okay to deprecate since the TracingPolicy can do everything one should need now. Note that the linter doesn't like that we use (our own) deprecated API, we should add a "ignore".

Thanks, added an ignore.

inliquid:
Thanks! We moved to TracingPolicy API beforehand so this doesn't affect us.

olsajiri (approved):
lgtm

Issue #2545, opened 2024-06-12
Dynamic log level change doesn't work the same for all environments


PhilipSchmid:
### What happened?

The Tetragon doc has a [Change log level dynamically](https://tetragon.io/docs/troubleshooting/#change-log-level-dynamically) section that explains how to dynamically change it using `SIGRTMIN+x`. However, `SIGRTMIN` isn't the same value for every environment:

> The  Linux  kernel  supports a range of 33 different real-time signals, numbered 32 to 64.  However, the glibc
POSIX threads implementation internally uses two (for NPTL) or three (for LinuxThreads) real-time signals (see
pthreads(7)), and adjusts the value of SIGRTMIN suitably (to 34 or 35).

It depends on libc, and, e.g., Alpine uses musl: https://git.musl-libc.org/cgit/musl/commit/src/signal/sigrtmin.c?id=016a5dc1925a66c7d1ffc14b862b1342f399cfda

We would likely need to develop a proper `tetra` CLI command to set the debug level in a unified and environment-independent way.

I ran into this issue while deploying Tetragon on a Talos Linux environment. For reproduction, you could use https://github.com/isovalent/terraform-aws-talos/tree/main/example.

Relevant Tetragon part: https://github.com/cilium/tetragon/blob/e44f7dfc0e8ee90ddcef1d915111d420917b9ae0/cmd/tetragon/main.go#L310-L329

### Tetragon Version

v1.1.0

### Kernel Version

6.6.29-talos

### Kubernetes Version

v1.29.3

PR #2643, opened 2024-07-05
tetra: Added dynamic log level change option

PhilipSchmid:
A tetra loglevel subcommand was added, which allows one to change Tetragon's log level dynamically without restarting it.

Supported subcommands:

```bash
  get         Prints the current log level
  set         Set the log level
```

Fixes: #2545

```release-note
tetra: Added dynamic log level change option
```

I manually tested it from a Linux and macOS client, issuing all sorts of `tetra loglevel *` commands. Worked flawlessly. Unfortunately, I don't have a Windows client, so I can't test it from there, but I'd expect the same results as it doesn't depend on local files or binaries.


pkg/server/server.go
```diff
@@ -378,3 +378,41 @@ func (s *Server) RuntimeHook(ctx context.Context, req *tetragon.RuntimeHookReque
...
+
+func (s *Server) SetLogLevel(ctx context.Context, req *tetragon.SetLogLevelRequest) (*tetragon.SetLogLevelResponse, error) {
+	currentLogLevel := logger.GetLogLevel()
+	changedLogLevel, err := logrus.ParseLevel(req.GetLevel())
+	if err != nil {
+		return nil, fmt.Errorf("invalid log level: %s", req.GetLevel())
+	}
+
+	// Temporarily set the logger to the lowest level to ensure the message is logged
+	logger.SetLogLevel(logrus.TraceLevel)
```
jrfastab, line 395: Will this impact other routines also using logger here? I'm not sure about internals of logging but we wouldn't want to shortly spew debug/trace info if another sensor is running.

Why not just make it a Warn if its critical to have printed? Seems cleaner than bouncing loglevel around.

pkg/server/server.go
```diff
@@ -378,3 +378,41 @@ func (s *Server) RuntimeHook(ctx context.Context, req *tetragon.RuntimeHookReque
...
+	logger.GetLogger().WithField("request", req).Infof("Log level changed from %s to %s", currentLogLevel, changedLogLevel.String())
+
+	// Set the logger back to the requested level
+	logger.SetLogLevel(changedLogLevel)
+	return &tetragon.SetLogLevelResponse{Level: changedLogLevel.String()}, nil
+}
+
+func (s *Server) ResetLogLevel(ctx context.Context, req *tetragon.ResetLogLevelRequest) (*tetragon.ResetLogLevelResponse, error) {
+	// TODO: Get the original log level with which Tetragon was originally started with in a nicer way than via "option.Config.LogOpts["level"]".
```
jrfastab, line 404: I would either drop the todo and create a first issue or do this ;)

jrfastab:
Like the idea started reviewing before I noticed it was draft. But would be a good addition thanks for working on it.

pkg/server/server.go
PhilipSchmid: Thanks for the input. Addressed via https://github.com/cilium/tetragon/pull/2643/commits/2b90834c022e35c50dc6515840a3e228d933b31d.

pkg/server/server.go
PhilipSchmid: Thanks for input. Addressed via https://github.com/cilium/tetragon/pull/2643/commits/d0c4133f016a1f4533a898440a4d22ad8d031f57.

PhilipSchmid:
Added minimal doc:
image

mtardy (commented):
Hey, sorry I have not reviewed that until now.

As we discussed offline I have no strong feeling over that and thought that it was maybe a bit overkill to dedicate 3 RPCs for this, I would see how we could extend that in the future to do more debugging things:
- enabling gops
- enable pprof

Maybe we could factor all that into `SetDebug` and have more options for evolution so that we don't need many calls. Not sure Reset or Get is particularly needed but maybe I'm missing something.

So to recap my opinion would be to reduce this to 1 RPC that could be extended. But honestly, I'm not sure that's better so that would be cool to have other peoples' opinion or just merge this if it's satisfying to someone. It would be cool to add a test though!

Overall the PR is great and thanks for adding docs that's amazing.

pkg/server/server.go
```diff
@@ -378,3 +378,30 @@ func (s *Server) RuntimeHook(ctx context.Context, req *tetragon.RuntimeHookReque
...
+}
+
+func (s *Server) SetLogLevel(_ context.Context, req *tetragon.SetLogLevelRequest) (*tetragon.SetLogLevelResponse, error) {
+	currentLogLevel := logger.GetLogLevel()
+	changedLogLevel, err := logrus.ParseLevel(req.GetLevel())
+	if err != nil {
+		return nil, fmt.Errorf("invalid log level: %s", req.GetLevel())
+	}
+	logger.SetLogLevel(changedLogLevel)
+	logger.GetLogger().WithField("request", req).Warnf("Log level changed from %s to %s", currentLogLevel, changedLogLevel.String())
```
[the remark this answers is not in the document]
PhilipSchmid: Getting the current value doesn't have any impact on the Tetragon instance, so debug should be sufficient, while changing the Tetragon instance's log level could have an impact in a way that warnings/errors are afterwards not seen anymore. Hence, I thought logging this change as a warning makes sense. Resetting the log level to the Tetragon instance's initial level should be a "neutral" operation. Hence, info level.

api/v1/tetragon/sensors.proto
```diff
@@ -153,4 +170,8 @@ service FineGuidanceSensors {
     rpc GetVersion(GetVersionRequest) returns (GetVersionResponse) {}
 
     rpc RuntimeHook(RuntimeHookRequest) returns (RuntimeHookResponse) {}
+
+	rpc GetLogLevel(GetLogLevelRequest) returns (GetLogLevelResponse) {}
+	rpc SetLogLevel(SetLogLevelRequest) returns (SetLogLevelResponse) {}
+	rpc ResetLogLevel(ResetLogLevelRequest) returns (ResetLogLevelResponse) {}
```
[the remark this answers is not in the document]
PhilipSchmid: Good catch! Fixed in the latest force push, in which I also rebased it to `main`.

PhilipSchmid:
@mtardy Thanks a lot for your valuable input!

You're right. Adding three different gRPC endpoints solely for this is an overkill 😅 . Let's also use this to enable gops and pprof on demand (if technically doable without a restart - I still need to check that).
Furthermore, I'd say we can reduce that to two generic gRPC endpoints. Maybe `SetDebugConfig` and `GetDebugConfig`? I still think we should have the GET option to provide better UX for the user using the `tetra` client. Only with this is it possible to fetch the current state of a specific debug-related option from the Tetragon instance. Don't you think?

PhilipSchmid:
(Fixed two failed CI tests with the latest two force pushes.)

PhilipSchmid:
I just added basic tests for `server.go`. See https://github.com/cilium/tetragon/pull/2643/commits/46be259261bf8fafca6efd84400057f86b7b62db. I didn't add tests for the `tetra` part as it would require some Mock setup, which, AFAICT, we currently don't have.

contrib/tetragon-rthooks/vendor/github.com/cilium/tetragon/api/v1/tetragon/sensors.proto
```diff
@@ -125,6 +125,38 @@ message GetVersionResponse{
 	string version = 1;
 }
 
+// For now, we only want to support debug-related config flags to be configurable.
+enum ConfigFlag {
+	CONFIG_FLAG_LOG_LEVEL = 0;
+}
+
+enum LogLevel {
```
PhilipSchmid, line 133: Maps to https://github.com/sirupsen/logrus/blob/master/logrus.go#L93-L112

pkg/server/server.go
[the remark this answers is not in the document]
jrfastab: ok I think I can believe above. We can change it later if we decide its not needed.

jrfastab (approved):

mtardy:
Thanks, sorry for the delay I was on PTO 🏝️

mtardy:
Amazing, it seems we have a first candidate to extend this debug RPC with process cache dump, Anastasios was thinking adding a RPC just for that might be overkill and we just have this now here, https://github.com/cilium/tetragon/pull/2246. Thanks @PhilipSchmid for making this more generic, it seems it will be useful.

kkourt:
See also: https://github.com/cilium/tetragon/pull/2876

PR #2657, opened 2024-07-11
Fix linter warnings in Dockerfiles

lambdanis:
Silly changes to fix linter warnings:
```
FromAsCasing: 'as' and 'FROM' keywords' casing do not match
LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format
```


mtardy (approved):
is it a linter you ran locally?

lambdanis:
> is it a linter you ran locally?

It's part of the images builds, locally and everywhere else.

PR #2664, opened 2024-07-12
release_template: also push API tag

kkourt:
See commit.


mtardy (approved):

mtardy (approved):
I think it's good because we filter correctly in the Makefile, it produced some issues to add new tags prefixes in the past but I think the regex is correct now.

kkourt:
> I think it's good because we filter correctly in the Makefile, it produced some issues to add new tags prefixes in the past but I think the regex is correct now.

Yap, I changed the VERSION command when I added yet another tag for rthooks images (rthooks/v*)

PR #2670, opened 2024-07-12
Makefile: Reduce the amount of vendoring calls

lambdanis:
When generating Go code for proto API or k8s CRDs, we don't need to run vendoring for all modules before and after. It should be enough if we run vendoring only for the relevant module (api or pkg/k8s) before, and only for the main module after.


mtardy (approved):
Nice, if this is enough, it will make the generation steps more lightweight.

PR #2673, opened 2024-07-14
Improve local k8s developer experience

lambdanis:
This set of changes is supposed to improve the developer experience using local kind clusters. It's based on my own experience, so if I'm breaking anyone's workflow then please shout.

The most important changes:
- `make kind-install-tetragon` and `install-tetragon.sh` script install Tetragon in tetragon namespace, not kube-system
- `make kind-install-tetragon` builds images by default
- You can pass additional Helm values to `make kind-install-tetragon` using `VALUES=<path>` option
- Default Helm values used are now in `values.yaml` file instead of being scattered around `install-tetragon.sh` script

Ref #1849


michi-covalent (approved):
i trust you

PR #2676, opened 2024-07-16
Remove deprecated gRPC `Dial` calls

### LINKED ISSUE #2675: Replace deprecated `grpc.DialContext`
### Context

[gRPC  Go]( https://github.com/grpc/grpc-go) deprecated `DialContext` with version `v1.63.0`.  We consistently use `DialContext` with `grpc.WithBlock()`. So switching to the recommended `NewClient` would change the behavior.

We need to figure out if we should change the behavior or somehow keep the current behavior while using `NewClient`.

### Acceptance Criteria

* The deprecated `DialContext` calls are replaced
* Related deprecated DialOptions are removed

### Related Links
* Upstream points out our current approach is an anti-pattern: https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md#dialing-in-grpc

mtardy:
Fixes #2675.

Consider also that now with those patches, the `retries` flag is no longer doing anything, since it was only retrying on the initial connection (while `timeout` now properly works on all calls). This could be a follow-up PR I think: remove it or add a helper to perform retry on the actual call in case of gRPC errors (and not the actual underlying method error)?


willfindlay:
Nice, taking a look shortly.

cmd/tetra/common/flags.go
```diff
@@ -26,3 +35,44 @@ var (
...
+
+		logger.GetLogger().WithFields(logrus.Fields{
+			"InitInfoFile":  defaults.InitInfoFile,
+			"ServerAddress": sa,
+		}).Debug("resolved server address using info file")
+		return sa, nil
+	}
+
+	return ServerAddress, nil
+}
```
willfindlay, line 78: We're always returning nil here. Perhaps we could consider removing the error return in the type signature?

willfindlay (commented):
>DialContext got deprecated, so we switch to NewClient that does not
connect to the server, it waits for the first RPC to be called to
establish a connection, thus the retry mechanism at connect no longer
make sense. It could be reimplemented as a retry helper for the RPC
call.

This makes sense to me. Assuming this is true and we establish a new connection on the first RPC call (and not a new connection on every subsequent call), why don't we have the client call a healthcheck RPC or similar (I think we already have one) and put the old retry logic in there?

cmd/tetra/common/flags.go
mtardy: indeed thanks

mtardy:
> why don't we have the client call a healthcheck RPC or similar (I think we already have one) and put the old retry logic in there

I would not do RPC checks that are not asked by the user, it would slow down and would be unexpected if it breaks no? I would maybe do the retry on the request itself or just not do it and let people use retry mechanism on `tetra` with external utility, using the status code returned for example.

willfindlay:
> > why don't we have the client call a healthcheck RPC or similar (I think we already have one) and put the old retry logic in there
> 
> I would not do RPC checks that are not asked by the user, it would slow down and would be unexpected if it breaks no? I would maybe do the retry on the request itself or just not do it and let people use retry mechanism on `tetra` with external utility, using the status code returned for example.

I don't have super strong feelings one way or another, but I do think it would be nice to support retries. Especially since we've had a lot of problems with connection timeouts in tetra in the past.

mtardy:
> > > why don't we have the client call a healthcheck RPC or similar (I think we already have one) and put the old retry logic in there
> > 
> > 
> > I would not do RPC checks that are not asked by the user, it would slow down and would be unexpected if it breaks no? I would maybe do the retry on the request itself or just not do it and let people use retry mechanism on `tetra` with external utility, using the status code returned for example.
> 
> I don't have super strong feelings one way or another, but I do think it would be nice to support retries. Especially since we've had a lot of problems with connection timeouts in tetra in the past.

This is an idea on how it could be done, c8110611fce5ac3db9114af1dcc36d69c720fef8. It would need a rewrite for those using `CliRun` that don't return the error out of the closure.

I'll see if I can make work gRPC build-in retry mechanism before trying to extend this. However this work as expected and works along interrupt and timeout which is nice (it wasn't the case before).

willfindlay:
Hm it seems that all the new OLM tests are failing. And the `Copy out e2e test logs` step seems to be broken. :cry:

mtardy:
> but I do think it would be nice to support retries. Especially since we've had a lot of problems with connection timeouts in tetra in the past.

Okay! I spent some amount of time to satisfy you on this one, I deep dive into their built-in backoff mechanism, see a696b32bc5a8256d9f666d767d042b5de4fa3fd5. Now it should work and it's transparent!

willfindlay:
> Now it should work and it's transparent!
That's amazing :open_mouth:

willfindlay (approved):

PR #2681, opened 2024-07-16
fix(docs): update getting started instructions

### LINKED ISSUE #2680: docs: getting started guide assumes a single node cluster
The [getting-started](https://tetragon.io/docs/getting-started/) guides, for example, the [execution monitoring](https://tetragon.io/docs/getting-started/execution/) assumes we are running a single node cluster. It might create confusion as events can be seen as missing if:
1. We are running a multi-node cluster.
2. The events created by the trigger mechanism are running on node A, for example the xwing pod is running on node A.
3. The `tetra getevents` command runs on another node, node B. Using `kubectl exec -ti -n kube-system ds/tetragon -c tetragon -- tetra getevents -o compact --pods xwing` that redirect to a random Pod from the daemonset might redirect to node B or C, etc.

The result is that `tetra getevents` from node B will not see the event triggered on node A and it looks like tetragon doesn't work.

Here are some suggestions:
- Modify the `kubectl exec -ti -n kube-system ds/tetragon -c tetragon` to make sure it picks the correct node instance (the one that the trigger pod is running on), might be the cleaner solution since all guides are using the xwing pod. I'm guessing we can have a reasonable command that: retrieves the xwing pod's node, run `kubectl exec` to the tetragon pod on that node.
- Aggregate the information from all tetragon pods using the logs, IIRC k8s can do that using `kubectl logs` on the ds. But the downside is that tetra must be installed locally if we want to display the data as compact and filter on it. Doing something like `kubectl logs ds/tetragon | tetra getevents -o compact --pods xwing`. Or maybe we can do something nasty and reuse any tetragon pod to run tetra without installing it like: `alias tetra="kubectl exec -ti -n kube-system ds/tetragon -c tetragon -- tetra"`, not sure it's nice...
- Add a huge warning in the docs that the guides only work for a single node cluster.

Thanks to @scottslowe for raising this issue! :)

scottslowe:
Thanks for contributing! Please ensure your pull request adheres to the following guidelines:

- [x] All commits contain a well written commit message and are signed-off (see [Submitting a pull request](https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/)).
- [ ] All code is covered by unit and/or end-to-end tests where feasible.
- [ ] All generated files are updated if needed (see [Making changes](https://tetragon.io/docs/contribution-guide/making-changes/)).
- [ ] Provide a title or release-note blurb suitable for the release notes (see [guidelines](https://tetragon.io/docs/contribution-guide/docs/release-notes/#release-note-blurb-in-pr)).
- [ ] Update documentation and write an upgrade note if needed (see [guidelines](https://tetragon.io/docs/contribution-guide/docs/release-notes/#upgrade-notes)).

Update Kubernetes install instructions to call out single node cluster assumption.

Update execution monitoring instructions to provide commands for using multi-node clusters.

Fixes #2680

### DIFF
--- a/docs/content/en/docs/getting-started/execution.md
+++ b/docs/content/en/docs/getting-started/execution.md
@@ -4,38 +4,50 @@ weight: 3
 description: "Execution traces with Tetragon"
 ---
 
-At the core of Tetragon is the tracking of all executions in a kubernetes cluster,
-virtual machines, and baremetal systems. This creates the foundation that allows
+At the core of Tetragon is the tracking of all executions in a Kubernetes cluster,
+virtual machines, and bare metal systems. This creates the foundation that allows
 Tetragon to attribute all system behavior back to a specific binary and its
 associated metadata (container, Pod, Node, and cluster).
 
-## Observe Tetragon Execution Events
+## Observe Tetragon execution events
 
-Tetragon exposes the execution trace over JSON logs and GRPC stream. The user
+Tetragon exposes the execution events over JSON logs and GRPC stream. The user
 can then observe all executions in the system.
 
-The following command can be used to observe exec events.
+Use the following instructions to observe execution events. These instructions
+assume you have deployed the Cilium [demo application](https://docs.cilium.io/en/stable/gettingstarted/demo/)
+in your environment.
 
 {{< tabpane text=true >}}
-{{% tab Kubernetes %}}
+{{% tab "Kubernetes (single node)" %}}
 
-For a single node Kubernetes cluster, you can target the Tetragon DaemonSet with a `kubectl exec` command:
+For a single node Kubernetes cluster, you can target the Tetragon DaemonSet with
+a `kubectl exec` command:
 
 ```shell
 kubectl exec -ti -n kube-system ds/tetragon -c tetragon -- tetra getevents -o compact --pods xwing
 ```
 
+This command runs `tetra getevents -o compact --pods xwing` in the single Pod
+that is a member of the Tetragon DaemonSet. Because there is only a single node
+in the cluster, it is guaranteed that the "xwing" Pod will also be running on
+the same node and that Tetragon will be able to capture and report execution
+events.
+
+{{% /tab %}}
+{{% tab "Kubernetes (multiple nodes)" %}}
+
 In a cluster with multiple nodes, you will first need to find the node where
-the workload is running. For example, when using Cilium's [demo application](https://docs.cilium.io/en/stable/gettingstarted/demo/),
-this command will show on which node the "xwing" Pod is running.
+the workload is running. This command shows the node where the "xwing" Pod is
+running.
 
 ```shell
 kubectl get pod xwing -o jsonpath='{.spec.nodeName}'
 ```
 
-You would then find the Tetragon Pod on the same node; this command returns a
-list of Tetragon Pods and the name of the Kubernetes node on which they are
-running.
+Next, you will need to find the Tetragon Pod on the same node. This command
+returns a list of Tetragon Pods and the name of the Kubernetes node on which
+they are running.
 
 ```shell
 kubectl -n kube-system get pods -o custom-columns=NAME:.metadata.name,NODE:.spec.nodeName | grep <name-of-node-with-workload>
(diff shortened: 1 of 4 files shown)


mtardy (changes requested):
Hey, thanks for opening that PR.

However, having those two different paths might overcomplicate the getting-started. I feel like we could use this command:

```shell
POD=$(kubectl get pod -l 'app.kubernetes.io/name=tetragon' -o name -n kube-system --field-selector spec.nodeName=$(kubectl get pod xwing -o=jsonpath='{.spec.nodeName}')); kubectl exec -ti -n kube-system $POD -c tetragon -- tetra getevents -o compact --pods xwing
```

To replace the existing:
```shell
kubectl exec -ti -n kube-system ds/tetragon -c tetragon -- tetra getevents -o compact --pods xwing
```

I haven't found a way to make it simpler, the better option would have been to create the xwing pod directly on the first node of the daemonset so that ds/tetragon directly selects the appropriate node but this is impossible if we continue to use the `.yaml` file of cilium. We could just deploy the xwing using `kubectl run` but the final getting started guide uses the deathstar thingie. So maybe it's simpler to just use a more complicated command to run on the appropriate tetragon node.

The other solution would have been to use the aggregated logs:
```shell
kubectl logs -l app.kubernetes.io/name=tetragon -c export-stdout -f -n kube-system | kubectl exec -i -n kube-system ds/tetragon -c tetragon -- tetra getevents -o compact --pods xwing
```
But this contains historic data which can be confusing for the user I guess!

So I would suggest to replace all the command with the one retrieving the correct tetragon POD. You can find all those references using:

scottslowe:
@mtardy I'm not sure that one long command is any better for users than taking a bit of space to explain what needs to be done and why. IMHO, it's clearer to explain what needs to be done, provide commands for each step, and then (at/near the end) provide a single command that encapsulates everything. I'll also point out that users with single node clusters can follow the multi-node instructions without any errors (they're just taking additional steps that aren't necessary _at this moment_).

To reduce any potential confusion, what do you think about restructuring the tabs to show:

* Kubernetes (single node)
* Kubernetes (multi-node)
* Docker

And then providing instructions in each section? It adds some complexity to the docs themselves, but provides clear distinction/separation for users on what set of instructions to follow.

scottslowe:
@mtardy I'd also suggest that we focus on fixing only the "Execution Monitoring" section in this PR. Once we are happy with how it looks and works, then I can file an additional PR to model the same changes in the other sections.

mtardy:
> @mtardy I'm not sure that one long command is any better for users than taking a bit of space to explain what needs to be done and why. IMHO, it's clearer to explain what needs to be done, provide commands for each step, and then (at/near the end) provide a single command that encapsulates everything. I'll also point out that users with single node clusters can follow the multi-node instructions without any errors (they're just taking additional steps that aren't necessary _at this moment_).
> 
> To reduce any potential confusion, what do you think about restructuring the tabs to show:
> 
> * Kubernetes (single node)
> * Kubernetes (multi-node)
> * Docker
> 
> And then providing instructions in each section? It adds some complexity to the docs themselves, but provides clear distinction/separation for users on what set of instructions to follow.

yeah that's a good idea as well. I just want to avoid the proposed changes that are confusing because we end up with 4 commands and a lot of text instead of one command.

On that:
> IMHO, it's clearer to explain what needs to be done, provide commands for each step, and then (at/near the end) provide a single command that encapsulates everything.

I think I don't agree, for a "getting started" quick guide that must show results as fast as possible. I would say for a more in-depth part of the doc yes but maybe not here.

> I'd also suggest that we focus on fixing only the "Execution Monitoring" section in this PR. Once we are happy with how it looks and works, then I can file an additional PR to model the same changes in the other sections. I think this would make reviews easier (smaller change sets to review).

It's up to you, I don't mind. I find it okay to review any case.

docs/content/en/docs/getting-started/execution.md
```diff
@@ -4,67 +4,128 @@ weight: 3
...
+```shell
+kubectl -n kube-system get pods -o custom-columns=NAME:.metadata.name,NODE:.spec.nodeName | grep <name-of-node-with-workload>
+```
+
+Once you have the identified the matching Pod, then target it with a `kubectl
+exec` to run the `tetra getevents` command.
+
+```shell
+kubectl exec -ti -n kube-system po/<tetragon-pod-name> -c tetragon -- tetra getevents -o compact --pods xwing
+```
```
mtardy, line 61: Could you condense those three commands into two using shell variable so that it's "copy pastable"?

```shell
POD=$(kubectl get pod -l 'app.kubernetes.io/name=tetragon' -o name -n kube-system --field-selector spec.nodeName=$(kubectl get pod xwing -o=jsonpath='{.spec.nodeName}'))
// stuff
kubectl exec -ti -n kube-system $POD -c tetragon -- tetra getevents -o compact --pods xwing
```

scottslowe: For the record, I'm not a fan of a single command that doesn't explain what it's doing.

I'd rather get these changes into the docs than get hung up on a minor difference of opinion, so I've added the two commands as suggested above in commit bc4204e.

mtardy (approved):
> For the record, I'm not a fan of a long command that doesn't explain what it's doing.
>
> I'd rather get these changes into the docs than get hung up on a minor difference of opinion, so I've added the two commands as suggested above in commit https://github.com/cilium/tetragon/commit/bc4204e896adbe58490e2a0ba5cdff26e657b4a3.

I agree this is debatable!

I'd even merge the Kubernetes single node and Kubernetes multiple node to simplify further but this is way better than previously thanks! :)

mtardy:
I forgot to ask for squashing the commits oups

docs/content/en/docs/getting-started/install-docker.md
```diff
@@ -23,15 +23,39 @@ The easiest way to start experimenting with Tetragon is to run it via Docker
 using the released container images.
 
 ```shell
-docker run --name tetragon-container --rm --pull always \
+docker run -d --name tetragon --rm --pull always \
```
mtardy, line 26: btw why did we change that? I think running interactively was fine to get the logs and everything.

docs/content/en/docs/getting-started/install-docker.md
```diff
@@ -23,15 +23,39 @@ The easiest way to start experimenting with Tetragon is to run it via Docker
...
+```shell
+curl -LO https://github.com/cilium/tetragon/raw/main/examples/quickstart/docker-compose.yaml
+docker compose -f docker-compose.yaml up -d
+```
+
+You can use `docker container ls` to verify that the containers are up and
+running. It might take a short amount of time for the container images to
+download.
+
+```shell
```
mtardy, line 52: forgot to add to remove shell

docs/content/en/docs/getting-started/install-docker.md
scottslowe: This made things a bit more similar between Docker and Kubernetes.

docs/content/en/docs/getting-started/install-docker.md
scottslowe: So you want no language designator here, just the backticks?

mtardy: yes, it's a nit again, but it makes the input look different than the output, this is subtle but this is the convention we have in the doc:
image

PR #2685, opened 2024-07-17
tests: replace `panic` with `t.Fatal`, upgrade to latest base image for vmtests and fix enforcer tests

kevsecurity:
Related to https://github.com/cilium/tetragon/issues/2686.

Panicking in tests is bad as it fails to tell the test handler that the test failed. Instead we should use t.Fatal(). This commit changes instances of panic in tests to t.Fatal().


mtardy (commented):
Okay good so the CI is failing as expected now. I still don't understand why "panic" made it not fail on main and v1.1, but fail on v1.0. I have a patch to add the `netcat.openbsd` https://github.com/cilium/little-vm-helper-images/pull/582 on base image which is a reasonable fix that we can integrate into that PR when it's ready.

Is it okay if I push to this PR?

kevsecurity:
> Okay good so the CI is failing as expected now. I still don't understand why "panic" made it not fail on main and v1.1, but fail on v1.0. I have a patch to add the `netcat.openbsd` [cilium/little-vm-helper-images#582](https://github.com/cilium/little-vm-helper-images/pull/582) on base image which is a reasonable fix that we can integrate into that PR when it's ready.
> 
> Is it okay if I push to this PR?

Push away; I made it for you!

mtardy:
Ok thanks.

Related to https://github.com/cilium/tetragon/issues/2686.

olsajiri (approved):
thanks

PR #2688, opened 2024-07-17
tests/e2e: clone proto event in rpcchecker

willfindlay:
We've occasionally been running into errors in e2e tests about invalid wire format messages and message size mismatches in protobuf. According to https://github.com/golang/protobuf/issues/1609, this is generally related to concurrency issues such as modifying a message during marshalling. Add a Clone() call before sending the event over the channel to mitigate this issue.

### DIFF
--- a/pkg/multiplexer/multiplexer.go
+++ b/pkg/multiplexer/multiplexer.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cilium/tetragon/pkg/logger"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/protobuf/proto"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/multiplexer/multiplexer.go
```diff
@@ -156,6 +157,9 @@ func (cm *ClientMultiplexer) GetEvents(ctx context.Context, allowList, denyList
 				default:
 				}
 				res, err := stream.Recv()
+				if err != nil {
+					res = proto.Clone(res).(*tetragon.GetEventsResponse)
+				}
```
mtardy, line 162: one nit: you could add a little comment with the git commit information since it might be easier to understand.

mtardy (approved):

olsajiri (approved):
fixes the issue for me ;-) thanks

pkg/multiplexer/multiplexer.go
willfindlay: I had the same thought this morning haha :sweat_smile: doing it now

willfindlay:
Only added a comment, so no need to wait for this to go through CI again. Merging.

Issue #2679, opened 2024-07-16
Support Postfix operator in matchBinaries selector


anfedotoff:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Is your feature request related to a problem?

Sometimes you know only binary name and this binary can be in different folders on different machines. In operator is limited by elements count.

### Describe the feature you would like

Selector can look like:

```yaml
- matchBinaries:
      - operator: "Postfix"
        values:
        - "qemu-system-x86_64"
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

mtardy:
Thanks for opening the issue, the foundation work can be found here:
- https://github.com/cilium/tetragon/pull/1731
- https://github.com/cilium/tetragon/pull/1732
- https://github.com/cilium/tetragon/pull/1926

As stated in #1731:
> Prepare the way to add the Prefix and NotPrefix operators. And after Postfix and NotPostfix.

You should be able to imitate #1732 for `Postfix` and `NotPostfix`.

As we talked about, I'll assign you to this since you wanted to work on it, please ping me for review when you have something. I can help you along the way if needed since this is something I wanted to eventually do (but I'm pleased you want to take care of it).

PR #2689, opened 2024-07-18
Postfix operator for matchBinaries selector

anfedotoff:
Details can be found in #2679.

- [x] BPF selector part
- [x] Tetragon selector part 
- [x]  Postfix operator test for matchBinaries selector
- [x] Update docs

```release-note
Add the Postfix and NotPostfix operators to the matchBinaries selector.
```

### DIFF
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -7,6 +7,7 @@
 #include "bpf_event.h"
 #include "bpf_helpers.h"
 #include "bpf_cred.h"
+#include "../process/string_maps.h"
 
 /* Applying 'packed' attribute to structs causes clang to write to the
  * members byte-by-byte, as offsets may not be aligned. This is bad for
@@ -274,6 +275,7 @@ struct msg_k8s {
 
 struct heap_exe {
 	char buf[BINARY_PATH_MAX_LEN];
+	char end[STRING_POSTFIX_MAX_LENGTH];
 	__u32 len;
 	__u32 error;
 }; // All fields aligned so no 'packed' attribute.
@@ -311,10 +313,16 @@ typedef __u64 mbset_t;
 struct binary {
 	// length of the path stored in path, this should be < BINARY_PATH_MAX_LEN
 	// but can contain negative value in case of copy error.
-	// While s16 would be sufficient, 64 bits are handy for alignment.
-	__s64 path_length;
+	// While s16 would be sufficient, 32 bits are handy for alignment.
+	__s32 path_length;
+	// if end_r contains reversed path postfix
+	__u32 reversed;
 	// BINARY_PATH_MAX_LEN first bytes of the path
 	char path[BINARY_PATH_MAX_LEN];
+	// STRING_POSTFIX_MAX_LENGTH last bytes of the path
+	char end[STRING_POSTFIX_MAX_LENGTH];
+	// STRING_POSTFIX_MAX_LENGTH reversed last bytes of the path
+	char end_r[STRING_POSTFIX_MAX_LENGTH];
 	// matchBinary bitset for binary
 	mbset_t mb_bitset;
 }; // All fields aligned so no 'packed' attribute
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -179,6 +179,8 @@ read_exe(struct task_struct *task, struct heap_exe *exe)
 {
 	struct file *file = BPF_CORE_READ(task, mm, exe_file);
 	struct path *path = __builtin_preserve_access_index(&file->f_path);
+	__u64 offset = 0;
+	__u64 revlen = STRING_POSTFIX_MAX_LENGTH - 1;
 
 	// we need to walk the complete 4096 len dentry in order to have an accurate
 	// matching on the prefix operators, even if we only keep a subset of that
@@ -188,14 +190,20 @@ read_exe(struct task_struct *task, struct heap_exe *exe)
 	if (!buffer)
 		return 0;
 
+	if (exe->len > STRING_POSTFIX_MAX_LENGTH - 1)
+		offset = exe->len - (STRING_POSTFIX_MAX_LENGTH - 1);
+	else
+		revlen = exe->len;
 	// buffer used by d_path_local can contain up to MAX_BUF_LEN i.e. 4096 we
 	// only keep the first 255 chars for our needs (we sacrifice one char to the
 	// verifier for the > 0 check)
-	if (exe->len > 255)
-		exe->len = 255;
+	if (exe->len > BINARY_PATH_MAX_LEN - 1)
+		exe->len = BINARY_PATH_MAX_LEN - 1;
 	asm volatile("%[len] &= 0xff;\n"
 		     : [len] "+r"(exe->len));
 	probe_read(exe->buf, exe->len, buffer);
+	if (revlen < STRING_POSTFIX_MAX_LENGTH)
+		probe_read(exe->end, revlen, (char *)(buffer + offset));
 
 	return exe->len;
 }
@@ -378,6 +386,11 @@ execve_send(void *ctx)
 			curr->bin.path_length = probe_read(curr->bin.path, event->exe.len, event->exe.buf);
 			if (curr->bin.path_length == 0)
 				curr->bin.path_length = event->exe.len;
+			__u64 revlen = event->exe.len;
+
+			if (event->exe.len > STRING_POSTFIX_MAX_LENGTH - 1)
+				revlen = STRING_POSTFIX_MAX_LENGTH - 1;
+			probe_read(curr->bin.end, revlen, event->exe.end);
 		}
 #else
 		// reuse p->args first string that contains the filename, this can't be
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -827,10 +827,7 @@ filter_char_buf_prefix(struct selector_arg_filter *filter, char *arg_str, uint a
 	return !!pass;
 }
 
-// Define a mask for the maximum path length on Linux.
-#define PATH_MASK (4096 - 1)
-
-FUNC_INLINE void copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
+FUNC_INLINE void __copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset, uint mask)
 {
 	uint i;
 
@@ -849,10 +846,23 @@ FUNC_INLINE void copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
 	// Alternative (prettier) fixes resulted in a confused verifier
 	// unfortunately.
 	for (i = 0; i < (STRING_POSTFIX_MAX_MATCH_LENGTH - 1); i++) {
-		dest[i & STRING_POSTFIX_MAX_MASK] = src[(len + offset - 1 - i) & PATH_MASK];
+		dest[i & STRING_POSTFIX_MAX_MASK] = src[(len + offset - 1 - i) & mask];
 		if (len + offset == (i + 1))
 			return;
 	}
+}
+
+// Define a mask for the maximum path length on Linux.
+#define PATH_MASK (4096 - 1)
+
+FUNC_INLINE void copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
+{
+	__copy_reverse(dest, len, src, offset, PATH_MASK);
+}
+
+FUNC_INLINE void file_copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
+{
+	__copy_reverse(dest, len, src, offset, STRING_POSTFIX_MAX_LENGTH - 1);
 }
 
 FUNC_LOCAL long
@@ -1554,7 +1564,10 @@ FUNC_INLINE int match_binaries(__u32 selidx)
 	__u8 *found_key;
 #ifdef __LARGE_BPF_PROG
 	struct string_prefix_lpm_trie prefix_key;
-	long ret;
+	struct string_postfix_lpm_trie *postfix_key;
+	__u64 postfix_len = STRING_POSTFIX_MAX_MATCH_LENGTH - 1;
+
+	int zero = 0;
 #endif /* __LARGE_BPF_PROG */
 
 	struct match_binaries_sel_opts *selector_options;
@@ -1602,10 +1615,29 @@ FUNC_INLINE int match_binaries(__u32 selidx)
 			// prepare the key on the stack to perform lookup in the LPM_TRIE
 			memset(&prefix_key, 0, sizeof(prefix_key));
 			prefix_key.prefixlen = current->bin.path_length * 8; // prefixlen is in bits
-			ret = probe_read(prefix_key.data, current->bin.path_length & (STRING_PREFIX_MAX_LENGTH - 1), current->bin.path);
-			if (ret < 0)
+			if (probe_read(prefix_key.data, current->bin.path_length & (STRING_PREFIX_MAX_LENGTH - 1), current->bin.path) < 0)
 				return 0;
 			found_key = map_lookup_elem(path_map, &prefix_key);
+			break;
+		case op_filter_str_postfix:
+		case op_filter_str_notpostfix:
+			path_map = map_lookup_elem(&string_postfix_maps, &selector_options->map_id);
+			if (!path_map)
+				return 0;
+			if (current->bin.path_length < STRING_POSTFIX_MAX_MATCH_LENGTH)
+				postfix_len = current->bin.path_length;
+			postfix_key = (struct string_postfix_lpm_trie *)map_lookup_elem(&string_postfix_maps_heap, &zero);
+			if (!postfix_key)
+				return 0;
+			postfix_key->prefixlen = postfix_len * 8; // prefixlen is in bits
+			if (!current->bin.reversed) {
+				file_copy_reverse((__u8 *)current->bin.end_r, postfix_len, (__u8 *)current->bin.end, current->bin.path_length - postfix_len);
+				current->bin.reversed = true;
+			}
+			if (postfix_len < STRING_POSTFIX_MAX_MATCH_LENGTH)
+				if (probe_read(postfix_key->data, postfix_len, current->bin.end_r) < 0)
+					return 0;
+			found_key = map_lookup_elem(path_map, postfix_key);
 			break;
 #endif /* __LARGE_BPF_PROG */
 		default:
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -158,9 +158,17 @@ calls and kernel functions that are coming from `cat` or `tail`.
     - "/usr/bin/tail"
 ```
 
-Currently, only the `In` operator type is supported and the `values` field has
-to be a map of `strings`. The default behaviour is `followForks: true`, so all
-the child processes are followed. The current limitation is 4 values.
+The available operators for `matchBinaries` are:
+- `In`
+- `NotIn`
+- `Prefix`
+- `NotPrefix`
+- `Postfix`
+- `NotPostfix`
+
+The `values` field has to be a map of `strings`. The default behaviour
+is `followForks: true`, so all the child processes are followed.
+The current limitation is 4 values.
 
 **Further examples**
 
--- a/pkg/api/processapi/processapi.go
+++ b/pkg/api/processapi/processapi.go
@@ -45,6 +45,8 @@ const (
 	MSG_COMMON_FLAG_USER_STACKTRACE   = 0x4
 
 	BINARY_PATH_MAX_LEN = 256
+
+	STRING_POSTFIX_MAX_LENGTH = 128
 )
 
 type MsgExec struct {
@@ -139,8 +141,11 @@ type MsgCapabilities struct {
 }
 
 type Binary struct {
-	PathLength int64
+	PathLength int32
+	Reversed   uint32
 	Path       [BINARY_PATH_MAX_LEN]byte
+	End        [STRING_POSTFIX_MAX_LENGTH]byte
+	End_r      [STRING_POSTFIX_MAX_LENGTH]byte
 	MBSet      uint64
 }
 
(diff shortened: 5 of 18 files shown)


bpf/process/types/basic.h
```diff
@@ -848,7 +845,7 @@ FUNC_INLINE void copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
 	// Alternative (prettier) fixes resulted in a confused verifier
 	// unfortunately.
 	for (i = 0; i < (STRING_POSTFIX_MAX_MATCH_LENGTH - 1); i++) {
-		dest[i & STRING_POSTFIX_MAX_MASK] = src[(len + offset - 1 - i) & PATH_MASK];
+		dest[i & STRING_POSTFIX_MAX_MASK] = src[(len + offset - 1 - i) & (BINARY_PATH_MAX_LEN - 1)];
```
mtardy, line 848: So this is more restrictive so it should be fine, and postfix on strings is also limited IIRC, would be nice to have @kevsecurity ack on that. BINARY_PATH_MAX_LEN is 256 which is < 128 (see below).

```C
#define STRING_POSTFIX_MAX_LENGTH 128
#define STRING_POSTFIX_MAX_MASK	  (STRING_POSTFIX_MAX_LENGTH - 1)
#ifdef __LARGE_BPF_PROG
#define STRING_POSTFIX_MAX_MATCH_LENGTH STRING_POSTFIX_MAX_LENGTH
#else
#define STRING_POSTFIX_MAX_MATCH_LENGTH 95
#endif
```

bpf/process/types/basic.h
```diff
@@ -1599,6 +1599,21 @@ FUNC_INLINE int match_binaries(__u32 selidx)
...
 			found_key = map_lookup_elem(path_map, &prefix_key);
 			break;
+		case op_filter_str_postfix:
+		case op_filter_str_notpostfix:
+			postfix_len = current->bin.path_length;
+			path_map = map_lookup_elem(&string_postfix_maps, &selector_options->map_id);
+			if (!path_map || !postfix_len)
+				return 0;
+			if (postfix_len >= STRING_POSTFIX_MAX_MATCH_LENGTH)
+				postfix_len = STRING_POSTFIX_MAX_MATCH_LENGTH - 1;
```
mtardy, line 1609: note for self: make sure we have a check userspace side that rejects those. Cropping like that could be very confusing.

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -3884,6 +3884,18 @@ func TestKprobeMatchBinaries(t *testing.T) {
 		}
 		matchBinariesTest(t, "NotPrefix", []string{"/usr/bin/t"}, createBinariesChecker("/usr/bin/head", "/etc/passwd"))
 	})
+	t.Run("Postfix", func(t *testing.T) {
+		if !kernels.EnableLargeProgs() {
+			t.Skip(skipMatchBinariesPrefix)
```
mtardy, line 3889: you can update the `skipMatchBinariesPrefix` const string to include Postfix/NotPostfix

pkg/sensors/tracing/kprobe_test.go
mtardy, line 1: Thanks that's cool you added tests, here you check:
- That the `Postfix` works when you specify a path and trigger a binary with that path
- That the `NotPostfix` works when you specify a path and trigger a binary without that path

You could use the `TestKprobeMatchBinariesPerfring`
- That the `NotPostfix` works when you specify a path and trigger a binary without that path **and correctly filters the event that could have been triggered from the binary from the specified path**. i.e. check for missing events (because NotPrefix filtered them).
- You can do the same with `PostFix` and check that the event outside bin/tail is not present and was filtered.

See `TestKprobeMatchBinariesPerfring` and `matchBinariesPerfringTest`, this is a bit more advanced and guarantees the thing behaves as thought.

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -158,9 +158,17 @@ calls and kernel functions that are coming from `cat` or `tail`.
...
+- `In`
+- `NotIn`
+- `Prefix`
+- `NotPrefix`
+- `Postfix`
+- `NotPostfix`
+
+The `values` field has to be a map of `strings`. The default behaviour
+is `followForks: true`, so all the child processes are followed.
+The current limitation is 4 values.
```
mtardy, line 171: I think the limitation here is outdated but it's out of the scope of this PR

mtardy (approved):
Thanks that's a great addition! Would love an ack from Kev since we are touching a few bits of its STRING code.

My main nit would be to add a test using the perfring framework to check for a missing event so that we are sure this thing works as intended, and I'm good! :)

bpf/process/types/basic.h
kevsecurity: I'd rather leave this at PATH_MASK if possible as it better represents the length of a path on Linux. Currently we appear to limit paths (when copying) to 256 bytes so this change would almost certainly work. But if we later increase path lengths to 4096, then this will also need changing.

kevsecurity: Okay, I can see that this doesn't work for the execve_map_value->bin.path because it is only BINARY_PATH_MAX_LENGTH in size. But string/charbuf buffers could be 4096 in size, and postfix matching needs to work on these too. I suggest a file_copy_reverse function that is used only in file/binary copying/matching, and maybe make binary postfix matching a LARGE_BPF_PROG option only to avoid hitting complexity limits on older kernels.

bpf/process/types/basic.h
```diff
@@ -1599,6 +1599,21 @@ FUNC_INLINE int match_binaries(__u32 selidx)
...
+			postfix_key = (struct string_postfix_lpm_trie *)map_lookup_elem(&string_postfix_maps_heap, &zero);
+			if (!postfix_key)
+				return 0;
+			postfix_key->prefixlen = postfix_len * 8; // prefixlen is in bits
+			copy_reverse(postfix_key->data, postfix_len, (__u8 *)current->bin.path, current->bin.path_length - postfix_len);
```
kevsecurity, line 1614: See earlier comment. I don't think current->bin.path is suitable for postfix matching. We need to store the end of the path string in a second (new) buffer in bin.

kevsecurity (changes requested):
I think we need to store the last 128 chars of the binary path in execve_send (bpf/process/bpf_execve_event.c) in a new part of the struct binary, specifically for postfix matching.
I think we need a copy of copy_reverse() (called file_copy_reverse()) that only copies file paths from buffer expected to be BINARY_PATH_MAX_LENGTH in size, leaving the original with the PATH_MASK position masking.

bpf/process/types/basic.h
anfedotoff: The reason I changed `PATH_MASK` is because verifier check failed. AFAIK,  `copy_reverse` is used only in one place (except new added call):

https://github.com/cilium/tetragon/blob/95f263729f129bb1b2d02ae157b9b7d24804832c/bpf/process/types/basic.h#L878

The verification in this case is OK, because it is enough place to hold the source buffer:

https://github.com/cilium/tetragon/blob/95f263729f129bb1b2d02ae157b9b7d24804832c/bpf/lib/generic.h#L50

I understand correctly, I suggest to create a function  `file_copy_reverse`, that like `copy_reverse`, but with (BINARY_PATH_MAX_LEN - 1) mask? If so, maybe rename `copy_reverse` to `arg_copy_reverse` than?

anfedotoff: I found the answer [here](https://github.com/cilium/tetragon/pull/2689#pullrequestreview-2191352985), thanks:).

kevsecurity: > AFAIK, `copy_reverse` is used only in one place (except new added call):
> 
> https://github.com/cilium/tetragon/blob/95f263729f129bb1b2d02ae157b9b7d24804832c/bpf/process/types/basic.h#L878

Except that handles character buffers up to 4096 chars in length.

> The verification in this case is OK, because it is enough place to hold the source buffer:
> 
> https://github.com/cilium/tetragon/blob/95f263729f129bb1b2d02ae157b9b7d24804832c/bpf/lib/generic.h#L50
> 
> If I understand correctly, you suggest to create a function `file_copy_reverse`, that like `copy_reverse`, but with (BINARY_PATH_MAX_LEN - 1) mask? If so, maybe rename `copy_reverse` to `arg_copy_reverse` than?

Yep. Then char buffers can be much longer and still do postfix matching correctly, without affecting the masking needed for binary postfix matching.

mtardy:
> I think we need to store the last 128 chars of the binary path in execve_send (bpf/process/bpf_execve_event.c) in a new part of the struct binary, specifically for postfix matching.
> 
> I think we need a copy of copy_reverse() (called file_copy_reverse()) that only copies file paths from buffer expected to be BINARY_PATH_MAX_LENGTH in size, leaving the original with the PATH_MASK position masking.

Thanks kev I completely overlooked this previous struggle. Indeed, for the binary we don't store the whole path so you have a few cases to take into account:
- path is short enough you can do what you did
- path is too long and you need to store at exec time the end of it (at least 128 chars). You can lazily reverse it when needed.

kevsecurity:
> Thanks kev I completely overlooked this previous struggle. Indeed, for the binary we don't store the whole path so you have a few cases to take into account:
> 
> * path is short enough you can do what you did
> * path is too long and you need to store at exec time the end of it (at least 128 chars). You can lazily reverse it when needed.

Personally, I'd always copy the end into the new end slot to save extra logic when matching. I agree that you should lazily reverse it the first time you do a postfix match on it, but this has race conditions. I think the simple solution (not necessarily the cheapest or best) is to have two extra slots, not one extra slot! So struct binary could look like:
```
struct binary {
	// length of the path stored in path, this should be < BINARY_PATH_MAX_LEN
	// but can contain negative value in case of copy error.
	// While s16 would be sufficient, 32 bits are handy for alignment.
	__s32 path_length;
	__u8 reversed;
	__u8 pad[3];
	// BINARY_PATH_MAX_LEN first bytes of the path
	char path[BINARY_PATH_MAX_LEN];
	char end[BINARY_PATH_MAX_LEN];
	char end_r[BINARY_PATH_MAX_LEN];
}; // All fields aligned so no 'packed' attribute
```

On execution, copy the end of the path into end. On first postfix lookup, reverse end into end_r and set reversed to true. If another thread is doing the same then end will be unaffected, so end_r will end up with the same values, even if one thread is still reversing end into end_r while another is using end_r. If reversed is true, then just match on end_r and assume it has already been reversed.

mtardy:
I just remember I had a very early draft branch from October but I don't think it'll really help you here, it was similar to the advice given here.

anfedotoff:
@mtardy , @kevsecurity , guys, can I ask you to review again, please?

mtardy:
> @mtardy , @kevsecurity , guys, can I ask you to review again, please?

sure, note that you can use that btw even if a message is fine of course.
image

bpf/lib/process.h
```diff
@@ -313,8 +314,15 @@ struct binary {
 	// but can contain negative value in case of copy error.
 	// While s16 would be sufficient, 64 bits are handy for alignment.
 	__s64 path_length;
+	// if end_r contains reversed path postfix
+	__u8 reversed;
+	__u8 pad[7];
```
mtardy, line 319: If you want you can also make `path_length` a u32 and reverse a u32 so that we are still aligned.

bpf/process/bpf_execve_event.c
```diff
@@ -191,11 +192,16 @@ read_exe(struct task_struct *task, struct heap_exe *exe)
 	// buffer used by d_path_local can contain up to MAX_BUF_LEN i.e. 4096 we
 	// only keep the first 255 chars for our needs (we sacrifice one char to the
 	// verifier for the > 0 check)
-	if (exe->len > 255)
+	if (exe->len > 255) {
+		offset = exe->len - 255;
 		exe->len = 255;
```
mtardy, line 197: a note not on your PR but general improvement, we could replace that with `BINARY_PATH_MAX_LEN - 1` here.

bpf/process/types/basic.h
```diff
@@ -1636,7 +1636,16 @@ FUNC_INLINE int match_binaries(__u32 selidx)
 			if (!postfix_key)
 				return 0;
 			postfix_key->prefixlen = postfix_len * 8; // prefixlen is in bits
-			file_copy_reverse(postfix_key->data, postfix_len, (__u8 *)current->bin.path, current->bin.path_length - postfix_len);
+			if (!current->bin.reversed) {
+				file_copy_reverse((__u8 *)current->bin.end_r, postfix_len, (__u8 *)current->bin.path, current->bin.path_length - postfix_len);
```
mtardy, line 1640: Wasn't the goal of keeping `end` to reverse `bin.end` and not `bin.path`? I think adding a test with a long binary like in 67348f6f826cc68c40625dbdeb8cc90dec5bc455 would have caught that.

bpf/process/types/basic.h
```diff
@@ -854,6 +854,21 @@ FUNC_INLINE void copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
 	}
 }
 
+#ifdef __LARGE_BPF_PROG
+FUNC_INLINE void file_copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
+{
+	uint i;
+
+	len &= STRING_POSTFIX_MAX_MASK;
+#pragma unroll
```
mtardy, line 863: Since you have a `__LARGE_BPF_PROG` macro around this, you should be able to just drop the pragma unroll and use bounded loops here.

bpf/process/types/basic.h
```diff
@@ -854,6 +854,21 @@ FUNC_INLINE void copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
 	}
 }
 
+#ifdef __LARGE_BPF_PROG
+FUNC_INLINE void file_copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
```
mtardy, line 858: we could maybe add a comment on the fact that it's almost a copy of:
https://github.com/cilium/tetragon/blob/c638b45a5e1e9467b4505ec3adb9dbf4a26823fa/bpf/process/types/basic.h#L833
but with a different PATH_MASK and why it's needed so that we don't get confused later on.

bpf/process/types/basic.h
```diff
@@ -1606,6 +1624,21 @@ FUNC_INLINE int match_binaries(__u32 selidx)
...
+			path_map = map_lookup_elem(&string_postfix_maps, &selector_options->map_id);
+			if (!path_map || !postfix_len)
+				return 0;
+			if (postfix_len >= STRING_POSTFIX_MAX_MATCH_LENGTH)
+				postfix_len = STRING_POSTFIX_MAX_MATCH_LENGTH - 1;
+			postfix_key = (struct string_postfix_lpm_trie *)map_lookup_elem(&string_postfix_maps_heap, &zero);
+			if (!postfix_key)
+				return 0;
+			postfix_key->prefixlen = postfix_len * 8; // prefixlen is in bits
+			file_copy_reverse(postfix_key->data, postfix_len, (__u8 *)current->bin.path, current->bin.path_length - postfix_len);
```
mtardy: maybe you want to merge that commit with the next one since it's maybe not nice to have this line that is "wrong" since it's just reversing `bin.path`.

mtardy (commented):
Thanks a lot for those updates :), here are some comments. My main point would be to add a test with a long binary like in https://github.com/cilium/tetragon/commit/67348f6f826cc68c40625dbdeb8cc90dec5bc455 so that we make sure this PR works and doesn't break.

bpf/process/types/basic.h
kevsecurity: Thinking about this (and accepting that I said we could do with a new function), as it's FUNC_INLINE we could probably get away with passing the mask as a parameter. So both copy_reverse and file_copy_reverse could just be calls to another FUNC_INLINE _copy_reverse which has the relevant mask as a parameter. WDYT?

bpf/process/types/basic.h
anfedotoff: Oh, thanks, my fault).

kevsecurity:
It's looking good. Address the few things including some tests and we should be good to merge.

bpf/lib/process.h
```diff
@@ -274,6 +274,7 @@ struct msg_k8s {
 
 struct heap_exe {
 	char buf[BINARY_PATH_MAX_LEN];
+	char end[BINARY_PATH_MAX_LEN];
```
kevsecurity, line 277: Sorry just noticed that the postfix LPM TRIE maps only go up to STRING_POSTFIX_MAX_LENGTH in size (= 128).
So here should probably be:
```
```

bpf/lib/process.h
```diff
@@ -311,10 +312,16 @@ typedef __u64 mbset_t;
...
+	__s32 path_length;
+	// if end_r contains reversed path postfix
+	__u32 reversed;
 	// BINARY_PATH_MAX_LEN first bytes of the path
 	char path[BINARY_PATH_MAX_LEN];
+	// BINARY_PATH_MAX_LEN last bytes of the path
+	char end[BINARY_PATH_MAX_LEN];
+	// BINARY_PATH_MAX_LEN reversed last bytes of the path
+	char end_r[BINARY_PATH_MAX_LEN];
 	// matchBinary bitset for binary
```
kevsecurity, line 325: Sorry just noticed that the postfix LPM TRIE maps only go up to STRING_POSTFIX_MAX_LENGTH in size (= 128).
So here should probably be:
```
```

bpf/process/bpf_execve_event.c
```diff
@@ -191,11 +192,16 @@ read_exe(struct task_struct *task, struct heap_exe *exe)
 	// buffer used by d_path_local can contain up to MAX_BUF_LEN i.e. 4096 we
 	// only keep the first 255 chars for our needs (we sacrifice one char to the
 	// verifier for the > 0 check)
-	if (exe->len > 255)
-		exe->len = 255;
+	if (exe->len > BINARY_PATH_MAX_LEN - 1) {
+		offset = exe->len - (BINARY_PATH_MAX_LEN - 1);
+		exe->len = BINARY_PATH_MAX_LEN - 1;
+	}
```
kevsecurity, line 198: Following on with the length of postfix, I think we need to move the setting of offset into a following check (i.e. one if block for normal exe->len setting, and another new if block for offset and revlen setting):
```
__u64 revlen = exe->len;
if (exe->len > STRING_POSTFIX_MAX_LENGTH - 1) {
    offset = exe->len - (STRING_POSTFIX_MAX_LENGTH - 1);
    revlen = STRING_POSTFIX_MAX_LENGTH - 1;
}

File: bpf/process/bpf_execve_event.c [hunk 15]
```diff
@@ -191,11 +192,16 @@ read_exe(struct task_struct *task, struct heap_exe *exe)
...
 	asm volatile("%[len] &= 0xff;\n"
 		     : [len] "+r"(exe->len));
 	probe_read(exe->buf, exe->len, buffer);
+	asm volatile("%[len] &= 0xff;\n"
+		     : [len] "+r"(exe->len));
+	probe_read(exe->end, exe->len, (char *)(buffer + offset));
```
kevsecurity, line 204: And then here use revlen instead of exe->len.

bpf/process/bpf_execve_event.c
```diff
@@ -378,6 +384,7 @@ execve_send(void *ctx)
 			curr->bin.path_length = probe_read(curr->bin.path, event->exe.len, event->exe.buf);
 			if (curr->bin.path_length == 0)
 				curr->bin.path_length = event->exe.len;
+			probe_read(curr->bin.end, event->exe.len & (BINARY_PATH_MAX_LEN - 1), event->exe.end);
```
kevsecurity, line 387: Swap BINARY_PATH_MAX_LEN for STRING_POSTFIX_MAX_LENGTH

bpf/process/types/basic.h
```diff
@@ -849,12 +846,25 @@ FUNC_INLINE void copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
...
+#define PATH_MASK (4096 - 1)
+
+FUNC_INLINE void copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
+{
+	__copy_reverse(dest, len, src, offset, PATH_MASK);
+}
+
+FUNC_INLINE void file_copy_reverse(__u8 *dest, uint len, __u8 *src, uint offset)
+{
+	__copy_reverse(dest, len, src, offset, BINARY_PATH_MAX_LEN - 1);
```
kevsecurity, line 865: Swap BINARY_PATH_MAX_LEN for STRING_POSTFIX_MAX_LEN

pkg/api/processapi/processapi.go
```diff
@@ -139,8 +139,11 @@ type MsgCapabilities struct {
 }
 
 type Binary struct {
-	PathLength int64
+	PathLength int32
+	Reversed   uint32
 	Path       [BINARY_PATH_MAX_LEN]byte
+	End        [BINARY_PATH_MAX_LEN]byte
+	End_r      [BINARY_PATH_MAX_LEN]byte
```
kevsecurity, line 146: For End and End_r swap BINARY_PATH_MAX_LEN for selectors.StringPostfixMaxLength or a new const = 128

bpf/process/bpf_execve_event.c
kevsecurity: Could maybe move the revlen declaration (I just put it there in the example but probably needs to go up the top after offset) to avoid the static check errors.

bpf/lib/process.h
anfedotoff: Yes, thank you, that's a good point. I'm fighting with verifier now).

kevsecurity: Ah sorry. Feel free to paste problem here; we might be able to help.

anfedotoff: Thanks, I'm think I managed to deal with verifier:)

kevsecurity (approved):
LGTM

mtardy (approved):
Thanks, it's almost ready, the only small blocker for me is to have a first commit that compiles & runs (the struct should be synchronized between kernel/user), the rest is nits and followups.

anfedotoff:
@mtardy , I think now, it's ready:)

mtardy:
> @mtardy , I think now it's ready:)

Yep thanks again, let's merge this! 🌮 🎉

PR #2692, opened 2024-07-18
pkg/sensors: reduce memory footprint of unused `override_tasks` maps

mtardy:
Some of the plumbing was already present. We now pin the map correctly.

```release-note
Reduce the kernel memory footprint (accounted by the cgroup v2 memory controller) of the override feature when unused (around ~3MB per kprobe).
```


mtardy:
I could update the test added here https://github.com/cilium/tetragon/commit/d63d82889e133b4f7656355b6d37f407cd20faa0.

olsajiri (approved):

PR #2695, opened 2024-07-18
tetragon: clone namespace improvements

tixxdz:
```release-note
store thread leader namespaces at fork and reduce false positives
```

### DIFF
--- a/bpf/process/bpf_fork.c
+++ b/bpf/process/bpf_fork.c
@@ -25,7 +25,6 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
 {
 	struct execve_map_value *curr, *parent;
 	struct msg_clone_event msg;
-	struct msg_capabilities caps;
 	u64 msg_size = sizeof(struct msg_clone_event);
 	struct msg_k8s kube;
 	u32 tgid = 0;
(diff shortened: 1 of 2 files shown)


tixxdz:
CI failure seems unrelated

bpf/process/bpf_fork.c
```diff
@@ -65,10 +64,12 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
 	 * before the execve hook point if they changed or not.
 	 * This needs to be converted later to credentials.
 	 */
-	get_current_subj_caps(&caps, task);
-	curr->caps.permitted = caps.permitted;
-	curr->caps.effective = caps.effective;
-	curr->caps.inheritable = caps.inheritable;
```
tpapagian, line 71: Unrelated to that PR, but these assignments seems to be redundant as we did `get_current_subj_caps`? And this is the reason that you removed those now, right?

tpapagian (approved):

bpf/process/bpf_fork.c
tixxdz: This just minor optimization, since old code was reading in new var caps then copying back to curr->caps.$field, see next line in the patch, now we read directly into curr->caps , so just saved some cycles.

tixxdz:
Thank you @tpapagian

PR #2703, opened 2024-07-20
tetra: LSM events compact print support

anfedotoff:
Adding LSM events compact printing like it is done for kprobes in default case.

It seems that I forgot about `tetra` compact  printing for LSM events.  Maybe we can change ❓ symbol to something else to distinguish LSM hooks and kprobes.

### DIFF
--- a/pkg/encoder/encoder.go
+++ b/pkg/encoder/encoder.go
@@ -491,6 +491,14 @@ func (p *CompactEncoder) EventToString(response *tetragon.GetEventsResponse) (st
 			event := p.Colorer.Blue.Sprintf("⁉️ %-7s", "tracepoint")
 			return CapTrailorPrinter(fmt.Sprintf("%s %s %s %s", event, processInfo, tp.Subsys, tp.Event), caps), nil
 		}
+	case *tetragon.GetEventsResponse_ProcessLsm:
+		lsm := response.GetProcessLsm()
+		if lsm.Process == nil {
+			return "", ErrMissingProcessInfo
+		}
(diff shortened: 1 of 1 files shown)


anfedotoff:
Hi @kkourt!  Please could you have a look?

pkg/encoder/encoder.go
```diff
@@ -491,6 +491,14 @@ func (p *CompactEncoder) EventToString(response *tetragon.GetEventsResponse) (st
 			event := p.Colorer.Blue.Sprintf("⁉️ %-7s", "tracepoint")
 			return CapTrailorPrinter(fmt.Sprintf("%s %s %s %s", event, processInfo, tp.Subsys, tp.Event), caps), nil
 		}
+	case *tetragon.GetEventsResponse_ProcessLsm:
+		lsm := response.GetProcessLsm()
+		if lsm.Process == nil {
+			return "", ErrMissingProcessInfo
+		}
+		processInfo, caps := p.Colorer.ProcessInfo(response.NodeName, lsm.Process)
+		event := p.Colorer.Blue.Sprintf("❓ %-7s", "LSM hook")
```
mtardy, line 500: This is just a proposition, since we are into choosing emoji ahah:

I would drop the "hook" part for brevity, maybe you can use the lock emoji for LSM? Not sure but this is what pops when you type "security" in the emojipedia https://emojipedia.org/search?q=security. I would avoid the police officer. 
```suggestion
		event := p.Colorer.Blue.Sprintf("🔒 %-7s", "LSM")
```

mtardy (approved):
Looks good, but I haven't tried it live! Thanks!

kkourt (approved):
LGTM, thanks!

PR #2705, opened 2024-07-22
tetragon: make resolving uid to username work with a processapi struct

tixxdz:
### Description
tetragon: make resolving uid to username work with a processapi struct


kkourt (approved):
lgtm, thanks.

PR #2709, opened 2024-07-22
Prevent search engines from indexing deploy preview URLs and force redirect already indexed URLs to the homepage.

paularah:
See Slack conversation reference from Scott Lowe 

This patch prevents the deploy previews from being indexed in search engines and force-redirects existing indexed deploy preview pages to the main Tetragon site.

1. updates the netlify config to include redirect rule

2. create a robots.txt file the prevents indexing deploy preview pages

### Description
Prevent search engines from indexing deploy preview URLs and force redirect already indexed URLs to the homepage.

### DIFF
--- a/docs/static/robots.txt
+++ b/docs/static/robots.txt
@@ -1,0 +1,2 @@
+User-agent: *
+Disallow: /deploy-preview-*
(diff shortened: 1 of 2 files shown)


mtardy:
> See Slack conversation reference from Scott Lowe

How can I see that?

mtardy (commented):
I don't think your patch works? Have you tried looking at the preview and its robots.txt?

mtardy:
This should do what you want https://github.com/cilium/tetragon/pull/2711.

mtardy:
Should be fixed by https://github.com/cilium/tetragon/pull/2711

Issue #2704, opened 2024-07-21
Development setup


spbkaizo:
I ran into an issue on building the docker images.

```
<simonb@odin:~/src/tetragon>
zsh/2 848  (git)-[main]-% make image
docker build -t "cilium/tetragon:latest" --target release --build-arg TETRAGON_VERSION=659704f9 --platform=linux/amd64 .
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/

Sending build context to Docker daemon  191.9MB
Step 1/39 : FROM --platform=$BUILDPLATFORM quay.io/cilium/clang:aeaada5cf60efe8d0e772d032fe3cc2bc613739c@sha256:b440ae7b3591a80ffef8120b2ac99e802bbd31dee10f5f15a48566832ae0866f AS bpf-builder
failed to parse platform : "" is an invalid component of "": platform specifier component must match "^[A-Za-z0-9_-]+$": invalid argument
make: *** [Makefile:172: image] Error 1
```

I've got a bit of a handcranked linux install, with docker-ce installed and running from a standalone binary.  The solution to get it to build was to install the buildx cli plugin.  I'm guessing this may come as part of the full docker desktop installations.

```
Download buildx from https://github.com/docker/buildx/releases
mv the buildx binary to ~/.docker/cli-plugins/docker-buildx
Check install with:
docker buildx version
```

This then allows the build to continue, as it now utilises the buildx cli plugin.

This might be helpful for others in case they encounter the same.

kkourt:
Thanks @spbkaizo!
Feel free to add this to https://tetragon.io/docs/contribution-guide/development-setup/#requirements (https://github.com/cilium/tetragon/blob/main/docs/content/en/docs/contribution-guide/development-setup.md)

PR #2713, opened 2024-07-22
doc: Add docker-buildx to requirements

spbkaizo:
The Docker buildx plugin is required as part of make image - it may or not be installed by a users docker installation.

This commit adds it to the requirements.

Fixes https://github.com/cilium/tetragon/issues/2704


mtardy (approved):
Thanks a lot!

PR #2714, opened 2024-07-22
Define health metrics group

lambdanis:
There are two things happening here:

1. Refactored custom collectors to use new helpers from pkg/metrics (commits 1-3). The main win here is simplifying how custom metrics are included in reference docs - we don't need to define a separate type and register it separately just for generating docs.
2. Defined a health metrics group (last commit). The idea is that in the future this group will have constrained cardinality and will be enabled by default (in contrast to another group with potentially high cardinality "debug" metrics). This commit only refactors the existing metrics initialization code to use the new framework. The health metrics group contains all metrics that were documented in the "health metrics" section, but in the future some of them will likely be moved to another group.

I had this PR already: #2604 stacked on top of #2606, but then I merged the latter, the base branch got deleted, the PR got closed and I can't reopen it, so opening it again.

### DIFF
--- a/cmd/tetragon-metrics-docs/main.go
+++ b/cmd/tetragon-metrics-docs/main.go
@@ -27,7 +27,7 @@ func main() {
 func initMetrics(target string, reg *prometheus.Registry, _ *slog.Logger) error {
 	switch target {
 	case "health":
-		metricsconfig.InitHealthMetricsForDocs(reg)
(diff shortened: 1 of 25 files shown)


kevsecurity (approved):
LGTM. Maybe worth getting a review from someone (Chance?) who knows more about this area than me though.

PR #2718, opened 2024-07-23
bpf: Fix Prefix operator for matchBinaries

anfedotoff:
It seems to me, that Prefix operator for matchBinaries selector is not applied if executable path length is greater than 256 bytes.


mtardy:
> I discovered this issue while working on  #2689. @mtardy , @kevsecurity , please, have a look.

Hey I haven't looked in details but a good way to present this would be:

1. A failing test
2. Fix the failure to show the test now succeed

anfedotoff:
> > I discovered this issue while working on  #2689. @mtardy , @kevsecurity , please, have a look.
> 
> Hey I haven't looked in details but a good way to present this would be:
> 
> 1. A failing test
> 2. Fix the failure to show the test now succeed

I have Tetragon running and it is armed with this policy:

```yaml
kind: TracingPolicy
metadata:
  name: "lsm"
spec:
  lsmhooks:
  - hook: "file_open"
    args:
      - index: 0
        type: "file"
    selectors:
    - matchBinaries:
      - operator: "Prefix"
        values:
        - "/home/anfedotoff/go/src/github.com/cilium/tetragon/contrib/tester-progs"
 ```

if I run this binary, the LSM is not triggered, but it seems to me it does.

```
/home/anfedotoff/go/src/github.com/cilium/tetragon/contrib/tester-progs/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa/bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb/ccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc/ddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd/direct-write-tester /tmp/test-test
```

**Comment by anfedotoff** [comment]
@mtardy, Hey! I added a test. It is failed on some VMs. I think, it because it  isn't allowed to run binaries from `/tmp`

```
kprobe_test.go:3923: failed to run direct-write-tester: exit status 255
```
But overall it shows the problem, I think. Maybe you have an idea where to create a large path to test binary?

**Comment by mtardy** [comment]
> @mtardy, Hey! I added a test. It is failed on some VMs. I think, it because it isn't allowed to run binaries from `/tmp`
> 
> ```
> kprobe_test.go:3923: failed to run direct-write-tester: exit status 255
> ```
> 
> But overall it shows the problem, I think. Maybe you have an idea where to create a large path to test binary?

yes, do you remember this https://github.com/cilium/tetragon/pull/2698?

https://github.com/cilium/tetragon/blob/b506aa1260c759cbd9a447034a4a7e0d50490469/pkg/sensors/tracing/enforcer_test.go#L300-L304

**Comment by mtardy** [comment]
> Maybe you have an idea where to create a large path to test binary?

I once created this thing that could be adapted or copy-pasted in our test base, as the comment below, you could use `/var/tmp` or just stop using direct write tester.

**Comment by anfedotoff** [comment]
> > Maybe you have an idea where to create a large path to test binary?
> 
> I once created this thing https://github.com/mtardy/pathgen that could be adapted or copy-pasted in our test base, as the comment below, you could use `/var/tmp` or just stop using direct write tester.
> 
> If you can avoid using the direct write tester I would as it uses a not very user-friendly Linux interface (breaks on tmpfs).

Ahh, thanks! I think I'll try to look at some other executable from `contrib/tester-progs`.

File: bpf/process/bpf_execve_event.c [hunk 1]
```diff
@@ -329,19 +330,29 @@ execve_send(void *ctx)
 		memset(&curr->bin, 0, sizeof(curr->bin));
 #ifdef __LARGE_BPF_PROG
 		// read from proc exe stored at execve time
-		if (event->exe.len <= BINARY_PATH_MAX_LEN) {
+		if (event->exe.len <= BINARY_PATH_MAX_LEN && !event->exe.error) {
 			curr->bin.path_length = probe_read(curr->bin.path, event->exe.len, event->exe.off);
 			if (curr->bin.path_length == 0)
 				curr->bin.path_length = event->exe.len;
+		} else {
+			heap = map_lookup_elem(&execve_heap, &zero);
```
mtardy, line 338: so I'm not sure it's really appropriate as a fix:
- The current implementation is indeed limited to 256 bytes on `exe` read, but this could be increased (maybe that's what you want to actually do). This is indeed bad that it's failing open for something as Prefix.
- Here you read `execve_heap` that contains the binary only because of the way we read args at `tracepoint/sys_execve`, I don't think there are any guarantees that this heap will actually contain this btw. We skip the binary name that is the first args, but this rollbacks to the old situation where we retrieved the filename from that args which could lead to read a relative path or a symlink (we switched to exe for that reason, reading the syscall args wasn't reliable). See: https://github.com/cilium/tetragon/blob/92a17c2386780497a5541dae5d0b0824f036dc0c/bpf/process/bpf_execve_event.c#L248-L254

I agree that there is a limitation on the current implementation which is pretty explicit (with `BINARY_PATH_MAX_LEN`), but falling back to the way we do things on <5.4 kernels will not help as it's also flawed and can be escaped with relative paths symlinks.

And since you `probe_read_str` using `BINARY_PATH_MAX_LEN` that is set to 256, I'm not sure to understand how that helps?

We have unit tests on the function that uses `read_exe` and the error should be filled if the path is too big to enter the buffer of 256 (it seems that the limit is 255) so we should be aware that we could not match properly.

mtardy (changes requested):
could you explain more your patch?

Thanks for the test, it seems to highlight the issue it's a nice start

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -3886,6 +3886,50 @@ func TestKprobeMatchBinaries(t *testing.T) {
...
 }
 
+func TestKprobeMatchBinariesPrefixLargePath(t *testing.T) {
+	if !kernels.EnableLargeProgs() {
+		t.Skip()
+	}
+	repoBinPath := testutils.RepoRootPath("contrib/tester-progs/nop")
+	tmpDir, err := os.MkdirTemp("/tmp", "tetragon-matchbinlarge-*")
+	assert.NoError(t, err)
+	defer os.RemoveAll(tmpDir)
```
mtardy, line 3896: use `t.Cleanup` instead of `defer` for this kind of thing in tests.

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -3886,6 +3886,50 @@ func TestKprobeMatchBinaries(t *testing.T) {
 	})
 }
 
+func TestKprobeMatchBinariesPrefixLargePath(t *testing.T) {
+	if !kernels.EnableLargeProgs() {
+		t.Skip()
+	}
+	repoBinPath := testutils.RepoRootPath("contrib/tester-progs/nop")
+	tmpDir, err := os.MkdirTemp("/tmp", "tetragon-matchbinlarge-*")
```
mtardy, line 3894: or you could use `t.TempDir`

bpf/process/bpf_execve_event.c
```diff
@@ -329,19 +330,29 @@ execve_send(void *ctx)
...
-			curr->bin.path_length--;
+		// reuse heap->maxpath that contains the filename.
+		heap = map_lookup_elem(&execve_heap, &zero);
+		if (heap) {
+			curr->bin.path_length = probe_read_str(curr->bin.path, BINARY_PATH_MAX_LEN, &heap->maxpath);
+			if (curr->bin.path_length > 1) {
+				// don't include the NULL byte in the length
+				curr->bin.path_length--;
+			}
 		}
```
mtardy, line 356: I don't understand the difference here between what you wrote and the existing code. iirc `&p->args` is just the args containing the filename (+ the rest of the args) that was put into `execve_heap` out of implementation details?

bpf/process/bpf_execve_event.c
mtardy: Thinking a bit more about this, it's indeed not great that we never check for the error at read from https://github.com/cilium/tetragon/pull/1926 and that fallback to the `args`' filename could be an idea but it looks flawed.  Indeed, we have unit tests on the function that uses `read_exe` and the error should be filled if the path is too big to enter the buffer of 256 (it seems that the limit is 255) so we should be aware that we could not match properly.

Because as your test highlights, an errored dentry will not work since it's reading from the end to the start, and thus the beginning of the path containing the prefix will not be here. Switching back to the old implem (as it seems you do even though you read from the execve_heap instead of p->args) could work until you don't use any relative path or symlink.

So I see two things:
- support the error more explicitly and return an error to the user so that he knows we might have skipped an event because of path len limitation
- fix the issue by increasing `BINARY_PATH_MAX_LEN` to 4096.

mtardy (changes requested):

bpf/process/bpf_execve_event.c
anfedotoff: Thanks for detailed review! Yes you are right, relative paths and symlinks might be a problem. It is good to find a better solution that I proposed. Let me explain what I've done.

IIUC, we read the first argument here.
https://github.com/cilium/tetragon/blob/1a7b15abb3d48802f7f44d44a1c2c7cb58ec51f7/bpf/process/bpf_execve_event.c#L59-L69

Here we think that p->args contains the first argument, but it seems to me that p->args contains arguments starting from the second one. So that's why I changed the logic for old kernels.
https://github.com/cilium/tetragon/blob/1a7b15abb3d48802f7f44d44a1c2c7cb58ec51f7/bpf/process/bpf_execve_event.c#L338

anfedotoff: > So I see two things:
support the error more explicitly and return an error to the user so that he knows we might have skipped an event because of path len limitation
fix the issue by increasing BINARY_PATH_MAX_LEN to 4096.

I think it's a good idea to increase BINARY_PATH_MAX_LEN to 4096.  If so we don't  need `file_copy_reverse` here. And also, we don't need to have extra variables to hold postfix path.

Do you see any limitations that can stop us to increase the BINARY_PATH_MAX_LEN?
IIUC, limitations with prefix/postfix matching we can avoid by copying exact amount of data (STRING_PREFIX_MAX_LENGTH / STRING_POSTFIX_MAX_LENGTH).

anfedotoff: We can have problems here, if  __LARGE_MAP_KEYS is not defined, I think:

https://github.com/cilium/tetragon/blob/1a7b15abb3d48802f7f44d44a1c2c7cb58ec51f7/bpf/process/types/basic.h#L1581-L1587.

But we can have a check, for this case.

bpf/process/bpf_execve_event.c
anfedotoff: I answered [here](https://github.com/cilium/tetragon/pull/2718#discussion_r1696680394).

bpf/process/bpf_execve_event.c
mtardy: > Here we think that p->args contains the first argument, but it seems to me that p->args contains arguments starting from the second one. So that's why I changed the logic for old kernels.

It's not the case, `event->process.args` is read in function `read_path` and the first string separated by `\0` is the first args, thus the binary given to sys_execve:

https://github.com/cilium/tetragon/blob/8e07d6cee05cda5f77a0191ca9250da5dc7c0b40/bpf/process/bpf_execve_event.c#L102-L111

mtardy: read_args, reads the args (hear the args of the execve binary not the args of the syscall), and thus use probe_read_str to remove the first part of the args, but does not touch it.

anfedotoff: Hmm... I see, thanks making it clear!

mtardy: > Do you see any limitations that can stop us to increase the BINARY_PATH_MAX_LEN?

It all boils down to make `prepend_name` support more than 256 bytes. I'm looking into it at the moment.

PR #2720, opened 2024-07-23
matchBinaries: support followChildren

kkourt:
This PR introduces a `FollowChildren:` attribute in `MatchBinaries`.
For now, only the `In` operator is supported but others can be added.
Another limitation is that only 64 of such sections are supported, but this number can also be increased.

```release-note
tracing: introduce FollowChildren attribute in MatchBinaries selector
```

### DIFF
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -31,6 +31,19 @@ struct {
 	__type(value, struct msg_data);
 } data_heap SEC(".maps");
 
+typedef __u64 mbset_t;
+
+/* tg_mbset_map holds a mapping from (binary) paths to a bitset of ids that it matches. The map is
+ * written by user-space and read in the exec hook to determine the bitset of ids of a binary that
+ * is executed.
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 1024);
+	__type(key, __u8[MATCH_BINARIES_PATH_MAX_LENGTH]);
+	__type(value, mbset_t);
+} tg_mbset_map SEC(".maps");
+
 FUNC_INLINE __u32
 read_args(void *ctx, struct msg_execve_event *event)
 {
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -31,8 +31,6 @@ struct {
 	__type(value, struct msg_data);
 } data_heap SEC(".maps");
 
-typedef __u64 mbset_t;
-
 /* tg_mbset_map holds a mapping from (binary) paths to a bitset of ids that it matches. The map is
  * written by user-space and read in the exec hook to determine the bitset of ids of a binary that
  * is executed.
@@ -271,6 +269,29 @@ execve_rate(void *ctx)
 	return 0;
 }
 
+/* update bitset mark */
+FUNC_INLINE
+void update_mb_bitset(struct binary *bin)
+{
+	__u64 *bitsetp;
+	struct execve_map_value *parent;
+
+	parent = event_find_parent();
+	if (parent) {
+		/* ->mb_bitset is used to track matchBinary matches to children (followChildren), so
+		 * here we propagate the parent value to the child.
+		 */
+		bin->mb_bitset = parent->bin.mb_bitset;
+	}
+
+	/* check the map and see if the binary path matches a binary
+	 * NB: only the In operator is supported for now
+	 */
+	bitsetp = map_lookup_elem(&tg_mbset_map, bin->path);
+	if (bitsetp)
+		bin->mb_bitset |= *bitsetp;
+}
+
 /**
  * execve_send() sends the collected execve event data.
  *
@@ -357,6 +378,8 @@ execve_send(void *ctx)
 			curr->bin.path_length--;
 		}
 #endif
+
+		update_mb_bitset(&curr->bin);
 	}
 
 	event->common.flags = 0;
--- a/bpf/process/bpf_process_event.h
+++ b/bpf/process/bpf_process_event.h
@@ -11,7 +11,8 @@
 
 #define ENAMETOOLONG 36 /* File name too long */
 
-#define MAX_BUF_LEN 256
+#define MATCH_BINARIES_PATH_MAX_LENGTH 256
+#define MAX_BUF_LEN		       256
 
 struct buffer_heap_map_value {
 	// Buffer is twice the needed size because of the verifier. In prepend_name
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1530,8 +1530,6 @@ struct {
 	__type(value, struct match_binaries_sel_opts);
 } tg_mb_sel_opts SEC(".maps");
 
-#define MATCH_BINARIES_PATH_MAX_LENGTH 256
-
 struct {
 	__uint(type, BPF_MAP_TYPE_ARRAY_OF_MAPS);
 	__uint(max_entries, MAX_SELECTORS); // only one matchBinaries per selector
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1520,6 +1520,7 @@ FUNC_INLINE size_t type_to_min_size(int type, int argm)
 struct match_binaries_sel_opts {
 	__u32 op;
 	__u32 map_id;
+	__u32 mbset_id;
 };
 
 // This map is used by the matchBinaries selectors to retrieve their options
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1578,6 +1578,14 @@ FUNC_INLINE int match_binaries(__u32 selidx)
 
 		switch (selector_options->op) {
 		case op_filter_in:
+			/* Check if we match the selector's bit in ->mb_bitset, which means that the
+			 * process matches a matchBinaries section with a followChidren:true
+			 * attribute either because the binary matches or because the binary of a
+			 * parent matched.
+			 */
+			if (current->bin.mb_bitset & (1UL << selector_options->mbset_id))
+				return 1;
+			fallthrough;
 		case op_filter_notin:
 			path_map = map_lookup_elem(&tg_mb_paths, &selidx);
 			if (!path_map)
--- a/pkg/mbset/mbset.go
+++ b/pkg/mbset/mbset.go
@@ -1,0 +1,130 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package mbset
+
+import (
+	"errors"
+	"fmt"
+	"path/filepath"
+	"sync"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/tetragon/pkg/api/processapi"
+	"github.com/cilium/tetragon/pkg/bpf"
+)
+
+const (
+	MapName   = "tg_mbset_map"
+	InvalidID = ^uint32(0)
+	MaxIDs    = 64 // this value should correspond to the number of bits we can fit in mbset_t
+)
+
+type bitSet = uint64
+
+func openMap() (*ebpf.Map, error) {
+	fname := filepath.Join(bpf.MapPrefixPath(), MapName)
+	ret, err := ebpf.LoadPinnedMap(fname, &ebpf.LoadPinOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w", fname, err)
+	}
+	return ret, nil
+}
+
+type state struct {
+	mu       sync.Mutex
+	nextID   uint32
+	mbsetMap *ebpf.Map
+}
+
+func newState() (*state, error) {
+	m, err := openMap()
+	if err != nil {
+		return nil, fmt.Errorf("failed to open map: %w", err)
+	}
+	return &state{
+		mbsetMap: m,
+	}, nil
+}
+
+func (s *state) Close() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.mbsetMap.Close()
+}
+
+// AllocID allocates a new ID
+func (s *state) AllocID() (uint32, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.nextID >= MaxIDs {
+		return InvalidID, fmt.Errorf("cannot allocate new id")
+	}
+	ret := s.nextID
+	s.nextID++
+	return ret, nil
+}
+
+// UpadteMap updates the map for a given id and its paths
+// (NB: only an In operator for the paths is supported)
+func (s *state) UpdateMap(id uint32, paths [][processapi.BINARY_PATH_MAX_LEN]byte) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if id == InvalidID {
+		return fmt.Errorf("invalid id")
+	} else if id >= MaxIDs {
+		return fmt.Errorf("unexpected id")
+	}
+
+	bit := uint64(1) << id
+	for _, path := range paths {
+		var val bitSet
+		var uflags ebpf.MapUpdateFlags
+		err := s.mbsetMap.Lookup(path, &val)
+		if errors.Is(err, ebpf.ErrKeyNotExist) {
+			val = bit
+			uflags = ebpf.UpdateNoExist
+		} else if err != nil {
+			return fmt.Errorf("failed to lookup mbset map: %w", err)
+		} else {
+			val |= bit
+			uflags = ebpf.UpdateExist
+		}
+
+		if err := s.mbsetMap.Update(path, val, uflags); err != nil {
+			return fmt.Errorf("failed to update mbset map: %w", err)
+		}
+	}
+
+	return nil
+}
+
+var (
+	glbSt          *state
+	glbErr         error
+	setGlobalState sync.Once
+)
+
+func getState() (*state, error) {
+	setGlobalState.Do(func() {
+		glbSt, glbErr = newState()
+	})
+
+	return glbSt, glbErr
+}
+
+func AllocID() (uint32, error) {
+	s, err := getState()
+	if err != nil {
+		return InvalidID, err
+	}
+	return s.AllocID()
+}
+
+func UpdateMap(id uint32, paths [][processapi.BINARY_PATH_MAX_LEN]byte) error {
+	s, err := getState()
+	if err != nil {
+		return err
+	}
+	return s.UpdateMap(id, paths)
+}
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cilium/tetragon/pkg/idtable"
 	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
 	"github.com/cilium/tetragon/pkg/kernels"
+	"github.com/cilium/tetragon/pkg/mbset"
 	"github.com/cilium/tetragon/pkg/reader/namespace"
 	"github.com/cilium/tetragon/pkg/reader/network"
 )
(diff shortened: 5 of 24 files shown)


bpf/process/bpf_process_event.h
```diff
@@ -11,7 +11,8 @@
 
 #define ENAMETOOLONG 36 /* File name too long */
 
-#define MAX_BUF_LEN 256
+#define MATCH_BINARIES_PATH_MAX_LENGTH 256
```
vparla, line 14: Do these lengths present a problem in terms of potential truncation?

kkourt: `MATCH_BINARIES_PATH_MAX` is the size of the paths users can specify in `values` under `MatchBinaries`. That is,  it only limits the size of the user-specified paths that we want to match against.

For example,

```
      - matchBinaries:
        - operator: "In"
          values:
          - "/usr/bin/bash"
 ```
 
 would work, but:
 
```
      - matchBinaries:
        - operator: "In"
          values:
          - "'/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo/foo"
 ```
 
 wouldn't.

dwindsor: Thank you for this work @kkourt! This will help us out immensely. 

One comment, could we in the future perhaps convert this to use the same string handling as was added in #2069 ? Not for now, just in the future.

kkourt: Yap, I've also thought the same :)

mtardy: so this is an issue on the prefix feature because of the way we read dentries, since we read them from end to start, you can escape a prefix match using a long path (>256). There are attempts to fix at the moment (https://github.com/cilium/tetragon/pull/2718) but the only durable fix would be to extend `MATCH_BINARIES_PATH_MAX_LENGTH` to 4096 somehow.

bpf/process/bpf_execve_event.c
```diff
@@ -31,6 +31,17 @@ struct {
...
 
+/* tg_mbset_map holds a mapping from (binary) paths to a bitset of ids that it matches. The map is
+ * written by user-space and read in the exec hook to determine the bitset of ids of a binary that
+ * is executed.
+ */
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 1024);
+	__type(key, __u8[MATCH_BINARIES_PATH_MAX_LENGTH]);
+	__type(value, mbset_t);
```
mtardy, line 42: it seems that you declare this map here with the type `mbset_t` but only define it a bit later in a7dbc8c7c5b8889baaaeb501c633eee9458448e2.

pkg/selectors/kernel.go
```diff
@@ -1181,6 +1182,16 @@ func ParseMatchBinary(k *KernelSelectorState, b *v1alpha1.BinarySelector, selIdx
...
 	sel := MatchBinariesSelectorOptions{}
 	sel.Op = op
+	sel.MBSetID = mbset.InvalidID
+	if b.FollowChildren {
+		if op != SelectorOpIn {
+			return fmt.Errorf("matchBinary: followChildren not yet implemented for operation '%s'", b.Operator)
+		}
+		sel.MBSetID, err = mbset.AllocID()
+		if err != nil {
+			return fmt.Errorf("matchBinary: %w", err)
```
mtardy, line 1192: nit: you can add something like `matchBinary: followChildren failed to allocate ID: %w`

bpf/process/bpf_execve_event.c
tpapagian, line 42: `mbset_t` is undefined here. Possibly move the definition in this patch instead of https://github.com/cilium/tetragon/pull/2720/commits/a7dbc8c7c5b8889baaaeb501c633eee9458448e2.

bpf/process/types/basic.h
```diff
@@ -1520,6 +1520,7 @@ FUNC_INLINE size_t type_to_min_size(int type, int argm)
 struct match_binaries_sel_opts {
 	__u32 op;
 	__u32 map_id;
+	__u32 mbset_id;
```
mtardy, line 1523: not a big fan of the naming here given it's already in the match binary selector option and it does not stand out that it's for the follow children feature. In general I think `mbset` for match binaries set does not give a strong idea it's behind the "follow children" feature.

bpf/process/bpf_execve_event.c
```diff
@@ -269,6 +269,26 @@ execve_rate(void *ctx)
 	return 0;
 }
 
+// update bitset mark
+FUNC_INLINE
+void update_mb_bitset(struct execve_map_value *curr)
```
mtardy, line 274: maybe take a pointer on `curr->bin` directly here?

bpf/process/bpf_execve_event.c
```diff
@@ -269,6 +269,26 @@ execve_rate(void *ctx)
...
+{
+	__u64 *bitsetp;
+	struct execve_map_value *parent;
+
+	parent = event_find_parent();
+	if (parent) {
+		// initialize bitset to parent's
+		curr->bin.mb_bitset = parent->bin.mb_bitset;
+	}
```
mtardy, line 283: Having a comment about why you do that instead of what is done could make this easier to get. If I understand here you do the case where the parent (let's say `/bin/bash`) was `In` MatchBinaries with `followChildren`. So you want the children to bear the bitset mark so that we can trigger an event on them, so you pass the bitset to the child.

bpf/process/bpf_execve_event.c
```diff
@@ -269,6 +269,26 @@ execve_rate(void *ctx)
...
+
+	// check the map and see if the binary path matches a binary
+	// NB: Only the In operator is supported for now
+	bitsetp = map_lookup_elem(&tg_mbset_map, curr->bin.path);
+	if (bitsetp)
+		curr->bin.mb_bitset |= *bitsetp;
```
mtardy, line 289: here you look at the map that was set by userspace on tracing policy loading, and if a new process is spawned after tetragon started, you add its bitset mark since it should be marked given the map.

bpf/process/bpf_execve_event.c
mtardy, line 274: could you make it more obvious that it's for the `followChildren` feature?

bpf/process/bpf_execve_event.c
```diff
@@ -355,6 +375,8 @@ execve_send(void *ctx)
 			curr->bin.path_length--;
 		}
 #endif
+
+		update_mb_bitset(curr);
```
mtardy, line 379: any particular reason to do it here at the very end or just, why not?

bpf/process/types/basic.h
```diff
@@ -1578,6 +1578,8 @@ FUNC_INLINE int match_binaries(__u32 selidx)
 
 		switch (selector_options->op) {
 		case op_filter_in:
+			if (current->bin.mb_bitset & (1UL << selector_options->mbset_id))
+				return 1;
```
mtardy, line 1582: A comment here would be welcomed to say that you short circuit matching since it's a child of a followed process. And I don't if we can write it but an explicit `[[fallthrough]]` would be amazing.

pkg/testutils/copyfile.go
```diff
@@ -28,3 +30,38 @@ func CopyFile(toPath, fromPath string, perm os.FileMode) error {
...
+	pattern := fmt.Sprintf("%s.*", bname)
+	outf, err := os.CreateTemp("", pattern)
+	if err != nil {
+		return "", err
+	}
+	defer outf.Close()
+
+	inf, err := os.Open(fname)
+	if err != nil {
+		os.Remove(outf.Name())
```
mtardy, line 45: why not using `defer` instead of having multiple `os.Remove(outf.Name())` at various return? Since they are done in reverse order you could:
```go
defer os.Remove(outf.Name())
defer outf.Close()
```

pkg/sensors/tracing/matchbinaries_follow_children_test.go
```diff
@@ -0,0 +1,86 @@
...
+	shPath, err := exec.LookPath("sh")
+	if err != nil {
+		t.Fatalf("failed to find 'sh' exec: %v", err)
+	}
+	tmpShPath, err := testutils.CopyFileToTmp(shPath)
+	if err != nil {
+		t.Fatalf("failed to copy 'sh' exec: %v", err)
+	}
+	t.Cleanup(func() {
+		// os.Remove(tmpShPath)
```
mtardy, line 38: I think you want to uncomment that, or pass `t` in your `testutils.CopyFileToTmp` to register the cleanup callback so that the caller doesn't need to.

pkg/mbset/mbset.go
```diff
@@ -0,0 +1,130 @@
...
+	return &State{
+		mbsetMap: m,
+	}, nil
+}
+
+func (s *State) Close() error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	return s.mbsetMap.Close()
+}
```
tpapagian, line 54: This does not seem to be used in this PR. Maybe remove that? As far as I can understand, the plan is to create the state once and have it throughout the execution of the agent. So there are no plans to close that.

pkg/sensors/tracing/matchbinaries_follow_children_test.go
mtardy, line 1: I'm curious about why in a separate new file? Most `matchBinaries` tests are in kprobe_test, so maybe you can put this one in `tracepoint_test` or we should have a separate test file that contain all the `matchBinaries` specific instead of grouping by hook which does not make a lot of sense here I agree!

pkg/sensors/tracing/matchbinaries_follow_children_test.go
tpapagian, line 38: Do we need that to be a comment?

mtardy (commented):
Thanks that's a cool feature!

I think my main point would be to make things a bit easier to understand as:
1. it's not easy to link the `mbset` naming to the follow children match binaries feature.
2. sometimes what we do in BPF can look a bit cryptic when it could be explained.

I think I understood the patch after taking a bit of time, but it might get complicated to get it in some time without more comments/better naming.

pkg/sensors/tracing/matchbinaries_follow_children_test.go
```diff
@@ -0,0 +1,86 @@
...
+	})
+	fmt.Printf("%s\n", tmpShPath)
```
tpapagian, line 40: Maybe remove that print?

pkg/sensors/tracing/matchbinaries_follow_children_test.go
```diff
@@ -0,0 +1,86 @@
...
+					FollowChildren: true,
+				}},
+			}},
+		}},
+	}
+
+	loadGenericSensorTest(t, spec)
+	t0 := time.Now()
+	loadElapsed := time.Since(t0)
+	t.Logf("loading sensors took: %s\n", loadElapsed)
```
tpapagian, line 63: I believe that you need to move `t0 := ...` before `loadGenericSensorTest`. But do we need to measure that?

tpapagian (approved):
LGTM, just added some small comments.

pkg/mbset/mbset.go
```diff
@@ -0,0 +1,130 @@
...
+
+// AllocID allocates a new ID
+func (s *State) AllocID() (uint32, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if s.nextID >= MaxIDs {
```
olsajiri, line 60: should we have some ID cleanup logic, because IIUC this will effectively fail to add policy with followChildren, when we cross the MaxIDs value, which is 64 now... we could reach that after 64 attempts to load policy through tetra interface right?

bpf/process/bpf_process_event.h
jrfastab: If we ifdef'd into 6.x kernels maybe 4096 is fine with better BPF? This way limitation only exists on <X.Y versions?

bpf/process/types/basic.h
kkourt: I have a different opinion. The mechanism is a matchBinaries set (`mbset`) and is used to implement the `followChidren` feature. I prefer to name things based on what they are rather than how they are used. Note that the latter can change. In any case, I'm happy to change it. What would be your suggestion of a better name?

bpf/process/bpf_process_event.h
mtardy: So I've been trying to extend the length to 4096, mostly making prepend_name (the function we use to read dentry) to support 4096 chars and I'm currently blocked when the compiler is replacing `memset` and `memcpy` when the object (from a map value) is too large.

bpf/process/bpf_execve_event.c
kkourt: Thanks, fixed!

pkg/selectors/kernel.go
kkourt: fixed, thanks!

bpf/process/bpf_execve_event.c
kkourt: Done, thanks.

bpf/process/bpf_execve_event.c
kkourt: Good point, added:
```
+               /* ->mb_bitset is used to track matchBinary matches to children (followChildren), so
+                * here we propagate the parent values to the child.
+                */

```

bpf/process/bpf_execve_event.c
kkourt: update_mb_bitset needs ->binary to be set for the map lookup. So it happens just after we set it.

bpf/process/bpf_execve_event.c
kkourt: https://github.com/cilium/tetragon/pull/2720#discussion_r1698160468 should make it clear, I believe.

bpf/process/types/basic.h
kkourt: It does not look like `[[fallthrough]]` is supported.

Added:
```
                case op_filter_in:
+                       /* Check if we match the selector's bit in ->mb_bitset, which means that the
+                        * process matches a matchBinaries section with a followChidren:true
+                        * attribute either because the binary matches or because the binary of a
+                        * parent matched.
+                        */
                        if (current->bin.mb_bitset & (1UL << selector_options->mbset_id))
                                return 1;
+                       /* fallthrough */
```

pkg/testutils/copyfile.go
kkourt: Dunno, for error handling I prefer using `os.Remove()` and t.Cleanup() for cleanup. I think the function is small enough that it does not matter.

pkg/sensors/tracing/matchbinaries_follow_children_test.go
kkourt: Yap, need to uncomment that. I guess it was a leftover.

pkg/sensors/tracing/matchbinaries_follow_children_test.go
kkourt: `tracepoint_test.go` is almost 1K lines now, and felt it would be better to start splitting tests up.

pkg/mbset/mbset.go
kkourt: I'd like to leave it so that if we ever use it directly (e.g,. for testing) we would now that we need to release resources for it.

Generally, when using global state I always try to implement everything as if we did not have global state, and then add a global variable on top of that. My intention is to i) make it easy to test, and ii) allow us to migrate to something that does not use global state, like Cilium's hive (https://github.com/cilium/hive). We are far from it at this point, but one can dream.

pkg/sensors/tracing/matchbinaries_follow_children_test.go
kkourt: Nope, removed it. Thanks!

pkg/sensors/tracing/matchbinaries_follow_children_test.go
kkourt: Yap, removed, thanks!

pkg/sensors/tracing/matchbinaries_follow_children_test.go
kkourt: I guess not, removed it. Thanks!

pkg/mbset/mbset.go
kkourt: Indeed we need cleanup logic to go over 64 policies but I left it as a followup since it's somewhat tricky. We can do it by introducing a lock to update clean up the bitset, but I'm trying to figure out if there is also a better way by passing the map updates to kernel side so that they can happen without a lock.

kkourt:
Thanks for the reviews, pushed a new version to address the feedback.

kkourt:
checkpatch complains with:
```
  Warning: WARNING:PREFER_FALLTHROUGH: Prefer 'fallthrough;' over fallthrough comment
```

But we do not seem to have support for it:

```
process/types/basic.h:1588:4: error: use of undeclared identifier 'fallthrough'                                                                                                                                                                                
                        fallthrough; 
```

mtardy:
> checkpatch complains with:
> 
> ```
>   Warning: WARNING:PREFER_FALLTHROUGH: Prefer 'fallthrough;' over fallthrough comment
> ```
> 
> But we do not seem to have support for it:
> 
> ```
> process/types/basic.h:1588:4: error: use of undeclared identifier 'fallthrough'                                                                                                                                                                                
>                         fallthrough; 
> ```

This is out of the scope of this PR and I don't want to block it on that but we could add what Cilium does in our compiler.h: https://github.com/cilium/cilium/blob/b189c578ea4da02467dc636e338683b8b6b4b012/bpf/include/bpf/compiler.h#L113-L115

pkg/mbset/mbset.go
olsajiri: ok, we have a sensor removal hook, maybe we could reclaim those ids from there

pkg/mbset/mbset.go
```diff
@@ -0,0 +1,130 @@
...
+func openMap() (*ebpf.Map, error) {
+	fname := filepath.Join(bpf.MapPrefixPath(), MapName)
+	ret, err := ebpf.LoadPinnedMap(fname, &ebpf.LoadPinOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w", fname, err)
+	}
+	return ret, nil
+}
+
+type State struct {
```
olsajiri, line 34: nit, could be local 'state' ?

olsajiri (approved):

pkg/mbset/mbset.go
kkourt: Yap, the tricky part is cleaning the bits from the processes in `execve_map`.

pkg/mbset/mbset.go
kkourt: done!

kkourt:
> > checkpatch complains with:
> > ```
> >   Warning: WARNING:PREFER_FALLTHROUGH: Prefer 'fallthrough;' over fallthrough comment
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > But we do not seem to have support for it:
> > ```
> > process/types/basic.h:1588:4: error: use of undeclared identifier 'fallthrough'                                                                                                                                                                                
> >                         fallthrough; 
> > ```
> 
> This is out of the scope of this PR and I don't want to block it on that but we could add what Cilium does in our compiler.h: https://github.com/cilium/cilium/blob/b189c578ea4da02467dc636e338683b8b6b4b012/bpf/include/bpf/compiler.h#L113-L115

Make sense, added `fallthrough`.

mtardy (approved):

lambdanis:
@kkourt I see `followChildren` is a required field, what's a breaking change and breaks existing policies. Can we make it not required and default to false?

kkourt:
> @kkourt I see `followChildren` is a required field in CRD - that's a breaking change and breaks existing policies. Can we make it not required and default to false?

Good point: https://github.com/cilium/tetragon/pull/2790

PR #2723, opened 2024-07-24
make: update install/k8s when running "make crds"

kkourt:


lambdanis (commented):
This is documented as two commands here: https://tetragon.io/docs/contribution-guide/making-changes/#making-changes-to-crds but making it one command makes sense. Could you also update the docs page?

kkourt:
> This is documented as two commands here: https://tetragon.io/docs/contribution-guide/making-changes/#making-changes-to-crds but making it one command makes sense. Could you also update the docs page?

Thanks Anna! I've updated the docs, and also changed the Makefile command to use the `tetragon/crds-yaml` target.

lambdanis (approved):
lgtm

Issue #2724, opened 2024-07-24
Getting Started Docs - Should remove the previous network policy before deploying enforcement


spbkaizo:
I was a little puzzled to find that the network enforcement policy doesn't block connections, until I realised the previous network policy was still applied.

On the enforcement page, the file access policy has a step to remove the previous policy.  The network section does not.

We should provide explicit instructions to remove the older (downloaded) policy.

I will create a PR.

PR #2725, opened 2024-07-24
doc: Add instructions to remove previous applied network policy.

spbkaizo:
Fixes: https://github.com/cilium/tetragon/issues/2724

### Description
The enforcement example for networking allows connections if the previous policy is left applied.


mtardy (commented):
Would be happy to merge that but now it needs a rebase as expected :)

mtardy:
Looks like it's been abandoned, I created https://github.com/cilium/tetragon/pull/3904 to fix this.

PR #2727, opened 2024-07-24
rthooks: image and helm fixes

kkourt:
see patches.


tpapagian (approved):
nit: The last 2 commits can be merged together in a single one. I believe that there will not be any case where reverting only one of those make sense. Feel free to ignore that for this PR, just a comment in general.

kkourt:
Thanks!

> nit: The last 2 commits can be merged together in a single one. I believe that there will not be any case where reverting only one of those make sense. Feel free to ignore that for this PR, just a comment in general.

I agree wrt to reverting. I have the habit of splitting patches that change code with auto-generation to better deal with conflicts. I'll leave it as and merge the PR, but maybe it would make sense to come up with a general rule on whether we want to split auto-generation with everything else or not. I'm personally on the former camp, but it's just a preference.

PR #2747, opened 2024-07-27
tetragon: only allow single instance to run on a node

### LINKED ISSUE #2672: Don't allow Tetragon instance to run when there is another one running

This is a follow up of discussion with @kkourt and @tixxdz. Currently Tetragon should be run as the single instance on a node (kernel). Otherwise its behavior can be unpredictable: instances can interfere on eBPF programs and/or maps, both on startup an cleanup. Tetragon writes an error to logfile if it sees another instance already up and running, but we would like to prevent it from starting. There is real-world case, our customer said they would like to "experiment" with Tetragon on a node while there is already an instance which protects this node. These experiments may lead to errors and they can break protection they have.

Tetragon should not start if it sees there is another instance already up and running. Error message should say that the reason for that, is possible unpredictable behavior because of the interference between instances.

_No response_

### Code of Conduct

inliquid:
### Description

This change will make Tetragon fail at startup if it finds out that another instance already created PID file and is still running. Previously it was only logging a warning.

Fixes #2672

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -222,19 +222,22 @@ func tetragonExecute() error {
 	proc.LogCurrentSecurityContext()
 
 	// When an instance terminates or restarts it may cleanup bpf programs,
-	// having a check here to see if another instance is already running, can
-	// help debug errors.
+	// having a check here to see if another instance is already running.
 	pid, err := pidfile.Create()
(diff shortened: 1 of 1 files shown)


jrfastab:
Agree sort of silly to launch if we know its going to be in a bad state.

PR #2748, opened 2024-07-29
tetragon: Assorted fixes

olsajiri:
adding missing docs and cgroup rate fixes

PR #2760, opened 2024-07-31
docs: fixes on new persistent enforcement and cgroup rate pages

mtardy:
cc @olsajiri, I was too slow to review #2748 properly and some things could have been enhanced on the docs side. Here are some changes.

Please see our `note` shortcodes and how shifting stuff can matter on the rendering of lists items. Also try to not group cmd and the output so that our users can copy-paste stuff.


olsajiri (approved):
lookgs good, thanks

PR #2764, opened 2024-08-02
bpf: bump prepend_name underlying buffer size 4096

### LINKED ISSUE #2758: matchBinaries `Prefix` operator fails to match path longer than 256 chars
The matchBinaries `Prefix` operator fails to match a path longer than 256 chars because of the way we read the binary path. We use the `exe` of the process and walk the dentry from end to beginning. Thus if the path is too long, the buffer contains an incorrect start.

A good situation looks like this:
```
 <--        buflen         -->
 -----------------------------
 |                /etc/passwd|
 -----------------------------
                 ^
                 |
                buf
 ```
 
 A bad situation looks like this:
```
 <- buflen ->
 -----------
 |tc/passwd|
 -----------
 error = 1
 ```

A permanent solution would be to increase the length `prepend_name` can work on. On a first attempt, it does not seem to be an issue. However, what is blocking is that `memcpy` and `memset` will fail on the larger struct (above 1K) even though we make sure those struct are not allocated on the stack (which is 512K).

mtardy:
Fixes #2758
### Description

The matchBinaries Prefix operator fails to match a path longer than 256 chars because of the way we read the binary path. We use the exe of the process and walk the dentry from end to beginning. Thus if the path is too long, the buffer contains an incorrect start.

### Changelog

```release-note
Fix a bug related to the matchBinaries Prefix operator by increasing the buffer size used by our dentry walk. Now the matchBinaries Prefix operator can correctly trigger a match on any path above 255 chars.
```

### DIFF
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -180,11 +180,22 @@ read_exe(struct task_struct *task, struct heap_exe *exe)
 	struct file *file = BPF_CORE_READ(task, mm, exe_file);
 	struct path *path = __builtin_preserve_access_index(&file->f_path);
 
-	exe->len = BINARY_PATH_MAX_LEN;
-	exe->off = (char *)&exe->buf;
-	exe->off = __d_path_local(path, exe->off, (int *)&exe->len, (int *)&exe->error);
-	if (exe->len > 0)
-		exe->len = BINARY_PATH_MAX_LEN - exe->len;
+	// we need to walk the complete 4096 len dentry in order to have an accurate
+	// matching on the prefix operators, even if we only keep a subset of that
+	char *buffer;
+
+	buffer = d_path_local(path, (int *)&exe->len, (int *)&exe->error);
+	if (!buffer)
+		return 0;
+
+	// buffer used by d_path_local can contain up to MAX_BUF_LEN i.e. 4096 we
+	// only keep the first 255 chars for our needs (we sacrifice one char to the
+	// verifier for the > 0 check)
+	if (exe->len > 255)
+		exe->len = 255;
+	asm volatile("%[len] &= 0xff;\n"
+		     : [len] "+r"(exe->len));
+	probe_read(exe->buf, exe->len, buffer);
 
 	return exe->len;
 }
@@ -364,7 +375,7 @@ execve_send(void *ctx)
 #ifdef __LARGE_BPF_PROG
 		// read from proc exe stored at execve time
 		if (event->exe.len <= BINARY_PATH_MAX_LEN) {
-			curr->bin.path_length = probe_read(curr->bin.path, event->exe.len, event->exe.off);
+			curr->bin.path_length = probe_read(curr->bin.path, event->exe.len, event->exe.buf);
 			if (curr->bin.path_length == 0)
 				curr->bin.path_length = event->exe.len;
 		}
--- a/.github/workflows/vmtests.yml
+++ b/.github/workflows/vmtests.yml
@@ -67,7 +67,7 @@ jobs:
         matrix:
            kernel:
               # renovate: datasource=docker depName=quay.io/lvh-images/kernel-images
-              - 'rhel8-20240620.102250'
+              - 'rhel8.9-20240806.173325'
               # renovate: datasource=docker depName=quay.io/lvh-images/kernel-images
               - 'bpf-next-20240624.013140'
               # renovate: datasource=docker depName=quay.io/lvh-images/kernel-images
--- a/Makefile
+++ b/Makefile
@@ -264,8 +264,10 @@ test: tester-progs tetragon-bpf ## Run Go tests.
 tester-progs: ## Compile helper programs for unit testing.
 	$(MAKE) -C $(TESTER_PROGS_DIR)
 
+## bpf-test: ## run BPF tests.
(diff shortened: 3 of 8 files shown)


mtardy:
cc @anfedotoff

anfedotoff:
Nice! Thanks, @mtardy. It seems that I can close #2718 you already cherry-picked the test.

mtardy:
> Nice! Thanks, @mtardy. It seems that I can close #2718 you already cherry-picked the test.

yeah I need to rework the test a little bit as well it seems to no work.

bpf/process/bpf_execve_event.c
```diff
@@ -169,11 +169,18 @@ read_exe(struct task_struct *task, struct heap_exe *exe)
...
+	// matching on the prefix operators, even if we only keep a subset of that
+	char *buffer;
+
+	buffer = d_path_local(path, (int *)&exe->len, (int *)&exe->error);
+	if (!buffer)
+		return 0;
+
+	// buffer used by d_path_local can contain up to MAX_BUF_LEN i.e. 4096 we
+	// only keep the first 255 chars for our needs
+	exe->len &= 0xFF;
```
kevsecurity, line 182: Masking to the lower 255 bits seems wrong. Surely if returned path is 257 in length, then masking it with 0xFF will make the length 1, instead of 255.

mtardy: hehe you caught the bug I caught during the rhel8 investigation indeed what I wanted here, nice catch thanks!
```diff
-       exe->len &= 0xFF;
+       bpf_printk("exe before mask: %d", exe->len);
+       if (exe->len >= 256) {
+               exe->len = 256;
+       }
```

mtardy:
I bumped into a rhel8 (rocky 8.6) bug, let's bump to a newer 8.9 that is fixed https://github.com/cilium/little-vm-helper-images/pull/612.

bpf/process/bpf_execve_event.c
kevsecurity: I think it should be exe->len = 255; (instead of 256) unless the comment is wrong?

kevsecurity:
CI looks pretty red ❓

bpf/process/bpf_execve_event.c
mtardy: I bumped the comment, why would you take 255 instead of 256?

mtardy:
> CI looks pretty red ❓

oh boy I just rebased and made the tiny change and indeed it seems everything is broken

bpf/process/bpf_execve_event.c
kevsecurity: It might not matter here, but often in BPF we set upper limits as one less than a power of two, so that we can do "&= MASK" type operations to constrain an index into the array (to satisfy the verifier as opposed to avoiding overflows). Maybe check that this buffer doesn't need this constraint?

mtardy: Indeed I didn't check not sure why but this is what makes the verifier unhappy. It's unhappy mostly because `exe->len` seems to be possibly negative. If I add a naive `if v < 0; v = 0` it gets optimized out.

kevsecurity: exe->len is a u32 so the compiler is absolutely sure it can't be <0 :D hence it gets rid of it. The verifier on the other hand isn't so sure (and rightly so).
Solution is something like:
```
if (exe->len > 255) {
    exe->len = 255;
}
asm volatile("%[len] &= 0xff;\n"
     : [len] "+r"(exe->len)
     :);
```
i.e. limit it to 255 to get the right logic, and then mask it to make the verifier happy, using "asm volatile" to force the compiler to insert it at this point.

mtardy: Yes this way it works I tried but I lost 1 char, I tried to make it work with 256 :D

mtardy: It seems that the inline asm is no longer needed, do you think we should continue use it like that?

kevsecurity: You only need the inline asm if the verifier complains. It varies between kernel versions so 🤷 .

mtardy: So apparently can:
- sacrifice 1 char to the verifier god and continue my day
- write a potentially broken bitwise operation on 32 bits and fight it

kevsecurity: I think sacrificing 1 byte (which we do everywhere) is less complex and therefore easier to maintain.

bpf/process/bpf_execve_event.c
```diff
@@ -180,11 +180,21 @@ read_exe(struct task_struct *task, struct heap_exe *exe)
...
+	char *buffer;
+
+	buffer = d_path_local(path, (int *)&exe->len, (int *)&exe->error);
+	if (!buffer)
+		return 0;
+
+	// buffer used by d_path_local can contain up to MAX_BUF_LEN i.e. 4096 we
+	// only keep the first 255 chars for our needs (we sacrifice one char to the
+	// verifier for the > 0 check)
+	if (exe->len >= 255)
```
kevsecurity, line 194: nit: if exe->len == 255, you don't need to set it to 255

mtardy: Ah yes it's not intentional

kevsecurity:
I think you need ```make format```

kevsecurity (approved):
LGTM

PR #2765, opened 2024-08-02
tetra: fix getevents cmd unexpected timeout

mtardy:
### Description

After commit 663a1e2ed51afa8590c07bac08822a98c0d12a43, we stop connecting to the server prior making the RPC. It creates a buggy situation when we are using streams: the connection get cancelled at timeout even if it's successfully streaming. Let's remove the timeout for the getevents command and only use the context based on the interrupt signals.

### Changelog

```release-note
Fix a bug where the tetra getevents command would timeout even if the connection was successful.
```

### DIFF
--- a/cmd/tetra/common/client.go
+++ b/cmd/tetra/common/client.go
@@ -72,15 +72,23 @@ func CliRun(fn func(ctx context.Context, cli tetragon.FineGuidanceSensorsClient)
 
 type ClientWithContext struct {
 	Client tetragon.FineGuidanceSensorsClient
-	Ctx    context.Context
-	conn   *grpc.ClientConn
-	cancel context.CancelFunc
+	// Ctx is a combination of the signal context and the timeout context
+	Ctx context.Context
+	// SignalCtx is only the signal context, you might want to use that context
+	// when the command should never timeout (like a stream command)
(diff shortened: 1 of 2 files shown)


cmd/tetra/common/client.go
```diff
@@ -94,11 +102,10 @@ func NewClientWithDefaultContextAndAddress() (*ClientWithContext, error) {
 func NewClient(ctx context.Context, address string, timeout time.Duration) (*ClientWithContext, error) {
 	c := &ClientWithContext{}
 
-	var timeoutContext context.Context
-	timeoutContext, c.cancel = context.WithTimeout(ctx, timeout)
+	c.SignalCtx, c.signalCancel = signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
 	// we don't need the cancelFunc here as calling cancel on timeout, the
 	// parent, will cancel its children.
```
willfindlay, line 107: This comment should be removed now I guess

mtardy: yeah correct the other should be enough
```
// The signal context is the parent of the timeout context, so cancelling
	// signal will cancel its child, timeout
```

kkourt (approved):

PR #2773, opened 2024-08-06
bpf: fix incorrect inline asm output operands sections 

mtardy:
In some places, certainly because of copy-paste we use the inputs operands instead of the output operands section.

I also spotted some incorrect constraints, `r+` instead of `+r` for example.


mtardy:
mmh they are more I missed many, I only searched for `rg "asm volatile.*::"` so far.

kevsecurity (approved):
Good work

mtardy:
> Good work

Thanks, I pushed a few [other fixes](https://github.com/cilium/tetragon/compare/41096481833d55d02c8eb4d6b4cdf4ba878c75eb..9c9656666bc7186428d078a932ba37a13088d348). I'll try to get John's review on this one.

PR #2776, opened 2024-08-07
[WIP] pod mapping via cgroup ids

kkourt:


.github/workflows/tetragon-rthook-pr.yaml
```diff
@@ -0,0 +1,83 @@
...
+
+    - name: extract build data
+      run: |
+        tar xf tetragon-rthooks.tar
+
+    - name: run test
+      run: |
+        cd tetragon-rthooks
+        ./scripts/minikube-install-hook.sh
+        ./scripts/minikube-test-hook.sh
```
kkourt, line 83: @cilium/github-sec :wave:

Is executing scripts from the repo in a `pull_request` target OK?

Based on offline discussions, I believe it should be OK but I might be missing something and, in any case, would be nice to verify.

joestringer: `pull_request` is the safer option ([docs](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request)) yep. 

Using `pull_request_target` is generally discouraged particularly due to the warning described in the docs: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target 

> Warning: For workflows that are triggered by the pull_request_target event, the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified and the workflow can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch. To help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered. For more information, see "[Keeping your GitHub Actions and workflows secure: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests)" on the GitHub Security Lab website.

pkg/cgidmap/cri.go
```diff
@@ -0,0 +1,123 @@
...
+	cond          sync.Cond
+	unresolvedIDs *list.List
+}
+
+type unmappedID struct {
+	podID  PodID
+	contID ContainerID
+}
+
+func (c *criResolver) enqeue(unmappedIDs []unmappedID) {
```
mtardy, line 38: this is a bit tricky to read, I would have appreciated maybe some comments like: in this case inputs bigger than maxQueue so recreating truncated queue. Fitting by removing first/last elements. Sorry I wasn't super familiar with container/list.

I discovered the `cond.Signal`/`cond.Wait` thing, I imagine there was no way to simplify this by using a Go channel? I imagine the issue is that it's not easy to manage the Go channel (remove entry, empty it, etc).

pkg/api/processapi/processapi.go
```diff
@@ -89,6 +89,7 @@ type MsgK8s struct {
 
 type MsgK8sUnix struct {
 	Docker string
+	Cgrpid uint64
```
mtardy, line 92: sorry this is opinionated nit: but why not just write `CgroupID` here? Is this user-facing? I honestly wondered why the name by reading it.

mtardy (approved):
Thanks, this is a nice improvement. I tried my best to review it.

On the nit side, I spotted many places where you use `ID` or `Id`, maybe that would be could to unify and use Golang convention `ID`. I saw some typos in the comments but not sure you want to spend the time fixing those in all those commits and didn't want to be annoying pointing them since we totally get the point anyway.

pkg/server/server.go
```diff
@@ -374,7 +376,11 @@ func (s *Server) RuntimeHook(ctx context.Context, req *tetragon.RuntimeHookReque
 	logger.GetLogger().WithField("request", req).Debug("Received a RuntimeHook request")
 	err := s.hookRunner.RunHooks(ctx, req)
 	if err != nil {
-		logger.GetLogger().WithField("request", req).WithError(err).Warn("Server RuntimeHook failed")
+		id := uuid.New()
+		logger.GetLogger().WithFields(logrus.Fields{
+			"logid": id,
+		}).WithError(err).Warn("server runtime hook failed")
+		return nil, fmt.Errorf("server runtime hook failed. Check agent logs with logid=%s for details", id)
```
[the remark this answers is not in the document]
kkourt: My thinking was that the error provides internal information for the server which maybe we don't want to make available to all clients. That being said, the gRPC interface is currently being considered trusted, so maybe above does not make a lot sense.

pkg/rthooks/rthooks.go
```diff
@@ -10,19 +10,18 @@
 // RuntimeHookRequest is issued.
 //
 // Specifically:
-//  - sensors can register their callbacks at init() using RegisterCallbacksAtInit
```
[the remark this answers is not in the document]
kkourt: I think this is `gofmt` doing this in my editor.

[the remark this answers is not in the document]
kkourt: Yap, it seems to be `gofmt`:

```diff
diff pkg/rthooks/rthooks.go.orig pkg/rthooks/rthooks.go                                                                                                                                                                                                                                                                        
--- pkg/rthooks/rthooks.go.orig                                                                                                                                                                                                                                                                                                
+++ pkg/rthooks/rthooks.go                                                                                                                                                                                                                                                                                                     
@@ -10,18 +10,18 @@                                                                                                                                                                                                                                                                                                            
 // RuntimeHookRequest is issued.                                                                                                                                                                                                                                                                                              
 //                                                                                                                                                                                                                                                                                                                            
 // Specifically:                                                                                                                                                                                                                                                                                                              
-//  - sensors can register their callbacks at init() using RegisterCallbacksAtInit                                                                                                                                                                                                                                            
-//    which registers in these hooks in globalRunner.                                                                                                                                                                                                                                                                         
-//  - after init(), GlobalRunner() can be used to retrieve this runner and pass                                                                                                                                                                                                                                               
-//    it to the gRPC server code so that it can execute these callbacks when a                                                                                                                                                                                                                                                
-//    RuntimeHookRequest is issued.                                                                                                                                                                                                                                                                                           
-//  - some of these hooks need access to pkg/watcher, so before passing the                                                                                                                                                                                                                                                   
-//    runner to gRPC server, we add the watcher as well. Hooks can access the                                                                                                                                                                                                                                                 
-//    watcher via the argument passed in the executed callback.                                                                                                                                                                                                                                                               
-//  - all callbacks are executed, i.e., if a callback returns an error execution of callbacks does                                                                                                                                                                                                                            
-//    not stop                                                                                                                                                                                                                                                                                                                
-//  - if any callback fails with an error, the gRPC server will return an error to the client (see                                                                                                                                                                                                                            
-//    pkg/server/server.go)                                                                                                                                                                                                                                                                                                   
+//   - sensors can register their callbacks at init() using RegisterCallbacksAtInit                                                                                                                                                                                                                                           
+//     which registers in these hooks in globalRunner.                                                                                                                                                                                                                                                                        
+//   - after init(), GlobalRunner() can be used to retrieve this runner and pass                                                                                                                                                                                                                                              
+//     it to the gRPC server code so that it can execute these callbacks when a                                                                                                                                                                                                                                               
+//     RuntimeHookRequest is issued.                                                                                                                                                                                                                                                                                          
+//   - some of these hooks need access to pkg/watcher, so before passing the                                                                                                                                                                                                                                                  
+//     runner to gRPC server, we add the watcher as well. Hooks can access the                                                                                                                                                                                                                                                
+//     watcher via the argument passed in the executed callback.                                                                                                                                                                                                                                                              
+//   - all callbacks are executed, i.e., if a callback returns an error execution of callbacks does                                                                                                                                                                                                                           
+//     not stop                                                                                                                                                                                                                                                                                                               
+//   - if any callback fails with an error, the gRPC server will return an error to the client (see                                                                                                                                                                                                                           
+//     pkg/server/server.go)                                                                                                                                                                                                                                                                                                  
 package rthooks                                                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                               
 import (       
 ```

File: pkg/rthooks/args_test.go [hunk 6]
```diff
@@ -0,0 +1,28 @@
...
+	ts := []test{{
+		path: "/kubepods/besteffort/pod05e102bf-8744-4942-a241-9b6f07983a53/f52a212505a606972cf8614c3cb856539e71b77ecae33436c5ac442232fbacf8",
+		id:   "05e102bf-8744-4942-a241-9b6f07983a53",
+	}, {
+		path: "/kubepods/besteffort/pod897277d4-5e6f-4999-a976-b8340e8d075e/crio-a4d6b686848a610472a2eed3ae20d4d64b6b4819feb9fdfc7fd7854deaf59ef3",
+		id:   "897277d4-5e6f-4999-a976-b8340e8d075e",
+	}}
+
+	for _, tc := range ts {
+		require.Equal(t, tc.id, podIdFromCgroupPath(tc.path))
```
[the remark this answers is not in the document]
kkourt: done

pkg/rthooks/args.go
```diff
@@ -59,7 +60,15 @@ func (arg *CreateContainerArg) PodID() (string, error) {
 	return podIdFromCgroupPath(arg.Req.CgroupsPath), nil
 }
 
+func containerIdFromCgroupPath(p string) string {
```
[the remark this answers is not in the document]
kkourt: Fixed.

pkg/rthooks/args.go
```diff
@@ -75,3 +79,28 @@ func (arg *CreateContainerArg) ContainerID() (string, error) {
...
+		return nil, err
+	}
+
+	nretries := 5
+	for i := 0; i < nretries; i++ {
+		pod, err = arg.Watcher.FindPod(podId)
+		if err == nil {
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
```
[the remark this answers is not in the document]
kkourt: We are not blocking for 50s, we are blocking for 50ms.

cmd/tetra/cri/cri.go
```diff
@@ -0,0 +1,79 @@
...
+	return ret
+}
+
+func versionCmd(flagVals *criFlags) *cobra.Command {
+	ret := &cobra.Command{
+		Use:   "version",
+		Short: "retrieve CRI version",
+		Args:  cobra.ExactArgs(0),
+		RunE: func(_ *cobra.Command, _ []string) error {
+			ctx := context.Background()
```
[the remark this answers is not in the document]
kkourt: Makes sense. Will leave it as a followup.

pkg/rthooks/args.go
```diff
@@ -93,14 +93,42 @@ func (arg *CreateContainerArg) Pod() (*corev1.Pod, error) {
...
 
 	return pod, err
 }
 
+func (arg *CreateContainerArg) findMirrorPod(hash string) (*corev1.Pod, error) {
+	var pod *corev1.Pod
+	var err error
+
+	nretries := 5
+	for i := 0; i < nretries; i++ {
```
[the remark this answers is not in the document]
kkourt: Added a patch at the end that uses  a common retry function.

pkg/api/processapi/processapi.go
kkourt: Because that's what we are already using in MsgK8s: https://github.com/cilium/tetragon/blob/ecdb38ff75056bdef8a7643cd46b3fe4489f9502/pkg/api/processapi/processapi.go#L88

pkg/cgidmap/cri.go
kkourt: I don't know a way to do this with go channels. Especially since what the code does is keep the more recent entries.

kkourt:
Thanks for the review @mtardy. Pushed a new version.

jrfastab:
Generic comment it would be best to keep patch max patch count around 10 or so this is getting rather large. Just for future.

pkg/rthooks/rthooks.go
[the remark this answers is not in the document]
willfindlay: Maybe a bit nitpicky but could we pick this out and squash it into the first commit to avoid the noise here?

willfindlay (approved):
LGTM overall. It's a little scary that we haven't tested this on cgroupv1 but I think it makes sense to mark as beta with a note in the documentation until we get the proper testing infra in place.

pkg/cgroups/cgroups.go
```diff
@@ -763,3 +764,84 @@ func HostCgroupRoot() (string, error) {
...
+
+// GetCgroupIdFromCgroupPath deals with some idiosyncrancies of container runtimes
+//
+// Typically, the container processes run in the cgroup path specified in the OCI spec under
+// cgroupsPath. crun, however, is an exception because it uses another directory (called subgroup)
+// under the cgroupsPath:
+// https://github.com/containers/crun/blob/main/crun.1.md#runocisystemdsubgroupsubgroup.
+//
+// This function deals with this by checking for a child directory. If it finds one (and only one)
+// it uses the cgroup id from the child.
```
tixxdz, line 815: So we did some hacks https://github.com/cilium/tetragon/blob/main/pkg/sensors/exec/procevents/proc.go#L96 , but not really convincing, the old branch that uses cgroup layer to tag subdirs with same cgroupID was also result of these runtime inconsistencies... 

So If I get you correctly here, you are using cgroupID of "container" subdir right?

I would rename this function to `GetCgroupIdFromSubGroup()`   less confusion with the other one

pkg/cgroups/cgroups.go
```diff
@@ -763,3 +764,84 @@ func HostCgroupRoot() (string, error) {
...
+	data, err := os.ReadFile(cgroupFile)
+	if err != nil {
+		return 0, err
+	}
+
+	pathPrefix := fmt.Sprintf("%s/1/root/sys/fs/cgroup", option.Config.ProcFS)
+
+	var path string
+	v2Prefix := "0::"
+	v1Prefix := fmt.Sprintf("%d:%s:", GetCgrpSubsystemIdx(), GetCgrpControllerName())
```
tixxdz, line 780: Why not call here if:  `GetCgroupMode() == CGROUP_UNIFIED`  assume only cgroupv2 and you will have one line to parse.

For everything else what you are already doing and it will be definitely under cgroupv1, even if systemd adds the cgroupv2 hierarchy, we don't use it, we follow the default k8s used one cgroupv1. So yeh looking for that:
`v1Prefix := fmt.Sprintf("%d:%s:", GetCgrpSubsystemIdx(), GetCgrpControllerName())`   will work, prefixing it later with pathprefix + controller name and the path.

pkg/cgroups/cgroups.go
```diff
@@ -763,3 +764,84 @@ func HostCgroupRoot() (string, error) {
...
+
+	lines := strings.Split(string(data), "\n")
+	for _, line := range lines {
+		if strings.HasPrefix(line, v2Prefix) {
+			path = fmt.Sprintf("%s/%s", pathPrefix, line[len(v2Prefix):])
+			break
+		}
+		// TODO: test cgroup v1 implementation
+		if strings.HasPrefix(line, v1Prefix) {
+			path = fmt.Sprintf("%s/%s/%s", pathPrefix, GetCgrpControllerName(), line[len(v2Prefix):])
```
tixxdz, line 790: here you are using v2prefix where it should be v1prefix

pkg/rthooks/args.go
```diff
@@ -0,0 +1,146 @@
...
+		return *p, nil
+	}
+
+	// retrieve the cgroup id from the host cgroup path.
+	//
+	// NB(kkourt): A better solution might be to hook into cgroup creation routines and create a
+	// mapping between directory and cgroup id that we maintain in user-space. Then, we can find
+	// the id using this mapping.
+	cgPath := arg.Req.CgroupsPath
+	cgRoot, err := cgroups.HostCgroupRoot()
```
tixxdz, line 43: Not related to this one, but we could cache the cgroup root, we already do that in other parts: https://github.com/cilium/tetragon/blob/c2f3dbd92890903780a2269e96d5a47ca5713b50/pkg/cgroups/cgroups.go#L165  and it won't change, so this part of code can also just cache it.

tixxdz (commented):
Quick review as I didn't have time to get all of it.

tixxdz:
Kornilios what you mean by "talking to the CRI to update missing mappings (this is useful for pods started before tetragon)" ?  and curious why we can't just do that from /proc?

pkg/cgroups/cgroups.go
tixxdz: Here it is a cgroupv1 bug, if fixed I don't see a reason why it should not work if all the logic runs after our cgroup environment discovery

pkg/rthooks/rthooks.go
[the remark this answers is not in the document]
kkourt: The reason, I have it in a different patch is that the:
> +//  - if any callback fails with an error, the gRPC server will return an error to the client (see                                                            
> +//    pkg/server/server.go)

is not true before. It becomes true after the changes are applied in the second patch where we actually return an error from the server.

pkg/cgroups/cgroups.go
kkourt: Done!

pkg/cgroups/cgroups.go
kkourt: Ah, nice catch! Thanks! Fixed.

pkg/rthooks/args.go
kkourt: That makes sense to me. There are a couple of more things that I would like to improve the cgroups package. I will leave this as a followup.

pkg/cgroups/cgroups.go
kkourt: > Why not call here if: GetCgroupMode() == CGROUP_UNIFIED assume only cgroupv2 and you will have one line to parse.

Haven't thought of that but, IMO, it makes the code more complicated so leaving things as is seems like the best option to me.

kkourt:
> Kornilios what you mean by "talking to the CRI to update missing mappings (this is useful for pods started before tetragon)" ? and curious why we can't just do that from /proc?

I mean that we are talking to the CRI (cri-o or containerd) socket to get information for existing containers. This information includes the cgroup path of the container, which we use to find the cgroup id. I think it should be possible to scan the filesystem and find the cgroup paths without talking to the CRI (ineed the policyfilter code does this), but the CRI is the source of truth for  this and using it seems like the best approach to  me.

pkg/cgroups/cgroups.go
tixxdz: yeh the problem is could be theoretical one has to check kernel sources, but there are hybrid installation where both are available, systemd mounts at it /sys/fs/cgroup/unified/ for cgroupv2 features and not resources management, now I can't recall how they show up inside the /proc/pid/cgroup or if it a named cgroup hierarchy, if they are by order, etc... I forget. But the point you may multiple lines there, so if we are checking that better improve it and separate it by:
1. if unified mode you only lookup for v2prefix
2. anything else you lookup for v1prefix

This way you don't parse them both at same time...

tixxdz: Looks perfect and nice style BTW ;-) !

kkourt: Thanks for the review!

kkourt:
Merging this. Left a couple of followups in the PR description based on the comments. Thanks everyone!

PR #2782, opened 2024-08-08
tetragon: Group missed metrics for policy and attach point

olsajiri:
Some of the sensors do not set policy for program and it could
    cause a problem with duplicate metrics values with empty policy.
    
    Making the missed metrics groups by policy and attach point
    values. Also storing only metrics that we managed to received
    correctly (instead of storing them with zero values).

### DIFF
--- a/pkg/metrics/kprobemetrics/collector.go
+++ b/pkg/metrics/kprobemetrics/collector.go
@@ -24,56 +24,86 @@ func NewBPFCollector() prometheus.Collector {
 	)
 }
 
-func collectLink(ch chan<- prometheus.Metric, load *program.Program) {
+func collectLink(load *program.Program) (uint64, bool) {
 	if load.Link == nil {
-		return
+		return 0, false
 	}
 
 	info, err := load.Link.Info()
 	if err != nil {
-		return
+		return 0, false
 	}
-
-	missed := uint64(0)
 
 	switch info.Type {
 	case link.PerfEventType:
 		if !bpf.HasMissedStatsPerfEvent() {
-			return
+			return 0, false
 		}
 		pevent := info.PerfEvent()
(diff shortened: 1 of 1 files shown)


pkg/metrics/kprobemetrics/collector.go
```diff
@@ -24,56 +24,91 @@ func NewBPFCollector() prometheus.Collector {
...
+		valLink, okLink := collectLink(load)
+		valProg, okProg := collectProg(load)
+
+		// Store metrics only when we retrieved them successfully.
+		if !okLink && !okProg {
+			continue
+		}
+
+		policy := load.Policy
+		if policy == "" {
```
lambdanis, line 92: I think it's fine to just leave an empty policy (I mean, technically you can create a policy named "unknown", then it would clash here).

pkg/metrics/kprobemetrics/collector.go
```diff
@@ -24,56 +24,91 @@ func NewBPFCollector() prometheus.Collector {
...
+		}
+
+		key := &missedKey{policy, load.Attach}
+
+		if okLink {
+			mapLink[key] = valLink
+		}
+		if okProg {
+			mapProg[key] = valProg
+		}
```
lambdanis, line 103: Should we add the value instead of overriding it if the key already exists? That's what I'd expect, but let me know if it doesn't make sense.

olsajiri: ugh.. thanks

pkg/metrics/kprobemetrics/collector.go
olsajiri: ok

lambdanis (approved):
lgtm

PR #2784, opened 2024-08-09
policyfiltermetrics: Refactor and rename metrics

lambdanis:
### Description

See commits.

### Changelog

```release-note
`tetragon_policyfilter_metrics_total` metric is renamed to `tetragon_policyfilter_operations_total`, and its `op` label is renamed to `operation`.
```

### DIFF
--- a/pkg/metrics/policyfiltermetrics/policyfiltermetrics.go
+++ b/pkg/metrics/policyfiltermetrics/policyfiltermetrics.go
@@ -4,6 +4,8 @@
 package policyfiltermetrics
 
 import (
+	"golang.org/x/exp/maps"
+
 	"github.com/cilium/tetragon/pkg/metrics"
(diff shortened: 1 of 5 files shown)


pkg/metrics/policyfiltermetrics/policyfiltermetrics.go
```diff
@@ -4,6 +4,8 @@
 package policyfiltermetrics
 
 import (
+	"golang.org/x/exp/maps"
```
mtardy, line 7: As a note (that we will forget) [maps.Values](https://pkg.go.dev/maps@master#Values) will be moved into the standard library for 1.23.

mtardy (approved):
Thanks, new changes in metrics pkg look cool, I took a look while reviewing! :)

PR #2790, opened 2024-08-10
crd: make followChildren optional

kkourt:
55117f55586f26ee9a8fc5eca734456eb55faf54 Introduced a new field (followChildren), but it was not marked as optional which is a breaking change.

Make the field optional and false by default.

fixes: 55117f55586f26ee9a8fc5eca734456eb55faf54

Reported by @lambdanis https://github.com/cilium/tetragon/pull/2720#issuecomment-2278425729


lambdanis (approved):
lgtm!

mtardy (approved):

PR #2791, opened 2024-08-11
chore(ci): Save time for build-clang workflow

Trung-DV:
### Description

Save time for https://github.com/cilium/tetragon/actions/workflows/build-clang-image.yaml, from minutes to seconds
image

### Changelog

```release-note
```


Trung-DV:
> .github/workflows/build-clang-image.yaml

I see that every PR relates to update Golang versions which modify `.github/workflows/build-clang-image.yaml` will trigger this workflow.

Or, may be we can use https://docs.docker.com/reference/dockerfile/#onbuild 🤔 ?

PR #2799, opened 2024-08-14
api/ops: Remove MsgOp* constants

lambdanis:
`MsgOp*` constants were duplicating `MSG_OP_*`. There is no point in maintaining two lists, and `MsgOp*` constants weren't added for some new features. Let's keep `MSG_OP_*` for consistency with other types and bpf.

Also remove `MsgOpKfreeSkb` (not used and not existing in bpf) and update metrics docs with missing opcodes.


jrfastab (approved):
Thanks!

PR #2818, opened 2024-08-16
IMA hashes in LSM events

anfedotoff:
Adding support for collecting file hashes calculated by Linux integrity subsystem (IMA).  IMA hashes will be contained in LSM events. To make it work you need to:

-  enable [LSM BPF](https://tetragon.io/docs/concepts/tracing-policy/hooks/#lsm-bpf)
- enable [IMA](https://sourceforge.net/p/linux-ima/wiki/Home/#enabling-ima-measurement)

Limitations of **bpf_ima** helpers (from #2409):

1. *bpf_ima_inode_hash/bpf_ima_file_hash* can be called only from **BPF_F_SLEEPABLE** lsm programs: lsm.s.
2. **lsm.s** programs are strictly limited for maps usage. Only arrays, hashes and ringbuffer are allowed. We can't use **BPF_MAP_TYPE_PROG_ARRAY** and **BPF_MAP_TYPE_PERF_EVENT_ARRAY**. It means for us that we can't use bpf-to-bpf calls from lsm.s programs and send events from lsm.s programs with perfbuffer.
3. only **trusted pointers** are allowed to call bpf_ima helpers. Some verifier error example:
 `38: (85) call bpf_ima_file_hash#193 R1 type=scalar expected=ptr_, trusted_ptr.`

Approach from #2409 is meant to overcome limitations above has a race condition. User mode handler can receive LSM event before IMA hash is stored to the separate map.

This approach was redesigned to get rid of race condition. The idea is follows. We split LSM generic sensor into three parts, which are loaded using bpf trampolines feature.

1. *bpf_generic_lsm_core*  does everything  but sending LSM event. It also initializes an empty *ima_hash*, which is used in later trampoline calls.
2. *bpf_generic_lsm_ima_\** (optional) is triggered after *bpf_generic_lsm_core*, calculates IMA hash and stores it in separate hash map.
3. *bpf_generic_lsm_output* just sends event using perf buffer. Also at this stage IMA hash is collected from the map (stored in p. 2).

Using this scheme we can avoid race condition. 
*Note: if we don't need to collect IMA hash, step 2 is skipped.*


anfedotoff:
@kkourt , Hi! May I ask you for a code review? Next Monday, on Tetragon community meeting I'll show a demo. I want to talk about some design decisions I made. Also I'm going  to speak about problems that I overcame and those are still exists.

kkourt:
> @kkourt , hi! May I ask you for a code review? Next Monday, on Tetragon community meeting I'll show a demo. I want to talk about some design decisions I made. Also I'm going to speak about problems that I overcame and those are still exists.

Hello :)

Thanks for the heads up! Will try and have a look before the community meeting. Looking forward to the demo!

bpf/process/bpf_generic_lsm.c
```diff
@@ -154,6 +168,75 @@ __attribute__((section("lsm/5"), used)) int
 generic_lsm_output(void *ctx)
 {
 	generic_output(ctx, (struct bpf_map_def *)&process_call_heap, MSG_OP_GENERIC_LSM);
+#ifdef __LARGE_MAP_KEYS
+	struct msg_generic_kprobe *e;
+	int zero = 0;
+
+	e = map_lookup_elem(&process_call_heap, &zero);
+	if (e && e->common.flags & MSG_COMMON_FLAG_IMA_HASH)
+		tail_call(ctx, &lsm_calls, TAIL_CALL_IMA_HASH);
```
kkourt, line 177: Isn't there a race with this approach?
That is, what guarantees are there that user-space will do the map lookup before the entry is added?

pkg/sensors/program/loader.go
```diff
@@ -423,6 +423,9 @@ func TracingAttach() AttachFunc {
 
 func LSMOpen(load *Program) OpenFunc {
 	return func(coll *ebpf.CollectionSpec) error {
+		delete(coll.Programs, "ima_file_open")
+		delete(coll.Programs, "ima_mmap_file")
+		delete(coll.Programs, "ima_bprm_check_security")
```
kkourt, line 428: Why do we remove the programs here?

kkourt (commented):
Thanks @anfedotoff. I had a first look. Please find some first comments below.

pkg/sensors/program/loader.go
anfedotoff: Because, we can't load them all (they are not generic). We need to load one of them only if `imaHash: true` in the policy. I load ima_* programs on demand, after all lsm generic programs are loaded.

bpf/process/bpf_generic_lsm.c
kkourt: Can we tailcall to a program that does the `generic_output`  from an lsm.s program?

anfedotoff: > Can we tailcall to a program that does the `generic_output` from an lsm.s program?

Unfortunately, no(. Becase we can't use `BPF_MAP_TYPE_PROG_ARRAY` maps in lsm.s programs. We can tail call in, but we can't taill call out :).

olsajiri:
I refreshed on LSM programs and I think we can do this differently

so LSM programs are attached through trampolines on specific bpf_lsm_XXX
function which is invoked as part of the lsm modules call chain on particular
lsm call back

and trampoline allows to attach and invoke multiple bpf programs to
specific function (or lsm hook)

so I think we could have the sleepable lsm hooks invoked first to get the
hash and update the map and then we will have generic lsm invoked that
would run the hash action, which would get the hash from the map and
send the event

```
  bpf_lsm_XXX
    -> trampoline
         run lsm.s/file_open
               map_update_elem(&ima_hash_map, &e->current, &hash, BPF_ANY);
         run lsm/generic_lsm
               map_lookup_elem(&ima_hash_map, &e->current, ...)
```

to ensure the generic lsm is invoked AFTER lsm.s/XXX hook we can rely
on kernel using standard lists when queuing programs for trampoline,
so you should just attach lsm.s hook first and then generic lsm

I guess there might be some hiccups when you'll implement that
(there always are) but so far it seems straightforward

WDYT?

pkg/sensors/tracing/genericlsm.go
```diff
@@ -102,6 +113,72 @@ func (k *observerLsmSensor) LoadProbe(args sensors.LoadProbeArgs) error {
...
+			var btfSpec *btf.Spec
+			if btfFilePath := cachedbtf.GetCachedBTFFile(); btfFilePath != "/sys/kernel/btf/vmlinux" {
+				// Non-standard path to BTF, open it and provide it as 'KernelTypes'.
+				var err error
+				btfSpec, err = btf.LoadSpec(btfFilePath)
+				if err != nil {
+					return fmt.Errorf("opening BTF file '%s' failed: %w", btfFilePath, err)
+				}
+			}
+			spec, err := ebpf.LoadCollectionSpec(args.Load.Name)
```
olsajiri, line 130: whatever solution we pick for this I think the IMA lsm hooks should go in separate object and we should be able to load it through normal LoadLSMProgram interface

olsajiri:
> When we move lsm.s/ima_* programe to the top, we will call helper no matter if tracing policy is satisfied or not.

anfedotoff:
> > When we move lsm.s/ima_* programe to the top, we will call helper no matter if tracing policy is satisfied or not.
> > ah ok, I did not realized you might want to filter :-\
> 
> could we split the lsm sensor into filtering program, then run the program with ima helper and then run a program that sends it out? bit of a stretch but trampolines seem to disable cpu migration so we could use the percpu heap map to store half-done event and send it out in the last program
> 
> IIUC trampolines with LSM programs will run registered lsm programs until one of them returns != 0 .. but will need to check on that
> 
> also I still need to check all the options for lsm sleepable programs

Hmm, maybe we can split sensor into three parts. If we can use the same percpu heap map this scheme might work. It seems to me, that your first solution is much easier to implement:).
I tried to get hashes for files that are out of scope `ima_policy=tcb`. Helper returned --EOPNOTSUPP. I think, that exactly what we want! Hash calculation is blocked. I'm  looking at the [__ima_inode_hash](https://github.com/torvalds/linux/blob/8d8d276ba2fb5f9ac4984f5c10ae60858090babc/security/integrity/ima/ima_main.c#L577) to make sure that hash calculation is block due to policy violation, but didn't succeed yet.

anfedotoff:
I found out that bpf_ima_file_hash must get hash despite ima policy. There is another reason that I can't get hash for a file that doesn't satisfy ima policy. Maybe I'll find time to debug helper to know the reason. But to put ima helper to the top it is not good... Maybe we should split it three part lsm sensor.

olsajiri:
> Hmm, maybe we can split sensor into three parts. If we can use the same percpu heap map this scheme might work. It seems to me, that your first solution is much easier to implement:).

yea, it might be better to do the initial solution first as a base and split to 3 parts lsm sensor as a follow up.. we might learn other things that could shape up the final solution when doing that ;-)

anfedotoff:
> > Hmm, maybe we can split sensor into three parts. If we can use the same percpu heap map this scheme might work. It seems to me, that your first solution is much easier to implement:).
> 
> yea, it might be better to do the initial solution first as a base and split to 3 parts lsm sensor as a follow up.. we might learn other things that could shape up the final solution when doing that ;-)

I thought, I can implement some example of our idea using ebpf-go, before coding in tetragon. Here is the repo:
https://github.com/anfedotoff/ima-ebpfgo-example, please have a look.

I have 3 bpf programs: three parts of LSM sensor. In the first program I emulate filtering and enforcing.  Second program is used to get hash and the last one to send event via perf buffer. What I noticed.

- we need to save information is event filtered or not in **test_msg** struct  (msg_generic_kprobe in case of tetragon). Don't forget about NoPost Action
- enforcing, filtering works fine.
-  it seems to me, that send can be earlier the hash calculation. (race condition now resides here).

Also, I'm not sure, that it is enough just to attach programs this [way](https://github.com/anfedotoff/ima-ebpfgo-example/blob/41af25c43f59c2de7158960ba264d06a656af07f/main.go#L43) to preserve the order when hook is triggered.

So far it seems that approach is not working for us, but I may be wrong.

olsajiri:
> So far it seems that approach is not working for us, but I may be wrong.

nice, but I'm not sure I understand what's wrong, could you please elaborate?  I'll try it anyway

anfedotoff:
> > So far it seems that approach is not working for us, but I may be wrong.
> 
> nice, but I'm not sure I understand what's wrong, could you please elaborate? I'll try it anyway

Yes, sure, I'll try to explain.
I attach three bpf programs using standard Atach interface for LSM programs in the following order:

1. init_bprm_check (create event, store filename of executed binary in the map, do filtering + blocking);
2. ima_bprm_check (**lsm.s** program, that calculates hashes and store it in the map);
3. send_bprm_check (send stored event in map using perfbuffer);

The question is can we rely on the order of execution when `bprm_check_security` is  triggered?

Here is some logs:

```
024/09/16 10:01:22 Waiting for events..
2024/09/16 10:01:44 Error file: ./tetragon. Code:0
2024/09/16 10:01:51 Error file: ./tetragon. Code:0
2024/09/16 10:01:57 Error file: ./tetragon. Code:0
2024/09/16 10:01:57 File: ./tetragon
2024/09/16 10:01:57 SHA256: 9c20c4a712cd95766b64bcddc1094b619092944b592be43d75c3f4b3fb1434d40000000000000000000000000000000000000000000000000000000000000000
2024/09/16 10:01:57 <=============>
2024/09/16 10:01:58 File: ./tetragon
2024/09/16 10:01:58 SHA256: 9c20c4a712cd95766b64bcddc1094b619092944b592be43d75c3f4b3fb1434d40000000000000000000000000000000000000000000000000000000000000000
2024/09/16 10:01:58 <=============>
2024/09/16 10:01:59 File: ./tetragon
2024/09/16 10:01:59 SHA256: 29aa689f38158d2e8941fa54e436f0260890af31cecad1e8799e5c2df7bc1ecc0000000000000000000000000000000000000000000000000000000000000000
2024/09/16 10:01:59 <=============>
2024/09/16 10:02:20 Error file: ./tetragon. Code:0
2024/09/16 10:02:22 File: ./tetragon
2024/09/16 10:02:22 SHA256: 4f291296e89b784cd35479fca606f228126e3641f5bcaee68dee36583d7c94830000000000000000000000000000000000000000000000000000000000000000
2024/09/16 10:02:22 <=============>
^C2024/09/16 10:02:31 Received signal, exiting..
```
According this logs we see, that the 1st program is always starts and finishes first.  Some times the third program starts before second is finished. `2024/09/16 10:02:20 Error file: ./tetragon. Code:0` this line means that there is no hash in event. It seems to me this because sleepable program sleeps and another bpf program beeing executed (this is send program in our case).

olsajiri:
> The question is can we rely on the order of execution when bprm_check_security is triggered?

I think we can, it's hardcoded in kernel.. and if it changes we could actually detect the way it's added
and use proper order attach

however it's the other way round (check https://pastebin.com/Y8EGQ0fg for the trampoline dump),
so when I reverse attach in your example I get proper order of executed programs

could you plz try the change below?

thanks,
jirka


```
diff --git a/main.go b/main.go
index 1d1fa138660d..0be26120a55f 100644
--- a/main.go
+++ b/main.go
@@ -40,11 +40,11 @@ func main() {
        }
        defer objs.Close()
 
-       initBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.InitBprmCheck})
+       sendBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.SendBprmCheck})
        if err != nil {
                log.Fatalf("linking LSM failed: %s", err)
        }
-       defer initBprmHook.Close()
+       defer sendBprmHook.Close()
 
        imaBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.ImaBprmCheck})
        if err != nil {
@@ -52,11 +52,11 @@ func main() {
        }
        defer imaBprmHook.Close()
 
-       sendBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.SendBprmCheck})
+       initBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.InitBprmCheck})
        if err != nil {
                log.Fatalf("linking LSM failed: %s", err)
        }
-       defer sendBprmHook.Close()
+       defer initBprmHook.Close()
 
        rd, err := perf.NewReader(objs.TcpmonMap, 65535)
        if err != nil {
```

anfedotoff:
> > The question is can we rely on the order of execution when bprm_check_security is triggered?
> 
> I think we can, it's hardcoded in kernel.. and if it changes we could actually detect the way it's added and use proper order attach
> 
> however it's the other way round (check https://pastebin.com/Y8EGQ0fg for the trampoline dump), so when I reverse attach in your example I get proper order of executed programs
> 
> could you plz try the change below?
> 
> thanks, jirka
> 
> ```
> diff --git a/main.go b/main.go
> index 1d1fa138660d..0be26120a55f 100644
> --- a/main.go
> +++ b/main.go
> @@ -40,11 +40,11 @@ func main() {
>         }
>         defer objs.Close()
>  
> -       initBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.InitBprmCheck})
> +       sendBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.SendBprmCheck})
>         if err != nil {
>                 log.Fatalf("linking LSM failed: %s", err)
>         }
> -       defer initBprmHook.Close()
> +       defer sendBprmHook.Close()
>  
>         imaBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.ImaBprmCheck})
>         if err != nil {
> @@ -52,11 +52,11 @@ func main() {
>         }
>         defer imaBprmHook.Close()
>  
> -       sendBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.SendBprmCheck})
> +       initBprmHook, err := link.AttachLSM(link.LSMOptions{Program: objs.imaPrograms.InitBprmCheck})
>         if err != nil {
>                 log.Fatalf("linking LSM failed: %s", err)
>         }
> -       defer sendBprmHook.Close()
> +       defer initBprmHook.Close()
>  
>         rd, err := perf.NewReader(objs.TcpmonMap, 65535)
>         if err != nil {
> ```

Hmm, I changed the order  (swap init and send programs) as you suggest. I got no events collected, but execution is blocked. Are you sure that the order is really proper? It seems to me, that send program is now executed before init program. (send -> ima -> init).
Anyway, the good news is that order is preserved. But other question is that sometimes we miss the hash in event. Does it mean that send program is started before ima program is finished?

olsajiri:
also I think we need to use hash for the data shared by the program, it looks like cpu migration is not actually disabled,
so we can't use per-cpu heap map in here

olsajiri:
hum, when I remove the 'filter' from init function, I get all executed files output:
(I'm still using the per-cpu heap, but I don't think we should rely on that)

```
root@amd:/home/jolsa/ima-ebpfgo-example# ./ima-test 
2024/09/16 12:04:42 Waiting for events..
2024/09/16 12:04:43 File: /usr/bin/ls
2024/09/16 12:04:43 SHA256: 12a6d908a68ccf6f9f3d799705577c28763f5deef6eddcff7643d6d8a6de543d0000000000000000000000000000000000000000000000000000000000000000
2024/09/16 12:04:43 <=============>
2024/09/16 12:04:45 File: /usr/bin/ls
2024/09/16 12:04:45 SHA256: 12a6d908a68ccf6f9f3d799705577c28763f5deef6eddcff7643d6d8a6de543d0000000000000000000000000000000000000000000000000000000000000000
2024/09/16 12:04:45 <=============>
2024/09/16 12:04:47 File: /usr/bin/ls
2024/09/16 12:04:47 SHA256: 12a6d908a68ccf6f9f3d799705577c28763f5deef6eddcff7643d6d8a6de543d0000000000000000000000000000000000000000000000000000000000000000
2024/09/16 12:04:47 <=============>
2024/09/16 12:04:49 File: /usr/bin/ls
2024/09/16 12:04:49 SHA256: 12a6d908a68ccf6f9f3d799705577c28763f5deef6eddcff7643d6d8a6de543d0000000000000000000000000000000000000000000000000000000000000000
2024/09/16 12:04:49 <=============>
2024/09/16 12:04:50 File: /usr/bin/w
2024/09/16 12:04:50 SHA256: ae49096b42cebab2f3a2557e32b0f3e3e5139fd7e22201c9dc5bcfab9701946c0000000000000000000000000000000000000000000000000000000000000000
2024/09/16 12:04:50 <=============>
2024/09/16 12:05:33 File: /usr/bin/git
2024/09/16 12:05:33 SHA256: 29aa689f38158d2e8941fa54e436f0260890af31cecad1e8799e5c2df7bc1ecc0000000000000000000000000000000000000000000000000000000000000000
2024/09/16 12:05:33 <=============>
2024/09/16 12:05:33 File: /usr/bin/pager
2024/09/16 12:05:33 SHA256: f95851f468afe5c6e287d741153bde008ef26844916f5dfcb916e849ef3ae8bb0000000000000000000000000000000000000000000000000000000000000000
2024/09/16 12:05:33 <=============>
2024/09/16 12:05:47 File: /usr/bin/vim
2024/09/16 12:05:47 SHA256: dfc4b48f9b7976f0b64f6e99702ad8b1e4fb1929958fe07da98dfeb88f0a23fb0000000000000000000000000000000000000000000000000000000000000000
2024/09/16 12:05:47 <=============>
```

anfedotoff:
> hum, when I remove the 'filter' from init function, I get all executed files output: (I'm still using the per-cpu heap, but I don't think we should rely on that)
> 
> ```
> root@amd:/home/jolsa/ima-ebpfgo-example# ./ima-test 
> 2024/09/16 12:04:42 Waiting for events..
> 2024/09/16 12:04:43 File: /usr/bin/ls
> 2024/09/16 12:04:43 SHA256: 12a6d908a68ccf6f9f3d799705577c28763f5deef6eddcff7643d6d8a6de543d0000000000000000000000000000000000000000000000000000000000000000
> 2024/09/16 12:04:43 <=============>
> 2024/09/16 12:04:45 File: /usr/bin/ls
> 2024/09/16 12:04:45 SHA256: 12a6d908a68ccf6f9f3d799705577c28763f5deef6eddcff7643d6d8a6de543d0000000000000000000000000000000000000000000000000000000000000000
> 2024/09/16 12:04:45 <=============>
> 2024/09/16 12:04:47 File: /usr/bin/ls
> 2024/09/16 12:04:47 SHA256: 12a6d908a68ccf6f9f3d799705577c28763f5deef6eddcff7643d6d8a6de543d0000000000000000000000000000000000000000000000000000000000000000
> 2024/09/16 12:04:47 <=============>
> 2024/09/16 12:04:49 File: /usr/bin/ls
> 2024/09/16 12:04:49 SHA256: 12a6d908a68ccf6f9f3d799705577c28763f5deef6eddcff7643d6d8a6de543d0000000000000000000000000000000000000000000000000000000000000000
> 2024/09/16 12:04:49 <=============>
> 2024/09/16 12:04:50 File: /usr/bin/w
> 2024/09/16 12:04:50 SHA256: ae49096b42cebab2f3a2557e32b0f3e3e5139fd7e22201c9dc5bcfab9701946c0000000000000000000000000000000000000000000000000000000000000000
> 2024/09/16 12:04:50 <=============>
> 2024/09/16 12:05:33 File: /usr/bin/git
> 2024/09/16 12:05:33 SHA256: 29aa689f38158d2e8941fa54e436f0260890af31cecad1e8799e5c2df7bc1ecc0000000000000000000000000000000000000000000000000000000000000000
> 2024/09/16 12:05:33 <=============>
> 2024/09/16 12:05:33 File: /usr/bin/pager
> 2024/09/16 12:05:33 SHA256: f95851f468afe5c6e287d741153bde008ef26844916f5dfcb916e849ef3ae8bb0000000000000000000000000000000000000000000000000000000000000000
> 2024/09/16 12:05:33 <=============>
> 2024/09/16 12:05:47 File: /usr/bin/vim
> 2024/09/16 12:05:47 SHA256: dfc4b48f9b7976f0b64f6e99702ad8b1e4fb1929958fe07da98dfeb88f0a23fb0000000000000000000000000000000000000000000000000000000000000000
> 2024/09/16 12:05:47 <=============>
> ```

Yes, that works for me too, no hashes are missed. I remove filter with out changing order of execution, some hashes were missed.

Now I'm looking at [__bpf_prog_enter_sleepable](https://github.com/torvalds/linux/blob/69a3a0a45a2f72412c2ba31761cc9193bb746fef/kernel/bpf/trampoline.c#L963) and [__bpf_prog_enter](https://github.com/torvalds/linux/blob/69a3a0a45a2f72412c2ba31761cc9193bb746fef/kernel/bpf/trampoline.c#L985)

> I think we need to use hash for the data shared by the program, it looks like cpu migration is not actually disabled,
so we can't use per-cpu heap map in here

It looks so... To use hash map is a good idea, but for now I don't know how to share key for  this map. If we can't use per-cpu heap.

olsajiri:
> It looks so... To use hash map is a good idea, but for now I don't know how to share key for this map. If we can't use per-cpu heap.

olsajiri:
> This Is the first thing that came up to my mind. But what if we have file_open hook and the same process opens several files will it be OK? I think I can try to check this, nevertheless.

hum, so if the key is thread id then single thread could have just one active file_open call and we should be fine, right?

anfedotoff:
> > This Is the first thing that came up to my mind. But what if we have file_open hook and the same process opens several files will it be OK? I think I can try to check this, nevertheless.
> 
> hum, so if the key is thread id then single thread could have just one active file_open call and we should be fine, right?

Ahh, yes, we have tid. You are right! It seems to be fine than. I'll try to fix my example. If every thing is fine, I'll start to refactor code in th PR.
Thank you for helping me to find the solution!

anfedotoff:
> > > This Is the first thing that came up to my mind. But what if we have file_open hook and the same process opens several files will it be OK? I think I can try to check this, nevertheless.
> > 
> > 
> > hum, so if the key is thread id then single thread could have just one active file_open call and we should be fine, right?
> 
> Ahh, yes, we have tid. You are right! It seems to be fine than. I'll try to fix my example. If every thing is fine, I'll start to refactor code in th PR. Thank you for helping me to find the solution!

Everything is fine! I can start refactoring :fireworks:

anfedotoff:
It works! No race condition! :sparkles:

bpf/process/bpf_generic_lsm_core.c
```diff
@@ -56,6 +56,15 @@ struct {
 	__type(value, struct event_config);
 } config_map SEC(".maps");
 
+#ifdef __LARGE_MAP_KEYS
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 1024);
+	__type(key, __u64);
+	__type(value, char[65]);
```
olsajiri, line 64: could we have a struct in here instead of just byte array? and some macro for the size

also I was wondering we should add some state to the hash value that would be increased each time it passed the bpf program in the chain and each program would make sure it is executed on the proper place of the chain

and we'd send some error in the event if it wasn't the case  ... so we would be sure the order of the programs in the trampoline chain is correct (in case it's evenr changed in future) .. wdyt?

bpf/process/types/basic.h
```diff
@@ -2286,6 +2296,13 @@ do_actions(void *ctx, struct selector_action *actions, struct generic_maps *maps
 {
 	bool post = true;
 	__u32 l, i = 0;
+	int zero = 0;
+	struct msg_generic_kprobe *e;
+
+	e = map_lookup_elem(maps->heap, &zero);
+	if (!e)
+		return 0;
+	e->post = true;
```
olsajiri, line 2305: I wonder we could do this directly in the lsm code code.. I think generic_actions could return post/nopost value
and lsm core code could use it to set e->post ... and all the other generic sensors could just ignore the generic_actions return value and return 0, which is what they do now anyway, wdyt?

pkg/sensors/tracing/genericlsm.go
```diff
@@ -342,23 +371,99 @@ func createGenericLsmSensor(
...
-		path.Join(option.Config.HubbleLib, loadProgName),
+		path.Join(option.Config.HubbleLib, loadProgCoreName),
 		lsmEntry.hook,
-		"lsm/generic_lsm",
+		"lsm/generic_lsm_core",
 		lsmEntry.hook,
 		"generic_lsm").
 		SetLoaderData(lsmEntry.tableId)
+
+	// Load ima program for hash calculating
```
olsajiri, line 442: could you please make some note in here that this is the part (I hope it is) where we set the order of the trampoline programs and that we might do feature detection for the programs order in the trampoline? thnx

bpf/process/types/basic.h
```diff
@@ -2211,6 +2215,12 @@ do_action(void *ctx, __u32 i, struct selector_action *actions,
 			e->common.flags |= MSG_COMMON_FLAG_USER_STACKTRACE;
 			e->user_stack_id = get_stackid(ctx, &stack_trace_map, BPF_F_USER_STACK);
 		}
+#ifdef __LARGE_MAP_KEYS
+		__u32 ima_hash = actions->act[++i];
+
+		if (ima_hash)
+			e->common.flags |= MSG_COMMON_FLAG_IMA_HASH;
```
olsajiri, line 2222: any reason why not to do separate action? IIUC it's just post action flag now, right?  I don't have strong opinion about that, just curious

pkg/sensors/tracing/lsm_test.go
```diff
@@ -236,3 +239,86 @@ spec:
 	err = jsonchecker.JsonTestCheck(t, ec.NewUnorderedEventChecker(lsmChecker))
 	assert.NoError(t, err)
 }
+
+func TestLSMIMAHash(t *testing.T) {
+	if !bpf.HasLSMPrograms() || !kernels.EnableLargeProgs() || !kernels.MinKernelVersion("6.0") {
+		t.Skip()
+	}
```
olsajiri, line 246: it'd be great to have this tested as part of CI, which I guess means booting with all lsm stuff.. I'll try to check on that

olsajiri (commented):
nice! left some comment, thanks

pkg/sensors/tracing/lsm_test.go
anfedotoff: For kernels 6+ bpf_ima_file_hash helper works with out IMA, so some vm tests are working already!

bpf/process/types/basic.h
anfedotoff: From my point of view, hash collection is like stacktrace collection. We already have a flag for it. With `NoPost` event neither hash collection nor stacktrace collection makes sense, because we need to post collected information in event. So, that's why I think Post action flag suits better than separate action.

bpf/process/types/basic.h
anfedotoff: Yes, I think, it's a good idea, I'll try!

bpf/process/bpf_generic_lsm_core.c
anfedotoff: Nice idea!

bpf/process/types/basic.h
```diff
@@ -2286,6 +2292,13 @@ do_actions(void *ctx, struct selector_action *actions, struct generic_maps *maps
 {
 	bool post = true;
 	__u32 l, i = 0;
+	int zero = 0;
+	struct msg_generic_kprobe *e;
+
+	e = map_lookup_elem(maps->heap, &zero);
+	if (!e)
+		return 0;
+	e->post = true;
```
olsajiri, line 2301: leftover? IIUC this hunk should go away right?

bpf/process/bpf_generic_lsm_ima_bprm.c
```diff
@@ -0,0 +1,53 @@
...
+	__u64 pid_tgid = get_current_pid_tgid();
+	struct ima_hash *dummy = map_lookup_elem(&ima_hash_map, &pid_tgid);
+
+	if (dummy && dummy->state == 1) {
+#ifdef __V61_BPF_PROG
+		hash.algo = ima_file_hash(bprm->file, &hash.value, 64);
+#else
+		hash.algo = ima_inode_hash(bprm->file->f_inode, &hash.value, 64);
+#endif
+		hash.state = 3; // 1 + 2
```
olsajiri, line 49: curious, any reason why it's + 2 ?

bpf/process/bpf_generic_lsm_core.c
```diff
@@ -144,16 +147,32 @@ generic_lsm_filter_arg(void *ctx)
...
+
+	e->post = postit;
+#ifdef __LARGE_MAP_KEYS
+	// Set dummy hash entry for ima program
+	if (e && e->common.flags & MSG_COMMON_FLAG_IMA_HASH && e->post) {
+		struct ima_hash hash;
+
+		__u64 pid_tgid = get_current_pid_tgid();
+
+		memset(&hash, 0, 66);
```
olsajiri, line 167: please use sizeof(struct ima_hash) instead of 66

bpf/process/bpf_generic_lsm_output.c
```diff
@@ -0,0 +1,93 @@
...
+	if (e && e->common.flags & MSG_COMMON_FLAG_IMA_HASH) {
+		__u64 pid_tgid = get_current_pid_tgid();
+		struct ima_hash *hash = map_lookup_elem(&ima_hash_map, &pid_tgid);
+
+		if (hash && hash->state == 3) {
+			// Copy hash after all arguments
+			if (e->common.size + 66 <= 16383) {
+				probe_read(&e->args[e->common.size & 16383], 66, (char *)hash);
+				e->common.size += 66;
+			}
```
olsajiri, line 85: please use sizeof instead of 66

cmd/tetra/getevents/getevents.go
```diff
@@ -36,18 +36,19 @@ type Opts struct {
 	Timestamps    bool
```
olsajiri, line 36: could you please some example output into changelog? generally all the commits should have some description so one does not need to parse the diff to get an idea ;-) thanks

bpf/process/types/basic.h
anfedotoff: Ahh, thanks! Exactly.

bpf/process/bpf_generic_lsm_ima_bprm.c
anfedotoff: I waited for this question:)). It seems, that I've done a bit overengineering. My idea was that if we sum different values, we can understand which program was not triggered (but we can't know if the order of execution was correct). I think just incrementing in is more intuitive. Anyway if error happens, we will need to debug. I'll put plus one at each stage.

cmd/tetra/getevents/getevents.go
anfedotoff: Yes, sure!

olsajiri (approved):
leeks great, thanks

bpf/process/bpf_generic_lsm_ima_bprm.c
```diff
@@ -0,0 +1,53 @@
...
+	struct ima_hash hash;
+	__u64 pid_tgid = get_current_pid_tgid();
+	struct ima_hash *dummy = map_lookup_elem(&ima_hash_map, &pid_tgid);
+
+	if (dummy && dummy->state == 1) {
+#ifdef __V61_BPF_PROG
+		hash.algo = ima_file_hash(bprm->file, &hash.value, MAX_IMA_HASH_SIZE);
+#else
+		hash.algo = ima_inode_hash(bprm->file->f_inode, &hash.value, MAX_IMA_HASH_SIZE);
+#endif
```
kkourt, line 48: Could we check this with CORE instead of adding a compile-time dependency?

api/v1/tetragon/tetragon.proto
```diff
@@ -546,6 +546,8 @@ message ProcessLsm {
     KprobeAction action = 8;
     // Tags of the Tracing Policy to categorize the event.
     repeated string tags = 9;
+    // IMA file hash. Format algorithm:value.
+    string ima_hash = 11;
```
kkourt, line 550: Is there a reason for not introducing a separate struct with the two fields?

E.g., 

```
message ImaFileHash {
    string (?) algorithm
    string (?) value
}
```

Question: Is the value of the hash encoded in  a string or is it bytes?

kkourt (changes requested):
Thanks, this looks good to me overall!

I have some requests for changes. Please let me know if they do not make sense to you. I have not followed the discussion with Jiri. Can you please add a summary of the discussion in the PR header. Having a discussion of what the previous approach was, why it did not work, and what this PR does that works would be very helpful!

Thanks!

anfedotoff:
> Thanks, this looks good to me overall!
> 
> I have some requests for changes. Please let me know if they do not make sense to you. I have not followed the discussion with Jiri. Can you please add a summary of the discussion in the PR header. Having a discussion of what the previous approach was, why it did not work, and what this PR does that works would be very helpful!
> 
> Thanks!

Sure! I updated the PR description. I hope, now it became more clear about what we've done to avoid the race condition.

bpf/process/bpf_generic_kprobe.c
```diff
@@ -165,7 +165,8 @@ generic_kprobe_filter_arg(void *ctx)
 __attribute__((section("kprobe/4"), used)) int
 generic_kprobe_actions(void *ctx)
 {
-	return generic_actions(ctx, &maps);
+	generic_actions(ctx, &maps);
+	return 0;
```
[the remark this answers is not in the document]
anfedotoff: Yes, sure I explain.  Please, have a look here:
https://github.com/cilium/tetragon/blob/19ba89615341a2d18c6b8f1037a55c657863eee8/bpf/process/types/basic.h#L2377-L2381

So, we have two options : tail call or always return 0. That's why we do not break anything. If you wish, I can make a separate PR for this change, or we can keep it this way?

api/v1/tetragon/tetragon.proto
anfedotoff: I took the default representation of IMA hashes from here: 

```
sudo head /sys/kernel/security/integrity/ima/ascii_runtime_measurements
10 bfc19aede7d85629786e0d59148778666c755db9 ima-ng sha256:70b6c3ed60645b8c33e4ccfc0af769960ace357b0d461a0d812000f5e1aabbc8 boot_aggregate
10 0b5a1ee5112b581fc74040146ce5bdb607c1e1ac ima-ng sha256:d9d1775d643f6f70a1a6f646dfe023052765f558a167ec58951ad2f1013b3e46 /init
10 fbc981e4b231cc9b0677302e85761ad99b628c6f ima-ng sha256:f5f25da321f9341462205d8fd6ec918bba89ad5ea5dc9f6d875d5174d1936422 /usr/bin/sh
```

Also, docker image ID is a hash, stored the same way:

```
docker images --no-trunc --quiet $IMAGE
sha256:90a85ddf9c2540f149a2de34c5d1e3d88f566bbed754cc2ebb56c439d8ca7f09
sha256:30e6fd08ea57517b0c3325ccff2d3dd351959f082a95c5a14a49694aab647a30
sha256:caa045052df8cfc98fe60290a5d99c9995c75811d4d2f6df1cf062549d1cbbc8
```

> Question: Is the value of the hash encoded in a string or is it bytes?

Hash value here is just a hex string, no encoding is here such as base64, etc.

bpf/process/bpf_generic_kprobe.c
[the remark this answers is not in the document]
kkourt: Let me try to explain my motivation for the above message.

From a reviewer's perspective what I see is a return value that used to not be ignored, now being ignored. Having this change in a separate patch (no need for a separate PR) with an explanation of why this is OK in the commit message makes things easier to review. Furthermore, if, at some point, we need to check `git` for changes having this as a separate change along with its reasoning will make it easier to understand the reasoning behind this change.

api/v1/tetragon/tetragon.proto
kkourt: I see, thanks! OK if it's the established way of doing things, let's leave it as is.

kkourt:
> > Thanks, this looks good to me overall!
> > I have some requests for changes. Please let me know if they do not make sense to you. I have not followed the discussion with Jiri. Can you please add a summary of the discussion in the PR header. Having a discussion of what the previous approach was, why it did not work, and what this PR does that works would be very helpful!
> > Thanks!
> 
> Sure! I updated the PR description. I hope, now it became more clear about what we've done to avoid the race condition.

Great Thanks! Can you please split the change that changes the return value into a separate commit with a short explanation of why it OK to do as discussed in another comment? Other than that, LGTM!

bpf/process/bpf_generic_kprobe.c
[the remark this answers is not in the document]
anfedotoff: Thanks for making it clear. I'll do a separate commit right before all other commits.

anfedotoff:
> > > Thanks, this looks good to me overall!
> > > I have some requests for changes. Please let me know if they do not make sense to you. I have not followed the discussion with Jiri. Can you please add a summary of the discussion in the PR header. Having a discussion of what the previous approach was, why it did not work, and what this PR does that works would be very helpful!
> > > Thanks!
> > 
> > 
> > Sure! I updated the PR description. I hope, now it became more clear about what we've done to avoid the race condition.
> 
> Great Thanks! Can you please split the change that changes the return value into a separate commit with a short explanation of why it OK to do as discussed in another comment? Other than that, LGTM!

@kkourt, can you look one more time plz)?

kkourt (approved):
Many thanks!

kkourt:
All tests ✅, merging! Thanks!

PR #2819, opened 2024-08-16
encoder: add compact fuzz tests and fix bugs related to unhandled event types

willfindlay:
### Description

We've been burned recently by unhandled event types in the compact encoder's code path. Introduce a new fuzz test for the compact encoder to ensure that we don't miss any in the future. A subsequent PR will introduce a new CI job that fuzzes for a little bit on every PR to try and catch any new cases that come up. For now, we use a static corpus.

The first fuzzing run found two unhandled event types:

- TEST events (which really probably don't need to be handled, but for completion let's just add a simple handler)
- uprobe events

Add an implementation for both of the above in the compact encoder.

### Changelog

```release-note
Fix missing cases in the compact encoder for tetra.
```


mtardy (commented):
Oh wow that's nice, are we sure we want to check in the corpus in the main repo though?

willfindlay:
> are we sure we want to check in the corpus in the main repo though

Do you have an alternative suggestion? I'd like to have these cases exercised in CI even under the normal `go test` command (without `-fuzz`).

jrfastab (approved):
Please for future at least put something in the commit msg body. Generally title without any details is slightly frowned on. Otherwise nice!

PR #2824, opened 2024-08-20
Watcher: fix NODE_NAME if missing

kevsecurity:
The pkg/reader/node package gets the value of the NODE_NAME environment variable (if the HUBBLE_NODE_NAME is empty). If it is empty, it gets the hostname and uses that. The pkg/watcher package just attempts to get the NODE_NAME environment variable and warns that it will not work as expected if it is empty or missing. This commit changes pkg/watcher so that it uses pkg/reader/node's function instead.

### DIFF
--- a/pkg/watcher/watcher.go
+++ b/pkg/watcher/watcher.go
@@ -6,13 +6,13 @@ package watcher
 import (
 	"errors"
 	"fmt"
-	"os"
 	"reflect"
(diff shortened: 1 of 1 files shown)


lambdanis (approved):

kkourt (approved):
LGTM, thanks.

PR #2827, opened 2024-08-20
Fix tetragon_process_cache_size metric

lambdanis:
tetragon_process_cache_size metric was increased on every add() and never decreased. This is incorrect - fix it, so that it's increased on add() only if there was no LRU eviction and it's decreased on remove().

### DIFF
--- a/pkg/process/cache.go
+++ b/pkg/process/cache.go
@@ -121,8 +121,9 @@ func (pc *Cache) refInc(p *ProcessInternal) {
 	atomic.AddUint32(&p.refcnt, 1)
 }
 
-func (pc *Cache) Purge() {
+func (pc *Cache) purge() {
 	pc.stopChan <- true
+	processCacheTotal.Set(0)
 }
 
 func NewCache(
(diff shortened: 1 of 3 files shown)


pkg/process/cache.go
```diff
@@ -121,8 +121,9 @@ func (pc *Cache) refInc(p *ProcessInternal) {
 	atomic.AddUint32(&p.refcnt, 1)
 }
 
-func (pc *Cache) Purge() {
+func (pc *Cache) purge() {
```
olsajiri, line 124: there's another Purge instance in here.. https://github.com/cilium/tetragon/blob/95bd82351ad8b39245f1feaf6c72ce2b4a3fcd99/pkg/process/cache.go#L49  but not sure it's needed

olsajiri (approved):

pkg/process/cache.go
lambdanis: The flow is: `purge()` sends to `stopChan`, then `cacheGarbageCollector()` reads from this channel and actually purges the inner `lru.Cache` (the line you pointed). So metric reset happens in a different goroutine than actual cleanup, but it should be fine.

Issue #1820, opened 2023-12-01
Helm: Refactor Tetragon DaemonSet label/annotation handling


PhilipSchmid:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the feature you would like

This is a follow-up item for https://github.com/cilium/tetragon/pull/1817#discussion_r1411124035.

Supersedes https://github.com/cilium/tetragon/issues/1257

### Describe your proposed solution

Refactor the Tetragon DaemonSet labels/annotations similarly to the Tetragon Operator (details: https://github.com/cilium/tetragon/pull/1817).

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #2830, opened 2024-08-21
helm: Support adding extra labels to ServiceMonitors

lambdanis:
Labels on both tetragon and tetragon-operator ServiceMonitors can be overridden
using tetragon[Operator].prometheus.ServiceMonitor.labelsOverride Helm values.
However, it's nice to be able to add custom labels without overriding all of
the default ones. Let's add such option via extraLabels Helm values.

While here, use only selector labels as selectors in ServiceMonitors.

Related: #1820


kkourt (approved):

kkourt:
@lambdanis LGTM, but I will leave merging to you :)

PR #2837, opened 2024-08-23
CI: Helm lint: Remove pipenv dependency

PhilipSchmid:
```release-note
CI: Helm lint: Remove pipenv dependency
```


lambdanis (approved):
thanks!

Issue #229, opened 2022-07-12
replace verifier.sh with a go verifier test that uses cilium/ebpf 


kkourt:
https://github.com/cilium/tetragon/blob/main/contrib/verify/verify.sh

mtardy:
Done by https://github.com/cilium/tetragon/pull/3433

PR #2838, opened 2024-08-23
verify.sh: Handle when bpf_verride_return is unavailable

russellb:
While taking a look at tetragon, I tried running verify.sh and
observed bpf_enforce and bpf_generic_kprobe* all fail because
bpf_override_return is not available. I also observed that this seems
be handled conditionally in tetragon via the
pkg.bpf.HasOverrideHelper() function.

This change updates verify.sh to check for bpf_override_return before
trying to load these programs. If it's not present, it will be skipped
cleanly since it's not expected to work.

A cleaner solution is to reuse the same check from the Go code.
Rewriting this script in Go is tracked in issue #229.

Signed-off-by: Russell Bryant <russell.bryant@gmail.com>

### DIFF
--- a/contrib/verify/verify.sh
+++ b/contrib/verify/verify.sh
@@ -8,6 +8,7 @@ shopt -s nullglob
 RED="\033[31m"
 BLUEUNDER="\033[34;4m"
 GREEN="\033[32m"
+YELLOW="\033[33m"
 NOCOLOR="\033[0m"
 TETRAGONDIR=/var/lib/tetragon
 DEBUG=0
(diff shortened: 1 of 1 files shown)


mtardy (approved):
Hey thanks for that, looks like a good patch for this script! FYI we have the same check that you can use in `tetra`, the CLI, you can use:

```shell-session
$ sudo ./tetra probe
override_return: true
buildid: true
kprobe_multi: false
uprobe_multi false
fmodret: true
fmodret_syscall: true
signal: true
large: true
link_pin: true
lsm: false
missed_stats_kprobe_multi: false
missed_stats_kprobe: false
```

but that's perfectly reasonable (better?) to use bpftool here directly

kkourt:
Thanks!

PR #2839, opened 2024-08-24
Expose kernel ringbuffer errors in metrics

lambdanis:
There are two metrics counting events lost in the ringbuffer:
* tetragon_missed_events_total, collected in BPF based on perf_event_output
  error
* tetragon_ringbuf_perf_event_lost_total, collected in observer based on
  Record.LostSamples from cilium/ebpf

When testing, I saw the former being higher than the latter. This might mean
there are failed writes to the ringbuffer that are not lost events seen by
Record.LostSamples. To investigate such issues, I'm adding error label to
tetragon_missed_events_total, representing kernel error returned by
perf_event_output.

According to @olsajiri EBUSY and ENOSPC would be the only we could really hit,
the rest is most likely due config error. So let's start with counting these
two, and aggregating other errors as "unknown". We can always split out more
errors in the future if needed.


bpf/lib/process.h
```diff
@@ -564,8 +564,13 @@ FUNC_INLINE struct execve_info *execve_joined_info_map_get(__u64 tid)
 _Static_assert(sizeof(struct execve_map_value) % 8 == 0,
 	       "struct execve_map_value should have size multiple of 8 bytes");
 
+#define SENT_FAILED_UNKNOWN 0 // unknown error
+#define SENT_FAILED_EBUSY   1 // EBUSY
+#define SENT_FAILED_ENOSPC  2 // ENOSPC
+#define SENT_FAILED_MAX	    3
```
olsajiri, line 570: nit, extra tab screwing the diff in terminal

bpf/lib/process.h
```diff
@@ -564,8 +564,13 @@ FUNC_INLINE struct execve_info *execve_joined_info_map_get(__u64 tid)
...
+
 struct kernel_stats {
-	__u64 sent_failed[256];
+	__u64 sent_failed[256][SENT_FAILED_MAX];
 };
```
olsajiri, line 574: just a note.. I think it's ok because it's still small map, but it is per cpu and this change multiplies the memory usage by 3

olsajiri (approved):
lgtm

olsajiri:
adding @tpapagian because I think he added those metrics originally

tpapagian (approved):

PR #2850, opened 2024-08-28
tracing: add flags to reject policies with signal and override actions

bentekkie:
Fixes 

### Description

This patch allows disabling of the 2 action types that allow modification of syscall behaviour.
This is desirable in certain scenarios to prevent privilage escalation via loading policies via the gRPC API.

This patch implements this by adding 2 flags to the tetragon binary
  - `--disable-signal-actions` which causes tetragon to reject any policy
  that specifies `Signal` or `Sigkill` actions
  - `--disable-override-actions` which causes tetragon to reject any policy
  that specifies `Override` actions

Rejection was preferred over silent removal of the disabled actions from the policy as that will have less unexpected behaviour.

### Changelog

```release-note
Add tetragon flags `--disable-signal-actions` and `--disable-override-actions` to allow rejection of policies with `Signal` or `Override` actions.
```

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -461,6 +461,7 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 	observer.GetSensorManager().LogSensorsAndProbes(ctx)
 	defer func() {
 		observer.RemoveSensors(ctx)
+		os.Remove(observerDir)
 	}()
 
 	pm, err := tetragonGrpc.NewProcessManager(


tixxdz:
Note that the grpc can be restricted by `--server-address unix:///var/run/tetragon/tetragon.sock` owned by root.

bentekkie:
This is not sufficient if we want a client running as a user to be able to stream events from tetragon

pkg/sensors/tracing/selectors.go
```diff
@@ -571,3 +574,37 @@ func populateStringPostfixFilterMap(
...
+		for _, selector := range selectors {
+			switch strings.ToLower(selector.Action) {
+			case "sigkill":
+				if option.Config.DisableSignalActions {
+					return fmt.Errorf("sigkill actions are disabled")
+				}
+			case "signal":
+				if option.Config.DisableSignalActions {
+					return fmt.Errorf("signal actions are disabled")
+				}
```
kkourt, line 593: nit: combine these two cases into a single case since they seem to be checking the same thing.

mtardy: I would also say, would we want to group these two options so that we don't end up with "too many flags"? Is it very likely that someone that would use one would like to use the other?

kkourt: I just meant that  having something like:
> case "sigkill", "signal":
> ...

PR #2856, opened 2024-09-01
Remove tetragon_msg_op_total metric

lambdanis:
tetragon_msg_op_total was counting events per opcode in the ring buffer queue. It wasn't particularly useful, as there are other metrics exposing similar numbers:
* tetragon_bpf_missed_events_total counting missed events per opcode in BPF
* tetragon_observer_ringbuf_queue_events_received_total counting total events received in the ring buffer queue
* tetragon_events_total counting events per event type in grpc

If needed, in the future we can add opcode label to metrics counting events in the observer:
* tetragon_observer_ringbuf_events_received_total
* tetragon_observer_ringbuf_queue_events_received_total
* tetragon_observer_ringbuf_queue_events_lost_total We could also add a metric counting all events (not only missed) per opcode in BPF. However, it's unclear if they could be useful - ringbuffer and events queue shouldn't discriminate different types of events, so having total counts of successful and missed events at each stage should be enough to troubleshoot capacity issues. There is still a per event type counter at the last stage, for monitoring overall data volume.


lambdanis:
Converting to draft as I need to rethink how these metrics should be used.

kevsecurity (approved):

PR #2858, opened 2024-09-02
docs for followChildren true

kkourt:
Docs for https://github.com/cilium/tetragon/pull/2720


mtardy (approved):
Thanks.

nit: we can use `caution` or `warning` block for the limitation so that they stand out but that is perfectly fine like this.

kkourt:
> Thanks.
> 
> nit: we can use `caution` or `warning` block for the limitation so that they stand out but that is perfectly fine like this.

Haven't thought of that but `caution` or `warning` seems more scary that I would like it to be. I don't feel strongly, but I think I'll keep it as is for now. We can always change it later.

PR #2859, opened 2024-09-02
helm: add an rthooks serviceAccount section

kkourt:


install/kubernetes/tetragon/templates/_helpers.tpl
```diff
@@ -82,6 +82,14 @@ ServiceAccounts
 {{- end -}}
 {{- end }}
 
+{{- define "rthooks.serviceAccount" -}}
+{{- if .Values.rthooks.serviceAccount.name -}}
+{{- printf  "%s" .Values.rthooks.serviceAccount.name -}}
+{{- else -}}
+{{- printf  "%s-rthooks-service-account" .Release.Name -}}
```
mtardy, line 89: nit: I would say you don't need to repeat "service-account" in the name since it is a service account object already, so it should be pretty obvious.

mtardy (approved):
Looks reasonable!

I would be curious about why we need this service account since it seems to be associated with no role.

install/kubernetes/tetragon/templates/_helpers.tpl
kkourt: We do it for the other service accounts, but I guess we can do the right thing here. Fixed.

kkourt:
Thanks @mtardy! pushed a new version that sets the default value of `create:` to false as discussed offline.

fgiloux (approved):
As Mahe wrote it may have been enough to make the SA name configurable in the rthooks DaemonSet but it still looks good to me.

install/kubernetes/tetragon/values.yaml
```diff
@@ -392,3 +392,8 @@ rthooks:
     override: ~
     repository: quay.io/cilium/tetragon-rthooks
     tag: v0.2
+  # -- rthooks service account.
+  serviceAccount:
+    create: false
+    annotations: {}
+    name: ""
```
mtardy, line 399: To imitate the others SA you can do (and that's evolutive)
```suggestion
  serviceAccount:
    name: ""
```
Or, if we never plan to add create and annotations (maybe not a good idea)
```suggestion
  serviceAccountName: ""
```

install/kubernetes/tetragon/templates/rthooks_serviceaccount.yaml
mtardy, line 1: remove that for now

install/kubernetes/tetragon/templates/_helpers.tpl
```diff
@@ -82,6 +82,14 @@ ServiceAccounts
...
 {{- end }}
 
+{{- define "rthooks.serviceAccount" -}}
+{{- if .Values.rthooks.serviceAccount.name -}}
+{{- printf  "%s" .Values.rthooks.serviceAccount.name -}}
+{{- else -}}
+{{- printf  "%s-rthooks" .Release.Name -}}
+{{- end -}}
+{{- end }}
+
```
mtardy, line 92:
```suggestion
```

install/kubernetes/tetragon/templates/rthooks-daemonset.yaml
```diff
@@ -37,6 +37,7 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      serviceAccountName: {{ include "rthooks.serviceAccount" . }}
```
mtardy, line 40:
```suggestion
      {{- with .Values.rthooks.serviceAccount.name }}
      serviceAccountName: {{ . }}
      {{- end }}
```

mtardy (commented):
If we can just make the `ServiceAccountName` configurable, I guess it would be enough?

kkourt:
Thanks Mahé! pushed a new version, PTAL.

kkourt:
Pushed a new version due to conflicts.

install/kubernetes/tetragon/values.yaml
```diff
@@ -395,3 +395,6 @@ rthooks:
     override: ~
     repository: quay.io/cilium/tetragon-rthooks
     tag: v0.3
+  # -- rthooks service account.
+  serviceAccount:
```
fgiloux, line 399: IIRC in the previous version the default was the release name, [like for the agent](https://github.com/cilium/tetragon/blob/main/install/kubernetes/tetragon/templates/_helpers.tpl#L69-L75).
With the new version you don't get the credentials of the tetragon SA to access the API. An alternative approach to achieve the same is with adding `automountServiceAccountToken: false` to spec.template.spec of the rthooks daemonset. This may be very specific to OpenShift but the SA identity will be checked for admitting a  pod using hostpath. Both the agent and the rthooks daemonset need it, which means that their respective SAs need to be bound to a convenient SCC.
The answer may be documentation but I wanted to mention this point.

fgiloux (approved):

install/kubernetes/tetragon/values.yaml
kkourt: > With the new version you don't get the credentials of the tetragon SA to access the API

API here being the k8s API server, correct? I think that's fine since rthooks does not need access to the API server. I think I'll just merge the change for now. Once/if we have a concrete problem, it would be easier to figure it out then.

Thanks!

Issue #2785, opened 2024-08-09
Improve monitoring errors


lambdanis:
- [ ] Review tetragon_errors_total and other "errors" metrics
- [ ] Stop reporting non-errors (as in: no action needed) as errors. Define separate metrics for "casual fails" if needed.
- [ ] Refactor errors metrics to make them more useful (all of these are subject to review)
  - Delete tetragon_errors_total{type="handler_error"} - it duplicates handler_errors_total
  - Consider further splitting tetragon_errors_total
  - Rename handler_errors_total to event_handler_errors_total?
  - Consider merging event cache errors metrics into one with error, event, entry labels
  - Consider merging kprobe ok and errors metrics into one with status label
- [ ] Standardize on `error` label (rather than `error_type`, `type`, etc)

AritraDey-Dev:
> Consider merging kprobe ok and errors metrics into one with status label

Hey, I started looking into it, but I wanted to get a quick opinion before I finish it up.

The issue is that the `errors` metric has a lot of specific labels (like `curr_fn`, `prev_fn`, etc.) to help debug, but the `ok` metric is just a simple counter.

If we merge them, the `ok` events are going to have empty strings for all those extra labels (like `{status="ok", curr_fn="", ...}`). It feels a little messy compared to just keeping them separate.

Is it worth merging them considering prometheus queries will be easier, or should we stick to the separate metrics?

mtardy:
> The issue is that the `errors` metric has a lot of specific labels (like `curr_fn`, `prev_fn`, etc.) to help debug, but the `ok` metric is just a simple counter.
> 
> If we merge them, the `ok` events are going to have empty strings for all those extra labels (like `{status="ok", curr_fn="", ...}`). It feels a little messy compared to just keeping them separate.
> 
> Is it worth merging them considering prometheus queries will be easier, or should we stick to the separate metrics?

Hey thanks for taking a look into this :)

Could you maybe just write how the metrics would look like in a text form here like:

### current metrics

- metric 1
- metric 2{label="something"}
- etc.

### new metrics

- metric{status="something",other_labels="...",etc.}

to see how it would look like quickly and see if it would be worth? Thanks a lot :)

AritraDey-Dev:
Hey thanks for looking into this,
Current metrics looks like this:
```
tetragon_generic_kprobe_merge_ok_total 0
[...]

PR #2862, opened 2024-09-02
Remove tetragon_errors_total{type="handler_error"} metric

lambdanis:
It was duplicating tetragon_handler_errors_total.

Part of #2785


kevsecurity (approved):

tpapagian (approved):

mtardy (approved):

olsajiri:
curious, can we still get the total number, like all the opcodes/err summed up?

lambdanis:
> curious, can we still get the total number, like all the opcodes/err summed up?

@olsajiri Yes, when querying metrics with PromQL you typically do something like
```
sum(rate(tetragon_bpf_missed_events_total[5m]))
```

In case of `tetragon_errors_total` metric however, such aggregation doesn't make much sense, because this metric is collecting various unrelated errors and non-errors. That's why I'm splitting it up, so that metrics queries are more meaningful.

PR #2864, opened 2024-09-02
renovate: add 'make metrics-docs' to post upgrade cmds

mtardy:
Also update target codegen to protogen and generate to crds.


mtardy:
Needed by https://github.com/cilium/tetragon/pull/2815, for example on why this is needed.

kkourt (approved):

PR #2887, opened 2024-09-07
chore(docs): rename killer to enforcer

### LINKED ISSUE #2517: update docs to use NotifyEnforcer
https://github.com/cilium/tetragon/blob/main/docs/content/en/docs/concepts/tracing-policy/selectors.md#notify-killer-action

see: https://github.com/cilium/tetragon/pull/2117

mozillazg:
Closes #2517

### Description

update docs to use NotifyEnforcer

### Changelog

```release-note
chore(docs): rename killer to enforcer
```

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -107,7 +107,7 @@ enum {
 	ACTION_SIGNAL = 9,
 	ACTION_TRACKSOCK = 10,
 	ACTION_UNTRACKSOCK = 11,
-	ACTION_NOTIFY_KILLER = 12,
+	ACTION_NOTIFY_ENFORCER = 12,
 };
 
 enum {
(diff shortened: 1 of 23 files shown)


bpf/process/types/basic.h
```diff
@@ -107,7 +107,7 @@ enum {
 	ACTION_SIGNAL = 9,
 	ACTION_TRACKSOCK = 10,
 	ACTION_UNTRACKSOCK = 11,
-	ACTION_NOTIFY_KILLER = 12,
+	ACTION_NOTIFY_ENFORCER = 12,
```
mtardy, line 110: is it okay @kkourt, I think we wanted to rename but this is technically a breaking change for API users now!

mtardy (commented):
Thanks for finishing this, let me just ask a few question first :)

bpf/process/types/basic.h
kkourt: It's  only internal, correct? So i think it should be fine.

mtardy:
Hey! I think we can merge that, thanks for the work! Could you bump the patch version of the CRD since those changes touch it, it's just here (don't forget to run `make crds` after that):

https://github.com/cilium/tetragon/blob/88036cd6e90f875744840d4ec168d8b50fa54050/pkg/k8s/apis/cilium.io/v1alpha1/version.go#L10

And can you please squash your commits into the first one when it's done?

bpf/process/types/basic.h
mtardy: nevermind I confused myself I thought it was a protobuf enum, no worry then.

mtardy (approved):
Thanks!

PR #2888, opened 2024-09-07
observer: Collect ring buffer metrics only once

### LINKED ISSUE #2834: observer: Collect ring buffer metrics only once
Currently observer collects ring buffer metrics - events received, events lost and errors - twice, as:
* [Prometheus metrics](https://github.com/cilium/tetragon/blob/ed287834b5e0a5b1264162f3958e5a99666b6a2f/pkg/observer/metrics.go#L20-L40)
* [Observer struct fields](https://github.com/cilium/tetragon/blob/198fd5f5da9a2a4abb171a5fb81269b76af4667b/pkg/observer/observer.go#L305-L307)

A small optimization would be to collect these numbers only once. The code can be refactored in different ways - e.g. embed Prometheus metrics in Observer struct or implement Prometheus metrics as a custom collector reading the Observer struct.

mozillazg:
Closes #2834

### Description

observer: Collect ring buffer metrics only once

### Changelog

```release-note
observer: Collect ring buffer metrics only once. The metrics values don't change, but we avoid keeping independent counters in the Observer struct.
```

### DIFF
--- a/pkg/observer/observer.go
+++ b/pkg/observer/observer.go
@@ -12,7 +12,6 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/cilium/ebpf"
@@ -27,6 +26,8 @@ import (
 	"github.com/cilium/tetragon/pkg/reader/notify"
 	"github.com/cilium/tetragon/pkg/sensors"
 	"github.com/cilium/tetragon/pkg/sensors/config/confmap"
+	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
 
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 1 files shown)


olsajiri (changes requested):
could you please add more comments in changelog on what's broken and how you try to fix that? thanks

mozillazg:
> could you please add more comments in changelog on what's broken and how you try to fix that? thanks

@lambdanis Could you give me some suggestions about how to write the changelog for this PR?

lambdanis (approved):
Thank you for the PR @mozillazg, looks good to me. I updated the changelog note.

@olsajiri Nothing is broken here, but it's a small optimization for better maintainability - instead of maintaining two sets of counters, we only increment metrics.

olsajiri:
> @olsajiri Nothing is broken here, but it's a small optimization for better maintainability - instead of maintaining two sets of counters, we only increment metrics.

would be great changelog ;-) thanks

olsajiri (approved):

Issue #2902, opened 2024-09-10
Missing pod info from process_kprobe events


tpapagian:
### What happened?

I have seen some `process_kprobe` events that are missing pod info. 

Full reproduction steps:

1. Create an EKS cluster (should be reproducible with other k8s installations as well):
```
$ cat eks-config.yaml
apiVersion: eksctl.io/v1alpha5
kind: ClusterConfig

metadata:
  name: apapag-test-fim
  region: eu-west-1
  version: "1.30"

managedNodeGroups:
- name: managed-ng-1
  instanceType: c6a.2xlarge
  desiredCapacity: 1
  amiFamily: AmazonLinux2023
  ssh: # import default public key (~/.ssh/id_rsa.pub)
    allow: true
$ eksctl create cluster -f eks-config.yaml
```

2. Enable [rthooks](https://tetragon.io/docs/concepts/runtime-hooks/) for the reproduction. I don't believe that this issue is related to the `rthooks`. To enable `rthooks`, I followed the instructions in https://tetragon.io/docs/installation/runtime-hooks/#containerd. More specifically, I logged in in the k8s node, updated `/etc/containerd/config.toml`, restarted `containerd`, and checked the `containerd` logs that everything is fine (and NRI is enabled).

3. Install Tetragon:
```
$ cat tetragon-oss-values.yaml
tetragon:
  enabled: true
  btf: /sys/kernel/btf/vmlinux
rthooks:
  enabled: true
  interface: "nri-hook"
$ helm install tetragon cilium/tetragon -n kube-system -f tetragon-oss-values.yaml
```

4. Create a `kprobe` based policy that runs only on pods with `app: "nginx"` labels and posts read/write calls that are done from `"/usr/sbin/nginx"` binary.

```
$ cat kprobe-file.yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "file-monitoring-nginx"
spec:
  podSelector:
    matchLabels:
      app: "nginx"
  kprobes:
  - call: "security_file_permission"
    syscall: false
    args:
    - index: 0
      type: "file"
    - index: 1
      type: "int"
    selectors:
    - matchBinaries:
      - operator: "In"
        values:
        - "/usr/sbin/nginx"
 $ kubectl apply -f kprobe-file.yaml
```

5. Waiting for events in another terminal with the following command:
```
[...]

PR #2904, opened 2024-09-10
[backport/v1.1][bugfix] Fix clone event caching due to missing pod info

tpapagian:
[upstream commit 20bba351bd5dd0476315bfb8334189cc351ba1b2]

The eventcache API provides 2 handlers.

These are:
RetryInternal -> called to setup process information Retry -> called to setup pod information

In the case of clone events, we used to have en empty implementation on the Retry handler. This results in an issue with missing pod information which is described in detail here: https://github.com/cilium/tetragon/issues/2902

This patch provides the proper Retry implementation to handle also those cases.

FIXES: https://github.com/cilium/tetragon/issues/2902


olsajiri (approved):
there's small conflict in process.go now, but other than that lgtm

PR #2907, opened 2024-09-11
helm: Remove deprecated tetragonOperator.skipCRDCreation value

BonySmoke:
### Description

In this PR https://github.com/cilium/tetragon/pull/2498, the `tetragon.skipCRDCreation` key was removed from the `values.yaml` file; however, there is still a reference to it in the operator config map.
From what I understand, the operator expects a boolean value, and in case the installation method is `operator`, it should be set to `false`.

The patch makes the following changes:
1. Remove deprecated `tetragon.skipCRDCreation` from the operator config map and set `false` as the static value in `skip-crd-creation`.
2. Rename `tetragon.skipCRDCreation` to `tetragonOperator.skipCRDCreation` in the upgrade notes for Tetragon `1.2.0`.

### Changelog

```release-note
Remove deprecated `tetragonOperator.skipCRDCreation` from the operator config map and set "false" when `.Values.crds.installMethod` is `operator`
```

### DIFF
--- a/contrib/upgrade-notes/v1.2.0.md
+++ b/contrib/upgrade-notes/v1.2.0.md
@@ -26,7 +26,7 @@ tetragon:
        - --retries
        - "5"
 ```
-* Deprecated `tetragon.skipCRDCreation` Helm value is removed. Use `crds.installMethod=none` instead.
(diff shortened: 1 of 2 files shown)


lambdanis (approved):
Thanks!

PR #2908, opened 2024-09-12
Added metrics for LRU data cache

### LINKED ISSUE #679: Add metrics for lru cache in pkg/observer/data.go
would be good to add metrics for lru cache in `pkg/observer/data.go`,
like we do in `pkg/process/cache.go`

so we can eventually watch and suggest better value for the cache

AshishNaware:
Fixes #679 

### Description
This PR adds additional metrics around LRU data cache such as size, capacity, evictions and cache miss in order to have a same set of metrics collection across the board.

### Changelog

```
Add metrics for LRU data cache
```

### DIFF
--- a/pkg/observer/cache.go
+++ b/pkg/observer/cache.go
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package observer
+
+import (
+	"fmt"
+
+	"github.com/cilium/tetragon/pkg/api/dataapi"
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+type cache struct {
+	cache *lru.Cache[dataapi.DataEventId, []byte]
+	size  int
+}
+
+// newCache constructs a cache of fixed size with the callback function that increments
+// data_cache_evictions_total counter every time the cache is evicted.
+func newCache(dataCacheSize int) (*cache, error) {
+	lruCache, err := lru.NewWithEvict(
+		dataCacheSize,
+		func(_ dataapi.DataEventId, _ []byte) {
+			dataCacheEvictions.Inc()
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+	cache := &cache{
+		cache: lruCache,
+		size:  dataCacheSize,
+	}
+	return cache, nil
+}
+
+func (c *cache) get(dataEventId dataapi.DataEventId) ([]byte, error) {
+	data, ok := c.cache.Get(dataEventId)
+	if !ok {
+		dataCacheMisses.WithLabelValues("get").Inc()
+		return nil, fmt.Errorf("data event with id : %v not found", dataEventId)
+	}
+	return data, nil
+}
+
+func (c *cache) add(id dataapi.DataEventId, msgData []byte) bool {
+	evicted := c.cache.Add(id, msgData)
+	if !evicted {
+		dataCacheTotal.Inc()
+	}
+	return evicted
+}
+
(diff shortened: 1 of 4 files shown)


pkg/observer/cache.go
```diff
@@ -0,0 +1,59 @@
+package observer
+
+import (
+	"fmt"
+
+	"github.com/cilium/tetragon/pkg/api/dataapi"
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+type Cache struct {
```
lambdanis, line 10: Everything here is in one package, so no need to export. Could you make it `cache` and `newCache()`, so that they're not used from other packages accidentally?

pkg/observer/cache.go
lambdanis, line 1: Tetragon Go files should start with a license header:
```
```
Could you add?

lambdanis (commented):
Thank you @AshishNaware, overall looks good to me. I left two comments about small details.

Tetragon metrics are also documented in a generated reference doc, so the new metrics must be added there too. To generate the metrics reference, run:
```
make metrics-docs
```
Then you should see updated doc files.

As a side note, at some point it might be nice to refactor, to have both process and observer cache use the same cache implementation, but with separate metrics. But this is out of scope in this PR.

pkg/observer/cache.go
AshishNaware: Good point. I made these changes in recent commit.

pkg/observer/cache.go
AshishNaware: done

pkg/observer/cache.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type cache struct {
+	cache *lru.Cache[dataapi.DataEventId, []byte]
+	size  int
+}
+
+func newCache(
+	dataCacheSize int,
+) (*cache, error) {
```
mtardy, line 20: nit: wrapping looks a bit funny here since it only has 1 arg

pkg/observer/cache.go
```diff
@@ -0,0 +1,62 @@
...
+	lruCache, err := lru.NewWithEvict(
```
mtardy, line 21: This changed from `lru.New` to `lru.NewWithEvict`, is it expected?

pkg/observer/cache.go
```diff
@@ -0,0 +1,62 @@
...
+	}
+	return data, nil
+}
+
+func (c *cache) add(id dataapi.DataEventId, msgData []byte) bool {
+	evicted := c.cache.Add(id, msgData)
+	if !evicted {
+		dataCacheTotal.Inc()
+	}
+	return evicted
```
mtardy, line 51: why don't we increment `dataCacheEvictions` here?

EDIT: oh this is the `NewWithEvict` I guess? It's a bit confusing here maybe we should add a comment there.

lambdanis (approved):
looks good to me, thanks!

lambdanis:
@AshishNaware Could you please squash the three commits and force push? That's a practice in the Tetragon project, to keep clean git history on the main branch. Currently one of the CI jobs is failing because one of the commit messages is too long.

pkg/observer/cache.go
AshishNaware: Resolving all threads as all of these were addressed in latest commits.

AshishNaware:
@lambdanis @mtardy - I have squashed the commits and rebased the branch. This PR has now passed all the checks.

mtardy (approved):
Thanks! :)

PR #2910, opened 2024-09-12
[v1.1] Fix tetragon_process_cache_size metric

lambdanis:
[ upstream commit: 9e35cbab946bf8c2d7d001dcf5a3b39b74768e8a ]

tetragon_process_cache_size metric was increased on every add() and never decreased. This is incorrect - fix it, so that it's increased on add() only if there was no LRU eviction and it's decreased on remove().


mtardy (approved):
Thanks!!

kkourt (approved):

PR #2928, opened 2024-09-16
Make event cache tunable

tpapagian:
This PR makes the event cache tunable. This includes the number of retries and the delay between each retry. For this reason we introduce 2 new command line flags:
```
      --event-cache-retries int                   Number of retries for event cache (default 15)
      --event-cache-retry-delay int               Delay in seconds between event cache retries (default 2)
```
This PR does not change the default values of those.

We also export those values through the Helm Chart.

```release-note
Make eventCache number of retries and delay between them tunable.
```

### DIFF
--- a/pkg/jsonchecker/jsonchecker.go
+++ b/pkg/jsonchecker/jsonchecker.go
@@ -15,15 +15,14 @@ import (
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
 	"github.com/cilium/tetragon/api/v1/tetragon/codegen/helpers"
-	"github.com/cilium/tetragon/pkg/eventcache"
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/testutils"
 	"github.com/sirupsen/logrus"
 )
 
-var (
+const (
 	Retries    = 13
-	RetryDelay = eventcache.EventRetryTimer + (1 * time.Second)
+	RetryDelay = 3 * time.Second
 )
 
 // DebugError is an error that will create a debug output message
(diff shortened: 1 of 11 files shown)


pkg/jsonchecker/jsonchecker.go
```diff
@@ -15,15 +15,14 @@ import (
...
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/testutils"
 	"github.com/sirupsen/logrus"
 )
 
-var (
+const (
 	Retries    = 13
-	RetryDelay = eventcache.EventRetryTimer + (1 * time.Second)
+	RetryDelay = 2 * time.Second
```
mtardy, line 25: This is a slight change from before where it was 3 seconds. Is this expected?

pkg/option/config.go
```diff
@@ -117,6 +120,11 @@ var (
 
 		// Enable all metrics labels by default
 		MetricsLabelFilter: DefaultLabelFilter(),
+
+		// set default valus for the event cache
+		// mainly used in the case of testing
+		EventCacheNumRetries: 15,
+		EventCacheRetryDelay: 2,
```
mtardy, line 127: nit: you could use the same const here and in the flag default value declaration, not sure it's important.

mtardy (approved):
Thanks! Lgtm! Out of curiosity, what was the reason behind this change?

pkg/jsonchecker/jsonchecker.go
tpapagian: Missed the `+1`. I have made that 3 again. I would say that as this is a test and the test passes, this was not a big deal, but I changed that to be on the safe side.

pkg/option/config.go
tpapagian: Yes makes sense. I fixed that.

tpapagian:
> Out of curiosity, what was the reason behind this change?

There may be cases where the k8s watcher is overloaded and the default settings may not enough. On the other hand, having a large delay may result on waiting too much for getting events (even without pod info). So I believe that this is a good idea to allow users to configure that based on their needs.

PR #2930, opened 2024-09-17
podinfo: add a deleted pod cache

kkourt:
Add a deleted pod cache for pod association. See commits.

### DIFF
--- a/pkg/watcher/watcher_test.go
+++ b/pkg/watcher/watcher_test.go
@@ -1,0 +1,220 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package watcher
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/stretchr/testify/require"
+
+	"github.com/google/uuid"
+	"github.com/sirupsen/logrus"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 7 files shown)


pkg/watcher/watcher_test.go
```diff
@@ -0,0 +1,221 @@
...
+	*testing.T
+	Logger *logrus.Logger
+}
+
+func (tl tlog) Write(p []byte) (n int, err error) {
+	tl.Log(string(p))
+	return len(p), nil
+}
+
+func TestFastK8s(t *testing.T) {
```
mtardy, line 39: nit: it's not super explicit what the test does from its name, maybe we can add some comments here from the commit message or change the name?

mtardy (approved):
That's cool! :) lgtm

pkg/watcher/watcher_test.go
```diff
@@ -0,0 +1,221 @@
...
+	informerFactory := informers.NewSharedInformerFactory(client, 0)
+	podInformer := informerFactory.Core().V1().Pods().Informer()
+	podInformer.AddEventHandler(ts.eventHandler())
+
+	watcher, err := newK8sWatcher(informerFactory)
+	require.Nil(t, err)
+	watcher.Start()
+
+	// Make sure informers are running.
+	//informerFactory.Start(ctx.Done())
```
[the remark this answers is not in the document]
kkourt: It's a leftover, removed.

pkg/watcher/watcher_test.go
kkourt: Added a comment, thanks!

kkourt:
Thanks for the review. Pushed a new version based on the feedback. Merging.

PR #2932, opened 2024-09-18
Improve rthooks docs

kkourt:
Improve rthooks docs.

### DIFF
--- a/docs/content/en/docs/installation/runtime-hooks.md
+++ b/docs/content/en/docs/installation/runtime-hooks.md
@@ -8,92 +8,34 @@ description: "Configure Runtime Hooks"
 See [Tetragon Runtime Hooks]({{< ref "/docs/concepts/runtime-hooks" >}}), for an introduction to
 the topic.
 
-We use `minikube` as the example platform because it supports both `cri-o` and `containerd`. Also,
-at the time of this writing, no images that support this have been released, so we build images
-locally within a checked out repo.
+
+## Install Tetragon with Runtime Hooks
+
+We use `minikube` as the example platform because it supports both `cri-o` and `containerd`, but the
+same steps can be applied in other platforms.
+
+### Setup Helm
 
 ```shell
-make image image-operator image-rthooks
+helm repo add cilium https://helm.cilium.io
+helm repo update
 ```
 
-## CRI-O
+### Setup cluster
 
-### Setup Minikube
+{{< tabpane text=true >}}
+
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/installation/runtime-hooks.md
```diff
@@ -8,99 +8,57 @@ description: "Configure Runtime Hooks"
...
-```shell
-make image image-operator image-rthooks
-```
-
-## CRI-O
+## Install Tetragon with Runtime Hooks
 
-### Setup Minikube
+We use `minikube` as the example platform because it supports both `cri-o` and `containerd`, but the
+same steps can be applied in other platforms.
```
mtardy, line 15: grammar nit

```suggestion
same steps can be applied to other platforms.
```

docs/content/en/docs/installation/runtime-hooks.md
```diff
@@ -132,44 +90,108 @@ minikube cp /tmp/new-config.toml /etc/containerd/config.toml
...
+{"time":"2024-07-01T10:57:21.435689144Z","level":"INFO","msg":"hook request to agent succeeded","hook":"create-container","start-time":"2024-07-01T10:57:21.433755984Z","req-cgroups":"/kubepods/besteffort/podd4e74de2-0db8-4143-ae55-695b2489c727/crio-828977b42e3149b502b31708778d0c057efbce038af80d0882ed3e0cb0ff8796","req-rootdir":"/run/containers/storage/overlay-containers/828977b42e3149b502b31708778d0c057efbce038af80d0882ed3e0cb0ff8796/userdata","req-containerName":"test"}
+```
+
+## Configuring Runtime Hooks installation
+
+### Installation directory (`installDir`)
+
+For tetragon runtime hooks to work, a binary (`tetragon-oci-hook`) needs to be installed on the
+host. This happens by the runtime hooks. This binary is installed in `/opt/tetragon` by default.
```
mtardy, line 149: "This happens by the runtime hooks"

I think a word may be missing there

mtardy (approved):
Thanks! Looks great.

Aside, I'm wondering if we should move "Installation -> Runtime Hooks" under "Installation -> Configuration -> Runtime Hooks"?

docs/content/en/docs/installation/runtime-hooks.md
kkourt: There was not :) rewarded to make it clearer. Thanks!

PR #2934, opened 2024-09-18
helm: Set rthooks.podSecurityContext to empty by default

michi-covalent:
[ upstream commit 255835998a8295f2500afc03d978f5a59b07dc85 ]

Set rthooks.podSecurityContext to empty by default to be consistent with the security context setting of Tetragon pods. Also note that "privileged" setting only applies to container security context [^1][^2].

[^1]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#podsecuritycontext-v1-core
[^2]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#securitycontext-v1-core


michi-covalent:
backporting https://github.com/cilium/tetragon/pull/2931 to v1.2 branch.

kkourt (approved):

PR #2935, opened 2024-09-19
k8s: Pick up github.com/cilium/cilium v1.16.1

michi-covalent:
This new version of Cilium updates k8s.io/code-generator dependency, and generate-groups.sh is no longer supported [^1]. Update pkg/k8s/Makefile to use kube_codegen.sh. One could only assume that this migration is successful since it's generating identical code. k8s-code-gen.sh is copied from cilium/cilium [^3] and modified for Tetragon.

See apiextensions docs [^2] regarding x-kubernetes-list-type field in CRDs.

[^1]: https://github.com/kubernetes/code-generator/commit/2e5be312636d3c77fb42ba37cedd4707968608a0
[^2]: https://pkg.go.dev/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions
[^3]: https://github.com/cilium/cilium/blob/main/contrib/scripts/k8s-code-gen.sh


mtardy:
oh no I already did this work, merged it, reversed it, and retried again recently, hope it didn't take you too much time. This should work now thanks to some fixes I did to controller-tools: https://github.com/cilium/tetragon/pull/2911. I need to rebase soonish.

PR #2937, opened 2024-09-19
Memory optimizations: remove BTF and kallsyms caches

tpapagian:
Do not cache large objects like the BTF and ksyms. Reload them when needed.

Before:
Screenshot 2024-09-19 at 11 09 15 AM

After:
Screenshot 2024-09-19 at 11 31 21 AM

### DIFF
--- a/pkg/ksyms/ksyms.go
+++ b/pkg/ksyms/ksyms.go
@@ -11,7 +11,6 @@ import (
 	"sort"
 	"strconv"
 	"strings"
-	"sync"
 
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/option"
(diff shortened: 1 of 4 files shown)


kkourt (approved):

pkg/ksyms/ksyms.go
```diff
@@ -11,25 +11,15 @@ import (
...
-)
-
 func KernelSymbols() (*Ksyms, error) {
-	var err error
-	setKernelSymbols.Do(func() {
-		kernelSymbols, err = NewKsyms(option.Config.ProcFS)
-	})
-	return kernelSymbols, err
+	return NewKsyms(option.Config.ProcFS)
 }
```
mtardy, line 23: nit: Could we add a comment there? it might be a little bit confusing why we had this function in the first place.

pkg/sensors/load.go
```diff
@@ -124,6 +125,9 @@ func (s *Sensor) Load(bpfDir string) error {
 	progsAdd(s.Progs)
 	AllMaps = append(AllMaps, s.Maps...)
 
+	// cleanup the BTF once we have loaded all sensor's program
+	btf.FlushKernelSpec()
```
mtardy, line 129: 😻 perfect!

pkg/sensors/program/loader.go
```diff
@@ -911,6 +911,11 @@ func doLoadProgram(
 
 	load.Prog = prog
 
+	// in KernelTypes, we use a non-standard BTF which is possibly annotated with symbols
+	// from kernel modules. At this point wewe don't need that anymore, so we can release
```
mtardy, line 915:
```suggestion
	// from kernel modules. At this point we don't need that anymore, so we can release
```

mtardy (approved):
Thanks!! I converged to writing this independently today, thanks for doing it :)

jrfastab (approved):

PR #2943, opened 2024-09-20
tracingpolicy: add BPF operations support

tixxdz:
### Description
Add BPF operations support.

### Changelog

```release-note
tracingpolicy: add BPF operations support 
```


olsajiri:
so IIUC you want the event to show the bpf command numbers ... IMO the types should be just for kernel objects
not for int values interpretation

could we do that without changing the ebpf code? we just need to get the int value from kernel (which we already have support for, no need to add new) and transform that number into bpf command enum value

I think we could factor the printer code to do that.. to have kernel type and user side type.. so you'd configure the rinter and argument to have int type value for kernel code and 'bpf_cmd' type for user space processing code

tixxdz:
> so IIUC you want the event to show the bpf command numbers ... IMO the types should be just for kernel objects not for int values interpretation
> 
> could we do that without changing the ebpf code? we just need to get the int value from kernel (which we already have support for, no need to add new) and transform that number into bpf command enum value
> 
> I think we could factor the printer code to do that.. to have kernel type and user side type.. so you'd configure the rinter and argument to have int type value for kernel code and 'bpf_cmd' type for user space processing code

Yes makes sense, can you please check again Jiri, thank you!

kaworu (approved):
protobuf changes LGTM

olsajiri (approved):

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -6970,3 +6970,67 @@ tetragon_missed_prog_probes_total{attach="wake_up_new_task",policy="__base__"} 0
 		prometheus.BuildFQName(consts.MetricsNamespace, "", "missed_prog_probes_total")))
 
 }
+
+func TestKprobeBpfCmd(t *testing.T) {
```
kkourt, line 6974: So I guess the idea of the test is to detect the tetragon bpf program / map loading?

kkourt (approved):
Thanks, LGTM.

pkg/sensors/tracing/kprobe_test.go
tixxdz: Yes kept it small, but it can be used to detect all bpf commands.

PR #2945, opened 2024-09-22
tetragon: Add map ownership

olsajiri:
Adding the concept of owning map and splitting Map object to owners and others.

The idea is that the 'owned' Map object takes care of creating/overwriting the pin,
while the 'other' Map object is just a user, that can read/update map but can't create it.

I'm not really happy with the 'other' naming, so any ideas are welcome ;-)

### DIFF
--- a/pkg/sensors/program/map.go
+++ b/pkg/sensors/program/map.go
@@ -75,6 +75,11 @@ const (
 	MapTypeProgram
 )
 
+type MapOpts struct {
+	Type  MapType
+	Owner bool
+}
+
 // Map represents BPF maps.
 type Map struct {
 	Name         string
@@ -85,6 +90,7 @@ type Map struct {
 	Entries      MaxEntries
 	InnerEntries MaxEntries
 	Type         MapType
+	Owner        bool
 }
 
 // Map holds pointer to Program object as a source of its ebpf object
@@ -100,8 +106,8 @@ type Map struct {
 //	p.PinMap["map2"] = &map2
 //	...
 //	p.PinMap["mapX"] = &mapX
-func mapBuilder(name string, ty MapType, lds ...*Program) *Map {
-	m := &Map{name, "", lds[0], Idle(), nil, MaxEntries{0, false}, MaxEntries{0, false}, ty}
+func mapBuilder(name string, ty MapType, owner bool, lds ...*Program) *Map {
+	m := &Map{name, "", lds[0], Idle(), nil, MaxEntries{0, false}, MaxEntries{0, false}, ty, owner}
 	for _, ld := range lds {
 		ld.PinMap[name] = m
 	}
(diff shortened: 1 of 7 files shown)


tixxdz:
> Adding the concept of owning map and splitting Map object to owners and others.
> 
> The idea is that the 'owned' Map object takes care of creating/overwriting the pin, while the 'other' Map object is just a user, that can read/update map but can't create it.

What are the possible use cases for it? restart ? or is it just to organize better?
 
> I'm not really happy with the 'other' naming, so any ideas are welcome ;-)

Hmm others could be fine, or "non-owners"?, but those are certainly better than "users" ?   or go with creators and users? ...

olsajiri:
> > Adding the concept of owning map and splitting Map object to owners and others.
> > The idea is that the 'owned' Map object takes care of creating/overwriting the pin, while the 'other' Map object is just a user, that can read/update map but can't create it.
> 
> What are the possible use cases for it? restart ? or is it just to organize better?

so at the moment any user of the map will overwrite the pin file with its spec and max entries,
while in reality we  always have one user that creates the map (owner) and other users that
expect the map to exist and want just to open and use it

it has also another aspect where you need all the users of the map to match the same spec
and same max entries value which is not convenient from configuration point of view, because
all users must have access to the spec and max entries

having the owner and other user force the config  duty only to owner and other users just use the map

> 
> > I'm not really happy with the 'other' naming, so any ideas are welcome ;-)
> 
> Hmm others could be fine, or "non-owners"?, but those are certainly better than "users" ? or go with creators and users? ...

hm, right.. maybe MapBuilder(..) for owner and MapUser(..) (instead MapOther) for other users.. with:

also perhaps the MapOwner should be just bool.. iota is perhaps too much

let's see.. thanks a lot for review

mtardy:
This could be beneficial for accounting which sensor/progs/policies are responsible for BPF map memory use, cc @kkourt

pkg/sensors/program/map.go
```diff
@@ -245,11 +245,12 @@ func isValidSubdir(d string) bool {
...
+func LoadOrCreatePinnedMap(pinPath string, mapSpec *ebpf.MapSpec, create bool) (*ebpf.Map, error) {
+	var err error
 	// Try to open the pinPath and if it exist use the previously
 	// pinned map otherwise pin the map and next user will find
 	// it here.
-	if _, err := os.Stat(pinPath); err == nil {
+	if _, err = os.Stat(pinPath); err == nil {
 		m, err := ebpf.LoadPinnedMap(pinPath, nil)
 		if err != nil {
 			return nil, fmt.Errorf("loading pinned map from path '%s' failed: %w", pinPath, err)
```
tixxdz, line 283: Does it make sense to re-create here if we have create true?

pkg/sensors/program/map.go
```diff
@@ -260,12 +261,23 @@ func LoadOrCreatePinnedMap(pinPath string, mapSpec *ebpf.MapSpec) (*ebpf.Map, er
 				"map-name": mapSpec.Name,
 			}).Warn("tetragon, incompatible map found: will delete and recreate")
```
tixxdz, line 289: Maybe this delete and recreate warn should be moved inside create true?

pkg/sensors/program/map.go
```diff
@@ -260,12 +261,23 @@ func LoadOrCreatePinnedMap(pinPath string, mapSpec *ebpf.MapSpec) (*ebpf.Map, er
...
+			if create {
+				os.Remove(pinPath)
+				return createPinnedMap(pinPath, mapSpec)
+			}
+			return nil, fmt.Errorf("incompatible map path '%s'", pinPath)
 		}
+		return m, nil
+	}
+	if create {
+		return createPinnedMap(pinPath, mapSpec)
```
tixxdz, line 302: Could move this create just after the state(pinPath) failure , and have the rest outside of if , just for readability up to you

tixxdz (approved):
LGTM, minor suggestions with one log message that should be fixed. Thank you!

pkg/sensors/program/map.go
olsajiri: hm, so this leg is when there is already an existing pin and we check if the pin is compatible with what we want to create, if it is, we are done, if not, we remove the pin and create new one

pkg/sensors/program/map.go
olsajiri: good catch! will move, thanks

pkg/sensors/program/map.go
olsajiri: hm, so it is already in the place after the stat failure.. not sure what you mean

pkg/sensors/program/map.go
```diff
@@ -47,6 +47,17 @@
 //      - map is local for program, not shared at all
 //
 //  NOTE Please do not share MapTypeProgram maps, it brings confusion.
+//
+//  Each map declares the ownership of the map. The map can be either
+//  owner of the map (via MapBuilder* helpers) or as an user (MapUser*
+//  helpers.
```
kkourt, line 53: Thanks for adding the comments!

kkourt (approved):
🚀

PR #2948, opened 2024-09-23
bpf: allow all operations for syscall64 type

kkourt:
syscall64 type allows a bit to be set to distinguish between 32- and 64-bit syscalls. Currently, the only operators that work are InMap and NotInMap. This commit extends support for other operations as well: EQ, NEQ, MASK, GT, LT.

Using the MASK operator specifically, allows us to write policies for all 32-bit syscalls.

For example:

```yaml
kind: TracingPolicy
metadata:
  name: syscalls32bit
spec:
  tracepoints:
  - subsystem: raw_syscalls
    event: sys_enter
    args:
    - index: 4
      type: syscall64
    selectors:
    - matchArgs:
      - index: 0
        operator: Mask
        values:
        - "2147483648" # IS_32BIT
```

```release-note
bpf: support all operators (including Mask) for the syscall64 type
```

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1268,14 +1268,18 @@ FUNC_INLINE bool is_signed_type(int type)
 
 // filter on values provided in the selector itself
 FUNC_LOCAL long
-filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
+filter_64ty_selector_val(struct selector_arg_filter *filter, char *args, bool set32bit)
 {
 	__u64 *v = (__u64 *)&filter->value;
 	int i, j = 0;
+	__u64 b32 = 0;
+
+	if (set32bit)
+		b32 |= IS_32BIT;
 
 #pragma unroll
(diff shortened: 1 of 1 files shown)


olsajiri:
it feels like we expose the implementation detail of syscall64 for 32 bit syscalls,
which is perhaps fine unless we want to change it, which might never happen ;-)

also LT,GT,EQ would have to count on the values having the bit set on which seems tricky,
I guess it's ok unless we document it, which you didn't, so it's fine ;-)

olsajiri (approved):

kkourt:
> it feels like we expose the implementation detail of syscall64 for 32 bit syscalls, which is perhaps fine unless we want to change it, which might never happen ;-)
> 
> also LT,GT,EQ would have to count on the values having the bit set on which seems tricky, I guess it's ok unless we document it, which you didn't, so it's fine ;-)

Agreed. I'm not sure if we want to expose the implementation detail of syscall64 but this PR just introduces the mechanism. Once we figure out how we want to properly expose this to users, we can use it. Until then, we can use what is introduced here to experiment. Thanks!

PR #2949, opened 2024-09-24
Add support to exclude valid processes from dump processCache

tpapagian:
### Description

When running ./tetra dump processcache we can get all processes in the user-space process cache. Out of these, we can use --skip-zero-refcnt to exclude processes with refCnt equals to 0. These are processes that will be removed from the processCache soon.

Out of these processes with refCnt not equal to 0, most of them are nornal running processes that also exist in the execve_map.

This patch adds another command line option
--exclude-execve-map-processes that excludes processes that exist in the execve_map.

The remaining processes (if any) have possibly an error in the reference counting.

```release-note
Add --exclude-execve-map-processes in tetra dump processcache command.
```


mtardy (approved):
This is great! :) Thanks

PR #2952, opened 2024-09-25
tetragon: Factor grpc exec events test

olsajiri:
Replace sleeps with waiting that can be interrupted by event delivery in notifier.

Adding bigger timeout to prevent flakes on slow machines.

### DIFF
--- a/pkg/grpc/exec/exec_test_helper.go
+++ b/pkg/grpc/exec/exec_test_helper.go
@@ -59,7 +59,42 @@ var (
 )
 
 type DummyNotifier[EXEC notify.Message, EXIT notify.Message] struct {
-	t *testing.T
+	t  *testing.T
+	ch chan bool
+}
+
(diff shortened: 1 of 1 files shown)


pkg/grpc/exec/exec_test_helper.go
```diff
@@ -415,7 +451,7 @@ func GrpcExecOutOfOrder[EXEC notify.Message, EXIT notify.Message](t *testing.T)
 		AllEvents = append(AllEvents, e)
 	}
 
-	time.Sleep(time.Millisecond * time.Duration((option.Config.EventCacheNumRetries+4)*CacheTimerMs)) // wait for cache to do it's work
+	dn.WaitNotifier(4)
```
tpapagian, line 454: Small question: where does this argument (4) comes from? In other cases it seems to be 1 or 2. Why don't we have the same number of retries in all cases?

olsajiri: it's the number of events we expect from the notifier

tpapagian (approved):
LGTM, thanks!

willfindlay (approved):

PR #2953, opened 2024-09-25
ProcessCache: Remove stale entries

kevsecurity:
Add stale entry removal functionality to the process cache. This won't affect normal operations, but in the event that a custom ref count has been placed on an entry and not removed (for whatever reason, maybe logic error, maybe missed event, etc), remove entries where the process and all its children have exited at least 10 minutes ago. Record a metric when entries are removed.

### DIFF
--- a/pkg/option/flags.go
+++ b/pkg/option/flags.go
@@ -115,6 +115,8 @@ const (
 
 	KeyEventCacheRetries    = "event-cache-retries"
 	KeyEventCacheRetryDelay = "event-cache-retry-delay"
+
+	KeyProcessCacheStaleInterval = "process-cache-stale-interval"
 )
 
 type UsernameMetadaCode int
@@ -244,6 +246,8 @@ func ReadAndSetFlags() error {
 
 	Config.EventCacheNumRetries = viper.GetInt(KeyEventCacheRetries)
 	Config.EventCacheRetryDelay = viper.GetInt(KeyEventCacheRetryDelay)
+
+	Config.ProcessCacheStaleInterval = viper.GetDuration(KeyProcessCacheStaleInterval)
 
 	return nil
 }
@@ -411,4 +415,6 @@ func AddFlags(flags *pflag.FlagSet) {
 
 	flags.Int(KeyEventCacheRetries, defaults.DefaultEventCacheNumRetries, "Number of retries for event cache")
 	flags.Int(KeyEventCacheRetryDelay, defaults.DefaultEventCacheRetryDelay, "Delay in seconds between event cache retries")
+
+	flags.Duration(KeyProcessCacheStaleInterval, defaults.DefaultProcessCacheStaleInterval, "Interval between stale process cache checks")
 }
--- a/pkg/process/cache.go
+++ b/pkg/process/cache.go
@@ -13,16 +13,18 @@ import (
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/defaults"
 	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/sensors/exec/execvemap"
 	lru "github.com/hashicorp/golang-lru/v2"
 	"google.golang.org/protobuf/types/known/wrapperspb"
 )
 
 type Cache struct {
-	cache      *lru.Cache[string, *ProcessInternal]
-	size       int
-	deleteChan chan *ProcessInternal
-	stopChan   chan bool
+	cache                 *lru.Cache[string, *ProcessInternal]
+	size                  int
+	deleteChan            chan *ProcessInternal
+	stopChan              chan bool
+	staleIntervalLastTime time.Time
 }
 
 // garbage collection states
@@ -40,9 +42,12 @@ var colorStr = map[int]string{
 	deleted:       "deleted",
 }
 
-// garbage collection run interval
 const (
+	// garbage collection run interval
 	intervalGC = time.Second * 30
+	// garbage collection stale entry max time.
+	// If a stale entry is older than the max time, then it will be removed.
+	staleIntervalMaxTime = time.Minute * 10
 )
 
 func (pc *Cache) cacheGarbageCollector() {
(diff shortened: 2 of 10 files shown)


mtardy:
> Record a metric when entries are removed.

Thanks, I was thinking about adding this. Thinking about this I also realize that it would make issues difficult to debug in the future because we'll struggle to see the actual events that were flushed, should we add an option to append those events to a (circular) debug file?

kkourt:
> > Record a metric when entries are removed.
> 
> Thanks, I was thinking about adding this. Thinking about this I also realize that it would make issues difficult to debug in the future because we'll struggle to see the actual events that were flushed, should we add an option to append those events to a (circular) debug file?

I agree. If we are doing this, it would also be nice to have a switch to configure the parameters that right now are hardcoded.

mtardy:
> > > Record a metric when entries are removed.
> > 
> > 
> > Thanks, I was thinking about adding this. Thinking about this I also realize that it would make issues difficult to debug in the future because we'll struggle to see the actual events that were flushed, should we add an option to append those events to a (circular) debug file?
> 
> I agree. If we are doing this, it would also be nice to have a switch to configure the parameters that right now are hardcoded.

we had this discussion offline and maybe having a switch to disable this could be enough to investigate and try to reproduce the cases where we see the metric goes out of the roof.

> it would also be nice to have a switch to configure the parameters that right now are hardcoded.

maybe you can just use the special GC interval value to 0 to disable it entirely so that you have a two in one option :D

kevsecurity:
> I agree. If we are doing this, it would also be nice to have a switch to configure the parameters that right now are hardcoded.

And also make the existing intervalGC value configurable? It's been hard-coded for ever.

pkg/process/cache.go
```diff
@@ -40,9 +41,15 @@ var colorStr = map[int]string{
...
-// garbage collection run interval
 const (
+	// garbage collection run interval
 	intervalGC = time.Second * 30
+	// garbage collection stale entry check interval.
+	// We will check for stale entries once per interval.
+	staleIntervalGC = time.Hour * 1
+	// garbage collection stale entry max time.
+	// If a stale entry is older than the max time, then it will be removed.
+	staleIntervalMaxTime = time.Minute * 10
```
mtardy, line 52: > > I agree. If we are doing this, it would also be nice to have a switch to configure the parameters that right now are hardcoded.
> 
> And also make the existing intervalGC value configurable? It's been hard-coded for ever.

I think we meant those ones. Or at least the first one `staleIntervalGC`

docs/content/en/docs/use-cases/network-observability/_index.md
mtardy, line 1: not sure you intended to commit this one :D

docs/content/en/docs/use-cases/network-observability/approved-dns-servers.md
mtardy, line 1: or this one!

pkg/option/flags.go
```diff
@@ -245,6 +247,8 @@ func ReadAndSetFlags() error {
 	Config.EventCacheNumRetries = viper.GetInt(KeyEventCacheRetries)
 	Config.EventCacheRetryDelay = viper.GetInt(KeyEventCacheRetryDelay)
 
+	Config.ProcessCacheStaleInterval = viper.GetUint(KeyProcessCacheStaleInterval)
```
mtardy, line 250: Could you use `GetDuration`, that takes care of all the seconds minutes whatever https://pkg.go.dev/github.com/spf13/viper#GetDuration?

pkg/option/flags.go
```diff
@@ -411,4 +415,6 @@ func AddFlags(flags *pflag.FlagSet) {
 
 	flags.Int(KeyEventCacheRetries, defaults.DefaultEventCacheNumRetries, "Number of retries for event cache")
 	flags.Int(KeyEventCacheRetryDelay, defaults.DefaultEventCacheRetryDelay, "Delay in seconds between event cache retries")
+
+	flags.Uint(KeyProcessCacheStaleInterval, defaults.DefaultProcessCacheStaleInterval, "Interval in minutes between stale process cache checks")
```
mtardy, line 419: same you can use duration and default to what's needed :)

pkg/process/cache.go
```diff
@@ -117,6 +128,37 @@ func (pc *Cache) cacheGarbageCollector() {
...
+	}
+}
+
+// Must be called with a lock held on p.refcntOpsLock
+func processAndChildrenHaveExited(p *ProcessInternal) bool {
+	if p.refcntOps["process++"] == p.refcntOps["process--"] &&
+		p.refcntOps["parent++"] == p.refcntOps["parent--"] {
+		return true
+	}
+	return false
```
mtardy, line 156: you could simplify this with the following:
```go
return p.refcntOps["process++"] == p.refcntOps["process--"] && p.refcntOps["parent++"] == p.refcntOps["parent--"]
```
curious that go simple isn't suggesting this 🤔

pkg/process/cache.go
```diff
@@ -117,6 +128,37 @@ func (pc *Cache) cacheGarbageCollector() {
...
+}
+
+func processOrChildExit(reason string) bool {
+	if reason == "process--" || reason == "parent--" {
+		return true
+	}
+	return false
```
mtardy, line 163: you can also simplify a little bit here
```go
return reason == "process--" || reason == "parent--"
```

pkg/process/cache.go
```diff
@@ -159,6 +163,13 @@ func processOrChildExit(reason string) bool {
 	return false
 }
 
+func processOrChildExec(reason string) bool {
+	if reason == "process++" || reason == "parent++" {
+		return true
+	}
+	return false
```
mtardy, line 170: same here

mtardy (commented):
looks good even though this thing looks a bit complicated, especially the second commit 😵 . Here are a few comments

docs/content/en/docs/use-cases/network-observability/_index.md
kevsecurity: Ah, they were in my tree. I'll get rid.

pkg/option/flags.go
kevsecurity: TBH I was just following how intervals were done for other switches, but yes that is a good idea!

pkg/process/cache.go
kevsecurity: Yep

pkg/process/cache.go
kevsecurity: Yep

mtardy (approved):
Thanks, that's great! It's a bit hard to review manually but thankfully you added tests so it should be fine :)!! Just a small fix and I think we can merge.

pkg/process/cache.go
```diff
@@ -92,6 +97,12 @@ func (pc *Cache) cacheGarbageCollector() {
 					}
 				}
 				deleteQueue = newQueue
+				// Check if it is time to clean stale entries.
+				if option.Config.ProcessCacheStaleInterval > 0 &&
+					pc.staleIntervalLastTime.Add(time.Duration(option.Config.ProcessCacheStaleInterval)*time.Minute).Before(time.Now()) {
```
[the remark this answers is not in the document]
kevsecurity: Probably. I will fix it now.

mtardy (approved):

kevsecurity:
Converted to draft. We may not need the second commit.

PR #2959, opened 2024-09-27
tetra: add a new debug maps command and move dump under debug

mtardy:
Here is the `help` for this new command:
```
Retrieve information about maps on the host related to tetragon.

- AllMaps: all the BPF maps loaded on the host
- PinnedProgsMaps: all the BPF maps used in programs pinned under /sys/fs/bpf/tetragon
- PinnedMaps: all the BPF maps pinned under /sys/fs/bpf/tetragon

The first table is the total of memlock bytes for the maps, those values are
accurate mostly from v6.4[^1].

The second table is in number of maps, inter, exter and union are the respective
math operations on the two first set, diff is the number of maps that were
detected as pinned maps but not used in pinned progs, ideally this number should
be 0. Note that the union should be the set of all the tetragon related maps on
the host if all the programs are properly pinned under the bpffs and all the
unused maps are also pinned under the bpffs.

The third table is the details of the maps that are in the diff, which means
maps that are pinned but unreferenced from pinned programs. Ideally, this table
should be empty.

The fourth table aggregate map by name and add the count column indicating how
many maps were aggregated for the given name. It ranks map from the one
consuming the most memory to the one consuming the less. Use the -n flag to
adjust the number of item in the table.

[^1]: https://lore.kernel.org/all/20230305124615.12358-1-laoar.shao@gmail.com/
```

Here is a typical output for tetragon loading with a TCP tracing policy:
```
AllMaps    PinnedProgsMaps   PinnedMaps
28981232   28972720          27344136

PinnedProgsMaps   PinnedMaps   Inter   Exter   Union   Diff
132               83           83      49      132     0

ID   Name   Type   KeySize   ValueSize   MaxEntries   Memlock

Name              Type          KeySize   ValueSize   MaxEntries   Count   TotalMemlock   PercentOfTotal
execve_map        Hash          4         632         32768        1       23073936       79.6%
socktrack_map     LRUHash       8         16          32000        1       2829376        9.8%
data_heap         PerCPUArray   4         32768       1            4       787744         2.7%
tg_execve_joine   PerCPUArray   4         8           2            2       722416         2.5%
tg_mbset_map      Hash          256       8           1024         1       338880         1.2%
string_maps_ro_   PerCPUArray   4         16384       1            3       295896         1.0%
string_maps_hea   PerCPUArray   4         16384       1            3       295896         1.0%
process_call_he   PerCPUArray   4         24312       1            1       146200         0.5%
retprobe_map      Hash          16        24          1024         1       108160         0.4%
buffer_heap_map   PerCPUArray   4         4352        1            4       105760         0.4%
```


cmd/tetra/debug/maps.go
```diff
@@ -0,0 +1,235 @@
...
+
+const tetragonBPFFS = "/sys/fs/bpf/tetragon"
+
+func NewMapCmd() *cobra.Command {
+	var lines int
+
+	cmd := cobra.Command{
+		Use:     "maps",
+		Aliases: []string{"map"},
+		Short:   "Retrieve information about maps on the host related to tetragon",
```
kkourt, line 26: s/maps/BPF maps/ ?

kkourt (approved):
Thanks!

This looks really, useful!

Some questions; Is this part of bugtool? Should we make it?

Can we have JSON output? JSON output helps if we want to write scripts to parse the data which will be useful if we run this via sysdmp/bugtool across the whole cluster.

cmd/tetra/debug/maps.go
```diff
@@ -0,0 +1,235 @@
...
+				len(inter(pinnedMapsSet, pinnedProgsMapsSet)),
+				len(exter(pinnedMapsSet, pinnedProgsMapsSet)),
+				len(union(pinnedMapsSet, pinnedProgsMapsSet)),
+				len(diff(pinnedMapsSet, pinnedProgsMapsSet)),
+			)
+			w.Flush()
+			cmd.Println()
+
+			// print details on the diff
+			w = tabwriter.NewWriter(cmd.OutOrStdout(), 0, 0, 3, ' ', 0)
```
olsajiri, line 120: I'm getting empty diff with just column output which is confusing,
```
PinnedProgsMaps   PinnedMaps   Inter   Exter   Union   Diff
62                31           31      31      62      0

ID   Name   Type   KeySize   ValueSize   MaxEntries   Memlock

Name              Type          KeySize   ValueSize   MaxEntries   Count   TotalMemlock   PercentOfTotal
execve_map        Hash          4         632         32768        1       23072416       83.6%
```

could we just say there's no diff or don't print those headers at all?

pkg/bugtool/maps.go
```diff
@@ -0,0 +1,355 @@
...
+		}
+		prog, err := ebpf.LoadPinnedProgram(path, &ebpf.LoadPinOptions{
+			ReadOnly: true,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to load pinned object %q: %w", path, err)
+		}
+		defer prog.Close()
+
+		if ok, err := isProg(prog.FD()); err != nil || !ok {
```
olsajiri, line 213: fyi I was using filename check for 'prog' or 'prog_override' to identify programs, your's is more resilient

cmd/tetra/debug/maps.go
```diff
@@ -0,0 +1,235 @@
...
+			w := tabwriter.NewWriter(cmd.OutOrStdout(), 0, 0, 3, ' ', 0)
+			fmt.Fprintln(w, "AllMaps\tPinnedProgsMaps\tPinnedMaps")
+			fmt.Fprintf(w, "%d\t%d\t%d\n", allMapsMem, pinnedProgsMapsMem, pinnedMapsMem)
+			w.Flush()
+			cmd.Println()
+
+			// print details on map distribution
+			pinnedProgsMapsSet := map[int]bugtool.ExtendedMapInfo{}
+			pinedProgsMaps, err := bugtool.FindMapsUsedByPinnedProgs(tetragonBPFFS)
+			if err != nil {
```
olsajiri, line 82: we already iterated that as part of TotalByteMemlockMapsFromProgs right? same for FindPinnedMaps and TotalByteMemlockMaps? could we perhaps run that just once at beginning and use the []ExtendedMapInfo for both memlock sum and printing later?

olsajiri (approved):
looks great, left few comments, thanks

mtardy:
Thanks for the reviews

> Some questions; Is this part of bugtool? Should we make it?

It's not currently as most commands parts of `dump` and I think it should.

> Can we have JSON output? JSON output helps if we want to write scripts to parse the data which will be useful if we run this via sysdmp/bugtool across the whole cluster.

agree

cmd/tetra/debug/maps.go
mtardy: agree but I then wonder how to describe the content of this output since I rely on first, second, third, etc table to describe it. Maybe printing there's no diff then in place of the table 🤔

pkg/bugtool/maps.go
mtardy: ok thanks didn't know! I think I'll integrate that into c/ebpf eventually https://github.com/cilium/ebpf/issues/1566.

cmd/tetra/debug/maps.go
mtardy: yep indeed the `Total`s are pretty much wrappers around the other function + sum. It could be optimized and only done once at the beginning and the sum computed locally with the results. Maybe I should remove the `Total` functions

mtardy: fixed it by removing them

cmd/tetra/debug/maps.go
mtardy: it will now print "Empty diff table" and "Empty BPF memory consumption table" in place of the two last tables if they are empty.

mtardy:
> Thanks for the reviews
> 
> > Some questions; Is this part of bugtool? Should we make it?
> 
> It's not currently as most commands parts of `dump` and I think it should.
> 
> > Can we have JSON output? JSON output helps if we want to write scripts to parse the data which will be useful if we run this via sysdmp/bugtool across the whole cluster.
> 
> agree

I will merge that and add that in a follow-up PR.

PR #2961, opened 2024-09-30
Implemented a possibility to disable gops via helm

XelK:
Fixes 

### Description

Implemented the possibility to disable gops server when installing tetragon via helm

### Changelog

```release-note
Add an "enabled" switch to enable/disable the gops server via the Helm chart. It is now disabled by default.
```

### DIFF
--- a/install/kubernetes/tetragon/values.yaml
+++ b/install/kubernetes/tetragon/values.yaml
@@ -175,6 +175,8 @@ tetragon:
     # -- The address at which to expose gRPC. Examples: localhost:54321, unix:///var/run/cilum/tetragon/tetragon.sock
     address: "localhost:54321"
   gops:
+    # -- Whether to enable exposing gops server.
+    enabled: true
     # -- The address at which to expose gops.
     address: "localhost"
     # -- The port at which to expose gops.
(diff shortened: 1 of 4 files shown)


mtardy:
Thanks for the patch following the discussion we had on Slack :)! However, it looks like your commit is not signed-off, to improve tracking of who did what, the Tetragon project uses a "sign-off" procedure which you can learn more about in the [Cilium project documentation](https://docs.cilium.io/en/latest/contributing/development/contributing_guide/#developer-s-certificate-of-origin).

<p>

```shell
```

```shell
git commit -s
```

```shell
git commit --amend -s
```

```
-s, --signoff, --no-signoff

   option on the command line.
```

</p>
</details>

mtardy:
Hey, please take a look at the CI output to update your patch and please in the end squash your commits into only one :)! Thanks

install/kubernetes/tetragon/values.yaml
```diff
@@ -175,6 +175,8 @@ tetragon:
     # -- The address at which to expose gRPC. Examples: localhost:54321, unix:///var/run/cilum/tetragon/tetragon.sock
     address: "localhost:54321"
   gops:
+     # -- Whether to enable exposing gops server.
+    enabled: false
```
mtardy, line 179: It will be user-facing change however, I imagine that the previous situation was making this enabled by default. However I think disabling it by default might be correct indeed!

mtardy (approved):
Thanks a lot for the work, if you want we could maybe disable that by default, thus we would need small changes in the e2e test framework.

PR #2962, opened 2024-09-30
Update Helm CRDs whenever code CRDs are updated

PhilipSchmid:
```release-note
Update Helm CRDs whenever code CRDs are updated
```


mtardy:
by the way, why do we need the json schema checked out in the git repo? can't we just generate them on the fly for kubeconform to run?

PhilipSchmid:
Superseded by https://github.com/cilium/tetragon/pull/2971

PR #2968, opened 2024-10-01
tetragon: sensor cleanup fixes

olsajiri:
do better job on cleaning up after sensors is unloaded or fails to load

### DIFF
--- a/pkg/sensors/load.go
+++ b/pkg/sensors/load.go
@@ -117,9 +117,20 @@ func (s *Sensor) Load(bpfDir string) (err error) {
 		return fmt.Errorf("tetragon, aborting minimum requirements not met: %w", err)
 	}
 
+	var (
+		loadedMaps  []*program.Map
+		loadedProgs []*program.Program
+	)
+
 	s.createDirs(bpfDir)
 	defer func() {
 		if err != nil {
+			for _, m := range loadedMaps {
+				m.Unload()
+			}
+			for _, p := range loadedProgs {
+				unloadProgram(p)
+			}
 			s.removeDirs()
 		}
 	}()
@@ -142,6 +153,7 @@ func (s *Sensor) Load(bpfDir string) (err error) {
 		if err = s.loadMap(bpfDir, m); err != nil {
 			return fmt.Errorf("tetragon, aborting could not load sensor BPF maps: %w", err)
 		}
+		loadedMaps = append(loadedMaps, m)
 	}
 
 	for _, p := range s.Progs {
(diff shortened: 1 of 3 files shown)


kkourt (approved):
Thanks, this looks good and I'm happy to merge it as is.

One thought I had was that maybe we would want to add a waning if the last sensor did not manage to remove the policy directory. So maybe something like:

```diff
diff --git a/pkg/sensors/collection.go b/pkg/sensors/collection.go
index 979517b61..6e4f8b2ba 100644
--- a/pkg/sensors/collection.go
+++ b/pkg/sensors/collection.go
@@ -5,9 +5,12 @@ package sensors
 
 import (
        "fmt"
+       "os"
+       "path/filepath"
        "sync"
 
        "github.com/cilium/tetragon/api/v1/tetragon"
+       "github.com/cilium/tetragon/pkg/option"
        "github.com/cilium/tetragon/pkg/tracingpolicy"
        "go.uber.org/multierr"
 )
@@ -134,6 +137,15 @@ func (c *collection) unload() error {
        if err != nil {
                return fmt.Errorf("failed to unload all sensors from collection %s: %w", c.name, err)
        }
+
+       if tp := c.tracingpolicy; tp != nil {
+               tpPath := filepath.Join(option.Config.BpfDir, sanitize(tp.TpName()))
+               _, err := os.Stat(path)
+               if os.IsNotExist(err) {
+                       return fmt.Erorrf("failed to remove policy direcotry: %s", tpPath)
+               }
+       }
+
        return nil
 }
 ```
 
 As I was writing the above, however, I realized that we might have a more significant problem. Two policies might have the same name if they belong to different k8s namespaces. See for example https://github.com/cilium/tetragon/pull/2337.
 
So I think we need something like:
```diff
diff --git a/pkg/sensors/tracing/policyhandler.go b/pkg/sensors/tracing/policyhandler.go
index 16603d3ab..a799a975b 100644
--- a/pkg/sensors/tracing/policyhandler.go
+++ b/pkg/sensors/tracing/policyhandler.go
@@ -25,6 +25,10 @@ func (h policyHandler) PolicyHandler(
 ) (sensors.SensorIface, error) {
 
        policyName := policy.TpName()
+       if tpNs, ok := tp.(tracingpolicy.TracingPolicyNamespaced); ok {
+               policyName = fmt.Printf("%s/%s", tpNs, policyName)
+       }
+
        spec := policy.TpSpec()
 ```

olsajiri:
yea, we need to fix that.. let's address that in separate PR, thanks

PR #2977, opened 2024-10-04
helm: Added dedicated flag for persistent enforcement

PhilipSchmid:
I added `tetragon.enableKeepSensorsOnExit` to enable persistent enforcement of the sensors even when the Tetragon process is gone/exited. Let's make this flag (`--keep-sensors-on-exit`) a proper Helm value instead of having users specify it via extraArgs. We _should_ also set `--release-pinned-bpf=false` in case `--keep-sensors-on-exit` gets enabled. See https://github.com/cilium/tetragon/commit/64d620e4278ca59fc8666d60daf40b63600943fa.

https://tetragon.io/docs/concepts/enforcement/persistent-enforcement/

```release-note
helm: Added dedicated persistent enforcement flag
```


olsajiri (approved):
lgtm, thanks

PR #2984, opened 2024-10-07
Expose BPF map kernel memory use by tracing policy

mtardy:
This allows users to understand how much kernel memory the BPF maps loaded by the program loaded by their TracingPolicy consume. This information can be retrieved from the gRPC API, by using `tetra` for example, or from the metrics server.

**gRPC API with `tetra`**

```shell-session
$ tetra tp list
ID   NAME                 STATE     FILTERID   NAMESPACE   SENSORS          KERNELMEMORY
1    file-monitoring      enabled   0          (global)    generic_kprobe   5.20 MB
2    sys-symlink-passwd   enabled   0          (global)    generic_kprobe   3.54 MB
3    accept               enabled   0          (global)    generic_kprobe   5.87 MB
4    connect              enabled   0          (global)    generic_kprobe   4.46 MB
```

**Metrics server using `curl`**

```shell-session
$ curl localhost:2112/metrics | grep kernel_memory
# HELP tetragon_tracingpolicy_kernel_memory_bytes The amount of kernel memory in bytes used by policy's sensors non-shared BPF maps (memlock).
# TYPE tetragon_tracingpolicy_kernel_memory_bytes gauge
tetragon_tracingpolicy_kernel_memory_bytes{policy="accept"} 5.865864e+06
tetragon_tracingpolicy_kernel_memory_bytes{policy="connect"} 4.458824e+06
tetragon_tracingpolicy_kernel_memory_bytes{policy="file-monitoring"} 5.197112e+06
tetragon_tracingpolicy_kernel_memory_bytes{policy="sys-symlink-passwd"} 3.543976e+06
```

Note: this PR looks a little bit like https://github.com/cilium/tetragon/pull/2090.

```release-note
Expose BPF map kernel memory use by tracing policy via the gRPC API and the metrics. Use `tetra tp list` to see the breakdown of BPF map memory use by policy or look for the `tetragon_tracingpolicy_kernel_memory_bytes` metric.
```


pkg/sensors/program/map.go
```diff
@@ -104,6 +104,10 @@ type Map struct {
 	Owner        bool
 }
 
+// GlobalMaps keeps a record of all global maps to exclude them from per policy
+// memory map accounting.
+var GlobalMaps map[string]bool
```
mtardy, line 109: should I put this behind a lock or not?

cmd/tetra/common/utils.go
```diff
@@ -0,0 +1,23 @@
...
+// Copyright Authors of Tetragon
+
+package common
+
+import "fmt"
+
+// HumanizeByteCount transforms bytes count into a quickly-readable version, for
+// example it transforms 4458824 into "4.46 MB". I copied this code from
+// https://yourbasic.org/golang/formatting-byte-size-to-human-readable-format/
+func HumanizeByteCount(b int) string {
```
kkourt, line 11: nit: maybe add some simple tests for this function?

pkg/sensors/program/loader.go
```diff
@@ -857,6 +857,50 @@ func doLoadProgram(
...
+		if !available {
+			logger.GetLogger().WithField("map", m.String()).Warn("failed to retrieve BPF map ID, you might be running <4.13")
+		}
+		collMaps[id] = m
+	}
+	load.LoadedMapsInfo = map[int]bpf.ExtendedMapInfo{}
+	for _, p := range coll.Programs {
+		i, err := p.Info()
+		if err != nil {
+			logger.GetLogger().WithError(err).WithField("program", p.String()).Warn("failed to retrieve BPF program info, you might be running <4.10")
```
kkourt, line 886: should we break here? ditto for the other error checks where we only emit a log message.

pkg/sensors/manager_test.go
```diff
@@ -419,3 +420,42 @@ func TestPolicyListingWhileLoadUnload(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, 0, len(l.Policies))
 }
+
+func TestPolicyKernelMemoryBytes(t *testing.T) {
```
kkourt, line 424: ❤️

pkg/sensors/program/map.go
kkourt: yes, please. And can we hide the global object and have functions to access it from other packages?


PS: I wonder if there is a better solution that keeping a global map.
PPS: Don't we need to delete maps from the map? I guess currently all global maps are part of the base sensor so it shouldn't matter, but maybe in the future we want global maps that are added/deleted at runtime (?)

kkourt (changes requested):
Thanks, this looks great!

pkg/sensors/program/loader.go
mtardy: yes good idea, let's give up there's almost no chance if can only fail one time.

pkg/sensors/program/map.go
mtardy: oh yeah it's better behind locks, and indeed I needed to add delete, I moved the logic to add in Load instead of at creation and delete at Unload! :) Please take a look again at that commit!

mtardy: > PS: I wonder if there is a better solution that keeping a global map.

On that I don't know :/ we don't have a central object that keeps references on maps so I guess that might be the only way?

kkourt (approved):

pkg/metrics/policymetrics/policymetrics.go
```diff
@@ -63,6 +70,7 @@ func collect(ch chan<- prometheus.Metric) {
 	for _, policy := range list.Policies {
 		state := policy.State
 		counters[state]++
+		ch <- policyKernelMemory.MustMetric(float64(policy.KernelMemoryBytes), policy.Name)
```
lambdanis, line 73: Does it make sense to report this metric for policies that are disabled or in error state?

mtardy: mmh I haven't tried, hopefully it should show up as zero if it's disabled since progs are unloaded and similarly for error 🤔

mtardy: Oh I tried and the memory state is actually not refreshed properly on disabled, it needs a fix, thanks for the remark.

PR #2994, opened 2024-10-11
enforcer: track missed notifications

kkourt:
```release-note
enforcer: add  `tetragon_enforcer_missed_notifications_total` metric
```


kkourt:
Moving to draft to figure out the CI failures.

pkg/sensors/sensors.go
```diff
@@ -241,3 +257,18 @@ func progsCleanup() {
 func AllPrograms() []*program.Program {
 	return append([]*program.Program{}, allPrograms...)
 }
+
+// sortSensors sort the sensors to enforce orderging constrains
+func sortSensors(sensors []SensorIface) {
+	sort.Slice(sensors, func(i, j int) bool {
```
olsajiri, line 263: seems ok, but I was wondering we want something more generic, like adding sort priority for the sensor.. could be follow up

pkg/sensors/sensors.go
```diff
@@ -69,6 +71,20 @@ type Sensor struct {
...
+		s.PostUnloadHook = hook
+		return
+	}
+
+	oldUnloadHook := s.PostUnloadHook
+	s.PostUnloadHook = func() error {
+		err1 := oldUnloadHook()
+		err2 := hook()
+		return errors.Join(err1, err2)
+	}
```
olsajiri, line 85: that's real nice way to hook the unload actions

olsajiri (approved):
lgtm, thanks

pkg/sensors/sensors.go
kkourt: Agreed. I think eventually we would want something more generic. I'm not sure what it would be though. Having a number seems hard to do.

Issue #2966, opened 2024-10-01
standalone-tarball-tests is flaky 


mtardy:
See https://github.com/cilium/tetragon/actions/runs/11123370825/job/30906850837?pr=2955.

Sometimes we see:
```
sudo tetra status
  sudo tetra tracingpolicy list | grep bpf -
  sudo tetra bugtool
  test $(stat -c %a /var/run/tetragon/tetragon.sock) -eq "[6](https://github.com/cilium/tetragon/actions/runs/11123370825/job/30906850837?pr=2955#step:10:6)60"
  sudo tetra tracingpolicy add examples/tracingpolicy/tcp-connect.yaml
  sudo tetra tracingpolicy list | grep connect -
  sudo grep "tetra" /var/log/tetragon/tetragon.log
  shell: /usr/bin/bash -e {0}
Health Status: running
Error: Process completed with exit code 1.
```

<img width="889" alt="image" src="https://github.com/user-attachments/assets/23aaea08-1ebb-4aab-86bd-04f52b399fcd">

Rerunning usually seems fine, but it would be better to have a reliable test.

mtardy:
Another more recent occurrence of this https://github.com/cilium/tetragon/actions/runs/11269374256/job/31338001835?pr=2985 if that can help,

PR #3001, opened 2024-10-14
ci:github: retry Test Tetragon on failure

tixxdz:
closes: #2966


tixxdz:
> sure I imagine we can merge this but it doesn't feel very satisfying as it's not fixing the underlying issue just automating the retries.
> 
> Since in this case the commands are simple enough, maybe we should understand which commands returned != 0 and why?

It depends since we parse logs and json events, ion tetragon startup logic, how long it takes..., or when loading policies dynamically how long it takes to add kprobes and have them on the output, so we can try to be smart, but I don't think it really worth it, or it matters since we just retry to see if events showed up or logs were successful or not.

PR #3002, opened 2024-10-15
tetragon: Setup tailcalls directly in bpf programs

olsajiri:
This is more straigt forward than current way (which still stays supported).
It also allows to override execve sensor functionality for projects that depend on tetragon.


kkourt (approved):
LGTM, thanks!

PR #3006, opened 2024-10-16
Fix metrics collection in case of duplicate policy names

lambdanis:
This fixes a metrics collection error that was occurring in case there were
multiple namespaced policies with the same name. The metrics collector is
iterating over all policies and reports tetragon_tracingpolicy_kernel_memory_bytes
metric for each of them. The metric used to be labeled only with the policy
name, so in case there was a second policy with the same name, the collector
reported a duplicate what causes the entire metrics collection job to fail.
Adding a policy_namespace label should prevent duplicates.

There is an underlying assumption that each policy returned by
ListTracingPolicies has a unique (name, namespace) tuple - if it's not the
case, then metrics collection might fail again. Also, I'm not sure if
per-policy memory metrics should be reported for policies in states different
than loaded. Currently they are - I didn't change that behaviour, but it's now
covered by tests.

Fixes: a5301b86f036 ("pkg/metrics: add metrics for policy kernel memory us")

### DIFF
--- a/pkg/testutils/sensors/sensors.go
+++ b/pkg/testutils/sensors/sensors.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/cilium/tetragon/pkg/bpf"
 	"github.com/cilium/tetragon/pkg/observer"
+	"github.com/cilium/tetragon/pkg/policyfilter"
 	"github.com/cilium/tetragon/pkg/sensors"
 )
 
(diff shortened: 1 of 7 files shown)


pkg/testutils/sensors/sensors.go
```diff
@@ -44,21 +44,41 @@ type TestSensorManager struct {
...
 func GetTestSensorManager(ctx context.Context, t *testing.T) *TestSensorManager {
-	if mgr := observer.GetSensorManager(); mgr != nil {
+	return getTestSensorManager(ctx, t, false)
+}
+
+func GetTestSensorManagerWithDummyPF(ctx context.Context, t *testing.T) *TestSensorManager {
+	return getTestSensorManager(ctx, t, true)
+}
+
+func getTestSensorManager(ctx context.Context, t *testing.T, useDummyPF bool) *TestSensorManager {
```
kkourt, line 54: Instead of a `bool`, I think it would be better to pass the interface implementation. WDYT?

kkourt (approved):
LGTM, thanks! Left a minor comment which would be great if fixed before merging.

willfindlay (approved):

pkg/testutils/sensors/sensors.go
lambdanis: Thanks, I changed it to pass `policyfilter.State` instead

mtardy:
oh thanks a lot!

PR #3012, opened 2024-10-18
renovate: add v1.2

kkourt:


mtardy (approved):
oh thanks, we forgot that

PR #3021, opened 2024-10-21
pkg/sensors: fix memory use of unloaded sensors

### LINKED ISSUE #3013: Disabled policies should update the tracing policy kernel memory consumption
Thanks to Anna for helping me see this issue.

_Originally posted by @lambdanis in https://github.com/cilium/tetragon/pull/2984#discussion_r1803875560_

mtardy:
Fixes #3013 

Sensors that are disabled are not loaded and should not use memory, thus we clear the LoadedMapsInfo on unloading programs.

### DIFF
--- a/pkg/sensors/program/program.go
+++ b/pkg/sensors/program/program.go
@@ -206,6 +206,11 @@ func (p *Program) Unload() error {
 	}
 	p.unloader = nil
 	p.unloaderOverride = nil
+	// The above unloader can succeed while not removing a pin to the program
+	// because of option.Config.KeepSensorsOnExit, and thus the maps remain.
+	if !p.Prog.IsPinned() {
+		p.LoadedMapsInfo = nil
+	}
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


mtardy:
Regarding the metrics talked here https://github.com/cilium/tetragon/pull/2984#discussion_r1803875560, I think it still makes sense to report 0 for a disabled policy.

pkg/sensors/program/program.go
```diff
@@ -206,6 +206,7 @@ func (p *Program) Unload() error {
 	}
 	p.unloader = nil
 	p.unloaderOverride = nil
+	p.LoadedMapsInfo = nil
```
mtardy, line 209: mmh just wait before merging, if the pinned is still there, this might be wrong to do that, I'll should check if ref are to 0 before considering that the maps have been removed.

mtardy: it should be good now, I just noticed that `--keep-sensors-on-exit` break disabling policies, but that's another issue.

lambdanis (approved):

PR #3025, opened 2024-10-22
Add an optional cluster_name field to GetEventsResponse

michi-covalent:
Add an optional cluster_name field to GetEventsResponse. To enable the field, set tetragon.clusterName Helm value.


jrfastab (approved):

install/kubernetes/tetragon/values.yaml
nvzflow, line 1: i wonder if some default name would be better than empty string. "default-cluster-name" or better. With empty string we also will trigger warning that cluster name was not provided so we cant differentiate pods of same id running in different clusters, like development and production etc.

nvzflow (approved):
looks good to me, just one comment regarding default name , otherwise all is good. Hopefully users will provide us cluster names, eventually they will.

install/kubernetes/tetragon/values.yaml
michi-covalent: what warning will it trigger? having default cluster name won't help differentiating events from different clusters right? 💭

PR #3031, opened 2024-10-22
workflows: simplify build image CI

mtardy:
Follow security best practices for this sensitive pull_request_target workflow and simplify it for easy maintenance.

### DIFF
--- a/.github/workflows/build-images-ci.yml
+++ b/.github/workflows/build-images-ci.yml
@@ -22,54 +22,40 @@ permissions:
   contents: read
 
 jobs:
-  build-and-push-prs:
+  build-and-push:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
         include:
           - name: tetragon
             dockerfile: ./Dockerfile
-
           - name: tetragon-operator
             dockerfile: ./Dockerfile.operator
-
           - name: tetragon-rthooks
             dockerfile: ./Dockerfile.rthooks
+    env:
+      IMAGE: quay.io/${{ github.repository_owner }}/${{ matrix.name }}-ci
 
     steps:
-      # https://github.com/docker/setup-qemu-action
       - name: Set up QEMU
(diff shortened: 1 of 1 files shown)


piergiorgioladisa:
> cc @piergiorgioladisa if you want to take a look at this simplification. I believe it's not fixing any actual security issue, but this would make the workflow more resilient for future evolution given the sensitivity since it's running in the context of `pull_request_target` and pulling untrusted code and potentially untrusted git metadata.

really great job in simplifying and securing the workflow. I can't see any issue, so everything looks good to me

.github/workflows/build-images-ci.yml
```diff
@@ -126,37 +115,6 @@ jobs:
...
             quay.io/${{ github.repository_owner }}/${{ matrix.name }}-ci:${{ steps.tag.outputs.name }}
 
-      - name: Sign Container Image
-        if: github.event_name == 'push'
-        env:
-          COSIGN_EXPERIMENTAL: "true"
-        run: |
-          cosign sign -y quay.io/${{ github.repository_owner }}/${{ matrix.name }}-ci@${{ steps.docker_build_ci_main.outputs.digest }}
-
-      - name: Generate SBOM
```
kkourt, line 136: A counter argument for removing this might be that we will get a chance to notice 'SBOM' breaking before an actual release. I don't mind that, and we can always try to do `-rc` prereleases to catch such issues.

kkourt (approved):
LGTM, thanks!

.github/workflows/build-images-ci.yml
mtardy: yep that's true

mtardy: However one of my arguments would be that maybe we should remove it everywhere since nobody's maintaining it, and if somebody actually needsit, maybe we should dedicate time and effort.

willfindlay (approved):
I love it

PR #3034, opened 2024-10-23
workflows: use GitHub arm64 runners instead of actuated

mtardy:
CNCF stopped funding actuated arm64 runners but GitHub now proposes their own runners part of the GitHub offering.


kkourt:
Disabled labels e2e test, see https://github.com/cilium/tetragon/issues/3060

kkourt (approved):

PR #3035, opened 2024-10-23
version: add version string to metrics

kkourt:
Before this patch:
tetragon_build_info{commit="61b575c72b22aa8bb6d7ed8af2d3dbfee43f6307",go_version="go1.23.1",modified="true",time="2024-10-23T06:20:18Z"} 1

After:
tetragon_build_info{commit="61b575c72b22aa8bb6d7ed8af2d3dbfee43f6307",go_version="go1.23.1",modified="true",time="2024-10-23T06:20:18Z",version="v1.3.0-pre.0-216-g61b575c72"} 1

```release-note
metrics: add version to build information
```

### DIFF
--- a/cmd/tetragon-metrics-docs/metricsmd/metricsmd.go
+++ b/cmd/tetragon-metrics-docs/metricsmd/metricsmd.go
@@ -45,6 +45,10 @@ func New(targets map[string]string, init initMetricsFunc) *cobra.Command {
 					Label:  "time",
 					Values: []string{"2022-05-13T15:54:45Z"},
 				},
+				{
+					Label:  "version",
+					Values: []string{"v1.2.0"},
+				},
 			},
 		},
 	}
--- a/docs/content/en/docs/reference/metrics.md
+++ b/docs/content/en/docs/reference/metrics.md
@@ -28,6 +28,7 @@ Build information about tetragon
 | `go_version` | `go1.22.0` |
 | `modified` | `false` |
 | `time ` | `2022-05-13T15:54:45Z` |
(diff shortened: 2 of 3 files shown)


mtardy (approved):
Sure!! You just miss some metrics doc generation.

kkourt:
> Sure!! You just miss some metrics doc generation

I think the problem is that the docs are based on the SHA of the version, so every time I update it the SHA changes. I'll try to find a workaround.

mtardy:
> > Sure!! You just miss some metrics doc generation
> 
> I think the problem is that the docs are based on the SHA of the version, so every time I update it the SHA changes. I'll try to find a workaround.

Oh yes put a constant in there :) https://github.com/cilium/tetragon/blob/c0e513f683a6b8dd4a837cafd4cd13b4b6b86234/cmd/tetragon-metrics-docs/metricsmd/metricsmd.go#L31-L49

mtardy:
> Did something else. Happy to change it to what you suggested if it's an issue, but I think overriding Version also works.

I think I would do it like Anna did with the rest of the values in there so that you don't touch the Makefile and we don't have two ways of doing the same thing.

kkourt:
> > Did something else. Happy to change it to what you suggested if it's an issue, but I think overriding Version also works.
> 
> I think I would do it like Anna did with the rest of the values in there so that you don't touch the Makefile and we don't have two ways of doing the same thing.

Good point! Pushed a new version.

lambdanis (approved):
that's a useful label, thanks!

PR #3039, opened 2024-10-24
tetragon: fix the process exit signal when core dumped

justin0u0:
### Description

The ProcessExit event returns a signal that the process received when it exited. However, core dumped signals (SIGSEGV, SIGILL, etc.) are not returned.

This patch fixes the issue by checking if the process exit code has the core dump bit set (7th bit, 0x80), and returns the correct signal. I used the kernel [code](https://elixir.bootlin.com/linux/v5.19/source/kernel/signal.c#L2076-L2078) as a reference to fix it.

### How to Verify

Launch the `tetra` CLI with the following command:

```bash
tetra getevents --process "run" --output compact
```

Then launch a random program and terminate it using different signals:

```bash
./run &
kill -9 $!
./run &
kill -11 $!
```

**Expected Behavior:**

Before: the signal for core dumped processes is not displayed correctly.

before

After: the correct signal for core dumped processes is displayed.
after

### DIFF
--- a/contrib/tester-progs/.gitignore
+++ b/contrib/tester-progs/.gitignore
@@ -23,3 +23,4 @@ drop-privileges
 change-capabilities
 direct-write-tester
 user-stacktrace
+pause
--- a/contrib/tester-progs/Makefile
+++ b/contrib/tester-progs/Makefile
@@ -24,7 +24,8 @@ PROGS = sigkill-tester \
 	getcpu \
 	direct-write-tester \
 	change-capabilities \
-	user-stacktrace
+	user-stacktrace \
+	pause
 
 
 all: $(PROGS)
--- a/contrib/tester-progs/pause.c
+++ b/contrib/tester-progs/pause.c
@@ -1,0 +1,7 @@
+#include <unistd.h>
+
+int main(int argc, char** argv)
+{
+	pause();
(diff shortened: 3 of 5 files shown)


justin0u0:
> Hey welcome to Tetragon 👋 , thanks for writing this patch!
> 
> Since you are adding something that was missing maybe we could add a test for your change, you can take a look at tests in `pkg/sensors/exec/excit_test.go` for example. We have some programs in `contrib/tester-progs` that you could reuse just to start some process that waits forever I think.
> 
> Please reach out if you need more pointers, that would be a pleasure.

Hi @mtardy,

I've added a test to verify process exit signals from numbers 1 to 15.

Since I couldn’t find a program in contrib/tester-progs that fits my use case, I added a simple pause program to achieve this.

Please take a look when you can. Thanks!

contrib/tester-progs/pause
mtardy, line 1: can you add this to the `contrib/tester-progs/.gitingore` and not commit the binary :)?

mtardy (commented):
Thanks that's great :) just a small git related change:

contrib/tester-progs/pause
justin0u0: Fixed it, thanks for the reminder!

PR #3043, opened 2024-10-24
Remove --enable-process-ancestors flag

michi-covalent:
Tetragon does not use this flag, so let's just remove it instead of deprecating it first.


tixxdz (approved):

jrfastab:
Yep wishful thinking it never existed.

PR #3044, opened 2024-10-25
tetra: fix `--policy-names` to apply all event types

justin0u0:
### Description

In #1867, the `--policy-names` flag was added to filter events based on the tracing policy. However, the filter was only appled to `kprobe` events.

This patch extends the filter to support all events types: `kprobe`, `tracepoint`, `uprobe` and `lsm`.

### DIFF
--- a/pkg/filters/filters.go
+++ b/pkg/filters/filters.go
@@ -127,5 +127,17 @@ func GetPolicyName(event *v1.Event) string {
 	if !ok {
 		return ""
 	}
-	return helpers.ResponseGetProcessKprobe(response).GetPolicyName()
+
+	switch ev := (response.Event).(type) {
+	case *tetragon.GetEventsResponse_ProcessKprobe:
+		return ev.ProcessKprobe.GetPolicyName()
+	case *tetragon.GetEventsResponse_ProcessTracepoint:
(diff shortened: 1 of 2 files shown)


pkg/filters/filters.go
```diff
@@ -127,5 +127,17 @@ func GetPolicyName(event *v1.Event) string {
 	if !ok {
 		return ""
 	}
-	return helpers.ResponseGetProcessKprobe(response).GetPolicyName()
+
+	switch ev := (response.Event).(type) {
+	case *tetragon.GetEventsResponse_ProcessKprobe:
+		return ev.ProcessKprobe.PolicyName
```
kkourt, line 133: Any reason for not using `GetPolicyName()` here? It's a bit safer if the pointer is `nil` which shouldn't happen normally but better to avoid crashing

kkourt (commented):
Thanks!

Other than a minor comment, this LGTM!

pkg/filters/filters.go
justin0u0: I think using `GetPolicyName()` is better, thanks!

kkourt (approved):
LGTM, thanks!

PR #3045, opened 2024-10-25
tetragon: Load base sensor via sensor manager

olsajiri:
Loading initial sensor via sensor manager so it stores all the
loaded sensors/programs. So far we kept it separated but having
it in one place is handy for metrics requesting this data.

plus related fixes


pkg/sensors/handler.go
```diff
@@ -19,6 +20,10 @@ type handler struct {
 	pfState      policyfilter.State
 }
 
+var (
+	muLoad sync.Mutex
```
kkourt, line 25: Could we add this to handler, instead of having a global lock?

kkourt (commented):
One comment from a quick look.

kkourt (approved):
🚀

mtardy (approved):

PR #3052, opened 2024-10-25
helm: Add part-of and component labels

lambdanis:
Also refactor the labels templates (deduplicate).


olsajiri (approved):
lgtm, perhaps we could add test for parseCgroupv2RootControllers ? we have some for parseCgroupv1SubSysIds

PR #3053, opened 2024-10-27
tetragon: improve how we handle cgroupv1 and cgroupv2

tixxdz:
kernel v6.11 introduced new CONFIG_MEMCG_V1 and CONFIG_CPUSETS_V1 [1]
that changed how userspace detects the compiled cgroup controllers,
if the new CONFIG_MEMCG_V1=n then memory and same for cpuset controller
will not be exported in /proc/cgroups.

In Cgroupv2 we automatically handle this with these changes. If your distribution
runs in Cgroupv1 then these changes also handle it assuming you have
both CONFIG_MEMCG=y and CONFIG_MEMCG_V1=y are set.  

[1] "af000ce85293b8e60" "cgroup: Do not report unavailable v1 controllers in /proc/cgroups"

Kernel v6.11 introduced a new CONFIG_MEMCG_V1 for legacy cgroup v1 memory controller which may affect Tetragon cgroup environment detection, process and container/pod association. Improve Tetragon so we handle the new configuration and allow process container association to properly work if the distribution boots in default unified hierarchy Cgroupv2.
If the distribution boots in Hybrid Cgroupv1 mode then ensure the appropriate controllers are exported otherwise report an error and print the kernel config that are needed, probably: CONFIG_MEMCG=y, CONFIG_MEMCG_V1=y, CONFIG_CPUSETS=y and CONFIG_CPUSETS_V1=y all must be set to have proper process and container/pod association. 

```release-note:
cgroups: fix kernel  config issue introduced in 6.11. For cgroup v1, setting CONFIG_MEMCG=y, CONFIG_MEMCG_V1=y, CONFIG_CPUSETS=y and CONFIG_CPUSETS_V1=y is required for Tetragon to work.
```

### DIFF
--- a/pkg/cgroups/cgroups.go
+++ b/pkg/cgroups/cgroups.go
@@ -215,7 +215,7 @@ func parseCgroupv1SubSysIds(filePath string) error {
 				/* We care only for the controllers that we want */
 				if idx >= CGROUP_SUBSYS_COUNT {
 					/* Maybe some cgroups are not upstream? */
-					return fmt.Errorf("Cgroup default subsystem '%s' is indexed at idx=%d higher than CGROUP_SUBSYS_COUNT=%d",
+					return fmt.Errorf("Cgroupv1 default subsystem '%s' is indexed at idx=%d higher than CGROUP_SUBSYS_COUNT=%d",
 						fields[0], idx, CGROUP_SUBSYS_COUNT)
 				}
 
@@ -229,7 +229,7 @@ func parseCgroupv1SubSysIds(filePath string) error {
 					logger.GetLogger().WithFields(logrus.Fields{
 						"cgroup.fs":              cgroupFSPath,
 						"cgroup.controller.name": controller.Name,
-					}).WithError(err).Warnf("parsing controller line from '%s' failed", filePath)
+					}).WithError(err).Warnf("Cgroupv1 parsing controller line from '%s' failed", filePath)
 				}
 			}
 		}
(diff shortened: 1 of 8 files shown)


olsajiri (approved):
lgtm, and perhaps we could add test for parseCgroupv2RootControllers

pkg/cgroups/cgroups.go
```diff
@@ -269,12 +269,90 @@ func parseCgroupSubSysIds(filePath string) error {
...
+		"cgroup.fs":          cgroupFSPath,
+		"cgroup.controllers": strings.Fields(activeControllers),
+	}).Debug("Cgroupv2 available controllers")
+
+	fixed := false
+	for i, controller := range CgroupControllers {
+		if strings.Contains(activeControllers, controller.Name) {
+			CgroupControllers[i].Id = 0
+			CgroupControllers[i].Active = true
+			fixed = true
```
kkourt, line 303: Do we need thus for v2? My understanding was that the first patch modified the bpf code to just use ->dfl_cgroup instead.

tixxdz: I simplified it a bit and removed this part, and kept the environment logging logic so in case of race conditions or any other container association failure we can try to reproduce.

tixxdz:
> lgtm, and perhaps we could add test for parseCgroupv2RootControllers

Thank you, yes I added it.

pkg/cgroups/cgroups.go
kkourt: > I simplified it a bit and removed this part, and kept the environment logging logic so in case of race conditions or any other container association failure we can try to reproduce.

Not sure I understand. Can you please explain?

tixxdz: > > I simplified it a bit and removed this part, and kept the environment logging logic so in case of race conditions or any other container association failure we can try to reproduce.
> 
> Not sure I understand. Can you please explain?

hmm what part exactly?  also the commit log contains explanation and some extra code comments

kkourt: Discussed offline. There seem to be some races, and we have logging to detect them.

kkourt (approved):

PR #3057, opened 2024-10-30
rthooks: Fix rootDir in createRuntime hook

tpapagian:
When using the createContainer hook in containerd or cri-o the rootDir points to the root directory of the container that is created.

In the case where we use the createRuntime hook in containerd the roodDir points to the location of the config.json file which is not the same as the container root directory. In order to fix that, we need to append the Root.Path from the spec.


kkourt (approved):
Thanks!

PR #3059, opened 2024-10-30
bpf: remove unused const PossibleCPUSysfsPath

tklauser:
It's unused since switching to use cilium/ebpf functionality to get the number of possible CPUs in commit 582da32e11bd ("tetragon: Implement bpf.GetNumPossibleCPUs through cilium/ebpf library").


kkourt (approved):
Thanks!

PR #3077, opened 2024-11-06
Renovate: Extend Helm files coverage

lambdanis:
Include Chart.yaml to automatically bump dependencies in case we add some. Also
include all directories under install/kubernetes.


mtardy (approved):
Thanks!

Issue #3033, opened 2024-10-23
Option `--keep-sensors-on-exit` doesn't work with disabling/enabling policies


mtardy:
I think I bumped into an unexpected thing while [fixing the memory use of disabled policies](https://github.com/cilium/tetragon/pull/3021#discussion_r1812430868).

If you use the `--keep-sensors-on-exit`, then when you use `tetra tp disable <policy-name>`, you trigger `.Unload()` on the program which is expecting mostly to be called on exit it seems since we don't unpin the progs if `option.Config.KeepSensorsOnExit` is here.

https://github.com/cilium/tetragon/blob/61b575c72b22aa8bb6d7ed8af2d3dbfee43f6307/pkg/sensors/unloader/unloader.go#L59-L65

Thus you didn't disable the policy, so `--keep-sensors-on-exit` doesn't work currently with the enable/disable feature.

cc @olsajiri

PR #3079, opened 2024-11-06
tetragon: un/pin fixes

olsajiri:
factor sensors unpin-ing

fixes https://github.com/cilium/tetragon/issues/3033

### DIFF
--- a/pkg/sensors/tracing/enforcer_test.go
+++ b/pkg/sensors/tracing/enforcer_test.go
@@ -827,6 +827,80 @@ func testEnforcerPersistentNoKeep(t *testing.T, builder func() *EnforcerSpecBuil
 	run(2, "exit status 22")
 }
 
+// We test following scenario:
+// - load enforcement policy
+// - 1st run of test binary, make sure enforcement policy is triggered
+// - disable enforcement policy via sensor manager
+// - 2nd run of test binary, no enforcement
+// - enable enforcement policy via sensor manager
+// - 3rd run of test binary, make sure enforcement policy is triggered
+// - remove enforcement policy via sensor manager
+// - 4th run of test binary, no enforcement
+func testEnforcerPersistentUnload(t *testing.T, builder func() *EnforcerSpecBuilder, expected, test string) {
+	testEnforcerCheckSkip(t)
+
+	if !bpf.HasLinkPin() {
+		t.Skip("skipping persistent enforcer test, link pin is not available")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	option.Config.KeepSensorsOnExit = true
+	defer func() { option.Config.KeepSensorsOnExit = false }()
+
+	tus.LoadInitialSensor(t)
+	path := bpf.MapPrefixPath()
+	mgr, err := sensors.StartSensorManager(path)
+	assert.NoError(t, err)
+
+	run := func(idx int, exp string) {
+		cmd := exec.Command(test, "0xfffe")
+		err := cmd.Run()
+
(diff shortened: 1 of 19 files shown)


pkg/sensors/tracing/enforcer_test.go
```diff
@@ -823,6 +823,56 @@ func testEnforcerPersistentNoKeep(t *testing.T, builder func() *EnforcerSpecBuil
...
+	assert.NoError(t, err)
+
+	err = mgr.AddTracingPolicy(ctx, tp)
+	assert.NoError(t, err)
+
+	// first run - sensors are loaded, we should get kill/override
+	run(1, expected)
+
+	// remove the policy and we should get rid of the enforcement
+	err = mgr.DeleteTracingPolicy(ctx, tp.TpName(), "")
```
mtardy, line 866: In the commit you specify "test that simulates enforcement policy unload (not exit)" but you use `Delete` instead of `DisableTracingPolicy`, is that intended? You don't exactly test the scenario here https://github.com/cilium/tetragon/issues/3033, maybe you want to modify one or add another one?

mtardy (approved):
looks good! The 2 sec sleeps in testing don't really look awesome though, isn't there a way to write something faster and more reliable?

pkg/sensors/tracing/enforcer_test.go
olsajiri: I see, it should be about the same, but I can add that scenario as well, thnx

olsajiri:
> looks good! The 2 sec sleeps in testing don't really look awesome though, isn't there a way to write something faster and more reliable?

yea can't think of any now.. but will give it some more thought

pkg/observer/observer.go
```diff
@@ -341,8 +341,8 @@ func (k *Observer) Start(ctx context.Context) error {
 }
 
 // InitSensorManager starts the sensor controller
-func (k *Observer) InitSensorManager() error {
-	mgr, err := sensors.StartSensorManager(option.Config.BpfDir)
+func (k *Observer) InitSensorManager(keepSensorsOnExit bool) error {
+	mgr, err := sensors.StartSensorManager(option.Config.BpfDir, keepSensorsOnExit)
```
kkourt, line 345: This is not an objection, but I would like for us to have a consistent way of accessing the config options. 

Some time ago @jrfastab did https://github.com/cilium/tetragon/commit/64b236be12e6feb440fbc718ac1f8460bd8e674a, where the idea (as I'm reading it) was to add configuration to under `option.Config` and use that from everywhere.

The above change goes into a different direction and passes `option.Config.KeepSensorsOnExit`. I don't have a strong opinion on what approach should we take, but I do think we should be consistent.

If we want to avoid pass the argument and use `option.Config.KeepSensorsOnExit` one complication is that it makes it tricky to do testing because the default config value might change and you might want to test a specific value.

If that's the issue here, one of the ways to deal with it is to save the default value, set the value you want and restore the default value after the test. See: https://github.com/cilium/tetragon/blob/93939d149633c648bc5cc17fd157145a2d034e4a/pkg/policyfilter/k8s_test.go#L712-L719

olsajiri: the issue is that we check option.Config.KeepSensorsOnExit un Unload but we don't know if we unload because tetragon is going down or we unload because we just removed policy.. both these cases have different path:
- when tetragon is going down we want to follow  KeepSensorsOnExit
- when we just unload sensor we DO NOT want to follow KeepSensorsOnExit

so that's why I store that in sensor manager and use it in tetragon exit path for sensor unload
I should have put this in some comments, sorry

olsajiri: ok, using KeepSensorsOnExit directly from SensorManager::RemoveAllSensors function, thanks

kkourt (approved):
Thanks!

PR #3080, opened 2024-11-07
Switch to Go 1.23+ stdlib `maps`/`slices` packages

tklauser:
Use the `maps` and `slices` packages from the Go standard library instead of the respective `golang.org/x/exp` packages.

Also add a linter check enforcing use of these packages.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -24,6 +24,7 @@ linters:
     - gosimple
     - makezero
     - dupword
+    - gomodguard
 
 linters-settings:
   goheader:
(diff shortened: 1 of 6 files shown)


tklauser:
> Thanks! Especially for the linter check, very nice! I need to read about this new iterator feature, I didn't know.

In case you didn't see it yet, https://go.dev/blog/range-functions is a nice introduction to the new iterators (also called range-over-func) in Go 1.23.

tixxdz (approved):
Much appreciated @tklauser

PR #3097, opened 2024-11-08
tetragon: Move procevents.GetRunningProcs call from base sensor load

olsajiri:
We need to call procevents.GetRunningProcs after server is started observer listeners are configured.

Currently it's called within base sensor load, but it's now loaded before server is started, so we don't get any initial exec events.

### DIFF
--- a/pkg/sensors/exec/threads_test.go
+++ b/pkg/sensors/exec/threads_test.go
@@ -18,7 +18,7 @@ import (
 	"github.com/cilium/tetragon/pkg/observer"
 	"github.com/cilium/tetragon/pkg/observer/observertesthelper"
 	"github.com/cilium/tetragon/pkg/option"
-	"github.com/cilium/tetragon/pkg/sensors/base"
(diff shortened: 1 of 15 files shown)


pkg/sensors/exec/threads_test.go
```diff
@@ -98,7 +98,12 @@ func TestMatchCloneThreadsIDs(t *testing.T) {
 	}
 
 	option.Config.HubbleLib = tus.Conf().TetragonLib
-	tus.LoadSensor(t, base.GetInitialSensor())
+	tus.LoadInitialSensor(t)
+
+	if err := procevents.GetRunningProcs(); err != nil {
```
tpapagian, line 103: Do we need that here? It seems that `tus.LoadInitialSensor(t)` also calls `procevents.GetRunningProcs()`.

olsajiri: ugh, leftover, will remove, thanks

kkourt (approved):
Thanks!

PR #3108, opened 2024-11-14
Add condition for procfs/version if procfs/version_signature fails

harisudarsan1:
### Description
Previously func GetKernelVersion(kernelVersion string, procfs string) (int, string, error) helper only checks the procfs/version_signature. I have added a condition to check procfs/version if first check fails.


olsajiri:
> func GetKernelVersion(kernelVersion string, procfs string) (int, string, error) helper only checks the procfs/version_signature. I have added a condition to check procfs/version

the current code falls back to get version from uname..  what's the benefit of reading /proc/version? is there some scenario where current code is broken/gives bad result?

harisudarsan1:
@olsajiri I added /proc/version because it is present in most of the distributions compared to /proc/version_signature just as an additional check.

harisudarsan1:
But I think if we are anyways falling back to uname then there is no need of /proc/version

Issue #2980, opened 2024-10-05
DOCS: update minikube installation scripts in k8s identity aware policies


paularah:
We now have a slight modification in the minikube installation scripts. The `minikube-containerd-install-hook.sh` file no longer exists. We should update the docs to reflect this.

d-cryptic:
Hi, Can you assign this task to me?

paularah:
> Hi, Can you assign this task to me?

cc: @mtardy

kkourt:
> Hi, Can you assign this task to me?

Sure, assigned!
Note that the new scripts now live in: https://github.com/cilium/tetragon/tree/main/contrib/tetragon-rthooks/scripts.

Thanks!

mtardy:
Fixed by https://github.com/cilium/tetragon/pull/3111

PR #3111, opened 2024-11-14
fix: correcting the script path for minikube installation steps in do…

d-cryptic:
### Describe what changes this pull request brings
- Updated the docs with the correct path for `minikube-install-hook.sh` script.

### Related issue
#2980 

### Before change
image

### After Change
image

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
+++ b/docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
@@ -59,7 +59,7 @@ First, let us start minikube, build and load images, and install Tetragon and OC
 
 ```shell
 minikube start --container-runtime=containerd
-./contrib/tetragon-rthooks/minikube-containerd-install-hook.sh
+./contrib/tetragon-rthooks/scripts/minikube-install-hook.sh
 make image image-operator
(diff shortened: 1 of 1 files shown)


d-cryptic:
Hi, 
I only found a single reference for install-hooks scripts.
I have a query regarding tetragon-oci-hook-setup in this [documentation page](https://tetragon.io/docs/installation/runtime-hooks/#setup-cluster).
This `./contrib/tetragon-rthooks/tetragon-oci-hook-setup` not present in current main branch.
Does this needs to be changed too? if so, what's the correct script path for this?


image

mtardy:
> I have a query regarding tetragon-oci-hook-setup in this [documentation page](https://tetragon.io/docs/installation/runtime-hooks/#setup-cluster).
> This `./contrib/tetragon-rthooks/tetragon-oci-hook-setup` not present in current main branch.
> Does this needs to be changed too? if so, what's the correct script path for this?

This is a binary you need to compile in order for it to be around :)

mtardy (approved):
Thanks for doing the change :)

mtardy (commented):
Could you shorten your commit message a little bit?

> Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 80)

paularah:
@d-cryptic do you need help with this?

Issue #2947, opened 2024-09-23
policies with the same symbols are broken 


kkourt:
Policies that use the same symbols do not currently work. Here's an example:

```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "multiple-symbols"
spec:
  kprobes:
  - call: __x64_sys_prctl
    args:
    - index: 0
      type: int64
    selectors:
    - matchArgs:
      - index: 0
        operator: Equal
        values:
        - "3"
    syscall: true
    tags:
    - pr_get_dumpable
  - call: __x64_sys_prctl
    args:
    - index: 0
      type: int64
    selectors:
    - matchArgs:
      - index: 0
        operator: Equal
        values:
        - "7"
    syscall: true
    tags:
    - pr_get_keepcaps
```

When using multi-kprobes, using this policy leads to the following error:
```
level=fatal msg="Failed to start tetragon" error="failed to get sensors from parser policy: sensor generic_kprobe from collection auditd-policy failed to load: failed prog /home/kkourt/src/hubble-fgs/bpf/objs/bpf_multi_kprobe_v61.o kern_version 394509 loadInstance: attaching 'generic_kprobe_event' failed: couldn't find one or more symbols: file does not exist" 
```

Disabling multi-kprobes, only part of the policy is applied (typically the one that is defined last).
The reason for this seems to be that when using the new bpffs hierarchy (https://github.com/cilium/tetragon/pull/2128), the two calls will end up in the same directory, using the same maps. 

Note that in `1.2` this works as expected, because we use different maps for each different hook. e.g., 
```
gkp-sensor-1-gkp-0-argfilter_maps
gkp-sensor-1-gkp-1-argfilter_maps
```

kkourt:
@olsajiri can you PTAL? 

When we are not using multi-kprobes, I think it would make sense to add an id to the call in the directory, so that we can separate the two.

PR #3121, opened 2024-11-15
tetragon: Allow multiple symbol instances in kprobe spec

olsajiri:
it's ok for non kprobe-multi specs and we retrn error in kprobe-multi mode

Fixes: https://github.com/cilium/tetragon/issues/2947

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -109,6 +109,7 @@ type genericKprobe struct {
 	argSigPrinters    []argPrinter
 	argReturnPrinters []argPrinter
 	funcName          string
+	instance          int
 
 	// for kprobes that have a retprobe, we maintain the enter events in
 	// the map, so that we can merge them when the return event is
@@ -626,6 +627,7 @@ func createGenericKprobeSensor(
 	}
 
 	has := hasMapsSetup(spec)
+	dups := make(map[string]int)
 
 	for i := range kprobes {
 		syms, syscall, err := getKprobeSymbols(kprobes[i].Call, kprobes[i].Syscall, lists)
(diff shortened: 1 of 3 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -897,11 +912,16 @@ func createKprobeSensorFromEntry(kprobeEntry *genericKprobe,
 	loadProgName, loadProgRetName := kernels.GenericKprobeObjs()
 	isSecurityFunc := strings.HasPrefix(kprobeEntry.funcName, "security_")
 
+	pinProg := kprobeEntry.funcName
+	if kprobeEntry.instance != 0 {
+		pinProg = fmt.Sprintf("%s_%d", kprobeEntry.funcName, kprobeEntry.instance)
```
kkourt, line 917: How about we use "%s:%d"? I'm just thinking that in theory there could be a fund name "foo_1" in the kernel that may cause collisions.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -986,6 +1006,9 @@ func createKprobeSensorFromEntry(kprobeEntry *genericKprobe,
 
 	if kprobeEntry.loadArgs.retprobe {
 		pinRetProg := sensors.PathJoin(fmt.Sprintf("%s_return", kprobeEntry.funcName))
+		if kprobeEntry.instance != 0 {
+			pinRetProg = sensors.PathJoin(fmt.Sprintf("%s_%d_return", kprobeEntry.funcName, kprobeEntry.instance))
```
kkourt, line 1010: ditto

kkourt (approved):
Left a minor comment, but otherwise LGTM.
Thanks!

pkg/sensors/tracing/generickprobe.go
olsajiri: ook, will change, thanks

PR #3122, opened 2024-11-15
Add tetra policyfilter listpolicies command

tpapagian:
Add `tetra policyfilter listpolicies` to determine which Kubernetes Identity Aware policies should be applied on a specific container.

Example:
```
$ kubectl exec -it ds/tetragon -n kube-system -c tetragon -- tetra policyfilter -r "unix:///procRoot/1/root/run/containerd/containerd.sock" listpolicies ff433e9e16467787a60ac853d9b313150091968731f620776d6d7c514b1e8d6c
ID   NAME                    STATE     FILTERID   NAMESPACE   SENSORS          KERNELMEMORY
5    lseek-podfilter-usage   enabled   5          (global)    generic_kprobe   1.72 MB
1    lseek-podfilter-app     enabled   1          (global)    generic_kprobe   1.72 MB
```

Details on how this works can be found on specific commits.

### DIFF
--- a/bpf/process/policy_filter.h
+++ b/bpf/process/policy_filter.h
@@ -9,6 +9,7 @@
 
 #define POLICY_FILTER_MAX_POLICIES   128
 #define POLICY_FILTER_MAX_NAMESPACES 1024
+#define POLICY_FILTER_MAX_CGROUP_IDS 1024
 
 struct {
 	__uint(type, BPF_MAP_TYPE_LRU_HASH);
@@ -29,6 +30,23 @@ struct {
 			__type(value, __u8); /* empty  */
 		});
 } policy_filter_maps SEC(".maps");
+
+// This map keeps exactly the same information as policy_filter_maps
+// but keeps the reverse mappings. i.e.
+// policy_filter_maps maps policy_id to cgroup_ids
+// policy_filter_cgroup_maps maps cgroup_id to policy_ids
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH_OF_MAPS);
+	__uint(max_entries, POLICY_FILTER_MAX_CGROUP_IDS);
+	__uint(key_size, sizeof(__u64)); /* cgroup id */
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, POLICY_FILTER_MAX_POLICIES);
+			__type(key, __u32); /* policy id */
+			__type(value, __u8); /* empty  */
+		});
+} policy_filter_cgroup_maps SEC(".maps");
 
 // policy_filter_check checks whether the policy applies on the current process.
 // Returns true if it does, false otherwise.
--- a/cmd/tetra/debug/dump.go
+++ b/cmd/tetra/debug/dump.go
@@ -200,18 +200,20 @@ func PolicyfilterState(fname string) {
 		fmt.Printf("%d: %s\n", polId, strings.Join(ids, ","))
 	}
 
-	fmt.Println("--- CgroupID to PolicyIDs mapping ---")
+	if data.Cgroup != nil {
+		fmt.Println("--- CgroupID to PolicyIDs mapping ---")
 
-	if len(data.Cgroup) == 0 {
-		fmt.Printf("(empty)\n")
-	}
-
-	for cgIDs, polIds := range data.Cgroup {
-		ids := make([]string, 0, len(polIds))
-		for id := range polIds {
-			ids = append(ids, strconv.FormatUint(uint64(id), 10))
+		if len(data.Cgroup) == 0 {
+			fmt.Printf("(empty)\n")
 		}
-		fmt.Printf("%d: %s\n", cgIDs, strings.Join(ids, ","))
+
+		for cgIDs, polIds := range data.Cgroup {
+			ids := make([]string, 0, len(polIds))
+			for id := range polIds {
+				ids = append(ids, strconv.FormatUint(uint64(id), 10))
+			}
+			fmt.Printf("%d: %s\n", cgIDs, strings.Join(ids, ","))
+		}
 	}
 }
 
--- a/cmd/tetra/debug/dump.go
+++ b/cmd/tetra/debug/dump.go
@@ -186,17 +186,32 @@ func PolicyfilterState(fname string) {
 		return
 	}
 
-	if len(data) == 0 {
+	fmt.Println("--- PolicyID to CgroupIDs mapping ---")
+
+	if len(data.Policy) == 0 {
 		fmt.Printf("(empty)\n")
-		return
 	}
 
-	for polId, cgIDs := range data {
+	for polId, cgIDs := range data.Policy {
 		ids := make([]string, 0, len(cgIDs))
 		for id := range cgIDs {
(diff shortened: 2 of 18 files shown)


bpf/process/policy_filter.h
```diff
@@ -30,6 +31,23 @@ struct {
...
+	__uint(key_size, sizeof(__u64)); /* cgroup id */
+	__array(
+		values, struct {
+			__uint(type, BPF_MAP_TYPE_HASH);
+			__uint(max_entries, POLICY_FILTER_MAX_POLICIES);
+			__type(key, __u32); /* policy id */
+			__type(value, __u8); /* empty  */
+		});
+} policy_filter_reverse_maps SEC(".maps");
+
```
kkourt, line 50: I think naming the map "reverse" will make the code/etc. harder to understand.
How about `cgroup_policies` or `policyfilter_cgoup_policies`?
What do you think?

cmd/tetra/debug/dump.go
```diff
@@ -186,18 +186,33 @@ func PolicyfilterState(fname string) {
...
-	for polId, cgIDs := range data {
+	for polId, cgIDs := range data.Direct {
 		ids := make([]string, 0, len(cgIDs))
 		for id := range cgIDs {
 			ids = append(ids, strconv.FormatUint(uint64(id), 10))
 		}
 		fmt.Printf("%d: %s\n", polId, strings.Join(ids, ","))
 	}
+
+	fmt.Println("--- Reverse Map ---")
```
kkourt: Even if we do not rename the map I find that "Direct" and "Reverse" would make things harder to understand. Let's describe what is the key and what is the value here to make it easier to understand the output.

pkg/policyfilter/map.go
```diff
@@ -95,36 +177,109 @@ func (m PfMap) newPolicyMap(polID PolicyID, cgIDs []CgroupID) (polMap, error) {
...
+	}
+
+	if err := inIter.Err(); err != nil {
+		return 0, fmt.Errorf("error iterating inner map: %w", err)
+	}
+
+	return mapSize, nil
+}
+
+func (m PfMap) deletePolicyIdInReverse(polID PolicyID) error {
```
mtardy, line 237: nit `deletepolicyIDInReverse` since we use `ID` everywhere else :)

pkg/policyfilter/map.go
```diff
@@ -95,36 +177,109 @@ func (m PfMap) newPolicyMap(polID PolicyID, cgIDs []CgroupID) (polMap, error) {
...
+		if err != nil {
+			return fmt.Errorf("error getting inner map size: %w", err)
+		}
+		if mapSize == 0 {
+			cgIDs = append(cgIDs, key)
+		}
+	}
+
+	if err := iter.Err(); err != nil {
+		return fmt.Errorf("deletePolicyIdInReverse: error iterating reverse map: %w", err)
```
mtardy, line 271: if you change, don't forget to change here.

pkg/policyfilter/map.go
```diff
@@ -78,6 +91,75 @@ func (m PfMap) release() error {
...
 
+// release closes the policy filter bpf map and remove (unpin) the bpffs file
+func (m PfMap) release() error {
+	if err := releaseMap(m.dir); err != nil {
+		return err
+	}
+
+	if err := releaseMap(m.rev); err != nil {
+		return err
+	}
```
mtardy, line 111: not sure this is actually relevant: should you try to release both maps here and maybe join error? So that you are still trying to release as maximum as possible in the case the first one fails?

cmd/tetra/policyfilter/policyfilter.go
```diff
@@ -138,3 +147,131 @@ func addCgroup(fname string, polID policyfilter.PolicyID, cgID policyfilter.Cgro
...
+					pol.Id,
+					pol.Name,
+					strings.TrimPrefix(strings.ToLower(pol.State.String()), "tp_state_"),
+					pol.FilterId,
+					namespace,
+					sensors,
+					common.HumanizeByteCount(int(pol.KernelMemoryBytes)),
+				)
+			}
+			w.Flush()
```
mtardy, line 266: Any chance we could reuse this code instead of copy pasting it?

mtardy (commented):
A few comments :)

cmd/tetra/debug/dump.go
tpapagian: I have renamed those to more descriptive headers.

pkg/policyfilter/map.go
tpapagian: Updated that!

pkg/policyfilter/map.go
tpapagian: Same here.

pkg/policyfilter/map.go
tpapagian: Yes, this makes sense. I have updated that to use `errors.Join`.

cmd/tetra/policyfilter/policyfilter.go
tpapagian: Yes, this makes sense. I have created a function to do the printing of tracing policies and used that here.

mtardy: lovely!

mtardy (approved):
Looks good to me! I'll let Kornilios review

bpf/process/policy_filter.h
```diff
@@ -9,6 +9,7 @@
 
 #define POLICY_FILTER_MAX_POLICIES   128
 #define POLICY_FILTER_MAX_NAMESPACES 1024
+#define POLICY_FILTER_MAX_CGROUP_IDS 32768 /* same as polMapSize in policyfilter/state.go */
```
kkourt, line 12: This seems high to me. According to https://kubernetes.io/docs/setup/best-practices/cluster-large/, a good limit for the number of pods per node is ~100. How about having something 512 or 1024 entries?

pkg/policyfilter/map.go
```diff
@@ -17,55 +18,68 @@ import (
...
+	if err != nil {
+		return PfMap{}, fmt.Errorf("loading spec for %s failed: %w", objPath, err)
+	}
+
+	var ret PfMap
+	if ret.policyMap, err = openMap(spec, MapName, polMapSize); err != nil {
+		return PfMap{}, fmt.Errorf("opening map %s failed: %w", MapName, err)
+	}
+	if ret.cgroupMap, err = openMap(spec, CgroupMapName, polMaxPolicies); err != nil {
+		return PfMap{}, fmt.Errorf("opening cgroup map %s failed: %w", MapName, err)
```
kkourt: Do we need to release `policyMap` here?

kkourt (commented):
Thanks!

bpf/process/policy_filter.h
tpapagian: Yes, this makes sense. Just tried to keep that consistent with what we had in `policyfilter/state.go`. 

Changed that to 1024.

pkg/policyfilter/map.go
tpapagian: Correct, fixed that.

kkourt (approved):

PR #3131, opened 2024-11-18
base sensor extensions

kkourt:
This PR introduces extensions for the base sensor. See commits.

### DIFF
--- a/pkg/sensors/base/base.go
+++ b/pkg/sensors/base/base.go
@@ -10,14 +10,9 @@ import (
 	"github.com/cilium/tetragon/pkg/ksyms"
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/mbset"
-	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/sensors"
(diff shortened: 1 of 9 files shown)


pkg/sensors/base/base.go
```diff
@@ -164,24 +154,21 @@ func GetDefaultMaps(cgroupRate bool) []*program.Map {
...
+	sensor.Progs = GetDefaultPrograms(option.CgroupRateEnabled())
+	sensor.Maps = GetDefaultMaps(option.CgroupRateEnabled())
 	return &sensor
 }
 
-func GetInitialSensorTest() *sensors.Sensor {
-	sensorTestInit.Do(func() {
-		setupPrograms()
-		sensorTest.Progs = GetDefaultPrograms(true)
-		sensorTest.Maps = GetDefaultMaps(true)
```
mtardy, line 181: I took a look at this early as it's still in draft but here we had both options forced to `true` in the test version, is that okay that we now always pass the options' value?

kkourt: There are only two tests that use this, and they are setting the configuration variable. The last patch ensures that we are generating a new list of programs/maps for each test, which is the right thing to do I think.

kevsecurity (approved):
LGTM

PR #3141, opened 2024-11-19
tetragon: Unflake the TestGeneratedExecEvents test

olsajiri:
On slow server the test might fails to start exporter fast enough making the test fail, because events did not get stored.

Moving the tetragonExecuteCtx into separate process and adding extra 5 tries on checking up on the exported file.

### DIFF
--- a/cmd/tetragon/main_test.go
+++ b/cmd/tetragon/main_test.go
@@ -6,7 +6,9 @@ package main
 import (
 	"context"
 	"os"
+	"sync"
 	"testing"
+	"time"
 
 	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
 	"github.com/cilium/tetragon/pkg/defaults"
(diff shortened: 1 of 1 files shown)


tpapagian (approved):

cmd/tetragon/main_test.go
```diff
@@ -52,15 +51,39 @@ func TestGeneratedExecEvents(t *testing.T) {
...
 		ec.NewProcessExecChecker("").WithProcess(
 			ec.NewProcessChecker().WithPid(1)),
 	)
 
-	err = jsonchecker.JsonTestCheck(t, checker)
+	// Try it 5 times and make sure exporter is up and processed all data
+	// in case it lags for some reason like slow CI server.
+	cnt := 0
+	for cnt < 5 {
+		if err = jsonchecker.JsonTestCheck(t, checker); err == nil {
```
mtardy, line 80:
```suggestion
		if jsonchecker.JsonTestCheck(t, checker) == nil {
```
FYI the error you will check on line 88 will not be this one as this one is scoped only for this block.

mtardy (commented):
lgtm except maybe something with `assert.NoError(t, err)` that might not be correct.

cmd/tetragon/main_test.go
olsajiri: @mtardy hum, err is created on line 43, line 80 won't create new scope ('=' vs ':=' ) .. what do I miss?

mtardy: ah sorry I missed it somehow I read `:=` because I always use that in if statement... you are right!

mtardy (approved):

Issue #3142, opened 2024-11-19
Add dynamic parameter extration for specific use cases


ScriptSathi:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

Hello ! I have not found any related topics, so I opening it.

For some CVE mitigations, it is necessary to have access to certain hooks, but if you can't reach the data, you won't be able to do anything.
So this feature should allow extracting any parameter from structure argument such as `linux_binprm` or `struct file` and gather any data inside.

### Describe the feature you would like

The feature could look like this

```yaml
# ...
spec:
  lsmhooks:
  - hook: "bprm_check_security"
    args:
      - index: 0
        type: "linux_binprm"
        extractParam: "file.f_path.dentry.d_name.name"
# Or    extractParam: "argc"
# ...
```

After this, we could match the data with the normal Tetragon behavior, like `Equal` or `Postfix`. 

### Describe your proposed solution

To do such a feature, it is possible to search for a specific type in BTF file and extract the offsets of the required data. Then, those offsets could be given to the BPF program that would use it to reach the data.

To make it simpler to understand, let's take an example using the above config
```
- Split extractParam string in order to have an array of childrens
- Search the parent, `linux_binprm` and look for any childrens called `file`.
  If found, we store the offset 
- If `file` parameter is found in `linux_binprm` search inside `file` to reach the child `f_path`
- And so on, until the end of the array.
- After all the offsets have been found, we gave it to the BPF program and loop on the argument buffer to search for the required data.
- Then we read the data and it should work
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #3143, opened 2024-11-19
Add dynamic extraction of a parameter attribute

ScriptSathi:
The discussion for this PR can be found here https://github.com/cilium/tetragon/issues/3142

Take this PR in the today state as a proof of concept for dynamic parameter extraction. I will continue to work on this PR until the below checks are done.

At the current state, the PR is able to
- [x] Extract parameters from LSM programs
- [x] Extract parameters from kprobes programs
- [x] Handle exception for uprobes as it is not part of BTF
- [x] Have unit tests for all cases
- [x] - Validate CI for all kernel versions

### Description
This PR introduce the dynamic extraction of a custom attribute

### Test the PR

You can use the following config
```yaml
kind: TracingPolicy
metadata:
  name: "lsm"
spec:
  lsmhooks:
  - hook: "bprm_check_security"
    args:
      - index: 0
        type: "string"
        resolve: "file.f_path.dentry.d_name.name"
    selectors:
      - matchArgs:
        - index: 0
          operator: "Postfix"
          values:
            - "ls"
            - "sh"
            - "bash"
```

### Release-note

```release-note
Add attribute resolution
```

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -63,6 +63,17 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 }
 
 FUNC_INLINE int
+extract_arg_depth(u32 i, struct extract_arg_data *data)
+{
+	if (i >= MAX_BTF_ARG_DEPTH || !data->btf_config[i].is_initialized)
+		return 1;
+	*data->arg = *data->arg + data->btf_config[i].offset;
+	if (data->btf_config[i].is_pointer)
+		probe_read((void *)data->arg, sizeof(char *), (void *)*data->arg);
+	return 0;
+}
+
+FUNC_INLINE int
 generic_process_event(void *ctx, struct bpf_map_def *tailcals)
 {
 	struct msg_generic_kprobe *e;
@@ -70,6 +81,9 @@ generic_process_event(void *ctx, struct bpf_map_def *tailcals)
 	int index, zero = 0;
 	unsigned long a;
 	long ty, total;
+#ifdef __LARGE_BPF_PROG
+	struct config_btf_arg *btf_config;
+#endif /* __LARGE_BPF_PROG */
 
 	e = map_lookup_elem(&process_call_heap, &zero);
 	if (!e)
@@ -86,6 +100,25 @@ generic_process_event(void *ctx, struct bpf_map_def *tailcals)
 
 	a = (&e->a0)[index];
 	total = e->common.size;
+
+#ifdef __LARGE_BPF_PROG
+	btf_config = config->btf_arg[index];
+	if (index < EVENT_CONFIG_MAX_ARG && btf_config->is_initialized) {
+		struct extract_arg_data extract_data = {
+			.btf_config = btf_config,
+			.arg = &a,
+		};
+#ifndef __V61_BPF_PROG
+#pragma unroll
+		for (int i = 0; i < MAX_BTF_ARG_DEPTH; ++i) {
+			if (extract_arg_depth(i, &extract_data))
+				break;
+		}
+#else
+		loop(MAX_BTF_ARG_DEPTH, extract_arg_depth, &extract_data, 0);
+#endif /* __V61_BPF_PROG */
+	}
+#endif /* __LARGE_BPF_PROG */
 
 	/* Read out args1-5 */
 	ty = (&config->arg0)[index];
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -148,6 +148,20 @@ struct selector_arg_filters {
 	__u32 argoff[5];
 } __attribute__((packed));
 
+struct config_btf_arg {
+	__u32 offset;
+	__u16 is_pointer;
+	__u16 is_initialized;
+} __attribute__((packed));
+
+struct extract_arg_data {
+	struct config_btf_arg *btf_config;
+	unsigned long *arg;
+};
+
+#define MAX_BTF_ARG_DEPTH    10
+#define EVENT_CONFIG_MAX_ARG 5
+
 struct event_config {
 	__u32 func_id;
 	__s32 arg0;
@@ -178,6 +192,7 @@ struct event_config {
 	 */
 	__u32 policy_id;
 	__u32 flags;
+	struct config_btf_arg btf_arg[EVENT_CONFIG_MAX_ARG][MAX_BTF_ARG_DEPTH];
 } __attribute__((packed));
 
 #define MAX_ARGS_SIZE	 80
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -400,6 +400,108 @@ The `maxData` flag does not work with `returnCopy` flag at the moment, so it's
 usable only for syscalls/functions that do not require return probe to read the
 data.
 
+### Attribute resolution
+
+{{< caution >}}
+- For kprobes, available only from kernel version 5.4.
+- For LSM, available only from kernel version 5.7.
+- For uprobes, this functionality is not supported.
+{{< /caution >}}
+
+This functionality allows you to dynamically extract specific attributes from
+kernel structures passed as parameters to Kprobes and LSM hooks.
+For example, when using the `bprm_check_security` LSM hook, you can access and
+display attributes from the parameter `struct linux_binprm *bprm` at index 0.
+This parameter contains many informations you may want to access.
+With the resolve flag, you can easily access fields such as
+`mm.owner.real_parent.comm`, which provides the parent process comm.
+
+#### How it works
+
+The following tracing policy demonstrates how to use the resolve flag to extract the
+parent process's comm during the execution of a binary:
+
+```yaml
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm"
+spec:
+  lsmhooks:
+  - hook: "bprm_check_security"
+    args:
+    - index: 0 # struct linux_binprm *bprm
+      type: "string"
+      resolve: "mm.owner.real_parent.comm"
+    selectors:
+    - matchActions:
+      - action: Post
+```
+- `index` flag : The parameter at index 0 is a pointer to the
+`struct linux_binprm`.
+- `resolve` flag : Using the resolve flag, the policy extracts the
+`mm.owner.real_parent.comm` field, representing the parent process's comm.
+
+{{< caution >}}
+- This feature requires you to read the kernel structure definitions to find what you're
+looking for in the hook parameter attributes. For instance, if you want to have a look
+at what is available inside `struct linux_binprm`, take a look at its definition in
+[include/linux/binfmts.h](https://elixir.bootlin.com/linux/v6.12.5/source/include/linux/binfmts.h#L18)
+- Some structures are dynamic. This means that they may change at runtime.
+{{< /caution >}}
+
+Tetragon can also handle some structures such as `struct file` or `struct
+path` and a few others. This means you can also extract the whole struct, if it is
+available in the attributes of the parameter, and set the type with the correct type
+like this :
+
+```yaml
+...
+  lsmhooks:
+  - hook: "bprm_check_security"
+    args:
+      - index: 0 # struct linux_binprm *bprm
+        type: "file"
+        resolve: "file"
+...
+```
+Or
+```yaml
+...
+  lsmhooks:
+  - hook: "bprm_check_security"
+    args:
+      - index: 0
+        type: "path"
+        resolve: "file.f_path"
+...
+```
+
+The following tracing policy demonstrates how to use the `resolve` flag on nested
+pointers. For exemple, the hook `security_inode_copy_up` (defined in
+[security/security.c](https://elixir.bootlin.com/linux/v6.12.5/source/security/security.c#L2753))
+takes two parameters: `struct dentry *src` and `struct cred **new`. The resolve
+flag allows you to access fields within these nested structures transparently.
+
+```yaml
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm"
+spec:
+  kprobes:
+  - call: "security_inode_copy_up"
+    args:
+    - index: 0 # struct dentry *src
+      type: "int64"
+    - index: 1 # struct cred **new
+      type: "int"
+      resolve: "user.uid.val"
+    selectors:
+    - matchActions:
+      - action: Post
+```
+
 ## Return values
 
 A `TracingPolicy` spec can specify that the return value should be reported in
--- a/examples/tracingpolicy/lsm_track_grandparent.yml
+++ b/examples/tracingpolicy/lsm_track_grandparent.yml
@@ -1,0 +1,14 @@
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm"
+spec:
+  lsmhooks:
+  - hook: "bprm_check_security"
+    args:
+      - index: 0
+        type: "string"
+        resolve: "mm.owner.real_parent.real_parent.comm"
+    selectors:
+      - matchActions:
+        - action: Post
--- a/pkg/api/tracingapi/client_kprobe.go
+++ b/pkg/api/tracingapi/client_kprobe.go
@@ -597,17 +597,25 @@ type KprobeArgs struct {
 	Args4 []byte
 }
 
+type ConfigBtfArg struct {
+	Offset        uint32 `align:"offset"`
+	IsPointer     uint16 `align:"is_pointer"`
+	IsInitialized uint16 `align:"is_initialized"`
+}
+
 const EventConfigMaxArgs = 5
+const MaxBtfArgDepth = 10 // Artificial value for compilation, may be extended
 
 type EventConfig struct {
-	FuncId          uint32                     `align:"func_id"`
-	Arg             [EventConfigMaxArgs]int32  `align:"arg0"`
-	ArgM            [EventConfigMaxArgs]uint32 `align:"arg0m"`
-	ArgTpCtxOff     [EventConfigMaxArgs]uint32 `align:"t_arg0_ctx_off"`
-	Syscall         uint32                     `align:"syscall"`
-	ArgReturnCopy   int32                      `align:"argreturncopy"`
-	ArgReturn       int32                      `align:"argreturn"`
-	ArgReturnAction int32                      `align:"argreturnaction"`
-	PolicyID        uint32                     `align:"policy_id"`
-	Flags           uint32                     `align:"flags"`
+	FuncId          uint32                                           `align:"func_id"`
+	Arg             [EventConfigMaxArgs]int32                        `align:"arg0"`
+	ArgM            [EventConfigMaxArgs]uint32                       `align:"arg0m"`
+	ArgTpCtxOff     [EventConfigMaxArgs]uint32                       `align:"t_arg0_ctx_off"`
+	Syscall         uint32                                           `align:"syscall"`
+	ArgReturnCopy   int32                                            `align:"argreturncopy"`
+	ArgReturn       int32                                            `align:"argreturn"`
+	ArgReturnAction int32                                            `align:"argreturnaction"`
+	PolicyID        uint32                                           `align:"policy_id"`
+	Flags           uint32                                           `align:"flags"`
+	BtfArg          [EventConfigMaxArgs][MaxBtfArgDepth]ConfigBtfArg `align:"btf_arg"`
 }
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -115,6 +115,32 @@ func LoadBTF() (*btf.Spec, error) {
 	return NewBTF()
 }
 
+func FindBtfFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
+	var hookFn *btf.Func
+	spec, err := LoadBTF()
+	if err != nil {
+		return nil, err
+	}
+
+	if err = spec.TypeByName(hook, &hookFn); err != nil {
+		return nil, fmt.Errorf("failed to find BTF type for hook %q: %w", hook, err)
+	}
+
+	btfHookProto, isBtfFuncProto := hookFn.Type.(*btf.FuncProto)
+	if !isBtfFuncProto {
+		return nil, fmt.Errorf("hook %q has no BTF type FuncProto", hook)
+	}
+	paramLen := len(btfHookProto.Params)
+	if argIndex > paramLen-1 {
+		parameter := "parameter"
+		if paramLen > 1 {
+			parameter += "s"
+		}
+		return nil, fmt.Errorf("index %d is out of range. The hook only have %d %q", argIndex, paramLen, parameter)
+	}
+	return &btfHookProto.Params[argIndex], nil
+}
+
 func ResolveNestedTypes(ty btf.Type) btf.Type {
 	switch t := ty.(type) {
 	case *btf.Restrict:
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -219,6 +219,8 @@ func processMembers(
 	if t, ok := currentType.(*btf.Pointer); ok {
 		(*btfArgs)[i].IsPointer = uint16(1)
 		currentType = t.Target
+	} else if _, ok := currentType.(*btf.Int); ok {
+		(*btfArgs)[i].IsPointer = uint16(1)
 	}
 	return &currentType, nil
 }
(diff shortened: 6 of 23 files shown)


bpf/process/generic_calls.h
```diff
@@ -89,6 +101,25 @@ generic_process_event(void *ctx, struct bpf_map_def *heap_map,
 	a = (&e->a0)[index];
 	total = e->common.size;
 
+	if (index <= 4) {
```
olsajiri, line 104: all the new extra processing should be optional, now we do that always.. maybe extra bool in event_config?

bpf/process/generic_calls.h
```diff
@@ -89,6 +101,25 @@ generic_process_event(void *ctx, struct bpf_map_def *heap_map,
...
+#ifndef __V61_BPF_PROG
+#pragma unroll
+		for (int i = 0; i < MAX_BTF_ARG_CHILD_DEPTH; ++i) {
+			if (extract_arg_depth(i, &extract_data))
+				break;
+		}
+#else
+		loop(MAX_BTF_ARG_CHILD_DEPTH, extract_arg_depth, &extract_data, 0);
+#endif /* __V61_BPF_PROG */
+		probe_read((void *)&a, sizeof(char *), (char *)a);
```
olsajiri, line 120: what's the purpose of this?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -758,6 +759,34 @@ func addKprobe(funcName string, instance int, f *v1alpha1.KProbeSpec, in *addKpr
...
+				return errFn(err)
+			}
+
+			if a.OverwriteType != "" {
+				argType = gt.GenericTypeFromString(a.OverwriteType)
+			} else {
+				argType = gt.GenericTypeFromBTF(*lastChild)
+			}
+		}
+
```
olsajiri, line 789: I think we should put this in function and use it from all the generic probes config code, now we just duplicate the code

olsajiri (commented):
I like it, left some comments

thanks for splitting the code in multiple logical changes, it's rare ;-)

we were originally thinking of using C expression parsing (there's some go lib for that) but I think that can be done later if it's ever needed, the basic parsing you did should be fine

please add tests, would be great to have some framework where we could easily add various 'ExtractParam' expressions for testing

bpf/process/types/basic.h
```diff
@@ -150,6 +150,19 @@ struct selector_arg_filters {
 	__u32 argoff[5];
 } __attribute__((packed));
 
+struct config_btf_arg_child {
+	__u8 offset;
```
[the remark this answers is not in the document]
ScriptSathi: When I was doing the POC for this I could use u32 and the verifier does not complain, but when i put this in tetragon it does, so I changed it to u8. Though I've seen this function `args_off` that format the int to 14 bits. So maybe the verifier let do math pointer until 1600 ? 
I will make more tests with u16 to push the verifier to the limits to try.

pkg/sensors/tracing/generickprobe.go
ScriptSathi: Yes ! I was thinking about this because it is very ugly this way. But after that, I though maybe there are things that can be mutualized between `generickprobe.go` and `genericlsm.go`. So I put the ugly code like this and later we do refacto.
But if you mention this, I'll start just for this haha

bpf/process/generic_calls.h
ScriptSathi: This is needed because pointer are dereferenced on the next iteration when called, not when we do math ptr.
To give more context, let's assume we have linux_binprm and want to follow this `linux_binprm.file.my_var` with `my_var` as string and `file` as `struct file *`. The `is_pointer` check in the loop is not done at `file` iteration but on the next with `my_var`. In order to get this pointer delay back, we make sure `my_var` is safe ptr.

In the tetragon logs it look like this

With this final `probe_read`
```json
{
  "process_lsm": {
    "args": [
      {
        "string_arg": "ls"
      }
    ],
  },
}
```

Without it
```json
    {
  "process_lsm": {
    "args": [
      {
        "string_arg": "xB�O�=�O�ls"
      }
    ],
  },
}
```

bpf/process/generic_calls.h
```diff
@@ -89,6 +101,25 @@ generic_process_event(void *ctx, struct bpf_map_def *heap_map,
 	a = (&e->a0)[index];
 	total = e->common.size;
 
+	if (index <= 4) {
+		btf_config = (&config->btf_arg_child0)[index];
+		struct extract_arg_data extract_data = {
+			.config = config,
```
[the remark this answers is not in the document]
ScriptSathi: `loop(MAX_BTF_ARG_CHILD_DEPTH, extract_arg_depth, &extract_data, 0);` 
loop takes only 4 params. I do not see how we could do it ? 
Another solution could be nested callback, but not sure it is better
Example :
`loop(MAX_BTF_ARG_CHILD_DEPTH, extract_arg_depth(btf_config), (void **)&a, 0);`
with `extract_arg_depth` that take `bpf_config` and return a function compatible with `loop`. But this look ugly as well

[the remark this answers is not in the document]
olsajiri: ugh, ok.. missed that, it's ok

examples/tracingpolicy/lsm_track_grandparent.yml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: "lsm"
+spec:
+  lsmhooks:
+  - hook: "bprm_check_security"
+    args:
+      - index: 0
+        type: "linux_binprm"
+        extractParam: "mm.owner.real_parent.real_parent.comm"
+        overwriteType: "string"
```
olsajiri, line 12: do we need overwriteType? could we maybe just do something like:

```
 - index: 0
   type: "string"
   extractParam: "linux_binprm.file.f_path.dentry.d_name.name"
```
the overwrite stuff might cause headaches later

also maybe rename `extractParam` to `resolve`? not sure ;-) 

cc @kkourt

bpf/process/types/basic.h
```diff
@@ -172,6 +185,11 @@ struct event_config {
 	 */
 	__u32 policy_id;
 	__u32 flags;
+	struct config_btf_arg btf_arg0[MAX_BTF_ARG_DEPTH];
+	struct config_btf_arg btf_arg1[MAX_BTF_ARG_DEPTH];
+	struct config_btf_arg btf_arg2[MAX_BTF_ARG_DEPTH];
+	struct config_btf_arg btf_arg3[MAX_BTF_ARG_DEPTH];
+	struct config_btf_arg btf_arg4[MAX_BTF_ARG_DEPTH];
```
olsajiri, line 192: why not follow the user space part and have just `btf_arg[5][MAX_BTF_ARG_DEPTH]`
I did not see `btf_arg[1-4]` being touched elsewhere..

olsajiri (changes requested):
looks great, I left some comments, thanks

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -400,6 +400,61 @@ The `maxData` flag does not work with `returnCopy` flag at the moment, so it's
 usable only for syscalls/functions that do not require return probe to read the
 data.
 
+### Advanced usage
+
+For specific use cases, you may want to extract a specific attribute from the argument.
+For instance you have `struct linux_binprm` as first argument and want to filter parent
+process name, you can do it as following.
```
mtardy, line 407:
```suggestion
You may want to extract a specific attribute from the argument structure. For instance,
you are hooking a function with `struct linux_binprm` as the first argument and want to
filter the parent process' name, you can do it as follows.
```

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -400,6 +400,61 @@ The `maxData` flag does not work with `returnCopy` flag at the moment, so it's
 usable only for syscalls/functions that do not require return probe to read the
 data.
 
+### Advanced usage
```
mtardy, line 403: I would maybe be more specific on this one, like "extracting structure fields/attribute", whatever you prefer that describes best this feature.

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -400,6 +400,61 @@ The `maxData` flag does not work with `returnCopy` flag at the moment, so it's
...
+        overwriteType: "string"
+    selectors:
+      - matchActions:
+        - action: Post
+```
+
+The above policy will display the parent process name every time the hook is called.
+The `extractParam` field is used to reach a specific data into the `struct
+linux_binprm`. It is important to set `overwriteType` as well to make sure the
+reached data is read correctly (as a string in this case).
```
mtardy, line 430:
```suggestion
The above policy will display the parent process's comm every time the hook is called.
The `extractParam` field is used to reach specific data into the `struct linux_binprm`. 
It is important to set `overwriteType` as well to make sure the reached data is read 
correctly (as a string in this case).
```

I would say `comm` maybe instead of name since this value can be changed by the thread itself, it's not a really safe way of identifying a process name.

examples/tracingpolicy/lsm_track_grandparent.yml
mtardy: I agree, but fyi it seems it's motivated by this https://github.com/cilium/tetragon/pull/3143/files#diff-3c0b10317f585c0016d0208630c3abbb0e8d39571e91254af88874710903a260R442-R455.

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -400,6 +400,61 @@ The `maxData` flag does not work with `returnCopy` flag at the moment, so it's
...
+
+{{< caution >}}
+- This feature requires you to know exactly what you are looking for in the attributes
+of the hook parameters. For instance, if you want to have a look on what is
+available inside `struct linux_binprm`, take a look at the
+[Bootlin website](https://elixir.bootlin.com/linux/v6.12.5/source/include/linux/binfmts.h#L18)
```
mtardy, line 436:
```suggestion
- This feature requires you to read the kernel structure definitions to find what you're
  looking for in the hook parameter attributes. For instance, if you want to have a look
  at what is available inside `struct linux_binprm`, take a look at its definition in
  [include/linux/binfmts.h](https://elixir.bootlin.com/linux/v6.12.5/source/include/linux/binfmts.h#L18)
```

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -400,6 +400,61 @@ The `maxData` flag does not work with `returnCopy` flag at the moment, so it's
...
+
+- Some structures are dynamic. This means that they may change at runtime. So you need to
+be aware of what you are looking for.
```
mtardy, line 439:
```suggestion
- Some structures are dynamic which means that they may change at runtime.
```

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -400,6 +400,61 @@ The `maxData` flag does not work with `returnCopy` flag at the moment, so it's
...
+{{< /caution >}}
+
+Tetragon can also handle some structures such as `struct file` or `struct
+path` and few others. This means you can also extract the whole struct, if it is
+available in the attributes of the parameter, and set the type with the correct type
+like this :
```
mtardy, line 445:
```suggestion
Tetragon can also handle some structures such as `struct file` or `struct
path` and a few others. This means you can also extract the whole struct if it is
available in the attributes of the parameter, and set the type with the correct type
like this :
```

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -400,6 +400,61 @@ The `maxData` flag does not work with `returnCopy` flag at the moment, so it's
...
+```yaml
+  - index: 0
+    type: "linux_binprm"
+    extractParam: "file"
+    overwriteType: "file"
+# Or
+    # extractParam: "file.f_path"
+    # overwriteType: "path"
+```
+
```
mtardy, line 456:
````suggestion
```yaml
- index: 0
  type: "linux_binprm"
  extractParam: "file"
  overwriteType: "file"
```
Or
```yaml
- index: 0
  extractParam: "file.f_path"
  overwriteType: "path"
```
````

Mmh not sure my suggestion is correct but I would split it in two.

**Review by mtardy — COMMENTED** [review]
I've been pulled-in for the docs change, here's a review. Thanks for adding docs!

File: pkg/sensors/tracing/generic_test.go [hunk 15]
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+// Copyright Orange
```
kkourt, line 3: We now have an AUTHORS file (https://github.com/cilium/tetragon/blob/main/AUTHORS) can you just use:
> // Copyright Authors of Tetragon

everywhere?

See https://github.com/cncf/foundation/blob/main/copyright-notices.md#copyright-notices for some context.

kkourt (changes requested):

olsajiri (approved):
looks good, one comment left, but could be fixed later if needed, thanks

bentekkie:
Hi, This feature is amazing, thanks for implementing it as it will open up usage of all the security_path_* kprobes that have dentry arguments, which is something I have been waiting for 😄 
Any idea when this will land?

olsajiri:
> Hi, This feature is amazing, thanks for implementing it as it will open up usage of all the security_path_* kprobes that have dentry arguments, which is something I have been waiting for 😄 Any idea when this will land?

I think @kkourt wanted to check that before merging

pkg/btf/btf.go
```diff
@@ -101,3 +104,118 @@ func InitCachedBTF(lib, btf string) error {
...
+		return processMembers(btfArgs, currentType, t.Members, pathToFound, i)
+	case *btf.Union:
+		return processMembers(btfArgs, currentType, t.Members, pathToFound, i)
+	case *btf.Pointer:
+		if len(pathToFound[i]) == 0 {
+			(*btfArgs)[i].IsPointer = uint16(1)
+			(*btfArgs)[i].IsInitialized = uint16(1)
+			return ResolveBtfPath(btfArgs, t.Target, pathToFound, i+1)
+		}
+		(*btfArgs)[i-1].IsPointer = uint16(1)
```
kkourt, line 185: Why is the above line needed?
As I commented in the bpf code, it's not clear to me what it means to have `IsPonter` equal to `false` but `IsIntialized` set to `true`.

pkg/btf/btf.go
```diff
@@ -216,6 +216,8 @@ func processMembers(
 	if t, ok := currentType.(*btf.Pointer); ok {
 		(*btfArgs)[i].IsPointer = uint16(1)
 		currentType = t.Target
+	} else if _, ok := currentType.(*btf.Int); ok {
+		(*btfArgs)[i].IsPointer = uint16(1)
```
kkourt, line 256: Any reason to not fold in this patch with the original version?

pkg/sensors/tracing/generic.go
```diff
@@ -0,0 +1,58 @@
...
+
+	rootType := param.Type
+	if rootTy, isPointer := param.Type.(*ebtf.Pointer); isPointer {
+		rootType = rootTy.Target
+	}
+
+	pathBase := strings.Split(arg.Resolve, ".")
+	path := addPaddingOnNestedPtr(rootType, pathBase)
+	if len(path) > api.MaxBtfArgDepth {
+		return nil, btfArg, fmt.Errorf("Unable to resolve \"%s\". The maximum depth allowed is %d", arg.Resolve, api.MaxBtfArgDepth)
```
kkourt, line 41: nit: you use `%q` instead of `\"%s\"`.

pkg/sensors/tracing/genericlsm.go
```diff
@@ -239,6 +239,12 @@ func addLsm(f *v1alpha1.LsmHookSpec, in *addLsmIn) (id idtable.EntryID, err erro
 		argType := gt.GenericTypeFromString(a.Type)
 
 		if a.Resolve != "" && j < api.EventConfigMaxArgs {
+			if !kernels.MinKernelVersion("5.8.0") {
+				// bpf_lsm_<hook_name> does not exist in BTF file before 5.8
+				// https://lore.kernel.org/bpf/20200329004356.27286-4-kpsingh@chromium.org/
+				return errFn(fmt.Errorf("Error: LSM programs could not use Resolve flag for your kernel version." +
+					"Please use Kprobe instead or update your kernel to version 5.8 or higher"))
+			}
```
kkourt, line 247: Checking kernel versions does not always work as expected because there are kernels with older versions that backport a lot of functionality.

Am I correct to assume that the `resolveBtfArg` would fail anyway? If so, I would suggest something like:
```
			lastBtfType, btfArg, err := resolveBtfArg("bpf_lsm_"+f.Hook, a)
                        if err != nil {
                                        return errFn(fmt.Errorf("could not resolve  path from btf. You can try to use a kprobe or update your kernel version to 5.8 or later"))
                        }
```

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -400,6 +400,102 @@ The `maxData` flag does not work with `returnCopy` flag at the moment, so it's
 usable only for syscalls/functions that do not require return probe to read the
 data.
 
+### Attribute resolution
```
kkourt, line 403: ❤️

kkourt (commented):
Thanks! This PR looks fantastic!

Please find some comments below. Other than a few minor things, I think we are good to merge this. Some of the comments refer to additional work, but I would suggest to leave this as follow-ups since the PR is already a lot of commits.

mtardy (approved):
Thanks, doc addition now looks good :)

bpf/process/generic_calls.h
```diff
@@ -62,11 +62,23 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
...
 }
 
+FUNC_INLINE int
+extract_arg_depth(u32 i, struct extract_arg_data *data)
+{
+	if (i >= MAX_BTF_ARG_DEPTH || !data->btf_config[i].is_initialized)
+		return 1;
+	*data->arg = *data->arg + data->btf_config[i].offset;
+	if (data->btf_config[i].is_pointer)
+		probe_read((void *)data->arg, sizeof(char *), (void *)*data->arg);
```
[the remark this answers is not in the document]
kkourt: > - When accessing `task_struct->comm` it is not needed since `comm` is not a pointer

https://elixir.bootlin.com/linux/v6.13-rc3/source/include/linux/sched.h#L1138:
> `char				comm[TASK_COMM_LEN];`

I'm sorry, but I still don't understand how things are supposed to work without pointers.
Can you provide an example of a hook and a "resolve:" field?

ScriptSathi:
I rebased from main to fix the CI issues after GitHub update (resolved with https://github.com/cilium/tetragon/pull/3387), but it seems `veristat` tests are still in error because of that (https://github.com/cilium/tetragon/actions/runs/13269998947/job/37046941582?pr=3143 and https://github.com/cilium/tetragon/actions/runs/13269998979/job/37046941723?pr=3143)

kkourt:
> I rebased from main to fix the CI issues after GitHub update (resolved with #3387), but it seems `veristat` tests are still in error because of that (https://github.com/cilium/tetragon/actions/runs/13269998947/job/37046941582?pr=3143 and https://github.com/cilium/tetragon/actions/runs/13269998979/job/37046941723?pr=3143)

I think the error is unrelated to this PR:
> Getting action download info
Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: 704facf57e6136b1bc63b828d79edcd491f0ee84`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down

kkourt:
> I rebased from main to fix the CI issues after GitHub update (resolved with #3387), but it seems `veristat` tests are still in error because of that (https://github.com/cilium/tetragon/actions/runs/13269998947/job/37046941582?pr=3143 and https://github.com/cilium/tetragon/actions/runs/13269998979/job/37046941723?pr=3143)

@mtardy AFAICT, even with the setup-bpftool update (https://github.com/mtardy/setup-bpftool/pull/7) being picked by renovate in tetragon (https://github.com/cilium/tetragon/pull/3387), the error remains?

In any case, the error seems unrelated to this PR.

mtardy:
> > I rebased from main to fix the CI issues after GitHub update (resolved with #3387), but it seems `veristat` tests are still in error because of that (https://github.com/cilium/tetragon/actions/runs/13269998947/job/37046941582?pr=3143 and https://github.com/cilium/tetragon/actions/runs/13269998979/job/37046941723?pr=3143)
> 
> @mtardy AFAICT, even with the setup-bpftool update ([mtardy/setup-bpftool#7](https://github.com/mtardy/setup-bpftool/pull/7)) being picked by renovate in tetragon (#3387), the error remains?
> 
> In any case, the error seems unrelated to this PR.

Sorry, it's my fault, you need https://github.com/cilium/tetragon/pull/3389 in.

pkg/api/tracingapi/client_kprobe.go
```diff
@@ -574,17 +574,25 @@ type KprobeArgs struct {
...
 }
 
+type ConfigBtfArg struct {
+	Offset        uint32 `align:"offset"`
+	IsPointer     uint16 `align:"is_pointer"`
+	IsInitialized uint16 `align:"is_initialized"`
+}
+
 const EventConfigMaxArgs = 5
+const MaxBtfArgDepth = 10 // Artificial value for compilation, may be extended
```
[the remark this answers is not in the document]
kkourt: Ah, I see what you mean.

When two values between bpf and user-space need to match, and there is no way to ensure they do, adding a comment helps so that folk know that if they need to change one, they need to change the other.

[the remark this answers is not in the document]
ScriptSathi: Sure ! I'll push this changes along with your additional comments if needed

bpf/process/generic_calls.h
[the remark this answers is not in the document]
kkourt: Thanks. I played with a few examples, and I think I understand now.

pkg/btf/btf.go
```diff
@@ -115,6 +115,32 @@ func LoadBTF() (*btf.Spec, error) {
 	return NewBTF()
 }
 
+func FindBtfFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
+	var hookFn *btf.Func
+	spec, err := LoadBTF()
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
kkourt: I would argue that for testing, it is much better/easier to avoid global state. The same can be said beyond testing of course, but this is a different dicussion. 

Concretely, if we wanted to test multiple files within a single test, we would have to change the global variable adn then perform the test. Having a function that accepts a btf file as an argument makes this easier: we just call the function with the proper argument. The function that uses the global btf file can be a simple wrapper over the function that does.

pkg/btf/btf.go
kkourt: I see. Thanks!

I ran a few examples, and I think I understand it better now.
One thing I was not able to figure out is when the "len(pathToFound[i]) == 0" condition will trigger.
Can you provide an example?

pkg/sensors/tracing/genericlsm.go
kkourt: One thing to note is that there are vendor kernels (e.g., rhel) that do not follow kernel.org kernels. So a kernel might report it's 4.18, for example, but have a number of patches that are not part of the 4.18 kernel. This is why, in many cases, kernel versions do not tell the full story and it's better to test whether a feature is available or not.

I think an error message like:
> fmt.Errorf("failed to find BTF type for hook %q: %w. Please check if the hook exists or if your kernel supports BTF for lsm hooks. As an alternative, consider switching to kprobes. ", hook, err)

is perfectly find, but either approach works.

kkourt (approved):
Thanks, LGTM!

This is a great new feature!

pkg/btf/btf.go
ScriptSathi: Yes, this condition mostly matches the "padding" system. This is the way I found to handle nested pointer such as `struct cred** new` https://elixir.bootlin.com/linux/v6.13-rc3/source/security/security.c#L2751

So in case of nested pointers, the path will look like this
["",..,"","my","path"]

So since current path is "", this means that the pointer is not the previous type but the current. So I add this specific use case.

ScriptSathi:
Thanks a lot for your precious reviews and time spend on it @olsajiri, @kkourt and @mtardy  🙏

PR #3146, opened 2024-11-20
cgrouprate: refactor code for registering into base

kkourt:
Instead of using base objects directly, it's cleaner to operate on the contents of the sensor object passed at registration time. Add some helper functions to identify these objects


pkg/cgrouprate/cgrouprate.go
```diff
@@ -117,6 +131,11 @@ func (r *CgroupRate) process(ctx context.Context) {
 	r.log.Infof("Cgroup rate started (%d/%s)",
 		r.opts.Events, time.Duration(r.opts.Interval).String())
 
+	defer func() {
+		// cleanup
+		glSt.handle = nil
+	}()
+
```
olsajiri, line 138: hum, why is this needed?

pkg/cgrouprate/cgrouprate.go
```diff
@@ -44,12 +47,19 @@ const (
...
 	cleanupInactiveTime = time.Minute
+	cgRateMaxEntries    = 32768 // this value could be fine tuned
 )
 
-var (
-	handle     *CgroupRate
-	handleLock sync.RWMutex
-)
+type globalState struct {
+	cgRmdirProg      *program.Program
```
olsajiri, line 54: does this prog need to be in the state struct? doesn't looks like it's used

pkg/cgrouprate/cgrouprate.go
```diff
@@ -280,27 +299,76 @@ func Check(kube *processapi.MsgK8s, ktime uint64) {
...
+	for _, p := range sensor.Progs {
+		if base.IsExecve(p) || base.IsFork(p) || base.IsExit(p) {
+			rateProgs = append(rateProgs, p)
+		}
+		if base.IsExecve(p) {
+			optsProgs = append(optsProgs, p)
+		}
+	}
+
+	if len(optsProgs) <= 0 {
```
olsajiri, line 350: should we check both opts and rate? len(xxx) can't be < 0 right?

pkg/cgrouprate/cgrouprate.go
kkourt: We can remove it if we want. I just added it because there used to be a global variable with it.

pkg/cgrouprate/cgrouprate.go
kkourt: So the new code has a check that glSt is not nil when we register:

```
	if glSt.handle != nil {
		return nil, fmt.Errorf("cgrouprate: handle is already set, need to cleanup first")
	}
```

This was failing for TestThrottle2 so I added the code above. The idea is that if the goroutine goes away, we can cleanup handle. Alternatively, I could also remove the above check.

pkg/cgrouprate/cgrouprate.go
kkourt: Yap, will fix.

pkg/cgrouprate/cgrouprate.go
olsajiri: right, but looks like only because we initialize it and used it on different places, I'd remove it to prevent future headaches

olsajiri (approved):

pkg/cgrouprate/cgrouprate.go
kkourt: OK, removed. Thanks!

Issue #3148, opened 2024-11-20
[fork_test.go:55] Return error <nil>


ScriptSathi:
### What happened?

While running `make test` on a PR, the tests crashing at fork_test.go:55 with the following line

```
fork_test.go:55: command failed with exit status 1. Context error: <nil>
```

https://github.com/cilium/tetragon/blob/main/pkg/sensors/exec/fork_test.go#L55

This occurs because `ctx.Err()` is displayed before it is done.

### Tetragon Version

git source

### More logs

```
...  
  logcapture.go:25: time="2024-11-21T12:56:40+01:00" level=info msg="Listening for events..."
    cmd.go:100: stdout>: parent: (pid:86725, ppid:85754) starts
    cmd.go:100: stdout>: child 1 (pid:86726) exits
    cmd.go:100: stdout>: parent: (pid:86725, ppid:85754) starts
    cmd.go:100: stdout>: parent: (pid:86725) child (86726) exited with: 0
    cmd.go:100: stderr>: got an alarm, bailing out
    cmd.go:100: stderr>: giving up on waiting our parent to die
    cmd.go:100: stdout>: parent: (pid:86725, ppid:85754) starts
    fork_test.go:55: command failed with exit status 1. Context error: <nil>
    logcapture.go:25: time="2024-11-21T12:57:10+01:00" level=info msg="Listening for events completed." error="context canceled"
...
```

PR #3149, opened 2024-11-20
fix(#3148) Update fork-tester.c replace systemd pid with parent process pid 

ScriptSathi:
Fix https://github.com/cilium/tetragon/issues/3148

### Description

A nil pointer is returned by `ctx.Err()` because `ctx` is not done when the error is displayed.

### More details

On some distributions, like Archlinux, systemd run with the flag `--user`. Thus, makes the user manager process a child of systemd process 1. So when you search for PPID, it will always be this child process.
According to the documentation below, systemd expose an environnement variable to help finding this children. So if we compare the value of this environment variable with the ppid, we should get the same result as doing `ppid == 1` 

https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#%24MANAGERPID

### DIFF
--- a/contrib/tester-progs/fork-tester.c
+++ b/contrib/tester-progs/fork-tester.c
@@ -2,99 +2,104 @@
 
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
-#include <sys/types.h>
-#include <sys/socket.h>
+#include <sys/prctl.h>
 #include <sys/wait.h>
-#include <arpa/inet.h>
-#include <fcntl.h>
+#include <unistd.h>
 #include <signal.h>
+#include <errno.h>
 
-// NB: global pipe for child 2 to notify parent that it has finished.
-int Pipe[2];
+#define TIMEOUT_SECONDS 30
 
-void client_run()
+volatile sig_atomic_t timeout_occurred = 0;
+
+void timeout_handler(int signo)
 {
-	// just print a message
-	printf("child 2 (pid:%d, ppid:%d) starts\n", getpid(), getppid());
-	printf("child 2 done\n");
-	return;
+	if (signo == SIGALRM) {
+		timeout_occurred = 1;
+	}
 }
 
-void child1()
+void child2(pid_t reaper_pid)
+{
+	pid_t initial_ppid = getppid();
+	pid_t pid = getpid();
+
+	printf("child 2 (pid:%d, ppid:%d) starts\n", pid, initial_ppid);
+
+	pid_t new_ppid;
+	for (int i = 0;; i++) {
+		new_ppid = getppid();
+		if (new_ppid == reaper_pid) {
+			break;
+		}
+		if (i == 30) {
+			fprintf(stderr, "giving up on waiting our parent to die\n");
+			exit(EXIT_FAILURE);
+		}
+		sleep(1);
+	}
+	printf("child 2 (pid:%d, ppid:%d) exits\n", pid, new_ppid);
+}
+
+void child1(pid_t ppid)
 {
 	pid_t pid;
 
-	if  ((pid = fork()) == -1) {
+	if ((pid = fork()) == -1) {
 		perror("fork");
 		exit(1);
 	} else if (pid == 0) {
-		for (int i=0; ;i++) {
-			int ppid = getppid();
-			if (ppid == 1) {
-				break;
-			}
-			if (i == 30) {
-				fprintf(stderr, "giving up on waiting our parent to die\n");
-				exit(1);
-			}
-			sleep(1);
-		}
-		client_run();
+		child2(ppid);
 	} else {
-		/* chilid 1 exits, after creating child 2 */
+		/* child 1 exits, after creating child 2 */
 		printf("child 1 (pid:%d) exits\n", getpid());
 		return;
 	}
-
 }
 
-void alarm_handler(int signum)
+void wait_children(void)
 {
-	fprintf(stderr, "got an alarm, bailing out\n");
-	exit(1);
-}
+	signal(SIGALRM, timeout_handler);
+	alarm(TIMEOUT_SECONDS);
 
-int main(int argc, char **argv)
-{
-	pid_t pid;
-
-	signal(SIGALRM, alarm_handler);
-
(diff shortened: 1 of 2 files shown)


contrib/tester-progs/fork-tester.c
```diff
@@ -33,6 +33,12 @@ void child1()
 			int ppid = getppid();
 			if (ppid == 1) {
 				break;
+			} else {
+				// If systemd run with --user
+				const char *manager_pid_str = getenv("MANAGERPID");
+				int manager_pid = atoi(manager_pid_str);
```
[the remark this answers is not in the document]
ScriptSathi: I thought about this, but since `atoi` can receive a NULL pointer, I was thinking to keep it basic this way. If `atoi` receive a NULL pointer or a string that could not be transformed to an integer, it returns 0 which is an invalid PID anyway.
But if you prefer the very explicit way, I have no problem adding those checks, what do you think ?

contrib/tester-progs/fork-tester.c
```diff
@@ -29,10 +29,20 @@ void child1()
...
 				break;
+			} else {
+				// If systemd run with --user
+				const char *manager_pid_str = getenv("MANAGERPID");
+				if (manager_pid_str) {
+					manager_pid = atoi(manager_pid_str);
+					if (ppid == manager_pid)
+						break;
+					manager_pid = 1;
+				}
```
kkourt, line 45: It's not that the forker-tester is a performance-critical program, but reading the environment variable on every iteration seems strange. Is there a reason for not reading it just once? You could also probably avoid the NULL check and just check whether atoi returns 0 and, if it does, just use pid=1.

ScriptSathi: You are right, I should not push code that late. Sorry, I hope the fix will be good enough

kkourt (changes requested):
Thanks! Can you please squash the two patches together?
I don't see a reason to keep them separate since the test will fail without the second patch.

contrib/tester-progs/fork-tester.c
```diff
@@ -2,99 +2,80 @@
...
-		/* chilid 1 exits, after creating child 2 */
+		/* child 1 exits, after creating child 2 */
 		printf("child 1 (pid:%d) exits\n", getpid());
 		return;
 	}
-
 }
 
-void alarm_handler(int signum)
+void wait_childrens(void)
```
kkourt, line 47:
```suggestion
void wait_children(void)
```

contrib/tester-progs/fork-tester.c
```diff
@@ -2,99 +2,80 @@
...
 
-int main(int argc, char **argv)
+int main(void)
 {
-	pid_t pid;
-
-	signal(SIGALRM, alarm_handler);
-
-	printf("parent: (pid:%d, ppid:%d) starts\n", getpid(), getppid());
+	pid_t fork_succeed;
```
kkourt, line 64:
```suggestion
	pid_t child_pid;
```

contrib/tester-progs/fork-tester.c
```diff
@@ -2,99 +2,80 @@
...
 
-	if (pipe2(Pipe, O_DIRECT) == -1) {
-		perror("pipe2");
-		exit(1);
-	}
+	printf("parent (pid:%d, ppid:%d) starts\n", pid, getppid());
+	prctl(PR_SET_CHILD_SUBREAPER, 1);
 
-	if  ((pid = fork()) == -1) {
+	if ((fork_succeed = fork()) == -1) {
```
kkourt, line 70:
```suggestion
	if ((child_pid = fork()) == -1) {
```

contrib/tester-progs/fork-tester.c
```diff
@@ -2,99 +2,80 @@
...
-
-	/* setup an alarm in case something goes wrong, and wait until the pipe
-	 * is closed. This will happen when all children terminate */
-	alarm(10);
-	char c;
-	int ret = read(Pipe[0], &c, 1);
-	if (ret < 0) {
-		perror("read");
-	}
+	wait_childrens();
```
kkourt, line 78:
```suggestion
	wait_children();
```

contrib/tester-progs/fork-tester.c
```diff
@@ -2,99 +2,80 @@
...
+		}
+	}
 }
 
-int main(int argc, char **argv)
+int main(void)
 {
-	pid_t pid;
-
-	signal(SIGALRM, alarm_handler);
```
kkourt, line 62: What's the reason that the `SIGALARM` was removed? It acted as a timeout.

contrib/tester-progs/fork-tester.c
```diff
@@ -2,99 +2,80 @@
...
+	while (1) {
+		int status;
+		int pid = wait(&status);
+		if (pid == -1) {
+			printf("parent (pid:%d) no more descendants\n", getpid());
+			break;
+		} else {
+			printf("parent (pid:%d) child (%d) exited with: %d\n", getpid(), pid,
+			       status);
+		}
```
kkourt, line 58: Can you add a sleep here? Having tight loops without sleeps where we print something can be problematic.

kkourt (changes requested):
Thanks!

Please find some additional comments.

contrib/tester-progs/fork-tester.c
ScriptSathi: I had no problem while testing because `wait` syscall is blocking as suggested in the documentation.
For me it is unnecessary to sleep. What kind of problem were you thinking of ?

```
# https://man7.org/linux/man-pages/man2/wait.2.html
If a child has already changed state, then these calls return immediately. 
Otherwise, they block until either a child changes state or a signal handler interrupts the call
```

kkourt: I see. Yap, that makes sense to me. Please ignore my comment above.

contrib/tester-progs/fork-tester.c
ScriptSathi: You are right, I forgot this one. I added it. 
For testing, if you put a sleep > than TIMEOUT_SECONDS before the wait, it will manually trigger the timeout.

PR #3159, opened 2024-11-24
tetragon: update bpf makefile

olsajiri:
bpf makefile changes, hopefully making it more readable

### DIFF
--- a/bpf/Makefile
+++ b/bpf/Makefile
@@ -67,7 +67,18 @@ CFLAGS_v53  = -D__LARGE_BPF_PROG
 CFLAGS_v511 = -D__LARGE_BPF_PROG -D__LARGE_MAP_KEYS
 CFLAGS_v61  = -D__LARGE_BPF_PROG -D__LARGE_MAP_KEYS -D__V61_BPF_PROG
 
+CFLAGS_bpf_enforcer.o           = -D__BPF_OVERRIDE_RETURN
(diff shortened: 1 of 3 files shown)


bpf/Makefile
```diff
@@ -88,104 +86,100 @@ $(eval $(call DEFINE_VARIANT,v511))
 $(eval $(call DEFINE_VARIANT,v61))
 
 # ALIGNCHECKER
-objs/%.ll: $(ALIGNCHECKERDIR)%.c
```
kkourt, line 92: I find the `.ll` step useful because it allows the inspection of the generated llvm bytecode. Can we keep it? Or at least have a way to generate llvm bytecode. 

That is, now I can  do:
```
make -C bpf objs/bpf_generic_kprobe.ll
llvm-dis bpf/objs/bpf_generic_kprobe.ll -o -
```

Would be nice to keep having something like the above.

kkourt (commented):
Nice cleanup! Left a comment below, but otherwise LGTM!

bpf/Makefile
olsajiri: if it's ok just to add target to generate .ll I'd just add that

kkourt: Yap, whatever works :)

kkourt (approved):
Thanks, LGTM

PR #3179, opened 2024-12-03
ProcessCache: Disable interval test in VM

kevsecurity:
The TestProcessCacheInterval test is flaky in VMs, probably because it relies on timing. This commit disables the test in the vmtests.


kkourt (approved):
lgtm, thanks!

mtardy:
These tests seem flaky in general. Can we make them better/remove them? https://github.com/cilium/tetragon/actions/runs/12203108081/job/34045336907?pr=3176#step:9:160

kevsecurity:
> These tests seem flaky in general. Can we make them better/remove them? https://github.com/cilium/tetragon/actions/runs/12203108081/job/34045336907?pr=3176#step:9:160

I'm open to ideas of how we test timing-related functionality, without waiting with sleeps which are inherently flaky on cloud runners and especially in VMs. Simulating the delay by running the ticker function isn't the same as the normal order of events and doesn't check that the ticker is actually working. But testing a ticker is flaky when timing is flaky.

PR #3181, opened 2024-12-03
[sensors] Add optional PostLoadHook

tpapagian:
This will provide the ability to run sensor-specific code right after loading all maps/programs.

This can be used for cleanup etc.


olsajiri (approved):
so now we just miss the pre-load hook ;-) looks good

PR #3186, opened 2024-12-05
exec: fix tracking of matchBinary children

kkourt:


tixxdz (approved):

bpf/lib/process.h
```diff
@@ -323,9 +323,22 @@ struct binary {
...
+FUNC_INLINE void
+binary_reset(struct binary *b)
+{
+	// buffer can be written at clone stage with parent's info, if previous path is longer than
+	// current, we can have leftovers at the end, so zero out bin structure.
+	//
+	// Do not zero the ->mb_bitset however, so that it can be inherited if exec() is called.
+	// This depends on ->mb_bitset being the last part of the struct.
+	memset(b, 0, sizeof(*b) - sizeof(b->mb_bitset));
+}
```
olsajiri, line 340: offsetof should be safer?

also this adds lot of instructions, we could add some generic memset function using the string_maps_ro_zero map .. could be follow up

kkourt: > offsetof should be safer?

You mean using something like `memset(b, 0, ofssetof(struct .., mb_bitset));`? Aren't both (sizeof and offsetof) computed in compile-time?

olsajiri: yes, but if there's alignment gaps in the struct (behind mb_bitset) you could end up on wrong offset?

tixxdz: (1) yes that could be problematic, why not embed the first part about paths on its struct bin_path inside bin and explicitly memset that.
(2) "// buffer can be written at clone stage with parent's info, if previous path is longer than
	// current, we can have leftovers at the end, so zero out bin structure."
	this comment is a bit cryptic and suggest an issue? thus explicitly having this separate could make it clear?

kkourt: > yes, but if there's alignment gaps in the struct (behind mb_bitset) you could end up on wrong offset?

OK, agreed: using offsetof is safer. Pushed a new version. Thanks.

olsajiri (approved):

PR #3197, opened 2024-12-08
tetragon: Remove not needed rule commands

olsajiri:
Keep just the dependency, the command is carried out by the default rule.


michi-covalent (approved):
looks innocent

kevsecurity (approved):

olsajiri:
> looks innocent

they always do ;-) thnx

PR #3205, opened 2024-12-09
BPF error metrics

kkourt:
Introduce bpf error metrics. See commits.

```
sudo ./tetra debug dump errmetrics                                                   
Location                        Error       Count
bpf/cgroup/bpf_cgtracker.c:64   E2BIG (7)   8
```

### DIFF
--- a/bpf/Makefile
+++ b/bpf/Makefile
@@ -120,15 +120,18 @@ CFLAGS_bpf_generic_lsm_output.o = -D__LARGE_BPF_PROG
 # Rules
 MTARGET_o  = $(patsubst $(DEPSDIR)%.d,$(OBJSDIR)%.o,$@)
 MTARGET_ll = $(patsubst $(DEPSDIR)%.d,$(OBJSDIR)%.ll,$@)
+MTARGET_i = $(patsubst $(DEPSDIR)%.d,$(OBJSDIR)%.i,$@)
 
 rule_d_CFLAGS  = $(CFLAGS_$(notdir $(MTARGET_o)))
 rule_o_CFLAGS  = $(CFLAGS_$(notdir $@))
 rule_ll_CFLAGS = $(CFLAGS_$(notdir $(patsubst %.ll,%.o,$@)))
 
 rule_d  = $(CLANG) $(CLANG_FLAGS) $(rule_d_CFLAGS) $1 -MM -MP -MT $(MTARGET_o)  $< >  $@ && \
-          $(CLANG) $(CLANG_FLAGS) $(rule_d_CFLAGS) $1 -MM -MP -MT $(MTARGET_ll) $< >> $@
+          $(CLANG) $(CLANG_FLAGS) $(rule_d_CFLAGS) $1 -MM -MP -MT $(MTARGET_ll) $< >> $@ && \
+          $(CLANG) $(CLANG_FLAGS) $(rule_d_CFLAGS) $1 -MM -MP -MT $(MTARGET_i) $< >> $@
 rule_o  = $(CLANG) $(CLANG_FLAGS) $(rule_o_CFLAGS) $1 -c $< -o $@
 rule_ll = $(CLANG) $(CLANG_FLAGS) $(rule_ll_CFLAGS) -emit-llvm $1 -c $< -o $@
+rule_i  = $(CLANG) $(CLANG_FLAGS) $(rule_o_CFLAGS) $1 -E $< -o $@
 
 VARIANT_CFLAGS = $$(CFLAGS_$1)
 
@@ -148,6 +151,9 @@ $(DEPSDIR)bpf_multi_uprobe_$$(VAR).d: $(PROCESSDIR)bpf_generic_uprobe.c
 # Object build rule for VARIANT objects
 $(OBJSDIR)%_$$(VAR).o:
 	$$(call rule_o,$(VARIANT_CFLAGS))
+
+$(OBJSDIR)%_$$(VAR).i:
+	$$(call rule_i,$(VARIANT_CFLAGS))
 
 $(OBJSDIR)%_$$(VAR).ll:
 	$$(call rule_ll,$(VARIANT_CFLAGS))
@@ -172,6 +178,9 @@ $(eval $(call DEFINE_VARIANT,v61))
 # Object build rule for non VARIANT objects
 objs/%.o:
 	$(rule_o)
+
+objs/%.i:
+	$(rule_i)
 
 objs/%.ll:
 	$(rule_ll)
--- a/bpf/cgroup/bpf_cgtracker.c
+++ b/bpf/cgroup/bpf_cgtracker.c
@@ -6,6 +6,7 @@
 #include "bpf_helpers.h"
 #include "bpf_cgroup.h"
 #include "bpf_tracing.h"
+#include "bpf_errmetrics.h"
 #include "cgtracker.h"
 
 char _license[] __attribute__((section(("license")), used)) = "GPL";
@@ -60,7 +61,7 @@ tg_cgtracker_cgroup_mkdir(struct bpf_raw_tracepoint_args *ctx)
 		return 0;
 	cgid_tracker = map_lookup_elem(&tg_cgtracker_map, &cgid_parent);
 	if (cgid_tracker)
-		map_update_elem(&tg_cgtracker_map, &cgid, cgid_tracker, BPF_ANY);
+		map_update_elem__errmetrics(&tg_cgtracker_map, &cgid, cgid_tracker, BPF_ANY);
 
 	return 0;
 }
--- a/bpf/lib/bpf_errmetrics.h
+++ b/bpf/lib/bpf_errmetrics.h
@@ -5,6 +5,7 @@
 
 #include "compiler.h"
 
+// should match: pkg/errmetrics/map.go:MapKey
 struct errmetrics_key {
 	__u16 error;
 	__u8 file_id;
@@ -29,7 +30,7 @@ struct {
 	__uint(type, BPF_MAP_TYPE_LRU_PERCPU_HASH);
 	__uint(max_entries, 1024);
 	__type(key, struct errmetrics_key);
-	__type(value, __u32);
+	__type(value, __u32); // counter, should match pkg/errmetrics/map.go:MapVal
 } tg_errmetrics_map SEC(".maps");
 
 FUNC_INLINE void
--- a/bpf/lib/bpf_errmetrics.h
+++ b/bpf/lib/bpf_errmetrics.h
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* Copyright Authors of Tetragon */
+#ifndef BPF_ERRMETRICS_H__
+#define BPF_ERRMETRICS_H__
+
+#include "compiler.h"
+
+struct errmetrics_key {
+	__u16 error;
+	__u8 file_id;
+	__u8 pad1;
+	__u16 line_nr;
+	__u16 pad2;
+} __attribute__((packed));
+
+FUNC_INLINE __u16
+get_fileid__(const char *const fname)
+{
+#define fileid__(f, id)                  \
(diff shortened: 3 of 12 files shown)


cmd/bpf-fileids/fileids.go
```diff
@@ -0,0 +1,75 @@
...
+
+// json entry
+type Entry struct {
+	ID       int    `json:"id"`
+	Filename string `json:"filename"`
+}
+
+func initFileIDs(fname string) []Entry {
+
+	//entryRe := regexp.MustCompile(`fileid__\("([^"])+", *([0-9]+)"\)`)
```
mtardy, line 24:
```suggestion
```

bpf/Makefile
```diff
@@ -120,15 +120,18 @@ CFLAGS_bpf_generic_lsm_output.o = -D__LARGE_BPF_PROG
...
+MTARGET_i = $(patsubst $(DEPSDIR)%.d,$(OBJSDIR)%.i,$@)
 
 rule_d_CFLAGS  = $(CFLAGS_$(notdir $(MTARGET_o)))
 rule_o_CFLAGS  = $(CFLAGS_$(notdir $@))
 rule_ll_CFLAGS = $(CFLAGS_$(notdir $(patsubst %.ll,%.o,$@)))
 
 rule_d  = $(CLANG) $(CLANG_FLAGS) $(rule_d_CFLAGS) $1 -MM -MP -MT $(MTARGET_o)  $< >  $@ && \
-          $(CLANG) $(CLANG_FLAGS) $(rule_d_CFLAGS) $1 -MM -MP -MT $(MTARGET_ll) $< >> $@
+          $(CLANG) $(CLANG_FLAGS) $(rule_d_CFLAGS) $1 -MM -MP -MT $(MTARGET_ll) $< >> $@ && \
+          $(CLANG) $(CLANG_FLAGS) $(rule_d_CFLAGS) $1 -MM -MP -MT $(MTARGET_i) $< >> $@
```
olsajiri, line 131: I wasn't sure about appending .ll rules to deps file.. and now we have one more ;-) but I think it's fine, it should not grow further I guess

bpf/lib/bpf_errmetrics.h
```diff
@@ -0,0 +1,77 @@
...
+	} while (0)
+
+#define map_update_elem__errmetrics(m, k, v, f) ({         \
+	int err;                                           \
+	__u16 fileid = get_fileid__(__FILE__);             \
+                                                           \
+	if (!__builtin_constant_p(fileid) || !fileid)      \
+		compile_error(__FILE__);                   \
+	err = map_update_elem(m, k, v, f);                 \
+	if (err)                                           \
```
olsajiri, line 72: not sure it's good for anything and probably expensive, but perhaps it might be interesting to see the count of success cases as well?

olsajiri (commented):
nice feature, let's spread it over the code ;-) I also like the idea someone mentioned in the mtg that we could store the stats for tests and compare on each run

bpf/Makefile
kkourt: The other option would be to have explicit rules for .i objects, and I'm not sure how to do that if we want to generate them with the appropriate CFLAGS (e.g., -Dpizzaisthebest).

bpf/lib/bpf_errmetrics.h
```diff
@@ -0,0 +1,77 @@
...
+#define compile_error(f)                                                                     \
+	do {                                                                                 \
+		extern __attribute__((__error__(xerrstr(f)))) void compile_time_error(void); \
+		compile_time_error();                                                        \
+	} while (0)
+
+#define map_update_elem__errmetrics(m, k, v, f) ({         \
+	int err;                                           \
+	__u16 fileid = get_fileid__(__FILE__);             \
+                                                           \
```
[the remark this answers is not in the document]
kkourt: `fileid` should be computed at compile-time. Otherwise, the check below (`!__builtin_constant_p(fileid)`) should fail. So it should always be zero overhead.

bpf/lib/fileids.h
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+// Copyright Authors of Tetragon
+//
+// source code file to id mapping for metrics. See bpf_errmetrics.h
+// inspired by cilium's bpf metrics. See: https://github.com/cilium/cilium/pull/30972
+// NB: keep this file simple: only single-line comments and fileid__ entries
+
+fileid__("cgroup/bpf_cgtracker.c", 112)
```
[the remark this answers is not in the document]
kkourt: It's just random. I thought we might want to have `1` for bpf_execve.c or similar but maybe that doesn't make sense. I can change it.

bpf/cgroup/bpf_cgtracker.c
```diff
@@ -60,7 +61,7 @@ tg_cgtracker_cgroup_mkdir(struct bpf_raw_tracepoint_args *ctx)
 		return 0;
 	cgid_tracker = map_lookup_elem(&tg_cgtracker_map, &cgid_parent);
 	if (cgid_tracker)
-		map_update_elem(&tg_cgtracker_map, &cgid, cgid_tracker, BPF_ANY);
+		map_update_elem__errmetrics(&tg_cgtracker_map, &cgid, cgid_tracker, BPF_ANY);
```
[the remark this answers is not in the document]
kkourt: Yes, it definitely does.  I just added what was needed for this PR. Is it OK to do  the rest in followups?

pkg/errmetrics/map.go
```diff
@@ -0,0 +1,70 @@
...
+	"syscall"
+
+	"github.com/cilium/ebpf"
+	"golang.org/x/sys/unix"
+)
+
+type Map struct {
+	*ebpf.Map
+}
+
```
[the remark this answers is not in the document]
kkourt: fwiw, I've been using the same approach in other places.
 * https://github.com/cilium/tetragon/blob/a1205b06bd5b2fb2868013e24bbfcde91c12f062/pkg/policyfilter/map.go#L25-L28
 * https://github.com/cilium/tetragon/blob/a1205b06bd5b2fb2868013e24bbfcde91c12f062/pkg/cgtracker/cgtracker.go#L33-L36

It's basically an easy way to provide some helper methods so that the map can be used by the CLI (`tetra`)

bpf/lib/bpf_errmetrics.h
```diff
@@ -0,0 +1,77 @@
...
+	__u16 line_nr;
+	__u16 pad2;
+} __attribute__((packed));
+
+FUNC_INLINE __u16
+get_fileid__(const char *const fname)
+{
+#define fileid__(f, id)                  \
+	if (!__builtin_strcmp(f, fname)) \
+		return id;
```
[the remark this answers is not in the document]
kkourt: I don't know, but as I mentioned below, the check happens at compile-time, so there wouldn't be any code for this generated in the object file.

bpf/lib/bpf_errmetrics.h
kkourt: One challenge is that the map has 1024 entries (I wanted to keep it small) and LRU, which means that success cases will evict the failure cases, which would be problematic.

bpf/lib/bpf_errmetrics.h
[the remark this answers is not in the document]
mtardy: fyi I took a look out of curiosity, I would expect it to be not efficient at all on older kernel like `__builtin_memset` and maybe use the helper `bpf_strncmp` on newer? https://docs.ebpf.io/linux/helper-function/bpf_strncmp/

On my recent kernel (6+) it did a function call
```asm
; 	int ret = __builtin_strcmp(bin->path, mycmp);
     163:	r1 = r6
     164:	r2 = 0x0 ll
     166:	call -0x1
```
But failed to change the function call at loading
```
Call at insn 162: symbol "strcmp": unsatisfied program reference
```

bpf/lib/bpf_errmetrics.h
```diff
@@ -0,0 +1,76 @@
...
+
+#define map_update_elem__errmetrics(m, k, v, f) ({         \
+	int err;                                           \
+	__u16 fileid = get_fileid__(__FILE__);             \
+                                                           \
+	if (!__builtin_constant_p(fileid) || !fileid)      \
+		compile_error(__FILE__);                   \
+	err = map_update_elem(m, k, v, f);                 \
+	if (err)                                           \
+		errmetrics_update(-err, fileid, __LINE__); \
```
mtardy, line 73: Maybe that would be good to be able to disable the error handling feature? If I understand correctly the `fileid` part should be simplified as a constant by the compiler (otherwise you call the magical `compile_time_error` that will generate an error?). But treating the actual error will have an overhead as `errmetrics_update` will take some instructions, especially if we spread this error handling. We could do it as a follow up if needed I imagine!

bpf/lib/fileids.h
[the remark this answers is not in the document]
mtardy: Yep maybe pre-generating them would also simplify the generation of the JSON on the `tetra` side? Just checking all files under BPF. So that we can just run a simple check validating that all files are here and in the `.JSON` asking the user to add it manually or automatically with a tool if he created a new file (like we do for the SPDX-License-Identifier header).

pkg/sensors/base/base.go
```diff
@@ -68,6 +68,8 @@ var (
 	StatsMap           = program.MapBuilder("tg_stats_map", Execve)
 
 	MatchBinariesSetMap = program.MapBuilder(mbset.MapName, Execve)
+
+	ErrMetricsMap = program.MapBuilder("errmetrics_map", Execve)
```
mtardy, line 72: should we use the `tg` prefix for newer maps?

pkg/errmetrics/errmetrics.go
```diff
@@ -0,0 +1,39 @@
...
+	Filename string `json:"filename"`
+}
+
+//go:embed fileids.json
+var fileIDsJSON []byte
+
+func initFileIDs() map[int]string {
+	var entries []Entry
+	if err := json.Unmarshal(fileIDsJSON, &entries); err != nil {
+		log.Panic(err)
```
mtardy, line 28: I imagine it's expected that we potentially panic on loading this package?

pkg/errmetrics/errmetrics.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	ret := map[int]string{}
+	for _, e := range entries {
+		ret[e.ID] = e.Filename
+	}
+	return ret
+}
+
+var fileIDs = initFileIDs()
```
mtardy, line 39: why not exposing this as a function and return an error if the file is invalid? You want it to panic on load?

bpf/cgroup/bpf_cgtracker.c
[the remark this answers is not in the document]
olsajiri: sure, thanks

cmd/bpf-fileids/fileids.go
kkourt: fixed!

bpf/lib/bpf_errmetrics.h
kkourt: I'd rather not disable it. This is intended for cases where something went wrong so the overhead should not happen in normal situations. If it happens more frequently than expected, we can add a flag to disable it.

pkg/sensors/base/base.go
kkourt: sure, fixed.

pkg/errmetrics/errmetrics.go
kkourt: OK, fixed to be prettier :)
(I didn't think it mattered since i) the file is static and ii) this is only used by tetra and the utility that generates ids, but it was easy to fix so fixed)

olsajiri (approved):
checkpatch is complaining, other than that looks good

kkourt:
> checkpatch is complaining, other than that looks good

Yeah, it's unfortunate but there is no way  to have both clang-format and checkpatch happy 🤖

PR #3209, opened 2024-12-10
introduce in_init_tree flag for process events

will-isovalent:
This PR introduces a new flag for process events, `in_init_tree`, which is propagated from BPF via the execve map. The flag indicates whether a process is a member of its container's initial process tree, or whether it was spawned externally. For instance:

- a host process' `in_init_tree` value is always false.
- a container's entrypoint or any process spawned from it has an `in_init_tree` value of true.
- a process injected into the container (e.g. via `docker exec` or `kubectl exec` or `nsenter`) has an `in_init_tree` value of false

To support unit testing this feature, we also introduce a new export filter for container IDs and use it to match container IDs in the unit test. This avoids flooding the unit test output with spurious events from the host system.

### Changelog

```release-note
- Introduce the in_init_tree flag for process events which indicates whether a process spawned from its container's init process tree
- Introduce a container_id export filter
- Introduce an in_init_tree export filter
```


will-isovalent:
Whoops, bad find and replace messed up some variable names. Fixing.

michi-covalent (approved):
api change looks good ✅

pkg/sensors/exec/exec_test.go
```diff
@@ -626,6 +626,86 @@ func TestDocker(t *testing.T) {
...
+		t.Skipf("docker not available. skipping test: %s", err)
+	}
+
+	var doneWG, readyWG sync.WaitGroup
+	defer doneWG.Wait()
+
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
+	containerID := observertesthelper.DockerCreate(t, "--name", "in-init-tree-test", "bash", "bash", "-c", "sleep 1 & sleep infinity")
```
mtardy, line 640: out of curiosity why the sleep 1 in `sleep 1 & sleep infinity`?

pkg/sensors/exec/exec_test.go
```diff
@@ -626,6 +626,86 @@ func TestDocker(t *testing.T) {
...
+	obs, err := observertesthelper.GetDefaultObserver(t, ctx, tus.Conf().TetragonLib, observertesthelper.WithContainerId(containerID[:24]))
+	if err != nil {
+		t.Fatalf("GetDefaultObserver error: %s", err)
+	}
+
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+
+	readyWG.Wait()
+	observertesthelper.DockerStart(t, "in-init-tree-test")
+	time.Sleep(1 * time.Second)
```
mtardy, line 654: I admit sleeping is easier. If this is one day flaky (because downloading the image took long on creation), with `docker ps -a` you can check that the container was created (image was downloaded) or that it was started with `docker ps`.

mtardy (approved):
lgtm!

pkg/sensors/exec/exec_test.go
will-isovalent: hm yeah, we're already sleeping in other docker tests. I'm gonna go through and refactor them all at some point so let's just fix that up then.

pkg/sensors/exec/exec_test.go
will-isovalent: think this is left over from testing actually, let me remove it.

PR #3210, opened 2024-12-10
tetragon: setup to let match binary names use args as well

jrfastab:
Setting up ability to match args as well as binary names. This is useful for matching 'java beaches.jar' or 'python palmTree.py' where the binary itself is an interpretor and the actual thing being called is what matters.

### DIFF
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -277,6 +277,8 @@ struct heap_exe {
 	char end[STRING_POSTFIX_MAX_LENGTH];
 	__u32 len;
 	__u32 error;
+	__u32 arg_len;
+	__u32 arg_start;
 }; // All fields aligned so no 'packed' attribute.
 
 struct msg_execve_event {
@@ -322,6 +324,8 @@ struct binary {
 	char end[STRING_POSTFIX_MAX_LENGTH];
 	// STRING_POSTFIX_MAX_LENGTH reversed last bytes of the path
 	char end_r[STRING_POSTFIX_MAX_LENGTH];
+	// args for the binary
+	char args[MAXARGLENGTH];
 	// matchBinary bitset for binary
 	// NB: everything after and including ->mb_bitset will not be zeroed on a new exec. See
 	// binary_reset().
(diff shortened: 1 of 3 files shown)


michi-covalent (approved):

bpf/lib/process.h
```diff
@@ -326,6 +328,10 @@ struct binary {
 	// NB: everything after and including ->mb_bitset will not be zeroed on a new exec. See
 	// binary_reset().
 	mbset_t mb_bitset;
+	// length of the args stored in args
+	__s64 args_length;
+	// args for the binary
+	char args[MAXARGLENGTH];
```
kkourt, line 334: Can you please place these above `->mb_bitset`? See comment above. The idea is that we zero out the paths but not `->mb_bitset` on a new exec.

kkourt (changes requested):

bpf/lib/process.h
jrfastab: I can do we need to zero out the paths though? Here I just null terminated it.

bpf/lib/process.h
```diff
@@ -322,6 +324,10 @@ struct binary {
 	char end[STRING_POSTFIX_MAX_LENGTH];
 	// STRING_POSTFIX_MAX_LENGTH reversed last bytes of the path
 	char end_r[STRING_POSTFIX_MAX_LENGTH];
+	// length of the args stored in args
+	__s64 args_length;
```
kkourt, line 328: Is this used somewhere?

bpf/process/bpf_execve_event.c
```diff
@@ -399,6 +407,15 @@ execve_send(void *ctx __arg_ctx)
...
 		}
+
+		off = event->exe.arg_start & 0xff;
+		len = event->exe.arg_len & 0xff;
+		probe_read(curr->bin.args, len, (char *)&event->process + off);
+
+		nullone = len + 1;
+		nulltwo = len + 2;
+		curr->bin.args[nullone & 0xff] = 0x00; // null terminate string
+		curr->bin.args[nulltwo & 0xff] = 0x00; // null terminate string
```
kkourt, line 418: Why do we need two consecutive nulls?

kkourt (changes requested):

bpf/lib/process.h
jrfastab: seems not , forgot to remove it when  I decided null terminating was easier.

bpf/process/bpf_execve_event.c
jrfastab: Appears individual args are null separated. But its also the Unicode null.

kkourt (approved):

PR #3211, opened 2024-12-11
filters/cel: add k8s ip and cidr library functions

will-isovalent:
This simple PR adds CIDR and IP helpers from the k8s CEL libraries to our CEL filters.
This should enable users to write export filters in CEL that match IPs in kprobe events
using CIDR ranges etc.

For example, the following CEL expression written as a CEL filter would match kprobe
events where the first argument contains a socket with source address in CIDR range
10.0.0.0/16:

    cidr('10.0.0.0/16').containsIP(process_kprobe.args[0].sock_arg.saddr)

The specific documentation for these CEL libraries is available here:

- https://pkg.go.dev/k8s.io/apiserver/pkg/cel/library#CIDR
- https://pkg.go.dev/k8s.io/apiserver/pkg/cel/library#IP

```release-note
- add IP and CIDR helpers to CEL filters
```


michi-covalent (approved):
i like it

PR #3216, opened 2024-12-12
docs: use label match in SERVICECIDR command for KinD

### LINKED ISSUE #3182: docs: Network Monitoring SERVICECIDR command for KinD doesn't work for named clusters
In the section [Monitoring Kubernetes Network Access](https://tetragon.io/docs/getting-started/network/#monitoring-kubernetes-network-access), the command in the KinD tab for the SERVICECIDR env var assumes the KinD cluster is created without a name.  

To avoid the naming issue it should match on the kube-apiserver component label instead like:

```
export SERVICECIDR=$(kubectl describe pod -n kube-system -l component=kube-apiserver | awk -F= '/--service-cluster-ip-range/ {print $2; }')
```

tiswanso:
Fixes #3182

### Description

Use `component=kube-apiserver` match in doc command for getting the SERVICECIDR on KinD clusters

Named KinD clusters will have kube-apiserver pods with the cluster name included--the format is `kube-apiserver-<CLUSTERNAME>-control-plane`.  The kube-system control-plane uses the label `component=kube-apiserver` on the apiserver pods regardless of the cluster name.

### DIFF
--- a/docs/content/en/docs/getting-started/network.md
+++ b/docs/content/en/docs/getting-started/network.md
@@ -31,7 +31,7 @@ export SERVICECIDR=$(gcloud container clusters describe ${NAME} --zone ${ZONE} |
 {{< /tab >}}
 
 {{< tab Kind >}}
-export SERVICECIDR=$(kubectl describe pod -n kube-system kube-apiserver-kind-control-plane | awk -F= '/--service-cluster-ip-range/ {print $2; }')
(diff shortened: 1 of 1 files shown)


mtardy (approved):
Thanks! That's a great fix, it actually fix the situation in the case you used `make kind-setup` so that's cool.

Issue #3234, opened 2024-12-17
tetra  debug progs bug


panzhenyu12:
### What happened?

1.kernel version 5.15.0, tetragon version 1.30
2.tp config
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "file-tamper-delete"
spec:
  kprobes:
  - call: "security_path_unlink"
    syscall: false
    return: true
    args:
    - index: 0
      type: "path" 
    - index: 1
      type: "nop" 
    returnArg:
      index: 0
      type: "int"
    returnArgAction: "Post"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/home/appuser/test"          
      matchActions:
      - action: Override
        argError: -1
3. run tetra debug progs, show msg info for /sys/fs/bpf/tetragon/file-tamper-delete/generic_kprobe/security_path_unlink/link_override: argument list too long

### Tetragon Version

v1.3.0

### Kernel Version

Linux 5.15.0-71-generic

### Bugtool

time="2024-12-17T16:32:00+08:00" level=info msg="saving init info"
time="2024-12-17T16:32:00+08:00" level=info msg="retrieving lib directory" libDir=/usr/local/lib/tetragon/bpf/
time="2024-12-17T16:32:00+08:00" level=warning msg="not an object file, ignoring" path=/usr/local/lib/tetragon/bpf/
time="2024-12-17T16:32:03+08:00" level=warning msg="no btf filename in tetragon config, attempting to fall back to /sys/kernel/btf/vmlinux"
time="2024-12-17T16:32:04+08:00" level=info msg="btf file added" btfFname=/sys/kernel/btf/vmlinux
time="2024-12-17T16:32:04+08:00" level=info msg="tetragon log file added" exportFname=/var/log/tetragon/tetragon.log
time="2024-12-17T16:32:04+08:00" level=info msg="executed command" cmd=/usr/bin/dmesg dstFname=dmesg.out ret=0
time="2024-12-17T16:32:04+08:00" level=info msg="executed command" cmd="/usr/sbin/tc filter show dev lo ingress" dstFname=tc-info.lo.ingress ret=0
time="2024-12-17T16:32:04+08:00" level=info msg="executed command" cmd="/usr/sbin/tc filter show dev lo egress" dstFname=tc-info.lo.egress ret=0
[...]

PR #3235, opened 2024-12-17
tetra: Fix tetra debug progs fail

olsajiri:
Recent ebpf changes make ebpf.LoadPinnedProgram fail on opening link path. We want to open only programs, so failing on opening links/maps is fine ad we just want to continue scanning.

Fixes: https://github.com/cilium/tetragon/issues/3234

### DIFF
--- a/cmd/tetra/debug/progs.go
+++ b/cmd/tetra/debug/progs.go
@@ -93,6 +93,18 @@ func detectBpffs() (string, error) {
 	return "", fmt.Errorf("bpffs mount not found")
 }
 
+func detectLib() (string, error) {
+	paths := []string{"/var/lib/tetragon", "./bpf/objs/"}
+
+	for _, path := range paths {
+		if _, err := os.Stat(filepath.Join(path, "bpf_prog_iter.o")); err != nil {
+			continue
+		}
+		return path, nil
+	}
+	return "", fmt.Errorf("lib directory mount not found")
+}
+
 func NewProgsCmd() *cobra.Command {
 	cmd := cobra.Command{
 		Use:     "progs",
@@ -129,6 +141,12 @@ Examples:
 				}
 			}
 
+			if cfg.lib == "" {
+				if cfg.lib, err = detectLib(); err != nil {
+					log.Fatal(err)
+				}
+			}
+
 			if err = runProgs(ctx); err != nil {
 				log.Fatal(err)
 			}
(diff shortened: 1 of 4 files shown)


mtardy (commented):
fyi: https://github.com/cilium/tetragon/blob/e4f084f713c480387a0bd758c6bc1cf3e8b50c65/pkg/bugtool/maps.go#L163-L169

Commit: https://github.com/cilium/tetragon/commit/ca9aaa04ac0c0183a7bf94278dedb0971485cac0

mtardy:
Also added in the meantime, something we could actually use to do thing properly https://github.com/cilium/ebpf/pull/1626.

cmd/tetra/debug/progs.go
```diff
@@ -331,6 +332,9 @@ func getTetragonProgs(base string) ([]*prog, error) {
 			if finfo.IsDir() {
 				return nil
 			}
+			if strings.HasSuffix(path, "link") {
```
kkourt, line 414: Maybe Having "/link" is safer?
(suggestion based on observing what exists under /sys/fs/bpf/tetragon)

cmd/tetra/debug/progs.go
```diff
@@ -96,7 +138,7 @@ Examples:
 	flags := cmd.Flags()
 	flags.BoolVar(&cfg.all, "all", false, "Get all programs")
 	flags.StringVar(&cfg.lib, "bpf-lib", "bpf/objs/", "Location of Tetragon libs (btf and bpf files)")
-	flags.StringVar(&cfg.bpffs, "bpf-dir", "/sys/fs/bpf/tetragon", "Location of bpffs tetragon directory")
+	flags.StringVar(&cfg.bpffs, "bpf-dir", "detect", "Location of bpffs tetragon directory")
```
kkourt, line 160: How instead of "detect", we use ""? Would be nice to also update the help text with something like "if left empty, Tetragon will try to auto-detect".

edit: same for bpf-lib

kkourt (commented):
LGTM, but left some comments.

cmd/tetra/debug/progs.go
olsajiri: right it is.. also we need to add '/link_override' .. it's not failing on it, but to make the change consistent

cmd/tetra/debug/progs.go
olsajiri: ok

cmd/tetra/debug/progs.go
```diff
@@ -137,7 +155,7 @@ Examples:
 
 	flags := cmd.Flags()
 	flags.BoolVar(&cfg.all, "all", false, "Get all programs")
-	flags.StringVar(&cfg.lib, "bpf-lib", "bpf/objs/", "Location of Tetragon libs (btf and bpf files)")
+	flags.StringVar(&cfg.lib, "bpf-lib", "?", "Location of Tetragon libs, btf and bpf files, '?' to auto detect")
```
kkourt, line 159: Any reason of why not using the empty string?
Not a blocker, just curious.

kkourt (approved):
LGTM, thanks.

cmd/tetra/debug/progs.go
olsajiri: heh, I thought that's what you asked for :-) but it's -- ""? --  I need eye checkup, will change

PR #3243, opened 2024-12-20
[optimize] Avoid heavy  kallsyms load multiple times

arthur-zhang:
Optimize heavy kallsyms initialization using sync.Once

### Description

Improve the performance by caching the result of NewKsyms operation using
sync.Once to ensure it's only executed once.

In current implement，kallsyms will be loaded multiple times

eg:
```
ksyms.NewKsyms (/home/arthur/tetra/tetragon-1.2.1/pkg/ksyms/ksyms.go:62)
ksyms.KernelSymbols (/home/arthur/tetra/tetragon-1.2.1/pkg/ksyms/ksyms.go:22)
base.setupPrograms (/home/arthur/tetra/tetragon-1.2.1/pkg/sensors/base/base.go:104)
base.GetInitialSensor.func1 (/home/arthur/tetra/tetragon-1.2.1/pkg/sensors/base/base.go:170)
```

and  multiple times in btf.ValidateKprobeSpec in every probe calls

```
ksyms.KernelSymbols (/home/arthur/tetra/tetragon-1.2.1/pkg/ksyms/ksyms.go:32)
btf.ValidateKprobeSpec (/home/arthur/tetra/tetragon-1.2.1/pkg/btf/validation.go:51)
tracing.preValidateKprobes (/home/arthur/tetra/tetragon-1.2.1/pkg/sensors/tracing/generickprobe.go:512)
```

and tracing...

### Changelog

```release-note
optimize kallsyms initialization using sync.Once
```

### DIFF
--- a/pkg/ksyms/ksyms.go
+++ b/pkg/ksyms/ksyms.go
@@ -11,6 +11,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/option"
@@ -18,8 +19,17 @@ import (
 	lru "github.com/hashicorp/golang-lru/v2"
 )
 
+var (
+	ksymsInstance *Ksyms
+	ksymsOnce     sync.Once
+	ksymsErr      error
+)
+
 func KernelSymbols() (*Ksyms, error) {
-	return NewKsyms(option.Config.ProcFS)
+	ksymsOnce.Do(func() {
+		ksymsInstance, ksymsErr = NewKsyms(option.Config.ProcFS)
+	})
+	return ksymsInstance, ksymsErr
 }
 
 type ksym struct {


kkourt (commented):
Thanks! This is something that we had done in the past but removed due to the memory overhead: https://github.com/cilium/tetragon/commit/f6bc9f72dd9375b08396d0322e548be6c742eede.

Is there a certain workflow that you are trying to optimize?

arthur-zhang:
> Thanks! This is something that we had done in the past but removed due to the memory overhead: [f6bc9f7](https://github.com/cilium/tetragon/commit/f6bc9f72dd9375b08396d0322e548be6c742eede).
> 
> Is there a certain workflow that you are trying to optimize?

we found it cause more higher cpu usage during startup time in cheap hardware devices，when we have a lot custom yaml rules to check validation. maybe we could add clean method when we don't need it?

mtardy:
Indeed this is something we have explicitly removed because of the memory consumption (around a few dozen MB). I see two ways if we want to put it back:
- Put this behind a flag for people to choose between memory and CPU
- Cache it with a timer cleanup callback of 30s/1m.

arthur-zhang:
> Indeed this is something we have explicitly removed because of the memory consumption (around a few dozen MB). I see two ways if we want to put it back:
> 
> * Put this behind a flag for people to choose between memory and CPU
> * Cache it with a timer cleanup callback of 30s/1m.

Got it. I will send a new pr when i get a better solution.

PR #3255, opened 2024-12-24
[bug] matchPIDs is using first pid only

arthur-zhang:
Fixes bug

### Description

In the code, the function next_pid_value(index, f, ty) is called multiple times to update the index, but the updated value is not written back to sel->index. This wil cause external callers to be unable to obtain the updated state.

matchPIDs only use first pid
```yaml
selectors:
- matchPIDs:
  - operator: NotIn
    values:
    - 0
    - 1
```

```
__u32 index = sel->index;

index = next_pid_value(index, f, ty);

FUNC_INLINE int next_pid_value(__u32 off, __u32 *f, __u32 ty)
{
	return off + 4;
}
```

offset use selecttor_filter->index

```
FUNC_INLINE int
process_filter_pid(struct selector_filter *sf, __u32 *f,
		   struct execve_map_value *enter, struct msg_ns *n,
		   struct msg_capabilities *c)
{
	__u32 sel, off = sf->index;  
	__u64 flags = sf->flags;
	__u64 pid;

	if (flags & PID_SELECTOR_FLAG_NSPID) {
		pid = enter->nspid;
	} else {
		pid = enter->key.pid;
	}

	if (off > 1000)
		sel = 0;
	else {
		__u64 o = (__u64)off;
		o = o / 4;
		asm volatile("%[o] &= 0x3ff;\n"
			     : [o] "+r"(o));
		sel = f[o];
	}
}
```

### Changelog

```
Fix selector filter to properly update and assign `sel->index` in `pfilter.h`
```

### DIFF
--- a/bpf/process/pfilter.h
+++ b/bpf/process/pfilter.h
@@ -342,15 +342,19 @@ selector_match(__u32 *f, struct selector_filter *sel,
 four:
 	res4 = process_filter(sel, f, enter, &msg->ns, &msg->caps);
 	index = next_pid_value(index, f, ty);
+	sel->index = index;
 three:
 	res3 = process_filter(sel, f, enter, &msg->ns, &msg->caps);
 	index = next_pid_value(index, f, ty);
+	sel->index = index;
 two:
 	res2 = process_filter(sel, f, enter, &msg->ns, &msg->caps);
 	index = next_pid_value(index, f, ty);
+	sel->index = index;
 one:
 	res1 = process_filter(sel, f, enter, &msg->ns, &msg->caps);
 	index = next_pid_value(index, f, ty);
+	sel->index = index;
 
 	if (ty == op_filter_notin)
 		return res1 & res2 & res3 & res4;
--- a/bpf/tests/pid_match_test.c
+++ b/bpf/tests/pid_match_test.c
@@ -1,0 +1,54 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* Copyright Authors of Cilium */
+
+//go:build ignore
+
+#include "vmlinux.h"
+#include "compiler.h"
+#include "bpf_core_read.h"
+#include "bpf_helpers.h"
+#include "process/retprobe_map.h"
+#include "process/types/basic.h"
+#include "process/pfilter.h"
+
+char _license[] __attribute__((section("license"), used)) = "Dual BSD/GPL";
+
+struct filter_map_value {
+	unsigned char buf[FILTER_SIZE];
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, int);
+	__type(value, struct filter_map_value);
+} test_filter_map SEC(".maps");
+
+__attribute__((section("raw_tracepoint/test"), used)) int
+test_pid_match()
+{
+	__u32 *f;
+	int zero = 0, index = 0;
+	struct pid_filter *pid;
+	struct execve_map_value *enter;
+
+	f = map_lookup_elem(&test_filter_map, &zero);
+	if (!f)
+		return 0;
+
+	pid = (struct pid_filter *)((u64)f + index);
+	index += sizeof(struct pid_filter);
+
+	enter = map_lookup_elem(&execve_map, &zero);
+	if (!enter)
+		return 0;
+
+	struct selector_filter sel = {
+		.index = index,
+		.ty = pid->op,
+		.flags = pid->flags,
+		.len = pid->len,
+	};
+
+	return selector_match(f, &sel, enter, NULL, &process_filter_pid);
+}
--- a/bpf/tests/pid_match_test.go
+++ b/bpf/tests/pid_match_test.go
@@ -1,0 +1,150 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package bpf
+
+import (
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/tetragon/pkg/api/processapi"
+	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
+	"github.com/cilium/tetragon/pkg/selectors"
+	"github.com/cilium/tetragon/pkg/sensors/exec/execvemap"
+	"github.com/stretchr/testify/assert"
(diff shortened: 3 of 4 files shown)


bpf/process/pfilter.h
```diff
@@ -342,15 +342,19 @@ selector_match(__u32 *f, struct selector_filter *sel,
 four:
 	res4 = process_filter(sel, f, enter, &msg->ns, &msg->caps);
 	index = next_pid_value(index, f, ty);
+	sel->index = index;
 three:
```
olsajiri, line 346: could we do directly `sel->index = next_pid_value(index, f, ty);`  ?

olsajiri (commented):
nice catch, could you please add test for the 2 pids matching spec? thanks

bpf/process/pfilter.h
arthur-zhang: maybe not. If we do that, index is not changed. and sel->index will be 0 and 4(using first two pids), not more pids will be choosen.

arthur-zhang:
> nice catch, could you please add test for the 2 pids matching spec? thanks

OK，I will spend some time to figure out how to do that.

arthur-zhang:
> > nice catch, could you please add test for the 2 pids matching spec? thanks
> 
> OK，I will spend some time to figure out how to do that.
write a simple test for that， please take some time to review it

bpf/tests/pid_match_test.go
mtardy, line 1: Hey thanks for taking the time to write such tests, I just wonder if there was a more straightforward way instead of adding a tests under `bpf/tests`. It's mostly because those are used for BPF unit tests and that would be maybe better to put the actual testing Go code under `pkg/someappropriatepkg` so that it gets tested along the other tests. Would you know a good place @olsajiri maybe?

bpf/tests/pid_match_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+func (ctx *testContext) runProg(selfPid uint32) error {
+	err := ctx.execvMap.Update(uint32(0), &execvemap.ExecveValue{
+		Process: processapi.MsgExecveKey{Pid: selfPid},
+	}, 0)
+	if err != nil {
+		return fmt.Errorf("failed to update execve map: %w", err)
+	}
+	_, err = ctx.prog.Run(&ebpf.RunOptions{})
+	return err
```
olsajiri, line 70: why not have test_pid_match to return the err and take get it in here,
and you could drop test_result_map map

bpf/tests/pid_match_test.c
```diff
@@ -0,0 +1,71 @@
...
+	struct selector_filter sel = {
+		.index = index,
+		.ty = pid->op,
+		.flags = pid->flags,
+		.len = pid->len,
+	};
+	res = selector_match(f, &sel, enter, NULL, &process_filter_pid);
+
+	err = map_update_elem(&test_result_map, &zero, &res, BPF_ANY);
+	if (err) {
```
olsajiri, line 66: no need for the condition, but please the other comment for runProg

olsajiri (commented):
looks good, left some comments
please split the fix and test code in separate commits

bpf/tests/pid_match_test.go
olsajiri: hum, I dont mind it there.. it's testing the bpf selector_match function, which seems ok

bpf/tests/pid_match_test.c
```diff
@@ -0,0 +1,71 @@
...
+__attribute__((section("raw_tracepoint/test"), used)) int
+test_pid_match()
+{
+	__u32 *f;
+	int zero = 0, index = 0, res = -1, err = 0;
+	struct pid_filter *pid;
+	struct execve_map_value *enter;
+
+	f = map_lookup_elem(&test_filter_map, &zero);
+
```
olsajiri, line 45: extra empty line

bpf/tests/pid_match_test.c
```diff
@@ -0,0 +1,71 @@
...
+	if (!f) {
+		return 0;
```
olsajiri, line 47: no need for {} in here

bpf/tests/pid_match_test.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+const programNamePidMatch = "test_pid_match"
+
+type testContext struct {
+	coll     *ebpf.Collection
+	prog     *ebpf.Program
+	execvMap *ebpf.Map
+	t        *testing.T
```
olsajiri, line 26: AFAICS there's no need to store `t` in `testContext`

bpf/tests/pid_match_test.c
```diff
@@ -0,0 +1,56 @@
...
+//go:build ignore
+
+#include "vmlinux.h"
+#include "compiler.h"
+#include "bpf_core_read.h"
+#include "bpf_helpers.h"
+#include "process/retprobe_map.h"
+
+#include "process/types/basic.h"
+#include "process/generic_calls.h"
```
olsajiri, line 13: I dont think this include is needed and AFAICS will cause compilation fail now

olsajiri (changes requested):
added some more comments, but looks good overall, please add missing changelogs

bpf/tests/pid_match_test.c
arthur-zhang: this is due to pfilter.h, it rely on
-  `struct execve_map_value` declare in lib/process.h
-  `op_filter_in` process/types/operations.h
- `INDEX_MASK` in `process/types/basic.h`

to make in compile, i add this headers. 

Any ideas to fix this?

bpf/tests/pid_match_test.go
arthur-zhang: refracted.

bpf/tests/pid_match_test.c
olsajiri: this does it for me

```
diff --git a/bpf/tests/pid_match_test.c b/bpf/tests/pid_match_test.c
index 1411d4a5351f..0af78469f7d8 100644
--- a/bpf/tests/pid_match_test.c
+++ b/bpf/tests/pid_match_test.c
@@ -10,7 +10,6 @@
 #include "process/retprobe_map.h"
 
 #include "process/types/basic.h"
-#include "process/generic_calls.h"
 #include "process/pfilter.h"
 
 char _license[] __attribute__((section("license"), used)) = "Dual BSD/GPL";

```

olsajiri (approved):

PR #3257, opened 2024-12-25
Pr/olsajiri/fixes

olsajiri:
Move generic maps in one place and use them statically in generic sensor code (not passed through argument or structure). This makes a difference in more complex kprobe session  code that's coming in shortly.


kkourt (approved):
LGTM, thanks!

PR #3258, opened 2024-12-25
refactor: remove deprecated installTailCalls function and its usage

arthur-zhang:
Since we do tailcall map initialize in ebpf side, we do not need do in go files.

### Description
we refract tail call map initialize in ebpf c code like this, and remove sec name from "kprobe/#idx" to "kprobe", the go code is useless.

```c
struct {
	__uint(type, BPF_MAP_TYPE_PROG_ARRAY);
	__uint(max_entries, 13);
	__uint(key_size, sizeof(__u32));
	__array(values, int(void *));
} kprobe_calls SEC(".maps") = {
	.values = {
		[0] = (void *)&generic_kprobe_setup_event,
		[1] = (void *)&generic_kprobe_process_event,
		[2] = (void *)&generic_kprobe_process_filter,
		[3] = (void *)&generic_kprobe_filter_arg,
		[4] = (void *)&generic_kprobe_actions,
		[5] = (void *)&generic_kprobe_output,
	},
};
```

previous:

```c
__attribute__((section("kprobe/0"), used)) int
generic_kprobe_setup_event(void *ctx)
```

current: 

```c
__attribute__((section("kprobe"), used)) int
generic_kprobe_setup_event(void *ctx)
```

### DIFF
--- a/pkg/sensors/program/loader.go
+++ b/pkg/sensors/program/loader.go
@@ -692,44 +692,6 @@ func slimVerifierError(errStr string) string {
 	return errStr[:headEnd] + "\n...\n" + errStr[tailStart:]
 }
 
-func installTailCalls(bpfDir string, spec *ebpf.CollectionSpec, coll *ebpf.Collection, load *Program) error {
-	// FIXME(JM): This should be replaced by using the cilium/ebpf prog array initialization.
-
-	secToProgName := make(map[string]string)
-	for name, prog := range spec.Programs {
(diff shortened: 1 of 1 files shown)


olsajiri (commented):
sorry but I'd like to keep that code around as a possibility to setup tailcalls,
there could also be some users of that.. plus it's not that much code

arthur-zhang:
got it

PR #3279, opened 2025-01-07
tetragon: harden actions a bit

olsajiri:
Adding new filter logic that triggers when process is not found in
execve_map (likely due to exhausted capacity).

In this case we currently exit the bpf program and that might leave
some crucial actions not executed (like sigkill).

We add new event_find_curr_probe call to gather process info in filter
context and use it to execute filters. The resulting kprobe event will
have minimal process data with 'unknown' flags field.


pkg/sensors/base/base.go
```diff
@@ -92,6 +105,10 @@ func setupPrograms() {
 		}
 	}
 	logger.GetLogger().Infof("Exit probe on %s", Exit.Attach)
+
+	threads := readFileDefault("/proc/sys/kernel/threads-max", 32768)
+	ExecveMap.SetMaxEntries(int(threads))
```
mtardy, line 110: Why do we need this? It seems it will make the size of the execve_map even larger as threads-max is often well above 32K. This will take significant space while running threads-max threads is pretty rare nop?

olsajiri: I tried this as mitigation for https://github.com/isovalent/security/issues/88  .. I think we need some combination of this change (with some reasonable size for execve_map) and other ways mentioned in the issue

mtardy: ah I see, maybe this is one of the cases where NO_PREALLOC could help so that we can dynamically size to a very large map. But I could see how this can lead to memory issues in the future. That's not an easy problem :/

pkg/option/flags.go
```diff
@@ -416,4 +419,6 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.Int(KeyEventCacheRetryDelay, defaults.DefaultEventCacheRetryDelay, "Delay in seconds between event cache retries")
 
 	flags.Bool(KeyCompatibilitySyscall64SizeType, false, "syscall64 type will produce output of type size (compatibility flag, will be removed in v1.4)")
+
+	flags.String(KeyExecveMapEntries, "", "Set entries for execve_map table (default 32768)")
```
kkourt, line 423: Can we have the default from the actual default value here? If we ever change it, the change should be reflected in help.

olsajiri:
note the checkpatch fails on container_of macro which is false alarm

bpf/process/generic_calls.h
```diff
@@ -579,7 +579,7 @@ FUNC_INLINE int generic_retkprobe(void *ctx, struct bpf_map_def *calls, unsigned
 	enter = event_find_curr(&ppid, &walker);
 
 	e->common.op = MSG_OP_GENERIC_KPROBE;
-	e->common.flags |= MSG_COMMON_FLAG_RETURN;
+	e->common.flags = MSG_COMMON_FLAG_RETURN;
```
mtardy, line 582: not sure to follow why we want this change?

bpf/process/bpf_process_event.h
```diff
@@ -288,4 +288,31 @@ set_in_init_tree(struct execve_map_value *curr, struct execve_map_value *parent)
...
+	read_exe((struct task_struct *)get_current_task(), &msg->exe);
+	return curr;
+}
+#else
+FUNC_INLINE struct execve_map_value *
+event_find_curr_probe(struct msg_generic_kprobe *msg)
+{
+	return NULL;
+}
+#endif
```
mtardy, line 317: what's the benefit of doing this instead of writing the function unconditionally and making the calling conditional to LARGE PROGS? Wouldn't that be more explicit that non large prog do not support that?

bpf/process/generic_calls.h
```diff
@@ -625,8 +625,12 @@ FUNC_INLINE int generic_process_filter(void)
 		return 0;
 
 	enter = event_find_curr(&ppid, &walker);
-	if (!enter)
-		return PFILTER_CURR_NOT_FOUND;
+	if (!enter) {
+		enter = event_find_curr_probe(msg);
+		if (!enter)
+			return PFILTER_CURR_NOT_FOUND;
```
mtardy, line 631: is it okay that this function always return `PFILTER_CURR_NOT_FOUND` on !__LARGE_BPF_PROG kernels? I guess this related to my other comment, there's something I don't understand I think

pkg/sensors/tracing/kprobe_sigkill_test.go
```diff
@@ -113,6 +114,41 @@ func TestKprobeSigkill(t *testing.T) {
 	testSigkill(t, makeSpecFile, checker)
 }
 
+func TestKprobeSigkillExecveMap1(t *testing.T) {
+	if !kernels.MinKernelVersion("5.3.0") {
+		t.Skip("sigkill requires at least 5.3.0 version")
+	}
+
+	// makeSpecFile creates a new spec file bsed on the template, and the provided arguments
```
mtardy, line 123: nit: we don't really care but since I saw it, little typo here `bsed` to `based`

pkg/sensors/tracing/kprobe_sigkill_test.go
```diff
@@ -113,6 +114,41 @@ func TestKprobeSigkill(t *testing.T) {
...
+				ec.NewKprobeArgumentChecker().WithIntArg(5555),
+			)).
+		WithAction(tetragon.KprobeAction_KPROBE_ACTION_SIGKILL).
+		WithProcess(ec.NewProcessChecker().WithFlags(sm.Full("unknown")))
+
+	checker := ec.NewUnorderedEventChecker(kpChecker)
+
+	option.Config.ExecveMapEntries = 1
+	testSigkill(t, makeSpecFile, checker)
+	option.Config.ExecveMapEntries = 0
```
mtardy, line 150: why the set to 0 here?

bpf/process/generic_calls.h
olsajiri: setting flags early in the processing so in following changes the filter tail call can change it (which is executed before current flags setup code) (from commit changelog)

bpf/process/bpf_process_event.h
olsajiri: well I think it's logically isolated piece of code which should be in function and is different for `__LARGE_BPF_PROG` define, so it makes sense to me to do the above

also I think it's better/readable/more manageble than having ifdefs inside the code

pkg/sensors/tracing/kprobe_sigkill_test.go
olsajiri: IIUC option.Config is global, so we need to reset `option.Config.ExecveMapEntries` so other tests will get default value

bpf/process/generic_calls.h
olsajiri: hum for that specific line we need to initialize the flags for return kprobe, I guess it could have probably went to separate patch

PR #3285, opened 2025-01-09
renovate: disable digest update on Dockerfiles

mtardy:
This only generates noise and is not needed except for our base image.


mtardy:
Hopefully this should make:
- https://github.com/cilium/tetragon/pull/3273
- https://github.com/cilium/tetragon/pull/3266
disappear

tixxdz (approved):

PR #3289, opened 2025-01-10
fix: improve error handling for tracing policies directory access

arthur-zhang:
improve error display when policy dir do not exist when starting tetragon.

### Description
Before the change，the error message is not clear to figure out what happens when custom tracing policy dir do not exist, the error message is 
```
time="2025-01-10T10:28:37+08:00" level=fatal msg="Failed to start tetragon" error="stat .: no such file or directory"
```
Now we check all directories upfront and provide better error context.

```
time="2025-01-10T10:38:06+08:00" level=fatal msg="Failed to start tetragon" error="Tracing Policies directory does not exist: /home/arthur/dev/tetragon/my_policy_dir"
```

### Changelog

```
Optimize directory existence check in loadTpFromDir
```

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -577,17 +577,18 @@ func waitCRDs(config *rest.Config) error {
 }
 
 func loadTpFromDir(ctx context.Context, dir string) error {
-	tpMaxDepth := 1
-	tpFS := os.DirFS(dir)
-
-	if dir == defaults.DefaultTpDir {
-		// If the default directory does not exist then do not fail
-		// Probably tetragon not fully installed, developers testing, etc
-		if _, err := os.Stat(dir); os.IsNotExist(err) {
+	if _, err := os.Stat(dir); err != nil {
+		// Do not fail if the default directory doesn't exist,
+		// it might because of developer setup or incomplete installation
+		if os.IsNotExist(err) && dir == defaults.DefaultTpDir {
 			log.WithField("tracing-policy-dir", dir).Info("Loading Tracing Policies from directory ignored, directory does not exist")
 			return nil
 		}
+		return fmt.Errorf("Failed to access tracing policies dir %s: %w", dir, err)
 	}
+
+	tpMaxDepth := 1
+	tpFS := os.DirFS(dir)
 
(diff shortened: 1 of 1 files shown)


cmd/tetragon/main.go
```diff
@@ -577,18 +577,22 @@ func waitCRDs(config *rest.Config) error {
...
+		if os.IsNotExist(err) {
+			// If the default directory does not exist then do not fail
+			// Probably tetragon not fully installed, developers testing, etc
+			if dir == defaults.DefaultTpDir {
+				log.WithField("tracing-policy-dir", dir).Info("Loading Tracing Policies from directory ignored, directory does not exist")
+				return nil
+			}
+			return fmt.Errorf("Tracing Policies directory does not exist: %s", dir)
 		}
+		return fmt.Errorf("Failed to access tracing policies directory: %w", err)
```
mtardy, line 590:
```suggestion
		// Do not fail if the default directory doesn't exist,
		// it might because of developer setup or incomplete installation
		if os.IsNotExist(err) && dir == defaults.DefaultTpDir {
			log.WithField("tracing-policy-dir", dir).Info("Loading Tracing Policies from directory ignored, directory does not exist")
			return nil
		}
		return fmt.Errorf("Failed to access tracing policies dir %s: %w", dir, err)
```

mtardy (commented):
Hey thanks! Not related to your change but the existing code: I'm not a huge fan of this flow as it looks unnecessarily complicated and I don't exactly get why we don't want to fail in a situation where the directory is given but does not exists. @tixxdz do you remember why you added this "return nil" case? Should we keep this?

You could simplify this a bit as suggested, it's a less explicit error msg but the code flow is a bit more understandable.

cmd/tetragon/main.go
tixxdz: I think this relates how the code evolved to not break various tetragon versions, but also in case you also just run ./tetragon directly.

As long as we don't fail if default dir doesn't exist and we fail for all the rest is fine.

Thank you @mtardy @arthur-zhang

mtardy (approved):
Lgtm!

PR #3298, opened 2025-01-10
tetragon: Fix override program pin for fmodret and kprobe multi

olsajiri:
Currently we pin override program with -override suffix, like:

  .../generic_kprobe/security_inode_mkdir-override

It's leftover from before we had the the bpffs directory structure, now we should pin it like:

  .../generic_kprobe/security_inode_mkdir/prog_override


mtardy (approved):
Thanks! Love the 2nd commit as I noticed that as well.

kevsecurity (approved):

PR #3305, opened 2025-01-15
tetragon: add get_current_task_btf

jrfastab:
Add get_current_task_btf so we can reduce bpf_probe_read usage and start using BTF pointer walking on newer kernels.


will-isovalent (approved):

michi-covalent (approved):
😇

olsajiri (approved):

Issue #786, opened 2023-03-10
ARM-specific vmlinux.h


kkourt:
#734 used the x86 vmlinux.h as a quick way to get the arm64 port going.

Long term, we might want to use a `vmlinux.h` generated from an `arm64` kernel.
Specifically:
 * generate an arm vmlinux.h and place it in include/arm64/vmlinux.h
 * move the old (x86) version to include/x86/vmlinux.h
 * have a include/vmlinux.h that picks the proper file gbased on the ARCH

@mtardy tried this with https://gist.github.com/mtardy/ec17aa35132b4fc769519f166b808f5b, but it was not trivial to use.

mtardy:
See this discussion also https://github.com/cilium/tetragon/pull/734#discussion_r1129606385, @tixxdz mentioned that the `vmlinux.h` in its current state after the specified PR merged might be wrong for arm64. Indeed, some arch-specific values in enums might exist (and miss in arm64) which might result in very difficult-to-debug issues. For example, [this enum value](https://elixir.bootlin.com/linux/latest/source/include/linux/page-flags.h#L125) might miss on arm64 and shift all the values by 1.

lmb:
> I guess an arch specific flag enum can easily cause trouble and hard to track...

Do you have an example in mind? The intended failure mode is that the program referring an arch specific enum that doesn't exist in the current kernel will fail to load with an error telling you which enum  value is missing. If that is currently not the case I'd consider it a bug and something we should fix for sure!

tixxdz:
> Do you have an example in mind? The intended failure mode is that the program referring an arch specific enum that doesn't exist in the current kernel will fail to load with an error telling you which enum value is missing. If that is currently not the case I'd consider it a bug and something we should fix for sure!

Ok, that's great then if CO-RE can adapt or error out on it!

So for now seems we are fine, thanks @lmb ;-) ! what remains is minor: as you noted if we start using arm specific bpf programs then having arm specific values will help to compile, and always remember to look at kernel and not vmlinux.h as source of truth to do Tetragon matchArgs on raw integer values (not symbols) in tracing policies...

These are minor issues, we can fix later or maybe some one will improve it.

kkourt:
OK, so it seems to me that the conclusion is that we are not aware of any potential issues with how things are now. So I'd suggest we close this issue. An objections?

Thanks @lmb!

[...]

PR #3308, opened 2025-01-16
tetragon: update vmlinux.h

olsajiri:
update vmlinux.h now both for both archs we build for

Fixes: https://github.com/cilium/tetragon/issues/786


olsajiri:
note the syntax checked does not like vmlinux.h formatting,
but I'd like to keep it, so the future update diff is reasonable

tpapagian (approved):

kkourt (approved):

PR #3321, opened 2025-01-21
fixes in server cleanup

kkourt:
See patches.

```release-note
tetragon: avoid the agent from hanging in some corner error conditions
```


kkourt:
clang-format issues are fixed in: https://github.com/cilium/tetragon/pull/3322 so we can ignore them.

kevsecurity (approved):
LGTM

PR #3328, opened 2025-01-23
tetragon: map changes

olsajiri:
Adding execve-map-entries option to setup entries of execve_map map, plus related loader changes.

### DIFF
--- a/pkg/option/flags.go
+++ b/pkg/option/flags.go
@@ -118,6 +118,9 @@ const (
 	KeyEventCacheRetryDelay = "event-cache-retry-delay"
 
 	KeyCompatibilitySyscall64SizeType = "enable-compatibility-syscall64-size-type"
+
+	KeyExecveMapEntries = "execve-map-entries"
+	KeyExecveMapSize    = "execve-map-size"
 )
 
 type UsernameMetadaCode int
@@ -252,6 +255,8 @@ func ReadAndSetFlags() error {
 
 	Config.CompatibilitySyscall64SizeType = viper.GetBool(KeyCompatibilitySyscall64SizeType)
 
+	Config.ExecveMapEntries = viper.GetInt(KeyExecveMapEntries)
+	Config.ExecveMapSize = viper.GetString(KeyExecveMapSize)
 	return nil
 }
 
@@ -419,4 +424,7 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.Int(KeyEventCacheRetryDelay, defaults.DefaultEventCacheRetryDelay, "Delay in seconds between event cache retries")
 
 	flags.Bool(KeyCompatibilitySyscall64SizeType, false, "syscall64 type will produce output of type size (compatibility flag, will be removed in v1.4)")
+
+	flags.Int(KeyExecveMapEntries, 0, "Set entries for execve_map table (default 32768)")
+	flags.String(KeyExecveMapSize, "", "Set size for execve_map table (allows K/M/G suffix)")
 }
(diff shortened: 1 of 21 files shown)


pkg/option/flags.go
```diff
@@ -252,6 +254,7 @@ func ReadAndSetFlags() error {
 
 	Config.CompatibilitySyscall64SizeType = viper.GetBool(KeyCompatibilitySyscall64SizeType)
 
+	Config.ExecveMapEntries = viper.GetString(KeyExecveMapEntries)
```
mtardy, line 257: proposition: maybe you could call `GetExecveMapEntries` so that you do the flag parsing here and you have a number in Config.ExecveMapEntries?

mtardy (approved):
Thanks, looks good! Just some nits or questions.

pkg/sensors/program/loader.go
```diff
@@ -804,8 +805,33 @@ func doLoadProgram(
...
+	// - maps passed to loader
+	//
+	// We take user maps from loadOpts.Maps and use them to resolve
+	// program maps in following order:
+	//   1) load.PinMap[name]
+	//   2) loadOpts.Maps user maps
+
+	maps := map[string]*Map{}
+	for _, pm := range loadOpts.Maps {
+		if !pm.IsOwner() {
```
kkourt, line 830: Out of curiosity, why are we skipping the "owned" maps?

pkg/sensors/base/base.go
```diff
@@ -77,6 +84,43 @@ var (
...
+	// follow threads-max entries
+	if str == "max" {
+		if val, err := readThreadsMax("/proc/sys/kernel/threads-max"); err == nil {
+			return int(val)
+		}
+		logger.GetLogger().Warn("Failed to read /proc/sys/kernel/threads-max file, falling back to default")
+		return ExecveMapMaxEntries
+	}
+	// set entries based on size
+	size, err := strutils.ParseSize(str)
```
kkourt, line 115: Do I understand correctly that we set entries based on size if there is a suffix? For example, `1024` is interpreted as the number of entries, but `1K` would be interpreted as the total size of the map?

It seems that this could cause confusion. An alternative would be to use two options, such as `--execve-map-max-entries` and `--execve-map-max-size`, and just keep one if both are set (maybe with a warning).

pkg/sensors/program/loader.go
olsajiri: so the program loader now gets all the sensor maps to it's possible to resolve 'not owned' maps,
so we can use easily other sensor maps via constructs like:
`
maps = append(maps, program.MapUserFrom(base.ExecveMap))
`
I tried to explain that in the comment above this code, I'll try to put more details in there

pkg/option/flags.go
```diff
@@ -419,4 +422,6 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.Int(KeyEventCacheRetryDelay, defaults.DefaultEventCacheRetryDelay, "Delay in seconds between event cache retries")
 
 	flags.Bool(KeyCompatibilitySyscall64SizeType, false, "syscall64 type will produce output of type size (compatibility flag, will be removed in v1.4)")
+
+	flags.String(KeyExecveMapEntries, "32768", "Set entries for execve_map table (entries/size/max)")
```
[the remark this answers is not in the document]
olsajiri: I'll add the extra option

pkg/sensors/program/loader.go
kkourt: Thanks!

kkourt (approved):
Thanks!

PR #3332, opened 2025-01-24
fix: Escape Markdown pipes

z63d:
### Description

There is a display problem because pipe is used in the Markdown table. The string after pipe is not displayed.
https://tetragon.io/docs/reference/helm-chart/
https://github.com/cilium/tetragon/blob/main/install/kubernetes/tetragon/README.md
Fix it by escaping pipe.

Also, if you escape pipe, the second half of the string will be recognized, and asterisk will make the following part bold.
```
**". For more control, you can select which binary or binaries should have their arguments redacted with the `binary_regex` field. NOTE: This feature uses RE2 as its regular expression library. Make sure that you follow RE2 regular expression guidelines as you may observe unexpected results otherwise. More information on RE2 syntax can be found [here](https://github.com/google/re2/wiki/Syntax). NOTE: When writing regular expressions in JSON, it is important to escape backslash characters. For instance `\Wpasswd\W?` would be written as `{"redact": "\\Wpasswd\\W?"}`. As a concrete example, the following will redact all passwords passed to processes with the "--password" argument: {"redact": ["--password(?:\\s+|=)(\\S*)"]} Now, an event which contains the string "--password=foo" would have that string replaced with "--password=**
```
So escape asterisk.


mtardy (changes requested):
Oh, both those files are generated from the values, you could modify the comment there or fix the genrator.

install/kubernetes/tetragon/values.yaml
```diff
@@ -124,15 +124,15 @@ tetragon:
   # As a concrete example, the following will redact all passwords passed to
   # processes with the "--password" argument:
   #
-  #   {"redact": ["--password(?:\\s+|=)(\\S*)"]}
+  #   {"redact": ["--password(?:\\s+\|=)(\\S*)"]}
```
mtardy, line 127: well this is not ideal since it eventually modify these regex (binary_regex) so that docs looks better but the direct docs on values isn't exact anymore. I'm afraid you'll have to fix the thing we use for the generation, it seems the issue already exists :/. This is a complete different task than just fixing this docs so don't feel like you have to fix this!

See:
- https://github.com/norwoodj/helm-docs/issues/118
- https://github.com/norwoodj/helm-docs/issues/171

z63d: Thanks! You are right.
Should I add this to the Issue?

mtardy: as you wish! if you don't consider fixing this other project's issue I might take a look eventually, tell me

kkourt:
@z63d I'm moving this to because it has not seen any activity for 3 weeks. Feel free to undraft it whenever is updated. Thanks!

Issue #3304, opened 2025-01-15
Network Monitoring - Typo in kubesystem namespace example


shmcfarl:
The instructions at https://tetragon.io/docs/getting-started/network/ refer to "kubesystem" as the namespace in the commands. It should be "kube-system".

Example: 

POD=$(kubectl -n **kubesystem** get pods -l 'app.kubernetes.io/name=tetragon' -o name --field-selector spec.nodeName=$(kubectl get pod xwing -o jsonpath='{.spec.nodeName}'))

kkourt:
Good catch, thanks!
Feel free to submit a PR that fixes it.

See https://tetragon.io/docs/contribution-guide/ for more details

PR #2998, opened 2024-10-14
Fix typo in execution.md

Golim:
Replaced kubesystem with kube-system in one command

Fixes 

### Changelog

```release-note
```

PR #3334, opened 2025-01-24
docs: fix typo referencing kube-system as kubesystem

z63d:
Fixes https://github.com/cilium/tetragon/issues/3304

### Description

Fix typo "kubesystem".


mtardy (approved):
Oh it looks like someone already did something like that recently but it was a bit incomplete https://github.com/cilium/tetragon/pull/2998.

It doesn't matter for this PR but for the next one if you can write a bit more explicit commit message like 'docs: fix typo referencing kube-system as kubesystem' or whatever it would be better :)

PR #3335, opened 2025-01-24
tetra: Set default for retries to 1

olsajiri:
The grpc update to v1.70.0 and especially [1] requires retries to be > 1 in order to have a valid retry policy.

As we already do +1 later, let's set the default to 1.

[1] https://github.com/grpc/grpc-go/pull/7905
Suggested-by: Anastasios Papagiannis <anastasios.papagiannis@isovalent.com>


lambdanis (approved):
With this change we could also revert 455458f4579cca73697edc67d7f41ebbc000ac88? Not needed though.

mtardy (approved):
thanks

tpapagian:
> With this change we could also revert 455458f4579cca73697edc67d7f41ebbc000ac88? Not needed though.

I believe that we still need that as in e2e tests we just call functions from tetra and not the binary itself. I will check that to be sure and revert that into f needed in a separate PR.

tpapagian (approved):

PR #3338, opened 2025-01-24
proc_reader: handle in_init_tree

will-isovalent:
Recently we introduced the in_init_tree flag into execve map values to indicate whether
a process is a member of the initial process tree for a container. This worked well for
containers started after Tetragon, but broke for cases where the container was started
before Tetragon, since our procfs walk did not account for the in_init_tree flag. Fix this
behaviour by introducing logic in the procfs walk to account for this.

```release-note
Fix in_init_tree flag for processes started before Tetragon.
```

### DIFF
--- a/pkg/observer/observertesthelper/observer_test_helper.go
+++ b/pkg/observer/observertesthelper/observer_test_helper.go
@@ -521,69 +521,6 @@ func ExecWGCurl(readyWG *sync.WaitGroup, retries uint, args ...string) error {
 	return err
 }
 
-// DockerCreate creates a new docker container in the background. The container will
-// be killed and removed on test cleanup.
-// It returns the containerId on success, or an error if spawning the container failed.
-func DockerCreate(tb testing.TB, args ...string) (containerId string) {
-	// note: we are not using `--rm` so we can choose to wait on the container
-	// with `docker wait`. We remove it manually below in t.Cleanup instead
-	args = append([]string{"create"}, args...)
-	id, err := exec.Command("docker", args...).Output()
-	if err != nil {
-		tb.Fatalf("failed to spawn docker container %v: %s", args, err)
-	}
-
-	containerId = strings.TrimSpace(string(id))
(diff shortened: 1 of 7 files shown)


will-isovalent:
@tixxdz pulling you in for review for the changes in the third patch.

michi-covalent (approved):

pkg/observer/observertesthelper/observer_test_helper.go
```diff
@@ -524,64 +525,29 @@ func ExecWGCurl(readyWG *sync.WaitGroup, retries uint, args ...string) error {
...
-		}
-	})
-
-	return containerId
+	return docker.Create(tb, args...)
 }
 
 // DockerStart starts a new docker container with a given ID.
+// DEPRECATED: use docker.DockerStart instead.
 func DockerStart(tb testing.TB, id string) {
```
olsajiri, line 535: is it necessary to keep old version? why not use docker package directly from current callers, seems like it's just one place

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -647,6 +661,10 @@ func listRunningProcs(procPath string) ([]procs, error) {
 
 	logger.GetLogger().Infof("Read ProcFS %s appended %d/%d entries", option.Config.ProcFS, len(processes), len(procFS))
 
+	slices.SortFunc(processes, func(a, b procs) int {
+		return cmp.Compare(a.pid, b.pid)
+	})
+
```
olsajiri, line 667: curious why is this needed

tixxdz (approved):
LGTM, see also Jiri comments! much appreciated @will-isovalent

pkg/observer/observertesthelper/observer_test_helper.go
will-isovalent: Just me being lazy, no other reason.

PR #3347, opened 2025-01-28
pkg/sensors: RawDetachUnloader do not detach on unpin false

mtardy:
I'm really unfamiliar with this part of the code, do we need something similar for tc? Does it make sense to not detach on `unpin=false`?

```release-note
Fix a bug where unloading programs where detaching them even in the case of unpin false (i.e.) --keep-sensors-on-exit
```

### DIFF
--- a/pkg/sensors/unloader/unloader.go
+++ b/pkg/sensors/unloader/unloader.go
@@ -82,15 +82,20 @@ type RawDetachUnloader struct {
 	AttachType ebpf.AttachType
 }
 
-func (rdu *RawDetachUnloader) Unload(_ bool) error {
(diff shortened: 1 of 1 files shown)


olsajiri (commented):
looks good, PROG_ATTACH does not return any link, so there's nothing to unpin, but we must skip the detach operation for 'unpin == false' otherwise the pinned program will be un-attached

mtardy:
Should I do the same patch for tc @olsajiri ?

olsajiri (approved):
could you please add similar comment to TcUnloader ? otherwise looks good, thanks

mtardy:
CI went green and then I added a comment, so merging this.

PR #3352, opened 2025-01-29
pkg/sensors: pin link in TracingAttach

mtardy:
Links were not pinned to the fs and thus --keep-sensors-on-exit was not working with program using such attach mechanism since they were garbage collected by the kernel on closing the file descriptor.


olsajiri (approved):
thanks

PR #3355, opened 2025-01-29
Pr/jrfastab/config map testing

jrfastab:
I got a bug report that a config file option did not work when pushed through the config map as a file in systemd. It turned out to be a typo but this adds infrastructure and a test to add go tests to check this works.

For now I just add one test but we can add more easily now.


michi-covalent (approved):
`:rubber-stamp:`

jrfastab:
close this for now.

PR #3358, opened 2025-01-30
Add socket and sockaddr types

kevsecurity:
### Description
The security_socket_* functions take pointers to struct socket and struct sockaddr, but we don't currently support these types. This series adds support for them. The socket support resolves the socket's struct sock pointer and uses that as per struct sock support. The sockaddr support uses the SAddr and SPort matchArgs operators to represent SinAddr and SinPort.

### Changelog
Add support for struct socket and struct sockaddr.

```release-note
Add support for struct socket and struct sockaddr.
```


will-isovalent (approved):
LGTM

PR #3367, opened 2025-02-05
docs: fix the Example jq filter in Observability Policies

z63d:
### Description

Fixed the Example jq filter command in Observability Policies "Privileges Escalation via File Capabilities Execution".


mtardy (approved):
Thanks!

PR #3368, opened 2025-02-05
api: build using buf

will-isovalent:
Refactor our build pipeline for the protobuf API to use buf[1] to do our code generation.

This approach has a few key advantages:

1. Simpler and more maintainable. We no longer need to maintain a complex makefile that
   gets more complex with every new .proto file.

2. Automatic checking for API breakages including JSON and wire compatibility.

3. Linting and automatic code formatting to improve code consistency and enforce best
   practices. Note: linting is currently disabled pending a larger incoming PR that will
   address current issues.

[1]: https://buf.build/docs/cli/


michi-covalent (approved):
✨

PR #3382, opened 2025-02-10
helm: add support  for cgidmap/cri

kkourt:
```release-note
helm: add cri.enabled, cri.socketHostPath, and cgidmap.enables variables
metrics: add metrics for cgidmap CRI resolution
```

### DIFF
--- a/install/kubernetes/tetragon/templates/_container_tetragon.tpl
+++ b/install/kubernetes/tetragon/templates/_container_tetragon.tpl
@@ -38,6 +38,10 @@
       name: export-logs
     - mountPath: "/procRoot"
       name: host-proc
+{{- if and (.Values.tetragon.cri.enabled) (.Values.tetragon.cri.socketHostPath) }}
+    - mountPath: {{ quote .Values.tetragon.cri.socketHostPath }}
+      name: cri-socket
+{{- end }}
 {{- range .Values.extraHostPathMounts }}
     - name: {{ .name }}
       mountPath: {{ .mountPath }}
--- a/install/kubernetes/tetragon/values.yaml
+++ b/install/kubernetes/tetragon/values.yaml
@@ -237,6 +237,16 @@ tetragon:
   enableKeepSensorsOnExit: false
   # -- Configure the interval (suffixed with s for seconds, m for minutes, etc) for the process cache garbage collector.
   processCacheGCInterval: 30s
+  # -- Configure tetragon pod so that it can contact the CRI running on the host
+  cri:
+    enabled: false
+    # -- path of the CRI socket on the host. This will typically be
+    # "/run/containerd/containerd.sock" for containerd or "/var/run/crio/crio.sock"  for crio.
+    socketHostPath: ""
+  # -- Enabling cgidmap instructs the Tetragon agent to use cgroup ids (instead of cgroup names) for
+  # pod association. This feature depends on cri being enabled.
+  cgidmap:
+    enabled: false
 # Tetragon Operator settings
 tetragonOperator:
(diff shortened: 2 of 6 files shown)


pkg/metrics/crimetrics/crimetrics.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/cilium/tetragon/pkg/metrics"
+	"github.com/cilium/tetragon/pkg/metrics/consts"
+)
+
+var (
+	CriResolutionsTotal = metrics.MustNewCounter(
+		metrics.NewOpts(
+			consts.MetricsNamespace,
+			"cri_cgidmap", "resolutions_total",
+			"number of total cgidmap CRI resolutions", nil, nil, nil),
```
mtardy, line 16: maybe you could explicit what `cgidmap` stands for? For someone reading these metrics out of the blue

install/kubernetes/tetragon/templates/_container_tetragon.tpl
```diff
@@ -38,6 +38,10 @@
       name: export-logs
     - mountPath: "/procRoot"
       name: host-proc
+{{- if ne .Values.tetragon.cri.socketHostPath "" }}
```
mtardy, line 41: if you want you can simplify with, or can you use `enabled` that you define for this field?
```suggestion
{{- if .Values.tetragon.cri.socketHostPath }}
```

install/kubernetes/tetragon/values.yaml
```diff
@@ -237,6 +237,16 @@ tetragon:
...
+  # -- Configure tetragon pod so that it can contact the CRI running on the host
+  cri:
+    enable: false
+    # -- path of the CRI socket on the host. This will typically be
+    # "/run/containerd/containerd.sock" for containerd or "/var/run/crio/crio.sock"  for crio.
+    socketHostPath: ""
+  # -- Enabling cgidmap instructs the Tetragon agent to use cgroup ids (instead of cgroup names) for
+  # pod association. This feature depends on cri being enabled.
+  cgidmap:
+    enable: false
```
mtardy, line 249: I think the convention is to use "enabled" and not "enable"
```suggestion
  cri:
    enabled: false
    # -- path of the CRI socket on the host. This will typically be
    # "/run/containerd/containerd.sock" for containerd or "/var/run/crio/crio.sock"  for crio.
    socketHostPath: ""
  # -- Enabling cgidmap instructs the Tetragon agent to use cgroup ids (instead of cgroup names) for
  # pod association. This feature depends on cri being enabled.
  cgidmap:
    enabled: false
```

pkg/metrics/crimetrics/crimetrics.go
kkourt: I've explained it better, but I'm still unsure if it helps. We need docs for this, but that's another PR.

install/kubernetes/tetragon/templates/_container_tetragon.tpl
kkourt: Makes sense. Used: `{{- if and (.Values.tetragon.cri.enabled) (.Values.tetragon.cri.socketHostPath) }}`

install/kubernetes/tetragon/values.yaml
kkourt: yap, fixed. Thanks!

kkourt:
@mtardy, thanks for the nice review! I've updated the PR, PTAL. I've also included a script I used to test this in the commit message. Maybe one day we 'll write some simple tests for the helm logic.

mtardy (approved):
looks fantastic

PR #3401, opened 2025-02-14
renovate: more robust parser for Go version in go.mod

mtardy:
Go automatically removes the toolchain when the go version is equal to the toolchain version so let's do best effort: try to parse the toolchain directive, if it fails, parse the golang version directive directly.


mtardy:
Needed for https://github.com/cilium/tetragon/pull/3361.

tpapagian (approved):

PR #3403, opened 2025-02-14
Refactor CRD defaulting and validation as generic

lambdanis:
Refactor the code for working with TracingPolicy CRD outside of Kubernetes
context to generically support any CRD. This includes loading custom resources
from YAML, applying defaults and validating them.

This design replaces the global state (tracingpolicy.defaultState), previously
initialized using sync.Once and used to cache validators and structural schemas,
with a generic struct (CRDContext[P]), initialized per CRD. CRDContext holds
structs necessary to process a particular CRD. CRDContexts for
TracingPolicy(Namespaced) are initialized in init().

The defaulting and validating logic is mostly unchanged and mimics the behavior
of Kubernetes apiserver. Two differences from the previous implementation maybe
worth noting:

- We now validate uniqueness invariants of arrays that are sets or maps
- To make FromYAML generic, but also avoid unnecessary copying of structs, we
  pass a pointer as a type parameter, but then use reflect to allocate the
  struct, to unmarshal into it. It's a bit ugly, but works.

To ensure that the generic functionality stays decoupled from TracingPolicy,
move it from pkg/tracingpolicy to a new pkg/crdutils package.

There is already another crdutils package under pkg/k8s module. However, it's
specific to Tetragon CRDs, while the defaulting and validating logic is
completely generic. Therefore, it seems better to keep it separate.


mtardy (approved):
Thanks, it looks cleaner like that! lgtm

PR #3404, opened 2025-02-15
helm: mount /sys/kernel/security/lsm for generic_lsm sensor

### LINKED ISSUE #3392: Problem with applying LSM policies in k8s cluster
### What happened?

To verify if LSM BPF is enabled we need to check  the contents of  `/sys/kernel/security/lsm` 

https://github.com/cilium/tetragon/blob/c713bf16feaa871dec7198a1c7ec228644a86e8a/pkg/bpf/detect.go#L234

This file is not mounted to the container so policy load is failing. Slack [thread](https://cilium.slack.com/archives/C03EV7KJPJ9/p1739004250156899).

### Tetragon Version

1.3.0

### Kernel Version

5.15.0-130-generic

### Kubernetes Version

v1.30.9

### Bugtool

_No response_

### Relevant log output

```shell

```

### Anything else?

Hot Fix that worked:

```
```

I see two options:

1. We can add this mount to helm chart. (I prefer this) 
2. We can add the fix above to documentation. If use want to use LSM BPF in k8s just add this directory to mountPath

anfedotoff:
We need to mount `/sys/kernel/security/lsm` from host  to check if lsm bpf is enabled.

Fixes  #3392

### DIFF
--- a/install/kubernetes/tetragon/templates/daemonset.yaml
+++ b/install/kubernetes/tetragon/templates/daemonset.yaml
@@ -114,6 +114,9 @@ spec:
           path: {{ quote .Values.tetragon.cri.socketHostPath }}
           type: Socket
 {{- end }}
+      - name: security-lsm
+        hostPath:
+          path: /sys/kernel/security/lsm
 {{- end }}
       {{- with .Values.extraVolumes }}
         {{- toYaml . | nindent 6 }}


install/kubernetes/tetragon/templates/daemonset.yaml
```diff
@@ -114,6 +114,9 @@ spec:
           path: {{ quote .Values.tetragon.cri.socketHostPath }}
           type: Socket
 {{- end }}
+      - name: security-lsm
+        hostPath:
+          path: /sys/kernel/security/lsm
```
mtardy, line 119: Considering that using `hostPath` is [a sensitive feature in k8s](https://kubernetes.io/docs/concepts/storage/volumes/#hostpath) and that tetragon is a runtime security product we should maybe do a few things here:
- not use the `type: ""` here and use something more specific if possible, read-only, see https://kubernetes.io/docs/concepts/storage/volumes/#hostpath-volume-types.
- make this optional and not mounted by default with a config "enabled" and such in the values.yaml.

mtardy (commented):
Hey, thanks for the PR.

install/kubernetes/tetragon/templates/daemonset.yaml
anfedotoff: Can I ask some questions, just to make it clear, please?

1.  I think, `type: "File` might be OK, but if `/sys/kernel/security/lsm` is not exists (in other words LSM is not enabled), than  Pod will not be loaded, right? I think it's not good.
2.  If we will use an option, it might fix the problem above, but what will be if we have nodes with different configuration? Some of the nodes have LSM enabled, but some don't?
3.  Am I right that this [config](https://github.com/cilium/tetragon/blob/main/install/kubernetes/tetragon/values.yaml) is the superset of this [config](https://github.com/cilium/tetragon/blob/main/examples/configuration/tetragon.yaml)? In other words, I can add some field to helm, but it is not necessary to add this field to tetragon.yaml?

mtardy: For point 3, just worry about https://github.com/cilium/tetragon/blob/main/install/kubernetes/tetragon/values.yaml, the other one is just an example helm config, I don't exactly know what is its use.

For your point 1 and 2, I did some research, and in any case, if the file the volume points to do not exists, the pod creation will fail, even without a type.

### Failed attempt 1

So ideally we would write something like this:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: lsm-pod
spec:
  containers:
    - name: nginx
      image: nginx
      volumeMounts:
        - name: lsm-volume
          mountPath: /sys/kernel/security/lsm
          readOnly: true
  volumes:
    - name: lsm-volume
      hostPath:
        path: /sys/kernel/security/lsm
        type: File
```
But for some reason I don't understand, containerd OCI runtime creation failed to create antyhing under `/sys/kernel/security/lsm` with
```
Error: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/sys/kernel/security/lsm" to rootfs at "/sys/kernel/security/lsm": open /run/containerd/io.containerd.runtime.v2.task/k8s.io/nginx/rootfs/sys/kernel/security/lsm: no such file or directory: unknown
```
My hypothesis is that the fs under `/sys` is readonly in the container fs, and only `/sys/kernel/security` exists.

### Failed attempt 2

So I tried something like
```yaml
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: lsm-pod
spec:
  containers:
    - name: nginx
      image: nginx
      volumeMounts:
        - name: lsm-volume
          mountPath: /sys/kernel/security/lsm
          subPath: lsm
          readOnly: true
  volumes:
    - name: lsm-volume
      hostPath:
        path: /sys/kernel/security
        type: Directory
```

But it's the same issue. Note that you can basically used the attempt 1 and attempt 2 if you want to mount the file anywhere else. It will work, the issue is with mounting under `/sys/kernel/security/lsm`.

### Attempt 3

It seems the only way is to mount under this exact path is to mount the entire security directory.
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: lsm-pod
spec:
  containers:
    - name: nginx
      image: nginx
      volumeMounts:
        - name: lsm-volume
          mountPath: /sys/kernel/security
          readOnly: true
  volumes:
    - name: lsm-volume
      hostPath:
        path: /sys/kernel/security
        type: Directory
```

#### Conclusions

So, would the `/sys/kernel/security` folder at least exist (if empty) on most kernel, even those without most security modules enabled? That could be a first solution if exposing the entire directory is not a security issue.

If not, it's possible to only mount `/sys/kernel/security/lsm` but you will bump into situation where it doesn't exist (and thus the deployment will fail) and it seems you cannot mount it at the exact same path in the container fs, so you'll need to mount it elsewhere and teach tetragon where to look (flag/hardcoding/etc).

#### Kind config

For you to retry this on a linux machine, you'll need to mount an extra volume on the kind cluster, you can use this config:
```yaml
apiVersion: kind.x-k8s.io/v1alpha4
kind: Cluster
nodes:
  - role: control-plane
    extraMounts:
      - hostPath: /sys/kernel/security
        containerPath: /sys/kernel/security
```
And then `kind create cluster --config kind.yaml`

anfedotoff: Thanks for such detailed explanation! I like you idea to mount the entire directory. Mounting file, we can come up to situation when pod will not start, because file doesn't exist. It's not good. The user might be confused.

And you are right, when we mount the entire directory we need to keep in mind that it is used to manage some LSM stuff (IMA, Apparmor, etc). From other side we will mount it in read-only mode and by default this is would be disabled. I need time to think on it a bit more).

Just a thing suddenly came up to my mind. Do we really need file check???
Please, have a look at the code below:

https://github.com/cilium/tetragon/blob/bd433e65bfdb1f25588356ca252b41f9906573e5/pkg/bpf/detect.go#L230-L254

We have 3 type of checks there. The last one is the most important: we try to load a simple LSM program.  Maybe it's an overkill to check the file?

Here we recommend to check if the LSM BPF is enabled in any case:

https://github.com/cilium/tetragon/blob/bd433e65bfdb1f25588356ca252b41f9906573e5/pkg/sensors/tracing/genericlsm.go#L356-L359

mtardy: I thought you added the file check ahah, what are we checking for? Can we just try to load the eBPF prog and give a helpful error message if it doesn't work? Maybe we can just use BPF progs to detect feature as we do in many places in cilium/ebpf

anfedotoff: Ahah, yeah  you are right that was me)). This file check looks if  LSM BPF is enabled on system or not. I though that it's a very cheep check. (At that time I can't imagine k8s problems)). Next check then tries to load LSM program.  This is the most reliable test. It covers the check above. And I think we can remove the file check at all. The last check would be enough.
I think we can add at the beginning something like "Unable to load simple LSM BPF program." + current message. Because the most common case is that LSM BPF is not enabled. 

This way can help us to avoid any security issues tight with mounting `/sys/kernel/security`.
I'll close this PR and open another one with appropriate fix for underlying problem. Thanks for helping).

ScriptSathi:
I guess this PR should be revived following the discussion https://github.com/cilium/tetragon/pull/3456

PR #3409, opened 2025-02-17
watcher: Fix K8sWatcher.FindPod

lambdanis:
The K8sWatcher.FindPod method first tries to find a pod using index, then, if
unsuccessful, falls back to walking the entire pods list. This fallback was
incorrect - fix it.

I'm not sure how often or if ever this code path was being hit in the wild - it
would be hit only if the index got out-of-sync. It would be good to understand
it, but for now let's just fix the bug.

### DIFF
--- a/pkg/watcher/fake.go
+++ b/pkg/watcher/fake.go
@@ -40,7 +40,7 @@ func (watcher *FakeK8sWatcher) FindMirrorPod(_ string) (*corev1.Pod, error) {
 }
 
 func (watcher *FakeK8sWatcher) FindPod(podID string) (*corev1.Pod, error) {
-	ret, ok := findPod(watcher.pods)
+	ret, ok := findPod(podID, watcher.pods)
 	if ok {
 		return ret, nil
 	}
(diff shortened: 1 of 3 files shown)


lambdanis:
I suppose walking the pod list might be applicable only if using `FakeK8sWatcher`? In such case we could remove the fallback from `K8sWatcher.FindPod`...

In any case, the code looks clearly incorrect and no tests have caught it.

kkourt (approved):
Nice catch, thanks!

I introduced the bug in: 175c5910e63c779068d3395306d2f4aced5075ce.

AFAICT, it was a typo and I ended using up the global variable (podIdx) instead of the argument (podID).

 https://github.com/cilium/tetragon/commit/175c5910e63c779068d3395306d2f4aced5075ce#diff-29ab67ad2861533ba7cfc10ed9d527f12a57b27cae23c566d373d82027c8f4c6R167-R177
 
 I guess at some point the argument was removed.
 
 I'll mark this PR for backports.

PR #3413, opened 2025-02-17
watcher: Refactor K8sWatcher to reuse config and factories

lambdanis:
The previous K8sWatcher implementation allowed extending it with extra
informers via AddInformers method, but didn't provide a way to reuse shared
informer factories between AddInformers calls. That's wasteful, as the whole
point of a factory is to be reused.

This PR adds shared informer factories to the K8sWatcher struct: one for
global k8s built-in resources, one for node-local k8s built-in resources, and
one for CRDs. They are initialized in NewK8sWatcher function. The AddInformers
method is replaced by AddInformer, which simply adds a single informer with
indexers to the list. The Start method starts all informer factories - thanks
to storing them, there is no more need to dynamically extend a start function.

Initialization of the pod informer is moved out of NewK8sWatcher to a dedicated
function called from main, AddPodInformer.

Watching TracingPolicy(Namespaced) now reuses the K8sWatcher. Before, we were
reading Kubernetes config, creating an informer factory and starting it, all in
the WatchTracePolicy function. Now it's replaced by AddTracingPolicyInformer,
which only adds informers to the K8sWatcher.


kkourt (approved):
LGTM, thanks.

Issue #3411, opened 2025-02-17
Remove LoadBTF and allow multiple btf files for testing


ScriptSathi:
This issue is the follow-up to this conversation https://github.com/cilium/tetragon/pull/3143#discussion_r1948539484

The idea is to make a more simple use of the `btfFile` global variable in case we want to test multiple BTF files in a single test.

I think what we can do here is using a callback to set the `btfFile`, then perform the test, and global variable like this.

```go // pkg/btf/btf_test.go
func RunTestOnSingleBtf(t *testing.T, btfFname string, test func(t *testing.T)) {
	if err := btfFileExists(btfFname); err != nil {
		t.Fatal(err)
	}
	btfFile = btfFname
        t.Run(btfFname, test)
	btfFile = ""
}

func TestFindBtfFuncParamFromHook(t *testing.T) {
	RunTestOnSingleBtf(t, defaults.DefaultBTFFile, testFindBtfFuncParamFromHook)
        RunTestOnSingleBtf(t, "/path/to/my/btf/2", testFindBtfFuncParamFromHook)
        RunTestOnSingleBtf(t, "/path/to/my/btf/3", testFindBtfFuncParamFromHook)
}
```

In `FindBtfFuncParamFromHook` we could also remove the use LoadBtf() by simply using NewBtf()
https://github.com/cilium/tetragon/blob/fc759c416af0b689c2d65b0c7f6b12f595ff956d/pkg/btf/btf.go#L118-L120

Similarly we could refactor this below code by using such a function
https://github.com/cilium/tetragon/blob/fc759c416af0b689c2d65b0c7f6b12f595ff956d/pkg/btf/validation_test.go#L56-L63

@kkourt, do you think it's close enough to what you were thinking of ?

kkourt:
> @kkourt, do you think it's close enough to what you were thinking of?
Yes, thanks for writing up!

Issue #1954, opened 2024-01-10
TestLabelsDemoApp failures (flake?)


kkourt:
Hit a `TestLabelsDemoApp` failure (https://github.com/cilium/tetragon/actions/runs/7472506683/job/20334842561?pr=1948) in https://github.com/cilium/tetragon/pull/1948. Seems like a flake.

Details:
```
 I0110 09:10:55.961095   14366 dumpinfo.go:240] contacting metrics serveraddrhttp://localhost:2112/metrics
--- FAIL: TestLabelsDemoApp (30.00s)
    --- FAIL: TestLabelsDemoApp/Run_Event_Checks (10.01s)
        --- FAIL: TestLabelsDemoApp/Run_Event_Checks/Run_Event_Checks (10.01s)
            rpcchecker.go:171: 
                	Error Trace:	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/tests/e2e/checker/rpcchecker.go:171
                	            				/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/vendor/sigs.k8s.io/e2e-framework/pkg/env/env.go:422
                	            				/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/vendor/sigs.k8s.io/e2e-framework/pkg/env/env.go:453
                	Error:      	Received unexpected error:
                	            	failed to get events after 10 tries
                	Test:       	TestLabelsDemoApp/Run_Event_Checks/Run_Event_Checks
                	Messages:   	checks should pass
    --- FAIL: TestLabelsDemoApp/Run_Workload (30.00s)
        --- FAIL: TestLabelsDemoApp/Run_Workload/Wait_for_Checker (30.00s)
            rpcchecker.go:107: 
                	Error Trace:	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/tests/e2e/checker/rpcchecker.go:107
                	            				/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/vendor/sigs.k8s.io/e2e-framework/pkg/env/env.go:422
                	            				/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/vendor/sigs.k8s.io/e2e-framework/pkg/env/env.go:453
                	Error:      	failed to wait for checker labelsEventChecker to start after 30s
                	Test:       	TestLabelsDemoApp/Run_Workload/Wait_for_Checker
[...]

PR #3414, opened 2025-02-18
Remove `LoadBtf()` and add test wrapper for single btf use

ScriptSathi:
Fix https://github.com/cilium/tetragon/issues/3411

### DIFF
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -34,12 +34,12 @@ func observerFindBTF(lib, btf string) (string, error) {
 	if btf == "" {
 		// Alternative to auto-discovery and/or command line argument we
 		// can also set via environment variable.
-		tetragonBtfEnv := os.Getenv("TETRAGON_BTF")
-		if tetragonBtfEnv != "" {
-			if _, err := os.Stat(tetragonBtfEnv); err != nil {
+		tetragonBTFEnv := os.Getenv("TETRAGON_BTF")
+		if tetragonBTFEnv != "" {
+			if _, err := os.Stat(tetragonBTFEnv); err != nil {
 				return btf, err
 			}
-			return tetragonBtfEnv, nil
+			return tetragonBTFEnv, nil
 		}
 
 		_, kernelVersion, err := kernels.GetKernelVersion(option.Config.KernelVersion, option.Config.ProcFS)
@@ -98,15 +98,15 @@ func GetCachedBTFFile() string {
 	return btfFile
 }
 
-func FindBtfFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
+func FindBTFFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
 	spec, err := NewBTF()
 	if err != nil {
 		return nil, err
 	}
-	return findBtfFuncParamFromHookWithSpec(spec, hook, argIndex)
+	return findBTFFuncParamFromHookWithSpec(spec, hook, argIndex)
 }
 
-func findBtfFuncParamFromHookWithSpec(spec *btf.Spec, hook string, argIndex int) (*btf.FuncParam, error) {
+func findBTFFuncParamFromHookWithSpec(spec *btf.Spec, hook string, argIndex int) (*btf.FuncParam, error) {
 	var hookFn *btf.Func
 
 	if err := spec.TypeByName(hook, &hookFn); err != nil {
@@ -118,8 +118,8 @@ func findBtfFuncParamFromHookWithSpec(spec *btf.Spec, hook string, argIndex int)
 		return nil, fmt.Errorf("failed to find BTF type for hook %q: %w", hook, err)
 	}
 
-	btfHookProto, isBtfFuncProto := hookFn.Type.(*btf.FuncProto)
-	if !isBtfFuncProto {
+	btfHookProto, isBTFFuncProto := hookFn.Type.(*btf.FuncProto)
+	if !isBTFFuncProto {
 		return nil, fmt.Errorf("hook %q has no BTF type FuncProto", hook)
 	}
 	paramLen := len(btfHookProto.Params)
@@ -147,7 +147,7 @@ func ResolveNestedTypes(ty btf.Type) btf.Type {
 	return ty
 }
 
-// ResolveBtfPath function recursively search in a btf structure in order to
+// ResolveBTFPath function recursively search in a btf structure in order to
 // found a specific path until it reach the target or fail.
 
 // The function also search in embedded anonymous structures or unions to cover as
@@ -162,8 +162,8 @@ func ResolveNestedTypes(ty btf.Type) btf.Type {
 // @i: The current depth, until last element of pathToFound.
 
 // Return: The last type found matching the path, or error.
-func ResolveBtfPath(
-	btfArgs *[api.MaxBtfArgDepth]api.ConfigBtfArg,
+func ResolveBTFPath(
+	btfArgs *[api.MaxBTFArgDepth]api.ConfigBTFArg,
 	currentType btf.Type,
 	pathToFound []string,
 	i int,
@@ -177,10 +177,10 @@ func ResolveBtfPath(
 		if len(pathToFound[i]) == 0 {
 			(*btfArgs)[i].IsPointer = uint16(1)
 			(*btfArgs)[i].IsInitialized = uint16(1)
-			return ResolveBtfPath(btfArgs, ResolveNestedTypes(t.Target), pathToFound, i+1)
+			return ResolveBTFPath(btfArgs, ResolveNestedTypes(t.Target), pathToFound, i+1)
 		}
 		(*btfArgs)[i-1].IsPointer = uint16(1)
-		return ResolveBtfPath(btfArgs, ResolveNestedTypes(t.Target), pathToFound, i)
+		return ResolveBTFPath(btfArgs, ResolveNestedTypes(t.Target), pathToFound, i)
 	default:
 		ty := currentType.TypeName()
 		if len(ty) == 0 {
@@ -195,7 +195,7 @@ func ResolveBtfPath(
 }
 
 func processMembers(
-	btfArgs *[api.MaxBtfArgDepth]api.ConfigBtfArg,
+	btfArgs *[api.MaxBTFArgDepth]api.ConfigBTFArg,
 	currentType btf.Type,
 	members []btf.Member,
 	pathToFound []string,
@@ -207,7 +207,7 @@ func processMembers(
 		if len(member.Name) == 0 { // If anonymous struct, fallthrough
 			(*btfArgs)[i].Offset = member.Offset.Bytes()
 			(*btfArgs)[i].IsInitialized = uint16(1)
-			lastTy, err := ResolveBtfPath(btfArgs, ResolveNestedTypes(member.Type), pathToFound, i)
+			lastTy, err := ResolveBTFPath(btfArgs, ResolveNestedTypes(member.Type), pathToFound, i)
 			if err != nil {
 				if lastError != nil {
 					idx := i + 1
@@ -225,9 +225,9 @@ func processMembers(
 			memberWasFound = true
 			(*btfArgs)[i].Offset = member.Offset.Bytes()
 			(*btfArgs)[i].IsInitialized = uint16(1)
-			isNotLastChild := i < len(pathToFound)-1 && i < api.MaxBtfArgDepth
+			isNotLastChild := i < len(pathToFound)-1 && i < api.MaxBTFArgDepth
 			if isNotLastChild {
-				return ResolveBtfPath(btfArgs, ResolveNestedTypes(member.Type), pathToFound, i+1)
+				return ResolveBTFPath(btfArgs, ResolveNestedTypes(member.Type), pathToFound, i+1)
 			}
 			currentType = ResolveNestedTypes(member.Type)
 			break
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -98,24 +98,18 @@ func GetCachedBTFFile() string {
 	return btfFile
 }
 
-func LoadBTF() (*btf.Spec, error) {
-	if btfFile != "" {
-		return NewBTF()
-	}
-	if err := InitCachedBTF(defaults.DefaultTetragonLib, ""); err != nil {
-		return nil, err
-	}
-	return NewBTF()
-}
-
 func FindBtfFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
-	var hookFn *btf.Func
-	spec, err := LoadBTF()
+	spec, err := NewBTF()
 	if err != nil {
 		return nil, err
 	}
+	return findBtfFuncParamFromHookWithSpec(spec, hook, argIndex)
+}
 
-	if err = spec.TypeByName(hook, &hookFn); err != nil {
+func findBtfFuncParamFromHookWithSpec(spec *btf.Spec, hook string, argIndex int) (*btf.FuncParam, error) {
+	var hookFn *btf.Func
+
+	if err := spec.TypeByName(hook, &hookFn); err != nil {
 		if strings.HasPrefix(hook, "bpf_lsm_") {
 			return nil, fmt.Errorf("failed to find BTF type for hook %q: %w."+
 				"Please check if the hook exists or if your kernel supports BTF for lsm hooks."+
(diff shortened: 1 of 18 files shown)


pkg/btf/btf_test.go
```diff
@@ -178,7 +178,16 @@ func genericTestFindBtfFuncParamFromHook(t *testing.T, hook string, argIndex int
 	return nil
 }
 
-func TestFindBtfFuncParamFromHook(t *testing.T) {
+func RunTestOnSingleBtf(t *testing.T, btfFname string, test func(t *testing.T)) {
+	if err := btfFileExists(btfFname); err != nil {
+		t.Skip(err)
```
ScriptSathi, line 183: If the BTF file does not exist at the specified path, the test is skipped. This means that those tests will be ignored for kernel 4.9 without complexity. But the downside of this, is if the path is wrong, the test is ignored, and we will not be noticed

pkg/btf/validation_test.go
```diff
@@ -48,37 +47,41 @@ var testFiles = []struct {
...
-			for ki := range spec.KProbes {
-				err = ValidateKprobeSpec(btf, spec.KProbes[ki].Call, &spec.KProbes[ki])
-				if checkErr := testFiles[fi].checkFn(t, err); checkErr != nil {
-					t.Fatal(checkErr)
-				}
-			}
-		})
-	}
+	RunTestOnSingleBtf(t, btfFname, genericTestSpecs(testdataPath))
 }
```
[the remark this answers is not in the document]
ScriptSathi: > I'm not sure we need to mess up with the btfFile variable (tends to end bad), why not just pass specific btf path to genericTestSpecs function?

This is because `FindBtfFuncParamFromHook()` use it here

https://github.com/cilium/tetragon/blob/b7f6a8d7579a059d718f59fdbb27dd660ccb406c/pkg/btf/btf.go#L118-L120

 
> like with the example from #3411 would something like this work? or I'm missing the point (likely)
> 
> ```
> func TestFindBtfFuncParamFromHook(t *testing.T) {
> 	t.Run(t, testFindBtfFuncParamFromHook(defaults.DefaultBTFFile))
>         t.Run(t, testFindBtfFuncParamFromHook("/path/to/my/btf/2"))
>         t.Run(t, testFindBtfFuncParamFromHook("/path/to/my/btf/3"))
> }
> ```
> 
> and having `btf, err := btf.LoadSpec(btfFname)` in `testFindBtfFuncParamFromHook`

But this means we will have to call the BTF spec before calling  `FindBtfFuncParamFromHook()`  and pass it to the function. And do the same in generic.go

https://github.com/cilium/tetragon/blob/b7f6a8d7579a059d718f59fdbb27dd660ccb406c/pkg/sensors/tracing/generic.go#L25-L28

IMO, this makes the code less resilient, and I find it a bit weird to call BTF file from generic.go.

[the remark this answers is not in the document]
ScriptSathi: The key point to do thing like this is to move `FindBtfFuncParamFromHook` in a wrapper like this

```
func FindBtfFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
	spec, err := NewBTF()
	if err != nil {
		return nil, err
	}
	return findBtfFuncParamFromHookWithSpec(spec, hook, argIndex)
}
```

So we just test `findBtfFuncParamFromHookWithSpec`
I did not thought about this at first, sorry for this.

I hope those new changes will be better.

pkg/btf/btf_test.go
```diff
@@ -178,34 +178,45 @@ func genericTestFindBtfFuncParamFromHook(t *testing.T, hook string, argIndex int
...
+		hook = "wake_up_new_task"
+		argIndex = 10
+		expectedName = "p"
+		err = genericTestFindBtfFuncParamFromHook(t, spec, hook, argIndex, expectedName)
+		assert.ErrorContains(t, err, fmt.Sprintf("index %d is out of range", argIndex))
+	}
+}
+
+func TestFindBtfFuncParamFromHook(t *testing.T) {
+	t.Run(defaults.DefaultBTFFile, testFindBtfFuncParamFromHook(defaults.DefaultBTFFile))
```
olsajiri, line 219: so we will eventually add more tests in here for other btf paths? could it be part of this change?

ScriptSathi: Sure, I thought I would simply improve this behavior and in another PR upload additionnal BTF files. 
Though, I've added a function `listBtfFiles` that return the list of btf files to test. 

At first, I was thinking of crawling the directory where the below btf files are uploading to make an automatic check. But with this we have less granularity.

https://github.com/cilium/tetragon/blob/a10ac161e2292b54fda6030cb0da582a117888ce/Makefile#L307-L308

The btf files I've added were from `observerFindBTF` and the `TestSpecs`. I believe you were thinking of those files?

pkg/btf/btf_test.go
```diff
@@ -44,6 +46,25 @@ func setupfiles() func(*testing.T, string, ...string) {
 	}
 }
 
+func listBtfFiles() ([]string, error) {
+	kernelVersion, err := forceConfiguredKernelVersion()
+	if err != nil {
+		return nil, err
+	}
+
+	_, testFname, _, _ := runtime.Caller(0)
```
olsajiri, line 55: could we just pass the name through argument?

pkg/btf/btf_test.go
```diff
@@ -44,6 +46,25 @@ func setupfiles() func(*testing.T, string, ...string) {
...
+	testdataPath := filepath.Join(filepath.Dir(testFname), "..", "..", "testdata")
+
+	btfFiles := []string{
+		defaults.DefaultBTFFile,
+		path.Join(defaults.DefaultTetragonLib, "btf"),
+		path.Join(defaults.DefaultTetragonLib, "metadata", "vmlinux-"+kernelVersion),
+		filepath.Join(testdataPath, "btf", "vmlinux-5.4.104+"),
```
olsajiri, line 62: so there will be some directory structure based on test name? could you describe that in detail?

pkg/btf/btf_test.go
olsajiri: > Sure, I thought I would simply improve this behavior and in another PR upload additionnal BTF files.
> Though, I've added a function listBtfFiles that return the list of btf files to test.

I'd suggest you already have the btf files? let's upload them to https://github.com/cilium/tetragon-testdata and put all pieces together

ScriptSathi: I do not since this was not part of the initial request (from https://github.com/cilium/tetragon/pull/3143#discussion_r1948539484). I'll be happy to add any files that you could suggest. 

But what are the needs of having more btf files than those in `listBtfFiles` ?

pkg/btf/btf.go
```diff
@@ -28,6 +28,22 @@ func btfFileExists(file string) error {
...
+		kernelVersion = option.Config.KernelVersion
+		return kernelVersion, nil
+	}
+	var uname unix.Utsname
+	err := unix.Uname(&uname)
+	if err != nil {
+		return "", fmt.Errorf("Kernel version lookup (uname -r) failing. Use '--kernel' to set manually: %w", err)
+	}
+	kernelVersion = unix.ByteSliceToString(uname.Release[:])
+	return kernelVersion, nil
```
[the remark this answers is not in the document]
ScriptSathi: I added it mostly for "code documentation" even though it is unnecessary. But i'll remove this

pkg/btf/btf.go
```diff
@@ -28,6 +28,22 @@ func btfFileExists(file string) error {
 	return err
 }
 
+func forceConfiguredKernelVersion() (string, error) {
```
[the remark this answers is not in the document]
ScriptSathi: Will this name be confusing with the `GetKernelVersion`function ?

https://github.com/cilium/tetragon/blob/fcb4cabe613ee8e92e1ee5858e36691db6cd7e85/pkg/kernels/kernels.go#L61-L67

I thought it would be more explicit to keep the same name as the comment suggest it to be as clear as possible.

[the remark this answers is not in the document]
olsajiri: hum, right.. why don't we use the `GetKernelVersion` directly like:

`_, kernelVersion, err := GetKernelVersion(option.Config.KernelVersion, option.Config.ProcFS)
`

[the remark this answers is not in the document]
ScriptSathi: I didn't want to alter the original code pushed with this commit https://github.com/cilium/tetragon/commit/ec77227304f1ef0b894d082b99c2a08c5eb1627d

Though, I think this commit was made because it was not possible to use `GetKernelVersion` because of the circular dependencie in `kernel.go`

https://github.com/cilium/tetragon/blob/fcb4cabe613ee8e92e1ee5858e36691db6cd7e85/pkg/kernels/kernels.go#L11

pkg/btf/btf_test.go
ScriptSathi: This is absolute paths to the files. Should I add some documentation above the code to describe it more ? 
Basically I simply took the code from elsewhere with no changes

pkg/btf/btf_test.go
ScriptSathi: But how can we do this ? 
We are looking for the project root absolute path so idk how this can be improved.
```go
	_, testFname, _, _ := runtime.Caller(0)
	testdataPath := filepath.Join(filepath.Dir(testFname), "..", "..", "testdata")
```

ScriptSathi:
@olsajiri I've changed the `forceConfiguredKernelVersion`to `minimalGetKernelVersion` and give more details in the commit message. I hope it looks clearer this way.

olsajiri:
> @olsajiri I've changed the `forceConfiguredKernelVersion`to `minimalGetKernelVersion` and give more details in the commit message. I hope it looks clearer this way.

I tried to move the functions a bit among packages and it seems to be ok https://github.com/cilium/tetragon/pull/3481

pkg/btf/btf_test.go
olsajiri: perhaps I don't follow the logic.. looks like you are trying to get to some place in the sources based on test file name, which seems strange to me
if we need to access some source location, I think we should rather use or add variable to defaults package?
in any case more details on what you're trying to do would be helpful, thanks

ScriptSathi: To give more documentation for this one. I copy and paste it from here

https://github.com/cilium/tetragon/blob/f22de557663314e870d5288fd851755752428776/pkg/btf/validation_test.go#L82-L84

Though, there is a helper to do so, but cannot be imported because of another circular import issue.

https://github.com/cilium/tetragon/blob/f22de557663314e870d5288fd851755752428776/pkg/testutils/path.go#L12-L15

pkg/btf/btf_test.go
ScriptSathi: `path.Join(defaults.DefaultTetragonLib, "btf"),` and `path.Join(defaults.DefaultTetragonLib, "metadata", "vmlinux-"+kernelVersion),`

Come from `observerFindBTF`

https://github.com/cilium/tetragon/blob/84212ddf3a8cb0fe5df2e77fe497f11e859996a5/pkg/btf/btf.go#L66
https://github.com/cilium/tetragon/blob/84212ddf3a8cb0fe5df2e77fe497f11e859996a5/pkg/btf/btf.go#L59

And `filepath.Join(testdataPath, "btf", "vmlinux-5.4.104+"),` 
From `TestSpec`
https://github.com/cilium/tetragon/blob/84212ddf3a8cb0fe5df2e77fe497f11e859996a5/pkg/btf/validation_test.go#L66

mtardy:
It's a nit but if you could write acronym in naming respecting this https://google.github.io/styleguide/go/decisions#initialisms that would be awesome, like somethingBTF and not somethingBtf

ScriptSathi:
> I'm ok with that `runtime.Caller` stuff .. sorry for the noise ;-) as far as for the TestSpec running on all btf files, sounds good if it's possible

This was no overkill finally, the TestSpecs test is failing for Ubuntu 22 
https://github.com/cilium/tetragon/actions/runs/13792365836/job/38575300776?pr=3414#step:9:80

Should I disable this tests for now ?

> It's a nit solely on style but if you could write acronym in naming respecting this https://google.github.io/styleguide/go/decisions#initialisms that would be awesome, like somethingBTF and not somethingBtf

Very great idea, thanks a lot. I'll push an extra commit that change `Btf` to `BTF`  in all the project, except for vendors and non Go related code

ScriptSathi:
After pushing the last commit the CI is failing on `TestLabelsDemoApp`

https://github.com/cilium/tetragon/actions/runs/13808479864/job/38624783675?pr=3414#step:6:397

But since this test works on ubuntu-22.04-arm64 and fails only on ubuntu-22.04, I'm not sure if it is related to this commit or not.

lambdanis:
> After pushing the last commit the CI is failing on `TestLabelsDemoApp`
> 
> https://github.com/cilium/tetragon/actions/runs/13808479864/job/38624783675?pr=3414#step:6:397
> 
> But since this test works on ubuntu-22.04-arm64 and fails only on ubuntu-22.04, I'm not sure if it is related to this commit or not.
> 
> Probably related to #1954

I rerun the job and it passed, so it seems the failure was an unrelated flake indeed.

olsajiri:
@ScriptSathi https://github.com/cilium/tetragon/pull/3414/commits/e868e9b14625e06572157e29e5163e1b058a98fb is causing the TestSpecs to fail ... there's some wrt to syscall names on arm? but it needs to be fixed

ScriptSathi:
@olsajiri I think the problem is more complex than it looks.
I pushed a commit to fix the first issue, but now the CI return this error
`validation_test.go:77: call "ksys_lseek" type name ksys_lseek: not found` for test `spec/lseek.yml
https://github.com/cilium/tetragon/actions/runs/13922039785/job/38957360036?pr=3414#step:9:80

Maybe `ksys_lseek` (https://elixir.bootlin.com/linux/v6.13.7/source/fs/read_write.c#L386) is inlined on arm64 and simply does not exists

ScriptSathi:
I've added a bypass for ARM in the last commit. What I think could be another solution would be to rewrite the spec files with other hooks. 

The CI fails here because of network issues it seems. But it is not related to this PR. The `Image CI Build` fails first and the other tests depends on it output. This explains the cascade failure.

pkg/btf/validation.go
```diff
@@ -145,7 +145,11 @@ func ValidateKprobeSpec(bspec *btf.Spec, call string, kspec *v1alpha1.KProbeSpec
 		// NB: this might change in different kernels so if we fail we treat it as a warning
 		prefix := "__x64_sys_"
 		if !strings.HasPrefix(call, prefix) {
-			return &ValidationWarn{s: fmt.Sprintf("could not get the function prototype for %s: arguments will not be verified", call)}
+			if strings.HasPrefix(call, "__arm64_sys_") {
```
mtardy, line 148: fyi: you have `pkg/arch::HasSyscallPrefix()` existing helper and other helpers if you need.

ScriptSathi:
To summarize quicly the latest changes, I have 
- Rebase from main
- Add 3 lasts commits to fix the `TestSpecs` bug

pkg/arch/arch.go
```diff
@@ -81,3 +81,14 @@ func HasSyscallPrefix(symbol string) bool {
 	}
 	return false
 }
+
+// Return the syscall name without arch prefix
+func TrimSyscallPrefix(symbol string) string {
```
olsajiri, line 86: we seem to have CutSyscallPrefix doing the same thing

ScriptSathi: If i'm correct `CutSyscallPrefix` in this case will turn `__x64_sys_something`  to `sys_something` where `TrimSyscallPrefix` will do return `something`

pkg/btf/validation_test.go
```diff
@@ -74,6 +75,10 @@ func genericTestSpecs(ks *ksyms.Ksyms, testdataPath string, btfFName string) fun
 				for ki := range spec.KProbes {
 					err = ValidateKprobeSpec(btf, spec.KProbes[ki].Call, &spec.KProbes[ki], ks)
 					if checkErr := testFiles[fi].checkFn(t, err); checkErr != nil {
+						// We skip these tests on ARM because the current testFiles include inlined functions for ARM
+						if strings.HasPrefix(runtime.GOARCH, "arm") {
```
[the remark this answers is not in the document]
olsajiri: I see, I guess it depends on you, if you want to fix the test first, we don't need that, thanks

ScriptSathi:
Thanks @olsajiri for your review.

There is another unrelated flake on `TestLabelsDemoApp` https://github.com/cilium/tetragon/actions/runs/14023155314/job/39257874085?pr=3414#step:6:339

PR #3415, opened 2025-02-18
Remove kernel version check for LSM Resolve flag

ScriptSathi:
This PR is the follow-up to this comment https://github.com/cilium/tetragon/pull/3143#discussion_r1948562130

In `genericlsm.go` we check the kernel version for the use of the Resolve flag. But it is not reliable because vendors may (like rhel) may have kernels with additional patches, but they are detected as lower than the required version.

### DIFF
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -123,6 +123,11 @@ func FindBtfFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error)
 	}
 
 	if err = spec.TypeByName(hook, &hookFn); err != nil {
+		if strings.HasPrefix(hook, "bpf_lsm_") {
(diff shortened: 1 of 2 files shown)


pkg/btf/btf.go
```diff
@@ -123,6 +123,11 @@ func FindBtfFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error)
 	}
 
 	if err = spec.TypeByName(hook, &hookFn); err != nil {
+		if strings.HasPrefix(hook, "bpf_lsm_") {
```
ScriptSathi, line 126: I think it is acceptable to such a check instead of passing a boolean directly to the function. I wanted to keep the function resilient

olsajiri (approved):
lgtm, thanks

PR #3422, opened 2025-02-19
pkg/bpf: remove redundant check from HasLSMPrograms

### LINKED ISSUE #3392: Problem with applying LSM policies in k8s cluster

anfedotoff:
Checking the contents of `/sys/kernel/security/lsm` file is redundant because later simple LSM program is loaded. File check makes problems when you trying to load LSM sensor in k8s cluster. We can remove this check without losing stability.

Fixes #3392 

Some discussion can be found here: #3404

### DIFF
--- a/pkg/bpf/detect.go
+++ b/pkg/bpf/detect.go
@@ -9,9 +9,7 @@ package bpf
 import (
 	"errors"
 	"fmt"
-	"os"
 	"path/filepath"
-	"strings"
 	"sync"
 	"unsafe"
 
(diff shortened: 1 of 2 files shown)


mtardy (approved):
Yeah thanks I fine it's fine to skip the `/sys/kernel/security/lsm` read and directly try to load a prog since this what we eventually want to do.

kkourt (approved):

PR #3423, opened 2025-02-20
add dentry type

olsajiri:
Adding support to extract dentry full path via new 'dentry' type,
which presents the data as as path_arg argument.

So following spec:

```
  spec:
    kprobes:
    - call: "security_path_unlink"
      syscall: false
      args:
      - index: 1
        type: "dentry"
```
gives you:

`   args{"path_arg":{"path":"/tmp/TestKprobeDentryPath4036433650/001/dentry-unlink872807283", "permission":"-rw-------"}}
`

**Note** dentry type can extract path only to the first mount point.

Note this patch is based on work of David Windsor, adding his SOB.

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -86,6 +86,8 @@ enum {
 	sockaddr_type = 40,
 	socket_type = 41,
 
+	dentry_type = 42,
+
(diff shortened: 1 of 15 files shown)


mtardy (approved):
sure, why not!

PR #3427, opened 2025-02-20
bpf: fix path truncation in cwd and copying path arguments

mtardy:
We bumped the size of the underlying prepend_name buffer to 4096 https://github.com/cilium/tetragon/pull/2764 but many users were still truncating the returned size to 255.

```release-note
Fix path truncations in event values for cwd and path/file function arguments. The function responsible for reading dentry was upgraded to 4096 but some users were still using the previous limitation of 256.
```


mtardy:
Hey @kevsecurity, could you check that 3e397417a5de0e4f2f4065863bd8c9b7a2253fa3 is okay for you? Not sure why those specific lines were added in the first place, it seems wrong.

kkourt:
Thanks @mtardy. Should we add backport tags for this PR?

tpapagian (approved):
LGTM, thanks!

mtardy:
> Thanks @mtardy. Should we add backport tags for this PR?

Mmh yes indeed, https://github.com/cilium/tetragon/pull/2764 is included until v1.2.0 so it seems it should be straightforward to backport until v1.2.0.

kevsecurity (approved):
LGTM

PR #3433, opened 2025-02-24
chore: added verifier tests

AshishNaware:
Fixes 
#229 replace verifier.sh with a go verifier test that uses cilium/ebpf

### Description

This change adds go tests using cilium/ebpf library as an alternative to verifier script. The test closely mimics all the actions of verifier.sh. `-d` flag produces more verbose output than the earlier script.


Makefile
```diff
@@ -277,7 +277,7 @@ bpf-test:
 
 .PHONY: verify
 verify: tetragon-bpf ## Verify BPF programs.
-	sudo contrib/verify/verify.sh bpf/objs
+	sudo TETRAGONDIR=$(CURDIR)/bpf/objs $(shell which go) test contrib/verify/verify_test.go -v -d
```
olsajiri, line 280: why use the -d for default? it's too verbose

contrib/verify/verify_test.go
```diff
@@ -0,0 +1,143 @@
...
+	TETRAGONDIR = "/var/lib/tetragon"
+)
+
+var (
+	DEBUG = flag.Bool("d", false, "debug")
+)
+
+func TestVerifyTetragonPrograms(t *testing.T) {
+
+	kernelVersion, err := strconv.ParseFloat(getKernelVersion(), 64)
```
olsajiri, line 32: you could use kernels.GetKernelVersion ?

contrib/verify/verify_test.go
```diff
@@ -0,0 +1,143 @@
...
+		if strings.HasPrefix(fileName, "bpf_generic_kprobe") || strings.HasPrefix(fileName, "bpf_enforcer") {
+			if err := features.HaveProgramHelper(ebpf.Kprobe, asm.FnOverrideReturn); err != nil {
+				continue
+			}
+		}
+
+		spec, err := ebpf.LoadCollectionSpec(tetragonDir + "/" + fileName)
+		require.NoError(t, err, "failed to parse elf file into collection spec")
+
+		if *DEBUG {
```
olsajiri, line 103: nit, I don't think we normally use all caps variable names, let's use 'debug' ? same for TETRAGONDIR

contrib/verify/verify_test.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		coll, err := ebpf.NewCollection(spec)
+		require.NoError(t, err, "failed to load resources into the kernel")
+
+		defer coll.Close()
+
+		for _, prog := range coll.Programs {
+			require.NotEqual(t, -1, prog.FD())
+			prog.Close()
```
olsajiri, line 116: is this close necessary ? coll.Close() should handle that

contrib/verify/verify_test.go
```diff
@@ -0,0 +1,143 @@
...
+			}
+		}
+
+		coll, err := ebpf.NewCollection(spec)
+		require.NoError(t, err, "failed to load resources into the kernel")
+
+		defer coll.Close()
+
+		for _, prog := range coll.Programs {
+			require.NotEqual(t, -1, prog.FD())
```
olsajiri, line 115: is this check necessary ? if the ebpf.NewCollection call does not return error, then all programs should be loaded no?

contrib/verify/verify.sh
```diff
@@ -1,123 +0,0 @@
...
-
-	# Check if bpf_override_return is available
-	if [[ "$B" == bpf_generic_kprobe* || "$B" == bpf_enforcer* ]]; then
-		if ! bpftool feature probe | grep -q "bpf_override_return"; then
-			echo -e "${YELLOW}bpf_override_return not available, skipping $B ...${NOCOLOR}\n"
-			continue
-		fi
-	fi
-
-	echo -e -n "Verifying $BLUEUNDER$obj$NOCOLOR... "
```
olsajiri, line 95: could we keep this output for less verbose output? not sure it's an option

contrib/verify/verify.sh
```diff
@@ -1,123 +0,0 @@
...
-		echo "Full output left in $OUT"
-		STATUS=1
-	else
-		echo -e "${GREEN}OK${NOCOLOR}"
-		awk '
-/^func/ { in_func=1 }
-/^;/ { if (in_func) { print "  " $0; in_func=0; } }
-/^verification time/ { verify_lines=3 }
-/^/ { if (verify_lines) { verify_lines -= 1; print "  " $0 } }' < "$OUT"
-		rm "$OUT"
```
olsajiri, line 116: not sure we could easily do this output, nor I'm sure we really need that.. I just don't see it in new code, so would be great to have some comment about that (like in changelog)

contrib/verify/verify.sh
AshishNaware: Do we need this? I don't think so. why do we want to check which file is currently being tested? if something goes wrong, we would get the failure messages right away.

contrib/verify/verify_test.go
AshishNaware: i agree. removed this for loop as it seems unnecessary.

contrib/verify/verify_test.go
AshishNaware: updated code with kernels.MinKernelVersion()

contrib/verify/verify.sh
AshishNaware: added comment in changelog

contrib/verify/verify.sh
olsajiri: ok, no need

contrib/verify/verify.sh
olsajiri: > added comment in changelog

@AshishNaware  aah I meant commit changelog (not PR) so we have the info in the git log, thanks

AshishNaware: done. :thumbsup:

AshishNaware: > please add a release-note/* label to the pull request

I am not able to add any labels to this PR

AshishNaware:
> please add a release-note/* label to the pull request

I am not able to add any labels to this PR

Makefile
```diff
@@ -277,7 +278,7 @@ bpf-test:
 
 .PHONY: verify
 verify: tetragon-bpf ## Verify BPF programs.
-	sudo contrib/verify/verify.sh bpf/objs
+	sudo DEBUG=${DEBUG} TETRAGONDIR=${TETRAGONDIR} $(shell which go) test contrib/verify/verify_test.go -v
```
olsajiri, line 281: please use `$(GO)`

Makefile
```diff
@@ -19,6 +19,7 @@ GO_TEST_TIMEOUT ?= 20m
 E2E_TEST_TIMEOUT ?= 20m
 BUILD_PKG_DIR ?= $(shell pwd)/build/$(TARGET_ARCH)
 VERSION ?= $(shell git describe --tags --always --exclude '*/*')
+TETRAGONDIR ?= $(CURDIR)/bpf/objs
```
olsajiri, line 22: no need for `TETRAGONDIR` variable, please use `$(CURDIR)/bpf/objs` directly

contrib/verify/verify_test.go
```diff
@@ -0,0 +1,112 @@
...
+			for _, progSpec := range spec.Programs {
+				fmt.Printf("%s\n", progSpec.Instructions.String())
+			}
+		}
+
+		collection, err := ebpf.NewCollection(spec)
+		require.NoError(t, err, "failed to load resources into the kernel")
+
+		collection.Close()
+
```
olsajiri, line 104: extra line

contrib/verify/verify_test.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
```
olsajiri, line 106: extra line

olsajiri (commented):
left few more comments, but otherwise lgtm, thanks

AshishNaware:
> left few more comments, but otherwise lgtm, thanks

added relevant changes in latest commit

olsajiri (approved):

AshishNaware:
I am not authorized to push to this branch. Can someone merge this PR if there are no more review comments?

PR #3434, opened 2025-02-24
[v1.3] bpf: fix path truncation in cwd and copying path arguments

mtardy:
Backport of https://github.com/cilium/tetragon/pull/3427

We bumped the size of the underlying prepend_name buffer to 4096 https://github.com/cilium/tetragon/pull/2764 but many users were still truncating the returned size to 255.

```release-note
Fix path truncations in event values for cwd and path/file function arguments. The function responsible for reading dentry was upgraded to 4096 but some users were still using the previous limitation of 256.
```


olsajiri:
heya,
should you include something like below to each commit?
`    [upstream commit xxx]
`

mtardy:
> heya, should you include something like below to each commit? ` [upstream commit xxx]`

oups thanks I forgot and in v1.2 as well, let me do that!

tpapagian (approved):

PR #3437, opened 2025-02-25
sensor cleanups

kkourt:
Misc code cleanups. See commits.

```release-note
gRPC: the deprecated sensors API is now removed.
```

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -20,7 +20,6 @@ import (
 	"github.com/cilium/tetragon/pkg/metrics/eventmetrics"
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/process"
-	"github.com/cilium/tetragon/pkg/sensors"
 	"github.com/cilium/tetragon/pkg/tracingpolicy"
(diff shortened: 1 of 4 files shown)


pkg/server/server.go
```diff
@@ -225,24 +218,9 @@ func (s *Server) GetHealth(_ context.Context, request *tetragon.GetHealthStatusR
...
-	for _, s := range *list {
-		sensors = append(sensors, &tetragon.SensorStatus{
-			Name:       s.Name,
-			Enabled:    s.Enabled,
-			Collection: s.Collection,
-		})
-	}
-
-	return &tetragon.ListSensorsResponse{Sensors: sensors}, nil
+	return nil, fmt.Errorf("ListSensors is deprecated")
```
mtardy, line 223: why don't we remove all this code?

kkourt: You mean the gRPC calls?
I tried to, but the `buf`linter said it's not OK because it breaks backwards compatibility: https://buf.build/docs/breaking/rules/#service_no_delete.

🤷🏼

olsajiri (approved):
lgtm

PR #3438, opened 2025-02-26
tetra/getevents: add a --reconnect option

kkourt:
In certain situations (e.g., development) it's useful to have a tetra client constantly running as you are restarting the tetragon agent. Add a --reconnect option in the tetra getevents command to support this.

```release-note
tetra: add a --reconnect option to getevents command
```

### DIFF
--- a/cmd/tetra/getevents/getevents.go
+++ b/cmd/tetra/getevents/getevents.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/cmd/tetra/common"
@@ -39,6 +40,8 @@ type Opts struct {
 	ImaHash       bool
 	PolicyNames   []string
 	CelExpression []string
+	Reconnect     bool
+	ReconnectWait time.Duration
 }
 
 var Options Opts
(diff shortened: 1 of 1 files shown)


mtardy (commented):
Could we make this the default, I think it would benefit everybody? Also it might be confusing with the gRPC retry already a flag.

kkourt:
> Could we make this the default, I think it would benefit everybody? Also it might be confusing with the gRPC retry already a flag.

I wouldn't make this the default because it breaks compatibility.
What does the gRPC retry flag do?

mtardy:
> > Could we make this the default, I think it would benefit everybody? Also it might be confusing with the gRPC retry already a flag.
> 
> I wouldn't make this the default because it breaks compatibility. What does the gRPC retry flag do?

It retries the attempt to connect to the gRPC server in the connection https://github.com/cilium/tetragon/commit/ebae77f7ba52b10bb0738572d8cd78a07a28c67f. It wouldn't retry after a successful attempt as you noticed, it's "inside" the gRPC connection thing.

kkourt:
> > > Could we make this the default, I think it would benefit everybody? Also it might be confusing with the gRPC retry already a flag.
> > 
> > 
> > I wouldn't make this the default because it breaks compatibility. What does the gRPC retry flag do?
> 
> It retries the attempt to connect to the gRPC server [ebae77f](https://github.com/cilium/tetragon/commit/ebae77f7ba52b10bb0738572d8cd78a07a28c67f). It wouldn't retry after a successful attempt as you noticed, it's for the initial connection to the gRPC server.

We now have:

```
      --retry                    Keep retrying even if an error occurred
      --retry-timeout duration   retry timeout in seconds (default 2s)
      --retries int             Connection retries with exponential backoff (default 1)
```

What would you suggest to change?

mtardy:
> ```
>       --retries int             Connection retries with exponential backoff (default 1)
> ```

maybe `--reconnect` with a bool input that tries to restart the connection indefinitely? I find having `--retry` and `--retries` confusing

mtardy:
> How about I change the `--retry` option to `--reconnect` (and `--retry-timeout` to `--reconnect-timeout`). Does that help?

I would keep `--retry` to explicitly show it's retrying when trying to establish a new connection (this can be done on every connection) and add new `--reconnect` that tries to restart the completely connection (with the retries) once it's been closed.

kkourt:
> > How about I change the `--retry` option to `--reconnect` (and `--retry-timeout` to `--reconnect-timeout`). Does that help?
> 
> EDIT: Yes agree same idea
> 
> I would keep `--retries` to explicitly show it's retrying when trying to establish a new connection (this can be done on every connection) and add new `--reconnect` that tries to restart the complet connection (with the retries) once it's been closed (what you want to add).

👍🏼 

Pushed a new version.

cmd/tetra/getevents/getevents.go
```diff
@@ -227,5 +249,7 @@ redirection of events to the stdin. Examples:
 	flags.BoolVar(&Options.ImaHash, "ima-hash", true, "Include ima hashes in compact output")
 	flags.StringSliceVar(&Options.PolicyNames, "policy-names", nil, "Get events by tracing policy names")
 	flags.StringSliceVar(&Options.CelExpression, "cel-expression", nil, "Get events satisfying the CEL expression")
+	flags.BoolVar(&Options.Reconnect, "reconnect", false, "Keep trying to connect even if an error occurred")
+	flags.DurationVar(&Options.ReconnectTimeout, "reconnect-timeout", 2*time.Second, "reconnect timeout")
```
mtardy, line 253: Just a semantic question: isn't a timeout a time until the thing end completely? Shouldn't it be named a backoff or something like that?

mtardy (approved):
Thanks looks good, just a question on the name "timeout"

cmd/tetra/getevents/getevents.go
kkourt: How about `reconnect-wait`?

mtardy: sure

mtardy: sorry for spotting this just now after the first review

olsajiri (approved):
nice, looks useful

PR #3440, opened 2025-02-26
Windows: Build tetragon on Windows

ExceptionalHandler:
### Description
This change attempts to build tetragon on Windows. Following is the strategy

    1. Code that is used in both platform, but is different, is separated by _linux.go and _windows.go suffixes

    2. Code that is unique to each OS is compiled with approrpiate //go:build directive

    3. certain constants are moved to a separate constants package

Caveats:
This is a compilation change. Tetragon.exe on windows runs and shows help message but does not monitor any activity. 

### Changelog

```
Tetragon.exe can now be compiled on Windows
```


mtardy (commented):
Hey Anadi thanks for that massive amount of work! 🪟 

What would you thinkk about maybe splitting this into different commits for example:
- commits for build directive/renaming _linux _windows
- many multiple commits for logical changes that were needed in the codebase

ExceptionalHandler:
@mtardy , That was the original plan, but the logical changes turned out to be minimal and restricted to only seven files. We needed those to build on Windows. Most changes are addition of `//go:build` directive

olsajiri:
heya,
looks great! quick comments before I dive in (might take a while):
- could we keep the linux end of lines, some files seem to have 0x0d 0x0a chars
- there's tetragon binary pushed in and removed, let's not add it in the first place
- second patch seems to be unrelated/separated fix, please add changelog to it

thanks

olsajiri:
> but the logical changes turned out to be minimal and restricted to only seven files. We needed those to build on Windows. Most changes are addition

+1 @mtardy .. I think it should be still possible to move the changes into several separate parts, does not need to be commit per package, but bigger some bigger commits granularity would ease the review

ExceptionalHandler:
Closing this PR. 
New PR on the way with removed _windows.go files and synced again main

mtardy:
> Closing this PR. New PR on the way with removed _windows.go files and synced again main

hey, you could rebase so that we have the history of the discussion in the same PR. You can use `--force-with-lease` to force push to your own branches :)

PR #3445, opened 2025-02-27
Windows: Build tetragon on Windows (Part -1)

ExceptionalHandler:
### Description
Originally was #3440 
This change attempts to build tetragon on Windows. Following is the strategy
    
    1. Code that is used in both platform, but is different, is separated by _linux.go and _windows.go suffixes.
    2. Some packages now therefore have three go files seperating common, linux and windows code
    3. the _windows.go files will be in a separate PR
    4. Code that is unique to each OS is compiled with approrpiate //go:build directive
    5. Certain constants are moved to a separate constants package

### Caveats
This is a compilation change to prepare for Windows build. This does not compile on Windows. 

### Changelog

```
Add build markers and suffixes for Linux specific code
```


ExceptionalHandler:
Closing this and re-opening #3440

ExceptionalHandler:
Reopening as #3440 Cannot be reopened.

mtardy:
So we can't reopen the previous PR because the original branch was overwritten here unfortunately, here's the old PR for history and comments https://github.com/cilium/tetragon/pull/3440.

pkg/constants/constants_linux.go
```diff
@@ -0,0 +1,58 @@
+package constants
```
olsajiri, line 1: I don't understand why we need this file? could you please provide more details on why is this needed?

pkg/grpc/tracing/tracing.go
```diff
@@ -467,9 +466,9 @@ func (msg *MsgGenericTracepointUnix) Retry(internal *process.ProcessInternal, ev
 
 func familyString(family uint16) string {
 	switch family {
-	case unix.AF_INET:
+	case constants.AF_INET:
 		return "AF_INET"
-	case unix.AF_INET6:
+	case constants.AF_INET6:
 		return "AF_INET6"
```
olsajiri, line 472: could you please explain what's the logic here? is there unix.AF_INET equivalent in windows? 

it seems like linux specific parts should go to x_linux.go instead of adding constants package, but I think I dont understand the reason for constant package yet

pkg/reader/caps/caps.go
```diff
@@ -13,15 +13,15 @@ import (
...
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/reader/namespace"
-	"golang.org/x/sys/unix"
 )
 
 var (
 	// Set default last capability based on upstream unix go library
-	cap_last_cap = int32(unix.CAP_LAST_CAP)
+	cap_last_cap = int32(constants.CAP_LAST_CAP)
 	lastCapOnce  sync.Once
```
olsajiri, line 25: same here, is there windows equivalent of `unix.CAP_LAST_CAP` ?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -56,33 +58,6 @@ func init() {
 	observer.RegisterEventHandlerAtInit(ops.MSG_OP_GENERIC_KPROBE, handleGenericKprobe)
 }
 
-const (
```
olsajiri, line 59: why is this move needed? these seem to be bpf program related constants, is it breaking build?

olsajiri (commented):
hi,
looks cleaner, but I'd suggest we split the patchset in following commits:

- single commit with all changes that mark files not to be compiled on windows:  
```
+//go:build !windows
+
```
- single commit with all pure/simple file moves `XXX.go -> XXX_linux.go`
- separate commit for each non-trivial change, like the one adding `constants_linux.go` with details describing the reason/aim in the changelog, AFAICS there's not too may of those

wdyt?

pkg/constants/constants_linux.go
ExceptionalHandler: This is needed for compilation on Windows with minimal changes.  
There are many .go files and packages  that are otherwise platform independent but would not compile on Windows  just because of use of unix. constants. Since go does not offer `#ifdef` based conditional compilation, there are only two ways to resolve this. 
One is to make entire go listing platform specific. That way we run the risk of rendering entire modules unavailable on Windows. We did not want some modules like observer or exec or even bpf unavailable on Windows. 
Second is to make some functions platform specific. We realized the only difference between the function definitions between Linux and Windows would be the use of platform specific constants. So we decided with a third option, which is to make constants package that would just declare some Linux constants in Windows too. That helped us minimize the changes we needed to make to compile _almost_ platform independent code on Windows.

pkg/sensors/tracing/generickprobe.go
ExceptionalHandler: Yes, they break build in an otherwise platform independent code listing

pkg/constants/constants_linux.go
ExceptionalHandler: The unix.WHATEVER  is in `"golang.org/x/sys/unix"` package which is not available on Windows

zodiacon (approved):

olsajiri (approved):
thanks, split is fine

please if possible add commit changelogs (at the moment none of the commits have any) to describe the change, some guidance is in here https://www.kernel.org/doc/html/latest/process/submitting-patches.html#describe-your-changes

Issue #3455, opened 2025-03-03
LSM programs could not load anymore


ScriptSathi:
### What happened?

When trying to start an lsm hook 
```sh
./tetra tracingpolicy add ./examples/tracingpolicy/lsm_track_grandparent.yml
```

I'm facing the following error. 

```
Error: failed to add tracing policy: rpc error: code = Unknown desc = policy handler 'tracing' failed loading policy 'lsm': Unable to load simple LSM BPF program. Does you kernel support the bpf LSM? You can enable LSM BPF by modifying the GRUB configuration /etc/default/grub with GRUB_CMDLINE_LINUX=lsm=bpf
```

### Tetragon Version

main

### Kernel Version

6.11

### Anything else?

```shell
$ cat /sys/kernel/security/lsm
lockdown,capability,landlock,yama,apparmor,bpf,ima,evm
```

Since  https://github.com/cilium/tetragon/pull/3422
LSM program could not load. The problem here is that `detectLSM`always return false. So `bpf.HasLSMPrograms()`also return false and is skipped from all the tests

PR #3456, opened 2025-03-03
fix : `detectLSM` never return true even when LSM exists

ScriptSathi:
A bug was introduced in commit 63f21187fc02a8d179d7ca587b0560fc151f678 The `detectLSM` function never return true when the LSM program is successfully loaded.

Fixes https://github.com/cilium/tetragon/issues/3455

### Description

`detectLSM` function never return true. This leads to `bpf.HasLSMPrograms()` returning false even if the LSM test program was successfully loaded


kevsecurity (approved):
LGTM – good catch.

kevsecurity (changes requested):
I guess the LSM test isn't sufficient and the Ubuntu 20.04 doesn't support LSM but Tetragon believes it does.

ScriptSathi:
Yes ! I think this patch is correct because it fixed the problem for me.
But I do not understand yet where the errors could come from. I'll investigate this.

ScriptSathi:
It seems `TestLSMOpenFile`is failing in the CI.

With the current setup  on my machine where LSM is enabled, the test is skipped because `bpf.HasLSMPrograms` return false. 

With the patch, the test run completely with no errors. So I believe the problem come from elsewhere

olsajiri:
@anfedotoff could you please check on this? thnx

anfedotoff:
Hey guys!
@ScriptSathi, thanks, nice catch! Your patch also works fine on my machine too.


It seems to me, I know the problem... The file check [5a54e09](https://github.com/cilium/tetragon/commit/5a54e098327defce00b8fee54e7c098f6339cb9e), the I remove was not redundant redundant.

Look this is my laptop:

```
 $ cat /sys/kernel/security/lsm 
lockdown,capability,bpf%
```

And this is Github CI agent:

```
Linux fv-az1787-661 6.8.0-1021-azure #25-Ubuntu SMP Wed Jan 15 20:45:09 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
test lsm:  lockdown,capability,landlock,yama,apparmor
```
We see no bpf here. It is really strange for me that we manage to load and attach  LSM bpf program, when ls bpf is not enabled.

I think we need to get back that file check, but that we will have problems in k8s while loading LSM bpf sensor...

ScriptSathi:
100% agree ! This makes no sense to me either. I'm wondering if this is something that can be push to [cilium/ebpf](https://github.com/cilium/ebpf) ?
> We see no bpf here. It is really strange for me that we manage to load and attach LSM bpf program, when lsm bpf is not enabled.

As an alternative for kubernetes, could it work to mount this file on the pod directly ? Something similar to what it already done in the daemonset deployment. It's just a thought but maybe it makes no sens. What do you think ?

https://github.com/cilium/tetragon/blob/fb54043d0c6332ecafffb9cdcee76d7b9c117e1a/install/kubernetes/tetragon/templates/daemonset.yaml#L93-L96

anfedotoff:
> 100% agree ! This makes no sense to me either. I'm wondering if this is something that can be push to [cilium/ebpf](https://github.com/cilium/ebpf) ?

It's a good question. As if there is a deeper problem here or this is by design. No matter what framework we use, it's allowed to load and attach lsm program despite lsm bpf is not enabled. Maybe be we need a kernel patch)?  

>As an alternative for kubernetes, could it work to mount this file on the pod directly ?

Technically, it's not a big problem. You can follow the discussion here #3404. We can mount `/sys/kernel/security` in read-only as a solution.

anfedotoff:
I think, this can be the best solution: https://docs.kernel.org/userspace-api/lsm.html#c.sys_lsm_list_modules. We can use this!

ScriptSathi:
> I think, this can be the best solution: https://docs.kernel.org/userspace-api/lsm.html#c.sys_lsm_list_modules. We can use this!

It's great, but this syscall does not exists on older kernel unfortunately
https://lore.kernel.org/lkml/20230629195535.2590-1-casey@schaufler-ca.com/T/

ScriptSathi:
I suggest revert the commit until we found a better solution. 

> We see no bpf here. It is really strange for me that we manage to load and attach LSM bpf program, when lsm bpf is not enabled.

But this is very annoying situation for this case.

anfedotoff:
> I suggest revert the commit until we found a better solution.
> 
> > We see no bpf here. It is really strange for me that we manage to load and attach LSM bpf program, when lsm bpf is not enabled.
> 
> But this is very annoying situation for this case.

Yes, I was very surprised about this.

Tetragon already has privileges if I 'm correct:

https://github.com/cilium/tetragon/blob/fb54043d0c6332ecafffb9cdcee76d7b9c117e1a/install/kubernetes/tetragon/values.yaml#L38-L52

So, mounting `/sys/kernel/security` in read-only mode from the node is not a security concern in this case.

kevsecurity (approved):

olsajiri:
> It's a good question. As if there is a deeper problem here or this is by design. No matter what framework we use, it's allowed to load and attach lsm program despite lsm bpf is not enabled. Maybe be we need a kernel patch)?

I'm guessing the bpf_lsm.. function to attach to is always there, so the attach works, it's just not called if the bpf lsm is not enabled

olsajiri (approved):
agreed to keep the bpf lsm check

PR #3460, opened 2025-03-05
bpf: verification fixes

kkourt:
See commits.

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -132,6 +132,9 @@ generic_process_event(void *ctx, struct bpf_map_def *tailcals)
 	extract_arg(config, index, &a);
 
 	/* Read out args1-5 */
+	asm volatile("%[index] &= %1 ;\n"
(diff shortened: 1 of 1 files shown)


bpf/process/generic_calls.h
```diff
@@ -652,8 +652,7 @@ FUNC_INLINE int generic_process_filter(void)
 	sel = &msg->sel;
 	current = &msg->current;
 
-	curr = sel->curr;
-	if (curr > MAX_SELECTORS)
+	if (sel->curr > MAX_SELECTORS)
```
kevsecurity, line 658: I'm surprised the compiler hasn't done this optimisation already?

kkourt: Yeah...

One possible explanation is that an optimization for bpf is not an optimization for x86, and it might be challenging to balance those if the optimizations happen before the target-specific backend. BPF probably takes less precendence over x86 or other hw archs.

PR #3470, opened 2025-03-06
docs: fix tracing policy options

z63d:
### Description
The location of the link and description for Uprobe Options is unnatural.
https://tetragon.io/docs/concepts/tracing-policy/options/


mtardy (approved):
Indeed, agree, thank you

PR #3481, opened 2025-03-07
reshufle some packages

olsajiri:
strip kernels package so it's usable from other places and we don't need to have multiple functions returning kernel version


kkourt (approved):
🚀

PR #3485, opened 2025-03-08
helm: enableLsmSensor flag

anfedotoff:
This flag allows to mount (read-only) `/sys/kernel/security` directory in Tetragon container.  `/sys/kernel/security/lsm` file is used to check if LSM BPF is enabled on system.

### Description

It was a long story about how we tried to fix LSM sensor for k8s environment: #3392 #3404 #3456.
The only way is to mount  `/sys/kernel/security directory`  to Tetragon container .  It works fine, I tested on my minikube cluster.

```bash
minikube start --driver=docker --mount --mount-string="/sys/kernel/security:/sys/kernel/security"   --memory=8096 --cpus=8
minikube ssh -- sudo mount bpffs -t bpf /sys/fs/bpf
```

Install Tetragon:

```bash
helm install --namespace kube-system \
        --set tetragon.grpc.address="unix:///var/run/cilium/tetragon/tetragon.sock" \
--set tetragon.enableLsmSensor=true \
tetragon .
```
Demo Pod:

```bash
kubectl create -f https://raw.githubusercontent.com/cilium/cilium/v1.15.3/examples/minikube/http-sw-app.yaml
```
Apply Policy and monitor logs:

```bash
kubectl apply -f ./go/src/github.com/cilium/tetragon/examples/tracingpolicy/lsm_file_open.yaml
kubectl exec -ti -n kube-system ds/tetragon -c tetragon -- tetra getevents -o compact --pods xwing
```
Check policy:
```bash
kubectl exec -ti xwing -- bash -c 'cat /etc/passwd'
```

```release-note/minor
```

### DIFF
--- a/install/kubernetes/tetragon/values.yaml
+++ b/install/kubernetes/tetragon/values.yaml
@@ -247,6 +247,8 @@ tetragon:
   # pod association. This feature depends on cri being enabled.
   cgidmap:
     enabled: false
+  # -- Read-only mount /sys/kernel/security into Tetragon container to check if LSM BPF is enabled on the host.
+  enableLsmSensor: false
 # Tetragon Operator settings
 tetragonOperator:
   # -- Enables the Tetragon Operator.
--- a/docs/content/en/docs/reference/helm-chart.md
+++ b/docs/content/en/docs/reference/helm-chart.md
@@ -80,6 +80,7 @@ To use [the values available](#values), with `helm install` or `helm upgrade`, u
 | tetragon.debug | bool | `false` | If you want to run Tetragon in debug mode change this value to true |
 | tetragon.enableK8sAPI | bool | `true` | Access Kubernetes API to associate Tetragon events with Kubernetes pods. |
 | tetragon.enableKeepSensorsOnExit | bool | `false` | Persistent enforcement to allow the enforcement policy to continue running even when its Tetragon process is gone. |
+| tetragon.enableLsmSensor | bool | `false` | Read-only mount /sys/kernel/security into Tetragon container to check if LSM BPF is enabled on the host. |
 | tetragon.enableMsgHandlingLatency | bool | `false` | Enable latency monitoring in message handling |
 | tetragon.enablePolicyFilter | bool | `true` | Enable policy filter. This is required for K8s namespace and pod-label filtering. |
 | tetragon.enablePolicyFilterCgroupMap | bool | `false` | Enable policy filter cgroup map. |
--- a/install/kubernetes/tetragon/README.md
+++ b/install/kubernetes/tetragon/README.md
@@ -62,6 +62,7 @@ Helm chart for Tetragon
 | tetragon.debug | bool | `false` | If you want to run Tetragon in debug mode change this value to true |
 | tetragon.enableK8sAPI | bool | `true` | Access Kubernetes API to associate Tetragon events with Kubernetes pods. |
 | tetragon.enableKeepSensorsOnExit | bool | `false` | Persistent enforcement to allow the enforcement policy to continue running even when its Tetragon process is gone. |
+| tetragon.enableLsmSensor | bool | `false` | Read-only mount /sys/kernel/security into Tetragon container to check if LSM BPF is enabled on the host. |
 | tetragon.enableMsgHandlingLatency | bool | `false` | Enable latency monitoring in message handling |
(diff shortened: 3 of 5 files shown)


install/kubernetes/tetragon/values.yaml
```diff
@@ -247,6 +247,8 @@ tetragon:
   # pod association. This feature depends on cri being enabled.
   cgidmap:
     enabled: false
+  # -- Read-only mount /sys/kernel/security into Tetragon container to check if LSM BPF is enabled on the host.
+  enableLsmSensor: false
```
lambdanis, line 251: Can we do
```
  lsm:
    enabled: false
```
instead? That's more conventional in Helm, and if we need more LSM sensor-related values in the future, they'll be nicely grouped.

tixxdz:
Hi @anfedotoff why do we need the flag? LSM if available then it should just be used or maybe due to some host mounts?

Another thing is tetragon should just mount security fs inside container, IIRC it is not virtualized, meaning we will always have one super block for securityfs and (last time I checked), if so we should just mount it if it is not already mounted.

tixxdz:
> > Hi @anfedotoff why do we need the flag? LSM if available then it should just be used or maybe due to some host mounts?
> 
> I would take a look at those issues where it was discussed:
> 
> > It was a long story about how we tried to fix LSM sensor for k8s environment: #3392 #3404 #3456.

Ok thanks I did go quickly over them.

So proper fix is first to try within Tetragon itself not helm.

Here https://github.com/cilium/tetragon/blob/main/pkg/bpf/bpffs_linux.go  add logic to do the work.

create a function that does  IsSecurityFSMounted()
1. stat /sys/kernel/security if it exist and a securityfs mount then return true
2. if it exists but not a securityfs mount (which should not happen) systemd ensures that. Warn and return false

then create two other functions MountSecurityFS()  and UnmountSecurityFS()
1. First one mounts the securityfs at /sys/kernel/security with MS_NOSUID|MS_NOEXEC|MS_NODEV flags.
2. Second one unmounts the securityfs

Then in bpf lsm detect just check first one IsSecurityFSMounted(), if true continue with detection and cache results for future. If not mounted explicitly mount it check lsm file content cache result if bpf lsm is enabled in kernel then unmount it in that same bpf lsm logic.

The unmount logic is to keep tetragon clean from any later things, since securityfs contains tpm ima etc. so we just check and unmount.

I think this should work transparently without any flag nor host mounts.

kkourt:
> It was a long story about how we tried to fix LSM sensor for k8s environment: https://github.com/cilium/tetragon/issues/3392  https://github.com/cilium/tetragon/pull/3404 https://github.com/cilium/tetragon/pull/3456.
> The only way is to mount /sys/kernel/security directory to Tetragon container

@anfedotoff can you please create a summary of this long story? It's not clear to me from reading the issues. Specifically, it would be nice to have a summary of the alternatives that were consider and why they do not work.

anfedotoff:
> > It was a long story about how we tried to fix LSM sensor for k8s environment: #3392  #3404 #3456.
> > The only way is to mount /sys/kernel/security directory to Tetragon container
> 
> @anfedotoff can you please create a summary of this long story? It's not clear to me from reading the issues. Specifically, it would be nice to have a summary of the alternatives that were consider and why they do not work.

Hey, sure! Long story short:

## The problem

LSM sensor doesn't work in k8s runtime with default tetragon deployment. The reason is that we need to check the contents of `/sys/kernel/security/lsm` from host. This file is not mounted in Tetragon container. Check always fails.

### Work-around

We can use  helm extra volumes for mounting

```bash
```

### Solutions that doesn't work

When Tetragon tries to find out whether lsm bpf is enabled or not it in addition to path check loads and attach simple LSM program. When Tetragon tries to find out whether lsm bpf is enabled or not it in addition to path check loads and attach simple LSM program.

[sys_lsm_list_modules](https://docs.kernel.org/userspace-api/lsm.html#c.sys_lsm_list_modules) can solve this problem but it is available since 6.8 kernel.

### Solutions that might work

We can have a helm flag for read-only mounting `/sys/kernel/security` from host to Tetragon container. This works fine, but we need to set this flag manually.

Solution with mounting/unmounting securityfs to Tetragon container proposed by  @tixxdz  looks good to me. I don't know if we need some extra functions or module for this (as we have for bpffs). I think we can just check if we have `/sys/kernel/security/` directory . If there is no directory than try to mount it and unmount at the end of `detectLSM`

PR #3488, opened 2025-03-10
Windows: Build tetragon on Windows (Part -2)

ExceptionalHandler:
### Description

This change attempts to port `ktime`, `constants` and `pkg/sensors` package to Windows following changes made via #3445. This is the strategy
    
    1. Windows specific .go files are added in this PR for ktime, constants and sensors package
    2. We bring back `pkg/sensors/unloader/unloader.go` , `pkg/sensors/tracing/policyhandler.go` and `pkg/sensors/sensors.go` and separate only platform specific code in these listings. 
    3. Lot of code, we know not used on Windows is removed. Examples include LSM, multi kprobes etc. 
    4. Functions required to build on Windows, but are not supported return "not supported on windows" error

### Caveats
This is a change to prepare for Windows build and port. This does not compile on Windows yet, as more changes to dependent packages will be added in subsequent PR. 

### Changelog

```
Add windows implementation to ktime, constants and sensor packages. 
```

### DIFF
--- a/pkg/constants/constants_windows.go
+++ b/pkg/constants/constants_windows.go
@@ -1,0 +1,20 @@
+package constants
+
+import (
+	"golang.org/x/sys/windows"
+)
+
+const (
+	AF_INET              = windows.AF_INET
+	AF_INET6             = windows.AF_INET6
+	PERF_MAX_STACK_DEPTH = 0x7f
+	CBitFieldMaskBit34   = 0x400000000
+	CAP_LAST_CAP         = 0x28
+	CAP_CHOWN            = 0
+	AF_UNIX              = windows.AF_UNIX
+	AF_NETBIOS           = windows.AF_NETBIOS
+	AF_IRDA              = windows.AF_IRDA
+	AF_BTH               = windows.AF_BTH
+	CGROUP2_SUPER_MAGIC  = 0x63677270
+	BPF_STATS_RUN_TIME   = 0
+)
--- a/pkg/ktime/ktime_test_windows.go
+++ b/pkg/ktime/ktime_test_windows.go
@@ -1,0 +1,24 @@
+package ktime
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestKtime(t *testing.T) {
+	time1, err := NanoTimeSince(0)
+	assert.NoError(t, err)
+	assert.Greater(t, time1.Milliseconds(), int64(0))
+	time2, err := NanoTimeSince(0)
+	assert.NoError(t, err)
+	assert.GreaterOrEqual(t, time2, time1)
+}
+
+func TestBoottime(t *testing.T) {
+	time1, err := NanoTimeSince(0)
+	assert.NoError(t, err)
+	ktime1, err := DecodeKtime(time1.Nanoseconds(), false)
+	assert.NoError(t, err)
+	assert.Greater(t, ktime1.UnixMilli(), int64(0))
+}
--- a/pkg/ktime/ktime_windows.go
+++ b/pkg/ktime/ktime_windows.go
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package ktime
+
+import (
+	"syscall"
+	"time"
+	"unsafe"
+)
+
+var (
+	dll            = syscall.MustLoadDLL("kernel32.dll")
+	queryCounter   = dll.MustFindProc("QueryPerformanceCounter")
+	queryFrequency = dll.MustFindProc("QueryPerformanceFrequency")
+)
+
+func NanoTimeSince(ktime int64) (time.Duration, error) {
+	nowTime := int64(time.Now().Nanosecond())
+	diff := nowTime - ktime
+	return time.Duration(diff), nil
+}
+
+func Monotonic() (time.Duration, error) {
+	return time.Duration(time.Now().Nanosecond()), nil
+}
+
+func getBootTimeNanoseconds() int64 {
+	var freq, counter int64
+	queryFrequency.Call(uintptr(unsafe.Pointer(&freq)))
+	queryCounter.Call(uintptr(unsafe.Pointer(&counter)))
+	return (counter * 1e9) / freq
+}
+
+func DecodeKtime(ktime int64, monotonic bool) (time.Time, error) {
+	var nowTime int64
+	if monotonic {
+		nowTime = int64(time.Now().Nanosecond())
+	} else {
+		nowTime = getBootTimeNanoseconds()
+	}
+	diff := ktime - nowTime
+	t := time.Now().Add(time.Duration(diff))
+	return t, nil
+}
--- a/pkg/sensors/exec/exec_windows.go
+++ b/pkg/sensors/exec/exec_windows.go
@@ -1,0 +1,149 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package exec
+
+import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
+	"strings"
+
+	"github.com/cilium/tetragon/pkg/api"
+	"github.com/cilium/tetragon/pkg/api/ops"
+	"github.com/cilium/tetragon/pkg/api/processapi"
+	exec "github.com/cilium/tetragon/pkg/grpc/exec"
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/observer"
+	"github.com/cilium/tetragon/pkg/sensors"
+	"github.com/cilium/tetragon/pkg/sensors/exec/userinfo"
+	"github.com/cilium/tetragon/pkg/sensors/program"
+	"github.com/cilium/tetragon/pkg/strutils"
+	"github.com/sirupsen/logrus"
+)
+
+func msgToExecveUnix(m *processapi.MsgExecveEvent) *exec.MsgExecveEventUnix {
+	unix := &exec.MsgExecveEventUnix{}
+	unix.Unix = &processapi.MsgExecveEventUnix{}
+	unix.Unix.Msg = m
+	return unix
+}
+
+func execParse(reader *bytes.Reader) (processapi.MsgProcess, bool, error) {
+	proc := processapi.MsgProcess{}
+	exec := processapi.MsgExec{}
+
+	if err := binary.Read(reader, binary.LittleEndian, &exec); err != nil {
+		logger.GetLogger().WithError(err).Debug("Failed to read exec event")
+		return proc, true, err
+	}
+
+	proc.Size = exec.Size
+	proc.PID = exec.PID
+	proc.TID = exec.TID
+	proc.NSPID = exec.NSPID
+	proc.UID = exec.UID
+	proc.Flags = exec.Flags
+	proc.Ktime = exec.Ktime
+	proc.AUID = exec.AUID
+	proc.SecureExec = exec.SecureExec
+	proc.Nlink = exec.Nlink
+	proc.Ino = exec.Ino
+
+	size := exec.Size - processapi.MSG_SIZEOF_EXECVE
+	if size > processapi.MSG_SIZEOF_BUFFER-processapi.MSG_SIZEOF_EXECVE {
+		err := fmt.Errorf("msg exec size larger than argsbuffer")
+		exec.Size = processapi.MSG_SIZEOF_EXECVE
+		proc.Args = "enomem enomem"
+		proc.Filename = "enomem"
+		return proc, false, err
+	}
+
+	args := make([]byte, size) //+2)
+	if err := binary.Read(reader, binary.LittleEndian, &args); err != nil {
+		proc.Size = processapi.MSG_SIZEOF_EXECVE
+		proc.Args = "enomem enomem"
+		proc.Filename = "enomem"
+		return proc, false, err
+	}
+
+	n := bytes.Index(args, []byte{0x00})
+	if n != -1 {
+		proc.Filename = strings.TrimPrefix(strutils.UTF8FromBPFBytes(args[:n]), "\\??\\")
+		args = args[n+1:]
+	}
+	remFileName := proc.Filename
+	if args[0] == '"' {
+		remFileName = "\"" + proc.Filename + "\""
+	}
+
+	proc.Args = strings.TrimPrefix(strutils.UTF8FromBPFBytes(args), remFileName)
+	proc.Args = strings.TrimPrefix(proc.Args, " ")
+	proc.Flags = api.EventNoCWDSupport
+	return proc, false, nil
+}
+
+func nopMsgProcess() processapi.MsgProcess {
+	return processapi.MsgProcess{
+		Filename: "<enomem>",
+		Args:     "<enomem>",
+	}
+}
+
+func handleExecve(r *bytes.Reader) ([]observer.Event, error) {
+	var empty bool
+
+	m := processapi.MsgExecveEvent{}
+	err := binary.Read(r, binary.LittleEndian, &m)
+	if err != nil {
+		return nil, err
+	}
+	//ToDo: Function Name
+	msgUnix := msgToExecveUnix(&m)
+	msgUnix.Unix.Process, empty, err = execParse(r)
+	if err != nil && empty {
+		msgUnix.Unix.Process = nopMsgProcess()
+	}
+	if err == nil && !empty {
+		err = userinfo.MsgToExecveAccountUnix(msgUnix.Unix)
+		if err != nil {
+			logger.GetLogger().WithFields(logrus.Fields{
+				"process.pid":    msgUnix.Unix.Process.PID,
+				"process.binary": msgUnix.Unix.Process.Filename,
+				"process.uid":    msgUnix.Unix.Process.UID,
+			}).WithError(err).Trace("Resolving process uid to username record failed")
+		}
+	}
+	return []observer.Event{msgUnix}, nil
+}
+
+func msgToExitUnix(m *processapi.MsgExitEvent) *exec.MsgExitEventUnix {
+	return &exec.MsgExitEventUnix{MsgExitEvent: *m}
+}
+
+func handleExit(r *bytes.Reader) ([]observer.Event, error) {
+	m := processapi.MsgExitEvent{}
+	err := binary.Read(r, binary.LittleEndian, &m)
+	if err != nil {
+		return nil, err
+	}
+	msgUnix := msgToExitUnix(&m)
+	return []observer.Event{msgUnix}, nil
+}
+
+type execProbe struct{}
+
+func (e *execProbe) LoadProbe(args sensors.LoadProbeArgs) error {
+	return program.LoadTracepointProgram(args.BPFDir, args.Load, args.Maps, args.Verbose)
+}
+
+func init() {
+	AddExec()
+}
+
+func AddExec() {
+	sensors.RegisterProbeType("execve", &execProbe{})
+
+	observer.RegisterEventHandlerAtInit(ops.MSG_OP_EXECVE, handleExecve)
+	observer.RegisterEventHandlerAtInit(ops.MSG_OP_EXIT, handleExit)
+}
--- a/pkg/sensors/load_windows.go
+++ b/pkg/sensors/load_windows.go
@@ -1,0 +1,131 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package sensors
+
+import (
+	"fmt"
+
+	"github.com/cilium/tetragon/pkg/kernels"
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/option"
+	"github.com/cilium/tetragon/pkg/sensors/program"
+
+	"github.com/sirupsen/logrus"
+)
+
+// Load loads the sensor, by loading all the BPF programs and maps.
+func (s *Sensor) Load(bpfDir string) (err error) {
+	if s == nil {
+		return nil
+	}
+
+	if s.Destroyed {
+		return fmt.Errorf("sensor %s has been previously destroyed, please recreate it before loading", s.Name)
+	}
+	if _, err = observerMinReqs(); err != nil {
+		return fmt.Errorf("tetragon, aborting minimum requirements not met: %w", err)
+	}
+
+	var (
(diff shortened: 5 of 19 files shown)


pkg/ktime/ktime_test_windows.go
```diff
@@ -0,0 +1,24 @@
...
+	time2, err := NanoTimeSince(0)
+	assert.Equal(t, err, nil)
+	assert.GreaterOrEqual(t, time2, time1)
+}
+
+func TestBoottime(t *testing.T) {
+	time1, err := NanoTimeSince(0)
+	assert.Equal(t, err, nil)
+	ktime1, err := DecodeKtime(time1.Nanoseconds(), false)
+	assert.Equal(t, err, nil)
```
kkourt, line 22: nit there is: assert.NoErr(t, err)

pkg/ktime/ktime_windows.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func ToProto(ktime uint64) *timestamppb.Timestamp {
+	return ToProtoOpt(ktime, true)
+}
+
+func ToProtoOpt(ktime uint64, monotonic bool) *timestamppb.Timestamp {
+	decodedTime, err := DecodeKtime(int64(ktime), monotonic)
+	if err != nil {
+		logrus.WithError(err).WithField("ktime", ktime).Warn("Failed to decode ktime")
```
kkourt, line 28: Seems better to return an error here?


As an aside, please  use logger.GetLogger(): https://github.com/cilium/tetragon/blob/84212ddf3a8cb0fe5df2e77fe497f11e859996a5/pkg/logger/log.go#L151
for logging.

pkg/sensors/tracing/lists_windows.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
+)
+
+// isList checks if a value specifies a list, and if so it returns it (or nil if list does not exist)
+func isList(val string, lists []v1alpha1.ListSpec) (bool, *v1alpha1.ListSpec) {
+	return false, nil
+}
+
+func preValidateLists(lists []v1alpha1.ListSpec) (err error) {
+	return fmt.Errorf("not supported on windows")
```
kkourt, line 18: Because, there are no arguments, you could use errors.New() here.
But, I think it would even be better to have a global variable and use that everywhere. I.e.,

```
var (
    notSupporterErr = errors.New("not supported on windows")
)
```

pkg/sensors/tracing/loader_windows.go
```diff
@@ -0,0 +1,6 @@
+package tracing
+
+func createLoaderEvents() error {
```
kkourt, line 3: Is this used somewhere?

pkg/sensors/tracing/policyhandler_windows.go
```diff
@@ -0,0 +1,49 @@
...
+	sections := 0
+	if len(spec.KProbes) > 0 {
+		sections++
+	}
+	if len(spec.Tracepoints) > 0 {
+		sections++
+	}
+	if len(spec.LsmHooks) > 0 {
+		sections++
+	}
```
kkourt, line 42: The original policyhandler.go code has recently changed. See, for example, https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/policyhandler_linux.go#L113-L115.

Were these changes left out intentionally?

Generally, we should avoid duplicating code if we can, so maybe using a helper function would help here.

pkg/sensors/tracing/policyhandler_windows.go
```diff
@@ -0,0 +1,49 @@
...
+	if sections > 1 {
+		return nil, errors.New("tracing policies with multiple sections of kprobes, tracepoints, or lsm hooks are currently not supported " + policyName + namespace)
+	}
+
+	//ToDo: Create Windows specific sensor here
+	return nil, nil
```
kkourt, line 48: As far as I can tell, the windows specific sensor is not created in this PR.
So I think the proper behavior is to return an error if `sections > 0` since none of these are currently supported.

pkg/sensors/exec/config/config_windows.go
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package config
+
+// ExecObj returns the exec object based on the kernel version
+func ExecObj() string {
+	return "process_monitor.sys"
```
kkourt, line 8: How is this file distributed?

pkg/sensors/exec/exec_windows.go
```diff
@@ -0,0 +1,151 @@
...
+	proc.PID = exec.PID
+	proc.TID = exec.TID
+	proc.NSPID = exec.NSPID
+	proc.UID = exec.UID
+	proc.Flags = exec.Flags
+	proc.Ktime = exec.Ktime
+	proc.AUID = exec.AUID
+	proc.SecureExec = exec.SecureExec
+	proc.Nlink = exec.Nlink
+	proc.Ino = exec.Ino
```
kkourt, line 51: Are all these fields relevant for windows?

pkg/sensors/program/loader_windows.go
```diff
@@ -0,0 +1,480 @@
...
+				Prog: prog,
+			},
+		}, nil
+	}
+}
+
+func TracingAttach() AttachFunc {
+	return func(_ *ebpf.Collection, _ *ebpf.CollectionSpec,
+		prog *ebpf.Program, spec *ebpf.ProgramSpec) (unloader.Unloader, error) {
+		return nil, fmt.Errorf("not supported on windows")
```
kkourt, line 145: See previous comment.
In this case, I think we could even have a single `AttachFunc` that just returns a not supported error, and return it.

pkg/sensors/load_windows.go
```diff
@@ -0,0 +1,353 @@
...
+	BPF_PROG_TYPE_SOCK_OPS                = 13
+	BPF_PROG_TYPE_SK_MSG                  = 16
+	BPF_PROG_TYPE_RAW_TRACEPOINT          = 17
+	BPF_PROG_TYPE_FLOW_DISSECTOR          = 22
+	BPF_PROG_TYPE_RAW_TRACEPOINT_WRITABLE = 24
+	BPF_PROG_TYPE_CGROUP_SOCKOPT          = 25
+	BPF_PROG_TYPE_TRACING                 = 26
+	BPF_PROG_TYPE_STRUCT_OPS              = 27
+	BPF_PROG_TYPE_EXT                     = 28
+	BPF_PROG_TYPE_LSM                     = 29
```
kkourt, line 37: Are these the same with linux? Can we move them in a common file?

pkg/sensors/sensors_windows.go
```diff
@@ -0,0 +1,314 @@
...
+
+	allPrograms = progs
+}
+
+func AllPrograms() []*program.Program {
+	return append([]*program.Program{}, allPrograms...)
+}
+
+// sortSensors sort the sensors to enforce orderging constrains
+func sortSensors(sensors []SensorIface) {
```
kkourt, line 302: Is this windows specific? Should we maybe move it to a common file for both archs?

pkg/sensors/sensors_windows.go
```diff
@@ -0,0 +1,314 @@
...
+//
+// Sensors are a mechanism for dynamically loading/unloading bpf programs.
+// Contrarily to low-level facilities like kprobes, sensors are meant to be
+// visible to end users who can enable/disable them.
+//
+// Sensor control operations are done in a separate goroutine which acts as a
+// serialization point for concurrent client requests.
+
+// Sensor is a set of BPF programs and maps that are managed as a unit.
+type Sensor struct {
```
kkourt, line 42: A lot of the code here seems to be the same as the Linux version. The `Sensor` struct seems to be an example. Is there a reason for this code to be duplicated? It's a good idea to avoid code duplication as much as possible.

kkourt (changes requested):
Thanks! Please find some comments bellow.

Generally, let's try to avoid code duplication as much as possible. If it's not possible, then let's at least document why.

pkg/sensors/program/loader_windows.go
```diff
@@ -0,0 +1,480 @@
...
+	bpf.SetExecCollection(coll)
+
+	collMaps := map[ebpf.MapID]*ebpf.Map{}
+	// we need a mapping by ID
+	for _, m := range coll.Maps {
+
+		info, err := m.Info()
+		if err != nil {
+			logger.GetLogger().WithError(err).WithField("map", m.String()).Warn("failed to retrieve BPF map info")
+			continue
```
olsajiri, line 307: linux code does `break` in here, does m.Info() work on windows? if not we could remove this completely for now

pkg/sensors/program/loader_windows.go
```diff
@@ -0,0 +1,480 @@
...
+		}
+		id, available := info.ID()
+		if !available {
+			logger.GetLogger().WithField("map", m.String()).Warn("failed to retrieve BPF map ID, you might be running <4.13")
+			continue
+		}
+		collMaps[id] = m
+
+		//ToDo: Uncomment after map pinning issue is fixed
```
olsajiri, line 316: also linkPin function is removed, is that not supported as well?

pkg/sensors/program/loader_windows.go
```diff
@@ -0,0 +1,480 @@
...
+	loadOpts *LoadOpts,
+	verbose int,
+) (*LoadedCollection, error) {
+
+	coll, err := ebpf.LoadCollection(load.Name)
+	if err != nil {
+		logger.GetLogger().WithError(err).WithField("Error ", err.Error()).Warn(" Failed to load Native Windows Collection ")
+		return nil, err
+	}
+	bpf.SetExecCollection(coll)
```
olsajiri, line 298: what's this function doing? I couldn't find it in ebpf sources (checked the windows branch)

pkg/sensors/sensors_windows.go
olsajiri: yes, when you make a diff of windows and linux version it's just `standardTypes` map plus one other change that suggests that the sensor.go used for windows version is missing last upstream change:
 4f6f97cb6b6d tetragon: Pass sensor maps to program loader

pkg/sensors/load_windows.go
```diff
@@ -0,0 +1,353 @@
...
+
+	l.WithField("name", s.Name).Info("Loading sensor")
+	if s.Loaded {
+		return fmt.Errorf("loading sensor %s failed: sensor already loaded", s.Name)
+	}
+
+	if err = s.FindPrograms(); err != nil {
+		return fmt.Errorf("tetragon, aborting could not find BPF programs: %w", err)
+	}
+
```
olsajiri, line 136: so the linux version calls loadMap in here for each map.. this perhaps skipped because of the map pinning issues? then would great to have some more comments on how the load looks like for wondow ;-)

pkg/ktime/ktime_windows.go
ExceptionalHandler: This function is common between Linux and Windows, only implementation differs. Changing prototype here without changing calling code.

pkg/sensors/sensors_windows.go
ExceptionalHandler: About code duplication. yes, there is a lot of common code that is duplicated. I have duplicated code and created separate _linux.go files where I think we will need to make big changes based on how ebpf on windows evolves. This include structures, enums and  `BPF_PROG_TYPE` constants. Windows has it's own `BPF_PROG_TYPE` consts, which may or may not have the same values. 
I have prioritized not making any changes to _linux code further down the road, as compared to making frequent changes in common code. We can move common code in common files when we find the 'real common code' between the two OS. 

But I am very happy to merge common code if team prioritizes code reuse.

pkg/sensors/exec/exec_windows.go
ExceptionalHandler: We set these to keep the protobuf event in parity with Linux

pkg/ktime/ktime_windows.go
jrfastab: I think a follow on PR to change the common code and each implementation to return the error would be fine.

pkg/sensors/program/loader_windows.go
lmb: What is the issue so I can look into it?

ExceptionalHandler: Pinned ringbuffer maps had a problem with out undocumented use of device io. I will verify once the sync ringbuffer API is made available.

pkg/sensors/program/loader_windows.go
ExceptionalHandler: This code is changing. I will add it it in a future PR.

pkg/sensors/load_windows.go
ExceptionalHandler: In windows, we load collection directly and do not load specs. The collection loads maps for us.

pkg/sensors/exec/config/config_windows.go
ExceptionalHandler: This change is removed from the PR. This will be distributed with the tetragon msi.

pkg/sensors/load_windows.go
ExceptionalHandler: I have removed types not used in Windows.

pkg/sensors/sensors_windows.go
ExceptionalHandler: I have merged latest upstream changes in Windows.

pkg/sensors/program/loader_windows.go
```diff
@@ -0,0 +1,480 @@
...
+
+func MultiKprobeAttach(load *Program, bpfDir string) AttachFunc {
+	return func(coll *ebpf.Collection, collSpec *ebpf.CollectionSpec,
+		prog *ebpf.Program, spec *ebpf.ProgramSpec) (unloader.Unloader, error) {
+		return nil, fmt.Errorf("not supported on windows")
+
+	}
+}
+
+func LoadTracepointProgram(bpfDir string, load *Program, verbose int) error {
```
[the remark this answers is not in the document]
ExceptionalHandler: I have added single attach function

pkg/sensors/exec/config/config_windows.go
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package config
+
+// ExecObj returns the exec object based on the kernel version
+func ExecObj() string {
```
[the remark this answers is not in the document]
ExceptionalHandler: I have removed this from PR

pkg/constants/constants_windows.go
```diff
@@ -1,20 +1,20 @@
...
+	CBitFieldMaskBit34   = 0x400000000
+	CAP_LAST_CAP         = 0x28
+	CAP_CHOWN            = 0
+	AF_UNIX              = windows.AF_UNIX
+	AF_NETBIOS           = windows.AF_NETBIOS
+	AF_IRDA              = windows.AF_IRDA
+	AF_BTH               = windows.AF_BTH
+	CGROUP2_SUPER_MAGIC  = 0x63677270
+	BPF_STATS_RUN_TIME   = 0
+)
```
mtardy, line 20: just fyi on this, I read through these patches very quickly, but you have somehow a NULL change here on this file? Maybe it's a whitespace change?

lmb: https://github.com/cilium/ebpf/commit/2b8ce32946a5a38840e907894a4df55bd3aed815 could help

ExceptionalHandler: Yes, I removed the CRLF. Which commit do you see that in, it is a new file?

pkg/sensors/exec/config/config_windows.go
olsajiri: hum, not sure what that means, how can we see that code?

pkg/constants/constants_windows.go
olsajiri: I see that in following commits:
- https://github.com/cilium/tetragon/pull/3488/commits/f345993fc0d43db51a92bb23cd59a4014dab90e3
- https://github.com/cilium/tetragon/pull/3488/commits/b274b117984ba63f6c633b577ee2388f0493f88c   (pkg/sensors/tracing/loader_windows.go)

please add the commit @lmb suggested above

olsajiri: > I see that in following commits:
> 
>     * [f345993](https://github.com/cilium/tetragon/commit/f345993fc0d43db51a92bb23cd59a4014dab90e3)
> 
>     * [b274b11](https://github.com/cilium/tetragon/commit/b274b117984ba63f6c633b577ee2388f0493f88c)   (pkg/sensors/tracing/loader_windows.go)
> 
> 
> please add the commit @lmb suggested above

ah you made all those changes that in last patch.. please squash those changes into original patches

pkg/sensors/load_windows.go
```diff
@@ -0,0 +1,356 @@
...
+	for _, p := range s.Progs {
+		if p.LoadState.IsLoaded() {
+			l.WithField("prog", p.Name).Info("BPF prog is already loaded, incrementing reference count")
+			p.LoadState.RefInc()
+			continue
+		}
+
+		if err = observerLoadInstance(bpfDir, p); err != nil {
```
olsajiri, line 147: I think I mentioned it already, but it seems you did not take the latest changes to make the window copy,
there's following change in latest source:
```
-               if err = observerLoadInstance(bpfDir, p); err != nil {
+               if err = observerLoadInstance(bpfDir, p, s.Maps); err != nil {
```

which I can't see in the windows source

pkg/sensors/load_windows.go
```diff
@@ -0,0 +1,356 @@
...
+	}
+	return nil
+}
+
+func (s *Sensor) policyDir() string {
+	if s.Namespace == "" {
+		return sanitize(s.Policy)
+	}
+	return fmt.Sprintf("%s:%s", s.Namespace, sanitize(s.Policy))
+}
```
olsajiri, line 54: there's been following fix in latest sources that came from 5f2bc5b78d50 policyconf: expose mode in grpc

```
 func (s *Sensor) policyDir() string {
-       if s.Namespace == "" {
-               return sanitize(s.Policy)
-       }
-       return fmt.Sprintf("%s:%s", s.Namespace, sanitize(s.Policy))
+       return tracingpolicy.PolicyDir(s.Namespace, s.Policy)
 }
```

seems like this could be shared by both linux and windows, but that could be done later,
for now let's make sure both linux and windows version have the latest code/changes

kkourt (changes requested):

kkourt (changes requested):
Apologies, some I made an error and some text was not submitted in the review.
Adding it here.

It seems to me that there is a lot of code duplication between the linux and the windows version. For example, looking at `load_windows.go` and `load_linux.go`, there are 15 and 17 functions, respectively. Out of those only 3 seem to have differences. Can we just add all the common code in a `load.go` file?

```
$ git grep ^func pkg/sensors/load_windows.go | wc -l
15
$ git grep ^func pkg/sensors/load_linux.go | wc -l
17
$  diff -u pkg/sensors/load_linux.go pkg/sensors/load_windows.go | grep func
-func (s *Sensor) setMapPinPath(m *program.Map) {
-func (s *Sensor) loadMap(bpfDir string, m *program.Map) error {
 func mergeSensors(sensors []*Sensor) *Sensor {
 func observerMinReqs() (bool, error) {
 ```

**Comment by ExceptionalHandler** [comment]
> It seems to me that there is a lot of code duplication between the linux and the windows version. For example, looking at `load_windows.go` and `load_linux.go`, there are 15 and 17 functions, respectively. Out of those only 3 seem to have differences. Can we just add all the common code in a `load.go` file?
 

'load.go' is added in latest commit.

File: pkg/sensors/load_windows.go [hunk 22]
```diff
@@ -0,0 +1,131 @@
...
+	l.WithFields(logrus.Fields{
+		"sensor": s.Name,
+		"maps":   loadedMaps,
+		"progs":  loadedProgs,
+	}).Infof("Loaded BPF maps and events for sensor successfully")
+	s.Loaded = true
+	return nil
+}
+
+func observerLoadInstance(bpfDir string, load *program.Program, maps []*program.Map) error {
```
olsajiri, line 95: note, linux version has some additional code (with unload for tracepoint),
but I think it can be removed (need to check) and we could move observerLoadInstance to load.go,
but I'll put this on my todo and submit that as follow up

pkg/sensors/sensor_types_windows.go
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
```
olsajiri, line 1: nit, could you plz rename rename 'sensor' to 'sensors' in the path, so we'd have:

```
pkg/sensors/sensors_types_linux.go      
pkg/sensors/sensors_types_windows.go    
```

olsajiri (approved):
looks good, left one minor comment, thanks

pkg/sensors/tracing/policyhandler_windows.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/cilium/tetragon/pkg/policyfilter"
+	"github.com/cilium/tetragon/pkg/sensors"
+	"github.com/cilium/tetragon/pkg/tracingpolicy"
+)
+
+func (h policyHandler) PolicyHandler(
+	policy tracingpolicy.TracingPolicy,
+	policyID policyfilter.PolicyID,
+) (sensors.SensorIface, error) {
+	return nil, errors.New("not supported on windows")
```
kkourt, line 18: Maybe use notSupportedWinErr here?

pkg/ktime/ktime_windows.go
```diff
@@ -0,0 +1,65 @@
...
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+var (
+	dll            = syscall.MustLoadDLL("kernel32.dll")
+	queryCounter   = dll.MustFindProc("QueryPerformanceCounter")
+	queryFrequency = dll.MustFindProc("QueryPerformanceFrequency")
+)
+
+func ToProto(ktime uint64) *timestamppb.Timestamp {
```
kkourt, line 21: Functions `ToProto`, `ToProtoOpt`, and `DiffKtime` seem to be the same across both versions. Any reason for not adding them to `pkt/ktime/ktime.go`?

kkourt (approved):
Just a couple of nits.

LGTM, thanks!

PR #3491, opened 2025-03-12
deprecate followFD and friends

kkourt:
This PR deprecates FollowFD, UnfollowFD, CopyFD. The
reason is that these actions are very difficult to implement correctly
(e.g., is hard to rack file descriptors across processes) and they,
also, have limited use.

This commit marks their use as deprecated in the ActionSelector struct
as well as well as in our docs. They are scheduled to be removed in 1.5,
but this might change based on user feedback.

```release-note
tracingpolicy: FollowFD, UnfollowFD, and CopyFD actions are deprecarted
```


pkg/testutils/wait.go
mtardy, line 1: is this expected to go with this PR? it's in your `authocore: make crds` commit.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -206,6 +206,8 @@ type ArgSelector struct {
 type ActionSelector struct {
 	// +kubebuilder:validation:Enum=Post;FollowFD;UnfollowFD;Sigkill;CopyFD;Override;GetUrl;DnsLookup;NoPost;Signal;TrackSock;UntrackSock;NotifyEnforcer;CleanupEnforcerNotification
 	// Action to execute.
+	// NOTE: actions FollowFD, UnfollowFD, and CopyFD are marked as deprecated and planned to
```
mtardy, line 209: could you try `kubebuilder:deprecatedversion`?

```
// +kubebuilder:deprecatedversion
marks this version as deprecated.
```

It's a bit unclear to me if that's appropriate for a field (apparently OpenAPIv3 support deprecated fields) or the whole CRD object.

pkg/testutils/wait.go
kkourt: nope, sorry :( will remove.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
mtardy: ah sorry indeed I didn't think correctly on this one, I went straight into deprecating a fields, nevermind!

pkg/testutils/wait.go
kkourt: should be fixed now!

kevsecurity (approved):

tpapagian (approved):

mtardy (approved):

PR #3509, opened 2025-03-14
ci: Refactor linters, formatters and generators checks

lambdanis:
The main change here is splitting the `generated-files.yaml` workflow into three: `lint-protobuf.yaml`, `lint-k8s.yaml` and `generated-docs.yaml`. By scoping the workflows to specific parts of the codebase and providing consistent names and output, we make it easier to understand the failures. I also made small changes to the triggers and the steps, see commits for details. We will stop running some unnecessary jobs on every push, although these are all cheap workflows, so maybe not that impactful.

Unfortunately renaming workflows and jobs breaks Github Actions history. To my best knowledge, none of Tetragon contributors are relying on it at the moment.


Makefile
```diff
@@ -464,6 +464,21 @@ lint-metrics-md: metrics-docs ## Check if metrics documentation is up to date.
 		false; \
 	fi
 
+.PHONY: validate
+validate: ## Convenience target running linters, formatters and generators across the codebase.
```
mtardy, line 461: any chance we can run some of those in parallel?

mtardy (approved):
Really love the change in workflow files, especially triggering them correctly.

My only question would be if the new Makefile could be accelerate by running some of the generation in parallel if they don't affect each others, like when declared as a target dep like this?

```Makefile
target: dep1 dep2
    something
```

Makefile
lambdanis: good point, yeah, I'll update :+1:

lambdanis:
One more fix because I forgot `/**` in the paths list. Now it should be all good.

PR #3512, opened 2025-03-15
pkg/bpf: mount securityfs to check lsm bpf

anfedotoff:
To check if lsm bpf is enabled /sys/kernel/security/lsm is used. Sometimes, securityfs is not mounted (eg. k8s, container deployments).

Detailed information in this [post](https://github.com/cilium/tetragon/pull/3485#issuecomment-2718236591).

### How to test

```
docker run --name tetragon --rm -it                  \
   --pid=host --cgroupns=host --privileged          \
   -v /sys/kernel/btf/vmlinux:/var/lib/tetragon/btf -v /home/fedotoff/go/src/github.com/cilium/tetragon/examples/tracingpolicy:/tracingpolicy \
cilium/tetragon \
/usr/bin/tetragon --debug --tracing-policy /tracingpolicy/lsm_file_open.yaml
```

### DIFF
--- a/pkg/bpf/detect_linux.go
+++ b/pkg/bpf/detect_linux.go
@@ -13,6 +13,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"syscall"
 	"unsafe"
 
 	"github.com/cilium/ebpf"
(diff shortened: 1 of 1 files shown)


pkg/bpf/detect_linux.go
```diff
@@ -231,6 +232,37 @@ func detectLSM() bool {
 	if features.HaveProgramType(ebpf.LSM) != nil {
 		return false
 	}
+	_, err := os.Stat("/sys/kernel/security")
+	if err != nil {
+		if os.IsNotExist(err) {
+			if err := os.Mkdir("/sys/kernel/security", 0755); err != nil {
```
olsajiri, line 238: is this needed? IIUC the `/sys/kernel/security` directory should be always there (I checked 4.19)
also I wonder we'd be allowed to create that directory anyway

PR #3524, opened 2025-03-18
main: Split k8s watchers into process metadata and policy

lambdanis:
Commit https://github.com/cilium/tetragon/commit/7dc13f82781ec6d0258e398281f847217d252513 ("crdwatcher: Reuse watcher.K8sWatcher to watch TracingPolicy")
reorganized the order of execution in the main function to start watching all
Kubernetes resources at once. In hindsight, this was a bit shortsighted. There
are two purposes Tetragon is watching Kubernetes API for:
1. retrieving Kubernetes metadata for processes
2. managing policies

These two should kick off at different times in the startup order. The former
should happen before the sensors are loaded, as otherwise events will be stuck
waiting for Kubernetes metadata. The latter should happen after the sensors are
loaded, as otherwise existing policies will fail to load on the first attempt.

To solve the problem, split informers used for these two purposes into separate
K8sWatcher instances, one started before the sensors are loaded, and one after.
Both use the same watcher.Watcher interface. Currently the "process metadata"
watcher watches only k8s built-in resources, and the "policy" watcher - only
Tetragon CRDs. However, it's quite likely to change in the future - the split
here is based on the purpose, not where the resources are defined.

The previous change wasn't released yet, so marking the PR as `release-note/misc`.

### DIFF
--- a/cmd/tetragon/main_linux.go
+++ b/cmd/tetragon/main_linux.go
@@ -55,7 +55,6 @@ import (
 	"github.com/cilium/tetragon/pkg/unixlisten"
 	"github.com/cilium/tetragon/pkg/version"
 	"github.com/cilium/tetragon/pkg/watcher"
-	k8sconf "github.com/cilium/tetragon/pkg/watcher/conf"
 	"github.com/cilium/tetragon/pkg/watcher/crdwatcher"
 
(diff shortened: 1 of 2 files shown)


mtardy (approved):
Thank you for the fix!

PR #3525, opened 2025-03-18
api: Copy API reference into docs 

lambdanis:
Commit https://github.com/cilium/tetragon/commit/eea39fd71e5b459675c048ede3160e2ec72ea4e5 ("api: build using buf") migrated proto API docs generation
to use buf, but also removed the script copying the generated README into the
docs directory from the Makefile. We still want API reference on the docs page,
so let's bring it back.

cc @will-isovalent


mtardy (approved):
agree

Issue #3492, opened 2025-03-12
TestSpecs is failing when using default BTF file for syslseek.yaml


ScriptSathi:
### What happened?

When updating  `TestSpecs` test to run on multiple BTF files in https://github.com/cilium/tetragon/pull/3414, it seems that the`specs/syslseek.yaml` test fails on Ubuntu 22 in the CI

https://github.com/cilium/tetragon/blob/e8f9dafd9f1fb6f540e71ad306daf960eb6dcef9/pkg/btf/validation_test.go#L47

https://github.com/cilium/tetragon/actions/runs/13808479848/job/38624371178?pr=3414#step:9:80 

### Tetragon Version

main branch

### Kernel Version

CI - Ubuntu 22

PR #3526, opened 2025-03-20
ci: In "Tetragon Go Test" add vmlinux in artifact when test fails

ScriptSathi:
To help debugging between arch, I think it is a great idea to upload the vmlinux file when this test fails

This will help fixing https://github.com/cilium/tetragon/issues/3492

### Motivation

`TestSpecs` uses `ksys_lseek`  symbol which is inlined (I think) on ARM but not on x86_64. If we upload the vmlinux, we can load the BTF spec and check what hooks are shared
https://github.com/cilium/tetragon/pull/3414#issuecomment-2733004356

### Description

Kprobes are not exactly the same between ARM and x86_64. To help understand why some tests fail on one architecture and not the other, it can be helpful to use the vmlinux file directly from the CI instead of deploying a specific VM with the same architecture for debugging.

### DIFF
--- a/.github/workflows/gotests.yml
+++ b/.github/workflows/gotests.yml
@@ -88,3 +88,11 @@ jobs:
         name: tetragon-bugtool
         path: /tmp/tetragon-bugtool*
         retention-days: 5
+
+    - name: Upload vmlinux file
+      if: failure()
+      uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+      with:
+        name: btf-vmlinux-${{ matrix.os }}
+        path: /sys/kernel/btf/vmlinux
+        retention-days: 5


.github/workflows/gotests.yml
```diff
@@ -88,3 +88,11 @@ jobs:
...
         path: /tmp/tetragon-bugtool*
         retention-days: 5
+
+    - name: Upload vmlinux file
+      if: failure()
+      uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
+      with:
+        name: vmlinux-${{ matrix.os }}
+        path: /sys/kernel/btf/vmlinux
+        retention-days: 5
```
kkourt, line 98: These are just the BTF definitions for the kernel, not the kernel itself, right?

Note that you can get the btf files from the lvh images.
For example:
```
x=$(docker create --platform linux/arm64  quay.io/lvh-images/kernel-images:6.12-20250319.212144)
docker cp $x:/data/kernels /tmp/deleteme
ls /tmp/deleteme/6.12/boot/btf-6.12.19
```

ScriptSathi: Yes it's just the 1.5MB BTF file

Thanks for this workaround solution ! I'll have a look

kkourt: Then, let's name the file so that it's clear that it's the btf file? (e.g., btf-vmlinux-${{ matrix.os }})?

kkourt (approved):

Issue #3316, opened 2025-01-17
Add pod annotation to the process execution event


cy83rc0llect0r:
Hi,

I want to add pod annotation as an extra field in the process execution event.
I thought that if I set the value of enable-pod-info to true, I could also have the pod's annotations. However, when I do this, the logs are not recorded at all, and for example, the 'whoami' log, which was recorded before, is no longer recorded.
Does anyone have a solution?

mtardy:
Hey, sorry this was left without a response. Indeed we only collect the pod labels in the event. If you think we should collect the annotation, we could discuss this: what was your use case, why did you need the annotations to be present?

cy83rc0llect0r:
Thank you for your response @mtardy 
I have multiple use cases. For instance, since Tetragon does not capture the Kubernetes user who execs into a pod, I have implemented an admission webhook that triggers a function to annotate the pod with the Kubernetes username and timestamp.
If it were possible to include this annotation as an extra field in the process execution event, identifying the user who executed into the pod would become significantly easier.
Would love to hear your thoughts on this approach.

mtardy:
That indeed is a sensible use case. If you'd like, you could try to write some patches to add this to Tetragon, I think you could get inspired by the way we add labels to the events since this is quite similar.

And if you can (schedule and timezone) you can even join the [next community call](https://isogo.to/tetragon-meeting-notes) and talk about your feature to show why it makes sense. If you cannot no worry but that could be helpful to get more guidance.

PR #3527, opened 2025-03-20
Enhance Tetragon Events with Pod Annotations Support

cy83rc0llect0r:
Enhancement #3316

### Description
This pull request implements support for including Kubernetes pod annotations in Tetragon event logs, addressing the enhancement request in issue #3316. The feature introduces an optional `--enable-pod-annotations` flag, allowing users to enrich process execution events with pod annotations for improved workload visibility and debugging in Kubernetes environments. The change is split into two commits for clarity:

1. **Core implementation**: Adds the `pod_annotations` field to the Tetragon protobuf definition and integrates it into the event generation logic.
2. **Helm chart support**: Provides a configuration option in the Helm chart to enable the feature.

The feature is disabled by default to maintain backward compatibility and can be activated via Helm or command-line flag.

**Use Case Example**: This enhancement enables tracking of Kubernetes users who execute commands in pods. For instance, a mutating admission webhook could annotate pods with the username (e.g., `exec-user`) and timestamp (e.g., `exec-time`) of an `exec` operation. With `--enable-pod-annotations` enabled, Tetragon would then include these annotations in its event logs, providing valuable audit trails for security and troubleshooting.

### Changelog
```release-note
Added support for including Kubernetes pod annotations in Tetragon event logs with the new `--enable-pod-annotations` flag (disabled by default). Enable it via Helm with `enablePodAnnotations=true`.

### CODE REVIEW

File: install/kubernetes/tetragon/templates/tetragon_configmap.yaml [hunk 1]
```diff
@@ -86,3 +86,6 @@ data:
   cri-endpoint: "unix://{{ .Values.tetragon.cri.socketHostPath }}"
 {{- end }}
   enable-cgidmap: {{ .Values.tetragon.cgidmap.enabled | quote }}
+{{- if .Values.tetragon.enablePodAnnotations }}
+  enable-pod-annotations: "true"
```
kkourt, line 90: I guess this works because the agent flag is by default false, but if that changes then a user setting `enablePodAnnotations` to false in helm would not get the expected result.

How about:
> `enable-pod-annotations: {{- if .Values.tetragon.enablePodAnnotations }}`?

install/kubernetes/tetragon/values.yaml
```diff
@@ -47,6 +47,7 @@ tetragon:
   # extraEnv:
   #   - name: foo
   #     value: bar
+  enablePodAnnotations: false
```
kkourt, line 50: I believe @lambdanis was advocating for having one object per configuration.

So:
```
podAnnotations:
    enabled: true
```

kkourt (commented):
Thanks for the PR!

Please find some comments below. Also, can you rebase instead of a merge commit?

install/kubernetes/tetragon/templates/tetragon_configmap.yaml
cy83rc0llect0r: Resolved all comments:
- Rebasing done on `upstream/main` to remove merge commit as requested.
- Fixed `tetragon_configmap.yaml` to always set `enable-pod-annotations` with `.Values.tetragon.podAnnotations.enabled`.
- Moved `enablePodAnnotations` to `tetragon.podAnnotations.enabled` in `values.yaml` for better structure.
- Excluded local changes to `podinfo_test` from the PR.
- Tested with Helm:
  - Deployed with `tetragon.podAnnotations.enabled: true` → annotations like `test-key=test-value` appear in logs.
  - Deployed with `tetragon.podAnnotations.enabled: false` → no annotations in logs.

Pushed to my fork (`github.com/cy83rc0llect0r/tetragon`). Ready for review!

install/kubernetes/tetragon/values.yaml
cy83rc0llect0r: Resolved all comments:
- Rebasing done on `upstream/main` to remove merge commit as requested.
- Fixed `tetragon_configmap.yaml` to always set `enable-pod-annotations` with `.Values.tetragon.podAnnotations.enabled`.
- Moved `enablePodAnnotations` to `tetragon.podAnnotations.enabled` in `values.yaml` for better structure.
- Excluded local changes to `podinfo_test` from the PR.
- Tested with Helm:
  - Deployed with `tetragon.podAnnotations.enabled: true` → annotations like `test-key=test-value` appear in logs.
  - Deployed with `tetragon.podAnnotations.enabled: false` → no annotations in logs.

Pushed to my fork (`github.com/cy83rc0llect0r/tetragon`). Ready for review!

install/kubernetes/tetragon/templates/tetragon_configmap.yaml
kkourt: I don't think this is resolved.

install/kubernetes/tetragon/values.yaml
kkourt: same for this comment, I don't think this is resolved.

install/kubernetes/tetragon/templates/tetragon_configmap.yaml
cy83rc0llect0r: @kkourt 
Could you please provide more details? I’ve tested various scenarios, and everything seems to be working as expected:

- Installed with tetragon.podAnnotations.enabled=false → No annotations appear in the logs.
- Upgraded to tetragon.podAnnotations.enabled=true → Annotations appear in the logs.
- Upgraded to tetragon.podAnnotations.enabled=false → Annotations are no longer present in new logs.
- Installed with tetragon.podAnnotations.enabled=true → Annotations appear in the logs.
- Upgraded to tetragon.podAnnotations.enabled=false → Annotations are removed from new logs.
- - Upgraded to tetragon.podAnnotations.enabled=true → Annotations appear in the logs.

kkourt: What will happen instead of:
>         flags.Bool(KeyEnablePodAnnotations, false, "Add pod annotations field to events.")
We have:
>               flags.Bool(KeyEnablePodAnnotations, *true*, "Add pod annotations field to events.")

And then use:
```
tetragon.podAnnotations.enabled=false
```

install/kubernetes/tetragon/values.yaml
cy83rc0llect0r: @kkourt  Following up on your comment:

Changed the code default to flags.Bool(KeyEnablePodAnnotations, true, "Add pod annotations field to events.").
Kept values.yaml default as tetragon.podAnnotations.enabled: false, which overrides the code default via Helm ( and also set to false again).
Testing:
helm install tetragon ./install/kubernetes/tetragon → No annotations (Helm sets false).
(after uninstall again:) helm install tetragon ./install/kubernetes/tetragon--set tetragon.podAnnotations.enabled=false → No annotations (Helm sets false).
helm upgrade tetragon ./install/kubernetes/tetragon --set tetragon.podAnnotations.enabled=true → Annotations (e.g., test-key=test-value) in logs.
helm upgrade tetragon ./install/kubernetes/tetragon --set tetragon.podAnnotations.enabled=false → No annotations in new logs.
Does this resolve your concern?

install/kubernetes/tetragon/templates/tetragon_configmap.yaml
kkourt: I don't think it does.
I guess my point is that the helm chart should work correctly regardless of the default value of the config.

cy83rc0llect0r: Thank you for responding.
@kkourt  I’ve ensured the Helm chart is fully independent of the code’s default:

- changed `tetragon_configmap.yaml`:
  ```yaml
  {{- if .Values.tetragon.podAnnotations.enabled }}
  enable-pod-annotations: {{ .Values.tetragon.podAnnotations.enabled | quote }}
  {{- else }}
  enable-pod-annotations: "false"
  {{- end }}

kkourt: Thank you.
Why is the `{{- else -}}` part needed?

kkourt: Yap, I think:
> enable-pod-annotations: {{ .Values.tetragon.podAnnotations.enabled | default "false" | quote }}

Seems like the best choice to me.

kkourt (changes requested):
Can you please squash the changes from the review feedback into the relevant original commits? (`git rebase --interactive` using the `squash` and `fixup` actions should help).

install/kubernetes/tetragon/templates/tetragon_configmap.yaml
cy83rc0llect0r: @kkourt Thanks for the feedback. Glad we settled on the simpler approach! I’ve updated the chart to:
```yaml
enable-pod-annotations: {{ .Values.tetragon.podAnnotations.enabled | default "false" | quote }}

File: pkg/option/flags.go [hunk 3]
```diff
@@ -352,7 +355,7 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.Bool(KeyEnableProcessCred, false, "Enable process_cred events")
 	flags.Bool(KeyEnableProcessNs, false, "Enable namespace information in process_exec and process_kprobe events")
 	flags.Uint(KeyEventQueueSize, 10000, "Set the size of the internal event queue.")
-
+        flags.Bool(KeyEnablePodAnnotations, true, "Add pod annotations field to events.")
```
kkourt, line 358: Can you please set this to `false` by default?

kkourt (changes requested):

pkg/option/flags.go
cy83rc0llect0r: Done!
Thank you.

kevsecurity (approved):

kevsecurity (changes requested):
I think you need ```make format && make -C install/kubernetes``` or something similar.

kevsecurity:
And ```make generate-flags```.
Then commit all the changes from those make commands.

kevsecurity:
And maybe ```make vendor```?

cy83rc0llect0r:
> And maybe `make vendor`?

Excuse me. I know.
I'm resolving all the issues.
When completed I will reply this comment.

kevsecurity:
> Excuse me. I know. I'm resolving all the issues. When completed I will reply this comment.

Sorry. Just trying to be helpful.

cy83rc0llect0r:
> > Excuse me. I know. I'm resolving all the issues. When completed I will reply this comment.
> 
> Sorry. Just trying to be helpful.

You are! More than you think <3
And I think it's Done!
Thank you.

cy83rc0llect0r:
> I'm really sorry, but I think you need to run `make format` on your first commit, because otherwise it consists of a change, that is then formatted in the second commit, and it makes it hard to understand why the format changes are in the second commit.

I think it's Done!
Let me know if there was another problem.
Thank you.

api/v1/tetragon/tetragon.proto
```diff
@@ -54,6 +54,8 @@ message Pod {
   string workload = 6;
   // Kubernetes workload kind (e.g. "Deployment", "DaemonSet") of the Pod.
   string workload_kind = 7;
+  // Contains all the annotationss of the pod.
```
mtardy, line 57: small typo
```suggestion
  // Contains all the annotations of the pod.
```

mtardy (commented):
PR looks good from my PoV, thank you!

I was just wondering if it's necessary to add yet another feature flag and not just add this by default to the events @kkourt @kevsecurity wdyt?

api/v1/tetragon/tetragon.proto
cy83rc0llect0r: Thank you!
Resolved.

mtardy (approved):
Looks good to me, I would argue that maybe we should make this the default?

And maybe the best addition to this would be a test that we have at least one annotation, maybe along the existing label tests in e2e.

cy83rc0llect0r:
> Looks good to me, I would argue that maybe we should make this the default?
> 
> And maybe the best addition to this would be a test that we have at least one annotation, maybe along the existing label tests in e2e.

Thank you @mtardy 
I think making it the default could make the logs heavier, especially with long or numerous pod annotations. That’s why I considered it as an optional feature to enable when needed. But if you prefer it as the default, I’m fine with that too.

kkourt (changes requested):

install/kubernetes/tetragon/values.yaml
```diff
@@ -47,6 +47,7 @@ tetragon:
   # extraEnv:
   #   - name: foo
   #     value: bar
+  podAnnotations: {}
```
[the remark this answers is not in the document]
cy83rc0llect0r: @kkourt Nice point.
Done!

kkourt (approved):

install/kubernetes/tetragon/values.yaml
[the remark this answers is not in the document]
kkourt: Thanks!

api/v1/tetragon/tetragon.proto
```diff
@@ -54,6 +54,8 @@ message Pod {
   string workload = 6;
   // Kubernetes workload kind (e.g. "Deployment", "DaemonSet") of the Pod.
   string workload_kind = 7;
+  // Contains all the annotations of the pod.
```
mtardy, line 57: could we add that it's not enabled by default and it's guarded by a flag?

PR #3529, opened 2025-03-20
tetragon/windows: Port pkg/kernels to Windows

ExceptionalHandler:
### Description
Adding windows specific kernel version retrieval via ntdll!RtlGetVersion Win32 API. 

The `kernels_common.go` file is added to convert string version to numeric across all OS.

kernels.go continues to be a //go:build !windows listing

### Changelog

```
Add windows implementation to kernels package.
```

### DIFF
--- a/pkg/kernels/kernels_common.go
+++ b/pkg/kernels/kernels_common.go
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package kernels
+
+import (
+	"strconv"
+	"strings"
+)
+
+func KernelStringToNumeric(ver string) int64 {
+	// vendors like to define kernel 4.14.128-foo but
+	// everything after '-' is meaningless from BPF
+	// side so toss it out.
+	release := strings.Split(ver, "-")
+	verStr := release[0]
+	numeric := strings.TrimRight(verStr, "+")
+	vers := strings.Split(numeric, ".")
+
+	// Split out major, minor, and patch versions
+	majorS := vers[0]
+	minorS := ""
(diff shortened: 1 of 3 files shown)


pkg/kernels/kernels_common.go
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: Apache-2.0
```
olsajiri, line 1: whats the difference between kernels_common.go and kernels.go? I though the pattern is:

kernels.go -> common code
kernels_linux.go -> linux code
kernels_windows.go -> windows code

ExceptionalHandler: kernels.go -> Linux and Mac ( Non, windows, as indicated by //go:build directive added in this [commit](https://github.com/cilium/tetragon/commit/218b6fae9afb511d921751bb5f1d6dc82e2c2af9)
kernels_common.go -> All three OS 
kernels_linux.go -> Linux
kernels_windows.go -> Windows.

olsajiri (approved):

pkg/kernels/kernels_common.go
mtardy: Just curious, why do you need darwin builds, does Tetragon actually builds on darwin? Maybe having kernels.go and `_linux` and `_windows` would be simpler than those combination.

mtardy: I just learned that you can use the `//go:build unix` (but not the `_unix.go` suffix) if that can help

ExceptionalHandler: Yes `cmd/tetra` does build on Darwin. There is CI for it. it imports a bunch of these packages.

PR #3530, opened 2025-03-20
tetragon/windows: Compilation only change for pkg/metrics/syscallmetrics

ExceptionalHandler:
### Description
Adding stub function to compile metrics package on Windows

### Changelog

```
Compile the package pkg/metrics on Windows
```

### DIFF
--- a/pkg/metrics/syscallmetrics/raw_syscall_name_windows.go
+++ b/pkg/metrics/syscallmetrics/raw_syscall_name_windows.go
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package syscallmetrics
+
+import (
+	"github.com/cilium/tetragon/api/v1/tetragon"
+)
+
+func rawSyscallName(tp *tetragon.ProcessTracepoint) string {
+	return ""
+}


pkg/metrics/syscallmetrics/raw_syscall_name_windows.go
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package syscallmetrics
```
olsajiri, line 4: are there syscalls on windows ? no idea ;-)

if we need syscallmetrics package in windows then I think the `syscallmetrics.Handle` should have _windows/_linux versions, the generic version is now checking for `raw_syscalls/sys_enter` which will not happen on windows

ExceptionalHandler: This is a compilation only change. We cannot remove the `syscallmetrics` code from Windows without impacting compilation of common code. Goland does not have `#ifdef` based conditional compilation, so we need stub packages to compile common code. .

olsajiri: ok, understood,  I just think it'd make more sense to me to do that with syscallmetrics.Handle than with rawSyscallName

ExceptionalHandler: yes, I considered that but this approach minimizes code duplication, which is the approach we agreed upon in team meeting.  This is the smallest change to to let the package compile n Windows. 

We cannot rule out use of `syscallmetrics` in Windows in the future entirely, so stubbing out the whole package might not be the best approach. 

Also we already committed `raw_syscall_name_linux.go`, wouldn't want to walk back on that....

olsajiri (approved):

PR #3531, opened 2025-03-20
tetragon/windows: Compilation only change for pkg/metricconfig package on Windows

ExceptionalHandler:
### Description
Compilation-only change.
Add windows specific stub-function: `registerHealthMetricsEx` in an otherwise platform independent metricsconfig package This enables compilation of `pkg/metricsconfig` on Windows. 

### Changelog

```
Compile the package pkg/metricsconfig on Windows
```

### DIFF
--- a/pkg/metricsconfig/healthmetrics.go
+++ b/pkg/metricsconfig/healthmetrics.go
@@ -15,7 +15,6 @@ import (
 	"github.com/cilium/tetragon/pkg/metrics/enforcermetrics"
 	"github.com/cilium/tetragon/pkg/metrics/errormetrics"
 	"github.com/cilium/tetragon/pkg/metrics/eventmetrics"
-	"github.com/cilium/tetragon/pkg/metrics/kprobemetrics"
 	"github.com/cilium/tetragon/pkg/metrics/opcodemetrics"
 	"github.com/cilium/tetragon/pkg/metrics/overhead"
 	"github.com/cilium/tetragon/pkg/metrics/policyfiltermetrics"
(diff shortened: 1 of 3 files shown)


pkg/metricsconfig/healthmetrics.go
```diff
@@ -87,15 +86,10 @@ func registerHealthMetrics(group metrics.Group) {
...
 	cgroupratemetrics.RegisterMetrics(group)
-	// kprobe metrics
-	kprobemetrics.RegisterMetrics(group)
-	group.ExtendInitForDocs(kprobemetrics.InitMetricsForDocs)
 	// policy metrics
 	group.MustRegister(policymetrics.NewPolicyCollector())
 	// gRPC metrics
 	group.MustRegister(grpcmetrics.NewServerMetrics())
-	// missed metris
-	group.MustRegister(kprobemetrics.NewBPFCollector())
```
olsajiri, line 98: I think the test on bpf-next is failing because kprobe metrics are not initialized now on linux,
I think this needs to stay, and the Init call above as well

ExceptionalHandler: Sorry, I needed to call the splitter function.

pkg/metricsconfig/healthmetrics_extended_linux.go
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
```
olsajiri, line 1: not sure about the `extended` suffix, let's just have following?

```
healthmetrics.go
healthmetrics_linux.go
healthmetrics_windows.go
```

olsajiri (approved):

PR #3532, opened 2025-03-20
tetragon/windows: Port pidfile package on Windows

ExceptionalHandler:
### Description
pidfile package works well on windows, except readPidFile() function, which needs to access the proc filesystem. 

In Windows, we directly query the os to check if the pid is running.

This change results in compiled pidfile package in Windows and passed unit tests.

### Changelog
```
Compile and run unit tests on the package pkg/pidfile on Windows
```

### DIFF
--- a/pkg/pidfile/pid_alive_linux.go
+++ b/pkg/pidfile/pid_alive_linux.go
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package pidfile
+
+import (
+	"os"
+	"path/filepath"
+
+	"github.com/cilium/tetragon/pkg/option"
+)
+
+func isPidAlive(pid string) bool {
+	_, err := os.Stat(filepath.Join(option.Config.ProcFS, pid))
+	return err == nil
+}
--- a/pkg/pidfile/pid_alive_windows.go
+++ b/pkg/pidfile/pid_alive_windows.go
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package pidfile
+
+import (
+	"os"
+	"strconv"
+)
+
+func isPidAlive(pid string) bool {
+	int32pid, err := strconv.ParseInt(pid, 0, 32)
+	if err == nil {
+		_, err = os.FindProcess(int(int32pid))
+	}
+	return err == nil
+}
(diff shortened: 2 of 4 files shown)


pkg/pidfile/pidfile.go
```diff
@@ -33,7 +34,15 @@ func readPidFile() (uint64, error) {
 	}
 
 	pid := string(bytes.TrimSpace(data))
-	_, err = os.Stat(filepath.Join(option.Config.ProcFS, pid))
+	if runtime.GOOS != "windows" {
```
olsajiri, line 37: I'm not sure, but maybe we should stick with _linux/_windows suffixes,
even thought it's small change I wonder we should have just single way
to separate linux and windows code .. just a hunch this has a potential
to turn bad in future ... thoughts? @kkourt @mtardy

ExceptionalHandler: I have created a new function and unit test

pkg/pidfile/pid_alive_linux.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/cilium/tetragon/pkg/option"
+)
+
+func isPidAlive(pid string) bool {
+	_, err := os.Stat(filepath.Join(option.Config.ProcFS, pid))
+	if err == nil {
+		return true
+	}
+	return false
```
mtardy, line 18:
```suggestion
	return err == nil
```

pkg/pidfile/pid_alive_windows.go
```diff
@@ -0,0 +1,20 @@
...
+
+func isPidAlive(pid string) bool {
+	int32pid, err := strconv.ParseInt(pid, 0, 32)
+	if err == nil {
+		_, err = os.FindProcess(int(int32pid))
+	}
+	if err != nil {
+		return false
+	}
+	return true
```
mtardy, line 19:
```suggestion
	return err == nil
```

mtardy (commented):
lint is just suggesting this simplification, it's not checking _windows file btw, maybe we should fix that.

pkg/pidfile/pid_alive_linux.go
ExceptionalHandler: I have made this change

pkg/pidfile/pid_alive_windows.go
ExceptionalHandler: I have made this change

PR #3533, opened 2025-03-20
tetragon/windows: Compilation only Change for pkg/procsyms on Windows

ExceptionalHandler:
### Description
Stubbing out the GetFnSymbol function on Windows to make sure pkg/procsysms compiles on Windows.

### Changelog

```
Compile the package pkg/procsyms on Windows
```

### DIFF
--- a/pkg/procsyms/procsyms.go
+++ b/pkg/procsyms/procsyms.go
@@ -1,0 +1,20 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package procsyms
+
+import (
+	"fmt"
+)
+
+// FnSym is a function location (function name, module path + offset)
+type FnSym struct {
+	Name   string
+	Module string
+	Offset uint64
+}
+
+// ToString returns a string representation of FnSym
+func (fsym *FnSym) ToString() string {
+	return fmt.Sprintf("%s (%s+0x%x)", fsym.Name, fsym.Module, fsym.Offset)
+}
--- a/pkg/procsyms/procsyms_linux.go
+++ b/pkg/procsyms/procsyms_linux.go
@@ -21,18 +21,6 @@ var (
 	setCache sync.Once
 )
 
-// FnSym is a function location (function name, module path + offset)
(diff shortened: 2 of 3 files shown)


pkg/procsyms/procsysms.go
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: Apache-2.0
```
olsajiri, line 1: file name typo s/procsysms/procsyms/

pkg/procsyms/procsysms.go
```diff
@@ -0,0 +1,31 @@
...
+package procsyms
+
+import (
+	"fmt"
+	"sync"
+
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+var (
```
olsajiri, line 13: this is not use in generic code, I think it should go to _linux.go object

ExceptionalHandler: This is used in the function signature of `GetFnSymbol(pid int, addr uint64) (*FnSym, error)` which is a package export and linked from various points.

pkg/procsyms/procsysms.go
ExceptionalHandler: Fixed

pkg/procsyms/procsysms.go
ExceptionalHandler: As per what we have agreed, I think we need to minimize the code added to platform specific files and maximize code added to platform neutral files. In that light, if a var is used only in Linux, but builds on Windows, shouldn't it stay in common code ?

olsajiri: hum, but both cache and setCache vars are in linux code only right? I don't why to have them in common code

ExceptionalHandler: They were originally in `procsyms.go` and I moved only Linux specific code - code that wouldn't compile on Windows, into `procsysms_linux.go`. 
I guess there are more than one right answers for this.

olsajiri (approved):

PR #3535, opened 2025-03-20
tetragon/windows: Compilation only change to build encoder package 

ExceptionalHandler:
### Description
Added windows specific stub of the function `rawSyscallEnter()` in `encoder/raw_syscall_enter.go` listing of encoder package to allow it to compile on Windows

### Changelog

```
Compile  the package pkg/encoder on Windows
```

### DIFF
--- a/pkg/encoder/raw_syscall_enter_windows.go
+++ b/pkg/encoder/raw_syscall_enter_windows.go
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package encoder
+
+import "github.com/cilium/tetragon/api/v1/tetragon"
+
+func rawSyscallEnter(tp *tetragon.ProcessTracepoint) string {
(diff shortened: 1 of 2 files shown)


olsajiri (approved):

pkg/encoder/raw_syscall_enter_windows.go
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package encoder
+
+import "github.com/cilium/tetragon/api/v1/tetragon"
+
+func rawSyscallEnter(tp *tetragon.ProcessTracepoint) string {
+	return "unknown win32"
+
```
olsajiri, line 10: why win32? unknown or 'not supported' would do?

ExceptionalHandler: That is generally what syscalls on Windows are called: win32 APIs

PR #3538, opened 2025-03-20
tetragon/windows: Add default definitions for Windows

ExceptionalHandler:
### Description
Moved over `defaults.go` to `defaults_linux.go` .
Added `defaults_windows.go` to add set of default paths and names for Windows. 
Some default definitions are stubbed out for compilation.

### Changelog

```
Port the package defaults to Windows
```

### DIFF
--- a/pkg/defaults/defaults_windows.go
+++ b/pkg/defaults/defaults_windows.go
@@ -1,15 +1,13 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of Cilium
 
-//go:build !windows
-
 package defaults
 
 import "time"
 
 const (
 	// DefaultMapRoot is the default path where BPFFS should be mounted
-	DefaultMapRoot = "/sys/fs/bpf"
+	DefaultMapRoot = `C:\Program Files\ebpf-for-windows`
 
 	// DefaultMapPrefix is the default path prefix where Tetragon maps should be pinned
 	DefaultMapPrefix = "tetragon"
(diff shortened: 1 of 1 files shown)


pkg/defaults/defaults_windows.go
```diff
@@ -0,0 +1,51 @@
...
+
+const (
+	// DefaultMapRoot is the default path where BPFFS should be mounted
+	DefaultMapRoot = `C:\Program Files\ebpf-for-windows`
+
+	// DefaultMapPrefix is the default path prefix where Tetragon maps should be pinned
+	DefaultMapPrefix = "tetragon"
+
+	// DefaultEventMap is the default name of the Event map
+	DefaultEventMap = "tcpmon"
```
jrfastab, line 16: perhaps map names can be shared?

pkg/defaults/defaults_windows.go
```diff
@@ -0,0 +1,51 @@
...
+
+	// Default secure export logs permissions
+	DefaultLogsPermission = "600"
+
+	// Pid file where to write tetragon main PID
+	DefaultPidFile = DefaultRunDir + "tetragon.pid"
+
+	// defaults for the event cache
+	DefaultEventCacheNumRetries = 15
+	DefaultEventCacheRetryDelay = 2
```
jrfastab, line 47: same here I'm not sure why these would actually be different?

jrfastab (approved):

jrfastab:
I'll merge this, I do wonder a bit why Windows would end up with different defaults for some of those things in the comments above. It makes sense unify them later.

PR #3547, opened 2025-03-21
tetragon/widows: Add signal translation for Windows

ExceptionalHandler:
### Description
Windows signal handler requires specific map to identify signal codes and convert into code and string. This PR adds that map and implements the signal code to string conversion


jrfastab (approved):
OK but surely there is a standard pkg for this we could use instead?

jrfastab:
Signal translation is included from unix pkg which doesn't make sense to use in _windows pkgs.

PR #3558, opened 2025-03-26
tetragon: add raw tracepoints

olsajiri:
adding support for raw tracepoints, which allows to use raw tracepoint arguments, like:
```
kind: TracingPolicy
metadata:
  name: "rawtp"
spec:
  tracepoints:
    - subsystem: "sched"
      event: "sched_process_exec"
      raw: true
      args:
        - index: 2
          type: "linux_binprm"
```

also there's support for argument resolve keyword

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -156,6 +156,107 @@ spec:
     - index: 4
       type: "int64"
 ```
+## Raw Tracepoints
+
+Raw tracepoints allow same attachment as tracepoints, but allow access to raw
+tracepoint arguments (compared to standard tracepoints that provide access to
+arguments through tracepoint format).
+
+Raw tracepoints are attach directly without the perf layer in the middle,
+so should be bit faster than standard tracepoints.
+
+Each tracepoint is defined in kernel with similar statement like following:
+```
+TRACE_EVENT(sched_process_exec,
+
+        TP_PROTO(struct task_struct *p, pid_t old_pid,
+                 struct linux_binprm *bprm),
+        ...
+);
+```
+
+which defines `sched:sched_process_exec` tracepoint with arguments.
+
+Raw tracepoints are configured by setting with `raw` spec tag.
+```yaml
+apiversion: cilium.io/v1alpha1
+kind: tracingpolicy
+metadata:
+  name: "rawtp"
+spec:
+  tracepoints:
+    - subsystem: "sched"
+      event: "sched_process_exec"
+      raw: true
+```
+
+We can use raw tracepoint to attach the tracepoint and access above arguments
+directly.
+
+Following example stores 3rd argument:
+```yaml
+apiversion: cilium.io/v1alpha1
+kind: tracingpolicy
+metadata:
+  name: "rawtp"
+spec:
+  tracepoints:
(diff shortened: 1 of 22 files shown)


pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -80,6 +80,9 @@ type TracingPolicySpec struct {
 	// A list of tracepoint specs.
 	Tracepoints []TracepointSpec `json:"tracepoints,omitempty"`
 	// +kubebuilder:validation:Optional
+	// A list of raw tracepoint specs.
+	RawTracepoints []TracepointSpec `json:"rawtracepoints,omitempty"`
```
kkourt, line 84: How about `raw_tracepoints`?

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -25,6 +25,9 @@ func (h policyHandler) PolicyHandler(
 	if len(spec.Tracepoints) > 0 {
 		sections++
 	}
+	if len(spec.RawTracepoints) > 0 {
+		sections++
+	}
```
kkourt, line 30: The original reason for checking the sections was that you couldn't have kprobes+tracepoints in the same policy due to some map conflicts. With all the map and sensor reorg work, where every sensor has its own directory, I'm not sure if this is needed any more. 

For this PR specifically, I wonder: Is this needed for tracepoints?

Could we have something like:

```
if len(spec.Tracepoints) > 0 || len(spec.RawTracepoints) > 0 {
   sections++
}
```

And then a single create function (`createGenericTracepointSensor`) that handles both?


Another approach would be not to add a new section but have a per-hook flag to point out whether this is a raw tracepoint or not:

```
traceponts:
  - event: "foo"
     subsystem: "lala"
     raw: true
```

But I don't know if that's better.

kkourt (commented):
Looks great, thanks!

Please find some comments/questions below.

pkg/sensors/tracing/policyhandler_linux.go
olsajiri: > The original reason for checking the sections was that you couldn't have kprobes+tracepoints in the same policy due to some map conflicts. With all the map and sensor reorg work, where every sensor has its own directory, I'm not sure if this is needed any more.

yea, let's check on that and perhaps remove the check

> Another approach would be not to add a new section but have a per-hook flag to point out whether this is a raw tracepoint or not:
...

I considered that before, not sure why I chose the other way ;-) but this might mean less changes in the end, I'll check on that, thanks

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -156,6 +156,104 @@ spec:
     - index: 4
       type: "int64"
 ```
+## Raw Tracepoints
+
+Raw tracepoints allow same attachment as tracepoints, but allow access to raw
+tracepoint arguments (compared to standard tracepoints that provide access to
+arguments through tracepoint format).
```
mtardy, line 163: nit: maybe we can explain why raw_tracepoint are interested compared to tracepoint (performance)?

mtardy (approved):
Thanks that's a cool addition, nice that it works with the resolver.

PR #3561, opened 2025-03-26
tetragon/windows: port package sensors/exec/procevents into Windows

ExceptionalHandler:
### Description
Port tetragon functions that use the proc-fs to query running processes or process properties into Windows to query the same using WIN32 APIs. The function to populate the exec map is stubbed out but not removed and will be implemented once the ebpf program on windows supports that.

### Changelog

```
Port the package sensor/exec into Windows. This allows Tetragon on Windows to load Windows based sensors.
```

### DIFF
--- a/pkg/sensors/exec/procevents/proc.go
+++ b/pkg/sensors/exec/procevents/proc.go
@@ -6,13 +6,9 @@ package procevents
 import (
 	"bytes"
 	"encoding/hex"
-	"fmt"
-	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/cilium/tetragon/pkg/api/processapi"
-	"github.com/cilium/tetragon/pkg/option"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


olsajiri (approved):
lgtm, thanks

will-isovalent:
Looks like this PR erroneously removed the logic to handle `inInitTree` in procfs events on Linux. I'm guessing maybe an old version was copy/pasted over somehow? @ExceptionalHandler it might be worth double checking to see if anything else was accidentally changed.

will-isovalent:
Here's a PR to fix the above: https://github.com/cilium/tetragon/pull/3827

ExceptionalHandler:
> Here's a PR to fix the above: #3827

Thanks for pointing this out and fixing. The move to `_linux.go` makes the diff from old version cumbersome. I'll generally check the diff manually before PR but somehow maybe missed this one 😞

PR #3562, opened 2025-03-27
tetragon/windows: Use a package-level 'not supported' error variable

ExceptionalHandler:
### Description
Certain stub functions on Windows that are added for compilation support return an `errors.New("not supported on windows")` error which can be a package-level variable. This PR attempts to declare and use such a variable for packages that use that error more than once.

### DIFF
--- a/pkg/cgroups/cgroups_windows.go
+++ b/pkg/cgroups/cgroups_windows.go
@@ -4,6 +4,8 @@ package cgroups
 
 import (
 	"errors"
+
+	"github.com/cilium/tetragon/pkg/constants"
 )
 
 func CgroupFsMagicStr(magic uint64) string {
@@ -11,7 +13,7 @@ func CgroupFsMagicStr(magic uint64) string {
 }
 
 func GetCgroupIdFromPath(cgroupPath string) (uint64, error) {
-	return 0, errors.New("not Supported on windows")
+	return 0, constants.ErrWindowsNotSupported
 }
 
 func DiscoverSubSysIds() error {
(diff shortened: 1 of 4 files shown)


pkg/cgroups/cgroups_windows.go
```diff
@@ -6,12 +6,16 @@ import (
 	"errors"
 )
 
+var (
+	notSupportedWinErr = errors.New("not supported on windows")
```
mtardy, line 10: Hey, little nit, I'd like to enable the linter for that on Windows but that name should be something like:
```suggestion
	ErrWindowsNotSupported = errors.New("not supported on Windows")
```
or
```suggestion
	ErrNotSupportedOnWindows = errors.New("not supported on Windows")
```
error name should start with "Err" https://go.dev/wiki/Errors#naming

ExceptionalHandler: Do you suggest we change the name in existing code too ?

mtardy: Apply the naming convention on other already existing error in the code? yeah we can but we can do it in another PR :)

mtardy: if you reuse this elsewhere, maybe it's best to export it and put this in an appropriate package

ExceptionalHandler: I have added an error in constnats package and used it globally. Thanks for the suggestion and guidance.

mtardy (approved):
Thanks it's better like this :)!

PR #3575, opened 2025-04-01
tetragon/windows: Port process cache package on Windows

ExceptionalHandler:
### Description
This PR compiles and ports process cache into Windows. The only difference on Windows is how the pid for cache is computed. Currently on Windows, we are not using ktime as an input into process identifier.
The PID re-use is not a very big problem on Windows and the current bpf program does not communicate the parent's ktime to usermode. 
So the cache pid  on windows is based on numeric PID issued by kernel.

### Changelog

```
Compile and port the pkg/process package into Windows
```

### DIFF
--- a/pkg/process/process.go
+++ b/pkg/process/process.go
@@ -4,7 +4,6 @@
 package process
 
 import (
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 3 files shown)


jrfastab (approved):
thats nice lots of good code reuse here.

PR #3578, opened 2025-04-02
tetragon/windows:  Support Windows create and exit process - sensor changes

ExceptionalHandler:
### Description
This PR lists sensor side changes to support CreateProcess and ExitProcess events. 
The program loader needs to load the entire collection of native Windows ebps program image using cilium/ebpf library. 
The specs for a native Windows bpf program is not available as it is not in ELF format 
This changes the order in which maps are loaded - collection is loaded first which loads maps and programs automatically.

### Changelog

```
Tetragon on Windows supports Create Process and Exit process events via bpf programs  specific to Windows.
```

### DIFF
--- a/pkg/sensors/program/loader.go
+++ b/pkg/sensors/program/loader.go
@@ -1,0 +1,122 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+package program
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/tetragon/pkg/sensors/unloader"
+)
+
+// AttachFunc is the type for the various attachment functions. The function is
+// given the program and it's up to it to close it.
+type AttachFunc func(*ebpf.Collection, *ebpf.CollectionSpec, *ebpf.Program, *ebpf.ProgramSpec) (unloader.Unloader, error)
+
+type OpenFunc func(*ebpf.CollectionSpec) error
+
+type LoadOpts struct {
+	Attach AttachFunc
+	Open   OpenFunc
+	Maps   []*Map
+}
+
+func linkPinPath(bpfDir string, load *Program, extra ...string) string {
+	pinPath := filepath.Join(bpfDir, load.PinPath, "link")
+	if len(extra) != 0 {
+		pinPath = pinPath + "_" + strings.Join(extra, "_")
+	}
+	return pinPath
+}
+
+func RawAttach(targetFD int) AttachFunc {
+	return RawAttachWithFlags(targetFD, 0)
+}
+
+func NoAttach() AttachFunc {
+	return func(_ *ebpf.Collection, _ *ebpf.CollectionSpec,
+		prog *ebpf.Program, _ *ebpf.ProgramSpec) (unloader.Unloader, error) {
+		return unloader.ChainUnloader{
+			unloader.ProgUnloader{
+				Prog: prog,
+			},
+		}, nil
+	}
+}
+
+// MissingConstantsError is returned by [rewriteConstants].
+type MissingConstantsError struct {
+	// The constants missing from .rodata.
+	Constants []string
+}
+
+func (m *MissingConstantsError) Error() string {
+	return fmt.Sprintf("some constants are missing from .rodata: %s", strings.Join(m.Constants, ", "))
+}
+
+// The loadProgram loads and attach bpf object @load. It is expected that user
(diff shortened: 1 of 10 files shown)


jrfastab:
OK fix up the linter. Are there any specfic changes on _linux and multi os loader side that need extra review or are we mostly moving code around? Just looking for a hint on how to review here. Can you add a bit more description to the PR title to let us know if (a) this is just code refactor to get _windows and shared or if we have logic changes on _linux side. Most important is not to introduce regressions here.

ExceptionalHandler:
Diff1
Diff2
The Load function is moved back to load.go, so a comparison from few versions ago might be easy to understand. I have attached a couple screenshots to help assess the difference. 

Code is moved around and logical changes are done to assist compilation with code re-use. 
Otherwise almost all most Linux code is same.

pkg/sensors/program/loader_windows.go
```diff
@@ -0,0 +1,281 @@
...
+	notSupportedWinErr     = errors.New("not supported on windows")
+	programTypeProcessGUID = makeGUID(0x22ea7b37, 0x1043, 0x4d0d, [8]byte{0xb6, 0x0d, 0xca, 0xfa, 0x1c, 0x7b, 0x63, 0x8e})
+	attachTypeProcessGUID  = makeGUID(0x66e20687, 0x9805, 0x4458, [8]byte{0xa0, 0xdb, 0x38, 0xe2, 0x20, 0xd3, 0x16, 0x85})
+)
+
+func makeGUID(data1 uint32, data2 uint16, data3 uint16, data4 [8]byte) windows.GUID {
+	return windows.GUID{Data1: data1, Data2: data2, Data3: data3, Data4: data4}
+}
+
+func WinAttachStub(_ *ebpf.Collection, _ *ebpf.CollectionSpec,
```
olsajiri, line 29: could be local, also maybe more decriptive name like windowsAttachNotSupp ?

pkg/sensors/program/loader.go
```diff
@@ -0,0 +1,205 @@
...
+		prog *ebpf.Program, _ *ebpf.ProgramSpec) (unloader.Unloader, error) {
+		return unloader.ChainUnloader{
+			unloader.ProgUnloader{
+				Prog: prog,
+			},
+		}, nil
+	}
+}
+
+func LoadTracepointProgram(bpfDir string, load *Program, maps []*Map, verbose int) error {
```
olsajiri, line 49: I think this should not be in generic code, let's have both linux and windows version
and windows version would look like:

```
func LoadTracepointProgram(bpfDir string, load *Program, maps []*Map, verbose int) error {
  return windowsAttachNotSupp
}
```

I think it's more straight than overloading attach functions.. also it's conveying the message what attachments are supported on each os

ditto for all the other linux specific function Load*Program

olsajiri (commented):
looks good, left some comment for the loader, thanks

pkg/sensors/program/loader_windows.go
ExceptionalHandler: Adding Not Supported to a function name seems little off. I think  the word 'stub' should communicate the meaning.

pkg/sensors/program/loader.go
ExceptionalHandler: I have made this change.

pkg/sensors/program/loader_windows.go
```diff
@@ -0,0 +1,322 @@
...
+	prog *ebpf.Program, spec *ebpf.ProgramSpec) (unloader.Unloader, error) {
+
+	return nil, notSupportedWinErr
+}
+
+func RawAttachWithFlags(targetFD int, flags uint32) AttachFunc {
+	return winAttachStub
+}
+
+func TracepointAttach(load *Program, bpfDir string) AttachFunc {
```
olsajiri, line 40: could we get away with removing this one and others, like suggested in here https://github.com/cilium/tetragon/commit/16b67c77bf4b7d265e930504377e2f29297deb53

I think we don't need those attach functions in windows code,
but I can't check and it's possible I'm missing something

it might be helpful to have target that would build windows go code on linux,
just to check if we broken something with our change

ExceptionalHandler: We need these as stub functions for compilation.

ExceptionalHandler: Actually I can remove some but not all stub functions. Let me change..

ExceptionalHandler: I have removed some more unused functions

olsajiri (approved):

PR #3581, opened 2025-04-03
helm: Add enable ancestors for process events values

t0x01:
This PR adds `tetragon.enableProcessAncestors`, `tetragon.enableProcessKprobeAncestors`, `tetragon.enableProcessTracepointAncestors`,  `tetragon.enableProcessUprobeAncestors` and `tetragon.enableProcessLsmAncestors` Helm values for `--enable-process-ancestors`, `--enable-process-kprobe-ancestors`, `--enable-process-tracepoint-ancestors`, `--enable-process-uprobe-ancestors` and `--enable-process-lsm-ancestors` flags added in [PR 2938](https://github.com/cilium/tetragon/pull/2938).

```release-note
helm: Add enable ancestors for process events values
```

### DIFF
--- a/docs/content/en/docs/reference/helm-chart.md
+++ b/docs/content/en/docs/reference/helm-chart.md
@@ -84,8 +84,13 @@ To use [the values available](#values), with `helm install` or `helm upgrade`, u
 | tetragon.enablePolicyFilter | bool | `true` | Enable policy filter. This is required for K8s namespace and pod-label filtering. |
 | tetragon.enablePolicyFilterCgroupMap | bool | `false` | Enable policy filter cgroup map. |
 | tetragon.enablePolicyFilterDebug | bool | `false` | Enable policy filter debug messages. |
+| tetragon.enableProcessAncestors | bool | `false` | Enable Ancestors visibility in exec and exit events. |
 | tetragon.enableProcessCred | bool | `false` | Enable Capabilities visibility in exec and kprobe events. |
+| tetragon.enableProcessKprobeAncestors | bool | `false` | Enable Ancestors visibility in kprobe events. |
+| tetragon.enableProcessLsmAncestors | bool | `false` | Enable Ancestors visibility in lsm events. |
 | tetragon.enableProcessNs | bool | `false` | Enable Namespaces visibility in exec and kprobe events. |
+| tetragon.enableProcessTracepointAncestors | bool | `false` | Enable Ancestors visibility in tracepoint events. |
+| tetragon.enableProcessUprobeAncestors | bool | `false` | Enable Ancestors visibility in uprobe events. |
 | tetragon.enabled | bool | `true` |  |
 | tetragon.eventCacheRetries | int | `15` | Configure the number of retries in tetragon's event cache. |
 | tetragon.eventCacheRetryDelay | int | `2` | Configure the delay (in seconds) between retires in tetragon's event cache. |
--- a/install/kubernetes/tetragon/README.md
+++ b/install/kubernetes/tetragon/README.md
@@ -66,8 +66,13 @@ Helm chart for Tetragon
 | tetragon.enablePolicyFilter | bool | `true` | Enable policy filter. This is required for K8s namespace and pod-label filtering. |
 | tetragon.enablePolicyFilterCgroupMap | bool | `false` | Enable policy filter cgroup map. |
 | tetragon.enablePolicyFilterDebug | bool | `false` | Enable policy filter debug messages. |
+| tetragon.enableProcessAncestors | bool | `false` | Enable Ancestors visibility in exec and exit events. |
 | tetragon.enableProcessCred | bool | `false` | Enable Capabilities visibility in exec and kprobe events. |
+| tetragon.enableProcessKprobeAncestors | bool | `false` | Enable Ancestors visibility in kprobe events. |
+| tetragon.enableProcessLsmAncestors | bool | `false` | Enable Ancestors visibility in lsm events. |
 | tetragon.enableProcessNs | bool | `false` | Enable Namespaces visibility in exec and kprobe events. |
+| tetragon.enableProcessTracepointAncestors | bool | `false` | Enable Ancestors visibility in tracepoint events. |
+| tetragon.enableProcessUprobeAncestors | bool | `false` | Enable Ancestors visibility in uprobe events. |
 | tetragon.enabled | bool | `true` |  |
 | tetragon.eventCacheRetries | int | `15` | Configure the number of retries in tetragon's event cache. |
 | tetragon.eventCacheRetryDelay | int | `2` | Configure the delay (in seconds) between retires in tetragon's event cache. |
--- a/install/kubernetes/tetragon/templates/tetragon_configmap.yaml
+++ b/install/kubernetes/tetragon/templates/tetragon_configmap.yaml
@@ -14,6 +14,11 @@ data:
   debug: {{ .Values.tetragon.debug | quote }}
   enable-process-cred: {{ .Values.tetragon.enableProcessCred | quote }}
   enable-process-ns: {{ .Values.tetragon.enableProcessNs | quote }}
+  enable-process-ancestors: {{ .Values.tetragon.enableProcessAncestors | quote }}
+  enable-process-kprobe-ancestors: {{ .Values.tetragon.enableProcessKprobeAncestors | quote }}
+  enable-process-tracepoint-ancestors: {{ .Values.tetragon.enableProcessTracepointAncestors | quote }}
+  enable-process-uprobe-ancestors: {{ .Values.tetragon.enableProcessUprobeAncestors | quote }}
+  enable-process-lsm-ancestors: {{ .Values.tetragon.enableProcessLsmAncestors | quote }}
   process-cache-size: {{ .Values.tetragon.processCacheSize | quote }}
 {{- if .Values.tetragon.exportFilename }}
   export-filename: {{ .Values.exportDirectory}}/{{ .Values.tetragon.exportFilename }}
--- a/install/kubernetes/tetragon/values.yaml
+++ b/install/kubernetes/tetragon/values.yaml
@@ -146,6 +146,16 @@ tetragon:
   enableProcessCred: false
   # -- Enable Namespaces visibility in exec and kprobe events.
   enableProcessNs: false
+  # -- Enable Ancestors visibility in exec and exit events.
+  enableProcessAncestors: false
+  # -- Enable Ancestors visibility in kprobe events.
+  enableProcessKprobeAncestors: false
+  # -- Enable Ancestors visibility in tracepoint events.
+  enableProcessTracepointAncestors: false
+  # -- Enable Ancestors visibility in uprobe events.
+  enableProcessUprobeAncestors: false
+  # -- Enable Ancestors visibility in lsm events.
+  enableProcessLsmAncestors: false
   # -- Set --btf option to explicitly specify an absolute path to a btf file. For advanced users only.
   btf: ""
   # -- Override the command. For advanced users only.


mtardy (commented):
Hey I didn't realized we added so many flags for one feature, would it be possible to group them under `--enable-ancestors=kprobe,tracepoints,lsm` etc, instead of boolean flags? This PR is fine but I would think that would be a better usability thing. You can use StringSlice in viper flags.

Would you think that would be better maybe? Since this is a recent feature, maybe it's fine to just change the flags. Not sure it's been released already or not, if yes we might need to deprecate them before removing next release.

t0x01:
Hey.

I think it should be possible to group them, yes. This feature was released with version 1.4.0 not so long ago. I'll see what i can do about it. Should i submit another PR or extend this one?

mtardy:
> Hey.
> 
> I think it should be possible to group them, yes. This feature was released with version 1.4.0 not so long ago. I'll see what i can do about it. Should i submit another PR or extend this one?

I think it would be reasonable to merge this, and create another PR with two things:
1. deprecate existing flag and add the list one that groups them.
2. update the helm charts flag so that it uses the new group flag implem with a better UX.

kkourt:
> but then translating these to a comma-separated flag is not as straightforward, so if we're changing the flag format then it seems better to keep the Helm value format same.

I don't think we need to translate it into a comma-seperated flag. Using multiple flags (`--enable-ancstors=kprobe --enable-ancestors=tracepoint`) should also work. If we go down this road, I think we would also need a `--disable-ancestors` flag (?).

mtardy:
> I don't think we need to translate it into a comma-seperated flag. Using multiple flags (`--enable-ancstors=kprobe --enable-ancestors=tracepoint`) should also work.

Yes iirc viper parses both `--flag=value1 --flag=value2` and `--flag=value1,value2` by using StringSlice.

> If we go down this road, I think we would also need a --disable-ancestors flag (?).

If the default is to be disabled, not sure we need a disable ancestor flags? Otherwise we need to resolve conflict etc, it sounds annoying and unnecessary.

t0x01:
Hey @mtardy , @lambdanis 

Thanks for the help. I added both new `--enable-ancestors` flag and `tetragon.processAncestors.enabled` helm value, as you suggested. I also added an upgrade note that indicates which flags are being deprecated. Let me know if these changes are ok.

pkg/grpc/tracing/tracing.go
```diff
@@ -327,8 +327,8 @@ func GetProcessKprobe(event *MsgGenericKprobeUnix) *tetragon.ProcessKprobe {
 
 	proc, parent, tetragonProcess, tetragonParent := getProcessParent(&event.Msg.ProcessKey, event.Msg.Common.Flags)
 
-	// Set the ancestors only if --enable-process-kprobe-ancestors flag is set.
-	if option.Config.EnableProcessKprobeAncestors && proc.NeededAncestors() {
+	// Set the ancestors only if --enable-ancestors flag includes 'kprobe'.
+	if option.Config.EnableAncestors["kprobe"] && proc.NeededAncestors() {
```
mtardy, line 331: I think internally, it's fine to keep the many config static variables `EnableProcessKprobeAncestors` and co, so that it avoids these dynamic string to be used with a map. Because otherwise we should use constant string and that for access I'm not sure it would be optimal

pkg/option/flags.go
```diff
@@ -159,12 +162,20 @@ func ReadAndSetFlags() error {
...
+	if viper.GetBool(KeyEnableProcessAncestors) {
+		Config.EnableAncestors["base"] = true
+		Config.EnableAncestors["kprobe"] = viper.GetBool(KeyEnableProcessKprobeAncestors)
+		Config.EnableAncestors["tracepoint"] = viper.GetBool(KeyEnableProcessTracepointAncestors)
+		Config.EnableAncestors["uprobe"] = viper.GetBool(KeyEnableProcessUprobeAncestors)
+		Config.EnableAncestors["lsm"] = viper.GetBool(KeyEnableProcessLsmAncestors)
+	}
+
+	// Override deprecated flags, if new --enable-ancestors flag is used
+	if enableAncestors := viper.GetString(KeyEnableAncestors); enableAncestors != "" {
```
mtardy, line 177: could you maybe use `viper.GetStringSlice(key string) : []string` here so that you don't need to split and everything but just validate the values?

pkg/option/flags.go
```diff
@@ -159,12 +162,20 @@ func ReadAndSetFlags() error {
...
-		Config.EnableProcessLsmAncestors = viper.GetBool(KeyEnableProcessLsmAncestors)
+	Config.EnableAncestors = DefaultEnableAncestors()
+	// NOTE: enable-process-ancestors flags are marked as deprecated and
+	// planned to be removed in version 1.6
+	if viper.GetBool(KeyEnableProcessAncestors) {
+		Config.EnableAncestors["base"] = true
+		Config.EnableAncestors["kprobe"] = viper.GetBool(KeyEnableProcessKprobeAncestors)
+		Config.EnableAncestors["tracepoint"] = viper.GetBool(KeyEnableProcessTracepointAncestors)
+		Config.EnableAncestors["uprobe"] = viper.GetBool(KeyEnableProcessUprobeAncestors)
+		Config.EnableAncestors["lsm"] = viper.GetBool(KeyEnableProcessLsmAncestors)
```
mtardy, line 173: so here you could still reuse the existing vars

pkg/option/ancestors_event_filter.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+var AncestorsEventTypeMap = map[tetragon.EventType]string{
+	tetragon.EventType_PROCESS_EXEC:       "base",
+	tetragon.EventType_PROCESS_EXIT:       "base",
+	tetragon.EventType_PROCESS_KPROBE:     "kprobe",
+	tetragon.EventType_PROCESS_TRACEPOINT: "tracepoint",
+	tetragon.EventType_PROCESS_UPROBE:     "uprobe",
+	tetragon.EventType_PROCESS_LSM:        "lsm",
+}
```
mtardy, line 21: you could reuse the `tetragon.EventType_PROCESS_EXEC.String()` directly, by removing useless suffix, lowercase, etc...

pkg/option/flags.go
```diff
@@ -357,12 +368,20 @@ func AddFlags(flags *pflag.FlagSet) {
...
-	flags.Bool(KeyEnableProcessUprobeAncestors, false, fmt.Sprintf("Include ancestors in process_uprobe events. Only used if '%s' is set to 'true'", KeyEnableProcessAncestors))
-	flags.Bool(KeyEnableProcessLsmAncestors, false, fmt.Sprintf("Include ancestors in process_lsm events. Only used if '%s' is set to 'true'", KeyEnableProcessAncestors))
+	flags.String(KeyEnableAncestors, "", "Comma-separated list of process event types to enable ancestors for. Supported event types are: base, kprobe, tracepoint, uprobe, lsm. Unknown event types will be ignored. Type 'base' enables ancestors for process_exec and process_exit events and is required by all other supported event types for correct reference counting. An empty string disables ancestors completely")
+
+	// NOTE: enable-process-ancestors flags are marked as deprecated and
+	// planned to be removed in version 1.6
+	flags.Bool(KeyEnableProcessAncestors, false, "Deprecated, please, use --enable-ancestors. Include ancestors in process_exec and process_exit events. Disabled by default. Required by other enable ancestors options for correct reference counting")
```
mtardy, line 375: Maybe add the keyword to add in the alternative?
```suggestion
	flags.Bool(KeyEnableProcessAncestors, false, "Deprecated, please, use --enable-ancestors=base. Include ancestors in process_exec and process_exit events. Disabled by default. Required by other enable ancestors options for correct reference counting")
```

pkg/option/flags.go
```diff
@@ -357,12 +368,20 @@ func AddFlags(flags *pflag.FlagSet) {
...
+	flags.Bool(KeyEnableProcessKprobeAncestors, false, fmt.Sprintf("Deprecated, please, use --enable-ancestors. Include ancestors in process_kprobe events. Only used if '%s' is set to 'true'", KeyEnableProcessAncestors))
+	flags.Bool(KeyEnableProcessTracepointAncestors, false, fmt.Sprintf("Deprecated, please, use --enable-ancestors. Include ancestors in process_tracepoint events. Only used if '%s' is set to 'true'", KeyEnableProcessAncestors))
+	flags.Bool(KeyEnableProcessUprobeAncestors, false, fmt.Sprintf("Deprecated, please, use --enable-ancestors. Include ancestors in process_uprobe events. Only used if '%s' is set to 'true'", KeyEnableProcessAncestors))
+	flags.Bool(KeyEnableProcessLsmAncestors, false, fmt.Sprintf("Deprecated, please, use --enable-ancestors. Include ancestors in process_lsm events. Only used if '%s' is set to 'true'", KeyEnableProcessAncestors))
+	flags.MarkDeprecated(KeyEnableProcessAncestors, "please use --enable-ancestors")
```
mtardy, line 380:
```suggestion
	flags.MarkDeprecated(KeyEnableProcessAncestors, "please use --enable-ancestors=base")
```

pkg/option/ancestors_event_filter.go
```diff
@@ -0,0 +1,74 @@
...
+
+func ParseEnableAncestors(eventTypesString string) []string {
+	eventTypes := []string{}
+	for _, t := range strings.Split(eventTypesString, ",") {
+		t = strings.TrimSpace(t)
+		eventTypes = append(eventTypes, t)
+	}
+	return eventTypes
+}
```
mtardy, line 46: I guess you can avoid writing all this by relying on viper.GetStringSlice

mtardy (commented):
Hey thanks a lot for the PR, it looks really clean!

However I think you could simplify by reusing the old variables internally (it's fine to have these variables, I think the important part is just avoid the UX of having all those flags). Also you can reuse the viper.GetStringSlice and just do validation on the values.

pkg/option/flags.go
t0x01: `viper.GetStringSlice(key string) : []string` only splits strings by spaces, if i'm not mistaken. I think i can use `pflag.StringSlice` and `viper.UnmarshalKey` instead to correctly parse strings with comma-separated values read from both the cmd flag and the config files. So it would be something like this:
```go
	var err error
	var enableAncestors []string
...
	if err = viper.UnmarshalKey(KeyEnableAncestors, &enableAncestors); err != nil {
		return fmt.Errorf("failed to parse enable-ancestors value: %w", err)
	}

	if slices.Contains(enableAncestors, "base") {
		Config.EnableProcessAncestors = true
		Config.EnableProcessKprobeAncestors = slices.Contains(enableAncestors, "kprobe")
		Config.EnableProcessTracepointAncestors = slices.Contains(enableAncestors, "tracepoint")
		Config.EnableProcessUprobeAncestors = slices.Contains(enableAncestors, "uprobe")
		Config.EnableProcessLsmAncestors = slices.Contains(enableAncestors, "lsm")
	}
...
	flags.StringSlice(KeyEnableAncestors, []string{}, "Comma-separated list of process event types to enable ancestors for. Supported event types are: base, kprobe, tracepoint, uprobe, lsm. Unknown event types will be ignored. Type 'base' enables ancestors for process_exec and process_exit events and is required by all other supported event types for correct reference counting. An empty string disables ancestors completely")
```
Would this be fine?

mtardy: > viper.GetStringSlice(key string) : []string only splits strings by spaces

I checked, it works fine when doing `--options A --options B` but also `--options A,B`. It doesn't use spaces.

t0x01: It works fine with cmd flags, yes. That's because the flag's value goes through `pflag` first. But, when the value is read from a config file, it doesn't go through `pflag`.

`viper.GetStringSlice` calls `cast.ToStringSlice`. [Here](https://github.com/spf13/viper/blob/v1.20.1/viper.go#L885).
`cast.ToStringSlice` then calls `cast.ToStringSliceE` [here](https://github.com/spf13/cast/blob/v1.7.1/cast.go#L161).

If an option's value is read from a config file, the value's type will be a `string`. And `cast.ToStringSliceE` will call `strings.Fields` [here](https://github.com/spf13/cast/blob/v1.7.1/caste.go#L1311).

I used this code to test it:
```go
package main

import (
        "fmt"
        "reflect"
        "slices"

        "github.com/spf13/pflag"
        "github.com/spf13/viper"
)

func main() {
        pflag.StringSlice("options", []string{}, "")
        viper.SetConfigName("config")
        viper.SetConfigType("yaml")
        viper.AddConfigPath(".")
        viper.ReadInConfig()

        pflag.Parse()
        viper.BindPFlags(pflag.CommandLine)

        options := viper.GetStringSlice("options")
        fmt.Printf("options type: %s\n", reflect.TypeOf(viper.Get("options")))
        fmt.Printf("options len = %d\n", len(options))
        fmt.Printf("options value = %s\n", options)
        fmt.Printf("options contains 'A': %t\n", slices.Contains(options, "A"))
}
```

Tests:
```bash
user@debian:~/test$ go run test.go --options A,B
options type: []string
options len = 2
options value = [A B]
options contains 'A': true
user@debian:~/test$ go run test.go --options A --options B
options type: []string
options len = 2
options value = [A B]
options contains 'A': true
user@debian:~/test$ echo 'options: A B' > config.yaml && go run test.go
options type: string
options len = 2
options value = [A B]
options contains 'A': true
user@debian:~/test$ echo 'options: A,B' > config.yaml && go run test.go
options type: string
options len = 1
options value = [A,B]
options contains 'A': false
user@debian:~/test$ echo 'options: [A,B]' > config.yaml && go run test.go
options type: []interface {}
options len = 2
options value = [A B]
options contains 'A': true
```

So, then the option's substrings must be separated with spaces or specified as yaml list elements in config files, if we use `viper.GetStringSlice` to parse it. With cmd flags, comma separated values should work fine. I'm just not sure what to choose here.

mtardy: I think it's fine as YAML list are never defined with comma separated values so it feels okay to ask for YAML list for this. Would that be an issue?

t0x01: No, i don't think it would be an issue. I'm just trying to clarify it preemptively to make sure i don't commit any unwanted changes. I'll try to update this PR in the next couple of days, thanks.

mtardy: yeah sure, I'm sure we can find a way to make this works fine with Helm as well! Cool thank you!

t0x01: So, about Helm. Since ConfigMap.data only allows UTF-8 strings, i can't provide a YAML list as a ConfigMap.data field value. I can concatenate list's elements into a string with a space delimiter, though.

In `values.yaml` i can make `tetragon.processAncestors.enabled` value be a YAML list, like so:
```yaml
tetragon:
  processAncestors:
    # -- List of process event types to enable ancestors for.
    # Supported event types are: base, kprobe, tracepoint, uprobe, lsm. Unknown event types will be ignored.
    # Type "base" is required by all other supported event types for correct reference counting.
    # Set it to [] to disable ancestors completely.
    enabled: []
```
And in the `tetragon_configmap.yaml` template, i can do this:
```yaml
data:
{{- if .Values.tetragon.processAncestors.enabled }}
  enable-ancestors: {{ join " " .Values.tetragon.processAncestors.enabled | quote }}
{{- end }}
```
So, the rendered tetragon configMap will look like this:
```bash
$ helm template -s templates/tetragon_configmap.yaml . --set "tetragon.processAncestors.enabled={base,kprobe,lsm}"
---
# Source: tetragon/templates/tetragon_configmap.yaml
apiVersion: v1
kind: ConfigMap
metadata:
  ...
data:
  ...
  enable-ancestors: "base kprobe lsm"
  ...
```

`viper.GetStringSlice` should parse that value correctly, but i guess it can be confusing if this is the only place where the value would not be a YAML list, but a string. Not sure if i can pass an actual YAML list through a configMap somehow. Would that ok?

t0x01: Yeah, i don't know. `viper.UnmarshalKey` seems to be more appropriate here if we want to parse both `pflag.StringSlice` slices from cmd flags and comma-separated strings from config files. And it looks like it's sort of a [recommended](https://github.com/spf13/viper/tree/v1.20.1?tab=readme-ov-file#decoding-custom-formats) approach in situations like this as well.

I can add a slightly changed [`mapstructure.StringToSliceHookFunc`](https://github.com/go-viper/mapstructure/blob/v2.2.1/decode_hooks.go#L138C6-L138C27), so it trims spaces as well, like `ParseMetricsLabelFilter`. Both `enable-ancestors` and `metrics-label-filter` will accept comma-separated strings of the same format. And `--enable-ancestors` cmd flag would additionally be usable as `--enable-ancestors=base enable-ancestors=kprobe`. `StringToSliceHookFunc` would also be reusable for new options, where a similar format is applicable.

What do you think?

mtardy: I'm not sure to get the issue here, if you have something like this:

```yaml
tetragon:
  processAncestors:
    # -- List of process event types to enable ancestors for.
    # Supported event types are: base, kprobe, tracepoint, uprobe, lsm. Unknown event types will be ignored.
    # Type "base" is required by all other supported event types for correct reference counting.
    # Set it to [] to disable ancestors completely.
    enabled: []
```

You should be able to generate a YAML list with something like

```yaml
data:
  enable-ancestors:
{{ toYaml .Values.tetragon.processAncestors.enabled | indent 4 }}
```

if you can just do the easiest solution that requires as low code as possible, that's good for me! :D

t0x01: But this won't work, no? `ConfigMap.data` is a `map[string]string` ([reference](https://kubernetes.io/docs/reference/kubernetes-api/config-and-storage-resources/config-map-v1/#ConfigMap)).

> * data (map[string]string)  
>   
>   Data contains the configuration data. Each key must consist of alphanumeric characters, '-', '_' or '.'. Values with non-UTF-8 byte sequences must use the BinaryData field. The keys stored in Data must not overlap with the keys in the BinaryData field, this is enforced during validation process.

`{{ toYaml .Values.tetragon.processAncestors.enabled | indent 4 }}` will generate a YAML list, yes, but then `helm install` will fail with an error like this:
```bash
* ConfigMap in version "v1" cannot be handled as a ConfigMap: json: cannot unmarshal array into Go struct field ConfigMap.data of type string
```

mtardy: ah yes I see the problem, perhaps `enable-ancestors: "base,kprobe,lsm"` with commas is the best. That ConfigMap limitation is pretty annoying indeed. Perhaps you can find inspiration in what https://github.com/cilium/cilium might do for a similar situation, or k8s, I guess we are not the first to discover this issue.

Anyway, do whatever is simplest, but I imagine using commas kind of recall the possible usage of commas when using the flag so it's the best option? If we can have `--flag=value1 --flag=value2`, `--flag=value1,value2` and also `flag: "value1,value2"` in the ConfigMap I imagine that's ideal?

mtardy:
(I'll be out for 15 days from today, if you want to move this forward in the meantime you might want to ping someone else)

pkg/option/flags.go
mtardy: up to you! :)

pkg/option/ancestors_event_filter.go
t0x01: I removed this map as i reverted to static boolean config variables. But i returned `AncestorsEventTypes` slice back, mainly for use inside `filters` to check if ancestors are enabled for all supported event types if no event types are specified in the filter. I think i can generate both this slice and the `AncestorsEnabled` function with protogen and move it to the helpers package. Should i do that? Or is there a better way of doing it?

pkg/option/flags.go
t0x01: Done.

pkg/option/ancestors_event_filter.go
t0x01: Replaced with `viper.UnmarshalKey`.

t0x01:
> (I'll be out for 15 days from today, if you want to move this forward in the meantime you might want to ping someone else)

I can wait. I really appreciate the help. Thank you.

examples/configuration/tetragon.yaml
```diff
@@ -16,11 +16,7 @@ debug: false
 disable-kprobe-multi: false
 enable-export-aggregation: false
 enable-k8s-api: false
-enable-process-ancestors: false
-enable-process-kprobe-ancestors: false
-enable-process-tracepoint-ancestors: false
-enable-process-uprobe-ancestors: false
-enable-process-lsm-ancestors: false
+enable-ancestors: "base,kprobe,tracepoint,uprobe,lsm"
```
mtardy, line 19: maybe here we should put an empty string to keep the default config?

pkg/option/ancestors_event_filter.go
mtardy, line 1: only nit is that maybe you don't need a new file for that and maybe avoid snake case filename even though arguably we are using this around...

mtardy (commented):
Thank you, looks good only a few comments. Sorry it was so complicated, I honestly thought it would be easier to do that!

mtardy:
also please ignore the windows test failing, you can rebase if you want and it should disappear.

pkg/option/ancestors_event_filter.go
mtardy: I think it's fine to not over engineer this and maybe put this into an existing file, not sure why you moved them in the first place. Was there an import cycle?

mtardy:
> > (I'll be out for 15 days from today, if you want to move this forward in the meantime you might want to ping someone else)
> 
> I can wait. I really appreciate the help. Thank you.

thanks btw for waiting :)

pkg/option/ancestors_event_filter.go
t0x01: There are 2 similar `enabledAncestors()` functions currently:
- in [`eventcache`](https://github.com/cilium/tetragon/blob/v1.4.0/pkg/eventcache/eventcache.go#L54) package
- in [`filters`](https://github.com/cilium/tetragon/blob/v1.4.0/pkg/filters/filters.go#L172) package

Now i made it so the same function can be used in both places. And `AncestorsEventTypes` slice is still used only in `filters`. I initially moved them to a separate file because there was more code, as the initial implementation was similar to metricslabels.

examples/configuration/tetragon.yaml
t0x01: Done.

t0x01:
Had to resolve conflicts in `go.mod` again. No other changes.

mtardy:
> Had to resolve conflicts in `go.mod` again. No other changes.

sorry we removed deps last week, you'll need to rebase once again 😔

mtardy:
Thank you for your patience and help!

PR #3591, opened 2025-04-04
tetragon/Windows: Add support for process create and exit - ring-buffer 

ExceptionalHandler:
### Description
This PR adds changes specific to Windows to support ring buffer reads from user-mode using synchronous APIS and also add a global ebpf collection object. 

Reading ringbuffer on Windows is implemented  using Windows device io control API. This is a transitory change as libbpf based implementation to read ringbuffer using Linux-style synchronous API is currently in the works.

The ringbuffer maps on Windows currently have some issues with pinning. As a result we do not pin maps after program load.
In order to read from those maps we need to keep a global collection object alive and use the FD in the collection object to read from those maps.
This is a transitory change till pinning issues is fixed on Windows.
After which, this global object will be removed and we will use pinned maps to read ringbuf maps.

### Changelog

```
Tetragon on Windows supports Create Process and Exit process events via bpf programs  specific to Windows.
```

### DIFF
--- a/pkg/bpf/coll_windows.go
+++ b/pkg/bpf/coll_windows.go
@@ -1,0 +1,19 @@
+//go:build windows
+
+package bpf
+
+import (
+	"github.com/cilium/ebpf"
+)
+
+var (
+	execColl *ebpf.Collection
+)
+
+func SetExecCollection(coll *ebpf.Collection) {
+	execColl = coll
+}
+
+func GetExecCollection() *ebpf.Collection {
+	return execColl
+}
--- a/pkg/bpf/ringbuf_windows.go
+++ b/pkg/bpf/ringbuf_windows.go
@@ -1,0 +1,357 @@
+//go:build windows
+
+package bpf
+
+import (
+	"fmt"
+	"sync/atomic"
+	"syscall"
+	"unsafe"
+
+	"github.com/cilium/tetragon/pkg/logger"
+	"golang.org/x/sys/windows"
+)
+
+var (
+	ModuleNt       = windows.NewLazySystemDLL("ntdll.dll")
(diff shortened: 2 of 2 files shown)


olsajiri:
@ExceptionalHandler so how is this plugged to tetragon, so far it's just standalone windows library code, right?

ExceptionalHandler:
> @ExceptionalHandler so how is this plugged to tetragon, so far it's just standalone windows library code, right?

The ringbuffer is used here, already added in existing PR
https://github.com/cilium/tetragon/blob/e78676257cacc313638e9dae2acfd5dc0faeea38/pkg/observer/observer_windows.go#L224

olsajiri:
> > @ExceptionalHandler so how is this plugged to tetragon, so far it's just standalone windows library code, right?
> 
> The ringbuffer is used here, already added in existing PR
> 
> https://github.com/cilium/tetragon/blob/e78676257cacc313638e9dae2acfd5dc0faeea38/pkg/observer/observer_windows.go#L224

hum can't see the PR, please paste the link in here.. also let's maybe merge those PRs together? would be better to see how this code is used in the first place, thanks

ExceptionalHandler:
> 
> hum can't see the PR, please paste the link in here.. also let's maybe merge those PRs together? would be better to see how this code is used in the first place, thanks

This was the PR this went in :
https://github.com/cilium/tetragon/pull/3577

olsajiri:
> > hum can't see the PR, please paste the link in here.. also let's maybe merge those PRs together? would be better to see how this code is used in the first place, thanks
> 
> This was the PR this went in : #3577

heh nice, we merged user before the library it uses.. let's stick them together in future for sanity ;-) thanks

olsajiri (approved):
lgtm, thanks

PR #3599, opened 2025-04-07
tetragon: fix path permissions

olsajiri:
fixing path's permissions retrieval

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -376,8 +376,8 @@ FUNC_INLINE long store_path(char *args, char *buffer, const struct path *arg,
 	 * -----------------------------------------
 	 * Next we set up the flags.
 	 */
-	args[size] = (u32)flags;
-	args[size + sizeof(u32)] = (u16)i_mode;
+	*(u32 *)&args[size] = (u32)flags;
+	*(u16 *)&args[size + sizeof(u32)] = (u16)i_mode;
 	size += sizeof(u32) + sizeof(u16); // for the flags + i_mode
 
 	return size;
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -11,6 +11,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io/fs"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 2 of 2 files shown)


olsajiri:
> Thank you @olsajiri , would it be possible that you add stat(/usr/bin/su) if file **exists** just open/fdinstall to ensure that fileperm bits are fine too?

if it's about the suid bit, we could add it in one of the max path tests,
would something like below be ok for you?

```
@@ -7482,7 +7483,7 @@ spec:
 
                pathRemoved := pathFull[:len(pathFull)-1] + "b"
 
-               if err := testutils.CopyFile(pathRemoved, pathFull, 0755); err != nil {
+               if err := testutils.CopyFile(pathRemoved, pathFull, 0755|fs.ModeSetuid); err != nil {
                        t.Fatalf("testutils.CopyFileerror: %s", err)
                }
 
@@ -7509,7 +7510,7 @@ spec:
 
                st, err := os.Stat(pathRemoved)
                require.NoError(t, err)
-               pathPerm := tpath.FilePathModeToStr(uint16(st.Mode() | syscall.S_IFREG))
+               pathPerm := tpath.FilePathModeToStr(uint16(st.Mode() | syscall.S_IFREG | syscall.S_ISUID))
 
                t.Logf("path full:     %s\n", pathFull)
                t.Logf("path removed:  %s\n", pathRemoved)
```

tixxdz:
> if it's about the suid bit, we could add it in one of the max path tests,

Yes that's great! so let's maybe add fs.ModeSetgid too ;-)

BTW can't we just do?
```
pathPerm := tpath.FilePathModeToStr(uint16(st.Mode()))
```

Last thing let's ensure please we create a tmp unpredictable dir for tests and we remove the suid binary at exit  if not already

olsajiri:
> > if it's about the suid bit, we could add it in one of the max path tests,
> 
> Yes that's great! so let's maybe add fs.ModeSetgid too ;-)

added

> 
> BTW can't we just do?
> 
> ```
> pathPerm := tpath.FilePathModeToStr(uint16(st.Mode()))
> ```

go seems to use its own bits, while the mode type bits use linux contants 

> 
> Last thing let's ensure please we create a tmp unpredictable dir for tests and we remove the suid binary at exit if not already

right.. added ;-) thanks

PR #3607, opened 2025-04-08
Update golangci-lint to v2 and fix newly discovered issues in the code base

mtardy:
This simplifies the code quite well, it's pretty efficient!


kkourt (approved):

tpapagian (approved):

michi-covalent (approved):

will-isovalent (approved):

tixxdz (approved):
Nice fixes! LGTM for cgroups and caps change!

If you refer to the caps yes correct, it is more or less a same minus the bitwise not I guess but more Go friendly indeed.
There is some history from C here where caps were typedef array of uint32[2] in old kernels ... so we try to auto handle in bpf, but for Go looks good, thank you ;-)

mtardy:
> it is more or less a same minus the bitwise not I guess but more Go friendly indeed

ah okay I understand the reason now :)

PR #3611, opened 2025-04-10
tetragon/windows: Add windows compile as a ci step

ExceptionalHandler:
### Description
This PR adds a CI step to cross compile tetragon.exe for Windows on Linux.

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -135,3 +135,14 @@ jobs:
 
       - name: Build CLI release binaries
         run: make cli-release
+
+  build-windows:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Build tetragon Windows binaries
+        run: |
+          GOOS=windows GOARCH=amd64 make tetragon
+          GOOS=windows GOARCH=arm64 make tetragon


Makefile.windows
mtardy, line 1: hey, why don't you use the existing facilities with a new GOOS (like with GOARCH) or something like that? I'm not sure to understand why we need a whole new Makefile.

.github/workflows/static-checks.yaml
```diff
@@ -135,3 +135,12 @@ jobs:
...
         run: make cli-release
+
+  build-windows:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Build tetragon Windows binaries
+        run: make tetragon-windows
```
mtardy, line 146: just doing `GOOS=windows make tetragon` should build tetragon for windows

mtardy (changes requested):
I see that you got inspired by the CLI thing. I would favor the existing build target we have for tetragon. wdyt?

Makefile.windows
ExceptionalHandler: Of course that is the easy change. 
I am following precedent set by how tetra.exe is built for Windows and Linux. 
Also, a new make file lets us add packaging and ebpf building targets later without disrupting other makefiles.

.github/workflows/static-checks.yaml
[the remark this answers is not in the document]
ExceptionalHandler: I have made this change.

mtardy (approved):
Thank you, also:
> please also add ./tetragon.exe just after ./tetragon in the clean: target

I imagine you can do as a followup, no big deal

PR #3621, opened 2025-04-11
tetragon/windows: Port unit tests in grpc package into  Windows

ExceptionalHandler:
### Description
This PR changs the unit tests in grpc package to also run and succeed on Windows. Some exec based unit tests are excluded due to ktime not being present in Windows exec event.

### DIFF
--- a/pkg/grpc/process_manager_test_windows.go
+++ b/pkg/grpc/process_manager_test_windows.go
@@ -1,0 +1,29 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Cilium
+
+package grpc
+
+import (
+	"encoding/base64"
+	"os"
+	"testing"
+
+	"github.com/cilium/tetragon/pkg/defaults"
+	"github.com/cilium/tetragon/pkg/process"
+	"github.com/cilium/tetragon/pkg/reader/node"
+	"github.com/cilium/tetragon/pkg/watcher"
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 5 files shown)


pkg/grpc/process_manager_test_windows.go
```diff
@@ -0,0 +1,29 @@
...
+	assert.NoError(t, os.Setenv("NODE_NAME", "my-node"))
+	node.SetExportNodeName()
+
+	err := process.InitCache(watcher.NewFakeK8sWatcher([]interface{}{}), 10, defaults.DefaultProcessCacheGCInterval)
+	assert.NoError(t, err)
+	defer process.FreeCache()
+	id := process.GetProcessID(1, 2)
+	decoded, err := base64.StdEncoding.DecodeString(id)
+	assert.NoError(t, err)
+	assert.Equal(t, "1:1:1", string(decoded))
```
olsajiri, line 28: missing `assert.NoError(t, os.Unsetenv("NODE_NAME"))` ?

olsajiri (approved):
lgtm, left one comment, thanks

pkg/grpc/process_manager_test_windows.go
ExceptionalHandler: Just following the UT in _linux

olsajiri: @ExceptionalHandler hm, but _linux version has it, right?

ExceptionalHandler: Yes it has. I missed that. I will add in the next PR.

PR #3625, opened 2025-04-13
tetragon: backport v1.3 stat map metrics

olsajiri:
backporting new stat map metrics


mtardy:
you can safely ignore the veristat failures

PR #3630, opened 2025-04-14
pkg: fix cilium/ebpf pin.WalkDir uses

mtardy:
Test https://github.com/cilium/ebpf/pull/1736


mtardy:
patch merged by https://github.com/cilium/tetragon/pull/3849

PR #3638, opened 2025-04-15
go.mod: Consistently add replace directives

michi-covalent:
The top level go.mod uses the v0.0.0-00010101000000-000000000000 pseudo version for github.com/cilium/tetragon/{api,pkg/k8s} packages, and replaces them with local directories. IDEs like GoLand runs "go list -m" on all the go.mod files. It fails in api/ and contrib/tetragon-rthooks/ with an error like this:

    % go list -m -mod=mod all
    go: github.com/cilium/tetragon/pkg/k8s@v0.0.0-00010101000000-000000000000: invalid version: unknown revision 000000000000

because they don't replace indirect dependencies. Replace these indirect dependencies to get rid of the IDE warning.

### DIFF
--- a/api/go.mod
+++ b/api/go.mod
@@ -23,4 +23,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-replace github.com/cilium/tetragon => ../
+replace (
+	github.com/cilium/tetragon => ../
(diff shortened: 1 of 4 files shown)


michi-covalent:
the warning looks like this in goland:

Screenshot 2025-04-14 at 6 48 13 PM

jrfastab (approved):

PR #3651, opened 2025-04-17
Get tracing policy informers from controller-runtime manager

michi-covalent:
Let controller-runtime manage tracing policy informers to simplify the code a bit. Eventually we'll convert these informers to controllers.


jrfastab (approved):
Looks harmless to me.

PR #3661, opened 2025-04-22
PodInfo: Add .process.pod.container.privileged field

tpapagian:
To be able to check if a container is priviledged.


tixxdz (approved):
LGTM!

Not sure if we should maybe match by Container ID ?

Also maybe name it pod.container.security_context.privileged ? to be a bit consistent with k8s specs and have a namespace prefix for security related metadata for later? just a suggestion... up to you

tpapagian:
> LGTM!

Thanks!

> Not sure if we should maybe match by Container ID ?

I have also the same thoughts at the beginning, but it seems that the container spec does not contain a container ID.

> Also maybe name it pod.container.security_context.privileged ?

Yes, this makes sense. I will update that before merging.

PR #3662, opened 2025-04-22
pkg/k8s: move TopLevelWorkload metadata in PodInfo

mtardy:
This is a follow-up on 067b84a9b0ef ("podinfo: Populate workload info"). Like this command I didn't want to use metav1.ObjectMeta but just the desired set of information, here Name and Namespace.

This basically moves those metadata under status as they should be and also implement GetName and GetNamespace on the workload so that it can be use in a generic way, being a subset of metav1.Object.


mtardy:
do not merge this

PR #3663, opened 2025-04-23
examples: use explicit wording about guarantees

kkourt:
Example policies are meant to illustrate the Tetragon features, and not to be used unmodified in for security. Clarify this in the README and in some of the policies that have TOCTU issues.


mtardy (approved):
Yep

PR #3675, opened 2025-04-29
helm: Change default metrics scrape interval to 60s

lambdanis:
The default metrics scrape interval was set to 10s in both agent and operator ServiceMonitors. It was reported to be quite short for large environments, so let's change the default to 60s. It remains configurable with Helm values, so users who require more frequent updates can override the value.


kevsecurity (approved):

michi-covalent (approved):
😇

PR #3682, opened 2025-04-30
buf: enable protoc-gen-go-tetragon

olsajiri:
When we transitioned in using buf in eea39fd71e5b, we did not add the plugin for the custom tetragon code generator for eventchecker and other helpers. This commit adds it.

Fixes: eea39fd71e5b ("api: build using buf")

### DIFF
--- a/api/buf.gen.yaml
+++ b/api/buf.gen.yaml
@@ -4,7 +4,7 @@ plugins:
     out: ./v1
     opt:
       - paths=source_relative
-  - local: protoc-gen-go-grpc
+  - local: /src/bin/protoc-gen-go-tetragon
     out: ./v1
     opt:
       - paths=source_relative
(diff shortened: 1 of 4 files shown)


api/buf.gen.yaml
```diff
@@ -4,7 +4,7 @@ plugins:
     out: ./v1
     opt:
       - paths=source_relative
-  - local: protoc-gen-go-grpc
+  - local: /src/bin/protoc-gen-go-tetragon
```
kkourt, line 7: I think we need:
```diff
diff --git a/api/buf.gen.yaml b/api/buf.gen.yaml
index 5d6f74f6f..7ae098e67 100644
--- a/api/buf.gen.yaml
+++ b/api/buf.gen.yaml
@@ -4,6 +4,10 @@ plugins:
     out: ./v1
     opt:
       - paths=source_relative
+  - local: /src/bin/protoc-gen-go-tetragon
+    out: ./v1
+    opt:
+      - paths=source_relative
   - local: protoc-gen-go-grpc
     out: ./v1
     opt:
```

PR #3684, opened 2025-04-30
docs: update rthooks installation

kkourt:
- add instructions for kind
- use the helm repo rather than local helm files (since our released helm charts support this now)

### DIFF
--- a/docs/content/en/docs/installation/runtime-hooks.md
+++ b/docs/content/en/docs/installation/runtime-hooks.md
@@ -89,6 +89,82 @@ Install the new configuration file and restart containerd
 minikube cp /tmp/new-config.toml /etc/containerd/config.toml
 minikube ssh sudo systemctl restart containerd
 ```
+{{% /tab %}}
+
+{{% tab "kind (with Containerd)" %}}
+
+Note: Kind [only supports
+containerd](https://kind.sigs.k8s.io/docs/design/principles/#target-cri-functionality)
+currently.
+
+```shell
+cat <<EOF > kind-config.yaml
+apiVersion: kind.x-k8s.io/v1alpha4
+kind: Cluster
+nodes:
+  - role: control-plane
+    extraMounts:
+      - hostPath: /proc
+        containerPath: /procHost
+EOF
+kind create cluster --config kind-config.yaml
+EXTRA_HELM_FLAGS=(--set tetragon.hostProcPath=/procHost) # flags for helm install
+```
+
+Tetragon Runtime Hooks use [NRI](https://github.com/containerd/nri). NRI is [enabled by
+default](https://github.com/containerd/containerd/blob/main/docs/NRI.md#disabling-nri-support-in-containerd)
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/installation/runtime-hooks.md
```diff
@@ -89,27 +89,106 @@ Install the new configuration file and restart containerd
...
+RuntimeName:  containerd
+RuntimeVersion:  v1.7.18
+RuntimeApiVersion:  v1
+```
+
+Assuming that the containerd version is earlier than 2.0,
+you can use the `tetragon-oci-hook-setup` to patch the configuration file:
+
+```shell
+docker cp  kind-control-plane:/etc/containerd/config.toml /tmp/old-config.toml
```
mtardy, line 134: nit: weird space
```suggestion
docker cp kind-control-plane:/etc/containerd/config.toml /tmp/old-config.toml
```

docs/content/en/docs/installation/runtime-hooks.md
```diff
@@ -89,27 +89,106 @@ Install the new configuration file and restart containerd
...
+         +  plugin_path = "/opt/nri/plugins"
+         +  plugin_registration_timeout = "5s"
+         +  plugin_request_timeout = "2s"
+         +  socket_path = "/var/run/nri/nri.sock"
+```
+
+Install the new configuration file and restart containerd
+
+```shell
+docker cp  /tmp/new-config.toml kind-control-plane:/etc/containerd/config.toml
```
mtardy, line 161:
```suggestion
docker cp /tmp/new-config.toml kind-control-plane:/etc/containerd/config.toml
```

docs/content/en/docs/installation/runtime-hooks.md
```diff
@@ -89,27 +89,106 @@ Install the new configuration file and restart containerd
 minikube cp /tmp/new-config.toml /etc/containerd/config.toml
 minikube ssh sudo systemctl restart containerd
 ```
+{{% /tab %}}
+
+{{% tab "kind with Containerd (default)" %}}
```
mtardy, line 94: I think what you want to convey here is that Containerd running with kind is the default (for kind). It looks a little bit from the tabs that it's the default way of installing runtime hooks (even if that makes little sense). Maybe just drop the (default) and add a small note at the start of the guide that Containerd is the default when deploying kind.

image

docs/content/en/docs/installation/runtime-hooks.md
```diff
@@ -89,27 +89,106 @@ Install the new configuration file and restart containerd
...
+
+You can check the containerd version using:
+
+```shell
+docker exec -it kind-control-plane crictl version
+```
+
+Output should be similar to:
+
+```text
```
mtardy, line 123: you can drop `text` so that it looks like above in this example, it's more clear that it's an output.

image

mtardy (approved):
Thank you! :) Only a few nits but this page looks good!

PR #3685, opened 2025-04-30
sensors: cache spec when loading maps

kkourt:
For every map we load, we find its program, load its spec and then load the map. This introduces unnecessary overhead when loading the maps. This commit, caches the specs and only loads them if we have not seen the program before.

This speeds up loading a sensor, but also tests.

For example, Without the patch, running:
  go test -exec sudo  ./pkg/sensors/tracing -bpf-lib $(pwd)/bpf/objs -test.run TestKprobeSelectors -count 1

three times, results in:
  ok      github.com/cilium/tetragon/pkg/sensors/tracing  8.773s
  ok      github.com/cilium/tetragon/pkg/sensors/tracing  8.703s
  ok      github.com/cilium/tetragon/pkg/sensors/tracing  8.739s

With the patch, the same command results in:
  ok      github.com/cilium/tetragon/pkg/sensors/tracing  7.419s
  ok      github.com/cilium/tetragon/pkg/sensors/tracing  7.532s
  ok      github.com/cilium/tetragon/pkg/sensors/tracing  7.491s

Which is a ~14% improvement.


olsajiri (approved):
looks good, I wonder we could put loaderCache in Sensor object and factor the code so we could use it from within doLoadProgram .. but that might up with some struggle, so could be a follow up ;-) thanks

jrfastab:
Might be nice to check memory usage from caching here.

kkourt:
> looks good, I wonder we could put loaderCache in Sensor object and factor the code so we could use it from within doLoadProgram .. but that might up with some struggle, so could be a follow up ;-) thanks

Yeah, my initial patch did this, but I thought it was more substantial changes for not much performance benefit. We can revisit though.

kkourt:
> Might be nice to check memory usage from caching here.

The cache is only maintained when we preload the maps of a sensor. When we are done with them, it goes away so there shouldn't be a memory issue here.

PR #3686, opened 2025-04-30
fix(deps): Remove metallb dependency

joestringer:
Cilium stopped depending on this package back in upstream commit
https://github.com/cilium/cilium/commit/c2b935042f3e ("bgp: remove metallb bgp integration"). No need to
replicate any go.mod replace tricks here for it or its dependencies.


olsajiri (approved):
lgtm, thanks

PR #3699, opened 2025-05-05
doc: fix cgroup rate explanation not match the parameter

tico88612:
### Description

As title, fixed parameter explanation wrong.


olsajiri (approved):
thanks

kkourt (approved):

PR #3701, opened 2025-05-05
[v1.3] helm: update rthooks image to v0.5

kkourt:


tpapagian (approved):
It seems that in https://github.com/cilium/tetragon/pull/3700 some readme files are also changed. Not sure why this is not the case here. Maybe something changed in v1.3 or we miss a `make` inside `./install/kubernetes`.

kkourt:
> It seems that in #3700 some readme files are also changed. Not sure why this is not the case here. Maybe something changed in v1.3 or we miss a `make` inside `./install/kubernetes`.

Good catch, thanks!

kkourt:
No need to run the full CI, since it was executed in the previous version. Merging.

Issue #3712, opened 2025-05-07
NotDAddr filter error with `::ffff:127.0.0.1`


kobrineli:
### What happened?

Hi! I've found a bug while filtering addresses for kprobe on `tcp_connect` function.
Consider we have tracing policy:

```
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "tcp-connect"
spec:
  kprobes:
    - call: "tcp_connect"
      syscall: false
      args:
        - index: 0
          type: "sock"
      selectors:
        - matchArgs:
            - index: 0
              operator: "NotDAddr"
              values:
                - 127.0.0.1
                - ::1
                - ::ffff:127.0.0.1
```

which filters ipv4 loopback adress, ipv6 loopback address and ipv4 loopback address in ipv6 form.

Now let's listen for connections with `netcat`:

```
nc -l 1234
```

And connect to  opened local port also with `netcat`:

```
nc ::ffff:127.0.0.1 1234
```

Despite filters on destination address, tetragon still generates event:
```
{
  "process_kprobe": {
    "process": {
      "exec_id": "bWluaWt1YmU6MzQwODY2MDY3MzU4MzQ4OjMwNjUwNDA=",
      "pid": 3065040,
      "uid": 1001,
      "cwd": "/home/kobrineli",
      "binary": "/usr/bin/nc",
      "arguments": "::ffff:127.0.0.1 1234",
      "flags": "execve clone",
      "start_time": "2025-05-07T12:11:58.495582072Z",
      "auid": 1001,
      "parent_exec_id": "bWluaWt1YmU6MzM5NDY4NTI4MjU1NTEwOjMwNDc0NTM=",
      "refcnt": 1,
      "cap": {},
      "tid": 3065040,
      "in_init_tree": false
    },
    "function_name": "tcp_connect",
    "args": [
      {
        "sock_arg": {
          "family": "AF_INET6",
          "type": "SOCK_STREAM",
          "protocol": "IPPROTO_TCP",
          "saddr": "127.0.0.1",
          "daddr": "127.0.0.1",
          "sport": 43318,
          "dport": 1234,
          "cookie": "18446638556892972544",
          "state": "TCP_SYN_SENT"
        }
      }
    ],
    "action": "KPROBE_ACTION_POST",
    "policy_name": "tcp-connect",
    "return_action": "KPROBE_ACTION_POST"
  },
  "node_name": "minikube",
[...]

PR #3714, opened 2025-05-07
selectors: support ipv4-mapped ipv6 addresses

kobrineli:
Fixes: #3712

### Description

Currently Tetragon parses ip address as only ipv4 or ipv6, while all ipv4 address can be used as ipv6 in `::ffff:X.X.X.X` form.
In Go `IP.To4()` method works correctly if the address have such form, so Tetragon will add this address only in IPv4 map.
Later, when this addressed is used with AF_INET6, Tetragon BPF code will try to use only ipv6 map, leading to false positive event.

To fix this, we need to add ipv4 address in both ipv4 and ipv6 maps in case the original string contains semicolon (meaning IPv4-mapped form). 

### Changelog

```release-note
Support IPv4-mapped IPv6 address form in selectors.
```


kobrineli:
With this fix policy from the issue works correctly:

```
kind: TracingPolicy
metadata:
  name: "tcp-connect"
spec:
  kprobes:
    - call: "tcp_connect"
      syscall: false
      args:
        - index: 0
          type: "sock"
      selectors:
        - matchArgs:
            - index: 0
              operator: "NotDAddr"
              values:
                - 127.0.0.1
                - ::1
                - ::ffff:127.0.0.1
```

pkg/selectors/kernel.go
```diff
@@ -575,46 +588,66 @@ func getBase(v string) int {
 	return 10
 }
 
-func parseAddr(v string) ([]byte, uint32, error) {
```
kkourt, line 578: Thanks for the patch!

The original code  has already some duplication and your changes seem to introduce more.

First, it's simpler to call Split() first, instead of having to call net.ParseIP() twice.
Then, I think we can avoid having two different functions doing roughly the same thing and use a single function as before.

How about:

```go
func parseAddr(v string) ([]byte, uint32, error) {
	vParts := strings.Split(v, "/")
	if len(vParts) == 0 || len(vParts) > 2 {
		return nil, 0, errors.New("IP address is not valid: should be in format ADDR or ADDR/MASKLEN")
	}

	var maskLen uint32
	hasMaskLen := false
	switch len(vParts) {
	case 1:
	case 2:
		x, err := strconv.ParseUint(vParts[1], 10, 32)
		if err != nil {
			return nil, 0, errors.New("IP CIDR is not valid: mask part does not parse")
		}
		maskLen = uint32(x)
		hasMaskLen = true
	default:
		return nil, 0, errors.New("IP address is not valid: does not parse as IPv4 or IPv6")
	}

	ipaddr := net.ParseIP(vParts[0])
	if ipaddr == nil {
		return nil, 0, errors.New("IP address is not valid: does not parse as IPv4 or IPv6")
	}

	if !strings.Contains(v, ":") {
		ipaddr4 := ipaddr.To4()
		if ipaddr4 != nil {
			return nil, 0, errors.New("IP address is not valid: does not parse as IPv4 or IPv6")
		}
		if !hasMaskLen {
			maskLen = 32
		} else if maskLen >= 32 {
			return nil, 0, errors.New("IP CIDR is not valid: IPv4 mask len must be <= 32")
		}
		return ipaddr4, maskLen, nil
	}

	ipaddr6 := ipaddr.To16()
	if ipaddr6 == nil {
		return nil, 0, errors.New("IP address is not valid: does not parse as IPv4 or IPv6")
	}
	if !hasMaskLen {
		maskLen = 128
	} else if maskLen >= 128 {
		return nil, 0, errors.New("IP CIDR is not valid: IPv6 mask len must be <= 128")
	}
	return ipaddr6, maskLen, nil

}
```

Haven't really tested it, but, even if it does not, it should be possible to tweek it so that it works.

One of the benefits of having a single function is that we can unit test it. Could you please also add some unit tests to this function? It will help people understand what is expected of the function and make it harder to break it.

kkourt (changes requested):
Thanks, the change makes sense to me but I believe the implementation could be improved. 

CC: @kevsecurity

kkourt:
> @kkourt Hi! Thanks for the review. I've made adjustments but with some changes. Moreover, I've noticed that my original proposal had a flaw: in case of invalid ipv4 address, if the original string didn't contain colon, `err6` would remain nil and the error wouldn't be returned.
> 
> Added unit tests, covering every case of the function.

Thanks!

Can you please squash the changes from my feedback into the relevant original commits? (`git rebase --interactive` using the `squash` and `fixup` actions should help). if you want, feel free to keep the unit test in a separate patch but if you opt to do so please add a descriptive commit message.

Also, once done, please force-push the new branch and notify the reviewers by clicking on the github button "Re-request review".

Thanks!

kobrineli:
@kkourt 
Done

pkg/selectors/kernel.go
```diff
@@ -536,17 +537,16 @@ func writeMatchAddrsInMap(k *KernelSelectorState, values []string) error {
...
+
+		if len(addr) == net.IPv4len {
 			val := KernelLPMTrie4{prefixLen: maskLen, addr: binary.LittleEndian.Uint32(addr)}
 			m4[val] = struct{}{}
-		} else if len(addr) == 16 {
+		} else {
 			val := KernelLPMTrie6{prefixLen: maskLen}
 			copy(val.addr[:], addr)
 			m6[val] = struct{}{}
-		} else {
```
kkourt, line 548: How about we keep this else? I know it's not strictly needed, but I feel that it makes the code more readable and provides a sanity check that `parseAddr()` works as expected.

kkourt (commented):
Thanks, this looks great to me!

Please find a minor comment below.

pkg/selectors/kernel.go
kobrineli: Agree, returned this `else`

kobrineli:
@kkourt 
Hi! Could you restart CI, please?
Errors don't seem to be related to PR changes.

kkourt:
> @kkourt Hi! Could you restart CI, please? Errors don't seem to be related to PR changes.

The errors keep happening, even though I agree that they do not seem to be related. Would need to investigate fruther.
I created an issue #3732 for this. CC: @tpapagian

pkg/selectors/kernel.go
```diff
@@ -576,45 +578,55 @@ func getBase(v string) int {
...
-		}
-		ipaddr6 := ipaddr.To16()
-		if ipaddr6 != nil {
-			return ipaddr6, 128, nil
-		}
-		return nil, 0, errors.New("IP address is not valid: does not parse as IPv4 or IPv6")
-	}
+	var maskLen uint32
+
 	vParts := strings.Split(v, "/")
```
kevsecurity, line 583: Can we rip most of this out and use net.ParseCIDR() instead? https://pkg.go.dev/net#ParseCIDR

kobrineli: Hm, I thought that `net.ParseCIDR()` will not work correctly with addresses like `::ffff:127.0.0.1/16`, but I was wrong, it returns correct mask of total length of 128 bits.
I'll try to apply this.

kobrineli:
@kkourt @kevsecurity 
Changed `parseAddr` method to make it use `netip.ParseAddr` and `net.ParseCIDR` methods, now it looks much better.
Unit tests for IPv4-mapped IPv6 address pass successfully.

kevsecurity:
Thanks. I hit merge as @kkourt had previously approved, and I wasn't sure if you have merge permissions.

kobrineli:
I don't, thanks!

kkourt:
> Thanks. I hit merge as @kkourt had previously approved, and I wasn't sure if you have merge permissions.

There were some CI issues that I was hoping to resolve before merging the PR.
And it seems that the CI never finished after the new changes were pushed (or maybe I'm missing something?)

PR #3716, opened 2025-05-07
tetragon/windows: Fix observer to make it event independent

ExceptionalHandler:
### Description
This PR has three commits:
1. Change the bpf program to use  `LIBBPF_PIN_BY_NAME` flag while creating ringbuf map in Windows. This lets the ringbuf map be re-usable across programs. 

2. While porting observer to windows, pkg/observer became dependent on event type and was highly specific to exec event. 
This PR fixes that by adding event specific parsing in sensor 

3. Fix the CI to match changes in bpf program

### DIFF
--- a/.github/workflows/windows-build-smoke-test.yml
+++ b/.github/workflows/windows-build-smoke-test.yml
@@ -28,6 +28,11 @@ jobs:
         run: mkdir D:\temp
         shell: pwsh
 
+      - name: Checkout Tetragon Repo 
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          path: go/src/github.com/cilium/tetragon/
+      
       - name: Set MSVC Environment Variables
         shell: cmd
         run: |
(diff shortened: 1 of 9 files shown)


.github/workflows/windows-build-smoke-test.yml
```diff
@@ -72,7 +84,7 @@ jobs:
           LDFLAGS: ${{ env.LD_FLAGS }}
         run: |
             .\scripts\initialize_repo.ps1
-
+      
       - name: Build
```
olsajiri, line 88: white space mixup

bpf/windows/process_monitor.c
```diff
@@ -90,47 +106,52 @@ get_scratch_space()
...
+int ProcessMonitor(process_md_t *ctx)
 {
-	process_event process_info;
-
-	memset(&process_info, 0, sizeof(process_info));
+	if (ctx->operation == PROCESS_OPERATION_CREATE) {
+		struct process_create_info_t process_create_info;
+
+		memset(&process_create_info, 0, sizeof(process_create_info));
+		process_create_info.common.op = 5;
```
olsajiri, line 116: could we use MSG_OP_EXECVE constant

pkg/bpf/ringbuf_windows.go
```diff
@@ -125,6 +137,24 @@ func GetNewWindowsRingBufReader() *WindowsRingBufReader {
 	return &reader
 }
 
+type Record struct {
+	// The CPU this record was generated on.
```
olsajiri, line 141: so this is duplicate of cilium/ebp Record struct right? why?

pkg/observer/observer_windows.go
```diff
@@ -251,39 +50,26 @@ func (observer *Observer) RunEvents(stopCtx context.Context, ready func()) error
 		defer wg.Done()
```
olsajiri, line 50: idea for future cleanup.. it looks like RunEvents could be generic and we could have just windows/linux ring buffer specific code

olsajiri (approved):
left few comments/questions, but can't really test it ;-) lgtm

bpf/windows/process_monitor.c
ExceptionalHandler: I am currently not including any Linux-side headers. Today the build system is complicated. Once it is simplified in a way that we can build the programs in bpf/windows folder directly, I'll include the `msg_types.h`.

pkg/observer/observer_windows.go
ExceptionalHandler: yes. I figured we could do that.

PR #3717, opened 2025-05-08
generic_calls.h : Extract arg in map instead of variable

tdaudi:
### Description

Currently, the argument in extracted and stored inside a simple variable. But if we want to reorder the tail calls and still be able to access the argument, we need to store it inside the map.

```release-note
Extract arg in map instead of variable
```

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -114,7 +114,7 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 	struct msg_generic_kprobe *e;
 	struct event_config *config;
 	int am, zero = 0;
-	unsigned long a;
+	unsigned long *a;
 	long ty;
 
 	e = map_lookup_elem(&process_call_heap, &zero);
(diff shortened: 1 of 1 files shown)


olsajiri (commented):
hum, I don't fully understand the change, what's broken? is it preparation for some upcoming changes?

tdaudi:
Sure, the reason for this is the where we would like to use the `args:` data to be used in the filter tail call.
To be more specific, the use case for this is to be able to match one of the extract attribute to `matchCapability` for example. A policy would allow detecting one `args:` and test against one capability for being able to deny running pods/docker containers with the `securityContext.privileged = true` or  `--privileged` flag.

So this commit is just a beginning

olsajiri:
> Sure, the reason for this is if we would like to use the `args:` data to be used in the filter tail call. To be more specific, the use case for this is to be able to match one of the extract attribute to `matchCapability` for example. A policy would allow detecting one `args:` and test against one capability for being able to deny running pods/docker containers with the `securityContext.privileged = true` or `--privileged` flag.
> 
> So this commit is just a beginning

ok, so I think let's push this together with the rest of the feature then

tdaudi:
Closing because the commit of this PR is now part of https://github.com/cilium/tetragon/pull/3772

Issue #3710, opened 2025-05-07
kprobes policy with out of order arguments does not work as expected


kkourt:
Consider this policy:

```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "sys-lseek"
spec:
  kprobes:
  - call: "sys_lseek"
    syscall: true
    args:
    - index: 0
      type: "int"
      label: "index 0"
    - index: 1
      type: "int"
      label: "index 1"
    - index: 2
      type: "int"
      label: "index 2"
```

If we execute:
```
echo "0 1 2" | ./contrib/tester-progs/lseek-pipe
```

The args in the event we get are:
```json
[
  {
    "int_arg": 0,
    "label": "index 0"
  },
  {
    "int_arg": 1,
    "label": "index 1"
  },
  {
    "int_arg": 2,
    "label": "index 2"
  }
]
```

Consider the same policy but with the arguments reversed:
```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "sys-lseek"
spec:
  kprobes:
  - call: "sys_lseek"
    syscall: true
    args:
    - index: 2
      type: "int"
      label: "index 2"
    - index: 1
      type: "int"
      label: "index 1"
    - index: 0
      type: "int"
      label: "index 0"
```

We then get:
```json
[
  {
    "int_arg": 0,
    "label": "index 2"
  },
  {
    "int_arg": 1,
    "label": "index 1"
  },
  {
    "int_arg": 2,
    "label": "index 0"
  }
]
```

Which is not what we would expect.

Issue #3711, opened 2025-05-07
kprobes: support having the same argument twice


kkourt:
If we write a policy with the same argument twice,

```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "sys-lseek"
spec:
  kprobes:
  - call: "sys_lseek"
    syscall: true
    args:
    - index: 0
      type: "int"
      label: "index 0"
    - index: 0
      type: "int"
      label: "index 0 (again)"

```

The agent will accept it, and then produce a wrong event. For example, if we execute:
```
echo "100 100 100" | ./contrib/tester-progs/lseek-pipe
```

We will get:

```json
[
  {
    "int_arg": 100,
    "label": "index 0"
  },
  {
    "int_arg": 0,
    "label": "index 0 (again)"
  }
]
```

And a warning in the logs:
```
time="2025-05-07T13:46:20+02:00" level=warning msg="Int type error" arg.usertype= error=EOF
```

Note that accessing the same argument twice is useful when used with "resolve:". For example:
```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "lsm"
spec:
  kprobes:
  - call: "security_bprm_check"
    syscall: false
    args:
      - index: 0
        type: "string"
        resolve: "mm.owner.comm"
        label: "proc"
      - index: 0
        type: "string"
        resolve: "mm.owner.real_parent.comm"
        label: "parent"
      - index: 0
        type: "string"
        resolve: "mm.owner.real_parent.real_parent.comm"
        label: "grand-parent"
    selectors:
      - matchActions:
        - action: Post
```

I'm marking this as a bug because we provide the wrong data to the user and we should (at minimum) reject the policy. The best solution, however, is to support multiple arguments on the same index (which, arguably, is a new feature).

See also: https://github.com/cilium/tetragon/issues/3710

olsajiri:
yes, with the resolve it makes sense now to allow to retrieve multiple argument values for single argument, which is new feature

but current code should fail for such spec now, will fix, thanks

tdaudi:
Hello ! Super glad this is part of the roadmap.
I've been working on this feature few months ago, but it stayed as a WIP and was not ready for upstream. The final policy look like this. I acknowledge that "duplicateIndex" is not very user-friendly, but it allows an easier reuse in the `matchArgs`.

@olsajiri If you have not started development on this yet, I'll be happy to continue pushing it.

```yml
[...]

PR #3725, opened 2025-05-11
tetragon: Fix kprobe argument printers order

olsajiri:
Arguments are appended based on the index value,
so the argument printers need to follow that.

Fixes: https://github.com/cilium/tetragon/issues/3710

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -7717,3 +7717,70 @@ spec:
 	err = jsonchecker.JsonTestCheck(t, checker)
 	assert.NoError(t, err)
 }
+
+func TestKprobeArgsReverse(t *testing.T) {
+	var doneWG, readyWG sync.WaitGroup
+	defer doneWG.Wait()
+
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
+	pidStr := strconv.Itoa(int(observertesthelper.GetMyPid()))
+	t.Logf("tester pid=%s\n", pidStr)
+
+	lseekConfigHook_ := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "sys-write"
+spec:
+  kprobes:
+  - call: "sys_lseek"
+    return: false
+    syscall: true
+    args:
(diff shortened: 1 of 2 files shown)


tixxdz (approved):

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -7717,3 +7717,70 @@ spec:
...
+	}
+
+	kpChecker := ec.NewProcessKprobeChecker("lseek-checker").
+		WithFunctionName(sm.Suffix("sys_lseek")).
+		WithArgs(ec.NewKprobeArgumentListMatcher().
+			WithOperator(lc.Ordered).
+			WithValues(
+				ec.NewKprobeArgumentChecker().WithIntArg(0).WithLabel(sm.Full("index 0")),
+				ec.NewKprobeArgumentChecker().WithIntArg(1).WithLabel(sm.Full("index 1")),
+				ec.NewKprobeArgumentChecker().WithIntArg(2).WithLabel(sm.Full("index 2")),
```
kkourt, line 7772: In https://github.com/cilium/tetragon/pull/3730#discussion_r2086564108, you mentioned:
> with this change (still need to push final commit) the order should be given by the position in the spec

My interpretation of the position in the spec is that the arguments are = {"index 2", "index 1", "index 0"}

But this test (and, AFAICT the code)  indicates that the arguments are ordered based on the index value not the spec order. Having the arguments in "spec order" seems like a better choice for me. Also, I'm thinking that if we change the order, we might break users that depend on it.

kkourt: Let's continue the discussion in https://github.com/cilium/tetragon/issues/3711#issuecomment-2879269243

olsajiri: so the order is always set by ebpf program from index 0,1... up to this change if you spcify different order, you will receive wrong data, because the printers will not match the ebpf program order

so this change fixes the printers order to match the ebpf argument retrieval and storage

kkourt: yap, I think I understand that part now. Thanks.

Since this fixes an obvious bug, we can merge it and then continue the discussion on what we want to do in https://github.com/cilium/tetragon/issues/3711#issuecomment-2879269243.

PR #3730, opened 2025-05-12
tetragon: factor args processing

olsajiri:
we're now reading only configured function arguments (instead of all)

got ~9% speedup for following spec when monitoring kernel build:
```
kind: TracingPolicy
metadata:
  name: "sys-read"
spec:
  kprobes:
  - call: "sys_read"
    syscall: true
  - call: "sys_write"
    syscall: true
```

stats:
```
without tetragon:

15,719,804,843,431      cycles:u                                                                (83.33%)
 2,598,845,153,040      cycles:k                                                                (83.33%)
19,842,037,958,091      instructions:u                   #    1.26  insn per cycle              (83.33%)
 1,273,595,233,682      instructions:k                   #    0.49  insn per cycle              (83.33%)
18,321,799,029,925      cycles                                                                  (83.33%)
21,111,722,586,302      instructions                     #    1.15  insn per cycle              (83.33%)

     380.714728532 seconds time elapsed


current code:

16,823,105,829,998      cycles:u                                                                (83.33%)
 3,263,453,744,849      cycles:k                                                                (83.33%)
20,909,182,069,407      instructions:u                   #    1.24  insn per cycle              (83.33%)
 1,599,988,474,696      instructions:k                   #    0.49  insn per cycle              (83.33%)
20,088,537,741,909      cycles                                                                  (83.33%)
22,506,465,170,136      instructions                     #    1.12  insn per cycle              (83.33%)

     463.132465533 seconds time elapsed


with the change:

16,757,140,024,302      cycles:u                                                                (83.33%)
 3,213,714,566,101      cycles:k                                                                (83.33%)
20,973,450,014,056      instructions:u                   #    1.25  insn per cycle              (83.33%)
 1,574,493,868,806      instructions:k                   #    0.49  insn per cycle              (83.33%)
19,973,638,645,054      cycles                                                                  (83.33%)
22,547,534,385,897      instructions                     #    1.13  insn per cycle              (83.33%)

     420.833966653 seconds time elapsed

```


bpf/process/generic_calls.h
```diff
@@ -130,14 +722,15 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 		     : "i"(MAX_SELECTORS_MASK));
 	ty = (&config->arg0)[index];
 
-	a = (&e->a0)[index];
+	a = __get_arg_value(ctx, config, index);
```
tdaudi, line 725: Is it not possible to keep using `(&e->a0)[index]` ? This is something I needed for this future PR because I will need to call it in `TAIL_CALL_FILTER`. 
https://github.com/cilium/tetragon/pull/3717

I personally like the idea to have atomic tailcalls that can be reordered if possible, and this map make it easy to understand.

olsajiri: hum, I'm not sure I understand what you need that for, could you please describe it with more details?

the idea of this change is to postpone argument retrieval until it is needed.. so we no longer have e->a0/a1/a2... anymore, just one 'a' value at the time it's needed

also I have one more change on top of this that limits tailcalls for arguments that we don't need (which we do now and it's a waste)

pkg/sensors/tracing/generickprobe.go
```diff
@@ -785,12 +786,18 @@ func addKprobe(funcName string, instance int, f *v1alpha1.KProbeSpec, in *addKpr
...
 		argSigPrinters = append(argSigPrinters, argP)
 
 		pathArgWarning(a.Index, argType, f.Selectors)
 	}
 
+	// Arguments are appended based on the index value,
+	// so the argument printers need to follow that
+	sort.Slice(argSigPrinters, func(i, j int) bool {
+		return argSigPrinters[i].index < argSigPrinters[j].index
+	})
```
kkourt, line 799: I've wrote some of my thoughts in: https://github.com/cilium/tetragon/issues/3711#issuecomment-2875584910.
One question I have is: if a user defines:
```
args:
 - index: 1
    type: ...
 - index: 0
   type: ...
```

What would be the order of the arguments in the ProcessKprobe event?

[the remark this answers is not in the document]
olsajiri: so at the moment (current main) it's always sorted from index 0,
with this change (still need to push final commit) the order should be given by the position in the spec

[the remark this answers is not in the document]
kkourt: Ah, I see. That makes sense to me!

bpf/process/generic_calls.h
tdaudi: This is the change I want to bring 
```yaml
      matchCapabilities:
      - type: Effective
        operator: In
        index: 2
        values:
        - "CAP_SYS_ADMIN"
```

I want to be capable of using matchCapabilities to match the extracted data earlier. For this, I also need to reorder tailcalls and put `TAIL_CALL_FILTER` after `TAIL_CALL_SETUP`. 
In my version of the code, I store the extracted data in `(&e->a0)[index]` and call it in `TAIL_CALL_FILTER` to check if it matches the capabilities.
It works fine in my code currently, but with your changes, I don't know how I could do this.

kkourt: Wouldn't be easier to use the Mask operator to check against the capability value?

tdaudi: No, because the value compared in the `process_filter_capabilities` is not `21`(=CAP_SYS_ADMIN) but `1U << 21` 

https://github.com/cilium/tetragon/blob/792590e57f484ac51f549f1da14aecc31ef9c246/bpf/process/pfilter.h#L235-L238

So if we want to do something similar, we need to do the mask on `1U << 21` which is less conveniant

olsajiri: > 
> 
> This is the change I want to bring
> 
>       matchCapabilities:
>       - type: Effective
>         operator: In
>         index: 2
>         values:
>         - "CAP_SYS_ADMIN"
> 
> I want to be capable of using matchCapabilities to match the extracted data earlier. For this, I also need to reorder tailcalls and put TAIL_CALL_FILTER after TAIL_CALL_SETUP.
> In my version of the code, I store the extracted data in (&e->a0)[index] and call it in TAIL_CALL_FILTER to check if it matches the capabilities.
> It works fine in my code currently, but with your changes, I don't know how I could do this.

so IIUC you want to extract credentials data from probed function argument (instead of reading current task) and match it against values? seems like we could do that with matchArgs with some additional value type support as @kkourt suggested ?

it's possible we still miss something.. would be great to see your changes

tdaudi: > so IIUC you want to extract credentials data from probed function argument (instead of reading current task) and match it against values? seems like we could do that with matchArgs with some additional value type support as @kkourt suggested ?
>
Yes, this is exactly right. IMO there are 2 interesting questions for this problem. What really means `matchCapabilites`/`matchNamespaces` ?
- Is it a "tool" that user can to match capabilities or namespaces (the way I suggested it, for example)
- Is it a process based filter only ? Which means I think we can clarify a bit this meaning by putting those in a specific part like this
```yaml
- matchProcess:
  operator: "Capabilities" # Or Namespaces
  values:
    - "CAP_SYS_ADMIN"
```

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -750,15 +747,11 @@ func (tp *genericTracepoint) eventConfig(config *tracingapi.EventConfig) (*traci
 		config.ArgTpCtxOff[i] = uint32(tpArg.CtxOffset)
 		config.Arg[i] = int32(tpArg.genericTypeId)
 		config.ArgM[i] = uint32(tpArg.MetaArg)
+		config.ArgIndex[i] = int32(tpArg.TpIdx)
 
-		tracepointLog.Debug(fmt.Sprintf("configured argument #%d: %+v (type:%d)", i, tpArg, tpArg.genericTypeId))
-	}
+		fmt.Printf("KRAVA i %d ArgIndex %d ArgTpCtxOff %d\n", i, config.ArgIndex[i], config.ArgTpCtxOff[i])
```
kkourt: leftover?

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -203,8 +203,11 @@ type PIDSelector struct {
 
 type ArgSelector struct {
 	// +kubebuilder:validation:Minimum=0
-	// Position of the argument to apply fhe filter to.
+	// Position of the argument (in function prototype) to apply fhe filter to.
 	Index uint32 `json:"index"`
+	// +kubebuilder:validation:Minimum=0
+	// Position of the argument (in spec file) to apply fhe filter to.
+	Arg *uint32 `json:"arg,omitempty"`
```
kkourt, line 210: Nice, I like this solution!

kkourt (approved):
> got ~9% speedup for following spec when monitoring kernel build:

Nice!

LGTM, thanks!

(please remove the Printf before merging)

bpf/process/generic_calls.h
```diff
@@ -598,6 +627,10 @@ generic_process_event_and_setup(struct pt_regs *ctx, struct bpf_map_def *tailcal
 	generic_process_init(e, MSG_OP_GENERIC_TRACEPOINT, config);
 #endif
 
+	/* No arguments, go send.. */
+	if (arg_idx(0) == -1)
+		tail_call(ctx, tailcals, TAIL_CALL_ARGS);
```
mtardy, line 632: I imagine it's correct, I'm not super familiar with our tail call ordering but the comment seems a bit contradictory to `TAIL_CALL_ARGS`? I imagine that's just the naming

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -43,6 +49,31 @@ selectors:
...
     - "/etc/shadow"
 ```
+In the next example, a selector is defined with a `matchArgs` filter that tells
+the BPF code to process only the function call for which the second spec argument,
+`arg` equals to 1 (which represents 1st function argument, `index` equals to 0),
+has value `0xcoffee`.
+
+```yaml
+selectors:
+- matchArgs:
```
mtardy: I think there is a small mistake in your example, you wanted to first specify args and then the selectors I imagine :)
```suggestion
- args:
```

pkg/sensors/tracing/generictracepoint.go
olsajiri: yep, will remove :) thanks

bpf/process/generic_calls.h
olsajiri: yea, it looks weird ;-) we're just skipping the argument retrieval and jump over to arg filtering like the rest of the code, I'll check if we can skip even further and go to send directly in follow up, there might be still some filtering in that code, thanks

docs/content/en/docs/concepts/tracing-policy/selectors.md
olsajiri: ah yea, nice :) thanks

PR #3736, opened 2025-05-13
tetragon/windows: CI Fix attempt by adding -Wait switch in racy steps

ExceptionalHandler:
Fixes 

### Description
The error `attach type not found for GUID` appears when the ebpf program being attached is not found in the Windows' ebpf store. We register the program in a step that calls `./setup-windows.ps1` script. However, the step appears immediately after the step to clear the ebpf store. This could be racy as Start-Process cmdlet is called with the -Wait switch.

This commit calls the ebpf store clear and registration commands with the -Wait switch to avoid any races


kkourt:
> This commit calls the ebpf store clear and registration commands with the -Wait switch to avoid any races

The PR message seems different from what the patch does. Am I missing something?

ExceptionalHandler:
> > This commit calls the ebpf store clear and registration commands with the -Wait switch to avoid any races
> 
> The PR message seems different from what the patch does. Am I missing something?

Sorry. Human Error. Fixed

kkourt (approved):

Issue #3734, opened 2025-05-13
bpf: issue with resolve on the second argument.


kkourt:
Consider the following two policies

First:
```yaml
apiVersion: cilium.io/v2alpha1
kind: TracingPolicy
metadata:
  name: "capset"
  annotations:
    author: "Isovalent"
spec:
  kprobes:
    - call: "security_capset"
      syscall: false
      selectors:
      args:
        - index: 1
          type: "cap_effective"
          resolve: "cap_effective"
          label: "old_creds.cap_effective"
```

Second:
```yaml
apiVersion: cilium.io/v2alpha1
kind: TracingPolicy
metadata:
  name: "capset"
  annotations:
    author: "Isovalent"
spec:
  kprobes:
    - call: "security_capset"
      syscall: false
      selectors:
      args:
        - index: 0
          type: "cap_effective"
          resolve: "cap_effective"
          label: "new_creds.cap_effective"
        - index: 1
          type: "cap_effective"
          resolve: "cap_effective"
          label: "old_creds.cap_effective"
```

Executing sudo /bin/true gives different results.

First:
```json
    "args": [
      {
        "cap_effective_arg": "0000000000000002",
        "label": "old_creds.cap_effective"
      }
    ],
```

Second:
```json
    "args": [
      {
        "cap_effective_arg": "000001ffffffffff",
        "label": "new_creds.cap_effective"
      },
      {
        "cap_effective_arg": "000001ffffffffff",
        "label": "old_creds.cap_effective"
      }
    ],
```

AFAICT, the  correct answer is the one provided by he second policy.

kkourt:
This seems to fix the issue:

```diff
diff --git a/pkg/sensors/tracing/generickprobe.go b/pkg/sensors/tracing/generickprobe.go
index b5cc06994..45d6a219f 100644
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -753,7 +753,7 @@ func addKprobe(funcName string, instance int, f *v1alpha1.KProbeSpec, in *addKpr
                        if err != nil {
                                return errFn(fmt.Errorf("error on hook %q for index %d : %w", f.Call, a.Index, err))
                        }
-                       allBTFArgs[j] = btfArg
+                       allBTFArgs[a.Index] = btfArg
                        argType = findTypeFromBTFType(a, lastBTFType)
                }
```

Will take a closer look and if everything checks out, I'll do a PR tomorrow.

[...]

PR #3737, opened 2025-05-14
tetragon: fix resolve issue with argument order

kkourt:
Running the following policy:
```
kind: TracingPolicy
metadata:
  name: "resolve-sockaddr"
spec:
  kprobes:
  - call: "security_socket_connect"
    syscall: false
    args:
    - index: 1
      type: "uint16"
      resolve: "sa_family"
```

Leads to an event wih the wrong value. The issue is that when setting allBTFArgs we need to use the index in the function order rather than the index in the spec order.

This commit fixes the issue, and adds a test for it.

Fixes: https://github.com/cilium/tetragon/issues/3734

```release-note
tracingpolicy: fix issue in argument order with the resolve argument option
```


kevsecurity (approved):

olsajiri (approved):
good catch, thnx

PR #3738, opened 2025-05-14
tetragon: Move some event_config values to arrays

olsajiri:
We plan to increase number of supported arguments and it's more convenient with array size.


kkourt (approved):
❤️

PR #3740, opened 2025-05-14
Temporarily disable controller-runtime metrics

michi-covalent:
controller-runtime runs its own metrics server on port 8080 by default. It would be good if we can avoid running multiple metric servers. I need some time to think about how to consolidate Tetragon's metric server and controller-runtime metric server.


sayboras (approved):
Thanks, we have done similar thing in cilium/cilium, additionally, the custom metric server is used.

https://github.com/cilium/cilium/pull/23501

kkourt (approved):

PR #3747, opened 2025-05-16
reduce image size by compressing bpf objs

kkourt:


Dockerfile
```diff
@@ -12,10 +12,12 @@
 # First builder (cross-)compile the BPF programs
 FROM --platform=$BUILDPLATFORM quay.io/cilium/clang:b97f5b3d5c38da62fb009f21a53cd42aefd54a2f@sha256:e1c8ed0acd2e24ed05377f2861d8174af28e09bef3bbc79649c8eba165207df0 AS bpf-builder
 WORKDIR /go/src/github.com/cilium/tetragon
-RUN apt-get update && apt-get install -y linux-libc-dev
+RUN apt-get update && apt-get install -y linux-libc-dev gzip
 COPY . ./
 ARG TARGETARCH
+ARG COMPRESS_BPF
 RUN make tetragon-bpf LOCAL_CLANG=1 TARGET_ARCH=$TARGETARCH
+RUN if [ "$COMPRESS_BPF" = "gzip" ]; then gzip bpf/objs/*.o; fi
```
mtardy, line 20: that's pretty convenient for this use case that gzip by default delete the original uncompressed file!

mtardy (approved):
This is really cool!

Dockerfile
kkourt: Most (if not all) the compression tools do this :)

PR #3748, opened 2025-05-16
tetragon/windows: Fix exec/exit event timestamp in event json

ExceptionalHandler:
### Description
On Windows the exec/exit event timestamp was serialized incorrectly. All timestamps in Windows are 100 Nanoseconds from 1 Jan 1600. The protobuf serialization routine assumed nanoseconds wrt boot time or monotonic time, which were then converted to Linux Epoch.

This change converts the Windows Epoch based 100 Nanoseconds to Linux Epoch based time before Event is generated. Both existing and new processes are made to report time wrt to Windows Epoch.

### DIFF
--- a/bpf/windows/process_monitor.c
+++ b/bpf/windows/process_monitor.c
@@ -115,13 +115,16 @@ int ProcessMonitor(process_md_t *ctx)
 	if (ctx->operation == PROCESS_OPERATION_CREATE) {
 		struct process_create_info_t process_create_info;
 
-		memset(&process_create_info, 0, sizeof(process_create_info));
+		int size = sizeof(process_create_info);
+
+		memset(&process_create_info, 0, size);
 		process_create_info.common.op = MSG_OP_EXECVE;
+		process_create_info.common.size = size;
+		process_create_info.common.ktime = ctx->creation_time;
 		process_create_info.process_id = ctx->process_id;
 		process_create_info.parent_process_id = ctx->parent_process_id;
 		process_create_info.creating_process_id = ctx->creating_process_id;
 		process_create_info.creating_thread_id = ctx->creating_thread_id;
-		process_create_info.common.ktime = ctx->creation_time;
 		process_create_info.creation_time = ctx->creation_time;
 
 		void *buffer = get_scratch_space();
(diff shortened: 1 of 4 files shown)


pkg/ktime/ktime_windows_test.go
```diff
@@ -25,3 +25,8 @@ func TestBoottime(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Greater(t, ktime1.UnixMilli(), int64(0))
 }
+
+func TestKernelime(t *testing.T) {
```
mtardy, line 29: looks like a small typo
```suggestion
func TestKernelTime(t *testing.T) {
```

bpf/windows/process_monitor.c
```diff
@@ -147,13 +151,15 @@ int ProcessMonitor(process_md_t *ctx)
...
 	} else if (ctx->operation == PROCESS_OPERATION_DELETE) {
 		struct process_exit_info_t process_exit_info;
+		int size = sizeof(process_exit_info);
 
-		memset(&process_exit_info, 0, sizeof(process_exit_info));
+		memset(&process_exit_info, 0, size);
 		process_exit_info.process_id = ctx->process_id;
+		process_exit_info.common.op = MSG_OP_EXIT;
 		process_exit_info.common.ktime = ctx->exit_time;
+		process_exit_info.common.size = size;
```
mtardy, line 160: btw might be okay to reuse `sizeof(static)` in multiple places since in this case it should be evaluated at compile time. I imagine the compiler will optimize the size variable away!

mtardy (approved):
looks good, only a few nits!

bpf/windows/process_monitor.c
```diff
@@ -115,13 +115,17 @@ int ProcessMonitor(process_md_t *ctx)
...
+		memset(&process_create_info, 0, size);
 		process_create_info.common.op = MSG_OP_EXECVE;
+		process_create_info.common.size = size;
+		process_create_info.common.ktime = ctx->creation_time;
 		process_create_info.process_id = ctx->process_id;
 		process_create_info.parent_process_id = ctx->parent_process_id;
 		process_create_info.creating_process_id = ctx->creating_process_id;
 		process_create_info.creating_thread_id = ctx->creating_thread_id;
-		process_create_info.common.ktime = ctx->creation_time;
+
```
mtardy, line 128: also a nit: whitespace churn here

PR #3760, opened 2025-05-22
Sensors: Use require over assert in tests

kevsecurity:
Most of the places where we have been using assert in tests, we really wanted the tests to fail at those points and not continue. In these cases, we should use require instead as it does just that. This commit changes a lot of asserts to requires.


lambdanis (approved):
looks innocent

Issue #2651, opened 2024-07-09
Cleanup unnecessary cilium/cilium dependencies


lambdanis:
I used [gomod](https://github.com/Helcaraxan/gomod) to check how Tetragon depends on cilium/cilium Go packages. Here we go:
```
gomod graph --style cluster=full -p 'deps(github.com/cilium/tetragon/**, 1) inter rdeps(github.com/cilium/cilium/**, 1)' > tetragon-cilium.dot && dot -Tpng -o tetragon-cilium.png tetragon-cilium.dot
```
![tetragon-cilium](https://github.com/cilium/tetragon/assets/3199463/8c84cd3d-7c8a-4cb0-adfd-3b3a8c17c460)
(To make it clear - these are dependencies in Go code only. Tetragon doesn't have runtime dependencies on Cilium).

Many of these are just small helpers imported from cilium/cilium for historical reasons. We should review them and remove unnecessary dependencies. If Tetragon is importing a package from cilium/cilium just to use a small helper, it might make sense to write such helper inside Tetragon codebase.

Some cilium/cilium dependencies were removed already in #794/#2580, but there are several remaining. What's particularly problematic are transitive dependencies of k8s libraries via cilium/cilium. Because not all interfaces are stable, this means version of k8s libraries is tied to cilium/cilium version. Here's a graph of such dependencies:
```
gomod graph --style cluster=full -p 'deps(github.com/cilium/tetragon/**, 1) inter rdeps(github.com/cilium/cilium/**, 1) inter (rdeps(k8s.io/**) + rdeps(sigs.k8s.io/**))' > tetragon-cilium-k8s.dot && dot -Tpng -o tetragon-cilium-k8s.png tetragon-cilium-k8s.dot
```
![tetragon-cilium-k8s](https://github.com/cilium/tetragon/assets/3199463/cf0db294-aab3-42da-bd6d-8b34f253c8b9)

mtardy:
Another dependency that is not great is within the bpf package https://github.com/cilium/tetragon/pull/2717#issuecomment-2244670751.

yashathwani:
Would like  to  work on this issue could you please assign it to me !!

mtardy:
> Would like to work on this issue could you please assign it to me !!

sure go ahead, please feel free to ask questions. :)

PR #3764, opened 2025-05-23
deps: Remove cilium hubble package dependency

sayboras:
### Description

The event and filter can be ported with minimum changes.

Relates: https://github.com/cilium/tetragon/issues/2651

### Changelog

```release-note
deps: Remove cilium hubble package dependency
```


sayboras:
Seems like the below job is timeout, is it known issue? If yes, can someone help to re-run? Thanks

https://github.com/cilium/tetragon/actions/runs/15210176533/job/42782653770?pr=3764

mtardy:
> Seems like the below job is timeout, is it known issue? If yes, can someone help to re-run? Thanks
> 
> https://github.com/cilium/tetragon/actions/runs/15210176533/job/42782653770?pr=3764

yep it can happen, I re ran

mtardy (approved):
Hopefully those filters have some unit tests on them! If they don't, maybe they deserve to break? Thanks that's a major dependency removal! If only we could remove cilium entirely!

mtardy:
you motivated me, I'm trying to see how difficult it would be to remove all `github.com/cilium/cilium`, let's see

sayboras:
> you motivated me, I'm trying to see how difficult it would be to remove all `github.com/cilium/cilium`, let's see

:+1: just a note that replacing slim/meta/v1 with k8s meta/v1 is a breaking change, as we have a custom LabelValue type for extra validation.

kkourt:
> > you motivated me, I'm trying to see how difficult it would be to remove all `github.com/cilium/cilium`, let's see
> 
> 👍 just a note that replacing slim/meta/v1 with k8s meta/v1 is a breaking change, as we have a custom LabelValue type for extra validation.

Would it be a breaking change if we just copy the exact types from the cilium repo?

PR #3766, opened 2025-05-24
deps: Remove Cilium slim k8s package

sayboras:
Relates: https://github.com/cilium/tetragon/issues/2651

Fixes 

### Changelog

```release-note
```


mtardy:
I was wondering why important the full package? It seems we just need something like 3 or 4 types declarations? I might be missing something but it seemed enough to have the LabelSelector and its deps?

kkourt (approved):

PR #3772, opened 2025-05-29
feat : Add `Capability` operator for `matchArgs`

tdaudi:
This PR follow the discussion started here https://github.com/cilium/tetragon/pull/3730#discussion_r2086015875

## Aim

The purpose of this PR is to allow a user to extract a capability field from the hook parameters and use it against the `matchCapabilities`  tool to compare the capability. This can be useful if you want to detect when a program requests a new capability (for example, in `security_capset`).

## How to test the PR

```bash
# Load the policy
./tetra tracingpolicy add examples/tracingpolicy/deny-privileged-containers.yml

# Try to run a normal container. It should work
docker run --rm hello-world

# Run a privileged container. It should return a permission denied
docker run --rm --privileged hello-world
```

The policy also works for KinD.

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -1537,6 +1537,7 @@ There are different types supported for each operator. In case of `matchArgs`:
 * Protocol
 * Family
 * State
+* CapabilityMask (see an [example](https://github.com/cilium/tetragon/blob/main/examples/tracingpolicy/deny-privileged-containers.yml))
 
 The operator types `Equal` and `NotEqual` are used to test whether the certain
 argument of a system call is equal to the defined value in the CR.
@@ -1631,6 +1632,14 @@ can be specified as either `AF_INET6` or 10.
 The `State` operator can accept integer values to match against or the equivalent
 TCP_ enumeration. For example, an established socket can be matched with
 `TCP_ESTABLISHED` or 1; a closed socket with `TCP_CLOSE` or 7.
+
+The `CapabilityMask` operator can be used to match a kernel capability from the hook.
+The concept is very similar to `matchCapabilities`, but is restricted to arguments you get from the hook.
+The filter accepts all the available capabilities (see also the kernel source).
+It performs a bitmask AND against the value. For example, if you want to match `CAP_SYS_ADMIN`,
+Tetragon will change the value to the mask, by doing `1 << 27` in this case, and use it to match the
+resolved value: `value & mask`. You can also directly use the `Mask` operator to perform
+such matching, by converting the capability to the mask and using it in the policy.
 
 In case of `matchPIDs`:
 


examples/tracingpolicy/deny-privileged-containers.yml
```diff
@@ -0,0 +1,29 @@
...
+      matchBinaries:
+      - operator: "In"
+        values:
+        - "/usr/local/sbin/runc" # runc in KinD
+        - "/usr/bin/runc"        # runc in Docker
+      matchCapabilities:
+      - index: 2
+        operator: In
+        values:
+        - "CAP_SYS_ADMIN"
```
olsajiri, line 25: I'm scratching my head how is this different from:
```
      matchArgs:
      - index: 2
        operator: "In"
        values:
        - "21"
```

bpf/process/generic_calls.h
```diff
@@ -130,14 +130,14 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 		     : "i"(MAX_SELECTORS_MASK));
 	ty = config->arg[index];
 
-	a = (&e->a0)[index];
-	extract_arg(config, index, &a);
+	a = &(&e->a0)[index];
+	extract_arg(config, index, a);
```
olsajiri: that clashes with the https://github.com/cilium/tetragon/pull/3730 where we allow to extract data multiple times from the same argument, s we need to keep the original value ... but I guess we could add extra values for this storage if it comes to that

bpf/process/generic_calls.h
```diff
@@ -62,7 +62,7 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 	msg->common.flags = 0;
 
 	/* Tail call into filters. */
-	tail_call(ctx, calls, TAIL_CALL_FILTER);
+	tail_call(ctx, calls, TAIL_CALL_SETUP);
 	return 0;
```
olsajiri: so all the tail call changes moves arguments retrieval before filter, right? that might have some performance implications, there's potentially lot of code and we currently read all the arguments regardless of the configuration.. even though this is going to change in https://github.com/cilium/tetragon/pull/3730 I think we should at least measure some common use cases

nit I wonder you'd need to make all the tail call changes within one commit to keep bisect working

bpf/process/pfilter.h
```diff
@@ -248,6 +250,10 @@ process_filter_capabilities(__u32 ty, __u32 op, __u32 ns, __u64 val,
 
 	caps = c->c[ty];
 
+	if (idx > -1 && idx < EVENT_CONFIG_MAX_ARG)
+		// Supported types kernel_cap_ty, cap_inh_ty, cap_prm_ty, cap_eff_ty
+		probe_read(&caps, sizeof(__u64), (char *)(&msg->a0)[idx]);
```
olsajiri: hum, why is the probe_read needed?

bpf/process/generic_calls.h
```diff
@@ -62,7 +62,7 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 	msg->common.flags = 0;
 
 	/* Tail call into filters. */
-	tail_call(ctx, calls, TAIL_CALL_FILTER);
```
olsajiri: missing comment change

olsajiri (commented):
heya, thanks for the change, I left some comments for discussion

examples/tracingpolicy/deny-privileged-containers.yml
tdaudi: 21 is the bit number set for the capability, not the value itself. The unit test I added may be more explicit, I think

bpf/process/generic_calls.h
tdaudi: Yes I know. This is why I wanted to finish developing this PR before you PR. Though I pushed an alternative (very cheap compare to yours, I acknowledge), https://github.com/cilium/tetragon/pull/3792

But does the changes you applied to your PR after the discussion https://github.com/cilium/tetragon/pull/3730#discussion_r2086015875 make it usable again ? I'm not sure to understand

bpf/process/generic_calls.h
tdaudi: > so all the tail call changes moves arguments retrieval before filter, right? that might have some performance implications, there's potentially lot of code and we currently read all the arguments regardless of the configuration.. even though this is going to change in #3730 I think we should at least measure some common use cases

Yes 100% agree. If I'm correct, the impact will only be on policies with the original `matchCapabilities` that fails (because if they don't fail it ran the same tailcalls right ?). But this is just a guess.

> 
> nit I wonder you'd need to make all the tail call changes within one commit to keep bisect working

Ok, I'll change this. Thanks

bpf/process/pfilter.h
tdaudi: Because kernel_cap_ty is as below. Initially I used `(&msg->a0)[idx]` only, but in the policy it means doing something less convenient / userfriendly  :

```yaml
- index: 2
  type: uint64
  resolve: val
```

```c
typedef struct { u64 val; } kernel_cap_t;
```

examples/tracingpolicy/deny-privileged-containers.yml
olsajiri: I see, but we do have the mask operator:
```
    matchArgs:
      - index: 2
        operator: "Mask"
        values:
        - "2097152" # 1 << 21
```

bpf/process/generic_calls.h
olsajiri: yes, it allows spec like:
```
    args:
    - index: 0
      type: "int"
      resolve: "mm.owner.pid"
    - index: 0
      type: "int"
      resolve: "mm.owner.mm.owner.pid"
```

so we need to keep the original argument value

tdaudi: Ah, I see your changes, I think you're right on this. I'll think about it and push a change.

>  but I guess we could add extra values for this storage if it comes to that

examples/tracingpolicy/deny-privileged-containers.yml
tdaudi: Yes, it does work with this configuration. But this seems pretty hardcore to use and to understand for a user.

```yaml
args:
    - index: 2
      type: "uint64" 
      resolve: val
...
      matchArgs:
      - index: 2
        operator: Mask
        values:
        - 2097152
```

kkourt: > But this seems pretty hardcore to use and to understand for a user.

That's a fair point, but why must we change the bpf path? If the current bpf path supports the functionality we want, it's a better approach to translate the policy-level improvements we want into the existing bpf path.

For example, we can use:

```
matchArgs:
 - index: 2
   operator: CapabilityMask
   values:
     - CAP_FOO
     - CAP_BAR
```

Or even:

```
matchArgs:
 - index: 2
   operator: CapabilityMask
   values:
     - CAP_FOO, CAP_BAR
```

And translate it into the Mask operator.

kkourt (changes requested):
Thanks for the PR!

I think having a more user-friendly way to match on capabilities arugments makes perfect sense.

However, one issue with the approach that the PR follows is that now we have two ways for matching arguments (`matchArgs` and `matchCapabilities`). Is there a reason why we cannot enhance the existing `matchArgs` mechanism to do what we want? Especially since it seems that doing so does not require any bpf changes. (See also my other comment.)

tdaudi:
> However, one issue with the approach that the PR follows is that now we have two ways for matching arguments (`matchArgs` and `matchCapabilities`). Is there a reason why we cannot enhance the existing `matchArgs` mechanism to do what we want? Especially since it seems that doing so does not require any bpf changes. (See also my other comment.)

In the documentation, it is not very clear that `matchCapbabilities` and `matchNamespaces` are meant for `task_struct` comparison. So since I wanted to compare a capability, I thought that using such a tool was the easiest way. 
So I was thinking that if I face such problem (compare `args:` capability), maybe someone else may found the same solution as me (using `matchCapability`).

Also, my second motivation for doing it this way is that turning `CAP_SYS_ADMIN` (= 21) to `2097152` is already done in the `matchCapabalities` so it involves less code duplication.

---

Other than that, I what you suggest @kkourt with `CapabilityMask` is very good. Probably the easiest solution. But to avoid confusions with `matchCapabilities` it will need some documentation updates.

kkourt:
> > However, one issue with the approach that the PR follows is that now we have two ways for matching arguments (`matchArgs` and `matchCapabilities`). Is there a reason why we cannot enhance the existing `matchArgs` mechanism to do what we want? Especially since it seems that doing so does not require any bpf changes. (See also my other comment.)
> 
> In the documentation, it is not very clear that `matchCapbabilities` and `matchNamespaces` are meant for `task_struct` comparison. So since I wanted to compare a capability, I thought that using such a tool was the easiest way. So I was thinking that if I face such problem (compare `args:` capability), maybe someone else may found the same solution as me (using `matchCapability`).
> 
> Also, my second motivation for doing it this way is that turning `CAP_SYS_ADMIN` (= 21) to `2097152` is already done in the `matchCapabalities` so it involves less code duplication.
> 
> Other than that, I what you suggest @kkourt with `CapabilityMask` is very good. Probably the easiest solution. But to avoid confusions with `matchCapabilities` it will need some documentation updates.

Indeed, we can update the docs to make things clearer.

> Also, my second motivation for doing it this way is that turning `CAP_SYS_ADMIN` (= 21) to `2097152` is already done in the `matchCapabalities` so it involves less code duplication.

Haven't looked at the code, but we should be able to refactor things so that the code that does the mapping is reused.

tdaudi:
Moving to draft until the changes discussed are applied

examples/tracingpolicy/deny-privileged-containers.yml
tdaudi: > For example, we can use:
> 
> ```
> matchArgs:
>  - index: 2
>    operator: CapabilityMask
>    values:
>      - CAP_FOO
>      - CAP_BAR
> ```

This is the way I implemented the operator.

> Or even:
> 
> ```
> matchArgs:
>  - index: 2
>    operator: CapabilityMask
>    values:
>      - CAP_FOO, CAP_BAR
> ```
> 
> And translate it into the Mask operator.

This version is less trivial. I took the same workflow as `matchCapabilities`. All the values for the same selector are concatenated in a single uint64 field and use it on the bpf side. 
If we would like to do the way you suggest, I think it would mean having a separate map for every selector (the same way as Prefix/Postfix selector maps works). I think this is good though, but maybe later if this is needed.

tdaudi:
The CI fails on 4.19. I guess I should restrict the operator to programs with the `__LARGE_BPF_PROG` flag ?

EDIT : Done. The feature requires a 5.4+ kernel

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1537,6 +1537,7 @@ There are different types supported for each operator. In case of `matchArgs`:
 * Protocol
 * Family
 * State
+* Capability (see an [example](https://github.com/cilium/tetragon/blob/main/examples/tracingpolicy/deny-privileged-containers.yml)) - Available for kernel >= 5.4
```
olsajiri: I think it deserves at least simple paragraph below as we have for '* State' and other bullets

bpf/process/types/basic.h
```diff
@@ -1211,6 +1211,18 @@ filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
 					return 1;
 			}
 			break;
+		case op_filter_capability: {
+			__u64 caps;
+
+			// We resolve the kernel_cap_t/cap_* here
+			probe_read(&caps, sizeof(__u64), args);
```
olsajiri: I still think this should work without probe_read, like that this should be enough:

```
+               case op_filter_capability: {
+                       if (*(u64 *)args & w)
+                               return 1;
+                       break;
+               }
```

so it could share the code with op_filter_mask and maybe 4.19 would work?

could you please explain in more details what's the idea behind the resolve?

pkg/selectors/kernel.go
```diff
@@ -200,6 +200,9 @@ const (
 	// more socket ops
 	SelectorOpFamily = 28
 	SelectorOpState  = 29
+
+	// capabilities mask
+	SelectorOpCapability = 30
```
olsajiri: just curious.. is there any point in having negative cap filter as well? to filter all the calls that are not CAP_ADMIN or such?

olsajiri (commented):
looks good, left some comments, thanks

bpf/process/types/basic.h
tdaudi: You are actually very right. I was pretty much sure that it was needed, but it works without it. I'll do the change, thank you 💯

pkg/selectors/kernel.go
tdaudi: This is a very good idea, I hadn't thought of it. I'll see how I can add this as well, thank you

tdaudi:
I pushed an intermediate version to see if 4.19 CI works. If it does, I'll add the `NotCapability` filter. But it will probably be necessary to restrict it to kernel >5.3 because the filter tail call has too many instructions already.

pkg/selectors/kernel.go
tdaudi: @olsajiri, sorry for the ping here, but do you have any recommendation on how we could add this new filter in the tailcal without restricting it to LARGE_BPF_PROG ? Or it does not matter if Capability is available for all the kernel versions but NotCapability only for 5.4+ ?

olsajiri: I guess it's fine.. and it's also fine if we do that as a follow up when we see that it's really needed, I dont insist ;-)

tdaudi: Oh, ok ! So this PR is good at this point.
Thank you for the reviews 💯

bpf/process/types/basic.h
tdaudi: Done, thank you

docs/content/en/docs/concepts/tracing-policy/selectors.md
tdaudi: I added a bit more documentation, thank you.
When I'll have time this summer, I'll try to push an update for all this part of the documentation because I think there are many interesting things to say.

bpf/process/types/basic.h
```diff
@@ -1221,6 +1221,7 @@ filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
 				return 1;
 			break;
 		case op_filter_mask:
+		case op_filter_capability:
```
kkourt: Is this bpf code change needed? Can't we juts use `op_filter_mask`? (We can still keep the Capability operator at the spec).

examples/tracingpolicy/deny-privileged-containers.yml
kkourt: > This version is less trivial. I took the same workflow as matchCapabilities. All the values for the same selector are concatenated in a single uint64 field and use it on the bpf side.
If we would like to do the way you suggest, I think it would mean having a separate map for every selector (the same way as Prefix/Postfix selector maps works). I think this is good though, but maybe later if this is needed.

Not sure I understand. One thing to keep in mind is that `values:` takes a list.

So what happens, if a user writes:

```
values:
  - CAP_FOO
  - CAP_BAR
```

Compared to writing:
```
values
 - CAP_FOO, CAP_BAR
```

Both cases seem the same to me, i.e., match if the capability is CAP_FOO _or_ CAP_BAR.
In which case, both can be translated as (cap & (CAP_FOO | CAP_BAR)).

Given above is correct, I would argue that it's better to use the existing list in the values, rather than introduce parsing for the `,`.


Moreover, using `OperatorMask` instead of `Operator` seems better to me because it hints that this is syntactic sugar for the Mask operator.

bpf/process/types/basic.h
tdaudi: I do not see any way to do it without code change.

https://github.com/cilium/tetragon/blob/50ad9dd37bcc88f0128ae21f159af46016ea5d78/pkg/selectors/kernel.go#L772-L783

In this code, the variable `op` store the operation to perform. This value is pushed with `WriteSelectorUint32(&k.data, op)` before the switch case that prepare the `values:`. And in this switch case, the mask operator, follow the default behaviour (see below), and we lose the information that say that this is Capability or this is a Mask.

https://github.com/cilium/tetragon/blob/50ad9dd37bcc88f0128ae21f159af46016ea5d78/pkg/selectors/kernel.go#L856-L861

tdaudi: One solution I see could be to move all this switch case before we store its value, and after it is done we register all the fields. Doing so, we will be able to rewrite the operator after building the values

examples/tracingpolicy/deny-privileged-containers.yml
tdaudi: > Given above is correct, I would argue that it's better to use the existing list in the values, rather than introduce parsing for the `,`.

Oh, sorry if I explain poorly for this.
I do not parse any `,`. The policy will simply look similarly to `matchCapabilities`. So capability will simply be specified as a list. 

> Moreover, using `OperatorMask` instead of `Operator` seems better to me because it hints that this is syntactic sugar for the Mask operator.

So my choice was motivated by the fact that I want users to require zero effort to understand how capability matching works.
I am afraid that `CapabilityMask` could be more confusing because you let the user this understanding of capabilities mask comparison, which requires more kernel understanding in my opinion.
Also, I thought that it was a good idea to stay closer to the name `matchCapabilities`

kkourt: In my opinion the fact that the selector below:

 ```
matchArgs:
 - index: 2
   operator: Capability
   values:
     - CAP_FOO
     - CAP_BAR
 ```
 
 matches only when *both* capabilities match is confusing. I think using `operator: Capability` can led to the impression the selector will match when either of the capabilities is set. 

Personally, I would prefer using `CapabilityMask` but if folk find this confusing, I would suggest that we use `operator: Capabilities`.

mtardy (commented):
Hey, I'm catching the discussion on the fly.

From what I read from the comments here, it looks like this change could be entirely a syntactic sugar done from userspace without touching the BPF (I know it's a bit frustrating hehe).

It's legitimate to think that this is not super good UX for matching capabilities over an argument:
```yaml
    matchArgs:
      - index: 2
        operator: "Mask"
        values:
        - "2097152" # 1 << 21
```

I imagine the next step is deciding on a word for the operator that makes it clear how the YAML list will be processed to translate to the mask.
```yaml
matchArgs:
- index: 2
  operator: <smartWording>
  values:
    - CAP_FOO
    - CAP_BAR
 ```
 
and just automatically translate that matchArgs into the other one suggested by Jiri in the policy processing. Does that sound fair or I'm missing something?

File: examples/tracingpolicy/deny-privileged-containers.yml (hunk 1 above)
Comment (mtardy) [reply examples/tracingpolicy/deny-privileged-containers.yml:25]: > I am afraid that CapabilityMask could be more confusing because you let the user this understanding of capabilities mask comparison, which requires more kernel understanding in my opinion.

yeah but the user needs to understand this! it needs to be clear that this capabilities are bitmask and that you are going to OR them! So yeah I kind of agree that `CapabilityMask` is more explicit, given we already chose mask and not bitmask for the other operator

Comment (tdaudi) [reply examples/tracingpolicy/deny-privileged-containers.yml:25]: Thanks for both your answers ! I think I see your point here. I'll push the change later today.

So we will have `CapabilityMask` in this PR and later `NotCapabilityMask` to match if the argument does not have the capabilities.
It seems a bit strange to me, but it follows the logic.

File: bpf/process/types/basic.h (hunk 9 above)
Comment (tdaudi) [reply bpf/process/types/basic.h]: Done, I think I implement the change the easiest way.
Let me know what you think

**Comment by tdaudi** [comment]
Latest change includes :

- Changing operator name from `Capability` to `CapabilityMask`
- Removing all reference to the new `op_filter_capability`. The workflow now follows `Mask` operator.
- `Mask` operator now handles `kernel_cap_t` and `cap_*` types
- Updating the docs details to explain this `Mask` behavior. (Let me know if it is understandable enough)

**Comment by kkourt** [comment]
@tdaudi I was trying to fix a bug (which now I realized that you had also fixed in your PR), and ended up with: https://github.com/cilium/tetragon/pull/3852. WDYT?

**Review by olsajiri — APPROVED** [review]
lgtm, needs rebase though

**Comment by tdaudi** [comment]
> lgtm, needs rebase though

Do you think the PR is good to merge with such implementation ? 
With the discussion in https://github.com/cilium/tetragon/pull/3852#issuecomment-3008509829, and with all the messages before, I think that the PR can  be improved.

To summarize,

### What is my problem

I need to check if one or many capabilities exist in a `u64` field.
Let's say the extracted `value = 0xffffffffffffffff`, but I want to see if only the `CAP_SYS_ADMIN` capability exists. For that, I need to convert `CAP_SYS_ADMIN` to a mask and use it.
The current state of the PR solves this problem.

But what if I want to check if both `CAP_A` AND `CAP_B` exists? The Mask operator does not provide a way to do so  (I give more context about it here https://github.com/cilium/tetragon/pull/3852#discussion_r2166696377).

Changing the `Mask` operator may be a breaking change because ATM it checks a single bit of the mask is set to true while I want all the bit of the mask to return true.

### What I suggest

Should we introduce a new mask operator for this specific case ?
Rename the current `Mask` operator to something like `MaskContain` (which means one bit of the mask exists) and create a new  name `MaskInclude` (all the mask exists in the extracted value). 
Later, another `MaskExclude` could make sense, as suggested in this comment https://github.com/cilium/tetragon/pull/3772#discussion_r2159708529

The `Include` and `Exclude` semantic also make sense for strings, so maybe there is something to push this way

What do you think is the best option here ?
CC: @kkourt 


EDIT : This small change may be good enough

```diff
diff --git a/bpf/process/types/basic.h b/bpf/process/types/basic.h
index 1bbea6abd..09a5ba93b 100644
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1221,7 +1221,7 @@ filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
                                return 1;
                        break;
                case op_filter_mask:
-                       if (*(u64 *)args & w)
+                       if ((*(u64 *)args & w) == w)
                                return 1;
                default:
                        break;
@@ -1317,7 +1317,7 @@ filter_32ty_selector_val(struct selector_arg_filter *filter, char *args)
                                return 1;
                        break;
                case op_filter_mask:
-                       if (*(u32 *)args & w)
+                       if ((*(u32 *)args & w) == w)
                                return 1;
                default:
                        break;
```

olsajiri:
> But what if I want to check if both `CAP_A` AND `CAP_B` exists? The Mask operator does not provide a way to do so (I give more context about it here [#3852 (comment)](https://github.com/cilium/tetragon/pull/3852#discussion_r2166696377)).

yea, current mask operator succeeds when one (or more) values are set,
if you need to check all the values, do we need to do something like:

```
@@ -1184,6 +1184,7 @@ FUNC_LOCAL long
 filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
 {
        __u64 *v = (__u64 *)&filter->value;
+       bool mask_all = true;
        int i, j = 0;
 
 #pragma unroll
@@ -1223,6 +1224,9 @@ filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
                case op_filter_mask:
                        if (*(u64 *)args & w)
                                return 1;
+                       break;
+               case op_filter_mask_all:
+                       mask_all &= *(u64 *)args & w;
                default:
                        break;
                }
@@ -1230,6 +1234,9 @@ filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
                if (j + 8 >= filter->vallen)
                        break;
        }
+
+       if (filter->op == op_filter_mask_all && mask_all)
+               return 1;
        return 0;
 }
```

kkourt:
@tdaudi do you still plan to merge this PR given that we merged https://github.com/cilium/tetragon/pull/3852?

tdaudi:
> @tdaudi do you still plan to merge this PR given that we merged #3852?

If no further changes are needed, I'm closing this PR.
Thanks a lot for #3852

PR #3785, opened 2025-05-30
tetragon/windows: add uid to exec events in Windows

ExceptionalHandler:
### Description
This commit consumes the changes made in process_monitor.c program to send the  user's login id as a part of exec event. 
This id is also enumerated for existing processes during enumeration when tetragon starts.

This uid is a unique Login Identifier extracted from user's token, and is same as Token AuthenticationId field of TOKEN_STATISTICS structure, available when token is queried with TokenStatistics information class.

This Authetication Id can be used to retrieve back the token, sid, user name etc. fields using the win32 API LsaGetLogonSessionData()

Since the Login ID as a uint64 is unique per user, and is different between privileged and non-privileged sessions of the same user, it seems to be  a good proxy for uid field in exec event.

### Changelog

```Exec events in Windows are now reported with uid to identify user launching the process. 
```

### DIFF
--- a/pkg/api/processapi/processapi_windows.go
+++ b/pkg/api/processapi/processapi_windows.go
@@ -9,6 +9,7 @@ type MsgCreateProcessEvent struct {
 	ParentProcessID   uint32
 	CreatingProcessID uint32
 	CreatingThreadID  uint32
+	UserLUID          uint64
 	CreationTime      uint64
 }
 
--- a/bpf/windows/process_monitor.c
+++ b/bpf/windows/process_monitor.c
@@ -125,6 +125,7 @@ int ProcessMonitor(process_md_t *ctx)
 		process_create_info.parent_process_id = ctx->parent_process_id;
 		process_create_info.creating_process_id = ctx->creating_process_id;
 		process_create_info.creating_thread_id = ctx->creating_thread_id;
+		process_create_info.user_luid = bpf_get_current_logon_id(ctx);
 		process_create_info.creation_time = ctx->creation_time;
 
 		void *buffer = get_scratch_space();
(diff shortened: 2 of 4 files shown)


pkg/api/processapi/processapi_windows.go
```diff
@@ -9,6 +9,7 @@ type MsgCreateProcessEvent struct {
 	ParentProcessID   uint32
 	CreatingProcessID uint32
 	CreatingThreadID  uint32
+	UserLUID          uint64
```
tixxdz, line 12: Hi @ExceptionalHandler sorry missed last windows dev, but I want to know is this how the field will be presented to users?  or is it internal to us only dev?

On linux loginuid https://tetragon.io/docs/reference/grpc-api/#process auid field is same but uid (effective uid) will change according to privileges, here it seems the opposite right? 

Feels more of an effective uid in Linux world... what are the other User IDs that we could report? just curious.

ExceptionalHandler: This is same as uid or effective UID. When user acquires new privileges, this uid will also change. It is derived from the current user's token, so LoginUID is a little misnomer. That is why I have added it in the uid field itself. 

Windows does not have a numeric UID (like 1000 on Linux). User identity is best expressed as SIDs which is a complex structure generally represented a string (like S-1-2-1000). Some software use the last value (1000) as a numeric id. Windows internally uses the above LoginUID where numeric IDs are needed. APIs are available for to-and-fro conversion from SID/token and how the numeric ID is calculated is opaque. (It is believed to be some kind of hash of the SID but that cannot be confirmed). 

We can find the token user logged in using the session identifier ( the real login id)  but that might require making changes to the ebpf extensions.

tixxdz: Alright got it! Thank you @ExceptionalHandler

PR #3791, opened 2025-05-31
docs: Remove EventChecker process_exec.yaml

sayboras:
### Description
There is no CRD EventChecker in cilium.io/v1alpha1, so probably we can just remove this example file.

### Changelog

```release-note
docs: Remove EventChecker process_exec.yaml
```


sayboras:
ah the file is used for test, k8s format is used.

PR #3798, opened 2025-06-03
Do proper cleanup on maps during sensor unload

tpapagian:
Please check commit messages.


pkg/sensors/tracing/generickprobe.go
```diff
@@ -658,6 +658,15 @@ func createGenericKprobeSensor(
...
 		DestroyHook: func() error {
+			for _, id := range ids {
+				if gk, err := genericKprobeTableGet(id); err == nil {
+					selectors.ClearKernelSelectorState(gk.loadArgs.selectors.entry)
+					if gk.loadArgs.retprobe {
+						selectors.ClearKernelSelectorState(gk.loadArgs.selectors.retrn)
+					}
+				}
+			}
+
```
kkourt: We do `genericKprobeTable.RemoveEntry(id)` in the code below.
Why do we also need to clear the selectors? Do we hold other referrences of these entries somewhere else?
Because I would expect GC to clean the selector state once there are no more references.

mtardy: > Do we hold other referrences of these entries somewhere else?

That might be the only explanation otherwise this should be garbage collected. Unfortunately there's no good Go tooling to see who's keeping a ref on something. So we will need to understand!

olsajiri: > We do genericKprobeTable.RemoveEntry(id) in the code below.

right, it should be taken care of by that removal.. do we have heap profile where this stands out?

another thing is that we could maybe cleanup selectors already in PostLoadHook, because it's no longer needed once the programs/maps are loaded

tpapagian: Yes, I agree that `genericKprobeTable.RemoveEntry(id)` should be enough but it seems that we miss something here. I spend time before the PR and after creating that to understand what we are missing. I will continue on that. 

> do we have heap profile where this stands out?

After running `for ((i=0;i<1000;i+=1)) do ./tetra tracingpolicy add ./pol.yaml && ./tetra tracingpolicy delete policy-example; done` where the policy is a simple kprobe based policy with matchArgs. 

A screenshot of the heap profile is
Screenshot 2025-06-04 at 1 15 01 PM

tpapagian: I believe that I found the real issue. This is again ready for review (this has some issues with the CI so I am  working on that).

tpapagian: Now, this is ready for review.

pkg/sensors/load.go
```diff
@@ -254,6 +254,24 @@ func (s *Sensor) Destroy(unpin bool) {
...
+	// s.Progs is a list of pointers to program.Program where each of these
+	// contains the PinMap field that is a map where its values are pointers
+	// to program.Map that points to the associated maps.
+	//
+	// To summarise, a map points to a program and the program points to a map.
+	// The next for loop breaks this circular dependency and allows the GC to
+	// reclaim that memory.
+	for _, m := range s.Maps {
+		m.Prog = nil
+	}
```
kkourt: The Golang GC should be able to deal with circular references, e.g., see:https://groups.google.com/g/golang-nuts/c/tzvgeBEW1WY, so it seems to me that we are still missing something

kkourt: This looks suspicious: https://github.com/cilium/tetragon/blob/ea9b9b6f5b58cb814967098e861ffadd0846fc5e/pkg/sensors/sensors.go#L24-L28. Maybe we are not removing the programs from `allPrograms`?

tpapagian: It seems that we did not cleanup `AllMaps`. Fixed that and this PR is again ready for review.

olsajiri: nice, lgtm.. I already approved ;-) thanks

Issue #3795, opened 2025-06-02
configure klog


michi-covalent:
client-go uses klog (for example https://github.com/kubernetes/client-go/blob/master/tools/cache/reflector.go), which might look something like this:
```
I0602 20:40:03.292162       1 reflector.go:556] "Warning: watch ended with error" reflector="sigs.k8s.io/controller-runtime/pkg/cache/internal/informers.go:108" type="*v1alpha1.TracingPolicy" err="an error on the server (\"unable to decode an event from the watch stream: http2: client connection lost\") has prevented the request from succeeding"
```
this doesn't match tetragon's log format. it's probably better to configure klog so that the log format is consistent.

just as a reference cilium configures klog like this https://github.com/cilium/cilium/blob/9e5fb2eab53f2eb5974db8900ee46ac24a240c51/pkg/logging/logging.go#L70-L102

sayboras:
this might be a regression when I removed cilium/cilium dep in https://github.com/cilium/tetragon/pull/3767

PR #3814, opened 2025-06-09
slog: Migrate from logrus to slog

sayboras:
Fixes: #3795

### Changelog

```release-note
```


pkg/bugtool/multilog_test.go
```diff
@@ -5,42 +5,45 @@ package bugtool
...
 
 	multiLog := MultiLog{
-		Logs: []logrus.FieldLogger{log1, log2},
+		Logs: []logger.FieldLogger{log1, log2},
 	}
 
 	multiLog.Info("foo")
 	multiLog.Warn("bar")
-	expected1 := "level=info msg=foo\nlevel=warning msg=bar\n"
+	expected1 := "level=INFO msg=foo\nlevel=WARN msg=bar\n"
```
mtardy: just to understand, the format will very slightly change? Like the case here?

Maybe we can indicate the change in the changelog if people are relying on those logs

pkg/logger/slog.go
```diff
@@ -137,3 +132,50 @@ func (w logSink) WithValues(keysAndValues ...any) logr.LogSink {
 func (w logSink) WithName(name string) logr.LogSink {
 	return logSink{w.LogSink.WithName(name)}
 }
+
+type FieldLogger interface {
+	Handler() slog.Handler
+	With(args ...any) *slog.Logger
+	WithGroup(name string) *slog.Logger
```
mtardy, line 139: I was wondering, why not redefine the `WithError` to ease the change? I guess there was already a debate on this when they created slog :)

cmd/tetragon/conf.go
```diff
@@ -51,9 +53,9 @@ func readConfigSettings(defaultConfDir string, defaultConfDropIn string, dropIns
 		if configDir != "" {
 			err := option.ReadConfigDir(configDir)
 			if err != nil {
-				log.WithField(option.KeyConfigDir, configDir).WithError(err).Fatal("Failed to read config from directory")
+				logger.Fatal(log, "Failed to read config from directory", option.KeyConfigDir, configDir, logfields.Error, err)
 			} else {
-				log.WithField(option.KeyConfigDir, configDir).Info("Loaded config from directory")
+				log.Info("Loaded config from directory", option.KeyConfigDir, configDir)
```
mtardy, line 58: How does that work? I'm a little bit confused about the use of `logger.` and `log.`

sayboras: logger.Fatal is syntactic sugar for the below, as slog didn't have Fatal method.

```
logger.Error(msg, args...)
os.Exit(-1)
```

pkg/logger/slog.go
sayboras: FieldLogger is implemented by slog. Yes, slog didn't offer WithError function sadly.

pkg/bugtool/multilog_test.go
sayboras: ah this is mainly in test code, the actual logger is already having lower case for level. Will add a note in changelog for logger change (e.g. warning -> warn)

slog.go
```
	case slog.LevelKey:
		// Lower-case the log level
		return slog.Attr{
			Key:   a.Key,
			Value: slog.StringValue(strings.ToLower(a.Value.String())),
		}
```

cmd/tetragon/conf.go
kevsecurity: Yep, I'm also confused. When would we use ```log.``` and when would we use ```logger.```? i.e. why isn't it ```logger.Info()``` on line 58?

sayboras: Sadly, we don't have log.{Fatal,Panic,Trace} in slog, so that is why we need to use logger.{Fatal,Panic,Trace} utility functions.

Ideally, there will be not much change related to developer experience, as log.{Error,Warn,Info,Debug} are more common.

mtardy: ok! For some reason I thought you were writing your own interface and implementing it with slog under the hood from https://github.com/cilium/tetragon/pull/3814#discussion_r2149358385

kevsecurity (approved):
:rubber-stamp:

PR #3815, opened 2025-06-09
e2e: Don't install Cilium

michi-covalent:
E2E tests don't rely on Cilium.


mtardy:
we can't really tell from the logs because those are not verbose in case of success but we should indeed aim for that and the flag is correct

PR #3821, opened 2025-06-10
tetragon: matchBinaries followChildren fixes

olsajiri:
adding removal of mbset IDs and support for NotIn operator

### DIFF
--- a/contrib/tester-progs/Makefile
+++ b/contrib/tester-progs/Makefile
@@ -27,7 +27,8 @@ PROGS = sigkill-tester \
 	user-stacktrace \
 	test-helper \
 	pause \
-	raisesigkill
+	raisesigkill \
+	forks
 
 
 all: $(PROGS)
--- a/contrib/tester-progs/forks.c
+++ b/contrib/tester-progs/forks.c
@@ -1,0 +1,49 @@
+#include <unistd.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <limits.h>
+#include <sys/wait.h>
+#include <signal.h>
+
+void sig(int signo)
+{
+	kill(0, SIGTERM);
+	waitpid(-1, NULL, 0);
+}
+
+int main(int argc, char **argv)
+{
+	long i, count = 10;
+
+
+	if (argc == 2) {
+		count = strtol(argv[1], NULL, 10);
+		if (count == LONG_MIN)
+			return -1;
+	}
+
+	printf("forking %ld\n", count);
+
+	/* Set new process group so kill(0, ...) works properly. */
+	setpgid(0, 0);
+
+	/* We expect SIGINT to trigger the kill for all spawned processes. */
+	signal(SIGINT, sig);
+
+	for (i = 1; i < count; i++) {
+		pid_t p = fork();
+
+		if (p < 0) {
+			perror("fork");
+			return -1;
+		}
+
+		if (p == 0) {
+			pause();
+			exit(0);
+		}
+	}
+
+	pause();
+	return 0;
+}
--- a/pkg/sensors/tracing/matchbinaries_follow_children_test.go
+++ b/pkg/sensors/tracing/matchbinaries_follow_children_test.go
@@ -9,12 +9,15 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"path/filepath"
 	"syscall"
 	"testing"
+	"time"
 	"unsafe"
 
 	"golang.org/x/sys/unix"
 
+	"github.com/cilium/ebpf"
 	"github.com/cilium/tetragon/pkg/bpf"
 	"github.com/cilium/tetragon/pkg/btf"
 	"github.com/cilium/tetragon/pkg/config"
@@ -25,6 +28,7 @@ import (
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/reader/namespace"
 	"github.com/cilium/tetragon/pkg/reader/notify"
+	"github.com/cilium/tetragon/pkg/sensors/exec/execvemap"
 	testsensor "github.com/cilium/tetragon/pkg/sensors/test"
 	"github.com/cilium/tetragon/pkg/testutils"
 	tuo "github.com/cilium/tetragon/pkg/testutils/observer"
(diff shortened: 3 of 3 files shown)


pkg/testutils/sensors/sensors.go
```diff
@@ -29,7 +29,7 @@ func LoadSensor(t *testing.T, sensori sensors.SensorIface) {
 	}
 
 	t.Cleanup(func() {
-		sensor.Unload(true)
+		sensor.Destroy(true)
```
kkourt: Is this a bugfix we need to backport?

pkg/mbset/mbset.go
```diff
@@ -32,24 +32,28 @@ func openMap(name string) (*ebpf.Map, error) {
 }
 
 type state struct {
-	mu     sync.Mutex
-	nextID uint32
+	mu  sync.Mutex
+	ids map[uint32]struct{}
```
kkourt, line 44: Maybe name this `usedIDs` for clarity?

pkg/mbset/mbset.go
```diff
@@ -32,24 +32,28 @@ func openMap(name string) (*ebpf.Map, error) {
...
+	for id := range uint32(MaxIDs) {
+		if _, ok := s.ids[id]; !ok {
+			s.ids[id] = struct{}{}
+			return id, nil
+		}
 	}
-	ret := s.nextID
-	s.nextID++
-	return ret, nil
+	return 0, errors.New("cannot allocate new id")
```
kkourt: nit: Previous code returned InvalidID. Not sure if this was changed intentionally or not. No strong feelings either way.

pkg/mbset/mbset.go
```diff
@@ -56,6 +58,106 @@ func (s *state) AllocID() (uint32, error) {
...
+				return fmt.Errorf("failed to update mbset map: %w", err)
+			}
+		} else {
+			if err := mbsetMap.Delete(path); err != nil {
+				return fmt.Errorf("failed to remove mbset map: %w", err)
+			}
+		}
+	}
+
+	// 2) Clean up execve_map_val
```
kkourt, line 119: I see two potential issues here.

One is that as we are iterating the execve entries, new programs might be spawned and the bits will be propagated to execve entries that we will never update. This can be addressed by doing multiple iterations until the bit is nowhere to be found. We don't have to block the policy deletion for that, we can do it in a different goroutine and once we are done we can mark the id as available for allocation.

Second, Unless I'm mistaken, up until know we 've been only updating the execve map entries from the bpf side. Now, we are also updating it from user-space, and I do think that it's possible to have a race. To avoid the race, I believe we need to do the '&=` operation atomically (maybe using a spinlock on the execve entry that we would need to introduce).

Here are two examples of a race:

First:
* userspace reads the value from the  execve map
* process dies, PID is reclyed
* bpf program adds new execve map value for new pid
* userspace write the old value in the execve map


Second, when we update the execve map (e.g., in matchCapabilitiesChanges [^1]):

* userspace reads the value from the execve map
* mathCapabilitiesChanges selector matches, and caps in execve map are updated
* userspace writes the value, and above change is lost


[^1]: https://github.com/cilium/tetragon/blob/06365d5602242f30ff0de287901348cdf7a36cd2/bpf/process/generic_calls.h#L838-L849

kkourt (changes requested):
Thanks!

I have some concerns about data races. Please see comments below.

pkg/testutils/sensors/sensors.go
olsajiri: it matters for following code, I don't think it's a problem for current code

pkg/mbset/mbset.go
olsajiri: seems bit redundant to me, but no problem

pkg/mbset/mbset.go
olsajiri: we only check for the error, but I think it should return the InvalidID, thanks

pkg/mbset/mbset.go
olsajiri: > I see two potential issues here.
> 
> One is that as we are iterating the execve entries, new programs might be spawned and the bits will be propagated to execve entries that we will never update. This can be addressed by doing multiple iterations until the bit is nowhere to be found. We don't have to block the policy deletion for that, we can do it in a different goroutine and once we are done we can mark the id as available for allocation.

so there's protection against following race (I think you mean the same one) in patch https://github.com/cilium/tetragon/pull/3821/commits/e8d9da371b00f2064bda1124a4bdeac515d29407

we always remove single ID, and during that removal we mask this ID to be used in any exec operation

olsajiri: > Second, Unless I'm mistaken, up until know we 've been only updating the execve map entries from the bpf side. Now, we are also updating it from user-space, and I do think that it's possible to have a race. To avoid the race, I believe we need to do the '&=` operation atomically (maybe using a spinlock on the execve entry that we would need to introduce).
> 
> Here are two examples of a race:
> 
> First:
> 
>     * userspace reads the value from the  execve map
> 
>     * process dies, PID is reclyed
> 
>     * bpf program adds new execve map value for new pid
> 
>     * userspace write the old value in the execve map
> 
> 
> Second, when we update the execve map (e.g., in matchCapabilitiesChanges [1](#user-content-fn-1-61e03e8234928bb4bad995e8b4c34b82)):
> 
>     * userspace reads the value from the execve map
> 
>     * mathCapabilitiesChanges selector matches, and caps in execve map are updated
> 
>     * userspace writes the value, and above change is lost
> 
> 
> ## Footnotes
> 
>     1. https://github.com/cilium/tetragon/blob/06365d5602242f30ff0de287901348cdf7a36cd2/bpf/process/generic_calls.h#L838-L849 [↩](#user-content-fnref-1-61e03e8234928bb4bad995e8b4c34b82)

ok, that's a problem.. I wonder we could prepare the update in userspace and actualy do it in ebpf code like we do in UpdateStatsMap

or we try the idea with bitmask with timestamp and have fast but (datahas current timestamp) and slow path (data is obsolete and we walk the tasks to the parent)

kkourt: > so there's protection against following race (I think you mean the same one) in patch https://github.com/cilium/tetragon/commit/e8d9da371b00f2064bda1124a4bdeac515d29407

Yap. Missed the above patch, thanks!

mtardy: took a look at the PR (we discussed it this morning with Kornilios shortly), nothing more to add, looks great overall but indeed
> up until know we 've been only updating the execve map entries from the bpf side. Now, we are also updating it from user-space, and I do think that it's possible to have a race.

It seems tricky anytime when you do RW from both kernel and userspace.

pkg/mbset/mbset.go
kkourt: Maybe a comment then?

pkg/mbset/mbset.go
```diff
@@ -60,6 +63,101 @@ func (s *state) AllocID() (uint32, error) {
...
+	//
+	// During the removal we can hit execve sensor which could use the removed
+	// id, so we iterate all pids after the update to make sure the id is removed.
+	// In unlikely case case that we still find records with the removed id,
+	// we repeat (4 extra times) the removal after 0.5 second delay.
+	// In case we start to hit this race more often, we sohuld consider go
+	// routine for the removal.
```
kkourt, line 126: AFAICT, we will be holding the `s.mu` mutex from this point on (including) when we sleep.
Is this intentional?

It seems to me that we can release it and then grab it again only when we do `delete(s.ids, id)`.
We can have the execve map cleanup code in a another function as well. For example:
```
s.mu.Unlock()
err := cleanupExecveMap(...)
s.mu.Lock()
if err != nil {
   return fmt.Errorf("failed to remove mbset id %d", id)
}
delete(s.ids, id)
return nil
```

pkg/sensors/base/update.go
```diff
@@ -0,0 +1,92 @@
...
+		if err := upd.Map.MapHandle.Update(key, val, 0); err != nil {
+			return errors.New("failed to update data map value")
+		}
+
+		ret, err := prog.Run(&opts)
+
+		// executed, but failed in bpf program above
+		if err == nil && ret != 0 {
+			return errors.New("failed to update map value")
+		}
```
kkourt, line 64: Do we need to check for `err != nil`?

pkg/selectors/kernel.go
```diff
@@ -1220,9 +1220,6 @@ func ParseMatchBinary(k *KernelSelectorState, b *v1alpha1.BinarySelector, selIdx
 	sel.Op = op
 	sel.MBSetID = mbset.InvalidID
 	if b.FollowChildren {
-		if op != SelectorOpIn {
```
kkourt, line 1231: Should this be `op != SelectorOpIn && op != SelectorOpNotIn`?

pkg/mbset/mbset.go
olsajiri: yes, I don't think we need to hold it during the cleanup, will change

pkg/sensors/base/update.go
olsajiri: we do, will add, thanks

pkg/selectors/kernel.go
olsajiri: ah we do have more operators :-D will change, thanks

bpf/process/bpf_execve_map_update.c
```diff
@@ -0,0 +1,94 @@
...
+
+__attribute__((section("seccomp"), used)) int
+execve_map_update(void *ctx)
+{
+	struct update_data *data;
+	__u32 idx = 0;
+
+	data = map_lookup_elem(&execve_map_update_data, &idx);
+	if (!data)
+		return 0;
```
kkourt: Should we return an error here so that it's captured by:
```
	ret, err := prog.Run(&opts)

	// executed, but failed in bpf program above
	if err != nil || ret != 0 {
```

bpf/process/types/basic.h
```diff
@@ -1506,6 +1506,12 @@ FUNC_INLINE int match_binaries(__u32 selidx, struct execve_map_value *current)
 				return 1;
 			fallthrough;
 		case op_filter_notin:
+			if (selector_options->op == op_filter_notin) {
+				if (selector_options->mbset_id != MBSET_INVALID_ID &&
+				    (current->bin.mb_bitset & (1UL << selector_options->mbset_id)))
+					return 0;
+			}
+
```
kkourt: should we add the `selector_options->mbset_id != MBSET_INVALID_ID` check the `op_filter_in` as well?

Assuming above is correct, would something like

```
		switch (selector_options->op) {
		case op_filter_in:
		case op_filter_notin:
			/* Check if we match the selector's bit in ->mb_bitset, which means that the
			 * process matches a matchBinaries section with a followChidren:true
			 * attribute either because the binary matches or because the binary of a
			 * parent matched.
			 */
			if (selector_options->mbset_id != MBSET_INVALID_ID &&
			    current->bin.mb_bitset & (1UL << selector_options->mbset_id)) {
				found_key = 0xc0ffee; /* invalid pointer, but we just check for NULL */
				break;
			}
```

work?

kkourt (commented):
Thanks!

bpf/process/bpf_execve_map_update.c
olsajiri: yes, that should never fail, but we need to return -1 to be consistent with user space, thanks

bpf/process/types/basic.h
olsajiri: yes, that works, and saves some instructions, nice, thanks

PR #3822, opened 2025-06-11
tetragon: assorted fixes

olsajiri:
assorted fixes, please see changelogs


pkg/sensors/tracing/generictracepoint.go
```diff
@@ -70,15 +66,14 @@ func init() {
 
 // genericTracepoint is the internal representation of a tracepoint
 type genericTracepoint struct {
+	tableId idtable.EntryID
```
mtardy, line 69: nit for coherence
```suggestion
	tableID idtable.EntryID
```

mtardy (approved):

pkg/sensors/tracing/generictracepoint.go
olsajiri: hum, seems like we use tableId everywhere:

```
pkg/sensors/tracing/genericlsm.go:      tableId   idtable.EntryID
pkg/sensors/tracing/generickprobe.go:   tableId idtable.EntryID
pkg/sensors/tracing/generictracepoint.go:       tableId idtable.EntryID
pkg/sensors/tracing/genericuprobe.go:   tableId      idtable.EntryID
```

olsajiri: ok, I'll spin a follow up if we decide we want that change

PR #3827, opened 2025-06-13
procevents: fix in-init-tree

will-isovalent:
8073fb2df erroneously removed the logic for handling inInitTree from proc events. Add that logic back here.

```release-note
Fix an issue where inInitTree was not properly accounting processes started before Tetragon.
```


kkourt:
Does this need to be backported?

kkourt (approved):

kkourt:
@will-isovalent merging this.  Should we do backports?

PR #3832, opened 2025-06-17
tetragon/windows: Support multiple programs from a single collection

ExceptionalHandler:
### Description
This PR changes the way programs are loaded on Windows in order to support loading multiple programs from a single collection. Earlier implementation assumed one program per collection.
In Windows, a collection can be loaded only once per process. In that light, if a sensor had two programs from the same sys file, the loading of the second program would fail, as loading the collection second time around fails.

In the PR, we cache loaded collections by path and check the cache before loading a new collection. The second run will use the cached collection object and load the passed in program.

### DIFF
--- a/pkg/bpf/coll_windows.go
+++ b/pkg/bpf/coll_windows.go
@@ -9,25 +9,33 @@ import (
 	"github.com/cilium/ebpf"
 )
 
-type CollectionStore struct {
-	collMap map[string]*ebpf.Collection
-}
-
 var (
-	store          CollectionStore
-	collstoreMutex sync.RWMutex
+	collectionByName map[string]*ebpf.Collection
+	collectionByPath map[string]*ebpf.Collection
+	collstoreMutex   sync.RWMutex
 )
 
-func SetCollection(name string, coll *ebpf.Collection) {
+func SetCollection(name string, path string, coll *ebpf.Collection) {
 	collstoreMutex.Lock()
-	store.collMap[name] = coll
-	collstoreMutex.Unlock()
+	defer collstoreMutex.Unlock()
+	collectionByName[name] = coll
+	collectionByPath[path] = coll
+}
+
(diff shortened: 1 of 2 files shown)


pkg/bpf/coll_windows.go
```diff
@@ -9,24 +9,32 @@ import (
...
 )
 
-type CollectionStore struct {
-	collMap map[string]*ebpf.Collection
-}
-
 var (
-	store          CollectionStore
+	store          map[string]*ebpf.Collection
+	pathStore      map[string]*ebpf.Collection
```
olsajiri: these variables could mimic the GetXXX functions, like:
```
  collectionByName map[string]*ebpf.Collection
  collectionByPath map[string]*ebpf.Collection

File: pkg/bpf/coll_windows.go [hunk 2]
```diff
@@ -9,24 +9,32 @@ import (
...
+func SetCollection(name string, path string, coll *ebpf.Collection) {
 	collstoreMutex.Lock()
-	store.collMap[name] = coll
+	store[name] = coll
+	pathStore[path] = coll
 	collstoreMutex.Unlock()
 }
 
+func GetCollectionByPath(path string) (*ebpf.Collection, error) {
+	collstoreMutex.RLock()
```
olsajiri, line 26: you could use the standard scheme:
```
  collstoreMutex.RLock()
  defer collstoreMutex.RUnlock()
  ...

same above
```

olsajiri (approved):
lgtm, left some comments

pkg/bpf/coll_windows.go
ExceptionalHandler: Done

pkg/bpf/coll_windows.go
ExceptionalHandler: Done

PR #3833, opened 2025-06-18
Use Go 1.19 atomic types

tklauser:
The Go standard library provides types for atomic access since Go 1.19, see https://go.dev/doc/go1.19#atomic_types and
https://pkg.go.dev/sync/atomic#pkg-types.

These make working with atomically accessed variables type-safe and more ergonomic.

This change also allows to get rid of the `go.uber.org/atomic` dependency.


.golangci.yml
```diff
@@ -55,6 +55,10 @@ linters:
             recommendations:
             - github.com/cilium/tetragon
             reason: cilium/cilium package might be overkilled, consider other alternatives
+        - go.uber.org/atomic
```
mtardy: surprisingly there's a colon missing here? didn't know it was the syntax hehe
```suggestion
        - go.uber.org/atomic:
```

tklauser: Thanks, fixed!

tklauser:
> I love the that we are removing those deps in favor of the stdlib! Next victim is `go.uber.org/multierr` maybe! :)

Yeah, multierr should be easy enough. I'll have a look :eyes:

mtardy:
> > I love the that we are removing those deps in favor of the stdlib! Next victim is `go.uber.org/multierr` maybe! :)
> 
> Yeah, multierr should be easy enough. I'll have a look 👀

I think the only downside is that errors don't exactly look the same, I think we took a look at it in the past and there was something like one was going to the line and not the other when logging or something...

mtardy (approved):
I love it! Thank you!

PR #3842, opened 2025-06-22
RFC tetragon: Do not rate limit exit events

olsajiri:
As reported by Balasaheb we could pollute process cache by rate limiting exit events.

It makes more sense to let all possible exit events go through, because they won't eat much resources and eventually release process cache records.

Otherwise we could end up with hanging process cache records.

Reported-by: Balasaheb Salunke ...

### DIFF
--- a/bpf/process/bpf_exit.h
+++ b/bpf/process/bpf_exit.h
@@ -69,14 +69,8 @@ FUNC_INLINE void event_exit_send(void *ctx, __u32 tgid)
 		probe_read(&exit->info.code, sizeof(exit->info.code),
 			   _(&task->exit_code));
 
-#ifndef __RHEL7_BPF_PROG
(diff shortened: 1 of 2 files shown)


olsajiri:
so with this change we might get burst of exit events, but keep process cache without stale processes,
or should we perhaps do periodical check if processes in cache are alive? (if we do not do that already,
I might have missed that ;-) @tpapagian  )

mtardy (approved):
Makes sense to process all exit events indeed! We do have a GC for the process cache but ended up not merging the cleanups proposed recently by https://github.com/cilium/tetragon/pull/2953.

tpapagian:
Yes, this sound reasonable to me as well. Even if we had a GC for the process cache, I would say that it is better to have those correct than relying on the GC to do the job as this would reveal possible bugs.

PR #3845, opened 2025-06-23
WIP

tixxdz:
Fixes 

### Changelog

```release-note
```


pkg/selectors/kernel.go
```diff
@@ -1195,6 +1209,130 @@
...
+
+	op, err := SelectorOp(uids.Operator)
+	if err != nil {
+		return fmt.Errorf("error: operator %w", err)
+	}
+
+	if (op != SelectorOpEQ) && (op != SelectorOpNEQ) {
+		return errors.New("error: operator supports only 'eq' and 'neq' operators")
+	}
+
```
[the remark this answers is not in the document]
tixxdz: ok thanks copilot the previous and current checks are enough

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -185,6 +188,24 @@ type CapabilitiesSelector struct {
 	Values []string `json:"values"`
 }
 
+type CredIdValues struct {
+	// +kubebuilder:validation:Enum=eq;neq;Equal;NotEqual
```
kkourt: Please don't add new operators. AFAICT, we don't use `eq` or `neq` anywhere else.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -185,6 +188,24 @@ type CapabilitiesSelector struct {
 	Values []string `json:"values"`
 }
 
+type CredIdValues struct {
```
kkourt: To make @mtardy's life easier for merging linters (https://github.com/cilium/tetragon/pull/3850), I think it's a good idea to respect go conventions, so please use `CredIDValues` instead. See: https://go.dev/wiki/CodeReviewComments#initialisms.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -185,6 +188,24 @@ type CapabilitiesSelector struct {
...
+	Operator string `json:"operator"`
+	// The UID or GID values to match
+	// To match against a range use 'min:max'. Supported values are 0:4294967295.
+	Values []string `json:"values,omitempty"`
+}
+
+type CredentialsSelector struct {
+	// +kubebuilder:validation:Optional
+	// Real UIDs to match.
+	Ruid []CredIdValues `json:"uid,omitempty"`
```
kkourt: Same as above. We can use `RealUIDs` or `RUIDs`. I prefer the former.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -185,6 +188,24 @@ type CapabilitiesSelector struct {
...
+	// +kubebuilder:validation:Optional
+	// Effective UIDs to match.
+	Euid []CredIdValues `json:"euid,omitempty"`
```
kkourt: Same as above, we can use `EffectiveUIDs` or `EUIDs`.

bpf/process/pfilter.h
```diff
@@ -254,15 +389,23 @@ process_filter_capabilities(__u32 ty, __u32 op, __u32 ns, __u64 val,
...
+	u32 ty; /* (i.e. effective, inheritable, or permitted) */
+	u32 op; /* op (i.e. op_filter_in or op_filter_notin) */
+	u32 ns; /* If ns == 0 <=> IsNamespaceCapability == false. Otheriwse it contains the value of host user namespace. */
+	u64 val; /* OR-ed capability values */
+} __attribute__((packed));
+
 #ifdef __CAP_CHANGES_FILTER
 FUNC_INLINE int
-process_filter_capability_change(__u32 ty, __u32 op, __u32 ns, __u64 val,
+process_filter_capability_change(struct caps_filter *caps,
```
kkourt: These changes seem unrelated to this patch which, according to the commit message, is about 'matchCurrentCred`. If you want to include them, please add a separate patch with a proper commit message.

pkg/selectors/kernel.go
```diff
@@ -1263,7 +1263,7 @@ func credUidSelectorValues(uids *v1alpha1.CredIdValues) ([]byte, uint32, error)
 			return nil, 0, fmt.Errorf("uids/gids values outside of range 0:%d", math.MaxUint32)
 		}
 		if minId > maxId {
-			minId = maxId
+			return nil, 0, fmt.Errorf("uids/gids values %q invalid: range should be 'min:max'", v)
```
kkourt: I believe this was introduced in a previous commit in this PR. Please fold the change in the original commit.

pkg/selectors/kernel_test.go
```diff
@@ -457,6 +457,198 @@ func TestParseMatchNamespaces(t *testing.T) {
 	}
 }
 
+func TestParseMatchCurrentCred(t *testing.T) {
```
kkourt: This test is hard to read.
I would suggest an array of test cases, each with the filter and the expected result to make it easier to read and understand.

kkourt (commented):
Thanks!

pkg/k8s/apis/cilium.io/v1alpha1/types.go
tixxdz: Sure, thanks

bpf/process/pfilter.h
```diff
@@ -114,6 +114,141 @@ struct selector_filter {
...
+		*index += sizeof(struct cred_id);
+		/* now index points at the end of cred_filter filter */
+	}
+
+	return ret;
+}
+
+FUNC_INLINE int
+process_filter_current_cred(__u32 *f, __u32 *index, __u32 len, struct msg_generic_kprobe *msg)
+{
```
olsajiri: msg argument is not needed?

bpf/process/pfilter.h
```diff
@@ -513,6 +650,27 @@ selector_process_filter(__u32 *f, __u32 index, struct execve_map_value *enter,
...
+	 *
+	 * (4 (length) + 4 (flags) + sizeof(ruid) + sizeof(rgid) + ...
+	 */
+	len = *(__u32 *)((__u64)f + (index & INDEX_MASK));
+	index += 4; /* 4: matchCurrentCred header */
+
+#ifdef __LARGE_BPF_PROG
+	/* Filter current task credentials */
+	if (len > 4) {
+		len -= 4;
```
olsajiri: should we follow the rest of the selectors and have the len -= 4 before the if condition and the condition have if (len > 0) ?

bpf/process/pfilter.h
```diff
@@ -114,6 +114,141 @@ struct selector_filter {
...
+	else if (cred_f->op == op_filter_neq && (uid < min || uid > max))
+		return PFILTER_ACCEPT;
+
+	return PFILTER_REJECT;
+}
+
+FUNC_INLINE int
+match_cred_uids(__u32 *f, __u32 *index, struct cred_filter *cred_f, __u32 uid)
+{
+	int i, ret = 0;
```
olsajiri: let's init ret with PFILTER_REJECT?

bpf/process/pfilter.h
```diff
@@ -114,6 +114,141 @@ struct selector_filter {
...
+		/* now index points at the end of cred_filter filter */
+	}
+
+	return ret;
+}
+
+FUNC_INLINE int
+process_filter_current_cred(__u32 *f, __u32 *index, __u32 len, struct msg_generic_kprobe *msg)
+{
+	__u32 flags, ret = 1, uid;
```
olsajiri: let's init ret with PFILTER_ACCEPT?

bpf/process/pfilter.h
```diff
@@ -114,6 +114,141 @@ struct selector_filter {
...
+
+	if (flags & PF_CRED_RUID) {
+		cred_f = (struct cred_filter *)((u64)f + (*index & INDEX_MASK));
+		/* 16: type (8 bytes), op (4 bytes), number of values (4 bytes) */
+		*index += sizeof(struct cred_filter);
+
+		if (cred_f->num > MAX_CRED_SELECTOR_VALUES)
+			return PFILTER_ERROR;
+		uid = BPF_CORE_READ(task, cred, uid.val);
+		ret &= match_cred_uids(f, index, cred_f, uid);
```
olsajiri: do you need &= in here? looks like simple assignment would work

bpf/process/pfilter.h
```diff
@@ -114,6 +114,141 @@ struct selector_filter {
...
+
+	if (flags & PF_CRED_EUID) {
+		cred_f = (struct cred_filter *)((u64)f + (*index & INDEX_MASK));
+		/* 16: type (8 bytes), op (4 bytes), number of values (4 bytes) */
+		*index += sizeof(struct cred_filter);
+
+		if (cred_f->num > MAX_CRED_SELECTOR_VALUES)
+			return PFILTER_ERROR;
+		uid = BPF_CORE_READ(task, cred, euid.val);
+		ret &= match_cred_uids(f, index, cred_f, uid);
```
olsajiri: do you need &= in here? looks like simple assignment would work

bpf/process/pfilter.h
```diff
@@ -114,6 +114,141 @@ struct selector_filter {
...
+
+	flags = *(__u32 *)((__u64)f + (*index & INDEX_MASK));
+	*index += 4; /* 4: matchCurrentCred flags */
+	/* Are there any valid credentials filters? */
+	if (!flags)
+		return 0;
+
+	/* Here we are at the beginning of matching credentials */
+
+	if (flags & PF_CRED_RUID) {
```
olsajiri: it seems like the code below this condition and the code under PF_CRED_EUID could be in the function that has the uid argument?

pkg/selectors/kernel_test.go
tixxdz: Sure, thanks

bpf/process/pfilter.h
tixxdz: I removed it, the code was failing with stack limits but I simplified the new code...

bpf/process/pfilter.h
tixxdz: Yes makes sense, thank you

bpf/process/pfilter.h
tixxdz: sure

bpf/process/pfilter.h
tixxdz: ok removed

bpf/process/pfilter.h
tixxdz: I ended up with PFILTER_REJECT and adapted the code below

bpf/process/pfilter.h
tixxdz: Ok I tried but it propagated PFILTER_ERROR inside, let me know what you think?  since we still need to handle in that generic calls the PFILTER_ERROR, IIRC right now PFILTER_ERROR could be PFILTER_ACCEPT in top callers.

tixxdz:
> Thanks!
> 
> Please find some comments below.

Much appreciated for the review, all comments handled let's see if tests pass.

bpf/process/pfilter.h
tixxdz: @olsajiri ok pushed let's see if tests pass, but also what you think about above

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -185,6 +188,24 @@ type CapabilitiesSelector struct {
...
+	Operator string `json:"operator"`
+	// The UID or GID values to match
+	// To match against a range use 'min:max'. Supported values are 0:4294967295.
+	Values []string `json:"values,omitempty"`
+}
+
+type CredentialsSelector struct {
+	// +kubebuilder:validation:Optional
+	// Real UIDs to match.
+	RUIDs []CredIDValues `json:"uid,omitempty"`
```
kkourt: I'm not sure what's the point of this being a list.


For example, if a user  specifies:
```
 - uid:
   - operator: Equal
      values:
       - "10"
       - "30"
   - operator: Equal
      values:
       - "20"
```

Then this either means: (uid == 10 OR uid == 30) AND (uid == 20) which will never match if the values are different.

Or it means (uid == 10 OR uid == 30) OR (uid == 20) which can just be written as a single selector: 

```
 - uid:
   - operator: Equal
      values:
       - "10"
       - "20"
       - "30"
   ```
   
   Also, is there a reason why the JSON field different? `uid` vs `RUIDs`?

File: pkg/k8s/apis/cilium.io/v1alpha1/types.go [hunk 17]
```diff
@@ -185,6 +188,24 @@ type CapabilitiesSelector struct {
...
+	// +kubebuilder:validation:Optional
+	// Effective UIDs to match.
+	EUIDs []CredIDValues `json:"euid,omitempty"`
```
kkourt: same as above

bpf/process/pfilter.h
```diff
@@ -114,6 +114,162 @@ struct selector_filter {
...
+	PF_CRED_EGID = 0x08, // Effective GIDs
+};
+
+struct cred_id {
+	__u32 start;
+	__u32 end;
+} __attribute__((packed));
+
+struct cred_filter {
+	__u64 ty; /* credentials type flags */
```
kkourt: What are the values of these flags?

Are they used somewhere?
The only thing I'm seeing is:

```
	if (!cred_f->ty)
		return PFILTER_REJECT;
```

so it seems to me that this can be omitted.
If they are used at a later patch, please introduce them there.

pkg/sensors/tracing/kprobe_process_credentials_test.go
```diff
@@ -337,3 +346,129 @@ spec:
...
+	tus.LoadInitialSensor(t)
+	tus.LoadSensor(t, testsensor.GetTestSensor())
+	sm := tuo.GetTestSensorManager(t)
+
+	ruid := []v1alpha1.CredIDValues{
+		{
+			Operator: "Equal",
+			Values:   []string{"0:0"},
+		},
+	}
```
kkourt: This test is hard to read.
I would suggest an array of test cases, each with the filter and the expected result to make it easier to read and understand.

pkg/selectors/kernel_test.go
```diff
@@ -457,6 +458,275 @@ func TestParseMatchNamespaces(t *testing.T) {
...
+		err := ParseMatchCurrentCred(k, credOldKernel)
+		require.Error(t, err)
+		return
+	}
+
+	tests := map[string]struct {
+		cred     []v1alpha1.CredentialsSelector
+		expected []byte
+		err      error
+	}{
```
kkourt: Thanks!

pkg/selectors/kernel_test.go
```diff
@@ -457,6 +458,275 @@ func TestParseMatchNamespaces(t *testing.T) {
...
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+				require.True(t, bytes.Equal(test.expected, d.e[0:d.off]))
+			}
+
+		})
+	}
+
+	/* Extra tests */
```
kkourt: Can you please use a similar structure as above? That is, something like:

```
tests := map[string]struct {
		cred     []v1alpha1.CredentialsSelector
		expected []byte
		err      error
```

pkg/sensors/tracing/kprobe_process_credentials_test.go
```diff
@@ -180,6 +181,11 @@ spec:
 		t.Fatalf("setgid(%d) error: %s", gid+1, err)
 	}
 
+	// Restore all gids to 0
+	if err := syscall.Setgid(0); err != nil {
+		t.Fatalf("setgid(0) error: %s", err)
+	}
```
kkourt, line 181: Is this a fix for an existing test? Please have it as a separate patch.

pkg/sensors/tracing/kprobe_process_credentials_test.go
```diff
@@ -342,6 +348,11 @@ spec:
...
```
kkourt, line 348: Similar as above: if this is a fix on an existing test, please add it as a separate patch.

kkourt (commented):
Thanks!

pkg/k8s/apis/cilium.io/v1alpha1/types.go
tixxdz: Hmm I was thinking an AND , I intended to disable it for now and open it later. To combine Equal and NotEqual

```
- uid:
   - operator: Equal
      values:
      -  "0:999" # system users
   - operator: NotEqual
      values:
      - "102"   # syslog
      - "995:998"  # systemd tools on ubuntu
```

Or let's say   Equal  "1000:65535"    users but NotEqual "2000" .  Of course we can make it in single Equal and use ranges but it is also easier if you use one or two single ranges, and exclude the special ones, so some flexibility for the future, especially if you have cases running with UIDs ranges...

I already check if we have already parsed uid field, let me add another check to ensure we have only one operator, and we can activate later. Will add code comments for it.


> Also, is there a reason why the JSON field different? uid vs RUIDs?

To be consistent with https://tetragon.io/docs/reference/grpc-api/#processcredentials

pkg/sensors/tracing/kprobe_process_credentials_test.go
tixxdz: Sure will do, missed it.

bpf/process/pfilter.h
tixxdz: Good catch, it should be used since we can't predict the order of passed credentials filters uid,euid,x,y or x,uid,y,euid...  in the tracing policy, so every time we iterate to the next one we should check it to know which field we are parsing... I intended to use it for that, will fix it, thanks!

pkg/selectors/kernel_test.go
tixxdz: Will try otherwise I separate them in different test, since they are a bit special...

pkg/sensors/tracing/kprobe_process_credentials_test.go
tixxdz: ok thanks

pkg/k8s/apis/cilium.io/v1alpha1/types.go
kkourt: > Or let's say Equal "1000:65535" users but NotEqual "2000"

OK, above makes sense to me, thanks!. Let's document that it's an AND,  add a test-case for it.

> To be consistent with https://tetragon.io/docs/reference/grpc-api/#processcredentials

They are consistent at the JSON level, but not at the gRPC level. I would suggest to name the JSON field `ruid` to make the mapping between gRPC and JSON fields obvious as we do in all other cases.

pkg/sensors/tracing/kprobe_process_credentials_test.go
tixxdz: done

pkg/k8s/apis/cilium.io/v1alpha1/types.go
tixxdz: > OK, above makes sense to me, thanks!. Let's document that it's an AND, add a test-case for it.

I just added a comment that in future we plan to do Equal and NotEqual , and added checks and test cases for now.

> They are consistent at the JSON level, but not at the gRPC level. I would suggest to name the JSON field ruid to make the mapping between gRPC and JSON fields obvious as we do in all other cases.

I see, I opted for UIDs and uid  then to be consistent with what we have and current kernel declaration.

bpf/process/pfilter.h
tixxdz: Ok fixed, now we don't expect any order. Thanks

tixxdz:
ok seems all tests are green now.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -185,6 +188,24 @@ type CapabilitiesSelector struct {
...
+	Values []string `json:"values,omitempty"`
+}
+
+type CredentialsSelector struct {
+	// +kubebuilder:validation:Optional
+	// Real UIDs to match.
+	UIDs []CredIDValues `json:"uid,omitempty"`
+	// +kubebuilder:validation:Optional
+	// Effective UIDs to match.
+	EUIDs []CredIDValues `json:"euid,omitempty"`
```
kkourt: Can  you document that these are AND?
(As discussed in https://github.com/cilium/tetragon/pull/3845#discussion_r2221556703)

tixxdz: done

tixxdz:
rebased.

tixxdz:
CLosing this since we merged the current resolve features.

kkourt:
Adding current PR for reference: ~https://github.com/cilium/tetragon/pull/4012~ https://github.com/cilium/tetragon/pull/4064

PR #3848, opened 2025-06-24
btf: skip arg validation if resolve is set

kkourt:
If a resolve path is set, skip argument validation.

It would be nice if we could check the type _after_ the path was resolved, but this requires more changes and is left as a followup.

### DIFF
--- a/pkg/btf/validation.go
+++ b/pkg/btf/validation.go
@@ -160,6 +160,13 @@ func ValidateKprobeSpec(bspec *btf.Spec, call string, kspec *v1alpha1.KProbeSpec
 		if specArg.Index >= fnNArgs {
 			return fmt.Errorf("kprobe arg %d has an invalid index: %d based on prototype: %s", i, specArg.Index, proto)
 		}
+
+		// If there is a resolve path defined by the user, we need to traverse it to match
+		// the types. This will happen later (e.g., see resolveBTFArg) so skip the check
+		// here.
+		if specArg.Resolve != "" {
+			continue
+		}
 		arg := proto.Params[int(specArg.Index)]
 		paramTyStr := getKernelType(arg.Type)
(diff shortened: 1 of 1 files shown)


kkourt:
CC: @tdaudi

michi-covalent (approved):
😇

tdaudi:
This is good, I like the idea of pre validation, so I took a closer look.

I think maybe we could add a field in `kpValidateInfo` to contain all BTF types for each call. So we resolve the attributes in this `ValidateKprobeSpec` and when we build `EventConfig` later we have minimal job to do.
https://github.com/cilium/tetragon/blob/bdb1ab277a11f4da1827c5806dbf4e42a7447b99/pkg/sensors/tracing/generickprobe.go#L417-L421

So instead of having
```go
err := btf.ValidateKprobeSpec(btfobj, call, f, ks)
```

Maybe we could have
```go
btfTypes, err := btf.ValidateKprobeSpec(btfobj, call, f, ks)

...

return &kpValidateInfo{
		calls:  &kpCall {
		    Name: name,
		    BTFTypes: btfTypes,
		},
...
	}, nil
```

I also like the idea of separating the logic between the "BTF logic" and `EventConfig` setup. But this is just my personal opinion. What do you think ?

kkourt:
> I also like the idea of separating the logic between the "BTF logic" and `EventConfig` setup. But this is just my personal opinion. What do you think ?


So if I understand correctly, the high level idea is to parse the needed resolve information in the validation step (and also validate types), and then pass it long when we load the sensor to properly build the corresponding map? That makes a lot of sense to me (but haven't checked the specifics in the code).

tdaudi:
Yes, exactly !
Unfortunately, this is not done similarly for other sensors. But think making a `Validation` interface for all sensors that do all the BTF checks can be great (and also centralize all the errors messages returned from different BTF configs)

PR #3850, opened 2025-06-24
linters/staticcheck: add checks for naming convention

### LINKED ISSUE #3490: Lint capitalization of acronyms in Go code
Tetragon Go code uses a mix of names with various case for acronyms like LSM or BPF. Following the [Go style guide](https://google.github.io/styleguide/go/decisions#initialisms) they should be always capitalized.

We can enforce it by adding a linter to the CI static checks. I don't know of a linter allowing to configure custom acronyms, so if a third-party doesn't exist we can write a custom one using go/analysis framework.

mtardy:
Fixes #3490.

https://google.github.io/styleguide/go/decisions#initialisms
https://go.dev/wiki/CodeReviewComments#initialisms


mtardy:
This is too opinionated, we decided to close this and issue https://github.com/cilium/tetragon/issues/3490.

This might be worthwhile https://github.com/cilium/tetragon/pull/3983

PR #3852, opened 2025-06-25
fix mask issue for capability types

kkourt:
Fix an issue where capabilities types in argument filter selectors are ignored.
Also, add some convenience parsing for capability types.

### DIFF
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -604,7 +604,7 @@ func parseAddr(v string) ([]byte, uint32, error) {
 func capsStrToUint64(values []string) (uint64, error) {
 	caps := uint64(0)
 	for _, v := range values {
-		valstr := strings.ToUpper(v)
+		valstr := strings.ToUpper(strings.TrimSpace(v))
 		c, ok := tetragon.CapabilitiesType_value[valstr]
 		if !ok {
 			return 0, fmt.Errorf("value %s unknown", valstr)
@@ -646,11 +646,11 @@ func writeMatchValues(k *KernelSelectorState, values []string, ty, op uint32) er
 		case gt.GenericSockType, gt.GenericSkbType, gt.GenericSockaddrType, gt.GenericSocketType, gt.GenericNetDev:
 			return fmt.Errorf("MatchArgs type sock, socket, skb, sockaddr and net_device do not support operator %s", selectorOpStringTable[op])
 		case gt.GenericKernelCap, gt.GenericCapInheritable, gt.GenericCapPermitted, gt.GenericCapEffective:
-			i, err := strconv.ParseUint(v, base, 64)
+			mask, err := parseCapabilitiesMask(v)
 			if err != nil {
-				return fmt.Errorf("MatchArgs value %s invalid: %w", v, err)
+				return fmt.Errorf("MatchArgs capabilities mask value %s invalid: %w", v, err)
 			}
-			WriteSelectorUint64(&k.data, uint64(i))
+			WriteSelectorUint64(&k.data, uint64(mask))
 		default:
 			return fmt.Errorf("MatchArgs type %s unsupported", gt.GenericTypeString(int(ty)))
 		}
@@ -1429,4 +1429,19 @@ func HasFilter(selectors []v1alpha1.KProbeSelector, index uint32) bool {
 		}
 	}
 	return false
+}
+
+// parseCapabilitiesMask create a capabilities mask
+func parseCapabilitiesMask(s string) (uint64, error) {
+	base := getBase(s)
+	mask, err := strconv.ParseUint(s, base, 64)
+	if err == nil {
+		return mask, nil
+	}
+	caps := strings.Split(s, ",")
+	mask, err = capsStrToUint64(caps)
+	if err != nil {
+		return uint64(0), fmt.Errorf("parseCapabilitiesMask: %w", err)
+	}
+	return mask, nil
 }
--- a/pkg/selectors/kernel_test.go
+++ b/pkg/selectors/kernel_test.go
@@ -15,6 +15,7 @@ import (
 	"testing"
 
 	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/cilium/tetragon/pkg/config"
@@ -1093,4 +1094,26 @@ func TestParseAddr(t *testing.T) {
 			}
 		})
 	}
+}
+
+func TestParseCapabilityMask(t *testing.T) {
+	v, err := parseCapabilitiesMask("100")
+	require.NoError(t, err)
+	assert.Equal(t, uint64(100), v)
+
+	v, err = parseCapabilitiesMask("CAP_SYS_ADMIN")
+	require.NoError(t, err)
+	assert.Equal(t, (uint64(1) << 21), v)
+
+	v, err = parseCapabilitiesMask("CAP_SYS_ADMIN,CAP_BPF")
+	require.NoError(t, err)
+	assert.Equal(t, (uint64(1)<<21)|(uint64(1)<<39), v)
+
+	// NB: spaces should be OK
+	v, err = parseCapabilitiesMask("CAP_SYS_ADMIN, CAP_BPF")
+	require.NoError(t, err)
+	assert.Equal(t, (uint64(1)<<21)|(uint64(1)<<39), v)
+
+	_, err = parseCapabilitiesMask("CAP_PIZZA")
+	assert.Error(t, err)
 }


pkg/selectors/kernel.go
```diff
@@ -1416,3 +1422,24 @@ func HasFilter(selectors []v1alpha1.KProbeSelector, index uint32) bool {
...
+	mask = uint64(0)
+	caps := strings.Split(s, ",")
+	for _, cap := range caps {
+		capstr := strings.ToUpper(cap)
+		capval, ok := tetragon.CapabilitiesType_value[capstr]
+		if !ok {
+			return uint64(0), fmt.Errorf("parseCapabilitiesMask: value %s unknown", capstr)
+		}
+		mask |= (1 << capval)
+	}
```
tdaudi: This seems to be a redundant code with values parsing for matchCapabilities.


https://github.com/cilium/tetragon/blob/b06bec3936609ad11ee3e93f30785a12b72349ec/pkg/selectors/kernel.go#L1163-L1170

I pushed a commit in https://github.com/cilium/tetragon/pull/3772 to move this code to a function (https://github.com/cilium/tetragon/pull/3772/commits/94cd499697dd5088133d91f2d564b1b01428ae6d).

pkg/selectors/kernel.go
```diff
@@ -632,8 +632,14 @@ func writeMatchValues(k *KernelSelectorState, values []string, ty, op uint32) er
 			WriteSelectorUint64(&k.data, uint64(i))
 		case gt.GenericSockType, gt.GenericSkbType, gt.GenericSockaddrType, gt.GenericSocketType, gt.GenericNetDev:
 			return fmt.Errorf("MatchArgs type sock, socket, skb, sockaddr and net_device do not support operator %s", selectorOpStringTable[op])
-		case gt.GenericCharIovec:
-			return fmt.Errorf("MatchArgs values %s unsupported", v)
+		case gt.GenericKernelCap, gt.GenericCapInheritable, gt.GenericCapPermitted, gt.GenericCapEffective:
```
tdaudi, line 648: The only downside I see doing this is if the hook is at some point where the capability is not yet typed with `kernel_cap_t`, the user will still have to write `type: kernel_cap_t` even though it is a `u64` type. But I can't say if it ever happens.

pkg/selectors/kernel_test.go
```diff
@@ -1094,3 +1095,17 @@ func TestParseAddr(t *testing.T) {
...
 }
+
+func TestParseCapabilityMask(t *testing.T) {
+	v, err := parseCapabilitiesMask("100")
+	require.NoError(t, err)
+	assert.Equal(t, uint64(100), v)
+	v, err = parseCapabilitiesMask("CAP_SYS_ADMIN")
+	require.NoError(t, err)
+	assert.Equal(t, (uint64(1) << 21), v)
+	v, err = parseCapabilitiesMask("CAP_SYS_ADMIN,CAP_BPF")
```
tdaudi, line 1108: If I'm correct, you will have the same result by setting the capability horizontally and vertically. I'm afraid that it can become confusing.

The 2 examples should produce the exact same result. So I'm wondering why the `,` could be necessary.

```yaml
matchArgs:
 - index: X
   operator: Mask
   values:
   - CAP_A,CAP_B
 ```
 
 And 

```yaml
matchArgs:
 - index: X
   operator: Mask
   values:
   - CAP_A
   - CAP_B
 ```

Comment (tdaudi) [reply pkg/selectors/kernel_test.go:1108]: Moreover, when I read `CAP_A,CAP_B`, I would naturally understand `if the value have both A and B capabilities` where it is `A or B`.

Comment (kkourt) [reply pkg/selectors/kernel_test.go:1108]: I think there are two different things: the capability value we construct and the operator we use.

 A value of  `CAP_A,CAP_B` means: build a capability value with `CAP_A` _and_ `CAP_B` set.
 
 Under the mask operator, the selector will be translated to:
`argX & ((1<<CAP_A)|(1<<CAP_B))` which matches if `CAP_A` _or_ `CAP_B` is set.


Similarly, under the mask operator writing:

```
operator: Mask
values:
   - CAP_A
   - CAP_B
```

And
```
operator: Mask
values:
 - CAP_A,CAP_B
 ```
 
 Has the same semantics, since the second selection expression (`argX & (1<<CAP_A) || argX & (1<<CAP_B)`) has the same result as the first.
 
 Note, however, that the versions have different semantics for the  `Equals` operator.

pkg/selectors/kernel.go
kkourt: Makes sense, will re-use your commit.

pkg/selectors/kernel.go
kkourt: I feel that's probably OK. There is a `u64` argument, and the user tells us to interpret it as a capability so making it explicit it the type seems sensible to me.

pkg/selectors/kernel_test.go
tdaudi: Maybe we need to update the mask from this :

https://github.com/cilium/tetragon/blob/b06bec3936609ad11ee3e93f30785a12b72349ec/bpf/process/types/basic.h#L1223-L1225

to

```c
case op_filter_mask:
     u64 ret = *(u64 *)args & w
     return ret == w;
```
So in this case if we have a mask like `CAP_A,CAP_B`, we ensure that both capabilities are set
It also works for a single cap because the output of the `&` will also be equal to the mask.

kkourt: I feel like this is a different operator. Not sure how I would name it though 🤔

pkg/selectors/kernel.go
kkourt: Pushed a new version to avoid code duplication that uses your commit.  Thanks!

michi-covalent (approved):

mtardy (approved):
lgtm, again "value1,value2" definitely looks odd for YAML but why not. Since it's the same semantic eventually would it be fine to just have the YAML list?

```yaml
operator: Mask
values:
   - CAP_A
   - CAP_B
```

Especially parsing "value1,value2" maybe we should trim the space etc... so if we leave it to YAML maybe that's for the best?

kkourt:
> lgtm, again "value1,value2" definitely looks odd for YAML but why not. Since it's the same semantic eventually would it be fine to just have the YAML list?
> 
> ```yaml
> operator: Mask
> values:
>    - CAP_A
>    - CAP_B
> ```
> 
> Especially parsing "value1,value2" maybe we should trim the space etc... so if we leave it to YAML maybe that's for the best?

As discussed in https://github.com/cilium/tetragon/pull/3852#discussion_r2166659814, `Mask` happens to have the same semantics (so a normal YAML list is fine) but for the Equal operator this is not the case.

Also, we are already trimming the whitespace so something like https://github.com/cilium/tetragon/blob/a993dc0b28db4dd956049b3befe668804887094c/pkg/selectors/kernel_test.go#L1113 work as expected.

mtardy:
> As discussed in [#3852 (comment)](https://github.com/cilium/tetragon/pull/3852#discussion_r2166659814), `Mask` happens to have the same semantics (so a normal YAML list is fine) but for the Equal operator this is not the case.

ah sorry I missed this could be used with another operator! ok cool then

> Also, we are already trimming the whitespace so something like

amazing

kkourt:
> > As discussed in [#3852 (comment)](https://github.com/cilium/tetragon/pull/3852#discussion_r2166659814), `Mask` happens to have the same semantics (so a normal YAML list is fine) but for the Equal operator this is not the case.
> 
> ah sorry I missed this could be used with another operator! ok cool then

To be fair, there are some operators (e.g., GT) where using this does not make a lot of sense. But `Equal` could be useful.

pkg/selectors/kernel_test.go
tdaudi: Do you think the behavior (the result) is different or the meaning of "mask" differ ?

kkourt:
Merging this. Happy to address additional feedback as followups.

tdaudi:
Sorry, I think I misunderstood something with this current implementation 

Having
```yaml
- "CAP_A,CAP_B
```
and
```yaml
- "CAP_A"
- "CAP_B"
```

This should produce the same result if I'm correct. So I do not understand why keep things this way ?

Or why not changing the mask operator as suggested in this comment https://github.com/cilium/tetragon/pull/3852#discussion_r2166696377 to have  "CAP_A,CAP_B" doing AND comparison and the other doing OR

kkourt:
> Sorry I think I miss understood something with this current implementation
> 
> Having
> 
> ```yaml
> - "CAP_A,CAP_B
> ```
> 
> and
> 
> ```yaml
> - "CAP_A"
> - "CAP_B"
> ```
> 
> This should produce the same result if i'm correct. So I do not understand why keep things this way ?

I don't think it would produce the same result. As I wrote in the comment above:

```yaml
operator: Equal
values:
 - "CAP_A,CAP_B"
 ```
 
 and
 
 ```yaml
 operator: Equal
 values:
 - "CAP_A"
 - "CAP_B"
 ```
 
 Will produce a different result.

tdaudi:
Ah, I'm sorry, I thought the implementation was still concerning the Mask operator. I apologize for this 🙏

kkourt:
> Ah, I'm sorry, I thought the implementation was still concerning the Mask operator. I apologize for this.

No need.

And to be fair, I do think that the `Mask` operator is probably the most useful operator in this context. But since we support other operators as well, and the `Equal` _does_ make same sense, I thought supporting the list in individual values was the best option. The other option would be to disallow users from using any other operator than `Mask`, but this felt too excessive,

tdaudi:
Do you think a new operator called `Include` in this case could make sense in this case ?
It does the same `&` mask operation for capabilities and iterate over characters when in it's a string.

kkourt:
> Do you think a new operator called `Include` in this case could make sense in this case ? It does the same `&` mask operation for capabilities and iterate over characters when in it's a string.

Can you provide an example?

tdaudi:
In my use case for containers, I want to check whether the attributes include a specific capability, but I don't want to list all of them.

In the following example:
https://github.com/cilium/tetragon/blob/1cd2ab3646c27ba9fe3e4282b33d475f11ca074c/examples/tracingpolicy/deny-privileged-containers.yml

The `--privileged` flag in Docker (I think) assigns all system capabilities to the container. However, I'm only interested in detecting a single one: `CAP_SYS_ADMIN`.
Using the Equal operator is less practical in this case because it requires listing all the capabilities, which is more complex and less maintainable.

And the cool thing you pointed with `CAP_A, CAP_B` is to check if both caps are included in the attribute

PR #3853, opened 2025-06-26
sensor: reduce logs in loading/unloading

kkourt:
Commit 04a1c6fa9dd2e49dba0ab112e34b97dffa2fc25c attempted to reduce logs by concentrating all the info in a single line. This is still too much information. This commit removes the sensors programs and maps from the loading/unloading log messages.

### DIFF
--- a/pkg/sensors/load.go
+++ b/pkg/sensors/load.go
@@ -171,8 +171,8 @@ func (s *Sensor) Load(bpfDir string) (err error) {
 	// cleanup the BTF once we have loaded all sensor's program
 	flushKernelSpec()
 
-	l.Info("Loaded BPF maps and events for sensor successfully",
-		"sensor", s.Name, "maps", loadedMaps, "progs", loadedProgs)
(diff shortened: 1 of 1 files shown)


pkg/sensors/load.go
```diff
@@ -218,7 +217,7 @@ func (s *Sensor) Unload(unpin bool) error {
 	}
 
 	cleanupProgsAndMaps()
-	logger.GetLogger().Info("Sensor unloaded", "maps", mapsOk, "maps-error", mapsErr, "progs", progs)
+	logger.GetLogger().Info("Sensor unloaded", "sensor", s.Name, "maps-error", mapsErr)
 	return nil
```
olsajiri, line 223: I just found those useful when checking on maps issue, could we leave the basic stuff as Info and add all other details to log.Debug call?

kkourt: Yap, makes sense.
Pushed a new version.

michi-covalent (approved):

olsajiri (approved):
thanks

PR #3859, opened 2025-06-27
tetragon/windows: Add sockops attach type GUID to windows loader

ExceptionalHandler:
### Description
This PR adds another program attach type - sockops for Windows based ebpf programs. Also, the attach type to guid string conversion now uses a map.


olsajiri (approved):
lgtm

PR #3604, opened 2025-04-07
Add nameOverride and fullnameOverride support to Tetragon Helm chart

cy83rc0llect0r:
**Enhancement #2951**

### Description  
This pull request adds support for `nameOverride` and `fullnameOverride` in the Tetragon Helm chart, addressing the enhancement request in issue #2951. This feature allows users to customize the names of Kubernetes resources deployed via Helm, improving flexibility and compatibility in diverse environments (e.g., when deploying multiple instances or following naming conventions).

The changes include:

- Addition of `nameOverride` and `fullnameOverride` fields to `values.yaml`.
- Introduction of a new `tetragon.fullname` helper in `_helpers.tpl` to unify naming logic across templates.
- Updates to all affected templates (DaemonSet, Deployment, ServiceAccount, etc.) to use the new helper instead of `.Release.Name`.
- Tested the changes using `helm template` to validate correct name rendering.

These changes are fully backward compatible: if `nameOverride` and `fullnameOverride` are not set, the chart behaves as before.

### Use Case Example  
A user deploying multiple instances of Tetragon (e.g., in different namespaces or clusters) can now avoid resource name collisions by customizing the full names of the Kubernetes objects. This is especially useful in CI/CD pipelines or large multi-tenant setups where naming conventions must be enforced.

### Changelog  
Added `nameOverride` and `fullnameOverride` support to the Tetragon Helm chart. Use these fields in `values.yaml` to customize resource names.

PR #3864, opened 2025-07-03
feat: add `nameOverride` support to helm templates

slntopp:
### Description

When using Tetragon as a Helm dependency, all resources are named after the release name, causing:
- Naming conflicts with parent application resources
- Confusion when managing multiple releases
- Resources appearing at the same level as the parent app instead of clearly being dependencies

#### Solution
Added `tetragon.nameOverride` and `tetragonOperator.nameOverride` fields to allow explicit control over resource naming independent of the Helm release name.

**Benefits:**
- Prevents resource naming conflicts
- Clearer separation between app and dependency resources  
- Better operational visibility in complex deployments
- Follows standard Helm naming patterns

**Example usage:**
```yaml
tetragon:
  nameOverride: "security-agent"
tetragonOperator:
  nameOverride: "security-operator"
```

This maintains backward compatibility (defaults to release name when empty) while solving the dependency naming pain point.

### Changelog
- adds `nameOverride` to tetragon and tetragonOperator Helm values.
- adds helper templates and makes resources names dependent on tetragon and operator names

```release-note
helm: add `tetragon.nameOverride` and `tetragonOperator.nameOverride`
```

### DIFF
--- a/docs/content/en/docs/reference/helm-chart.md
+++ b/docs/content/en/docs/reference/helm-chart.md
@@ -114,6 +114,7 @@ To use [the values available](#values), with `helm install` or `helm upgrade`, u
 | tetragon.image.repository | string | `"quay.io/cilium/tetragon"` |  |
 | tetragon.image.tag | string | `"v1.4.1"` |  |
 | tetragon.livenessProbe | object | `{}` | Overrides the default livenessProbe for the tetragon container. |
+| tetragon.nameOverride | string | `""` |  |
 | tetragon.podAnnotations.enabled | bool | `false` |  |
 | tetragon.pprof.address | string | `"localhost"` | The address at which to expose pprof. |
 | tetragon.pprof.enabled | bool | `false` | Whether to enable exposing pprof server. |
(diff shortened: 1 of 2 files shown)


mtardy:
Hey just fyi we have a draft PR opened for that https://github.com/cilium/tetragon/pull/3604 which seems a bit abandoned. Will review this one thanks! :)

michi-covalent (approved):
looks great 🥰

install/kubernetes/tetragon/templates/_helpers.tpl
```diff
@@ -1,3 +1,35 @@
+{{/*
+Resources names
+*/}}
+{{- define "tetragon.name" -}}
+{{- default .Release.Name .Values.tetragon.nameOverride | trunc 63 | trimSuffix "-" }}
```
kkourt, line 5: Why is `trimSuffix` needed here?

slntopp: Most k8s object names and ids must start and end with alphanumeric chars, this is common failsafe.

> You could theoretically set your release name or nameOverride to `amazing-tetragon-deployment-named-long-with-dashes-because-why-not` (66 chars), truncated to 63 would result into `amazing-tetragon-deployment-named-long-with-dashes-because-why-`, which would fail

install/kubernetes/tetragon/templates/_helpers.tpl
```diff
@@ -1,3 +1,35 @@
...
+{{- end }}
+
+{{- define "tetragon.configMapName" -}}
+{{- printf "%s-config" (include "tetragon.name" .) | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{- define "tetragon.clusterRole" -}}
+{{- include "tetragon.name" . }}
+{{- end }}
```
mtardy, line 14: nit: maybe we don't need those aliases with tetragon-operator? not sure

slntopp:
Best practice is for `app.kubernetes.io/instance` to be the same across all workloads deployed under the same release, so you can do things like `kubectl get all -l app.kubernetes.io/instance=my-release`

install/kubernetes/tetragon/templates/_helpers.tpl
slntopp: do you mean these?
https://github.com/cilium/tetragon/pull/3864/files/0da7b72a1b392a00f6b424eace8a0b7f78694235#diff-6e9c141c310a62601f514de8bdac44a904783e9c65e76895f8f76817cc7c994eR20-R22

mtardy: as I understood `tetragon.clusterRole` is just an alias to `tetragon.name` right now? I was wondering if it was necessary to create those but who cares, let's merge this

PR #3875, opened 2025-07-10
cel: Apply EvalCEL only on events that are related to the rule

tpapagian:
Now filterByCELExpression we run every cel rule on all events. In some cases, it is clear that there is no value to test that as it will not match. For example we have `process_exec.process.bad_field_name == 'curl'` and we try to match that on a process_exit event.

To avoid those overheads, we change CompileCEL to also return the list of events that are related to the rule that we compiled. This allows us to avoid unnecessary calls to EvalCEL.

### DIFF
--- a/pkg/filters/cel_expression.go
+++ b/pkg/filters/cel_expression.go
@@ -6,7 +6,9 @@ package filters
 import (
 	"context"
 	"fmt"
+	"maps"
 	"reflect"
+	"slices"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/api/v1/tetragon/codegen/helpers"
@@ -58,14 +60,19 @@ func EvalCEL(ctx context.Context, program cel.Program, eventMap map[string]any) 
 	return false, nil
 }
 
+type celProgEv struct {
+	program    cel.Program
+	eventNames []string
+}
+
 func (c *CELExpressionFilter) filterByCELExpression(ctx context.Context, log logger.FieldLogger, exprs []string) (FilterFunc, error) {
-	var programs []cel.Program
+	var programs []celProgEv
 	for _, expr := range exprs {
-		prg, err := c.CompileCEL(expr)
+		prog, eventNames, err := c.CompileCEL(expr)
 		if err != nil {
 			return nil, err
 		}
-		programs = append(programs, prg)
+		programs = append(programs, celProgEv{program: prog, eventNames: eventNames})
 	}
 
 	return func(ev *event.Event) bool {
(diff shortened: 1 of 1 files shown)


pkg/filters/cel_expression.go
```diff
@@ -116,24 +131,32 @@ func NewCELExpressionFilter(log logger.FieldLogger) *CELExpressionFilter {
...
-		log:    log,
-		celEnv: celEnv,
+		log:       log,
+		evToProto: evToProto,
+		celEnv:    celEnv,
 	}
 }
 
-func (c *CELExpressionFilter) CompileCEL(expr string) (cel.Program, error) {
+func (c *CELExpressionFilter) CompileCEL(expr string) (cel.Program, []string, error) {
```
kkourt, line 142: Please add a comment of what the 2nd return value is.

pkg/filters/cel_expression.go
```diff
@@ -116,24 +131,32 @@ func NewCELExpressionFilter(log logger.FieldLogger) *CELExpressionFilter {
...
 	// we want filters to be boolean expressions, so check the type of the
 	// expression before proceeding
 	ast, err := compile(c.celEnv, expr, cel.BoolType)
 	if err != nil {
-		return nil, fmt.Errorf("error compiling CEL expression: %w", err)
+		return nil, []string{}, fmt.Errorf("error compiling CEL expression: %w", err)
```
kkourt: Here, we can just return nil, nil, fmt.Errorf("...

pkg/filters/cel_expression.go
```diff
@@ -116,24 +131,32 @@ func NewCELExpressionFilter(log logger.FieldLogger) *CELExpressionFilter {
...
+	}
+
+	uniqEvents := map[string]struct{}{}
+	for _, tp := range ast.NativeRep().TypeMap() {
+		if ev, ok := c.evToProto[tp.TypeName()]; ok {
+			uniqEvents[ev] = struct{}{}
+		}
```
kkourt, line 154: OK, so *if* I understand correctly, TypeMap() provides a mapping from all expressions (i.e., their id) to their type. if an event type is not in this map, then it's not used in the expression and we can ignore it.

Nice!!!

Can you please add a brief explanation on how this works for our future selves?

pkg/filters/cel_expression.go
tpapagian: Added a comment there.

pkg/filters/cel_expression.go
tpapagian: Changed that to nil.

pkg/filters/cel_expression.go
tpapagian: Added some comments there.

> if an event type is not in this map, then it's not used in the expression and we can ignore it.

The actual check happens in `filterByCELExpression` (i.e. before calling `EvalCEL`).

PR #3887, opened 2025-07-11
[RFC] add CapabiliitesGained operator

kkourt:
Change the newly introduced `Arg` to be an array so that we can support operators that work with multiple arguments.
(The Arg change has not yet been released, so this is not a breaking change).

Using above change, implement a `CapabilitiesGained` operator.

See commits.

Example policy:
```yaml
kind: TracingPolicy
metadata:
  name: "capabilites-gained"
spec:
  kprobes:
  - call: "security_capset"
    syscall: false
    return: true
    args:
      - index: 1
        type: "cap_effective"
        resolve: "cap_effective"
        label: "old_creds.cap_effective"
      - index: 2
        type: "cap_effective"
        label: "cap_effective"
    returnArg:
      index: 0
      type: "int"
    selectors:
      - matchActions:
        - action: Post
        matchArgs:
        - operator: CapabilitiesGained
          args:
          - 0
          - 1
    message: capability escalation
```


bpf/process/types/basic.h
```diff
@@ -1650,13 +1655,21 @@ selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx,
...
 			break;
-		case syscall64_type:
-		case s64_ty:
-		case u64_ty:
-		case kernel_cap_ty:
 		case cap_inh_ty:
 		case cap_prm_ty:
 		case cap_eff_ty:
+		case kernel_cap_ty:
+			if (filter->op == op_capabilities_gained) {
```
olsajiri, line 1663: I guess having op_capabilities_gained in ops implies filter->vallen == 4, so no need to check?

pkg/selectors/kernel.go
```diff
@@ -758,14 +762,30 @@ func checkOp(op uint32) error {
...
+
 func argIndexType(arg *v1alpha1.ArgSelector, sig []v1alpha1.KProbeArg) (uint32, uint32, error) {
-	if arg.Arg != nil {
-		index := *arg.Arg
-		if index >= uint32(len(sig)) {
-			return 0, 0, fmt.Errorf("wrong ArgSelector.Arg value %d, max(%d)", index, uint32(len(sig)))
-		}
-		ty := sig[index].Type
-		return index, uint32(gt.GenericTypeFromString(ty)), nil
+	if len(arg.Args) > 0 {
```
olsajiri, line 792: should we bail with error in case of len(arg.Args) != 0 ?

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -205,10 +205,10 @@ type ArgSelector struct {
 	// +kubebuilder:validation:Minimum=0
 	// Position of the argument (in function prototype) to apply fhe filter to.
 	Index uint32 `json:"index"`
-	// +kubebuilder:validation:Minimum=0
-	// Position of the argument (in spec file) to apply fhe filter to.
-	Arg *uint32 `json:"arg,omitempty"`
-	// +kubebuilder:validation:Enum=Equal;NotEqual;Prefix;NotPrefix;Postfix;NotPostfix;GreaterThan;LessThan;GT;LT;Mask;SPort;NotSPort;SPortPriv;NotSportPriv;DPort;NotDPort;DPortPriv;NotDPortPriv;SAddr;NotSAddr;DAddr;NotDAddr;Protocol;Family;State;InMap;NotInMap
+	// +kubebuilder:validation:items:Minimum=0
+	// Position of the operator arguments (in spec file) to apply fhe filter to.
+	Args []uint32 `json:"args,omitempty"`
```
olsajiri, line 210: not sure it'd bring more complications, but could we have both? Arg and Args

olsajiri (commented):
looks good, left some comments, thanks

bpf/process/types/basic.h
kkourt: Yap, that's the idea. Happy to add a check, but not sure what the reasonable thing to would be if the check fails.

pkg/selectors/kernel.go
kkourt: I think not. The idea is that len(arg.Args) == 0 is the same as the case where arg.Arg == nil in the old code, so we just use the `Index`.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
kkourt: Yap, we could have both. In that case, I would opt for `Arg` and `ExtraArgs`.
I think having a simple field is cleaner from the user perspective, but happy to add a new field if folk feel  it's better.

kkourt:
Pushed a new version where I: merged commits so that all commits compile, added a test, constrained this feature only for kernels that support large programs (since 4.19 was failing to load the prog).

mtardy (approved):
looks good to me! (indeed missing the updated docs for now)

pkg/k8s/apis/cilium.io/v1alpha1/types.go
mtardy: Even if it makes the UX a bit more complicated with `Args` when specifying only one arg, I guess it's also simpler only have one the array!

kkourt:
> looks good to me! (indeed missing the updated docs for now)

Thanks! Pushed a new commit that also updates that docs.

PR #3888, opened 2025-07-13
docs: Clarify export filter scope (JSON vs gRPC)

### LINKED ISSUE #3742: docs: add warning/caution that denylist/allowlist only affect the file export
I can't count the time people have been confused by that: 

1. they are applying a denylist/allowlist;
2. they are checking the event through the gRPC via tetra;
3. they still see the event they are trying to filter;
4. they are confused 😵‍💫.

We should be very clear that those filters only affect the event exported in the JSON file.

We should be very clear in the documentation of those filters which part they affect, we can use the `{{< caution >}}` and co to highlight this.

f4r00q:
Fixes #3742

### Description
This PR clarifies the scope of `denylist` and `allowlist` export filters in the Tetragon documentation. Users were confused about why filtered events still appeared in `tetra` CLI output, not realizing these filters only apply to JSON file exports and not to gRPC streaming.

The changes add:
- A prominent caution note in `docs/content/en/docs/events.md` using the `{{< caution >}}` shortcode to clearly explain that export filters only affect JSON file sinks
- Concise inline notes in the Helm chart value tables in both `helm-chart.md` and `README.md` to indicate the filters are "for file sinks only"

These documentation updates will help users understand the intended behavior and avoid confusion when using the `tetra` CLI alongside export filters.

### Changelog
```release-note
docs: Clarify that export filters (denylist/allowlist) only apply to JSON file exports, not gRPC streaming
```

### DIFF
--- a/docs/content/en/docs/concepts/events.md
+++ b/docs/content/en/docs/concepts/events.md
@@ -146,6 +146,15 @@ in cases where two filter configurations match on the same event.
 You can configure export filters using the provided helm options, command line
 flags, or environment variables.
 
+{{< caution >}}
+The `denylist` and `allowlist` filters mentioned above only apply to events
+exported as JSON files (e.g., through a configured `file` sink). These filters
+do not affect events streamed via the gRPC API, which are consumed by tools
+like the `tetra` CLI. You will still see unfiltered events when using the
+`tetra` CLI, even if those events would be filtered out by `denylist` or
+`allowlist` for JSON export.
+{{< /caution >}}
+
 ##### List of Process Event Filters
 
 The type for each filter attribute can be determined by referring to the 
--- a/docs/content/en/docs/reference/helm-chart.md
+++ b/docs/content/en/docs/reference/helm-chart.md
@@ -90,7 +90,7 @@ To use [the values available](#values), with `helm install` or `helm upgrade`, u
 | tetragon.eventCacheRetries | int | `15` | Configure the number of retries in tetragon's event cache. |
 | tetragon.eventCacheRetryDelay | int | `2` | Configure the delay (in seconds) between retires in tetragon's event cache. |
 | tetragon.exportAllowList | string | `"{\"event_set\":[\"PROCESS_EXEC\", \"PROCESS_EXIT\", \"PROCESS_KPROBE\", \"PROCESS_UPROBE\", \"PROCESS_TRACEPOINT\", \"PROCESS_LSM\"]}"` | Allowlist for JSON export. For example, to export only process_connect events from the default namespace:  exportAllowList: |   {"namespace":["default"],"event_set":["PROCESS_EXEC"]} |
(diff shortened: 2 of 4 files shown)


kkourt (approved):
Thanks! Wanted add something like this for a long time.

docs/content/en/docs/concepts/events.md
```diff
@@ -146,6 +146,16 @@ in cases where two filter configurations match on the same event.
 You can configure export filters using the provided helm options, command line
 flags, or environment variables.
 
+{{< caution >}}
+**Important Note on Export Filters:**
```
mtardy: Just a nit, could you maybe remove the title "important note on export filters", it makes the card a double title that is unusual in the docs:

image

You could just start the sentence, "On export filter, the denylist..."

mtardy (commented):
Thanks a lot for taking the time to do this, just a little nit

docs/content/en/docs/concepts/events.md
calghar: Made the change.  Please let me know if you see any other issues

mtardy:
and can you fixup your second commit (minor edit) into the first one please? we just rebase and merge by default :)

docs/content/en/docs/concepts/events.md
mtardy: thanks, that's very nice, just that and we can merge https://github.com/cilium/tetragon/pull/3888#issuecomment-3074362166

mtardy:
Oh I realized you added stuff manually to generated files `docs/content/en/docs/reference/helm-chart.md` and `install/kubernetes/tetragon/README.md`. Please add the comment directly to values.yaml https://github.com/cilium/tetragon/blob/main/install/kubernetes/tetragon/values.yaml and then `make -C install/kubernetes` otherwise it will get overwritten.

We could have a CI check for that sorry, it's somehow written on them that they are generated but maybe that's not extremely obvious!

calghar:
> Oh I realized you added stuff manually to generated files `docs/content/en/docs/reference/helm-chart.md` and `install/kubernetes/tetragon/README.md`. Please add the comment directly to values.yaml https://github.com/cilium/tetragon/blob/main/install/kubernetes/tetragon/values.yaml and then `make -C install/kubernetes` otherwise it will get overwritten.
> 
> We could have a CI check for that sorry, it's somehow written on them that they are generated but maybe that's not extremely obvious!

Yeah, no worries 🙂. That was my bad, I should have seen it. Will make another commit and then squash it all into one. Sorry for the noise

mtardy:
> Yeah, no worries 🙂. That was my bad, I should have seen it. Will make another commit and then squash it all into one. Sorry for the noise

really no noise, it's me dropping the review in slices unfortunately (and again it's written like at the very end for the README and at the very top for the other so it's not amazing dev ux)! And don't be sorry it's super helpful, I agree with Kornilios that
> [We] wanted to add something like this for a long time.

mtardy (approved):
Thank you that's perfect!

calghar:
> Thank you that's perfect!

Awesome. 🙂
Also really appreciate the work the team has put into the project. Have been using Tetragon quite a lot recently and its really amazing. 👏
Glad to be able to contribute.

PR #3895, opened 2025-07-16
tracingpolicy: respect syscall attribute in lists

kkourt:
If we have a policy such as:

```
kind: TracingPolicy
metadata:
 name: "unload-module"
spec:
  lists:
    - name: "sys_delete_module" type: "generated_ftrace" pattern: "sys_delete_module" kprobes:
  - call: "list:sys_delete_module" syscall: true args:
      - index: 0  # module (user-space pointer, so unsafe) type: "string"
      - index: 1  # flags type: "uint32"
```

The "syscall: true" is not respected when generating the calls. This commit fixes this.

```release-note
tracinpolicy: respect syscall attribute in lists
```


olsajiri (approved):
thanks

PR #3901, opened 2025-07-16
tetragon: add support to follow children of old process

olsajiri:
- add v6.12 objects
- add support to follow children of old process

### DIFF
--- a/bpf/process/bpf_execve_map_update.c
+++ b/bpf/process/bpf_execve_map_update.c
@@ -24,31 +24,12 @@ struct {
 
 char _license[] __attribute__((section("license"), used)) = "Dual BSD/GPL";
 
-#ifdef __V63_BPF_PROG
-FUNC_INLINE void
-__execve_map_update(struct update_data *data)
-{
-	struct execve_map_value *curr;
-	struct bpf_iter_num it;
-	int *v;
-
-	bpf_iter_num_new(&it, 0, MAX_PIDS);
-	for (v = bpf_iter_num_next(&it); v; v = bpf_iter_num_next(&it)) {
-		__u32 pid, idx = (__u32)*v;
-
-		asm volatile("%[idx] &= 0x7fff;\n"
-			     : [idx] "+r"(idx));
-		if (data->cnt == idx)
-			break;
-
-		pid = data->pids[idx];
-		curr = execve_map_get_noinit(pid);
-		if (curr)
-			lock_and(&curr->bin.mb_bitset, ~(1 << data->bit));
-	}
-	bpf_iter_num_destroy(&it);
-}
+#ifdef __V612_BPF_PROG
+#define LOOPS MAX_PIDS
 #else
+#define LOOPS 1024
+#endif
+
 #ifdef __LARGE_BPF_PROG
 FUNC_INLINE void
 __execve_map_update(struct update_data *data)
@@ -56,7 +37,8 @@ __execve_map_update(struct update_data *data)
 	struct execve_map_value *curr;
 	__u32 idx, pid;
 
-	for (idx = 0; idx < 1024; idx++) {
+	bpf_for(idx, 0, LOOPS)
+	{
 		asm volatile("%[idx] &= 0x7fff;\n"
 			     : [idx] "+r"(idx));
 		if (data->cnt == idx)
@@ -74,7 +56,6 @@ __execve_map_update(struct update_data *data)
 {
 }
 #endif /* __LARGE_BPF_PROG */
-#endif /* __V63_BPF_PROG */
 
 __attribute__((section("seccomp"), used)) int
 execve_map_update(void *ctx)
--- a/bpf/process/bpf_mbset.h
+++ b/bpf/process/bpf_mbset.h
@@ -15,6 +15,13 @@ struct {
 	__type(value, mbset_t);
 } tg_mbset_map SEC(".maps");
 
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, __u32);
+	__type(value, __u64);
+} tg_mbset_gen SEC(".maps");
+
 /* update bitset mark */
 FUNC_INLINE
 void update_mb_bitset(struct binary *bin)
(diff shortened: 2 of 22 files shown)


pkg/config/config_linux.go
```diff
@@ -121,6 +140,14 @@ func EnableV61Progs() bool {
 	return (int64(kernelVer) >= kernels.KernelStringToNumeric("6.1.0"))
 }
 
+func EnableV612Progs() bool {
+	if option.Config.ForceSmallProgs {
+		return false
+	}
+	kernelVer, _, _ := kernels.GetKernelVersion(option.Config.KernelVersion, option.Config.ProcFS)
```
kkourt, line 147: Can we add a probe for this instead of relying to the kernel version?

Generally, I think we should 6.12 (and all other) objects rely on a set of features that can we can probe at runtime.

Also, is 6.12 a new variant?

bpf/process/bpf_execve_map_update.c
```diff
@@ -24,39 +24,21 @@ struct {
...
+
 #ifdef __LARGE_BPF_PROG
 FUNC_INLINE void
 __execve_map_update(struct update_data *data)
 {
 	struct execve_map_value *curr;
 	__u32 idx, pid;
 
-	for (idx = 0; idx < 1024; idx++) {
+	bpf_for(idx, 0, LOOPS)
```
kkourt, line 40: Why is this better than the `for` loop? Could you add some info in the commit message?

bpf/process/bpf_mbset.h
```diff
@@ -38,4 +45,40 @@ void update_mb_bitset(struct binary *bin)
...
+		bitsetp = map_lookup_elem(&tg_mbset_map, parent->bin.path);
+		if (bitsetp && *bitsetp)
+			lock_or(&task->bin.mb_bitset, *bitsetp);
+		last = parent;
+	}
+
+	task->bin.mb_gen = *gen;
+}
+#else
+#define update_mb_task(task)
```
kkourt, line 80: is this supposed to be empty code? How do you feel about defining this to a `do {} while (0)`

bpf/process/bpf_mbset.h
```diff
@@ -0,0 +1,41 @@
...
+	parent = event_find_parent();
+	if (parent) {
+		/* ->mb_bitset is used to track matchBinary matches to children (followChildren), so
+		 * here we propagate the parent value to the child.
+		 */
+		lock_or(&bin->mb_bitset, parent->bin.mb_bitset);
+	}
+
+	/* check the map and see if the binary path matches a binary
+	 * NB: only the In operator is supported for now
```
kkourt: I think what we do for `In` should also work for `NotIn` that we are currently also supporting. Do you agree? :)

contrib/tester-progs/follow_children_1.c
```diff
@@ -0,0 +1,50 @@
...
+	p = fork();
+	if (p < 0) {
+		perror("fork");
+		return -1;
+	}
+
+	if (p == 0) {
+		read(kick[0], &c, 1);
+		prctl(0xdead, 0, 0, 0, 0);
+		pause();
```
kkourt, line 53: Is the intention here to continue with the rest of the code that is also executed in the parent or should we exit the child?

contrib/tester-progs/follow_children_1.c
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	pause();
```
kkourt, line 57: Can you add some comments to figure out how the different sections match the test code?
For example: /* parent wait point p1 */

pkg/sensors/tracing/matchbinaries_follow_children_test.go
```diff
@@ -323,3 +324,125 @@ func TestMatchBinariesFollowChildrenUpdate(t *testing.T) {
...
+	err = sm.Manager.AddTracingPolicy(ctx, &tp)
+	require.NoError(t, err)
+
+	// Make sure it's properly terminated
+	t.Cleanup(func() {
+		cmd.Process.Signal(syscall.Signal(2))
+		cmd.Process.Wait()
+	})
+
+	cmd.Process.Signal(syscall.Signal(10))
```
kkourt: // wake parent from parent wait point p1

contrib/tester-progs/follow_children_1.c
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	p = fork();
+	if (p < 0) {
+		perror("fork");
+		return -1;
+	}
+
+	if (p == 0) {
+		read(kick[0], &c, 1);
```
kkourt, line 50: /* child wait point c1 */

contrib/tester-progs/follow_children_1.c
```diff
@@ -0,0 +1,50 @@
...
+		prctl(0xdead, 0, 0, 0, 0);
+		pause();
+	}
+
+	pause();
+	write(kick[1], &c, 1);
```
kkourt, line 59: /* wake up child from child wait point c1 */

kkourt (commented):
Thanks!

Overall, LGTM. Please find some small comments below.

pkg/config/config_linux.go
olsajiri: I need to use open numeric interator (added around 6.3) so I picked the latest longterm stable and used that as new variant ... that iteration takes place in the middle of the generic event code, so I'm not sure I could do it some other way ... are you concerned about another variant object? @kkourt

bpf/process/bpf_execve_map_update.c
olsajiri: with the iterator we could do 100k loops and verifier won't complain, while 1024 seems to be the max in this code for large programs

bpf/process/bpf_mbset.h
olsajiri: hum, what's the benefit?

bpf/process/bpf_mbset.h
olsajiri: i agree ;-) will remove that comment

contrib/tester-progs/follow_children_1.c
olsajiri: I think the test kills the process when it hungs in pause, I'll double check and add some comment

contrib/tester-progs/follow_children_1.c
olsajiri: will do, sorry

pkg/config/config_linux.go
kkourt: I was just curious if this was a new variant. I don't have any specific concerns about adding a new one.

bpf/process/bpf_mbset.h
kkourt: Can't think of any. Let's leave it as is :)

PR #3909, opened 2025-07-17
Fix/operator non root 3385

### LINKED ISSUE #3385: How to run terragon operator deployment as user.. not as root

Hi, 
We would like to know, How to run terragon operator deployment as user.. not as root

Want to know how to install tetragon as user not as root

_No response_

### Code of Conduct

f4r00q:
## Summary

This PR adds support for running the Tetragon operator as a non-root user, addressing the security enhancement request in #3385. The implementation follows Kubernetes security best practices while maintaining full backward compatibility.

## Related Issue

Fixes #3385

## Proposed Changes

- **Helm Chart Updates**:
  - Added security context configuration options with non-root user using UID/GID 65532
  - Added commented examples in `values.yaml` to guide users on non-root configuration
  
- **Container Image**:
  - Updated `Dockerfile.operator` to create a dedicated non-root user (UID/GID 65532)
  - Set USER directive to ensure container runs as non-root by default

- **Documentation**:
  - Added new file to the installation guide.

## Testing Performed

- [x] Built and tested operator container with non-root user
- [x] Verified container runs as UID 65532 using `docker inspect`
- [x] Deployed to Kind cluster and verified functionality
- [x] Confirmed operator starts successfully with non-root user
- [x] Tested with both pod and container security contexts
- [x] Verified operator can manage all Tetragon resources (CRDs, deployments, etc.)

## Security Considerations

- Uses UID/GID 65532 following distroless conventions
- No privilege escalation allowed

## Backward Compatibility

- Changes are fully opt-in via Helm values
- Default `values.yaml` keeps security contexts empty (current behavior)
- No breaking changes to existing deployments
- Users must explicitly enable non-root mode

### DIFF
--- a/Dockerfile.operator
+++ b/Dockerfile.operator
@@ -38,6 +38,8 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 COPY --from=builder /out/${TARGETOS}/${TARGETARCH}/usr/bin/tetragon-operator /usr/bin/tetragon-operator
 WORKDIR /
 ENV GOPS_CONFIG_DIR=/
+# Avoid root privliges
+USER 65532:65532
 ENTRYPOINT ["/usr/bin/tetragon-operator"]
 
 FROM release
--- a/docs/content/en/docs/installation/kubernetes-operator-security.md
+++ b/docs/content/en/docs/installation/kubernetes-operator-security.md
@@ -1,0 +1,55 @@
+---
+title: "Running Tetragon Operator as Non-Root"
+weight: 6
+description: "Configure the Tetragon operator to run with non-root privileges"
+---
+
+## Running Tetragon Operator as Non-Root
+
+Starting with version X.Y.Z, the Tetragon operator supports running as a non-root user, enhancing the security posture of your Kubernetes deployments.
+
+## Configuration
+
+To run the operator as non-root, configure the following Helm values:
+
+```yaml
+tetragonOperator:
+  podSecurityContext:
+    runAsNonRoot: true
+    runAsUser: 65532
+    runAsGroup: 65532
+    fsGroup: 65532
+    seccompProfile:
+      type: RuntimeDefault
+
+  containerSecurityContext:
+    allowPrivilegeEscalation: false
+    readOnlyRootFilesystem: true
+    capabilities:
+      drop:
+      - ALL
+```
+
+## Installation
+
+Install Tetragon with non-root operator:
+
+```bash
+helm install tetragon cilium/tetragon \
+  --namespace kube-system \
+  --values non-root-values.yaml
+```
+
+## Verification
+
+Verify the operator is running as non-root:
+
+```bash
+kubectl exec -n kube-system deployment/tetragon-operator -- id
+```
+
+Expected output:
+
+```txt
+uid=65532 gid=65532 groups=65532
+```
--- a/install/kubernetes/tetragon/templates/operator_deployment.yaml
+++ b/install/kubernetes/tetragon/templates/operator_deployment.yaml
@@ -49,7 +49,7 @@ spec:
           {{- with .Values.tetragonOperator.extraVolumeMounts }}
             {{- toYaml . | nindent 10 }}
           {{- end }}
-        {{- with .Values.tetragonOperator.securityContext }}
+         {{- with .Values.tetragonOperator.containerSecurityContext}}
         securityContext:
           {{- toYaml . | nindent 10 }}
         {{- end }}
--- a/install/kubernetes/tetragon/values.yaml
+++ b/install/kubernetes/tetragon/values.yaml
@@ -269,14 +269,27 @@ tetragonOperator:
     create: true
     annotations: {}
     name: ""
-  # -- securityContext for the Tetragon Operator Deployment Pods.
+  # -- securityContext for the Tetragon Operator Deployment Pods. (commented out for backwards compatability)
   podSecurityContext: {}
-  # -- securityContext for the Tetragon Operator Deployment Pod container.
-  securityContext:
+  #  runAsNonRoot: true
+  #  runAsUser: 65532
+  #  runAsGroup: 65532
+  #  fsGroup: 65532
+  #  seccompProfile:
+  #    type: RuntimeDefault
+  # -- securityContext for the Tetragon Operator Deployment Pod container. (commented out for backwards compatability)
+  containerSecurityContext:
     allowPrivilegeEscalation: false
     capabilities:
       drop:
         - "ALL"
+  #  readOnlyRootFilesystem: true
+  #  runAsNonRoot: true
+  #  runAsUser: 65532
+  #  runAsGroup: 65532
+
+  # Deprecated: Use containerSecurityContext instead
+  # securityContext: {}
   # -- resources for the Tetragon Operator Deployment Pod container.
   resources:
     limits:


install/kubernetes/tetragon/values.yaml
```diff
@@ -269,14 +269,27 @@ tetragonOperator:
...
-  # -- securityContext for the Tetragon Operator Deployment Pod container.
-  securityContext:
+  #  runAsNonRoot: true
+  #  runAsUser: 65532
+  #  runAsGroup: 65532
+  #  fsGroup: 65532
+  #  seccompProfile:
+  #    type: RuntimeDefault
+  # -- securityContext for the Tetragon Operator Deployment Pod container. (commented out for backwards compatability)
+  containerSecurityContext:
```
mtardy, line 278: so I'm not against deprecating `operator.securityContext` and use `operator.containerSecurityContext` so that it's more explicit but could you move the deprecation in a separate earlier commit and make the fallback natural: meaning that if `containerSecurityContext` is not defined, templating should still use `securityContext` so that it's backward compatible? Thanks

install/kubernetes/tetragon/templates/operator_deployment.yaml
```diff
@@ -49,7 +49,7 @@ spec:
           {{- with .Values.tetragonOperator.extraVolumeMounts }}
             {{- toYaml . | nindent 10 }}
           {{- end }}
-        {{- with .Values.tetragonOperator.securityContext }}
+         {{- with .Values.tetragonOperator.containerSecurityContext}}
```
mtardy: you have a space here, not sure it's intended. Anyway should be change for backward compatibility on previous field name.

Dockerfile.operator
```diff
@@ -38,6 +38,8 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
 COPY --from=builder /out/${TARGETOS}/${TARGETARCH}/usr/bin/tetragon-operator /usr/bin/tetragon-operator
 WORKDIR /
 ENV GOPS_CONFIG_DIR=/
+# Avoid root privliges
```
mtardy: little typo

Dockerfile.operator
```diff
@@ -38,6 +38,8 @@ COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifica
...
+USER 65532:65532
```
mtardy: maybe instead of using a number you can create a un-privileged user by name like `operator` and use it here (making sure it has permission to access the binary)?

docs/content/en/docs/installation/kubernetes-operator-security.md
```diff
@@ -0,0 +1,55 @@
+---
+title: "Running Tetragon Operator as Non-Root"
+weight: 6
+description: "Configure the Tetragon operator to run with non-root privileges"
+---
+
+## Running Tetragon Operator as Non-Root
+
+Starting with version X.Y.Z, the Tetragon operator supports running as a non-root user, enhancing the security posture of your Kubernetes deployments.
```
mtardy: could you fold long lines around 80 chars?

docs/content/en/docs/installation/kubernetes-operator-security.md
```diff
@@ -0,0 +1,55 @@
...
+
+Verify the operator is running as non-root:
+
+```bash
+kubectl exec -n kube-system deployment/tetragon-operator -- id
+```
+
+Expected output:
+
+```txt
```
mtardy: you can remove `txt` here to have a nicely formatted output text.

mtardy (commented):
Thanks for adding the doc, I have a few comments :)

calghar:
Tried to fix all issues @mtardy . Please let me know if you still need changes.
 Thanks!😊

calghar:
Any updates here @mtardy

mtardy:
> Any updates here @mtardy

ah I'm sorry it felt between the cracks, you can re-request review, I check my review list almost every day
image

calghar:
> > Any updates here @mtardy
> 
> ah I'm sorry it felt between the cracks, you can re-request review, I check my review list almost every day image

No worries 😊. I have re-requested the review

mtardy (commented):
ah, can you rebase your PR on top of main and put the fixes into the appropriate commits? We merge rebase so we need a clean commit history. (sorry it took all this time for this remark...)

calghar:
> ah, can you rebase your PR on top of main and put the fixes into the appropriate commits? We merge rebase so we need a clean commit history. (sorry it took all this time for this remark...)

Done 😊

mtardy:
Sorry I'm on PTO next week, coming back end of august, ask for someone else review if you want in the meantime, otherwise see you in a week 👋

install/kubernetes/tetragon/templates/operator_deployment.yaml
```diff
@@ -111,4 +114,4 @@ spec:
   strategy:
     {{- toYaml . | trim | nindent 4 }}
   {{- end }}
-{{- end }}
+{{- end }}
```
mtardy: if you can avoid the last line removal here

install/kubernetes/tetragon/values.yaml
```diff
@@ -274,12 +274,25 @@ tetragonOperator:
...
-  securityContext:
+  containerSecurityContext:
     allowPrivilegeEscalation: false
     capabilities:
       drop:
         - "ALL"
+  #  readOnlyRootFilesystem: true
+  #  runAsNonRoot: true
+  #  runAsUser: 65532
+  #  runAsGroup: 65532
```
mtardy: I see the idea behind those comments but maybe that's giving mixed signals, maybe we should either have the security context default or not?

install/kubernetes/tetragon/values.yaml
```diff
@@ -274,12 +274,25 @@ tetragonOperator:
...
+
+  # -- securityContext for the Tetragon Operator Deployment Pod container. (DEPRECATED: Use containerSecurityContext instead)
+  securityContext: {}
```
mtardy: you can tag this in the comment to be removed in release N+1

install/kubernetes/tetragon/templates/operator_deployment.yaml
```diff
@@ -49,9 +49,12 @@ spec:
...
             {{- toYaml . | nindent 10 }}
           {{- end }}
-        {{- with .Values.tetragonOperator.securityContext }}
+        {{- if .Values.tetragonOperator.containerSecurityContext }}
         securityContext:
-          {{- toYaml . | nindent 10 }}
+          {{- toYaml .Values.tetragonOperator.containerSecurityContext | nindent 10 }}
+        {{- else if .Values.tetragonOperator.securityContext }}
+        securityContext:
+          {{- toYaml .Values.tetragonOperator.securityContext | nindent 10 }}
```
mtardy, line 57: Thanks, it's great like this!

Dockerfile.operator
mtardy: still the typo :D

Dockerfile.operator
```diff
@@ -22,6 +22,11 @@ RUN --mount=type=bind,readwrite,target=/go/src/github.com/cilium/tetragon --moun
     make GOARCH=${TARGETARCH} VERSION=${TETRAGON_VERSION} tetragon-operator \
     && mkdir -p /out/${TARGETOS}/${TARGETARCH}/usr/bin && mv tetragon-operator /out/${TARGETOS}/${TARGETARCH}/usr/bin
 
+# User/group files for image
+RUN mkdir -p /out/etc && \
+    echo 'operator:x:65532:65532:operator:/home/operator:/sbin/nologin' > /out/etc/passwd && \
+    echo 'operator:x:65532:' > /out/etc/group
```
mtardy: question: any reason to not use adduser/groupadd & cie and write by hand like this?

Dockerfile.operator
```diff
@@ -36,8 +41,13 @@ ARG TARGETARCH
...
+COPY --from=builder /out/etc/group /etc/group
+
 WORKDIR /
 ENV GOPS_CONFIG_DIR=/
+# Avoid root privliges
+USER operator:operator
 ENTRYPOINT ["/usr/bin/tetragon-operator"]
 
-FROM release
+FROM release
```
mtardy: same, if you can avoid these empty line changes

docs/content/en/docs/installation/kubernetes-operator-security.md
```diff
@@ -0,0 +1,79 @@
+---
+title: "Running Tetragon Operator as Non-Root"
+weight: 6
+description: "Configure the Tetragon operator to run with non-root privileges"
+---
+
+## Running Tetragon Operator as Non-Root
+
+Starting with version X.Y.Z, the Tetragon operator supports running as a
```
mtardy: Just as a reminder to change X.Y.Z when we merge this.

docs/content/en/docs/installation/kubernetes-operator-security.md
```diff
@@ -0,0 +1,79 @@
...
+
+```yaml
+tetragonOperator:
+  podSecurityContext:
+    runAsNonRoot: true
+    runAsUser: 65532
+    runAsGroup: 65532
+    fsGroup: 65532
+    seccompProfile:
+      type: RuntimeDefault
```
mtardy: maybe having the information in the docs as this is enough and could be pointed to instead of having these commented in the values.yaml?

docs/content/en/docs/installation/kubernetes-operator-security.md
```diff
@@ -0,0 +1,79 @@
...
+    runAsUser: 65532
+```
+
+To:
+
+```yaml
+tetragonOperator:
+  containerSecurityContext:
+    runAsUser: 65532
+```
```
mtardy: thanks, could you write this to https://github.com/cilium/tetragon/blob/main/contrib/upgrade-notes/latest.md as well for next release? :)

install/kubernetes/tetragon/values.yaml
calghar: Sure thing, I thought it would be good to have this in at least the first version where this change occurs.

docs/content/en/docs/installation/kubernetes-operator-security.md
calghar: That makes sense. Like the previous instance just for the first version which had the change I thought it would make sense to keep it

Dockerfile.operator
calghar: Cant remember the exact reason but I build kept on failing. I believe adding user group that way was not possible because of the base image being used. In any case this is more explicit and should hopefully avoid any issues down the line imo.

install/kubernetes/tetragon/values.yaml
calghar: Made this to be 1.6.1 assuming these changes will be incorporated in a minor release but of course this can be changed before merge

calghar:
Anything else that needs to be changed? @mtardy

contrib/upgrade-notes/latest.md
```diff
@@ -11,7 +11,7 @@ Depending on your setup, changes listed here might require a manual intervention
 
 ### Helm Values
 
-* TBD
+* The `tetragonOperator.securityContext` field has been deprecated in favor of `tetragonOperator.containerSecurityContext` for clarity. The old field is still supported for backward compatibility but might be removed in a future release. Users should migrate their configurations to use the new field.
```
mtardy: nit: could you fold this?

Dockerfile.operator
mtardy: I'm sorry it seems there's still 2 empty lines changes here:
1. You are removing line 42
2.You are removing the empty line after line 43

image

Dockerfile.operator
```diff
@@ -36,8 +41,12 @@ ARG TARGETARCH
 LABEL maintainer="maintainer@cilium.io"
 COPY --from=certs /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 COPY --from=builder /out/${TARGETOS}/${TARGETARCH}/usr/bin/tetragon-operator /usr/bin/tetragon-operator
+COPY --from=builder /out/etc/passwd /etc/passwd
+COPY --from=builder /out/etc/group /etc/group
+
 WORKDIR /
 ENV GOPS_CONFIG_DIR=/
+# Avoid root privileges
+USER operator:operator
```
mtardy: so on
> Changes are fully opt-in via Helm values
Default values.yaml keeps security contexts empty (current behavior)
No breaking changes to existing deployments
Users must explicitly enable non-root mode

Because you use this USER directive here, it will be the default deployment mode for starting the tetragon-operator container. So with this, any deployment will be deployed as non-root.

Two things:
- Maybe that's okay and we just switch to non-root for everyone (so you don't need the helm chart shenanigans)
- Or we remove this and ask k8s to use a non-root user via the securityContext with what you did.

docs/content/en/docs/installation/kubernetes-operator-security.md
```diff
@@ -0,0 +1,79 @@
...
+    fsGroup: 65532
+    seccompProfile:
+      type: RuntimeDefault
+
+  containerSecurityContext:
+    allowPrivilegeEscalation: false
+    readOnlyRootFilesystem: true
+    capabilities:
+      drop:
+      - ALL
```
mtardy: if we decide on going non root for everyone, maybe we should put this in a separate patch as well and apply this by default also (always with the ability to override it).

mtardy:
> Anything else that needs to be changed? @mtardy

sorry this is taking significant time and going back and forth a little bit but I think it's a really nice change :)

calghar:
> A few remarks, I didn't initially understand that setting the `USER` in the image will start any non-overriding user deployment with the user.
> 
> We could do this in two steps by having the backward compatible thing or just move everyone toward non root operator for next minor version. I think it should be fine? In this case we could default to non-root and show how to get back to root user with a config example to get the previous situation for people that care about that.
> 
> Like a default well configured security context (you don't need to specify the user and group if the image has been built with the USER thing it seems) and the possibility to override this with the previous config.
> 
> wdyt?

Yeah, given the docker set user can be overwritten by k8s at runtime, we can set up the docker image to use a non-root user and then provide a patch to say "hey if you wish to use a root user you can override the settings in the helm chart using this config", while keeping the helm chart values as I have currently set them up. This ensures that while we force a non-root user we also provide options to override it if needed by someone. For consistency it would still be better to use the same user and group in helm since K8s will anyways override it, should a root config be applied.

Dockerfile.operator
calghar: This work although might be better to go with the non-root image and then also provide non-root defaults for helm imho.

calghar: Also makes it easier for people who are worried about compliance etc. like the person who raised the issue originally

calghar:
> > Anything else that needs to be changed? @mtardy
> 
> sorry this is taking significant time and going back and forth a little bit but I think it's a really nice change :)

No problemo. 😄

calghar:
> > A few remarks, I didn't initially understand that setting the `USER` in the image will start any non-overriding user deployment with the user.
> > We could do this in two steps by having the backward compatible thing or just move everyone toward non root operator for next minor version. I think it should be fine? In this case we could default to non-root and show how to get back to root user with a config example to get the previous situation for people that care about that.
> > Like a default well configured security context (you don't need to specify the user and group if the image has been built with the USER thing it seems) and the possibility to override this with the previous config.
> > wdyt?
> 
> Yeah, given the docker set user can be overwritten by k8s at runtime, we can set up the docker image to use a non-root user and then provide a patch to say "hey if you wish to use a root user you can override the settings in the helm chart using this config", while keeping the helm chart values as I have currently set them up. This ensures that while we force a non-root user we also provide options to override it if needed by someone. For consistency it would still be better to use the same user and group in helm since K8s will anyways override it, should a root config be applied.

So i gave it some thought and I am currently not sure which approach is the best. If one anyways has the ability to override the non-root user set via docker then wouldn't it be better to have this configuration in helm. So set a boolean value which can determine wether to use non-root or root user instead of going with that rather involved method in Docker. I would prefer to just add this change to the helm chart and revert the changes to Docker if that works for you. @mtardy

jrfastab:
@mtardy ping just to keep the discussion going or we merge :) Thanks for all the work @calghar

docs/content/en/docs/installation/kubernetes-operator-security.md
mtardy, line 1: Thank you, I think the docs is really nice. Could we put this piece of docs under the "Configuration" page in installation? Like as a section equal to "Enable Process Credentials"? As a "##" section and then everything under is "###" or "####".

Dockerfile.operator
mtardy: I can remove this change for you if you want.

calghar:
> > So i gave it some thought and I am currently not sure which approach is the best. If one anyways has the ability to override the non-root user set via docker then wouldn't it be better to have this configuration in helm. So set a boolean value which can determine wether to use non-root or root user instead of going with that rather involved method in Docker. I would prefer to just add this change to the helm chart and revert the changes to Docker if that works for you. @mtardy
> 
> Yeah either way is fine. Not sure which is best, so I think we could merge it this like that :)!
> 
> Just a small docs nit and we can merge.
> 
> > just to keep the discussion going or we merge :)
> 
> agree!!

Done with the doc changes 😊. 
Not sure what you meant with the Dockerfile.operator comment, i am using the old version now 😅

mtardy (approved):
I removed the dockerfile change and did just a few modifications to the docs part (and rebased to remove the TracingPolicy unrelated generated changes), let's merge this when green! Thanks a lot for the work :)

calghar:
> I removed the dockerfile change and did just a few modifications to the docs part (and rebased to remove the TracingPolicy unrelated generated changes), let's merge this when green! Thanks a lot for the work :)

Awesome. Thanks for the feedback, it was fun. Onwards!

PR #3918, opened 2025-07-20
refactor: Clean up composite literals and nil comparisons

yeongjukang:
### Description
While exploring the Tetragon codebase, gopls returned 105 issues. I'm addressing 104 of these warnings.

- Category 1: Redundant Types in Composite Literals (104 Warnings)
  - The first category consists of 104 warnings for 'redundant type from array, slice, or map composite literal'. While these might have been kept for readability, I believe removing them will improve the overall code quality and make the codebase more approachable for new contributors by eliminating these gopls warnings.

- Category 2: Impossible Condition: nil != nil (genericlsm.go)
  - The second category is the 'impossible condition: nil != nil' warning in genericlsm.go. This error occurs because new err variables are declared within the blocks of the preceding two for loops. When an error exists, there's no need to return it, and the err variable within the createGenericLsmSensor function also appears to be unnecessary, so I've removed it.

- Remaining Issue
  - The single remaining issue has appropriate comments, so I kept it as it is. (const genericTP_OutputSize)

Thank you for your review.


mtardy (approved):
Hey, could you put the PR message in the commit description? Thank you!

yeongjukang:
@mtardy Thank you for the review, I updated it!

PR #3922, opened 2025-07-21
tetra: Add refmap command

olsajiri:
wip

### DIFF
--- a/pkg/sensors/program/loader_linux.go
+++ b/pkg/sensors/program/loader_linux.go
@@ -997,7 +997,7 @@ func doLoadProgram(
 		if pm, ok := load.PinMap[mapLoad.Name]; ok {
 			pinPath = pm.PinPath
 		}
-		if m, ok := coll.Maps[mapLoad.Name]; ok {
+		if m, ok := pinnedMaps[mapLoad.Name]; ok {
 			if err := mapLoad.Load(m, pinPath); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/tetra/debug/refmap.go
```diff
@@ -0,0 +1,48 @@
...
+
+				if idx != 0 {
+					fmt.Printf("\n")
+				}
+				fmt.Printf("%s: ", prog.Name)
+				for m, _ := range refMaps {
+					fmt.Printf("%s ", m)
+				}
+				idx += 1
+			}
```
mtardy: nit
```suggestion
			for idx, prog := range spec.Programs {
				refMaps := make(map[string]bool)
				for _, inst := range prog.Instructions {
					if inst.Reference() != "" {
						refMaps[inst.Reference()] = true
					}
				}

				if idx != 0 {
					fmt.Printf("\n")
				}
				fmt.Printf("%s: ", prog.Name)
				for m, _ := range refMaps {
					fmt.Printf("%s ", m)
				}
			}
```

mtardy (commented):
looks like a nice patch, would be useful indeed :)

cmd/tetra/debug/refmap.go
olsajiri: hum, `spec.Programs` is `map[string]*ProgramSpec `, so can't replace that `int` I need

mtardy: ah oups sorry

PR #3935, opened 2025-07-24
fix: Force remove tetragon-clang container for the tetragon-bpf target in case it's still running.

acamatcisco:
### Description
After `make clean`, I got an error when trying to rebuild the tetragon-bpf target:

```
acam@lima-default:~/src/hubble-fgs/modules/tetragon-oss$ make -j3 tetragon-bpf
docker rm tetragon-clang || true
Error response from daemon: cannot remove container "tetragon-clang": container is running: stop the container before removing or force remove
```

This PR contains a simple change to force-remove the tetragon-clang container if it's already running.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,9 +136,9 @@ tetragon-bpf-local:
 
 .PHONY: tetragon-bpf-container
 tetragon-bpf-container:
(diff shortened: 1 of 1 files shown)


mtardy (approved):
sure

PR #3938, opened 2025-07-25
bpf: Use bpf_ktime_get_boot_ns when available

tpapagian:
Fixes 

### Changelog

```release-note
```


kkourt (approved):
Thanks!

bpf/process/bpf_execve_event.c
```diff
@@ -212,7 +212,11 @@ event_execve(struct exec_ctx_struct *ctx)
 	p->pid = pid >> 32;
 	p->tid = (__u32)pid;
 	p->nspid = get_task_pid_vnr_curr();
+#ifdef __LARGE_BPF_PROG
+	p->ktime = (bpf_core_type_exists(btf_bpf_ktime_get_boot_ns)) ? (ktime_get_boot_ns()) : (ktime_get_ns());
```
olsajiri: should we use this in other places as well then? generic sensors and fork/exit?

tpapagian: I believe that this makes sense. I will update.

mtardy (approved):
lgtm! Agree with Jiri's comment as well :)

bpf/process/bpf_execve_event.c
tpapagian: Done!

PR #3947, opened 2025-07-28
bpf: fix issue with multiple inactive selectors

kkourt:
The current code only tail calls into the check for the next selector if it is active. However, we can have more than two selector which means that we could be in a situation where: the first two are 0 and the third is active.

The commit changes the code to tail call into the next active selector.

There a potential optimization for this, were we save the index of the active selectors in an array and just jump to the next one. This is left as future work.

TODO:
 - [x] Add a test for this case
 - [x] Add docs that specify the correct (expected) behavior

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -4,11 +4,13 @@ weight: 3
 description: "Perform in-kernel BPF filtering and actions on events"
 ---
 
-Selectors are a way to perform in-kernel BPF filtering on the events to
-export, or on the events on which to apply an action.
+Selectors enable per-hook in-kernel BPF filtering and actions. Each selector defines a set of
+filters as well as a set of (optional) actions to be performed if the selector filters match. Each
+hook can contain up to 5 selectors. If no selectors are defined on a hook, the default action
+(`Post`, i.e., post an event) will be used.
 
-A `TracingPolicy` can contain from 0 to 5 selectors. A selector is composed of
-1 or more filters. The available filters are the following:
+
+Each selector comprises a set of filters:
 - [`matchArgs`](#arguments-filter): filter on the value of arguments.
 - [`matchReturnArgs`](#return-args-filter): filter on the return value.
 - [`matchPIDs`](#pids-filter): filter on PID.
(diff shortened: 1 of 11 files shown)


kkourt:
The checkpatch failure is:
```
   CHECK:CAMELCASE: Avoid CamelCase: <_Static_assert>
  #27: FILE: bpf/process/generic_calls.h:1091:
  +_Static_assert(5 == MAX_SELECTORS, "update selidx_next()");
```

I don't think we have a non camel-case wrapper for this AFAIK, and would be nice to keep this in case we ever opt to change MAX_SELECTORS.

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -17,9 +19,42 @@ A `TracingPolicy` can contain from 0 to 5 selectors. A selector is composed of
...
 - [`matchNamespaceChanges`](#namespace-changes-filter): filter on Linux namespaces changes.
 - [`matchCapabilityChanges`](#capability-changes-filter): filter on Linux capabilities changes.
+
+And a set of actions that will be performed if the specified filters match:
 - [`matchActions`](#actions-filter): apply an action on selector matching.
 - [`matchReturnActions`](#return-actions-filter): apply an action on return selector matching.
 
+For a selector to match, all of its filters must match (AND operator). If a selector defines no
+filters, it matches. If multiple selectors are defined in a hook, only the action defined in the
+first matching selector will be applied (short-circuited OR). If a selector defines no action, the
```
mtardy, line 29: nit: I would maybe put this in a note or caution block maybe?

mtardy (approved):
docs changes look good, I would wait for Jiri's review on the BPF patch!

bpf/process/generic_calls.h
```diff
@@ -1088,6 +1088,46 @@ FUNC_INLINE int filter_args(struct msg_generic_kprobe *e, int selidx, bool is_en
...
+	case 4:
+		if (e->sel.active[idx & MAX_SELECTORS_MASK])
+			break;
+		idx++;
+		fallthrough;
+	case 5:
+		if (e->sel.active[idx & MAX_SELECTORS_MASK])
+			break;
+		idx++;
+	}
```
mtardy, line 1126: your switch case looks like a loop no?
```suggestion
	for (int i = idx; i <= 5; i++) {
		if (e->sel.active[idx & MAX_SELECTORS_MASK])
			break;
		idx++;
	}
```

kkourt: Indeed, but I wasn't sure how well the compiler would unroll it given the (int i = idx), so I just unrolled it on my own.

mtardy: oh okay nevermind :)

PR #3948, opened 2025-07-28
Optimize Kprobe Rate Limit Test Performance

### LINKED ISSUE #2690: tests: `pkg.sensors.tracing:TestKprobeRateLimit` and `pkg.sensors.tracing:TestKprobeNoRateLimit` could be optimized
Those tests use the checker to check that events are missing: this is suboptimal as it leads to an unnecessarily long test as the checker will retry multiple times to have the missing events. I see two options:
- Use the perfring test framework to check quickly and efficiently for missing events
- Find a way to disable the retry mechanism and limit the test in time (this is less reliable)

AritraDey-Dev:
Thanks for contributing! Please ensure your pull request adheres to the following guidelines:

- [x] All commits contain a well written commit message and are signed-off (see [Submitting a pull request](https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/)).
- [x] All code is covered by unit and/or end-to-end tests where feasible.
- [x] All generated files are updated if needed (see [Making changes](https://tetragon.io/docs/contribution-guide/making-changes/)).
- [x] Provide a title or release-note blurb suitable for the release notes (see [guidelines](https://tetragon.io/docs/contribution-guide/release-notes/#release-note-blurb-in-pr)).
- [ ] Update documentation and write an upgrade note if needed (see [guidelines](https://tetragon.io/docs/contribution-guide/release-notes/#upgrade-notes)).

### Description

optimize kprobe rate limit tests with perfring framework

Fixes #2690 
### Changelog

```
Optimized kprobe rate limit tests with perfring framework
```

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/cilium/ebpf/asm"
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
+	"github.com/cilium/tetragon/pkg/api/tracingapi"
 	"github.com/cilium/tetragon/pkg/arch"
 	"github.com/cilium/tetragon/pkg/bpf"
 	"github.com/cilium/tetragon/pkg/config"
@@ -4734,6 +4735,10 @@ spec:
 }
 
 func testKprobeRateLimit(t *testing.T, rateLimit bool) {
+	testutils.CaptureLog(t, logger.GetLogger())
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
 	hook := `apiVersion: cilium.io/v1alpha1
 kind: TracingPolicy
 metadata:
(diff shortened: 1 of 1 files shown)


pkg/sensors/tracing/kprobe_test.go
```diff
@@ -4771,68 +4776,75 @@ spec:
...
-	// Generate 5 datagrams
-	socket, err := net.Dial("udp", "127.0.0.1:9468")
-	if err != nil {
-		t.Fatalf("failed dialing socket: %s", err)
-	}
+	ops := func() {
+		server := "nc.openbsd"
+		cmdServer := exec.Command(server, "-unvlp", "9468", "-s", "127.0.0.1")
+		require.NoError(t, cmdServer.Start())
+		time.Sleep(1 * time.Second)
```
olsajiri, line 4803: is this sleep needed? any chance the dial call below would wait long enough to connect?

olsajiri (commented):
lgtm, please add changelog description (text from the related issue would be enough)

pkg/sensors/tracing/kprobe_test.go
AritraDey-Dev: As UDP `net.Dial() `succeeds immediately even if no server is listening, so without the sleep, packets would be sent before` nc.openbsd` finishes starting up

AritraDey-Dev:
@mtardy, from my testing it took around 40–42 seconds without this patch, and with the patch it only took around 3–5 seconds, approximately 20x faster.

pkg/sensors/tracing/kprobe_test.go
mtardy: yeah that would be nice to have a proper "ready" probe because these things tend to create unnecessary slow and potentially flaky tests but the improvement is worth this, especially this is already in the old code so...

mtardy (approved):
Okay! Changes look good thank you!

PR #3952, opened 2025-07-29
bpf: remove unused func UpdateElementFromPointers

tklauser:
It's a leftover of the `bpf.Map` interface that was removed in commit 0143c1c466ae ("tetragon: Remove bpf.Map module").


pkg/bpf/perf_windows.go
```diff
@@ -1,24 +0,0 @@
...
-
-import (
-	"fmt"
-
-	"golang.org/x/sys/windows"
-)
-
-var (
-	EbpfApi = windows.NewLazyDLL("ebpfapi.dll")
-	BPF     = EbpfApi.NewProc("bpf")
```
mtardy, line 14: looks like you have some windows stuff that we need somehow

tklauser: Whoopsie, fixed.

PR #3956, opened 2025-07-29
fix: Refactor SIZEOF_EVENT constant to not be a hard-coded value.

acamatcisco:
### Description
This is less brittle in the case where the struct's size is later modified.

### DIFF
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -95,16 +95,16 @@
  *
  * Phew all clear now?
  */
-#define CWD_MAX	     4096
-#define BUFFER	     1024
-#define SIZEOF_EVENT 56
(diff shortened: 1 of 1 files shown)


bpf/lib/process.h
```diff
@@ -95,16 +95,16 @@
  *
  * Phew all clear now?
  */
-#define CWD_MAX	     4096
-#define BUFFER	     1024
-#define SIZEOF_EVENT 56
+#define CWD_MAX		   4096
+#define BUFFER		   1024
```
acamatcisco, line 99: Formatting job was failing without these whitespace changes.

tpapagian (approved):

PR #3957, opened 2025-07-29
Fix a typo in kubectl in the runtime hook documentation

uhlhosting:
Fixes 

### Description

Line 200 fixed broken type kubecl to kubectl


mtardy (approved):
Indeed thanks

PR #3962, opened 2025-07-30
rthooks: Log container ID as a key-value pair

michi-covalent:
Move the container ID out of log messages so that they aggregate better.

### DIFF
--- a/pkg/policyfilter/rthooks/rthooks.go
+++ b/pkg/policyfilter/rthooks/rthooks.go
@@ -5,7 +5,6 @@ package rthooks
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


kkourt (approved):

sayboras (approved):

sayboras:
Nice, going forward we probably should follow key-value pair practice in logging.

mtardy:
> Nice, going forward we probably should follow key-value pair practice in logging.

Agree, one way to improve would be to search for `printf` inside logging statement and see if we should change them. Now the complicated thing is to keep something consistent when displaying field, like `container-id` vs `containerID` vs whatever so that parsing the logs is not too tedious!

PR #3974, opened 2025-07-31
fix: Refactor length calculation logic when reading process args.

acamatcisco:
### Description
There were 3 bugs here:

1. An off-by-one in computing the NUL termination bytes offset
2. If arg length >= 256, we would read the wrong number of kernel bytes
3. If arg length exactly 254 or 255, we would incorrectly overwrite the
first and/or second bytes of the args array with a NUL byte

Also added a comment about the reason for the double null termination.

Signed-off-by: Aaron Campbell <acam@cisco.com>

### Changelog

```release-note
More robust process argument parsing logic.
```

### DIFF
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -333,7 +333,7 @@ execve_send(void *ctx __arg_ctx)
 		/* zero out previous paths in ->bin */
 		binary_reset(&curr->bin);
 #ifdef __LARGE_BPF_PROG
-		__u32 nullone, nulltwo, off, len;
+		__u32 off, len;
 
 		// read from proc exe stored at execve time
 		if (event->exe.len <= BINARY_PATH_MAX_LEN) {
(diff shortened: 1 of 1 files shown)


bpf/process/bpf_execve_event.c
```diff
@@ -348,13 +348,16 @@ execve_send(void *ctx __arg_ctx)
...
 		probe_read(curr->bin.args, len, (char *)&event->process + off);
 
-		nullone = len + 1;
-		nulltwo = len + 2;
-		curr->bin.args[nullone & 0xff] = 0x00; // null terminate string
-		curr->bin.args[nulltwo & 0xff] = 0x00; // null terminate string
+		nullone = len;
+		nulltwo = len + 1;
+		curr->bin.args[nullone] = 0x00; // null terminate string
+		curr->bin.args[nulltwo] = 0x00; // null terminate string
```
olsajiri: it seems ok, but I'm scratching my head why we need two null bytes at the end @jrfastab

acamatcisco: @olsajiri AFAICT, there is one NUL byte between each argv element, so this technique let's you find the terminal point by scanning for two. We do that at `pkg/model/server/server.go:116` in GetProcessMap() for example:

```
m := bytes.Index(value.Args[:], []byte{0x00, 0x00})
```

Probably I should add a comment for this. :-) [EDIT: Done.]

Issue #3836, opened 2025-06-19
Support for sockaddr_un struct


genie-zs:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

No

### Describe the feature you would like

To be able to get unix socket file path (sun_path), we need to add support for the [sockaddr_un](https://elixir.bootlin.com/linux/v5.15.185/source/include/uapi/linux/un.h#L9) struct.

Recently, tetragon received support for the struct sockaddr in https://github.com/cilium/tetragon/pull/3358, but to fully support UNIX sockets, we also need to support the sockaddr_un structure. 

That will allow us to develop policies to detect and filter operations on specific UNIX sockets such as `docker.sock` and `containerd.sock` by hooking into functions such as  [unix_stream_connect()](https://elixir.bootlin.com/linux/v5.15.185/source/net/unix/af_unix.c#L1310) .

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

sayboras:
This should be done as mentioned in #3980

PR #3980, opened 2025-08-04
feat: add support for AF_UNIX socket path via sockaddr_un

AshishNaware:
Fixes 
#3836 

### Description

This PR adds the support for  [sockaddr_un](https://elixir.bootlin.com/linux/v5.15.185/source/include/uapi/linux/un.h#L9) alongside the existing sockaddr_in logic. While Tetragon previously added support for sockaddr_in (#3358), this enhancement extends visibility to UNIX domain sockets, enabling detection and filtering on socket paths such as:

```
/run/containerd/containerd.sock
/var/run/nscd/socket
/dev/log
```
Selectors on `sockaddr_un` arguments support `Equal`, `NotEqual`, `Prefix`, `NotPrefix` on the path, and `Family` 
Note: This feature is available on linux 5.11 or newer.

### Changelog

```release-note
- Added decoding support for `sockaddr_un` in `security_socket_connect` events.
  - Tetragon can now extract and emit UNIX socket paths from AF_UNIX addresses.
  - Supports observability into socket-based interactions like containerd.sock and docker.sock.
  - Selectors support Family and Equal, NotEqual, Prefix, NotPrefix on the path
  - Requires linux 5.11+
```


AshishNaware:
This PR is still work in progress as I'm currently addressing the test failures, but I’d appreciate early feedback on the overall struct parsing and decoding approach.

bpf/process/types/sockaddr.h
```diff
@@ -13,34 +13,52 @@ struct sockaddr_in_type {
...
+		probe_read(&event->in.sin_port, sizeof(event->in.sin_port), _(&ipv4->sin_port));
 		break;
 	case AF_INET6:
-		probe_read(&event->sin_addr, sizeof(event->sin_addr), _(&ipv6->sin6_addr));
-		probe_read(&event->sin_port, sizeof(event->sin_port), _(&ipv6->sin6_port));
+		probe_read(&event->in.sin_addr, sizeof(event->in.sin_addr), _(&ipv6->sin6_addr));
+		probe_read(&event->in.sin_port, sizeof(event->in.sin_port), _(&ipv6->sin6_port));
+		break;
+	case AF_UNIX:
+		probe_read_str(event->un.sun_path, sizeof(event->un.sun_path), un->sun_path);
```
kevsecurity: This doesn't feel safe as sockaddrs are often user-controlled, and the sun_path might be 108 chars long and not null-terminated (if I've understood correctly).

Also, maybe we should support abstract sockets where the first byte is 0x00, and path name can contain null bytes?

pkg/api/tracingapi/client_kprobe.go
```diff
@@ -255,19 +255,31 @@ func (m MsgGenericKprobeArgSkb) IsReturnArg() bool {
...
+type SockaddrUn struct {
+	SunFamily uint16
+	SunPath   [108]byte
+}
+
+type SockaddrEvent struct {
+	Family uint16
+
+	In SockaddrIn
+	Un SockaddrUn
```
kevsecurity: This doesn't appear to match the struct sockaddr_event in sockaddr.h as there In and Un are unioned, but here appear sequential?

pkg/sensors/tracing/args_linux.go
```diff
@@ -263,18 +264,43 @@ func getArg(r *bytes.Reader, a argPrinter) api.MsgGenericKprobeArg {
...
+			arg.SinAddr = network.GetIP(in.SinAddr, family).String()
+			arg.SinPort = uint32(in.SinPort)
+
+		case unixsys.AF_UNIX:
+			var un api.SockaddrUn
+			if err := binary.Read(r, binary.LittleEndian, &un); err != nil {
+				logger.GetLogger().Warn("failed to read sockaddr_un", logfields.Error, err)
+				return nil
+			}
+			arg.Path = string(bytes.Trim(un.SunPath[:], "\x00"))
```
kevsecurity: This feels a bit brittle. I feel that we could add a new type (gt.GenericSockaddrUnType) and rename the existing gt.GenericSockaddrType to gt.GenericSockaddrInType, then we can just binary.Read() directly into appropriate matching data structures in the appropriate cases.

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -970,17 +972,42 @@ func handleMsgGenericTracepoint(
 			unix.Args = append(unix.Args, arg)
 
 		case gt.GenericSockaddrType:
-			var address tracingapi.MsgGenericKprobeSockaddr
+			var family uint16
+			if err := binary.Read(r, binary.LittleEndian, &family); err != nil {
```
kevsecurity: Same comment as per kprobe.

api/v1/tetragon/tetragon.proto
```diff
@@ -355,11 +355,24 @@ message KprobeSkb {
   string family = 13;
 }
 
+message SockaddrIn {
+  uint32 sin_port = 1;
+  uint64 sin_addr_0 = 2;
+  uint64 sin_addr_1 = 3;
+}
```
kevsecurity: Why are we adding this?

api/v1/tetragon/tetragon.proto
```diff
@@ -355,11 +355,24 @@ message KprobeSkb {
...
+
 message KprobeSockaddr {
   string family = 1;
   string addr = 2;
   uint32 port = 3;
   string path = 4;
+
+  SockaddrIn in = 5;
+  SockaddrUn un = 6;
```
kevsecurity: Why are we adding these?

kevsecurity (changes requested):
Thanks for adding this. Overall I think it makes sense, but I'm concerned about how we interpret the buffer in the BPF code (let's be more robust here), and I think we should probably add a new message type (from BPF to user space) instead of overloading the one supporting sockaddr_in/6.

kevsecurity:
Also, as well as passing existing tests, please could you add some tests for the new functionality?

AshishNaware:
Hi @kevsecurity - I’ve marked this PR ready for review. I updated the parsing logic per your earlier feedback and added tests. This is my first PR touching Cilium eBPF, so please call out any edge cases I may have missed. Thanks!

olsajiri (changes requested):
please separate the change into several commits
typically at least the generated changes go in separate commit, but also single logic change should go in separated commit if possible,  thanks

pkg/selectors/kernel.go
```diff
@@ -856,8 +856,8 @@ func ParseMatchArg(k *KernelSelectorState, arg *v1alpha1.ArgSelector, sig []v1al
 			return fmt.Errorf("writePostfixStrings error: %w", err)
 		}
 	case SelectorOpSport, SelectorOpDport, SelectorOpNotSport, SelectorOpNotDport, SelectorOpProtocol, SelectorOpFamily, SelectorOpState:
-		if ty != gt.GenericSockType && ty != gt.GenericSkbType && ty != gt.GenericSockaddrType && ty != gt.GenericSocketType {
-			return errors.New("sock/socket/skb/sockaddr operators specified for non-sock/socket/skb/sockaddr type")
+		if ty != gt.GenericSockType && ty != gt.GenericSkbType && ty != gt.GenericSockaddrType && ty != gt.GenericSocketType && ty != gt.GenericSockaddrUnType {
+			return errors.New("sock/socket/skb/sockaddr/sockaddr_un operators specified for non-sock/socket/skb/sockaddr/sockaddr_un type")
```
kevsecurity, line 1026: Until we have selectors for sockaddr_un, I think we can leave it out of this if-statement.

pkg/sensors/tracing/args_linux.go
```diff
@@ -276,6 +276,19 @@ func getArg(r *bytes.Reader, a argPrinter) api.MsgGenericKprobeArg {
...
+		var arg api.MsgGenericKprobeArgSockaddrUn
+
+		err := binary.Read(r, binary.LittleEndian, &sockaddr)
+		if err != nil {
+			logger.GetLogger().Warn("sockaddrun type err", logfields.Error, err)
+		}
+
+		arg.Index = uint64(a.index)
+		arg.Family = sockaddr.Family
+		arg.Path = string(NormalizeSockaddrUnPath(sockaddr.Path[:]))
```
kevsecurity: It is probably worth noting that this only supports named unix sockets, and we'd need a bit more code to handle abstract sockets, and to separate them from unnamed sockets. That can be a follow-on patch though.

kevsecurity: My bad; didn't read on!

pkg/sensors/tracing/kprobe_net_test.go
```diff
@@ -1326,3 +1376,148 @@ spec:
...
+    - matchArgs:
+      - index: 0
+        operator: "Family"
+        values: ["AF_UNIX"]
+      - index: 1
+        operator: "Family"
+        values: ["AF_UNIX"]
+      - index: 1
+        operator: "Prefix"
+        values: ["/"]
```
kevsecurity: The selectors matching index 1 will match everything because there is no specific matching in bpf/process/types/basic.h.

kevsecurity (changes requested):
It looks good - thanks! I agree that it would be better divided into a few commits. Did you want to add a case to the switch in selector_arg_offset() (in basic.h) to filter the sockaddr_un parameters (the selectors in your test policy implies you might do)? Let me know if you want help or a chat about that. But otherwise, other than dividing it up, it is looking good.

kevsecurity:
@AshishNaware looks like your new test is failing on a few older kernels. It should be possible to use LVH to download and run up VMs with the same kernels to run the tests locally. Let me know if you need help with that.

AshishNaware:
Hi. I haven’t had much bandwidth recently, but I plan to work through the comments and push the changes within the next two weeks. thanks

AshishNaware:
I’ve addressed the review comments, and the filters are now working as expected in my local environment. With the current setup, we can filter socket paths by prefix and family. I think this could be extended in the future to support additional path filters (e.g., Equal, NotEqual, Postfix), but I’d prefer to keep that work for a separate PR.

At the moment, I’m working through some pipeline failures and have reached out to @kevsecurity on Slack for guidance.

mtardy (commented):
> I’ve addressed the review comments, and the filters are now working as expected in my local environment. With the current setup, we can filter socket paths by prefix and family. I think this could be extended in the future to support additional path filters (e.g., Equal, NotEqual, Postfix), but I’d prefer to keep that work for a separate PR.
>
> At the moment, I’m working through some pipeline failures and have reached out to @kevsecurity on Slack for guidance.

Kev will be soon available for review (I would expect next week).

In the meantime, you should be able to now rebase your PR, we fixed something on the generated TracingPolicy docs.

Also if you rebase, if you can be a little bit more specific in your patches titles that would help for when we want to merge this "added tests" or "pr checks" is not sufficient as title/description. Maybe you want to squash the "pr checks" one into other commits above btw.

pkg/selectors/kernel.go
AshishNaware: resolving this comment as i have added changes to filter out values using selectors.

pkg/sensors/tracing/kprobe_net_test.go
AshishNaware: i have added changes to filter out those values using selectors.

AshishNaware:
@mtardy / @kevsecurity  - The pipeline is green now. I’ve gated the feature behind the `__LARGE_BPF_PROG` macro to prevent adding excessive instructions on the 4.19 kernel. The change is ready for review.

bpf/process/types/basic.h
```diff
@@ -1108,6 +1121,66 @@ filter_inet(struct selector_arg_filter *filter, char *args)
...
+	struct sockaddr_un_type *address = (struct sockaddr_un_type *)args;
+	char path_buffer[UNIX_PATH_MAX];
+	int err;
+
+	switch (filter->type) {
+	case sockaddr_un_type:
+		break;
+	default:
+		return 0;
+	}
```
kevsecurity: We only call filter_sockaddr_un() if filter->type == sockaddr_un_type, so this is not needed. The similar construction in filter_inet() is there because we call that with a handle of different types.

bpf/process/types/sockaddr_un.h
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: (GPL-2.0-only OR BSD-2-Clause)
+/* Copyright Authors of Cilium */
+
+#ifndef __SOCKADDR_UN_H__
+#define __SOCKADDR_UN_H__
+
+#define UNIX_PATH_MAX 108
```
kevsecurity: This name could be more descriptive, as I think it will confuse people as it is. Maybe it could be UNIX_SOCKET_PATH_MAX instead?

bpf/process/types/basic.h
```diff
@@ -1108,6 +1121,66 @@ filter_inet(struct selector_arg_filter *filter, char *args)
...
+	switch (filter->op) {
+	case op_filter_str_prefix: {
+		uint path_len = 0;
+		char *path_to_check = path_buffer;
+
+		if (path_buffer[0] == '\0') {
+			// ignore leading null byte for abstract sockets
+			path_to_check = &path_buffer[1];
+			for (int i = 1; i < UNIX_PATH_MAX; i++) {
+				if (path_buffer[i] == '\0') {
```
kevsecurity: Abstract socket names could include null bytes, according to the man page. But if we allow that, then we won't be able to do string matching (as they won't be C strings). I think your approach is the best we can easily do for now. Making this note to remember that we might need to revisit this.

bpf/process/types/basic.h
```diff
@@ -1108,6 +1121,66 @@ filter_inet(struct selector_arg_filter *filter, char *args)
...
+		break;
+	default:
+		return 0;
+	}
+
+	err = probe_read(path_buffer, UNIX_PATH_MAX, address->sun_path);
+	if (err != 0)
+		return 0;
+
+	switch (filter->op) {
```
kevsecurity, line 1187: Feels like we should have a case for equals at least. Maybe take a look at filter_char_buf() and implement those cases here? In which case, probably do path checking and counting first.

bpf/process/types/basic.h
kevsecurity: I think we should include a flag in the sockaddr_un_type to indicate if the socket is abstract.

bpf/process/types/basic.h
```diff
@@ -1108,6 +1121,66 @@ filter_inet(struct selector_arg_filter *filter, char *args)
...
+		} else {
+			// process the full path for filesystem sockets
+			path_to_check = path_buffer;
+			for (int i = 0; i < UNIX_PATH_MAX; i++) {
+				if (path_buffer[i] == '\0') {
+					path_len = i;
+					break;
+				}
+			}
+			if (path_len == 0 && path_buffer[0] != '\0')
```
kevsecurity: path_buffer[0] cannot equal '\0' otherwise we'd be in the other branch of the if.

pkg/api/tracingapi/client_kprobe.go
```diff
@@ -262,6 +262,11 @@ type MsgGenericKprobeSockaddr struct {
 	SinAddr   [2]uint64
 }
 
+type MsgGenericKprobeSockaddrUn struct {
+	Family uint16
```
kevsecurity: For sockaddr_un, Family is always AF_UNIX (according to the man page). Do we need to record it? In inet we do because it tells us whether we're IPv4 or IPv6 (or potentially something we don't understand).

pkg/selectors/kernel.go
kevsecurity: Okay, I think this needs a nested if within the if to return an error if sockaddr_un is being used with OpSport, OpDport, etc. Really, it should only be string matchers (if we drop family) so maybe this shouldn't be in this if at all?

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -1083,3 +1097,35 @@ func handleMsgGenericTracepoint(
 func (t *observerTracepointSensor) LoadProbe(args sensors.LoadProbeArgs) error {
 	return LoadGenericTracepointSensor(args.BPFDir, args.Load, args.Maps, args.Verbose)
 }
+
+// NormalizeSockaddrUnPath returns a display-friendly byte slice:
+// "@name" for abstract UNIX sockets (trimming trailing NUL padding),
+// or the pathname up to the first NUL for filesystem sockets.
+func NormalizeSockaddrUnPath(raw []byte) []byte {
```
kevsecurity: If we had a flag for abstract sockets then we could use that to spot an abstract socket.

kevsecurity (changes requested):
This is looking really good. I've left a handful of comments to look at. Happy to discuss them. A bit of a tweak and this can be merged.

genie-zs:
@AshishNaware Any luck that it will land any time soon?

AshishNaware:
> @AshishNaware Any luck that it will land any time soon?

sorry for the delay. I will work on this over the coming weekend.

AshishNaware:
Addressed review comments with below changes-

- Renamed variables and droped redundant checks.
- Modified sockaddr_un_type struct with path length and abstract-socket flag.
- Added support to filter path equal, not equal, prefix, not prefix and family along with tests.

**Pipeline failures:**
e2e tests are successful locally. is there any known pipeline issue? can we retrigger it?

```
go list ./tests/e2e/tests/... | xargs -Ipkg go test  -gcflags= -timeout 20m -failfast -cover pkg  -fail-fast -tetragon.helm.set tetragon.image.override=""cilium/tetragon:latest"" -tetragon.helm.set tetragonOperator.image.override=""cilium/tetragon-operator:latest"" -tetragon.helm.url="" -tetragon.helm.chart="/home/ashish/github.com/cilium/tetragon/install/kubernetes/tetragon" ""
?   	github.com/cilium/tetragon/tests/e2e/tests	[no test files]
ok  	github.com/cilium/tetragon/tests/e2e/tests/labels	289.880s	coverage: [no statements]
ok  	github.com/cilium/tetragon/tests/e2e/tests/policyfilter	129.138s	coverage: [no statements]
ok  	github.com/cilium/tetragon/tests/e2e/tests/skeleton	54.043s	coverage: [no statements]
```

Tests failing on kernels 4.1 seems unrelated to my changes. Any guidance on this?

kevsecurity:
> **Pipeline failures:** e2e tests are successful locally. is there any known pipeline issue? can we retrigger it?

I think you might need a rebase to pick up some CI changes.

> Tests failing on kernels 4.1 seems unrelated to my changes. Any guidance on this?

Without looking properly, my guess is that the generic kprobe program no longer loads on 4.19, possibly because it has more than 4096 instructions maybe? Does it work locally?

mtardy:
hey @AshishNaware when you think this is ready, do not forget to click those for us to know about it

image

AshishNaware:
Hi @mtardy I am currently observing some failures on older kernels, especially regarding unsafe memory access - `value -2147483648 makes map_value pointer be out of bounds` . Wondering if I could use some help from maintainers to debug the issue?

pkg/sensors/tracing/kprobe_net_test.go
```diff
@@ -1326,3 +1377,314 @@ spec:
 	err := jsonchecker.JsonTestCheckExpectWithKeep(suite.T(), checker, false, false)
 	suite.Require().NoError(err)
 }
+
+func (suite *KprobeNet) TestKprobeSocketAndSockaddrUn() {
+	if !config.EnableLargeProgs() {
+		suite.T().Skip("skipping test as older kernels may not support Prefix selector argument processing")
+	}
+	// Kernel <5.5 lacks bpf_probe_read_user; see TestKprobeSockaddrUnPathEquals.
+	if !kernels.MinKernelVersion("5.5") {
```
kevsecurity: Your last commit gates the functionality at v5.11 and the tests are failing with missing sockaddr_un arguments on kernel v5.10 (```logcapture.go:24: time=2026-03-20T15:54:01.165Z level=WARN msg="sockaddrun type err" error="unexpected EOF"```).

I think the solution might be to gate your tests at v5.11. :D

kevsecurity:
👋 Hey, this is looking good. I left a comment that you're gating the functionality at 5.11 but testing it from 5.5, probably causing the failures on the 5.10 kernel.
If this is the case, changing the gate in the tests should fix it. Please then update the PR description (and particularly the change log part) and relevant commit descriptions to state that the functionality is gated at v5.11.
And then I think we should be good to go! I really appreciate the effort that has gone into this!

pkg/sensors/tracing/kprobe_net_test.go
AshishNaware: that worked!!! thanks

AshishNaware:
All tests are now passing. I have updated the release note and the PR description. Let me know if that looks good. Thanks for all the help.

kevsecurity:
> All tests are now passing. I have updated the release note and the PR description. Let me know if that looks good. Thanks for all the help.

Yep, I think this looks good. I would however change the following:
* I think the first commit title should include "unix sockets" somewhere so after the merge it is obvious what the protobuf changes were for.
* I think the last commit should really be merged into the second and third commits. It doesn't make sense to me to introduce the feature gated at 5.5 and then subsequently change it to 5.11 in the same PR, when it could just be introduced with the feature gated at 5.11.

What do you think? I'll ask some other people to also take a look, but this feels like the final step.

bpf/process/types/sockaddr_un.h
```diff
@@ -0,0 +1,58 @@
...
+ */
+FUNC_INLINE void
+set_event_from_sockaddr_un(struct sockaddr_un_type *event, struct sockaddr *address)
+{
+	struct sockaddr_un *un = (struct sockaddr_un *)address;
+
+	memset(event, 0, sizeof(*event));
+	if (probe_read(&event->family, sizeof(event->family), _(&address->sa_family)) < 0)
+		return;
+	probe_read(event->sun_path, sizeof(event->sun_path), un->sun_path);
```
FedeDP: Since here we are not checking for `probe_read` errors, if we fail to read `un->sun_path` for any reason, we might end up with `event->sun_path == '\0'`; that also means that we will set `is_abstract` as true.
I don't think this is what we want (?), but am not sure.

FedeDP: Also, small suggestion: to avoid all the
```
startIdx := 0
if sockaddr.IsAbstract {
	startIdx = 1 // Skip leading null byte for abstract sockets
}
path := make([]byte, int(sockaddr.PathLen))
copy(path, sockaddr.Path[startIdx:startIdx+int(sockaddr.PathLen)])
```
around, couldn't we:
* read first byte here
* if it's `'\0'`, set the abstract flag
* go on reading all other bytes
* push to userspace only the real bytes (without the leading empty byte for abstract sockets)

Wdyt? I think this will help clean a bit the userspace event parsing code.

FedeDP (commented):
LGTM, left a suggestion and a comment, PTAL 🙏

bpf/process/types/sockaddr_un.h
AshishNaware: @FedeDP - that was a good catch, thanks for the review. I have handled the probe_read error in recent changes. Also implemented your suggestion regarding sending real bytes to userspace.

AshishNaware:
>     I think the first commit title should include "unix sockets" somewhere so after the merge it is obvious what the protobuf changes were for.

I have updated commit messages

FedeDP:
> I think the last commit should really be merged into the second and third commits. It doesn't make sense to me to introduce the feature gated at 5.5 and then subsequently change it to 5.11 in the same PR, when it could just be introduced with the feature gated at 5.11.

I agree with @kevsecurity  but it seems like you already fixed it :)

FedeDP (approved):
LGTM! Thank you very much :) Let's wait for @kevsecurity 🚀

kevsecurity (approved):
🎉  Looks good!

bpf/process/generic_calls.h
```diff
@@ -222,6 +222,9 @@ FUNC_INLINE bool is_read_arg_1(long type)
 	case skb_type:
 	case sock_type:
 	case sockaddr_type:
+#if defined(__LARGE_BPF_PROG) && defined(__V511_BPF_PROG)
```
olsajiri: `__V511_BPF_PROG` implies `__LARGE_BPF_PROG`, please fix also all the other places

bpf/process/types/sockaddr_un.h
```diff
@@ -0,0 +1,54 @@
...
+		return;
+	if (probe_read(&first, sizeof(first), src_path) < 0)
+		return;
+	event->is_abstract = (first == '\0');
+	if (event->is_abstract) {
+		src_path = un->sun_path + 1;
+		max_len = UNIX_SOCKET_PATH_MAX - 1;
+	}
+	if (probe_read(event->sun_path, max_len, src_path) < 0)
+		return;
```
olsajiri: could we use `probe_read_str` instead (and use return value for path_len) and drop that code below?

pkg/sensors/tracing/kprobe_net_test.go
```diff
@@ -1326,3 +1377,311 @@ spec:
...
+	suite.readyWG.Wait()
+
+	tp := suite.addTracingPolicy(hook)
+	defer suite.deleteTracingPolicy(tp)
+
+	const fsSock = "/tmp/tg.sock"
+	const absSock = "@tetragon-abs"
+
+	cleanupFS := startUnixServer(suite.T(), fsSock)
+	defer cleanupFS()
```
olsajiri, line 1425: curious, could this be just `defer startUnixServer(suite.T(), fsSock)` ?

pkg/sensors/tracing/kprobe_net_test.go
```diff
@@ -1326,3 +1377,311 @@ spec:
...
+			),
+		)
+
+	checker := ec.NewUnorderedEventChecker(kpFS, kpAbs)
+	err := jsonchecker.JsonTestCheckExpectWithKeep(suite.T(), checker, false, false)
+	suite.Require().NoError(err)
+}
+
+func (suite *KprobeNet) TestKprobeSockaddrUnPathNotEqual() {
+	if !config.EnableLargeProgs() {
```
olsajiri: 5.11 kernel check implies large kernel

pkg/sensors/tracing/kprobe_net_test.go
```diff
@@ -1326,3 +1377,311 @@ spec:
...
+			),
+		)
+
+	checker := ec.NewUnorderedEventChecker(kpFS, kpAbs)
+	err := jsonchecker.JsonTestCheckExpectWithKeep(suite.T(), checker, false, false)
+	suite.Require().NoError(err)
+
+}
+
+func (suite *KprobeNet) TestKprobeSockaddrUnPathEquals() {
```
olsajiri, line 1462: there's also family filtering support, is there test for that?

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -999,6 +1000,27 @@ func handleMsgGenericTracepoint(
 			arg.SinPort = uint32(address.SinPort)
 			unix.Args = append(unix.Args, arg)
 
+		case gt.GenericSockaddrUnType:
```
olsajiri, line 1103: this seems like same code as in args_linux.go, please put that in function

olsajiri (commented):
looks good, left some comments, please provide decriptive changelogs to each commit, thanks

mtardy (commented):
can you update the docs with that new type in a sep commit? https://tetragon.io/docs/concepts/tracing-policy/argument_types/

bpf/process/types/sockaddr_un.h
AshishNaware: I believe `probe_read_str` may not behave correctly for abstract sockets, since they can include a leading NUL byte.

pkg/sensors/tracing/kprobe_net_test.go
AshishNaware: `defer startUnixServer(suite.T(), fsSock)` would not work correctly here as the server won't be listening until the test returns and could fail the test.

pkg/sensors/tracing/kprobe_net_test.go
AshishNaware: added separate test for family filtering

pkg/sensors/tracing/generictracepoint.go
AshishNaware: moved common code under `decodeSockaddrUnPath`

mtardy (approved):
docs look ok! thanks

bpf/process/types/sockaddr_un.h
kevsecurity: Abstract socket names don't have to end in NUL so probe_read_str() probably wouldn't work.

olsajiri: hum, with leading NULL byte the length is zero also with your current code, could you please explain in more detail what's the issue? thnx

AshishNaware: Sorry, I was wrong to reason that `probe_read_str` may not work due to leading `NUL` byte as the `probe_read` is reading full path (or from `sun_path + 1` in case of abstract sockets) until the first `NUL` byte is detected.

As @kevsecurity pointed out in current thread and [past comment ](https://github.com/cilium/tetragon/pull/3980#discussion_r2251003784) `probe_read_str()` may not work in case of abstract sockets not ending with `NUL`. 

However, after reading below comment from the codebase, 
```
// probe_read_str() always nul-terminates the string.
// So add one to the length to allow for it. This should
// result in us honouring our max_size correctly.
```
I am kind of thinking whether probe_read_str() is a good choice as when there exists an abstract path (without `NUL` terminator)  i.e. `NUL + 107 bytes`, it would just return a path of size 108. (`107 bytes + NUL`)

kevsecurity: When the man page (https://man7.org/linux/man-pages/man7/unix.7.html) states:

```
Null bytes in the name have no special significance.
```

does this mean the abstract path can contain null bytes? I read it that it does, so a probe_read_str() that stops on a null byte would be wrong IMO. Also, if it does not, then why not specify the length using strnlen() (similar to what the normal socket path description does, but with a max length, n)? I think this is strong evidence that the abstract path can contain null characters.

AshishNaware: > The name of the socket is contained in the first (addrlen - sizeof(sa_family_t)) bytes of sun_path. 

So we need addrlen to extract the actual path and use `probe_read` instead of `probe_read_str` (since null bytes can appear in the name). Any suggestions on how to obtain addrlen?

kevsecurity: Often when a function takes a sockaddr, it also takes a len. We have other data types (one I think) in Tetragon where they require a length parameter (memory pointer, for example). I'll take a look at the code next week and see if we can do the same thing.

pkg/grpc/tracing/tracing.go
```diff
@@ -634,6 +644,16 @@ func (msg *MsgGenericTracepointUnix) HandleMessage() *tetragon.GetEventsResponse
 				SockaddrArg: &address,
 			}})
 
+		case tracingapi.MsgGenericKprobeArgSockaddrUn:
+			address := tetragon.KprobeSockaddrUn{
+				Family: familyString(v.Family),
+				Path:   strutils.UTF8FromBPFBytes([]byte(v.Path)),
```
olsajiri: is this path already converted?

olsajiri (commented):
left one comment, otherwise lgtm, thanks

pkg/grpc/tracing/tracing.go
AshishNaware: yep. fixed it. 

I see the CI is failing on windows build. Can we re-trigger the workflows if those are flaky tests? I don't think it has something to do with this fix.

FedeDP:
I am not sure whether the test:
```
--- FAIL: Test_rateLimitExport (0.05s)
    --- FAIL: Test_rateLimitExport/rate_limit_(100_events,_10_rate_limit) (0.05s)
        exporter_test.go:277: 
            	Error Trace:	D:/a/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/exporter/exporter_test.go:196
            	            				D:/a/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/exporter/exporter_test.go:277
            	Error:      	Not equal: 
            	            	expected: 10
            	            	actual  : 12
            	Test:       	Test_rateLimitExport/rate_limit_(100_events,_10_rate_limit)
            	Messages:   	number of events
        exporter_test.go:277: 
            	Error Trace:	D:/a/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/exporter/exporter_test.go:198
            	            				D:/a/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/exporter/exporter_test.go:277
            	Error:      	Not equal: 
            	            	expected: 0x5a
            	            	actual  : 0x58
            	Test:       	Test_rateLimitExport/rate_limit_(100_events,_10_rate_limit)
            	Messages:   	number of dropped events
FAIL
FAIL	github.com/cilium/tetragon/pkg/exporter	0.380s
```
is flaky; never seen that failure (@mtardy do you?).

olsajiri (approved):

mtardy:
> is flaky; never seen that failure (@mtardy do you?).

never seen that one in particular nop, looks like it was flaky on windows also

bpf/process/types/sockaddr_un.h
mtardy: So I'm not exactly sure on how this ended up being fixed, related comment here as well So I'm not exactly sure on how this ended up being fixed.

Could you clarify what what was decided?

bpf/process/types/sockaddr_un.h
```diff
@@ -0,0 +1,42 @@
...
+	event->is_abstract = (first == '\0');
+	if (event->is_abstract)
+		src_path = un->sun_path + 1;
+
+	ret = probe_read_str(event->sun_path, UNIX_SOCKET_PATH_MAX, src_path);
+	if (ret < 0)
+		return;
+	event->path_len = (__u8)(ret - 1);
+}
```
mtardy, line 50: So just before merging, could we clarify that this satisfies everyone, I was not sure of the outcome of this discussion https://github.com/cilium/tetragon/pull/3980#discussion_r3008914504.

cc @kevsecurity @olsajiri

mtardy: From offline discussions I think we can always open a separate issue to take a closer look on this if needed but we can move forward with this PR.

mtardy: Also for next time Ashish, you don't need to add "changelog" to your git commit description, and please wrap the descriptions so that it fits ~80 chars.

bpf/process/types/sockaddr_un.h
kevsecurity: So, unfortunately I think ```probe_read_str()``` is wrong because the data could contain nuls. The best solution would be to ```probe_read_kernel()``` with a length that is 3 less than the supplied length param (2 less because that is the size of the family parameter, plus 1 for the leading nul), but that adds the complexity of needing the length parameter.

I think maybe the easiest answer here is to ```probe_read_kernel()``` the full 107 byte buffer (starting from offset 1) and deal with it in user space. I imagine we might not display abstract paths with nuls correctly at this point anyway. Maybe if we make this small change (probe_read_str -> probe_read_kernel with len 107), then we can merge and create an issue for someone to look at abstract sockets more closely?

olsajiri: > I think maybe the easiest answer here is to probe_read_kernel() the full 107 byte buffer (starting from offset 1) and deal with it in user space. I imagine we might not display abstract paths with nuls correctly at this point anyway. Maybe if we make this small change (probe_read_str -> probe_read_kernel with len 107), then we can merge and create an issue for someone to look at abstract sockets more closely?

sounds good

bpf/process/types/sockaddr_un.h
kevsecurity: This looks great! Thanks.

kevsecurity:
I think you might only need a rebase to get through the CI. Once tests pass, I'm happy to merge.

PR #3984, opened 2025-08-04
deps: remove direct gopkg.in/yaml.v2 dep

mtardy:
The project is unmaintained, and we regularly use the k8s package elsewhere, let's use it everywhere.

This will auto-close https://github.com/cilium/tetragon/pull/3436.


michi-covalent (approved):
😇

PR #3986, opened 2025-08-05
pin github actions by hash

datosh:
Fixes 

### Description

I've noticed that this project is pinning most of its GitHub Action dependencies by referencing a commit hash. This is great, as it ensures that the workflows are both stable and secure. It is a security best practice, [endorsed by GitHub](https://docs.github.com/en/actions/how-tos/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), and helps prevent security incidents such as [CVE-2025-30066](https://www.wiz.io/blog/github-action-tj-actions-changed-files-supply-chain-attack-cve-2025-30066), aka the "tj-actions/changed-files supply chain attack".

I was not able to find any legitimate reason in the git history or issues, so I assume these were just overlooked.

This PR pins the actions to match the best practice followed by the other workflows in this repository.

Pinning this and tracking the master branch is also no problem for renovate. It will recommend new versions as new commits are made to master branch.

### Changelog

```release-note
pin alexellis/arkade-get github action by hash
```


.github/workflows/run-e2e-tests.yaml
```diff
@@ -89,7 +89,7 @@ jobs:
         go-version: '1.24.5'
 
     - name: Install kubectl, kind and cilium CLI
-      uses: alexellis/arkade-get@master
+      uses: alexellis/arkade-get@67512250416516917cfcda50a3ee1144ce39585a # master
```
mtardy: can you indicate the version in the magic comment nearby with `# vX.Y.Z`, it's for renovate to be able to update it.

mtardy (changes requested):

mtardy:
Thanks for taking the time to do this and improve our CI :)

.github/workflows/run-e2e-tests.yaml
datosh: This is definitely the best way to do it, @mtardy , but sadly this project [does not provide any tags](https://github.com/alexellis/arkade-get/tags) we could reference here.

This is not an issue for renovate, though! It will happily track the `master` branch and propose updates as new commits are made. Please see here for how this will look:
1. https://github.com/datosh/pinned-actions/blob/main/.github/workflows/track-main.yml#L18
2. https://github.com/datosh/pinned-actions/issues/5
3. https://github.com/datosh/pinned-actions/pull/6

I can create an issue with the project, and ask them to release / tag a version.

mtardy: ah so that's why I didn't use it in the first place... Yeah yeah I see but that would be annoying to have constant renovate updates popping up (we would not review the code manually anyway so I don't think it would make a difference security wise). Given we no longer use Cilium in the e2e tests after
```
make e2e-test E2E_TESTS=${{matrix.package.f}} E2E_BUILD_IMAGES=0 E2E_AGENT=${{ needs.prepare.outputs.agentImage }} E2E_OPERATOR=${{ needs.prepare.outputs.operatorImage }} EXTRA_TESTFLAGS="-cluster-name=${{ env.clusterName }} -args -v=4 -tetragon.install-cilium=false"
```

maybe we could remove it altogether and install helm/kubectl/kind with our own step or with their respective GitHub action step. It would just need to work both on amd64 and arm64.

mtardy: https://github.com/Azure/setup-kubectl
https://github.com/Azure/setup-helm
https://github.com/helm/kind-action with `install_only`

datosh: That makes a lot of sense to me!
I have pushed a new version. Let me know your thoughts.

mtardy (commented):
Hey the patch looks great, unfortunately we merged this in the meantime (it's kinda my fault hehe) https://github.com/cilium/tetragon/pull/3987 so you might need to rebase, but CI passes, everything looks good!

datosh:
PTAL

mtardy (approved):
Thank you, let's see if the CI is green :)

PR #4000, opened 2025-08-08
tetragon: Change generic usdt op number

olsajiri:
Let's follow the latest op event number added.


pkg/api/ops/ops.go
```diff
@@ -28,14 +28,14 @@ const (
 	MSG_OP_GENERIC_TRACEPOINT = 14
 	MSG_OP_GENERIC_UPROBE     = 15
 	MSG_OP_GENERIC_LSM        = 16
-	MSG_OP_GENERIC_USDT       = 17
```
kkourt, line 31: Let's add a comment for future reference here?

PR #4005, opened 2025-08-12
tetragon: add support for usdt set action

olsajiri:
Adding support for usdt set action which allows to set/write
value to user space usdt argument.

The set action is defined by index (argIndex), which specifies
usdt output argument and its value (argValue) which will be set 
by this action, like:
```
  - action: Set 
    argIndex: 0
    argValue: 32

```
Above set action sets '32' value to usdt frst argument.

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -621,6 +621,7 @@ matches. They are defined under `matchActions` and currently, the following
 - [TrackSock action](#tracksock-action)
 - [UntrackSock action](#untracksock-action)
 - [Notify Enforcer action](#notify-enforcer-action)
+- [USDT Set action](#usdt-set-action)
 
 {{< warning >}}
 The FollowFD and related (UnfollowFD, CopyFD) actions have been deprecated due to being unsafe and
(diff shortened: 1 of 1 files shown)


kkourt:
Tried this out and it seems to work 🎉 

One thing I noticed is that the kernel has lockdown enabled, the program refuses to load:
> [66658.017991] Lockdown: tetragon: use of bpf to write user RAM is restricted; see man kernel_lockdown.7

This happens even if we don't use the "Set" action. So we might want to have a way to check this at runtime and set a constant in the bpf code to prove to the verifier that the write is actually dead code, so that we can load policies that do not use "Set" in lockdown kernels. It can be a followup to this PR though.

api/v1/tetragon/tetragon.proto
```diff
@@ -510,6 +510,8 @@ enum KprobeAction {
   KPROBE_ACTION_NOTIFYENFORCER = 13;
   // CleanupEnforcerNotification action cleanups any state left by NotifyEnforcer
   KPROBE_ACTION_CLEANUPENFORCERNOTIFICATION = 14;
+  // Set action sets first USDT argument
+  KPROBE_ACTION_SET = 15;
```
mtardy, line 514: >  - action: Set
   argIndex: 0
   argValue: 32
>
> Above set action sets '32' value to usdt frst argument.

Could you explain what's the use case? I don't have much context on that, and to me USDT are user space tracepoint, so I don't fully get why we would like to set a value to a USDT argument and what it would mean?

pkg/sensors/program/loader_linux.go
```diff
@@ -282,100 +282,196 @@ func KprobeAttach(load *Program, bpfDir string) AttachFunc {
...
+		RelinkFn:   linkFn,
+	}, nil
+}
+
+func uprobeAttachWriteOffload(load *Program, bpfDir string,
+	coll *ebpf.Collection, collSpec *ebpf.CollectionSpec) error {
+
+	spec, ok := collSpec.Programs["generic_usdt_write_offload"]
+	if !ok {
+		return errors.New("spec for generic_usdt_write_offload  program not found")
```
mtardy: nit
```suggestion
		return errors.New("spec for generic_usdt_write_offload: program not found")
```
or
```suggestion
		return errors.New("spec for generic_usdt_write_offload program not found")
```

pkg/sensors/program/loader_linux.go
```diff
@@ -282,100 +282,196 @@ func KprobeAttach(load *Program, bpfDir string) AttachFunc {
...
+	}
+
+	prog, ok := coll.Programs["generic_usdt_write_offload"]
+	if !ok {
+		return errors.New("program generic_usdt_write_offload  not found")
```
mtardy: same nit here

mtardy (approved):
looks cool, so we can modify the first arg value so that user can take different decision based on Tetragon hook?

api/v1/tetragon/tetragon.proto
olsajiri: so it's basically a way to propagate decision from bpf code to user code through the first usdt argument, and the code can check the argument value after the usdt probe is hit and make decision based on that, like return from the function or such ...  I'll push some docs for this in separate PR if it's ok

mtardy: yeah of course it's ok :) thanks for the explanations Jiri!

bpf/process/generic_calls.h
```diff
@@ -779,6 +779,38 @@ do_override_action(__s32 error)
...
+		.value = value,
+
+	};
+	__u64 id = get_current_pid_tgid();
+
+	/*
+	 * TODO: this should not happen, it means that the override
+	 * program was not executed for some reason, we should do
+	 * warning in here
+	 */
```
kkourt, line 797: No need to do it in this PR, but this seems like a good use-case for errmetrics (see https://github.com/cilium/tetragon/pull/4120) CC: @FedeDP

mtardy: yep, here is the API https://github.com/cilium/tetragon/blob/be913d5898901e924fed4e5be8b548d1e86d1d53/bpf/lib/errmetrics.h#L60-L78

You can use `with_errmetrics` with a helper or just `errmetrics(ERRNO)`!

PR #4008, opened 2025-08-12
fix: reject NotifyEnforcer kprobe action without an Enforcer

### LINKED ISSUE #4007: TracingPolicy: reject policies when NotifyEnforcers is specified but no Enforcer present
Currently, `TracingPolicy`'s with a `NotifyEnforcer` action but no corresponding `enforcers` clause will load without error. We should at least produce a warning here to warn the user that their intended enforcement actions will be absent without the `enforcers` clause.

dwindsor:
Fixes #4007 

### Changelog

```release-note
```

### Testing
This policy fails to load (it's missing `enforcers`):
```yaml
kind: TracingPolicy
metadata:
  name: "linux-ssh-key-config-deletion-detection"
  annotations:
    author: "Isovalent now part of Cisco"
spec:
  kprobes:
  - call: "security_path_unlink"
    syscall: false
    return: false
    tags: [ "attack.T1070.004", "attack.T1485", "attack.defense_evasion", "attack.impact" ]
    args:
    - index: 0
      type: "path"
    selectors:
    # System SSH keys/config files
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/etc/ssh/ssh_host_"
        - "/etc/ssh/sshd_config"
        - "/etc/ssh/ssh_config"
      matchBinaries:
        - operator: "NotIn"
          values: ["/usr/bin/ssh-keygen", "/usr/sbin/sshd", "/usr/bin/dpkg"]
      matchActions:
      - action: NotifyEnforcer
 ```
 
Failure to load due to missing `enforcers`:
```
level=error msg="Failed to execute tetragon" error="policy handler 'tracing' failed loading policy 'linux-ssh-key-config-deletion-detection': NotifyEnforcer action specified, but spec contains no enforcers"
```

The same policy loads after adding a `enforcers`:
```yaml
kind: TracingPolicy
metadata:
  name: "linux-ssh-key-config-deletion-detection"
  annotations:
    author: "Isovalent now part of Cisco"
spec:
  enforcers:
    - calls:
      - security_path_unlink
  kprobes:
  - call: "security_path_unlink"
    syscall: false
    return: false
    tags: [ "attack.T1070.004", "attack.T1485", "attack.defense_evasion", "attack.impact" ]
    args:
    - index: 0
      type: "path"
    selectors:
    # System SSH keys/config files
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/etc/ssh/ssh_host_"
        - "/etc/ssh/sshd_config"
        - "/etc/ssh/ssh_config"
      matchBinaries:
        - operator: "NotIn"
          values: ["/usr/bin/ssh-keygen", "/usr/sbin/sshd", "/usr/bin/dpkg"]
      matchActions:
      - action: NotifyEnforcer
```

Loaded successfully:

```
dave@cartman:~/src/github.com/cilium/tetragon$ sudo ./tetra tp list
ID   NAME                                      STATE     FILTERID   NAMESPACE   SENSORS                       KERNELMEMORY   MODE
1    linux-ssh-key-config-deletion-detection   enabled   0          (global)    __enforcer__,generic_kprobe   3.91 MB        enforce
```

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -522,8 +522,8 @@ func allKprobesIgnored(info []*kpValidateInfo) bool {
 
 // preValidateKprobes pre-validates the semantics and BTF information of a Kprobe spec
 // Furthermore, it does some preprocessing of the calls and returns one kpValidateInfo struct per
-// kprobe
-func preValidateKprobes(log logger.FieldLogger, kprobes []v1alpha1.KProbeSpec, lists []v1alpha1.ListSpec) ([]*kpValidateInfo, error) {
+// kprobe. It also validates that if any selector uses NotifyEnforcer action, the spec contains enforcers.
+func preValidateKprobes(log logger.FieldLogger, kprobes []v1alpha1.KProbeSpec, lists []v1alpha1.ListSpec, enforcers []v1alpha1.EnforcerSpec) ([]*kpValidateInfo, error) {
 	btfobj, err := btf.NewBTF()
 	if err != nil {
 		return nil, err
@@ -539,6 +539,15 @@ func preValidateKprobes(log logger.FieldLogger, kprobes []v1alpha1.KProbeSpec, l
 	ks, err := ksyms.KernelSymbols()
 	if err != nil {
 		return nil, fmt.Errorf("validateKprobeSpec: ksyms.KernelSymbols: %w", err)
+	}
+
+	// If the NotifyEnforcer action is specified, there must be at least one enforcer.
+	for _, kprobe := range kprobes {
+		if selectors.HasNotifyEnforcerAction(&kprobe) {
+			if len(enforcers) == 0 {
+				return nil, errors.New("NotifyEnforcer action specified, but spec contains no enforcers")
+			}
+		}
 	}
 
 	ret := make([]*kpValidateInfo, len(kprobes))
--- a/pkg/sensors/tracing/policyhandler_linux.go
+++ b/pkg/sensors/tracing/policyhandler_linux.go
@@ -126,7 +126,7 @@ func (h policyHandler) PolicyHandler(
 			"policy", tracingpolicy.TpLongname(policy),
 			"sensor", name,
 		)
-		validateInfo, err := preValidateKprobes(log, spec.KProbes, spec.Lists)
+		validateInfo, err := preValidateKprobes(log, spec.KProbes, spec.Lists, spec.Enforcers)
 		if err != nil {
 			return nil, fmt.Errorf("validation failed: %w", err)
 		}
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -1497,6 +1497,18 @@ func HasFilter(selectors []v1alpha1.KProbeSelector, index uint32) bool {
 	return false
 }
 
+// HasNotifyEnforcerAction returns true if any selector in the KProbeSpec has a NotifyEnforcer action
+func HasNotifyEnforcerAction(kspec *v1alpha1.KProbeSpec) bool {
+	for _, s := range kspec.Selectors {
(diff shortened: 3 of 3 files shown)


pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -126,6 +126,27 @@ func (h policyHandler) PolicyHandler(
 			"policy", tracingpolicy.TpLongname(policy),
 			"sensor", name,
 		)
+
+		// Check if any selector uses NotifyEnforcer action
```
kkourt: The check makes sense to me. Can you add it to the `preValidateKprobes` function?
I think it's better to have all validation in a single place.

kkourt (changes requested):
Thanks!

pkg/sensors/tracing/policyhandler_linux.go
dwindsor: Yeah makes sense! Thanks

pkg/sensors/tracing/generickprobe.go
```diff
@@ -541,6 +541,29 @@ func preValidateKprobes(log logger.FieldLogger, kprobes []v1alpha1.KProbeSpec, l
 		return nil, fmt.Errorf("validateKprobeSpec: ksyms.KernelSymbols: %w", err)
 	}
 
+	// Check if any selector uses NotifyEnforcer action
+	hasNotifyEnforcer := false
+	for _, kprobe := range kprobes {
```
kkourt, line 545: How about using a [label](https://go.dev/ref/spec#Break_statements) to simplify breaking out of multiple loops?

kkourt (changes requested):
Thanks!

Can you also please squash the changes form my original feedback into the relevant original commits? (`git rebase --interactive` using the `squash` and `fixup` actions should help).

dwindsor:
qq @kkourt do you think it's worth adding a ci test for this? not sure if it's worth it (ie try to load a policy w/ notifyenforcers and no enforcers)

kkourt:
> qq @kkourt do you think it's worth adding a ci test for this? not sure if it's worth it (ie try to load a policy w/ notifyenforcers and no enforcers)

I think trying to load a policy is a lot of effort to test this. If we want to test this, I would suggest to just write some unit tests for the `preValidateKprobes` function.

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -126,10 +126,12 @@ func (h policyHandler) PolicyHandler(
 			"policy", tracingpolicy.TpLongname(policy),
 			"sensor", name,
 		)
-		validateInfo, err := preValidateKprobes(log, spec.KProbes, spec.Lists)
+
```
mtardy: nit: whiteline changes

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -126,10 +126,12 @@ func (h policyHandler) PolicyHandler(
...
+		validateInfo, err := preValidateKprobes(log, spec.KProbes, spec.Lists, spec.Enforcers)
 		if err != nil {
 			return nil, fmt.Errorf("validation failed: %w", err)
 		}
+
```
mtardy: here as well

pkg/sensors/tracing/generickprobe.go
olsajiri: we do have already some of selectors.Has* functions, we could add it there
also we do have `func HasSigkillAction(kspec *v1alpha1.KProbeSpec) bool` maybe we could add generic `HasAction` helper

dwindsor: Thanks just added `HasNotifyEnforcerAction`, I agree that `HasAction` would be useful but could we do that in another PR?

PR #4012, opened 2025-08-14
tetragon: add attribute source

olsajiri:
There's a special type of argument that retrieves its data from current process
object and is being treated as argument value when it comes to selector filters.

Following example attaches to `sys_prctl` syscall and resolves `mm.owner.pid` field
from current process object.

```yaml
- call: "sys_prctl"
  syscall: true
  args:
  - type: "int"
    source: "current_task"
    resolve: "mm.owner.pid"
  selectors:
  - matchArgs:
    - index: 0
      operator: "Equal"
        values: 1234
```

The argument value is then used in selector filter as standard argument value.

By default (`source` attribute unset) the argument value is retrieved from
function's arguments.

If `source` attribute is set to `current_task` (currently the only possible configuration),
tetragon get current process address and resolves specified field path.

The `resolve` attribute is mandatory for `current` source.

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -679,6 +679,38 @@ spec:
       - action: Post
 ```
 
+## Argument source
+
+There's a special type of argument that retrieves its data from current process
+object and is being treated as argument value when it comes to selector filters.
+
+Following example attaches to `sys_prctl` syscall and resolves `mm.owner.pid` field
+from current process object.
+
+```yaml
+- call: "sys_prctl"
+  syscall: true
+  args:
+  - type: "int"
+    source: "current_task"
+    resolve: "mm.owner.pid"
+  selectors:
+  - matchArgs:
+    - index: 0
+      operator: "Equal"
+        values: 1234
+```
+
+The argument value is then used in selector filter as standard argument value.
+
+By default (`source` attribute unset) the argument value is retrieved from
+function's arguments.
+
+If `source` attribute is set to `current` (currently the only possible configuration),
+tetragon get current process address and resolves specified field path.
+
+The `resolve` attribute is mandatory for `current` source.
+
 ## Return values
 
 A `TracingPolicy` spec can specify that the return value should be reported in


pkg/sensors/tracing/kprobe_test.go
```diff
@@ -6897,3 +6897,63 @@ func TestCapabilitiesGained(t *testing.T) {
...
+metadata:
+  name: "resolve-parent-comm"
+spec:
+  kprobes:
+  - call: "security_task_getscheduler"
+    syscall: false
+    args:
+    - type: "int"
+      source: "current"
+      resolve: "mm.owner.pid"
```
tixxdz, line 6925: So here:

source: current , current is obvious for a few developers only I guess, maybe we add current_process or something else? hmm but again this could bring it back near matchCurrentProcess...  so no strong opinion here, but source could be "current_function" (default optional)  or "current_task|process"

So if you advance with this, we first test the equal operator and I can rework the other matchCurrentCred PR to only have tests, and we iterate over the range later, first step will be good to have at least one value matching.

olsajiri: ok, btw looks like we have range operator already
```
    - matchArgs:
      - index: 0
        operator: "InMap"
        values:
        - "0:2"
```

olsajiri: > source: current , current is obvious for a few developers only I guess, maybe we add current_process or something else? hmm but again this could bring it back near matchCurrentProcess... so no strong opinion here, but source could be "current_function" (default optional) or "current_task|process"

`current_task` sounds good

tixxdz: > ok, btw looks like we have range operator already

great, supported range should be in: 0:4294967295

tixxdz: and I guess with NotInMap we do the whitelisting ...

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -101,6 +101,9 @@ type KProbeArg struct {
 	// +kubebuilder:validation:Optional
 	// Label to output in the JSON
 	Label string `json:"label"`
+	// +kubebuilder:validation:Optional
+	// Source of the data, if missing the default if function arguments
```
mtardy, line 105: nit: small typo
```suggestion
	// Source of the data, if missing the default is the function arguments
```

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -384,6 +385,22 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 		if argType == gt.GenericInvalidType {
 			return nil, fmt.Errorf("Arg(%d) type '%s' unsupported", i, a.Type)
 		}
+
+		if hasCurrentTaskSource(&a) {
+			if a.Resolve == "" {
+				return nil, errors.New("error: source 'current' can't be used without resolve attribute")
```
mtardy, line 391: nit: maybe try to grep for "current" in the changes to update to current_task

mtardy (commented):
I see how this is very useful and powerful. It just feel strange to use "args" as it's ultimately not related to args iiuc. It's a weird and cryptic API to me but it's not like the TracingPolicy is pristine already.

olsajiri:
> I see how this is very useful and powerful. It just feel strange to use "args" as it's ultimately not related to args iiuc. It's a cryptic API to me but it's not like the TracingPolicy is pristine already and it's nice to reuse all the rest of the facilities.

yea, I'm not happy about that as well ;-\ but atm I don't see better way.. the main driving points for me is that there's already logistics for getting the data to event and we are able to filter on it

we would need to come with spec changes that allow to define the data and filter on it.. plus bpf/user support to carry this  data from kernel to user space event .. and meanwhile it's all in the args ... or maybe we could come up with better naming for the `source` or `args` ;-)

olsajiri:
I wonder we could actualy add new spec stuff like:
```
  args:
  - index: 0
    type: int
  - index: 1
    type: int
  data:
  - type: int
    source: "current_task"
    resolve: "pid"
  selectors:
    - matchArgs:
      - index: 1
        operator: "In"
        values:
        - "0"
    - matchData:
      - index: 0
        operator: "NotIn"
        values:
        - "0"
```
and just pretend on bpf/selector/user side that it's all arguments, will give it a try

olsajiri:
> and just pretend on bpf/selector/user side that it's all arguments, will give it a try

kprobe support https://github.com/cilium/tetragon/pull/4064 , still not sure it's worth the effort ;-)

tixxdz:
> > and just pretend on bpf/selector/user side that it's all arguments, will give it a try
> 
> kprobe support #4064 , still not sure it's worth the effort ;-)

up to you ;-) ! me at this point I just want something that I can use in policies , whatever you agree on would work for me!

olsajiri:
closing in favor of https://github.com/cilium/tetragon/pull/4064

PR #4013, opened 2025-08-14
helm: Quote tetragon.processAncestors.enabled

michi-covalent:
Kubernetes removes configmap entries if the value is null. Some GitOps systems (ArgoCD in my case) consider the configmap out of sync because the desired manifest contains:

    enable-ancestors:

but the realized configmap doesn't contain this entry. Quote the Helm value tetragon.processAncestors.enabled so that the entry gets rendered as:

    enable-ancestors: ""

instead so that Kubernetes does not remove the entry from the configmap.


sayboras (approved):

mtardy:
Ah we had https://github.com/cilium/tetragon/pull/4006 on the way 😅

PR #4020, opened 2025-08-18
tetragon: add HasAuditLoginuid function

### LINKED ISSUE #2087: pkg/detect: add a new package to detect kernel config
As we detect BPF features using `pkg/bpf/detect.go` for now, it would be interesting to add more detection logic for other kernel configs that Tetragon makes use of. Indeed we have some users who cannot start Tetragon or see warnings without understanding how it's related to their kernel configuration.

This originates from [this Slack conversation](https://cilium.slack.com/archives/C03EV7KJPJ9/p1707486918770869) where user seemed to miss:
- `CONFIG_AUDIT` to get `/proc/*/loginuid`.
- `CONFIG_SECURITY` to have `security_bprm_committing_creds` available.

zeyuwzy:
1.Add `HasAuditLoginuid()` to check kernel support for `CONFIG_AUDIT` when auid is unavailable.
2.Add `DetectConfig()` ,in pkg/kernels/config.go, to detect kernel configuration status, enabling other detect functions(e.g., HasAuditLoginuid) to access configs of interest.
3.Add `audit_loginuid` field in `tetra probe`.
4.Add `tetra probe config` command to list availability of Linux kernel configuration options.

Fixes #2087

### DIFF
--- a/pkg/bpf/detect_linux.go
+++ b/pkg/bpf/detect_linux.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/cilium/tetragon/pkg/arch"
 	"github.com/cilium/tetragon/pkg/btf"
+	"github.com/cilium/tetragon/pkg/kernels"
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/logger/logfields"
 )
@@ -47,6 +48,7 @@ var (
 	missedStatsKprobeMulti Feature
 	batchUpdate            Feature
 	uprobeRefCtrOffset     Feature
+	auditLoginuid          Feature
 )
 
 func HasOverrideHelper() bool {
@@ -508,14 +510,36 @@ func HasUprobeRefCtrOffset() bool {
 	return uprobeRefCtrOffset.detected
 }
 
+func detectAuditLoginuid() bool {
+	if kernels.MinKernelVersion("5.1") && kernels.DetectConfig(kernels.CONFIG_AUDIT) {
+		return true
+	} else if kernels.DetectConfig(kernels.CONFIG_AUDITSYSCALL) {
+		return true
+	}
+
+	_, err := os.Stat("/proc/self/loginuid")
+	return err == nil
+}
+
+func detectAuditLoginuidOnce() {
+	auditLoginuid.init.Do(func() {
+		auditLoginuid.detected = detectAuditLoginuid()
+	})
+}
+
+func HasAuditLoginuid() bool {
+	detectAuditLoginuidOnce()
+	return auditLoginuid.detected
+}
+
 func LogFeatures() string {
 	// once we have detected all features, flush the BTF spec
 	// we cache all values so calling again a Has* function will
 	// not load the BTF again
 	defer ebtf.FlushKernelSpec()
-	return fmt.Sprintf("override_return: %t, buildid: %t, kprobe_multi: %t, uprobe_multi %t, fmodret: %t, fmodret_syscall: %t, signal: %t, large: %t, link_pin: %t, lsm: %t, missed_stats_kprobe_multi: %t, missed_stats_kprobe: %t, batch_update: %t, uprobe_refctroff: %t",
+	return fmt.Sprintf("override_return: %t, buildid: %t, kprobe_multi: %t, uprobe_multi %t, fmodret: %t, fmodret_syscall: %t, signal: %t, large: %t, link_pin: %t, lsm: %t, missed_stats_kprobe_multi: %t, missed_stats_kprobe: %t, batch_update: %t, uprobe_refctroff: %t, audit_loginuid: %t",
 		HasOverrideHelper(), HasBuildId(), HasKprobeMulti(), HasUprobeMulti(),
 		HasModifyReturn(), HasModifyReturnSyscall(), HasSignalHelper(), HasProgramLargeSize(),
 		HasLinkPin(), HasLSMPrograms(), HasMissedStatsKprobeMulti(), HasMissedStatsPerfEvent(),
-		HasBatchAPI(), HasUprobeRefCtrOffset())
+		HasBatchAPI(), HasUprobeRefCtrOffset(), HasAuditLoginuid())
 }
--- a/pkg/kernels/config.go
+++ b/pkg/kernels/config.go
@@ -1,0 +1,165 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build !windows
+
+package kernels
+
+import (
+	"bufio"
+	"compress/gzip"
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/logger/logfields"
+
+	"golang.org/x/sys/unix"
+)
+
+type Config string
+
+const (
+	CONFIG_AUDIT                  Config = "CONFIG_AUDIT"
+	CONFIG_AUDITSYSCALL           Config = "CONFIG_AUDITSYSCALL"
+	CONFIG_BPF                    Config = "CONFIG_BPF"
+	CONFIG_BPF_EVENTS             Config = "CONFIG_BPF_EVENTS"
+	CONFIG_BPF_JIT                Config = "CONFIG_BPF_JIT"
+	CONFIG_BPF_JIT_DEFAULT_ON     Config = "CONFIG_BPF_JIT_DEFAULT_ON"
+	CONFIG_BPF_KPROBE_OVERRIDE    Config = "CONFIG_BPF_KPROBE_OVERRIDE"
+	CONFIG_BPF_SYSCALL            Config = "CONFIG_BPF_SYSCALL"
+	CONFIG_CGROUPS                Config = "CONFIG_CGROUPS"
+	CONFIG_DEBUG_INFO_BTF         Config = "CONFIG_DEBUG_INFO_BTF"
+	CONFIG_DEBUG_INFO_BTF_MODULES Config = "CONFIG_DEBUG_INFO_BTF_MODULES"
+	CONFIG_FTRACE_SYSCALLS        Config = "CONFIG_FTRACE_SYSCALLS"
+	CONFIG_HAVE_BPF_JIT           Config = "CONFIG_HAVE_BPF_JIT"
+	CONFIG_HAVE_EBPF_JIT          Config = "CONFIG_HAVE_EBPF_JIT"
+	CONFIG_SECURITY               Config = "CONFIG_SECURITY"
+)
+
+var (
+	getKernelConfig = sync.OnceValue(initKernelConfig)
+	kernelConfigMap = map[Config]string{
+		CONFIG_AUDIT:                  "n",
+		CONFIG_AUDITSYSCALL:           "n",
+		CONFIG_BPF:                    "n",
+		CONFIG_BPF_EVENTS:             "n",
+		CONFIG_BPF_JIT:                "n",
+		CONFIG_BPF_JIT_DEFAULT_ON:     "n",
+		CONFIG_BPF_KPROBE_OVERRIDE:    "n",
+		CONFIG_BPF_SYSCALL:            "n",
+		CONFIG_CGROUPS:                "n",
+		CONFIG_DEBUG_INFO_BTF:         "n",
+		CONFIG_DEBUG_INFO_BTF_MODULES: "n",
+		CONFIG_FTRACE_SYSCALLS:        "n",
+		CONFIG_HAVE_BPF_JIT:           "n",
+		CONFIG_HAVE_EBPF_JIT:          "n",
+		CONFIG_SECURITY:               "n",
+	}
+)
+
+func scanKernelConfig(reader io.Reader) error {
+	scanner := bufio.NewScanner(reader)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if line == "" || strings.HasPrefix(line, "#") {
+			continue
+		}
+
+		parts := strings.SplitN(line, "=", 2)
+		if len(parts) == 2 {
+			key := strings.TrimSpace(parts[0])
+			value := strings.TrimSpace(parts[1])
+
+			if key == "" || value == "" {
+				continue
+			}
+
+			if _, ok := kernelConfigMap[Config(key)]; ok {
+				kernelConfigMap[Config(key)] = value
+			}
+		}
+	}
+
+	return scanner.Err()
+}
+
+func initKernelProcConfigGz() error {
+	file, err := os.Open("/proc/config.gz")
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	gzReader, err := gzip.NewReader(file)
+	if err != nil {
+		return err
+	}
+	defer gzReader.Close()
+
+	return scanKernelConfig(gzReader)
+}
+
+func initKernelBootConfig() error {
+	var uname unix.Utsname
+	if err := unix.Uname(&uname); err != nil {
+		return err
+	}
+
+	kernelRelease := strings.TrimRight(string(uname.Release[:]), "\x00")
+	configPath := "/boot/config-" + kernelRelease
+	file, err := os.Open(configPath)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	return scanKernelConfig(file)
+}
+
+func initKernelConfig() error {
+	if err := initKernelProcConfigGz(); err != nil {
+		return initKernelBootConfig()
+	}
+
+	return nil
+}
+
+// DetectConfig checks if a kernel config option is enabled
+// It first tries /proc/config.gz, then looks for config file in /boot for the current kernel
+func DetectConfig(conf Config) bool {
+	if err := getKernelConfig(); err != nil {
+		logger.GetLogger().Error("Detecting kernel config failed", logfields.Error, err)
+		return false
+	}
+
+	if val, ok := kernelConfigMap[conf]; ok {
+		// Only check for CONFIG_XXX=y or CONFIG_XXX=m. When the value is a string or number,
+		// it's mostly treated as a kernel parameter rather than a feature switch,
+		// so we temporarily filter these out.
+		if val == "y" || val == "m" {
+			return true
+		}
+	}
+
+	return false
+}
+
+func LogConfigs() string {
+	if err := getKernelConfig(); err != nil {
+		logger.GetLogger().Error("Detecting kernel config failed", logfields.Error, err)
+		return ""
+	}
+
+	var items []string
+	for k, v := range kernelConfigMap {
+		items = append(items, fmt.Sprintf("%s:%s", k, v))
+	}
+	sort.Strings(items)
+
+	return strings.Join(items, " ")
+}
--- a/cmd/tetra/probe/config.go
+++ b/cmd/tetra/probe/config.go
@@ -1,0 +1,26 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build !windows
+
+package probe
+
+import (
+	"strings"
+
+	"github.com/cilium/tetragon/pkg/kernels"
+
+	"github.com/spf13/cobra"
+)
+
+func NewConfigCmd() *cobra.Command {
+	cmd := cobra.Command{
+		Use:   "config",
+		Short: "Probe for the availability of Linux kernel configuration options",
+		Run: func(cmd *cobra.Command, _ []string) {
+			cmd.Println(strings.ReplaceAll(kernels.LogConfigs(), " ", "\n"))
+		},
+	}
+
+	return &cmd
+}
--- a/cmd/tetra/probe/probe.go
+++ b/cmd/tetra/probe/probe.go
@@ -42,5 +42,7 @@ func New() *cobra.Command {
 			cmd.Println(strings.ReplaceAll(bpf.LogFeatures(), ", ", "\n"))
 		},
 	}
+
+	cmd.AddCommand(NewConfigCmd())
 	return &cmd
 }


zeyuwzy:
> we could also try to read /boot/config-`uname..` or /proc/config.gz first to get all config values that we are interested in and have this change as a fallback for CONFIG_AUDIT in case none of those files is present

Thank you for the suggestion. I will update the PR to add an interface for detecting kernel configuration, making it easier for future checks to use and get the kernel configs they need.

pkg/bpf/detect_linux.go
```diff
@@ -508,14 +511,98 @@ func HasUprobeRefCtrOffset() bool {
...
+		if strings.HasPrefix(line, configOption+"=y") || strings.HasPrefix(line, configOption+"=m") {
+			return true, nil
+		}
+	}
+
+	return false, scanner.Err()
+}
+
+func detectKernelBootConfig(configOption string) (bool, error) {
+	kernelRelease, err := os.ReadFile("/proc/sys/kernel/osrelease")
```
olsajiri: uname will return that as well but with just one syscall, right?

pkg/bpf/detect_linux.go
```diff
@@ -508,14 +511,98 @@ func HasUprobeRefCtrOffset() bool {
...
+
+	scanner := bufio.NewScanner(gzReader)
+	for scanner.Scan() {
+		line := scanner.Text()
+
+		if strings.HasPrefix(line, configOption+"=y") || strings.HasPrefix(line, configOption+"=m") {
+			return true, nil
+		}
+	}
+
```
olsajiri: the loop above should be shared (in function or otherwise) with detectKernelBootConfig

pkg/bpf/detect_linux.go
```diff
@@ -508,14 +511,98 @@ func HasUprobeRefCtrOffset() bool {
...
+func detectKernelConfig(configOption string) (bool, error) {
+	if enabled, err := detectKernelProcConfigGz(configOption); err == nil {
+		return enabled, nil
+	}
+
+	return detectKernelBootConfig(configOption)
+}
+
+func detectAuditLoginuid() bool {
+	enabled, err := detectKernelConfig("CONFIG_AUDIT")
```
olsajiri: would be great to have a support to call this just once for all the config options we want to detect, so we can parse those files just once, but let's do that when we have the need in the future

pkg/bpf/detect_linux.go
```diff
@@ -508,14 +511,98 @@ func HasUprobeRefCtrOffset() bool {
...
+			return true, nil
+		}
+	}
+
+	return false, scanner.Err()
+}
+
+// detectKernelConfig checks if a kernel config option is enabled
+// It first tries /proc/config.gz, then looks for config file in /boot for the current kernel
+func detectKernelConfig(configOption string) (bool, error) {
```
olsajiri: could we move this to pkg/kernels/config.go or such? we could call `kernels.DetectConfig("...")`

olsajiri (dismissed):
looks good, left few comments, thanks

pkg/bpf/detect_linux.go
zeyuwzy: Thanks for the suggestions. I have created a struct in pkg/kernels/config.go to store the parsing results of the kernel configuration file.  If there are other related use cases in the future, this structure can be exported by providing a new function, thanks

pkg/bpf/detect_linux.go
zeyuwzy: A more universal approach, thanks

pkg/bpf/detect_linux.go
mtardy: Just as a note, that would be nice to know approximately how much space this might take (you should be able to see easily [with pprof](https://mtardy.com/posts/memory-golang/#memory-profiling-with-pprof)). In the past we had to flush the kallsyms cache because of memory use, this should be an order of magnitude smaller but still might be significant.
```
mtardy@lima-default:/boot $ cat config-6.8.0-71-generic | wc -l
14390
mtardy@lima-default:/boot $ cat /proc/kallsyms | wc -l
194742
```

zeyuwzy: If Tetragon were to invoke the initialization interface, it would result in an increase of approximately 500 KB in resident memory (based on my configuration of 14,649 lines). However, in the current PR, this interface is only being called within the tetra CLI. Just wanted to check, is this memory increase within an acceptable range?

`zeyu@zeyu-Parallels-ARM-Virtual-Machine:~/work/tetragon$ go tool pprof -base before -top after
tetra
Build ID: c48ec0b28a7dbf1e79c4884e12be41fbbbd78343
Type: inuse_space
Time: 2025-08-26 21:39:26 CST
Showing nodes accounting for 532.26kB, 100% of 532.26kB total
      flat  flat%   sum%        cum   cum%
  532.26kB   100%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/kernels.scanKernelConfig
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/bpf.HasAuditLoginuid (inline)
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/bpf.LogFeatures
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/bpf.detectAuditLoginuid
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/bpf.detectAuditLoginuidOnce
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/bpf.detectAuditLoginuidOnce.func1
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/kernels.DetectConfig
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/kernels.init.OnceValues[go.shape.*uint8,go.shape.interface { Error string }].func1
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/kernels.init.OnceValues[go.shape.*uint8,go.shape.interface { Error string }].func1.1
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/kernels.initKernelBootConfig
         0     0%   100%   532.26kB   100%  github.com/cilium/tetragon/pkg/kernels.initKernelConfig
         0     0%   100%   532.26kB   100%  github.com/spf13/cobra.(*Command).Execute (inline)
         0     0%   100%   532.26kB   100%  github.com/spf13/cobra.(*Command).ExecuteC
         0     0%   100%   532.26kB   100%  github.com/spf13/cobra.(*Command).execute
         0     0%   100%   532.26kB   100%  main.addCommands.New.func2
         0     0%   100%   532.26kB   100%  main.main
         0     0%   100%   532.26kB   100%  runtime.main
         0     0%   100%   532.26kB   100%  sync.(*Once).Do (partial-inline)
         0     0%   100%   532.26kB   100%  sync.(*Once).doSlow

zeyu@zeyu-Parallels-ARM-Virtual-Machine:~/work/tetragon$ cat /boot/config-6.11.0-29-generic | wc -l
14649`

mtardy: ah if the memory is just in the tetra CLI then it's fine

mtardy: I also feel like caching the config (long term) is not super useful since:
- for tetra it might be shortlived anyway
- for tetragon, it will use unnecessary memory (532.26kB is around 1MB of impact end of the line with the GC of Go)

zeyuwzy:
> I don't want to block this since I think that's a nice improvement and paves the way for more config detection.
> 
> Maybe what we want is static config detection and just cache the result for the ones we are interested about (so that you parse the file onetime for all your wanted configs) and not cache the whole config. You should be able to know at build time which config you want to look for so that should be doable. It's just that [1 more MB](https://github.com/cilium/tetragon/pull/4020#discussion_r2303253932) for that use might be overkill I think.

Maybe we can use a static watchlist in code to pick out just the configs Tetragon cares about and save those to the map. How about we start by adding the configs mentioned on this Tetragon FAQ page (https://tetragon.io/docs/installation/faq/) to our watchlist. This way, we'd use a lot less memory , and it would still do exactly what we need for now. If we need to add more later, we just expand this watchlist.

pkg/kernels/config.go
```diff
@@ -0,0 +1,164 @@
...
+
+func setKernelConfigMap(key, value string) {
+	for index, config := range kernelConfigMap {
+		if config.name == key {
+			config.value = value
+			kernelConfigMap[index] = config
+			break
+		}
+	}
+}
```
mtardy: Could you simplify this by having just a `kernelConfigMap = map[string]string` with the key being the name and the value being `m`, `y` or `n`? Not sure what's the benefit of having the extra int enum?

mtardy (commented):
This looks good thank you, mostly one remark to simplify a little bit maybe?

pkg/kernels/config.go
zeyuwzy: thanks , actually, I just want to redefine the parameter type for func DetectConfig(conf Config) by enum instead of directly using string. This will clearly indicate which configuration parameters are currently supported for detection.

maybe we can simplify it by defining Config as string and making kernelConfigMap a map[Config]string, while retaining the Config type.

pkg/kernels/config.go
```diff
@@ -0,0 +1,155 @@
...
+		CONFIG_HAVE_EBPF_JIT:          "n",
+		CONFIG_SECURITY:               "n",
+	}
+)
+
+func setKernelConfigMap(key, value string) {
+	if _, ok := kernelConfigMap[Config(key)]; ok {
+		kernelConfigMap[Config(key)] = value
+	}
+}
```
mtardy: nit: maybe inline this func as it's not doing much anymore, wdyt?

pkg/bpf/detect_linux.go
```diff
@@ -508,14 +510,38 @@ func HasUprobeRefCtrOffset() bool {
 	return uprobeRefCtrOffset.detected
 }
 
+func detectAuditLoginuid() bool {
+	enabled, err := kernels.DetectConfig(kernels.CONFIG_AUDIT)
+	if err != nil {
+		if _, err := os.Stat("/proc/self/loginuid"); err != nil {
+			logger.GetLogger().Error("failed to stat /proc/self/loginuid", logfields.Error, err)
```
mtardy: nit: maybe remove the log here or if the error is relevant: return an error and log it at the very last step. Or if it's not important, we can just return false here. It should be fine in this case.

mtardy (approved):
Thanks that's great like that, thank you! :)

pkg/kernels/config.go
```diff
@@ -0,0 +1,155 @@
...
+	CONFIG_BPF_JIT_DEFAULT_ON     Config = "CONFIG_BPF_JIT_DEFAULT_ON"
+	CONFIG_BPF_KPROBE_OVERRIDE    Config = "CONFIG_BPF_KPROBE_OVERRIDE"
+	CONFIG_BPF_SYSCALL            Config = "CONFIG_BPF_SYSCALL"
+	CONFIG_CGROUPS                Config = "CONFIG_CGROUPS"
+	CONFIG_DEBUG_INFO_BTF         Config = "CONFIG_DEBUG_INFO_BTF"
+	CONFIG_DEBUG_INFO_BTF_MODULES Config = "CONFIG_DEBUG_INFO_BTF_MODULES"
+	CONFIG_FTRACE_SYSCALLS        Config = "CONFIG_FTRACE_SYSCALLS"
+	CONFIG_HAVE_BPF_JIT           Config = "CONFIG_HAVE_BPF_JIT"
+	CONFIG_HAVE_EBPF_JIT          Config = "CONFIG_HAVE_EBPF_JIT"
+	CONFIG_SECURITY               Config = "CONFIG_SECURITY"
```
mtardy: A great followup for https://github.com/cilium/tetragon/issues/2087 (not for you specifically, just thinking) would be to have a tetra command to probe all these :), could be `tetra probe config` or something like that!

olsajiri: yes, we detect all those config options above but we just display audit, I'd think let's add config option only when we use that, otherwise we spend cycles for nothing ... not sure we want to display all these config options at start ? @kkourt , but `tetra probe config` sounds like nice idea

pkg/kernels/config.go
olsajiri: +1 I don't think this needs to be in function

pkg/kernels/config.go
```diff
@@ -0,0 +1,155 @@
...
+
+	kernelRelease := strings.TrimRight(string(uname.Release[:]), "\x00")
+	configPath := "/boot/config-" + kernelRelease
+	file, err := os.Open(configPath)
+	if err != nil {
+		return err
+	}
+	defer file.Close()
+
+	err = scanKernelConfig(file)
```
olsajiri: could be just `return scanKernelConfig(file)` ?

pkg/kernels/config.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+	defer file.Close()
+
+	gzReader, err := gzip.NewReader(file)
+	if err != nil {
+		return err
+	}
+	defer gzReader.Close()
+
+	err = scanKernelConfig(gzReader)
```
olsajiri: could be just `return scanKernelConfig(gzReader)` ?

olsajiri (commented):
looks good, thanks, left few comments

tixxdz (commented):
Comments here:

- Detection based on kernel config file should not fail, so if you don't detect the file you log and explicitly return non error. It is a a CONFIG option after all IIRC, and some distros or users do hide their configs explicitly and it is not a mandatory feature to run Tetragon.
- The loginuid if it fails, we should continue to work and return explicitly uint32(-1) not set , IIRC the code explicitly initialize all fields to this value, from Tetragon perspective loginuid not compiled or not set is same.
- The CONFIG_SECURITY should probably be able to detect by checking /sys/kernel/security/lsm or something like that.

Fine to log all features this helps debugging.

pkg/kernels/config.go
kkourt: Are we talking about having a `tetra probe config` that reads kernel configuration and prints out options that are relevant for tetragon? 👍🏼 makes sense to me.

pkg/kernels/config.go
zeyuwzy: Thanks for the suggestion! I'll fix that soon.

pkg/bpf/detect_linux.go
zeyuwzy: thanks, I also found this function about get loginuid in tetragon , perhaps I can use this interface to reimplement  this part https://github.com/cilium/tetragon/blob/6a6bae661e3d2fb854b4788828dcf3d89e070628/pkg/reader/proc/proc.go#L98C1-L112C1

pkg/kernels/config.go
zeyuwzy: That sounds like a good idea. I'm willing to implement it following this approach.

pkg/kernels/config.go
zeyuwzy: Thanks for the suggestion! I'll fix that soon.

zeyuwzy:
> Comments here:
> 
> * Detection based on kernel config file should not fail, so if you don't detect the file you log and explicitly return non error. It is a a CONFIG option after all IIRC, and some distros or users do hide their configs explicitly and it is not a mandatory feature to run Tetragon.
> * The loginuid if it fails, we should continue to work and return explicitly uint32(-1) not set , IIRC the code explicitly initialize all fields to this value, from Tetragon perspective loginuid not compiled or not set is same.
> * The CONFIG_SECURITY should probably be able to detect by checking /sys/kernel/security/lsm or something like that.
> 
> Fine to log all features this helps debugging.

- Thanks for the suggestion! Sorry I missed the case where the kernel config file might be hidden.The kernel config checks more as a backup to other detect methods. Given that, it makes sense to remove the err return from the interface(and add some log). 
- Perhaps an appropriate approach is to first validate the loginuid uint32(-1) and then use the kernel configuration check as a backup.
- yes, thank you for the suggestion, detection functions could be added for certain kernel config items, for example, to handle cases where files cannot be detected .For now,the kernel config detect was initially considered as a backup to other detection methods. Maybe add other checks here later?

thanks for you comments

tixxdz:
> > Comments here:
> > 
> > * Detection based on kernel config file should not fail, so if you don't detect the file you log and explicitly return non error. It is a a CONFIG option after all IIRC, and some distros or users do hide their configs explicitly and it is not a mandatory feature to run Tetragon.
> > * The loginuid if it fails, we should continue to work and return explicitly uint32(-1) not set , IIRC the code explicitly initialize all fields to this value, from Tetragon perspective loginuid not compiled or not set is same.
> > * The CONFIG_SECURITY should probably be able to detect by checking /sys/kernel/security/lsm or something like that.
> > 
> > Fine to log all features this helps debugging.
> 
>     * Thanks for the suggestion! Sorry I missed the case where the kernel config file might be hidden.The kernel config checks more as a backup to other detect methods. Given that, it makes sense to remove the err return from the interface(and add some log).
> 
>     * Perhaps an appropriate approach is to first validate the loginuid uint32(-1) and then use the kernel configuration check as a backup.

I would stay stating just /proc/self/loginuid exists is enough IIRC, otherwise backup check kernel configuration if you want.

The loginuid uint32(-1) could say:  it is not compiled in or it is not set. From Tetragon perspective it is same, but from you feature detection it is not, better be _explicit_ if it was compiled in or not, cause if compiled in then some values could be uint32(-1) and others could be set which helps a lot debugging weird cases...  
 
>     * yes, thank you for the suggestion, detection functions could be added for certain kernel config items, for example, to handle cases where files cannot be detected .For now,the kernel config detect was initially considered as a backup to other detection methods. Maybe add other checks here later?

Yes fine later and thank you for the work.

So probably some comments to fix, but the feature detection logging without failing at startup or tetra probe config LGTM!

>

zeyuwzy:
> The loginuid uint32(-1) could say: it is not compiled in or it is not set. From Tetragon perspective it is same, but from you feature detection it is not, better be _explicit_ if it was compiled in or not, cause if compiled in then some values could be uint32(-1) and others could be set which helps a lot debugging weird cases...

thanks, I understand you point, but I'm still a bit confused. Currently, HasAuditLoginuid() only returns a bool indicating whether loginuid is valid (from Tetragon's perspective). I am not sure is it necessary to add another bool return value for the compilation status? (The CONFIG_AUDIT compiled detection is provided by DetectConfig()). 
I've added some logs in HasAuditLoginuid that print the CONFIG_AUDIT compiled status when loginuid is missing or equals uint32(-1), Could this be useful for debugging case?

pkg/kernels/config.go
```diff
@@ -0,0 +1,164 @@
...
+			return true
+		}
+	}
+
+	return false
+}
+
+func LogConfigs() string {
+	if err := getKernelConfig(); err != nil {
+		logger.GetLogger().Error("initKernelConfig failed", logfields.Error, err)
```
tixxdz: maybe here say: "Detecting kernel config failed..."

pkg/bpf/detect_linux.go
```diff
@@ -508,14 +511,47 @@ func HasUprobeRefCtrOffset() bool {
 	return uprobeRefCtrOffset.detected
 }
 
+func detectAuditLoginuid() bool {
+	compiled := kernels.DetectConfig(kernels.CONFIG_AUDIT)
+
```
tixxdz: Here I would just do:

```
compiled := kernels.DetectConfig(kernels.CONFIG_AUDIT)
if compiled {
    return true
}

_, err := os.Stat("/proc/self/loginuid")
if err == nil {
   return true
}
return false
```

Make it simple, since loginuid is created per the CONFIG_AUDIT https://elixir.bootlin.com/linux/v6.16.6/source/fs/proc/base.c#L3387   and remove the extra checks. With this patch LGTM

tixxdz:
> > The loginuid uint32(-1) could say: it is not compiled in or it is not set. From Tetragon perspective it is same, but from you feature detection it is not, better be _explicit_ if it was compiled in or not, cause if compiled in then some values could be uint32(-1) and others could be set which helps a lot debugging weird cases...
> 
> thanks, I understand you point, but I'm still a bit confused. Currently, HasAuditLoginuid() only returns a bool indicating whether loginuid is valid (from Tetragon's perspective). I am not sure is it necessary to add another bool return value for the compilation status? (The CONFIG_AUDIT compiled detection is provided by DetectConfig()). I've added some logs in HasAuditLoginuid that print the CONFIG_AUDIT compiled status when loginuid is missing or equals uint32(-1), Could this be useful for debugging case?

Ah sorry what I meant is Tetragon prints the process.auid regardless, we had some bugs before where users parsed these fields IIRC and we returned zero 0 accidentally, so we explicitly initialize things now to invalid UID.

But for your detection just check if compiled or not, or if /proc/self/loginuid CONFIG_AUDIT is present or not as suggested above in the comment. Yes this improves debugging ;-)

With this change LGTM.

tixxdz:
BTW loginuid was under CONFIG_AUDITSYSCALL and changed in kernel 5.1 under CONFIG_AUDIT https://lkml.org/lkml/2019/1/22/1021  , so for that kernel version I would check this CONFIG add it in the map log it without changing that much logic.

As a follow up maybe we add a CI test job @olsajiri  to ensure if a kernel config is renamed somehow then we catch it?

pkg/bpf/detect_linux.go
zeyuwzy: Thanks for your suggestion! I will follow this approach to make the modifications and add a kernel version check for CONFIG_AUDITSYSCALL

mtardy (approved):
Looks good to me! Jiri, who is blocking the merge is off and might be back on Thursday I think, so we can probably merge end of week.

The only minor nit to me is that you could technically have separated this into two commits:
- all your new `pkg/kernels/config.go` + the `tetra probe config`.
- the whole auditLoginuid detection probe that seems specific to a use case (basically your changes from pkg/bpf/detect_linux.go`).

But that's just a nit, not a blocker, I'd merge like that.

> As a follow up maybe we add a CI test job @olsajiri to ensure if a kernel config is renamed somehow then we catch it?

I would indeed do it as a follow up and not add more things to this PR that would be great to merge after one month and all this work :)

mtardy:
You can ignore the failure from the [generated docs](https://github.com/cilium/tetragon/actions/runs/17720791881/job/50352574481?pr=4020) or rebase your PR to make it disappear, that's unrelated to this patch.

mtardy:
Thanks a lot for these patches! 🎉 

If you ever want to follow-up with something, I think that an update to the documentation FAQ item for kernel config would be nice to mention the great `tetra probe config` first things: https://tetragon.io/docs/installation/faq/#what-are-the-linux-kernel-configuration-options-needed-to-run-tetragon. We could even consider moving that FAQ entry to another page like "verification" or "configuration" in the future if it becomes too large! :)
- https://github.com/cilium/tetragon/issues/4114

PR #4022, opened 2025-08-18
selectors: Ignore empty matchBinaries

tpapagian:
When tried to use the following tracing policy:
```yaml
kind: TracingPolicy
metadata:
  name: "lsm"
spec:
  lsmhooks:
  - hook: "bprm_check_security"
    args:
      - index: 0
        type: "linux_binprm"
    selectors:
    - matchArgs:
        - index: 0
          operator: "Postfix"
          values:
            - "/true"
      matchBinaries:
        - operator: NotIn
          values: []
```

I didn't get any events when running `/usr/bin/true`.

Removing the empty matchBinaries fixes that issue.

In order to fix that, this patch ignores empty (i.e. no values) matchBinaries selectors.

### DIFF
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -1268,6 +1268,11 @@ func ParseMatchBinary(k *KernelSelectorState, b *v1alpha1.BinarySelector, selIdx
 		return fmt.Errorf("matchBinary error: %w", err)
 	}
 
+	// ignore matchBinaries selectors with no values
+	if len(b.Values) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


kkourt:
@tpapagian should we backport this?

tpapagian:
> @tpapagian should we backport this?

~~Sure, I will do that. In all supported versions (i.e. 1.5, 1.4, and 1.3)?~~

Backports:
https://github.com/cilium/tetragon/pull/4048
https://github.com/cilium/tetragon/pull/4049
https://github.com/cilium/tetragon/pull/4050

PR #4036, opened 2025-08-25
fix: detectUprobeRefCtrOffsetOnce init logic

zeyuwzy:
Fix the issue where the detectUprobeRefCtrOffsetOnce function never executes its initialization logic due to incorrectly using an initialization variable from batchUpdate.

### DIFF
--- a/pkg/bpf/detect_linux.go
+++ b/pkg/bpf/detect_linux.go
@@ -498,7 +498,7 @@ func detectUprobeRefCtrOffset() bool {
 }
 
 func detectUprobeRefCtrOffsetOnce() {
-	batchUpdate.init.Do(func() {
+	uprobeRefCtrOffset.init.Do(func() {
 		uprobeRefCtrOffset.detected = detectUprobeRefCtrOffset()
 	})
 }


pkg/bpf/detect_linux.go
```diff
@@ -498,7 +498,7 @@ func detectUprobeRefCtrOffset() bool {
 }
 
 func detectUprobeRefCtrOffsetOnce() {
-	batchUpdate.init.Do(func() {
+	uprobeRefCtrOffset.init.Do(func() {
```
kkourt, line 501: LGTM, but I wonder: Can we use `OnceFunc`  https://go.dev/pkg/sync/?m=old#OnceFunc for this and other detection functions?

olsajiri (approved):
ugh copy/paste error, thanks a lot for fixing that

pkg/bpf/detect_linux.go
mtardy: I think this is what it is, it's just wrapped into some struct

https://github.com/cilium/tetragon/blob/dc4e70d3c3131eb25b7c925b99198db0e4c86b10/pkg/bpf/detect_linux.go#L33-L36

EDIT: ah nevermind you mention OnceFunc, not just once. Yeah I guess we could use this directly!

olsajiri:
the ci fail should be fixed by https://github.com/cilium/tetragon/pull/4047, merging

PR #4039, opened 2025-08-25
tetragon: Add fentry sensor

olsajiri:
Adding support for fentry sensor. The configuration mirrors kprobes at the moment, except for enforcement, which will come later.

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -528,6 +528,7 @@ type hasMaps struct {
 	override   bool
 	sockTrack  bool
 	selector   bool
+	fentry     bool
 }
 
 // hasMapsSetup setups the has maps for the per policy maps. The per kprobe maps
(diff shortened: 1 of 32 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -528,6 +528,7 @@ type hasMaps struct {
 	override   bool
 	sockTrack  bool
 	selector   bool
+	fentry     bool
```
kkourt, line 531: Can we have this as an enum type with two values: fentry and kprobe? I think it improves readability compared to a boolean.

For example, seeing:
> loadGenericKprobeSensor(args.BPFDir, args.Load, args.Maps, args.Verbose, false)

You don't really know what `false` means.

olsajiri: yep, will change, thanks

kkourt: From what I can tell, this was not changed:
> func loadGenericKprobeSensor(bpfDir string, load *program.Program, maps []*program.Map, verbose int, fentry bool) error {

kkourt (approved):

PR #4051, opened 2025-08-26
new(tetra/getevents): allow to filter events by container name regex.

### LINKED ISSUE #3451: CLI filter with container name

_No response_

To be able to do some quick things with CLI it'd be very handy to specify a container name.

There is no `jq` inside the tetragon pod, which and even having installed it one needs to figure out a filter expression and cover the types of events user is interested in also.
It's also not obvious that one can use compact output format using a pipe. It'd be easier if there was a flag.

Here is what I had to use:
```
tetra getevents --namespace foo  --pod=foo | jq 'select(.process_exec.process.pod.container.name == "bar", .process_exit.process.pod.container.name == "bar")' | tetra getevents -o compact
```
It's workable, but having a simple flag would have been nicer, of course.

_No response_

### Code of Conduct

FedeDP:
Fixes #3451

### Description
As requested by the user, allow `tetra getevents` command to filter by container name regex.

### Changelog

```release-note
new(tetra/getevents): allow to filter events by container name regex.
```


api/v1/tetragon/events.proto
```diff
@@ -80,6 +80,9 @@ message Filter {
   google.protobuf.BoolValue in_init_tree = 16;
   // Filter ancestor processes' binaries using RE2 regular expression syntax.
   repeated string ancestor_binary_regex = 17;
+  // Filter by the container name in the process.pod.container field using RE2 regular expression syntax:
+  // https://github.com/google/re2/wiki/Syntax
+  repeated string container_name = 18;
```
mtardy: nit: maybe add the `_regex` suffix like others (I see that container_id doesn't but maybe that's a better idea to be explicit)
```suggestion
  repeated string container_name_regex = 18;
```

FedeDP: Yes exactly i followed the `container_id` convention! But i agree it is better to be explicit! Going to push a commit to fix that!

FedeDP: Pushed :)

mtardy:
ah also we just "rebase and merge" so if you can squash your commits so that they look like they should be merged on top of main that would be great. Thank you!!

PR #1852, opened 2023-12-06
workflows: add independent go-version for setup-go actions

mtardy:
This reverts commit https://github.com/cilium/tetragon/commit/dd5ddc26585820afd0e4b84a2675e482c8dae236.

This was a good idea but https://github.com/cilium/tetragon/pull/1748 impacts this as we no longer update the patch version in the go.mod directive. Thus the CI is running outdated Go versions.

PR #4052, opened 2025-08-26
chore(ci): rely upon `go-version-file` instead of manually embedding everywhere the go version

FedeDP:
### Description

`actions/setup-go` provides the `go-version-file` input to fetch the go version from the provided `go.mod` : https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

IMHO it is better to rely upon it than to embed everywhere the go version string since it is less error prone (even if the go version is automatically bumped in our case).


.github/workflows/gotests.yml
```diff
@@ -103,17 +102,15 @@ jobs:
...
-      with:
-        # renovate: datasource=golang-version depName=go
-        go-version: '1.25.0'
-
     - name: Checkout code
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
       with:
         path: go/src/github.com/cilium/tetragon/
 
+    - name: Install Go
```
FedeDP, line 110: Moved **after** repo checkout otherwise `go.mod` is not present.

.github/workflows/windows-build-smoke-test.yml
```diff
@@ -137,15 +137,15 @@ jobs:
...
-        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
-        with:
-          go-version: '1.24.2'
-
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
           path: go/src/github.com/cilium/tetragon/
+
+      - name: Install Go
```
FedeDP, line 145: Same as above: moved **after** checkout. Also, bumped to use latest go version instead of 1.24.2.

mtardy:
So we did that and we already reverted it, I would need to find out why and if it still applies

mtardy:
ah this is the reason https://github.com/cilium/tetragon/pull/1852. We don't usually bump our main Go version (for reasons, you should find details in the commits history) with patch release (while we bump the toolchain) but we want the CI to be up to date.

.github/workflows/gotests.yml
mtardy: usually doing this is a best practice for cache reasons, maybe you can put that in a separate commit and merge it?

FedeDP:
Oh sorry, had a quick look at the blame but didn't notice that we already did that.
It makes sense, let me close this one then! Sorry for the noise!

.github/workflows/gotests.yml
FedeDP: I will!

mtardy:
> Oh sorry, had a quick look at the blame but didn't notice that we already did that. It makes sense, let me close this one then! Sorry for the noise!

ah no problem, love the energy especially when it's targeted toward CI and this one is not easy to see why we didn't do that in the first place! Really no prob, keep proposing stuff on things you might see 🤗!

PR #4064, opened 2025-08-28
tetragon: support for current task data

olsajiri:
adding support to resolve and filter on current task data and include it in the event, with spec like:
 
```
spec:
  kprobes:
  - call: "sys_prctl"
    syscall: true
    args:
    - index: 0
      type: "int"
    - index: 1
      type: "int"
    data:
    - source: "current_task"
      resolve: "pid"
      type: "int"
    - source: "current_task"
      resolve: "mm.owner.pid"
      type: "int"
    selectors:
    - matchArgs:
      - index: 1
        operator: "Equal"
        values:
        - 1
      matchData:
      - args:
        - 1
        operator: "NotEqual"
        values:
        - 1
```


olsajiri:
@tixxdz could you please try your tests on this one?

tixxdz:
> @tixxdz could you please try your tests on this one?

Sure will test it today, thank you very much Jiri

mtardy (approved):
> The matchData is just syntax sugar on top of argument filtering, but in reality it just defines filter on specific argument value which happens to be taken from current task.

Thanks I think it looks cleaner that way! :) I'll let Djalal review this.

PR #4070, opened 2025-08-29
new(ci): add gh workflow to sync bpf headers from upstream

### LINKED ISSUE #3306: tetragon: sync api.h with upstream kernels

no

It would be nice if we updated vmlinux and api.h to a match latest kernel. Bonus points for having a script do it or a some better way than just manually syncing headers. Regardless even just updating them manually would be useful.

bpf/include/api does not have all the possible definitions for later kernels. I think it would be useful to just have them all so we can skip writing the extra patch to update them. I also use Tetragon to load experiments sometimes because its an easy way to load BPF programs and distribute code into a K8s network so just having all the apis makes sense to me.

### Code of Conduct

FedeDP:
Fixes #3306

### Description

A new github workflow is created; it runs monthly and:
* generates `bpf/include/vmlinux_generated_$arch` from latest Opensuse tumbleweed kernel images
* generates `bpf/include/bpf_helper_defs.h` from libbpf upstream master

It can also be triggered manually by maintainers.

Also, `bpf/include/api.h` has been split and now all helper definitions live in `bpf/include/bpf_helper_defs.h` to allow us to automatically update them.

See workflow run (on my fork): https://github.com/FedeDP/tetragon/actions/runs/17318134700/job/49165107914, and the resulting PR: https://github.com/FedeDP/tetragon/pull/22.

NOTE: `bpf/include/bpf_helper_defs.h` was already upgraded to latest libbpf.


.github/workflows/sync-api-vmlinux-cron.yaml
```diff
@@ -0,0 +1,99 @@
+name: Sync api.h and vmlinux.h from latest kernel
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 2 1 * *' # first day of each month
+
+jobs:
+  sync-api-vmlinux:
+    runs-on: ${{ (matrix.arch == 'arm64' && 'ubuntu-24.04-arm') || 'ubuntu-24.04' }}
```
FedeDP, line 9: While i could do everything in a single x86_64 ubuntu job, i chose to run it in parallel on both architectures and then have a single job collect built headers and create a PR; this way the permissions to write pull_requests are only available to a small scope job.

.github/workflows/sync-api-vmlinux-cron.yaml
```diff
@@ -0,0 +1,99 @@
...
+        run: |
+          image=$(wget -qO- ${{ matrix.tumbleweed_repo }} | grep "kernel-default-base-*" | head -n1 | awk '{split($0,file,"\""); print file[4]}')
+          wget -O image.rpm ${{ matrix.tumbleweed_repo }}$image
+          mkdir image && cd image
+          rpm2cpio ../image.rpm | cpio -idmv
+          prefix="./kernel-default-base-"
+          suffix=".rpm"
+          version=${image#"$prefix"}
+          version=${version%"$suffix"}
+          echo "KERNEL_VERSION=$version" >> $GITHUB_ENV
```
FedeDP, line 34: We also save the downloaded kernel version in github env to later use it as comment to the generated headers.

.github/workflows/sync-api-vmlinux-cron.yaml
```diff
@@ -0,0 +1,99 @@
...
+          ./extract-vmlinux image/usr/lib/modules/*/vmlinux.xz > decomp-vmlinuz
+
+      - name: Generate updated vmlinux_generated
+        run: |
+          mkdir out
+          bpftool btf dump file decomp-vmlinuz format c > out/vmlinux_generated_${{ matrix.arch }}.h
+          sed -i '1i// Autogenerated from opensuse tumbleweed ${{ env.KERNEL_VERSION }} kernel' out/vmlinux_generated_${{ matrix.arch }}.h
+
+      - name: Sync bpf_helper_defs.h from upstream libbpf repo (x86 only)
+        if: matrix.arch == 'x86'
```
FedeDP, line 54: We only need to do this once, hence the x86 only.

.github/workflows/sync-api-vmlinux-cron.yaml
```diff
@@ -0,0 +1,104 @@
...
+      - name: Download arm64 artifacts
+        uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          name: sync_arm64
+          path: bpf/include/
+
+      - name: Run clang format on generated sources
+        env:
+          LOCAL_CLANG_FORMAT: 1
+        run: make clang-format
```
FedeDP, line 104: Make sure to format the new code!

FedeDP:
Didn't expect to see verifier failures on this one :laughing: anyway, here is a snippet of a failure:
```
logcapture.go:24: time=2025-08-29T08:25:59.548Z level=INFO msg="Opening collection failed, dumping verifier log."
load program: invalid argument:
	0: (bf) r6 = r1
	1: (85) call bpf_get_current_task#35
	2: (bf) r9 = r0
	3: (7b) *(u64 *)(r10 -160) = r6
	4: (61) r8 = *(u32 *)(r6 +8)
	5: (b7) r1 = 0
	6: (63) *(u32 *)(r10 -148) = r1
	7: (bf) r2 = r10
	8: (07) r2 += -148
	9: (18) r1 = 0xffff88812d42d200
	11: (85) call bpf_map_lookup_elem#1
	12: (bf) r7 = r0
	13: (15) if r7 == 0x0 goto pc+2603
	 R0=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R6=ctx(id=0,off=0,imm=0) R7=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R8=inv(id=0,umax_value=4294967295,var_off=(0x0; 0xffffffff)) R9=inv(id=0) R10=fp0,call_-1 fp-160=ctx
	14: (85) call bpf_get_current_pid_tgid#14
	15: (bf) r6 = r0
	16: (85) call bpf_get_current_task#35
	17: (7b) *(u64 *)(r10 -144) = r0
	18: (b7) r1 = 1208
	19: (0f) r0 += r1
	20: (bf) r1 = r10
	21: (07) r1 += -144
	22: (b7) r2 = 8
	23: (bf) r3 = r0
	24: (85) call bpf_probe_read#4
	25: (7b) *(u64 *)(r10 -176) = r6
	26: (79) r3 = *(u64 *)(r10 -144)
	27: (7b) *(u64 *)(r10 -168) = r9
	28: (15) if r3 == 0x0 goto pc+96
...
	525: (77) r3 >>= 16
	526: (73) *(u8 *)(r7 +390) = r3
	 R0=map_value(id=0,off=1152,ks=4,vs=12288,imm=0) R1=inv0 R2=inv(id=0,smin_value=-2147483648,smax_value=2147483647) R3_w=inv(id=0,umax_value=281474976710655,var_off=(0x0; 0xffffffffffff)) R6=ctx(id=0,off=0,umax_value=65535,var_off=(0x0; 0xffff)) R7=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R8=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R9=inv(id=0) R10=fp0,call_-1 fp-160=ctx fp-176=map_value fp-184=inv
	527: (73) *(u8 *)(r7 +388) = r2
	 R0=map_value(id=0,off=1152,ks=4,vs=12288,imm=0) R1=inv0 R2=inv(id=0,smin_value=-2147483648,smax_value=2147483647) R3_w=inv(id=0,umax_value=281474976710655,var_off=(0x0; 0xffffffffffff)) R6=ctx(id=0,off=0,umax_value=65535,var_off=(0x0; 0xffff)) R7=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R8=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R9=inv(id=0) R10=fp0,call_-1 fp-160=ctx fp-176=map_value fp-184=inv
	528: (77) r2 >>= 8
	529: (73) *(u8 *)(r7 +389) = r2
	 R0=map_value(id=0,off=1152,ks=4,vs=12288,imm=0) R1=inv0 R2_w=inv(id=0,umax_value=72057594037927935,var_off=(0x0; 0xffffffffffffff)) R3_w=inv(id=0,umax_value=281474976710655,var_off=(0x0; 0xffffffffffff)) R6=ctx(id=0,off=0,umax_value=65535,var_off=(0x0; 0xffff)) R7=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R8=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R9=inv(id=0) R10=fp0,call_-1 fp-160=ctx fp-176=map_value fp-184=inv
	530: (bf) r2 = r1
	531: (77) r2 >>= 24
	532: (73) *(u8 *)(r7 +379) = r2
	 R0=map_value(id=0,off=1152,ks=4,vs=12288,imm=0) R1=inv0 R2_w=inv0 R3_w=inv(id=0,umax_value=281474976710655,var_off=(0x0; 0xffffffffffff)) R6=ctx(id=0,off=0,umax_value=65535,var_off=(0x0; 0xffff)) R7=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R8=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R9=inv(id=0) R10=fp0,call_-1 fp-160=ctx fp-176=map_value fp-184=inv
	533: (bf) r2 = r1
	534: (77) r2 >>= 16
	535: (73) *(u8 *)(r7 +378) = r2
	 R0=map_value(id=0,off=1152,ks=4,vs=12288,imm=0) R1=inv0 R2_w=inv0 R3_w=inv(id=0,umax_value=281474976710655,var_off=(0x0; 0xffffffffffff)) R6=ctx(id=0,off=0,umax_value=65535,var_off=(0x0; 0xffff)) R7=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R8=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R9=inv(id=0) R10=fp0,call_-1 fp-160=ctx fp-176=map_value fp-184=inv
	536: (73) *(u8 *)(r7 +376) = r1
	 R0=map_value(id=0,off=1152,ks=4,vs=12288,imm=0) R1=inv0 R2_w=inv0 R3_w=inv(id=0,umax_value=281474976710655,var_off=(0x0; 0xffffffffffff)) R6=ctx(id=0,off=0,umax_value=65535,var_off=(0x0; 0xffff)) R7=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R8=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R9=inv(id=0) R10=fp0,call_-1 fp-160=ctx fp-176=map_value fp-184=inv
	537: (77) r1 >>= 8
	538: (73) *(u8 *)(r7 +377) = r1
	 R0=map_value(id=0,off=1152,ks=4,vs=12288,imm=0) R1_w=inv0 R2_w=inv0 R3_w=inv(id=0,umax_value=281474976710655,var_off=(0x0; 0xffffffffffff)) R6=ctx(id=0,off=0,umax_value=65535,var_off=(0x0; 0xffff)) R7=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R8=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R9=inv(id=0) R10=fp0,call_-1 fp-160=ctx fp-176=map_value fp-184=inv
	539: (b7) r8 = 32
	540: (b7) r1 = 8388608
	541: safe
	
	from 503 to 614: R0=map_value(id=0,off=1152,ks=4,vs=12288,imm=0) R1=inv(id=0) R2=inv(id=0) R3=inv32735 R6=ctx(id=0,off=0,umax_value=65535,var_off=(0x0; 0xffff)) R7=map_value(id=0,off=0,ks=4,vs=5808,imm=0) R8=map_value(id=0,off=0,ks=4,vs=32768,imm=0) R9=inv32735 R10=fp0,call_-1 fp-160=ctx fp-176=map_value fp-184=inv
	614: (bf) r9 = r2
	615: (bf) r3 = r9
	616: (0f) r3 += r6
	math between ctx pointer and register with unbounded min value is not allowed
    observer_test_helper.go:421: Load base error: failed prog /home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/bpf/objs/bpf_execve_event.o kern_version 267263 loadInstance: opening collection '/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/bpf/objs/bpf_execve_event.o' failed: program event_execve: load program: invalid argument: math between ctx pointer and register with unbounded min value is not allowed (12893 line(s) omitted)
    filenames.go:49: keeping export file for TestUprobeArgsWithSymbol (/tmp/tetragon.gotest.TestUprobeArgsWithSymbol.299077108.json)
    base.go:196: cleanup: unloading base sensor
level=info msg="Unloading sensor __base__"
```

bpf/include/api.h
FedeDP, line 1: The only differences should be:
* `__maybe_unused` got lost
* `skb_event_output` (BPF_FUNC2) got lost

Possibly one of those is causing verifier issues caught by vmtests?

FedeDP: I'll manually add back skb_event_output to `api.h`. Let's see.

FedeDP: `skb_event_output` readded.
I am now going to add back `__maybe_unused`  too.

FedeDP: Added back.

FedeDP:
Uh also the vmlinux_generated_x86.h is missing some symbols.
I'll try to fix all of the above :)

FedeDP:
> Uh also the vmlinux_generated_x86.h is missing some symbols.

Spotted the issue! Fixing it!

.clang-format-ignore
```diff
@@ -1,6 +1,7 @@
 bpf/include/vmlinux.h
 bpf/include/vmlinux_generated_arm64.h
 bpf/include/vmlinux_generated_x86.h
+bpf/include/bpf_helper_defs.h
```
FedeDP: This is generated. Ignore it.

bpf/include/vmlinux.h
```diff
@@ -1,6 +1,6 @@
 /* SPDX-License-Identifier: (LGPL-2.1 OR BSD-2-Clause) */
-#ifndef __VMLINUX_H__
-#define __VMLINUX_H__
+#ifndef __VMLINUX_GENERIC_H__
```
FedeDP: Since 2 generated headers come both with `__VMLINUX_H__` include guard, rename this one that is easier to modify.

.github/workflows/sync-api-vmlinux-cron.yaml
```diff
@@ -0,0 +1,106 @@
...
+        run: |
+          ./extract-vmlinux image/usr/lib/modules/*/vmlinux.xz > decomp-vmlinuz
+
+      - name: Generate updated vmlinux_generated
+        run: |
+          mkdir out
+          bpftool btf dump file decomp-vmlinuz format c > out/vmlinux_generated_${{ matrix.arch }}.h
+          # Add an autogenerated comment
+          sed -i '1i// Autogenerated from opensuse tumbleweed ${{ env.KERNEL_VERSION }} kernel' out/vmlinux_generated_${{ matrix.arch }}.h
+          sed -i 's/__VMLINUX_H__/__VMLINUX_GENERATED_H__/g' out/vmlinux_generated_${{ matrix.arch }}.h
```
FedeDP, line 51: Update include guard since `__VMLINUX_H__` is already taken by `bpf/include/vmlinux.h`.

bpf/include/bpf_helper_defs.h
```diff
@@ -0,0 +1,212 @@
+// Autogenerated from libbpf sources
+static void *(* const map_lookup_elem)(void *map, const void *key) __maybe_unused = (void *) 1;
+static long (* const map_update_elem)(void *map, const void *key, const void *value, __u64 flags) __maybe_unused = (void *) 2;
+static long (* const map_delete_elem)(void *map, const void *key) __maybe_unused = (void *) 3;
```
FedeDP: Both of these functions have changed their return type from `int` to `long`.
Since libbpf is the source of truth, this should be ok (also: https://docs.ebpf.io/linux/helper-function/bpf_map_update_elem/)

FedeDP: Also `probe_read`, `probe_read_str`, `probe_read_kernel` and `probe_read_user`

FedeDP:
I think that the vmtests failures are linked to the new signature for some bpf helpers, as explained here: https://github.com/cilium/tetragon/pull/4070#discussion_r2310038045
EDIT: this is the list of **used** helpers whose usage needs to be adapted to the new signature (return type `int` -> `long`):
```
static int BPF_FUNC(map_update_elem, void *map, const void *key,
		    const void *value, uint32_t flags);
static int BPF_FUNC(map_delete_elem, void *map, const void *key);
static int BPF_FUNC(probe_read, void *dst, uint32_t size, const void *src);
static int BPF_FUNC(probe_read_str, void *dst, int size, const void *src);
static int BPF_FUNC(probe_read_kernel, void *dst, uint32_t size, const void *src);
static int BPF_FUNC(probe_read_user, void *dst, uint32_t size, const void *src);
static int BPF_FUNC(perf_event_output, void *ctx, void *map, uint64_t flags, void *data, uint64_t size);
```

FedeDP:
So, reverting to previous signatures for bpf helpers is working. Probably cannot use new definitions on old kernels.
Will fix the workflow to patch those signatures to revert the `static long` change.

.github/workflows/sync-api-vmlinux-cron.yaml
```diff
@@ -0,0 +1,115 @@
...
+
+      - name: Sync bpf_helper_defs.h from upstream libbpf repo (x86 only)
+        if: matrix.arch == 'x86'
+        env:
+          PATCH_LONG_INT: 'map_update_elem;map_delete_elem;probe_read;perf_event_output;probe_read_str;probe_read_user;probe_read_kernel'
```
FedeDP, line 56: On older kernels using new signature for these helpers causes troubles with the verifier.
This env variable is later used to patch each of the element to return `int` instead of `long`.

.github/workflows/sync-api-vmlinux-cron.yaml
FedeDP: (use local clang-format!)

.github/workflows/sync-api-vmlinux-cron.yaml
```diff
@@ -0,0 +1,117 @@
...
+          # drop `bpf_` prefix from all helpers, since we don't use it
+          sed -i 's/const bpf_/const /g' out/bpf_helper_defs.h
+          # add __maybe_unused
+          sed -i 's/) =/) __maybe_unused =/g' out/bpf_helper_defs.h
+          # Add an autogenerated comment
+          sed -i '1i// Autogenerated from libbpf sources' out/bpf_helper_defs.h
+          # Patch various long return value to old signatures (int)
+          # to avoid eBPF verifier issues on old kernels
+          IFS=';' read -r -a to_be_patched <<< "${{ env.PATCH_LONG_INT }}"
+          for element in "${to_be_patched[@]}"; do sed -i "s/static long (\* const $element)/static int (\* const $element)/g" out/bpf_helper_defs.h; done
```
kkourt, line 70: In my eyes, these steps seem brittle and easy to break in the future.

Given the above, I think it's worth taking a step back and asking, is this automation really needed?
What we have done so far is to update the headers as we use the relevant features in our code, which worked quite well, IMO. 

I'm CCing: @kevsecurity @olsajiri @mtardy for any thoughts on their side.

FedeDP: I agree with you; while the vmlinux are worth auto updating, the bpf helpers are a bit complex and possibly not worth it.
Note however that this "patch" step is pretty easy to fix up in the eventuality that upstream libbpf changes its format. At the same time, i tried fixing our code to use `long` return code for those helpers, and it broke old kernels verifier anyway (as expected because on those kernels the helpers returned an `int`).

All in all, i think maintainers have a better vision than me about this one and i trust your opinion :)

kevsecurity: I don't really understand the aim of automatically updating either the helpers or the vmlinux.h. I don't think it is too much trouble to update either as required, but if a tool would help, then great. I just don't really see this as a fully-automated task.

FedeDP: I have no strong opinion on this one; i created the PR to fix https://github.com/cilium/tetragon/issues/3306.
If maintainers feel like this is not really a thing they want, i am :100: on board with that :+1:

mtardy: I think I would agree, with the experience of maintaining CI, maybe that's a bit fragile and would require too much attention in comparison to just sync manually when we need it. I'm not sure it provides much value to sync it that frequently. I'm sorry we make this comment after you spent some consequent time on this... but at least now we know it's not trivial thanks to your work.

Let's close the initial issue https://github.com/cilium/tetragon/issues/3306. Sorry and thanks for the work. Maybe we'll change our mind in the future and we'll have this somewhere ready :)

FedeDP: No prob, i'll close this one then :)

PR #4074, opened 2025-09-02
policy counters

kkourt:
See commits.

```release-note
tracingpolicy: add counters about actions performed for every policy
```

### DIFF
--- a/bpf/alignchecker/bpf_alignchecker.c
+++ b/bpf/alignchecker/bpf_alignchecker.c
@@ -7,6 +7,7 @@
 #include "lib/bpf_cred.h"
 #include "process/retprobe_map.h"
 #include "process/types/basic.h"
+#include "policy_stats.h"
 
 // event messages
 struct msg_generic_kprobe _msg_generic_kprobe;
@@ -24,3 +25,4 @@ struct event_config _event_config;
 struct tetragon_conf _tetragon_conf;
 struct cgroup_tracking_value _cgroup_tracking_value;
 struct kernel_stats _kernel_stats;
+struct policy_stats _policy_stats;
--- a/pkg/alignchecker/alignchecker.go
+++ b/pkg/alignchecker/alignchecker.go
@@ -13,6 +13,7 @@ import (
 	"github.com/cilium/tetragon/pkg/api/processapi"
 	"github.com/cilium/tetragon/pkg/api/testapi"
 	"github.com/cilium/tetragon/pkg/api/tracingapi"
+	"github.com/cilium/tetragon/pkg/policystats"
 	"github.com/cilium/tetragon/pkg/sensors/cgroup/cgrouptrackmap"
 	"github.com/cilium/tetragon/pkg/sensors/config/confmap"
 	"github.com/cilium/tetragon/pkg/sensors/exec/execvemap"
@@ -36,6 +37,9 @@ var defaultChecks = map[string][]any{
 
 	// metrics
 	"kernel_stats": {processapi.KernelStats{}},
+
+	// policy stats
+	"policy_stats": {policystats.PolicyStats{}},
 }
 
 // CheckStructAlignmentsDefault calls CheckStructAlignments with the default alignment defaultChecks.


pkg/policystats/policystats.go
```diff
@@ -0,0 +1,56 @@
...
+	PolicySignal                PolicyAction = 2
+	PolicyMonitorSignal         PolicyAction = 3
+	PolicyOverride              PolicyAction = 4
+	PolicyMonitorOverride       PolicyAction = 5
+	PolicyNotifyEnforcer        PolicyAction = 6
+	PolicyMonitorNotifyEnforcer PolicyAction = 7
+	PolicyActionsNr                          = 8
+)
+
+type PolicyStats struct {
```
tpapagian, line 34: Nit: It could make sense to compare struct policy_stats and PolicyStats in [alignchecker](https://github.com/cilium/tetragon/blob/6a6bae661e3d2fb854b4788828dcf3d89e070628/pkg/alignchecker/alignchecker.go#L38). This could help to avoid adding things and forget to update one of the structs.

bpf/lib/policy_stats.h
```diff
@@ -0,0 +1,32 @@
...
+	POLICY_MONITOR_NOTIFY_ENFORCER = 7, /* policy did not notify the enforcer because it was in monitor mode */
+	POLICY_NACTIONS_,
+};
+
+struct policy_stats {
+	u64 act_cnt[POLICY_NACTIONS_];
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_ARRAY);
```
tpapagian, line 29: Thinking out loud: Could it make sense to user a per-cpu array instead to avoid contention with `lock_add`? Possibly not as generating events is not the common path?

tpapagian (approved):

tixxdz (approved):

pkg/policystats/policystats.go
will-isovalent: +1 this needs to be added to alignchecker imo

pkg/policystats/policystats.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+const (
+	PolicyStatsMapName = "policy_stats"
+)
+
+type PolicyAction uint8
+
+const (
+	// NB: values below should match the ones in bpf/lib/policy_conf.h
```
will-isovalent: This comment says they need to match policy_conf.h, but the actual file where they're defined is policy_stats.h

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -71,7 +71,7 @@ func (pi *policyInfo) policyConfMap(prog *program.Program) *program.Map {
 	if pi.policyConf != nil {
 		return program.MapUserFrom(pi.policyConf)
 	}
-	pi.policyConf = program.MapBuilderPolicy("policy_conf", prog)
+	pi.policyConf = program.MapBuilderPolicy(policyconf.PolicyConfMapName, prog)
```
will-isovalent, line 85: Minor nit (not related to this line but to the commit itself). The commit has a typo in the title. `policyfonf -> policyconf`

api/v1/tetragon/sensors.proto
```diff
@@ -55,6 +55,27 @@ enum TracingPolicyMode {
...
+  uint64 monitor_override = 5;
+  // number of enforcer notifications triggered from the policy
+  uint64 notify_enforcer = 6;
+  // number of enforcer notifications that did not occur because the policy was in monitor mode
+  uint64 monitor_notify_enforcer = 7;
+}
+
+message TracingPolicyStats {
+  TracingPolicyActionCounters action_counters = 1;
+}
```
will-isovalent, line 77: Just thinking out loud here, but maybe we want a comment somewhere in the BPF header instructing developers to make sure they update protobuf in addition to the userspace map reader when adding a new action counter.

cmd/tetra/common/utils.go
```diff
@@ -70,7 +71,8 @@ func PrintTracingPolicies(output io.Writer, policies []*tetragon.TracingPolicySt
 			}
 		}
 
-		fmt.Fprintf(w, "%d\t%s\t%s\t%d\t%s\t%s\t%s\t%s\t\n",
+		counters := pol.GetStats().GetActionCounters()
```
will-isovalent, line 74: Do we need nil checks here?

pkg/policystats/policystats.go
kkourt: Added alignchecker support. Needed support for multiple types, but it wasn't too complicated.

bpf/lib/policy_stats.h
kkourt: I thought about it, but then user-space ends up being more complicated. I would leave it as a potential future optimization.

pkg/policystats/policystats.go
kkourt: Thanks, fixed.

pkg/sensors/tracing/policyhandler_linux.go
kkourt: Thanks! Fixed

api/v1/tetragon/sensors.proto
kkourt: makes sense, added.

cmd/tetra/common/utils.go
kkourt: I don't think so, because the gRPC methods handle it:
```
func (x *TracingPolicyStatus) GetStats() *TracingPolicyStats {
	if x != nil {
		return x.Stats
	}
	return nil
}
```

PR #4075, opened 2025-09-02
tetragon: add bpf ring buffer support

kevsecurity:
### Description
Adds support for bpf ring buffer and sets that as the default from kernels v5.11 onwards. CLI switch to prefer perf ring buffer. The bpf ring buffer has two advantages over the perf ring buffer; better memory usage and guaranteed event order. The drawback is potentially efficiency as the multiple-provider aspect requires synchronisation.

### Changelog

```release-note
Adds support for bpf ring buffer and sets that as the default from kernels v5.11 onwards.
```

### DIFF
--- a/pkg/vmtests/skip.go
+++ b/pkg/vmtests/skip.go
@@ -17,7 +17,7 @@ type skipRule struct {
 
 // We should probably have this in a json file, but for now we keep it here
 var rules = []skipRule{
-	skipRule{TestNameRe: "pkg.sensors.tracing.TestLoader", KernelRe: "(6\\.6|6\\.1)"},
+	skipRule{TestNameRe: "pkg.sensors.tracing.TestLoader", KernelRe: ""},
 	skipRule{TestNameRe: "pkg.sensors.exec.TestProcessCacheInterval", KernelRe: ""},
 	skipRule{TestNameRe: "pkg.watcher.TestFastK8s", KernelRe: ""},
 }


pkg/sensors/base/base_linux.go
```diff
@@ -33,6 +41,35 @@ func GetDefaultMaps() []*program.Map {
 		MatchBinariesGenMap,
 		ErrMetricsMap,
 	}
+	// The BPF ring buffer is available from v5.8, but rather than add another set of
```
dwindsor, line 40: qq does this leave kernels `5.8>=`, `<5.11` unsupported? not sure it matters much just curious

pkg/observer/observer.go
```diff
@@ -149,8 +150,9 @@ func (k *Observer) getRBQueueSize() int {
 // notified of their corresponding events.
 type Observer struct {
 	/* Configuration */
-	listeners  map[Listener]struct{}
-	PerfConfig *bpf.PerfEventConfig
+	listeners      map[Listener]struct{}
+	PerfConfig     *bpf.PerfEventConfig
+	RingBufMapName string
```
dwindsor: qq, have we given any thought to having multiple CPUs service the same ring? Definitely not for this PR, just curious. There may be a benefit to doing it, if we haven't checked yet. cc @kkourt

kevsecurity:
Sorry, this was supposed to be created as draft as I wanted an early check on how the CI tests would fair. Feel free to unsubscribe as there might be many updates.

dwindsor:
> Sorry, this was supposed to be created as draft as I wanted an early check on how the CI tests would fair. Feel free to unsubscribe as there might be many updates.

My fault! =)

pkg/sensors/base/base_linux.go
kevsecurity: Yes, all kernels <v5.11 will use the perf ring buffer. Kernels from 5.11+ will default to the bpf ring buffer, but can be switched to perf ring buffer. If someone has a use case for using the bpf ring buffer with kernels <v5.11 then it is easy to add an extra set of programs. It might be nice if we avoid them if we don't need them though.

pkg/config/config_linux.go
```diff
@@ -147,6 +147,14 @@ func EnableRhel7Progs() bool {
 	return (int64(kernelVer) < kernels.KernelStringToNumeric("3.11.0"))
 }
 
+func EnableV511Progs() bool {
```
mtardy, line 206: could you maybe add a comment to tell how it's related to the ring buffer?

bpf/lib/bpf_event.h
```diff
@@ -16,4 +17,31 @@ struct {
...
 } tcpmon_map SEC(".maps");
 
+#ifdef __V511_BPF_PROG
+// The ring buffer needs to have a size that is a multiple of a page size, and is a power of 2.
+// Zero isn't acceptable, so we arbitrarily choose 4K as that is a common/default page size.
+// This will be resized in user space.
+struct {
+	__uint(type, BPF_MAP_TYPE_RINGBUF);
+	__uint(max_entries, 4096);
+} tg_rb_events SEC(".maps");
```
mtardy: nit: I see there's the related comment below, but we have this convention so that you can spot at quick glance if sizes are definitive or not.
```suggestion
struct {
	__uint(type, BPF_MAP_TYPE_RINGBUF);
	__uint(max_entries, 4096); // This will be resized in user space.
} tg_rb_events SEC(".maps");
```

pkg/observer/observer.go
mtardy: nit
```suggestion
	RingBufMapPath string
```

pkg/observer/observer_linux.go
```diff
@@ -74,6 +76,22 @@ func (k *Observer) RunEvents(stopCtx context.Context, ready func()) error {
 		return fmt.Errorf("creating perf array reader failed: %w", err)
 	}
 
+	var ringBufReader *ringbuf.Reader
+	var ringBufMap *ebpf.Map
+	if kernels.MinKernelVersion("5.11") {
```
mtardy: maybe you want to use your newly added `EnableV511Progs()`? Maybe we can change the name to something ring buf related?

pkg/observer/observer_linux.go
```diff
@@ -119,6 +137,35 @@ func (k *Observer) RunEvents(stopCtx context.Context, ready func()) error {
 		}
 	}()
 
+	if kernels.MinKernelVersion("5.11") {
```
mtardy: here as well `EnableV511Progs()`? and at close step below

olsajiri:
@kevsecurity overall looks good, will check more later.. but before we switch, do we have performance data to justify that?

pkg/observer/observer_linux.go
```diff
@@ -80,53 +97,82 @@ func (k *Observer) RunEvents(stopCtx context.Context, ready func()) error {
...
-					k.log.Warn("Reading bpf events failed", "errors", errorCnt, logfields.Error, err)
-				}
-			} else {
-				if len(record.RawSample) > 0 {
-					select {
-					case eventsQueue <- &record:
-					default:
-						// eventsQueue channel is full, drop the event
-						queueLost.Inc()
+	if useBpfRingBuffer {
```
FedeDP: I think it would be great to introduce an interface on top of `ringBufReader` and `perfReader` here to avoid the doubled, very similar, `go func`; wdyt?

Also, perhaps it's correct, but:
```
if record.LostSamples > 0 {
	RingbufLost.Add(float64(record.LostSamples))
}
```
is only present in the perf reader path.

pkg/observer/observer.go
FedeDP: In this regard, please note that `ringbuf` maps data pages twice, making process `rss` account twice for them.
This means increased OOM score in case of low memory on the system.

Refer to: https://lore.kernel.org/bpf/CAEf4BzbYoGY2e_Nf1EOgzDdjGOLbSjJ5OmHRDFnYUA_VDYD6zQ@mail.gmail.com/T/#t for more info.

pkg/observer/observer_linux.go
kevsecurity: As you note, the two functions are already different. I imagine them deviating further over time.

FedeDP: Thanks, that makes sense!

pkg/observer/observer.go
kevsecurity: The size of the ring buffer is user-configured from the CLI switches. Using the defaults, the oom_score is the same with this series, whether --use-perf-ring-buffer is true or false, as it is on the main branch (also true with a variety of ring buffer sizes). I assume this means it is no more likely to be killed than the current main branch. (Only checked on 5.15, but did use a tracing policy that generated lots of rapid events to touch all of the ring buffer memory.)
Even though the memory contributes to the RSS, maybe the oom_score ignores them as they are shared with the kernel?

FedeDP: Perhaps is due to some later kernel patch involving oom score computation? I remember i saw something about it; let me look it up!

kevsecurity:
> @kevsecurity overall looks good, will check more later.. but before we switch, do we have performance data to justify that?

For those with perf concerns in situations where the perf ring buffer is better, they can use the --use-perf-ring-buffer switch. We should consider this when benchmarks and comparisons are made.

Meta reported that the perf was better with the bpf ring buffer and it handled spikes better.

I ran a stress test comprising 1024 iterations of dd copying 32K from /dev/urandom to files in /tmp, while running Tetragon with raw_syscalls tracing policy example. The idea here was to make the ring buffer busy (reporting every syscall), and then time the script running the dds. The script itself takes about 1 minute to run and there were no significant differences between the time taken with the perf ring or with the bpf ring. Equally, CPU usage by Tetragon before running the script was about 184% (2xCPU KVM), which dropped to about 135% while running the script, implying resources were starved. If anything, Tetragon with bpf ring got slightly more CPU implying the underlying ring was more efficient. More tests are welcome, but on the surface it looks like it will make no difference.

The main justification is that the bpf ring buffer will permit sending events from bpf callbacks, which the perf ring will not.

olsajiri:
> The main justification is that the bpf ring buffer will permit sending events from bpf callbacks, which the perf ring will not.

do you have example of such callback? thanks ... edited ah timer ;-)

pkg/config/config_linux.go
```diff
@@ -142,11 +154,54 @@ func GenericLsmObjs() (string, string) {
 	return "bpf_generic_lsm_core.o", "bpf_generic_lsm_output.o"
 }
 
+func CGroupMkdirObj() string {
```
dwindsor, line 166: Not for this PR but we should consider creating a generic `SelectBpfObj()` interface here

dwindsor:
Overall this looks really good to me! 

> > @kevsecurity overall looks good, will check more later.. but before we switch, do we have performance data to justify that?
> 
> For those with perf concerns in situations where the perf ring buffer is better, they can use the --use-perf-ring-buffer switch. We should consider this when benchmarks and comparisons are made.
> 
> Meta reported that the perf was better with the bpf ring buffer and it handled spikes better.
> 
> I ran a stress test comprising 1024 iterations of dd copying 32K from /dev/urandom to files in /tmp, while running Tetragon with raw_syscalls tracing policy example. The idea here was to make the ring buffer busy (reporting every syscall), and then time the script running the dds. The script itself takes about 1 minute to run and there were no significant differences between the time taken with the perf ring or with the bpf ring. Equally, CPU usage by Tetragon before running the script was about 184% (2xCPU KVM), which dropped to about 135% while running the script, implying resources were starved. If anything, Tetragon with bpf ring got slightly more CPU implying the underlying ring was more efficient. More tests are welcome, but on the surface it looks like it will make no difference.
> 
> The main justification is that the bpf ring buffer will permit sending events from bpf callbacks, which the perf ring will not.

qq did you do any test for dropped events in perf vs ringbuf during heavy load? No need to hold up the PR just curious if you looked at this

kevsecurity:
> qq did you do any test for dropped events in perf vs ringbuf during heavy load? No need to hold up the PR just curious if you looked at this

I didn't, so I ran some tests just now to compare. In all cases I ran the same raw_syscalls tracing policy. I took a couple of measurements a few seconds after starting up, a couple of seconds apart (not very scientific, admittedly). I then ran the stress test and took a measurement after.

main branch, before stress:
```
root@e338dac4c545:~# curl localhost:2112/metrics 2>/dev/null | grep -i missed_events\\\|overflowed_events\\\|ringbuf                   
# HELP tetragon_bpf_missed_events_total Number of Tetragon perf events that are failed to be sent from the kernel.                     
# TYPE tetragon_bpf_missed_events_total counter                                                                                        
tetragon_bpf_missed_events_total{error="ENOENT",msg_op="14"} 380                                                                       
tetragon_bpf_missed_events_total{error="ENOENT",msg_op="23"} 5                                                                         
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="14"} 4077                                                                      
# HELP tetragon_notify_overflowed_events_total The total number of events dropped because listener buffer was full                     
# TYPE tetragon_notify_overflowed_events_total counter                                                                                 
tetragon_notify_overflowed_events_total 3711                                                                                           
# HELP tetragon_observer_ringbuf_errors_total Number of errors when reading Tetragon ring buffer.                                      
# TYPE tetragon_observer_ringbuf_errors_total counter                                                                                  
tetragon_observer_ringbuf_errors_total 0                                                                                               
# HELP tetragon_observer_ringbuf_events_lost_total Number of perf events Tetragon ring buffer lost.                                    
# TYPE tetragon_observer_ringbuf_events_lost_total counter                                                                             
tetragon_observer_ringbuf_events_lost_total 4077                                                                                       
# HELP tetragon_observer_ringbuf_events_received_total Number of perf events Tetragon ring buffer received.                            
# TYPE tetragon_observer_ringbuf_events_received_total counter                                                                         
tetragon_observer_ringbuf_events_received_total 273705                                                                                 
# HELP tetragon_observer_ringbuf_queue_events_lost_total Number of perf events Tetragon ring buffer events queue lost.                 
# TYPE tetragon_observer_ringbuf_queue_events_lost_total counter                                                                       
tetragon_observer_ringbuf_queue_events_lost_total 193067                                                                               
# HELP tetragon_observer_ringbuf_queue_events_received_total Number of perf events Tetragon ring buffer events queue received.         
# TYPE tetragon_observer_ringbuf_queue_events_received_total counter                                                                   
tetragon_observer_ringbuf_queue_events_received_total 18460                                                                            
root@e338dac4c545:~# curl localhost:2112/metrics 2>/dev/null | grep -i missed_events\\\|overflowed_events\\\|ringbuf                   
# HELP tetragon_bpf_missed_events_total Number of Tetragon perf events that are failed to be sent from the kernel.                     
# TYPE tetragon_bpf_missed_events_total counter                                                                                        
tetragon_bpf_missed_events_total{error="ENOENT",msg_op="14"} 380                                                                       
tetragon_bpf_missed_events_total{error="ENOENT",msg_op="23"} 5                                                                         
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="14"} 10472                                                                     
# HELP tetragon_notify_overflowed_events_total The total number of events dropped because listener buffer was full                     
# TYPE tetragon_notify_overflowed_events_total counter                                                                                 
tetragon_notify_overflowed_events_total 18827                                                                                          
# HELP tetragon_observer_ringbuf_errors_total Number of errors when reading Tetragon ring buffer.                                      
# TYPE tetragon_observer_ringbuf_errors_total counter                                                                                  
tetragon_observer_ringbuf_errors_total 0                                                                                               
# HELP tetragon_observer_ringbuf_events_lost_total Number of perf events Tetragon ring buffer lost.                                    
# TYPE tetragon_observer_ringbuf_events_lost_total counter                                                                             
tetragon_observer_ringbuf_events_lost_total 10472                                                                                      
# HELP tetragon_observer_ringbuf_events_received_total Number of perf events Tetragon ring buffer received.                            
# TYPE tetragon_observer_ringbuf_events_received_total counter                                                                         
tetragon_observer_ringbuf_events_received_total 549324                                                                                 
# HELP tetragon_observer_ringbuf_queue_events_lost_total Number of perf events Tetragon ring buffer events queue lost.                 
# TYPE tetragon_observer_ringbuf_queue_events_lost_total counter                                                                       
tetragon_observer_ringbuf_queue_events_lost_total 445867                                                                               
# HELP tetragon_observer_ringbuf_queue_events_received_total Number of perf events Tetragon ring buffer events queue received.         
# TYPE tetragon_observer_ringbuf_queue_events_received_total counter                                                                   
tetragon_observer_ringbuf_queue_events_received_total 37995                                                                            
```

main branch, after stress:
```
root@e338dac4c545:~# curl localhost:2112/metrics 2>/dev/null | grep -i missed_events\\\|overflowed_events\\\|ringbuf                   
# HELP tetragon_bpf_missed_events_total Number of Tetragon perf events that are failed to be sent from the kernel.                     
# TYPE tetragon_bpf_missed_events_total counter                                                                                        
tetragon_bpf_missed_events_total{error="ENOENT",msg_op="14"} 380                                                                       
tetragon_bpf_missed_events_total{error="ENOENT",msg_op="23"} 5                                                                         
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="14"} 133168                                                                    
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="23"} 5                                                                         
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="5"} 6                                                                          
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="7"} 4                                                                          
# HELP tetragon_notify_overflowed_events_total The total number of events dropped because listener buffer was full                     
# TYPE tetragon_notify_overflowed_events_total counter                                                                                 
tetragon_notify_overflowed_events_total 246697                                                                                         
# HELP tetragon_observer_ringbuf_errors_total Number of errors when reading Tetragon ring buffer.                                      
# TYPE tetragon_observer_ringbuf_errors_total counter                                                                                  
tetragon_observer_ringbuf_errors_total 0                                                                                               
# HELP tetragon_observer_ringbuf_events_lost_total Number of perf events Tetragon ring buffer lost.                                    
# TYPE tetragon_observer_ringbuf_events_lost_total counter                                                                             
tetragon_observer_ringbuf_events_lost_total 133183                                                                                     
# HELP tetragon_observer_ringbuf_events_received_total Number of perf events Tetragon ring buffer received.                            
# TYPE tetragon_observer_ringbuf_events_received_total counter                                                                         
tetragon_observer_ringbuf_events_received_total 5.490327e+06                                                                           
# HELP tetragon_observer_ringbuf_queue_events_lost_total Number of perf events Tetragon ring buffer events queue lost.                 
# TYPE tetragon_observer_ringbuf_queue_events_lost_total counter                                                                       
tetragon_observer_ringbuf_queue_events_lost_total 5.087394e+06                                                                         
# HELP tetragon_observer_ringbuf_queue_events_received_total Number of perf events Tetragon ring buffer events queue received.         
# TYPE tetragon_observer_ringbuf_queue_events_received_total counter                                                                   
tetragon_observer_ringbuf_queue_events_received_total 337351                                                                           
```

this PR with perf ring buffer, before stress:
```
root@e338dac4c545:~# curl localhost:2112/metrics 2>/dev/null | grep -i missed_events\\\|overflowed_events\\\|ringbuf                   
# HELP tetragon_bpf_missed_events_total Number of Tetragon perf events that are failed to be sent from the kernel.                     
# TYPE tetragon_bpf_missed_events_total counter                                                                                        
tetragon_bpf_missed_events_total{error="ENOENT",msg_op="14"} 201                                                                       
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="14"} 4826                                                                      
# HELP tetragon_notify_overflowed_events_total The total number of events dropped because listener buffer was full                     
# TYPE tetragon_notify_overflowed_events_total counter                                                                                 
tetragon_notify_overflowed_events_total 4171                                                                                           
# HELP tetragon_observer_ringbuf_errors_total Number of errors when reading Tetragon ring buffer.                                      
# TYPE tetragon_observer_ringbuf_errors_total counter                                                                                  
tetragon_observer_ringbuf_errors_total 0                                                                                               
# HELP tetragon_observer_ringbuf_events_lost_total Number of perf events Tetragon ring buffer lost.                                    
# TYPE tetragon_observer_ringbuf_events_lost_total counter                                                                             
tetragon_observer_ringbuf_events_lost_total 4826                                                                                       
# HELP tetragon_observer_ringbuf_events_received_total Number of perf events Tetragon ring buffer received.                            
# TYPE tetragon_observer_ringbuf_events_received_total counter                                                                         
tetragon_observer_ringbuf_events_received_total 261631                                                                                 
# HELP tetragon_observer_ringbuf_queue_events_lost_total Number of perf events Tetragon ring buffer events queue lost.                 
# TYPE tetragon_observer_ringbuf_queue_events_lost_total counter                                                                       
tetragon_observer_ringbuf_queue_events_lost_total 177102                                                                               
# HELP tetragon_observer_ringbuf_queue_events_received_total Number of perf events Tetragon ring buffer events queue received.         
# TYPE tetragon_observer_ringbuf_queue_events_received_total counter                                                                   
tetragon_observer_ringbuf_queue_events_received_total 19052                                                                            
root@e338dac4c545:~# curl localhost:2112/metrics 2>/dev/null | grep -i missed_events\\\|overflowed_events\\\|ringbuf                   
# HELP tetragon_bpf_missed_events_total Number of Tetragon perf events that are failed to be sent from the kernel.                     
# TYPE tetragon_bpf_missed_events_total counter                                                                                        
tetragon_bpf_missed_events_total{error="ENOENT",msg_op="14"} 201                                                                       
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="14"} 11048                                                                     
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="5"} 1                                                                          
# HELP tetragon_notify_overflowed_events_total The total number of events dropped because listener buffer was full                     
# TYPE tetragon_notify_overflowed_events_total counter                                                                                 
tetragon_notify_overflowed_events_total 19876                                                                                          
# HELP tetragon_observer_ringbuf_errors_total Number of errors when reading Tetragon ring buffer.                                      
# TYPE tetragon_observer_ringbuf_errors_total counter                                                                                  
tetragon_observer_ringbuf_errors_total 0                                                                                               
# HELP tetragon_observer_ringbuf_events_lost_total Number of perf events Tetragon ring buffer lost.                                    
# TYPE tetragon_observer_ringbuf_events_lost_total counter                                                                             
tetragon_observer_ringbuf_events_lost_total 11049                                                                                      
# HELP tetragon_observer_ringbuf_events_received_total Number of perf events Tetragon ring buffer received.                            
# TYPE tetragon_observer_ringbuf_events_received_total counter                                                                         
tetragon_observer_ringbuf_events_received_total 520404                                                                                 
# HELP tetragon_observer_ringbuf_queue_events_lost_total Number of perf events Tetragon ring buffer events queue lost.                 
# TYPE tetragon_observer_ringbuf_queue_events_lost_total counter                                                                       
tetragon_observer_ringbuf_queue_events_lost_total 415922                                                                               
# HELP tetragon_observer_ringbuf_queue_events_received_total Number of perf events Tetragon ring buffer events queue received.         
# TYPE tetragon_observer_ringbuf_queue_events_received_total counter                                                                   
tetragon_observer_ringbuf_queue_events_received_total 39047                                                                            
```

this PR with perf ring buffer, after stress:
```
root@e338dac4c545:~# curl localhost:2112/metrics 2>/dev/null | grep -i missed_events\\\|overflowed_events\\\|ringbuf                   
# HELP tetragon_bpf_missed_events_total Number of Tetragon perf events that are failed to be sent from the kernel.                     
# TYPE tetragon_bpf_missed_events_total counter                                                                                        
tetragon_bpf_missed_events_total{error="ENOENT",msg_op="14"} 201                                                                       
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="14"} 97153                                                                     
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="23"} 8                                                                         
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="5"} 8                                                                          
tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="7"} 4                                                                          
# HELP tetragon_notify_overflowed_events_total The total number of events dropped because listener buffer was full                     
# TYPE tetragon_notify_overflowed_events_total counter
tetragon_notify_overflowed_events_total 222975
# HELP tetragon_observer_ringbuf_errors_total Number of errors when reading Tetragon ring buffer.
# TYPE tetragon_observer_ringbuf_errors_total counter
tetragon_observer_ringbuf_errors_total 0
# HELP tetragon_observer_ringbuf_events_lost_total Number of perf events Tetragon ring buffer lost.
# TYPE tetragon_observer_ringbuf_events_lost_total counter
tetragon_observer_ringbuf_events_lost_total 97173
# HELP tetragon_observer_ringbuf_events_received_total Number of perf events Tetragon ring buffer received.
# TYPE tetragon_observer_ringbuf_events_received_total counter
tetragon_observer_ringbuf_events_received_total 5.294368e+06
# HELP tetragon_observer_ringbuf_queue_events_lost_total Number of perf events Tetragon ring buffer events queue lost.
# TYPE tetragon_observer_ringbuf_queue_events_lost_total counter
tetragon_observer_ringbuf_queue_events_lost_total 4.917314e+06
# HELP tetragon_observer_ringbuf_queue_events_received_total Number of perf events Tetragon ring buffer events queue received.
# TYPE tetragon_observer_ringbuf_queue_events_received_total counter
tetragon_observer_ringbuf_queue_events_received_total 308629
```

this PR with bpf ring buffer, before stress:
```
root@e338dac4c545:~# curl localhost:2112/metrics 2>/dev/null | grep -i missed_events\\\|overflowed_events\\\|ringbuf
# HELP tetragon_bpf_missed_events_total Number of Tetragon perf events that are failed to be sent from the kernel.
# TYPE tetragon_bpf_missed_events_total counter
tetragon_bpf_missed_events_total{error="unknown",msg_op="14"} 465
tetragon_bpf_missed_events_total{error="unknown",msg_op="5"} 1
# HELP tetragon_notify_overflowed_events_total The total number of events dropped because listener buffer was full
# TYPE tetragon_notify_overflowed_events_total counter
tetragon_notify_overflowed_events_total 0
# HELP tetragon_observer_ringbuf_errors_total Number of errors when reading Tetragon ring buffer.
# TYPE tetragon_observer_ringbuf_errors_total counter
tetragon_observer_ringbuf_errors_total 0
# HELP tetragon_observer_ringbuf_events_lost_total Number of perf events Tetragon ring buffer lost.
# TYPE tetragon_observer_ringbuf_events_lost_total counter
tetragon_observer_ringbuf_events_lost_total 0
# HELP tetragon_observer_ringbuf_events_received_total Number of perf events Tetragon ring buffer received.
# TYPE tetragon_observer_ringbuf_events_received_total counter
tetragon_observer_ringbuf_events_received_total 238622
# HELP tetragon_observer_ringbuf_queue_events_lost_total Number of perf events Tetragon ring buffer events queue lost.
# TYPE tetragon_observer_ringbuf_queue_events_lost_total counter
tetragon_observer_ringbuf_queue_events_lost_total 163442
# HELP tetragon_observer_ringbuf_queue_events_received_total Number of perf events Tetragon ring buffer events queue received.
# TYPE tetragon_observer_ringbuf_queue_events_received_total counter
tetragon_observer_ringbuf_queue_events_received_total 11762
root@e338dac4c545:~# curl localhost:2112/metrics 2>/dev/null | grep -i missed_events\\\|overflowed_events\\\|ringbuf
# HELP tetragon_bpf_missed_events_total Number of Tetragon perf events that are failed to be sent from the kernel.
# TYPE tetragon_bpf_missed_events_total counter
tetragon_bpf_missed_events_total{error="unknown",msg_op="14"} 2003
tetragon_bpf_missed_events_total{error="unknown",msg_op="5"} 1
# HELP tetragon_notify_overflowed_events_total The total number of events dropped because listener buffer was full
# TYPE tetragon_notify_overflowed_events_total counter
tetragon_notify_overflowed_events_total 14885
# HELP tetragon_observer_ringbuf_errors_total Number of errors when reading Tetragon ring buffer.
# TYPE tetragon_observer_ringbuf_errors_total counter
tetragon_observer_ringbuf_errors_total 0
# HELP tetragon_observer_ringbuf_events_lost_total Number of perf events Tetragon ring buffer lost.
# TYPE tetragon_observer_ringbuf_events_lost_total counter
tetragon_observer_ringbuf_events_lost_total 0
# HELP tetragon_observer_ringbuf_events_received_total Number of perf events Tetragon ring buffer received.
# TYPE tetragon_observer_ringbuf_events_received_total counter
tetragon_observer_ringbuf_events_received_total 587417
# HELP tetragon_observer_ringbuf_queue_events_lost_total Number of perf events Tetragon ring buffer events queue lost.
# TYPE tetragon_observer_ringbuf_queue_events_lost_total counter
tetragon_observer_ringbuf_queue_events_lost_total 489162
# HELP tetragon_observer_ringbuf_queue_events_received_total Number of perf events Tetragon ring buffer events queue received.
# TYPE tetragon_observer_ringbuf_queue_events_received_total counter
tetragon_observer_ringbuf_queue_events_received_total 32886
```

this PR with bpf ring buffer, after stress:
```
root@e338dac4c545:~# curl localhost:2112/metrics 2>/dev/null | grep -i missed_events\\\|overflowed_events\\\|ringbuf
# HELP tetragon_bpf_missed_events_total Number of Tetragon perf events that are failed to be sent from the kernel.
# TYPE tetragon_bpf_missed_events_total counter
tetragon_bpf_missed_events_total{error="unknown",msg_op="14"} 61017 
tetragon_bpf_missed_events_total{error="unknown",msg_op="23"} 23
tetragon_bpf_missed_events_total{error="unknown",msg_op="5"} 27
tetragon_bpf_missed_events_total{error="unknown",msg_op="7"} 24
# HELP tetragon_notify_overflowed_events_total The total number of events dropped because listener buffer was full
# TYPE tetragon_notify_overflowed_events_total counter
tetragon_notify_overflowed_events_total 274750
# HELP tetragon_observer_ringbuf_errors_total Number of errors when reading Tetragon ring buffer.
# TYPE tetragon_observer_ringbuf_errors_total counter
tetragon_observer_ringbuf_errors_total 0
# HELP tetragon_observer_ringbuf_events_lost_total Number of perf events Tetragon ring buffer lost.
# TYPE tetragon_observer_ringbuf_events_lost_total counter
tetragon_observer_ringbuf_events_lost_total 0
# HELP tetragon_observer_ringbuf_events_received_total Number of perf events Tetragon ring buffer received.
# TYPE tetragon_observer_ringbuf_events_received_total counter
tetragon_observer_ringbuf_events_received_total 6.634135e+06
# HELP tetragon_observer_ringbuf_queue_events_lost_total Number of perf events Tetragon ring buffer events queue lost.
# TYPE tetragon_observer_ringbuf_queue_events_lost_total counter
tetragon_observer_ringbuf_queue_events_lost_total 6.212769e+06
# HELP tetragon_observer_ringbuf_queue_events_received_total Number of perf events Tetragon ring buffer events queue received.
# TYPE tetragon_observer_ringbuf_queue_events_received_total counter
tetragon_observer_ringbuf_queue_events_received_total 355830
```

The ringbuf_events_lost_total is 0 for the bpf ring buffer because the bpf ring buffer doesn't lose events (it prevents new events being inserted when full); instead we need to check the bpf_missed_events_total values as these count errors on the BPF side when submitting events.

Overall, it appears that the bpf ring loses fewer events in the ring buffer (60k vs 100k-130k), receives more events from the ring buffer (6.6e6 vs 5.3e6-5.5e6), and manages to deliver more events to the event queue (350k vs 310k-340k).

pkg/observer/observer.go
FedeDP: Couldn't find anything :/ anway, this is not a big deal and something we can fix later if issues arise. 
Thank you very much for looking into that!

dwindsor:
> I didn't, so I ran some tests just now to compare. In all cases I ran the same raw_syscalls tracing policy. I took a couple of measurements a few seconds after starting up, a couple of seconds apart (not very scientific, admittedly). I then ran the stress test and took a measurement after.

You didn't have to do that! Thanks though, I think these numbers justify the feature

bpf/lib/bpf_event.h
```diff
@@ -16,4 +17,31 @@ struct {
...
+
+FUNC_INLINE long event_output(void *ctx, void *data, uint64_t size)
+{
+	struct tetragon_conf *conf;
+	int zero = 0;
+
+	conf = map_lookup_elem(&tg_conf_map, &zero);
+	if (conf && conf->use_perf_ring_buf)
```
FedeDP: Question: when do we expect the user to prefer `perf ring_buf` in place of `bpf_ringbuf` when the latter is supported? I mean, do we really need an option to force-enable the perf ring buf?

FedeDP: Perhaps it is just for a smooth transition?

pkg/bpf/ringbuf_linux.go
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package bpf
+
+import "path/filepath"
+
+var (
+	ringbufEventsMapName = "tg_rb_events"
+	RingBufEventsMapName = filepath.Join(MapPrefixPath(), ringbufEventsMapName)
```
olsajiri: nit, this gets removed/changed in the next commit, looks like it belongs to the next commit

bpf/lib/bpf_event.h
```diff
@@ -16,4 +17,31 @@ struct {
...
+		return perf_event_output(ctx, &tcpmon_map, BPF_F_CURRENT_CPU, data, size);
+	return ringbuf_output(&tg_rb_events, data, size, 0);
```
olsajiri: this can return -EAGAIN and we should handle that in perf_event_output_update_error_metric

kevsecurity:
> With the `reserve` API, we can directly read data onto the ring buffer. Of course, you can only `preserve` space for data whose size is known at compile time (eg: if you are pushing an integer parameter).

Yep, but many of the events are variable sizes. I'll update any fixed size ones in a follow-up PR.

> EDIT: oh i just saw we are indeed using a single ring buffer shared among all CPUs, right? Didn't catch it, i was expecting a ring buffer for each CPU :D that's why i had concerns about memory usage.

Yes, we're using the default model. If it causes problems, we can look at more advanced architectures. :D

FedeDP:
Makes 100% sense, thank you very much!

bpf/lib/bpf_event.h
kevsecurity: The reason is really in case a problem arises with the BPF ring buffer. Not expected, but this would permit quick resolution while we find a code solution.

pkg/observer/observer.go
```diff
@@ -149,8 +150,9 @@ func (k *Observer) getRBQueueSize() int {
 // notified of their corresponding events.
 type Observer struct {
 	/* Configuration */
-	listeners  map[Listener]struct{}
```
olsajiri, line 152: could you please remove the 'Perf ring buffer..' prefix from log line few lines above, it's confusing now

olsajiri:
I did quick benchmark by monitoring some of the syscalls by tetragon during and running kernel kernel build on on 16 cpus

I got slightly more events and better cycles spent results using ringbuf, but its overall time is about 4% longer.. not sure why yet, perhaps the contention on the ringbuf, but then it's supposed to return EAGAIN in that case.. need to check more

```
ringbuf:

        11,060,653,512,288      cycles:u
         2,042,302,383,017      cycles:k
        13,105,302,080,370      cycles
        14,127,809,786,880      instructions                     #    1.08  insn per cycle

             269.098160020 seconds time elapsed

        tetragon_observer_ringbuf_events_received_total: 23812513

        tetragon_bpf_missed_events_total{error="unknown",msg_op="13"} 10392
        tetragon_bpf_missed_events_total{error="unknown",msg_op="23"} 1
        tetragon_bpf_missed_events_total{error="unknown",msg_op="5"} 2
        tetragon_bpf_missed_events_total{error="unknown",msg_op="7"} 7

perf:

        11,293,047,239,564      cycles:u
         2,173,144,746,229      cycles:k
        13,468,566,933,095      cycles
        14,212,137,152,521      instructions                     #    1.06  insn per cycle

             258.944280935 seconds time elapsed

        tetragon_observer_ringbuf_events_received_total: 23775446
        tetragon_observer_ringbuf_events_lost_total 31224

        tetragon_bpf_missed_events_total{error="ENOENT",msg_op="13"} 30
        tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="13"} 31198
        tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="23"} 7
        tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="24"} 4
        tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="5"} 8
        tetragon_bpf_missed_events_total{error="ENOSPC",msg_op="7"} 7

without tetragon:

        10,066,987,517,786      cycles:u
         1,740,507,945,446      cycles:k
        11,809,381,832,745      cycles
        12,564,083,574,553      instructions                     #    1.06  insn per cycle

             228.918539640 seconds time elapsed
```

pkg/bpf/ringbuf_linux.go
kevsecurity: Yep, will fix.

bpf/lib/bpf_event.h
kevsecurity: Yep, will add.

pkg/observer/observer.go
kevsecurity: Yep.

kevsecurity:
> I did quick benchmark by monitoring some of the syscalls by tetragon during and running kernel kernel build on on 16 cpus
> 
> I got slightly more events and better cycles spent results using ringbuf, but its overall time is about 4% longer.. not sure why yet, perhaps the contention on the ringbuf, but then it's supposed to return EAGAIN in that case.. need to check more

Thanks. If you only ran the bench once per condition, then I'd recommend (if you don't mind) running it a handful of times per condition and taking the means. I found there was quite a lot of noise in my tests and averaging over 3 runs gave a cheap average that showed the extent of the noise (and that the two conditions were similar overall).

pkg/config/config_linux.go
```diff
@@ -147,6 +147,14 @@ func EnableRhel7Progs() bool {
 	return (int64(kernelVer) < kernels.KernelStringToNumeric("3.11.0"))
 }
 
+func EnableV511Progs() bool {
+	if option.Config.ForceSmallProgs {
+		return false
+	}
+	kernelVer, _, _ := kernels.GetKernelVersion(option.Config.KernelVersion, option.Config.ProcFS)
+	return (int64(kernelVer) >= kernels.KernelStringToNumeric("5.11.0"))
+}
```
kkourt, line 212: From what I can see, there are still checks for `kernels.MinKernelVersion("5.11")` in your branch.

For example in `ExecObj()`:

```shell
// ExecObj returns the exec object based on the kernel version
func ExecObj() string {
	if EnableRhel7Progs() {
		return "bpf_execve_event_v310.o"
	} else if EnableV612Progs() {
		return "bpf_execve_event_v612.o"
	} else if EnableV61Progs() {
		return "bpf_execve_event_v61.o"
	} else if kernels.MinKernelVersion("5.11") {
		return "bpf_execve_event_v511.o"
	} else if EnableLargeProgs() {
		return "bpf_execve_event_v53.o"
	}
	return "bpf_execve_event.o"
}
```


I think having both checks can be confusing. I would suggest having a preparation patch where `EnableV511Progs()` is added and used for all the isntances where we use _v511.o variants.

bpf/lib/environ_conf.h
```diff
@@ -25,6 +25,8 @@ struct tetragon_conf {
 	__u32 tg_cgrp_level; /* Tetragon cgroup level */
 	__u64 tg_cgrpid; /* Tetragon current cgroup ID to avoid filtering blocking itself */
 	__u64 cgrp_fs_magic; /* Cgroupv1 or Cgroupv2 */
+	__u8 use_perf_ring_buf; /* Use perf ring buffer rather than the bpf ring buffer */
```
mtardy, line 28: idea: you could use global const in place and then profit from dead code elimination (from 5.4).

cmd/tetragon/main.go
```diff
@@ -808,6 +809,13 @@ func execute() error {
 			if err := option.ReadAndSetFlags(); err != nil {
 				logger.Fatal(log, "Failed to parse command line flags", logfields.Error, err)
 			}
+			// Override perf ring buffer choice if only the perf ring is available.
+			// NB: can't do this in option.ReadAndSetFlags() as it causes an import cycle.
```
mtardy, line 818: mmh this is annoying (indeed not a huge deal), we should be able to call the `kernels` pkg from this if we have a function to validate flags and stop tetragon starting. It can be useful in the future, maybe that's worth investigating and fixing?

pkg/observer/observer_linux.go
```diff
@@ -63,33 +63,32 @@ func perfBufferSize(perCPUBuffer int) int {
...
-		return fmt.Errorf("opening pinned map '%s' failed: %w", k.PerfConfig.MapName, err)
-	}
-	defer perfMap.Close()
 
-	rbSize := k.getRBSize(int(perfMap.MaxEntries()))
-	perfReader, err := perf.NewReader(perfMap, rbSize)
+	var ringBufMapName string
+	var ringBufReader *ringbuf.Reader
+	var perfReader *perf.Reader
+	useBpfRingBuffer := config.EnableV511Progs() && !option.Config.UsePerfRingBuffer
```
mtardy: naming nit: try to use BPF in place of Bpf

mtardy (commented):
Oups I realized I never submitted the few nits I wrote a while ago, I guess I didn't finish it properly. Hope it's not too much noise as it's a lot of nits you can ignore and you already have plenty of feedback on the PR.

On the message from Kornilios:
> Now, we don't have to fix all the functions in this PR, but if we think the above approach makes sense (CC: @olsajiri @mtardy ), we can try and have all new features being detected by runtime checks rather than kernel versions.
>
> Hence, for this PR specifically, I would argue for adding a runtime check for the bpf ringbuffer check in EnableV511Progs()

Agree!

kevsecurity:
> This is a good opportunity to discuss the kernel variants we have for the bpf code.
> 
> ...
>
> Hence, for this PR specifically, I would argue for adding a runtime check for the bpf ringbuffer check in `EnableV511Progs()`
> 
> Thoughts?

I agree with the idea; I don't agree that it should be in this PR.

My rationale is that a) the other version based checks all need replacing too; and b) this PR is about adding support for BPF ring buffer; why complicate it further with kernel feature detection?

pkg/config/config_linux.go
kevsecurity: Yes these are errors.

pkg/config/config_linux.go
kevsecurity: I've added a comment to the commit.

bpf/lib/environ_conf.h
kevsecurity: Yep, one at a time. This map already existed.

kkourt:
> > This is a good opportunity to discuss the kernel variants we have for the bpf code.
> > ...
> > Hence, for this PR specifically, I would argue for adding a runtime check for the bpf ringbuffer check in `EnableV511Progs()`
> > Thoughts?
> 
> I agree with the idea; I don't agree that it should be in this PR.
> 
> My rationale is that a) the other version based checks all need replacing too; and b) this PR is about adding support for BPF ring buffer; why complicate it further with kernel feature detection?

In my opinion, if the PR that introduces use of a feature (in this case, the bpf ring buffer) does not also introduce its runtime detection, runtime detection for this particular feature is unlikely to be added in the future.

kevsecurity:
> In my opinion, if the PR that introduces use of a feature (in this case, the bpf ring buffer) does not also introduce its runtime detection, runtime detection for this particular feature is unlikely to be added in the future.

Do we not have lots of counterexamples? For example, we used to do "equals" matches on strings by checking each byte one by one, limiting the length of strings and the number of strings to compare to; but I built a PR that changed that to a hash. And then added prefix using an LPM map. And then added postfix by inverting the string for an LPM map. And elsewhere I've put in quite a bit of effort to build feature detection reliably.

My opinion was that this feature detection might not be trivial, so why add it to an already complicated PR? It is also my opinion that, due to distributions like RHEL, feature detection will become more important and possibly essential. I therefore can't conceive of a situation where this doesn't get implemented. Also, we could create an issue to track it.

kevsecurity:
> In my opinion, if the PR that introduces use of a feature (in this case, the bpf ring buffer) does not also introduce its runtime detection, runtime detection for this particular feature is unlikely to be added in the future.

I mean, it does introduce runtime detection, just based on kernel version, like all the others.

kevsecurity:
> I got slightly more events and better cycles spent results using ringbuf, but its overall time is about 4% longer.. not sure why yet, perhaps the contention on the ringbuf, but then it's supposed to return EAGAIN in that case.. need to check more

I've been seeing that some tests were taking longer for a particular event to be seen. I think I have observed that the greater throughput of the bpf ring buffer is adding to the load on the event processor. When there is a huge volume of events, more arrive in user space with the bpf ring buffer (it's more efficient, so more are handled). This puts a bigger load on the event queue and event processor. Handling all those extra events takes CPU, so everything takes longer.

PR #4079, opened 2025-09-03
new(cmd/tetra,pkg/bugtool): allow to extend bugtool with custom commands and grpc calls.

FedeDP:
### Description

It can be useful for downstream consumers to extend content of `bugtool` generated tar archive with custom commands or grpc calls.

### Changelog

```release-note
new(cmd/tetra,pkg/bugtool): allow to extend bugtool with custom commands and grpc calls.
```


cmd/tetra/commands_linux.go
```diff
@@ -18,7 +18,7 @@ import (
 
 func addCommands(rootCmd *cobra.Command) {
 	addBaseCommands(rootCmd)
-	rootCmd.AddCommand(bugtool.New())
+	rootCmd.AddCommand(bugtool.New().Command())
```
FedeDP, line 21: Ideally, it would be used like this:
```
rootCmd.AddCommand(bugtool.New().
		WithCommandAction(ifConfig).
		WithGRPCAction(grpcAdditionalAction).
		Command())

...

func ifConfig(commander bugtoolpkg.Commander, tarWriter *tar.Writer) error {
	return commander.ExecCmd(tarWriter, "ifconfig.out", "ifconfig", "-a")
}

func grpcAdditionalAction(grpcer bugtoolpkg.GRPCer, tarWriter *tar.Writer) error {
	c, err := common.NewClient(context.Background(), "myaddr", 5*time.Second)
	if err != nil {
		return err
	}
	defer c.Close()
    
	res, err := c.Client.ListSensors(c.Ctx, &tetragon.ListSensorsRequest{})
	if err != nil || res == nil {
		return errors.New("failed to list sensors")
	}

	fname := "sensors.json"
	return grpcer.TarAddJson(tarWriter, fname, res)
}
```

pkg/bugtool/bugtool.go
```diff
@@ -126,6 +126,7 @@ type bugtoolInfo struct {
 	info      *InitInfo
 	prefixDir string
 	multiLog  MultiLog
+	tarWriter *tar.Writer
```
FedeDP, line 129: Stored `tarWriter` inside `bugtoolInfo` struct to avoid passing it everywhere; this way the new 2 interfaces `Commander` and `GRPCer` are leaner too.

cmd/tetra/commands_linux.go
mtardy: Hey! I noticed you like putting notes on your GH PRs, would that be possible to include some of those in the commit description so that it persists in history if that's useful notes on your patches?

I see some of that are intended toward reviewer while some other might be useful to find later on

pkg/bugtool/bugtool.go
```diff
@@ -195,8 +196,20 @@ func (s *bugtoolInfo) tarAddFile(tarWriter *tar.Writer, fnameSrc string, fnameDs
...
 
-// Bugtool gathers information and writes it as a tar archive in the given filename
-func Bugtool(outFname string, bpftool string, gops string) error {
+type Commander interface {
+	ExecCmd(dstFname string, cmdName string, cmdArgs ...string) error
+}
+
+type GRPCer interface {
+	TarAddJson(fname string, obj interface{}) error
+}
```
FedeDP, line 205: bugtoolInfo struct implements both interfaces thus can be passed (as pointer) as either interface when calling additional command/grpc Actions.

FedeDP: I don't really like `GRPCer` naming since all the interface allows to do is to append a json to the tar archive. At the same time, i wanted to strongly tie this one with the `GRPCAction` , at least naming-wise.

pkg/bugtool/bugtool.go
FedeDP: Most of the changes in this file are caused by this change, or by the `ExecCmd` and `TarAddJson` renaming.

cmd/tetra/commands_linux.go
FedeDP: I think that's a great idea! Yes i LOVE to make self-reviews, it is helpful for reviewers and for myself (to find bugs or better solutions :D)

Let me amend the commit with some more info.

kkourt: yap, wanted to say the same thing as Mahé. All of this is great context so it's great if it becomes part of the commit message.

kkourt:
@FedeDP just a note: if the PR is ready for review, you can notify the reviewers by clicking on the github button "Re-request review". (At least for me, this ensures that the review is back in my queue)

FedeDP:
Thanks @kkourt ! You are totally right :)

pkg/bugtool/bugtool.go
mtardy: > Stored tarWriter inside bugtoolInfo struct to avoid passing it everywhere;

I feel like it would have been nice to put that in a separate commit

Maybe more generally, if you need to write review comments on your PR that could mean that a separate patch could be useful with that commit description on why you do it.

FedeDP: You are right, my bad!
I can always rebase and extrapolate the commit if you wish, it's not a problem!

FedeDP:
> I was just genuinely wondering what's the benefit of creating this specialized extension mechanism while you could extend the cobra command?

I think it feels more straighforward having an API for that; like an API covered case and not just a forced workaround, IMHO.

FedeDP:
@mtardy split the PR in 2 commits, as advised! Thank you very much!

PR #4081, opened 2025-09-05
install/kubernetes/tetragon: Helm chart add serviceLabels

yieon-lyon:
charts: add serviceLabels for additive Service label management

Tetragon Helm chart currently exposes only serviceLabelsOverride which forces full replacement of the Service labels and makes day to day label management error prone.

This patch fixes this by adding:

A new values key serviceLabels as an empty map in values.yaml to allow appending labels without losing chart defaults. Example:

```
serviceLabels: {}
# serviceLabels:
#   team: sre
#   owner: platform
```

Service template logic that preserves compatibility. If serviceLabels is set it is merged with existing chart labels or serviceLabelsOverride to produce the final label set. additional serviceLabels can simplify label based integration such as Prometheus.

Signed-off-by: Lyon <yieon1230@gmail.com>

### DIFF
--- a/install/kubernetes/tetragon/templates/service.yaml
+++ b/install/kubernetes/tetragon/templates/service.yaml
@@ -23,5 +23,8 @@ spec:
     {{- else }}
     {{- include "tetragon.labels" . | nindent 4 }}
     {{- end }}
+    {{- with .Values.serviceLabels }}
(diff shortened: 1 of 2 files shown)


yieon-lyon:
Delete PR due to incorrect commit and recreate it

PR #4095, opened 2025-09-08
tetragon: allow to parse usdt sib argument

olsajiri:
Borrowing changelog from kernel commit [1]:

    On x86-64, USDT arguments can be specified using Scale-Index-Base (SIB)
    addressing, e.g. "1@-96(%rbp,%rax,8)". The current USDT implementation
    in libbpf cannot parse this format, causing `bpf_program__attach_usdt()`
    to fail with -ENOENT (unrecognized register).

    This patch fixes this by implementing the necessary changes:
    - add correct handling for SIB-addressed arguments in `bpf_usdt_arg`.
    - add adaptive support to `__bpf_usdt_arg_type` and 
      `__bpf_usdt_arg_spec` to represent SIB addressing parameters.

Adding same change to our usdt parsing code.

[1] 758acb9ccfdb libbpf: Fix USDT SIB argument handling causing unrecognized register error

### DIFF
--- a/contrib/tester-progs/sdt.h
+++ b/contrib/tester-progs/sdt.h
@@ -1,0 +1,515 @@
+/* <sys/sdt.h> - Systemtap static probe definition macros.
+
+   This file is dedicated to the public domain, pursuant to CC0
+   (https://creativecommons.org/publicdomain/zero/1.0/)
+*/
+
+#ifndef _SYS_SDT_H
+#define _SYS_SDT_H    1
+
+/*
+  This file defines a family of macros
+
+       STAP_PROBEn(op1, ..., opn)
+
+  that emit a nop into the instruction stream, and some data into an auxiliary
+  note section.  The data in the note section describes the operands, in terms
+  of size and location.  Each location is encoded as assembler operand string.
+  Consumer tools such as gdb or systemtap insert breakpoints on top of
+  the nop, and decode the location operand-strings, like an assembler,
+  to find the values being passed.
+
+  The operand strings are selected by the compiler for each operand.
+  They are constrained by gcc inline-assembler codes.  The default is:
+
+  #define STAP_SDT_ARG_CONSTRAINT nor
+
+  This is a good default if the operands tend to be integral and
+  moderate in number (smaller than number of registers).  In other
+  cases, the compiler may report "'asm' requires impossible reload" or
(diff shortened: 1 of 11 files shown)


pkg/elf/usdt_args_amd64.go
```diff
@@ -97,6 +102,57 @@ func parseCommon(sz int, arg *UsdtArg) error {
...
+				if n, _ = fmt.Sscanf(str, "%d@(%%%s,%%%s)", &sz, reg.Reset(), regIdx.Reset()); n != 3 {
+					return errNext
+				}
+			}
+		}
+	}
+
+	arg.Type = USDT_ARG_TYPE_SIB
+	arg.ValOff = uint64(off)
+	arg.RegOff, ok = resolveReg(reg.name)
```
kkourt, line 132: At the bpf side, I believe we mask these with  `0xfff`. Can we reject values that are larger than 4096 here?

contrib/tester-progs/sdt.h
```diff
@@ -0,0 +1,515 @@
+/* <sys/sdt.h> - Systemtap static probe definition macros.
+
+   This file is dedicated to the public domain, pursuant to CC0
+   (https://creativecommons.org/publicdomain/zero/1.0/)
+*/
```
kkourt, line 5: > We already have usdt.h macro from [1] but for the SIB testing it's
> very useful to have the sdt.h from [1] which allows to define
> argument string directly.

nit: the [1] reference seems to be missing (in the commit message).

pkg/elf/usdt_args_amd64.go
olsajiri: the values are restricted to the size of the `struct pt_regs` , so no problem there

the bpf code you probably refer to is this one:
`
off = arg->reg_off & 0xfff
`
if it's not present the older kernels verifier will fail with:
`
math between ctx pointer and register with unbounded min value is not allowed
`
I can't find a way to make unsigned log value out of context pointer so that's why I mask the offset instead,while it's not really needed

contrib/tester-progs/sdt.h
olsajiri: ugh, added, thanks

kkourt (approved):

PR #4098, opened 2025-09-10
new(tests/e2e): add a metrics checker on e2e tests.

FedeDP:
### Description

Implement a metricschecker for e2e tests.


mtardy:
Hey @will-isovalent could you take a look at that since you are the issue author :) ?

tests/e2e/metricschecker/metricschecker.go
```diff
@@ -0,0 +1,218 @@
...
+	rc.timeLimit = limit
+	return rc
+}
+
+type metricOp int
+
+const (
+	opEqual   metricOp = 1
+	opGreater          = 1 << iota
+	opLess
```
kkourt: I guess we are using bits to capture things like `>=` as `opEqual | opLess`? Maybe add a comment here since it's  unconventional? (at least to me)

kkourt (approved):
LGTM, thanks.

tests/e2e/metricschecker/metricschecker.go
FedeDP: Yep i'll do! :)

FedeDP: Renamed so that `bitmask` usage is more obvious.

PR #4107, opened 2025-09-12
policy_stats: use the map only for policy sensors

kkourt:
perf_event_output_metric is defined in lib/process.h that is included from the code in the base sensor as well.

As a result, some programs from the exec sensor end up having private copies of the policy_stats map, which is just waster memory. Add some ifdefs to only enable the policy_stats map for the tracing (generic) sensors.

Fixes: b0e50ee7f3a1 ("bpf: policy stats")

### DIFF
--- a/bpf/lib/policy_stats.h
+++ b/bpf/lib/policy_stats.h
@@ -6,6 +6,10 @@
 
 #include "policy_conf.h"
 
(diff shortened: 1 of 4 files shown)


kkourt:
Generated docs failure is fixed by https://github.com/cilium/tetragon/pull/4108

PR #4109, opened 2025-09-12
tetragon: add range filter

olsajiri:
adding InRange and NotInrage operators that allows to filter for 4 ranges in match selectors, like: 
```
      matchData:
      - index: 0
        operator: NotInRange
        values:
        - "0:187904818"
        - "187904818:1879048187"
        - "1879048189:4294967295"
```

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -6997,3 +6997,104 @@ spec:
 	err = jsonchecker.JsonTestCheck(t, checker)
 	require.NoError(t, err)
 }
+
+func testKprobeRangeOp(t *testing.T, in bool) {
+
+	if !config.EnableLargeProgs() {
+		t.Skipf("Skipping test since it needs kernel >= 5.3")
+	}
+
+	var doneWG, readyWG sync.WaitGroup
+	defer doneWG.Wait()
+
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
+	op := "InRange"
+	if !in {
+		op = "NotInRange"
+	}
+
+	hook := `apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "resolve-parent-comm"
+spec:
+  kprobes:
+  - call: "sys_prctl"
+    syscall: true
+    args:
+    - index: 0
+      type: "int"
+    selectors:
+    - matchArgs:
+      - index: 0
+        operator: "` + op + `"
+        values:
+        - 0xffff0:0xffff1
+        - 0xffff5:0xffff6
+        - 0xffff8:0xffff9
+        - 0xffffe:0xfffff
+`
+
+	createCrdFile(t, hook)
+
+	obs, err := observertesthelper.GetDefaultObserverWithFile(t, ctx, testConfigFile, tus.Conf().TetragonLib)
+	if err != nil {
+		t.Fatalf("GetDefaultObserverWithFile error: %s", err)
+	}
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	for i := range 0x10 {
+		unix.Prctl(0xffff0+i, 0, 0, 0, 0)
+	}
+
+	getChecker := func(list []int) *ec.UnorderedEventChecker {
+		checkers := []ec.EventChecker{}
+
+		for _, v := range list {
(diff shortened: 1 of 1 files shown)


bpf/process/types/basic.h
```diff
@@ -1277,6 +1277,74 @@ filter_64ty_map(struct selector_arg_filter *filter, char *args)
...
+
+#pragma unroll
+	for (i = 0; i < MAX_MATCH_VALUES; i += 2) {
+		__u64 start = v[i], end = v[i + 1];
+		bool res;
+
+		if (is_signed_type(filter->type))
+			res = ((__s64)start <= *(s64 *)args) && (*(s64 *)args <= (__s64)end);
+		else
+			res = (start <= *(s64 *)args) && (*(s64 *)args <= end);
```
kkourt: Shouldn't this be (u64 *)args?

olsajiri: it should, will fix thanks

bpf/process/types/basic.h
```diff
@@ -1277,6 +1277,74 @@ filter_64ty_map(struct selector_arg_filter *filter, char *args)
...
 
+#ifdef __LARGE_BPF_PROG
+FUNC_LOCAL long
+filter_64ty_range(struct selector_arg_filter *filter, char *args)
+{
+	__u64 *v = (__u64 *)&filter->value;
+	int i, j = 0;
+
+#pragma unroll
+	for (i = 0; i < MAX_MATCH_VALUES; i += 2) {
```
mtardy: As I read it, `MAX_MATCH_VALUES` being `4` and since you increment by `i += 2`, to me you can support only up to 2 ranges and not 4 ranges. Maybe I'm missing something because I see you have three ranges in the last tests but wondering if it is actually triggered 🤔 ?

olsajiri: ah yea, and there's more issues.. I'll push new version, thanks

bpf/process/types/basic.h
```diff
@@ -1280,6 +1280,78 @@ filter_64ty_map(struct selector_arg_filter *filter, char *args)
...
+	__u64 *v = (__u64 *)&filter->value;
+	int all = 0, i, j = 0;
+
+#pragma unroll
+	for (i = 0; i < MAX_MATCH_VALUES; i++) {
+		__u64 start = v[i * 2], end = v[i * 2 + 1];
+		bool res;
+
+		if (is_signed_type(filter->type))
+			res = ((__s64)start <= *(s64 *)args) && (*(s64 *)args <= (__s64)end);
```
mtardy, line 1296: since your range is inclusive on both ends, how would that work with:
```
    values:
    - "0:187904818"
    - "187904818:1879048187"
 ```
technically both range would match for number `187904818`. I imagine it's not a big deal since you short-circuit as soon as you match. Maybe you want Go-like range with `slice[0:5]` being `0, 1, 2, 3, 4` elements. Anyway that would be great eventually to have it documented somewhere how they behave.

File: bpf/process/types/basic.h [hunk 4]
```diff
@@ -1280,6 +1280,78 @@ filter_64ty_map(struct selector_arg_filter *filter, char *args)
...
+		else
+			res = (start <= *(u64 *)args) && (*(u64 *)args <= end);
+
+		if (filter->op == op_in_range && res)
+			return 1;
+		all |= res;
+
+		j += 16;
+		if (j + 8 /* vallen+type */ >= filter->vallen)
```
mtardy: nit: maybe writing `j += sizeof(int64_t) * 2` would make this more explicit and self documented?

using `sizeof(whatever is 64 bits)` instead of 8 maybe is better

bpf/process/types/basic.h
```diff
@@ -1280,6 +1280,78 @@ filter_64ty_map(struct selector_arg_filter *filter, char *args)
...
+		if (is_signed_type(filter->type))
+			res = ((__s32)start <= *(s32 *)args) && (*(s32 *)args <= (__s32)end);
+		else
+			res = (start <= *(u32 *)args) && (*(u32 *)args <= end);
+
+		if (filter->op == op_in_range && res)
+			return 1;
+		all |= res;
+
+		j += 8;
```
mtardy: so that here it can be `sizeof(uint32_t) * 2`

mtardy (approved):
LGTM! Just a few remarks (sorry for the late review)

bpf/process/types/basic.h
olsajiri: the [min,max] range seems more user friendly for me.. I added docs

PR #4116, opened 2025-09-17
k8s: Add alias for getting k8s config

sayboras:
Similar to what controller-runtime is having (e.g. alias.go), this will help with testing and extension later. One of the use case is to support more auth methods to k8s.

### Description
Similar to what controller-runtime is having (e.g. alias.go), this will help with testing and extension later. One of the use case is to support more auth methods to k8s.

### Changelog

```release-note
k8s: Add alias for getting k8s config
```


mtardy (approved):
Yep sure!

PR #4120, opened 2025-09-18
new(bpf): errmetrics improvements

FedeDP:
### Description

This PR improves bpf errmetrics intoduced in #3205.
It does so with multiple things (split by commit):
* added a generic `with_errmetrics` macro that supersedes `map_update_elem__errmetrics`
* expose bpf helper informations too in errmetrics
* use `__FILE_NAME__` macro instead of just `__FILE__`; this prevents issues with relative includes: prior to this, some files (eg: `bpf_process_event.h`) were also included with relative path (`../bpf_process_event.h`), and we would've been required to add them twice to the `fileids` with both paths. Now, we will only report the filename. We don't expect any name clash between bpf sources!
* the trick used to trigger a comp time error in `with_errmetrics` macro worked fine until we had multiple comp time errors in the same source file. At that point, the compiler complained about multiple `extern` function definition with different `error` attributes. I fixed it by appending a `__COUNTER__` to the end of the comp time error function name
* `bpf/tetragon/fileids.h` should not be formatted; added it to the ignore list
* expanded usage of `with_errmetrics` macro for `map_update_elem`
* updated `cmd/bpf-fileids` go command to leverage `cgo` instead of manually parsing the C header. Only downside: we are no more able to customize `fileids.h` path. I don't think that is needed at all since this command is mostly made to be automatically ran by makefile.

### DIFF
--- a/cmd/bpf-fileids/fileids.go
+++ b/cmd/bpf-fileids/fileids.go
@@ -3,14 +3,38 @@
 
 package main
 
+/*
+typedef struct {
+	const char *filename;
+	int id;
+}  fileid_t;
+
+const fileid_t *get_fileids() {
+    static fileid_t fileids[] = {
+#define fileid__(f, id)                  { f, id },
+#include "../../bpf/tetragon/fileids.h"
+#undef fileid__
+		{ NULL, -1 }
+	};
+	return fileids;
+}
+
+size_t get_fileids_len() {
+    const fileid_t *fileids = get_fileids();
+	size_t len = 0;
+	while (fileids[len].filename != NULL) {
+		len++;
+	}
+	return len;
+}
+*/
+import "C"
+
 import (
-	"bufio"
 	"encoding/json"
 	"fmt"
 	"os"
-	"regexp"
-	"strconv"
-	"strings"
+	"unsafe"
 )
 
 // json entry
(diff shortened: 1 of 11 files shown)


FedeDP:
cc @kkourt  since he made the initial PR :)

FedeDP:
Just force-pushed rewording a commit :) 
> [Check pull request commits / checkpatch (pull_request)](https://github.com/cilium/tetragon/actions/runs/17826173892/job/50679759521?pr=4120)Failing after 14s

It is complaining about `MACRO_ARG_REUSE`. I think that is a false-positive.

cmd/bpf-fileids/fileids.go
```diff
@@ -19,52 +43,37 @@ type Entry struct {
...
-		m := entryRe.FindStringSubmatch(line)
-		if len(m) == 3 {
-			id, err := strconv.ParseInt(m[2], 0, 32)
-			if err != nil {
-				continue
-			}
-			ret = append(ret, Entry{int(id), m[1]})
-		}
-
+	goSlice := (*[1 << 30]C.fileid_t)(unsafe.Pointer(cArrayPtr))[:fileidsLen:fileidsLen]
```
mtardy, line 54: I was confused about that line, maybe drop a comment, especially for the `(*[1 << 30]C.fileid_t)` type casting part which seems typical CGO stuff.

mtardy (approved):
Overall this looks good to me, that's a nice improvement and I want this to be used more widely also :)!

For the CGO part, we tried to get rid of CGO in tetragon itself. I guess it's fine to use it as part of the `cmd/bpf-fileids`. Given the simple format of the header file, maybe the switch from regex to CGO is a bit overkill and arguably not super worth but since you did it already. I imagine it's more resilient now?

Aside: checkpatch is mostly complaining about too long commit titles
> Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 91)

Try to make them fit within ~70 chars if possible :)

FedeDP:
> Given the simple format of the header file, maybe the switch from regex to CGO is a bit overkill and arguably not super worth but since you did it already. I imagine it's more resilient now?

Yes it is overkill for sure! But well, i think that parsing a C header leveraging CGO is much better than doing it by hand (and is more robust, eg: we now support multiline comments too); also, we could support more complex data structures in the future, since we are now just including the C header through CGO.
Not sure we are gonna do that, but at least we can :)

FedeDP:
> Try to make them fit within ~70 chars if possible :)

Uh didn't notice this error! Will reword the commit as needed!

cmd/bpf-fileids/fileids.go
FedeDP: Done!

FedeDP:
Pushed the requested changes:)

mtardy:
At first I wanted to disable the checkpatch `MACRO_ARG_REUSE` as Cilium does https://github.com/cilium/cilium/blob/d0f466773e2df4c3dc3823e1014dc86d45e063f7/bpf/Makefile.bpf#L60 but maybe that's an actually nice check. It's okay to ignore checkpatch sometimes but at least it makes us think about stuff 🤔 💭 

> It is complaining about MACRO_ARG_REUSE. I think that is a false-positive.

Yeah it's a bit weird as a check... It's because you reuse the same arg of the macro twice and technically someone could misuse your macro by passing an expression like `macro(x++)` and then if x was equal to 0 before, it's now 2 instead of expected 1. We can safely ignore this but still worth noting maybe...

cmd/bpf-fileids/fileids.go
```diff
@@ -3,14 +3,38 @@
...
+
+size_t get_fileids_len() {
+    const fileid_t *fileids = get_fileids();
+	size_t len = 0;
+	while (fileids[len].filename != NULL) {
+		len++;
+	}
+	return len;
+}
+*/
```
kkourt, line 30: Nice! We don't use cgo in the agent for many reasons, but that's a great use. Thanks,

PR #4124, opened 2025-09-23
kprobe: Add support for bpf_prog argument

tpapagian:
Example usage:
```yaml
kind: TracingPolicy
metadata:
  name: "example-bpf"
spec:
  kprobes:
  - call: "security_bpf_prog"
    syscall: false
    args:
    - index: 0
      type: "bpf_prog"
```


pkg/generictypes/generictypes.go
```diff
@@ -171,6 +174,7 @@ var genericTypeToStringTable = map[int]string{
 	GenericSockaddrType:    "sockaddr",
 	GenericSocketType:      "socket",
 	GenericDentryType:      "dentry",
+	GenericBpfProg:         "bpf_prog",
```
mtardy: nit: Seems the convention is not respected everywhere but I think it's better with the type suffix
```suggestion
	GenericBpfProgType:         "bpf_prog",
```

bpf/libbpf/bpf_core_read.h
```diff
@@ -163,7 +163,7 @@ enum bpf_type_info_kind {
  * argument.
  */
 #define bpf_core_read_str(dst, sz, src)					    \
-	bpf_probe_read_str(dst, sz,					    \
+	probe_read_str(dst, sz,					    \
```
mtardy: what happened here? we just never noticed?

tpapagian: Yes, it seems that we have never used `BPF_CORE_READ_STR_INTO` macro before.

pkg/generictypes/generictypes.go
tpapagian: I used `GenericBpfAttr` as an example to create that PR but almost all of the rest have a `Type` suffix. I have updated that PR.

Issue #4119, opened 2025-09-18
perf_event_arg config field value meaning


genie-zs:
Hello! Recently I did some experiments with [bpf policy](https://github.com/cilium/tetragon/blob/main/examples/policylibrary/bpf.yaml) and came accros strange set of numbers in [perf_event_arg](https://man7.org/linux/man-pages/man2/perf_event_open.2.html) config field for a tracepoint event.

*Output below is intentionally ommited*
```json
"function_name": "security_perf_event_alloc",
"args": [
  {
	"perf_event_arg": {
	  "Type": "PERF_TYPE_TRACEPOINT",
	  "Config": "670014898176" <---
	}
  }
]
```	
It seems ([1](https://man7.org/linux/man-pages/man2/perf_event_open.2.html), [2](https://github.com/cilium/tetragon/blob/698a6ba914fec27980f2c119296f4dd210657302/bpf/process/types/basic.h#L1172)) that it should contain tracepoint id matching `/sys/kernel/tracing/events/*/*/id`.

Is this somekind of bug or I just misunderstand something?

I've encountered this behaviour on next systems:

Tetragon Version: 1.5.0
Kernel: 5.14.0-362.13.1.el9_3.x86_64 
Linux flavor: Rocky 9.3

Tetragon Version: 1.4.1
Kernel: 5.8.0-63-generic
Linux flavor: Ubuntu 20.10

olsajiri:
yep, it's a bug, thanks for the report, should be fixed by https://github.com/cilium/tetragon/pull/4126

genie-zs:
Nice! Thank you!

PR #4126, opened 2025-09-25
tetragon: Fix struct perf_event_info_type layout

olsajiri:
We have a hole in perf_event_info_type which results in wrong number in its go counterpart MsgGenericKprobePerfEvent which is aligned differently.

Make sure the C object does not have any holes.

Fixes: https://github.com/cilium/tetragon/issues/4119


kevsecurity:
How did we not notice before?

olsajiri:
nobody checked that field I guess.. ?

PR #4132, opened 2025-09-26
chore(bpf, pkg/errmetrics): some probe_read() bpf errmetrics

FedeDP:
### Description

Also, fixes a small issue with the newly introduced cgo variant for the bpf-fileids go helper (see https://github.com/cilium/tetragon/commit/303ce626af6a7caf896fdd7b36f93098e68f05ee): since we are now embedding the fileids.h header through cgo, we need a non-cached run to be able to catch fileids.h updates.


kkourt (approved):
Thanks!

PR #4133, opened 2025-09-26
ci: always upload Go test artifacts for easier debugging

### LINKED ISSUE #2806: ci: Go tests does not return artifacts on failed execution
Here is an example https://github.com/cilium/tetragon/actions/runs/10416242015?pr=2804.

image

AritraDey-Dev:
Fixes #2806 

### Description
This PR updates the Go test workflow to always upload test artifacts (logs, bugtool dumps, vmlinux) regardless of test success or failure.

### DIFF
--- a/.github/workflows/gotests.yml
+++ b/.github/workflows/gotests.yml
@@ -77,7 +77,7 @@ jobs:
       if: failure()
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
       with:
-        name: tetragon-json
+        name: tetragon-json-${{ matrix.os }}
         path: /tmp/tetragon.gotest*
         retention-days: 5
 
@@ -85,7 +85,7 @@ jobs:
       if: failure()
       uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
       with:
-        name: tetragon-bugtool
+        name: tetragon-bugtool-${{ matrix.os }}
         path: /tmp/tetragon-bugtool*
         retention-days: 5
 


kevsecurity:
Is the benefit worth the storage costs?

mtardy (commented):
Thanks, I agree with the unique name fix but maybe not with the `always` change. Wdyt?

AritraDey-Dev:
> Thanks, I agree with the unique name fix but maybe not with the `always` change. Wdyt?

As the corresponding issue is mentioned, I'd prefer to keep `if: always()`  where we might need to compare logs from both successful and failed runs.  And also helps us spot warnings in passing tests.

AritraDey-Dev:
> Is the benefit worth the storage costs?

I believe, since the artifacts are retained for 5 days, this keeps long-term storage costs low.

mtardy (changes requested):
In the case of success you won't have any bugtool related information https://github.com/cilium/tetragon/actions/runs/18046445028?pr=4133. So maybe put it abck to failure() only. Not convinced about the usefulness of having the BTF and events in case of success as well. I'll drop e6ab3fd7be1d933349454af803ebbf6a9626bc31 all together.

Image

Thank you!

mtardy (commented):
hey could we just drop the rest and merge ad69544d437bfe6eb5ce72bff27c32013a7899a7?

AritraDey-Dev:
> hey could we just drop the rest and merge [ad69544](https://github.com/cilium/tetragon/commit/ad69544d437bfe6eb5ce72bff27c32013a7899a7)?

yes,I was overthinking it -` if: failure()` should work fine without continue-on-error: true and the final exit 1 step. I'll simplify this back to using if: failure()

PR #4127, opened 2025-09-25
pkg/errmetrics: expose error metrics via Prometheus metrics

mtardy:
This exposes the error metrics via Prometheus metrics and also add a new `errmetrics(ERR)` macro.

PR #4135, opened 2025-09-28
k8s: Add retry support for ControllerManager

shpalani:
### Changelog
```Support ControllerManager retry logic.
```

### Description
Problem:
  ControllerManager lacks support for retry. The retry option is required for projects where the on-prem controller can be delayed to come up.
Solution:
- Added retry support for ControllerManager. 
- The retry logic uses expontential backoff to avoid hammering the API Server.
- The retry logic uses Jitter, and logs retry attempt with time duration.
- Default for RetryCount is 1, where 1 means no retries,  just one execution of the connection attempt.
- Default for backoff is retry.DefaultBackoff

- This PR ensures the backward compatability for client's retry of ControllerManager is 1.
With this implementation, the Client's can set their desired values using the k8sConfigRetry alias to override.

Testing
1. Default behavior
```
hyperadmin@simulator:~/tetragon$ sudo ./tetragon --bpf-lib bpf/objs --k8s-kubeconfig-path /home/hyperadmin/tetragon/config --enable-application-model
level=info msg="Starting tetragon" version=71ec615fc
level=info msg="config settings" config="map[bpf-dir:tetragon bpf-lib:bpf/objs btf: cgroup-rate: cluster-name: config-dir: cpuprofile: cri-endpoint: data-cache-size:1024 debug:false disable-kprobe-multi:false enable-ancestors:[] enable-cgidmap:false enable-cgidmap-debug:false enable-cgtrackerid:true enable-compatibility-syscall64-size-type:false enable-cri:false enable-export-aggregation:false enable-k8s-api:false enable-msg-handling-latency:false enable-pid-set-filter:false enable-pod-annotations:false enable-pod-info:false enable-policy-filter:false enable-policy-filter-cgroup-map:false enable-policy-filter-debug:false enable-process-cred:false enable-process-ns:false enable-tracing-policy-crd:true event-cache-retries:15 event-cache-retry-delay:2 event-queue-size:10000 execve-map-entries:0 execve-map-size: export-aggregation-buffer-size:10000 export-aggregation-window-size:15s export-allowlist: export-denylist: export-file-compress:false export-file-max-backups:5 export-file-max-size-mb:10 export-file-perm:600 export-file-rotation-interval:0s export-filename: export-rate-limit:-1 expose-stack-addresses:false field-filters: force-large-progs:false force-small-progs:false generate-docs:false gops-address: health-server-address::6789 health-server-interval:10 k8s-kubeconfig-path:/home/hyperadmin/tetragon/config keep-sensors-on-exit:false kernel: log-format:text log-level:info memprofile: metrics-label-filter:namespace,workload,pod,binary metrics-server: netns-dir:/var/run/docker/netns/ pprof-address: process-cache-gc-interval:30s process-cache-size:65536 procfs:/proc/ rb-queue-size:65535 rb-size:0 rb-size-total:0 redaction-filters: release-pinned-bpf:true server-address:localhost:54321 tracing-policy: tracing-policy-dir:/etc/tetragon/tetragon.tp.d username-metadata:disabled verbose:0]"
level=info msg="Tetragon current security context" SELinux=unconfined AppArmor=unconfined Smack="" Lockdown=none
level=info msg="Tetragon pid file creation succeeded" pid=344077 pidfile=/var/run/tetragon/tetragon.pid
level=info msg="BPF: successfully released pinned BPF programs and maps" bpf-dir=/sys/fs/bpf/tetragon
level=info msg="BTF discovery: default kernel btf file found" btf-file=/sys/kernel/btf/vmlinux
level=info msg="BPF detected features: override_return: true, buildid: true, kprobe_multi: true, uprobe_multi true, fmodret: true, fmodret_syscall: true, signal: true, large: true, link_pin: true, lsm: false, missed_stats_kprobe_multi: true, missed_stats_kprobe: true, batch_update: true, uprobe_refctroff: true, audit_loginuid: true"
level=info msg="Cgroup mode detection succeeded" cgroup.fs=/sys/fs/cgroup cgroup.mode="Unified mode (Cgroupv2)"
level=info msg="Cgroupv2 supported controllers detected successfully" cgroup.fs=/sys/fs/cgroup cgroup.path=/proc/1/root/sys/fs/cgroup cgroup.controllers="[cpuset cpu io memory hugetlb pids rdma misc]" cgroup.hierarchyID=0
level=info msg="Cgroupv2 supported controllers detected successfully" cgroup.fs=/sys/fs/cgroup cgroup.path=/sys/fs/cgroup/user.slice/user-1001.slice/session-1175.scope cgroup.controllers="[memory pids]" cgroup.hierarchyID=0
level=info msg="Cgroupv2 hierarchy validated successfully" cgroup.fs=/sys/fs/cgroup cgroup.path=/sys/fs/cgroup/user.slice/user-1001.slice/session-1175.scope
level=info msg="Deployment mode detection succeeded" cgroup.fs=/sys/fs/cgroup deployment.mode="systemd user session"
level=info msg="Updated TetragonConf map successfully" confmap-update=tg_conf_map deployment.mode="systemd user session" log.level=0 cgroup.fs.magic=Cgroupv2 cgroup.hierarchyID=0 NSPID=344077
level=info msg="Enabling Kubernetes API"
level=info msg="created controller manager" attempts=1 duration=248.54994ms
level=info msg="Configured redaction filters" redactionFilters=""
```

2. With --k8s-controlplane-retry 1
```
hyperadmin@simulator:~/tetragon$ sudo ./tetragon --bpf-lib bpf/objs --k8s-kubeconfig-path /home/hyperadmin/tetragon/config --k8s-controlplane-retry 1
level=info msg="Starting tetragon" version=71ec615fc
...
level=info msg="Enabling Kubernetes API"
level=info msg="created controller manager" attempts=1 duration=249.152782ms
level=info msg="Configured redaction filters" redactionFilters=""
...
```

3. With --k8s-controlplane-retry 2
```
hyperadmin@simulator:~/tetragon$ sudo ./tetragon --bpf-lib bpf/objs --k8s-kubeconfig-path /home/hyperadmin/tetragon/config --k8s-controlplane-retry 2
...
level=info msg="Enabling Kubernetes API"
level=info msg="created controller manager" attempts=1 duration=250.076563ms
level=info msg="Configured redaction filters" redactionFilters=""
...
```

4. --k8s-controlplane-retry 0
```
hyperadmin@simulator:~/tetragon$ sudo ./tetragon --bpf-lib bpf/objs --k8s-kubeconfig-path /home/hyperadmin/tetragon/config --k8s-controlplane-retry 0
...
level=info msg="Enabling Kubernetes API"
level=info msg="retryCount is zero, which is invalid. Defaulting to 1"
level=info msg="created controller manager" attempts=1 duration=249.37557ms
```

5 With infinite retries, set to -1,  --k8s-controlplane-retry -1
```
hyperadmin@simulator:~/tetragon$ sudo ./tetragon --bpf-lib bpf/objs --k8s-kubeconfig-path /home/hyperadmin/tetragon/config --k8s-controlplane-retry -1
level=info msg="Enabling Kubernetes API"
level=info msg="created controller manager" attempts=1 duration=247.925219ms
level=info msg="Configured redaction filters" redactionFilters=""
...
```

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -6,7 +6,9 @@ package manager
 import (
 	"context"
 	"fmt"
+	"math"
 	"sync"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
@@ -15,7 +17,9 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	cmCache "sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -85,12 +89,10 @@ func newControllerManager() (*ControllerManager, error) {
 		cfg = ctrl.GetConfigOrDie()
 		inCluster = true
 	}
-	controllerManager, err := ctrl.NewManager(cfg, controllerOptions)
+	// Try to initialize the controller manager with retries
+	manager, err := newControllerManagerWithRetry(cfg, controllerOptions)
 	if err != nil {
 		return nil, err
-	}
-	manager = &ControllerManager{
-		Manager: controllerManager,
 	}
 	if inCluster {
 		err = manager.addPodInformer()
@@ -215,4 +217,67 @@ func (cm *ControllerManager) FindPod(podID string) (*corev1.Pod, error) {
 
 func (cm *ControllerManager) FindMirrorPod(hash string) (*corev1.Pod, error) {
 	return watcher.FindMirrorPod(hash, cm.podInformer)
+}
+
+// newControllerManagerWithRetry attempts to create a new ControllerManager instance with retry logic.
+// It uses the provided Kubernetes REST config and controller options. The number of retry attempts is
+// determined by conf.K8sConfigRetry(): a negative value means infinite retries, zero is invalid, and
+// a positive value specifies the maximum number of retries. The function applies exponential backoff
+// between retries. If the ControllerManager is created successfully, it returns the instance; otherwise,
+// it returns an error after exhausting all retries or encountering a fatal error.
+//
+// Parameters:
+//   - cfg: Kubernetes REST configuration.
+//   - controllerOptions: Options for the controller manager.
+//
+// Returns:
+//   - *ControllerManager: The created controller manager instance, or nil on failure.
+//   - error: An error if the controller manager could not be created.
+func newControllerManagerWithRetry(cfg *rest.Config, controllerOptions ctrl.Options) (cm *ControllerManager, err error) {
+
+	retryCount := conf.K8sConfigRetry()
+	if retryCount < 0 {
+		// max int32 retries, until connection succeeds.
+		logger.GetLogger().Info("setting retryCount to max int32 retries")
+		retryCount = math.MaxInt32
+	} else if retryCount == 0 {
+		// 1 means no retries, just one connection attempt.
+		logger.GetLogger().Info("retryCount is zero, which is invalid. Defaulting to 1")
+		retryCount = 1
+	}
+
+	var (
+		attempts  = 0
+		startTime = time.Now()
+	)
+
+	backoff := retry.DefaultBackoff
+	// localBackoff is a copy of backoff for retries.
+	localBackoff := backoff
+	localBackoff.Steps = retryCount
+	err = retry.RetryOnConflict(localBackoff, func() error {
+		attempts++
+		mgr, mgrErr := ctrl.NewManager(cfg, controllerOptions)
+		if mgrErr != nil {
+			logger.GetLogger().Warn("failed to create controller manager", logfields.Error, mgrErr,
+				"attempt", attempts)
+			// retry upon error
+			return mgrErr
+		}
+		cm = &ControllerManager{
+			Manager: mgr,
+		}
+		// success
+		return nil
+	})
+
+	duration := time.Since(startTime)
+	if err != nil {
+		logger.GetLogger().Error("failed to create controller manager", logfields.Error, err,
+			"attempts", attempts, "duration", duration.String(), "retryCount", retryCount)
+		return nil, err
+	}
+	logger.GetLogger().Info("created controller manager", "attempts", attempts,
+		"duration", duration.String())
+	return cm, nil
 }
--- a/pkg/watcher/conf/conf.go
+++ b/pkg/watcher/conf/conf.go
@@ -13,6 +13,9 @@ import (
 // K8sConfig alias is mainly for testing and extension if required.
 var K8sConfig = k8sConfig
 
+// K8sConfigRetry alias is for clients to override the retry default, if required
+var K8sConfigRetry = k8sConfigRetry
+
 // K8sConfig returns Kubernetes client configuration. If running in-cluster, the
 // second return value is true, otherwise false.
 func k8sConfig() (*rest.Config, bool, error) {
@@ -22,4 +25,12 @@ func k8sConfig() (*rest.Config, bool, error) {
 	}
 	cfg, err := rest.InClusterConfig()
 	return cfg, true, err
+}
+
+// k8sConfigRetry returns the number of attempts configured for the Kubernetes control plane.
+// It retrieves this value from the global option.Config.K8sControlPlaneRetry setting.
+// Negative values indicate infinite retries, zero is invalid, and positive values specify the maximum number of attempts.
+// Default behavior is 1 attempt.
+func k8sConfigRetry() (retryAttempts int) {
+	return option.Config.K8sControlPlaneRetry
 }
--- a/docs/data/tetragon_flags.yaml
+++ b/docs/data/tetragon_flags.yaml
@@ -168,6 +168,10 @@ options:
       shorthand: h
       default_value: "false"
       usage: help for tetragon
+    - name: k8s-controlplane-retry
+      default_value: "1"
+      usage: |
+        Number of attempts for Kubernetes control plane connection (negative for infinite, zero is invalid, positive for max attempts)
     - name: k8s-kubeconfig-path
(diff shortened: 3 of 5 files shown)


pkg/manager/manager.go
```diff
@@ -216,3 +218,53 @@ func (cm *ControllerManager) FindPod(podID string) (*corev1.Pod, error) {
...
+		}
+		cm = &ControllerManager{
+			Manager: mgr,
+		}
+		success = true
+		return true, nil
+	})
+
+	duration := time.Since(startTime)
+	if !success {
```
kkourt: Why do we need `success`? Can't we just check `backofErr`?
Also, why do we need `lastError`? Can't we just return `backofErr`?

ref: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoffWithContext, and https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoff.

shpalani: Thanks @kkourt for your comment.
```
- lastError is returned by mgrErr (in line# 244 lastError = mgrErr, giving why the connection failed).
- backoffErr is returned by wait.ExponentialBackoffWithContext (in line# 234).

Since the above errors give slightly different msg, I introduced success variable, so 
the final errToReturn has the right error. This helps the client code to have metrics/logging captured.
```
---
Info:
function wait.ExponentialBackoffWithContext
- It returns true, in which case ExponentialBackoffWithContext returns nil.
- The context ctx is canceled or times out, in which case it returns a non-nil error.
- The backoff limit is reached (based on the Backoff struct), in which case it also returns a non-nil error.

shpalani:
> let's use https://github.com/cilium/tetragon/blob/main/vendor/k8s.io/client-go/util/retry/util.go instead of implementing our own retry logic

Fixed. Thanks.

pkg/manager/manager.go
shpalani: > Why do we need `success`? Can't we just check `backofErr`? Also, why do we need `lastError`? Can't we just return `backofErr`?
> 
> ref: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoffWithContext, and https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait#ExponentialBackoff.

Fixed. Thanks.

shpalani:
> > let's use https://github.com/cilium/tetragon/blob/main/vendor/k8s.io/client-go/util/retry/util.go instead of implementing our own retry logic
> 
> Fixed. Thanks.

pkg/elf/usdt_args.go
```diff
@@ -5,6 +5,6 @@
 
 package elf
 
-func parseArgs(spec *UsdtSpec) error {
+func parseArgs(_ *UsdtSpec) error {
```
kkourt, line 8: FWIW, unrelated changes are better placed in separate patches.
In any case, this was already fixed by another PR: https://github.com/cilium/tetragon/pull/4127/commits/c4bbdde7e6c1d449a457cf3664e31c099adade8b.

pkg/watcher/conf/conf.go
```diff
@@ -13,6 +13,9 @@ import (
 // K8sConfig alias is mainly for testing and extension if required.
 var K8sConfig = k8sConfig
 
+// K8sConfigRetry alias is for clients to override the retry default, if required
+var K8sConfigRetry = k8sConfigRetry
+
```
kkourt, line 18: nit: something like
```
var K8sConfigRetry = func()  (retryAttempts int) {
    return 1
}
```

is slightly more readable.

michi-covalent: also, there should be a corresponding command line flag too right? 💭 users should be able to set the retry config without code change

pkg/manager/manager.go
```diff
@@ -85,13 +90,12 @@ func newControllerManager() (*ControllerManager, error) {
...
-	controllerManager, err := ctrl.NewManager(cfg, controllerOptions)
+	// Try to initialize the controller manager with retries
+	cm, err := newControllerManagerWithRetry(cfg, controllerOptions)
 	if err != nil {
 		return nil, err
 	}
-	manager = &ControllerManager{
-		Manager: controllerManager,
-	}
+	manager = cm
```
michi-covalent:
```suggestion
```

pkg/manager/manager.go
```diff
@@ -85,13 +90,12 @@ func newControllerManager() (*ControllerManager, error) {
 		cfg = ctrl.GetConfigOrDie()
 		inCluster = true
 	}
-	controllerManager, err := ctrl.NewManager(cfg, controllerOptions)
+	// Try to initialize the controller manager with retries
+	cm, err := newControllerManagerWithRetry(cfg, controllerOptions)
```
michi-covalent:
```suggestion
	manager, err := newControllerManagerWithRetry(cfg, controllerOptions)
```

pkg/manager/manager.go
```diff
@@ -216,3 +220,64 @@ func (cm *ControllerManager) FindPod(podID string) (*corev1.Pod, error) {
...
+func newControllerManagerWithRetry(cfg *rest.Config, controllerOptions ctrl.Options) (cm *ControllerManager, err error) {
+
+	retryCount := conf.K8sConfigRetry()
+	if retryCount < 0 {
+		// Infinite retries, until connection succeeds.
+		logger.GetLogger().Info("setting retryCount to max int64 for infinite retries")
+		retryCount = math.MaxInt64
+	} else if retryCount == 0 {
+		return nil, errors.New("retryCount must be a positive integer or negative for infinite retries (0 is invalid)")
+	}
```
michi-covalent, line 247: probably makes sense to take `0` as a valid config. `0` retries should map to the current behavior of not retrying 💭

michi-covalent (commented):
thanks! it looks great, just a couple of minor nits ✨

pkg/watcher/conf/conf.go
shpalani: Do you want a configurable retry config? There were so many flags, hence I was debating?

As per the design, clients in code can override this function
Sample, 
K8sConfigRetry() (retryAttempts int) { 
        // Infinite retries until success.
        return -1
}

pkg/manager/manager.go
shpalani: 1 means no retries, just one execution of the function.  0 will break backward compatability.

pkg/elf/usdt_args.go
shpalani: Ok, removed.

pkg/watcher/conf/conf.go
shpalani: Added a command line flag for configurable retry.

michi-covalent (approved):
thanks! ❤️

PR #4136, opened 2025-09-29
policies: only warn once for stats and mode

kkourt:
If, for whatever, reason the bpf maps for stats and mode do not exist in the BPF filesystem, we issue a warning. In the cases where this happens, the warnings can get noisy. Introduce an atomic variable in every collection to only issue the warning once per poilcy.


kkourt:
> LGTM, maybe this is something we want to extend like kernel WARN_ONCE. I wonder if we can have something satisfactory in Golang!

:nod:

PR #4138, opened 2025-09-29
Pr/olsajiri/flags

olsajiri:
wip


tixxdz (approved):
Thanks @mtardy @olsajiri , the flag has not been used lately! we transparently collect the values now! so LGTM

PR #4140, opened 2025-10-01
pkg/crdutils: fix standalone custom resources validation

mtardy:
This fixes commit 29b76c402b68 ("Refactor CRD defaulting and validation as generic") that was enhancing commit a9c9a0bd83a8 ("pkg/tracingpolicy: add k8s validation for meta and spec").

I think it's a typo that was introduced because when do the validation of the object, we give the function two versions of that object, cr that is typed and was created by yaml.UnmarshalStrict the JSON object with K8s default, and unstr that is the unstructured object version of what was passed as input, but with K8s default. We need these two objects to do the validation respectively on the ObjectMeta and the Spec.

When Unmarshaling the JSON object to the GenericTracingPolicy (or others) types, all fields are set to their Golang defaults (!) in addition to the K8s default that were already applied. So some missing fields, that were defaulted by the K8s default were injected in the process and thus only partial validation was done.

### DIFF
--- a/examples/tracingpolicy/security-socket-connect-block-others.yaml
+++ b/examples/tracingpolicy/security-socket-connect-block-others.yaml
@@ -31,7 +31,7 @@ spec:
       - index: 1
         operator: "SPort"
         values:
-        - 80
+        - "80"
       matchBinaries:
       - operator: "In"
         values:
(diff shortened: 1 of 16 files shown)


mtardy:
Interestingly it found issues in our own tests and examples https://github.com/cilium/tetragon/pull/4140/commits/276584087e9e5b5055b80fa320b78fc44392c21a.

PR #4150, opened 2025-10-07
k8s: Avoid hard coded CRD.spec.group

sayboras:
### Description
The value should be coming from embed file instead.

### Changelog

```release-note
k8s: Avoid hard coded CRD.spec.group
```


michi-covalent (commented):
do we still need this `CustomResourceDefinitionSchemaVersionKey` const defined?

sayboras:
> do we still need this `CustomResourceDefinitionSchemaVersionKey` const defined?

it's still used to check the version :(

PR #4151, opened 2025-10-07
fix: always close the bpf link in `detectKprobeMulti` before returning

Andreagit97:
### Description

This PR ensures we always close the bpf link in `detectKprobeMulti` before returning

### DIFF
--- a/pkg/bpf/detect_linux.go
+++ b/pkg/bpf/detect_linux.go
@@ -76,8 +76,12 @@ func detectKprobeMulti() bool {
 	syms := []string{"vprintk"}
 	opts := link.KprobeMultiOptions{Symbols: syms}
 
-	_, err = link.KprobeMulti(prog, opts)
-	return err == nil
+	link, err := link.KprobeMulti(prog, opts)
+	if err != nil {
+		return false
+	}
+	link.Close()
+	return true
 }
 
 func HasKprobeMulti() bool {


pkg/bpf/detect_linux.go
```diff
@@ -78,8 +76,12 @@ func detectKprobeMulti() bool {
...
 	opts := link.KprobeMultiOptions{Symbols: syms}
 
-	_, err = link.KprobeMulti(prog, opts)
-	return err == nil
+	link, err := link.KprobeMulti(prog, opts)
+	if err != nil {
+		return false
+	}
+	link.Close()
+	return true
```
mtardy, line 84: nit: maybe keep it close to existing code?
```suggestion
	link, err = link.KprobeMulti(prog, opts)
	defer link.Close()
	return err == nil
```

mtardy (commented):
Thanks! Technically it won't change the code as the link was GCed and thus closed but that's a cleanup :)!

pkg/bpf/detect_linux.go
andreaterzolo: uhm if `err!=nil`, `link` would probably be `nil`, so there is the risk of ending up calling `link.Close()` on a nil pointer

andreaterzolo: BTW, I agree, I just noticed it running a test and looking at the loaded ebpf progs with bpftool, otherwise I would have never seen it

mtardy: > uhm if err!=nil, link would probably be nil, so there is the risk of ending up calling link.Close() on a nil pointer

you are right sorry!

PR #4154, opened 2025-10-08
fix

olsajiri:
wip


mtardy:
could you push this patch on top of https://github.com/cilium/tetragon/pull/4137?

PR #4155, opened 2025-10-08
tetragon: assorted fixes

olsajiri:
assorted fixes for https://github.com/cilium/tetragon/pull/4137


mtardy:
I'd rather merge https://github.com/cilium/tetragon/pull/4137 so that it doesn't need another rebase afterward but that's almost the same since both needs rebase anyway rn.

PR #4158, opened 2025-10-09
feat: allow LSM funcs to be overridden on SLE

### LINKED ISSUE #4157: allow override on LSM functions if CONFIG_BPF_KPROBE_OVERRIDE is not enabled

With #1609 , we're able to override security_ LSM functions using BPF_MODIFY_RETURN.  However, the `preValidateKprobes()` still blocks a policy that hooks on LSM functions with override, if kernel CONFIG_BPF_KPROBE_OVERRIDE compile option is not enabled in the Linux kernel.  The error message and the tracing policy is as below. 

```
level=warn msg="adding tracing policy failed" error="policy handler 'tracing' failed loading policy 'deploy-ubuntu-deployment': validation failed: error in spec.kprobes[0]: error override action not supported, bpf_override_return helper not available"
```

```
kind: TracingPolicyNamespaced
metadata:
  name: deploy-ubuntu-deployment
spec:
  kprobes:
  - args:
    - index: 0
      type: linux_binprm
    call: security_bprm_creds_for_exec
    selectors:
    - matchActions:
      - action: Override
        argError: -1
    syscall: false
  options:
  - name: disable-kprobe-multi
    value: "1"
  podSelector:
    matchLabels:
      app: ubuntu
```

The policy should be able to be supported because it relies on fmod_ret instead of CONFIG_BPF_KPROBE_OVERRIDE.

`preValidateKprobes()` might be able to be changed, so this kind of policy will be allowed if kprobe override is not available. 

### Code of Conduct

holyspectral:
Fixes #4157

### Description

The override action in kprobe sensor, when hooked on LSM functions, is implemented based on fmod_ret instead of kprobe_override.

However, the preValidateKprobes function still rejects this kind of policy if CONFIG_BPF_KPROBE_OVERRIDE is not enabled in the kernel.

This commit allows this kind of policy to be loaded if CONFIG_BPF_KPROBE_OVERRIDE is not enabled.

### DIFF
--- a/pkg/sensors/tracing/generickprobe_test.go
+++ b/pkg/sensors/tracing/generickprobe_test.go
@@ -236,3 +236,140 @@ func Test_DisableEnablePolicy_KernelMemoryBytes(t *testing.T) {
 	assert.Equal(t, tetragon.TracingPolicyState_TP_STATE_ENABLED, list.Policies[0].State)
 	assert.NotZero(t, list.Policies[0].KernelMemoryBytes)
 }
+
+func Test_validateOverride(t *testing.T) {
+	tests := []struct {
+		name     string
+		f        *v1alpha1.KProbeSpec
+		funcName string
+		useMulti bool
+		wantErr  bool
+	}{
+		{
+			name: "override on syscalls",
+			f: &v1alpha1.KProbeSpec{
+				Call:    "sys_execve",
+				Return:  false,
+				Syscall: true,
+				Message: "",
+				Args: []v1alpha1.KProbeArg{
+					{
+						Index: 0,
+						Type:  "string",
+					},
+				},
+				Selectors: []v1alpha1.KProbeSelector{
+					{
+						MatchActions: []v1alpha1.ActionSelector{
+							{
+								Action:   "Override",
+								ArgError: -1,
+							},
+						},
+					},
+				},
+			},
+			funcName: "sys_execve",
+			useMulti: false,
+			wantErr:  !bpf.HasOverrideHelper(),
+		},
+		{
+			name: "override on lsm funcs",
+			f: &v1alpha1.KProbeSpec{
+				Call:    "security_bprm_creds_for_exec",
+				Return:  false,
+				Syscall: false,
+				Message: "",
+				Args: []v1alpha1.KProbeArg{
+					{
+						Index: 0,
+						Type:  "linux_binprm",
+					},
+				},
+				Selectors: []v1alpha1.KProbeSelector{
+					{
+						MatchActions: []v1alpha1.ActionSelector{
+							{
+								Action:   "Override",
+								ArgError: -1,
+							},
+						},
+					},
+				},
+			},
+			funcName: "security_bprm_creds_for_exec",
+			useMulti: false,
+			wantErr:  !bpf.HasModifyReturn(),
+		},
+		{
+			name: "override on lsm funcs with multi-kprobe enabled should be denied",
+			f: &v1alpha1.KProbeSpec{
+				Call:    "security_bprm_creds_for_exec",
+				Return:  false,
+				Syscall: false,
+				Message: "",
+				Args: []v1alpha1.KProbeArg{
+					{
+						Index: 0,
+						Type:  "linux_binprm",
+					},
+				},
+				Selectors: []v1alpha1.KProbeSelector{
+					{
+						MatchActions: []v1alpha1.ActionSelector{
+							{
+								Action:   "Override",
+								ArgError: -1,
+							},
+						},
+					},
+				},
+			},
+			funcName: "security_bprm_creds_for_exec",
+			useMulti: true,
+			wantErr:  true,
+		},
+		{
+			name: "override on non-lsm and non-syscall functions should be denied",
+			f: &v1alpha1.KProbeSpec{
+				Call:    "dentry_open",
+				Return:  false,
+				Syscall: false,
+				Message: "",
+				Args: []v1alpha1.KProbeArg{
+					{
+						Index: 0,
+						Type:  "linux_binprm",
+					},
+				},
+				Selectors: []v1alpha1.KProbeSelector{
+					{
+						MatchActions: []v1alpha1.ActionSelector{
+							{
+								Action:   "Override",
+								ArgError: -1,
+							},
+						},
+					},
+				},
+			},
+			funcName: "dentry_open",
+			useMulti: true,
+			wantErr:  true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotErr := validateOverride(tt.f, tt.funcName, tt.useMulti)
+			if gotErr != nil {
+				if !tt.wantErr {
+					t.Errorf("validateOverride() failed: %v", gotErr)
+				}
+				return
+			}
+			if tt.wantErr {
+				t.Fatal("validateOverride() succeeded unexpectedly")
+			}
+		})
+	}
+}


holyspectral:
@olsajiri thanks for the input!  My intention was to catch corner cases, e.g., a policy that overrides return values from both syscall and lsm hooks. To do that, I would need to check each call.  

d08a005cdb1e6cdd49c12d7a593f219c1087f093 would defer the check to verifier when I try to override a return value on a syscall without kprobe_override enabled, which might be a little less than ideal. 

```
level=warn msg="updateTracingPolicy: failed to add new policy" new-name=deploy-ubuntu-deployment error="sensor generic_kprobe from collection deploy-ubuntu-deployment failed to load: failed prog /var/lib/tetragon/bpf_generic_kprobe_v61.o kern_version 394240 loadInstance: opening collection '/var/lib/tetragon/bpf_generic_kprobe_v61.o' failed: program generic_kprobe_override: load program: invalid argument: unknown func bpf_override_return#58 (24 line(s) omitted)"
```

I hope this makes sense to you.  And thanks for your input again!

olsajiri:
> @olsajiri thanks for the input! My intention was to catch corner cases, e.g., a policy that overrides return values from both syscall and lsm hooks. To do that, I would need to check each call.
> 
> [d08a005](https://github.com/cilium/tetragon/commit/d08a005cdb1e6cdd49c12d7a593f219c1087f093) would defer the check to verifier when I try to override a return value on a syscall without kprobe_override enabled, which might be a little less than ideal.

could we do that in addKprobe? it has already some override checks, would be great to have it together

holyspectral:
@olsajiri Thanks for pointing that out!  I've changed the code to include `addKprobe()` in scope.  I keep the validation code in `preValidateKprobes()` because I'm not sure if `preValidateKprobes()` would be called in the future to validate the v1alpha1.KProbeSpec, say, in a dry run, but feel free to let me know if we want to remove it.

holyspectral:
Sorry I messed up the diff a little bit.  The change since the last change can be seen here: https://github.com/cilium/tetragon/compare/754dea7..883265fba3437f04773d864c48822998788806fc

holyspectral:
Looks like many tests are failing.  Will take a look.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -408,6 +408,30 @@ type kpValidateInfo struct {
...
+		}
+	} else {
+		if !bpf.HasOverrideHelper() {
+			return errors.New("override action not supported on syscalls, bpf_override_return helper not available")
+		}
+
+		if !f.Syscall {
+			return errors.New("override action can be used only with syscalls and security_ hooks")
+		}
+	}
```
kkourt, line 439: nit: how about structuring this as:

```go
if isSecurityFunc {
    ...
} else if f.Syscall {
    ...
} else {
    return errors.New("override action can be used only with syscalls and security_ hooks") 
}
```

kkourt (approved):
LGTM, thanks.

kkourt:
Thanks! Let's  wait for @olsajiri's review as well.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -733,16 +750,9 @@ func addKprobe(funcName string, instance int, f *v1alpha1.KProbeSpec, in *addKpr
...
-
 	if selectors.HasOverride(f) {
-		if isSecurityFunc && in.useMulti {
-			return errFn(fmt.Errorf("error: can't override '%s' function with kprobe_multi, use --disable-kprobe-multi option",
-				funcName))
-		}
-		if isSecurityFunc && !bpf.HasModifyReturn() {
-			return errFn(fmt.Errorf("error: can't override '%s' function without fmodret support",
-				funcName))
+		if err := validateOverride(f, funcName); err != nil {
```
olsajiri: do we still need this, I think we either need to check in preValidateKprobe or in addKprobe, or do I miss some scenario where first checkwon't catch it?

pkg/sensors/tracing/generickprobe_test.go
```diff
@@ -236,3 +237,135 @@ func Test_DisableEnablePolicy_KernelMemoryBytes(t *testing.T) {
...
+						},
+					},
+				},
+			},
+			wantErr: !bpf.HasModifyReturn() || !bpf.HasOverrideHelper(),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := preValidateKprobes(logger.DefaultSlogLogger, tt.kprobes, nil, nil)
```
mtardy: This is typically when writing tests that you would realize that you shouldn't pass the logger as the first arg... But not at all related to this PR, just thinking out loud on the existing code.

pkg/sensors/tracing/generickprobe.go
holyspectral: No we don't.  The check was kept in two places just for sanity, like if someone uses the `preValidateKprobe()` or `addKprobe` directly in the future then the check would still be present.  

Having said that, I think it makes sense to keep it simple.  I will remove the check in `addKprobe()`.   Thanks!

pkg/sensors/tracing/generickprobe.go
```diff
@@ -733,19 +750,6 @@ func addKprobe(funcName string, instance int, f *v1alpha1.KProbeSpec, in *addKpr
 		}
 	}
 
-	isSecurityFunc := strings.HasPrefix(funcName, "security_")
-
-	if selectors.HasOverride(f) {
-		if isSecurityFunc && in.useMulti {
-			return errFn(fmt.Errorf("error: can't override '%s' function with kprobe_multi, use --disable-kprobe-multi option",
```
olsajiri, line 748: so this check is missing in `validateOverride`, I it needs `in.useMulti` so I think we need to move `validateOverride` to `addKprobe`

holyspectral: Thanks for catching it.  I'll change accordingly along with the unit-test.

holyspectral:
Moved the check to addKprobe as @olsajiri suggested.  @mtardy please feel free to take a look since the test is rewritten to accommodate the change.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -414,6 +414,33 @@ type kpValidateInfo struct {
...
+	isSecurityFunc := strings.HasPrefix(funcName, "security_")
+
+	if isSecurityFunc {
+		// LSM functions
+		if !bpf.HasModifyReturn() {
+			return errors.New("override action not supported on security_ hooks, fmod_ret not available")
+		}
+
+		if useMulti {
+			return fmt.Errorf("can't override '%s' function with kprobe_multi, use --disable-kprobe-multi option", funcName)
```
mtardy: small nit: if you can use the real flagKey here

```suggestion
			return fmt.Errorf("can't override '%s' function with kprobe_multi, use --%s option", funcName, config.keyDisableKprobeblab...)
```

mtardy (approved):
lgtm, only a small nit, tell me if you want to fix it or not :)

holyspectral:
@mtardy l've implemented the small change.  Feel free to let me know if there is anything else!

mtardy:
> @mtardy l've implemented the small change. Feel free to let me know if there is anything else!

perfect! any chance you can squash this new commit into the first commit, we rebase and merge (see https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/#commit-changes).

holyspectral:
@mtardy sure!  I was thinking the same thing but not sure because everyone has different preference.  I will do that in a minute.

holyspectral:
@mtardy please feel free to let me know if there is anything that I can do to get this PR merged.  Much appreciated.

mtardy:
> @mtardy please feel free to let me know if there is anything that I can do to get this PR merged. Much appreciated.

ah sorry I forgot to merge this indeed! Thank you for the reminder :)

PR #4170, opened 2025-10-10
selectors: fix off by one bounds check

andrewstrohman:
This prevents a panic when tetragon is
supplied an invalid data filter within
a tracing policy.

### DIFF
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -857,7 +857,7 @@ func dataIndexType(arg *v1alpha1.ArgSelector, sig []v1alpha1.KProbeArg) (uint32,
 	if len(arg.Args) > 0 {
 		return argIndexTypeFromArgs(arg, 0, sig)
 	}
-	if arg.Index > uint32(len(sig)) {
+	if arg.Index >= uint32(len(sig)) {
 		return 0, 0, errors.New("argFilter for unknown index")
 	}
 	ty := sig[arg.Index].Type


mtardy:
cc @olsajiri

andrewstrohman:
> thanks, but from quick look it wouldn't panic, would it?

The following policy causes a panic without this patch:

```
kind: TracingPolicy
metadata:
  name: matchdata-pid-not-1
spec:
  kprobes:
    - call: "sys_openat"
      syscall: true
      args:
        - index: 1
          type: "string"
      data:
        - type: "string"
          source: "current_task"
          resolve: "pid"
          index: 0 # I don't understand why this is required, or what it does.
      selectors:
        - matchData:
            - index: 1
              operator: "NotEqual"
              values:
                - "1"
        - matchArgs:
            - index: 1
              operator: "Prefix"
              values:
                - "/etc"

```

andrewstrohman:
> index: 0 # I don't understand why this is required, or what it does.

I think the problem stems from the fact that args and data both use `KProbeArg struct`. In the args case, mandating index makes sense, but I don't think it makes sense for data.

The following "fixes" this for me, but I don't like that it also applies to args as well.

```
andy@xps:~/src/tetragon$ git diff pkg/k8s/apis/cilium.io/v1alpha1/types.go
diff --git a/pkg/k8s/apis/cilium.io/v1alpha1/types.go b/pkg/k8s/apis/cilium.io/v1alpha1/types.go
index 675c1eb18..11fbf3407 100644
--- a/pkg/k8s/apis/cilium.io/v1alpha1/types.go
+++ b/pkg/k8s/apis/cilium.io/v1alpha1/types.go
@@ -70,6 +70,7 @@ type KProbeSpec struct {
 }

 type KProbeArg struct {
+       // +kubebuilder:validation:Optional
        // Position of the argument.
        Index uint32 `json:"index"`
```

I think we might have made validation changes very recently that caused this. @olsajiri  Should I create an issue for this?

olsajiri:
> > thanks, but from quick look it wouldn't panic, would it?
> 
> The following policy causes a panic without this patch:
> 
> ```
> kind: TracingPolicy
> metadata:
>   name: matchdata-pid-not-1
> spec:
>   kprobes:
>     - call: "sys_openat"
>       syscall: true
>       args:
>         - index: 1
>           type: "string"
>       data:
>         - type: "string"
>           source: "current_task"
>           resolve: "pid"
>           index: 0 # I don't understand why this is required, or what it does.
>       selectors:
>         - matchData:
>             - index: 1
>               operator: "NotEqual"
>               values:
>                 - "1"
>         - matchArgs:
>             - index: 1
>               operator: "Prefix"
>               values:
>                 - "/etc"
> ```

ugh.. missed the rest of the function, great thanks

olsajiri:
> I think we might have made validation changes very recently that caused this. @olsajiri Should I create an issue for this?

we discussed this with @kkourt , I need to come up with some validation rule that keeps the minimum for index.. will try to post it soon

PR #4171, opened 2025-10-11
tetragon: support (u)int8/16 selectors

andrewstrohman:
This allows us to author configs like the following:

```
kind: TracingPolicy
metadata:
  name: "test2"
spec:
  kprobes:
  - call: "br_switchdev_port_vlan_del"
    syscall: false
    args:
    - index: 1
      type: "uint16"
    selectors:
      - matchArgs:
        - args: [0]
          operator: "Equal"
          values:
            - "7"
```

### DIFF
--- a/pkg/sensors/tracing/tracepoint_test.go
+++ b/pkg/sensors/tracing/tracepoint_test.go
@@ -831,6 +831,12 @@ func TestUInt16Tracepoint(t *testing.T) {
 					Operator: "Equal",
 					Values:   []string{"7777"},
 				}},
+			}, {
+				MatchArgs: []v1alpha1.ArgSelector{{
+					Args:     []uint32{uint32(0)},
+					Operator: "InRange",
+					Values:   []string{"7779:7780"},
+				}},
 			}},
 		}},
 	}
@@ -852,15 +858,17 @@ func TestUInt16Tracepoint(t *testing.T) {
 	}
 
 	ops := func() {
-		listener, err := net.Listen("tcp", ":7777")
-		if err != nil {
-			panic(err)
+		for port := 7776; port <= 7781; port++ {
+			listener, err := net.Listen("tcp", fmt.Sprintf(":%d", port))
+			if err != nil {
+				panic(err)
+			}
+			listener.Close()
 		}
-		listener.Close()
 	}
 
 	perfring.RunTest(t, ctx, ops, eventFn)
-	require.Equal(t, 1, count, "expected single event")
+	require.Equal(t, 3, count, "expected three events: one for each of 7777, 7779 and 7780")
 }
 
 func testListSyscallsDupsRange(t *testing.T, checker *ec.UnorderedEventChecker, configHook string) {


pkg/selectors/kernel_test.go
```diff
@@ -43,6 +43,44 @@ func TestWriteSelectorUint32(t *testing.T) {
 	}
 }
 
+func TestWriteSelectorUint16(t *testing.T) {
```
olsajiri: please put tests in separate commit

bpf/process/types/basic.h
```diff
@@ -1273,6 +1273,218 @@ filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
...
+	case op_filter_lt:
+	case op_filter_gt:
+	case op_filter_eq:
+	case op_filter_neq:
+	case op_filter_mask:
+		return filter_8ty_selector_val(filter, args);
+	case op_filter_inmap:
+	case op_filter_notinmap:
+		return filter_8ty_map(filter, args);
+	}
```
olsajiri: would be great to add range support like we do for 32/64 bits

bpf/process/types/basic.h
```diff
@@ -1273,6 +1273,218 @@ filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
...
+
+	return 0;
+}
+
+// filter on values provided in the selector itself
+FUNC_LOCAL long
+filter_16ty_selector_val(struct selector_arg_filter *filter, char *args)
```
olsajiri, line 1366: not sure why exactly 4.19 CI tests are failing but my guess is complexity,
we might need to have this only for LARGE setup.. or do you need this in 4.19?

bpf/process/types/basic.h
andrewstrohman: Thanks for catching this! I'll add the range support.

pkg/selectors/kernel_test.go
andrewstrohman: Sounds good. My next push will have the tests broken out into a separate commit.

bpf/process/types/basic.h
andrewstrohman: When I ran:
`../veristat/src/veristat bpf/objs/bpf_generic_kprobe.o | grep generic_kprobe_filter_arg`
I saw:
```
bpf_generic_kprobe.o  generic_kprobe_filter_arg      failure              0      0       0          4500           0
```

So I think I exceeded the 4K instruction limit. (However, the column that corresponds to 4500 is the "Program Size" column, not the "Insns" column.) After adding `#ifdef __LARGE_BPF_PROG`, I see it's below 4096, and 4.19 vmtests passes now.

bpf/process/types/basic.h
```diff
@@ -1273,6 +1273,182 @@ filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
...
+	__u8 *v = (__u8 *)&filter->value;
+	int i, j = 0;
+
+#pragma unroll
+	for (i = 0; i < MAX_MATCH_VALUES; i++) {
+		__u8 w = v[i];
+		bool res;
+
+		switch (filter->op) {
+#ifdef __LARGE_BPF_PROG
```
olsajiri: let's remove this ifdef, filter_8ty_selector_val will not get compiled for 4.19 now,
so let's not confuse the source for future

there's couple of similar instances in this change

andrewstrohman: I removed here and also in filter_16ty_selector_val. But I didn't find any additional locations where I could remove it.

andrewstrohman:
I've changed the description in the commit messages to reflect that fact that this change is more general than resolve -- it's general selector support.

bpf/process/types/basic.h
```diff
@@ -1497,6 +1618,27 @@ filter_32ty(struct selector_arg_filter *filter, char *args)
...
+	case op_filter_eq:
+	case op_filter_neq:
+	case op_filter_mask:
+		return filter_16ty_selector_val(filter, args);
+	case op_filter_inmap:
+	case op_filter_notinmap:
+		return filter_16ty_map(filter, args);
+	case op_in_range:
+	case op_notin_range:
+		return filter_16ty_range(filter, args);
```
kkourt, line 1740: nit: can we also have `#ifdef __LARGE_BPF_PROG` here instead of a dummy `filter_16ty_range` function similar to what we do in `selector_arg_offset`?

kevsecurity:
In your examples, if you put "int" instead of "uint16", does it just work? If so (and it should), then (I think) you only need to add a switch case in pkg/selectors/kernel.go:writeMatchValues() for gt.GenericU16Type (I would put it on the same case as gt.GenericU32Type with no extra code; it should magically work). I think this previously worked when validation was more lax  (as the K8s validation includes uint16, for example).

If that works, then the same can be done for the signed versions (maybe) and the uint8 type.

andrewstrohman:
> In your examples, if you put "int" instead of "uint16", does it just work? If so (and it should), then (I think) you only need to add a switch case in pkg/selectors/kernel.go:writeMatchValues() for gt.GenericU16Type (I would put it on the same case as gt.GenericU32Type with no extra code; it should magically work). I think this previously worked when validation was more lax (as the K8s validation includes uint16, for example).
> 
> If that works, then the same can be done for the signed versions (maybe) and the uint8 type.

Thanks for this explanation. Your suggestion works for having tetragon accept the policy and reporting the args in the event correctly. But the selector filter logic is still missing [here](https://github.com/cilium/tetragon/blob/27c9abe39c448c6f823c607eaad483d8c5717ecb/bpf/process/types/basic.h#L1793-L1803). As a consequence of this, we get filter matches when we shouldn't.

So, I'm going to use your suggestion to simplify the userspace side of things. I'll keep the bpf side of this patch, but adapt it to cope with the fact that when matching against (u)int8/16 types, the configured matches are enlarged to 32 bit types when passed to the bpf program.

kevsecurity:
> Thanks for this explanation. Your suggestion works for having tetragon accept the policy and reporting the args in the event correctly. But the selector filter logic is still missing [here](https://github.com/cilium/tetragon/blob/27c9abe39c448c6f823c607eaad483d8c5717ecb/bpf/process/types/basic.h#L1793-L1803). As a consequence of this, we get filter matches when we shouldn't.

This makes sense. I think maybe I only added enough to match values within a struct sock or struct sk_buff (eg port number, protocol, family, state), and left generic u8 and u16 for the future.

> So, I'm going to use your suggestion to simplify the userspace side of things. I'll keep the bpf side of this patch, but adapt it to cope with the fact that when matching against (u)int8/16 types, the configured matches are enlarged to 32 bit types when passed to the bpf program.

Sounds good.

kkourt:
@andrewstrohman Thanks! Can you please add some context about the discussions here and the approach in the commit messages?

andrewstrohman:
> Can we try to figure out what the existing uses of `gt.GenericU16Type` mean and if they are used in the code?
> 
> (The same is true for `u8`).

An example for each of the uses follow.

> In `writeRangeInMap`:
> 
> https://github.com/cilium/tetragon/blob/77579570474e3ee2b130edd8553e752adccab603/pkg/selectors/kernel.go#L428
> 
> https://github.com/cilium/tetragon/blob/77579570474e3ee2b130edd8553e752adccab603/pkg/selectors/kernel.go#L449

The idea here is: instead of sending the range to the bpf program, we store all the values in the range within the map. So if the range is large, we have a lot of keys in the map.

```
kind: TracingPolicy
metadata:
  name: "test"
spec:
  tracepoints:
  - subsystem: "tcp"
    event: "tcp_destroy_sock"
    args:
    - index: 5
      type: "uint16"
    - index: 6
      type: "uint16"
    - index: 8
      type: "uint32"
    - index: 9
      type: "uint32"
    selectors:
      - matchArgs:
        - args: [1]
          operator: "InMap"
          values:
          - "79:81"
 ```  


> And in `writeMatchRangesInMap`:
> 
> https://github.com/cilium/tetragon/blob/77579570474e3ee2b130edd8553e752adccab603/pkg/selectors/kernel.go#L489

This is one is actually in writeListValuesInMap()
The idea here is that would could do something like:
```
kind: TracingPolicy
metadata:
  name: "test"
spec:
  lists:
  - name: "ports"
    type: "uint16"
    values:
    - "77"
    - "81"
  tracepoints:
  - subsystem: "tcp"
    event: "tcp_destroy_sock"
    args:
    - index: 5
      type: "uint16"
    - index: 6
      type: "uint16"
    - index: 8
      type: "uint32"
    - index: 9
      type: "uint32"
    selectors:
      - matchArgs:
        - args: [1]
          operator: "InMap"
          values:
          - "list:ports"
```

But, it doesn't work because:
> validation failed: spec.lists[0].type in body should be one of [syscalls generated_syscalls generated_ftrace]

We don't support lists of ints. So we could either remove this reference of uint8/16 or we could support lists of ints. It looks like lists are only supported for tracepoints.

> https://github.com/cilium/tetragon/blob/77579570474e3ee2b130edd8553e752adccab603/pkg/selectors/kernel.go#L569
> 

This one really is in writeMatchValuesInMap()

```
kind: TracingPolicy
metadata:
  name: "test"
spec:
  tracepoints:
  - subsystem: "tcp"
    event: "tcp_destroy_sock"
    args:
    - index: 5
      type: "uint16"
    - index: 6
      type: "uint16"
    - index: 8
      type: "uint32"
    - index: 9
      type: "uint32"
    selectors:
      - matchArgs:
        - args: [1]
          operator: "InMap"
          values:
          - "77"
          - "81"
 ```

This exercise made me realize that I needed to add more on the userspace side to support ranges of uint16. 

```
kind: TracingPolicy
metadata:
  name: "test"
spec:
  tracepoints:
  - subsystem: "tcp"
    event: "tcp_destroy_sock"
    args:
    - index: 5
      type: "uint16"
    - index: 6
      type: "uint16"
    - index: 8
      type: "uint32"
    - index: 9
      type: "uint32"
    selectors:
      - matchArgs:
        - args: [1]
          operator: "InRange"
          values:
          - "77:81"
```
This works if I do:
```
diff --git a/pkg/selectors/kernel.go b/pkg/selectors/kernel.go
index fb8b68d7b..869766c36 100644
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -521,7 +521,7 @@ func writeMatchValuesRange(k *KernelSelectorState, values []string, ty uint32) e
                }

                switch ty {
-               case gt.GenericIntType, gt.GenericS32Type, gt.GenericSizeType, gt.GenericU32Type:
+               case gt.GenericIntType, gt.GenericS32Type, gt.GenericSizeType, gt.GenericU32Type, gt.GenericU16Type, gt.GenericU8Type, gt.GenericS16Type, gt.GenericS8Type:
                        if rangeMax > math.MaxUint32 || rangeMin > math.MaxUint32 {
                                return fmt.Errorf("range out of range (%d:%d) allowed max %d", rangeMin, rangeMax, math.MaxUint32)
                        }
```

So the only open item is whether we want to support lists of integers. If not, I can remove the reference here: https://github.com/cilium/tetragon/blob/77579570474e3ee2b130edd8553e752adccab603/pkg/selectors/kernel.go#L489

bpf/process/types/basic.h
andrewstrohman: I did this to match the existing convention used for filter_(32|64)ty_range. I'll tack on a commit that does this for the 32 and 64 bit versions as well, for consistency.

andrewstrohman:
> @andrewstrohman Thanks! Can you please add some context about the discussions here and the approach in the commit messages?

Yes, I added this to the commit messages:
> To simplify the userspace side, when we send
(u)int8/16 match values to the bpf program,
we expand them to 32 bit integers.

kevsecurity (approved):

olsajiri:
looks good, please put changelog in each commit, like https://github.com/cilium/tetragon/pull/4171/commits/dbeb746199adab27ea945090aa44f9260e0adab3 is a mystery on first sight

pkg/selectors/kernel.go
```diff
@@ -521,7 +521,7 @@ func writeMatchValuesRange(k *KernelSelectorState, values []string, ty uint32) e
 		}
 
 		switch ty {
-		case gt.GenericIntType, gt.GenericS32Type, gt.GenericSizeType, gt.GenericU32Type:
+		case gt.GenericIntType, gt.GenericS32Type, gt.GenericSizeType, gt.GenericU32Type, gt.GenericU16Type, gt.GenericU8Type, gt.GenericS16Type, gt.GenericS8Type:
```
kkourt, line 524: suggestion: Would it make sense to extend the TestUInt16Tracepoint test in this commit to also add a filter to test this? (I'm assuming this adds support for filters).

andrewstrohman: Yeap! Next push will have this.

andrewstrohman:
I've added more context to the commit messages.

olsajiri (approved):
looks good ,thanks

PR #4173, opened 2025-10-11
tetragon: add uprobe override action

olsajiri:
Adding support for uprobe override action.

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -720,6 +720,10 @@ this action for enforcement.
 
 ### Override action
 
+Override action is defined for kprobe and uprobe selectors.
+
+#### Override action for kprobe
+
 `Override` action allows to modify the return value of call. While `Sigkill`
 will terminate the entire process responsible for making the call, `Override`
 will run in place of the original kprobed function and return the value
@@ -769,6 +773,64 @@ Starting from kernel version `5.7` overriding `security_` hooks is also possible
 For kernel developers: if you want to override your kernel functions then
 ensure they properly follow the [Error Injectable Functions](https://docs.kernel.org/fault-injection/fault-injection.html#error-injectable-functions) guide.
 {{< /caution >}}
+
+#### Override action for uprobe
+
+{{< warning >}}
+Beware, here be dragons!!! Use with caution, it could easily crash traced application.
+{{< /warning >}}
+
+Note that `Override` action can be used only on `x86_64` architecture.
+
+Similar to kprobe, the uprobe `Override` action allows to modify the return value
+of user space call with `argError` argument, like:
+
+```yaml
+uprobes:
+- path: "test"
+  symbols:
+  - "func"
+  selectors:
+  - matchActions:
+    - action: Override
+      argError: 123
+```
+
+Note that it can be successfully used only when following conditions are met:
+- uprobe is attached to the beggining of the user space function;
+- user space function is called via `call` instruction;
+- user space function returns `int` type.
+
+It's possible to override specific registers with arbitrary value with `argRegs`
+argument, like:
+
+```yaml
+uprobes:
+- path: "test"
+  symbols:
+  - "func"
+  selectors:
+  - matchActions:
+    - action: Override
+      argRegs:
+      - "rax=11"
+      - "rbp=(%rsp)"
+      - "rip=8(%rsp)"
+      - "rsp=8%rsp"
+```
+
+The `argRegs` argument is an array of strings where each string start with destination
+register name followed by `=` and an assignment expression, which can be one of the
+following types:
+
+- constant `rax=11`
+- register `rbp=(%rsp)`
+- register plus offset `rip=8(%rsp)`
+- dereference of register plus offset `rsp=8%rsp`
+
+{{< note >}}
+This interface is likely to be changed in the future.
+{{< /note >}}
 
 ### FollowFD action
 


pkg/bpf/detect_linux.go
```diff
@@ -536,14 +537,43 @@ func HasAuditLoginuid() bool {
 	return auditLoginuid.detected
 }
 
+func detectUprobeRegsChange() bool {
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Name: "uprobe_regs",
+		Type: ebpf.Kprobe,
+		Instructions: asm.Instructions{
+			asm.StoreImm(asm.R1, 0x50 /* rax */, 0xdead, 8),
```
mtardy: could you use your `asm.RegOffset` helper for `rax` since we have it. I didn't get how it was related to uprobes_regs just by reading but I'll trust you on that!

pkg/asm/assignment_amd64_test.go
```diff
@@ -0,0 +1,82 @@
...
+package asm
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAssignment(t *testing.T) {
```
mtardy, line 15: It seems that `ParseAssignment` would be a perfect candidate for a little fuzzing tests, checking that at least basic fuzzing do not make it crash quickly https://go.dev/doc/security/fuzz/.

I wrote a little patch that you can merge into your current PR: https://github.com/cilium/tetragon/pull/4209 so that you can rebase or do anything you want with it.

pkg/asm/assignment_amd64.go
```diff
@@ -0,0 +1,184 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build amd64 && linux
+// +build amd64,linux
```
mtardy: you can drop the old style now in your commits, the new `go:build` was added in 1.17
```
// +build amd64,linux
```

bpf/process/uprobe_offload.h
```diff
@@ -0,0 +1,194 @@
...
+FUNC_INLINE void do_uprobe_override(void *ctx, __u32 idx)
+{
+	__u64 id = get_current_pid_tgid();
+	__u32 *idxp;
+
+	/*
+	 * TODO: this should not happen, it means that the override
+	 * program was not executed for some reason, we should do
+	 * warning in here
+	 */
```
mtardy, line 54: please use `errmetrics(something)` then

pkg/api/processapi/processapi_regs_amd64.go
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build amd64 && linux
+// +build amd64,linux
```
mtardy: in this commit as well you have some old school constraints

pkg/sensors/tracing/genericuprobe_amd64.go
```diff
@@ -0,0 +1,26 @@
...
+	processapi "github.com/cilium/tetragon/pkg/api/processapi"
+	"github.com/cilium/tetragon/pkg/logger"
+)
+
+func populateUprobeRegs(m *ebpf.Map, regs []processapi.RegAssignment) error {
+	uprobeRegs := processapi.UprobeRegs{}
+
+	n := copy(uprobeRegs.Ass[:], regs)
+	if n != len(regs) {
+		logger.GetLogger().Warn(fmt.Sprintf("register assignments count mismatch #regs %d", len(regs)))
```
mtardy: please use structured logging instead of sprintf

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -770,6 +774,60 @@ For kernel developers: if you want to override your kernel functions then
 ensure they properly follow the [Error Injectable Functions](https://docs.kernel.org/fault-injection/fault-injection.html#error-injectable-functions) guide.
 {{< /caution >}}
 
+#### Override action for uprobe
+
+**BEWARE, HERE BE DRAGONS!!! USE WITH CAUTION COULD EASILY CRASH TRACED APPLICATION**
```
```suggestion
{{< warning >}}
Beware, here be dragons!!! Use with caution, it could easily crash traced application.
{{< /warning >}}
```

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -770,6 +774,60 @@ For kernel developers: if you want to override your kernel functions then
...
+  selectors:
+  - matchActions:
+    - action: Override
+      argError: 123
+```
+
+Note that it can be successfully used only when following conditions are met:
+- uprobe is attached to the beggining of the user space function
+- user space function is called via `call` instruction
+- user space function returns `int` type
```
```suggestion
- uprobe is attached to the beggining of the user space function;
- user space function is called via `call` instruction;
- user space function returns `int` type.
```

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -770,6 +774,60 @@ For kernel developers: if you want to override your kernel functions then
...
+      argRegs:
+      - "rax=11"
+      - "rbp=(%rsp)"
+      - "rip=8(%rsp)"
+      - "rsp=8%rsp"
+```
+
+The `argRegs` argument is array of strings where each string start with destination
+register name followed by `=` and assignment expression, which can be one of the
+following types:
```
```suggestion
The `argRegs` argument is an array of strings where each string start with destination
register name followed by `=` and an assignment expression, which can be one of the
following types:
```

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -770,6 +774,60 @@ For kernel developers: if you want to override your kernel functions then
...
 {{< /caution >}}
 
+#### Override action for uprobe
+
+**BEWARE, HERE BE DRAGONS!!! USE WITH CAUTION COULD EASILY CRASH TRACED APPLICATION**
+
+Note that `Override` action can be used only on `x86_64` architecture.
+
+Similar to kprobe the uprobe `Override` action allows to modify the return value
+of user space call with `argError` argument, like:
```
```suggestion
Similar to kprobe, the uprobe `Override` action allows to modify the return value
of user space call with `argError` argument, like:
```

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -770,6 +774,60 @@ For kernel developers: if you want to override your kernel functions then
...
+The `argRegs` argument is array of strings where each string start with destination
+register name followed by `=` and assignment expression, which can be one of the
+following types:
+
+- constant `rax=11`
+- register `rbp=(%rsp)`
+- register plus offset `rip=8(%rsp)`
+- dereference of register plus offset `rsp=8%rsp`
+
+Note that this interface is likely to be changed in the future.
```
```suggestion
{{< caution >}}
This interface is likely to be changed in the future.
{{< /warning >}}
```
or
```suggestion
{{< note >}}
This interface is likely to be changed in the future.
{{< /note >}}
```

pkg/asm/assignment_amd64_test.go
```diff
@@ -0,0 +1,82 @@
...
+
```
mtardy: nit: whitespace

mtardy (approved):
This is really fun tracing feature. I'd add a little fuzzing thing https://github.com/cilium/tetragon/pull/4209. It's mostly nits.

pkg/asm/reg_offset_amd64.go
```diff
@@ -0,0 +1,51 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build amd64 && linux
+// +build amd64,linux
```
kkourt: As @mtardy  said in another comment, the `+build amd64,linux` part is not needed. Line 4 alone should work.

pkg/selectors/kernel.go
```diff
@@ -1434,6 +1434,7 @@ type KernelSelectorArgs struct {
 	ActionArgTable *idtable.Table
 	ListReader     ValueReader
 	Maps           *KernelSelectorMaps
+	IsUprobe       bool
```
kkourt, line 1445: nit: In these situations, I think it's better to pass an enum with the type of the hook instead of a bool, since the enum is more extensible.

That being said, this can be done as a followup if and whenever it is needed.

bpf/process/generic_calls.h
```diff
@@ -934,13 +934,19 @@ do_action(void *ctx, __u32 i, struct selector_action *actions, bool *post, bool
 		}
 		break;
 	case ACTION_OVERRIDE:
+#if defined(GENERIC_UPROBE) && defined(__TARGET_ARCH_x86)
+		index = actions->act[++i];
+		do_uprobe_override(ctx, index);
+		polacct = POLICY_OVERRIDE;
```
kkourt: Can we respect monitor/enforce mode here as well? Or is there a reason we don't want to?

bpf/process/uprobe_offload.h
```diff
@@ -0,0 +1,194 @@
...
+
+FUNC_INLINE int
+write_reg(struct pt_regs *ctx, __u32 dst, __u64 val)
+{
+#define WRITE_REG(reg)                                                   \
+	asm volatile("if %[dst] != %[off] goto +1\n"                     \
+		     "*(u64 *)(%[ctx] + %[off]) = %[val]\n"              \
+		     : [ctx] "+r"(ctx), [dst] "+r"(dst), [val] "+r"(val) \
+		     : [off] "i"(offsetof(struct pt_regs, reg))          \
+		     :);
```
kkourt, line 77: Maybe I'm missing something, but this seems inneficient.

For example:
```
     344:	55 02 01 00 00 00 00 00	if r2 != 0x0 goto +0x1 <LBB7_51+0x28>
     345:	7b 13 00 00 00 00 00 00	*(u64 *)(r3 + 0x0) = r1
     346:	55 02 01 00 08 00 00 00	if r2 != 0x8 goto +0x1 <LBB7_51+0x38>
     347:	7b 13 08 00 00 00 00 00	*(u64 *)(r3 + 0x8) = r1
     348:	55 02 01 00 10 00 00 00	if r2 != 0x10 goto +0x1 <LBB7_51+0x48>
     349:	7b 13 10 00 00 00 00 00	*(u64 *)(r3 + 0x10) = r1
```

If `r2 == 0x0`, then we do `*(u64 *)(r3 + 0x0) = r1`, but we still check all other cases.

Can't we just do:
```
asm volatile("*(u64 *)(%[ctx] + %[dst]) = %[val]\n")
```

Or, if we don't trust `dst`, 
```
switch (dst) {
     case offsetof(stuct pt_regs, r15):
     case offsetof(stuct pt_regs, r14):
     ...
     case offsetof(stuct pt_regs, sp):
     asm volatile("*(u64 *)(%[ctx] + %[dst]) = %[val]\n")
}
```

Maybe it's to help verification?

pkg/sensors/tracing/genericuprobe_arm64.go
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build arm64 && linux
+// +build arm64,linux
```
kkourt: nit: no need for the `+build comment`.

pkg/selectors/kernel.go
olsajiri: yea, I wanted to keep it simple for now, it's simple change if/when we want to extend

bpf/process/uprobe_offload.h
olsajiri: > Maybe it's to help verification?

yes, I used to have there big switch for offset which clang 18 will optimize in such way that will put context pointer + arbitrary offset in register and use that with the rest of the offset value to access the context and  that will fail the verifier - which will only work for 'context pointer + offset'

clang 20 will fix that and we could use the big switch, but clang 18 needs this workaround

as for the efficiency.. I did not find an easy way to do return from assemble in inline function after we read the value, so it will go through all the offsets, but I don't think it's big hold up for now

bpf/process/generic_calls.h
olsajiri: hum, not sure what you mean.. @kkourt

pkg/bpf/detect_linux.go
olsajiri: I think it's better we do not do this for specific arch, I used offset 0 to write into first context register

pkg/asm/assignment_amd64_test.go
olsajiri: looks good, let's push it in separate PR

bpf/process/uprobe_offload.h
kkourt: OK, makes sense to me. If you do another pass, can you add a comment?

bpf/process/generic_calls.h
kkourt: I mean doing:
```
        index = actions->act[++i];
		if (enforce_mode) {
			do_uprobe_override(ctx, index);
			polacct = POLICY_OVERRIDE;
		} else {
			polacct = POLICY_MONITOR_OVERRIDE;
		}
		break;
```

So that enforce/monitor mode works for uprobe override as well.

olsajiri: ah nevermind got it.. will change

bpf/process/generic_calls.h
```diff
@@ -935,10 +935,15 @@ do_action(void *ctx, __u32 i, struct selector_action *actions, bool *post, bool
...
-		error = actions->act[++i];
 		if (enforce_mode) {
+#if defined(GENERIC_UPROBE) && defined(__TARGET_ARCH_x86)
+			index = actions->act[++i];
+			do_uprobe_override(ctx, index);
+#else
+			error = actions->act[++i];
 			do_override_action(error);
 			polacct = POLICY_OVERRIDE;
+#endif
```
kkourt, line 944: I don't think this is correct. For one, it does not  set `pollact = POLICY_OVERRIDE` for uprobes. 

Moreover, it does not update `i` in monitor mode. At the end of the function we do:
```C
	if (!err) {
		e->action = action;
		return ++i;
	}
```

So we need to maintain the correct value of `i` regardless of whether we are in monitor or enforce mode. That is, we need to do the `i++` in `error = actions->act[++i];` in both enforce and monitor cases.

So I think we need:
```C
	case ACTION_OVERRIDE:
		arg = actions->act[++i];
		if (enforce_mode) {
#if defined(GENERIC_UPROBE) && defined(__TARGET_ARCH_x86)
			do_uprobe_override(ctx, arg);
#else
			do_override_action(arg);
#endif
			polacct = POLICY_OVERRIDE;
		} else {
			polacct = POLICY_MONITOR_OVERRIDE;
		}
```

olsajiri: @kkourt  ah right, how about:
```
                if (enforce_mode) {
#if defined(GENERIC_UPROBE) && defined(__TARGET_ARCH_x86)
                        index = actions->act[++i];
                        do_uprobe_override(ctx, index);
#else
                        error = actions->act[++i];
                        do_override_action(error);
#endif
                        polacct = POLICY_OVERRIDE;
                } else {
                        polacct = POLICY_MONITOR_OVERRIDE;
                }

```

kkourt: I don't think that above would work because we do not increase `i` when in monitor mode, so when we return `i` and go for the next action, we would have wrong data because we didn't skip the `index` (or `error`).

PR #4183, opened 2025-10-13
check-links: fix the periodic check issue creation

mtardy:
For some reason (didn't investigate more) the workflow was broken and the issue was never created while it was failing for a long time recently. This should fix it.

Detected it thanks to https://github.com/cilium/tetragon/pull/4177 and fixed recent failures in https://github.com/cilium/tetragon/pull/4181


mtardy:
https://github.com/cilium/tetragon/actions/runs/18471165911/job/52625028008?pr=4183 passed, let's early merge

PR #4184, opened 2025-10-13
tetragon: Add support to retrieve environment variables

olsajiri:
storing environment variables into process event data, enabled with `--enable-process-environment-variables` option,
and we allow to redact info from the variables 

reworked a bit the exec data parsing and added the environment variables parsing contributed by @slntopp,
I left the SOB info in some changes, @slntopp please shout if you want to make some changes

### DIFF
--- a/docs/content/en/docs/concepts/events.md
+++ b/docs/content/en/docs/concepts/events.md
@@ -223,9 +223,9 @@ Since Tetragon traces the entire system, event exports might sometimes contain
 sensitive information (for example, a secret passed via a command line argument
 to a process). To prevent this information from being exfiltrated via Tetragon
 JSON export, Tetragon provides a mechanism called Redaction Filters which can be
-used to string patterns to redact from exported process arguments. These filters are written
-in JSON and passed to the Tetragon agent via the `--redaction-filters` command
-line flag or the `redactionFilters` Helm value.
+used to string patterns to redact from exported process arguments and environment
+variables. These filters are written in JSON and passed to the Tetragon agent via
+the `--redaction-filters` command line flag or the `redactionFilters` Helm value.
 
 To perform redactions, redaction filters define RE2 regular expressions in the
 `redact` field. Any capture groups in these RE2 regular expressions are redacted and
@@ -243,7 +243,7 @@ characters. For instance `\Wpasswd\W?` would be written as `{"redact": "\\Wpassw
 {{< /warning >}}
 
 For more control, you can select which binary or binaries should have their
-arguments redacted with the `binary_regex` field.
+arguments or environment variables redacted with the `binary_regex` field.
 
 As a concrete example, the following will redact all passwords passed to
 processes with the `"--password"` argument:
@@ -264,6 +264,18 @@ We can also redact these as follows:
 
 With both of the above redaction filters in place, we are now redacting all
 password arguments.
+
+Another example is to redact `SSHPASS` environment variable with:
+
+```json
+{"redact": ["(?:SSHPASS=)+(\\S+)"]}
+```
+
+Now, an event that contains the string `"SSHPASS=password"` would have that string
+replaced with `"SSHPASS=*****"`.
+
+It's also possible to store only requested environment variables with
+'--filter-environment-variables VAR1[,VAR2..]` option.
 
 ### `tetra` CLI
 


bpf/process/bpf_execve_event.c
```diff
@@ -126,6 +127,71 @@ read_args(void *ctx, struct msg_execve_event *event)
...
+
+	with_errmetrics(probe_read, &env_start, sizeof(env_start), _(&mm->env_start));
+	with_errmetrics(probe_read, &env_end, sizeof(env_end), _(&mm->env_end));
+
+	if (!env_start || !env_end)
+		return 0;
+
+	envs = (char *)p + p->size;
+	if (envs >= (char *)&event->process + BUFFER)
+		return 0;
```
FedeDP: I'd move this before the `probe_read` to avoid wasting time on reading env start and end when not needed.

FedeDP: Right after the `conf->env_vars_enabled` check

bpf/process/bpf_execve_event.c
```diff
@@ -126,6 +127,71 @@ read_args(void *ctx, struct msg_execve_event *event)
...
+
+	free_size = (char *)&event->process + BUFFER - envs;
+	envs_size = env_end - env_start;
+
+	if (envs_size < BUFFER && envs_size < free_size) {
```
FedeDP, line 165: Don't we need to set 
`flags |= EVENT_ENVS_DATA;`
in this branch too?

olsajiri: so now the envs data is marked by having `p->size_envs != 0`, the `EVENT_ENVS_DATA` bit is set just if the envs goes through the data event

but I did forget to put `api.EventErrorEnvs` in `FlagStrings` .. added ;-)

FedeDP: I see, it makes sense. Thanks!

FedeDP (approved):

bpf/process/bpf_execve_event.c
```diff
@@ -103,6 +103,8 @@ read_args(void *ctx, struct msg_execve_event *event)
 	args_size = end_stack - start_stack;
 
 	if (args_size < BUFFER && args_size < free_size) {
+		if (args_size)
+			args_size -= 1;
```
kkourt, line 108: question: I guess we are skipping the NULL here now that we have sizes?

pkg/sensors/exec/exec_linux.go
```diff
@@ -109,59 +110,76 @@ func execParse(reader *bytes.Reader) (processapi.MsgProcess, error) {
...
+		err := fmt.Errorf("msg exec size larger than argsbuffer, size %d != %d, SizePath %d, SizeArgs %d, SizeCwd %d",
+			size, exec.SizePath+exec.SizeArgs+exec.SizeCwd, exec.SizePath, exec.SizeArgs, exec.SizeCwd)
 		return proc, err
 	}
 
-	if exec.Flags&api.EventDataFilename != 0 {
-		var desc dataapi.DataEventDesc
-
-		dr := bytes.NewReader(args)
+	if exec.SizePath != 0 {
```
kkourt, line 131: Are there legitimate cases where this can be zero? (similarly for the other added sizes)

bpf/process/bpf_execve_event.c
```diff
@@ -126,6 +127,71 @@ read_args(void *ctx, struct msg_execve_event *event)
 	return size;
 }
 
+#ifdef __LARGE_BPF_PROG
+FUNC_INLINE __u32 read_envs(void *ctx, struct msg_execve_event *event)
```
kkourt, line 133: Will this work correctly for execs happening from inside the kernel (e.g, call_usrmodehelper and friends).

api/v1/tetragon/tetragon.proto
```diff
@@ -297,6 +297,8 @@ message Process {
   // for example, you wish to discern whether a process was spawned using a
   // tool like nsenter or kubectl exec.
   google.protobuf.BoolValue in_init_tree = 20;
+  // Environment variables passed to the binary at execution.
+  string environment_variables = 21;
```
kkourt: Question: Should we  parse this in user-space and offer an array to the user?
E.g., 

```
message EnvEntry {
   string Key = 1;
   string Value = 2;
}
```

And return:
```
    repeated EnvEntry environment_variables = 21
```

pkg/option/flags.go
```diff
@@ -390,6 +394,8 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.Bool(KeyEnablePodAnnotations, false, "Add pod annotations field to events.")
 	flags.StringSlice(KeyEnableAncestors, []string{}, "Comma-separated list of process event types to enable ancestors for. Supported event types are: base, kprobe, tracepoint, uprobe, lsm, usdt. Unknown event types will be ignored. Type 'base' enables ancestors for process_exec and process_exit events and is required by all other supported event types for correct reference counting. An empty string disables ancestors completely")
 
+	flags.Bool(KeyEnableProcessEnvironmentVariables, false, "Include environment variables in process_exec events. Disabled by default. Note that this option can significantly increase the size of the events and may impact performance")
+
```
kkourt: Let's add a note about sensitive information in the flag.

E.g., "Include environment variables in process_exec events. Disabled by default. Note that this option can significantly increase the size of the events and may impact performance, as well as capture sensitive information such as passwords in the events"

api/v1/tetragon/tetragon.proto
olsajiri: not sure.. as you already mentioned catching this impacts performance already,
also it's sort of similar to args string which we handle the same way together with the redaction support

bpf/process/bpf_execve_event.c
olsajiri: yes

pkg/sensors/exec/exec_linux.go
olsajiri: yes, if probe read fails on the envs data, then EVENT_ERROR_ARGS should be set

pkg/option/flags.go
olsajiri: ok

pkg/sensors/exec/exec_linux.go
kkourt: But we continue reading everything else if I understand correctly from the code?

api/v1/tetragon/tetragon.proto
kkourt: Yeah, but splitting a string in userspace (I'm guessing based on a ` `) character shouldn't have a big impact on performance, no?

bpf/process/bpf_execve_event.c
```diff
@@ -126,6 +127,71 @@ read_args(void *ctx, struct msg_execve_event *event)
 	return size;
 }
 
+#ifdef __LARGE_BPF_PROG
```
mtardy, line 130: if we are guarding this with `__LARGE_BPF_PROG` anyway, why can't we stop using `tetragon_conf` and use proper runtime constants https://ebpf-go.dev/concepts/global-variables/#runtime-constants ? (so that we benefit from dead code elimination etc.)

I think we wanted to switch existing code so that would be cool to not add more code using the legacy style if we can? wdyt?

bpf/process/bpf_execve_event.c
```diff
@@ -359,10 +359,9 @@ execve_send(void *ctx __arg_ctx)
 		curr->bin.args[len] = 0x00;
 		curr->bin.args[len + 1] = 0x00;
 #else
-		// reuse p->args first string that contains the filename, this can't be
-		// above 256 in size (otherwise the complete will be send via data msg)
-		// which is okay because we need the 256 first bytes.
-		curr->bin.path_length = probe_read_str(curr->bin.path, BINARY_PATH_MAX_LEN, &p->args);
+		char *filename = (char *)ctx + (_(ctx->__data_loc_filename) & 0xFFFF);
+
+		curr->bin.path_length = probe_read_str(curr->bin.path, BINARY_PATH_MAX_LEN, (void *)filename);
```
mtardy, line 442: ah I remember seeing this change in another patch set and couldn't get why doing that. So here I understand that it simplifies your changes from the next commit!

api/v1/tetragon/tetragon.proto
mtardy: Yeah I'm wondering too if it's not a nice UX improvement that is pretty cheap considering everything that we already do in this gRPC conversion. No strong feelings.

mtardy (commented):
Thanks, here are a few comments, I think using size based parsing is cleaner indeed and easier to understand than searching for those NULL bytes.

pkg/option/flags.go
mtardy: "as well as capture sensitive information such as passwords in the events" might be super verbose but we can also mention the feature we have for redacting such content here.

bpf/process/bpf_execve_event.c
olsajiri: yep, environment is stored the same way, so we get it.. like:

`{"process_exec":{"process":{"exec_id":"cWVtdS0xOjUyNzE3MDc3ODA3MjoxMjYxOA==","pid":12618,"uid":0,"cwd":"/","binary":"/usr/lib/systemd/systemd-coredump","arguments":"12617 1000 1000 11 1763675427 102400000000 qemu-1 1 3","flags":"execve rootcwd clone","start_time":"2025-11-20T21:50:27.329868374Z","auid":4294967295,"parent_exec_id":"cWVtdS0xOjI5NzAwMDAwMDA6Mg==","tid":12618,"in_init_tree":false,"environment_variables":"HOME=/ TERM=linux PATH=/sbin:/usr/sbin:/bin:/usr/bin"},"parent":{"exec_id":"cWVtdS0xOjI5NzAwMDAwMDA6Mg==","pid":2,"uid":0,"binary":"[kthreadd]","flags":"procFS","start_time":"2025-11-20T21:41:43.129089721Z","auid":4294967295,"parent_exec_id":"cWVtdS0xOjE6MA==","tid":2,"in_init_tree":false}},"node_name":"qemu-1","time":"2025-11-20T21:50:27.329867762Z"}
`

bpf/process/bpf_execve_event.c
olsajiri: probably good idea :) will check, thanks

api/v1/tetragon/tetragon.proto
olsajiri: ok, will add that

pkg/sensors/exec/exec_linux.go
olsajiri: yes

pkg/sensors/base/base.go
```diff
@@ -154,6 +154,10 @@ func setupSensor() {
 
 	logger.GetLogger().Info(fmt.Sprintf("Set execve_map entries %d", entries),
 		"size", strutils.SizeWithSuffix(entries*int(unsafe.Sizeof(execvemap.ExecveValue{}))))
+
+	if option.Config.EnableProcessEnvironmentVariables {
+		Execve.RewriteConstants["ENV_VARS_ENABLED"] = uint8(1)
```
mtardy, line 159: nice thanks

docs/content/en/docs/concepts/events.md
```diff
@@ -265,6 +265,18 @@ We can also redact these as follows:
...
+
+```json
+{"redact": ["(?:SSHPASS=)+(\\S+)"]}
+```
+
+Now, an event that contains the string `"SSHPASS=password"` would have that string
+replaced with `"SSHPASS=*****"`.
+
+It's also possible to store only requested environment variables with
+'--filter-environment-variables VAR1[,VAR2..]` option.
```
mtardy, line 278: nit: feel free to ignore
```suggestion
It's also possible to store only requested environment variables with the
'--filter-environment-variables VAR1[,VAR2..]` option.
```

Also this could be documented in the part that describes this feature, but if we don't have any, that's cool

bpf/process/bpf_execve_event.c
```diff
@@ -126,6 +127,71 @@ read_args(void *ctx, struct msg_execve_event *event)
...
+	if (envs >= (char *)&event->process + BUFFER)
+		return 0;
+
+	task = (struct task_struct *)get_current_task();
+	probe_read(&mm, sizeof(mm), _(&task->mm));
+	if (!mm)
+		return 0;
+
+	with_errmetrics(probe_read, &env_start, sizeof(env_start), _(&mm->env_start));
+	with_errmetrics(probe_read, &env_end, sizeof(env_end), _(&mm->env_end));
```
mtardy, line 157: cool to use errmetrics here

tpapagian (approved):

cmd/tetragon/main.go
```diff
@@ -230,6 +230,10 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 		logger.Fatal(log, "Can't specify --execve-map-entries and --execve-map-size together")
 	}
 
+	if option.Config.EnableProcessEnvironmentVariables && !config.EnableLargeProgs() {
+		logger.Fatal(log, "Can't specify --enable-process-environment-variables on small kernels (<v5.3)")
```
kkourt: non-blocking nit: on early kernels?

pkg/process/process.go
```diff
@@ -268,6 +268,26 @@ func GetExecIDFromKey(key *tetragonAPI.MsgExecveKey) string {
 	return GetProcessID(key.Pid, key.Ktime)
 }
 
+func getEnvironmentVariables(envs []string) []*tetragon.EnvVar {
+	res := []*tetragon.EnvVar{}
+
+	for _, v := range envs {
+		var key, val string
+
+		idx := strings.Index(v, "=")
```
kkourt, line 278: non blocking nit: would strings.Split be better?

olsajiri: nope, because I'm lookng for the first '=' and the variable value could have additional '='

cmd/tetragon/main.go
olsajiri: early sounds better ;-) thanks

pkg/process/process.go
kkourt: SplitN then? (https://pkg.go.dev/strings#SplitN)

olsajiri: hm, then in case of error (sep not found) I have to reorg the returned slice.. not sure it's easier

PR #4193, opened 2025-10-15
new: uretprobe support

### LINKED ISSUE #4118: Support for monitoring bash user input with `uretprobe` (function return value capture)

_No response_

adding support for `uretprobes` so that we can monitoring bash user input via `readline()` return values

I have been experimenting with tracing `bash` input using Tetragon policies.
For example, attaching an `uprobe` to `/bin/bash` with `symbol: readline`:

```yaml
kind: TracingPolicy
metadata:
  name: trace-bash-readline
spec:
  uprobes:
  - path: "/bin/bash"
    symbols:
    - "readline"
    args:
    - index: 0
      type: "string"
      returnCopy: true
```

What I observe is that the event only contains the **prompt string** (`PS1` with escape codes), not the actual user input typed (e.g. `echo hello`).

From my understanding of how GNU Readline works:

* `arg0` = the prompt string (what bash prints before waiting for input).
* **return value** of `readline()` = the actual line typed by the user.

Currently, Tetragon `UProbeSpec` only has `Args []KProbeArg` and `returnCopy: true` within `KProbeArg`, but it seems that `uretprobe` (function return value probes) is not supported, at least for uprobes.

---

**My question:**

* Is there a reason Tetragon does not support `uretprobe` for uprobes (similar to kretprobes)?
* Are there technical limitations (e.g. performance, stability, verifier issues) that prevent return-value capture for uprobes?
* Would you consider adding support for `uretprobes` so that monitoring bash user input via `readline()` return values becomes possible?

---

**Use case:**
Monitoring interactive shell input (commands as typed by the user) is valuable for security auditing and intrusion detection. At the moment, only the prompt string can be captured, which isn’t very useful.

### Code of Conduct

FedeDP:
Fixes #4118 

### Description

We previously missed `uretprobe` support. This PR adds support for it, commit by commit.
Basically, it adds support for:
```
return: true
returnArg
args.returnCopy
```

### Changelog

```release-note
Add uretprobe support for uprobes, like kprobes, which means supports for `return`, `returnArg` and `args.returnCopy`.
```

### DIFF
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -74,16 +74,7 @@ type genericUprobe struct {
 	// the map, so that we can merge them when the return event is
 	// generated. The events are maintained in the map below, using
 	// the retprobe_id (thread_id) and the enter ktime as the key.
-	pendingEvents *lru.Cache[pendingEventKey, uprobePendingEvent]
-}
-
-// pendingEvent is an event waiting to be merged with another event.
-// This is needed for retprobe probes that generate two events: one at the
-// function entry, and one at the function return. We merge these events into
-// one, before returning it to the user.
-type uprobePendingEvent struct {
-	ev          *tracing.MsgGenericUprobeUnix
-	returnEvent bool
+	pendingEvents *lru.Cache[pendingEventKey, pendingEvent[*tracing.MsgGenericUprobeUnix]]
 }
 
 func (g *genericUprobe) SetID(id idtable.EntryID) {
@@ -163,18 +154,13 @@ func handleGenericUprobe(r *bytes.Reader) ([]observer.Event, error) {
 	// Cache return value on merge and run return filters below before
 	// passing up to notify hooks.
 	if uprobeEntry.loadArgs.retprobe {
-		// if an event exist already, try to merge them. Otherwise, add
-		// the one we have in the map.
-		curr := uprobePendingEvent{ev: unix, returnEvent: returnEvent}
-		key := pendingEventKey{eventId: m.RetProbeId, ktimeEnter: ktimeEnter}
-
-		if prev, exists := uprobeEntry.pendingEvents.Get(key); exists {
-			uprobeEntry.pendingEvents.Remove(key)
-			unix = uretprobeMerge(prev, curr)
-		} else {
-			uprobeEntry.pendingEvents.Add(key, curr)
-			unix = nil
-		}
+		_, unix, _ = retprobeMergeEvents[*tracing.MsgGenericUprobeUnix](
+			unix,
+			uprobeEntry.pendingEvents,
+			returnEvent,
+			m.RetProbeId,
+			ktimeEnter,
+			reportUprobeMergeError)
 	}
 	if unix == nil {
 		return []observer.Event{}, err
@@ -655,7 +641,7 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 			pendingEvents:     nil,
 		}
 
-		uprobeEntry.pendingEvents, err = lru.New[pendingEventKey, uprobePendingEvent](4096)
+		uprobeEntry.pendingEvents, err = lru.New[pendingEventKey, pendingEvent[*tracing.MsgGenericUprobeUnix]](4096)
 		if err != nil {
 			return err
 		}
@@ -868,7 +854,7 @@ func createUprobeSensorFromEntry(uprobeEntry *genericUprobe,
 	return progs, maps
 }
 
-func reportUprobeMergeError(curr uprobePendingEvent, prev uprobePendingEvent) {
+func reportUprobeMergeError(curr pendingEvent[*tracing.MsgGenericUprobeUnix], prev pendingEvent[*tracing.MsgGenericUprobeUnix]) {
 	currSymbol := "UNKNOWN"
 	if curr.ev != nil {
 		currSymbol = curr.ev.Symbol
@@ -892,31 +878,4 @@ func reportUprobeMergeError(curr uprobePendingEvent, prev uprobePendingEvent) {
 		"currType", currType.String(),
 		"prevSymbol", prevSymbol,
 		"prevType", prevType.String())
-}
-
-// uretprobeMerge merges the two events: the one from the entry probe with the one from the return probe
-// TODO: find a way to merge this with kretprobeMerge
-func uretprobeMerge(prev uprobePendingEvent, curr uprobePendingEvent) *tracing.MsgGenericUprobeUnix {
-	var retEv, enterEv *tracing.MsgGenericUprobeUnix
-
-	if prev.returnEvent && !curr.returnEvent {
-		retEv = prev.ev
-		enterEv = curr.ev
-	} else if !prev.returnEvent && curr.returnEvent {
-		retEv = curr.ev
-		enterEv = prev.ev
-	} else {
-		reportUprobeMergeError(curr, prev)
-		return nil
-	}
-
-	for _, retArg := range retEv.Args {
-		index := retArg.GetIndex()
-		if uint64(len(enterEv.Args)) > index {
-			enterEv.Args[index] = retArg
-		} else {
-			enterEv.Args = append(enterEv.Args, retArg)
-		}
-	}
-	return enterEv
 }
--- a/pkg/grpc/tracing/tracing.go
+++ b/pkg/grpc/tracing/tracing.go
@@ -687,6 +687,10 @@ type MsgGenericKprobeUnix struct {
 	Tags             []string
 }
 
+func (msg *MsgGenericKprobeUnix) GetArgs() *[]tracingapi.MsgGenericKprobeArg {
+	return &msg.Args
+}
+
 func (msg *MsgGenericKprobeUnix) Notify() bool {
 	return true
 }
@@ -811,6 +815,10 @@ type MsgGenericUprobeUnix struct {
 	Message      string
 	Args         []tracingapi.MsgGenericKprobeArg
 	Tags         []string
+}
+
+func (msg *MsgGenericUprobeUnix) GetArgs() *[]tracingapi.MsgGenericKprobeArg {
+	return &msg.Args
 }
 
 func (msg *MsgGenericUprobeUnix) Notify() bool {
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -96,7 +96,7 @@ type genericKprobe struct {
 	// the map, so that we can merge them when the return event is
 	// generated. The events are maintained in the map below, using
 	// the retprobe_id (thread_id) and the enter ktime as the key.
-	pendingEvents *lru.Cache[pendingEventKey, pendingEvent]
+	pendingEvents *lru.Cache[pendingEventKey, pendingEvent[*tracing.MsgGenericKprobeUnix]]
 
 	tableId idtable.EntryID
 
@@ -132,8 +132,8 @@ type genericKprobe struct {
 // This is needed for retprobe probes that generate two events: one at the
 // function entry, and one at the function return. We merge these events into
 // one, before returning it to the user.
-type pendingEvent struct {
-	ev          *tracing.MsgGenericKprobeUnix
+type pendingEvent[T evArgsRetriever] struct {
+	ev          T
 	returnEvent bool
 }
 
@@ -962,7 +962,7 @@ func addKprobe(funcName string, instance int, f *v1alpha1.KProbeSpec, in *addKpr
 		}
 	}
 
-	kprobeEntry.pendingEvents, err = lru.New[pendingEventKey, pendingEvent](4096)
+	kprobeEntry.pendingEvents, err = lru.New[pendingEventKey, pendingEvent[*tracing.MsgGenericKprobeUnix]](4096)
 	if err != nil {
 		return errFn(err)
 	}
@@ -1375,18 +1375,22 @@ func handleMsgGenericKprobe(m *api.MsgGenericKprobe, gk *genericKprobe, r *bytes
 
 	// there are two events for this probe (entry and return)
 	if gk.loadArgs.retprobe {
-		// if an event exist already, try to merge them. Otherwise, add
-		// the one we have in the map.
-		curr := pendingEvent{ev: unix, returnEvent: returnEvent}
-		key := pendingEventKey{eventId: m.RetProbeId, ktimeEnter: ktimeEnter}
-
-		if prev, exists := gk.pendingEvents.Get(key); exists {
-			gk.pendingEvents.Remove(key)
-			unix = kretprobeMerge(prev, curr)
-		} else {
-			gk.pendingEvents.Add(key, curr)
+		var (
+			other  *tracing.MsgGenericKprobeUnix
+			merged bool
+		)
+		merged, unix, other = retprobeMergeEvents[*tracing.MsgGenericKprobeUnix](
+			unix,
+			gk.pendingEvents,
+			returnEvent,
+			m.RetProbeId,
+			ktimeEnter,
+			reportKprobeMergeError)
+		if unix != nil {
+			kprobemetrics.MergeOkTotalInc()
+			unix.ReturnAction = other.Msg.ActionId
+		} else if !merged {
 			kprobemetrics.MergePushedInc()
-			unix = nil
 		}
 	}
 	if unix == nil {
@@ -1402,7 +1406,7 @@ func handleMsgGenericKprobe(m *api.MsgGenericKprobe, gk *genericKprobe, r *bytes
 	return []observer.Event{unix}, err
 }
 
-func reportMergeError(curr pendingEvent, prev pendingEvent) {
+func reportKprobeMergeError(curr pendingEvent[*tracing.MsgGenericKprobeUnix], prev pendingEvent[*tracing.MsgGenericKprobeUnix]) {
 	currFn := "UNKNOWN"
 	if curr.ev != nil {
 		currFn = curr.ev.FuncName
@@ -1429,9 +1433,18 @@ func reportMergeError(curr pendingEvent, prev pendingEvent) {
 		"prevType", prevType.String())
 }
 
-// kretprobeMerge merges the two events: the one from the entry probe with the one from the return probe
-func kretprobeMerge(prev pendingEvent, curr pendingEvent) *tracing.MsgGenericKprobeUnix {
-	var retEv, enterEv *tracing.MsgGenericKprobeUnix
+type reportMergeErorrFn[T evArgsRetriever] func(curr pendingEvent[T], prev pendingEvent[T])
+
+type evArgsRetriever interface {
+	GetArgs() *[]api.MsgGenericKprobeArg
+	// This constraint allows us to return nil from methods
+	*tracing.MsgGenericKprobeUnix | *tracing.MsgGenericUprobeUnix
+}
+
+// retprobeMerge merges the two events: the one from the entry probe with the one from the return probe
+func retprobeMerge[T evArgsRetriever](prev pendingEvent[T], curr pendingEvent[T],
+	onMergeError reportMergeErorrFn[T]) (T, T) {
+	var retEv, enterEv T
 
 	if prev.returnEvent && !curr.returnEvent {
 		retEv = prev.ev
@@ -1440,22 +1453,37 @@ func kretprobeMerge(prev pendingEvent, curr pendingEvent) *tracing.MsgGenericKpr
 		retEv = curr.ev
 		enterEv = prev.ev
 	} else {
-		reportMergeError(curr, prev)
-		return nil
+		onMergeError(curr, prev)
+		return nil, nil
 	}
 
-	kprobemetrics.MergeOkTotalInc()
-
-	for _, retArg := range retEv.Args {
+	retArgs := retEv.GetArgs()
+	enterArgs := enterEv.GetArgs()
+	for _, retArg := range *retArgs {
 		index := retArg.GetIndex()
-		if uint64(len(enterEv.Args)) > index {
-			enterEv.Args[index] = retArg
+		if uint64(len(*enterArgs)) > index {
+			(*enterArgs)[index] = retArg
 		} else {
-			enterEv.Args = append(enterEv.Args, retArg)
+			*enterArgs = append(*enterArgs, retArg)
 		}
 	}
-	enterEv.ReturnAction = retEv.Msg.ActionId
-	return enterEv
+	return enterEv, retEv
+}
+
+func retprobeMergeEvents[T evArgsRetriever](unix T, pendingEvents *lru.Cache[pendingEventKey, pendingEvent[T]],
+	returnEvent bool, retprobeId, ktimeEnter uint64, onMergeError reportMergeErorrFn[T]) (bool, T, T) {
+	// if an event exist already, try to merge them. Otherwise, add
+	// the one we have in the map.
+	curr := pendingEvent[T]{ev: unix, returnEvent: returnEvent}
+	key := pendingEventKey{eventId: retprobeId, ktimeEnter: ktimeEnter}
+
+	if prev, exists := pendingEvents.Get(key); exists {
+		pendingEvents.Remove(key)
+		enter, exit := retprobeMerge[T](prev, curr, onMergeError)
+		return true, enter, exit
+	}
+	pendingEvents.Add(key, curr)
+	return false, nil, nil
 }
 
 func (k *observerKprobeSensor) LoadProbe(args sensors.LoadProbeArgs) error {


FedeDP:
Will resolve rebase conflict and eventual CI issues asap :)

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -142,6 +160,26 @@ func handleGenericUprobe(r *bytes.Reader) ([]observer.Event, error) {
 		unix.Args = append(unix.Args, arg)
 	}
 
+	// Cache return value on merge and run return filters below before
+	// passing up to notify hooks.
+	if uprobeEntry.loadArgs.retprobe {
```
FedeDP, line 156: This is exact same code as generickprobe.go.
Couldn't find a way to make it generic enough (without making weird dances with interfaces and go generics that were at the end larger than current code) to reuse same code in both.

bpf/process/bpf_generic_retuprobe.c
```diff
@@ -0,0 +1,71 @@
...
+#define GENERIC_URETPROBE
+
+#include "compiler.h"
+#include "bpf_tracing.h"
+#include "bpf_event.h"
+#include "bpf_task.h"
+#include "retprobe_map.h"
+#include "types/basic.h"
+
+#define MAX_FILENAME 8096
```
olsajiri: not needed, please remove it also from bpf_generic_retkprobe.c

bpf/process/generic_calls.h
```diff
@@ -1206,6 +1213,88 @@ FUNC_INLINE int generic_retkprobe(void *ctx, struct bpf_map_def *calls, unsigned
 	return 1;
 }
 
+FUNC_INLINE int generic_retuprobe(void *ctx, struct bpf_map_def *calls, unsigned long ret)
```
olsajiri: I think this could be shared by both retkprobe/retuprobe as generic_retprobe with few ifdefs noted below

the socket related code would be ifdef GENERIC_KRETPROBE

bpf/process/generic_calls.h
```diff
@@ -1206,6 +1213,88 @@ FUNC_INLINE int generic_retkprobe(void *ctx, struct bpf_map_def *calls, unsigned
...
+	case char_iovec:
+		size += __copy_char_iovec(size, info.ptr, info.cnt, ret, e);
+	default:
+		break;
+	}
+
+	/* Complete message header and send */
+	enter = event_find_curr(&ppid, &walker);
+
+	e->common.op = MSG_OP_GENERIC_UPROBE;
```
olsajiri, line 1204: this would have special block for GENERIC_KRETPROBE and for GENERIC_URETPROBE

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -442,50 +543,115 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+		uprobeEntry.pendingEvents, err = lru.New[pendingEventKey, uprobePendingEvent](4096)
+		if err != nil {
+			return err
 		}
 
 		uprobeTable.AddEntry(uprobeEntry)
 		id := uprobeEntry.tableId
 
-		config.FuncId = uint32(id.ID)
+		eventConfig.FuncId = uint32(id.ID)
```
olsajiri, line 652: why not keep it as config, would be less changes

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -43,22 +47,43 @@ var (
...
 )
 
+type uprobeLoadArgs struct {
+	selectors kprobeSelectors
+	retprobe  bool
+	config    *api.EventConfig
+}
+
 type genericUprobe struct {
+	loadArgs     uprobeLoadArgs
```
olsajiri, line 57: this change seems like it could be split into several changes so it's less complicated, one of them might be to introduce uprobeLoadArgs in separate change without functional changes?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -558,25 +769,15 @@ func createSingleUprobeSensor(ids []idtable.EntryID) ([]*program.Program, []*pro
...
-	attachData := &program.UprobeAttachData{
-		Path:         uprobeEntry.path,
-		Symbol:       symbol,
-		Offset:       offset,
-		Address:      uprobeEntry.address,
-		RefCtrOffset: uprobeEntry.refCtrOffset,
-	}
+	loadProgName, loadProgRetName := config.GenericUprobeObjs(false)
 
+	pinProg := fmt.Sprintf("%d-%s", uprobeEntry.tableId.ID, uprobeEntry.symbol)
```
olsajiri: let's not complicate the change, this is not really needed, is it?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -558,25 +769,15 @@ func createSingleUprobeSensor(ids []idtable.EntryID) ([]*program.Program, []*pro
...
 	load := program.Builder(
 		path.Join(option.Config.HubbleLib, loadProgName),
 		fmt.Sprintf("%s %s", uprobeEntry.path, uprobeEntry.symbol),
 		"uprobe/generic_uprobe",
-		fmt.Sprintf("%d-%s", uprobeEntry.tableId.ID, uprobeEntry.symbol),
+		pinProg,
 		"generic_uprobe").
-		SetAttachData(attachData).
```
olsajiri, line 625: I dont understand why is this moved, would be great to have some explanation in changelog, in all of them ;-)

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -585,7 +786,87 @@ func createUprobeSensorFromEntry(uprobeEntry *genericUprobe,
...
+
+	logger.GetLogger().Debug("failed to merge events",
+		"currSymbol", currSymbol,
+		"currType", currType.String(),
+		"prevSymbol", prevSymbol,
+		"prevType", prevType.String())
+}
+
+// uretprobeMerge merges the two events: the one from the entry probe with the one from the return probe
+func uretprobeMerge(prev uprobePendingEvent, curr uprobePendingEvent) *tracing.MsgGenericUprobeUnix {
```
olsajiri: I think this could be shared with kprobe no? perhaps with some func callback for retrieve specific probe data? not sure..  would be great to share this, but we can always do it as follow up if it's too complicated

contrib/tester-progs/uretprobe.c
```diff
@@ -0,0 +1,28 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+char *return_string(char *param) {
```
olsajiri: please keep the same kernel source style as we do for the rest of the test programs
you also mixed tabs and spaces

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -442,50 +543,115 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 		argPrinters = append(argPrinters, argPrinter{index: i, ty: argType})
 	}
 
-	addUprobeEntry := func(sym string, offset uint64, idx int) {
+	eventConfig := initEventConfig()
+
+	// Parse ReturnArg, we have two types of return arg parsing. We
+	// support populating a kprobe buffer from kretprobe hooks. This
```
olsajiri: kprobe/kretprobe ?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -442,50 +543,115 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+	if argRetprobe != nil {
+		setRetprobe = true
+
+		argType := gt.GenericTypeFromString(argRetprobe.Type)
+		eventConfig.ArgReturnCopy = int32(argType)
+
+		argP := argPrinter{index: int(argRetprobe.Index), ty: argType, label: argRetprobe.Label}
+		argReturnPrinters = append(argReturnPrinters, argP)
+	} else {
+		eventConfig.ArgReturnCopy = int32(0)
```
olsajiri: hm, perhaps we should add GenericUnsetType/unset_ty or such

pkg/sensors/tracing/uprobe_test.go
```diff
@@ -196,6 +198,118 @@ spec:
...
+
+	if err := exec.Command(testUretprobe).Run(); err != nil {
+		t.Fatalf("Failed to execute test binary: %s\n", err)
+	}
+
+	err = jsonchecker.JsonTestCheck(t, checker)
+	require.NoError(t, err)
+}
+
+func TestUretprobeRetCopy(t *testing.T) {
```
olsajiri, line 256: could you please also add tests for return selector and action?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -499,22 +665,36 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+			}
 		}
 	}
 
 	return ids, nil
 }
 
 func multiUprobePinPath(sensorPath string) string {
-	return sensors.PathJoin(sensorPath, "multi_kprobe")
+	return sensors.PathJoin(sensorPath, "multi_uprobe")
```
olsajiri, line 698: ugh, please send this in separate change, thanks

pkg/sensors/tracing/genericuprobe.go
FedeDP: Yep i didn't want to over-complicate it since the PR is already big enough.
I can leave a TODO if you wish!

pkg/sensors/tracing/genericuprobe.go
FedeDP: Makes sense.

pkg/sensors/tracing/uprobe_test.go
FedeDP: `returnArgAction` is not supported by uretprobes (at least, not in this PR).
Also, uprobe only supports matchPIDs selector: https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/genericuprobe.go#L271

pkg/sensors/tracing/genericuprobe.go
FedeDP: Or, if you don't mind over-complicating the PR, i can work on it now :)

pkg/sensors/tracing/genericuprobe.go
FedeDP: Split the creation of `uprobeLoadArgs` to its own commit.

FedeDP:
Seeing https://github.com/cilium/tetragon/pull/4207 changes, i also ported `InitKernelReturnSelectorState` to use the newly introduced `KernelSelectorArgs` struct.

pkg/sensors/tracing/uprobe_test.go
FedeDP: (Also, since `returnArgAction` seems to only be supported by kretprobes with only `ACTION_TRACKSOCK` and `ACTION_UNTRACKSOCK` values, i don't think it is worth to support it in uretprobes)

FedeDP:
Rebased on top of latest main.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -759,7 +759,7 @@ func createMultiUprobeSensor(sensorPath string, multiIDs []idtable.EntryID, poli
 		regsMap := program.MapBuilderProgram("regs_map", load)
 		sleepableOffloadMap := program.MapBuilderProgram("sleepable_offload", load)
 		sleepableOffloadMap.SetMaxEntries(sleepableOffloadMaxEntries)
-		maps = append(maps, regsMap)
+		maps = append(maps, regsMap, sleepableOffloadMap)
```
FedeDP, line 744: @olsajiri seems like in #4173 `sleepableOffloadMap` was missing for multi uprobe sensors.
I added the fix in its own commit (ready to drop it in case it was a desired behavior!)

olsajiri: ah nice.. could you please push separate fix for uprobe_multi part that goes before your changes? the change for uretprobe below would not be necessary, thanks

FedeDP: Do you mean opening a new PR with the 
> 		maps = append(maps, regsMap, sleepableOffloadMap)

patch?

FedeDP: Oh you meant this one: https://github.com/cilium/tetragon/pull/4193/commits/847a83572859845b9f229809fa678fabd913e277?

FedeDP: :point_up:  done here: https://github.com/cilium/tetragon/pull/4225

FedeDP: and https://github.com/cilium/tetragon/pull/4226 for the `sleepableOffloadMap` one.

FedeDP:
Rebased on top of latest main fixes.

pkg/sensors/tracing/genericuprobe.go
FedeDP: > the change for uretprobe below would not be necessary, thanks

Do you mean that i should not support `sleepable_offload` for uretprobe?

FedeDP: Yep, private chat confirmed :point_up: .
Dropped support for `SleepableOffload` in retuprobe.

FedeDP:
@olsajiri in the latest commit, i tried a solution to unify `uretprobeMerge` and `kretprobeMerge` helpers. Let me know WDYT :)

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -203,6 +267,7 @@ func loadMultiUprobeSensor(ids []idtable.EntryID, args sensors.LoadProbeArgs) er
 	data := &program.MultiUprobeAttachData{}
 	data.Attach = make(map[string]*program.MultiUprobeAttachSymbolsCookies)
 
+	binBuf := make([]bytes.Buffer, len(ids))
```
olsajiri: I dont understand the point of this change.. why to have slice instead of using configData in each iteration?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -139,26 +183,46 @@ func loadSingleUprobeSensor(uprobeEntry *genericUprobe, args sensors.LoadProbeAr
...
+		mapLoad = append(mapLoad, &program.MapLoad{
 			Name: "filter_map",
 			Load: func(m *ebpf.Map, _ string) error {
-				return m.Update(uint32(0), selBuff[:], ebpf.UpdateAny)
+				return m.Update(uint32(0), selector.Buffer(), ebpf.UpdateAny)
 			},
-		},
+		})
+
+		if load.SleepableOffload {
```
olsajiri, line 203: sleepable offload was is tested on entry probe so far.. I haven't thought about using that on uretprobe, let's keep it just for entry probe for now

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -211,39 +276,46 @@ func loadMultiUprobeSensor(ids []idtable.EntryID, args sensors.LoadProbeArgs) er
...
-
-		if load.SleepableOffload {
-			load.MapLoad = append(load.MapLoad,
-				&program.MapLoad{
-					Name: "regs_map",
-					Load: func(m *ebpf.Map, _ string) error {
-						return populateUprobeRegs(m, uprobeEntry.selectors.Regs())
+			})
+
+			if load.SleepableOffload {
```
olsajiri, line 305: ditto for sleepable offload just for entry uprobes

install/kubernetes/tetragon/crds-yaml/cilium.io_tracingpolicies.yaml
```diff
@@ -2740,6 +2740,12 @@ spec:
                     returnArg:
                       description: A return argument to include in the trace output.
                       properties:
+                        btfType:
```
olsajiri, line 2743: should this be part of commit:
new(pkg,install,vendor): added `return` and `returnArg` to UprobeSpec. 
?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -479,50 +566,115 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+		eventConfig.FuncId = uint32(id.ID)
 
 		ids = append(ids, id)
+		return nil
 	}
 
 	if symbols != 0 {
 		for idx, sym := range spec.Symbols {
-			if err := checkSymbol(sym); err != nil {
+			if err = checkSymbol(sym); err != nil {
```
olsajiri: why is this needed? there's other places like that you did not change.. I think this could stay?

olsajiri (commented):
looks good, I saw u added another commit, so I'll wait for that

install/kubernetes/tetragon/crds-yaml/cilium.io_tracingpolicies.yaml
FedeDP: You're right!! Thanks

pkg/sensors/tracing/genericuprobe.go
FedeDP: Yup here `load.SleepableOffload` will be true only for entry probes; same below.

pkg/sensors/tracing/genericuprobe.go
FedeDP: Right, let me revert that bit!

pkg/sensors/tracing/genericuprobe.go
FedeDP: Let me revert this one; it was to make it more similar to generickprobe, but it's now worth it considering how big this PR already is.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -479,50 +566,115 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+		argType := gt.GenericTypeFromString(argRetprobe.Type)
+		eventConfig.ArgReturnCopy = int32(argType)
+
+		argP := argPrinter{index: int(argRetprobe.Index), ty: argType, label: argRetprobe.Label}
+		argReturnPrinters = append(argReturnPrinters, argP)
+	} else {
+		eventConfig.ArgReturnCopy = int32(gt.GenericUnsetType)
+	}
+
+	// Write attributes into BTF ptr for use with load
```
olsajiri: nit, not sure how that comment is related to the code below,
also we could do just `setRetprobe := spec.Return` before argRetprobe check

bpf/process/bpf_generic_retuprobe.c
```diff
@@ -13,8 +13,6 @@
 #include "retprobe_map.h"
 #include "types/basic.h"
 
-#define MAX_FILENAME 8096
```
olsajiri: nit, you didn't need to add it in the first place in the earlier commit

FedeDP: Right! Let me fix it, and sorry for the confusion!

pkg/sensors/tracing/genericuprobe.go
FedeDP: I like it!

mtardy:
> also I think it's good to have some description in each changelogs to ease up the future headaches

yep agree, I know it can look like repetition in the context of the PR but having the maximum of independent "why" in the commit descriptions for each change will help later on. Of course you can skip that for "chore" and generated stuff. Don't need to make up long sentences if not needed but having a bit of more context on why is always nice.

FedeDP:
I will update the PR changelog accordingly, thanks!

FedeDP:
It is updated; let me know whether it works fine or if you want more in-depth descriptions!

mtardy:
ah! we are not speaking about the changelog but the commit descriptions of each commits. I think Jiri said changelog and but meant commit description.

FedeDP:
Ah i see, let me go through all of them; i though i had been pretty precise in the commit messages :laughing: 

EDIT: ok eg: [this](https://github.com/cilium/tetragon/pull/4193/commits/e4ef1d491aa0a89e058942c5df846dfb13ea447a) needs some rewordings! Will push updated commits soon :tm:

FedeDP:
I force-pushed my reworded commits; PTAL! Thanks!

olsajiri:
> I force-pushed my reworded commits; PTAL! Thanks!

looks good, thanks

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -608,6 +646,12 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 			argReturnPrinters: argReturnPrinters,
 			retprobe:          setRetprobe,
 			tags:              tagsField,
+			pendingEvents:     nil,
+		}
+
+		uprobeEntry.pendingEvents, err = lru.New[pendingEventKey, uprobePendingEvent](4096)
```
mtardy: sorry I didn't check in detail, is this how we do it for kprobes?

I'm just thinking of these cache because I just fixed memory leaks, could it be that one of the events get lost and this LRU end up being filled and never cleaned? Do we need 4096 entries to work in normal situations? Should we be able to tunes this value (maybe with the kprobe one along in a first place?)

pkg/sensors/tracing/uprobe_test.go
```diff
@@ -196,6 +198,118 @@ spec:
...
+
+	obs, err := observertesthelper.GetDefaultObserverWithFile(t, ctx, testConfigFile, tus.Conf().TetragonLib, observertesthelper.WithMyPid())
+	if err != nil {
+		t.Fatalf("GetDefaultObserverWithFile error: %s", err)
+	}
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	if err := exec.Command(testUretprobe).Run(); err != nil {
+		t.Fatalf("Failed to execute test binary: %s\n", err)
```
mtardy: fyi: the tests from which you copy pasted this are having issues in vmtests, I think because of panic/or fatal, not sure. you might want to lookup the vmtests results for 5.4 for example even if it's green and switch to more traditional `require.NoError` or `t.Error`.

pkg/k8s/apis/cilium.io/v1alpha1/version.go
```diff
@@ -7,4 +7,4 @@ package v1alpha1
 // Used to determine if CRD needs to be updated in cluster
 //
 // Developers: Bump patch for each change in the CRD schema.
-const CustomResourceDefinitionSchemaVersion = "1.7.0"
+const CustomResourceDefinitionSchemaVersion = "1.7.1"
```
mtardy, line 10: can you move this commit just after your CRD change patch, or squash it with it if you want as well.

pkg/sensors/tracing/genericuprobe.go
FedeDP: Yep this is copy pasted from the kprobe code. I understand your sentiment; i think we can rethink it later on (going to take a note about it to avoid forgetting it)!
Don't want to pollute this already-big PR with more changes :)

mtardy (approved):
Let's just check what vmtests says (even if it's green) for the new test but otherwise let's merge this :)

pkg/sensors/tracing/genericuprobe.go
FedeDP: Since we expect to merge events that share the same `RetProbeId`, and given that `retprobe_id` is set by bpf to `get_current_pid_tgid` (https://github.com/cilium/tetragon/blob/9d68fba8c4ce37e0b0298f4148e661b93d90b7d0/bpf/process/retprobe_map.h#L120), `4096` is a generic "number of threads in the system performing our uprobe hook point".

Perhaps we could lower it; maybe @olsajiri has more insights

FedeDP:
Failed:
> ❌	pkg.bugtool.TestFindMaps													(total:3 failed:2 skipped:0)	1.406979587s	6.808598856s
├─❌	TestFindMaps/BaseSensorMemlock													960ms				(960ms)
└─❌	TestFindMaps															960ms

pkg/sensors/tracing/genericuprobe.go
olsajiri: not sure how we ended up with 4096, but perhaps let's lower it and make it configurable in separate pr ;-) thanks

mtardy:
> Failed:
> 
> > ❌	pkg.bugtool.TestFindMaps													(total:3 failed:2 skipped:0)	1.406979587s	6.808598856s
> > ├─✅	TestFindMaps/NoSuchFile														0s				(0s)
> > ├─❌	TestFindMaps/BaseSensorMemlock													960ms				(960ms)
> > └─❌	TestFindMaps															960ms

yeah... https://github.com/cilium/tetragon/issues/4220

pkg/sensors/tracing/genericuprobe.go
FedeDP: Of course i'll do!

PR #4194, opened 2025-10-16
tracepoint: support (u)int8/16 args

andrewstrohman:
On top of #4171 . Once that PR is merged, I will rebase this one over main so that only the two top most commits in the series remain.

### Description
This allows us to support (u)int8/16 args for tracepoints.

For example, 
```
kind: TracingPolicy
metadata:
  name: "test"
spec:
  tracepoints:
  - subsystem: "tcp"
    event: "tcp_destroy_sock"
    args:
    - index: 5
      type: "uint16"
    - index: 6
      type: "uint16"
```
 works with this patch.

### DIFF
--- a/pkg/sensors/tracing/tracepoint_test.go
+++ b/pkg/sensors/tracing/tracepoint_test.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"net"
 	"os"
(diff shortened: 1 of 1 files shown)


andrewstrohman:
I'm closing this because i've consolidated the changes in https://github.com/cilium/tetragon/pull/4171

PR #4196, opened 2025-10-16
Dockerfile.clang: upgrade to clang-20

olsajiri:


olsajiri:
note arm CI is failing with flaky TestKprobeMatchArgsFdPrefix test

mtardy:
Thanks for the update, looks good and it's great we have no patch to do at all, it seems there's a just a small CI issue, not sure it's important but would be nice to fix https://github.com/cilium/tetragon/actions/runs/18563370292/job/52922499566.

Issue #4176, opened 2025-10-13
tests: `TestKprobeMatchArgsFdEqual` is flaky


mtardy:
https://github.com/cilium/tetragon/actions/runs/18402760356/job/52435626625?pr=4159

mtardy:
Also https://github.com/cilium/tetragon/actions/runs/18521420319/job/52782001240 `TestKprobeMatchArgsFdPrefix`

kkourt:
All failures seem to be in "ubuntu-24.04-arm64"

kkourt:
```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 24.04.3 LTS
Release:        24.04
Codename:       noble
$ uname -m
aarch64
```

```
$ go test -tags sudo_tests -exec sudo -test.run TestKprobeMatchArgsFdEqual  ./pkg/sensors/tracing/   -count 20
ok      github.com/cilium/tetragon/pkg/sensors/tracing  90.114s
```

🤔

kkourt:
Let's reopen it if we see a new one.

PR #4198, opened 2025-10-17
usdt: support for `resolve:`

kkourt:
Support using the `resolve:` attribute in USDT policies.

```release-note
policies: support for resolve: in USDT policies
```

### DIFF
--- a/contrib/tester-progs/usdt-resolve.c
+++ b/contrib/tester-progs/usdt-resolve.c
@@ -2,25 +2,62 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
+#include <inttypes.h>
 
 #include "usdt.h"
 
-struct mystruct {
-	unsigned long a, b, c;
+struct mysubstruct {
+	uint64_t v64;
+	uint8_t  v8;
+	uint32_t v32;
 };
+
+struct mystruct {
+	uint8_t  v8;
+	uint16_t v16;
+	uint32_t v32;
+	uint64_t v64;
+	struct mysubstruct sub;
+};
+
+void usage(char *argv0)
+{
+	fprintf(stderr, "Usage: %s <field> <val>\n", argv0);
+	fprintf(stderr, "field can be one of: v8, v16, v32, v64, sub.v32\n");
+}
 
 int main(int argc, char *argv[])
 {
 	volatile int ret = 0;
-	if (argc < 2) {
-		fprintf(stderr, "Usage: %s <int>\n", argv[0]);
+	if (argc < 3) {
+		usage(argv[0]);
 		exit(1);
 	}
-	struct mystruct s = {
-		.a = rand(),
-		.b = atoi(argv[1]),
-		.c = rand(),
-	};
+
+	char *field = argv[1];
+	long val = atol(argv[2]);
+	if (!val) {
+		usage(argv[0]);
+		exit(1);
+	}
+
+	struct mystruct s = {0};
+	if (!strcmp(field, "v8")) {
+		s.v8 = val;
+	} else if (!strcmp(field, "v16")) {
+		s.v16 = val;
+	} else if (!strcmp(field, "v32")) {
+		s.v32 = val;
+	} else if (!strcmp(field, "v64")) {
+		s.v64 = val;
+	} else if (!strcmp(field, "sub.v32")) {
+		s.sub.v32 = val;
+	} else {
+		usage(argv[0]);
+		exit(1);
+	}
+
 	USDT(tetragon, test, ret, &s);
 	printf("ret=%d\n", ret);
 	return ret;
--- a/pkg/sensors/tracing/usdt_amd64_test.go
+++ b/pkg/sensors/tracing/usdt_amd64_test.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"os"
 	"os/exec"
+	"strconv"
 	"sync"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


kkourt:
Failures:

vmtests, 6.1:
```
 ❌	pkg.bugtool.TestFindMaps													(total:3 failed:2 skipped:0)	1.049115874s	6.638355964s
├─❌	TestFindMaps/BaseSensorMemlock													550ms				(550ms)
└─❌	TestFindMaps		
```

And `TestKprobeMatchArgsFdPrefix`  https://github.com/cilium/tetragon/issues/4176.

Restarting both.

pkg/bpf/detect_linux.go
```diff
@@ -536,14 +536,18 @@ func HasAuditLoginuid() bool {
 	return auditLoginuid.detected
 }
 
+func HasProbeWriteUser() bool {
+	return features.HaveProgramHelper(ebpf.Kprobe, asm.FnProbeWriteUser) == nil
```
olsajiri: do we really needs this? it was added in 4.8

but maybe for rhel7? ;-) let's call it HasProbeWriteUserHelper and move it up so it's together with HasSignalHelper and others?

pkg/sensors/tracing/genericusdt.go
```diff
@@ -324,20 +325,35 @@ func addUsdt(spec *v1alpha1.UsdtSpec, in *addUsdtIn, ids []idtable.EntryID) ([]i
...
+					spec.BTFPath,
+					arg.Resolve,
+				)
+				if err != nil {
+					return nil, err
+				}
+
+				allBTFArgs[cfgIdx] = btfArg
+				argType = findTypeFromBTFType(&arg, lastBTFType)
+
```
olsajiri: nit extra line

pkg/sensors/tracing/generic.go
```diff
@@ -33,6 +33,30 @@ func hasCurrentTaskSource(arg *v1alpha1.KProbeArg) bool {
 	return arg.Source == "current_task"
 }
 
+func resolveUserBTFArg(tyName string, btfFile string, resolvePath string) (*ebtf.Type, [api.MaxBTFArgDepth]api.ConfigBTFArg, error) {
```
olsajiri: could we pass 'arg' directly same as we do for resolveBTFArg? then the bottom of the function could be shared (from strings.Split) by both resolveUserBTFArg and resolveBTFArg

let's use btfPath instead of btfFile? it comes from spec.BTFPath

contrib/tester-progs/usdt-resolve.c
```diff
@@ -0,0 +1,27 @@
+//go:build ignore
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "usdt.h"
+
+struct mystruct {
+	unsigned long a, b, c;
```
olsajiri: could you add more types? could be separate test that parse out values from POD types? or at least 5.. which I guess is the max for the number of usdt arguments

olsajiri (commented):
looks good, left few comments, thanks

pkg/bpf/detect_linux.go
kkourt: I would say we needed not only for older kernel versions, but also for kernels that do not allow this due to lockdown:
```Lockdown: tetragon: use of bpf to write user RAM is restricted; see man kernel_lockdown.7```

> let's call it HasProbeWriteUserHelper and move it up so it's together with HasSignalHelper and others?

ack

File: pkg/sensors/tracing/usdt_amd64_test.go [hunk 5]
```diff
@@ -354,3 +358,94 @@ spec:
...
+        operator: "Equal"
+        values:
+        - "10"
+      matchActions:
+      - action: Set
+        argIndex: 0
+        argValue: 120
+`
+
+	t.Logf("write: %t", bpf.HasProbeWriteUser())
```
mtardy: maybe drop this line since you have t.Skip on it anyway a few lines before

pkg/sensors/tracing/usdt_amd64_test.go
```diff
@@ -354,3 +358,94 @@ spec:
...
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	// if the argument is 10, the program should fail
+	cmd := exec.Command(usdt, "10")
+	cmdErr := testutils.RunCmdAndLogOutput(t, cmd)
+	require.Error(t, cmdErr)
+	require.Equal(t, 120, cmd.ProcessState.ExitCode())
+
+	// if hte argument is not 10, then the program should succeed
```
mtardy: small typo

pkg/sensors/tracing/generic.go
kkourt: done, thanks.

contrib/tester-progs/usdt-resolve.c
kkourt: Thanks! I added one for u32, but I think we need https://github.com/cilium/tetragon/pull/4171 for other types. I also added a test for a nested struct.

pkg/sensors/tracing/usdt_amd64_test.go
kkourt: fixed, thanks!

pkg/sensors/tracing/usdt_amd64_test.go
kkourt: fixed!

kkourt:
Failures are due to [quay issues](https://status.redhat.com/incidents/24r8xgknfd4z), merging.

PR #4199, opened 2025-10-17
docs(assets/scss): Remove useless horizontal scrollbar

u-damien:
### Description
Added `overflow-x: hidden;` style to the `body` element to remove the useless horizontal scrollbar.

Before patch:
image

After patch:
image

### DIFF
--- a/docs/assets/scss/_styles_project.scss
+++ b/docs/assets/scss/_styles_project.scss
@@ -1,3 +1,4 @@
+@import "body";
 @import "button";
 @import "community-banner";
 @import "home";
--- a/docs/assets/scss/body.scss
+++ b/docs/assets/scss/body.scss
@@ -1,0 +1,3 @@
+body {
+  overflow-x: hidden;
+}


mtardy (changes requested):
Hey thanks for this, while I agree with this change, I think it would be better to fix the cause of this instead of just masking the symptoms 

image

image

To what I understand it's the `Get hands-on with Tetragon` that is too wide and is getting extra space for some reason, would you like to take a look by any chance? :)

u-damien:
Hello,

I saw that some elements width with 100vw overflow the parent element, I think that's the problem. I might take a look in the following days.
For example, in https://tetragon.io/docs/getting-started/ if you change the footer width from 100vw to 100%, it will remove the scrollbar. I tried in the others website paths (didn't worked only with footer) but didn't had the time to check deeper.

mtardy:
> Hello,
> 
> I saw that some elements width with 100vw overflow the parent element, I think that's the problem. I might take a look in the following days. For example, in https://tetragon.io/docs/getting-started/ if you change the footer width from 100vw to 100%, it will remove the scrollbar. I tried in the others website paths (didn't worked only with footer) but didn't had the time to check deeper.

If you want to look it up that would be amazing otherwise we can always open an issue for it later on. No pressure.

u-damien:
I will take a look maybe this week or the next week, yes!

kkourt:
Moving this to draft since it has ben idle for some time now.

PR #4203, opened 2025-10-17
new(pkg/elf): add usdt_args_arm64 implementation

### LINKED ISSUE #4192: USDT argument parsing for arm64

No

USDT argument parsing for arm64 has not been implemented yet. This request is to implement it.

```
astrohma@lima-default:~/src/tetragon$ ls pkg/elf/
file.go  usdt_args.go        usdt_args_amd64_test.go  usdt_notsupp.go
usdt.go  usdt_args_amd64.go  usdt_linux.go
```
We need a usdt_args_arm64.go source file to implement the arg parsing for arm64.

### Code of Conduct

FedeDP:
Fixes #4192

### Description

Also, since we now support multiple architectures, refactored the package a bit to allow easier integration.

Reference: https://github.com/iovisor/bcc/blob/b63d7e38e8a0f6339fbd57f3a1ae7297e1993d92/src/cc/usdt/usdt_args.cc#L224

### Changelog

```release-note
new(pkg/elf): add usdt_args_arm64 impl.
```


pkg/elf/usdtargs_linux.go
FedeDP, line 1: This file now contains some generic helpers to be used by arch-specific implementations.

pkg/elf/usdt_args.go
```diff
@@ -1,10 +1,82 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of Tetragon
 
-//go:build !amd64
+//go:build linux && (amd64 || arm64)
```
FedeDP: We only build this file on linux amd64|arm64.

pkg/elf/usdtargs_linux.go
```diff
@@ -1,10 +1,82 @@
...
+	"io"
+	"strings"
+)
+
+// Arg parsing entrypoint
+func parseArgs(spec *UsdtSpec) error {
+	for idx, str := range strings.Split(spec.ArgsStr, " ") {
+		arg := &spec.Args[idx]
+
+		for _, parse := range parsers {
```
FedeDP, line 18: `parsers` are set by arch-specific source files as package-wide variable.

pkg/elf/usdt_args.go
```diff
@@ -1,10 +1,82 @@
...
+	return nil
+}
+
+var errNext = errors.New("next")
+
+type fn func(str string, arg *UsdtArg) error
+
+type RegScanner struct {
+	name        string
+	closingRune rune
```
FedeDP: `)` for amd64 and `]` for arm64.

pkg/elf/usdtargs_linux_amd64.go
```diff
@@ -21,7 +18,15 @@ type reg struct {
 	off  uint16
 }
 
-var ptregs unix.PtraceRegs
+var (
+	ptregs  unix.PtraceRegs
+	parsers = []fn{
```
FedeDP, line 15: Definition of `parsers` for amd64.

pkg/elf/usdtargs_linux_arm64.go
```diff
@@ -0,0 +1,125 @@
...
+	"fmt"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+)
+
+var (
+	ptregs unix.PtraceRegs
+	// Parsers sorting IS important!
+	parsers = []fn{
```
FedeDP, line 18: Definition of `parsers` for arm64.

pkg/elf/usdtargs_linux.go
```diff
@@ -1,10 +1,81 @@
...
+func (sc *RegScanner) Scan(state fmt.ScanState, _ rune) error {
+	for {
+		r, _, err := state.ReadRune()
+		if errors.Is(err, io.EOF) {
+			break
+		}
+		if err != nil {
+			return err
+		}
+		if r == ',' || r == ' ' || r == closingRune {
```
FedeDP, line 72: `closingRune` is an arch-specific constant set by arch-specific go source files.

pkg/elf/usdt_args_amd64.go
```diff
@@ -7,61 +7,21 @@
...
-	"io"
-	"strings"
 
 	"github.com/cilium/tetragon/pkg/asm"
 )
 
-var errNext = errors.New("next")
+const closingRune = ')'
 
-type fn func(str string, arg *UsdtArg) error
```
olsajiri, line 19: could you please split this move in to separate change and add your arm changes on top? otherwise looks good, thanks

FedeDP: Of course!

FedeDP: Done :)

olsajiri (approved):
thanks

mtardy (commented):
So this a bit out of scope of this PR but since we have a bit of crazy file naming going on here we could use the implicit naming convention of Go files https://pkg.go.dev/cmd/go#hdr-Build_constraints

e.g. `reg_offset_amd64.go` becomes `regoffset_linux_amd64.go` and we can get rid of the build constraint altogether, same for the `*_test.go` files. Also the `FuzzAssignment` could be shared between amd64 and arm64 tests.

I'll suggest avoiding snake case in file names (since it's used by GOOS and GOARCH or _test), let's do regoffset and usdtargs or something like that

I could fix that after we merge this, it doesn't need to be in there. Tell me what you prefer, you can do a third commit :) otherwise we could merge this!

FedeDP:
> I could fix that after we merge this, it doesn't need to be in there. Tell me what you prefer, you can do a third commit :) otherwise we could merge this!

I 100% agree; let me fix it :)

FedeDP:
> Also the FuzzAssignment could be shared between amd64 and arm64 tests.

Btw right now `assignment` is amd64 only, thus i'll leave `FuzzAssignment` amd64 only :)

pkg/elf/usdt_args_others.go
FedeDP, line 1: Unused since `parseArgs` is only called by `usdt_linux.go`.

PR #4082, opened 2025-09-05
install/kubernetes/tetragon: Helm chart add serviceAnnotations and serviceLabels

yieon-lyon:
charts: add serviceAnnotations and serviceLabels for additive Service metadata management

Tetragon Helm chart currently exposes only serviceLabelsOverride which forces full replacement of the Service labels and makes day to day label management error prone.

This patch fixes this by adding:

A new values key serviceLabels as an empty map in values.yaml to allow appending labels without losing chart defaults.

A new values key serviceAnnotations as an empty map in values.yaml to allow appending annotations without losing chart defaults.

Service template logic that preserves compatibility. If serviceLabels is set it is merged with existing chart labels or serviceLabelsOverride to produce the final label set. If serviceAnnotations is set it is merged with existing chart annotations to produce the final annotations set.

Example values:
```
serviceLabels: {}
# serviceLabels:
#   k8s-app: tetragon

serviceAnnotations: {}
# serviceAnnotations:
#  prometheus.io/port: "2112"
#  prometheus.io/scrape: "true"
```

As a secondary note, these additive fields can simplify label based selection in systems such as Prometheus.

PR #4208, opened 2025-10-20
install/kubernetes/tetragon: Helm chart add serviceAnnotations and serviceLabels

yieon-lyon:
charts: add serviceAnnotations and serviceLabels for additive Service metadata management

Tetragon Helm chart currently exposes only serviceLabelsOverride which forces full replacement of the Service labels and makes day to day label management error prone.

This patch fixes this by adding:

A new values key serviceLabels as an empty map in values.yaml to allow appending labels without losing chart defaults.

A new values key serviceAnnotations as an empty map in values.yaml to allow appending annotations without losing chart defaults.

Service template logic that preserves compatibility. If serviceLabels is set it is merged with existing chart labels or serviceLabelsOverride to produce the final label set. If serviceAnnotations is set it is merged with existing chart annotations to produce the final annotations set.

Example values:
```
serviceLabels: {}
# serviceLabels:
# k8s-app: tetragon

serviceAnnotations: {}
# serviceAnnotations:
# prometheus.io/port: "2112"
# prometheus.io/scrape: "true"
```

As a secondary note, these additive fields can simplify label based selection in systems such as Prometheus.

### DIFF
--- a/docs/content/en/docs/reference/helm-chart.md
+++ b/docs/content/en/docs/reference/helm-chart.md
@@ -69,6 +69,8 @@ To use [the values available](#values), with `helm install` or `helm upgrade`, u
 | serviceAccount.annotations | object | `{}` |  |
 | serviceAccount.create | bool | `true` |  |
 | serviceAccount.name | string | `""` |  |
+| serviceAnnotations | object | `{}` |  |
+| serviceLabels | object | `{}` |  |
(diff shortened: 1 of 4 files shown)


mtardy:
Could you just `--force-with-lease` push on the previous PR instead of opening another https://github.com/cilium/tetragon/pull/4082 so that we keep history?

PR #4213, opened 2025-10-21
feat: align the argument types for k8s context

kemingy:
Fixes N/A

### Description

Update the argument types doc according to pkg/k8s/apis/cilium.io/v1alpha1/types.go

### Changelog

```release-note
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/argument_types.md
+++ b/docs/content/en/docs/concepts/tracing-policy/argument_types.md
@@ -31,6 +31,7 @@ List of described data types:
 - [`bpf_attr`](#bpf_attr)
 - [`perf_event`](#perf_event)
 - [`bpf_map`](#bpf_map)
+- [`bpf_prog`](#bpf_prog)
 - [`user_namespace`](#user_namespace)
 - [`capability`](#capability)
 - [`kiocb`](#kiocb)
@@ -144,6 +145,10 @@ TBD
 
 TBD
 
+## `bpf_prog`
+
+TBD
+
 ## `user_namespace`
 
 TBD
@@ -193,7 +198,7 @@ TBD
 The `linux_binprm` data type represents kernel `struct linux_binprm` object
 and retrieves the `struct linux_binprm::file` full path.
 
-See general path limitations in [path retrieval limits](#pathlimits))
+See general path limitations in [path retrieval limits](#pathlimits).
 
 ## `data_loc`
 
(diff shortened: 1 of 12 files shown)


mtardy (commented):
Thanks! Ah indeed some of the types are not reflected into the TracingPolicy API, @olsajiri since you created that page, I imagine you based yourself on https://github.com/cilium/tetragon/blob/27c9abe39c448c6f823c607eaad483d8c5717ecb/pkg/generictypes/generictypes.go#L81-L132

Do you know why we have this difference?

docs/content/en/docs/concepts/tracing-policy/argument_types.md
```diff
@@ -193,8 +187,6 @@ TBD
 The `linux_binprm` data type represents kernel `struct linux_binprm` object
 and retrieves the `struct linux_binprm::file` full path.
 
-See general path limitations in [path retrieval limits](#pathlimits))
-
```
mtardy, line 197: why are you removing this?

mtardy (changes requested):
So apparently instead of just deleting it from the docs we should either:
- document that these types are only supported in non-k8s context
- add support for k8s context, I don't think it should be too hard, it's just some plumbing missing

docs/content/en/docs/concepts/tracing-policy/argument_types.md
kemingy: Restored with the correct link.

kemingy:
> So apparently instead of just deleting it from the docs we should either:
> 
>     * document that these types are only supported in non-k8s context
> 
>     * add support for k8s context, I don't think it should be too hard, it's just some plumbing missing

Will try to support those types for k8s context.

Actually, I'm not sure what the non-k8s context means. I run it with the local Docker container, but clearly it requires the tracepoints policy YAML file, and those types are rejected since they violate the protobuf.

mtardy:
> Will try to support those types for k8s context.

Cool! I think it's the better option here.

> Actually, I'm not sure what the non-k8s context means. I run it with the local Docker container, but clearly it requires the tracepoints policy YAML file, and those types are rejected since they violate the protobuf.

Yeah previously they weren't validated because of a bug but indeed now they should be validated against the CRD so it should be rejected anyway.

mtardy (approved):
Thanks a lot for doing the wiring :)!!

PR #4219, opened 2025-10-21
fix(bpf/process): fix some missing `break` statements.

FedeDP:
Not a real fix since the `default` case would cover us, still it is error prone.


olsajiri (approved):
thanks

FedeDP:
> I'd also think about using fallthrough maybe and see if we can make clang warn us in case we don't.

Yup 100%: the more explicit we are, the better!
I'll leave it for a follow up PR!

PR #4221, opened 2025-10-21
tetragon: wrap ring samples in an event struct

kevsecurity:
Prior to adding support for the BPF ring buffer, the RunEvents() function in pkg/observer/observer_linux.go sent pointers to perfReader Records through the eventsQueue channel. As the perf ring and the BPF ring use different types, but the eventsQueue reader only needed the RawSample byte slice within, this was changed to send a pointer to the RawSample byte slice instead, which both the perf ring and BPF ring could provide.

We have seen panics where the slice that arrives out of the eventsQueue is empty, yet we have length checks prior to sending the slices to prevent this.

Perhaps passing pointers to slices through channels leads to reuse problems, with invalidated slices ending up with length 0? And perhaps sending the slices within a type makes a difference?

### DIFF
--- a/pkg/observer/observer_linux.go
+++ b/pkg/observer/observer_linux.go
@@ -26,6 +26,10 @@ const (
 	perCPUBufferBytes = 65535
 )
 
+type eventChannelRecord struct {
+	data []byte
+}
+
 func (k *Observer) getRBSize(cpus int) int {
 	var size int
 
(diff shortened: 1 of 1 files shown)


kevsecurity:
The real solution is: https://github.com/cilium/tetragon/pull/4231.

PR #4222, opened 2025-10-21
metrics: migrate Prometheus metrics to pkg/metrics helpers

### LINKED ISSUE #2798: Migrate metrics to use pkg/metrics helpers
`pkg/metrics` package provides functions to define metrics, which provide a few features on top of regular prometheus functions:
* define possible values for labels with known values
* define example values for metrics with unknown values
* initialize metrics with known labelsets at startup - for metrics with only constrained labels this is done automatically for all label combinations, but initialization function can be also provided explicitly

Ideally metrics with variable labels (defined with `prometheus.New*Vec`) should be migrated to use new helpers. This gives us a few benefits:
* Metrics are initialized at startup without the need for a separate initialization function defined and called somewhere else in the code (this was the tricky part of defining metrics so far, devs just needed to know that they should write initialization code).
* Metrics reference doc is generated using labels provided in the metric definition, not a separate initialization function (this is particularly useful for metrics with unconstrained labels, as they needed separate "example" label values for docs, and custom collectors, they needed whole separate types for docs).
* For labels with known values these values are constrained at runtime (this ensures that metrics cardinality and related resources usage are stable).

Examples:
tetragon_policyfilter_operations_total: https://github.com/cilium/tetragon/pull/2784/commits/6feacec60f1d3fc6a580014a86123a6faa8b7b88

AritraDey-Dev:
Fixes #2798

### Description

This PR migrates Prometheus metrics to use pkg/metrics helpers.

### Changelog

```release-note
Migrated metrics to use pkg/metrics helpers 
```

### DIFF
--- a/pkg/metrics/opcodemetrics/opcodemetrics.go
+++ b/pkg/metrics/opcodemetrics/opcodemetrics.go
@@ -14,20 +14,44 @@ import (
 )
 
 var (
-	MsgOpsCount = prometheus.NewCounterVec(prometheus.CounterOpts{
-		Namespace:   consts.MetricsNamespace,
-		Name:        "msg_op_total",
-		Help:        "The total number of times we encounter a given message opcode. For internal use only.",
-		ConstLabels: nil,
-	}, []string{"msg_op"})
+	MsgOpsCount = metrics.MustNewCounter(
+		metrics.NewOpts(
+			consts.MetricsNamespace, "", "msg_op_total",
+			"The total number of times we encounter a given message opcode. For internal use only.",
+			nil, []metrics.ConstrainedLabel{{Name: "msg_op", Values: func() []string {
+				res := make([]string, 0, len(ops.OpCodeStrings))
+				for opcode := range ops.OpCodeStrings {
+					// Exclude MSG_OP_UNDEF (0) as it's not a valid operational opcode - only used for error tracking.
+					// Also exclude MSG_OP_TEST as it's only used for testing purposes.
+					if opcode != ops.MSG_OP_UNDEF && opcode != ops.MSG_OP_TEST {
+						res = append(res, strconv.Itoa(int(int32(opcode))))
+					}
+				}
+				return res
+			}()}}, nil,
+		),
+		nil,
+	)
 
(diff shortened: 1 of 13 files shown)


mtardy:
Hey, could you wrap your commit description (use `gq` in vim for that for example)?

Also could the commit description contains more of the "why" of the change instead of the what (that should come from the title already), see more https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/#write-a-commit-message.

Thank you :)

pkg/metrics/opcodemetrics/opcodemetrics.go
```diff
@@ -14,20 +14,42 @@ import (
...
+
+	LatencyStats = metrics.MustNewHistogram(
+		metrics.HistogramOpts{
+			Opts: metrics.NewOpts(
+				consts.MetricsNamespace, "", "handling_latency",
+				"The latency of handling messages in us.",
+				nil, []metrics.ConstrainedLabel{{Name: "op", Values: func() []string {
+					res := make([]string, 0, len(ops.OpCodeStrings))
+					for opcode := range ops.OpCodeStrings {
+						if opcode != ops.MSG_OP_UNDEF && opcode != ops.MSG_OP_TEST {
```
kkourt, line 44: question(non-blocking): In errormetrics, you include undef (`// include UNDEF (0) to represent unknown opcodes in docs/metrics`) but here you do not. Any specific reason?

kkourt (approved):
Thanks!

LGTM, but this is really tricky to review because to check for correctness, you need to check, for example, that the `Inc()` method uses the same type as the one we use to generate the constrained labels. I wonder if there is a way to avoid this  duplication, but it's not obvious to me how.

pkg/metrics/opcodemetrics/opcodemetrics.go
AritraDey-Dev: In error metrics, we include `MSG_OP_UNDEF` to track unknown opcodes.

In opcode metrics, we exclude because it's not a valid operational opcode - it's a sentinel value for undefined/unknown operations.` MSG_OP_TEST` is also excluded as it's for testing.

I'll add a comment to opcodemetrics.go to make this clear.

PR #4223, opened 2025-10-22
fix(bpf): force explicit switch case fallthrough

FedeDP:
Added `-Wimplicit-fallthrough` clang flag to enforce it.

### Changelog

```release-note
fix(bpf): force explicit switch case fallthrough
```

### DIFF
--- a/bpf/lib/process.h
+++ b/bpf/lib/process.h
@@ -601,6 +601,7 @@ event_output_update_error_metric(u8 msg_op, long err)
 			break;
 		default:
 			lock_add(&valp->sent_failed[msg_op][SENT_FAILED_UNKNOWN], 1);
+			break;
 		}
 	}
 }
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1757,6 +1757,7 @@ selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx,
 		case fd_ty:
 			/* Advance args past fd */
 			args += 4;
+			fallthrough;
 		case file_ty:
 		case path_ty:
 		case dentry_type:
@@ -1790,7 +1791,7 @@ selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx,
 				pass = !!((cap_old ^ cap_new) & cap_new);
 				break;
 			}
-			/* falltrough */
+			fallthrough;
 #endif
 		case syscall64_type:
 		case s64_ty:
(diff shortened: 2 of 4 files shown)


FedeDP:
See https://github.com/cilium/tetragon/pull/4219#pullrequestreview-3361613205 request :) 
cc @mtardy

mtardy (commented):
looks good but I checkout-ed this, tried to remove fall throughs, compiled the BPF progs (with `make tetragon-bpf`) and didn't got any warning 🤔 .

EDIT: ah as long as the block is empty it does not generate a warning, nevermind

bpf/process/types/basic.h
```diff
@@ -1757,16 +1757,22 @@ selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx,
...
 			/* Advance args past fd */
 			args += 4;
+			fallthrough;
 		case file_ty:
+			fallthrough;
 		case path_ty:
+			fallthrough;
 		case dentry_type:
 #ifdef __LARGE_BPF_PROG
+			fallthrough;
```
mtardy: so those are superfluous maybe we can not add them on empty blocks in general? I think that's a good compromise to have neat switch case at the same and we have the compiler screaming if someone's changing this

FedeDP:
Yes, example removing the `break` statement that would've prevented github.com/cilium/tetragon/pull/4218:
```
In file included from process/bpf_generic_lsm_core.c:47:
process/generic_calls.h:974:2: error: unannotated fall-through between switch labels [-Werror,-Wimplicit-fallthrough]
  974 |         case ACTION_SET:
      |         ^
process/generic_calls.h:974:2: note: insert '__attribute__((fallthrough));' to silence this warning
  974 |         case ACTION_SET:
      |         ^
      |         __attribute__((fallthrough)); 
process/generic_calls.h:974:2: note: insert 'break;' to avoid fall-through
  974 |         case ACTION_SET:
      |         ^
      |         break; 
1 error generated.
```

bpf/process/types/basic.h
FedeDP: I agree, i went to berserk mode on that one :imp:

bpf/lib/process.h
```diff
@@ -601,6 +601,7 @@ event_output_update_error_metric(u8 msg_op, long err)
 			break;
 		default:
 			lock_add(&valp->sent_failed[msg_op][SENT_FAILED_UNKNOWN], 1);
+			break;
```
mtardy, line 604: minor nit sorry: this one's useless as well

FedeDP: Yes but i'd prefer to keep the `break` even for the `default` case!

mtardy: ok as you wish, we have `-Wimplicit-fallthrough` to watch our back now here but I'm fine with this! let's merge this once it's green then

PR #4226, opened 2025-10-22
fix(pkg/sensors): account for sleepableOffloadMap in multi uprobe

FedeDP:
Extracted from #4193 

### Description

This was a leftover from https://github.com/cilium/tetragon/pull/4173

### Changelog

```release-note
fix(pkg/sensors): account for sleepableOffloadMap in multi uprobe
```


mtardy (commented):
For reference:

> ah nice.. could you please push separate fix for uprobe_multi part that goes before your changes? the change for uretprobe below would not be necessary, thanks

_Originally posted by @olsajiri in https://github.com/cilium/tetragon/pull/4193#discussion_r2449260642_

PR #4230, opened 2025-10-22
tetragon: send slices to the events queue

kevsecurity:
When reading the ring buffers, we check the length of the byte slice is greater than 0, and then send a reference to it to the events queue channel. Sometimes on retrieval, the byte slice has a length of 0.

This commit attempts to resolve this by sending the byte slice to the events queue channel instead, on the assumption that this will make tracking the backing array easier for golang, so it shouldn't invalidate it before we use it.


kevsecurity:
The real solution is https://github.com/cilium/tetragon/pull/4231.

PR #4233, opened 2025-10-22
workflows: fix release image display digest step

mtardy:
Docker build push action thought it was a good idea to add a build summary and uploading artifacts that can't be downloaded via the GitHub actions/download-artifact so we need to adapt each of our workflows.

See https://github.com/docker/build-push-action?tab=readme-ov-file#summaries and https://github.com/actions/toolkit/pull/1874.

### DIFF
--- a/.github/workflows/build-images-releases.yml
+++ b/.github/workflows/build-images-releases.yml
@@ -62,6 +62,8 @@ jobs:
       - name: Release Build ${{ matrix.name }}
         uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6.18.0
         id: docker_build_release
+        env:
+          DOCKER_BUILD_RECORD_UPLOAD: false
         with:
           provenance: false
           context: .


kkourt (approved):

FedeDP:
What about setting the `DOCKER_BUILD_SUMMARY` env variable to false instead?
> Summaries are enabled by default, but can be disabled with the DOCKER_BUILD_SUMMARY [environment variable](https://github.com/docker/build-push-action?tab=readme-ov-file#environment-variables).

https://github.com/docker/build-push-action?tab=readme-ov-file#environment-variables

Perhaps even `DOCKER_BUILD_RECORD_UPLOAD` might work fine if we want to retain the summary but kill the upload (?)


EDIT: either case, we could just set them as repository-wide envs and we should be good to go, without touching any workflow.

mtardy:
> Perhaps even `DOCKER_BUILD_RECORD_UPLOAD` might work fine if we want to retain the summary but kill the upload (?)

oh this is what I was looking for but in the args not in the env variables

mtardy:
You're right, I set up the DOCKER_BUILD_RECORD_UPLOAD variable repository wide

mtardy:
lol it didn't work I wonder why https://github.com/cilium/tetragon/actions/runs/18725426276/job/53410703924

FedeDP:
> Check build summary support
  Build summary supported!

I did not expect anything different :rofl: 
That's weird btw, seems like the action is not seeing `DOCKER_BUILD_RECORD_UPLOAD=false` being set :/

EDIT: it just checks `process.env.FOO`: https://github.com/docker/build-push-action/blob/9e436ba9f2d7bcd1d038c8e55d039d37896ddc5d/src/main.ts#L277

PR #4234, opened 2025-10-22
tetragon: Fix check_cap tester program call

kevsecurity:
Adding missing cap argument to check_cap function and while at it make all local functions static.

[Upstream commit: 9a5fdcfff]


mtardy (approved):
Thank you

Issue #4204, opened 2025-10-17
Support more than 38 override matchActions on LSM function


holyspectral:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

Currently the override action of kprobe sensors on LSM functions are based on `fmod_ret` eBPF programs. This feature is great because it allows users to write rules based on the arguments of LSM functions and to deny unwanted behavior.  The issues like the TOCTOU race or symlink is not a problem, because LSM already handles them.  

However, for this feature to be very useful, there are still some issues.
- The `fmod_ret` program is implemented based on eBPF trampoline.  Linux kernel only allows up to [38 eBPF trampolines](https://elixir.bootlin.com/linux/v6.14.11/source/include/linux/bpf.h#L1138).  That means that we can only have 38 fmod_ret programs attached by Tetragon at maximum.
- Under the current Tetragon design, for each override matchAction specified on lsm functions, a `fmod_ret` program is created and attached by Tetragon.  That means when a new policy is created on the same LSM function, a new fmod_ret programs will still be created and take a seat of eBPF trampolines.

For example, when we have two tracing policies dealing with same two LSM functions (A & B), we would have the graph below.  Despite only two LSM functions used, 4 `override_tasks` maps will be created, and 4 seats of eBPF trampolines will be taken by `generic_fmodret_override` programs.

```mermaid
  graph TD;
      TracingPolicy1-->singleKprobeSensor1;
      TracingPolicy1-->singleKprobeSensor2;
      singleKprobeSensor1-->map1[(override_tasks on A function)];
      singleKprobeSensor2-->map2[(override_tasks on B function)];
      map1-->fmod_ret1[generic_fmodret_override on A function]
      map2-->fmod_ret2[generic_fmodret_override on B function]

      TracingPolicy2-->singleprobeSensor3;
      TracingPolicy2-->singleKprobeSensor4;
      singleprobeSensor3-->map3[(override_tasks on A function)];
      singleKprobeSensor4-->map4[(override_tasks on B function)];
      map3-->fmod_ret3[generic_fmodret_override on A function]
      map4-->fmod_ret4[generic_fmodret_override on B function]
```

### Describe the feature you would like

We would like to have more than 38 policies to override return values on lsm functions.

### Describe your proposed solution

I think we can allow each policy to share `generic_fmodret_override` program and related maps, as long as they're attached to the same function.

[...]

PR #4244, opened 2025-10-23
feat(sensors): support 38+ override on lsm funcs

holyspectral:
Fixes  #4204 

### Description

As mentioned in #4204, this PR makes multiple kprobe sensors to share their `fmod_ret` programs, as long as they attach to the same function, so we can allow more than 38+ overrides on lsm functions.  

#### (update Dec. 11, 2025) The changes since draft PR

- Refactored and moved most logic to pkg/sensors/program/
- Introduced an override ID in both userspace and ebpf to make sure that an override action is only handled by correct override programs.
- Fixed an issue in override_tasks map's lifecycle. 

#### The changes:

1. Now override programs are maintained separately from the kprobe sensors.  By saying that, it means:
  - `override_tasks` becomes a shared global map.
  - override programs (both kprobe and fmod_ret) are created and maintained separately.  So multiple kprobe sensors can share the same override program.
 
2. The pinned map/programs's location will be as below:

```
/bpffs/tetragon/__override__
/bpffs/tetragon/__override__/kprobe
/bpffs/tetragon/__override__/kprobe/__x64_sys_symlinkat
/bpffs/tetragon/__override__/kprobe/__x64_sys_symlinkat/link_override
/bpffs/tetragon/__override__/kprobe/__x64_sys_symlinkat/prog_override
/bpffs/tetragon/__override__/kprobe/__x64_sys_execve
/bpffs/tetragon/__override__/kprobe/__x64_sys_execve/link_override
/bpffs/tetragon/__override__/kprobe/__x64_sys_execve/prog_override
/bpffs/tetragon/__override__/override_tasks
/bpffs/tetragon/__override__/fmod_ret
/bpffs/tetragon/__override__/fmod_ret/security_bprm_creds_for_exec
/bpffs/tetragon/__override__/fmod_ret/security_bprm_creds_for_exec/prog
```

### Changelog

```release-note
```


holyspectral:
@olsajiri may I have your feedback on this one?  Thanks a lot.

pkg/sensors/tracing/genericfmod_ret.go
```diff
@@ -0,0 +1,55 @@
...
+	if fmodretMap == nil {
+		fmodretMap = make(map[string]*program.Program)
+	}
+
+	loadProgName, _ := config.GenericKprobeObjs(false)
+
+	if fmodret, ok = fmodretMap[attachFunc]; !ok {
+
+		fmodret = program.Builder(
+			path.Join(option.Config.HubbleLib, loadProgName),
```
olsajiri: since we already add new program for this, could we move override programs into separated object?
that might speed up the load

pkg/sensors/tracing/genericfmod_ret.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"path"
+
+	"github.com/cilium/tetragon/pkg/config"
+	"github.com/cilium/tetragon/pkg/option"
+	"github.com/cilium/tetragon/pkg/sensors/program"
+)
+
+var fmodretMap map[string]*program.Program
```
olsajiri: I was thinking about something similar but bit more generic to cover also standard kprobe override programs.. there's no limitation for attached kprobe programs, but it would at least benefit from having just single copy of the program and reduce the footprint

pkg/sensors/tracing/genericfmod_ret.go
```diff
@@ -0,0 +1,55 @@
...
+			path.Join(option.Config.HubbleLib, loadProgName),
+			attachFunc,
+			"fmod_ret/security_task_prctl",
+			"fmod_ret/"+attachFunc,
+			"generic_fmod_ret")
+
+		fmodret.PinPath = "fmod_ret/" + attachFunc
+
+		fmodretmap := program.MapBuilder("override_tasks", fmodret)
+		fmodretmap.PinPath = path.Join("fmod_ret/", attachFunc, "override_tasks")
```
olsajiri: i'd rather place it in some visibly special place like we do for base sensor, perhaps something like:

```
__override__/fmodret/security_xxx
__override__/fmodret/security_yyy
               ...
__override__/kprobe/ksys_read
__override__/kprobe/ksys_write
               ...
__override__/override_tasks
```

pkg/sensors/load.go
```diff
@@ -65,8 +65,10 @@ func (s *Sensor) policyDir() string {
 
 func (s *Sensor) createDirs(bpfDir string) {
 	for _, p := range s.Progs {
-		// setup sensor based program pin path
-		p.PinPath = filepath.Join(s.policyDir(), s.Name, p.PinName)
+		// setup sensor based program pin path if it's not specified
+		if p.PinPath == "" {
```
olsajiri, line 69: is this needed, or it's just a leftover?  PinPath should be always defined even with your change no?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1234,6 +1259,21 @@ func loadMultiKprobeSensor(ids []idtable.EntryID, bpfDir string, load *program.P
...
+func loadGenericFmodRetProgram(bpfDir string, load *program.Program, maps []*program.Map, verbose int) error {
+	if load.LoadState.IsLoaded() {
+		logger.GetLogger().Info(fmt.Sprintf("The generic fmodify return program on %s has been loaded", load.Attach))
+		return nil
+	}
+
+	logger.GetLogger().Info("loading generic fmod ret program", "prog", load)
+
+	unload := func() {
+		deleteFmodRetProg(load.Attach)
```
olsajiri: could we use program.Program.unloaderOverride or add something in there so it gets removed during program's unload?

pkg/sensors/tracing/genericfmod_ret.go
holyspectral: Sure.  I think it makes a lot of sense to do this.  Thanks!

pkg/sensors/tracing/genericfmod_ret.go
holyspectral: I think that's a good idea.  Let me look into this.

pkg/sensors/tracing/generickprobe.go
holyspectral: Ah I didn't notice this function exists when I wrote the code.  From its name it's definitely more suitable.  Let me look into this.

pkg/sensors/load.go
holyspectral: yeah there might be a better way to implement this, but the code here is basically a lazy initialization.  When the PinPath is already given, it uses the existing one.  For override program, we prefill the `PinPath` so it falls under this category.  

For other programs, the default `PinPath` under `s.policyDir()` will be used in Ln70.

pkg/sensors/tracing/genericfmod_ret.go
holyspectral: One issue I noticed is that, if I'm not mistaken, for kprobe override programs, when multi-kprobe is enabled, each kprobe attach points actually share the same `override_maps`.  Do you think we should change that and use per-hookpoint map instead?

olsajiri: hum not sure what you mean..

holyspectral: Ah, I wasn't aware that the creation and loading of maps and programs in multi kprobe flows are based on the data type of `load.LoaderData`.  That makes sense now.  Thanks a lot!

holyspectral:
Changes since v1:

1. Allow kprobe to use shared override programs.
2. Instead of having one override_tasks map for each hook point, in v2 each override program shares the same override_tasks map.  This helps to support multi-kprobe scenario.
3. The pinned path of ebpf programs and maps will be as below:

```
/bpffs/tetragon/__override__
/bpffs/tetragon/__override__/kprobe
/bpffs/tetragon/__override__/kprobe/__x64_sys_symlinkat
/bpffs/tetragon/__override__/kprobe/__x64_sys_symlinkat/link_override
/bpffs/tetragon/__override__/kprobe/__x64_sys_symlinkat/prog_override
/bpffs/tetragon/__override__/kprobe/__x64_sys_execve
/bpffs/tetragon/__override__/kprobe/__x64_sys_execve/link_override
/bpffs/tetragon/__override__/kprobe/__x64_sys_execve/prog_override
/bpffs/tetragon/__override__/override_tasks
/bpffs/tetragon/__override__/fmod_ret
/bpffs/tetragon/__override__/fmod_ret/security_bprm_creds_for_exec
/bpffs/tetragon/__override__/fmod_ret/security_bprm_creds_for_exec/prog
```

4. Make sure that unused override programs are cleaned up via
unloaderOverride().

5. Move bpf functions regarding overrides into another object file.

holyspectral:
I still keep this as draft because of an item that I'd like to discuss first.  Say we have a scenario is like the below (keep in mind that in v2 we use a shared override_tasks map):

- We have two tracing policies A & B.
    - PolicyA is hooked on `sys_execve` and policyB is on `sys_symlinkat`.
    - They both have their own override action in matchActions. 
- A process triggers the override action of policy A via `sys_execve`, and has an item inserted into the `override_tasks` map.
    - override_tasks[<pid_tgid>] = -EPERM
- Before the program triggers the override program attached to `sys_execve`, a policy change on A happens and the override program on `sys_execve` is removed from another core.
    - the content of the map stays the same, override_tasks[<pid_tgid>] = -EPERM
    - but the override program on `sys_execve` is removed.
- When the program triggers policy B via `sys_symlinkat` later, because `override_tasks[<pid_tgid>] = -EPERM`, the action will be denied unexpectedly, even if nothing is matched.

I can think of a few directions:

1. Maybe this is not a real issue, the window is too short, or Tetragon has already handled this, so we don't have to care about it. 

2. Change the `override_tasks` to be a BPF_MAP_TYPE_HASH_OF_MAPS and let its inner map be associated to each hook points, so we can clean up the content of override_tasks map when we remove an override program.  For example,

```
    override_tasks: {
        "syscall:sys_execve": <inner map like the current override_tasks map>,
        "syscall:sys_symlinkat": <inner map like the current override_tasks map>,
        "fmod_ret:security_bprm_creds_for_exec": <inner map like the current override_tasks map>,
    }
```

3. We don't delete the override programs immediately when its policy is removed.  This will give the override programs some time to remove the items.

@olsajiri do you think this is a real issue that we should address?  I'd love to know your thoughts on this.

olsajiri:
> I can think of a few directions:
> 
>     1. Maybe this is not a real issue, the window is too short, or Tetragon has already handled this, so we don't have to care about it.
> 
>     2. Change the `override_tasks` to be a BPF_MAP_TYPE_HASH_OF_MAPS and let its inner map be associated to each hook points, so we can clean up the content of override_tasks map when we remove an override program.  For example,
> 
> 
> ```
>     override_tasks: {
>         "syscall:sys_execve": <inner map like the current override_tasks map>,
>         "syscall:sys_symlinkat": <inner map like the current override_tasks map>,
>         "fmod_ret:security_bprm_creds_for_exec": <inner map like the current override_tasks map>,
>     }
> ```
> 
>     3. We don't delete the override programs immediately when its policy is removed.  This will give the override programs some time to remove the items.
> 
> 
> @olsajiri do you think this is a real issue that we should address? I'd love to know your thoughts on this.

yea, that seems like a problem.. so at the moment override_task is program's map, so each override program has its own copy, I'd suggest to have some state of this change doing the same, and adding a change to single map on top of that

perhaps we could have policy id as part of the override_task value and have sensor unload to cleanup its records before it unloads the override program..  something like you suggest in 2) but not sure what's benefit of inner map
 
I'll check on your change in more detail but on first glance please try to split the change into more logical changes/commits, it's easier to review, thanks

holyspectral:
> perhaps we could have policy id as part of the override_task value and have sensor unload to cleanup its records before it unloads the override program.. something like you suggest in 2) but not sure what's benefit of inner map

Thanks.  I think that makes sense.  Let me see if I can add an extra field to the key or value of override_tasks map.  

> I'll check on your change in more detail but on first glance please try to split the change into more logical changes/commits, it's easier to review, thanks

I've split them into multiple commits.  Please feel free to let me know if anything is not clear!

pkg/sensors/program/loader_linux.go
```diff
@@ -692,13 +536,59 @@ func LoadUprobeProgram(bpfDir string, load *Program, maps []*Map, verbose int) e
...
+func LoadKProbeOverrideProgram(bpfDir string, load *Program, maps []*Map, progName string, verbose int, unloadFunc func(bool) error) error {
+	opts := &LoadOpts{
+		Attach: func(
+			_ *ebpf.Collection,
+			_ *ebpf.CollectionSpec,
+			prog *ebpf.Program,
+			spec *ebpf.ProgramSpec,
+		) (unloader.Unloader, error) {
+
+			pinPath := filepath.Join(bpfDir, load.PinPath, "prog_override")
```
olsajiri: perhaps no need for the override suffix now? also for link

pkg/sensors/program/loader_linux.go
```diff
@@ -692,13 +536,59 @@ func LoadUprobeProgram(bpfDir string, load *Program, maps []*Map, verbose int) e
...
+		},
+		Maps: maps,
 	}
 	return loadProgram(bpfDir, load, opts, verbose)
 }
 
-func LoadFmodRetProgram(bpfDir string, load *Program, maps []*Map, progName string, verbose int) error {
+func LoadKProbeOverrideProgram(bpfDir string, load *Program, maps []*Map, progName string, verbose int, unloadFunc func(bool) error) error {
+	opts := &LoadOpts{
+		Attach: func(
```
olsajiri: we tend to put callbacks code in functions, please add override attach and open fnuctions

pkg/sensors/program/loader_linux.go
```diff
@@ -692,13 +536,59 @@ func LoadUprobeProgram(bpfDir string, load *Program, maps []*Map, verbose int) e
...
+			}
+
+			return unload, err
+		},
+		Open: func(coll *ebpf.CollectionSpec) error {
+			progSpec, ok := coll.Programs[progName]
+			if !ok {
+				return fmt.Errorf("progName %s not in collecition spec programs: %+v", progName, coll.Programs)
+			}
+			progSpec.AttachTo = load.Attach
```
olsajiri: is this needed for kprobe?

pkg/sensors/tracing/genericoverride.go
```diff
@@ -0,0 +1,102 @@
...
+
+func createFmodRetOverrideProg(attachFunc string) *program.Program {
+	overrideProg := program.Builder(
+		"bpf_generic_override.o",
+		attachFunc,
+		"fmod_ret/security_task_prctl",
+		"fmod_ret/"+attachFunc,
+		"generic_fmod_ret")
+
+	overrideProg.PinPath = path.Join("__override__", "fmod_ret", attachFunc)
```
olsajiri: I think it's probably ok, but could you please check tetrag debug progs/maps works properly with this?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1007,6 +1015,38 @@ func addKprobe(funcName string, instance int, f *v1alpha1.KProbeSpec, in *addKpr
...
+	overrideTasksMap.PinPath = kprobeOverrideMap.PinPath
+	overrideTasksMap.SetMaxEntries(overrideMapMaxEntries)
+
+	maps = append(maps, overrideTasksMap)
+
+	return progs, maps
+}
+
+func createFmodRetOverrideProgramFromEntry(load *program.Program, attachFunc string, progs []*program.Program, maps []*program.Map) ([]*program.Program, []*program.Map) {
+	// setup fmodret program and its input
```
olsajiri: looks like this function and the above createKProbeOverrideProgramFromEntry could be unified in just one function with load.OverrideFmodRet as bool argument

pkg/sensors/program/loader_linux.go
```diff
@@ -715,6 +605,11 @@ func LoadFmodRetProgram(bpfDir string, load *Program, maps []*Map, progName stri
 			if err != nil {
 				return nil, fmt.Errorf("attaching '%s' failed: %w", spec.Name, err)
 			}
+			if unloadFunc != nil {
+				load.unloaderOverride = &unloader.CustomUnloader{
+					UnloadFunc: unloadFunc,
```
olsajiri: could we just set load.unloaderOverride in getOverrideProg ? then we could leave LoadFmodRetProgram as is

pkg/sensors/program/loader_linux.go
```diff
@@ -692,13 +536,59 @@ func LoadUprobeProgram(bpfDir string, load *Program, maps []*Map, verbose int) e
...
+		Open: func(coll *ebpf.CollectionSpec) error {
+			return nil
+		},
+		Maps: maps,
 	}
 	return loadProgram(bpfDir, load, opts, verbose)
 }
 
-func LoadFmodRetProgram(bpfDir string, load *Program, maps []*Map, progName string, verbose int) error {
+func LoadKProbeOverrideProgram(bpfDir string, load *Program, maps []*Map, progName string, verbose int, unloadFunc func(bool) error) error {
```
olsajiri: hm why do we need special override function for kprobe and not for fmodret? can't we just use LoadKprobeProgram?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1484,39 +1548,32 @@ func retprobeMerge[T evArgsRetriever](prev pendingEvent[T], curr pendingEvent[T]
...
-	}
-	pendingEvents.Add(key, curr)
-	return false, nil, nil
+func (k *fmodRetProgram) LoadProbe(args sensors.LoadProbeArgs) error {
+	return loadGenericFmodRetProgram(args.BPFDir, args.Load, args.Maps, args.Verbose)
 }
 
-func (k *observerKprobeSensor) LoadProbe(args sensors.LoadProbeArgs) error {
-	return loadGenericKprobeSensor(args.BPFDir, args.Load, args.Maps, args.Verbose)
+func (k *kprobeOverrideProgram) LoadProbe(args sensors.LoadProbeArgs) error {
```
olsajiri, line 1621: so the kprobe override program depends on being loaded 'before' the actual program so it gets in the kprobe list behind the actual program, which is laded after.. tricky ;-) I guess if tests are passing we're doing it somehow, but I wonder we specifically ensure this

holyspectral: yeah exactly.  I will double check to make sure it's loaded in the right order.

pkg/sensors/program/loader_linux.go
holyspectral: Thanks for the feedback!  In the current code base, in addition to override programs, `LoadKprobeProgram()` loads kprobe programs used in kprobe sensors and `LoadFmodRetProgram()` loads the override programs for enforcer, so it was an attempt to ensure that the `unloaderOverride()` is only specified and called in the override program scenario.

Combining all the comments together, now I think it makes more sense to move the `genericoverride.go` to under `github.com/cilium/tetragon/pkg/sensors/program`.  This way I can assign unexported `unloaderOverride` there, and then I don't need many changes in this version.

Thanks for your feedback again!

holyspectral:
Hi @olsajiri thanks for your patience.  I've updated v3 in this PR and changed this PR as ready for review.  The change since  last time:

- Refactored and moved most logic to pkg/sensors/program/
- Introduced an override ID in both userspace and ebpf to make sure that an override action is only handled by correct override programs.
- Fixed an issue in override_tasks map's lifecycle.

I'll appreciate any feedback on this.

olsajiri:
> Hi @olsajiri thanks for your patience. I've updated v3 in this PR and changed this PR as ready for review. The change since last time:
> 
>     * Refactored and moved most logic to pkg/sensors/program/
> 
>     * Introduced an override ID in both userspace and ebpf to make sure that an override action is only handled by correct override programs.
> 
>     * Fixed an issue in override_tasks map's lifecycle.
> 
> 
> I'll appreciate any feedback on this.

@holyspectral I'm checking on that, would you mind to rebase it? it'd be easier for me to run it, thanks

holyspectral:
Looks like there are issues from vmtests on older kernels.  I will take a look.

pkg/sensors/program/loader_linux.go
olsajiri: same question with new version..

pkg/sensors/program/loader_linux.go
```diff
@@ -683,7 +683,51 @@ func LoadMultiKprobeProgram(bpfDir string, load *Program, maps []*Map, verbose i
...
```
olsajiri, line 686: why is this added when it's completely removed in one of the following commit?
also the commit log mentions only unloader, not this function

pkg/sensors/tracing/generickprobe.go
```diff
@@ -345,11 +365,23 @@ func createMultiKprobeSensor(polInfo *policyInfo, multiIDs []idtable.EntryID, ha
...
+			gk, err := genericKprobeTableGet(id)
+			if err != nil {
+				return nil, nil, err
+			}
+			gk.data = &genericKprobeData{}
+
+			progs, maps = createKProbeOverrideProgramFromEntry(load, gk.funcName, progs, maps)
+		}
+	} else {
+		overrideTasksMap := program.MapBuilderProgram("override_tasks", load)
```
olsajiri, line 379: seems wrong.. if we have load.override we want to ping the override_tasks map, right? you're doing it for the !load.override case

pkg/sensors/tracing/generickprobe.go
```diff
@@ -345,11 +365,23 @@ func createMultiKprobeSensor(polInfo *policyInfo, multiIDs []idtable.EntryID, ha
...
 	if has.override {
-		overrideTasksMap.SetMaxEntries(overrideMapMaxEntries)
+		for _, id := range multiIDs {
+			gk, err := genericKprobeTableGet(id)
+			if err != nil {
+				return nil, nil, err
+			}
+			gk.data = &genericKprobeData{}
+
+			progs, maps = createKProbeOverrideProgramFromEntry(load, gk.funcName, progs, maps)
```
olsajiri, line 376: should we have the load.OverrideFmodRet check in here and call createFmodRetOverrideProgramFromEntry?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1520,3 +1621,11 @@ func retprobeMergeEvents[T evArgsRetriever](unix T, pendingEvents *lru.Cache[pen
 func (k *observerKprobeSensor) LoadProbe(args sensors.LoadProbeArgs) error {
 	return loadGenericKprobeSensor(args.BPFDir, args.Load, args.Maps, args.Verbose)
 }
+
+func (k *fmodRetProgram) LoadProbe(args sensors.LoadProbeArgs) error {
+	return loadGenericFmodRetProgram(args.BPFDir, args.Load, args.Maps, args.Verbose)
+}
+
+func (k *kprobeOverrideProgram) LoadProbe(args sensors.LoadProbeArgs) error {
```
olsajiri, line 1629: so how do we know this gets called before the main program load to ensure it's called in the right order?

pkg/sensors/program/genericoverride.go
```diff
@@ -1,68 +1,82 @@
 // SPDX-License-Identifier: Apache-2.0
```
olsajiri, line 1: maybe rename genericoverride.go to just override.go

pkg/sensors/program/genericoverride.go
```diff
@@ -59,24 +80,24 @@ func createKProbeOverrideProg(attachFunc string) *Program {
...
 }
 
 func GetOverrideProg(overrideType OverrideType, attachFunc string) (*Program, *Map) {
-	var overrideProg *Program
+	var overrideProg *genericOverride
 	var ok bool
 
 	if overrideProgMap == nil {
-		overrideProgMap = make(map[string]*Program)
+		overrideProgMap = make(map[string]*genericOverride)
```
olsajiri, line 100: this needs more explanation on why is this needed.. please use commit log message to describe what you are doing and why.. for now and for future ;-)

bpf/process/types/basic.h
```diff
@@ -264,7 +264,7 @@ struct msg_linux_binprm {
 #ifdef __MULTI_KPROBE
 FUNC_INLINE __u32 get_index(void *ctx)
 {
-	return (__u32)get_attach_cookie(ctx);
+	return (__u32)(get_attach_cookie(ctx) & 0x0f);
```
olsajiri, line 267: that's real scary ;-)

again, please explain in the commit log message what you are doing and why.. I fail to see why this is needed

bpf/process/types/basic.h
```diff
@@ -2503,20 +2503,4 @@ FUNC_INLINE const struct path *get_path(long type, unsigned long arg, struct pat
 			     : "+r"(x));                \
 	})
 
-FUNC_INLINE int try_override(void *ctx, struct bpf_map_def *override_tasks)
```
olsajiri, line 2506: please make the change frst before moving the function, it's hard to review what changed

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1302,8 +1309,11 @@ func loadMultiKprobeSensor(ids []idtable.EntryID, bpfDir string, load *program.P
 		}
 		load.MapLoad = append(load.MapLoad, config)
 
+		// 0-3: index, 4-11: override_id, others: reserved
+		cookies := index&0x0f | (gk.loadArgs.overrideID << 4)
```
olsajiri, line 1313: what's the use case for this? please explain

pkg/sensors/program/genericoverride.go
```diff
@@ -163,10 +165,40 @@ func deleteOverrideProg(overrideType OverrideType, attachFunc string, override *
...
+
 	_, err = overrideIDTable.RemoveEntry(override.tableId)
 
 	if err != nil {
 		return fmt.Errorf("failed to remove entry: %w", err)
 	}
 	return nil
 }
+
+func cleanupPendingDeletionOverrideMap(id int) error {
```
olsajiri, line 181: this should be part of some of the previous commits right?

olsajiri (changes requested):
the first few changes look good but then I fail to see the reason for some of the following changes, I left some comments, please check

I think last 3 commits are fixes of your previous commits and should be squash where they belong

mtardy:
Could you reorganize your PR commits, squash the things that need to be squashed (lints, back and forth change in the code), separate the others correctly? It's a bit hard to review as of now, I see Jiri proposed some ideas. Also some commits seems misstitled ("kprobe: support 38+ override on lsm funcs"). It should appear in the patch set how we should merge it to the main branch. Thanks a lot!

kkourt:
Marking this PR as draft. @holyspectral feel free to undraft when ready.

PR #4252, opened 2025-10-27
new(bpf,pkg,docs): use tracefs instead of debugfs.

### LINKED ISSUE #4250: Tetragon seems to depend on `TRACEFS_AUTOMOUNT_DEPRECATED`
I had Tetragon's daemonset on a crashloop with 

`level=error msg="Failed to execute tetragon" error="sensor __base__ from collection __base__ failed to load: failed prog /var/lib/tetragon/bpf_execve_event_v612.o kern_version 397824 loadInstance: attaching 'event_execve' failed: neither debugfs nor tracefs are mounted"` 

until I enabled `CONFIG_TRACEFS_AUTOMOUNT_DEPRECATED` in the kernel config.

`TRACEFS_AUTOMOUNT_DEPRECATED` was introduced in Linux v6.17-rc1
See https://github.com/torvalds/linux/commit/9ba817fb7c6afd3c86a6d4c3b822924b87ef0348

It is curious how I ended up without it (it's `default y` in Kconfig) but maybe also worth documenting?

When enabled, dmesg spits out `NOTICE: Automounting of tracing to debugfs is deprecated and will be removed in 2030` when Tetragon comes up.

FedeDP:
Fixes #4250

### Description
TraceFS (`/sys/kernel/tracing` mount) is available since linux 4.1.
DebugFS automout of tracefs (ie: `/sys/kernel/debug/tracing`) is going to get sunset in 2030, and is now behind `CONFIG_TRACEFS_AUTOMOUNT_DEPRECATED`.
See https://github.com/torvalds/linux/commit/9ba817fb7c6afd3c86a6d4c3b822924b87ef0348.
Drop debugfs altogether and just 

### Changelog

```release-note
new(bpf,pkg,docs): use tracefs instead of debugfs.
```

### DIFF
--- a/pkg/sensors/test/checker_test.go
+++ b/pkg/sensors/test/checker_test.go
@@ -9,7 +9,6 @@ import (
 	"context"
 	"errors"
 	"log/slog"
-	"os"
 	"sync"
 	"testing"
 
@@ -22,7 +21,7 @@ import (
 
 // TestTestChecker tests the test checker
 func TestTestChecker(t *testing.T) {
-	if _, err := os.Stat("/sys/kernel/tracing/events/syscalls"); os.IsNotExist(err) {
+	if !testutils.CheckKernelTracingExists() {
 		t.Skip("cannot use syscall tracepoints (consider enabling CONFIG_FTRACE_SYSCALLS)")
 	}
 
--- a/pkg/sensors/test/lseek_test.go
+++ b/pkg/sensors/test/lseek_test.go
@@ -14,6 +14,8 @@ import (
 	"github.com/stretchr/testify/require"
 	"golang.org/x/sys/unix"
 
+	"github.com/cilium/tetragon/pkg/testutils"
+
 	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
 	"github.com/cilium/tetragon/pkg/jsonchecker"
 	"github.com/cilium/tetragon/pkg/observer/observertesthelper"
@@ -30,7 +32,7 @@ func TestMain(m *testing.M) {
 }
 
 func TestSensorLseekLoad(t *testing.T) {
-	if _, err := os.Stat("/sys/kernel/tracing/events/syscalls"); os.IsNotExist(err) {
+	if !testutils.CheckKernelTracingExists() {
 		t.Skip("cannot use syscall tracepoints (consider enabling CONFIG_FTRACE_SYSCALLS)")
 	}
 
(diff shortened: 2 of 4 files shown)


FedeDP:
I think it is safer to add a fallback to old debugfs path. I will push the changes soon :)

pkg/tracepoint/tracepoint.go
```diff
@@ -43,6 +48,16 @@ type FieldFormat struct {
 	IsSigned bool
 }
 
+func init() {
+	// Try to use traceFS first
+	_, err := os.Stat(tracepointsPath)
+	if os.IsNotExist(err) {
+		// Fallback at using debugFS
+		usedTracepointPath = debugFsTracepointPath
+		logger.GetLogger().Warn("TraceFS not available; fallback at using debugFS", "tracefs", tracepointsPath, "debugfs", debugFsTracepointPath)
```
FedeDP: Do we want a more explicit "ehy we will remove this in the near future" warning?

cmd/tetragon/main.go
```diff
@@ -297,6 +297,7 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 	checkProcFS()
 	// Setup file system mounts
 	bpf.CheckOrMountFS("")
+	bpf.CheckOrMountTraceFS()
 	bpf.CheckOrMountDebugFS()
```
kkourt: How about we don't mount both filesystems?

Would it make sense to have `CheckOrMountTraceFS()` that first tries tracefs, and then debugfs. It then sets a private value with the tracefs path and an error so that we can have something like `GetTraceFSPath() (string, error)` for users.

FedeDP: It makes sense! I was thinking that perpahs we used debugfs for other things too, but cannot find any reference in the code.  Should be safe to kill!

pkg/tracepoint/tracepoint.go
kkourt: A good rule of thumb is to make warnings actionable. For example, we can have something that encourages users to let us know if this warning is triggered in a GH issue.

pkg/bpf/bpffs_linux.go
```diff
@@ -111,7 +113,26 @@ func checkOrMountDebugFSDefaultLocations() error {
 		return mountFS(debugFSRoot, mountinfo.FilesystemTypeDebugFS)
 	}
 	if !debugfsInstance {
-		return errors.New("instance exists with othe type")
+		return errors.New("instance exists with other type")
```
FedeDP, line 117: `othe` -> `other`

PR #4263, opened 2025-10-28
tetragon: improve ring reliability in tests

kevsecurity:
Ensures tests that interact with the ring buffer directly receive all events and don't stop prematurely. See commits for details.


pkg/testutils/perfring/perfring.go
```diff
@@ -184,16 +184,26 @@ func ProcessEvents(t *testing.T, ctx context.Context, eventFn EventFn, wgStarted
 		}()
 	}
 
+	complChanCount := 0
+	maxComplChan := 1
+	if useBPFRingBuffer {
+		maxComplChan = 2
```
kkourt, line 190: question(non-blocking): are there tests that use both ring buffers? Shouldn't be one or the other?

kkourt:
I believe we need to backport this to 1.6. Please remove the label if that's not the case.

PR #4265, opened 2025-10-29
nok8s: build without k8s

kkourt:
This PR introduces a `nok8s` build tag that allows building Tetragon with no k8s support, to reduce the image size. 

The reduction in the tetragon binary is shown below.
59M	tetragon
33M	tetragon-nok8s

### DIFF
--- a/.github/workflows/static-checks.yaml
+++ b/.github/workflows/static-checks.yaml
@@ -152,6 +152,17 @@ jobs:
           GOOS=windows GOARCH=amd64 make tetragon
           GOOS=windows GOARCH=arm64 make tetragon
 
+  build-nok8s:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Build tetragon Windows binaries
+        run: |
+          GOARCH=amd64 make tetragon-nok8s tetra-nok8s
+          GOARCH=arm64 make tetragon-nok8s tetra-nok8s
+
   errmetrics-fileids-check:
     runs-on: ubuntu-latest
     steps:


mtardy:
<details><summary>Made me think of these kind of memes</summary>

image

</details>

pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -93,5 +95,4 @@ func TestEmptyTracingPolicy(t *testing.T) {
 	path := tempfile.CreateTempFile(t, "")
 	_, err := FromFile(path)
 	require.Error(t, err)
-	assert.Contains(t, err.Error(), "unknown CRD kind: ")
```
mtardy, line 96: why remove that?

pkg/tracingpolicy/nok8s.go
```diff
@@ -0,0 +1,35 @@
...
+	case v1alpha1.TPKindDefinition:
+		var gtp GenericTracingPolicy
+		if err := json.Unmarshal(jsonBytes, &gtp); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal TracingPolicy: %w", err)
+		}
+		return &gtp, nil
+	case v1alpha1.TPNamespacedKindDefinition:
+		return nil, fmt.Errorf("namespaced tracing policies not supported in non-k8s builds: %s", kind)
+	default:
+		return nil, fmt.Errorf("unknown kind: %s", kind)
```
mtardy, line 33: maybe you can keep the same error so that the test still works on both versions `return nil, fmt.Errorf("unknown CRD kind: %s", unstr.GetKind())`?

pkg/tracingpolicy/yaml_test.go
mtardy, line 1: why the indent needed to change in this file? It's not valid with the new parser?

pkg/cgtracker/test/cgtracker_test.go
```diff
@@ -214,6 +215,11 @@ func doProgTest(t *testing.T, cgfsPath string) {
 
 // TestCgTrackerPolicyFilter checks that cgroup tracking works with policyfilter
 func TestCgTrackerPolicyFilter(t *testing.T) {
+
```
mtardy: nit: whiteline

pkg/cgtracker/test/cgtracker_test.go
```diff
@@ -214,6 +215,11 @@ func doProgTest(t *testing.T, cgfsPath string) {
...
+	if !build.K8sEnabled() {
+		t.Skip("non-k8s build: skipping policy filter")
```
mtardy: since this string is used many many times, it would have been nice to have it in `build` pkg maybe? or bundle even the whole

```go
if !build.K8sEnabled() {
		t.Skip("non-k8s build: skipping policy filter")
}
```

in some func like SkipK8sDisabled?

pkg/policyfilter/nok8s.go
```diff
@@ -0,0 +1,55 @@
...
+	glblState   State
+	glblError   error // nolint:errname
+	setGlobalPf sync.Once
+)
+
+// GetState returns global state for policyfilter
+func GetState() (State, error) {
+	setGlobalPf.Do(func() {
+		if option.Config.EnablePolicyFilter {
+			glblError = fmt.Errorf("policyfilter is not suported in non-k8s build")
```
mtardy, line 30: as well we have many places where we have this "unsupported in non k8s build", could we use the same error everywhere like we do for windows? Could be in your build package also. You could enrich the error with something else with fmt.Errorf but this is still detectable by callers. Not sure it's super useful for callers but would make it easy to see everywhere where we have something that is unsupported because of the build (like for windows).

cmd/tetra/common/commands.go
```diff
@@ -0,0 +1,18 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package common
+
+import (
+	"github.com/spf13/cobra"
+)
+
+func AddSubCommandIfNotNil(parent *cobra.Command, cmds ...*cobra.Command) {
```
mtardy, line 10: From what you did, does AddCommand panics when you add a nil command? Just curious

pkg/tracingpolicy/generictracingpolicy_test.go
kkourt: Because the error string is different in the two cases. I think it's enough to check that there is an error.

pkg/tracingpolicy/nok8s.go
kkourt: The reason I didn't do it, because the error "unknown CRD kind" is misleading. There is no CRD in this case.

pkg/tracingpolicy/yaml_test.go
kkourt: Yes, because we changed the yaml library the output is different.

cmd/tetra/common/commands.go
kkourt: yes

pkg/policyfilter/nok8s.go
kkourt: I would like to be able to grep for error lines, and I think what you suggest breaks that.

pkg/watcher/pod_nok8s.go
```diff
@@ -0,0 +1,41 @@
...
+
+//go:build nok8s
+
+package watcher
+
+import (
+	"errors"
+)
+
+var nonK8sErr = errors.New("non k8s build")
```
mtardy, line 12: might be nice to have that into your new build pkg to be reused

cmd/tetra/commands.go
```diff
@@ -25,7 +26,7 @@ func addBaseCommands(rootCmd *cobra.Command) {
 	rootCmd.AddCommand(sensors.New())
 	rootCmd.AddCommand(status.New())
 	rootCmd.AddCommand(rthooks.New())
-	rootCmd.AddCommand(explain.New())
```
mtardy, line 28: tbh if this adds 15M in general, we could just remove it altogether, I don't really mind and it doesn't feel like it's worth 15M

pkg/policyfilter/nok8s.go
mtardy: I still think that sharing the same error would make sense and then you can just grep for `build.NoK8sError` or whatever, or you can just remove it and see where it breaks. basically a bit like what you did with `var nonK8sErr = errors.New("non k8s build")` already here. just a nit, feel free to ignore and merge like that we can always change that.

PR #4266, opened 2025-10-29
new: retprobes events cache size option

FedeDP:
### Description

In #4193 we noticed that the default size for the lru cache used by kretprobes and uretprobes is very large (4096).
This PR moves the default to 64, and adds an option key to set it to a desired value.

Context: https://github.com/cilium/tetragon/pull/4193#discussion_r2470185219

### Changelog

```release-note
new: retprobes events cache size option
```

### DIFF
--- a/docs/data/tetragon_flags.yaml
+++ b/docs/data/tetragon_flags.yaml
@@ -226,6 +226,9 @@ options:
       default_value: "true"
       usage: |
         Release all pinned BPF programs and maps in Tetragon BPF directory. Enabled by default. Set to false to disable
+    - name: retprobes-cache-size
+      default_value: "4096"
+      usage: Set {k,u}retprobes events cache maximum size
     - name: server-address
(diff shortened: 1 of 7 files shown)


FedeDP:
No prob!

FedeDP:
cc @olsajiri we discussed this one in #4193 !

olsajiri:
> cc @olsajiri we discussed this one in #4193 !

IIUC the new value is 64, could we back it up with some tests? or we just think 4096 is too much?

FedeDP:
perhaps we could just half it and see what happens? ;-) @kkourt 

also perhaps in future we could need separate options for uprobe a kprobe, but not sure it's common to have them configured on the same side

FedeDP:
Moved back default to 4096.

FedeDP:
Rebased on top of latest main that should fix ci.

PR #4267, opened 2025-10-29
tetragon: kprobe/uprobe registers data retrieval

olsajiri:
adding support to:
- work with all register sizes in override regs
- pt_regs data source to retrieve register as data

### DIFF
--- a/pkg/sensors/tracing/kprobe_amd64_test.go
+++ b/pkg/sensors/tracing/kprobe_amd64_test.go
@@ -7,6 +7,7 @@ package tracing
 
 import (
 	"context"
+	"os"
 	"os/exec"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


bpf/process/uprobe_offload.h
```diff
@@ -119,15 +131,20 @@ write_reg(struct pt_regs *ctx, __u32 dst, __u64 val)
...
+
 	/* Using inlined asm for same reason we use WRITE_REG above. */
 #define READ_REG(reg) ({                                        \
 	__u64 val;                                              \
 	asm volatile("%[val] = *(u64 *)(%[ctx] + %[off])\n"     \
 		     : [ctx] "+r"(ctx), [val] "+r"(val)         \
 		     : [off] "i"(offsetof(struct pt_regs, reg)) \
 		     :);                                        \
+	val <<= shift;                                          \
+	val >>= shift;                                          \
```
kkourt, line 147: question: I guess that's to zero the rest of the bits?

bpf/process/generic_calls.h
```diff
@@ -449,6 +454,95 @@ FUNC_INLINE int arg_idx(int index)
...
+	probe_read(&val, sizeof(val), _ctx + offset);
+	val <<= shift;
+	val >>= shift;
+	return val;
+}
+
+// TODO let's unite this with read_reg in bpf/process/uprobe_offload.h
+#if defined(__TARGET_ARCH_x86) && (defined GENERIC_KPROBE || defined GENERIC_UPROBE)
+FUNC_INLINE long get_pt_regs_arg(struct pt_regs *ctx, struct event_config *config, int index)
+{
```
kkourt, line 475: For kprobes: do I understand correctly that the above applies to the user-space registers?
What happens if there is no user context when we execute the kprobe?

bpf/process/uprobe_offload.h
olsajiri: yes, the shift is set based on the value size, we do the same thing for usdt arguments

bpf/process/generic_calls.h
olsajiri: no for kprobes it's pt_regs at the point the probe was hit, which is bit different for kprobe through int3 and kprobe through fprobe/ftrace, but it's still inside kernel

PR #4280, opened 2025-10-31
exporter: fix flaky Test_rateLimitExport with polling

### LINKED ISSUE #2789: tests: `pkg.sensors.tracing.TestKprobeRateLimit` is flaky
https://github.com/cilium/tetragon/actions/runs/10316714693/job/28559512984?pr=2787

[tetragon-bugtool.zip](https://github.com/user-attachments/files/16562376/tetragon-bugtool.zip)
[tetragon-json.zip](https://github.com/user-attachments/files/16562378/tetragon-json.zip)

calghar:
## Summary
This PR fixes the flaky Test_rateLimitExport test by replacing the fixed sleep duration with a polling-based synchronization mechanism, addressing the timing race condition reported in #2789.

## Related Issue
Fixes #2789

## Root Cause
The test used a fixed 200ms sleep which allowed multiple rate limiter ticker intervals (50ms each) to fire during the wait period. This created a timing window where:
- Multiple tickers could emit rate-limit-info messages if events were still processing
- A race condition at ticker boundaries could cause off-by-one errors in event counts
- No synchronization existed between "events sent" and "events fully processed"

## Proposed Changes
- **Added countEvents() helper function**: Non-blocking function to count events and rate-limit-info messages without assertions
- **Replaced fixed sleep with polling**: Poll every 10ms until expected number of events and rate-limit-info messages are received
- **Added timeout with clear diagnostics**: 500ms timeout (2.5× original sleep) with descriptive error message showing actual vs. expected counts

## Testing Performed
- Test passes 20 consecutive runs without failures (verified via Docker with golang:1.25)
- No performance degradation - typically should complete faster than the original
- Deterministic behavior regardless of system timing or load

## Backward Compatibility
No breaking changes. The fix only modifies the test implementation, not the rate limiter functionality itself.

### DIFF
--- a/pkg/exporter/exporter_test.go
+++ b/pkg/exporter/exporter_test.go
@@ -120,6 +120,35 @@ type jsonEvent struct {
 
 const nodeName = "test-node-name"
 
+// countEvents counts the number of events and rate-limit-info messages in the given slice.
+// It returns (gotEvents, gotRateLimitInfo, gotDropped).
+func countEvents(eventsJSON []string) (int, int, uint64) {
+	gotEvents, gotRateLimitInfo, gotDropped := 0, 0, uint64(0)
+	for _, event := range eventsJSON {
+		if event == "" {
+			continue
+		}
+
+		var ev jsonEvent
+		if err := json.Unmarshal([]byte(event), &ev); err != nil {
+			continue
+		}
+
+		if len(ev.Event) > 0 {
+			gotEvents++
+		}
+		if len(ev.RateLimitInfo) > 0 {
+			var res tetragon.GetEventsResponse
+			if err := json.Unmarshal([]byte(event), &res); err != nil {
+				continue
+			}
+			gotRateLimitInfo++
+			gotDropped += res.GetRateLimitInfo().GetNumberOfDroppedProcessEvents()
+		}
+	}
+	return gotEvents, gotRateLimitInfo, gotDropped
+}
+
 func checkEvents(t *testing.T, eventsJSON []string, wantEvents, wantRateLimitInfo int, wantDropped uint64) {
 	t.Helper()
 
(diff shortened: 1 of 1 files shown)


calghar:
This PR fixes a flaky test with no user-facing changes. Could a maintainer please add the `release-note/misc` label? Thanks!

FedeDP (approved):
LGTM, thanks!

pkg/exporter/exporter_test.go
```diff
@@ -120,6 +120,35 @@ type jsonEvent struct {
 
 const nodeName = "test-node-name"
 
+// countEvents counts the number of events and rate-limit-info messages in the given slice.
+// It returns (gotEvents, gotRateLimitInfo, gotDropped).
+func countEvents(eventsJSON []string) (int, int, uint64) {
+	gotEvents, gotRateLimitInfo, gotDropped := 0, 0, uint64(0)
```
mtardy: I'd used the named return parameters feature if you are anyway going to do it
```suggestion
func countEvents(eventsJSON []string) (gotEvents int, gotRateLimitInfo int, gotDropped uint64) {
```

pkg/exporter/exporter_test.go
```diff
@@ -214,9 +243,28 @@ func Test_rateLimitExport(t *testing.T) {
...
+			timeout := time.After(500 * time.Millisecond)
+			ticker := time.NewTicker(10 * time.Millisecond)
+			defer ticker.Stop()
+
+		pollLoop:
+			for {
+				gotEvents, gotRateLimitInfo, _ := countEvents(results.items)
+				if gotEvents >= tt.wantEvents && gotRateLimitInfo >= tt.wantRateLimitInfo {
+					// We have all the expected output, proceed to cleanup and assertions.
+					break pollLoop
```
mtardy: I don't think you need to name it in your case?
```suggestion
			for {
				gotEvents, gotRateLimitInfo, _ := countEvents(results.items)
				if gotEvents >= tt.wantEvents && gotRateLimitInfo >= tt.wantRateLimitInfo {
					// We have all the expected output, proceed to cleanup and assertions.
					break
```

mtardy (approved):
lgtm as well, only a few golang nits. Will merge this later, give you some time to update the patch or not depending on what you prefer! thanks

calghar:
> lgtm as well, only a few golang nits. Will merge this later, give you some time to update the patch or not depending on what you prefer! thanks

Sorry for the late reply. Was on vacation. Have incorporated your suggestions 🙂

Issue #4284, opened 2025-11-03
tests: `TestModeEnforcer` started failing on arm64 main: override action is broken on recent arm64 kernels


mtardy:
It just started failing between less than 6 hours ago and 10/20min. I first found out on an unrelated PR https://github.com/cilium/tetragon/pull/4275 and saw it was failing on main.

This runner might have various underlying changes since this is an arm64 runner provided by GitHub with a non-fixed kernel version and environment. We don't have vmtests with stable kernels on arm64.

This is not a flake though and seems to consistently fail

https://github.com/cilium/tetragon/actions/runs/19039330941/job/54371958716#step:9:168

cc @kkourt

mtardy:
So taking a look at this I have what happens is this:

```
    logcapture.go:24: time=2025-11-03T16:54:36.223Z level=INFO msg="Loaded sensor successfully" sensor=generic_kprobe
    mode_test.go:176: command getcpu out:"getcpu returned: err:errno 0" err:<nil>
    mode_test.go:195: 
        	Error Trace:	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/policyconf/test/mode_test.go:195
        	            				/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/policyconf/test/mode_test.go:228
        	Error:      	"getcpu returned: err:errno 0" does not contain "operation not permitted"
        	Test:       	TestModeEnforcer
```

And the test is doing something like this
```
	checkEnforce()
	resetEnforcerMap()
	policyconf.SetPolicyMode(tp, policyconf.MonitorMode)
	checkMonitor()
	resetEnforcerMap()
	policyconf.SetPolicyMode(tp, policyconf.EnforceMode)
	checkEnforce()
```

So the first enforce seems to not work and let everything pass. I couldn't repro this on arm64 6.8.0 (the test passes) but the GitHub runner now runs `Linux runnervmwmpaq 6.14.0-1012-azure #12~24.04.1-Ubuntu SMP Tue Sep 9 19:15:52 UTC 2025 aarch64 aarch64 aarch64 GNU/Linux`, so I tried on `bpf-next-20251103.013252` and could repro the issue. So something has changed between 6.8 and 6.14 that makes the enforcer mode fail it seems.

mtardy:
Ah this is a broader problem than the enforcer, it seems that the `override` action is broken on arm64 more generally.

I tried this with `sys_getcpu`
[...]

PR #4285, opened 2025-11-03
helm: Handle tetragon.enabled value properly

michi-covalent:
Don't create Tetragon daemonset/configmap/service if tetragon.enabled is set to false. This allows users to manage CRDs using Tetragon Helm chart without running Tetragon agent/operator by setting crds.installMethod to "helm" and disabling everything else. For example:

    % helm template tetragon ./install/kubernetes/tetragon \
      --set crds.installMethod=helm \
      --set tetragon.enabled=false \
      --set tetragonOperator.enabled=false \
      --set serviceAccount.create=false

### DIFF
--- a/install/kubernetes/tetragon/templates/daemonset.yaml
+++ b/install/kubernetes/tetragon/templates/daemonset.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.tetragon.enabled }}
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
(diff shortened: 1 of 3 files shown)


sayboras (approved):
Makes sense to me ✅

michi-covalent:
https://github.com/cilium/tetragon/actions/runs/19051826722/job/54415152138?pr=4285 is hitting https://github.com/cilium/tetragon/issues/4284

michi-covalent:
added `needs-backport/1.6` label. without this patch `tetragon.enabled=false` can generate invalid k8s manifests, so this can be considered a bugfix 🔧

PR #4287, opened 2025-11-04
bug: Avoid overwritten slogLeveler

sayboras:
This commit is to use SetLogLevel func to set the log level, instead of overwrite opts.Lelvel, which overwrites slogLeveler function i.e. we can't change the log level later during runtime.

Fixes: 09226dd9116da81b29ddede896f7827f7a1a62af

```release-note
bug: Avoid overwritten slogLeveler
```


michi-covalent:
https://github.com/cilium/tetragon/actions/runs/19067742913/job/54522306256?pr=4287 is a known failure: https://github.com/cilium/tetragon/issues/4284. merging.

PR #4288, opened 2025-11-04
fix(docs): correct apiVersion and kind casing in TracingPolicy examples

littlejo:
Fixes 

### Changelog

```release-note
```


mtardy (commented):
Thank you, , it's a relatively straightforward patch but could you be more explicit in the commit title on what your fix does for git history?

mtardy (approved):

PR #4289, opened 2025-11-04
new(pkg/sensors): add a `ListCollections` method on sensors Manager.

FedeDP:
### Description
While not exposed via grpc, it can be used internally to access the list of Collection (deep copied to avoid sync issues).

### DIFF
--- a/pkg/sensors/collection.go
+++ b/pkg/sensors/collection.go
@@ -11,6 +11,8 @@ import (
 
 	"go.uber.org/multierr"
 
+	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
+
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/policyconf"
@@ -60,6 +62,18 @@ func (ck *collectionKey) String() string {
 		return fmt.Sprintf("%s/%s", ck.namespace, ck.name)
 	}
 	return ck.name
+}
+
+// Exposed via ListCollections()
+type Collection struct {
+	Name              string
+	TracingpolicyName string
+	TracingpolicySpec *v1alpha1.TracingPolicySpec
+	TracingpolicyMode tetragon.TracingPolicyMode
+	TracingpolicyID   uint64
+	PolicyfilterID    uint64
+	State             TracingPolicyState
+	Err               string
 }
 
 // collection is a collection of sensors
--- a/pkg/sensors/handler.go
+++ b/pkg/sensors/handler.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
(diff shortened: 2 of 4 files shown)


FedeDP:
cc @kkourt

pkg/sensors/collection.go
```diff
@@ -62,6 +64,18 @@ func (ck *collectionKey) String() string {
 	return ck.name
 }
 
+// Exposed via ListCollections()
+type Collection struct {
+	Name              string
+	TracingpolicyName string
+	TracingpolicySpec *v1alpha1.TracingPolicySpec
```
kkourt, line 71: How about we do something like:

```
type Collection {
      tracingpolicy tracingpolicy.TracingPolicy
}
```

And then:
```
func (c *Collection) Spec() *v1alpha1.TracingPolicySpec {
   // deepcopy goes here
} 
```

My thinking is that we would be able to implement `listPolicies` using `listCollections` without having to do an unnecessary copy. The copy would happen only if a caller requested the spec.

WDYT?

pkg/sensors/handler.go
```diff
@@ -427,6 +429,43 @@ func (h *handler) listPolicies() []*tetragon.TracingPolicyStatus {
 	return ret
 }
 
+func (h *handler) listCollections(policyOnly bool) []*Collection {
```
kkourt, line 432: thought (not-blocking): Should we pass a closure here to keep things more flexible, or is it an overkill?

Edit: Probably an overkill, and we can always filter at the caller.

kkourt (commented):
Thanks!

pkg/sensors/handler.go
FedeDP: Also since this is not exposed via grpc, we can always update it later if we need to!

pkg/sensors/collection.go
FedeDP: Neat idea! We reduce code duplication a bit :)

FedeDP: So, on a second thought, it's not that easy because:
* if we want `listPolicies` to use `listCollections` we would have a double lock, one for `listCollections` and one while iterating on the returned collections to build the `TracingPolicyStatus` slice, since we access fields like `col.sensors` that can be updated in the meantime
* there is no easy way to deep copy the sensors slice since it is a slice of interfaces
* `Collection` and `collection` overlap a lot (ie: from `listCollections` it would be good to return not only the `tracingpolicy` field but more, like i did in this PR)  and should probably be merged into a single type

Regarding the last bullet point, at that point it would be better to let the currentyl existing `collection` type implement a `Collectioner` interface like:
```go
type Collectioner interface {
    Spec() *v1alpha1.TracingPolicySpec
    Mode() tetragon.TracingPolicyMode
}
```
and more, if needed, and do not even touch `collection` struct, but let the world calling `ListCollections` expect it to return a `Collectioner`.

kkourt: So let's keep the original approach then?

kkourt: That is, have  a separate function and always copy. Or is there something better?

FedeDP: yes i'd go with this at the moment. We can always revisit it later (avoiding breaking changes of course)!

PR #4290, opened 2025-11-04
tetragon: arm fixes

olsajiri:
fix/workaround kprobe multi related tests on arm

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -5244,11 +5244,20 @@ func TestKprobeKernelStackTrace(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
 	defer cancel()
 
+	disableKprobeMulti := ""
+
+	if isArm() {
+		disableKprobeMulti = `
+  options:
+    - name: "disable-kprobe-multi"
+      value: "1"`
+	}
+
 	tracingPolicy := `apiVersion: cilium.io/v1alpha1
 kind: TracingPolicy
 metadata:
   name: uname
-spec:
+spec: ` + disableKprobeMulti + `
   kprobes:
     - call: sys_newuname
       selectors:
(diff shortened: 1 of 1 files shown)


pkg/sensors/tracing/consts.go
```diff
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of Tetragon
 
+//go:build !windows
```
mtardy, line 4: nit: you can use `_linux.go` postfix in Go files so that it only build on linux and is pretty explicit, for example here it would be `pkg/sensors/tracing/consts_linux.go`.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -610,6 +611,10 @@ func hasMapsSetup(spec *v1alpha1.TracingPolicySpec) hasMaps {
 	return has
 }
 
+func isArm() bool {
+	return runtime.GOARCH == "arm64"
+}
```
mtardy, line 616: nit: I would inline this `==` equal check where you use it or put it in `pkg/arch` under `IsARM64` or something like that. We have some places in the code where we just check `runtime.GOARCH == "arm64"`, I think it's simple enough to inline.

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -5244,11 +5244,20 @@ func TestKprobeKernelStackTrace(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
 	defer cancel()
 
+	disableKprobeMulti := ""
+
+	if isArm() {
```
mtardy, line 5249: ah I see you use it there as well

kkourt:
Thanks! Marked for backport.

pkg/sensors/tracing/consts.go
olsajiri: hum, is there preferred way? I think we have other examples of this.. I'd prefer to decide what way is better and we could convert all the sites

mtardy:
> Thanks! Marked for backport.

fyi it needs backports in all stable branches @olsajiri

PR #4291, opened 2025-11-04
tetragon: handle test events sequentially

kevsecurity:
The test infrastructure that monitors the ring buffers and handles events during tests does not mirror how tetragon does this in production. When monitoring events for real, we have an event cache channel that the ring buffer monitors write to, and the event handler reads from. This means that events are handled sequentially.

In the test infrastructure, the ring buffer monitors handle the events directly, resulting in two events able to be handle concurrently, which causes problems. This commit introduces an event cache to the test infrastructure and uses that to handle events sequentially, avoiding the problem.

### DIFF
--- a/pkg/testutils/perfring/perfring.go
+++ b/pkg/testutils/perfring/perfring.go
@@ -100,14 +100,27 @@ func ProcessEvents(t *testing.T, ctx context.Context, eventFn EventFn, wgStarted
 
 	wgStarted.Done()
 
-	errChan := make(chan error)
+	errChan := make(chan error, 2)
 	defer close(errChan)
 
-	complChan := make(chan bool)
+	complChan := make(chan bool, 2)
 	defer close(complChan)
+
+	// Create an events queue that both the ring buffers can send to.
+	// This means we can handle the events sequentially by reading from the
+	// queue, avoiding issues of handling events concurrently.
+	type testEvent struct {
+		RawSample    *[]byte
+		ComplChecker *testsensor.CompletionChecker
+		Ctx          context.Context
+		CancelFunc   context.CancelFunc
+	}
+
+	eventsQueue := make(chan *testEvent, 65536) // arbitrary size for tests
 
 	var wg sync.WaitGroup
 	wg.Add(1)
+	ctxPerfRing, cancelPerfRing := context.WithCancel(ctx)
 	defer wg.Wait()
 
 	go func() {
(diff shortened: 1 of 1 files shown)


kevsecurity:
ARM failure is a known issue.

pkg/testutils/perfring/perfring.go
```diff
@@ -100,83 +100,77 @@ func ProcessEvents(t *testing.T, ctx context.Context, eventFn EventFn, wgStarted
...
-				if handlerErr != nil {
-					errChan <- fmt.Errorf("error handling perfring data: %w", handlerErr)
-					break
-				}
-				err = loopEvents(events, eventFn, complChecker)
-				if err != nil {
-					errChan <- fmt.Errorf("error loop event function returned: %w", err)
-					break
-				}
+			if complChecker.Done() || ctxPerfRing.Err() != nil {
```
kkourt, line 141: What happens if there is a `ctxPerfRing` error? Do we propagate it to the test  somehow?
The benefit of having `t *testing.T` is that we can just fail the test, but we do not seem to be doing that.

pkg/testutils/perfring/perfring.go
```diff
@@ -100,83 +100,77 @@ func ProcessEvents(t *testing.T, ctx context.Context, eventFn EventFn, wgStarted
...
-					if handlerErr != nil {
-						errChan <- fmt.Errorf("error handling ringbuf data: %w", handlerErr)
-						break
-					}
-					err = loopEvents(events, eventFn, complChecker)
-					if err != nil {
-						errChan <- fmt.Errorf("error loop event function returned: %w", err)
-						break
-					}
+				if complChecker.Done() || ctxBPFRing.Err() != nil {
```
kkourt, line 168: Similarly for this error here.

pkg/testutils/perfring/perfring.go
kevsecurity: We use ctxPerfRing only to indicate that we are done. I don't think we need to propagate any errors. What errors did you have in mind?

kkourt: > || ctxPerfRing.Err() != nil

Above is an error, no?

kevsecurity: > Above is an error, no?

Isn't this the context has been closed elsewhere? What errors do contexts throw?

Ultimately, it is in a test, and the test will (I assume) fail if an error (other than context is done) has occurred in servicing the ring buffer. I don't really see the point in catching and handling hypothetical errors in tests, unless the tests are failing.

kkourt: Why do you say the errors are hypothetical?

kevsecurity: > Why do you say the errors are hypothetical?

According to https://pkg.go.dev/context, the only errors thrown by the context are Cancelled and DeadlineExceeded. Neither of these are actual errors; they are signals that the context has closed.

kkourt: OK, but why are they hypothetical?

> 	ctxPerfRing, cancelPerfRing := context.WithCancel(ctx)

Either the cancel function is used, so it's not hypothetical or the cancel function is not needed so we shouldn't use it.

kevsecurity: Okay, this is starting to get philosophical. Both Cancelled and DeadlineExceeded are circumstances that are expected, and by that nature IMHO are not "errors" in the sense that no error has occurred. They are simply signals. As such, I referred to them as "hypothetical errors" because they are not errors, but are called errors by the language.

The process involves threads. When one thread detects that the process has finished, it calls the cancel functions on the contexts that the other threads run within. This signals to those threads that the process is done.

Therefore, I believe we need the cancel function; but the result of the cancel function being called isn't a real error (it's a signal) but is returned as an error type in golang; this error type is expected and therefore does not need to be propagated. As the context won't (according to the docs) throw any other errors, we don't need to check the error; we can just accept that the context has completed.

Happy to discuss this offline btw.

kkourt: I did not intend for a philosophical discussion :)

> Therefore, I believe we need the cancel function; but the result of the cancel function being called isn't a real error (it's a signal) but is returned as an error type in golang; this error type is expected and therefore does not need to be propagated. As the context won't (according to the docs) throw any other errors, we don't need to check the error; we can just accept that the context has completed.

I see -- makes sense.

Why do we need the context, though? Can we just call `.Close()` from the event handling loop? (We seem to be already doing so).

The current code seems to be:
```go
		wg.Go(func() {

			complChecker := testsensor.NewCompletionChecker()

			for ctxBPFRing.Err() == nil {

				record, err := ringBufReader.Read()
				if err != nil {
					if ctxBPFRing.Err() == nil && !errors.Is(err, os.ErrClosed) {
						errChan <- fmt.Errorf("error reading ringbuf data: %w", err)
					}
					break
				}

				if complChecker.Done() || ctxBPFRing.Err() != nil {
					break
				}

				if len(record.RawSample) > 0 {
					eventsQueue <- &testEvent{&record.RawSample, complChecker, ctxBPFRing, cancelBPFRing}
				}
			}
		})
```

Can't we just do something like:
```go

		wg.Go(func() {

			complChecker := testsensor.NewCompletionChecker()

			for {

				record, err := ringBufReader.Read()
				if err != nil {
					if !errors.Is(err, os.ErrClosed) {
						errChan <- fmt.Errorf("error reading ringbuf data: %w", err)
					}
                    break
				}

				if len(record.RawSample) > 0 {
					eventsQueue <- &testEvent{&record.RawSample, complChecker, ctxBPFRing, cancelBPFRing}
				}
			}
		})
```

I'm not saying that the code will not work as it is, but I feel like there is room for simplification to make it easier to understand.

kevsecurity: My patch extends the existing logic; the previous version (with a single ring buffer) used the context to stop the read loop, so I did the same but with two derived contexts, one for each ring buffer. If closing the ring buffer is suitable, then I wonder why that was rejected as a simplification in the original code? Perhaps it would introduce unintended consequences? I'd rather merge as is, given it works and is an evolution of the existing code.

kevsecurity: Also, this is a test util; how perfect does this need to be? It feels like a lot of cycles are being used in this discussion that could be better used elsewhere.

kkourt: > If closing the ring buffer is suitable, then I wonder why that was rejected as a simplification in the original code? 

I _think_ it's because we needed to deal with a timeout: https://github.com/cilium/tetragon/commit/92a2503f7788e3707f61c99cd2751667f9b326cc.  We operated on the `ctx` passed by the user, while in your commit you pass one context on each ring buffer goroutine.

> Also, this is a test util; how perfect does this need to be? It feels like a lot of cycles are being used in this discussion that could be better used elsewhere.

It does not have to be perfect, but because testing is shared infrastructure, I think it's worth getting it as understandable as we can. If another test fails in the future, it would be nice to understand why. My opinion as a reviewer (who might have to deal with this code in the future) is that it is difficult to understand,  and I believe it can be simplified (offered concrete suggestions).

kevsecurity: Thanks for the context, but I'm struggling to understand how "as understandable as we can" is not the definition of "perfect" in this example?

The contexts provided to the rings are derived from the user supplied one, meaning cancelling the user supplied one should also cancel these routines (as it originally did).

This patch fixes a real bug that results in a panic. I don't have a great deal of cycles to attend to it, so I think it would make sense to merge as-is (given the debate appears to be entirely about style than functionality now) and create an issue for a tidy-up when someone has cycles.

kkourt: > Thanks for the context, but I'm struggling to understand how "as understandable as we can" is not the definition of "perfect" in this example?
> 
> The contexts provided to the rings are derived from the user supplied one, meaning cancelling the user supplied one should also cancel these routines (as it originally did).
> 
> This patch fixes a real bug that results in a panic. I don't have a great deal of cycles to attend to it, so I think it would make sense to merge as-is (given the debate appears to be entirely about style than functionality now) and create an issue for a tidy-up when someone has cycles.
> 

I'm not going for perfect. One observation I had that would improve the code, but I did not request to expedite merging of the PR, was that the two goroutines  for reading the ring buffers could be implemented via a single function to follow DRY. 

It's also not a stylistic change. A style change would be to request to use the same approach in both goroutines:

```
	go func() {
		defer wg.Done()
```

```
		wg.Go(func() {
``` 

Which I also did not request in the interest of expediency.

To move things forward, and since this addresses an actual issue, feel free to merge the PR if you want (Federico has ACKed it in any case).

PR #4292, opened 2025-11-04
helm: Handle tetragon.enabled value properly

michi-covalent:
[ upstream commit e3a065383068f05455d4af835efd56f73baf66ea ]

Don't create Tetragon daemonset/configmap/service if tetragon.enabled is set to false. This allows users to manage CRDs using Tetragon Helm chart without running Tetragon agent/operator by setting crds.installMethod to "helm" and disabling everything else. For example:

    % helm template tetragon ./install/kubernetes/tetragon \
      --set crds.installMethod=helm \
      --set tetragon.enabled=false \
      --set tetragonOperator.enabled=false \
      --set serviceAccount.create=false

### DIFF
--- a/install/kubernetes/tetragon/templates/daemonset.yaml
+++ b/install/kubernetes/tetragon/templates/daemonset.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.tetragon.enabled }}
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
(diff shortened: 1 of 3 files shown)


michi-covalent:
https://github.com/cilium/tetragon/actions/runs/19077838575/job/54498200138?pr=4292 is a known failure: https://github.com/cilium/tetragon/issues/4284

PR #4296, opened 2025-11-05
uprobe: support match args

andrewstrohman:
### Description
This change allows filtering uprobe events via matchArgs selector.

### DIFF
--- a/pkg/sensors/tracing/uprobe_test.go
+++ b/pkg/sensors/tracing/uprobe_test.go
@@ -832,3 +832,70 @@ func TestUprobeArgsWithAddress(t *testing.T) {
 
 	testUprobeArgs(t, checkers, tp)
 }
+
+func uprobeArgsMatch(t *testing.T, argOneVal int, expectCheckerFailure bool) error {
+	uprobeTest1 := testutils.RepoRootPath("contrib/tester-progs/uprobe-test-1")
+	libUprobe := testutils.RepoRootPath("contrib/tester-progs/libuprobe.so")
+
+	pathHook := `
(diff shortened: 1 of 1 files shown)


pkg/sensors/tracing/uprobe_test.go
```diff
@@ -832,3 +832,70 @@ func TestUprobeArgsWithAddress(t *testing.T) {
...
+}
+
+func TestUprobeArgsMatch(t *testing.T) {
+	err := uprobeArgsMatch(t, 123)
+	require.NoError(t, err)
+}
+
+func TestUprobeArgsMatchNot(t *testing.T) {
+	err := uprobeArgsMatch(t, 124)
+	require.Error(t, err)
```
kkourt: Having negative tests is great; thanks for adding them!

The proper way to do that, however, is use `jsonchecker.JsonTestExpect` with a `true` argument. 


See: https://github.com/cilium/tetragon/blob/83b5251fffeba767fdf1c4f78a79414de8ed897d/pkg/jsonchecker/jsonchecker.go#L210-L212


The reason is that jsonchecker does some tricks to keep the events file when there is a failure, so it needs to understand what failure is. My guess is that this comes from copy-paste existing, since I've seen this before, so I created: https://github.com/cilium/tetragon/issues/4297.

kkourt (changes requested):
Thanks, just one requested change.

pkg/sensors/tracing/uprobe_test.go
andrewstrohman: Thanks. I've corrected this.

PR #4313, opened 2025-11-07
tetragon: Display error when usdt probe is not found in binary

olsajiri:
Currently we won't hit error when there's not configured usdt probe in the monitored binary, adding it:
```
level=error msg="Failed to execute tetragon" error="policy handler 'tracing' \
                 failed loading policy 'usdts': failed to configure usdt 'tetragon/test', \
                 not found in the binary: '/home/jolsa/tetragon/contrib/tester-progs/usdt-override'"
```


kkourt:
First patch was backported in https://github.com/cilium/tetragon/pull/4318.

PR #4316, opened 2025-11-07
new: monitor only mode

FedeDP:
### Description

Today, we default at `enforce` even for policies that do not support any enforcement.
Introduce a `monitor_only` mode that:
* is set to ebpf map during startup for policies that do not support any enforcement
* denies further update to policy mode since the policy does not support enforcement

Example output:
```
$ ./tetra tracingpolicy list
ID   NAME              STATE     FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE           NPOST   NENFORCE   NMONITOR
1    read-etc-passwd   enabled   0          (global)    generic_kprobe   1.87 MB        monitor_only   0       0          0

$ ./tetra tracingpolicy set-mode read-etc-passwd enforce
Error: failed set mode to "enforce" tracing policy: rpc error: code = Unknown desc = unexpected error: setMode called in a collection that is monitor only
```

### Changelog

```release-note
new: monitor only mode
```

### DIFF
--- a/pkg/policyconf/test/mode_test.go
+++ b/pkg/policyconf/test/mode_test.go
@@ -11,6 +11,10 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/bpf"
 	"github.com/cilium/tetragon/pkg/grpc/tracing"
 	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
@@ -24,9 +28,6 @@ import (
 	pft "github.com/cilium/tetragon/pkg/testutils/policyfilter/tester"
 	tus "github.com/cilium/tetragon/pkg/testutils/sensors"
 	"github.com/cilium/tetragon/pkg/tracingpolicy"
-
-	"github.com/stretchr/testify/require"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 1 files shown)


pkg/selectors/kernel.go
```diff
@@ -1627,6 +1627,18 @@ func HasSigkillAction(kspec *v1alpha1.KProbeSpec) bool {
 	return false
 }
 
+func HasEnforcementAction(selectors []v1alpha1.KProbeSelector) bool {
+	for _, s := range selectors {
+		for _, action := range s.MatchActions {
+			act := actionTypeTable[strings.ToLower(action.Action)]
+			if act == ActionTypeSigKill || act == ActionTypeSignal || act == ActionTypeNotifyEnforcer {
```
kkourt: I think the `Override` action, as well as the `Set` action for uprobes, should also be here.

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -49,6 +50,44 @@ func newPolicyInfo(
 
 }
 
+func isEnforcementModeAllowed(spec *v1alpha1.TracingPolicySpec) bool {
+	if len(spec.Enforcers) > 0 {
+		return true
+	}
```
kkourt: Maybe we don't need this? A policy that has enforcers without a notify enforcer action shouldn't be able to enforce.

pkg/sensors/sensors.go
```diff
@@ -56,6 +56,8 @@ type Sensor struct {
 	Loaded bool
 	// Destroyed indicates whether the sensor had been destroyed.
 	Destroyed bool
+	// Enforcement indicates whether the sensor supports enforcement mode.
+	Enforcement bool
```
kkourt: nit: Maybe we can name this `SupportsEnforcement`?

kkourt (commented):
Thanks!

Left some comments, but LGTM.

pkg/sensors/tracing/policyhandler_linux.go
FedeDP: Makes sense!

pkg/sensors/sensors.go
FedeDP: Do you want me to also update new SensorIface method `HasEnforcement()` to `SupportsEnforcement()`?

kkourt: I don't think you are allowed to have a method with the same name as a struct member?
(Or at least that was my assumption, otherwise we can use `HasEnforcement` everywhere)

FedeDP: Oh you are right :laughing:

FedeDP:
> One more comment: can we have monitor_only in the list?

Well, right now the `mode` we see there is the one seen by bpf. 
I'd avoid having something like:
```
func (c *collection) mode() tetragon.TracingPolicyMode {
	if c.tracingpolicy == nil || c.state != EnabledState {
		return tetragon.TracingPolicyMode_TP_MODE_UNKNOWN
	}
	if !c.allowModeUpdate {
		return tetragon.TracingPolicyMode_TP_MODE_MONITOR_ONLY
	}

	mode, err := policyconf.PolicyMode(c.tracingpolicy)
```

because:
* we'd need to add `TracingPolicyMode_TP_MODE_MONITOR_ONLY` to our grpc api
* we are no more exposing the real bpf value
* i'd prefer to keep the `allowModeUpdate` information bit private

Do you think that, from an UX perspective, is it that bad not to expose `TracingPolicyMode_TP_MODE_MONITOR_ONLY` ?

kkourt:
> > One more comment: can we have monitor_only in the list?
> 
> Well, right now the `mode` we see there is the one seen by bpf. I'd avoid having something like:
> 
> ```
> func (c *collection) mode() tetragon.TracingPolicyMode {
> 	if c.tracingpolicy == nil || c.state != EnabledState {
> 		return tetragon.TracingPolicyMode_TP_MODE_UNKNOWN
> 	}
> 	if !c.allowModeUpdate {
> 		return tetragon.TracingPolicyMode_TP_MODE_MONITOR_ONLY
> 	}
> 
> 	mode, err := policyconf.PolicyMode(c.tracingpolicy)
> ```
> 
> because:
> 
>     * we'd need to add `TracingPolicyMode_TP_MODE_MONITOR_ONLY` to our grpc api
> 
>     * we are no more exposing the real bpf value
> 
>     * i'd prefer to keep the `allowModeUpdate` information bit private
> 
> 
> Do you think that, from an UX perspective, is it that bad not to expose `TracingPolicyMode_TP_MODE_MONITOR_ONLY` ?

I would say exposing MONITOR_ONY is good to have (but not too bad if we don't do it).
If we do it, however, I would also read the bpf value and check that things make sense.
That is, if bpf is in enforcement mode but userspace is in monitor_only, then something is probably wrong.

FedeDP:
> If we do it, however, I would also read the bpf value and check that things make sense.
That is, if bpf is in enforcement mode but userspace is in monitor_only, then something is probably wrong.

Well i like this use of the new flag. I think it makes sense then, to have this check! Let me update the PR :)

FedeDP:
Pushed the changes @kkourt !

Also, updated PR description with an updated output from `tetra tracingpolicy list`.

FedeDP:
Windows static checks CI is having some problems tracked here: https://github.com/cilium/tetragon/issues/4330
We can skip it i guess.

pkg/policyconf/policyconf.go
```diff
@@ -15,11 +15,16 @@ import (
...
+}
+
 const (
 	InvalidMode Mode = Mode(^uint8(0))
 	// NB: values below should match the ones in bpf/lib/policy_conf.h
-	EnforceMode Mode = 0
-	MonitorMode Mode = 1
+	EnforceMode     Mode = 0
+	MonitorMode     Mode = 1
+	MonitorOnlyMode Mode = 2 // monitor and cannot be updated to enforce. Not visible to bpf.
```
kkourt: Can you please add some information in the commit message for ` new(pkg/policyconf): introduce MonitorOnlyMode. `  on what is MonitorOnly mode and what is the motivation? I know it seems obvious now, but providing context is helpful both for people without the context we have now and also for our future selves.

Similarly for the other commits. Note that "why" we are doing something is equally (if not more) important that what. Usually, the what can be figured out by looking at the code, but the why is not so easy.

pkg/sensors/collection.go
```diff
@@ -89,7 +89,8 @@ type collection struct {
 	// if this is not zero, then the policy is filtered
 	policyfilterID uint64
 	// state indicates the state of the collection
-	state TracingPolicyState
+	state           TracingPolicyState
+	allowModeUpdate bool
```
kkourt: Suggestion: How about we keep the mode here? That is, monitor, monitor_only, enforce.

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -49,6 +50,40 @@ func newPolicyInfo(
 
 }
 
+func isEnforcementModeAllowed(spec *v1alpha1.TracingPolicySpec) bool {
```
kkourt: Nit: Can we call this `hasEnforcementActions`?

I think it's more apt since the question is whether a certain policy has enforcement actions.

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -89,7 +130,12 @@ func (pi *policyInfo) policyConfMap(prog *program.Program) *program.Map {
 			mode := policyconf.EnforceMode
 			if pi.specOpts != nil {
 				mode = pi.specOpts.policyMode
+				if mode == policyconf.MonitorOnlyMode {
+					// We don't expose MonitorOnly to ebpf
+					mode = policyconf.MonitorMode
+				}
 			}
+
```
kkourt: nit: whitespace

pkg/sensors/collection.go
```diff
@@ -128,6 +128,13 @@ func (c *collection) mode() tetragon.TracingPolicyMode {
 		return tetragon.TracingPolicyMode_TP_MODE_UNKNOWN
 	}
 
+	if !c.allowModeUpdate {
+		if mode != policyconf.MonitorMode {
+			logger.GetLogger().Warn("monitor-only bit is set but bpf map is not in monitor mode", "policy", c.name)
+		}
+		return tetragon.TracingPolicyMode_TP_MODE_MONITOR_ONLY
```
kkourt: comment: I wonder if we could return something else here if the mode is not monitor mode, but I don't have any good ideas. We can leave it as is and revisit as needed.

kkourt (commented):
Thanks!

pkg/sensors/collection.go
FedeDP: I agree, don't really know what else to return :)

pkg/sensors/collection.go
FedeDP: I think that using the ebpf map as source of truth is a good thing; i'd avoid storing that info bit in userspace.

pkg/policyconf/policyconf.go
FedeDP: I force-pushed the reworded commits; feel free to tell me if there's anything else that you feel like should be improved :)

pkg/sensors/collection.go
FedeDP: If we want the mode to be everywhere, we should really push it down to ebpf; lots of ugliness in this PR would go away then.

FedeDP: I force-pushed a new PR version with the above implemented. It is IMHO much easier.

mtardy:
Just took a look and this is nice! thank you!

PR #4317, opened 2025-11-07
Makefile.cli: make CLI compilation parallelizable

mtardy:
Our CLI release/compilation was always a bit slow because it could be started in parallel.

Unfortunately looks like it doesn't make a difference on the slow GitHub runner with only 4 cores... but it does on large machines. I find it better but not a lot of args here.

### DIFF
--- a/Makefile.cli
+++ b/Makefile.cli
@@ -4,7 +4,22 @@
 CLI_GO_BUILD = CGO_ENABLED=0 $(GO) build
 # renovate: datasource=docker
 GO_IMAGE = docker.io/library/golang:1.25.4-alpine@sha256:d3f0cf7723f3429e3f9ed846243970b20a2de7bae6a5b66fc5914e228d831bbb
-TARGET=tetra
+TARGET = tetra
+
+RELEASE_FOLDER = release
+# Avoid shooting yourself in the foot with rm in cli-clean or
+# creating/overwriting files from the root folder
+ifeq ($(strip $(RELEASE_FOLDER)),)
+$(error RELEASE_FOLDER is not set or is empty)
+endif
+
+RELEASES := \
+	$(RELEASE_FOLDER)/$(TARGET)-darwin-arm64.tar.gz \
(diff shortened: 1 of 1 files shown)


Makefile.cli
```diff
@@ -4,51 +4,48 @@
 CLI_GO_BUILD = CGO_ENABLED=0 $(GO) build
 # renovate: datasource=docker
 GO_IMAGE = docker.io/library/golang:1.25.4-alpine@sha256:d3f0cf7723f3429e3f9ed846243970b20a2de7bae6a5b66fc5914e228d831bbb
-TARGET=tetra
+TARGET = tetra
+
+RELEASE_FOLDER = release
```
FedeDP, line 9: Since the target name below (`release/$(TARGET)-%.tar.gz:`) mandates this, i'd prefer to be explicit instead of adding a variable.
Ie: i'd drop `RELEASE_FOLDER = release` and just do
```
RELEASES := \
	release/$(TARGET)-darwin-arm64.tar.gz \
...
```

FedeDP: Otherwise the variables gives the false impression that by updating its value, you are ok while you are not :)

mtardy: ah yeah good catch, I added the $RELEASE_FOLDER variable everywhere so that you can change it on the fly!

mtardy: maybe I should add `./` before $(RELEASE_FOLDER) in other commands or check globally that it's set to something non emtpy to avoid root mkdir/write etc.

FedeDP (approved):

Makefile.cli
FedeDP: Mmh it would be helpful, yes :)

mtardy: https://github.com/cilium/tetragon/compare/9e89e8fda59155eeb64f2ff18a67a680907a3345..3e9769156e4e81ef972d30338cd1f0f2f9e4b440

PR #4320, opened 2025-11-09
tetragon: enable loader test

olsajiri:
and one doc fix

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -826,9 +826,10 @@ register name followed by `=` and an assignment expression, which can be one of 
 following types:
 
 - constant `rax=11`
-- register `rbp=(%rsp)`
-- register plus offset `rip=8(%rsp)`
-- dereference of register plus offset `rsp=8%rsp`
+- register `rbp=%rsp`
+- register plus offset `rip=8%rsp`
+- dereference of register `rbp=(%rsp)`
+- dereference of register plus offset `rsp=8(%rsp)`
 
 {{< note >}}
 This interface is likely to be changed in the future.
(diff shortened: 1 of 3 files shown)


kkourt (approved):

FedeDP (approved):

pkg/sensors/tracing/loader_test.go
```diff
@@ -95,6 +96,16 @@ func TestLoader(t *testing.T) {
 		t.Skip("no support for loader events")
 	}
 
+	// Vmtests runs on 2 cpus system with ringbuffer size if 128K (which is
+	// 64K per cpu) and that might not be enough when all the high volume
+	// loader traffic gets stuck on single cpu so we might lose the test
+	// loader event.. 10M per cpu seems to be plenty for the test to pass.
```
mtardy: little typo that made it hard to read that long sentence
```suggestion
	// Vmtests runs on 2 cpus system with ringbuffer size of 128K (which is
	// 64K per cpu) which might not be enough when all the high volume
	// loader traffic gets stuck on single cpu so we might lose the test
	// loader event.. 10M per cpu seems to be plenty for the test to pass.
```
btw 64K from 10M is x150, it seems really large!

mtardy (approved):
Thank you!!

pkg/sensors/tracing/loader_test.go
olsajiri: yes.. still I think it's ok wrt we are sure we have plenty and the total of 4GB that vm runs with

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -827,8 +827,8 @@ following types:
 
 - constant `rax=11`
 - register `rbp=(%rsp)`
```
andrewstrohman: It might be worth mentioning that this is a dereference too.

andrewstrohman (approved):

Issue #3728, opened 2025-05-12
Tetragon does not raise an event when the returned value is null


tdaudi:
### What happened?

Let's take an example to illustrate this with the below policy. To trigger this policy, run any command in bash.

In `security_bprm_check` the `bprm.interpreter` attribute is null, but Tetragon does not raise the error
https://elixir.bootlin.com/linux/v6.14-rc3/source/include/linux/binfmts.h#L54

```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "lsm"
spec:
  lsmhooks:
  - hook: "bprm_check_security"
    args:
    - index: 0
      type: "string"
      resolve: "executable.f_path.dentry.d_name.name"
    selectors:
    - matchActions:
      - action: Post
```

No events are raised when this policy is applied. 
If you try 
```yaml
    - index: 0
      type: "uint32"
      resolve: "executable.f_mode"
```

A `0` is raised where it should not, because here this is not the `f_mode` that is null but the `executable`.

This happened in the below code.
https://github.com/cilium/tetragon/blob/main/bpf/process/generic_calls.h#L70-L78

We do not track errors on the extract or later on the copy value.

### Tetragon Version

latest

### Kernel Version

6.14

### Relevant log output

```shell

```

andrewstrohman:
Hi @tdaudi. Thanks for creating this issue and giving me you insights in #4327.  It's great getting your feedback early.

> In security_bprm_check the bprm.interpreter attribute is null,

I think you meant `bprm.executable` here, not `bprm.interpreter`

>  but Tetragon does not raise the error

Does this mean generate an event? I see you said that in your next sentence. I just want to make sure that I correctly understand the expectation.

> A 0 is raised where it should not, because here this is not the f_mode that is null but the executable.

[...]

PR #4327, opened 2025-11-11
handle resolve of NULL pointers

andrewstrohman:
The resolve logic follows pointers by dereferencing. Some of these pointer values may be NULL, or otherwise impossible to dereference. In such cases, matchArgs selectors should fail to match, because the argument was not really resolved.

When we fail to dereference during resolve, report the depth at which the failure occurred. Also, when argument reading fails, the event's kprobe argument type will be converted to KprobeError, instead of what was specified in the tracing policy.

The intention is to have a follow-up PR that converts the depth of failure to a string that tells the user which pointer could not be dereferenced.

### DIFF
--- a/pkg/sensors/tracing/uprobe_reg_test.go
+++ b/pkg/sensors/tracing/uprobe_reg_test.go
@@ -1233,3 +1233,169 @@ func TestUprobePtRegsPreloadSubstringOverrideSingle(t *testing.T) {
 func TestUprobePtRegsPreloadSubstringOverrideMulti(t *testing.T) {
 	testUprobePtRegsPreloadSubstringOverride(t, false)
 }
+
+func TestUprobeResolveNull(t *testing.T) {
+	if !config.EnableLargeProgs() || !bpf.HasUprobeRefCtrOffset() {
+		t.Skip("Need 5.3 or newer kernel for uprobe ref_ctr_off support for this test.")
+	}
+
+	uprobe := testutils.RepoRootPath("contrib/tester-progs/uprobe-null")
+	uprobeBtf := testutils.RepoRootPath("contrib/tester-progs/uprobe-null.btf")
+
+	tt := []struct {
+		arg    string
+		kpArgs []*ec.KprobeArgumentChecker
+	}{
+		{"first", []*ec.KprobeArgumentChecker{
+			ec.NewKprobeArgumentChecker().WithErrorArg(ec.NewKprobeErrorChecker().WithMessage(sm.Full("1"))),
+		}},
+		{"second", []*ec.KprobeArgumentChecker{
+			ec.NewKprobeArgumentChecker().WithErrorArg(ec.NewKprobeErrorChecker().WithMessage(sm.Full("2"))),
+		}},
+		{"third", []*ec.KprobeArgumentChecker{
+			ec.NewKprobeArgumentChecker().WithErrorArg(ec.NewKprobeErrorChecker().WithMessage(sm.Full("3"))),
+		}},
+		{"nonull", []*ec.KprobeArgumentChecker{
+			ec.NewKprobeArgumentChecker().WithIntArg(0),
+		}},
+	}
+
+	uprobeHook := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "uprobe-null"
+spec:
+  uprobes:
+  - path: "` + uprobe + `"
+    btfPath: "` + uprobeBtf + `"
+    symbols:
+    - "func"
+    args:
+    - index: 0
+      type: "int32"
+      btfType: "first"
+      resolve: "second.third.val"
+`
+
+	uprobeConfigHook := []byte(uprobeHook)
+	err := os.WriteFile(testConfigFile, uprobeConfigHook, 0644)
+	if err != nil {
+		t.Fatalf("writeFile(%s): err %s", testConfigFile, err)
+	}
+
+	var checkers []ec.EventChecker
+	for i := range tt {
+		checkers = append(checkers, ec.NewProcessUprobeChecker("uprobe-null").
+			WithProcess(ec.NewProcessChecker().
+				WithBinary(sm.Full(uprobe)).
+				WithArguments(
+					sm.Full(tt[i].arg)),
+			).WithArgs(ec.NewKprobeArgumentListMatcher().
+			WithOperator(lc.Ordered).
+			WithValues(tt[i].kpArgs...)))
+	}
+
+	var doneWG, readyWG sync.WaitGroup
+	defer doneWG.Wait()
+
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
+	obs, err := observertesthelper.GetDefaultObserverWithFile(t, ctx, testConfigFile, tus.Conf().TetragonLib, observertesthelper.WithMyPid())
+	if err != nil {
+		t.Fatalf("GetDefaultObserverWithFile error: %s", err)
+	}
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	for i := range tt {
+		cmd := exec.Command(uprobe, tt[i].arg)
+		cmdErr := testutils.RunCmdAndLogOutput(t, cmd)
+		require.NoError(t, cmdErr)
+	}
+
+	err = jsonchecker.JsonTestCheck(t, ec.NewUnorderedEventChecker(checkers...))
+	require.NoError(t, err)
+}
+
+func UprobeResolveNullMatch(t *testing.T, expectCheckerFailure bool, arg string) {
+	if !config.EnableLargeProgs() || !bpf.HasUprobeRefCtrOffset() {
+		t.Skip("Need 5.3 or newer kernel for uprobe ref_ctr_off support for this test.")
+	}
+
+	uprobe := testutils.RepoRootPath("contrib/tester-progs/uprobe-null")
+	uprobeBtf := testutils.RepoRootPath("contrib/tester-progs/uprobe-null.btf")
+
+	uprobeHook := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "uprobe-null"
+spec:
+  uprobes:
+  - path: "` + uprobe + `"
+    btfPath: "` + uprobeBtf + `"
+    symbols:
+    - "func"
+    args:
+    - index: 0
+      type: "int32"
+      btfType: "first"
+      resolve: "second.third.val"
+    selectors:
+    - matchArgs:
+      - args: [0]
+        operator: "Equal"
+        values:
+          - "0"
+`
+
+	uprobeConfigHook := []byte(uprobeHook)
+	err := os.WriteFile(testConfigFile, uprobeConfigHook, 0644)
+	if err != nil {
+		t.Fatalf("writeFile(%s): err %s", testConfigFile, err)
+	}
+
+	kpArgs := []*ec.KprobeArgumentChecker{
+		ec.NewKprobeArgumentChecker(),
+	}
+
+	var checkers []ec.EventChecker
+	checkers = append(checkers, ec.NewProcessUprobeChecker("uprobe-null").
+		WithProcess(ec.NewProcessChecker().
+			WithBinary(sm.Full(uprobe)).
+			WithArguments(
+				sm.Full(arg)),
+		).WithArgs(ec.NewKprobeArgumentListMatcher().
+		WithOperator(lc.Ordered).
+		WithValues(kpArgs...)))
+
+	var doneWG, readyWG sync.WaitGroup
+	defer doneWG.Wait()
+
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
+	obs, err := observertesthelper.GetDefaultObserverWithFile(t, ctx, testConfigFile, tus.Conf().TetragonLib, observertesthelper.WithMyPid())
+	if err != nil {
+		t.Fatalf("GetDefaultObserverWithFile error: %s", err)
+	}
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	cmd := exec.Command(uprobe, arg)
+	cmdErr := testutils.RunCmdAndLogOutput(t, cmd)
+	require.NoError(t, cmdErr)
+
+	err = jsonchecker.JsonTestCheckExpect(t, ec.NewUnorderedEventChecker(checkers...), expectCheckerFailure)
+	require.NoError(t, err)
+}
+
+func TestUprobeResolveNullMatchPositive(t *testing.T) {
+	UprobeResolveNullMatch(t, false, "nonull")
+}
+
+func TestUprobeResolveNullMatchNegative(t *testing.T) {
+	UprobeResolveNullMatch(t, true, "first")
+}


bpf/process/generic_calls.h
```diff
@@ -391,14 +391,20 @@ extract_arg_depth(u32 i, struct extract_arg_data *data)
 {
 	if (i >= MAX_BTF_ARG_DEPTH || !data->btf_config[i].is_initialized)
 		return 1;
+	/* NULL pointer */
+	if (*data->arg == 0) {
+		*data->null_ptr_found = true;
```
tdaudi: If the last `btf_config` resolve in a null pointer, you would not reach that code right ?

bpf/process/generic_calls.h
```diff
@@ -413,6 +419,7 @@ FUNC_INLINE void extract_arg(struct event_config *config, int index, unsigned lo
 		struct extract_arg_data extract_data = {
 			.btf_config = btf_config,
 			.arg = a,
+			.null_ptr_found = null_ptr_found,
```
tdaudi: I think the field should probably be called `err` because you catch the error from `probe_read`

pkg/sensors/tracing/uprobe_amd64_test.go
```diff
@@ -222,6 +248,17 @@ spec:
       type: "` + tt[2].specTy + `"
       btfType: "mystruct"
       resolve: "` + tt[2].field + `"
+    - index: 1
+      type: "uint8"
+      btfType: "mystruct"
+      resolve: "sub.v8"
```
tdaudi: I think 2 test cases should be added. This one is good, but if `sub` alone is resolved, does it raise an error too ? For instance, if `type: file` is null, does Tetragon display null as well ?

andrewstrohman:
> Thank you for taking time to write this PR. I opened an issue for this few month ago, see #3728.

Thanks for bringing this to my attention, and your feedback/ideas here. 

> * First, what happen if the null structure is not the last resolving field. How can you track where it ends ?

I've changed things so that I keep track of which level of depth we are unable to dereference. I use a sentinel value of -1 to indicated that no issues were detected. This is then used as an indicator that selectors against this arg shouldn't match. I'm thinking that perhaps this level of depth of failure information could be put in the event not only to distinguish resolve failures from success, but also what depth the failure occurred.

> * Also, what happened if you are resolving `int *value` and the ptr is null ?

I'm investigated this, and it seems that this scenario doesn't work for the non-NULL case. So, I want to fix that first, before considering how to handle the NULL case. This is what I'm trying:

```
struct mystruct {
	uint8_t  v8;
	uint16_t v16;
	uint32_t v32;
	uint64_t v64;
	struct mysubstruct sub;
	struct mysubstruct *subp;
	uint32_t *v32p;
};
```
In the userspace program I'm testing against, I do this:
```
struct mystruct s = {0};
uint32_t value = 3;
s.v32p = &value;
```
The policy's arg config:
```
    - index: 1
      type: "uint32"
      btfType: "mystruct"
      resolve: "v32p"
```
Is this the scenario you're describing here?

It seems like we might need to do one more iteration of calling extract_arg_depth and dereference again to make this work as expected. Put another way, I think that we are acting on a pointer as if it's an integer.

> * And if you are resolving a struct (e.g. `type: file`) you need to let `extract_arg` ends correctly, because at this point you have no idea if `extract_arg` return a legitimate `0` or a null struct. But when resolving it's fields, you will figure out that it is actually NULL.

I'll look into this more closely, but from reading the code it seems that we don't handle the `type: file` NULL case correctly, even outside of resolving. At first glance, I'm not seeing checks for `probe_read()` return values even though this type requires dereference. Overall, I thinking that my strategy is to fix issues in `read_arg()` where we should be watching the return value of `probe_read()` (types that require dereference), if they exist, independently from catching dereference problems in `extract_arg_depth`. I think the combination of these two methods will cover all bases. What do you think about this approach? 

> The potential solution I was thinking of for this was to take all the return values of the `probe_read` and find a way to send this error to the userspace if it is `< 0`. 

Thanks for this idea. I think this is the right approach, and I have started moving in this direction.

> So in case of a null pointer, you can have a new error code that the kernel does not cover, for instance `-100`, that is dedicated to it.

I'm not sure if I follow this. I think we need to communicate the error out of band of the arg value or else we will have a collision. Perhaps we can report the depth at which dereferencing failed in the event, in order to distinguish between resolve success and failure.


I'm trying to figure out what is the right behavior is for when we cannot resolve due to a NULL pointer. My current approach is to indicate there was a resolve error, but not let that prevent the event from firing, unless there is a matchArgs selector associated with the arg. In this method, I would communicate that the arg was not actually resolved in the event somehow. But, I realized from the issue that you created about this, that NULL string pointers prevent an event from firing, regardless of selectors. So maybe that's a more appropriate outcome.

kkourt:
> I'm trying to figure out what is the right behavior is for when we cannot resolve due to a NULL pointer. My current approach is to indicate there was a resolve error, but not let that prevent the event from firing, unless there is a matchArgs selector associated with the arg. In this method, I would communicate that the arg was not actually resolved in the event somehow.

Above makes sense to me.

>  But, I realized from the issue that you created about this, that NULL string pointers prevent an event from firing, regardless of selectors. So maybe that's a more appropriate outcome.

I don't think that the appropriate outcome.

I'll use the example from the issue for clarity:

```yaml
  lsmhooks:
  - hook: "bprm_check_security"
    args:
    - index: 0
      type: "string"
      resolve: "executable.f_path.dentry.d_name.name"
    selectors:
    - matchActions:
      - action: Post
```

In this case (that is, where there no matchArgs), I think we should generate an event (perhaps with an empty argument, or even a special value to indicate that some resolution went wrong).

bpf/process/generic_calls.h
andrewstrohman: Yes, this is good point. I think my new placement for error detection takes care of this problem.

pkg/sensors/tracing/uprobe_amd64_test.go
andrewstrohman: > I think 2 test cases should be added.

You go on to describe what sounds like one test case. What is the other test case you're thinking about here?

>  but if sub alone is resolved, does it raise an error too ? For instance, if type: file is null, does Tetragon display null as well ?

As I mentioned in https://github.com/cilium/tetragon/issues/3728, I think that we should break this up into two independent tasks. One is to handle the dereferencing in extract_arg_depth, and the other is to handle issues for the terminal types reach by resolve (in `read_arg`), which would happen even without the use of the resolve feature. With pointer types, read_arg() has some issues when the pointer is NULL. Because of this, I don't want to test scenarios where we resolve to a NULL pointer and pass that NULL pointer to read_arg() (yet).

Also, for the example you have above:
> what happened if you are resolving int *value and the ptr is null ?

From my testing, I see that `read_arg()` is passed a pointer to an int in this scenario. If we want to resolve the int value that the pointer points to, I think we need to call `extract_arg_depth` one more time in order to dereference, because `read_arg()` takes the integer "by value" instead of by pointer to an integer. So, I don't want to test this scenario until we make this work in the non-NULL case.

So, I think that I should add a test for when the first pointer is NULL, and when the second to last pointer is NULL, but I don't want to test the scenario where the last pointer is NULL, because that problem is not specific to resolve, so I want to solve that issue independently.

andrewstrohman:
I'm converting back to draft, as I just realized a problem with my approach. I need to think about tracepoints and usdt a bit more.

olsajiri:
hi, just some thoughts for discussion.. I dont feel strongly about it, I'm not sure what the right solution is ;-)

do we need that error-depth number in the final event? I wonder if you see it in event in most cases you'll know what failed.. in which case using some flag instead would be enough?

and if there's reason to keep it (I guess long derefs would justify that) could we maybe store the part of the deref chain (string) that failed instead of depth (we have that deref chain info already)

and how about using metric instead, something like `deref_fails_cnt[policy,sensor,arg-idx,deref-chain]` .. as I'm not sure this info should be part of the final event.. but I could be easily wrong

andrewstrohman:
> hi, just some thoughts for discussion.. I dont feel strongly about it, I'm not sure what the right solution is ;-)

Thanks for the review and helping me think this through.


> do we need that error-depth number in the final event? I wonder if you see it in event in most cases you'll know what failed.. in which case using some flag instead would be enough?

I'd be OK with just a boolean flag that indicates that resolving failed. My primary goal here is to indicate that the argument value is bogus. I can't just remove the bogus arg value from the event because that would shift some args values to the left of what was configured in the spec, causing a mis-match with expectations.

I think I went down this path of recording which dereference failed, because of @tdaudi's earlier comment. He said:
> First, what happen if the null structure is not the last resolving field. How can you track where it ends ?

@tdaudi, did I understand your comment correctly? Is this something that is important to you? If this is not what you meant, or you don't feel strongly about this, I think we should just go with a boolean flag. I initially thought this might be nice to expose, but it's not important to me.

If users really want to know where the NULL deference happened, they can resolve the intermediate pointers to see which one was NULL, if the they have the args to spare (we have a max of 5). This way, they can also run a selector against it too, whereas just reporting the depth does not allow them to filter based on it.

> and if there's reason to keep it (I guess long derefs would justify that) could we maybe store the part of the deref chain (string) that failed instead of depth (we have that deref chain info already)

When you say "long derefs" here, I think you mean multiple dereferences per resolve. Please let me know if I misunderstood. I agree that if we decide to pinpoint the exact deference that failed it would be better to go all the way and show them where it failed via a prefix of the resolve configuration string. Just reporting the depth requires the user to know too much (and do extra investigation) in order to be able to interpret it in a useful way. But I'm currently leaning toward just changing it to be a boolean.

> and how about using metric instead, something like `deref_fails_cnt[policy,sensor,arg-idx,deref-chain]` .. as I'm not sure this info should be part of the final event.. but I could be easily wrong

When you say "this info", I think you mean the depth of dereference failure here. I think we agree that we need some signal in the event that indicates that the arg value is bogus. Please correct me if I misunderstand your position here.

I would prefer your first or second second suggestion over this, because the user cannot definitely know which event the counter increment was related to.


When we fix `read_arg()`,so that NULL pointers won't prevent events(see [here](https://github.com/cilium/tetragon/issues/3728#issuecomment-3529351256) for background), we can use this same boolean flag to indicate the dereference failure, unifying the two sources of dereference failures.

tdaudi:
> @tdaudi, did I understand your comment correctly? Is this something that is important to you? If this is not what you meant, or you don't feel strongly about this, I think we should just go with a boolean flag. I initially thought this might be nice to expose, but it's not important to me.
> 
> If users really want to know where the NULL deference happened, they can resolve the intermediate pointers to see which one was NULL, if the they have the args to spare (we have a max of 5). This way, they can also run a selector against it too, whereas just reporting the depth does not allow them to filter based on it.

My initial idea was to have a convinient way to know which field in the resolve path is null to avoid spending time on updating the policy and testing again. But for production, I think it is interesting to know where the resolve fails when the null value is unexpected. Knowing where the null appears help understanding the reason and avoid the difficulty of having to reproduce it, especially if it occurs rarely.

bpf/process/generic_calls.h
```diff
@@ -588,7 +605,7 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 	else
 		a = (&e->a0)[arg_index];
 
-	extract_arg(config, index, &a);
+	extract_arg(config, index, &a, &e->resolve_err_depth[index]);
```
olsajiri: so if we fail to resolve  we still continue and store the 'unresolved data' which likely null, right? I think we can skip it

but how about we make this more generic and we start each argument data with '__u32 error' status and in case of failure we write just error != 0 as argument value.. in case of resolve failure we encode the depth into it

then on user space side the getArg would just read first uint32 from event reader to get the argument status
so all the extra retrieval of error depth from each probe type would not be needed, also probably the argument index will be clear, because you have the error for argument you are about to read, or skip reading if it's != 0

we could use this to store other errors that happen during argument encoding, which there are plenty

and into final event instead of the depth value, I'd add error string that in case of resolve failure would contain something like: "failed to resolve current->file->f_inode"

seems like this could save some cycles and prevent bogus arguments values, because IIUC if we bail from argument storing in the middle on kernel side the user side still tries to read the whole part

andrewstrohman: > so if we fail to resolve we still continue and store the 'unresolved data' which likely null, right? I think we can skip it

I think "skip it" means skipping calling `read_arg` for the arg. This way, `e->args` isn't filled with the bogus value. 

> but how about we make this more generic and we start each argument data with '__u32 error' status and in case of failure we write just error != 0 as argument value.. in case of resolve failure we encode the depth into it

OK, so each argument value in `e->args` is proceeded by a u32 that holds the depth of failure, or 0 if no failure occurred. If the u32 is non-zero, no bogus arg value follows -- the next bytes in `e->args` are about the subsequent arg.

> we could use this to store other errors that happen during argument encoding, which there are plenty

I think these other errors mean passing NULL pointers to `read_arg()` (which are problems that pre-dated resolve functionality). Is that what you're thinking here or something else?

> and into final event instead of the depth value, I'd add error string that in case of resolve failure would contain something like: "failed to resolve current->file->f_inode"

Yes, I agree adding a string that shows where we failed to dererefernce, is better than just exposing the depth of the error. So, from an implementation perspective, this means instead of adding member `int32 resolve_err_depth` to `KprobeArgument`, I would add a string member/attribute to `KprobeArgument` that describes the error that happened when reading the arg value. If the string is empty, it means there was no error reading the arg. See below for another possible way to interpret this. Regardless, in userspace, I would use the resolve failure depth send from the bpf program to determine where the NULL pointer was and only expose that to the user.

> seems like this could save some cycles and prevent bogus arguments values, because IIUC if we bail from argument storing in the middle on kernel side the user side still tries to read the whole part

In your suggested approach, I like that we don't send a bogus value from bpf to userspace. I like that the argument read outcome sits directly in front of the argument value in `e->args`, so that they are coupled and are less like to get mixed up.

What should `getArg()` return when it sees a non-zero u32 proceeding the value it wants to read? Should I create something like:
```
type MsgGenericKprobeArgFailed struct {
	Index         uint64
	Value      string
	Label         string
}
```
In this approach, there wouldn't be an error string associated with every KprobeArgument. The error string would only be associated with KprobeArgument_ArgFailed. I'm not sure if there would be a problem if the user expected a receive an KprobeArgument_UintArg, for example, in a specific position within the args list, but instead found type KprobeArgument_ArgFailed? If you don't see an issue with this way of communicating the failure, I think it's the cleanest method, because we don't need to fabricate a bogus arg value in userspace as a placeholder.

While I like not transmitting a bogus argument to userspace within `e->args`, and I like not mixing up the arg read outcomes, I am worried that this approach will require changes in bpf programs which will result in their instruction count exceeded 4k (on kernels that don't support __LARGE_BPF_PROG).

While experimenting with what I currently have, I found that in many code paths, I didn't have any instructions to spare for 4.19. I couldn't even check for an int32 being 0 without crossing the 4k boundary. In the current approach, I was able to avoid adding any new instructions in the mostly full programs by completely isolating the change from 4.19 with `__LARGE_BPF_PROG`. But with this proposal, I think I'll need to make 4.19 aware of the uint32 value that proceeds each arg, so I worry that I won't be able to make it fit. I'll keep a close eye on this and let you know if I hit this 4k instruction count problem and cannot get around it.

olsajiri: > I think these other errors mean passing NULL pointers to read_arg() (which are problems that pre-dated resolve functionality). Is that what you're thinking here or something else?

I was thinking about it more like general approach to for read_arg plus for resolve status, but your point about 4.19 seems valid.. I guess we could start with just resolve error if the rest wont fit in.. perhaps there's a way to split read_arg into 2 tailcalls .. on the other hand it's just 4.19 issue, and I wonder how long this will be around

or we make it just for `__LARGE_BPF_PROG` which might mean bit uglier code for user space, but I guess we have to try first to see ;-)

> I'm not sure if there would be a problem if the user expected a receive an KprobeArgument_UintArg, for example, in a specific position within the args list, but instead found type KprobeArgument_ArgFailed?

the KprobeArgument_ArgFailed makes sense to me, I think it's ok to find unexpected KprobeArgument_UintArg, because we were not able to read expected type anyway.. @kkourt any idea about that?

kkourt: Is the idea to have an `KprobeArgument_ArgFailed` type for arguments that we failed to read (e.g., in the case of a resolve that hit a NULL value)? If so, makes sense to me.

andrewstrohman: Yeap, that's the idea. Thanks for the feedback.

andrewstrohman: @olsajiri I changed it, so that there is an arg status header before each arg that indicates whether the arg was able to be read. This way, I was able to avoid sending a bogus arg value from the bpf program to userspace.

Since the instruction count budget is really tight, I decided to exclude 4.19 from this change.

Please let me know what you think.

bpf/process/types/basic.h
```diff
@@ -2298,9 +2304,11 @@ rate_limit(__u64 ratelimit_interval, __u64 ratelimit_scope, struct msg_generic_k
 	dst = key->data;
 
 	for (i = 0; i < MAX_POSSIBLE_ARGS; i++) {
+		if (config->idx[i] == -1)
+			break;
 		if (e->argsoff[i] >= e->common.size)
 			break;
-		if (i < MAX_POSSIBLE_ARGS - 1)
+		if (config->idx[i + 1] != -1)
```
kkourt: Hmmm... for the last iteration, config->idx[i+1] is out of bounds, no?
I mean, it's still valid access in the map (so validation does not fail) but I think it's outside of ->idx.
I think this what the initial `i < MAX_POSSIBLE_ARGS - 1` check was trying to avoid.

bpf/process/generic_calls.h
```diff
@@ -190,6 +190,17 @@ copy_iov_iter(void *ctx, long off, unsigned long arg, int argm, struct msg_gener
 #define copy_iov_iter(ctx, orig_off, arg, argm, e) 0
 #endif /* __LARGE_BPF_PROG */
 
+FUNC_INLINE long write_arg_status(struct msg_generic_kprobe *e, unsigned long offset, __u32 status)
+{
+#ifdef __LARGE_BPF_PROG
+	char *args = args_off(e, offset);
+	*(__u32 *)args = status;
+	return (offset + sizeof(__u32)) & 16383;
+#else
```
kkourt: We seem to have a different behavior for different kernels. Please document this in the commit message.

bpf/lib/generic.h
```diff
@@ -65,6 +65,7 @@ struct msg_generic_kprobe {
 	char args[24000];
 	unsigned long a0, a1, a2, a3, a4;
 	long argsoff[MAX_POSSIBLE_ARGS];
+	__u32 arg_error_status[MAX_POSSIBLE_ARGS];
```
kkourt: From what I can tell, we write the status in ->args[]  and I guess user-spaces reads it from there?
Can't we just add the `__u32 arg_error_status` before args and make it part of `MsgGenericKprobe` and friends?
I think it  would simplify the code significantly.

bpf/process/types/basic.h
andrewstrohman: Yes, you're right. Thank you and good catch. I've updated the condition to be:
```
if (i < MAX_POSSIBLE_ARGS - 1 && config->idx[i + 1] != -1)
```

bpf/process/generic_calls.h
andrewstrohman: I've added the following to the commit message for the bpf change commit:

```
Since the instruction count budget is extremely tight for 4.19, this
change only applies to later kernels which can accommodate the
instruction count increase.
```

bpf/lib/generic.h
andrewstrohman: > From what I can tell, we write the status in ->args[] and I guess user-spaces reads it from there?

Yes, that's right. 

> Can't we just add the __u32 arg_error_status before args and make it part of MsgGenericKprobe and friends?
I think it would simplify the code significantly.

This was my first approach, but @olsajiri preferred that I change it to be this way. See the discussion [here](https://github.com/cilium/tetragon/pull/4327#discussion_r2554150947). If the error status is not a header for the arg values within `->args[]`, it's more likely that we can mix up which arg the error status applies to.

See [here](https://github.com/andrewstrohman/tetragon/commits/andrewstrohman/resolve-null-backup/) for a backup of my first approach where I tried this.

kkourt: Ah, sorry, missed that initial discussion.

@olsajiri thoughts?

I think that prepending a status before every argument in ->args makes it more complicated than it needs to be (especially with all the ifdefs for it).

Maybe we have something like ->error_status as part of the visible message (MsgGenericKprobe), but have it ordered by how arguments are stored in ->args? How do we feel about this option?

olsajiri: hm, it does seem more complicated than I expected, lemme check on that

bpf/process/types/basic.h
```diff
@@ -2298,9 +2304,11 @@ rate_limit(__u64 ratelimit_interval, __u64 ratelimit_scope, struct msg_generic_k
 	dst = key->data;
 
 	for (i = 0; i < MAX_POSSIBLE_ARGS; i++) {
+		if (config->idx[i] == -1)
```
olsajiri: we have arg_idx function for, could we use it? also below

bpf/lib/generic.h
olsajiri: so we need plenty of large prog ifdefs because it's too big..  and it's a problem for TAIL_CALL_PROCESS call right? I might have an idea how to split it, how much in hurry are we with this change? ;-)

I think having the status before each arg is still better option, that ease up the arg processing

bpf/process/types/basic.h
olsajiri: we have arg_idx function for, could we use it? also below

bpf/lib/generic.h
olsajiri: > I think that prepending a status before every argument in ->args makes it more complicated than it needs to be (especially with all the ifdefs for it).

yea, there's lot of ifdefs.. I have an idea on splitting of TAIL_CALL_PROCESS which should help,

> Maybe we have something like ->error_status as part of the visible message (MsgGenericKprobe), but have it ordered by how arguments are stored in ->args? How do we feel about this option?

the original version was quite complex change on user side, I think having the status before each argument is a better way

olsajiri: > yea, there's lot of ifdefs.. I have an idea on splitting of TAIL_CALL_PROCESS which should help,

draft in here https://github.com/cilium/tetragon/pull/4439

bpf/process/types/basic.h
andrewstrohman: I avoided this initially, because it caused a circular import problem. But now that you brought this up, I broke some code out into a new header file called event_config.h in order to do it.

bpf/lib/generic.h
andrewstrohman: The draft looks good to me. Thanks!

olsajiri:
@andrewstrohman looks good, could you please move the args tail-call split into separate PR? I think it's good to be pushed and would make the rest easier to review/ack, thanks

andrewstrohman:
> @andrewstrohman looks good, could you please move the args tail-call split into separate PR? I think it's good to be pushed and would make the rest easier to review/ack, thanks

I broke this out into [this](https://github.com/cilium/tetragon/pull/4527) PR.

andrewstrohman:
@olsajiri and @kkourt The requested prerequisite PR to break up the arg filter tail call has been merged. So, this PR is ready for review again.

pkg/sensors/tracing/generic.go
```diff
@@ -89,37 +89,39 @@ func hasPtRegsSource(arg *v1alpha1.KProbeArg) bool {
...
 	if err != nil {
-		return nil, [api.MaxBTFArgDepth]api.ConfigBTFArg{}, err
+		return nil, [api.MaxBTFArgDepth]api.ConfigBTFArg{}, [api.MaxBTFArgDepth]string{}, err
 	}
 	ty := ebtf.Type(st)
 	return resolveBTFType(arg, ty)
 }
 
-func resolveBTFArg(hook string, arg *v1alpha1.KProbeArg, tp bool) (*ebtf.Type, [api.MaxBTFArgDepth]api.ConfigBTFArg, error) {
+func resolveBTFArg(hook string, arg *v1alpha1.KProbeArg, tp bool) (*ebtf.Type, [api.MaxBTFArgDepth]api.ConfigBTFArg, [api.MaxBTFArgDepth]string, error) {
```
olsajiri: there's one more user now added in https://github.com/cilium/tetragon/commit/7d4eafe0ce28c95de0fd3ba698e1ae283801afa2

bpf/process/types/basic.h
```diff
@@ -2029,6 +1966,10 @@ filter_arg_1(struct msg_generic_kprobe *e, struct selector_arg_filter *filter, c
 		if (filter->op == op_capabilities_gained) {
 			__u64 cap_old = *(__u64 *)args;
 			__u32 index2 = *((__u32 *)&filter->value);
+
+			asm volatile("%[index2] &= 0x7;\n" : [index2] "+r"(index2));
+			if (e->arg_error_status[index2])
+				return 0;
```
olsajiri, line 2012: don't we check this already in selector_arg_offset ? what do I miss?

bpf/process/generic_calls.h
```diff
@@ -472,28 +486,38 @@ extract_arg_depth(u32 i, struct extract_arg_data *data)
...
 	if (i >= MAX_BTF_ARG_DEPTH || !data->btf_config[i].is_initialized)
 		return 1;
+
 	*data->arg = *data->arg + data->btf_config[i].offset;
-	if (data->btf_config[i].is_pointer)
-		probe_read((void *)data->arg, sizeof(char *), (void *)*data->arg);
+
+	if (data->btf_config[i].is_pointer) {
+		if (probe_read((void *)data->arg, sizeof(char *), (void *)*data->arg) < 0) {
+			*data->arg_error_status = i + 1;
```
olsajiri: should we store in `arg_error_status` error value from `probe_read` instead?

bpf/process/generic_calls.h
```diff
@@ -1352,9 +1361,11 @@ FUNC_INLINE int generic_retprobe(void *ctx, struct bpf_map_def *calls, unsigned
 
 	switch (do_copy) {
 	case char_buf:
+		size = write_arg_status(e, size, 0);
```
olsajiri, line 1230: could you split the initial change - the NULL resolve error - and the rest of the changes like this one?

bpf/lib/generic.h
olsajiri: maybe shorter name like  `arg_status` would be better?

bpf/process/generic_calls.h
```diff
@@ -442,15 +452,19 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
...
+	args = args_off(e, orig_off);
 
 	path_arg = get_path(type, arg, &path_buf);
 	if (path_arg)
-		return copy_path(args, path_arg);
-
+		return copy_path(args, path_arg) + sizeof(__u32);
 	/*
 	 * Separate argument processing based on the process const
 	 * for 4.19 kernels..
```
olsajiri, line 473: I'm still staring on that, but I wonder we could also do something like this:
(this is by no means functional code, just to show what I mean)

```
@@ -470,15 +470,22 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
         * for 4.19 kernels..
         */
        if (process == __READ_ARG_1)
-               return __read_arg_1(ctx, type, orig_off, arg, argm, args);
+               err = __read_arg_1(ctx, type, orig_off, arg, argm, args);
        if (process == __READ_ARG_2)
-               return __read_arg_2(ctx, type, orig_off, arg, argm, args);
+               err = __read_arg_2(ctx, type, orig_off, arg, argm, args);
 
        /* .. and the rest of the world */
        if (is_read_arg_1(type))
-               return __read_arg_1(ctx, type, orig_off, arg, argm, args);
+               err = __read_arg_1(ctx, type, orig_off, arg, argm, args);
        else
-               return __read_arg_2(ctx, type, orig_off, arg, argm, args);
+               err = __read_arg_2(ctx, type, orig_off, arg, argm, args);
+
+       if (err < 0) {
+                write_arg_status(e, saved_orig_off, e->arg_error_status[index]);
+                e->arg_error_status[index] = err;
+                return sizeof(__u32);
+       }
+       return err;
+
 }
```

that would mean first convert all arguments to return < 0 in case of error,
also maybe above could also be in separate function

pkg/sensors/tracing/generic.go
andrewstrohman: Thanks for the heads up. I fixed this upon rebase.

bpf/process/types/basic.h
andrewstrohman: The `CapabilitiesGained` operator is the only operator that takes more than one `args` value. The first arg was checked in `selector_arg_offset`, and the second arg is being checked here.  If either one could not be read, we should default to not matching on the selector.

bpf/process/generic_calls.h
andrewstrohman: We need to store the "level" at which it failed, because that's what's used to know which dereference failed. Resolve can traverse multiple pointers, so without this information I won't be able to pinpoint which pointer failed to derefernce.

IMO, the return value of `probe_read` is not very useful. I think it's either -EFAULT or -EINVAL. But I could include this information in addition to the "level", if you think that would be useful.

bpf/process/generic_calls.h
andrewstrohman: Will do. I interpret this to mean that you want to the change that alters the selector outcome based on resolve failure to be separate from the change the communicates that failure to userspace. Please let me know if I misunderstood.

bpf/lib/generic.h
andrewstrohman: Sounds good.

bpf/process/generic_calls.h
andrewstrohman: In the @tdaudi 's [issue](https://github.com/cilium/tetragon/issues/3728) he talks about the case where events are dropped. See [here](https://github.com/cilium/tetragon/issues/3728#issuecomment-3525223316) for my analysis on this. From a recent, unrelated group slack chat about uprobe events, I know that you are aware of this problem too. (I'm talking about how we return from `generic_process_event` when these functions return a negative value.) So, I agree it would be good to do this for that reason.

However, this PR has been open for a while and scope has already been increased significantly beyond what was originally envision. So I would like to ask that I work on this needed, but separate, aspect that you're bringing up here in a follow-up PR.

With that being said, I do think that now is the right time to discuss how to communicate those `__read_arg_#()` failures and how distinguished between `__read_arg_#()` failures and failure brought on by dereferencing during resolve.

Like I mentioned in an above response, I want to record the depth or level at which the resolve failure occurs so I can pinpoint which pointer failed to dereference. With this in mind, how should we represent these `__read_arg_#()` failures? In many cases, it makes sense to assign arg_status "1", because the first pointer failed. But maybe there is a better way to unify or separate these different cases. If we have a plan on how to move forward in this aspect, I think we can be confident that this change puts us in the right direction for the future.

What do you think?

bpf/process/types/basic.h
olsajiri: ah it's the other arg, ok

bpf/process/generic_calls.h
olsajiri: actualy I think it's ok, I got confused seeing this change in generic_retprobe and thought you added extra checks and  but I guess it needs to be there

bpf/lib/generic.h
olsajiri: so this is needed before we access argument data in actions and other parts

```
+               if (e->arg_error_status[name])
+                       return 0;
```

I wonder we could add function for that with descriptive name like

```
+               if (!is_arg_ok(e, name))
+                       return 0;
```

bpf/process/generic_calls.h
olsajiri: > In the @tdaudi 's [issue](https://github.com/cilium/tetragon/issues/3728) he talks about the case where events are dropped. See [here](https://github.com/cilium/tetragon/issues/3728#issuecomment-3525223316) for my analysis on this. From a recent, unrelated group slack chat about uprobe events, I know that you are aware of this problem too. (I'm talking about how we return from `generic_process_event` when these functions return a negative value.) So, I agree it would be good to do this for that reason.
> 
> However, this PR has been open for a while and scope has already been increased significantly beyond what was originally envision. So I would like to ask that I work on this needed, but separate, aspect that you're bringing up here in a follow-up PR.

yes, I meant that as a follow up.. I think let's finish error report for resolve and then figure out what we want to do for the rest

> 
> With that being said, I do think that now is the right time to discuss how to communicate those `__read_arg_#()` failures and how distinguished between `__read_arg_#()` failures and failure brought on by dereferencing during resolve.
> 
> Like I mentioned in an above response, I want to record the depth or level at which the resolve failure occurs so I can pinpoint which pointer failed to dereference. With this in mind, how should we represent these `__read_arg_#()` failures? In many cases, it makes sense to assign arg_status "1", because the first pointer failed. 
> But maybe there is a better way to unify or separate these different cases. If we have a plan on how to move forward in this aspect, I think we can be confident that this change puts us in the right direction for the future.
> 
> What do you think?

I agree,  let's have the level value now and see how useful that will be and if we want to add error value in future

bpf/process/generic_calls.h
olsajiri: ok, let's keep the level value for now

bpf/process/generic_calls.h
andrewstrohman: OK, I'll squash the broken commit back together.

bpf/lib/generic.h
andrewstrohman: Sounds good. Done!

bpf/process/generic_calls.h
andrewstrohman: I've created [this issue](https://github.com/cilium/tetragon/issues/4579) and assigned it to myself for follow up, so I don't forget about this.

bpf/process/types/basic.h
```diff
@@ -1962,6 +1966,10 @@ filter_arg_1(struct msg_generic_kprobe *e, struct selector_arg_filter *filter, c
 		if (filter->op == op_capabilities_gained) {
 			__u64 cap_old = *(__u64 *)args;
 			__u32 index2 = *((__u32 *)&filter->value);
+
+			asm volatile("%[index2] &= 0x7;\n" : [index2] "+r"(index2));
```
kkourt: This makes me somewhat scared that if we ever change MAX_POSSIBLE_ARGS to something larger than 7, the code above will silently break.

How about:

```diff
diff --git a/bpf/lib/generic.h b/bpf/lib/generic.h
index 00e751a80..0734d8f0b 100644
--- a/bpf/lib/generic.h
+++ b/bpf/lib/generic.h
@@ -21,6 +21,10 @@
 #define SELECTORS_ACTIVE	 31
 #define MAX_CONFIGURED_SELECTORS MAX_POSSIBLE_SELECTORS + 1
 
+/* convenience mask for verifier appeasing*/
+#define MAX_POSSIBLE_ARGS_MASK 0x7
+_Static_assert(MAX_POSSIBLE_ARGS_MASK >= MAX_POSSIBLE_ARGS, "Need to update MAX_POSSIBLE_ARGS_MASK");
+
 struct msg_selector_data {
 	__u64 curr;
 	bool pass;
diff --git a/bpf/process/types/basic.h b/bpf/process/types/basic.h
index 5bb827070..85f138165 100644
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -1967,7 +1967,9 @@ filter_arg_1(struct msg_generic_kprobe *e, struct selector_arg_filter *filter, c
 			__u64 cap_old = *(__u64 *)args;
 			__u32 index2 = *((__u32 *)&filter->value);
 
-			asm volatile("%[index2] &= 0x7;\n" : [index2] "+r"(index2));
+			asm volatile("%[index2] &= %[mask];\n"
+				: [index2] "+r"(index2)
+				: [mask] "i"(MAX_POSSIBLE_ARGS_MASK));
 			if (!is_arg_ok(e, index2))
 				return 0;
 			__u64 cap_new = *(__u64 *)get_arg(e, index2);
```

kkourt: From what I can see in the code, there are a few more instances of the above pattern but we can fix them as a followup PR.

bpf/process/types/basic.h
kkourt, line 2209: Which reminds me that all of this code is deprecated. Created https://github.com/cilium/tetragon/issues/4580

bpf/process/generic_calls.h
```diff
@@ -409,7 +416,7 @@ __read_arg_2(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 		size = 0;
 		break;
 	}
-	return size;
+	return size + sizeof(__u32);
```
kkourt: Reading the PR, I understand that those `+ sizeof(__u32)` are accounting for the space of the status header in arg byte-array, but reading the code without the PR in mind makes it difficult to discern their purpose.

Would it make sense to add a comment?
If we don't want to be too verbose, I think something like
```
typedef __u32 arg_status_t;
...
return size + sizeof(arg_status_t);
```

Will also make things more clear in the code.
WDYT?

andrewstrohman: That's a good idea. I created the typedef and used it.

bpf/process/types/basic.h
andrewstrohman: That's a good point. I do think `MAX_POSSIBLE_ARGS_MASK >= MAX_POSSIBLE_ARGS` is off by one though,
So I changed it to `MAX_POSSIBLE_ARGS_MASK >= MAX_POSSIBLE_ARGS - 1`.

I went though and tried to use MAX_POSSIBLE_ARGS_MASK as much as possible.

I consolidated EVENT_CONFIG_MAX_ARG and MAX_POSSIBLE_ARGS to one macro because these two values cannot deviate, and there are cases were we mask an index to be used for both.

I found that MAX_SELECTORS_MASK was being used for MAX_POSSIBLE_ARGS sized arrays, so I fixed that as well.

Finally, there is a case where we mask an arg index to be used for two different contexts: arg config index, and arg function signature index. I tried to decouple this as well.

I added two commits to the series to address these things. I see you suggested follow-up PR, so let me know if you prefer that I move these two these commits to a separate PR.

kkourt: Thanks.

I think it's OK to have it in the same PR.

kkourt: We  can also leave it as a follow-up since I'm seeing some test failures for the rhel kernel.

andrewstrohman: The rhel failures are confusing to me. When I run the rhel VM locally, regardless of whether I use main or this branch, `TestUprobeArgsMatch` passes for me. But for the other 3 tests that fail the CI, they fail locally for me regardless of whether I run main or this branch.

andrewstrohman: The problem with my local VMs was due to soft links and `MatchBinaries`. The problem with rhel was that it was not writing to memory (`e->arg_status[index] = 0;`) without an upper bound check, which I don't fully understand. I noticed that if I added a bpf_printk() to log `e->arg_status[index] `'s value immediately after it was initialized to 0, the problem when away. This clue lead me to a suggestion to do an upper bounds check (which also worked and is what I settled on).

bpf/lib/generic.h
```diff
@@ -63,8 +69,13 @@ struct msg_generic_kprobe {
 	__u64 user_stack_id; // User Stack trace ID
 	/* anything above is shared with the userspace so it should match structs MsgGenericKprobe and MsgGenericTracepoint in Go */
 	char args[24000];
+#define MAX_ACCESSIBLE_ARGS 5
+/* convenience mask for verifier appeasing*/
+#define MAX_ACCESSIBLE_ARGS_MASK 0x7
```
olsajiri: how is this one different from MAX_POSSIBLE_ARGS_MASK ?
also let's not add defines in the middle of the struct

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -852,12 +858,32 @@ func handleMsgGenericTracepoint(
...
 		}
 
+		err := binary.Read(r, binary.LittleEndian, &status)
+
+		if err != nil {
+			logger.GetLogger().Warn(fmt.Sprintf("Arg status size type error sizeof %d", m.Common.Size), logfields.Error, err)
+			break
+		}
+
+		if status != 0 {
```
olsajiri: looks like we could have function for this and use it in here and in getArg?

pkg/btf/btf.go
```diff
@@ -341,15 +370,18 @@ func processMembers(
 	}
 	if t, ok := currentType.(*btf.Pointer); ok {
 		btfArgs[i].IsPointer = uint16(1)
+		(*btfPtrNames)[i] = GetPtrName(pathToFound[:i])
```
olsajiri: do we need to get this info from in here? seems like it could be retrieved within `resolveBTFType` from `path` cc @tdaudi

tdaudi: I'm not sure I understand why it is necessary ?
If you take `strings.Split(arg.Resolve, ".")`, don't we get the same array as btfPtrNames ? I probably misunderstand something, so apologies if I'm wrong.

tdaudi: The `arg.Resolve` is a bit tweaked before going through the Resolve algo, but the "final" might be similar to the steps you try to resolve
https://github.com/cilium/tetragon/blob/645d787c9c5acd4f89e59744aae3ae0a72477791/pkg/sensors/tracing/generic.go#L98-L103

bpf/process/types/basic.h
andrewstrohman: Per @olsajiri's request, I have moved these changes into a [different PR](https://github.com/cilium/tetragon/pull/4607).

andrewstrohman:
> also there seems to be some unrelated fixes, which could go in separate PR (faster)?
> 
> * the MAX_ARGS stuff
> * tetragon: make GetIndex() consistent

I've removed these commits from this series. The max args stuff is now [here](https://github.com/cilium/tetragon/pull/4607) and the returnCopy fix is now [here](https://github.com/cilium/tetragon/pull/4608).

bpf/lib/generic.h
andrewstrohman: > how is this one different from MAX_POSSIBLE_ARGS_MASK ?

MAX_ACCESSIBLE_ARGS is the max number of accessible (reachable) args within the function signature. We cannot read the sixth argument of a hooked function because we only support reading the first 5.

MAX_POSSIBLE_ARGS_MASK is the max number of configured arguments in the tracing policy. This also happens to be 5, but these two concepts could deviate in the future. For example, we might support reaching deeper arguments within the function signature  (ie sixth arg) without increasing the maximum amount of args allowed to be configured within the tracing policy.

> also let's not add defines in the middle of the struct

As you've requested, I've moved these commits to a [different PR](https://github.com/cilium/tetragon/pull/4607), so let's continue this conversation there. I've moved these defines above the struct definition, and left a comment about how they are related to the a[0-4] members of the struct.

pkg/sensors/tracing/generictracepoint.go
andrewstrohman: Sounds good. I did this.

pkg/btf/btf.go
andrewstrohman: I found the resolve configuration logic to be extremely difficult to follow and reason about. Also, I know from previous conversations that there are some cases that don't currently work right. 

Because of this, I decided to follow (bolt onto) the resolve configuration logic, creating a mapping from the depths that require a dereference, to the name of the pointer that caused the deference to happen. This way, I can be sure that I have the right names available for these depths without have to "externally" reason about the resolve configuration outcome after it's concluded. Because of the complexity of how resolve configuration is determined, I don't think that there is an appropriate pre-existing data structure that can give me an accurate pointer name.

So, I would appreciate if you guys could be very specific with your suggestions here. I'm happy to try them out and report back what happens.

andrewstrohman: > If you take strings.Split(arg.Resolve, "."), don't we get the same array as btfPtrNames ? I probably misunderstand something, so apologies if I'm wrong.

The dots are ambiguous. They can mean either `.` or `->`. So, I don't think this method cleanly maps resolve depth to variable names.

bpf/process/generic_calls.h
```diff
@@ -472,28 +487,38 @@ extract_arg_depth(u32 i, struct extract_arg_data *data)
...
 	if (i >= MAX_BTF_ARG_DEPTH || !data->btf_config[i].is_initialized)
 		return 1;
+
 	*data->arg = *data->arg + data->btf_config[i].offset;
-	if (data->btf_config[i].is_pointer)
-		probe_read((void *)data->arg, sizeof(char *), (void *)*data->arg);
+
+	if (data->btf_config[i].is_pointer) {
+		if (probe_read((void *)data->arg, sizeof(char *), (void *)*data->arg) < 0) {
+			*data->arg_status = i + 1;
```
tdaudi: I try to understand a bit more the code to give a better answer to https://github.com/cilium/tetragon/pull/4327/changes#r2767188698

I know this piece of code was made quite some time ago, but do you recall why you do `i + 1` here but later in userspace you do `status-1` (in `arg_linux.go`) ?

andrewstrohman: @olsajiri requested that I used an unsigned integer for the status. So, zero represents that the arg was able to be read without issue. Any non-zero represents that an error occurred during reading the arg. Further, that non-zero value communicates the depth at which the error occurred. Since dereference can fail at depth zero, the depth reported from the bpf program needs to have one added to it to distinguish between the success case and the failure at depth zero case.

pkg/btf/btf.go
tdaudi: ## Method A

So the test pass with the following diff.

```diff
diff --git a/pkg/sensors/tracing/generic.go b/pkg/sensors/tracing/generic.go
index 454a8bd77..c6846a239 100644
--- a/pkg/sensors/tracing/generic.go
+++ b/pkg/sensors/tracing/generic.go
@@ -103,6 +103,7 @@ func resolveBTFType(arg *v1alpha1.KProbeArg, ty ebtf.Type) (*ebtf.Type, [api.Max
        }
 
        lastBTFType, err := resolveBTFPath(&btfArg, &btfPtrName, btf.ResolveNestedTypes(ty), path)
+       copy(btfPtrName[1:], path)
        return lastBTFType, btfArg, btfPtrName, err
 }
 
diff --git a/pkg/sensors/tracing/uprobe_reg_test.go b/pkg/sensors/tracing/uprobe_reg_test.go
index adc3db935..93afd0c05 100644
--- a/pkg/sensors/tracing/uprobe_reg_test.go
+++ b/pkg/sensors/tracing/uprobe_reg_test.go
@@ -766,7 +766,7 @@ func TestUprobeResolveNull(t *testing.T) {
                        ec.NewKprobeArgumentChecker().WithErrorArg(ec.NewKprobeErrorChecker().WithMessage(sm.Full("failed to dereference second"))),
                }},
                {"third", []*ec.KprobeArgumentChecker{
-                       ec.NewKprobeArgumentChecker().WithErrorArg(ec.NewKprobeErrorChecker().WithMessage(sm.Full("failed to dereference second.third"))),
+                       ec.NewKprobeArgumentChecker().WithErrorArg(ec.NewKprobeErrorChecker().WithMessage(sm.Full("failed to dereference third"))),
                }},
                {"nonull", []*ec.KprobeArgumentChecker{
                        ec.NewKprobeArgumentChecker().WithIntArg(0),
```

What I do not understand is why the `1` is needed in `copy(btfPtrName[1:], path)`. Because you try to resolve 3 steps, the first is for the `second` field, then `third` and then `val`. 
The []api.ConfigBTFArg has 3 steps which corresponds to what I'm saying. But in you test you have 4 steps. But if I remove the  `first` entry in `TestUprobeResolveNull` test and remove this `1` in the copy, the test fails but I don't get it. Do you have an idea ?

## Method B

Alternatively I think your approach is more robust and it will help for futur improvements, like with this PR https://github.com/cilium/tetragon/pull/4464.

Though I personally would have prefer wrapping the `btfArg` in a bigger struct where you can store your additional data.

Here is a pseudo-code diff just to be more pragmatic

```diff
diff --git a/pkg/btf/btf.go b/pkg/btf/btf.go
index 59f6cd5a7..de0abca00 100644
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -254,6 +254,11 @@ func GetPtrName(parts []string) string {
        return ret.String()
 }
 
+type BTFStep struct {
+       Name string
+       Resolve api.ConfigBTFArg
+}
+
 // ResolveBTFPath function recursively search in a btf structure in order to
 // found a specific path until it reach the target or fail.
 
@@ -270,7 +275,7 @@ func GetPtrName(parts []string) string {
 
 // Return: The last type found matching the path, or error.
 func ResolveBTFPath(
-       btfArgs *[api.MaxBTFArgDepth]api.ConfigBTFArg,
+       btfSteps *[api.MaxBTFArgDepth]BTFStep,
        btfPtrNames *[api.MaxBTFArgDepth]string,
        currentType btf.Type,
        pathToFound []string,
@@ -318,7 +323,7 @@ func ResolveBTFPath(
 }
 
 func processMembers(
-       btfArgs *[api.MaxBTFArgDepth]api.ConfigBTFArg,
+       btfArgs *[api.MaxBTFArgDepth]BTFStep,
        btfPtrNames *[api.MaxBTFArgDepth]string,
        currentType btf.Type,
        members []btf.Member,
@@ -347,8 +352,9 @@ func processMembers(
                }
                if member.Name == pathToFound[i] {
                        memberWasFound = true
-                       btfArgs[i].Offset = member.Offset.Bytes()
-                       btfArgs[i].IsInitialized = uint16(1)
+                       btfArgs[i].Resolve.Offset = member.Offset.Bytes()
+                       btfArgs[i].Resolve.IsInitialized = uint16(1)
+                       btfArgs[i].Name = pathToFound[i]
                        isNotLastChild := i < len(pathToFound)-1 && i < api.MaxBTFArgDepth
                        if isNotLastChild {
                                return ResolveBTFPath(btfArgs, btfPtrNames, ResolveNestedTypes(member.Type), pathToFound, i+1)
diff --git a/pkg/sensors/tracing/generic.go b/pkg/sensors/tracing/generic.go
index 454a8bd77..f2c99188f 100644
--- a/pkg/sensors/tracing/generic.go
+++ b/pkg/sensors/tracing/generic.go
@@ -154,6 +154,7 @@ func resolveBTFArg(hook string, arg *v1alpha1.KProbeArg, tp bool) (*ebtf.Type, [
 }
 
 func resolveBTFPath(btfArg *[api.MaxBTFArgDepth]api.ConfigBTFArg, btfPtrName *[api.MaxBTFArgDepth]string, rootType ebtf.Type, path []string) (*ebtf.Type, error) {
+       // Add a function here that separate the Name from Resolve in 2 different arrays
        return btf.ResolveBTFPath(btfArg, btfPtrName, rootType, path, 0)
 }
```

Sorry if I can't more specific here, I do not have much time to spent on dev currently.
I hope it helps a bit.

andrewstrohman: Thanks for this explanation. I went for Method A, which is the same thing that Jiri was asking about in the beginning of this thread.

> What I do not understand is why the 1 is needed in copy(btfPtrName[1:], path). Because you try to resolve 3 steps, the first is for the second field, then third and then val.
The []api.ConfigBTFArg has 3 steps which corresponds to what I'm saying. But in you test you have 4 steps. But if I remove the first entry in TestUprobeResolveNull test and remove this 1 in the copy, the test fails but I don't get it. Do you have an idea ?

The policy config is:
```
btfType: "first"
resolve: "second.third.val"
```
When we pass "third" to the test program, we expect message "failed to dereference second.third", because passing "third" causes `second.third = NULL;`.

In this scenario, path looks like ["second", "third", "val"].

As you already mentioned: btfArg=[{0 1 1} {0 1 1} {0 1 1}],
```
type ConfigBTFArg struct {
	Offset        uint32 `align:"offset"`
	IsPointer     uint16 `align:"is_pointer"`
	IsInitialized uint16 `align:"is_initialized"`
}
```
Like you said, it's three configs for 3 pointers. It is the final pointer that we cannot dereference (at index 2). The path component at index 2 is "val", but "val" is not a pointer -- "third" is the last pointer. This is why you need to shift things right by one.

Another way to think about it, is that the first pointer variable name doesn't have its name in `path` (the variable name is not found in the policy at all). This why we have `failed to dereference pointer` outcome. This is also why you needed to comment out the `first` test case.

I need to be honest here. Despite the fact that CI is passing, I don't feel confident that my current method of determining the name of the pointer will work for every valid resolve configuration, because I still haven't developed a strong intuition about how resolve configuration creation is supposed to work. I get the sense from this conversation that we expect an instance of ConfigBTFArg struct to correspond to the tokens after splitting on every dot or array index in the resolve string. This approach hinges on this, and I'm just not sure if this is always true.

From code inspection, I don't think this current approach will work if the argument is a nested pointer. Incidentally, I don't think that resolve, itself, will work as expected if it encounters a nested pointer in the middle a resolve string.

I'm worried about different combinations that haven't been tested -- nested pointers, arrays, and, having non-pointers in the middle of the resolve path.

Yes, I can add more unit tests to gain confidence, but this is getting too far removed from my original goal with this PR. So, I would like to request that I remove the commit that translates the depth to the string, and  put that commit in a follow-up PR that incorporates more unit testing, and possibly changes to how resolve works itself. If we do it this way, only the depth will be reported in the error message until the follow-up PR is merged.

tdaudi: > The path component at index 2 is "val", but "val" is not a pointer -- "third" is the last pointer. This is why you need to shift things right by one.

In the original implementation of the Resolve algo, we blindly tested the first `probe_read` in `extract_arg_depth`, but we never take the information if the initial pointer (`first` in our case), is actually non-null. So, did you add this extra prefix for that reason ? I just want to be sure that if there are 10 entries to resolve, nothing unexpected happen with `btfPtrName`.

> From code inspection, I don't think this current approach will work if the argument is a nested pointer. Incidentally, I don't think that resolve, itself, will work as expected if it encounters a nested pointer in the middle a resolve string.

If the nested ptr is at the beginning of the path, it should be prefixed with `[0]`.
https://github.com/cilium/tetragon/blob/46b1886880e7e6c16a7b1894cae2d4189f51ac28/pkg/sensors/tracing/generic.go#L78

When it's a nested pointer inside of the resolve, I believe it should fail here if the user set `[XX]` manually. 
https://github.com/cilium/tetragon/blob/46b1886880e7e6c16a7b1894cae2d4189f51ac28/pkg/btf/btf.go#L264-L268

Though, I agree with you, there is a hole if the user does not set this array brackets, I think there is a bug logic in the resolve algo because no new entry is added but we still do `t.Target` (which indicates a dereference).
https://github.com/cilium/tetragon/blob/46b1886880e7e6c16a7b1894cae2d4189f51ac28/pkg/btf/btf.go#L269

> Yes, I can add more unit tests to gain confidence

That make sense, I also feel like it's needed.

andrewstrohman: >  So, did you add this extra prefix for that reason ? 

I didn't add a prefix. When you were experimenting, you added a prefix to shift the components of the path to the right. I'm accomplishing the same thing by stopping one index to the left of the depth when I construct the full name.

Consider the case where it's the first pointer that cannot dereference. We shouldn't construct the string from the first element of path in this case. path doesn't include a string that represents that that first variable name that is dereferenced, because that variable name corresponding to that failed dereference does not appear in the policy at all. This base case should give you an intuition about why I stop combining the string one index to the left of the depth, which is the same reason why you needed to do `1:`. 

> If the nested ptr is at the beginning of the path, it should be prefixed with [0].

Yes, that's why I'm saying the code I currently have won't work, as-is. The string would start with a [0].

If we are OK with dropping this commit for now, I will add more tests in the follow-up PR to cover the cases that I'm worried about.

kkourt: > I found the resolve configuration logic to be extremely difficult to follow and reason about. Also, I know from previous conversations that there are some cases that don't currently work right.

I would agree with the above statements. I haven't followed all the details, but my take is that we should probably rethink the resolve implementation so that it's clearer (with fewer corner cases). That being said, I suggest we prioritize merging this PR since it's a clear improvement and leave the rethinking of resolve as a follow-up PR as suggested by Andy.

tdaudi: > I didn't add a prefix. When you were experimenting, you added a prefix to shift the components of the path to the right. I'm accomplishing the same thing by stopping one index to the left of the depth when I construct the full name.

I'm not sure this is how the resolve works. In the BPF code, I don't see a check on the initial struct to make sure it's not null (`first`), so you have to rely on the `N + 1` to determine that `N` is null which is not the correct approach IMO. So it means you determine that `first` is null when resolving second in the test. If you modifiy `tester-progs/uprobe-null.c` and add non-pointers between `first` and `second`, when running `./uprobe-null first`, the resolve output will be this new field. I think the approach that works here is check `first` is null before entering the `extract_arg_depth` and if you do, you rely on the `path` to have the appropriate fields.

Maybe this can be assign in a bugfix followup PR, but that's not for me to say.

andrewstrohman: To recap, I initially wanted to just deliver the depth at which the dereferenced failed. However, I was encouraged to convert this to a string that represents which pointer could not be dereferenced. This is a good suggestion in that it's something that the user can understand, whereas just reporting the depth is not as helpful in  terms of being able to determine which pointer could not be dereferenced. I think this was originally suggested because we thought that this mapping already existed and was easily available to use for translation.

Given the conversation in this this thread, I think we all agree that it's not a simple as we initially hoped. Being able to prevent bogus argument values in the event and avoiding false selector matches is a valuable step forward in itself. As such, I have removed the the commit that does the translation from depth to the string that represents the pointer that could not dereference.

We can continue this conversation in a follow-up PR where we translate the resolve failure's depth to a string that represents the pointer that could not be dereferenced.

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -858,6 +858,13 @@ func handleMsgGenericTracepoint(
 			continue
 		}
 
+		if errorArg, isReadErr := getArgStatus(r); isReadErr {
```
olsajiri: we do the error/warn log in this function, so could getArgStatus returl (arg, error) instead?
also let's use just `err` instead of `isReadErr`

bpf/process/types/basic.h
```diff
@@ -291,7 +224,7 @@ args_off(struct msg_generic_kprobe *e, unsigned long off)
 {
 	asm volatile("%[off] &= 0x3fff;\n"
 		     : [off] "+r"(off));
-	return e->args + off;
+	return e->args + ((unsigned int)off & 0x3fff);
```
olsajiri: makes me wonder, do we still need that asm statement above then?

pkg/sensors/tracing/generictracepoint.go
andrewstrohman: Sounds good. I did both of these things.

bpf/process/types/basic.h
andrewstrohman: I just doubled checked this. On 5.4, I get the following error when I tried removing the above asm statement:
> program generic_uprobe_process_event: load program: invalid argument: math between map_value pointer and register with unbounded min value is not allowed

andrewstrohman:
> please add some changelog to following commits:

I did this.

> otherwise looks good, let's merge it soon ;-)

I really want to merge this, but this most recent rebase exposed a conflict between preload logic and this change. Now that extract_arg is executed in preload for uprobe arg strings, I need a way to communicate the depth of those resolve failures to the "normal" (not sleepable) arg reading code, so that the arg status can be set appropriately. This is my top priority and I will follow up here once I have a solution for it.

UPDATE: I handled the above described conflict by relaying the arg status within `struct preload_data`. So it turned out to be pretty straight forward.

While analyzing this, I saw a problem with a change that I recently merged. See [here](https://github.com/cilium/tetragon/pull/4638) for the PR to fix it. I'm not sure if I should continue keeping that PR separate, or if it makes more sense to instead include it in this series. That fix is about preload arg reading needing to run `extract_arg` for the `pt_regs` case too. Once that's in place, I'll update the `pt_regs` preload case to store the extract_arg error depth status into `struct preload_data` as well.

andrewstrohman:
@olsajiri after the latest rebase, I was having verifier issues that affects 4.19 only. So I added an additional commit to this series to show what I had to do to overcome it. After fixing 4.19 with repetitive masking, then 5.4 started complaining, so I had to add yet another mask for 5.4. If you have any suggestions about how to potentially avoid this, I would appreciate it. I don't see this repetitive masking elsewhere in the code.

pkg/sensors/tracing/generictracepoint.go
olsajiri: hum I still see the log message inside `getArgStatus`

andrewstrohman: OK, I've moved the logging to the call site of `getArgStatus`.

andrewstrohman:
I re-investigated all the additional masking and bounds checks that I originally added, and was able to reduce it significantly. I first got 4.19 working. Then, I add a commit that added the changes required for 5.4. Finally, I added a commit specific to rhel8.9. I think this is the best I can do in this regard.

Issue #4260, opened 2025-10-28
Allow configuring policy filter map size


kyledong-suse:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

Currently, Tetragon limits the number of concurrent policies to 128 in the ebpf program for `policy_filter_maps`:
```
#define POLICY_FILTER_MAX_POLICIES   128

struct {
	__uint(type, BPF_MAP_TYPE_HASH_OF_MAPS);
	__uint(max_entries, POLICY_FILTER_MAX_POLICIES);
	__type(key, u32); /* policy id */
	__array(
		values, struct {
			__uint(type, BPF_MAP_TYPE_HASH);
			__uint(max_entries, 1);
			__type(key, __u64); /* cgroup id */
			__type(value, __u8); /* empty  */
		});
} policy_filter_maps SEC(".maps");
```

The current cap is too restrictive. Users operating with large policy sets can quickly reach this limit, causing subsequent policies to be dropped or rejected.

### Describe the feature you would like

The idea is to make the policy filter map size configurable like [execve_map](https://github.com/cilium/tetragon/blob/main/bpf/lib/process.h#L363)

### Describe your proposed solution

- Replace the fixed `__uint(max_entries, POLICY_FILTER_MAX_POLICIES)` with `__uint(max_entries, 1)` in bpf/process/policy_filter.h
- Add a new option flag "policy-filter-map-entries" refers to existing implementation for [execve-map-entries](https://github.com/cilium/tetragon/blob/main/pkg/option/flags.go#L134)
- Default the map entries to 128 to remain backward compatible
- Let the agent set `spec.Maps["policy_filter_maps"].MaxEntries` before program is loaded, based on the new configuration value

This proposal will

- Removes the static 128-policy cap.
- Enables dynamic scaling of policy capacity and gives users the flexibility to configure limits according to their deployment scale.
- No verifier or eBPF logic changes required.
- Backward compatible (defaults remain unchanged).

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

kyledong-suse:
Hi @mtardy, just wanted to follow up on this issue to see if there’s been any update or thoughts from the team.

I’ve been revisiting this recently and still seeing the same limitation. Wondering if this is on the radar or if there’s a known workaround. If it helps, I can try to provide more logs or a minimal reproducer. Please let me know what would be most useful.

Thanks again for your time!

kkourt:
[...]

PR #4331, opened 2025-11-11
pkg/option: allow policy-filter-map-entries configurable via flag

kyledong-suse:
### Description

This commit introduces a new flag to configure the number of entries in policy filter maps. This allows users to tune the map size based on workload scale and system resources, improving flexibility in policy handling.

Note: this commit only affects policies with k8s segmentation primitives (i.e., either podSelectors or namespaced policies).

Fixes: #4260

### DIFF
--- a/pkg/policyfilter/map.go
+++ b/pkg/policyfilter/map.go
@@ -85,7 +85,7 @@ func newPfMap(enableCgroupMap bool) (PfMap, error) {
 	if enableCgroupMap {
 		if ret.cgroupMap, err = openMap(spec, CgroupMapName, polMaxPolicies); err != nil {
 			releaseMap(ret.policyMap)
-			return PfMap{}, fmt.Errorf("opening cgroup map %s failed: %w", MapName, err)
+			return PfMap{}, fmt.Errorf("opening cgroup map %s failed: %w", CgroupMapName, err)
 		}
 	}
 


kyledong-suse:
@mtardy and @kkourt, would you please review this PR when you get a chance. Thanks!

bpf/process/policy_filter.h
```diff
@@ -20,7 +20,7 @@ struct {
 
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH_OF_MAPS);
-	__uint(max_entries, POLICY_FILTER_MAX_POLICIES);
+	__uint(max_entries, 1); // will be resized by agent when needed
```
mtardy, line 23: This is used by `policy_filter_cgroup_maps` just nearby. Maybe we would need to update both and remove that const?

pkg/sensors/program/loader_linux.go
```diff
@@ -976,6 +977,12 @@ func doLoadProgram(
 		}
 	}
 
+	// Set MaxEntries for policy_filter_maps if it exists in the spec.
+	// This ensures the spec matches the user-defined value.
+	if ms, ok := spec.Maps["policy_filter_maps"]; ok {
+		ms.MaxEntries = uint32(option.Config.PolicyFilterMapEntries)
+	}
+
```
mtardy: ah yes already have a facility to do that higher level in the loader (it's just above your new code btw), please use the `SetMaxEntries` on the map itself directly, you need to have the map handle.

pkg/policyfilter/map.go
```diff
@@ -72,6 +73,10 @@ func newPfMap(enableCgroupMap bool) (PfMap, error) {
 		return PfMap{}, fmt.Errorf("loading spec for %s failed: %w", objPath, err)
 	}
 
+	if _, ok := spec.Maps["policy_filter_maps"]; ok {
+		spec.Maps["policy_filter_maps"].MaxEntries = uint32(option.Config.PolicyFilterMapEntries)
+	}
```
mtardy, line 78: see other comment about using `SetMaxEntries`, you should see example by grepping

pkg/policyfilter/map.go
```diff
@@ -80,7 +85,7 @@ func newPfMap(enableCgroupMap bool) (PfMap, error) {
 	if enableCgroupMap {
 		if ret.cgroupMap, err = openMap(spec, CgroupMapName, polMaxPolicies); err != nil {
 			releaseMap(ret.policyMap)
-			return PfMap{}, fmt.Errorf("opening cgroup map %s failed: %w", MapName, err)
+			return PfMap{}, fmt.Errorf("opening cgroup map %s failed: %w", CgroupMapName, err)
```
mtardy, line 88: maybe this could be part of another commit fixing this typo?

pkg/option/flags.go
```diff
@@ -483,4 +487,6 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.String(KeyExecveMapSize, "", "Set size for execve_map table (allows K/M/G suffix)")
 
 	flags.Int(KeyRetprobesCacheSize, defaults.DefaultRetprobesCacheSize, "Set {k,u}retprobes events cache maximum size")
+
+	flags.Int(KeyPolicyFilterMapEntries, defaults.DefaultPolicyFilterMapEntries, "Set entries for policy_filter_map table (default 128)")
```
mtardy: could you be a little bit more explicit in the help what this map is for and why changing its size would matter (why you want to increase or decrease its size). Succinct is still better than extra verbose, but out of context on the map, it's not clear why this matters with this help (at least for me :)).

pkg/sensors/program/loader_linux.go
kyledong-suse: I was thinking to reuse `SetMaxEntries`. However, after digging into the code, I had the following findings:
- `policy_filter_maps` is not a `program.Map`.
- `policy_filter_maps` is created as raw `ebpf.Map` in [newPfMap()](https://github.com/cilium/tetragon/blob/main/pkg/policyfilter/map.go#L62), which is called during `policyfilter.New()`
- `policyfilter.GetState()` is called in [StartSensorManager()](https://github.com/cilium/tetragon/blob/main/pkg/sensors/manager.go#L35), which happens before sensors load.

So your suggestion is to refactor `policy_filter_maps` from `ebpf.Map` to `program.Map`? Also need to refactor this [PfMap](https://github.com/cilium/tetragon/blob/main/pkg/policyfilter/map.go#L27) struct.

pkg/policyfilter/map.go
kyledong-suse: Sure, I'll remove the typo fix from this commit. 
Do you like another commit to fix this typo in the same PR or a separate PR?

pkg/option/flags.go
kyledong-suse: WDYT about this one?
"Set maximum number of policies in policy_filter_maps (default 128). This map restricts tracing policies to specific pods/containers. Increase if you have many policies, decrease to save memory if you have few policies."

bpf/process/policy_filter.h
kyledong-suse: Sure, I think it makes sense to me. I'll update both of them and remove that const.

pkg/sensors/program/loader_linux.go
mtardy: Let's ask since we can before investing time in this :)!

Hey @tpapagian and @kkourt, I see you are the one you touched this code, any reason why you didn't use program.Map in the first place?

pkg/policyfilter/map.go
mtardy: same PR is fine :)

pkg/option/flags.go
mtardy: yep it's a nice suggestion! Don't need to repeat the default as cobra should display it in the help.

pkg/sensors/program/loader_linux.go
kkourt: Not reason I remember. Is the idea be to make the map a part of the base/exec sensor?

kyledong-suse: IMO, I think it makes sense to add `policy_filter_maps` to base sensor (similar to `execve_map`).

kyledong-suse: What do you think @kkourt and @tpapagian?

andreaterzolo: Sorry if I jump in, but AFAIK `policy_filter_maps` is a global map and its lifecycle is managed by the policy filter.  Updating the maxEntries in `pkg/policyfilter/map.go` as you did should be enough. Why are we bumping the number of entries also here?

andreaterzolo: Oh, I see, even if we use the `MapReplacements`, we need the maxEntries to match the original spec...directly from the cilium/ebpf doc

```
	// MapReplacements takes a set of Maps that will be used instead of
	// creating new ones when loading the CollectionSpec.
	//
	// For each given Map, there must be a corresponding MapSpec in
	// CollectionSpec.Maps, and its type, key/value size, max entries and flags
	// must match the values of the MapSpec.
	//
	// The given Maps are Clone()d before being used in the Collection, so the
	// caller can Close() them freely when they are no longer needed.
	MapReplacements map[string]*Map
```

Sorry for the noise

kyledong-suse: @mtardy @kkourt I’ve been thinking more about the refactoring in the past week. IMO, it’s better to create a separate issue and PR for that work. The current issue and PR focus on adding a user-configurable knob for the size of `policy_filter_maps`. If we include the refactoring in the same PR, it will introduce more churn and go beyond the original scope. What do you think?

bpf/process/policy_filter.h
kyledong-suse: @mtardy Given another think, the `POLICY_FILTER_MAX_POLICIES` in `policy_filter_cgroup_maps` represents how many policies can apply to a single cgroup (inner map size of `policy_filter_cgroup_maps`). 
I’m fine to simplify and drive both `policy_filter_maps` and `policy_filter_cgroup_maps` from the same `policy-filter-map-entries` setting and remove that const. However, I just want you to aware that if we want to update both, it would be mixing a global capacity knob with a per-cgroup limit, which could waste memory if we config more policies globally.

pkg/sensors/program/loader_linux.go
kkourt: Agreed! Let's create a follow-up issue.

kyledong-suse: Issue #4398 is created to track the refactoring

bpf/process/policy_filter.h
kyledong-suse: @mtardy Please feel free to add you comment. Thanks

kkourt: > I’m fine to simplify and drive both policy_filter_maps and policy_filter_cgroup_maps from the same policy-filter-map-entries setting and remove that const. However, I just want you to aware that if we want to update both, it would be mixing a global capacity knob with a per-cgroup limit, which could waste memory if we config more policies globally.

My reading is that that `policy_filter_cgroup_maps` uses `POLICY_FILTER_MAX_POLICIES` since it's the most pessimistic approach. This means, that there will be no capacity issues (i.e., a failure to insert) in `policy_filter_cgroup_maps` without hitting a capacity issue in `policy_filter_maps`.

I think there are three options here:

1. Have the run-time size to determine the size of policy_filter_maps and policy_filter_cgroup_maps. That is, have the run-time size instead of POLICY_FILTER_MAX_POLICIES.
2. Have a different run-time size for both. In this case, we would need to check the error path of when the capacity of `policy_filter_cgroup_maps` is not enough and do something reasonable. This can be a follow-up PR.
3. Leave it as is. In this case, we would still need to check the error path since we might end up with a case where the capacity of  `policy_filter_cgroup_maps` is not enough.

Note that option (2.) can be done in a followup PR after (1.).

CCing @tpapagian who I believe introduced this change.

kyledong-suse: My preference would be to leave it as-is for now, and keep `policy_filter_cgroup_maps` sized using `POLICY_FILTER_MAX_POLICIES`.

A few reasons for this:
- Using `POLICY_FILTER_MAX_POLICIES` for `policy_filter_cgroup_maps` is intentionally pessimistic and gives us a safety guarantee.
- While driving both maps from a single run-time knob is appealing, it mixes two different semantics: a global policy capacity, and a per-cgroup upper bound. This coupling can indeed lead to wasted memory when the global configuration is large, and I’m not fully convinced that trade-off is worth it yet.
- Introducing separate run-time sizing (option 2) would require us to define and carefully handle a new failure mode (partial success where the global map accepts entries but the cgroup map does not). That feels like a larger behavioural change than this PR intends.

I agree that we should eventually harden the error path around `policy_filter_cgroup_maps`, but I think that can be done independently of changing sizing semantics, and safely as a follow-up.

pkg/sensors/program/loader_linux.go
mtardy: Okay I imagine we can merge like that but this is a bit ugly that we have this `if ms, ok := spec.Maps["policy_filter_maps"];` in the middle of the generic loader code. One last ask before merging like this would be to remove the comment here and replace it with a TODO linking to https://github.com/cilium/tetragon/issues/4398.

Just CCing @olsajiri so he's not too surprised when he'll bump against this!

kyledong-suse:
@mtardy Thank you so much for taking time to review and discuss this PR. I have edit the comment as you suggested. Please take another look when you have time. Thanks!

pkg/option/flags.go
```diff
@@ -305,6 +307,8 @@ func ReadAndSetFlags() error {
 	Config.ExecveMapSize = viper.GetString(KeyExecveMapSize)
 
 	Config.RetprobesCacheSize = viper.GetInt(KeyRetprobesCacheSize)
+
+	Config.PolicyFilterMapEntries = viper.GetInt(KeyPolicyFilterMapEntries)
```
kkourt, line 311: Given the discussion in https://github.com/cilium/tetragon/pull/4331/changes/9dc973541bfee666d380d6c73e1e0be775971d32#r2629371514, should we reject values that are larger than larger than POLICY_FILTER_MAX_POLICIES?

kyledong-suse: I don’t think we should reject values larger than POLICY_FILTER_MAX_POLICIES. The goal of this PR is to allow users to configure the number of entries in the policy filter maps. The original limit of 128 is sometimes too small, and users may legitimately have more policies in their maps.

mtardy:
> please ping back here beginning of Jan if no progress

I sent a message to Jiri to progress on this

pkg/sensors/program/loader_linux.go
```diff
@@ -961,6 +962,11 @@ func doLoadProgram(
 		}
 	}
 
+	// TODO: remove this special case handling (see #4398)
```
olsajiri, line 965: this is bad.. would it help moving this all under program.Map ? there's draft PR for that already https://github.com/cilium/tetragon/pull/4501 @sayboras

mtardy: Yeah, it was discussed here for info https://github.com/cilium/tetragon/pull/4331#discussion_r2547138530

pkg/option/flags.go
kkourt: OK, but given that we have:

```
// This map keeps exactly the same information as policy_filter_maps
// but keeps the reverse mappings. i.e.
// policy_filter_maps maps policy_id to cgroup_ids
// policy_filter_cgroup_maps maps cgroup_id to policy_ids
struct {
	__uint(type, BPF_MAP_TYPE_HASH_OF_MAPS);
	__uint(max_entries, POLICY_FILTER_MAX_CGROUP_IDS);
	__type(key, __u64); /* cgroup id */
	__array(
		values, struct {
			__uint(max_entries, POLICY_FILTER_MAX_POLICIES);
			__type(key, __u32); /* policy id */
			__type(value, __u8); /* empty  */
		});
} policy_filter_cgroup_maps SEC(".maps");
```

Doesn't this mean that for a value larger than POLICY_FILTER_MAX_POLICIES, things will not work as expected when considering the `policy_filter_cgoup_maps`?

pkg/sensors/program/loader_linux.go
sayboras: oh right, I was waiting for this to land in first, just to avoid any potential conflict here.

kkourt:
Drafting the PR. @kyledong-suse feel free to unfraft when ready.

Issue #4249, opened 2025-10-25
Optimize inner policy maps with BPF_F_NO_PREALLOC to reduce memory usage


kyledong-suse:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

As mentioned part of #4191, while testing Tetragon’s policy filter implementation (`pkg/policyfilter/map.go`), I noticed that the inner per-policy maps (`policy_%d_map`) currently use a fixed size(`32768`) and the default preallocated hash map mode:
```
// addPolicyMap adds and initializes a new policy map
func (m PfMap) newPolicyMap(polID PolicyID, cgIDs []CgroupID) (polMap, error) {
	name := fmt.Sprintf("policy_%d_map", polID)
	innerSpec := &ebpf.MapSpec{
		Name:       name,
		Type:       ebpf.Hash,
		KeySize:    uint32(unsafe.Sizeof(CgroupID(0))),
		ValueSize:  uint32(1),
		MaxEntries: uint32(polMapSize),   // currently const = 32768
	}
        ...
}
```
This causes significant memory preallocation even when the number of tracked cgroups per policy is small.

I tested using the following tracing policy:
```
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "policy-1"
spec:
  podSelector:
    matchLabels:
      app: "ubuntu"
  kprobes:
  - call: "security_bprm_creds_for_exec"
    syscall: false
    args:
    - index: 0
      type: "linux_binprm"
    selectors:
    - matchArgs:
      - index: 0
        operator: "NotEqual"
        values:
        - "/usr/bin/sleep"
        - "/usr/bin/cat"
        - "/usr/bin/my-server-1"
      matchActions:
      - action: Override
        argError: -1
  options:
  - name: disable-kprobe-multi
    value: "1"
```

After applying the policy, the resulting map allocation was:
```
278096: hash  name policy_1_map  flags 0x0
	key 8B  value 1B  max_entries 32768  memlock 2622752B
	pids tetragon(1208075)
```
This corresponds to **~2.6 MB** of memory preallocated for a single inner map, which is excessive given that only a handful of cgroups are typically tracked per policy.

### Describe the feature you would like

We want to optimize the pre-allocation memory only for actual entries, reducing the footprint.

### Describe your proposed solution

[...]

PR #4340, opened 2025-11-13
tetragon/pkg: add user-configurable BPF_F_NO_PREALLOC flag support

kyledong-suse:
### Description

This commit adds support for controlling the BPF_F_NO_PREALLOC flag on BPF maps through CLI flags. Users can now disable preallocation globally or for specific maps to reduce memory usage.

Fixes: #4249

### DIFF
--- a/pkg/sensors/program/map.go
+++ b/pkg/sensors/program/map.go
@@ -166,7 +166,18 @@ func mapBuilder(name string, ty MapType, owner bool, lds ...*Program) *Map {
 	if len(lds) != 0 {
 		prog = lds[0]
 	}
-	m := &Map{name, "", prog, Idle(), nil, MaxEntries{0, false}, MaxEntries{0, false}, ty, owner, false}
+	m := &Map{
+		Name:         name,
+		PinPath:      "",
+		Prog:         prog,
+		PinState:     Idle(),
+		MapHandle:    nil,
+		Entries:      MaxEntries{0, false},
+		InnerEntries: MaxEntries{0, false},
+		Type:         ty,
+		Owner:        owner,
+		NoPrealloc:   false,
+	}
 	for _, ld := range lds {
 		ld.PinMap[name] = m
 	}
@@ -198,7 +209,18 @@ func MapBuilderOpts(name string, opts MapOpts, lds ...*Program) *Map {
 }
 
 func mapUser(name string, ty MapType, prog *Program) *Map {
-	return &Map{name, "", prog, Idle(), nil, MaxEntries{0, false}, MaxEntries{0, false}, ty, false, false}
+	return &Map{
+		Name:         name,
+		PinPath:      "",
+		Prog:         prog,
+		PinState:     Idle(),
+		MapHandle:    nil,
+		Entries:      MaxEntries{0, false},
+		InnerEntries: MaxEntries{0, false},
+		Type:         ty,
+		Owner:        false,
+		NoPrealloc:   false,
+	}
 }
 
 func MapUser(name string, prog *Program) *Map {


mtardy (commented):
Looks fine to me but let's see if that would be better to have a global flag instead of a per map flag for NO_PREALLOC disable since you mentioned another map in the issue. I'm really not sure which one is best 🤔 

We could even extend this flag to the map we already have using no_prealloc, but maybe that's too much for this patch set and out of scope!

kyledong-suse:
Thanks for taking time to review this PR @mtardy!
Let's wait @kkourt for his comment about this [idea](https://github.com/cilium/tetragon/issues/4249#issuecomment-3559821386).
If you all agree to have a generic flag to enable/disable `BPF_F_NO_PREALLOC`, I can modify the title of the issue as well as this PR and commit messages.

mtardy:
is it ready for re-review @kyledong-suse ? :)

kyledong-suse:
@mtardy @kkourt I have modified the implementation regarding to the [discussion](https://github.com/cilium/tetragon/issues/4249#issuecomment-3569948894) in #4249 
It's ready for re-review. Thanks!

pkg/sensors/program/loader_linux.go
```diff
@@ -974,6 +974,14 @@ func doLoadProgram(
...
 			ms.InnerMap.MaxEntries = innerMax
 		}
+
+		// Apply BPF_F_NO_PREALLOC flag if configured
+		if m.GetNoPrealloc() {
+			ms.Flags = bpf.BPF_F_NO_PREALLOC
+		} else {
+			// Explicitly clear BPF_F_NO_PREALLOC if not configured
+			ms.Flags &^= bpf.BPF_F_NO_PREALLOC
+		}
```
olsajiri: so this is on many places now.. could we add something like: 

```
func (m *Map) GetPreallocFlags(flags uint32) uint32 {
  if m.NoPrealloc {
   flags |= ..  
  } else {
   flags &^= ..
  }
  return flags
}
```

and call it like.. `ms.Flags = m.GetPreallocFlags(ms.Flags)`

ot something similar to get rid of the duplicity

pkg/sensors/program/map.go
```diff
@@ -164,7 +165,18 @@ func mapBuilder(name string, ty MapType, owner bool, lds ...*Program) *Map {
...
+		PinPath:      "",
+		Prog:         prog,
+		PinState:     Idle(),
+		MapHandle:    nil,
+		Entries:      MaxEntries{0, false},
+		InnerEntries: MaxEntries{0, false},
+		Type:         ty,
+		Owner:        owner,
+		NoPrealloc:   false,
+	}
```
olsajiri, line 180: could you plz put this hunk and the one below in separate commit?

pkg/option/flags.go
```diff
@@ -237,6 +245,24 @@ func ReadAndSetFlags() error {
...
+	var bpfMapsDisablePrealloc []string
+	if err = viper.UnmarshalKey(KeyBpfMapsDisablePrealloc, &bpfMapsDisablePrealloc, viper.DecodeHook(stringToSliceHookFunc(","))); err != nil {
+		return fmt.Errorf("failed to parse bpf-maps-disable-prealloc value: %w", err)
+	}
+	Config.BpfMapsDisablePrealloc = bpfMapsDisablePrealloc
+
+	// Warn if both flags are set and conflict
+	if !Config.BpfMapsPrealloc && len(Config.BpfMapsDisablePrealloc) > 0 {
+		logger.GetLogger().Warn("--bpf-maps-prealloc=false and --bpf-maps-disable-prealloc are both set. The disable-prealloc list will be ignored since preallocation is already disabled for all supported maps.")
+	}
```
olsajiri, line 273: hum, what does it mean supported maps? looks like I can run:
`
./tetragon --bpf-lib ./bpf/objs/ --bpf-maps-disable-prealloc="execve_map"`

pkg/option/flags.go
```diff
@@ -137,6 +139,12 @@ const (
 	KeyRetprobesCacheSize = "retprobes-cache-size"
 )
 
+// SupportedBpfMapsForNoPrealloc lists the BPF maps that support BPF_F_NO_PREALLOC flag.
+var SupportedBpfMapsForNoPrealloc = []string{
+	"policy_filter_maps", // Applies to inner maps (policy_%d_map)
+	"override_tasks",
+}
```
olsajiri: this seems to be used only to generate string to option string ?

pkg/sensors/load_linux.go
```diff
@@ -80,6 +81,14 @@ func (s *Sensor) loadMap(bpfDir string, loaderCache *loaderCache, m *program.Map
...
+		// Apply BPF_F_NO_PREALLOC flag if configured
+		if m.GetNoPrealloc() {
+			mapSpec.Flags |= bpf.BPF_F_NO_PREALLOC
+		} else {
+			// Explicitly clear BPF_F_NO_PREALLOC if not configured
+			mapSpec.Flags &^= bpf.BPF_F_NO_PREALLOC
+		}
 	} else {
 		// If map is NOT the owner we follow the maximum entries
 		// of the pinned map and update the spec with that.
```
olsajiri, line 88: there is a l.Debug output below:

`l.Debug("tetragon, map loaded.", "sensor", s.Name, "map", m.Name, "path", pinPath, "max", m.Entries)
`

 could you plz add noprealloc output to it?

olsajiri (changes requested):
heya, looks good, left some comments

I need to check the issue properly, but did we discuss performance implications? I guess `policy_filter_maps` is fine, I'm not sure `override_task` is that straight forward

thanks

pkg/sensors/program/loader_linux.go
kyledong-suse: I'll add a new helper for program.Map
```
func (m *Map) GetPreallocFlags(flags uint32) uint32 {
	if m.GetNoPrealloc() {
		return flags | bpf.BPF_F_NO_PREALLOC
	}
	return flags &^ bpf.BPF_F_NO_PREALLOC
}
```
However, for policyfilter/map.go, since it's using raw ebpf.Map, so this helper won't affect them for now. 
Please refer to [this discussion](https://github.com/cilium/tetragon/pull/4331#discussion_r2577385729) for the potential refactoring for policyfilter/map.go. I'll suggest to create a separate issue to work on it.

pkg/option/flags.go
kyledong-suse: I was thinking to provide more details to user in CLI, but seems we don't need this string list as it's already mentioned in the document. I can remove it.

pkg/option/flags.go
kyledong-suse: Currently, only `policy_filter_maps` and `override_tasks` will call `ShouldUseNoPrealloc()`.
Basically user can add any map names, but they just won't be affected which I think it is a bit confused to user. So that's the main reason that I add "supported maps" in the log and document.

pkg/sensors/program/map.go
kyledong-suse: Absolutely, I just want to confirm that you want me to put changes for line 168-179 and line 211-222 in a separate commit?

kyledong-suse:
>I need to check the issue properly, but did we discuss performance implications? I guess policy_filter_maps is fine, I'm not sure override_task is that straight forward

It has been discussed a bit in the issue. Additionally, #4204 proposed an idea to use shared `override_task` map, which will optimize both memory utilization and CPU consumption.

pkg/sensors/program/loader_linux.go
kyledong-suse: Issue https://github.com/cilium/tetragon/issues/4398 is created to track the refactoring

olsajiri:
@kyledong-suse I think it's ok, I have one more idea for the code, please check https://github.com/cilium/tetragon/pull/4463/changes/4705b872db8cf9e147221832cd8657b57c330378

- let's mark program.Map object with `SetNoPreAlloc` when we allow to use `BPF_F_NO_PREALLOC`, like:
  `overrideTasksMap := program.MapBuilderProgram("override_tasks", load).SetNoPreAlloc()`
- perhaps `GetPreallocFlags` could have generic name, maybe just `Flags` ?
- I think we should change `pkg/policyfilter` to use program.Map object, so we do not have separate  `BPF_F_NO_PREALLOC` code in there, but that can be follow up, meanwhile just move the `BPF_F_NO_PREALLOC` setup in single function please check https://github.com/cilium/tetragon/pull/4463/changes/4705b872db8cf9e147221832cd8657b57c330378


just to make sure I understand the options semantics.. we either enable no-prealloc for all allowed maps (via SetNoPreAlloc) with `--bpf-maps-prealloc=false` or specificaly enable separated maps with `--bpf-maps-disable-prealloc=map1,map2..`

kkourt:
The PR does not seem to have been update in a while, so I'm moving it to draft.
@kyledong-suse please move it out of draft when it is ready for reviewing.

PR #4356, opened 2025-11-17
tetragon/windows: Upgrade CI to use efw 1.0.0-rc1

ExceptionalHandler:
This PR upgrades the ebpf-for-windows version used to build ebpf programs to v1.0.0.-rc1 which is closer ro what will be mabe available as part of Windows OS

### DIFF
--- a/.github/workflows/windows-build-smoke-test.yml
+++ b/.github/workflows/windows-build-smoke-test.yml
@@ -80,7 +80,7 @@ jobs:
         run: | 
           git clone --recursive https://github.com/microsoft/ntosebpfext.git
           cd ${{ env.TEMP }}\ntosebpfext
-          git checkout e7dc209a8be0da2ff5d75f5772a0ee0bf4a10383
+          git checkout b4b240c99321f43b073d792a2f8c194e46a1b6ce
       
       - name: Copy Process_monitor.c file
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/windows-build-smoke-test.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         run: | 
           git clone --recursive https://github.com/microsoft/ntosebpfext.git
           cd ${{ env.TEMP }}\ntosebpfext
-          git checkout e7dc209a8be0da2ff5d75f5772a0ee0bf4a10383
+          git checkout tags/v0.6.0
```
FedeDP: Can we retain the usage of the commit hash instead?
Should be b4b240c99321f43b073d792a2f8c194e46a1b6ce

ExceptionalHandler: I ask for my education: What is the advantage of using commit hashes instead of tags ?

FedeDP: The reason is pretty simple: tags can be re-pushed, hashes don't (ie: they are immutable). Less risk of supply chain attacks.

ExceptionalHandler: I have addressed this.

FedeDP: Thank you!

ygvalent: @ExceptionalHandler I had the same question some time ago, here's a good thread about it: https://isovalent.slack.com/archives/C01742ANV9P/p1755200254977689

Issue #4220, opened 2025-10-21
tests: `pkg.bugtool.TestFindMaps` is flaky


mtardy:
This one's on me 😔 

https://github.com/cilium/tetragon/actions/runs/18681499471/job/53264252392?pr=4216

FedeDP:
For reference, this is the test failure log:
```
level=info msg="BTF discovery: default kernel btf file found" btf-file=/sys/kernel/btf/vmlinux
=== RUN   TestFindMaps
=== RUN   TestFindMaps/NoSuchFile
=== RUN   TestFindMaps/BaseSensorMemlock
level=info msg="Exit probe on acct_process"
level=info msg="Set execve_map entries 32768" size=28M
level=info msg="BPF ring buffer size (bytes)" total=128K
level=info msg="BTF file: using metadata file" metadata=/sys/kernel/btf/vmlinux
level=info msg="Loading sensor" name=__base__
level=info msg="Loading kernel version 6.1.155"
level=info msg="Loaded sensor successfully" sensor=__base__
level=info msg="Read ProcFS /proc appended 71/127 entries"
    maps_test.go:48: 
        	Error Trace:	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/bugtool/maps_test.go:48
        	Error:      	Should not be zero, but was 0
        	Test:       	TestFindMaps/BaseSensorMemlock
level=info msg="Unloading sensor __base__"
level=info msg="Sensor unloaded" sensor=__base__ maps-error=[]
    base.go:198: cleanup: unloading base sensor
level=info msg="Unloading sensor __base__"
--- FAIL: TestFindMaps (0.62s)
    --- PASS: TestFindMaps/NoSuchFile (0.00s)
    --- FAIL: TestFindMaps/BaseSensorMemlock (0.62s)
FAIL
```

That means the failure is here:
```go
mapsUsedByProgs, err := FindMapsUsedByPinnedProgs(path)
require.NoError(t, err)
if assert.NotEmpty(t, mapsUsedByProgs) {
	assert.NotZero(t, mapsUsedByProgs[0].Memlock)
}
```

PR #4362, opened 2025-11-20
cmd/tetra:  add explain command for standalone CRD documentation

### LINKED ISSUE #2073: New command in Tetragon CLI: tetra explain

Currently the best exact description of Tetragon's CRD TraingPolicy resides in the [source code](https://github.com/cilium/tetragon/tree/main/pkg/k8s/apis/cilium.io/v1alpha1). This is suboptimal; the proposed command `tetra explain` would provide easier access to the same description.

See also related thread in Slack [here](https://cilium.slack.com/archives/C03EV7KJPJ9/p1707300049141239).

When working with tracing policies, and exact understanding of the fields in Tetragon's CRD TracingPolicy and their semantics is mandatory. The Kubernetes client `kubectl` offers a command kubectl explain to get hold or CRD descriptions. For cases where Tetragon serves independently of Kubernetes it would be helpful if its client `tetra` could support such a command as well: e.g. `tetra explain tracingpolicy.spec.call.syscall`.

`tetra explain` should render the description from comments that already reside in the source code. An implementation could perhaps build on top of (call out to code "inside") Kubernetes' `kubectl explain`. Yet `tetra explain` must not introduce a strong dependency on Kubernetes: the command should work in deployments both with and without.

### Code of Conduct

AritraDey-Dev:
Fixes #2073

### Description
See commit.

### Changelog

```release-note
new(cmd/tetra): add explain command for standalone CRD documentation
```

### DIFF
--- a/cmd/tetra/explain/explain_test.go
+++ b/cmd/tetra/explain/explain_test.go
@@ -1,0 +1,133 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package explain
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestConvertBracketsToDots(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
+			name:     "simple bracket",
+			input:    "tracingpolicy[spec]",
+			expected: "tracingpolicy.spec",
+		},
+		{
+			name:     "nested brackets",
+			input:    "tracingpolicy[spec][kprobes]",
+			expected: "tracingpolicy.spec.kprobes",
+		},
+		{
+			name:     "mixed notation",
+			input:    "tracingpolicy.spec[kprobes]",
+			expected: "tracingpolicy.spec.kprobes",
+		},
+		{
+			name:     "empty bracket",
+			input:    "tracingpolicy[]",
+			expected: "tracingpolicy",
+		},
+		{
+			name:     "numeric index",
+			input:    "tracingpolicy.spec.kprobes[0]",
+			expected: "tracingpolicy.spec.kprobes",
+		},
+		{
+			name:     "numeric index in middle",
+			input:    "tracingpolicy.spec.kprobes[0].call",
+			expected: "tracingpolicy.spec.kprobes.call",
+		},
+		{
+			name:     "no brackets",
+			input:    "tracingpolicy.spec.kprobes",
+			expected: "tracingpolicy.spec.kprobes",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := convertBracketsToDots(tt.input)
+			if got != tt.expected {
+				t.Errorf("convertBracketsToDots(%q) = %q, want %q", tt.input, got, tt.expected)
+			}
+		})
+	}
+}
+
+func TestNormalizePath(t *testing.T) {
+	tests := []struct {
+		name    string
+		input   string
+		want    []string
+		wantErr bool
+	}{
+		{
+			name:    "simple path",
+			input:   "tracingpolicy.spec",
+			want:    []string{"tracingpolicy", "spec"},
+			wantErr: false,
+		},
+		{
+			name:    "bracket notation",
+			input:   "tracingpolicy[spec][kprobes]",
+			want:    []string{"tracingpolicy", "spec", "kprobes"},
+			wantErr: false,
+		},
+		{
+			name:    "mixed notation",
+			input:   "tracingpolicy.spec[kprobes]",
+			want:    []string{"tracingpolicy", "spec", "kprobes"},
+			wantErr: false,
+		},
+		{
+			name:    "empty path",
+			input:   "",
+			want:    nil,
+			wantErr: true,
+		},
+		{
+			name:    "starts with dot",
+			input:   ".tracingpolicy.spec",
+			want:    nil,
+			wantErr: true,
+		},
+		{
+			name:    "ends with dot",
+			input:   "tracingpolicy.spec.",
+			want:    nil,
+			wantErr: true,
+		},
+		{
+			name:    "empty segment",
+			input:   "tracingpolicy..spec",
+			want:    []string{"tracingpolicy", "spec"},
+			wantErr: false,
+		},
+		{
+			name:    "whitespace",
+			input:   " tracingpolicy . spec ",
+			want:    []string{"tracingpolicy", "spec"},
+			wantErr: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := normalizePath(tt.input)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("normalizePath() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("normalizePath() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}


cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,603 @@
...
+		for resName, info := range supportedResources {
+			supported = append(supported, resName)
+			supported = append(supported, info.aliases...)
+		}
+		sort.Strings(supported)
+		return fmt.Errorf("unsupported resource: %s (supported resources: %s)", resourceName, strings.Join(supported, ", "))
+	}
+
+	// Try to find and load CRD file
+	crdData, err := findCRDFile(resourceInfo.crdFiles)
```
FedeDP: This will never work outside of development build (ie: when you build tetragon from within its repo root).

FedeDP: You can probably directly leverage `client.TracingPolicyCRD` (see github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/client), and its generated json schema:
```go
version := client.TracingPolicyCRD.Definition.Spec.Versions[0] // latest version
if version.Schema != nil && version.Schema.OpenAPIV3Schema != nil {
	jsonSchema := version.Schema.OpenAPIV3Schema
	fmt.Printf("JSON Schema for CRD '%s', Version '%s':\n", client.TracingPolicyCRD.Definition.Name, version.Name)

	schemaBytes, _ := json.MarshalIndent(jsonSchema, "", "  ")
	fmt.Println(string(schemaBytes))
}
```
By using `jsonSchema` you have access to whatever fields + description you need.

FedeDP: At that point, `handleFallback` can go away.

FedeDP (changes requested):
Left a suggestion to improve the implementation.

cmd/tetra/explain/explain.go
AritraDey-Dev: Thanks for the suggestion. With this approach, there's also no need to use the hard-coded paths i have used.

FedeDP:
Can you squash the commits into a single one?

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,582 @@
...
+	cmd.Flags().StringVarP(&outputFormat, "output", "o", "", "Output format. One of: json|yaml")
+	cmd.Flags().BoolVarP(&recursive, "recursive", "R", false, "Print the fields of fields recursively")
+	cmd.Flags().BoolVar(&showExample, "example", false, "Show example usage")
+	cmd.Flags().BoolVar(&listMode, "list", false, "List all supported resources")
+	cmd.Flags().StringVar(&apiVersion, "api-version", "", "Specify the API version to use")
+
+	return cmd
+}
+
+type crdDef struct {
```
FedeDP: Do we really need this type? Can't we just use `CustomResourceDefinition` ?

AritraDey-Dev: yes not needed.Can use the existing one.

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,563 @@
...
+	cmd.Flags().StringVarP(&outputFormat, "output", "o", "", "Output format. One of: json|yaml")
+	cmd.Flags().BoolVarP(&recursive, "recursive", "R", false, "Print the fields of fields recursively")
+	cmd.Flags().BoolVar(&showExample, "example", false, "Show example usage")
+	cmd.Flags().BoolVar(&listMode, "list", false, "List all supported resources")
+	cmd.Flags().StringVar(&apiVersion, "api-version", "", "Specify the API version to use")
+
+	return cmd
+}
+
+type schema struct {
```
FedeDP: I think we can also drop this `schema` usage, and use `*apiextensionsv1.JSONSchemaProps`.

I'll attach a small patch for that, let me know if you are ok with it.
[explain.patch](https://github.com/user-attachments/files/23674114/explain.patch)

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,563 @@
...
+	Items                *schema            `json:"items"`
+	AdditionalProperties *schema            `json:"additionalProperties"`
+	Required             []string           `json:"required"`
+	Example              any                `json:"example"`
+	Enum                 []any              `json:"enum"`
+	Format               string             `json:"format"`
+	Default              any                `json:"default"`
+}
+
+var supportedResources = map[string]struct {
```
FedeDP: Since this unnamed struct is also used below as:
```
var resourceInfo struct {
		kind    string
		aliases []string
		getCRD  func() *apiextensionsv1.CustomResourceDefinition
	}
```
can we define it only once and use it elsewhere?

AritraDey-Dev: Nice catch.Completely agree.

AritraDey-Dev: Implemented in next commit.

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,485 @@
...
+		part = strings.TrimSpace(part)
+		if part == "" {
+			if i == 0 || i == len(parts)-1 {
+				return nil, errors.New("invalid path: cannot start or end with '.'")
+			}
+			continue // skip empty segments in middle
+		}
+
+		// Handle array notation like field[] or field[0]
+		if strings.Contains(part, "[") && strings.Contains(part, "]") {
```
FedeDP: Question about this one: what are we supporting by doing so (considering we are leaving out the string between "[" and "]")?

Should we expand this a little bit to support eg:
```
./tetra explain tracingpolicy[spec][uprobes]
```

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,485 @@
...
+	parts, err := normalizePath(path)
+	if err != nil {
+		return err
+	}
+
+	resourceName := strings.ToLower(parts[0])
+
+	// Find the resource by name or alias
+	var resInfo resourceInfo
+	var found bool
```
FedeDP:
```suggestion
	var resInfo *resourceInfo
```
And below set it `resInfo = &info`, and finally check:
`if resInfo == nil { ... }` instead of `if !found`.
Basically, let's kill `var found bool`.

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,485 @@
...
+		version = &crd.Spec.Versions[len(crd.Spec.Versions)-1]
+	}
+
+	if version.Schema == nil || version.Schema.OpenAPIV3Schema == nil {
+		return errors.New("no schema found for CRD version")
+	}
+
+	rootSchema := version.Schema.OpenAPIV3Schema
+
+	current := rootSchema
```
FedeDP:
```suggestion
	current := version.Schema.OpenAPIV3Schema
```
`rootSchema` is unused.

FedeDP:
This is quite ready; left some comments :) 
Btw let me take one moment and thank you for this! Not only does this close a requested feature from long ago, but i think it is really useful! :rocket:

cmd/tetra/explain/explain.go
AritraDey-Dev: Yes, this is a limitation and needs to be expanded. It requires a separate function to handle it. I will implement this in the next commit.

AritraDey-Dev: Implemented!
```
./tetra explain tracingpolicy[spec][uprobes]
KIND:     TracingPolicy
VERSION:  v1alpha1

FIELD:    uprobes <[]object>

DESCRIPTION:
     A list of uprobe specs.

FIELDS:
   addrs        <[]integer>
     List of the traced addresses

   args <[]object>
     A list of function arguments to include in the trace output.

   btfPath      <string>
     path for a BTF file for the traced binary

   data <[]object>
     A list of data to include in the trace output.

   message      <string>
     A short message of 256 characters max that will be included
in the event output to inform users what is going on.

   offsets      <[]integer>
     List of the traced offsets

   path <string> -required-
     Name of the traced binary

   refCtrOffsets        <[]integer>
     List of the traced ref_ctr_offsets

   return       <boolean>
     Indicates whether to collect return value of the traced function.

   returnArg    <object>
     A return argument to include in the trace output.

   selectors    <[]object>
     Selectors to apply before producing trace output

   symbols      <[]string>
     List of the traced symbols

   tags <[]string>
     Tags to categorize the event, will be include in the event output.
Maximum of 16 Tags are supported.
```

FedeDP: :rocket:

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,519 @@
...
+	if s == "" {
+		return false
+	}
+	for _, r := range s {
+		if r < '0' || r > '9' {
+			return false
+		}
+	}
+	return true
+}
```
FedeDP, line 126:
```suggestion
func isNumeric(s string) bool {
	_, err := strconv.Atoi(s)
	return err == nil
}
```

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,519 @@
...
+		info := supportedResources[name]
+		shortNames := strings.Join(info.aliases, ",")
+		fmt.Fprintf(tw, "%s\t%s\t%s\n", name, shortNames, info.kind)
+	}
+	tw.Flush()
+}
+
+// convertBracketsToDots converts bracket notation to dot notation using regex
+func convertBracketsToDots(path string) string {
+	path = strings.ReplaceAll(path, "][", ".")
```
FedeDP, line 100: Since after this we end up with, eg: `tracingpolicy[spec][uprobes][tags]` -> `tracingpolicy[spec.uprobes.tags]`, is it really worth it all the regex-related logic? After all, we can just safely replace `[` with `.` and drop last `]`, right?
Or is there an edge case i am not considering?

AritraDey-Dev: Yes `ReplaceAll` would work for most cases! I kept the regex mainly for input validation (handling malformed input like field(something like unclosed brackets)).Also something like field[0] → field.0 is harmless.

cmd/tetra/explain/explain.go
AritraDey-Dev: This will require another extra import.

FedeDP: Not a big deal since it's from stdlib :)

FedeDP (commented):
Wdyt if we drop `supportedResources` in favor of using `client.AllCRDs`?
I got a patch that should work fine. At that point we basically have no custom code in `explain.go` :)
[explain.patch](https://github.com/user-attachments/files/23709583/explain.patch)


We also use aliases (ie: short names) from the crds.

Example:
```
$ ./tetra explain --list
NAME                     SHORTNAMES
TracingPolicy            tgtp
TracingPolicyNamespaced  tgtpn
PodInfo                  tgpi
```

And
```
$ ./tetra explain tracingpolicies
Error: unsupported resource: tracingpolicies (supported resources: PodInfo, TracingPolicy, TracingPolicyNamespaced, tgpi, tgtp, tgtpn)
```

And:
```
$ ./tetra explain tracingpolicy
KIND:     TracingPolicy
VERSION:  v1alpha1

DESCRIPTION:


FIELDS:
   apiVersion	<string>
     APIVersion defines the versioned schema of this representation of an object.

   kind	<string>
     Kind is a string value representing the REST resource this object represents.

   metadata	<object> -required-

   spec	<object> -required-
     Tracing policy specification.
```

AritraDey-Dev:
Got it — using` client.AllCRDs `makes sense and removes the need for maintaining `supportedResources`.
I’ll update the PR accordingly.

FedeDP:
Thanks! And sorry for the scattered review :) I am going through it piece by piece!
After this we will be finally over! Thanks for your patience!

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,452 @@
...
+		return outputJSON(w, current)
+	case "yaml":
+		return outputYAML(w, current)
+	default:
+		return outputDefault(w, crd.Spec.Names.Kind, version.Name, parts, current)
+	}
+}
+
+func outputJSON(w io.Writer, s *apiextensionsv1.JSONSchemaProps) error {
+	data, err := json.MarshalIndent(s, "", "  ")
```
mtardy: nit: might be better to just use `json.Marshal` and let the user reindent if he wants with tools like `jq`. I would assume that if you need JSON output, you don't really care about beautifying the output here.

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,452 @@
...
+
+		normalized = append(normalized, part)
+	}
+
+	if len(normalized) == 0 {
+		return nil, errors.New("invalid path: no valid segments found")
+	}
+
+	return normalized, nil
+}
```
mtardy, line 157: The code is extremely critical so it's not really required but could have been nice to add a few units for those parsing functions in general

cmd/tetra/explain/explain.go
```diff
@@ -0,0 +1,452 @@
...
+
+func formatDescription(desc string) string {
+	if desc == "" {
+		return ""
+	}
+
+	// Return the full description for proper multi-line handling
+	return desc
+}
```
mtardy: I'm not sure to understand why this function is needed, to my understanding it does nothing?

cmd/tetra/commands_darwin.go
```diff
@@ -14,4 +15,5 @@ func addCommands(rootCmd *cobra.Command) {
 	addBaseCommands(rootCmd)
 	rootCmd.AddCommand(tracingpolicy.New())
 	rootCmd.AddCommand(loglevel.New())
+	rootCmd.AddCommand(explain.New())
```
mtardy: why can't we add this command to `commands.go` to support Windows as well? Is there a blocker?

cmd/tetra/explain/explain.go
AritraDey-Dev: Yes, it seems to be a leftover from earlier, where I probably intended to add some specific formatting logic. Since the `indent` function already handles multi-line strings correctly, this helper is redundant. I’ll remove it.

cmd/tetra/explain/explain.go
AritraDey-Dev: Yes, definitely. I was planning to add tests in a separate PR once this is merged, but if it’s not possible to proceed without tests, I can include them in this PR as well. Please let me know.

cmd/tetra/commands_darwin.go
AritraDey-Dev: I'm on Linux and can't verify Windows support, so I played it safe. If you can test it on Windows, can move it to commands.go

cmd/tetra/explain/explain.go
mtardy: oups I wanted to write:
> The code is **not** extremely critical

😅 

> I was planning to add tests in a separate PR once this is merged

I think including them with the new code would make sense in the same PR. No need for anything crazy, just a few test cases to make sure the examples work.. :)

cmd/tetra/commands_darwin.go
mtardy: I don't see why it wouldn't work on Windows? Let's include it!

cmd/tetra/explain/explain.go
AritraDey-Dev: Added in https://github.com/cilium/tetragon/pull/4362/commits/0540480aac6a93d72f3b0dbfad115488f50082df Does this need to be squashed, or can we keep it as a separate commit?

AritraDey-Dev:
https://github.com/cilium/tetragon/actions/runs/19829777729/job/56813511234?pr=4362
> ❌	pkg.bugtool.TestFindMaps													(total:3 failed:2 skipped:0)	1.170776882s	6.350378272s
├─❌	TestFindMaps/BaseSensorMemlock													820ms				(820ms)
└─❌	TestFindMaps		

ig  #4220

PR #4376, opened 2025-11-24
tetragon: assorted fixes

olsajiri:
adding fixes/cleanup for loader events and register read/write

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -13,6 +13,7 @@
 #include "policy_stats.h"
 #include "generic_path.h"
 #include "bpf_ktime.h"
+#include "regs.h"
 
 #define MAX_TOTAL 9000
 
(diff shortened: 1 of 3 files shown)


pkg/sensors/tracing/genericuprobe.go
```diff
@@ -406,23 +411,17 @@ func createGenericUprobeSensor(
...
 		}
-
-		hasRegsOverrideAction = hasRegsOverrideAction || selectors.HasOverride(uprobe.Selectors)
 	}
 
-	hasSleepableOffload := hasRegsOverrideAction && bpf.HasUprobeRegsChange()
-
 	if in.useMulti {
-		progs, maps, err = createMultiUprobeSensor(name, ids, polInfo.name, hasSleepableOffload)
+		progs, maps, err = createMultiUprobeSensor(name, ids, polInfo.name, &has)
```
FedeDP: I'd pass `has` by value here; while it might be a little worse in term of performance, it clearly states that we are not going to play with the pointed data (ie: that `has` as seen from the caller is const). WDYT?

FedeDP: Same below.

FedeDP (approved):

tpapagian (approved):

Issue #2265, opened 2024-03-28
doc: document why Tetragon will fail to load its BPF programs with operation not permitted


tixxdz:
We should add in our https://tetragon.io/docs/faq/ some documentation why Tetragon will fail with operation not permitted when loading BPF programs on some machines.

Users have been reporting this https://github.com/cilium/tetragon/issues/2263 and it is due to lockdown in confidentiality mode, or could be to some other LSM context.

The PR https://github.com/cilium/tetragon/pull/2149  prints the security context of Tetragon at startup, so having an example of output in the FAQ will help.

We can use @kkourt text:
"lockdown mode seems to be set to 'confidentiality', which does not allow Tetragon to operate because it prohibits loading BPF programs. See https://github.com/iovisor/bcc/issues/2565 for more details".

dangome3:
I will be in charge of this ticket

mtardy:
> I will be in charge of this ticket

Sure! This mentions lockdown mode on which we could improve the detection and print more debug messages (always better to detect and log than writing in the FAQ if we can, or at least with a probe we can have action item on how to detect at runtime in the FAQ). I imagine it can extend to other security settings that could affect BPF. Could be an interesting investigation :)

dangome3:
Implementation in https://github.com/cilium/tetragon/pull/4381

dangome3:
@mtardy @tixxdz the PR has been already merged. Please let me know if there's anything else I can do or if we can close this issue :)

mtardy:
Yep, fixed by https://github.com/cilium/tetragon/pull/4381

PR #4381, opened 2025-11-24
docs: add FAQ entry for kernel lockdown and improve warning message

dangome3:
Implementation for https://github.com/cilium/tetragon/issues/2265 
### Description
Adds documentation for why Tetragon will fail with operation not permitted when loading BPF programs on some machines.

### FAQ Section Rendered
Here how the documentation looks https://deploy-preview-4381--tetragon.netlify.app/docs/installation/faq/#kernel-lockdown

### Debug message
Here the info message updated when the lockdown is set to confidential:

Screenshot 2025-11-24 at 6 34 31 PM

### DIFF
--- a/docs/content/en/docs/installation/faq.md
+++ b/docs/content/en/docs/installation/faq.md
@@ -247,3 +247,31 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
     nsenter -t 1 -m -u -n -i sh -c 'ls -la /sys/kernel/btf/vmlinux'
 -r--r--r--    1 root     root       4988627 Nov 21 20:33 /sys/kernel/btf/vmlinux
 ```
+
+### Why Tetragon fails with "operation not permitted" when loading BPF programs {#kernel-lockdown}
+
+This error indicates that Linux Kernel Lockdown is preventing BPF programs from loading.
+Kernel Lockdown is a security feature designed to prevent even the root user from modifying
+the kernel or reading its internal secrets. It operates in two distinct modes:
+
+- Integrity: Prevents the user from modifying the running kernel. For example, it blocks
+  writing to `/dev/mem` or loading unsigned kernel modules.
+- Confidentiality: Includes all integrity protections but goes further by preventing the user
+  from reading information from the kernel that could reveal secrets.
+
+On systems with Kernel Lockdown enabled in `confidentiality` mode, Tetragon will fail with `EPERM`.
+To fix this, use `integrity` mode.
+
+You can verify the Lockdown status in the Tetragon startup logs.
+A warning is printed when `confidentiality` mode is active:
+
+```
+level=warn msg="Kernel Lockdown is in 'confidentiality' mode; Tetragon will fail to load BPF programs. See https://tetragon.io/docs/installation/faq/#kernel-lockdown for details."
+level=info msg="Tetragon current security context" SELinux=unconfined AppArmor=unconfined Smack="" Lockdown=confidentiality
+```
+
+You can also check the status of the Kernel Lockdown by running:
+
+```shell
+cat /sys/kernel/security/lockdown
+```
--- a/pkg/reader/proc/proc_linux.go
+++ b/pkg/reader/proc/proc_linux.go
@@ -178,7 +178,7 @@ func LogCurrentSecurityContext() {
 			logLSM = true
 		}
 		if lockdown == "confidentiality" {
-			logger.GetLogger().Warn("Kernel Lockdown is in 'confidentiality' mode, Tetragon will fail to load BPF programs")
+			logger.GetLogger().Warn("Kernel Lockdown is in 'confidentiality' mode; Tetragon will fail to load BPF programs. See https://tetragon.io/docs/installation/faq/#kernel-lockdown for details.")
 		}
 	}
 


docs/content/en/docs/installation/faq.md
```diff
@@ -247,3 +247,52 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
     nsenter -t 1 -m -u -n -i sh -c 'ls -la /sys/kernel/btf/vmlinux'
 -r--r--r--    1 root     root       4988627 Nov 21 20:33 /sys/kernel/btf/vmlinux
 ```
+
+### Why Tetragon fails with "operation not permitted" when loading BPF programs {#kernel-lockdown}
+
+Tetragon relies on eBPF (Extended Berkeley Packet Filter) to observe system behavior by attaching
+probes to the kernel and reading arguments, return values, and internal structures. Linux Kernel
+Lockdown is a security feature designed to prevent even the root user (UID 0) from modifying the
+kernel or reading its internal secrets. It effectively draws a line where "root" privileges stop,
```
tixxdz: First paragraph good, but I would remove from "It effectively..."  since in reality it doesn't, the lockdown is just a set of security checks performed where at random places where such leaks could happen. A compromised root can do way more... but eventually overwrite lockdown protection itself...

docs/content/en/docs/installation/faq.md
```diff
@@ -247,3 +247,52 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
...
+cryptographic keys stored in kernel memory.
+
+Kernel Lockdown operates in two distinct modes:
+
+- **Integrity**: Prevents the user from modifying the running kernel. For example, it blocks
+  writing to `/dev/mem` or loading unsigned kernel modules.
+- **Confidentiality**: Includes all integrity protections but goes further by preventing the user
+  from reading information from the kernel that could reveal secrets.
+
+On systems with **Kernel Lockdown** enabled in `confidentiality` mode (common with Secure Boot),
```
tixxdz: Please just re-write this as: On systems with **Kernel Lockdown** enabled in `confidentiality` mode Tetragon will fail with EPERM...

No need to further explain.

docs/content/en/docs/installation/faq.md
```diff
@@ -247,3 +247,52 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
...
+cat /sys/kernel/security/lockdown
+```
+
+The output will show the current mode:
+
+* `[none]` - Lockdown is disabled (Tetragon will work)
+* `[integrity]` - Only integrity is protected (Tetragon will work)
+* `[confidentiality]` - Full lockdown (Tetragon will **NOT** work)
+
+#### Resolution Options
```
tixxdz: Here just say run in integrity mode  lockdown=integrity  and that's it.

No need to mention the rest.

tixxdz (commented):
Much appreciated @dangome3 for the change! some comments to fix.

Then LGTM.

docs/content/en/docs/installation/faq.md
```diff
@@ -247,3 +247,52 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
...
+`EPERM (Operation not permitted)`. This causes Tetragon to fail when trying to load its BPF
+programs at startup.
+
+#### Checking Lockdown Status
+
+When Tetragon starts on a system with lockdown in `confidentiality` mode, it will log a warning similar to:
+
+```text
+level=warn msg="Kernel Lockdown is in 'confidentiality' mode; Tetragon will fail to load BPF programs. See https://tetragon.io/docs/installation/faq/#kernel-lockdown for details."
+```
```
mtardy: please remove `text` for output, the rendering is different then.
````suggestion
```
level=warn msg="Kernel Lockdown is in 'confidentiality' mode; Tetragon will fail to load BPF programs. See https://tetragon.io/docs/installation/faq/#kernel-lockdown for details."
```
````

File: docs/content/en/docs/installation/faq.md [hunk 5]
```diff
@@ -247,3 +247,52 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
     nsenter -t 1 -m -u -n -i sh -c 'ls -la /sys/kernel/btf/vmlinux'
 -r--r--r--    1 root     root       4988627 Nov 21 20:33 /sys/kernel/btf/vmlinux
 ```
+
+### Why Tetragon fails with "operation not permitted" when loading BPF programs {#kernel-lockdown}
+
+Tetragon relies on eBPF (Extended Berkeley Packet Filter) to observe system behavior by attaching
```
mtardy: I think you can be more succinct here

```suggestion
Tetragon relies on eBPF to observe system behavior by attaching
```

docs/content/en/docs/installation/faq.md
mtardy: I think no need for bold here also
```suggestion
On systems with Kernel Lockdown enabled in `confidentiality` mode (common with Secure Boot),
```

docs/content/en/docs/installation/faq.md
mtardy: and btw I thought we didn't have this yet (per my issue comment), it seems it's already done, cool

docs/content/en/docs/installation/faq.md
```diff
@@ -247,3 +247,52 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
...
+
+```shell
+cat /sys/kernel/security/lockdown
+```
+
+The output will show the current mode:
+
+* `[none]` - Lockdown is disabled (Tetragon will work)
+* `[integrity]` - Only integrity is protected (Tetragon will work)
+* `[confidentiality]` - Full lockdown (Tetragon will **NOT** work)
```
mtardy:
```suggestion
* `[confidentiality]` - Full lockdown (Tetragon will not work)
```

docs/content/en/docs/installation/faq.md
```diff
@@ -247,3 +247,52 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
...
+
+#### Resolution Options
+
+1. **Disable Secure Boot** (in UEFI/BIOS settings) - This will disable lockdown entirely.
+2. **Use `integrity` mode instead of `confidentiality`** - Add `lockdown=integrity` to kernel boot parameters.
+3. **Use a custom kernel** without `CONFIG_SECURITY_LOCKDOWN_LSM` enabled.
```
mtardy: See above comment from Djalal but just mentioning:

I'd drop the bold statement as well also just with plain simple text
```suggestion
1. Disable Secure Boot (in UEFI/BIOS settings) - This will disable lockdown entirely.
2. Use `integrity` mode instead of `confidentiality` - Add `lockdown=integrity` to kernel boot parameters.
3. Use a custom kernel without `CONFIG_SECURITY_LOCKDOWN_LSM` enabled.
```

mtardy (commented):
Thanks for doing this, indeed maybe that's a bit too verbose (e.g. you don't need to recall what eBPF is and repeat the information multiple time) but this is a nice entry thanks! And looks like the detection mechanism is already there so it's cool, thank you for looking :)

mtardy:
btw we have a bot rendering the website for you for each PR so you can simply link to it by editing your PR like this: https://deploy-preview-4381--tetragon.netlify.app/docs/installation/faq/#kernel-lockdown, no need for screenshot for the next one! :)

docs/content/en/docs/installation/faq.md
```diff
@@ -247,3 +247,30 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
...
+This error indicates that Linux Kernel Lockdown is preventing BPF programs from loading.
+Kernel Lockdown is a security feature designed to prevent even the root user from modifying
+the kernel or reading its internal secrets. It operates in two distinct modes:
+
+- Integrity: Prevents the user from modifying the running kernel. For example, it blocks
+  writing to `/dev/mem` or loading unsigned kernel modules.
+- Confidentiality: Includes all integrity protections but goes further by preventing the user
+  from reading information from the kernel that could reveal secrets.
+
+On systems with Kernel Lockdown enabled in confidentiality mode, Tetragon will fail with EPERM. To fix this, use integrity mode.
```
FedeDP: I think it would be better to highlight keyword here somehow: 
```suggestion
On systems with Kernel Lockdown enabled in *confidentiality* mode, Tetragon will fail with `EPERM`. To fix this, use integrity mode.
```
wdyt?

dangome3: I liked the idea! but I didn't use the bold as for previous comments I understood that is better to keep plain text.

Please take a look at the new changes :)

docs/content/en/docs/installation/faq.md
```diff
@@ -247,3 +247,30 @@ $ docker run -it --rm --privileged --pid=host ubuntu \
...
+This error indicates that Linux Kernel Lockdown is preventing BPF programs from loading.
+Kernel Lockdown is a security feature designed to prevent even the root user from modifying
+the kernel or reading its internal secrets. It operates in two distinct modes:
+
+- Integrity: Prevents the user from modifying the running kernel. For example, it blocks
+  writing to `/dev/mem` or loading unsigned kernel modules.
+- Confidentiality: Includes all integrity protections but goes further by preventing the user
+  from reading information from the kernel that could reveal secrets.
+
+On systems with Kernel Lockdown enabled in `confidentiality` mode, Tetragon will fail with `EPERM`. To fix this, use `integrity` mode.
```
mtardy: small nit: wrap text at around 80 chars
```suggestion
On systems with Kernel Lockdown enabled in `confidentiality` mode, Tetragon will fail with `EPERM`.
To fix this, use `integrity` mode.
```

dangome3: Thank you for your comment! That was fixed already :)

PR #4382, opened 2025-11-26
metrics: Add node_name label

### LINKED ISSUE #2214: Add node_name label to metrics
Currently metrics exposed by Tetragon don't have a label with the node name. `node` label is added by a relabeling in the [default ServiceMonitor](https://github.com/cilium/tetragon/blob/main/install/kubernetes/tetragon/templates/servicemonitor.yaml#L21-L24). This works well enough, however:
* In deployments not using the default ServiceMonitor, users have to manually configure scrape config to add such relabeling.
* In some deployments (non-k8s in particular) node name might be not available to Prometheus at the scrape time.
* `node` label on metrics might end up being different than `node_name` field in JSON events.
* It's always nice to have metrics labels matching events fields.

So I'm suggesting to add `node_name` label to all metrics exposed by Tetragon - it should be retrieved in the same name as `node_name` in events.

AritraDey-Dev:
Fixes #2214 
### Description

This PR adds `node_name` label to metrics to match events.
### Changelog

```release-note
metrics: new node_name label
```

### DIFF
--- a/pkg/metrics/eventmetrics/eventmetrics.go
+++ b/pkg/metrics/eventmetrics/eventmetrics.go
@@ -115,7 +115,7 @@ func InitEventsMetricsForDocs(registry *prometheus.Registry) {
 	InitEventsMetrics(registry)
 
 	// Initialize metrics with example labels
-	processLabels := option.CreateProcessLabels(consts.ExampleNamespace, consts.ExampleWorkload, consts.ExamplePod, consts.ExampleBinary)
+	processLabels := option.CreateProcessLabels(consts.ExampleNamespace, consts.ExampleWorkload, consts.ExamplePod, consts.ExampleBinary, consts.ExampleNodeName)
 	for ev, evString := range tetragon.EventType_name {
 		if tetragon.EventType(ev) != tetragon.EventType_UNDEF && tetragon.EventType(ev) != tetragon.EventType_TEST {
 			EventsProcessed.WithLabelValues(processLabels, evString).Add(0)
@@ -150,7 +150,7 @@ func handleOriginalEvent(originalEvent any) {
 }
 
 func handleProcessedEvent(pInfo *tracingpolicy.PolicyInfo, processedEvent any) {
-	var eventType, namespace, workload, pod, binary string
+	var eventType, namespace, workload, pod, binary, nodeName string
 	switch ev := processedEvent.(type) {
 	case *tetragon.GetEventsResponse:
 		binary, pod, workload, namespace = GetProcessInfo(filters.GetProcess(&event.Event{Event: ev}))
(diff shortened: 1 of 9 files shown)


pkg/metrics/filteredlabels.go
```diff
@@ -25,33 +25,36 @@ func (l NilLabels) Values() []string { return []string{} }
...
 
 type ProcessLabels struct {
-	Namespace string
-	Workload  string
-	Pod       string
 	Binary    string
+	Pod       string
+	Workload  string
+	Namespace string
+	NodeName  string
```
FedeDP: Can we reduce the diff by avoiding these unnecessary changes?
The sorting of ProcessLabels fields seemed to be the same sorting of `func NewProcessLabels` arguments.
Or is this a reason for the change?

AritraDey-Dev: I reordered to match `NewProcessLabels()` parameter order, but you're right,diff should be minimized, i will revert that part.

pkg/metrics/eventmetrics/eventmetrics.go
```diff
@@ -160,10 +160,13 @@ func handleProcessedEvent(pInfo *tracingpolicy.PolicyInfo, processedEvent any) {
 			logger.GetLogger().Warn("metrics: handleProcessedEvent: unhandled event", "event", processedEvent, logfields.Error, err)
 			eventType = "unhandled"
 		}
+		if ev.NodeName != "" {
+			nodeName = ev.NodeName
+		}
```
FedeDP:
```suggestion
	nodeName = ev.NodeName
```
After all the default value is already empty string.

AritraDey-Dev: In` syscallmetrics.go` too.Fixed.

pkg/metrics/syscallmetrics/syscallmetrics.go
```diff
@@ -57,11 +57,12 @@ func Handle(event any) {
 				}
 				binary = tpEvent.Process.Binary
 			}
+			nodeName = ev.NodeName
```
FedeDP, line 60: Just a question: why are we only setting `nodeName` when:
`if tpEvent.Subsys == "raw_syscalls" && tpEvent.Event == "sys_enter" {`?
Can't we set it always? After all, we are using `ev` here (not `tpEvent`) that is readily available from the start of the function.

AritraDey-Dev: I kept it inside the if block to follow the same pattern as the other fields (namespace, workload....) and only set it when we're actually going to emit a metric (when syscall != ""). But you're right that since ev is readily available from the start.Should we move it outside the block?

FedeDP: Ah i see it now, thanks. Then it's good to go like it is now, sorry for the noise!

AritraDey-Dev:
> Thanks lgtm as well, did you check that it includes all the concerned metrics? Should we eventually remove the relabelling mentioned in #2214? (I think no but just asking).

yes i did, `events_total`, `policy_events_total`, and `syscalls_total` are all covered since they use `ProcessLabels`.I just wrote a demo script to verify that.
Regarding the relabeling, i agree we should keep it for backward compatibility.you might consider deprecating it in next release.

mtardy:
> > Thanks lgtm as well, did you check that it includes all the concerned metrics? Should we eventually remove the relabelling mentioned in #2214? (I think no but just asking).
> 
> yes i did, `events_total`, `policy_events_total`, and `syscalls_total` are all covered since they use `ProcessLabels`.I just wrote a demo script to verify that. Regarding the relabeling, i agree we should keep it for backward compatibility.you might consider deprecating it in next release.

thank you!

PR #4385, opened 2025-11-27
new(pkg/sensors): sensors maps declarative deps

FedeDP:
### Description

Am just playing with this one :) 
I think it would be very nice to be able to quickly tell `X sensor uses Y,Z,foo,bar maps`. Also, it would help in preventing bugs when we forget to add a map to a sensor (and our tests don't catch it).

### Changelog

```release-note
```

### DIFF
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -40,6 +40,35 @@ import (
 	"github.com/cilium/tetragon/pkg/sensors/program"
 )
 
+var uprobeMapsDeps = mapDeps[uprobeHas]{
+	"uprobe_calls": {},
+	"regs_map": {
+		enabled: func(_ []idtable.EntryID, has uprobeHas) bool {
+			return has.sleepableOffload
+		},
+	},
+	"sleepable_offload": {
+		enabled: func(_ []idtable.EntryID, has uprobeHas) bool {
+			return has.sleepableOffload
+		},
+		process: func(p *program.Map, _ []idtable.EntryID, _ uprobeHas) {
+			p.SetMaxEntries(sleepableOffloadMaxEntries)
+		},
+	},
+	"tg_mb_sel_opts": {
+		enabled: func(ids []idtable.EntryID, _ uprobeHas) bool {
+			// only for single sensors (non multi)
+			return ids == nil
+		},
+	},
+}.withConfigMap().withFilterMap().withRetprobeMap()
+
+var retuprobeMapsDeps = mapDeps[uprobeHas]{
+	"retuprobe_calls": {
+		flags: sensorMap,
+	},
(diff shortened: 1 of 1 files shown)


pkg/sensors/tracing/mapdeps.go
```diff
@@ -0,0 +1,92 @@
...
+
+type mapDep[HASMAP any] struct {
+	enabled mapEnabledFn[HASMAP]
+	process mapProcess[HASMAP]
+	flags   uint8
+}
+
+type mapDeps[HASMAP any] map[string]mapDep[HASMAP]
+
+func depsToMaps[HASMAP any](deps mapDeps[HASMAP], load *program.Program, multiIDs []idtable.EntryID, has HASMAP) []*program.Map {
```
FedeDP: Also, in the future this might contain a common ground for all sensors; for example enabling `cgtracker` map if its option is enabled:
```go
if option.Config.EnableCgTrackerID {
    maps = append(maps, mapDep[HASMAP]{
		enabled: func(_ []idtable.EntryID, _ HASMAP) bool {
			return option.Config.EnableCgTrackerID
		},
		flags: userMap,
	})
}
```

FedeDP:
For example a fix like https://github.com/cilium/tetragon/pull/4391 would become a single line entry in the `genericlsm` sensor with this one. Also, it would be much more discoverable imo.

dwindsor:
This seems like a lot of code to switch to a declarative model. IIUC in the current way of doing things, creation of a new sensor essentially is:

```go
myMap := program.MapBuilderSensor("my_map", load)
if someCondition {
    myMap.SetMaxEntries(1000)
}
maps = append(maps, myMap)
```

The new way is:

```go
type fooHasMaps struct {
    myFeature bool 
}

var fooMapsDeps = mapDeps[fooHasMaps]{
    "my_map": { 
        process: func(p *program.Map, _ []idtable.EntryID, has fooHasMaps) {
            if has.myFeature {
                p.SetMaxEntries(1000)
            }
        },
        flags: sensorMap,
    },
}

has := fooHasMaps{myFeature: someCondition} 
maps := fooMapsDeps.depsToMaps(load, nil, has) 
```

Not worth it imo, and now we have an extra layer to maintain just to catch these errors at build, when they should be fairly obvious imo (we wire up all maps in sensor creation/initialization methods).

FedeDP:
> Not worth it imo, and now we have an extra layer to maintain just to catch these errors at build, when they should be fairly obvious imo (we wire up all maps in sensor creation/initialization methods).

While i am not against this opinion, at the same time i feel like it would be good to be able to quickly know which maps are being used by a sensor by just giving a quick glance at the top of the file.
It's not a perfect solution for probably not even a big problem; is more like a refactor for an improved (IMO) dev experience while working with these files.

FedeDP:
On a second thought, after some private discussions, this is not really needed.
Going to close this one!

PR #4390, opened 2025-11-28
tetragon: assorted fixes

olsajiri:
assorted fixes:

```
tetragon: Move program.KeepCollection to option.Config object
    Moving program.KeepCollection to option.Config object and enabling
    it only for tests that needs it, not for all.

tetragon: Use string_maps_heap map in match_binaries
    We put struct string_prefix_lpm_trie variabe on stack (256+ bytes),
    which is not healthy, let's use heap map.

tetragon: Fix asm parseConst parsing
    Andrew reported we can't properly parse 0xffffffffffffffff constant
    in register assignment, fixing that.

tetragon: Pin selector maps only when they are used
tetragon: Pin override_tasks only when it's used
tetragon: Pin fdinstall_map only when it's used
tetragon: Pin stack_trace_map only when it's used
    There's no need to polute bpf fs with maps that we do not use.

tetragon: Fix usdt-set.yaml example policy
   Example fix.

tetragon: Use read_reg function in usdt set action
    We read register value and we already have read_reg function for
    that, let's use it.
```

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -51,7 +51,6 @@ import (
 	"github.com/cilium/tetragon/pkg/rthooks"
 	"github.com/cilium/tetragon/pkg/sensors/base"
 	"github.com/cilium/tetragon/pkg/sensors/exec/procevents"
-	"github.com/cilium/tetragon/pkg/sensors/program"
 	"github.com/cilium/tetragon/pkg/server"
 	"github.com/cilium/tetragon/pkg/tracingpolicy"
 	"github.com/cilium/tetragon/pkg/unixlisten"
@@ -236,7 +235,7 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 
 	// enable extra programs/maps loading debug output
 	if logger.GetLogger().Enabled(ctx, slog.LevelDebug) {
-		program.KeepCollection = true
+		option.Config.KeepCollection = true
 	}
 
 	log.Info("Starting tetragon", "version", version.Version)
--- a/pkg/observer/observertesthelper/observer_test_helper.go
+++ b/pkg/observer/observertesthelper/observer_test_helper.go
@@ -56,6 +56,7 @@ type testObserverOptions struct {
 	config              string
 	lib                 string
 	procCacheGCInterval time.Duration
+	keepCollection      bool
 }
 
 type testExporterOptions struct {
@@ -77,6 +78,13 @@ func WithMyPid() TestOption {
 		o.exporter.allowList = append(o.exporter.allowList, &tetragon.Filter{
 			PidSet: []uint32{GetMyPid()},
 		})
+	}
+}
+
+// Enable KeepCollection config flag
+func WithKeepCollection() TestOption {
+	return func(o *TestOptions) {
+		o.observer.keepCollection = true
 	}
 }
 
@@ -161,8 +169,9 @@ func newDefaultTestOptions(opts ...TestOption) *TestOptions {
 	// default values
 	options := &TestOptions{
 		observer: testObserverOptions{
-			config: "",
-			lib:    "",
+			config:         "",
+			lib:            "",
+			keepCollection: false,
 		},
 		exporter: testExporterOptions{
 			podAccessor: watcher.NewFakeK8sWatcher(nil),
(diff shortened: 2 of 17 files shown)


bpf/process/regs.h
```diff
@@ -133,4 +133,17 @@ write_reg(struct pt_regs *ctx, __u32 dst, __u8 size, __u64 val)
...
 
 #endif /* __TARGET_ARCH_x86 */
+
+#if defined(__TARGET_ARCH_arm64)
+FUNC_INLINE __u64
+read_reg(struct pt_regs *ctx, __u32 src, __u8 shift)
+{
+	unsigned long val = (unsigned long)-1UL, off = src & 0xfff;
+
+	probe_read_kernel(&val, sizeof(val), (void *)ctx + off);
```
FedeDP: Q: previous code was error checking the `probe_read_kernel` call; do we want to restore it?

olsajiri: hm I thought it's not used.. we do not test 'set' action on arm right? val will be 0 on error, which is likely to fail on write_user_arg, but you're right we should check it

I wonder if we could add read_reg/write_reg interface for arm.. and 'set' action test

andrewstrohman: >  val will be 0 on error,

Why do you say this? It seems like it will be ULONG_MAX, because that's what it's initialized to. I think your point still stands, that write_user_arg will fail.

For the x86 version of this function, we do return 0 if the switch statement fails to find a match. So, should we initialize val to 0?

olsajiri: I did not check across al lversions but latest upstream zeros out the dst in case of error

```
static __always_inline int
bpf_probe_read_kernel_common(void *dst, u32 size, const void *unsafe_ptr)
{
        int ret = -EFAULT;
        
        if (IS_ENABLED(CONFIG_BPF_EVENTS))
                ret = copy_from_kernel_nofault(dst, unsafe_ptr, size);
        if (unlikely(ret < 0))
                memset(dst, 0, size);
        return ret;     
}       
```

examples/tracingpolicy/symlink-override.yaml
```diff
@@ -18,12 +18,3 @@ spec:
...
       type: "string"
-    selectors:
-    - matchArgs:
-      - index: 0
-        operator: "Equal"
-        values:
-        - "/etc/passwd"
-      matchActions:
-      - action: Override
-        argError: -1
```
mtardy, line 29: why did you remove this?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -310,14 +310,13 @@ func createMultiKprobeSensor(polInfo *policyInfo, multiIDs []idtable.EntryID, ha
...
 	if has.stackTrace {
+		stackTraceMap := program.MapBuilderProgram("stack_trace_map", load)
 		stackTraceMap.SetMaxEntries(stackTraceMapMaxEntries)
+		maps = append(maps, stackTraceMap)
+		data.stackTraceMap = stackTraceMap
 	}
 
-	maps = append(maps, stackTraceMap)
-	data.stackTraceMap = stackTraceMap
-
```
mtardy, line 320: ah indeed, how did you bump into this (edit: ah looks like you bumped into another one and fixed all of them)? I wonder why it wasn't done like this in the first place, I imagine we added `has.stackTrace` after? 🤔

examples/tracingpolicy/symlink-override.yaml
olsajiri: typo, bigger one.. ;-) thnx

pkg/sensors/tracing/generickprobe.go
olsajiri: I guess.. I'd like to consolidate/cleanup the tetragon dir a bit, makes other things easier

olsajiri:
@kkourt @FedeDP I did the change and also I merged in other fixes, plz take a look

bpf/process/types/basic.h
```diff
@@ -1924,11 +1923,14 @@ FUNC_INLINE int match_binaries(__u32 selidx, struct execve_map_value *current)
 			if (!path_map)
 				return 0;
 			// prepare the key on the stack to perform lookup in the LPM_TRIE
```
mtardy: nit: I'd remove the comment then or just "on the stack"

olsajiri:
I had to rebase on top of new override changes, the change in  https://github.com/cilium/tetragon/pull/4390/changes/11d751c99fa568875424aa0b322ef78e02d29f13 is more straight forward now

PR #4393, opened 2025-11-29
tetragon: add support for substring operators

olsajiri:
Adding support for SubString/IgnCase operators for matchArg selector, to be used on top of strings like:

```
    data:
    - index: 0
      type: "string"
      source: "pt_regs"
      resolve: "rdi"
    selectors:
    - matchData:
      - index: 0
        operator: "SubString"
        values:
        - "_3_"
```
The SubString matches exact string, SubStringIgnCase matches string regardless the case.

On top of https://github.com/cilium/tetragon/pull/4489

### DIFF
--- a/pkg/sensors/tracing/uprobe_reg_test.go
+++ b/pkg/sensors/tracing/uprobe_reg_test.go
@@ -929,3 +929,115 @@ func TestUprobePtRegsPreloadSubstringIgnCaseNotMatch(t *testing.T) {
 		false, /* single */
 	)
 }
+
+func testUprobePtRegsPreloadSubstringOverride(t *testing.T, single bool) {
+	if !bpf.HasKfunc("bpf_copy_from_user_str") {
+		t.Skip("skipping, no string preload support")
+	}
+	if !bpf.HasUprobeRegsChange() {
+		t.Skip("skipping, no regs change support in kernel")
+	}
+	if !bpf.HasKfunc("bpf_strnstr") {
+		t.Skip("skipping, no bpf_strnstr kfunc in kernel")
+	}
+
+	testBinary := testutils.RepoRootPath("contrib/tester-progs/regs-override")
+
+	// Put uprobe in test_3 function at:
+	//
+	//      static const char *test_3_string = "test_3_string";
+	//
+	//      "push   %%rbp\n"          /* +0  55                            */
+	//      "mov    %%rsp, %%rbp\n"   /* +1  48 89 e5                      */
+	//      "mov    %[str], %%rdi\n"  /* +4  48 8b 3d 96 2e 00 00          */
+	//      "pop    %%rbp\n"          /* +11 5d                            */
+	//      "mov    $0x0,%%rax\n"     /* +12 48 c7 c0 00 00 00 00          */
+	// -->  "mov    $0xff,%%rax\n"    /* +19 48 c7 c0 ff 00 00 00          */
+	//      "ret\n"                   /* +26 c3                            */
+	//      :
+	//      : [str] "m" (test_3_string)
+	//
+	// Make sure we retrieve data with eax value (1) as int argument
+	// and match the expected value via matchData.
+
+	options := ""
+	if single {
+		options = `  options:
+  - name: "disable-uprobe-multi"
+    value: "1"`
+	}
+
+	pathHook := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "uprobe"
+spec:
+` + options + `
+  uprobes:
+  - path: "` + testBinary + `"
+    symbols:
+    - "test_3+19"
+    data:
+    - index: 0
+      type: "string"
+      source: "pt_regs"
+      resolve: "rdi"
+    selectors:
+    - matchData:
+      - index: 0
+        operator: "SubString"
+        values:
+        - "_3_"
+      matchActions:
+      - action: Override
+        argRegs:
+        - "rip=7%rip"
+`
+
+	pathConfigHook := []byte(pathHook)
+	err := os.WriteFile(testConfigFile, pathConfigHook, 0644)
+	if err != nil {
+		t.Fatalf("writeFile(%s): err %s", testConfigFile, err)
+	}
+
+	upChecker := ec.NewProcessUprobeChecker("UPROBE_DATA_MATCH").
+		WithProcess(ec.NewProcessChecker().
+			WithBinary(sm.Full(testBinary))).
+		WithSymbol(sm.Full("test_3+19")).
+		WithData(ec.NewKprobeArgumentListMatcher().
+			WithOperator(lc.Ordered).
+			WithValues(
+				ec.NewKprobeArgumentChecker().WithStringArg(sm.Full("test_3_string_CASE")),
+			))
+
+	checker := ec.NewUnorderedEventChecker(upChecker)
+
+	var doneWG, readyWG sync.WaitGroup
+	defer doneWG.Wait()
+
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
+	obs, err := observertesthelper.GetDefaultObserverWithFile(t, ctx, testConfigFile, tus.Conf().TetragonLib)
+	if err != nil {
+		t.Fatalf("GetDefaultObserverWithFile error: %s", err)
+	}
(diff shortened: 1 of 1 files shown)


bpf/process/generic_calls.h
```diff
@@ -579,6 +580,30 @@ FUNC_INLINE long get_pt_regs_arg(struct pt_regs *ctx, struct event_config *confi
...
+	unsigned long arg = 0;
+
+	if (ty == string_type) {
+		arg = preload_string_arg(ctx);
+
+		// Make verifier to believe it's just an ordinary number and not
+		// a pointer to the map. The rest of the argument code might do
+		// some arithmetics on it which would fail for pointer, but it's
+		// always using probe_read, so it's safe.
+		probe_read(&arg, sizeof(arg), &arg);
```
kevsecurity, line 595: Not a problem: just surprised probe_read can have @arg as source and dest! I will borrow this myself.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -210,12 +210,7 @@ func loadSingleUprobeSensor(uprobeEntry *genericUprobe, args sensors.LoadProbeAr
 		selector = uprobeEntry.loadArgs.selectors.entry
 	}
 	if selector != nil {
-		mapLoad = append(mapLoad, &program.MapLoad{
-			Name: "filter_map",
-			Load: func(m *ebpf.Map, _ string) error {
-				return m.Update(uint32(0), selector.Buffer(), ebpf.UpdateAny)
-			},
-		})
+		mapLoad = append(mapLoad, selectorsMaploads(selector, 0)...)
```
andrewstrohman, line 216: This will cause the selector maps to be included in a `program.Program`'s `MapLoad`. A `collectionLoader`'s maps is populated by a `CollectionSpec`'s maps. `CollectionSpec`'s maps are determined by the elf file.

Maps found in a `program.Program`'s MapLoad will be loaded if they are also found in the `ebpf.Collection`'s `Maps`.  `ebpf.Collection` `Maps` is populated by `collectionLoader`'s `maps`. So, just adding this is enough to load the maps.

However, when we create the sensor, we are not including the selector maps in the `sensors.Sensor`'s `Maps` member. As such, we are not doing the max entry logic, like how we do for kprobes. See [here](https://github.com/cilium/tetragon/pull/4546/commits/2f08b5f954fb6e0c9e733cff1c1f8504cabeb2d6) and [here](https://github.com/cilium/tetragon/pull/4546/commits/b8d7d4c1ac609217ccf27ca84b4b42452bec390a) for how I was thinking about handling this.

What do you think? My WIP does not prevent duplicates of the "filter_map" `program.MapLoad` like yours does. Also, there isn't an exact match between maps taken care of in `selectorsMaploads` vs my `filterMaps` function, but despite these shortcomings, it demonstrates the idea.

pkg/selectors/kernel.go
```diff
@@ -891,6 +901,21 @@ func ParseMatchData(k *KernelSelectorState, arg *v1alpha1.ArgSelector, data []v1
 	return parseMatchArg(k, arg, data, ty)
 }
 
+func writeMatchSubString(k *KernelSelectorState, values []string) error {
+	for _, v := range values {
+		id := len(k.subStrs)
+		if id >= SubstringMapEntries-1 {
```
andrewstrohman: I think this is off by one.

I think it should instead be: `if id >= SubstringMapEntries`

Consider the following example. `SubstringMapEntries` is 100. It's used to set the max number of entries in the map. So the number of subStrs should not exceed 100.

If we want to add exactly 100 subStrs, on the last iteration, `id` will be 99.
99 >= (100 - 1), is true, so we return error. But we could add that last entry without going over the map's capacity.

pkg/sensors/tracing/genericuprobe.go
olsajiri: so kprobe has that HasSelector condition for selector maps.. are you suggestion something similar for uprobe?

why do you keep the filter_map load in  https://github.com/cilium/tetragon/pull/4546/changes/b8d7d4c1ac609217ccf27ca84b4b42452bec390a ? it's part of `selectorsMaploads` right?

pkg/selectors/kernel.go
olsajiri: yep, thanks

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: > why do you keep the filter_map load in https://github.com/cilium/tetragon/commit/b8d7d4c1ac609217ccf27ca84b4b42452bec390a ? it's part of selectorsMaploads right?

Yes, that's a problem with my WIP. I agree that I shouldn't do that. I would fix that if we decided the overall approach makes sense.

> so kprobe has that HasSelector condition for selector maps.. are you suggestion something similar for uprobe?

Yes, that seems like a good way to conditionally load the selector maps based on whether they are needed.

But what I'm really getting at here is that we are loading the maps without configuring the max entries. So, I think we need to address that. It looks like the logic of loading selector maps would be the same across all hook point types. So, I'm proposing that we create a unified way to do this.

We already have `selectorsMaploads`, a unified way for creating `program.MapLoad`.  I think we should have the equivalent for generating `program.Map`. Then all hook types could reuse the same code for generating both `program.MapLoad` and `program.Map` for selector maps.

olsajiri: right, that sounds good.. let's see how it looks like wrt that each sensor needs slightly different maps, but I think it would be good to have something like that

olsajiri:
I changed the code to use core instead of the CONFIG variables, it's simpler

pkg/selectors/kernel.go
```diff
@@ -883,6 +893,21 @@ func ParseMatchData(k *KernelSelectorState, arg *v1alpha1.ArgSelector, data []v1
 	return parseMatchArg(k, arg, data, ty)
 }
 
+func writeMatchSubString(k *KernelSelectorState, values []string) error {
+	for _, v := range values {
+		id := len(k.subStrs)
+		if id >= SubstringMapEntries {
+			return fmt.Errorf("substring error: Only %d substrings allowed", SubstringMapEntries)
+		}
+		if len(v) >= 100 {
```
kkourt: Please add a constant for this, ideally with a comment pointing to the corresponding BPF value.

pkg/selectors/kernel.go
```diff
@@ -1639,6 +1672,21 @@ func HasOverride(selectors []v1alpha1.KProbeSelector) bool {
 	return false
 }
 
+func HasOperator(selectors []v1alpha1.KProbeSelector, op uint32) bool {
+	for _, s := range selectors {
+		for _, a := range slices.Concat(s.MatchArgs, s.MatchData, s.MatchReturnArgs) {
+			argOp, err := SelectorOp(a.Operator)
+			if err != nil {
+				return false
```
kkourt, line 1684: Shouldn't we continue looking at the other selectors instead of returning here?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -915,8 +936,15 @@ func createUprobeSensorFromEntry(uprobeEntry *genericUprobe,
 	filterMap := program.MapBuilderProgram("filter_map", load)
 	retProbe := program.MapBuilderSensor("retprobe_map", load)
 	selMatchBinariesMap := program.MapBuilderProgram("tg_mb_sel_opts", load)
+
 	maps = append(maps, configMap, tailCalls, filterMap, selMatchBinariesMap, retProbe)
 
+	if has.substring {
+		substringMap := program.MapBuilderSensor("substring_map", load)
+		substringMap.SetMaxEntries(selectors.SubstringMapEntries)
```
kkourt: Can't we just size this to the number of entries in the selector?

pkg/selectors/kernel.go
olsajiri: that means the selector operator is wrong and the selector should already fail or we fail shortly.. so I think there's no point to continue

pkg/sensors/tracing/genericuprobe.go
olsajiri: true, will change, thanks

pkg/selectors/kernel.go
kkourt: OK, in that case, shouldn't we return an error?

olsajiri:
note the ci failure is just syntax false alarm

PR #4395, opened 2025-12-02
Fix: Add exponential backoff to controller manager connection retry

### LINKED ISSUE #4396: Add exponential backoff to controller manager connection
Problem domain:
 The default retry duration for the controller manager to connect to the controller is 10 ms. This is too aggressive to the clients with indefinite retries.

Request:
 - To add exponential backoff retry setting.
 - Handle context cancellation between retries.

shpalani:
[This is a follow-up to PR: #4135 and #4153]

This PR contains
- Added exponential backoff retry starting with 10ms, factor of 2 and max cap of 2 secs.
- Added context cancellation before each retry.
- During backoff sleep - to check if context is concelled while waiting between retries.

Fixes #4396

### Description
1. Added exponential backoff retry settings. The progression will be
Attempt 1: No delay (immediate)
Attempt 2: 10ms delay
Attempt 3: 20ms delay
Attempt 4: 40ms delay
Attempt 5: 80ms delay
Attempt 6: 160ms delay
Attempt 7: 320ms delay
Attempt 8: 640ms delay
Attempt 9: 1.28s delay
Attempt 10: 2s delay
Attempt 11: 2s delay
Attempt 12: 2s delay
...

2. Before each retry attempt, to check if context is cancelled.

### Testing

1. On tetragon with 1 attempt. 
```
hyperadmin@simulator:~/tetragon$ sudo ./tetragon --bpf-lib bpf/objs --k8s-kubeconfig-path /home/hyperadmin/tetragon/config
level=info msg="Starting tetragon" version=42effd0cc
level=info msg="config settings" config="map[bpf-dir:tetragon bpf-lib:bpf/objs btf: cgroup-rate: cluster-name: config-dir: cpuprofile: cri-endpoint: data-cache-size:1024 debug:false disable-kprobe-multi:false enable-ancestors:[] enable-cgidmap:false enable-cgidmap-debug:false enable-cgtrackerid:true enable-compatibility-syscall64-size-type:false enable-cri:false enable-export-aggregation:false enable-k8s-api:false enable-msg-handling-latency:false enable-pid-set-filter:false enable-pod-annotations:false enable-pod-info:false enable-policy-filter:false enable-policy-filter-cgroup-map:false enable-policy-filter-debug:false enable-process-cred:false enable-process-ns:false enable-tracing-policy-crd:true event-cache-retries:15 event-cache-retry-delay:2 event-queue-size:10000 execve-map-entries:0 execve-map-size: export-aggregation-buffer-size:10000 export-aggregation-window-size:15s export-allowlist: export-denylist: export-file-compress:false export-file-max-backups:5 export-file-max-size-mb:10 export-file-perm:600 export-file-rotation-interval:0s export-filename: export-rate-limit:-1 expose-stack-addresses:false field-filters: force-large-progs:false force-small-progs:false generate-docs:false gops-address: health-server-address::6789 health-server-interval:10 k8s-controlplane-retry:1 k8s-kubeconfig-path:/home/hyperadmin/tetragon/config keep-sensors-on-exit:false kernel: log-format:text log-level:info memprofile: metrics-label-filter:namespace,workload,pod,binary metrics-server: netns-dir:/var/run/docker/netns/ pprof-address: process-cache-gc-interval:30s process-cache-size:65536 procfs:/proc/ rb-queue-size:65535 rb-size:0 rb-size-total:0 redaction-filters: release-pinned-bpf:true retprobes-cache-size:4096 server-address:localhost:54321 tracing-policy: tracing-policy-dir:/etc/tetragon/tetragon.tp.d use-perf-ring-buffer:false username-metadata:disabled verbose:0]"
level=info msg="Tetragon current security context" SELinux=unconfined AppArmor=unconfined Smack="" Lockdown=none
level=info msg="Tetragon pid file creation succeeded" pid=1267624 pidfile=/var/run/tetragon/tetragon.pid
level=info msg="BPF: successfully released pinned BPF programs and maps" bpf-dir=/sys/fs/bpf/tetragon
level=info msg="BTF discovery: default kernel btf file found" btf-file=/sys/kernel/btf/vmlinux
level=info msg="BPF detected features: override_return: true, buildid: true, kprobe_multi: true, uprobe_multi: true, fmodret: true, fmodret_syscall: true, signal: true, large: true, link_pin: true, lsm: false, missed_stats_kprobe_multi: true, missed_stats_kprobe: true, batch_update: true, uprobe_refctroff: true, audit_loginuid: true, probe_write_user: true, uprobe_regs_change: false"
level=info msg="Cgroup mode detection succeeded" cgroup.fs=/sys/fs/cgroup cgroup.mode="Unified mode (Cgroupv2)"
level=info msg="Cgroupv2 supported controllers detected successfully" cgroup.fs=/sys/fs/cgroup cgroup.path=/proc/1/root/sys/fs/cgroup cgroup.controllers="[cpuset cpu io memory hugetlb pids rdma misc]" cgroup.hierarchyID=0
level=info msg="Cgroupv2 supported controllers detected successfully" cgroup.fs=/sys/fs/cgroup cgroup.path=/sys/fs/cgroup/user.slice/user-1001.slice/session-3014.scope cgroup.controllers="[memory pids]" cgroup.hierarchyID=0
level=info msg="Cgroupv2 hierarchy validated successfully" cgroup.fs=/sys/fs/cgroup cgroup.path=/sys/fs/cgroup/user.slice/user-1001.slice/session-3014.scope
level=info msg="Deployment mode detection succeeded" cgroup.fs=/sys/fs/cgroup deployment.mode="systemd user session"
level=info msg="Updated TetragonConf map successfully" confmap-update=tg_conf_map deployment.mode="systemd user session" log.level=0 cgroup.fs.magic=Cgroupv2 cgroup.hierarchyID=0 NSPID=1267624
level=info msg="Enabling Kubernetes API"
level=info msg="created controller manager" attempts=1 duration=250.303089ms
level=info msg="Configured redaction filters" redactionFilters=""
level=info msg="Exit probe on acct_process"
level=info msg="Set execve_map entries 32768" size=28M
level=info msg="BPF ring buffer size (bytes)" total=256K
level=info msg="Loading sensor" name=__base__
level=info msg="Loading kernel version 6.8.12"
level=info msg="Loaded sensor successfully" sensor=__base__
level=info msg="Available sensors" sensors=__base__
level=info msg="Registered sensors (policy-handlers)" policy-handlers="enforcer, loader sensor, tracing"
level=info msg="Registered probe types" types="enforcer, generic_kprobe, generic_lsm, generic_tracepoint, generic_uprobe, generic_usdt, loader, execve"
level=info msg="Creating new EventCache" retries=15 delay=2s
level=info msg="Starting process manager" enableK8s=false enableProcessCred=false enableProcessNs=false
level=info msg="Exporter configuration" enabled=false fileName=""
level=info msg="Successfully detected bpftool path" bpftool=/usr/sbin/bpftool
level=warn msg="failed to locate gops binary, on bugtool debugging ensure you have gops installed"
level=info msg="Enabling policy informers"
level=info msg="Starting gRPC health server" address=:6789 interval=10
level=info msg="Starting gRPC server" protocol=tcp address=localhost:54321
level=info msg="BPF: found active BPF resources" bpf-dir=/sys/fs/bpf/tetragon pinned-bpf="[__base__ execve_map execve_map_stats execve_map_update_data tcpmon_map tg_conf_map tg_errmetrics_map tg_execve_joined_info_map tg_execve_joined_info_map_stats tg_mbset_gen tg_mbset_map tg_rb_events tg_stats_map]"
level=info msg="Read ProcFS /proc/ appended 174/236 entries"
level=info msg="Cgroup rate disabled (0/0s)"
level=info msg="Loading Tracing Policies from directory ignored, directory does not exist" tracing-policy-dir=/etc/tetragon/tetragon.tp.d
level=info msg="Perf ring buffer size (bytes)" percpu=68K total=272K
level=info msg="Events queue size (events)" size=63K
```

2. Tested with indefinite retries until successful controller connection is made.
```
time=2025-12-01T23:16:48Z level=info msg="setting retryCount to max int32 retries" 
time=2025-12-01T23:17:18Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=1
time=2025-12-01T23:17:48Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=2
time=2025-12-01T23:17:48Z level=info msg="Backing off before retry" duration=10.797783ms attempt=2
time=2025-12-01T23:18:18Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=3
time=2025-12-01T23:18:18Z level=info msg="Backing off before retry" duration=20.187635ms attempt=3
time=2025-12-01T23:18:48Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=4
time=2025-12-01T23:18:48Z level=info msg="Backing off before retry" duration=43.604029ms attempt=4
time=2025-12-01T23:19:18Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=5
time=2025-12-01T23:19:18Z level=info msg="Backing off before retry" duration=87.271282ms attempt=5
time=2025-12-01T23:19:49Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=6
time=2025-12-01T23:19:49Z level=info msg="Backing off before retry" duration=168.207238ms attempt=6
time=2025-12-01T23:20:19Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=7
time=2025-12-01T23:20:19Z level=info msg="Backing off before retry" duration=320.485942ms attempt=7
time=2025-12-01T23:20:49Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=8
time=2025-12-01T23:20:49Z level=info msg="Backing off before retry" duration=672.479544ms attempt=8
time=2025-12-01T23:21:20Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=9
time=2025-12-01T23:21:20Z level=info msg="Backing off before retry" duration=1.322916074s attempt=9
time=2025-12-01T23:21:51Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=10
time=2025-12-01T23:21:51Z level=info msg="Backing off before retry" duration=2.156202296s attempt=10
time=2025-12-01T23:22:23Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=11
time=2025-12-01T23:22:23Z level=info msg="Backing off before retry" duration=2.018142281s attempt=11
time=2025-12-01T23:22:55Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=12
time=2025-12-01T23:22:55Z level=info msg="Backing off before retry" duration=2.07709506s attempt=12
time=2025-12-01T23:23:27Z level=warn msg="failed to create controller manager" error="failed to determine if *v1.Pod is namespaced: failed to get restmapping: failed to get server groups: Get \"https://10.96.0.1:6443/api\": dial tcp 10.96.0.1:6443: i/o timeout" attempt=13
time=2025-12-01T23:23:27Z level=info msg="Backing off before retry" duration=2.10744545s attempt=13
```

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -237,9 +237,9 @@ func newControllerManagerWithRetry(ctx context.Context, cfg *rest.Config, contro
 
 	retryCount := conf.K8sConfigRetry()
 	if retryCount < 0 {
-		// max int32 retries, until connection succeeds.
-		logger.GetLogger().Info("setting retryCount to max int32 retries")
-		retryCount = math.MaxInt32
+		// max int retries, until connection succeeds.
+		logger.GetLogger().Info("setting retryCount to max int retries")
+		retryCount = math.MaxInt
 	} else if retryCount == 0 {
 		// 1 means no retries, just one connection attempt.
 		logger.GetLogger().Info("retryCount is zero, which is invalid. Defaulting to 1")
@@ -250,22 +250,27 @@ func newControllerManagerWithRetry(ctx context.Context, cfg *rest.Config, contro
 		attempts  = 0
 		startTime = time.Now()
 	)
+	const (
+		backoffFactor   = 2.0
+		backoffInterval = 2 * time.Second
+		backoffDuration = 10 * time.Millisecond
+	)
 
 	defaultRetry := retry.DefaultRetry
 	// Create a copy of the default retry with modified steps.
 	// This is to ensure that we do not modify the global default retry.
-	// We only want to change the number of steps (i.e., retries).
+	// Change the number of steps (i.e., retries) and Duration.
 	localRetry := defaultRetry
 	localRetry.Steps = retryCount
+	localRetry.Duration = 0 // Avoid default backoff time
 
 	defaultBackoff := retry.DefaultBackoff
 	// localBackoff is a copy of backoff for retries.
 	localBackoff := defaultBackoff
 	localBackoff.Steps = localRetry.Steps - 1
-	// Set exponential backoff with cap at 2 seconds
-	localBackoff.Duration = 10 * time.Millisecond // Start with 10ms
-	localBackoff.Factor = 2.0                     // Double each time
-	localBackoff.Cap = 2 * time.Second            // Cap at 2 seconds
+	localBackoff.Duration = backoffDuration // Start with backoffDuration
+	localBackoff.Factor = backoffFactor     // Multiply each time with backoffFactor
+	localBackoff.Cap = backoffInterval      // Max retry interval of backoffCap
 	logger.GetLogger().Debug("Using local retry and backoff settings",
 		"retrySteps", localRetry.Steps,
 		"backoffSteps", localBackoff.Steps,
@@ -284,16 +289,12 @@ func newControllerManagerWithRetry(ctx context.Context, cfg *rest.Config, contro
 			// Check if context is cancelled before attempting retry
 			select {
 			case <-ctx.Done():
-				logger.GetLogger().Warn("retry operation cancelled", logfields.Error, ctx.Err())
+				logger.GetLogger().Info("retry operation cancelled", logfields.Error, ctx.Err())
 				return ctx.Err()
 			default:
 			}
 
 			attempts++
-			// Wrap attempts counter to prevent integer overflow
-			if attempts > math.MaxInt32 {
-				attempts = 0
-			}
 			mgr, mgrErr := ctrl.NewManager(cfg, controllerOptions)
 			if mgrErr != nil {
 				now := time.Now()
@@ -306,14 +307,14 @@ func newControllerManagerWithRetry(ctx context.Context, cfg *rest.Config, contro
 				// Apply exponential backoff before retry
 				if attempts > 1 {
 					backoffDuration := localBackoff.Step()
-					logger.GetLogger().Info("Backing off before retry",
+					logger.GetLogger().Debug("Backoff before retry",
 						"duration", backoffDuration,
 						"attempt", attempts)
 
 					// Use context-aware sleep
 					select {
 					case <-ctx.Done():
-						logger.GetLogger().Warn("controller manager creation cancelled", logfields.Error, ctx.Err())
+						logger.GetLogger().Info("retry operation cancelled", logfields.Error, ctx.Err())
 						return ctx.Err()
 					case <-time.After(backoffDuration):
 						// Continue with retry
@@ -322,6 +323,7 @@ func newControllerManagerWithRetry(ctx context.Context, cfg *rest.Config, contro
 				// retry upon error
 				return mgrErr
 			}
+
 			cm = &ControllerManager{
 				Manager: mgr,
 			}


pkg/manager/manager.go
```diff
@@ -262,10 +262,16 @@ func newControllerManagerWithRetry(cfg *rest.Config, controllerOptions ctrl.Opti
 	// localBackoff is a copy of backoff for retries.
 	localBackoff := defaultBackoff
 	localBackoff.Steps = localRetry.Steps - 1
+	// Set exponential backoff with cap at 2 seconds
+	localBackoff.Duration = 10 * time.Millisecond // Start with 10ms
+	localBackoff.Factor = 2.0                     // Double each time
+	localBackoff.Cap = 2 * time.Second            // Cap at 2 seconds
```
FedeDP: Can we use `const`s for these values so that they are easier to spot and eventually update?

pkg/manager/manager.go
```diff
@@ -275,7 +281,19 @@ func newControllerManagerWithRetry(cfg *rest.Config, controllerOptions ctrl.Opti
...
+				logger.GetLogger().Warn("retry operation cancelled", logfields.Error, ctx.Err())
+				return ctx.Err()
+			default:
+			}
+
 			attempts++
+			// Wrap attempts counter to prevent integer overflow
+			if attempts > math.MaxInt32 {
+				attempts = 0
+			}
```
FedeDP: So let's say we do `math.MaxInt32` attempts, the next one will have no delay, and then we will restart the exponential backoff ladder, right?

pkg/manager/manager.go
```diff
@@ -285,6 +303,22 @@ func newControllerManagerWithRetry(cfg *rest.Config, controllerOptions ctrl.Opti
...
+					logger.GetLogger().Info("Backing off before retry",
+						"duration", backoffDuration,
+						"attempt", attempts)
+
+					// Use context-aware sleep
+					select {
+					case <-ctx.Done():
+						logger.GetLogger().Warn("controller manager creation cancelled", logfields.Error, ctx.Err())
+						return ctx.Err()
+					case <-time.After(backoffDuration):
```
FedeDP, line 319: Isn't it `retry.OnError` duty to wait for the `backoff.Step()`?
I see it:
```
func OnError(backoff wait.Backoff, retriable func(error) bool, fn func() error) error {
	var lastErr error
	err := wait.ExponentialBackoff(backoff, func() (bool, error) {
        err := fn()
``` 
and
```
func ExponentialBackoff(backoff Backoff, condition ConditionFunc) error {
	for backoff.Steps > 0 {
		if ok, err := runConditionWithCrashProtection(condition); err != nil || ok {
			return err
		}
		if backoff.Steps == 1 {
			break
		}
		time.Sleep(backoff.Step())
	}
	return ErrWaitTimeout
}
```

So, by sleeping us too, we are sleeping in both `fn()` called above, and `ExponentialBackoff`, right?

acamatcisco:
> Attempt 10+: 2s delay (capped)

I think 2s is a reasonable cap. If someone is actively trying to fix the connectivity, that's fast enough that they won't be confused that nothing is happening. But it's slow enough to not hammer anything like it is currently. It could probably be capped at 5 seconds, 10 seconds, or maybe even 30 seconds and be fine, but I like starting with a snappier value of 2s.

pkg/manager/manager.go
shpalani: Good catch! Looking into it.

shpalani: I digged further and tried out. The reason for adding our own:
```time.After(backoffDuration):```

retry.OnError() has a ExponentialBackoff() which does NOT support the requirement of max Cap of retry interval (in our case, it is 2 secs)
 - 3. a sleep truncated by the cap on duration has been completed.

So,
1.  If I set the localRetry.Cap = 2 secs, with localRetry.Factor = 2,, then the iteration STOPS when it reaches 2 secs. 

10 ms --> 20 ms --> 40 ms --> 80 --> 160 ms --> 320 ms --> 640ms --> 1.28s --> 2s (capped). Retry stops and returns to caller.

2. If I set the localRetry.Factor = 2 with Default duration of 10 ms, then the iteration goes beyond... This is not what we want.

10 ms --> 20 ms --> 40 ms --> 80 --> 160 ms --> 320 ms --> 640ms --> 1.28s --> 2s --> 4s --> 8s --> ... --> 1 min --> 2 min ---> 4m --> ...

---
Hence the only way to achieve the requirement of 2 secs max cap of retry interval is to have our own time.After(backoffDuration) as specified in this PR:  with

+ localRetry.Duration = 1 * time.Millisecond
+ localRetry.Factor = 1
...

Comment (FedeDP) [reply pkg/manager/manager.go:319]: I think the cap is set by the second parameter to `retry.OnError()`, ie: `retriable func(error) bool`.
It tells whether the error is a retriable one; in our case, when we reach our limit, instead of returning `true`, we should return `false`.
At that point, the `retry.OnError()` loop breaks and we are good to go. Does that work?

In any case, as the code currently is, we are both manually sleeping + we have the sleep within `OnError` and i bet this is not the desired behavior.

Comment (shpalani) [reply pkg/manager/manager.go:319]: > In any case, as the code currently is, we are both manually sleeping + we have the sleep within OnError and i bet this is not the desired behavior.

The sleep within OnError does sleep for only 10ms. It can also be configured to 0ms.
Hence the manual sleep (time.After(backoffDuration)) is to ensure we have 2 secs retry interval.

Comment (shpalani) [reply pkg/manager/manager.go:319]: > I think the cap is set by the second parameter to `retry.OnError()`, ie: `retriable func(error) bool`.

No. It is set in the localRetry struct. In this case, it was never set to any value.
```
err = retry.OnError(
		**localRetry**,
```

Comment (FedeDP) [reply pkg/manager/manager.go:319]: Sorry i explained myself bad :) 
What i meant is that we can leverage the second parameter to `OnError` to stop the exponential backoff when we are ready (ie: when we reached the cap).
Right now we pass in a `func(_ error) bool { return true },` but instead of blindly returning true, we might want to `return false` when we are done.

But perhaps i completely misunderstood the code requirement, since we don't want to ever break the wait loop, but just keep waiting until either the context gets canceled or the connection is successfully made, right?

Anyway, i think it feels a bit ankward to have the double sleep (one by us here and the other from within the `wait.Backoff::OnError()`. 
I mean, we are basically implementing the exponential backoff ourselves at this point.

Comment (shpalani) [reply pkg/manager/manager.go:319]: **Let me summarize my observations:**
Existing functionality:
1. The "Cap" in Backoff struct (defined in k8s.io/apimachinery/pkg/uti/wait/backoff.go) functionality is to retry upto the Cap time specified. **It will STOP once that CAP is reached, and return**
2. The ExponentialBackoff() (in k8s.io/apimachinery/pkg/uti/wait/backoff.go) has time.Sleep() that will **sleep for the duration set in first arg in retry.OnError(localRetry,...)** The default is 10 ms, if not set.

The requirement/request:
1. The controller manager need to retry **indefinitely** when retryCount was set to -1, until the controller connection is successful.
2. The exponential backoff of 2 secs is a RETRY INTERVAL.
10 ms --> 20 ms --> 40 ms --> 80ms --> 160 ms --> 320 ms --> 640ms --> 1.28s --> 2s --> 2s --> 2s --> 

---
With the changes,
1. The existing "Cap" in Backoff struct was not intended for RETRY INTERVAL functionality. Hence the additional  "time.After(backoffDuration)" with cancel context aware is needed to be implemented in newControllerManagerWithRetry()

Comment (FedeDP) [reply pkg/manager/manager.go:319]: > The controller manager need to retry indefinitely when retryCount was set to -1, until the controller connection is successful.

Nice, thanks i got it now :) 
As i said, using `wait.Backoff::OnError()` feels weird since we are not actually leveraging its exponential backoff feature (we are manually computing the backoff and sleeping now). 
Do you think we actually need it?

What i am proposing:
* directly use `ExponentialBackoffWithContext()` (that is context aware just like we want)
* run it in a `for` loop while the returned error is `ErrWaitTimeout`.

At that point, when the `backoff.Cap` is reached, `ExponentialBackoffWithContext` will return `ErrWaitTimeout` but our `for` loop will enforce it to restart once again.

This means:
* no custom code to handle the context errors
* no custom code to compute and wait
* we rely upon a tested code

I think my proposed change should 100% align with your requirements :) WDYT?

Comment (shpalani) [reply pkg/manager/manager.go:319]: I have addressed the comment with the fix to avoid the 2 sleeps. Please review.

In initial PR in #4135 , I used ExponentialBackoffWithContext direct call, but was proposed to change this way.

------
The provided proposal doesn't meet the requirement to maintain the RETRY INTERVAL until connection success. It will restart again at 10 ms.
for loop {
  10 ms --> 20 ms --> 40 ms --> 80ms --> 160 ms --> 320 ms --> 640ms --> 1.28s --> 2s
}

File: pkg/manager/manager.go (hunk 1 above)
Comment (shpalani) [reply pkg/manager/manager.go]: Fixed. Thanks.

File: pkg/manager/manager.go (hunk 3 above)
Comment (FedeDP) [reply pkg/manager/manager.go:319]: Linking here my proposed changes on top of this PR: https://github.com/cilium/tetragon/commit/6fd39c86f8877cb07f699cd7ae43c10ab9c9ce67

> In initial PR in #4135 , I used ExponentialBackoffWithContext direct call, but was proposed to change this way.

Probably it made sense at that time, because we were 100% relying upon it , but now that the patch you are suggesting is reimplementing the backoff wait algorithm, i think retry.OnError seems a bit pointless now.

But i don't want to block this one, let me ask for another review (cc @kkourt )

Comment (shpalani) [reply pkg/manager/manager.go]: Another suggestion is to use  ExponentialBackoffWithContext instead of ExponentialBackoff in OnError(), for all clients usage for retry.OnError(). IMHO, Cancel context aware is a must. 
```
func **OnError**(backoff wait.Backoff, retriable func(error) bool, fn func() error) error {
	var lastErr error
	err := wait.ExponentialBackoffWithContext(backoff, func() (bool, error) {
```

Comment (shpalani) [reply pkg/manager/manager.go:319]: @FedeDP The proposed fix 6fd39c8 does not work. Reason as already stated, the "Cap" (with 2 secs) is not intended for our requirement. With your diff, it runs only 8 times, and exits with panic.

10 ms --> 20 ms --> 40 ms --> 80ms --> 160 ms --> 320 ms --> 640ms --> 1.28s -- exit with panic

Comment (kkourt) [reply pkg/manager/manager.go:319]: I would agree with @FedeDP. Sleeping (waiting) in the `retriable` function passed in `wait.OnError()` seems counter-intuitive.

Comment (shpalani) [reply pkg/manager/manager.go:319]: Thanks for the follow up. So what is the next step?  Any plan to fix the ""Cap" functionality which is a blocker
?

Comment (kkourt) [reply pkg/manager/manager.go:319]: How about:

```go
func retry() error {
	var backoff = wait.Backoff{
		Steps:    10,
		Duration: 1 * time.Millisecond,
		Factor:   5.0,
		Jitter:   0.1,
	}
	durMax, _ := time.ParseDuration("2s")
	for backoff.Steps > 0 {
		// do thing, if succesful return nil
		if backoff.Steps == 1 {
			break
		}
		dur := backoff.Step()
		if dur > durMax {
			dur = durMax
		}
		time.Sleep(dur)
	}

	return wait.ErrWaitTimeout
}
```

shpalani: @kkourt : It is not cancel context-aware. Sleep is a blocking call. We would still like to keep retry.OnError. The diff in my PR serves both.

shpalani: > Sleeping (waiting) in the retriable function passed in wait.OnError() seems counter-intuitive.

@kkourt  If you look at my 2nd diff, there is no sleeping. It is made 0. I have addressed it.
localRetry.Duration = 0 // Avoid default backoff time

shpalani:
Can someone add release-note/* label? Thanks.

PR #4397, opened 2025-12-02
fix(pkg/sensors): fixed `ret{k,u}probe_calls` map builder invocation.

FedeDP:
### Description
We should be using `MapBuilderProgram` instead of `MapBuilderSensor` for multi sensors too, just like we do for single sensors.

### Changelog

```release-note
fix(pkg/sensors): fixed `ret{k,u}probe_calls` map builder invocation.
```


kevsecurity (approved):
LGTM

PR #4409, opened 2025-12-06
pkg/btf : Add resolution of array

tdaudi:
## Description

This PR add the resolution of attributes for both arrays and dynamic arrays.
It is implemented the work like so :

```yaml
- index: 0
  resolve: some.array[123].sub.field
  type: uint64
```

## Motivation

Sometimes the data you are looking for is stored inside of those arrays, you may want to have a look inside.

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -679,6 +679,15 @@ spec:
       - action: Post
 ```
 
+It is also possible to resolve arrays (`int arr[100]`) and dynamic arrays
+(`int **dyn_arr`) by using square bracket notations, similarly as follows:
+
+```yaml
+- index: 0
+  resolve: some.field[12].some.subfield
+  type: uint32
+```
+
 ## Data
 
 Kprobes allow definition of `data` fields and following `matchData` selector


mtardy:
I haven't review the content but just wanted to mention that usually the syntax is `array[idx].field` instead of `array.[idx].field`. Is the syntax with the `.` before the array index a valid JSON query?

tdaudi:
> Is the syntax with the . before the array index a valid JSON query?

Why JSON query is important in this context ?

The resolve string works using a split for every steps
https://github.com/cilium/tetragon/blob/281b3973079914fcf2a79b1a8f6eedb41f207a77/pkg/sensors/tracing/generic.go#L41

Having this syntax avoid the complexity of having to manipulate too much the string. Another reason is that because it is not possible to start the `resolve` field by using the hook parameter name, if this param is an array, it make sense for me to have something like `[idx].some.sub.field`.

mtardy:
> Why JSON query is important in this context ?

I think it's just to follow convention since nobody uses `array.[idx].field`, from a UX PoV I think it makes sense for things to just work as people would expect when using them. I understand it complicates how you parse it.

tdaudi:
What would be a valid syntax for you if one of the parameter is an array ? Because resolve always start by the resolve of the attributes, we never uses the name of the parameter. In this PR I suggest something like
`[123].sub.field`
By doing `myparam[123]` in the resolve, it also forces to do a check on `myparam` to ensure it is valid according to the `index` defined by the user. It is an unnecessary information for the `resolve` feature as well.

olsajiri:
I'd survive `array.[idx].xxx` but `array[idx]` would be better ;-) from quick look it seems not so terrible, could we try?
also AFAICS the first element is never array, right? so we don't need to care about `resolve: [xxx].field.` stuff

tdaudi:
> I'd survive array.[idx].xxx but array[idx] would be better ;-) from quick look it seems not so terrible, could we try?

No problem, i'll try to work on this later in this week

>  also AFAICS the first element is never array, right? so we don't need to care about resolve: [xxx].field. stuff

Well using [security_inode_copy_up](https://elixir.bootlin.com/linux/v6.18/source/security/security.c#L2825), for exemple we could imagine doing like this :

```yaml
- index: 1 # struct cred **
  resolve: [123].some.field
```

Even if it does not makes any sense, it should be valid

tdaudi:
Changes from v1

- Add one separate commit to change `strings.Split(arg.Resolve, ".")` in a separate function. Because we add more complexity on the path. I added a dedicated test for this.
- The ptr padding in the Resolve path is changed from "" to "[0]". Like this nested pointers will be handled as arrays

pkg/btf/btf.go
```diff
@@ -228,13 +270,40 @@ func ResolveBTFPath(
 	case *btf.Union:
 		return processMembers(btfArgs, currentType, t.Members, pathToFound, i)
 	case *btf.Pointer:
-		if len(pathToFound[i]) == 0 {
+		if idx, err := parseArrayIdxStr(pathToFound[i]); err == nil {
```
olsajiri, line 264: so this is related to the [0] change in `addPaddingOnNestedPtr` that' separate change, right? if yes, let's move it into separate commit

pkg/btf/btf.go
```diff
@@ -228,13 +270,40 @@ func ResolveBTFPath(
...
 		return ResolveBTFPath(btfArgs, ResolveNestedTypes(t.Target), pathToFound, i)
+	case *btf.Array:
+		idx, err := parseArrayIdxStr(pathToFound[i])
+		if err != nil {
+			return nil, fmt.Errorf("fail parsing array index : %w", err)
+		}
+		if idx >= t.Nelems {
+			return nil, fmt.Errorf("array index out of bound. Nelems=%d, got=%d", t.Nelems, idx)
+		}
+		target := ResolveNestedTypes(t.Type)
```
olsajiri: this deserves some more comments.. also in the commit log.. in few months we will be lost in here ;-) thanks

pkg/btf/btf.go
```diff
@@ -228,13 +270,40 @@ func ResolveBTFPath(
...
+		(*btfArgs)[i].IsInitialized = uint16(1)
+		(*btfArgs)[i].Offset = uint32(getSizeofType(target) * idx)
+		if ptr, ok := t.Type.(*btf.Pointer); ok {
+			(*btfArgs)[i].IsPointer = uint16(1)
+			target = ResolveNestedTypes(ptr.Target)
+		}
+		if len(pathToFound) > i+1 {
+			return ResolveBTFPath(btfArgs, target, pathToFound, i+1)
```
olsajiri: I wonder if it would make the code easier if we move this check in the beginning of the `ResolveBTFPath` and if there's no more work we will just return currentType as last and we would get rid of checks like:

```
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -278,10 +278,7 @@ func ResolveBTFPath(
                        if i > 0 {
                                (*btfArgs)[i-1].IsPointer = uint16(1)
                        }
-                       if len(pathToFound) > i+1 {
-                               return ResolveBTFPath(btfArgs, target, pathToFound, i+1)
-                       }
-                       return &target, nil
+                       return ResolveBTFPath(btfArgs, target, pathToFound, i+1)
                }
                (*btfArgs)[i-1].IsPointer = uint16(1)
                return ResolveBTFPath(btfArgs, ResolveNestedTypes(t.Target), pathToFound, i)
@@ -300,10 +297,7 @@ func ResolveBTFPath(
                        (*btfArgs)[i].IsPointer = uint16(1)
                        target = ResolveNestedTypes(ptr.Target)
                }
-               if len(pathToFound) > i+1 {
-                       return ResolveBTFPath(btfArgs, target, pathToFound, i+1)
-               }
-               return &target, nil
+               return ResolveBTFPath(btfArgs, target, pathToFound, i+1)
        default:
                ty := currentType.TypeName()
                if len(ty) == 0 {

```

same for `processMembers` ... I might be overlooking something, so if it's not better, feel free to ignore

pkg/sensors/tracing/generic.go
```diff
@@ -38,7 +75,7 @@ func hasPtRegsSource(arg *v1alpha1.KProbeArg) bool {
 
 func resolveBTFType(arg *v1alpha1.KProbeArg, ty ebtf.Type) (*ebtf.Type, [api.MaxBTFArgDepth]api.ConfigBTFArg, error) {
 	btfArg := [api.MaxBTFArgDepth]api.ConfigBTFArg{}
-	pathBase := strings.Split(arg.Resolve, ".")
+	pathBase := formatBTFPath(arg.Resolve)
```
olsajiri: so this is prereq change for the array parsing support, right? but in commit log I see it in commit after the support is added.. could you please re-arange?

pkg/btf/btf.go
tdaudi: From my tests, my frankenstein algo has become too complex to do this. Though I put this common logic into `processArray`. Like this it looks a bit cleaner. Let me know what you think

pkg/btf/btf.go
tdaudi: Yes sure sorry. I hope now it's a bit more straight forward. Please let me know !

pkg/btf/btf.go
tdaudi: You are absolutely correct, thank you for noticing. This should be better now

pkg/sensors/tracing/generic.go
tdaudi: Sorry thank you for noticing. I re-arange all of this.

pkg/sensors/tracing/generic_test.go
```diff
@@ -16,6 +16,36 @@ import (
...
+		{
+			name:   "array as first resolve",
+			input:  "[1].path.to.my.field",
+			output: []string{"[1]", "path", "to", "my", "field"},
+		},
+		{
+			name:   "array in the resolve path",
+			input:  "path.to[123].my.field",
+			output: []string{"path", "to", "[123]", "my", "field"},
+		},
```
mtardy, line 85: I'd maybe add a few tests cases to be sure that you are okay with the output with weird inputs like double dots, nested arrays, dots within arrays, etc.

pkg/sensors/tracing/generic.go
mtardy, line 1: On the first commit message

```
Because we want to include argument filtering in the resolve path, we
need to make a split on the "." as long as one on the "[" when it
matches a bracket. This commit add a function like this, as long the
test.
One other thing is that it changes the way the "padding" in the resolve
path is registered. Until now, nested ptr such as (struct cred**) were
registered as "" in the btf path. Now, we consider the same as getting
the first parameter of an array. So "" is replaced by "[0]". But the
behaviour remain the same.
```

Just a remark: I think you want to say "as well" or "along" instead "as long"

Thanks for writing commit description! Also nit advice, but you can remove "One other thing is that" or "this commit adds" or "the idea of this commit" by just saying the sentence without starting with these introduction words.

pkg/btf/btf.go
```diff
@@ -224,6 +224,25 @@ func parseArrayIdxStr(s string) (uint32, error) {
...
+	case *btf.Struct:
+		return t.Size
+	case *btf.Void:
+		return 0
+	case *btf.Float:
+		return t.Size
+	default:
+		return 0
+	}
+}
```
mtardy, line 230: Could you use `btf.Sizeof` directly?

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -679,6 +679,16 @@ spec:
       - action: Post
 ```
 
+It is also possible to resolve arrays (`int arr[100]`) and dynamic arrays (`int **dyn_arr`) by resolving like so :
```
mtardy:
```suggestion
It is also possible to resolve arrays (`int arr[100]`) and dynamic arrays (`int **dyn_arr`) 
by using square bracket notations, similarly as follows:
```

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -679,6 +679,16 @@ spec:
...
+
+```yaml
+- index: 0
+  resolve: some.field[12].some.subfield
+  type: uint32
+```
+
+It will work exactly the same way for both `arr` and `dyn_arr`.
+
```
mtardy:
```suggestion
```

pkg/btf/btf.go
```diff
@@ -329,3 +355,30 @@ func processMembers(
...
+) (*btf.Type, error) {
+	var target btf.Type
+	if t, ok := currentType.(*btf.Array); ok {
+		target = ResolveNestedTypes(t.Type)
+	} else if t, ok := currentType.(*btf.Pointer); ok {
+		target = ResolveNestedTypes(t.Target)
+		(*btfArgs)[i].IsPointer = uint16(1)
+		if i > 0 {
+			(*btfArgs)[i-1].IsPointer = uint16(1)
+		}
```
mtardy: so all the work done with the pointer to you fixed size array looks a bit hard to read, I don't have much time to propose something now but I wonder if you could simplify this code.

Do you really need a pointer to your array? It's already pointing to your array elements.

pkg/sensors/tracing/generic_test.go
tdaudi: Initially the goal of the func was not to test the input was correct but simply that the resolve path is formated correctly. Because in any case an error would occures on the Resolve algo later.

But seeing your comment it's probably better to enforce the parsing with this format check. So, I changd dthe func to handle wrong formats and return an error. 
Let me know what you think

pkg/sensors/tracing/generic.go
tdaudi: Arf yes thank you for noticing. I'll try to do better next time

docs/content/en/docs/concepts/tracing-policy/hooks.md
tdaudi: I removed it, thanks

pkg/btf/btf.go
tdaudi: The idea it to always stay ahead on the dereference mechanism, so we need `(*btfArgs)[i].IsPointer = uint16(1)`. 
And because we are in a case where previous elem and current are pointers, `(*btfArgs)[i-1].IsPointer = uint16(1)` is also needed.

I refactor the `processArray` to make it shorter. I would not say it makes the reading easier but since I added few more comments for this behaviour I hope it will help for the review.
Thanks

docs/content/en/docs/concepts/tracing-policy/hooks.md
tdaudi: Done, thank you !

pkg/btf/btf.go
tdaudi: Nice catch thank you ! 
I kept getSizeofType though because btf.Sizeof return 0 when an error occures which is fine (no error should occurs either).
Also, in the futur, i'd like to pimp a bit getSizeofType to handle `void *` offset calculation.

contrib/tester-progs/uprobe-resolve.c
```diff
@@ -58,6 +60,19 @@ int main(int argc, char *argv[])
 		s.v64 = val;
 	} else if (!strcmp(field, "sub.v32")) {
 		s.sub.v32 = val;
+	} else if (
+		(field[0] == 'a' && field[1] == 'r' && field[2] == 'r') || (field[0] == 'd' && field[1] == 'y' && field[2] == 'n')) {
```
olsajiri: please keep it on the same line with `if (` and split the condition if the line's too long

pkg/btf/btf.go
```diff
@@ -306,3 +347,18 @@ func processMembers(
...
+	i int,
+	idx uint32,
+) (*btf.Type, error) {
+	(*btfArgs)[i].IsInitialized = uint16(1)
+	(*btfArgs)[i].Offset = getSizeofType(targetType) * idx
+	if len(pathToFound) > i+1 {
+		return ResolveBTFPath(btfArgs, targetType, pathToFound, i+1)
+	}
+	return &targetType, nil
+}
```
olsajiri, line 364: it's great you did it with single `processArray` function, I just tried to make `processArray` local function in `ResolveBTFPath` and it seems bit more readable to me (less arguments to pass) but I don't mind keeping it this way as well

olsajiri (approved):
looks good, left 2 nits, thanks

pkg/sensors/tracing/generic_test.go
mtardy: ah I think it was okay to try your best without returning an explicit error, just making sure that weird syntax were simplified as expected. But it's also good like that! :)

pkg/btf/btf.go
mtardy: ah I was confused about differences between slices and arrays in Go. With slices you wouldn't need to pass pointers, with arrays you do need to do what you do!

Small nit: you can simplify your `(*array)[n] = x` assignments: https://go.dev/play/p/3qRC3yHegrm, see touch1, you don't need to dereference the array identifier.

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -679,6 +679,14 @@ spec:
       - action: Post
 ```
 
+It is also possible to resolve arrays (`int arr[100]`) and dynamic arrays (`int **dyn_arr`) by using square bracket notations, similarly as follows:
```
mtardy: small nit: if you can wrap at ~80 chars that would be amazing.

pkg/sensors/tracing/generic.go
mtardy: no prob, it's better to focus on what to say in the commit description than how to say it, we are all making typos in these :)

contrib/tester-progs/uprobe-resolve.c
tdaudi: Ah yes it should be better now, thanks.
Though, the indent on GH does not reflect the indent on Git.
- GitHub and Vscode display the code the same way
- For Vim & Git, the condition is not aligned

No big deal anyway. I'm guessing it's a font issue but that's not ideal

pkg/btf/btf.go
tdaudi: If you don't mind I'd prefer keep it split and see the switch as a router with less logic as possible. I personally like the idea of spiting every behavior in different "branches" to not be confused on which part of the code is doing what.
Also in the future I'd like to pimp `processArray` to handle casting attributes and have a better control on void* and buffers in general.

docs/content/en/docs/concepts/tracing-policy/hooks.md
tdaudi: Thank you, done

pkg/btf/btf.go
tdaudi: Oh cool that's definitely better, thank you.
I changed all the `(*array)[n]` to `array[n]` in the algo as well in the commit

tdaudi:
Thank you for the reviews.
I applied the few nit changes, let me know if you see anything else

PR #4416, opened 2025-12-09
metrics: Add debug metric for non-critical errors

AritraDey-Dev:
Part of #2785 
>  Stop reporting non-errors (as in: no action needed) as errors. Define separate metrics for "casual fails" if needed.

### Description

Adds `tetragon_debug_events_total` metric for non-critical errors to reduce noise in `tetragon_errors_total`. (see commit)

Example Output:
```
# HELP tetragon_debug_events_total The total number of Tetragon debug events. For internal use only.
# TYPE tetragon_debug_events_total counter
tetragon_debug_events_total{type="process_metadata_username_ignored_not_in_host_namespaces"} 1
```

### Changelog

```release-note
Added tetragon_debug_events_total metric to separate non-critical issues from actual errors.
```

### DIFF
--- a/pkg/metrics/errormetrics/errormetrics.go
+++ b/pkg/metrics/errormetrics/errormetrics.go
@@ -24,16 +24,12 @@ const (
 	EventFinalizeProcessInfoFailed
 	// Failed to resolve Process uid to username
 	ProcessMetadataUsernameFailed
-	// The username resolution was skipped since the process is not in host
-	// namespaces.
-	ProcessMetadataUsernameIgnoredNotInHost
 )
 
 var errorTypeLabelValues = map[ErrorType]string{
-	ProcessPidTidMismatch:                   "process_pid_tid_mismatch",
-	EventFinalizeProcessInfoFailed:          "event_finalize_process_info_failed",
-	ProcessMetadataUsernameFailed:           "process_metadata_username_failed",
-	ProcessMetadataUsernameIgnoredNotInHost: "process_metadata_username_ignored_not_in_host_namespaces",
+	ProcessPidTidMismatch:          "process_pid_tid_mismatch",
+	EventFinalizeProcessInfoFailed: "event_finalize_process_info_failed",
+	ProcessMetadataUsernameFailed:  "process_metadata_username_failed",
 }
 
 func (e ErrorType) String() string {
@@ -106,6 +102,7 @@ var (
 func RegisterMetrics(group metrics.Group) {
 	group.MustRegister(ErrorTotal)
 	group.MustRegister(HandlerErrors)
+	group.MustRegister(DebugTotal)
 }
 
 func InitMetrics() {
@@ -121,6 +118,10 @@ func InitMetrics() {
 	// NB: We initialize only ops.MSG_OP_UNDEF here, but unknown_opcode can occur for any opcode
 	// that is not explicitly handled.
 	GetHandlerErrors(ops.MSG_OP_UNDEF, HandlePerfUnknownOp).Add(0)
+
+	for er := range debugTypeLabelValues {
+		GetDebugTotal(er).Add(0)
+	}
 }
 
 // Get a new handle on an ErrorTotal metric for an ErrorType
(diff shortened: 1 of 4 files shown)


mtardy (commented):
I agree this is an improvement thanks! But just taking a step back here do we care about this warning at all in general? Should we just drop this altogether instead?

It looks like it was added there https://github.com/cilium/tetragon/pull/2588/changes/28ce5efe6d0c1d586adc9cb50ea49c572cb3ccf5. It looks like it was seen as useful during the review @kkourt, any memory?

pkg/metrics/errormetrics/errormetrics.go
```diff
@@ -103,9 +99,63 @@ var (
...
+func GetHandlerErrors(opcode ops.OpCode, er EventHandlerError) prometheus.Counter {
+	return HandlerErrors.WithLabelValues(strconv.Itoa(int(int32(opcode))), er.String())
+}
+
+type WarningType int
+
+const (
+	// The username resolution was skipped since the process is not in host
+	// namespaces.
+	ProcessMetadataUsernameIgnoredNotInHost WarningType = iota
```
mtardy:
```suggestion
	ProcessMetadataUsernameIgnoredNotInHost DebugType = iota
```

mtardy (commented):
maybe we can move this from error to debug/note or an independent metric (with debug in its name or something)

AritraDey-Dev:
> maybe we can move this from error to debug/note or an independent metric (with debug in its name or something)

`tetragon_debug_events_total` how does this sound?

"There are 2 hard problems in computer science: cache invalidation, naming things, and off-by-1 errors." --- Leon Bambrick😅

pkg/metrics/errormetrics/errormetrics.go
```diff
@@ -121,24 +171,18 @@ func InitMetrics() {
...
-	GetErrorTotal(er).Inc()
+	for er := range debugTypeLabelValues {
+		GetDebugTotal(er).Add(0)
+	}
 }
 
-// Get a new handle on the HandlerErrors metric
-func GetHandlerErrors(opcode ops.OpCode, er EventHandlerError) prometheus.Counter {
-	return HandlerErrors.WithLabelValues(strconv.Itoa(int(int32(opcode))), er.String())
+// Get a new handle on an DebugTotal metric for an DebugType
```
mtardy:
```suggestion
// Get a new handle on a DebugTotal metric for a DebugType
```

pkg/metrics/errormetrics/errormetrics.go
```diff
@@ -121,24 +171,18 @@ func InitMetrics() {
...
+func GetDebugTotal(er DebugType) prometheus.Counter {
+	return DebugTotal.WithLabelValues(er.String())
 }
 
-// Increment the HandlerErrors metric
-func HandlerErrorsInc(opcode ops.OpCode, er EventHandlerError) {
-	GetHandlerErrors(opcode, er).Inc()
+// Increment an DebugTotal for an DebugType
```
mtardy:
```suggestion
// Increment a DebugTotal for a DebugType
```

mtardy (commented):
yeah just two nits that should be quick to fix if you're okay:
- small typos in comment
- could you put this stuff into a new `debugmetrics.go` file in the same `errormetrics` pkg?

I'm also ok to merge like that if you feel that's better

AritraDey-Dev:
>could you put this stuff into a new debugmetrics.go file in the same errormetrics pkg?

Yeah, that would be much more structured.

pkg/metrics/errormetrics/errormetrics.go
```diff
@@ -103,9 +99,30 @@ var (
...
+// Increment an ErrorTotal for an ErrorType
+func ErrorTotalInc(er ErrorType) {
+	GetErrorTotal(er).Inc()
+}
+
+// Get a new handle on the HandlerErrors metric
+func GetHandlerErrors(opcode ops.OpCode, er EventHandlerError) prometheus.Counter {
+	return HandlerErrors.WithLabelValues(strconv.Itoa(int(int32(opcode))), er.String())
+}
+
```
mtardy: Then you don't need those move lines changes hopefully

mtardy (approved):
awesome let's merge like this once green, thanks again

PR #4417, opened 2025-12-09
fix(tests): Adjust skeleton test timeouts to reduce flakyness

dangome3:
Increase the curl event checker time limit from 30 to 60 seconds and change the curl container imagePullPolicy from Always to IfNotPresent.

Tested several times:
```shiell
--- PASS: TestSkeletonBasic (15.05s)
    --- PASS: TestSkeletonBasic/Get_Minimum_Kernel_Version (0.00s)
        --- PASS: TestSkeletonBasic/Get_Minimum_Kernel_Version/Lookup_Kernel_Version (0.00s)
    --- PASS: TestSkeletonBasic/Run_Event_Checks (2.98s)
        --- PASS: TestSkeletonBasic/Run_Event_Checks/Run_Event_Checks (2.98s)
    --- PASS: TestSkeletonBasic/Run_Workload (15.05s)
        --- PASS: TestSkeletonBasic/Run_Workload/Wait_for_Checker (0.00s)
        --- PASS: TestSkeletonBasic/Run_Workload/Run_Workload (10.02s)
        --- PASS: TestSkeletonBasic/Run_Workload/Run_Metrics_Checks (0.02s)
        --- PASS: TestSkeletonBasic/Run_Workload/Uninstall_policy (5.01s)
```

Found several pipelines failing due to timeout issue so I increased them. The execution time should not be affected when the network is working properly, the current change just add additional changes to make the tests work

### DIFF
--- a/tests/e2e/tests/skeleton/skeleton_test.go
+++ b/tests/e2e/tests/skeleton/skeleton_test.go
@@ -87,8 +87,8 @@ func TestMain(m *testing.M) {
 func TestSkeletonBasic(t *testing.T) {
 	// Grab the minimum kernel version in all cluster nodes and define an RPC checker with it
 	kversion := helpers.GetMinKernelVersion(t, runner.Environment)
-	// Create an curl event checker with a limit or 10 events or 30 seconds, whichever comes first
-	curlChecker := curlEventChecker(kversion).WithEventLimit(100).WithTimeLimit(30 * time.Second)
+	// Create an curl event checker with a limit or 200 events or 120 seconds, whichever comes first
+	curlChecker := curlEventChecker(kversion).WithEventLimit(200).WithTimeLimit(120 * time.Second)
 
 	metricsChecker := metricschecker.NewMetricsChecker("skeletonMetricsChecker")
 
@@ -103,8 +103,8 @@ func TestSkeletonBasic(t *testing.T) {
 
 	// This feature waits for curlChecker to start then runs a custom workload.
 	runWorkload := features.New("Run Workload").
-		/* Wait up to 30 seconds for the event checker to start before continuing */
-		Assess("Wait for Checker", curlChecker.Wait(30*time.Second)).
+		/* Wait up to 60 seconds for the event checker to start before continuing */
+		Assess("Wait for Checker", curlChecker.Wait(60*time.Second)).
 		/* Run the workload */
 		Assess("Run Workload", func(ctx context.Context, _ *testing.T, c *envconf.Config) context.Context {
 			ctx, err := helpers.LoadCRDString(namespace, curlPod, true)(ctx, c)


FedeDP:
> Tested several times:

Question: were you able to reproduce the issue before your changes?

dangome3:
> > Tested several times:
> 
> Question: were you able to reproduce the issue before your changes?

There are some pipelines that failed

tests/e2e/tests/skeleton/skeleton_test.go
```diff
@@ -87,8 +87,8 @@ func TestMain(m *testing.M) {
 func TestSkeletonBasic(t *testing.T) {
 	// Grab the minimum kernel version in all cluster nodes and define an RPC checker with it
 	kversion := helpers.GetMinKernelVersion(t, runner.Environment)
-	// Create an curl event checker with a limit or 10 events or 30 seconds, whichever comes first
-	curlChecker := curlEventChecker(kversion).WithEventLimit(100).WithTimeLimit(30 * time.Second)
+	// Create an curl event checker with a limit or 200 events or 120 seconds, whichever comes first
+	curlChecker := curlEventChecker(kversion).WithEventLimit(200).WithTimeLimit(120 * time.Second)
```
dangome3, line 91: Adjusted the time to make it more stable. It will not affect the time when it works, only give more chances to success when the infraestructure/network is a bit slower

tests/e2e/tests/skeleton/skeleton_test.go
```diff
@@ -174,7 +174,7 @@ spec:
       containers:
         - name: curl
           image: docker.io/curlimages/curl:latest
-          imagePullPolicy: Always
+          imagePullPolicy: IfNotPresent
```
AritraDey-Dev: Why is `imagePullPolicy` changed to `IfNotPresent`? Since we are using the latest tag, if a node has a stale image cached, it won't pull the update. This could lead to flaky tests for local testing too.

dangome3: I was thinking it is a good optimization but you are right!
I already updated it to `Always`

Thanks for the suggestion

tests/e2e/tests/skeleton/skeleton_test.go
mtardy: should the time limit be the same as the wait limit nearby `Wait(60*time.Second)`? So 120 s everywhere or 200 s everywhere? I don't exactly knows what `.Wait` does on a checker out of the top of my head.

dangome3: Hey @mtardy!

Thanks for your comment. The `.Wait` method defines how long to wait for the checker to start. So it can be shorter as it should start almost instantly once the gRPC connection is established

We can also update it to 120s, that shouldn't affect the execution time either but I didn't see any timeout related to the checker starting phase, so it is unnecessary IMO

Please let me know if you have other opinions :)

mtardy:
> > > Tested several times:
> > 
> > 
> > Question: were you able to reproduce the issue before your changes?
> 
> There are some pipelines that failed

Yeah I was also curious about that, did it failed locally when testing several time? Not sure it will improve the situation otherwise but this looks 😇 enough.

PR #4421, opened 2025-12-10
Separate ProcessLoader ancestors configuration

t0x01:
### Description
Add a separate configuration option for enabling ancestors for `ProcessLoader` events similar to other event types with existing ancestors support. Ancestors support for `ProcessLoader` was added in [PR 4376](https://github.com/cilium/tetragon/pull/4376) and currently the `EnableProcessUprobeAncestors` option is used for both `ProcessUprobe` and `ProcessLoader` events.

### Changelog
```release-note
Add a separate configuration option for enabling ancestors for ProcessLoader events
```


FedeDP (approved):
Let me cc @olsajiri to make sure he is ok with this :) but LGTM!

PR #4426, opened 2025-12-13
elf: handle spaces found in arm64 stapsdt

andrewstrohman:
When parsing arguments, we first split the string found in the stapdt section by a space before trying to match against the resulting tokens. On arm64, I encountered some spaces between the reg (sp) and index_reg (12).

objdump -s -j .note.stapsdt arp.o

arp.o:     file format elf64-littleaarch64

Contents of section .note.stapsdt:
 0000 08000000 36000000 03000000 73746170  ....6.......stap
 0010 73647400 00000000 00000000 00000000  sdt.............
 0020 00000000 00000000 00000000 74657472  ............tetr
 0030 61676f6e 00617270 002d3440 5b73702c  agon.arp.-4@[sp,
 0040 2031325d 20384078 31000000            12] 8@x1...

This change removes spaces found within the brackets so that downstream parsing works as expected.

I discovered this on arm64 while attempting to use the Set matchAction to write to USDT argument. 
```
        matchActions:
          - action: Set
            argIndex: 0
            argValue: 1

```
I wasn't able to load the policy prior to this change. It failed with this error:
```
level=error msg="Failed to start tetragon" error="add TracingPolicy failed: failed to get sensors from (/host/src/live-protect/examples/v1.19/arp.yaml) parser policy: policy handler 'tracing' failed loading policy 'arp': failed to configure usdt 'tetragon/arp', set action argument is not 'deref' type: ''"
```
With this change, I verified that the Set action worked as expected.


pkg/elf/usdtargs_linux.go
```diff
@@ -10,9 +10,30 @@ import (
...
+			inside = true
+		case ']':
+			inside = false
+		}
+		if inside && letter == ' ' {
+			continue
+		}
+		ret.WriteRune(letter)
+	}
+	return ret.String()
```
olsajiri, line 30: yea, that's pain of the string parsing.. there was similar issue in register parsing, that's why I added RegScanner .. not sure there's better way to handle all this, looks like fmtSScanf can't handle blanks nicely

could you please add tests for this? thanks

olsajiri (approved):
great, thanks

mtardy (approved):

PR #4427, opened 2025-12-14
tetragon: factor some execve tests

olsajiri:
using single observer for multiple tests, it saves time and lines


olsajiri:
note the CI fail is just for too long subjects

FedeDP (approved):
Love it, thanks!

PR #4439, opened 2025-12-16
tetragon: split process event tail call

olsajiri:
Split generic_[ku]probe_process_event into two separate tail calls, so we have space (there's 4096 instructions limit on 4.19 kernels) for new features in processing event code.
 
```
Before:
      generic_kprobe_process_event 4029 insns
      generic_uprobe_process_event 4029 insns
 
Now:
      generic_uprobe_process_event   2795 insns
      generic_uprobe_process_event_2 3147 insns
    
      generic_kprobe_process_event   2795 insns
      generic_kprobe_process_event_2 3147 insns
```

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -423,7 +423,7 @@ __read_arg_2(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
  * Returns the size of data appended to @args.
  */
 FUNC_INLINE long
-read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int argm)
+read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int argm, int process)
 {
 	size_t min_size = type_to_min_size(type, argm);
 	struct msg_generic_kprobe *e;
@@ -449,6 +449,16 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
 	if (path_arg)
 		return copy_path(args, path_arg);
 
+	/*
+	 * Separate argument processing based on the process const
+	 * for 4.19 kernels..
+	 */
+	if (process == __READ_ARG_1)
+		return __read_arg_1(ctx, type, orig_off, arg, argm, args);
+	if (process == __READ_ARG_2)
+		return __read_arg_2(ctx, type, orig_off, arg, argm, args);
+
+	/* .. and the rest of the world */
 	if (is_read_arg_1(type))
 		return __read_arg_1(ctx, type, orig_off, arg, argm, args);
 	else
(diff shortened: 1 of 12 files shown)


bpf/process/generic_calls.h
```diff
@@ -422,6 +450,11 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
...
 		return copy_path(args, path_arg);
 
+	if (process == __READ_ARG_1)
+		return __read_arg_1(ctx, type, orig_off, arg, argm, args);
+	if (process == __READ_ARG_2)
+		return __read_arg_2(ctx, type, orig_off, arg, argm, args);
+
 	size = __read_arg_1(ctx, type, orig_off, arg, argm, args);
 	if (!size)
 		size = __read_arg_2(ctx, type, orig_off, arg, argm, args);
```
andrewstrohman: Maybe it's worth considering doing this instead:

```
if is_read_arg_1(type) {
    return __read_arg_1(ctx, type, orig_off, arg, argm, args);
} else {
    return __read_arg_2(ctx, type, orig_off, arg, argm, args);
}
```

I suggest this because it makes it more clear that if you handle a new type in `__read_arg_1`, you also need to update `is_read_arg_1`. Otherwise, if a developer's preliminary testing is on a modern kernel, they may not discover the connection between these two functions until they run vmtests and see that 4.19 doesn't work as expected.

Also, while returning zero from `__read_arg_1` doesn't make sense for success, and therefore is a suitable way to signal that `__read_arg_2` should be called, we could avoid making this new meaning/connection by being explicit when checking `is_read_arg_1`. This way, we would be free to use the zero return value case for some other meaning in the future. (A meaning that applies both to `__read_arg_1` and `__read_arg_2`)

olsajiri: makes sense, will change, thanks

bpf/process/generic_calls.h
```diff
@@ -392,6 +409,62 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
...
+	/*
+	 * Separate argument processing based on the process const
+	 * for 4.19 kernels..
+	 */
+	if (process == __READ_ARG_1)
+		return __read_arg_1(ctx, type, orig_off, arg, argm, args);
+	if (process == __READ_ARG_2)
+		return __read_arg_2(ctx, type, orig_off, arg, argm, args);
+
+	/* .. and the rest of the world */
```
kkourt, line 461:
```suggestion
	/* .. and the rest of the world (i.e., process == __READ_ARG_ALL) */
```

bpf/process/generic_calls.h
```diff
@@ -517,6 +591,13 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 	ty = config->arg[index];
 	am = config->arm[index];
 
+#ifndef __LARGE_BPF_PROG
+#if defined(GENERIC_KPROBE) || defined(GENERIC_UPROBE)
```
andrewstrohman: This is missing GENERIC_TRACEPOINT. It might be the reason why the CI is failing after tacking on the latest commit to enable this for tracepoints as well.

PR #4443, opened 2025-12-19
support uprobe register override on arm64

andrewstrohman:
This PR adds support for register/return override on arm64. Where x86 specific code could be re-used, it was. I adapted the related x86_64-specific unit tests to also work for arm64.

Since `struct pt_regs` on arm64 doesn't have member names that correspond to registers, and instead stores all general purpose registers in an array, our data resolve functionality does not work well to match on register values. I think we have room for improvement in this regard.

### DIFF
--- a/pkg/sensors/tracing/uprobe_reg_test.go
+++ b/pkg/sensors/tracing/uprobe_reg_test.go
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of Tetragon
 
-//go:build amd64 && linux
+//go:build linux
 
 package tracing
 
@@ -9,6 +9,7 @@ import (
 	"context"
 	"os"
 	"os/exec"
+	"runtime"
 	"strconv"
 	"sync"
 	"testing"
@@ -92,17 +93,7 @@ func TestUprobeOverrideRegsAction(t *testing.T) {
 
 	testBinary := testutils.RepoRootPath("contrib/tester-progs/regs-override")
 
-	// Put uprobe in test_1 function at:
-	//
-	//     "push   %rbp\n"         /* +0  55             */
-	//     "mov    %rsp,%rbp\n"    /* +1  48 89 e5       */
-	//     "mov    $0x1,%eax\n"    /* +4  b8 01 00 00 00 */
-	// --> "mov    $0x3,%eax\n"    /* +9  b8 03 00 00 00 */
-	//     "pop    %rbp\n"         /* +14 5d             */
-	//     "ret\n"                 /* +15 c3             */
-	//
-	// Make sure uprobe overrides test_1 return value (with 11)
-	// and the rest of the function is not executed.
+	var symbol string
 
 	pathHook := `
 apiVersion: cilium.io/v1alpha1
@@ -112,8 +103,24 @@ metadata:
 spec:
   uprobes:
   - path: "` + testBinary + `"
+`
+	switch runtime.GOARCH {
+	case "amd64":
+		// Put uprobe in test_1 function at:
+		//
+		//     "push   %rbp\n"         /* +0  55             */
+		//     "mov    %rsp,%rbp\n"    /* +1  48 89 e5       */
+		//     "mov    $0x1,%eax\n"    /* +4  b8 01 00 00 00 */
+		// --> "mov    $0x3,%eax\n"    /* +9  b8 03 00 00 00 */
+		//     "pop    %rbp\n"         /* +14 5d             */
+		//     "ret\n"                 /* +15 c3             */
+		//
+		// Make sure uprobe overrides test_1 return value (with 11)
+		// and the rest of the function is not executed.
+		symbol = "test_1+9"
+		pathHook += `
     symbols:
-    - "test_1+9"
+    - "` + symbol + `"
     selectors:
     - matchActions:
       - action: Override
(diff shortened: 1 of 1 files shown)


pkg/asm/assignment_linux_arm64_test.go
mtardy, line 1: instead of duplicating this file, can we just remove `_arm64` and `_amd64` here and have one file for linux? Do they need to be adapted or is this a copy of the other one (I see the fuzzing test is copied at least from memory).

andrewstrohman: It's not a copy. These tests are specific to the arch's registers and `struct pt_regs`.

```
astrohma@lima-default:~/src/tetragon$ diff pkg/asm/assignment_linux_amd64_test.go pkg/asm/assignment_linux_arm64_test.go 
14,17c14,17
<       f.Add("rax=1")
<       f.Add("rbp=128%rax")
<       f.Add("rbp=0x20(%rsp)")
<       f.Add("rsp=-1372(%rbp)")
---
>       f.Add("x0=1")
>       f.Add("x29=128%x0")
>       f.Add("x29=0x20(%sp)")
>       f.Add("sp=-1372(%x29)")
33c33
<       ass, err = ParseAssignment("rax=1")
---
>       ass, err = ParseAssignment("x0=1")
37c37
<       assert.Equal(t, uint16(0x50), ass.Dst)
---
>       assert.Equal(t, uint16(0x0), ass.Dst)
40c40
<       ass, err = ParseAssignment("rbx=-1")
---
>       ass, err = ParseAssignment("x19=-1")
44c44
<       assert.Equal(t, uint16(0x28), ass.Dst)
---
>       assert.Equal(t, uint16(0x98), ass.Dst)
47c47
<       ass, err = ParseAssignment("rcx=0x123")
---
>       ass, err = ParseAssignment("x20=0x123")
51c51
<       assert.Equal(t, uint16(0x58), ass.Dst)
---
>       assert.Equal(t, uint16(0xa0), ass.Dst)
55c55
<       ass, err = ParseAssignment("rsp=%rax")
---
>       ass, err = ParseAssignment("sp=%x0")
58,59c58,59
<       assert.Equal(t, uint16(0x98), ass.Dst)
<       assert.Equal(t, uint16(0x50), ass.Src)
---
>       assert.Equal(t, uint16(0xf8), ass.Dst)
>       assert.Equal(t, uint16(0x0), ass.Src)
63c63
<       ass, err = ParseAssignment("rbp=128%rax")
---
>       ass, err = ParseAssignment("x29=128%x0")
66,67c66,67
<       assert.Equal(t, uint16(0x20), ass.Dst)
<       assert.Equal(t, uint16(0x50), ass.Src)
---
>       assert.Equal(t, uint16(0xe8), ass.Dst)
>       assert.Equal(t, uint16(0x0), ass.Src)
70c70
<       ass, err = ParseAssignment("rax=0x80%rax")
---
>       ass, err = ParseAssignment("x1=0x80%x1")
73,74c73,74
<       assert.Equal(t, uint16(0x50), ass.Dst)
<       assert.Equal(t, uint16(0x50), ass.Src)
---
>       assert.Equal(t, uint16(0x8), ass.Dst)
>       assert.Equal(t, uint16(0x8), ass.Src)
78c78
<       ass, err = ParseAssignment("rsp=-1372(%rbp)")
---
>       ass, err = ParseAssignment("sp=-1372(%x29)")
81,82c81,82
<       assert.Equal(t, uint16(0x98), ass.Dst)
<       assert.Equal(t, uint16(0x20), ass.Src)
---
>       assert.Equal(t, uint16(0xf8), ass.Dst)
>       assert.Equal(t, uint16(0xe8), ass.Src)
85c85
<       ass, err = ParseAssignment("rbp=0x20(%rsp)")
---
>       ass, err = ParseAssignment("x29=0x20(%sp)")
88,89c88,89
<       assert.Equal(t, uint16(0x20), ass.Dst)
<       assert.Equal(t, uint16(0x98), ass.Src)
---
>       assert.Equal(t, uint16(0xe8), ass.Dst)
>       assert.Equal(t, uint16(0xf8), ass.Src)
```

In every other case, I did merge into one file in hopes of deduplicating as much as possible by checking the arch with `runtime.GOARCH`. But in this case, I did not think the common code was enough to justify it.

That being said, I can adapt this to a single file if we still feel there is value in that.

mtardy: ah ok nevermind, I didn't look too deep :) looks good then!

olsajiri (approved):
looks great, thanks

pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -335,51 +377,95 @@ spec:
...
+		//   d65f03c0        ret
+		symbol = "test_1+12"
+		pathHook += `
+    symbols:
+    - "` + symbol + `"
+    data:
+    - index: 0
+      type: "int"
+      source: "pt_regs"
+      resolve: "w0"`
```
mtardy: Hey, is this expected? You `t.Skip` but still set a few variables after that? I imagine you can just remove this since we stop from there?

andrewstrohman: It's like this because I wanted to see if I could find a way to make it work. But after I found out that I couldn't get it work (for now) I just added the `t.Skip`.  I thought that it was worth keeping so that we could easily enable the test if/when we got register resolve working later.

I'll remove it to avoid confusion. We can just re-create it later if/when we pursue making resolving registers work on arm64.

mtardy: ah I see! It's up to you then, you can add a comment to explain that after the skip and keep (or not), as you wish :)

PR #4466, opened 2026-01-05
BPF: zero cleanup_key on exec

kevsecurity:
On process execution (execve) we send a message to user space with the details. In the case where a process hasn't previously forked (script execution, for example), we will already have an entry in our maps for this PID. We set the cleanup_key to this entry so that user space can harmonise the processes and set the referece counting flags accordingly.

If, however, the process isn't in our map, then we still send the message to user space but we fail to zero the cleanup_key, thereby sending the previous value that a previous execution had set (the heap memory is reused). This causes user space confusion, so this commit resets the cleanup_key before checking if the process is in the map. This should correct the reference counting issue where a process is marked with an additional "parent--" count.

### DIFF
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -271,6 +271,7 @@ event_execve(struct exec_ctx_struct *ctx)
 	if (parent) {
 		event->parent = parent->key;
 		update_mb_task(parent);
+		event->parent_flags = 0;
 	} else {
 		event_minimal_parent(event, task);
 	}
@@ -280,6 +281,7 @@ event_execve(struct exec_ctx_struct *ctx)
 	p->size_path = 0;
 	p->size_args = 0;
 	p->size_cwd = 0;
+	p->size_envs = 0;
 
 	/**
 	 * Per thread tracking rules TID == PID :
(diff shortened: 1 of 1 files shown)


bpf/process/bpf_execve_event.c
```diff
@@ -372,6 +372,8 @@ execve_send(struct exec_ctx_struct *ctx __arg_ctx)
 
 	pid = (get_current_pid_tgid() >> 32);
 
+	// Zero the cleanup key to prevent user space confusion.
+	event->cleanup_key = (struct msg_execve_key){ 0 };
```
olsajiri: nice catch, would it be better to do this in event_execve with other init code?

mtardy (approved):
Ah yeah, that same issue happened to me already, nice! If it does fix stuff, could we backport this? Would there be a way this is actually systematically testable and we could write a regression test for this?

Also just asking out loud to double check, is there anything else that would need to be zeroed out?

bpf/process/bpf_execve_event.c
kevsecurity: Yes, I'll do that.

kevsecurity:
> Ah yeah, that same issue happened to me already, nice! If it does fix stuff, could we backport this? Would there be a way this is actually systematically testable and we could write a regression test for this?
> 
> Also just asking out loud to double check, is there anything else that would need to be zeroed out?

I can take a look.

PR #4489, opened 2026-01-08
tetragon: add support to preload user strings

olsajiri:
Adding uprobe preload program to allow loading data in sleepable context. The preload program is executed before the main program loads the data and passes it via map to main program.
    
So far we allow only string type data to be preloaded.

### DIFF
--- a/pkg/sensors/tracing/uprobe_reg_test.go
+++ b/pkg/sensors/tracing/uprobe_reg_test.go
@@ -645,3 +645,104 @@ func TestUprobePtRegsDataMatch(t *testing.T) {
 func TestUprobePtRegsDataNotMatch(t *testing.T) {
 	testUprobePtRegsMatch(t, 10, true)
 }
+
+func testUprobePtRegsPreload(t *testing.T, multi bool) {
+	if !bpf.HasKfunc("bpf_copy_from_user_str") {
+		t.Skip("skipping")
+	}
+
+	testBinary := testutils.RepoRootPath("contrib/tester-progs/regs-override")
+
+	disableUprobeMulti := ""
+
+	if !multi {
+		disableUprobeMulti = `
+  options:
+    - name: "disable-uprobe-multi"
+      value: "1"`
+	}
+
+	// Put uprobe in test_3 function at:
+	//
+	//      static const char *test_3_string = "test_3_string_CASE";
+	//
+	//      "push   %%rbp\n"          /* +0  55                            */
+	//      "mov    %%rsp, %%rbp\n"   /* +1  48 89 e5                      */
+	//      "mov    %[str], %%rdi\n"  /* +4  48 8b 3d 96 2e 00 00          */
+	//      "pop    %%rbp\n"          /* +11 5d                            */
+	// -->  "mov    $0x0,%%rax\n"     /* +12 48 c7 c0 00 00 00 00          */
+	//      "mov    $0xff,%%rax\n"    /* +19 48 c7 c0 ff 00 00 00          */
+	//      "ret\n"                   /* +26 c3                            */
+	//      :
+	//      : [str] "m" (test_3_string)
+	//
+	// Make sure we retrieve data with eax value (1) as int argument
+	// and match the expected value via matchData.
+
+	pathHook := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "uprobe"
+spec: ` + disableUprobeMulti + `
+  uprobes:
+  - path: "` + testBinary + `"
+    symbols:
+    - "test_3+12"
+    data:
+    - index: 0
+      type: "string"
+      source: "pt_regs"
+      resolve: "rdi"
+`
+
+	pathConfigHook := []byte(pathHook)
+	err := os.WriteFile(testConfigFile, pathConfigHook, 0644)
+	if err != nil {
+		t.Fatalf("writeFile(%s): err %s", testConfigFile, err)
+	}
+
+	upChecker := ec.NewProcessUprobeChecker("UPROBE_DATA_MATCH").
+		WithProcess(ec.NewProcessChecker().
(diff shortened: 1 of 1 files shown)


bpf/process/uprobe_preload.h
```diff
@@ -0,0 +1,115 @@
...
+	__type(key, __u64);
+	__type(value, struct preload_data);
+} sleepable_preload SEC(".maps");
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, 1);
+	__type(key, __u32);
+	__type(value, struct preload_data);
+} sleepable_preload_heap SEC(".maps");
```
andrewstrohman: I don't see that this is used anywhere. I do see that `process_call_heap` was used.

bpf/process/uprobe_preload.h
```diff
@@ -0,0 +1,115 @@
...
+	__u64 id = get_current_pid_tgid();
+	struct preload_data *data;
+	__u32 zero = 0;
+
+	data = map_lookup_elem(&process_call_heap, &zero);
+	if (!data)
+		return 0;
+
+	bpf_copy_from_user_str(data, sizeof(*data), (const void *)val, 0);
+	map_update_elem(&sleepable_preload, &id, data, BPF_ANY);
```
andrewstrohman: I think that this is prone to a race. If after line 49, the context is preempted due to the fact that we called `bpf_copy_from_user*` (which can trigger a page fault) another context can overwrite what's stored in `data` before line 50 is executed.

I'm not sure if we can safely use `BPF_MAP_TYPE_PERCPU_ARRAY` for a heap in this case.

andrewstrohman:
I think that it would be helpful to explicitly call out the motivation here. I _think_ we are doing this in order to be able to page in userspace memory for the cases where we would encounter a page fault in order to access memory.  As this problem is a common for reading strings, they are the first type wired to use this. Is this an accurate understanding, or am I missing the point?

andrewstrohman:
Why is this for x86_64 only? Recently, I've been experimenting with [reading string args](https://github.com/cilium/tetragon/pull/4546). From my limited testing, I saw that page faults only happened on x86_64. (It didn't seem to happen on arm64). Will arm64 ever need this functionality?

andrewstrohman:
This PR is specific to data source pt_regs. I think it would be nice to extend similar logic for string arg reading as well.

Recently, I experimented with trying to read string arguments from uprobes using the `args` definition (as opposed to `data`). I noticed on x86_64  that I couldn't do this because the string was not already paged into memory. I think this approach would fix that problem for me. I understand that we can use this patch to read string argument via the `data` definition, but I think it would be nice to do it with the `args` syntax as well. This way, they can just read string args using the same syntax that they would for kprobe hooks.

What do you think about me trying to extend this in that way?

bpf/process/uprobe_preload.h
olsajiri: nah I messed up the setup.. copy paste error, thanks

bpf/process/uprobe_preload.h
olsajiri: yes, you're right.. I switched to using just the hash based per process/thread, thanks

olsajiri:
> I think that it would be helpful to explicitly call out the motivation here. I _think_ we are doing this in order to be able to page in userspace memory for the cases where we would encounter a page fault in order to access memory. As this problem is a common for reading strings, they are the first type wired to use this. Is this an accurate understanding, or am I missing the point?

I tried to describe that in the PR description.. so ATM we need to delegate reading from user space to sleepable program, because now we can't make the main generic sensor sleepable ... there's pending kernel change for that, so we might be able to do that in the future, but so far we need the preload bit

olsajiri:
> Why is this for x86_64 only? Recently, I've been experimenting with [reading string args](https://github.com/cilium/tetragon/pull/4546). From my limited testing, I saw that page faults only happened on x86_64. (It didn't seem to happen on arm64). Will arm64 ever need this functionality?

I'm not sure about arm, that's why I made it x86 specific, because it's problem in there AFAIK.. if you think it's needed for arm as well, it should be easy to switch on

olsajiri:
> This PR is specific to data source pt_regs. I think it would be nice to extend similar logic for string arg reading as well.
> 
> Recently, I experimented with trying to read string arguments from uprobes using the `args` definition (as opposed to `data`). I noticed on x86_64 that I couldn't do this because the string was not already paged into memory. I think this approach would fix that problem for me. I understand that we can use this patch to read string argument via the `data` definition, but I think it would be nice to do it with the `args` syntax as well. This way, they can just read string args using the same syntax that they would for kprobe hooks.
> 
> What do you think about me trying to extend this in that way?

sure, I added just string so far but I think we will need other types soon

as I said earlier there's pending kernel change that will allow uprobe sensor to be sleepable, but we will still need preload for old kernels ... and we should somehow make those 2 approaches co-exists together

andrewstrohman (approved):
nice!

andrewstrohman:
>  so ATM we need to delegate reading from user space to sleepable program, because now we can't make the main generic sensor sleepable 

I'm trying to confirm the motivation for making reading from user space sleepable. I think it's so we can access any memory within the user space program. Otherwise, we would not be able to access memory that hasn't been paged in before the uprobe hook point. Is this the motivation?

olsajiri:
> the motivation for making reading from user space sleepable. I think it's so we can access any memory within the user space program. Otherwise, we would not be able to access memory that hasn't been paged in before the uprobe hook point

yep, exactly

mtardy:
> I'm trying to confirm the motivation for making reading from user space sleepable. I think it's so we can access any memory within the user space program. Otherwise, we would not be able to access memory that hasn't been paged in before the uprobe hook point. Is this the motivation?

this recap would have been awesome from the start to understand the motivation directly 😅

PR #4494, opened 2026-01-09
fix(pkg/sensors): fixed {k,u}retprobe args merge helper.

### LINKED ISSUE #4488: having returnCopy makes some args missing from tetragon events

Having `returnCopy: true` makes the following arg disappear from tetragon events.

To reproduce:

### Happy path:
```
kind: TracingPolicy
metadata:
  name: "write-content"
spec:
  kprobes:
  - call: "vfs_write"
    syscall: false
    args:
    - index: 1
      type: "char_buf"
      label: kfunc_char_buf
      sizeArgIndex: 3
      returnCopy: false
    - index: 2
      type: "size_t"
      label: kfunc_size_t
    selectors:
    - matchNamespaces:
      - namespace: Pid
        operator: NotIn
        values:
        - "host_ns"
  - call: "sys_write"
    syscall: true
    args:
    - index: 1
      type: "char_buf"
      label: sys_char_buf
      returnCopy: false
      sizeArgIndex: 3
    - index: 2
      type: "size_t"
      label: sys_size_t
    selectors:
    - matchNamespaces:
      - namespace: Pid
        operator: NotIn
        values:
        - "host_ns"
```

then start tetragon, start tetra, invoke the kfunc and syscall: `unshare -Umuinpf --mount-proc sh -c 'echo foobar'`,  we get the correct tetragon events

```
"function_name":"vfs_write","args":[{"bytes_arg":"Zm9vYmFyCg==","label":"kfunc_char_buf"},{"size_arg":"7","label":"kfunc_size_t"}]...

"function_name":"__x64_sys_write","args":[{"bytes_arg":"Zm9vYmFyCg==","label":"sys_char_buf"},{"size_arg":"7","label":"sys_size_t"}]...
```
 
### Error Path:

But if we add the `retunCopy: true`, the sys_size_t arg is missing:

```
kind: TracingPolicy
metadata:
  name: "write-content"
spec:
  kprobes:
  - call: "vfs_write"
    syscall: false
    args:
    - index: 1
      type: "char_buf"
      label: kfunc_char_buf
      sizeArgIndex: 3
      returnCopy: true
    - index: 2
      type: "size_t"
      label: kfunc_size_t
    selectors:
    - matchNamespaces:
      - namespace: Pid
        operator: NotIn
        values:
        - "host_ns"
  - call: "sys_write"
    syscall: true
    args:
    - index: 1
      type: "char_buf"
      label: sys_char_buf
      returnCopy: true
      sizeArgIndex: 3
    - index: 2
      type: "size_t"
      label: sys_size_t
    selectors:
    - matchNamespaces:
      - namespace: Pid
        operator: NotIn
        values:
        - "host_ns"
```

```
"function_name":"vfs_write","args":[{"bytes_arg":"","label":"kfunc_char_buf"},{"bytes_arg":"Zm9vYmFyCg==","label":"kfunc_char_buf"}] ...

"function_name":"__x64_sys_write","args":[{"bytes_arg":"","label":"sys_char_buf"},{"bytes_arg":"Zm9vYmFyCg==","label":"sys_char_buf"}] ...
```

### Tetragon Version

I am building tetra, tetragon at head

### Kernel Version

6.16.12-1rodete2-amd64

### Kubernetes Version

NA, running directly on linux

### Bugtool

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

FedeDP:
Fixes #4488

### Description
Since the arg `Index` refers to the syscall signature index, and not the index in the args slice, we should lookup the correct enterArg with same index, instead of just overriding the `[index]` slice elem.

At first, i thought the bug was introduced by my refactor in: https://github.com/cilium/tetragon/pull/4193, but that part of the code was already in place.

### Changelog

```release-note
fix(pkg/sensors): fixed {k,u}retprobe args merge helper.
```

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -605,6 +605,48 @@ spec:
 	runKprobeObjectRead(t, readHook, checker, fd, fd2)
 }
 
+// Differently from other tests using returnCopy,
+// this one skips index 0 element to avoid future regressions
+// like https://github.com/cilium/tetragon/issues/4488.
+func TestKprobeObjectReturnCopy(t *testing.T) {
+	fd, fd2, _ := createTestFile(t)
+	pidStr := strconv.Itoa(int(observertesthelper.GetMyPid()))
+	readHook := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "sys-read"
+spec:
+  kprobes:
+  - call: "sys_read"
+    syscall: true
+    args:
+    - index: 1
+      type: "char_buf"
+      returnCopy: true
+    - index: 2
+      type: "size_t"
+    selectors:
+    - matchPIDs:
+      - operator: In
+        followForks: true
+        values:
+        - ` + pidStr + `
+`
+
+	kpChecker := ec.NewProcessKprobeChecker("").
+		WithFunctionName(sm.Full(arch.AddSyscallPrefixTestHelper(t, "sys_read"))).
+		WithArgs(ec.NewKprobeArgumentListMatcher().
+			WithOperator(lc.Ordered).
+			WithValues(
+				ec.NewKprobeArgumentChecker().WithBytesArg(bc.Full([]byte("hello world"))),
+				ec.NewKprobeArgumentChecker().WithSizeArg(100),
+			))
+	checker := ec.NewUnorderedEventChecker(kpChecker)
+
+	runKprobeObjectRead(t, readHook, checker, fd, fd2)
+}
+
 // sys_openat trace
 func getOpenatChecker(t *testing.T, dir string) ec.MultiEventChecker {
 	kpChecker := ec.NewProcessKprobeChecker("").


mtardy (commented):
Thanks a lot for looking into this, any chance you could add a regression test from the issue examples? :)

FedeDP:
Yep done right now :)

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -605,6 +605,45 @@ spec:
...
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "sys-read"
+spec:
+  kprobes:
+  - call: "sys_read"
+    syscall: true
+    args:
+    - index: 1
```
olsajiri, line 624: so the other returnCopy tests work because those policies define all read arguments? could you plz make the note in the comment about that

FedeDP: Exactly!
> could you plz make the note in the comment about that

100%!

mtardy:
Does this fix need backporting?

FedeDP:
Given the issue has been there forever, i don't think a backport is worth it (it's not like this is a bug introduced in latest release after all). WDYT?

mtardy:
> Given the issue has been there forever, i don't think a backport is worth it (it's not like this is a bug introduced in latest release after all). WDYT?

ah if the issue was already there I think it would be nice to have it patched on supported versions. Even if we didn't introduce them lately, we still bug fix the supported versions when we can.

PR #4500, opened 2026-01-10
metrics: clean up policy metrics when tracing policies are deleted

### LINKED ISSUE #2256: Delete metrics with policy label on policy delete
`tetragon_policy_events_total` metric has a `policy` label. When a TracingPolicy is deleted, the corresponding metrics are still exposed by Tetragon, leading to an overhead (usually not big, as policies are not churned frequently, but might be significant in some cases).

There was a similar problem with metrics that have a `pod` label. Now when a pod is deleted, Tetragon deletes its metrics from the registry (see https://github.com/cilium/tetragon/blob/main/pkg/metrics/metricwithpod.go).

We should add a similar handler for policy delete events. One small gotcha is that at the moment metrics don't distinguish between clusterwide and namespaced policies, so we should first add another label to distinguish between different policy kinds.

gr3chk:
Policy-labeled metric series are currently kept after a TracingPolicy is deleted, which can cause stale metrics to remain exported.

Register tracing-policy metrics for cleanup and removes their series from the registry when a policy is deleted, regardless of how the policy was created or removed.

Fixes: #2256

### Description

Ensure policy-labeled metric series are properly cleaned up when a TracingPolicy is deleted, preventing stale metrics from remaining exported.

### Changelog

```release-note
Fix cleanup of policy-labeled metric series when tracing policies are deleted.
```

### DIFF
--- a/pkg/metrics/counter.go
+++ b/pkg/metrics/counter.go
@@ -9,6 +9,12 @@ import (
 
 type initCounterFunc func(*prometheus.CounterVec)
 
+func RegisterCounterVecForPodCleanup(metric *prometheus.CounterVec) {
+	metricsWithPodMutex.Lock()
+	metricsWithPod = append(metricsWithPod, metric.MetricVec)
+	metricsWithPodMutex.Unlock()
+}
+
 // NewCounterVecWithPod is a wrapper around prometheus.NewCounterVec that also
 // registers the metric to be cleaned up when a pod is deleted.
 //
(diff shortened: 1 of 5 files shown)


kkourt (commented):
Hi @gr3chk,

Apologies for the delayed reply, The PR makes sense to me, but, unfortunately, it needs to be rebased. Are you interested in continuing your work on it?

PR #4501, opened 2026-01-10
refactor: Replace ebpf.Map with program.Map for policy filter

sayboras:
Fixes 

### Changelog

```release-note
refactor: Replace ebpf.Map with program.Map for policy filter
```

### DIFF
--- a/pkg/policyfilter/map.go
+++ b/pkg/policyfilter/map.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/cilium/tetragon/pkg/bpf"
 	"github.com/cilium/tetragon/pkg/config"
+	"github.com/cilium/tetragon/pkg/sensors/program"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


olsajiri:
I recall mentioning this but forgot the motivation for this change.. and seeing the patches it does not look like we'll be in a better state.. was it the maxEntries setup we wanted this for? @kkourt I'll try to check some more

mtardy:
> I recall mentioning this but forgot the motivation for this change.. and seeing the patches it does not look like we'll be in a better state.. was it the maxEntries setup we wanted this for? @kkourt I'll try to check some more

this is the motivation https://github.com/cilium/tetragon/pull/4331#discussion_r2698313480

olsajiri:
> > I recall mentioning this but forgot the motivation for this change.. and seeing the patches it does not look like we'll be in a better state.. was it the maxEntries setup we wanted this for? @kkourt I'll try to check some more
> 
> this is the motivation [#4331 (comment)](https://github.com/cilium/tetragon/pull/4331#discussion_r2698313480)

ah ok, thanks.. so I think we need master builder of that map so the sensors could use it and declare their map with MapUserFrom and that way we will have the max entries set on just master object

FedeDP (approved):
LGTM

PR #4504, opened 2026-01-12
celbpf: evaluate CEL expressions in bpf

kkourt:
This is a first PR to compile CEL expressions in BPF in the context of Tetragon policies.

This is heavily based on work done by @YutaroHayakawa.

The PR creates a simple CEL to BPF compiler and uses it to evaluate CEL expressions in TracingPolicies.
For now it supports:
- 64- and 32-bit signed and unsigned integers
- addition and subtraction
- comparison operators (==, !=, >, >=, <, <=)
- logical operators (&&, ||, !)

For now, this only works on kprobes (but not on multi-kprobe).

followup work: https://github.com/cilium/tetragon/issues/4661

Dependencies:
 - https://github.com/cilium/tetragon/pull/4526

### DIFF
--- a/bpf/process/cel_expr.h
+++ b/bpf/process/cel_expr.h
@@ -8,7 +8,7 @@
  * The code for the cel_expr_N functions will be generated and linked when the policy is loaded
  */
 
-#if defined(GENERIC_KPROBE) && !defined(__MULTI_KPROBE)
+#if defined(__LARGE_BPF_PROG) && (defined(GENERIC_KPROBE) && !defined(__MULTI_KPROBE))
 int cel_expr_0(long *argsoff, char *args);
 int cel_expr_1(long *argsoff, char *args);
 int cel_expr_2(long *argsoff, char *args);
@@ -18,8 +18,7 @@ int cel_expr_5(long *argsoff, char *args);
 int cel_expr_6(long *argsoff, char *args);
 int cel_expr_7(long *argsoff, char *args);
 
-FUNC_LOCAL int
-cel_expr(int id, long *argsoff, char *args)
+static __attribute__((noinline)) __attribute__((__unused__)) int cel_expr(int id, long *argsoff, char *args)
 {
 	switch (id) {
 	case 0:
@@ -44,7 +43,7 @@ cel_expr(int id, long *argsoff, char *args)
 	}
 }
 #else
-FUNC_LOCAL int
+FUNC_INLINE int
 cel_expr(int id, long *argsoff, char *args)
 {
 	return 0;
--- a/pkg/bpf/detect_linux.go
+++ b/pkg/bpf/detect_linux.go
@@ -56,6 +56,7 @@ var (
 	auditLoginuid          Feature
 	uprobeRegsChange       Feature
 	kfuncs                 FeatureKfuncs
+	mixBpfAndTailCalls     Feature
 )
 
 func HasOverrideHelper() bool {
@@ -610,12 +611,82 @@ func HasKfunc(name string) bool {
 	return detected
 }
 
+func detectMixBpfAndTailCalls() bool {
+	// create a tail call map
+	tcMap, err := ebpf.NewMap(&ebpf.MapSpec{
+		Type:       ebpf.ProgramArray,
+		KeySize:    4,
+		ValueSize:  4,
+		MaxEntries: 1,
+	})
+	if err != nil {
+		fmt.Printf("#0 err=%+v\n", err)
+		return false
+	}
+	defer tcMap.Close()
+
+	// create a program
+	tcProg, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Type: ebpf.RawTracepoint,
+		Instructions: asm.Instructions{
+			asm.Mov.Imm(asm.R0, 0),
+			asm.Return(),
+		},
+		License: "Dual BSD/GPL",
+	})
+	if err != nil {
+		return false
+	}
+	defer tcProg.Close()
+
+	// insert the program into the map
+	zero := uint32(0)
+	fd := int32(tcProg.FD())
+	err = tcMap.Update(&zero, &fd, 0)
+	if err != nil {
+		return false
+	}
+
+	// create a program that does both a bpf-to-bpf call and a tail call
+	prog, err := ebpf.NewProgram(&ebpf.ProgramSpec{
+		Instructions: asm.Instructions{
+			asm.Mov.Reg(asm.R6, asm.R1),
+			asm.Call.Label("myfn"),
+			asm.Mov.Reg(asm.R1, asm.R6),
+			asm.LoadMapPtr(asm.R2, tcMap.FD()),
+			asm.Mov.Imm(asm.R3, 0),
+			asm.FnTailCall.Call(),
+			asm.Mov.Imm(asm.R0, 0),
+			asm.Return(),
+			asm.Mov.Imm(asm.R0, 0).WithSymbol("myfn"),
+			asm.Return(),
+		},
+		Type:    ebpf.RawTracepoint,
+		License: "Dual BSD/GPL",
+	})
+	if err != nil {
+		return false
+	}
+	defer prog.Close()
+	return true
+}
+
+func detectMixBpfAndTailCallsOnce() {
+	mixBpfAndTailCalls.init.Do(func() {
+		mixBpfAndTailCalls.detected = detectMixBpfAndTailCalls()
+	})
+}
+
+func DetectMixBpfAndTailCalls() bool {
+	detectMixBpfAndTailCallsOnce()
+	return mixBpfAndTailCalls.detected
+}
+
 func LogFeatures() string {
 	// once we have detected all features, flush the BTF spec
 	// we cache all values so calling again a Has* function will
 	// not load the BTF again
 	defer ebtf.FlushKernelSpec()
-
 	probeStr := func(p *FeatureProbe) string {
 		return p.Name + ": " + strconv.FormatBool(p.Fn())
 	}
@@ -658,4 +729,5 @@ var FeatureProbes = []FeatureProbe{
 	{"audit_loginuid", HasAuditLoginuid},
 	{"probe_write_user", HasProbeWriteUserHelper},
 	{"uprobe_regs_change", HasUprobeRegsChange},
+	{"mix_bpf_and_tail_calls", DetectMixBpfAndTailCalls},
 }
--- a/pkg/celbpf/celbpf.go
+++ b/pkg/celbpf/celbpf.go
@@ -13,6 +13,9 @@ import (
 	cgChecker "github.com/google/cel-go/checker"
 	cgCommon "github.com/google/cel-go/common"
 	cgParser "github.com/google/cel-go/parser"
+
+	"github.com/cilium/tetragon/pkg/bpf"
+	"github.com/cilium/tetragon/pkg/config"
 )
 
 type Env struct{}
@@ -20,6 +23,19 @@ type Env struct{}
 type ExprArg struct {
 	GenTy     int
 	ArgOffset int
+}
+
+// EnabledInBPF returns true if celbpf is supported in the BPF programs
+func EnabledInBPF() bool {
+	return config.EnableLargeProgs()
+}
+
+// Supported returns true if celbpf is supported at runtime
+// NB: The cel_expr code uses bpf-to-bpf calls, so we need to detect whether mixing bpf-to-bpf calls
+// and tail-calls is allowed.
+// See: https://lore.kernel.org/bpf/20200829231925.GB31692@ranger.igk.intel.com/T/
+func Supported() bool {
+	return bpf.DetectMixBpfAndTailCalls()
 }
 
 func Compile(celExpr string, args []ExprArg, labelPrefix string) (asm.Instructions, error) {
--- a/pkg/celbpf/celbpf_test.go
+++ b/pkg/celbpf/celbpf_test.go
@@ -123,6 +123,9 @@ func btfTestArgExprFnTy(fnName string) *btf.Func {
 }
 
 func TestArgExprs(t *testing.T) {
+	if !Supported() {
+		t.Skip()
+	}
 	testCase := []struct {
 		expr     string
 		ret      uint32
--- a/pkg/selectors/celexpr.go
+++ b/pkg/selectors/celexpr.go
@@ -70,8 +70,11 @@ func parseMatchCelExpr(
 	arg *v1alpha1.ArgSelector,
 	sig []v1alpha1.KProbeArg,
 ) error {
-	exprs, err := addMatchCelExpr(k.celExprFunctions, arg, sig)
+	if !celbpf.Supported() {
+		return errors.New("celbpf not supported in this kernel")
+	}
 
+	exprs, err := addMatchCelExpr(k.celExprFunctions, arg, sig)
 	if err != nil {
 		return err
 	}
@@ -86,7 +89,37 @@ func parseMatchCelExpr(
 	return nil
 }
 
+func removeCelFunction(prog *ebpf.ProgramSpec) {
+	newInstructions := make(asm.Instructions, 0, len(prog.Instructions))
+	skip := false
+	for _, ins := range prog.Instructions {
+		sym := ins.Symbol()
+		if sym == "cel_expr" {
+			skip = true
+			// skip cel_expr()
+			continue
+		} else if skip && sym == "" {
+			// continue skipping cel_expr()
+			continue
+		}
+
+		// found new symbol (new func), stop skipping if we were
+		skip = false
+
+		if ins.IsFunctionCall() && ins.Reference() == "cel_expr" {
+			ins = asm.Mov.Imm(asm.R0, 0)
+		}
+		newInstructions = append(newInstructions, ins)
+	}
+	prog.Instructions = newInstructions
+}
+
 func (cefs CelExprFunctions) RewriteProg(prog *ebpf.ProgramSpec) error {
+	if !celbpf.Supported() {
+		removeCelFunction(prog)
+		return nil
+	}
+
 	// add the generated functions as new instructions
 	for _, insns := range cefs.functions() {
 		prog.Instructions = append(prog.Instructions, insns...)
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -29,6 +29,7 @@ import (
 	"github.com/cilium/tetragon/pkg/asm"
 	"github.com/cilium/tetragon/pkg/bpf"
 	"github.com/cilium/tetragon/pkg/btf"
+	"github.com/cilium/tetragon/pkg/celbpf"
 	"github.com/cilium/tetragon/pkg/cgtracker"
 	"github.com/cilium/tetragon/pkg/config"
 	"github.com/cilium/tetragon/pkg/eventhandler"
@@ -1225,8 +1226,9 @@ func loadSingleKprobeSensor(id idtable.EntryID, bpfDir string, load *program.Pro
 
 	rewriteProg := make(map[string]func(prog *ebpf.ProgramSpec) error)
 	if entry := gk.loadArgs.selectors.entry; entry != nil {
-		// functions generated for CEL expressions
-		rewriteProg["generic_kprobe_filter_arg"] = entry.CelExprFunctions().RewriteProg
+		if celbpf.EnabledInBPF() {
+			rewriteProg["generic_kprobe_filter_arg"] = entry.CelExprFunctions().RewriteProg
+		}
 	}
 	load.RewriteProg = rewriteProg
 
--- a/contrib/verify/verify_test.go
+++ b/contrib/verify/verify_test.go
@@ -116,11 +116,13 @@ func TestVerifyTetragonPrograms(t *testing.T) {
 		}
 
 		if strings.HasPrefix(fileName, "bpf_generic_kprobe") {
-			for _, prog := range spec.Programs {
-				var exprs selectors.CelExprFunctions
-				if prog.Name == "generic_kprobe_filter_arg" {
-					err := exprs.RewriteProg(prog)
-					require.NoError(t, err, "failed to rewrite program for empty CEL expressions")
+			if fileName != "bpf_generic_kprobe.o" { // 4.19 version does not need to be rewritten
+				for _, prog := range spec.Programs {
+					var exprs selectors.CelExprFunctions
+					if prog.Name == "generic_kprobe_filter_arg" {
+						err := exprs.RewriteProg(prog)
+						require.NoError(t, err, "failed to rewrite program for empty CEL expressions")
+					}
 				}
 			}
 		}
--- a/pkg/bpf/detect_windows.go
+++ b/pkg/bpf/detect_windows.go
@@ -41,8 +41,12 @@ func HasBatchAPI() bool {
 	return false
 }
 
+func DetectMixBpfAndTailCalls() bool {
+	return false
+}
+
 func LogFeatures() string {
-	return fmt.Sprintf("override_return: %t, buildid: %t, fmodret: %t, fmodret_syscall: %t, signal: %t, large: %t, link_pin: %t",
+	return fmt.Sprintf("override_return: %t, buildid: %t, fmodret: %t, fmodret_syscall: %t, signal: %t, large: %t, link_pin: %t mix_bpf_and_tail_calls: %t",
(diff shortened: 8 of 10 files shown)


olsajiri:
nice, I was surprised how easy the program patching was with cilium/ebpf, really cool
could you split the first 3 fixes in separate PR?

kkourt:
> nice, I was surprised how easy the program patching was with cilium/ebpf, really cool could you split the first 3 fixes in separate PR?

Yap, makes sense: https://github.com/cilium/tetragon/pull/4526

pkg/celbpf/codegen.go
```diff
@@ -29,15 +42,45 @@ func (g *codeGenerator) emitRaw(insts ...asm.Instruction) {
...
+	imm := int32(0)
 	if val {
 		imm = 1
 	}
 	g.stackTop -= 8
-	g.emitRaw(asm.StoreImm(asm.R10, g.stackTop, imm, asm.DWord))
+	g.emitRaw(
+		asm.Mov.Imm(scratchRegs[0], imm),
+		asm.StoreMem(asm.R10, g.stackTop, scratchRegs[0], asm.DWord),
+	)
```
YutaroHayakawa: Why did you change this?

pkg/celbpf/codegen.go
```diff
@@ -29,15 +42,45 @@ func (g *codeGenerator) emitRaw(insts ...asm.Instruction) {
...
+func (g *codeGenerator) emitBranchEquals(reg0 asm.Register, reg1 asm.Register) {
+	// make space to push a boolean (8 bytes)
+	g.stackTop -= 8
+	label := g.generateLabel()
+	g.emitRaw(
+		asm.JEq.Reg(reg0, reg1, label),
+		asm.LoadImm(asm.R3, 0, asm.DWord),
+		asm.Instruction{OpCode: asm.Ja.Op(asm.ImmSource), Offset: 2},
+		asm.LoadImm(asm.R3, 1, asm.DWord).WithSymbol(label),
+		asm.StoreMem(asm.R10, g.stackTop, asm.R3, asm.DWord),
```
YutaroHayakawa: This hard-coded R3 may become a problem in the future (maybe this is coming from my original code, sorry about that). I think it's safer to pass free scratch register through argument.

pkg/celbpf/compiler.go
```diff
@@ -35,14 +36,80 @@ func (c *compiler) compileLiteral(lit cgRef.Val) error {
...
+	callArgs := call.Args()
+	argTypes := make([]*cgTypes.Type, 0, len(callArgs))
+	for _, arg := range callArgs {
+		ty := c.ast.GetType(arg.ID())
+		err := c.compileExpr(arg)
+		if err != nil {
+			return err
+		}
+		argTypes = append(argTypes, ty)
+	}
```
YutaroHayakawa: Just trying to remember, the number of argument is already validated by type checker before compilation right? So, at this point, the number of the argument is guaranteed to be 2 for `==`?

pkg/celbpf/compiler.go
```diff
@@ -35,14 +36,80 @@ func (c *compiler) compileLiteral(lit cgRef.Val) error {
...
+		return errors.New("expression Kind 'IdentKind' not supported")
+	case cgAst.ListKind:
+		return errors.New("expression Kind 'ListKind' not supported")
+	case cgAst.MapKind:
+		return errors.New("expression Kind 'MapKind' not supported")
+	case cgAst.SelectKind:
+		return errors.New("expression Kind 'SelectKind' not supported")
+	case cgAst.StructKind:
+		return errors.New("expression Kind 'StructKind' not supported")
+
```
YutaroHayakawa, line 207: I guess we don't need these as we handle unsupported expression error below?

pkg/celbpf/env.go
```diff
@@ -10,8 +10,36 @@ import (
...
 
+func checkerAddFunctions(env *cgChecker.Env) error {
+	paramA := cgTypes.NewTypeParamType("A")
+
+	fnsOpts := []struct {
+		name string
+		opts []cgDecls.FunctionOpt
+	}{
+		{name: cgOperators.Equals, opts: []cgDecls.FunctionOpt{
+			cgDecls.Overload(cgOverloads.Equals, []*cgTypes.Type{paramA, paramA}, cgTypes.BoolType),
```
YutaroHayakawa, line 34: What do you think about enumerating all possible type combinations here instead of using parameterized type here?

```golang
			cgDecls.Overload(cgOverloads.Equals, []*cgTypes.Type{cgTypes.IntType, cgTypes.IntType}, cgTypes.BoolType),
			cgDecls.Overload(cgOverloads.Equals, []*cgTypes.Type{cgTypes.UintType, cgTypes.UintType}, cgTypes.BoolType),
			cgDecls.Overload(cgOverloads.Equals, []*cgTypes.Type{cgTypes.BoolType, cgTypes.BoolType}, cgTypes.BoolType),
```

It's annoying to add types to here every time we add the new type support, but at least, we don't accidentally support the operation we never intended to support (in fact, we forgot to cover bool == bool in the test case in this commit). Thinking about this become the user-facing API, being subjective about supporting specific operation for specific type feels better to me.

pkg/celbpf/compiler.go
```diff
@@ -35,14 +36,80 @@ func (c *compiler) compileLiteral(lit cgRef.Val) error {
...
+		}
+		argTypes = append(argTypes, ty)
+	}
+
+	// pop arguments (reverse order)
+	for j := range callArgs {
+		i := len(callArgs) - j - 1
+		ty := argTypes[i]
+		switch ty.TypeName() {
+		case "int", "uint":
```
YutaroHayakawa, line 150: I think we may have a `bool` case here right (like `true == false` or `(1 == 2) == (3 == 4)`)? At least type checker doesn't enforce that this must be `int` or `uint`.

pkg/celbpf/celbpf_test.go
```diff
@@ -22,18 +22,23 @@ func TestExprs(t *testing.T) {
 	}{
 		{"true", 1},
 		{"false", 0},
+		{"10 == 10", 1},
+		{"10 == 5", 0},
+		{"10u == 10u", 1},
+		{"10u == 5u", 0},
```
YutaroHayakawa: Needs test cases for bool == bool case.

pkg/celbpf/compiler.go
```diff
@@ -107,6 +114,29 @@ func (c *compiler) compileCall(expr cgAst.Expr) error {
...
+
+func (c *compiler) compileIdent(s string) error {
+	if strings.HasPrefix(s, "arg") {
+		idx, err := strconv.Atoi(s[3:])
+		if err != nil {
+			return fmt.Errorf("invalid argument (%s): %w", s, err)
+		}
+		return c.compileArg(idx)
+	}
+	return fmt.Errorf("ident %q unknown", s)
```
YutaroHayakawa: I think this can be caught in the checker, so worth adding `BUG:` prefix?

pkg/celbpf/codegen.go
```diff
@@ -132,3 +135,36 @@ func (g *codeGenerator) emitU32(reg asm.Register, regTy *cgTypes.Type) error {
...
+		g.stackTop -= 8
+		g.emitRaw(
+			// r3 = *(u64 *)(argArgsOff + idx)
+			asm.LoadMem(asm.R3, argArgsOff, int16(argOffset*8), asm.DWord),
+			asm.And.Imm(asm.R3, 0x7ff),
+			// r3 += args
+			asm.Add.Reg(asm.R3, argArgs),
+			// r4 = *(u64 *)(r3)
+			asm.LoadMem(asm.R4, asm.R3, 0, asm.DWord),
+			asm.StoreMem(asm.R10, g.stackTop, asm.R4, asm.DWord),
```
YutaroHayakawa: Same here for the register management. Maybe safer to let compiler to provide free scratch registers instead of hard-coding R3 and R4.

pkg/celbpf/codegen.go
```diff
@@ -132,3 +135,36 @@ func (g *codeGenerator) emitU32(reg asm.Register, regTy *cgTypes.Type) error {
...
 	return nil
 }
+
+func (g *codeGenerator) pushArg(argTy *cgTypes.Type, argOffset int) error {
+	switch argTy {
+	case u32Ty, s32Ty:
+		g.stackTop -= 8
+		g.emitRaw(
+			// r3 = *(u64 *)(argArgsOff + idx)
+			asm.LoadMem(asm.R3, argArgsOff, int16(argOffset*8), asm.DWord),
```
YutaroHayakawa: argArgsOff == R1, so this means the R1 always pointing to the argument throughout the evaluation right? Are we planning to save and restore this pointer after every BPF helper call?

pkg/celbpf/codegen.go
```diff
@@ -168,3 +168,51 @@ func (g *codeGenerator) pushArg(argTy *cgTypes.Type, argOffset int) error {
...
+// emit addition
+func (g *codeGenerator) emitAdd(
+	r1 asm.Register, ty1 *cgTypes.Type,
+	r2 asm.Register, ty2 *cgTypes.Type,
+) error {
+	switch {
+	case ty1.TypeName() == s64Ty.TypeName() && ty2.TypeName() == s64Ty.TypeName(),
+		ty1.TypeName() == s32Ty.TypeName() && ty2.TypeName() == s32Ty.TypeName(),
+		ty1.TypeName() == u64Ty.TypeName() && ty2.TypeName() == u64Ty.TypeName(),
+		ty1.TypeName() == u32Ty.TypeName() && ty2.TypeName() == u32Ty.TypeName():
```
YutaroHayakawa, line 215: This type checking is already performed in the checker. Do we need to double-check it here?

YutaroHayakawa (changes requested):
Exciting to see this finally landed to Tetragon! I reviewed the compiler part and made some comments. I couldn't review the last few commits as I'm not that familiar with the Tetragon internals.

I think the code to wire-up CEL to Tetragon (CRD generation, CRD => loader interface and C interface) can be separated into different PR.

Currently, there's no description around how each concept in the cel-go library works (like TypeProvider, Overload, Parser/Checker, AST, etc), so from those who first see this, it's pretty hard to review I assume. Maybe better to have more description around that in the code.

pkg/celbpf/codegen.go
kkourt: Because it results to an invalid opcode: https://github.com/cilium/ebpf/blob/61c4efeaf4c2993d0e74300003e29a3164b60ac9/asm/load_store.go#L316-L318

I'll fold the change so that we don't commit the problematic code.

pkg/celbpf/codegen.go
kkourt: Makes sense. I'll pass a scratch register as an argument. At some point, I'll write a trivial register allocator as well.

pkg/celbpf/compiler.go
kkourt: Yes, I believe typing ensures the right number of arguments.

pkg/celbpf/compiler.go
kkourt: We don't, but I find it convenient to have a breakdown of the different types (instead of having something like "%T" in the common error).

pkg/celbpf/env.go
kkourt: For equality, I think it's fine to have it as a parameterized overload. And I'm trying to write the code in such as a way so that if we don't support something, it will fail on the code generator anyway.

pkg/celbpf/compiler.go
kkourt: The "bool" case for this was added in the patch that introduced logical operators. Let me try to push it earlier where it probably belongs (together with the test cases that you mentioned).

pkg/celbpf/codegen.go
kkourt: Yap, makes sense. Changed it.

pkg/celbpf/codegen.go
kkourt: Yes, for simplicity I'm never modifying R1 and R2 in this first version. If we ever want to add a helper call (or need more registers), we would need to figure it out. An alternative would be to save the arguments into a map.

pkg/celbpf/codegen.go
kkourt: I think so. Because depending on the representation, we might want to generate different code for different types (e.g., if we ever want to use 4 bytes for {u,s}32 in the stack).

kkourt:
> Exciting to see this finally landed to Tetragon! I reviewed the compiler part and made some comments. I couldn't review the last few commits as I'm not that familiar with the Tetragon internals.

Thanks for the review, and sharing your code for this! Pushed a new version that addresses your feedback.

> I think the code to wire-up CEL to Tetragon (CRD generation, CRD => loader interface and C interface) can be separated into different PR.

I generally prefer having the user of a facility in the same PR as the introduction of the facility, but if folk find it hard to review, I'm happy to split it.
 
> Currently, there's no description around how each concept in the cel-go library works (like TypeProvider, Overload, Parser/Checker, AST, etc), so from those who first see this, it's pretty hard to review I assume. Maybe better to have more description around that in the code.

Could be the case, yes. Let's see what the Tetragon folk think. If it's not clear, I can try to explain it better.

pkg/celbpf/codegen.go
YutaroHayakawa: Still using R3?

pkg/celbpf/codegen.go
YutaroHayakawa: But we don't have such a case now right? Then, it might be simpler to just remove the unnecessary code?

pkg/celbpf/compiler.go
YutaroHayakawa: Can't we just consolidate it to `fmt.Errorf("expression kind '%T' not supported")`? Seems like it just expands the type name within the single quote.

pkg/celbpf/env.go
YutaroHayakawa: > it will fail on the code generator anyway

I think it's better to capture this kind of the error in the type checker as much as possible because it's a typing problem. Handling type error in the code generator makes the code generator implementation complex and it's very easy for us to forget which check is done in the type checker and which is in the code generator.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1223,10 +1224,16 @@ func loadSingleKprobeSensor(id idtable.EntryID, bpfDir string, load *program.Pro
 		return err
 	}
 
+	rewriteProg := make(map[string]func(prog *ebpf.ProgramSpec) error)
 	if entry := gk.loadArgs.selectors.entry; entry != nil {
-		// functions generated for CEL expressions
-		load.RewriteProg["generic_kprobe_filter_arg"] = entry.CelExprFunctions.RewriteProg
+		if celbpf.EnabledInBPF() {
```
andrewstrohman, line 1229: The nested if condition is the same as the outer if condition.

pkg/celbpf/celbpf.go
```diff
@@ -0,0 +1,117 @@
...
+
+	"github.com/cilium/tetragon/pkg/bpf"
+	"github.com/cilium/tetragon/pkg/config"
+)
+
+type Env struct{}
+
+type ExprArg struct {
+	GenTy     int
+	ArgOffset int
```
YutaroHayakawa, line 25: This offset will be ultimately used as an offset parameter of Load instruction which is 16bit. How about change this to int16 to reduce the risk of overflowing?

andrewstrohman:
For commit "sensors: support for CEL expressions in kprobes", the first part of the commit message:
> This commit adds the necessary code to load the CEL functions before we
load the program. We just append the generated instructions to the
program instructions and let the cilium/ebpf loader to do the rest.

This doesn't actually happen in this commit.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1223,6 +1224,17 @@ func loadSingleKprobeSensor(id idtable.EntryID, bpfDir string, load *program.Pro
 		return err
 	}
 
+	rewriteProg := make(map[string]func(prog *ebpf.ProgramSpec) error)
+	if entry := gk.loadArgs.selectors.entry; entry != nil {
+		if celbpf.EnabledInBPF() {
+			// functions generated for CEL expressions
+			if celbpf.EnabledInBPF() {
+				rewriteProg["generic_kprobe_filter_arg"] = entry.CelExprFunctions.RewriteProg
```
olsajiri: we have access functions for KernelSelectorState members, let's add one for CelExprFunctions ?

bpf/process/cel_expr.h
```diff
@@ -0,0 +1,53 @@
...
+
+#ifndef __CEL_EXPR_H__
+#define __CEL_EXPR_H__
+
+/*
+ * The code for the cel_expr_N functions will be generated and linked when the policy is loaded
+ */
+
+#if defined(__LARGE_BPF_PROG) && (defined(GENERIC_KPROBE) && !defined(__MULTI_KPROBE))
+int cel_expr_0(long *argsoff, char *args);
```
olsajiri, line 12: curious what's the problem with multi kprobes?

bpf/process/types/basic.h
```diff
@@ -2043,6 +2044,14 @@ selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx,
 		margsoff = (seloff + argsoff) & INDEX_MASK;
 		filter = (struct selector_arg_filter *)&f[margsoff];
 
+#ifdef __LARGE_BPF_PROG
+		if (filter->op == op_cel_expr) {
+			int ret = cel_expr(filter->index, e->argsoff, e->args);
+
```
olsajiri: I think we were discussing printing size of the programs to the tetragon output in future,
do you have any idea about the bpf program sizes? I expect it's not much, just curious

pkg/bpf/detect_linux.go
```diff
@@ -606,15 +607,86 @@ func HasKfunc(name string) bool {
...
 	// not load the BTF again
 	defer ebtf.FlushKernelSpec()
-	return fmt.Sprintf("override_return: %t, buildid: %t, kprobe_multi: %t, uprobe_multi: %t, fmodret: %t, fmodret_syscall: %t, signal: %t, large: %t, link_pin: %t, lsm: %t, missed_stats_kprobe_multi: %t, missed_stats_kprobe: %t, batch_update: %t, uprobe_refctroff: %t, audit_loginuid: %t, probe_write_user: %t, uprobe_regs_change: %t",
+	return fmt.Sprintf("override_return: %t, buildid: %t, kprobe_multi: %t, uprobe_multi: %t, fmodret: %t, fmodret_syscall: %t, signal: %t, large: %t, link_pin: %t, lsm: %t, missed_stats_kprobe_multi: %t, missed_stats_kprobe: %t, batch_update: %t, uprobe_refctroff: %t, audit_loginuid: %t, probe_write_user: %t, uprobe_regs_change: %t, mix_bpf_and_tail_calls: %t",
 		HasOverrideHelper(), HasBuildId(), HasKprobeMulti(), HasUprobeMulti(),
 		HasModifyReturn(), HasModifyReturnSyscall(), HasSignalHelper(), HasProgramLargeSize(),
 		HasLinkPin(), HasLSMPrograms(), HasMissedStatsKprobeMulti(), HasMissedStatsPerfEvent(),
 		HasBatchAPI(), HasUprobeRefCtrOffset(), HasAuditLoginuid(), HasProbeWriteUserHelper(),
-		HasUprobeRegsChange())
+		HasUprobeRegsChange(), DetectMixBpfAndTailCalls())
```
olsajiri: I wonder we should stop doing/showing feature detection at each tetragon start with default log level and do not disrupt the system.. maybe just print it in bugtool or such

pkg/selectors/celexpr.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (cefs CelExprFunctions) functions() map[string]asm.Instructions {
+	fns := make(map[string]asm.Instructions)
+	// first, fill in the functions for which we have an expression
+	for i, insns := range cefs {
+		fns[CelExprFuncName(i)] = insns
+	}
+	// next, fill in default functions (returning zero) for the rest of expressions
+	// NB: this is needed, because otherwise loading will fail due to unresolved symbols.
+	for i := len(fns); i < MaxCelExprFunctions; i++ {
```
olsajiri, line 138: is there any way we could remove those symbols instead? it's empty function, so not big deal, just curious ;-)

kkourt: That's the first thing I tried but couldn't make it work. If I remember correctly, there are some metadata for the functions that the kernel was expecting to find, and I couldn't figure out how to remove them from the spec.

pkg/celbpf/codegen.go
kkourt: Indeed. I am not sure how I dropped this, but it should be properly fixed in the next push. Thanks!

pkg/celbpf/compiler.go
kkourt: I have a slight preference as it is now, and I don't think it's a big deal.

pkg/celbpf/env.go
kkourt: I agree that we should catch it in the type checker if we can. I was commenting on the possibility that the type checker has errors. I think parametric equality makes sense for now. If we ever have a type that does not support equality, we can revisit.

pkg/celbpf/codegen.go
kkourt: We do not, but I think the code is more explicit (and clear) this way.

pkg/sensors/tracing/generickprobe.go
kkourt: nice catch, thanks!

pkg/celbpf/celbpf.go
kkourt: Nice catch!

I did the following:
```diff
diff --git a/pkg/celbpf/codegen.go b/pkg/celbpf/codegen.go
index 9407d105e..8dbc95809 100644
--- a/pkg/celbpf/codegen.go
+++ b/pkg/celbpf/codegen.go
@@ -7,6 +7,7 @@ package celbpf
 
 import (
 	"fmt"
+	"math"
 
 	"github.com/cilium/ebpf/asm"
 	cgOperators "github.com/google/cel-go/common/operators"
@@ -140,12 +141,18 @@ func (g *codeGenerator) emitU32(reg asm.Register, regTy *cgTypes.Type) error {
 }
 
 func (g *codeGenerator) pushArg(argTy *cgTypes.Type, argOffset int, tmp1, tmp2 asm.Register) error {
+	off := argOffset * 8
+	if off > math.MaxInt16 {
+		return fmt.Errorf("offset %d overflows 16-bits", off)
+	}
+	off16 := int16(off)
+
 	switch argTy {
 	case u32Ty, s32Ty:
 		g.stackTop -= 8
 		g.emitRaw(
 			// tmp1 = *(u64 *)(argArgsOff + idx)
-			asm.LoadMem(tmp1, argArgsOff, int16(argOffset*8), asm.DWord),
+			asm.LoadMem(tmp1, argArgsOff, off16, asm.DWord),
 			asm.And.Imm(tmp1, 0x7ff),
 			// tmp1 += args
 			asm.Add.Reg(tmp1, argArgs),
@@ -157,7 +164,7 @@ func (g *codeGenerator) pushArg(argTy *cgTypes.Type, argOffset int, tmp1, tmp2 a
 		g.stackTop -= 8
 		g.emitRaw(
 			// tmp1 = *(u64 *)(argArgsOff + idx)
-			asm.LoadMem(tmp1, argArgsOff, int16(argOffset*8), asm.DWord),
+			asm.LoadMem(tmp1, argArgsOff, off16, asm.DWord),
 			asm.And.Imm(tmp1, 0x7ff),
 			// tmp1 += args
 			asm.Add.Reg(tmp1, argArgs),

```

which feels safer to me, becase we return an error.

pkg/sensors/tracing/generickprobe.go
kkourt: Done

bpf/process/cel_expr.h
kkourt: We are using an id to pass to the bpf side what celbpf function to call in the bpf code.
This id is currently kept at the KernelSelectorState struct.
For multi-kprobes (I think) we have multiple kernelSelector structs so both selectors (in different kprobes) would get an id of 0, which wouldn't work because the expressions are different.

bpf/process/types/basic.h
kkourt: You mean with the CEL expressions compiled in? It depends on the expression. The code generator is not at all optimized at the moment, so it uses many instructions but I think that's fixable.

kkourt:
Pushed a new version. Thanks for the reviews!

PR #4508, opened 2026-01-12
new: event log grpc service

FedeDP:
### Description
This is a new grpc service that allows to request rotation log params changes to event log exporter.

### Changelog

```release-note
new: event log grpc service
```

### DIFF
--- a/pkg/exporter/exporter_test.go
+++ b/pkg/exporter/exporter_test.go
@@ -8,13 +8,20 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 	"testing"
+	"testing/synctest"
 	"time"
 
+	"github.com/cilium/lumberjack/v2"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/cilium/tetragon/pkg/server/eventlog"
+
+	"github.com/cilium/tetragon/pkg/option"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/encoder"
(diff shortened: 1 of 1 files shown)


cmd/tetragon/main.go
```diff
@@ -690,38 +705,7 @@ func startExporter(ctx context.Context, server *server.Server) error {
 	finfo, err := os.Stat(filepath.Clean(option.Config.ExportFilename))
 	if err == nil && finfo.IsDir() {
 		// Error if exportFilename points to a directory
-		return errors.New("passed export JSON logs file point to a directory")
-	}
-	logFile := filepath.Base(option.Config.ExportFilename)
```
FedeDP, line 699: Moved to be internally managed by exporter.

pkg/server/eventlog/server.go
```diff
@@ -0,0 +1,80 @@
...
+	// Update server params
+	if req.MaxSize != nil {
+		s.MaxSize = *req.MaxSize
+	}
+	if req.MaxBackups != nil {
+		s.MaxBackups = *req.MaxBackups
+	}
+	if req.RotationInterval != nil {
+		s.RotationInterval = *req.RotationInterval
+	}
```
kkourt, line 104: Does it make sense to check for values that may cause problems? E.g., req.MaxSize == 0?

pkg/server/eventlog/server.go
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package eventlog
```
kkourt, line 4: Would this package better placed under server? i.e., pkg/server/eventlog?

pkg/exporter/exporter.go
```diff
@@ -92,3 +145,29 @@ func (e *Exporter) SendMsg(_ any) error {
...
+
+	if params.MaxBackups != nil {
+		writer.MaxBackups = int(*params.MaxBackups)
+	}
+
+	if params.RotationInterval != nil {
+		e.rotateTimer = time.AfterFunc(
+			*params.RotationInterval,
+			e.rotate,
+		)
```
kkourt, line 178: Should we use e.rotateTimer?Stop() if rotateTimer is not nil?

pkg/server/eventlog/server.go
FedeDP: I don't think that is necessary; we do not check their value from `option.Config` either.
Also:
* `MaxSize == 0` would enforce default value to be used (100MB):
```go
// max returns the maximum size in bytes of log files before rolling.
func (l *Logger) max() int64 {
	if l.MaxSize == 0 {
		return int64(defaultMaxSize * megabyte)
	}
	return int64(l.MaxSize) * int64(megabyte)
}
```

pkg/server/eventlog/server.go
FedeDP: Yes it might be a better place, thanks!

pkg/exporter/exporter.go
FedeDP: Sure, that makes sense!

FedeDP:
The test is expected to badly fail because of a goroutine leak in lumberjack code; see here for the fix: https://github.com/cilium/lumberjack/pull/7

FedeDP:
`cilium/lumberjack` had a patch release (v2.4.2) that fixed the goroutine leak. I upgraded it in latest commit. :pray: 

EDIT: since the update reached main, rebased on top of main and removed this commit.

pkg/exporter/exporter.go
```diff
@@ -92,3 +150,33 @@ func (e *Exporter) SendMsg(_ any) error {
 func (e *Exporter) RecvMsg(_ any) error {
 	return nil
 }
+
+func (e *Exporter) SetLogParams(params eventlog.Params) error {
```
kkourt, line 154: Do we need to synchronize here? Is it possible to get multiple of these requests via gRPC?

FedeDP: I think you are right; i was not sure whether grpc requests were running concurrently, but i bet so.
Let me fix this.

FedeDP: Pushed the changes.

FedeDP: Mmh i am wondering whether it is better to sync at the `eventlog/server` level instead; this leaves less freedom to `eventLogParamsSetter` interface implementers, but at the same time gives much more guarantees that every call to this grpc API is not concurrent.

FedeDP: IMHO :point_up: is the best solution; this way implementers are not aware of any sync method since it is managed above them. 
Pushed the changes.

pkg/server/eventlog/server.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+type Params struct {
+	MaxSize          *int32
+	RotationInterval *time.Duration
+	MaxBackups       *int32
+}
+
+type eventLogParamsSetter interface {
+	SetLogParams(params Params) error
```
kkourt, line 26: Can we document that the caller needs to serialize the calls to this method here?

cmd/tetragon/main.go
```diff
@@ -495,11 +496,25 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
...
+	// Fetch the exporter if needed
+	var exporter *exporter.Exporter
+	if option.Config.ExportFilename != "" {
+		exporter, err = getExporter(ctx, pm.Server)
+		if err != nil {
+			return fmt.Errorf("failed to fetch json exporter: %w", err)
+		}
+	}
+
+	logSrv := eventlog.New(exporter)
```
kkourt, line 509: If I do: --export-filename "", and then `.tetra eventlog set`, I get a SIGSEGV.
Let's fix this.

Also: I'm a bit confused as to why the export file is set by default.
At least in help, it is mentioned that this is disabled by default:
```
      --export-filename string                     Filename for JSON export. Disabled by default
```

FedeDP: > If I do: --export-filename "", and then .tetra eventlog set, I get a SIGSEGV.

Uh let me double check; thanks!

pkg/server/eventlog/server.go
FedeDP: Mmh what do you mean?

cmd/tetragon/main.go
FedeDP: > Also: I'm a bit confused as to why the export file is set by default.

It is not set by default though. Did you see the opposite?

FedeDP: I chose to validate the passed `setters` to be non-nil interface and non-pointing-to-nil values. 
This enforces the check even for (possible) clients that make use of the eventlog grpc service.

pkg/observer/observertesthelper/observer_test_helper.go
```diff
@@ -413,7 +413,7 @@ func loadExporter(tb testing.TB, ctx context.Context, obs *observer.Observer, op
 	encoder := encoder.NewProtojsonEncoder(outF)
 
 	req := tetragon.GetEventsRequest{AllowList: opts.allowList, DenyList: opts.denyList}
-	exporter := exporter.NewExporter(ctx, &req, processManager.Server, encoder, outF, nil)
+	exporter, _ := exporter.NewExporter(ctx, &req, processManager.Server, encoder, outF, nil)
```
kkourt: I guess we could `Fatal()` here if there is an error?

FedeDP: Makes sense :) 
we can just return the `err` since it is managed by the callers.

PR #4510, opened 2026-01-12
docs: exclude nakryiko.com from link checker

puwun:
The nakryiko.com blog has transient certificate issues that cause false positives in the automated link checker.

Fixes: #4502

### Description
Added `nakryiko.com` to the lychee link checker exclude list in `.github/lychee.toml`. The automated link checker periodically reports certificate/network errors for links to this domain, even though the site and links are valid. This follows the existing pattern used for other domains with similar transient issues (e.g., `virtualbox.org`, `slack.cilium.io`)

### Changelog

```release-note
docs: Exclude nakryiko.com from automated link checker to prevent false positives
```

### DIFF
--- a/.github/lychee.toml
+++ b/.github/lychee.toml
@@ -11,6 +11,8 @@ exclude = [
   'docs/assets/icons/logo.*\.svg', 
   # the virtualbox website sends 500 regularly
   'virtualbox.org',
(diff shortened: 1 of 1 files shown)


mtardy (approved):
Thanks, that exactly what I wanted to do :)

Issue #4509, opened 2026-01-12
perfring test failure not handled correctly


kkourt:
Consider the following failed test:

```diff
diff --git a/pkg/sensors/tracing/tracepoint_test.go b/pkg/sensors/tracing/tracepoint_test.go
index 9f0d3c9c6..cc4ee5faf 100644
--- a/pkg/sensors/tracing/tracepoint_test.go
+++ b/pkg/sensors/tracing/tracepoint_test.go
@@ -1019,3 +1019,34 @@ spec:
        err = jsonchecker.JsonTestCheck(t, checker)
        require.NoError(t, err)
 }
+
+func TestTracepointShouldFail(t *testing.T) {
+       testutils.CaptureLog(t, logger.GetLogger())
+       ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+       defer cancel()
+
+       mypid := int(observertesthelper.GetMyPid())
+       t.Logf("filtering for my pid (%d)", mypid)
+
+       spec := &v1alpha1.TracingPolicySpec{
+               Tracepoints: []v1alpha1.TracepointSpec{{
+                       Subsystem: "raw_syscalls",
+                       Event:     "sys_enter",
+               }},
+       }
+
+       loadGenericSensorTest(t, spec)
+       t0 := time.Now()
+       loadElapsed := time.Since(t0)
+       t.Logf("loading sensors took: %s\n", loadElapsed)
+
+       eventFn := func(ev notify.Message) error {
+               return fmt.Errorf("FAILURE TEST")
+       }
+
+       ops := func() {
+               time.Sleep(10)
+       }
+
+       perfring.RunTest(t, ctx, ops, eventFn)
+}
```

perfring.RunTest() does not terminate even though eventFn returns an error.

```
    logcapture.go:24: time=2026-01-12T17:45:17.406+01:00 level=INFO msg="Loaded sensor successfully" sensor=generic_tracepoint
    tracepoint_test.go:1041: loading sensors took: 50ns
    perfring.go:202: error loop event function returned: FAILURE TEST
```

kevsecurity:
#4511 is fix.

mtardy:
> [#4511](https://github.com/cilium/tetragon/pull/4511) is fix.

fix got merged, closing this!

PR #4511, opened 2026-01-12
Tetragon: Perf ring testutil exit on error

kevsecurity:
The ring buffer test util in pkg/testutil/perfring do not exit correctly if a test fails. This is partly because the test function is launched in a goroutine, but mainly because we fail to cancel the ring buffer consumers' contexts. This commit addresses both issues.


kkourt (approved):
Thanks! tried it out and fixes https://github.com/cilium/tetragon/issues/4509! 🚀

PR #4516, opened 2026-01-13
test: add synthetic event generators for OOO scenarios

ariosmon:
Adds bash scripts to generate JSONL event streams for testing Tetragon's ProcessCache behavior under Out-Of-Order (OOO) conditions. Includes:
- Single scenario generator (smoke test).
- Stress test generator (load/memory leak test).
- Documentation (README).


ariosmon:
> Hey, just a comment, I saw it's a draft and I imagine it's been derived from your recent bash scripts but if we decide to merge as part of the project eventually I think it would be nice to use Go. In the past we had to migrate bash scripts to Go already so maybe we could do it in Go in the first place.

Hello @mtardy, we are working on the [synthetic event injection API](https://github.com/cilium/tetragon/issues/4497) and need to generate a JSON file with some async events. My initial approach was to create it using a Bash script, but we could also move to Go. If this approach is approved after some testing, I can migrate it to Go at that point and merge it into the master branch. @sglushko

Issue #4442, opened 2025-12-18
Support for `int32_arr` argument type to capture integer array (file descriptor array) output parameters


ywmei-brt1:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem?

Yes. Currently, Tetragon lacks the ability to natively capture arguments that are pointers to arrays of integers, specifically when those arrays are used as output parameters populated by the kernel. 

Standard examples include:

1. `pipe2(int pipefd[2], int flags)`: The first argument, `pipefd`, is a pointer to an array of two integers.
2. `socketpair(int domain, int type, int protocol, int sv[2])`: The fourth argument, `sv`, is a pointer to an array of two integers.

In both cases, the kernel populates these arrays with new file descriptors during syscall execution, the values in the array can only be read via returnCopy.  Without support for an `int32_arr` type, users cannot observe the actual file descriptors returned by these syscalls, creating a gap in visibility regarding process inter-communication channels and resource allocation.

### Describe the feature you would like

Introduce a new argument type `int32_arr` to Tetragon's TracingPolicy configuration. This feature should:

1.  **Support Array Types**: Allow `TracingPolicy` arguments to be defined with `type: "int32_arr"` and a `size` field (defines how many elements we read out of this int32_arr).
2.  **Support Return Copy**: Support `returnCopy: true` for this type. Since these arrays are often output parameters (like in `pipe2`), the BPF collector must be able to defer reading the memory until the function returns (kretprobe) to capture the values populated by the kernel.
3.  **API Visibility**: Expose these values in the Tetragon API (gRPC and JSON export) as a structured list of integers, rather than raw byte buffers.

Basically, say, if we want to know which two file descriptors `pipe2` syscall used, we want to be able to use a tetragon policy like this:

```
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "pipe-monitoring"
spec:
  lists:
  - name: "pipes"
    type: "syscalls"
    values:
    - "sys_pipe"
    - "sys_pipe2"
  kprobes:
  - call: "list:pipes"
    syscall: true
    args:
    - index: 0
      type: "int32_arr"
      size: 2
      returnCopy: true
      label: "pipefd"
```

to get the two file descriptor being used (say, they are `[3, 4]`) from a tetragon event like this:

```
...
  "function_name": "__x64_sys_pipe2",
  "args": [
    {
      "int32_list_arg": {"values": [3,4]},
[...]

PR #4519, opened 2026-01-14
fix(pkg/sensors): skip nested ptr detection for indexed array access.

FedeDP:
Refs #4442

### Description

When we want to resolve a precise index of an array, like for example `do_pipe2` with `resolve: "[1]"`, we want to skip nested pointer logic, since we are already going to dereference the correct index. 
Also, updated `TestResolveBTFArgFromKprobePolicy` to test the behavior.

### DIFF
--- a/pkg/sensors/tracing/generic.go
+++ b/pkg/sensors/tracing/generic.go
@@ -73,10 +73,21 @@ func formatBTFPath(resolvePath string) ([]string, error) {
 	return path, nil
 }
 
+// First argument is added to enforce the method to be called on a pointer type
+func isPointerToIndexedArray(_ *ebtf.Pointer, firstResolvePath string) bool {
+	var idx int
+	n, _ := fmt.Sscanf(firstResolvePath, `[%d]`, &idx)
+	return n == 1
+}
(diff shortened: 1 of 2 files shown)


mtardy (commented):
Thanks! Indeed it fixes our issue there but I was wondering, is there a way to simplify the code so that it works in all situations instead of having those nested checks?

Maybe I'm wrong since I'm not familiar with this but it looks to me:
1. We are trying to do something specific to a particular situation
2. We patch this particular situation to be less specific in some cases

Do you think we could simplify it?

If not, the patch looks great

FedeDP:
>    We are trying to do something specific to a particular situation
>    We patch this particular situation to be less specific in some cases

I agree with you, but couldn't find a quick better way to deal with the problem. 
I'll try again tomorrow and let you know my findings :D

FedeDP:
@andrewstrohman FYI

FedeDP:
@mtardy couldn't find a better, working, way to do this :/

mtardy (approved):
yes lgtm then! thank you for investigating this

Issue #3493, opened 2025-03-12
Document socket tracking limitations


kevsecurity:
Socket tracking is intended to create a mapping between network sockets and the processes that created them. When reporting a socket, Tetragon will overwrite the process information with the details from its mapping. This is usually required because networking operations are often asynchronous and occur in a process context that isn't the process that created the socket.

This has the following limitations (please add):

- The socket mappings are stored in a LRU hash in the kernel. This has an upper limit for entries. When full, old entries are evicted to make space for new entries. An adversary could overflow the map so that their evil socket is no longer tracked, so events related to it will likely be ascribed to the wrong process. Solutions include blocking socket creation if the map is full (switch to a non-LRU hash and check return value of map update), or alerting on it, noting in the event that the process information is questionable.
- Sockets do not actually belong to processes (contradicting the above point to some degree) but are shared between processes. This happens on fork if parent and child both keep the file descriptor open. It also happens if a process shares a file descriptor with another via IPC. We attribute all socket activity to the process that created the socket, but that process could exit and another could continue to service the socket. After exit our mapping would no longer make sense. An adversary might make use of the confusion to hide an attack.

PR #4521, opened 2026-01-14
docs: document socket tracking limitations

puwun:
Add documentation for socket tracking limitations.

The new 'Limitations' section covers:

- LRU map overflow risks: Socket mappings use a fixed-size LRU hash that can evict old entries when full, potentially causing events to be attributed to the wrong process.

- Socket sharing behavior: Sockets can be shared between processes via fork() or IPC, and Tetragon attributes activity to the creating process which may lead to stale mappings.

Both limitations are documented with appropriate `warning` and `caution` blocks to ensure visibility.

Fixes: #3493

### Changelog

```release-note
docs: document socket tracking limitations including LRU map overflow risks and socket sharing behavior
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -769,7 +769,30 @@ See [`TrackSock`](/docs/concepts/tracing-policy/selectors/#tracksock-action) and
   returnArgAction: TrackSock
 ```
 
-Socket tracking is only available on kernels >=5.3.
+#### Limitations
+
+Socket tracking has the following limitations:
+
+- Kernel version: socket tracking is only available on kernels >=5.3.
+
+- LRU map overflow: socket mappings are stored in an LRU (Least Recently Used)
+  hash map in the kernel with a fixed upper limit for entries. When the map is
+  full, old entries are evicted to make space for new ones. If many sockets are
+  created in a short period, older socket mappings may be lost and network
+  events may be attributed to the wrong process.
+
+- Socket sharing between processes: sockets are not strictly owned by a single
+  process—they can be shared via `fork()` (when both parent and child keep the
+  file descriptor open) or via IPC file descriptor passing. Tetragon attributes
+  all socket activity to the process that originally created the socket. If that
+  process exits while another continues using the socket, the mapping references
+  a process that no longer exists.
+
+{{< warning >}}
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -771,6 +771,39 @@ See [`TrackSock`](/docs/concepts/tracing-policy/selectors/#tracksock-action) and
 
 Socket tracking is only available on kernels >=5.3.
```
mtardy: This could be integrated into the limitations? wdyt

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -771,6 +771,39 @@ See [`TrackSock`](/docs/concepts/tracing-policy/selectors/#tracksock-action) and
...
+keep the file descriptor open, or when a process shares a file descriptor with
+another process via IPC mechanisms.
+
+Tetragon attributes all socket activity to the process that originally created
+the socket. However, if that process exits while another process continues to
+use the socket, the mapping will reference a process that no longer exists. This
+can lead to events being associated with stale process information. In security
+contexts, an adversary might exploit this behavior to obscure the true source of
+network activity.
+{{< /caution >}}
```
mtardy: While I really appreciate that you found and reused `warning` and `caution`, for such amount of text I think it would be better to use plain paragraphs or list. Also avoid using bold text.

I think it could be nice to explain the limitations as you did with a few paragraphs, or a list with elements starting with the title you put in bold for example. Something a bit like:

> - LRU map overflow: socket mappings are stored in an LRU (Least Recently Used) hash map in the kernel with a fixed upper limit for entries. When the map is full, old entries are evicted to make space for new ones. This means that if many sockets are [...]
> - Socket sharing between processes: sockets are not strictly owned by a single process—they can be shared between processes. This happens when a process calls `fork()` and both parent and child [...]

You can have multiple paragraphs in a list item. But if you can also manage to make it less verbose that would be great, if it's not possible it's okay, we need to stay accurate.

You can always use `warning` at the end, insisting that for A and B limitations, you have security implications.

mtardy (commented):
Thanks a lot for doing, just a few comments

puwun:
@mtardy thanks, updated 👍🏼

PR #4527, opened 2026-01-15
bpf: break up arg filter tail call to increase instruction budget

andrewstrohman:
### Description
We are running very short on instruction budget for the args tail call on 4.19 kernels. As a result, we haven't been able to add a lot of recent features for 4.19. In addition to missing features on 4.19, the complexity caused by `#ifdef __LARGE_BPF_PROG` makes it difficult to follow code. As I developed [exposing NULL dereferencing](https://github.com/cilium/tetragon/pull/4327#issuecomment-3753583454), we found the code hard to review because of the logic being split.

As a first step to solving this problem, @olsajiri created a [PR](https://github.com/cilium/tetragon/pull/4439) to address this issue for process tail call. This PR is modeled after that change, but instead breaks up the args tail call, which filters args.

While preparing for this change, I found bug related to the op_capabilities_gained filter's behavior, which I fixed in this series.

Finally, an optimization is included in this change, where we stop evaluating a selector's filters (return early) after a selector's filter has failed, because all of a selector's filters need to pass in order for the selector to match.

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -139,6 +139,7 @@ enum {
 	TAIL_CALL_SEND = 5,
 	TAIL_CALL_PATH = 6,
 	TAIL_CALL_PROCESS_2 = 7,
+	TAIL_CALL_ARGS_2 = 8,
 };
 
 struct selector_action {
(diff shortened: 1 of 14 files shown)


bpf/process/types/basic.h
```diff
@@ -2072,7 +2072,7 @@ selector_arg_offset(__u8 *f, struct msg_generic_kprobe *e, __u32 selidx,
 				__u32 index2 = *((__u32 *)&filter->value);
 				__u64 cap_new = *(__u64 *)get_arg(e, index2);
 
-				pass = !!((cap_old ^ cap_new) & cap_new);
+				pass &= !!((cap_old ^ cap_new) & cap_new);
```
kkourt, line 2075: Thanks for fixing this!

We would probably need to backport this change?
We typically backport at the PR level, so would it make sense to backport the full PR?

kkourt (approved):
LGTM, thanks!

bpf/process/types/basic.h
andrewstrohman: I think that it would be better to backport just the commit that fixes this, instead of the entire PR. I say this because it's the only part of this PR which is a fix (the rest is enhancement) and I thought we usually only backport fixes.

That being said, I'm happy to backport the entire PR if the team/community thinks there is value in that.

olsajiri: +1 for backporting just the commit, no need for the whole pr

kkourt:
Thanks @andrewstrohman! Merging! Can you please do the backport?

andrewstrohman:
> Thanks @andrewstrohman! Merging! Can you please do the backport?

[here](https://github.com/cilium/tetragon/pull/4564) it is.

PR #4540, opened 2026-01-18
cgroup: Check cgroup level with tracking level

sayboras:
```release-note
cgroup: Check cgroup level with tracking level
```

### DIFF
--- a/pkg/sensors/exec/cgroups_test.go
+++ b/pkg/sensors/exec/cgroups_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/stretchr/testify/require"
 
 	"github.com/cilium/tetragon/pkg/bpf"
+	"github.com/cilium/tetragon/pkg/errmetrics"
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/sensors/cgroup/cgrouptrackmap"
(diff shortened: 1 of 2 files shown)


bpf/cgroup/bpf_cgroup_release.c
```diff
@@ -33,12 +34,24 @@ tg_tp_cgrp_release(struct bpf_raw_tracepoint_args *ctx)
 		return 0;
 
 	cgrp_track = map_lookup_elem(&tg_cgrps_tracking_map, &cgrpid);
-	/* TODO: check cgroup level if it is under our tracking level
-	 *   then we probably did miss it and should report this.
-	 *   Otherwise the cgroup was never tracked and let's exit.
-	 */
-	if (!cgrp_track)
+	if (!cgrp_track) {
```
olsajiri, line 37: looks like this could be function used in here and in tg_tp_cgrp_rmdir?

sayboras: thanks, I will push one update :100:

bpf/cgroup/bpf_cgroup_release.c
```diff
@@ -33,12 +34,10 @@ tg_tp_cgrp_release(struct bpf_raw_tracepoint_args *ctx)
...
+		check_cgroup_tracking_miss(cgrp);
 		return 0;
```
olsajiri, line 40: nit, would it make any sense to get the error metric from particular file, like to see if it was rmdir or release who trigger that? maybe we could call errmetrics from outside `check_cgroup_tracking_miss`, like:

```
        cgrp_track = map_lookup_elem(&tg_cgrps_tracking_map, &cgrpid);
        if (!cgrp_track) {
-               check_cgroup_tracking_miss(cgrp);
+               if (check_cgroup_tracking_miss(cgrp))
+                       errmetrics(ENOENT);
                return 0;
        }
```

pkg/sensors/exec/cgroups_test.go
```diff
@@ -1338,3 +1339,94 @@ func TestCgroupv2ExecK8sHierarchyInUnified(t *testing.T) {
 
 	testCgroupv2K8sHierarchy(ctx, t, cgroups.CGROUP_UNIFIED, true)
 }
+
+// TestCgroupMissedTrackingErrMetrics verifies that when a cgroup that should
+// have been tracked (based on level and hierarchy) is removed without being
+// in the tracking map, an error metric is recorded.
+func TestCgroupMissedTrackingErrMetrics(t *testing.T) {
+	testutils.CaptureLog(t, logger.GetLogger())
```
olsajiri, line 1347: I think it's better to split test into separate commit for review and possible backports

olsajiri (commented):
ltgm, left 2 comments

PR #4545, opened 2026-01-19
fix: Change kernel config detection failure to info level

dangome3:
Tetragon attempts to read the kernel config by trying a couple of places where it is found. This should be `Info` instead of `Error`:

```
level=error msg="Detecting kernel config failed" error="open /boot/config: no such file or directory"
```

This PR changes `Error` for `Info` and fixes this log to be shown twice

### DIFF
--- a/pkg/bpf/detect_linux.go
+++ b/pkg/bpf/detect_linux.go
@@ -518,20 +518,23 @@ func HasUprobeRefCtrOffset() bool {
 	return uprobeRefCtrOffset.detected
 }
 
-func detectAuditLoginuid() bool {
-	if kernels.MinKernelVersion("5.1") && kernels.DetectConfig(kernels.CONFIG_AUDIT) {
-		return true
-	} else if kernels.DetectConfig(kernels.CONFIG_AUDITSYSCALL) {
-		return true
-	}
-
-	_, err := os.Stat("/proc/self/loginuid")
-	return err == nil
-}
-
 func detectAuditLoginuidOnce() {
 	auditLoginuid.init.Do(func() {
-		auditLoginuid.detected = detectAuditLoginuid()
+		cfg := kernels.CONFIG_AUDITSYSCALL
+		if kernels.MinKernelVersion("5.1") {
+			cfg = kernels.CONFIG_AUDIT
+		}
+		if ok, errConfig := kernels.DetectConfig(cfg); ok {
+			auditLoginuid.detected = true
+		} else if _, errStat := os.Stat("/proc/self/loginuid"); errStat == nil {
+			auditLoginuid.detected = true
+		} else {
+			if errConfig != nil {
+				logger.GetLogger().Info("failed to detect config and to stat loginuid", "config_error", errConfig, "config", string(cfg), "stat_error", errStat)
+			} else {
+				logger.GetLogger().Info("failed to stat loginuid", logfields.Error, errStat)
+			}
+		}
 	})
 }
 
--- a/pkg/kernels/config.go
+++ b/pkg/kernels/config.go
@@ -131,10 +131,9 @@ func initKernelConfig() error {
 
 // DetectConfig checks if a kernel config option is enabled
 // It first tries /proc/config.gz, then looks for config file in /boot for the current kernel
-func DetectConfig(conf Config) bool {
+func DetectConfig(conf Config) (bool, error) {
 	if err := getKernelConfig(); err != nil {
-		logger.GetLogger().Error("Detecting kernel config failed", logfields.Error, err)
-		return false
+		return false, fmt.Errorf("getting kernel config failed: %w", err)
 	}
 
 	if val, ok := kernelConfigMap[conf]; ok {
@@ -142,11 +141,11 @@ func DetectConfig(conf Config) bool {
 		// it's mostly treated as a kernel parameter rather than a feature switch,
 		// so we temporarily filter these out.
 		if val == "y" || val == "m" {
-			return true
+			return true, nil
 		}
 	}
 
-	return false
+	return false, nil
 }
 
 func LogConfigs() string {


dangome3:
Doing some tests mocking the failure:

Screenshot 2026-01-19 at 4 34 59 PM

Now it is prevented from showing twice

pkg/bpf/detect_linux.go
```diff
@@ -519,8 +519,10 @@ func HasUprobeRefCtrOffset() bool {
 }
 
 func detectAuditLoginuid() bool {
-	if kernels.MinKernelVersion("5.1") && kernels.DetectConfig(kernels.CONFIG_AUDIT) {
-		return true
+	if kernels.MinKernelVersion("5.1") {
+		if kernels.DetectConfig(kernels.CONFIG_AUDIT) {
+			return true
+		}
```
kkourt: The old and the new code in this change seem equivalent to me. Am I missing something?

kkourt (commented):
Thanks @dangome3!

> Doing some tests mocking the failure:
> Now it is prevented from showing twice

Is there a specific change that achieves that?
It's typically useful to add the details of these tests and their results in the commit message for future reference.

pkg/kernels/config.go
```diff
@@ -133,7 +133,7 @@ func initKernelConfig() error {
 // It first tries /proc/config.gz, then looks for config file in /boot for the current kernel
 func DetectConfig(conf Config) bool {
 	if err := getKernelConfig(); err != nil {
-		logger.GetLogger().Error("Detecting kernel config failed", logfields.Error, err)
+		logger.GetLogger().Info("Detecting kernel config failed", logfields.Error, err)
```
mtardy: While not a strong blocker since the patch is better: I still thing it would be better to not log stuff in pkg exported functions, otherwise it really prevents from reusing it. Returning an error would be the right thing to do and decide on the caller function that are not exported stuff that could be reused but the end user.

pkg/bpf/detect_linux.go
dangome3: The change is pretty minor, the version before called `kernels.DetectConfig(kernels.CONFIG_AUDIT)` even if `kernels.MinKernelVersion("5.1")` is false, therefore there was an unnecessary call there and that's why the error message was shown twice.

This minor change assures that `kernels.DetectConfig(kernels.CONFIG_AUDIT)` is only called when the other condition is true, making the log showing just once, in the else statement

pkg/kernels/config.go
dangome3: Sure, I'll follow that approach then

Thanks for taking the time to point the issue!

pkg/bpf/detect_linux.go
kkourt: Logical operators such as `&&`  are typically [short-circuited](https://en.wikipedia.org/wiki/Short-circuit_evaluation), which is the case in go as well: https://go.dev/play/p/2NsumSeF_E0

dangome3: Got it! thanks for clarifying that!

That being said, I think it either way needs to be a nested condition now that I'm returning error from `kernels.DetectConfig` function

pkg/bpf/detect_linux.go
```diff
@@ -525,10 +519,22 @@ func HasUprobeRefCtrOffset() bool {
...
+	} else {
+		ok, err := kernels.DetectConfig(kernels.CONFIG_AUDITSYSCALL)
+		if err != nil {
+			logger.GetLogger().Info("Detecting kernel config failed from CONFIG_AUDITSYSCALL", logfields.Error, err)
+		}
+		if ok {
+			return true
+		}
 	}
```
mtardy: For the new thing, you could write something like that:

```suggestion
	cfg := kernels.CONFIG_AUDITSYSCALL
	if kernels.MinKernelVersion("5.1") {
		cfg = kernels.CONFIG_AUDIT
	}

	ok, err := kernels.DetectConfig(cfg)
	if err != nil {
    		logger.GetLogger().Info("Detecting kernel config failed", "config", string(cfg), logfields.Error, err)
	}
	return ok
}
```

pkg/bpf/detect_linux.go
mtardy, line 538: It's really nitpicking at this stage but we could argue that the logging should happen here so you are sure that you print only once but since both functions are unexported, it doesn't make a difference. Maybe `detectAuditLoginuid` should be inlined here if it's only used here btw.

pkg/kernels/config.go
```diff
@@ -131,22 +131,21 @@ func initKernelConfig() error {
 
 // DetectConfig checks if a kernel config option is enabled
 // It first tries /proc/config.gz, then looks for config file in /boot for the current kernel
-func DetectConfig(conf Config) bool {
+func DetectConfig(conf Config) (bool, error) {
 	if err := getKernelConfig(); err != nil {
-		logger.GetLogger().Error("Detecting kernel config failed", logfields.Error, err)
-		return false
+		return false, err
```
mtardy: fyi: can be useful sometimes to wrap errors, maybe it's not that useful here tbh since there's only one path.
```suggestion
		return false, fmt.Errorf("getting kernel config failed: %w", err)
```

pkg/bpf/detect_linux.go
dangome3: Thanks for the suggestion, I applied the proposed change. Taking care that the remaining code in the function still runs

```go
_, err = os.Stat("/proc/self/loginuid")
```

pkg/bpf/detect_linux.go
dangome3: Moved into `detectAuditLoginuidOnce`, now I think the info message makes more sense. Please let me know if you have any feedback!

pkg/bpf/detect_linux.go
```diff
@@ -524,20 +518,32 @@ func HasUprobeRefCtrOffset() bool {
...
+			} else if errStat != nil {
+				return false, fmt.Errorf("reading loginuid failed: %w", errStat)
+			}
+			return true, nil
+		}()
+
+		if err != nil {
+			logger.GetLogger().Info("audit loginuid not detected", logfields.Error, err)
+		}
+		auditLoginuid.detected = detected
```
mtardy: assuming auditLoginuid.detected is initialized as false (should be the case in Go if we don't anything particular) you could simplify like that to make the logic a bit simpler to read, you don't need a local function either.


```suggestion
cfg := kernels.CONFIG_AUDITSYSCALL
if kernels.MinKernelVersion("5.1") {
    cfg = kernels.CONFIG_AUDIT
}

if ok, errConfig := kernels.DetectConfig(cfg); ok {
    auditLoginuid.detected = true
} else if _, errStat := os.Stat("/proc/self/loginuid"); errStat == nil {
    auditLoginuid.detected = true
} else {
    if errConfig != nil {
        logger.GetLogger().Info("failed to detect config and to stat loginuid", "config_error", errConfig, "config", string(cfg), "stat_error", errStat)
    } else {
        logger.GetLogger().Info("failed to to stat loginuid", logfields.Error, errStat)
    }
}
```

dangome3: Thanks for this suggestion, you are right it is more readable now. The changes have been applied!

PR #4546, opened 2026-01-19
Enable uprobe preload arg reading for "args" config and for arm64

andrewstrohman:
### Description
Previously, in [this PR](https://github.com/cilium/tetragon/pull/4489), we added support for reading string arguments in a sleep-able "preload" uprobe program. This allowed us to read a string argument that are not paged in when the hook fires.

That previous change was specific to `data` tracing policy configurations that used source `pt_regs`. But uprobe `args` configurations suffer from the same problem for the same reason. This PR extends the previous PR to also allow for `string` type `args` configurations as well.

Initial investigation found the issue to be specific to x86, but the problem does indeed apply to all architectures. When swap is enabled, we have no guarantee that a pointer will point to paged-in memory at the time of the hook. As such, this PR enables preload arg reading for arm64 as well.

While these changes are specific to string arguments, the problem is not -- it can happen to any pointer type. The intention here is to add the other pointer types in the future.

This same issue, of trying to read user space memory that isn't paged in, applies to USDT as well. The intention is to enable the sleep-able preload programs for USDT in a follow-up PR.

### DIFF
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -574,11 +574,10 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 				// If we are getting string type from pt_regs register we can safely assume
 				// it's from user address, so we need to read it through preload.
 				if argType == gt.GenericStringType {
-					if bpf.HasKfunc("bpf_copy_from_user_str") {
-						preload = true
-					} else {
-						logger.GetLogger().Warn("can't preload string argument, might be wrong")
+					if !bpf.HasKfunc("bpf_copy_from_user_str") {
+						return fmt.Errorf("can't preload string for argument %d", i)
 					}
+					preload = true
 				}
 			} else if hasCurrentTaskSource(a) {
 				if !bpf.HasProgramLargeSize() {
@@ -604,11 +603,10 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 			}
 
 			if argType == gt.GenericStringType {
-				if bpf.HasKfunc("bpf_copy_from_user_str") {
-					preload = true
-				} else {
-					logger.GetLogger().Warn("can't preload string argument, might be wrong")
+				if !bpf.HasKfunc("bpf_copy_from_user_str") {
+					return fmt.Errorf("can't preload string for argument %d", i)
 				}
+				preload = true
 			}
 		}
 


bpf/process/generic_calls.h
```diff
@@ -584,7 +584,7 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 
 #if defined(GENERIC_TRACEPOINT) || defined(GENERIC_USDT)
 	a = (&e->a0)[index];
-	extract_arg(config, index, &a);
+	extract_arg(config, index, &a, false);
```
kkourt, line 587: Shouldn't can_sleep be true for USDTs?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -601,14 +601,22 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 				allBTFArgs[i] = btfArg
 				argType = findTypeFromBTFType(a, lastBTFType)
 			}
+
+			if argType == gt.GenericStringType {
+				if bpf.HasKfunc("bpf_copy_from_user_str") && runtime.GOARCH == "amd64" {
+					preload = true
+				} else {
+					logger.GetLogger().Warn("can't preload string argument, might be wrong")
```
kkourt: Can we bail out here and not load the policy? I think it's the right thing to do because we have a pretty strong signal that the policy will not work reliably.

CC: @olsajiri since  we discussed this offline

contrib/tester-progs/uprobe-test.c
```diff
@@ -8,13 +13,65 @@ int uprobe_test_lib_arg2(char a1, short a2);
...
 int uprobe_test_lib_arg4(long a1, int a2, char a3, void *a4);
 int uprobe_test_lib_arg5(int a1, char a2, unsigned long a3, short a4, void *a5);
+int uprobe_test_lib_string_arg(char *str);
+
+/* Page out a memory region containing 'data' of length 'len'.
+ * Returns the address of the paged-out region.
+ * The temp file backing is used to work around the fact that our test VMs
+ * don't have any disk backed storage to use for swap.
+ */
+void *pageout(void *data, size_t len) {
```
kkourt, line 23: Nice!

pkg/sensors/tracing/genericuprobe.go
olsajiri: yea, I already have change for the current code https://github.com/cilium/tetragon/pull/4575/changes/8e269ab5bb8ef066afec063093374f597c39872d 

perhaps you can add it to your pr and change yours

bpf/process/uprobe_preload.h
```diff
@@ -96,10 +141,21 @@ uprobe_preload(struct pt_regs *ctx)
 	if (config->arm[4] & ARGM_PT_REGS_PRELOAD)
 		preload_pt_regs_arg(ctx, config, 4);
 
+	if (config->arm[0] & ARGM_PRELOAD)
```
olsajiri: I paused on this because in theory one PRELOAD bit should be enough.. we just read register and then the data.. would be great to unify `ARGM_PT_REGS_PRELOAD` and `ARGM_PRELOAD` but I don't see an easy way for this, so I guess we're fine ;-)

bpf/process/generic_calls.h
```diff
@@ -651,16 +599,17 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 	 *   - current task object
 	 *   - real argument value
 	 */
-	if (am & ARGM_PT_REGS_PRELOAD)
+	if (am & (ARGM_PT_REGS_PRELOAD | ARGM_PRELOAD)) {
 		a = get_pt_regs_preload_arg(ctx, ty);
```
olsajiri: let's rename `get_pt_regs_preload_arg` to `get_preload_arg`

contrib/tester-progs/uprobe-test.c
olsajiri: +1, please use the tabs instead of space for indent

bpf/process/generic_calls.h
olsajiri: I think it needs support in the usdt sensor first.. that should be probably follow up change

andrewstrohman: Yes, I was planning on supporting USDT in a follow-up change.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Thanks Jiri. I cherry-picked your commit into this series. I adapted it to also apply the same change to my case as well (the non-data, args case).|

I edited the commit message a little to shorten the subject and remove the bit about being x86 only, because that no longer applies when the patch was put on top of this series.

contrib/tester-progs/uprobe-test.c
andrewstrohman: I've converted all the spaces to tabs.

bpf/process/generic_calls.h
andrewstrohman: Sounds good. Done!

bpf/process/uprobe_preload.h
andrewstrohman: I added a commit to the end of this series, with subject "DO NOT MERGE", that combines them into one bit. Please let me know what you think. If you think it's an improvement, I squash the change into earlier commit(s) before merging.

bpf/process/uprobe_preload.h
```diff
@@ -120,6 +120,27 @@ preload_arg(struct pt_regs *ctx, struct event_config *config, int index)
 	return 0;
 }
 
+struct preload_arg_data {
+	struct event_config *config;
+	struct pt_regs *ctx;
+};
+
+FUNC_INLINE int
```
olsajiri: maybe FUNC_LOCAL would be better for verifier in newer kernels

bpf/process/uprobe_preload.h
```diff
@@ -130,28 +151,18 @@ uprobe_preload(struct pt_regs *ctx)
...
-	if (config->arm[4] & ARGM_PRELOAD)
-		preload_arg(ctx, config, 4);
-
+	struct preload_arg_data preload_data = {
+		.config = config,
+		.ctx = ctx,
+	};
+
+#ifndef __V61_BPF_PROG
+#pragma unroll
```
olsajiri, line 164: I guess if we are doing that let's add the whole arsenal (the bpf_for path) like we do in prepend_path and other places

bpf/process/uprobe_preload.h
```diff
@@ -120,6 +120,27 @@ preload_arg(struct pt_regs *ctx, struct event_config *config, int index)
...
+
+FUNC_INLINE int
+try_preload_arg(int idx, struct preload_arg_data *data)
+{
+	asm volatile("%[idx] &= %1 ;\n"
+		     : [idx] "+r"(idx)
+		     : "i"(MAX_POSSIBLE_ARGS));
+
+	if (data->config->arm[idx] & ARGM_PRELOAD) {
+		if (data->config->arm[idx] & ARGM_PT_REGS)
```
olsajiri, line 136: ah right, we do have another bit just for PT_REGS, nice that should work

bpf/process/uprobe_preload.h
olsajiri: yep, I think it's better with that, thanks

bpf/process/generic_calls.h
kkourt: Sounds good!

bpf/process/uprobe_preload.h
andrewstrohman: OK, thanks, I did this. 

I added another commit to this series that introduces a macro that allows us to de-dup  many instances of this boiler plate code for doing loops. Please let me know what you think. There is a conflict between clang-format and checkpatch for that change. I'm not sure how to resolve it. Perhaps this a legit case for overriding the CI.

bpf/process/uprobe_preload.h
andrewstrohman: I did this, but I don't understand why this is better for the verifier in newer kernels, so I'm hoping you can help me understand it better.

I know that FUNC_LOCAL will cause the function to not be inlined in newer kernels. I think that this means we will do bpf-to-bpf calls as a result. I can see how this helps with instruction count budget, but the instruction count budget does not seem very constraining in these modern kernels.

I think that the verifier will need to re-validate offsets when transitioning from one bpf program to another. So in this sense, it seems like we may need to re-teach the verifier the bound checking in the called program, even if this was known by the calling program. Whereas, if we kept it inlined, we would not have to do that in the called program. Is this an actual downside of using FUNC_LOCAL? Am I thinking about this right?

bpf/process/uprobe_preload.h
andrewstrohman: @kkourt 
Given that you said:
> LGTM (assuming Jiri and CI are also happy)

I want to know you opinion about overriding the checkpatch CI for the last commit in this series. There seems to be a conflict between what clang-format and checkpatch wants.

andrewstrohman: Upon further reflection, I thought I didn't need a macro to accomplish the de-duping that I wanted to do. So, I converted the the macro to a function that took the callback as a parameter, and therefore eliminated the issues with clang-format and checkpatch. 

However, it turned out that this angered the 5.4 verifier. So, I ended up just backing out the de-duping change altogether, as it doesn't seem worth it any more. 

Sorry for the noise.

PR #4548, opened 2026-01-20
selectors: add selectors macros syntactic sugar

### LINKED ISSUE #4323: TracingPolicy: add global selectors syntactic sugar

_No response_

Sometimes in tetragon policies we need to write same selectors for each hook we want to intercept and report. For example, we may want to filter out all events, where process binary has specific prefix, e.g. we consider that all events from some directory are false positives.

For instance, we want to intercept `call_1`, `call_2`, `call_3`, and we always want to filter out events where binary locates in `dir`. 

The policy would look like:
```
spec:
  kprobes:
  - call: <call_1>
    selectors:
    - matchBinaries:
      - operator: "NotPrefix"
        values:
        - "dir"
  - call: <call_2>
    selectors:
    - matchBinaries:
      - operator: "NotPrefix"
        values:
        - "dir"
  - call: <call_3>
    selectors:
    - matchBinaries:
      - operator: "NotPrefix"
        values:
        - "dir"
```

To make such policies prettier and to improve experience of writing new policies, we could add **global selectors** - selectors that are declared in the beginning of specification and that are appended to every selector in the policy.

With global selectors, the policy would look like this:
```
spec:
  globalSelectors:
    matchBinaries:
	- operator: "NotPrefix"
	  values:
	  - "dir"
  kprobes:
  - call: <call_1>
  - call: <call_2>
  - call: <call_3>
```

Currently in tetragon we have an ability to create lists of calls and to have same selectors for every call in the list (https://tetragon.io/docs/concepts/tracing-policy/hooks/#lists), and it can solve the described problem as well, but if we want to have specific selectors for different calls in the list besides common selectors, it is not sufficient.

For example, we want to intercept system calls `open` and `openat`, which have different signatures: `open` system call has `pathname` as first argument, and `openat` has `pathname` as second argument. We want to catch event when `pathname` ends with `passwd`, but we want to filter out events where binary has prefix `/opt/myapp`, because we consider that our application can read `passwd` file safely. Because of the fact that these system calls signatures are different, we cannot create a list and have same `matchArgs` selector, because indices of `pathname` are different.

With global selectors, the policy would look like this:

```
spec:
  globalSelectors:
    matchBinaries:
      - operator: "NotPrefix"
        values: "/opt/myapp"
  kprobes:
  - call: "sys_open"
    args:
    - index: 0
      type: "string"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Postfix"
        values:
        - "passwd"
  - call: "sys_openat"
    args:
    - index: 1
      type: "string"
    selectors:
    - matchArgs:
      - index: 1
        operator: "Postfix"
        values:
        - "passwd" 
```

To solve the problem and to add such global selectors, we could add global selectors field `GlobalSelectors` to [TracingPolicySpec](https://github.com/cilium/tetragon/blob/main/pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go#L76) with type [KProbeSelector](https://github.com/cilium/tetragon/blob/main/pkg/k8s/apis/cilium.io/v1alpha1/types.go#L131).

Every time we parse selectors from the specification, we will append to currently parsed selectors selectors from `GlobalSelectors` field, and parse them as if they were specified together in same field. E.g.:

```
func ParseMatchArgs(k *KernelSelectorState, matchArgs []v1alpha1.ArgSelector, matchData []v1alpha1.ArgSelector,
	args []v1alpha1.KProbeArg, data []v1alpha1.KProbeArg, globalSelectors *KProbeSelector) error {

  matchArgs = append(matchArgs, globalSelectors.MatchArgs)
  matchData = append(matchData, globalSelectors.MatchData)
  ...
}
```

### Code of Conduct

kobrineli:
Fixes #4323

### Description

Sometimes in tetragon policies we need to write same selectors for each hook we want to intercept and report. For example, we may want to filter out all events, where process binary has specific prefix, e.g. we consider that all events from some directory are false positives.

Consider we have tracing policy, where we want to intercept kernel functions `call_1`, `call_2`, `call_3`, but target binary must not be located in directory `dir`. For that we define the following policy:
    
 ```yaml
 spec:
   kprobes:
   - call: <call_1>
     selectors:
     - matchBinaries:
       - operator: "NotPrefix"
         values:
         - "dir"
   - call: <call_2>
     selectors:
     - matchBinaries:
       - operator: "NotPrefix"
         values:
         - "dir"
   - call: <call_3>
     selectors:
     - matchBinaries:
       - operator: "NotPrefix"
         values:
         - "dir"
 ```

With macros, we can rewrite this policy in more convenient way:

```yaml
spec:
  selectorsMacros:
    dir:
      matchBinaries:
      - operator: "NotPrefix"
        values:
        - "dir"
  kprobes:
  - call: <call_1>
    selectors:
    - marcos: [ dir ]
  - call: <call_2>
    selectors:
    - marcos: [ dir ]
  - call: <call_3>
    selectors:
    - marcos: [ dir ]
```

For more real example, consider we want to intercept system calls `open` and `openat`, which have different signatures:
`open` system call has `pathname` as a first argument, and `openat` has `pathname` as a second argument. We want to catch
event when `pathname` ends with `passwd`, but we want to filter out events where binary has prefix `/opt/myapp`, because
we consider that our application can read `passwd` file safely. Because these system calls signatures are different, we
cannot create a list and have same `matchArgs` selector, because indices of `pathname` are different.

With macros, the policy would look like this:

```yaml
spec:
  selectorsMacros:
    myappExclusion:
      matchBinaries:
      - operator: "NotPrefix"
        values: "/opt/myapp"
  kprobes:
  - call: "sys_open"
    args:
    - index: 0
      type: "string"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Postfix"
        values:
        - "passwd"
      macros: ["myappExclusion"]
  - call: "sys_openat"
    args:
    - index: 1
      type: "string"
    selectors:
    - matchArgs:
      - index: 1
        operator: "Postfix"
        values:
        - "passwd"
      macros: ["myappExclusion"]
```

### Changelog

```release-note
selectors: add selectors macros syntactic sugar
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -1062,3 +1062,164 @@ spec:
         values:
         - "list:dups"
 ```
+
+## Selectors macros
+
+It's possible to define selectors macros in `TracingPolicy` specification that
+contain selectors definitions, which can be used in `kprobes`, `tracepoints`,
+`uprobes`, `usdts`, and `lsmhooks` as a part of their selectors by macro name.
+The content of the macro will be  substituted in target selectors.
+
+{{< caution >}}
+The same field cannot be present in a macro definition and in a policy selector
+that uses the macro. See more info in [macros
+limitations](/docs/concepts/tracing-policy/hooks/#macros-limitations).
+{{< /caution >}}
+
+Consider we have tracing policy, where we want to intercept kernel functions
+`call_1`, `call_2`, `call_3`, but target binary must not be located in
+directory `dir`. For that we define the following policy:
+    
+ ```yaml
+ spec:
+   kprobes:
+   - call: <call_1>
+     selectors:
+     - matchBinaries:
+       - operator: "NotPrefix"
+         values:
+         - "dir"
+   - call: <call_2>
+     selectors:
+     - matchBinaries:
+       - operator: "NotPrefix"
+         values:
+         - "dir"
+   - call: <call_3>
+     selectors:
+     - matchBinaries:
+       - operator: "NotPrefix"
+         values:
+         - "dir"
+ ```
+
+With macros, we can rewrite this policy in more convenient way:
+
+```yaml
+spec:
+  selectorsMacros:
+    dir:
+      matchBinaries:
+      - operator: "NotPrefix"
+        values:
+        - "dir"
+  kprobes:
+  - call: <call_1>
+    selectors:
+    - macros: [ dir ]
+  - call: <call_2>
+    selectors:
+    - macros: [ dir ]
+  - call: <call_3>
+    selectors:
+    - macros: [ dir ]
+```
+
+For a more realistic example, consider we want to intercept system calls `open`
+and `openat`, which have different signatures: `open` system call has
+`pathname` as a first argument, and `openat` has `pathname` as a second
+argument. We want to catch event when `pathname` ends with `passwd`, but we
+want to filter out events where binary has prefix `/opt/myapp`, because we
+consider that our application can read `passwd` file safely. Because these
+system calls signatures are different, we cannot create a list and have same
+`matchArgs` selector, because indices of `pathname` are different.
+
+With macros, the policy would look like this:
+
+```yaml
+spec:
+  selectorsMacros:
+    myappExclusion:
+      matchBinaries:
+      - operator: "NotPrefix"
+        values: "/opt/myapp"
+  kprobes:
+  - call: "sys_open"
+    args:
+    - index: 0
+      type: "string"
+    selectors:
+    - matchArgs:
+      - index: 0
+        operator: "Postfix"
+        values:
+        - "passwd"
+      macros: ["myappExclusion"]
+  - call: "sys_openat"
+    args:
+    - index: 1
+      type: "string"
+    selectors:
+    - matchArgs:
+      - index: 1
+        operator: "Postfix"
+        values:
+        - "passwd"
+      macros: ["myappExclusion"]
+```
+
+### Limitations {#macros-limitations}
+
+Having a same field in a hook selector and in a macro definition, or using
+different macros with same fields defined will result in an error.
+
+The following policy will be considered as **incorrect**, because there is
+`matchArgs` field in both macro and policy selector:
+```yaml
+spec:
+  selectorsMacros:
+    argMacro:
+      matchArgs:
+        - index: 0
+          operator: "NotEqual"
+          values:
+            - <value-1>
+  kprobes:
+  - call: <call>
+    args:
+      - index: 0
+        type: int
+    selectors:
+    - macros: [ argMacro ]
+      matchArgs:
+        - index: 0
+          operator: "NotEqual"
+          values:
+            - <value-1>
+```
+
+The following policy will be considered as **incorrect**, because different
+macros contain `matchArgs` field and used in same selector:
+```yaml
+spec:
+  selectorsMacros:
+    argMacro1:
+      matchArgs:
+        - index: 0
+          operator: "NotEqual"
+          values:
+            - <value-1>
+    argMacro2:
+      matchArgs:
+        - index: 0
+          operator: "NotEqual"
+          values:
+            - <value-2>
+  kprobes:
+  - call: <call>
+    args:
+      - index: 0
+        type: int
+    selectors:
+    - macros: [ argMacro1, argMacro2 ]
+```


kobrineli:
@kkourt @mtardy 
Hi! I've implemented selectors macros which we discussed in https://github.com/cilium/tetragon/issues/4323

Cannot add release note label by myself

kobrineli:
@kkourt @mtardy 
Hi! Could you check this PR, please?

pkg/sensors/tracing/generic.go
```diff
@@ -173,3 +174,34 @@ func pathArgWarning(index uint32, ty int, s []v1alpha1.KProbeSelector) {
...
+				return fmt.Errorf("macro '%s' is used in selector, but no macros were defined in policy spec", macroName)
+			}
+			macro, ok := macros[macroName]
+			if !ok {
+				return fmt.Errorf("undefined macro '%s'", macroName)
+			}
+			if len(macro.Macros) > 0 {
+				return errors.New("macro definition cannot use other macros")
+			}
+			selector.MatchPIDs = append(selector.MatchPIDs, macro.MatchPIDs...)
```
kkourt: I wonder if append() is the right thing to do here because adding multiple selectors might yield unexpected results.

Alternatively, we could return an error if a field is defined in both the macro and the policy selectors.

```go
if len(macro.macro.MatchPIDs) > 0 {
    if (len(selector.MatchPIDs) > 0) {
       return errors.New("") 
   }
}
```

Thoughts?

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -162,6 +162,10 @@ type KProbeSelector struct {
 	// +kubebuilder:validation:Optional
 	// IDs for capabilities changes
 	MatchCapabilityChanges []CapabilitiesSelector `json:"matchCapabilityChanges,omitempty"`
+	// +kubbuilder:validation:Optional
```
mtardy: there's a typo in `kubebuilder` that will make the validation not happen!

mtardy: you also have typos in the example of your commit description `marcos`->`macros` that you might want to fix.

docs/content/en/docs/reference/tracing-policy.md
mtardy, line 1: please bundle this with the 2nd commit since you separate the generated stuff from the changes of the types from commit 1.

pkg/sensors/tracing/generic.go
```diff
@@ -173,3 +174,34 @@ func pathArgWarning(index uint32, ty int, s []v1alpha1.KProbeSelector) {
 			name, index))
 	}
 }
+
+func appendMacrosSelectors(selectors []v1alpha1.KProbeSelector, macros map[string]v1alpha1.KProbeSelector) error {
+	for i := range selectors {
+		selector := &selectors[i]
+		for _, macroName := range selector.Macros {
+			if macros == nil {
+				return fmt.Errorf("macro '%s' is used in selector, but no macros were defined in policy spec", macroName)
```
mtardy: might be worth also putting that message when macro is not nil but an empty map (`len(0)`) otherwise you will just bump into the check just after, which is okay also!

pkg/sensors/tracing/genericlsm.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/cilium/ebpf"
 
 	"github.com/cilium/tetragon/pkg/api/ops"
-	processapi "github.com/cilium/tetragon/pkg/api/processapi"
+	"github.com/cilium/tetragon/pkg/api/processapi"
```
mtardy, line 21: I agree with the formatting change but maybe to bundle with another commit or PR?

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -1017,3 +1017,100 @@ spec:
...
+  selectorsMacros:
+    dir:
+      matchBinaries:
+      - operator: "NotPrefix"
+        values:
+        - "dir"
+  kprobes:
+  - call: <call_1>
+    selectors:
+    - marcos: [ dir ]
```
mtardy: small typo here as well

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -1017,3 +1017,100 @@ spec:
...
+
+## Selectors macros
+
+It's possible to define selectors macros in `TracingPolicy` specification that contain selectors definitions, which can
+be used in `kprobes`, `tracepoints`, `uprobes`, `usdts`, and `lsmhooks` as a part of their selectors by macro name.
+The content of the macro will be substituted in target selectors, and all filters will be appended to corresponding ones.
+
+Consider we have tracing policy, where we want to intercept kernel
+functions `call_1`, `call_2`, `call_3`, but target binary must not be
+located in directory `dir`. For that we define the following policy:
```
mtardy: nit: could you wrap consistently? in `vim` you can use `gq` after selecting the text to wrap around 80 chars

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -1017,3 +1017,100 @@ spec:
...
+    - marcos: [ dir ]
+  - call: <call_2>
+    selectors:
+    - marcos: [ dir ]
+  - call: <call_3>
+    selectors:
+    - marcos: [ dir ]
+```
+
+For more real example, consider we want to intercept system calls `open` and `openat`, which have different signatures:
```
mtardy: maybe "For a more realistic example"?

pkg/sensors/tracing/generic.go
kobrineli: I considered that having macros filters appended to selectors filters was part of the idea, but we could add this check.
But I think appending can be useful if you have multiple use cases of macros in a policy, and in one of them you also want to have some additional filters of same type.

kobrineli:
> Hey, looks good, here's some comments. Thanks for the PR!
> 
> I know this might sound a bit negative and weird after the good feedback from the issue but the more I see this the more I doubt it's worth adding. Let's see what other people think! Not strongly against but not strongly in favor.

Maybe we could discuss it on 9 February on the community meeting?
At least, having such macros would be very useful personally for me, because for some policies with multiple hooks I have same binary exclusions for every hook in policy, so being able to use named macro would really simplify writing such policies.

pkg/sensors/tracing/generic.go
kobrineli: changed check for `len(macros) == 0` only, because it covers both cases when `macros` is nil and when `macros` is not nil but empty

pkg/sensors/tracing/genericlsm.go
kobrineli: agree, reverted it in current PR

pkg/sensors/tracing/generic.go
kkourt: Could you provide an example where appending would be useful?

kobrineli: Cannot come up with something quickly, just thought that such semantics would make macros more flexible to use. For my use cases, described in issue, it's not so important, so I'm okay with not appending macro to filters.

So do I understand correctly, that using two macros in selector, when same filters are defined in both macros, should also result in an error?

kobrineli: Added check whether field definition is exclusive for each selector

kobrineli:
@kkourt
Hi! I've added check for exclusive field definition in policy selectors and macro as we discussed.
Also added limitations section in docs with examples of incorrect policies.

pkg/sensors/tracing/generic.go
```diff
@@ -173,3 +175,91 @@ func pathArgWarning(index uint32, ty int, s []v1alpha1.KProbeSelector) {
...
+			}
+		}
+	}
+
+	return nil
+}
+
+func useMacro[T any](filters []T, macrosFilters []T) ([]T, error) {
+	if len(filters) > 0 && len(macrosFilters) > 0 {
+		return nil, errors.New("field is defined in both macro and policy selector")
```
kkourt: nit: I think we will hit this error even if the same selector is in multiple macros. So maybe the error should be something like:
```
errors.New("field is defined in multiple macros and/or policy selectors")
```

nit^2:
You could even have something like:
```
return fmt.Errorf("%T:  field is defined in multiple macros and/or policy selectors", filters[0])
```

To avoid the fmt.Errorf() in the caller.

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -1017,3 +1017,166 @@ spec:
...
+
+## Selectors macros
+
+It's possible to define selectors macros in `TracingPolicy` specification that
+contain selectors definitions, which can be used in `kprobes`, `tracepoints`,
+`uprobes`, `usdts`, and `lsmhooks` as a part of their selectors by macro name.
+The content of the macro will be  substituted in target selectors.
+
+{{< warning >}}
+Same field cannot be present in a macro definition and in a policy selector
```
kkourt: nit: "The same field"

kkourt (approved):
Thanks! LGTM!

pkg/sensors/tracing/generic.go
kobrineli: the problem with %T is that we would have `v1alpha1.ArgSelector`, for example, for both `MatchArgs` and `MatchReturnArgs`, but I think this is ok, because it is actual only for few fields

kobrineli:
CI is failing on job which never failed before, cannot restart it myself

kobrineli:
@kkourt 
Hi! What blocks this PR from being merged?
(except of failed packages e2e tests CI job, could you restart it, please?..)

kkourt:
> Hi! What blocks this PR from being merged?
> (except of failed packages e2e tests CI job, could you restart it, please?..)

I was waiting for @mtardy to take a look.
The CI failure is a known issue -- I don't think it's worth restarting.

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -1062,3 +1062,166 @@ spec:
...
+
+It's possible to define selectors macros in `TracingPolicy` specification that
+contain selectors definitions, which can be used in `kprobes`, `tracepoints`,
+`uprobes`, `usdts`, and `lsmhooks` as a part of their selectors by macro name.
+The content of the macro will be  substituted in target selectors.
+
+{{< warning >}}
+The same field cannot be present in a macro definition and in a policy selector
+that uses the macro.  More info
+[here](/docs/concepts/tracing-policy/hooks/#macros-limitations).
```
mtardy: nit: I would replace "here" with the actual content like "See more info in [macros limitations](...)"

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -1062,3 +1062,166 @@ spec:
...
 ```
+
+## Selectors macros
+
+It's possible to define selectors macros in `TracingPolicy` specification that
+contain selectors definitions, which can be used in `kprobes`, `tracepoints`,
+`uprobes`, `usdts`, and `lsmhooks` as a part of their selectors by macro name.
+The content of the macro will be  substituted in target selectors.
+
+{{< warning >}}
```
mtardy: nit: this is a bit personal appreciation but I feel it's more a caution than a warning (warning is for kinda "dangerous" stuff)

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -1062,3 +1062,166 @@ spec:
...
+```yaml
+spec:
+  selectorsMacros:
+    argMacro:
+      matchArgs:
+        - index: 0
+          operator: "NotEqual"
+          values:
+            - <value-1>
+
```
mtardy: is the empty line on purpose on both examples?

mtardy:
> > Hi! What blocks this PR from being merged?
> > (except of failed packages e2e tests CI job, could you restart it, please?..)
> 
> I was waiting for @mtardy to take a look. The CI failure is a known issue -- I don't think it's worth restarting.

I opened https://github.com/cilium/tetragon/issues/4693.

docs/content/en/docs/concepts/tracing-policy/hooks.md
kobrineli: no, just by accident. I'll remove

mtardy:
Could you update the changelog in the first message to add a short sentence to explain this new feature? thanks!

mtardy:
sorry somehow it fails to merge with 
image

I guess it's just a transient issue at Github??

pkg/k8s/apis/cilium.io/v1alpha1/version.go
```diff
@@ -7,4 +7,4 @@ package v1alpha1
 // Used to determine if CRD needs to be updated in cluster
 //
 // Developers: Bump patch for each change in the CRD schema.
-const CustomResourceDefinitionSchemaVersion = "1.7.9"
+const CustomResourceDefinitionSchemaVersion = "1.7.10"
```
AritraDey-Dev: Hey, since https://github.com/cilium/tetragon/pull/4609 was recently merged, I think you need to bump the version. I believe that might also be the reason @mtardy mentioned here https://github.com/cilium/tetragon/pull/4548#issuecomment-3958929866 . thanks

mtardy: ah yeah, weird that we don't see the usual "conflict" thing, but might be that thanks, let's rebase @kobrineli ?

kobrineli: thanks! rebased, waiting for ci

kobrineli: @mtardy 
I've rebased, CI finally passed

Issue #709, opened 2023-02-21
Test and document the limitations of the fields in `TracingPolicy`


mtardy:
The documentation on `TracingPolicy` limitation might be outdated. It would be interesting to update the documentation and add tests case to make sure the limitations are prevented from the agent side instead of failing silently or creating an error on the BPF side.

I see two ways of doing that:
- experimentally, trying to break TracingPolicy and then someone can help to link to the BPF limitation.
- reading the BPF code to retrieve the limits.

Limitations include for example:
- max numbers of selectors
- max PID values for `matchPIDs`
- max number of `matchArgs` per selector
- max values for `matchBinaries`
- etc.
```[tasklist]
### Tasks
- [ ] Perform better user space TracingPolicy limitation validation
- [ ] Faster: annotate the doc to mention it is outdated
- [ ] Better: update the documentation
```

kevsecurity:
filter_file_buf() in basic.h limits the number of file matches to MAX_MATCH_FILE_VALUES == 2.

kevsecurity:
Postfix file match limited to 50 chars on kernels >=5.2 and 40 on kernels <5.2.

kevsecurity:
File match equal and prefix limited to 128 chars on kernels >=5.2 and 32 on kernels <5.2.

kevsecurity:
#1408 will alleviate some of these limitations. e.g. number of matches becomes unbounded; equal and prefix lengths become 128 on all kernels; and postfix length becomes 128 characters on kernels >=5.2 and 96 on kernels <5.2.

jrfastab:
This is old. File specific issues are refile with up to date info now.

mtardy:
well, it was the point of this issue because documentation is currently outdated.

PR #4549, opened 2026-01-20
docs: update TracingPolicy selector limitations

puwun:
### Description
Update the Limitations section in the selectors documentation:

- Remove outdated warning banner referencing  #709
- Correct `Max Selectors` from 8 to 5 (per MAX_SELECTORS in `basic.h`)
- Correct `Max MatchArg Values` from 1 to 4 (per MAX_MATCH_VALUES in `basic.h`)
- Add note about `InMap/NotInMap` operators for unlimited values

Fixes: #709

### Changelog

```release-note
docs: Fix outdated TracingPolicy selector limitations (Max Selectors: 8 -> 5, Max MatchArg Values: 1 -> 4)
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -2044,18 +2044,17 @@ The above would be executed in kernel as:
 
 ### Limitations
 
-{{% pageinfo %}}
-Those limitations might be outdated, see [issue #709](https://github.com/cilium/tetragon/issues/709).
-{{% /pageinfo %}}
-
 Because BPF must be bounded we have to place limits on how many selectors can
 exist.
 
-- Max Selectors 8.
+- Max Selectors 5.
 - Max PID values per selector 4
 - Max MatchArgs per selector 5 (one per index)
-- Max MatchArg Values per MatchArgs 1 (limiting initial implementation can bump
-  to 16 or so)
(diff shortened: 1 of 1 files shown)


mtardy (approved):
Thanks! Technically it does not fixes completely the issue since we might have other undocumented limitation but let's close it.

mtardy:
If you can document this part, if it's not already documented that would be fantastic, otherwise it's already a nice improvement. Thanks!

image

puwun:
@mtardy no worries
Just making sure I've got it correct, the changes to be made are:
- Max file match values: 8 on kernels ≥ 5.3, 2 on kernels < 5.3
- String prefix max length: 256 chars (all kernels, after  #1408)
- String postfix max length: 128 chars (all kernels, after  #1408)

If it helps, I can open a follow up PR just to add these kernel version-specific limits  ?

mtardy:
> @mtardy no worries Just making sure I've got it correct, the changes to be made are:
> 
> * Max file match values: 8 on kernels ≥ 5.3, 2 on kernels < 5.3
> * String prefix max length: 256 chars (all kernels, after  [Convert string and char_buf matches to hash look ups #1408](https://github.com/cilium/tetragon/pull/1408))
> * String postfix max length: 128 chars (all kernels, after  [Convert string and char_buf matches to hash look ups #1408](https://github.com/cilium/tetragon/pull/1408))
> 
> If it helps, I can open a follow up PR just to add these kernel version-specific limits ?

Yep something like that, reflecting at least everything that was discussed in this issue in the docs, or check if it's not already btw!

PR #4554, opened 2026-01-21
Add regex support for namespace filtering

AritraDey-Dev:
### Description

This PR adds regex-based namespace filtering for events. Previously, the export allow/deny list only supported exact namespace matches, which made managing multiple similarly named namespaces difficult.(please review commit-wise).

```
tetra getevents --namespace-regex 'prod-.*'
```

### Changelog

```release-note
Add regex support for namespace filtering in export allow/deny lists
```

### DIFF
--- a/pkg/filters/namespace_test.go
+++ b/pkg/filters/namespace_test.go
@@ -66,3 +66,102 @@ func TestNamespace(t *testing.T) {
 	ev = event.Event{Event: &tetragon.GetEventsResponse{Event: &tetragon.GetEventsResponse_ProcessExec{ProcessExec: &tetragon.ProcessExec{Process: &tetragon.Process{}}}}}
 	assert.True(t, fl.MatchOne(&ev))
 }
+
+func TestNamespaceRegex(t *testing.T) {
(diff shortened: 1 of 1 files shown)


AritraDey-Dev:
> would be easier to review if you split the generated changes to the separated commit

done.

pkg/filters/namespace_regex_test.go
mtardy, line 1: nit: maybe you can put this into the already existing `namespace_test.go`.

mtardy:
you need to satisfy the CI however, you might need to bundle the generated changes with the actual changes to make every commit to build. You'll need to generate docs etc. The output should be understandable

AritraDey-Dev:
>  you might need to bundle the generated changes with the actual changes to make every commit to build

Yes, for every commit to build, the generated changes should be bundled into a single commit. I didn’t think about that while splitting this.should I squash it back into one?

The other CI failure looks like a gitHub status issue; there seems to be some disruption according to https://www.githubstatus.com/ .

AritraDey-Dev:
Updated now! the generated code(api changes) into its own commit and kept the regex logic in a separate commit.

PR #4565, opened 2026-01-23
support extended selectors for uprobes

AritraDey-Dev:
### Description

This PR enables full selector support for uprobes in TracingPolicies, removing the restriction that only allowed matchPIDs. Users can now filter uprobe events by arguments (matchArgs), return values, namespaces, and more, matching the capabilities of kprobes.

Example:
```
spec:
  uprobes:
  - path: "/bin/bash"
    symbols: ["readline"]
    args: [{index: 0, type: "string"}]
    selectors:
    - matchArgs:     # <- Now supported!
      - {index: 0, operator: "Equal", values: ["secret"]}
```
### Changelog

```release-note
Added support for all selectors (like argument matching) in `uprobe` policies.
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -313,6 +313,50 @@ spec:
 This example shows how to use uprobes to hook into the readline function
 running in all the bash shells.
 
+### Selectors
+
+Uprobes support the full range of selectors available for filtering events,
+including `matchArgs`, `matchReturnArgs`, `matchBinaries`, `matchNamespaces`,
+`matchCapabilities`, and more. This allows you to precisely filter uprobe
+events based on function arguments, return values, or process context.
+
+Here is an example that only generates events when the `readline` function
+is called from a specific binary and returns a string starting with "sudo":
+
+```yaml
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "uprobe-with-selectors"
+spec:
+  uprobes:
+  - path: "/bin/bash"
+    symbols:
+    - "readline"
+    return: true
+    returnArg:
+      index: 0
+      type: "string"
+    selectors:
+    - matchBinaries:
+      - operator: "In"
+        values:
+        - "/bin/bash"
+        - "/usr/bin/bash"
+      matchReturnArgs:
+      - index: 0
+        operator: "Prefix"
+        values:
+        - "sudo"
+```
+
+This policy will only generate events when `readline` is called from `/bin/bash`
+and returns a string starting with "sudo", which can be useful for monitoring potentially
+sensitive command input.
+
+For more details on available selectors and their usage, see the
+[Selectors]({{< ref "/docs/concepts/tracing-policy/selectors" >}}) documentation.
+
 ## USDTs
 
 Tetragon allows to attach and monitor USDT (User Statically-Defined Tracing) probes.


pkg/sensors/tracing/genericuprobe.go
```diff
@@ -250,6 +246,100 @@ func loadSingleUprobeSensor(uprobeEntry *genericUprobe, args sensors.LoadProbeAr
...
+	if uprobeEntry != nil {
+		if load.RetProbe {
+			return uprobeEntry.loadArgs.selectors.retrn
+		}
+		return uprobeEntry.loadArgs.selectors.entry
+	}
+	return nil
+}
+
+func uprobeFilterMaps(load *program.Program, uprobeEntry *genericUprobe) []*program.Map {
```
olsajiri: so this is 3rd (AFAICS) function doing the same thing, I think we want to share some of that setup,
please check filterMapsForLsm@genericlsm.go filterMaps@generickprobe.go and try to share it

AritraDey-Dev: Hey thanks for bringing this up !,I will look into this.

AritraDey-Dev: So I think we can keep a shared implementation of this in `selectors.go `, and then use it in the relevant places.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -178,78 +177,8 @@ func getProgramSelector(load *program.Program, kprobeEntry *genericKprobe) *sele
...
-	if state != nil && !kernels.MinKernelVersion("5.9") {
-		// Versions before 5.9 do not allow inner maps to have different sizes.
-		// See: https://lore.kernel.org/bpf/20200828011800.1970018-1-kafai@fb.com/
-		maxEntries := state.StringPostfixMapsMaxEntries()
-		stringPostfixFilterMaps.SetInnerMaxEntries(maxEntries)
-	}
-	maps = append(maps, stringPostfixFilterMaps)
-
-	return maps
+	return createSelectorMaps(load, state)
```
olsajiri: let's remove filterMaps and call createSelectorMaps directly like we do in tracepoint sensor?

pkg/sensors/tracing/genericlsm.go
```diff
@@ -551,69 +542,5 @@ func createLsmSensorFromEntry(polInfo *policyInfo, lsmEntry *genericLsm,
...
-	if !kernels.MinKernelVersion("5.9") {
-		// Versions before 5.9 do not allow inner maps to have different sizes.
-		// See: https://lore.kernel.org/bpf/20200828011800.1970018-1-kafai@fb.com/
-		maxEntries := lsmEntry.selectors.StringPostfixMapsMaxEntries()
-		stringPostfixFilterMaps.SetInnerMaxEntries(maxEntries)
-	}
-	maps = append(maps, stringPostfixFilterMaps)
-
-	return maps
+	return createSelectorMaps(load, lsmEntry.selectors)
```
olsajiri: let's remove filterMapsForLsm and call createSelectorMaps directly?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -245,6 +245,21 @@ func loadSingleUprobeSensor(uprobeEntry *genericUprobe, args sensors.LoadProbeAr
...
+	state := getUprobeProgramSelector(load, uprobeEntry)
+	return createSelectorMaps(load, state)
+}
```
olsajiri: same, I don't think we need uprobeFilterMaps function, let's call createSelectorMaps directly

pkg/sensors/tracing/generickprobe.go
AritraDey-Dev: Sure!

olsajiri (approved):
thanks

olsajiri:
@AritraDey-Dev thanks for the doc, the selector code changed, please rebase

PR #4566, opened 2026-01-25
chore: Add sloglint in golangci-lint

sayboras:
The migration from logrus to slog is done recently, add linter for consistency in the future.

```release-note
chore: Add sloglint in golangci-lint
```

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -23,6 +23,7 @@ linters:
     - modernize
     - perfsprint
     - revive
+    - sloglint
     - staticcheck
     - testifylint
     - unused


.golangci.yml
```diff
@@ -6,6 +6,7 @@ output:
 linters:
   default: none
   enable:
+    - copyloopvar
```
mtardy: you also added that one in the same commit, is it expected?

sayboras: ah sorry, I messed up the commits :face_exhaling:

PR #4568, opened 2026-01-25
procsyms: Implement caching for process memory maps

AritraDey-Dev:
### Description

This PR implements a TODO to make symbol resolution much faster by caching memory maps.

Right now, we read the `/proc/[pid]/maps` file every time we need to find a symbol. This is slow because it involves reading from the disk over and over again.

Implemented a cache that stores the memory maps for each process ID so we don't have to read the file every time. The cache automatically clears itself when a process dies or starts a new program (execve) to ensure we always have the correct data. (split this into multiple commits, so please review them commit-wise.) 

**Output of benchmark test**(commit https://github.com/cilium/tetragon/commit/1be92a42bf7e33e2df65d9e24471c14ef5f4fb58)

**With this patch**
```
go test -bench=. -benchmem -count=1 ./pkg/procsyms/...
goos: linux
goarch: amd64
pkg: github.com/cilium/tetragon/pkg/procsyms
cpu: 13th Gen Intel(R) Core(TM) i5-13420H
BenchmarkGetProcMaps-12         41458305                27.40 ns/op            0 B/op          0 allocs/op
PASS
ok      github.com/cilium/tetragon/pkg/procsyms 2.089s
```
**without this patch**
```
go test -bench=. -benchmem -count=1 ./pkg/procsyms/...
goos: linux
goarch: amd64
pkg: github.com/cilium/tetragon/pkg/procsyms
cpu: 13th Gen Intel(R) Core(TM) i5-13420H
BenchmarkGetProcMaps-12    
   22100             58114 ns/op           33469 B/op        499 allocs/op
PASS
ok      github.com/cilium/tetragon/pkg/procsyms 2.949s
```

### Changelog

```release-note
Made symbol lookups faster by caching memory maps.
```

### DIFF
--- a/pkg/sensors/exec/exec_linux.go
+++ b/pkg/sensors/exec/exec_linux.go
@@ -23,6 +23,7 @@ import (
 	"github.com/cilium/tetragon/pkg/observer"
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/process"
+	"github.com/cilium/tetragon/pkg/procsyms"
 	"github.com/cilium/tetragon/pkg/sensors"
 	"github.com/cilium/tetragon/pkg/sensors/exec/procevents"
 	"github.com/cilium/tetragon/pkg/sensors/exec/userinfo"
@@ -210,6 +211,8 @@ func handleExecve(r *bytes.Reader) ([]observer.Event, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Invalidate cached memory maps since exec replaces the address space
+	procsyms.InvalidatePID(int(m.Parent.Pid))
 	msgUnix := msgToExecveUnix(&m)
 	msgUnix.Unix.Process, err = execParse(r)
 	if err == nil {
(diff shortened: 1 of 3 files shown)


mtardy (changes requested):
Hey 👋  yep so in general we had the tendency to do the contrary: remove caching in favor of re-reading stuff (for example we do that for `/proc/kallsyms`). The reason is that for Tetragon it's not really a problem to take ~50 more micro seconds if it can save memory that will reside in heap for a longer lifetime.

So thanks a lot for opening that PR, it looks pretty good but on the principle I'm not sure we would like that. Sorry :(

Especially maintaining a cache of entries related to `procs` as been proven to be extremely difficult and error prone because `exit`s don't exactly match `exec`s for reasons so it has a tendency to leak over time and needs special GC... it's a bit of a pain.

pkg/sensors/exec/exec_linux.go
```diff
@@ -210,6 +211,8 @@ func handleExecve(r *bytes.Reader) ([]observer.Event, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Invalidate cached memory maps since exec replaces the address space
+	procsyms.InvalidatePID(int(m.Parent.Pid))
```
olsajiri, line 215: hum, how about other ways of maps being changes during the life of the process? like dlopen-ed libraries, I'm not sure we could account for that in an easy way

kkourt:
The PR has remained idle for some time. Moving it to draft. @AritraDey-Dev if you plan to continue working on it, please undraft it. thanks!

PR #4569, opened 2026-01-26
feat : Add support to resolve attributes in kernel modules 

tdaudi:
Enhance the resolve to allow resolving module hooks.
Currently if Tetragon support hooking into kernel modules, it is not allowed to resolve any fields because the BTF file used contains the information of the kernel, not the module. In order to load the appropriate file, we search for the `kallsyms` if the symbol is a module, and if it, we load the appropriate BTF file.

### DIFF
--- a/pkg/btf/btf_test.go
+++ b/pkg/btf/btf_test.go
@@ -252,6 +252,19 @@ func TestFindBTFFuncParamFromHook(t *testing.T) {
 	for _, btfFile := range btfFiles {
 		t.Run(btfFile, testFindBTFFuncParamFromHook(btfFile))
 	}
+
+	t.Run("ResolveModuleIsSupported", func(t *testing.T) {
+		module := "overlay"
+		hook := "ovl_create"
+
+		// Check if BTF is available for the module
+		if _, err := os.Stat("/sys/kernel/btf/" + module); err != nil {
+			t.Skipf("BTF for module %s not found", module)
+		}
+
+		_, err := FindBTFFuncParamFromHook(hook, 0)
+		require.NoError(t, err)
+	})
 }
 
 func fatalOnError(t *testing.T, err error) {


pkg/btf/btf.go
```diff
@@ -156,9 +157,24 @@ func firstTypeByName(spec *btf.Spec, name string, typ any) error {
 }
 
 func FindBTFFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
-	spec, err := NewBTF()
+	var spec *btf.Spec
+
+	ks, err := ksyms.KernelSymbols()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("fail to load ksyms.KernelSymbols: %w", err)
```
FedeDP: Do we want to error out here? Or it would be better to fallback at the current method?
I mean, i'd avoid being in a situation where, because `ksyms.KernelSymbols()` fails, we lose support for current behavior.

tdaudi: Good question ! I basically copy/paste the code from here 
https://github.com/cilium/tetragon/blob/94f1e6be16825d63710e2231328f7fc5fb6789cc/pkg/btf/validation.go#L58-L69

So I think if we reach that point, it's already safe to load it a second time.

FedeDP: I agree, still i'd use something like:
```go
var (
  spec *btf.Spec
  err error
)

if ks, err := ksyms.KernelSymbols(); err == nil {
  if kmod, err := ks.GetKmod(hook); err == nil {
		spec, err = btf.LoadKernelModuleSpec(kmod)
  }
}
if spec == nil {
      spec, err = NewBTF()
}
if err != nil {
	return nil, err
}
```

FedeDP: but am not sure :)

tdaudi: Why not ! It's shorter and more stable so let's do it.
Thanks !

pkg/btf/btf.go
```diff
@@ -156,7 +157,18 @@ func firstTypeByName(spec *btf.Spec, name string, typ any) error {
...
 
 func FindBTFFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
-	spec, err := NewBTF()
+	var (
+		spec *btf.Spec
+		err  error
+	)
+	if ks, err := ksyms.KernelSymbols(); err == nil {
+		if kmod, err := ks.GetKmod(hook); err == nil {
+			spec, _ = btf.LoadKernelModuleSpec(kmod)
```
FedeDP:
```suggestion
			spec, err = btf.LoadKernelModuleSpec(kmod)
```
to propagate `err` to the check at L172

tdaudi: My point was, if err != nil there, spec == nil, so it will fall into the `newBTF` code. So this `err` is never used

tdaudi: But ofc I don't mind adding it anyway

tdaudi: Done ! :)

FedeDP: oh makes sense, didn't see that! Thanks

tdaudi: It seems the CI complains about the change
```
  pkg/btf/btf.go:166:10  ineffassign  ineffectual assignment to err
  pkg/btf/btf.go:166:4   staticcheck  SA4006: this value of err is never used
```

Though, there is one behavior that may cause a crash is the `err !=nil` but with a spec defined. It will go neither in the following if.

FedeDP: Weird, shouldn't it be used later? :/

tdaudi: Oh oups my bad, it's fine.
Maybe the static test just check LoadKernelModuleSpec never return a spec, a stop the processing after the `step == nil` branch. But I think it's safe to ignore.
Thanks :)

FedeDP: @mtardy perhaps has more hints!

mtardy: mmh indeed the code is hard to parse for a human here tbh, in some cases the value on line 166 is indeed never used because you may override it at `spec, err = NewBTF()` and the general rule of thumbs is to always error check before reading another return param. Also in general in Go it's not a good practice to assume that if `err != nil` then the other parameters are nil or initialized at 0 and the linter might just assume this without knowing what this function actually returns.

Could you rewrite this by using two differents variables name for errors so that it's just easier to read? 

Also personally not a big fan to use this construct in functions as well, we generally only use it in files global scope.
```
var (
  stuff
  stuff
)
```

```go
	var (
		spec *btf.Spec
		err  error
	)
	if ks, err := ksyms.KernelSymbols(); err == nil {
		if kmod, err := ks.GetKmod(hook); err == nil {
			spec, err = btf.LoadKernelModuleSpec(kmod)
		}
	}
	if spec == nil {
		spec, err = NewBTF()
	}
	if err != nil {
		return nil, err
	}
```

pkg/btf/btf.go
```diff
@@ -156,7 +157,18 @@ func firstTypeByName(spec *btf.Spec, name string, typ any) error {
...
+		err  error
+	)
+	if ks, err := ksyms.KernelSymbols(); err == nil {
+		if kmod, err := ks.GetKmod(hook); err == nil {
+			spec, err = btf.LoadKernelModuleSpec(kmod)
+		}
+	}
+	if spec == nil {
+		spec, err = NewBTF()
+	}
```
mtardy: if spec is nil which might be the case if it returned an `err` (I haven't checked the function), then the `err` value on line 166 is never read and overwritten at line 170 to be checked at line 172. Is it expected? I'm sure we can simplify this!

tdaudi: Thank you for the review. 

Initially I skipped the error. (see https://github.com/cilium/tetragon/pull/4569#discussion_r2727764120)
But I'm not sure what changes you expect beyond this

```diff
diff --git a/pkg/btf/btf.go b/pkg/btf/btf.go
index b3a559007..690f26ae0 100644
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -157,10 +157,9 @@ func firstTypeByName(spec *btf.Spec, name string, typ any) error {
 }
 
 func FindBTFFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
-       var (
-               spec *btf.Spec
-               err  error
-       )
+       var spec *btf.Spec
+       var err error
+
```

mtardy: would something like that convey what you want to do?

```go
func FindBTFFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
    // Try module BTF
    if ks, err := ksyms.KernelSymbols(); err == nil {
        if kmod, err := ks.GetKmod(hook); err == nil {
            if spec, err := btf.LoadKernelModuleSpec(kmod); err == nil {
                return findBTFFuncParamFromHookWithSpec(spec, hook, argIndex)
            }
        }
    }

    // In case of failure, default to kernel BTF
    spec, err := NewBTF()
    if err != nil {
        return nil, err // or wrap the error with fmt.Errorf
    }
    return findBTFFuncParamFromHookWithSpec(spec, hook, argIndex)
}
```

If you want to add the info that you tried the kernel module BTF you can also assign some kind of err on the final failed path and append it to the error you eventually might return.

mtardy: and yes sorry I didn't even read fully the thread but you said exactly that here lol https://github.com/cilium/tetragon/pull/4569#discussion_r2727772458

tdaudi: Perfect thanks ! 

> and yes sorry I didn't even read fully the thread but you said exactly that here lol https://github.com/cilium/tetragon/pull/4569#discussion_r2727772458

No worries :)

pkg/btf/btf.go
```diff
@@ -156,6 +157,15 @@ func firstTypeByName(spec *btf.Spec, name string, typ any) error {
 }
 
 func FindBTFFuncParamFromHook(hook string, argIndex int) (*btf.FuncParam, error) {
+	// If the hook is part of a kernel module, load its BTF file
+	if ks, err := ksyms.KernelSymbols(); err == nil {
+		if kmod, err := ks.GetKmod(hook); err == nil {
+			if spec, err := btf.LoadKernelModuleSpec(kmod); err == nil {
+				return findBTFFuncParamFromHookWithSpec(spec, hook, argIndex)
+			}
```
olsajiri, line 166: should we do `return nil, err` in else path inhere?

tdaudi: As suggested here https://github.com/cilium/tetragon/pull/4569#discussion_r2727108244, the `NewBTF` is load by default if the `GetKmod` nested ifs fails. Do you think we should not do that ?

olsajiri: hm, so on line 163 we already now the function/hook is in module, right? how will the fallback to kernel btf help with that? seems like we just make findBTFFuncParamFromHookWithSpec later on to fail instead? probably with misleading error ... but I might be missing something

tdaudi:
@olsajiri I assigned your change request, but the CI seems to fail due to an unrelated problem. Should we ignore it or I rebase and rerun the tests ?

olsajiri:
> @olsajiri I assigned your change request, but the CI seems to fail due to an unrelated problem. Should we ignore it or I rebase and rerun the tests ?

I think it's unrelated, but let's rebase

PR #4583, opened 2026-01-29
gRPC info

kkourt:
Add a gRPC call to get information about the agent and its environment.

### DIFF
--- a/cmd/tetra/info/info.go
+++ b/cmd/tetra/info/info.go
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Hubble
+
+package info
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/cilium/tetragon/cmd/tetra/common"
+	"github.com/cilium/tetragon/pkg/tetragoninfo"
+
+	"github.com/spf13/cobra"
+)
+
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "info",
+		Short: "Retrieve information from the server",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, _ []string) error {
+			cli, err := common.NewClientWithDefaultContextAndAddress()
+			if err != nil {
+				return fmt.Errorf("failed to create client: %w", err)
+			}
+
+			res, err := cli.Client.GetInfo(cli.Ctx, &tetragon.GetInfoRequest{})
+			if err != nil {
+				return fmt.Errorf("failed to retrieve info: %w", err)
+			}
+
+			info := tetragoninfo.Decode(res)
+			b, err := json.Marshal(info)
+			if err != nil {
+				return fmt.Errorf("failed to marshal info: %w", err)
+			}
+			cmd.Println(string(b))
+			return nil
+		},
+	}
+	return cmd
+}
--- a/cmd/tetra/commands.go
+++ b/cmd/tetra/commands.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/cilium/tetragon/cmd/tetra/explain"
 	"github.com/cilium/tetragon/cmd/tetra/getevents"
+	"github.com/cilium/tetragon/cmd/tetra/info"
 	"github.com/cilium/tetragon/cmd/tetra/rthooks"
 	"github.com/cilium/tetragon/cmd/tetra/sensors"
 	"github.com/cilium/tetragon/cmd/tetra/stacktracetree"
@@ -25,6 +26,7 @@ func addBaseCommands(rootCmd *cobra.Command) {
 	rootCmd.AddCommand(status.New())
 	rootCmd.AddCommand(rthooks.New())
 	rootCmd.AddCommand(explain.New())
+	rootCmd.AddCommand(info.New())
 
 	// bugtool technically builds on darwin and windows but makes no sense since
 	// it's supposed to be run on the machine running Tetragon, using


pkg/tetragoninfo/info.go
```diff
@@ -0,0 +1,136 @@
...
+		Build: version.BuildInfo{
+			GoVersion: res.GetBuild().GoVersion,
+			Commit:    res.GetBuild().Commit,
+			Time:      res.GetBuild().Time,
+			Modified:  res.GetBuild().Modified,
+		},
+	}
+}
+
+// decodeConf decodes the confiugration part of info
```
mtardy: nit: small typo on config

cmd/tetra/info/info.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/logger/logfields"
+	"github.com/cilium/tetragon/pkg/tetragoninfo"
+
+	"github.com/spf13/cobra"
+)
+
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "info",
```
mtardy, line 19: is this command going to be relevant for most users? I would say mostly yes so keeping it as a top level command is fine, otherwise we could put it under `debug`.

cmd/tetra/info/info.go
```diff
@@ -0,0 +1,48 @@
...
+					info := tetragoninfo.Decode(res)
+					b, err := json.Marshal(info)
+					if err != nil {
+						logger.GetLogger().Error("failed to generate json", logfields.Error, err)
+						return
+					}
+					cmd.Println(string(b))
+				},
+				func(err error) {
+					logger.GetLogger().Error("error retrieving server info", logfields.Error, err)
```
mtardy: why not just return the error here? otherwise your `tetra info` will return `0` as exit.

pkg/bpf/detect_linux.go
```diff
@@ -614,10 +615,47 @@ func LogFeatures() string {
...
+	case 1:
+		return probeStr(&FeatureProbes[0])
+	}
+
+	var bld strings.Builder
+	bld.WriteString(probeStr(&FeatureProbes[0]))
+	for _, p := range FeatureProbes[1:] {
+		bld.WriteString(", " + probeStr(&p))
+	}
+	return bld.String()
```
mtardy, line 635:
```suggestion
	if len(FeatureProbes) == 0 {
		return ""
	}

	var probes []string
	for i := range FeatureProbes {
		probes = append(probes, probeStr(&FeatureProbes[i])) // can simplify if you don't use pointers
	}

	return strings.Join(probes, ", ")
```

pkg/bpf/detect_linux.go
```diff
@@ -614,10 +615,47 @@ func LogFeatures() string {
...
-	return fmt.Sprintf("override_return: %t, buildid: %t, kprobe_multi: %t, uprobe_multi: %t, fmodret: %t, fmodret_syscall: %t, signal: %t, large: %t, link_pin: %t, lsm: %t, missed_stats_kprobe_multi: %t, missed_stats_kprobe: %t, batch_update: %t, uprobe_refctroff: %t, audit_loginuid: %t, probe_write_user: %t, uprobe_regs_change: %t",
-		HasOverrideHelper(), HasBuildId(), HasKprobeMulti(), HasUprobeMulti(),
-		HasModifyReturn(), HasModifyReturnSyscall(), HasSignalHelper(), HasProgramLargeSize(),
-		HasLinkPin(), HasLSMPrograms(), HasMissedStatsKprobeMulti(), HasMissedStatsPerfEvent(),
-		HasBatchAPI(), HasUprobeRefCtrOffset(), HasAuditLoginuid(), HasProbeWriteUserHelper(),
-		HasUprobeRegsChange())
+
+	probeStr := func(p *FeatureProbe) string {
+		return p.Name + ": " + strconv.FormatBool(p.Fn())
+	}
```
mtardy, line 621: You could also define the stringer for `FeatureProbe` directly

mtardy: also why take a pointer, will the function once thing will not work with a copy? I imagine that's why you need this altogether, because stringer with side effect wouldn't work hehe

mtardy (commented):
Thanks overall looks good, a few comments

pkg/bpf/detect_linux.go
kkourt: The whole point of this was to avoid using `strings.Join()` :)

pkg/bpf/detect_linux.go
kkourt: > You could also define the stringer for FeatureProbe directly

Why? 

This is not a canonical representation of the object -- I just use it to create the log message.

> also why take a pointer, will the function once thing will not work with a copy? I imagine that's why you need this altogether, because stringer with side effect wouldn't work hehe

Why do a copy?

mtardy: > Why do a copy?

if you don't modify your input, why not?

pkg/bpf/detect_linux.go
mtardy: why are you avoiding `strings.Join()`?

kkourt: Because the version above is more efficient. In strings.Join you first need to build all the individual strings and then join them. In the above version you just append what you want in the final string.

pkg/bpf/detect_linux.go
kkourt: To avoid the (potential) copy. Passing a pointer is more efficient than passing a structure if you need to copy it.

pkg/bpf/detect_linux.go
mtardy: So for fun I benchmarked this to see the reality, I wrote it super fast, maybe I'm missing something in the benchmark.

```go
package main

import "strings"

func korni(slice []string) string {
	switch len(slice) {
	case 0:
		return ""
	case 1:
		return slice[0]
	}

	var bld strings.Builder
	bld.WriteString(slice[0])
	for _, p := range slice[1:] {
		bld.WriteString(", " + p)
	}
	return bld.String()
}

func mahe(slice []string) string {
	if len(slice) == 0 {
		return ""
	}
	var probes []string
	for _, i := range slice {
		probes = append(probes, i)
	}
	return strings.Join(probes, ", ")
}

func maheOpt(slice []string) string {
	if len(slice) == 0 {
		return ""
	}
	probes := make([]string, 0, len(slice))
	for _, i := range slice {
		probes = append(probes, i)
	}
	return strings.Join(probes, ", ")
}
```

with 

```go
func generateStringSlice(n int, strLen int) []string {
	const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
	res := make([]string, n)
	seed := rand.New(rand.NewSource(time.Now().UnixNano()))

	for i := 0; i < n; i++ {
		b := make([]byte, strLen)
		for j := range b {
			b[j] = charset[seed.Intn(len(charset))]
		}
		res[i] = string(b)
	}
	return res
}

func Benchmark(b *testing.B) {
	// Define the arbitrary sizes you want to test
	sizes := []int{10, 100, 1000, 10000}

	for _, size := range sizes {
		data := generateStringSlice(size, 10)

		b.Run(fmt.Sprintf("kornilios size%d", size), func(b *testing.B) {
			b.ReportAllocs()
			for i := 0; i < b.N; i++ {
				korni(data)
			}
		})

		b.Run(fmt.Sprintf("mahe size%d", size), func(b *testing.B) {
			b.ReportAllocs()
			for i := 0; i < b.N; i++ {
				mahe(data)
			}
		})

		b.Run(fmt.Sprintf("maheOpt size%d", size), func(b *testing.B) {
			b.ReportAllocs()
			for i := 0; i < b.N; i++ {
				maheOpt(data)
			}
		})
	}
}
```

And the result is that it's not more efficient in performance or allocation most of the time (only time is with 10000 len strings in comparison to naive implementation that doesn't pre-allocate in my case):

```
goos: darwin
goarch: arm64
pkg: benchmark-deleteme
cpu: Apple M3 Max
Benchmark/kornilios_size10-14            6407629               163.9 ns/op           240 B/op          4 allocs/op
Benchmark/mahe_size10-14                 7358383               163.4 ns/op           576 B/op          4 allocs/op
Benchmark/maheOpt_size10-14             11312643               106.7 ns/op           288 B/op          2 allocs/op
Benchmark/kornilios_size100-14            804900              1406 ns/op            3312 B/op          8 allocs/op
Benchmark/mahe_size100-14                 887532              1303 ns/op            5696 B/op          7 allocs/op
Benchmark/maheOpt_size100-14             1255794               953.1 ns/op          3072 B/op          2 allocs/op
Benchmark/kornilios_size1000-14            80457             14419 ns/op           46576 B/op         15 allocs/op
Benchmark/mahe_size1000-14                105189             11302 ns/op           47424 B/op         10 allocs/op
Benchmark/maheOpt_size1000-14             127783              9189 ns/op           28672 B/op          2 allocs/op
Benchmark/kornilios_size10000-14            7392            148060 ns/op          514800 B/op         23 allocs/op
Benchmark/mahe_size10000-14                 6666            166958 ns/op          788801 B/op         17 allocs/op
Benchmark/maheOpt_size10000-14             10000            100162 ns/op          286720 B/op          2 allocs/op
PASS
ok      benchmark-deleteme      15.753s
```

pkg/bpf/detect_linux.go
mtardy: > Passing a pointer is more efficient than passing a structure if you need to copy it

Fyi: I've been reading many times the contrary in Go that when dealing with small structs, copying and passing by value is actually faster when trying in benchmarks:
- https://blog.boot.dev/golang/pointers-faster-than-values/
- https://medium.com/@harshithgowdakt/go-pointers-vs-values-the-complete-developers-guide-be71ceb11a4d
- https://preslav.me/2026/01/08/golang-structs-vs-pointers-pointer-first/

But anyway, I don't really mind it's just that quite often the conclusion is "don't try to optimize because often the actual perf is counter intuitive", so I'll just write what's simplest and convey intent, explaining why I wondered the pointer here.

pkg/bpf/detect_linux.go
mtardy: re-reading this my message looks a bit pedantic but I really was curious because I have been told also that using the string builder is more efficient.

kkourt: From what I can tell, your benchmark does not include the intermediate string generation, that is `probeStr(&FeatureProbes[i])`. Am I missing something?

kkourt: Here's my attempt:

<details>

```
package main

import (
	"strconv"
	"strings"
)

type t struct {
	b bool
	s string
}

func korni(t []t) string {
	switch len(t) {
	case 0:
		return ""
	case 1:
		return t[0].s + ": " + strconv.FormatBool(t[0].b)
	}

	var bld strings.Builder
	bld.WriteString(t[0].s + ": " + strconv.FormatBool(t[0].b))

	for _, ti := range t[1:] {
		bld.WriteString(ti.s + ": " + strconv.FormatBool(ti.b))
	}
	return bld.String()
}

func mahe(t []t) string {
	if len(t) == 0 {
		return ""
	}
	var probes []string
	for _, ti := range t {
		p := ti.s + ": " + strconv.FormatBool(ti.b)
		probes = append(probes, p)
	}
	return strings.Join(probes, ", ")
}
```

```
package main

import (
	"fmt"
	"math/rand"
	"testing"
	"time"
)

func generateTs(n int, strLen int) []t {
	const charset = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"
	res := make([]t, n)
	seed := rand.New(rand.NewSource(time.Now().UnixNano()))
	for i := range n {
		str := make([]byte, strLen)
		for j := range strLen {
			str[j] = charset[seed.Intn(len(charset))]
		}
		var b bool
		if seed.Intn(2) == 1 {
			b = true
		}
		res[i] = t{b: b, s: string(str)}
	}
	return res
}

func Benchmark(b *testing.B) {
	// Define the arbitrary sizes you want to test
	sizes := []int{10, 100, 1000, 10000}

	for _, size := range sizes {
		data := generateTs(size, 10)

		b.Run(fmt.Sprintf("kornilios size%d", size), func(b *testing.B) {
			b.ReportAllocs()
			for i := 0; i < b.N; i++ {
				korni(data)
			}
		})

		b.Run(fmt.Sprintf("mahe size%d", size), func(b *testing.B) {
			b.ReportAllocs()
			for i := 0; i < b.N; i++ {
				mahe(data)
			}
		})

	}
}
```

</details>

Results:
```
goos: linux
goarch: amd64
pkg: splitmark
cpu: AMD Ryzen 7 PRO 8840HS w/ Radeon 780M Graphics
Benchmark/kornilios_size10-16         	 4527552	       261.1 ns/op	     360 B/op	       4 allocs/op
Benchmark/mahe_size10-16              	 2386431	       500.5 ns/op	     840 B/op	      14 allocs/op
Benchmark/kornilios_size100-16        	  433150	      2491 ns/op	    4392 B/op	       8 allocs/op
Benchmark/mahe_size100-16             	  278076	      4544 ns/op	    8440 B/op	     107 allocs/op
Benchmark/kornilios_size1000-16       	   42097	     28060 ns/op	   83368 B/op	      16 allocs/op
Benchmark/mahe_size1000-16            	   27807	     44064 ns/op	   74408 B/op	    1010 allocs/op
Benchmark/kornilios_size10000-16      	    3774	    293967 ns/op	  685480 B/op	      23 allocs/op
Benchmark/mahe_size10000-16           	    2097	    539098 ns/op	 1054178 B/op	   10017 allocs/op
```

I wrote this  hastily, so it might be the case that I'm missing something.

kkourt: Also did a quick test to compare pass-by-value and pass-by-pointer: https://gist.github.com/kkourt/59e26b9006208a00f16c21118c0f55b0.

The time is slightly better when passing by value, but it results in more instructions. Overall, I don't think it makes   a lot of difference.

mtardy: Ah you are completely right, as soon as you create those new strings (what we actually do in the code) my implementation is way less performant, the allocation difference is insane!

pkg/tetragoninfo/info.go
kkourt: Thanks!

PS: If I ever do a PR without a typo, it probably means that my account was hacked.

cmd/tetra/info/info.go
kkourt: It also seems better at the top level to me as well.

pkg/bpf/detect_linux.go
kkourt: OK, i don't think it makes a big difference (see my other comment), but if you don't really mind I will leave it as is to expedite things.

cmd/tetra/info/info.go
kkourt: The problem is that the wrapper we are using:

> func CliRunErr(fn func(ctx context.Context, cli tetragon.FineGuidanceSensorsClient), fnErr func(err error)) {

Does not have the ability to return an error so that we can use `RunE` instead of `Run`. We should probably change that, but I would leave it as a follow up.

kkourt:
Thanks @mtardy! Pushed a new version.

cmd/tetra/info/info.go
mtardy: Yeah I see, I'm fine with it but just know this is the legacy way of writing gRPC style command. Now you can just use `common.NewClientWithDefaultContextAndAddress` and get your client without the closure style thing. Grep for example, it's actually more simple and then you can do whatever you like.

pkg/tetragoninfo/info.go
mtardy: ahah noted, I'll be watching now

cmd/tetra/info/info.go
kkourt: fair enough. Fixed!

PR #4593, opened 2026-01-29
fix(grpc/exec): fix RefDec in exit's eventcache path 

dwindsor:
https://github.com/isovalent/hubble-fgs/issues/7063

### Description

Make the exit eventcache path mirror the exec path with respect to refcount ops.

When events get kicked to the eventcache, they will be retried a certain number of times to see if pod/parent info has become available. There are 2 methods for retrying: `RetryInternal()` and `Retry()`, which are interfaces. `RetryInternal()` is sometimes [called before](https://github.com/cilium/tetragon/blob/eb584c98e16894ecfb2b9ae79012241d00d1ef5b/pkg/eventcache/eventcache.go#L154) `Retry()`, depending on the circumstances.

For exec, we have the `parent.RefInc("parent")` being [called in Retry()](https://github.com/cilium/tetragon/blob/eb584c98e16894ecfb2b9ae79012241d00d1ef5b/pkg/grpc/exec/exec.go#L229), after the [pod info check](https://github.com/cilium/tetragon/blob/eb584c98e16894ecfb2b9ae79012241d00d1ef5b/pkg/grpc/exec/exec.go#L208C5-L208C6). In exit, we have `parent.RefDec("parent")` [called in RetryInternal()](https://github.com/cilium/tetragon/blob/eb584c98e16894ecfb2b9ae79012241d00d1ef5b/pkg/grpc/exec/exec.go#L541) before the pod info check that happens in `Retry()`, which in the exit event's case just happens to be a [call](https://github.com/cilium/tetragon/blob/eb584c98e16894ecfb2b9ae79012241d00d1ef5b/pkg/grpc/exec/exec.go#L567) to `eventcache.HandleGenericEvent`.

Make the exit path balance with the exec path in that it should only call `parent.RefDec("parent")` after it's confirmed that pod info is present.

### Results

_Before Fix_
```
tetragon_event_cache_inserts_total 161844
tetragon_event_cache_fetch_retries_total{entry_type="ancestors_info"} 0
tetragon_event_cache_fetch_retries_total{entry_type="parent_info"} 1020
tetragon_event_cache_fetch_retries_total{entry_type="pod_info"} 55169
tetragon_event_cache_fetch_retries_total{entry_type="process_info"} 0
tetragon_event_cache_fetch_failures_total{entry_type="pod_info",event_type="PROCESS_EXEC"} 509
tetragon_event_cache_fetch_failures_total{entry_type="pod_info",event_type="PROCESS_EXIT"} 458

# how many refcnt underflows were detected?
$ tetra debug dump processcache 2>/dev/null | jq -s '[.[] | select(.refcnt > 2147483647)] | length'
565

  {
    "process": {
      "exec_id": "Y3ZlLXRlc3QtY29udHJvbC1wbGFuZTo2MjA3MzQ5NDMxNTc0OTQ6MzY1ODY2MQ==",
      "pid": 3658661,
      "uid": 0,
      "cwd": "/",
      "binary": "/bin/sh",
      "arguments": "/scripts/churn.sh",
      "flags": "execve rootcwd clone inInitTree",
      "start_time": "2026-01-29T19:54:58.884463084Z",
      "auid": 4294967295,
      "pod": {
  ...
     }
    },
    "color": "deletePending",
    "refcnt": 4294967295,
    "refcnt_ops": {
      "parent++": 202,
      "parent--": 203,
      "process++": 1,
      "process--": 1
    }
  },
```

_After Fix_
```
tetragon_event_cache_inserts_total 161948
tetragon_event_cache_fetch_retries_total{entry_type="ancestors_info"} 0
tetragon_event_cache_fetch_retries_total{entry_type="parent_info"} 4556
tetragon_event_cache_fetch_retries_total{entry_type="pod_info"} 53862
tetragon_event_cache_fetch_retries_total{entry_type="process_info"} 0
tetragon_event_cache_fetch_failures_total{entry_type="pod_info",event_type="PROCESS_EXEC"} 508
tetragon_event_cache_fetch_failures_total{entry_type="pod_info",event_type="PROCESS_EXIT"} 450

# how many refcnt underflows were detected? 
$ tetra debug dump processcache 2>/dev/null | jq -s '[.[] | select(.refcnt > 2147483647)] | length'
0
```

This is the workload that was used to generate these metrics:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: churn-scripts
  namespace: default
data:
  churn.sh: |
    #!/bin/sh
    for i in $(seq 1 100); do
      /bin/true
    done
---
apiVersion: batch/v1
kind: Job
metadata:
  name: churn-job
  namespace: default
spec:
  parallelism: 50
  completions: 500
  backoffLimit: 500
  template:
    spec:
      restartPolicy: Never
      volumes:
        - name: scripts
          configMap:
            name: churn-scripts
            defaultMode: 0755
      containers:
        - name: churn
          image: alpine:latest
          command: ["/bin/sh", "/scripts/churn.sh"]
          volumeMounts:
            - name: scripts
              mountPath: /scripts
```

### Changelog

```release-note
```

### DIFF
--- a/pkg/grpc/exec/exec.go
+++ b/pkg/grpc/exec/exec.go
@@ -520,15 +520,13 @@ func (msg *MsgExitEventUnix) RetryInternal(ev notify.Event, timestamp uint64) (*
 	proc, parent := process.GetParentProcessInternal(msg.ProcessKey.Pid, timestamp)
 	var err error
 
-	if option.Config.EnableProcessAncestors && proc.NeededAncestors() && !msg.RefCntDone[AncestorsRefCnt] {
+	if option.Config.EnableProcessAncestors && proc.NeededAncestors() {
 		if ancestors, perr := process.GetAncestorProcessesInternal(proc.UnsafeGetProcess().ParentExecId); perr == nil {
 			var tetragonAncestors []*tetragon.Process
 			for _, ancestor := range ancestors {
 				tetragonAncestors = append(tetragonAncestors, ancestor.UnsafeGetProcess())
-				ancestor.RefDec("ancestor")
 			}
 			ev.SetAncestors(tetragonAncestors)
-			msg.RefCntDone[AncestorsRefCnt] = true
 		} else {
 			eventcache.CacheRetries(eventcache.AncestorsInfo).Inc()
 			err = eventcache.ErrFailedToGetAncestorsInfo
@@ -537,10 +535,6 @@ func (msg *MsgExitEventUnix) RetryInternal(ev notify.Event, timestamp uint64) (*
 
 	if parent != nil {
 		ev.SetParent(parent.UnsafeGetProcess())
-		if !msg.RefCntDone[ParentRefCnt] {
-			parent.RefDec("parent")
-			msg.RefCntDone[ParentRefCnt] = true
-		}
 	} else {
 		eventcache.CacheRetries(eventcache.ParentInfo).Inc()
 		err = eventcache.ErrFailedToGetParentInfo
@@ -549,10 +543,6 @@ func (msg *MsgExitEventUnix) RetryInternal(ev notify.Event, timestamp uint64) (*
 	if proc != nil {
 		// Use cached version of the process
 		ev.SetProcess(proc.UnsafeGetProcess())
-		if !msg.RefCntDone[ProcessRefCnt] {
-			proc.RefDec("process")
-			msg.RefCntDone[ProcessRefCnt] = true
-		}
 	} else {
 		eventcache.CacheRetries(eventcache.ProcessInfo).Inc()
 		err = eventcache.ErrFailedToGetProcessInfo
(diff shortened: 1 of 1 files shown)


pkg/grpc/exec/exec.go
```diff
@@ -565,7 +555,34 @@ func (msg *MsgExitEventUnix) RetryInternal(ev notify.Event, timestamp uint64) (*
 }
 
 func (msg *MsgExitEventUnix) Retry(internal *process.ProcessInternal, ev notify.Event) error {
-	return eventcache.HandleGenericEvent(internal, ev, nil)
+	if err := eventcache.HandleGenericEvent(internal, ev, nil); err != nil {
```
dwindsor, line 558: pod info check happens in `HandleGenericEvent`

pkg/grpc/exec/exec.go
```diff
@@ -520,15 +520,13 @@ func (msg *MsgExitEventUnix) RetryInternal(ev notify.Event, timestamp uint64) (*
 	proc, parent := process.GetParentProcessInternal(msg.ProcessKey.Pid, timestamp)
 	var err error
 
-	if option.Config.EnableProcessAncestors && proc.NeededAncestors() && !msg.RefCntDone[AncestorsRefCnt] {
```
dwindsor, line 523: Move ancestor accounting to `Retry()` to mirror exec path

tpapagian (approved):
Thanks, this sounds reasonable. 

We have some tests in https://github.com/cilium/tetragon/blob/main/pkg/grpc/exec/exec_test.go and https://github.com/cilium/tetragon/blob/main/pkg/grpc/exec/exec_test_helper.go that checks reference counting. It would be great to have a new test (or modify an existing one if this makes more sense) that shows the issue (i.e. fails without your fix and pass with your fix) in order to avoid breaking it again in the future.

dwindsor:
> It would be great to have a new test (or modify an existing one if this makes more sense) that shows the issue (i.e. fails without your fix and pass with your fix) in order to avoid breaking it again in the future.

Agreed, looking into this, it requires us to sleep for a while, which we likely shouldn't do in unit tests.

The closest test function we have is `GrpcExecPodOutOfOrder` but we need something like `GrpcExecPodInfoNeverArrives`, which would require us to sleep until the max retries in the eventcache was reached. Tried reworking it using testing/synctest but there were some non-trivial interactions w/ the eventcache and processcache, causing deadlocks.

tpapagian:
> > It would be great to have a new test (or modify an existing one if this makes more sense) that shows the issue (i.e. fails without your fix and pass with your fix) in order to avoid breaking it again in the future.
> 
> Agreed, looking into this, it requires us to sleep for a while, which we likely shouldn't do in unit tests.
> 
> The closest test function we have is `GrpcExecPodOutOfOrder` but we need something like `GrpcExecPodInfoNeverArrives`, which would require us to sleep until the max retries in the eventcache was reached. Tried reworking it using testing/synctest but there were some non-trivial interactions w/ the eventcache and processcache, causing deadlocks.

Let's don't block on that. We can re-consider that in a future PR.

PR #4597, opened 2026-01-31
synthetic: add event recording and replay support

### LINKED ISSUE #4497: Add synthetic event injection API
The `ProcessCache` is implemented as a [color-based](https://github.com/cilium/tetragon/blob/2c3ccdc12061890cc69d4eef0be768505a58d33c/pkg/process/cache.go#L31) (`inUse`, `deletePending`, `deleteReady`, `deleted`) LRU cache whose purpose is to track process state over time. It has a garbage collector ([`CacheGarbageCollector`](https://github.com/cilium/tetragon/blob/2c3ccdc12061890cc69d4eef0be768505a58d33c/pkg/process/cache.go#L46)) that uses a [`time.Ticker`](https://github.com/cilium/tetragon/blob/2c3ccdc12061890cc69d4eef0be768505a58d33c/pkg/process/cache.go#L47) to move each cache entry through its lifecycle (`inUse` -> `deletePending` -> ... -> `deleted`); only those entries with color `deleteReady` are actually removed.

In order for an entry to be tagged with `deleteReady`, its refcount must be 0. Refcounts can become unbalanced due to either missing/dropped bpf events or [out-of-order](https://github.com/cilium/tetragon/blob/2c3ccdc12061890cc69d4eef0be768505a58d33c/pkg/process/cache.go#L62) bpf events.

In order to reason more about cache behavior for OOO events in particular, it would be useful to give Tetragon the ability to receive its events from a text file containing a synthetic event steam, rather than from the bpf subsystem itself. 

This would allow us to test hypotheses about abnormal event ordering and their effects on cache state without having to actually reproduce the conditions that lead to the OOO events, which can be very difficult sometimes.

cc @kevsecurity @kkourt @dangome3 @FedeDP @ariosmon

sglushko:
Fixes: cilium#4497

### Description
Added a full synthetic events record/replay pipeline for deterministic debugging and testing.  
This includes a JSONL recorder (WritingListener), a replay observer (ReadingObserver), type‑preserving JSON serialization for interface fields, and factory helpers wired to CLI flags.

**How to run (record):**
```bash
sudo ./tetragon --synthetic-events-log /tmp/synthetic-events.jsonl --synthetic-events-verify-roundtrip
```
- If the log file already exists, it is renamed with a timestamp prefix (e.g. `2026-02-02_11-05-30_synthetic-events.jsonl`) and a fresh file is created.
- The `--synthetic-events-verify-roundtrip` flag enables a check to ensure each recorded event is marshaled/unmarshaled and verified to be identical to the original event. In case of a mismatch, the following warnings will be logged: `Roundtrip verification: unmarshal failed` or `Roundtrip verification: mismatch detected`.
- If a recorded event contains an unregistered type, round‑trip verification logs `Roundtrip verification: unmarshal failed` (error `unknown type: <Type>`). Register new event types in `pkg/synthetic/types_linux.go` via `RegisterType((*NewType)(nil))`.

**How to run (replay):**
```bash
sudo ./tetragon --synthetic-events-source /tmp/synthetic-events.jsonl
```

**Expected logs:**
- `Starting synthetic event replay`
- `Finished synthetic event replay` (with event count)
- `Failed to unmarshal event` `unknown type:` (needs to register new event types in `pkg/synthetic/types_linux.go`) 

**Event Log format:**
JSONL, with type‑preserving wrappers:
```json
{"synthetic_type":"<type>","synthetic_value":{...}}
```

**Unit tests (example):**
```bash
go test -coverprofile=coverage.out ./pkg/synthetic/... && go tool cover -func=coverage.out
```
Example output:
```
ok      github.com/cilium/tetragon/pkg/synthetic        0.053s  coverage: 91.0% of statements
github.com/cilium/tetragon/pkg/synthetic/factory_linux.go:18:           NewReadingObserverFromFile      0.0%
github.com/cilium/tetragon/pkg/synthetic/factory_linux.go:32:           NewWritingListenerToFile        0.0%
github.com/cilium/tetragon/pkg/synthetic/factory_linux.go:46:           openSourceFile                  0.0%
github.com/cilium/tetragon/pkg/synthetic/factory_linux.go:56:           openLogFile                     0.0%
github.com/cilium/tetragon/pkg/synthetic/reading_observer_linux.go:35:  NewReadingObserver              100.0%
github.com/cilium/tetragon/pkg/synthetic/reading_observer_linux.go:45:  StartReady                      100.0%
github.com/cilium/tetragon/pkg/synthetic/reading_observer_linux.go:88:  observerListeners               100.0%
github.com/cilium/tetragon/pkg/synthetic/reading_observer_linux.go:98:  AddListener                     100.0%
github.com/cilium/tetragon/pkg/synthetic/reading_observer_linux.go:104: RemoveListener                  100.0%
github.com/cilium/tetragon/pkg/synthetic/serializer.go:24:              Marshal                         100.0%
github.com/cilium/tetragon/pkg/synthetic/serializer.go:34:              Unmarshal                       100.0%
github.com/cilium/tetragon/pkg/synthetic/serializer.go:57:              unmarshalWithTypes              100.0%
github.com/cilium/tetragon/pkg/synthetic/serializer.go:76:              getTypeName                     100.0%
github.com/cilium/tetragon/pkg/synthetic/serializer.go:81:              wrapValue                       100.0%
github.com/cilium/tetragon/pkg/synthetic/serializer.go:174:             unwrapInto                      100.0%
github.com/cilium/tetragon/pkg/synthetic/serializer.go:280:             unwrapTypedValue                100.0%
github.com/cilium/tetragon/pkg/synthetic/serializer.go:304:             getJSONFieldName                100.0%
github.com/cilium/tetragon/pkg/synthetic/serializer.go:317:             setBasicValue                   100.0%
github.com/cilium/tetragon/pkg/synthetic/types.go:34:                   RegisterType                    100.0%
github.com/cilium/tetragon/pkg/synthetic/types_linux.go:15:             init                            100.0%
github.com/cilium/tetragon/pkg/synthetic/writing_listener_linux.go:33:  NewWritingListener              100.0%
github.com/cilium/tetragon/pkg/synthetic/writing_listener_linux.go:49:  Notify                          100.0%
github.com/cilium/tetragon/pkg/synthetic/writing_listener_linux.go:75:  Close                           100.0%
github.com/cilium/tetragon/pkg/synthetic/writing_listener_linux.go:83:  WithVerifyRoundtrip             100.0%
github.com/cilium/tetragon/pkg/synthetic/writing_listener_linux.go:90:  WithDelegate                    100.0%
github.com/cilium/tetragon/pkg/synthetic/writing_listener_linux.go:97:  verifyRoundtripEquality         100.0%
total:                                                                  (statements)                    91.0%
```

Also here is the post describing the Usage & Testing Guide. https://github.com/cilium/tetragon/issues/4497#issuecomment-3836611060

**release-note**
```release-note
Added synthetic event recording/replay via --synthetic-events-log and --synthetic-events-source, with optional --synthetic-events-verify-roundtrip validation.
```

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -366,7 +366,12 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 	}
 
 	// Get observer from configFile
-	obs := observer.NewObserver()
+	var obs observer.EventObserver
+	obs = observer.NewObserver()
+	obs, err = setupObserver(ctx, obs, log)
+	if err != nil {
+		return err
+	}
 	defer func() {
 		obs.PrintStats()
 	}()
@@ -509,7 +514,9 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 	}
 
 	log.Info("Exporter configuration", "enabled", option.Config.ExportFilename != "", "fileName", option.Config.ExportFilename)
-	obs.AddListener(pm)
+	if err := setupListener(ctx, obs, pm, log); err != nil {
+		return err
+	}
 	saveInitInfo()
 
 	// Initialize a k8s watcher used to manage policies. This should happen
@@ -632,7 +639,7 @@ func addTracingPolicy(ctx context.Context, file string) error {
 // Periodically log current status every 24 hours. For lost or error
 // events we ratelimit statistics to 1 message per every 1hour and
 // only if they increase, to inform users that events are being lost.
-func logStatus(ctx context.Context, obs *observer.Observer) {
+func logStatus(ctx context.Context, obs observer.EventObserver) {
 	prevLost := uint64(0)
 	prevErrors := uint64(0)
 	lostTicker := time.NewTicker(1 * time.Hour)
@@ -808,7 +815,7 @@ func startGopsServer() error {
 
 func execute() error {
 	rootCmd := &cobra.Command{
-		Use:   "tetragon",
+		Use:   binaryName(),
 		Short: "Tetragon - eBPF-based Security Observability and Runtime Enforcement",
 		Run: func(cmd *cobra.Command, _ []string) {
 			if viper.GetBool(option.KeyGenerateDocs) {
@@ -820,6 +827,9 @@ func execute() error {
 
 			if err := option.ReadAndSetFlags(); err != nil {
 				logger.Fatal(log, "Failed to parse command line flags", logfields.Error, err)
+			}
+			if err := readExtraFlags(); err != nil {
+				logger.Fatal(log, "Failed to parse extra command line flags", logfields.Error, err)
 			}
 			// Override perf ring buffer choice if only the perf ring is available.
 			// NB: can't do this in option.ReadAndSetFlags() as it causes an import cycle.
@@ -844,6 +854,7 @@ func execute() error {
 
 	flags := rootCmd.PersistentFlags()
 	option.AddFlags(flags)
+	addExtraFlags(flags)
 	viper.BindPFlags(flags)
 	return rootCmd.Execute()
 }
--- a/pkg/synthetic/serializer.go
+++ b/pkg/synthetic/serializer.go
@@ -1,0 +1,361 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+// serializer.go provides JSON serialization and deserialization with type preservation
+// for interface fields. This enables roundtrip of events through JSON while maintaining
+// concrete type information.
+
+package synthetic
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+)
+
+// Serializer implements Codec with type-preserving serialization.
+type Serializer struct{}
+
+// Marshal serializes any value with type wrapper at top level and for interface fields.
+func (Serializer) Marshal(v any) ([]byte, error) {
+	ifaceVal := reflect.ValueOf(&v).Elem()
+	wrapped, err := wrapValue(ifaceVal)
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(wrapped)
+}
+
+// Unmarshal deserializes JSON with type wrappers.
+func (Serializer) Unmarshal(data []byte) (any, error) {
+	var wrapper TypedValue
+	if err := json.Unmarshal(data, &wrapper); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal event wrapper: %w", err)
+	}
+
+	if wrapper.Type == "" {
+		return nil, errors.New("missing type in event wrapper")
+	}
+
+	factory, ok := InterfaceRegistry[wrapper.Type]
+	if !ok {
+		return nil, fmt.Errorf("unknown type: %s", wrapper.Type)
+	}
+
+	event := factory()
+	if err := unmarshalWithTypes(wrapper.Value, event); err != nil {
+		return nil, err
+	}
+	return event, nil
+}
+
+// unmarshalWithTypes deserializes JSON with type wrappers back to the target.
+func unmarshalWithTypes(data []byte, v any) error {
+	rv := reflect.ValueOf(v)
+	if rv.Kind() != reflect.Ptr || rv.IsNil() {
+		return errors.New("target must be non-nil pointer")
+	}
+
+	// Use Decoder with UseNumber to preserve precision for large integers (uint64)
+	dec := json.NewDecoder(bytes.NewReader(data))
+	dec.UseNumber()
+
+	var raw any
+	if err := dec.Decode(&raw); err != nil {
+		return err
+	}
+
+	return unwrapInto(raw, rv.Elem())
+}
+
+// getTypeName returns the type name for a value (e.g. "*tracing.MsgGenericKprobeUnix").
+func getTypeName(v any) string {
+	return reflect.TypeOf(v).String()
+}
+
+// wrapValue recursively wraps interface values with type info.
+func wrapValue(v reflect.Value) (any, error) {
+	switch v.Kind() {
+	case reflect.Interface:
+		if v.IsNil() {
+			return nil, nil
+		}
+		elem := v.Elem()
+		// Wrap interface value with type info
+		wrapped, err := wrapValue(elem)
+		if err != nil {
+			return nil, err
+		}
+		data, err := json.Marshal(wrapped)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal interface value: %w", err)
+		}
+		return TypedValue{
+			Type:  getTypeName(elem.Interface()),
+			Value: data,
+		}, nil
+
+	case reflect.Ptr:
+		if v.IsNil() {
+			return nil, nil
+		}
+		// For pointers, wrap the element and return as-is (json.Marshal handles pointers)
+		return wrapValue(v.Elem())
+
+	case reflect.Struct:
+		result := make(map[string]any)
+		t := v.Type()
+		for i := range v.NumField() {
+			field := t.Field(i)
+			if !field.IsExported() {
+				continue
+			}
+			name := getJSONFieldName(field)
+			if name == "-" {
+				continue
+			}
+			wrapped, err := wrapValue(v.Field(i))
+			if err != nil {
+				return nil, fmt.Errorf("field %s: %w", name, err)
+			}
+			result[name] = wrapped
+		}
+		return result, nil
+
+	case reflect.Slice:
+		if v.IsNil() {
+			return nil, nil
+		}
+		result := make([]any, v.Len())
+		for i := range v.Len() {
+			wrapped, err := wrapValue(v.Index(i))
+			if err != nil {
+				return nil, fmt.Errorf("index %d: %w", i, err)
+			}
+			result[i] = wrapped
+		}
+		return result, nil
+
+	case reflect.Map:
+		if v.IsNil() {
+			return nil, nil
+		}
+		result := make(map[string]any)
+		for _, key := range v.MapKeys() {
+			wrapped, err := wrapValue(v.MapIndex(key))
+			if err != nil {
+				return nil, fmt.Errorf("map key %v: %w", key.Interface(), err)
+			}
+			result[fmt.Sprint(key.Interface())] = wrapped
+		}
+		return result, nil
+
+	case reflect.Array:
+		result := make([]any, v.Len())
+		for i := range v.Len() {
+			wrapped, err := wrapValue(v.Index(i))
+			if err != nil {
+				return nil, fmt.Errorf("index %d: %w", i, err)
+			}
+			result[i] = wrapped
+		}
+		return result, nil
+
+	default:
+		return v.Interface(), nil
+	}
+}
+
+// unwrapInto recursively unwraps typed values into the target.
+func unwrapInto(raw any, target reflect.Value) error {
+	if raw == nil {
+		return nil
+	}
+
+	// Check if this is a typed wrapper
+	if m, ok := raw.(map[string]any); ok {
+		if typeName, hasType := m["synthetic_type"]; hasType {
+			if valueRaw, hasValue := m["synthetic_value"]; hasValue {
+				typeNameStr, ok := typeName.(string)
+				if !ok {
+					return fmt.Errorf("synthetic_type must be string, got %T", typeName)
+				}
+				return unwrapTypedValue(typeNameStr, valueRaw, target)
+			}
+		}
+	}
+
+	switch target.Kind() {
+	case reflect.Interface:
+		// Should not happen if serialization is correct - all interfaces should have typed wrappers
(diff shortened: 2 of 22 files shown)


sglushko:
I don't have permission to add labels to this PR. Could a maintainer please add the appropriate `release-note/*` label (probably `release-note/minor` for this new feature)?

Thanks!

cmd/tetragon/main.go
```diff
@@ -366,7 +367,16 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
...
+	var obs observer.EventObserver
+	obs = observer.NewObserver()
+	if option.Config.SyntheticEventsSource != "" {
+		readingObs, err := synthetic.NewReadingObserverFromFile(ctx, option.Config.SyntheticEventsSource, log)
+		if err != nil {
+			return err
+		}
+		readingObs.Observer = obs.(*observer.Observer)
+		obs = readingObs
+	}
```
kkourt, line 374: I feel like this shouldn't be in the main tetragon binary.

Can we have this functionality as a go test instead of having it in the tetragon binary?
Alternatively, if that's not possible, I would suggest creating another binary specific to the test?

sglushko: Thanks for the feedback.

A couple of questions first:

Does this concern apply only to replay, or to recording as well?
Is there an established pattern for test utilities in Tetragon I should follow?

For accurate recording and replay, the binary needs to share as much code as possible with the real tetragon — ideally it should be tetragon itself with minimal changes. If a separate binary can achieve this, I think that's possible.

mtardy: I +1 this, you should be able to bundle this as part of go test, you can even add flags and whatever you like for your use case. You can use `go test` for whatever you like, for example here are how we run the BPF unit tests https://github.com/cilium/tetragon/blob/430e8c1c432b033cf83df5486fb25588a4824d1c/bpf/tests/prepend_name_test.go#L115-L325

mtardy (commented):
I agree with Kornilios, maybe try to consider another design where you can use go test to start your thing, it might be a bit tricky to understand which part of tetragon you need, but you can take inspiration on existing units tests that are testing the full "load BPF progs with such policy and check for such events" or on the main function. See https://github.com/cilium/tetragon/pull/4597#discussion_r2775664703.

sglushko:
@kkourt @mtardy 
Thanks, that makes sense. I'll rework the PR to use `go test` for both recording and replay. The core `pkg/synthetic/` package (serializer, WritingListener, ReadingObserver) stays the same — I'll just wire them through test helpers instead of CLI flags. I'll remove the changes to `cmd/tetragon/` and `pkg/option/`.
 
Just to clarify — did you have a chance to look at the current version? It already uses a separate binary with a `//go:build synthetic` tag ([mode_synthetic.go](cci:7://file:///Users/sglushko/.windsurf/worktrees/tetragon/tetragon-e8966f97/cmd/tetragon/mode_synthetic.go:0:0-0:0) / [mode_default.go](cci:7://file:///Users/sglushko/.windsurf/worktrees/tetragon/tetragon-e8966f97/cmd/tetragon/mode_default.go:0:0-0:0)), so the main tetragon binary is completely unaffected. But I understand the concern about keeping `cmd/tetragon/` clean, so I'm happy to move to the `go test` approach.

sglushko:
One point I’ve been considering is that go test might not be the best fit for recording. The intended use case is to run Tetragon in a cluster for an extended period (hours or days), recording events continuously while monitoring cache behavior. When an anomaly is detected (e.g., a refcount imbalance), the recorded event stream can be replayed locally to reproduce the issue. This is more of a long-running debugging workflow in a cluster environment than a standard test scenario.


Replay also benefits from running as a real Tetragon instance rather than within a go test harness. With the full gRPC server, metrics, and process cache accessible, you can inspect the cache state, counters, and event output exactly as you would during a live test run in the cluster. Inside a test harness, this level of observability is lost.


How would you suggest handling this? I am happy to discuss any options.

kkourt:
> Added a full synthetic events record/replay pipeline for deterministic debugging and testing.

...

> One point I’ve been considering is that go test might not be the best fit for recording. The intended use case is to run Tetragon in a cluster for an extended period (hours or days), recording events continuously while monitoring cache behavior.

The above statements seem contradictory to me. My interpretation of "synthetic events" is that they are artificial, while above you seem to be saying that you want to gather the events from a live cluster. Can you please clarify?

kkourt:
> Added a full synthetic events record/replay pipeline for deterministic debugging and testing.

...

>The intended use case is to run Tetragon in a cluster for an extended period (hours or days), recording events continuously while monitoring cache behavior.

The above statements seem contradictory to me. My interpretation of "synthetic events" is that they are artificial, while above you seem to be saying that you want to gather the events from a live cluster. Can you please clarify?

sglushko:
GitHub seems to be having some issues right now. Apologies if my messages end up being duplicated.

>The above statements seem contradictory to me. My interpretation of "synthetic events" is that they are artificial, while above you seem to be saying that you want to gather the events from a live cluster. Can you please clarify?

That’s a fair point, and I see where the confusion comes from.

By "synthetic events" I am referring to the fact that during replay, these events are injected into the pipeline from an external source (the JSONL file) instead of the BPF subsystem. From the perspective of the Tetragon observer, they are "synthetic" because they don't come from the kernel in real-time.

However, you are right that during recording, we are capturing real events from a live cluster. It is worth noting that the original task was focused only on injecting events from a file. During development, the idea to also record events emerged naturally, as it helped significantly with testing and debugging the implementation. This allowed us to capture real-world scenarios and then replay them. This doesn't contradict the original requirement to replay events; instead, it complements it by providing a reliable way to generate the necessary event streams.

kkourt:
> GitHub seems to be having some issues right now. Apologies if my messages end up being duplicated.
> 
> > The above statements seem contradictory to me. My interpretation of "synthetic events" is that they are artificial, while above you seem to be saying that you want to gather the events from a live cluster. Can you please clarify?
> 
> That’s a fair point, and I see where the confusion comes from.
> 
> By "synthetic events" I am referring to the fact that during replay, these events are injected into the pipeline from an external source (the JSONL file) instead of the BPF subsystem. From the perspective of the Tetragon observer, they are "synthetic" because they don't come from the kernel in real-time.
> 
> However, you are right that during recording, we are capturing real events from a live cluster. It is worth noting that the original task was focused only on injecting events from a file. During development, the idea to also record events emerged naturally, as it helped significantly with testing and debugging the implementation. This allowed us to capture real-world scenarios and then replay them. This doesn't contradict the original requirement to replay events; instead, it complements it by providing a reliable way to generate the necessary event streams.

What is the format of these events? Are these normal Tetragon events (e.g., PROCESS_EXEC)?

sglushko:
>What is the format of these events? Are these normal Tetragon events (e.g., PROCESS_EXEC)?

The events are captured here: https://github.com/cilium/tetragon/blob/main/pkg/observer/observer.go#L138. They implement the `notify.Message interface`, but since they can have various underlying types, I implemented a custom serializer and registered the supported types to ensure correct deserialization: https://github.com/cilium/tetragon/pull/4597/changes#diff-2734a53467c88b9adab56c2ee026ed2161cc9fe8ebcd309c345d816131d54ec5R17-R23


The serialized format is JSON, but a custom serializer is required because some struct fields use interface types, which the standard library cannot deserialize unambiguously. To handle this, I added a wrapper that includes type information:
```json 
{"synthetic_type":"<type>","synthetic_value":{...}}
```

<details>
<summary>Here is an example of a serialized PROCESS_EXEC event (`*exec.MsgExecveEventUnix`):</summary>

```json
{"synthetic_type":"*exec.MsgExecveEventUnix","synthetic_value":{"Unix":{"Kube":{"CgrpTrackerID":0,"Cgrpid":4429,"Docker":""},"Msg":{"CleanupProcess":{"Ktime":288251396523616,"Pad":0,"Pid":2332806},"Common":{"Flags":0,"Ktime":288251396707535,"Op":5,"PadV2":[0,0],"Size":874},"Creds":{"Cap":{"Effective":2199023255551,"Inheritable":0,"Permitted":2199023255551},"Egid":0,"Euid":0,"FSgid":0,"FSuid":0,"Gid":0,"Pad":0,"SecureBits":0,"Sgid":0,"Suid":0,"Uid":0,"UserNs":{"Gid":0,"Level":0,"NsInum":0,"Uid":0}},"Kube":{"CgrpTrackerID":0,"Cgrpid":4429,"Docker":[99,114,105,111,46,115,101,114,118,105,99,101,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]},"Namespaces":{"CgroupInum":4026531835,"IpcInum":4026531839,"MntInum":4026531841,"NetInum":4026531840,"PidChildInum":4026531836,"PidInum":4026531836,"TimeChildInum":4026531834,"TimeInum":4026531834,"UserInum":4026531837,"UtsInum":4026531838},"Parent":{"Ktime":7740000000,"Pad":0,"Pid":1341},"ParentFlags":0},"Process":{"AUID":4294967295,"Args":"-c\u0000daad67d6256338e99345908fdefdf9196110f7c516cd8ebce941b9ca5d6a256f\u0000-n\u0000k8s_prometheus_prometheus-k8s-0_openshift-monitoring_3a3be45e-5c76-4cc7-b0e5-ba358bd99479_0\u0000-r\u0000/usr/bin/runc\u0000-p\u0000/tmp/pidfile3593012291\u0000-e\u0000-l\u0000/tmp/crio-log-daad67d6256338e99345908fdefdf9196110f7c516cd8ebce941b9ca5d6a256f3065360050\u0000--socket-dir-path\u0000/var/run/crio\u0000--log-level\u0000info\u0000--sync\u0000--log-global-size-max\u000016777216\u0000-T\u00003\u0000-s\u0000--exec-process-spec\u0000/tmp/exec-process-2433818724\u0000--runtime-arg\u0000--root=/run/runc","Envs":null,"Filename":"/usr/bin/conmon","Flags":20481,"Ino":0,"Ktime":288251396707535,"NSPID":0,"Nlink":0,"PID":2332806,"SecureExec":0,"Size":554,"TID":2332806,"UID":0,"User":{"Name":""}}}}}
```
</details>

kkourt:
I see, thanks. 

Can you walk me through a use-case for the recording and replay feature? Is this a feature that targets Tetragon developers or users?

sglushko:
> Can you walk me through a use-case for the recording and replay feature? Is this a feature that targets Tetragon developers or users?

The primary use case is for developers to debug complex issues, such as refcounter bugs in cache dumps. I build a custom image with recording enabled—and as you requested, this functionality is not included in the default Tetragon binary. The workflow is similar to the one I linked in the PR description (https://github.com/cilium/tetragon/issues/4497#issuecomment-3836611060), though the steps have changed slightly since the code was removed from the main binary. Essentially, you run the custom image in a cluster to record events, copy the file locally, and then replay it in a local Docker container. This provides a reliable way to reproduce and debug issues locally.

pkg/synthetic/serializer.go
```diff
@@ -0,0 +1,361 @@
...
+func (Serializer) Marshal(v any) ([]byte, error) {
+	ifaceVal := reflect.ValueOf(&v).Elem()
+	wrapped, err := wrapValue(ifaceVal)
+	if err != nil {
+		return nil, err
+	}
+	return json.Marshal(wrapped)
+}
+
+// Unmarshal deserializes JSON with type wrappers.
```
kkourt, line 33: What's the reason to use json? 

Would it make sense to use gob (https://pkg.go.dev/encoding/gob) in this case?
We know that both producer and consumer will be written in go.
My motivation here is to avoid adding any marshall/unmarshall code.

sglushko: I looked into gob, and it also requires type registration for interfaces. However, unlike my current implementation, it requires registration during encoding as well. This isn't a major issue, but it might complicate recording events for new sensors.


The pros of gob are that it’s more compact and likely faster. On the downside, it would need to be implemented and isn't human-readable. One of the requirements was the ability to manually create events, reorder them, and easily inspect the file to get a high-level understanding of the event flow. Given that this is all for debugging, size and speed are less important than human readability. What do you think?

kkourt: > It requires registration during encoding as well. This isn't a major issue, but it might complicate recording events for new sensors.

I would say that part is fine. 

> Given that this is all for debugging, size and speed are less important than human readability. What do you think?

Yes, agreed. With gob we lose readability, which is a good property to have.

sglushko: I’ve looked at the current implementation, and it seems we could use any serializer that implements the Codec interface with almost no changes. We could create a gob implementation for it and plug it in wherever needed using DI.

kkourt: I think JSON is OK. I was hoping to avoid adding the serializer code, but your argument about using JSON for readability makes sense to me.

PR #4600, opened 2026-02-02
Adjusting raw_syscalls policy to use syscall type

### LINKED ISSUE #539: properly map syscall ids to syscall name at the CLI
#483 defined SyscallNames for dwarin, but that's probably not what we want. 

Commit https://github.com/cilium/tetragon/commit/f56c0f5e35455208f8190285d70f4f1de8bb0595, added

[tetragon/pkg/encoder/encoder.go](https://github.com/cilium/tetragon/blob/27facf9a04c32e3d6f391ed824f915616e65d10b/pkg/encoder/encoder.go#L287)

Line 287 in [27facf9](https://github.com/cilium/tetragon/commit/27facf9a04c32e3d6f391ed824f915616e65d10b)
 if name := syscallinfo.GetSyscallName(int(sysID)); name != "" { 

Which translates system call ids to names, and it uses that to figure out the arguments as well. So (at least in the above case) we would need the syscallNames of Linux (rather than dwarvin). Not sure what the best solution for this. It would probaby be better if we did this translation (id->name) on the agent.

bar234:
Using the syscall64 type ensures the client maps the syscall to the correct name.

Fixes #539 

### Description
If a client running on a different architecture from the agent received events from the `raw_syscalls.yaml` policy, it would map the syscall id to the name based on the client ABI.

```
syscall x86-vm /home/berose/Development/tetragon/tetragon mlock(unsigned long start=?, size_t len=?)(228)
```
(228 should be `clock_gettime` on x86)

By changing the policy's syscall argument from `int64` to `syscall64`, the ABI of the agent is provided in the event passed to the client and the correct syscall name is displayed.

```
syscall x86-vm /home/berose/Development/tetragon/tetragon x64/clock_gettime(228)
```

### DIFF
--- a/examples/tracingpolicy/raw_syscalls.yaml
+++ b/examples/tracingpolicy/raw_syscalls.yaml
@@ -9,7 +9,7 @@ spec:
       # args: add both the syscall id, and the array with the arguments
       args:
         - index: 4
-          type: "int64"
+          type: "syscall64"
       #  - index: 5
       #selectors:
       #- matchPIDs:


olsajiri:
the signed-off-by must have same email as patch author, please change, otherwise looks good

bar234:
I think I updated the commit correctly. Sorry about that, wasn't paying attention to which git environment I was using. Let me know if you still see a mismatch or any other issues!

PR #4601, opened 2026-02-03
pkg/sensors: reject unsupported returnArgAction values

shuv-amp:
This series tightens `ReturnArgAction` handling for kprobe return paths.

The return-path implementation only has explicit behavior for the
socket-tracking actions (`TrackSock` and `UntrackSock`), but the schema
and examples also suggested that `Post` was meaningful there. Based on
the review feedback in this PR, this series rejects unsupported values
instead of accepting a silent no-op.

Changes:
- reject unsupported `returnArgAction` values in kprobe return-path validation
- clarify that `returnArg` exports the return value, while `returnArgAction` is only for socket-tracking actions on that value
- regenerate CRDs/docs and sync the vendored `pkg/k8s` CRD copies
- remove redundant `returnArgAction: "Post"` from examples

Policies that explicitly set `returnArgAction: "Post"` should omit the
field instead.

Testing:
- `go test ./apis/cilium.io/v1alpha1 -count=1` from `pkg/k8s/`
- the equivalent pre-cleanup stack passed the full PR CI before this reviewer cleanup

```release-note
ReturnArgAction now only accepts TrackSock/UntrackSock for kprobe return args; unsupported values like Post are rejected to prevent silent no-ops.
```

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -1253,6 +1253,10 @@ FUNC_INLINE int generic_retprobe(void *ctx, struct bpf_map_def *calls, unsigned 
 #if defined(__LARGE_BPF_PROG) && (defined(GENERIC_KRETPROBE) || defined(GENERIC_FEXIT))
 		struct socket_owner owner;
 
+		/* ReturnArgAction only has return-path handling for socket tracking.
+		 * Userspace validation limits it to ACTION_TRACKSOCK and
+		 * ACTION_UNTRACKSOCK.
+		 */
 		switch (config->argreturnaction) {
 		case ACTION_TRACKSOCK:
 			owner.pid = e->current.pid;
--- a/contrib/upgrade-notes/latest.md
+++ b/contrib/upgrade-notes/latest.md
@@ -17,7 +17,10 @@ See the [Stack Traces](https://tetragon.io/docs/concepts/tracing-policy/selector
 
 ### TracingPolicy (k8s CRD)
 
-* TBD
+* `returnArgAction` no longer accepts `Post`. Use `returnArg` to include the
+  return value in events, and omit `returnArgAction` for the default return-value
+  behavior. Only `TrackSock` and `UntrackSock` are supported. Existing policies
+  that set `returnArgAction: "Post"` should remove the field.
 
 ### Events (protobuf API)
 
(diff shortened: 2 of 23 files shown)


shuv-amp:
Updates pushed to address CI failures:

- Synced `vendor/` directory.
- Bumped CRD schema version in `version.go`.
- Regenerated `tracing-policy.md` using the project's make target.

All checks are passing locally. The `validate labels` check currently fails as expected, pending the addition of a `release-note/*` label.

kkourt (commented):
Thanks!

The change makes sense to me overall, but I would like to get some feedback from the original author of this, which was introduced in https://github.com/cilium/tetragon/pull/1060/changes/f85118919556cbbae66a631f9175b951bac2136b#diff-bc54ba3950ae7bb8c718f46e20d8c315f8ff7661d6950fe362276e8160b26227.

@kevsecurity any thoughts on whether we need a Post action in `ReturnArgAction`?
I think it would be useful to clarify the intended semantics here.

@shuv-amp please split the PR into multiple commits so that it's easier to review? see https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/#commit-changes for details.

kevsecurity:
> @kevsecurity any thoughts on whether we need a Post action in `ReturnArgAction`? I think it would be useful to clarify the intended semantics here.

It was added for socket tracking where we needed to be able to store the return value from ```sk_alloc()```. We don't need to allow Post etc.

Am I right there are two ways of specifying/interacting with return values? If so, we should probably combine to reduce confusion.

kevsecurity (changes requested):
Please add descriptions to each commit to explain what each commit is doing; it'll save future readers from having to read the code to understand why/how it achieved. Thanks!

bpf/process/generic_calls.h
```diff
@@ -1255,6 +1259,8 @@ FUNC_INLINE int generic_retprobe(void *ctx, struct bpf_map_def *calls, unsigned
 		case ACTION_UNTRACKSOCK:
 			map_delete_elem(&socktrack_map, &ret);
 			break;
+		default:
+			break;
```
mtardy: Looks like these lines are doing nothing

mtardy (commented):
I'll agree with Kev maybe finish to format this PR https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/ and re-ask for a review, thanks!

shuv-amp:
Addressed the review feedback:
- removed the no-op default arm in `generic_retprobe()`
- added descriptions to each commit and kept the series split by logical change

Please take another look when you have time.

kevsecurity:
Hello. This is really good, thanks!
I think the first commit message mentions removing the default arm, but that isn't in the commit itself (I think because you originally added it in your commit, so removing it from it results in no change). So I think you can lose that one sentence and then we're good.

kevsecurity (approved):
LGTM

kkourt (approved):

PR #4606, opened 2026-02-04
Enable sleep-able USDT argument reading

andrewstrohman:
Like with uprobe, USDT is used as a hook for user space programs. User space programs can have their memory paged out to swap. This means that at the point where a USDT hook fires, all the memory of the user space program may not be loaded into memory. As such, the bpf program might not be able to dereference argument pointers without being able to do a page fault. Sleep-able bpf programs allow us to trigger a page fault when user memory is not paged in, so that we can reliably access the data that the argument pointers point at.

This change attempts to do as much code (and map) reuse as possible with respect to the existing uprobe implementation. Some files/functions are renamed to reflect that they apply to all user space programs instead of just uprobe.

### DIFF
--- a/contrib/verify/verify_test.go
+++ b/contrib/verify/verify_test.go
@@ -104,6 +104,13 @@ func TestVerifyTetragonPrograms(t *testing.T) {
 			continue
 		}
 
+		// Check if bpf_copy_from_user_str is available
+		if (strings.HasPrefix(fileName, "bpf_generic_uprobe") ||
+			strings.HasPrefix(fileName, "bpf_generic_usdt")) &&
+			!bpf.HasKfunc("bpf_copy_from_user_str") {
+			continue
+		}
+
 		spec, err := ebpf.LoadCollectionSpec(tetragonDir + "/" + fileName)
 		require.NoError(t, err, "failed to parse elf file into collection spec")
 		require.NotNil(t, spec, "collection spec should not be nil")
--- a/pkg/sensors/tracing/genericusdt.go
+++ b/pkg/sensors/tracing/genericusdt.go
@@ -42,6 +42,7 @@ type observerUsdtSensor struct {
 
 type usdtHas struct {
 	sleepableOffload bool
+	sleepablePreload bool
 }
 
 var (
@@ -117,7 +118,7 @@ func createGenericUsdtSensor(
 	hasSetAction := false
 
 	for _, usdt := range spec.Usdts {
-		ids, err = addUsdt(&usdt, &in, ids)
+		ids, err = addUsdt(&usdt, &in, ids, &has)
 		if err != nil {
 			return nil, err
 		}
@@ -170,6 +171,7 @@ func createMultiUsdtSensor(multiIDs []idtable.EntryID, policyName string, has us
 		SetPolicy(policyName)
 
 	load.SleepableOffload = has.sleepableOffload
+	load.SleepablePreload = has.sleepablePreload
 
 	progs = append(progs, load)
 
@@ -186,6 +188,12 @@ func createMultiUsdtSensor(multiIDs []idtable.EntryID, policyName string, has us
 		sleepableOffloadMap := program.MapBuilderProgram("write_offload", load)
 		sleepableOffloadMap.SetMaxEntries(sleepableOffloadMaxEntries)
 		maps = append(maps, sleepableOffloadMap)
+	}
+
+	if has.sleepablePreload {
+		sleepablePreloadMap := program.MapBuilderProgram("sleepable_preload", load)
+		sleepablePreloadMap.SetMaxEntries(sleepablePreloadMaxEntries)
+		maps = append(maps, sleepablePreloadMap)
 	}
 
 	if option.Config.EnableCgTrackerID {
@@ -232,6 +240,7 @@ func createUsdtSensorFromEntry(usdtEntry *genericUsdt,
 		SetPolicy(usdtEntry.policyName)
 
 	load.SleepableOffload = has.sleepableOffload
+	load.SleepablePreload = has.sleepablePreload
 
 	progs = append(progs, load)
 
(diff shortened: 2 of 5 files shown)


andrewstrohman:
This PR is ready for review. I've converted to draft until its dependency is merged.

bpf/process/generic_calls.h
```diff
@@ -588,7 +588,10 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 
 #if defined(GENERIC_TRACEPOINT) || defined(GENERIC_USDT)
 	a = (&e->a0)[index];
-	extract_arg(config, index, &a, false);
+	if (am & ARGM_PRELOAD)
+		a = get_preload_arg(ctx, ty);
```
kkourt, line 592: Does this code apply for GENERIC_TRACEPOINT? Or do we need to splt the GENERIC_TRACEPOINT and GENERIC_USDT implementation?

contrib/tester-progs/usdt-args.c
```diff
@@ -0,0 +1,84 @@
...
+	fprintf(stderr, "Usage: %s <type> <val>\n", argv0);
+	fprintf(stderr, "type can be one of: string\n");
+}
+
+/* Page out a memory region containing 'data' of length 'len'.
+ * Returns the address of the paged-out region.
+ * The temp file backing is used to work around the fact that our test VMs
+ * don't have any disk backed storage to use for swap.
+ */
+void *pageout(void *data, size_t len) {
```
andrewstrohman: This is a copy/paste of what's in contrib/tester-progs/uprobe-test.c. I needed this function in yet another test program as well, so I've broken this logic out into a library, so the code can be re-used, instead of repeated. See here: https://github.com/cilium/tetragon/pull/4666/changes/f7d4173b9d915f0f064fcbae4e3d8016be9ab1a0

I'd like this to use libtester.so here as well, to eliminate the repetition.

bpf/process/generic_calls.h
andrewstrohman: > Does this code apply for GENERIC_TRACEPOINT?

It does not apply to GENERIC_TRACEPOINT. ARGM_PRELOAD can only be set for uprobe/USDT.

> Or do we need to splt the GENERIC_TRACEPOINT and GENERIC_USDT implementation?

I'm not sure why we need to split it. I think that splitting it up would reduce instruction count for tracepoint, but I don't think there is a functional need to do it. But, maybe I'm missing something.

With that being said, I don't see an issue with splitting it (with `#if defined`), if we feel that's the best path forward. Just let me know.

andrewstrohman:
@kkourt I've made two changes since you approved.

The first is to reject policies that specify multiple preload argument per hook point. I'm following [this](https://github.com/cilium/tetragon/pull/4635/changes/98e7a31e55917baabd6008bdd76a6f9ee8213478) as an example.

The second was to move the `pageout` function into a library for re-use. I found that I need this function for [this PR](https://github.com/cilium/tetragon/pull/4666) as well, so I didn't want to continue duplicating that code.

pkg/sensors/tracing/genericusdt.go
```diff
@@ -365,12 +386,30 @@ func addUsdt(spec *v1alpha1.UsdtSpec, in *addUsdtIn, ids []idtable.EntryID) ([]i
 				cfgArg.Signed = 0
 			}
 
+			if argType == gt.GenericStringType {
+				if !bpf.HasKfunc("bpf_copy_from_user_str") {
+					return nil, fmt.Errorf("can't preload string for argument %d", cfgIdx)
+				}
+
+				if preload {
```
olsajiri, line 394: would it be better to do it outside this if like for uprobes? https://github.com/cilium/tetragon/pull/4635/changes/a4d6c0f7901c049e334e589e22ca75671e596f2c

contrib/verify/verify_test.go
```diff
@@ -104,6 +104,11 @@ func TestVerifyTetragonPrograms(t *testing.T) {
 			continue
 		}
 
+		// Check if bpf_copy_from_user_str is available
+		if (strings.HasPrefix(fileName, "bpf_generic_uprobe") || strings.HasPrefix(fileName, "bpf_generic_usdt")) && !bpf.HasKfunc("bpf_copy_from_user_str") {
```
olsajiri: please split the line, AFAICS  we try to follow the 100 columns max

bpf/process/generic_calls.h
olsajiri: I stumbled on this hunk as well, I think we should factor `get_preload_arg` to be more readable.. maybe each sensor having its own version or something like that.. in future ;-)

olsajiri (commented):
looks good, left some minor comments

bpf/process/generic_calls.h
kkourt: Mostly for clarity, but I'm happy to leave it as is, and if we want, we can tackle in the future.

contrib/verify/verify_test.go
andrewstrohman: Ok, done.

pkg/sensors/tracing/genericusdt.go
andrewstrohman: I did it this way in order to avoid creating another variable.

The `has` variable applies to all USDT specs within a tracing policy spec. Currently, the `preload` variable is specific to the current USDT spec. So, in order to check the problematic condition in a new sequential (not nested) `if` statement, I would need an additional variable to signify whether I've already encountered preload variable for the current USDT spec.

Put another way, checking `preload` in a sequential (not nested) `if` would be ambiguous because we don't know if `preload` was true in a previous iteration of args, and `has.sleepablePreload` could be true because a preload arg was found in a different USDT spec within the same tracing policy.

kkourt (approved):

PR #4607, opened 2026-02-05
bpf: improve arg index masking

andrewstrohman:
This PR is meant to address some [feedback](https://github.com/cilium/tetragon/pull/4327#discussion_r2740137016) from a different [PR](https://github.com/cilium/tetragon/pull/4327). I initially developed this change within that original PR, but a another reviewer requested that this be handled in a seperate PR.

These changes replace mask constants used for array index masking with preprocessor macros for clarity and safety. The mask macros are tied to the size of the arrays that they help provide the indexes into. This way, we will have a compilation error instead of silent failures in the event that the array size is increased without also increasing the size of the corresponding mask.

While investigating this, I found that we had other mask related issues/confusions. For example, MAX_SELECTORS_MASK is frequently used for argument indexes. This mask just happens to be appropriate for argument indexes today, but the concept of max selectors is unrelated to the concept of max supported args. As such, these limits could deviate in the future.

Similarly, a new mask macro was created to distinguish between the number of arguments that can be reached from a function signature perspective vs the number of arguments that can be configured in the tracing policy.

Finally, EVENT_CONFIG_MAX_ARG is replaced everywhere with MAX_POSSIBLE_ARGS because these two values cannot deviate.

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -2274,7 +2274,7 @@ installfd(struct msg_generic_kprobe *e, int fd, int name, bool follow)
 	asm volatile("%[fd] &= %[mask];\n"
 		     : [fd] "+r"(fd)
 		     : [mask] "i"(MAX_POSSIBLE_ARGS_MASK));
-	if (fd > MAX_POSSIBLE_ARGS)
+	if (fd >= MAX_POSSIBLE_ARGS)
 		return 0;
 
 	if (!is_arg_ok(e, fd))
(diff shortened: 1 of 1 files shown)


kkourt:
Thanks Andy! 

If this depends on https://github.com/cilium/tetragon/pull/4327, can we move it to draft until #4327 is merged?

andrewstrohman:
> If this depends on https://github.com/cilium/tetragon/pull/4327, can we move it to draft until https://github.com/cilium/tetragon/pull/4327 is merged?

Yes, sounds good. Thanks for doing that.

Issue #4359, opened 2025-11-19
returnCopy does not update the appropriate arg value within the event


andrewstrohman:
### What happened? How can we reproduce this?

With the following policy:
```
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: ret-test
spec:
  kprobes:
    - call: "sys_read"
      syscall: true
      args:
#      - index: 0
#        type: "int"
      - index: 1
        type: "char_buf"
        returnCopy: true
        sizeArgIndex: 3
      - index: 2
        type: "size_t"
```
The args in the events look like:

```
"args":[{"bytes_arg":""},{"bytes_arg":"read string"}]
```

This is happening [here](https://github.com/cilium/tetragon/blob/b82408601fe54ece9ae1aa0e94b6d9fe52be3e5f/pkg/sensors/tracing/generickprobe.go#L1495-L1496). `enterArgs` is ordered based on the ordering of the spec file -- it's not indexed based on the args position within the function signature.

So this issue happens when a kprobe `returnCopy` arg specifies an index that is not the same as that arg's offset within the policy. 

retArg.GetIndex() has different semantics depending on what type of hookpoint is used. For uprobe and lsm, GetIndex() returns the position within the spec. For kprobe and usdt, GetIndex() returns the index value specified for the arg in the spec. Put another way, it's the arg position within the function signature.

This means that for uprobe, returnCopy works as expected because the retArg.GetIndex() returns the arg's position within `enterArgs`. But for kprobe, retArg.GetIndex() returns the index of the argument within the function signature, which might not coincide the arg's position within the spec file.

Perhaps we should change the semantics of kprobe argument index to be the same as uprobe.

### Tetragon Version

v1.7.0-pre.0-133-gb82408601

### Kernel Version

6.8.0-87-generic, but the issue is not specific to kernel version.

### Kubernetes Version

N/A

### Relevant log output

```shell

```

andrewstrohman:
I'm trying to address this as a part of https://github.com/cilium/tetragon/pull/4327

PR #4608, opened 2026-02-05
Fix issue with returnCopy

andrewstrohman:
This fixes a returnCopy issue where we were overwriting the wrong arg. We were confusing the two different meanings of index. It can mean the index of the arg within the function signature or it can mean the the position of the configured arg within the tracing policy.

This change allows returnCopy to work when index within the function signature does not match the index of the configured arg within the tracing policy.

fixes: https://github.com/cilium/tetragon/issues/4359

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -555,6 +555,52 @@ spec:
 	runKprobeObjectRead(t, readHook, checker, fd, fd2)
 }
 
+func TestKprobeObjectReadIdxMismatch(t *testing.T) {
+	fd, fd2, fdString := createTestFile(t)
+	pidStr := strconv.Itoa(int(observertesthelper.GetMyPid()))
+	readHook := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "sys-read"
+spec:
+  kprobes:
+  - call: "sys_read"
+    syscall: true
+    args:
+    - index: 1
+      type: "char_buf"
+      returnCopy: true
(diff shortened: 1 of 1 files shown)


andrewstrohman:
It turns out that this PR did not fix returnCopy, because [another PR](https://github.com/cilium/tetragon/pull/4494) raced with it, and fixed it with an earlier merge. I confirmed this by reverted the functional change in this PR, but keeping the test, and the test passed.

I don't see an issue with keeping both of these changes because there is no functional issue. But my change does make the logic introduced in https://github.com/cilium/tetragon/pull/4494 a little confusing now, because it's no long necessary to do that in order to determine the index to overwrite. I'm hesitant to revert this one because I think there is value in making the semantics of `GetIndex` consistent regardless of how we fix the return copy issue. @FedeDP what do you think?

FedeDP:
>  I'm hesitant to revert this one because I think there is value in making the semantics of GetIndex consistent regardless of how we fix the return copy issue.

I agree! thanks for looking into this btw; also, the changes in this PR will help with: https://github.com/cilium/tetragon/issues/4442 (i had similar changes in my local branch); specifically, it helps with second point for https://github.com/cilium/tetragon/issues/4442#issuecomment-3743826160.
Of course we need many more adjustments to fix that issue, but still it's a step in the right direction :)
Thank you very much!!

andrewstrohman:
OK, thanks. I've created [this PR](https://github.com/cilium/tetragon/pull/4633) for consideration. If we want to keep this one, then the other commit will be confusing because it's determining the configured index by iterating through all the args, but the configured index is now is available via `GetIndex`.

PR #4609, opened 2026-02-05
tracing: add `FileType` and `NotFileType` support

### LINKED ISSUE #736: Add the ability to specify file type (i.e. pipe) in matchArgs
Now only regular files are supported in matchArgs.

AritraDey-Dev:
### Description

Adds `FileType`and `NotFileType` operators to tetragon's  `matchArgs` selectors. Now user can choose to see events only for certain types of files, like pipes, sockets, or regular files.
**Example:**
Filter vfs_write events to only capture writes to pipes or sockets:
```
kind: TracingPolicy
metadata:
  name: "filter-pipes-and-sockets"
spec:
  kprobes:
  - call: "vfs_write"
    syscall: false
    args: [{index: 0, type: "file"}]
    selectors:
    - matchArgs:
      - index: 0
        operator: "FileType"
        values: ["pipe", "socket"]
```
Fixes #736 

### Changelog

```release-note
Add `FileType` and `NotFileType` selectors to filter events by file type (e.g., pipes, sockets, or regular files).
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/argument_types.md
+++ b/docs/content/en/docs/concepts/tracing-policy/argument_types.md
@@ -222,6 +222,7 @@ The `file` data type represents kernel `struct file` object and retrieves
 the file's full path.
 
 See general path limitations in [path retrieval limits](#pathlimits).
+Support for [file type filtering]({{< ref "selectors.md#file-type-filtering" >}}) is also available for this type.
 
 ## `dentry`
 
@@ -238,6 +239,7 @@ See general path limitations in [path retrieval limits](#pathlimits).
 
 The `path` data type represents kernel `struct path` object retrieves
 the related path.
+Support for [file type filtering]({{< ref "selectors.md#file-type-filtering" >}}) is also available for this type.
 
 <a name="pathlimits"></a>
 
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -120,6 +120,8 @@ The available operators for `matchArgs` are:
 - `Prefix`
 - `Postfix`
 - `Mask`
+- `FileType`
+- `NotFileType`
 
 **Further examples**
 
@@ -151,6 +153,32 @@ with the previous example.
     operator: "Prefix"
     values:
     - "/etc"
+```
+
+### File type filtering
+
+The `FileType` and `NotFileType` operators allow filtering based on the type of a
+file (e.g., regular file, pipe, socket, etc.). These operators can only be used
+with arguments of type `file` or `path`.
+
+Matching file types:
+- `sock` or `socket`: Socket
+- `lnk` or `link`: Symbolic link
+- `reg` or `regular`: Regular file
+- `blk` or `block`: Block device
+- `dir`: Directory
+- `chr` or `char`: Character device
+- `fifo` or `pipe`: FIFO/Pipe
+
+In the following example, we monitor `vfs_write` only for regular files.
+
+```yaml
+selectors:
+- matchArgs:
+  - index: 0
+    operator: "FileType"
+    values:
+    - "reg"
 ```
 
 ## Data filter


bpf/process/types/basic.h
```diff
@@ -958,6 +960,36 @@ filter_file_buf(struct selector_arg_filter *filter, struct string_buf *args)
...
+		// Use probe_read to avoid verifier unbounded memory access error
+		// Address is: args->buf (start of path) + args->len (end of path) + 4 (sizeof flags)
+		probe_read(&mode, sizeof(mode), (void *)((char *)args->buf + args->len + 4));
+
+		__u32 *v = (__u32 *)&filter->value;
+#pragma unroll
+		for (int i = 0; i < MAX_MATCH_VALUES; i++) {
+			u32 val = 0;
+
+			__builtin_memcpy(&val, &v[i], sizeof(val));
```
andrewstrohman: Why are we doing this copy? Can we avoid the copy with the following?

```
index 6a9a250da..af1024d21 100644
@@ -977,12 +977,9 @@ filter_file_buf(struct selector_arg_filter *filter, struct string_buf *args)
                __u32 *v = (__u32 *)&filter->value;
 #pragma unroll
                for (int i = 0; i < MAX_MATCH_VALUES; i++) {
-                       u32 val = 0;
-
-                       __builtin_memcpy(&val, &v[i], sizeof(val));
-                       if (val == 0)
+                       if (v[i] == 0)
                                break;
-                       if ((mode & 0170000) == val) {
+                       if ((mode & 0170000) == v[i]) {
                                match = 1;
                                break;
                        }
```

pkg/selectors/kernel.go
```diff
@@ -805,6 +814,34 @@ func writePostfixStrings(k *KernelSelectorState, values []string, ty uint32) err
...
+	"reg":     32768,
+	"regular": 32768,
+	"blk":     24576,
+	"block":   24576,
+	"dir":     16384,
+	"chr":     8192,
+	"char":    8192,
+	"fifo":    4096,
+	"pipe":    4096,
+}
```
andrewstrohman, line 847: I think it would be better to use the syscall package, instead of hard coding these.

```
andy@builder:~/src$ go run test-2.go
CHR:8192
DIR:16384
IFO:4096
LNK:40960
MT:61440
REG:32768
SOCK:49152
BLK:24576
andy@builder:~/src$ cat test-2.go
package main

import (
        "fmt"
        "syscall"
)

func main() {
        fmt.Printf("CHR:%d\n", syscall.S_IFCHR)
        fmt.Printf("DIR:%d\n", syscall.S_IFDIR)
        fmt.Printf("IFO:%d\n", syscall.S_IFIFO)
        fmt.Printf("LNK:%d\n", syscall.S_IFLNK)
        fmt.Printf("MT:%d\n", syscall.S_IFMT)
        fmt.Printf("REG:%d\n", syscall.S_IFREG)
        fmt.Printf("SOCK:%d\n", syscall.S_IFSOCK)
        fmt.Printf("BLK:%d\n", syscall.S_IFBLK)
}

```

bpf/process/types/basic.h
```diff
@@ -958,6 +960,36 @@ filter_file_buf(struct selector_arg_filter *filter, struct string_buf *args)
...
+			if (val == 0)
```
andrewstrohman: We are not writing a zero value at the end of the list of match values to signal the list of match values is done.

The way to confirm that we should stop iteration is to check the iteration against `filter->vallen`. See an example [here](https://github.com/cilium/tetragon/blob/bb17a8f6d37946d1e45f371f549696055b6be247/bpf/process/types/basic.h#L1537-L1538).

bpf/process/types/basic.h
```diff
@@ -958,6 +960,36 @@ filter_file_buf(struct selector_arg_filter *filter, struct string_buf *args)
...
+		// size + sizeof(u32) + sizeof(u16)
+		// mode is at the end, so we can access it using the length of the string
+		// plus the size of the flags.
+		u16 mode = 0;
+		// 4 bytes for length, plus string length, plus 4 bytes for flags
+		// u32 *flags = (u32 *)&args->buf[args->len];
+
+		// Use probe_read to avoid verifier unbounded memory access error
+		// Address is: args->buf (start of path) + args->len (end of path) + 4 (sizeof flags)
+		probe_read(&mode, sizeof(mode), (void *)((char *)args->buf + args->len + 4));
```
andrewstrohman: We can end up here for type fd_ty. In that case, the mode is not positioned like this (or available at all), because that type doesn't use `store_path` for reading the arg.

andrewstrohman:
For commit message with subject: "tests: Add e2e tests for FileType filtering", I don't think those added unit tests are e2e tests. So, I think we should change the phrasing from "e2e tests" to "unit tests".

bpf/process/types/basic.h
AritraDey-Dev: Thanks! I'd love to simplify it, but since it's a 'packed' struct, the BPF verifier might complain about unaligned access. We usually use `memcpy` as a safe pattern here to avoid those errors. Since the compiler simplifies it anyway, I thought this was safer. Hope that's cool!

bpf/process/types/basic.h
AritraDey-Dev: Thanks! updated the loop to check `filter->vallen` instead of the zero value.

bpf/process/types/basic.h
AritraDey-Dev: Yeah, `fd_ty` is risky bcoz it doesn't always have the mode info.  I updated the Go control plane validation (pkg/selectors/kernel.go) to explicitly disallow `FileType` and `NotFileType` selectors on 
`fd` type arguments.. Now it only runs on `file` and `path` types where we know it's safe.

AritraDey-Dev: https://github.com/cilium/tetragon/pull/4609/changes/5fdce95758552c1b9fe47a631f2f81b802cee6ce#diff-27cf3aef45267b713359a8b5e3cbf2932e0d809290418cc042c7f000d01dbdd2L961

bpf/process/types/basic.h
```diff
@@ -958,6 +960,39 @@ filter_file_buf(struct selector_arg_filter *filter, struct string_buf *args)
 	case op_filter_str_notpostfix:
 		match = filter_char_buf_postfix(filter, args->buf, args->len);
 		break;
+	case op_filter_file_type:
+	case op_filter_not_file_type: {
```
kkourt: Can you please move the code here in a new inline function as is done with the other operators above?

pkg/sensors/tracing/kprobe_filetype_test.go
```diff
@@ -0,0 +1,484 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
+	pidStr := strconv.Itoa(int(observertesthelper.GetMyPid()))
+
+	tmpFile, err := os.CreateTemp("", "tetragon-filetype-test-*")
+	require.NoError(t, err)
+	defer os.Remove(tmpFile.Name())
+	defer tmpFile.Close()
```
kkourt: Would defer functions be called if a require.* assertion fails?
An alterntive would be to use t.Cleanup()

AritraDey-Dev: >Would defer functions be called if a require.* assertion fails?

Yes, defer functions still run even if a require check fails. But` t.Cleanup() `is the better way to do it in Go tests.

> An alterntive would be to use t.Cleanup()

Thanks! Updated it.

bpf/process/types/basic.h
mtardy: have you tried in the CI to simplify to see on which kernel it complains?

pkg/sensors/tracing/kprobe_filetype_test.go
mtardy, line 1: nit: I wouldn't add a new file for a new test given we have plenty, maybe you can put them with other type tests? Not sure where they are

pkg/sensors/tracing/kprobe_filetype_test.go
```diff
@@ -0,0 +1,492 @@
...
+	defer cancel()
+
+	pidStr := strconv.Itoa(int(observertesthelper.GetMyPid()))
+
+	tmpFile, err := os.CreateTemp("", "tetragon-filetype-notpipe-*")
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		os.Remove(tmpFile.Name())
+		tmpFile.Close()
+	})
```
mtardy: nit: might be worth considering `t.TempDir`, not sure it's better here since you need a file in the dir so even removing the manual cleanup bits might not actually help lol in LOC comparison.

pkg/sensors/tracing/kprobe_filetype_test.go
AritraDey-Dev: Done,Added in kprobe_test.go.

pkg/sensors/tracing/kprobe_filetype_test.go
AritraDey-Dev: Thanks!  switched to `t.TempDir()` ,cleanup will be much easier too.

bpf/process/types/basic.h
AritraDey-Dev: Just pushed the simplified version too. https://github.com/cilium/tetragon/pull/4609/changes/1bf20ee220fc8d8de699e119676b87735cc0a3d0. https://github.com/cilium/tetragon/actions/runs/21765781990/job/62801937574?pr=4609 CI passed with both versions!
 I'd still like to keep the memcpy though , it's just a safer pattern for BPF code. The [kernel docs](https://docs.kernel.org/core-api/unaligned-memory-access.html) suggest using `memcpy` when dealing with packed structs, and older kernels like 4.19 can be picky about alignment.

>If use of such macros is not convenient, another option is to use [memcpy()](https://docs.kernel.org/core-api/kernel-api.html#c.memcpy), where the source or destination (or both) are of type u8* or unsigned char*. Due to the byte-wise nature of this operation, unaligned accesses are avoided.

Also see https://docs.kernel.org/bpf/verifier.html

 The compiler optimizes it away anyway, so no perf hit... just a bit of extra safety for edge cases we might not catch in CI.

andrewstrohman: That doc is talking about unaligned access in general kernel programming.

I don't think the verifier will load the bpf program if unaligned access could happen. I think this is one of the things it verifies. On arm64, unaligned access would cause a bus error or alignment trap, and I think these things would cause a panic. Hence the verifier needs to confirm that this will not happen.

AritraDey-Dev: Thanks, I did some digging, the BPF verifier already ensures safety here as long as bounds are checked if `(args->len > MAX_STRING)`, so the memcpy isn't needed.Also in other parts of `basic.h` it use the simplified direct access pattern too,updated the patch :)

bpf/process/types/basic.h
```diff
@@ -973,6 +973,40 @@ struct string_buf {
...
+
+	probe_read(&mode, sizeof(mode), (void *)((char *)args->buf + args->len + 4));
+
+	__u32 *v = (__u32 *)&filter->value;
+#pragma unroll
+	for (int i = 0; i < MAX_MATCH_VALUES; i++) {
+		if (v[i] == 0)
+			break;
+
+		if ((mode & 0170000) == v[i])
```
olsajiri, line 1010: where's this coming from? please put some comments on why it's needed

AritraDey-Dev:
@kkourt @mtardy Hi! You can remove the "needs-rebase" label since all the conflicts have been resolved. Is there anything that is still blocking this PR from being merged?

mtardy:
> @kkourt @mtardy Hi! You can remove the "needs-rebase" label since all the conflicts have been resolved. Is there anything that is still blocking this PR from being merged?

hey I think this can be merged, thanks for the nice PR!

PR #4623, opened 2026-02-08
bpf: Auto-match IPv4 CIDRs against IPv4-mapped IPv6 addresses

### LINKED ISSUE #4618: IPv4 should be registered in both IPv4/IPv6 format to handle `AF_INET`/`AF_INET6` sockets

On Linux it's possible to use IPv4 address on a `AF_INET6` socket. But currently, Tetragon does not make the configuration of the TP easy.

Here is an event I have :

```json
{
    ...
    "args": [
        {
            "sock_arg": {
                "family": "AF_INET6",
                "type": "SOCK_STREAM",
                "protocol": "IPPROTO_TCP",
                "saddr": "172.31.2.173",
                "daddr": "172.31.2.204",
                "sport": 3000,
                "dport": 54430,
                "cookie": "18446626630453924992",
                "state": "TCP_ESTABLISHED"
              }
            }
        ],
    ...
}
```

In my policy, I try to filter `172.16.0.0/12`. It works fine on an IPv4 socket, but I receive an event when it's IPv6.
The workaround is to use both  `172.16.0.0/12` and  `::ffff:172.16.0.0/12` as filter which seems quite confusing.

I think what should be great, when adding an IPv4, is to add silently the `::ffff:<IPv4>` in the `addr6lpm_maps` (or the approriate IPv6 corresponding map, sorry if i'm incorrect,) to make sure the user only need to register the IPv4 once and can also handle both format.

### Tetragon Version

Latest

sayboras:
### Description

Dual-stack applications often use AF_INET6 sockets for IPv4 traffic
too, and the kernel then reports the addresses as ::ffff:x.x.x.x. An
IPv4 CIDR in SAddr or DAddr did not match those, so every CIDR had to
be listed twice, e.g. 172.16.0.0/12 and ::ffff:172.16.0.0/12.

When an IPv6 address misses the IPv6 LPM map and is IPv4-mapped, look
up its embedded IPv4 address in the IPv4 map instead. A single IPv4
CIDR then matches both socket families.

Fixes: https://github.com/cilium/tetragon/issues/4618

---

```release-note
bpf: Auto-match IPv4 CIDRs against IPv4-mapped IPv6 addresses
```


docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1862,6 +1862,41 @@ There are different types supported for each operator. In case of `matchArgs`:
 * InRange - In interval range
 * NotInRange - Not in interval range
 
+##### IPv4-Mapped IPv6 Addresses
+
+When filtering socket addresses with `SAddr` or `DAddr` operators, Tetragon automatically
+handles IPv4-mapped IPv6 addresses (`::ffff:x.x.x.x`). This is important for modern
+dual-stack networking where applications may use IPv6 sockets to handle both IPv4 and IPv6 traffic.
+
+**Automatic Normalization**: When you register an IPv4 CIDR (e.g., `172.16.0.0/12`), it will
```
mtardy: Hey could you make normal the bold stuff in the text? There are other occurrences later
```suggestion
Automatic Normalization: When you register an IPv4 CIDR (e.g., `172.16.0.0/12`), it will
```

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1862,6 +1862,41 @@ There are different types supported for each operator. In case of `matchArgs`:
...
+- matchArgs:
+  - index: 0
+    operator: "DAddr"
+    values:
+    - "10.0.0.0/8"      # Matches both 10.x.x.x and ::ffff:10.x.x.x
+    - "172.16.0.0/12"   # Matches both 172.16-31.x.x and ::ffff:172.16-31.x.x
+    - "::1/128"         # Matches only native IPv6 loopback
+```
+
+**Native IPv6**: Pure IPv6 addresses (like `2001:db8::1` or `::1`) are NOT affected by this
```
```suggestion
Native IPv6: Pure IPv6 addresses (like `2001:db8::1` or `::1`) are not affected by this
```

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1862,6 +1862,41 @@ There are different types supported for each operator. In case of `matchArgs`:
...
+normalization. They will only match IPv6 CIDR filters, not IPv4 CIDRs. The automatic matching
+only applies to IPv4-mapped IPv6 addresses (`::ffff:x.x.x.x`).
+
+**Why This Matters**: Modern applications using dual-stack networking may create `AF_INET6` sockets
```
mtardy: maybe instead of having "why this matters" we could put that into a note or caution?

mtardy (commented):
Hey thanks! Could you split the patches with the fix, the docs, the test?

bpf/process/types/tuple.h
```diff
@@ -39,4 +39,36 @@ FUNC_INLINE void write_ipv6_addr32(u32 *dest, u32 *src)
...
+		return false;
+
+	// Next 2 bytes must be 0, then 2 bytes of 0xffff (in little-endian: lower 32 bits = 0xffff0000)
+	// Mask out the IPv4 address (upper 32 bits) and check the pattern
+	return (addr[1] & 0xffffffffULL) == 0xffff0000ULL;
+}
+
+// Extract IPv4 address from IPv4-mapped IPv6 address
+// Returns the 32-bit IPv4 address from the last 4 bytes
+// On little-endian: addr[1] has format 0xDDCCBBAA_FFFF0000, we want 0xAABBCCDD
```
mtardy: you do not byte swap in the function, is this expected? maybe change the result of what we "want" if you still keep it network byte order

bpf/process/types/basic.h
```diff
@@ -1034,20 +1035,46 @@ filter_addr_map(struct selector_arg_filter *filter, __u64 *addr, __u16 family)
...
+			addrmap = map_lookup_elem(&addr4lpm_maps, &map_idx);
+			if (!addrmap)
+				return filter_addr_op_mod(filter->op, 0);
+
+			arg4.prefix = 32;
+			arg4.addr = extract_ipv4_from_mapped(addr);
+			arg = &arg4;
+
+			exists = (long)map_lookup_elem(addrmap, arg);
+			return filter_addr_op_mod(filter->op, exists);
```
mtardy: here you could `break` then?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1890,6 +1890,43 @@ There are different types supported for each operator. In case of `matchArgs`:
 * InRange - In interval range
 * NotInRange - Not in interval range
 
+##### IPv4-Mapped IPv6 Addresses
```
mtardy: if you introduce a new title, I'd introduce others subtitles for the rest of the content otherwise the text after looks like it's part of this section.

docs/content/en/docs/concepts/tracing-policy/selectors.md
mtardy, line 1921: Not for this PR specifically but I feel like this should be its own section, not under `### Multiple operators` and maybe not under `## Selector Semantics` (not sure on that one, does operator belong to semantic definition? maybe?). Either way it should become a `###` or `##` of tis own.

mtardy (commented):
I think overall LGTM, just a few nits mostly on docs to merge this :) sorry for the review delay

docs/content/en/docs/concepts/tracing-policy/selectors.md
sayboras: what do you think about having it at the end of Operator types section ?

Issue #4617, opened 2026-02-06
`NotEqual` operator on int does not work as expected


tdaudi:
## Description

The expected behavior for the `NotEqual` operator  is to be true when `not A and not B`. It works as expected when the operator is used with strings, but it behave differently when the values are integers.
When the operator is an integer, the filter seems doing `not A or not B`. So when the value is A, B return true, and when the value is B, A return true. So it never filter properly.

The workaround is to use `NotInMap`.

## Step to reproduce

Let's take an example policy 

```yaml
apiVersion: cilium.io/v1alpha1
kind: TracingPolicy
metadata:
  name: "test"
spec:
  kprobes:
  - call: security_file_permission
    syscall: false
    args:
    - index: 0
      type: uint64
      resolve: f_inode.i_sb.s_magic
    selectors:
    - matchActions:
      - action: Post
      matchArgs:
      - args: [0]
        operator: NotEqual
        values:
        - "7377"       
        - "1397703499"
      matchBinaries:
      - operator: Postfix
        values:
        - curl
```

Test with `curl https://google.com`. 

If the bug is triggered, you should receive events for `args: [0]` when it's equal to `7377` or `1397703499`

### Tetragon Version

Latest

### Kernel Version

NA

### Kubernetes Version

v1.34

AritraDey-Dev:
Yes, I can confirm this bug exists. I was able to reproduce the bug where the `NotEqual` operator for integer args fails when multiple values are provided.
The bug is in the ebpf comparison logic
https://github.com/cilium/tetragon/blob/46b1886880e7e6c16a7b1894cae2d4189f51ac28/bpf/process/types/basic.h#L1658-L1665
The same issue exists for[ 64 bit](https://github.com/cilium/tetragon/blob/46b1886880e7e6c16a7b1894cae2d4189f51ac28/bpf/process/types/basic.h#L1267-L1274), [16 bit](https://github.com/cilium/tetragon/blob/46b1886880e7e6c16a7b1894cae2d4189f51ac28/bpf/process/types/basic.h#L1407-L1414), [8 bit](https://github.com/cilium/tetragon/blob/46b1886880e7e6c16a7b1894cae2d4189f51ac28/bpf/process/types/basic.h#L1320-L1327) 

[...]

PR #4624, opened 2026-02-08
bpf: fix `NotEqual` operator for int args with multiple values

AritraDey-Dev:
### Description

The `NotEqual` logic was incorrectly passing events if the value didn't match at least one item in the list. Fixed to only pass if the value matches none of the items.For more detail see https://github.com/cilium/tetragon/issues/4617#issuecomment-3864082238, explained it here.

Tested with the policy from the issue using `security_file_permission` .

Fixes: #4617

### Changelog

```release-note
Fixed NotEqual operator for integer matchArgs with multiple values
```

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -7971,3 +7971,61 @@ spec:
 	err = jsonchecker.JsonTestCheck(t, checker)
 	require.NoError(t, err)
 }
+
+func TestKprobeNotEqualMultipleValues(t *testing.T) {
+	if !kernels.MinKernelVersion("5.0") {
+		t.Skip("Test fails on older kernels (like 4.19) due to missing BTF resolution for f_inode.i_sb.s_magic")
+	}
+
+	var doneWG, readyWG sync.WaitGroup
+	defer doneWG.Wait()
+
+	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
+	defer cancel()
+
+	tracingPolicy := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "test-notequal-multiple"
+spec:
+  kprobes:
+  - call: security_file_permission
+    syscall: false
+    args:
+    - index: 1
+      type: int
+    selectors:
+    - matchActions:
+      - action: Post
+      matchArgs:
+      - index: 1
+        operator: NotEqual
+        # curl naturally triggers security_file_permission with MAY_READ (4).
+        # We filter out MAY_WRITE (2) and MAY_READ|MAY_WRITE (6) to verify NotEqual allows 4.
+        values:
+        - "2"
+        - "6"
+      matchBinaries:
+      - operator: Postfix
+        values:
+        - curl
+`
+	createCrdFile(t, tracingPolicy)
+
+	obs, err := observertesthelper.GetDefaultObserverWithFile(t, ctx, testConfigFile, tus.Conf().TetragonLib)
+	if err != nil {
+		t.Fatalf("GetDefaultObserverWithFile error: %s", err)
+	}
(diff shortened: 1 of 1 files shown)


kevsecurity (approved):

bpf/process/types/basic.h
```diff
@@ -1266,11 +1266,13 @@ filter_64ty_selector_val(struct selector_arg_filter *filter, char *args)
 #endif // __LARGE_BPF_PROG
 		case op_filter_eq:
 		case op_filter_neq:
+		case op_filter_notin:
```
olsajiri: do we use NotIn for numbers? what's the difference to NotEqual? otherwise looks good, thanks

AritraDey-Dev: I usually see `NotEqual` used for numbers in the policies, but I added `NotIn` to the bpf code as well just for consistency since they are logically the same when checking a list of values. It also makes it easier to use the `is_not_operator()` check at the end of the func.

olsajiri: I think it's not allowed under k8s for argument filtering, is it?
in any case, let's fix just NotEqual in this patch and please add separate commit/pr for adding NotIn operator in there, thanks

AritraDey-Dev: the K8s API currently only exposes `NotEqual` for arguments. removed op_filter_notin from this patch.thanks

bpf/process/types/basic.h
```diff
@@ -1366,10 +1367,11 @@ filter_8ty_selector_val(struct selector_arg_filter *filter, char *args)
 		case op_filter_eq:
 		case op_filter_neq:
 			res = (*(u8 *)args == w);
-			if (filter->op == op_filter_eq && res)
-				return 1;
-			if (filter->op == op_filter_neq && !res)
-				return 1;
+			if (res) {
```
olsajiri: nit, I wonder something like that would be better, but I'm not convinced myself.. up to you ;-)

```
-                       if (res) {
-                               if (filter->op == op_filter_eq)
-                                       return 1;
-                               return 0;
-                       }
+                       if (res)
+                               return filter->op == op_filter_eq ? 1 : 0;
```

AritraDey-Dev: yes sure! this is much better.

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -7971,3 +7971,65 @@ spec:
...
+	createCrdFile(t, tracingPolicy)
+
+	obs, err := observertesthelper.GetDefaultObserverWithFile(t, ctx, testConfigFile, tus.Conf().TetragonLib)
+	if err != nil {
+		t.Fatalf("GetDefaultObserverWithFile error: %s", err)
+	}
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	testCurl := testutils.RepoRootPath("contrib/tester-progs/curl-test.sh")
```
olsajiri: we don't seem to have `contrib/tester-progs/curl-test.sh` in the repo.. what do I miss?

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -7971,3 +7971,65 @@ spec:
...
+      type: int
+    selectors:
+    - matchActions:
+      - action: Post
+      matchArgs:
+      - index: 1
+        operator: NotEqual
+        values:
+        - "2"
+        - "6"
```
olsajiri, line 8008: please add some comments on why the hook is executed during the curl and what's the expected argument values we filter for, thanks

pkg/sensors/tracing/kprobe_test.go
AritraDey-Dev: Ah, thanks. I was just using a sample script (`curl-test.sh`) to test this locally, so that part was just leftover. Since it doesn’t exist, it fell back to the `curl` command in the test. Thanks for pointing that out.

PR #4633, opened 2026-02-09
Revert "fix(pkg/sensors): fixed {k,u}retprobe args merge helper."

andrewstrohman:
This reverts commit 4157093815eebb134e7e8a5be23b835d88a255b6.

This commit being reverted raced another approach, Commit 3368d2f062d3 ("tetragon: make GetIndex() consistent"). While the commit being reverted in this patch is still functionally correct/valid (with or without 3368d2f062d3), it's no longer needed with 3368d2f062d3 applied because GetIndex() now returns the index within the tracing policy. As such we don't need to iterate through the args to determine the configured index.

So, revert this commit to avoid confusion in the future.


andrewstrohman:
@FedeDP please let me know what you think about this. I think it's OK from a functional perspective to keep this commit, but I also think that it will be confusing, because it's no longer necessary.

PR #4634, opened 2026-02-09
fix(process/cache): balance parent refcount ops during LRU eviction

dwindsor:
Update `process.Cache` to properly perform refcount operations in the eviction path.

When a process tracked by `process.Cache` gets LRU evicted due to pressure before its exit event arrives, we currently only [update metrics](https://github.com/cilium/tetragon/blob/9f3c6e1308ce815132d41384a5ec03ad875d74dc/pkg/process/cache.go#L151) and move on without doing any refcount operations.

When the child's exit event later arrives, the [exit handler ](https://github.com/cilium/tetragon/blob/9f3c6e1308ce815132d41384a5ec03ad875d74dc/pkg/grpc/exec/exec.go#L407)cannot find the child in the cache, so [GetParentProcessInternal](https://github.com/cilium/tetragon/blob/9f3c6e1308ce815132d41384a5ec03ad875d74dc/pkg/process/process.go#L506) returns `nil` and the `parent--` is [skipped](https://github.com/cilium/tetragon/blob/9f3c6e1308ce815132d41384a5ec03ad875d74dc/pkg/grpc/exec/exec.go#L502). This leads to an imbalance as `process.cacheGarbageCollector` will only clean those entries with balanced refcounts (refcnt=0). The entries remain in the cache longer than they should, flooding the cache with invalid entries and increasing eviction pressure. 

To fix this, we decrement the parent refcount in the eviction callback, but only for entries still marked `inUse`. An unconditional decrement in the callback would double-count `parent--` for normally-exiting processes,
driving the still-live parent's refcount to zero prematurely and causing it to be garbage-collected while still running.

This workload was used to produce this imbalance:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: process-churn
  namespace: default
spec:
  replicas: 3
  selector:
    matchLabels:
      app: process-churn
  template:
    metadata:
      labels:
        app: process-churn
    spec:
      containers:
        - name: churn
          image: alpine:latest
          command: ["/bin/sh", "-c"]
          args:
            - |
              while true; do
                for i in $(seq 1 50); do
                  /bin/sleep 30 & # ensure children stick around long enough to be evicted while inUse
                done
                sleep 0.5
              done
```

### Metrics
_Before Fix_
```
tetragon_process_cache_capacity         8192
tetragon_process_cache_size             8192
tetragon_process_cache_evictions_total  359930
tetragon_process_cache_misses{get}      978647
tetragon_process_cache_misses{remove}   96045
```

_After Fix_
```
tetragon_process_cache_capacity         8192
tetragon_process_cache_size             8192
tetragon_process_cache_evictions_total  217714
tetragon_process_cache_misses{get}      677400
tetragon_process_cache_misses{remove}   35563
```

Note the dramatic reduction in `tetragon_process_cache_evictions_total` and `tetragon_process_cache_misses{remove}`. Tests repeated w/ similar results.

```release-note
```

### DIFF
--- a/pkg/process/cache.go
+++ b/pkg/process/cache.go
@@ -145,19 +145,41 @@ func NewCache(
 	processCacheSize int,
 	GCInterval time.Duration,
 ) (*Cache, error) {
+	// Stash a reference to the Cache to refer to later in the eviction closure.
+	pm := &Cache{
+		size: processCacheSize,
+	}
+
 	lruCache, err := lru.NewWithEvict(
 		processCacheSize,
-		func(_ string, _ *ProcessInternal) {
+		func(_ string, evicted *ProcessInternal) {
 			processCacheEvictions.Inc()
+
+			// Perform parent-- for LRU-evicted entries that will never
+			// reach the exit handler.
+
+			// Skip non-inUse entries whose exit path already performed parent--
+			if evicted.color != inUse {
+				return
+			}
+
+			// Is the parent still in the cache?
+			if evicted.process == nil {
+				return
+			}
+			parent, ok := pm.cache.Peek(evicted.process.ParentExecId)
+			if !ok {
+				return
+			}
+
+			pm.refDec(parent, "parent--")
 		},
 	)
 	if err != nil {
 		return nil, err
 	}
-	pm := &Cache{
-		cache: lruCache,
-		size:  processCacheSize,
-	}
+
+	pm.cache = lruCache
 	pm.cacheGarbageCollector(GCInterval)
 	return pm, nil
 }


pkg/process/cache.go
```diff
@@ -145,19 +145,40 @@ func NewCache(
...
+	pm := &Cache{
+		size: processCacheSize,
+	}
+
 	lruCache, err := lru.NewWithEvict(
 		processCacheSize,
-		func(_ string, _ *ProcessInternal) {
+		func(_ string, evicted *ProcessInternal) {
+			// Perform parent-- for LRU-evicted entries that will never
+			// reach the exit handler
```
kkourt: Seeing this patch, I understand the above comment but  I think the resulting code looks confusing:

```go
			// Perform parent-- for LRU-evicted entries that will never
			// reach the exit handler
			processCacheEvictions.Inc()
           ...
```

How about:

```go
			processCacheEvictions.Inc()

			// Perform parent-- for LRU-evicted entries that will never
			// reach the exit handler
           ...
           
```

dwindsor:
> Also, this seems like a bug. Should we backport it?

Yeah I think so, should we do that concurrently with this fix or merge this and then backport?

pkg/process/cache.go
dwindsor: Updated, thanks!

dwindsor:
```
Process launched with PID: 7196
Looking for regex: \{\"process_exec\"\:\{\"process\"\:\{\"exec_id\"\:\".{16,30}\"\,.{0,1}\"pid\"\:7196\,.{0,1}\"uid\"\:[0-9]{0,9}\,.{0,1}\"binary\"\:\"C:\\\\Windows\\\\system32\\\\notepad.exe\"
Get-Content: D:\a\_temp\93c40fd5-5cad-4e8d-a38a-39932e18727b.ps1:39
Line |
```

Windows CI failing, seems unrelated

PR #4635, opened 2026-02-10
tetragon: assorted fixes

olsajiri:
- gt operator fix foe unsigned values
- fail uprobe policy with multiple preload arguments

### DIFF
--- a/pkg/sensors/tracing/uprobe_reg_test.go
+++ b/pkg/sensors/tracing/uprobe_reg_test.go
@@ -747,6 +747,115 @@ func TestUprobePtRegsPreloadMulti(t *testing.T) {
 	testUprobePtRegsPreload(t, true)
 }
 
+func testUprobePtRegsPreloadDouble(t *testing.T, multi bool) {
+	if !bpf.HasKfunc("bpf_copy_from_user_str") {
+		t.Skip("skipping")
+	}
+
+	testBinary := testutils.RepoRootPath("contrib/tester-progs/regs-override")
+
+	disableUprobeMulti := ""
+
+	if !multi {
+		disableUprobeMulti = `
+  options:
+    - name: "disable-uprobe-multi"
+      value: "1"`
+	}
+
+	// Put uprobe in test_3 function at:
+	//
+	//      static const char *test_3_string = "test_3_string_CASE";
+	//
+	//      "push   %%rbp\n"          /* +0  55                            */
+	//      "mov    %%rsp, %%rbp\n"   /* +1  48 89 e5                      */
+	//      "mov    %[str], %%rdi\n"  /* +4  48 8b 3d 96 2e 00 00          */
+	//      "pop    %%rbp\n"          /* +11 5d                            */
+	// -->  "mov    $0x0,%%rax\n"     /* +12 48 c7 c0 00 00 00 00          */
+	//      "mov    $0xff,%%rax\n"    /* +19 48 c7 c0 ff 00 00 00          */
+	//      "ret\n"                   /* +26 c3                            */
+	//      :
+	//      : [str] "m" (test_3_string)
+	//
+	// Make sure we retrieve data with eax value (1) as int argument
+	// and match the expected value via matchData.
+	//
+	// This test does the same thing as testUprobePtRegsPreload but instead
+	// of single uprobe with preload argument it adds two uprobes with preload
+	// argument.
+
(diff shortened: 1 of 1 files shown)


pkg/sensors/tracing/kprobe_test.go
```diff
@@ -7437,3 +7437,70 @@ func TestKprobeRangeIn(t *testing.T) {
...
+	createCrdFile(t, hook)
+
+	obs, err := observertesthelper.GetDefaultObserverWithFile(t, ctx, testConfigFile, tus.Conf().TetragonLib)
+	if err != nil {
+		t.Fatalf("GetDefaultObserverWithFile error: %s", err)
+	}
+	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
+	readyWG.Wait()
+
+	unix.Prctl(0xffff0, uintptr(value), 0, 0, 0)
```
andrewstrohman, line 7482: Why did you choose to pass 0xffff0 as the first arg? It doesn't look like it's a valid `option` value for `prctl`. Since it's the same value that is used for comparison, this feels a little confusing, because there is no relationship between these two things.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -619,7 +619,13 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 			}
 		}
 
-		has.sleepablePreload = has.sleepablePreload || preload
+		if preload {
+			// We do not support more than one preloaded argument, fail if there's more.
+			if has.sleepablePreload {
```
andrewstrohman: The `has` variable is used for all uprobe specs within a tracing policy spec. As such, I think this check makes us more prohibitive than necessary. I think we need to only prohibit multiple args per uprobe spec, not multiple args per tracing policy.

olsajiri: ok, I assume it probably works but let's add test for that.. spec with multiple uprobes on same place with preload argument

pkg/sensors/tracing/kprobe_test.go
olsajiri: yea, there's no relation.. I think I copied it from some other test, do we really care? we get sensor hit regardless of what prctl does later on

andrewstrohman: > do we really care?

No, I agree it's OK as-is. I just wanted to confirm my understanding. Thanks.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I wasn't considering a tracing spec where the multiple uprobes within it are attached to the exact same hook point, until you brought it up here. I was just thinking about different hook points within the same spec. But now that you bring this up, along these same lines, we could also have multiple, distinct tracing policies with the exact same hook point, which could cause a similar problem as well.

It looks like the order of execution for multiple uprobes at the same hook point is that they are executed in the reverse order that they are registered.

But, if we ever end up registering the probes in a different order, then the current version of the test may not catch a problem.

Let me explain with an example. Let's say there are two uprobes P1 and P2 that are attached to the exact same uprobe hook point. If, in the future, we change our attach ordering to:
1) attach P1 generic_sleepable_preload_cleanup
2) attach P2 generic_sleepable_preload_cleanup
3) attach P1 generic_uprobe_setup_event
4) attach P2 generic_uprobe_setup_event
5) attach P1 generic_sleepable_preload
6) attach P2 generic_sleepable_preload

Then the execution order would be the reverse of this. In this scenario, the test would not fail, when it should, because both probes store the same arg value in the map, at the same key:

```
    - index: 0
      type: "string"
      source: "pt_regs"
      resolve: "rdi"
 ```
If instead, we changed the test so that the second uprobe spec, which is attached to the same hook point, to have a different data configuration to extract a string from a different register, then we would could catch this problem. That being said, making this specific bug in the future is unlikely.

If you think this is worth caring about, I'm happy to make this change to the test in a follow-up. Just let me know.

olsajiri: I added that test to be sure such scenario actually works ale wrt the rest of the preload code,
but it might be good to have code that verifies we have correct order of the preload/cleanup attachments .. I think it's ok either to change this test later or add new one, thanks

PR #4639, opened 2026-02-11
policytest: new testing framework for Tetragon

kkourt:
## Examples

```
root@kind-bpf-next:/host/src/tetragon-trees/newtest# ./tetra  pt list
uprobe-generic [uprobes]
uprobe-override [uprobes]
root@kind-bpf-next:/host/src/tetragon-trees/newtest# ./tetra  pt run uprobe-override  uprobe-generic
P: uprobe-generic                                         ✅   1/1 scenario(s) succeeded
└S: execute nop and check events                          🟢
P: uprobe-override                                        ✅   1/1 scenario(s) succeeded
└S: execute uprobe-simple, check enforcement and events   🟢
```

```
 ./tetra pt run uprobe-generic uprobe-override
P: uprobe-generic                  ✅    1/1 scenario(s) succeeded
└S: execute nop and check events   🟢
P: uprobe-override                 ⏭️    uprobes cannot change registers
```

NB:  ⏭️ means that the policy was skipped, because there is no kernel support.

## New policy testing framework.

The idea  is to convert the observer tests into policy tests, that work by communicating with a tetragon agent over gRPC.

```go
// Trigger is a way to trigger a policy
type Trigger interface {
	Trigger(ctx context.Context) error
}

type Scenario struct {
	Name         string
	Trigger      Trigger
	EventChecker ec.MultiEventChecker
}

// Policies are represented as strings, because that's how they are loaded via gRPC
type Policy string

type Label string

// T defines a policy test
type T struct {
	// Name returns the name of the test
	Name string
	// Labels is a set of labels for the test
	Labels []Label

	// ShouldSkip returns true if the test is to be skipped for a certain environment
	ShouldSkip func(info *tetragoninfo.Info) bool

	// Policy generates a policy for this test
	Policy func(c *Conf) (Policy, error)

	// Scenarios returns a list of scenarios to test the generated policy
	Scenarios []func(c *Conf) *Scenario
}
```

There is an ergonomic builder for creating a policytest:
```go
var _ = policytest.NewBuilder("uprobe-generic").WithLabels("uprobes").WithPolicyTemplate(`
kind: TracingPolicy
metadata:
  name: "uprobe-generic"
spec:
  uprobes:
  - path: {{ testBinary "nop" }}
    symbols:
    - "main"
`).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
	nop := c.TestBinary("nop")
	upChecker := ec.NewProcessUprobeChecker("UPROBE_GENERIC").
			WithBinary(sm.Full(nop))).
		WithSymbol(sm.Full("main"))
	return &policytest.Scenario{
		Name:         "execute nop",
		Trigger:      policytest.NewCmdTrigger(nop),
		EventChecker: ec.NewUnorderedEventChecker(upChecker),
	}
}).RegisterAtInit()
```

Until we have proper CI for the polciytests, we can use the run them as observer tests using a wrapper:
```go
func TestUprobeGeneric(t *testing.T) {
	policytest.AllPolicyTests.DoObserverTest(t, "uprobe-generic")
}
```

Execution example:
```
./tetra policytest run uprobe-generic uprobe-generic
uprobe-generic      : &{Skipped:false Err:<nil> ScenariosRes:[{Name:execute nop CheckerErr:<nil>}]}
```
 
## TODO

- [ ] CI integration (in GHA)

### Tests

Add:
 - [x] test with enforcement
 - [x] test with ShouldSkip
 - [x] test with monitor mode

#### List policy tests

```
# filter by label
tetra policytest list --label foo
```

```
# list scenarios
tetra policytest list --scenarios
```

#### Run policy tests

```
tetra policytest run
	--grpc-server         # specify gRPC server
	--label               # run tests that match label (may be repeated)
	--all                 # run all tests
	--load-policy=true    # if set to false, the policy will not be loaded
	--unload-policy=true  # if set to false, the policy will not be unloaded
	--force-skippable     # execute tests even if they are skipped
```

#### Download binaries

The idea  is to build an OCI image with the test binaries

```
# copy test binaries from an OCI image
tetra policytest fetch-binaries --from oci://image --to ssh:///path/to/dir
tetra policytest fetch-binaries --from oci://image --to file://path/to/dir
```

### DIFF
--- a/pkg/testutils/policytest/runner.go
+++ b/pkg/testutils/policytest/runner.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"os"
 	"sync"
 	"time"
 
@@ -139,6 +140,13 @@ func (r *LocalRunner) AddPolicy(l *slog.Logger, test *T) (*PolicyHandler, error)
 		return nil, err
 	}
 	tpName := tp.TpName()
+
+	if r.conf.DumpPolicyPath != "" {
+		err := os.WriteFile(r.conf.DumpPolicyPath, []byte(pol), 0644)
+		if err != nil {
+			l.Warn("failed to dump policy", "err", err)
+		}
+	}
 
 	_, err = r.cli.Client.AddTracingPolicy(r.cli.Ctx, &tetragon.AddTracingPolicyRequest{
 		Yaml: string(pol),
--- a/cmd/tetra/policytest/policytest.go
+++ b/cmd/tetra/policytest/policytest.go
@@ -52,6 +52,7 @@ func listCmd() *cobra.Command {
 func runCmd() *cobra.Command {
 	cwd, _ := os.Getwd()
 	testBinsPath := filepath.Join(cwd, "contrib/tester-progs")
+	dumpPolicyPath := ""
 	monitorMode := false
 	cmd := cobra.Command{
 		Use:   "run",
@@ -79,8 +80,9 @@ func runCmd() *cobra.Command {
 				return ok
 			})
 			runner, err := policytest.NewLocalRunner(ctx, log, &policytest.Conf{
-				GrpcAddr: common.ServerAddress,
-				BinsDir:  testBinsPath,
+				GrpcAddr:       common.ServerAddress,
+				BinsDir:        testBinsPath,
+				DumpPolicyPath: dumpPolicyPath,
 			})
 			if err != nil {
 				return fmt.Errorf("failed to start local runner: %w", err)
@@ -101,6 +103,7 @@ func runCmd() *cobra.Command {
 	}
 	flags := cmd.Flags()
 	flags.StringVar(&testBinsPath, "bindir", testBinsPath, "path for test binaries directory")
+	flags.StringVar(&dumpPolicyPath, "dump-policy-path", dumpPolicyPath, "save the policy in the provided path")
 	flags.BoolVar(&monitorMode, "monitor-mode", monitorMode, "set the policy(-ies) in monitor mode before running the test(s)")
 	return &cmd
 }
--- a/pkg/testutils/policytest/conf.go
+++ b/pkg/testutils/policytest/conf.go
@@ -18,6 +18,9 @@ type Conf struct {
 
 	// Test run configuration
 	RunConf *RunConf
+
+	// Path to save the generated policy
+	DumpPolicyPath string
 }
 
 func (c *Conf) TestBinary(s string) string {


pkg/testutils/policytest/runner.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+
+	var res Result
+	for _, sc := range test.Scenarios {
+		scenario := sc(r.conf)
+
+		// set the scenario timeout to 10s
+		// TODO: make it configurable
+		ctx, cancel := context.WithTimeout(r.cli.Ctx, time.Second*10)
+		defer cancel()
```
dwindsor: This is interesting, IIUC `cancel()` won't get called until this function returns, wouldn't we need to wrap this for loop's body in a closure to have it trigger on each loop iteration? It probably won't matter because `test.Scenarios` won't be huge

pkg/testutils/policytest/runner.go
```diff
@@ -0,0 +1,449 @@
...
+//   - forward (runFwd)
+//   - check (runCheck)
+//
+// receive blocks on the event stream and passes all events to the forward goroutine.
+// forward either drops receieved events, or forwards them to the check goroutine
+// check runs the EventChecker that is part of the test.
+//
+// When a new scenario starts, a new check goroutine is created and the forwarder is configured to
+// forward events to it.
+func NewLocalRuner(
```
dwindsor: nit: typo "Runner"

pkg/testutils/policytest/runner.go
```diff
@@ -0,0 +1,449 @@
...
+	}
+	if grpcStatus.Code(err) != grpcCodes.DeadlineExceeded {
+		return fmt.Errorf("failed to unload policy: %w", err)
+	}
+
+	// deadline exceeded: let's try to reconnect to unload the policy
+	client, err = cli.NewClient(context.Background(), conf.GrpcAddr, time.Second*20)
+	if err != nil {
+		return fmt.Errorf("failed to create a new client to unload policy: %w", err)
+	}
```
will-isovalent, line 408: The retry client created here is never closed after use. Consider adding a `defer client.Close()` after the successful `NewClient` call to avoid leaking the gRPC connection.

pkg/testutils/policytest/runner.go
kkourt: Nice catch, thanks! Pushed a new version that moves the inner parts of the loop into a separate function

pkg/testutils/policytest/runner.go
kkourt: Fixed! thanks!

pkg/testutils/policytest/runner.go
kkourt: Good catch, thanks! Pushed a new version with a defer.

pkg/testutils/policytest/observertest.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Fatalf("failed to generate policy: %s", err)
+	}
+
+	policyFile, err := os.CreateTemp("", "teragon-policy-"+t.Name()+"-*.txt")
+	if err != nil {
+		t.Fatalf("failed to create polcy file: %s", err)
+	}
+	policyFile.WriteString(string(policyStr))
+	policyFile.Close()
+	policyFname := policyFile.Name()
```
mtardy, line 52: this is outside of the scope of this PR but could be get rid of this one day, why do we need to write the policy to a file to be passed to our observer? It always looked to me like we are doing write/read for nothing

pkg/testutils/policytest/observertest.go
```diff
@@ -0,0 +1,75 @@
...
+		BinsDir: testutils.RepoRootPath("contrib/tester-progs"),
+	}
+	policyStr, err := pt.Policy(conf)
+	if err != nil {
+		t.Fatalf("failed to generate policy: %s", err)
+	}
+
+	policyFile, err := os.CreateTemp("", "teragon-policy-"+t.Name()+"-*.txt")
+	if err != nil {
+		t.Fatalf("failed to create polcy file: %s", err)
```
mtardy, line 48: small harmless typo

kkourt:
@mtardy @dwindsor @kevsecurity any additional feedback?

pkg/testutils/policytest/dump.go
mtardy, line 1: nit: would have been cool to have an example of the output in the commit desc (like what we have in the PR body)

dwindsor (approved):

pkg/testutils/policytest/observertest.go
kkourt: I believe it's useful so that we can have the policy as part of the artifacts if the test fails.

pkg/testutils/policytest/dump.go
kkourt: good point, added.

PR #4640, opened 2026-02-11
docs: add missing index field to data section examples

dangome3:
### Description
The data section examples were missing the required index field. Added index: 0 to both examples in hooks.md and selectors.md to match the correct data section format.

When trying to load a policy with previous example Tetragon triggers this error:

```shell
level=error msg="Failed to execute tetragon" error="validation failed: spec.kprobes[0].data[0].index in body is required"
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -699,7 +699,8 @@ kernel `current_task` object.
 
 ```yaml
 data:
-- type: "string"
+- index: 0
+  type: "string"
   source: "current_task"
   resolve: "comm"
 selectors:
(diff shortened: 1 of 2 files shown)


andrewstrohman:
I don't think that it makes sense to be forced to specify an index in the data section of the config. I'm not sure what the meaning of it would be. If folks agree with this, I think that we should instead fix the validation logic to not require index for the `data` section of the config.

olsajiri:
> I don't think that it makes sense to be forced to specify an index in the data section of the config. I'm not sure what the meaning of it would be. If folks agree with this, I think that we should instead fix the validation logic to not require index for the `data` section of the config.

I think you're right, let's try to get rid of it, but meanwhile let's fix the docs

mtardy:
> I don't think that it makes sense to be forced to specify an index in the data section of the config. I'm not sure what the meaning of it would be. If folks agree with this, I think that we should instead fix the validation logic to not require index for the `data` section of the config.

Yep we can create a new type for this instead of reusing the matchArgs thing

PR #4642, opened 2026-02-11
bpf: use the correct index for pt_regs config

andrewstrohman:
The configuration for pt_regs is stored at the index that corresponds to its position within the configured args (data args have a later position than args). While the configuration validation requires the user to specify an index, it should be ignored.


andrewstrohman:
@olsajiri I added an additional commit to this PR that changes one of our existing tests to detect the error that is fixed in the original/first commit. This way, we can detect if the same problem comes back later.

PR #4645, opened 2026-02-12
operator: Replace hasAllRequiredFields requeue loop with a predicate

mlguerrero12:
Move the hasAllRequiredFields check from inside Reconcile to a Pod predicate filter. This prevents pods without required fields (IP, NodeName, etc.) from entering the work queue, avoiding unnecessary reconcile cycles and work queue churn during pod startup.


mlguerrero12:
Thanks @FedeDP. Only the validate labels job is failing now.

mtardy:
cc @michi-covalent, this might be something we want to do elsewhere!

PR #4654, opened 2026-02-13
fix: handle matchBinaries NotIn when execve_map misses process (#4641)

### LINKED ISSUE #4641: matchBinaries -> NotIn does not work for /usr/bin/runc

Actual: events with binary `/usr/bin/runc` are produced from a policy that has it in `matchBinaries` -> `NotIn` selector.
Expected: no such events are produced.

#### Environment
* k8s: EKS 1.32
* kernel: 6.1.159
* OS: Bottlerocket

#### Policy example

```yaml
kind: TracingPolicyNamespaced
metadata:
  name: "do-not-match-runc"
  namespace: my-namespace
spec:
  kprobes:
    - call: "sys_setuid"
      syscall: true
      args: [{index: 0, type: "int"}]
      selectors:
        - matchBinaries:
            - operator: NotIn
              values:
                - "/usr/bin/runc"
          matchActions:
            - action: Post
              rateLimit: "1m"
```

same for `sys_setgid` and `cap_capable`

#### Before

```shell
$ tetra tracingpolicy list | grep -E '(NAME|do-not-match-runc)'
ID   NAME                       STATE     FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE      NPOST   NENFORCE   NMONITOR
6    do-not-match-runc          enabled   6          my-namespace     generic_kprobe   9.27 MB        enforce   0       0          0
```

#### Perform dummy action
```shell
$ kubectl exec debug -c debug -- bash -c 'echo something'
```

#### After
```shell
$ tetra tracingpolicy list | grep -E '(NAME|do-not-match-runc)'
ID   NAME                       STATE     FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE      NPOST   NENFORCE   NMONITOR
6    do-not-match-runc          enabled   6          my-namespace     generic_kprobe   9.27 MB        enforce   6       0          0
```

#### Logs

_btw_ it's not possible to remove `node_labels` and `node_name` fields even when specifying `-f` since they're outside `process_kprobe`:

```shell
$ tetra getevents --policy-names do-not-match-runc -f policy_name,process.binary | jq -c 'del(.node_labels,.node_name)'

{"process_kprobe":{"process":{"binary":"/usr/bin/runc"},"policy_name":"do-not-match-runc"}},"time":"2026-02-11T21:20:36.661058273Z"}
{"process_kprobe":{"process":{"binary":"/usr/bin/runc"},"policy_name":"do-not-match-runc"}},"time":"2026-02-11T21:20:36.661059891Z"}
{"process_kprobe":{"process":{"binary":"/usr/bin/runc"},"policy_name":"do-not-match-runc"}},"time":"2026-02-11T21:20:36.661060021Z"}
{"process_kprobe":{"process":{"binary":"/usr/bin/runc"},"policy_name":"do-not-match-runc"}},"time":"2026-02-11T21:20:36.661059911Z"}
{"process_kprobe":{"process":{"binary":"/usr/bin/runc"},"policy_name":"do-not-match-runc"}},"time":"2026-02-11T21:20:36.661067622Z"}
{"process_kprobe":{"process":{"binary":"/usr/bin/runc"},"policy_name":"do-not-match-runc"}},"time":"2026-02-11T21:20:36.661167397Z"}
```

### Tetragon Version

1.6.0

### Kernel Version

6.1.159

### Kubernetes Version

EKS 1.32

### Bugtool

_No response_

### Relevant log output

```shell
level=info msg="adding namespaced tracing policy" name=do-not-match-runc info="do-not-match-runc (object:1/6856848f-d42b-4af5-ac40-de8f4322cbd8) (type:/)" namespace=my-namespace
level=info msg="Added kprobe" return=false function=__x64_sys_setuid override=false
level=info msg="Added generic kprobe sensor: /var/lib/tetragon/bpf_generic_kprobe_v61.o -> __x64_sys_setuid" override=false
level=info msg="Loading sensor" name=generic_kprobe
level=info msg="Loading kernel version 6.1.159"
level=info msg="Loaded generic kprobe program: /var/lib/tetragon/bpf_generic_kprobe_v61.o -> __x64_sys_setuid"
level=info msg="Loaded sensor successfully" sensor=generic_kprobe
```

### Anything else?

looks similar to https://github.com/cilium/tetragon/issues/1535

mann-ds:
Fixes #4641

### Description

root cause analysis:

* `matchBinaries` filtering is evaluated against `enter->bin` in `match_binaries()` (in [bpf/process/types/basic.h](https://github.com/cilium/tetragon/blob/main/bpf/process/types/basic.h)).
* for processes missing from `execve_map` (not found via normal lookup, common for processes started before Tetragon), code falls back to `event_find_curr_probe()` (in [bpf/process/bpf_process_event.h](https://github.com/cilium/tetragon/blob/main/bpf/process/bpf_process_event.h)).
* that fallback populated PID/ns/caps, but **did not populate** `curr.bin` with the executable path used by `matchBinaries`.
* so `NotIn` was evaluated against empty/stale binary data; and for `NotIn`, “not found in selector values” is treated as pass (`!match`), which let events through incorrectly.

why this matched the report:
* `runc` can be long-lived / pre-existing from Tetragon’s perspective, so it frequently hits this fallback path.
* result: false positives for policies like `matchBinaries: NotIn: ["/usr/bin/runc"]`.

changes:

* populate binary metadata in the process-not-found kprobe fallback path
* add `NotIn` regression tests for both perfring and early-exec cases

### Changelog

```release-note
fix: handle matchBinaries NotIn when execve_map misses process (#4641)
bpf: zero stale pkey and flags in event_find_curr_probe
tests: add matchBinaries NotIn tests for perfring and early-exec cases
```


bpf/process/bpf_process_event.h
```diff
@@ -345,17 +345,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
...
 
-	read_exe((struct task_struct *)get_current_task(), &msg->exe);
+	/* Reset binary info and populate from task's exe path for
+	 * matchBinaries filtering. Without this, matchBinaries selectors
+	 * (especially NotIn) won't work correctly when the process is
+	 * not found in execve_map.
+	 */
+	binary_reset(&curr->bin);
+	curr->bin.path_length = -1;
+	exe_len = read_exe(task, &msg->exe);
```
olsajiri: ok, I guess there are still some instructions left in filter tail call to burn ;-)

bpf/process/bpf_process_event.h
```diff
@@ -345,17 +345,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
...
+	if (exe_len > 0 && exe_len <= BINARY_PATH_MAX_LEN) {
+		int err;
+
+		asm volatile("%[len] &= 0xff;\n" : [len] "+r"(exe_len));
+		err = probe_read(curr->bin.path, exe_len, msg->exe.buf);
+		if (err == 0) {
+			curr->bin.path_length = exe_len;
+			if (exe_len > STRING_POSTFIX_MAX_LENGTH - 1)
```
olsajiri: could this and below be just for `__LARGE_BPF_PROG` ?

bpf/process/bpf_process_event.h
```diff
@@ -345,17 +345,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
...
+	 * (especially NotIn) won't work correctly when the process is
+	 * not found in execve_map.
+	 */
+	binary_reset(&curr->bin);
+	curr->bin.path_length = -1;
+	exe_len = read_exe(task, &msg->exe);
+	if (exe_len > 0 && exe_len <= BINARY_PATH_MAX_LEN) {
+		int err;
+
+		asm volatile("%[len] &= 0xff;\n" : [len] "+r"(exe_len));
```
olsajiri: is this necessary? I see similar code around without the asm masking for verifier, same for below

bpf/process/bpf_process_event.h
```diff
@@ -345,17 +345,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
...
+		err = probe_read(curr->bin.path, exe_len, msg->exe.buf);
+		if (err == 0) {
+			curr->bin.path_length = exe_len;
+			if (exe_len > STRING_POSTFIX_MAX_LENGTH - 1)
+				revlen = STRING_POSTFIX_MAX_LENGTH - 1;
+			else
+				revlen = exe_len;
+			asm volatile("%[revlen] &= 0x7f;\n" : [revlen] "+r"(revlen));
+			probe_read(curr->bin.end, revlen, msg->exe.end);
```
olsajiri: we use the `with_errmetrics` wrapper now for probe_read

olsajiri (commented):
looks good, left some comments, also please separate test from the actual fix, thanks

bpf/process/bpf_process_event.h
mann-ds: IIUC the whole `event_find_curr_probe` function is under `#ifdef __LARGE_BPF_PROG`, so should be already set.

bpf/process/bpf_process_event.h
mann-ds: thanks, I've wrapped it with `with_errmetrics`

bpf/process/bpf_process_event.h
mann-ds: Without adding these I get the following error when checking locally:

```shell
    logcapture.go:24: time=2026-02-17T12:37:35.170+01:00 level=INFO msg="Loading kernel version 6.18.8"
    cgtracker_test.go:240: 
        	Error Trace:	pkg/cgtracker/test/cgtracker_test.go:240
        	            	sensor generic_kprobe from collection lseek-test failed to load: failed prog bpf/objs/bpf_multi_kprobe_v61.o kern_version 397832 loadInstance: opening collection 'bpf/objs/bpf_multi_kprobe_v61.o' failed: program generic_kprobe_process_filter: load program: permission denied: 933: (85) call bpf_probe_read#4: R2 unbounded memory access, use 'var &= const' or 'if (var < const)' (889 line(s) omitted)
        	Test:       	TestCgTrackerPolicyFilter

```

I'm getting same errors in other places too, so it might be a local set-up/OS/kernel version issue, though.

So please let me know if you prefer to see these lines removed.

mann-ds:
> looks good, left some comments, also please separate test from the actual fix, thanks

Thank you for reviewing the PR, I've adjusted the contents and also removed the tests for now, so they could be added separately.

bpf/process/bpf_process_event.h
olsajiri: ok, let's keep it

olsajiri:
> Thank you for reviewing the PR, I've adjusted the contents and also removed the tests for now, so they could be added separately.

please keep the test in separate commit if possible

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -4428,6 +4431,202 @@ func TestKprobeMatchBinariesEarlyExec(t *testing.T) {
 	t.Error("events triggered by process executed before Tetragon should not be ignored because of matchBinaries")
 }
 
+// matchBinariesNotInPerfringTest checks that the matchBinaries NotIn operator
+// correctly filters events. For NotIn, events from binaries in the values list
+// should be filtered out, while events from other binaries should be allowed.
+func matchBinariesNotInPerfringTest(t *testing.T, values []string) {
```
mann-ds, line 4570: similar to `matchBinariesPerfringTest`

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -4428,6 +4431,202 @@ func TestKprobeMatchBinariesEarlyExec(t *testing.T) {
...
+		}
+	}
+	if !headEventExist {
+		t.Error("kprobe event triggered by /usr/bin/head should be present, unfiltered by the matchBinaries NotIn selector")
+	}
+}
+
+// TestKprobeMatchBinariesEarlyExecNotIn checks that the matchBinaries NotIn
+// operator filters events triggered by processes started before Tetragon.
+func TestKprobeMatchBinariesEarlyExecNotIn(t *testing.T) {
```
mann-ds, line 4664: similar to `TestKprobeMatchBinariesEarlyExec`

mann-ds:
@olsajiri commits were missing the `Signed-off-by: <...>` lines, now it should be good to go

mann-ds:
@olsajiri @kkourt do you think it's ready for merging?

bpf/process/bpf_process_event.h
```diff
@@ -345,17 +346,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
 {
 	struct task_struct *task = (struct task_struct *)get_current_task();
 	struct execve_map_value *curr;
+	__u32 exe_len;
+	__u32 revlen = STRING_POSTFIX_MAX_LENGTH - 1;
```
olsajiri: no need to initialize revlen?

bpf/process/bpf_process_event.h
```diff
@@ -345,17 +346,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
...
-	read_exe((struct task_struct *)get_current_task(), &msg->exe);
+	/* Reset binary info and populate from task's exe path for
+	 * matchBinaries filtering. Without this, matchBinaries selectors
+	 * (especially NotIn) won't work correctly when the process is
+	 * not found in execve_map.
+	 */
+	binary_reset(&curr->bin);
+	curr->bin.path_length = -1;
+	exe_len = read_exe(task, &msg->exe);
+	if (exe_len > 0 && exe_len <= BINARY_PATH_MAX_LEN) {
```
olsajiri: exec_len is unsigned, so no need to check > 0

bpf/process/bpf_process_event.h
```diff
@@ -345,17 +346,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
...
+		err = with_errmetrics(probe_read, curr->bin.path, exe_len, msg->exe.buf);
+		if (err == 0) {
+			curr->bin.path_length = exe_len;
+			if (exe_len > STRING_POSTFIX_MAX_LENGTH - 1)
+				revlen = STRING_POSTFIX_MAX_LENGTH - 1;
+			else
+				revlen = exe_len;
+			asm volatile("%[revlen] &= 0x7f;\n" : [revlen] "+r"(revlen));
+			with_errmetrics(probe_read, curr->bin.end, revlen, msg->exe.end);
+		}
```
olsajiri: there's same code in bpf/process/bpf_execve_event.c:426 doing the same thing (plus argument setup),
i think we should add function for that and have just one copy of that

also I'm not fully clear on why we reset all the binary data just before we read it

olsajiri (commented):
left some other comments, please also add some changelog for the test commit, thanks

bpf/process/bpf_process_event.h
```diff
@@ -345,17 +346,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
...
 	struct execve_map_value *curr;
+	__u32 exe_len;
+	__u32 revlen = STRING_POSTFIX_MAX_LENGTH - 1;
 
 	curr = &msg->curr;
 	curr->key.pid = BPF_CORE_READ(task, tgid);
 	curr->key.ktime = tg_get_ktime();
+	curr->pkey.pid = 0;
+	curr->pkey.ktime = 0;
+	curr->flags = 0;
```
mtardy, line 391: should it be part of a separate patch or unecessary?

bpf/process/bpf_process_event.h
```diff
@@ -345,17 +346,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
...
 
 	get_current_subj_caps(&curr->caps, task);
 	get_namespaces(&curr->ns, task);
 	set_in_init_tree(curr, NULL);
 
-	read_exe((struct task_struct *)get_current_task(), &msg->exe);
+	/* Reset binary info and populate from task's exe path for
+	 * matchBinaries filtering. Without this, matchBinaries selectors
+	 * (especially NotIn) won't work correctly when the process is
+	 * not found in execve_map.
```
mtardy: I'd be a bit more precise than "won't work correctly" if possible?

bpf/process/bpf_process_event.h
mtardy: so for the reset part, bin->path and co are used as a key at some point (in `match_binaries`) so it's important that what is at the end of string is NULL terminated, and the place we write is a heap map that is reused multiple time without be completely zeroed I think. So this part makes sense to me

mtardy: > there's same code in bpf/process/bpf_execve_event.c:426 doing the same thing (plus argument setup),
i think we should add function for that and have just one copy of that

but +1 on that, would be nice, this code is a bit complicated because of the zeroing, the rev parts etc, would be cool to reuse.

pkg/sensors/tracing/kprobe_test.go
mtardy, line 4570: am I correct saying that this test should already be succeed because it's not a process that started before tetragon like the other tests that is testing what you fixed before?

mtardy (commented):
A few comments, I agree with Jiri points. I see the issue but maybe we could reuse existing code to fix it. Could you also use a real email address in the sign-off? Thank you!

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -4428,6 +4431,202 @@ func TestKprobeMatchBinariesEarlyExec(t *testing.T) {
...
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		tailFile.Close()
+		os.Remove(tailFile.Name())
+	})
+
+	catRead, catWrite, err := os.Pipe()
+	require.NoError(t, err)
+	t.Cleanup(func() {
+		catRead.Close()
```
kalakotima: `catRead` is closed twice — once explicitly after `catCommand.Start()`
(correct, to release the local pipe end) and again inside `t.Cleanup`.
The second close will return an error on an already-closed file descriptor.

Suggested fix — only close `catWrite` in Cleanup since `catRead` is
already closed manually:

t.Cleanup(func() {
    catWrite.Close() // catRead already closed after Start()
})

bpf/process/bpf_process_event.h
kalakotima: The initial value of `revlen` is dead code. It is always overwritten
before use inside the `if (err == 0)` block:

  if (exe_len > STRING_POSTFIX_MAX_LENGTH - 1)
      revlen = STRING_POSTFIX_MAX_LENGTH - 1;  // same as initial
  else
      revlen = exe_len;                        // different value

The initialization implies there is a fallback path that uses the
default value — there is not. Declare at point of use instead:

  // remove top-level: __u32 revlen = STRING_POSTFIX_MAX_LENGTH - 1;

  // inside if (err == 0):
  __u32 revlen = (exe_len > STRING_POSTFIX_MAX_LENGTH - 1)
      ? STRING_POSTFIX_MAX_LENGTH - 1
      : exe_len;

bpf/process/bpf_process_event.h
```diff
@@ -345,17 +346,44 @@ event_find_curr_probe(struct msg_generic_kprobe *msg)
...
+	 */
+	binary_reset(&curr->bin);
+	curr->bin.path_length = -1;
```
kalakotima: If `path_length` is an unsigned type (__u32 or __u16), assigning -1
wraps to 0xFFFFFFFF or 0xFFFF. Any downstream check of the form
`path_length < 0` will never be true on an unsigned type, making
this sentinel value silently ineffective.

Please document the type of `path_length` here and confirm the
sentinel comparison used downstream is compatible, or use an explicit
named constant:

  #define BINARY_PATH_LENGTH_UNSET UINT32_MAX
  curr->bin.path_length = BINARY_PATH_LENGTH_UNSET;

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -4428,6 +4431,202 @@ func TestKprobeMatchBinariesEarlyExec(t *testing.T) {
...
+		headCmd := exec.Command("/usr/bin/head", "/etc/passwd")
+
+		err := tailCmd.Start()
+		require.NoError(t, err)
+		tailPID = tailCmd.Process.Pid
+		err = headCmd.Start()
+		require.NoError(t, err)
+		headPID = headCmd.Process.Pid
+
+		err = tailCmd.Wait()
```
kalakotima: Both `/usr/bin/tail /etc/passwd` and `/usr/bin/head /etc/passwd`
exit immediately after reading the file. By the time
`perfring.RunTestEvents` returns and PIDs are validated, the kernel
may have recycled both PIDs and assigned them to unrelated processes.

This makes the test non-deterministic — it could pass or fail based
on what new process happens to reuse tailPID or headPID.

Suggested fix -> use a long-running command to keep the PID alive
through the validation window, similar to how
`TestKprobeMatchBinariesEarlyExecNotIn` uses `tail -f`:

  tailCmd := exec.Command("/usr/bin/tail", "-f", "/dev/null")
  headCmd := exec.Command("/usr/bin/head", "-c", "0", "/dev/zero")

bpf/process/bpf_process_event.h
mann-ds: kept in the fix commit and added a comment explaining this, can split into its own commit/PR if you'd still prefer, though

pkg/sensors/tracing/kprobe_test.go
mann-ds: yes, I added it here similar to `In`/`Prefix`/`Postfix` to avoid regressions in `copy_exe_to_bin`. `matchBinariesNotInPerfringTest` should exercise the normal path, when the process is already in `execve_map`, which was working before the fix in this PR

test for this PR is in `TestKprobeMatchBinariesEarlyExecNotIn`, where the process starts before Tetragon and hits `event_find_curr_probe`

bpf/process/bpf_process_event.h
mann-ds: extracted to `copy_exe_to_bin`

mann-ds:
:wave: I updated the PR to address the comments, please take a look.

pkg/sensors/tracing/kprobe_test.go
mtardy: Could you add those explanations in the commit description?

mtardy (commented):
Thanks, I think it's ready to merge, could you please add all the details of "why" in each commit descriptions? A summary of the PR body would be good in the first commit. And maybe the reply to my question on the tests in the test commit.

bpf/process/bpf_process_event.h
```diff
@@ -340,6 +341,33 @@ set_in_init_tree(struct execve_map_value *curr, struct execve_map_value *parent)
...
+{
+	__u32 len = exe->len;
+	__u32 revlen;
+
+	if (len == 0 || len > BINARY_PATH_MAX_LEN) {
+		bin->path_length = -1;
+		return;
+	}
+
+	asm volatile("%[len] &= 0xff;\n" : [len] "+r"(len));
```
kevsecurity: It would be nice to not have the literal 0xff in this asm. Maybe something like:

```
	asm volatile("%[len] &= %1;\n"
		     : [len] "+r"(len)
		     : "i"(BINARY_PATH_MAX_LEN - 1));
```

?

bpf/process/bpf_process_event.h
```diff
@@ -340,6 +341,33 @@ set_in_init_tree(struct execve_map_value *curr, struct execve_map_value *parent)
...
+	bin->path_length = with_errmetrics(probe_read, bin->path, len, exe->buf);
+	if (bin->path_length == 0)
+		bin->path_length = len;
+
+	revlen = len > STRING_POSTFIX_MAX_LENGTH - 1 ? STRING_POSTFIX_MAX_LENGTH - 1 : len;
+	asm volatile("%[revlen] &= 0x7f;\n" : [revlen] "+r"(revlen));
```
kevsecurity: Ditto here:

```
	asm volatile("%[revlen] &= %1;\n"
		     : [revlen] "+r"(revlen)
		     : "i"(STRING_POSTFIX_MAX_LENGTH - 1));
```

bpf/process/bpf_process_event.h
kevsecurity: I tend to agree with @mtardy. Please put these lines and the comment in another commit. I appreciate it is relevant to your fix, but then the whole PR is relevant, so they could just be in another commit to make reading the commit log easier.

kevsecurity (changes requested):
Looks great and thanks! Please could you address the literals in the asm statements – I appreciate we have widespread examples of it, but let's not add more while we try to clean up! Also, I do think it is worth putting the extra init code into a separate commit for clarity.

bpf/process/bpf_process_event.h
mann-ds: done

bpf/process/bpf_process_event.h
mann-ds: done

bpf/process/bpf_process_event.h
mann-ds: done

pkg/sensors/tracing/kprobe_test.go
mann-ds: done

mann-ds:
> Looks great and thanks! Please could you address the literals in the asm statements – I appreciate we have widespread examples of it, but let's not add more while we try to clean up! Also, I do think it is worth putting the extra init code into a separate commit for clarity.

comments are addressed, I've added a separate commit as well

kevsecurity (approved):
LGTM

mann-ds:
@kevsecurity thank you for approving the PR, can you also merge it please? I don't have permissions, unfortunately

kevsecurity:
> @kevsecurity thank you for approving the PR, can you also merge it please? I don't have permissions, unfortunately

I'd like to give @olsajiri and @mtardy the opportunity to confirm first.

olsajiri (approved):
lgtm, thanks

mtardy (approved):
thanks!

mann-ds:
great news, it's merged now :tada: so the next question is of course if it's possible to backport it to the `v1.6` branch and make a new 1.6.x release?

Issue #4656, opened 2026-02-15
ci: E2E test failure due to failed to load image


sayboras:
### Description

Sample runs in two different PRs

https://github.com/cilium/tetragon/actions/runs/21984718952/job/63516638332?pr=4653
https://github.com/cilium/tetragon/actions/runs/22028698220/job/63649958653?pr=4655

```
I0215 03:13:55.576053   13349 tetragon.go:162] "Loading image into kind cluster" cluster="tetragon-ci-99cd" image="quay.io/cilium/tetragon-ci:49341497a7a34c34b88af3f30dd1c14a435d7013" helm="tetragon.image.override"
E0215 03:13:59.266565   13349 env.go:414] Setup failure: failed to load image quay.io/cilium/tetragon-ci:49341497a7a34c34b88af3f30dd1c14a435d7013 into cluster tetragon-ci-99cd: load image: kind: load docker-image quay.io/cilium/tetragon-ci:49341497a7a34c34b88af3f30dd1c14a435d7013 failed: exit status 1: Image: "quay.io/cilium/tetragon-ci:49341497a7a34c34b88af3f30dd1c14a435d7013" with ID "sha256:d96f38d9bdeb80bb5a26b8ec0f874e86682516c724498b18b9fe962c984b0f58" not yet present on node "tetragon-ci-99cd-control-plane", loading...
ERROR: failed to load image: command "docker exec --privileged -i tetragon-ci-99cd-control-plane ctr --namespace=k8s.io images import --all-platforms --digests --snapshotter=overlayfs -" failed with error: exit status 1
```

PR #4657, opened 2026-02-15
ci/e2e: Fallback to load images from registry

sayboras:
### Description

Don't fail the test if the image is not loaded into kind, as the error could be related other upstream issues (e.g. kind, containerd). Similar approach is already used if the images are not available locally.

Fixes: https://github.com/cilium/tetragon/issues/4656

```release-note
ci/e2e: Fallback to load images from registry
```

### DIFF
--- a/tests/e2e/install/tetragon/tetragon.go
+++ b/tests/e2e/install/tetragon/tetragon.go
@@ -168,7 +168,14 @@ func Install(opts ...Option) env.Func {
 							klog.InfoS("Image is not present locally, attempting to install Tetragon regardless", "cluster", clusterName, "image", v, "helm", k)
 							break
 						}
-						return ctx, fmt.Errorf("failed to load image %s into cluster %s: %w", v, clusterName, err)
+
+						// If failed to load the image, this could be related to kind/containerd issues, so just log
+						// the message and attempt to install Tetragon regardless. See
+						// https://github.com/kubernetes-sigs/kind/issues/3795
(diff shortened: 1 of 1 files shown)


tests/e2e/install/tetragon/tetragon.go
```diff
@@ -168,7 +168,14 @@ func Install(opts ...Option) env.Func {
...
 						}
-						return ctx, fmt.Errorf("failed to load image %s into cluster %s: %w", v, clusterName, err)
+
+						// If failed to load the image, this could be related to kind/containerd issues, so just log
+						// the message and attempt to install Tetragon regardless. See
+						// https://github.com/kubernetes-sigs/kind/issues/3795
+						if err != nil {
+							klog.InfoS("Failed to load image into kind cluster", "cluster", clusterName, "image", v, "error", err)
+							break
+						}
```
kkourt, line 178: So could it happen that we get an error, but the image is actually installed so tetragon installation would work?

sayboras: based on the linked issue, I don't think the image is loaded into the cluster. Some mention the workaround is to load the images using crt command, but it's not ideal in our case

https://github.com/kubernetes-sigs/kind/issues/3795
```
$ docker save -o nginx.tar nginx:1.27.0
$ sudo ctr images import nginx.tar
...
```

kkourt (approved):

PR #4663, opened 2026-02-17
bpf: Fix with_errmetrics_ptr to handle NULL pointers

sayboras:
### Description

The with_errmetrics_ptr macro used IS_ERR() which only catches error
pointers. However, helpers like map_lookup_elem return NULL on failure,
not an error pointer. Since IS_ERR(NULL) is always false, errmetrics
tracking was never triggered for these helpers.

Add IS_ERR_OR_NULL() and use it in with_errmetrics_ptr. When the result
is an error pointer, the error code is extracted via PTR_ERR.
When the result is NULL, ENOENT is reported.

### Changelog

```release-note
bpf: Fix with_errmetrics_ptr to handle NULL pointers
```

### DIFF
--- a/bpf/lib/err.h
+++ b/bpf/lib/err.h
@@ -31,4 +31,9 @@ static inline long PTR_ERR(const void *ptr)
 	return (long)ptr;
 }
 
+static inline bool IS_ERR_OR_NULL(const void *ptr)
+{
+	return unlikely(!ptr) || IS_ERR_VALUE((unsigned long)ptr);
+}
+
 #endif /* BPF_ERR_H__ */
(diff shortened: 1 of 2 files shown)


bpf/process/uprobe_offload.h
```diff
@@ -51,11 +51,7 @@ FUNC_INLINE void do_uprobe_override(void *ctx, __u32 idx)
 	__u64 id = get_current_pid_tgid();
 	__u32 *idxp;
 
-	/*
-	 * This should not happen, it means that the override program was
-	 * not executed for some reason.
-	 */
-	idxp = with_errmetrics_ptr(map_lookup_elem, &sleepable_offload, &id);
+	idxp = map_lookup_elem(&sleepable_offload, &id);
```
olsajiri: ugh, sorry.. how about we steal^wborrow `IS_ERR_OR_NULL` and use it in with_errmetrics_ptr instead?

sayboras: ah it should work as well, let me send one update :100:

PR #4664, opened 2026-02-17
fix(pkg/sensors): force `__base__` sensor to be unloaded last.

FedeDP:
### Description

Since many sensors depend on its maps and progs to be available, always unload it last.

The patch is the easiest possible, but we could also add a way, when creating sensors, to specify a `priority`. 
For now, that seems like a little over-engineered since we don't have any real use case for it. 

### Changelog

```release-note
fix(pkg/sensors): force `__base__` sensor to be unloaded last.
```

### DIFF
--- a/pkg/sensors/base/base.go
+++ b/pkg/sensors/base/base.go
@@ -29,15 +29,13 @@ const (
 )
 
 var (
-	basePolicy = "__base__"
-
 	Execve = program.Builder(
 		config.ExecObj(),
 		"sched/sched_process_exec",
 		"tracepoint/sys_execve",
 		"event_execve",
 		"execve",
-	).SetPolicy(basePolicy)
+	).SetPolicy(sensors.BaseSensorName)
 
 	ExecveMapUpdate = program.Builder(
 		config.ExecUpdateObj(),
@@ -45,7 +43,7 @@ var (
 		"seccomp",
 		"execve_map_update",
 		"seccomp",
-	).SetPolicy(basePolicy)
+	).SetPolicy(sensors.BaseSensorName)
 
 	ExecveBprmCommit = program.Builder(
 		"bpf_execve_bprm_commit_creds.o",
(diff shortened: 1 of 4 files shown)


pkg/sensors/handler.go
```diff
@@ -291,8 +291,16 @@ func removeAllSensors(h *handler, unpin bool) {
 	defer h.collections.mu.Unlock()
 	collections := h.collections.c
 	for ck, col := range collections {
-		col.destroy(unpin)
-		delete(collections, ck)
+		if col.name == "__base__" {
+			// Base sensor always unloaded last
+			defer func(ck collectionKey, col *collection) {
```
FedeDP, line 298: `defer` in a for loop is ok in this case since we are only deferring once.

pkg/sensors/handler.go
```diff
@@ -291,8 +291,16 @@ func removeAllSensors(h *handler, unpin bool) {
 	defer h.collections.mu.Unlock()
 	collections := h.collections.c
 	for ck, col := range collections {
-		col.destroy(unpin)
-		delete(collections, ck)
+		if col.name == "__base__" {
```
FedeDP: Perhaps we might move `__base__` string variable to be held by the `pkg/sensors` package, instead of the `base` one, so that we can use the variable here.

kevsecurity: Yes I think we should use a const here.

FedeDP: Done!

kevsecurity (approved):

pkg/sensors/handler.go
olsajiri: will it still happen under the `h.collections.mu` lock? edit: nah, probably yes ;-)

FedeDP: Yes since 
"Deferred function calls are pushed onto a stack. When a function returns, its deferred calls are executed in last-in-first-out order."
(https://go.dev/tour/flowcontrol/13)

pkg/sensors/base/base.go
```diff
@@ -29,47 +29,45 @@ const (
 )
 
 var (
-	basePolicy = "__base__"
```
olsajiri, line 32: is there some reason for moving this out? seems to me that base package is better place for that

FedeDP: I can't reference `base` package from within the `sensors` package because that would be a dep cycle.
But i wanted to address https://github.com/cilium/tetragon/pull/4664#discussion_r2816400088

PR #4670, opened 2026-02-19
fix(eventcache): resolve missing container metadata in kprobe events

### LINKED ISSUE #4152: container metadata is missing from events if the policy is matched during kubectl exec

1. In a kubernetes environment (verified on kind + Ubuntu 24.04 and EKS), deploy tetragon.
2. Deploy the tracing policy below, which blocks all executables matching the k8s pod label `app: ubuntu`

```
kind: TracingPolicyNamespaced
metadata:
  name: deploy-ubuntu-deployment
spec:
  kprobes:
  - args:
    - index: 0
      type: linux_binprm
    call: security_bprm_creds_for_exec
    selectors:
    - matchActions:
      - action: Override
        argError: -1
    syscall: false
  options:
  - name: disable-kprobe-multi
    value: "1"
  podSelector:
    matchLabels:
      app: ubuntu
```

3. Run a program in a pod that matches `app: ubuntu` using `kubectl exec`.
4. The executable is blocked correctly, but its event doesn't have any information associated with the pod, as below:

[kubectl-exec.json](https://github.com/user-attachments/files/22754047/kubectl-exec.json)

5. If I remove the override part from the tracing policy, the program can run without problem, as expected.  However, its event still doesn't come with its container metadata.
6. If I run the program not through `kubectl exec` directly, for example, from a shell that I've established before, the event comes with container metadata correctly. 

[normal.json](https://github.com/user-attachments/files/22754065/normal.json)

### Tetragon Version

```
/ # tetra version
CLI version: v1.5.0
```

### Kernel Version

/ # tetra version
CLI version: v1.5.0

(EKS)
```
/ # uname -a
Linux ip-xxxxxx.compute.internal 6.1.150-174.273.amzn2023.x86_64 #1 SMP PREEMPT_DYNAMIC Tue Sep  9 12:21:26 UTC 2025 x86_64 Linux
```

(kind)
```
/ # uname -a
Linux kind-control-plane 6.14.0-32-generic #32~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Sep  2 14:21:04 UTC 2 x86_64 Linux
```

### Kubernetes Version

(EKS)
```
$ kubectl version
Client Version: v1.33.3
Kustomize Version: v5.6.0
Server Version: v1.32.9-eks-113cf36
```

(kind)
```
$ kubectl version
Client Version: v1.33.3
Kustomize Version: v5.6.0
Server Version: v1.33.1
```

### Bugtool

This command runs on EKS:

```
/ # tetra bugtool
level=info msg="saving init info"
level=info msg="retrieving lib directory" libDir=/var/lib/tetragon
level=warn msg="not an object file, ignoring" path=/var/lib/tetragon
level=warn msg="no btf filename in tetragon config, attempting to fall back to /sys/kernel/btf/vmlinux"
level=info msg="btf file added" btfFname=/sys/kernel/btf/vmlinux
level=info msg="tetragon log file added" exportFname=/var/run/cilium/tetragon/tetragon.log
level=info msg="contacting metrics server" metricsAddr=http://localhost:2112/metrics
level=info msg="executed command" cmd=/bin/dmesg ret=0 dstFname=dmesg.out
level=info msg="executed command" cmd="/sbin/tc filter show dev lo ingress" ret=0 dstFname=tc-info.lo.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev lo egress" ret=0 dstFname=tc-info.lo.egress
level=info msg="executed command" cmd="/sbin/tc filter show dev ens5 ingress" ret=0 dstFname=tc-info.ens5.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev ens5 egress" ret=0 dstFname=tc-info.ens5.egress
level=info msg="executed command" cmd="/sbin/tc filter show dev enif20915368b5 ingress" ret=0 dstFname=tc-info.enif20915368b5.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev enif20915368b5 egress" ret=0 dstFname=tc-info.enif20915368b5.egress
level=info msg="executed command" cmd="/sbin/tc filter show dev enieddc18b2fc6 ingress" ret=0 dstFname=tc-info.enieddc18b2fc6.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev enieddc18b2fc6 egress" ret=0 dstFname=tc-info.enieddc18b2fc6.egress
level=info msg="executed command" cmd="/sbin/tc filter show dev ens6 ingress" ret=0 dstFname=tc-info.ens6.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev ens6 egress" ret=0 dstFname=tc-info.ens6.egress
level=info msg="executed command" cmd="/sbin/tc filter show dev eni43ddcaf029e ingress" ret=0 dstFname=tc-info.eni43ddcaf029e.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev eni43ddcaf029e egress" ret=0 dstFname=tc-info.eni43ddcaf029e.egress
level=info msg="executed command" cmd="/sbin/tc filter show dev enifaffa0817c2 ingress" ret=0 dstFname=tc-info.enifaffa0817c2.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev enifaffa0817c2 egress" ret=0 dstFname=tc-info.enifaffa0817c2.egress
level=info msg="executed command" cmd="/sbin/tc filter show dev eni3fbdac6a2ec ingress" ret=0 dstFname=tc-info.eni3fbdac6a2ec.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev eni3fbdac6a2ec egress" ret=0 dstFname=tc-info.eni3fbdac6a2ec.egress
level=info msg="executed command" cmd="/sbin/tc filter show dev eni85d6be82070 ingress" ret=0 dstFname=tc-info.eni85d6be82070.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev eni85d6be82070 egress" ret=0 dstFname=tc-info.eni85d6be82070.egress
level=info msg="executed command" cmd="/sbin/tc filter show dev enid197514c16a ingress" ret=0 dstFname=tc-info.enid197514c16a.ingress
level=info msg="executed command" cmd="/sbin/tc filter show dev enid197514c16a egress" ret=0 dstFname=tc-info.enid197514c16a.egress
level=info msg="executed command" cmd="/usr/bin/bpftool map show -j" ret=0 dstFname=bpftool-maps.json
level=info msg="executed command" cmd="/usr/bin/bpftool prog show -j" ret=0 dstFname=bpftool-progs.json
level=info msg="executed command" cmd="/usr/bin/bpftool cgroup tree -j" ret=0 dstFname=bpftool-cgroups.json
level=info msg="Dumping gops information" gops-address=localhost:8118 gops-path=/usr/bin/gops
level=info msg="executed command" cmd="/usr/bin/gops stack localhost:8118" ret=0 dstFname=gops.stack
level=info msg="executed command" cmd="/usr/bin/gops stats localhost:8118" ret=0 dstFname=gops.stats
level=info msg="executed command" cmd="/usr/bin/gops memstats localhost:8118" ret=0 dstFname=gops.memstats
level=info msg="Contacting gops server for pprof dump" gops-address=localhost:8118
level=info msg="Successfully dumped gops pprof" gops-address=localhost:8118 gops-path=/usr/bin/gops profile=cpu
level=info msg="Contacting gops server for pprof dump" gops-address=localhost:8118
level=info msg="Successfully dumped gops pprof" gops-address=localhost:8118 gops-path=/usr/bin/gops profile=heap
level=info msg="dumped tracing policies in tracing-policies.json"
level=info msg="executed command" cmd="/usr/bin/pmap -x 1" ret=0 dstFname=pmap.out
level=info msg="cgroup file added" file=memory.current
level=info msg="cgroup file added" file=memory.stat
level=info msg="BPF maps checks added" file=debugmaps.json
level=info msg="executed command" cmd="/bin/cat /sys/kernel/tracing/trace" ret="exit status 1" dstFname=trace
```

### Relevant log output

```shell

```

### Anything else?

_No response_

puwun:
Fixes: #4152

### Description

When a kprobe fires during execve (e.g. `security_bprm_creds_for_exec` triggered by `kubectl exec`), the process entry in the cache comes from the clone event and has no Docker ID or Pod info. The exec event creates a separate cache entry under a different ExecId (different ktime), so the kprobe handler only finds the stale clone-based entry. These events bypassed the event cache entirely because `Needed()` only cached events with a non-empty Docker ID, so they were emitted immediately with missing container metadata.

### Changes
- Add `cgrpid` field to BPF `msg_generic_kprobe` and set it via  `tg_get_current_cgroup_id()` in `generic_start_process_filter`
- Add matching `Cgrpid` field to Go `MsgGenericKprobe` and  `MsgGenericTracepoint` structs
- Force-cache kprobe events where Docker and Pod are both empty  and cgrpid is non-zero
- Use `cgidmap.Get(cgrpid)` in `HandleGenericEvent` on retry to  resolve container ID (with CgTrackerID support), then resolve  pod info from the container ID
- Remove `GetByPidWithContainerInfo` PID-based lookup

Requires `--enable-cgidmap` and `--enable-cri`

### Changelog

```release-note
Fixed missing container metadata (pod name, namespace, container ID) in kprobe events when a tracing policy matches during `kubectl exec`
```

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -43,7 +43,8 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 	config = map_lookup_elem(&config_map, &msg->idx);
 	if (!config)
 		return 0;
-	if (!policy_filter_check(config->policy_id))
+	msg->cgrp_tracker_id = tg_get_current_cgrp_tracker_id();
+	if (!do_policy_filter_check(config->policy_id, msg->cgrp_tracker_id))
 		return 0;
 	msg->func_id = config->func_id;
 	msg->retprobe_id = 0;
--- a/bpf/process/policy_filter.h
+++ b/bpf/process/policy_filter.h
@@ -51,30 +51,40 @@ struct {
 		});
 } policy_filter_cgroup_maps SEC(".maps");
 
-// policy_filter_check checks whether the policy applies on the current process.
-// Returns true if it does, false otherwise.
-
-FUNC_INLINE bool policy_filter_check(u32 policy_id)
+/**
+ * tg_get_current_cgrp_tracker_id() Returns tracker cgroup id of the current task or 0 on failure.
+ * If the cgroup tracker is not enabled, it returns the cgroup id of the current task.
+ */
+FUNC_INLINE __u64 tg_get_current_cgrp_tracker_id(void)
 {
-	void *policy_map;
 	__u64 cgroupid, trackerid;
 
-	if (!policy_id)
-		return true;
-
-	policy_map = map_lookup_elem(&policy_filter_maps, &policy_id);
-	if (!policy_map)
-		return false;
-
 	cgroupid = tg_get_current_cgroup_id();
-	if (!cgroupid) {
-		errmetrics(ENOENT);
-		return false;
-	}
+	if (!cgroupid)
+		return 0;
 
 	trackerid = cgrp_get_tracker_id(cgroupid);
 	if (trackerid)
 		cgroupid = trackerid;
+
+	return cgroupid;
+}
+
+// do_policy_filter_check checks whether the policy applies on a given cgroup id.
+// Returns true if it does, false otherwise.
+FUNC_INLINE bool do_policy_filter_check(__u32 policy_id, __u64 cgroupid)
+{
+	void *policy_map;
+
+	if (!policy_id)
+		return true;
+
+	if (!cgroupid)
+		return false;
+
+	policy_map = map_lookup_elem(&policy_filter_maps, &policy_id);
+	if (!policy_map)
+		return false;
 
 	if (map_lookup_elem(policy_map, &cgroupid))
 		return true; // We have a match from the podSelector and/or the containerSelector.
@@ -82,17 +92,33 @@ FUNC_INLINE bool policy_filter_check(u32 policy_id)
 	// We didn't match on the podSelector and/or the containerSelector.
 	// Now we need to check if we have a hostSelector match.
 
-	trackerid = HOST_SELECTOR_MODE;
-	if (!map_lookup_elem(policy_map, &trackerid))
-		return false; // Cannot find the match mode of the hostSelector so we do not care to match any host workloads.
+	{
+		__u64 trackerid = HOST_SELECTOR_MODE;
 
-	policy_id = ALL_PODS_POLICY_ID;
-	policy_map = map_lookup_elem(&policy_filter_maps, &policy_id);
-	if (!policy_map)
-		return false; // Cannot find the cgroupids of all containers inside all pods. This should not happen.
+		if (!map_lookup_elem(policy_map, &trackerid))
+			return false; // Cannot find the match mode of the hostSelector so we do not care to match any host workloads.
+	}
+
+	{
+		__u32 all_pods_id = ALL_PODS_POLICY_ID;
+
+		policy_map = map_lookup_elem(&policy_filter_maps, &all_pods_id);
+		if (!policy_map)
+			return false; // Cannot find the cgroupids of all containers inside all pods. This should not happen.
+	}
 
 	// If !map_lookup_elem(policy_map, &cgroupid) then our cgroupid belongs to a host workload.
 	return !map_lookup_elem(policy_map, &cgroupid);
+}
+
+// policy_filter_check checks whether the policy applies on the current process.
+// Returns true if it does, false otherwise.
+FUNC_INLINE bool policy_filter_check(__u32 policy_id)
+{
+	__u64 cgroupid;
+
+	cgroupid = tg_get_current_cgrp_tracker_id();
+	return do_policy_filter_check(policy_id, cgroupid);
 }
 
 struct {
--- a/pkg/eventcache/eventcache.go
+++ b/pkg/eventcache/eventcache.go
@@ -113,6 +113,16 @@ func HandleGenericInternal(ev notify.Event, pid uint32, tid *uint32, timestamp u
 	return nil, err
 }
 
+// CgrpTrackerGetter is implemented by events that carry a cgroup tracker ID
+// from BPF for container metadata resolution.
+type CgrpTrackerGetter interface {
+	GetCgrpTrackerID() uint64
+}
+
+// ContainerIDResolver resolves a cgroup tracker ID to a container ID.
+// Set by the tracing package at init time to avoid circular imports.
+var ContainerIDResolver func(uint64) string
+
 // Generic Event handler without any extra msg specific details or debugging
 // so we only need to wait for the internal link to the process context to
 // resolve PodInfo. This happens when the msg populates the internal state
@@ -120,8 +130,28 @@ func HandleGenericInternal(ev notify.Event, pid uint32, tid *uint32, timestamp u
 func HandleGenericEvent(internal *process.ProcessInternal, ev notify.Event, tid *uint32) error {
 	p := internal.UnsafeGetProcess()
 	if option.Config.EnableK8s && p.Pod == nil {
-		CacheRetries(PodInfo).Inc()
-		return ErrFailedToGetPodInfo
+		docker := p.Docker
+		if docker == "" {
+			if g, ok := ev.(CgrpTrackerGetter); ok && ContainerIDResolver != nil {
+				if id := g.GetCgrpTrackerID(); id != 0 {
+					docker = ContainerIDResolver(id)
+				}
+			}
+		}
+		if docker != "" {
+			podInfo := process.GetPodInfo(docker, p.Binary, p.Arguments, 0)
+			if podInfo == nil {
+				CacheRetries(PodInfo).Inc()
+				return ErrFailedToGetPodInfo
+			}
+			internal.AddPodInfo(podInfo)
+			if p.Docker == "" {
+				internal.UnsafeGetProcess().Docker = docker
+			}
+		} else {
+			CacheRetries(PodInfo).Inc()
+			return ErrFailedToGetPodInfo
+		}
 	}
 
 	// When we report the per thread fields, take a copy
--- a/pkg/grpc/tracing/tracing.go
+++ b/pkg/grpc/tracing/tracing.go
@@ -399,7 +399,10 @@ func GetProcessKprobe(event *MsgGenericKprobeUnix) *tetragon.ProcessKprobe {
 	if ec := eventcache.Get(); ec != nil && !isUnknown(tetragonProcess) &&
 		(ec.Needed(tetragonProcess) ||
 			(tetragonProcess.Pid.Value > 1 && ec.Needed(tetragonParent)) ||
-			(option.Config.EnableProcessKprobeAncestors && ec.NeededAncestors(parent, ancestors))) {
+			(option.Config.EnableProcessKprobeAncestors && ec.NeededAncestors(parent, ancestors)) ||
+			// Clone-to-exec race: cache events with no container
+			// metadata so retry can resolve via cgidmap.
+			(option.Config.EnableK8s && tetragonProcess.Docker == "" && tetragonProcess.Pod == nil && event.Msg.CgrpTrackerID != 0)) {
 		ec.Add(nil, tetragonEvent, event.Msg.Common.Ktime, event.Msg.ProcessKey.Ktime, event)
 		return nil
 	}
@@ -511,6 +514,10 @@ func (msg *MsgGenericTracepointUnix) RetryInternal(ev notify.Event, timestamp ui
 
 func (msg *MsgGenericTracepointUnix) Retry(internal *process.ProcessInternal, ev notify.Event) error {
 	return eventcache.HandleGenericEvent(internal, ev, &msg.Msg.Tid)
+}
+
+func (msg *MsgGenericTracepointUnix) GetCgrpTrackerID() uint64 {
+	return msg.Msg.CgrpTrackerID
 }
 
 func familyString(family uint16) string {
@@ -774,6 +781,10 @@ func (msg *MsgGenericKprobeUnix) Retry(internal *process.ProcessInternal, ev not
 	return eventcache.HandleGenericEvent(internal, ev, &msg.Msg.Tid)
 }
 
+func (msg *MsgGenericKprobeUnix) GetCgrpTrackerID() uint64 {
+	return msg.Msg.CgrpTrackerID
+}
+
 func (msg *MsgGenericKprobeUnix) HandleMessage() *tetragon.GetEventsResponse {
 	k := GetProcessKprobe(msg)
 	if k == nil {
@@ -909,6 +920,10 @@ func (msg *MsgGenericUprobeUnix) Retry(internal *process.ProcessInternal, ev not
 	return eventcache.HandleGenericEvent(internal, ev, &msg.Msg.Tid)
 }
 
+func (msg *MsgGenericUprobeUnix) GetCgrpTrackerID() uint64 {
+	return msg.Msg.CgrpTrackerID
+}
+
 func GetProcessUprobe(event *MsgGenericUprobeUnix) *tetragon.ProcessUprobe {
 	var ancestors []*process.ProcessInternal
 	var tetragonAncestors []*tetragon.Process
@@ -1020,6 +1035,10 @@ func (msg *MsgGenericUsdtUnix) Retry(internal *process.ProcessInternal, ev notif
 	return eventcache.HandleGenericEvent(internal, ev, &msg.Msg.Tid)
 }
 
+func (msg *MsgGenericUsdtUnix) GetCgrpTrackerID() uint64 {
+	return msg.Msg.CgrpTrackerID
+}
+
 func GetProcessUsdt(event *MsgGenericUsdtUnix) *tetragon.ProcessUsdt {
 	var ancestors []*process.ProcessInternal
 	var tetragonAncestors []*tetragon.Process
@@ -1120,6 +1139,10 @@ func (msg *MsgGenericLsmUnix) RetryInternal(ev notify.Event, timestamp uint64) (
 
 func (msg *MsgGenericLsmUnix) Retry(internal *process.ProcessInternal, ev notify.Event) error {
 	return eventcache.HandleGenericEvent(internal, ev, &msg.Msg.Tid)
+}
+
+func (msg *MsgGenericLsmUnix) GetCgrpTrackerID() uint64 {
+	return msg.Msg.CgrpTrackerID
 }
 
 func (msg *MsgGenericLsmUnix) HandleMessage() *tetragon.GetEventsResponse {
--- a/bpf/lib/generic.h
+++ b/bpf/lib/generic.h
@@ -78,6 +78,7 @@ struct msg_generic_kprobe {
 	__u32 tid; // Thread ID that triggered the event
 	__u64 kernel_stack_id; // Kernel stack trace ID on u32 and potential error, see flag in msg_common.flags
 	__u64 user_stack_id; // User Stack trace ID
+	__u64 cgrp_tracker_id; // Cgroup tracker ID of the process that triggered the event
 	/* anything above is shared with the userspace so it should match structs MsgGenericKprobe and MsgGenericTracepoint in Go */
 	char args[24000];
 	unsigned long a0, a1, a2, a3, a4;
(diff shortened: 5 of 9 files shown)


kkourt:
Thanks! Can you please provide a sequence of events that cause is?
If I understand correctly, this is caused by the `security_bprm_creds_for_exec` being executed _after_ the clone event but _before_ the execve is generated?

puwun:
@kkourt correct, the sequence is something like:

1. `clone()` → creates `ProcessInternal` with empty Docker/Pod
2. `security_bprm_creds_for_exec` kprobe fires → looks up process, finds the clone entry → `Needed()` returns false (Docker is empty) → event emitted immediately without container metadata
3. `execve` event arrives → creates a new `ProcessInternal` (different ExecId) with proper Docker/Pod

The fix force-caches the event at step 2 so the retry path can find the entry from step 3 via a PID-based lookup.

kkourt:
At which step between the three above steps is the "Docker" value set?
In other words, when we are executing `security_bprm_creds_for_exec`, are we in the host or in the container cgroup?

puwun:
@kkourt Docker is set at step 3 when the BPF exec handler reads the cgroup, `initProcessInternalExec` sets it from `event.Kube.Docker`. The clone entry just copies the parent (`cloneInternalProcessCopy`), which doesn't have a container ID.
The process should already be in the container cgroup when `security_bprm_creds_for_exec` fires, but the kprobe handler only sends the `ProcessKey` so userspace ends up looking up the stale clone entry.

kkourt:
> The process should already be in the container cgroup when security_bprm_creds_for_exec fires (...)

So then, this problem does not exist when using cgidmap?
(see https://github.com/cilium/tetragon/pull/2776 for some details)

pkg/eventcache/eventcache.go
```diff
@@ -120,8 +120,38 @@ func HandleGenericInternal(ev notify.Event, pid uint32, tid *uint32, timestamp u
...
-		return ErrFailedToGetPodInfo
+		if p.Docker == "" && p.Pid != nil {
+			// Stale clone entry: look up by PID to find the
+			// exec-based entry with container metadata.
+			updated := process.GetByPidWithContainerInfo(p.Pid.Value)
+			if updated == nil {
+				CacheRetries(PodInfo).Inc()
+				return ErrFailedToGetPodInfo
+			}
+			up := updated.UnsafeGetProcess()
```
kalakotima: UnsafeGetProcess() can return nil — the "Unsafe" prefix exists
precisely because it makes no guarantees. Accessing up.Pod,
up.Docker, up.Binary, up.Arguments without a nil check will
panic at runtime.

Suggested fix:

up := updated.UnsafeGetProcess()
if up == nil {
CacheRetries(PodInfo).Inc()
return ErrFailedToGetPodInfo
}

pkg/process/process.go
```diff
@@ -587,6 +587,18 @@ func Get(execId string) (*ProcessInternal, error) {
 	return procCache.get(execId)
 }
 
+// GetByPidWithContainerInfo scans the process cache for a ProcessInternal
+// matching the given PID that has a non-empty Docker ID.
+func GetByPidWithContainerInfo(pid uint32) *ProcessInternal {
```
kalakotima: This performs a full O(n) linear scan of the process cache on every
call. Given the cache can hold up to `processCacheSize` entries
(default 32768), and this is called from the event processing hot
path in `HandleGenericEvent`, this will cause measurable latency
under high event load.

The cache is keyed by `ExecId` not PID. A PID-based lookup requires
either a secondary index or accepting the linear scan cost with a
clear comment explaining why it is acceptable here.

At minimum, please add a comment documenting the O(n) complexity
and why it is acceptable, or consider a secondary pid→execId index.

kalakotima: `GetByPidWithContainerInfo` returns entries in any GC state including
`deletePending`, `deleteReady`, and `deleted`. A process being
returned here could be concurrently removed by the GC goroutine
between this scan and the caller using the result.

The eviction callback in `NewCache` already demonstrates the correct
pattern:

if evicted.color != inUse {
    return
}

The same guard should apply here:

if v.process != nil && v.process.Pid != nil &&
    v.process.Pid.Value == pid && v.process.Docker != "" &&
    v.color == inUse {    // ← add this
    return v
}

pkg/grpc/tracing/tracing.go
```diff
@@ -388,7 +388,10 @@ func GetProcessKprobe(event *MsgGenericKprobeUnix) *tetragon.ProcessKprobe {
 	if ec := eventcache.Get(); ec != nil && !isUnknown(tetragonProcess) &&
 		(ec.Needed(tetragonProcess) ||
 			(tetragonProcess.Pid.Value > 1 && ec.Needed(tetragonParent)) ||
-			(option.Config.EnableProcessKprobeAncestors && ec.NeededAncestors(parent, ancestors))) {
+			(option.Config.EnableProcessKprobeAncestors && ec.NeededAncestors(parent, ancestors)) ||
```
kalakotima, line 402: This condition caches for retry any event where `Docker == ""` and
`Pod == nil`. This includes host-level processes (PID 1, kernel
threads, system daemons) that will never have container metadata.

These will be retried until `MaxRetries` is exhausted, generating
unnecessary `ErrFailedToGetPodInfo` churn and inflating
`CacheRetries(PodInfo)` metrics with false positives.

Consider adding a guard to limit this to processes that are
reasonably expected to be containerized — for example checking
that the process is not a host-network process, or using an
allowlist of known non-container conditions to skip caching.

puwun:
@kkourt I don't think so, the issue isn't how Docker gets resolved from the cgroup, it's that the kprobe handler looks up the `ProcessInternal` by `pid+clone_ktime` and finds the clone entry. 
`MsgCloneEvent` doesn't carry any cgroup info (`Kube/Cgrpid`), and the `MsgGenericKprobe` doesn't either. So cgidmap wouldn't have anything to resolve from at that point, the container ID resolution (whether via cgroup name or cgidmap) only happens in `msgToExecveKubeUnix` for exec events.

kkourt:
> @kkourt I don't think so, the issue isn't how Docker gets resolved from the cgroup, it's that the kprobe handler looks up the `ProcessInternal` by `pid+clone_ktime` and finds the clone entry. `MsgCloneEvent` doesn't carry any cgroup info (`Kube/Cgrpid`), and the `MsgGenericKprobe` doesn't either. So cgidmap wouldn't have anything to resolve from at that point, the container ID resolution (whether via cgroup name or cgidmap) only happens in `msgToExecveKubeUnix` for exec events.

@puwun I see, thanks.

The problem with the PID lookup is that it's not reliable, because the pid might be recycled by the time we check.

I think the best approach to tackle this is to:
* have bpf pass the cgroup id on the kprobe (and all events)
    * we already retrieve it in the policyfilter check (this is why the podSelector works)
 * If we have the cgroup id, we can figure out the container ID using the cgdimap, and, from that, find the pod.
 
 It's a bit different from what we've been doing so far, i.e., associating pod information using only information from exec events, but, at least for the cases where no pod information exists, I don't see a downside to do this.

kkourt (commented):
(see https://github.com/cilium/tetragon/pull/4670#issuecomment-4301993085)

puwun:
thanks @kkourt. Updated the changes, now passing `cgrpid` from BPF via `tg_get_current_cgroup_id()` and using `cgidmap` (with CgTrackerID support) in `HandleGenericEvent` to resolve container metadata on retry. Removed `GetByPidWithContainerInfo`
Verified e2e on kind with `--enable-cgidmap` and `--enable-cri`.

bpf/process/generic_calls.h
```diff
@@ -63,6 +57,7 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 	get_namespaces(&msg->ns, task);
 	/* Initialize capabilities to apply filters on them */
 	get_current_subj_caps(&msg->caps, task);
+	msg->cgrpid = tg_get_current_cgroup_id();
 #ifdef __NS_CHANGES_FILTER
 	msg->sel.match_ns = 0;
 #endif
```
kkourt, line 69: We already do a lookup for the cgroup id and the tracker id in policy_filter_check().
How about we just reuse that?

Maybe something like:

First commit: introduce tg_get_current_cgrp_tracker_id() and refactor policy_filter_check() to compute cgroupid first

```diff
diff --git a/bpf/lib/bpf_cgroup.h b/bpf/lib/bpf_cgroup.h
index 2a5d6c898..3f757f14f 100644
--- a/bpf/lib/bpf_cgroup.h
+++ b/bpf/lib/bpf_cgroup.h
@@ -387,6 +387,25 @@ FUNC_INLINE __u64 tg_get_current_cgroup_id(void)
 	return get_cgroup_id(cgrp);
 }
 
+/**
+ * tg_get_current_cgrp_tracker_id() Returns tracker cgroup id of the current task or 0 on failure
+ */
+FUNC_INLINE __u64 tg_get_current_cgrp_tracker_id(void)
+{
+	__u64 cgroupid, trackerid;
+
+	cgroupid = tg_get_current_cgroup_id();
+	if (!cgroupid)
+		return 0;
+
+	trackerid = cgrp_get_tracker_id(cgroupid);
+	if (trackerid)
+		cgroupid = trackerid;
+
+	return cgroupid;
+}
+
+
 /**
  * __get_cgrp_tracking_val_heap() Get a cgroup_tracking_val from the
  * tg_cgrps_tracking_heap map while setting its fields.
diff --git a/bpf/process/policy_filter.h b/bpf/process/policy_filter.h
index 95cba33b2..cfe3385c7 100644
--- a/bpf/process/policy_filter.h
+++ b/bpf/process/policy_filter.h
@@ -56,8 +56,12 @@ struct {
 
 FUNC_INLINE bool policy_filter_check(u32 policy_id)
 {
+	u64 cgroupid
 	void *policy_map;
-	__u64 cgroupid, trackerid;
+
+	cgroupid = tg_get_current_cgrp_tracker_id();
+	if (!cgroupid)
+		return false
 
 	if (!policy_id)
 		return true;
@@ -66,14 +70,6 @@ FUNC_INLINE bool policy_filter_check(u32 policy_id)
 	if (!policy_map)
 		return false;
 
-	cgroupid = tg_get_current_cgroup_id();
-	if (!cgroupid)
-		return false;
-
-	trackerid = cgrp_get_tracker_id(cgroupid);
-	if (trackerid)
-		cgroupid = trackerid;
-
 	if (map_lookup_elem(policy_map, &cgroupid))
 		return true; // We have a match from the podSelector and/or the containerSelector.
 ```

Second commit: set msg->cgrp_tracker_id
```diff
diff --git a/bpf/process/generic_calls.h b/bpf/process/generic_calls.h
index 932f170fc..d2d3c1637 100644
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -43,7 +43,8 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 	config = map_lookup_elem(&config_map, &msg->idx);
 	if (!config)
 		return 0;
-	if (!policy_filter_check(config->policy_id))
+	msg->cgrp_tracker_id = tg_get_current_cgrp_tracker_id()
+	if (!policy_filter_check__(config->policy_id, msg->cgrp_tracker_id))
 		return 0;
 	msg->func_id = config->func_id;
 	msg->retprobe_id = 0;
diff --git a/bpf/process/policy_filter.h b/bpf/process/policy_filter.h
index cfe3385c7..bcd9a0d39 100644
--- a/bpf/process/policy_filter.h
+++ b/bpf/process/policy_filter.h
@@ -51,15 +51,10 @@ struct {
 		});
 } policy_filter_cgroup_maps SEC(".maps");
 
-// policy_filter_check checks whether the policy applies on the current process.
+// policy_filter_check__ checks whether the policy on a given cgroup id
 // Returns true if it does, false otherwise.
-
-FUNC_INLINE bool policy_filter_check(u32 policy_id)
+FUNC_INLINE bool policy_filter_check__(u32 policy_id, u64 cgroupid)
 {
-	u64 cgroupid
-	void *policy_map;
-
-	cgroupid = tg_get_current_cgrp_tracker_id();
 	if (!cgroupid)
 		return false
 
@@ -89,4 +84,15 @@ FUNC_INLINE bool policy_filter_check(u32 policy_id)
 	return !map_lookup_elem(policy_map, &cgroupid);
 }
 
+// policy_filter_check checks whether the policy applies on the current process.
+// Returns true if it does, false otherwise.
+FUNC_INLINE bool policy_filter_check(u32 policy_id)
+{
+	u64 cgroupid
+	void *policy_map;
+
+	cgroupid = tg_get_current_cgrp_tracker_id();
+	return policy_filter_check__(policy_id, cgroupid);
+}
+
 #endif /* POLICY_FILTER_MAPS_H__ */
```

kkourt (changes requested):
Thanks!

Can you please split the user-space changes into a different commit? Take a look at https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/.

Also, please have a look at the CI static failure.

puwun:
Thanks @kkourt. Updated and split into two commits. 
Placed `tg_get_current_cgrp_tracker_id()` in `policy_filter.h` instead of `bpf_cgroup.h` since `bpf_cgroup.h` doesn't include `cgtracker.h`.

kkourt:
> Thanks @kkourt. Updated and split into two commits. Placed `tg_get_current_cgrp_tracker_id()` in `policy_filter.h` instead of `bpf_cgroup.h` since `bpf_cgroup.h` doesn't include `cgtracker.h`.

Thanks! Could you also rebase to resolve the conflict?

kkourt:
Once done, please force-push the new branch and notify the reviewers by clicking on the github button "Re-request review".

bpf/process/policy_filter.h
```diff
@@ -51,46 +51,73 @@ struct {
...
+	if (trackerid)
+		cgroupid = trackerid;
 
-FUNC_INLINE bool policy_filter_check(u32 policy_id)
+	return cgroupid;
+}
+
+// policy_filter_check__ checks whether the policy applies on a given cgroup id.
+// Returns true if it does, false otherwise.
+FUNC_INLINE bool policy_filter_check__(__u32 policy_id, __u64 cgroupid)
```
kkourt: Can you split the BPF patch into two commits:
1. One that introduces the policy_filter_check__ function
2. A subsequent commit that uses it?

olsajiri: what's the idea behind the '__' function suffix? we have one function like that.. I'd vote against it ;-)

kkourt: @olsajiri would `do_policy_filter_check` work better? If not, any suggestions?

bpf/process/policy_filter.h
```diff
@@ -51,46 +51,73 @@ struct {
...
 		return true; // We have a match from the podSelector and/or the containerSelector.
 
 	// We didn't match on the podSelector and/or the containerSelector.
 	// Now we need to check if we have a hostSelector match.
 
-	trackerid = HOST_SELECTOR_MODE;
-	if (!map_lookup_elem(policy_map, &trackerid))
-		return false; // Cannot find the match mode of the hostSelector so we do not care to match any host workloads.
+	{
+		__u64 trackerid = HOST_SELECTOR_MODE;
```
kkourt, line 96: @tpapagian could please you take a look at the changes here?

bpf/process/policy_filter.h
```diff
@@ -51,46 +51,73 @@ struct {
 		});
 } policy_filter_cgroup_maps SEC(".maps");
 
-// policy_filter_check checks whether the policy applies on the current process.
-// Returns true if it does, false otherwise.
+/**
+ * tg_get_current_cgrp_tracker_id() Returns tracker cgroup id of the current task or 0 on failure
```
kkourt:
```suggestion
 * tg_get_current_cgrp_tracker_id() Returns tracker cgroup id of the current task or 0 on failure
 * If the cgroup tracker is not enabled, it returns the cgroup id of the current task.
```

pkg/eventcache/eventcache.go
```diff
@@ -117,11 +117,27 @@ func HandleGenericInternal(ev notify.Event, pid uint32, tid *uint32, timestamp u
 // so we only need to wait for the internal link to the process context to
 // resolve PodInfo. This happens when the msg populates the internal state
 // but that event is not fully populated yet.
-func HandleGenericEvent(internal *process.ProcessInternal, ev notify.Event, tid *uint32) error {
+func HandleGenericEvent(internal *process.ProcessInternal, ev notify.Event, tid *uint32, containerID string) error {
```
kkourt: Please avoid the extra argument, and add a type switch on the event where we get the containerID if the type has it.

bpf/process/policy_filter.h
tpapagian: These change LGTM.

PR #4672, opened 2026-02-19
cleanup(pkg/observer): use `sync.Once` in `getDefaultObserver()`.

FedeDP:
### Description
Instead of manually using a global boolean flag to start metrics server during observer tests, we now use a `sync.Once` variable.

### DIFF
--- a/pkg/observer/observertesthelper/observer_test_helper.go
+++ b/pkg/observer/observertesthelper/observer_test_helper.go
@@ -48,8 +48,8 @@ import (
 )
 
 var (
-	metricsAddr    = "localhost:2112"
-	metricsEnabled = false
+	metricsAddr       = "localhost:2112"
+	metricsEnableOnce sync.Once
 )
 
 type testObserverOptions struct {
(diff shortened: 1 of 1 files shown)


pkg/observer/observertesthelper/observer_test_helper.go
```diff
@@ -236,19 +236,10 @@ func getDefaultObserver(tb testing.TB, ctx context.Context, initialSensor *senso
...
 	saveInitInfo(o, exportFname)
 
-	// There doesn't appear to be a better way to enable the metrics server once and only
-	// once at the beginning of the observer tests. My initial thought was to use the init
-	// function in this file, however that actually ends up interfering with the Tetragon agent
-	// since it get compiled into the observer package.
-	//
-	// This is horrifically ugly, so we may want to figure out a better way to do this
-	// at some point in the future. I just don't see a better way that doesn't involve
-	// a lot of code changes in a lot of a files.
```
kkourt, line 246: Any reason for not keeping the comment?
Even though I did not write it, I think it relates to having to start the metrics server from the the observer, rather than having a boolean variable vs a sync.Once.

FedeDP: Uh i thought it was talking about using the bool variable, but yep no problem in reverting the change.

olsajiri (approved):

Issue #4247, opened 2025-10-24
Missing file access events when running Tetragon on kind


AritraDey-Dev:
### What happened? How can we reproduce this?

Following the Tetragon setup in the official docs on a local kind cluster, I tried monitoring file access events as described in [Filename Access Use Case](https://tetragon.io/docs/use-cases/filename-access/)

When attempting to view logs as in the [Examples section](https://tetragon.io/docs/use-cases/filename-access/#examples) I only see:
```
🚀 process default/file-access /bin/cat /etc/passwd
💥 exit    default/file-access /bin/cat /etc/passwd 0
```
No logs for read or write operations are generated, even when performing actions like:
`vi /etc/passwd` or  `echo "test" >> /etc/passwd`  or `cat /etc/passwd`

### Tetragon Version

latest

### Kernel Version

6.8.0-50-generic

### Kubernetes Version

v1.34

### Relevant log output

```shell
🚀 process default/file-access /bin/cat /etc/passwd
💥 exit    default/file-access /bin/cat /etc/passwd 0
```

sayboras:
I tried with main branch, but am unable to reproduce the issue though

kex deployment/netshot -- sh -c 'echo "test" >> /etc/passwd'

```
➜  tetragon git:(main) kubectl exec -it -n tetragon ds/tetragon -c tetragon -- tetra getevents -o compact --namespace default
🚀 process default/netshot-7fd446d5dc-dxqvq /bin/sh -c "echo "test" >> /etc/passwd" 
📚 read    default/netshot-7fd446d5dc-dxqvq /bin/sh /etc/passwd           
📝 write   default/netshot-7fd446d5dc-dxqvq /bin/sh /etc/passwd           
💥 exit    default/netshot-7fd446d5dc-dxqvq /bin/sh -c "echo "test" >> /etc/passwd" 0 
```

dangome3:
Hey @AritraDey-Dev 

I was also able to see the policy events:

```
🚀 process kind-control-plane /usr/bin/cat /etc/passwd
📝 mmap-r  kind-control-plane /usr/bin/cat /etc/ld.so.cache
📚 read    kind-control-plane /usr/bin/cat /etc/locale.alias
📚 read    kind-control-plane /usr/bin/cat /etc/locale.alias
📚 read    kind-control-plane /usr/bin/cat /etc/passwd
📚 read    kind-control-plane /usr/bin/cat /etc/passwd
```

Can you confirm that you still face this issue?
[...]

PR #4679, opened 2026-02-20
filename monitoring policies for kernels >= 6.2

AritraDey-Dev:
### Description

This fixes https://github.com/cilium/tetragon/issues/4247 and for details see https://github.com/cilium/tetragon/issues/4247#issuecomment-3801537861.
This was discussed with @kkourt in the latest tetragon community meeting, where it was agreed that we should update the docs and example policies to handle the `security_path_truncate` → `security_file_truncate` change in linux 6.2+.

### Changelog

```release-note
Document kernel `compatibility` issue for `security_path_truncate` on kernels 6.2+ and add compact encoder support for security_file_truncate.
```

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -45,6 +45,7 @@ import (
 	"github.com/cilium/tetragon/pkg/grpc/tracing"
 	"github.com/cilium/tetragon/pkg/jsonchecker"
 	"github.com/cilium/tetragon/pkg/kernels"
+	"github.com/cilium/tetragon/pkg/ksyms"
 	"github.com/cilium/tetragon/pkg/logger"
 	bc "github.com/cilium/tetragon/pkg/matchers/bytesmatcher"
 	lc "github.com/cilium/tetragon/pkg/matchers/listmatcher"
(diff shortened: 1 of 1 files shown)


docs/content/en/docs/use-cases/filename-access.md
```diff
@@ -233,6 +236,58 @@ Another example of a [similar
 policy](https://raw.githubusercontent.com/cilium/tetragon/main/examples/tracingpolicy/filename_monitoring_filtered.yaml)
 can be found in our examples folder.
 
+## Kernel compatibility
+
+The `security_path_truncate` hook used in the truncation monitoring was
+[refactored in Linux 6.2](https://github.com/torvalds/linux/commit/3350607dc5637be2563f484dcfe2fed456f3d4ff)
+and replaced by `security_file_truncate`. On kernels **>= 6.2**, the symbol
```
mtardy:
```suggestion
and replaced by `security_file_truncate`. On kernels >= 6.2, the symbol
```

docs/content/en/docs/use-cases/filename-access.md
```diff
@@ -233,6 +236,58 @@ Another example of a [similar
...
 
+## Kernel compatibility
+
+The `security_path_truncate` hook used in the truncation monitoring was
+[refactored in Linux 6.2](https://github.com/torvalds/linux/commit/3350607dc5637be2563f484dcfe2fed456f3d4ff)
+and replaced by `security_file_truncate`. On kernels >= 6.2, the symbol
+`security_path_truncate` is no longer available as a traceable kprobe entry point
+— it may be inlined or its visibility changed. When the tracing policy references
+a hook that does not exist in the running kernel, the **entire policy fails to
+load**, which means that no events (including reads and writes from the other
```
mtardy:
```suggestion
`security_path_truncate` is no longer available as a traceable kprobe entry point,
it may be inlined or its visibility changed. When the tracing policy references
a hook that does not exist in the running kernel, the entire policy fails to
load, which means that no events (including reads and writes from the other
```

AritraDey-Dev:
>Also if you want you can update directly https://raw.githubusercontent.com/cilium/tetragon/main/examples/tracingpolicy/filename_monitoring.yaml so that it works transparently for everyone? Instead of having them to see they need to change something from the docs? wdyt

Yeah, initially I thought of doing that, but I didn’t want to touch that policy since it is specific to filename monitoring. And the fix we are adding in this PR is specifically for the events like `write` and `read` that are not being fired for kernel 6.2+. So I’d prefer to keep them separate. If anyone encounters this issue, they can simply apply the fix mentioned in the docs.Let me know if that sounds okay.

PR #4686, opened 2026-02-20
Revert "ci: Fix package e2e tests GHA"

andrewstrohman:
This reverts commit 854fe5f40336f0c136bf624e609d199f72cea863.

https://github.com/cilium/tetragon/commit/854fe5f40336f0c136bf624e609d199f72cea863 was required because the tarball was double gzip'ed. The
tarball is no longer double gzip'ed, so revert.

If we run into this problem again, we should handle it by ensuring that the tarball is only gzip'ed once.

[Here](https://github.com/cilium/tetragon/actions/runs/22236687712/job/64330369791?pr=4666) is an example of how CI is failing without this revert.

```
gzip -dc tetragon-0f16f81-arm64.tar.gz | tar -zxvf - 
  sudo ./tetragon-0f16f81-arm64/install.sh
  shell: /usr/bin/bash -e {0}

gzip: stdin: not in gzip format
tar: Child died with signal 13
tar: Error is not recoverable: exiting now
```

### DIFF
--- a/.github/workflows/packages-e2e-tests.yaml
+++ b/.github/workflows/packages-e2e-tests.yaml
@@ -108,7 +108,7 @@ jobs:
 
       - name: Install Tetragon Tarball
         run: |
-          gzip -dc tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz | tar -zxvf - 
+          tar zxvf tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz
           sudo ./tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}/install.sh
         working-directory: ./build/${{ matrix.arch }}/linux-tarball/
 


kkourt:
> If we run into this problem again, we should handle it by ensuring that the tarball is only gzip'ed once.

@andrewstrohman how do we do this?
We are seeing the same failures in v1.6: https://github.com/cilium/tetragon/pull/4689#issuecomment-3945215765

andrewstrohman:
> > If we run into this problem again, we should handle it by ensuring that the tarball is only gzip'ed once.
> 
> @andrewstrohman how do we do this? We are seeing the same failures in v1.6: [#4689 (comment)](https://github.com/cilium/tetragon/pull/4689#issuecomment-3945215765)

That failure is happening to v1.6 because the same offending commit was applied to that branch as well. So, the fix is to [revert in v1.6](https://github.com/cilium/tetragon/pull/4695) too. I checked the other release branches that we support, and they do not have the offending commit applied, so no revert is required for them.

To answer you question about what to do if the double gzip tarball comes back. The double gzip problem happens to me locallly, when I run `tarball-release`.

With the following patch:
```
diff --git a/Makefile b/Makefile
index 8194e1ccf..47493030a 100644
--- a/Makefile
+++ b/Makefile
@@ -217,6 +217,7 @@ tarball: tarball-clean image ## Build Tetragon compressed tarball.
        sync $(BUILD_PKG_DIR)/docker/manifest.json
        cat $(BUILD_PKG_DIR)/docker/manifest.json
        cp "${BUILD_PKG_DIR}/docker/$$(jq -r '.[].Layers[0]' "${BUILD_PKG_DIR}/docker/manifest.json")" ${BUILD_PKG_DIR}/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar
+       file ${BUILD_PKG_DIR}/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar
        @tar -tf ${BUILD_PKG_DIR}/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar | grep "/usr/local/bin/tetragon" - \
                || (echo "make: '$@' Error: could not find tetragon inside generated tarball"; exit 1)
        $(QUIET)rm -fr $(BUILD_PKG_DIR)/tetragon-$(VERSION)-$(TARGET_ARCH).tmp.tar
```
I see:
> /home/andy/build/src/tetragon/build/amd64/linux-tarball/tetragon-v1.7.0-pre.0-526-g4098d301f-amd64.tar: gzip compressed data, original size modulo 2^32 250636288 gzip compressed data, reserved method, has CRC, from FAT filesystem (MS-DOS, OS/2, NT), original size modulo 2^32 250636288

So, I think that this issue that continues to affect my local environment, temporarily affected CI and then went away.

If it comes back, maybe we can do something like the following, so it works either way:

```
diff --git a/Makefile b/Makefile
index 8194e1ccf..fbe9a84fa 100644
--- a/Makefile
+++ b/Makefile
@@ -220,7 +220,11 @@ tarball: tarball-clean image ## Build Tetragon compressed tarball.
        @tar -tf ${BUILD_PKG_DIR}/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar | grep "/usr/local/bin/tetragon" - \
                || (echo "make: '$@' Error: could not find tetragon inside generated tarball"; exit 1)
        $(QUIET)rm -fr $(BUILD_PKG_DIR)/tetragon-$(VERSION)-$(TARGET_ARCH).tmp.tar
-       gzip -6 $(BUILD_PKG_DIR)/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar
+       if ! file $(BUILD_PKG_DIR)/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar | grep -q gzip; then \
+               gzip -6 $(BUILD_PKG_DIR)/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar; \
+       else \
+               mv $(BUILD_PKG_DIR)/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar $(BUILD_PKG_DIR)/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar.gz; \
+       fi
        @echo "Tetragon tarball is ready: $(BUILD_PKG_DIR)/linux-tarball/tetragon-$(VERSION)-$(TARGET_ARCH).tar.gz"

 .PHONY: tarball-release
```

I've confirmed that this change fixes my local environment.

PR #4689, opened 2026-02-23
fix(process/cache): balance parent refcount ops during LRU eviction (1.6)

dwindsor:
## Description

Backport of 8c9cf10b7 from `main` to `v1.6`.

When a child process gets LRU-evicted under cache pressure before its exit event arrives, the eviction callback only increments a metric without performing the parent refcount decrement. The child's exit handler later cannot find the child in the cache, so `parent--` never fires. This leaves the parent's refcount permanently elevated, preventing garbage collection and causing stale entries to accumulate — increasing eviction pressure further.

This fix decrements the parent refcount in the eviction callback, but only for entries still marked `inUse` (whose normal exit path has not already handled it), avoiding a double-decrement for normally-exiting processes.

### DIFF
--- a/pkg/process/cache.go
+++ b/pkg/process/cache.go
@@ -145,19 +145,41 @@ func NewCache(
 	processCacheSize int,
 	GCInterval time.Duration,
 ) (*Cache, error) {
+	// Stash a reference to the Cache to refer to later in the eviction closure.
+	pm := &Cache{
+		size: processCacheSize,
+	}
+
 	lruCache, err := lru.NewWithEvict(
 		processCacheSize,
-		func(_ string, _ *ProcessInternal) {
+		func(_ string, evicted *ProcessInternal) {
 			processCacheEvictions.Inc()
+
+			// Perform parent-- for LRU-evicted entries that will never
+			// reach the exit handler.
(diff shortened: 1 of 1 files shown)


dwindsor:
[standalone-tarball-tests (ubuntu-22.04, amd64, x86-64, no, upload/)](https://github.com/cilium/tetragon/actions/runs/22310666320/job/64542164984?pr=4689#logs) failing, looks unrelated:

```
Run gzip -dc tetragon-9482675-amd64.tar.gz | tar -zxvf - 

gzip: stdin: not in gzip format
tar: Child died with signal 13
tar: Error is not recoverable: exiting now

gzip: stdout: Broken pipe
Error: Process completed with exit code 2.
```

kkourt:
> [standalone-tarball-tests (ubuntu-22.04, amd64, x86-64, no, upload/)](https://github.com/cilium/tetragon/actions/runs/22310666320/job/64542164984?pr=4689#logs) failing, looks unrelated:


Yap it's unrelated, although I'm not sure what the fix is (https://github.com/cilium/tetragon/pull/4686#issuecomment-3949626993)

PR #4692, opened 2026-02-24
tetragon: string related fixes

olsajiri:
- fix for substring matching
- fix for empty string matching


olsajiri:
note CI lint is failing on string += op, but fixing that would mean rewiting the test, so let's do that in some follow up

PR #4703, opened 2026-02-27
tetragon: Fix substring filter to match all possible spec values

olsajiri:
We allow 100 substring values, but bpf filter code uses just 4, let's fix it.


olsajiri:
note there's unrelated CI lint fail.. I'll submit separate change for the test

PR #4704, opened 2026-02-27
bugtool: add extra files support

will-isovalent:
Add a file-based mechanism for the daemon to register additional files for collection in bugtool archives. At startup, the daemon writes a JSON map of name/path pairs to a well-known file (`tetragon-bugtool-extra-files.json` in the run directory). At bugtool time, the CLI reads this file and streams each registered entry into the tar archive.

This uses a dedicated file rather than extending `InitInfo` to keep initialization config and bugtool file registration as independent concerns. The extra files file is optional: if it doesn't exist, the bugtool silently continues, so existing deployments are unaffected.

Callers invoke `SaveExtraFiles` at startup to register files whose paths are only known from runtime config. Since the CLI can't discover these paths without reading something the daemon wrote, the file-based IPC works even when the daemon has crashed, which is a key bugtool use case.

```release-note
Bugtool now supports collecting extra files registered by the daemon at startup via a separate JSON manifest file.
```

### DIFF
--- a/pkg/bugtool/bugtool_test.go
+++ b/pkg/bugtool/bugtool_test.go
@@ -4,8 +4,10 @@
 package bugtool
 
 import (
+	"encoding/json"
 	"io"
 	"os"
+	"path/filepath"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


will-isovalent:
The lint failure is a pre-existing issue and is completely unrelated to any code touched in this branch.

pkg/bugtool/bugtool.go
```diff
@@ -43,17 +43,18 @@ import (
...
+	BTFFname    string            `json:"btf_fname"`
+	ServerAddr  string            `json:"server_address"`
+	MetricsAddr string            `json:"metrics_address"`
+	GopsAddr    string            `json:"gops_address"`
+	MapDir      string            `json:"map_dir"`
+	BpfToolPath string            `json:"bpftool_path"`
+	GopsPath    string            `json:"gops_path"`
+	MaxRecvSize int               `json:"max_recv_size"`
+	PID         int               `json:"pid"`
+	ExtraFiles  map[string]string `json:"extra_files,omitempty"`
```
mtardy: do you really need the "name" for docs purposes of what is this extra files? Otherwise I imagine you can just the path base.

mtardy:
> The lint failure is a pre-existing issue and is completely unrelated to any code touched in this branch.

fyi should go away if you rebase, it's been fixed on main

pkg/bugtool/bugtool.go
will-isovalent: Yes indeed, the motivation is to be able to provide a consistent and descriptive name (e.g. foo_export.json rather than whatever might be configured in tetragon.options.FooExportFilename).

will-isovalent:
> fyi should go away if you rebase, it's been fixed on main

Yep I saw that, thanks!

will-isovalent:
> looks a bit hacky to be in this init file but works!

Yeah I agree, it is definitely a hack (partially why I left this in draft). Let me see if I can come up with something that works a bit better.

PR #4706, opened 2026-02-28
fix: Added WaitCRDsWithResync with resync period

### LINKED ISSUE #4707: Controller manager's WaitCRDs gives error "failed to add event handler"
Error message in Client
```
time=2026-02-24T19:58:12Z level=info msg="Waiting for required CRDs" crds=map[smartswitchnetworkpolicies.isovalent.com:{}]
...
time=2026-02-24T19:58:13Z level=error msg="failed to add event handler" error="handler {0x18139c0 <nil> <nil>}
```

Impact: When an error is returned from WaitCRDs, the client exits and restarts.

shpalani:
Fixes #4707

### Description
Problem:
Client uses Tetragon controller manager for connecting to external K8s api_server and getting resources from CRDs.
There appears to be a race condition:
If WaitCRDs is called early when initial sync might still be in progress, but the Informer's AddFunc only triggers on new CRD events. Then we'd miss that CRD and wait for the CRD that's already there. 
Also, error we see is from s.stopped where the informer is getting stopped, as a result of RemoveInformer. This can cause race condition using the shared informer, where one stops it while another tries to add handler, from the same informer cache. The client has more than 1 CRD, trying to access at startup. 

Error message:
```
time=2026-02-24T19:58:13Z level=error msg="failed to add event handler" error="handler {0x18139c0 <nil> <nil>} 
```
 
 The default resync period for AddEventHandler() in controller-runtime is 10 hours (defaultSyncPeriod = 10 * time.Hour).
In controller-runtime/pkg/cache/cache.go
var (
	defaultSyncPeriod = 10 * time.Hour
)

### Solution:
- Added a new function WaitCRDsWithResync, instead of modifying existing WaitCRDs.
- Use AddEventHandlerWithResyncPeriod with configurable resync period from caller, instead of default 10h
- Add completion flag to prevent duplicate processing after CRDs found
- Add context cancellation handling for graceful shutdown in the client

Behavior Flow Before (AddEventHandler):
1. Informer starts
2. Initial cache sync happens → AddFunc called for existing CRDs. The CRD we are looking is not present.
3. Wait indefinitely for new CRD events
4. Only get AddFunc calls when NEW CRDs are created.
5. Resync every 10 HOURS (default) → AddFunc called for ALL existing CRDs
6. RemoveInformer, when found.

Behavior Flow After (WaitCRDsWithResync with AddEventHandlerWithResyncPeriod):
1. Informer starts  
2. Check if CRDs already in cache -> We have what we need, return immediately. No event handler needed.
3. For CRDs missing from cache, 
  4. Initial cache sync happens → AddFunc called for existing CRDs. The CRD we are looking is not present.
  5. Wait for new CRD events OR resync
  7. Get AddFunc calls when NEW CRDs are created
  8. Resync every configurable SECONDS → AddFunc called for ALL existing CRDs. The CRD can be found.
  9. Context cancellation handling → graceful shutdown

### Logs from Client invocation:
err = kubernetesManager.WaitCRDsWithResync(ctx, crds, 60*time.Second)
```
time=2026-02-28T17:20:55Z level=info msg="starting Kubernetes Manager for on-prem deployment"                         
...
time=2026-02-28T17:20:55Z level=info msg="Waiting for required CRDs with resync" crds=map[smartswitchnetworkpolicies.isovalent.com:{}] resyncPeriod=1m0s
time=2026-02-28T17:20:56Z level=info msg="Found CRD in cache" crd=smartswitchnetworkpolicies.isovalent.com            
time=2026-02-28T17:20:56Z level=info msg="Found all required CRDs in cache"  
```
```
time=2026-02-28T01:11:50Z level=info msg="Waiting for required CRDs with resync" crds=map[smartswitchnetworkpolicies.isovalent.com:{}] resyncPeriod=1m0s                   
...                                    
time=2026-02-28T01:11:50Z level=info msg="\"Warning: resync period is smaller than resync check period and the informer has already started. Changing it to the resync check period\" resyncPeriod=\"30s\" resyncCheckPeriod=\"9h27m38.395867694s\"" subsys=klog                                                                                                                             
time=2026-02-28T01:11:50Z level=info msg="Found CRD" crd=smartswitchnetworkpolicies.isovalent.com                               
time=2026-02-28T01:11:50Z level=info msg="Found all the required CRDs"
```

Detailed notes of the fix:
```
### Introduces New WaitCRDsWithResync function that addresses CRD Wait Mechanism
#### Fixes Missing CRDs During Early Initialization in WaitCRDsWithResync
- **Problem**: If `WaitCRDs` was called early when initial sync might still be in progress, the `AddFunc` would only trigger on new CRD events, causing the function to miss CRDs that were already present in the cluster
- **Solution**: Added cache pre-check that queries existing CRDs from the informer cache before setting up event handlers
- **Cache validation**: Check `informer.HasSynced()` to detect if cache is not yet synchronized
- **Immediate detection**: CRDs already present in the cache are detected immediately without waiting for events
- **Fallback mechanism**: If cache listing fails, the function gracefully falls back to event-driven approach

#### Fixes SharedIndexInformer Race Condition in multiple waitCRDs calls
 - **Problem**: The `WaitCRDs` function was calling `RemoveInformer()` after waiting for CRDs, which could cause race conditions where other parts of the code would get "handler was not added to shared informer because it has stopped already" intermittent errors when trying to add event handlers to the same shared informer
- **Root Cause**: Controller-runtime returns shared informers via `GetInformer()`, and removing them affects other consumers to use same shared CRD informer later
- **Solution**: Did not add `cm.Manager.GetCache().RemoveInformer(ctx, &apiextensionsv1.CustomResourceDefinition{})` call in `WaitCRDsWithResync`

#### New AddEventHandlerWithResyncPeriod in WaitCRDsWithResync
- **Resync support**: Enhanced `WaitCRDsWithResync` function now uses `AddEventHandlerWithResyncPeriod` instead of basic `AddEventHandler`
- **Configurable resync period**: Allows specification of resync period for CRD discovery operations to handle missed events
- **Improved reliability**: Periodic resync ensures CRDs are detected even if initial events are missed during informer startup
- **Context Cancel Aware**: Handling context cancel during resync wait period
```

### Release note
```release-note
Introduces WaitCRDsWithResync function with cache pre-check, configurable resync period, cancel context aware, and eliminates SharedIndexInformer race conditions by avoiding RemoveInformer calls.
```

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -138,6 +138,26 @@ func (cm *ControllerManager) ListNamespaces() ([]corev1.Namespace, error) {
 	return namespaceList.Items, nil
 }
 
+// WaitCRDs waits for the specified Custom Resource Definitions (CRDs) to become available
+// in the Kubernetes cluster using a simple event-driven approach. This method sets up an
+// event handler that responds to CRD addition events and completes when all required CRDs
+// are found. It does not perform periodic resyncs or cache refreshes, making it suitable
+// for scenarios where CRDs are expected to be created shortly after startup and network
+// connectivity is stable.
+//
+// Use WaitCRDs when:
+//   - You expect CRDs to be available soon after application startup
+//   - Network connectivity to the API server is reliable
+//   - You want minimal resource overhead (no periodic resyncs)
+//   - The application can tolerate potentially missing CRDs that exist but weren't detected
+//     due to missed events or temporary network issues
+//
+// Parameters:
+//   - ctx: Context for cancellation
+//   - crds: Map of CRD names to wait for (map keys are CRD names, values are ignored)
+//
+// Returns:
+//   - error: Non-nil if waiting fails or context is cancelled
 func (cm *ControllerManager) WaitCRDs(ctx context.Context, crds map[string]struct{}) error {
 	log := logger.GetLogger()
 	log.Info("Waiting for required CRDs", "crds", crds)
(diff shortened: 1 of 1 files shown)


pkg/manager/manager.go
```diff
@@ -176,6 +176,144 @@ func (cm *ControllerManager) WaitCRDs(ctx context.Context, crds map[string]struc
 	return nil
 }
 
+// WaitCRDsWithResync waits for CRDs with configurable resync period and context cancellation logic
+func (cm *ControllerManager) WaitCRDsWithResync(ctx context.Context, crds map[string]struct{}, resyncPeriod time.Duration) error {
```
will-isovalent: Rather than adding a new function, it might be cleaner to update the existing caller and replace WaitCRDs with this new version. Otherwise, ostensibly we are both duplicating code here and leaving an older version in place that is theoretically vulnerable to the race you describe in the original issue.

pkg/manager/manager.go
```diff
@@ -176,6 +176,144 @@ func (cm *ControllerManager) WaitCRDs(ctx context.Context, crds map[string]struc
...
+			// CRDs found, normal completion
+		}
+	}()
+
+	wg.Wait()
+
+	// Do not remove the informer, as GetInformer may return a shared informer that is used elsewhere.
+	// We're only adding an event handler, not owning the informer.
+	// The controller-runtime manager handles informer cleanup automatically when stopped
+	return nil
```
will-isovalent, line 341: When the context is cancelled while waiting for CRDs, the goroutine at line 293 unblocks the WaitGroup, but the function unconditionally returns nil at line 314. The caller has no way to distinguish "all CRDs found" from "context was cancelled before finding CRDs". It should return `ctx.Err()` when cancellation caused the unblock.

pkg/manager/manager.go
```diff
@@ -176,6 +176,144 @@ func (cm *ControllerManager) WaitCRDs(ctx context.Context, crds map[string]struc
...
+					log.Debug("Resync event detected",
+						"resyncPeriod", resyncPeriod,
+						"resyncCount", resyncCount,
+						"stillWaiting", len(remainingCRDs),
+						"timeSinceLastResync", now.Sub(lastResyncTime))
+					lastResyncTime = now
+				}
+			}
+		},
+	}, resyncPeriod)
```
will-isovalent, line 308: Maybe it would be cleaner to use `sync.Once()` here instead of this completion flag mechanism?

will-isovalent:
Also, please remember to add a release note to your PR description and set the appropriate PR labels. Please also squash those earlier commits before merging.

The lint failure can be fixed by a rebase, it's unrelated to these changes.

pkg/manager/manager.go
shpalani: My thought flow was to provide a new function, as it replaces  AddEventHandler with AddEventHandlerWithResyncPeriod from controller-runtime, to maintain backward compatability. 
Also, the waitCRDs does not provide cancel context aware and cache lookup.

I have added comments in both the functions. It is upto the caller of this controller manager to use the right informer function.

pkg/manager/manager.go
shpalani: The current approach is reasonable because
1. The boolean flag is simple and clear to to prevent further processing.
2. The select pattern is  actually quite readable than do once

```
completed := false    // Boolean flag

// In AddFunc:
if completed {
    return           // Short-circuit further processing
}
// ... process CRD ...
if len(remainingCRDs) == 0 {
    completed = true
    select {
    case <-done:
        // Already completed  
    default:
        close(done)   // Ensure only called once
        wg.Done()
    }
}

// In context cancellation goroutine:
select {
case <-done:
    // Already completed
default:
    close(done)       // Ensure only called once  
    wg.Done()
}
```

pkg/manager/manager.go
shpalani: My thought flow here was,
the application/caller is issuing the context cancel upon an exit/restart, hence the return error doesn't matter. 

But it is a good point, I will fix it.

shpalani:
> Thanks for the fix. Left a few comments.
> 
> The logic here is also pretty complex. It would probably be worth adding some tests.

In testing in client, the CRD is retrieved mostly from cache (improves performance), so the event handler add is just a fallback.
err = cm.Manager.GetCache().List(ctx, existingCRDs)


And also in this case, the client is present outside of the cluster and has multiple CRDs, so there needs to be reliability.

shpalani:
> Also, please remember to add a release note to your PR description and set the appropriate PR labels. Please also squash those earlier commits before merging.
> 
> The lint failure can be fixed by a rebase, it's unrelated to these changes.

I have rebased. Can you please add the release/* in the labels, as I do not have permission.

will-isovalent:
@shpalani for the release-note section, we're looking for a 1 line plaintext description of what was added. I think what you have right now (the detailed markdown writeup) will probably break our automation pipeline that creates the release notes by reading these code blocks.

pkg/manager/manager.go
will-isovalent: @shpalani after looking at it again, I think there actually might be a concurrency bug here. It would be possible for both the select in AddFunc and the select in context cancellation to hit the default branch concurrently and cause a double close. Here's a go playground link to demonstrate the issue: https://go.dev/play/p/NyiOZzzl5CV

shpalani: Fixed. Thanks for the playground link.

shpalani:
> @shpalani for the release-note section, we're looking for a 1 line plaintext description of what was added. I think what you have right now (the detailed markdown writeup) will probably break our automation pipeline that creates the release notes by reading these code blocks.

Added a one line description. Thanks.

PR #830, opened 2023-03-22
kprobes: add a prevalidate kprobe semantics phase

tixxdz:
kprobes: pre-validate kprobe semantics
    
This adds a pre-validate kprobe semantics phase in order to:
1. Separate kprobe semantics errors from BPF related ones.
2. Fail early without probing BPF on some errors and avoid loading BPF resources
3. Reduce user confusion if a policy was loaded or not

Before:
```
time="2023-03-22T12:48:47+01:00" level=debug msg="Received an AddTracingPolicy request" request="yaml:\"apiVersion: cilium.io/v1alpha1\\nkind: TracingPolicy\\nmetadata:\\n name: \\\"fake\\\"\\nspec:\\n kprobes:\\n - call: \\\"sys_fake\\\"\\n   syscall: true\\n\""
time="2023-03-22T12:48:47+01:00" level=warning msg="invalid or old kprobe spec: kprobe spec validation failed: call __x64_sys_fake not found"
time="2023-03-22T12:48:47+01:00" level=debug msg="Kprobe spec pre-validation succeeded" name=gkp-sensor-1
time="2023-03-22T12:48:47+01:00" level=info msg="Added generic kprobe sensor: bpf/objs/bpf_generic_kprobe_v53.o -> __x64_sys_fake"
time="2023-03-22T12:48:47+01:00" level=debug msg="Checking for bpf file" file=bpf/objs/bpf_generic_kprobe_v53.o
time="2023-03-22T12:48:47+01:00" level=debug msg="Found bpf file" file=bpf/objs/bpf_generic_kprobe_v53.o
...
time="2023-03-22T12:48:47+01:00" level=debug msg="Found bpf file" file=bpf/objs/bpf_generic_kprobe_v53.o
time="2023-03-22T12:48:47+01:00" level=info msg="tetragon, map loaded." map=fdinstall_map path=/sys/fs/bpf/tetragon/gkp-sensor-1-fdinstall_map sensor=gkp-sensor-1
time="2023-03-22T12:48:47+01:00" level=info msg="tetragon, map loaded." map=config_map path=/sys/fs/bpf/tetragon/gkp-sensor-1-gkp-0-config_map sensor=gkp-sensor-1
time="2023-03-22T12:48:47+01:00" level=info msg="tetragon, map loaded." map=kprobe_calls path=/sys/fs/bpf/tetragon/gkp-sensor-1-gkp-0-kp_calls sensor=gkp-sensor-1
[...]

PR #4708, opened 2026-02-28
tracing: add pre-validation for tracepoints and LSM hooks

AritraDey-Dev:
### Description

This PR introduces pre validation for tracepoints and LSM hooks for tracingpolicy, following up on the initial pre-validation work that was done for kprobes in PR #830(See commit descriptions  for more details).

### Changelog

```release-note
Added pre-validation for tracepoints and LSM hooks to reject invalid `TracingPolicies` before BPF resources are created.
```

### DIFF
--- a/pkg/sensors/tracing/lsm_validation_test.go
+++ b/pkg/sensors/tracing/lsm_validation_test.go
@@ -1,0 +1,145 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build !windows
+
+package tracing
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/cilium/tetragon/pkg/bpf"
+	"github.com/cilium/tetragon/pkg/config"
+	"github.com/cilium/tetragon/pkg/tracingpolicy"
+)
+
+func TestLsmValidationBogusHook(t *testing.T) {
+	if !bpf.HasLSMPrograms() || !config.EnableLargeProgs() {
+		t.Skip("LSM programs not supported on this kernel")
+	}
+
+	crd := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm-bogus-hook"
+spec:
+  lsmhooks:
+  - hook: "bogus_nonexistent_hook_xyz"
+`
+
+	err := checkCrd(t, crd)
+	require.Error(t, err)
+}
+
+func TestLsmValidationEmptyHook(t *testing.T) {
+	if !bpf.HasLSMPrograms() || !config.EnableLargeProgs() {
+		t.Skip("LSM programs not supported on this kernel")
+	}
+
+	crd := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm-empty-hook"
+spec:
+  lsmhooks:
+  - hook: ""
+`
+
+	err := checkCrd(t, crd)
+	require.Error(t, err)
+}
+
+func TestLsmValidationInvalidSelector(t *testing.T) {
+	if !bpf.HasLSMPrograms() || !config.EnableLargeProgs() {
+		t.Skip("LSM programs not supported on this kernel")
+	}
+
+	crd := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm-bad-selector"
+spec:
+  lsmhooks:
+  - hook: "file_open"
+    selectors:
+    - matchReturnArgs:
+      - index: 0
+        operator: "Equal"
+        values:
+        - "0"
+`
+
+	err := checkCrd(t, crd)
+	require.Error(t, err)
+}
+
+func TestLsmValidationArgIndexOutOfBounds(t *testing.T) {
+	if !bpf.HasLSMPrograms() || !config.EnableLargeProgs() {
+		t.Skip("LSM programs not supported on this kernel")
+	}
+
+	crd := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm-arg-oob"
+spec:
+  lsmhooks:
+  - hook: "file_open"
+    args:
+    - index: 5
+      type: "int"
+`
+
+	err := checkCrd(t, crd)
+	require.Error(t, err)
+}
+
+func TestLsmValidationInvalidArgType(t *testing.T) {
+	if !bpf.HasLSMPrograms() || !config.EnableLargeProgs() {
+		t.Skip("LSM programs not supported on this kernel")
+	}
+
+	crd := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm-bad-argtype"
+spec:
+  lsmhooks:
+  - hook: "file_open"
+    args:
+    - index: 0
+      type: "bogus_type_xyz"
+`
+
+	_, err := tracingpolicy.FromYAML(crd)
+	require.Error(t, err)
+}
+
+func TestLsmValidationValidPolicy(t *testing.T) {
+	if !bpf.HasLSMPrograms() || !config.EnableLargeProgs() {
+		t.Skip("LSM programs not supported on this kernel")
+	}
+
+	crd := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm-valid"
+spec:
+  lsmhooks:
+  - hook: "file_open"
+    args:
+    - index: 0
+      type: "file"
+`
+
+	err := checkCrd(t, crd)
+	require.NoError(t, err)
+}
--- a/pkg/sensors/tracing/tracepoint_validation_test.go
+++ b/pkg/sensors/tracing/tracepoint_validation_test.go
@@ -1,0 +1,157 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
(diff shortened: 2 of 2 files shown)


pkg/sensors/tracing/generictracepoint.go
```diff
@@ -360,20 +363,102 @@ func buildArgsRaw(info *tracepoint.Tracepoint, specArgs []v1alpha1.KProbeArg) ([
...
+			if arg.Index >= nfields {
+				return nil, fmt.Errorf("tracepoint %s/%s has %d fields but field %d was requested in args[%d]",
+					spec.Subsystem, spec.Event, nfields, arg.Index, i)
+			}
+		}
+	} else {
+		// For raw tracepoints, argument index must be <= 5
+		for i, arg := range spec.Args {
+			if arg.Index > 5 {
+				return nil, fmt.Errorf("raw tracepoint %s/%s can read up to 5 arguments, but index %d was requested in args[%d]",
```
olsajiri: checks and errors like this are now doubled, also in lsm case, is there any way we could have preValidate hook and probe setup code sharing the same check?

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -360,20 +363,102 @@ func buildArgsRaw(info *tracepoint.Tracepoint, specArgs []v1alpha1.KProbeArg) ([
...
+
+// preValidateTracepoints pre-validates the semantics of tracepoint specs.
+// It checks that each tracepoint subsystem/event exists and that arguments
+// are valid. It also validates that NotifyEnforcer actions have enforcers.
+func preValidateTracepoints(tracepoints []v1alpha1.TracepointSpec, enforcers []v1alpha1.EnforcerSpec) ([]*tpValidateInfo, error) {
+	ret := make([]*tpValidateInfo, len(tracepoints))
+	for i := range tracepoints {
+		for _, sel := range tracepoints[i].Selectors {
+			for _, act := range sel.MatchActions {
+				if act.Action == "NotifyEnforcer" && len(enforcers) == 0 {
```
olsajiri: there's `HasNotifyEnforcerAction` , it's kprobe specific, but perhaps we could change the argument to be spec.Selectors directly

pkg/sensors/tracing/kprobe_validation_test.go
```diff
@@ -387,3 +387,291 @@ spec:
 	err := checkCrd(t, crd)
 	require.Error(t, err)
 }
+
+func TestTracepointValidationWrongSubsystem(t *testing.T) {
```
olsajiri: we have validation tests in separate files for each probe, let's add files for tracepoint and lsm as well

olsajiri (commented):
thanks, left some comments

pkg/sensors/tracing/generictracepoint.go
AritraDey-Dev: yes done thanks.

pkg/sensors/tracing/kprobe_validation_test.go
AritraDey-Dev: yeah sounds good!

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -245,8 +265,8 @@ func buildArgs(info *tracepoint.Tracepoint, specArgs []v1alpha1.KProbeArg) ([]ge
 
 	for argIdx := range specArgs {
 		specArg := &specArgs[argIdx]
-		if specArg.Index >= nfields {
-			return nil, fmt.Errorf("tracepoint %s/%s has %d fields but field %d was requested", info.Subsys, info.Event, nfields, specArg.Index)
+		if err := validateTracepointArg(info.Subsys, info.Event, nfields, specArg.Index, argIdx); err != nil {
```
olsajiri, line 271: I think these changes should be squashed into previous commit

AritraDey-Dev:
CI failure seems to be unrelated.

pkg/sensors/tracing/genericlsm.go
```diff
@@ -205,6 +207,72 @@ func isValidLsmSelectors(selectors []v1alpha1.KProbeSelector) error {
...
+	}
+
+	return nil
+}
+
+// preValidateLsmHooks pre-validates the semantics of LSM hook specs.
+// It checks that the kernel supports BPF LSM, that each hook exists in BTF,
+// and that arguments and selectors are valid.
+func preValidateLsmHooks(lsmHooks []v1alpha1.LsmHookSpec) error {
+	if !bpf.HasLSMPrograms() || !config.EnableLargeProgs() {
```
olsajiri: we have the same check in `createGenericLsmSensor`, let's keep just one,
I have slight preference on keep the one in `createGenericLsmSensor` .. but not really strong ;-)

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -407,7 +407,6 @@ func preValidateTracepoint(spec *v1alpha1.TracepointSpec) error {
 			}
 		}
 	} else {
-		// For raw tracepoints, argument index must be <= 5
```
olsajiri: this belongs to previous commit

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -360,20 +382,94 @@ func buildArgsRaw(info *tracepoint.Tracepoint, specArgs []v1alpha1.KProbeArg) ([
...
 		return nil, errors.New("failed creating generic tracepoint, conf is nil")
 	}
 
-	tp := tracepoint.Tracepoint{
-		Subsys: conf.Subsystem,
-		Event:  conf.Event,
+	// Use the pre-loaded tracepoint info from validation if available,
+	// otherwise create a new one (format will be loaded in buildGenericTracepointArgs).
+	var tp tracepoint.Tracepoint
+	if valInfo != nil && valInfo.tp != nil {
```
olsajiri: could `valInfo.tp` be nil ?

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -360,20 +382,94 @@ func buildArgsRaw(info *tracepoint.Tracepoint, specArgs []v1alpha1.KProbeArg) ([
...
 	if conf == nil {
 		return nil, errors.New("failed creating generic tracepoint, conf is nil")
 	}
 
-	tp := tracepoint.Tracepoint{
-		Subsys: conf.Subsystem,
-		Event:  conf.Event,
+	// Use the pre-loaded tracepoint info from validation if available,
+	// otherwise create a new one (format will be loaded in buildGenericTracepointArgs).
+	var tp tracepoint.Tracepoint
```
olsajiri: the rest of the code uses &tp, so let's start with that and use the pointer from the beginning

pkg/sensors/tracing/generictracepoint.go
AritraDey-Dev: No, `valInfo.tp` should never be nil if it comes from the pre-validation step. I only added the` valInfo.tp != nil `check to be extra safe and avoid any potential nil pointer panics in case someone manually calls this function without running validation first.If you think it's unnecessary, I can remove it!

pkg/sensors/tracing/generictracepoint.go
AritraDey-Dev: Thanks,this perfectly avoids copying the struct.

pkg/sensors/tracing/genericlsm.go
```diff
@@ -205,6 +207,67 @@ func isValidLsmSelectors(selectors []v1alpha1.KProbeSelector) error {
...
+		return errors.New("lsm hook name is empty")
+	}
+
+	// Validate selectors
+	if err := isValidLsmSelectors(f.Selectors); err != nil {
+		return fmt.Errorf("invalid selectors: %w", err)
+	}
+
+	// Validate the hook exists in BTF by looking up bpf_lsm_<hook>
+	btfFunc := "bpf_lsm_" + f.Hook
```
olsajiri: nit, we do this also in `addLsm` would be good to have helper that returns the name

pkg/sensors/tracing/generictracepoint.go
```diff
@@ -379,59 +382,71 @@ func buildArgsRaw(info *tracepoint.Tracepoint, specArgs []v1alpha1.KProbeArg) ([
 	return ret, nil
 }
 
+// tpValidateInfo holds pre-validated tracepoint information from the validation
+// phase. It is passed to createGenericTracepoint so that we don't need to load
+// the tracepoint format from tracefs a second time.
+type tpValidateInfo struct {
+	tp *tracepoint.Tracepoint
```
mtardy: would that be more simple if your struct just hold the type instead of a pointer to the type? is it useful that to be able to be null?

pkg/sensors/tracing/lsm_validation_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/cilium/tetragon/pkg/bpf"
+	"github.com/cilium/tetragon/pkg/config"
+	"github.com/cilium/tetragon/pkg/tracingpolicy"
+)
+
+func TestLsmValidationBogusHook(t *testing.T) {
+
```
mtardy: nit: remove all the empty lines after the function name

pkg/sensors/tracing/generictracepoint.go
AritraDey-Dev: Yes,there's no need for tp to be nullable inside since the entire  `tpValidateInfo` wrapper can just be nil if validation fails,updated it!

AritraDey-Dev:
@mtardy since it's already approved,is there anything blocking this PR to merge

PR #4717, opened 2026-03-02
celbpf: support multi-kprobes

kkourt:
multi-kprobe support for celbpf

### DIFF
--- a/pkg/sensors/tracing/celbpf_test.go
+++ b/pkg/sensors/tracing/celbpf_test.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"fmt"
 	"testing"
+	"unsafe"
 
 	"github.com/stretchr/testify/require"
 	"golang.org/x/sys/unix"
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
LGTM!

pkg/sensors/tracing/celbpf_test.go
```diff
@@ -55,10 +56,6 @@ func TestCelExpr(t *testing.T) {
 			Name: "lseek-celexpr",
 		},
 		Spec: v1alpha1.TracingPolicySpec{
-			Options: []v1alpha1.OptionSpec{{
```
olsajiri, line 58: should we keep this one to make sure kprobe_multi is disabled for this test? you just added kprobe-multi specific tests, right?

kkourt: yap, let's keep it. Pushed a new version.

PR #4719, opened 2026-03-02
bpf: use FNV-1a variant for ratelimit hash

msuozzo:
FNV-1a is a very simple non-cryptographic hash algorithm. It accumulates
each byte of input using an XOR and a multiply. As a result, its code
size is ~10x smaller than comparable hash functions.

However byte-by-byte hash functions often struggle over large input.
Here, FNV-1a is nearly 10x slower than a comparable algorithm, Murmur
Hash, for 1K inputs (about the upper bound of what we'd care about
here).

Word-by-word hash functions like Murmur Hash can be much faster but risk
insufficient "mixing" across bits within each word. To provide this
bit-mixing, Murmur Hash defines a final mixer ("fmix").

A relatively common way of repurposing a byte-by-byte hasher to word
size is to add a mixer and doing so here with "fmix" on top of FNV-1a
has great results with no loss of collision-resistence, randomness, or
skewedness.

Code size:

FNV-1a 8mix - 100 ins
FNV-1a      -  35 ins
MurmurHash3 - 350 ins
Jenkins     - 400 ins

1K Hash Speed:

FNV-1a 8mix -  150 ns
FNV-1a      - 1500 ns
MurmurHash3 -  175 ns
Jenkins     - 2500 ns

> Parents: [#4696](https://redirect.github.com/cilium/tetragon/pull/4696)

### DIFF
--- a/bpf/process/heap.h
+++ b/bpf/process/heap.h
@@ -9,7 +9,7 @@
 struct heap_ro_value {
 	union {
 		char string_maps_heap[STRING_MAPS_HEAP_SIZE];
-		char ratelimit_heap[sizeof(struct ratelimit_key) + 128];
+		char ratelimit_heap[sizeof(struct ratelimit_key) + MAX_HASH_BYTES];
 	};
 };
 
--- a/bpf/process/ratelimit_maps.h
+++ b/bpf/process/ratelimit_maps.h
@@ -4,12 +4,6 @@
 #ifndef __RATELIMIT_MAPS_H__
 #define __RATELIMIT_MAPS_H__
 
-/* The number of bytes per argument to include in the key
- * that we use to check for repeating data.
- * 40 is good for IPv6 data.
- */
-#define KEY_BYTES_PER_ARG 40
-
 /* Can't really include generic.h in here..
  */
 #define MAX_POSSIBLE_ARGS 5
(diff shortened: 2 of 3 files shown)


msuozzo:
bump

mtardy (commented):
Hey 👋  so I think the reasoning behind the PR seems to make sense, I lack a bit of context on this and would need to investigate a bit more on how rate limit currently work. 

Anyway I haven't verified this but if it's true, it totally makes sense to hash here.
```
The existing logic used the first 40 arg bytes concatenated as the
ratelimit cache key. One difficulty with this approach is that file
paths often share substantial prefixes causing spurious drops for
distinct events. In order to reduce false negatives for these cases, we
can hash argument contents instead.
```

Could @kevsecurity take a look and say what he thinks about that? We also have performance implications here on that hotpath?

Also I'd reorganize a bit your PR maybe:
- first patch: introduce your new BPF function for hashing
- second patch: wire your new hashing function into the rate limit feature

msuozzo:
Thanks @mtardy! I'd be happy to split it up.

Was especially interested to see CI results here but it looks like it may have broken a bunch of tests 🙁 . I was only able to get it working on an older kernel (early 5.X?) using --force-small-progs so that may be a issue that needs addressing.

mtardy:
> Was especially interested to see CI results here but it looks like it may have broken a bunch of tests 🙁 . I was only able to get it working on an older kernel (early 5.X?) using --force-small-progs so that may be a issue that needs addressing.

yeah that's tricky, feel free to push updates here I'll try to click the button so that the CI runs. But to make sure kernel versions works for you, I'd use [cilium/little-vm-helper](https://github.com/cilium/little-vm-helper), the README will explain how to download the images used in the CI and with host-mount and port options it's easy to SSH into the machine and share the folder you are working on (like tetragon) to run inside the VM:

On x86_64 that would be something like that (see the readme for more info)
```
lvh run --image ~/_data/images/<your image>.qcow2 --port 2222:22 --host-mount ~/tetragon --cpu 4 --mem 8G
```

kevsecurity:
👋 Sorry I've not been giving this attention; was unfortunately busy with other things. I've had a look through and I like the idea. I think the performance issues @mtardy are only limited to people choosing to use rate limiting, and they are likely trading off one perf issue for another, so I don't think it is an issue for them.

To move forwards, I think we only need one hash in the PR, so if the mix version makes more sense, let's just introduce that. I'd agree to use pass-through for vars <= 8 bytes as it's cheaper and we're only really using a hash to shorten the input. As @mtardy suggested, maybe commit 1 could be "add the hash function but don't use it", and commit 2 could be "change rate limit to use the new hash". It would be nice to add tests in a further commit: tests that show the hash is working, and tests that show rate limiting is working with parameters <= 8 bytes and total args > 40 bytes.

My advice around complexity would be two-fold: first, unless you have particularly old kernels, you only need to introduce this functionality on later kernels, using an existing macro like ```__V61_BPF_PROG``` to compile the new version or continue with the existing code. But second, I would consider (assuming your use case supports it) implementing the hash with ```bpf_iter_num``` and ```bpf_dynptr_from_mem``` as per the example provided by @dylandreimerink in the Slack thread. It should hopefully be possible to gate this with ```__V61_BPF_PROG``` and then use a flag to indicate whether to use the new or old, with that flag based on availability of these helpers (which could be ascertained in user space and set in a const in the BPF on load, or could potentially be detected with BTF helpers in the BPF program itself).

Happy to discuss these on the Slack thread. Also, I think it is a good idea for you to come to a Tetragon community meeting if you can. @mtardy do you think we could discuss this at the next one to get other people's opinions on this?

I'll endeavour to review quickly when you push an update. Thanks for the work – I think we'll get there – and apologies again for being slow to review.

kevsecurity (changes requested):
As per comment.

PR #4721, opened 2026-03-02
feat: Extract uprobe target names and offsets from stripped go binaries

### LINKED ISSUE #4720: Unable to attach uprobes to stripped go binaries
Uprobes require symbols to attach to a binary. go maintains a Program Counter Line Table that survives stripping. Use that to calculate uprobe target function names and offsets.

dwindsor:
Fixes #4720 
https://github.com/cilium/tetragon-testdata/pull/1
https://github.com/isovalent/cilium-enterprise-demos/pull/210
### Description

Go binaries embed a Program Counter Line Table (.gopclntab) that
the runtime uses for stack traces [1]; this section survives stripping.

Add new SafeELFFile methods to extract uprobe offsets from .gopclntab.
The parsed data is cached on the SafeELFFile for the lifetime of the object.

Since Go 1.17, the compiler emits both ABIInternal and ABI0 pclntab
entries for the same function [2]; the parser deduplicates these by
selecting the largest function body (the real implementation).

Prior to 1.26, the pclntab header embedded a textStart field that
stored the relocated address of runtime.text; parsing it required
version-specific magic number checks. Go 1.26 zeroes this field [3],
so we resolve the text base from the .text section vaddr instead.

CGo binaries are explicitly rejected because the external linker
controls the final layout and the .text vaddr is no longer a reliable
proxy for runtime.text.

Therefore, we restrict this feature to pure Go binaries compiled with
go >= 1.26.

[1] Link: https://groups.google.com/g/golang-dev/c/vuR4CmMp_34/m/NhIsb2IGZGUJ?pli=1
[2] Link: https://go.googlesource.com/proposal/+/master/design/27539-internal-abi.md
[3] Link: https://github.com/golang/go/commit/0e1bd8b5f17e337df0ffb57af03419b96c695fe4

### Results

```yaml
# https://nvd.nist.gov/vuln/detail/CVE-2024-45337
# https://github.com/charmbracelet/soft-serve/security/advisories/GHSA-pchf-49fh-w34r
kind: TracingPolicy
metadata:
  name: "observe-cve-2024-45337"
spec:
  uprobes:
  - path: "/usr/local/bin/soft"
    symbols:
    - "golang.org/x/crypto/ssh.(*connection).serverAuthenticate"
    selectors:
    - matchActions:
      - action: Post
```
```json
{
    "process_uprobe": {
        "process": {
            "exec_id": "Y3ZlLWRlbW8tY29udHJvbC1wbGFuZTozMzQ5NDgxODI1NzkxMDQ6MTg1MDMzNw==",
            "pid": 1850337,
            "uid": 1000,
            "cwd": "/home/git",
            "binary": "/usr/local/bin/soft",
            "arguments": "serve",
            "flags": "execve clone inInitTree",
            "start_time": "2026-03-16T16:33:23.674216738Z",
            "auid": 4294967295,
            "pod": {
                "namespace": "default",
                "name": "soft-serve",
                "uid": "204cabf4-899d-485e-8e30-41d9bbade55e",
                "container": {
                    "id": "containerd://c8a52a2de3652c8aa1b762aab4026e852c603ba1638c38193083a687278c7c67",
                    "name": "soft-serve",
                    "image": {
                        "id": "sha256:03ea700acf322639e7a4cc2b81622da4df4348c8bf61799bfe7e953ec73b85b4",
                        "name": "docker.io/library/soft-serve-vuln:latest"
                    },
                    "start_time": "2026-03-16T16:33:23Z",
                    "pid": 1,
                    "security_context": {}
                },
                "pod_labels": {
                    "app": "soft-serve"
                },
                "workload": "soft-serve",
                "workload_kind": "Pod"
            },
            "docker": "c8a52a2de3652c8aa1b762aab4026e8",
            "parent_exec_id": "Y3ZlLWRlbW8tY29udHJvbC1wbGFuZTozMzQ5NDc5NzMyMzI2NjE6MTg1MDI4Ng==",
            "refcnt": 1,
            "tid": 1850368,
            "in_init_tree": true
        },
        "parent": {
            "exec_id": "Y3ZlLWRlbW8tY29udHJvbC1wbGFuZTozMzQ5NDc5NzMyMzI2NjE6MTg1MDI4Ng==",
            "pid": 1850286,
            "uid": 0,
            "cwd": "/run/containerd/io.containerd.runtime.v2.task/k8s.io/88820bd0ebad3a57198342e9aa203fbbf363081968d3f1fb21b713260002b25a",
            "binary": "/usr/local/bin/containerd-shim-runc-v2",
            "arguments": "-namespace k8s.io -id 88820bd0ebad3a57198342e9aa203fbbf363081968d3f1fb21b713260002b25a -address /run/containerd/containerd.sock",
            "flags": "execve clone",
            "start_time": "2026-03-16T16:33:23.464869664Z",
            "auid": 4294967295,
            "parent_exec_id": "Y3ZlLWRlbW8tY29udHJvbC1wbGFuZTozMzQ5NDc5NzAyODQ1NzE6MTg1MDI3OA==",
            "tid": 1850286,
            "in_init_tree": false
        },
        "path": "/procRoot/89110/root/usr/local/bin/soft",
        "symbol": "golang.org/x/crypto/ssh.(*connection).serverAuthenticate",
        "policy_name": "observe-cve-2024-45337",
        "offset": "3504448",
    },
    "node_name": "cve-demo-control-plane",
    "time": "2026-03-16T16:33:30.151051435Z",
    "node_labels": {
        "beta.kubernetes.io/arch": "amd64",
        "beta.kubernetes.io/os": "linux",
        "kubernetes.io/arch": "amd64",
        "kubernetes.io/hostname": "cve-demo-control-plane",
        "kubernetes.io/os": "linux",
        "node-role.kubernetes.io/control-plane": ""
    }
}
```

### Changelog

```release-note
```

### DIFF
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -340,12 +340,12 @@ func loadMultiUprobeSensor(ids []idtable.EntryID, args sensors.LoadProbeArgs) er
 			attach = &program.MultiUprobeAttachSymbolsCookies{}
 		}
 
-		if uprobeEntry.symbol != "" {
+		if uprobeEntry.address != 0 {
+			attach.Addresses = append(attach.Addresses, uprobeEntry.address)
+		} else if uprobeEntry.symbol != "" {
 			symbol, offset := resolveSymbol(uprobeEntry.symbol)
 			attach.Symbols = append(attach.Symbols, symbol)
 			attach.Offsets = append(attach.Offsets, offset)
-		} else {
-			attach.Addresses = append(attach.Addresses, uprobeEntry.address)
 		}
 
 		if uprobeEntry.refCtrOffset != 0 {
@@ -775,7 +775,31 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 		return nil
 	}
 
-	if symbols != 0 {
+	f, err := elf.OpenSafeELFFile(spec.Path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	if symbols != 0 && f.IsStrippedPureGoBinary() {
+		tbl, err := f.Pclntab()
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse pclntab: %w", err)
+		}
+		for idx, sym := range spec.Symbols {
+			if err := checkSymbol(sym); err != nil {
+				return nil, fmt.Errorf("failed to parse symbol: %w", err)
+			}
+			off, ok := tbl.OffsetByName(sym)
+			if !ok {
+				return nil, fmt.Errorf("failed to resolve symbol: %w", err)
+			}
+			err = addUprobeEntry(sym, off, idx)
+			if err != nil {
+				return nil, err
+			}
+		}
+	} else if symbols != 0 {
 		for idx, sym := range spec.Symbols {
 			if err := checkSymbol(sym); err != nil {
 				return nil, fmt.Errorf("failed to parse symbol: %w", err)
@@ -793,12 +817,6 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 			}
 		}
 	} else if addrs != 0 {
-		f, err := elf.OpenSafeELFFile(spec.Path)
-		if err != nil {
-			return nil, err
-		}
-		defer f.Close()
-
 		for idx, addr := range spec.Addrs {
 			off, err := f.OffsetFromAddr(addr)
 			if err != nil {


pkg/pclntab/pclntab.go
```diff
@@ -0,0 +1,245 @@
...
+	if textStart != 0 {
+		return errors.New("pre-Go 1.26 binary (textStart != 0)")
+	}
+	return nil
+}
+
+// Returns the .text section vaddr for pure-Go binaries
+func resolveTextStart(f *elf.File) (uint64, error) {
+	// ignore CGo binaries
+	if f.Section(".plt") != nil || f.Section(".init") != nil {
```
dwindsor: I'm not actually sure if there's a better way to do this (detecting whether a binary was build with cGO). 
This is just comparing ELF sections in a few known cGO binaries vs known pure Go binaries.

pkg/pclntab/pclntab.go
```diff
@@ -0,0 +1,245 @@
...
+
+// Table containing function offsets and an ABI-deduplicated view of symbols
+type offsetTable struct {
+	funcs  []FuncOffset
+	byName map[string]FuncOffset // deduplicated: largest function per name
+}
+
+var (
+	mu          sync.RWMutex
+	offsetCache = make(map[string]*offsetTable)
```
dwindsor: On second thought I'm not sure we actually need a cache here as we only parse the line table at policy load time. It's an expensive operation though, my thinking is that the cache would be useful if we reload policies. Then again, if the binary in question has been updated, the offsets in `offsetCache` will be stale.

We should probably just remove this. Thoughts?

olsajiri: yep, also it does not seem to have removal logic, so could explode one day

I wonder we could do similar thing that ebpf/cilium is doing with uprobe link cacheAddresses,
and have this functionality under our SafeELFFile (pkg/elf/file.go) or some child of that

this way symbols would be cached only within the life time of its elf object and we could cache them if we think it's a good idea later

pkg/pclntab/pclntab.go
```diff
@@ -0,0 +1,245 @@
...
+)
+
+// Parses .gopclntab for path into an offset table, caching per binary path
+func loadOffsetsFromTable(path string) (*offsetTable, error) {
+	mu.RLock()
+	idx, ok := offsetCache[path]
+	mu.RUnlock()
```
olsajiri: hm, IIUC 2 threads could call this in the same time and if the path is not cached they would race and execute code below

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -776,11 +777,27 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 	}
 
 	if symbols != 0 {
+		pclntabResolved := make(map[string]uint64)
```
olsajiri: instead of allocating allways the map, how about we special case it like:
```
-       if symbols != 0 {
+       if symbols != 0 && pclntab.IsStrippedGoBinary(spec.Path) {
+               pclntabResolved := make(map[string]uint64)
+               if resolved, err := pclntab.Lookup(spec.Path, spec.Symbols); err == nil {

...


+       } else if symbols != 0 {

...
```

pkg/pclntab/pclntab.go
```diff
@@ -0,0 +1,245 @@
...
+	if err != nil {
+		return nil, err
+	}
+	out := make([]FuncOffset, len(ot.funcs))
+	copy(out, ot.funcs)
+	return out, nil
+}
+
+// Resolves the given symbol names to deduplicated FuncOffsets via .gopclntab
+func Lookup(path string, names []string) ([]FuncOffset, error) {
```
olsajiri: perhaps it does not matter much, but I'd do just single symbol resolve not an array

if we go with the symbol cache in elf file, we could just export the ot.byName function below and be done with that.. and the uprobe code (user of this)  iterates the symbols anyway, so maybe we could do something like:

```
    f, err := elf.OpenSafeELFFile(spec.Path) // or overloaded like elf.OpenStrippedGoBinary(spec.Path)
    if err != nil {
            return nil, err
    }
            if err := checkSymbol(sym); err != nil {
                    return nil, fmt.Errorf("failed to parse symbol: %w", err)
            }           
            off, ok := f.OffsetByName(sym)
            if !ok {    
                    return nil, fmt.Errorf("failed to resolve symbol: %w", err)
            }           
            err = addUprobeEntry("", off, idx)
            if err != nil { 
                    return nil, err
            }           
    }
```

pkg/elf/pclntab_test.go
```diff
@@ -0,0 +1,203 @@
+// SPDX-License-Identifier: Apache-2.0
```
olsajiri, line 1: I'd split the code/tests in separate commits, but maybe we decided not to, not sure.. I'll live ;-)

pkg/pclntab/pclntab.go
dwindsor: > I wonder we could do similar thing that ebpf/cilium is doing with uprobe link cacheAddresses,
and have this functionality under our SafeELFFile (pkg/elf/file.go) or some child of that

~~I was thinking of leaving it a separate package for now.~~

Didn't understand what you were originally suggesting =). Moved to `pkg/elf/pclntab.go` and plumbed into `SafeELFFile`. The cache should now be scoped to the lifetime of the `elf.File` and as such won't be in memory long.

Does Cilium need this functionality as well? I see they have uprobes code but am not familiar with the use cases.

pkg/elf/pclntab_test.go
dwindsor: Done =)

pkg/pclntab/pclntab.go
dwindsor: ~~See above comment about Cilium possibly using/needing this code~~

I think I've updated this as you intended. `Lookup()` has been removed and replaced with `OffsetByName()`, which doesn't do a bulk lookup.

New API is:

```go
ef, err := OpenSafeELFFile(path)
if err != nil {
	t.Fatalf("open %s: %v", path, err)
}

tbl, err := ef.Pclntab()
if err != nil {
	return nil, fmt.Errorf("pclntab: %w", err)
}

	if err := checkSymbol(sym); err != nil {
		return nil, fmt.Errorf("failed to parse symbol: %w", err)
	}
	if off, ok := tbl.OffsetByName(sym); ok {
		err = addUprobeEntry("", off, idx)
	} else {
		err = addUprobeEntry(sym, 0, idx)
	}
	if err != nil {
		return nil, err
	}
}
```

pkg/sensors/tracing/genericuprobe.go
dwindsor: Makes sense, updated, thanks

pkg/pclntab/pclntab.go
dwindsor: The offset cache has been moved into `pkg/elf `and access is serialized by [`sync.Once`](https://github.com/dwindsor/tetragon/blob/f5debe59ec442c0b7728c58baaf30a4464d91130/pkg/elf/pclntab.go#L87)

pkg/pclntab/pclntab.go
dwindsor: edit: I think actually this check for `.plt `and `.init` is sufficient -- it tells us that an external linker was used. `cmd/link` (used in pure Go) won't insert these sections

pkg/elf/pclntab_test.go
```diff
@@ -0,0 +1,225 @@
...
+// Copyright Authors of Tetragon
+
+//go:build linux
```
dwindsor, line 4: Without this, Windows CI will fail (one of the tests uses /bin/sh)

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -775,7 +775,31 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 		return nil
 	}
 
-	if symbols != 0 {
+	ef, elfErr := elf.OpenSafeELFFile(spec.Path)
```
olsajiri: lets use `err` , it's all over the place in this function already
also I think using `file` or just `f` for file is bit better than `ef`, like we did below

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -775,7 +775,31 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+		}
+
+		for idx, sym := range spec.Symbols {
+			if err := checkSymbol(sym); err != nil {
+				return nil, fmt.Errorf("failed to parse symbol: %w", err)
+			}
+			if off, ok := tbl.OffsetByName(sym); ok {
+				err = addUprobeEntry("", off, idx)
+			} else {
+				err = addUprobeEntry(sym, 0, idx)
```
olsajiri: is there any point in doing the fallback with symbol if we failed to resolve with specialized function?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -775,7 +775,31 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+	ef, elfErr := elf.OpenSafeELFFile(spec.Path)
+	if elfErr == nil {
+		defer ef.Close()
+	}
+
+	if symbols != 0 && elfErr == nil && ef.IsStrippedPureGoBinary() {
+		tbl, err := ef.Pclntab()
+		if err != nil {
+			return nil, fmt.Errorf("pclntab: %w", err)
+		}
```
olsajiri, line 788: could all this be hidden under `ef.Offset(symbol)` ? like we would check if this is stripped go binary in `OpenSafeELFFile` and then if Offset is called we would do all the magic for the symbol resolve?
and same for `ef.Symbols` for the policy generation path

pkg/elf/testdata/Makefile
```diff
@@ -0,0 +1,15 @@
+TESTDATA_URL = https://github.com/cilium/tetragon-testdata/raw/main/pclntab
+
+all: node_exporter-stripped node_exporter-unstripped node_exporter-go122-stripped
```
olsajiri: this gives me:

```
jolsa@ubuntu-24:~/tetragon/pkg/elf/testdata$ make
wget -nv -O node_exporter-stripped.gz 'https://github.com/cilium/tetragon-testdata/raw/main/pclntab/node_exporter-stripped.gz'
https://github.com/cilium/tetragon-testdata/raw/main/pclntab/node_exporter-stripped.gz:
2026-03-11 11:50:30 ERROR 404: Not Found.
make: *** [Makefile:6: node_exporter-stripped.gz] Error 8
rm node_exporter-stripped.gz
```

is it that big? could we make somehow have some small test binary in `contrib/tester-progs`

pkg/elf/pclntab_test.go
```diff
@@ -0,0 +1,236 @@
...
+	t.Helper()
+	f, err := OpenSafeELFFile(path)
+	if err != nil {
+		t.Fatalf("open %s: %v", path, err)
+	}
+	t.Cleanup(func() { f.Close() })
+	return f
+}
+
+func TestIsStrippedPureGoBinary(t *testing.T) {
```
olsajiri, line 40: it'd be great to have also test for uprobe on top of the stripped go binary

pkg/elf/testdata/Makefile
dwindsor: If you're okay with me making crafted binaries for testing, definitely we should do this. The reason I chose these binaries is because they're from the wild, so works well in a PoC demo.

Perhaps I create small test binaries in `contrib` and keep the production binaries in another demos repository.

pkg/sensors/tracing/genericuprobe.go
dwindsor: > could all this be hidden under ef.Offset(symbol) ?

Not easily, moving the entire thing into Offset is a bit awkward in terms of error handling and flow. It ends up being awkward because we're adding another clause to a large if statement and error handling is non-obvious.

> like we would check if this is stripped go binary in OpenSafeELFFile and then if Offset is called we would do all the magic for the symbol resolve?

We'd have to call `IsStrippedPureGoBinary` again in Offset I think, because before reading the line table we need to know it's pure go (if we try reading .text address from a cGO binary things will break). 

I'm going to leave it the way it is now and think about it some more. I don't think it's that bad now, which is the first version you suggested.

pkg/elf/testdata/Makefile
dwindsor: Removed the testdata harness and now build test binaries in CI, so no binary checkins needed, excepting `pclntab-go122-stripped` which cannot easily be built in CI and isn't executed, only parsed

pkg/elf/pclntab_test.go
dwindsor: Added this test to policytests/uprobes.go. It attaches a uprobe and observes main.main being called:

```
jsonTestCheck/line:0004 event:PROCESS_UPROBE =>  FINAL MATCH
```

dwindsor:
CI failing on 4.19 and 5.4, will investigate

dwindsor:
> CI failing on 4.19 and 5.4, will investigate

Fixed: "." was an illegal character in some bpffs nodes in 5.4 and 4.19, so sanitize the filename to convert "." -> "_".

Windows CI failing, not sure if it's related:

```
✓ Tetragon is Running
Process launched with PID: 8832
Looking for regex: \{\"process_exec\"\:\{\"process\"\:\{\"exec_id\"\:\".{16,30}\"\,.{0,1}\"pid\"\:8832\,.{0,1}\"uid\"\:[0-9]{0,9}\,.{0,1}\"binary\"\:\"C:\\\\Windows\\\\system32\\\\notepad.exe\"
Get-Content: D:\a\_temp\f097882b-9bdf-412c-a535-e6f7eee18cff.ps1:39
Line |
```

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -776,23 +779,47 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
 		for idx, sym := range spec.Symbols {
+			// Is this a valid symbol name?
 			if err := checkSymbol(sym); err != nil {
 				return nil, fmt.Errorf("failed to parse symbol: %w", err)
 			}
-			err = addUprobeEntry(sym, 0, idx)
-			if err != nil {
+
+			off := uint64(0)
+			if err == nil {
```
olsajiri: hm, what's the check for?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -950,11 +977,12 @@ func createUprobeSensorFromEntry(polInfo *policyInfo, uprobeEntry *genericUprobe
 
 	loadProgName, loadProgRetName := config.GenericUprobeObjs(false)
 
+	pinSymbol := strings.ReplaceAll(uprobeEntry.symbol, ".", "_")
```
olsajiri, line 971: please put this in separate commit

pkg/sensors/tracing/genericuprobe.go
olsajiri: fair enough, but I'd like still not to mix this case with the generic one.. could we do this instead? https://github.com/cilium/tetragon/commit/1eb429ade6b2969c2b3d58dd5e21a615262981a9

pkg/sensors/tracing/genericuprobe.go
dwindsor: done

pkg/sensors/tracing/genericuprobe.go
dwindsor: It was left over from a cherrypick, removed

pkg/sensors/tracing/genericuprobe.go
dwindsor: applied, thanks

pkg/elf/pclntab.go
```diff
@@ -0,0 +1,218 @@
...
+			End:    pclntabVAToFileOffset(se.File, fn.End),
+		})
+	}
+	sort.Slice(funcs, func(i, j int) bool {
+		return funcs[i].Offset < funcs[j].Offset
+	})
+
+	// When Go emits multiple pclntab entries for the same name
+	// (ABIInternal + ABI0 wrapper), keep the largest function
+	// which is he real implementation, not the wrapper
```
mtardy, line 111: small typo on the "the real implem"

contrib/tester-progs/pclntab-go122-stripped
mtardy, line 1: ah please do not commit any binaries to the repo, compile them as part of your tests like we do for the other contrib/tester-progs.

.github/workflows/static-checks.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
         run: |
           echo "Checking for any checked in binary files not in allowlist..." 
           outfile="$(mktemp)"
-          find . -type f -size +0 -not -wholename '**/vendor/**' -not -wholename '**/_vendor/**' -not -wholename '**/.git/**' -not -name '*.png' -not -name '*.jpg' -not -name '*.ico' | xargs -n 100 grep -IL '' | tee "$outfile"
+          find . -type f -size +0 -not -wholename '**/vendor/**' -not -wholename '**/_vendor/**' -not -wholename '**/.git/**' -not -name '*.png' -not -name '*.jpg' -not -name '*.ico' -not -wholename '**/contrib/tester-progs/**' | xargs -n 100 grep -IL '' | tee "$outfile"
```
mtardy: this check is often really useful on avoiding committing binaries from the contrib tester progs, please revert that change

cmd/tetra/tracingpolicy/generate/generate.go
```diff
@@ -169,6 +169,24 @@ func New() *cobra.Command {
 				uprobe.Symbols = append(uprobe.Symbols, sym.Name)
 			}
 
+			// Stripped Go binary with no symtab: fall back to .gopclntab
+			if len(uprobe.Symbols) == 0 {
+				file, err := telf.OpenSafeELFFile(uprobesBinary)
```
mtardy, line 174: just double checking that you don't want to do anything on `err != nil` from here?

contrib/tester-progs/pclntab-go122-stripped
dwindsor: Yeah, I didn't really want to do this but thought it might be acceptable in this one case. We need this binary to be compiled with a specific version of go, so building it in CI would require either bundling that version of go, or perhaps running a separate build in a container. 

Thoughts on this? Not really sure myself 🤔

mtardy: ah I see, given the test just checks that the thing doesn't work and you have a test in which the thing should work I'd just remove it maybe? I see the idea of checking the error path as well but maybe it's not worth bothering in this case.

```go
func TestPclntabRejectPreGo126(t *testing.T) {
	preGo126 := testutils.RepoRootPath("contrib/tester-progs/pclntab-go122-stripped")
	if _, err := os.Stat(preGo126); err != nil {
		t.Skipf("pclntab-go122-stripped not found: %v", err)
	}
	f := openSafe(t, preGo126)
	_, err := f.Pclntab()
	if err == nil {
		t.Fatal("expected error for pre-Go 1.26 binary")
	}
	t.Logf("correctly rejected: %v", err)
}
```

dwindsor: I'm fine to remove it, but the thing it checks for is actually pretty critical. If we try attaching to a pre-1.26 go binary, there's a chance that the offsets we read from `.pclntab` aren't right. If we try attaching to that address it wont work. So we check that the binary gets explicitly rejected here.

Would it be okay to check in this one binary, but do it in a subdirectory of `contrib/tester-progs` so we don't have to allowlist the whole thing?

mtardy: Is what you are trying to test against is someone modifying your function and the function not to fail anymore and then the thing to be broken?

okay so could you extract maybe exactly the bits in the binary part that you need for your function to fail and just feed it from the Go test file statically? Do you need a full ELF of almost 1MB?

mtardy: or emulate what would happen with an old binary from the test perspective so that it fails if someone changes the function? Since you are committing something static anyway that would be similar.

dwindsor: > Is what you are trying to test against is someone modifying your function and the function not to fail anymore and then the thing to be broken?

Nah it’s more about testing the case that a customer tries to write a TracingPolicy against a binary they think is go 1.26+ but actually isn’t. 

Thinking about this more, i think you’re right that we can just remove the check. pclntab() will just return err and the offsets aren’t passed to cilium/ebpf for attachment. 👍

dwindsor: > okay so could you extract maybe exactly the bits in the binary part that you need for your function to fail and just feed it from the Go test file statically? Do you need a full ELF of almost 1MB?

This is interesting, if you think the test is still worth it, i can do something like this (hard-code the few bits of pclntab i’m interested in)

dwindsor: > Thinking about this more, i think you’re right that we can just remove the check. pclntab() will just return err and the offsets aren’t passed to cilium/ebpf for attachment. 👍

I removed the test

.github/workflows/static-checks.yaml
dwindsor: reverted

cmd/tetra/tracingpolicy/generate/generate.go
dwindsor: Good catch -- yes we want to handle that. It wasn't fatal before because we end up passing the  target binary to `cilium/ebpf` anyway for further handling, at which point it would fail.

PR #4728, opened 2026-03-05
cel: add uprobe support and make filter not match if arg could not be read

andrewstrohman:
This mimics the logic that we have for other selector filter types where the filter does not match when its corresponding args could not be read.

When CEL expressions are used, the args that they reference need to be declared in the args section of the corresponding filter. Relay this information to the BPF program, so that it can confirm the args were successfully read before allowing a match.

The most flexible way to test this is to use uprobes. So, I added uprobe support for CEL in order to test that we won't match on a CEL expression when its corresponding arg could not be read (due to NULL pointers).

Since I added CEL uprobe support, I added a test to confirm the multi-uprobe works properly with CEL. Since we want to use policytests going forward, I extend the policytest framework in order to support running multiple scenarios (from legacy, observer tests) and doing negative tests.

Here is how I tested the policytests change to support expecting no match against the eventchecker:
```
root@kind-6:/host/build/src/tetragon# ./tetra policytest run cel-multi-uprobe-one-match
P: cel-multi-uprobe-one-match                         ✅   2/2 scenario(s) succeeded
├S: check uprobe_test_lib_arg2 event                  🟢
└S: check uprobe_test_lib_arg1 event does not occur   🟢
root@kind-6:/host/build/src/tetragon# ./tetra policytest run cel-multi-uprobe-both-match
P: cel-multi-uprobe-both-match   ✅   1/1 scenario(s) succeeded
└S: check both events occur      🟢
```

The other policytest change, where we can run multiple scenarios from a legacy, observer unit test, is used by the `TestCelExprMultiUprobeOneMatch` unit test.

### DIFF
--- a/tests/policytests/uprobes.go
+++ b/tests/policytests/uprobes.go
@@ -78,3 +78,119 @@ spec:
 		},
 	}
 }).RegisterAtInit()
+
+var _ = policytest.NewBuilder("cel-multi-uprobe-one-match").WithLabels("uprobes").WithPolicyTemplate(`
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "cel-multi-uprobe"
+spec:
+  uprobes:
+  - path: {{ testBinary "libuprobe.so" }}
+    symbols:
+    - "uprobe_test_lib_arg1"
+    selectors:
+    - matchArgs:
+      - operator: CelExpr
+        values:
+        - "false"
+  - path: {{ testBinary "libuprobe.so" }}
+    symbols:
+    - "uprobe_test_lib_arg2"
+    selectors:
+    - matchArgs:
+      - operator: CelExpr
+        values:
+        - "true"
+`).WithSkip(func(si *policytest.SkipInfo) string {
+	if !si.AgentInfo.Probes[bpf.LargeProgsProbe] {
+		return "need 5.3 or newer kernel"
+	}
+
+	if !si.AgentInfo.Probes[bpf.UprobeRefCtrOffsetProbe] {
+		return "need uprobe ref_ctr_off support"
+	}
+
+	if !si.AgentInfo.Probes[bpf.MixBPFAndTailCallsProbe] {
+		return "need kernel where we can mix bpf and tail calls"
+	}
+
+	return ""
+}).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
+	myBin := c.TestBinary("uprobe-test-1")
+	upChecker := ec.NewProcessUprobeChecker("cel-multi-uprobe").
+		WithProcess(ec.NewProcessChecker().
+			WithBinary(sm.Full(myBin))).WithSymbol(sm.Full("uprobe_test_lib_arg2"))
+
+	return &policytest.Scenario{
+		Name:         "check uprobe_test_lib_arg2 event",
+		Trigger:      policytest.NewCmdTrigger(myBin).ExpectExitCode(0),
+		EventChecker: ec.NewUnorderedEventChecker(upChecker),
+	}
+}).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
+	myBin := c.TestBinary("uprobe-test-1")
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
LGTM!

bpf/process/types/basic.h
```diff
@@ -2206,6 +2206,23 @@ selector_arg_offset(void *ctx, struct bpf_map_def *tailcalls,
 		 * to a cel expression identifier (rather than an argument)
 		 */
 		if (filter->op == op_cel_expr) {
+			/* if the cel expression references an arg that could not be read,
+			 * the selector does not match
+			 */
+			__u32 *v = (__u32 *)&filter->value;
+			int j = 0;
+
+			for (int k = 0; k < MAX_POSSIBLE_ARGS; k++) {
```
olsajiri: nit, maybe we could add int j in the for loop as well to show it belongs there

`for (int k = 0, j = 0; k < MAX_POSSIBLE_ARGS; k++) {
`

kkourt:
Conflicts caused by me merging https://github.com/cilium/tetragon/pull/4717. Happy to help resolve conflict if it ends up being tricky,

bpf/process/types/basic.h
andrewstrohman: I've made this change. Thanks.

andrewstrohman:
> Conflicts caused by me merging #4717. Happy to help resolve conflict if it ends up being tricky,

Thanks for the heads up. I've adjusted this change to support multi-uprobe as well.

andrewstrohman:
I've converted this to a draft because I'm working on adding a policy test for the multi-uprobe case.

andrewstrohman:
@kkourt I added multi-uprobe support in order to resolve the rebase conflict. I wanted to add a test for this, and I know that we want to use policytests going forward, so I extended the policytests framework a little in order to get the test functionality that I wanted.

Since you are SME for all these things, I'm requesting that you please review again.

Thanks!

tests/policytests/uprobes.go
```diff
@@ -78,3 +80,111 @@ spec:
...
+        values:
+        - "true"
+`).WithSkip(func(_ *policytest.SkipInfo) string {
+	if !config.EnableLargeProgs() || !bpf.HasUprobeRefCtrOffset() {
+		return "Need 5.3 or newer kernel for uprobe ref_ctr_off support for this test"
+	}
+
+	if !celbpf.Supported() {
+		return "Need CEL BPF support for this test"
+	}
```
kkourt, line 178: The way you check skip functionality here requires being able to load BPF programs. I think it's better if we have the skip function operation only on its input.

Can you change it to something like:
```diff
diff --git a/tests/policytests/uprobes.go b/tests/policytests/uprobes.go
index 9ec0f7936..822402f4f 100644
--- a/tests/policytests/uprobes.go
+++ b/tests/policytests/uprobes.go
@@ -104,13 +104,17 @@ spec:
       - operator: CelExpr
         values:
         - "true"
-`).WithSkip(func(_ *policytest.SkipInfo) string {
-       if !config.EnableLargeProgs() || !bpf.HasUprobeRefCtrOffset() {
-               return "Need 5.3 or newer kernel for uprobe ref_ctr_off support for this test"
+`).WithSkip(func(si *policytest.SkipInfo) string {
+       if !si.AgentInfo.Probes[bpf.LargeProgsProbe] {
+               return "Need 5.3 or newer kernel"
        }

-       if !celbpf.Supported() {
-               return "Need CEL BPF support for this test"
+       if !si.AgentInfo.Probes[bpf.UprobeRefCtrOffsetProbe] {
+               return "need uprobe ref_ctr_off support"
+       }
+
+       if !si.AgentInfo.Proibes[bpf.MixBPFAndTailCallsProbe] {
+               return "need kernel where we can mix bpf and tail calls"
        }
```

This way, we don't need CAP_BPF or require to execute the skip function on the same node as where the agent runs.

andrewstrohman: Thank for catching this! Makes sense. I've updated both tests to be like this.

PR #4737, opened 2026-03-06
Makefile: don't rely on PWD

andrewstrohman:
When renovate runs, this env variable is not present. So, use CURDIR instead, which is Make's internal variable.


mtardy:
Looks like @dangome3 went to the same conclusion https://github.com/cilium/tetragon/pull/4778. Merging this since it's older.

PR #4742, opened 2026-03-07
tetragon: tracing tests cleanup

olsajiri:
small tracing tests cleanup before bigger fentry changes

### DIFF
--- a/pkg/sensors/sensors.go
+++ b/pkg/sensors/sensors.go
@@ -237,24 +237,6 @@ type LoadProbeArgs struct {
 	Version, Verbose int
 }
 
-func GetMergedSensorFromParserPolicy(tp tracingpolicy.TracingPolicy) (SensorIface, error) {
-	// NB: use a filter id of 0, so no filtering will happen
-	sis, err := SensorsFromPolicy(tp, policyfilter.NoFilterID)
-	if err != nil {
-		return nil, err
-	}
-	sensors := make([]*Sensor, 0, len(sis))
-	for _, si := range sis {
-		s, ok := si.(*Sensor)
-		if !ok {
-			return nil, fmt.Errorf("cannot merge sensor of type %T", si)
-		}
-		sensors = append(sensors, s)
-	}
-
-	return SensorCombine(tp, tp.TpName(), sensors...), nil
-}
-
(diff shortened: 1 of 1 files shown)


pkg/sensors/tracing/kprobe_test.go
```diff
@@ -4556,15 +4503,23 @@ spec:
...
+	}
+	s := sens[0]
+
+	err = s.Load(option.Config.BpfDir)
 	if err != nil {
 		return err
 	}
-	return sens.Unload(true)
+	s.Destroy(true)
+	return nil
```
FedeDP: The only difference here is that we returned the eventual `sens.Unload` error before, while we now just log a message (given `Destroy` impl).
Is it ok?

pkg/sensors/tracing/enforcer_test.go
```diff
@@ -68,10 +68,7 @@ func testEnforcer(t *testing.T, configHook string,
 	ctx, cancel := context.WithTimeout(context.Background(), tus.Conf().CmdWaitTime)
 	defer cancel()
 
-	err := os.WriteFile(testConfigFile, []byte(configHook), 0644)
-	if err != nil {
-		t.Fatalf("writeFile(%s): err %s", testConfigFile, err)
-	}
+	createCrdFile(t, configHook)
```
FedeDP, line 71: Wondering whether it would be better to call it `mustCreateCrdFile` since we fail out (`t.Fatal`) error.
Anyway, i see it is widely used thus a name change is a bit too much :)

pkg/sensors/tracing/kprobe_test.go
olsajiri: yea, it's unfortunate the sensor::Destroy does not return error, maybe we can fix that, will check, thnx

FedeDP: Not a big deal IMO.

olsajiri: added the change, I tihnk it's better, plz check @FedeDP  thnx

FedeDP: LGTM, thanks for dealing with it :)

PR #4757, opened 2026-03-12
metrics: consolidate kprobe and uprobe merge logic and metrics

AritraDey-Dev:
Part of #2785 
> Consider merging kprobe ok and errors metrics into one with status label

### Description

This PR unify `kprobe` and `uprobe` merge logic and metrics, including a fix for a nil `dereference` in the merge path.

Example prometheus output with this patch:
```
tetragon_generic_kprobe_merge_total{curr_fn="vfs_read",curr_type="exit",prev_fn="vfs_read",prev_type="enter",status="ok"} 142
tetragon_generic_kprobe_merge_total{curr_fn="vfs_read",curr_type="enter",prev_fn="vfs_read",prev_type="exit",status="ok"} 12
tetragon_generic_kprobe_merge_total{curr_fn="vfs_read",curr_type="exit",prev_fn="vfs_read",prev_type="exit",status="error"} 2
```

/cc @mtardy 

### Changelog

```release-note
consolidate kprobe and uprobe merge logic and metrics into a unified format.
```

### DIFF
--- a/contrib/upgrade-notes/latest.md
+++ b/contrib/upgrade-notes/latest.md
@@ -28,4 +28,9 @@ See the [Stack Traces](https://tetragon.io/docs/concepts/tracing-policy/selector
 
 ### Metrics
 
-* TBD
+* Kprobe and uprobe merge metrics have been consolidated.
+  The following metrics were removed:
+  - `tetragon_generic_kprobe_merge_errors_total`
+  - `tetragon_generic_kprobe_merge_ok_total`
+  They are replaced by `tetragon_generic_kprobe_merge_total` which includes a `status` label with values `ok` or `error`.
+  The new metric also includes labels `curr_type`, `prev_type` (either `enter` or `exit`), `curr_fn`, and `prev_fn`.


pkg/sensors/tracing/generickprobe.go
```diff
@@ -1399,44 +1393,34 @@ func handleMsgGenericKprobe(m *api.MsgGenericKprobe, gk *genericKprobe, r *bytes
...
 }
 
-type reportMergeErorrFn[T evArgsRetriever] func(curr pendingEvent[T], prev pendingEvent[T])
+type reportMergeErrorFn[T evArgsRetriever] func(curr pendingEvent[T], prev pendingEvent[T])
 
 type evArgsRetriever interface {
 	GetArgs() *[]api.MsgGenericKprobeArg
-	// This constraint allows us to return nil from methods
-	*tracing.MsgGenericKprobeUnix | *tracing.MsgGenericUprobeUnix
+	GetName() string
```
FedeDP: I'd rename the interface to `evtNameArgsRetriever` now :)

AritraDey-Dev:
> We might want to move all the retprobe merge stuff (interfaces, structures and helpers) to their own source file

yes, definitely better to have a separate file for the retprobe logic. I’ll add those changes to this PR itself. thanks

pkg/sensors/tracing/generickprobe.go
```diff
@@ -137,10 +131,6 @@ type genericKprobe struct {
 // This is needed for retprobe probes that generate two events: one at the
 // function entry, and one at the function return. We merge these events into
 // one, before returning it to the user.
```
mtardy: ah don't forget to remove that comment along

pkg/metrics/kprobemetrics/kprobemetrics.go
```diff
@@ -8,37 +8,67 @@ import (
 	"slices"
 
 	"github.com/prometheus/client_golang/prometheus"
+	dto "github.com/prometheus/client_model/go"
```
mtardy, line 11: what does dto stand for?

pkg/metrics/kprobemetrics/kprobemetrics.go
```diff
@@ -8,37 +8,67 @@ import (
...
-var mergeErrorTypeLabelValues = map[MergeErrorType]string{
-	MergeErrorTypeEnter: "enter",
-	MergeErrorTypeExit:  "exit",
+// Legacy alias for compatibility during refactor
+type MergeErrorType = MergeProbeType
+
+const (
+	MergeErrorTypeEnter = MergeProbeTypeEnter
+	MergeErrorTypeExit  = MergeProbeTypeExit
+)
```
mtardy: is there a way to not add new stuff during that refactor? It makes the patch a bit difficult to follow imo

pkg/grpc/tracing/tracing.go
```diff
@@ -740,6 +740,10 @@ func (msg *MsgGenericKprobeUnix) GetArgs() *[]tracingapi.MsgGenericKprobeArg {
 	return &msg.Args
 }
 
+func (msg *MsgGenericKprobeUnix) GetName() string {
```
mtardy, line 743: nice!

pkg/metrics/kprobemetrics/kprobemetrics.go
AritraDey-Dev: yeah, I kept the aliases on purpose so each commit stays focused on one thing, commit 2 only touches the metrics layer, commit 4 handles the callers. Mixing both in one commit would make the diff harder to review. The aliases are gone by the end of the series so the final state is clean either way.

pkg/metrics/kprobemetrics/kprobemetrics.go
AritraDey-Dev: data transfer object
https://github.com/cilium/tetragon/blob/34e9c4ace3003b120760bd54ceb5ac8890f775c8/pkg/bench/summary.go#L19

AritraDey-Dev:
>Also do you think it would be possible to add tests for the report*Merge functions?

Added in generickprobe_test.go,i think that is the right place.

pkg/metrics/kprobemetrics/kprobemetrics.go
```diff
@@ -54,44 +84,101 @@ var (
...
+	}
+	prevType := MergeProbeTypeEnter
+	if prevIsReturn {
+		prevType = MergeProbeTypeExit
+	}
+	MergeErrorsInc(currFn, prevFn, currType, prevType)
+}
+
+// TotalMergeStatus returns the total count for a given status
+func TotalMergeStatus(status string) uint64 {
```
mtardy: nit: only thing that looks a bit weird to me is that you wrote dedicated functions so that users of this don't have to set the status string by themselves because they call the "MergeError" or "MergeOk" function which are setting that variable, but for the `TotalMergeStatus` you need to give the string in freeform like here https://github.com/cilium/tetragon/pull/4757/changes/979efc8963d40efabf806f0545542912fee71dfa#diff-b5ee06defc106b7c31fe9a4cff34d7e2654a9e8cd0321f38dc2059adb0f85590R88-R89.

Maybe you can write two small wrapper around private `totalMergeStatus` like `TotalMergeOk` and `TotalMergeError`? And also maybe put the strings "ok", "error" into private consts here.

By the way all this code is so complicated for just two counters but since we are already there now.

mtardy:
> > Also do you think it would be possible to add tests for the report*Merge functions?
> 
> Added in generickprobe_test.go,i think that is the right place.

If this is just unit testing functions from `kprobemetrics.go`, I'll put it along in `kprobemetrics_test.go`. Btw, it's my fault, but now seeing it, I'm not sure that test is that useful in the end but always nice to have it, thanks.

AritraDey-Dev:
>Maybe you can write two small wrapper around private totalMergeStatus like TotalMergeOk and TotalMergeError? And also maybe put the strings "ok", "error" into private consts here.

yes agree,it looks like good approach,will do that.

AritraDey-Dev:
>If this is just unit testing functions from kprobemetrics.go, I'll put it along in kprobemetrics_test.go. Btw, it's my fault, but now seeing it, I'm not sure that test is that useful in the end but always nice to have it, thanks.

Not sure if we should add a separate file for testing this. Since there aren’t any dedicated tests for the kprobemetrics part, I think we can skip adding tests here as well. wdyt? We could drop the test commit?

mtardy:
> > If this is just unit testing functions from kprobemetrics.go, I'll put it along in kprobemetrics_test.go. Btw, it's my fault, but now seeing it, I'm not sure that test is that useful in the end but always nice to have it, thanks.
> 
> Not sure if we should add a separate file for testing this. Since there aren’t any dedicated tests for the kprobemetrics part, I think we can skip adding tests here as well. wdyt? We could drop the test commit?

yeah we can drop that commit I think it's fine

contrib/upgrade-notes/latest.md
```diff
@@ -28,4 +28,10 @@ See the [Stack Traces](https://tetragon.io/docs/concepts/tracing-policy/selector
...
 ### Metrics
 
-* TBD
+* Kprobe and uprobe merge metrics have been consolidated.
+  The following metrics were removed:
+  - `tetragon_generic_kprobe_merge_errors_total`
+  - `tetragon_generic_kprobe_merge_ok_total`
+  They are replaced by `tetragon_generic_kprobe_merge_total` which includes a `status` label with values `ok` or `error`.
+  The new metric also includes labels `curr_type`, `prev_type` (either `enter` or `exit`), `curr_fn`, and `prev_fn`.
+* A new metric `tetragon_generic_kprobe_merge_pushed_total` was added to track events pushed to the internal merge map.
```
mtardy: I wanted to merge this but sorry just noticed this, why do you say `tetragon_generic_kprobe_merge_pushed_total` is a new metric? It seems it already exists to me and you didn't touch it right?

AritraDey-Dev: Sorry about that! I must have mixed it up while writing the notes,since we were touching all the merge metrics I assumed it was new too, but didn't double check. Fixed now!

PR #4762, opened 2026-03-13
add set action stats

kkourt:
Fixes:
 * add the policy_conf and policy_stats for USDTs
 * add support for Set / MonitorSet action acounters

### DIFF
--- a/pkg/testutils/policytest/observertest.go
+++ b/pkg/testutils/policytest/observertest.go
@@ -17,7 +17,14 @@ import (
 	"github.com/cilium/tetragon/pkg/observer/observertesthelper"
 	"github.com/cilium/tetragon/pkg/testutils"
 	tus "github.com/cilium/tetragon/pkg/testutils/sensors"
+	"github.com/cilium/tetragon/pkg/tetragoninfo"
 )
+
+var getSkipInfo = sync.OnceValue(func() *SkipInfo {
+	res := tetragoninfo.Gather()
+	info := tetragoninfo.Decode(res)
+	return &SkipInfo{info}
+})
 
 func (rpt *RegisteredPolicyTests) DoObserverTest(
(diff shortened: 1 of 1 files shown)


bpf/process/generic_calls.h
```diff
@@ -1022,7 +1022,12 @@ do_action(void *ctx, __u32 i, struct selector_action *actions, bool *post, bool
 	case ACTION_SET:
 		index = actions->act[++i];
 		value = actions->act[++i];
-		do_set_action(ctx, e, index, value);
+		if (enforce_mode) {
```
FedeDP, line 1025: This is a fix right? I mean, skipping the `do_set_action` when `!enforce_mode`.

FedeDP: If yes, we might want to move it to its own commit?

kkourt: Are you saying that we split this commit into to parts:
* respect the enforce mode
* Set polacct so that the counters are increased?

I feel both are bugfixes, so my take is to keep it as a single commit.

FedeDP: Ah as far as i got it, the 
> respect the enforce mode

is a fix on a previous bug; while `polacct`  change is related to this PR. Isn't it?

kkourt: I consider the fact that we were keeping no action counters for certain actions to be a bug. Specifically, NENEFORCE actions in `tetra tp list` would be zero even if we had a `Set` action, which is enforcement.

FedeDP: Makes sense then! Thanks

kkourt:
> could you please add missing changelogs? otherwise lgtm

Added changelogs. Merging. Thanks!

PR #4763, opened 2026-03-14
eventchecker: support generic map types

AritraDey-Dev:
### Description

This PR extends the `eventchecker` to support generic map fields. Previously,PR #260 hardcoded support only for `map[string]string`. 

This PR introduces a generic `MapMatcher` to automatically handle different key-value types (like int and bool).

**Note**: `MapMatcher` is made generic so future proto map fields with non-string value types (e.g. `map<string, uint32>`) are automatically supported by codegen without additional matcher changes.

### Changelog

```release-note
eventchecker: support generic map types
```

### DIFF
--- a/api/vendor/github.com/cilium/tetragon/pkg/matchers/mapmatcher/mapmatcher.go
+++ b/api/vendor/github.com/cilium/tetragon/pkg/matchers/mapmatcher/mapmatcher.go
@@ -1,0 +1,109 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package mapmatcher
+
+import (
+	"fmt"
+
+	"sigs.k8s.io/yaml"
+)
+
+// Matcher is a generic interface for matching values of type T.
+type Matcher[T any] interface {
+	Match(T) error
+}
+
+// MapMatcher matches a map of type map[K]V using matchers for each value.
+// K is the key type, V is the value type, and M is the matcher type for V.
+type MapMatcher[K comparable, V any, M any] map[K]M
+
+// Match matches the actual map against the expected matchers.
+// It iterates over all expected matchers and ensures they match the values in the actual map.
+func (m MapMatcher[K, V, M]) Match(actual map[K]V) error {
+	var unmatched []K
+
+	for key, matcher := range m {
+		val, ok := actual[key]
+		if !ok {
+			unmatched = append(unmatched, key)
+			continue
+		}
+
+		// Try to match using the value or its pointer. This allows MapMatcher to work
+		// with matchers that have pointer receivers even when stored as values in the map.
+		var err error
+		if im, ok := any(matcher).(interface{ Match(V) error }); ok {
+			err = im.Match(val)
+		} else if im, ok := any(&matcher).(interface{ Match(V) error }); ok {
+			err = im.Match(val)
+		} else {
+			return fmt.Errorf("key %v: matcher of type %T does not implement Match(%T) error", key, matcher, val)
+		}
+
+		if err != nil {
+			return fmt.Errorf("key %v match failed: %w", key, err)
+		}
+	}
+
+	if len(unmatched) > 0 {
+		return fmt.Errorf("unmatched keys: %v", unmatched)
+	}
+
+	return nil
+}
+
+// UnmarshalJSON implements json.Unmarshaler for MapMatcher.
+func (m *MapMatcher[K, V, M]) UnmarshalJSON(b []byte) error {
+	type Alias MapMatcher[K, V, M]
+	var alias Alias
+	if err := yaml.UnmarshalStrict(b, &alias); err != nil {
+		return fmt.Errorf("unmarshal MapMatcher: %w", err)
+	}
+	*m = MapMatcher[K, V, M](alias)
+	return nil
+}
+
+// PrimitiveMatcher is a generic matcher for primitive types (string, int, bool, etc.)
+// that performs a simple equality check.
+type PrimitiveMatcher[V comparable] struct {
+	Value V `json:"value"`
+}
+
+// Match implements the Matcher interface for PrimitiveMatcher.
+func (m PrimitiveMatcher[V]) Match(actual V) error {
+	if m.Value != actual {
+		return fmt.Errorf("expected %v, got %v", m.Value, actual)
+	}
+	return nil
+}
+
+// UnmarshalJSON implements json.Unmarshaler for PrimitiveMatcher.
+// It allows the user to provide a plain value as a shorthand for the matcher.
+func (m *PrimitiveMatcher[V]) UnmarshalJSON(b []byte) error {
+	// Try to unmarshal as a plain value first
+	var val V
+	if err := yaml.UnmarshalStrict(b, &val); err == nil {
+		m.Value = val
+		return nil
+	}
+
+	// Otherwise, unmarshal as a struct
+	type Alias PrimitiveMatcher[V]
+	var alias Alias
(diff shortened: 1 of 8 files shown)


andrewstrohman:
This is note for myself:

I was able to generate the last commit in this series by executing:
`BUF_BREAKING_AGAINST_BRANCH=AritraDey-Dev/pr/aritra/eventchecker-generic-map-support make codegen && make vendor`

There is a comment in the Makefile for the protogen target:
```
 # Need to call vendor twice here, once before and once after codegen the reason
 # being we need to grab changes first plus pull in whatever gets generated here.
```
The code originally did this when this comment was introduced, and has since changed to not do this.

tools/protoc-gen-go-tetragon/eventchecker/field.go
```diff
@@ -70,8 +70,8 @@ func (field *Field) generateFrom(g *protogen.GeneratedFile, msg *CheckedMessage)
 }
 
 func (field *Field) getFrom(g *protogen.GeneratedFile, msg *CheckedMessage) (string, error) {
-	checkerVar := fmt.Sprintf("%s.%s", checkerVarName, field.GoName)
-	eventVar := fmt.Sprintf("%s.%s", eventVarName, field.GoName)
+	checkerVar := checkerVarName + "." + field.GoName
+	eventVar := eventVarName + "." + field.GoName
```
andrewstrohman, line 74: There are a lot of changes in this commit like this, where we change `Sprintf()` calls to string concatenation ("+"). It doesn't seem like the motivation to do this is related to what is described in the commit message. Should these changes be moved out into a separate commit?

tools/protoc-gen-go-tetragon/eventchecker/field.go
```diff
@@ -857,6 +832,83 @@ func (field *Field) typeName(g *protogen.GeneratedFile) (string, error) {
...
+            splitN := strings.SplitN
+            for _, s := range event.PodLabels {
+                 kv := splitN(s, "=", 2)
+                 if len(kv) == 2 {
+                     checkerVar[kv[0]] = *` + smatcher + `(kv[1])
+                 }
+            }
+            checker.PodLabels = checkerVar
+        }`, nil
+	}
```
andrewstrohman: I don't see this in the generated output. Why is this code not currently used?

api/vendor/github.com/cilium/tetragon/pkg/matchers/mapmatcher/mapmatcher.go
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: Apache-2.0
```
andrewstrohman, line 1: It would probably be better if this was in a different commit; one commit for your changes and another for the result of "make vendor".

I do it this way because I find it easier to rebase / cherry-pick. This way, I don't have to handle conflicts for the generated files. Instead, I just run "make vendor" to regenerate the commit that corresponds to the original "make vendor" commit.

pkg/matchers/mapmatcher/mapmatcher.go
```diff
@@ -0,0 +1,109 @@
...
+import (
+	"fmt"
+
+	"sigs.k8s.io/yaml"
+)
+
+// Matcher is a generic interface for matching values of type T.
+type Matcher[T any] interface {
+	Match(T) error
+}
```
andrewstrohman, line 15: I see that this type is never used. I think you created this for clarity and/or documentation purposes, which makes sense to me. I just want to confirm that this is intentional, and not something left over from an earlier approach.

pkg/matchers/mapmatcher/mapmatcher_test.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (m mockMatcher) Match(actual string) error {
+	if m.expected != actual {
+		return assert.AnError
+	}
+	return nil
+}
+
+func TestMapMatcher(t *testing.T) {
```
andrewstrohman, line 25: Should this test also test `UnmarshalJSON` ( via`yaml.Unmarshal`), like how is done in `TestPrimitiveMatcher`?

andrewstrohman:
I see the benefit of making more generic code that can be re-used. I also like how the From method is implemented for maps, in case we want to use that in the future.

That being said, I'm wondering if there any other motivation that I'm not aware of. Do you plan to use this for another change in the future? If so, please add that information to the PR, so that reviewers can better understand where you are headed.

andrewstrohman (approved):
I left some comments, but overall LGTM.

That being said, all this code and concepts were new to me as I had never looked at the protoc-gen-go-tetragon part of the codebase before reviewing this. So, it would be great if we could get another review from someone with more background/expertise in this this area before merging.

tools/protoc-gen-go-tetragon/eventchecker/field.go
AritraDey-Dev: Great catch! It's dead code left over from earlier iterations,the struct field is actually named PodLabels (so "Pod_Labels" never matches) and new generic map logic handles the native map perfectly, so I'll just remove it.

pkg/matchers/mapmatcher/mapmatcher.go
AritraDey-Dev: You're spot on! `Matcher[T]` is indeed currently unused as a formal type constraint,it's mostly there for docs to show the expected shape of the `matchers` that go into `MapMatcher`.
The main reason I used any for M in `MapMatcher[K, V, M] `instead of the `Matcher` interface is to give us max flexibility. Specifically, it lets this work seamlessly with existing matchers (like `StringMatcher`) that have pointer receivers, without us having to tweak their source files at all. Since the `Match` method dynamically checks both the value and its pointer during execution, having the formal interface constraint wasn't strictly required to get the job done.

pkg/matchers/mapmatcher/mapmatcher_test.go
AritraDey-Dev: I didn't add a specific unmarshal test for MapMatcher because its UnmarshalJSON method is just standard boilerplate mapping (using the common `Alias` type trick to prevent infinite recursion).
It has no custom branching or fallback logic to verify. `PrimitiveMatcher`, on the other hand, absolutely needed a specific test because it has handwritten, custom fallback logic to handle two completely different YAML formats (parsing both shorthand bare values and full structs).

tools/protoc-gen-go-tetragon/eventchecker/field.go
AritraDey-Dev: Yes moved this to a separate commit,thanks.

AritraDey-Dev:
>That being said, I'm wondering if there any other motivation that I'm not aware of. Do you plan to use this for another change in the future? If so, please add that information to the PR, so that reviewers can better understand where you are headed.

Hey, the main fix was `PodLabels`/`PodAnnotations`, but this patch made `MapMatcher` generic so any future proto map fields with non-string values (e.g. `map<string, uint32>`) just work without touching the matcher code again. Will update the PR description!

andrewstrohman (approved):

will-isovalent (commented):
Hey, thanks so much for the contribution. The changes make sense to me.

If it's not too much effort, I think we probably also need to think about how you could express integer ranges and inequalities here in addition to raw equality checks. This could also be done as a follow-up though.

kkourt:
> Hey, thanks so much for the contribution. The changes make sense to me.
> 
> If it's not too much effort, I think we probably also need to think about how you could express integer ranges and inequalities here in addition to raw equality checks. This could also be done as a follow-up though.


@will-isovalent do you think this should be in the context of this PR, or can it be a follow-up?

PR #4765, opened 2026-03-17
fix(pkg/ksyms): avoid kallsyms error when addresses are not readable.

FedeDP:
### Description
When `kptr_restrict` is set to 2, `/proc/kallsyms` shows 0 for all addresses (see here for more info: https://docs.kernel.org/admin-guide/sysctl/kernel.html#kptr-restrict). Instead of failing, log a warn message (just once).

Symbol address is indeed only used:
* [to sort the symbols list](https://github.com/cilium/tetragon/blob/main/pkg/ksyms/ksyms.go#L124)
* [to get a function offset](https://github.com/cilium/tetragon/blob/main/pkg/ksyms/ksyms.go#L161)

Both of them are not really used for validation.
The latter is used by `GetFnOffset`, in turn used by cmd/ksyms and [grpc `GetProcessKprobe::kernelStack()`](https://github.com/cilium/tetragon/blob/main/pkg/grpc/tracing/tracing.go#L367). `GetFnOffset` is update to fail with a specific error: "kernel symbols addresses are not available"

All of this allows tetragon to run on systems where `kptr_restrict` is set to 2.

### Changelog

```release-note
tetragon: policies can now be loaded even if kptr_restrict is equal to 2
```

Tested by manually setting `sudo sysctl kernel.kptr_restrict=2` on my laptop, and seeing tetragon fail with:
```
"level=warn msg="adding tracing policy failed" error="policy handler 'tracing' failed loading policy 'test-pol': validation failed: validateKprobeSpec: ksyms.KernelSymbols: no kernel symbols found: function srso_alias_untrain_ret reported at address 0, insufficient permissions? If not, kptr_restrict might be set to 2"
```

After the fix, the policy is working fine and we have this warning in the logs:
```
level=warn msg="function srso_alias_untrain_ret reported at address 0, insufficient permissions? If not, kptr_restrict might be set to 2"
level=info msg="Added kprobe" return=false function=security_file_permission override=false
```

Also, `kptr` executable with `kptr_restrict==2` fails with:
```
$ sudo ./ksyms 0xffffffffa870f925
level=warn msg="function srso_alias_untrain_ret reported at address 0, insufficient permissions? If not, kptr_restrict might be set to 2"
addr 0xffffffffa870f925: error: kernel symbols addresses are not available
```

### DIFF
--- a/pkg/grpc/tracing/tracing.go
+++ b/pkg/grpc/tracing/tracing.go
@@ -438,7 +438,7 @@ func kernelStack(event *MsgGenericKprobeUnix) []*tetragon.StackTraceEntry {
 		fnOffset, err := kernelSymbols.GetFnOffset(addr)
 		if err != nil {
 			// maybe group those errors as they might come in pack
-			logger.GetLogger().Warn("stacktrace: failed to retrieve symbol and offset", "address", fmt.Sprintf("0x%x", addr))
+			logger.GetLogger().Warn("stacktrace: failed to retrieve symbol and offset", "address", fmt.Sprintf("0x%x", addr), "err", err.Error())
 			continue
 		}
 		entry := &tetragon.StackTraceEntry{


tpapagian:
The approach seems reasonable, but I believe the failures in `TestKprobeKernelStackTrace` might be related.

FedeDP:
Yup :D looking into it rn!

FedeDP:
That test seems flaky, or is just me?
I am having trouble making it pass successfully on my local machine even on main branch :/ 
It sometimes does pass, sometimes doesn't.
I am on a 6.17 kernel btw

FedeDP:
Moving to draft until i understand what's happening with the bugged test.

FedeDP:
Ok the local issue is due to ubuntu backporting `6d08340d1e35 Revert "perf/x86: Always store regs->ip in perf_callchain_kernel()"` to 6.17 kernel, while our test check is:
```go
var symbolMatcher *sm.StringMatcher
	if kernels.MinKernelVersion("6.17") {
		// On 6.18+ x86, the probed function may not appear due to perf_callchain_kernel() changes
		// Check for syscall entry points instead
		if isArm() {
			symbolMatcher = sm.Prefix("el0_svc")
		} else {
			symbolMatcher = sm.Prefix("entry_SYSCALL_64")
		}
	} else {
		symbolMatcher = sm.Suffix("sys_newuname")
	}
``` 

EDIT: problem is, locally i cannot repro the CI issue :/

pkg/ksyms/ksyms.go
```diff
@@ -159,6 +160,10 @@ func (k *Ksyms) GetFnOffset(addr uint64) (*FnOffset, error) {
 
 // GetFnOffset -- retruns the FnOffset for a given address
 func (k *Ksyms) getFnOffset(addr uint64) (*FnOffset, error) {
+	if k.fnAddrDenied {
```
FedeDP, line 195: @tpapagian i fixed the test issue: i was previously checking `k.table[0].addr == 0` here; but it turns out that sometimes the very first line is indeed a 0 (but it's not a function).
I now use a proper boolean.

pkg/ksyms/ksyms.go
```diff
@@ -89,9 +90,9 @@ func NewKsyms(procfs string) (*Ksyms, error) {
 		sym.ty = fields[1]
 		sym.name = fields[2]
 
-		if sym.isFunction() && sym.addr == 0 {
-			err = fmt.Errorf("function %s reported at address 0, insufficient permissions? If not, kptr_restrict might be set to 2", sym.name)
-			break
+		if sym.isFunction() && sym.addr == 0 && !ksyms.fnAddrDenied {
+			ksyms.fnAddrDenied = true
+			logger.GetLogger().Warn(fmt.Sprintf("function %s reported at address 0, insufficient permissions? If not, kptr_restrict might be set to 2", sym.name))
```
olsajiri: nit looks good, but we could be proactive and try to read/display kptr_restrict value

pkg/ksyms/ksyms.go
```diff
@@ -159,6 +160,10 @@ func (k *Ksyms) GetFnOffset(addr uint64) (*FnOffset, error) {
 
 // GetFnOffset -- retruns the FnOffset for a given address
 func (k *Ksyms) getFnOffset(addr uint64) (*FnOffset, error) {
+	if k.fnAddrDenied {
+		return nil, errors.New("kernel symbols addresses are not available")
```
olsajiri, line 196: there seems to be just one user of the `GetFnOffset` in `kernelStack` and it does not display the error:

```
                fnOffset, err := kernelSymbols.GetFnOffset(addr)
                if err != nil {
                        // maybe group those errors as they might come in pack
                        logger.GetLogger().Warn("stacktrace: failed to retrieve symbol and offset", "address", fmt.Sprintf("0x%x", addr))
                        continue
                }
```

pkg/ksyms/ksyms.go
FedeDP: Makes sense, let me try to get it.

pkg/ksyms/ksyms.go
```diff
@@ -89,9 +113,18 @@ func NewKsyms(procfs string) (*Ksyms, error) {
...
 		sym.name = fields[2]
 
-		if sym.isFunction() && sym.addr == 0 {
-			err = fmt.Errorf("function %s reported at address 0, insufficient permissions? If not, kptr_restrict might be set to 2", sym.name)
-			break
+		if sym.isFunction() && sym.addr == 0 && !fnAddr0Warned {
+			fnAddr0Warned = true
+			// If we faild to red kptr_restrict value,
+			// try to proceed anyway.
+			if kptrRestrict == kptrRestrictUnknown || kptrRestrict == kptrRestrictStrict {
```
FedeDP, line 120: @olsajiri i decided to only enable the new behavior when either:
* we cannot properly read `kptr_restrict` value
* `kptr_restrict == 2`

In other cases, ie `[0,1]`, we will fail like we did before; because in those cases we **should** be able to read symbols addresses, and if we are not, it's an error IMO.

pkg/grpc/tracing/tracing.go
```diff
@@ -421,6 +421,11 @@ func kernelStack(event *MsgGenericKprobeUnix) []*tetragon.StackTraceEntry {
 	}
 
 	var stackTrace []*tetragon.StackTraceEntry
+	kernelSymbols, err := ksyms.KernelSymbols()
+	if err != nil {
+		logger.GetLogger().Warn("stacktrace: failed to read kernel symbols", logfields.Error, err)
+		return stackTrace
+	}
```
mtardy, line 428: thanks!!!

PR #4766, opened 2026-03-18
helm: fix rthooks container resources not being applied

PhilipSchmid:
The resources block in the tetragon-rthooks container template was using an invalid YAML construct:

  resources: {}
    {{- toYaml . | nindent 4 }}

This results in a YAML parse error when rthooks.resources is set, because a flow-style empty map ({}) cannot have block-style children.

Fix by removing the erroneous {} so the toYaml output is correctly nested under the resources key.

```release-note
helm: fix rthooks container resources not being applied
```


PhilipSchmid:
Let's please also backport this :). TIA!

FedeDP (approved):

PR #4778, opened 2026-03-19
fix(Makefile): use CURDIR instead of PWD for container volume mounts

dangome3:
### Description
Replace `$(PWD)` with `$(CURDIR)` in metrics-docs target container commands. CURDIR is more reliable as it's set by make itself, while PWD depends on the shell environment. 

`$(PWD)` was causing the following error in renovate job:

```
  Command failed: make gen-docs-references
  docker: invalid reference format
  Run 'docker run --help' for more information
  make: *** [Makefile:513: metrics-docs] Error 125
```

See [renovate job](https://github.com/cilium/tetragon/actions/runs/23285626838/job/67708291508#step:8:39595)

Looking closely you'll see that when the docker command run, it tried to run the following

```
docker run --rm -v : -w  docker.io/library/golang:1.26.1-alpine@sha256:xxxx ./tetragon-metrics-docs health >> docs/content/en/docs/reference/metrics.md\n"
```

`$(PWD)` is empty, that's why it has been changed to `$(CURDIR)`. See also [What is the difference between PWD and CURDIR?](https://stackoverflow.com/questions/52437728/bash-what-is-the-difference-between-pwd-and-curdir)

### Changelog

```release-note
use CURDIR instead of PWD for container volume mounts
```


mtardy:
I'll close this now since we already have 434785d9cd7e54ee588d711dc62eae4b6bac74f4. Thanks a lot for the investigation!!

PR #4779, opened 2026-03-20
bugtool: decouple pprof stats from gops

kkourt:


pkg/bugtool/bugtool.go
```diff
@@ -619,6 +620,27 @@ func (s *bugtoolInfo) getPProf(file string, gopsSignal byte) error {
 	return s.tarAddBuff(file, buff)
 }
 
+func (s *bugtoolInfo) addPprofInfo() {
```
mtardy: ah you know I cannot not mention that `s.getPProf` has different case lol, please ignore me

pkg/bugtool/bugtool.go
```diff
@@ -641,27 +663,6 @@ func (s *bugtoolInfo) addGopsInfo() {
...
-				WithError(err).
-				Warn("Failed to dump gops pprof")
-		} else {
-			s.multiLog.
-				WithField("gops-address", s.info.GopsAddr).
-				WithField("gops-path", s.info.GopsPath).
-				WithField("profile", name).
-				Info("Successfully dumped gops pprof")
-		}
-	}
```
mtardy, line 664: yeah or we could have move this up just after

```
	if s.info.GopsAddr == "" {
		s.multiLog.Info("Skipping gops dump info as daemon is running without gops, use --gops-address to enable gops")
		return
	}
```

I imagine having a separate method is cleaner.

PR #4782, opened 2026-03-20
Makefile: don't rely on PWD

dangome3:
[ upstream commit 4b8ccd9641b8e8135bd8574025a914899c333b79 ]


mtardy:
lol you can guess I have a weird connection from a train

PR #4786, opened 2026-03-21
bpf: Remove un-used memcmp macro

sayboras:
### Description

This should be fixed in Clang 15 as per below issue. Tetragon is using clang 20 as per Dockerfile.clang.

Relates: https://github.com/llvm/llvm-project/issues/26592

### Changelog

```release-note
bpf: enable __builtin_memcmp() macro definition
```


sayboras:
> we don't use any `memcmp` macro in the code, should we just remove it?

Yes, we can. 

(Oh no, need to use my github searching skills to find the patch :cry: )

PR #4788, opened 2026-03-22
renovate: Bump golang to 1.25 for all stable versions

sayboras:
### Description
Golang 1.24 is EOL wheng 1.26 is released.

### Changelog

```release-note
renovate: Bump golang to 1.25 for all stable versions
```

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -308,13 +308,13 @@
       ]
     },
     {
-      // update go patch for 1.24 for stable branches
+      // update go patch for 1.25 for stable branches
       "enabled": true,
       "matchPackageNames": [
         "go",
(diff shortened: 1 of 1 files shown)


mtardy:
well too late lol

sayboras:
> In the past, mostly because we had cilium it created real issues because some deps couldn't be updated anymore so we had to revert. Most of it should be okay now but would be worth testing

Ah this is actually motivated from Cilium recently as part of releases, as for now, all Cilium stable releases are with newer go version 1.25+, I think it should be fine for cilium/tetragon. 

https://github.com/cilium/cilium/pull/44801
https://github.com/cilium/proxy/pull/1796

PR #4794, opened 2026-03-23
process: fix unreachable deleted state check in GC and add observability metric

kalakotima:
## Summary

Fix a logic bug in the cache garbage collector where the `deleted`
state check was unreachable, add the missing observability counter
that the existing TBD comment called for, and register two metrics
that were defined but never exposed to Prometheus.

### Problem

In `cacheGarbageCollector`, the `deleteChan` select case had two
consecutive checks in the wrong order:

```go
// This catches ALL non-inUse states including deleted
if p.color != inUse {
    p.color = deletePending  // ← deleted processes land here
    continue                 // ← color silently reset, never tracked
}

// THIS IS NEVER REACHED
// deleted != inUse, so it already matched the branch above
if p.color == deleted {
    continue
}

This patch:
- Moves the deleted check before the != inUse check (fixes dead code)
- Adds processCacheEarlyDeletion counter (fulfills existing TBD comment)
- Registers processCacheCapacityEvictions which was defined but never exposed

PR #4795, opened 2026-03-23
process: use chan struct{} for stopChan signal channel

kalakotima:
## Summary

Replace `chan bool` with `chan struct{}` for the `stopChan` signal
channel in the process cache.

## Background

The `stopChan` field in `Cache` is used exclusively as a shutdown
signal — no boolean value is ever read from it. Using `chan bool`
implies the value being sent carries meaning, which is misleading
to readers of the code.

## Changes

- Change `stopChan` field type from `chan bool` to `chan struct{}`
- Change `make(chan bool)` to `make(chan struct{})` in
  `cacheGarbageCollector`
- Change `pc.stopChan <- true` to `pc.stopChan <- struct{}{}`
  in `purge()`

## Why This Matters

`chan struct{}` is the idiomatic Go pattern for signal-only channels
because:

1. It explicitly communicates intent — no value, signal only
2. `struct{}` occupies zero bytes of memory vs `bool` which
   occupies 1 byte per send
3. Sending `true` implies there could be a `false` case — there
   is not, making `chan bool` semantically wrong here

### DIFF
--- a/pkg/process/cache.go
+++ b/pkg/process/cache.go
@@ -25,7 +25,7 @@ type Cache struct {
 	cache      *lru.Cache[string, *ProcessInternal]
 	size       int
 	deleteChan chan *ProcessInternal
-	stopChan   chan bool
+	stopChan   chan struct{} // ← CHANGED from chan bool
 }
 
 // garbage collection states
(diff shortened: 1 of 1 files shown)


mtardy:
https://github.com/cilium/tetragon/pull/4794#issuecomment-4114284207

PR #4798, opened 2026-03-24
tetra: add global --max-recv-size flag

will-isovalent:
Add a global `--max-recv-size` persistent flag to the `tetra` CLI (default 10MB). Previously, the gRPC client connections used gRPC's hardcoded 4MB default, causing `ResourceExhausted` errors when a server response exceeds that limit.

The limit is now applied at the connection level via `WithDefaultCallOptions(MaxCallRecvMsgSize(...))` in `common.NewClient()`, so all commands benefit without any per-command changes.

The existing per-command `--max-recv-size` flags on `tetra bugtool` and `tetra debug dump processcache` are removed, as they are superseded by the new global flag. Both commands now read `common.MaxRecvMsgSize` directly.

```release-note
Add global `--max-recv-size` flag to `tetra` CLI to configure the maximum gRPC receive message size (default 10MB, up from 4MB).
```

### DIFF
--- a/cmd/tetra/common/flags.go
+++ b/cmd/tetra/common/flags.go
@@ -15,15 +15,16 @@ import (
 )
 
 const (
-	KeyColor         = "color"          // string
-	KeyDebug         = "debug"          // bool
-	KeyOutput        = "output"         // string
-	KeyTty           = "tty-encode"     // string
-	KeyServerAddress = "server-address" // string
-	KeyTimeout       = "timeout"        // duration
-	KeyRetries       = "retries"        // int
-	KeyNamespace     = "namespace"      // string
-	KeyLogLevel      = "loglevel"       // string
+	KeyColor          = "color"          // string
+	KeyDebug          = "debug"          // bool
+	KeyOutput         = "output"         // string
+	KeyTty            = "tty-encode"     // string
+	KeyServerAddress  = "server-address" // string
+	KeyTimeout        = "timeout"        // duration
+	KeyRetries        = "retries"        // int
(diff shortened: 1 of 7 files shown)


cmd/tetra/main.go
```diff
@@ -47,5 +47,6 @@ func New() *cobra.Command {
 	flags.StringVar(&common.ServerAddress, common.KeyServerAddress, "", "gRPC server address")
 	flags.DurationVar(&common.Timeout, common.KeyTimeout, 30*time.Second, "Connection timeout")
 	flags.IntVar(&common.Retries, common.KeyRetries, 1, "Connection retries with exponential backoff")
+	flags.IntVar(&common.MaxRecvMsgSize, common.KeyMaxRecvMsgSize, 10*1024*1024, "Maximum gRPC message size in bytes the client can receive")
```
mtardy: hey you could add this `10*1024*1024` as a defaults in https://github.com/isovalent/hubble-fgs/blob/master/pkg/defaults/defaults.go and put that here

cmd/tetra/common/flags.go
```diff
@@ -15,26 +15,28 @@ import (
...
 var (
-	Debug         bool
-	ServerAddress string
-	Timeout       time.Duration
-	Retries       int
+	Debug          bool
+	ServerAddress  string
+	Timeout        time.Duration
+	Retries        int
+	MaxRecvMsgSize = 10 * 1024 * 1024 // 10MB default
```
mtardy: and here

PR #4800, opened 2026-03-26
fix : add enum support for resolve

### LINKED ISSUE #4799: Enum support for uprobe resolve

When using uprobes with BTF resolve support, enum is not a recognized type and this makes the resolve function not return the right value:

```yaml
kind: TracingPolicy
metadata:
  name: "uprobe-example"
spec:
  uprobes:
  - path: "/home/litios/testbin"
    btfPath: "/home/litios/testbin.btf"
    symbols:
    - "testfunc"
    args:
    - index: 0
      type: "uint64"
      btfType: "test_struct"
      resolve: "testenum"
    selectors:
    - matchActions:
      - action: Post  
```

```c
#include <stdint.h>

enum test_enum {
    TEST_ENUM_0 = 0,
    TEST_ENUM_1 = 1,
    TEST_ENUM_2 = 2,
};

struct test_struct {
    int fd;
    char* test;
    enum test_enum testenum;
};

int testfunc(struct test_struct s) {
    return s.testenum;
}

void main() {
    struct test_struct ctx;
    ctx.fd = 1;
    ctx.test = "dada";
    ctx.testenum = TEST_ENUM_1;
    testfunc(ctx);
}
```

```
[11] ENUM 'test_enum' encoding=UNSIGNED size=4 vlen=3
	'TEST_ENUM_0' val=0
	'TEST_ENUM_1' val=1
	'TEST_ENUM_2' val=2
[12] STRUCT 'test_struct' size=24 vlen=3
	'fd' type_id=7 bits_offset=0
	'test' type_id=9 bits_offset=64
	'testenum' type_id=11 bits_offset=128
```

```yaml
    "policy_name": "uprobe-example",
    "args": [
      {
        "size_arg": "17"
      }
    ],
```

```
pwndbg> x/4gx &ctx
0x7fffffffe070:	0x0000000000000001	0x0000555555556004
0x7fffffffe080:	0x00007fff00000001	0x00007fffffffe1b8
pwndbg> p ctx
$3 = {
  fd = 1,
  test = 0x555555556004 "dada",
  testenum = TEST_ENUM_1
}
pwndbg> p (int)TEST_ENUM_1
$4 = 1
```

It looks like `enum` is not being resolved properly with uprobes resolve (BTF).

(@tdaudi provided a solution that seems to work already! PR pending)

_No response_

### Code of Conduct

tdaudi:
Fixes https://github.com/cilium/tetragon/issues/4799

## Description

Enums are uint32. Currently, because of the way integer are handled on BPF, we need to add an extra dereferencing in the resolve algo to have it working properly. For this reason I did not add tests as this little hack won't be necessary after a fix will be done.

```
[11] ENUM 'test_enum' encoding=UNSIGNED size=4 vlen=3
```

### DIFF
--- a/contrib/tester-progs/uprobe-resolve.c
+++ b/contrib/tester-progs/uprobe-resolve.c
@@ -14,11 +14,17 @@ struct mysubstruct {
 	char     *buff;
 };
 
+enum myenum_u32 {
+	SUCCESS = 1234,
+	FAILURE = 4321,
+};
+
 struct mystruct {
 	uint8_t  v8;
 	uint16_t v16;
 	uint32_t v32;
 	uint64_t v64;
+	enum myenum_u32 enum_u32;
 	struct mysubstruct sub;
 	struct mysubstruct *arr[10];
 	struct mysubstruct **dyn;
@@ -28,7 +34,7 @@ struct mystruct {
 void usage(char *argv0)
 {
 	fprintf(stderr, "Usage: %s <field> <val>\n", argv0);
-	fprintf(stderr, "field can be one of: v8, v16, v32, v64, sub.v32 arr[idx].v64 dyn[idx].v64, subp.buff, subp.v64\n");
+	fprintf(stderr, "field can be one of: v8, v16, v32, v64, enum_u32, sub.v32 arr[idx].v64 dyn[idx].v64, subp.buff, subp.v64\n");
 }
 
 // without noinline, the symbol is found, but no event fires
@@ -68,6 +74,8 @@ int main(int argc, char *argv[])
 		s.v32 = val;
 	} else if (!strcmp(field, "v64")) {
 		s.v64 = val;
+	} else if (!strcmp(field, "enum_u32")) {
+		s.enum_u32 = (val == SUCCESS) ? SUCCESS : FAILURE;
 	} else if (!strcmp(field, "subp.v64")) {
 		struct mysubstruct s2 = {
 			.v64 = val,
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -351,10 +351,11 @@ func processMembers(
 			members,
 		)
 	}
-	if t, ok := currentType.(*btf.Pointer); ok {
+	switch t := currentType.(type) {
+	case *btf.Pointer:
 		btfArgs[i].IsPointer = uint16(1)
 		currentType = t.Target
-	} else if _, ok := currentType.(*btf.Int); ok {
+	case *btf.Int, *btf.Enum:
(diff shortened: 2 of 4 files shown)


andrewstrohman (approved):

andrewstrohman:
> Enums are uint32.

I don't think this is always true. I think they can be other integer types too. I think that's why the definition of `Enum` specifies `Size` and `Signed`:

```
// Enum lists possible values.
type Enum struct {
	Name string
	// Size of the enum value in bytes.
	Size uint32
	// True if the values should be interpreted as signed integers.
	Signed bool
	Values []EnumValue
}
```


But given your debug output:

> [11] ENUM 'test_enum' encoding=UNSIGNED size=4 vlen=3

I agree the user would configure the type as `uint32` in their policy for this case. I'm just pointing out that the correct type to configure in the policy seems to be situational.

olsajiri (approved):
lgtm, thanks

FedeDP (approved):

kkourt (commented):
> Enums are uint32. Currently, because of the way integer are handled on
> BPF, we need to add an extra dereferencing in the resolve algo to have
> it working properly.

That's the most common case, but it's not universal as Andy pointed out.

```
sudo bpftool btf dump file /sys/kernel/btf/vmlinux | grep 'ENUM ' | grep -v 'size=4'
[1148] ENUM 'rw_hint' encoding=UNSIGNED size=1 vlen=6
[2359] ENUM 'blk_integrity_checksum' encoding=UNSIGNED size=1 vlen=4
[4871] ENUM '_cache_table_type' encoding=UNSIGNED size=1 vlen=4
[4872] ENUM '_tlb_table_type' encoding=UNSIGNED size=1 vlen=15
[23364] ENUM 'scsi_cmnd_submitter' encoding=UNSIGNED size=1 vlen=3
[115702] ENUM 'hub_led_mode' encoding=UNSIGNED size=1 vlen=8
```

Shouldn't we check the enum size and act appropriately?

tdaudi:
Thank you for your clarifications. I think it's better this way indeed.

Changes includes :

- One commit for the `type:` transformation when [`lastBTFType`](https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/generickprobe.go#L789) is an `*btf.Enum`.  (I give details in the commit message)
- I "recycle" the `v32` field in `TestUprobeResolve` to validate the enum with uint32 format.

andrewstrohman (approved):

kkourt (approved):
LGTM, but can you please rephrase the first commit message that still mentions:
> Enums are uint32.

tdaudi:
Sorry, it should be better now. Thanks

PR #4805, opened 2026-03-26
Honor stderrthreshold when logtostderr is enabled

pierluigilenoci:
## Description

klog has a long-standing bug where setting `-logtostderr=true` causes the `-stderrthreshold` flag to be silently ignored. All log messages are sent to stderr regardless of their severity.

klog v2.140.0 introduced two new flags that allow callers to opt into the correct behavior:
- `-legacy_stderr_threshold_behavior=false`
- `-stderrthreshold=INFO`

This PR sets them in `initializeKLog()` (inside `pkg/logger/klog_bridge.go`), right after `klog.InitFlags()` is called.

### Defensive fix

While Tetragon currently sets `logtostderr=false` and uses `SetOutputBySeverity` to route klog output through its own logging pipeline, this fix is defensive: if the klog configuration changes in the future, `-stderrthreshold` will be honored correctly.

### go.mod

No dependency bump needed — Tetragon already uses `k8s.io/klog/v2 v2.140.0`.

Reference: kubernetes/klog#212

/cc @tklauser @mtardy

### DIFF
--- a/pkg/logger/klog_bridge.go
+++ b/pkg/logger/klog_bridge.go
@@ -36,6 +36,10 @@ func initializeKLog(logger *slog.Logger) error {
 	// update them from this file.
 	klog.InitFlags(klogFlags)
 
+	// Opt into fixed stderrthreshold behavior (kubernetes/klog#212).
+	klogFlags.Set("legacy_stderr_threshold_behavior", "false")
+	klogFlags.Set("stderrthreshold", "INFO")
+
 	// Make sure klog does not log to stderr as we want it to control the output
 	// of klog so we want klog to log the errors to each writer of each level.
 	klogFlags.Set("logtostderr", "false")


pierluigilenoci:
Hi @olsajiri, @kkourt — gentle ping. Would you be able to review this when you get a chance? Thank you!

kkourt:
Ah, I see your PR description now.
> While Tetragon currently sets logtostderr=false and uses SetOutputBySeverity to route klog output through its own logging pipeline, this fix is defensive: if the klog configuration changes in the future, -stderrthreshold will be honored correctly.

I don't think we need to be "defensive" about this. We are setting the right option, so IMO there is no need for additional flags.

pierluigilenoci:
@kkourt Thank you for the review — that's a fair point. You're right that Tetragon explicitly sets `logtostderr=false` and routes klog output via `SetOutputBySeverity`, so the klog bug with `stderrthreshold` being silently ignored doesn't actually affect the current logging behavior.

I included this as a defensive measure in case the klog configuration is changed in the future, but I understand that adding "just in case" code isn't necessarily a net positive — it adds complexity without solving a real problem today.

If you'd prefer not to include this change, I completely understand. Should I close this PR?

mtardy:
> If you'd prefer not to include this change, I completely understand. Should I close this PR?

yes

I think that would be nice from your side to check whether your change actually make sense instead of flooding dozens of project with the same PR just because they use a dependency. I understand the idea but please do not offload the work to check "if this actually makes sense" to the maintainers.

image

PR #4809, opened 2026-03-30
tetragon: Remove the bpf_kfunc tag check from detectKfunc

olsajiri:
It's enough to check function presence in kernel's BTF.

The check for bpf_kfunc tag is nice to have but was introduced after kfuncs, so not all kernels have it. Tetragon uses only straight forward kfuncs (without kernel functions name collisions) so just the BTF check is fine.

Plus other related fixes.

### DIFF
--- a/pkg/bpf/detect_linux.go
+++ b/pkg/bpf/detect_linux.go
@@ -622,8 +622,11 @@ func detectKfunc(name string) bool {
 		return false
 	}
 
-	// kfunc has bpf_kfunc tag attached
-	return len(fn.Tags) == 1 && fn.Tags[0] == "bpf_kfunc"
+	// We used to check for 'bpf_kfunc' declaration tag, which is nice to
+	// have but was introduced after kfuncs, so not all kernels have it.
+	// It's enough to check function presence in kernel's BTF.
+
+	return true
 }
 
 func HasKfunc(name string) bool {


pkg/sensors/program/loader_linux.go
```diff
@@ -910,7 +911,10 @@ func rewriteConstants(spec *ebpf.CollectionSpec, consts map[string]any) error {
 
 	confs := map[string]func(v *ebpf.VariableSpec) error{
 		"CONFIG_ITER_NUM": func(v *ebpf.VariableSpec) error {
-			enabled := bpf.HasKfunc("bpf_iter_num_new")
+			// We can't use numberic iterator until we get following fix from 6.9 kernel:
```
mtardy: fyi: you wrote numberic here and in the commit description

pkg/bpf/detect_linux.go
```diff
@@ -622,8 +622,7 @@ func detectKfunc(name string) bool {
 		return false
 	}
 
-	// kfunc has bpf_kfunc tag attached
-	return len(fn.Tags) == 1 && fn.Tags[0] == "bpf_kfunc"
+	return true
```
mtardy, line 629: nit: feels like a comment would be useful here for the explanation to stick around on this simplification

olsajiri: ok, will add, thnx

PR #4811, opened 2026-03-31
new: domain-sharded tp management

### LINKED ISSUE #4808: Manage mixage of CRD and locally pushed tracing policies (tetra)

Today, you can use `tetra tracingpolicy` to load a policy; but you can also install new policies through CRD.
The problem is: given a policy has been installed via k8s, do we allow `tetra tracingpolicy ...` to modify it?

Also, this is a future-proofing request, since new tracingpolicy sources can be added later on and amplify the issue.

My proposal is for tetragon Sensors Manager to track the "source" of a tracingpolicy.

Let's add a new `source` (or `domain`) string to the [Sensors Manager collectionKey structure](https://github.com/cilium/tetragon/blob/main/pkg/sensors/collection.go#L56).

My proposal is to update the code to:
* set the source to `tetra-cli` for policies registered via tetra / initial command line argument (`--tracing-policy`)
* set the source to `k8s` for both `TracingPolicyNamespaced` and normal (ie: non-Namespaced) `TracingPolicy` pushed via CRD
* add a cli switch to `tetra tracingpolicy` like: `--force` to enforce the override for any source (ie: if you always invoke `tetra tracingpolicy --force ...` , the behavior will be exactly the current one)

At that point, Sensors Manager will differentiate between all of these policies that will be effectively managed specifically for each source, and each source can only operate on its policies, unless the user knowingly enforces the old behavior via `tetra tracingpolict --force` cli flag.

### Code of Conduct

FedeDP:
Fixes #4808

### Description

Add `domain` sharding for tracing policies in sensors manager.
Each "domain" can only see and work on its own policies; `tetra tracingpolicy` sub commands gained a new `--domain` flag; by default it will only act upon `grpc` domain (enforced by the grpc API), but it can be enforced to work eg: on k8s domain, for debug purposes.

Also a new `tetra tracingpolicy domains` command has been added to list all available domains (this is a dynamic list, ie: it depends upon loaded policies domains).

Examples of current impl:
```
$ sudo ./tetragon --btf /sys/kernel/btf/vmlinux  --bpf-lib ./bpf/objs/ --tracing-policy /home/fdipierr/tp_passwd.yaml
```

We will then have:
```
$ ./tetra tracingpolicy domains
[static]
```

Then we load a new policy via tetra:
```
$ ./tetra tracingpolicy add ~/tp_uprobe_ret_copy.yaml 
tracing policy "/home/fdipierr/tp_uprobe_ret_copy.yaml" added
```

And we now have a new domain:
```
$ ./tetra tracingpolicy domains
[grpc static]
```

We then try to remove the tetra policy but looking for it in the `static` domain:
```
$ ./tetra tracingpolicy delete trace-bash-readline --domain static
Error: failed to delete tracing policy: rpc error: code = Unknown desc = tracing policy {trace-bash-readline  static} does not exist
```
But in the correct domain (`grpc` is the default for grpc connections)
```
$ ./tetra tracingpolicy delete trace-bash-readline
tracing policy "trace-bash-readline" deleted
```

### Changelog

```release-note
new: domain-sharded tp management
```


FedeDP:
Rebased on top of latest main after the landing of big https://github.com/cilium/tetragon/pull/4265 :)

pkg/metrics/policymetrics/policymetrics_test.go
FedeDP, line 1: Updated to declare tracingpolicies before using them, since we are going to need to call `tp.TpDomain()`.

api/v1/tetragon/sensors.proto
```diff
@@ -30,7 +30,10 @@ message ListSensorsResponse {
   repeated SensorStatus sensors = 1;
 }
 
-message ListTracingPoliciesRequest {}
+message ListTracingPoliciesRequest {
+  // domain to be listed; empty to list all domains
+  string domain = 1;
```
FedeDP, line 35: We have basically 3 domains for policies right now:
* "static" domain for tracing policies loaded by configured folder/file
* "grpc" for grpc-loaded policies (eg: tetra CLI)
* "k8s" for policies loaded by the crd watcher

In the tests this knowledge may be leaked, but normally this is all transparent in the code, ie:
* "k8s" is enforced by the k8s `TracingPolicy::TpDomain` method on addition, and by the crd watcher on deletion
* "static" is enforced by the normal `TracingPolicy::TpDomain` method on addition (there is no deletion path for static policies)
* "grpc" is enforced by grpc server for both additions and deletions

pkg/sensors/collection.go
```diff
@@ -54,14 +54,21 @@ func (s TracingPolicyState) ToTetragonState() tetragon.TracingPolicyState {
 // collectionKey is the unique key for sensors
 // this enables policies with the same name for different namespaces
 type collectionKey struct {
-	name, namespace string
+	name, namespace, domain string
```
FedeDP, line 57: Each tracing policy is now referenced by its domain too.

pkg/sensors/handler.go
```diff
@@ -14,7 +16,10 @@ import (
 	"github.com/cilium/tetragon/pkg/tracingpolicy"
 )
 
-var BaseSensorName = "__base__"
+const (
+	BaseSensorName = "__base__"
+	sensorsDomain  = "sensors"
```
FedeDP, line 21: As explained above, the `sensors` domain is the one enforced by Sensors related APIs.

pkg/server/server.go
```diff
@@ -238,7 +254,12 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
 		"metadata.namespace", namespace,
 		"metadata.name", tp.TpName())
 
-	if err := s.observer.AddTracingPolicy(ctx, tp); err != nil {
+	gtp := GRPCTracingPolicy{tp, grpcDomain}
+	if req.GetDomain() != "" {
```
FedeDP, line 254: Normally (and by default) it is just `grpc`, but for debug purposes eg: via tetra, users can customize the `domain` they want to act upon.

pkg/server/server.go
```diff
@@ -223,6 +226,15 @@ func (s *Server) ListSensors(_ context.Context, _ *tetragon.ListSensorsRequest)
 	return nil, errors.New("ListSensors is deprecated")
 }
 
+type GRPCTracingPolicy struct {
```
FedeDP, line 229: Small wrapper around `TracingPolicy` to enforce grpc specific domain.

FedeDP:
Q: do we want to expose the domain in the `tetra tracingpolicy list` command? Right now, i chose not to update the [`TracingPolicyStatus` message](https://github.com/cilium/tetragon/blob/06be127c7e51a7b957456c1dcf6dddb69b58962a/api/v1/tetragon/sensors.proto#L87).

pkg/tracingpolicy/generictracingpolicy.go
```diff
@@ -14,6 +14,8 @@ import (
 type TypeMeta = metav1.TypeMeta
 type ObjectMeta = metav1.ObjectMeta
 
+const staticDomain = "static"
```
kkourt: Let's add this in a file like `tracingpolicy.go` where we don't have to duplicate it for the k8s and nok8s builds.

api/v1/tetragon/sensors.proto
kkourt, line 111: How about we add a domain in `TracingPolicyStatus`?

I think it would be useful to do something like `tetra tp list --all-domains" that sets the gRPC request's domain to "", so that we can get all policies from all domains.

cmd/tetra/tracingpolicy/tracingpolicy.go
```diff
@@ -134,19 +144,23 @@ func tpDelCmd() *cobra.Command {
 	}
 	flags := ret.Flags()
 	flags.StringVarP(&namespace, common.KeyNamespace, "n", "", "Namespace of the tracing policy.")
+	flags.StringVarP(&domain, "domain", "", "", "Domain to be used. Use k8s to act against CRD policies.")
```
kkourt: `s/against/on/`?

kkourt (commented):
Thanks!

Can you please add a commit message in the first commit that introduces commits with the motivation for this change?

Things like https://github.com/cilium/tetragon/pull/4811/changes/6a84712d31b4ae75a3fa60076d3d8f674095b18a#r3052523498, are also well-suited for commit messages. Making the context part of the git history is very valuable.


Also, do you think we should enforce some constraints on what is allowed (e.g., characters) in a domain name? I think that's probably a good idea.

api/v1/tetragon/sensors.proto
kkourt: Saw https://github.com/cilium/tetragon/pull/4811#issuecomment-4221962045, after my review. I think it makes sense to add it.

FedeDP: Yup, i asked for it here: https://github.com/cilium/tetragon/pull/4811#issuecomment-4221962045
Thanks for answering :laughing: 
I'll do it!

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -148,6 +154,10 @@ func (tp *TracingPolicy) TpInfo() string {
 	return fmt.Sprintf("%s (object:%d/%s) (type:%s/%s)", tp.ObjectMeta.Name, tp.ObjectMeta.Generation, tp.ObjectMeta.UID, tp.TypeMeta.Kind, tp.TypeMeta.APIVersion)
 }
 
+func (tp *TracingPolicy) TpDomain() string {
+	return k8sDomain
```
kkourt, line 158: I find this confusing. Looking at the subsequent patches, it seems that what you want to do is have all other users of TracingPolicies use a different type that returns a different TpDomain().

Would it be a better choice to add a Domain field in the TracingPolicy? We don't have to expose it to users. It can be a private field (with getters and setters) or maybe marked with `json:"-"`?

FedeDP:
> Also, do you think we should enforce some constraints on what is allowed (e.g., characters) in a domain name? I think that's probably a good idea.

I fully agree!

EDIT: but also, since we are the ones setting them and they are just returned from interface method implementations, i am not sure whether we need it. After all, users cannot change it and they are all statically lived strings.

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
FedeDP: Why is it confusing? I mean, i don't find it confusing obviously :) Mind to share your concerns?

Indeed i find it rather simple and straightforward: each implementation of the interface will have its own specific domain.

kkourt: That's a good question.

One part is that historically, we used the same type for multiple "domains" (e.g., gRPC and static). I also feel that at least for the `GenericTracingPolicy` we can have the same type everywhere, and it can be used to make the domains more dynamic in the future.

kkourt: But I don't feel strongly about it. I do think, however, that we would need to document the new behavior in the commit message (and maybe the types) because this is a change from what we did before.

FedeDP: >  also feel that at least for the GenericTracingPolicy we can have the same type everywhere, and it can be used to make the domains more dynamic in the future.

What i don't like is that if we allow people to customize the domain, sooner or later we might end up with mixed domains over and over (ie: "oh i added a `GenericTracingPolicy { Domain: "k8s" }` because x,y,z".
I prefer the "one type for each" in this case; but again, just like you, this is not a strong opinion. 
Luckily, we can change in the future without any breaking change.

>  I do think, however, that we would need to document the new behavior in the commit message (and maybe the types) because this is a change from what we did before.

Fully agree! Let me improve the commit messages :) 

Anyway, would love to hear more feedback on this topic; let me ping @mtardy @will-isovalent 
Thanks!

FedeDP: I tried to add more details to first 2 commits: https://github.com/cilium/tetragon/pull/4811/changes/fc1ce375f0ce8ee44907c61d0bf0840db7f5fd72 and https://github.com/cilium/tetragon/pull/4811/changes/653f046890db026ec44a11ecd8499700b5d58454; PTAL :)

cmd/tetra/tracingpolicy/tracingpolicy.go
```diff
@@ -104,11 +109,15 @@ func tpAddCmd() *cobra.Command {
 	}
 	flags := ret.Flags()
 	flags.StringVarP(&mode, "mode", "m", "", "Tracing policy mode (enforce|monitor)")
+	flags.StringVarP(&domain, "domain", "", "", "Domain to be used. Use k8s to act on CRD policies.")
```
kkourt: Should we also mention what happens if the domain is left empty?

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -32,6 +32,8 @@ const (
 
 	// TPKindDefinition is the kind name of Cilium Tracing Policy
 	TPNamespacedKindDefinition = "TracingPolicyNamespaced"
+
+	k8sDomain = "k8s"
```
kkourt, line 36: nit: How about we add a new package with all the domain constants? This means that we can use them in the CLI code as well.

pkg/server/server.go
```diff
@@ -44,14 +46,14 @@ type observer interface {
 	AddTracingPolicy(ctx context.Context, policy tracingpolicy.TracingPolicy) error
 	// DeleteTracingPolicy deletes a tracing policy that was added with
 	// AddTracingPolicy as defined by its name (policy.TpName()).
-	DeleteTracingPolicy(ctx context.Context, name string, namespace string) error
+	DeleteTracingPolicy(ctx context.Context, name string, namespace string, domain string) error
 	// ListTracingPolicies lists active traing policies
-	ListTracingPolicies(ctx context.Context) (*tetragon.ListTracingPoliciesResponse, error)
+	ListTracingPolicies(ctx context.Context, domain string) (*tetragon.ListTracingPoliciesResponse, error)
```
kkourt, line 52: nit: can we document in the comment what happens if the domain is ""?

cmd/tetra/tracingpolicy/tracingpolicy.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/cilium/tetragon/pkg/tracingpolicy"
 )
 
-func tpConfigure(name, namespace string, enable *bool, mode *tetragon.TracingPolicyMode) error {
+func tpConfigure(name, namespace, domain string, enable *bool, mode *tetragon.TracingPolicyMode) error {
```
kkourt, line 20: nit: any reason why the CLI changes were not split in another commit?

kkourt (approved):
LGTM, thanks!

Left some nits, but it's up to you if you want to address them.

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
FedeDP: Indeed i tried hard not to expose them :) i don't think we should care about what the value is from the outside.

> This means that we can use them in the CLI code as well.

Do we need to?

cmd/tetra/tracingpolicy/tracingpolicy.go
FedeDP: Makes sense!

cmd/tetra/tracingpolicy/tracingpolicy.go
FedeDP: You are right, let me split them!

FedeDP:
Also, rebased on top of main.

api/v1/tetragon/sensors.proto
mtardy: why not using an enum in the proto definition?

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -66,6 +68,10 @@ func (tp *TracingPolicyNamespaced) TpInfo() string {
 	return fmt.Sprintf("%s (object:%d/%s) (type:%s/%s)", tp.ObjectMeta.Name, tp.ObjectMeta.Generation, tp.ObjectMeta.UID, tp.TypeMeta.Kind, tp.TypeMeta.APIVersion)
 }
 
+func (tp *TracingPolicyNamespaced) TpDomain() string {
```
mtardy, line 71: just fyi we named it `TpSomething` because `Name` and `Namespace` where already taken on the object, I guess you can follow the convention or just name it Domain here.

pkg/sensors/collection.go
```diff
@@ -54,14 +54,14 @@ func (s TracingPolicyState) ToTetragonState() tetragon.TracingPolicyState {
...
```
mtardy, line 57: same, should it be its own type here instead of a string? if it's meant to be a freeform thing why not a string, if it's backed by actual types, maybe an enum is better?

cmd/tetra/tracingpolicy/tracingpolicy.go
```diff
@@ -134,18 +144,22 @@ func tpDelCmd() *cobra.Command {
 	}
 	flags := ret.Flags()
 	flags.StringVarP(&namespace, common.KeyNamespace, "n", "", "Namespace of the tracing policy.")
+	flags.StringVarP(&domain, "domain", "", "", "Domain to be used. Use k8s to act on CRD policies. By default only acts against grpc domain.")
```
mtardy, line 147: nit: for user facing stuff I'd write `gRPC`.

EDIT: ah but it's the domain thing, maybe make it case insensitive, or just do nothing... nevermind

pkg/sensors/manager_test.go
```diff
@@ -79,6 +80,52 @@ func TestAddPolicies(t *testing.T) {
...
+	err = mgr.AddTracingPolicy(ctx, &policy)
+	require.NoError(t, err)
+
+	// Adding it once again to the same domain should fail
+	err = mgr.AddTracingPolicy(ctx, &policy)
+	require.Error(t, err)
+
+	// Adding it to a new domain is ok
+	// Use the GRPCTracingPolicy wrapper to set the domain.
+	gtp := server.GRPCTracingPolicy{TracingPolicy: &policy, Domain: "test"}
```
mtardy: shouldn't the type you used be enough to check which domain from `AddTracingPolicy`? You could detect the underlying type and associate automatically with the correct domain nop?

mtardy (commented):
Thanks look good, am I correct now assuming that we can have two policies with the same name but in different domain then? Name/Namespace combination is not unique anymore?

Here are a few comments.

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
FedeDP: I think `TpDomain` feels right even if we don't really need to name it like that!

FedeDP: Thanks for pointing that out though!

FedeDP:
> am I correct now assuming that we can have two policies with the same name but in different domain then? Name/Namespace combination is not unique anymore?

Exactly!

pkg/sensors/manager_test.go
FedeDP: Yes but i didn't want to leak that knowledge down to sensors manager.
Indeed what i can do is letting `GRPCTracingPolicy.TpDomain()` return `grpcDomain` if its `Domain` field is empty.
`GRPCTracingPolicy` is mostly used to allow `tetra` to customize the `Domain`.

pkg/sensors/collection.go
FedeDP: Here it's a branch in my fork with the changes: https://github.com/FedeDP/tetragon/pull/28
I agree that using an enum is great in this case; feel free to quickly review that PR, most of the changes are in the first commit (the one introducing the proto definitions) and last 3 `wip` commits.

The only thing i don't really like is that adding a dep on `github.com/cilium/tetragon/api/v1/tetragon` in the `v1alpha1` package (https://github.com/FedeDP/tetragon/pull/28/changes/eb4aeff4e708861b24628801f85f78fd9856284e#diff-d21a1eb20132f75090ef83b70e1c4a20cc100b0e561d416b9987ee55a15fefd5) vendors the `github.com/cilium/tetragon/api/v1/tetragon` folder plus its deps. Probably not a big deal but wanted to point that out.

@mtardy let me know if you are ok with the changes, i'll correctly squash them and push them over here.
Thank you so much!

mtardy: while it looks a bit cleaner because we don't have dangling strings all around the places, I think this:

> The only thing i don't really like is that adding a dep on github.com/cilium/tetragon/api/v1/tetragon in the v1alpha1 package (https://github.com/FedeDP/tetragon/commit/eb4aeff4e708861b24628801f85f78fd9856284e#diff-d21a1eb20132f75090ef83b70e1c4a20cc100b0e561d416b9987ee55a15fefd5) vendors the github.com/cilium/tetragon/api/v1/tetragon folder plus its deps. Probably not a big deal but wanted to point that out.

Is bad and we should avoid it for now. Maybe we use the enum elsewhere and in the k8s API you just return a const with a comment (like we do with the string currently) so that we avoid the extra import?

FedeDP: Mmh the problem is that if we return an integer, we don't satisfy the interface (`TpDomain() tetragon.TracingPolicyDomain`), right?

mtardy: yeah indeed, it's a bit sad but maybe we can stay with the strings...

FedeDP: Do you think that vendoring api/v1 under v1alpha1 is that bad?

mtardy: yeah it feels off to me, especially for just an enum int, bringing the full gRPC dependency into our k8s type definition for an int is weird.

FedeDP: Backported a small fix from https://github.com/FedeDP/tetragon/pull/28 (moved the filtering of private-internal `sensorsDomain` directly in the `listDomains` handler instead of in tetra).
The PR is good now.

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
kkourt: We are already using them, no?

> 	flags.StringVarP(&domain, "domain", "", "", "Domain to be used. Use k8s to act on CRD policies. By default only acts against grpc domain.")

FedeDP: Well, `tetra tracingpolicy` just takes a `string` domain; since the `domain` param is mostly for debug purposes, i'd go the other way and mark `domain` flag hidden.

mtardy (approved):
Yeah thanks, with the enum approach it would have been more static and thus not needing to list domains and everything, but who knows, maybe someday we will add dynamic domain and this will be possible since we went with strings! Anyway we tried and the enum approach was bad because of ... reasons in the above threads :)!

Let's do it like that!

mtardy:
maybe it's worth documenting also?

FedeDP:
> Yeah thanks, with the enum approach it would have been more static and thus not needing to list domains and everything, but who knows, maybe someday we will add dynamic domain and this will be possible since we went with strings! Anyway we tried and the enum approach was bad because of ... reasons in the above threads :)!

Yes, having tried that path, i feel actually bad throwing it all away since it was clearly a better design, even if, as you noted, less dynamic.

> maybe it's worth documenting also?

Do you mean updating `docs/`? Yes i think it makes sense. I am trying to find the right document :D

mtardy:
> > Yeah thanks, with the enum approach it would have been more static and thus not needing to list domains and everything, but who knows, maybe someday we will add dynamic domain and this will be possible since we went with strings! Anyway we tried and the enum approach was bad because of ... reasons in the above threads :)!
> 
> Yes, having tried that path, i feel actually bad throwing it all away since it was clearly a better design, even if, as you noted, less dynamic.

at least we tried, this has its advantages also!

> 
> > maybe it's worth documenting also?
> 
> Do you mean updating `docs/`? Yes i think it makes sense. I am trying to find the right document :D

Yeah, not sure where it would be relevant to talk about this. We can merge this kind of feature I think, might be worth a note somewhere on "why you don't see all the policies" if we have some documentation about listing policies maybe? idk exactly

FedeDP:
I think i found the right place :) PTAL latest commit, thanks!

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
kkourt: I don't mind having it hidden or not, but I would have this as:
```go
flags.StringVarP(&domain, "domain", "", "",  fmt.Sprintf("Domain to be used. Use %s to act on CRD policies. By default only acts against %s domain.", tpdomains.K8s, tpdomains.Grpc))
```

This is not so much because the variable can change, but it's easier to go back and forth between packages that use this (e.g., using gopls symbol references). In other words, I find it convenient to use `tpdomains.K8s` to see all the places that this is used.

kkourt: In any case, that's a nit.

See my original message:
> Left some nits, but it's up to you if you want to address them.

PR #4812, opened 2026-03-31
rthooks: add --fail-allow-namespaces-regex flag

PhilipSchmid:
## Description

Add `--fail-allow-namespaces-regex` to `tetragon-oci-hook`, accepting RE2 regex patterns for namespaces that should not cause the hook to fail when Tetragon is unreachable.

Patterns use substring matching by default; use `^` and `$` anchors for full-string matching:
- `^kube-.*$` — prefix match
- `^.*-system$` — suffix match
- `^foo-.*-bar$` — prefix and suffix match
- `.*` — matches any namespace

The flag can be used standalone or combined with the existing `--fail-allow-namespaces` exact-match list. A container is allowed if its namespace matches either an exact entry or any pattern. `--fail-cel-expr` takes precedence over both flags. Both flags are evaluated through a single CEL program using CEL's native `.matches()` function, keeping all fail-check logic within the existing CEL evaluation path.

If the namespace annotation key is absent from the container annotations the hook always fails (safe default). Invalid patterns cause a CEL runtime error which also falls back to the safe default. This is consistent with how Tetragon handles CEL expressions at other places — expressions are not pre-validated against an external regex engine at startup (invalid inputs surface as CEL compile or runtime errors).

## Changes

- `cel.go`: fix two pre-existing bugs in `celUserExpr` (unchecked `cel.NewEnv` error, wrong error variable in compile error path); implement `celAllowNamespacesWithPatterns` using CEL's `.matches()` for regex pattern evaluation alongside exact namespace matching
- `main.go`: add `FailAllowNamespacesRegex` to `cliConf`; pass it to `celAllowNamespacesWithPatterns`
- `cel_test.go`: table-driven tests covering exact namespaces, regex patterns (prefix, suffix, combined anchors, substring, unanchored), combined usage, missing annotation keys, and annotation key priority
- `values.yaml` + `_container_rthooks.tpl`: add `rthooks.failAllowNamespacesRegex` Helm value; pass each entry as a repeated flag to the hook binary
- `runtime-hooks.md`: document the new option, pattern semantics, usage examples, and precedence rules

## Motivation

Previously, the only way to exclude a group of namespaces from hook failures was `--fail-cel-expr`, which requires users to write a raw CEL expression. The new flag covers common patterns without requiring any CEL knowledge, using standard RE2 regex that integrates naturally with the existing CEL infrastructure:

```yaml
rthooks:
  enabled: true
  failAllowNamespacesRegex:
    - "^kube-.*$"
    - "^prod-.*$"
    - "^.*-system$"
```

## Release Note
```release-note
rthooks: add --fail-allow-namespaces-regex flag
```


contrib/tetragon-rthooks/cmd/oci-hook/main.go
```diff
@@ -439,6 +439,17 @@ func main() {
...
+	case cliConf.FailCelUser != "":
+		log.Info("fail check: using user-provided CEL expression", "expr", cliConf.FailCelUser)
+	case len(cliConf.FailAllowNamespacesPatterns) > 0:
+		log.Info("fail check: using namespace allow-list with patterns",
+			"namespaces", cliConf.FailAllowNamespaces,
+			"patterns", cliConf.FailAllowNamespacesPatterns,
+		)
+	default:
+		log.Info("fail check: using namespace allow-list", "namespaces", cliConf.FailAllowNamespaces)
+	}
```
kkourt: Why is this added?

contrib/tetragon-rthooks/cmd/oci-hook/main.go
```diff
@@ -362,15 +363,14 @@ func checkFail(log *slog.Logger, prog *celProg, annotations map[string]string) e
...
-		ret, err = celUserExpr(expr)
-	} else {
-		ret, err = celAllowNamespaces(cliConf.FailAllowNamespaces)
+		return celUserExpr(expr)
+	}
+	if len(cliConf.FailAllowNamespacesPatterns) > 0 {
+		return celAllowNamespacesWithPatterns(cliConf.FailAllowNamespaces, cliConf.FailAllowNamespacesPatterns)
 	}
-	return ret, err
+	return celAllowNamespaces(cliConf.FailAllowNamespaces)
```
kkourt: This does not make sense to me. 

We either:
* use CEL expressions for everything, or 
* move the --fail-allow-namespaces code to use the open-ended function you introduced, and then add support for patterns.

docs/content/en/docs/installation/runtime-hooks.md
```diff
@@ -213,18 +213,50 @@ To understand these messages, consider what `tetragon-oci-hook` should do if it
 cannot contact the Tetragon agent.
 
 You may want to stop certain workloads from running. For other workloads (for example, the
-tetragon pod itself) you probably want to do the opposite and let the them start. To this end,
+tetragon pod itself) you probably want to do the opposite and let them start. To this end,
```
kkourt, line 216: Please split fixes on existing code in different patches.

kkourt (changes requested):
Thanks!

Can you split the PR into different commits so that it can be easily reviewed?

It feels to me that the PR was generated by an LLM. Please document the influence level (see: https://danielmiessler.com/blog/ai-influence-level-ail) of the PR.

docs/content/en/docs/installation/runtime-hooks.md
PhilipSchmid: I first thought doing a separate git commit for a single word removal might be an overkill (the rest of the changes in `### Failure check (failAllowNamespaces and failAllowNamespacesPatterns)` are directly related to this PR), but sure, will do 😄.

contrib/tetragon-rthooks/cmd/oci-hook/main.go
PhilipSchmid: Good point! The original idea was to show the user the current fail check configuration against we're checking the to-be-started containers. However, back then, I didn't realize this would get triggered on every container start instead instead of just during startup of the `tetragon-rthook` DaemonSet (`/setup/main.go` != `oci-hook/main.go` 😅). Here it just brings a lot of noise (at `Info` log level) for every container being started. Also, the complete config is already printed a few lines above (at `Debug` log level — `log.Debug("starting", "conf", fmt.Sprintf("%+v", cliConf))`). Removed it.

PhilipSchmid (commented):
Thanks for the feedback and input!

> Can you split the PR into different commits so that it can be easily reviewed?

Absolutely, I just did that and force-pushed 4 granular commits.

> It feels to me that the PR was generated by an LLM. Please document the influence level (see: https://danielmiessler.com/blog/ai-influence-level-ail) of the PR.

💯, I used an LLM create that. I'd say it's an AIL 3-4. IMO, it's probably obvious that close to no human would have added so many unit tests 😅. Still, stuff like this was all implemented because I gave pretty detailed instructions to the LLM to do so, and I also reviewed them all and they make sense to me (according to me limited golang experience). Is there a specific way you want this AIL being mentioned in the PR and/or git commits? Maybe it's also worth to add a brief section about that in https://tetragon.io/docs/contribution-guide/ as I assume I'm not the first and won't for sure be the last one doing that?

contrib/tetragon-rthooks/cmd/oci-hook/main.go
PhilipSchmid: Fair point! This complicates things unnecessarily in the code. Honestly, I should likely also not over-engineer with regex either, since we (for now) only need to allow patterns like `namespace-*` (or `*-namespace`), suffix/prefix wildcard matching should be sufficient. So, keeping everything within CEL would probably be best.

PhilipSchmid: Just changed it to a pure CEL-based approach.

docs/content/en/docs/installation/runtime-hooks.md
```diff
@@ -219,12 +219,50 @@ belongs in the same namespace as Tetragon. The previous messages concern the tet
...
+  enabled: true
+  failAllowNamespaces: kube-system
+  failAllowNamespacesPatterns:
+    - "acme-*"
+    - "dev-*"
+```
+
+> **Note:** If the pod namespace annotation is missing from the container annotations, the hook
+> will always fail the container (safe default). Both options are ignored when
+> `extraHookArgs` sets `fail-cel-expr`, which always takes precedence.
```
mtardy: please use the actual `note` docs block

docs/content/en/docs/installation/runtime-hooks.md
```diff
@@ -219,12 +219,50 @@ belongs in the same namespace as Tetragon. The previous messages concern the tet
...
+```yaml
+rthooks:
+  enabled: true
+  failAllowNamespacesPatterns:
+    - "acme-*"
+    - "dev-*"
+```
+
+The two options can be combined. A container is allowed to start if its namespace matches
+**either** an exact name **or** any pattern:
```
mtardy: nit: not sure you need the bold here

contrib/tetragon-rthooks/cmd/oci-hook/cel.go
```diff
@@ -6,15 +6,39 @@ package main
...
 type celProg struct {
 	p      cel.Program
 	values map[string]interface{}
 }
 
+// namespaceMatchesGlob reports whether ns matches the glob pattern p.
+// The only supported wildcard is a single '*', which matches any sequence of
+// characters. Patterns with more than one '*' are rejected at construction time
+// (see celAllowNamespacesWithPatterns), so this function assumes at most one.
+func namespaceMatchesGlob(ns, p string) bool {
```
mtardy, line 25: your commit description says you fixed something in a function that you just added? I'm a bit confused.

> Fix a false-match bug in namespaceMatchesGlob for contains patterns
> where the prefix and suffix overlap on short namespace strings.

mtardy (commented):
I think I was mostly pulled for the docs change and I don't much about how these flags works (and how exactly you can use `--fail-cel-expr`) but couldn't we just adapt this so that it uses CEL `.matches`

https://github.com/cilium/tetragon/blob/3613098d6f27e0e39c4569ed36da2c57fa5d4364/contrib/tetragon-rthooks/cmd/oci-hook/cel.go#L42-L71

Or even now, what can a user already do with `--fail-cel-expr`?

I feel like we could define the namespace env variable within the user expr[^1] and that someone could reuse that variable with CEL `.matches` to do it?

Maybe I'm completely missing something I didn't spend so much time getting 100% of how we use CEL here.

[^1]: it seems it's even already the case https://github.com/cilium/tetragon/blob/3613098d6f27e0e39c4569ed36da2c57fa5d4364/contrib/tetragon-rthooks/cmd/oci-hook/cel.go#L21

docs/content/en/docs/installation/runtime-hooks.md
PhilipSchmid: Ah, sure, that makes sense. I just changed it to a `{{< note >}}` block.

docs/content/en/docs/installation/runtime-hooks.md
PhilipSchmid: The "either/or" feels important to me, but I agree that we should avoid using bold here to maintain consistency with the rest of the documentation.

contrib/tetragon-rthooks/cmd/oci-hook/cel.go
PhilipSchmid: Ah, damn, right. That was a git commit message leftover from an earlier version. I removed it. Thanks!

PhilipSchmid:
Thanks a lot for the review/input, @mtardy!

> I think I was mostly pulled for the docs change and I don't much about how these flags works (and how exactly you can use `--fail-cel-expr`) but couldn't we just adapt this so that it uses CEL `.matches`
> 
> https://github.com/cilium/tetragon/blob/3613098d6f27e0e39c4569ed36da2c57fa5d4364/contrib/tetragon-rthooks/cmd/oci-hook/cel.go#L42-L71

Generally speaking, yes, CEL with `.matches` would work. It would be similar to my initial approach, which allowed full (RE2) regular expressions. However, I think we should try to keep the user experience simple by only allowing globbing and not full regular expressions.

> Or even now, what can a user already do with `--fail-cel-expr`?

Technically, it would be possible to use the `--fail-cell-expr` option to achieve the same result. However, it is not intuitive. The user would need to understand the internal workings, and the option is prone to errors. Here is an example for comparison:

```
rthooks:
  enabled: true
  interface: "nri-hook"
  extraHookArgs:
    fail-cel-expr: >-
      !(annotations.exists(k, k in ["io.kubernetes.pod.namespace",
      "io.kubernetes.cri.sandbox-namespace"] && (annotations[k] == "kube-system" ||
      annotations[k].matches("^acme-") ||
      annotations[k].matches("^dev-"))))
```

vs.

```
rthooks:
  enabled: true
  failAllowNamespaces: kube-system
  failAllowNamespacesPatterns:
    - "acme-*"
    - "dev-*"
```

contrib/tetragon-rthooks/cmd/oci-hook/cel.go
```diff
@@ -39,21 +63,59 @@ func celUserExpr(expr string) (*celProg, error) {
...
 	}
 
-	//expr := `!(annotations_namespace_key in annotations && annotations[annotations_namespace_key] in allow_labels)`
-	expr := `annotations_namespace_keys.all(key, !(key in annotations && annotations[key] in allow_labels))`
+	// The expression fails (returns true) unless a matching namespace key is
+	// found AND the namespace is in the exact allow list or matches a glob.
+	expr := `annotations_namespace_keys.all(key,
+		!(key in annotations && (
+			annotations[key] in allow_namespaces ||
+			allow_patterns.exists(p, namespace_matches_glob(annotations[key], p))
```
kkourt: As Mahé already said, why can't we just use CEL's `.matches`?
See https://github.com/google/cel-spec/blob/master/doc/langdef.md#list-of-standard-definitions

kkourt (changes requested):

kkourt:
> Generally speaking, yes, CEL with `.matches` would work. It would be similar to my initial approach, which allowed full (RE2) regular expressions. However, I think we should try to keep the user experience simple by only allowing globbing and not full regular expressions.

Why? using `.matches` has a much simpler implementation.
Also,  the `glob` implementation in this PR does not seem correct to me. See https://man7.org/linux/man-pages/man7/glob.7.html.

contrib/tetragon-rthooks/cmd/oci-hook/cel.go
PhilipSchmid: When switching to `.matches`, which by default does [substring matching using RE2](https://github.com/google/cel-spec/blob/master/doc/langdef.md#regular-expressions), what configuration option would you like to expose to the user?

1. Fully blown RE2 pattern capability?
2. "Glob-**style**" like `*` matching? Basically what it currently supports via the non-ideal implementation. (Supporting a single `*` wildcard: `kube-*` (prefix), `*-system` (suffix), `foo-*-bar` (contains), or `*` alone to match any namespace.)
3. Full glob support like described in https://man7.org/linux/man-pages/man7/glob.7.html?
4. Something else?

Personally, I'd still vote for number 2 from an UX perspective, but this would then require some "glob-style" pattern to RE2 pattern parsing, right? For example, to ensure it's explicitly anchored using `^` and `$`, so that a `failAllowNamespacesPatterns` set to `infra-*` wouldn't match `blub-infra-system` but only `infra-system` (besides mapping `*` to `.*`).

kkourt: Why is RE2 not preferable from a UX perspective? It seems to me that it's the simpler implementation since CEL already supports it, and I don't understand what the downside is.

I'm not particularly fond of 2. because it's something that we need to define in an ad-hoc manner, both in terms of the implementation and also in terms of the pattern matching semantics.

PhilipSchmid: > Why is RE2 not preferable from a UX perspective? It seems to me that it's the simpler implementation since CEL already supports it, and I don't understand what the downside is.

My original thought was that it offers flexibility that we don't need since we're only matching K8s NS names, which can only be in [a specific format](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names). As often times, flexibility brings complexity, and as we don't need a majority the `.matches`' flexibility, we could also avoid the complexity. In this case, it's the complexity for the users because writing correct RE2 syntax is, in my opinion, more complex and error-prone than the introduced simplified syntax.

> it's something that we need to define in an ad-hoc manner, both in terms of the implementation and also in terms of the pattern matching semantics

True that, it's increasing the complexity on our end — the Tetragon implementation. Additionally, the introduced pattern-matching semantics can be considered "random" since it's not a complete implementation of a well-known principle, such as RE2, globbing, or an RFC. Perhaps we can find a middle ground by exposing the `.matches` method to users while also implementing guardrails to reduce the likelihood of misconfigurations.

* Hints in the Helm values commends
* Some notes about potential pitfalls in the docs
* Some basic extra input validation and logging so that a user is more likely to see the potentially unintended behavior

I'll switch it to the `.matches` approach and try to some up with some basic things for the bullet points mentioned above.

PhilipSchmid: Okay, I just force pushed such a new version. One note on the three mentioned bullet points: I didn't really include the 3rd one in order to be more consistent with how Tetragon handles CEL expressions at other places. There, expressions are also not pre-validated against an external regex engine at startup (invalid inputs surface as CEL compile or runtime errors).

contrib/tetragon-rthooks/cmd/oci-hook/cel.go
```diff
@@ -42,21 +42,38 @@ func celUserExpr(expr string) (*celProg, error) {
...
+//   - Namespace matches an exact allow entry  -> do not fail (false)
+//   - Namespace matches a regex pattern       -> do not fail (false)
+//   - Namespace matches neither               -> fail (true)
+func celAllowNamespacesWithPatterns(namespaces []string, patterns []string) (*celProg, error) {
 	env, err := cel.NewEnv(
 		cel.Variable("annotations", cel.MapType(cel.StringType, cel.StringType)),
 		cel.Variable("annotations_namespace_keys", cel.ListType(cel.StringType)),
-		cel.Variable("allow_labels", cel.ListType(cel.StringType)),
+		cel.Variable("allow_namespaces", cel.ListType(cel.StringType)),
+		cel.Variable("allow_patterns", cel.ListType(cel.StringType)),
```
kkourt: We are still matching namespaces here, right? How about `allow_namepace_patterns`?

contrib/tetragon-rthooks/cmd/oci-hook/cel.go
```diff
@@ -67,8 +84,9 @@ func celAllowNamespaces(vals []string) (*celProg, error) {
 	return &celProg{
 		p: p,
 		values: map[string]interface{}{
-			"allow_labels":               vals,
 			"annotations_namespace_keys": cliConf.AnnNamespaceKeys,
+			"allow_namespaces":           namespaces,
+			"allow_patterns":             patterns,
```
kkourt: ditto here for the name

contrib/tetragon-rthooks/cmd/oci-hook/cel_test.go
```diff
@@ -6,48 +6,265 @@ package main
...
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
+// defaultAnnKeys mirrors the production default for annotation namespace keys.
+var defaultAnnKeys = []string{
+	"io.kubernetes.pod.namespace",
+	"io.kubernetes.cri.sandbox-namespace",
+}
```
kkourt: How about we define 

```go
defaultAnnotationsNamespaceKeys = []string{
	"io.kubernetes.pod.namespace",
	"io.kubernetes.cri.sandbox-namespace",
}
```

in main and re-use it for tests?

We can have:
```go
		kong.Vars{
			"defAnnotationsNamespaceKeys": strings.Join(",", defaultAnnotationsNamespaceKeys),
		},
```

contrib/tetragon-rthooks/cmd/oci-hook/cel_test.go
```diff
@@ -6,48 +6,265 @@ package main
...
+// annPod builds an annotation map with the pod.namespace key (cri-o style).
+func annPod(ns string) map[string]string {
+	return map[string]string{"io.kubernetes.pod.namespace": ns}
+}
+
+// -----------------------------------------------------------------------
+// TestCelAllowNamespaces - exact-match behaviour
+// -----------------------------------------------------------------------
+
+func TestCelAllowNamespaces(t *testing.T) {
```
kkourt, line 58: Can you please separate the tests in a different commit?

kkourt (changes requested):
Just a few nits, but LGTM, thanks!

contrib/tetragon-rthooks/cmd/oci-hook/cel.go
PhilipSchmid: Absolutely, changed!

contrib/tetragon-rthooks/cmd/oci-hook/cel_test.go
PhilipSchmid: Makes sense, changed.

install/kubernetes/tetragon/templates/_container_rthooks.tpl
```diff
@@ -15,6 +15,9 @@
     - --grpc-address={{ .Values.tetragon.grpc.address }}
     - --fail-allow-namespaces
     - {{ if .Values.rthooks.failAllowNamespaces }}{{ printf "%s,%s" .Release.Namespace .Values.rthooks.failAllowNamespaces }}{{ else }}{{ .Release.Namespace }}{{ end }}
+   {{- range .Values.rthooks.failAllowNamespacesRegex }}
+    - --fail-allow-namespaces-regex={{ . }}
+   {{- end }}
```
kkourt, line 20: I'm fine with keeping this, but keep in mind that if someone specifies it, it will fail until we tag a new rthooks image and update values.yaml: https://github.com/cilium/tetragon/blob/a46b00ecb53a9ac2a3cb1121a7cd838ae1ba9a41/install/kubernetes/tetragon/values.yaml#L473-L474

kkourt (approved):
LGTM, thanks!

install/kubernetes/tetragon/templates/_container_rthooks.tpl
PhilipSchmid: Good point! Are you aware of any other rthooks-related activities/changes we need to wait for before creating a new `quay.io/cilium/tetragon-rthooks` release right after merging this PR?

kkourt: Nope, we can just tag once we merge.

contrib/tetragon-rthooks/cmd/oci-hook/cel_test.go
```diff
@@ -6,48 +6,259 @@ package main
...
+}
+
+// annPod builds an annotation map with the pod.namespace key (cri-o style).
+func annPod(ns string) map[string]string {
+	return map[string]string{"io.kubernetes.pod.namespace": ns}
+}
+
+// -----------------------------------------------------------------------
+// TestCelAllowNamespaces - exact-match behaviour
+// -----------------------------------------------------------------------
```
mtardy: I'm not a big fan of these detached comments, I know LLM are producing this, wondering why. If you feel like documenting the func, please add normal comments like
```go
// TestCelAllowNamespaces - exact-match behaviour
func TestCelAllowNamespaces(t *testing.T) {
...
```

Hopefully tests names should document themselves

contrib/tetragon-rthooks/cmd/oci-hook/cel_test.go
```diff
@@ -6,48 +6,259 @@ package main
...
+	nsKubeSystem = "kube-system"
+	nsDefault    = "default"
+	nsMonitor    = "monitoring"
+
+	// RE2 regex pattern examples. Anchored with ^ and $ for full-string matching.
+	patternAcme = "^acme-.*$"
+	patternDev  = "^dev-.*$"
+)
+
+// helpers ----------------------------------------------------------------
```
mtardy: same for here, I think it's useless comments, we already call `t.Helper()` and it's obvious.

mtardy (approved):
Thanks I think it looks good as well like this. Just small nits on comments.

Sorry for the delay!

contrib/tetragon-rthooks/cmd/oci-hook/cel_test.go
PhilipSchmid: Sounds reasonable. Changed in the latest git force push 😄.

install/kubernetes/tetragon/templates/_container_rthooks.tpl
PhilipSchmid: Created a release PR here: https://github.com/cilium/tetragon/pull/5014

PR #4813, opened 2026-03-31
cleanup(pkg): deny calls to already-deprecated gRPC methods.

FedeDP:
### Description

The methods:
* EnableTracingPolicy
* DisableTracingPolicy

are marked as deprecated in our [proto files](https://github.com/cilium/tetragon/blob/main/api/v1/tetragon/sensors.proto#L286); deny their usage.

Also, introduce a new option to smooth the transition:
* `enable-deprecated-tracingpolicy-grpc` to enable tp-related deprecated grpc methods

### Changelog

```release-note
cleanup(pkg): deny calls to already-deprecated gRPC methods.
```


FedeDP:
While we are at it, what about dropping the already-deprecated `tetra sensors`? https://github.com/cilium/tetragon/blob/main/cmd/tetra/sensors/sensors.go

It has been deprecated in 2023: https://github.com/cilium/tetragon/commit/ba8b45466dea666ea94ade00e59018b8571ab3a4

cc @kkourt

PR #4814, opened 2026-03-31
Add support for spec.hostSelector

tpapagian:
This PR adds support for `spec.hostSelector` to enable users to write policies that are applied only to host workloads and/or pod workloads (with additional filtering for the containers if needed).

The syntax is:
```yaml
spec:
  hostSelector: {}
  podSelector: {}
  containerSelector: {}
```
 and the previous example will match all workloads either in the host or inside pods without any filtering. This is also the default (i.e. if the user does not define any of those in a tracing policy).

If we need to match only host workloads we need to have something like:
```yaml
spec:
  hostSelector: {}
  podSelector: null
  containerSelector: null
```

For now we only support 2 types of `hostSelector`. These are the `{}` (everything) and `null` (nothing). Specific filters are not supported yet.

In order to make that work, we need to make some changes in the behaviour of the `podSelector`. Currently, `podSelector: {}` and `podSelector: null` are exactly the same thing and will match all workloads (including host workloads). Based on this behaviour, now we don't have a way to match on all pod workloads (excluding host workloads).

In this PR we change that behaviour. Now `podSelector: {}` will match all pod workloads and `podSelector: null` will not match any pod workloads. The behaviour of `containerSelector` is similar, but what actually does is a second level filtering on the `podSelector`. For this reason:
```yaml
spec:
  hostSelector: ~
  podSelector: ~
  containerSelector:
    matchExpressions:
      - key: name
        operator: In
        values:
        - ubuntu
```
will not match anything as podSelector is null. We need to have something like 
```yaml
spec:
  hostSelector: ~
  podSelector: {}
  containerSelector:
    matchExpressions:
      - key: name
        operator: In
        values:
        - ubuntu
```
to filter containers.

This PR also provides the ability to make `hostSelector`, `podSelector`, and `containerSelector` equals to `null` which was not possible before. 

## Implementation

In order to make that work, this PR adds one more entry in `policy_filter_maps` with key equals to `ALL_PODS_POLICY_ID` (`UINT32_MAX`). The inner map of that entry contains all `cgroup_id`s of all pods in the system. If a `cgroup_id` is part of that set, this means that it is a pod workload. If not this means that this is part of the host workload. Additionally, each policy contains an invalid `cgroup_id` entry with value `HOST_SELECTOR_MODE` (`UINT64_MAX`) if the policy itself cares about host workloads. 

Then we update `policy_filter_check` to check those as well. More specifically what policy_filter_check does is:
1. check if our cgroup_id match on the policy build from podSelector/containerSelector
    1. return true if we match
    2. continue to check if we care about host workloads
2. check if the policy cares also for host workloads
    1. return false if not
    2. continue to check if our cgroup_id is a host workload  
3. check if our cgroup_id is part of ALL_PODS_POLICY_ID
    1. return true if not (this means that out cgroup_id is not a pod workload --> so this is a host workload)
    2. return false if it is (this means that out cgroup_id is a pod workload)

Any updates needed for the policy_filter_maps are handled from the agent in a similar way to podSelector/containerSelector.

## Testing

All tests related to the policyfiler have been updated to reflect those changes. Some additional tests have been added to check the hostSelector explicitly.

## Limitations

This approach has similar issues with [Kubernetes Identity Aware Policies](https://tetragon.io/docs/concepts/tracing-policy/k8s-filtering/) and more specifically if we rely on the k8s API server, there may be a race on when we add a cgroup_id to our maps and when the pod actually starts. The hostSelector has similar issues. All of these issues can be resolved by using [Runtime Hooks](https://tetragon.io/docs/concepts/runtime-hooks/) instead of the K8s API server.

## Alternatives

Another way to do that is to use [bpf_current_task_under_cgroup](https://docs.ebpf.io/linux/helper-function/bpf_get_current_ancestor_cgroup_id/) or [bpf_get_current_ancestor_cgroup_id](https://docs.ebpf.io/linux/helper-function/bpf_get_current_ancestor_cgroup_id/) eBPF helpers. This allows us to check if a `cgroup_id` is somewhere under a specific cgroup subtree. If this specific cgroup subtree is where k8s adds new pods, then we can achieve the same results. The challenge here is to find the `cgroup_id` that k8s use to keep all pods. In most of the cases this is `/sys/fs/cgroup/kubepods.slice/` but this is not always the case in all deployments. It seems that there are ways to get that inside Tetragon but there may be a bit complicated. 

For this reason, we are starting with the approach proposed in this PR. It works in all cases, although it introduces a small increase in memory usage. At this point, the impact does not seem significant enough to justify a more complex design. We can always improve it in future PRs if needed.

### DIFF
--- a/bpf/process/policy_filter.h
+++ b/bpf/process/policy_filter.h
@@ -10,11 +10,22 @@
 #define POLICY_FILTER_MAX_POLICIES   128
 #define POLICY_FILTER_MAX_CGROUP_IDS 1024
 
+#define ALL_PODS_POLICY_ID 0x01
+#define HOST_SELECTOR_MODE 0xFFFFFFFFFFFFFFFFull
+
+// In order to implement the hostSelector we add one more entry in the outer map
+// that is not related to any specific policy. This entry has policy_id equals
+// to ALL_PODS_POLICY_ID. In that case the inner map contains the cgroup_ids for
+// *all* containers inside *all* pods. This allows us to generate a mechanism to
+// match (i) on all pods or (ii) in none of the pods (which is the same as the
+// host workload).
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH_OF_MAPS);
 	__uint(max_entries, POLICY_FILTER_MAX_POLICIES);
 	__type(key, u32); /* policy id */
 	__array(
+		// If a specific policy needs to match on host workloads as well we also
+		// add an entry with key HOST_SELECTOR_MODE (UINT64_MAX).
 		values, struct {
 			__uint(type, BPF_MAP_TYPE_HASH);
 			__uint(max_entries, 1);
@@ -63,7 +74,23 @@ FUNC_INLINE bool policy_filter_check(u32 policy_id)
 	if (trackerid)
 		cgroupid = trackerid;
 
-	return map_lookup_elem(policy_map, &cgroupid);
+	if (map_lookup_elem(policy_map, &cgroupid))
+		return true; // We have a match from the podSelector and/or the containerSelector.
+
+	// We didn't match on the podSelector and/or the containerSelector.
+	// Now we need to check if we have a hostSelector match.
+
+	trackerid = HOST_SELECTOR_MODE;
+	if (!map_lookup_elem(policy_map, &trackerid))
+		return false; // Cannot find the match mode of the hostSelector so we do not care to match any host workloads.
+
+	policy_id = ALL_PODS_POLICY_ID;
+	policy_map = map_lookup_elem(&policy_filter_maps, &policy_id);
+	if (!policy_map)
+		return false; // Cannot find the cgroupids of all containers inside all pods. This should not happen.
+
+	// If !map_lookup_elem(policy_map, &cgroupid) then our cgroupid belongs to a host workload.
+	return !map_lookup_elem(policy_map, &cgroupid);
 }
 
 #endif /* POLICY_FILTER_MAPS_H__ */
--- a/docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
+++ b/docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
@@ -49,6 +49,67 @@ the policy is applied to.
 
 For container field filters, we use the `containerSelector` field of tracing policies to select the containers that the policy is applied to. At the moment, the only supported fields are `name` and `repo` which refers to the container repository.
 
+## Host workload filters
+
+To filter host workloads we use the `hostSelector` field of tracing policies to select if a policy
+should be applied to host workloads or not. For now this only supports `{}` to match all host workloads
+and `null` to match none of the host workloads.
+
+Based on these, the user can filter out specific workloads.
+
+## Examples
+
+By default, a policy match on all workloads, similar to the following example:
+
+```yaml
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "workload-filtering"
+spec:
+  hostSelector: {}
+  podSelector: {}
+  containerSelector: {}
+  kprobes:
+```
+
+This is equivalent as if the user does not define any of `hostSelector`, `podSelector`, and
+`containerSelector`.
+
+The following example will match only host workloads.
+
+```yaml
+spec:
+  hostSelector: {}
+  podSelector: null
+  containerSelector: null
+```
+
+The following example will match only pod workloads.
+
+```yaml
+spec:
+  hostSelector: null
+  podSelector: {}
+  containerSelector: {}
+```
+
+`containerSelector` acts as a second level filtering on the `podSelector`. This means that first the
+`podSelector` is evaluated and if pod match we then apply the `containerSelector`. The following example
+will match all pods inside the `kube-system` namespace and all host workloads.
+
+```yaml
+spec:
+  hostSelector: {}
+  podSelector:
+    matchExpressions:
+    - key: "k8s:io.kubernetes.pod.namespace"
+      operator: In
+      values:
+      - "kube-system"
+  containerSelector: {} # this can be also omitted as the default value is {}
+```
+
 ## Demo
 
 ### Setup
--- a/pkg/policyfilter/map.go
+++ b/pkg/policyfilter/map.go
@@ -84,6 +84,19 @@ func newPfMap(enableCgroupMap bool) (PfMap, error) {
 		}
 	}
 
+	// Create the reserved entry once at initialization time.
+	allPodsMap, err := ret.newPolicyMap(AllPodsID, nil)
+	if err != nil {
+		ret.release()
+		return PfMap{}, fmt.Errorf("opening all-pods policy map failed: %w", err)
+	}
+
+	// We just initialized allPodsMap but we don't want that now so let's close that.
+	if err := allPodsMap.Inner.Close(); err != nil {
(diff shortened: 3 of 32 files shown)


bpf/process/policy_filter.h
```diff
@@ -10,11 +10,22 @@
 #define POLICY_FILTER_MAX_POLICIES   128
 #define POLICY_FILTER_MAX_CGROUP_IDS 1024
 
+#define ALL_PODS_POLICY_ID 0xFFFFFFFFul
+#define HOST_SELECTOR_MODE 0xFFFFFFFFFFFFFFFFull
```
mtardy: With the way it's structured, you could have used `1` instead of UINT32/64 max. Doesn't make a huge difference but maybe is cleaner?

https://github.com/cilium/tetragon/blob/3613098d6f27e0e39c4569ed36da2c57fa5d4364/pkg/sensors/handler.go#L36-L41

pkg/policyfilter/map.go
```diff
@@ -84,6 +84,19 @@ func newPfMap(enableCgroupMap bool) (PfMap, error) {
 		}
 	}
 
+	// Create the reserved UINT32_MAX entry once at initialization time.
+	allPodsMap, err := ret.newPolicyMap(AllPodsPolicyID, nil)
+	if err != nil {
+		ret.release()
```
mtardy, line 90: before they call `releaseMap(ret.PolicyMap)` function, do you want to follow that pattern here as well?

pkg/policyfilter/state.go
```diff
@@ -401,10 +401,14 @@ func (m *state) delPod(id PodID) *podInfo {
...
 // AddPolicy adds a policy
 func (m *state) AddPolicy(polID PolicyID, namespace string, podLabelSelector *slimv1.LabelSelector,
-	containerLabelSelector *slimv1.LabelSelector) error {
+	containerLabelSelector *slimv1.LabelSelector, hostLabelSelector *slimv1.LabelSelector) error {
 	m.mu.Lock()
 	defer m.mu.Unlock()
 
+	if polID == AllPodsPolicyID {
+		return fmt.Errorf("policy id %d is reserved for the all-pods map entry", polID)
+	}
```
mtardy: so that you might be able to even drop that check

pkg/sensors/k8s.go
```diff
@@ -25,30 +27,48 @@ func (h *handler) updatePolicyFilter(tp tracingpolicy.TracingPolicy, tpID uint64
...
+	// matches everything        | tp.TpSpec().ContainerSelector != nil && (len(ps.MatchLabels) + len(ps.MatchExpressions) == 0)
+	// matches based on selector | tp.TpSpec().ContainerSelector != nil && (len(ps.MatchLabels) + len(ps.MatchExpressions) != 0)
+	containerSelector := tp.TpSpec().ContainerSelector
+
+	// matches nothing           | tp.TpSpec().HostSelector == nil
+	// matches everything        | tp.TpSpec().HostSelector != nil && (len(ps.MatchLabels) + len(ps.MatchExpressions) == 0)
+	// matches based on selector | Not supported yet
+	hostSelector := tp.TpSpec().HostSelector
+	if hostSelector != nil && (len(hostSelector.MatchLabels)+len(hostSelector.MatchExpressions) > 0) {
+		return policyfilter.NoFilterID, errors.New("spec.hostSelector does not support arbitrary labels. Only ~ (empy) and {} (all) is supported for now")
```
mtardy: small user facing typo `empy` to `empty`

pkg/policyfilter/state_test.go
```diff
@@ -114,3 +114,266 @@ func TestState(t *testing.T) {
 	require.Empty(t, s.policies)
 	require.Empty(t, s.pods)
 }
+
+func TestStateAllPodsPolicyEntry(t *testing.T) {
+	s, err := New(true)
+	if err != nil {
+		t.Skipf("failed to inialize policy filter state: %s", err)
```
mtardy: small typo: initialize

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
```diff
@@ -49,6 +49,63 @@ the policy is applied to.
...
+metadata:
+  name: "workload-filtering"
+spec:
+  hostSelector: {}
+  podSelector: {}
+  containerSelector: {}
+  kprobes:
+```
+
+This is the same as if the user does not define any of `hostSelector`, `podSelector`, and `containerSelector`. Based on that the user can filter out specific workloads.
```
mtardy: can you wrap the docs line plz like you did at the beginning?

mtardy: I would cut the part `Based on that the user can filter out specific workloads.` to another paragraph just before the examples.

And put `This is equivalent` instead of `This is the same`

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
```diff
@@ -49,6 +49,63 @@ the policy is applied to.
...
+The following example will match only pod workloads.
+
+```yaml
+spec:
+  hostSelector: null
+  podSelector: {}
+  containerSelector: {}
+```
+
+`containerSelector` acts as a second level filtering on the `podSelector`. This means that first the podSelector is evaluated and if pod match we then apply the `containerSelector`.
```
mtardy: in most places you used `podSelector` but here it's missing the backticks

mtardy (commented):
Here are few comments, I think this is a good patch set. I'd just like to discuss a little bit offline about the alternative you proposed and that was already discussed internally. :)

pkg/policyfilter/map.go
tpapagian: At this point, if we fail we need to release both `ret.PolicyMap` and `ret.cgroupMap`. `releaseMap(ret.PolicyMap)` will release only the first, while `ret.release()` will release both.

pkg/sensors/k8s.go
tpapagian: Fixed that.

pkg/policyfilter/state_test.go
tpapagian: Fixed that.

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
tpapagian: Fixed that.

tpapagian: Fixed that.

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
tpapagian: Fixed that.

bpf/process/policy_filter.h
tpapagian: I have changed `ALL_PODS_POLICY_ID` to 1 which make sense.

For  `HOST_SELECTOR_MODE` I want a value that will not collide with any real cgroupid. My understanding is that cgroupid 1 exists (or may exist) and this can cause issues. I would propose to keep it as it is to be on the safe side.

pkg/policyfilter/state.go
tpapagian: Updated `AllPodsPolicyID` and dropped that.

mtardy (approved):
Thank you!! Looks good

PR #4817, opened 2026-03-31
feat(uprobes): add Go string parsing and inline modification

dwindsor:
Fixes #4827 

### Description
With the pclntab changes being merged, we now can write policies that target stripped Go binaries. The problem is, pclntab doesn't contain function signature metadata, only function location data (offset + size).

This series adds support for the `go_string` type and the `clearGoString` action to `Override` in TracingPolicy. It requires a fairly complicated descent into the Go ABI internals, but I think the ABI has been stable enough for a while (since 1.18) to make this safe. Basically, since we have the function start location from pclntab, we just need to be able to isolate go string parameters, which are stored as a (`[word][u32]`) tuple (data_ptr+len). The new Go ABI (ABIInternal) is register based, with predictable slot assignments for types that haven't changed since 1.18. We go:generate the ABI slot mapping at build time, so no hard-coding.

The following figure illustrates how Go datatypes map to register slots on Go 1.18+ (ABIInternal):

ServeContent

 With this, we can use [7384893](https://lkml.org/lkml/2025/9/8/871#:~:text=LKML:%20Jiri%20Olsa:%20%5BPATCHv2,app%20registers%20from%20uprobe%20registers) `(bpf: Allow uprobe program to change context registers)` to add the `clearGoString` action to `go_string` parameters that contain suspicious (i.e. SQL injection, environment var leak, malicious AI prompt) strings.  We don't return early here, we just make the string inert by clearing it, allowing the application to handle cleanup etc. It's difficult to return from non-LSM functions (lsm gives a lot of guarantees - locks held, refcounts guaranteed, etc). 

Screenshot 2026-04-22 at 4 56 53 PM

### Changelog

```release-note
```


dwindsor:
vmtests CI failing due to a verifier error on < 6.12, will fix:

> ❌	pkg.sensors.tracing.TestUprobeGoStringArg

dwindsor:
vmtests failing on kernels that don't have `bpf_copy_from_user_str`

```
424: (85) call unknown#233685214
invalid func unknown#233685214
```

Gating `go_string` support on the presence of `bpf_copy_from_user_str`.

bpf/process/event_config.h
```diff
@@ -39,6 +39,13 @@ struct config_reg_arg {
 	__u8 pad;
 } __attribute__((packed));
 
+#define GO_ARG_MAX_REGS 3
+
+struct config_go_arg {
+	__u16 regs[GO_ARG_MAX_REGS];
+	__u16 field_off;
+} __attribute__((packed));
```
olsajiri, line 47: looks like we use only regs[0/1] so let's scratch the rest?
also is there reason to have an array for that?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -670,6 +726,9 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 		if err != nil {
 			return err
 		}
+		if argType == gt.GenericGoStringType {
```
olsajiri, line 728: is there a reason this needs to be pt_regs type? I think if we make it just ordinary type that will read the data from preconfigured registers it will be more simple

bpf/process/user_preload.h
```diff
@@ -60,11 +101,20 @@ preload_pt_regs_arg(struct pt_regs *ctx, struct event_config *config, int index)
 		     : [index] "+r"(index)
 		     : "i"(EVENT_CONFIG_MAX_REG_ARG_MASK));
 
+	ty = config->arg[index];
+
+	if (ty == go_string_type) {
```
olsajiri, line 106: as I said in other comment, let's make it an ordinary type and move this under preload_arg

pkg/sensors/tracing/goabi_slots.go
```diff
@@ -0,0 +1,52 @@
...
+	offsets, ok := goABIKnownFuncs[symbol]
+	if !ok || argIndex >= len(offsets) {
+		return -1
+	}
+	return offsets[argIndex]
+}
+
+// https://go.googlesource.com/go/+/refs/heads/master/src/cmd/compile/abi-internal.md
+var goABIIntRegs = []string{
+	"rax", "rbx", "rcx", "rdi", "rsi", "r8", "r9", "r10", "r11",
```
olsajiri, line 36: looks like this should be amd64 specific ? this is the only supported arch right? what will happen on other archs?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -811,6 +871,15 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
 	defer f.Close()
 
+	for _, a := range spec.Args {
+		if a.Type == "go_string" {
+			if err := f.ValidateGoABI(); err != nil {
+				return nil, fmt.Errorf("go_string: %w", err)
+			}
+			break
+		}
+	}
```
olsajiri, line 880: let's open the file earlier and call this in addArg together with go_string code? no need for looping args twice

pkg/api/tracingapi/client_kprobe.go
```diff
@@ -681,4 +688,5 @@ type EventConfig struct {
 	BTFArg          [EventConfigMaxArgs][MaxBTFArgDepth]ConfigBTFArg `align:"btf_arg"`
 	UsdtArg         [EventConfigMaxUsdtArgs]ConfigUsdtArg            `align:"usdt_arg"`
 	RegArg          [EventConfigMaxRegArgs]ConfigRegArg              `align:"reg_arg"`
+	GoArg           [EventConfigMaxRegArgs]ConfigGoArg               `align:"go_arg"`
```
olsajiri, line 691: go still can't do unions right? looks like at some point we should consolidate this mess

contrib/tester-progs/go/servecontent-test/main.go
```diff
@@ -0,0 +1,43 @@
...
+	if len(os.Args) > 1 {
+		name = os.Args[1]
+	}
+
+	content := bytes.NewReader([]byte("test"))
+	modtime := time.Now()
+
+	handler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// Calls ServeContent(w, r, name, modtime, content)
+		http.ServeContent(w, r, name, modtime, content)
```
olsajiri, line 28: is there any reason why we don't just test function with string argument? web server seems like overkill and just eating cycles for no reason

pkg/selectors/kernel.go
```diff
@@ -66,6 +66,7 @@ var actionTypeTable = map[string]uint32{
 	"notifyenforcer":              ActionTypeNotifyEnforcer,
 	"cleanupenforcernotification": ActionTypeCleanupEnforcerNotification,
 	"set":                         ActionTypeSet,
+	"cleargostring":               ActionTypeOverride,
```
olsajiri, line 69: hum, do we have use case for this?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -503,6 +503,33 @@ func createGenericUprobeSensor(
...
+			slot := GoABISlotForArg(sym, int(action.ArgIndex))
+			if slot < 0 {
+				return fmt.Errorf("ClearGoString: unknown Go ABI layout for %s arg %d", sym, action.ArgIndex)
+			}
+			_, lenReg, err := GoABISlotRegNames(slot)
+			if err != nil {
+				return fmt.Errorf("ClearGoString: %w", err)
+			}
+			action.Action = "Override"
+			action.ArgRegs = []string{lenReg + "=0"}
```
olsajiri, line 526: let's not do that.. I think it's better to add new argument to override action that says it's "go_string_clear" and setup the regs accordingly in  `ParseMatchAction`

pkg/sensors/tracing/goabi_slots_test.go
```diff
@@ -0,0 +1,88 @@
...
+		got := GoABISlotForArg(tt.symbol, tt.argIndex)
+		found := got >= 0
+		if found != tt.wantFound || (found && got != tt.wantSlot) {
+			t.Errorf("GoABISlotForArg(%q, %d) = %d (found=%v), want %d (found=%v)",
+				tt.symbol, tt.argIndex, got, found, tt.wantSlot, tt.wantFound)
+		}
+	}
+}
+
+func TestExpandClearGoStringActions(t *testing.T) {
```
olsajiri, line 43: can't see actual observer test for zeroing go_string, please add
also please separate tests into separate commit

pkg/sensors/tracing/uprobe_test.go
```diff
@@ -1127,3 +1128,65 @@ spec:
 	err = jsonchecker.JsonTestCheck(t, checker)
 	require.NoError(t, err)
 }
+
+func TestUprobeGoStringArg(t *testing.T) {
```
olsajiri, line 1132: please add this using `tests/policytests`

pkg/sensors/tracing/goabi_slots_gen.go
```diff
@@ -0,0 +1,20 @@
+// Code generated by goabi-gen; DO NOT EDIT.
```
olsajiri, line 1: does it need to be in git if it's generated? we could also skip the format hunk then

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -657,6 +689,30 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+					return errors.New("go_string: sleepable preload requires bpf_copy_from_user_str (kernel >= 6.12)")
+				}
+				if preload {
+					return errors.New("error: can't preload more than one argument")
+				}
+				if len(spec.Symbols) == 0 {
+					return errors.New("go_string type requires at least one symbol")
+				}
+				sym := spec.Symbols[0]
+				slot := GoABISlotForArg(sym, int(a.Index))
```
olsajiri, line 703: how will this work for multiple symbols? should this be limited just for uprobe with one symbol?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -657,6 +689,30 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
 			}
+
+			if argType == gt.GenericGoStringType {
+				if !bpf.HasKfunc("bpf_copy_from_user_str") {
+					return errors.New("go_string: sleepable preload requires bpf_copy_from_user_str (kernel >= 6.12)")
+				}
+				if preload {
+					return errors.New("error: can't preload more than one argument")
+				}
+				if len(spec.Symbols) == 0 {
```
olsajiri, line 699: there's `symbols` variable for that

pkg/selectors/kernel.go
dwindsor: Clearing a string that contains a suspicious payload (i.e. "1=1" for SQL injection, etc)

bpf/process/event_config.h
dwindsor: ~~Makes sense, thanks. I set `GO_ARG_MAX_REGS` to 3 because slices consume 3 registers (ptr, len, cap) and we might want support for that later, but we can add it later if we want.~~

We set GO_ARG_MAX_REGS to 3 here because that's the max # of slots Go native datatypes can consume (slices take 3). 

I'd rather keep just one structure for go arguments, not have multiple ones (eg config_go_string_arg, config_go_int_arg, etc).

wdyt?

pkg/sensors/tracing/goabi_slots.go
dwindsor: Thanks, I forgot to add arch guards here. There's already a guard in `uprobe_tests.go`:

```
if runtime.GOARCH != "amd64" {
	t.Skip("go_string ABI register mapping is amd64-only")
}
```

Doing something similar here.

contrib/tester-progs/go/servecontent-test/main.go
dwindsor: Because we have ABI tests, we want to see how the Go compiler lays things out.

pkg/sensors/tracing/goabi_slots.go
dwindsor: Added `archHasGoString` that enforces `amd64` arch. Will allowlist `arm64` as well when we add arch support.

pkg/sensors/tracing/genericuprobe.go
dwindsor: Reworked, thanks.

New policy:

```
matchActions:
  - action: Override
    clearGoString: true
    argIndex: 2
```

bpf/process/user_preload.h
dwindsor: Updated, thanks. Removed `argPtRegsBit` tagging for `go_string`.

pkg/sensors/tracing/genericuprobe.go
dwindsor: Updated to open the file at the beginning of [addUprobe](https://github.com/dwindsor/tetragon/blob/44411522f10a46dcc40e1cf145094775979ae92e/pkg/sensors/tracing/genericuprobe.go#L533); args are only looped once.

pkg/sensors/tracing/goabi_slots_test.go
dwindsor: added observer test and separated into separate commit

pkg/sensors/tracing/genericuprobe.go
dwindsor: Yes we only support extraction of one symbol per function. This is enough for many use cases. Perhaps if we identify functions later where we want to extract/modify multiple parameters we can expand this.

pkg/sensors/tracing/genericuprobe.go
dwindsor: see below comment

pkg/sensors/tracing/genericuprobe.go
dwindsor: fixed

pkg/sensors/tracing/uprobe_test.go
dwindsor: done

pkg/sensors/tracing/goabi_slots_gen.go
dwindsor: probably not but we have a bunch of different generated files in git, just following that pattern by default

dwindsor:
Thank you for the review @olsajiri! Comments mostly all addressed, a few I have questions on.

CI should be fine, just missing a pr label. Windows CI failing, unrelated:

```
Looking for regex: \{\"process_exec\"\:\{\"process\"\:\{\"exec_id\"\:\".{16,30}\"\,.{0,1}\"pid\"\:5772\,.{0,1}\"uid\"\:[0-9]{0,9}\,.{0,1}\"binary\"\:\"C:\\\\Windows\\\\system32\\\\notepad.exe\"
Get-Content: D:\a\_temp\f4d56df1-39bf-46f4-8c85-7f80bf400926.ps1:39
Line |
```

bpf/process/event_config.h
olsajiri: so you plan on implementing other args then? I guess that's fine, but it'd be shame to keep it like that and never use it

contrib/tester-progs/go/servecontent-test/main.go
olsajiri: could you be more specific in how that's different to some simpler call?

bpf/process/event_config.h
dwindsor: Yes, I already have a branch that adds `setGoInt` to Override actions. After that we could do slices or interfaces.

contrib/tester-progs/go/servecontent-test/main.go
dwindsor: Ahh sorry, you're right, I wasn't reading this carefully. I thought you meant, why can't we hardcode something rather than compiling something.

Reduced this test program as suggested to just target `string.ToUpper()` rather than writing an entire server. 

FWIW the server was used because it uses `net.ServeContent`, which is used in many of the unit tests. But you're right, it's overkill and a waste of cycles.

pkg/sensors/tracing/uprobe_test.go
```diff
@@ -1134,63 +1133,5 @@ spec:
...
-	}
-	observertesthelper.LoopEvents(ctx, t, &doneWG, &readyWG, obs)
-	readyWG.Wait()
-
-	err = exec.Command(testBin, pathArg).Run()
-	require.NoError(t, err)
-
-	err = jsonchecker.JsonTestCheck(t, checker)
-	require.NoError(t, err)
+	policytest.AllPolicyTests.DoObserverTest(t, "uprobe-go-string-arg", nil)
```
olsajiri, line 1286: please squash this with the commit that adds this test

bpf/process/user_preload.h
```diff
@@ -47,6 +47,47 @@ preload_string_type(struct pt_regs *ctx, struct event_config *config, unsigned l
...
+		 * program can copy the pre-formatted [len][data] directly
+		 * to the event buffer without probe_read_str.
+		 */
+		long err;
+
+		if (len > sizeof(data->data) - 4)
+			len = sizeof(data->data) - 4;
+		asm volatile("%[len] &= 0xfff;\n"
+			     : [len] "+r"(len));
+		err = probe_read_user(&data->data[4], len, (void *)ptr);
```
olsajiri: probe_read_user does not allow faults and the point of preload is that we allow faults,
I think we need to use copy_from_user_task instead

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -656,6 +681,36 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 				}
 				preloadArg = true
 			}
+
+			if argType == gt.GenericGoStringType {
+				if !bpf.HasKfunc("bpf_copy_from_user_str") {
```
olsajiri, line 685: this kfunc seems unrelated to the go_string preload

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -656,6 +681,36 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+				if !goABIValidated {
+					if err := f.ValidateGoABI(); err != nil {
+						return fmt.Errorf("go_string: %w", err)
+					}
+					goABIValidated = true
+				}
+				if symbols == 0 {
+					return errors.New("go_string type requires at least one symbol")
+				}
+				sym := spec.Symbols[0]
```
olsajiri, line 700: what if there's more symbols in the uprobe spec? I think we should fail the spec in such case

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -656,6 +681,36 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
...
+
+			if argType == gt.GenericGoStringType {
+				if !bpf.HasKfunc("bpf_copy_from_user_str") {
+					return errors.New("go_string: sleepable preload requires bpf_copy_from_user_str (kernel >= 6.12)")
+				}
+				if preload {
+					return errors.New("error: can't preload more than one argument")
+				}
+				if !goABIValidated {
+					if err := f.ValidateGoABI(); err != nil {
```
olsajiri, line 692: could this be called later underIsStrippedPureGoBinary  condition?  

```
        if symbols != 0 && f.IsStrippedPureGoBinary() {
                tbl, err := f.Pclntab()
                if err != nil {
                        return nil, fmt.Errorf("failed to parse pclntab: %w", err)
                }

                -> in here
```

cmd/goabi-gen/main.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+		return n
+	case *types.Map, *types.Chan:
+		return 1
+	case *types.Signature:
+		return 1
+	case *types.Array:
+		return int(u.Len()) * intRegSlots(u.Elem())
+	}
+	return 1
```
olsajiri: curious.. looks like in here we should return 0, warn or error?

pkg/selectors/kernel.go
```diff
@@ -1194,7 +1200,21 @@ func ParseMatchAction(k *KernelSelectorState, action *v1alpha1.ActionSelector, a
...
+			argRegs := action.ArgRegs
+			if action.ClearGoString {
+				if len(argRegs) > 0 {
+					return errors.New("parseMatchAction: clearGoString cannot be combined with argRegs")
+				}
+				if k.uprobeGoStringClearRegs == nil {
+					return errors.New("parseMatchAction: clearGoString requires uprobe symbol context")
+				}
+				var err error
+				argRegs, err = k.uprobeGoStringClearRegs(action.ArgIndex)
```
olsajiri: this seems worse to me.. I can't find your previous version, but I think now this is more complicated, sorry..

how about we leave user to configure `action.ClearGoString` and we validate that user did not provide `argsRegs` and then we populate `argsRegs` with the needed register  assignment?

pkg/sensors/tracing/goabi_slots.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// GoABISlotRegNames returns the ptr and len register names for a Go string at slot.
+func GoABISlotRegNames(slot int) (ptrReg, lenReg string, err error) {
+	if !archHasGoString() {
```
olsajiri, line 44: it's bit of a shortcut now, but I guess full arch split can wait when we get another arch support

bpf/process/user_preload.h
dwindsor: Thanks! copy_from_user_task does appear to allow faults (it runs with `.might_sleep = true`). Updated to use that.

pkg/sensors/tracing/genericuprobe.go
dwindsor: Yes, this is strange because the go_string preload indeed doesn't call bpf_copy_from_user_str, yet without this, vmtests fail in CI on kernels that are missing copy_from_user_str. I haven't quite figured out why that is the case yet. See this above comment: https://github.com/cilium/tetragon/pull/4817#issuecomment-4193653476

pkg/sensors/tracing/genericuprobe.go
dwindsor: Good point. Updated to reject if the # of arguments isn't exactly 1.

pkg/sensors/tracing/goabi_slots.go
dwindsor: Yeah other arch support will be tiny PRs compared to this.

The reason this PR is so massive is that it adds the Go ABI slot generation code and updates CRDs. I didn't want to add that in a separate PR without a user, hence we have it here in this PR with the most useful (in my opinion) type to write policies against (strings).

pkg/selectors/kernel.go
dwindsor: That's fine, I have no strong opinion on how the action gets translated to registers, I just would rather not allow the user to populate regs whenever it can be fixed up behind the scenes. We'll find a way that works =).

pkg/sensors/tracing/uprobe_test.go
dwindsor: done

pkg/sensors/tracing/genericuprobe.go
dwindsor: Not without the side effect of removing ABI validation on non-stripped binaries. It would be nice if this worked transparently for both cases.

cmd/goabi-gen/main.go
dwindsor: Yes, you're right. This is a leftover hack from when I was trying to get this working with non-stdlib types.

Returning error now when an unknown type is encountered. We don't want to generate a slot table that's based on guessed slot counts.

Added a unit test to explicitly test for this condition: slot generation for an unknown type.

pkg/selectors/kernel.go
dwindsor: Hmm, but aren't we already doing that?

We validate that they didn't provide `argRegs`:
```
if len(argRegs) > 0 {
	return errors.New("parseMatchAction: clearGoString cannot be combined with argRegs")
}
```

Immediately after we populate `argRegs` using `uprobeGoStringClearRegs`. I might be missing something you were thinking about though?

olsajiri: yes, but I think it's better to do it in uprobe code without theuprobeGoStringClearRegs  callback ... just prepare argRegs in uprobe sensor code

dwindsor: Ahh I see. Moved the args fixup code to `expandClearGoStringActions` and call that in `addUprobe` after `elf.SafeOpenFile`.

pkg/sensors/tracing/goabi_slots.go
```diff
@@ -0,0 +1,105 @@
...
+	return offsets[argIndex]
+}
+
+// https://go.googlesource.com/go/+/refs/heads/master/src/cmd/compile/abi-internal.md
+var goABIIntRegs = []string{
+	"rax", "rbx", "rcx", "rdi", "rsi", "r8", "r9", "r10", "r11",
+}
+
+// GoABISlotRegNames returns the ptr and len register names for a Go string at slot.
+func GoABISlotRegNames(slot int) (ptrReg, lenReg string, err error) {
```
olsajiri, line 43: nit, no need for this to be global

bpf/process/user_preload.h
```diff
@@ -65,11 +107,12 @@ preload_pt_regs_arg(struct pt_regs *ctx, struct event_config *config, int index)
 		     : [index] "+r"(index)
 		     : "i"(EVENT_CONFIG_MAX_REG_ARG_MASK));
 
+	ty = config->arg[index];
+
 	reg = &config->reg_arg[index];
 	shift = 64 - reg->size * 8;
 
 	val = read_reg(ctx, reg->offset, shift);
-	ty = config->arg[index];
```
olsajiri, line 72: I dont understand this change.. no nee dto move this right?

bpf/process/generic_calls.h
```diff
@@ -294,6 +295,27 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 	case string_type:
 		size = copy_strings(args, (char *)arg, MAX_STRING);
 		break;
+	case go_string_type: {
```
olsajiri, line 298: I think we need to add go_string_type to type_to_min_size

pkg/sensors/tracing/goabi_slots.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// expandClearGoStringActions turns clearGoString into a plain Override on the
+// Go string's len register so the selector layer just sees argRegs.
+func expandClearGoStringActions(spec *v1alpha1.UProbeSpec) error {
+	for si := range spec.Selectors {
+		sel := &spec.Selectors[si]
+		for ai := range sel.MatchActions {
+			act := &sel.MatchActions[ai]
+			if !act.ClearGoString {
```
olsajiri, line 60: I think this looks better now, but should it also check there is actualy argument with go_string in the spec? otherwise the ValidateGoABI won't be called later on

bpf/process/user_preload.h
```diff
@@ -52,6 +52,48 @@ preload_string_type(struct pt_regs *ctx, struct event_config *config, unsigned l
...
+		long err;
+
+		if (len > sizeof(data->data) - 4)
+			len = sizeof(data->data) - 4;
+		asm volatile("%[len] &= 0xfff;\n"
+			     : [len] "+r"(len));
+		err = copy_from_user_task(&data->data[4], len, (void *)ptr, task, 0);
+		if (err) {
+			*(__u32 *)data->data = 0;
+			data->status = (__u32)-err;
```
olsajiri, line 89: I think we expect  (__u32) -1 as error.. anything else is interpreted as depth error during resolve (getArgStatus)

bpf/include/api.h
```diff
@@ -117,6 +117,9 @@ static int BPF_FUNC(probe_read_kernel, void *dst, uint32_t size, const void *src
 static int BPF_FUNC(probe_read_user, void *dst, uint32_t size, const void *src);
 static int BPF_FUNC(probe_write_user, void *dst, const void *src, uint32_t len);
 static int BPF_FUNC(copy_from_user, void *dst, uint32_t size, const void *src);
+struct task_struct;
```
olsajiri, line 120: is this needed? I get clean build without it

cmd/goabi-gen/main.go
```diff
@@ -0,0 +1,256 @@
+// SPDX-License-Identifier: Apache-2.0
```
olsajiri, line 1: this seems to break `make codegen`:

```
...
go generate ./pkg/sensors/tracing/...
2026/05/14 09:35:16 write pkg/sensors/tracing/goabi_slots_gen.go: open pkg/sensors/tracing/goabi_slots_gen.go: no such file or directory
exit status 1
```

PR #4828, opened 2026-04-07
sensors: fix matchParentBinaries test

### LINKED ISSUE #4761: In-Kernel Policy Metrics: Counting fast events without the log overhead

Right now, if I want to count how many times something happens (like a file being read or a network connection), tetragon has to send a full log to user-space every single time. 

If something happens very fast (like 1000 of times per sec), the logs fill up quickly and the CPU gets very busy just processing all those events. Most of the time, I don't need all those logs; I just want to know "how many times did this happen?"

Probably a good way to solve this  is to add a feature **In-Kernel Policy Metrics** which lets tetragon count things directly in the kernel without sending a log for every single event

-  **In the Kernel (BPF)**: We could add a small array to the `policy_stats` map. When a policy rule matches, the BPF code would just increment a number in that array. This should be atomic and very fast.
-  **In the Agent (Go)**: We could create a system to track which metric name belongs to which index in the BPF map. 
-  **Prometheus**: A new collector could read the BPF maps directly and export the counters to Prometheus. This would give us a new metric like `tetragon_tracingpolicy_custom_metric_total`.

example policy:
```yaml
spec:
  kprobes:
  - call: "sys_execve"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Equal"
        values: ["/usr/bin/touch"]
      matchActions:
      - action: PostMetric
        metricName: "touch_calls_total"
```

### Code of Conduct

ethical-buddy:
Fixes #4761

### Description

`TestKprobeMatchParentBinaries` used the exported `ProcessKprobe` parent as a proxy for `matchParentBinaries`. That is not reliable for the exec-without-clone path: the selector uses the previous binary saved in `tg_parents_bin`, while the exported parent is resolved from the task parent.

This updates the test to assert the selector behavior through the emitted `fd_install` event instead. Each case runs the shell that should match the selector against a known file, then checks that the expected `/usr/bin/tail` event for that file is reported.

The test also restores `option.Config.ParentsMapEnabled` after it runs so the package does not leak global state into later tracing tests.

### Changelog

```release-note
None
```

### DIFF
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -4043,15 +4043,31 @@ spec:
 	return configHook.String()
 }
 
-func createParentsChecker(parent, binary string) *ec.ProcessKprobeChecker {
+func createMatchParentBinariesChecker(binary, filename string) *ec.ProcessKprobeChecker {
 	kpChecker := ec.NewProcessKprobeChecker("").
-		WithParent(ec.NewProcessChecker().WithBinary(sm.Full(parent))).
 		WithProcess(ec.NewProcessChecker().WithBinary(sm.Full(binary))).
-		WithFunctionName(sm.Full("fd_install"))
+		WithFunctionName(sm.Full("fd_install")).
+		WithArgs(ec.NewKprobeArgumentListMatcher().
+			WithOperator(lc.Subset).
+			WithValues(
+				ec.NewKprobeArgumentChecker().WithFileArg(ec.NewKprobeFileChecker().WithPath(sm.Full(filename))),
+			))
 	return kpChecker
 }
 
-func matchParentBinariesTest(t *testing.T, operator string, values []string, kpChecker *ec.ProcessKprobeChecker, newProcess, fentry bool) {
+func runShellTail(t *testing.T, shell, file string, newProcess bool) {
+	args := []string{"-c"}
+	if newProcess {
+		args = append(args, fmt.Sprintf("echo '/usr/bin/tail %s' | %s", file, shell))
+	} else {
+		args = append(args, "/usr/bin/tail "+file)
+	}
+	if err := exec.Command(shell, args...).Run(); err != nil {
+		t.Fatalf("failed to run tail %s with %s: %s", file, shell, err)
+	}
+}
+
+func matchParentBinariesTest(t *testing.T, operator string, values []string, matchingShell string, newProcess, fentry bool) {
 	var doneWG, readyWG sync.WaitGroup
 	defer doneWG.Wait()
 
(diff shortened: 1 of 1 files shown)


FedeDP:
Ehy thanks for this PR!
It seems like the modified tests are now failing in CI; can you please take a look?

kevsecurity (changes requested):
Please can you add commit descriptions to explain what each commit addresses? Thanks!

ethical-buddy:
Thanks for the review. I cleaned this up into a single signed-off commit with a commit body explaining the selector fix and test changes. I also fixed the lint issue and tightened the BPF selector path so a missing parent-binary map entry does not bypass matchParentBinaries.

kevsecurity:
> Thanks for the review. I cleaned this up into a single signed-off commit with a commit body explaining the selector fix and test changes. I also fixed the lint issue and tightened the BPF selector path so a missing parent-binary map entry does not bypass matchParentBinaries.

Thanks, I kicked off the tests again. Assuming they come back green, I think we really prefer commit messages to be limited to 80 chars; would you be able to wrap you commit message? (I'd wait for the tests to run first though.)

bpf/process/pfilter.h
```diff
@@ -428,21 +427,17 @@ selector_process_filter(__u32 *f, __u32 index, struct execve_map_value *enter,
...
 	if (PARENTS_MAP_ENABLED) {
 		struct binary *parent_bin = map_lookup_elem(&tg_parents_bin, &enter->key.pid);
 
-		if (parent_bin)
-			/* matchParentBinaries key is in range [MAX_SELECTORS; MAX_SELECTORS * 2) */
-			if (!match_binaries(index + MAX_SELECTORS, enter, parent_bin))
-				return 0;
+		/* matchParentBinaries key is in range [MAX_SELECTORS; MAX_SELECTORS * 2) */
+		if (!match_binaries(index + MAX_SELECTORS, enter, parent_bin))
+			return 0;
```
olsajiri: so this is all guarded with `PARENTS_MAP_ENABLED` but that is not enabled for kprobes ATM,
I think we need following fix for that https://github.com/cilium/tetragon/pull/4888

so I wonder we enable it by accident with exec sensor or we need some other fix for the issue you described

ethical-buddy:
Thanks, I checked the failures from the last run. The extra BPF-side change was too broad and caused unrelated runtime tests to fail, so I dropped it and kept the patch focused on the flaky test. I also wrapped the commit message body and force-pushed the cleaned single commit.

bpf/process/pfilter.h
olsajiri: @ethical-buddy I think your change was right, I separated it in https://github.com/cilium/tetragon/pull/4888/changes/08234eb78786c286836cfa74fdee21d8b8bf6aa7

it fixes this issue https://github.com/cilium/tetragon/issues/5069 for me, @tdaudi could you plz check?

I think we need changes from https://github.com/cilium/tetragon/pull/4888 as well
could you please add your test changes on top of it? thanks

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -4105,100 +4101,91 @@ func testKprobeMatchParentBinaries(t *testing.T, fentry bool) {
 	tests := map[string]struct {
 		operator                string
 		values                  []string
-		expectedParent          string
-		binary                  string
+		matchingShell           string
 		parentCreatesNewProcess bool
```
kkourt, line 4105: I find this hard to follow.

Can we change this to have:
* The command that we execute
* What we want to match in the event checker

To make the test cases more obvious?

PR #4832, opened 2026-04-07
renovate: add gomodVendor to postUpdateTasks

dangome3:
### Description
Sometimes the following error is shown when doing a major release update:

  go: inconsistent vendoring in ...

This is triggered because some tasks like updating the imports with 'mod' requires vendor packages to be updated too. This change shouldn't have repercusions as 'make vendor' runs anyway in postUpgradeTasks


FedeDP (approved):
LGTM thanks!

mtardy (approved):

PR #4833, opened 2026-04-08
Add support for matchWorkloads selector

tpapagian:
This works in a similar way to global workload selectors but allows to do fine-grain workload selection inside a single tracing policy.

Example:
```yaml
kind: TracingPolicy
metadata:
  name: "file-monitoring-filtered"
spec:
  kprobes:
  - call: "security_file_permission"
    syscall: false
    return: true
    args:
    - index: 0
      type: "file" # (struct file *) used for getting the path
    - index: 1
      type: "int" # 0x04 is MAY_READ, 0x02 is MAY_WRITE
    returnArg:
      index: 0
      type: "int"
    selectors:
    - matchArgs:      
      - index: 0
        operator: "Prefix"
        values:
        - "/etc/shadow"
      - index: 1
        operator: "Equal"
        values:
        - "2" # MAY_WRITE
      matchWorkloads: # match only host workloads
        hostSelector: {}
    - matchArgs:      
      - index: 0
        operator: "Prefix"
        values:
        - "/etc/shadow"
      - index: 1
        operator: "Equal"
        values:
        - "4" # MAY_READ
      matchWorkloads: # match host workloads and pods inside "kube-system" namespace
        hostSelector: {}
        podSelector:
          matchExpressions:
          - key: "k8s:io.kubernetes.pod.namespace"
            operator: In
            values:
            - "kube-system"
```

More details can be found in each patch.


pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -209,6 +213,22 @@ type CapabilitiesSelector struct {
...
+	PodSelector *slimv1.LabelSelector `json:"podSelector"`
+	// +kubebuilder:validation:Optional
+	// ContainerSelector selects containers that this policy applies to.
+	// A map of container fields will be constructed in the same way as a map of labels.
+	// The name of the field represents the label "key", and the value of the field - label "value".
+	// Currently, only the "name" field is supported.
+	ContainerSelector *slimv1.LabelSelector `json:"containerSelector"`
+	// +kubebuilder:validation:Optional
+	// HostSelector selects hosts that this policy applies to.
+	// For now only ~ (none) and {} (all) is supported.
```
mtardy, line 234: note: maybe we can X-Validation that only `~` and `{}` is supported?

pkg/selectors/kernel.go
```diff
@@ -1263,6 +1265,65 @@ func ParseMatchAction(k *KernelSelectorState, action *v1alpha1.ActionSelector, a
...
+
+	podSelector := workloads[0].PodSelector
+	containerSelector := workloads[0].ContainerSelector
+	hostSelector := workloads[0].HostSelector
+
+	// The user explicitly defined all podSelector, containerSelector, and hostSelector
+	// to be {}. This will match everything and for this reason we should not use
+	// a policyfilter at all.
+	matchAll := func(s *slimv1.LabelSelector) bool {
+		return (s != nil && (len(s.MatchLabels)+len(s.MatchExpressions) == 0))
```
mtardy: Does that work here to compare to an empty LabelSelector? like `LabelSelector{}`?

pkg/selectors/kernel.go
```diff
@@ -1263,6 +1265,65 @@ func ParseMatchAction(k *KernelSelectorState, action *v1alpha1.ActionSelector, a
...
+	// This covers the "special" case where all of podSelector, containerSelector, hostSelector
+	// are nil (default). In that case we match everything so no need to apply a policyfilter.
+	matchNothing := func(s *slimv1.LabelSelector) bool {
+		return s == nil
+	}
+	if matchNothing(podSelector) && matchNothing(containerSelector) && matchNothing(hostSelector) {
+		return nil
+	}
+
+	// If the user specifies a podSelector but don't specify a containerSelector,
```
mtardy: small typo: dont -> doesn't. Also in next comment

pkg/selectors/kernel.go
```diff
@@ -1263,6 +1265,65 @@ func ParseMatchAction(k *KernelSelectorState, action *v1alpha1.ActionSelector, a
...
+	// we assume that the user cares for all containers inside the pods that match.
+	if podSelector != nil && containerSelector == nil {
```
mtardy: maybe reuse you `matchNothing` method here for containerSelector? It's fairly trivial but I think it documents well the code. Same after

pkg/sensors/tracing/generickprobe.go
```diff
@@ -1036,6 +1036,9 @@ func createKprobeSensorFromEntry(polInfo *policyInfo, kprobeEntry *genericKprobe
 	filterMap := program.MapBuilderProgram("filter_map", load)
 	maps = append(maps, filterMap)
 
+	workloadsMap := program.MapBuilderProgram("workloads_map", load)
```
mtardy, line 1043: maybe make `workloads_map` a const somewhere and reference it every time you use it?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -718,6 +719,32 @@ matchCapabilityChanges:
 See a [demonstration example](https://github.com/cilium/tetragon/blob/main/examples/tracingpolicy/fd_install_cap_changes.yaml)
 of this feature.
 
+## Workloads filter
+
+Workloads filter can be specified under the `matchWorkloads` field and provide
```
mtardy: provide -> provides

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -718,6 +719,32 @@ matchCapabilityChanges:
...
+filtering based on Kubernetes workloads. Inside `matchWorkloads` the user can
+define a `hostSelector`, a `podSelector`, and a `containerSelector`.
+
+This works in a similar way to global workload selectors such as `spec.hostSelector`,
+`spec.podSelector`, and `spec.containerSelector`. More details on these
+can be found in [Filtering semantics](https://tetragon.io/docs/concepts/tracing-policy/k8s-filtering/#filtering-semantics).
```
mtardy: please reference the rest of the docs using relative path, here would be 

[Filtering semantics]({{< ref "/docs/concepts/tracing-policy/k8s-filtering/#filtering-semantics" >}})

This allows the doc to know at build time if things still exist more importantly and domain name to be changed (if we rename tetragon lol).

mtardy (commented):
Thanks!! A few comments, I think it's almost good to go :)

pkg/k8s/apis/cilium.io/v1alpha1/types.go
tpapagian: Yes, this make sense, I have added:
```
// +kubebuilder:validation:XValidation:rule="!has(self.matchLabels) && !has(self.matchExpressions)",message="The hostSelector should be either null or {}."
```
that seems to work. 

I have also created https://github.com/cilium/tetragon/pull/4947 to add the same validation check in the global `spec.hostSelector`.

pkg/selectors/kernel.go
tpapagian: Fixed

pkg/selectors/kernel.go
tpapagian: Fixed

docs/content/en/docs/concepts/tracing-policy/selectors.md
tpapagian: Fixed

docs/content/en/docs/concepts/tracing-policy/selectors.md
tpapagian: Fixed

pkg/sensors/tracing/generickprobe.go
tpapagian: I would propose to leave it as it is for now because it is not clear to me on where is a good place to add this `const`. Possibly `pkg/selectors/kernel.go ` but this will not make things consistent as it is now in the previous `program.MapBuilderProgram` calls in the same file. If you have anything specific in mind, just let me know.

pkg/selectors/kernel.go
tpapagian: Seems to work with `DeepEqual`. I have changed that as it make the code cleaner.

tpapagian:
> Thanks!! A few comments, I think it's almost good to go :)

Thanks for the review. I believe I have addressed all of your comments.

mtardy (approved):
Cool thanks!

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -169,6 +170,9 @@ type KProbeSelector struct {
 	// IDs for capabilities changes
 	MatchCapabilityChanges []CapabilitiesSelector `json:"matchCapabilityChanges,omitempty"`
 	// +kubebuilder:validation:Optional
+	// Workloads to match
+	MatchWorkloads []WorkloadsSelector `json:"matchWorkloads,omitempty"`
```
kkourt: Is this a `AND` (set intersection) or an `OR` (set union)?

bpf/process/policy_filter.h
```diff
@@ -93,4 +93,22 @@ FUNC_INLINE bool policy_filter_check(u32 policy_id)
 	return !map_lookup_elem(policy_map, &cgroupid);
 }
 
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
```
kkourt: Selector IDs are consecutive, right? If so, I think having this as an array is better because it's a more performant lookup.

bpf/process/policy_filter.h
```diff
@@ -93,4 +93,22 @@ FUNC_INLINE bool policy_filter_check(u32 policy_id)
...
+	__uint(max_entries, MAX_SELECTORS);
+	__type(key, __u32); /* selector id */
+	__type(value, __u32); /* policy_id */
+} workloads_map SEC(".maps");
+
+FUNC_INLINE int match_workloads(__u32 selector_id)
+{
+	__u32 *pol_id = 0;
+
+	pol_id = map_lookup_elem(&workloads_map, &selector_id);
```
kkourt, line 107: non-blocking comment: I think there are some tricks we could play here to avoid the double lookup in the future. I am not sure if it matters, however, if it's an array and not a hash.

pkg/selectors/helpers.go
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package selectors
+
+import slimv1 "github.com/cilium/tetragon/pkg/k8s/slim/k8s/apis/meta/v1"
+
+func MatchAllLabelSelector(s *slimv1.LabelSelector) bool {
```
kkourt, line 8: The commit message is "Extract common functionality"

Can you please reword it so that it is more specific?

pkg/policyfilter/selectors.go
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build !nok8s
+
+package policyfilter
+
+import "sync/atomic"
+
+var ops atomic.Uint32
```
kkourt: What does ops stand for? I tend to associate ops with "operations" so I had to look at the code below to understand what this does.

pkg/policyfilter/selectors.go
```diff
@@ -0,0 +1,14 @@
...
+
+func GetSelectorPolicyID() PolicyID {
+	return PolicyID(ops.Add(1) + polMaxPolicies)
```
kkourt: Is there a reason for not having a single counter for everything?

pkg/policyfilter/selectors.go
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build !nok8s
+
+package policyfilter
```
kkourt, line 6: non-blocking comment: would it make sense to rename this to workloadselector or something? The `policyfilter` term was always confusing and with this change it does not really make sense anymore.

pkg/selectors/kernel.go
```diff
@@ -1263,6 +1265,59 @@ func ParseMatchAction(k *KernelSelectorState, action *v1alpha1.ActionSelector, a
 	return nil
 }
 
+func ParseMatchWorkloads(k *KernelSelectorState, workloads []v1alpha1.WorkloadsSelector, selIdx int) error {
+	if len(workloads) > 1 {
+		return errors.New("only a single selector under matchWorkloads is supported")
```
kkourt: Is there a use-case for having more than one selector? If not, maybe we should not have an array there?

kkourt (commented):
Thanks!

pkg/k8s/apis/cilium.io/v1alpha1/types.go
tpapagian: Do you mean for the array? If yes, you other comment `Is there a use-case for having more than one selector? If not, maybe we should not have an array there?` is relevant to that as well.

I am not really sure that it make sense to have more than one for now. So to answer your question `AND (set intersection) or an OR (set union)` I would say this is not defined in that PR. 

I just followed the common syntax that we use for the other types as well to be an array just to be consistent. This also results in a more consistent syntax in out tracing policies. But we will have the ability to add more later without breaking the user API if we need to support more complex selectors.

If this does not make sense, I can remove the array.

pkg/policyfilter/selectors.go
tpapagian: We can have a single counter. The only reason that I did that is that now tracing policies have sequential IDs. If we have the same counter then tracing policies may don't have sequential IDs. I believe that we don't rely on that anywhere so possibly we can remove that.

pkg/policyfilter/selectors.go
tpapagian: Do you mean to rename `pkg/policyfilter` to `pkg/workloadselector`? I believe that it makes sense to do that change. I would prefer to do that in a separate PR as I believe that this would result in several changes that are unrelated to that PR.

pkg/policyfilter/selectors.go
tpapagian: Yes, `ops` is not a very good name. If we decide to use a separate counter, I will update that to a better name.

pkg/selectors/kernel.go
tpapagian: Replied in https://github.com/cilium/tetragon/pull/4833#discussion_r3273115723

bpf/process/policy_filter.h
tpapagian: It's a bit easier to check if this selector has `matchWorkloads`. But we can change that to an array and use a dummy value for not defined. I will do that change as it will have a better performance.

tpapagian: Changed that to `BPF_MAP_TYPE_ARRAY`.

pkg/selectors/helpers.go
tpapagian: Fixed

pkg/policyfilter/selectors.go
tpapagian: Fixed

pkg/k8s/apis/cilium.io/v1alpha1/types.go
kkourt: If we think there is no point (either now or in the future) to have an array (which I think it's the case, workloadSelector itself can contain mulitple matchers), then I would say remove the array. One of the benefits of the array is that it has an obvious empty value. If we want this, we can have a pointer to a `WorklodSelector`:

```
MatchWorkloads *WorkloadSelector
```

pkg/policyfilter/selectors.go
kkourt: Up to you! We can always change it later. I don't think we rely in sequential IDs in policies (indeed, we shouldn't)

pkg/k8s/apis/cilium.io/v1alpha1/types.go
tpapagian: I removed the array for here and used a pointer to WorkloadSelector instead.

tpapagian:
Merging that. All comments have been addressed. If there are any items missing, we can always address them in followup PRs.

mann-ds:
:wave: when will there be a new release with this PR changes included?

mann-ds:
@mtardy please include this one as well :slightly_smiling_face:

mtardy:
> @mtardy please include this one as well 🙂

Hey I've seen your messages here and there. Unfortunately you'll have to wait until the next release, usually it's at least 3 months between two releases and the last one was released April 29th. You can use CI images in the meantime from main if you want to test dev features.

PR #4834, opened 2026-04-08
tetragon: Load properly on kernels without module support

### LINKED ISSUE #4802: Support for kernels without module support

My kernel does not have module support (`CONFIG_MODULES=n`) and also some other tailoring configurations. I also build Tetragon from source. Before `make` I dump my kernels BTF and use that instead of the `vmlinux_generated_x86.h` provided in the Tetragon source. However, the build fails with:
```
error: incomplete definition of type 'const struct load_info'
```
Because my kernel does not have module support the BTF does not contain `struct load_info`.

As a workaround I use indirection: I dump my kernels BTF to `bpf/include/vmlinux_generated_x86.h.custom` and add a new file `bpf/include/module_defs.h`:
```
#ifndef __VMLINUX_MODULE_DEFS_H__
#define __VMLINUX_MODULE_DEFS_H__

struct load_info {
	...
};

#endif
```
Then, I overwrite the provided `vmlinux_generated_x86.h` like so:
```
#ifndef __VMLINUX_GENERATED_X86_H__
#define __VMLINUX_GENERATED_X86_H__

#include "vmlinux_generated.h.custom"
#include "module_defs.h"

#endif
```
That way the build works. I think the dependency to `load_info` was introduced via [#1433](https://github.com/cilium/tetragon/issues/1433).

Could you please introduce some kind of feature flag to allow building Tetragon from source for kernels that do not have module support?

A flag for make that controls a preprocessor directive perhaps?

### Code of Conduct

olsajiri:
Make sure we load on kernels without module support.

The 'struct load_info' is internal modules struct available only if there's module support compiled in (CONFIG_MODULES).

Fixes: #4802

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -480,6 +480,14 @@ FUNC_INLINE long copy_load_module(char *args, unsigned long arg)
 	const struct load_info *mod = (struct load_info *)arg;
 	struct tg_kernel_module *info = (struct tg_kernel_module *)args;
 
+	/*
+	 * Make sure we don't crash on kernel without module support,
+	 * 'struct load_info' is internal modules struct available only
+	 * if there's module support compiled in.
+	 */
+	if (!bpf_core_type_exists(struct load_info))
+		return 0;
+
 	memset(info, 0, sizeof(struct tg_kernel_module));
 
 	if (BPF_CORE_READ_INTO(&name, mod, name) != 0)


bpf/process/types/basic.h
```diff
@@ -480,6 +480,14 @@ FUNC_INLINE long copy_load_module(char *args, unsigned long arg)
 	const struct load_info *mod = (struct load_info *)arg;
 	struct tg_kernel_module *info = (struct tg_kernel_module *)args;
 
+	/*
+	 * Make sure we don't crash on kernel without module support,
+	 * 'struct load_info' is internal modules struct available only
+	 * if there's module support compiled in.
+	 */
+	if (!bpf_core_type_exists(struct load_info))
```
FedeDP, line 488: We are already casting `arg` as `struct load_info` above:
```
const struct load_info *mod = (struct load_info *)arg;
```
I think it would be better to move the cast after the check.

andrewstrohman (approved):

bpf/process/types/basic.h
olsajiri: hm,  I think the casting is not a problem, we do have the struct definition.. it's the core deref below that fails, because the struct is not in btf, what do I miss?

FedeDP: Oh yes i get it, it's just to be more clear about our intent. Anyway, no strong opinion!

FedeDP (approved):

PR #4838, opened 2026-04-09
fix: update buf breaking to point to v1.5 branch

dangome3:
'make protogen' was failing because the variable BUF_BREAKING_AGAINST_BRANCH was pointing to the wrong branch


mtardy (approved):
thanks, ~~are all stables branches good now?~~ nevermind just saw your comment here https://github.com/cilium/tetragon/pull/4835#issuecomment-4216180979.

PR #4848, opened 2026-04-10
pkg/selectors: fix CPU hangs

### LINKED ISSUE #4699: OOM and CPU Hang via Unbounded `InMap` Ranges in TracingPolicy

Tetragon's `InMap` selector incorrectly parses unbounded massive ranges and maximum int bounds, leading to OOM or CPU hang (infinite loops) due to unrestrained map allocs and int overflows.

Apply this tracing policy
```yaml
kind: TracingPolicy
metadata:
  name: "test"
spec:
  kprobes:
  - call: "sys_prctl"
    syscall: true
    args:
    - index: 0
      type: "int"
    selectors:
    - matchArgs:
      - index: 0
        operator: "InMap"
        values:
        - "0:100000000"
```
This results in:
```
  NAME             READY   STATUS      RESTARTS      AGE
  tetragon-fpn9z   1/2     OOMKilled   1             10m
```

Also
```yaml
kind: TracingPolicy
metadata:
  name: "range-infinite-loop-bug"
spec:
  kprobes:
  - call: "sys_prctl"
    syscall: true
    args:
    - index: 0
      type: "uint64"
    selectors:
    - matchArgs:
      - index: 0
        operator: "InMap"
        values:
        - "18446744073709551610:18446744073709551615"
```

Since `0 <= 18446744073709551615` inherently evaluates to `true`, the `for` loop resets and continues iterating without skipping a beat forever.

In this case tetragon ceasing to process any further security logs because its primary processing thread is locked natively at 100% CPU usage spinning endless circles.
```
kubectl top pods -n tetragon
  NAME                                 CPU(cores)   MEMORY(bytes)   
  tetragon-fpn9z                       976m         952Mi   
```

### Tetragon Version

latest

### Kernel Version

6.8.0-50-generic

### Kubernetes Version

_No response_

### Bugtool

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

3u13r:
Fixes #4699

### Description
Fixes an infinite loop for ranges with `math.MaxUint64` as upper bound.

Also documents that using range notation with the `InMap` operator is memory intensive. See the commit message for details.

### Changelog

```release-note
Fix a infinite loop when using range notation with math.MaxUint64 as upper bound
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -2113,8 +2113,17 @@ exist.
 - String prefix max length: 256 chars
 - String postfix max length: 128 chars
 
-For an unlimited number of values, consider using the `InMap` or `NotInMap`
+For larger sets of values, consider using the `InMap` or `NotInMap`
 operators which store values in a BPF map.
+These are limited only by the amount of available memory.
+
+{{< caution >}} 
+The `InMap` and `NotInMap` operators also support the range notation described
+for the `InRange` operator. However, using range notation with `InMap` or
+`NotInMap` consumes more memory, because each value in the range is added
+individually to the map. For large ranges, prefer the `InRange` operator
+instead.
+{{< /caution >}}
 
 
 ## Return Actions filter
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -466,20 +466,26 @@ func writeRangeInMap(v string, ty uint32, op uint32, m *ValueMap) error {
 		if sRangeVal[0] > sRangeVal[1] {
 			sRangeVal[0], sRangeVal[1] = sRangeVal[1], sRangeVal[0]
 		}
-		for val := sRangeVal[0]; val <= sRangeVal[1]; val++ {
+		for val := sRangeVal[0]; ; val++ {
 			var valByte [8]byte
 			binary.LittleEndian.PutUint64(valByte[:], uint64(val))
 			m.Data[valByte] = struct{}{}
+			if val >= sRangeVal[1] {
+				break
+			}
 		}
 
 	case gt.GenericU64Type, gt.GenericU32Type, gt.GenericU16Type, gt.GenericU8Type:
 		if uRangeVal[0] > uRangeVal[1] {
 			uRangeVal[0], uRangeVal[1] = uRangeVal[1], uRangeVal[0]
 		}
-		for val := uRangeVal[0]; val <= uRangeVal[1]; val++ {
+		for val := uRangeVal[0]; ; val++ {
 			var valByte [8]byte
 			binary.LittleEndian.PutUint64(valByte[:], val)
 			m.Data[valByte] = struct{}{}
+			if val >= uRangeVal[1] {
+				break
+			}
 		}
 	}
 	return nil
--- a/pkg/selectors/kernel_test.go
+++ b/pkg/selectors/kernel_test.go
@@ -12,6 +12,9 @@ import (
 	"bytes"
 	"encoding/binary"
 	"errors"
+	"fmt"
+	"maps"
+	"math"
 	"strings"
 	"syscall"
 	"testing"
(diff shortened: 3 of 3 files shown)


pkg/selectors/kernel.go
```diff
@@ -461,24 +461,37 @@ func writeRangeInMap(v string, ty uint32, op uint32, m *ValueMap) error {
 			return fmt.Errorf("unknown type: %d", ty)
 		}
 	}
+
+	// Arbitrary limit of the range of value that we insert for a single range value
+	// For large ranges the user should use the InRange operator.
+	const rangeLimit = 1000
```
3u13r: Depending on how much memory the user has and is willing to use for large ranges, we break current users. 
Should start out giving a warning in this release and then prohibit it in the next?

Generally: What is a good value here? I think that people likely should use the InRange selector for everything over 10, but that doesn't mean that we shouldn't allow larger values.

3u13r:
As discussed in the community meeting, I'll think about if the range limit is a good idea at all.
Therefore moving this to draft for now.

pkg/selectors/kernel.go
3u13r: Decided not to limit the user but instead only document the behavior for now. See the 2nd commit message for problems identified with this approach.

3u13r:
Ready to review again. Especially let me know if you think we should handle the OOM behavior instead of only documenting it (see [2nd commit](https://github.com/cilium/tetragon/pull/4848/changes/31ee5e5691be62143ae118a2fefa71161cbabde2)).

pkg/selectors/kernel.go
```diff
@@ -466,20 +466,26 @@ func writeRangeInMap(v string, ty uint32, op uint32, m *ValueMap) error {
 		if sRangeVal[0] > sRangeVal[1] {
 			sRangeVal[0], sRangeVal[1] = sRangeVal[1], sRangeVal[0]
 		}
-		for val := sRangeVal[0]; val <= sRangeVal[1]; val++ {
+		for val := sRangeVal[0]; ; val++ {
 			var valByte [8]byte
-			binary.LittleEndian.PutUint64(valByte[:], uint64(val))
+			binary.LittleEndian.PutUint64(valByte[:], uint64(sRangeVal[0]+val))
```
mtardy: I don't understand that change

pkg/selectors/kernel.go
```diff
@@ -466,20 +466,26 @@ func writeRangeInMap(v string, ty uint32, op uint32, m *ValueMap) error {
...
 
 	case gt.GenericU64Type, gt.GenericU32Type, gt.GenericU16Type, gt.GenericU8Type:
 		if uRangeVal[0] > uRangeVal[1] {
 			uRangeVal[0], uRangeVal[1] = uRangeVal[1], uRangeVal[0]
 		}
-		for val := uRangeVal[0]; val <= uRangeVal[1]; val++ {
+		for val := uRangeVal[0]; ; val++ {
 			var valByte [8]byte
-			binary.LittleEndian.PutUint64(valByte[:], val)
+			binary.LittleEndian.PutUint64(valByte[:], uRangeVal[0]+val)
```
mtardy: neither this one then symetrically

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -2113,8 +2113,16 @@ exist.
...
+For larger sets of values, consider using the `InMap` or `NotInMap`
 operators which store values in a BPF map.
+These are limited only by the amount of available memory.
+
+
+The `InMap` and `NotInMap` operators also support the range notation 
+described for the `InRange` operator. However, using range notation 
+with `InMap` or `NotInMap` is more memory-intensive, since every value 
+in the range is individually added to the map. For large ranges, 
+prefer the `InRange` operator instead.
```
mtardy: You could use the caution bloc for this and remove the empty line
```suggestion
{{< caution >}} 
The `InMap` and `NotInMap` operators also support the range notation described
for the `InRange` operator. However, using range notation with `InMap` or
`NotInMap` consumes more memory, because each value in the range is added
individually to the map. For large ranges, prefer the `InRange` operator
instead.
{{< /caution >}}
```

mtardy:
> Ready to review again. Especially let me know if you think we should handle the OOM behavior instead of only documenting it (see [2nd commit](https://github.com/cilium/tetragon/pull/4848/changes/31ee5e5691be62143ae118a2fefa71161cbabde2)).

You have a bunch of issues in the CI that would be worth investigating maybe before, something's wrong

pkg/selectors/kernel.go
3u13r: My bad this is bogus and from an earlier version. Removed this and added to the test that we also assert the created map to catch this in the future.

pkg/selectors/kernel.go
3u13r: see https://github.com/cilium/tetragon/pull/4848#discussion_r3095167637

docs/content/en/docs/concepts/tracing-policy/selectors.md
3u13r: Done, thanks for the suggestion!

3u13r:
> You have a bunch of issues in the CI that would be worth investigating maybe before, something's wrong


Seems like the the CI failed due to the error I corrected [here](https://github.com/cilium/tetragon/pull/4848#discussion_r3095167637)

mtardy (approved):
Thanks!

PR #4854, opened 2026-04-10
Dockerfile: use BuildKit cache mounts

3u13r:
### Description

Due to the `COPY . ./` we thrash the docker layers cache even if we only touch unrelated files like READMEs.

This change uses BuildKit's cache mounts and ccache to give clang and go a place to store their compiler cache.

This is mostly important for development environments since those caches are not automatically exported and shared between GitHub runners.

In total this change reduces compile times from 1m 20s to 25s for Go changes and from 1m20s to 40s for a change to the eBPF headers.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,11 +12,12 @@
 # First builder (cross-)compile the BPF programs
 FROM --platform=$BUILDPLATFORM quay.io/cilium/clang:b97f5b3d5c38da62fb009f21a53cd42aefd54a2f@sha256:e1c8ed0acd2e24ed05377f2861d8174af28e09bef3bbc79649c8eba165207df0 AS bpf-builder
 WORKDIR /go/src/github.com/cilium/tetragon
-RUN apt-get update && apt-get install -y linux-libc-dev gzip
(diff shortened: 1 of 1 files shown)


mtardy (commented):
Thanks! Maybe also explain the ccache change in the patch?

mtardy (approved):
lgtm

mtardy:
Restarting some jobs, it looks like something went wrong infra side when they got triggered

mtardy:
Btw can you check what kind of speedup we can see in the CI timing? Always nice numbers to show.

3u13r:
I agree that CI speedup numbers would be nice, but I don't think this improves the CI. 
I've written in the commit message:
```
This is mostly important for development environments since those caches
are not automatically exported and shared between GitHub runners.
```
The problem is described in more detail (more details [here](https://github.com/moby/buildkit/issues/1512).

Do we have dedicated runners or share the BuildKit cache somehow? Do we have workflows that run e.g. `make image` multiple times but cannot cache the whole image layer?

Also I think permissions to (manually) trigger the CI would be immensely helpful to get the numbers.

PR #4858, opened 2026-04-14
cleanup(ci): avoid x-compilation in packages-e2e-tests CI.

### LINKED ISSUE #4693: tests: Packages e2e Tests `standalone-tarball-tests` is flaky
It's been happening quite a while lately, we need to address this either by disabling it or fixing it.

See examples https://github.com/cilium/tetragon/actions/runs/22216418576?pr=4548.

FedeDP:
Fixes: #4693

Not 100% sure whether it fixes the issue but seems to help (and it's also a nice cleanup):
immagine

### DIFF
--- a/.github/workflows/packages-e2e-tests.yaml
+++ b/.github/workflows/packages-e2e-tests.yaml
@@ -8,28 +8,12 @@ on:
 
 jobs:
   standalone-tarball-builds:
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ (matrix.arch == 'arm64' && 'ubuntu-22.04-arm') || 'ubuntu-22.04' }}
     strategy:
       fail-fast: false
       matrix:
-        include:
-          # We use the native arch build
-          - os: ubuntu-22.04
-            arch: amd64
-            match_arch: x86-64
-            cross_compile: no
-            upload_path: upload/
-          - os: ubuntu-22.04
-            arch: arm64
-            match_arch: arm64
-            cross_compile: yes
-            upload_path: upload-cross-compile/
-
+        arch: [amd64, arm64]
     steps:
-      # https://github.com/docker/setup-buildx-action
-      - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
-
       - name: Checkout Source Code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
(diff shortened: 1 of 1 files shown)


.github/workflows/packages-e2e-tests.yaml
```diff
@@ -8,28 +8,12 @@ on:
...
-            match_arch: x86-64
-            cross_compile: no
-            upload_path: upload/
-          - os: ubuntu-22.04
-            arch: arm64
-            match_arch: arm64
-            cross_compile: yes
-            upload_path: upload-cross-compile/
-
+        arch: [amd64, arm64]
```
mtardy, line 15: you could directly put the name of the runners here if you want, we usually do it like that

.github/workflows/packages-e2e-tests.yaml
```diff
@@ -38,50 +22,32 @@ jobs:
...
-        if: ${{ matrix.cross_compile == 'yes' }}
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y qemu qemu-user-static binfmt-support
-          sudo update-binfmts --display
-          make TARGET_ARCH=${{ matrix.arch }} tarball
-          mkdir ${{ matrix.upload_path }}
-          mv ./build/${{ matrix.arch }}/linux-tarball/tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz ./${{ matrix.upload_path }}
+          mkdir upload/
+          mv ./build/${{ matrix.arch }}/linux-tarball/tetragon-${{ steps.tag.outputs.tag }}-${{ matrix.arch }}.tar.gz ./upload/
```
mtardy, line 30: ah at the same time it's useful to have the arch name here

.github/workflows/packages-e2e-tests.yaml
```diff
@@ -38,50 +22,32 @@ jobs:
...
           retention-days: 1
 
   standalone-tarball-tests:
     needs: [standalone-tarball-builds]
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ (matrix.arch == 'arm64' && 'ubuntu-22.04-arm') || 'ubuntu-22.04' }}
     strategy:
       matrix:
+        arch: [amd64, arm64]
         include:
```
mtardy, line 46: do you still need the include here?

mtardy (approved):
Thanks overall looks good, this might just fix the issue indeed. Just some small questions.

mtardy:
thanks for taking the time to make the CI better btw

.github/workflows/packages-e2e-tests.yaml
FedeDP: Yup exactly, wanted to avoid the `include` thing here since we only need the arch after all.

.github/workflows/packages-e2e-tests.yaml
FedeDP: AFAIK that's the only way to specify a `match_arch`, unless we directly use `arch: x86-64` and we adapt the `runs-on:` entry.
But i feel like this is more readable!

FedeDP: Oh you mean we could fully drop its check?
```
# Ensure that default native builds work
file /usr/local/bin/tetragon | grep ${{ matrix.match_arch }} -
```

mtardy: it felt like a hack but include apparently is also for adding config not just selecting what should be run from the matrix https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-expanding-configurations so I imagine it's fine

PR #4859, opened 2026-04-14
tetragon/windows: Fix process_exit event on Windows

ExceptionalHandler:
### Description
In some cases, with PROCESS_OPERATION_DELETE, the `ctx->process_id` is null and that causes the bpf program to exit prematurely without clearing command and image path maps. In this patch, we use the `bpf_get_current_pid_tgid()` helper to get the pid if it is not present in ctx.

### DIFF
--- a/bpf/windows/process_monitor.c
+++ b/bpf/windows/process_monitor.c
@@ -183,22 +183,24 @@ int ProcessMonitor(process_md_t *ctx)
 	} else if (ctx->operation == PROCESS_OPERATION_DELETE) {
 		struct process_exit_info_t process_exit_info;
 		int size = sizeof(process_exit_info);
-		uint32_t *pid = NULL;
+		uint32_t pid = 0;
 
-		if ((ctx) && ctx->process_id)
-			*pid = ctx->process_id;
-		else
-			return 0;
+		if (ctx->process_id != 0) {
+			pid = ctx->process_id;
+		} else {
+			uint64_t pid_tgid = bpf_get_current_pid_tgid();
 
+			pid = pid_tgid >> 32;
+		}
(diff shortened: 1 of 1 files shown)


andrewstrohman:
Hi @ExceptionalHandler 

> In some cases, with PROCESS_OPERATION_DELETE, the ctx->process_id is null a

Can you please explain a little more about the situations that lead to this? Can these same scenarios affect the `PROCESS_OPERATION_CREATE` operation as well?

aboch (approved):

ExceptionalHandler:
> Can you please explain a little more about the situations that lead to this? Can these same scenarios affect the PROCESS_OPERATION_CREATE operation as well?

The `ctx->process_id` corresponds to the `process_id` parameter in the callback registered via `PsSetCreateProcessNotifyRoutineEx` in the extension driver here:
https://github.com/microsoft/ntosebpfext/blob/b4b240c99321f43b073d792a2f8c194e46a1b6ce/ebpf_extensions/ntosebpfext/ntos_ebpf_ext_process.c#L332

In our testing we saw some cases where this parameter was 0 in the bpf program during process termination leading to the maps not being cleared on process exit. I have not investigated why the `process_id` was zero since it is anyways not recommended to expect the `process_id` parameter in callback to be always non-zero when process is exiting. 

During process creation the `process_id` and `create_info` structures are expected to be non null. The extension driver checks for non null `create_info` structure before invoking the bpf program. We have also not seen any issues with process creation during our testing.

FedeDP (approved):
LGTM but let's wait for @andrewstrohman :)

andrewstrohman (approved):

andrewstrohman:
Thanks for the explanation. LGTM!

PR #4862, opened 2026-04-15
ci: add prep job in vmtests to cache packages

### LINKED ISSUE #4860: Investigate vmtests workflow
Sometimes the vm tests workflow takes more than 30 minutes in the `apt-get update` stage. This should take less than a minute: 
https://github.com/cilium/tetragon/actions/runs/23806730293/job/69383738858?pr=4816

dangome3:
Fixes #4860

### Description
From time to time the 'apt-get update' step takes more than 30 minutes due to network variability in the jobs. This pattern allow to cache package dependencies so the 'VM Install test dependencies' step is as short as possible.

### DIFF
--- a/.github/workflows/vmtests.yml
+++ b/.github/workflows/vmtests.yml
@@ -9,6 +9,12 @@ on:
     paths-ignore:
       - 'docs/**'
 
+env:
+  VMTEST_APT_PACKAGES: >-
+    mmdebstrap libguestfs-tools qemu-utils qemu-system-x86 cpu-checker qemu-kvm
+    libvirt-daemon-system libvirt-clients bridge-utils virtinst virt-manager
+    isc-dhcp-client
+
 jobs:
   build:
     name: Build tetragon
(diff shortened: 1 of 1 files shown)


dangome3:
> I'd rather try to understand what is happening before applying a patch on top of this

Will move to draft while I get more context of the issue

mtardy (commented):
Not sure we want to do this but just for that I did some research and I think you could just leverage the `cache` github action to cache the `/var/cache/apt/archives` folder basically putting this before the `apt-get install` step.

```
  - run: echo "week=$(date +%Y-%W)" >> "$GITHUB_OUTPUT"
    id: week

  - uses: actions/cache@v4 # use real sha here
    with:
      path: /var/cache/apt/archives
      key: apt-${{ runner.os }}-${{ hashFiles('.github/workflows/vmtests.yaml') }}-${{ steps.week.outputs.week }}
      restore-keys: |
        apt-${{ runner.os }}-${{ hashFiles('.github/workflows/vmtests.yaml') }}-
        apt-${{ runner.os }}-
```

Issue #4579, opened 2026-01-28
Don't drop event when encountering arg read failures


andrewstrohman:
### What happened? How can we reproduce this?

See discussion [here](https://github.com/cilium/tetragon/pull/4327#discussion_r2732670290). We want to catch read arg failures in `read_arg` and expose the failure using arg_status, instead of aborting the event.

This issue was originally reported in [this issue](https://github.com/cilium/tetragon/issues/3728). That issue is a little more broad, so this issue has been created to focus on this specific aspect.

### Tetragon Version

All

### Kernel Version

All

### Relevant log output

```shell

```

kkourt:
@andrewstrohman I believe this issue can be closed. WDYT?

andrewstrohman:
> @andrewstrohman I believe this issue can be closed. WDYT?

Yes, I'll close it now.

I've created a follow up [here](https://github.com/cilium/tetragon/issues/4923) for doing some more error checking. Since this issue is specifically about dropped events, it makes sense to close this one, as that issue has been resolved.

PR #4863, opened 2026-04-16
Handle NULL pointers for string args

andrewstrohman:
This change lays the ground work for addressing https://github.com/cilium/tetragon/issues/4579. After I created the error arg type with [this change](https://github.com/cilium/tetragon/pull/4327), the problem described in that issue mutated such that we send events, but they are corrupted.

While looking into that problem, I found another issue with https://github.com/cilium/tetragon/pull/4327 where returnArgs can be overwritten when the first configured arg read results in an error arg.

Userspace probes have different string reading logic than kprobes. While events were not dropped in the uprobe case when NULL pointers are passed, empty strings were being reported in the event. So, I made uprobes consistent with kprobes in this regard, by returning an error arg when a NULL pointer is passed. This way, we can distinguish between empty strings and NULL pointers.

While creating the uprobe test I noticed a bug where I was masking against a macro that was not actually a mask. I corrected this as well.

I wanted to make the uprobe tests policytests, but I couldn't because there is no way to determine kfunc availability for policy tests yet.

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -276,8 +276,7 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 			size += 4;
 		} else {
 			/* If filter specification is fd type then we
-			 * expect the fd has been previously followed
-			 * otherwise drop the event.
+			 * prevent the filter from matching
 			 */
 			return -1;
 		}
@@ -345,7 +344,7 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 		size = 0;
 		break;
 	}
-	return size + sizeof(arg_status_t);
+	return size;
 }
 
 FUNC_INLINE long
@@ -423,7 +422,7 @@ __read_arg_2(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 		size = 0;
 		break;
 	}
-	return size + sizeof(arg_status_t);
+	return size;
 }
 
 /**
@@ -450,6 +449,7 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
 	const struct path *path_arg = 0;
 	struct path path_buf;
 	int zero = 0;
+	int ret;
 
 	e = map_lookup_elem(&process_call_heap, &zero);
 	if (!e)
@@ -467,22 +467,32 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
 	args = args_off(e, orig_off);
 
 	path_arg = get_path(type, arg, &path_buf);
-	if (path_arg)
-		return copy_path(args, path_arg) + sizeof(arg_status_t);
 	/*
-	 * Separate argument processing based on the process const
-	 * for 4.19 kernels..
+	 * If not path_ag, separate argument processing based on the process
+	 * const for 4.19 kernels..
 	 */
-	if (process == __READ_ARG_1)
-		return __read_arg_1(ctx, type, orig_off, arg, argm, args);
-	if (process == __READ_ARG_2)
-		return __read_arg_2(ctx, type, orig_off, arg, argm, args);
+	if (path_arg) {
+		ret = copy_path(args, path_arg);
+	} else if (process == __READ_ARG_1) {
+		ret = __read_arg_1(ctx, type, orig_off, arg, argm, args);
+	} else if (process == __READ_ARG_2) {
+		ret = __read_arg_2(ctx, type, orig_off, arg, argm, args);
+	} else {
+		/* .. and the rest of the world */
+		if (is_read_arg_1(type))
+			ret = __read_arg_1(ctx, type, orig_off, arg, argm, args);
+		else
+			ret = __read_arg_2(ctx, type, orig_off, arg, argm, args);
+	}
 
-	/* .. and the rest of the world */
-	if (is_read_arg_1(type))
-		return __read_arg_1(ctx, type, orig_off, arg, argm, args);
-	else
-		return __read_arg_2(ctx, type, orig_off, arg, argm, args);
+	if (ret < 0) {
+		e->arg_status[index & MAX_POSSIBLE_ARGS_MASK] = -1;
+		/* update the arg status to reflect the detected fault */
+		write_arg_status(e, orig_off - sizeof(arg_status_t), e->arg_status[index & MAX_POSSIBLE_ARGS_MASK]);
+		return sizeof(arg_status_t);
+	}
+
+	return ret + sizeof(arg_status_t);
 }
 
 FUNC_INLINE long get_pt_regs_arg_syscall(struct pt_regs *ctx, __u16 offset, __u8 shift)
@@ -657,13 +667,6 @@ generic_process_event(void *ctx, struct bpf_map_def *tailcals, int process)
 		errv = generic_read_arg(ctx, index, total, tailcals, process);
 		if (errv > 0)
 			total += errv;
-		/* Follow filter lookup failed so lets abort the event.
-		 * From high-level this is a filter and should be in the
-		 * filter block, but its just easier to do here so lets
-		 * do it where it makes most sense.
-		 */
-		if (errv < 0)
-			return filter_args_reject(e->func_id);
 	}
 	e->common.size = total;
 	/* Continue to process other arguments. */
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -1240,6 +1240,7 @@ FUNC_INLINE int generic_retprobe(void *ctx, struct bpf_map_def *calls, unsigned 
 	ty_arg = config->argreturn;
 	do_copy = config->argreturncopy;
 	if (ty_arg) {
+		e->arg_status[0] = 0;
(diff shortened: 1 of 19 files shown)


bpf/process/types/basic.h
```diff
@@ -365,8 +365,8 @@ FUNC_INLINE long copy_strings(char *args, char *arg, int max_size)
 	// So add one to the length to allow for it. This should
 	// result in us honouring our max_size correctly.
 	size = probe_read_str(&args[4], max_size + 1, arg);
-	if (size <= 0)
-		return invalid_ty;
+	if (size < 0)
+		return size;
```
kkourt, line 369: This changes a `<=` check to a `<` check, and below we do `size--`. Is there  a chance for an overflow?

bpf/process/generic_calls.h
```diff
@@ -657,13 +667,6 @@ generic_process_event(void *ctx, struct bpf_map_def *tailcals, int process)
 		errv = generic_read_arg(ctx, index, total, tailcals, process);
 		if (errv > 0)
 			total += errv;
-		/* Follow filter lookup failed so lets abort the event.
-		 * From high-level this is a filter and should be in the
-		 * filter block, but its just easier to do here so lets
-		 * do it where it makes most sense.
-		 */
-		if (errv < 0)
-			return filter_args_reject(e->func_id);
```
kkourt, line 666: Why is this removed? Can you please add details in the commit message?

bpf/process/generic_calls.h
```diff
@@ -467,22 +467,32 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
...
-		return __read_arg_1(ctx, type, orig_off, arg, argm, args);
-	if (process == __READ_ARG_2)
-		return __read_arg_2(ctx, type, orig_off, arg, argm, args);
-
-	/* .. and the rest of the world */
-	if (is_read_arg_1(type))
-		return __read_arg_1(ctx, type, orig_off, arg, argm, args);
-	else
-		return __read_arg_2(ctx, type, orig_off, arg, argm, args);
+	if (path_arg) {
```
kkourt, line 474: Why is path_arg special? Can you add a commit or the specifics in the commit message?

pkg/sensors/tracing/tracepoint_test.go
```diff
@@ -479,7 +479,7 @@ func TestLoadTracepointSensor(t *testing.T) {
 		sensorMaps = append(sensorMaps, tus.SensorMap{Name: "tp_calls", Progs: []uint{0, 1, 2, 3, 4, 6, 7}})
 
 		// all but generic_tracepoint_event,generic_tracepoint_filter
-		sensorMaps = append(sensorMaps, tus.SensorMap{Name: "retprobe_map", Progs: []uint{1, 2, 6, 7}})
+		sensorMaps = append(sensorMaps, tus.SensorMap{Name: "retprobe_map", Progs: []uint{1, 6, 7}})
```
kkourt, line 482: > This is the problematic code path that caused dropped events, which the previous commit fixed.

Thanks for fixing this, but it's unclear to me what the issue was. Can you add some details in the commit message?

pkg/sensors/tracing/args_linux.go
```diff
@@ -112,7 +112,11 @@ func getArgStatus(r *bytes.Reader) (*api.MsgGenericKprobeArgError, error) {
 	}
 
 	if status != 0 {
-		arg.Message = strconv.FormatUint(uint64(status), 10)
+		if status == ^uint32(0) {
```
kkourt, line 115: Is this the return from `probe_read_str`? Or we have a convention for status?

kkourt (commented):
Thanks Andy!

I'm having some difficulty following this PR. I'm not super familiar with the code, so maybe I'm missing something obvious. Could you please add some explanations to help me understand the change (e.g., iin commit messages)?

bpf/process/types/basic.h
andrewstrohman: `bpf_probe_read_str` will not return 0 under any condition. See [here](https://docs.ebpf.io/linux/helper-function/bpf_probe_read_str/#:~:text=On%20success%2C%20the%20strictly%20positive,On%20error%2C%20a%20negative%20value.). It either returns a positive number, to indicate success and the number of bytes copieid (including NULL terminator) or it returns a negative number to indicate failure. 

The reason the success case is a positive number (as opposed to 0 being valid) is because it includes the NULL terminator.

It would have been OK to leave the condition as it was, so if you prefer it to remain that way, I'll revert that part. I don't think there is any functional/behavioral difference though.

bpf/process/generic_calls.h
andrewstrohman: I removed this because it's dead code now. `generic_read_arg` will no longer return a negative return value under after this change, so I thought removing it would prevent confusion. I have added this information into the commit message.

bpf/process/generic_calls.h
andrewstrohman: path_arg is not special. In the future, I'd like `copy_path` to return a negative value if we encounter a fault. If I kept the early return for `if (path_arg)`, then I would need to add the fault detection logic and arg status fixup for that early return too. This way, there is a single point where we evaluate if a fault occurred and update the status accordingly.

I've updated the commit message to describe this thinking.

pkg/sensors/tracing/tracepoint_test.go
andrewstrohman: Prior to the arg status change, when `generic_read_arg` returned a negative value (ie NULL string), we would `return filter_args_reject(e->func_id)`.  The `return` part was the cause of the event drop. The calling of `filter_args_reject(e->func_id)` made the generic_tracepoint_process_event program reference the retprobe_map.

Now that `generic_read_arg` will no longer return negative values, the dropping issue is fixed, and the program no longer has a connection to the `retprobe_map` map.

I've tried to improve the commit message to explain this.

pkg/sensors/tracing/args_linux.go
andrewstrohman: With my current approach, on the BPF side, if a negative return code is encountered from `probe_read_str`, we set arg status code to -1, regardless of the reason for error.

We could preserve/plumb the error codes in the arg status if you think that's valuable. We currently don't relay the error code for the resolve dereference error case either. So if we want to plumb the return code so that it can be reflected in the event, I'd like to do the same for resolve induced errors as well.

bpf/process/types/basic.h
kkourt: If it's not possible to return `0`, no need to change it, thanks!

pkg/sensors/tracing/uprobe_test.go
```diff
@@ -1127,3 +1127,63 @@ spec:
 	err = jsonchecker.JsonTestCheck(t, checker)
 	require.NoError(t, err)
 }
+
+func TestUprobeNULLStringAndReturnArg(t *testing.T) {
```
kkourt, line 1221: Is there a reason that we are not using the new policytest format?

kkourt (approved):
Just a comment on the test, but otherwise this looks good to me, thanks!

bpf/process/user_preload.h
```diff
@@ -140,7 +140,7 @@ try_preload_arg(int idx, struct preload_arg_data *data)
 {
 	asm volatile("%[idx] &= %1 ;\n"
 		     : [idx] "+r"(idx)
-		     : "i"(MAX_POSSIBLE_ARGS));
+		     : "i"(MAX_POSSIBLE_ARGS_MASK));
```
olsajiri, line 143: nice, so we never caught it because we always used index 0? I guess we should add more tests for other than zero arg/data

olsajiri (approved):

pkg/sensors/tracing/uprobe_test.go
andrewstrohman: Yes, I put the reason in the description of this PR:

> I wanted to make the uprobe test a policytest, but I couldn't because there is no way to determine kfunc availability for policy tests yet.

Do you know of a way that I can do this? If not, I can think about adding support for this detection in future, but I didn't want to gate these fixes on that improvement.

bpf/process/user_preload.h
andrewstrohman: > so we never caught it because we always used index 0?

Yes, this issue only affected args configured at indexes 2 and 3.

> I guess we should add more tests for other than zero arg/data

I've added additional tests to the end of this series. They check that we can handle string args at any position within the function signature and within any configuration position within the args section of the policy.

Please let me know what you think.

pkg/sensors/tracing/uprobe_test.go
olsajiri: there's WithSkip where you can do the detection IIUC the problem

andrewstrohman: Thanks. I'm aware of WithSkip, but the code within WithSkip is run on the local machine, not the agent that's being tested. We need a way to query the agent's features. `si.AgentInfo.Probes` was relayed from the agent to local, but it does not include the kfunc availablity of the agent.

kkourt: OK, thanks! @andrewstrohman can you create an issue for this?

andrewstrohman: Yes, [here](https://github.com/cilium/tetragon/issues/4913) is the issue.

PR #4881, opened 2026-04-17
uprobe-test-1: have libs relative to binary

kkourt:
We want to be able to execute uprobe-test-1 from different directories. That is, copy the binary and the .so file in different directories and execute them from them.

From https://sourceware.org/binutils/docs/ld/Options.html

-rpath-link=dir
 ...
    The tokens $ORIGIN and $LIB can appear in these search directories.
    They will be replaced by the full path to the directory containing
    the program or shared object in the case of $ORIGIN and either ‘lib’
    - for 32-bit binaries - or ‘lib64’ - for 64-bit binaries - in the case of $LIB.


kkourt:
CC: @andrewstrohman

FedeDP (approved):

olsajiri (approved):

PR #4884, opened 2026-04-20
process/cache: add metric and fix early deletion detection in GC

lbcristaldo:
### Description
This PR adds a new metric tetragon_process_cache_early_deletions_total to track cases where the process cache GC attempts to delete a process that was already marked as deleted.
While adding the metric, a bug was found: the deleted color check in the deleteChan handler was unreachable because color == deleted also satisfies color != inUse, causing the handler to exit early. The fix moves the deleted check before the inUse check so the metric is actually reachable.
Previously this case only had a silent TODO comment.

**Testing**
go test ./pkg/process -run TestProcessCache -v
go test ./pkg/process/...
go test -race ./pkg/process/...

### Changelog
```release-note
Add `tetragon_process_cache_early_deletions_total` metric to improve observability of process cache GC early deletions.
```

### DIFF
--- a/pkg/process/cache_test.go
+++ b/pkg/process/cache_test.go
@@ -4,11 +4,15 @@
 package process
 
 import (
+	"strings"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 	"google.golang.org/protobuf/types/known/wrapperspb"
+
+	"github.com/prometheus/client_golang/prometheus/testutil"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/defaults"
@@ -18,6 +22,7 @@ func TestProcessCache(t *testing.T) {
 	// add a process to the cache.
 	cache, err := NewCache(10, defaults.DefaultProcessCacheGCInterval)
 	require.NoError(t, err)
+	defer cache.purge()
 	pid := wrapperspb.UInt32Value{Value: 1234}
 	execID := "process1"
 	proc := ProcessInternal{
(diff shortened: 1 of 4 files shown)


FedeDP (approved):
LGTM, love nice little (as in LoC) fixes!

cc @dwindsor @sglushko that recently worked on this area of the code! 🙏

FedeDP:
Oh sorry, i think you'll have to update the metrics documentation via `make metrics-docs`.

sglushko:
> LGTM, love nice little (as in LoC) fixes!
> 
> cc @dwindsor @sglushko that recently worked on this area of the code! 🙏

> The fix moves the deleted check before the inUse check so the metric is actually reachable.

Hi, yes, this looks similar to what I found as well.

lbcristaldo:
> Oh sorry, i think you'll have to update the metrics documentation via `make metrics-docs`.

Done. I ran `make metrics-docs` and fixed the goimports formatting. Thanks!

olsajiri (commented):
please provide details into commit's changelog

mtardy (approved):

pkg/process/cache_test.go
```diff
@@ -46,3 +50,28 @@ func TestProcessCache(t *testing.T) {
...
+	proc.color = deleted
+	cache.deletePending(&proc)
+
+	time.Sleep(50 * time.Millisecond)
+
+	expected := strings.NewReader(`# HELP tetragon_process_cache_early_deletions_total Number of times the GC attempted to delete a process already marked as deleted. May indicate the GC is deleting processes too early.
+# TYPE tetragon_process_cache_early_deletions_total counter
+tetragon_process_cache_early_deletions_total 1
+`)
+	require.NoError(t, testutil.CollectAndCompare(processCacheEarlyDeletions, expected))
```
olsajiri, line 78: should we call cache.purge in here? same for the existing test

lbcristaldo: Good point! Added defer cache.purge() to both tests

tpapagian (approved):

PR #4885, opened 2026-04-20
cleanup: drop TracingPolicyNamespaced interface

### LINKED ISSUE #4864: Drop TracingPolicyNamespaced interface
Instead, we can just add the `TpNamespace()` method to the `TracingPolicy` interface, and return empty string for non-namespaced TPs.
This leads to at least 2 obvious improvements:
* less code, thus less burden 
* drop lots of runtime interface assertions/checks that are inherently slow

FedeDP:
Fixes #4864

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -585,14 +585,9 @@ func addTracingPolicy(ctx context.Context, file string) error {
 		return err
 	}
 
-	namespace := ""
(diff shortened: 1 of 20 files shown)


olsajiri (approved):
nice cleanup, thanks

mtardy (approved):
Indeed cool cleanup :) thanks

PR #4888, opened 2026-04-20
tetragon: Add missing setup for matchParentBinaries selector

olsajiri:
Tracing sensors (kprobe, lsm, uprobe, usdt, tracepoint) include
pfilter.h which references the PARENTS_MAP_ENABLED rodata constant to
gate parent-binary selector matching in BPF. The constant was only set
for the execve program in pkg/sensors/base/base.go, leaving it at zero
in all tracing sensors even when --parents-map-enabled is set.

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -504,13 +504,14 @@ type addKprobeIn struct {
 }
 
 type hasMaps struct {
-	stackTrace bool
-	rateLimit  bool
-	enforcer   bool
-	override   bool
-	sockTrack  bool
-	selector   bool
-	fentry     bool
+	stackTrace     bool
+	rateLimit      bool
+	enforcer       bool
+	override       bool
+	sockTrack      bool
+	selector       bool
+	fentry         bool
+	parentBinaries bool
 }
 
 // hasMapsSetup setups the has maps for the per policy maps. The per kprobe maps
@@ -523,6 +524,7 @@ func hasMapsSetup(spec *v1alpha1.TracingPolicySpec, kprobes []v1alpha1.KProbeSpe
 		has.sockTrack = has.sockTrack || selectors.HasSockTrack(&kprobe)
 		has.override = has.override || selectors.HasOverride(kprobe.Selectors)
 		has.selector = has.selector || selectors.HasSelector(&kprobe)
+		has.parentBinaries = has.parentBinaries || selectors.HasMatchParentBinaries(kprobe.Selectors)
 	}
 	return has
 }
@@ -644,9 +646,7 @@ func createGenericKprobeSensor(
 		maps = append(maps, program.MapUserFrom(base.RingBufEvents))
 	}
 
-	if option.Config.ParentsMapEnabled {
-		maps = append(maps, program.MapUserFrom(base.ParentBinariesMap))
-	}
+	maps = setParentsMap(progs, maps, has.parentBinaries)
 
 	return &sensors.Sensor{
 		Name:      name,
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -4207,6 +4207,10 @@ func TestKprobeMatchParentBinariesCombined(t *testing.T) {
 	policytest.AllPolicyTests.DoObserverTest(t, "kprobe-match-parent-binaries-combined", nil)
 }
 
+func TestKprobeMatchParentBinariesFollowChildrenCombined(t *testing.T) {
+	policytest.AllPolicyTests.DoObserverTest(t, "kprobe-match-parent-binaries-followchildren-combined", nil)
+}
+
 func TestMatchParentBinariesSetParentsMap(t *testing.T) {
 	specDisabled := &v1alpha1.TracingPolicySpec{
 		KProbes: []v1alpha1.KProbeSpec{
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -4203,6 +4203,10 @@ func TestKprobeMatchParentBinaries(t *testing.T) {
 	testKprobeMatchParentBinaries(t, false)
 }
 
+func TestKprobeMatchParentBinariesCombined(t *testing.T) {
+	policytest.AllPolicyTests.DoObserverTest(t, "kprobe-match-parent-binaries-combined", nil)
+}
+
 func TestMatchParentBinariesSetParentsMap(t *testing.T) {
 	specDisabled := &v1alpha1.TracingPolicySpec{
 		KProbes: []v1alpha1.KProbeSpec{
--- a/pkg/sensors/tracing/kprobe_test.go
+++ b/pkg/sensors/tracing/kprobe_test.go
@@ -4203,6 +4203,82 @@ func TestKprobeMatchParentBinaries(t *testing.T) {
 	testKprobeMatchParentBinaries(t, false)
 }
 
+func TestMatchParentBinariesSetParentsMap(t *testing.T) {
+	specDisabled := &v1alpha1.TracingPolicySpec{
+		KProbes: []v1alpha1.KProbeSpec{
+			{
+				Call:    "test_symbol",
+				Syscall: false,
+			},
+		},
+	}
+	specEnabled := &v1alpha1.TracingPolicySpec{
+		KProbes: []v1alpha1.KProbeSpec{
+			{
+				Call:    "test_symbol",
+				Syscall: false,
+				Selectors: []v1alpha1.KProbeSelector{
+					{
+						MatchParentBinaries: []v1alpha1.BinarySelector{
+							{
+								Operator: "In",
+								Values:   []string{"test"},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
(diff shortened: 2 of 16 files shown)


kobrineli:
Oh, didnt notice that.. thanks!
But how did it work at all then?
I've tested it locally, and tests for parent binaries also passed

olsajiri:
> Oh, didnt notice that.. thanks! But how did it work at all then? I've tested it locally, and tests for parent binaries also passed

yea, I think the existing tests use the positive assertions (event present), but do not check that  some event was filtered out because of matchParentBinary, like what TestKprobeMatchParentBinariesCombined does:

    - lseek-pipe run via bash parent: events arrive (selector matches)
    - lseek-pipe run via sh parent: no events arrive (selector filters out)

olsajiri:
@tdaudi @ethical-buddy

tdaudi:
Hm, I'm still receiving unrelated events with the policy I used in https://github.com/cilium/tetragon/issues/5069 after I compiled from your branch.

olsajiri:
> Hm, I'm still receiving unrelated events with the policy I used in #5069 after I compiled from your branch.

I vibe coded the test https://github.com/cilium/tetragon/pull/4888/changes/3aab0df36cc5ee57a9c01b3fbf131a9e399de3b7 that tries to mimic the the issue described in #5069

tdaudi:
Awesome thanks ! That seems to fix the issue

pkg/sensors/tracing/generickprobe.go
```diff
@@ -660,9 +662,7 @@ func createGenericKprobeSensor(
 		maps = append(maps, program.MapUserFrom(base.RingBufEvents))
 	}
 
-	if option.Config.ParentsMapEnabled {
-		maps = append(maps, program.MapUserFrom(base.ParentBinariesMap))
-	}
+	maps = setParentsMap(progs, maps, has.parentBinaries)
```
kkourt, line 649: Do we reject the policy if the selectors have parentBinaries but the config option is not set?

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -4207,6 +4207,80 @@ func TestKprobeMatchParentBinaries(t *testing.T) {
...
+			if m.Name == base.ParentBinariesMap.Name {
+				return true
+			}
+		}
+		return false
+	}
+
+	t.Run("disabled", func(t *testing.T) {
+		option.Config.ParentsMapEnabled = false
+		t.Cleanup(func() { option.Config.ParentsMapEnabled = false })
```
kkourt: nit: let's save the previous value here and restore it in case we ever change the default.

pkg/sensors/tracing/kprobe_test.go
```diff
@@ -4207,6 +4207,80 @@ func TestKprobeMatchParentBinaries(t *testing.T) {
...
+		for _, p := range sensor.Progs {
+			_, ok := p.RewriteConstants["PARENTS_MAP_ENABLED"]
+			assert.False(t, ok, "PARENTS_MAP_ENABLED should not be set for prog %s", p.Name)
+		}
+		assert.False(t, hasParentsMap(sensor), "ParentBinariesMap should not be in sensor maps")
+	})
+
+	t.Run("enabled", func(t *testing.T) {
+		option.Config.ParentsMapEnabled = true
+		t.Cleanup(func() { option.Config.ParentsMapEnabled = false })
```
kkourt: ditto: let's restore the previous default value.

tests/policytests/kprobe_matchparentbinaries.go
```diff
@@ -0,0 +1,83 @@
...
+var _ = policytest.NewBuilder("kprobe-match-parent-binaries-combined").
+	WithLabels("kprobes").
+	WithSkip(func(_ *policytest.SkipInfo) string {
+		if !config.EnableLargeProgs() {
+			return "kernels without large progs do not support matchParentBinaries selector"
+		}
+		return ""
+	}).
+	WithSetup(func() func() {
+		option.Config.ParentsMapEnabled = true
```
kkourt, line 27: This does not work for when executing the test with the CLI (tetra policytest). One option might be to check the agent configuration and skip the test if parents are not enabled.

kkourt (changes requested):

pkg/sensors/tracing/generickprobe.go
kobrineli: Yeah, we do

https://github.com/cilium/tetragon/blob/848c3f2151804791db414163266f266741af7675/pkg/selectors/kernel.go#L1615

litios:
Thanks for the PR!

I found a weird issue with this when using `followChildren` when a process doing execve after fork:

```yaml
spec:
  kprobes:
  - call: "security_bprm_creds_from_file"
    syscall: false
    args:
    - index: 0
      type: "linux_binprm"
      label: "binary"
    data:
    - index: 0
      type: linux_binprm
      source: current_task
      resolve: real_parent.real_parent.mm.exe_file
      label: grand_parent
    selectors:
    - matchParentBinaries:
      - operator: "In"
        values:
        - "/usr/bin/node"
        followChildren: true
      matchArgs:
      - index: 0
        operator: "Postfix"
        values:
        - "ssh"
      matchActions:
      - action: Override
        argError: -1
```

This doesn't fire but it does with `matchBinaries`:

```yaml
spec:
  kprobes:
  - call: "security_bprm_creds_from_file"
    syscall: false
    args:
    - index: 0
      type: "linux_binprm"
      label: "binary"
    data:
    - index: 0
      type: linux_binprm
      source: current_task
      resolve: real_parent.real_parent.mm.exe_file
      label: grand_parent
    selectors:
    - matchBinaries:
      - operator: "In"
        values:
        - "/usr/bin/node"
        followChildren: true
      matchArgs:
      - index: 0
        operator: "Postfix"
        values:
        - "ssh"
      matchActions:
      - action: Override
        argError: -1
```


```json
{
  "process_kprobe": {
    "process": {
      "exec_id": "dWJ1bnR1MjQwNDoxMTM3NTE2ODgzNDYxOTQ6MTI1MTQx",
      "pid": 125141,
      "uid": 0,
      "cwd": "/test/test-package",
      "binary": "/usr/bin/python3",
      "arguments": "/test/test-package/scripts/install.py",
      "flags": "execve inInitTree",
      "start_time": "2026-06-23T14:23:25.534965434Z",
      "auid": 4294967295,
      "docker": "10fbd4f58218c73d362c64fe64dff85",
      "parent_exec_id": "dWJ1bnR1MjQwNDoxMTM3NTE2NDM1NTAzODc6MTI1MTMy",
      "refcnt": 1,
      "tid": 125141,
      "in_init_tree": true
    },
    "parent": {
      "exec_id": "dWJ1bnR1MjQwNDoxMTM3NTE2NDM1NTAzODc6MTI1MTMy",
      "pid": 125132,
      "uid": 0,
      "cwd": "/test/test-package",
      "binary": "/usr/bin/python3",
      "arguments": "/test/test-package/scripts/install.py",
      "flags": "execve clone inInitTree",
      "start_time": "2026-06-23T14:23:25.490169333Z",
      "auid": 4294967295,
      "docker": "10fbd4f58218c73d362c64fe64dff85",
      "parent_exec_id": "dWJ1bnR1MjQwNDoxMTM3NTE1MjkzODE2NjU6MTI1MTI1",
      "tid": 125132,
      "in_init_tree": true
    },
    "function_name": "security_bprm_creds_from_file",
    "args": [
      {
        "linux_binprm_arg": {
          "path": "/usr/bin/ssh",
          "permission": "-rwxr-xr-x"
        },
        "label": "binary"
      }
    ],
    "action": "KPROBE_ACTION_OVERRIDE",
    "policy_name": "node-cli-children-suspicious-exec",
    "return_action": "KPROBE_ACTION_POST",
    "data": [
      {
        "file_arg": {
          "path": "/usr/bin/node",
          "permission": "-rwxr-xr-x"
        },
        "label": "grand_parent"
      }
    ]
  },
  "node_name": "ubuntu2404",
  "time": "2026-06-23T14:23:25.535930051Z"
}
```

olsajiri:
> I found a weird issue with this when using followChildren when a process doing execve after fork:

@litios I'll need more info on that, like what did you run and what you expected to happen

litios:
In this case:

```yaml
spec:
  kprobes:
  - call: "security_bprm_creds_from_file"
    syscall: false
    args:
    - index: 0
      type: "linux_binprm"
      label: "binary"
    data:
    - index: 0
      type: linux_binprm
      source: current_task
      resolve: real_parent.real_parent.mm.exe_file
      label: grand_parent
    selectors:
    - matchParentBinaries:
      - operator: "In"
        values:
        - "/usr/bin/node"
        followChildren: true
      matchArgs:
      - index: 0
        operator: "Postfix"
        values:
        - "ssh"
      matchActions:
      - action: Override
        argError: -1
 ```

I designed the policy with the idea that this should match when any of the children of `node` try to execute `ssh` binary, since `followChildren: true`, but `node` should be able to.

So I read it like:
1) If `node` executes `ssh`, it shouldn't fire/block.
2) If `node` executes `python3` which executes `ssh`, it should block.
3) If we have the following execution pattern `node` -> `python3` -> `java` -> `ssh`, it should block since `java` grand-parent is `node`

I feel that 3) should match with `matchParentBinaries` + `followChildren`, unless I'm misunderstanding the feature.

> 3) worked fine with `matchBinaries` and `followChildren` (see event in comment above), but it doesn't allow to express 1)

PR #4889, opened 2026-04-21
Fix null spec.{pod, container, host}Selectors

tpapagian:
After https://github.com/cilium/tetragon/pull/4814 when a user applies a policy similar to the following

```
spec:
  hostSelector: null
  kprobes:
```

we expect that hostSelector is null and podSelector to be {}. This is the case when we run:
kubectl create -f pol.yaml
kubectl apply --server-size=true -f pol.yaml
and in all cases outside of k8s.

When a user loads such a policy with kubectl apply -f pol.yaml it seems that the client side of k8s sees hostSelector to be null and thus it is applying the default value which is {}.

A way to fix that without changing the user API, we use the "last-applied-configuration" annotation which keeps the original policy that the user applies. If we find out that spec.{pod, container, host}Selectors was null, we override that value to be nul again. It seems that in that in the client-side apply we cannot affect the order of when defaults are applied.

Tested manually and this seems to work as expected in all cases.

### DIFF
--- a/pkg/watcher/crdwatcher/lastapplied.go
+++ b/pkg/watcher/crdwatcher/lastapplied.go
@@ -1,0 +1,59 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build !nok8s
+
+package crdwatcher
+
+import (
+	"encoding/json"
+
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
+)
+
+type Tp[T any] interface {
+	TpSpec() *v1alpha1.TracingPolicySpec
+	DeepCopy() T
+	GetAnnotations() map[string]string
+}
+
+func restoreNullSelectors[T Tp[T]](tp T) T {
+	ret := tp.DeepCopy()
+	restoreNullSelectorsFromLastApplied(ret.GetAnnotations(), ret.TpSpec())
+	return ret
+}
+
+// restoreNullSelectorsFromLastApplied recovers explicit null selectors from the
+// kubectl last-applied annotation. This is needed because client-side
+// `kubectl apply` removes the distinction between explicit null and omitted
+// fields for CRD defaults before the object reaches Tetragon.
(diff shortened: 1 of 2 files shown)


FedeDP (approved):

mtardy (commented):
> A way to fix that without changing the user API, we use the "last-applied-configuration" annotation which keeps the original policy that the user applies. If we find out that spec.{pod, container, host}Selectors was null, we override that value to be nul again. It seems that in that in the client-side apply we cannot affect the order of when defaults are applied.

It seems a bit hacky, would changing the default value work? Even if it's a breaking change maybe it's necessary in this case because we change the behavior anyway nop? wdyt?

tpapagian:
> It seems a bit hacky, would changing the default value work? Even if it's a breaking change maybe it's necessary in this case because we change the behavior anyway nop? wdyt?

We need a way to distinguish between everything and nothing. 

The behaviour that I have in mind (and implemented in https://github.com/cilium/tetragon/pull/4814) is:
```yaml
spec:
  hostSelector: {}
```
to denote everything and 
```yaml
spec:
  hostSelector: null
```
to denote nothing. And "everything" (`{}`) be the default value.

The issue here that null is not handled as expected in all cases regarding on when k8s applies the default values.

Another way to fix that (I have tested that in https://github.com/cilium/tetragon/pull/4887) is:
```yaml
spec:
  hostSelector: {}
```
to denote everything and 
```yaml
spec:
  hostSelector:
    empty: true
```
to denote nothing. In that case, we will not use null in any case. This should be also applied in all of s`pec.{pod, container, host}Selectors` and the [PR](https://github.com/cilium/tetragon/pull/4887) that I provided before managed to keep the same syntax that we already have in podSelector to define labels etc.

The reason that choose to go with this PR is that it keeps the tracing policy simpler. But I am fine with the other way if this sounds better to you.

mtardy (approved):
By searching we found https://github.com/kubernetes/kubernetes/issues/129456 this issue. It seems that the client side cannot really know that a field is nullable so it just blindly apply the default, and the `--server-side` should be preferred but I guess they don't want to break existing behaviors. So I think the hack is okay since this is kinda broken.

kkourt (approved):

PR #4890, opened 2026-04-21
improved logging on event messages

kkourt:
```release-note
tracing: add policy information when logging errors in event parsing
```


FedeDP (approved):
I really like it, thanks!

PR #4894, opened 2026-04-22
ci: remove azure apt mirrors

### LINKED ISSUE #4860: Investigate vmtests workflow

dangome3:
### Description

From https://github.com/actions/runner-images/issues/12949

> Package downloads from http://azure.archive.ubuntu.com/ubuntu on
GitHub-hosted runners can sometimes be slow. This happens due to performance issues with the Azure Ubuntu archive mirror, not with the GitHub Actions runner images.

This change removes the azure.archive mirrors so the apt-get update uses other mirrors.

Fixes https://github.com/cilium/tetragon/issues/4860

### DIFF
--- a/.github/workflows/vmtests.yml
+++ b/.github/workflows/vmtests.yml
@@ -96,6 +96,16 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
+    - name: Remove slow Azure apt mirrors
+      run: |
+        # The default 'azure.archive.ubuntu.com' mirrors can be really slow.
+        # Removing azure.archive will pioritize the official Ubuntu mirrors.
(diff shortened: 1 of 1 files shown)


FedeDP:
Thanks!
Let me re-trigger the vmtests 4-5 times to check that PR is indeed working :)

mtardy (approved):
It looks good from your 3 runs, I'd be happy to merge it like this since we know it worked as well in LVH :)

FedeDP (approved):
Let's go then!

PR #4895, opened 2026-04-23
Load programs that reference kfuncs when an external BTF file is used.

andrewstrohman:
We had a report from a user who did not enable CONFIG_DEBUG_INFO_BTF in their kernel, and as a result, was unable to add kprobe policies because the loader rejected the collection. The loader rejected the collection because for kfunc fixups, the BTF file must be located in hard-coded paths.

The fix for the loader is to poison kfunc calls when an external BTF file is used, instead of refusing to load the collection. We cannot allow the poisoned kfunc calls to be reachable, so additional checks were put in place within the bpf code to prevent this.

Finally, the HasKfunc() function was updated to detect when a external BTF is used. This way, features that depend on kfuncs will have the correct detection outcome, so that we don't allow users to load policies that depend on kfuncs when an external BTF is used.

## Testing
The following cilium/ebpf patch was used to simulate the lack of `/sys/kernel/btf/vmlinux`:
```
diff --git a/btf/kernel.go b/btf/kernel.go
index bb7368b..488a5f1 100644
--- a/btf/kernel.go
+++ b/btf/kernel.go
@@ -127,7 +127,7 @@ func loadKernelSpec() (*Spec, error) {
                return nil, internal.ErrNotSupportedOnOS
        }

-       fh, err := os.Open("/sys/kernel/btf/vmlinux")
+       fh, err := os.Open("/sys/kernel/btf/vmlinux-does-not-exist")
        if err == nil {
                defer fh.Close()

@@ -178,7 +178,7 @@ func loadKernelModuleSpec(module string, base *Spec) (*Spec, error) {
                return nil, fmt.Errorf("invalid module name %q", module)
        }

-       fh, err := os.Open(filepath.Join("/sys/kernel/btf", module))
+       fh, err := os.Open(filepath.Join("/sys/kernel/btf-does-not-exist", module))
        if err != nil {
                return nil, err
        }
@@ -200,15 +200,7 @@ func findVMLinux() (*os.File, error) {

        // use same list of locations as libbpf
        // https://github.com/libbpf/libbpf/blob/9a3a42608dbe3731256a5682a125ac1e23bced8f/src/btf.c#L3114-L3122
-       locations := []string{
-               "/boot/vmlinux-%s",
-               "/lib/modules/%s/vmlinux-%[1]s",
-               "/lib/modules/%s/build/vmlinux",
-               "/usr/lib/modules/%s/kernel/vmlinux",
-               "/usr/lib/debug/boot/vmlinux-%s",
-               "/usr/lib/debug/boot/vmlinux-%s.debug",
-               "/usr/lib/debug/lib/modules/%s/vmlinux",
-       }
+       locations := []string{}

        for _, loc := range locations {
                file, err := os.Open(fmt.Sprintf(loc, release))
--
2.43.0
``` 

I tested this on a 5.4 and 6.18.12 kernels to confirm that I could start tetragon and load kprobe policies.

## Related PR
https://github.com/cilium/ebpf/pull/1990


kkourt:
Thanks @andrewstrohman!

I've moved this to draft, since I don't think we want to merge it with:

```
> # github.com/cilium/ebpf v0.21.0 => github.com/andrewstrohman/ebpf v0.0.0-20260422233615-d6968610e447
```

I assume that when https://github.com/cilium/ebpf/pull/1990 is merged, we will update the `cilium/ebpf` dependency and mark this PR as ready to merge.

Please let me know if that's not the case.

pkg/bpf/detect_linux.go
```diff
@@ -630,6 +630,11 @@ func detectKfunc(name string) bool {
 }
 
 func HasKfunc(name string) bool {
+
+	if btfFilePath := btf.GetCachedBTFFile(); btfFilePath != "/sys/kernel/btf/vmlinux" {
```
olsajiri, line 638: please put some comment on why we do that in here, thanks

olsajiri (approved):
lgtm, left one comment, thanks

pkg/bpf/detect_linux.go
andrewstrohman: Thanks, good call. I did this.

andrewstrohman:
> Thanks @andrewstrohman!
> 
> I've moved this to draft, since I don't think we want to merge it with:
> 
> ```
> > # github.com/cilium/ebpf v0.21.0 => github.com/andrewstrohman/ebpf v0.0.0-20260422233615-d6968610e447
> ```
> 
> I assume that when [cilium/ebpf#1990](https://github.com/cilium/ebpf/pull/1990) is merged, we will update the `cilium/ebpf` dependency and mark this PR as ready to merge.
> 
> Please let me know if that's not the case.

I did that for two reasons:
1) I wasn't sure how long it would take to merge.
2) I wanted to only add the commit on top of the release we were using, without bringing in less-tested, mainline changes that have not been incorporated into a official cilium/ebpf release yet.


For 1, the equivalent fix was merged this morning. For 2, maybe it's not that important.

Now that the fix is merged, I updated the version of the cilium/ebpf module to point at the commit of the merged fix in mainline. Please let me know if reason 2 is something that we should care about.

olsajiri:
> Now that the fix is merged, I updated the version of the cilium/ebpf module to point at the commit of the merged fix in mainline. Please let me know if reason 2 is something that we should care about.

I think we did that before and it was ok, I'm good either way.. not sure how long you can wait with this fix before there's new ebpf/cilium release

kkourt (approved):
> I wanted to only add the commit on top of the release we were using, without bringing in less-tested, mainline changes that have not been incorporated into a official cilium/ebpf release yet.

Since it fixes an actual issue, I think it's OK to use an version that is part of the `main` branch, but not yet released.

kkourt:
@andrewstrohman IMO, we should merge this as is. Any objections?

andrewstrohman:
> @andrewstrohman IMO, we should merge this as is. Any objections?

No objections. Thanks!

PR #4896, opened 2026-04-23
Warn on HostSelector != nil in TracingPolicyNamespaced

tpapagian:


FedeDP (approved):

pkg/sensors/k8s.go
```diff
@@ -22,6 +23,12 @@ import (
 //	policyfilter.PolicyID(tpID), nil if filtering is needed and policyfilter has been successfully set up
 //	_, err if an error occurred
 func (h *handler) updatePolicyFilter(tp tracingpolicy.TracingPolicy, tpID uint64) (policyfilter.PolicyID, error) {
+	hostSelector := tp.TpSpec().HostSelector
+	if tp.TpNamespace() != "" && tp.TpSpec().HostSelector != nil {
+		logger.GetLogger().Warn("TracingPolicyNamespaced cannot match host workloads. Explicitly set `spec.hostSelector: null` to remove this warning. Overrided spec.hostSelector with null before applying this policy.")
+		hostSelector = nil
+	}
```
kkourt, line 35: What's the default? Will this warning be emitted for namespaced policies taht do not define a `HostSelector`?

tpapagian: The default is `{}` as defined a previous PR. The user will get this warning, but the policy will load and the `hostSelector` will be removed in `TracingPolicyNamespaced`.

To remove that warning users need to define `spec.hostSelector: null` explicitly.

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
```diff
@@ -55,6 +55,8 @@ To filter host workloads we use the `hostSelector` field of tracing policies to
 should be applied to host workloads or not. For now this only supports `{}` to match all host workloads
 and `null` to match none of the host workloads.
 
+If `hostSelector` is not `null` in `TracingPolicyNamespaced` users will get a warning and `hostSelector` will be overridden to `null`.
```
mtardy:
```suggestion
{{< note >}}
Since a `TracingPolicyNamespaced` can only apply to Pods within a specific
Kubernetes namespaces, any non-null `hostSelector` will trigger a warning and
and be overriden to `null`.
{{< /note >}}
```

pkg/sensors/k8s.go
```diff
@@ -22,6 +23,12 @@ import (
 //	policyfilter.PolicyID(tpID), nil if filtering is needed and policyfilter has been successfully set up
 //	_, err if an error occurred
 func (h *handler) updatePolicyFilter(tp tracingpolicy.TracingPolicy, tpID uint64) (policyfilter.PolicyID, error) {
+	hostSelector := tp.TpSpec().HostSelector
+	if tp.TpNamespace() != "" && tp.TpSpec().HostSelector != nil {
```
mtardy, line 30: Idea: couldn't we make this impossible by design having different kubebuilder validation on the namespaced version? Or adding X-Validation?

mtardy (commented):
I know the why might be obvious for someone with context but would be worth adding (to the commit descr as well as the docs)

pkg/sensors/k8s.go
kkourt: OK, can we then have the warning show only a single time in the agent's lifetime?

tpapagian: Sure, I will do that.

tpapagian: Done

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
tpapagian: Done

pkg/sensors/k8s.go
tpapagian: I have done several attends to fix that in a cleaner way. 

Possibly a better way is to have different defaults for `TracingPolicy` and `TracingPolicyNamespaced` for the `spec.hostSelector `. I can do that with having separate `TracingPolicySpec` for each of these policies. For the global selectors this will be not a big deal regarding code duplication. But I am thinking also for https://github.com/cilium/tetragon/pull/4833 that will have the same issues. In that case we need to create a lot of duplicate structs which does not sound ideal to me.

Regarding X-Validation, I tried something like:
```diff
diff --git a/pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go b/pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
index 6f2376ab4..e189786bd 100644
--- a/pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
+++ b/pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
@@ -50,6 +50,7 @@ type TracingPolicy struct {
 // +genclient:noStatus
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 // +kubebuilder:resource:categories={tetragon},singular="tracingpolicynamespaced",path="tracingpoliciesnamespaced",scope="Namespaced",shortName={tgtpn}
+// +kubebuilder:validation:XValidation:rule="spec.hostSelector == null",message="TracingPolicyNamespaced should have null spec.hostSelector."
 type TracingPolicyNamespaced struct {
        metav1.TypeMeta   `json:",inline"`
        metav1.ObjectMeta `json:"metadata"`
```
But this also does not work as expected. Possibly related to https://github.com/cilium/tetragon/pull/4889#pullrequestreview-4161392836 but I haven't spend more time on that yet to be honest. 

I just created this RP to make sure that things work as expected and I will continue with a cleaner solution on a followup PR if this makes sense.

mtardy (approved):

pkg/sensors/k8s.go
mtardy: oh okay cool :) all right indeed that looks easier like this!

pkg/sensors/k8s.go
```diff
@@ -22,6 +26,14 @@ import (
...
+		warnNonNilHostSelector.Do(func() {
+			logger.GetLogger().Warn("TracingPolicyNamespaced cannot match host workloads. Explicitly set `spec.hostSelector: null` to remove this warning. Overrided spec.hostSelector with null before applying this policy.")
```
kkourt, line 32: How about:
> TracingPolicyNamespaced cannot match host workloads. All namespaced policies will ignore this field. Explicitly set `spec.hostSelector: null` to remove this warning.

tpapagian: https://github.com/cilium/tetragon/pull/4918

PR #4900, opened 2026-04-23
tetragon: fixes for the btf resolve

olsajiri:
Fixes for the btf resolve code, which broke with the latest bpf-next/master update https://github.com/cilium/tetragon/pull/4897

### DIFF
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -320,11 +320,8 @@ func processMembers(
 	i int,
 ) (*btf.Type, error) {
 	var lastError *resolveError
-	memberWasFound := false
 	for _, member := range members {
-		if len(member.Name) == 0 { // If anonymous struct, fallthrough
-			btfArgs[i].Offset = member.Offset.Bytes()
-			btfArgs[i].IsInitialized = uint16(1)
+		if len(member.Name) == 0 { // anonymous struct/union, fallthrough
 			lastTy, err := ResolveBTFPath(btfArgs, member.Type, pathToFound, i)
 			if err != nil {
 				// Propagate the deepest error for both resolve and non-resolve error.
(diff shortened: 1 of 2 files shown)


tdaudi:
Awesome, thanks a lot for this PR ! Should we merge https://github.com/cilium/tetragon/pull/4800 before this one ?

olsajiri:
> Awesome, thanks a lot for this PR ! Should we merge #4800 before this one ?

I think so, looks like it could go in already

andrewstrohman:
I think that this is the commit that triggered the bug: https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=8228e42b5f88aa68708ced277399ee3b59748627

andrewstrohman:
In the first commit message:
> [1] 4231aba000f5 ("powerpc/64s: Fix page table fragment refcount race vs speculative references")

I don't understand the relevance of this.

andrewstrohman (approved):
LGMT. Thanks!

olsajiri:
> In the first commit message:
> 
>     [1] 4231aba000f5 ("powerpc/64s: Fix page table fragment refcount race vs speculative references")
> 
> I don't understand the relevance of this.
> 

heh, me neither ;-) removed thanks

pkg/btf/btf.go
```diff
@@ -301,64 +303,67 @@ func ResolveBTFPath(
...
 	i int,
 ) (*btf.Type, error) {
-	var lastError error
-	memberWasFound := false
+	var lastError *resolveError
 	for _, member := range members {
-		if len(member.Name) == 0 { // If anonymous struct, fallthrough
+		if len(member.Name) == 0 { // anonymous struct/union, fallthrough
 			btfArgs[i].Offset = member.Offset.Bytes()
 			btfArgs[i].IsInitialized = uint16(1)
```
tdaudi:
```suggestion
```
I'd like to take the opportunity of your PR to also remove this useless check, if you're ok.

I think it's safe if we remove it because we will go through `ResolveBTFPath` again and fall into the `if member.Name == pathToFound[i]` in any case if no error is returned.

tdaudi (approved):
I think indeed this is much better, thank you very much 💯

I just have a small suggestion but it's not directly related to the fix.

pkg/btf/btf.go
olsajiri: yep, seems good, will check, thanks

PR #4915, opened 2026-04-27
tetragon: add sleepable uprobe sensor

olsajiri:
adding new uprobe sleepable sensors in following objects:
```
      bpf_generic_uprobe_sleepable.o
      bpf_generic_retuprobe_sleepable.o
      bpf_multi_uprobe_sleepable.o
      bpf_multi_retuprobe_sleepable.o
```

now we can retrieve user space data directly from sensor and do not
need preload probes (extra 2 probes on same target)

it's available from kernel v6.19 and the policy can't have stacktrace atm

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -289,10 +289,14 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 
 		probe_read(&file, sizeof(file), &arg);
 		probe_read(&arg, sizeof(arg), &file->name);
-	}
-		fallthrough;
-	case string_type:
 		size = copy_strings(args, (char *)arg, MAX_STRING);
+		break;
+	}
+	case string_type:
+		if (argm & ARGM_USER)
+			size = copy_strings_user(args, (char *)arg, MAX_STRING);
+		else
+			size = copy_strings(args, (char *)arg, MAX_STRING);
 		break;
 	case net_dev_ty: {
 		struct net_device *dev = (struct net_device *)arg;
@@ -575,6 +579,14 @@ FUNC_INLINE long get_preload_arg(struct pt_regs *ctx, long ty, arg_status_t *sta
 }
 #endif
 
+__maybe_unused static bool get_can_sleep(void)
+{
+#ifdef __SLEEPABLE
+	return true;
+#endif
+	return false;
+}
+
 FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map_def *tailcals,
 				  int process)
 {
@@ -633,6 +645,12 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 	if (am & ARGM_PRELOAD) {
 		a = get_preload_arg(ctx, ty, &e->arg_status[index & MAX_POSSIBLE_ARGS_MASK]);
 	} else {
+		/*
+		 * can_sleep is true if the sensor can sleep and reliably read data
+		 * from user space.
+		 */
+		bool can_sleep = get_can_sleep();
+
 		asm volatile("%[index] &= %1 ;\n"
 			     : [index] "+r"(index)
 			     : "i"(MAX_POSSIBLE_ARGS_MASK));
@@ -640,11 +658,16 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 			a = get_pt_regs_arg(ctx, config, index);
 		} else if (am & ARGM_CURRENT_TASK) {
 			a = get_current_task();
+			/*
+			 * We are getting data from kernel current object, we don't
+			 * need to sleep and need to use bpf_probe_read.
+			 */
+			can_sleep = false;
 		} else {
 			arg_index = config->idx[index & MAX_POSSIBLE_ARGS_MASK];
 			a = (&e->a0)[arg_index & MAX_ACCESSIBLE_ARGS_MASK];
 		}
-		extract_arg(config, index, &a, false, &e->arg_status[index & MAX_POSSIBLE_ARGS_MASK]);
+		extract_arg(config, index, &a, can_sleep, &e->arg_status[index & MAX_POSSIBLE_ARGS_MASK]);
 	}
 
 	if (should_offload_path(ty))
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -974,8 +974,9 @@ do_action(void *ctx, __u32 i, struct selector_action *actions, bool *post, bool 
 		if (rate_limit(ratelimit_interval, ratelimit_scope, e))
 			*post = false;
 #endif /* __LARGE_BPF_PROG */
-		__u32 kernel_stack_trace = actions->act[++i];
+		__u32 kernel_stack_trace __maybe_unused = actions->act[++i];
 
+#ifndef __NO_STACKTRACE
 		if (kernel_stack_trace) {
 			// Stack id 0 is valid so we need a flag.
 			e->common.flags |= MSG_COMMON_FLAG_KERNEL_STACKTRACE;
@@ -987,13 +988,16 @@ do_action(void *ctx, __u32 i, struct selector_action *actions, bool *post, bool 
 			// Here we just signal that there was a collision returning -EEXIST.
 			e->kernel_stack_id = get_stackid(ctx, &stack_trace_map, 0);
 		}
+#endif /* __NO_STACKTRACE */
 
-		__u32 user_stack_trace = actions->act[++i];
+		__u32 user_stack_trace __maybe_unused = actions->act[++i];
 
+#ifndef __NO_STACKTRACE
 		if (user_stack_trace) {
 			e->common.flags |= MSG_COMMON_FLAG_USER_STACKTRACE;
 			e->user_stack_id = get_stackid(ctx, &stack_trace_map, BPF_F_USER_STACK);
 		}
+#endif /* __NO_STACKTRACE */
 #ifdef __V511_BPF_PROG
 		__u32 ima_hash = actions->act[++i];
 
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -543,7 +543,7 @@ FUNC_INLINE long get_pt_regs_arg(struct pt_regs *ctx, struct event_config *confi
 }
 #endif /* __TARGET_ARCH_x86 && (GENERIC_KPROBE || GENERIC_UPROBE) */
 
-#if defined(GENERIC_UPROBE) || defined(GENERIC_USDT)
+#if !defined(__NO_PRELOAD) && (defined(GENERIC_UPROBE) || defined(GENERIC_USDT))
 FUNC_INLINE unsigned long get_preload_arg(struct pt_regs *ctx, long ty, arg_status_t *status)
 {
 	unsigned long arg = 0;
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -379,6 +379,28 @@ FUNC_INLINE long copy_strings(char *args, char *arg, int max_size)
 	return size + 4;
 }
 
+#ifdef __SLEEPABLE
+FUNC_INLINE long copy_strings_user(char *args, char *arg, int max_size)
+{
+	int *s = (int *)args;
+	long size;
+
+	size = bpf_copy_from_user_str(&args[4], max_size + 1, (const void *)arg, 0);
+	if (size <= 0)
+		return invalid_ty;
+	// Remove the nul character from end.
+	size--;
+	*s = size;
+	// Initial 4 bytes hold string length
+	return size + 4;
+}
+#else
+FUNC_INLINE long copy_strings_user(char *args, char *arg, int max_size)
+{
+	return 0;
+}
+#endif
+
 FUNC_INLINE long copy_skb(char *args, unsigned long arg)
 {
 	struct sk_buff *skb = (struct sk_buff *)arg;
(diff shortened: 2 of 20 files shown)


pkg/sensors/tracing/genericuprobe.go
```diff
@@ -422,6 +422,19 @@ type uprobeHas struct {
 	sleepableOffload bool
 	sleepablePreload bool
 	substring        bool
+	sleepable        bool
+}
+
+func uprobeHasSetup(uprobes []v1alpha1.UProbeSpec, opts *specOptions) uprobeHas {
+	hasStackTrace := bool(false)
```
mtardy: I was confused on why the explicit type conversion here
```suggestion
	hasStackTrace := false
```

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -877,6 +903,19 @@ func multiUprobePinPath(sensorPath string) string {
...
+		if sleepable {
+			return "uprobe.multi.s/generic_uprobe", "uprobe.multi.s/generic_retuprobe"
+		}
+		return "uprobe.multi/generic_uprobe", "uprobe.multi/generic_retuprobe"
+	}
+	if sleepable {
+		return "uprobe.s/generic_uprobe", "uprobe.s/generic_retuprobe"
+	}
+	return "uprobe/generic_uprobe", "uprobe/generic_retuprobe"
+}
```
mtardy, line 917: how about
```suggestion
func uprobeLabel(multi, sleepable bool) (string, string) {
	prefix := "uprobe"

	if multi {
		prefix += ".multi"
	}

	if sleepable {
		prefix += ".s"
	}

	return prefix + "/generic_uprobe",
		prefix + "/generic_retuprobe"
}
```

bpf/process/generic_calls.h
```diff
@@ -633,18 +645,21 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
...
+		bool can_sleep = get_can_sleep();
+
 		asm volatile("%[index] &= %1 ;\n"
 			     : [index] "+r"(index)
 			     : "i"(MAX_POSSIBLE_ARGS_MASK));
 		if (am & ARGM_PT_REGS) {
 			a = get_pt_regs_arg(ctx, config, index);
 		} else if (am & ARGM_CURRENT_TASK) {
 			a = get_current_task();
+			can_sleep = false;
```
mtardy, line 665: could you comment this one? now sure to follow, maybe I miss context on this

docs/content/en/docs/concepts/tracing-policy/options.md
```diff
@@ -44,6 +44,7 @@ Example:
 ## Uprobe options
 
 - [`disable-uprobe-multi`](#disable-uprobe-multi): disable uprobe multi link
+- [`disable-sleepable`](#disable-sleepable): disable sleepable uprobe programs
```
mtardy, line 47: I was wondering about this if this should be an agent flag or a per policy option, but indeed this feels more like a per-policy thing, why not...

mtardy (commented):
Thanks look good! Here are a few comments/questions.

pkg/sensors/tracing/genericuprobe.go
olsajiri: I'd rather keep it this way, it's simple and follows `Generic*Objs` functions

pkg/sensors/tracing/genericuprobe.go
olsajiri: yep, will fix

bpf/process/generic_calls.h
olsajiri: it's for extract_arg_depth to use either `copy_from_user` (in sensors that can sleep  can_sleep=true) or just `probe_read`, I'll add comment in the code

this particular  `can_sleep = false` case, we are getting data from kernel current object, so we need to use the normal probe_read

docs/content/en/docs/concepts/tracing-policy/options.md
olsajiri: yep, I think we discussed that for other flags and it's easier in the policy than passing the arguments

mtardy (approved):

ExceptionalHandler (approved):

3u13r:
quick question, this is available from 7.0 and not 6.19 onwards, right? At least I cannot find it [in 6.19.14](https://elixir.bootlin.com/linux/v6.19.14/source/kernel/bpf/core.c#L2427-L2428).

pkg/bpf/detect_linux.go
```diff
@@ -734,6 +734,45 @@ func HasGetFuncRetHelper() bool {
...
+			asm.Return(),
+		},
+		Type:    ebpf.Kprobe,
+		Flags:   unix.BPF_F_SLEEPABLE,
+		License: "Dual BSD/GPL",
+	})
+	if err != nil {
+		return false
+	}
+	defer prog.Close()
```
andrewstrohman, line 765: I don't think we need to defer here. We can just call `prog.Close()` without defer'ng, right? In most other detect functions, we use defer because prog is used for a subsequent link call. But we aren't doing that here.

I do see that detectMixBpfAndTailCalls does this seemingly unnecessary defer as well.

andrewstrohman:
I think that this will be a problematic in combination with our heap:
```
struct {
    __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
    __uint(max_entries, 1);
    __type(key, __u32);
    __type(value, struct msg_generic_kprobe);
} process_call_heap SEC(".maps");
```

BPF_MAP_TYPE_PERCPU_ARRAY works as a heap under the assumption that the bpf program will not be interrupted or preempted. But with this change, the bpf program can sleep when encountering a fault. Then, another instance of the program on the same CPU can write to the heap while the initial instance is sleeping, causing corruption of the data on the heap.

olsajiri:
> I think that this will be a problematic in combination with our heap:
> 
> ```
> struct {
>     __uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
>     __uint(max_entries, 1);
>     __type(key, __u32);
>     __type(value, struct msg_generic_kprobe);
> } process_call_heap SEC(".maps");
> ```
> 
> BPF_MAP_TYPE_PERCPU_ARRAY works as a heap under the assumption that the bpf program will not be interrupted or preempted. But with this change, the bpf program can sleep when encountering a fault. Then, another instance of the program on the same CPU can write to the heap while the initial instance is sleeping, causing corruption of the data on the heap.

you're right.. I focused on the migrate_disable which keeps the program on same cpu but forgot about the preemption.. I think we could use task local storage in here, but seems its size is limited to 2 pages and we need like 20K or so.. need to re-think this.. thanks

bpf/process/types/basic.h
```diff
@@ -379,6 +379,28 @@ FUNC_INLINE long copy_strings(char *args, char *arg, int max_size)
...
 
+#ifdef __SLEEPABLE
+FUNC_INLINE long copy_strings_user(char *args, char *arg, int max_size)
+{
+	int *s = (int *)args;
+	long size;
+
+	size = bpf_copy_from_user_str(&args[4], max_size + 1, (const void *)arg, 0);
+	if (size <= 0)
+		return invalid_ty;
```
andrewstrohman, line 390: I recently changed `copy_strings` to return size, instead of invalid_ty. That was the last user of invalid_ty until this resurrected its usage.

Currently, all that matters in terms of arg error status reporting, is that we return a negative number, and returning invalid_ty satisfies that. The reason is, that I've made it so that all negative return values bubble up as the same argument status value (-1).

However, folks have suggested that we relay the error code in the status so we can be confident what error occurred. So, we might want to consider returning size here in case we decide to relay error codes in the future.

andrewstrohman: I got thrown off by incorrect documentation for `bpf_copy_from_user_str`. I created a [PR](https://github.com/isovalent/ebpf-docs/pull/274) to update the docs.

andrewstrohman: The reason I cared about the return value, is that I don't think that `bpf_copy_from_user_str` can return zero (as long as the second parameter is non-zero). So, we don't need to worry about returning 0 here if we choose to return size instead of invalid_ty.

bpf/process/types/basic.h
```diff
@@ -379,6 +379,28 @@ FUNC_INLINE long copy_strings(char *args, char *arg, int max_size)
 	return size + 4;
 }
 
+#ifdef __SLEEPABLE
+FUNC_INLINE long copy_strings_user(char *args, char *arg, int max_size)
+{
+	int *s = (int *)args;
+	long size;
+
+	size = bpf_copy_from_user_str(&args[4], max_size + 1, (const void *)arg, 0);
```
andrewstrohman, line 388: Everywhere else in the code, we guard calling of kfuncs with `bpf_ksym_exists` (or similar) before calling the func. This is needed because if the kfunc is not available, then the loader will poison the kfunc call. The `bpf_ksym_exists` guard prevents the verifier from seeing the poisoned call (due to dead code elimination), and rejecting the program.

`detectSleepableTailCalls` does not depend on kfunc availability (as it shouldn't), so we can end up trying to load the program with the poisoned kfunc call.

I simulated not having a `/sys/kernel/btf` with:
```
diff --git a/vendor/github.com/cilium/ebpf/btf/kernel.go b/vendor/github.com/cilium/ebpf/btf/kernel.go
index 2089d64c8..4ad9feb08 100644
--- a/vendor/github.com/cilium/ebpf/btf/kernel.go
+++ b/vendor/github.com/cilium/ebpf/btf/kernel.go
@@ -128,7 +128,7 @@ func loadKernelSpec() (*Spec, error) {
                return nil, internal.ErrNotSupportedOnOS
        }

-       fh, err := os.Open("/sys/kernel/btf/vmlinux")
+       fh, err := os.Open("/sys/kernel/btf/vmlinux-does-not-exist")
        if err == nil {
                defer fh.Close()

@@ -179,7 +179,7 @@ func loadKernelModuleSpec(module string, base *Spec) (*Spec, error) {
                return nil, fmt.Errorf("invalid module name %q", module)
        }

-       fh, err := os.Open(filepath.Join("/sys/kernel/btf", module))
+       fh, err := os.Open(filepath.Join("/sys/kernel/btf-does-not-exist", module))
        if err != nil {
                return nil, err
        }
@@ -201,15 +201,7 @@ func findVMLinux() (*os.File, error) {

        // use same list of locations as libbpf
        // https://github.com/libbpf/libbpf/blob/9a3a42608dbe3731256a5682a125ac1e23bced8f/src/btf.c#L3114-L3122
-       locations := []string{
-               "/boot/vmlinux-%s",
-               "/lib/modules/%s/vmlinux-%[1]s",
-               "/lib/modules/%s/build/vmlinux",
-               "/usr/lib/modules/%s/kernel/vmlinux",
-               "/usr/lib/debug/boot/vmlinux-%s",
-               "/usr/lib/debug/boot/vmlinux-%s.debug",
-               "/usr/lib/debug/lib/modules/%s/vmlinux",
-       }
+       locations := []string{}

        for _, loc := range locations {
                file, err := os.Open(fmt.Sprintf(loc, release))

```

Then, on a bpf-next kernel, I loaded a uprobe policy that does not specify string args:

```
kind: TracingPolicy
metadata:
  name: "uprobe"
spec:
  uprobes:
  - path: "/host/uprobe"
    symbols:
    - "func_int"
    args:
    - index: 0
      type: "int32"
    selectors:
    - matchArgs:
      - args: [0]
        operator: "GreaterThan"
        values:
          - "0"
      - args: [0]
        operator: "LessThan"
        values:
          - "1025"
      matchActions:
      - action: NoPost
    - matchActions:
      - action: Override
        argError: -2
```
This resulted in:
```
level=error msg="Failed to execute tetragon" error="sensor generic_uprobe from collection uprobe failed to load: failed prog bpf/objs/bpf_multi_uprobe_sleepable.o kern_version 459008 loadInstance: opening collection 'bpf/objs/bpf_multi_uprobe_sleepable.o' failed: program generic_uprobe_process_event: load program: unknown kfunc: invalid func unknown#233685214 (229 line(s) omitted)"
```
When I changed the code like this:
```
index db8e69f73..56a822813 100644
@@ -293,9 +293,10 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
                break;
        }
        case string_type:
-               if (argm & ARGM_USER)
-                       size = copy_strings_user(args, (char *)arg, MAX_STRING);
-               else
+               if (argm & ARGM_USER) {
+                       if (bpf_ksym_exists(bpf_copy_from_user_str))
+                               size = copy_strings_user(args, (char *)arg, MAX_STRING);
+               } else
                        size = copy_strings(args, (char *)arg, MAX_STRING);
                break;
        case net_dev_ty: {
```
We can load the program. Since we check for `bpf_copy_from_user_str` if there is a string arg specified in the policy (and reject policy if not available), I think that doing something like above would be OK, because that code path will not be taken without a string arg configured in the policy.

andrewstrohman: I probably need to create a unit test to catch problems like this.

olsajiri: souds good.. I ended up with kernel change I need to be pushed first, so will resume this after that, thanks

olsajiri:
> quick question, this is available from 7.0 and not 6.19 onwards, right? At least I cannot find it [in 6.19.14](https://elixir.bootlin.com/linux/v6.19.14/source/kernel/bpf/core.c#L2427-L2428).

this one is in v6.19:
`0f0c332992b8 bpf: Allow sleepable programs to use tail calls
`
but we need another fix, will update the PR later

docs/content/en/docs/concepts/tracing-policy/options.md
```diff
@@ -60,3 +61,19 @@ Example:
     - name: "disable-uprobe-multi"
       value: "1"
 ```
+
+### disable-sleepable
+
+This option disables sleepable uprobe programs for all the uprobes defined in
```
kkourt, line 67: Should we maybe name the option disable-uprobe-sleepable?

PR #4917, opened 2026-04-27
Simplify workload selectors

tpapagian:
https://github.com/cilium/tetragon/pull/4814 added support for spec.hostSelector. The idea was to have by default `spec.podSelector: {}`, `spec.containerSelector: {}`, and `spec.hostSelector: {}`. But this has some unintended side effects that complicate our codebase (i.e. https://github.com/cilium/tetragon/pull/4889 and https://github.com/cilium/tetragon/pull/4896). 

In order to fix that, we change the defaults of `spec.{pod, container, host}Selector` from `{}` to `null`. The new behaviour is described in the table below:

| `hostSelector`   | `podSelector`    | `containerSelector` | Result                                                                                                                 |
| ---------------- | ---------------- | ------------------- | ---------------------------------------------------------------------------------------------------------------------- |
| `null` (default) | `null` (default) | `null` (default)    | Select all host, pod, and container workloads.                                                                         |
| `null` (default) | `null` (default) | `{...}`             | Select all containers that match `containerSelector` across all pods. No host workloads are selected.                  |
| `null` (default) | `{...}`          | `null` (default)    | Select all containers in pods that match `podSelector`. No host workloads are selected.                                |
| `null` (default) | `{...}`          | `{...}`             | Select all containers that match `containerSelector` in pods that match `podSelector`. No host workloads are selected. |
| `{}`             | `null` (default) | `null` (default)    | Select all host workloads.                                                                                             |
| `{}`             | `null` (default) | `{...}`             | Select all host workloads, plus all containers that match `containerSelector` across all pods.                         |
| `{}`             | `{...}`          | `null` (default)    | Select all host workloads, plus all containers in pods that match `podSelector`.                                       |
| `{}`             | `{...}`          | `{...}`             | Select all host workloads, plus all containers that match `containerSelector` in pods that match `podSelector`.        |

This change will also allow us to remove the code introduced in https://github.com/cilium/tetragon/pull/4889 and be able to check the `hostSelector` value inside `TracingPolicyNamespaced` with a `kubebuilder` `XValidation` (i.e. https://github.com/cilium/tetragon/pull/4896#discussion_r3147258823).


pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpoliciesnamespaced.yaml
```diff
@@ -6321,5 +6315,8 @@ spec:
         - metadata
         - spec
         type: object
+        x-kubernetes-validations:
+        - message: TracingPolicyNamespaced should have a null spec.hostSelector.
+          rule: '!has(self.spec.hostSelector)'
```
mtardy, line 6320: have you tested that it works? does it work on k8s and in standalone?

mtardy (commented):
This feel reasonable and would simplify things, we discussed offline maybe merging the table because it's not obvious to me when I read this. Just another comment:

pkg/k8s/apis/cilium.io/client/crds/v1alpha1/cilium.io_tracingpoliciesnamespaced.yaml
tpapagian: Yes in k8s:

```bash
$ cat example1.yaml
kind: TracingPolicyNamespaced
metadata:
  name: "example"
spec:
  kprobes:
  - call: "security_file_permission"
    syscall: false
$ kubectl apply -n default -f example1.yaml
tracingpolicynamespaced.cilium.io/example created
$ kubectl delete -n default -f example1.yaml
tracingpolicynamespaced.cilium.io "example" deleted from default namespace
$ cat example2.yaml
kind: TracingPolicyNamespaced
metadata:
  name: "example"
spec:
  hostSelector: {}
  kprobes:
  - call: "security_file_permission"
    syscall: false
$ kubectl apply -n default -f example2.yaml
The TracingPolicyNamespaced "example" is invalid: <nil>: Invalid value: "object": TracingPolicyNamespaced should have a null spec.hostSelector.
```

Not sure if it make sense to have `TracingPolicyNamespaced` in standalone. i.e.
```bash
$ sudo ./tetragon --btf /sys/kernel/btf/vmlinux  --bpf-lib ./bpf/objs/ --tracing-policy ./example1.yaml
[...]
level=error msg="Failed to execute tetragon" error="validation failed: metadata.namespace: Required value"
```
What use-case do you have in mind?

tpapagian: Something like that in standalone?

```bash
$ cat example1.yaml
kind: TracingPolicyNamespaced
metadata:
  name: "example"
  namespace: "default"
spec:
  kprobes:
  - call: "security_file_permission"
    syscall: false
$ sudo ./tetragon --btf /sys/kernel/btf/vmlinux  --bpf-lib ./bpf/objs/ --tracing-policy ./example1.yaml --enable-policy-filter
[...]
level=info msg="Events queue size (events)" size=63K
```

and 

```bash
$ cat example2.yaml
kind: TracingPolicyNamespaced
metadata:
  name: "example"
  namespace: "default"
spec:
  hostSelector: {}
  kprobes:
  - call: "security_file_permission"
    syscall: false
$ sudo ./tetragon --btf /sys/kernel/btf/vmlinux  --bpf-lib ./bpf/objs/ --tracing-policy ./example2.yaml --enable-policy-filter
[...]
level=error msg="Failed to execute tetragon" error="TracingPolicyNamespaced cannot match host workloads so spec.hostSelector should be null"
```

mtardy: Amazing, I had the fear at some point that X-Validation was not working on standalone! thanks a lot!

mtardy (approved):
I think that's indeed better.

So as I understand you do not need to revert 564b89ad373fcda7adb53ccabad8c414e1aa4bd3 as well?

Maybe we should highlight that the default `null`/`null`/`null` is a special case with a note or caution?

Should you also not merge https://github.com/cilium/tetragon/pull/4918 and revert the PR now since you shouldn't be able to get any policy not respecting the validation?

tpapagian:
> So as I understand you do not need to revert https://github.com/cilium/tetragon/commit/564b89ad373fcda7adb53ccabad8c414e1aa4bd3 as well?

Possibly part of that. I will check now. Not everything as this also contains fixes/changes for the policyfilter implementation to support the hostSelector (i.e. an entry in the policy_filter_map with all pod cgroup IDs).

> Maybe we should highlight that the default null/null/null is a special case with a note or caution?

Yes, I can do that.

> Should you also not merge https://github.com/cilium/tetragon/pull/4918 and revert the PR now since you shouldn't be able to get any policy not respecting the validation?

I have closed https://github.com/cilium/tetragon/pull/4918 as it does not make sense to merge it now. I have kept the check in the agent but now it is an error. Just in case somehow we overcome the XValidation check.

tpapagian:
> > So as I understand you do not need to revert [564b89a](https://github.com/cilium/tetragon/commit/564b89ad373fcda7adb53ccabad8c414e1aa4bd3) as well?
> 
> Possibly part of that. I will check now. Not everything as this also contains fixes/changes for the policyfilter implementation to support the hostSelector (i.e. an entry in the policy_filter_map with all pod cgroup IDs).
> 

Most of that was already done in the first patch. I have added a comment in the first patch description.

> > Maybe we should highlight that the default null/null/null is a special case with a note or caution?
> 
> Yes, I can do that.
> 

Done.

kkourt (approved):

PR #4922, opened 2026-04-28
helm/tetragon: expose tetragon.sock to host

3u13r:
### Description
See commit message for context.

Since we've not yet released a version with the broken commit, I don't think a release message is needed.

What is still needed is adding CI around the rthooks. But I wanted to get the fix in before the next release and adding CI is more work (that I'm starting now).


kkourt (approved):
Thanks!

PR #4925, opened 2026-04-28
nok8s: don't return an error for label filters

kkourt:
The current code results in:
> Error: failed to receive events: rpc error: code = Unknown desc = label filters not supported in nok8s build

When doing "tetra getevents" for the nok8s build.

Instead of return an error, just ignore the unsupported filters.

Fixes: d833af47155e ("nok8s: update pkg/filters")


sayboras (approved):
Looks legit

PR #4926, opened 2026-04-28
add flag to disable event metrics

dangome3:
## Description

Add --enable-event-metrics flag (default: true) to allow disabling event metrics while keeping health and resource metrics available.

Split InitAllMetrics into InitHealthMetrics and InitEventsMetrics. Health metrics and resource metrics are always initialized when --metrics-server is set.

Unit tests for metrics config added in 'pkg/metrics/metricwithpod_test.go' also manually tested in a no-k8s environment.

### Changelog

```release-note
metrics: add agent flag to disable event metrics while keeping health metrics
```

### DIFF
--- a/docs/content/en/docs/installation/metrics.md
+++ b/docs/content/en/docs/installation/metrics.md
@@ -10,6 +10,12 @@ Tetragon exposes a number of Prometheus metrics that can be used for two main pu
 1. Monitoring the health of Tetragon itself
 2. Monitoring the activity of processes observed by Tetragon
 
+{{< note >}}
+  When the metrics server is started, Tetragon exposes both health and event
+  metrics by default. You can disable event metrics while keeping health metrics
+  by passing the `--enable-event-metrics=false` flag.
+{{< /note >}}
+
 For the full list, refer to [metrics reference]({{< ref "/docs/reference/metrics" >}}).
 
 ## Enable/Disable Metrics
--- a/pkg/observer/observertesthelper/observer_test_helper.go
+++ b/pkg/observer/observertesthelper/observer_test_helper.go
@@ -246,7 +246,8 @@ func getDefaultObserver(tb testing.TB, ctx context.Context, initialSensor *senso
 	// a lot of code changes in a lot of a files.
 	metricsEnableOnce.Do(func() {
 		go metricsconfig.EnableMetrics(metricsAddr)
-		metricsconfig.InitAllMetrics(metricsconfig.GetRegistry())
+		metricsconfig.InitHealthMetrics(metricsconfig.GetRegistry())
+		metricsconfig.InitEventsMetrics(metricsconfig.GetRegistry())
 	})
 
 	tb.Cleanup(func() {
--- a/pkg/option/flags.go
+++ b/pkg/option/flags.go
@@ -48,6 +48,7 @@ const (
 	KeyEnablePodAnnotations = "enable-pod-annotations"
 
 	KeyMetricsServer      = "metrics-server"
+	KeyEnableEventMetrics = "enable-event-metrics"
 	KeyMetricsLabelFilter = "metrics-label-filter"
 	KeyServerAddress      = "server-address"
 	KeyGopsAddr           = "gops-address"
@@ -234,6 +235,7 @@ func ReadAndSetFlags() error {
 	}
 
 	Config.MetricsServer = viper.GetString(KeyMetricsServer)
+	Config.EnableEventMetrics = viper.GetBool(KeyEnableEventMetrics)
 	Config.MetricsLabelFilter = DefaultLabelFilter().WithEnabledLabels(ParseMetricsLabelFilter(viper.GetString(KeyMetricsLabelFilter)))
 	Config.ServerAddress = viper.GetString(KeyServerAddress)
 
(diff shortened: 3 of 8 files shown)


kkourt (commented):
Thanks @dangome3!

Can you please split the first patch in two patches? One for the new functionality, and one for the unit tests?

dangome3:
Hey @kkourt ! sure! that's already done. Thanks for the suggestion

pkg/observer/observertesthelper/observer_test_helper.go
```diff
@@ -246,7 +246,8 @@ func getDefaultObserver(tb testing.TB, ctx context.Context, initialSensor *senso
 	// a lot of code changes in a lot of a files.
 	metricsEnableOnce.Do(func() {
 		go metricsconfig.EnableMetrics(metricsAddr)
-		metricsconfig.InitAllMetrics(metricsconfig.GetRegistry())
+		metricsconfig.InitHealthMetrics(metricsconfig.GetRegistry())
+		metricsconfig.InitEventsMetrics(metricsconfig.GetRegistry())
```
mtardy, line 250: question: we don't want to respect the flag from testing?

docs/content/en/docs/installation/metrics.md
```diff
@@ -50,6 +50,9 @@ address of the Tetragon Agent to, for example, `:2112`, via the `--metrics-serve
 If using [systemd]({{< ref "/docs/installation/package" >}}), set the `metrics-address` entry in a file under the
 `/etc/tetragon/tetragon.conf.d/` directory.
 
+By default, Tetragon exposes both health and events metrics. You can disable events metrics while keeping
+health metrics by adding `--enable-event-metrics=false`.
```
mtardy: nit: wrapping is a bit wrong here (like in the rest of that doc btw)
```suggestion
By default, Tetragon exposes both health and events metrics. You can disable
events metrics while keeping health metrics by adding `--enable-event-metrics=false`.
```

docs/content/en/docs/installation/metrics.md
mtardy, line 11: I feel like you could add a note there just after 1. and 2. instead of later in the `## Enable/Disable Metrics` and `### Non-Kubernetes` section.

mtardy (commented):
Overall looks good thanks, just a few comments. 

Also I know the patch is quite simple but the first one could have been also sliced into two, your commit description is a hint for this, your commit does two things:
```
Add --enable-event-metrics flag (default: true) to allow disabling
event metrics while keeping health and resource metrics available.

Split InitAllMetrics into InitHealthMetrics and InitEventsMetrics.
Health metrics and resource metrics are always initialized when
--metrics-server is set.
```
I would have split first, and then added the flag in two separate commits.

pkg/observer/observertesthelper/observer_test_helper.go
dangome3: This was intentional as this file contains only tests helper functions, and maybe any test could use the event metrics too

Please let me know if my assumption is wrong and if we should respect the flag here too

dangome3:
Thanks @mtardy for the suggestions. I updated the PR

kkourt:
> Also I know the patch is quite simple but the first one could have been also sliced into two, your commit description is a hint for this, your commit does two things:
> 
> ```
> Add --enable-event-metrics flag (default: true) to allow disabling
> event metrics while keeping health and resource metrics available.
> 
> Split InitAllMetrics into InitHealthMetrics and InitEventsMetrics.
> Health metrics and resource metrics are always initialized when
> --metrics-server is set.
> ```
> 
> I would have split first, and then added the flag in two separate commits.

I believe what Mahè meant here is to:
1. Split the implementation of InitAllMetrics into InitHealthMetrics and InitEventsMetrics into one commit
2. Add the flag, and use it to determine whether InitEventMetrics or not in the second commit.

In the current split, the flag is introduced in the first patch, and then used in the second patch, which I find more confusing that having both changes in a single patch (as originally).

dangome3:
> In the current split, the flag is introduced in the first patch, and then used in the second patch, which I find more confusing that having both changes in a single patch (as originally).

That's 💯 right, sorry for the confusion. That should be fixed now

pkg/option/flags.go
```diff
@@ -409,6 +411,7 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.String(KeyK8sKubeConfigPath, "", "Absolute path of the kubernetes kubeconfig file")
 	flags.Int(KeyK8sControlPlaneRetry, 1, "Number of attempts for Kubernetes control plane connection (negative for infinite, zero is invalid, positive for max attempts)")
 	flags.String(KeyMetricsServer, "", "Metrics server address (e.g. ':2112'). Disabled by default")
+	flags.Bool(KeyEnableEventMetrics, true, "Enable per-event metrics. Enabled by default. Health and resource metrics are always available when --metrics-server is set.")
```
mtardy: sorry missed it: could you just use fmt.Sprintf("... when --%s ...", keyMetricsServer) here instead of hardcoding it.

mtardy (approved):
thanks look good, just one small nit that I missed, otherwise than that now it's a very nice PR to review cut like that thanks.

pkg/option/flags.go
dangome3: Good catch! thanks for the suggestion

mtardy (approved):

PR #4933, opened 2026-04-30
tetragoninfo: add darwin stub for bpfProbes

### LINKED ISSUE #4931: v1.7.0 build failed on macOS
Hello, this is a homebrew maintainer.

It seems that the new version didn't expected to build on macOS, right?

```
pkg/tetragoninfo/info.go:84:12: undefined: bpfProbes
```

Is there any plan to support or fix for macOS?
Thanks,

chenrui333:
Add `info_darwin.go` with a nil-returning `bpfProbes()` stub, matching the existing `info_windows.go`. Without this, the tetra CLI fails to build on macOS:

```
# github.com/cilium/tetragon/pkg/tetragoninfo
pkg/tetragoninfo/info.go:84:12: undefined: bpfProbes
```

Closes #4931
Ref: https://github.com/Homebrew/homebrew-core/pull/280098

### DIFF
--- a/pkg/tetragoninfo/info_darwin.go
+++ b/pkg/tetragoninfo/info_darwin.go
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+package tetragoninfo
+
+import (
+	"github.com/cilium/tetragon/api/v1/tetragon"
+)
+
+func bpfProbes() []*tetragon.GetInfoResponse_Probe {
+	return nil
+}


kkourt (approved):

FedeDP (approved):

mtardy (approved):
The CI was supposed to catch that but it's not "failing" properly: meaning we ignore the failure see https://github.com/cilium/tetragon/issues/4934. Thanks a lot for the note and patch.

mtardy (commented):
nit: Wait could you reuse the `_windows.go` file maybe? with just a !linux tag?

EDIT: I mean it's pretty similar, nevermind we can merge like that.

chenrui333:
> nit: Wait could you reuse the `_windows.go` file maybe? with just a !linux tag?
> 
> EDIT: I mean it's pretty similar, nevermind we can merge like that.

yeah, I was thinking about that, but it would be good to separate for the long run.

PR #4937, opened 2026-04-30
tests/e2e: introduce rthooks e2e test

3u13r:
### Description
We recently regressed in the integration between the rthooks and Tetragon agent (see [here](https://github.com/cilium/tetragon/pull/4922/changes/918870344043fe5b19dbf62960596e39eddaef66)).
This PR replaces the existing rthook test, that didn't cover the integration between hook and agent, with a new e2e test. As the replaced test, this e2e test runs inside minikube.  

In an effort to make this error more debuggable, this PR also adds a log container to the rthooks daemonset that tails the logfile of the hook binary.


3u13r:
~Still a draft because I first want to make sure the tests pass and I want to do a first review myself with semi-fresh eyes.~ Done.

The new CI pipeline is [here](https://github.com/cilium/tetragon/actions/runs/25353777771/job/74338727468?pr=4937).



The rthook e2e test can only be run in minikube. Therefore, we skip it when we don't detect the `-minikube` flag. This leads to superfluous CI runs like [this](https://github.com/cilium/tetragon/actions/runs/25353777772/job/74339372640?pr=4937). 

A solution I came up with is to change the folder structure so that we have `e2e/minikube/...` and `e2e/kind/...` targets to pass via `E2E_TESTS`.
The other solution is to just switch all e2e test to use minikube.
Do you have a preference?

tests/e2e/helpers/grpcbridge/grpcbridge.go
```diff
@@ -0,0 +1,128 @@
...
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{Labels: labels},
+				Spec: corev1.PodSpec{
+					Containers: []corev1.Container{{
+						Name:  "socat",
+						Image: "docker.io/library/alpine:3.23.4@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11",
+						Command: []string{
+							"sh", "-c",
+							"apk add --no-cache -q socat && " +
+								"exec socat TCP-LISTEN:54321,reuseaddr,fork " +
```
will-isovalent: Consider using SocatPort instead of 54321 here to make sure it's in sync.

Makefile
```diff
@@ -337,11 +339,38 @@ ls-e2e-test:
 ## e2e-test E2E_TESTS=./tests/e2e/tests/skeleton: ## run a specific e2e test
 .PHONY: e2e-test
 ifneq ($(E2E_BUILD_IMAGES), 0)
-e2e-test: image image-operator
+e2e-test: image image-operator image-rthooks
```
will-isovalent: How about we make image-rthooks a requirement of e2e-test-minikube instead?

tests/e2e/helpers/cluster.go
```diff
@@ -185,3 +187,25 @@ func GetTempKindClusterName(ctx context.Context) string {
...
+		klog.InfoS("Loading image into minikube", "image", image)
+
+		cmd := exec.Command("minikube", "image", "load", image)
+		if out, err := cmd.CombinedOutput(); err != nil {
+			return ctx, fmt.Errorf("minikube image load %s: %w\n%s", image, err, out)
+		}
+
+		// cri-o on minikube automatically prefixes loaded images with "localhost/",
+		// so align the image tag with the one in the helm chart and the one when using containerd.
+		cmd = exec.Command("minikube", "image", "tag", "localhost/"+image, "docker.io/"+image)
```
will-isovalent, line 224: Is docker.io strictly correct here? A lot of our images are on quay.io

will-isovalent (commented):
Thanks @3u13r, this looks great! Just a few questions/comments.

tests/e2e/helpers/cluster.go
3u13r: That is a very good question. I'll investigate this as this PR is a few days old :smile:.
I think we are good, since this fix is needed for images that we push from the host and we only do that for locally build images. The problem with those is that they only have an implicit Repo in their name. I.e. in docker the image `cilium/tetragon:latest` can also be referred to via `docker.io/cilium/tetragon:latest`. 

Now those images that don't have a repo "get one" when pushing them into minikube.

crio: `cilium/tetragon:latest` turns into the more explicit `localhost/cilium/tetragon`
containerd: `cilium/tetragon:latest` turns into `docker.io/cilium/tetragon`

Note that under containerd you can still refer to the `docker.io/cilium/tetragon` via `cilium/tetragon`. Since the helm-chart value is overwritten in the Makefile to also `cilium/tetragon` this works well.

But under crio you cannot refer to the `localhost/cilium/tetragon` image via the value from the helm chart `cilium/tetragon`. But they implement the same default as containerd in that you can refer to a `docker.io/cilium/tetragon` image via `cilium/tetragon`. Therefore, we re-tag it as such.

For images that refer to a remote registry this re-tagging fails (as it also does when using containerd). But this doesn't matter since the image reference matches the one in helm and can just be pulled. At least as long as the registry doesn't require auth. But for one this auth case doesn't work currently so we don't break it with this change. 

I notice that this answer is very long. Let me (on Friday):
1. Try out if this minikube e2e test works with remote non-auth image references
2. Try to add more info from this answer into the comment

tests/e2e/helpers/grpcbridge/grpcbridge.go
3u13r: Good catch!

Makefile
3u13r: Good point :+1:

tests/e2e/helpers/cluster.go
3u13r: Both the default images in the deplotment, e.g. `quay.io/cilium/hubble-export-stdout` and manually setting the Tetragon agent image to a remote registry works fine.

I've added some information of the long comment above as a comment since it's not obvious what we are doing here. It also includes why remote registries work.

will-isovalent (approved):
Let's do it!

will-isovalent:
Regarding your question about minikube in CI: I think splitting it like this is a fine approach.

3u13r:
Rebased on the latest main. There was one failing test before that seemed unrelated to this PR. Let's see.

PR #4941, opened 2026-05-01
Do not update pinned digest for select go modules

andrewstrohman:
Do not update digest for cilium/little-vm-helper and ciliium/ebpf. We frequently need to pin a digest for these repos, but don't want to update again until the following release.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -375,6 +375,13 @@
       "enabled": false
     },
     {
+      // do not update digest versions for these go modules. We frequently need to pin a digest, but don't want to update again until the following release
+      "enabled": false,
+      "matchPackageNames": ["github.com/cilium/ebpf", "github.com/cilium/little-vm-helper"],
+      "matchManagers": ["gomod"],
+      "matchUpdateTypes": ["digest"],
+    },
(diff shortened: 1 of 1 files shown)


michi-covalent (approved):
👍

.github/renovate.json5
```diff
@@ -374,6 +374,20 @@
...
+      "matchManagers": ["gomod"],
+      "matchUpdateTypes": ["digest"],
+    },
+    {
+      // do not update digest versions for cilium/little-vm-helper. We frequently need to pin a digest, but don't want to update again until the following release
+      "enabled": false,
+      "matchPackageNames": ["github.com/cilium/little-vm-helper"],
+      "matchManagers": ["gomod"],
+      "matchUpdateTypes": ["digest"],
+    },
```
mtardy, line 383: Ah you can merge this package rule since `matchPackageNames` is an array so that we have only one.

andrewstrohman: OK, thanks, merged.

tpapagian (approved):

mtardy (approved):
Thank you!!

PR #4943, opened 2026-05-01
metrics: add missing HandlePerfEmptyData error label and initialization

lbcristaldo:
### Description
`HandlePerfEmptyData` was defined and used in `observer.go` but was missing from `eventHandlerErrorLabelValues`, causing `String()` to return an empty label. Add the missing label and initialize the metric in `InitMetrics()` along with the other handler error types.

Add unit tests for the new label and metric initialization.

**Testing**
- `go test ./pkg/metrics/errormetrics/...`
- `go test -race ./pkg/metrics/errormetrics/...`
- `make check`

### Changelog
```release-note
Fix missing label and initialization for HandlePerfEmptyData metric.
```

### DIFF
--- a/pkg/metrics/errormetrics/errormetrics.go
+++ b/pkg/metrics/errormetrics/errormetrics.go
@@ -51,6 +51,7 @@ const (
 
 var eventHandlerErrorLabelValues = map[EventHandlerError]string{
 	HandlePerfUnknownOp:    "unknown_opcode",
+	HandlePerfEmptyData:    "perf_empty_data",
 	HandlePerfHandlerError: "event_handler_failed",
 }
 
@@ -122,6 +123,7 @@ func InitMetrics() {
 	// NB: We initialize only ops.MSG_OP_UNDEF here, but unknown_opcode can occur for any opcode
 	// that is not explicitly handled.
 	GetHandlerErrors(ops.MSG_OP_UNDEF, HandlePerfUnknownOp).Add(0)
+	GetHandlerErrors(ops.MSG_OP_UNDEF, HandlePerfEmptyData).Add(0)
 
 	for er := range debugTypeLabelValues {
 		GetDebugTotal(er).Add(0)
--- a/pkg/metrics/errormetrics/errormetrics_test.go
+++ b/pkg/metrics/errormetrics/errormetrics_test.go
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package errormetrics
+
+import (
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/cilium/tetragon/pkg/api/ops"
+)
+
+func TestHandlePerfEmptyData(t *testing.T) {
+	// verify label exists and is correct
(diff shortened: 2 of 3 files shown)


FedeDP:
Ehy thanks for this!
Can you shorten commit message?
> Error: ERROR:CUSTOM: Please avoid long commit subjects (max: 75, found: 82)

Thanks!

FedeDP (approved):
LGTM.

pkg/metrics/errormetrics/errormetrics_test.go
```diff
@@ -0,0 +1,25 @@
...
+func TestHandlePerfEmptyData(t *testing.T) {
+	// verify label exists and is correct
+	assert.Equal(t, "perf_empty_data", HandlePerfEmptyData.String())
+
+	// use relative comparison to avoid depending on global state from InitMetrics()
+	// verify metric initializes to 0
+	before := testutil.ToFloat64(GetHandlerErrors(ops.MSG_OP_EXECVE, HandlePerfEmptyData))
+	HandlerErrorsInc(ops.MSG_OP_EXECVE, HandlePerfEmptyData)
+	after := testutil.ToFloat64(GetHandlerErrors(ops.MSG_OP_EXECVE, HandlePerfEmptyData))
+	assert.InDelta(t, before+1, after, 1e-9)
```
FedeDP, line 24: Q: why are you using `InDelta` here?

lbcristaldo: i'm using InDelta because testifylint requires it for float64 comparisons, since prometheus counters store values as floats, using assert.Equal could lead to flaky tests due to precision issues

pkg/metrics/errormetrics/errormetrics.go
```diff
@@ -117,6 +118,7 @@ func InitMetrics() {
 	for opcode := range ops.OpCodeStrings {
 		if opcode != ops.MSG_OP_UNDEF && opcode != ops.MSG_OP_TEST {
 			GetHandlerErrors(opcode, HandlePerfHandlerError).Add(0)
+			GetHandlerErrors(opcode, HandlePerfEmptyData).Add(0)
```
kkourt: In observer.go we have:

```go
	if len(data) == 0 {
		return ops.MSG_OP_UNDEF, nil, &HandlePerfError{
			kind:   errormetrics.HandlePerfEmptyData,
			err:    errors.New("empty perf data"),
			opcode: ops.MSG_OP_UNDEF,
		}
	}
```

Should't `GetHandlerErrors(opcode, HandlePerfEmptyData).Add(0)` be added for `ops.MSG_OP_UNDEF` as we do for `HandlePerfUnknownOp`?

kkourt (changes requested):

kkourt (commented):
Can you please reply to my comment in https://github.com/cilium/tetragon/pull/4943#discussion_r3186332538? 

If you addressed it, please let me know how. Thanks!

pkg/metrics/errormetrics/errormetrics.go
lbcristaldo: yes, I moved HandlePerfEmptyData initialization to MSG_OP_UNDEF 
since that's the only opcode used in observer.go. Also updated the test 
to use MSG_OP_UNDEF accordingly

kkourt (approved):

PR #4947, opened 2026-05-04
Use kubebuilder XValidation for spec.hostSelector

tpapagian:
As we already described in the documentation, spec.hostSelector only supports null and {} as values and this is checked in the agent as well.

This patch adds a kubebuilder validation as well.

### DIFF
--- a/docs/content/en/docs/reference/tracing-policy.md
+++ b/docs/content/en/docs/reference/tracing-policy.md
@@ -129,6 +129,8 @@ Currently, only the "name" field is supported.<br/>
         <td>
           HostSelector selects hosts that this policy applies to.
 For now only ~ (none) and {} (all) is supported.<br/>
+          <br/>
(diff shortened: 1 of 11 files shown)


kkourt (approved):
Thanks!

Do we need to backport this to 1.7?

mtardy (approved):

mtardy:
> Thanks!
> 
> Do we need to backport this to 1.7?

I think it's a bonus UX thing, I wouldn't bother backporting it

tpapagian:
> > Thanks!
> > Do we need to backport this to 1.7?
> 
> I think it's a bonus UX thing, I wouldn't bother backporting it

I agree, this is not a bugfix. Even without that the agent will reject the policy. It is just a matter of better UX. So technicality speaking there is not need to backport that. But it is straightforward if we want to do that.

PR #4950, opened 2026-05-05
grpc: Support mTLS for tetragon grpc server

sayboras:
### Description

This PR is to add TLS/mTLS support for tetragon grpc server, below points worth highlighting:

- grpc TLS feature is disabled by default, mainly for upgrade scenario if existing installation is already configured with TCP (instead of unix socket)
- dual listeners if TCP + TLS combination is used i.e. unix domain socket is enabled as second listener. This is to avoid any potential issue for in-pod tetra command.

### Changelog

```release-note
grpc: Support mTLS for tetragon grpc server
```

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -120,15 +120,21 @@ func absPath(p string) string {
 	return ret
 }
 
-// Save daemon information so it is used by client cli but
-// also by bugtool
+// saveInitInfo writes daemon info read by tetra and bugtool. ServerAddr
+// advertises the unix listener (when one runs) so in-pod tooling
+// defaults to the trusted IPC channel rather than the TCP listener,
+// which may require TLS / mTLS.
 func saveInitInfo() error {
+	addr := ""
+	if path, ok := resolveUnixSocketPath(option.Config.ServerAddress); ok {
+		addr = "unix://" + path
+	}
 	info := bugtool.InitInfo{
 		ExportFname: absPath(option.Config.ExportFilename),
 		LibDir:      absPath(option.Config.HubbleLib),
 		BTFFname:    absPath(option.Config.BTF),
 		MetricsAddr: option.Config.MetricsServer,
-		ServerAddr:  option.Config.ServerAddress,
+		ServerAddr:  addr,
 		GopsAddr:    option.Config.GopsAddr,
 		MapDir:      absPath(bpf.MapPrefixPath()),
 		PID:         os.Getpid(),
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -63,7 +63,10 @@ import (
 	"github.com/spf13/cobra/doc"
 	"github.com/spf13/viper"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
 	"google.golang.org/protobuf/types/known/durationpb"
+
+	"github.com/cilium/tetragon/pkg/certloader"
 )
 
 var (
@@ -688,45 +691,130 @@ func getExporter(ctx context.Context, server *server.Server) (*exporter.Exporter
 	return exporter.NewExporter(ctx, &req, server, encoder, writer, rateLimiter)
 }
 
-func Serve(ctx context.Context, listenAddr string, srv *server.Server, logSrv *eventlog.Server) error {
-	// we use an empty listen address to effectively disable the gRPC server
-	if len(listenAddr) == 0 {
+// Serve starts the Tetragon gRPC server. listenAddr drives the topology:
+// "" disables gRPC; "unix://X" runs a single plaintext listener at X;
+// on Linux, a TCP address runs that listener plus an in-pod plaintext
+// unix socket at the platform default path, with --server-tls-* gating
+// the TCP path. Windows skips the sidecar unix listener.
+//
+// extraOpts apply to every listener; callers should not pass grpc.Creds
+// here. TLS is attached to the TCP path via buildServerTLSOptions.
+func Serve(ctx context.Context, listenAddr string, srv *server.Server, logSrv *eventlog.Server, extraOpts ...grpc.ServerOption) error {
+	if listenAddr == "" {
 		return nil
 	}
-	grpcServer := grpc.NewServer()
-	tetragon.RegisterFineGuidanceSensorsServer(grpcServer, srv)
-	tetragon.RegisterEventLogServiceServer(grpcServer, logSrv)
-
+	register := func(s *grpc.Server) {
+		tetragon.RegisterFineGuidanceSensorsServer(s, srv)
+		tetragon.RegisterEventLogServiceServer(s, logSrv)
+	}
 	proto, addr, err := server.SplitListenAddr(listenAddr)
 	if err != nil {
 		return fmt.Errorf("failed to parse listen address: %w", err)
 	}
-	go func(proto, addr string) {
-		var listener net.Listener
-		var err error
-		if proto == "unix" {
-			listener, err = unixlisten.ListenWithRename(addr, 0660)
-		} else {
-			listener, err = net.Listen(proto, addr)
+
+	if proto == "unix" {
+		if err := serveOne(ctx, "unix", addr, extraOpts, register); err != nil {
+			return fmt.Errorf("starting unix gRPC listener: %w", err)
 		}
-		if err != nil {
-			logger.Fatal(log, "Failed to start gRPC server", "protocol", proto, "address", addr, logfields.Error, err)
+		return nil
+	}
+
+	if sockPath, ok := resolveUnixSocketPath(listenAddr); ok {
+		if err := serveOne(ctx, "unix", sockPath, extraOpts, register); err != nil {
+			return fmt.Errorf("starting unix gRPC listener: %w", err)
 		}
+	}
+
+	tlsOpts, tlsEnabled, err := buildServerTLSOptions(ctx)
+	if err != nil {
+		return err
+	}
+	if !tlsEnabled {
+		log.Warn("Tetragon gRPC TCP listener is exposing the API without TLS; configure --"+
+			option.KeyServerTLSCertFile+" and --"+
+			option.KeyServerTLSKeyFile+" to enable it",
+			"address", addr)
+	}
+	grpcOpts := append([]grpc.ServerOption{}, extraOpts...)
+	grpcOpts = append(grpcOpts, tlsOpts...)
+	if err := serveOne(ctx, proto, addr, grpcOpts, register); err != nil {
+		return fmt.Errorf("starting TCP gRPC listener: %w", err)
+	}
+	return nil
+}
+
+// serveOne binds the listener synchronously so bind failures surface as
+// a returned error rather than a runtime fatal.
+func serveOne(
+	ctx context.Context,
+	proto, addr string,
+	grpcOpts []grpc.ServerOption,
+	register func(*grpc.Server),
+) error {
+	var listener net.Listener
+	var err error
+	if proto == "unix" {
+		listener, err = unixlisten.ListenWithRename(addr, 0660)
+	} else {
+		listener, err = net.Listen(proto, addr)
+	}
+	if err != nil {
+		return fmt.Errorf("listen %s://%s: %w", proto, addr, err)
+	}
+	grpcServer := grpc.NewServer(grpcOpts...)
+	register(grpcServer)
+
+	go func() {
 		log.Info("Starting gRPC server", "protocol", proto, "address", addr)
-		if err = grpcServer.Serve(listener); err != nil {
-			log.Error("Failed to close gRPC server", logfields.Error, err)
+		if err := grpcServer.Serve(listener); err != nil {
+			log.Error("gRPC Serve returned", logfields.Error, err)
 		}
-	}(proto, addr)
-	go func(proto, addr string) {
+	}()
+	go func() {
 		<-ctx.Done()
 		grpcServer.Stop()
-		// if proto is unix, ListenWithRename() creates the socket
-		// then renames it, so explicitly clean it up.
 		if proto == "unix" {
 			os.Remove(addr)
 		}
-	}(proto, addr)
+	}()
 	return nil
+}
+
+// buildServerTLSOptions returns the credentials option for the TCP
+// listener (or nil when TLS is disabled). The boolean reports whether
+// TLS is actually active for logging accuracy.
+func buildServerTLSOptions(ctx context.Context) ([]grpc.ServerOption, bool, error) {
+	cfg := certloader.Config{
+		CertFile:          option.Config.ServerTLSCertFile,
+		KeyFile:           option.Config.ServerTLSKeyFile,
+		ClientCAFiles:     option.Config.ServerTLSClientCAFiles,
+		RequireClientCert: option.Config.ServerTLSRequireClientCert,
+	}
+	if !cfg.Enabled() {
+		return nil, false, nil
+	}
+	// Lazy load tolerates a provisioner (cert-manager, cilium-certgen Job)
+	// that writes the Secret after the agent starts; Watch promotes the
+	// Reloader to Ready as soon as the files appear.
+	reloader, err := certloader.NewReloaderLazy(cfg)
+	if err != nil {
+		return nil, false, fmt.Errorf("preparing gRPC TLS: %w", err)
+	}
+	certloader.Watch(ctx, reloader)
+	log.Info("gRPC TLS enabled",
+		"mtls", cfg.RequireClientCert,
+		"cert", cfg.CertFile,
+		"key", cfg.KeyFile,
+		"client-ca-files", len(cfg.ClientCAFiles),
+		"ready", reloader.Ready(),
+	)
+	if !reloader.Ready() {
+		log.Warn("gRPC TLS material not yet on disk; handshakes will fail until files appear at the configured paths",
+			"cert", cfg.CertFile,
+			"key", cfg.KeyFile,
+		)
+	}
+	return []grpc.ServerOption{grpc.Creds(credentials.NewTLS(reloader.ServerConfig()))}, true, nil
 }
 
 func startGopsServer() error {
--- a/cmd/tetragon/main_linux.go
+++ b/cmd/tetragon/main_linux.go
@@ -12,9 +12,28 @@ import (
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/reader/namespace"
 	"github.com/cilium/tetragon/pkg/reader/proc"
+	"github.com/cilium/tetragon/pkg/server"
 
 	"github.com/spf13/viper"
 )
+
+// resolveUnixSocketPath returns the unix socket path the agent should
+// expose for in-pod tooling and whether one should be started, derived
+// from listenAddr: a "unix://X" address yields X; a TCP address yields
+// defaults.DefaultUnixSocket; an empty or invalid address yields ("", false).
+func resolveUnixSocketPath(listenAddr string) (string, bool) {
+	if listenAddr == "" {
+		return "", false
+	}
+	proto, addr, err := server.SplitListenAddr(listenAddr)
+	if err != nil {
+		return "", false
+	}
+	if proto == "unix" {
+		return addr, true
+	}
+	return defaults.DefaultUnixSocket, true
+}
 
 func logCurrentSecurityContext() {
 	proc.LogCurrentSecurityContext()
--- a/docs/content/en/docs/installation/grpc-tls.md
+++ b/docs/content/en/docs/installation/grpc-tls.md
@@ -1,0 +1,135 @@
+---
+title: "Secure the gRPC API with TLS"
+linkTitle: "gRPC TLS / mTLS"
+weight: 8
+description: "Enable TLS or mTLS on the Tetragon gRPC TCP listener"
+---
+
+Tetragon serves its gRPC API on two listeners:
+
+* An always-on unix-domain socket at `/var/run/tetragon/tetragon.sock`,
+  reserved for in-pod IPC and protected by file permissions.
+* An optional TCP listener for off-host clients. TLS only applies here.
+
+This page shows how to enable TLS, switch to mTLS, and connect with `tetra`.
+
+{{< caution >}}
+TLS settings are silently ignored when `--server-address` points at a
+`unix://` socket. Configure a TCP address (for example `0.0.0.0:54321`)
+before enabling TLS.
+{{< /caution >}}
+
+## Quick start (Helm, auto-provisioned)
+
+The chart can issue and rotate certificates for you. The default
+`auto.method=helm` is the simplest option and requires no extra components.
+
+```shell
+helm upgrade --install tetragon cilium/tetragon -n kube-system \
+  --set tetragon.grpc.address=0.0.0.0:54321 \
+  --set tetragon.grpc.tls.enabled=true
+```
+
+The chart generates a CA and a server certificate with a wildcard SAN over
+`*.tetragon-grpc.cilium.io`, and writes them to a Secret mounted at
+`/var/lib/tetragon/tls/` inside each agent pod. Clients must override SNI to
+match this domain (see [Connect with `tetra`](#connect-with-tetra)).
+
+## Enable mTLS
+
+mTLS makes the agent reject TCP connections that do not present a client
+certificate signed by the bundled CA.
+
+```shell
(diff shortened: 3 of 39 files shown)


cmd/tetragon/main.go
```diff
@@ -117,15 +120,21 @@ func absPath(p string) string {
...
 
-// Save daemon information so it is used by client cli but
-// also by bugtool
+// saveInitInfo writes daemon info read by tetra and bugtool. ServerAddr
+// advertises the unix listener (when one runs) so in-pod tooling
+// defaults to the trusted IPC channel rather than the TCP listener,
+// which may require TLS / mTLS.
 func saveInitInfo() error {
+	addr := ""
+	if path, ok := resolveUnixSocketPath(option.Config.ServerAddress); ok {
```
ExceptionalHandler, line 129: We need to make these changes specific to Linux. 
On Windows, tetragon does not start with the error:
`level=error source=C:/git/tetragon/pkg/logger/slog.go:179 msg="Failed to execute tetragon" error="starting unix gRPC listener: listen unix:///var/run/tetragon/tetragon.sock: GetFileAttributesEx \\var\\run\\tetragon: The system cannot find the path specified."`

ExceptionalHandler (changes requested):

cmd/tetragon/main.go
sayboras: ah good point :100:, no wonder the windows CI job is failing :100:

pkg/certloader/watcher.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func run(ctx context.Context, w *fsnotify.Watcher, tracked []string, log slogger, r *Reloader) {
+	defer func() { _ = w.Close() }()
+	// Drive the debounce timer from this goroutine so ctx.Done()
+	// deterministically preempts a pending reload.
+	timer := time.NewTimer(time.Hour)
+	if !timer.Stop() {
+		<-timer.C
+	}
```
kkourt: I don't understand why we start a timer and then immediately stop it. Could you maybe add a comment to explain it?

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/containerd/cgroups/v3 v3.1.3
 	github.com/deckarep/golang-set/v2 v2.9.0
 	github.com/fatih/color v1.19.0
+	github.com/fsnotify/fsnotify v1.10.0
```
mtardy: fyi some stuff has been going on lately with fsnotify https://github.com/fsnotify/fsnotify/issues/757, might be worth pining the version or not. But would be cool to take a look and see what we do if we decide to explicitly depend on it. We had some discussion off-line that I could tell you as well if needed.

pkg/certloader/watcher.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+// slogger is the subset of slog used here, fakeable in tests.
+type slogger interface {
+	Info(msg string, args ...any)
+	Error(msg string, args ...any)
+	Debug(msg string, args ...any)
+}
+
+func run(ctx context.Context, w *fsnotify.Watcher, tracked []string, log slogger, r *Reloader) {
```
kkourt: I can't wrap my head around this funciton. Can we add some comments to make its intention more clear?

pkg/certloader/watcher.go
```diff
@@ -0,0 +1,139 @@
...
+	// deterministically preempts a pending reload.
+	timer := time.NewTimer(time.Hour)
+	if !timer.Stop() {
+		<-timer.C
+	}
+	pending := false
+	schedule := func() {
+		if pending && !timer.Stop() {
+			select {
+			case <-timer.C:
```
kkourt: Why is this needed?

Reading from https://pkg.go.dev/time#Timer.Stop:

> For a chan-based timer created with NewTimer(d), as of Go 1.23, any receive from t.C after Stop has returned is guaranteed to block rather than receive a stale time value from before the Stop; if the program has not received from t.C already and the timer is running, Stop is guaranteed to return true. Before Go 1.23, the only safe way to use Stop was insert an extra <-t.C if Stop returned false to drain a potential stale value. See the [NewTimer](https://pkg.go.dev/time#NewTimer) documentation for more details.

Could this be something that was needed post v1.23?

cmd/tetragon/main.go
```diff
@@ -681,47 +684,134 @@ func getExporter(ctx context.Context, server *server.Server) (*exporter.Exporter
...
 }
 
-func Serve(ctx context.Context, listenAddr string, srv *server.Server, logSrv *eventlog.Server) error {
-	// we use an empty listen address to effectively disable the gRPC server
-	if len(listenAddr) == 0 {
+// Serve starts the Tetragon gRPC server. listenAddr drives the topology:
+// "" disables gRPC; "unix://X" runs a single plaintext listener at X;
+// on Linux, a TCP address runs that listener plus an in-pod plaintext
+// unix socket at the platform default path, with --server-tls-* gating
+// the TCP path. Windows skips the sidecar unix listener.
```
kkourt, line 704: If I understand correctly, the service on the Unix domain socket will run implicitly when the listen address is specified? This seems too convoluted and, I think it means that there is no way to run a TCP (tls or otherswise) gRPC server alone (without a unix socket)?

Here's a couple of alternatives for discussion:
1. allow for multiple instances of --server-address (so we can have one for unix, one for TCP)
2. add a --tls-server-address for specifying a TCP listening address for the TLS endpoint.

2. makes more sense to me.

kkourt (commented):
Thanks! 

cmd/tetragon/main.go
sayboras: yes, your understanding is correct. My main reason is to avoid any breakage with in-pod tetra command, it's in trusted IPC in local host. 

Re 1) I thought about option 1, but then we need to have TLS settings for each server address as well, I think it's a little bit overkill and potentially introduce some breaking change in helm (or make helm config more complicated with 2 phase rollout)

Re 2) Option 2 can work with another helm flag grpc.tls.address. I didn't think about this approach tbh. The only concern I have is that non-tls address then could be with TCP, and it's kind of a `workaround` and less secure if you also have TLS enabled.

Re 3) Apart from avoiding any breaking change, it's actually how we did with hubble i.e. unix socket domain is configured automatically in addition to TLS listener.

pkg/certloader/watcher.go
sayboras: ah good point :100:

pkg/certloader/watcher.go
sayboras: sure, my intention is to make sure the reload only runs once the timer fires without further activity for debounceWindow, in case of multiple events happening sequentially.

pkg/certloader/watcher.go
sayboras: this is to avoid stale value, but it's no longer required after go 1.23.

pkg/certloader/watcher.go
```diff
@@ -0,0 +1,127 @@
...
+	dirs := map[string]struct{}{}
+	for _, p := range tracked {
+		dirs[filepath.Dir(p)] = struct{}{}
+	}
+	for d := range dirs {
+		if err := w.Add(d); err != nil {
+			_ = w.Close()
+			return fmt.Errorf("certloader: watching dir %q: %w", d, err)
+		}
+	}
```
mtardy: why the use of the intermediary map, does fsnotify screams if you pass two times the same dir?

pkg/certloader/watcher.go
```diff
@@ -0,0 +1,127 @@
...
+
+	log := logger.GetLogger().With("component", "certloader.Watch")
+	go run(ctx, w, tracked, log, r)
+	return nil
+}
+
+// absolutePaths normalizes the configured TLS file paths to absolute form.
+func absolutePaths(cfg Config) ([]string, error) {
```
mtardy: this feels like it should be a method on Config nop? I thought it was a generic function but it's actually returning stuff from the Config in a special form.

pkg/certloader/watcher.go
```diff
@@ -0,0 +1,127 @@
...
+			log.Error("fsnotify watcher error, scheduling reconciliation reload", logfields.Error, err)
+			timer.Reset(debounceWindow)
+		}
+	}
+}
+
+// triggersReload reports whether an fsnotify event concerns a file we must
+// react to: a directly tracked TLS file, or the ..data symlink swapped
+// atomically by Kubernetes Secret projection.
+func triggersReload(ev fsnotify.Event, tracked []string) bool {
```
mtardy: I feel like you could find a more explicit name for this maybe? It feels like this thing is like "ignore this if specific event happens"

pkg/certloader/watcher.go
```diff
@@ -0,0 +1,127 @@
...
+
+		case ev, ok := <-w.Events:
+			// fsnotify Events channel closed: watcher is gone, exit.
+			if !ok {
+				return
+			}
+			// Ignore unrelated paths; arm the debounce timer for ours.
+			if triggersReload(ev, tracked) {
+				log.Debug("TLS file change detected", "event", ev.String())
+				timer.Reset(debounceWindow)
```
mtardy: shouldn't you reload the file here then? I think I'm mis-understanding this

pkg/certloader/tls.go
```diff
@@ -0,0 +1,187 @@
...
+// NewReloader loads the cert material described by cfg. The TLS files must
+// exist on disk; use [NewReloaderLazy] when material is provisioned out of
+// band (cert-manager, cilium-certgen) and may not be present at startup.
+func NewReloader(cfg Config) (*Reloader, error) {
+	if err := cfg.Validate(); err != nil {
+		return nil, err
+	}
+	if !cfg.Enabled() {
+		return nil, errors.New("TLS is not enabled: server cert and key are required")
+	}
```
mtardy: I feel like you should swap those two nop? Validate is doing the same check anyway returning an error so the second one might be useless?

pkg/certloader/tls.go
```diff
@@ -0,0 +1,187 @@
...
+		return errors.New("client CA bundle is only honored when require-client-cert is true")
+	}
+	return nil
+}
+
+// Reloader holds TLS material and atomically swaps it on Reload, so a
+// rotation does not interrupt in-flight handshakes. Safe for concurrent use.
+type Reloader struct {
+	cfg     Config
+	snap    atomic.Pointer[snapshot]
```
mtardy, line 56: why do you have this on top of the lock, efficiency?

pkg/option/validate_test.go
```diff
@@ -0,0 +1,42 @@
...
+		{name: "mtls ok", c: config{ServerAddress: tcp, ServerTLSCertFile: "a", ServerTLSKeyFile: "b", ServerTLSRequireClientCert: true, ServerTLSClientCAFiles: []string{"c"}}},
+		{name: "mtls without cert", c: config{ServerAddress: tcp, ServerTLSRequireClientCert: true, ServerTLSClientCAFiles: []string{"c"}}, wantErr: true},
+		{name: "mtls without ca", c: config{ServerAddress: tcp, ServerTLSCertFile: "a", ServerTLSKeyFile: "b", ServerTLSRequireClientCert: true}, wantErr: true},
+		{name: "ca without mtls", c: config{ServerAddress: tcp, ServerTLSCertFile: "a", ServerTLSKeyFile: "b", ServerTLSClientCAFiles: []string{"c"}}, wantErr: true},
+		{name: "tls with unix address", c: config{ServerAddress: "unix:///run/tetragon.sock", ServerTLSCertFile: "a", ServerTLSKeyFile: "b"}, wantErr: true},
+		{name: "tls with empty address", c: config{ServerAddress: "", ServerTLSCertFile: "a", ServerTLSKeyFile: "b"}, wantErr: true},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
```
mtardy, line 33: this feels a bit overkill to use `t.Parallel()` here, unsure

pkg/option/validate_test.go
mtardy, line 1: I would call the file `validate_test.go` to welcome other tests later

cmd/tetragon/main_linux.go
```diff
@@ -12,10 +12,33 @@ import (
...
 	"github.com/cilium/tetragon/pkg/reader/namespace"
 	"github.com/cilium/tetragon/pkg/reader/proc"
+	"github.com/cilium/tetragon/pkg/server"
 
 	"github.com/spf13/viper"
 )
 
+// defaultUnixSocketPath is the in-pod IPC socket. tetragon-info.json
+// advertises it so co-located tooling does not have to ship TLS material.
+const defaultUnixSocketPath = "/var/run/tetragon/tetragon.sock"
```
mtardy: fwiw we have a `pkg/defaults` with similar consts https://github.com/cilium/tetragon/blob/main/pkg/defaults/defaults.go

docs/content/en/docs/installation/grpc-tls.md
```diff
@@ -0,0 +1,135 @@
...
+
+Tetragon serves its gRPC API on two listeners:
+
+* An always-on **unix-domain socket** at `/var/run/tetragon/tetragon.sock`,
+  reserved for in-pod IPC and protected by file permissions.
+* An optional **TCP listener** for off-host clients. TLS only applies here.
+
+This page shows how to enable TLS, switch to mTLS, and connect with `tetra`.
+
+{{< note >}}
```
mtardy: nit: I'd make this a `caution` maybe

docs/content/en/docs/installation/grpc-tls.md
mtardy, line 1: in general for your doc (that is great btw thanks), I'll reserve bold for https://kubernetes.io/docs/contribute/style/style-guide/#use-bold-for-user-interface-elements. Otherwise looks good

mtardy (commented):
Here are a few comments, mostly reviewed the first commit and the last with the docs.

I'm curious about the use of `t.Parallel()` everywhere, is this something we should do elsewhere, how much time does it save? Is it a good pattern to follow?

go.mod
sayboras: good point :100:, I spend sometime today to check how fsnotify is used, seems like I can just use the below package directly

sigs.k8s.io/controller-runtime/pkg/certwatcher

pkg/certloader/watcher.go
sayboras: I just found out that the below package is available, so remove the hand rolled implementation

sigs.k8s.io/controller-runtime/pkg/certwatcher

pkg/certloader/watcher.go
sayboras: based on my local testing, it's not, but duplicated event unnecessary.

Related comment https://github.com/cilium/tetragon/pull/4950#discussion_r3257021733

pkg/certloader/watcher.go
sayboras: https://github.com/cilium/tetragon/pull/4950#discussion_r3257021733

pkg/certloader/tls.go
sayboras: good catch :100:

pkg/option/validate_test.go
sayboras: it's not much different, tbh the test skeleton was from Goland template (especially if i generate tabular tests). While there is some value (e.g. -race), I don't think we run such setting in CI, removing these t.Parrallel(), thanks.

pkg/option/validate_test.go
sayboras: :100:

cmd/tetragon/main_linux.go
sayboras: good point :100:

docs/content/en/docs/installation/grpc-tls.md
sayboras: ah noted and thanks :100:

pkg/certloader/tls.go
sayboras: Yes, iiuc it's the go-to for read-heavy workloads; handshakes only need a consistent snapshot, not strictly the latest one.

cmd/tetragon/main.go
kkourt: >> 3. makes more sense to me.

Apologies, this was a typo. I meant to write that "Option 2. makes more sense to me."

> Re 2) Option 2 can work with another helm flag grpc.tls.address. I didn't think about this approach tbh. The only concern I have is that non-tls address then could be with TCP, and it's kind of a workaround and less secure if you also have TLS enabled.

If understand correctly, your concern is that it would be possible to have TCP addresses that are not over TLS? I think that's fine. Indeed, I would say that's desirable since it does not break users that rely on this.

@mtardy @3u13r any opinions on this?

sayboras: yes, it's exactly my concern. 

Assume that TLS/mTLS is enabled, the all the connectivity **must** be either via TLS/mTLS or unix domain socket (which is more secure as local host access and perms are required).

3u13r: I think having both at the same time makes sense. 
Security-wise it also makes sense that the UDS is always-on, since it should usually require root and root can also restart Tetragon with UDS as server-address. 

From a yaml POV I'm thinking along the lines of
```yaml
grpc:
  server-socket-address: /var/run/tetragon/tetragon.sock # or "" to disable
  server-address:
    address: localhost:54321
    insecure-disable-tls: false # user need to define "insecure-disable-tls: true"
```
We can then have the same split between the UDS path and server address in the agent flags.
For compatibility reasons, we can also opt for disabling tls by default for now. Note that this breaks the existing values.yaml structure. Do we know how problematic this is for users to adapt?

No matter what, I think we need to adapt the [server-address docs](https://github.com/cilium/tetragon/pull/4950/changes#diff-86f174fc9775bbb185c8b78215839423e2171c9570422da90fe41a80a9c6985bR187)

ExceptionalHandler (approved):

cmd/tetragon/main.go
sayboras: I see two open questions here:

1. Should we allow a TLS + non-TLS (non-UDS) listener together?
2. Should we auto-configure a UDS listener when TLS is enabled?

On (1): This is a decision we need to agree on. My opinion is no, for security reasons. The only use case that I think of is for internal and external traffic, but best practice is to keep TLS on even for internal traffic (e.g. the motivation of service mesh projects). Worth noting: since the TLS flag is off by default, there's no breaking change for existing plain-text users on the upgrade path.

On (2): Agreed — and Leonard's comment seems aligned. 

Based on the outcome of 1) and 2), I can try to come up with helm structure to avoid/minimize breaking changes.

3u13r: I don't have a super strong opinion on (1).
Given the case that the user uses a service mesh, they should just not configure the tls section.

Then the only reason I see, as you also already identified, a split traffic (internal vs. external) situation. I'd think that usually the external traffic is encrypted at the Gateway/Ingress level and then becomes internal traffic, i.e. they can still only encrypt "external" traffic by putting a Gateway in front of it.

I guess I agree, let's go with not allowing TLS and non-TLS (non-UDS) together.

kkourt: > On (1): This is a decision we need to agree on. My opinion is no, for security reasons. The only use case that I think of is for internal and external traffic, but best practice is to keep TLS on even for internal traffic (e.g. the motivation of service mesh projects). Worth noting: since the TLS flag is off by default, there's no breaking change for existing plain-text users on the upgrade path.

I cannot think of a legitimate use-case for having a TLS and a non-TLS listener together, so I'm OK with disallowing it.

pkg/option/validate_test.go
mtardy: you still have plenty of them in the PR, I'm not against adding those if they make things more efficient or are correct, I just never used them but maybe we should!

mtardy (approved):
I see some comments were not answered but you replace maybe some of your code with new controller runtime pkg. It's actually hard to see in the GitHub compare since there is also rebases going on. So not sure if some comments are not relevant anymore.

I think that would be good to have multiple acks on that one, I mostly ack for the docs that is top notch and the general Go which overall makes sense to me. But maybe the certs/TLS things would be nice to have a bit more scrutiny from someone with experience on this that could review after you added that new pkg :). Also see this discussion needs to be resolved https://github.com/cilium/tetragon/pull/4950#discussion_r3234284567.

pkg/certloader/tls.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// Reload re-reads the cert/key and CA bundle from disk and atomically
+// replaces the active snapshot. Concurrent calls are serialized by reloadMu.
+func (r *Reloader) Reload() error {
+	r.reloadMu.Lock()
+	defer r.reloadMu.Unlock()
+	cert, err := tls.LoadX509KeyPair(r.cfg.CertFile, r.cfg.KeyFile)
+	if err != nil {
+		return fmt.Errorf("loading server cert/key: %w", err)
```
kkourt, line 104: Should we set the r.snap to nil here?
If there was a valid file before, but now there is not, should we continue using the old file?

The same question can also be asked for every failure in this function.

pkg/certloader/tls.go
```diff
@@ -0,0 +1,187 @@
...
+type snapshot struct {
+	cert    *tls.Certificate
+	caPool  *x509.CertPool
+	require bool
+}
+
+// NewReloader loads the cert material described by cfg. The TLS files must
+// exist on disk; use [NewReloaderLazy] when material is provisioned out of
+// band (cert-manager, cilium-certgen) and may not be present at startup.
+func NewReloader(cfg Config) (*Reloader, error) {
```
kkourt: AFIACT, this method is only used for tests?

If so, could we have it as a private method and add an appropriate comment? We could also move it to a _test.go file.

pkg/certloader/export_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+package certloader
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+)
+
+// Snapshot exposes the current snapshot for in-package tests.
+func (r *Reloader) Snapshot() (*tls.Certificate, *x509.CertPool, bool) {
```
kkourt: Does this method need to be public?

install/kubernetes/tetragon/values.yaml
```diff
@@ -481,3 +544,37 @@ rthooks:
...
     name: ""
+
+# -- cilium-certgen settings used by tetragon.grpc.tls.auto.method=cronJob.
+# Mirrors the same block in cilium/cilium so operators familiar with the
+# hubble TLS workflow find consistent knobs.
+certgen:
+  image:
+    override: ~
+    repository: quay.io/cilium/certgen
+    tag: v0.2.5
```
kkourt, line 556: Do we need a renovate tag here?

kkourt (approved):
Thanks! 🚀

This looks great! I would love to have the different auto methods introduced as different patches, but I think that's too much to ask. I'm also assuming that a lot of these are used in cilium/cilium, so they are already battle-tested (at least to some extent)

Other than some minor comments, LGTM!

pkg/certloader/export_test.go
sayboras: Good point :100:

install/kubernetes/tetragon/values.yaml
sayboras: yes, I will address this one separately, tbh I don't know how to do multi-line renovate upgrade now :sweat:.

pkg/certloader/tls.go
sayboras: ah good point :100:

pkg/certloader/tls.go
sayboras: my thinking was to keep severing the old but valid cert till it's expired

install/kubernetes/tetragon/values.yaml
sayboras: ah I was overthinking, probably the below will just work

```
# renovate: datasource=docker depName=quay.io/cilium/certgen
```

sayboras:
Merge this in as baseline, I will take a look at CI flakes/issues/improvement if any.

install/kubernetes/tetragon/values.yaml
sayboras: yes, it is working as expected https://github.com/cilium/tetragon/pull/5063

pkg/certloader/tls.go
kkourt: Makes sense, thanks!

kkourt: I thought of another potential reason for keep serving the old valid cert.

AFAIU, we need to update two files so it's possible by the time we read them only one is updated while the other is not. We will eventually retry, but until we do -- it's a good idea to not disrupt operations.

PR #4949, opened 2026-05-05
tetragon: add btfType and btfTypeModule

olsajiri:
- allow to change base resolve type with `btfType`
- allow to specify kernel module with `btfTypeModule` as source of the `btfType` type

example:
```
args:
- index: 1
  type: "uint16"
  label: "sockaddr_alg.salg_family"
  btfType: "sockaddr_alg_new"
  btfTypeModule: "af_alg"
  resolve: "salg_family"
- index: 1
  type: "string"
  label: "sockaddr_alg.salg_name"
  btfType: "sockaddr_alg_new"
  btfTypeModule: "salg_family"
  resolve: "salg_name"
```

PR #4956, opened 2026-05-06
feat: Add support for sockaddr_alg type

kevsecurity:
### Description
Add support for the sockaddr_alg type. Supported operators are ```Family``` (= "AF_ALG"), ```Equal/NotEqual/Prefix/NotPrefix``` (for the salg_name), ```ALGType``` (salg_type), ```ALGFeat``` (salg_feat) and ```ALGMask``` (salg_mask).

Special credit to @AshishNaware who inspired this PR. I basically copied #3980 and replaced all the sockaddr_un references with sockaddr_alg references, and now we have basic support! Without Ashish's PR, this would probably not have happened (unless someone decided to vibe code it quickly!).

### Changelog
Add support for the sockaddr_alg type.

```release-note
Add support for the sockaddr_alg type
```


kevsecurity:
I added a test program that creates an AF_ALG socket and, if successful, attempts to bind it. This doubles as detection for AF_ALG support in the kernel, and as a trigger for the unit test. When the test program is built (alg-bind), the Makefile runs it and echos its output. This reveals whether the kernel has AF_ALG support. By echoing its output, we ignore its exit code so that ```make``` doesn't confuse it with a build error.

kobrineli (approved):

kevsecurity:
Closed in favour of #4949 which will provide better access in a more generic form.

PR #4959, opened 2026-05-08
docs: Remove reference to followForks

andrewstrohman:
This configuration only applies to the PID selector. So, it doesn't make sense to mention it for the binary selectors.


mtardy (approved):
yep thanks

PR #4973, opened 2026-05-08
btf  : Fix union/anonymous struct resolution

### LINKED ISSUE #4974: Resolve fails with structs that have unions

When handling structs with union inside (like `key`) `resolve` returns an error when trying to access the elements inside the union.

```yaml
kind: TracingPolicy
metadata:
  name: "test-key"
spec:
  options:
  - name: disable-kprobe-multi
    value: "1"
  kprobes:
  - call: "security_key_alloc"
    syscall: false
    args:
    - index: 0
      type: "string"
      resolve: "type.name"
```

```
"function_name": "security_key_alloc",
    "args": [
      {
        "error_arg": {
          "Message": "2"
        }
      }
    ],
```

### Tetragon Version

Tested with main

### Kernel Version

6.19.0

### Kubernetes Version

_No response_

### Bugtool

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

tdaudi:
Fixes https://github.com/cilium/tetragon/issues/4974

## Description

Since the resolve algorithm was implemented, Union and anonymous struct types were incorrectly resolved. The reason for this is that the final element found offset were written on top on the union/anonymous struct.

With this fix, the offset of the union/anonymous struct is incremented only when the final element matches.

No test were added because one already exist for union/anonymous struct. Therefore I just updated the logic of the test because this one was also broken.

## Example

With the following policy
```yaml
kprobes:
  - call: "security_key_alloc"
    syscall: false
    args:
    - index: 0
      type: "string"
      resolve: "type.name"
```

### Before

The event look like this

```
"function_name": "security_key_alloc",
    "args": [
      {
        "error_arg": {
          "Message": "2"
        }
      }
    ],
```

### After

```
"function_name": "security_key_alloc",
    "args": [
      {
        "string_arg": "rxrpc"
      }
    ],
```

### DIFF
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -334,6 +334,7 @@ func processMembers(
 				}
 				continue
 			}
+			btfArgs[i].Offset += member.Offset.Bytes()
 			return lastTy, nil
 		}
 		if member.Name != pathToFound[i] {
--- a/pkg/btf/btf_test.go
+++ b/pkg/btf/btf_test.go
@@ -290,29 +290,29 @@ func getBTFPointer(ty btf.Type) (*btf.Pointer, error) {
 	return nil, fmt.Errorf("Invalid type for \"%v\", expected \"*btf.Pointer\", got %q", t, reflect.TypeOf(ty).String())
 }
 
-func findMemberInBTFStruct(structTy *btf.Struct, memberName string) (*btf.Member, error) {
+func findMemberInBTFStruct(structTy *btf.Struct, memberName string) (*btf.Member, uint32, error) {
 	for _, member := range structTy.Members {
 		if member.Name == memberName {
-			return &member, nil
+			return &member, 0, nil
 		}
 
 		if anonymousStructTy, ok := member.Type.(*btf.Struct); ok && len(member.Name) == 0 {
 			for _, m := range anonymousStructTy.Members {
 				if m.Name == memberName {
-					return &m, nil
+					return &m, uint32(member.Offset.Bytes()), nil
 				}
 			}
 		}
 
-		if unionTy, ok := member.Type.(*btf.Union); ok {
+		if unionTy, ok := member.Type.(*btf.Union); ok && len(member.Name) == 0 {
 			for _, m := range unionTy.Members {
 				if m.Name == memberName {
-					return &m, nil
+					return &m, uint32(member.Offset.Bytes()), nil
 				}
 			}
 		}
 	}
-	return nil, fmt.Errorf("Member %q not found in struct %v", memberName, structTy)
+	return nil, 0, fmt.Errorf("Member %q not found in struct %v", memberName, structTy)
 }
 
 func getBTFPointerAndSetConfig(ty btf.Type, btfConfig *api.ConfigBTFArg) (*btf.Pointer, error) {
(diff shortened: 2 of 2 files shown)


andrewstrohman:
I tested this change with:
```
#include <stdio.h>
#include <stdint.h>

struct yetanotherstruct {
        char *first;
        int second;
};

struct anotherstruct {
        int someint;
        struct {
                char achar;
        };
        struct {
                uint32_t someuint;
                struct {
                        char * charp;
                        struct {
                                int first;
                                struct yetanotherstruct second;
                        };
                };
        };
};


struct mystruct {
        int someint;
        struct {
                struct {
                        int anotherint;
                        struct anotherstruct another;
                };
        };
};



void passit(struct mystruct *s) {
        printf("%d\n", s->another.second.second);
}


int main(int argc, char **argv) {
        struct mystruct mys;
        mys.another.second.second = 7;
        passit(&mys);
}
```
And policy:
```
kind: TracingPolicy
metadata:
  name: "uprobe"
spec:
  uprobes:
  - path: "/home/andy/src/nested-union-test"
    BTFPath: "/home/andy/src/nested-union-test.btf"
    symbols:
    - "passit"
    args:
    - index: 0
      type: "int"
      resolve: "another.second.second"
      BTFType: "mystruct"
```

And the event was as expected:
```
{"process_uprobe":{"process":{"exec_id":"YnVpbGRlcjo5OTIzODU1MDc5MDYwNzc4OjQxNTg3NDE=","pid":4158741,"uid":1000,"cwd":"/home/andy/build/src","binary":"/home/andy/build/src/nested-union-test","flags":"execve clone","start_time":"2026-05-08T20:32:48.215629845Z","auid":1000,"parent_exec_id":"YnVpbGRlcjoyNzUzMDAwMDAwMDoxNjEz","refcnt":1,"tid":4158741,"in_init_tree":false},"parent":{"exec_id":"YnVpbGRlcjoyNzUzMDAwMDAwMDoxNjEz","pid":1613,"uid":1000,"cwd":"/home/andy/build/src","binary":"/usr/bin/bash","flags":"procFS","start_time":"2026-01-13T23:55:40.666568747Z","auid":1000,"parent_exec_id":"YnVpbGRlcjoyNzQ2MDAwMDAwMDoxNTU4","tid":1613,"in_init_tree":false},"path":"/home/andy/src/nested-union-test","symbol":"passit","policy_name":"uprobe","args":[{"int_arg":7}],"action":"KPROBE_ACTION_POST"},"node_name":"builder","time":"2026-05-08T20:32:48.215891402Z"}
```


I also tested another variant of that program, where pointers are used, and that works too:
```
#include <stdio.h>
#include <stdint.h>

struct yetanotherstruct {
        char *first;
        int second;
};

struct anotherstruct {
        int someint;
        struct {
                char achar;
        };
        struct {
                uint32_t someuint;
                struct {
                        char * charp;
                        struct {
                                int first;
                                struct yetanotherstruct *second;
                        };
                };
        };
};


struct mystruct {
        int someint;
        struct {
                struct {
                        int anotherint;
                        struct anotherstruct *another;
                };
        };
};



void passit(struct mystruct *s) {
        printf("%d\n", s->another->second->second);
}


int main(int argc, char **argv) {
        struct mystruct mys;
        struct anotherstruct mya;
        struct yetanotherstruct myy;

        mys.another = &mya;

        mya.second = &myy;

        myy.second = 7;
        passit(&mys);
}
```

andrewstrohman (approved):

kkourt:
Added a `needs-backport` label for both 1.7 and 1.6. I am not sure how easy the backport is for 1.6 though.

kkourt:
> I tested this change with:
[...]

Thanks, @andrewstrohman!

Do you think we can add these tests in our go tests?

kkourt:
> Added a `needs-backport` label for both 1.7 and 1.6. I am not sure how easy the backport is for 1.6 though.

At least for 1.7, I had to backport https://github.com/cilium/tetragon/pull/4900 as well.

andrewstrohman:
> Do you think we can add these tests in our go tests?

[Here](https://github.com/cilium/tetragon/pull/4987) is the PR to do it.

PR #4977, opened 2026-05-09
k8s: Migrate to controller-runtime reconcilers and abstract informer pattern

sayboras:
### Description

The main goal is to leverage controller-runtime framework for reconciling resources. Additionally, this PR is also to hide client-go informer from consumer, with one benefit of avoiding init() registration.

### Changelog

```release-note
k8s: Migrate to controller-runtime reconcilers and abstract informer pattern
```


cmd/tetragon/main_k8s.go
```diff
@@ -60,13 +59,18 @@ func initK8s(ctx context.Context) (watcher.PodAccessor, error) {
 	return podAccessor, nil
 }
 
-func initK8sPolicyWatcher(ctx context.Context) error {
+func initK8sPolicyWatcher(_ context.Context) error {
```
mtardy: maybe remove the arg completely? the nok8s version is doing nothing

pkg/watcher/crdwatcher/tracingpolicy_namespaced_reconciler.go
```diff
@@ -0,0 +1,86 @@
...
+// resources. Structurally identical to TracingPolicyReconciler, with the
+// namespace propagated through Add/Delete calls into sensors.Manager.
+type TracingPolicyNamespacedReconciler struct {
+	Client  client.Client
+	Sensors sensorManager
+}
+
+// Reconcile drives Add/Delete on sensors.Manager from the live state of the
+// TracingPolicyNamespaced CR.
+func (r *TracingPolicyNamespacedReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
mtardy, line 33: could you share the code between the two reconcilers? you could also probably share the unit tests so we don't need to have everything twice?

Maybe having integrations tests for both is good on the other side like we have now.

pkg/manager/podevents.go
```diff
@@ -0,0 +1,106 @@
...
+	if _, err := p.sink.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj any) {
+			pod, ok := obj.(*corev1.Pod)
+			if !ok {
+				return
+			}
+			handler(pod)
+		},
+	}); err != nil {
+		logger.GetLogger().Warn("PodEventSource: failed to register Add handler", logfields.Error, err)
```
mtardy: should we actually bubble errors from there? Is it possible to change the interface so that you bubble from here to `RegisterPodDeleteHandler` to the main initialization function?

I don't see the point of logging here, I think we should error if it's used during initialization.

pkg/cgidmap/register.go
```diff
@@ -0,0 +1,73 @@
...
+
+// PodEventSource is the narrow capability cgidmap needs to keep its pod→
+// container-cgroup mappings up to date. The concrete adapter lives in
+// pkg/manager; tests can pass a hand-rolled fake satisfying the same
+// interface.
+type PodEventSource interface {
+	OnPodAdd(handler func(pod *corev1.Pod))
+	OnPodUpdate(handler func(oldPod, newPod *corev1.Pod))
+	OnPodDelete(handler func(pod *corev1.Pod))
+}
```
mtardy: why do we redefine one if it's exactly the same set of needed methods as `pkg/manager/podevents.go.PodEventSource`? I think I understood for the other one as it's only using OnPodDelete but not sure to follow here

cmd/tetragon/main_k8s_unix.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/cilium/tetragon/pkg/cgidmap"
+	"github.com/cilium/tetragon/pkg/logger/logfields"
+	"github.com/cilium/tetragon/pkg/manager"
+	"github.com/cilium/tetragon/pkg/option"
+)
+
+// registerCgidmapPodHandlers wires cgidmap into the supplied pod-event source
+// when the cgidmap feature is enabled. cgidmap is unavailable on Windows, so
+// this function is built only for non-Windows targets.
+func registerCgidmapPodHandlers(events manager.PodEventSource) {
```
mtardy: nit: I know the naming is all over the place but you could follow the existing `EnableCgIDmap` case thing

cmd/tetragon/main_k8s_unix.go
```diff
@@ -0,0 +1,25 @@
...
+	if !option.Config.EnableCgIDmap {
+		return
+	}
+	if err := cgidmap.Register(events); err != nil {
+		log.Warn("failed to register cgidmap pod handlers", logfields.Error, err)
```
mtardy: why not return an error here as well? I think it's important to stop here if the flag didn't work

pkg/cgidmap/register.go
```diff
@@ -0,0 +1,73 @@
...
+// Register wires cgidmap into the supplied pod event source. The caller is
+// expected to gate this call on cgidmap being enabled, so a no-op
+// registration when disabled does not appear in the startup log.
+func Register(events PodEventSource) error {
+	m, err := GlobalMap()
+	if err != nil {
+		// If cgidmap is disabled, this should not be called at all — the
+		// caller is responsible for the gate. Surface the error so a
+		// misconfiguration fails loudly instead of silently no-opping.
+		if errors.Is(err, cgidDisabled) {
```
mtardy: kinda out of scope, but if you'd like to rename this var properly, it's confusing as of now that it's an error
https://github.com/cilium/tetragon/blob/557cb8727957fa3ecc65ec92e20faebf6ee51047/pkg/cgidmap/cgidmap.go#L246

basically removing the linter exception

pkg/cgidmap/register.go
```diff
@@ -0,0 +1,73 @@
...
+			return err
+		}
+		logger.GetLogger().Warn("failed to retrieve cgidmap, not registering pod handlers", logfields.Error, err)
+		return err
```
mtardy: why the logging + returning the error? Maybe just `return fmt.Errorf("...")`?

pkg/policyfilter/state.go
```diff
@@ -307,54 +291,47 @@ func (m *state) updatePodHandler(pod *v1.Pod) error {
...
-				logger.GetLogger().Warn(fmt.Sprintf("policyfilter, delete-pod handler: unexpected object type: %T", pod))
-				return
-			}
-			podID, err := uuid.Parse(string(pod.UID))
-			if err != nil {
-				logger.GetLogger().Warn("policyfilter, delete-pod: failed to parse id", "pod-id", pod.UID, logfields.Error, err)
-				return
-			}
+// onPodAdd handles a new pod observed by the cluster. The handler refreshes
+// pod→policy mappings from the pod's current labels and containers.
```
mtardy: nit: I've seen in some other comments you've been using non ascii characters, I'd do simple `->` if that's possible

pkg/watcher/crdwatcher/tracingpolicy_reconciler.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Spec changed (or new object): mirror the previous informer handler's
+	// Delete+Add cycle. Delete is best-effort; Add is the real outcome.
+	if delErr := r.Sensors.DeleteTracingPolicy(ctx, tp.TpName(), ""); delErr != nil {
+		log.Debug("delete-before-add returned an error", logfields.Error, delErr)
+	}
+	log.Info("adding tracing policy", "info", tp.TpInfo())
+	if addErr := r.Sensors.AddTracingPolicy(ctx, tp); addErr != nil {
+		log.Warn("adding tracing policy failed", logfields.Error, addErr)
+	}
```
mtardy, line 70: question: so we reuse the same pattern as previously, not retrying, should we now retry since we use the reconciler or not? maybe we should try to install the TP as long as the user deployed it? idk if in some situations it might be actually useful or not, I imagine the reason why it errored would not change over time in this case.

pkg/policyfilter/policyfilter.go
```diff
@@ -8,12 +8,23 @@ package policyfilter
...
 )
 
+// PodEventSource is the narrow capability policyfilter needs to consume pod
+// lifecycle events. The concrete adapter lives in pkg/manager; tests can pass
+// a hand-rolled fake satisfying the same interface.
+type PodEventSource interface {
+	OnPodAdd(handler func(pod *v1.Pod))
+	OnPodUpdate(handler func(oldPod, newPod *v1.Pod))
+	OnPodDelete(handler func(pod *v1.Pod))
+}
```
mtardy: another copy, maybe we want to reuse the og?

pkg/manager/manager.go
```diff
@@ -370,10 +394,17 @@ func (cm *ControllerManager) addPodInformer() error {
 	if err != nil {
 		return nil
```
mtardy: pre-existing bug it seems
```suggestion
		return err
```

mtardy (commented):
Thanks, this seems like a very nice cleanup, I have some questions, not sure I followed everything :)!!

pkg/watcher/crdwatcher/tracingpolicy_namespaced_reconciler.go
sayboras: The reconcile method for each is pretty small, let's start with integration test first. 

Basically, having generic in a Reconciler is kind of un-common

pkg/cgidmap/register.go
sayboras: ah I was thinking about consumer defined interface pattern, which helps with unit tests, and also cyclic import issue.

pkg/policyfilter/state.go
sayboras: oh thanks, updated my goland setting :100:

pkg/watcher/crdwatcher/tracingpolicy_reconciler.go
sayboras: good point, I think you are right, let's make use of the framework with retry + exponential backoff :100:

pkg/cgidmap/register.go
mtardy: yeah maybe that's the way to go!

pkg/watcher/crdwatcher/tracingpolicy_reconciler.go
mtardy: So I'm really not sure this is the right call here, could something change on the host so that the error could go away at some point? Because if it's just to have tetragon retry forever, not sure it's good.

mtardy (commented):
Thanks, I would approve I think we should just discuss a little bit more on https://github.com/cilium/tetragon/pull/4977#discussion_r3267934122. Not sure changing the behavior is the right call, sorry was just a question originally to spark a discussion.

pkg/watcher/crdwatcher/tracingpolicy_reconciler.go
sayboras: ah i got what you mean, I cannot think of any case that the error is resolved over time. 

Let me wrap the error with TerminalError() so that requeue is not happening.

pkg/manager/manager.go
```diff
@@ -116,6 +117,22 @@ func (cm *ControllerManager) Start(ctx context.Context) {
 	})
 }
 
+// RegisterControllerWhenCRDReady runs setup once crdName is observed. The
+// wait is a manager Runnable, so registration works before or after
+// Manager.Start. Use this for Reconcilers whose CRD may not yet be installed
+// at agent startup.
+func (cm *ControllerManager) RegisterControllerWhenCRDReady(crdName string, setup func(ctrlManager.Manager) error) error {
+	return cm.Manager.Add(ctrlManager.RunnableFunc(func(ctx context.Context) error {
+		if err := cm.WaitCRDs(ctx, map[string]struct{}{crdName: {}}); err != nil {
```
will-isovalent: Consider using WaitCRDsResync here. Since this gets called twice, I think there is a race where the informer gets removed for one subroutine while the other is still waiting on it. I believe WaitCRDsResync solves this exact problem from what I remember.

mtardy (commented):
maybe I'm wrong but I see no changes push since my last review, any other reasons you re-ask for review?

sayboras:
> maybe I'm wrong but I see no changes push since my last review, any other reasons you re-ask for review?

Oh I think the only changes I made was to use [TerminalError](https://github.com/cilium/tetragon/pull/4977#discussion_r3310627228)

mtardy (commented):
> > maybe I'm wrong but I see no changes push since my last review, any other reasons you re-ask for review?

> Oh I think the only changes I made was to use https://github.com/cilium/tetragon/pull/4977#discussion_r3310627228

I don't see that you pushed anything to this PR for 3 weeks

pkg/manager/podevents.go
mtardy: you didn't change this is it expected?

pkg/cgidmap/register.go
```diff
@@ -0,0 +1,57 @@
...
+	})
+	return nil
+}
+
+func deletePodHandler(m Map, pod *corev1.Pod) {
+	podID, err := uuid.Parse(string(pod.UID))
+	if err != nil {
+		logger.GetLogger().Warn("cgidmap, podDeleted: failed to parse pod id", "pod-id", pod.UID, logfields.Error, err)
+		return
+	}
```
mtardy, line 49: so what was the decision against not returning error and logging on these handlers?

mtardy (commented):
just some questions, then in general I think it's pretty good. Would love another pair of eyes on this maybe?

pkg/watcher/crdwatcher/tracingpolicy_reconciler_integration_test.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+		if time.Now().After(deadline) {
+			return fmt.Errorf("update conflict after deadline: %w", err)
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+}
+
+// TestReconciler_PredicateBehavior asserts the four key behaviors documented
+// in the issue for both Reconciler implementations: create -> Add; status /
```
kkourt: Is this a GH issue?

pkg/watcher/crdwatcher/tracingpolicy_reconciler_integration_test.go
```diff
@@ -0,0 +1,242 @@
...
+// annotation patch -> silent; spec mutate -> Delete+Add; delete -> Delete.
+//
+// Requires a Kind cluster with the TracingPolicy and TracingPolicyNamespaced
+// CRDs installed (same scaffolding pkg/manager/manager_integration_test.go
+// relies on).
+func TestReconciler_PredicateBehavior(t *testing.T) {
```
kkourt, line 172: This test seems complicated :)
Can you add a summary of what are the things that it tests and how?

pkg/watcher/crdwatcher/tracingpolicy_reconciler_test.go
```diff
@@ -0,0 +1,231 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
+	"github.com/cilium/tetragon/pkg/tracingpolicy"
+)
+
```
kkourt, line 29: Ditto: these test seem complicated :)
Can you add a summary of what they test and how?

kkourt (approved):
Couple of nits, but otherwise LGTM. Thanks!

pkg/cgidmap/register.go
sayboras: The main reason is mainly due to underlying cache.ResourceEventHandler interface has no return at all.

pkg/watcher/crdwatcher/tracingpolicy_reconciler_integration_test.go
sayboras: sure, its mainly to test the trigger and expected action. For example, if there is annotation change and no generation bump, the reconcilation should not be triggered and no Add or Delete policy happens.

pkg/watcher/crdwatcher/tracingpolicy_reconciler_test.go
sayboras: Sure, I have added a short summary. 

Most of patterns are from my previous work in [Cilium Gateway API](https://github.com/cilium/cilium/tree/main/operator/pkg/gateway-api), we actually have quite a number of unit tests (~459)

pkg/manager/podevents.go
sayboras: oh sorry, my bad.

mtardy (approved):
overall lgtm. Btw because you rebase before force pushing every time it makes it almost impossible to see what you changed through GitHub UI. I'll suggest not rebasing unless it's useful, or pushing this separately so that we can see the changes :). That's the downside of only rebase and merge...

sayboras:
> I'll suggest not rebasing unless it's useful, or pushing this separately so that we can see the changes :). That's the downside of only rebase and merge...

Good point, noted and thanks :100:

PR #4979, opened 2026-05-10
test: use testing/synctest for deterministic time

sayboras:
### Description
Convert tests that rely on time.Sleep / wall-clock polling to testing/synctest so they run deterministically and near-instantly, removing flake potential in CI.

### Changelog

```release-note
test: use testing/synctest for deterministic time
```


FedeDP (approved):
Yay! Thank you so much!

PR #4988, opened 2026-05-11
fix(manager): resolve race conditions in CRD waiting and manager initialization

kalakotima:
PR Summary

This PR fixes five production-grade bugs discovered in manager.go through static analysis and race detector review. The bugs span across three functions- Get(), WaitCRDs(), and WaitCRDsWithResync() .

All bugs are reproducible under go test race and represent real failure modes in multi-go routine Kubernetes controller environments where informer callbacks and the main goroutine operate concurrently.

requesting to find fixes mentioned below.

1. Get() — Silent panic on initialization failure
// BEFORE: error swallowed, process panics with no recovery path
initOnce.Do(func() {
    manager, err = newControllerManager()
    if err != nil {
        panic(err)  // ❌ callers have no chance to handle this
    }
})
return manager
err was scoped inside initOnce.Do(...) and never surfaced to the caller
Any failure in newControllerManager() caused an unrecoverable panic

2. WaitCRDs() — Unsynchronized map mutation from informer goroutine
// BEFORE: crds map written from informer goroutine, read from caller goroutine
AddFunc: func(obj any) {
    delete(crds, crdObject.Name)  // ❌ no lock — DATA RACE
    if len(crds) == 0 {
        wg.Done()                 // ❌ wg.Done() can underflow if called twice
    }
}
// ...
wg.Wait() // ❌ blocks forever if ctx is cancelled
The caller's crds map was directly mutated inside the informer callback (different goroutine) with no synchronization
wg.Wait() had no context cancellation path — goroutine leak on shutdown

3. WaitCRDsWithResync() — Race on remainingCRDs map and completed bool
// BEFORE: both accessed from informer goroutine without a lock
completed := false          // ❌ plain bool, written from informer goroutine

AddFunc: func(obj any) {
    if completed { return } // ❌ read without lock
    delete(remainingCRDs, crdObject.Name) // ❌ write without lock
    if len(remainingCRDs) == 0 {
        completed = true    // ❌ write without lock
        finish()
    }
}
remainingCRDs map and completed bool both had concurrent read/write from the informer goroutine with no mutex
Detectable by go test -race — classified as undefined behaviour in Go

4. WaitCRDsWithResync() — TOCTOU window in completion signalling
// BEFORE: wg + separate goroutine has a race window
go func() {
    select {
    case <-ctx.Done():
        finish()       // goroutine may not run before wg.Wait() unblocks
    case <-done:
    }
}()
wg.Wait()

if ctx.Err() != nil { // ❌ could be nil even after cancellation
    return ctx.Err()
}
A separate goroutine was used to handle ctx.Done() but wg.Wait() could unblock before that goroutine was scheduled
ctx.Err() check after wg.Wait() had a TOCTOU window — cancellation could be missed

also please find the 1 liner summary fixes 

Thankyou.

### DIFF
--- a/pkg/manager/manager.go
+++ b/pkg/manager/manager.go
@@ -1,6 +1,8 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright Authors of Tetragon
 
+//go:build !nok8s
+
 package manager
 
 import (
@@ -39,6 +41,7 @@ import (
 var (
 	initOnce, startOnce sync.Once
 	manager             *ControllerManager
+	managerErr          error // : Capture init error for safe retrieval
 	_                   watcher.PodAccessor = (*ControllerManager)(nil)
 )
 
@@ -51,15 +54,14 @@ type ControllerManager struct {
 	podInformer     cache.SharedIndexInformer
 }
 
-func Get() *ControllerManager {
-	var err error
+// Get returns the singleton ControllerManager, initializing it on first call.
+//  Now returns error instead of silently panicking on init failure,
+// giving callers a chance to handle the error gracefully.
+func Get() (*ControllerManager, error) {
 	initOnce.Do(func() {
-		manager, err = newControllerManager()
-		if err != nil {
-			panic(err)
-		}
+		manager, managerErr = newControllerManager()
 	})
-	return manager
+	return manager, managerErr
 }
 
 // newControllerManager creates a new controller manager. The enableMetrics flag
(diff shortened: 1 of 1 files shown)


pkg/manager/manager.go
```diff
@@ -41,6 +41,7 @@ import (
 var (
 	initOnce, startOnce sync.Once
 	manager             *ControllerManager
+	managerErr          error // : Capture init error for safe retrieval
```
ExceptionalHandler, line 44: Does this even compile? 
I do not see a corresponding main.go change!

ExceptionalHandler (changes requested):

sayboras:
This race condition is probably addressed by https://github.com/cilium/tetragon/pull/4977

Issue #4923, opened 2026-04-28
check and report faults encountered in read_arg()


andrewstrohman:
### What happened? How can we reproduce this?

read_arg(), and generic_path_offload() indirectly call probe_read() and do not check the return value. Many of the types that these functions read are only found in kernel space, so memory will not need to be paged in. But even for types that only apply to kernel memory, the pointer may be NULL.

So, it would be best to check the return value of functions like probe_read() and if the return value indicates error, set the corresponding arg status to signify that the arg could not be read.

### Tetragon Version

v1.6

### Kernel Version

all

### Relevant log output

```shell

```

andrewstrohman:
We should decide if want to enhance error reporting by relaying the error code to be visible in the event. Currently, all error status codes cause the arg status to be -1, so the reason for error is lost.

PR #4989, opened 2026-05-12
bpf: add errmetrics more generously

### LINKED ISSUE #4952: add errmetrics in the BPF code where they are missing.
https://github.com/cilium/tetragon/commit/494cbe5bfcb754cc57dd12c6ae5ea17fb2ac019d introduced errmetrics and they were later improved by https://github.com/cilium/tetragon/pull/4120, but there are still locations in the BPF code where we don't track unexpected failures.

E.g., calls to `cgrp_get_tracker_id`.

3u13r:
Fixes #4952

### Description
Add errmetrics around function calls that don't yet handle the return code and can fail silently.

Those are mostly calls of `probe_read` and `map_update_elem`. Note that we didn't add this to all such cases since we run against the instruction limit on 4.19, but also on 5.10 if we try to add errmetrics to all cases.

### Changelog

```release-note
bpf: add errmetrics to more function calls
```


3u13r:
Converted to draft until we are aligned [in the issue](https://github.com/cilium/tetragon/issues/4952#issuecomment-4426880889).

andrewstrohman:
For the failures that occur in `read_arg()` and the functions that it calls (directly and indirectly), I have a plan to communicate the error via arg status. See [here](https://github.com/cilium/tetragon/issues/4923). This is preferable to setting the metric because it will prevent bogus argument values in the events. Instead, users will see that the arg could not be read.

In [this PR](https://github.com/cilium/tetragon/pull/4863) I laid the groundwork to do this, and handled a `probe_read_str` error within `copy_strings`. With this groundwork, we just need to return a negative value from these functions in order to plumb the error to userspace.

3u13r:
This is ready to review. The last remaining question is if we want to use up all the instruction space with covering more but not all of the probe_reads.

@andrewstrohman Thanks a lot for the pointer! I converted the errmetrics in that path to just return the error.

bpf/cgroup/bpf_cgroup_events.h
```diff
@@ -49,7 +49,7 @@ send_cgrp_event(struct bpf_raw_tracepoint_args *ctx,
 	msg->cgrp_data.level = cgrp_track->level;
 	msg->cgrp_data.hierarchy_id = cgrp_track->hierarchy_id;
 	memcpy(&msg->cgrp_data.name, &cgrp_track->name, KN_NAME_LENGTH);
-	probe_read_str(&msg->path, PATH_MAP_SIZE - 1, path);
+	with_errmetrics(probe_read_str, &msg->path, PATH_MAP_SIZE - 1, path);
```
andrewstrohman, line 52: https://docs.ebpf.io/linux/helper-function/bpf_probe_read_str/

bpf_probe_read_str() returns a positive number on success.

`with_errmetrics` will call `errmetrics_update` if the return code is not 0.

Maybe we should change `with_errmetrics` to call `errmetrics_update` only if the return value is negative (instead of non-zero).

bpf/process/generic_calls.h
```diff
@@ -131,7 +132,9 @@ copy_char_buf(void *ctx, long off, unsigned long arg, int argm,
 		return return_error(s, char_buf_saved_for_retprobe);
 	}
 	meta = get_arg_meta(argm, e);
-	probe_read(&bytes, sizeof(bytes), &meta);
+	ret = probe_read(&bytes, sizeof(bytes), &meta);
```
andrewstrohman: I don't understand why we use probe_read() for this, because this is reading from a map value. I see this in other places too, so it's probably just a lack of understanding on my part.

This is not something that needs to understood for this PR, but I thought I would mention it in case someone knows.

bpf/cgroup/bpf_cgroup_events.h
3u13r: Ah [this](https://github.com/cilium/tetragon/commit/13d7a4de0dc409544ce399423c33c484c1580efe#diff-e955cd8cfebb6836258b4a95ad9a7b2c58ee4745c28220dc0b938fcdfa8b8facR56) changed the comment for `with_errmetrics` slightly so that I understood it that is just handles negative values as errors.

+1 to your suggestion, let me see if this breaks anything.

3u13r:
```
git grep -h 'with_errmetrics(' | grep -oP 'with_errmetrics\(\s*\K[^,)]+' | sed 's/^\s*//;s/\s*$//' | sort | uniq -c | sort -rn
```
| Occurrences  | function | docs | Uses positive (>=0) returns for error? |
| ------------- | ------------- | ------------- |  ------------- |
|48 | probe_read| https://docs.ebpf.io/linux/helper-function/bpf_probe_read/ | No |
|10 | map_update_elem| https://docs.ebpf.io/linux/helper-function/bpf_map_update_elem/ | No |
|6 | probe_read_kernel| https://docs.ebpf.io/linux/helper-function/bpf_probe_read_kernel/ | No |
|2 | probe_read_str| https://docs.ebpf.io/linux/helper-function/bpf_probe_read_str/ | No |

Looks good to me.

bpf/process/generic_calls.h
3u13r: So, changing this to e.g.
```c
bytes = meta;
```
passes all e2e tests on rhel8.10. 
Let me create a second PR where we can discuss this/get a second opinion. I couldn't track the original author though, since this function tracks back to the initial Tetragon commit.

bpf/cgroup/bpf_cgroup_events.h
3u13r: See https://github.com/cilium/tetragon/pull/4989/changes/5d1dc4490dd840172e0b2dfbf1c360f0601e282b

bpf/process/generic_calls.h
3u13r: see: https://github.com/cilium/tetragon/pull/5018

bpf/process/generic_calls.h
```diff
@@ -152,7 +156,9 @@ copy_iov_iter(void *ctx, long off, unsigned long arg, int argm, struct msg_gener
 		goto nodata;
 
 	tmp = _(&iov_iter->iter_type);
-	probe_read(&iter_type, sizeof(iter_type), tmp);
+	ret = probe_read(&iter_type, sizeof(iter_type), tmp);
```
andrewstrohman, line 152: types char_buf and char_iovec are special because they have a type-specific way of reporting faults to userspace which predated the generic method of reporting arg status that I recently created. See the call sites of `return_error` to see how it works. A positive number is returned, but the type-specific messages sent to userspace encodes that a fault occurred.

So in these two cases we might want to be consistent in the error reporting method by doing `return return_error(s, char_buf_pagefault);` instead of returning a negative number.

I think that long term, it would be nice to remove these special cases, and make all errors surface through the same generic status value.

bpf/process/generic_calls.h
```diff
@@ -259,23 +276,31 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 		__u32 fd;
 
 		key.tid = get_current_pid_tgid() >> 32;
-		probe_read(&fd, sizeof(__u32), &arg);
+		ret = probe_read(&fd, sizeof(__u32), &arg);
```
andrewstrohman: This is another example of using bpf_probe_read() to read something that is stored in map value. This is so prevalent in the code, there must some reason that I'm not aware of.

@olsajiri do you happen to know why we are using probe_read for reading data that resides in map values? 

I understand why we need this for pointer types, but I don't understand why we need it for integer types, because the arg value is coming from a map value.

bpf/process/bpf_fork.c
```diff
@@ -95,7 +95,8 @@ BPF_KPROBE(event_wake_up_new_task, struct task_struct *task)
 #ifndef __RHEL7_BPF_PROG
 	struct msg_k8s kube;
 
-	__event_get_cgroup_info(task, &kube);
+	if (!__event_get_cgroup_info(task, &kube))
```
andrewstrohman: It looks like `__event_get_cgroup_info` doesn't have a consistent meaning for its return value.

In some error cases, it returns 0, and in others it returns a bitmask that encodes the encountered errors.

bpf/process/generic_calls.h
3u13r: Isn't `arg` even stack-local, since with all inline functions, this boils down to [this](https://github.com/3u13r/tetragon/blob/e20efdcc95034354ee721b19ba7806a7ff7b06ea/bpf/process/generic_calls.h#L603)?

But yeah, this function has a lot of such usage.

bpf/process/bpf_fork.c
3u13r:
```
	cgrp = get_task_cgroup(task, cgrpfs_magic, subsys_idx, &flags);
```
already provides us with error flags, so let's use those to always return non-zero in case of error and 0 in case of success.

3u13r: fixed [here](https://github.com/cilium/tetragon/pull/4989/commits/1e2a6f8b0517ba6cad849d588ee0461314e9a5bb)

bpf/process/generic_calls.h
3u13r: Thanks for the explanation, changed it. Yeah it would make sense to reduce the number of ways we do error reporting eventually :tm:.

bpf/process/generic_calls.h
andrewstrohman: Yes, you're right.  It's stack local, because it got copied into a variable.

I don't understand why we need to probe_read() for reading from the stack-local variable. I see that you have removed this and all the tests pass.

bpf/process/bpf_fork.c
andrewstrohman: Your fix makes `__event_get_cgroup_info` return 0 on success and non-zero on failure. As such, we are calling errmetrics for the success case here.
So, we should remove the `!` from this if condition, right?

3u13r: Dang it, yeah you're right. Will do another round of review myself again before pushing, sorry for the work.
Though it also maybe shows that the test coverage isn't too great on the errmetrics side.

bpf/process/generic_calls.h
3u13r: I can only hypothesize along the lines of
1. Maybe old verifiers didn't like converting e.g., from a long to int
2. it was used as a general purpose memcpy

I think the [original finding](https://github.com/cilium/tetragon/pull/5018/changes) was even more egregious, since we read to one local var and then copy it to another, converting from `long` to `size_t` and using it as argument for a `long` parameter.


I've also tried with this instance, the tests passed too.
Let me create a PR with all instances that I can spot and then we can discuss further in the PR.

bpf/process/bpf_process_event.h
```diff
@@ -311,7 +314,7 @@ __event_get_cgroup_info(struct task_struct *task, struct msg_k8s *kube)
 
 	cgrp = get_task_cgroup(task, cgrpfs_magic, subsys_idx, &flags);
 	if (!cgrp)
-		return 0;
+		return flags;
```
andrewstrohman, line 317: There are two callers of __event_get_cgroup_info. [This caller](https://github.com/cilium/tetragon/blob/d800078bbfe9b005705f14c9c95ffa76052e4a4a/bpf/process/bpf_execve_event.c#L318) alters p->flags with the return value. So, I wanted to confirm that this change in behavior would not cause an unintended consequence for this usage of __event_get_group_info.

On the userspace side, for event reception, we have [this corresponding struct](https://github.com/cilium/tetragon/blob/d800078bbfe9b005705f14c9c95ffa76052e4a4a/pkg/api/processapi/processapi.go#L131-L139). Compare this with the bpf side: https://github.com/cilium/tetragon/blob/d800078bbfe9b005705f14c9c95ffa76052e4a4a/bpf/lib/process.h#L276-L296. So, I think it looks like userspace will ignore the flags and everything else within `struct msg_process process;`. One the bpf side of things, I checked for behavior differences based on flags, and only found [this](https://github.com/cilium/tetragon/blob/d800078bbfe9b005705f14c9c95ffa76052e4a4a/bpf/process/bpf_execve_event.c#L228-L229), and that's unrelated to what's changing here.

So, I don't think there will be functional or user-visible difference because of this. But, since my analysis seems to indicate that most of the flags don't have any effect, I'm worried that I've missed something.

Can you please double check my findings here?

andrewstrohman (approved):
Thanks, LGTM. Since there is a lot going on in this PR, I won't merge this for few days in order to give others an opportunity to review it as well.

Please help remind me to merge this next week, if no one else reviews.

bpf/process/bpf_process_event.h
3u13r: I think they are [parsed](https://github.com/3u13r/tetragon/blob/54f0f77234e9ad71edca6279b0ee0ac63859da26/pkg/process/process.go#L407) into [here](https://github.com/3u13r/tetragon/blob/54f0f77234e9ad71edca6279b0ee0ac63859da26/vendor/github.com/cilium/tetragon/api/v1/tetragon/tetragon.pb.go#L1378-L1381) and user-visible [here](https://github.com/3u13r/tetragon/blob/54f0f77234e9ad71edca6279b0ee0ac63859da26/vendor/github.com/cilium/tetragon/api/v1/tetragon/tetragon.pb.go#L1378-L1381).

Though since we use the those flags as `error_flags`  in [get_task_cgroup](https://github.com/3u13r/tetragon/blob/54f0f77234e9ad71edca6279b0ee0ac63859da26/bpf/lib/bpf_cgroup.h#L258) I think we just get more information on errors.

We might even want to wire the flags in `event_wake_up_new_task` at some point.


So there should be no functional difference, only a log output difference, that we warn the user about to not depend upon.

PR #4990, opened 2026-05-12
new: argNew{Symbol,Addr,Offset} support in uprobe override action

FedeDP:
### Description

This PR introduces new arguments for `Override` action for uprobes.
They allow to redirect the `symbol` call to a new symbol.
```yaml
uprobes:
- path: "test"
  symbols:
  - "malloc"
  selectors:
  - matchActions:
    - action: Override
      argNewSymbol: "malloc_patched"
```

Shortcomings:
* **the new symbol must be binary-compatible with the hooked one! (no check is in place!)**
* the new symbol must be already present in the binary
* uprobe only
* only one uprobe hook symbol/addr/offset is supported when using the new action
* kernel support to mess with raw regs is needed

### Changelog

```release-note
new: argNew{Symbol,Addr,Offset} support in uprobe override action
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -911,13 +911,19 @@ this action for enforcement.
 
 ### Override action
 
-Override action is defined for kprobe and uprobe selectors.
+Override action is defined for kprobe, uprobe and lsmhooks selectors, each one
+of them supporting different arguments:
+| Hook   | Arguments |
+| ------ | --------- |
+| kprobe | `argError` |
+| uprobe | `argError,argRegs,argNewSymbol,argNewAddr,argNewOffset` |
+| lsmhooks | `argError` |
 
-#### Override action for kprobe
+#### Override action for kprobe and lsmhooks
 
 `Override` action allows to modify the return value of call. While `Sigkill`
 will terminate the entire process responsible for making the call, `Override`
-will run in place of the original kprobed function and return the value
+will run in place of the original probed function and return the value
 specified in the `argError` field. It's then up to the code path or the user
 space process handling the returned value to whether stop or proceed with the
 execution.
@@ -971,8 +977,6 @@ ensure they properly follow the [Error Injectable Functions](https://docs.kernel
 Beware, here be dragons!!! Use with caution, it could easily crash traced application.
 {{< /warning >}}
 
-Note that `Override` action can be used only on `x86_64` architecture.
-
 Similar to kprobe, the uprobe `Override` action allows to modify the return value
 of user space call with `argError` argument, like:
 
@@ -988,9 +992,57 @@ uprobes:
 ```
 
 Note that it can be successfully used only when following conditions are met:
-- uprobe is attached to the beggining of the user space function;
+- uprobe is attached to the beginning of the user space function;
 - user space function is called via `call` instruction;
-- user space function returns `int` type.
+- user space function returns `int` type;
+- kernel support to manipulate raw registers is available (6.18+).
+
+It's also possible to override the traced symbol with a new symbol call, like:
+```yaml
+uprobes:
+- path: "test"
+  symbols:
+  - "malloc"
+  selectors:
+  - matchActions:
+    - action: Override
+      argNewSymbol: "malloc_patched"
+```
+
+or even with a new symbol offset or address, eg:
+```yaml
+uprobes:
+- path: "test"
+  symbols:
+  - "malloc"
+  selectors:
+  - matchActions:
+    - action: Override
+      argNewOffset: "0x0000115c"
+```
+
+or:
+
+```yaml
+uprobes:
+- path: "test"
+  symbols:
+  - "malloc"
+  selectors:
+  - matchActions:
+    - action: Override
+      argNewAddr: "0x00001137"
+```
+
+There are, however, some restrictions:
+- kernel support to manipulate raw registers is required (6.18+);
+- new symbol must be already present in the binary;
+- new symbol must have the same signature as the original symbol.
+
+{{< warning >}}
+Since tetragon enforces no verification for the last point, this can lead to
+crashing the traced application.
+{{< /warning >}}
 
 It's possible to override specific registers with arbitrary value with `argRegs`
 argument, like:
--- a/docs/content/en/docs/reference/tracing-policy.md
+++ b/docs/content/en/docs/reference/tracing-policy.md
@@ -955,7 +955,20 @@ Filters specified in macros will be appended to corresponding filters of the sel
         <td><b>action</b></td>
         <td>enum</td>
         <td>
-          Action to execute.<br/>
+          Action to execute.
+The Override action has three variants, depending on what arguments are set
+  1. Override the return value of function
+       Supported hooks: kprobes, uprobes, lsm
+       Arguments: ArgError (return value)
+  2. Override the value of a register
+       Supported hooks: uprobes
+       Arguments: ArgRegs
+  3. Override a function call
+       Supported hooks: uprobes
+       Arguments: One of:
+       - ArgNewSymbol: override call to a new symbol (in the binary)
+	      - ArgNewAddr: override call to a new address (in the binary)
+	      - ArgNewOffset: override call to an offset (in the binary)<br/>
           <br/>
             <i>Enum</i>: Post, Sigkill, Override, GetUrl, DnsLookup, NoPost, Signal, TrackSock, UntrackSock, NotifyEnforcer, CleanupEnforcerNotification, Set<br/>
         </td>
@@ -986,10 +999,41 @@ Filters specified in macros will be appended to corresponding filters of the sel
         </td>
         <td>false</td>
       </tr><tr>
+        <td><b>argNewAddr</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's address.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewOffset</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's offset.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewSymbol</b></td>
+        <td>string</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol name.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
         <td><b>argRegs</b></td>
         <td>[]string</td>
         <td>
-          An arg value for the regs action<br/>
+          An arg value for the override action, uprobe only.<br/>
         </td>
         <td>false</td>
       </tr><tr>
@@ -1524,7 +1568,20 @@ Only valid with the post action and with a rateLimit specified.<br/>
         <td><b>action</b></td>
         <td>enum</td>
         <td>
-          Action to execute.<br/>
+          Action to execute.
+The Override action has three variants, depending on what arguments are set
+  1. Override the return value of function
+       Supported hooks: kprobes, uprobes, lsm
+       Arguments: ArgError (return value)
+  2. Override the value of a register
+       Supported hooks: uprobes
+       Arguments: ArgRegs
+  3. Override a function call
+       Supported hooks: uprobes
+       Arguments: One of:
+       - ArgNewSymbol: override call to a new symbol (in the binary)
+	      - ArgNewAddr: override call to a new address (in the binary)
+	      - ArgNewOffset: override call to an offset (in the binary)<br/>
           <br/>
             <i>Enum</i>: Post, Sigkill, Override, GetUrl, DnsLookup, NoPost, Signal, TrackSock, UntrackSock, NotifyEnforcer, CleanupEnforcerNotification, Set<br/>
         </td>
@@ -1555,10 +1612,41 @@ Only valid with the post action and with a rateLimit specified.<br/>
         </td>
         <td>false</td>
       </tr><tr>
+        <td><b>argNewAddr</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's address.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewOffset</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's offset.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewSymbol</b></td>
+        <td>string</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol name.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
         <td><b>argRegs</b></td>
         <td>[]string</td>
         <td>
-          An arg value for the regs action<br/>
+          An arg value for the override action, uprobe only.<br/>
         </td>
         <td>false</td>
       </tr><tr>
@@ -2693,7 +2781,20 @@ Filters specified in macros will be appended to corresponding filters of the sel
         <td><b>action</b></td>
         <td>enum</td>
         <td>
-          Action to execute.<br/>
+          Action to execute.
+The Override action has three variants, depending on what arguments are set
+  1. Override the return value of function
+       Supported hooks: kprobes, uprobes, lsm
+       Arguments: ArgError (return value)
+  2. Override the value of a register
+       Supported hooks: uprobes
+       Arguments: ArgRegs
+  3. Override a function call
+       Supported hooks: uprobes
+       Arguments: One of:
+       - ArgNewSymbol: override call to a new symbol (in the binary)
+	      - ArgNewAddr: override call to a new address (in the binary)
+	      - ArgNewOffset: override call to an offset (in the binary)<br/>
           <br/>
             <i>Enum</i>: Post, Sigkill, Override, GetUrl, DnsLookup, NoPost, Signal, TrackSock, UntrackSock, NotifyEnforcer, CleanupEnforcerNotification, Set<br/>
         </td>
@@ -2724,10 +2825,41 @@ Filters specified in macros will be appended to corresponding filters of the sel
         </td>
         <td>false</td>
       </tr><tr>
+        <td><b>argNewAddr</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's address.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewOffset</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's offset.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewSymbol</b></td>
+        <td>string</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol name.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
         <td><b>argRegs</b></td>
         <td>[]string</td>
         <td>
-          An arg value for the regs action<br/>
+          An arg value for the override action, uprobe only.<br/>
         </td>
         <td>false</td>
       </tr><tr>
@@ -3262,7 +3394,20 @@ Only valid with the post action and with a rateLimit specified.<br/>
         <td><b>action</b></td>
         <td>enum</td>
         <td>
-          Action to execute.<br/>
+          Action to execute.
+The Override action has three variants, depending on what arguments are set
+  1. Override the return value of function
+       Supported hooks: kprobes, uprobes, lsm
+       Arguments: ArgError (return value)
+  2. Override the value of a register
+       Supported hooks: uprobes
+       Arguments: ArgRegs
+  3. Override a function call
+       Supported hooks: uprobes
+       Arguments: One of:
+       - ArgNewSymbol: override call to a new symbol (in the binary)
+	      - ArgNewAddr: override call to a new address (in the binary)
+	      - ArgNewOffset: override call to an offset (in the binary)<br/>
           <br/>
             <i>Enum</i>: Post, Sigkill, Override, GetUrl, DnsLookup, NoPost, Signal, TrackSock, UntrackSock, NotifyEnforcer, CleanupEnforcerNotification, Set<br/>
         </td>
@@ -3293,10 +3438,41 @@ Only valid with the post action and with a rateLimit specified.<br/>
         </td>
         <td>false</td>
       </tr><tr>
+        <td><b>argNewAddr</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's address.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewOffset</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's offset.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewSymbol</b></td>
+        <td>string</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol name.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
         <td><b>argRegs</b></td>
         <td>[]string</td>
         <td>
-          An arg value for the regs action<br/>
+          An arg value for the override action, uprobe only.<br/>
         </td>
         <td>false</td>
       </tr><tr>
@@ -4100,7 +4276,20 @@ Filters specified in macros will be appended to corresponding filters of the sel
         <td><b>action</b></td>
         <td>enum</td>
         <td>
-          Action to execute.<br/>
+          Action to execute.
+The Override action has three variants, depending on what arguments are set
+  1. Override the return value of function
+       Supported hooks: kprobes, uprobes, lsm
+       Arguments: ArgError (return value)
+  2. Override the value of a register
+       Supported hooks: uprobes
+       Arguments: ArgRegs
+  3. Override a function call
+       Supported hooks: uprobes
+       Arguments: One of:
+       - ArgNewSymbol: override call to a new symbol (in the binary)
+	      - ArgNewAddr: override call to a new address (in the binary)
+	      - ArgNewOffset: override call to an offset (in the binary)<br/>
           <br/>
             <i>Enum</i>: Post, Sigkill, Override, GetUrl, DnsLookup, NoPost, Signal, TrackSock, UntrackSock, NotifyEnforcer, CleanupEnforcerNotification, Set<br/>
         </td>
@@ -4131,10 +4320,41 @@ Filters specified in macros will be appended to corresponding filters of the sel
         </td>
         <td>false</td>
       </tr><tr>
+        <td><b>argNewAddr</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's address.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewOffset</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's offset.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewSymbol</b></td>
+        <td>string</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol name.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
         <td><b>argRegs</b></td>
         <td>[]string</td>
         <td>
-          An arg value for the regs action<br/>
+          An arg value for the override action, uprobe only.<br/>
         </td>
         <td>false</td>
       </tr><tr>
@@ -4669,7 +4889,20 @@ Only valid with the post action and with a rateLimit specified.<br/>
         <td><b>action</b></td>
         <td>enum</td>
         <td>
-          Action to execute.<br/>
+          Action to execute.
+The Override action has three variants, depending on what arguments are set
+  1. Override the return value of function
+       Supported hooks: kprobes, uprobes, lsm
+       Arguments: ArgError (return value)
+  2. Override the value of a register
+       Supported hooks: uprobes
+       Arguments: ArgRegs
+  3. Override a function call
+       Supported hooks: uprobes
+       Arguments: One of:
+       - ArgNewSymbol: override call to a new symbol (in the binary)
+	      - ArgNewAddr: override call to a new address (in the binary)
+	      - ArgNewOffset: override call to an offset (in the binary)<br/>
           <br/>
             <i>Enum</i>: Post, Sigkill, Override, GetUrl, DnsLookup, NoPost, Signal, TrackSock, UntrackSock, NotifyEnforcer, CleanupEnforcerNotification, Set<br/>
         </td>
@@ -4700,10 +4933,41 @@ Only valid with the post action and with a rateLimit specified.<br/>
         </td>
         <td>false</td>
       </tr><tr>
+        <td><b>argNewAddr</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's address.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewOffset</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's offset.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewSymbol</b></td>
+        <td>string</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol name.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
         <td><b>argRegs</b></td>
         <td>[]string</td>
         <td>
-          An arg value for the regs action<br/>
+          An arg value for the override action, uprobe only.<br/>
         </td>
         <td>false</td>
       </tr><tr>
@@ -5486,7 +5750,20 @@ Filters specified in macros will be appended to corresponding filters of the sel
         <td><b>action</b></td>
         <td>enum</td>
         <td>
-          Action to execute.<br/>
+          Action to execute.
+The Override action has three variants, depending on what arguments are set
+  1. Override the return value of function
+       Supported hooks: kprobes, uprobes, lsm
+       Arguments: ArgError (return value)
+  2. Override the value of a register
+       Supported hooks: uprobes
+       Arguments: ArgRegs
+  3. Override a function call
+       Supported hooks: uprobes
+       Arguments: One of:
+       - ArgNewSymbol: override call to a new symbol (in the binary)
+	      - ArgNewAddr: override call to a new address (in the binary)
+	      - ArgNewOffset: override call to an offset (in the binary)<br/>
           <br/>
             <i>Enum</i>: Post, Sigkill, Override, GetUrl, DnsLookup, NoPost, Signal, TrackSock, UntrackSock, NotifyEnforcer, CleanupEnforcerNotification, Set<br/>
         </td>
@@ -5517,10 +5794,41 @@ Filters specified in macros will be appended to corresponding filters of the sel
         </td>
         <td>false</td>
       </tr><tr>
+        <td><b>argNewAddr</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's address.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewOffset</b></td>
+        <td>integer</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol's offset.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+          <br/>
+            <i>Format</i>: int64<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
+        <td><b>argNewSymbol</b></td>
+        <td>string</td>
+        <td>
+          An arg value for the override action, uprobe only.
+The new symbol name.
+Beware that the symbol MUST be binary compatible with the traced uprobe symbol.<br/>
+        </td>
+        <td>false</td>
+      </tr><tr>
         <td><b>argRegs</b></td>
         <td>[]string</td>
         <td>
-          An arg value for the regs action<br/>
+          An arg value for the override action, uprobe only.<br/>
         </td>
         <td>false</td>
       </tr><tr>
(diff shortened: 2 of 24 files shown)


mtardy:
Might be worth chatting with @dwindsor about the "override" API also used from https://github.com/cilium/tetragon/pull/4817.

FedeDP:
Thanks @mtardy for the pointer, didn't notice that PR had a new `Override` sub-action (`clearGoString`).
Will chat with David for sure!

dwindsor:
Not sure what this PR is doing but the overall shape of the clearGoString override is:

```
- matchActions:
 - action: Override
    argIndex: 2
    clearGoString: true
```

dwindsor:
Interesting! A few comments.

1. What runtimes are we supporting with this? Seems like symtab would still be necessary to make this work.

2. > the new symbol must be already present in the binary

 Does this mean a security team needs to pre-compile and seed the binary?

3. Some of the assumptions regarding target function address calculation and`PT_LOAD` will break down unless it's statically-compiled Go. `Offset()` in file.go finds the first `PT_LOAD` segment containing the `vaddr` and uses this as the base. This assumption won't hold across all runtimes / linkers, for instance when creating shared libraries if multiple PT_LOAD segments are created, a vaddr in one could easily collide with a vaddr in another. If the symbol we're replacing (malloc) is in a different segment than the symbol we're replacing it with (malloc_patched), the replacement works. But if it's in another segment, the RIP calculation will silently fail.

FedeDP:
> What runtimes are we supporting with this? Seems like symtab would still be necessary to make this work.

Yes, exactly. In case the binary is going to be stripped, one can provide directly the `Offset`.

> Does this mean a security team needs to pre-compile and seed the binary?

Yes; this is mostly an initial feature set, of course. We can expand it later!

olsajiri:
I think we can already divert to another virtual address with (not sure about the number syntax now): 
```
    - action: Override
      argRegs:
      - "rip=0x400345"
```

I think it's ok to have another action like:
```
    - action: OverrideCall
      newSymbol: "malloc_patched"
```

but let's maybe do what @dwindsor did in here https://github.com/cilium/tetragon/pull/4817/changes#diff-f5be2d582d38138315344f2c3c43f51ea9a300a144358f0b7e17591ff2bbcea9R55 and just set argRegs accordingly for standard uprobe override? we won't need new ebpf code 

also I wonder we could do something like below, but not sure ;-)
```
    - action: Override
      argRegs:
      - "rip=symbol:func_2"
      - "rip=offset:0x35a234"
```

I can see this helpful if we want to jump to existing symbol/offset.. but your spec suggests the binary has already patched function loaded and I'm not sure how would that code get in running process

FedeDP:
> I think we can already divert to another virtual address with (not sure about the number syntax now):

Yep!

>  just set argRegs accordingly for standard uprobe override? we won't need new ebpf code

That is exactly what i did; if you see, bpf code has basically no patch at all, because the new actions is managed exactly like an `Override` for uprobe. See https://github.com/cilium/tetragon/pull/4990/changes#diff-27cf3aef45267b713359a8b5e3cbf2932e0d809290418cc042c7f000d01dbdd2R1264 .

> I can see this helpful if we want to jump to existing symbol/offset.. but your spec suggests the binary has already patched function loaded and I'm not sure how would that code get in running process

Yes, this is just a first PR introducing the new action; right now it is not super useful, it's mostly a wrapper around `Override` action as you noted.

pkg/selectors/kernel.go
```diff
@@ -1260,6 +1261,28 @@ func ParseMatchAction(k *KernelSelectorState, action *v1alpha1.ActionSelector, a
 	case ActionTypeSet:
 		WriteSelectorUint32(&k.data, action.ArgIndex)
 		WriteSelectorUint32(&k.data, action.ArgValue)
+	case ActionTypeOverrideCall:
+		// Only supported for entry uprobes right now
+		if k.isUprobe && action.NewOffset != 0 {
```
mtardy: I feel like you could enforce this on the k8s API with some X-Validation

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1727,6 +1728,31 @@ The `Set` action uses following arguments:
...
+  symbols:
+  - "malloc"
+  selectors:
+  - matchActions:
+    - action: OverrideCall
+      newSymbol: "malloc_patched"
+```
+
+There are, however, some restrictions:
+- Kernel support to mess with raw registers is required.
```
mtardy: "to mess with" seems a bit familiar, maybe "to manipulate" or something else?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1727,6 +1728,31 @@ The `Set` action uses following arguments:
...
+- The new symbol must be already present in the binary.
+- The new symbol must be compatible with the original symbol.
```
mtardy: maybe add details on this?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1727,6 +1728,31 @@ The `Set` action uses following arguments:
...
+- Only a single Symbol/Addr/Offset is supported when using the `OverrideCall` action.
```
mtardy: you should be able to express that with kubebuilder annotation or X-Validation also

mtardy (commented):
Here are some comments, I'll let the uprobe people approve this after discussing with them, not sure I'm relevant merging this :)

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -278,6 +278,17 @@ type ActionSelector struct {
...
+
+	// The new symbol for the overrideCall action.
+	// Beware that the symbol MUST be binary compatible with the traced uprobe symbol.
+	NewSymbol string `json:"newSymbol,omitempty"`
+	// The new symbol's address for the overrideCall action.
+	// Beware that the called symbol MUST be binary compatible with the traced uprobe symbol.
+	NewAddr uint64 `json:"newAddr,omitempty"`
+	// The new symbol's offset for the overrideCall action.
+	// Beware that the new symbol MUST be binary compatible with the traced uprobe symbol.
+	NewOffset int64 `json:"newOffset,omitempty"`
```
kkourt: In terms of policy syntax, how about 

```go
struct OverrideCallArg {
	// Beware that the symbol MUST be binary compatible with the traced uprobe symbol.
	NewSymbol string `json:"newSymbol,omitempty"`
	// The new symbol's address for the overrideCall action.
	// Beware that the called symbol MUST be binary compatible with the traced uprobe symbol.
	NewAddr uint64 `json:"newAddr,omitempty"`
	// The new symbol's offset for the overrideCall action.
	// Beware that the new symbol MUST be binary compatible with the traced uprobe symbol.
	NewOffset int64 `json:"newOffset,omitempty"`
}
```

And then:


```go
type ActionSelector struct {
...
 OverrideCallArg *OverrideCallArg `json:"overrideCall,omitempty"`
}
```

kkourt (commented):
Thanks! I have a comment below about policy syntax.

A few more notes:
* Why were the first and the second patch separated? From my end, it makes reviewing harder.
* Could you please add commit message?

FedeDP:
> Why were the first and the second patch separated? From my end, it makes reviewing harder.

I can merge them, no problem on my side!

> Could you please add commit message?

Yeah...thanks!

pkg/k8s/apis/cilium.io/v1alpha1/types.go
FedeDP: Mmmh that would be the first case where we use a secondary struct inside `ActionSelector`. Also,  i'm not sure i like the updated policy syntax:
```
uprobes:
- path: "test"
  symbols:
  - "malloc"
  selectors:
  - matchActions:
    - action: OverrideCall
      overrideCall:
        newSymbol: "malloc_patched"
```

pkg/selectors/kernel.go
FedeDP: I am not sure whether i can enforce this via X-validation; i should check from within the `ActionSelector` struct if `action == ActionTypeOverrideCall` then check whether the value is embedded in a `UProbeSpec` or not.
I don't think we can express it.

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: As above (https://github.com/cilium/tetragon/pull/4990#discussion_r3303518589) how can we check via x-validation that if `UProbeSpec->[]KProbeSelector->[]MatchActions->Action==OverrideCall` then `(len(symbols) + len(addr) + len(offset)) == 1`?

pkg/k8s/apis/cilium.io/v1alpha1/types.go
kkourt: > Also, i'm not sure i like the updated policy syntax:

Why not?
I feel it's much cleaner that what we did so far because, otherwise, we need to check all fields to see what is relevant for what action.

FedeDP: Nothing against it actually, is just that it feels a little bit off since it is the first time we do something like this (at least for `matchActions`).

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1754,6 +1755,31 @@ The `Set` action uses following arguments:
 - The `argIndex` defines position of the return argument.
 - The `argValue` defined the actual value to write.
 
+### OverrideCall action
+
+OverrideCall action is only defined for uprobe selectors and allows tetragon to route the
+traced symbol calls to a new symbol.
+For example, to override `malloc` calls to `malloc_patched` function,
+the following policy can be used:
```
mtardy: wrapping looks a little bit funny, in vim you should be able to use `gq` to wrap properly

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1754,6 +1755,31 @@ The `Set` action uses following arguments:
...
+  symbols:
+  - "malloc"
+  selectors:
+  - matchActions:
+    - action: OverrideCall
+      newSymbol: "malloc_patched"
+```
+
+There are, however, some restrictions:
+- Kernel support to manipulate raw registers is required.
```
mtardy: do we have the kernel version for this?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1754,6 +1755,31 @@ The `Set` action uses following arguments:
...
+- The new symbol must be already present in the binary.
+- The new symbol must be compatible with the original symbol (ie: same signature); tetragon enforces no verification.
```
mtardy: ie -> i.e. but in general we tend to avoid latin expressions https://kubernetes.io/docs/contribute/style/style-guide/#avoid-latin-phrases

docs/content/en/docs/concepts/tracing-policy/selectors.md
mtardy: What would you think of something like this?

```go
// +kubebuilder:validation:XValidation:rule="(self.action == 'OverrideCall') == (((has(self.newSymbol) && self.newSymbol != '') ? 1 : 0) + (has(self.newAddr) ? 1 : 0) + (has(self.newOffset) ? 1 : 0) == 1)",message="newSymbol, newAddr, or newOffset must be set iff action is OverrideCall"
type ActionSelector struct {
```

or more strict like
```
// +kubebuilder:validation:XValidation:rule="(self.action == 'OverrideCall' && (((has(self.newSymbol) && self.newSymbol != '') ? 1 : 0) + (has(self.newAddr) ? 1 : 0) + (has(self.newOffset) ? 1 : 0)) == 1) || (self.action != 'OverrideCall' && !has(self.newSymbol) && !has(self.newAddr) && !has(self.newOffset))",message="exactly one of newSymbol, newAddr, or newOffset must be set for OverrideCall, and none otherwise"
```

mtardy (commented):
I think overall looks good, only a few nits on the docs. Again I can approve but I think that would be better for other people working on uprobe to greenlight it.

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: Oh right, that comment in the doc is referring to the fact that we only allow a single `uprobe` hook point (either symbol or addr or offset) when `OverrideCall` is used.
It is not referring to the `OverrideCall` action `newSymbol/newAddr/newOffset`. The latter is already managed in my latest push: https://github.com/cilium/tetragon/pull/4990/changes/ab0d909fedadfd89764210f8f4c30633bce3b92e#diff-bc54ba3950ae7bb8c718f46e20d8c315f8ff7661d6950fe362276e8160b26227R270
```
// +kubebuilder:validation:AtMostOneOf=newSymbol;newAddr;newOffset
type ActionSelector struct {
```
In the end, we just need to validate that there can be at most one between them; i did not check that they must be set only when action is `OverrideCall` (we never raised an error if the fields are ignored after all, AFAIK)

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: You are right, it should be 6.18+ unless backported. Let me state it!

pkg/k8s/apis/cilium.io/v1alpha1/types.go
kkourt: @mtardy I think we need one more opinion here :)

mtardy: > I feel it's much cleaner that what we did so far because, otherwise, we need to check all fields to see what is relevant for what action.

I think I agree. It would have been better to group multiple args (even single one) for some action under their own name, otherwise theoretically we should have `overrideCallNewSymbol`, `overrideCallNewAddr`, etc.

It would also make XValidation easier to write like "if you use this action", then "this arg should not been a null struct".

> is just that it feels a little bit off since it is the first time we do something like this

indeed that's true that we never grouped them so far... like many mistakes in this custom resource!

I guess both are fine, it's either a route of:
- committing to the existing messy way
- trying to group them and make it better now

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -83,12 +84,23 @@ type genericUprobe struct {
...
 
-	n := copy(uprobeRegs.Ass[:], regs)
-	if n != len(regs) {
-		logger.GetLogger().Warn("register assignments count mismatch", "#regs", len(regs))
+	// We convert the map of slices to a single flat slice.
+	// This is because we will execute all the registers override
+	// in the same `uprobe_offload` context.
+	// This means that if we have 5 regs Override and 1 OverrideCall,
+	// the length of regsMap[0] will be 6,
+	// and `uprobe_offload` will execute the requested assignment to all of them.
```
olsajiri: this feels like unnecessary complication.. and source of future  headaches
are we expecting Override and OverrideCall in the same spec? is there some use-case?

the way I understand this feature is that we want to make it easier to divert the execution to another offset/address/symbol

I think this could be done just by adding those fields to `ActionSelector` as you did and then just use Override action to detect if any of them is defined and use it to populate instruction pointer register.. perhaps warn/error if current override regs have already instruction pointer register assigned as well

FedeDP: > are we expecting Override and OverrideCall in the same spec? is there some use-case?

Not now; but in the future i am planning to add a `OverrideArgs` (that similar to OverrideCall allows to override argument values); in that case, we will need to support both of them at the same time.

I can post-pone the patch to later, but seems right a good moment now to be future proof.

(Instead, `Override` and `OverrideCall` cannot be set together; it is gated behind a check while creating the uprobe sensor)

olsajiri: hum, still I don't get why we can't have just single Override action that we configure for what we want

FedeDP: I think the idea is to improve the UX for end users; i don't have strong opinions about it though

FedeDP: I see 3 options here:

1. use already-existing `Override` action for everything:

```yaml
- action: Override
  argRegs:
  - "rip=0x400345"
```

and eventually add new arg directly to it for easier UX:
```yaml
- action: Override
  argCall: "malloc_patched"
```

2. add `OverrideCall` (and eventually `OverrideArgs` in the future) to later deprecate `Override.argRegs`:

```yaml
- action: OverrideCall
  newSymbol: "malloc_patched"
```

3. same as 2 but change the policy syntax:

```yaml
- action: OverrideCall
  overrideCall:
        newSymbol: "malloc_patched"
```


This PR implements 2. My idea is that from an UX pov, `override` action should only allow to override the return value (like we do for kprobes), and we introduce new "overrideFoo" specific actions to override symbol call and call arguments.

About 3., my opinion is that it's a good change but would feel a bit out of context considering it would be the very first one with the updated syntax. Anyway, if we want to adapt the new syntax we have to start somewhere :) 

cc @mtardy @kkourt

kkourt: I'm leaning towards option 1), and adding a comment to specify what happens. Maybe something like:

```go
		// The Override action has three variants, depending on what arguments are set
		//   - ArgError -> override the return value of a function (e.g., kprobes)
		//   - ArgRegs  -> override the value of a register (uprobes / USDTs)
		//   - ArgNewCall -> overrides attached call to a function (uprobes / ...)
```

But, all three options are fine, so @FedeDP if you feel strongly about any of them, I'm totally fine.

FedeDP: I'll go with 1 since Jiri already stated that he preferred it too :)

FedeDP: Pushed the changes! :rocket:

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -536,6 +537,32 @@ func validateUprobeSpec(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) err
...
+						numAddressMethods++
+					}
+					if matchAct.ArgNewOffset != 0 {
+						numAddressMethods++
+					}
+					if numAddressMethods > 1 {
+						return errors.New("uprobe Override action needs exactly one of either argNewSymbol, argNewOffset or argNewAddr defined")
+					}
+					state.overrideSymbol = numAddressMethods == 1
+					break outerLoop
```
mtardy: can you just return also here?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -536,6 +537,32 @@ func validateUprobeSpec(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) err
...
+				}
+			}
```
mtardy: nit: this could be a function in the style of HasSomething function we have that take the selectors as input.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -564,7 +591,67 @@ func validateUprobeFeatures(spec *v1alpha1.UProbeSpec, has *uprobeHas) error {
 	return nil
 }
 
+func initOverrideSymbolOffset(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) error {
+	if state.overrideSymbol {
```
mtardy: could you maybe write `if !state.overrideSymbol` and return immediately Go-style so that we don't have an indent here.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -564,7 +591,67 @@ func validateUprobeFeatures(spec *v1alpha1.UProbeSpec, has *uprobeHas) error {
...
+
+					// Store the relative-to-traced-symbol delta.
+					// In bpf, we will compute the override symbol address as:
+					// * curr_ip = PT_REGS_IP(ctx);
+					// * next_ip = curr_ip + uprobeOverrideCallAddr
+					sel.MatchActions[i].ArgNewOffset = int64(overrideSymbOffset) - symbolOffset
+					break outerLoop
+				}
+			}
+		}
```
mtardy: I also feel it would be better as a function for readability.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -564,7 +591,67 @@ func validateUprobeFeatures(spec *v1alpha1.UProbeSpec, has *uprobeHas) error {
...
+			symbolOffset = int64(off)
+		} else if state.offsets > 0 {
+			symbolOffset = int64(spec.Offsets[0])
+		} else if state.addrs > 0 {
+			off, err := f.OffsetFromAddr(spec.Addrs[0])
+			if err != nil {
+				return err
+			}
+			symbolOffset = int64(off)
+		}
```
mtardy: I'm not sure to follow why we copy the length of the object into some var and then check the var to retrieve the value from the object. Should the offsets/addrs/symbols should be named offsetsNumber, addrsLen or something like that then? It's also okay to check len(spec.Offsets[0]) many time, it's fairly efficient in go.

pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -180,6 +181,73 @@ spec:
...
+	if bytes, err = exec.Command(uprobeTest1).Output(); err != nil {
+		t.Fatalf("Failed to execute test binary: %s\n", err)
+	}
+
+	if strings.Contains(string(bytes), "uprobe_test_lib_string_arg_empty") {
+		t.Fatalf("uprobe_test_lib_string_arg_empty should not be called")
+	}
+	if !strings.Contains(string(bytes), "uprobe_test_lib_string_arg__") {
+		t.Fatalf("uprobe_test_lib_string_arg__ should be called")
+	}
```
mtardy: just a note, I guess this was added before but I was wondering why we didn't rely on return code instead of strings, but not sure it makes a bit difference in reliability for testing.

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -991,6 +989,24 @@ Note that it can be successfully used only when following conditions are met:
 - uprobe is attached to the beggining of the user space function;
 - user space function is called via `call` instruction;
 - user space function returns `int` type.
+- kernel support to manipulate raw registers is available (6.18+ unless backported).
```
mtardy: nit: I'd avoid "unless backported". Maybe link to the patch needed from upstream if you want to add more context as it's useful to check if the commit sha is present in our tree or not but since anything could be backported I'd just add 6.18+. bpf docs links to github linux mirror.

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -991,6 +989,24 @@ Note that it can be successfully used only when following conditions are met:
...
+  selectors:
+  - matchActions:
+    - action: Override
+      argNewSymbol: "malloc_patched"
+```
+
+There are, however, some restrictions:
+- kernel support to manipulate raw registers is required (6.18+ unless backported);
+- the new symbol must be already present in the binary;
+- the new symbol must be compatible (same signature) with the original symbol, as tetragon enforces no verification;
```
mtardy: nit: could be wrapped.

Also what happens then? seg fault?

docs/content/en/docs/reference/tracing-policy.md
mtardy, line 1: this is a detail but I'd bundle that with the commit where you added that.

pkg/tracingpolicy/k8s_xvalidation_test.go
```diff
@@ -81,3 +83,49 @@ spec:
...
+`
+	if withSymbol {
+		crd += "        argNewSymbol: \"test_3\"\n"
+	}
+	if withAdrr {
+		crd += "        argNewAddr: 0x985256\n"
+	}
+	if withOff {
+		crd += "        argNewOffset: 0x9156366\n"
+	}
```
mtardy, line 112: nit: maybe using sprintf would be a bit less tricky with spacing here, it's fine either way.

pkg/tracingpolicy/k8s_xvalidation_test.go
```diff
@@ -33,7 +33,8 @@ metadata:
   name: "uprobe"
 spec:
   uprobes:
-  - path: "/usr/bin/test"`
+  - path: "/usr/bin/test"
+`
```
mtardy, line 37: churn change

pkg/tracingpolicy/k8s_xvalidation_test.go
```diff
@@ -81,3 +83,49 @@ spec:
...
+	testUprobeValidationOverrideArgNewSymbolAddrOffset(t, true, true, false)
+}
+
+func TestUprobeValidationOverrideArgNewSymbolOffset(t *testing.T) {
+	testUprobeValidationOverrideArgNewSymbolAddrOffset(t, true, false, true)
+}
+
+func TestUprobeValidationOverrideArgNewAddrOffset(t *testing.T) {
+	testUprobeValidationOverrideArgNewSymbolAddrOffset(t, false, true, true)
+}
```
mtardy, line 135: maybe since those are related
```go
func TestUprobeValidationOverrideArgNewSymbolAddrOffset(t *testing.T) {
	t.Run("NewSymbolAddrOffset", func(t *testing.T) {
		testUprobeValidationOverrideArgNewSymbolAddrOffset(t, true, true, true)
	})

	t.Run("NewSymbolAddr", func(t *testing.T) {
		testUprobeValidationOverrideArgNewSymbolAddrOffset(t, true, true, false)
	})

	t.Run("NewSymbolOffset", func(t *testing.T) {
		testUprobeValidationOverrideArgNewSymbolAddrOffset(t, true, false, true)
	})

	t.Run("NewAddrOffset", func(t *testing.T) {
		testUprobeValidationOverrideArgNewSymbolAddrOffset(t, false, true, true)
	})
}
```

mtardy (commented):
Cool! Just a few comments on code form.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -564,7 +591,67 @@ func validateUprobeFeatures(spec *v1alpha1.UProbeSpec, has *uprobeHas) error {
...
+					}
+					if err != nil {
+						return err
+					}
+
+					// Store the relative-to-traced-symbol delta.
+					// In bpf, we will compute the override symbol address as:
+					// * curr_ip = PT_REGS_IP(ctx);
+					// * next_ip = curr_ip + uprobeOverrideCallAddr
+					sel.MatchActions[i].ArgNewOffset = int64(overrideSymbOffset) - symbolOffset
```
olsajiri: hum, I'm not sure it's always that simple delta.. like that delta between file offsets matches the delta between virtual addresses.. if the new symbol is in another section for whatever reason, the delta would not match, but I haven't check

I'd remove the New, feels redundant, but don't have strong opinion on that

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -564,7 +591,67 @@ func validateUprobeFeatures(spec *v1alpha1.UProbeSpec, has *uprobeHas) error {
 	return nil
 }
 
+func initOverrideSymbolOffset(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) error {
+	if state.overrideSymbol {
+		f, err := elf.OpenSafeELFFile(spec.Path)
```
olsajiri: let's open the file just once in `addUprobe` and pass the file pointer to functions that need it like `addUprobeEntries` and this one

olsajiri (commented):
thanks, left some comments, also some of those commits would deserve changelog ;-)

pkg/sensors/tracing/genericuprobe.go
FedeDP: That would make sense but we still need to count the `numAddressMethods` (bad naming, going to change it btw); unless we call it `selectors.CountOverrideActionNewMethods` or something like that.

pkg/sensors/tracing/genericuprobe.go
FedeDP: Yeah i don't really know either; they were already there before my changes and didn't want to touch them to avoid increasing the patch size.

FedeDP: I think we can eventually address this in a follow up PR.

pkg/sensors/tracing/uprobe_reg_test.go
FedeDP: Oh that was much easier, right :) I didn't think of it actually. Let me fix it!

docs/content/en/docs/reference/tracing-policy.md
FedeDP: Right!

pkg/tracingpolicy/k8s_xvalidation_test.go
FedeDP: Yes not a big deal, i'd leave it as-is :)

pkg/sensors/tracing/genericuprobe.go
FedeDP: I like it!

pkg/tracingpolicy/k8s_xvalidation_test.go
mtardy: misread!! thanks @FedeDP to explain offline

FedeDP: The change is indeed correct since i needed to add a newline.
Basically, this test was faling because the yaml was bad, not because the verification was failing :)

pkg/sensors/tracing/genericuprobe.go
FedeDP: > if the new symbol is in another section for whatever reason, the delta would not match, but I haven't check

Yes that's right; unfortunately i couldn't find a better way to do that. I am not sure whether using `Address()` instead of `Offset()` would work better (ie: using `sym.Value` instead of computing the offset from within the segment).

> I'd remove the New, feels redundant, but don't have strong opinion on that

I think it gives a little more insight, but yes it's not super useful. @kkourt wdyt?

pkg/k8s/apis/cilium.io/v1alpha1/types.go
kkourt, line 291: Can we add a summary of the Override Action here? See https://github.com/cilium/tetragon/pull/4990#discussion_r3458473426.

Maybe something like:
```
// The Override action has three variants, depending on what arguments are set
//   1. Override the return value of function
//        Supported hooks: kprobes
//         Arguments: ArgError (return value)
//   2. Override the value of a register
//      Supported hooks: (uprobes / USDTs)
//      Arguments: ArgRegs
//   3. Override a function call
//      Supported hooks: ...
//      Arguments: One of:
//        - ArgNewSymbol: override call to a new symbol (in the binary)
//	      - ArgNewAddr: override call to a new address (in the binary)
//	     - ArgNewOffset: override call to an offset (in the binary)
```

What is the difference between offset and address? How do these things interact with things like ASLR?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -545,6 +546,31 @@ func validateUprobeSpec(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) err
...
+						numArgNewArgs++
+					}
+					if matchAct.ArgNewOffset != 0 {
+						numArgNewArgs++
+					}
+					if numArgNewArgs > 1 {
+						return errors.New("uprobe Override action needs exactly one of either argNewSymbol, argNewOffset or argNewAddr defined")
+					}
+					state.overrideSymbol = numArgNewArgs == 1
+					return nil
```
kkourt: Can we also validate having multiple override actions? I think we need to reject policies that define multiple override actions.

pkg/sensors/tracing/genericuprobe.go
kkourt: I agree, let's pull this code into a function.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -1026,22 +1111,6 @@ func addUprobeEntries(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, state *u
 		return nil
 	}
 
-	var f *elf.SafeELFFile
```
kkourt, line 1185: is there a reason that this code was moved below?

pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -180,6 +181,68 @@ spec:
 	require.NoError(t, err)
 }
 
+func TestUprobeOverrideNewSymbolAction(t *testing.T) {
```
kkourt: Can we implement this with a policytest?
See https://github.com/cilium/tetragon/blob/main/tests/policytests/uprobes.go

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -988,9 +986,32 @@ uprobes:
...
+  selectors:
+  - matchActions:
+    - action: Override
+      argNewSymbol: "malloc_patched"
+```
+
+There are, however, some restrictions:
+- kernel support to manipulate raw registers is required (6.18+);
+- new symbol must be already present in the binary;
+- new symbol must have the same signature as the original symbol;
```
kkourt: Can we also document argNewAddr and argNewOffset?

docs/content/en/docs/concepts/tracing-policy/selectors.md
kkourt, line 912: A summary of the different override variants and in what hooks they are supported would be useful here, I think.

kkourt (commented):
Thanks!

pkg/sensors/tracing/uprobe_reg_test.go
FedeDP: 100%!

pkg/sensors/tracing/genericuprobe.go
olsajiri: I think you need to compute the delta from virtual addresses, so convert the uprobe and the override location  (if needed) to virtual address and use them for the delta

pkg/k8s/apis/cilium.io/v1alpha1/types.go
FedeDP: Sorry i completely lost track of it! I'll add it now!

> What is the difference between offset and address? How do these things interact with things like ASLR?

Address is the symbol virtual address, while offset is the symbol offset from the start of the segment it is contained in. They are the same of `uprobeSpec.Offsets/Addrs`; the same restrictions we have when using `uprobeSpec.Offsets/Addrs` should apply, since we use the very same methods.

pkg/sensors/tracing/genericuprobe.go
FedeDP: Do we? We never did that before (afaik that could be a broader scope PR, ie: deny all usage of multiple same actions).
I am not sure whether this PR is a good fit anyway.

pkg/sensors/tracing/genericuprobe.go
FedeDP: see: https://github.com/cilium/tetragon/pull/4990#discussion_r3567103013

pkg/k8s/apis/cilium.io/v1alpha1/types.go
FedeDP: Also, considering we compute the delta between traced symbol VA and new symbol VA, and then we jump from IP of the traced symbol + delta, this way ASLR should not be a problem.

pkg/sensors/tracing/genericuprobe.go
FedeDP: Last push fixed this, thank you so much also for the offline pointers :)

pkg/sensors/tracing/uprobe_reg_test.go
FedeDP: So, since this needs to run in vmtests and policytest don't  for now (https://github.com/cilium/tetragon/pull/5247), i opted to add 2 test commits:
* add the policytest for the `argNewSymbol` feat
* add "normal" uprobe_reg_test that also tests `argNewOffset` and `argNewAddr`

pkg/sensors/tracing/genericuprobe.go
kkourt: I don't mind we do it in a different PR, but I don't see a user defining multiple override actions is sensible.

pkg/elf/file.go
```diff
@@ -128,8 +128,6 @@ func (se *SafeELFFile) Offset(name string) (uint64, error) {
 }
 
 func (se *SafeELFFile) OffsetFromAddr(addr uint64) (uint64, error) {
-	offset := uint64(0)
```
kkourt, line 131: This seems like a fix for the `OffsetFromAddr` to support an offset of zero? Maybe it's worth being in its own patch?

pkg/selectors/kernel.go
```diff
@@ -1807,6 +1807,28 @@ func HasOverride(selectors []v1alpha1.KProbeSelector) bool {
...
+					numArgNewArgs++
+				}
+				if action.ArgNewOffset != 0 {
+					numArgNewArgs++
+				}
+				return numArgNewArgs
+			}
+		}
+	}
+	return numArgNewArgs
```
kkourt: Is this always `return 0`?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -573,7 +579,64 @@ func validateUprobeFeatures(spec *v1alpha1.UProbeSpec, has *uprobeHas) error {
...
+				}
+				if err != nil {
+					return err
+				}
+
+				// Store the relative-to-traced-symbol delta.
+				// In bpf, we will compute the override symbol address as:
+				// * curr_ip = PT_REGS_IP(ctx);
+				// * next_ip = curr_ip + uprobeOverrideCallAddr
+				sel.MatchActions[i].ArgNewOffset = int64(overrideSymbAddr - symbolAddr)
```
kkourt: I'm somewhat confused.

There seem to be two "offsets":
1. the `ArgNewOffset` that the user can specify, which seems to be always positive and it's the offset from the start of the ELF segment that the original call is in? That's what we seem to be passing in `AddrFromOffset`
2. there is the `sel.MatchActions[i].ArgNewOffset` which is where we jump relative to the original IP? And this offset can be negative?

Assuming my understanding is correct, can we use appropriately named variables to disambiguate these two offsets?

kkourt (commented):
No specific objections, but the code in this PR is hard for me to follow and would be nice if we could find ways to improve that.

pkg/selectors/kernel.go
FedeDP: Yes definitely, let me make it more obvious.

pkg/elf/file.go
FedeDP: Oh no it's just to cut a couple of lines, since i introduced `AddrFromOffset` i tried to make it as similar as possible. No functional changes.

pkg/sensors/tracing/genericuprobe.go
FedeDP: > there is the sel.MatchActions[i].ArgNewOffset which is where we jump relative to the original IP? And this offset can be negative?

Exactly.

> Assuming my understanding is correct, can we use appropriately named variables to disambiguate these two offsets?

I didn't want to introduce eg: `action.argIPOffset` for this, and neither passing new values to `ParseMatchAction` (eg: by updating `KernelSelectorArgs` structure.) Therefore in the end i settled with reusing the `ArgNewOffset` with a slightly different meaning (note that this is internal only).

FedeDP:
> No specific objections, but the code in this PR is hard for me to follow and would be nice if we could find ways to improve that.

Except from your review comments, is there anything else that you find confusing?

pkg/elf/file.go
FedeDP: I can split this commit into 2, one for `elf/file.go` and one for the everything else. Let me do that!

FedeDP: Done

pkg/sensors/tracing/genericuprobe.go
FedeDP: I can introduce eg: `KernelSelectorArgs.UprobeIPDelta` or something like that if you prefer.

kkourt:
> > No specific objections, but the code in this PR is hard for me to follow and would be nice if we could find ways to improve that.
> 
> Except from your review comments, is there anything else that you find confusing?

Not really. I've tried to flag some of the things that I believe could be improved. To be fair, the code is already complex. Thanks for bearing with me!

pkg/sensors/tracing/genericuprobe.go
FedeDP: Done!

FedeDP:
@kkourt i tried to split the commits in a better way, and fixed your review comments :) PTAL :eyes:

pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -180,6 +184,98 @@ spec:
...
+	// It will return 100 when the symbol is overridden.
+	// It will be considered an error by `cmd.Run()`.
+	require.Error(t, cmd.Run())
+	require.Equal(t, 100, cmd.ProcessState.ExitCode())
+
+	err = jsonchecker.JsonTestCheck(t, checker)
+	require.NoError(t, err)
+}
+
+func TestUprobeOverrideNewSymbolAddrOffsetAction(t *testing.T) {
```
FedeDP, line 277: Couldn't add offset and args test as policytests for the moment; see https://github.com/cilium/tetragon/issues/5284

kkourt (approved):

pkg/sensors/tracing/genericuprobe.go
FedeDP: @kkourt i totally forgot that the check is already in place deep down: https://github.com/cilium/tetragon/blob/main/pkg/selectors/kernel_regs_amd64.go#L17

FedeDP:
Rebased on top of main to fix conflicts.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -545,6 +546,11 @@ func validateUprobeSpec(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) err
 		}
 	}
 
+	numArgNewArgs := selectors.CountOverrideArgNewSymbolAddrOffset(spec.Selectors)
+	if numArgNewArgs > 1 {
+		return errors.New("uprobe Override action needs exactly one of either argNewSymbol, argNewOffset or argNewAddr defined")
```
mtardy: maybe here you can report `numArgNewArgs` you counted for the user?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -573,14 +579,74 @@ func validateUprobeFeatures(spec *v1alpha1.UProbeSpec, has *uprobeHas) error {
...
+func initOverrideSymbolOffset(spec *v1alpha1.UProbeSpec, state *uprobeConfigState, f *elf.SafeELFFile) (int64, error) {
+	if !state.overrideSymbol {
+		return 0, nil
+	}
+
+	// Load probed symbol offset
+	var (
+		symbolAddr uint64
+		err        error
+	)
```
mtardy, line 650: nit: I'd tend to avoid this style for not global level var but that's personal preference here I see we use it already in some places. Maybe we could enable a linter for that.

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -988,9 +991,57 @@ uprobes:
...
+  selectors:
+  - matchActions:
+    - action: Override
+      argNewAddr: "0x00001137"
+```
+
+There are, however, some restrictions:
+- kernel support to manipulate raw registers is required (6.18+);
+- new symbol must be already present in the binary;
+- new symbol must have the same signature as the original symbol;
```
mtardy: nit: if you use `;` you need `.` at the last item

I mean I have been following this rule forever, is it correct?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -988,9 +991,57 @@ uprobes:
 ```
 
 Note that it can be successfully used only when following conditions are met:
-- uprobe is attached to the beggining of the user space function;
+- uprobe is attached to the beginning of the user space function;
 - user space function is called via `call` instruction;
-- user space function returns `int` type.
+- user space function returns `int` type;
+- kernel support to manipulate raw registers is available (6.18+);
```
mtardy: same here with final `.`

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -911,13 +911,18 @@ this action for enforcement.
 
 ### Override action
 
-Override action is defined for kprobe and uprobe selectors.
+Override action is defined for kprobe, uprobe and lsmhooks selectors, each one of them supporting different arguments:
```
```suggestion
Override action is defined for kprobe, uprobe and lsmhooks selectors, each one
of them supporting different arguments:
```

pkg/elf/file.go
```diff
@@ -101,25 +101,9 @@ func (se *SafeELFFile) Offset(name string) (uint64, error) {
...
-				//
-				// fn symbol offset = fn symbol VA - .text VA + .text offset
-				//
-				// stackoverflow.com/a/40249502
-				offset = sym.Value - prog.Vaddr + prog.Off
-				break
-			}
+		offset, err := se.OffsetFromAddr(sym.Value)
+		if err != nil {
+			offset = sym.Value
```
mtardy, line 106: nice

mtardy (approved):
Only nits you can apply/ignore at your conveniance, this LGTM! thanks!

nit: please wrap your git commit log (for example in `new(pkg/sensors): uprobe implementation for override argNewSymbol.`)

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: Yeah i guess so! Let me amend this!

pkg/sensors/tracing/genericuprobe.go
FedeDP: We could enable a linter, but indeed i find it more readable like that, because you know that's a `var` block and all function explicitly declared vars are expected to be there.

PR #4994, opened 2026-05-12
add uprobe digest verification

andrewstrohman:
In some cases, a uprobe policy is only intended to apply to a specific version of a binary, and could have adverse effect if attached to a different version of the binary.

For users that want to only apply the policy for a specific build, they can specify a digest type and value to validate the target binary. The following example policy demonstrates this functionality.

```
kind: TracingPolicy
metadata:
    name: "uprobe-digest-example"
spec:
    uprobes:
    - path: "/bin/example-target-binary"
      symbols:
      - "main"
      binaryDigests:
```

When the above policy is loaded, tetragon will calculate the sha256 hash of `/bin/example-target-binary` and compare it against all the binaryDigests configured within the `binaryDigests` list. If the calculated digest does not match any configured digest, that uprobe is skipped and a warning is reported in policy status.

As you can see from the above example, the format of the entries of the `binaryDigests` list is
`<digest type>:<digest>`.

The following digest types are supported:
- sha1
- sha256
- sha384
- sha512
- build-id (found in the ELF note section named `.note.gnu.build-id`)

When a uprobe is attached, the kernel keeps track of the target by its inode. It's best practice for installation software (such as rpm, deb, install, tar) to unlink a pre-existing target before installing a new version. This ensures that the new version of the binary will have a new inode, so the uprobe will not fire for it.

But, if the target binary is updated without an unlink, which would happen if `cp` is used or if the target is opened and updated in-place, the inode will be retained. In this case, the uprobe will continue firing for the updated binary, potentially causing unintended effect. As such, it's critical that users of this feature ensure their installation method will not retain the inode of the target binary that they want to update.

If the user wants the policy to apply to a new version of the binary post-installation, they need to reload the policy after installation in order to attach to the new version of the binary.

## Implementation Details

In order to prevent TOCTOU issues where the sequence is:
1) we check the digest
2) target binary gets updated
3) attach the binary

I open the target binary and obtain a file descriptor, and then use that file descriptor instead of the path to attach.

A similar problem can happen for elf offset determination, so I open the file description during configuration parsing while creating the sensor. Then I pass a path that points at this open file down to the attach logic that happens during load.

Because I need to keep the file open, to prevent its file descriptor from being repelased, there is a brief period of time when tetragon has an open file for every target file that the sensor attaches to. The files are all closed immediately after loading the sensor.

## Note for reviewers
[This issue](https://github.com/cilium/tetragon/issues/5005) still remains after this refactor. For now, I've just prevented the offending configurations instead of fixing the underlying issue.

## Children PRs
Given feedback to break this into more manageable PRs, this PR will not be merged, and instead it was broken up into the following children PRs

 - https://github.com/cilium/tetragon/pull/5301
 - https://github.com/cilium/tetragon/pull/5280
 - https://github.com/cilium/tetragon/pull/5212
 - https://github.com/cilium/tetragon/pull/5183
 - https://github.com/cilium/tetragon/pull/5182
 - https://github.com/cilium/tetragon/pull/5181
 - https://github.com/cilium/tetragon/pull/5191 (the last two commits came from this PR, while Jiri's commits that proceeded them were created to help review those commits)


pkg/sensors/program/loader_linux.go
```diff
@@ -316,7 +316,21 @@ func uprobeAttachSingle(load *Program, prog *ebpf.Program, spec *ebpf.ProgramSpe
...
+		// to reference the file descriptor, which is a stable reference to the
+		// inode even if the path's inode changes.
+		// This trick ensures that the uprobe attachment is not affected by TOCTOU issues
+		// with the target file.
+		f, err := os.Open(data.Path)
+		if err != nil {
+			return nil, fmt.Errorf("open executable %s: %w", data.Path, err)
+		}
+		defer f.Close()
+		fdPath := fmt.Sprintf("/proc/self/fd/%d", f.Fd())
```
olsajiri: I think we should take into account `option.Config.ProcFS`?

also I wonder it'd make sense just to add fd to the kernel interface.. seems useful to me, will check

pkg/sensors/program/loader_uprobe_linux.go
```diff
@@ -0,0 +1,616 @@
...
+	// - 4 bytes: description size (build ID length)
+	// - 4 bytes: note type (3 for NT_GNU_BUILD_ID)
+	// - 4 bytes: "GNU\0"
+	// - N bytes: build ID data
+	if len(data) < 16 {
+		return "", errors.New("invalid .note.gnu.build-id section: too small")
+	}
+
+	// Skip the header (16 bytes) and extract the build ID
+	buildID := data[16:]
```
olsajiri: IIUC we should parse that section.. plz check sensors/tracing/loader_test.go

pkg/sensors/program/loader_uprobe_linux.go
```diff
@@ -0,0 +1,616 @@
...
+
+func MultiUprobeAttach(load *Program, bpfDir string) AttachFunc {
+	return func(coll *ebpf.Collection, collSpec *ebpf.CollectionSpec,
+		prog *ebpf.Program, spec *ebpf.ProgramSpec) (unloader.Unloader, error) {
+
+		return uprobeAttach(load, bpfDir, coll, collSpec, prog, spec, uprobeAttachMulti)
+	}
+}
+
+func attachSingleMultiUprobe(load *Program, prog *ebpf.Program, path string, attach *MultiUprobeAttachSymbolsCookies, digests []string, bpfDir string, idx int, extra ...string) (link.Link, error) {
```
olsajiri: im bad at naming functions, but I think we should use something else ;-) `attachMultiUpobeLink` ?

pkg/sensors/program/loader_uprobe_linux.go
```diff
@@ -0,0 +1,616 @@
...
+			}
+			symbol = "" // symbol is not used for Go binaries, as we attach by offset only
+		}
+	case Offset:
+		address = configOffset
+		if addressOffset != nil {
+			*addressOffset = address
+		}
+	case Address:
+		address, err = elfFile.OffsetFromAddr(configAddress)
```
olsajiri: hum, so address translation to offset is actually the only reason why we do all this, right? because symbol and offset does not depend on elf file version.. but address might translate into different offset

I wonder it'd be easier change just to make it exception case and translate it to offset in here if it's used to define uprobe, but keep everything else (symbols/offset) where it was

also that AddressOffsets backpointer sucks.. but not sure what else we can do

could we deprecate uprobes specified by address and just ignore them? ;-) any idea if anyone is using that? @kkourt

pkg/sensors/program/loader_linux.go
andrewstrohman: > I think we should take into account option.Config.ProcFS

Yes, thank you. I've done this.

> also I wonder it'd make sense just to add fd to the kernel interface.. seems useful to me, will check

That would nice.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -407,6 +407,9 @@ type UProbeSpec struct {
 	// +kubebuilder:validation:Optional
 	// A return argument to include in the trace output.
 	ReturnArg *KProbeArg `json:"returnArg,omitempty"`
+	// +kubebuilder:validation:Optional
+	// Digests of the traced binary. If any digest matches, verification passes.
+	Digests []string `json:"digests,omitempty"`
```
kkourt: How about `BinaryDigests` here to make this more explicit?

pkg/sensors/program/loader_uprobe_linux.go
kkourt: > hum, so address translation to offset is actually the only reason why we do all this, right? because symbol and offset does not depend on elf file version.. but address might translate into different offset

Not necessarily; I can think of use-cases where you want to load by symbol only on certain binaries (e.g., only on binaries that a certain issue exists).

olsajiri: hum, maybe I should have be more specific.. I meant the move of the resolving code into the link layer.. only the address needs the translation, symbol and offset are fixed.. so only because of address translation we move all this resolve code in the link layer after the digest check

now I wonder we could do that the other way round and get the file descriptor before we try to link the uprobe.. the resolve code stays and we should be able to get path from file descriptor .. and when we get the new kernel interface via fd, we just use it .. could this work @andrewstrohman ?

pkg/k8s/apis/cilium.io/v1alpha1/types.go
andrewstrohman: Ok, sounds good. I did this.

pkg/sensors/program/loader_uprobe_linux.go
andrewstrohman: Yeah, that's better. Thanks, done.

pkg/sensors/program/loader_uprobe_linux.go
andrewstrohman: Thanks. I made a small adjustment to sensors/tracing/loader_test.go, and then moved that build-id extraction logic out into pkg/elf so it could be re-used by this change.

pkg/sensors/program/loader_uprobe_linux.go
andrewstrohman: > only the address needs the translation, symbol and offset are fixed..

There is also the pure go binary case. It resolves the symbols into offsets by walking the elf file.

> now I wonder we could do that the other way round and get the file descriptor before we try to link the uprobe.. the resolve code stays and we should be able to get path from file descriptor .. and when we get the new kernel interface via fd, we just use it .. could this work @andrewstrohman ?

Yes, I think this could work. I will do the digest verification right before the resolve logic, change the resolve logic to operate on the *os.File instead of the path, but leave that resolve logic located where it's currently at. I'll then pass the *os.File down to the link logic instead of the path. Did I interpret your suggestion correctly?

I'll save what I have now in a backup branch, and try it this way. I agree the approach your proposing here is probably better, especially considering the AddressOffsets backpointer.

olsajiri: > Yes, I think this could work. I will do the digest verification right before the resolve logic, change the resolve logic to operate on the *os.File instead of the path, but leave that resolve logic located where it's currently at. I'll then pass the *os.File down to the link logic instead of the path. Did I interpret your suggestion correctly?

yes exactly.. this way the link layer stays with single purpose

also I was wondering do we want some periodic check the digest did not change? not sure what the action would be if the digest does not match, but I guess at least warning for start

andrewstrohman: > also I was wondering do we want some periodic check the digest did not change? not sure what the action would be if the digest does not match, but I guess at least warning for start

I would like to address this in a follow-up. I have created [this issue](https://github.com/cilium/tetragon/issues/5102) to track this. The way I see it, there are two ways that the digest changes:
1) The file is unlinked and a new file (with a new inode) is placed at the same path. In this situation, the probe is no longer attached to anything, so tetragon will not interject.
2) The file is updated in-place. Examples of this include opening the file and writing to it or using the `cp` command. This results in probes being placed at unintended points of the program, resulting in unintended behavior.

In this PR, I have a commit that updated the concepts docs to talk about this, so the user is warned about the impact of doing these things. This way, the user is informed about the consequences until we can put the suggested automated detection into place.

andrewstrohman: @olsajiri I have made the changes that I described above. I'm still testing and cleaning it up a bit, and I need to add some logging for the digest mismatches. But it's very close to being ready.

The downside of this approach is that I need to keep all the target binaries open until the sensor is loaded (as opposed to opening one file at a time, as was done with the earlier approach). Also, I think we have a higher risk of a FD leak in the future.

That being said, I still think this is preferable to the initial approach. Please let me know what you think.

andrewstrohman: > Also, I think we have a higher risk of a FD leak in the future.

I was able to improve on this concern by closing the files in the PostLoadHook, so I feel better about this aspect.

api/v1/tetragon/sensors.proto
```diff
@@ -111,6 +132,8 @@ message TracingPolicyStatus {
   optional TracingPolicyStats stats = 12;
   // domain of the policy
   string domain = 13;
+  // uprobe sensor status
+  UprobeStatus uprobe_status = 14;
```
FedeDP: I am not sure i like the sensor-specific status here but i don't think there is a better solution :/

FedeDP (commented):
I feel like this PR could be split in 3-4:
* first 2 commits
* loader_test fixes
* digest implementation
* uprobe hook status

WDYT?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -456,13 +479,21 @@ func createGenericUprobeSensor(
...
+			if cleanupErr := cleanupUprobeEntries(ids); cleanupErr != nil {
+				return nil, errors.Join(err, cleanupErr)
+			}
+			return nil, err
 		}
 
-		ids, err = addUprobe(&uprobe, ids, &in, &has)
-		if err != nil {
-			return nil, err
+		nextIDs, addErr := addUprobe(&uprobe, ids, &in, &has)
```
kkourt: I'm a bit confused by this, we seem to be calling `cleanupUprobeEntries` both inside `addUprobe` and outside. Would it be possible to do the cleanup in a single place?

pkg/sensors/load.go
```diff
@@ -229,6 +229,10 @@ func (s *Sensor) Unload(unpin bool) error {
 func (s *Sensor) Destroy(unpin bool) error {
 	var errs error
 
+	if s.Destroyed {
+		return nil
+	}
```
kkourt, line 233: Do we know cases where .Destroy() will be called twice? If so, could you add them i a comment here for future reference?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -505,6 +576,30 @@ func createGenericUprobeSensor(
 			openedFiles[uprobe.Path] = entryFile
 		}
 
+		if len(uprobe.BinaryDigests) > 0 {
```
kkourt: This function is getting failry large. Can we move this code into a helper function?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -453,6 +456,74 @@ func cleanupUprobeEntries(ids []idtable.EntryID) error {
...
 }
 
+// verifyFileDigest verifies that an open file's digest matches the configured digest.
+// digestConfig format is "<algo>:<hash>" (e.g., "sha256:abc123..." or "build-id:deadbeef...")
+func verifyFileDigest(file *os.File, digestConfig string, fileHashCache map[string]string) error {
+	if digestConfig == "" {
+		return nil
+	}
+
+	parts := strings.SplitN(digestConfig, ":", 2)
```
kkourt: maybe using strings.Cut is more convinient here?

api/v1/tetragon/sensors.proto
```diff
@@ -84,6 +84,27 @@ message TracingPolicyStats {
   TracingPolicyActionCounters action_counters = 1;
 }
 
+enum UprobeHookState {
+  // unknown or unspecified state
```
kkourt, line 106: Do we need this to be uprobe specific? Can't we have a single state for all hooks?

api/v1/tetragon/sensors.proto
```diff
@@ -84,6 +84,27 @@ message TracingPolicyStats {
...
+  STATUS_LOADED = 1;
+  // uprobe was not loaded due to digest mismatch
+  STATUS_DIGEST_REJECTED = 2;
+}
+
+message UprobeHookStatus {
+  // hook probe status
+  UprobeHookState state = 1;
+  // hook path
+  string path = 2;
```
kkourt: How about a single status for all types of hooks as well?

For example:
```
message HookStatus {
     string section = 1;
     HookState state = 2;
     hookDescription string = 3; // for uprobes it can be path.
}
```

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -638,18 +635,40 @@ func createGenericUprobeSensor(
...
 				for algo, hash := range digestCache {
 					actualHashes = append(actualHashes, fmt.Sprintf("%s:%s", algo, hash))
 				}
-				return nil, fmt.Errorf(
-					"digest verification failed for %s: none of the configured digests matched\nconfigured: [%s]\nactual: [%s]",
-					uprobe.Path,
-					strings.Join(uprobe.BinaryDigests, ", "),
-					strings.Join(actualHashes, ", "),
-				)
+				logger.GetLogger().Warn(
```
kkourt: I'm not sure if this warrants a "Warning". how about Info?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -638,18 +635,40 @@ func createGenericUprobeSensor(
 				}
 			}
 			if !matchFound {
+				if !alreadyOpened {
```
kkourt: This code is becoming difficult to follow. It would be nice if we could factor it so that it's easier to understand (maybe some helper functions would help?)

kkourt (commented):
Thanks Andy!

The overall approach makes sense to me. Please find some comments below.
For what it's worth, I do agree with Fede that it might be helpful to split this into multiple PRs.

And one more note: maybe it makes sense to remove the changes that reject the policy if we end up not doing it? Note also that the k8s description seems to be out of date:

```
	// Digests of the traced binary. If any digest matches, verification passes.
	BinaryDigests []string `json:"binaryDigests,omitempty"`
```

api/v1/tetragon/sensors.proto
FedeDP: That would be really great and also addresses my concern about the sensor-specific status there.

pkg/sensors/handler.go
```diff
@@ -125,6 +125,7 @@ func (h *handler) addTracingPolicy(op *tracingPolicyAdd) error {
 	if err != nil {
 		col.err = err
 		col.state = LoadErrorState
+		col.destroy(true)
```
olsajiri, line 149: please separate generic sensors changes from the uprobe changes and explain them in the changelog

pkg/sensors/tracing/genericuprobe.go
olsajiri: +1, could we just have the defer func logic you added in `addUprobe` moved up to `createGenericUprobeSensor` ?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -491,33 +525,40 @@ func createGenericUprobeSensor(
...
+
+	baseLen := len(ids)
+	committed := false
+
+	defer func() {
+		if committed {
+			return
+		}
+
+		if len(ids) > baseLen {
```
olsajiri: if we move this one level up to `createGenericUprobeSensor` we could just clean everything right? no need to do partial cleaning

pkg/sensors/tracing/genericusdt.go
```diff
@@ -295,7 +331,30 @@ func createUsdtSensorFromEntry(polInfo *policyInfo, usdtEntry *genericUsdt,
...
 }
 
-func addUsdt(spec *v1alpha1.UsdtSpec, in *addUsdtIn, ids []idtable.EntryID, has *usdtHas) ([]idtable.EntryID, error) {
+func addUsdt(spec *v1alpha1.UsdtSpec, in *addUsdtIn, ids []idtable.EntryID, has *usdtHas) (retIDs []idtable.EntryID, retErr error) {
+	var state *selectors.KernelSelectorState
+
+	baseLen := len(ids)
+	committed := false
+
+	defer func() {
```
olsajiri, line 346: same as for uprobes, why not move it to the upper layer and do the cleanup in `createGenericUsdtSensor` ?

pkg/sensors/program/loader_linux.go
```diff
@@ -368,32 +422,11 @@ func uprobeAttachMulti(load *Program, prog *ebpf.Program, spec *ebpf.ProgramSpec
...
-			} else {
-				lnk, err = exec.UprobeMulti(attach.Symbols, prog, opts)
-			}
-			if err != nil {
-				return nil, err
-			}
-			pinExtra := append(extra, strconv.Itoa(idx))
-			err = LinkPin(lnk, bpfDir, load, pinExtra...)
+		for file, attach := range data.Attach {
+			lnk, err := attachMultiUpobeLink(load, prog, file, attach, bpfDir, idx, extra...)
```
olsajiri: please put the `attachMultiUpobeLink` factor/addition into separate commit, even if it's just code move with no changes

pkg/sensors/program/loader_linux.go
```diff
@@ -316,10 +321,21 @@ func uprobeAttachSingle(load *Program, prog *ebpf.Program, spec *ebpf.ProgramSpe
...
+		// The kernel tracks uprobe targets by inode. When we open a file,
+		// its file descriptor points to an inode.
+		// The loader API requires that a path is provided. It will not accept
+		// a file descriptor directly. But we can use the procfs self/fd/<N> symlink
+		// to reference the file descriptor, which is a stable reference to the
+		// inode even if the path's inode changes.
+		// This trick ensures that the uprobe attachment is not affected by TOCTOU issues
+		// with the target file.
+
+		fdPath := procSelfFDPath(data.File)
```
olsajiri: would `data.File.Name()` just do?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -481,15 +488,31 @@ func createGenericUprobeSensor(
...
 			err = fmt.Errorf("append macros selectors: %w", err)
 			if cleanupErr := cleanupUprobeEntries(ids); cleanupErr != nil {
-				return nil, errors.Join(err, cleanupErr)
+				err = errors.Join(err, cleanupErr)
 			}
 			return nil, err
 		}
 
-		nextIDs, addErr := addUprobe(&uprobe, ids, &in, &has)
+		entryFile, ok := openedFiles[uprobe.Path]
```
olsajiri: nit, would it make sense to get absolute path from `uprobe.Path` first?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -144,7 +146,7 @@ func handleGenericUprobe(r *bytes.Reader) ([]observer.Event, error) {
 
 	unix := &tracing.MsgGenericUprobeUnix{}
 	unix.Msg = &m
-	unix.Path = uprobeEntry.path
+	unix.Path = uprobeEntry.file.Name()
```
olsajiri: iiuc we close the file descriptor after we load the sensor, so this `file` in here is closed, right?
I think we should keep the uprobeEntry.path for that, and just add `uprobeEntry.file`

pkg/sensors/tracing/genericusdt.go
```diff
@@ -652,7 +688,7 @@ func handleGenericUsdt(r *bytes.Reader) ([]observer.Event, error) {
 
 	unix := &tracing.MsgGenericUsdtUnix{}
 	unix.Msg = &m
-	unix.Path = uprobeUsdt.path
+	unix.Path = uprobeUsdt.file.Name()
```
olsajiri: same as for uprobes, let's keep `path`, `file` is closed by now

pkg/elf/build_id.go
```diff
@@ -0,0 +1,81 @@
...
+		if note.Typ == 3 &&
+			note.Namesz == 4 &&
+			bytes.Equal(name, []byte{'G', 'N', 'U', 0}) &&
+			note.Descsz > 0 {
+			return desc, true
+		}
+	}
+}
+
+func ParseBuildId(file *os.File) ([]byte, error) {
```
olsajiri: would it make sense to have this as another method of `SafeELFFile` ?

pkg/sensors/tracing/genericuprobe.go
olsajiri: +1 this can be in a single function

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -505,6 +576,30 @@ func createGenericUprobeSensor(
...
+				if err := verifyFileDigest(entryFile, digest, digestCache); err == nil {
+					matchFound = true
+					break
+				}
+			}
+			if !matchFound {
+				// Build a summary of actual hashes computed during verification
+				var actualHashes []string
+				for algo, hash := range digestCache {
+					actualHashes = append(actualHashes, fmt.Sprintf("%s:%s", algo, hash))
```
olsajiri: do we need to print out all these hash details ? maybe that could be just Debug log in the `verifyFileDigest` 

also would it make sense to have `tetra` command that displays those hashes for given binary?

pkg/sensors/sensors.go
```diff
@@ -113,6 +116,15 @@ type SensorIface interface {
 	// the sensor's programs.
 	TotalMemlock() uint64
 	Overhead() ([]ProgOverhead, bool)
+	SetStatus(policyStatus *tetragon.TracingPolicyStatus)
```
olsajiri, line 124: looks like generic feature, please split that into separate commit with changelog that explains that and then add the uprobe related changes

andrewstrohman:
Thanks all for your feedback. Given the request to break this up into multiple PRs, I'll address these comments in this PR first, then close this PR, and finally open several smaller PRs.

andrewstrohman:
> For what it's worth, I do agree with Fede that it might be helpful to split this into multiple PRs.

Ack, I'll break this up into multiple PRs.

> And one more note: maybe it makes sense to remove the changes that reject the policy if we end up not doing it? 

Given the feedback to break this into multiple PRs, I think it makes sense to keep the intermediate state where we reject the policy if any hook's digest does not match. But if you'd prefer to just log the mismatch and not load the probe until we get to the final PR (which exposes the hook status via our API), that works for me too. Just let me know.

> Note also that the k8s description seems to be out of date:

>	// +kubebuilder:validation:Optional
	// Digests of the traced binary. If any digest matches, verification passes.
	BinaryDigests []string `json:"binaryDigests,omitempty"`
	
Why does this seem out of date? This is meant to describe that users can specify multiple digests, and only one needs to match in order for the hook to load.

pkg/sensors/load.go
andrewstrohman: Ok, I added the following comment:
```
andy@builder:~/src/tetragon$ git diff
diff --git a/pkg/sensors/load.go b/pkg/sensors/load.go
index 0e0d19686..128e11be2 100644
--- a/pkg/sensors/load.go
+++ b/pkg/sensors/load.go
@@ -229,6 +229,8 @@ func (s *Sensor) Unload(unpin bool) error {
 func (s *Sensor) Destroy(unpin bool) error {
        var errs error
 
+       // If a sensor fails to load, it will be destroyed immediately, in order to clean up its associated resources.
+       // If/when the sensor's policy is removed, this function will be called again when its collection is destroyed.
        if s.Destroyed {
                return nil
        }
```

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Sounds good. I moved this logic out into its own helper function.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: sounds good. done.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Sounds good. I changed this to Info.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: This logic has been moved out into a helper function. See [here](https://github.com/cilium/tetragon/pull/4994#discussion_r3442014618).

api/v1/tetragon/sensors.proto
andrewstrohman: Ok, I did this.

api/v1/tetragon/sensors.proto
andrewstrohman: I did this and also added another field:
```
// the index at which this hook is configured within its section
  uint32 hook_idx = 4;
}
```
To distinguish which hook the status is for, when we have multiple occurrences of the same hook. I didn't need this previously, because I matched the status index with the tracing policy configuration index. But now that we theoretically can have multiple sensor's status in the same status array, I can no longer just use ordering to distinguish.

The API user would iterate through the statuses, and and check the section and hook_idx in order to determine which hook the status belongs to.

api/v1/tetragon/sensors.proto
andrewstrohman: I tried to address this [here](https://github.com/cilium/tetragon/pull/4994#discussion_r3456394562).

pkg/sensors/program/loader_linux.go
andrewstrohman: OK, done.

pkg/sensors/program/loader_linux.go
andrewstrohman: No, because then we could link a different inode than what was checked.

I'm trying to avoid a TOCTOU problem that occurs where a new file, with a new inode is installed at the target path after the digest check is done, but before the path is linked. By keeping the file open and using it's FD instead of it's path, this cannot happen.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: > iiuc we close the file descriptor after we load the sensor, so this file in here is closed, right?

Yes, that's right.

> I think we should keep the uprobeEntry.path for that, and just add uprobeEntry.file

When one calls .Name() on a closed *os.File, it still returns the path that was opened. So there is no need to store the path separately.

pkg/sensors/tracing/genericusdt.go
andrewstrohman: See above response. There is no problem with calling `Name()` on a closed file.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Ok, I changed this to use the absolute path. The absolute path is not unique path because of symlinks. If your intention was to make the path unique, we should use `filepath.EvalSymlinks` too.

pkg/elf/build_id.go
andrewstrohman: ok, I did this.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: When I asked users how they would like to be notified about digest failures, they did say they want to see the digest failures in the log in addition to having a way to query the hook status. So, I kept a more brief `Info` log of the digest failure, and put the hash details in a `Debug` level log.

> also would it make sense to have tetra command that displays those hashes for given binary?

I've created a follow-up issue for this [here](https://github.com/cilium/tetragon/issues/5180). Please take a look and add your thoughts to it.

pkg/sensors/sensors.go
andrewstrohman: Ok, done.

pkg/sensors/handler.go
andrewstrohman: Ok, done.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: The reason that I initially avoided doing this is because it requires that `addUprobe` returns all the ids in the error case. I have now changed `addUprobe` to do this in order to simplify like you guys are asking for.

I wasn't able to totally eliminate the defer in `addUprobe`, but I was able to limit it to only doing CleanupKernelSelectorState, for the case where we bail before the first uprobe entry is created.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I did this. See [here](https://github.com/cilium/tetragon/pull/4994#discussion_r3464858633) for more details.

pkg/sensors/tracing/genericusdt.go
andrewstrohman: I did this. See [here](https://github.com/cilium/tetragon/pull/4994#discussion_r3464858633) for more details.

andrewstrohman:
I've addressed all the feedback. Please hold your follow-up feedback for the 4 PRs that I will break this one into. Since most of these smaller PRs depend on the the previous commits, this will have to be a serial process -- in most cases a PR with subsequent commits will not be authored until its dependencies are merged.

In each of these smaller PRs, I will link this one to give the reviewers the larger context of where I'm heading. I will continue working on this brach as I address feedback in the smaller PRs in order to test the changes, but I won't merge this one.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -424,11 +424,34 @@ type uprobeHas struct {
...
+	var errs error
+
+	for _, id := range ids {
+		uprobeEntry, err := genericUprobeTableGet(id)
+		if err != nil {
+			errs = errors.Join(errs, err)
+			continue
+		}
+
+		if err = selectors.CleanupKernelSelectorState(uprobeEntry.loadArgs.selectors.entry); err != nil {
```
olsajiri, line 645: I think we  need to cleanup return selectors as well

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -491,33 +522,26 @@ func createGenericUprobeSensor(
...
 	var argRetprobe *v1alpha1.KProbeArg
 	var argRetprobeIdx int
 	var setRetprobe bool
+	var err error
+	var uprobeSelectorState *selectors.KernelSelectorState
+	var uprobeRetSelectorState *selectors.KernelSelectorState
+
+	defer func() {
+		if retErr != nil && uprobeSelectorState != nil {
+			if cleanupErr := selectors.CleanupKernelSelectorState(uprobeSelectorState); cleanupErr != nil {
```
olsajiri: I think we need to cleanup return selector as well

olsajiri (commented):
I put your cleanup code on top of addUprobe refactor #5191, it made it bit easier
I'm still not sure the cleanup is correct for IDs sharing the selectors

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: `CleanupKernelSelectorState` only cleans up state related to MatchBinaries and MatchWorkLoadIDs. Since these are not set for return selectors, I made a conscious decision to not call `CleanupKernelSelectorState` for the return selectors.

But, I don't think it hurts to call `CleanupKernelSelectorState` on the return selector state, even though we don't currently need that. It seems like a good future-proofing measure, so I'll update my change to do this.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: See above response to the same suggestion. I did this for future-proofing, but it's not currently needed.

andrewstrohman:
> I put your cleanup code on top of addUprobe refactor #5191, it made it bit easier 

Thanks. I took a look at your PR. The code is definitely better with smaller functions like that. But, I think I did find an issue with the newly introduced `added`. I left a comment in that PR explaining why I think we can have a leak because of it.

> I'm still not sure the cleanup is correct for IDs sharing the selectors

I interpret this to mean that you're worried about trying to cleanup the same selector state multiple times. This is a pre-existing issue, because the destroy hook would have the same problem. I see that if that happens, we will get some error messages when we shouldn't. So this aspect does need to be fixed.

Here is an example:
```
kind: TracingPolicy
metadata:
  name: "uprobe-selectors"
spec:
  uprobes:
    - path: "/home/andy/src/sleep"
      symbols:
      - "main"
      - "do_sleep"
      selectors:
      - matchBinaries:
        - operator: "In"
          values:
          - "/usr/bin/dash"
          followChildren: true

```
When I use this policy with a mainline build, I get:
> level=warn msg="Destroy hook failed" sensor=generic_uprobe error="failed to lookup mbset map: lookup: key does not exist"

When I delete the policy.

Currently, every hook type, except for uprobe, instantiates selector state per-entry. uprobe is special because we can have multiple entries per hook.

With USDT, we could theoretically re-use the selector state for the case where one entry (provider/name) has multiple matches in the binary. I'm not sure if we meant to allow the user to define multiple USDT tracepoints with the same name and provider. Since the configuration parsing logic gives me the sense that it's our intention to allow that, I tried it out and found a small bug that arises from that configuration. I added a small fix for this to the series. Please let me know if we want to support multiple identical provider/name or not. If not, I'll remove the commit that removes the error message  and break out of the loop in `func addUsdt` after the first target match has occurred.

I think we should decide if we want entries to share selector state when we have multiple entries per hook. I think the answer to this is "yes". If so, and we want usdt to support multiple tracepoints with the same name/provider (this causes multple entries per hook), then we should change usdt to re-use the same selector state, like how is done for uprobe. Regardless of whether we want to support multiple of the same provider/name in usdt, we need to make sure that we don't try to re-delete the shared selector state. I've modified my approach to do this.

Since this PR will not be merged, and instead https://github.com/cilium/tetragon/pull/5182 will be merged, let's continue this discussion over there.

kkourt:
@andrewstrohman can you please add all the children PRs in the description of this PR?

PR #4995, opened 2026-05-13
observer: add YARA rule-based process execution scanning

### LINKED ISSUE #1319: Yara rule support for detection and blocking file executions

_No response_

Hi tetragon team,

I was wondering if you had considered adding Yara rule (https://yara.readthedocs.io/en/stable/writingrules.html) support to tetragon. I feel like being able to check a file against a list of Yara rules at execution would be pretty useful.
There are a few Yara parsers for go around, but I have no idea how hard this would be to implement. Just throwing this out there to see what you think.

Regards,
Aidan

_No response_

### Code of Conduct

4ykutG:
Fixes #1319

Description
Adds YARA rule-based scanning for process execution events. When a new process starts, its binary is scanned asynchronously against YARA rules. On a match, SIGKILL is sent to terminate the process.

pkg/observer/yara_scanner.go: YARA engine with 4-worker pool, buffered queue (256), per-worker Scanner, fail-safe drop on full queue
pkg/observer/observer.go: whitelist filter + async scan on execve events
pkg/option, pkg/defaults: --yara-rules-dir flag (default /etc/tetragon/yara-rules.d)
cmd/tetragon/main.go: engine initialization at startup (non-fatal if directory missing)
yara-rules/ransomware.yar: detection rules for WannaCry, LockBit, REvil, shadow copy deletion, ransom note names, encryption+mass-write patterns
Known limitation: userspace SIGKILL has a TOCTOU race — the binary starts executing before we can kill it. A proper solution would use eBPF LSM hooks for pre-execution blocking, which could be a follow-up.

Dependency: requires libyara >= 4.2 on the host system.

Changelog

Add YARA rule-based process execution scanning via --yara-rules-dir flag.


Dockerfile
mtardy, line 1: why did you remove the whole Dockerfile?

mtardy (changes requested):
Thanks for the proposal. Could you speak a bit about why you think this is the
right way to solve the problem?

Additionally, if you are using Generative AI tools to create this contribution,
please describe how involved you were using the [AI Influence Level], and
describe how you reviewed the output before submission.

[AI Influence Level]: https://danielmiessler.com/blog/ai-influence-level-ail

Could you also in the future split your commits into logical changes, please read https://tetragon.io/docs/contribution-guide/ before resubmitting. Thanks!

PR #4996, opened 2026-05-13
Add a method to remove a single pod from the fake watcher

mstemm:
This is useful for frameworks utilizing FakeK8sWatcher that need to remove individual pods.

### Description
Add a method to the fake watcher `watcher.RemovePod` that can be used to remove a single previously added pod. This is useful for frameworks using fake watcher that need to change the set of pods.

### Changelog

```
Add a method `(watcher *FakeK8sWatcher) RemovePod` that can be used to remove a previously added pod. This is useful for frameworks using fake watcher that need to change the set of pods.
```

### DIFF
--- a/pkg/watcher/fake.go
+++ b/pkg/watcher/fake.go
@@ -50,6 +50,16 @@ func (watcher *FakeK8sWatcher) AddPod(pod *corev1.Pod) {
 	watcher.pods = append(watcher.pods, pod)
 }
 
+func (watcher *FakeK8sWatcher) RemovePod(rem *corev1.Pod) {
+	for i, p := range watcher.pods {
+		pod := p.(*corev1.Pod)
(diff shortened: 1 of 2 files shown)


pkg/watcher/fake.go
```diff
@@ -50,6 +50,15 @@ func (watcher *FakeK8sWatcher) AddPod(pod *corev1.Pod) {
 	watcher.pods = append(watcher.pods, pod)
 }
 
+func (watcher *FakeK8sWatcher) RemovePod(pod *corev1.Pod) {
+	for i, p := range watcher.pods {
+		if p.(*corev1.Pod).Name == pod.Name {
```
tpapagian: Does it make sense to compare the Namespace as well?

mstemm: Oh sure, fixed.

tpapagian (approved):

mtardy (approved):

PR #4998, opened 2026-05-13
chore(pkg/sensors): emit a warning when `monitor_only` mode is enforced.

### LINKED ISSUE #4997: Policy should be rejected if enforcement mode is requested but not supported

A policy that is valid only for Monitor mode is accepted when added with Protect/Enforce mode. Instead of rejecting the request, Tetragon silently installs the policy in monitor_only mode. A later mode change to enforce is correctly rejected. The initial add operation should also reject the policy when Protect/Enforce mode is requested.

```
root@ubuntu2404:/tmp/tetragon-policies# tetra tp add test.yaml -m enforce
tracing policy "test.yaml" added
root@ubuntu2404:/tmp/tetragon-policies# tetra tp list
ID   NAME                               STATE     FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE           NPOST   NENFORCE   NMONITOR
63   test   enabled   0          (global)    generic_kprobe   693.10 kB      monitor_only   0       0          0
```

### Tetragon Version

Latest

### Kernel Version

6.19.0

### Kubernetes Version

_No response_

### Bugtool

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

FedeDP:
Fixes #4997

### DIFF
--- a/pkg/sensors/tracing/policyhandler_linux.go
+++ b/pkg/sensors/tracing/policyhandler_linux.go
@@ -92,6 +92,7 @@ func newPolicyInfoFromSpec(
 
 	// If enforcement is not allowed, force monitor only
 	if !hasEnforcementActions(spec) {
+		logger.GetLogger().Warn("Enforcement actions are not configured for this policy, forcing monitor_only mode", "policy", name)
 		opts.policyMode = policyconf.MonitorOnlyMode
 	}
 


tpapagian (approved):

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -92,6 +92,7 @@ func newPolicyInfoFromSpec(
 
 	// If enforcement is not allowed, force monitor only
 	if !hasEnforcementActions(spec) {
+		logger.GetLogger().Warn("Enforcement actions are not configured for this policy, forcing monitor_only mode")
```
olsajiri: nit maybe we could print policyname so people do not need to search around in the log?

olsajiri (approved):
nit maybe you

pkg/sensors/tracing/policyhandler_linux.go
FedeDP: Sure, makes sense!

PR #5001, opened 2026-05-13
cmd/tetragon-tester: improve sysrq-trigger shutdown failure log

JoonsooWon:
While reading the `cmd/tetragon-tester` code, I noticed two small issues in the error log printed

1. The binary name was misspelled as `syseq-trigger` instead of `sysrq-trigger`.
2. The captured error value was not included in the log, which makes debugging harder

This PR fixes the typo and includes the error using `%v`


mtardy (approved):
sure thanks, did you see the error in some logs?

JoonsooWon:
Thanks
No actual logs, I was reading through the code to learn about tetragon and eBPF, and I noticed it.

FedeDP (approved):

PR #5002, opened 2026-05-13
fix for multiple uprobes in one policy with different targets

andrewstrohman:
This is meant to fix an issue regarding multiple probes within a single, multi-probe enabled uprobe policy. If the probes are for different paths, then the policy will not load.

This issue was caused by the fact that we did not use a unique path when pinning the bpf links.

### DIFF
--- a/pkg/sensors/program/loader_linux.go
+++ b/pkg/sensors/program/loader_linux.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/cilium/ebpf"
@@ -369,6 +370,7 @@ func uprobeAttachMulti(load *Program, prog *ebpf.Program, spec *ebpf.ProgramSpec
 		var links []link.Link
 		var lnk link.Link
 
+		idx := 0
 		for path, attach := range data.Attach {
 			exec, err := link.OpenExecutable(path)
 			if err != nil {
@@ -388,12 +390,14 @@ func uprobeAttachMulti(load *Program, prog *ebpf.Program, spec *ebpf.ProgramSpec
 			if err != nil {
 				return nil, err
 			}
-			err = LinkPin(lnk, bpfDir, load, extra...)
+			pinExtra := append(extra, strconv.Itoa(idx))
+			err = LinkPin(lnk, bpfDir, load, pinExtra...)
 			if err != nil {
 				lnk.Close()
 				return nil, err
 			}
 			links = append(links, lnk)
+			idx++
 		}
 		return links, nil
 	}
--- a/tests/policytests/uprobes.go
+++ b/tests/policytests/uprobes.go
@@ -264,3 +264,34 @@ spec:
 		EventChecker: ec.NewUnorderedEventChecker(checker),
 	}
 }).RegisterAtInit()
+
+var _ = policytest.NewBuilder("uprobe-multiple-targets").WithLabels("uprobes").WithPolicyTemplate(`
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "uprobe-multiple-targets"
+spec:
+  uprobes:
+  - path: {{ testBinary "uprobe-resolve" }}
+    symbols:
+    - "func"
+  - path: {{ testBinary "nop" }}
+    symbols:
+    - "main"
+`).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
+	binOne := c.TestBinary("uprobe-resolve")
+	binTwo := c.TestBinary("nop")
+	up1Checker := ec.NewProcessUprobeChecker(binOne).
+		WithProcess(ec.NewProcessChecker().
+			WithBinary(sm.Full(binOne))).WithSymbol(sm.Full("func"))
+
+	up2Checker := ec.NewProcessUprobeChecker(binTwo).
+		WithProcess(ec.NewProcessChecker().
+			WithBinary(sm.Full(binTwo))).WithSymbol(sm.Full("main"))
+
+	return &policytest.Scenario{
+		Name:         "check both events occur",
+		Trigger:      policytest.NewMultiCmdTrigger(map[string][]string{binOne: {"v8", "7"}, binTwo: {}}),
+		EventChecker: ec.NewUnorderedEventChecker(up1Checker, up2Checker),
+	}
+}).RegisterAtInit()
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -524,13 +524,22 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 	addrs := len(spec.Addrs)
 	refCtrOffsets := len(spec.RefCtrOffsets)
 
-	// uprobe definition spec usanity checks
-	if symbols == 0 && offsets == 0 && addrs == 0 {
-		return nil, errors.New("uprobe need either Symbols, Offsets or Addrs defined")
+	// uprobe definition spec sanity check
+	numAddressMethods := 0
+	if symbols != 0 {
+		numAddressMethods++
 	}
-	if symbols != 0 && offsets != 0 && addrs != 0 {
-		return nil, errors.New("uprobe is defined either only with Symbols, Offsets or Addrs")
+	if offsets != 0 {
+		numAddressMethods++
 	}
+	if addrs != 0 {
+		numAddressMethods++
+	}
+
+	if numAddressMethods != 1 {
+		return nil, errors.New("uprobe needs exactly one of either Symbols, Offsets or Addrs defined")
+	}
+
 	if refCtrOffsets != 0 {
 		if symbols != 0 && symbols != refCtrOffsets {
 			return nil, fmt.Errorf("RefCtrOffsets(%d) has different dimension than Symbols(%d)",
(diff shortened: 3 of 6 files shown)


FedeDP:
I agree that 2. feels really weird and we'd better not support it; after all this (taken from the newly added test):
```
uprobes:
  - path: {{ testBinary "uprobe-resolve" }}
    symbols:
    - "func"
  - path: {{ testBinary "uprobe-resolve" }}
    addrs:
    - 0x1140 # addr of main in uprobe-resolve
```
is a way to overcome this check https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/genericuprobe.go#L531; ie this:
```
uprobes:
  - path: {{ testBinary "uprobe-resolve" }}
    symbols:
    - "func"
    addrs:
    - 0x1140 # addr of main in uprobe-resolve
```
is already rejected (even if the check is a bit stricter, checking that all of {symbols,addrs,offsets} are not empty; i am wondering whether the if condition is actually the desired one).

My 2c: i'd kill 2. support by erroring out, and i'd fix the aforementioned check that IMHO is too strict, we should just check that at least 2 of them are defined.

pkg/sensors/program/program.go
```diff
@@ -102,7 +103,7 @@ type MultiUprobeAttachSymbolsCookies struct {
 
 type MultiUprobeAttachData struct {
 	// Path -> []{Symbol,Cookie}
```
olsajiri, line 104: so the comment makes no longer sense

tests/policytests/uprobes.go
```diff
@@ -221,3 +221,106 @@ spec:
...
+metadata:
+  name: "uprobe-multiple-targets-same-path"
+spec:
+  uprobes:
+  - path: {{ testBinary "uprobe-resolve" }}
+    symbols:
+    - "func"
+  - path: {{ testBinary "uprobe-resolve" }}
+    addrs:
+    - 0x1140 # addr of main in uprobe-resolve
```
olsajiri: so that will probably change for each build?

pkg/sensors/program/loader_linux.go
```diff
@@ -388,7 +389,8 @@ func uprobeAttachMulti(load *Program, prog *ebpf.Program, spec *ebpf.ProgramSpec
 			if err != nil {
 				return nil, err
```
olsajiri, line 391: unrelated to your change, but I wonder we do proper cleanup in here in case some targets already succeeded.. looks like we should clean it up before return

tests/policytests/uprobes.go
```diff
@@ -221,3 +221,106 @@ spec:
...
+		WithProcess(ec.NewProcessChecker().
+			WithBinary(sm.Full(bin)))
+
+	expected_count := uint64(2)
+
+	return &policytest.Scenario{
+		Name:         "check both events occur",
+		Trigger:      policytest.NewCmdTrigger(bin, "v8", "7").ExpectExitCode(0),
+		EventChecker: ec.NewUnorderedEventChecker(upChecker),
+		ActCountChecker: policytest.ActionCounts{
```
olsajiri: is this supported in observer test? can't see it `DoObserverTest`

olsajiri (commented):
> In particular, I'd like feedback about 2. I think it's unlikely that a user would use a mixture of symbols vs addrs. It's unfortunate that the current version of the fix does not attempt to combine regardless if there is a mix or not.

I think it makes sense to support that.. unless user specify all uprobes one by one with with same path and cancel the purpose of multi uprobe ;-) but I guess you can have just offsets for some probes and symbols for others

> I'm wondering if we really want to support this use case. And if so, should I combine the multiple probes into a single link (like how is done prior to this change) where the target path is the same and the address specification method (symbols vs addrs) is the same.

ok, then you'd need to convert symbols to offsets right? we'd be little bit faster with just single uprobe_multi link.. maybe it could make the code easier?

andrewstrohman:
> is already rejected (even if the check is a bit stricter, checking that all of {symbols,addrs,offsets} are not empty; i am wondering whether the if condition is actually the desired one).
> 
> My 2c: i'd kill 2. support by erroring out, and i'd fix the aforementioned check that IMHO is too strict, we should just check that at least 2 of them are defined.


Ultimately, the loader gets three values passed to it for each probe's link:
1) a symbol. 
2) an address. This is the a substitute for symbol. if it's defined, the symbol is ignored. Confusingly, the address is not the virtual address -- it's the offset within the elf file.
3) an offset. This is the offset from the symbol's address, (or the address, if it's defined).

From the loader API perspective, we need to pass either a symbol or an address(of a symbol). Regardless of which is passed, we can pass an optional offset from that symbol. As noted above, if we pass both a symbol and an address, then the symbol is ignored.

Things get a little confusing because of conflicting meanings of address and offset. In our policy spec, address means virtual address and offset means offset within the elf. For the loader, address means the offset within the elf file and offset means an offset from the symbol address.

If symbols are provided in the spec, then we split that string, ie "main+0x7" into a loader symbol and loader offset. If the target binary is pure go, then the offset of the symbol is determined, and we use that.
If offsets are provided in the spec, we use them as loader addresses, without supplying a loader offset. This makes sense given the mapping I described above.
If addresses are provided in the spec, we calculate the elf file offset from the address and use this as loader addresses without offset.

So if users specify an address via addrs or offsets, they need to include the offset within that address. When addrs or offsets are supplied in the spec, there is no way for the user to specify an offset based on these addresses. 

So, given this, I think it makes sense that we require to have exactly one of symbols, offsets, and addresses in the spec file. They all individually describe enough information to determine the address to attach the probe to.

But if you meant that we should allow combining offsets and addrs, then maybe that makes sense too, in that  addrs are converted to offsets. But if we go down that road, you could say the same about pure go binaries.

I think that in order to make the explanation simple, we should either allow all combinations of addrs/offset/symbols or no combination of them.

tests/policytests/uprobes.go
andrewstrohman: It's not. This check is only done if you run the policy test directly, ie `./tetra policytest run uprobe-multiple-targets-same-path-x86`.  So when we run this via the shim for our existing unit tests, it won't fail if the counts are wrong.

We do have a couple other preexisting policy tests that utilize this check, with one of them using the shim for unit tests. So this wouldn't be the first use of this check mechanism that won't work as a unit test.

With that being said, since the current failure mode is that the policy won't load, the unit test can still be useful. This is why I felt it was OK to do. But I agree this is not ideal.

It seems we have three options:
1) Accept this shortcoming, and leave it as-is because it can still detect policy load errors.
2) Make the observer test mode check this. This probably the best option, but requires some additional effort now.
3) Convert the test to be a legacy unit test, where it's easy to count the number of posts.

pkg/sensors/program/loader_linux.go
andrewstrohman: I see what you're saying and I agree this lack of cleanup is a problem.

tests/policytests/uprobes.go
andrewstrohman: Yes, this is a good point. I did overlook this.

Perhaps, I should a create new simple tester-prog, that has a comment that its code should never be updated. That would prevent problems due to code change. But it wouldn't prevent problems when the compiler evolves and generates different binaries for the same code.

I'm not sure how to get around this problem without checking the binary into the repo, which I don't think we want to do.

andrewstrohman:
> > In particular, I'd like feedback about 2. I think it's unlikely that a user would use a mixture of symbols vs addrs. It's unfortunate that the current version of the fix does not attempt to combine regardless if there is a mix or not.
> 
> I think it makes sense to support that.. unless user specify all uprobes one by one with with same path and cancel the purpose of multi uprobe ;-) but I guess you can have just offsets for some probes and symbols for others

Sorry, but I don't fully understand your point here. It seems like you do want to prevent this conflict from occurring, in some scenarios, but not others. Can you please give me concrete example of what configuration should be allowed, and what configuration should be disallowed?

It seems you are trying to make the distinction between multiple probes in a (multi-uprobe enabled )uprobe policy vs a single uprobe that targets multiple symbols/addrs/offets. The former is the scenario that I'm trying to address here. The latter is not currently a problem because we don't allow them to configure more than one of symbols, offsets and addrs. 

For the former, Federico is suggesting (in this PR) that I just reject the policies that do this. Do you think it's better to just reject those configs, rather than trying to cope?

> > I'm wondering if we really want to support this use case. And if so, should I combine the multiple probes into a single link (like how is done prior to this change) where the target path is the same and the address specification method (symbols vs addrs) is the same.
> 
> ok, then you'd need to convert symbols to offsets right? we'd be little bit faster with just single uprobe_multi link.. maybe it could make the code easier?

I was initially thinking that I would combine to have at most two links for multi-attach: one for symbols and one for addrs/offsets. But what you're describing seems better: converting the symbols into addresses, so that they can all be combined.

andrewstrohman:
Thanks @FedeDP and @olsajiri for your feedback. I've decided to handle issue 2) in a follow up PR. I've created [this issue](https://github.com/cilium/tetragon/issues/5005) as a reminder and summarized all the feedback I received here in it.

FedeDP:
> So, given this, I think it makes sense that we require to have exactly one of symbols, offsets, and addresses in the spec file. They all individually describe enough information to determine the address to attach the probe to.

I 100% agree, but that's not what the code is doing there, right? 
```go
if symbols != 0 && offsets != 0 && addrs != 0 {
```
This will pass fine if eg: both `symbols` and `offsets` are not 0, since it checks that all 3 of them are `!=0`, thus it will only fail when users specifies all 3 of them.

kkourt (approved):
Thanks!

FedeDP (approved):
Changes LGTM!

andrewstrohman:
> I 100% agree, but that's not what the code is doing there, right?
> 
> ```go
> if symbols != 0 && offsets != 0 && addrs != 0 {
> ```
> 
> This will pass fine if eg: both `symbols` and `offsets` are not 0, since it checks that all 3 of them are `!=0`, thus it will only fail when users specifies all 3 of them.

Ok, I understand now. Yes, you're right. I'll add a commit to address this.

FedeDP (approved):

PR #5003, opened 2026-05-14
chore: fix duplicated "an" in process.go GetPodInfo doc comment

lphuc2250gma:
Removes a doubled `an an` from the doc comment for `GetPodInfo` in `pkg/process/process.go`. Comment-only, no behavior change.


FedeDP (approved):
Thanks!!

PR #5013, opened 2026-05-17
helm: Disable gops debug interface by default

peoyekunle:
The gops listener is unauthenticated and reachable by any hostNetwork pod on the same node, exposing runtime internals without authentication.  It can be re-enabled during install.

Reported-by: Banh Do <<bangdo.dxb2@gmail.com>>

### DIFF
--- a/install/kubernetes/tetragon/values.yaml
+++ b/install/kubernetes/tetragon/values.yaml
@@ -253,7 +253,7 @@ tetragon:
         extraIpAddresses: []
   gops:
     # -- Whether to enable exposing gops server.
-    enabled: true
+    enabled: false
     # -- The address at which to expose gops.
     address: "localhost"
     # -- The port at which to expose gops.
--- a/Makefile
+++ b/Makefile
@@ -341,7 +341,7 @@ e2e-test: image image-operator
 else
 e2e-test:
 endif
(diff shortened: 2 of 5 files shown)


mtardy (commented):
Thanks, please take a look at the CI errs, you might need to generate some files.

```sh
make -C install/kubernetes
```

mtardy (commented):
Sorry I missed it in the first review.

1. can you add why the change in the commit description
2. can you add a note in there (in another commit if you want) for people that are going to upgrade to know about this change https://github.com/cilium/tetragon/blob/main/contrib/upgrade-notes/latest.md.

mtardy:
image
when ready

install/kubernetes/tetragon/values.yaml
```diff
@@ -188,7 +188,7 @@ tetragon:
     address: "unix:///var/run/tetragon/tetragon.sock"
   gops:
     # -- Whether to enable exposing gops server.
-    enabled: true
+    enabled: false
```
olsajiri, line 256: curious what's the reason for that, please elaborate in the commit log as requested by @mtardy

mtardy: I think this is because of a security concern, would be cool to have it explicitly stated in the commit log.

peoyekunle: Added this to the commit message -  "The gops listener is unauthenticated and reachable by any
hostNetwork pod on the same node, exposing runtime internals
without authentication."

mtardy (approved):
lgtm thanks

mtardy:
@peoyekunle can you re-force push (just `commit --amend` your last commit and repush it without any modification), I think GitHub was ahving troubles when you last pushed and thus the CI didn't run

peoyekunle:
All good now @mtardy

PR #5018, opened 2026-05-19
bpf: simplify copy_char_buf

3u13r:
Thanks to @andrewstrohman for [finding this issue](https://github.com/cilium/tetragon/pull/4989#discussion_r3262416749).

### Description
This function has been largely untouched since the first Tetragon commit:
https://github.com/cilium/tetragon/blob/931b70f2c9878ba985ba6b589827bea17da6ec33/bpf/process/types/basic.h#L439

We used the `probe_read()` function to the value of meta, which in turn is read from the `msg_generic_krobe *e` value which we read from a eBPF map in `__read_arg_2`. Therefore, we can access the value directly.


mtardy (approved):
thanks, indeed seems the value is coming from a map anway.

PR #5021, opened 2026-05-19
tetragon: factor register override parsing

olsajiri:
change register assignment parsign to use small cursor parsing
instead of fmt.Sscanf which can't handle trailing input properly

### DIFF
--- a/pkg/asm/assignment_linux.go
+++ b/pkg/asm/assignment_linux.go
@@ -129,24 +129,40 @@ func parseReg(str string, ass *Assignment) error {
 	return nil
 }
 
+// parseConst parses constant assignment forms such as "1" and "-1".
+// Constants keep base-0 parsing, so 0x and leading-zero forms are accepted.
+// Examples: "1", "-1", "0x20", "010".
 func parseConst(str string, ass *Assignment) error {
-
-	var (
-		uoff uint64
-		soff int64
-		err  error
-	)
-
-	if uoff, err = strconv.ParseUint(str, 0, 64); err != nil {
-		if soff, err = strconv.ParseInt(str, 0, 64); err != nil {
-			return errNext
-		}
-		uoff = uint64(soff)
+	uoff, err := parseOffset(strings.TrimSpace(str))
+	if err != nil {
+		return errNext
 	}
 
 	ass.Type = ASM_ASSIGNMENT_TYPE_CONST
 	ass.Off = uoff
 	return nil
+}
+
+// parseOffset parses decimal, hex, octal, signed, and full-width unsigned
+// offsets. It intentionally uses strconv base 0, so offsets accept "0x20",
+// "010", "-1", and "0xffffffffffffffff" forms.
+func parseOffset(str string) (uint64, error) {
+	if str == "" {
+		return 0, strconv.ErrSyntax
+	}
+
+	// Parse negative numbers as int64 first because ParseUint rejects the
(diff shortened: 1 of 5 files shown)


andrewstrohman (approved):
Thanks, LGTM.

pkg/asm/assignment_linux.go
```diff
@@ -129,26 +129,57 @@ func parseReg(str string, ass *Assignment) error {
...
+	// int64.
+	off, err := strconv.ParseUint(str, 0, 64)
+	if err == nil {
+		return off, nil
+	}
+
+	// Fall back to signed parsing for base-0 forms that ParseUint rejected but
+	// ParseInt can still represent. Return the original unsigned error if both
+	// parsers fail; it describes the attempted uint64 result.
+	soff, serr := strconv.ParseInt(str, 0, 64)
```
kkourt: What is a value that ParseUint will reject, but ParseInit will parse at this point (given that we 've already parsed negative numbers above)?

pkg/asm/assignment_linux.go
```diff
@@ -89,26 +89,53 @@ func parseRegDeref(str string, ass *Assignment) error {
...
 
-	if n, _ = fmt.Sscanf(str, "0x%x%%%s", &off, &reg); n != 2 {
-		if n, _ = fmt.Sscanf(str, "%d%%%s", &off, reg.Reset()); n != 2 {
-			return errNext
-		}
+	p := cursorparser.New(str)
+
+	// Split "off%reg" at the percent marker. There must be a non-empty
+	// offset before it and a non-empty register name after it.
+	offStr, ok := p.ReadUntil('%')
```
kkourt: In a subsequent patch, you use `strings.Cut()`, couldn't we use `strings.Cut('%')` here as well?

pkg/asm/assignment_linux.go
```diff
@@ -36,56 +35,62 @@ type Assignment struct {
 
 type fn func(str string, ass *Assignment) error
 
-type RegScanner struct {
-	name string
-}
+// parseRegDeref parses dereference forms "(%reg)" and "off(%reg)".
+// Examples: "(%rsp)", "0x20(%rsp)", "0x20 ( %rsp )".
+func parseRegDeref(str string, ass *Assignment) error {
```
kkourt, line 40: nit: would be nice to have tests for the Examples.

pkg/asm/assignment_linux.go
olsajiri: hum, no value.. the ParseUInt is enough, thanks

pkg/asm/assignment_linux.go
olsajiri: yep, sohuld make easier code, thanks

kkourt (approved):

olsajiri:
@kkourt I noticed we could remove the bit below as well, otherwise there's no change

```
diff --git a/pkg/asm/assignment_linux.go b/pkg/asm/assignment_linux.go
index f8e33d8d1b28..30653527dd5b 100644
--- a/pkg/asm/assignment_linux.go
+++ b/pkg/asm/assignment_linux.go
@@ -133,10 +133,6 @@ func parseRegOff(str string, ass *Assignment) error {
 // parseReg parses register assignment forms "%reg".
 // Example: "%rax".
 func parseReg(str string, ass *Assignment) error {
-       var (
-               ok bool
-       )
-
        p := cursorparser.New(str)

        // Register assignments must start with a percent marker.

```

PR #5024, opened 2026-05-19
renovate: use installTools for golang instead of manual install

### LINKED ISSUE #5023: renovate: refactor install tool command for post-upgrade tasks
renovate added a better option to install tools in post upgrade tasks: https://docs.renovatebot.com/configuration-options/#postupgradetasksinstalltools

We can replace old install tool commands:
https://github.com/cilium/tetragon/blob/d800078bbfe9b005705f14c9c95ffa76052e4a4a/.github/renovate.json5#L19

dangome3:
### Description

Replace manual golang installation commands with Renovate's built-in installTools configuration.

Fixes https://github.com/cilium/tetragon/issues/5023

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -16,7 +16,6 @@
     "^make vendor$",
     "^make -C install/kubernetes$",
     "^go mod vendor$",
-    "^install-tool golang \\$\\(grep -oP '\\^toolchain go\\\\K\\.\\+\\$' go.mod \\|\\| grep -oP '\\^go \\\\K\\.\\+\\$' go.mod\\)$",
     "^make gen-docs-references$",
   ],
   // repository configuration
@@ -185,7 +184,7 @@
         "gomodVendor",
         "gomodUpdateImportPaths",
       ],
-      postUpgradeTasks: {
+      "postUpgradeTasks": {
         "commands": ["/tmp/install-buildx", "make vendor", "make protogen", "make crds", "make gen-docs-references"],
         "fileFilters": ["**/**"],
         "executionMode": "branch"
@@ -213,9 +212,12 @@
         // We need to trigger a golang install manually here because in some
         // cases it might not be preinstalled, see:
         // https://github.com/renovatebot/renovate/discussions/23485
-        "commands": ["install-tool golang $(grep -oP '^toolchain go\\K.+$' go.mod || grep -oP '^go \\K.+$' go.mod)", "make vendor", "make gen-docs-references"],
+        "commands": ["make vendor", "make gen-docs-references"],
         "fileFilters": ["**/**"],
-        "executionMode": "branch"
+        "executionMode": "branch",
+        "installTools": {
+          "golang": {}
+        }
       },
     },
     {
@@ -234,9 +236,12 @@
         // We need to trigger a golang install manually here because in some
         // cases it might not be preinstalled, see:
         // https://github.com/renovatebot/renovate/discussions/23485
-        "commands": ["install-tool golang $(grep -oP '^toolchain go\\K.+$' go.mod || grep -oP '^go \\K.+$' go.mod)", "make vendor", "make gen-docs-references"],
+        "commands": ["make vendor", "make gen-docs-references"],
         "fileFilters": ["**/**"],
-        "executionMode": "branch"
+        "executionMode": "branch",
+        "installTools": {
+          "golang": {}
+        }
       },
     },
     {


dangome3:
cc: @mtardy

.github/renovate.json5
```diff
@@ -213,9 +212,12 @@
         // We need to trigger a golang install manually here because in some
         // cases it might not be preinstalled, see:
         // https://github.com/renovatebot/renovate/discussions/23485
-        "commands": ["install-tool golang $(grep -oP '^toolchain go\\K.+$' go.mod || grep -oP '^go \\K.+$' go.mod)", "make vendor", "make gen-docs-references"],
+        "commands": ["make vendor", "make gen-docs-references"],
         "fileFilters": ["**/**"],
-        "executionMode": "branch"
+        "executionMode": "branch",
+        "installTools": {
+          "golang": {}
```
sayboras, line 219: is the automatically detecting golang version as what we have with install-tool ?

mtardy: Yep that's a good question, I think it might use their install-tool thing internally so maybe it's worth investigating if this script for go will not install the version specified in go.mod now that Go can do such things. Might be worth taking a look at that script.

https://github.com/renovatebot/renovate/blob/c6304b78d359961b7aa95bb03b45c8f309471264/lib/workers/repository/update/branch/execute-post-upgrade-commands.ts#L148-L174

dangome3: Hey! Thanks both for commenting

Yes, as @mtardy said, renovate uses internal `install-tool` for the `containerbase/base` image. That is pointed in the docs https://docs.renovatebot.com/configuration-options/#postupgradetasksinstalltools in which you can specify the version to install, and if not it installs the latest. See https://github.com/containerbase/base/blob/main/docs/tools.md#install

I have been researching how renovate uses this tool and found that it passes an argument which is `constrains` to the install tool command. See [here](https://github.com/renovatebot/renovate/blob/818b0197df031709e2b56ef2522f4cdda606a5c7/lib/util/exec/containerbase.ts#L377-L390)

```node
export async function generateInstallCommands(
  toolConstraints: Opt<ToolConstraint[]>,
): Promise<string[]> {
  ...
      const toolVersion = await resolveConstraint(toolConstraint);
      const { toolName } = toolConstraint;
      const installCommand = `install-tool ${toolName} ${quote(toolVersion)}`;
      installCommands.push(installCommand);
    }
  }
  return installCommands;
}
```

I also found that that for golang it extracts the constrains [here](https://github.com/renovatebot/renovate/blob/4929140eef60c3d6444141860dc294ccd30c44ee/lib/modules/manager/gomod/extract.ts#L85-L107)

```node
  const toolchainDirective = packageFile.deps.find(
    (dep) =>
      dep.depName === 'go' &&
      dep.depType === 'toolchain' &&
      dep.datasource === 'golang-version',
  );
  if (toolchainDirective?.currentValue) {
    packageFile.extractedConstraints ??= {};
    packageFile.extractedConstraints.golang = toolchainDirective.currentValue;
```

Which are loaded to the `upgrade.constrains` attribute [here](https://github.com/renovatebot/renovate/blob/818b0197df031709e2b56ef2522f4cdda606a5c7/lib/workers/repository/process/fetch.ts#L124-L129)

The main highlight IMO is that It uses toolchain directive. If any case we could use it to make sure it installs the same go version as in `go.mod`

I think we can use something like this:

```diff
diff --git a/go.mod b/go.mod
index a73316c93..4f919d510 100644
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,8 @@ module github.com/cilium/tetragon
 // renovate: datasource=golang-version depName=go
 go 1.26.0

+toolchain go1.26.0
+
 require (
        github.com/alecthomas/kong v1.15.0
        github.com/cilium/ebpf v0.21.1-0.20260501093402-9f87aafaeb37
```

and we could use `renovate` to bump that when `go` version is updated too.

Otherwise I think it will install the latests version, not sure if that would be an issue. 

What do you think? should we add that constrain?

mtardy: I think we want to avoid the toolchain thing https://github.com/cilium/tetragon/pull/1748, let's try maybe it just works

sayboras: ah thanks for your detailed explanation, installing latest versions for main branch is fine to me.

sayboras (approved):
👍

.github/renovate.json5
```diff
@@ -185,7 +184,7 @@
         "gomodVendor",
         "gomodUpdateImportPaths",
       ],
-      postUpgradeTasks: {
+      "postUpgradeTasks": {
```
mtardy, line 187: nit: ideally I would have shipped into a different commit but it's okay

mtardy (approved):
Thanks, let's see if it works, we can always revert if it doesn't :)

PR #5025, opened 2026-05-19
helm: Adjust mounting of socket to survive daemon restarts for consistent metadata collection

gregswift:
### Description

If/when a containerd or cri daemon restarts without completely restarting the node it breaks the mounted socket into the daemonset. This happens because linux bind mounts on inode, not the path itself.

When the disconnect happens, metadata is no longer collected to be submitted.

This implements the similar socket mount setup as cilium-agent for spire-agent and envoy.

* [volumes](https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml#L931-L943)
* [volumeMounts](https://github.com/cilium/cilium/blob/main/install/kubernetes/cilium/templates/cilium-agent/daemonset.yaml#L312-L321)

Worth noting:
1. Any orgs running Kyverno/OPA/Gatekeeper rules that allow hostPath mounts by explicit path (/run/containerd/containerd.sock) need their allowlist updated to the directory path.

2. `type: Socket` → `type: Directory` removes a kubelet-side integrity check. With `Socket`, kubelet refuses to start the pod unless the file is a unix socket at that exact path. With `Directory`, kubelet only checks that a directory exists.
   a. For CRI socket, this seems the safer choice because it would be better to fail loudly if the runtime path is wrong rather than silently creating an empty directory and not collecting metadata which leads to other failure scenarios.
   b. Security-wise this is a small reduction in defense-in-depth (you could conceivably bind-mount a directory whose containerd.sock was a regular file pointing somewhere unexpected, which requires host root already).

3. Mounting the directory does open access such as an easy path log/stdio scraping if tetragon itself is compromised, however, as procfs is already mounted the access already exists.

### Changelog

```release-note
helm: Adjust mounting of socket to survive daemon restarts for consistent metadata collection
```

### DIFF
--- a/contrib/upgrade-notes/latest.md
+++ b/contrib/upgrade-notes/latest.md
@@ -11,7 +11,10 @@ Depending on your setup, changes listed here might require a manual intervention
 
 ### Helm Values
 
-* TBD
+* CRI users: Mounting of the Container Runtime socket is moving to a directory path.
+  If your systems are running Kyverno/OPA/Gatekeeper rules that allow hostPath mounts
+  only by explicit paths, then you will need to extend your allowlist with the
+  directory containing the socket passed to `tetragon.cri.socketHostPath`.
 
 ### TracingPolicy (k8s CRD)
 
--- a/install/kubernetes/tetragon/templates/_container_tetragon.tpl
+++ b/install/kubernetes/tetragon/templates/_container_tetragon.tpl
@@ -41,7 +41,7 @@
     - mountPath: "/procRoot"
       name: host-proc
 {{- if and (.Values.tetragon.cri.enabled) (.Values.tetragon.cri.socketHostPath) }}
-    - mountPath: {{ quote .Values.tetragon.cri.socketHostPath }}
+    - mountPath: {{ dir .Values.tetragon.cri.socketHostPath | quote }}
       name: cri-socket
 {{- end }}
 {{- range .Values.extraHostPathMounts }}
--- a/install/kubernetes/tetragon/templates/daemonset.yaml
+++ b/install/kubernetes/tetragon/templates/daemonset.yaml
@@ -101,8 +101,8 @@ spec:
 {{- if and (.Values.tetragon.cri.enabled) (.Values.tetragon.cri.socketHostPath) }}
       - name: cri-socket
         hostPath:
-          path: {{ quote .Values.tetragon.cri.socketHostPath }}
-          type: Socket
+          path: {{ dir .Values.tetragon.cri.socketHostPath | quote }}
+          type: Directory
 {{- end }}
       {{- with .Values.extraVolumes }}
         {{- toYaml . | nindent 6 }}


USERS.md
```diff
@@ -76,12 +76,17 @@ Example entry:
...
     * N: Stream Security
       D: Stream Security uses Tetragon for Network, Process, and File observability and protection in Kubernetes clusters
       U: Network monitoring, Process monitoring, File access monitoring
       L: https://www.stream.security/
       Q: @vitali-streamsec
+
+    * N: Xata
+      D: Xata uses Tetragon to identify security risks and policy violations
+      U: Runtime security
+      L: https://xata.io
```
sayboras: Thanks for adding USERS.md details.

We normally split this into dedicated PR.

sayboras (commented):
Understand that it's still in draft, so I didn't spend much time on main changes, just one small comment on USERS.md update.

USERS.md
gregswift: ahh. the pr template called it out so I included it 👍

mtardy: yeah or at least separate patch, also if you can avoid touching the other lines, or do the cleanup in another separate patch, that would be perfect :)

gregswift: I removed those from this for now. can add as a separate PR.

When you say separate patch, you mean i can do 2 commits on another PR, one to remove the trailing whitespace, another to add the new user?

mtardy: ok cool would appreciate a PR for USERS.md, always nice to have some data like that. 

> When you say separate patch, you mean i can do 2 commits on another PR, one to remove the trailing whitespace, another to add the new user?

yeah that would be ideal

contrib/upgrade-notes/latest.md
```diff
@@ -11,7 +11,10 @@ Depending on your setup, changes listed here might require a manual intervention
 
 ### Helm Values
 
-* TBD
+* Mounting of the Container Runtime socket is moving to a directory path. If your
+  systems are running Kyverno/OPA/Gatekeeper rules that allow hostPath mounts only
+  by explicit paths, then you will need to extend your allowlist with the directory
+  containing the socket passed to `tetragon.cri.socketHostPath`.
```
mtardy: maybe we can be a bit more explicit that this affects you only if you use the `cri` feature at the beginning of that paragraph.

mtardy (commented):
Thanks a lot for the PR, patch is clean with a clear explanation of the problem and the benefits/downsides, that's nice :)!

I'll ping @kkourt since I think he added this.

mtardy:
Ah also you need to sign off the commit https://github.com/cilium/tetragon/actions/runs/26164259092/job/77104552577?pr=5025.

The Tetragon project uses a "sign-off" procedure which you can learn more about in the [Cilium project documentation](https://docs.cilium.io/en/latest/contributing/development/contributing_guide/#developer-s-certificate-of-origin).

<p>

```shell
```

```shell
git commit -s
```

```shell
git commit --amend -s
```

```
-s, --signoff, --no-signoff

   option on the command line.
```

</p>
</details>

kkourt:
> I'll ping @kkourt since I think he added this.

Thanks for the detailed description in the PR! The approach makes sense to me.

I wondered if mounting the dir is  enough (instead of the scoket) and whether  we woudl also need to have a facility to reconnect to the socket if it goes away, but I guess grpc.NewClient does this for us: https://pkg.go.dev/google.golang.org/grpc#ClientConn

USERS.md
gregswift: #5056 submitted

gregswift:
I think I've caught everything as asked 👍

sayboras (approved):
Thanks a lot for your contribution 💯

mtardy (approved):
I'll let @kkourt merge this he thinks it's good

kkourt (approved):
Thanks!

PR #5028, opened 2026-05-20
docs: Add fentry/fexit sensor documentation

sayboras:
### Description
I was giving it a try with fentry/fexit feature, and notice that docs/example is not there.

### Changelog

```release-note
docs: Add fentry/fexit sensor documentation
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -4,10 +4,11 @@ weight: 2
 description: "Hook points for Tracing Policies and arguments description"
 ---
 
-Tetragon can hook into the kernel using `kprobes` and `tracepoints`, as well as in user-space
-programs using `uprobes`. Users can configure these hook points using the correspodning sections of
-the `TracingPolicy` specification (`.spec`). These hook points include arguments and return values
-that can be specified using the `args` and `returnArg` fields as detailed in the following sections.
+Tetragon can hook into the kernel using `kprobes`, `fentry/fexit`, and `tracepoints`, as well as in
+user-space programs using `uprobes`. Users can configure these hook points using the corresponding
+sections of the `TracingPolicy` specification (`.spec`). These hook points include arguments and
+return values that can be specified using the `args` and `returnArg` fields as detailed in the
+following sections.
 
 {{< warning >}}
 Hooking a system call can introduce time-of-check to time-of-use (TOCTOU)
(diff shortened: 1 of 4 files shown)


tpapagian (approved):

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -89,6 +90,107 @@ spec:
     # [...]
 ```
 
+## Fentry/Fexit
+
+Fentry (BPF fentry/fexit) programs attach to kernel functions using BTF (BPF Type Format). They
+are a modern alternative to kprobes with lower overhead. Instead of using breakpoints and
+instruction patching like kprobes, fentry programs are called directly from a trampoline, which
+makes them faster and safer.
```
mtardy, line 98:
```suggestion
Fentry/fexit programs attach to kernel functions using BTF (BPF Type Format).
They are a modern alternative to kprobes with lower overhead. Instead of using
breakpoints and instruction patching like kprobes, fentry programs rely on the
ftrace kernel infrastructure which calls the BPF progs directly from a
trampoline making them faster and safer.
```

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -89,6 +90,107 @@ spec:
...
+- No BTF support: fentry requires the kernel to be built with `CONFIG_DEBUG_INFO_BTF=y`.
+- Enforcement is needed: fentry does not yet support enforcement (override) actions.
+
+### Kernel requirements
+
+To use fentry, the following kernel requirements must be met:
+
+- Kernel version: >= 6.1 is required for Tetragon's fentry implementation.
+- BTF support: the kernel must be built with BTF enabled (`CONFIG_DEBUG_INFO_BTF=y`). You can
+  verify this with:
```
mtardy, line 120: I feel like we kinda repeat ourselves at a few lines of distance

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -89,6 +90,107 @@ spec:
...
+The output should include:
+
+```
+CONFIG_DEBUG_INFO_BTF=y
+```
+
+### Limitations
+
+- No enforcement/override support: unlike kprobes, fentry does not yet support enforcement
+  actions such as `Override` or `Signal`.
```
mtardy, line 135: This as well is repeated

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -89,6 +90,107 @@ spec:
...
+      index: 0
+      type: "int"
+```
+
+{{< note >}}
+Unlike kprobes where `syscall` can be set to `true` to trace system call entry points, fentry is
+typically used for internal kernel functions and `syscall` should be set to `false`. This is
+because fentry attaches via BTF-based trampolines that are suited for tracing internal kernel
+function boundaries rather than the syscall ABI layer.
+{{< /note >}}
```
mtardy, line 166: question: can you trace syscall with fentry? it's not clear to me when reading this

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -89,6 +90,107 @@ spec:
...
+
+{{< note >}}
+The `fentry` program attaches at function *entry*, while its `fexit` counterpart attaches at
+function *exit*, where the return value is available. To capture a function's return value, set
```
mtardy, line 170: https://kubernetes.io/docs/contribute/style/style-guide/#use-italics-to-define-or-introduce-new-terms
```suggestion
The `fentry` program attaches at function entry, while its `fexit` counterpart attaches at
function exit, where the return value is available. To capture a function's return value, set
```

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -89,6 +90,107 @@ spec:
...
+`return: true` — this is what attaches the `fexit` program — and define `returnArg` with the
+return value's `type`, as shown in the example above. When `return` is left unset (it defaults to
+`false`), only the `fentry` program is attached and any `returnArg` is ignored. See
+[Return values](#return-values) for more details on `return` and `returnArg`.
+{{< /note >}}
```
mtardy, line 175: My opinion but I feel like it could be a normal paragraph, looks like important information to explain how to do fexit with the tracingpolicy!
```suggestion
The `fentry` program attaches at function *entry*, while its `fexit` counterpart attaches at
function *exit*, where the return value is available. To capture a function's return value, set
`return: true` — this is what attaches the `fexit` program — and define `returnArg` with the
return value's `type`, as shown in the example above. When `return` is left unset (it defaults to
`false`), only the `fentry` program is attached and any `returnArg` is ignored. See
[Return values](#return-values) for more details on `return` and `returnArg`.
```

mtardy (commented):
Thanks a lot for keeping the docs up to date :)!! Here are a few comments, I feel like @olsajiri should take a look as he introduced the code changes.

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -89,6 +90,107 @@ spec:
     # [...]
 ```
 
+## Fentry/Fexit
+
+Fentry (BPF fentry/fexit) programs attach to kernel functions using BTF (BPF Type Format). They
+are a modern alternative to kprobes with lower overhead. Instead of using breakpoints and
```
olsajiri, line 96: let's just state trampolines are faster alternative to kprobes, the breakpoint part is bit more difficult and not really accurate, also safer is arguable

docs/content/en/docs/concepts/tracing-policy/hooks.md
olsajiri: yes, the `__x64_sys_*` are traceable and need special processing to get the syscall arguments,
so you need to use the syscall flag for fentry sensor as well

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -89,6 +90,107 @@ spec:
...
+instruction patching like kprobes, fentry programs are called directly from a trampoline, which
+makes them faster and safer.
+
+### When to use fentry vs kprobes
+
+Fentry is preferred over kprobes when the environment supports it, because:
+
+- Lower overhead: no breakpoint or instruction patching is needed; the BPF program is invoked
+  directly from a trampoline.
+- Safer: avoids the complexity and potential issues of runtime instruction patching.
```
olsajiri, line 106: arguable.. kprobes on function entries use ftrace in a similar way like trampolines,
maybe you mean the kprobe attach complexity when it's attached in the middle of the function
on specific instruction, than it's tricky.. but still safe

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -89,6 +90,107 @@ spec:
...
+
+{{< note >}}
+Unlike kprobes where `syscall` can be set to `true` to trace system call entry points, fentry is
+typically used for internal kernel functions and `syscall` should be set to `false`. This is
+because fentry attaches via BTF-based trampolines that are suited for tracing internal kernel
+function boundaries rather than the syscall ABI layer.
+{{< /note >}}
+
+{{< note >}}
+The `fentry` program attaches at function *entry*, while its `fexit` counterpart attaches at
```
kevsecurity, line 169: Maybe I'm missing it, but I can't see a corresponding explanation for kretprobes; I think we just silently hide the kretprobe functionality behind capturing the return value in a kprobe. Therefore, I question whether we need this discussion about fexit here, as the same could apply to fentry?

mtardy (approved):
Thanks Tam!

olsajiri (approved):
thanks

PR #5029, opened 2026-05-20
pkg/bugtool: bug fixes to `tetra debug maps` output

mtardy:
See patches.

```release-note
Fixed incorrect total memlock reporting for aggregated maps in Bugtool and tetra debug maps, where only a single instance’s memlock usage was previously shown.
```


FedeDP (approved):
Cool, thank you very much!

kkourt (approved):
Thanks!
I think it's worth backporting this to the latest release if it's not too complicated. WDYT?

mtardy:
> Thanks! I think it's worth backporting this to the latest release if it's not too complicated. WDYT?

yeah, I'll do that

PR #5033, opened 2026-05-20
new: move exportDirectory to non tmpfs folder

FedeDP:
### Description

Since `tetragon.log` could grow indefinitely, we'd better avoid using a tmpfs-backed location for it.

### Changelog

```release-note
new: move exportDirectory to non tmpfs folder
```

### DIFF
--- a/contrib/upgrade-notes/latest.md
+++ b/contrib/upgrade-notes/latest.md
@@ -17,6 +17,8 @@ Depending on your setup, changes listed here might require a manual intervention
   directory containing the socket passed to `tetragon.cri.socketHostPath`.
 * `tetragon.gops.enabled` now defaults to `false`. To re-enable the gops debug
   interface, set `--set tetragon.gops.enabled=true` at install or upgrade time.
+* `exportDirectory` default value has been updated to `/var/log/tetragon` from 
+  `/var/run/cilium/tetragon/` to avoid writing to tmpfs.
 
 ### TracingPolicy (k8s CRD)
 
--- a/docs/content/en/docs/concepts/events.md
+++ b/docs/content/en/docs/concepts/events.md
@@ -114,7 +114,7 @@ in a Kubernetes cluster  `process_exec.process.pod`. The binary and args being e
 the event here `process_exec.process.binary` and `process_exec.process.args`. Finally, a `node_name`
 and `time` provide the location and time for the event and will be present in all event types.
 
-A default deployment writes the JSON log to `/var/run/cilium/tetragon/tetragon.log` where it can
+A default deployment writes the JSON log to `/var/log/tetragon/tetragon.log` where it can
 be exported through normal log collection tooling, e.g. fluentd, logstash, etc. The file will
 be rotated and compressed by default. See [Helm Options] for details on how to customize this location.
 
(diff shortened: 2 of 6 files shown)


sayboras (approved):

mtardy (commented):
Do we also need to touch:

https://github.com/cilium/tetragon/blob/71eaaaba91d79dfbd9147aa318c8ef782207f3e5/install/kubernetes/tetragon/templates/daemonset.yaml#L78-L81

and

https://github.com/cilium/tetragon/blob/71eaaaba91d79dfbd9147aa318c8ef782207f3e5/install/kubernetes/tetragon/templates/_container_tetragon.tpl#L35-L36

or are they for different purpose?

Also should we add the symlink? I'm wondering if it's okay for a new release or if we should do the whole deprecation dance, add the symlink for one release and then remove.

FedeDP:
Nope, the volume we are referring to is:
```
      - name: export-logs
        hostPath:
          path: {{ .Values.exportDirectory }}
          type: DirectoryOrCreate
``` 
that is already templated to `exportDirectory`.
The `/var/run/cilium/` folder remains for some sockets etc etc, we only moved the export files here.

> Also should we add the symlink? I'm wondering if it's okay for a new release or if we should do the whole deprecation dance, add the symlink for one release and then remove.

I am a bit unsure about this one; i am open to do that, it's not a big problem. At the same time, i'd do that only if we really need to!

FedeDP:
Well a
```
$ grep -R "/var/run/cilium" *
grep: examples/configuration/README.md: File o directory non esistente
install/kubernetes/tetragon/templates/_container_tetragon.tpl:    - mountPath: "/var/run/cilium"
install/kubernetes/tetragon/templates/daemonset.yaml:          path: /var/run/cilium
```
only found these 2 instances indeed. Perhaps we can kill the volume? But since it was there, i don't think we should touch it (at least, not in this PR!). WDYT?

mtardy:
> Well a
> 
> ```
> $ grep -R "/var/run/cilium" *
> grep: examples/configuration/README.md: File o directory non esistente
> install/kubernetes/tetragon/templates/_container_tetragon.tpl:    - mountPath: "/var/run/cilium"
> install/kubernetes/tetragon/templates/daemonset.yaml:          path: /var/run/cilium
> ```
> 
> only found these 2 instances indeed. Perhaps we can kill the volume? But since it was there, i don't think we should touch it (at least, not in this PR!). WDYT?

no no it's fine if it's used for sockets or whatever other things that are virtually 0 byte or small

kkourt (commented):
Thanks!

Can you please add some information on the commit message about the motivation behind this change?

mtardy:
> > Also should we add the symlink? I'm wondering if it's okay for a new release or if we should do the whole deprecation dance, add the symlink for one release and then remove.
> 
> I am a bit unsure about this one; i am open to do that, it's not a big problem. At the same time, i'd do that only if we really need to!

It all depends if we consider this a breaking change, if yes then it's better to do the whole deprecation thing, let's see other people's opinion.

FedeDP:
@kkourt done!

contrib/upgrade-notes/latest.md
```diff
@@ -11,7 +11,7 @@ Depending on your setup, changes listed here might require a manual intervention
 
 ### Helm Values
 
-* TBD
+* `exportDirectory` default value has been updated to `/var/log/tetragon` to avoid writing to tmpfs.
```
mtardy: maybe tell in which directory it was before in this small note

mtardy (approved):
lgtm! would wait on discussing at meetings/or here if we need to do the deprecation thing before merging

kkourt (approved):

contrib/upgrade-notes/latest.md
FedeDP: Right!

kkourt:
> I'm wondering if it's okay for a new release or if we should do the whole deprecation dance, add the symlink for one release and then remove.

In my mind, we don't have to since the user can just change the default to the old value.

FedeDP:
I agree with @kkourt !
@mtardy wdyt?

FedeDP:
I think we can go on with this; we can always revert it later if needed.

PR #5034, opened 2026-05-20
bpf: use void *ctx for fentry/fexit programs

sayboras:
### Description

fentry/fexit BPF programs don't receive a struct pt_regs * context — their context is an array of function arguments, not register state. This updates the generic fentry/fexit programs to use void
   *ctx, which more accurately reflects the actual context type:

- bpf_generic_fentry.c: change generic_fentry_event signature from struct pt_regs *ctx to void *ctx.
- bpf_generic_fexit.c: change the fexit_calls prog-array values type from int(struct pt_regs *) to int(void *) to match.

No functional change — the context pointer is passed through to generic_start_process_filter unchanged.

### Changelog

```release-note
bpf: use void *ctx for fentry/fexit programs
```

### DIFF
--- a/bpf/process/bpf_generic_fentry.c
+++ b/bpf/process/bpf_generic_fentry.c
@@ -50,7 +50,7 @@ struct {
 #define SECTION_TAIL  "fentry/generic_fentry_tail"
 
 __attribute__((section((SECTION_ENTRY)), used)) int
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
Thanks!

olsajiri (approved):
oops ;-) thanks

PR #5036, opened 2026-05-21
ci: get chart version before second checkout

3u13r:
### Description
Move the chart version calculation to before the second checkout.


sayboras (approved):
Oh nice 💯

Issue #4747, opened 2026-03-09
docs: make `matchData` use cases more discoverable


mtardy:
One idea would be to add some kind of notes around the existing selectors, saying that you can extend those by using the generic matchData https://tetragon.io/docs/concepts/tracing-policy/selectors/#data-filter.

And provide more examples in matchData like the one reproducing matchUID.

> I found a solution following [#1943 (comment)](https://github.com/cilium/tetragon/issues/1943#issuecomment-3571923641)
> 
> It works properly with
> 
>     data:
>     - type: "int"
>       index: 0
>       source: "current_task"
>       resolve: "cred.uid.val"
>     selectors:
>       matchData:
>       - index: 0
>         operator: "GreaterThan"
>         values:
>           - "1000"
> I hope it could help I hope this can help someone looking for the same thing as us.
> 
> I close the issue. Regards.
> 
> great! :) 

 _Originally posted by @mtardy in [#4455](https://github.com/cilium/tetragon/issues/4455#issuecomment-3686106533)_

JoonsooWon:
Hello @mtardy.
I'm learning eBPF and Tetragon and want to tackle this as part of my learning.
Could you assign this issue to me?

PR #5039, opened 2026-05-21
docs: expand matchData section with examples and operator reference

JoonsooWon:
related issue: https://github.com/cilium/tetragon/issues/4747

### Description

Expand the `Data filter` section. (`selectors.md`)
Added examples:
- Filtering for UIDs at an LSM hook
- Matching specific process names

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -186,16 +186,27 @@ selectors:
 Data filters can be specified under the `matchData` field and provide
 filtering based on the value of the specified `data` field.
 
-You can specify the argument either in the `index` or in `args` field. Both `index` and
-`args` field denote the argument position within the spec file.
+The `data` block allows you to filter events based on fields of
+kernel data structures rather than function arguments — for example,
+the UID of the current task at a hook that does not take credentials
+as an argument. Each entry in the `data` block specifies:
 
-In the following example we extra pid value from `current_task` and filter
-on all values except for `1`.
+- `source`: where to read from. Supported values are `current_task`
+  (the process's `task_struct`) and `pt_regs` (the register state at
+  the hook point).
+- `resolve`: which field of the source structure to read.
+- `type`: how the resolved value should be interpreted.
+
+A `matchData` selector then refers to a `data` entry by its `index`
+and applies an operator.
+
+In the following example we extract the `pid` value from `current_task`
+and filter on all values except for `1`.
 
 ```yaml
 data:
 - index: 0
-  type: "string"
+  type: "int"
   source: "current_task"
   resolve: "pid"
 selectors:
@@ -205,6 +216,52 @@ selectors:
     values:
     - "1"
 ```
+
+Nested fields can be accessed through the `.` separator, as described
+in [Attribute resolution]({{< ref "/docs/concepts/tracing-policy/hooks#attribute-resolution" >}}).
+In the following example we extract the UID of the current task via
+`cred.uid.val` and filter for non-system users (UID greater than
+`1000`), hooked at `security_bprm_committed_creds` — an LSM hook
+called after a process's credentials are committed.
+
+```yaml
+kprobes:
+- call: "security_bprm_committed_creds"
+  syscall: false
+  data:
+  - index: 0
+    type: "int"
+    source: "current_task"
+    resolve: "cred.uid.val"
+  selectors:
+  - matchData:
+    - index: 0
+      operator: "GreaterThan"
+      values:
+      - "1000"
+```
+
+The same pattern works for non-integer types. The following example
+reads the process name from `comm` (a string field of `task_struct`)
+and matches when it equals a specific value.
+
+```yaml
+data:
+- index: 0
+  type: "string"
+  source: "current_task"
+  resolve: "comm"
+selectors:
+- matchData:
+  - index: 0
+    operator: "Equal"
+    values:
+    - "cat"
+```
+
+The available operators for `matchData` are the same as those listed under
+[Arguments filter](#arguments-filter), and apply according to the `type`
+declared in the corresponding `data` entry.
 
 ## Return args filter
 


docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -186,16 +186,22 @@ selectors:
 Data filters can be specified under the `matchData` field and provide
 filtering based on the value of the specified `data` field.
 
-You can specify the argument either in the `index` or in `args` field. Both `index` and
-`args` field denote the argument position within the spec file.
+The `data` block is used when the value you want to filter on is not exposed as a function argument. For example, the UID of the current task at a hook that does not take credentials as an argument. Each entry in the `data` block specifies:
```
mtardy: So I don't think that's the right way to phrase this. "The `data` block is used when the value you want to filter on is not exposed as a function argument.".

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -186,16 +186,22 @@ selectors:
...
 
-In the following example we extra pid value from `current_task` and filter
-on all values except for `1`.
+- `source`: where to read from. The typical value is `current_task`, which refers to the `task_struct` of the process executing the hooked code path
+
+- `resolve`: which field of the source structure to read. Nested fields can be addressed using `.` separators like `cred.uid.val`
+
+- `type`: how the resolved value should be interpreted
```
mtardy: do not put empty lines between list items

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -186,16 +186,22 @@ selectors:
 Data filters can be specified under the `matchData` field and provide
 filtering based on the value of the specified `data` field.
 
-You can specify the argument either in the `index` or in `args` field. Both `index` and
-`args` field denote the argument position within the spec file.
+The `data` block is used when the value you want to filter on is not exposed as a function argument. For example, the UID of the current task at a hook that does not take credentials as an argument. Each entry in the `data` block specifies:
 
-In the following example we extra pid value from `current_task` and filter
-on all values except for `1`.
+- `source`: where to read from. The typical value is `current_task`, which refers to the `task_struct` of the process executing the hooked code path
```
mtardy: I wouldn't say the "typical value".

Do you know what are the other things you can read from data?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -206,6 +212,43 @@ selectors:
     - "1"
 ```
 
+Nested fields can be accessed through the `.` separator. In the following example we extract the UID of the current task via `cred.uid.val` and filter for non-system users (UID greater than `1000`), hooked at `security_bprm_committed_creds` — an LSM hook called after a process's credentials are committed.
```
mtardy: you can point to the docs of resolving nested fields here

mtardy (changes requested):
thanks, I think that's a nice addition but this needs just a bit more work, some of the phrasing should be changed.

Can you wrap the lines like the rest of the file?

Thank you!

mtardy:
Maybe @olsajiri you can help here as well since you added this.

docs/content/en/docs/concepts/tracing-policy/selectors.md
JoonsooWon: Thanks!
I've fixed it

docs/content/en/docs/concepts/tracing-policy/selectors.md
JoonsooWon: I've add a link to the [Attribute resolution](https://tetragon.io/docs/concepts/tracing-policy/hooks/#attribute-resolution) section.

docs/content/en/docs/concepts/tracing-policy/selectors.md
JoonsooWon: Thanks!
I updated the phrasing to emphasize its primary purpose.

[changed version](https://github.com/cilium/tetragon/pull/5039/changes#diff-37a73e4a96ceb7cf01d09a92d49a12534ccf8f26638c5442e3c59508e838a381R189)

docs/content/en/docs/concepts/tracing-policy/selectors.md
JoonsooWon: I noticed `pt_regs` earlier, but left it out because I wasn't sure how it was being used in this specific path.
I appreciate you pointing it out.

[changed version](https://github.com/cilium/tetragon/pull/5039/changes#diff-37a73e4a96ceb7cf01d09a92d49a12534ccf8f26638c5442e3c59508e838a381R194)

JoonsooWon:
Thanks for the feedback.
I've addressed everything and updated the PR.
Would you mind reviewing again when you have time?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -186,16 +186,28 @@ selectors:
...
+The `data` block allows you to filter events based on fields of
+kernel data structures rather than function arguments — for example,
+the UID of the current task at a hook that does not take credentials
+as an argument. Each entry in the `data` block specifies:
 
-In the following example we extra pid value from `current_task` and filter
-on all values except for `1`.
+- `source`: where to read from. Supported values are `current_task`
+  (the process's `task_struct`) and `pt_regs` (the register state at
+  the hook point).
```
mtardy: Let's just make sure this is the only supported values if we say that, otherwise we can propose those values maybe not say it's the implicitly "only" supported values, like we "for example" or something like that.

Maybe you can take a look at the code to have a bit more info on that :)?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -206,6 +218,50 @@ selectors:
...
+  resolve: "comm"
+selectors:
+- matchData:
+  - index: 0
+    operator: "Equal"
+    values:
+    - "cat"
+```
+
+The available operators for `matchData` are the same as those listed under [Arguments filter](#arguments-filter), and apply according to the `type` declared in the corresponding `data` entry.
```
mtardy: please wrap this also

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -186,16 +186,28 @@ selectors:
...
-on all values except for `1`.
+- `source`: where to read from. Supported values are `current_task`
+  (the process's `task_struct`) and `pt_regs` (the register state at
+  the hook point).
+- `resolve`: which field of the source structure to read.
+- `type`: how the resolved value should be interpreted.
+
+A `matchData` selector then refers to a `data` entry by its `index`
+and applies an operator. This makes it possible to filter on values
+that are not exposed by dedicated selectors.
```
mtardy: > This makes it possible to filter on values that are not exposed by dedicated selectors

I don't think the word selector is correct. matchData is a selector. Maybe drop that sentence?

mtardy (commented):
Thanks a lot for the work, this is better, still a few comments :)!

docs/content/en/docs/concepts/tracing-policy/selectors.md
JoonsooWon: Sorry about that.
I've fixed this.

[changed](https://github.com/cilium/tetragon/pull/5039/changes#diff-37a73e4a96ceb7cf01d09a92d49a12534ccf8f26638c5442e3c59508e838a381R262-R264)

docs/content/en/docs/concepts/tracing-policy/selectors.md
JoonsooWon: I've dropped this sentence as you suggested.
Thanks!

[changed](https://github.com/cilium/tetragon/pull/5039/changes#diff-37a73e4a96ceb7cf01d09a92d49a12534ccf8f26638c5442e3c59508e838a381R200-R201)

docs/content/en/docs/concepts/tracing-policy/selectors.md
JoonsooWon: Thanks for the feedback.
I've checked the code one more.

Two helpers are simple string equality checks
```go
func hasCurrentTaskSource(arg *v1alpha1.KProbeArg) bool {
	return arg.Source == "current_task"
}

func hasPtRegsSource(arg *v1alpha1.KProbeArg) bool {
	return arg.Source == "pt_regs"
}
```
in `genericuprobe.go` and `generickprobe.go`
https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/genericuprobe.go#L725-L727
https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/generickprobe.go#L857-L859

Currently, only these two values pass validation.
I've used 'for example' to keep it extensible, as you suggested.

[changed](https://github.com/cilium/tetragon/pull/5039/changes#diff-37a73e4a96ceb7cf01d09a92d49a12534ccf8f26638c5442e3c59508e838a381R194-R196)

JoonsooWon:
Thanks!
I've addressed all the comments and updated the PR.
Could you review the changes again?

docs/content/en/docs/concepts/tracing-policy/selectors.md
mtardy: ah you were right, so I think the previous formulation was more correct then sorry!

mtardy (commented):
almost there I'm sorry can you just revert the change that you checked then we can merge this :)

docs/content/en/docs/concepts/tracing-policy/selectors.md
JoonsooWon: No worries!
I've reverted it :)

JoonsooWon:
Thanks for the review :)
I've fixed all
Could you review the changes again?

mtardy (approved):
Thanks!

PR #5046, opened 2026-05-22
pkg/selectors: fix Equals selector for dentry

### LINKED ISSUE #5041: `type: dentry` does not support being used in `matchArgs`

The following policy works 

```yaml
kind: TracingPolicy
metadata:
    name: test
spec:
    kprobes:
      - call: "security_inode_link"
        syscall: false
        args:
        - index: 0
          type: dentry
        selectors:
        - matchActions:
            - action: Post
```

This one does not

```yaml
kind: TracingPolicy
metadata:
    name: test
spec:
    kprobes:
      - call: "security_inode_link"
        syscall: false
        args:
        - index: 0
          type: dentry
        selectors:
        - matchArgs:
            - operator: Equal
              index: 0
              values:
                - /abc/def
          matchActions:
            - action: Post
```

### Tetragon Version

Latest

### Kernel Version

6.17

### Kubernetes Version

_No response_

### Bugtool

_No response_

### Relevant log output

```shell
level=warn msg="Server AddTracingPolicy request failed" error="policy handler 'tracing' failed loading policy 'test': parseMatchArgs  error: writeMatchValues error: MatchArgs type dentry unsupported" metadata.namespace="" metadata.name=test
```

### Anything else?

_No response_

3u13r:
Fixes #5041

### Description
When choosing if we want to write the selector value as value or string, we missed `GenericDentryType` in the list of types that should be written as strings.

Note that this is also documented to work, and given that we support the Equals operator for other path types, it makes sense to add support.

### Changelog

```release-note
pkg/selectors: fix Equals selector for dentry
```

### Open questions
Do we want to add tests for every combination of data type and selector?
Also note that this isn't a regression, this has worked, as far as I can tell ([PR1](https://github.com/cilium/tetragon/pull/3423), [PR2](https://github.com/cilium/tetragon/pull/3450)).

### DIFF
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -1000,7 +1000,7 @@ func parseMatchArg(k *KernelSelectorState, arg *v1alpha1.ArgSelector, sig []v1al
 		}
 	case SelectorOpEQ, SelectorOpNEQ:
 		switch ty {
-		case gt.GenericFdType, gt.GenericFileType, gt.GenericPathType, gt.GenericStringType, gt.GenericCharBuffer, gt.GenericLinuxBinprmType, gt.GenericDataLoc, gt.GenericNetDev, gt.GenericSockaddrUnType:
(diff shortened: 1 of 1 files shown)


kkourt (approved):
Thanks! I think we need to backport this, so I added the respective labels.

PR #5058, opened 2026-05-25
sensors: fix LoadedMapsInfo data race in TotalMemlock

### LINKED ISSUE #5037: Agent panics with `concurrent map iteration and map write`
## What happened? How can we reproduce this?

The Tetragon agent panics and the `tetragon` container is restarted by the kubelet when a TracingPolicy is applied/deleted in a tight loop while a client (e.g. `tetra tracingpolicy list`) polls policy status concurrently.

The crash originates in `Sensor.TotalMemlock`, which copies a per-program `LoadedMapsInfo` map without holding any lock. The policy load/unload path mutates that same map at the same time, which the Go runtime detects and aborts the process.

### Root cause

`Sensor.TotalMemlock` (`pkg/sensors/sensors.go:152`):

```go
func (s Sensor) TotalMemlock() int {
    uniqueMap := map[int]bpf.ExtendedMapInfo{}
    for _, p := range s.Progs {
        // we could first check that it exist then write but all maps with
        // same ID on the kernel should share the same info
        maps.Copy(uniqueMap, p.LoadedMapsInfo)
    }
    ...
}
```

`p.LoadedMapsInfo` (`pkg/sensors/program/program.go:182`, `map[int]bpf.ExtendedMapInfo`) is

- written on the load path at `pkg/sensors/program/loader_linux.go:1150-1172`, and
- set to `nil` on the unload path at `pkg/sensors/program/program.go:237`,

with no synchronization for either side.

`TotalMemlock` is called from the gRPC `ListTracingPolicies` handler
(`pkg/sensors/handler.go:389`) which only holds `h.collections.mu.RLock()` —
that lock guards the collections registry, not the per-program internals.
The `maps.Copy` therefore iterates a map that the load goroutine can mutate
concurrently Go runtime fatals with `concurrent map iteration and map write`.

### Impact / severity

- Remotely-triggerable agent crash via the agent's gRPC API + a normal CR
  apply/delete pattern (CI, GitOps reconciliation, the operator itself).
- Brief observability gap on every restart; all sensors reload.
- Severity: **HIGH**.

## Tetragon Version

```text
$ tetra version
CLI version: v1.8.0-pre.0-81-g88395efcc
```

(Built from `main` at commit `88395efcc` — `e2e: add log message on tetragon installation failure`.)

## Kernel Version

```text
$ uname -a
Linux tetragon-dev-control-plane 6.18.31-1-lts #1 SMP PREEMPT_DYNAMIC Fri, 15 May 2026 15:01:51 +0000 x86_64 Linux
```

## Kubernetes Version

```text
$ kubectl version
Client Version: v1.36.0
Kustomize Version: v5.8.1
Server Version: v1.35.1
```

(kind cluster, single node `tetragon-dev-control-plane`.)

## Bugtool

Not collected for this report; the issue is a pure Go data race and is
fully described by the previous-container stack trace below. Available on
request.

## Relevant log output

```shell
fatal error: concurrent map iteration and map write

goroutine 1589 [running]:
internal/runtime/maps.fatal({0x24a96a5?, 0x48ec8f?})
	/usr/local/go/src/runtime/panic.go:1181 +0x18
internal/runtime/maps.(*Iter).Next(0x92e40a8f418?)
	/usr/local/go/src/internal/runtime/maps/table.go:819 +0x86
maps.Copy[...](...)
	/usr/local/go/src/maps/maps.go:63
github.com/cilium/tetragon/pkg/sensors.Sensor.TotalMemlock({{0x2469961, 0xe}, {0x0, 0x0}, {0x92e456cbcd0, 0x10}, {0x92e3e2d8120, 0x1a}, {0x92e43d645e0, 0x1, ...}, ...})
	/go/src/github.com/cilium/tetragon/pkg/sensors/sensors.go:157 +0x153
github.com/cilium/tetragon/pkg/sensors.(*handler).listPolicies(0x38b6?)
	/go/src/github.com/cilium/tetragon/pkg/sensors/handler.go:389 +0x483
github.com/cilium/tetragon/pkg/sensors.(*Manager).ListTracingPolicies(0x92e3e40a0a0, {0x25f5fb8?, 0x3db9640?})
	/go/src/github.com/cilium/tetragon/pkg/sensors/manager.go:160 +0x31
github.com/cilium/tetragon/pkg/server.(*Server).ListTracingPolicies(0x92e461454a0, {0x25f6108, 0x92e3fe02840}, 0x92e47104900?)
	/go/src/github.com/cilium/tetragon/pkg/server/server.go:298 +0xcd
github.com/cilium/tetragon/api/v1/tetragon._FineGuidanceSensors_ListTracingPolicies_Handler({0x23af340, 0x92e461454a0}, {0x25f6108, 0x92e3fe02840}, 0x92e414bb000, 0x0)
	/go/src/github.com/cilium/tetragon/vendor/github.com/cilium/tetragon/api/v1/tetragon/sensors_grpc.pb.go:460 +0x1a9
google.golang.org/grpc.(*Server).processUnaryRPC(0x92e3e64e008, {0x25f6108, 0x92e46eb6630}, 0x92e436d2000, 0x92e46194b40, 0x3d76128, 0x0)
	/go/src/github.com/cilium/tetragon/vendor/google.golang.org/grpc/server.go:1430 +0x11dc
google.golang.org/grpc.(*Server).handleStream(0x92e3e64e008, {0x25f6c38, 0x92e3e1f2340}, 0x92e436d2000)
	/go/src/github.com/cilium/tetragon/vendor/google.golang.org/grpc/server.go:1856 +0xbc6
google.golang.org/grpc.(*Server).serveStreams.func2.1()
	/go/src/github.com/cilium/tetragon/vendor/google.golang.org/grpc/server.go:1065 +0x7f
created by google.golang.org/grpc.(*Server).serveStreams.func2 in goroutine 1574
	/go/src/github.com/cilium/tetragon/vendor/google.golang.org/grpc/server.go:1076 +0x11d
```

sayboras:
### Description

Fixes: https://github.com/cilium/tetragon/issues/5037

### Changelog

```release-note
sensors: fix LoadedMapsInfo data race in TotalMemlock
```

### DIFF
--- a/.github/workflows/gotests.yml
+++ b/.github/workflows/gotests.yml
@@ -122,3 +122,56 @@ jobs:
       working-directory: ${{ github.workspace }}\go\src\github.com\cilium\tetragon
       run: |
         go test --failfast .\pkg\... .\cmd\...
+
+  race:
+    name: Race detector
+    runs-on: ubuntu-24.04
+    timeout-minutes: 50
+    # Advisory for now: a detected race should surface in the job logs without
+    # blocking the PR while we shake out the initial package list. Flip to
+    # false (or remove the line) once the default RACE_PKGS set is stable.
+    continue-on-error: true
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      with:
+        path: go/src/github.com/cilium/tetragon/
+
+    - name: Install Go
+      uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
+      with:
+        # renovate: datasource=golang-version depName=go
+        go-version: '1.26.3'
+
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get -y install libelf-dev netcat-traditional libcap-dev gcc libc6-dev-i386 pahole llvm
+
+        sudo sed -i '/secure_path/d' /etc/sudoers
+        sudo sed -i '/env_reset/d' /etc/sudoers
+
+    - name: Install bpftool
+      uses: mtardy/setup-bpftool@9bc044304197616a6321d452e04cc059e006d8e2 # v1.0.4
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+
+    - name: Run go tests with -race
+      env:
+        GOPATH: ${{ env.GITHUB_WORKSPACE }}/go
+        SUDO: sudo -E
+      run: |
+        go clean -cache -modcache -testcache -fuzzcache
+        cd go/src/github.com/cilium/tetragon/
+        sudo -E echo "run race tests: " `uname -a`
+        sudo -E go mod verify
+        export TETRAGON_LIB=$(realpath "bpf/objs/")
+        make test-race RACE_TEST_TIMEOUT=45m
+
+    - name: Upload Tetragon logs
+      if: failure()
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: tetragon-race-logs
+        path: /tmp/tetragon.gotest*
+        retention-days: 5
--- a/Makefile
+++ b/Makefile
@@ -271,6 +271,15 @@ copy-golangci-lint:
 test: tester-progs tetragon-bpf ## Run Go tests.
 	$(GO) test -exec "$(SUDO)" -p 1 -parallel 1 $(GOFLAGS) -gcflags=$(GO_BUILD_GCFLAGS) -timeout $(GO_TEST_TIMEOUT) -failfast -cover ./pkg/... ./cmd/... ./operator/... ${EXTRA_TESTFLAGS}
 
+# Packages exercised by `make test-race`. Override on the command line to
+# expand or narrow the scope.
+RACE_PKGS ?= ./pkg/sensors ./pkg/metrics/... ./pkg/process/... ./pkg/reader/... ./pkg/server/... ./pkg/ratelimit/...
+RACE_TEST_TIMEOUT ?= 30m
+
+.PHONY: test-race
+test-race: tester-progs tetragon-bpf ## Run Go tests with the race detector (use RACE_PKGS=... to override scope).
+	$(GO) test $(GOFLAGS) -gcflags=$(GO_BUILD_GCFLAGS) -timeout $(RACE_TEST_TIMEOUT) -race $(RACE_PKGS) ${EXTRA_TESTFLAGS}
+
 .PHONY: tester-progs
 tester-progs: ## Compile helper programs for unit testing.
 	$(MAKE) -C $(TESTER_PROGS_DIR)
@@ -543,6 +552,8 @@ help: ## Display this help, based on https://www.thapaliya.com/en/writings/well-
 	$(call print_help_option,EXTRA_GO_BUILD_LDFLAGS,extra flags to pass to the Go linker)
 	$(call print_help_option,EXTRA_GO_BUILD_FLAGS,extra flags to pass to the Go builder)
 	$(call print_help_option,EXTRA_TESTFLAGS,extra flags to pass to the test binary)
+	$(call print_help_option,RACE_PKGS,packages exercised by test-race target)
+	$(call print_help_option,RACE_TEST_TIMEOUT,timeout for test-race target)
 
 .PHONY: version chart-version
 version: ## Print Tetragon version.
--- a/pkg/sensors/program/program.go
+++ b/pkg/sensors/program/program.go
@@ -30,6 +30,8 @@ package program
 
 import (
 	"fmt"
+	"maps"
+	"sync"
 
 	"github.com/cilium/ebpf"
 	"github.com/cilium/ebpf/btf"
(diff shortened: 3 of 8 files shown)


FedeDP (approved):
Great fix, thank you so much!
And also, the new `test-race` LGTM :)

tpapagian (approved):

ExceptionalHandler (approved):

pkg/sensors/sensors_race_test.go
```diff
@@ -0,0 +1,55 @@
...
+//
+// Run with -race to lock the regression in:
+//
+//	go test -race ./pkg/sensors/ -run TestSensor_TotalMemlock
+func TestSensor_TotalMemlock(_ *testing.T) {
+	prog := &program.Program{}
+	s := Sensor{Progs: []*program.Program{prog}}
+
+	const goroutines = 8
+	const iters = 5_000
```
mtardy: that's big numbers, I would assume you might need something way lower when running with race? Just making sure the test is not too slow when running usually.

pkg/sensors/program/program.go
```diff
@@ -177,9 +179,13 @@ type Program struct {
...
-	// Information of all maps used and loaded by this program by map ID. This
-	// is populated after load and used for map memlock accounting.
-	LoadedMapsInfo map[int]bpf.ExtendedMapInfo
+	// loadedMapsInfo holds info of all maps used and loaded by this program
+	// by map ID. It is populated after load and used for map memlock
+	// accounting. Unexported so all access goes through SetLoadedMapsInfo and
+	// CopyLoadedMapsInfo, which take mapsMu and prevent the data race between
+	// load/unload (writer) and gRPC ListTracingPolicies (reader).
+	mapsMu         sync.RWMutex
+	loadedMapsInfo map[int]bpf.ExtendedMapInfo
```
mtardy: maybe swap those two lines so that the comment goes into the loadedMapsInfo field?

pkg/sensors/sensors.go
```diff
@@ -153,8 +152,11 @@ func (s Sensor) TotalMemlock() uint64 {
 	uniqueMap := map[int]bpf.ExtendedMapInfo{}
 	for _, p := range s.Progs {
 		// we could first check that it exist then write but all maps with
-		// same ID on the kernel should share the same info
-		maps.Copy(uniqueMap, p.LoadedMapsInfo)
+		// same ID on the kernel should share the same info.
+		// Use CopyLoadedMapsInfo so we don't iterate the program's map while
+		// the load/unload path is mutating it (was: maps.Copy of a shared
+		// field, which racing readers could panic on).
```
mtardy: nit: Maybe we can remove this part, it seems this comment is repeated already.

Makefile
```diff
@@ -271,6 +271,16 @@ copy-golangci-lint:
 test: tester-progs tetragon-bpf ## Run Go tests.
 	$(GO) test -exec "$(SUDO)" -p 1 -parallel 1 $(GOFLAGS) -gcflags=$(GO_BUILD_GCFLAGS) -timeout $(GO_TEST_TIMEOUT) -failfast -cover ./pkg/... ./cmd/... ./operator/... ${EXTRA_TESTFLAGS}
 
+# Packages exercised by `make test-race`. Override on the command line to
+# expand or narrow the scope, e.g.:
+#   make test-race RACE_PKGS="./pkg/sensors ./pkg/grpc/..."
```
mtardy: nit: unnecessary maybe also since you already added the comment on the target (thanksk!!)

Makefile
```diff
@@ -271,6 +271,16 @@ copy-golangci-lint:
...
+RACE_PKGS ?= ./pkg/sensors ./pkg/metrics/... ./pkg/process/... ./pkg/reader/... ./pkg/server/... ./pkg/ratelimit/...
+RACE_TEST_TIMEOUT ?= 30m
+
+.PHONY: test-race
+test-race: tester-progs tetragon-bpf ## Run Go tests with the race detector (use RACE_PKGS=... to override scope).
+	$(GO) test -exec "$(SUDO)" -p 1 -parallel 1 $(GOFLAGS) -gcflags=$(GO_BUILD_GCFLAGS) -timeout $(RACE_TEST_TIMEOUT) -race $(RACE_PKGS) ${EXTRA_TESTFLAGS}
```
mtardy: > Default scope covers known-concurrent userspace packages; tracing/ BPF tests are excluded because race-instrumentation slowdown destabilises kernel-event timing.

This is really good thanks, maybe you can drop `-p 1 -parallel 1` then? Since it's userspace pkgs, we can test them in parallel this shouldn't be an issue.

.github/workflows/gotests.yml
```diff
@@ -122,3 +122,56 @@ jobs:
...
         go test --failfast .\pkg\... .\cmd\...
+
+  race:
+    name: Race detector
+    runs-on: ubuntu-24.04
+    timeout-minutes: 50
+    # Advisory for now: a detected race should surface in the job logs without
+    # blocking the PR while we shake out the initial package list. Flip to
+    # false (or remove the line) once the default RACE_PKGS set is stable.
+    continue-on-error: true
```
mtardy, line 133: ah this is nice, so we will see red but this won't not really fail the job?

Makefile
mtardy: and do you need `-exec "$(SUDO)"` also? maybe we can remove it

mtardy (commented):
Thanks a lot, I wanted to do that, but indeed the issue is that those run are way more slow. Since you selected the pkg, if the runtime is reasonable, it makes sense to run them on every PR!

A few comments

pkg/sensors/program/program.go
sayboras: sure :100:

Makefile
sayboras: ah good catch :100:

.github/workflows/gotests.yml
sayboras: yes, I am planning to add some more race tests, would explain some flaky at first.

mtardy (approved):
lgtm

PR #5065, opened 2026-05-26
pkg/crdutils: validate XValidation CEL rules in standalone

mtardy:
Previously, XValidation rules added in the CRDs were only checked from the k8s API server, this addition now run them in standalone (i.e. when passing a policy via --tracing-policy, or through the gRPC API for example).

```release-note
Fixed XValidation CEL rules not being enforced when loading tracing policies in standalone mode outside Kubernetes.
```

### DIFF
--- a/pkg/crdutils/crdutils.go
+++ b/pkg/crdutils/crdutils.go
@@ -9,6 +9,7 @@
 package crdutils
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
Great fix!

tpapagian (approved):

kkourt:
Does it make sense to backport this to v1.7? If not, please remove the label.

mtardy:
> Does it make sense to backport this to v1.7? If not, please remove the label.

I wondered the same, let's backport as this is was expected, should be easy

PR #5066, opened 2026-05-26
[v1.7] pkg/crdutils: validate XValidation CEL rules in standalone

mtardy:
Backport of https://github.com/cilium/tetragon/pull/5065

Previously, XValidation rules added in the CRDs were only checked from the k8s API server, this addition now run them in standalone (i.e. when passing a policy via --tracing-policy, or through the gRPC API for example).

```release-note
Fixed XValidation CEL rules not being enforced when loading tracing policies in standalone mode outside Kubernetes.
```

### DIFF
--- a/pkg/crdutils/crdutils.go
+++ b/pkg/crdutils/crdutils.go
@@ -9,6 +9,7 @@
 package crdutils
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


mtardy:
see https://www.githubstatus.com/

tpapagian (approved):

FedeDP (approved):

mtardy:
Ah the XValidation was missing on the tested field because it was not backported but we still have one that would require a change to the test (adding some helpers etc...) so I just dropped the test since we know the patch is good from main.

PR #5068, opened 2026-05-26
cmd/tetra: enable TLS for all TLS-related client flags

immanuwell:
### Description

Small fix for a TLS client footgun in `tetra`.

PR #4950 says any `--tls-*` flag should switch the client to TLS. In practice, `--tls-server-name` or `--tls-skip-verify` alone still kept insecure creds, so `tetra` could talk plaintext by accident. Kinda sneaky.

This makes any TLS-related client flag enable TLS, keeps the zero-flag plaintext path as-is, and adds tests plus docs.

### Repro

1. Start a TLS gRPC listener.
2. Dial it with one of these:
   - `tetra --server-address 127.0.0.1:54321 --tls-skip-verify getevents`
   - `tetra --server-address 127.0.0.1:54321 --tls-server-name node.tetragon-grpc.cilium.io getevents`
3. Before this patch, TLS is not enabled in those modes, so the client falls back to plaintext and the TLS connection cant work.

I re-checked it with a local TLS gRPC server too. After this patch, `--tls-skip-verify` alone reaches `READY`.

### Changelog

```release-note
Fix tetra TLS client flag handling for TLS-only connections
```

### DIFF
--- a/cmd/tetra/common/tls_test.go
+++ b/cmd/tetra/common/tls_test.go
@@ -27,8 +27,8 @@ func TestTLSConfigEnabled(t *testing.T) {
 		want bool
 	}{
 		{name: "zero", cfg: TLSConfig{}},
-		{name: "skip-verify alone", cfg: TLSConfig{SkipVerify: true}},
-		{name: "server-name alone", cfg: TLSConfig{ServerName: "x"}},
+		{name: "skip-verify alone", cfg: TLSConfig{SkipVerify: true}, want: true},
+		{name: "server-name alone", cfg: TLSConfig{ServerName: "x"}, want: true},
 		{name: "ca alone", cfg: TLSConfig{CAFiles: []string{"a"}}, want: true},
 		{name: "cert alone", cfg: TLSConfig{CertFile: "a"}, want: true},
 		{name: "key alone", cfg: TLSConfig{KeyFile: "a"}, want: true},
@@ -88,6 +88,25 @@ func TestTLSCredentialsLoadsCAAndCert(t *testing.T) {
 	})
 	require.NoError(t, err)
 	require.NotNil(t, creds)
+}
+
+func TestTLSCredentialsBuildsForTLSOnlyFlags(t *testing.T) {
+	t.Parallel()
+	cases := []struct {
+		name string
+		cfg  TLSConfig
+	}{
+		{name: "skip-verify only", cfg: TLSConfig{SkipVerify: true}},
+		{name: "server-name only", cfg: TLSConfig{ServerName: "node.tetragon-grpc.cilium.io"}},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			creds, err := TLSCredentials(tc.cfg)
+			require.NoError(t, err)
+			require.NotNil(t, creds)
+		})
+	}
 }
 
 func TestTLSCredentialsRejectsMissingFiles(t *testing.T) {
--- a/docs/content/en/docs/installation/grpc-tls.md
+++ b/docs/content/en/docs/installation/grpc-tls.md
@@ -98,6 +98,9 @@ tetra \
 ```
 
 Drop `--tls-cert-file` / `--tls-key-file` if the server is in TLS-only mode.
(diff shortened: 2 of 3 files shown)


cmd/tetra/common/tls_test.go
```diff
@@ -27,8 +27,8 @@ func TestTLSConfigEnabled(t *testing.T) {
 		want bool
 	}{
 		{name: "zero", cfg: TLSConfig{}},
-		{name: "skip-verify alone", cfg: TLSConfig{SkipVerify: true}},
-		{name: "server-name alone", cfg: TLSConfig{ServerName: "x"}},
+		{name: "skip-verify alone", cfg: TLSConfig{SkipVerify: true}, want: true},
+		{name: "server-name alone", cfg: TLSConfig{ServerName: "x"}, want: true},
```
mtardy, line 31: since this a change in behavior that was explicitly tested, I'll ask @sayboras to review.

docs/content/en/docs/installation/grpc-tls.md
```diff
@@ -98,6 +98,9 @@ tetra \
 ```
 
 Drop `--tls-cert-file` / `--tls-key-file` if the server is in TLS-only mode.
+If the server certificate chains to the platform trust store, you can also use
```
mtardy: question, what is the platform trust store?

mtardy (commented):
code makes sense to me but I'm missing something (see question). However I'll let Tam say whether the behavior change is okay.

cmd/tetra/common/tls_test.go
sayboras: thanks for the ping, I am taking a look :100:

docs/content/en/docs/installation/grpc-tls.md
sayboras: is it host/system level cert (e.g. /etc/ssl/...) ?

contrib/upgrade-notes/latest.md
```diff
@@ -7,7 +7,10 @@ Depending on your setup, changes listed here might require a manual intervention
 
 ### Agent Options
 
-* TBD
+* `tetra` now enables TLS when `--tls-server-name` or `--tls-skip-verify`
+  is used, even without `--tls-ca-cert-files` or client mTLS material.
+  This fixes TLS-only client connections that rely on the platform trust
+  store or dev-only skip-verify mode.
```
sayboras: we probably can remove this note, as the original PR adding mTLS feature is only in main branch now.

docs/content/en/docs/installation/grpc-tls.md
immanuwell: upd, I replaced "platform trust store" with more explicit "host OS trust store" / "system CA bundle", I meant place like /etc/ssl/...

contrib/upgrade-notes/latest.md
immanuwell: agreed, removed it

immanuwell:
@mtardy @sayboras done, addressed your suggestions

sayboras:
> done, addressed your suggestions

Somehow the commit message is having \n character in plain text. Can you help to update?

immanuwell:
@sayboras done, I removed \n from the commit message

mtardy (approved):
Fix looks good to me, there's an issue with the CI. I see @sayboras that you re-ran it multiple time, let's either find if there's an issue related to this PR or create an issue for the flake or broken test.

Otherwise I think we can merge but I'll let Tam do it.

sayboras:
Yeah, I meant to take a look at the CI failures, will share my findings here.

sayboras:
I think i know why the CI is failing

- The PR branched from a 2026-05-26 merge-base.
- The matchWorkloads feature — both the CRD field (commit 3100a903a) and the e2e test (commit b3bff5801) — landed on main on 2026-05-27, after the PR was cut.
- The e2e job checks out the merged source (so it runs the new TestMatchWorkloadsSelector and ships the new policy YAML), but the operator image that installs the CRD is built from the PR head, which predates the matchWorkloads CRD field. The operator therefore installs an old CRD without matchWorkloads, the API server prunes the field, and the test fails.

sayboras (approved):
Thanks 💯

PR #5071, opened 2026-05-27
cleanup: moved kprobe and uprobe validation to k8s x-validation markers.

FedeDP:
### Description

Move to use k8s x-validation where possible.
The only field we can pre-validate are:
* kprobe `Return==true` -> requires `returnArg`
* uprobe `Return==true` -> requires `returnArg`
* uprobe -> at most one between `addrs,symbols,offsets`

### Changelog

```release-note
cleanup: moved kprobe and uprobe validation to k8s x-validation markers.
```

### DIFF
--- a/pkg/crdutils/crdutils_test.go
+++ b/pkg/crdutils/crdutils_test.go
@@ -202,6 +202,9 @@ spec:
       type: "string"
     - index: 4
       type: "skb"
+    returnArg:
+      index: 0
+      type: "int"
   - call: "another_func"
     return: false
     syscall: false
@@ -289,6 +292,10 @@ var expectedData = GenericTracingPolicy{
 						Index: 4,
 						Type:  "skb",
 					},
+				},
+				ReturnArg: &v1alpha1.KProbeArg{
+					Index: 0,
+					Type:  "int",
 				},
 			},
 			{
--- a/pkg/k8s/apis/cilium.io/v1alpha1/types.go
+++ b/pkg/k8s/apis/cilium.io/v1alpha1/types.go
@@ -29,6 +29,7 @@ type KprobeIgnore struct {
 	CallNotFound bool `json:"callNotFound,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="!self.return || has(self.returnArg)",message="ReturnArg not specified with Return=true."
 type KProbeSpec struct {
 	// Name of the function to apply the kprobe spec to.
 	Call string `json:"call"`
@@ -364,6 +365,8 @@ type TracepointSpec struct {
 	Raw bool `json:"raw,omitempty"`
 }
 
+// +kubebuilder:validation:XValidation:rule="!(has(self.symbols) && has(self.addrs)) && !(has(self.symbols) && has(self.offsets)) && !(has(self.addrs) && has(self.offsets))",message="uprobe needs at most one of symbols, addrs or offsets defined"
+// +kubebuilder:validation:XValidation:rule="!self.return || has(self.returnArg)",message="ReturnArg not specified with Return=true."
 type UProbeSpec struct {
 	// Name of the traced binary
 	Path string `json:"path"`
--- a/pkg/tracingpolicy/k8s_xvalidation_test.go
+++ b/pkg/tracingpolicy/k8s_xvalidation_test.go
@@ -1,0 +1,83 @@
+//go:build !nok8s
+
+package tracingpolicy
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestKprobeValidationReturnWithoutArg(t *testing.T) {
+	// missing returnArg while having return: true
+	crd := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "missing-returnarg"
+spec:
+  kprobes:
+  - call: "sys_openat"
+    return: true
+    syscall: true
+`
+	_, err := FromYAML(crd)
+	require.Error(t, err)
+}
+
+func testUprobeValidationSymbolsAddrsOffsets(t *testing.T, withSymbol, withAdrr, withOff bool) {
+	crd := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "uprobe"
+spec:
+  uprobes:
+  - path: "/usr/bin/test"`
+	if withSymbol {
+		crd += "    symbols: [\"test_3\"]\n"
+	}
+	if withAdrr {
+		crd += "    addrs: [0x985256]\n"
+	}
+	if withOff {
+		crd += "    offsets: [0x9156366]\n"
+	}
+
+	_, err := FromYAML(crd)
+	require.Error(t, err)
+}
+
+func TestUprobeValidationSymbolsAddrsOffsets(t *testing.T) {
+	testUprobeValidationSymbolsAddrsOffsets(t, true, true, true)
+}
+
+func TestUprobeValidationSymbolsAddrs(t *testing.T) {
+	testUprobeValidationSymbolsAddrsOffsets(t, true, true, false)
+}
+
+func TestUprobeValidationSymbolsOffsets(t *testing.T) {
+	testUprobeValidationSymbolsAddrsOffsets(t, true, false, true)
+}
+
+func TestUprobeValidationAddrsOffsets(t *testing.T) {
+	testUprobeValidationSymbolsAddrsOffsets(t, false, true, true)
+}
+
+func TestUprobeValidationReturnWithoutArg(t *testing.T) {
+	crd := `
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "uprobe"
+spec:
+  uprobes:
+  - path: "/usr/bin/test"
+    symbols:
+    - "test_3"
+    return: true
+`
+
+	_, err := FromYAML(crd)
+	require.Error(t, err)
+}
--- a/install/kubernetes/tetragon/crds-yaml/cilium.io_tracingpolicies.yaml
+++ b/install/kubernetes/tetragon/crds-yaml/cilium.io_tracingpolicies.yaml
@@ -1332,6 +1332,9 @@ spec:
                   required:
                   - call
                   type: object
+                  x-kubernetes-validations:
+                  - message: ReturnArg not specified with Return=true.
+                    rule: '!self.return || has(self.returnArg)'
                 type: array
               hostSelector:
                 description: |-
@@ -2611,6 +2614,9 @@ spec:
                   required:
                   - call
                   type: object
+                  x-kubernetes-validations:
+                  - message: ReturnArg not specified with Return=true.
+                    rule: '!self.return || has(self.returnArg)'
                 type: array
               lists:
                 description: A list of list specs.
@@ -6687,6 +6693,13 @@ spec:
                   required:
                   - path
                   type: object
+                  x-kubernetes-validations:
+                  - message: uprobe needs at most one of symbols, addrs or offsets
+                      defined
+                    rule: '!(has(self.symbols) && has(self.addrs)) && !(has(self.symbols)
+                      && has(self.offsets)) && !(has(self.addrs) && has(self.offsets))'
+                  - message: ReturnArg not specified with Return=true.
+                    rule: '!self.return || has(self.returnArg)'
                 type: array
               usdts:
                 description: A list of usdt specs.
--- a/install/kubernetes/tetragon/crds-yaml/cilium.io_tracingpoliciesnamespaced.yaml
+++ b/install/kubernetes/tetragon/crds-yaml/cilium.io_tracingpoliciesnamespaced.yaml
@@ -1332,6 +1332,9 @@ spec:
                   required:
                   - call
                   type: object
+                  x-kubernetes-validations:
+                  - message: ReturnArg not specified with Return=true.
+                    rule: '!self.return || has(self.returnArg)'
                 type: array
               hostSelector:
                 description: |-
@@ -2611,6 +2614,9 @@ spec:
                   required:
                   - call
                   type: object
+                  x-kubernetes-validations:
+                  - message: ReturnArg not specified with Return=true.
+                    rule: '!self.return || has(self.returnArg)'
                 type: array
               lists:
                 description: A list of list specs.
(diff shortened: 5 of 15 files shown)


mtardy:
just noting: it's a very good idea now that we know it's supported properly

FedeDP:
I am still trying to understand why CI is failing on my PRs :/ Anyway, this is ready for review now.

sayboras:
> I am still trying to understand why CI is failing on my PRs :/ Anyway, this is ready for review now.

Probably due to the below error

```
level=error msg="Unable to Register CRDs" subsys=crd error="failed to create TracingPolicy/v1alpha1: CustomResourceDefinition.apiextensions.k8s.io \"tracingpolicies.cilium.io\" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[uprobes].items.x-kubernetes-validations[1].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.6x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)\nfailed to create TracingPolicyNamespaced/v1alpha1: CustomResourceDefinition.apiextensions.k8s.io \"tracingpoliciesnamespaced.cilium.io\" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[uprobes].items.x-kubernetes-validations[1].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.6x (try simplifying the rule, or adding maxItems, maxProperties, and maxLength where arrays, maps, and strings are declared)"
...
```

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -364,6 +365,8 @@ type TracepointSpec struct {
 	Raw bool `json:"raw,omitempty"`
 }
 
+// +kubebuilder:validation:AtMostOneOf=symbols;addrs;offsets
+// +kubebuilder:validation:XValidation:rule="!self.return || has(self.returnArg)",message="ReturnArg not specified with Return=true."
```
sayboras, line 369: this might cause complexity issue

```
level=error msg="Unable to Register CRDs" subsys=crd error="failed to create TracingPolicy/v1alpha1: CustomResourceDefinition.apiextensions.k8s.io \"tracingpolicies.cilium.io\" is invalid: spec.validation.openAPIV3Schema.properties[spec].properties[uprobes].items.x-kubernetes-validations[1].rule: Forbidden: estimated rule cost exceeds budget by factor of 1.6x
```

FedeDP: Uh great catch! Can you point me to this CI failure? I cannot find this log line in the CI :/

sayboras: ah I faced similar issue before in cilium, so just run `make kind-setup` locally from this branch and check the logs.

CI might not have previous log captured.

FedeDP: Oh thanks! I have a path forward at least now :heart:

FedeDP: The one breaking the complexity is:
```
// +kubebuilder:validation:AtMostOneOf=symbols;addrs;offsets
type UProbeSpec struct {
```

Adding `// +kubebuilder:validation:MaxItems=16` on `symbols`, `addrs` and `offsets` does not help either :/
I will probably drop that check from kubebuilder and restore the old one in genericuprobe.

sayboras: or give it a try with something like below instead ? 

```
// +kubebuilder:validation:XValidation:rule="!(has(self.symbols) && has(self.addrs)) && !(has(self.symbols) && has(self.offsets)) && !(has(self.addrs) && has(self.offsets))",message="uprobe needs at most one of symbols, addrs or offsets"
```

FedeDP: Yay using it plain like you suggested (getting rid `AtMostOneOf`) works fine! Thanks!

FedeDP: I discovered that the issue was due to the `AtMostOneOf` validation on `TracingPolicySpec` type that then gets expanded to the x-validation for `UProbeSpec` and gets too complex.
Another fix would've been to add a `MaxItems` on `TracingPolicySpec.{kprobes,uprobes,...}` fields, but your solution is fine and we don't have to add any limit.

pkg/sensors/tracing/generickprobe.go
```diff
@@ -878,9 +878,7 @@ func addKprobe(funcName string, instance int, f *v1alpha1.KProbeSpec, in *addKpr
 	// instructs the BPF kretprobe program which type of copy to use. And
 	// argReturnPrinters tell golang printer piece how to print the event.
 	if f.Return {
-		if f.ReturnArg == nil {
-			return errFn(errors.New("ReturnArg not specified with Return=true"))
-		}
+		// ReturnArg != nil is specified from crd validation (see TestUprobeValidationReturnWithoutArg())
```
sayboras: we might need to keep this check, mainly for nok8s build

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -746,9 +730,7 @@ func addUprobe(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, in *addUprobeIn
 	// instructs the BPF uretprobe program which type of copy to use. And
 	// argReturnPrinters tell golang printer piece how to print the event.
 	if spec.Return {
-		if spec.ReturnArg == nil {
-			return nil, errors.New("ReturnArg not specified with Return=true")
-		}
+		// ReturnArg != nil is specified from crd validation (see TestUprobeValidationReturnWithoutArg())
```
sayboras: same as above

sayboras (commented):
probably need to double check the code path for nok8s if we need to keep some of the checks

pkg/sensors/tracing/generickprobe.go
FedeDP: oh right, that's quite a problem with leveraging k8s validation :/ didn't think about that.
@mtardy wdyt? I'd close this one; if we have to repeat the check in both code and x-validation, it makes no sense.

kkourt: Let's keep both.

For k8s, it's very helpful to have the validation checks because they happen before the policy reaches the agent (at the API server), which is much more convenient for the end user.

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -143,28 +142,6 @@ func (h policyHandler) PolicyHandler(
...
-	}
-	if len(spec.Usdts) > 0 {
-		sections++
-	}
-	if len(spec.Fentries) > 0 {
-		sections++
-	}
-	if sections > 1 {
-		return nil, errors.New("tracing policies with multiple sections of kprobes, tracepoints, lsm hooks, uprobes or usdts are currently not supported")
-	}
```
kkourt, line 168: As an aside, having policies with only one section was due to some issues with maps in the past. I believe these issues should have been fixed now, so we could add some tests and remove this limitation.

@olsajiri Am I missing something?

pkg/sensors/tracing/generickprobe.go
FedeDP: Makes sense @kkourt , let me update the PR.

pkg/sensors/tracing/policyhandler_linux.go
FedeDP: Side question: do we want to support tracing policies without at least one of `{kprobes,tracepoints,uprobes,lsm,usdts,fentries}`? I don't think that makes sense.
Today we just return `nil,nil` : https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/policyhandler_linux.go#L225

I'd expect it to throw an error in that specific case.

kkourt: For some cases (e.g., when templating) it might be useful to support "empty" policies.

FedeDP: I see, thanks for the clarification!

pkg/sensors/tracing/generickprobe.go
mtardy: Yeah I agree that x-validation is useful indeed in a k8s context anyway! That's a bit annoying that we now again have to do both while we did everything to support it on standalone to simplify it...

pkg/crdutils/crdutils_test.go
```diff
@@ -189,7 +189,7 @@ metadata:
 spec:
   kprobes:
   - call: "example_func"
-    return: true
+    return: true
```
mtardy: nit: whitespace change

pkg/sensors/tracing/policyhandler_linux.go
mtardy: I'd put these changes into a separate commit

pkg/k8s/apis/cilium.io/v1alpha1/types.go
mtardy: might be interesting to see if we don't introduce too much CEL complexity by doing this, maybe we should measure it? We don't want the k8s API to struggle on validating our CRD or our own agent.

mtardy (commented):
agree with the rest of the feedback here. I'd maybe split this into:
- new x-validation in the CRDs + generation
- changes in non valid policies in tests
- changes in the runtime checks
- changes about the policies main sections? (maybe group the last two)

pkg/sensors/tracing/policyhandler_linux.go
FedeDP: Sure :)

pkg/k8s/apis/cilium.io/v1alpha1/types.go
FedeDP: Do you have any idea about how to measure it?
Btw if we get rid of the `AtMostOneOf` (as suggested by @kkourt ) we should be good to go.

pkg/sensors/tracing/policyhandler_linux.go
FedeDP: > As an aside, having policies with only one section was due to some issues with maps in the past. I believe these issues should have been fixed now, so we could add some tests and remove this limitation.

So, changing this, requires updating `PolicyHandler` signature to:
```
func (h policyHandler) PolicyHandler(
	policy tracingpolicy.TracingPolicy,
	policyID policyfilter.PolicyID,
) ([]sensors.SensorIface, error) {
```
, and that means updating `type policyHandler interface` too (and touching all policyHandler implementations).
I am not sure whether we want to do that in the context of this PR.
WDYT?

FedeDP: (i got the commit ready in any case; just let me know if you want it pushed as part of this PR or in a new one)

kkourt: I would handle this in another PR.

> So, changing this, requires updating PolicyHandler signature [...]

Why? Can't we install multiple handlers? That is, one per section?

FedeDP: Then this PR is ready! :)

> Why? Can't we install multiple handlers? That is, one per section?

Mmh what do you mean?
Today, `PolicyHandler` takes a `TracingPolicy` and returns a single `SensorIface`. With my patch, that would return a `[]SensorIface`. I think it makes sense since a single tracing policy is now going to build multiple sensors (one for each section).

kkourt: I mean that we don't have to change the interface. We can instead register multiple policyHandlers: one for kprobes, one for tracepoints, and so on so forth.

FedeDP: Oh i see, that seems a better approach indeed. I'll dig into it!

FedeDP: Here is the PR: https://github.com/cilium/tetragon/pull/5184

pkg/sensors/tracing/uprobe_validation_test.go
```diff
@@ -42,3 +44,62 @@ spec:
...
+  uprobes:
+  - path: "` + uprobe + `"
+    symbols:
+    - "test_3"
+    return: true
+`
+
+	_, err := tracingpolicy.FromYAML(crd)
+	require.Error(t, err)
+}
```
mtardy: I think the fact that your added tests are not calling anything from pkg/sensors/tracing but mostly `tracingpolicy` showcases that.

mtardy (commented):
overall lgtm, just one last thing: maybe would be worth testing all those validations you added to make sure they do what we want?

btw we use the validation tests from the sensor because we have the runtime validations here but would be maybe better to put the validation with the code that is running the CRDs CEL checks (the FromYAML)? maybe in https://github.com/cilium/tetragon/tree/main/pkg/tracingpolicy

kkourt (approved):
LGTM, thanks!

pkg/sensors/tracing/uprobe_validation_test.go
FedeDP: You are right, moved under `pkg/tracingpolicy`. Also, only test them on !nok8s builds :)

pkg/tracingpolicy/k8s_xvalidation_test.go
```diff
@@ -0,0 +1,85 @@
...
+
+package tracingpolicy
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// In this file we test k8s API x-validation.
```
mtardy: sorry a nit, for next PR: but I'll try to avoid unnecessary comment: this is already the name of the file.

mtardy (approved):
lgtm! thanks, only some CI failure to fix, not sure why but should be good to go eventually :)

pkg/tracingpolicy/k8s_xvalidation_test.go
FedeDP: Yes fixed haha thanks!

FedeDP:
> lgtm! thanks, only some CI failure to fix, not sure why but should be good to go eventually :)

yes the problem is that `checkCrd()` used by `kprobe_validation_test` is going to fail at 
```
tp, err := tracingpolicy.FromYAML(crd)
```
in k8s builds.

I now added a `build.K8sEnabled()` check and fixed `TestKprobeValidationMissingReturnArg` consequently.

PR #5073, opened 2026-05-28
bpf: cleanup more bpf_probe_read

3u13r:
### Description

Follow up from this discussion: https://github.com/cilium/tetragon/pull/4989#discussion_r3264329057

I couldn't convert some of the `bpf_probe_read`s because while they were reading from e.g. eBPF map values, they had a dynamic size to be copied. For `memcpy` the size must be constexpr.

#### Questions for the reviewers
1. Did I miss any `probe_read`s, that still can be converted?
2. Did I convert `probe_read`s in place where I shouldn't?

### Changelog

```release-note
bpf: replace some bpf_probe_read calls
```

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -321,26 +321,26 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 	case size_type:
 	case s64_ty:
 	case u64_ty:
-		probe_read(args, sizeof(__u64), &arg);
+		*(__u64 *)args = (__u64)arg;
 		size = sizeof(__u64);
 		break;
 	/* Consolidate all the types to save instructions */
 	case int_type:
 	case s32_ty:
 	case u32_ty:
-		probe_read(args, sizeof(__u32), &arg);
+		*(__u32 *)args = (__u32)arg;
 		size = sizeof(__u32);
 		break;
 	case s16_ty:
 	case u16_ty:
 		/* read 2 bytes, but send 4 to keep alignment */
-		probe_read(args, sizeof(__u16), &arg);
+		*(__u16 *)args = (__u16)arg;
 		size = sizeof(__u32);
 		break;
 	case s8_ty:
 	case u8_ty:
 		/* read 1 byte, but send 4 to keep alignment */
-		probe_read(args, sizeof(__u8), &arg);
+		*(__u8 *)args = (__u8)arg;
 		size = sizeof(__u32);
 		break;
 	case skb_type:
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -277,16 +277,13 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 			asm volatile("%[bytes] &= 0xfff;\n"
 				     : [bytes] "+r"(bytes)
 				     :);
-			ret = probe_read(&args[4], bytes + 4, (char *)&val->file[0]);
+			// size + file path + flags
+			size = 4 + bytes + 4;
+			ret = probe_read(&args[4], size, (char *)&val->file[0]);
 			if (ret < 0)
 				return ret;
-			size = bytes + 4 + 4;
 
-			// flags
-			ret = probe_read(&args[size], 4,
-					 (char *)&val->file[size - 4]);
-			if (ret < 0)
-				return ret;
+			// account for fd written at args[0]
 			size += 4;
 		} else {
 			/* If filter specification is fd type then we
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -298,7 +298,7 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 	case filename_ty: {
 		struct filename *file;
 
-		probe_read(&file, sizeof(file), &arg);
+		file = (struct filename *)arg;
 		probe_read(&arg, sizeof(arg), &file->name);
 	}
 		fallthrough;
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -929,6 +929,7 @@ filter_char_buf(struct selector_arg_filter *filter, char *args, int value_off)
 	return is_not_operator(filter->op) ? !match : match;
 }
 
+// This struct captures the layout documented in store_path().
 struct string_buf {
 	__u32 len;
 	char buf[];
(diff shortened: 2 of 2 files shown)


bpf/process/types/basic.h
```diff
@@ -2514,7 +2522,7 @@ rate_limit(__u64 ratelimit_interval, __u64 ratelimit_scope, struct msg_generic_k
 	}
 
 	// Clean the heap
-	probe_read(key->data, MAX_POSSIBLE_ARGS * KEY_BYTES_PER_ARG, ro_heap);
+	memcpy(key->data, ro_heap, MAX_POSSIBLE_ARGS * KEY_BYTES_PER_ARG);
```
olsajiri: let's keep this one, that memcpy translates to 400 instructions:

```
;       memcpy(key->data, ro_heap, MAX_POSSIBLE_ARGS * KEY_BYTES_PER_ARG);
    4705:       71 01 07 00 00 00 00 00 r1 = *(u8 *)(r0 + 0x7)
    4706:       73 19 1f 00 00 00 00 00 *(u8 *)(r9 + 0x1f) = r1
    4707:       71 01 06 00 00 00 00 00 r1 = *(u8 *)(r0 + 0x6)
    4708:       73 19 1e 00 00 00 00 00 *(u8 *)(r9 + 0x1e) = r1
    4709:       71 01 05 00 00 00 00 00 r1 = *(u8 *)(r0 + 0x5)
    4710:       73 19 1d 00 00 00 00 00 *(u8 *)(r9 + 0x1d) = r1
    4711:       71 01 04 00 00 00 00 00 r1 = *(u8 *)(r0 + 0x4)
    4712:       73 19 1c 00 00 00 00 00 *(u8 *)(r9 + 0x1c) = r1
    4713:       71 01 03 00 00 00 00 00 r1 = *(u8 *)(r0 + 0x3)
    4714:       73 19 1b 00 00 00 00 00 *(u8 *)(r9 + 0x1b) = r1
    4715:       71 01 02 00 00 00 00 00 r1 = *(u8 *)(r0 + 0x2)
    4716:       73 19 1a 00 00 00 00 00 *(u8 *)(r9 + 0x1a) = r1
    4717:       71 01 01 00 00 00 00 00 r1 = *(u8 *)(r0 + 0x1)
    ...
```

bpf/process/types/basic.h
```diff
@@ -2327,7 +2330,10 @@ struct fdinstall_key {
 };
 
 struct fdinstall_value {
```
olsajiri, line 2352: there's PR removing all this fd tracking stuff https://github.com/cilium/tetragon/pull/4718 , not sure why it got stalled

bpf/process/types/basic.h
```diff
@@ -2514,7 +2520,7 @@ rate_limit(__u64 ratelimit_interval, __u64 ratelimit_scope, struct msg_generic_k
 	}
 
 	// Clean the heap
-	probe_read(key->data, MAX_POSSIBLE_ARGS * KEY_BYTES_PER_ARG, ro_heap);
+	memcpy(key->data, ro_heap, MAX_POSSIBLE_ARGS * KEY_BYTES_PER_ARG);
```
kevsecurity: This will be quite a big memcpy. The point of using probe_read() here is to trade the small CPU cost of launching a helper for the huge complexity win of reducing instruction count. We are practically at the point where we are failing to add functionality to older kernels because of complexity (and it being easier to just gate it at a later kernel). I would then implore you to not add to this by swapping probe_read()s for memcpy()s when the cost is significant.

bpf/process/generic_calls.h
```diff
@@ -321,26 +321,26 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 	case size_type:
 	case s64_ty:
 	case u64_ty:
-		probe_read(args, sizeof(__u64), &arg);
+		*(__u64 *)args = (__u64)arg;
```
kevsecurity, line 321: Does this work for uprobes just as well? I always assumed these probe_read()s had some purpose.

3u13r: From the xlated view I'd argue that it's a u64 variable that was already copied to the stack:
```
; *(__u64 *)args = (__u64)arg; [file:/home/lcohnen/projects/tetragon2/tetragon/bpf/process/generic_calls.h line_num:321 line_col:18]
1100: (79) r1 = *(u64 *)(r10 -184)
1101: (7b) *(u64 *)(r8 +0) = r1
```

Could you maybe elaborate on what the assumption is why this is needed?

bpf/process/types/basic.h
3u13r: Yeah I think that @FedeDP is working on this currently [see Slack](https://isovalent.slack.com/archives/C01N6G0CHFV/p1779953133493359).

Should we omit the changes here? Otherwise I'd just leave them and let them be removed with the rest of the feature. 
Otherwise we could also add a comment about the deprecation of that feature.

bpf/process/generic_calls.h
kevsecurity: > Could you maybe elaborate on what the assumption is why this is needed?

No idea; it was that way before I started working on it.

bpf/process/types/basic.h
3u13r: Oh interesting, I thought that it would do u64 sized copies. Maybe we need something like [this](https://github.com/cilium/cilium/commit/9c85721783447e10e359f723a1ee7d465c704ccc) in Tetragon eventually.

Will change it back to using the helper.

bpf/process/types/basic.h
3u13r: x-ref: https://github.com/cilium/tetragon/pull/5073#discussion_r3334387441

Good point, removed the change.

3u13r:
I'm not sure how the failing test relates to my change. I think the error in the test is:
```
observer_test_helper.go:449: SensorManager.AddTracingPolicy error: policy handler 'tracing' failed loading policy 'uprobe': error: Failed to retrieve register argument 'rdi'
```
And the arm version of the code does not resolve 'rdi' and returns an error, which seems correct for this input. But I wonder why this is the input.

EDIT: see [PR](https://github.com/cilium/tetragon/pull/5082)

bpf/process/types/basic.h
FedeDP: Oh didn't notice there was already a PR around that thing; i think my PR will supersedes that one then :) 
> Should we omit the changes here? Otherwise I'd just leave them and let them be removed with the rest of the feature.

100% this, i'll take care of rebasing mine and dropping the new code.

3u13r: Thanks!

3u13r:
CI seems good now.
Regarding labels I'd request "release-note/misc".
I don't think there's any need for backports.

FedeDP (approved):
Great job! LGTM

olsajiri (approved):

bpf/process/types/basic.h
```diff
@@ -2516,6 +2520,8 @@ rate_limit(__u64 ratelimit_interval, __u64 ratelimit_scope, struct msg_generic_k
 	if (!key)
 		return false;
 	ro_heap = map_lookup_elem(&heap_ro_zero, &zero);
+	if (!ro_heap)
```
kevsecurity: This ```map_lookup_elem()``` cannot really fail if the code managed to compile (unless we've done something terribly funky in the loader) and is JITed to a direct memory load in the loader. We only really need the test if the verifier is complaining, but it doesn't mind when it is used in a ```probe_read()```. I guess this was left over from when you swapped this ```probe_read()``` to a ```memcpy()```.

3u13r: Good point, I guess if this fails then there's something truly broken that we catch in the CI. Let me remove this again.

bpf/process/types/basic.h
mtardy: you can hint clang to generate u64 size copies by casting the type in the memcpy

mtardy (approved):
lgtm!

PR #5080, opened 2026-06-01
renovate: add dependency cooldown

peoyekunle:
- adds a 5-day dependency cooldown to the Renovate config for stability and security to prevent supply chain vulnerabilities
- Vulnerability updates are exempt from this cooldown by default.

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -85,6 +85,11 @@
   //   important rules override settings from earlier rules if needed.
   "packageRules": [
     {
+      // Dependency cooldown: skip versions published less than 5 days ago
+      "matchUpdateTypes": ["major", "minor", "patch"],
+      "minimumReleaseAge": "5 days"
+    },
+    {
       "groupName": "all github action dependencies",
       "groupSlug": "all-github-action",
       "matchFileNames": [
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -90,6 +90,15 @@
       "minimumReleaseAge": "5 days"
     },
     {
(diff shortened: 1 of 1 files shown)


.github/renovate.json5
```diff
@@ -84,6 +84,11 @@
   //   are at the top, and the most important rules at the bottom. This way
   //   important rules override settings from earlier rules if needed.
   "packageRules": [
+    {
+      // Dependency cooldown: skip versions published less than 5 days ago
+      "matchUpdateTypes": ["major", "minor", "patch"],
```
mtardy, line 89: maybe even digest?

.github/renovate.json5
```diff
@@ -84,6 +84,11 @@
...
+      "minimumReleaseAge": "5 days"
```
mtardy, line 90: just curious: why 5 days?

mtardy (approved):
Yeah I generally agree, I wanted to add something like that.

mtardy (approved):
thanks, should we merge this? 

also please next time wrap the text in the commit logs

.github/renovate.json5
peoyekunle: Recent supply chain compromises have often been discovered within the first three days, I added a 2-day buffer. [Dependabot](https://github.blog/security/supply-chain-security/the-case-for-a-cooldown-why-dependabot-now-waits-before-issuing-version-updates/) uses a three-day delay. I don’t have a strong opinion on the exact number of days, from a security perspective, anything longer than three days is fine.

PR #5088, opened 2026-06-02
Check for images tags when updated in helm values

### LINKED ISSUE #5020: CI doesn't fail when rthooks image tag points to non-existent version
### Problem

In #5014 the tetragon-rthooks image tag was bumped to v0.9 before that tag was actually published. CI passed regardless.
Suspected cause: CI doesn't deploy with rthooks enabled, so the missing image is never pulled.

### Impact

Broken/missing rthooks image tags ship to users without CI signal. Also, bad helm chart values can merge silently.

### Expected

CI should fail (or at least warn) when a referenced image tag doesn't exist in the registry — at minimum for the rthooks deployment path.

### Proposed directions

* Enable rthooks in an existing e2e/install CI job so image pull fails surface.
* Or add a lightweight job that resolves every image tag referenced in  against the registry (manifest HEAD).

### References

* PR #5014 (the bump)
* PR #5017 (Renovate equivalent)
* Review: https://github.com/cilium/tetragon/pull/5014#pullrequestreview-4312290899

dangome3:
### Description
Adds check to run when the kubernetes `values.yaml` is updated. This verifies that all images tags in the file are published in their registry.

Fixes #5020

### Testing

This is a successful run:
```
make -C install/kubernetes check-images-tags                                                                      
Checking quay.io/cilium/tetragon:v1.7.0
Checking quay.io/cilium/tetragon-operator:v1.7.0
Checking quay.io/cilium/hubble-export-stdout:v1.1.1
Checking quay.io/cilium/tetragon-rthooks:v0.9
Checking quay.io/cilium/certgen:v0.4.5
```

This is a run when there is a non existing image:

```
make -C install/kubernetes check-images-tags
Checking quay.io/cilium/tetragon:v1.7.0
Checking quay.io/cilium/tetragon-operator:v1.7.0
Checking quay.io/cilium/hubble-export-stdout:v1.1.1
Checking quay.io/cilium/tetragon-rthooks:v0.10
  MISSING: quay.io/cilium/tetragon-rthooks:v0.10
make: *** [check-images-tags] Error 1
```
Here an example run https://github.com/cilium/tetragon/actions/runs/26846697582/job/79168054353?pr=5088#step:6:13

### DIFF
--- a/.github/workflows/lint-helm.yaml
+++ b/.github/workflows/lint-helm.yaml
@@ -61,6 +61,8 @@ jobs:
         run: |
           set -ex
           make -C install/kubernetes lint
+      - name: Check images tags exists
+        run: make -C install/kubernetes check-images-tags
       - name: Run Kubeconform with minimum supported K8s version
         if: success() || steps.validate_openapi2jsonschema_script.outcome == 'failure'
         id: kubeconform_min_k8s_version
--- a/install/kubernetes/Makefile
+++ b/install/kubernetes/Makefile
@@ -84,6 +84,18 @@ openapi2jsonschema.py: ## openapi2jsonschema.py script generating JSON schema fr
 lint: ## Lint the Helm chart
 	$(HELM) lint . --with-subcharts
 
+.PHONY: check-images-tags
+check-images-tags: ## Verify all image:tag references in values.yaml exist in their registries
+	@set -e; \
+	IMAGES=$$(yq -e '.. | select(has("repository") and has("tag")) | .repository + ":" + .tag' tetragon/values.yaml); \
+	for img in $$IMAGES; do \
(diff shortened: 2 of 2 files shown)


PhilipSchmid (approved):
Overall, LGTM, thanks!

One thing to mention: Since it parses `tetragon/values.yaml` instead of the final rendered image values from the K8s resources, it won't detect non-existent images specified via `image.override`. However, I think this should be fine since we're unlikely to ever put a value in the `override` field in the publicly hosted Helm chart template. Also, `check-images-tags` is not included in `make all`, so it should also not block local development when playing around with local dev images.

FedeDP (approved):
LGTM! Thanks!

PR #5090, opened 2026-06-03
cleanup: drop deprecated fd actions

### LINKED ISSUE #4580: remove followFD and friends from BPF

FedeDP:
### Description

This PR aims to drop deprecated `followfd`, `unfollowfd` and `copyfd` actions.

### Changelog

```release-note
cleanup: drop deprecated actions
```


FedeDP:
Waiting on https://github.com/cilium/tetragon/pull/5109 to eventually rebase and make this ready for review.

bpf/process/types/basic.h
```diff
@@ -117,12 +117,12 @@ enum {
 
 enum {
 	ACTION_POST = 0,
-	ACTION_FOLLOWFD = 1,
+	// ACTION_FOLLOWFD = 1, deprecated
```
kkourt, line 120: ~I think we can reuse these because it's an internal interface between the agent and the bpf. If so, do you want to add a comment mentioning that these IDs can be reused if needed?~

I guess we can't reuse these because they are defined in .proto? Can you add a comment to that effect?

contrib/upgrade-notes/latest.md
```diff
@@ -24,9 +24,10 @@ Depending on your setup, changes listed here might require a manual intervention
   return value in events, and omit `returnArgAction` for the default return-value
   behavior. Only `TrackSock` and `UntrackSock` are supported. Existing policies
   that set `returnArgAction: "Post"` should remove the field.
-
 * `fd` type no longer needs the pre `FollowFD` and post `UnfollowFD` actions to resolve the
   path and works on its own. Existing policies that used `fd` type should be updated to remove the hooks associated with the `FollowFD` and `UnfollowFD` actions. The structure around the use of `fd` hasn't changed and no update should be necessary to the hook that used the type. 
+* Actions `followfd`, `unfollowfd` and `copyfd` are no longer supported.
+  They were previously deprecated and to be removed in tetragon v1.5.
```
kkourt: How about:

* Actions `followfd`, `unfollowfd` and `copyfd` were removed (they were deprecated in v1.5).

kkourt (approved):
LGTM, thanks!

bpf/process/types/basic.h
FedeDP: Exactly, let me add a comment.
Do you prefer an inline comment here or in the commit message?

contrib/upgrade-notes/latest.md
FedeDP: Simpler :) let me push it!

bpf/process/types/basic.h
FedeDP: I chose to expand the commit message.

pkg/selectors/kernel.go
```diff
@@ -53,12 +53,12 @@ const (
...
-	"unfollowfd":                  ActionTypeUnfollowFd,
-	"sigkill":                     ActionTypeSigKill,
-	"override":                    ActionTypeOverride,
-	"copyfd":                      ActionTypeCopyFd,
+	"post": ActionTypePost,
+	// "followfd":                    ActionTypeFollowFd, deprecated
+	// "unfollowfd":                  ActionTypeUnfollowFd, deprecated
+	"sigkill":  ActionTypeSigKill,
+	"override": ActionTypeOverride,
+	// "copyfd":                      ActionTypeCopyFd, deprecated
```
mtardy: I think for the map[uint32]string you can remove the entries safely

bpf/process/types/basic.h
mtardy: > I guess we can't reuse these because they are defined in .proto?

Just noting those are not exactly the same (unfortunately), we have a off by one gap here. But not changing the discussion anyway.

https://github.com/cilium/tetragon/blob/632033a8a61ea31fe4f91d099c6f9efedc11d1ca/api/v1/tetragon/tetragon.proto#L500-L506

docs/security-observability-with-ebpf/README.md
```diff
@@ -2,3 +2,8 @@
 
 This directory contains example TracingPolicies linked from the [O'Reilly report on Security Observability with eBPF](https://www.oreilly.com/library/view/security-observability-with/9781492096719/). 
 These examples were contributed by the report's authors @sharlns and @jedsalazar. 
+
+{{< warning >}}
+Beware, these examples are not production-ready. They were intended to be consumed as a reference for the aforementioned book.
+The examples are not guaranteed to work as-is. Please do not treat them as documentation.
+{{< /warning >}}
```
mtardy: very good to add this: this won't ship in the docs website however, so this is just intended for consumption through the github UI so you can use Github markdown warning for that

```markdown
> [!WARNING]
> Urgent info that needs immediate user attention to avoid problems.
> Beware, these examples are not production-ready. They were intended to be
> consumed as a reference for the aforementioned book. The examples are not
> guaranteed to work as-is. Please do not treat them as documentation.
```

pkg/selectors/kernel_test.go
mtardy, line 1: nit: maybe you can add why in the commit changelog here, because the commit on its own is a bit mysterious

mtardy (commented):
thanks, mostly lgtm :)!

pkg/selectors/kernel.go
FedeDP: Sure! Was leaving them there for long-term memory :) but we can get rid of them, we already have plenty of "THESE WERE DEPRECATED" over the code to remember us!

pkg/selectors/kernel_test.go
FedeDP: Tried my best :) PTAL!

mtardy (approved):
lovely thanks

FedeDP:
Re-bumping CustomResourceDefinitionSchemaVersion after main rebase.

PR #5104, opened 2026-06-04
exporter: fix data race on exporterStartErr in Start()

irozzo-1A:
Fixes 

### Description

The original GetEventsWG called readyWG.Done() on its three early-error paths (filter list build, aggregator creation) and then returned the error. The goroutine in exporter.Start() captured that error and wrote exporterStartErr *after* GetEventsWG returned, but readyWG.Done() may already have unblocked the main goroutine, which reads exporterStartErr concurrently. Under the Go memory model this is a data race; in practice the error is silently lost and the daemon starts with a broken, non-exporting exporter.

### Changelog

```release-note
fix: race condition on GetEventsWG startup error handling
```

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -438,7 +438,7 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 
 	// The "defer cleanupWg.Wait()" above, might introduce deadlocks if an error happens.
 	// This is because cancel() will not be called until cleanupWg.Wait() returns.
-	// But, the code in server/server.go:GetEventsWG() will only call cleanupWg.Done() if ctx.Done()
+	// But, the code in server/server.go:GetEventsListener() will only call cleanupWg.Done() if ctx is done
 	// Which causes a deadlock. To fix this, we add a new ctx and we pass that to the rest of the
 	// initialization functions. This means that we can cancel them without causing a deadlock
 	// using cancel2.
--- a/pkg/exporter/exporter.go
+++ b/pkg/exporter/exporter.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/cilium/lumberjack/v2"
@@ -30,15 +31,18 @@ type ExportEncoder interface {
 }
 
 type Exporter struct {
-	ctx              context.Context
-	request          *tetragon.GetEventsRequest
-	server           *server.Server
-	encoder          ExportEncoder
-	closer           io.Closer
-	rateLimiter      *ratelimit.RateLimiter
+	ctx         context.Context
+	request     *tetragon.GetEventsRequest
+	server      *server.Server
+	encoder     ExportEncoder
+	closer      io.Closer
+	rateLimiter *ratelimit.RateLimiter
+	logFile     string
+	logsDir     string
+
+	// mu protects rotateTimer and rotationInterval for concurrent access.
+	mu               sync.Mutex
 	rotateTimer      *time.Timer
-	logFile          string
-	logsDir          string
 	rotationInterval time.Duration
 }
 
@@ -79,14 +83,17 @@ func NewExporter(
 }
 
 func (e *Exporter) Start() error {
-	// Start the rotation timer if needed
+	// Start the rotation timer if needed. Hold mu to honor the invariant that
+	// rotateTimer and rotationInterval are only accessed while holding it.
+	e.mu.Lock()
 	if e.rotationInterval > 0 {
-		_, ok := e.closer.(*lumberjack.Logger)
-		if !ok {
+		if _, ok := e.closer.(*lumberjack.Logger); !ok {
+			e.mu.Unlock()
 			return fmt.Errorf("writer must be of type lumberjack.Logger but got %T", e.closer)
 		}
 		e.rotateTimer = time.AfterFunc(e.rotationInterval, e.rotate)
 	}
+	e.mu.Unlock()
 
 	run, err := e.server.GetEventsListener(e.request, e, e.closer)
 	if err != nil {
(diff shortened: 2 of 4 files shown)


pkg/server/server.go
```diff
@@ -123,34 +123,41 @@ func (s *Server) removeNotifierAndDrain(l *getEventsListener) {
...
 		"events.allow_list", request.GetAllowList(),
 		"events.deny_list", request.GetDenyList(),
 		"events.field_filters", request.GetFieldFilters(),
 		"events.aggregation_options", request.GetAggregationOptions())
+
+	signalError := func(err error) {
+		if ready != nil {
+			ready <- err
+		}
+	}
```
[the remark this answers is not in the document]
will-isovalent: I don't think this is really a concern in practice. The conditions for this to be problematic are essentially impossible with the current usage afaict.

pkg/exporter/exporter.go
```diff
@@ -89,18 +88,14 @@ func (e *Exporter) Start() error {
...
-	var readyWG sync.WaitGroup
-	var exporterStartErr error
-	readyWG.Add(1)
+	ready := make(chan error, 1)
 	go func() {
-		if err := e.server.GetEventsWG(e.request, e, e.closer, &readyWG); err != nil {
-			exporterStartErr = fmt.Errorf("error starting JSON exporter: %w", err)
-		}
+		e.server.GetEventsWithStartupCh(e.request, e, e.closer, ready)
 	}()
```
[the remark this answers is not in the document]
will-isovalent: I don't think this is actually a regression, based on the old behaviour here. But it still might be worth fixing. @irozzo-1A would you be willing to fix this here?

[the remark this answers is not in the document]
irozzo-1A: yep, sure

will-isovalent (approved):
Thanks!

irozzo-1A:
oups, let me fix the static checks failures 😄

FedeDP (approved):
LGTM, i like the separation between the setup phase and the run function.
The new version is both easier to read and simpler to manage.

(Triggered the CI!)

FedeDP:
> [Tetragon Go Test / build_windows (pull_request)](https://github.com/cilium/tetragon/actions/runs/27335378850/job/81386309752?pr=5104)Failing after 13m

@irozzo-1A  Can you investigate windows CI failures? Thanks!

pkg/exporter/exporter.go
```diff
@@ -107,7 +123,22 @@ func (e *Exporter) rotate() {
...
 		logger.GetLogger().Warn("Failed to rotate JSON export file", "file", option.Config.ExportFilename, logfields.Error, rotationErr)
 	}
-	e.rotateTimer = time.AfterFunc(e.rotationInterval, e.rotate)
+	e.mu.Lock()
+	// Do not reschedule once the context is cancelled, otherwise the timer
+	// keeps rescheduling itself forever even after the exporter is stopped.
+	if e.ctx.Err() == nil {
+		e.rotateTimer = time.AfterFunc(e.rotationInterval, e.rotate)
+	}
+	e.mu.Unlock()
```
mtardy: why not moving the lock/unlock inside the if statement?

pkg/exporter/exporter.go
```diff
@@ -107,7 +123,22 @@ func (e *Exporter) rotate() {
...
+}
+
+// stopRotateTimer stops the self-rescheduling rotation timer, if any.
+func (e *Exporter) stopRotateTimer() {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+	if e.rotateTimer != nil {
+		e.rotateTimer.Stop()
+	}
```
mtardy, line 140: same here

pkg/exporter/exporter.go
mtardy, line 46: could we group and move these three together maybe at the end of the struct?

cmd/tetragon/main.go
```diff
@@ -438,7 +438,7 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 
 	// The "defer cleanupWg.Wait()" above, might introduce deadlocks if an error happens.
 	// This is because cancel() will not be called until cleanupWg.Wait() returns.
-	// But, the code in server/server.go:GetEventsWithReadyChan() will only call cleanupWg.Done() if ctx is done
+	// But, the code in server/server.go:GetEventsListener() will only call cleanupWg.Done() if ctx is done
```
mtardy, line 441: could you squash that in the appropriate commit, see https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/#commit-changes

mtardy (commented):
just a few comments on form, other people seems to be happy with it otherwise, I don't have a lot of experience on the exporter.

pkg/exporter/exporter.go
irozzo-1A: good catch, thanks

pkg/exporter/exporter.go
irozzo-1A: This case is different: the if condition reads `rotateTimer`, which is guarded by `mu`. I think we should keep as is

pkg/exporter/exporter.go
irozzo-1A: Actually it was needed, the check must happen under `mu` so it cannot race with `stopRotateTimer()`: otherwise `rotate()` could observe a live context, get preempted by `stopRotateTimer()`, and then install a new timer that outlives shutdown. Let me revert.

irozzo-1A:
I had to rebase because there was a skew between the level of code used to run the e2e tests (merge ref) and the one used to build the container image (PR head) causing  [tests](https://github.com/cilium/tetragon/actions/runs/28896826237) to fail.

FedeDP (approved):

pkg/exporter/exporter.go
mtardy: not sure to understand why, your lock protects `e.ctx` also?

irozzo-1A: No, not exactly @mtardy. The mutex doesn't protect `e.ctx`, it makes the context cancellation check and the timer re-scheduling atomic with respect to `stopRotateTimer()`. If the lock is moved inside the if, the context check is done outside the lock, so this interleaving becomes possible:
1. rotate() reads `e.ctx.Err() == nil` → true (ctx still live), lock not yet held
2. the context is cancelled, `stopRotateTimer()` acquires the lock and calls `e.rotateTimer.Stop()`
3. `rotate()` finally acquires the lock and assigns a fresh timer to `e.rotateTimer`

That timer now outlives shutdown. Minor in practice, but this makes the shutdown deterministic at very small cost. Let me know if you want me to clarify it with a comment or if you have other suggestions.

mtardy (approved):
thanks!

PR #5106, opened 2026-06-05
pkg/process: use atomic for color field in ProcessInternal

### LINKED ISSUE #5089: pkg/process: use atomic for color field in ProcessInternal
### Description
  The color field in the ProcessInternal struct is currently used to track the garbage collection state of a process. While it is primarily accessed within the cacheGarbageCollector loop, recent reviews in PR #5057 highlighted that it is mutated without explicit synchronization, which could lead to race conditions or visibility issues across goroutines.
### Proposed Changes
   - Refactor the color field in ProcessInternal (defined in pkg/process/process.go) from int to an atomic type (e.g., atomic.Int32 or atomic.Uint32).
  - Update all sites that access or modify p.color (primarily in pkg/process/cache.go) to use Load() and Store().
  - Ensure the color constants (inUse, deletePending, etc.) are compatible with the new atomic type.
  ### Context
  This was identified as a follow-up item during the review of PR #5057.

arpitjain099:
## Description

The `color` field in `ProcessInternal` (`pkg/process/process.go`) tracks the garbage collection state of a process. It is written by the `cacheGarbageCollector` goroutine, but it is also read concurrently from other goroutines:

- `dump()` and `getEntries()` read `v.color` when building the API response for a process-cache dump.
- the LRU eviction callback in `NewCache` reads `evicted.color` to decide whether to perform the parent refcount decrement.

None of this access was synchronized. This was raised during the review of #5057: once GC could transition a process back to `inUse` (the fix in that PR), the unsynchronized access became a correctness concern, not just a memory-visibility one, because eviction can incorrectly skip or apply parent refcount adjustments depending on a torn or stale read of `color`. Reviewers (@kevsecurity, @tpapagian, @will-isovalent) agreed that switching `color` to an atomic in a follow-up made sense, and that follow-up is tracked in #5089.

## Changes

- Store `color` as `atomic.Int32` in `ProcessInternal` and route every read/write through two small helpers, `getColor()` and `setColor()`, so all access is synchronized. The GC state machine and its transitions are unchanged.
- Introduce a named `processColor` type for the state constants (`inUse`, `deletePending`, `deleteReady`, `deleted`) so the colors are constrained to the valid set instead of plain `int`. This also picks up the suggestion in #5089 to make the color enum a real type.

All `color` access sites were updated: the GC loop in `cacheGarbageCollector` (the `<-ticker.C` and `<-pc.deleteChan` cases), the LRU eviction callback, and the `dump()` / `getEntries()` read paths.

## Testing

Added `TestProcessCacheColorDataRace`, which runs the GC writing `color` (driven by a refcnt-churn goroutine) concurrently with a goroutine reading `color` through `getEntries()`, the same read path used by `dump()`.

Before the change, the race detector flags the conflict:

```
WARNING: DATA RACE
Write at 0x... by goroutine 27:
  pkg/process.(*Cache).cacheGarbageCollector.func1()
      pkg/process/cache.go:110

Previous read at 0x... by goroutine 28:
  pkg/process.(*Cache).getEntries()
      pkg/process/cache.go:273
```

After the change:

```
$ go test -race -count=1 ./pkg/process/...
ok  	github.com/cilium/tetragon/pkg/process
```

`go test -race ./pkg/process/...` is clean, `go build ./...` passes, and `go vet ./pkg/process/...` is clean. Reverting only the source change (keeping the new test) makes the race reappear, which confirms the test guards against a regression.

Fixes: #5089

### DIFF
--- a/pkg/process/cache.go
+++ b/pkg/process/cache.go
@@ -28,15 +28,19 @@ type Cache struct {
 	stopChan   chan bool
 }
 
+// processColor tracks the garbage collection state of a process. It is stored
+// in ProcessInternal as an atomic and accessed via getColor/setColor.
+type processColor int32
+
 // garbage collection states
 const (
-	inUse = iota
+	inUse processColor = iota
 	deletePending
 	deleteReady
 	deleted
 )
 
-var colorStr = map[int]string{
+var colorStr = map[processColor]string{
 	inUse:         "inUse",
 	deletePending: "deletePending",
 	deleteReady:   "deleteReady",
@@ -77,15 +81,15 @@ func (pc *Cache) cacheGarbageCollector(intervalGC time.Duration) {
 					 * process a second time, but that is harmless.
 					 */
 					if p.refcnt.Load() != 0 {
-						p.color = inUse
+						p.setColor(inUse)
 						continue
 					}
-					if p.color == deleteReady {
-						p.color = deleted
+					if p.getColor() == deleteReady {
+						p.setColor(deleted)
 						pc.remove(p.process)
 					} else {
 						newQueue = append(newQueue, p)
-						p.color = deleteReady
+						p.setColor(deleteReady)
 					}
 				}
 				deleteQueue = newQueue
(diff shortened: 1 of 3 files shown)


kevsecurity:
Looks good. I'll review when it goes green.

kevsecurity (changes requested):
Looks like just the linter errors. Otherwise LGTM.

arpitjain099:
Fixed the linter findings - the modernize linter flagged the test, so I switched the setup loop to range-over-int and the two goroutines to `WaitGroup.Go`. Verified with `gofmt` and `GOOS=linux go vet ./pkg/process/`. Thanks @kevsecurity!

arpitjain099:
> Fixed the linter findings - the modernize linter flagged the test, so I switched the setup loop to range-over-int and the two goroutines to `WaitGroup.Go`. Verified with `gofmt` and `GOOS=linux go vet ./pkg/process/`. Thanks @kevsecurity!

Please review this PR when you can @kevsecurity

tpapagian (approved):
Thanks!

kevsecurity (approved):

kkourt:
@tpapagian @kevsecurity I'm marking this for backports, similarly to https://github.com/cilium/tetragon/pull/5057. Please let me know otherwise. Thanks!

PR #5109, opened 2026-06-05
sensors: add support for fd type to retrieve file*

litios:
Currently, `fd` type would only be populated through the deprecated FollowFD feature. 
We can instead resolve the `fd` directly against the current_task (`files->fdt->fd[n]`) and get the file* pointer.

Add tests for the feature too.

Example event generated:

```yaml
{
  "process_kprobe": {
    "process": {
      "exec_id": "dWJ1bnR1MjQwNDoxNjQ4MjkxNTAwMDAwMDA6MzIxNzQ=",
      "pid": 32174,
      "uid": 1000,
      "cwd": "/home/litios",
      "binary": "/usr/bin/zsh",
      "flags": "procFS",
      "start_time": "2026-06-05T09:17:06.837726801Z",
      "auid": 1000,
      "parent_exec_id": "dWJ1bnR1MjQwNDoxNjQ4MjkxMDAwMDAwMDA6MzIxNzM=",
      "refcnt": 1,
      "tid": 32174,
      "in_init_tree": false
    },
    "parent": {
      "exec_id": "dWJ1bnR1MjQwNDoxNjQ4MjkxMDAwMDAwMDA6MzIxNzM=",
      "pid": 32173,
      "uid": 1000,
      "cwd": "/",
      "binary": "/usr/sbin/sshd",
      "flags": "procFS rootcwd",
      "start_time": "2026-06-05T09:17:06.787726790Z",
      "auid": 1000,
      "parent_exec_id": "dWJ1bnR1MjQwNDoxNjQ4MjgwNzAwMDAwMDA6MzIwNzQ=",
      "tid": 32173,
      "in_init_tree": false
    },
    "function_name": "__x64_sys_write",
    "args": [
      {
        "file_arg": {
          "path": "/tmp/target",
          "permission": "-rw-rw-r--"
        }
      },
      {
        "bytes_arg": "QUFvQQo="
      },
      {
        "size_arg": "5"
      }
    ],
    "action": "KPROBE_ACTION_POST",
    "policy_name": "kprobe-fd-arg-live-file",
    "return_action": "KPROBE_ACTION_POST"
  },
  "node_name": "ubuntu2404",
  "time": "2026-06-05T09:36:41.053843626Z"
}
```

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -251,7 +251,6 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 {
 	struct msg_generic_kprobe *e;
 	long size = -1;
-	long ret = 0;
 	int zero = 0;
 
 	e = map_lookup_elem(&process_call_heap, &zero);
@@ -262,36 +261,6 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 	case iov_iter_type:
 		size = copy_iov_iter(ctx, orig_off, arg, argm, e);
 		break;
-	case fd_ty: {
-		struct fdinstall_key key = { 0 };
-		struct fdinstall_value *val;
-
-		key.tid = get_current_pid_tgid() >> 32;
-		key.fd = arg;
-
-		val = map_lookup_elem(&fdinstall_map, &key);
-		if (val) {
-			__u32 bytes = *((__u32 *)&val->file[0]);
-
-			memcpy(&args[0], &key.fd, sizeof(key.fd));
-			asm volatile("%[bytes] &= 0xfff;\n"
-				     : [bytes] "+r"(bytes)
-				     :);
-			// size + file path + flags
-			size = 4 + bytes + 4;
-			ret = probe_read(&args[4], size, (char *)&val->file[0]);
-			if (ret < 0)
-				return ret;
-
-			// account for fd written at args[0]
-			size += 4;
-		} else {
-			/* If filter specification is fd type then we
-			 * prevent the filter from matching
-			 */
-			return -1;
-		}
-	} break;
 	case filename_ty: {
 		struct filename *file;
 
@@ -502,9 +471,9 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
 	}
 
 	if (ret < 0) {
-		/* fd_ty returns -1 when the fd is not in the fdinstall_map,
-		 * meaning the filter should not match (event should be dropped).
-		 * This is a filter decision, not a fault.
+		/* fd_ty returns -1 when the fd cannot be resolved, meaning the
+		 * filter should not match (event should be dropped). This is a
+		 * filter decision, not a fault.
 		 */
 		if (type == fd_ty)
 			return -1;
--- a/bpf/process/generic_path.h
+++ b/bpf/process/generic_path.h
@@ -227,6 +227,7 @@ FUNC_INLINE bool should_offload_path(long type)
 	case path_ty:
 	case dentry_type:
 	case linux_binprm_type:
+	case fd_ty:
 		return true;
 	}
 	return false;
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -2238,9 +2238,6 @@ filter_arg_2(struct msg_generic_kprobe *e, struct selector_arg_filter *filter, c
 {
 	switch (filter->type) {
 	case fd_ty:
-		/* Advance args past fd */
-		args += 4;
-		fallthrough;
 	case file_ty:
 	case path_ty:
 	case dentry_type:
@@ -2766,6 +2763,38 @@ FUNC_INLINE void path_from_dentry(struct dentry *dentry, struct path *path_buf)
 	path_buf->dentry = dentry;
 }
 
+FUNC_INLINE const struct path *path_from_fd(unsigned long fd)
+{
+	struct task_struct *task = (struct task_struct *)get_current_task();
+	struct files_struct *files = NULL;
+	struct fdtable *fdt = NULL;
+	struct file **fd_array = NULL;
+	struct file *file = NULL;
+	__u32 fd_idx;
+	__u32 max_fds;
+
+	if (fd > 0xffffffff)
+		return NULL;
+	fd_idx = (__u32)fd;
+
+	if (BPF_CORE_READ_INTO(&files, task, files) != 0 || !files)
+		return NULL;
+	if (BPF_CORE_READ_INTO(&fdt, files, fdt) != 0 || !fdt)
+		return NULL;
+	if (BPF_CORE_READ_INTO(&max_fds, fdt, max_fds) != 0)
+		return NULL;
+	if (fd_idx >= max_fds)
+		return NULL;
+	if (BPF_CORE_READ_INTO(&fd_array, fdt, fd) != 0 || !fd_array)
+		return NULL;
+
+	probe_read(&file, sizeof(file), &fd_array[fd_idx]);
+	if (!file)
+		return NULL;
+
+	return _(&file->f_path);
+}
+
 FUNC_INLINE const struct path *get_path(long type, unsigned long arg, struct path *path_buf)
 {
 	const struct path *path_arg = 0;
@@ -2789,6 +2818,9 @@ FUNC_INLINE const struct path *get_path(long type, unsigned long arg, struct pat
 	case dentry_type:
 		path_from_dentry((struct dentry *)arg, path_buf);
 		path_arg = path_buf;
+		break;
+	case fd_ty:
+		path_arg = path_from_fd(arg);
 		break;
 #ifdef __LARGE_BPF_PROG
 	case linux_binprm_type: {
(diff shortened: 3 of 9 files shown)


FedeDP:
Thanks for this PR!
I bet there is a reason why we decided to go with the `FollowFD` approach instead, but worth a try! @olsajiri perhaps has more context :) 

Seems like multiple uprobe related vmtests are failing now :(
```
    observer_test_helper.go:449: SensorManager.AddTracingPolicy error: sensor generic_uprobe from collection uprobe-generic failed to load: failed prog /home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/bpf/objs/bpf_generic_uprobe.o kern_version 267263 loadInstance: opening collection '/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/bpf/objs/bpf_generic_uprobe.o' failed: program generic_uprobe_process_event: load program: argument list too long
```

Also, can you rebase the PR and fix the `golangci-lint` issues? 
Thank you so much!

bpf/process/generic_calls.h
```diff
@@ -263,35 +292,18 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
...
-				return ret;
-			size += 4;
+		struct file *file = get_current_file_from_fd(arg);
+
+		if (file) {
+			const struct path *path = _(&file->f_path);
+			__u32 fd = (__u32)arg;
+
+			memcpy(&args[0], &fd, sizeof(fd));
+			ret = copy_path(&args[4], path);
```
olsajiri: I wonder it'd be better to move this `read_arg` and somehow treat it together with `path_arg`, and keep just single `copy_path` invocation, because it's expensive

olsajiri (commented):
> I bet there is a reason why we decided to go with the `FollowFD` approach instead, but worth a try! @olsajiri perhaps has more context :)

it happened before me ;-) @kkourt ?
maybe it's bit more reliable than reading fdtable, but it's tracing, so best effort 

at the first glance the change seems ok, could you also remove that fdinstall* machinery?

FedeDP:
>  could you also remove that fdinstall* machinery?

If you are talking about `followfd` related actions (and related `fdinstall` maps and types), it will be done in https://github.com/cilium/tetragon/pull/5090 once this is cleared :)

bpf/process/generic_calls.h
litios: Indeed, that makes a lot more sense, added there and everything working fine! Ty

litios:
Sorry it took a minute to fix the pipeline issues, rebased from main, fixed lint issues and applied @olsajiri suggestion!
i'll see if the uprobe issues show again

FedeDP:
Windows CI is fixed by https://github.com/cilium/tetragon/pull/5118; the vmtests failures must be investigated one by one.
For example, `pkg.sensors.tracing.TestTracepointListSyscallDups` fails with:
```
    logcapture.go:24: time=2026-06-10T15:37:58.900Z level=INFO msg="Listening for events..."
    jsonchecker.go:179: jsonTestCheck: opening: /tmp/tetragon.gotest.TestTracepointListSyscallDups.3869185167.json
    jsonchecker.go:130: JsonCheck (retry=0) took 12.433µs
    jsonchecker.go:154: JsonCheck (retry=1) failed: JsonEOF: failed to match after 0 events: err:UnorderedEventChecker: 3/3 checks remain. Retrying after 3s
    jsonchecker.go:130: JsonCheck (retry=1) took 28.934µs
    tracepoint_amd64_test.go:46: 
        	Error Trace:	/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/sensors/tracing/tracepoint_amd64_test.go:46
        	            				/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/pkg/sensors/tracing/tracepoint_amd64_test.go:124
        	            	JsonTestCheck failed in retry cnt=1 and there were no new events from previous try: JsonEOF: failed to match after 0 events: err:UnorderedEventChecker: 3/3 checks remain
        	Test:       	TestTracepointListSyscallDups
    filenames.go:49: keeping export file for TestTracepointListSyscallDups (/tmp/tetragon.gotest.TestTracepointListSyscallDups.3869185167.json)
    logcapture.go:24: time=2026-06-10T15:38:01.911Z level=INFO msg="Listening for events completed." error="context canceled"
```

One thing i noted is that only "old" kernels fail while new ones (6.x+) pass fine.

FedeDP:
Uh i just saw same issues on main CI: https://github.com/cilium/tetragon/actions/runs/27328812117/job/80737366990 but only on 5.15.

FedeDP:
>  failed to match after 0 events

It seems like we are not able to get any events when the issue happens :/

kkourt:
> > I bet there is a reason why we decided to go with the `FollowFD` approach instead, but worth a try! @olsajiri perhaps has more context :)
> 
> it happened before me ;-) @kkourt ? maybe it's bit more reliable than reading fdtable, but it's tracing, so best effort

I don't think we ever consider iterating the fd table as in the current PR.

bpf/process/generic_calls.h
```diff
@@ -502,9 +471,9 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
...
 
 	if (ret < 0) {
-		/* fd_ty returns -1 when the fd is not in the fdinstall_map,
-		 * meaning the filter should not match (event should be dropped).
-		 * This is a filter decision, not a fault.
+		/* fd_ty returns -1 when the fd cannot be resolved, meaning the
+		 * filter should not match (event should be dropped). This is a
+		 * filter decision, not a fault.
 		 */
 		if (type == fd_ty)
```
litios, line 478: @FedeDP this is a remaining piece of the followfd feature, in case you want to change this decision. I didn't want to delay the PR more since you will handle the actual removal of FollowFD, we can have the discussion in that new PR since this is just a design decision.

FedeDP (approved):

litios:
On this last commit, I removed completely the `int` push before the path for `fd` because it's no longer needed (was used by the followFD) and simplifies the `fd` type to behave like any of the other path-like types.

litios:
@olsajiri @kkourt soft ping, just in case this fell through the cracks :smiley:

kkourt (commented):
Thanks!

I feel this changes the semantics of the fd type to some extent. Do we think it's a good idea to rename the type? If not, let's add something to the upgrade notes. Also, I feel it's worth updating the docs.

As a minor note, we have been using "type" to say things like, "give me the filename of this struct file", or, in this case, "give the filename of this file descriptor", which seems strange to me because we are not defining the type of the argument, but, instead, we define what information to get out of it. I don't think we should address this in this PR, but might be an interesting topic to think about.

litios:
> I feel this changes the semantics of the fd type to some extent. Do we think it's a good idea to rename the type? If not, let's add something to the upgrade notes. .

That's a good point, in the sense that `fd` is not really the type of the argument. Maybe we can rename to `fd_file`?
The old `fd` type will _disappear_ with followfd feature being deprecated so that would leave us with only `fd_file`.

> Also, I feel it's worth updating the docs

I feel that the current description matches more the new `fd_file` type than anything else :laughing: 

`The fd data type represents a file descriptor. Tetragon resolves the file descriptor to retrieve the associated file path.`

If we agree with the name change, maybe I can just update the the existing `fd` docs to point to `fd_file`

FedeDP:
> That's a good point, in the sense that fd is not really the type of the argument. Maybe we can rename to fd_file?
The old fd type will disappear with followfd feature being deprecated so that would leave us with only fd_file.

Since `fd_ty` is basically already deprected and was to be removed, i don't feel the semantic change is really problematic after all.
I think it's better to keep `fd_ty` (whose description, i agree with David, is already ok with the new implementation) with the new semantic and kill with :fire: the old `followfd` related actions.

I agree with Kornilios to add an item to upgrade notes though :)

litios:
If it's worth anything, my personal preference is to keep it with the existing type `fd` too, mainly because it follows what we said `fd` was doing already, just without the `FollowFD`/`UnfollowFD` bits. The part of the policy that used the `fd` is exactly the same before and after this change:

```yaml
kprobes:
- call: "fd_install"
  syscall: false
  args:
  - index: 0
    type: int
  - index: 1
    type: "file"
  selectors:
  - matchArgs:
    - index: 1
      operator: "Equal"
      values:
      - "/tmp/passwd"
    matchActions:
    - action: FollowFD
      argFd: 0
      argName: 1
- call: "sys_write"
  syscall: true
  args:
  - index: 0
    type: "fd"
  - index: 1
    type: "char_buf"
    sizeArgIndex: 3
  - index: 2
    type: "size_t"
  selectors:
  - matchArgs:
    - index: 0
      operator: "Equal"
      values:
      - "/tmp/passwd"
    matchActions:
    - action: Sigkill
- call: "sys_close"
  syscall: true
  args:
  - index: 0
     type: "int"
  selectors:
  - matchActions:
    - action: UnfollowFD
      argFd: 0
```

litios:
Added a small comment to the upgrade notes

olsajiri:
> ... with the new semantic and kill with 🔥 the old followfd related actions.
I'd keep fd_ty with new semantics as well.. @litios could we remove the followfd stuff in this pr or somebody else is doing that elsewhere?

litios:
@FedeDP is taking care of deprecating FollowFD after this PR :)

FedeDP:
Yes @olsajiri , will rebase https://github.com/cilium/tetragon/pull/5090 as soon as this is merged :)

litios:
@olsajiri / @kkourt any issues you see with it or should we merge? :eyes:
To avoid delaying @FedeDP followup PR any further

olsajiri (approved):

PR #5114, opened 2026-06-08
lsm: fixing bug with multiple lsm hook entries

### LINKED ISSUE #4578: Duplicate LSM hook entries cause event amplification while only the last entry is applied
### What happened? ###

When defining multiple `lsmhooks` entries with the same `hook` (`file_open`) in a single `TracingPolicy`, Tetragon posts events N times per matching event (where N is the number of duplicate `lsmhooks` entries), but only the selector from the final entry is applied.

**Example:** With the 3 duplicate `file_open` entries provided below:
- Opening `/tmp/foo` or `/tmp/bar` triggers no events.
- Opening `/tmp/baz` triggers 3 duplicate Post events.
- If a 4th entry were added, it would trigger 4 Post events for `/tmp/baz`.

```yaml
kind: TracingPolicy
metadata:
  name: "lsm-conflict-repro"
spec:
  lsmhooks:
    - hook: file_open
      args:
        - index: 0
          type: file
      selectors:
        - matchArgs:
            - args: [0]
              operator: Equal
              values: [/tmp/foo]
          matchActions:
            - action: Post
    - hook: file_open
      args:
        - index: 0
          type: file
      selectors:
        - matchArgs:
            - args: [0]
              operator: Equal
              values: [/tmp/bar]
          matchActions:
            - action: Post
    - hook: file_open
      args:
        - index: 0
          type: file
      selectors:
        - matchArgs:
            - args: [0]
              operator: Equal
              values: [/tmp/baz]
          matchActions:
            - action: Post

```

### What I expected to happen ###

One of the following would be acceptable, but not the current behavior:

- Coexistence: each `lsmhooks` entry behaves independently (no “last selector wins”, and no duplication), or
- Fail fast: policy is rejected with a validation error explaining duplicate LSM hooks in one policy are unsupported.

### Tetragon Version

`v1.6.0`

### Kernel Version

`6.18.6`

### Kubernetes Version

_No response_

### Bugtool

_No response_

### Relevant log output

```shell

```

### Related Context

There have been prior discussions and fixes regarding duplicate-instance handling for kprobes (#2947 / PR #3121). This report highlights a similar issue specifically for LSM hooks, which additionally includes event amplification proportional to the number of entries. I have not yet confirmed if this also affects `tracepoints`.

### Motivation

Currently, selectors are limited to 5 per hook. In production environments, it is common to require more than 5 independent match conditions for high-volume hooks. While splitting rules across multiple `TracingPolicy` objects is a viable workaround, being able to "shard" conditions across multiple entries in a single policy would significantly improve manageability.

bar234:
Fixing bug with multiple lsm hook entries

Fixes #4578

### Description
This commit fixes a bug where events were not being emitted when multiple entries were configured for the same lsm hook

### Changelog

```
lsm: fix bug with duplicate hooked functions
```

### DIFF
--- a/pkg/sensors/tracing/genericlsm.go
+++ b/pkg/sensors/tracing/genericlsm.go
@@ -74,6 +74,8 @@ type genericLsm struct {
 	tags []string
 	// is IMA hash collector program needed to load
 	imaProgLoad bool
+	// instance identifier to mitigate duplicate hooks at same location
+	instance int
 }
 
 func (g *genericLsm) SetID(id idtable.EntryID) {
@@ -290,7 +292,7 @@ type addLsmIn struct {
 	selectorStatsBase uint32
 }
 
-func addLsm(f *v1alpha1.LsmHookSpec, in *addLsmIn) (id idtable.EntryID, err error) {
+func addLsm(f *v1alpha1.LsmHookSpec, instance int, in *addLsmIn) (id idtable.EntryID, err error) {
 	var argSigPrinters []argPrinter
 	var argsBTFSet [api.MaxArgsSupported]bool
 	var allBTFArgs [api.EventConfigMaxArgs][api.MaxBTFArgDepth]api.ConfigBTFArg
@@ -379,6 +381,7 @@ func addLsm(f *v1alpha1.LsmHookSpec, in *addLsmIn) (id idtable.EntryID, err erro
 		message:     msgField,
 		tags:        tagsField,
 		imaProgLoad: false,
+		instance:    instance,
 	}
 
 	for _, sel := range f.Selectors {
@@ -434,6 +437,7 @@ func createGenericLsmSensor(
 	}
 
 	var selectorStatsBase uint32
+	dups := make(map[string]int)
 	for _, hook := range lsmHooks {
 		if err := appendMacrosSelectors(hook.Selectors, spec.SelectorsMacros); err != nil {
 			return nil, fmt.Errorf("append macros selectors: %w", err)
@@ -442,7 +446,13 @@ func createGenericLsmSensor(
 		in.selectorStatsBase = selectorStatsBase
 		selectorStatsBase += uint32(len(hook.Selectors))
 
-		id, err := addLsm(&hook, &in)
+		instance, ok := dups[hook.Hook]
+		if ok {
+			instance++
+		}
+		dups[hook.Hook] = instance
+
+		id, err := addLsm(&hook, instance, &in)
 		if err != nil {
 			return nil, err
 		}
@@ -537,6 +547,10 @@ func createLsmSensorFromEntry(polInfo *policyInfo, lsmEntry *genericLsm,
 	progs []*program.Program, maps []*program.Map) ([]*program.Program, []*program.Map) {
 
 	loadProgCoreName, loadProgOutputName := config.GenericLsmObjs()
+	pinName := lsmEntry.hook
+	if lsmEntry.instance != 0 {
+		pinName = fmt.Sprintf("%s:%d", lsmEntry.hook, lsmEntry.instance)
+	}
 
 	/* We need to load LSM programs in the following order:
 	   1. bpf_generic_lsm_output
@@ -547,7 +561,7 @@ func createLsmSensorFromEntry(polInfo *policyInfo, lsmEntry *genericLsm,
 		path.Join(option.Config.HubbleLib, loadProgOutputName),
 		lsmEntry.hook,
 		"lsm/generic_lsm_output",
-		lsmEntry.hook,
+		pinName,
 		"generic_lsm").
 		SetLoaderData(lsmEntry.tableId).
 		SetPolicy(lsmEntry.policyName)
@@ -557,7 +571,7 @@ func createLsmSensorFromEntry(polInfo *policyInfo, lsmEntry *genericLsm,
 		path.Join(option.Config.HubbleLib, loadProgCoreName),
 		lsmEntry.hook,
 		"lsm/generic_lsm_core",
-		lsmEntry.hook,
+		pinName,
 		"generic_lsm").
 		SetLoaderData(lsmEntry.tableId).
 		SetPolicy(lsmEntry.policyName)
@@ -571,7 +585,7 @@ func createLsmSensorFromEntry(polInfo *policyInfo, lsmEntry *genericLsm,
 				path.Join(option.Config.HubbleLib, loadProgImaName),
 				lsmEntry.hook,
 				"lsm.s/generic_lsm_ima_"+loadProgImaType,
-				lsmEntry.hook,
+				pinName,
 				"generic_lsm").
 				SetLoaderData(lsmEntry.tableId).
 				SetPolicy(lsmEntry.policyName)
--- a/pkg/sensors/tracing/lsm_test.go
+++ b/pkg/sensors/tracing/lsm_test.go
@@ -32,6 +32,7 @@ import (
 	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/sensors"
 	"github.com/cilium/tetragon/pkg/testutils"
+	"github.com/cilium/tetragon/pkg/testutils/policytest"
 	tus "github.com/cilium/tetragon/pkg/testutils/sensors"
 )
 
@@ -453,4 +454,8 @@ spec:
 		return true
 	}
 	require.Condition(t, checkFunc)
+}
+
+func TestLSMDuplicateHooks(t *testing.T) {
+	policytest.AllPolicyTests.DoObserverTest(t, "lsm-dup-hooks", nil)
 }
--- a/pkg/testutils/policytest/builder.go
+++ b/pkg/testutils/policytest/builder.go
@@ -39,17 +39,31 @@ func (b *Builder) WithParameter(p Parameter) *Builder {
 //
 // In the template, the following functions are supported
 //   - testBinary: generate a test binary path from the binary name (Conf.TestBinary())
+//   - tempFile: generate a temporary file and save it in the conf
 func (b *Builder) WithPolicyTemplate(tmpl string) *Builder {
 	policyTest := b.policytest
-	policyTest.Policy = func(c *Conf) (Policy, error) {
+	policyTest.Policy = func(c *Conf) (Policy, PolicyCleanupFn, error) {
 		funcMap := template.FuncMap{
 			"testBinary": func(s string) string {
 				return c.TestBinary(s)
 			},
+			"tempFile": func(key string) (string, error) {
+				return c.TempFile(key)
+			},
 		}
+
 		t, err := template.New("testpolicy").Funcs(funcMap).Parse(tmpl)
+
+		// Clean up any temp files created if this function returns due to an
+		// error
+		defer func() {
+			if err != nil {
+				c.CleanupTempFiles()
+			}
+		}()
+
 		if err != nil {
-			return Policy(""), fmt.Errorf("failed to parse template: %w", err)
+			return Policy(""), nil, fmt.Errorf("failed to parse template: %w", err)
 		}
 
 		// fill in params
@@ -65,9 +79,9 @@ func (b *Builder) WithPolicyTemplate(tmpl string) *Builder {
 		var buf bytes.Buffer
 		err = t.Execute(&buf, params)
 		if err != nil {
-			return Policy(""), fmt.Errorf("failed to execute template: %w", err)
+			return Policy(""), nil, fmt.Errorf("failed to execute template: %w", err)
 		}
-		return Policy(buf.String()), nil
+		return Policy(buf.String()), c.CleanupTempFiles, nil
 	}
 	return b
 }
--- a/pkg/testutils/policytest/conf.go
+++ b/pkg/testutils/policytest/conf.go
@@ -7,8 +7,10 @@ package policytest
 
 import (
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 )
 
 type ParamVals map[string]any
@@ -43,8 +45,54 @@ type Conf struct {
 
 	// Path to save the generated policy
 	DumpPolicyPath string
+
+	// Map of temporary files generated for this generated policy. Key is
+	// unique identifier for the temp file, value is the generated temp
+	// filename
+	tempFiles   map[string]string
+	tempFilesMu sync.Mutex
 }
 
 func (c *Conf) TestBinary(s string) string {
 	return filepath.Join(c.BinsDir, s)
+}
+
+func (c *Conf) TempFile(key string) (string, error) {
+	c.tempFilesMu.Lock()
+	defer c.tempFilesMu.Unlock()
+
+	if c.tempFiles == nil {
+		c.tempFiles = make(map[string]string)
+	} else if path, ok := c.tempFiles[key]; ok {
+		return path, nil // already created — return cached path
+	}
+	tempFile, err := os.CreateTemp("", "tetragon-testfile-*")
+	if err != nil {
+		return "", err
+	}
+	tempFile.Close()
+	path := tempFile.Name()
+	c.tempFiles[key] = path
+	return path, nil
+}
+
+func (c *Conf) TempFileMust(key string) string {
+	ret, err := c.TempFile(key)
+	if err != nil {
+		panic(err)
+	}
+	return ret
+}
+
+func (c *Conf) CleanupTempFiles() {
+	c.tempFilesMu.Lock()
+	defer c.tempFilesMu.Unlock()
+
+	if c.tempFiles == nil {
+		return
+	}
+	for _, tempFile := range c.tempFiles {
+		os.Remove(tempFile)
+	}
+	c.tempFiles = nil
 }
--- a/pkg/testutils/policytest/observertest.go
+++ b/pkg/testutils/policytest/observertest.go
@@ -58,10 +58,11 @@ func (rpt *RegisteredPolicyTests) DoObserverTest(
 			ParamValues: params,
 		},
 	}
-	policyStr, err := pt.Policy(conf)
+	policyStr, cleanupFn, err := pt.Policy(conf)
 	if err != nil {
 		t.Fatalf("failed to generate policy: %s", err)
 	}
+	t.Cleanup(cleanupFn)
 
 	policyFile, err := os.CreateTemp("", "tetragon-policy-"+t.Name()+"-*.txt")
 	if err != nil {
--- a/pkg/testutils/policytest/runner.go
+++ b/pkg/testutils/policytest/runner.go
@@ -115,11 +115,20 @@ func (r *LocalRunner) Close() {
 
 func (r *LocalRunner) AddPolicy(l *slog.Logger, test *T) (*PolicyHandler, error) {
 	// generate policy
-	pol, err := test.Policy(r.conf)
+	pol, cleanup, err := test.Policy(r.conf)
 	if err != nil {
 		err = fmt.Errorf("failed to create policy for test %q: %w", test.Name, err)
 		return nil, err
 	}
+
+	// call cleanup if we're returning due to an error. Note that this won't
+	// run if the test doesn't have a policy but there should be nothing to
+	// clean up in this case
+	defer func() {
+		if err != nil {
+			cleanup()
+		}
+	}()
 
 	// allow for tests that do not have a policy
 	if len(pol) == 0 {
@@ -154,6 +163,7 @@ func (r *LocalRunner) AddPolicy(l *slog.Logger, test *T) (*PolicyHandler, error)
 	return &PolicyHandler{
 		tpName:      tpName,
 		tpNamespace: "", // TODO: change this when we add support for namespaced policies
+		cleanup:     cleanup,
 	}, nil
 }
 
@@ -385,9 +395,15 @@ func runFwd(
 type PolicyHandler struct {
 	tpName      string
 	tpNamespace string
+	cleanup     PolicyCleanupFn
 }
 
 func (ph *PolicyHandler) Cleanup(l *slog.Logger, conf *Conf, client *cli.ClientWithContext) error {
+
+	// call policy cleanup after we are done
+	if ph.cleanup != nil {
+		defer ph.cleanup()
+	}
 
 	_, err := client.Client.DeleteTracingPolicy(client.Ctx, &tetragon.DeleteTracingPolicyRequest{
 		Name:      ph.tpName,
--- a/cmd/tetra/policytest/policytest.go
+++ b/cmd/tetra/policytest/policytest.go
@@ -104,11 +104,12 @@ func dumpPolicyCmd() *cobra.Command {
 			}
 
 			for _, t := range tests {
-				pol, err := t.Policy(&conf)
+				pol, cleanup, err := t.Policy(&conf)
 				if err != nil {
 					return err
 				}
 				cmd.OutOrStdout().Write([]byte(pol))
+				cleanup()
 			}
 			return nil
 		},
--- a/pkg/testutils/policytest/policytest.go
+++ b/pkg/testutils/policytest/policytest.go
@@ -31,6 +31,8 @@ type Scenario struct {
 // Policies are represented as strings, because that's how they are loaded via gRPC
 type Policy string
 
+type PolicyCleanupFn func()
+
 type Label string
 
 type SkipInfo struct {
@@ -69,7 +71,7 @@ type T struct {
 	ShouldSkip func(info *SkipInfo) string
 
 	// Policy generates a policy for this test
-	Policy func(c *Conf) (Policy, error)
+	Policy func(c *Conf) (Policy, PolicyCleanupFn, error)
 
 	Params []Parameter
 
--- a/tests/policytests/lsm.go
+++ b/tests/policytests/lsm.go
@@ -1,0 +1,96 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build !windows
+
+package tests
+
+import (
+	"os/exec"
+
+	"github.com/cilium/tetragon/pkg/bpf"
+
+	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
+	lc "github.com/cilium/tetragon/pkg/matchers/listmatcher"
+	sm "github.com/cilium/tetragon/pkg/matchers/stringmatcher"
+	"github.com/cilium/tetragon/pkg/testutils/policytest"
+)
+
+// This file contains tests on lsm hooks
+
+var _ = policytest.NewBuilder("lsm-dup-hooks").WithLabels("lsm").WithPolicyTemplate(`
+apiVersion: cilium.io/v1alpha1
+kind: TracingPolicy
+metadata:
+  name: "lsm-dup-hooks"
+spec:
+  lsmhooks:
+  - hook: "file_open"
+    args:
+      - index: 0
+        type: "file"
+    selectors:
+    - matchArgs:
+      - args: [0]
+        operator: "Equal"
+        values: [{{ tempFile "file1"}}]
+      matchActions:
+        - action: Post
+  - hook: "file_open"
+    args:
+      - index: 0
+        type: "file"
+    selectors:
+    - matchArgs:
+      - args: [0]
+        operator: "Equal"
+        values: [{{ tempFile "file2" }}]
+      matchActions:
+        - action: Post
+`).WithSkip(func(si *policytest.SkipInfo) string {
+	if !si.AgentInfo.Probes[bpf.LargeProgsProbe] {
+		return "need 5.3 or newer kernel"
+	}
+	if !si.AgentInfo.Probes[bpf.LsmProbe] {
+		return "Need LSM Support"
+	}
+	return ""
+}).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
+	catBin, err := exec.LookPath("cat")
+	if err != nil {
+		panic(err)
+	}
+	tempFile := c.TempFileMust("file1")
+	fileChecker := ec.NewProcessLsmChecker("lsm-file1-checker").
+		WithFunctionName(sm.Suffix("file_open")).
+		WithProcess(ec.NewProcessChecker().WithBinary(sm.Full(catBin))).
+		WithArgs(ec.NewKprobeArgumentListMatcher().
+			WithOperator(lc.Ordered).
+			WithValues(
+				ec.NewKprobeArgumentChecker().WithFileArg(ec.NewKprobeFileChecker().WithPath(sm.Full(tempFile)))))
+
+	return &policytest.Scenario{
+		Name:         "checking first lsm hook event fired for duplicate hook policy",
+		Trigger:      policytest.NewCmdTrigger(catBin, tempFile).ExpectExitCode(0),
+		EventChecker: ec.NewUnorderedEventChecker(fileChecker),
+	}
+}).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
+	catBin, err := exec.LookPath("cat")
+	if err != nil {
+		panic(err)
+	}
+	tempFile := c.TempFileMust("file2")
+	fileChecker := ec.NewProcessLsmChecker("lsm-file2-checker").
+		WithFunctionName(sm.Suffix("file_open")).
+		WithProcess(ec.NewProcessChecker().WithBinary(sm.Full(catBin))).
+		WithArgs(ec.NewKprobeArgumentListMatcher().
+			WithOperator(lc.Ordered).
+			WithValues(
+				ec.NewKprobeArgumentChecker().WithFileArg(ec.NewKprobeFileChecker().WithPath(sm.Full(tempFile)))))
+
+	return &policytest.Scenario{
+		Name:         "checking second lsm hook event fired for duplicate hook policy",
+		Trigger:      policytest.NewCmdTrigger(catBin, tempFile).ExpectExitCode(0),
+		EventChecker: ec.NewUnorderedEventChecker(fileChecker),
+	}
+}).RegisterAtInit()


bar234:
As discussed in monthly meeting, I'll fix duplicate uprobe support next, then look to combine these very similar test cases with a `policytest` template

pkg/sensors/tracing/genericlsm.go
```diff
@@ -431,12 +434,15 @@ func createGenericLsmSensor(
...
+	dups := make(map[string]int)
 	for _, hook := range lsmHooks {
 		if err := appendMacrosSelectors(hook.Selectors, spec.SelectorsMacros); err != nil {
 			return nil, fmt.Errorf("append macros selectors: %w", err)
 		}
 
-		id, err := addLsm(&hook, &in)
+		instance := dups[hook.Hook]
+		dups[hook.Hook] = instance + 1
+		id, err := addLsm(&hook, instance, &in)
```
olsajiri, line 455: let's keep the same logic with generickprobe sensor and have `instance` and `dups[hook]` with the same value? seems less confusing to me

pkg/sensors/tracing/lsm_test.go
```diff
@@ -454,3 +454,93 @@ spec:
 	}
 	require.Condition(t, checkFunc)
 }
+
+func TestLSMDuplicateHooks(t *testing.T) {
```
olsajiri, line 459: please add this as policytest

olsajiri (commented):
not sure there's any gain in having this unified across the sensors, let's check at least ;-)
I'd separate the code and tests, thanks

pkg/sensors/tracing/genericlsm.go
mtardy: yep agree, logic looks like this

https://github.com/cilium/tetragon/blob/19ca889d72e24b4d12ca95a7f10d7502d94d6647/pkg/sensors/tracing/generickprobe.go#L632-L639

could technically reuse some generic function but this is a small piece of code anyway

mtardy (commented):
agree with Jiri, for more context policytest is the new way of testing that was added recently, feel free to ask if you don't find enough info.

pkg/sensors/tracing/genericlsm.go
bar234: No problem. I looked at kprobe to understand the logic, but I should have circled back and copied the implementation.

bar234:
Just so I understand the `policytest` approach, it looks like `tests/policytests/*` (maybe a new `lsm.go` file here?) is the place to implement a new builder, then call it from `pkg/sensors/tracing/lsm_test.go`. I'm looking at the `kprobe-null-string` builder as an example to follow.

Taking a stab at this and the first hiccup I encountered is the creation of temporary files. I don't see an implementation of this in policytest, so I'm thinking create these outside of the policytest and pass them in as parameters.

Let me know if this sounds reasonable or if there's a better approach here, @mtardy. Thanks

bar234:
@mtardy @olsajiri I moved the tests to `policytest` and took a stab at adding support for tempfiles. I can make adjustments/scale this back if you feel this is necessary, but I wanted to use this as an opportunity to understand the new test framework a bit better.

I see a couple static check CI tests failing ~~but I might need some help understanding them. Vendoring looks like it might just be a hiccup. The `golangci-lint (windows)` one isn't happy with how I'm importing `policytests` into my lsm test file. I see other policytests in that directory are skipped for windows build. If this is fine I can make this change.~~

I see `policytest` is not built for windows anyways, so made this change

Thanks!

pkg/testutils/policytest/builder.go
```diff
@@ -39,13 +40,26 @@ func (b *Builder) WithParameter(p Parameter) *Builder {
...
+//   - tempFile: generate a temporary file and save it in the conf
 func (b *Builder) WithPolicyTemplate(tmpl string) *Builder {
 	policyTest := b.policytest
 	policyTest.Policy = func(c *Conf) (Policy, error) {
 		funcMap := template.FuncMap{
 			"testBinary": func(s string) string {
 				return c.TestBinary(s)
 			},
+			"tempFile": func() (string, error) {
+				tempFile, err := os.CreateTemp("/var/tmp", "tetragon-testfile-*")
```
olsajiri: im not sure it's the best way.. what if you have template with multiple temporary files and you need to know which one is used on specific place so you could setup checker properly

this might be also wrong, but would it be easier to define something like
below with `WithTempFile` creating temporary file associated by name? @kkourt 

```
 // This file contains tests on lsm hooks
 
-var _ = policytest.NewBuilder("lsm-dup-hooks").WithLabels("lsm").WithPolicyTemplate(`
+var _ = policytest.NewBuilder("lsm-dup-hooks").WithLabels("lsm").WithTempFile("file1").WithTempFile("file2").WithPolicyTemplate(`
 kind: TracingPolicy
 metadata:
@@ -31,7 +31,7 @@ spec:
     - matchArgs:
       - args: [0]
         operator: "Equal"
-        values: [{{ tempFile }}]
+        values: [{{ file1 }}]
       matchActions:
         - action: Post
   - hook: "file_open"
@@ -42,7 +42,7 @@ spec:
     - matchArgs:
       - args: [0]
         operator: "Equal"
-        values: [{{ tempFile }}]
+        values: [{{ file2 }}]
       matchActions:
         - action: Post
 `).WithSkip(func(si *policytest.SkipInfo) string {
@@ -55,7 +55,7 @@ spec:
        return ""
 }).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
        myBin := c.TestBinary("direct-write-tester")
-       tempFile := c.TempFiles[0]
+       tempFile := c.TempFiles["file1"]
        fileChecker := ec.NewProcessLsmChecker("lsm-file1-checker").
                WithFunctionName(sm.Suffix("file_open")).
                WithProcess(ec.NewProcessChecker().WithBinary(sm.Full(myBin))).
@@ -71,7 +71,7 @@ spec:
        }
 }).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
        myBin := c.TestBinary("direct-write-tester")
-       tempFile := c.TempFiles[1]
+       tempFile := c.TempFiles["file2"]
        fileChecker := ec.NewProcessLsmChecker("lsm-file2-checker").
                WithFunctionName(sm.Suffix("file_open")).
                WithProcess(ec.NewProcessChecker().WithBinary(sm.Full(myBin))).

```

pkg/sensors/tracing/genericlsm.go
```diff
@@ -440,8 +440,12 @@ func createGenericLsmSensor(
 			return nil, fmt.Errorf("append macros selectors: %w", err)
 		}
 
-		instance := dups[hook.Hook]
```
olsajiri: let's not add code and remove it again in the same PR, let's squash that change together and keep just the final shape in the change

pkg/sensors/tracing/lsm_test.go
```diff
@@ -456,91 +457,5 @@ spec:
 }
 
 func TestLSMDuplicateHooks(t *testing.T) {
-	if !bpf.HasLSMPrograms() || !config.EnableLargeProgs() {
```
olsajiri: same, no need to add full observer test if you're removing it in the same PR, let's just add the policy test

pkg/sensors/tracing/genericlsm.go
bar234: Thanks for the tip, I've re-organized my branch locally to be two commits, one for the LSM fix and the other for the temp file changes in policytest. I'll push it when I make some of the tempfile changes you suggested

pkg/testutils/policytest/builder.go
bar234: Oh I like that a lot! I was using the indexes but that has multiple problems if the template changes. I didn't think about adding the `WithTempFile` methods to the builder. I'll poke around with that a bit and see how it looks. Thank you!

bar234: I ran into some issues creating the tempfiles on `policytest`, but not having access to them when I created the scenario (`AddScenario` only has access to the `conf`, not `policytest`).

Also I was adding awkward collision handling code if a temp file was named `testBinary` since that was another key added to the `funcMap`

So I tried to just follow the `testBinary` example:

```diff
--- a/tests/policytests/lsm.go
+++ b/tests/policytests/lsm.go
@@ -31,7 +31,7 @@ spec:
     - matchArgs:
       - args: [0]
         operator: "Equal"
-        values: [{{ tempFile }}]
+        values: [{{ tempFile "file1"}}]
       matchActions:
         - action: Post
   - hook: "file_open"
@@ -42,7 +42,7 @@ spec:
     - matchArgs:
       - args: [0]
         operator: "Equal"
-        values: [{{ tempFile }}]
+        values: [{{ tempFile "file2" }}]
       matchActions:
         - action: Post
 `).WithSkip(func(si *policytest.SkipInfo) string {
@@ -55,7 +55,7 @@ spec:
        return ""
 }).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
        myBin := c.TestBinary("direct-write-tester")
-       tempFile := c.TempFiles[0]
+       tempFile := c.TempFiles["file1"]
        fileChecker := ec.NewProcessLsmChecker("lsm-file1-checker").
                WithFunctionName(sm.Suffix("file_open")).
                WithProcess(ec.NewProcessChecker().WithBinary(sm.Full(myBin))).
@@ -71,7 +71,7 @@ spec:
        }
 }).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
        myBin := c.TestBinary("direct-write-tester")
-       tempFile := c.TempFiles[1]
+       tempFile := c.TempFiles["file2"]
```

I made the changes in my local branch which I'll test and push shortly so you can see the full picture

pkg/testutils/policytest/observertest.go
```diff
@@ -57,6 +57,7 @@ func (rpt *RegisteredPolicyTests) DoObserverTest(
 		TestConf: &TestConf{
 			ParamValues: params,
 		},
+		TempFiles: make(map[string]string),
```
kkourt: This is  hidden in the PRs, but the purpose of policytest is to be able to run tests via the CLI on TetragonServer.

The `DoObserverTest` is a wrapper we use so that we can execute the tests we migrate using `go test`.

Note that if we do:
```
sudo ./tetra policytest run lsm-dup-hooks
```

The test fails with:
```
P: lsm-dup-hooks   ❌   failed to create policy for test "lsm-dup-hooks": failed to execute template: template: testpolicy:16:20: executing "testpolicy" at <tempFile "file1">: error calling tempFile: assignment to entry in nil map
```

So I would suggest something like:
```diff
-                               if path, ok := c.TempFiles[key]; ok {
+                               if c.TempFiles == nil {
+                                       c.TempFiles = make(map[string]string)
+                               } else if path, ok := c.TempFiles[key]; ok {
                                        return path, nil // already created — return cached path
                                }
```

instead.

pkg/testutils/policytest/observertest.go
```diff
@@ -89,4 +90,11 @@ func (rpt *RegisteredPolicyTests) DoObserverTest(
 		err = jsonchecker.JsonTestCheckExpect(t, scenario.EventChecker, scenario.ExpectCheckerFailure)
 		require.NoError(t, err)
 	}
+
+	// Clean up any temp files created when test runs cleanup
+	for _, tempFile := range conf.TempFiles {
+		t.Cleanup(func() {
+			os.Remove(tempFile)
+		})
+	}
```
kkourt: Similarly, we would need to remove the files for the `./tetra policytest` as well.

kkourt (commented):
Thanks!

I think adding the ability for temporary files in policytests is useful, but given that we also need to execute the tests from the CLI, the approach in this PR does not work as it is.

My suggestion would be to extend the Policy() method to also return a cleanup function.

Something like:

```diff
diff --git a/cmd/tetra/policytest/policytest.go b/cmd/tetra/policytest/policytest.go
index acc1e0cd2..42ec5410c 100644
--- a/cmd/tetra/policytest/policytest.go
+++ b/cmd/tetra/policytest/policytest.go
@@ -96,11 +96,12 @@ func dumpPolicyCmd() *cobra.Command {
 			}
 
 			for _, t := range tests {
-				pol, err := t.Policy(&conf)
+				pol, cleanup, err := t.Policy(&conf)
 				if err != nil {
 					return err
 				}
 				cmd.OutOrStdout().Write([]byte(pol))
+				cleanup()
 			}
 			return nil
 		},
diff --git a/pkg/testutils/policytest/builder.go b/pkg/testutils/policytest/builder.go
index f191b5c58..cf9e6fff4 100644
--- a/pkg/testutils/policytest/builder.go
+++ b/pkg/testutils/policytest/builder.go
@@ -8,7 +8,6 @@ package policytest
 import (
 	"bytes"
 	"fmt"
-	"os"
 	"text/template"
 )
 
@@ -43,28 +42,18 @@ func (b *Builder) WithParameter(p Parameter) *Builder {
 //   - tempFile: generate a temporary file and save it in the conf
 func (b *Builder) WithPolicyTemplate(tmpl string) *Builder {
 	policyTest := b.policytest
-	policyTest.Policy = func(c *Conf) (Policy, error) {
+	policyTest.Policy = func(c *Conf) (Policy, PolicyCleanupFn, error) {
 		funcMap := template.FuncMap{
 			"testBinary": func(s string) string {
 				return c.TestBinary(s)
 			},
 			"tempFile": func(key string) (string, error) {
-				if path, ok := c.TempFiles[key]; ok {
-					return path, nil // already created — return cached path
-				}
-				tempFile, err := os.CreateTemp("/var/tmp", "tetragon-testfile-*")
-				if err != nil {
-					// non-nil error here will be returned by Template.Parse
-					return "", err
-				}
-				tempFile.Close()
-				c.TempFiles[key] = tempFile.Name()
-				return tempFile.Name(), nil
+				return c.TempFile(key)
 			},
 		}
 		t, err := template.New("testpolicy").Funcs(funcMap).Parse(tmpl)
 		if err != nil {
-			return Policy(""), fmt.Errorf("failed to parse template: %w", err)
+			return Policy(""), nil, fmt.Errorf("failed to parse template: %w", err)
 		}
 
 		// fill in params
@@ -80,9 +69,12 @@ func (b *Builder) WithPolicyTemplate(tmpl string) *Builder {
 		var buf bytes.Buffer
 		err = t.Execute(&buf, params)
 		if err != nil {
-			return Policy(""), fmt.Errorf("failed to execute template: %w", err)
+			return Policy(""), nil, fmt.Errorf("failed to execute template: %w", err)
 		}
-		return Policy(buf.String()), nil
+		cleanupFn := func() {
+			c.CleanupTempFiles()
+		}
+		return Policy(buf.String()), cleanupFn, nil
 	}
 	return b
 }
diff --git a/pkg/testutils/policytest/conf.go b/pkg/testutils/policytest/conf.go
index 8a98b22fb..3a6bbe6c7 100644
--- a/pkg/testutils/policytest/conf.go
+++ b/pkg/testutils/policytest/conf.go
@@ -6,6 +6,7 @@
 package policytest
 
 import (
+	"os"
 	"path/filepath"
 )
 
@@ -31,9 +32,44 @@ type Conf struct {
 	// Map of temporary files generated for this generated policy. Key is
 	// unique identifier for the temp file, value is the generated temp
 	// filename
-	TempFiles map[string]string
+	tempFiles map[string]string
 }
 
 func (c *Conf) TestBinary(s string) string {
 	return filepath.Join(c.BinsDir, s)
 }
+
+func (c *Conf) TempFile(key string) (string, error) {
+	if c.tempFiles == nil {
+		c.tempFiles = make(map[string]string)
+	} else if path, ok := c.tempFiles[key]; ok {
+		return path, nil // already created — return cached path
+	}
+	tempFile, err := os.CreateTemp("/var/tmp", "tetragon-testfile-*")
+	if err != nil {
+		// non-nil error here will be returned by Template.Parse
+		return "", err
+	}
+	tempFile.Close()
+	path := tempFile.Name()
+	c.tempFiles[key] = path
+	return path, nil
+}
+
+func (c *Conf) TempFileMust(key string) string {
+	ret, err := c.TempFile(key)
+	if err != nil {
+		panic(err)
+	}
+	return ret
+}
+
+func (c *Conf) CleanupTempFiles() {
+	if c.tempFiles == nil {
+		return
+	}
+	for _, tempFile := range c.tempFiles {
+		os.Remove(tempFile)
+	}
+	c.tempFiles = nil
+}
diff --git a/pkg/testutils/policytest/observertest.go b/pkg/testutils/policytest/observertest.go
index 51575b2e8..d4ab2c651 100644
--- a/pkg/testutils/policytest/observertest.go
+++ b/pkg/testutils/policytest/observertest.go
@@ -57,12 +57,12 @@ func (rpt *RegisteredPolicyTests) DoObserverTest(
 		TestConf: &TestConf{
 			ParamValues: params,
 		},
-		TempFiles: make(map[string]string),
 	}
-	policyStr, err := pt.Policy(conf)
+	policyStr, cleanupFn, err := pt.Policy(conf)
 	if err != nil {
 		t.Fatalf("failed to generate policy: %s", err)
 	}
+	t.Cleanup(cleanupFn)
 
 	policyFile, err := os.CreateTemp("", "tetragon-policy-"+t.Name()+"-*.txt")
 	if err != nil {
@@ -90,11 +90,4 @@ func (rpt *RegisteredPolicyTests) DoObserverTest(
 		err = jsonchecker.JsonTestCheckExpect(t, scenario.EventChecker, scenario.ExpectCheckerFailure)
 		require.NoError(t, err)
 	}
-
-	// Clean up any temp files created when test runs cleanup
-	for _, tempFile := range conf.TempFiles {
-		t.Cleanup(func() {
-			os.Remove(tempFile)
-		})
-	}
 }
diff --git a/pkg/testutils/policytest/policytest.go b/pkg/testutils/policytest/policytest.go
index c10ecfe6a..bbe2ae2e2 100644
--- a/pkg/testutils/policytest/policytest.go
+++ b/pkg/testutils/policytest/policytest.go
@@ -29,6 +29,8 @@ type Scenario struct {
 // Policies are represented as strings, because that's how they are loaded via gRPC
 type Policy string
 
+type PolicyCleanupFn func()
+
 type Label string
 
 type SkipInfo struct {
@@ -53,7 +55,7 @@ type T struct {
 	ShouldSkip func(info *SkipInfo) string
 
 	// Policy generates a policy for this test
-	Policy func(c *Conf) (Policy, error)
+	Policy func(c *Conf) (Policy, PolicyCleanupFn, error)
 
 	Params []Parameter
 
diff --git a/pkg/testutils/policytest/runner.go b/pkg/testutils/policytest/runner.go
index 3b57946e6..75f40a691 100644
--- a/pkg/testutils/policytest/runner.go
+++ b/pkg/testutils/policytest/runner.go
@@ -115,7 +115,7 @@ func (r *LocalRunner) Close() {
 
 func (r *LocalRunner) AddPolicy(l *slog.Logger, test *T) (*PolicyHandler, error) {
 	// generate policy
-	pol, err := test.Policy(r.conf)
+	pol, cleanup, err := test.Policy(r.conf)
 	if err != nil {
 		err = fmt.Errorf("failed to create policy for test %q: %w", test.Name, err)
 		return nil, err
@@ -154,6 +154,7 @@ func (r *LocalRunner) AddPolicy(l *slog.Logger, test *T) (*PolicyHandler, error)
 	return &PolicyHandler{
 		tpName:      tpName,
 		tpNamespace: "", // TODO: change this when we add support for namespaced policies
+		cleanup:     cleanup,
 	}, nil
 }
@@ -385,10 +386,14 @@ func runFwd(
 type PolicyHandler struct {
 	tpName      string
 	tpNamespace string
+	cleanup     PolicyCleanupFn
 }
 
 func (ph *PolicyHandler) Cleanup(l *slog.Logger, conf *Conf, client *cli.ClientWithContext) error {
 
+	// call policy cleanup after we are done
+	defer ph.cleanup()
+
 	_, err := client.Client.DeleteTracingPolicy(client.Ctx, &tetragon.DeleteTracingPolicyRequest{
 		Name:      ph.tpName,
 		Namespace: ph.tpNamespace,
diff --git a/tests/policytests/lsm.go b/tests/policytests/lsm.go
index 3adbabeb4..cfece2e93 100644
--- a/tests/policytests/lsm.go
+++ b/tests/policytests/lsm.go
@@ -55,7 +55,7 @@ spec:
 	return ""
 }).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
 	myBin := c.TestBinary("direct-write-tester")
-	tempFile := c.TempFiles["file1"]
+	tempFile := c.TempFileMust("file1")
 	fileChecker := ec.NewProcessLsmChecker("lsm-file1-checker").
 		WithFunctionName(sm.Suffix("file_open")).
 		WithProcess(ec.NewProcessChecker().WithBinary(sm.Full(myBin))).
@@ -71,7 +71,7 @@ spec:
 	}
 }).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
 	myBin := c.TestBinary("direct-write-tester")
-	tempFile := c.TempFiles["file2"]
+	tempFile := c.TempFileMust("file2")
 	fileChecker := ec.NewProcessLsmChecker("lsm-file2-checker").
 		WithFunctionName(sm.Suffix("file_open")).
 		WithProcess(ec.NewProcessChecker().WithBinary(sm.Full(myBin))).
```

Thoughts?

kkourt:
As an aside, I'm getting:

```
level=warn msg="Failed to remove program dir" prog=file_open dir=lsm-dup-hooks/generic_lsm/file_open error="remove /sys/fs/bpf/tetragon/lsm-dup-hooks/generic_lsm/file_open: no such file or directory"
level=warn msg="Failed to remove program dir" prog=file_open:1 dir=lsm-dup-hooks/generic_lsm/file_open:1 error="remove /sys/fs/bpf/tetragon/lsm-dup-hooks/generic_lsm/file_open:1: no such file or directory"
```

When running the test with `tetra policytest`. I don't think it relates to this PR though.

pkg/testutils/policytest/observertest.go
bar234: Oh thanks for the CLI info. I didn't realize the tests could be run this way. I like the new architecture you suggested. Let me work on this a bit and I'll provide an update

bar234:
> As an aside, I'm getting:
> 
> ```
> level=warn msg="Failed to remove program dir" prog=file_open dir=lsm-dup-hooks/generic_lsm/file_open error="remove /sys/fs/bpf/tetragon/lsm-dup-hooks/generic_lsm/file_open: no such file or directory"
> level=warn msg="Failed to remove program dir" prog=file_open:1 dir=lsm-dup-hooks/generic_lsm/file_open:1 error="remove /sys/fs/bpf/tetragon/lsm-dup-hooks/generic_lsm/file_open:1: no such file or directory"
> ```
> 
> When running the test with `tetra policytest`. I don't think it relates to this PR though.

Confirmed these errors are happening outside of this PR when a policy is unloaded. Can open it as an issue if you'd like. My guess is it's a cleanup ordering issue

pkg/testutils/policytest/observertest.go
bar234: I implemented and tested the changes suggested. Only thing I added was a temp file cleanup in case there's an error before returning from the function assigned to `policyTest.Policy`:

```diff
--- a/pkg/testutils/policytest/builder.go
+++ b/pkg/testutils/policytest/builder.go
@@ -39,17 +39,31 @@ func (b *Builder) WithParameter(p Parameter) *Builder {
 //
 // In the template, the following functions are supported
 //   - testBinary: generate a test binary path from the binary name (Conf.TestBinary())
+//   - tempFile: generate a temporary file and save it in the conf
 func (b *Builder) WithPolicyTemplate(tmpl string) *Builder {
        policyTest := b.policytest
-       policyTest.Policy = func(c *Conf) (Policy, error) {
+       policyTest.Policy = func(c *Conf) (Policy, PolicyCleanupFn, error) {
                funcMap := template.FuncMap{
                        "testBinary": func(s string) string {
                                return c.TestBinary(s)
                        },
+                       "tempFile": func(key string) (string, error) {
+                               return c.TempFile(key)
+                       },
                }
+
                t, err := template.New("testpolicy").Funcs(funcMap).Parse(tmpl)
+
+               // Clean up any temp files created if this function returns due to an
+               // error
+               defer func() {
+                       if err != nil {
+                               c.CleanupTempFiles()
+                       }
+               }()
+
                if err != nil {
-                       return Policy(""), fmt.Errorf("failed to parse template: %w", err)
+                       return Policy(""), nil, fmt.Errorf("failed to parse template: %w", err)
                }

                // fill in params
```

FedeDP:
Changes LGTM; re-triggering review from other reviewers :) 
Thanks!

mtardy:
btw @kkourt

> I think adding the ability for temporary files in policytests is useful, but given that we also need to execute the tests from the CLI, the approach in this PR does not work as it is.

Shouldn't we have this tested in the CI so that we know we don't break this? I guess it wouldn't catch new stuff added like that until we add it to this test? 🤔

pkg/testutils/policytest/conf.go
```diff
@@ -27,8 +28,47 @@ type Conf struct {
...
 	return filepath.Join(c.BinsDir, s)
 }
+
+func (c *Conf) TempFile(key string) (string, error) {
+	if c.tempFiles == nil {
+		c.tempFiles = make(map[string]string)
+	} else if path, ok := c.tempFiles[key]; ok {
+		return path, nil // already created — return cached path
+	}
+	tempFile, err := os.CreateTemp("/var/tmp", "tetragon-testfile-*")
```
mtardy: why do we set `/var/tmp` here? I think you could leave it empty I think

pkg/testutils/policytest/conf.go
```diff
@@ -27,8 +28,47 @@ type Conf struct {
...
+	if err != nil {
+		return "", err
+	}
+	tempFile.Close()
+	path := tempFile.Name()
+	c.tempFiles[key] = path
```
mtardy, line 75: since you don't protect tempFiles, you have a TOCTOU race here. Might be harmless in your case, wdyt?

pkg/testutils/policytest/builder.go
```diff
@@ -39,17 +39,31 @@ func (b *Builder) WithParameter(p Parameter) *Builder {
...
+
 		t, err := template.New("testpolicy").Funcs(funcMap).Parse(tmpl)
+
+		// Clean up any temp files created if this function returns due to an
+		// error
+		defer func() {
+			if err != nil {
+				c.CleanupTempFiles()
+			}
+		}()
```
mtardy, line 63: why doing this with defer in this case?

pkg/testutils/policytest/runner.go
```diff
@@ -385,10 +386,14 @@ func runFwd(
 type PolicyHandler struct {
 	tpName      string
 	tpNamespace string
+	cleanup     PolicyCleanupFn
 }
 
 func (ph *PolicyHandler) Cleanup(l *slog.Logger, conf *Conf, client *cli.ClientWithContext) error {
 
+	// call policy cleanup after we are done
+	defer ph.cleanup()
```
mtardy: ph might be nil

pkg/testutils/policytest/runner.go
```diff
@@ -154,6 +154,7 @@ func (r *LocalRunner) AddPolicy(l *slog.Logger, test *T) (*PolicyHandler, error)
 	return &PolicyHandler{
 		tpName:      tpName,
 		tpNamespace: "", // TODO: change this when we add support for namespaced policies
+		cleanup:     cleanup,
 	}, nil
```
mtardy, line 167: so here, if test.Policy() call failed, you indeed cleanup the tmp files, but if it fails in between line 124 and 154, then we don't run the cleanup method.

mtardy (commented):
a few comments

pkg/testutils/policytest/conf.go
bar234: Oh good point, looks like empty string will use `os.TempDir`. I'll make this change

pkg/testutils/policytest/conf.go
bar234: I can protect `tempFiles` with a mutex which should mitigate this.

Looking at how the tests run, I don't *believe* we have parallel accesses to this map. `DoObserverTest` creates the `conf` (initializes the `tempFiles` map) and then generates the policy (where the `tempFiles` would be populated) serially. I checked out `template.Parse` to see if it parallelizes any parsing and I don't believe it does.

However, I'm ok adding a mutex here in case the use changes in the future or I missed something in my analysis

pkg/testutils/policytest/builder.go
bar234: Honestly I wasn't sure the right approach for running this code on an error return. Alternatively I can just call `c.CleanupTempFiles()` at the error returns from `Parse(tmpl)` and `t.Execute`

pkg/testutils/policytest/runner.go
bar234: Good call, checking for `nil` before issuing this `defer`

pkg/testutils/policytest/runner.go
bar234: Hmmm, so I can add calls to `cleanup` if it's not nil at each of these error returns. However I wonder if there's an architectural adjustment that can be made here. Cleanup doesn't get tied to `PolicyHandler` until this function succeeds, but creation of the policy is where the temp files are created. And this needs to happen at the beginning of `AddPolicy` so a valid policy yaml object is ready to load into the sensors manager.

I don't see a good shift to make here but I'm open to feedback. In the short term I'll add calls to `cleanup` in these error paths but I'll keep thinking about this

kkourt:
> btw @kkourt
> 
> > I think adding the ability for temporary files in policytests is useful, but given that we also need to execute the tests from the CLI, the approach in this PR does not work as it is.
> 
> Shouldn't we have this tested in the CI so that we know we don't break this? I guess it wouldn't catch new stuff added like that until we add it to this test? 🤔

We should. I'm working on it :)

pkg/testutils/policytest/builder.go
kkourt: I like `defer`. It basically says: if there is an error, clean up stuff so that you don't have to clean up things every time there is an error. @mtardy what's your concern?

pkg/testutils/policytest/builder.go
```diff
@@ -65,9 +79,12 @@ func (b *Builder) WithPolicyTemplate(tmpl string) *Builder {
...
 		err = t.Execute(&buf, params)
 		if err != nil {
-			return Policy(""), fmt.Errorf("failed to execute template: %w", err)
+			return Policy(""), nil, fmt.Errorf("failed to execute template: %w", err)
+		}
+		cleanupFn := func() {
+			c.CleanupTempFiles()
 		}
-		return Policy(buf.String()), nil
+		return Policy(buf.String()), cleanupFn, nil
```
kkourt: nit: I think we can even do `return Policy(buf.String()), c.CLeanupTempFiles, nil` here.

pkg/testutils/policytest/runner.go
kkourt: How about the defer trick?

```go
	pol, cleanup, err := test.Policy(r.conf)
	if err != nil {
		err = fmt.Errorf("failed to create policy for test %q: %w", test.Name, err)
		return nil, err
	}
	defer func() {
		if err != nil {
			cleanup()
		}
	}()
```

kkourt (commented):
Thanks!

pkg/testutils/policytest/builder.go
mtardy: ah indeed nevermind, I guess I got confused about closure and defer again, I didn't think it would print "new error": https://go.dev/play/p/e-FFiSSlVrU

pkg/testutils/policytest/runner.go
bar234: If we're all on board with the defer trick I'm happy to use that. It was something I just discovered and was waiting to resolve Mahe's comment on it before deciding it was a good pattern to use elsewhere.

Sounds like we like it :-)

bar234:
Sorry for the delay, just got back from a week of vacation. Pushed code changes and rebased to latest `main` to resolve some conflicts but now I'm seeing some CI failures. I'll take a look at these ASAP

FedeDP:
> ❌	pkg.sensors.tracing.TestLSMDuplicateHooks											(total:1 failed:1 skipped:0)	10.955846176s	22m48.748049952s

The newly introduced test is failing on some kernels :( can you take a look? thanks!

bar234:
I've been trying to reproduce these failures locally without much luck:

```
berose@lima-tetragon:/Users/berose/Development/tetragon$ ./tests/vmtests/tetragon-vmtests-run   --kernel tests/vmtests/test-data/kernels/6.1/boot/vmlinuz-6.1.177 --kernel-ver 6.1-20260706.013554 --base tests/vmte
sts/test-data/images/base.qcow2 --mem 4G --btf-file tests/vmtests/test-data/kernels/6.1/boot/btf-6.1.177 --testsfile tests/vmtests/repeat-lsm-dup-hooks --qemu-disable-kvm
...
results directory: /Users/berose/Development/tetragon/tests/vmtests/vmtests-results-3627767058
✅      pkg.sensors.tracing.TestLSMDuplicateHooks       4.101s  (4.101s)
✅      pkg.sensors.tracing.TestLSMDuplicateHooks       4.076s  (8.178s)
✅      pkg.sensors.tracing.TestLSMDuplicateHooks       4.013s  (12.19s)
✅      pkg.sensors.tracing.TestLSMDuplicateHooks       4.128s  (16.319s)
✅      pkg.sensors.tracing.TestLSMDuplicateHooks       4.168s  (20.487s)
All 5 tests succeeded! 🎉🚢🍕 (took: 1m14.195s, skipped:0)
```

However these are all `arm64` whereas the github runner is `amd64`. So I'm trying to get `amd64` versions of the containers running on my M4 MBP. I'm still poking around with the containers generated and docker configs but if there's a known good way to do this I'm all ears!

FedeDP:
So, from the uploaded vmtests result (5.10), error is:
```
    logcapture.go:24: time=2026-07-06T23:36:56.477Z level=INFO msg="Listening for events..."
    lsm_test.go:460: failed to trigger scenario checking first lsm hook event fired for duplicate hook policy: command /home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/contrib/tester-progs/direct-write-tester /tmp/tetragon-testfile-3184857336 terminated normally with 255, but was expected to terminate with 0
    logcapture.go:24: time=2026-07-06T23:36:56.491Z level=INFO msg="Listening for events completed." error="context canceled"
```
in the tetragon.log there is indeed a single `process_lsm` event:
```json
{"process_lsm":{"process":{"exec_id":"dGV0cmFnb246MTUyNDM4NjMyNTgyMjoxMjY2Mw==","pid":12663,"uid":0,"cwd":"/","binary":"/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/contrib/tester-progs/direct-write-tester","arguments":"/tmp/tetragon-testfile-3184857336","flags":"execve rootcwd clone","start_time":"2026-07-06T23:36:56.486083373Z","auid":4294967295,"parent_exec_id":"dGV0cmFnb246MTUyMDE5MDAwMDAwMDoxMjY1NA==","refcnt":1,"cap":{"permitted":["CAP_CHOWN","DAC_OVERRIDE","CAP_DAC_READ_SEARCH","CAP_FOWNER","CAP_FSETID","CAP_KILL","CAP_SETGID","CAP_SETUID","CAP_SETPCAP","CAP_LINUX_IMMUTABLE","CAP_NET_BIND_SERVICE","CAP_NET_BROADCAST","CAP_NET_ADMIN","CAP_NET_RAW","CAP_IPC_LOCK","CAP_IPC_OWNER","CAP_SYS_MODULE","CAP_SYS_RAWIO","CAP_SYS_CHROOT","CAP_SYS_PTRACE","CAP_SYS_PACCT","CAP_SYS_ADMIN","CAP_SYS_BOOT","CAP_SYS_NICE","CAP_SYS_RESOURCE","CAP_SYS_TIME","CAP_SYS_TTY_CONFIG","CAP_MKNOD","CAP_LEASE","CAP_AUDIT_WRITE","CAP_AUDIT_CONTROL","CAP_SETFCAP","CAP_MAC_OVERRIDE","CAP_MAC_ADMIN","CAP_SYSLOG","CAP_WAKE_ALARM","CAP_BLOCK_SUSPEND","CAP_AUDIT_READ","CAP_PERFMON","CAP_BPF","CAP_CHECKPOINT_RESTORE"],"effective":["CAP_CHOWN","DAC_OVERRIDE","CAP_DAC_READ_SEARCH","CAP_FOWNER","CAP_FSETID","CAP_KILL","CAP_SETGID","CAP_SETUID","CAP_SETPCAP","CAP_LINUX_IMMUTABLE","CAP_NET_BIND_SERVICE","CAP_NET_BROADCAST","CAP_NET_ADMIN","CAP_NET_RAW","CAP_IPC_LOCK","CAP_IPC_OWNER","CAP_SYS_MODULE","CAP_SYS_RAWIO","CAP_SYS_CHROOT","CAP_SYS_PTRACE","CAP_SYS_PACCT","CAP_SYS_ADMIN","CAP_SYS_BOOT","CAP_SYS_NICE","CAP_SYS_RESOURCE","CAP_SYS_TIME","CAP_SYS_TTY_CONFIG","CAP_MKNOD","CAP_LEASE","CAP_AUDIT_WRITE","CAP_AUDIT_CONTROL","CAP_SETFCAP","CAP_MAC_OVERRIDE","CAP_MAC_ADMIN","CAP_SYSLOG","CAP_WAKE_ALARM","CAP_BLOCK_SUSPEND","CAP_AUDIT_READ","CAP_PERFMON","CAP_BPF","CAP_CHECKPOINT_RESTORE"]},"ns":{"uts":{"inum":4026531838,"is_host":true},"ipc":{"inum":4026531839,"is_host":true},"mnt":{"inum":4026531840,"is_host":true},"pid":{"inum":4026531836,"is_host":true},"pid_for_children":{"inum":4026531836,"is_host":true},"net":{"inum":4026531992,"is_host":true},"time":{"inum":4026531834,"is_host":true},"time_for_children":{"inum":4026531834,"is_host":true},"cgroup":{"inum":4026531835,"is_host":true},"user":{"inum":4026531837,"is_host":true}},"tid":12663,"process_credentials":{"uid":0,"gid":0,"euid":0,"egid":0,"suid":0,"sgid":0,"fsuid":0,"fsgid":0},"in_init_tree":false},"parent":{"exec_id":"dGV0cmFnb246MTUyMDE5MDAwMDAwMDoxMjY1NA==","pid":12654,"uid":0,"cwd":"/","binary":"/home/runner/work/tetragon/tetragon/go/src/github.com/cilium/tetragon/go-tests/pkg.sensors.tracing","arguments":"-test.v -test.run ^TestLSMDuplicateHooks$","flags":"procFS rootcwd","start_time":"2026-07-06T23:36:52.289757356Z","auid":4294967295,"parent_exec_id":"dGV0cmFnb246MzM4MDAwMDAwMDoyNDE=","cap":{"permitted":["CAP_CHOWN","DAC_OVERRIDE","CAP_DAC_READ_SEARCH","CAP_FOWNER","CAP_FSETID","CAP_KILL","CAP_SETGID","CAP_SETUID","CAP_SETPCAP","CAP_LINUX_IMMUTABLE","CAP_NET_BIND_SERVICE","CAP_NET_BROADCAST","CAP_NET_ADMIN","CAP_NET_RAW","CAP_IPC_LOCK","CAP_IPC_OWNER","CAP_SYS_MODULE","CAP_SYS_RAWIO","CAP_SYS_CHROOT","CAP_SYS_PTRACE","CAP_SYS_PACCT","CAP_SYS_ADMIN","CAP_SYS_BOOT","CAP_SYS_NICE","CAP_SYS_RESOURCE","CAP_SYS_TIME","CAP_SYS_TTY_CONFIG","CAP_MKNOD","CAP_LEASE","CAP_AUDIT_WRITE","CAP_AUDIT_CONTROL","CAP_SETFCAP","CAP_MAC_OVERRIDE","CAP_MAC_ADMIN","CAP_SYSLOG","CAP_WAKE_ALARM","CAP_BLOCK_SUSPEND","CAP_AUDIT_READ","CAP_PERFMON","CAP_BPF","CAP_CHECKPOINT_RESTORE"],"effective":["CAP_CHOWN","DAC_OVERRIDE","CAP_DAC_READ_SEARCH","CAP_FOWNER","CAP_FSETID","CAP_KILL","CAP_SETGID","CAP_SETUID","CAP_SETPCAP","CAP_LINUX_IMMUTABLE","CAP_NET_BIND_SERVICE","CAP_NET_BROADCAST","CAP_NET_ADMIN","CAP_NET_RAW","CAP_IPC_LOCK","CAP_IPC_OWNER","CAP_SYS_MODULE","CAP_SYS_RAWIO","CAP_SYS_CHROOT","CAP_SYS_PTRACE","CAP_SYS_PACCT","CAP_SYS_ADMIN","CAP_SYS_BOOT","CAP_SYS_NICE","CAP_SYS_RESOURCE","CAP_SYS_TIME","CAP_SYS_TTY_CONFIG","CAP_MKNOD","CAP_LEASE","CAP_AUDIT_WRITE","CAP_AUDIT_CONTROL","CAP_SETFCAP","CAP_MAC_OVERRIDE","CAP_MAC_ADMIN","CAP_SYSLOG","CAP_WAKE_ALARM","CAP_BLOCK_SUSPEND","CAP_AUDIT_READ","CAP_PERFMON","CAP_BPF","CAP_CHECKPOINT_RESTORE"]},"ns":{"uts":{"inum":4026531838,"is_host":true},"ipc":{"inum":4026531839,"is_host":true},"mnt":{"inum":4026531840,"is_host":true},"pid":{"inum":4026531836,"is_host":true},"pid_for_children":{"inum":4026531836,"is_host":true},"net":{"inum":4026531992,"is_host":true},"cgroup":{"inum":4026531835,"is_host":true},"user":{"inum":4026531837,"is_host":true}},"tid":12654,"process_credentials":{"uid":0,"gid":0,"euid":0,"egid":0,"suid":0,"sgid":0,"fsuid":0,"fsgid":0},"in_init_tree":false},"function_name":"file_open","policy_name":"lsm-dup-hooks","args":[{"file_arg":{"path":"/tmp/tetragon-testfile-3184857336","permission":"-rw-------"}}],"action":"KPROBE_ACTION_POST"},"node_name":"tetragon","time":"2026-07-06T23:36:56.486997652Z"}
```

it seems like `direct-write-tester` is failing due to context canceled.

FedeDP:
Re-triggered the job to check if it was a transient issue.

bar234:
Thanks for retrying the tests. My suspicion is it wasn't transient if it was just my new test. I've almost got the kernels that are failing running locally so I can debug why the temp files are disappearing before `direct-write-tester` can touch them on these specific kernels

bar234:
Alright, I think I got it. Using `direct-write-tester` on a `tmpfs` filesystem is throwing errors because of `O_DIRECT` lack of support on certain kernels. I'll work on a fix and push ASAP

FedeDP (approved):
LGTM, waiting for other reviewers to approve :)

pkg/testutils/policytest/conf.go
```diff
@@ -31,6 +33,12 @@ type TestConf struct {
 	ParamValues ParamVals
 }
 
+// Tie a mutex to a map of temp file paths to protect it from race conditions
```
mtardy: nit: you can drop this comment

pkg/testutils/policytest/conf.go
```diff
@@ -43,8 +51,54 @@ type Conf struct {
...
+	// filename
+	tempFiles SafeMap
 }
 
 func (c *Conf) TestBinary(s string) string {
 	return filepath.Join(c.BinsDir, s)
 }
+
+func (c *Conf) TempFile(key string) (string, error) {
+	// Protect the temp files map while we're reading and possibly updating it
```
mtardy: nit: same

pkg/sensors/tracing/genericlsm.go
```diff
@@ -442,7 +446,13 @@ func createGenericLsmSensor(
 		in.selectorStatsBase = selectorStatsBase
 		selectorStatsBase += uint32(len(hook.Selectors))
 
-		id, err := addLsm(&hook, &in)
+		instance, ok := dups[hook.Hook]
+		if ok {
+			instance = instance + 1
```
mtardy: nit: instance++?

mtardy (approved):
lgtm overall, I'll let kornilios or jiri merge it

pkg/sensors/tracing/genericlsm.go
bar234: Thanks Mahe. Pushed a fix for these nits

bar234:
@kkourt @olsajiri Small nudge for your reviews. This seems pretty close so I'll start applying a similar fix to uprobes

pkg/testutils/policytest/conf.go
```diff
@@ -43,8 +50,53 @@ type Conf struct {
 
 	// Path to save the generated policy
 	DumpPolicyPath string
+
+	// Map of temporary files generated for this generated policy. Key is
+	// unique identifier for the temp file, value is the generated temp
+	// filename
+	tempFiles SafeMap
```
kkourt: nit: I think it's cleaner if we have the two fields in the `Conf` struct, i.e., 

```go
type Conf struct {
...
     tempFiles map[string]string
     tempFilesMu sync.Mutex
}
```

kkourt (approved):
Just a small nit, but otherwise LGTM.

pkg/testutils/policytest/conf.go
bar234: No problem, just pushed this change

pkg/testutils/policytest/conf.go
```diff
@@ -43,8 +45,54 @@ type Conf struct {
...
+	if err != nil {
+		return "", err
+	}
+	tempFile.Close()
+	path := tempFile.Name()
+	c.tempFiles[key] = path
+	return path, nil
+}
+
+func (c *Conf) TempFileMust(key string) string {
```
olsajiri, line 79: nit, I'd keep just TempFileMust, also do we have something in policytest to fail the test, panic looks like too much @kkourt  ?

olsajiri (approved):
please split test in separate commit plus I left one small comment, otherwise lgtm

pkg/testutils/policytest/conf.go
bar234: Just to clarify, are you ok keeping `TempFileMust` as-is but are looking for a test fail trigger instead of `panic`?

In other tests I see errors from the `AddScenario` function returned as `Scenario` objects with errors in the name and trigger of `/usr/bin/false`:

https://github.com/cilium/tetragon/blob/d4c67247adeffff267b64fe2aed4bffb328a4540/tests/policytests/kprobes.go#L147-L169

Not sure if that's a pattern to replicate. Alternatively I wonder if we could check for an error on return from the Scenario function (`s(conf)`/`sc(r.conf)`) and throw an error to the test runner/result structure at that point:
https://github.com/cilium/tetragon/blob/d4c67247adeffff267b64fe2aed4bffb328a4540/pkg/testutils/policytest/observertest.go#L83-L87
https://github.com/cilium/tetragon/blob/d4c67247adeffff267b64fe2aed4bffb328a4540/pkg/testutils/policytest/runner.go#L189-L196

I can spin that off into a new PR and even retroactively apply to the kprobe tests if interested

olsajiri: > Just to clarify, are you ok keeping `TempFileMust` as-is but are looking for a test fail trigger instead of `panic`?

yes, let's keep TempFileMust, I think when we are not able to create temp file it's ok to fail the test, but I wonder we could do something else than panic

kkourt: I believe `TempFileMust` was my idea (https://github.com/cilium/tetragon/pull/5114#pullrequestreview-4530721018).
I think it's the best option to keep writing the tests as declarative as possible.

kkourt:
I believe we need to backport this.

kkourt:
We have ACKs from everyone, so I'm merging. Thanks Ben! If there are additional items, let's do them as followups.

FedeDP:
Cherry picked in https://github.com/cilium/tetragon/pull/5294

PR #5116, opened 2026-06-09
tetragon: use standard iteration pattern

olsajiri:
add our iteration pattern to two more places to save some verifier cycles

### DIFF
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -890,33 +890,59 @@ filter_char_buf_postfix(struct selector_arg_filter *filter, char *arg_str, uint 
 }
 
 #ifdef __LARGE_BPF_PROG
+struct filter_char_substring_data {
+	struct selector_arg_filter *filter;
+	char *arg_str;
+	uint arg_len;
+	bool igncase;
+};
+
+FUNC_LOCAL long
+do_filter_char_substring(int i, struct filter_char_substring_data *data, bool igncase)
+{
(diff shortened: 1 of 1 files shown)


bpf/process/types/basic.h
```diff
@@ -890,33 +890,59 @@ filter_char_buf_postfix(struct selector_arg_filter *filter, char *arg_str, uint
...
+		}
+	} else {
+		for (i = 0; i < MAX_SUBSTRING_VALUES; i++) {
+			if (do_filter_char_substring(i, &data, igncase))
+				return 1;
+			j += 4;
+			if (j + 8 >= data.filter->vallen)
+				break;
+		}
 	}
```
andrewstrohman, line 946: Why are we not doing `bpf_loop` #ifdef __V61_BPF_PROG?

andrewstrohman: I guess the complication is that we would need to store a pointer to the return value in the ctx being passed to `bpf_loop`

andrewstrohman (approved):
LGTM.

FedeDP (approved):

bpf/process/types/basic.h
olsajiri: it's the `igncase` constant.. I'd need to pass it through the ctx and I think verifier would not treat it as a constant and `do_filter_char_substring` might not eliminate the other kfunc if it's missing.. but haven't tried it.. I just need to use the bpf_loop on latest kernel ;-)

PR #5117, opened 2026-06-09
renovate: add rule for mikefarah/yq updates

dangome3:
### Description

Updates for yq image are happening quite often, and besides security updates we don't need to bump the version so frequently.

This commit sets the schedule for updating yq once every 3 months.


mtardy (approved):
Thanks! I've been tired of this update :)

PR #5125, opened 2026-06-10
linters: add sloglint settings

3u13r:
### Description

Configure sloglint to enforce e.g. not re-using the built-in key names.
Also, enforce that the slog.DefaultLogger is not used directly. I replaced those invocations with our logger package. If we don't initialize the logger we fall back to the same behavior by using the slog.DefaultLogger.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -43,6 +43,15 @@ linters:
         - "on"
         - at
         - from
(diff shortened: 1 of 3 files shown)


sayboras (approved):
Thanks

PR #5132, opened 2026-06-13
tetragon: Fix readArg handling for fd_ty type

olsajiri:
Added a check for fd_ty before the generic fault-handling.

When fd_ty returns -1, it means the fd was not found in the
fdinstall_map — this is a filter decision (don't emit the event),
not a fault. We propagate -1 so the caller knows to reject.

Fixes: https://github.com/cilium/tetragon/commit/e6a89ab2b8c9012a2c22d1fbd7010f7a1153e69f ("bpf: detect faults for basic types")

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -502,6 +502,12 @@ read_arg(void *ctx, int index, int type, long orig_off, unsigned long arg, int a
 	}
 
 	if (ret < 0) {
+		/* fd_ty returns -1 when the fd is not in the fdinstall_map,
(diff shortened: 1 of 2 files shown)


andrewstrohman:
I deleted my original comment about https://github.com/cilium/tetragon/blob/8f6a9f6fb1e0c753b4f76295818091c905bbd031/bpf/process/generic_calls.h#L282-L284 because I realize now that:
1) this error path is not about reading args, but rather about reading/writing map values
2) because of 1), we don't need to read call probe_read here. Instead, we could/should just use memcpy.

andrewstrohman (approved):

PR #5082, opened 2026-06-01
pkg/sensors/tracing: skip x86 specific tests on non-x86

3u13r:
### Description
GitHub upgraded the kernel in our CI to 6.17. This kernel now supports the kfuncs needed for the tests, hence they are enabled.

These tests are now failing because they use x86 specific registers in the policy.
Therefore, skip them on non-x86 platforms.

PR #5137, opened 2026-06-17
pkg/sensors: Add arm64 policies for uprobe regs tests

### LINKED ISSUE #5083: ci: add arm-specific policies in pkg/sensors/tracing tests
In https://github.com/cilium/tetragon/pull/5082 we've skipped some tests on non-x86 because the policy was referencing x86-specific registers.

Some tests in that file already provide arm-specific version of the policy. 

Add arm-specific policies to the currently skipped tests in that file.

sayboras:
### Description

PR #5082 skipped four uprobe register tests on arm64 because their policies referenced x86-specific registers. The deeper issue was that they all rely on test_3, which only had a real x86 implementation (arm64 got a stub returning 0).

Add an aarch64 test_3 that loads the string pointer into x1 and returns 0xff, then provide arm64 policy variants (symbol offsets, resolve and override registers) for the four tests so they run instead of skipping.

Fixes: https://github.com/cilium/tetragon/issues/5083

### Changelog

```release-note
pkg/sensors: Add arm64 policies for uprobe regs tests
```

### DIFF
--- a/pkg/sensors/tracing/uprobe_reg_test.go
+++ b/pkg/sensors/tracing/uprobe_reg_test.go
@@ -825,10 +825,6 @@ func testUprobePtRegsPreload(t *testing.T, multi bool) {
 		t.Skip("skipping")
 	}
 
-	if runtime.GOARCH == "arm64" {
-		t.Skip("skipping, x86_64 only test")
-	}
-
 	testBinary := testutils.RepoRootPath("contrib/tester-progs/regs-override")
 
 	disableUprobeMulti := ""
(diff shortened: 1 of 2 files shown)


FedeDP (approved):

pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -840,22 +836,13 @@ func testUprobePtRegsPreload(t *testing.T, multi bool) {
...
-	//      "pop    %%rbp\n"          /* +11 5d                            */
-	// -->  "mov    $0x0,%%rax\n"     /* +12 48 c7 c0 00 00 00 00          */
-	//      "mov    $0xff,%%rax\n"    /* +19 48 c7 c0 ff 00 00 00          */
-	//      "ret\n"                   /* +26 c3                            */
-	//      :
-	//      : [str] "m" (test_3_string)
-	//
-	// Make sure we retrieve data with eax value (1) as int argument
-	// and match the expected value via matchData.
+	// Attach where the test_3 string pointer is still in a register (rdi on
```
olsajiri: let's not do this, it will be nightmare to debug later

if you want to add arm64 tests just do it in separate ifdef and add all the comments as for the x86_64 case

pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -941,26 +924,12 @@ func testUprobePtRegsPreloadDouble(t *testing.T, multi bool) {
...
-	//      "ret\n"                   /* +26 c3                            */
-	//      :
-	//      : [str] "m" (test_3_string)
-	//
-	// Make sure we retrieve data with eax value (1) as int argument
-	// and match the expected value via matchData.
-	//
-	// This test does the same thing as testUprobePtRegsPreload but instead
-	// of single uprobe with preload argument it adds two uprobes with preload
-	// argument.
```
olsajiri, line 963: please keep the x86_64 stuff, there's no point in removing these comments,
just add separate policies for arm if needed, bit of duplication is ok

pkg/sensors/tracing/uprobe_reg_test.go
sayboras: good point :100:

pkg/sensors/tracing/uprobe_reg_test.go
sayboras: sure, I will revert this change

kkourt (approved):

olsajiri (approved):

PR #5140, opened 2026-06-17
Add per-selector action counters

tpapagian:
Tested with:
```yaml
kind: TracingPolicy
metadata:
  name: "lsm-file-open"
spec:
  lsmhooks:
  - hook: "file_permission"
    args:
    - index: 0
      type: "file"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/opt/"
      matchActions:
      - action: Post
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/etc/"
      matchActions:
      - action: NoPost
  - hook: "file_open"
    args:
      - index: 0
        type: "file"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/home/"
      matchActions:
      - action: NoPost
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/etc/"
      matchActions:
      - action: Post
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/mnt/"
      matchActions:
      - action: Override
        argError: -1
      - action: Post
```

Example output (with `sudo ./tetra tp list -o json | jq .`):
```json
{
  "policies": [
    {
      "id": "2",
      "name": "lsm-file-open",
      "sensors": [
        "generic_lsm"
      ],
      "enabled": true,
      "state": "TP_STATE_ENABLED",
      "kernel_memory_bytes": "2323184",
      "mode": "TP_MODE_ENFORCE",
      "stats": {
        "action_counters": {
          "post": "457",
          "override": "1",
          "nopost": "849"
        },
        "selector_action_counters": [
          {
            "hook": "lsm:file_permission",
            "hook_index": 0,
            "selector_index": 0,
            "action_counters": {
              "post": "1"
            }
          },
          {
            "hook": "lsm:file_permission",
            "hook_index": 0,
            "selector_index": 1,
            "action_counters": {
              "nopost": "802"
            }
          },
          {
            "hook": "lsm:file_open",
            "hook_index": 1,
            "selector_index": 0,
            "action_counters": {
              "nopost": "47"
            }
          },
          {
            "hook": "lsm:file_open",
            "hook_index": 1,
            "selector_index": 1,
            "action_counters": {
              "post": "455"
            }
          },
          {
            "hook": "lsm:file_open",
            "hook_index": 1,
            "selector_index": 2,
            "action_counters": {
              "post": "1",
              "override": "1"
            }
          }
        ]
      },
      "domain": "static"
    }
  ]
}
```

Note: per-selector action counter only are only shown with `-o json` argument in `tetra tp list`, otherwise the output is much complicated to show. 

The same per-selector action counters are also reported as metrics:
```console
$ curl localhost:2112/metrics 2> /dev/null | grep tracingpolicy_selector_actions_total
# HELP tetragon_tracingpolicy_selector_actions_total The total number of tracing policy actions observed per selector.
# TYPE tetragon_tracingpolicy_selector_actions_total counter
tetragon_tracingpolicy_selector_actions_total{action="nopost",hook="lsm:file_open",hook_index="1",policy="lsm-file-open",policy_namespace="",selector_index="0"} 51
tetragon_tracingpolicy_selector_actions_total{action="nopost",hook="lsm:file_permission",hook_index="0",policy="lsm-file-open",policy_namespace="",selector_index="1"} 816
tetragon_tracingpolicy_selector_actions_total{action="override",hook="lsm:file_open",hook_index="1",policy="lsm-file-open",policy_namespace="",selector_index="2"} 1
tetragon_tracingpolicy_selector_actions_total{action="post",hook="lsm:file_open",hook_index="1",policy="lsm-file-open",policy_namespace="",selector_index="1"} 462
tetragon_tracingpolicy_selector_actions_total{action="post",hook="lsm:file_open",hook_index="1",policy="lsm-file-open",policy_namespace="",selector_index="2"} 1
tetragon_tracingpolicy_selector_actions_total{action="post",hook="lsm:file_permission",hook_index="0",policy="lsm-file-open",policy_namespace="",selector_index="0"} 1
```


api/v1/tetragon/sensors.proto
```diff
@@ -78,10 +78,22 @@ message TracingPolicyActionCounters {
   uint64 set = 8;
   // number of set actions that did not occur because the policy was in monitor mode
   uint64 monitor_set = 9;
+  // number of events suppressed by NoPost actions
+  uint64 nopost = 10;
```
kkourt, line 82: Can you split the addition of a nopost event with the per-selector action counters in different patches?

bpf/process/generic_calls.h
```diff
@@ -1099,6 +1112,7 @@ do_action(void *ctx, __u32 i, struct selector_action *actions, bool *post, bool
 
 	if (polacct != POLICY_INVALID_ACT_) {
 		policy_stats_update(polacct);
+		policy_selector_stats_update(polacct, generic_selector_stats_id(e));
```
kkourt, line 1070: Do we need policy_stats_update()? Can't we just do the addition in user-space?

kkourt (commented):
Thanks!

I believe we can now remove the old policyStats map, no?

api/v1/tetragon/sensors.proto
tpapagian: Done

bpf/process/generic_calls.h
tpapagian: Removed that. Also removed the policyStats as you proposed.

api/v1/tetragon/sensors.proto
kkourt: Apologies if I wasn't clear. i meant not only the .proto patches but also the functionality. For example, the "add_selectors" commit also includes the introduction of the no post action counter in BPF side.

bpf/lib/policy_stats.h
```diff
@@ -34,6 +34,7 @@ enum policy_actions {
 	POLICY_MONITOR_NOTIFY_ENFORCER = 7, /* policy did not notify the enforcer because it was in monitor mode */
 	POLICY_SET = 8, /* policy set user-space values */
 	POLICY_MONITOR_SET = 9, /* policy did not set user-space values because it was in monitor mode */
+	POLICY_NOPOST = 10, /* policy suppressed an event */
```
kkourt, line 38: I feel like the introduction of POLICY_NOPST warrants its own patch, and it's own commit message.

bpf/lib/process.h
```diff
@@ -655,7 +656,7 @@ event_output(void *ctx, void *data, u64 size)
 	return ringbuf_output(&tg_rb_events, data, size, 0);
 }
 
-FUNC_INLINE void
+FUNC_INLINE bool
```
kkourt, line 658: This change also seems to warrant its own patch and commit message.

pkg/sensors/collection.go
```diff
@@ -160,19 +168,120 @@ func (c *collection) stats() *tetragon.TracingPolicyStats {
...
+		ret = appendSelectorStatsMetadata(ret, "tracepoint:"+tp.Subsystem+"/"+tp.Event, tp.Selectors)
+	}
+	for _, lsm := range spec.LsmHooks {
+		ret = appendSelectorStatsMetadata(ret, "lsm:"+lsm.Hook, lsm.Selectors)
+	}
+	for _, usdt := range spec.Usdts {
+		ret = appendSelectorStatsMetadata(ret, "usdt:"+usdt.Path+":"+usdt.Provider+"/"+usdt.Name, usdt.Selectors)
+	}
+	return ret
+}
```
kkourt, line 252: @andrewstrohman I feel like the strings that @tpapagian generates here are similar to what we are discussing in https://github.com/cilium/tetragon/pull/4994 for hook description, maybe it's worth reusing them?

@tpapagian can you please update the commit message with more details about what this commit does?

bpf/lib/policy_stats.h
tpapagian: Done

bpf/lib/process.h
tpapagian: Done

pkg/sensors/collection.go
tpapagian: Done

andrewstrohman:
Users can specify the same hook multiple times within one policy. Without specifying a hook index, or ordering the status of the hooks in the same order that they are configured in the policy, I'm not sure how users can differentiate.

I have this same concern for reporting uprobe hook status. My current approach to solve this is to list the hook status in the same order as they are configured within the policy. I see that from your example that the status is in the same order as what was configured in the policy. 

What would be our recommendation to users about how to match the action counts to the hook when they have the same hook configured multiple times in one policy?

Since the metrics are not ordered, it seems like we would need a hook_index label in order to differentiate for the metrics case.

pkg/sensors/collection.go
andrewstrohman: >  maybe it's worth reusing them?

Yes, definitely. I'll re-use these for the description. Even with adding the symbols/addrs/offsets to the hook description, I think we can still have situation where it's ambiguous which hook the hook name corresponds to. I already left another comment in this PR about about.

tpapagian:
> Users can specify the same hook multiple times within one policy. Without specifying a hook index, or ordering the status of the hooks in the same order that they are configured in the policy, I'm not sure how users can differentiate.
> 
> I have this same concern for reporting uprobe hook status. My current approach to solve this is to list the hook status in the same order as they are configured within the policy. I see that from your example that the status is in the same order as what was configured in the policy.
> 
> What would be our recommendation to users about how to match the action counts to the hook when they have the same hook configured multiple times in one policy?
> 
> Since the metrics are not ordered, it seems like we would need a hook_index label in order to differentiate for the metrics case.

Yes, this makes sense. The only reason that we can do that is that we don't allow different hook types inside the same policy. For example this policy: https://gist.github.com/tpapagian/915774a7bf2fa9b1a9b0ec22947ddd8d fails to load with `error="policy handler 'tracing' failed loading policy 'lsm-file-open': tracing policies with multiple sections of kprobes, tracepoints, lsm hooks, uprobes or usdts are currently not supported"`.

This is because, we cannot define a `hook_index` for all hooks across all hook types. These are parsed in `selectorStatsMetadataFromSpec` (part of this PR) where I believe that we do not have the information on the order of the hook types but only for the hooks inside a specific hook type.

A proper way to fix that is to add a label per-selector which would be user-defined. This is something that I am working on in https://github.com/cilium/tetragon/pull/5158. I just choose to split that in 2 PRs as this becomes large enough.

To summarise, I will also add a hook_index in this PR and we will have labels in the next PR as well. Does this sound reasonable?

tpapagian:
> To summarise, I will also add a hook_index in this PR and we will have labels in the next PR as well. Does this sound reasonable?

I have updated the PR to include `hook_index` as well.

andrewstrohman:
> Yes, this makes sense. The only reason that we can do that is that we don't allow different hook types inside the same policy. For example this policy: https://gist.github.com/tpapagian/915774a7bf2fa9b1a9b0ec22947ddd8d fails to load with error="policy handler 'tracing' failed loading policy 'lsm-file-open': tracing policies with multiple sections of kprobes, tracepoints, lsm hooks, uprobes or usdts are currently not supported".

I assume that we are talking about the approach where the ordering of the status reflects the ordering of the config. Yesterday morning, when I wrote that, I had a uprobe-specific status in the tracing policy. So, there was a consideration for allowing multiple hooks types per policy in the future. Along these lines, it's also worth noting that currently one policy can create multiple sensors. See [here](https://github.com/isovalent/hubble-fgs/issues/8171#issuecomment-4657324124) for more details on that.

But given review feedback that I read after writing my comment here, I changed the structure to have a generic status, instead of uprobe-specific. In this new approach, I'm adding an additional piece of information to each status entry that describes what section of the config that the status is about. I also added a hook_index in order cope with multiple sensors or a future where we allow policies with multiple hook types.


> This is because, we cannot define a hook_index for all hooks across all hook types. These are parsed in selectorStatsMetadataFromSpec (part of this PR) where I believe that we do not have the information on the order of the hook types but only for the hooks inside a specific hook type.


Makes sense. I agree it would be best to only index within the context of a hook type. As I described above, I'm planning on doing the same, along with specifying which spec section the hook lives in. But, looking at the prefix of the "hook" string in your output could serve the same purpose as explicitly calling out the section of the config. 

> A proper way to fix that is to add a label per-selector which would be user-defined. This is something that I am working on in https://github.com/cilium/tetragon/pull/5158. I just choose to split that in 2 PRs as this becomes large enough.

Ok, makes sense. I think the combination of both the label and hook_index is good, because that way the user doesn't need to use labels to differentiate. Then optionally, they can add the label for convenience and readability.


> To summarise, I will also add a hook_index in this PR and we will have labels in the next PR as well. Does this sound reasonable?

Sounds good to me!

andrewstrohman:
I'm going to remove myself as a reviewer here, because I think you were only asking for me to respond to a specific aspect of this PR (how to determine which hook a status belongs to).

If you want me to review this PR in its entirely, just add me back as a reviewer, and I'll take the time to carefully review all aspects of the change.

kkourt:
> If you want me to review this PR in its entirely, just add me back as a reviewer, and I'll take the time to carefully review all aspects of the change.

Please do. I think it would be helpful to have another pair of eyes on this. Thanks!

bpf/lib/policy_stats.h
```diff
@@ -60,11 +67,26 @@ policy_stats_update(int act)
...
 }
+
+FUNC_INLINE void
+policy_selector_stats_update(int act, __u32 selector_stats_id)
+{
+	struct policy_stats *pstats;
+
+	pstats = map_lookup_elem(&selector_stats, &selector_stats_id);
+	if (pstats)
+		lock_add(&pstats->act_cnt[act], 1);
```
kkourt, line 62: Would it make sense to have an errmetrics here?

api/v1/tetragon/sensors.proto
```diff
@@ -82,8 +82,20 @@ message TracingPolicyActionCounters {
...
+message TracingPolicySelectorActionCounters {
+  // hook is the policy hook that owns the selector
+  string hook = 1;
+  // hook_index is the hook index withing a policy
+  google.protobuf.UInt32Value hook_index = 2;
+  // selector_index is the selector index within the hook
+  google.protobuf.UInt32Value selector_index = 3;
+  // action counters for the selector
+  TracingPolicyActionCounters action_counters = 4;
+}
```
kkourt, line 94: What do you think of having something like

```
message TracingPolicySelectorActionCounters {
  // tracing policy section (kprobes, tracepoints, etc.)
  string section = 1;
  // hook is the policy hook that owns the selector
  string hook = 2;
  // hook_index is the hook index within a section
  google.protobuf.UInt32Value hook_index = 2;

// same as before
}
```

That is, instead of having a global hook_index, add a section and then a hook index within this section.
I think this is clearer for the user, but it means that we also don't need to pass the base selector index in the different sensors.

kkourt (commented):
LGTM, thanks!

Just a couple of suggestions. Let me know what you think.

bpf/lib/policy_stats.h
tpapagian: Yes, this makes sense. I will add that.

api/v1/tetragon/sensors.proto
tpapagian: > instead of having a global hook_index, add a section and then a hook index within this section.

The `hook_index` as it is now, is not global but per-section. I need to fix the comment in the proto file and more specifically replace `policy` with `section`. This is what I also described in https://github.com/cilium/tetragon/pull/5140#issuecomment-4776087575: `This is because, we cannot define a hook_index for all hooks across all hook types. These are parsed in selectorStatsMetadataFromSpec (part of this PR) where I believe that we do not have the information on the order of the hook types but only for the hooks inside a specific hook type.`

I am fine with adding `string section = 1;` but this will be always the same as the prefix of `hook` (from the beginning until the first `:`). 

> but it means that we also don't need to pass the base selector index in the different sensors.

We don't pass the base selector index in different sensors. We pass the base selector index in the different hooks of a single sensor. Or I don't understand the comment correctly 😕

pkg/sensors/collection.go
```diff
@@ -180,6 +196,97 @@ func actionCountersFromStats(stats *policystats.PolicyStats) *tetragon.TracingPo
...
+	metadata := selectorStatsMetadataFromSpec(c.tracingpolicy.TpSpec())
+	if len(metadata) == 0 {
+		return nil
+	}
+
+	stats, err := policystats.GetPolicySelectorStats(c.tracingpolicy)
+	if err != nil {
+		if !c.warnedOnStatsRetrievalFailure.Load() {
+			c.warnedOnStatsRetrievalFailure.Store(true)
+			logger.GetLogger().Warn("failed to retrieve policy selector stats", "err", err, "policy", c.name)
```
andrewstrohman: This will stop spamming of the log, but if we do it with `c.warnedOnStatsRetrievalFailure.CompareAndSwap(false, true)` instead, we guarantee only one log line. This caught my attention because we put the effort into using atomics, but then don't actually get the benefit of using them.

I see that this was modeled after warnedOnModeRetrievalFailure, but I have the same assessment for that case as well.

pkg/sensors/tracing/policyhandler_linux.go
```diff
@@ -115,6 +143,19 @@ func (pi *policyInfo) policyStatsMap(prog *program.Program) *program.Map {
...
 
+func (pi *policyInfo) selectorStatsMap(prog *program.Program) *program.Map {
+	if pi.selectorStats != nil {
+		return program.MapUserFrom(pi.selectorStats)
+	}
+	pi.selectorStats = program.MapBuilderPolicy(policystats.PolicySelectorStatsMapName, prog)
+	entries := pi.selectorCount
+	if entries == 0 {
+		entries = 1
+	}
```
andrewstrohman, line 146: I guess this is because we cannot have a map with zero values? Have you considered not loading the map if there is no selector configuration? The benefit of doing that may not be worth the additional complexity required to do it.

api/v1/tetragon/sensors.proto
```diff
@@ -82,8 +82,20 @@ message TracingPolicyActionCounters {
   uint64 nopost = 10;
 }
 
+message TracingPolicySelectorActionCounters {
+  // hook is the policy hook that owns the selector
+  string hook = 1;
+  // hook_index is the hook index withing a policy
+  google.protobuf.UInt32Value hook_index = 2;
+  // selector_index is the selector index within the hook
+  google.protobuf.UInt32Value selector_index = 3;
```
andrewstrohman, line 91: I'm thinking that you've chosen `google.protobuf.UInt32Value` over `uint32` here to see the zero values in `sudo ./tetra tp list -o json`. 

An alternative to this, is to use marshalling options like:
```
	marshaller := protojson.MarshalOptions{
		EmitUnpopulated: true,
	}

	jsonData, err := marshaller.Marshal(res)
``` 
[here](https://github.com/cilium/tetragon/blob/0a0dfdc9cc07dbaa1721839432170b792ce08cd0/cmd/tetra/tracingpolicy/tracingpolicy.go#L232).

However, because this will print all the zero values, that may not be acceptable. I just wanted to make you aware of this, in case you're not.

andrewstrohman (approved):
LGTM!

andrewstrohman:
Prior to this change, the policy-wide POLICY_POST was incremented every time an event was posted. So for example, the loader messages incremented the counter. Now, it's only incremented when we send events because we called the TAIL_CALL_SEND tail call.

bpf/lib/policy_stats.h
tpapagian: Done.

pkg/sensors/collection.go
tpapagian: Yes, this makes sense. I have added one more patch to fix both `warnedOnStatsRetrievalFailure` and `warnedOnModeRetrievalFailure`.

api/v1/tetragon/sensors.proto
tpapagian: Yes, as you said, this may have unintended side-effects and this is the reason that I used `google.protobuf.UInt32Value` over `uint32`.

pkg/sensors/tracing/policyhandler_linux.go
tpapagian: Yes, this makes sense. I haven't checked the details but it worth thinking of that Tetragon-wide as I believe that we are using this pattern in several other places.

tpapagian:
> Prior to this change, the policy-wide POLICY_POST was incremented every time an event was posted. So for example, the loader messages incremented the counter. Now, it's only incremented when we send events because we called the TAIL_CALL_SEND tail call.

It seems that only loader messages are affected by that (the rest are not policy based). The truth, as it is now, is that even when loader is loaded, the policy_stats map is not populated and thus they are ignored. It would be good to have this in mind if we need to change that at some point.

tpapagian:
Merging that and we can do any any followups on https://github.com/cilium/tetragon/pull/5158 or a separate PR. Thanks for the reviews!

andrewstrohman:
> It seems that only loader messages are affected by that (the rest are not policy based). 

I see the following, which are not policy based:
MSG_OP_CGROUP
MSG_OP_EXECVE
MSG_OP_EXIT
MSG_OP_CLONE
MSG_OP_LOADER
MSG_OP_THROTTLE
MSG_OP_DATA


> The truth, as it is now, is that even when loader is loaded, the policy_stats map is not populated and thus they are ignored.

Thank for pointing this out. I see now that ~~most of~~ the programs that emit the above messages do not have HAS_POLICY_STATS defined, so it's a no-op. ~~It looks like MSG_OP_DATA might be an exception though. That being said,~~ I don't think this is worth worrying about anymore.


Thanks for the explanation.

mann-ds:
:wave: when will there be a new release with this PR changes included?

mtardy:
missed that PR but would be cool to add some documentation around this new feature.

PR #5148, opened 2026-06-17
tracing: add native DNS query parsing for UDP kprobes

### LINKED ISSUE #4948: Feature: Native DNS query parsing in TracingPolicy, capture queried domain names from udp_sendmsg/udp_recvmsg payloads

_Note:_ 

This request stems from [this](https://cilium.slack.com/archives/C03EV7KJPJ9/p1774298767772689) discussion on the Cilium Slack (Tetragon channel). I'm raising it here because filing it on GitHub gives everyone interested a public place to subscribe to and watch progress, rather than having the discussion get lost in Slack history. It also gives maintainers something concrete to reference from PRs or related efforts. I'm personally interested in seeing this feature land, which is the other reason I'm filing it.

Tetragon TracingPolicies can observe DNS-related socket activity today (e.g. by hooking udp_sendmsg and filtering on DPort: 53), but there is no supported way to extract the actual DNS payload  i.e. the domain name being queried  from a hooked function.

The use case originated on the Cilium Slack, where a user running Tetragon v1.6.0 was attempting to monitor per-pod DNS lookups for observability purposes, specifically, to log which domains are being queried by which pods, with the longer-term goal of detecting things like typosquatting attempts. They were able to capture socket metadata (src/dst IP, port 53) reliably by hooking `udp_sendmsg` with a `sock arg`, but extracting the DNS payload itself proved unworkable with the current argument types:

* type: `sock` on arg `0` of `udp_sendmsg` — returns IP/port only, no payload.
* type: `skb` on `ip_output` — returns header metadata only.
* type: `char_buf` / `char_iovec` on arg 1 (`msghdr`) of `udp_sendmsg` — either errors with `CharBufErrorBufTooLarge` or reads the struct memory rather than the underlying message data.

@kevsecurity responded in the same thread, confirming that this isn't currently supported in open-source Tetragon, and indicating that DNS-specific support is something the team could probably attempt — likely via a hook type that can reliably access skb packet data (e.g. CGroupSKB) and read it via skb_load_bytes(), paired with some form of payload-matching logic.

A first-class way, inside a TracingPolicy, to observe DNS queries and responses on the workload's network path and emit them as structured fields in the resulting Tetragon event.
Concretely:

* Per-pod and per-process visibility into the domain names being queried, with the standard Tetragon process and Kubernetes context attached.
* Structured output in the gRPC and JSON event schema (queried name, qtype, qclass, transaction ID, rcode), rather than a raw byte buffer the user has to decode.
* Filtering on the queried name and qtype in selectors, so policies can scope events (for example, only *.example.com, or only A and AAAA lookups).
* A TracingPolicy example and accompanying documentation showing per-pod DNS observability end-to-end.

DoT and DoH (encrypted DNS) are out of scope.

No concrete proposal from my side. The Slack thread already surfaced a likely direction (an skb-data-capable hook such as CGroupSKB combined with a small payload-matching mechanism, with DNS as the first concrete consumer), and the design 
is best left to the maintainers.

One possible shape worth mentioning, purely as a discussion starter and not as the expected outcome:

A pluggable `parser` concept attached to an argument or return value in a TracingPolicy. A parser would handle the protocol-specific decoding on the code side and expose structured fields that existing selectors (matchArgs, matchReturnArgs) can reference. New parsers could be added incrementally as use cases come up, without growing the policy schema for each one.

For DNS, a parser would take the relevant payload (for example, arg 1 of udp_sendmsg or udp_recvmsg, reached via msghdr and iovec, or the data portion of an skb) and expose fields like domain, qtype, and qclass to the selector layer.

```
....

selectors:
  - matchArgs:
      - index: 1
        parser: udp_sendmsg_dns     # parser used to interpret the argument
        parsedField: domain         # field exposed by the parser's output object
        operator: "Equal"
        values:
          - "example.com."
      - index: 1
        parser: udp_sendmsg_dns
        parsedField: qtype
        operator: "Equal"
        values:
          - "A"
          
```

The same concept could later extend to uprobes on TLS libraries (for example, SSL_read and SSL_write in OpenSSL), where a parser would decode the decrypted payload and expose application-layer fields such as HTTP method, path, headers or host. This is out of scope for this request and listed only to show that the parser idea generalizes beyond DNS.

### Code of Conduct

litemars:
Introduce a new `dns` KProbeArg type that decodes the first question of DNS-over-UDP messages directly in BPF when attached to udp_sendmsg / udp_recvmsg (and IPv6 variants). The parsed query name, type, class, flags, and tx_id are exposed as a new KprobeDns argument on ProcessKprobe events, and dns_query_name can be matched with the existing string operators (Equal, NotEqual, Prefix, Postfix, SubString). The on-path parser lives in bpf/process/types/dns.h. It validates the DNS header (opcode, qdcount), walks the QNAME with a single bounded loop sized for verifier complexity, and lowercases labels as it writes.

Fixes: #4948 

```release-note
Add a new `dns` kprobe argument type for TracingPolicy that decodes DNS-over-UDP queries directly in BPF, exposing `query_name`/`query_type` and supporting Equal/Prefix/Postfix/SubString selectors on the queried domain.
```

### DIFF
--- a/api/v1/tetragon/tetragon.proto
+++ b/api/v1/tetragon/tetragon.proto
@@ -459,6 +459,18 @@ message SyscallId {
   string abi = 2;
 }
 
+message KprobeDns {
+  string query_name = 1;
+  uint32 query_type = 2;
+  string query_type_str = 3;
+  uint32 query_class = 4;
+  uint32 tx_id = 5;
+  uint32 flags = 6;
+  bool response = 7;
+  bool truncated = 8;
+  bool parsed = 9;
+}
+
 message KprobeArgument {
   oneof arg {
     string string_arg = 1;
@@ -493,6 +505,7 @@ message KprobeArgument {
     KprobeBpfProg bpf_prog_arg = 31;
     KprobeError error_arg = 32;
     KprobeSockaddrUn sockaddrun_arg = 33;
+    KprobeDns dns_arg = 34;
   }
   string label = 18;
 }
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -356,6 +356,11 @@ __read_arg_1(void *ctx, int type, long orig_off, unsigned long arg, int argm, ch
 		size = copy_sockaddr_un(args, arg);
 		break;
 #endif
+#ifdef __LARGE_BPF_PROG
+	case dns_type:
+		size = copy_dns(args, arg);
+		break;
+#endif
 	default:
 		size = 0;
 		break;
--- a/bpf/process/types/basic.h
+++ b/bpf/process/types/basic.h
@@ -14,6 +14,7 @@
 #include "sockaddr_un.h"
 #endif
 #include "socket.h"
+#include "dns.h"
 #include "net_device.h"
 #include "../bpf_process_event.h"
 #include "bpfattr.h"
@@ -101,6 +102,8 @@ enum {
 	bpf_prog_type = 43,
 
 	sockaddr_un_type = 44,
+
+	dns_type = 45,
 
 	nop_s64_ty = -10,
 	nop_u64_ty = -11,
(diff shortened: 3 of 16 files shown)


kkourt (commented):
As a first comment, this PR is relatively large and difficult to review as two commits.
I think splitting the PR into multiple commits will make reviewing easier.

litemars:
Hi @kkourt,

the main modification of the PR comes down to the first commit only. The second commit is just the output of make protogen, make codegen and make generate.

mtardy:
> the main modification of the PR comes down to the first commit only.

hello, this is what could use some splitting.

mtardy:
converting to draft as it doesn't build

litemars:
Hi @kkourt,

I split up the commit into multiple small ones - please let me know if you have any question about this? feel free to close the PR if this feature is not needed anymore

PR #5153, opened 2026-06-18
Makefile.defs: update BUF_BREAKING_AGAINST_BRANCH

andrewstrohman:
To reflect the fact that we should be comparing against the release branch.

### DIFF
--- a/Makefile.defs
+++ b/Makefile.defs
@@ -13,6 +13,6 @@ GO ?= go
 CONTAINER_ENGINE ?= docker
 
 BUF ?= buf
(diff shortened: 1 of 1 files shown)


andrewstrohman:
[This](https://github.com/cilium/tetragon/pull/5152#issuecomment-4743366827) is how this came to my attention.

tpapagian (approved):

andrewstrohman:
[here](https://github.com/cilium/tetragon/pull/5154) is the fix for the windows CI failure.

PR #5161, opened 2026-06-19
cleanup(pkg/sensors): refactor sensors dup instance code

FedeDP:
### Description

Seeing https://github.com/cilium/tetragon/pull/5114, and considering we need the same fix for other sensors, make it easier to embed it into other sensors.

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -87,7 +87,7 @@ type genericKprobe struct {
 	argSigPrinters    []argPrinter
 	argReturnPrinters []argPrinter
 	funcName          string
-	instance          int
+	instance          InstanceID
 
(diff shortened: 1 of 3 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -81,13 +81,40 @@ type genericKprobeData struct {
 	stackTraceMap *program.Map
 }
 
+type InstanceID int
```
olsajiri: maybe `pkg/sensors/tracing/generic.go` or separate file is better place?

olsajiri (approved):
left one comment, otherwise lgtm

pkg/sensors/tracing/generickprobe.go
FedeDP: Thanks, you are right! Let me move it

PR #5166, opened 2026-06-19
feat: add support for nodeSelector in tracing policy

### LINKED ISSUE #4935: Add nodeSelector support to TracingPolicy and TracingPolicyNamespaced

_No response_

Tetragon TracingPolicy resources currently apply broadly to all nodes running Tetragon, unless scoped by existing workload-level selectors such as podSelector, containerSelector etc.

However, there is no policy-level way to restrict a TracingPolicy to only the Kubernetes nodes that match a given node label selector and apply policies to all workloads running on a node group. This makes it difficult to target policies at specific node groups, such as worker nodes, canary nodes, high-risk nodes, GPU nodes, or nodes in a particular pool with a different architecture (arm, amd etc) or OS. 

This is needed because some tracing policies may be expensive, noisy, or only relevant on a subset of nodes. Without a node-level selector, users must rely on operational workarounds such as: deploying different Tetragon configurations per node group, manually applying and removing policies,
or accepting that the policy is loaded cluster-wide even when only some nodes need it. This limits policy targeting and makes gradual rollout or node-pool-specific tracing harder to manage.

_No response_

### Code of Conduct

SamrathPalSingh:
Fixes #4935 

### Description

Add `nodeSelector` support to `TracingPolicy` so cluster-scoped tracing policies can be limited to nodes matching Kubernetes node labels.

Previously, a `TracingPolicy` was considered for all nodes. This change allows users to target policies more precisely by setting `spec.nodeSelector`, using the same label selector semantics as Kubernetes. An unset or empty selector continues to match all nodes.

This also adds unit coverage for the node matching behavior, including:

- unset selectors matching all nodes
- empty selectors matching all nodes
- `matchLabels` selecting and rejecting nodes
- `matchExpressions` selecting and rejecting nodes
- invalid selector operators returning an error

### Changelog

```release-note
Add nodeSelector support to TracingPolicy for limiting cluster-scoped tracing policies to matching Kubernetes nodes.

### DIFF
--- a/install/kubernetes/tetragon/crds-yaml/cilium.io_tracingpolicies.yaml
+++ b/install/kubernetes/tetragon/crds-yaml/cilium.io_tracingpolicies.yaml
@@ -3609,6 +3609,61 @@ spec:
                   - hook
                   type: object
                 type: array
+              nodeSelector:
+                description: NodeSelector selects nodes that this policy applies to.
+                properties:
+                  matchExpressions:
+                    description: matchExpressions is a list of label selector requirements.
+                      The requirements are ANDed.
+                    items:
+                      description: |-
+                        A label selector requirement is a selector that contains values, a key, and an operator that
+                        relates the key and values.
+                      properties:
+                        key:
+                          description: key is the label key that the selector applies
+                            to.
+                          type: string
+                        operator:
+                          description: |-
+                            operator represents a key's relationship to a set of values.
+                            Valid operators are In, NotIn, Exists and DoesNotExist.
+                          enum:
+                          - In
+                          - NotIn
+                          - Exists
+                          - DoesNotExist
+                          type: string
+                        values:
+                          description: |-
+                            values is an array of string values. If the operator is In or NotIn,
+                            the values array must be non-empty. If the operator is Exists or DoesNotExist,
+                            the values array must be empty. This array is replaced during a strategic
+                            merge patch.
+                          items:
+                            type: string
+                          type: array
+                          x-kubernetes-list-type: atomic
+                      required:
+                      - key
+                      - operator
+                      type: object
+                    type: array
+                    x-kubernetes-list-type: atomic
+                  matchLabels:
+                    additionalProperties:
+                      description: MatchLabelsValue represents the value from the
+                        MatchLabels {key,value} pair.
+                      maxLength: 63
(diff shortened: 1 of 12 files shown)


olsajiri (changes requested):
please split your change in logical commits
- behaviour change
- generated change
- tests
- ...
with each commit having changelog describing the change

it's harder to find reviewer than to prompt AI these days, so the idea is to save reviewer's time ;-)

SamrathPalSingh:
adding description to help with the review:

### feat: support tracing policy nodeSelector
- Added optional `.spec.nodeSelector` support to `TracingPolicy` and `TracingPolicyNamespaced`.
- Policies now only apply to the local node when their `nodeSelector` matches the node labels.
- Nil or omitted `nodeSelector` keeps existing behavior and applies the policy to all nodes.
- Add/update handling now evaluates the selector before adding, replacing, or removing tracing policies from the sensor manager.
- Added node update reconciliation so policy application is adjusted when local node labels change

### chore: regenerate tracing policy CRDs
- Regenerated tracing policy CRD YAMLs to expose the new `nodeSelector` field in the public Kubernetes schema.
- Updated generated deepcopy code so `TracingPolicySpec.NodeSelector` is copied correctly.
- Mirrored the generated API/CRD updates into the vendored Tetragon API files.

### test: tracing policy nodeSelector matching
- Added unit coverage for tracing policy node selector evaluation.

sayboras:
Hi @SamrathPalSingh, thank you for picking this up and for the thorough test coverage — much appreciated! 🙏                                                                                                                                  
                                                                                                                                                                                                                                                
A quick heads-up on timing: the TracingPolicy watch layer was refactored on main fairly recently. The hand-rolled informer in pkg/watcher/crdwatcher/tracingpolicy.go that this PR builds on has been migrated to controller-runtime          
  reconcilers (tracingpolicy_reconciler.go and tracingpolicy_namespaced_reconciler.go), so the integration point for the nodeSelector gating has moved quite a bit. As-is, this PR would need a non-trivial rebase onto the new structure.      
                                                                                                                                                                                                                                                
Since this overlaps with some in-flight work in that area, I'll take it forward from here and carry the nodeSelector support onto the new controller-runtime reconcilers (including handling node re-labelling at runtime).  https://github.com/cilium/tetragon/pull/5189                                                                                  
                                                                                                                                                                                                                                                
Thanks again for moving this forward, and apologies that the refactor landed on top of your PR!

PR #5171, opened 2026-06-22
tetragon: sleepable_preload fixes

olsajiri:
make sleepable_preload bpf map marked as not preloaded in the kernel
and make it shared between all the policies by default


kkourt (approved):
LGTM, thanks!

FedeDP (approved):

PR #5173, opened 2026-06-22
fix: skip malformed stdin events in tetra getevents

immanuwell:
Fixes none

### Description
`tetra getevents` reading from stdin could turn a bad json line into an empty event and print `{}` when debug is off. Small papercut, but it can mess with pipes.

This skips lines that fail unmarshal, same as debug mode already kinda did, and adds a regression test.

Repro:
```
go build -o /tmp/tetra ./cmd/tetra
printf 'not-json\n{"process_exec":{"process":{"binary":"/usr/bin/netserver"}}}\n' | /tmp/tetra getevents
```

Before:
```
{}
{"process_exec":{"process":{"binary":"/usr/bin/netserver"}}}
```

After:
```
{"process_exec":{"process":{"binary":"/usr/bin/netserver"}}}
```

### Changelog
```release-note
tetra getevents now skips malformed JSON lines from stdin instead of emitting empty events
```

### DIFF
--- a/cmd/tetra/getevents/io_reader_client.go
+++ b/cmd/tetra/getevents/io_reader_client.go
@@ -112,8 +112,10 @@ func (i *ioReaderClient) Recv() (*tetragon.GetEventsResponse, error) {
 		res := &tetragon.GetEventsResponse{}
 		line := i.scanner.Bytes()
 		err := i.unmarshaller.Unmarshal(line, res)
-		if err != nil && i.debug {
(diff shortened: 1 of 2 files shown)


mtardy (approved):
looks good to me, thanks for the fix and the associated test :)

ExceptionalHandler (approved):

FedeDP (approved):

PR #5177, opened 2026-06-23
btf : Add type casting in `resolve`

tdaudi:
## Description

- Add support to cast a resolve attributes.
- Add support to use `pt_regs` along with the resolve 

It works as follow

```yaml
- index: 0
  type: "uint8"
  resolve: "((char *)buffer)[12]"
```

## Motivation

We need support to retrieve values from `rsp`. By default it's not accessible because the types are `unsigned long`. This PR allow to cast `rsp` to a `char*` and then act as an array to get the value we need.

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/hooks.md
+++ b/docs/content/en/docs/concepts/tracing-policy/hooks.md
@@ -777,6 +777,47 @@ spec:
       resolve: "sin_addr.s_addr"
 ```
 
+#### Type casts in resolve paths
+
+The `resolve` path can cast an intermediate field to another BTF type before
+continuing the resolution. This is useful when a field is typed as a generic
+pointer, such as `void *`, but the policy author knows the concrete structure
+stored at runtime.
+
+Type casts use a C-style syntax:
+
+```text
+((type)field-or-subpath).next_field
+```
+
+```yaml
+args:
+- index: 1
+  btfType: "foo"
+  type: "uint32"
+  resolve: "((struct bar *)myparam).myvalue"
+```
+
+In this example, Tetragon resolves `myparam`, treats the value as a
+`struct bar *`, dereferences it, and then reads the `myvalue` field.
+
+The cast is part of the `resolve` path, so it can also be used with indexes:
+
+```yaml
+args:
+- index: 0
+  type: "uint8"
+  resolve: "((char *)buffer)[12]"
+```
+
+This resolves `buffer`, treats it as a `char *`, and reads element `12`.
+
+{{< note >}}
+For uprobes and USDT probes, type names used in casts are resolved first from
+the BTF file configured with `spec.uprobes.btfPath`. If the type is not found there, Tetragon
+falls back to the kernel vmlinux BTF.
+{{< /note >}}
+
 #### Kernel module BTF types
 
 For kprobe arguments, use `btfTypeModule` with `btfType` when the structure is
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -272,6 +272,7 @@ type btfResolver struct {
 	btfArgs     *[api.MaxBTFArgDepth]api.ConfigBTFArg
 	currentType btf.Type
 	pathToFound []string
+	spec        *btf.Spec
 }
 
 // ResolveBTFPath function recursively search in a btf structure in order to
@@ -292,11 +293,13 @@ func ResolveBTFPath(
 	btfArgs *[api.MaxBTFArgDepth]api.ConfigBTFArg,
 	currentType btf.Type,
 	pathToFound []string,
+	spec *btf.Spec,
 ) (*btf.Type, error) {
 	resolver := btfResolver{
 		btfArgs:     btfArgs,
 		currentType: currentType,
 		pathToFound: pathToFound,
+		spec:        spec,
 	}
 	return resolver.resolve(currentType, 0)
 }
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -28,6 +28,8 @@ import (
 
 var (
 	btfFile string
+
+	errNotACast = errors.New("not a cast")
 )
 
 func btfFileExists(file string) error {
@@ -309,6 +311,12 @@ func (r *btfResolver) resolve(
 	i int,
 ) (*btf.Type, error) {
 	currentType = ResolveNestedTypes(currentType)
+	if castType, err := r.parseBTFTypeCast(i); !errors.Is(err, errNotACast) {
+		if err != nil {
+			return nil, err
+		}
+		return r.resolve(castType, i)
+	}
 	switch t := currentType.(type) {
 	case *btf.Struct:
 		return r.processMembers(currentType, t.Members, i)
@@ -429,4 +437,37 @@ func (r *btfResolver) processArray(
 		r.btfArgs[i].IsPointer = uint16(1)
 	}
 	return &targetType, nil
+}
+
+func (r *btfResolver) parseBTFTypeCast(i int) (btf.Type, error) {
+	pathElem := r.pathToFind[i]
+
+	if !strings.HasPrefix(pathElem, "(") || !strings.HasSuffix(pathElem, ")") {
+		if strings.ContainsAny(pathElem, "()") {
+			return nil, fmt.Errorf("invalid BTF type cast %q (must be: \"(type)\")", pathElem)
+		}
+		return nil, errNotACast
+	}
+
+	typeExpr := strings.TrimSpace(pathElem[1 : len(pathElem)-1])
+	if typeExpr == "" {
+		return nil, errors.New("empty BTF type cast")
+	}
+
+	if r.spec == nil {
+		spec, err := NewBTF()
+		if err != nil {
+			return nil, err
+		}
+		r.spec = spec
+	}
+
+	ty, err := ParseBTFType(r.spec, typeExpr)
+	if err != nil {
+		return nil, err
+	}
+
+	r.pathToFind = append(r.pathToFind[:i], r.pathToFind[i+1:]...)
+
+	return ty, nil
 }
--- a/pkg/btf/btf.go
+++ b/pkg/btf/btf.go
@@ -268,47 +268,63 @@ func getSizeofType(t btf.Type) uint32 {
 	return uint32(ret)
 }
 
+type btfResolver struct {
+	btfArgs     *[api.MaxBTFArgDepth]api.ConfigBTFArg
+	currentType btf.Type
+	pathToFound []string
+}
+
 // ResolveBTFPath function recursively search in a btf structure in order to
 // found a specific path until it reach the target or fail.
-
+//
 // The function also search in embedded anonymous structures or unions to cover as
 // much use cases as possible. For instance, mm_struct have 2 fields, anonymous
 // struct and another type. But you are still able to look into the anonymous
 // struct by specifying a path like "mm.pgd.pgd".
-
+//
 // @btfArgs: dest array for storing btf informations to reach the target on the
 // bpf side.
 // @currentType: The current type being proccessed, starts with root type.
 // @pathToFound: The string representation of the path to reach in the structures.
-// @i: The current depth, until last element of pathToFound.
-
+//
 // Return: The last type found matching the path, or error.
 func ResolveBTFPath(
 	btfArgs *[api.MaxBTFArgDepth]api.ConfigBTFArg,
 	currentType btf.Type,
 	pathToFound []string,
+) (*btf.Type, error) {
+	resolver := btfResolver{
+		btfArgs:     btfArgs,
+		currentType: currentType,
+		pathToFound: pathToFound,
+	}
+	return resolver.resolve(currentType, 0)
+}
+
+func (r *btfResolver) resolve(
+	currentType btf.Type,
 	i int,
 ) (*btf.Type, error) {
 	currentType = ResolveNestedTypes(currentType)
 	switch t := currentType.(type) {
 	case *btf.Struct:
-		return processMembers(btfArgs, currentType, t.Members, pathToFound, i)
+		return r.processMembers(currentType, t.Members, i)
 	case *btf.Union:
-		return processMembers(btfArgs, currentType, t.Members, pathToFound, i)
+		return r.processMembers(currentType, t.Members, i)
 	case *btf.Pointer:
 		if i > 0 {
 			// To avoid adding an extra BTFArg entry only for doing this
 			// dereferencing, we mark the previous element as pointer.
-			btfArgs[i-1].IsPointer = uint16(1)
+			r.btfArgs[i-1].IsPointer = uint16(1)
 		}
-		if idx, err := parseArrayIdxStr(pathToFound[i]); err == nil {
+		if idx, err := parseArrayIdxStr(r.pathToFound[i]); err == nil {
 			// To stay ahead on the dereferecing, we mark the current btfArg as pointer
-			btfArgs[i].IsPointer = uint16(1)
-			return processArray(btfArgs, t.Target, pathToFound, i, idx)
+			r.btfArgs[i].IsPointer = uint16(1)
+			return r.processArray(t.Target, i, idx)
 		}
-		return ResolveBTFPath(btfArgs, t.Target, pathToFound, i)
+		return r.resolve(t.Target, i)
 	case *btf.Array:
-		idx, err := parseArrayIdxStr(pathToFound[i])
+		idx, err := parseArrayIdxStr(r.pathToFound[i])
 		if err != nil {
 			return nil, fmt.Errorf("fail parsing array index : %w", err)
 		}
@@ -317,15 +333,15 @@ func ResolveBTFPath(
 		if t.Nelems != 0 && idx >= t.Nelems {
 			return nil, fmt.Errorf("array index out of bound. Nelems=%d, got=%d", t.Nelems, idx)
 		}
-		return processArray(btfArgs, t.Type, pathToFound, i, idx)
+		return r.processArray(t.Type, i, idx)
 	default:
 		ty := currentType.TypeName()
 		if len(ty) == 0 {
 			ty = reflect.TypeOf(currentType).String()
 		}
-		currentPath := pathToFound[i]
+		currentPath := r.pathToFound[i]
 		if i > 0 {
-			currentPath = pathToFound[i-1]
+			currentPath = r.pathToFound[i-1]
 		}
 		return nil, fmt.Errorf("unexpected type : %q has type %q", currentPath, ty)
 	}
@@ -340,17 +356,15 @@ func (e *resolveError) Error() string {
 	return e.str
 }
 
-func processMembers(
-	btfArgs *[api.MaxBTFArgDepth]api.ConfigBTFArg,
+func (r *btfResolver) processMembers(
 	currentType btf.Type,
 	members []btf.Member,
-	pathToFound []string,
 	i int,
 ) (*btf.Type, error) {
 	var lastError *resolveError
 	for _, member := range members {
 		if len(member.Name) == 0 { // anonymous struct/union, fallthrough
-			lastTy, err := ResolveBTFPath(btfArgs, member.Type, pathToFound, i)
+			lastTy, err := r.resolve(member.Type, i)
 			if err != nil {
 				// Propagate the deepest error for both resolve and non-resolve error.
 				if err2, ok := errors.AsType[*resolveError](err); ok {
@@ -362,24 +376,24 @@ func processMembers(
 				}
 				continue
 			}
-			btfArgs[i].Offset += member.Offset.Bytes()
+			r.btfArgs[i].Offset += member.Offset.Bytes()
 			return lastTy, nil
 		}
-		if member.Name != pathToFound[i] {
+		if member.Name != r.pathToFound[i] {
 			continue
 		}
-		btfArgs[i].Offset = member.Offset.Bytes()
-		btfArgs[i].IsInitialized = uint16(1)
-		if i < len(pathToFound)-1 && i < api.MaxBTFArgDepth {
-			return ResolveBTFPath(btfArgs, member.Type, pathToFound, i+1)
+		r.btfArgs[i].Offset = member.Offset.Bytes()
+		r.btfArgs[i].IsInitialized = uint16(1)
+		if i < len(r.pathToFound)-1 && i < api.MaxBTFArgDepth {
+			return r.resolve(member.Type, i+1)
 		}
 		memberType := ResolveNestedTypes(member.Type)
 		switch t := memberType.(type) {
 		case *btf.Pointer:
-			btfArgs[i].IsPointer = uint16(1)
+			r.btfArgs[i].IsPointer = uint16(1)
 			memberType = t.Target
 		case *btf.Int, *btf.Enum:
-			btfArgs[i].IsPointer = uint16(1)
+			r.btfArgs[i].IsPointer = uint16(1)
 		}
 		return &memberType, nil
 	}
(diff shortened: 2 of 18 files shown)


tdaudi:
The CI failure https://github.com/cilium/tetragon/actions/runs/28157736041/job/83390331898?pr=5177 doesn't seems related.

pkg/btf/type_builder.go
```diff
@@ -0,0 +1,319 @@
...
+	fallbackSpec     *btf.Spec
+	errFallbackSpec  error
+)
+
+var loadFallbackBTFSpec = func() (*btf.Spec, error) {
+	fallbackSpecInit.Do(func() {
+		fallbackSpec, errFallbackSpec = NewBTF()
+	})
+	return fallbackSpec, errFallbackSpec
+}
```
mtardy: don't we already have a global with `btf.Spec` already in the same package?

pkg/btf/btf.go
```diff
@@ -280,35 +286,45 @@ func getSizeofType(t btf.Type) uint32 {
 // bpf side.
 // @currentType: The current type being proccessed, starts with root type.
 // @pathToFound: The string representation of the path to reach in the structures.
-// @i: The current depth, until last element of pathToFound.
 
 // Return: The last type found matching the path, or error.
 func ResolveBTFPath(
```
mtardy, line 294: could you make this comment correct to be associated with the function, either add empty comment line `//` or remove the whiteline between the comments otherwise

pkg/btf/btf.go
```diff
@@ -430,3 +436,28 @@ func (r *btfResolver) processArray(
 	}
 	return &targetType, nil
 }
+
+func (r *btfResolver) parseBTFTypeCast(i int) (btf.Type, bool, error) {
+	pathElem := r.pathToFound[i]
```
mtardy: should you do a check to avoid potential panic here?

pkg/btf/btf.go
```diff
@@ -268,6 +268,12 @@ func getSizeofType(t btf.Type) uint32 {
 	return uint32(ret)
 }
 
+type btfResolver struct {
+	btfArgs     *[api.MaxBTFArgDepth]api.ConfigBTFArg
+	currentType btf.Type
+	pathToFound []string
```
mtardy: pathToFind maybe? not sure to understand what it's suppose to hold

pkg/btf/btf_test.go
```diff
@@ -681,6 +681,111 @@ func TestParseArrayIdxStr(t *testing.T) {
...
+			if tt.wantErr {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+
+			wantPath := tt.input
+			if tt.wantOK && !tt.wantErr {
+				require.NotNil(t, resolver.spec, "should have initialized resolver.spec")
+				wantPath = append(append([]string(nil), tt.input[:tt.idx]...), tt.input[tt.idx+1:]...)
```
mtardy: I think it would be a bit easier to read like that
```
input := slices.Clone(tt.input)
wantPath = append(input[:tt.idx], input[tt.idx+1:]...)
```

pkg/sensors/tracing/generic.go
```diff
@@ -23,55 +23,97 @@ import (
...
-		switch r {
-		case '.':
-			if inBracket || i > 0 && resolvePath[i-1] == '.' {
-				invalidFormat = true
-				break
+	i := 0
+	lastWasDot := false
+
+	var parse func(stopAtCloseParen bool) error
+	parse = func(stopAtCloseParen bool) error {
```
mtardy, line 36: honestly this function is really complicated to review if not impossible taking a lot of time. Maybe you should:
1. make it an actual function instead of anonymous one
2. add some fuzzing harness for your parser function to make sure it doesn't crash?

pkg/sensors/tracing/generic.go
mtardy, line 1: small typos in your git commit
```patch
generic : Refactor resolve for PtRegs
- When we have `resolve: rsp`, we know that rsp as type `unsigned long`.
+ When we have `resolve: rsp`, we know that rsp has type `unsigned long`.
- In that case we want to cast it's value to a `char*` to go through it's
+ In that case we want to cast its value to a `char*` to go through its
```

But I had trouble understanding what you meant, is it what you mean?
> rsp is typed like an integer, but we want to use its value as a pointer. So we strip rsp from the path, cast the value to char *, and let the resolver continue from that casted pointer type. Void is just a dummy starting type because the cast will overwrite it.

Could you give an example?

pkg/sensors/tracing/generic.go
```diff
@@ -144,6 +146,44 @@ func hasPtRegsSource(arg *v1alpha1.KProbeArg) bool {
...
+		return regArg, btfArg, false, fmt.Errorf("failed to retrieve register argument %q", resolve)
+	}
+
+	path = path[1:]
+	if len(path) == 0 {
+		return regArg, btfArg, false, nil
+	}
+	if !bpf.HasProgramLargeSize() {
+		return regArg, btfArg, false, errors.New("resolve flag can't be used for your kernel version. Please update to version 5.4 or higher or disable Resolve flag")
+	}
```
mtardy, line 175: should you have a dedicated check for resolve feature? do you have any other place where you return the same error maybe?

pkg/sensors/tracing/generic.go
```diff
@@ -144,6 +146,44 @@ func hasPtRegsSource(arg *v1alpha1.KProbeArg) bool {
 	return arg.Source == "pt_regs"
 }
 
+func resolvePtRegsArg(resolve string) (api.ConfigRegArg, [api.MaxBTFArgDepth]api.ConfigBTFArg, bool, error) {
```
mtardy: could you maybe document that function, it's not easy to understand

pkg/sensors/tracing/generic_test.go
```diff
@@ -43,6 +43,42 @@ func TestFormatBTFPath(t *testing.T) {
 			want:    []string{"path", "to", "[123]", "my", "field"},
 			wantErr: false,
 		},
+		{
+			name:    "reject suffix type cast",
+			input:   "rsp(char*)[16]",
+			want:    []string{"rsp", "[16]"},
+			wantErr: true,
```
mtardy, line 50: the function is both returning an error and a result?

pkg/sensors/tracing/generic_test.go
```diff
@@ -43,6 +43,42 @@ func TestFormatBTFPath(t *testing.T) {
...
+		},
+		{
+			name:    "reject suffix pointer to array type cast",
+			input:   "rsp(*char[64])[16]",
+			want:    []string{"rsp", "(*char[64])", "[16]"},
+			wantErr: true,
```
mtardy, line 56: same here on both error and result

pkg/sensors/tracing/generic.go
mtardy: maybe adding test for this function would make it easier to understand?

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -777,6 +777,47 @@ spec:
...
+- index: 0
+  type: "uint8"
+  resolve: "((char *)buffer)[12]"
+```
+
+This resolves `buffer`, treats it as a `char *`, and reads element `12`.
+
+{{< note >}}
+For uprobes and USDT probes, type names used in casts are resolved first from
+the BTF file configured with `btfPath`. If the type is not found there, Tetragon
```
mtardy: > the BTF file configured with `btfPath`

that might not be very transparent for the user, what does it corresponds to?

pkg/sensors/tracing/generic.go
```diff
@@ -113,23 +195,18 @@ func resolveBTFType(arg *v1alpha1.KProbeArg, ty ebtf.Type) (*ebtf.Type, [api.Max
...
-func resolveUserBTFArg(arg *v1alpha1.KProbeArg, btfPath string) (*ebtf.Type, [api.MaxBTFArgDepth]api.ConfigBTFArg, error) {
-	spec, err := ebtf.LoadSpec(btfPath)
-	if err != nil {
-		return nil, [api.MaxBTFArgDepth]api.ConfigBTFArg{}, err
-	}
-
+func resolveUserBTFArg(arg *v1alpha1.KProbeArg, userBTFSpec *ebtf.Spec) (*ebtf.Type, [api.MaxBTFArgDepth]api.ConfigBTFArg, error) {
 	var st *ebtf.Struct
-	err = spec.TypeByName(arg.BTFType, &st)
+	err := userBTFSpec.TypeByName(arg.BTFType, &st)
```
mtardy, line 204: this panics if userBTFSpec is nil

pkg/btf/btf_test.go
```diff
@@ -681,6 +681,111 @@ func TestParseArrayIdxStr(t *testing.T) {
...
+			idx:     0,
+			wantOK:  false,
+			wantErr: false,
+		},
+		{
+			name:    "Invalid format - trailing token",
+			input:   []string{"field", "(char ')*"},
+			idx:     1,
+			wantOK:  true,
+			wantErr: true,
```
mtardy: what does it mean semantically that your function return an error and still return ok?

pkg/btf/btf.go
```diff
@@ -280,35 +286,45 @@ func getSizeofType(t btf.Type) uint32 {
...
+) (*btf.Type, error) {
+	resolver := btfResolver{
+		btfArgs:     btfArgs,
+		currentType: currentType,
+		pathToFound: pathToFound,
+	}
+	return resolver.resolve(currentType, 0)
+}
+
+func (r *btfResolver) resolve(
```
mtardy, line 309: imho it would be easier to pass your BTF spec as args to function as it's easier to follow its lifecycle that way instead of creating struct and methods. Also it makes the recursive nature easier to follow with pure functions.

pkg/btf/btf.go
```diff
@@ -430,3 +436,28 @@ func (r *btfResolver) processArray(
 	}
 	return &targetType, nil
 }
+
+func (r *btfResolver) parseBTFTypeCast(i int) (btf.Type, bool, error) {
```
mtardy: I feel like it would be easier to understand as a pure function nop? why a method there?

mtardy (commented):
Hey thanks. Here are a few comments, generally I feel like this is a bit tricky to follow without high context on this codebase.

You have multiple functions doing parsing of strings now, I think it's globally a good idea to isolate them as much as possible and provide comprehensive testing and fuzzing for each of them. I think it's already kinda the case in some places.

Also I'm a bit confused on the ordering:
- `btf : Add support to override the BTF type` is added first and seems that parseBTFTypeCast will parse a determined position element in path, get its BTF type and then remove it from the path?
- `sensors/tracing: support prefixed BTF casts in resolve paths` to actually parse the type casting in the strings is added later?

Maybe we should have a chat so that you can explain me this PR :)

pkg/btf/type_builder.go
tdaudi: I don't think we're caching the file. I think it's reloaded every time
https://github.com/cilium/tetragon/blob/d20ea1f27ca596bc8abcaa77a9a4c3ff14942e0c/pkg/sensors/tracing/generickprobe.go#L456-L460

Maybe this fallback function is not the best approach, but if you have suggestions i'd be happy to change things

pkg/btf/btf.go
tdaudi: I don't think `i` can be out of bound here. I'd say it's fine. Did I missed something ?

pkg/btf/btf.go
tdaudi: Indeed maybe a little typo issue added on the initial implem of the algo, will update

pkg/sensors/tracing/generic.go
tdaudi: > add some fuzzing harness for your parser function to make sure it doesn't crash?

The commit includes a lot of different test cases, I think it's important that it doesn't break anything and we assert all expected input/output behave has expected, which I think is the case here.

> make it an actual function instead of anonymous one

Because it's a recursive function I choose to keep it "small" with everything in the same function. But the string parsing logic to make this work is also not trivial to do. I feel like have both the `parse` function and `formatBTFPath` separated will make it more complex. Especially because the file already have various unrelated helpers.

pkg/sensors/tracing/generic.go
tdaudi: > small typos in your git commit

Thanks !

> Could you give an example?

When we have `resolve: rsp` what we get is a `uint64` value representing a pointer. Because we know it's a pointer we can turn it into one with the cast. So step by step it do exactly what the resolve algo ingest

`[]string{"rsp", "(char*)", "[123]"}`

Our entry point is a `uint64`, we turn it into a `char*` and then act as if it's an array to get the data at offset `123`.

pkg/sensors/tracing/generic.go
tdaudi: Indeed this check is done at other places of the code. We could delay this check to the entry point of the resolve algo in `btf.ResolveBTFPath` but that involves a lot of unnecessary code processing prior to returning an error which i'm not a big fan of (for example in `resolveBTFArg` function).

https://github.com/cilium/tetragon/blob/3620c1776ee67f784a0a6213f0afc156ad828a8b/pkg/sensors/tracing/generic.go#L159

pkg/sensors/tracing/generic.go
tdaudi: Not sure I follow. I'm not sure this case can happen. Let me know if I'm missing something here

pkg/btf/btf_test.go
tdaudi: The ok here is used to determine if this is a cast to perform or no, where the error expect an error returned by the function.
So in that case we need both to make sure it fits the requirements in `ResolveBTFPath`

pkg/btf/btf.go
tdaudi: I feel like as the function grow it's less readable to pass as arguments a lot of parameters. Especially when they are not part of the resolve iterations but are static for all the duration of the function.
IMO it's a little bit more cleaner this way and it also helps for future work on the algo if we want to modify something with less changes

pkg/btf/btf.go
tdaudi: I wanted something independent from the resolve logic and that act as the entry point of the BTF casting workflow.
This way it's isolated from the resolve logic and can be tested as one independently in `TestParseBTFTypeCast`

pkg/sensors/tracing/generic_test.go
tdaudi: Thanks, I missed that one

pkg/sensors/tracing/generic_test.go
tdaudi: Miss that one too, thanks

pkg/btf/type_builder.go
```diff
@@ -0,0 +1,319 @@
...
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/cilium/ebpf/btf"
+)
+
+type btfTypeParser struct {
```
olsajiri, line 19: would it make sense to inherit from `pkg/cursorparser/parser.go` ?

pkg/btf/btf.go
```diff
@@ -280,35 +287,53 @@ func getSizeofType(t btf.Type) uint32 {
...
+	}
+	return resolver.resolve(currentType, 0)
+}
+
+func (r *btfResolver) resolve(
+	currentType btf.Type,
 	i int,
 ) (*btf.Type, error) {
 	currentType = ResolveNestedTypes(currentType)
+	if castType, ok, err := r.parseBTFTypeCast(i); ok {
```
olsajiri: returning `ok` and `err` seems redundant

also this can return `nil, false, nil` .. wouldn't it panic on the  `r.resolve(castType, i)` call below

pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -820,6 +896,87 @@ func TestUprobePtRegsDataNotMatch(t *testing.T) {
...
+	//      <+21>: mov    rax,QWORD PTR [rip+0x2df4]        # stack_source
+	//      <+28>: mov    QWORD PTR [rsp+0x10],rax
+	//      <+33>: call   stack_slot_barrier
+	// -->  <+38>: mov    rax,QWORD PTR [rsp+0x10]
+	//      <+43>: add    rsp,0x28
+	//      <+47>: ret
+	//
+	// Make sure we resolve the expected value from the stack slot
+	// at [rsp+0x10] via the current stack pointer.
+	symbol := "stack_slot_value+38"
```
olsajiri: I don't think it's good idea to rely on particular compilation layout, could change with other systems build flags and compiler versions.. please use inline asm as we did for the other tests

pkg/btf/btf.go
tdaudi: Mahé suggested the same thing in https://github.com/cilium/tetragon/pull/5177/changes#r3513761882.
The problem is that there is 2 different things, the first is "is this a type to cast?" and the second "does an error occurs will parsing?". If we only keep the error, we'll have 2 options, skip the error message to avoid killing the resolve algo, or doing an error filtering to only return the error if it's not "not a cast" (or something similar).
What we can do instead is having a second function (ex : isACast) that return a boolean and call parseBTFTypeCast after. 

```go
if isACast(pathToFind[i]) {
    castType,, err := r.parseBTFTypeCast(i)
    if err != nil {
	  return nil, err
	}
	return r.resolve(castType, i)
}
```

olsajiri: could we add special error for no cast, like errNext in pkg/asm/assignment_linux.go ?

tdaudi: Sounds good, thanks for the advice, will update

pkg/btf/type_builder.go
tdaudi: That make a lot of sense, I add 3 helpers in cursorparser to help with the parsing and pushed it in the same commit as `type_builder.go`. Let me know if you prefer having it separated.

pkg/btf/btf.go
tdaudi: Done, thanks

pkg/sensors/tracing/uprobe_reg_test.go
tdaudi: Indeed I didn't thought about that, I changed it and it should be fine, thanks

pkg/btf/type_builder.go
kkourt: The reason we don't cache the file is that it wastes a lot of memory: https://github.com/cilium/tetragon/pull/2937.

tdaudi: That makes a lot of sense. So in the latest push I removed this part of the code and choose to load it in `parseBTFType` instead. The file will be loaded only during the `resolve` phase and only if it includes a cast in the field.

```go
func (r *btfResolver) parseBTFTypeCast(i int) (btf.Type, error) {
...
	if r.spec == nil {
		spec, err := NewBTF()
		if err != nil {
			return nil, err
		}
		r.spec = spec
	}
...
}
```

However, we also could refactor the code to load the BTF file in [policyHandler.PolicyHandler](https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/policyhandler_linux.go#L225) (in `sensors/tracing/policyhandler_linux.go`) to load the BTF file, pass it as an argument to `preValidateKprobes` and `createGenericKprobeSensor` (and co.). But there is quite some changes to do just to bring the file into the resolve algo.

Let me know if one or the approach makes more sense

docs/content/en/docs/concepts/tracing-policy/hooks.md
tdaudi: I add a fix, let me know if that's better, thanks

pkg/btf/btf_test.go
tdaudi: Indeed it's not ideal, it should be fixed, thanks

pkg/btf/btf.go
tdaudi: Done, thanks

pkg/btf/type_builder.go
```diff
@@ -0,0 +1,261 @@
...
+
+// ParseBTFType builds a BTF type from a C-like type expression.
+//
+// Supported constructs are primitive types and BTF spec types. For example:
+//
+//   - "char*"
+//   - "(*char[64])"
+//   - "uint64_t[4]"
+//   - "struct task_struct *"
+func ParseBTFType(spec *btf.Spec, typeExpr string) (btf.Type, error) {
```
kkourt, line 32: I'm guessing this is used to parse the type in the casting?

Can we have clear semantics for the clear here? "C-like type expression" leaves some ambiguity and, for example, it's not clear to me what "(*char[64])" is. Note also, that "uint64_t[4]" is not something you can use when casting in C.

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -777,6 +777,47 @@ spec:
       resolve: "sin_addr.s_addr"
 ```
 
+#### Type casts in resolve paths
```
kkourt, line 780: Let's use this opportunity to properly document the resolve operator, at the top level (I believe this is in `### Attribute resolution`) so that we clearly define the syntax and semantics of the operator.

kkourt (commented):
Some initial comments from my side, thanks!

pkg/btf/type_builder.go
tdaudi: > Can we have clear semantics for the clear here? "

The logic of the string parsing is splitted in 2 parts :

- The `[formatBTFPath](https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/generic.go#L25-L76)` in generic.go format the string into an array  of "pathToFind" that the resolve will consome. This part has the responsibility to both split  the string and make sure the types are correctly formatted. Please see the tests in https://github.com/cilium/tetragon/pull/5177/changes/20c5289d9e0d29585dfcdb3696eafd32e606ede6. The advantage of this approach is that we are returning the error early without waiting to reach `ParseBTFType` and see that something is not formatted correctly.
- The `ParseBTFType` is more generic and can take a type in whatever format, and return BTF type.

So from the user perspective it will works strictly by doing `(char*[64])` but `(*char[64])` will return an error.

kkourt: Sorry meant:
> Can we have clear semantics for the casting here? "

> The logic of the string parsing is splitted in 2 parts :

OK, but above is referring to the implementation.

I would like to understand what the language of the resolve is (specifically with respect to casting) without having to go over the implementation. What is the syntax for specifying a cast?

tdaudi: It's a custom type casting format inspired from the C type casting mechanism. I added exemples in `TestParseBTFTypePointerArrayCombinations` (https://github.com/cilium/tetragon/pull/5177/changes/e2767fbf9a790dd251a2443d66796bdc2460858d).

But if it's just a documentation problem I can rewrite this comment, will do that

kkourt: Examples help, but we need to specify the exact syntax. Otherwise, we can end up with ambiguities that are not covered by the examples.

docs/content/en/docs/concepts/tracing-policy/hooks.md
tdaudi: (this is still a WIP in the latest changes)

pkg/btf/type_builder.go
```diff
@@ -0,0 +1,256 @@
...
+		"long long", "long long int", "signed long long", "signed long long int",
+		"int64", "int64_t", "s64", "__s64":
+		return btfSignedInt("signed long long int", 8), true
+	case "unsigned long", "unsigned long int",
+		"unsigned long long", "unsigned long long int",
+		"uint64", "uint64_t", "u64", "__u64", "size_t":
+		return btfUnsignedInt("unsigned long long int", 8), true
+
+	default:
+		return nil, false
```
3u13r: Do we need the boolean here? Checking if the first return value is nil should have the same effect, right?

pkg/btf/type_builder.go
```diff
@@ -0,0 +1,256 @@
...
+		return btfUnsignedInt("unsigned short int", 2), true
+
+	// 4 Bytes
+	case "int", "signed int", "signed", "int32", "int32_t", "s32", "__s32":
+		return btfSignedInt("signed int", 4), true
+	case "unsigned int", "unsigned", "uint32", "uint32_t", "u32", "__u32":
+		return btfUnsignedInt("unsigned int", 4), true
+
+	// 8 Bytes
+	case "long", "long int", "signed long", "signed long int",
```
3u13r: In my opinion we should remove some of the types here and just list the ones we support, for example just the `stdint.h` names like "uint32_t" or the list of integer types the [user has available for args](https://tetragon.io/docs/reference/tracing-policy/#tracingpolicyspecfentriesindexargsindex).

pkg/sensors/tracing/generic_test.go
```diff
@@ -18,84 +18,111 @@ import (
...
-			name:    "unclosed bracket",
-			input:   "my.field[123",
-			want:    []string{},
-			wantErr: true,
+			input: "((struct task_struct*)((struct my_struct**)((char*)field)[8])[2]).comm",
+			want:  []string{"field", "(char*)", "[8]", "(struct my_struct**)", "[2]", "(struct task_struct*)", "comm"},
 		},
+
+		// cast mid-path
+		{input: "path.((char*)target)[8].end", want: []string{"path", "target", "(char*)", "[8]", "end"}},
```
3u13r: Not sure if we want synthetic test cases, if not I'd argue to remove this case, because accessing `.end` after having casted the previous element to a char array doesn't make sense.

pkg/btf/btf.go
```diff
@@ -388,30 +413,61 @@ func processMembers(
...
+		r.btfArgs[i].IsPointer = uint16(1)
 		targetType = t.Target
 	case *btf.Int, *btf.Enum:
-		btfArgs[i].IsPointer = uint16(1)
+		r.btfArgs[i].IsPointer = uint16(1)
 	}
 	return &targetType, nil
 }
+
+func (r *btfResolver) parseBTFTypeCast(i int) (btf.Type, error) {
```
3u13r: Personally, I'd like comments even on those internal functions with the intend to make it easier for a human reader to know what the purpose is and where it roughly sits in the callstack.
I.e. here I'd write something like:

```go
// parseBTFTypeCast parses the path element at i as a type cast. On success,
// it removes the cast from the path and returns its BTF type. If the element
// is not a cast, it returns errNotACast.
//
// For example, given {"field", "(char*)", "[1]"} and i=1, it returns a
// char* BTF type and changes the path to {"field", "[1]"}.
```

PR #5179, opened 2026-06-23
fix: prevent nil panic on closed stdin and sort event-types error in getevents

immanuwell:
### Description

Two small bugs in `tetra getevents`:

1. `os.Stdin.Stat()` error is silently dropped, so when stdin fd is closed `fi` is nil and `fi.Mode()` panics with a nil pointer dereference. Fix: check `err == nil` before touching `fi`.

2. Supported event types in the `--event-types` validation error were pulled from a map, so the order was random on every run. Fix: collect names, sort them, join - same pattern as `explain.go` already uses.

**Reproduce the panic:**
```
# close fd 0 before running tetra
bash -c 'exec 0<&-; tetra getevents'
# runtime error: invalid memory address or nil pointer dereference
```

**Reproduce the non-determinism:**
```
# run a few times with a bad event type and watch the order change
tetra getevents --event-types INVALID
```

### Changelog

```release-note
fix: `tetra getevents` no longer panics when stdin fd is closed; supported event types in error messages are now sorted alphabetically
```

### DIFF
--- a/cmd/tetra/getevents/getevents.go
+++ b/cmd/tetra/getevents/getevents.go
@@ -199,8 +199,8 @@ redirection of events to the stdin. Examples:
 			return nil
 		},
 		RunE: func(_ *cobra.Command, _ []string) error {
-			fi, _ := os.Stdin.Stat()
-			if fi.Mode()&os.ModeNamedPipe != 0 {
+			fi, err := os.Stdin.Stat()
+			if err == nil && fi.Mode()&os.ModeNamedPipe != 0 {
 				// read events from stdin
 				return getEvents(context.Background(), newIOReaderClient(os.Stdin, common.Debug))
 			}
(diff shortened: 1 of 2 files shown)


FedeDP:
Thanks for this patch!
Since it addresses 2 different things, can you split into 2 commits?

Also, can you fix the commit message to adhere to rule:
> Please avoid long commit subjects (max: 75, found: 78)

Thank you so much!

olsajiri:
> Thanks for this patch! Since it addresses 2 different things, can you split into 2 commits?
> 
> Also, can you fix the commit message to adhere to rule:
> 
> > Please avoid long commit subjects (max: 75, found: 78)
> 
> Thank you so much!

+1 also please add some description to commits changelogs

mtardy (changes requested):
agree with both reviews up there, marking this as request change for now

immanuwell:
done, addressed all the requested changes, PTAL

mtardy (approved):
lgtm thanks :)

FedeDP (approved):
Thanks!

PR #5182, opened 2026-06-25
Link uprobe target by file descriptor

andrewstrohman:
In a future PR, I will be doing uprobe target digest verification, where we support a new tracing policy configuration that allows users to supply digests, to be confirmed by tetragon before it links the target binary specified by `path`. This is useful for folks who only want to load policies for specific versions of a binary. See [this PR](https://github.com/cilium/tetragon/pull/4994) for the complete picture of where this is heading. Reviewers in that superset PR suggested that the change be broken up into multiple PRs. So, this is the first of a series of PRs.

The linux kernel tracks uprobe binary targets by inode. When you open a file, the kernel allocates a file descriptor.  That file descriptor is locked onto an inode. By linking the target via file descriptor, we ensure that the inode being linked is the same inode that was inspected for digest verification. This avoids a TOCTOU problem that would happen otherwise, where we could link a target binary that we did not check the digest of.

### DIFF
--- a/pkg/sensors/collection.go
+++ b/pkg/sensors/collection.go
@@ -103,6 +103,8 @@ type collection struct {
 
 	warnedOnModeRetrievalFailure  atomic.Bool
 	warnedOnStatsRetrievalFailure atomic.Bool
+	// if non-empty, this indicates that the collection cannot be disabled, and the string is the reason why
+	disableNotAllowed string
 }
 
 type selectorStatsMetadata struct {
--- a/pkg/sensors/handler.go
+++ b/pkg/sensors/handler.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"maps"
 	"slices"
+	"strings"
 	"sync"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
@@ -117,6 +118,15 @@ func (h *handler) addTracingPolicy(op *tracingPolicyAdd) error {
 	col.sensors = append(col.sensors, sensors...)
 	col.state = LoadingState
 
+	notAllowedReasons := make([]string, 0, len(col.sensors))
+	for _, s := range col.sensors {
+		notAllowedReason := s.DisableNotAllowed()
+		if notAllowedReason != "" {
+			notAllowedReasons = append(notAllowedReasons, notAllowedReason)
+		}
+	}
+	col.disableNotAllowed = strings.Join(notAllowedReasons, "; ")
+
 	// unlock so that policyLister can access the collections (read-only) while we are loading.
 	h.collections.mu.Unlock()
 	err = h.load(&col)
@@ -160,6 +170,10 @@ func (h *handler) deleteTracingPolicy(op *tracingPolicyDelete) error {
 func (h *handler) doDisableTracingPolicy(col *collection) error {
 	if col.state != EnabledState {
 		return fmt.Errorf("tracing policy %s is not enabled", col.name)
+	}
+
+	if col.disableNotAllowed != "" {
+		return fmt.Errorf("tracing policy %s cannot be disabled: %s", col.name, col.disableNotAllowed)
 	}
 
 	col.state = UnloadingState
--- a/pkg/sensors/program/loader_linux.go
+++ b/pkg/sensors/program/loader_linux.go
@@ -358,6 +358,38 @@ func MultiUprobeAttach(load *Program, bpfDir string) AttachFunc {
 	}
 }
 
+func attachMultiUpobeLink(load *Program, prog *ebpf.Program, path string,
+	attach *MultiUprobeAttachSymbolsCookies, bpfDir string,
+	idx int, extra ...string) (link.Link, error) {
+
+	exec, err := link.OpenExecutable(path)
+	if err != nil {
+		return nil, err
+	}
+	opts := &link.UprobeMultiOptions{
+		Addresses:     attach.Addresses,
+		Offsets:       attach.Offsets,
+		RefCtrOffsets: attach.RefCtrOffsets,
+		Cookies:       attach.Cookies,
+	}
+	var lnk link.Link
+	if load.RetProbe {
+		lnk, err = exec.UretprobeMulti(attach.Symbols, prog, opts)
+	} else {
+		lnk, err = exec.UprobeMulti(attach.Symbols, prog, opts)
+	}
+	if err != nil {
+		return nil, err
+	}
+	pinExtra := append(extra, strconv.Itoa(idx))
+	err = LinkPin(lnk, bpfDir, load, pinExtra...)
+	if err != nil {
+		lnk.Close()
+		return nil, err
+	}
+	return lnk, nil
+}
+
 func uprobeAttachMulti(load *Program, prog *ebpf.Program, spec *ebpf.ProgramSpec,
 	bpfDir string, extra ...string) (unloader.Unloader, error) {
 
@@ -368,32 +400,11 @@ func uprobeAttachMulti(load *Program, prog *ebpf.Program, spec *ebpf.ProgramSpec
 
 	linkFn := func() ([]link.Link, error) {
 		var links []link.Link
-		var lnk link.Link
 
 		idx := 0
 		for path, attach := range data.Attach {
-			exec, err := link.OpenExecutable(path)
+			lnk, err := attachMultiUpobeLink(load, prog, path, attach, bpfDir, idx, extra...)
 			if err != nil {
-				return nil, err
-			}
-			opts := &link.UprobeMultiOptions{
-				Addresses:     attach.Addresses,
-				Offsets:       attach.Offsets,
-				RefCtrOffsets: attach.RefCtrOffsets,
-				Cookies:       attach.Cookies,
-			}
-			if load.RetProbe {
-				lnk, err = exec.UretprobeMulti(attach.Symbols, prog, opts)
-			} else {
-				lnk, err = exec.UprobeMulti(attach.Symbols, prog, opts)
-			}
-			if err != nil {
-				return nil, err
-			}
-			pinExtra := append(extra, strconv.Itoa(idx))
-			err = LinkPin(lnk, bpfDir, load, pinExtra...)
-			if err != nil {
-				lnk.Close()
 				return nil, err
 			}
 			links = append(links, lnk)
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -690,6 +690,12 @@ func createGenericUprobeSensor(
 		in.selectorStatsBase = selectorStatsBase
 		selectorStatsBase += uint32(len(uprobe.Selectors))
 
+		absPath, err := filepath.Abs(uprobe.Path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to resolve absolute path for %q: %w", uprobe.Path, err)
+		}
+		uprobe.Path = absPath
+
 		var entryFile *os.File
 
 		ids, err = addUprobe(&uprobe, entryFile, ids, &in, &has)
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -12,7 +12,9 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"os"
 	"path"
+	"path/filepath"
 	"strconv"
 	"strings"
 
@@ -49,6 +51,8 @@ type observerUprobeSensor struct {
 	name string
 }
 
+const disableNotAllowedReasonLinkByFD = "target linked by file descriptor"
+
 var (
 	uprobeTable idtable.Table
 )
@@ -62,7 +66,8 @@ type uprobeLoadArgs struct {
 type genericUprobe struct {
 	loadArgs     uprobeLoadArgs
 	tableId      idtable.EntryID
-	path         string
+	targetPath   string
+	linkPath     string
 	symbol       string
 	address      uint64
 	refCtrOffset uint64
@@ -101,7 +106,8 @@ func (g *genericUprobe) SetID(id idtable.EntryID) {
 func (g *genericUprobe) LogAttrs(level slog.Level, msg string, attrs ...slog.Attr) {
 	attrs = append(attrs,
 		slog.Attr{Key: "policy_name", Value: slog.StringValue(g.policyName)},
-		slog.Attr{Key: "path", Value: slog.StringValue(g.path)},
+		slog.Attr{Key: "path", Value: slog.StringValue(g.targetPath)},
+
 		slog.Attr{Key: "symbol", Value: slog.StringValue(g.symbol)},
 		slog.Attr{Key: "address", Value: slog.Uint64Value(g.address)},
 	)
@@ -144,7 +150,7 @@ func handleGenericUprobe(r *bytes.Reader) ([]observer.Event, error) {
 
 	unix := &tracing.MsgGenericUprobeUnix{}
 	unix.Msg = &m
-	unix.Path = uprobeEntry.path
+	unix.Path = uprobeEntry.targetPath
 	unix.Symbol = uprobeEntry.symbol
 	unix.Offset = uprobeEntry.address
 	unix.RefCtrOffset = uprobeEntry.refCtrOffset
@@ -250,7 +256,7 @@ func loadSingleUprobeSensor(uprobeEntry *genericUprobe, args sensors.LoadProbeAr
 
 	symbol, offset := resolveSymbol(uprobeEntry.symbol)
 	attachData := &program.UprobeAttachData{
-		Path:         uprobeEntry.path,
+		Path:         uprobeEntry.linkPath,
 		Symbol:       symbol,
 		Offset:       offset,
 		Address:      uprobeEntry.address,
@@ -262,7 +268,7 @@ func loadSingleUprobeSensor(uprobeEntry *genericUprobe, args sensors.LoadProbeAr
 		return err
 	}
 
-	logger.GetLogger().Info(fmt.Sprintf("Loaded generic uprobe program: %s -> %s [%s]", args.Load.Name, uprobeEntry.path, uprobeEntry.symbol))
+	logger.GetLogger().Info(fmt.Sprintf("Loaded generic uprobe program: %s -> %s [%s]", args.Load.Name, uprobeEntry.targetPath, uprobeEntry.symbol))
 	return nil
 }
 
@@ -365,7 +371,7 @@ func loadMultiUprobeSensor(ids []idtable.EntryID, args sensors.LoadProbeArgs) er
 
 		load.MapLoad = append(load.MapLoad, mapLoad...)
 
-		attach, ok := data.Attach[uprobeEntry.path]
+		attach, ok := data.Attach[uprobeEntry.linkPath]
 		if !ok {
 			attach = &program.MultiUprobeAttachSymbolsCookies{}
 		}
@@ -384,7 +390,7 @@ func loadMultiUprobeSensor(ids []idtable.EntryID, args sensors.LoadProbeArgs) er
 
 		attach.Cookies = append(attach.Cookies, uint64(index))
 
-		data.Attach[uprobeEntry.path] = attach
+		data.Attach[uprobeEntry.linkPath] = attach
 	}
 
 	load.SetAttachData(data)
@@ -418,10 +424,11 @@ type addUprobeIn struct {
 }
 
 type uprobeHas struct {
-	sleepableOffload     bool
-	sleepablePreload     bool
-	substring            bool
-	sleepablePreloadSize int
+	sleepableOffload        bool
+	sleepablePreload        bool
+	substring               bool
+	sleepablePreloadSize    int
+	disableNotAllowedReason string
 }
 
 func validateMultiUprobeConsistency(uprobes []v1alpha1.UProbeSpec) error {
@@ -487,6 +494,8 @@ type uprobeConfigState struct {
 	argReturnPrinters []argPrinter
 
 	policyName string
+
+	entryFile *os.File
 }
 
 type uprobeArgConfig struct {
@@ -594,7 +603,7 @@ func initUprobeSelectors(spec *v1alpha1.UProbeSpec, in *addUprobeIn, state *upro
 	return nil
 }
 
-func cleanupUprobeEntries(ids []idtable.EntryID) error {
+func cleanupUprobeEntries(ids []idtable.EntryID, openedFiles []*os.File) error {
 	var errs error
 
 	for _, id := range ids {
@@ -614,6 +623,12 @@ func cleanupUprobeEntries(ids []idtable.EntryID) error {
 		_, err = uprobeTable.RemoveEntry(id)
 		if err != nil {
 			errs = errors.Join(errs, err)
+		}
+	}
+
+	for _, entryFile := range openedFiles {
+		if err := entryFile.Close(); err != nil && !errors.Is(err, os.ErrClosed) {
+			errs = errors.Join(errs, fmt.Errorf("problem closing path %q: %w", entryFile.Name(), err))
 		}
 	}
 
@@ -656,9 +671,11 @@ func createGenericUprobeSensor(
 		}
 	}
 
+	var openedFiles []*os.File
+
 	defer func() {
 		if retErr != nil {
-			if cleanupErr := cleanupUprobeEntries(ids); cleanupErr != nil {
+			if cleanupErr := cleanupUprobeEntries(ids, openedFiles); cleanupErr != nil {
 				retErr = errors.Join(retErr, cleanupErr)
 			}
 		}
@@ -673,9 +690,20 @@ func createGenericUprobeSensor(
 		in.selectorStatsBase = selectorStatsBase
 		selectorStatsBase += uint32(len(uprobe.Selectors))
 
-		ids, err = addUprobe(&uprobe, ids, &in, &has)
+		var entryFile *os.File
+
+		ids, err = addUprobe(&uprobe, entryFile, ids, &in, &has)
 		if err != nil {
+			if entryFile != nil {
+				if closeErr := entryFile.Close(); closeErr != nil && !errors.Is(closeErr, os.ErrClosed) {
+					err = errors.Join(err, closeErr)
+				}
+			}
 			return nil, err
+		}
+
+		if entryFile != nil {
+			openedFiles = append(openedFiles, entryFile)
 		}
 	}
 
@@ -699,13 +727,23 @@ func createGenericUprobeSensor(
 	}
 
 	return &sensors.Sensor{
-		Name:      name,
-		Progs:     progs,
-		Maps:      maps,
-		Policy:    polInfo.name,
-		Namespace: polInfo.namespace,
+		Name:                    name,
+		Progs:                   progs,
+		Maps:                    maps,
+		Policy:                  polInfo.name,
+		Namespace:               polInfo.namespace,
+		DisableNotAllowedReason: has.disableNotAllowedReason,
 		DestroyHook: func() error {
-			return cleanupUprobeEntries(ids)
+			return cleanupUprobeEntries(ids, openedFiles)
+		},
+		PostLoadHook: func() error {
+			var errs error
+			for _, entryFile := range openedFiles {
+				if err = entryFile.Close(); err != nil && !errors.Is(err, os.ErrClosed) {
+					errs = errors.Join(errs, fmt.Errorf("problem closing path %q: %w", entryFile.Name(), err))
+				}
+			}
+			return errs
 		},
 	}, nil
 }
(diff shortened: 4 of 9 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -285,6 +285,7 @@ func createMultiKprobeSensor(polInfo *policyInfo, multiIDs []idtable.EntryID, ha
 			SetRetProbe(true).
 			SetLoaderData(multiRetIDs).
 			SetPolicy(polInfo.name)
+
```
FedeDP: Can you drop this empty line?

andrewstrohman: Ok, will do in the the next push.

andrewstrohman:
There is a problem with this approach. I did not take into consideration the disable/enable tracing policy functionality. If the user disables the tracing policy, and then re-enables it, then the collection is re-used instead of re-created. That collection will contain sensors that have *os.File which are closed after the load.

I had an offline discussion about this with some folks. I'm going to summarize that here, so that everyone can get up-to-date and participate in the discussion.

There were some ideas about how to handle this:
1) I go back to my original approach, where I pushed all the address resolution and digest calculations down close to where we link the target binary. See [here](https://github.com/cilium/tetragon/tree/pr/andrewstrohman/uprobe-progs-by-fd-backup) for a backup of that approach.
2) We deprecate the enable/disable functionality. Users could just remove/add the policy back as an alternative.
3) We disallow disabling uprobe policies.
4) We only disallow disabling uprobe policies that contain digest configurations.
5) We re-open and check the digest upon enable. If we do this, we also need to recalculate elf offsets for when users pass addrs or when we have pure go binaries.

For 5, since we will support multiple digests, we have have a problematic sequence like this:
i) User loads policy, with two digests, digest1 and digest2 configured for a hook. We read the binary, calculate elf file offsets, and load policy because digest1 is verified.
ii) User disables policy
iii) User updates binary to one with digest2, but the elf file offsets are different
iv) User re-enables policy and we attach to the wrong address

This is why I believe that I would need to move the elf offset calculations from where they currently reside. But moving that code was the reason why I abandoned my original approach. So, maybe we should reconsider 1) again.

I'm going to experiment with approach 4 in this PR to see how we feel about it. If a policy does not have digests configured, I will link by path instead of file descriptor.

If anyone else has an idea, please let me know!

kkourt:
Thanks for the great summary @andrewstrohman! Can you please add all of this in the commit message when you are done? I think it's valuable context and we should have it in the git log.

andrewstrohman:
> Thanks for the great summary @andrewstrohman! Can you please add all of this in the commit message when you are done? I think it's valuable context and we should have it in the git log.

Ok, done!

andrewstrohman:
I moved this to draft because @olsajiri cleaned up the base that this change belongs on top of in [this PR](https://github.com/cilium/tetragon/pull/5191#pullrequestreview-4604610705). I've rebased my changes on top of his, and will re-open this PR once he merges his.

pkg/sensors/program/loader_linux.go
```diff
@@ -358,6 +381,44 @@ func MultiUprobeAttach(load *Program, bpfDir string) AttachFunc {
 	}
 }
 
+func attachMultiUpobeLink(load *Program, prog *ebpf.Program, file *os.File, attach *MultiUprobeAttachSymbolsCookies, bpfDir string, idx int, extra ...string) (link.Link, error) {
```
olsajiri: nit, lets split this in 2 lines? we don't have strict rule, but I think this one's too long

pkg/sensors/program/loader_linux.go
```diff
@@ -307,6 +308,17 @@ func UprobeAttach(load *Program, bpfDir string) AttachFunc {
...
 
+func procSelfFDPath(f *os.File) string {
+	return filepath.Join(option.Config.ProcFS, "self", "fd", strconv.FormatUint(uint64(f.Fd()), 10))
+}
+
+func getTargetPath(file *os.File, linkByFD bool) string {
+	if linkByFD {
+		return procSelfFDPath(file)
+	}
+	return file.Name()
```
olsajiri: hum, I don't understand the purpose of `linkByFD` , also why we can't always use the file.Name()

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -673,10 +690,30 @@ func createGenericUprobeSensor(
...
+		entryFile, ok := openedFiles[absPath]
+		if !ok {
+			entryFile, err = os.Open(absPath)
+			if err != nil {
+				return nil, err
+			}
+			openedFiles[absPath] = entryFile
+		}
+
+		ids, dna, err = addUprobe(&uprobe, entryFile, ids, &in, &has)
```
olsajiri: we set flags in `has` within `addUprobe` , let's have this `disableNotAllowed` as one of the `has` flags?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -144,7 +148,7 @@ func handleGenericUprobe(r *bytes.Reader) ([]observer.Event, error) {
 
 	unix := &tracing.MsgGenericUprobeUnix{}
 	unix.Msg = &m
-	unix.Path = uprobeEntry.path
+	unix.Path = uprobeEntry.file.Name()
```
olsajiri: ok we close the file after we load it.. but I guess it's fine to still get the path from the object

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -673,10 +690,30 @@ func createGenericUprobeSensor(
...
 		selectorStatsBase += uint32(len(uprobe.Selectors))
 
-		ids, err = addUprobe(&uprobe, ids, &in, &has)
+		absPath, err := filepath.Abs(uprobe.Path)
 		if err != nil {
+			return nil, fmt.Errorf("failed to resolve absolute path for %q: %w", uprobe.Path, err)
+		}
+		uprobe.Path = absPath
+
+		entryFile, ok := openedFiles[absPath]
```
olsajiri: this seems like optimization that could come on top of this change, or do we need it in here? this change is already big

pkg/sensors/tracing/genericusdt.go
```diff
@@ -137,8 +174,24 @@ func createGenericUsdtSensor(
...
 		selectorStatsBase += uint32(len(usdt.Selectors))
 
-		ids, err = addUsdt(&usdt, &in, ids, &has)
+		absPath, err := filepath.Abs(usdt.Path)
 		if err != nil {
+			return nil, fmt.Errorf("failed to resolve absolute path for %q: %w", usdt.Path, err)
+		}
+		usdt.Path = absPath
+
+		entryFile, ok := openedFiles[absPath]
```
olsajiri: same comment as for uprobes, also maybe both sensors could share a function for this?

pkg/sensors/tracing/uprobe_test.go
```diff
@@ -1378,3 +1379,56 @@ spec:` + opts + `
 		assert.Equal(t, uint32(1024), getMaxEntries(m))
 	})
 }
+
+func TestDisableEnablePolicyUprobe(t *testing.T) {
```
olsajiri, line 1386: would be great to have some comment explaining why we test this
also would ease up the change to have it in separate commit, but I feel like I'm loosing this battle, so up to you ;-)

pkg/sensors/program/loader_linux.go
andrewstrohman: Ok, sounds good. The next push will have it split.

pkg/sensors/program/loader_linux.go
andrewstrohman: > why we can't always use the file.Name()

From this PR description:
> The linux kernel tracks uprobe binary targets by inode. When you open a file, the kernel allocates a file descriptor. That file descriptor is locked onto an inode. By linking the target via file descriptor, we ensure that the inode being linked is the same inode that was inspected for digest verification. This avoids a TOCTOU problem that would happen otherwise, where we could link a target binary that we did not check the digest of.

I did answer this same question in the previous PR [here](https://github.com/cilium/tetragon/pull/4994#discussion_r3456557208).

It's important that we prevent any scenario where we link the wrong binary and/or elf offset. The kernel tracks the target binary by inode -- not path. While the kernel API to link a target binary currently requires a path, it translates that path into an inode, and then locks onto the inode instead of path that it was configured to link against.

This means that we need to ensure that the kernel will link the same inode that we computed the digest of and we determined the elf offset of.

Consider this sequence that could happen if we always link by the target path (`file.Name()`):
1. We open the target path, calculate the digest and compare it against the configured digests.
2. We then determine the elf offsets of the attach point given this already opened file. It's required that we do this with the already opened file, because otherwise we could have a situation where we check the digest, it passes, but then the user replaces the target binary before re-open the file to determine the elf offsets.
3. The user either deletes or moves the target binary.
4. The user replaces the binary at the target path with a new binary.
5. We link via user specified path. This means that we are linking a different binary/inode than what was digest checked and what we computed the elf offsets of. This causes malfunction when the new binary is executed.

This cannot happen when we use `procSelfFDPath(file)`. With this approach, when we reach step 5, we link the binary that was digest checked an we determined the elf offsets of -- not the new binary with new inode. If the user deleted the file in step 3, then after tetragon closes the file, it's inode will be released, and the probe will not fire during the new binary's (or any binaries) execution. 

If the user moved the binary that was checked in step 1  to a new path in step 3, then the probe will only fire for invocations of the original binary at the new path -- it will not interject for the new binary stored at the target path.

Let's come to agreement on this before proceeding further. If what I wrote here is wrong, it completely invalidates this PR . I don't think that there is any benefit of holding the file open during linking, if we don't link the target by the (checked) open file's file descriptor.

>  I don't understand the purpose of linkByFD 

See [this comment](https://github.com/cilium/tetragon/pull/5182#issuecomment-4803518535). This approach of opening the file during sensor creation and then closing it after loading the first time does not work correctly when combined with the ability to disable a policy because when we re-enable the sensor the file is closed. The file needs to remain open in order to link by file descriptor.

linkByFD will be set in a future commit based on whether a digest configuration was found. This is explained in this corresponding commit message. The idea is that when there is no digest verification configuration, we don't need to worry about TOCTOU, and can just link by path. Since we can link by path in this scenario, it's OK to allow disable/enable.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Yes, that's much better. Thanks.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: > I guess it's fine to still get the path from the object

Yes, that's right. See [here](https://github.com/cilium/tetragon/pull/4994#discussion_r3456563666) for the answer to this the last time you asked it.

pkg/sensors/tracing/uprobe_test.go
andrewstrohman: > would be great to have some comment explaining why we test this

Done

> also would ease up the change to have it in separate commit

This test is/was already in its own, separate commit. So, I'm not sure what the suggestion is here. If you meant PR instead of commit, this commit is apart of this PR because this PR allows us to disallow uprobe disablement. So, this test is meant to confirm that pre-existing uprobe policies can still be disabled after that change.

> but I feel like I'm loosing this battle

I hope you don't feel like I'm resisting your suggestions. I value your suggestions and try to do my best to address each one of them as completely as possible. So, please don't give up -- it's important to me that I produce the best quality that I can.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I've removed this optimization and added it back in subsequent commit.

pkg/sensors/tracing/genericusdt.go
andrewstrohman: I put the this logic into a shared function to avoid duplication.

pkg/sensors/program/loader_linux.go
olsajiri: I think I got confused by having the choice of getting the path via file.Name and by fd, I think the concept of passing fd is fine, let's keep it

if we are changing the interface to have always the File pointer (in `UprobeAttachData` and `MultiUprobeAttachData`) , why don't we always get the path from fd? no need to pass linkByFd flag

pkg/sensors/tracing/uprobe_test.go
olsajiri: nah, that's just general sigh, no worries, it's great

ugh must have mixed up what I see in global github view and in terminal sry ;-)

pkg/sensors/program/loader_linux.go
andrewstrohman: > why don't we always get the path from fd? no need to pass linkByFd flag

Once we close the file after the first load, the file descriptor is released. So, in order to use a file descriptor, we would need to re-open the file, and pass this new file handle's file descriptor to the loader API.

I could omit the linkByFd flag by checking the file descriptor, and if it's -1, re-open the file. Does that sound better? My only hesitation with this approach is that if we create a bug in the future where a digest-checked file is closed before we reach the link logic, we will re-open the file because of this, and won't get any signal that there is a bug. We will silently allow the TOCTOU problem to happen.

olsajiri: > > why don't we always get the path from fd? no need to pass linkByFd flag
> 
> Once we close the file after the first load, the file descriptor is released. So, in order to use a file descriptor, we would need to re-open the file, and pass this new file handle's file descriptor to the loader API.

hum, I still don't see that.. we open the file in the config phase and it's closed in  sensor's PostLoadHook, so it's opened during all the loads, no?

also would be great to have something like  `link.OpenExecutable(file)` but it looks like bigger change

andrewstrohman: When I wrote "first load" I'm talking about the first time the policy is added. Subsequent loads with closed files happen when the user re-enables the policy after disabling it.

This change allows users to continue disable/enable uprobe policies which don't specify digest verification.

In a subsequent PR, where I introduce the digest check, linkByFd will be true for uprobes that specify digest validation. But policies that do not specify digest verification will continue to be linked by path, so that they can be disabled and re-enabled without issue.

olsajiri: I'd suggest to keep this PR contained and if the change makes sense for the digest feature, then let's introduce it with the rest of the digest changes in its PR

andrewstrohman: > I'd suggest to keep this PR contained

This PR is about making it so that we can link by file descriptor. But we cannot link pre-existing uprobe policies by file descriptor because that conflicts with the ability to disable/enable a policy. So, from my perspective, linkByFD is apart of this, and therefore should be contained within this PR.

>  and if the change makes sense for the digest feature

If linkByFD does not make sense for the digest feature, then we shouldn't merge this PR. The only reason we set up this infrastructure is to use it for digest verification. So, if you disagree about this approach, I'd like to settle this here and now instead of deferring that decision to later PR. I don't feel comfortable merging this unless I know it will be useful for my goal of enabling digest verification. I listed the other ways to handle the conflict between linking by FD and the disable/enable policy feature in [this comment](https://github.com/cilium/tetragon/pull/5182#issuecomment-4803518535). Do you think a different approach is better?

pkg/sensors/tracing/genericusdt.go
```diff
@@ -137,10 +167,10 @@ func createGenericUsdtSensor(
 		in.selectorStatsBase = selectorStatsBase
 		selectorStatsBase += uint32(len(usdt.Selectors))
 
-		ids, err = addUsdt(&usdt, &in, ids, &has)
-		if err != nil {
+		if ids, err = addUsdt(&usdt, &in, ids, &has); err != nil {
 			return nil, err
 		}
+
```
kkourt, line 176: The code here seems equivalent before and after. If we are not modifying the semantics of the code, I would argue to avoid these stylistic changes because they introduce unnecessary churn.

pkg/sensors/tracing/genericusdt.go
```diff
@@ -300,32 +333,42 @@ func createUsdtSensorFromEntry(polInfo *policyInfo, usdtEntry *genericUsdt,
...
+			if cleanupErr := selectors.CleanupKernelSelectorState(state); cleanupErr != nil {
+				retErr = errors.Join(retErr, cleanupErr)
+			}
+		}
+	}()
+
 	se, err := elf.OpenSafeELFFile(spec.Path)
 	if err != nil {
-		return nil, err
+		return ids, err
```
kkourt, line 356: Why are we returning the partial list of IDs? I don't see it being used in the caller. Am I missing something?

pkg/sensors/collection.go
```diff
@@ -103,6 +103,7 @@ type collection struct {
 
 	warnedOnModeRetrievalFailure  atomic.Bool
 	warnedOnStatsRetrievalFailure atomic.Bool
+	disableNotAllowed             bool
```
kkourt: non-blocking suggestion: would it make sense to add a string (instead of a bool) so that we can explain user why disabling is not allowed?

pkg/sensors/program/loader_linux.go
```diff
@@ -307,6 +308,17 @@ func UprobeAttach(load *Program, bpfDir string) AttachFunc {
 	}
 }
 
+func procSelfFDPath(f *os.File) string {
+	return filepath.Join(option.Config.ProcFS, "self", "fd", strconv.FormatUint(uint64(f.Fd()), 10))
```
kkourt: Do we need the host procfs here or just `/proc`?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -945,6 +989,8 @@ func addUprobeEntries(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, state *u
 			pendingEvents:     nil,
 		}
 
+		has.disableNotAllowed = has.disableNotAllowed || uprobeEntry.loadArgs.linkByFD
```
kkourt: Can you add a comment here on why we do not allow disabling in this case?

kkourt (commented):
A couple of minor comments, but overall LGTM. Thanks!

pkg/sensors/program/loader_linux.go
olsajiri: I think the concept is ok, I just don't understand why we need to pass both file and LinkByFD to loader

iiuc we do not allow to disable policy if there's one sensor with `DisablePolicyNotAllowed` set in the collection

and we pass file pointer to `UprobeAttachData/MultiUprobeAttachData` structs

then `linkByFD` and `file` is used in `getTargetPath` to get the path for uprobe

why don't we keep Path in  UprobeAttachData/MultiUprobeAttachData and set it in config time to either:
- user specified path in spec, if we allow policy disable/enable
- procSelfFDPath, if we do NOT allow policy disable/enable

then the loader has always proper path without extra `file/linkByFD` in the `UprobeAttachData/MultiUprobeAttachData`

andrewstrohman: Ok, thanks for this explanation.

That will work, and sounds better. But of course we need to keep the file open until after we link. Otherwise, we won't link the correct file. Also, we need to close the file after we link, so that we don't keep the target file open forever, and accumulate a bunch of open files across all policies.

So, I'll change it to do it they way you describe here, and keep memory of the open files at the sensor level instead of the uprobe level, so that the post-load hook can close the files. (the `*os.File`'s will not be passed down to the linking logic)

pkg/sensors/tracing/genericusdt.go
andrewstrohman: Ok, make sense. I removed this stylistic change.

pkg/sensors/tracing/genericusdt.go
andrewstrohman: The caller has a defer'ed function that cleans up based on the returned ids. If we don't return the partial list (which is a list of entries that have been created thus far), then we won't clean up the entries that were created in the error path.

The same approach was used for uprobes [here](https://github.com/cilium/tetragon/pull/5191/changes/4b8dee283b03a32c6e38b2aab04a4a489de0be16).

pkg/sensors/collection.go
andrewstrohman: Yes, that's a good suggestion, and I did this.

pkg/sensors/program/loader_linux.go
andrewstrohman: I did this because of [this suggestion](https://github.com/cilium/tetragon/pull/4994#discussion_r3330874714). cc:@olsajiri

My understanding of this option is that it's a way for the user to specify where the host procfs is mounted, when it's not mounted at /proc.

My understanding is that tetragon always runs in the host namespace. Since the tetragon process opens these files, host procfs seems like the right place to find tetragon's pid.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Yeap, this is the comment that I added:
```
		// Depending on configuration (ie binaryDigests) we will link via file descriptor, instead of by the path configured in the policy,
		// in order to avoid a TOCTOU race. These files are closed after the sensor is loaded when the tracing policy is initially added,
		// and as such their file descriptors are no longer valid after that point. So, it's not safe to load the sensor again
		// when we re-enable a policy after it's been disabled, because the configured targetPaths would point to closed/re-used file descriptors.
```

pkg/sensors/tracing/genericusdt.go
```diff
@@ -171,6 +201,9 @@ func createGenericUsdtSensor(
 		Maps:      maps,
 		Policy:    polInfo.name,
 		Namespace: polInfo.namespace,
+		DestroyHook: func() error {
+			return cleanupUsdtEntries(ids)
```
FedeDP, line 212: Question: since in case of error during `createGenericUsdtSensor` we already called `cleanupUsdtEntries`, aren't we going to end up in this branch for every `id`:
```go
usdtEntry, err := genericUsdtTableGet(id)
if err != nil {
	errs = errors.Join(errs, err)
	continue
}
```
thus returning a list of errors in the DestroyHook?

pkg/sensors/handler.go
```diff
@@ -117,6 +117,14 @@ func (h *handler) addTracingPolicy(op *tracingPolicyAdd) error {
 	col.sensors = append(col.sensors, sensors...)
 	col.state = LoadingState
 
+	for _, s := range col.sensors {
+		notAllowedReason := s.DisableNotAllowed()
+		if notAllowedReason != "" {
+			col.disableNotAllowed = notAllowedReason
```
FedeDP: Do we want to collect all reasons here (in case any subsequent sensor has another reason)?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -945,6 +1003,14 @@ func addUprobeEntries(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, state *u
 			pendingEvents:     nil,
 		}
 
+		// Depending on configuration (ie binaryDigests) we will link via file descriptor, instead of by the path configured in the policy,
+		// in order to avoid a TOCTOU race. These files are closed after the sensor is loaded when the tracing policy is initially added,
+		// and as such their file descriptors are no longer valid after that point. So, it's not safe to load the sensor again
+		// when we re-enable a policy after it's been disabled, because the configured targetPaths would point to closed/re-used file descriptors.
+		if has.disableNotAllowedReason == "" && state.entryFile != nil {
+			has.disableNotAllowedReason = "binaryDigests configured"
```
FedeDP: Can we introduce a constant for this strng?

pkg/sensors/program/loader_linux.go
kkourt: > My understanding of this option is that it's a way for the user to specify where the host procfs is mounted, when it's not mounted at /proc.

Correct.

> My understanding is that tetragon always runs in the host namespace.

It depends on the namespace. In K8s, Tetragon does not run in the host pid namespace which is why we need to have a path to the host's /proc?

> Since the tetragon process opens these files, host procfs seems like the right place to find tetragon's pid.

Why do we need tetragon's pid?
Is there a reason why `/proc/self/fd/...` in the PID namespace that tetragon runs would not work?

andrewstrohman: Ok, thanks for the explanation. I was confused before.

It seems that we can use either procfs, and self will work.

I did:
```
sudo mkdir -p /mnt/host_proc
sudo unshare --mount --pid --fork bash
mount --make-rprivate /
mount --bind /proc /mnt/host_proc
mount -t proc proc /proc
```
To simulate tetragon's operating conditions when it's run within its own pid namespace, and also has access to the host procfs (host pid namespace).

both /proc/self/fd/... and /mnt/host_proc/fd/.... were accessible and showed an expected open file.

So I'll change it back to "/proc".

pkg/sensors/tracing/genericusdt.go
andrewstrohman: If an error occurs in `createGenericUsdtSensor`, we won't return a sensor from that function. We will return `nil, err` in the error case. The caller, `sensorsFromPolicyHandlers` sees the error and will destroy the sensors that it created in previous iterations, but will not attempt to destroy the `nil` sensor that was returned along with the err.

Since the sensor is never created in this scenario, it won't be destroyed.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Yeap, done.

pkg/sensors/handler.go
andrewstrohman: Ok, sounds good, done.

pkg/sensors/tracing/genericusdt.go
FedeDP: Ah here it is! Thank you so much for the explanation!

FedeDP (approved):
LGTM, let's wait for other reviewers :)

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -49,6 +52,21 @@ type observerUprobeSensor struct {
 	name string
 }
 
+const disableNotAllowedReasonBinaryDigests = "binaryDigests configured"
```
olsajiri: hum, this comes later in another PR, right?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -673,10 +704,34 @@ func createGenericUprobeSensor(
...
 
-		ids, err = addUprobe(&uprobe, ids, &in, &has)
+		absPath, err := filepath.Abs(uprobe.Path)
 		if err != nil {
+			return nil, fmt.Errorf("failed to resolve absolute path for %q: %w", uprobe.Path, err)
+		}
+		uprobe.Path = absPath
+
+		var entryFile *os.File
+		/*
```
olsajiri: leftover?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -945,6 +1025,14 @@ func addUprobeEntries(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, state *u
 			pendingEvents:     nil,
 		}
 
+		// Depending on configuration (ie binaryDigests) we will link via file descriptor, instead of by the path configured in the policy,
+		// in order to avoid a TOCTOU race. These files are closed after the sensor is loaded when the tracing policy is initially added,
+		// and as such their file descriptors are no longer valid after that point. So, it's not safe to load the sensor again
+		// when we re-enable a policy after it's been disabled, because the configured targetPaths would point to closed/re-used file descriptors.
+		if state.entryFile != nil {
+			has.disableNotAllowedReason = appendDisableNotAllowedReason(has.disableNotAllowedReason, disableNotAllowedReasonBinaryDigests)
```
olsajiri: let's not complicate this now.. we will have one reason, let's handle more reasons when we get there.. also I think this should be part of the digest PR

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I'm not sure exactly what you are thinking is a leftover.

`uprobe.Path = absPath` is not leftover. I added this per your previous suggestion to do so. I didn't think that suggestion was related to passing the *os.File to the link code, but maybe I misunderstood. I thought you just thought it would be more clear to show an absolute instead of relative path.

`var entryFile *os.File` is not leftover. This PR is about setting up the infrastructure to link by file descriptor.  Without the using a file handle, I'm not able to link by file descriptor.

`/*` I removed this comment.  The comment was intentional because I thought it helps people understand how to use this new infrastructure. But I don't feel strongly about this, so I removed it.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I've changed this to "target linked by file descriptor". In a later PR (when digest verification is done), it is changed back to "binaryDigests configured"

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I removed the appendDisableNotAllowedReason function and just set it to a string that says the target is linked by file descriptor.

olsajiri (approved):
thanks

kkourt (approved):

kkourt:
Merging, thanks Andy!

PR #5184, opened 2026-06-25
new: handle multiple sensors within same tracingpolicy spec

FedeDP:
### Changelog

```release-note
new: handle multiple sensors within same tracingpolicy spec
```


FedeDP:
Do we want to add an upgrade-notes for this?

pkg/sensors/tracing/policyhandler_test.go
```diff
@@ -0,0 +1,75 @@
...
+		case 0:
+			require.Equal(t, "generic_kprobe", sens.GetName())
+		case 1:
+			require.Equal(t, "generic_tracepoint", sens.GetName())
+		case 2:
+			require.Equal(t, "generic_uprobe", sens.GetName())
+		case 3:
+			require.Equal(t, "generic_usdt", sens.GetName())
+		}
+	}
```
kkourt, line 110: Should we test that we get events as well?

kkourt (commented):
Thanks! Overall, LGTM.

As an aside, this limitation was introduced due to some issues with shared maps in the past. I'm not aware of any other reason of why it wouldn't work. Let's see!

pkg/sensors/tracing/policyhandler_test.go
FedeDP: Oh yes we can test it too!

FedeDP: Done!

pkg/sensors/tracing/policyhandler_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"github.com/cilium/tetragon/pkg/policyfilter"
+	"github.com/cilium/tetragon/pkg/sensors"
+	"github.com/cilium/tetragon/pkg/testutils"
+	"github.com/cilium/tetragon/pkg/tracingpolicy"
+)
+
+func TestPolicyHandler(t *testing.T) {
+	if !config.EnableLargeProgs() || !bpf.HasUprobeRefCtrOffset() {
+		t.Skip("Need 5.3 or newer kernel for usdt and uprobe ref_ctr_off support for this test.")
+	}
```
FedeDP, line 36: Min requirements for usdt hook.
We could test the same on <5.3 with some `if` around, but didn't want to create useless complexity.

olsajiri:
> Thanks! Overall, LGTM.
> 
> As an aside, this limitation was introduced due to some issues with shared maps in the past. I'm not aware of any other reason of why it wouldn't work. Let's see!

I guess there was a reason ;-) but can't recall either and can't find it in quick git log search

still do we really need that? we might just create new bugs for no reason.. maybe let's allow that specifically for the use case?

FedeDP:
There is no hurry to merge this IMO; it just popped out in the context of https://github.com/cilium/tetragon/pull/5071 (specifically https://github.com/cilium/tetragon/pull/5071#discussion_r3441765023) thus i gave it a spin.

kkourt:
> > Thanks! Overall, LGTM.
> > As an aside, this limitation was introduced due to some issues with shared maps in the past. I'm not aware of any other reason of why it wouldn't work. Let's see!
> 
> I guess there was a reason ;-) but can't recall either and can't find it in quick git log search
> 
> still do we really need that? we might just create new bugs for no reason.. maybe let's allow that specifically for the use case?

I'm not aware of a specific use-case, so we can also wait until one appears :)

FedeDP:
Let me move this to draft!

PR #5189, opened 2026-06-27
policy: Add support for nodeSelector

### LINKED ISSUE #4935: Add nodeSelector support to TracingPolicy and TracingPolicyNamespaced

sayboras:
### Description

Fixes https://github.com/cilium/tetragon/issues/4935

### Changelog

```release-note
policy: Add support for nodeSelector
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
+++ b/docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
@@ -62,6 +62,65 @@ Kubernetes namespace, any non-null `hostSelector` will trigger an error.
 
 Based on these, the user can choose on which workloads does a policy apply.
 
+## Node filtering
+
+The selectors above (`podSelector`, `containerSelector`, `hostSelector`) filter
+which workloads a policy applies to; the policy is still loaded by the
+Tetragon agent on every node. To restrict policy loading to specific nodes, use
+the `nodeSelector` field.
+
+`nodeSelector` is a standard Kubernetes label selector matched against the
+labels of the node each Tetragon agent runs on. An agent loads the policy only
+when its own node's labels match. This is useful for policies that are
+expensive, noisy, or only relevant on a subset of nodes — for example GPU
+nodes, a specific architecture or OS, or a canary node pool.
+
+Unlike `hostSelector`, arbitrary `matchLabels` and `matchExpressions` are
+supported. If `nodeSelector` is empty (`{}`) or omitted, the policy is loaded on
+all nodes, preserving the default behavior. `nodeSelector` is supported on both
+`TracingPolicy` and `TracingPolicyNamespaced`.
+
+The following example loads the policy only on nodes labeled `nodepool: gpu`
+with an `amd64` architecture:
+
+```yaml
+spec:
+  nodeSelector:
+    matchLabels:
+      nodepool: gpu
+    matchExpressions:
+    - key: kubernetes.io/arch
+      operator: In
+      values:
+      - amd64
+  kprobes:
+```
+
+When a node is relabeled at runtime, each agent re-evaluates its policies and
+loads or unloads them accordingly.
+
+{{< note >}}
+`nodeSelector` is evaluated by the Tetragon agent against the Kubernetes node it
+runs on, so it only applies to policies managed through the Kubernetes API
+(`TracingPolicy` and `TracingPolicyNamespaced`). Policies loaded directly from a
+file or through the gRPC API are not subject to `nodeSelector`.
+{{< /note >}}
+
+{{< note >}}
+If the agent cannot evaluate the selector (for example, the node object is
+temporarily unavailable), it loads the policy on that node — `nodeSelector`
+fails open so a transient error never silently drops a policy.
+{{< /note >}}
+
+{{< warning >}}
+Because `nodeSelector` keys are read from the node's labels, any user that
+can edit node labels can change which policies an agent loads — including
+unloading an enforcement policy by relabeling the node. When gating enforcement
+policies (for example `Sigkill`), restrict who can modify the node labels used
+in the selector, and prefer labels that are set at node provisioning time and
+not writable by untrusted workloads.
+{{< /warning >}}
+
 ## Filtering semantics
 
 The following table summarizes how different combinations of `hostSelector`, `podSelector`, and `containerSelector` behave.
--- a/pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
+++ b/pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
@@ -121,6 +121,19 @@ type TracingPolicySpec struct {
 	HostSelector *slimv1.LabelSelector `json:"hostSelector,omitempty"`
 
 	// +kubebuilder:validation:Optional
+	// NodeSelector selects the nodes, by label, on which Tetragon agents load
+	// this policy. If empty or unset, the policy is loaded on all nodes. This differs in
+	// purpose from hostSelector: nodeSelector controls where a policy is loaded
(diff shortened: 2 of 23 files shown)


pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -120,6 +120,19 @@ type TracingPolicySpec struct {
 	// For now only ~ (none) and {} (all) is supported.
 	HostSelector *slimv1.LabelSelector `json:"hostSelector,omitempty"`
 
+	// +kubebuilder:validation:Optional
+	// NodeSelector selects, by node label, which nodes' Tetragon agents load
+	// this policy; if empty or unset it is loaded on all nodes. This differs in
```
mtardy: The wording could be simplified a bit
```suggestion
	// NodeSelector selects the nodes, by label, on which Tetragon agents load
	// this policy. If empty or unset, the policy is loaded on all nodes. This differs in
```

pkg/k8s/apis/cilium.io/v1alpha1/tracing_policy_types.go
```diff
@@ -120,6 +120,19 @@ type TracingPolicySpec struct {
...
+	// purpose from hostSelector: nodeSelector controls where a policy is loaded
+	// (on which nodes), whereas hostSelector controls which workloads a loaded
+	// policy applies to (host vs pod workloads) and does not affect whether the
+	// policy is loaded on a node. Use nodeSelector to target a node group such
+	// as GPU nodes, a specific architecture or OS, or a canary pool; use
+	// hostSelector to scope a loaded policy to host workloads. Unlike
+	// hostSelector, nodeSelector supports arbitrary matchLabels and
+	// matchExpressions.
```
mtardy, line 133: Thanks I think this is a really good idea to immediately clarify the difference between the two.

pkg/tracingpolicy/generictracingpolicy_test.go
```diff
@@ -94,6 +94,36 @@ func TestEmptyTracingPolicy(t *testing.T) {
...
+func TestYamlNodeSelector(t *testing.T) {
+	build.SkipIfK8sDisabled(t)
+	tp, err := FromYAML(tpNodeSelector)
+	require.NoError(t, err)
+	sel := tp.TpSpec().NodeSelector
+	require.NotNil(t, sel, "nodeSelector should be parsed")
+	require.Equal(t, "gpu", sel.MatchLabels["node-role"])
+	require.Len(t, sel.MatchExpressions, 1)
+	require.Equal(t, "kubernetes.io/arch", sel.MatchExpressions[0].Key)
+}
```
mtardy, line 125: great thanks

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
```diff
@@ -62,6 +62,63 @@ Kubernetes namespace, any non-null `hostSelector` will trigger an error.
...
+```
+
+When a node is relabeled at runtime, each agent re-evaluates its policies and
+loads or unloads them accordingly.
+
+{{< note >}}
+`nodeSelector` is evaluated by the Tetragon agent against the Kubernetes node it
+runs on, so it only applies to policies managed through the Kubernetes API
+(`TracingPolicy` and `TracingPolicyNamespaced`). Policies loaded directly from a
+file or through the gRPC API are not subject to `nodeSelector`.
```
mtardy, line 106: could we add runtime checks then to reject such policies loaded not via the k8s API?

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
```diff
@@ -62,6 +62,63 @@ Kubernetes namespace, any non-null `hostSelector` will trigger an error.
...
+
+If the agent cannot evaluate the selector (for example, the node object is
+temporarily unavailable), it loads the policy on that node — `nodeSelector`
+fails open so a transient error never silently drops a policy.
```
mtardy, line 112: nit: I would split this note into a separate one as it's not too much related to above apragraph

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
```diff
@@ -62,6 +62,63 @@ Kubernetes namespace, any non-null `hostSelector` will trigger an error.
...
+{{< /note >}}
+
+{{< warning >}}
+Because `nodeSelector` keys are read from the node's labels, any principal that
```
mtardy: What is "principal"? maybe you wanted to say individual or person or operator or user?

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
```diff
@@ -62,6 +62,63 @@ Kubernetes namespace, any non-null `hostSelector` will trigger an error.
...
+can edit node labels can change which policies an agent loads — including
+unloading an enforcement policy by relabeling the node. When gating enforcement
+policies (for example `Sigkill`), restrict who can modify the node labels used
+in the selector, and prefer labels that are set at node provisioning time and
+not writable by untrusted workloads.
+{{< /warning >}}
```
mtardy, line 122: really important warning here! thanks

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
```diff
@@ -62,6 +62,63 @@ Kubernetes namespace, any non-null `hostSelector` will trigger an error.
 
 Based on these, the user can choose on which workloads does a policy apply.
 
+## Node filtering
+
+The selectors above (`podSelector`, `containerSelector`, `hostSelector`) filter
+*which workloads* a policy applies to; the policy is still loaded by the
+Tetragon agent on every node. To instead control *which nodes* load a policy at
+all, use the `nodeSelector` field.
```
mtardy: maybe "To restrict policy loading to specific nodes, use the `nodeSelector` field."

mtardy: per the style guide we tend to avoid using bold for emphasize.

mtardy (approved):
Cool! Only a few nits and comments. Individual patches looks really good :) thanks for the e2e test as well!

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
sayboras: ah i mean k8s principal credentials, updated to user to avoid any confusion :100:

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
sayboras: Thanks, I will take this as follow-up, need to refresh my memory if there was a use case to support VM deployments with selector as hostname, IP, etc.

PR #5190, opened 2026-06-28
uprobe: Add resolvePathInContainer support

### LINKED ISSUE #5105: How can I hook a function from user-space executable in other pods.
Hello.

Now I'm trying to hook a function from user-space executable in other pods.
Target function in test is pam_authenticate function in libpam library.
This function will be called when user try to do "su -".

This is a TracingPolicy for it.
```
kind: TracingPolicy
metadata:
  name: "up-pam"
spec:
  options:
  - name: "disable-uprobe-multi"
    value: "1"
  uprobes:
  - path: "/usr/lib64/libpam.so.0.85.1"
    symbols:
    - "pam_authenticate"
```

It works well in bare-metal linux.
My problem is that I'm not sure what is the best way to hook a function from user-space executable 'in other pods.'

Initially, I struggled with an error: 'Failed to register TracingPolicy.'
And I realized that Tetragon in the Tetragon-pod can not see libpam library in user-space of the host-node in kubernetes
I tried to describe the situation with following figure.

Image

In the figure, purple box indicates Tetragon-pod, orange box indicates other-pod (where I want to hook something)
Gray color process is a process in the host-node area.

In a Kubernetes environment, while the kernel is shared, the user-space is not shared.
Like I mentioned, Tetragon in the Tetragon-pod can not see libpam library in user-space ('libpam.0' in the figure)
So, Tetragon couldn't register UPROBE TracingPolicy,
Resultingly, executing 'su -' couldn't fire hooking both host-node area process(gray) and other-pod area process(orange)

So, I mount hostPath containing libpam.0 to the Tetragon-pod like below:

Image

Then, 'Failed to register' error disappeared, finally, host-node area process(gray) fired event.
But, you know, other-pod(orange) is isolated in the kubernetes system.
I couldn't register and fire libpam in the other-pod ('libpam.2' in the orange box)

Actually my goal is hooking function in libpam.2 in the other-pod.

 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
I have some ideas or advice from the internet, but they are not good for me.

* Advice 1: Mount root-layover-directory of other-pod.
Advice said;
  You can find real file path of 'other-pod' from the overlay directory.
  ("/var/lib/containers/storage/overlay")
  Search it and mount this hostpath to the tetragon pod and make tracingpolicy 
  according to the path for executable you want to hook
  
But, you know it cannot be practical solution. right?
If I want to deploy 'other-pod' again, I have to do it again and again.
Search path from the overlay, make new tracingpolicy for it ...

* Advice 2: Mount /proc/<PID>/root path of 'other-pod'
Advice said;
  You can also see the 'other-pod' path by mount '/proc/<PID>/root'
  <PID> is a process id of main-process or 'other-pod'
  Search it and mount this hostpath to the tetragon pod and make tracingpolicy 
  according to the path for executable you want to hook

I think this advice has a same issue.
If I want to deploy 'other-pod' again, I have to do it again and again.
Search PID from the process list of host-node, make new tracingpolicy for it ...
It's almost impossible in the practical field.

* Idea 1: Put executables on the host-node side and mount them in the Tetragon-pod and other-pod.
  And then, Tetragon can register executables, and when these executables are called by process in other-pod,
  It might make fire events, and Tetragon could get it.

It's a good idea, but the thing is that I cannot handle or change 'other-pod' in practical field.
So, it's not good as well.
 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

May I have some advice for it?
What is the best way to do this?

Thanks.

sayboras:
### Description
Uprobe paths are interpreted in the agent's own mount namespace, so binaries that only exist inside other pods' images cannot be hooked.

Add a `resolvePathInContainer` option to uprobe specs. When set, `path` is resolved relative to the root filesystem of each container selected by the policy's `podSelector`, and the uprobe is attached per matching container. Container roots are resolved via runtime hooks, with CRI as fallback. The option requires a `podSelector` and comes with documentation, an example policy, and an e2e test backed by a new tester-progs image for test workloads.

Fixes: #5105

### Changelog

```release-note
uprobe: Add resolvePathInContainer option to resolve uprobe paths inside selected containers
```


sayboras:
Forgot updating the docs, marking it as draft.

sayboras:
Seems like I messed up the rebase, and CI is failing for new tests :thinking:

PR #5191, opened 2026-06-28
uprobe: factor addUprobe code

olsajiri:
it grew a lot, making it more structured

### DIFF
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -469,6 +469,48 @@ func validateMultiUprobeConsistency(uprobes []v1alpha1.UProbeSpec) error {
 	return nil
 }
 
+type uprobeConfigState struct {
+	symbols       int
+	offsets       int
+	addrs         int
+	refCtrOffsets int
+}
+
+func validateUprobeSpec(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) error {
+	state.symbols = len(spec.Symbols)
+	state.offsets = len(spec.Offsets)
+	state.addrs = len(spec.Addrs)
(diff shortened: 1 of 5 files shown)


pkg/sensors/tracing/genericuprobe.go
```diff
@@ -875,90 +893,102 @@ func addUprobeEntries(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, state *u
...
 			if err != nil {
-				return nil, err
+				return ids, err
 			}
 		}
 	} else if state.symbols != 0 {
 		for idx, sym := range spec.Symbols {
 			if err := checkSymbol(sym); err != nil {
-				return nil, fmt.Errorf("failed to parse symbol: %w", err)
+				return ids, fmt.Errorf("failed to parse symbol: %w", err)
```
andrewstrohman, line 985: returning ids inside loops like this will cause a leak when the failure happens after the first iteration.

andrewstrohman: so, I don't think we should introduce `added` and instead just append to `ids`.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -875,90 +893,102 @@ func addUprobeEntries(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, state *u
 
 		state.eventConfig.FuncId = uint32(id.ID)
 
-		ids = append(ids, id)
+		added = append(added, id)
```
andrewstrohman: See comment below. Using `added` here instead of `ids` will result in a leak if a failure occurs after the first iteration, when iterating over symbols/addrs/offsets.

andrewstrohman (approved):
Thanks, LGTM.

PR #5196, opened 2026-06-29
chore(deps): update all lvh-images main

sayboras:


bpf/process/bpf_generic_lsm_ima_bprm.c
```diff
@@ -24,10 +24,12 @@ BPF_PROG(ima_bprm, struct linux_binprm *bprm)
 	struct ima_hash *dummy = map_lookup_elem(&ima_hash_map, &pid_tgid);
 
 	if (dummy && dummy->state == 1) {
+		struct file *file = *__builtin_preserve_access_index(&bprm->file);
```
mtardy: tip: we have the `_` macro for that https://github.com/cilium/tetragon/blob/c091a957d6158087e91125f7853dd9fa74e1ff3a/bpf/lib/bpf_helpers.h#L56

mtardy: so maybe you can use it inline like previously

sayboras: ah nice :100:

bpf/process/bpf_process_event.h
```diff
@@ -263,7 +263,14 @@ get_namespaces(struct msg_ns *msg, struct task_struct *task)
...
 		} else {
-			msg->user_inum = 0;
+			// Since kernel 7.1 user_ns was dropped from mm_struct;
+			// read it from the task credentials instead.
+			const struct cred *cred;
+
+			probe_read(&cred, sizeof(cred), _(&task->cred));
+			probe_read(&user_ns, sizeof(user_ns), _(&cred->user_ns));
+			probe_read(&msg->user_inum, sizeof(msg->user_inum),
+				   _(&user_ns->ns.inum));
```
mtardy: if we use the core macros, you could write this with the following:
```suggestion
			msg->user_inum = BPF_CORE_READ(task, cred, user_ns, ns.inum);
```

mtardy (approved):
lgtm, just one remark, mostly nit

PR #5199, opened 2026-06-29
[v1.7] backport fd_ty read_arg fix

andrewstrohman:
Upstream PR: https://github.com/cilium/tetragon/pull/5132

Jiri fixed a bug that I created in the last release. I'd like to get this fix into v1.7 (the only affected release branch) because the bug has the potential to generate a lot of events.


olsajiri (approved):
thanks

PR #5212, opened 2026-06-30
usdt: unify selector state when there are multiple entries per hook

andrewstrohman:
While probably a rare use-case, we support users setting multiple USDT tracepoints with the same provider/name. When this happens, we create multiple usdt entries for the hook. These entries share the same selector, because they were generated from the same hook.

We have a similar situation for uprobes, where one hook has multiple entries. In that case, we re-use the selector state. This change makes usdt like uprobes in this way. It makes sense to represent a selector with a single selector state.

While investigating this, I found that an error message was emitted when I created a binary that has multiple tracepoints with the same provider/name. This change contains a fix for that as well.


FedeDP (approved):

olsajiri (approved):
thanks

PR #5214, opened 2026-07-01
add read-only/frozen config map

olsajiri:
Adding support to store read-only config data into single map that's shared among all sensors.

The idea is that base sensor creates read-only/frozen global map pinned in rodata_config,
and other sensors use just it instead of having their own .rodata maps.

```
struct rodata_config {
	__u8 ITER_NUM;
	__u8 PARENTS_MAP_ENABLED;
	__u8 ENV_VARS_ENABLED;
	__u8 pad[5];
};
```

### DIFF
--- a/bpf/lib/config.h
+++ b/bpf/lib/config.h
@@ -32,7 +32,8 @@ struct rodata_config {
 	__u8 ITER_NUM;
 	__u8 USE_PERF_RING_BUF;
 	__u8 ENV_VARS_ENABLED;
-	__u8 pad[5];
+	__u8 PARENTS_MAP_ENABLED;
+	__u8 pad[4];
 };
 
 volatile const struct rodata_config rodata_config
@@ -53,11 +54,12 @@ volatile const struct rodata_config rodata_config
  * without compile-time elimination both get verified, which failed to
  * load on a real v5.4 kernel in CI.
  *
- * ENV_VARS_ENABLED doesn't have that problem - its call site is a plain
- * guard with no alternate branch, so it works fine as an ordinary
- * runtime value on any kernel.
+ * PARENTS_MAP_ENABLED/ENV_VARS_ENABLED don't have that problem - their
+ * call sites are plain guards with no alternate branch, so they work
+ * fine as ordinary runtime values on any kernel.
  */
 #ifdef __LARGE_BPF_PROG
+volatile const __u8 PARENTS_MAP_ENABLED;
 volatile const __u8 ENV_VARS_ENABLED;
 #define ITER_NUM     0
 #define CONFIG(name) name
--- a/bpf/lib/config.h
+++ b/bpf/lib/config.h
@@ -3,34 +3,41 @@
 
 #pragma once
 
-#ifdef __V511_BPF_PROG
-
-#define DECLARE_CONFIG(type, name) \
-	volatile const type CONFIG_##name;
+#define __field(name) name
 
 /*
- * Reconstruct the rodata pointer on each access to prevent the compiler
- * from reusing the pointer across multiple accesses in short
- * succession. This makes branches based on config variables more likely
- * to be predictable using backtracking, since load/deref/branch will be
- * close to each other and the pointer won't be reused across basic
- * blocks.
+ * Read a field from a frozen, read-only rodata map, reconstructing the map
+ * pointer on each access to prevent the compiler from reusing it across
+ * multiple accesses in short succession. This makes branches based on
+ * config variables more likely to be predictable using backtracking, since
+ * load/deref/branch will be close to each other and the pointer won't be
+ * reused across basic blocks.
  *
- * Specifying the global var ptr as an asm input gives enough
+ * Specifying the global map pointer as an asm input gives enough
  * information to the compiler to allow it to reuse the pointer across
  * blocks, so opt for a direct symbol reference instead. We need the
  * pointer reconstructed in bytecode on every access.
  */
-#define CONFIG(name)                                                  \
-	(*({                                                          \
-		void *out;                                            \
-		asm volatile("%0 = " __stringify(CONFIG_##name) " ll" \
-			     : "=r"(out));                            \
-		(typeof(CONFIG_##name) *)out;                         \
-	}))
+#define __CONFIG_FIELD(map, type, field)                      \
+	({                                                    \
+		void *out;                                    \
+		asm volatile("%0 = " #map " ll"               \
+			     : "=r"(out));                    \
+		((volatile const type *)out)->__field(field); \
+	})
 
-DECLARE_CONFIG(bool, ITER_NUM);
-DECLARE_CONFIG(bool, USE_PERF_RING_BUF);
+#ifdef __V511_BPF_PROG
+
+struct rodata_config {
+	__u8 ITER_NUM;
+	__u8 USE_PERF_RING_BUF;
+	__u8 pad[6];
+};
+
+volatile const struct rodata_config rodata_config
+	__attribute__((section(".rodata.config"), used));
+
+#define CONFIG(name) __CONFIG_FIELD(rodata_config, struct rodata_config, name)
 
 #else
 
--- a/bpf/lib/config.h
+++ b/bpf/lib/config.h
@@ -31,7 +31,8 @@
 struct rodata_config {
 	__u8 ITER_NUM;
 	__u8 USE_PERF_RING_BUF;
-	__u8 pad[6];
+	__u8 ENV_VARS_ENABLED;
+	__u8 pad[5];
 };
 
 volatile const struct rodata_config rodata_config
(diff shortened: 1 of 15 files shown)


bpf/lib/config.h
```diff
@@ -15,21 +22,14 @@
...
-		(typeof(CONFIG_##name) *)out;                         \
-	}))
-
-DECLARE_CONFIG(bool, ITER_NUM);
+#define CONFIG(name)                                                         \
+	({                                                                   \
+		void *out;                                                   \
+		asm volatile("%0 = rodata_config ll"                         \
+			     : "=r"(out));                                   \
+		((volatile const struct rodata_config *)out)->__field(name); \
```
kkourt: > #define __field(name) name

Can we just use ->name here?

kkourt (commented):
Overall, LGTM, just a small comment.

As an aside, I think the loading code becomes increasingly complex so maybe we can rethink its structure.

olsajiri:
> Overall, LGTM, just a small comment.
> 
> As an aside, I think the loading code becomes increasingly complex so maybe we can rethink its structure.

it's complex, but I was thinking the same ;-) will try to come up with some reorg

bpf/lib/config.h
olsajiri: it's for the linter, otherwise it's complaining that name should be in `()` and we can't disable it just for this file

kkourt (approved):

mtardy:
I'll take a look this afternoon!

pkg/sensors/program/map.go
```diff
@@ -99,8 +99,13 @@ type MapOpts struct {
 }
 
 // Map represents BPF maps.
+// Name
+// - ELF name, map's ebpf program handle
+// PinName
+// - bpffs filename, it may differ from the ELF map name.
```
mtardy, line 108: maybe explain why here since you have a comment from what we have in the commit log?

bpf/lib/config.h
```diff
@@ -5,8 +5,15 @@
 
 #ifdef __V511_BPF_PROG
 
-#define DECLARE_CONFIG(type, name) \
-	volatile const type CONFIG_##name;
+struct rodata_config {
+	__u8 ITER_NUM;
+	__u8 pad[7];
+};
```
mtardy, line 37: so all global config must be declared in the same struct now instead of here and there with a macro?

bpf/lib/config.h
```diff
@@ -15,21 +22,14 @@
...
-	}))
-
-DECLARE_CONFIG(bool, ITER_NUM);
+#define CONFIG(name)                                                         \
+	({                                                                   \
+		void *out;                                                   \
+		asm volatile("%0 = rodata_config ll"                         \
+			     : "=r"(out));                                   \
+		((volatile const struct rodata_config *)out)->__field(name); \
+	})
```
mtardy: I'm not sure to get the __field(name) macro? why do we need it?

pkg/sensors/program/map.go
```diff
@@ -372,22 +399,26 @@ func LoadOrCreatePinnedMap(pinPath string, mapSpec *ebpf.MapSpec, create bool) (
 			if create {
 				logger.GetLogger().Warn("will delete and recreate", "map", mapSpec.Name)
 				os.Remove(pinPath)
-				return createPinnedMap(pinPath, mapSpec)
+				return createPinnedMap(pinPath, mapSpec, validate)
 			}
-			return nil, fmt.Errorf("incompatible map '%s'", pinPath)
+			return nil, fmt.Errorf("incompatible map '%s': %w", pinPath, err)
```
mtardy, line 407: I agree with Kornilios that it's a bit harsh to follow part of this code. Maybe in some place where you return in the error case, following the traditional:

```go
err := function()
if err != nil {
  blabla
  return something;
}

do something else
```

instead of checking for `err == nil` would make it easier (at least for me to understand)

mtardy (approved):
I think overall lgtm, appreciate how [rodata_linux.go](https://github.com/cilium/tetragon/pull/5214/changes/11c6099765b2a60fc3d53ff4bec9cce53534291e#diff-d4ee481932cacd4921d55cfddc754309debbd973578623a808f2bd1b76cc3dbc) centralize some of this stuff.

Agree with Kornilios that this code is a bit hard to browse now, maybe would be worth to have some kind of simplification or more docs on what maps/progs are going through when being loaded by the loader (what is check at what step), I usually have to re-learn if usually when I need to interact with this (if not asking you lol). Maybe some LLM can help us there now.

bpf/lib/config.h
olsajiri: it's for the linter, otherwise it's complaining that name should be in () and we can't disable it just for this file

bpf/lib/config.h
olsajiri: yes, the global config for verifier is now in the `rodata_config struct` which we load with base sensor and all the other sensors use that, you still use the `CONFIG(NEW_VARIABLE)` to use the config

pkg/sensors/program/map.go
olsajiri: could you point me to the place/instance you refer to,  I'm not sure I see a place where this could make current code better

pkg/sensors/program/map.go
olsajiri: ok, I added bit more comment to the source code

pkg/sensors/program/map.go
olsajiri: ok, let me try the oss sync first and merge and we can discuss this later

olsajiri:
@mtardy @kkourt I had to rework it to interface better with ee, please recheck

kkourt:
> @mtardy @kkourt I had to rework it to interface better with ee, please recheck

BTW, if you want to re-request review you notify the reviewers by clicking on the github button "Re-request review".

olsajiri:
> > @mtardy @kkourt I had to rework it to interface better with ee, please recheck
> 
> BTW, if you want to re-request review you notify the reviewers by clicking on the github button "Re-request review".
> 

yea, I know, but I felt like writing ;-)

pkg/sensors/program/rodata_linux.go
```diff
@@ -0,0 +1,96 @@
...
+
+// MapBuilderRodataConfig creates a shared, read-only, frozen configuration
+// map with the given ELF and pin names, populated by load func
+//
+// T must be a fixed-layout struct usable with encoding/binary.Write
+// (fixed-width integer fields and fixed-size arrays thereof only - no
+// strings, slices, maps, or pointers). The encoded byte length of T must
+// exactly match the compiled size of the corresponding ".rodata.<name>"
+// section in the target BPF object, including any C-side padding.
+func MapBuilderRodataConfig[T any](elfName, pinName string, load func() (T, error), lds ...*Program) *Map {
```
kkourt: Why do we need generics here?

Can't we just use `load func() (any, error)`?

pkg/sensors/program/rodata_linux.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/features"
+)
+
+var rodataConfigMaps = map[string]string{} // elfName -> pinName
+
+func registerRodataConfigMap(elfName, pinName string) {
+	if existing, ok := rodataConfigMaps[elfName]; ok {
+		if existing != pinName {
+			panic(fmt.Sprintf("required shared map %q already registered with pin %q, got %q", elfName, existing, pinName))
```
kkourt, line 28: Since this can panic, we should only be using it at initialization time (which seems to be the case with the current code): `registerRodataConfigMap` is called by `MapBuilderRodataConfig` which is called by `MapBuilderRodataConfig` which is called by:

```go
// RodataConfigMap is Tetragon's own shared, read-only rodata configuration map.
var RodataConfigMap = MapBuilderRodataConfig(Execve)
```

Should we add a comment this effect or maybe change the function names to reflect that?
(e.g., I like to use an `AtInit`  suffix and a comment in the cases)

pkg/sensors/program/rodata_linux.go
```diff
@@ -0,0 +1,96 @@
...
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/features"
+)
+
+var rodataConfigMaps = map[string]string{} // elfName -> pinName
```
kkourt, line 18: I guess we don't need a lock here because all loading functions are serialized in the sensor manager?

kkourt (approved):
Thanks!

Just a couple of minor comments, but LGTM!

pkg/sensors/program/rodata_linux.go
olsajiri: I think that would work as well, is that a preferred way?

pkg/sensors/program/rodata_linux.go
olsajiri: ok, sounds good

pkg/sensors/program/rodata_linux.go
olsajiri: yes, no locking is needed because sensor manager is doing that for us.. I'll add comment about that

pkg/sensors/program/rodata_linux.go
kkourt: Only reason being that it's simpler.

bpf/lib/config.h
mtardy: would have been nice to add a comment in there or in the commit log but not a hard req

mtardy (approved):
thanks, lgtm

kkourt (approved):

PR #5218, opened 2026-07-02
Do not require enable-policy-filter on hostSelector

tpapagian:
Now if we run Tetragon on a non-k8s env with a policy such as:
```yaml
    kind: TracingPolicy
    metadata:
      name: "sys-linkat-passwd"
    spec:
      hostSelector: {}
      kprobes:
      - call: "sys_linkat"
        syscall: true
```
It will fail with an error message: `level=error msg="Failed to execute tetragon" error="policyfilter is disabled"`

In that case, we can ignore `spec.hostSelector` as all workloads are considered host workloads.

On the other hand, if a policy defines `podSelector` and/or `containerSelector` in a non-k8s env Tetragon will return an error.

This is aligned with what we do in non-k8s builds (i.e. `tetragon-nok8s`).

### DIFF
--- a/pkg/cgtracker/test/cgtracker_test.go
+++ b/pkg/cgtracker/test/cgtracker_test.go
@@ -215,6 +215,11 @@ func doProgTest(t *testing.T, cgfsPath string) {
 // TestCgTrackerPolicyFilter checks that cgroup tracking works with policyfilter
 func TestCgTrackerPolicyFilter(t *testing.T) {
 	build.SkipIfK8sDisabled(t)
+	oldEnableK8s := option.Config.EnableK8s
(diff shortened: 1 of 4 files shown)


mtardy (approved):
thanks lgtm!!

FedeDP (approved):
LGTM, thank you!

kkourt (approved):

kkourt:
@tpapagian do we need to backport this?

tpapagian:
> @tpapagian do we need to backport this?

Not a bugfix but it will definitely simplify policy deployment in some cases. As it is easy to backport I have done this in https://github.com/cilium/tetragon/pull/5238

PR #5219, opened 2026-07-02
fix: reject malformed field filter JSON

immanuwell:
Fixes

### Description
Malformed `fieldFilters` JSON was getting swallowed in the snake_case fixup path, so Tetragon ended up acting like no field filters were set. Kinda rough, bad config should fail loud not fail open.

This makes `ParseFieldFilterList` return the JSON decode error instead, and adds a small regression test.

Repro:
`go test ./pkg/fieldfilters/...`

or with a tiny snippet:
```go
_, err := fieldfilters.ParseFieldFilterList("{\"fields\":\"process.pid\",\"action\":\"INCLUDE\"")
fmt.Println(err)
```

Before: `nil`
After: `unexpected EOF`

### Changelog
```release-note
Fix malformed field filter JSON being silently ignored.
```


kevsecurity (approved):

FedeDP (approved):
LGTM, thank you!

will-isovalent (approved):
Thank you!

PR #5220, opened 2026-07-03
cmd/tetra: return bugtool command errors

immanuwell:
Fixes

### Description
`tetra bugtool` was swallowing `bugtool.Bugtool(...)` errors and still exiting 0.
That is kinda rough for scripts and CI.

This switches the command to `RunE` and returns the error.
Added a small test for the fail path and the ok path.

Repro:
1. Run `go run ./cmd/tetra bugtool -o /proc/1/tetragon-bugtool.tar.gz`
2. Before this patch it exits 0, even though no archive is created.
3. With this patch it exits 1 and prints `failed to create bugtool archive: ...`

### Changelog
```release-note
Fix `tetra bugtool` to return a non-zero exit code when archive creation fails
```

### DIFF
--- a/cmd/tetra/bugtool/bugtool.go
+++ b/cmd/tetra/bugtool/bugtool.go
@@ -43,8 +43,8 @@ func New() *Command {
 	bugtoolCmd.command = &cobra.Command{
 		Use:   "bugtool",
 		Short: "Produce a tar archive with debug information",
-		Run: func(_ *cobra.Command, _ []string) {
-			bugtool.Bugtool(outFile, bpfTool, gops, common.MaxRecvMsgSize, bugtoolCmd.CommandActions, bugtoolCmd.GRPCActions)
+		RunE: func(_ *cobra.Command, _ []string) error {
+			return bugtool.Bugtool(outFile, bpfTool, gops, common.MaxRecvMsgSize, bugtoolCmd.CommandActions, bugtoolCmd.GRPCActions)
 		},
 	}
 


cmd/tetra/bugtool/bugtool.go
```diff
@@ -4,29 +4,32 @@
 package bugtool
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 
 	"github.com/cilium/tetragon/cmd/tetra/common"
-	"github.com/cilium/tetragon/pkg/bugtool"
+	bugtoolpkg "github.com/cilium/tetragon/pkg/bugtool"
```
tpapagian: Do we really need this rename of the import here?

cmd/tetra/bugtool/bugtool.go
```diff
@@ -4,29 +4,32 @@
...
 )
 
 var (
-	outFile string
-	bpfTool string
-	gops    string
+	outFile    string
+	bpfTool    string
+	gops       string
+	runBugtool = bugtoolpkg.Bugtool
```
tpapagian: Possibly rename that to a more descriptive name i.e. `bugtoolFn`?

cmd/tetra/bugtool/bugtool_test.go
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: Apache-2.0
```
tpapagian: You can possibly add this test on a separate commit.

tpapagian (commented):
Thanks for the PR! I have added some minor comments.

mtardy (changes requested):
Thanks I think this is good since I was trying to clean up the cmd recently https://github.com/cilium/tetragon/pull/5188. However I feel like in this case you can drop the test, we know RunE should work. Similarly I agree with Anastasios maybe renaming the import might be unecessary.

I think a 2/3 lines change to just use RunE and return the value from Bugtool should be enough here :) Thanks!

cmd/tetra/bugtool/bugtool.go
immanuwell: got it

cmd/tetra/bugtool/bugtool.go
immanuwell: removed the runBugtool indirection

cmd/tetra/bugtool/bugtool_test.go
immanuwell: removed the test file entirely

immanuwell:
@tpapagian @mtardy addressed all suggestions, please take a look

tpapagian (approved):
LGTM, thanks! There are some issues related to the CI that are unrelated to that PR.

mtardy (approved):
Thanks!

tpapagian:
Can you please rebase that PR? With that, CI should be good.

immanuwell:
@tpapagian done, rebaised

tpapagian (approved):

PR #5231, opened 2026-07-03
cgidmap: Add cgroupfs fallback for pod association

### LINKED ISSUE #3381: cgidmap depends on cri for pod association.
https://github.com/cilium/tetragon/pull/2776 Introduced pod association via cgroup ids.

Currently, there are two ways that the cgidmap can be updated.

By contacting the CRI: https://github.com/cilium/tetragon/blob/0bf690aed79a6e0cf4ff9199454eb25d3063fd96/pkg/cgidmap/cri.go#L105
Or, in the runtime hooks: https://github.com/cilium/tetragon/blob/0bf690aed79a6e0cf4ff9199454eb25d3063fd96/pkg/cgidmap/rthooks.go#L70

This means that without `--enable-cri` being set, tetragon will not be able to do pod association for existing pods (even if runtime hooks are used).

One of the benefits of talking to the CRI is that it provides authoritative answers.

If `--enable-cri` is not set, we could scan the cgroup filesystem as we do in the policyfilter code: https://github.com/cilium/tetragon/blob/0bf690aed79a6e0cf4ff9199454eb25d3063fd96/pkg/policyfilter/cgroupid.go#L24-L36

sayboras:
### Description

Without `--enable-cri`, cgidmap could not associate existing pods. Generalize the async resolver into a backend-agnostic queue/worker and add a best-effort cgroupfs-scan fallback (mirroring policyfilter), so pod association keeps working for existing pods when the CRI is disabled. Covered by a new e2e test.

Fixes #3381

### Changelog

```release-note
cgidmap: Add cgroupfs fallback for pod association when CRI is disabled
```

### DIFF
--- a/pkg/cgidmap/resolver.go
+++ b/pkg/cgidmap/resolver.go
@@ -1,0 +1,137 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build !windows && !nok8s
+
+package cgidmap
+
+import (
+	"fmt"
+	"slices"
+	"sync"
+
+	"github.com/cilium/tetragon/pkg/cgroups"
+	"github.com/cilium/tetragon/pkg/cgtracker"
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/logger/logfields"
+	"github.com/cilium/tetragon/pkg/metrics"
+	"github.com/cilium/tetragon/pkg/option"
+)
+
+// asynchronous resolution of unmapped container ids to cgroup ids. The queue and
+// worker are backend-agnostic; a backend only provides a containerPathFn (how to
+// find a container's cgroup path) and metric counters.
+
+const (
+	// if, for whatever reason, we cannot resolve ids we dont want the queue to grow
+	// unbounded. Hence, we only keep the last 128 unresolved ids added. If we manage to catch
+	// up, subsequent sync updates from the pod hooks will ensure that the ids that were dropped
+	// will be added if they are still alive.
+	maxUnmappedIDs = 128
+)
+
+type unmappedID struct {
+	podID  PodID
+	contID ContainerID
+}
+
+// containerPathFn returns the absolute host cgroup path for a container. It is the
+// only part of resolution that differs between the CRI and cgroupfs backends.
+type containerPathFn func(unmappedID) (string, error)
+
+type resolver struct {
+	mu   sync.Mutex
+	cond sync.Cond
+	// unresolvedIDs holds pending ids, oldest first; the worker pops the
+	// newest from the end (LIFO).
+	unresolvedIDs []unmappedID
+
+	m             Map
+	containerPath containerPathFn
+	// getCgroupID derives the cgroup id from a path. The choice depends only
+	// on how cgidmap keys its map (EnableCgTrackerID), not on the backend.
+	getCgroupID func(string) (uint64, error)
+	attempted   *metrics.Counter
+	errored     *metrics.Counter
+}
+
+func newResolver(m Map, containerPath containerPathFn, attempted, errored *metrics.Counter) *resolver {
+	getCgroupID := cgroups.GetCgroupIDFromSubCgroup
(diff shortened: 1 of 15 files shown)


andrewstrohman:
>  add a best-effort cgroupfs-scan fallback (mirroring policyfilter),

Why do you consider this best effort? What's an example scenario where cgroupfs-scan would not work correctly? Is it possible to make the cgoupfs-scan method reliable and always use that method regardless if `--enable-cri` is enabled?

pkg/cgidmap/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+
+// containerPathFn returns the absolute host cgroup path for a container. It is the
+// only part of resolution that differs between the CRI and cgroupfs backends.
+type containerPathFn func(unmappedID) (string, error)
+
+type resolver struct {
+	mu   sync.Mutex
+	cond sync.Cond
+	// unresolvedIDs holds pending ids, oldest first; the worker pops the
+	// newest from the end (LIFO).
```
andrewstrohman, line 46: Why is LIFO preferable to FIFO?

tests/e2e/tests/cgidmap/cgidmap_test.go
```diff
@@ -0,0 +1,116 @@
...
+		WithTimeLimit(2 * time.Minute)
+
+	metricsChecker := metricschecker.NewMetricsChecker("cgidmapFallbackMetrics")
+
+	runEventChecker := features.New("Check pod association").
+		Assess("Exec events carry pod info", rpcChecker.CheckInNamespace(1*time.Minute, namespace)).
+		Feature()
+
+	checkMetrics := features.New("Check cgfs resolver metrics").
+		Assess("Wait for event checker", rpcChecker.Wait(60*time.Second)).
```
andrewstrohman, line 110: I think the idea here is that we want to check metrics after all the events have arrived.

But Wait() will return after the event checker starts, not after the event checker concludes. So, it seems like we can have a race where the event checker has started, but we haven't received events yet, we check metrics and see the value is 0, and then the first event is generated.

pkg/cgidmap/resolver.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return fmt.Errorf("find container path: %w", err)
+	}
+
+	cgID, err := r.getCgroupID(path)
+	if err != nil {
+		return fmt.Errorf("get cgroup id: %w", err)
+	}
+
+	if err := cgtracker.AddCgroupTrackerPath(path); err != nil {
```
FedeDP, line 132: What if we pass this helper just like you did for `criContainerPath`? Ie: we do not invoke directly the `cgtracker.Foo` helper, but instead we receive the function to be called in the `newResolver` constructor?
At that point, you could drop `AddCgroupTrackerPath` changes and it's IMO cleaner since `resolver` is not going to depend upon `cgtracker` anymore.

tests/e2e/tests/cgidmap/cgidmap_test.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+			ctx, err = helpers.LoadCRDString(namespace, workloadPod, true)(ctx, cfg)
+			if err != nil {
+				return ctx, fmt.Errorf("failed to create workload pod: %w", err)
+			}
+			return tetragon.Install(
+				tetragon.WithHelmOptions(map[string]string{
+					"tetragon.exportAllowList": "",
+					"tetragon.cgidmap.enabled": "true",
+					// tetragon.cri.enabled stays false (chart default)
```
FedeDP, line 74: I would be explicit; if we would ever update the default we would no more test the case here.

pkg/cgidmap/resolver.go
kkourt: The thinking is that it's better to drop old request where things might have changed rather than new requests.

kkourt (commented):
Thanks for working on this!

> cgidmap: generalize the async resolver

The above patch is not easy for me to review. Can you split it into patches: One where you introduce the "backend-agnostic" resolver (presumably, with no functional changes in the code), and one or more patches with the rest of the changes?

PR #5246, opened 2026-07-06
 ci: set new checkout action flag to allow fork checkouts

3u13r:
I'm pretty sure we've broken the buiold-ci-images workflow because we've upgraded to v7 of the checkout action: https://github.com/actions/checkout/pull/2454

### Description
With v7 of the actions/checkout action one has to set the new
`allow-unsafe-pr-checkout: true` flag in order to checkout branches from
forks in a workflow with pull_request_target.

### DIFF
--- a/.github/workflows/build-images-ci.yml
+++ b/.github/workflows/build-images-ci.yml
@@ -92,6 +92,7 @@ jobs:
           persist-credentials: false
           ref: ${{ steps.tag.outputs.tag }}
           fetch-depth: 0
+          allow-unsafe-pr-checkout: true
 
       - name: Login to quay.io for CI
         uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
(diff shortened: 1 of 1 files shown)


tpapagian:
Yes, this is done in https://github.com/cilium/tetragon/pull/5233. But things are still failing on this PR. We can also revert that if we need to do that.

3u13r:
The checkout still fails because 

> This event runs in the context of the default branch of the base repository, rather than in the context of the merge commit

So we need the change in the workflow file on `main` in order to affect e.g., this PR.

I'd argue with merging this PR instead of staying behind on the checkout action version, because we want other updates and this change "just" uses the security assumptions from the previous version.

mtardy (approved):
yep! thanks

tpapagian:
> I'd argue with merging this PR instead of staying behind on the checkout action version, because we want other updates and this change "just" uses the security assumptions from the previous version.

Let's merge it!

tpapagian (approved):

PR #5247, opened 2026-07-07
support for policytest in tetragon-vmtest-run

kkourt:
This PR introduces a first version for running policytests in LVH.

Dependency PRs:
 - [x] https://github.com/cilium/tetragon/pull/5255
 - [x] https://github.com/cilium/tetragon/pull/5256
 - [x] https://github.com/cilium/tetragon/pull/5312
 
 
TODO (followups):
 - [ ] Create a common workflow for tarball creation to avoid duplicate work
 - [ ] remove fetch-data.sh script (also from vmtests workflow)
 - [ ] use policytest for packages e2e test


kkourt:
Let me know if I should split into multiple PRs.

cmd/tetragon-vmtests-run/conf.go
```diff
@@ -35,6 +36,22 @@ type testConf struct {
 	filesystems []QemuFS
 }
 
+func cmdAddTestConfFlags(cmd *cobra.Command, cnf *testConf) {
```
FedeDP, line 45: What if we make this a method on `testConf` that takes `flags` instead of while `cmd` as input param?

cmd/tetragon-vmtests-run/policytest.go
```diff
@@ -264,6 +264,38 @@ func mustCopyTesterProgsCmd(hostDir string, tmpDir string) images.Action {
 	}}
 }
 
+func buildTetragonActions(ptConf *PolicyTestConf, tmpDir string) ([]images.Action, error) {
```
FedeDP: I'd possibly merge this change with the policytest introduction; after all it is just an internal refactor that isn't introducing any new feat.

cmd/tetragon-vmtests-run/policytest.go
```diff
@@ -0,0 +1,391 @@
...
+	return fname, nil
+}
+
+func mustMakeTetragonServiceFile(fname string) string {
+	fname, err := makeTetragonServiceFile(fname)
+	if err != nil {
+		panic(err)
+	}
+	return fname
+}
```
FedeDP: IMHO i'd move `tetragonService` stuff to a new `service.go` file; same for `policytester_service.go` and `tester_progs.go`; perhaps even `build_image.go`?

Also, i'd dump the systemd units as `.service` files and `go:embed` them.

mtardy:
> Let me know if I should split into multiple PRs.

I think it would indeed lower the barrier to review this. Maybe cut it in two?

cmd/tetragon-vmtests-run/conf.go
kkourt: How do we deal with `MargFlagRequired`?

```go
func cmdAddTestConfFlags(cmd *cobra.Command, cnf *testConf) {
	cmd.Flags().StringVar(&cnf.baseImageFilename, "base", "", "base image filename")
	cmd.MarkFlagRequired("base")
```

cmd/tetragon-vmtests-run/policytest.go
kkourt: done

cmd/tetragon-vmtests-run/conf.go
FedeDP: Oh i see, let's pass cmd to it then :)

cmd/tetragon-vmtests-run/policytest.go
kkourt: OK, created `policytester_service.go` `tester_progs.go` and `tetragon_service.go`.

> Also, i'd dump the systemd units as .service files and go:embed them.

Any particular reasons for doing that? I like having them in the same file because it's easier to inspect the templates values in the template and what we set when we render it.

FedeDP: Mmh it's because if you have them in a `.service` file, then you have syntax highlighting, and it's IMO easier to inspect for script errors. Also, that way the systemd unit is going to be versioned by itself and it's easier to follow its evolution (even if i don't think we'll need to touch it). Not a big deal, though, feel free to skip the suggestion :)

kkourt:
> > Let me know if I should split into multiple PRs.
> 
> I think it would indeed lower the barrier to review this. Maybe cut it in two?

https://github.com/cilium/tetragon/pull/5312

FedeDP (approved):
LGTM, thank you!

PR #5248, opened 2026-07-07
add ability to disable process cache

andrewstrohman:
The process cache can cause a lot of memory usage when many processes are executed on a system. The cache itself, and its deleteQueue expand under rapid process creation. Process churn can also cause significant memory fragmentation. This potential for an explosion of memory usage makes it difficult to estimate an upper bound of how much memory tetragon will use.

This change allows users to disable the process cache in order prevent its memory usage and fragmentation at the cost of not being able to reliably enrich events with process information.

## Testing

See [this PR](https://github.com/cilium/tetragon/pull/5354) for how I'm testing this change.

## Issue
https://github.com/cilium/tetragon/issues/5420

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -531,6 +531,10 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 
 	if option.Config.DisableProcessCache {
 		log.Info("Process cache is disabled")
+		// The k8s watcher is used to retrieve pod metadata independently of
+		// the process cache, so it must still be set for pod info to be
+		// attached to events.
+		process.SetK8sWatcher(podAccessor)
 	} else {
 		if err := process.InitCache(podAccessor, option.Config.ProcessCacheSize, pcGCInterval); err != nil {
 			return err
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -529,8 +529,12 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 		pcGCInterval = defaults.DefaultProcessCacheGCInterval
 	}
 
-	if err := process.InitCache(podAccessor, option.Config.ProcessCacheSize, pcGCInterval); err != nil {
-		return err
+	if option.Config.DisableProcessCache {
+		log.Info("Process cache is disabled")
+	} else {
+		if err := process.InitCache(podAccessor, option.Config.ProcessCacheSize, pcGCInterval); err != nil {
+			return err
+		}
 	}
 
 	// cleanupWg is needed to ensure that gRPC code cleanly finishes before we exit (e.g,
--- a/pkg/grpc/exec/exec.go
+++ b/pkg/grpc/exec/exec.go
@@ -44,6 +44,8 @@ func GetProcessExec(event *MsgExecveEventUnix, useCache bool) *tetragon.ProcessE
 	var tetragonParent *tetragon.Process
 	var tetragonAncestors []*tetragon.Process
 	var ancestors []*process.ProcessInternal
+	var parent *process.ProcessInternal
+	var err error
 
 	proc := process.AddExecEvent(event.Unix)
 	tetragonProcess := proc.UnsafeGetProcess()
@@ -51,9 +53,11 @@ func GetProcessExec(event *MsgExecveEventUnix, useCache bool) *tetragon.ProcessE
 	parentId := tetragonProcess.ParentExecId
 	processId := tetragonProcess.ExecId
 
-	parent, err := process.Get(parentId)
-	if err == nil {
-		tetragonParent = parent.UnsafeGetProcess()
+	if !option.Config.DisableProcessCache {
+		parent, err = process.Get(parentId)
+		if err == nil {
+			tetragonParent = parent.UnsafeGetProcess()
+		}
 	}
 
 	// Set the ancestors only if --enable-ancestors flag includes 'base'.
@@ -367,6 +371,10 @@ func (msg *MsgCloneEventUnix) Retry(internal *process.ProcessInternal, _ notify.
 func (msg *MsgCloneEventUnix) HandleMessage() *tetragon.GetEventsResponse {
 	var ancestors []*process.ProcessInternal
 
+	if option.Config.DisableProcessCache {
+		return nil
+	}
+
 	proc, _ := process.AddCloneEvent(&msg.MsgCloneEvent)
 	if option.Config.EnableProcessAncestors && proc.NeededAncestors() {
 		ancestors, _ = process.GetAncestorProcessesInternal(proc.UnsafeGetProcess().ParentExecId)
@@ -664,6 +672,10 @@ func (msg *MsgProcessCleanupEventUnix) Retry(_ *process.ProcessInternal, _ notif
 
 func (msg *MsgProcessCleanupEventUnix) HandleMessage() *tetragon.GetEventsResponse {
 	var ancestors []*process.ProcessInternal
+
+	if option.Config.DisableProcessCache {
+		return nil
+	}
 
 	msg.RefCntDone = [3]bool{false, false, false}
 	proc, parent := process.GetParentProcessInternal(msg.PID, msg.Ktime)
--- a/docs/data/tetragon_flags.yaml
+++ b/docs/data/tetragon_flags.yaml
@@ -35,6 +35,9 @@ options:
     - name: disable-kprobe-multi
       default_value: "false"
       usage: Allow to disable kprobe multi interface
+    - name: disable-process-cache
+      default_value: "false"
+      usage: Disable process cache
     - name: enable-ancestors
       default_value: '[]'
       usage: |
--- a/pkg/api/flags.go
+++ b/pkg/api/flags.go
@@ -82,7 +82,7 @@ const (
 	// EventDataArgs indicates args are received with data event
 	EventDataArgs = 0x1000000
 
-	// EventDataArgs indicates that a clone event's process is in its
+	// EventInInitTree indicates that a clone event's process is in its
 	// container's init process tree.
 	EventInInitTree = 0x2000000
 )
--- a/pkg/grpc/exec/kthread_init.go
+++ b/pkg/grpc/exec/kthread_init.go
@@ -9,6 +9,7 @@ import (
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/api/processapi"
 	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/option"
 	"github.com/cilium/tetragon/pkg/process"
 	"github.com/cilium/tetragon/pkg/reader/notify"
 )
@@ -21,6 +22,9 @@ type MsgKThreadInitUnix struct {
 }
 
 func (msg *MsgKThreadInitUnix) HandleMessage() *tetragon.GetEventsResponse {
+	if option.Config.DisableProcessCache {
+		return nil
+	}
 	proc := process.AddExecEvent(msg.Unix)
 	parent, err := process.Get(proc.UnsafeGetProcess().ParentExecId)
 	if err != nil {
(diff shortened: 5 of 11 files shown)


pkg/grpc/exec/exec.go
```diff
@@ -45,19 +45,21 @@ func GetProcessExec(event *MsgExecveEventUnix, useCache bool) *tetragon.ProcessE
...
 
-	if !option.Config.DisableProcessCache {
-		parent, err = process.Get(parentId)
-		if err == nil {
-			tetragonParent = parent.UnsafeGetProcess()
-		}
+	if option.Config.DisableProcessCache {
+		// best effort: fall back to reading the parent from execve_map
+		parent, _ = process.GetExecParentInternalFromExecveMap(event.Unix)
+	} else if p, err := process.Get(parentId); err == nil {
```
FedeDP: I am wondering whether adding an interface with:
```go
type ProcessFetcher interface {
  GetProcessExec(event *MsgExecveEventUnix, useCache bool) 
  GetParentProcessInternal(pid uint32, ktime uint64) (*ProcessInternal, *ProcessInternal)
}
```
would fit better.

We could have a `CacheProcessFetcher` and an `BPFProcessFetcher`; imo the could would become more readable and it would be up to the API caller to use the desired interface.
Also, it could help with tests possibly.

WDYT?

I imagined something like:
```go
if option.Config.DisableProcessCache {
		log.Info("Process cache is disabled")
		// The k8s watcher is used to retrieve pod metadata independently of
		// the process cache, so it must still be set for pod info to be
		// attached to events.
		process.SetK8sWatcher(podAccessor)
        process.SetFetcher(process.NewBPFProcessFetcher())
	} else {
        // InitCache would internally call
        // process.SetFetcher(process.NewCacheProcessFetcher())
		if err := process.InitCache(podAccessor, option.Config.ProcessCacheSize, pcGCInterval); err != nil {
			return err
		}
	}
```

While this would work fine for `GetParentProcessInternal` (that lives in `process` package), i'm not sure how to let this work for `GetProcessExec` that lives in `exec` package though.

andrewstrohman: > I am wondering whether adding an interface with:

> type ProcessFetcher interface {
  GetProcessExec(event *MsgExecveEventUnix, useCache bool) 
  GetParentProcessInternal(pid uint32, ktime uint64) (*ProcessInternal, *ProcessInternal)
}
would fit better.


GetProcessExec is not completely different based on whether the process cache is disabled. Regardless, we want to call `proc.AnnotateProcess(option.Config.EnableProcessCred, option.Config.EnableProcessNs)` and `event.finalize(tetragonEvent, proc, eventcache.NO_EV_CACHE)`. If I understand you suggestion correctly, both implementations of this interface would need to make these calls. And in the future, if more common logic is necessary, we will have to be careful to add it to both implementations.

On the other hand, GetParentProcessInternal is completely different depending on mode, so I think it makes more sense to have two separate implementations for that one.

> We could have a CacheProcessFetcher and an BPFProcessFetcher; imo the could would become more readable and it would be up to the API caller to use the desired interface.

This sounds like you're saying that at each call site, we would check whether the process cache is disabled, and depending on that use use a CacheProcessFetcher or a BPFProcessFetcher. Later on, you suggest a ` process.SetFetcher` to set the fetcher. Is the idea that we also have a `process.GetFetcher`, so that we don't need to check  whether the process cache is disabled, and completely avoid the branching logic at the call site?

> Also, it could help with tests possibly.

I understand that using indirection can help testing (ie mocking), but it's not clear to me what you're think ing here. Can you please elaborate with a specific example?

>I imagined something like:
> ...
>        // InitCache would internally call
>        // process.SetFetcher(process.NewCacheProcessFetcher())

Why is it better to call `SetFetcher` within InitCache as opposed to calling it where this comment is? Is it so we can avoid updating the other `InitCache` call sites?

> While this would work fine for GetParentProcessInternal (that lives in process package), i'm not sure how to let this work for GetProcessExec that lives in exec package though.

I'm thinking that you are saying that `process.SetFetcher` would set a private (un-exported) variable in the process package, and we would use this variable when calling the new interface's functions. If this is the concern then we could instead store it in exported variable or create a `process.GetFetcher` function to get access to it to use it outside the package.

FedeDP: > GetProcessExec is not completely different based on whether the process cache is disabled. Regardless, we want to call proc.AnnotateProcess(option.Config.EnableProcessCred, option.Config.EnableProcessNs) and event.finalize(tetragonEvent, proc, eventcache.NO_EV_CACHE). If I understand you suggestion correctly, both implementations of this interface would need to make these calls. And in the future, if more common logic is necessary, we will have to be careful to add it to both implementations.

Yes, i agree it's not perfect but it gets more readable IMO. Unless we are able to cut the "fetcher-specific" code in `GetProcessExec` and let the interface only implement that.
After a check, i think we could let the interface be something like:
```
type ProcessFetcher interface {
  GetParent(event *MsgExecveEventUnix, parentID uint64) (*process.ProcessInternal, cleanupEvent, error)
  GetParentProcessInternal(pid uint32, ktime uint64) (*ProcessInternal, *ProcessInternal)
}
```
at that point, `GetProcessExec` becomes:
```go
parent, cleanup, err := procFetcher.GetParent(event, parentId)
if err == nil {
	tetragonParent = parent.UnsafeGetProcess()
}

...

// do we need to cleanup anything?
if cleanupEvent != nil {
	cleanupEvent.HandleMessage()
}
```


> Is the idea that we also have a process.GetFetcher, so that we don't need to check whether the process cache is disabled, and completely avoid the branching logic at the call site?

Exactly!

> Can you please elaborate with a specific example?

Was exactly talking about mocking :)
No need to touch `option.Config.DisableProcessCache` at that point, we can just `process.SetFetcher(myFetcher)`.


> Why is it better to call SetFetcher within InitCache as opposed to calling it where this comment is? Is it so we can avoid updating the other InitCache call sites?

Yes exactly, also since we are going to `InitCache()`, it seems like a good point in the code to create the `CacheProcessFetcher`. But no strong opinions.

> I'm thinking that you are saying that process.SetFetcher would set a private (un-exported) variable in the process package, and we would use this variable when calling the new interface's functions. If this is the concern then we could instead store it in exported variable or create a process.GetFetcher function to get access to it to use it outside the package.

Yes exactly, that would work; didn't want to expose it but i think that's the only option.

Also, feel free to just discard my suggestion if that seems too complex. My idea was to make the code more readable, but if the outcome is the opposite, let's skip it!

andrewstrohman: OK, thanks for this thoughtful suggestion and responses. I'll give it a try.

> at that point, GetProcessExec becomes:

> parent, cleanup, err := procFetcher.GetParent(event, parentId)
if err == nil {
	tetragonParent = parent.UnsafeGetProcess()
}

> ...

> // do we need to cleanup anything?
if cleanupEvent != nil {
	cleanupEvent.HandleMessage()
}

I see that you have cleanupEvent returned from GetParent in order to prevent the gate of calling cleanup.HandleMessage(). Instead, I'll move the gate to inside the cleanupEvent.HandleMessage() function, which seems simple and more resilient to future changes.

FedeDP: > Instead, I'll move the gate to inside the cleanupEvent.HandleMessage() function, which seems simple and more resilient to future changes.

I agree, thanks for the changes.

FedeDP (approved):
LGTM, thank you!

andrewstrohman:
cilium/tetragon-committers, please do not merge this PR -- I'll take care of merging it. I want to give others a chance to review, and want to do some more testing myself.

kkourt:
> cilium/tetragon-committers, please do not merge this PR -- I'll take care of merging it. I want to give others a chance to review, and want to do some more testing myself.

One good way to indicate that the PR is not to be merged is to move it to Draft. Doing so.

pkg/grpc/exec/exec.go
```diff
@@ -367,6 +371,10 @@ func (msg *MsgCloneEventUnix) Retry(internal *process.ProcessInternal, _ notify.
 func (msg *MsgCloneEventUnix) HandleMessage() *tetragon.GetEventsResponse {
 	var ancestors []*process.ProcessInternal
 
+	if option.Config.DisableProcessCache {
```
kkourt, line 374: Instead of having these checks in `pkg/grpc`, it would be nice if we had the `DisableProcessCache` only inside `pkg/process`. Ideally, only inside `pkg/process/cache.go`. The next commit seems to do something to that effect, but see my other comment regarding the last commit.

cmd/tetragon/main.go
```diff
@@ -527,6 +527,7 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 		// the process cache, so it must still be set for pod info to be
 		// attached to events.
 		process.SetK8sWatcher(podAccessor)
+		process.SetFetcher(&process.BPFProcessFetcher{})
```
kkourt: If I understand correctly, we are reading a BPF map for every event? I'm thinking that this could cause a significant CPU overhead (especially for high-event counts) and potential contention among different cores. Thoughts?

Maybe it's  worth having a first PR where we always omit process information for events other than exec and revisiting the last patch in a different PR?

kkourt (commented):
Thanks! Please find some first comments below.

cmd/tetragon/main.go
andrewstrohman: > If I understand correctly, we are reading a BPF map for every event? I'm thinking that this could cause a significant CPU overhead (especially for high-event counts) and potential contention among different cores. Thoughts?

I agree this needs to be evaluated before this part of the change can be accepted. 

> Maybe it's worth having a first PR where we always omit process information for events other than exec and revisiting the last patch in a different PR?

Sounds good. I dropped the last commit that attempted best-effort retrieval of process info from the execve map. I created [this issue](https://github.com/cilium/tetragon/issues/5457), in case we want to revisit this later. 

With regard to the the exec events, here is an example of what I currently have without best effort:
```
{
   "node_name" : "builder",
   "process_exec" : {
      "process" : {
         "arguments" : "180",
         "auid" : 1000,
         "binary" : "/usr/bin/sleep",
         "cwd" : "/home/andy",
         "exec_id" : "YnVpbGRlcjozNDQ5ODM4MjEzMjk1NTkzOjI2MDgxOA==",
         "flags" : "execve clone",
         "in_init_tree" : false,
         "parent_exec_id" : "YnVpbGRlcjozNDQ4MDM4MTQwMDAwMDAwOjI0NjY2NQ==",
         "pid" : 260818,
         "start_time" : "2026-08-18T22:19:02.343808236Z",
         "tid" : 260818,
         "uid" : 1000
      }
   },
   "time" : "2026-08-18T22:19:02.343807821Z"
}
```
Notice how there is no `parent` included. At one point, we talked about how users could manually reconstruct the process ancestry by examining the events, if need be. They could still do this by base64 decoding `parent_exec_id`, but I want to make sure you don't think omitting the parent information (in particular the the parent pid) is a problem.

pkg/grpc/exec/exec.go
andrewstrohman: To summarize our offline discussion about this:
- Doing it this way makes the code more resilient to future changes, which might not account for the case of the process cache being disabled.
- We talked about whether the Fetcher interface would work for this. The idea was that we still would use the Fetcher interface, but have it return nil, instead of best effort from the execve map, when the process cache is disabled. This approach would not completely satisfy this concern as-is, because of the early return cases. You mentioned that if the abstraction is not complete, you would not want to proceed with something partial. We talked about refactoring in order to have a complete, centralized abstraction, but I'm not seeing an easy way to do this at this time. As such, I'd prefer to not gate this PR on achieving this.
- To some extent, adding policytest invocations to our CI, where it targets a tetragon instance running with the process cache disabled, alleviates this concern. I've created [this issue](https://github.com/cilium/tetragon/issues/5458) to keep track of this.

cmd/tetragon/main.go
kkourt: > I want to make sure you don't think omitting the parent information (in particular the the parent pid) is a problem.

Thanks for flagging this. Including some parent information (e.g., the parent pid or the parent exec_id) might be a useful thing to do, but, from my perspective, I'm fine if we leave it as a potential follow-up.

kkourt (approved):

PR #5252, opened 2026-07-08
Pr/mstemm/make deleted pod cache size configurable

### LINKED ISSUE #5251: Deleted pod lru cache can be too small for busy environments
In a busy environment with lots of pods/containers coming and going, the default size of 128 for the deleted pod lru cache at https://github.com/cilium/tetragon/blob/ad83b7679333f4e6d63cbc55d941b779c583f81c/pkg/watcher/deleted_pod_cache.go#L27 might be too small.

Make the cache size configurable (and bigger) and add logs/metrics to show how often items are being evicted from the cache.

mstemm:
Fixes #5251.

### Description
In busy environments, there is a lot of stress put on the deleted pod lru cache, leading to evictions pretty quickly. Make the size configurable, increase the default size, and add logging/metrics to track when items are evicted.

```release-note
Make the size of the deleted pod lru cache configurable and increase the default to 1024 items. Add additional debug logging and metrics to track when items are evicted from the deleted pod lru cache.
```

Note: This pull request was authored with the assistance of Github Copilot. Reviewers should keep this in mind during code review.

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package constants
+
+const (
+	WatcherDeletedPodCacheSize = 1024
+)
--- a/pkg/metrics/watchermetrics/watchermetrics.go
+++ b/pkg/metrics/watchermetrics/watchermetrics.go
@@ -61,12 +61,19 @@ var (
 		"The total hits for pod information in the deleted pod cache.",
 		nil, nil, nil,
 	), nil)
+
+	WatcherDeletedPodCacheEvictions = metrics.MustNewCounter(metrics.NewOpts(
+		consts.MetricsNamespace, "", "watcher_delete_pod_cache_evictions",
+		"The total evictions from the deleted pod cache.",
+		nil, nil, nil,
+	), nil)
 )
 
 func RegisterMetrics(group metrics.Group) {
 	group.MustRegister(WatcherErrors)
 	group.MustRegister(WatcherEvents)
 	group.MustRegister(WatcherDeletedPodCacheHits)
+	group.MustRegister(WatcherDeletedPodCacheEvictions)
 }
 
 func InitMetrics() {
@@ -74,6 +81,7 @@ func InitMetrics() {
 	GetWatcherEvents(K8sWatcher).Add(0)
 	GetWatcherErrors(K8sWatcher, FailedToGetPodError).Add(0)
 	GetWatcherDeletedPodCacheHits().Add(0)
+	GetWatcherDeletedPodCacheEvictions().Add(0)
 }
 
 // Get a new handle on an WatcherEvents metric for a watcher type
@@ -88,4 +96,8 @@ func GetWatcherErrors(watcherType Watcher, watcherError ErrorType) prometheus.Co
 
 func GetWatcherDeletedPodCacheHits() prometheus.Counter {
 	return WatcherDeletedPodCacheHits.WithLabelValues()
+}
+
+func GetWatcherDeletedPodCacheEvictions() prometheus.Counter {
+	return WatcherDeletedPodCacheEvictions.WithLabelValues()
 }
--- a/pkg/option/flags.go
+++ b/pkg/option/flags.go
@@ -16,6 +16,7 @@ import (
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 
+	"github.com/cilium/tetragon/pkg/constants"
 	"github.com/cilium/tetragon/pkg/defaults"
 	"github.com/cilium/tetragon/pkg/logger"
 	"github.com/cilium/tetragon/pkg/logger/logfields"
@@ -32,6 +33,7 @@ const (
 	KeyVerbosity              = "verbose"
 	KeyProcessCacheSize       = "process-cache-size"
 	KeyDataCacheSize          = "data-cache-size"
+	KeyDeletedPodCacheSize    = "deleted-pod-cache-size"
 	KeyProcessCacheGCInterval = "process-cache-gc-interval"
 	KeyForceSmallProgs        = "force-small-progs"
 	KeyForceLargeProgs        = "force-large-progs"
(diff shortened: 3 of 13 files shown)


FedeDP (approved):
LGTM

pkg/option/flags.go
```diff
@@ -455,6 +457,7 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.Int(KeyVerbosity, 0, "set verbosity level for eBPF verifier dumps. Pass 0 for silent, 1 for truncated logs, 2 for a full dump")
 	flags.Int(KeyProcessCacheSize, 65536, "Size of the process cache")
 	flags.Int(KeyDataCacheSize, 1024, "Size of the data events cache")
+	flags.Int(KeyDeletedPodCacheSize, 1024, "Size of the deleted pod cache")
```
mtardy: nit: I've seen you put those 1024 two times in the Go code at least, just fyi we have `pkg/constants`

pkg/metrics/watchermetrics/watchermetrics.go
```diff
@@ -61,19 +61,27 @@ var (
 		"The total hits for pod information in the deleted pod cache.",
 		nil, nil, nil,
 	), nil)
+
+	WatcherDeletedPodCacheEvictions = metrics.MustNewCounter(metrics.NewOpts(
+		consts.MetricsNamespace, "", "watcher_delete_pod_cache_evictions",
+		"The total evictions from the deleted pod cache.",
+		nil, nil, nil,
+	), nil)
```
mtardy, line 69: ahah I was going to comment on the previous patch "you could add a metric for that" lol, cool

pkg/manager/manager_integration_test.go
```diff
@@ -34,13 +37,16 @@ const (
...
+	testEnv                     *envtest.Environment
+	manager                     *ControllerManager
+	originalDeletedPodCacheSize int
 }
 
 func (suite *ManagerTestSuite) SetupSuite() {
 	err := os.Setenv("NODE_NAME", nodeName)
 	require.NoError(suite.T(), err)
+	suite.originalDeletedPodCacheSize = option.Config.DeletedPodCacheSize
+	option.Config.DeletedPodCacheSize = 1
```
mtardy: I feel like this should not be in the full test suite but only in the appropriate test nop?

pkg/manager/manager_integration_test.go
```diff
@@ -135,6 +141,81 @@ func (suite *ManagerTestSuite) TestFindContainer() {
 	assert.Equal(suite.T(), pod.Spec.Containers[0].Name, container.Name)
 }
 
+func (suite *ManagerTestSuite) TestWatcherMetrics() {
```
mtardy: do we really need a full integration test for that? what's the value against having a simple unit test?

I feel like if we already have integration tests to make sure that pod lifecycle is connected correctly to the watcher, for your metric you can just simulate the lifecycle from the pod cache methods and check that metric is okay. wdyt?

mtardy (commented):
looks good overall, thanks for thinking about both debug + metrics here :)

mostly some nits and questions

pkg/option/flags.go
mstemm: Sure, added a constant to pkg/constants and used it in the two go code places that refer to it. I don't know of a way to plumb through the constants into the non go code, but if you have an example let me know.

pkg/manager/manager_integration_test.go
mstemm: Sure, moved to a more limited standalone test that works only with a Deleted Pod Cache.

pkg/manager/manager_integration_test.go
mstemm: See above, replaced the test with a more limited one.

pkg/watcher/deleted_pod_cache_test.go
```diff
@@ -0,0 +1,88 @@
...
+	suite.cache.Add(secondContainerID, deletedPodCacheEntry{
+		pod:        secondPod,
+		contStatus: &secondPod.Status.ContainerStatuses[0],
+	})
+
+	_, _, found = suite.cache.FindContainer(firstContainerID)
+	suite.False(found)
+
+	evictionsAfter := testutil.ToFloat64(watchermetrics.GetWatcherDeletedPodCacheEvictions())
+	suite.InDelta(evictionsBefore+1, evictionsAfter, 1e-9)
```
andrewstrohman, line 79: Can we avoid specifying a tolerance with:
```
suite.Equal(1.0, hitsAfterFirstFind-hitsBefore)
suite.Equal(1.0, evictionsAfter-evictionsBefore)
```

andrewstrohman (approved):

pkg/watcher/deleted_pod_cache_test.go
mstemm: I could be wrong but I believe things get weird once you look for specific values for floats. It's probably not very likely to come up, so I could just look for the specific values, but the test might start failing strangely afterward.

pkg/constants/constants.go
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package constants
+
+const (
+	WATCHER_DELETED_POD_CACHE_SIZE = 1024
```
mtardy: ah just a style thing in Go we shouldn't use uppercase with snake case, just use `WatcherDeletedPodCacheSize`.

Sorry might have been confusing because there are a lot of system constants defined but we generally follow this for the other consts.

pkg/watcher/deleted_pod_cache_test.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+func (suite *DeletedPodCacheTestSuite) TearDownSuite() {
+	option.Config.DeletedPodCacheSize = suite.originalDeletedPodCacheSize
+}
+
+func TestDeletedPodCacheSuite(t *testing.T) {
+	suite.Run(t, new(DeletedPodCacheTestSuite))
+}
```
mtardy, line 88: sorry might be a dumb question: why do you need a suite and not just a normal test?

mtardy (commented):
just a few remarks

pkg/constants/constants.go
mstemm: Sure, fixed.

pkg/watcher/deleted_pod_cache_test.go
mstemm: So I can change the default deleted pod cache size to 1 before the test and back to its original value afterward. I suppose I could move the creation of the pod cache into a single test but this allows for creating additional test cases later.

mtardy: yeah I guess I would do this and transform the test into a suite in the future when needed but that's okay

mtardy (approved):

PR #5253, opened 2026-07-09
fix(cmd/tetra): accept redirected file stdin in getevents

immanuwell:
stdin replay in `tetra getevents` only kicked in for named pipes, so `tetra getevents < events.json` would go try gRPC instead

This makes any non tty stdin use the replay path, so pipes still work and regular file redirection works too

Repro:
1. `printf '%s\n' '{"process_exec":{"process":{"binary":"/bin/echo"}}}' > /tmp/tetra-event.json`
2. `go run ./cmd/tetra getevents < /tmp/tetra-event.json`
3. before this patch it dials `localhost:54321` and fails. with this patch it prints the event.

### DIFF
--- a/cmd/tetra/getevents/getevents.go
+++ b/cmd/tetra/getevents/getevents.go
@@ -201,7 +201,7 @@ redirection of events to the stdin. Examples:
 		},
 		RunE: func(_ *cobra.Command, _ []string) error {
 			fi, err := os.Stdin.Stat()
-			if err == nil && fi.Mode()&os.ModeNamedPipe != 0 {
+			if err == nil && fi.Mode()&os.ModeCharDevice == 0 {
 				// read events from stdin
 				return getEvents(context.Background(), newIOReaderClient(os.Stdin, common.Debug))
 			}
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
Changes LGTM; let me ask for a review by @mtardy too :)

cmd/tetra/getevents/getevents.go
```diff
@@ -201,8 +201,8 @@ redirection of events to the stdin. Examples:
 		},
 		RunE: func(_ *cobra.Command, _ []string) error {
 			fi, err := os.Stdin.Stat()
-			if err == nil && fi.Mode()&os.ModeNamedPipe != 0 {
-				// read events from stdin
+			if err == nil && fi.Mode()&os.ModeCharDevice == 0 {
+				// read events from redirected stdin
```
mtardy: I might just keep the comment as it didn't change, we are still reading from stdin in the case the os.Stdin is not in ModeCharDevice
```suggestion
				// read events from stdin
```

mtardy (commented):
Thanks patch and test looks good! Could you put the comment of the PR details in the commit log? You can put everything, just wrap the text.

immanuwell:
@mtardy done

mtardy (approved):

mtardy:
Thanks a lot!

PR #5254, opened 2026-07-09
sensors: keep LSM enforcement active after agent exit

tpapagian:
When starting tetragon with `./tetragon --btf /sys/kernel/btf/vmlinux --bpf-lib ./bpf/objs/ --keep-sensors-on-exit` and load the following policy with `./tetra tp add pol-kprobes.yaml`

```yaml
kind: TracingPolicy
metadata:
  name: "kprobes-file-open"
spec:
  kprobes:
  - call: "security_file_open"
    syscall: false
    args:
      - index: 0
        type: "file"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/mnt/"
      matchActions:
      - action: Override
        argError: -1
      - action: Post
```

The expected behaviour is to block all open calls to files under `/mnt/`. For example: 
```
$ cat /mnt/aa.txt
cat: /mnt/aa.txt: Operation not permitted (os error 1)
```

As we also use `--keep-sensors-on-exit` if we kill tetragon agent, this bpf programs should remain there and still block that operation. And this is the case.

On the other hand, if we use a lsm based policy such as: 
```yaml
kind: TracingPolicy
metadata:
  name: "lsm-file-open"
spec:
  lsmhooks:
  - hook: "file_open"
    args:
      - index: 0
        type: "file"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/mnt/"
      matchActions:
      - action: Override
        argError: -1
      - action: Post
```
The behaviour should be the same, but enforcement does not work after agent stops.

This patch fixes that by pinning LSM links so enforcement remains active after Tetragon exits with `--keep-sensors-on-exit`. Previously, only the programs and maps remained pinned, while closing the unpinned link detached the LSM programs.

### DIFF
--- a/pkg/sensors/program/loader_linux.go
+++ b/pkg/sensors/program/loader_linux.go
@@ -555,7 +555,7 @@ func LSMOpen(load *Program) OpenFunc {
 	}
 }
 
-func LSMAttach() AttachFunc {
+func LSMAttach(load *Program, bpfDir string) AttachFunc {
 	return func(_ *ebpf.Collection, _ *ebpf.CollectionSpec,
 		prog *ebpf.Program, spec *ebpf.ProgramSpec) (unloader.Unloader, error) {
 
@@ -567,6 +567,10 @@ func LSMAttach() AttachFunc {
 		lnk, err := linkFn()
 		if err != nil {
 			return nil, fmt.Errorf("attaching '%s' failed: %w", spec.Name, err)
+		}
+		if err := LinkPin(lnk, bpfDir, load, spec.Name); err != nil {
+			lnk.Close()
+			return nil, err
 		}
 		return &unloader.RelinkUnloader{
 			UnloadProg: unloader.ProgUnloader{Prog: prog}.Unload,
@@ -779,7 +783,7 @@ func LoadTracingProgram(bpfDir string, load *Program, maps []*Map, verbose int) 
 
 func LoadLSMProgram(bpfDir string, load *Program, maps []*Map, verbose int) error {
 	opts := &LoadOpts{
-		Attach: LSMAttach(),
+		Attach: LSMAttach(load, bpfDir),
 		Open:   LSMOpen(load),
 		Maps:   maps,
 	}
@@ -788,7 +792,7 @@ func LoadLSMProgram(bpfDir string, load *Program, maps []*Map, verbose int) erro
 
 func LoadLSMProgramSimple(bpfDir string, load *Program, maps []*Map, verbose int) error {
 	opts := &LoadOpts{
-		Attach: LSMAttach(),
+		Attach: LSMAttach(load, bpfDir),
 		Maps:   maps,
 	}
 	return loadProgram(bpfDir, load, opts, verbose)


pkg/sensors/program/loader_linux.go
```diff
@@ -568,6 +568,10 @@ func LSMAttach() AttachFunc {
 		if err != nil {
 			return nil, fmt.Errorf("attaching '%s' failed: %w", spec.Name, err)
 		}
+		if err := LinkPin(lnk, bpfDir, load, spec.Name); err != nil {
```
olsajiri, line 571: what will be the final link pin path ? curious what will be after '../link_" because I'm not sure what's in spec.Name for lsm program

I wonder spec.AttachTo would be another choice

tpapagian: By using:
```yaml
spec:
  lsmhooks:
  - hook: "file_open"
    args:
      - index: 0
        type: "file"
```

I added some print statements before `LinkPin(lnk, bpfDir, load, spec.Name)` and the output is:
```
----> spec.Name: generic_lsm_output
----> spec.AttachTo: file_open
level=info msg="Loaded generic LSM program: bpf/objs/bpf_generic_lsm_output_v61.o -> file_open"
----> spec.Name: generic_lsm_event
----> spec.AttachTo: file_open
level=info msg="Loaded generic LSM program: bpf/objs/bpf_generic_lsm_core_v61.o -> file_open"
```

And the programs are:
```
$ sudo tree /sys/fs/bpf/
/sys/fs/bpf/
└── tetragon
    ├── __base__
    │   ├── event_execve
    │   │   ├── execve_calls
    │   │   ├── link
    │   │   └── prog
    │   ├── event_exit
    │   │   ├── link
    │   │   └── prog
    │   ├── execve_map_update
    │   │   └── prog
    │   ├── kprobe_pid_clear
    │   │   ├── link
    │   │   └── prog
    │   └── tg_kp_bprm_committing_creds
    │       ├── link
    │       └── prog
    ├── execve_map
    ├── execve_map_stats
    ├── execve_map_update_data
    ├── lsm-file-open
    │   ├── generic_lsm
    │   │   └── file_open
    │   │       ├── addr4lpm_maps
    │   │       ├── addr6lpm_maps
    │   │       ├── argfilter_maps
    │   │       ├── config_map
    │   │       ├── filter_map
    │   │       ├── link_generic_lsm_event
    │   │       ├── link_generic_lsm_output
    │   │       ├── lsm_calls
    │   │       ├── override_tasks
    │   │       ├── process_call_heap
    │   │       ├── prog
    │   │       ├── string_maps_0
    │   │       ├── string_maps_1
    │   │       ├── string_maps_10
    │   │       ├── string_maps_2
    │   │       ├── string_maps_3
    │   │       ├── string_maps_4
    │   │       ├── string_maps_5
    │   │       ├── string_maps_6
    │   │       ├── string_maps_7
    │   │       ├── string_maps_8
    │   │       ├── string_maps_9
    │   │       ├── string_postfix_maps
    │   │       ├── string_prefix_maps
    │   │       ├── tg_mb_paths
    │   │       ├── tg_mb_sel_opts
    │   │       └── workloads_map
    │   ├── policy_conf
    │   └── selector_stats
    ├── tcpmon_map
    ├── tg_conf_map
    ├── tg_errmetrics_map
    ├── tg_execve_joined_info_map
    ├── tg_execve_joined_info_map_stats
    ├── tg_mbset_gen
    ├── tg_mbset_map
    ├── tg_rb_events
    └── tg_stats_map

11 directories, 51 files
```

So `link_generic_lsm_event` and `link_generic_lsm_output`. By using only `spec.AttachTo` they will possibly collide. As each policy and hook is under a different directory, I don't believe that this cause any issues.

olsajiri: yep should be fine, thnx

olsajiri (approved):

PR #5255, opened 2026-07-09
policytest updates

kkourt:
See commits.
Needed for https://github.com/cilium/tetragon/pull/5247

### DIFF
--- a/cmd/tetra/policytest/policytest.go
+++ b/cmd/tetra/policytest/policytest.go
@@ -9,7 +9,9 @@ import (
 
 	// import tests
 	"context"
+	"errors"
 	"fmt"
+	"iter"
 	"log/slog"
 	"os"
 	"path/filepath"
@@ -119,6 +121,7 @@ func runCmd() *cobra.Command {
 	testBinsPath := filepath.Join(cwd, "contrib/tester-progs")
 	dumpPolicyPath := ""
 	monitorMode := false
+	allParams := false
 	var params map[string]string
 	cmd := cobra.Command{
 		Use:   "run",
@@ -136,10 +139,24 @@ func runCmd() *cobra.Command {
 				},
 			))
 
-			// NB: parameters are applied to all policies
-			paramValues := make(map[string]any)
-			for k, v := range params {
-				paramValues[k] = v
+			var getParamValues func(t *policytest.T) iter.Seq[policytest.ParamVals]
+			if len(params) > 0 && allParams {
+				return errors.New("setting --params conflicts with --all-params")
+			} else if allParams {
+				getParamValues = func(t *policytest.T) iter.Seq[policytest.ParamVals] {
+					return t.AllParamValues()
+				}
+			} else {
+				// NB: parameters are applied to all policies
+				paramValues := make(map[string]any)
+				for k, v := range params {
+					paramValues[k] = v
+				}
+				getParamValues = func(_ *policytest.T) iter.Seq[policytest.ParamVals] {
+					return func(yield func(policytest.ParamVals) bool) {
+						yield(paramValues)
+					}
+				}
 			}
 
 			ctx := context.Background()
@@ -162,12 +179,15 @@ func runCmd() *cobra.Command {
 			var results []*policytest.Result
 			var ptNames []string
 			for _, t := range tests {
-				ptNames = append(ptNames, t.Name)
-				res := runner.RunTest(log, t, &policytest.TestConf{
-					MonitorMode: monitorMode,
-					ParamValues: paramValues,
-				})
-				results = append(results, res)
+				for paramValues := range getParamValues(t) {
+					ptName := fmt.Sprintf("%s (%s)", t.Name, paramValues)
+					ptNames = append(ptNames, ptName)
+					res := runner.RunTest(log, t, &policytest.TestConf{
+						MonitorMode: monitorMode,
+						ParamValues: paramValues,
+					})
+					results = append(results, res)
+				}
 			}
 			runner.Close()
 			policytest.DumpResults(cmd.OutOrStdout(), ptNames, results)
@@ -179,5 +199,6 @@ func runCmd() *cobra.Command {
 	flags.StringVar(&dumpPolicyPath, "dump-policy-path", dumpPolicyPath, "save the policy in the provided path")
 	flags.BoolVar(&monitorMode, "monitor-mode", monitorMode, "set the policy(-ies) in monitor mode before running the test(s)")
 	flags.StringToStringVar(&params, "set-param", map[string]string{}, "Set a policy parameter")
+	flags.BoolVar(&allParams, "all-params", allParams, "Run policy tests using all available parameters")
 	return &cmd
 }
--- a/cmd/tetra/policytest/policytest.go
+++ b/cmd/tetra/policytest/policytest.go
@@ -176,13 +176,14 @@ func runCmd() *cobra.Command {
 			if err != nil {
 				return fmt.Errorf("failed to start local runner: %w", err)
 			}
-			var results []*policytest.Result
-			var ptNames []string
+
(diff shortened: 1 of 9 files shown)


pkg/testutils/policytest/policytest.go
```diff
@@ -39,6 +40,20 @@ type Parameter struct {
 	Name    string
 	Default any
 	Help    string
+	// Values, if set, is used to generate values for this parameter when testing.
+	// Values is meant to hold all values (including the one specified by Default)
+	Values []any
```
FedeDP, line 46: I was thinking if we maybe could use a generic for it, like:
```go
type Parameter[T any] struct {
  ...
  Values []T
}

type T[P any] struct {
  ...
  Params []Parameter[P]
}
```
and then one could instantiate a parameter like:
```go
policytest.NewBuilder[string]("kprobe-lseek").WithLabels("kprobes").
	WithParameter(policytest.Parameter{
		Name:    "Hook",
		Default: "kprobes",
		Values:  []string{"kprobes", "fentries"},
		Help:    "type of hook to use in the policy",
	})
```

or something similar.

cmd/tetra/policytest/policytest.go
```diff
@@ -123,10 +125,17 @@ func runCmd() *cobra.Command {
 	monitorMode := false
 	allParams := false
 	var params map[string]string
+	var outputFormat = "text"
 	cmd := cobra.Command{
 		Use:   "run",
 		Short: "Run Tetragon policy test(s)",
 		Long:  "Run Tetragon policy test(s)",
+		PreRunE: func(_ *cobra.Command, _ []string) error {
+			if outputFormat != "json" && outputFormat != "text" {
```
FedeDP: I don't know whether we ever do it, but the "cobra" way of dealing with enum-like flags is to create a custom enum type and implement the `pflag.Value` interface.
Just a nit, in case you feel adventurous 😆

cmd/tetra/policytest/policytest.go
```diff
@@ -201,17 +202,25 @@ func runCmd() *cobra.Command {
 			}
 			runner.Close()
 
+			out := cmd.OutOrStdout()
+			if outputFile != "" {
+				out, err = os.OpenFile(outputFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0666)
```
FedeDP: Do we need to `defer out.Close()` ?

cmd/tetra/policytest/policytest.go
FedeDP: Oh i saw we do like you did everytime we deal with enum-like flags.
It's good to know; perhaps i'll take my time and propose a patch then :)

pkg/testutils/policytest/policytest.go
kkourt: >   Params []Parameter[P]

Doesn't this mean that all parameters would need to have the same underlying type? In theory, we might want one parameter to take string and another to take int, no?

FedeDP: Yep, i thought that wouldn't be a problem, but you are right, in that case it wouldn't scale. 
Thanks for clarifying it!

cmd/tetra/policytest/policytest.go
kkourt: Did:

```diff
diff --git a/cmd/tetra/policytest/policytest.go b/cmd/tetra/policytest/policytest.go
index 7b347bf7e..72aaa3867 100644
--- a/cmd/tetra/policytest/policytest.go
+++ b/cmd/tetra/policytest/policytest.go
@@ -25,6 +25,13 @@ import (
 	_ "github.com/cilium/tetragon/tests/policytests" // so that tests can be registered
 )
 
+type outputFmt int
+
+const (
+	outputText outputFmt = iota
+	outputJSON
+)
+
 func New() *cobra.Command {
 	tpCmd := &cobra.Command{
 		Use:     "policytest",
@@ -125,17 +132,11 @@ func runCmd() *cobra.Command {
 	monitorMode := false
 	allParams := false
 	var params map[string]string
-	var outputFormat = "text"
+	var outputFormat = outputText
 	cmd := cobra.Command{
 		Use:   "run",
 		Short: "Run Tetragon policy test(s)",
 		Long:  "Run Tetragon policy test(s)",
-		PreRunE: func(_ *cobra.Command, _ []string) error {
-			if outputFormat != "json" && outputFormat != "text" {
-				return fmt.Errorf("invalid value for %q flag: %s", common.KeyOutput, outputFormat)
-			}
-			return nil
-		},
 		RunE: func(cmd *cobra.Command, args []string) error {
 			logLevel := slog.LevelInfo
 			if common.Debug {
@@ -202,7 +203,7 @@ func runCmd() *cobra.Command {
 			runner.Close()
 
 			switch outputFormat {
-			case "json":
+			case outputJSON:
 				for _, res := range results {
 					b, err := json.Marshal(res)
 					if err != nil {
@@ -210,7 +211,7 @@ func runCmd() *cobra.Command {
 					}
 					cmd.Println(string(b))
 				}
-			case "text":
+			case outputText:
 				policytest.DumpResults(cmd.OutOrStdout(), results)
 			}
 			return nil
@@ -222,6 +223,34 @@ func runCmd() *cobra.Command {
 	flags.BoolVar(&monitorMode, "monitor-mode", monitorMode, "set the policy(-ies) in monitor mode before running the test(s)")
 	flags.StringToStringVar(&params, "set-param", map[string]string{}, "Set a policy parameter")
 	flags.BoolVar(&allParams, "all-params", allParams, "Run policy tests using all available parameters")
-	flags.StringVar(&outputFormat, common.KeyOutput, outputFormat, "output format (text|json)")
+	flags.Var(&outputFormat, "output", "output format (text|json)")
 	return &cmd
 }
+
+func (of *outputFmt) Set(v string) error {
+	switch v {
+	case "text":
+		*of = outputText
+	case "json":
+		*of = outputJSON
+	default:
+		return fmt.Errorf("output format must be either \"text\" or \"json\"")
+	}
+
+	return nil
+}
+
+func (of *outputFmt) String() string {
+	switch *of {
+	case outputText:
+		return "text"
+	case outputJSON:
+		return "json"
+	default:
+		return ""
+	}
+}
+
+func (of *outputFmt) Type() string {
+	return "output"
+}
```

cmd/tetra/policytest/policytest.go
kkourt: Let's :)

Pushed:
```diff
diff --git a/cmd/tetra/policytest/policytest.go b/cmd/tetra/policytest/policytest.go
index ab916764f..18fe9a418 100644
--- a/cmd/tetra/policytest/policytest.go
+++ b/cmd/tetra/policytest/policytest.go
@@ -205,10 +205,12 @@ func runCmd() *cobra.Command {
 
 			out := cmd.OutOrStdout()
 			if outputFile != "" {
-				out, err = os.OpenFile(outputFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0666)
+				f, err := os.OpenFile(outputFile, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0666)
 				if err != nil {
 					return fmt.Errorf("error opening output file: %w", err)
 				}
+				defer f.Close()
+				out = f
 			}
 
 			switch outputFormat {
```

cmd/tetra/policytest/policytest.go
FedeDP: I am introducing a generic way of doing that: https://github.com/cilium/tetragon/pull/5258
But this is good :) thank you for taking the time to do that!

FedeDP: (will eventually take care of rebasing mine once yours is merged to port this to the new `option.Enum`!)

FedeDP (approved):
LGTM CI aside :)

PR #5256, opened 2026-07-09
contrib/tester-progs: have libs relative to binary

kkourt:
We want to be able to execute tester progs from different directories. That is, copy the binary and the .so file in different directories and execute them from them.

See 061e7fc4680a65ccec406e28b6b9bbf22cff1f96

### DIFF
--- a/contrib/tester-progs/Makefile
+++ b/contrib/tester-progs/Makefile
@@ -106,7 +106,7 @@ usdt: usdt.c
 	$(GCC) -Wall $< -o $@
 
 usdt-args: usdt-args.c tester-lib.h libtester.so
-	$(GCC) -Wall $< -o $@ $(CURDIR)/libtester.so
+	$(GCC) -Wall $< -o $@ -L . -ltester -Wl,-rpath,'$$ORIGIN'
 
 capabilities-ioperm: capabilities-ioperm.c
 	$(GCC) -Wall $< -o $@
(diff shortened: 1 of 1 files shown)


FedeDP (approved):
Lovely :)

andrewstrohman:
The loader will prefer LD_LIBRARY_PATH over $ORIGIN, unless we specify `--disable-new-dtags`.

I'm not sure if we care about this, but I thought it's worth pointing out.

andrewstrohman (approved):

kkourt:
> The loader will prefer LD_LIBRARY_PATH over $ORIGIN, unless we specify `--disable-new-dtags`.
> 
> I'm not sure if we care about this, but I thought it's worth pointing out.

What's the implication of the above?

At least in my tests,  dropping the `.so` and the binary file in a dir somewhere seems to work. Do you mean that this will break if LD_LIBRARY_PATH is set?

andrewstrohman:
> What's the implication of the above?
> 
> At least in my tests, dropping the `.so` and the binary file in a dir somewhere seems to work. Do you mean that this will break if LD_LIBRARY_PATH is set?

I don't think it will break, unless there is a different `libtester.so` in a directory specified by LD_LIBRARY_PATH.

I think `--disable-new-dtags` makes the loader prefer the shared library in the same directory as the binary over another version of it in a path specified by LD_LIBRARY_PATH.

A demonstration follows.

```
andy@builder:~$ env | grep LD_LIBRARY_PATH
LD_LIBRARY_PATH=/home/andy/ld_path
andy@builder:~$ pwd
/home/andy
andy@builder:~$ gcc -Wall maintest.c -o maintest -L . -ltest -Wl,-rpath,'$ORIGIN',--disable-new-dtags
andy@builder:~$ ldd maintest | grep test
        libtest.so => /home/andy/./libtest.so (0x0000777c5fb97000)
andy@builder:~$ gcc -Wall maintest.c -o maintest -L . -ltest -Wl,-rpath,'$ORIGIN'
andy@builder:~$ ldd maintest | grep test
        libtest.so => /home/andy/ld_path/libtest.so (0x000079b8b5743000)
```

So, if we want to prioritize the version of the library in the same directory of the binary, we could use `--disable-new-dtags` to do that.

PR #5257, opened 2026-07-09
policy: Report node filtered policies as skipped

sayboras:
### Description
A TracingPolicy whose nodeSelector does not match a node was silently absent from `tetra tracingpolicy list`, making a gated-out policy indistinguishable from a missing one. Track such a policy in the sensor manager with a new TP_STATE_SKIPPED state without loading any BPF.

Relates: https://github.com/cilium/tetragon/pull/5189

### Changelog

```release-note
policy: Report node filtered policies as skipped
```

----
### Testing

Testing was done locally, please find below sample output

```
tetragon-dev-control-plane:/# tetra tracingpolicy list
ID   NAME                  DOMAIN   STATE     FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE           NPOST      NENFORCE   NMONITOR
3    ns-test-namespaced    k8s      skipped   0          default                      0 B            unknown        0          0          0
4    ns-test-match         k8s      skipped   0          (global)                     0 B            unknown        0          0          0
5    ns-test-no-selector   k8s      enabled   0          (global)    generic_kprobe   2.75 MB        monitor_only   93302064   0          0
2    ns-test-empty         k8s      enabled   0          (global)    generic_kprobe   2.75 MB        monitor_only   93302928   0          0

# Update node label to match all above policies
tetragon-dev-control-plane:/# tetra tracingpolicy list
ID   NAME                  DOMAIN   STATE     FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE           NPOST      NENFORCE   NMONITOR
7    ns-test-empty         k8s      enabled   0          (global)    generic_kprobe   2.75 MB        monitor_only   267659     0          0
6    ns-test-namespaced    k8s      enabled   6          default     generic_kprobe   2.75 MB        monitor_only   0          0          0
8    ns-test-match         k8s      enabled   0          (global)    generic_kprobe   2.75 MB        monitor_only   239368     0          0
5    ns-test-no-selector   k8s      enabled   0          (global)    generic_kprobe   2.75 MB        monitor_only   94512235   0          0

```


pkg/metrics/policymetrics/policymetrics.go
```diff
@@ -131,22 +144,9 @@ func collect(ch chan<- prometheus.Metric) {
...
-		float64(counters[tetragon.TracingPolicyState_TP_STATE_DISABLED]),
-		strings.TrimPrefix(strings.ToLower(tetragon.TracingPolicyState_TP_STATE_DISABLED.String()), "tp_state_"),
-	)
-	ch <- policyState.MustMetric(
-		float64(counters[tetragon.TracingPolicyState_TP_STATE_ENABLED]),
-		strings.TrimPrefix(strings.ToLower(tetragon.TracingPolicyState_TP_STATE_ENABLED.String()), "tp_state_"),
-	)
+	for i, state := range policyStates {
+		ch <- policyState.MustMetric(float64(counters[state]), policyStateNames[i])
+	}
```
mtardy, line 149: I'd put these changes in a separate commit before the main commit if possible

pkg/metrics/policymetrics/policymetrics_test.go
```diff
@@ -21,20 +21,26 @@ import (
...
+			skippedKernelMemory = "tetragon_tracingpolicy_kernel_memory_bytes{policy=\"salad\", policy_namespace=\"\"} 0\n"
+		}
 		return strings.NewReader(fmt.Sprintf(`# HELP tetragon_tracingpolicy_kernel_memory_bytes The amount of kernel memory in bytes used by policy's sensors non-shared BPF maps (memlock).
 # TYPE tetragon_tracingpolicy_kernel_memory_bytes gauge
 tetragon_tracingpolicy_kernel_memory_bytes{policy="pizza", policy_namespace=""} 0
 tetragon_tracingpolicy_kernel_memory_bytes{policy="amazing-one", policy_namespace=""} 0
 tetragon_tracingpolicy_kernel_memory_bytes{policy="amazing-one", policy_namespace="default"} 0
 tetragon_tracingpolicy_kernel_memory_bytes{policy="amazing-one", policy_namespace="kube-system"} 0
-# HELP tetragon_tracingpolicy_loaded The number of loaded tracing policy by state.
+%s# HELP tetragon_tracingpolicy_loaded The number of loaded tracing policy by state.
```
mtardy: what prevent us to make this non-conditional like the other metrics at 0?

pkg/sensors/collection.go
```diff
@@ -33,6 +33,8 @@ const (
 	ErrorState
 	LoadingState
 	UnloadingState
+	// SkippedState means the policy is tracked but not loaded on this node.
+	SkippedState
```
mtardy: could you find a better explicit name that would require no comment?

pkg/sensors/handler.go
```diff
@@ -72,23 +72,32 @@ func SensorsFromPolicy(tp tracingpolicy.TracingPolicy, filterID policyfilter.Pol
 
 // revive:enable:exported
 
-func (h *handler) addTracingPolicy(op *tracingPolicyAdd) error {
-	h.collections.mu.Lock()
-	defer h.collections.mu.Unlock()
+// newCollection registers a collection for op, allocating a policy ID for it.
+// Should be called with h.collections.mu locked (for writing).
+func (h *handler) newCollection(op *tracingPolicyAdd) (*collection, error) {
```
mtardy: I don't exactly get the slicing and the rename, could you maybe put it in a preparatory commit before explaining in the git log the needs?

mtardy (commented):
Thanks, I think it makes sense to not make them disappear from listing. Could you please just slice this commit for reviews. I think it's fine to bundle changes together but maybe 700+ lines in the same commit is a lot and there are some changes that are kinda unrelated to the addition that could be separated. I didn't finish the full review.

I'm fine if you don't want to separate tests and changes but if you can have:
- commit 1 unrelated changes
- commit 2 related changes to pave the way
- commit 3 changes

or something like that

I feel like it would be easier to understand and review. Also imho I would try to go less verbose on comments, even if it's good having some of them.

pkg/sensors/collection.go
mtardy: or do you think it's better to have a generic name that would fit future use case in which we choose to skip a TP?

pkg/metrics/policymetrics/policymetrics.go
sayboras: Good point :100:

pkg/sensors/handler.go
sayboras: Good point :100:

pkg/sensors/collection.go
sayboras: TBH, I can't think of any other reason/selector to skip policy. Seems like we have everything covered. 

NodeSkippedState is the best I can think of right now.

pkg/metrics/policymetrics/policymetrics_test.go
sayboras: good point :100:

pkg/sensors/handler.go
```diff
@@ -72,23 +72,32 @@ func SensorsFromPolicy(tp tracingpolicy.TracingPolicy, filterID policyfilter.Pol
...
-	tpID := h.allocPolicyID()
 
-	col := collection{
+	col := &collection{
 		name:            op.ck.name,
 		tracingpolicy:   op.tp,
-		tracingpolicyID: uint64(tpID),
+		tracingpolicyID: h.allocPolicyID(),
+	}
+	collections[op.ck] = col
```
FedeDP: I feel like this line should be done by the caller; i wouldn't expect a method called `newCollection` to also register it in our collection map.

pkg/sensors/handler.go
```diff
@@ -141,6 +141,21 @@ func (h *handler) addTracingPolicy(op *tracingPolicyAdd) error {
...
 }
 
+// addSkippedTracingPolicy tracks a policy without loading it on this node. The
+// collection carries SkippedState so the policy is reported (e.g. by
+// ListTracingPolicies) as skipped rather than being invisible; no sensors or
+// policyfilter state are created.
+func (h *handler) addSkippedTracingPolicy(op *tracingPolicyAdd) error {
+	h.collections.mu.Lock()
+	defer h.collections.mu.Unlock()
+	col, err := h.newCollection(op)
```
FedeDP: Oh in the next commit i got why `newCollection` also adds it to the register.
Perhaps we could rename `newCollection` to something more explicit, like eg: `registerNewCollection()`?

pkg/sensors/collection.go
kkourt: >  it's better to have a generic name that would fit future use case in which we choose to skip a TP?

Above would be my preference. Maybe we never get another use-case, but if we do I'd rather have a single field and if we want more info we can add a string reason somewhere.

docs/content/en/docs/concepts/tracing-policy/k8s-filtering.md
```diff
@@ -99,6 +99,11 @@ spec:
 When a node is relabeled at runtime, each agent re-evaluates its policies and
 loads or unloads them accordingly.
 
+A policy that a node does not load because its `nodeSelector` does not match is
+not silently dropped: it is reported with the `skipped` state (rather than being
+absent) in `tetra tracingpolicy list`, so a gated-out policy is distinguishable
+from a missing one.
```
kkourt, line 105: ❤️

pkg/sensors/handler.go
```diff
@@ -72,23 +72,32 @@ func SensorsFromPolicy(tp tracingpolicy.TracingPolicy, filterID policyfilter.Pol
...
-func (h *handler) addTracingPolicy(op *tracingPolicyAdd) error {
-	h.collections.mu.Lock()
-	defer h.collections.mu.Unlock()
+// newCollection registers a collection for op, allocating a policy ID for it.
+// Should be called with h.collections.mu locked (for writing).
+func (h *handler) newCollection(op *tracingPolicyAdd) (*collection, error) {
 	collections := h.collections.c
 	// allow overriding existing policy collection that resulted in an error
 	// during the loading state
 	if col, exists := collections[op.ck]; exists && col.state != LoadErrorState {
```
kkourt: The previous code fails to add a policy with the same name, unless the previous policy was in error.
With the introduction of the changes in this PR, should we also include the case were the policy was skipped? That is, if a policy is skipped and a user tried to add it again, we don't add an error.

Note that this path is also used for loading policies via gRPC.

I don't have a strong opinion either way, but I think it's worth discussing it.

kkourt (commented):
Thanks!

pkg/sensors/handler.go
sayboras: :thinking: good point, I think we can allow overriding an existing policy if it's skipped

pkg/watcher/crdwatcher/tracingpolicy_reconciler.go
```diff
@@ -28,6 +28,7 @@ import (
 // Defined where it is consumed so the Reconciler can be unit-tested with a fake.
 type sensorManager interface {
 	AddTracingPolicy(ctx context.Context, tp tracingpolicy.TracingPolicy) error
+	AddSkippedTracingPolicy(ctx context.Context, tp tracingpolicy.TracingPolicy) error
```
mtardy, line 31: thinking here: you could have implemented this on the existing "AddTracingPolicy" with a "skipped" boolean flag. Maybe it would have been less moving/new code around. Not sure it would make a big difference, just realized when seeing this new method on the interface that is very similar to the existing one. Just a design choice I imagine.

mtardy (approved):
I think it looks good overall, thanks for the slicing it's way easier to review now. I was thinking about maybe a part of this could have been a new boolean flag on `AddTracingPolicy` instead of these new functions, not sure it makes a big diff in the end.

My last nit maybe for next time is that the e2e changes could have been sliced as well, the preparation for using the state and then the actual change for the new one. Just for making reviewers' job easier. Now no need to re-send something. Thanks!!

FedeDP (approved):
LGTM! Thanks

kkourt (approved):

PR #5258, opened 2026-07-10
new: enum like flags

FedeDP:
### Description

This PR introduces support for enum-like flags setting by introducing new `option.Enum` and `option.SliceEnum` structures that implement the `pflag.Value` interface.
It helped us clean up a little bit a couple of tetra commands.
Also, this is the cobra-ish way of doing that :)

Example output:
```
$ ./tetra tracingpolicy list --output foo
Error: invalid argument "foo" for "-o, --output" flag: invalid argument jsonn, please provide one of (text, json)

$ ./tetra getevents --event-types foo
Error: invalid argument "foo" for "-e, --event-types" flag: invalid argument foo, please provide one of (PROCESS_EXEC, PROCESS_EXIT, PROCESS_KPROBE, PROCESS_LOADER, PROCESS_LSM, PROCESS_THROTTLE, PROCESS_TRACEPOINT, PROCESS_UPROBE, PROCESS_USDT, RATE_LIMIT_INFO, TEST, UNDEF)
```

### DIFF
--- a/cmd/tetra/getevents/getevents.go
+++ b/cmd/tetra/getevents/getevents.go
@@ -10,7 +10,6 @@ import (
 	"io"
 	"os"
 	"sort"
-	"strings"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -18,16 +17,18 @@ import (
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/fieldmaskpb"
 
+	"github.com/cilium/tetragon/pkg/option"
+
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/cmd/tetra/common"
 	"github.com/cilium/tetragon/pkg/encoder"
 )
 
 type Opts struct {
-	Output         string
-	Color          string
+	Output         *option.Enum
+	Color          *option.Enum
 	IncludeFields  []string
-	EventTypes     []string
+	EventTypes     *option.SliceEnum
 	ExcludeFields  []string
 	Namespaces     []string
 	Namespace      []string // deprecated: use Namespaces
@@ -86,10 +87,10 @@ var GetFilter = func() *tetragon.Filter {
 	}
 
 	// Is used to filter on the event types i.e. PROCESS_EXEC, PROCESS_EXIT etc.
-	if len(Options.EventTypes) > 0 {
+	if len(Options.EventTypes.Values) > 0 {
 		var eventType tetragon.EventType
 
-		for _, v := range Options.EventTypes {
+		for _, v := range Options.EventTypes.Values {
 			eventType = tetragon.EventType(tetragon.EventType_value[v])
 			filter.EventSet = append(filter.EventSet, eventType)
 		}
@@ -140,7 +141,7 @@ func getEvents(ctx context.Context, client tetragon.FineGuidanceSensorsClient) e
 	if err != nil {
 		return fmt.Errorf("failed to call GetEvents: %w", err)
 	}
-	eventEncoder := GetEncoder(os.Stdout, encoder.ColorMode(Options.Color), Options.Timestamps, Options.Output == "compact", Options.TTYEncode, Options.StackTraces, Options.ImaHash)
+	eventEncoder := GetEncoder(os.Stdout, encoder.ColorMode(Options.Color.Value), Options.Timestamps, Options.Output.Value == "compact", Options.TTYEncode, Options.StackTraces, Options.ImaHash)
 	for {
 		res, err := stream.Recv()
 		if err != nil {
@@ -174,24 +175,6 @@ redirection of events to the stdin. Examples:
   # Include only process and parent.pod fields
   tetra getevents -f process,parent.pod`,
 		PreRunE: func(_ *cobra.Command, _ []string) error {
-			if Options.Output != "json" && Options.Output != "compact" {
-				return fmt.Errorf("invalid value for %q flag: %s", common.KeyOutput, Options.Output)
-			}
-			if Options.Color != "auto" && Options.Color != "always" && Options.Color != "never" {
-				return fmt.Errorf("invalid value for %q flag: %s", "color", Options.Color)
-			}
-
-			for _, v := range Options.EventTypes {
-				if _, found := tetragon.EventType_value[v]; !found {
-					supported := make([]string, 0, len(tetragon.EventType_value))
-					for name := range tetragon.EventType_value {
-						supported = append(supported, name)
-					}
-					sort.Strings(supported)
-					return fmt.Errorf("invalid value for %q flag: %s. Supported are %s", "event-types", v, strings.Join(supported, ", "))
-				}
-			}
-
 			// merge deprecated to new flags, appending since order does not matter
 			Options.Namespaces = append(Options.Namespace, Options.Namespaces...)
 			Options.Pods = append(Options.Pod, Options.Pods...)
@@ -234,11 +217,21 @@ redirection of events to the stdin. Examples:
 		},
 	}
 
+	// Prepare enum-like flags
+	Options.Output, _ = option.NewEnum([]string{"json", "compact"}, "json")
+	Options.Color, _ = option.NewEnum([]string{"auto", "always", "never"}, "auto")
+	supported := make([]string, 0, len(tetragon.EventType_value))
+	for name := range tetragon.EventType_value {
+		supported = append(supported, name)
+	}
+	sort.Strings(supported)
+	Options.EventTypes, _ = option.NewSliceEnum(supported, nil)
+
 	flags := cmd.Flags()
-	flags.StringVarP(&Options.Output, common.KeyOutput, "o", "json", "Output format. json or compact")
-	flags.StringVar(&Options.Color, "color", "auto", "Colorize compact output. auto, always, or never")
+	flags.VarP(Options.Output, common.KeyOutput, "o", "Output format "+Options.Output.Allowed())
+	flags.Var(Options.Color, "color", "Colorize compact output "+Options.Color.Allowed())
 	flags.StringSliceVarP(&Options.IncludeFields, "include-fields", "f", nil, "Include only fields in events")
-	flags.StringSliceVarP(&Options.EventTypes, "event-types", "e", nil, "Include only events of given types")
+	flags.VarP(Options.EventTypes, "event-types", "e", "Include only events of given types")
 	flags.StringSliceVarP(&Options.ExcludeFields, "exclude-fields", "F", nil, "Exclude fields from events")
 
 	flags.StringSliceVarP(&Options.Namespace, "namespace", "n", nil, "Get events by Kubernetes namespace")
--- a/pkg/option/enum.go
+++ b/pkg/option/enum.go
@@ -1,0 +1,90 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package option
+
+import (
+	"fmt"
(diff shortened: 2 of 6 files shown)


pkg/option/enum.go
```diff
@@ -0,0 +1,83 @@
...
+	allowed := make([]string, len(e.allowed))
+	for i, v := range e.allowed {
+		allowed[i] = fmt.Sprintf("%v", v)
+	}
+	return fmt.Sprintf("(%s)", strings.Join(allowed, ", "))
+}
+
+func (e *Enum[T]) Set(p T) error {
+	if !slices.Contains(e.allowed, p) {
+		return fmt.Errorf("invalid argument %v, please provide one of %s", p, e.Allowed())
```
FedeDP: Note: for this line, and for the same line for `SliceEnum`, govet is giving a warning:
>  pkg/option/enum.go:39:21      govet     inline: cannot inline: type parameter inference is not yet supported
>  pkg/option/enum.go:75:21      govet     inline: cannot inline: type parameter inference is not yet supported
 
This is a known limitation of the go compiler as of today, that is targeted to be fixed by go1.27: https://github.com/golang/go/issues/68236

Since it has no real functional impact, and our usage scope is pretty limited, i think we can ignore it.

pkg/option/enum_test.go
```diff
@@ -0,0 +1,54 @@
...
+
+func TestEnum(t *testing.T) {
+	allowed := []string{"foo", "bar", "baz"}
+	e := NewEnum(allowed, "foo")
+
+	if e.Value != "foo" {
+		t.Fatalf("expected default value %q, got %q", "foo", e.Value)
+	}
+
+	if err := e.Set("bar"); err != nil {
```
andrewstrohman: should use a variable with the value "bar", instead of repeating the same string literal?

pkg/option/enum_test.go
```diff
@@ -0,0 +1,54 @@
...
+	e := NewSliceEnum(allowed, []string{"foo"})
+
+	if len(e.Values) != 1 || e.Values[0] != "foo" {
+		t.Fatalf("expected default values [foo], got %v", e.Values)
+	}
+
+	if err := e.Set("bar"); err != nil {
+		t.Fatalf("unexpected error setting valid value: %v", err)
+	}
+	if len(e.Values) != 2 || e.Values[1] != "bar" {
```
andrewstrohman: should we also check that e.Values[0] has remained "foo" after the `Set`?

andrewstrohman (approved):

pkg/option/enum_test.go
FedeDP: Makes sense, let me fix it :)

FedeDP:
Uh i noticed that the `Value.Value` interface **requires** a `Set(string)` method. Therefore the whole templated logic is not going to work.
I am going to fix up the code for that.

See:
```
// Value is the interface to the dynamic value stored in a flag.
// (The default value is represented as a string.)
type Value interface {
	String() string
	Set(string) error
	Type() string
}
```

FedeDP:
Fixed + added many more tests.

pkg/option/enum.go
```diff
@@ -0,0 +1,88 @@
...
+	_, err := fmt.Sscanf(p, "%v", &v)
+	if err != nil {
+		return err
+	}
+	if !slices.Contains(e.allowed, v) {
+		return fmt.Errorf("invalid argument %v, please provide one of %s", v, e.Allowed())
+	}
+	e.Value = v
+	return nil
+}
```
mtardy: Mmh why do you need Sscanf here? is it only strings anyway?
```suggestion
func (e *Enum[T]) Set(p string) error {
	for _, v := range e.allowed {
		if fmt.Sprint(v) == p {
			e.Value = v
			return nil
		}
	}

	return fmt.Errorf("invalid argument %q, please provide one of %s", p, e.Allowed())
}
```

pkg/option/enum.go
```diff
@@ -0,0 +1,88 @@
...
+	return fmt.Sprintf("%v", e.Value)
+}
+
+func (e *Enum[T]) Allowed() string {
+	allowed := make([]string, len(e.allowed))
+	for i, v := range e.allowed {
+		allowed[i] = fmt.Sprintf("%v", v)
+	}
+	return fmt.Sprintf("(%s)", strings.Join(allowed, ", "))
+}
```
mtardy: you have duplicate code between Allowed() for Enum and SliceEnum

pkg/option/enum.go
```diff
@@ -0,0 +1,88 @@
...
+	"fmt"
+	"reflect"
+	"slices"
+	"strings"
+)
+
+type Enum[T comparable] struct {
+	allowed []T
+	Value   T
+}
```
mtardy: do we really need to support `T comparable` and not just strings?

pkg/option/enum.go
```diff
@@ -0,0 +1,88 @@
...
+
+func NewEnum[T comparable](allowed []T, d T) *Enum[T] {
+	return &Enum[T]{
+		allowed: allowed,
+		Value:   d,
+	}
+}
```
mtardy: should you validate that d belongs to allowed here?

pkg/option/enum.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (e *SliceEnum[T]) Allowed() string {
+	allowed := make([]string, len(e.allowed))
+	for i, v := range e.allowed {
+		allowed[i] = fmt.Sprintf("%v", v)
+	}
+	return fmt.Sprintf("(%s)", strings.Join(allowed, ", "))
+}
+
+func (e *SliceEnum[T]) Set(p T) error {
```
mtardy: here you take `T` but in the Enum version you take a `string`, is it expected?

cmd/tetra/getevents/getevents.go
```diff
@@ -234,11 +217,21 @@ redirection of events to the stdin. Examples:
...
+		supported = append(supported, name)
+	}
+	sort.Strings(supported)
+	Options.EventTypes = option.NewSliceEnum(supported, nil)
+
 	flags := cmd.Flags()
-	flags.StringVarP(&Options.Output, common.KeyOutput, "o", "json", "Output format. json or compact")
-	flags.StringVar(&Options.Color, "color", "auto", "Colorize compact output. auto, always, or never")
+	flags.VarP(Options.Output, common.KeyOutput, "o", "Output format. json or compact")
+	flags.Var(Options.Color, "color", "Colorize compact output. auto, always, or never")
```
mtardy: should you use allowed in a sprintf here instead of hardcoding the allowed in the help?

mtardy (commented):
nice addition, I think it's useful but maybe we can simplify it a little bit?

pkg/option/enum.go
FedeDP: I think it might be useful to support any kind of enum in the future (for example, loglevel as number instead of string).
We do not have any requirement for it right now but since it's not that complex, i think it's good to have.

pkg/option/enum.go
FedeDP: right!

pkg/option/enum.go
FedeDP: Yep i need to fix the SliceEnum impl :) thanks for catching it!

pkg/option/enum.go
FedeDP: Thanks! Nice simplification

FedeDP: Oh the problem with this is that:
```
t.Run("float enum pflag", func(t *testing.T) {
		os.Args = []string{"test", "--foo", "2.1"}
		testEnumPflag(t, []float64{1.1, 2.1, 3.1}, 1.1, 2.1)
	})

```
this passes fine but this one does not:
```
t.Run("float enum pflag", func(t *testing.T) {
		os.Args = []string{"test", "--foo", "2.0"}
		testEnumPflag(t, []float64{1.0, 2.0, 3.0}, 1.0, 2.0)
	})
```
That's because `fmt.Sprint(v)` returns `1,2,3` instead of `1.0,2.0,3.0` in this scenario.

pkg/option/enum.go
mtardy: I'm still not sure to get the benefit of having enum that are anything other than strings for flag? If it was loglevel then we would use the enum name like "debug"/"info" not the underlying int nop?

Regarding your comment here https://github.com/cilium/tetragon/pull/5258#discussion_r3570019043, if it was float64, we wouldn't have an enum but would use float64. I really wonder how to justify having a float64 enum.

maybe push it as a string thing, simplify most of the code and extend it if one day we find any other types that would be useful as enum, but I thought the point of enum is to use string to represent numbers?

pkg/option/enum.go
mtardy: see https://github.com/cilium/tetragon/pull/5258#discussion_r3586517545

mtardy (commented):
think it looks good but still find it hard to justify any other types than strings for enums, and that would simplify the code.

pkg/option/enum.go
FedeDP: It's not a big deal to me, we can simplify now and fix it later; but i find it simple enough that i would not worry about "premature optimization is root of all evil" thing :) 
In this case i prefer an "oh we already supported that, handy!" approach whenever we'll need it.

Let me ask @kkourt what does he thing! (thanks for the precious feedback btw!)

kkourt: If we have just strings right now, I would opt to leave `T comparable` for a later PR (whenever we need it).

FedeDP: 2 vs 1 :) 
Let's drop the generics :laughing:

FedeDP:
@mtardy after a discussion with @kkourt i dropped the generics and just implemented string flags as you suggested! Thanks for the discussion :)

pkg/option/enum.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	return &e, nil
+}
+
+func (e *Enum) String() string {
+	return e.Value
+}
+
+func (e *Enum) Allowed() string {
+	return fmt.Sprintf("(%s)", strings.Join(e.allowed, ", "))
```
mtardy, line 37: nit: I see that you added the () here, maybe the caller could add these if it's needed and here just return strings.Join as it's also used within your errors in this files and the parenthesis might just be there for the CLI help.

mtardy (approved):
lgtm! thanks

pkg/option/enum.go
FedeDP: Right, makes sense! Doing it now!
EDIT: oh, actually it's always used with parenthesis after all, i don't think it is worth changing it! Unless we'd want this:
```
./tetra getevents --event-types HEH
Error: invalid argument "HEH" for "-e, --event-types" flag: invalid argument HEH, please provide one of (PROCESS_EXEC, PROCESS_EXIT, PROCESS_KPROBE, PROCESS_LOADER, PROCESS_LSM, PROCESS_THROTTLE, PROCESS_TRACEPOINT, PROCESS_UPROBE, PROCESS_USDT, RATE_LIMIT_INFO, TEST, UNDEF)
```
to become
```
./tetra getevents --event-types HEH
Error: invalid argument "HEH" for "-e, --event-types" flag: invalid argument HEH, please provide one of PROCESS_EXEC, PROCESS_EXIT, PROCESS_KPROBE, PROCESS_LOADER, PROCESS_LSM, PROCESS_THROTTLE, PROCESS_TRACEPOINT, PROCESS_UPROBE, PROCESS_USDT, RATE_LIMIT_INFO, TEST, UNDEF
```
but it's better like it is today IMO.

PR #5271, opened 2026-07-10
fix(cmd/tetra): reject invalid explain output formats

immanuwell:
### Description
`tetra explain` says `-o` only accepts `json|yaml`, but typos still slipped through and the command quietly fell back to the default text output

This fix makes bad `-o` values fail fast, and adds a small regression test.

Repro:
`go run ./cmd/tetra explain tracingpolicy -o bogus`

Before:
exit `0`, regular text output

After:
exit `1`, `Error: invalid value for "output" flag: bogus`

### Changelog
```release-note
Fix `tetra explain` so invalid `--output` values fail instead of silently falling back to text output.
```

### DIFF
--- a/cmd/tetra/explain/explain.go
+++ b/cmd/tetra/explain/explain.go
@@ -63,6 +63,9 @@ Examples:
 			}
 			return cobra.ExactArgs(1)(cmd, args)
 		},
+		PreRunE: func(_ *cobra.Command, _ []string) error {
+			return validateOutputFormat(outputFormat)
+		},
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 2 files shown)


immanuwell:
@FedeDP WDYT?

FedeDP:
Ehy @immanuwell thanks for this!
Last friday, i opened https://github.com/cilium/tetragon/pull/5258 to introduce an `Enum` like pflag; i just pushed a commit to use it for `tetra explain --output` too :)
Can you give it a look? It should supersed this PR.

immanuwell:
fun coincidence) so I'm going to close this PR

PR #5279, opened 2026-07-13
Add support for gRPC-based persistent policies

tpapagian:
This PR adds persistence for tracing policies installed through the gRPC API, including policies managed with `tetra tracingpolicy`.

Persisted policies are restored when Tetragon restarts, including their state (enabled or disabled) and configured policy mode. This allows policy enforcement to remain active across agent restarts when combined with the existing pinned BPF handoff.

Example configuration:

```bash
tetragon [...] --keep-sensors-on-exit --release-pinned-bpf=false --persist-grpc-policies --persist-grpc-policies-dir /var/run/tetragon/grpc-policies
```

Details on the command line arguments:

1. `--keep-sensors-on-exit` keeps the existing BPF programs active when Tetragon stops
2. `--release-pinned-bpf=false` prevents the previous BPF programs from being removed when Tetragon starts
3. `--persist-grpc-policies` enables persistence and restoration of gRPC-based policies
4. `--persist-grpc-policies-dir` configures the directory containing the persistent policies

This PR adds support for `--persist-grpc-policies` and `--persist-grpc-policies-dir`. The remaining arguments already existed in Tetragon.

During a Tetragon restart, the following steps take place:
1. When Tetragon stops, `/sys/fs/bpf/tetragon` is not deleted and for this reason enforcement is still active even if the agent is not running. This happens because of `--keep-sensors-on-exit`.
2. When Tetragon re-starts, the existing `/sys/fs/bpf/tetragon` directory is renamed to `/sys/fs/bpf/tetragon_old`. Enforcement remains active. This happens because of `--release-pinned-bpf=false`. 
3. Tetragon re-creates a new `/sys/fs/bpf/tetragon` and restores the policies from the persistent store. This happens because of `--persist-grpc-policies`.
4. After policy loading completes, `/sys/fs/bpf/tetragon_old` is removed.

There is a short interval during which both the old and new programs are active. This overlap is intentional and avoids an enforcement gap during the restart.

More details can be found on the commit messages.

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -25,6 +25,7 @@ import (
 	"github.com/cilium/tetragon/api/v1/tetragon"
 	"github.com/cilium/tetragon/pkg/policystore"
 	"github.com/cilium/tetragon/pkg/rthooks"
+	"github.com/cilium/tetragon/pkg/sensors"
 	"github.com/cilium/tetragon/pkg/server/eventlog"
 
 	"github.com/cilium/tetragon/pkg/bpf"
@@ -226,6 +227,77 @@ func openGRPCPolicyStore() (*policystore.Store, error) {
 		"policies", len(store.List()))
 
 	return store, nil
+}
+
+type persistedTracingPolicy struct {
+	policy  tracingpolicy.TracingPolicy
+	enabled bool
+}
+
+func restoreGRPCPolicies(ctx context.Context, store *policystore.Store, manager *sensors.Manager) error {
+	if store == nil {
+		return nil
+	}
+
+	// Validate all records before loading any of them. This avoids partially
+	// restoring a store when a later record is malformed.
+	records := store.List()
+	log.Info("Starting persisted gRPC policy restoration",
+		"policies", len(records))
+	policies := make([]persistedTracingPolicy, 0, len(records))
+	for _, entry := range records {
+		policy, err := tracingpolicy.FromYAML(entry.Pol.YAML)
+		if err != nil {
+			return fmt.Errorf("restore persisted gRPC policy %s: parse YAML: %w", entry.ID.Name, err)
+		}
+		if policy.TpName() != entry.ID.Name || policy.TpNamespace() != entry.ID.Namespace {
+			return fmt.Errorf(
+				"restore persisted gRPC policy %s: record identity does not match YAML identity %s/%s",
+				entry.ID.Name, policy.TpNamespace(), policy.TpName())
+		}
+
+		policies = append(policies, persistedTracingPolicy{
+			policy: &server.GRPCTracingPolicy{
+				TracingPolicy: policy,
+				Domain:        entry.ID.Domain,
+			},
+			enabled: entry.Pol.Enabled,
+		})
+	}
+
+	for i, persisted := range policies {
+		policy := persisted.policy
+		state := sensors.DisabledState
+		if persisted.enabled {
+			state = sensors.EnabledState
+		}
+		log.Info("Loading persisted gRPC policy",
+			"name", policy.TpName(),
+			"namespace", policy.TpNamespace(),
+			"domain", policy.TpDomain(),
+			"enabled", persisted.enabled)
+		if err := manager.AddTracingPolicyWithState(ctx, policy, state); err != nil {
+			// as we want all-or-nothing semantics a single policy load error has to
+			// delete all previously loaded policies
+			var rollbackErr error
+			for j := i - 1; j >= 0; j-- {
+				restored := policies[j].policy
+				if err := manager.DeleteTracingPolicy(ctx, restored.TpName(), restored.TpNamespace(), restored.TpDomain()); err != nil {
+					rollbackErr = errors.Join(rollbackErr, fmt.Errorf("roll back restored gRPC policy %s: %w", tracingpolicy.TpLongname(restored), err))
+				}
+			}
+			return errors.Join(fmt.Errorf("restore persisted gRPC policy %s: load: %w", tracingpolicy.TpLongname(policy), err), rollbackErr)
+		}
+		log.Info("Restored persisted gRPC policy",
+			"name", policy.TpName(),
+			"namespace", policy.TpNamespace(),
+			"domain", policy.TpDomain(),
+			"enabled", persisted.enabled)
+	}
+	log.Info("Completed persisted gRPC policy restoration",
+		"policies", len(policies))
+
+	return nil
 }
 
 func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready func()) error {
@@ -480,11 +552,14 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 	if err = loadInitialSensor(ctx); err != nil {
 		return err
 	}
-	observer.GetSensorManager().LogSensorsAndProbes(ctx)
 	defer func() {
 		observer.RemoveSensors(ctx)
 		os.Remove(observerDir)
 	}()
+	if err = restoreGRPCPolicies(ctx, grpcPolicyStore, observer.GetSensorManager()); err != nil {
+		return err
+	}
+	observer.GetSensorManager().LogSensorsAndProbes(ctx)
 
 	pm, err := tetragonGrpc.NewProcessManager(
 		ctx,
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/cilium/tetragon/pkg/policystore"
 	"github.com/cilium/tetragon/pkg/rthooks"
 	"github.com/cilium/tetragon/pkg/server/eventlog"
 
@@ -210,6 +211,23 @@ func tetragonExecute() error {
 	return tetragonExecuteCtx(ctx, cancel, func() {})
 }
 
+func openGRPCPolicyStore() (*policystore.Store, error) {
+	if !option.Config.PersistGRPCPolicies {
+		return nil, nil
+	}
+
+	store, err := policystore.OpenAndLoad(option.Config.PersistGRPCPoliciesDir)
+	if err != nil {
+		return nil, fmt.Errorf("open persistent policy store %q: %w", option.Config.PersistGRPCPoliciesDir, err)
+	}
+
+	log.Info("Opened persistent policy store",
+		"directory", option.Config.PersistGRPCPoliciesDir,
+		"policies", len(store.List()))
+
+	return store, nil
+}
+
 func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready func()) error {
 	sigs := make(chan os.Signal, 1)
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
@@ -223,6 +241,11 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 		logger.Fatal(log, fmt.Sprintf("Failed path specified by --tracing-policy-dir '%q' is not absolute", option.Config.TracingPolicyDir))
 	}
 	option.Config.TracingPolicyDir = filepath.Clean(option.Config.TracingPolicyDir)
+
+	grpcPolicyStore, err := openGRPCPolicyStore()
+	if err != nil {
+		return err
+	}
 
 	if option.Config.RBSize != 0 && option.Config.RBSizeTotal != 0 {
 		logger.Fatal(log, "Can't specify --rb-size and --rb-size-total together")
@@ -467,7 +490,8 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 		ctx,
 		&cleanupWg,
 		observer.GetSensorManager(),
-		hookRunner)
+		hookRunner,
+		grpcPolicyStore)
 	if err != nil {
 		return err
 	}
--- a/docs/content/en/docs/concepts/enforcement/persistent-grpc-policies.md
+++ b/docs/content/en/docs/concepts/enforcement/persistent-grpc-policies.md
@@ -1,0 +1,138 @@
+---
+title: "Persistent gRPC policies"
+weight: 2
+description: "Persist and restore tracing policies installed through gRPC"
+---
+
+Tetragon can persist tracing policies installed through the gRPC API and
+restore their desired state after the agent restarts. Combined with persistent
+enforcement, this allows the old BPF programs to remain active while Tetragon
+loads replacement programs from the stored policies.
+
+This feature applies only to policies installed through gRPC, including
+policies added with `tetra tracingpolicy add`. Kubernetes and statically loaded
+policies continue to be managed by their respective policy sources.
+
+## Configuration
+
+Start Tetragon with the following options:
+
+```shell
+sudo ./tetragon \
+  --btf /sys/kernel/btf/vmlinux \
+  --bpf-lib ./bpf/objs/ \
+  --keep-sensors-on-exit \
+  --release-pinned-bpf=false \
+  --persist-grpc-policies \
+  --persist-grpc-policies-dir /var/run/tetragon/grpc-policies
+```
+
+Use the same options each time Tetragon starts.
+
+{{< caution >}}
+`--release-pinned-bpf` defaults to `true`. Set it explicitly to `false` when
+continuous enforcement across an agent restart is required. Enabling only the
+policy store restores policies, but does not preserve the previous BPF tree
+during startup.
+{{< /caution >}}
+
+The options have the following roles:
+
+- `--keep-sensors-on-exit` leaves pinned sensors active when Tetragon exits.
+- `--release-pinned-bpf=false` prevents Tetragon from removing the previous BPF
+  tree at the beginning of startup. Tetragon moves the tree aside and removes
+  it only after replacement policies have loaded.
+- `--persist-grpc-policies` enables saving and restoring policies installed
+  through gRPC.
+- `--persist-grpc-policies-dir` selects the policy store directory. Its default
+  value is `/var/run/tetragon/grpc-policies`.
+
+## Restart sequence
+
+On a normal restart, Tetragon performs the following handoff:
+
+1. Rename `/sys/fs/bpf/tetragon` to `/sys/fs/bpf/tetragon_old`. The links pinned
+   below this directory keep the previous BPF programs attached.
+2. Create a new `/sys/fs/bpf/tetragon` tree and load the initial sensors.
+3. Read and validate all records in the gRPC policy store.
+4. Recreate each stored policy in the new BPF tree. Tetragon registers disabled
+   policies without loading their BPF programs.
+5. Remove `/sys/fs/bpf/tetragon_old`, which detaches the previous programs.
+
+The policies are rebuilt from the persistent store; Tetragon does not reload
+them from `tetragon_old`. The old BPF tree exists only to maintain coverage
+during the handoff.
+
+{{< note >}}
+There is an interval during startup in which both the old and replacement
+programs are attached. This avoids an enforcement gap.
+{{< /note >}}
+
+## Manage persisted policies
+
+Policies added through gRPC are persisted automatically after they load
+successfully:
+
+```shell
+sudo ./tetra tracingpolicy add policy.yaml
+```
+
+Deleting a policy removes it from both the running agent and the persistent
+store:
+
+```shell
+sudo ./tetra tracingpolicy delete policy-name
+```
+
+Enable and disable operations update the stored desired state:
+
+```shell
+sudo ./tetra tracingpolicy disable policy-name
+sudo ./tetra tracingpolicy enable policy-name
+```
+
+Mode changes also update the stored policy YAML:
+
+```shell
+sudo ./tetra tracingpolicy set-mode policy-name monitor
+sudo ./tetra tracingpolicy set-mode policy-name enforce
+```
+
+## Stored desired state
+
+The store uses one JSON record per policy. A policy is uniquely identified by
+its domain, namespace, and name.
+
+Each record contains:
+
+- `yaml`: the complete policy, including its configured policy mode;
+- `enabled`: whether the policy is enabled or not.
+
+The policy identity is encoded in the record filename as
+`<name>:<namespace>:<domain>.json`.
+
+For example:
+
+```json
+{
+  "yaml": "apiVersion: cilium.io/v1alpha1\nkind: TracingPolicy\nmetadata:\n  name: lsm-file-open\nspec:\n  options:\n  - name: policy-mode\n    value: monitor\n  lsmhooks:\n  - hook: file_open\n",
+  "enabled": true
+}
+```
+
+## Failure behavior
+
+The running policy and its stored desired state are updated together:
+
+- If adding a new record to the store fails, Tetragon removes the newly loaded
+  runtime policy and restores any previous store state.
+- Enable, disable, and mode changes write the desired state before changing the
+  running policy. If the runtime change fails, Tetragon restores the previous
+  record.
+- Deleting a policy removes its record before deleting the running policy. If
+  the runtime deletion fails, Tetragon restores the record.
+
+At startup, Tetragon validates every stored policy before loading any of them.
+If loading a policy fails, it rolls back policies restored earlier in the same
+startup attempt and reports the error. The old BPF tree is removed only after
+startup policy loading succeeds.
--- a/pkg/policystore/format.go
+++ b/pkg/policystore/format.go
@@ -1,0 +1,36 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package policystore
+
+import (
+	"fmt"
+)
+
+type PolicyID struct {
+	Name      string // policy name
+	Namespace string // can be empty
+	Domain    string // i.e. grpc, static, k8s
+}
+
+// returns the string representation of the policy ID
+func (id PolicyID) String() string {
+	return fmt.Sprintf("%s:%s:%s", id.Name, id.Namespace, id.Domain)
+}
+
+type PolicyWithState struct {
+	YAML    string `json:"yaml"`
+	Enabled bool   `json:"enabled"`
+}
+
+type PolicyEntry struct {
+	ID  PolicyID
+	Pol PolicyWithState
+}
+
+type PolicyStore interface {
+	Get(id PolicyID) (PolicyWithState, bool)
+	Put(id PolicyID, state PolicyWithState) error
+	Delete(id PolicyID) error
+	List() []PolicyEntry
+}
--- a/pkg/policystore/store.go
+++ b/pkg/policystore/store.go
@@ -1,0 +1,252 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package policystore
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/logger/logfields"
+)
+
+const (
+	recordFileSuffix     = ".json"
+	recordTempFileSuffix = ".json.tmp"
+	recordDirMode        = 0o700
+	recordFileMode       = 0o600
+)
+
+var errDstRemoved = errors.New("destination removed")
+
+type Store struct {
+	dir     string
+	mu      sync.Mutex
+	records map[PolicyID]PolicyWithState
+}
+
+// gets aaa:bbb:ccc.json and returns (aaa, bbb, ccc)
+func decodeFilename(name string) (string, string, string, error) {
+	base := strings.TrimSuffix(name, recordFileSuffix)
+	parts := strings.SplitN(base, ":", 3)
+	if len(parts) != 3 {
+		return "", "", "", fmt.Errorf("invalid filename %s: expected format aaa:bbb:ccc.json", name)
+	}
+
+	return parts[0], parts[1], parts[2], nil
+}
+
+func OpenAndLoad(dir string) (*Store, error) {
+	if strings.TrimSpace(dir) == "" {
+		return nil, errors.New("policy store directory must not be empty")
+	}
+
+	dir = filepath.Clean(dir)
+	if err := os.MkdirAll(dir, recordDirMode); err != nil {
+		return nil, fmt.Errorf("create policy store directory %s: %w", dir, err)
+	}
+
+	info, err := os.Stat(dir)
+	if err != nil {
+		return nil, fmt.Errorf("stat policy store directory %s: %w", dir, err)
+	}
+	if !info.IsDir() {
+		return nil, fmt.Errorf("policy store path %s is not a directory", dir)
+	}
+
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return nil, fmt.Errorf("read policy store directory %s: %w", dir, err)
+	}
+
+	s := &Store{
+		dir:     dir,
+		records: make(map[PolicyID]PolicyWithState),
+	}
+
+	for _, entry := range entries {
+		name := entry.Name()
+		if strings.HasSuffix(name, recordTempFileSuffix) || !strings.HasSuffix(name, recordFileSuffix) {
+			logger.GetLogger().Warn("unrecognized file in policy store",
+				"filename", name)
+			continue
+		}
+
+		info, err := entry.Info()
+		if err != nil {
+			logger.GetLogger().Warn("stat policy record failed",
+				logfields.Error, err,
+				"path", filepath.Join(dir, name))
+			continue
+		}
+		if !info.Mode().IsRegular() {
+			logger.GetLogger().Warn("policy record is not a regular file",
+				"path", filepath.Join(dir, name))
+			continue
+		}
+
+		path := filepath.Join(dir, name)
+		data, err := os.ReadFile(path)
+		if err != nil {
+			return nil, fmt.Errorf("read policy record %s: %w", path, err)
+		}
+
+		var pol PolicyWithState
+		if err := json.Unmarshal(data, &pol); err != nil {
+			return nil, fmt.Errorf("decode policy record %s: %w", path, err)
+		}
+
+		polName, polNamespae, polDomain, polErr := decodeFilename(name)
+		if polErr != nil {
+			return nil, fmt.Errorf("policy record %s has unexpected filename", path)
+		}
+		id := PolicyID{
+			Name:      polName,
+			Namespace: polNamespae,
+			Domain:    polDomain,
+		}
+		if _, exists := s.records[id]; exists {
+			return nil, fmt.Errorf("duplicate policy record for %s", id)
+		}
+
+		s.records[id] = pol
+	}
+
+	return s, nil
+}
+
+func (s *Store) List() []PolicyEntry {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	entries := make([]PolicyEntry, 0, len(s.records))
+	for id, pol := range s.records {
+		entries = append(entries, PolicyEntry{ID: id, Pol: pol})
+	}
+	return entries
+}
+
+func (s *Store) Get(id PolicyID) (PolicyWithState, bool) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	state, ok := s.records[id]
+	return state, ok
+}
+
+func (s *Store) Put(id PolicyID, state PolicyWithState) error {
+	data, err := json.MarshalIndent(state, "", "  ")
+	if err != nil {
+		return fmt.Errorf("encode policy record for %s: %w", id, err)
+	}
+	data = append(data, '\n')
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if err := writeAtomic(s.dir, id.String(), data); err != nil {
+		if errors.Is(err, errDstRemoved) {
+			delete(s.records, id)
+		}
+		return fmt.Errorf("write policy record for %s: %w", id, err)
+	}
+	s.records[id] = state
+	return nil
+}
+
+func (s *Store) Delete(id PolicyID) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	path := filepath.Join(s.dir, id.String()+recordFileSuffix)
+	err := os.Remove(path)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("delete policy record for %s: %w", id, err)
+	}
+
+	delete(s.records, id)
+	if err == nil {
+		if err := syncDir(s.dir); err != nil {
+			return fmt.Errorf("sync policy store after deleting %s: %w", id, err)
+		}
+	}
+	return nil
+}
+
+func chmodWriteSync(f *os.File, data []byte) error {
+	defer f.Close()
+	if err := f.Chmod(recordFileMode); err != nil {
+		return err
+	}
+	if _, err := f.Write(data); err != nil {
+		return err
+	}
+	if err := f.Sync(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// writeAtomic attempts to write data atomicaly to a file
+// if an error happens, the caller should try to put the previous
+// record if anything exists
+func writeAtomic(dir, name string, data []byte) (retErr error) {
+	tmp, err := os.CreateTemp(dir, name+".*"+recordTempFileSuffix)
+	if err != nil {
+		retErr = fmt.Errorf("failed to create tempfile in dir %q for name %q: %w", dir, name, err)
+		return
+	}
+
+	tmpName := tmp.Name()
+	fileToRemove := tmpName
+	defer func() {
+		if retErr == nil {
+			return
+		}
+		if err := os.Remove(fileToRemove); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("failed to remove %q: %w", fileToRemove, err))
+		} else {
+			if tmpName != fileToRemove {
+				retErr = errors.Join(retErr, errDstRemoved)
+			}
+		}
+		// make an attempt to sync the dir
+		syncDir(dir)
+	}()
+
+	if err := chmodWriteSync(tmp, data); err != nil {
+		retErr = fmt.Errorf("failed to write data to tempfile %q: %w", tmpName, err)
+		return
+	}
+
+	dst := filepath.Join(dir, name+recordFileSuffix)
+	if err := os.Rename(tmpName, dst); err != nil {
+		retErr = fmt.Errorf("failed rename tempfile %q to %q: %w", tmpName, dst, err)
+		return
+	}
+
+	// now the proper file exists, we need to sync the directory to make this fully atomic in
+	// the rare occasion that something bad happens, we remove the file
+	fileToRemove = dst
+	if err := syncDir(dir); err != nil {
+		retErr = fmt.Errorf("failed to sync directory: %w", err)
+		return
+	}
+
+	return
+}
+
+func syncDir(dir string) error {
+	f, err := os.Open(dir)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+	return f.Sync()
+}
--- a/pkg/sensors/manager.go
+++ b/pkg/sensors/manager.go
@@ -132,23 +132,25 @@ func newTracingPolicyAdd(ctx context.Context, tp tracingpolicy.TracingPolicy) (*
 
 // AddTracingPolicy adds a new sensor based on a tracing policy
 func (h *Manager) AddTracingPolicy(ctx context.Context, tp tracingpolicy.TracingPolicy) error {
-	op, err := newTracingPolicyAdd(ctx, tp)
-	if err != nil {
-		return err
-	}
-
-	return h.handler.addTracingPolicy(op)
+	return h.AddTracingPolicyWithState(ctx, tp, EnabledState)
 }
 
-// AddSkippedTracingPolicy tracks tp as skipped, without loading any BPF, so
-// that it is reported instead of being absent on this node.
-func (h *Manager) AddSkippedTracingPolicy(ctx context.Context, tp tracingpolicy.TracingPolicy) error {
+// AddTracingPolicyWithState adds a new sensor based on a tracing policy and
+// controls its initial state. A disabled policy remains registered and can
+// subsequently be enabled with EnableTracingPolicy. A skipped policy is
+// tracked without creating sensors or policy filter state.
+func (h *Manager) AddTracingPolicyWithState(ctx context.Context, tp tracingpolicy.TracingPolicy, state TracingPolicyState) error {
+	if state != EnabledState && state != DisabledState && state != SkippedState {
+		return fmt.Errorf("invalid initial tracing policy state: %v", state)
+	}
+
 	op, err := newTracingPolicyAdd(ctx, tp)
 	if err != nil {
 		return err
 	}
 
-	return h.handler.addSkippedTracingPolicy(op)
+	op.state = state
+	return h.handler.addTracingPolicy(op)
 }
 
 // DeleteTracingPolicy deletes a new sensor based on a tracing policy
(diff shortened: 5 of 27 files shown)


pkg/policystore/store.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	s := &Store{
+		dir:     dir,
+		records: make(map[Key]Record),
+	}
+
+	for _, entry := range entries {
+		name := entry.Name()
+		if isTemporaryRecordFile(name) || !strings.HasSuffix(name, recordFileSuffix) {
```
kkourt: This seems unnecessarily complicated to me, how about we have:

```
	recordFileSuffix = ".json"
    recordTempFileSuffix = ".json.tmp"
```

And just check the suffix?

pkg/policystore/store.go
```diff
@@ -0,0 +1,195 @@
...
+			continue
+		}
+
+		info, err := entry.Info()
+		if err != nil {
+			return nil, fmt.Errorf("stat policy record %q: %w", filepath.Join(dir, name), err)
+		}
+		if !info.Mode().IsRegular() {
+			return nil, fmt.Errorf("policy record %q is not a regular file", filepath.Join(dir, name))
+		}
```
kkourt, line 92: I think the behavior we want here is to continue iterating all the directory entries, even if we encounter an error on one.

pkg/policystore/format.go
```diff
@@ -0,0 +1,45 @@
...
+	hash := sha256.New()
+	hash.Write([]byte(k.Domain))
+	hash.Write([]byte{0})
+	hash.Write([]byte(k.Namespace))
+	hash.Write([]byte{0})
+	hash.Write([]byte(k.Name))
+	return hex.EncodeToString(hash.Sum(nil))
+}
+
+type Record struct {
```
kkourt: non-blocking nit: the `Record` name feels too generic to me.
Feel free to leave it as is if  you prefer it though.

pkg/policystore/store.go
```diff
@@ -0,0 +1,195 @@
...
+	if err == nil {
+		if err := syncDir(s.dir); err != nil {
+			return fmt.Errorf("sync policy store after deleting %s: %w", key.ToString(), err)
+		}
+	}
+	return nil
+}
+
+func recordFileName(key Key) string {
+	return fmt.Sprintf("%s%s", key.ToHash(), recordFileSuffix)
```
kkourt: What's the reason for hashing here?
I feel like using something like '<domain>:<namespace>:<name>` will make the contents of the directory better to understand, and we will, also, not have to store the key in the record (although that's a minor thing).

pkg/policystore/store.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+func isTemporaryRecordFile(name string) bool {
+	return strings.HasPrefix(name, ".") && strings.Contains(name, ".tmp-")
+}
+
+func writeAtomic(dir, name string, data []byte) (bool, error) {
+	tmp, err := os.CreateTemp(dir, "."+name+".tmp-")
```
kkourt: Just noting that initially I thought that we could create the tempfile in `/tmp` or something, so to avoid ignoring temp files, but then the rename would not be atomic so we need the tempfile and final file to be in the same dir.

pkg/policystore/store.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	data = append(data, '\n')
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	committed, err := writeAtomic(s.dir, recordFileName(record.Id), data)
+	if committed {
+		s.records[record.Id] = record
+	}
```
kkourt: What is the case where `comitted == true` but `err != nil`?

pkg/policystore/store.go
```diff
@@ -0,0 +1,195 @@
...
+
+	s := &Store{
+		dir:     dir,
+		records: make(map[Key]Record),
+	}
+
+	for _, entry := range entries {
+		name := entry.Name()
+		if isTemporaryRecordFile(name) || !strings.HasSuffix(name, recordFileSuffix) {
+			continue
```
kkourt, line 78: Might worth adding a logging message here?

cmd/tetragon/main.go
```diff
@@ -224,6 +237,16 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
...
 	option.Config.TracingPolicyDir = filepath.Clean(option.Config.TracingPolicyDir)
 
+	grpcPolicyStore, err := openGRPCPolicyStore()
+	if err != nil {
+		return err
+	}
+	if grpcPolicyStore != nil {
+		log.Info("Opened persistent policy store",
+			"directory", option.Config.PersistGRPCPoliciesDir,
+			"policies", len(grpcPolicyStore.List()))
```
kkourt: nit: Any reason for not having this inside `openGRPCPolicyStore`?

I feel like the less things we have in `tetragonEecuteCtx` the better :)

pkg/server/server.go
```diff
@@ -243,6 +247,101 @@ func (gtp *GRPCTracingPolicy) TpDomain() string {
...
+	}
+	return fmt.Errorf("restore persisted state for policy %s: %w", key.Name, err)
+}
+
+func wrapPolicyRuntimeRollbackError(key policystore.Key, err error) error {
+	if err == nil {
+		return nil
+	}
+	return fmt.Errorf("roll back runtime policy %s: %w", key.Name, err)
+}
```
kkourt: nit: how about

```go
func wrapError(err error, fmt string, args ...any) error {
    if err == nil {
        return nil;
    }
   return fmt.Errorf("%s: %w", fmt.Sprintf(fmt, args), err)
}
```

instead of the two helpers above?

pkg/server/server.go
```diff
@@ -258,6 +357,17 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
 	if req.GetDomain() != "" {
 		gtp.Domain = req.GetDomain()
 	}
+	key := grpcPolicyKey(tp.TpName(), tp.TpNamespace(), gtp.TpDomain())
+
+	s.policyMu.Lock()
+	defer s.policyMu.Unlock()
```
kkourt: Why are we locking here?

pkg/server/server.go
```diff
@@ -258,6 +357,17 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
...
+
+	// check if a policy with the same name-namespace-domain already exists
+	var previous policystore.Record
+	var previousExisted bool
+	if s.policyStore != nil {
+		previous, previousExisted = s.policyStore.Get(key)
+	}
```
kkourt: Can't we do this after we add the policy, in the `s.policyStore != nil` below?

pkg/server/server.go
```diff
@@ -266,6 +376,26 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
...
+		record := policystore.Record{
+			Id:      key,
+			YAML:    req.GetYaml(),
+			Enabled: true,
+		}
+		if err := s.policyStore.Put(record); err != nil {
+			// A failed write may still have changed the store.
+			// Restore the record that existed before this add request.
+			// A previous record may exist when retrying a policy in LoadErrorState.
+			storeRollbackErr := s.restorePolicyRecord(key, previous, previousExisted)
```
kkourt: Not sure I understand, why do we need to restore the record here?

kkourt (commented):
Thanks!

Please find some first comments below.

pkg/server/server.go
tpapagian: Both `s.observer.AddTracingPolicy` and `s.policyStore.Put(record)` are thread-safe on their own. But we also need to make everything inside each function atomic.

Let's say that we have 2 concurrent requests for the same policy. One will disable that and one will set mode = monitor. Both of these have 3 steps: get the current policy from the store, update the runtime, update the store. We don't want those to be interleaved because this can lead to incosidtent state in the runtime and in the store.

cmd/tetragon/main.go
tpapagian: I have moved that inside `openGRPCPolicyStore`.

pkg/server/server.go
tpapagian: Make sense, I have used wrapError instead of these.

pkg/server/server.go
tpapagian: Yes, I have moved that in the `s.policyStore != nil` below.

pkg/policystore/store.go
tpapagian: With `./tetra tracingpolicy add ./pol1.yaml --domain something` users can add a policy with arbitrary domain name. This can possibly lead to malformed filenames. On the other hand, `tetra tracingpolicy add` is a privileged command so this is possibly an overkill. I will remove that.

pkg/server/server.go
tpapagian: If a policy with the same `<name,namespace,domain>` exists, then `s.observer.AddTracingPolicy` will fail and this code will never run.

There is one case, where the policy is in `load_error` state where we can re-add a policy with the same `<name,namespace,domain>`.

Example:
```yaml
kind: TracingPolicy
metadata:
  name: "lsm-file-open"
spec:
  lsmhooks:
  - hook: "file_open_missing"
    args:
      - index: 0
        type: "file"
```

```console
$ sudo ./tetra tracingpolicy add ./pol1.yaml
Error: failed to add tracing policy: rpc error: code = Unknown desc = policy handler 'tracing' failed loading policy 'lsm-file-open': lsm validation failed: error in spec.lsmHooks[0]: lsm hook "file_open_missing" not found in BTF (looked up "bpf_lsm_file_open_missing"): not found, please check if the hook exists or if your kernel supports BTF for lsm hooks, as an alternative, consider switching to kprobes
```

and 

```console
$ sudo ./tetra tracingpolicy list
ID   NAME            DOMAIN   STATE        FILTERID   NAMESPACE   SENSORS   KERNELMEMORY   MODE      NPOST   NENFORCE   NMONITOR
2    lsm-file-open   grpc     load_error   0          (global)              0 B            unknown   0       0          0
```

then we fix the policy (i.e. rename file_open_missing to file_open).

```console
$ sudo ./tetra tracingpolicy add ./pol1.yaml
tracing policy "./pol1.yaml" added
```

and 

```console
$ sudo ./tetra tracingpolicy list
ID   NAME            DOMAIN   STATE     FILTERID   NAMESPACE   SENSORS       KERNELMEMORY   MODE      NPOST   NENFORCE   NMONITOR
3    lsm-file-open   grpc     enabled   0          (global)    generic_lsm   1.12 MB        enforce   0       0          0
```

If the `s.policyStore.Put` for the corrected policy fails, we need to restore in the store the old policy with the load_error and call s.observer.DeleteTracingPolicy on the corrected policy. There is no real reason to re-apply the wrong policy. But on an agent restart, we will still have the wrong policy loaded.

pkg/policystore/store.go
tpapagian: This the case where the rename succeeds and the directory sync fails. Currently we update the in-memory representation of the store and return an error. 

I believe that this is handled appropriately in the caller side. More specifically in AddTracingPolicy we get the previous record before the Put. In the case where Put fails (either with committed equals to true or false) we delete the new entry and call DeleteTracingPolicy on the applied policy. We do something similar in EnableTracingPolicy, ConfigureTracingPolicy, and DisableTracingPolicy.

There is a case where Put fails and the restore fails. Not sure what will happen in that case, but in order to make that case work as expected we possibly need a more complex store and I am not sure if it worth the effort. 

If a rename succeeds and a sync fails, I am not sure that Tetragon will manage to restart in that node..

pkg/policystore/store.go
tpapagian: Yes, a rename between different filesystem mounts is not atomic.

pkg/policystore/store.go
tpapagian: Done.

pkg/policystore/store.go
tpapagian: Changed that. We now print a warning for each of these cases and continue.

pkg/policystore/format.go
tpapagian: Renamed to PolicyState.

pkg/policystore/store.go
tpapagian: Changed that to use "name:namespace:domain.json" naming. I have also removed the key from the record.

pkg/policystore/store.go
tpapagian: Done

pkg/sensors/manager.go
mtardy, line 154: so we just added this new AddSkippedTracingPolicy. I suggested adding a boolean on the interface to say if should be added skipped or not (but we went with the new AddSkipped instead of reusing Add and adding a skipped boolean). And now we add another boolean (as opposed to a new method). Maybe we should reuse the other one or align on whether if we should use new methods or new args. Enabled/Skipped is fairly similar maybe we can do something here?

pkg/policystore/format.go
```diff
@@ -0,0 +1,36 @@
...
+package policystore
+
+import (
+	"fmt"
+)
+
+type PolicyID struct {
+	Name      string // policy name
+	Namespace string // can be empty
+	Domain    string // i.e. grpc, static, k8s
```
mtardy, line 13: out of this scope: but I wish we use an enum here internally even if the API exposes a string.

pkg/policystore/format.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// returns the string representation of the policy ID
+func (id PolicyID) ToString() string {
```
mtardy: maybe use `String()` so that it's called automatically as the stringer on this struct when you print with "%s", https://go.dev/tour/methods/17 and you can remove the manual call.

pkg/policystore/store.go
```diff
@@ -0,0 +1,216 @@
...
+func (s *Store) Get(id PolicyID) (PolicyState, bool) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	state, ok := s.records[id]
+	return state, ok
+}
+
+func (s *Store) Put(id PolicyID, state PolicyState) error {
+	data, err := json.MarshalIndent(state, "", "  ")
```
mtardy, line 144: do we care about it being human readable with indent?

pkg/policystore/store.go
```diff
@@ -0,0 +1,216 @@
...
+func writeAtomic(dir, name string, data []byte) (bool, error) {
+	tmp, err := os.CreateTemp(dir, name+".*"+recordTempFileSuffix)
+	if err != nil {
+		return false, err
+	}
+	tmpName := tmp.Name()
+	defer os.Remove(tmpName)
+
+	if err := tmp.Chmod(recordFileMode); err != nil {
+		tmp.Close()
```
mtardy: any chance we could wrap these close into the defer? maybe skipping it we reach close?

pkg/policystore/store.go
```diff
@@ -0,0 +1,216 @@
...
+		return fmt.Errorf("write policy record for %s: %w", id.ToString(), err)
+	}
+	return nil
+}
+
+func (s *Store) Delete(id PolicyID) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	path := filepath.Join(s.dir, fmt.Sprintf("%s%s", id.ToString(), recordFileSuffix))
```
mtardy: path := filepath.Join(s.dir, id.ToString() + recordFileSuffix)

pkg/policystore/format.go
```diff
@@ -0,0 +1,36 @@
...
+	return fmt.Sprintf("%s:%s:%s", id.Name, id.Namespace, id.Domain)
+}
+
+type PolicyState struct {
```
mtardy: i'd rename this maybe because:
- initially I thought it was going to be the global "state" because of the name
- it's containing the state (enable or not, btw should you have more details, not just bool, here as the state of a policy can be also skipped etc?) but also the YAML

pkg/policystore/store.go
```diff
@@ -0,0 +1,216 @@
...
+	base := strings.TrimSuffix(name, recordFileSuffix)
+	parts := strings.SplitN(base, ":", 3)
+	if len(parts) != 3 {
+		return "", "", "", fmt.Errorf("invalid filename %s: expected format aaa:bbb:ccc.json", name)
+	}
+
+	return parts[0], parts[1], parts[2], nil
+}
+
+func Open(dir string) (*Store, error) {
```
mtardy: this is doing more than Opening the dir here: parsing the content etc, maybe change the name?

pkg/policystore/format.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"fmt"
+)
+
+type PolicyID struct {
+	Name      string // policy name
+	Namespace string // can be empty
+	Domain    string // i.e. grpc, static, k8s
+}
```
mtardy, line 14: doesn't this triplet representing a policy already exist in the codebase?

cmd/tetragon/main.go
```diff
@@ -210,6 +211,23 @@ func tetragonExecute() error {
...
+	}
+
+	store, err := policystore.Open(option.Config.PersistGRPCPoliciesDir)
+	if err != nil {
+		return nil, fmt.Errorf("open persistent policy store %q: %w", option.Config.PersistGRPCPoliciesDir, err)
+	}
+
+	log.Info("Opened persistent policy store",
+		"directory", option.Config.PersistGRPCPoliciesDir,
+		"policies", len(store.List()))
```
mtardy, line 227: nit: I'd log from the caller and not from the function. 

EDIT: lol just saw https://github.com/cilium/tetragon/pull/5279#discussion_r3595153779

pkg/server/server.go
```diff
@@ -243,6 +247,94 @@ func (gtp *GRPCTracingPolicy) TpDomain() string {
...
+func grpcPolicyID(name, namespace, domain string) policystore.PolicyID {
+	if domain == "" {
+		domain = GrpcDomain
+	}
+	return policystore.PolicyID{
+		Domain:    domain,
+		Namespace: namespace,
+		Name:      name,
+	}
+}
```
mtardy: I feel this should belong to policystore pkg nop?

pkg/server/server.go
```diff
@@ -310,8 +465,18 @@ func (s *Server) ConfigureTracingPolicy(ctx context.Context, req *tetragon.Confi
...
-	if err := s.observer.ConfigureTracingPolicy(ctx, req); err != nil {
+	id := grpcPolicyID(req.GetName(), req.GetNamespace(), req.GetDomain())
+
+	s.policyMu.Lock()
+	defer s.policyMu.Unlock()
+
+	if err := s.updatePersistedPolicy(id,
+		func(state *policystore.PolicyState) error {
+			return updateStateFromConfigure(state, req)
+		},
```
mtardy: nit: I find the code structure complicated to follow here, having a function taking a function as argument that just return directly from another function.

pkg/server/server.go
```diff
@@ -243,6 +247,94 @@ func (gtp *GRPCTracingPolicy) TpDomain() string {
...
+
+// updatePersistedPolicy writes the desired state before applying the runtime
+// change. If the runtime change fails, the previous durable state is restored.
+// Policies that are not already in the store retain their existing behavior
+// and are not implicitly adopted into gRPC persistence.
+func (s *Server) updatePersistedPolicy(
+	id policystore.PolicyID,
+	update func(*policystore.PolicyState) error,
+	apply func() error,
+) error {
```
mtardy: I get the idea behind this function but I don't get why you call apply on all first error paths?

pkg/server/server.go
```diff
@@ -243,6 +247,94 @@ func (gtp *GRPCTracingPolicy) TpDomain() string {
...
+	}
+	return nil
+}
+
+func wrapError(err error, msg string, args ...any) error {
+	if err == nil {
+		return nil
+	}
+	return fmt.Errorf("%s: %w", fmt.Sprintf(msg, args...), err)
+}
```
mtardy: I guess it's personal but I think these kind of helpers make the code flow more complicated to read and are not really helpful. See https://github.com/cilium/tetragon/pull/5279/changes/b27e256409e977c7145f5fed6630dc9717935ca1#r3603569163

pkg/server/server.go
```diff
@@ -243,6 +247,94 @@ func (gtp *GRPCTracingPolicy) TpDomain() string {
...
+	next := previous
+	if err := update(&next); err != nil {
+		return err
+	}
+	if err := s.policyStore.Put(id, next); err != nil {
+		restoreErr := s.policyStore.Put(id, previous)
+		return errors.Join(
+			fmt.Errorf("persist desired state for policy %s: %w", id.Name, err),
+			wrapError(restoreErr, "restore persisted state for policy %s", id.Name),
+		)
```
mtardy: maybe it feels better to write but can you directly inline the code in those place so it's easier to understand quickly how you want to propagate the errors? Tbh when I read this it's quite unclear to me.

pkg/server/server.go
```diff
@@ -243,6 +247,94 @@ func (gtp *GRPCTracingPolicy) TpDomain() string {
...
+
+	var mode string
+	switch req.GetMode() {
+	case tetragon.TracingPolicyMode_TP_MODE_ENFORCE:
+		mode = "enforce"
+	case tetragon.TracingPolicyMode_TP_MODE_MONITOR:
+		mode = "monitor"
+	default:
+		return fmt.Errorf("unexpected mode: %v", req.GetMode())
+	}
```
mtardy: I saw we have a similar construct (the opposite) somewhere else in the code https://github.com/cilium/tetragon/blob/b2d5916b100b8ba77ffc0a8ad52b4a0e1da3dc4b/cmd/tetra/tracingpolicy/tracingpolicy.go#L299-L308, maybe we should write those two functions and reuse them for this type to string?

cmd/tetragon/main.go
```diff
@@ -228,6 +229,73 @@ func openGRPCPolicyStore() (*policystore.Store, error) {
...
+		}
+
+		policies = append(policies, persistedTracingPolicy{
+			policy: &server.GRPCTracingPolicy{
+				TracingPolicy: policy,
+				Domain:        entry.ID.Domain,
+			},
+			enabled: entry.State.Enabled,
+		})
+	}
```
mtardy, line 266: maybe it's a dumb question: why doesn't your store already contain `persistedTracingPolicy` and you just manage the serialization in the back when writing/reading to disk? Does it need to expose YAML?

pkg/server/server_test.go
```diff
@@ -64,3 +65,36 @@ func TestSetDebug(t *testing.T) {
...
+	enabled := false
+	mode := tetragon.TracingPolicyMode_TP_MODE_MONITOR
+
+	err := updateStateFromConfigure(&state, &tetragon.ConfigureTracingPolicyRequest{
+		Enable: &enabled,
+		Mode:   &mode,
+	})
+	require.NoError(t, err)
+	require.False(t, state.Enabled)
+	require.YAMLEq(t, `apiVersion: cilium.io/v1alpha1
```
mtardy, line 104: oh wow didn't know about this

docs/content/en/docs/concepts/enforcement/persistent-grpc-policies.md
```diff
@@ -0,0 +1,136 @@
...
+4. Recreate each stored policy in the new BPF tree. Tetragon registers disabled
+   policies without loading their BPF programs.
+5. Remove `/sys/fs/bpf/tetragon_old`, which detaches the previous programs.
+
+The policies are rebuilt from the persistent store; Tetragon does not reload
+them from `tetragon_old`. The old BPF tree exists only to maintain coverage
+during the handoff.
+
+There is an interval during startup in which both the old and replacement
+programs are attached. This avoids an enforcement gap.
```
mtardy, line 68: nit: could put this in `note`

mtardy (commented):
Thanks this is useful, however I had troubles reviewing `b27e256409e977c7145f5fed6630dc9717935ca1` maybe I should just spend more time to understand it better (but I get that the rollback parts make things complicated). Here are some suggestions.

Also something I'm not a huge fan is the confusion we now get from this addition:
- we have mode: monitor or enforce
- we have status: https://deploy-preview-5279--tetragon.netlify.app/docs/reference/grpc-api/#tetragon-TracingPolicyState with the big enum
- we now have this new enabled or not boolean from the PolicyState.
- we have the AddSkipped and the Add + enabled: false/true.

It's a bit hard to follow what is what tbh now.

tpapagian:
Thanks @mtardy for the review! I will address all of you comments.

> Thanks this is useful, however I had troubles reviewing `b27e256409e977c7145f5fed6630dc9717935ca1` maybe I should just spend more time to understand it better (but I get that the rollback parts make things complicated). Here are some suggestions.

Yes, I agree that `b27e256409e977c7145f5fed6630dc9717935ca1` is a bit complicated. I will make a try to rewrite this patch in order to make it simpler.

Another thought that I had (based on the comments from @kkourt as well) is that we could add a dependency to a persistent key-value store that solves all of these issues regarding persistence internally and just use that here. It would simplify things but I don't believe that we need such a complicated dependency as this PR requires simple semantics without the need to store a large amount of data and/or have any performance requirements. If you believe that this is a good idea, I can make a try towards this direction. 

> 
> Also something I'm not a huge fan is the confusion we now get from this addition:
> 
> * we have mode: monitor or enforce
> * we have status: https://deploy-preview-5279--tetragon.netlify.app/docs/reference/grpc-api/#tetragon-TracingPolicyState with the big enum
> * we now have this new enabled or not boolean from the PolicyState.
> * we have the AddSkipped and the Add + enabled: false/true.
> 
> It's a bit hard to follow what is what tbh now.

I don't feel that this PR complicates things regarding those items. Even before that PR we have the ability to have policies enabled or disabled. We used to have only AddTracingPolicy which always adds a policy in enabled state. This PR only adds  AddTracingPolicyWithState (without too much code duplication) that provides the ability to add a tracing policy in disabled state (which can be enabled later by using the gRPC API). 

UPDATE: After reading https://github.com/cilium/tetragon/pull/5279#discussion_r3603331941 again I can see the issue here. `AddSkippedTracingPolicy` is introduced after the initial version of this PR so I implemented `AddTracingPolicyWithState` to solve the same issue. I will remove `AddTracingPolicyWithState` and reuse `AddSkippedTracingPolicy` when needed. With that in place, I don't believe that this PR complicates things regarding what you mentioned before.

pkg/sensors/manager.go
tpapagian: Merged AddTracingPolicyWithState and AddSkippedTracingPolicy into a single function that takes as argument the state (i.e. enabled, disabled, skipped).

pkg/policystore/format.go
tpapagian: Done

pkg/policystore/store.go
tpapagian: Done

pkg/policystore/store.go
tpapagian: Renamed that to OpenAndLoad

cmd/tetragon/main.go
tpapagian: Don't have strong opinions on that. If you want to change that back to the original version, just let me know.

pkg/server/server.go
tpapagian: Moved that to policystore

pkg/server/server.go
tpapagian: Created pkg/tracingpolicy/mode.go with these helpers and reuse the in both places.

docs/content/en/docs/concepts/enforcement/persistent-grpc-policies.md
tpapagian: Done

pkg/policystore/store.go
tpapagian: I would say no as this is not intended to be read or modified by users.

pkg/policystore/format.go
tpapagian: Renamed that to `PolicyWithState`.

> it's containing the state (enable or not, btw should you have more details, not just bool, here as the state of a policy can be also skipped etc?) but also the YAML

I would say that there is no need to change that. A skipped policy is because of a nodeSelector that does not match the local machine. So on agent restart we will try to apply the policy, the nodeSelector will not match, and the policy will end up again in skipped state.

pkg/policystore/store.go
tpapagian: Done

pkg/policystore/format.go
tpapagian: There is also 

https://github.com/cilium/tetragon/blob/f4c3713f50e7ab38e498607ae5df37b8c68492e6/pkg/sensors/collection.go#L62-L64

but I don't believe that it worth make that public and reuse that as it does not result in a large code repetition.

pkg/server/server.go
tpapagian: Removed wrapError and do all error handling inline.

tpapagian:
@mtardy Thanks again for the review! I believe that I have addressed all of your comments.

I have also made a try to simplify things a bit and also make some patches smaller.

pkg/policystore/store.go
kkourt: > I believe that this is handled appropriately in the caller side. More specifically in AddTracingPolicy we get the previous record before the Put. In the case where Put fails (either with committed equals to true or false) we delete the new entry and call DeleteTracingPolicy on the applied policy.

How about if there is an error `writeAtomic` we try to put the filesystem in its previous state. I think this simplifies the code.

For example:

```diff
diff --git a/pkg/policystore/store.go b/pkg/policystore/store.go
index 8d266b4e9..ede502b6b 100644
--- a/pkg/policystore/store.go
+++ b/pkg/policystore/store.go
@@ -148,13 +148,10 @@ func (s *Store) Put(id PolicyID, state PolicyWithState) error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	committed, err := writeAtomic(s.dir, id.String(), data)
-	if committed {
-		s.records[id] = state
-	}
-	if err != nil {
+	if err := writeAtomic(s.dir, id.String(), data); err != nil {
 		return fmt.Errorf("write policy record for %s: %w", id, err)
 	}
+	s.records[id] = state
 	return nil
 }
 
@@ -191,22 +188,50 @@ func chmodWriteSync(f *os.File, data []byte) error {
 	return nil
 }
 
-func writeAtomic(dir, name string, data []byte) (bool, error) {
+// writeAtomic attempts to write data atomicaly to a file
+// if an error happens, we try our best to leave the filesystem in its previous state
+func writeAtomic(dir, name string, data []byte) (retErr error) {
+
 	tmp, err := os.CreateTemp(dir, name+".*"+recordTempFileSuffix)
 	if err != nil {
-		return false, err
+		retErr = fmt.Errorf("failed to create tempfile in dir %q for name %q: %w", dir, name, err)
+		return
 	}
+
 	tmpName := tmp.Name()
-	defer os.Remove(tmpName)
+	fileToRemove := tmpName
+	defer func() {
+		if fileToRemove == "" {
+			return
+		}
+		if err := os.Remove(fileToRemove); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("failed to remove %q", fileToRemove))
+		}
+		// make an attempt to sync the dir
+		syncDir(dir)
+	}()
 
 	if err := chmodWriteSync(tmp, data); err != nil {
-		return false, err
+		retErr = fmt.Errorf("failed to write data to tempfile %q: %w", tmpName, err)
+		return
 	}
+
 	dst := filepath.Join(dir, name+recordFileSuffix)
 	if err := os.Rename(tmpName, dst); err != nil {
-		return false, err
+		retErr = fmt.Errorf("failed rename tempfile %q to %q: %w", tmpName, dst, err)
+		return
 	}
-	return true, syncDir(dir)
+
+	// now the proper file exists, we need to sync the directory to make this fully atomic in
+	// the rare occasion that something bad happens, we remove the file
+	fileToRemove = dst
+	if err := syncDir(dir); err != nil {
+		retErr = fmt.Errorf("failed to sync directory: %w", err)
+		return
+	}
+
+	fileToRemove = ""
+	return
 }
 
 func syncDir(dir string) error {
diff --git a/pkg/server/server.go b/pkg/server/server.go
index e9a410dac..9ebbea261 100644
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -290,8 +290,6 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
 			var storeRollbackErr error
 			if exists { // previous entry existed to put it back and this will override anything that already exists there
 				storeRollbackErr = s.policyStore.Put(id, previous)
-			} else { // nothing exists there before so just delete if anything exists now
-				storeRollbackErr = s.policyStore.Delete(id)
 			}
 			if storeRollbackErr != nil {
 				storeRollbackErr = fmt.Errorf("restore persisted state for policy %s: %w", id.Name, storeRollbackErr)
```

tpapagian: I believe that this make sense.

> we try our best to leave the filesystem in its previous state

This is not totally correct. In the case of a failure, we leave the store in either of the following states:
1. If the failure happens before or at the rename, we will keep the store as it was
2. If the failure is at the sync or the remove inside the defer, we will remove the entry entirely and the caller should put the previous entry again. If the remove is successful we should also call delete(s.records, id) in Put() when writeAtomic fails for the persistent and in-memory representation to be consistent.

In either cases, a Put for the previous entry from the caller side is enough to restore the previous entry. 

With some fixes, I ended up to the following diff:
```diff
diff --git a/pkg/policystore/store.go b/pkg/policystore/store.go
index 8d266b4e9..4924e5b49 100644
--- a/pkg/policystore/store.go
+++ b/pkg/policystore/store.go
@@ -23,6 +23,8 @@ const (
        recordFileMode       = 0o600
 )

+var errDstRemoved = errors.New("destination removed")
+
 type Store struct {
        dir     string
        mu      sync.Mutex
@@ -148,13 +150,13 @@ func (s *Store) Put(id PolicyID, state PolicyWithState) error {
        s.mu.Lock()
        defer s.mu.Unlock()

-       committed, err := writeAtomic(s.dir, id.String(), data)
-       if committed {
-               s.records[id] = state
-       }
-       if err != nil {
+       if err := writeAtomic(s.dir, id.String(), data); err != nil {
+               if errors.Is(err, errDstRemoved) {
+                       delete(s.records, id)
+               }
                return fmt.Errorf("write policy record for %s: %w", id, err)
        }
+       s.records[id] = state
        return nil
 }
@@ -191,22 +193,53 @@ func chmodWriteSync(f *os.File, data []byte) error {
        return nil
 }

-func writeAtomic(dir, name string, data []byte) (bool, error) {
+// writeAtomic attempts to write data atomicaly to a file
+// if an error happens, the caller should try to put the previous
+// record if anything exists
+func writeAtomic(dir, name string, data []byte) (retErr error) {
        tmp, err := os.CreateTemp(dir, name+".*"+recordTempFileSuffix)
        if err != nil {
-               return false, err
+               retErr = fmt.Errorf("failed to create tempfile in dir %q for name %q: %w", dir, name, err)
+               return
        }
+
        tmpName := tmp.Name()
-       defer os.Remove(tmpName)
+       fileToRemove := tmpName
+       defer func() {
+               if retErr == nil {
+                       return
+               }
+               if err := os.Remove(fileToRemove); err != nil {
+                       retErr = errors.Join(retErr, fmt.Errorf("failed to remove %q: %w", fileToRemove, err))
+               } else {
+                       if tmpName != fileToRemove {
+                               retErr = errors.Join(retErr, errDstRemoved)
+                       }
+               }
+               // make an attempt to sync the dir
+               syncDir(dir)
+       }()

        if err := chmodWriteSync(tmp, data); err != nil {
-               return false, err
+               retErr = fmt.Errorf("failed to write data to tempfile %q: %w", tmpName, err)
+               return
        }
+
        dst := filepath.Join(dir, name+recordFileSuffix)
        if err := os.Rename(tmpName, dst); err != nil {
-               return false, err
+               retErr = fmt.Errorf("failed rename tempfile %q to %q: %w", tmpName, dst, err)
+               return
        }
-       return true, syncDir(dir)
+
+       // now the proper file exists, we need to sync the directory to make this fully atomic in
+       // the rare occasion that something bad happens, we remove the file
+       fileToRemove = dst
+       if err := syncDir(dir); err != nil {
+               retErr = fmt.Errorf("failed to sync directory: %w", err)
+               return
+       }
+
+       return
 }

 func syncDir(dir string) error {
diff --git a/pkg/server/server.go b/pkg/server/server.go
index e9a410dac..9ebbea261 100644
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -290,8 +290,6 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
                        var storeRollbackErr error
                        if exists { // previous entry existed to put it back and this will override anything that already exists there
                                storeRollbackErr = s.policyStore.Put(id, previous)
-                       } else { // nothing exists there before so just delete if anything exists now
-                               storeRollbackErr = s.policyStore.Delete(id)
                        }
                        if storeRollbackErr != nil {
                                storeRollbackErr = fmt.Errorf("restore persisted state for policy %s: %w", id.Name, storeRollbackErr)
```

Let me know If this seems reasonable from you side and I will apply this to the PR as well.

tpapagian: @kkourt I have applied the diff in the PR so you can check that from there.

pkg/server/server.go
kkourt: > If the s.policyStore.Put for the corrected policy fails, we need to restore in the store the old policy with the load_error and call s.observer.DeleteTracingPolicy on the corrected policy

Does the policy with `load_error` exist in the store?

tpapagian: > Does the policy with load_error exist in the store?

After testing that again no. Possibly I can do some cleanup on this code. I will have a look.

UPDATE: I believe we can remove the following entirely.

```go
			// A failed write may still have changed the store.
			// Restore the record that existed before this add request.
			// A previous record may exist when retrying a policy in LoadErrorState.
			var storeRollbackErr error
			if exists { // previous entry existed to put it back and this will override anything that already exists there
				storeRollbackErr = s.policyStore.Put(id, previous)
			}
			if storeRollbackErr != nil {
				storeRollbackErr = fmt.Errorf("restore persisted state for policy %s: %w", id.Name, storeRollbackErr)
			}
```

I have updated the PR accordingly.

pkg/server/server.go
```diff
@@ -258,6 +262,10 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
 	if req.GetDomain() != "" {
 		gtp.Domain = req.GetDomain()
 	}
+	id := policystore.GrpcPolicyID(tp.TpName(), tp.TpNamespace(), gtp.TpDomain())
```
kkourt: One thing to note is that here, we persist policies that are not part of the gRPC domain. I think that's fine, but something to be aware of.

CC: @FedeDP who did the domain implementation.

pkg/policystore/format.go
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package policystore
+
+import (
+	"fmt"
+)
+
+const GrpcDomain = "grpc"
```
kkourt: This is also defined in server.go. I guess there will be an import cycle if we try to import it from there. Do we need it? See also below.

pkg/policystore/format.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// returns the string representation of the policy ID
+func (id PolicyID) String() string {
+	return fmt.Sprintf("%s:%s:%s", id.Name, id.Namespace, id.Domain)
+}
+
+func GrpcPolicyID(name, namespace, domain string) PolicyID {
+	if domain == "" {
+		domain = GrpcDomain
```
kkourt: is there a case where this is called with an empty domain?

pkg/server/server.go
```diff
@@ -258,6 +262,10 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
...
+
+	s.policyMu.Lock()
+	defer s.policyMu.Unlock()
```
kkourt: Can we move the the lockling and the `id := ` also under `if s.policyStore != nil`?

pkg/server/server.go
```diff
@@ -266,20 +274,79 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
...
 	if req.GetDomain() != "" {
 		domain = req.GetDomain()
 	}
+	id := policystore.GrpcPolicyID(req.GetName(), req.GetNamespace(), domain)
+
+	s.policyMu.Lock()
+	defer s.policyMu.Unlock()
+
+	var previous policystore.PolicyWithState
+	var exists bool
```
kkourt: same as before, can we move everything below if s.policyStore != nil?

pkg/server/server.go
```diff
@@ -266,20 +274,79 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
...
+				if restoreErr != nil {
+					restoreErr = fmt.Errorf("restore persisted state for policy %s: %w", id.Name, restoreErr)
+				}
+				return nil, errors.Join(
+					fmt.Errorf("delete persisted policy %s: %w", id.Name, err),
+					restoreErr,
+				)
+			}
+		}
+	}
```
kkourt, line 355: So in `AddTracingPolicy`, we first do the observer operation and, if successful, we then persist the policy in the store. If peristing the policy fails, we reverse the `AddTracingPolicy`.

Here, we do the reverse: we first remove the policy from the store, and then attempt to delete it from the observer.

Is there a reason for the reverse order of operations in Add vs Del?

I'm not sure what happens if we fail to delete a policy from the observer. Does the policy remain active?

tpapagian: > Is there a reason for the reverse order of operations in Add vs Del?

My thinking was to do the more "risky" part first. What I mean by that:

In Add, the AddTracingPolicy can fail for many reasons that can be related to the kernel, to the policy itself, to the runtime, etc. So I try to do the difficult thing first (i.e. add the policy in the runtime) and then update the store (which sounds to me that it has less chances to fail). If the runtime operation fail, then we have nothing to do.

In Delete, the DeleteTracingPolicy call will be likely successful in all cases. So try the more "risky" part first (update the store) and then run the less risky one. If I change the order in the Delete path, and do first the DeleteTracingPolicy (that succeeds), and then the store update (that fails), then we need to Add the policy back again to the runtime which sounds to me more complicated.

Same idea applies to Enable/Disable/Configure.

pkg/server/server.go
FedeDP: > One thing to note is that here, we persist policies that are not part of the gRPC domain. I think that's fine, but something to be aware of.

Mmmh why do you say that? `gtp.TpDomain()` can be `grpc` if the request has no domain field set.

pkg/policystore/format.go
tpapagian: After running some tests, it seems no. So I removed that. 

I initially used https://github.com/cilium/tetragon/blob/0f804fc285d41b492e94f9ed6fac5c4a496f8e7d/pkg/server/server.go#L239-L244 as a reference to do that.

pkg/policystore/format.go
tpapagian: Removed that because of https://github.com/cilium/tetragon/pull/5279#discussion_r3631278762

pkg/server/server.go
tpapagian: Yes, and I have updated that.

pkg/server/server.go
tpapagian: Depends on what we decide in https://github.com/cilium/tetragon/pull/5279#discussion_r3629894151. If we keep it as it is, we need `id`, `previous`, and `exists` in restorePreviousEntry which is called after `DeleteTracingPolicy`.

pkg/server/server.go
kkourt: Users can specify the domain in the gRPC call, no?

tpapagian: Seems yes:
```
$ sudo ./tetra tracingpolicy add ./pol1.yaml --domain "abc"
tracing policy "./pol1.yaml" added
$ sudo ./tetra tracingpolicy list
ID   NAME            DOMAIN   STATE     FILTERID   NAMESPACE   SENSORS       KERNELMEMORY   MODE      NPOST   NENFORCE   NMONITOR
2    lsm-file-open   abc    enabled   0          (global)    generic_lsm   1.12 MB        enforce   0       0          0
```
but the default (i.e. not defined) seems to be grpc.

FedeDP: Exactly :point_up:

kkourt: And would expect that `pol1.yaml` will be persisted, hence my comment:

> we persist policies that are not part of the gRPC domain. I think that's fine, but something to be aware of.

pkg/server/server.go
kkourt: OK, that's fine with me. Thanks for the explanation.

pkg/server/server.go
FedeDP: Understood what you meant; yes that's a correct observation; i think that's fine. Note however that we internally never expect free-domains to be actually used; the `tetra tracingpolicy add --domain` switch is mostly for debug purposes.

pkg/server/server.go
```diff
@@ -68,20 +69,23 @@ type Server struct {
 	notifier     Notifier
 	observer     observer
 	hookRunner   hookRunner
+	policyStore  *policystore.Store
+	policyMu     sync.Mutex
 	tetragon.UnimplementedFineGuidanceSensorsServer
```
kkourt, line 82: Can you please add a comment on why we need the lock? (see https://github.com/cilium/tetragon/pull/5279#discussion_r3595844944)

pkg/server/server.go
```diff
@@ -266,20 +270,84 @@ func (s *Server) AddTracingPolicy(ctx context.Context, req *tetragon.AddTracingP
...
+				)
+			}
+		}
+	}
 
 	if err := s.observer.DeleteTracingPolicy(ctx, req.GetName(), req.GetNamespace(), domain); err != nil {
+		restoreErr := restorePreviousEntry(s.policyStore, id, previous, exists)
 		logger.GetLogger().Warn("Server DeleteTracingPolicy request failed", "name", req.GetName(), logfields.Error, err)
-		return nil, err
+		return nil, errors.Join(err, restoreErr)
```
kkourt: non-blocking nit: 

IMO:
```go
if store != nil && exists {
        if restoreErr := store.Put(id, previous); restoreErr != nil {
                    err = errors.Join(err, fmt.Errorf("restore persisted state for policy %s: %w", id.Name, restoreErr))
        }
}
return nil, err
```

Makes the code more clear, compared to using the helper function.

cmd/tetragon/main.go
```diff
@@ -228,6 +229,77 @@ func openGRPCPolicyStore() (*policystore.Store, error) {
...
+			state = sensors.EnabledState
+		}
+		log.Info("Loading persisted gRPC policy",
+			"name", policy.TpName(),
+			"namespace", policy.TpNamespace(),
+			"domain", policy.TpDomain(),
+			"enabled", persisted.enabled)
+		if err := manager.AddTracingPolicyWithState(ctx, policy, state); err != nil {
+			// as we want all-or-nothing semantics a single policy load error has to
+			// delete all previously loaded policies
```
kkourt, line 281: Hm, do we? 

I was thinking that we would load all the policies we could, and let the user know about failing ones and ask them to inspect them on the file-system.

Not sure what the best approach is here.
@mtardy @FedeDP Thoughts?

kkourt (commented):
Thanks!

Please find some (more) comments below.

pkg/server/server.go
tpapagian: Done

pkg/server/server.go
tpapagian: Done

cmd/tetragon/main.go
tpapagian: These are correct policies that have been loaded successfully before. So the expected thing is that all of them will load without any issues. If any of them fails to load this is possibly a temporary issue with the agent or the OS.

If we keep the behaviour as it is now, in the case where a single policy fails to load, the agent will fail to start but the previous policies will be still there (providing enforcement). Then a retry to start the agent will possibly be successful and we avoid having any enforcement gaps.

If we change the behaviour to allow policies to fail to load (and report errors to the user), then the agent will start successfully and remove all previous policies. Then there will be a gap in enforcement for the policy (or policies) that we fail to load. In the case where a policy cannot load at all, the user can remove that from the store manually (after inspection) and retry to start the agent.

This is my reasoning to have all-or-nothing semantics here.

kkourt: > These are correct policies that have been loaded successfully before. So the expected thing is that all of them will load without any issues. If any of them fails to load this is possibly a temporary issue with the agent or the OS.

A case where this might happen is an agent being downgraded, and ending up with a policy that is not supported in the previous version.

> If we keep the behaviour as it is now, in the case where a single policy fails to load, the agent will fail to start 

Ah, so we fail the agent from starting? Sorry missed that part.

> but the previous policies will be still there (providing enforcement).

OK, that makes sense to me. Thanks.

kkourt (approved):
Thanks! LGTM!

mtardy (approved):
Reviewed quickly via phone but it looks like you addressed all the comments!! Lgtm :) thanks!

PR #5280, opened 2026-07-13
Uprobe binary digest verification

andrewstrohman:
In some cases, a uprobe policy is only intended to apply to a specific version of a binary, and could have adverse effect if attached to a different version of the binary.

For users that want to only apply the policy for a specific build, they can specify a digest type and value to validate the target binary. The following example policy demonstrates this functionality.

```
kind: TracingPolicy
metadata:
    name: "uprobe-digest-example"
spec:
    uprobes:
    - path: "/bin/example-target-binary"
      symbols:
      - "main"
      binaryDigests:
```

When the above policy is loaded, tetragon will calculate the sha256 hash of `/bin/example-target-binary` and compare it against all the binaryDigests configured within the `binaryDigests` list. If the calculated digest does not match any configured digest, the policy is rejected in its entirety.

As you can see from the above example, the format of the entries of the `binaryDigests` list is
`<digest type>:<digest>`.

The following digest types are supported:
- sha1
- sha256
- sha384
- sha512
- build-id (found in the ELF note section named `.note.gnu.build-id`)

When a uprobe is attached, the kernel keeps track of the target by its inode. It's best practice for installation software (such as rpm, deb, install, tar) to unlink a pre-existing target before installing a new version. This ensures that the new version of the binary will have a new inode, so the uprobe will not fire for it.

But, if the target binary is updated without an unlink, which would happen if `cp` is used or if the target is opened and updated in-place, the inode will be retained. In this case, the uprobe will continue firing for the updated binary, potentially causing unintended effect. As such, it's critical that users of this feature ensure their installation method will not retain the inode of the target binary that they want to update.

If the user wants the policy to apply to a new version of the binary post-installation, they need to reload the policy after installation in order to attach to the new version of the binary.

## Note to reviewers

See [this PR](https://github.com/cilium/tetragon/pull/4994) for the big picture of where this change is heading. If/when this PR is merged, we will reject policies which configured digests, and none of those configured digests match the actual digest of the binary. There will be a follow-up PR which will contain the additional commits from [here](https://github.com/cilium/tetragon/pull/4994) which are not present in this PR. This follow-up will change the behavior such that digest mismatch will result in partial load of a policy, instead of rejecting the policy in its entirety. Only the probes/hooks which failed digest verification are not loaded, and the rest of the hooks are loaded. That follow-up change will also make the hook status (loaded vs digest rejected) visible to the user via our tracing policy status API.

The benefit of having these two separate PRs is that it makes it easier for the reviewer to review (less cognitive load) and it allows me to make incremental progress.
On the other hand, the consequence of doing this is that there will be a period of time where the main branch does digest verification, but rejects policies upon verification failure. So, I'm also open to squashing my changes into a single PR so that we never reject policies due to digest mismatch, if we think that's a better way to go. Just let me know.


pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -405,6 +405,9 @@ type UProbeSpec struct {
 	// +kubebuilder:validation:Optional
 	// A return argument to include in the trace output.
 	ReturnArg *KProbeArg `json:"returnArg,omitempty"`
+	// +kubebuilder:validation:Optional
+	// Digests of the traced binary. If any digest matches, verification passes.
```
kkourt: I find "verification passes" somewhat ambiguous. 

How about:
```
// BinaryDigests specifies a set of digests for the traced binary.
// The uprobe is installed only the digest of the traced binary matches a digest in the set.
// Currently, if the digest is not matched, the policy is rejected. Subsequent work will skip loading the uprobe instead of rejecting the policy.
```

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -635,6 +638,112 @@ func cleanupUprobeEntries(ids []idtable.EntryID, openedFiles []*os.File) error {
...
+	if entryFile == nil {
+		return nil
+	}
+
+	digestCache := make(map[string]string)
+	matchFound := false
+	for _, digest := range uprobe.BinaryDigests {
+		if err := verifyFileDigest(entryFile, digest, digestCache); err == nil {
+			matchFound = true
+			break
```
kkourt: nit: can we just `return nil` here?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -698,8 +807,23 @@ func createGenericUprobeSensor(
...
-		ids, err = addUprobe(&uprobe, entryFile, ids, &in, &has)
-		if err != nil {
+		if len(uprobe.BinaryDigests) != 0 {
+			entryFile, err = os.Open(absPath)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		if err := verifyBinaryDigests(&uprobe, entryFile, cfgIdx); err != nil {
```
kkourt: At this point, if I understand correctly by looking at `verifyFileDigest`, `entryFile` might not point at the start of the file. Is this an issue for when we pass `entryFile` in `addUprobe`?

If it is, we could write `verifyFileDigest` so that it restores the file pointer to its previous location when done (using SEEK_SET with offset 0 to get the current location).

kkourt (commented):
LGTM, just a couple of comments. Thanks!

pkg/k8s/apis/cilium.io/v1alpha1/types.go
andrewstrohman: Yes, that's much more clear. Thanks.

I added that, but included the word "if" here:
> The uprobe is installed only **if** the digest

Because I think that you meant to write that.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Yes, that's better. Done.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: > At this point, if I understand correctly by looking at verifyFileDigest, entryFile might not point at the start of the file.

yes, I agree with this.

>  Is this an issue for when we pass entryFile in addUprobe?

I don't think so, despite the fact that the file is used in addUprobe. It is used for resolving the elf offsets, but elf library reads from explicit byte offsets, so it effectively ignores the current position of the file that is passed to it.

> If it is, we could write verifyFileDigest so that it restores the file pointer to its previous location when done (using SEEK_SET with offset 0 to get the current location).

I'm going to do this even though I don't think it's currently required. This seems like a good defensive measure for a future where the file offset would matter.

Prior to this change, verifyFileDigest did `file.Seek(0, 0)` in order to ensure that it would read from the beginning. I'm keeping that, and in addition, I'll remember the offset of the file when the function was entered (should be always be zero for now), and set that offset before returning.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -698,8 +813,23 @@ func createGenericUprobeSensor(
...
+		if len(uprobe.BinaryDigests) != 0 {
+			entryFile, err = getOrOpenFile(absPath, openedFilesMap)
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		if err := verifyBinaryDigests(&uprobe, entryFile, cfgIdx); err != nil {
+			if entryFile != nil {
+				if closeErr := entryFile.Close(); closeErr != nil && !errors.Is(closeErr, os.ErrClosed) {
```
olsajiri: I wonder could the entryfile be closed at this point ? (wrt os.ErrClosed check)

nit since it's error path I'm not sure how helpful it is to gather errors from close call,
maybe we could just close it.. but don't have strong feelings about that

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -635,6 +638,117 @@ func cleanupUprobeEntries(ids []idtable.EntryID, openedFiles []*os.File) error {
...
+	defer func() {
+		if _, err := file.Seek(currentOffset, io.SeekStart); err != nil {
+			retErr = errors.Join(retErr, fmt.Errorf("failed to restore file offset: %w", err))
+		}
+	}()
+
+	var calculatedHash string
+
+	if algo == "build-id" {
+		safeELF, err := elf.NewSafeELFFile(file)
```
olsajiri: IIUC we don't need to close safeELF right?

  ▎ Close(): If the File was created using NewFile directly instead of Open, Close has no effect.

could you put comment in here, I don't think it's obvious

olsajiri (approved):
thanks, looks good, left 2 small comments

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: > IIUC we don't need to close safeELF right?

Yes, we don't need to close it, and if we try to close it anyway there is no effect.

> could you put comment in here, I don't think it's obvious

Ok, done. This is the comment I added:
```
               // There is no need to call safeELF.Close() because that function only closes the underlying *os.File.
               // Further, safeELF.Close() won't attempt to close the underlying file when the SafeELFFile is created with
               // NewSafeELFFile().
               // We don't want to close the underlying file here because it needs to remain open for the uprobe to be
               // attached to its file descriptor.
```

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: > I wonder could the entryfile be closed at this point ? (wrt os.ErrClosed check)

Good catch. I don't think that it could be closed at this point. So, I removed the os.ErrClosed check. I also removed the os.ErrClosed check after the following call of `addUprobe` for the same reason.

> nit since it's error path I'm not sure how helpful it is to gather errors from close call,
maybe we could just close it.. but don't have strong feelings about that

The reason that I appended these errors is because otherwise, it's hidden that we expected to be able to close the file, but could not. I thought this might help us troubleshoot what's going wrong in case of  cascading failures. So, I've changed this to log the error instead appending to the error that's returned.  This way, we don't hide the issue, and also don't include/append the cascading error to the return value, because it's irrelevant to the caller.

mtardy:
> If the calculated digest does not match any configured digest, that uprobe is skipped and a warning is reported in policy status.

On this, I just thought on this new status that we just added that could be reused here https://github.com/cilium/tetragon/commit/e3b2e489565eb6f5b7c68e5eeb185c219c8ee8d4#diff-b51cb791919665a2013217e038932107da1e4bdde09646ea959d94c714caca72R53-R54

andrewstrohman:
> > If the calculated digest does not match any configured digest, that uprobe is skipped and a warning is reported in policy status.
> 
> On this, I just thought on this new status that we just added that could be reused here [e3b2e48#diff-b51cb791919665a2013217e038932107da1e4bdde09646ea959d94c714caca72R53-R54](https://github.com/cilium/tetragon/commit/e3b2e489565eb6f5b7c68e5eeb185c219c8ee8d4#diff-b51cb791919665a2013217e038932107da1e4bdde09646ea959d94c714caca72R53-R54)

In the next PR, the tracing policy will be partially loaded, so I don't think we need to worry too much about this intermediate state. If I changed the tracing policy status to TP_STATE_SKIPPED, I would need to undo that in the next PR, so I don't think there is a lot of value in perfecting this. At this point in the series of PRs, the tracing policy status for digest rejected policies is the same as when the target binary doesn't exist.

Maybe in the next PR, I could add a TracingPolicyState TP_STATE_PARTIALLY_LOADED. The current way that I'm proposing to communicate the partial load is via a per-hook status, but I could also mark the policy as partially loaded too.

You pointing this out does remind me that I need to change that wording for this PR to reflect the intermediate state we are in (no information in policy status).

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -635,6 +638,122 @@ func cleanupUprobeEntries(ids []idtable.EntryID, openedFiles []*os.File) error {
...
+		return nil
+	}
+
+	algo, expectedHash, found := strings.Cut(digestConfig, ":")
+	if !found {
+		return fmt.Errorf("invalid digest format, expected '<algo>:<hash>' but got '%s'", digestConfig)
+	}
+
+	algo = strings.ToLower(algo)
+	expectedHash = strings.ToLower(expectedHash)
```
mtardy: maybe useless note: if you want to accept `  sha256 :    abc123 ` with random space you can also `strings.TrimSpace` here.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -635,6 +638,122 @@ func cleanupUprobeEntries(ids []idtable.EntryID, openedFiles []*os.File) error {
...
+func verifyBinaryDigests(uprobe *v1alpha1.UProbeSpec, entryFile *os.File, cfgIdx int) error {
+	if entryFile == nil {
+		return nil
+	}
+
+	digestCache := make(map[string]string)
+	for _, digest := range uprobe.BinaryDigests {
+		if err := verifyFileDigest(entryFile, digest, digestCache); err == nil {
+			return nil
+		}
```
mtardy: shouldn't we surface the errors returned error on != nil somehow, maybe at a later stage in the final returned error line 754 and in the debug log?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -635,6 +638,122 @@ func cleanupUprobeEntries(ids []idtable.EntryID, openedFiles []*os.File) error {
...
+		actualHashes = append(actualHashes, fmt.Sprintf("%s:%s", algo, hash))
+	}
+	logger.GetLogger().Debug(
+		"digest verification failed: none of the configured digests matched",
+		"path", uprobe.Path,
+		"index", cfgIdx,
+		"configured", strings.Join(uprobe.BinaryDigests, ", "),
+		"actual", strings.Join(actualHashes, ", "),
+	)
+	return fmt.Errorf("digest verification failed for path %q at hook index %d", uprobe.Path, cfgIdx)
```
mtardy: also hook index %d could be spec.Uprobes[%d], I feel sometime it's helpful to stick to the spec definition so it's easy to locate the issue, like we do for TP validation.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -698,11 +817,26 @@ func createGenericUprobeSensor(
...
+			}
+		}
+
+		if err := verifyBinaryDigests(&uprobe, entryFile, cfgIdx); err != nil {
+			if entryFile != nil {
+				if closeErr := entryFile.Close(); closeErr != nil {
+					logger.GetLogger().Warn("failed to close file after binary digest verification failure", logfields.Error, closeErr, "path", uprobe.Path)
+				}
+			}
+			return nil, err
```
mtardy: I feel like this `return nil, err` is where you would actually enrich your error with the cfgIdx since this is where you are looping over the uprobes instead of passing this by arg and doing it within the function.

However for your logging you might be able to provide less details depending if you move it here or keep it in the function.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -1009,7 +1143,7 @@ func addUprobeEntries(spec *v1alpha1.UProbeSpec, ids []idtable.EntryID, state *u
 		// and as such their file descriptors are no longer valid after that point. So, it's not safe to load the sensor again
 		// when we re-enable a policy after it's been disabled, because the configured targetPaths would point to closed/re-used file descriptors.
 		if state.entryFile != nil {
-			has.disableNotAllowedReason = disableNotAllowedReasonLinkByFD
```
mtardy, line 1013: q: why the old reason is abandoned and replaced and the new one not added on top of it depending on the situation? I haven't read the full function so might be missing stuff here

pkg/sensors/tracing/uprobe_digest_test.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func TestUprobeDigestMultipleDigestsAllWrong(t *testing.T) {
+	// All digests are wrong
+	wrongDigests := []string{
+		"sha256:1111111111111111111111111111111111111111111111111111111111111111",
+		"sha256:2222222222222222222222222222222222222222222222222222222222222222",
+	}
+	policyLoadMatchesExpectation(t, uprobePolicy(wrongDigests), false)
+}
```
mtardy: nit: not crazy preference but I feel this could be one test with subtests since it's testing the same thing in different scenarios, potentially sharing the test.go file with the rest of the test. wdyt?

pkg/sensors/tracing/genericuprobe.go
mtardy: with the "tracing: avoid opening a file multiple times" commit here you are still closing a file that is stored in your cache right?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -765,6 +765,7 @@ func createGenericUprobeSensor(
 	var err error
 	var has uprobeHas
 	var celExprs *selectors.CelExprFunctions
+	openedFilesMap := make(map[string]*os.File)
```
mtardy: who is taking care of cleaning up this?

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -366,6 +366,44 @@ sensitive command input.
 For more details on available selectors and their usage, see the
 [Selectors]({{< ref "/docs/concepts/tracing-policy/selectors" >}}) documentation.
 
+### Target Binary Digest Verification
+
+In some cases, a uprobe policy is only intended to apply to a specific version of a binary, and could have adverse effect if attached to a different version of the binary.
```
mtardy: please wrap the lines at 80 chars

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -366,6 +366,44 @@ sensitive command input.
...
+
+For users that want to only apply the policy for a specific build, they can specify a digest type and value to validate the target binary. The following example policy demonstrates this functionality.
+
+```
```
mtardy: add `yaml` please after the three backtickts

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -366,6 +366,44 @@ sensitive command input.
...
+spec:
+    uprobes:
+    - path: "/bin/example-target-binary"
+      symbols:
+      - "main"
+      binaryDigests:
+        - "sha256:1111111111111111111111111111111111111111111111111111111111111111"
+        - "sha256:2222222222222222222222222222222222222222222222222222222222222222"
+```
+When the above policy is loaded, tetragon will calculate the sha256 hash of `/bin/example-target-binary` and compare it against all the binaryDigests configured within the `binaryDigests` list. The policy will be rejected if the calculated digest does not match any of the configured binaryDigests.
```
mtardy: it against all the binaryDigests configured within the `binaryDigests` list -> it against all the binary digests configured

mtardy: maybe specify what it means explicitly what it means to be rejected? is it still around when listing policies, is it just ignored?

docs/content/en/docs/concepts/tracing-policy/hooks.md
```diff
@@ -366,6 +366,44 @@ sensitive command input.
...
+The following digest types are supported:
+- sha1
+- sha256
+- sha384
+- sha512
+- build-id (found in the ELF note section named `.note.gnu.build-id`)
+
+When a uprobe is attached, the kernel keeps track of the target by its inode. It's best practice for installation software (such as rpm, deb, install, tar) to unlink a pre-existing target before installing a new version. This ensures that the new version of the binary will have a new inode, so the uprobe will not fire for it.
+
+But, if the target binary is updated without an unlink, which would happen if `cp` is used or if the target is opened and updated in-place, the inode will be retained. In this case, the uprobe will continue firing for the updated binary, potentially causing unintended effect. As such, it's critical that users of this feature ensure their installation method will not retain the inode of the target binary that they want to update.
```
mtardy: maybe it's worth putting this into a `caution` block?

mtardy (commented):
Thanks! This is a neat. A few comments:

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: sounds good. done.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: yes, that's better. done.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: `openedFilesMap`, itself, does not need to be cleaned up. Its purpose is to prevent opening the same file twice when the same hook is present within one policy.

The files that are opened need to be closed. This is tracked with `openedFiles`. `openedFiles` is captured in the PostLoadHook so that those files can be closed.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I did it the way that you described originally, but the feedback I received was to just replace it because there is currently only one reason, so there is no need to introduced that complexity yet.

See [here](https://github.com/cilium/tetragon/pull/5182#discussion_r3559320881) for that earlier discussion.

pkg/sensors/tracing/uprobe_digest_test.go
andrewstrohman: >  I feel this could be one test with subtests since it's testing the same thing in different scenarios

Ok, I made these sub-tests.

> potentially sharing the test.go file with the rest of the test

I'm not able to follow this suggestion. Can you re-phrase it please? FWIW, I'd like to keep these tests in this file instead of adding to `pkg/sensors/tracing/uprobe_test.go`, because uprobe_test.go is starting to get large.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Yes, if the file was already opened in a previous iteration, then we will close here, and then try to close again in the defer'ed function. That defered cleanup checked for `!errors.Is(closeErr, os.ErrClosed)` which made the behavior functionally correct, but sloppy.

I've improved this by only closing the file once and I removed the checks for `!errors.Is(closeErr, os.ErrClosed)` because this cannot happen anymore.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I ended up consolidating `openFilesMap` with `openFiles` so that we don't have two different data structures tracking the open files.

docs/content/en/docs/concepts/tracing-policy/hooks.md
andrewstrohman: I did this (except for the example policy, which has line length over 80).

docs/content/en/docs/concepts/tracing-policy/hooks.md
andrewstrohman: done

docs/content/en/docs/concepts/tracing-policy/hooks.md
andrewstrohman: I changed it to: "it against all the configured binary digests" because I thought that wording is a little more clear.

docs/content/en/docs/concepts/tracing-policy/hooks.md
andrewstrohman: Good idea, done.

docs/content/en/docs/concepts/tracing-policy/hooks.md
andrewstrohman: Ok, I explained that the policy will show as `load_error` when listing the policies.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Yeah, I changed it so that operational errors (as opposed to digest mismatch) are returned.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Sounds good. I did the enrichment at the caller, so that cfgIdx is not passed.

kkourt (approved):
Thanks Andy! 🚀

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -626,15 +639,138 @@ func cleanupUprobeEntries(ids []idtable.EntryID, openedFiles []*os.File) error {
...
+			return fmt.Errorf("unsupported digest algorithm '%s'", algo)
+		}
+
+		h := hashType.New()
+		if _, err := io.Copy(h, file); err != nil {
+			return fmt.Errorf("failed to calculate digest: %w", err)
+		}
+
+		calculatedHash = hex.EncodeToString(h.Sum(nil))
+	}
```
FedeDP: I'd move this block of code into a `computeHash() string` helper function.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -626,15 +639,138 @@ func cleanupUprobeEntries(ids []idtable.EntryID, openedFiles []*os.File) error {
...
+		err := verifyFileDigest(entryFile, digest, digestCache)
+		if err == nil {
+			return nil
+		}
+
+		// Keep trying other configured digests only when the current digest
+		// mismatches. Operational errors should be returned immediately.
+		if !errors.As(err, &mismatchErr) {
+			return err
+		}
```
FedeDP:
```suggestion
if _, ok := errors.AsType[*DigestMismatchError](err); !ok {
	return err
}
```
and we can drop `var mismatchErr *DigestMismatchError`

pkg/sensors/tracing/generic.go
```diff
@@ -22,6 +23,20 @@ import (
 	"github.com/cilium/tetragon/pkg/selectors"
 )
 
+func getOrOpenFile(path string, openedFilesMap map[string]*os.File) (*os.File, error) {
```
FedeDP: Are we planning to use this helper elsewhere? Otherwise i'd move it in genericuprobe.go.

pkg/sensors/tracing/genericuprobe.go
FedeDP: (killed the indentation in the process, sorry!)

mtardy (approved):
Cool thanks!

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Sounds good. Done.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Yes, that's better. I did this.

pkg/sensors/tracing/generic.go
andrewstrohman: Initially, I put in generic.go  because this was for both uprobe and usdt. But the code has changed, and now  it's only for uprobe, so I agree it should be in genericuprobe.go. So, I moved it there.

PR #5175, opened 2026-06-23
Implement matchCaller filter

3u13r:
### Description

Current state: Doing the first review by myself (with some AI). Feel free to already have a look.
There are some architecture changes to the eBPF filter matching that might be controversial and should be discussed or at least ACK'd by a few maintainers.

### TODOs

- [x] User-facing docs
    1. Because of https://github.com/torvalds/linux/commit/cfa7f3d2c526c224a6271cc78a4a27a0de06f4f0 one can only match on the immediate caller when the platform is amd64 and has a kernel version >=6.12
    2. Because of https://github.com/torvalds/linux/commit/b33164f2bd1cedb094c32cb466287116164457ae in kernel versions <5.9 we get absolute offsets from bpf_get_stack, since the helper only looks into the first NOTE section for the build id. But the build is in the second NOTE section by default with e.g., gcc 13.3.0. If the binary doesn't fulfill this requirement on <5.9 Tetragon rejects the policy .

### Changelog

```release-note
Add TracingPolicy feature filter for caller functions in uprobe selectors
```

PR #5283, opened 2026-07-16
misc: various fixes

3u13r:
### Description

A collection of various fixed I did during https://github.com/cilium/tetragon/pull/5175/commits. Since all of them are unrelated to the actual feature, we moved them here.

Let me know if you prefer one PR for each change.

### Changelog

```release-note
tracingpolicy: fix some typos in KProbeSelector and ArgSelector CRD docs
```

### DIFF
--- a/pkg/elf/build_id.go
+++ b/pkg/elf/build_id.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"debug/elf"
 	"encoding/binary"
+	"errors"
 	"io"
 )
 
@@ -24,7 +25,7 @@ func align[I Integer](n, alignment I) I {
 type note struct {
 	Namesz uint32
 	Descsz uint32
-	Typ    uint32
+	Type   uint32
 }
 
 // ParseBuildIdFromNotes returns the GNU build ID found in a raw ELF notes blob,
@@ -51,7 +52,7 @@ func ParseBuildIdFromNotes(dat []byte, order binary.ByteOrder) ([]byte, bool) {
 			return []byte{}, false
 		}
 
-		if note.Typ == 3 &&
+		if note.Type == 3 && // NT_GNU_BUILD_ID = 3
 			note.Namesz == 4 &&
 			bytes.Equal(name, []byte{'G', 'N', 'U', 0}) &&
 			note.Descsz > 0 {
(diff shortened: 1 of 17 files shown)


pkg/testutils/policytest/conf.go
```diff
@@ -43,6 +44,9 @@ type Conf struct {
 
 	// Path to save the generated policy
 	DumpPolicyPath string
+
+	// GRPCTimeout to use while waiting for a scenario's expected events
+	GRPCTimeout time.Duration
```
kkourt: Heh: https://github.com/cilium/tetragon/pull/5247/changes/2cad7b99f6c60d1eac2e4b6429301fa4d739ee96

pkg/elf/build_id.go
```diff
@@ -24,7 +25,7 @@ func align[I Integer](n, alignment I) I {
 type note struct {
 	Namesz uint32
 	Descsz uint32
-	Typ    uint32
+	Type   uint32
```
kkourt, line 28: Why the renaming? I think it's clear that `Typ` refers to the type.
IMO, we should only renaming things if we are making functional changes in the code anyway.

kkourt (approved):

pkg/elf/build_id.go
3u13r: There's not much more to it than what I wrote in the commit message.

My thought is basically:
 1. Oh this doesn't align with the spec
 2. Is there a reason that it doesn't align with the spec?
 3. No? Then align it to save others the thought process of 1. and 2.

Also while `Type` is not a very specific name, it also might help if one grep's for certain names in the repo.

In this PR it's more out-of-the blue than in the original PR, since there I at least touched this file and used this variable.

But there doesn't need to be a big discussion about this. ~I'll merged it since you already approved it, otherwise I'd just have removed it.~
Since you need to touch this PR again to merge it, feel free to either ping me or directly edit this PR and drop the commit

pkg/testutils/policytest/conf.go
3u13r: Let me drop it from my set, then you have less issues with rebasing.

pkg/elf/build_id.go
3u13r: Just want to mention, if we really don't like polluting our history, there is the options to exclude certain changes from `git blame` (also from GitHub's integrated blame) via a `.git-blame-ignore-revs` file.

pkg/testutils/policytest/conf.go
kkourt: I don't mind if you needed for the other PR. I'm happy to rebase https://github.com/cilium/tetragon/pull/5247.

pkg/elf/build_id.go
kkourt: > But there doesn't need to be a big discussion about this

Agreed. I'm not arguing to remove it (this is why I clicked Approved rather than Request Changes), but I think it's worth having a discussion on how we want to deal with these types of changes in general (rather than in this specific commit).

> there is the options to exclude certain changes from git blame (also from GitHub's integrated blame) via a .git-blame-ignore-revs file.

Interesting. Can you provide an example?

kkourt:
~Note: I split the release-notes into two lines: one for each change.~

Nevermind, you removed the policytest --timeout change.

pkg/elf/build_id.go
3u13r: [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/.git-blame-ignore-revs) uses this a lot especially when they finished their formatter and formatted the whole repository.

PR #5286, opened 2026-07-16
fix: uprobe multi hooks `override` action overwritten by latest one

### LINKED ISSUE #5285: Specs with multiple uprobe hooks with an override matchAction only accounts for last set

Reported by @litios in https://github.com/cilium/tetragon/issues/4578#issuecomment-4449957373.
Basically, when a Spec has multiple uprobe hooks each one with its `Override` matchAction, the last one overwrites all of them.
Example:
```yaml
kind: TracingPolicy
metadata:
  name: "test-binary"
spec:
  uprobes:
  - path: "/home/litios/git/pocs/burning/binary"
    addrs:
    - 0x00000000000011c9
    selectors:
      - matchActions:
        - action: Override
          argError: 1111
  - path: "/home/litios/git/pocs/burning/binary"
    addrs:
    - 0x000000000000121c
    selectors:
      - matchActions:
        - action: Override
          argError: 2222
```
In this case, if the first one gets triggered, its return code will be 2222 instead of 1111.

The issue seem to only appear for multi uprobes: https://github.com/cilium/tetragon/issues/4578#issuecomment-4990297862

### Tetragon Version

Latest main version.

### Kernel Version

```
Linux fdipierr-l-pw0bf5nd 6.17.0-40-generic #40~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Jun 23 16:48:12 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

FedeDP:
Fixes #5285

### Description

As described in the issue, whenever a spec with multiple uprobes, each of one with its own `override` match action, was applied, the last uprobe.matchAction.Override arguments would overwrite the override arguments for all other uprobes.
The problem is that `populateUprobeRegs` always overwrote the `0` index of the `regs_map`; that is correct in the single sensor, but for multi, each uprobe entry should fill its own index, otherwise the last entry would take over.
Needed also to pass the `uprobeID` (the index of the current uprobe entry we are creating) to `InitKernelSelectorState` to allow `selectors.parseOverrideRegs()` to correctly store the index to be looked for in ebpf.

### Changelog

```release-note
fix: uprobe multi hooks `override` action overwritten by latest one
```

### DIFF
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -1620,6 +1620,7 @@ type KernelSelectorArgs struct {
 	ListReader     ValueReader
 	Maps           *KernelSelectorMaps
 	IsUprobe       bool
+	UprobeID       int
 	CelExprs       *CelExprFunctions
 }
 
@@ -1684,13 +1685,14 @@ func createKernelSelectorState(
 	listReader ValueReader,
 	maps *KernelSelectorMaps,
 	isUprobe bool,
+	uprobeID int,
 	celExprs *CelExprFunctions,
 	parseSelector func(k *KernelSelectorState, selectors *v1alpha1.KProbeSelector, selIdx int) error,
 ) (*KernelSelectorState, error) {
 	if len(selectors) > MaxSelectors {
 		return nil, fmt.Errorf("no more than %d selectors supported (%d provided)", MaxSelectors, len(selectors))
 	}
-	state := NewKernelSelectorState(listReader, maps, isUprobe, celExprs)
+	state := NewKernelSelectorState(listReader, maps, isUprobe, uprobeID, celExprs)
 
 	WriteSelectorUint32(&state.data, uint32(len(selectors)))
 	soff := make([]uint32, len(selectors))
@@ -1744,7 +1746,7 @@ func InitKernelSelectorState(args *KernelSelectorArgs) (*KernelSelectorState, er
 		return nil
 	}
 
-	return createKernelSelectorState(args.Selectors, args.ListReader, args.Maps, args.IsUprobe, args.CelExprs, parse)
+	return createKernelSelectorState(args.Selectors, args.ListReader, args.Maps, args.IsUprobe, args.UprobeID, args.CelExprs, parse)
 }
 
 func InitKernelReturnSelectorState(selectors []v1alpha1.KProbeSelector, returnArg *v1alpha1.KProbeArg,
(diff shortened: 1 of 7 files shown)


andrewstrohman (approved):

andrewstrohman:
I think this approach is functionally correct, but somewhat wasteful in map memory.

The regs_map has one entry per uprobe entry. However, some or many of those entries may not correspond to an override-regs action, so they end up carrying no meaningful override configuration.

Also consider the case where one uprobe spec entry has multiple addrs/offsets/symbols. In that case, a single spec entry generates multiple uprobe IDs (one per attach point) that all share the same selectors.

When this happens, regs_map gets repetitive entries because the same selector configuration is duplicated across those generated probes. The selector state sent through filter_map uses the first probe index in that sequence, so the BPF override path ends up consulting that first regs_map entry regardless of which attach point fired. The later duplicate regs_map entries are therefore effectively unused.

A leaner approach would be to populate regs_map only with non-duplicate, actually used entries. I do not think the memory savings are worth the extra complexity right now, but it seemed worth calling out.

kkourt (approved):

kkourt:
> I think this approach is functionally correct, but somewhat wasteful in map memory.

[...]
Thanks for the analysis Andy!

> I do not think the memory savings are worth the extra complexity right now, but it seemed worth calling out.

If I understand correctly, the memory overhead is one sizeof(pt_regs) per uprobe entry in the policy, while it could be one sizeof(pt_regs) per uprobe entry with an override in the policy? If that's the case, sizeof(pt_regs) seems to be 168 bytes on x86_64, so I would agree that it does not worth the extra complexity at this point.

kkourt:
@FedeDP I think we need to backport this?

FedeDP:
Thanks @andrewstrohman for your in depth insight. Yes i chose the simple solution over the complex one since the memory footprint of `pt_regs` is very small.
At the same time, i always forgot that we can have a single uprobe spec with multiple addrs/offsets/symbols :/ thanks for pointing that out!

FedeDP:
Cherry picked in https://github.com/cilium/tetragon/pull/5294

andrewstrohman:
> If I understand correctly, the memory overhead is one sizeof(pt_regs) per uprobe entry in the policy, while it could be one sizeof(pt_regs) per uprobe entry with an override in the policy? If that's the case, sizeof(pt_regs) seems to be 168 bytes on x86_64, so I would agree that it does not worth the extra complexity at this point.

> Yes i chose the simple solution over the complex one since the memory footprint of pt_regs is very small.

Yes, I agree with this. The amount of wasted memory is very small.

PR #5300, opened 2026-07-17
option: validate export-rate-limit and warn on invalid or zero values

### LINKED ISSUE #5281: export-rate-limit: invalid config value silently parsed as 0, dropping all events

👋 hi tetragon team! 

Setting `--export-rate-limit: 50000,1s` (an invalid value) in `tetragon.yaml` causes 100% event drop rate. No warning or error is logged as far as I can tell. Removing the flag entirely lets events through normally. After looking through the code,  the format `50000,1s` actually does not apply to `--export-rate-limit`, thats just for `--cgroup-rate`, however the docs didn't explicitly state as much and I had made an assumption that was not correct when configuring tetragon.

  To reproduce:
  1. Add an invalid config such as `export-rate-limit: 50000,1s` in `tetragon.yaml`
  2. Run Tetragon and generate events
  3. Observe 0 events exported and the `tetragon_export_ratelimit_events_dropped_total` prom metric go up.

  Expected: either events flow through, or at minimum a config error is logged.

  Root cause (traced in code):

It appears that `viper.GetInt` [silently returns 0 ](https://github.com/cilium/tetragon/blob/1de2ed8ebea18e56257dc59597aa13bf8f0e471e/pkg/option/flags.go#L245) when it can't parse the value as a plain integer. That 0 then passes the >= 0 guard in [cmd/tetragon/main.go:668](https://github.com/cilium/tetragon/blob/1de2ed8ebea18e56257dc59597aa13bf8f0e471e/cmd/tetragon/main.go#L668-L669), creating a rate limiter with  zero burst capacity. `Allow()` always returns false and everything gets dropped. Nothing is logged.

  Fix:

  - Log a warning at startup if ExportRateLimit parsed to 0 so operators aren't left guessing
  - Consider treating 0 the same as -1 (disabled) rather than "drop everything"
  - Update docs to clarify the value must be a plain integer (events per minute, no unit suffix)

  Happy to put up a PR for this.

### Tetragon Version

`1.7.0`

### Kernel Version

`6.12.92`

I don't think this applies however as the issue is in userspace golang.

### Kubernetes Version

n/a

### Bugtool

n/a

### Relevant log output

```shell
n/a
```

### Anything else?

Happy to put together a PR to update docs/do input validation on the config value.

stefanamaerz:
### Description
Fixes: #5281

`export-rate-limit` is registered as an integer flag, but when set via a config file with a non-integer value (e.g. `50000,1s`), `viper.GetInt` silently returns `0`. A rate limit of `0` creates a zero-burst-capacity limiter where `Allow()` always returns false. This silently drops 100% of events with no indication why.

This PR uses `cast.ToIntE(viper.Get(...))` so the same parse path handles both CLI and config-file values and actually reports parse failures. On a parse error, rate limiting is disabled (falls back to -1) and a startup warning is logged. A value of 0 disables JSON export entirely (no exporter is created), instead of building a rate limiter that would silently reject every event.

Flag description, generated docs, and Helm values are updated to document the plain-integer requirement.

### DIFF
--- a/pkg/option/flags.go
+++ b/pkg/option/flags.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/go-viper/mapstructure/v2"
+	"github.com/spf13/cast"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
 
@@ -252,7 +253,13 @@ func ReadAndSetFlags() error {
 	Config.ExportFileRotationInterval = viper.GetDuration(KeyExportFileRotationInterval)
 	Config.ExportFileMaxBackups = viper.GetInt(KeyExportFileMaxBackups)
 	Config.ExportFileCompress = viper.GetBool(KeyExportFileCompress)
-	Config.ExportRateLimit = viper.GetInt(KeyExportRateLimit)
+	value := viper.Get(KeyExportRateLimit)
+	Config.ExportRateLimit, err = cast.ToIntE(value)
+	if err != nil {
+		logger.GetLogger().Warn(fmt.Sprintf("failed to parse %s '%v', falling back to -1 (no rate limiting): set 0 to disable JSON export, or a positive integer to rate limit it", KeyExportRateLimit, value),
+			logfields.Error, err)
+		Config.ExportRateLimit = -1
+	}
 	Config.ExportFilePerm = viper.GetString(KeyExportFilePerm)
 
 	Config.EnableExportAggregation = viper.GetBool(KeyEnableExportAggregation)
@@ -464,7 +471,7 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.Int(KeyExportFileMaxBackups, 5, "Number of rotated JSON export files to retain")
 	flags.Bool(KeyExportFileCompress, false, "Compress rotated JSON export files")
 	flags.String(KeyExportFilePerm, defaults.DefaultLogsPermission, "Access permissions on JSON export files")
-	flags.Int(KeyExportRateLimit, -1, "Rate limit (per minute) for event export. Set to -1 to disable")
+	flags.Int(KeyExportRateLimit, -1, "Rate limit (per minute) for event export. Set to -1 to disable rate limiting, 0 to disable JSON export, or a positive integer to rate limit event export")
 	flags.String(KeyLogLevel, "info", "Set log level")
 	flags.String(KeyLogFormat, "text", "Set log format")
 	flags.String(KeyLogFile, "", "Set log file where tetragon agent logs will be written (in addition to stdout or stderr)")
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -472,9 +472,11 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 		return err
 	}
 
-	// Fetch the exporter if needed
+	// Fetch the exporter if needed. An export rate limit of 0 disables JSON
+	// export altogether, so avoid creating an exporter that would drop every
+	// event.
 	var exporter *exporter.Exporter
-	if option.Config.ExportFilename != "" {
+	if option.Config.ExportFilename != "" && option.Config.ExportRateLimit != 0 {
 		exporter, err = getExporter(ctx, pm.Server)
 		if err != nil {
 			return fmt.Errorf("failed to fetch json exporter: %w", err)
@@ -681,7 +683,7 @@ func getExporter(ctx context.Context, server *server.Server) (*exporter.Exporter
 	encoderWriter := exporter.NewExportedBytesTotalWriter(writer)
 	encoder := encoder.NewProtojsonEncoder(encoderWriter)
 	var rateLimiter *ratelimit.RateLimiter
-	if option.Config.ExportRateLimit >= 0 {
+	if option.Config.ExportRateLimit > 0 {
 		rateLimiter = ratelimit.NewRateLimiter(ctx, 1*time.Minute, option.Config.ExportRateLimit, encoder)
 	}
 	var aggregationOptions *tetragon.AggregationOptions
--- a/docs/content/en/docs/reference/helm-chart.md
+++ b/docs/content/en/docs/reference/helm-chart.md
@@ -118,7 +118,7 @@ To use [the values available](#values), with `helm install` or `helm upgrade`, u
 | tetragon.exportFileMaxSizeMB | int | `10` | Size in megabytes at which to rotate JSON export files. |
 | tetragon.exportFilePerm | string | `"600"` | JSON export file permissions as a string. Typically it's either "600" (to restrict access to owner) or "640"/"644" (to allow read access by logs collector or another agent). |
 | tetragon.exportFilename | string | `"tetragon.log"` | JSON export filename. Set it to an empty string to disable JSON export altogether. |
-| tetragon.exportRateLimit | int | `-1` | Rate-limit event export (events per minute), Set to -1 to export all events. |
+| tetragon.exportRateLimit | int | `-1` | Rate-limit event export (events per minute). Must be a plain integer: -1 disables rate limiting and exports all events (default), 0 disables JSON export entirely, and a positive integer limits export to that many events per minute. Invalid (non-integer) values fall back to -1 and log a warning. |
 | tetragon.extraArgs | object | `{}` |  |
 | tetragon.extraEnv | list | `[]` |  |
 | tetragon.extraVolumeMounts | list | `[]` |  |
--- a/docs/data/tetragon_flags.yaml
+++ b/docs/data/tetragon_flags.yaml
@@ -145,7 +145,7 @@ options:
     - name: export-rate-limit
       default_value: "-1"
       usage: |
-        Rate limit (per minute) for event export. Set to -1 to disable
+        Rate limit (per minute) for event export. Set to -1 to disable rate limiting, 0 to disable JSON export, or a positive integer to rate limit event export
     - name: expose-stack-addresses
       default_value: "false"
       usage: Expose real linear addresses in events stack traces
--- a/go.mod
+++ b/go.mod
@@ -31,6 +31,7 @@ require (
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.70.0
 	github.com/prometheus/procfs v0.21.1
+	github.com/spf13/cast v1.10.0
 	github.com/spf13/cobra v1.10.2
 	github.com/spf13/pflag v1.0.10
 	github.com/spf13/viper v1.21.0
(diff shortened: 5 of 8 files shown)


FedeDP (changes requested):
Ehy thanks for this!
So, i'd squash 1,2,4 commits into a single one; also, the unit-test related commit contains stuff that is not meant to be there; i'd edit the commit to just keep changes to `pkg/option/export_rate_limit_test.go`.
I am not even sure whether we need a test like that, after all we are testing that `strconv.Atoi` works fine, aren't we?

Finally, i am wondering whether we'd better just update `pkg/ratelimit/ratelimit.go::getLimit()` to return `rate.Inf` in case the numEvents is 0:
```go
func getLimit(numEvents int, interval time.Duration) rate.Limit {
	if numEvents == 0 {
		return rate.Inf
	}
	return rate.Every(interval / time.Duration(numEvents))
}
```
since the current check:
```go
var rateLimiter *ratelimit.RateLimiter
	if option.Config.ExportRateLimit >= 0 {
		rateLimiter = ratelimit.NewRateLimiter(ctx, 1*time.Minute, option.Config.ExportRateLimit, encoder)
	}
```
seems to allow a `option.Config.ExportRateLimit` value of 0 (that imho, if as effect produces a rateLimiter that limits everything, it would be better to not even create the ratelimiter in the first place!)

stefanamaerz:
@FedeDP , thank you for the thoughtful review, much appreciated! 

I've implemented your suggestion and squashed into a single commit and removed the test. Let me know if that looks good on your end!

pkg/ratelimit/ratelimit.go
```diff
@@ -29,7 +29,7 @@ type RateLimiter struct {
 // representing number of events per second.
 func getLimit(numEvents int, interval time.Duration) rate.Limit {
 	if numEvents == 0 {
-		return 0
+		return rate.Inf
```
FedeDP: @kkourt perhaps you got more context on the choice to return `0` when `numEvents` was 0?
It felt weird that we allowed to create a ratelimiter that basically dropped all events, but maybe it was somehow a desired behavior.

kkourt: I don't have additional context. Is `numEvents` a user-controlled parameter?

FedeDP: Yes `numEvents` comes from `option.Config.ExportRateLimit` and the ratelimiter gets created like this:
```
var rateLimiter *ratelimit.RateLimiter
if option.Config.ExportRateLimit >= 0 {
	rateLimiter = ratelimit.NewRateLimiter(ctx, 1*time.Minute, option.Config.ExportRateLimit, encoder)
}
```
In the current code, a  `option.Config.ExportRateLimit == 0` would've the effect of "reject all"; i am not sure whether it was intentional (probably) and/or it makes sense.
Perhaps it is a use case that we want to keep, ie:
* -1 (default value) -> no ratelimit in place
* 0 -> always reject
* 1+ -> number of allowed events per minute

Flag description:
> flags.Int(KeyExportRateLimit, -1, "Rate limit (per minute) for event export. Set to -1 to disable")

My proposal implemented by @stefanamaerz (sorry for the back and forth btw!) is to make `0` have the same effect of -1, but it we think "always reject" is a valid behavior, we should retain it.
At that point, perhaps, if `option.Config.ExportRateLimit == 0` i would short-circuit the creation of the `newExporter` and avoid it at all, since we would create an exporter that rejects all events.

stefanamaerz: IMO: the `0 -> always reject` rate limiter doesn't make sense. The only usecase I could see is if someone wanted to use this to disable logging. 

However I think they could simply disable grpc/json logging via:

```
# disable JSON export
--export-filename=""
# disable gRPC
--server-address=""
```

https://tetragon.io/docs/reference/daemon-configuration/

FedeDP: I agree with you; so in the end `option.Config.ExportRateLimit == 0` means either:
* shortcircuit for exporter without rate limiter (proposed behavior)
* shortcircuit for no exporter (similar to today's behavior but avoid creating the exporter at all, instead of creating an exported that rejects everything)

FedeDP: After a discussion with @kkourt , the idea is to let `option.Config.ExportRateLimit == 0`:
> shortcircuit for no exporter (similar to today's behavior but avoid creating the exporter at all, instead of creating an exporter that rejects everything)

This way we have no breaking change in UX, but we still clear a bit the expectations.
Basically, `main.go` code would become:
```go
	// Fetch the exporter if needed
	var exporter *exporter.Exporter
	if option.Config.ExportFilename != "" && option.Config.ExportRateLimit != 0 {
		exporter, err = getExporter(ctx, pm.Server)
		if err != nil {
			return fmt.Errorf("failed to fetch json exporter: %w", err)
		}
	}

```

@stefanamaerz does it make sense to you?
Perhaps we can also update `KeyExportRateLimit` usage string stating that `0` value will disable JSON export.

stefanamaerz: @FedeDP  thanks for the clarification! I think I understand now. Let me reiterate to make sure we're aligned before I push the next update:
- `-1` (default): no rate limiting, JSON export is enabled.
- `0`: explicitly disable JSON export entirely (short-circuit the exporter creation). This preserves the observable behavior of the current bug/code (no events exported) but does so cleanly (and more efficently/no json serialization overhead) instead of creating a rate limiter that rejects everything.
- positive integer: rate-limit JSON export to N events per minute.
- invalid/non-integer value (e.g. `50000,1s`): log a startup warning and fall back to the default -1 (no rate limiting; JSON export enabled).


This means the current PR needs two high-level changes:
1. Keep the `viper.GetString` + `strconv.Atoi` validation and warning for invalid values, falling back to `-1`.
2. Revert `pkg/ratelimit/ratelimit.go` back to returning `rate.Limit(0)` for `numEvents == 0`.
3. In `cmd/tetragon/main.go`, skip `getExporter(...)` when `option.Config.ExportRateLimit == 0` so no exporter is created at all (and document that `0` disables JSON export in the flag/Helm docs).


Does that capture it correctly? If so I'll update the PR accordingly.

FedeDP: Exactly, thank you so much and sorry if i had you waste some time on this one :)

stefanamaerz: @FedeDP no worries at all, I very much appreciate your guidance here as this is my first PR to the tetragon repo!

I pushed a squashed commit; let me know if that looks good to you!

pkg/option/flags.go
```diff
@@ -257,7 +252,7 @@ func ReadAndSetFlags() error {
 	Config.ExportFileRotationInterval = viper.GetDuration(KeyExportFileRotationInterval)
 	Config.ExportFileMaxBackups = viper.GetInt(KeyExportFileMaxBackups)
 	Config.ExportFileCompress = viper.GetBool(KeyExportFileCompress)
-	Config.ExportRateLimit = viper.GetInt(KeyExportRateLimit)
+	Config.ExportRateLimit = parseExportRateLimit(viper.GetString(KeyExportRateLimit))
```
FedeDP: I am still unsure about this change; i found this comment with a proposed workaround: https://github.com/spf13/viper/issues/1243#issuecomment-968842010
```go
Config.ExportRateLimit, err = cast.ToIntE(viper.Get(key))
if err != nil {
    // it's not an int
    logger.GetLogger().Warn(fmt.Sprintf("failed to parse %s '%s', falling back to -1 (no rate limiting): set 0 to disable JSON export, or a positive integer to rate limit it", KeyExportRateLimit, value),
			logfields.Error, err)
    Config.ExportRateLimit = -1
}
```
What do you think about using this?
I'd like to avoid the configuration breaking change (after all we are going to have the same "cast to int" issue for every non-string flags, it's not very doable to change them all. I understand `KeyExportRateLimit` is problematic because a value of 0 would kill the export file though).

stefanamaerz: Hi @FedeDP, thanks for the suggestion! I agree, that's much cleaner. I vaguely recall this issue, but I think I missed it. This is cleaner and avoids a needless breaking change.

I pushed an update which uses `cast.ToIntE(viper.Get())` per your suggestion. On parse error we warn and fall back to -1. Let me know if there's anything else you'd like adjusted!

pkg/option/flags.go
```diff
@@ -474,7 +471,7 @@ func AddFlags(flags *pflag.FlagSet) {
 	flags.Int(KeyExportFileMaxBackups, 5, "Number of rotated JSON export files to retain")
 	flags.Bool(KeyExportFileCompress, false, "Compress rotated JSON export files")
 	flags.String(KeyExportFilePerm, defaults.DefaultLogsPermission, "Access permissions on JSON export files")
-	flags.Int(KeyExportRateLimit, -1, "Rate limit (per minute) for event export. Set to -1 to disable")
+	flags.Int(KeyExportRateLimit, -1, "Rate limit (per minute) for event export. Set to -1 to disable, 0 to disable JSON export, or a positive integer to rate limit event export")
```
FedeDP:
```suggestion
	flags.Int(KeyExportRateLimit, -1, "Rate limit (per minute) for event export. Set to -1 to disable rate limiting, 0 to disable JSON export, or a positive integer to rate limit event export")
```

pkg/option/flags.go
FedeDP: I think we are a 
* `make vendor`
* `make generate-flags`
* "Signed-off-by" commit

from a :green_circle:  CI !! :smile:

stefanamaerz: Ok I ran make vendor/generate-flags, signed the commit, [incorporated your language suggestions](https://github.com/cilium/tetragon/pull/5300#pullrequestreview-4808923748) and squashed the commits.

Appreciate it!

stefanamaerz: I noticed CI is still failing and saying to run `make generate-flags`.


I reran `make generate-flags` (this time from an x86 VM intead of arm, not sure if that makes a difference):
https://github.com/cilium/tetragon/pull/5300/commits/bd6e01c4175fd60461bc57c0f7f0767c68d7dc0e

FedeDP:
@stefanamaerz care to squash the commits together? Thank you!

stefanamaerz:
> @stefanamaerz care to squash the commits together? Thank you!

@FedeDP done!
https://github.com/cilium/tetragon/pull/5300/changes/01c3a06b7507a3b62d8d4bd3e23099d9dce5a098

FedeDP (approved):
LGTM!
Only thing i'd still do, is to update the commit message to properly state the issue we are fixing (ie: that a wrong value for `KeyExportRateLimit` would otherwise return `0` and silently disable event exports).
Thank you so much!

stefanamaerz:
> LGTM! Only thing i'd still do, is to update the commit message to properly state the issue we are fixing (ie: that a wrong value for `KeyExportRateLimit` would otherwise return `0` and silently disable event exports). Thank you so much!

Commit message updated: https://github.com/cilium/tetragon/pull/5300/changes/a20736365ea36e0319b7e06da6aac660b6098c1b

kkourt (approved):

PR #3825, opened 2025-06-12
Pr/kkourt/kprobe ignore missing call

kkourt:
wip

PR #5301, opened 2026-07-17
uprobe digest partial load

andrewstrohman:
This is the final PR of a series of PRs that implement uprobe binary digest verification. See [this parent PR](https://github.com/cilium/tetragon/pull/4994) for the list of related PRs and the initial feedback I received before breaking that large parent PR into a series of smaller children PRs.

The core uprobe binary digest validation logic in now in the `main` branch. The current behavior is to reject a policy in its entirety if any uprobes within the policy fail digest verification. This PR changes that behavior such that we salvage the uprobes that don't fail verification, and therefore partially load the policy. If all the hooks fail digest verification, the policy will still be considered loaded, even though none of its hooks were attached to.

Since a policy can be partially loaded for the first time, we need a way to communicate which hooks were attached to and which hooks we did not attach to (due to digest verification failure). This PR extends our API's `TracingPolicyStatus` to expose each hook's status. Initially, the hook status will only be exposed for uprobe and kprobe policies, because  these are the only type of policies that can currently have partial loads. We can use this framework to expose other policy type's hook status if/when they need the ability to do partial loading.

Since our API now exposes the hook status, this PR extends the `tetra tracingpolicy list` command by interleaving the hook status within its pre-existing output when a policy is found to be partially loaded.

The following policy and command output demonstrates how this works.
```
kind: TracingPolicy
metadata:
  name: "uprobe-digest-example"
spec:
  uprobes:
    - path: "/home/andy/src/sleep"
      symbols:
      - "main"
      binaryDigests:
      ignore:
        digestVerificationFailure: true
    - path: "/home/andy/src/sleep"
      symbols:
      - "main"
      binaryDigests:
        - "sha256:8bc8c492483d324eecd6a3630e61f43524da590e3cafeb298e43f5d0d63f6814"
    - path: "/home/andy/src/my_program"
      symbols:
      - "main"
      binaryDigests:
        - "sha256:77603eb0809fb6b95c69d79772943a11f18ad22cbd5d0c2efc50171760ff98e4"

```

```
andy@builder:~/src/tetragon$ sudo ./tetra tracingpolicy add ~/uprobe-digest-mismatch.yaml
tracing policy "/home/andy/uprobe-digest-mismatch.yaml" added
andy@builder:~/src/tetragon$ sudo ./tetra tracingpolicy list
ID   NAME                    DOMAIN   STATE               FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE           NPOST   NENFORCE   NMONITOR
2    uprobe-digest-example   grpc     partially_enabled   0          (global)    generic_uprobe   380.79 kB      monitor_only   0       0          0
     SECTION   CFGIDX   DESCRIPTION                 STATUS
     uprobes   0        /home/andy/src/sleep        digest_rejected
     uprobes   1        /home/andy/src/sleep        loaded
     uprobes   2        /home/andy/src/my_program   loaded
```

Tetragon also logs which hooks failed verification when the policy is loaded:
```
level=info msg="spec.uprobes[0]: digest verification failed for path \"/home/andy/src/sleep\""
```


pkg/sensors/tracing/genericuprobe.go
```diff
@@ -860,32 +862,54 @@ func createGenericUprobeSensor(
...
 			return nil, err
 		}
 	}
 
-	if useMulti {
-		progs, maps, err = createMultiUprobeSensor(polInfo, name, ids, has)
-	} else {
-		progs, maps, err = createSingleUprobeSensor(polInfo, ids, has)
-	}
+	if len(ids) != 0 {
```
FedeDP, line 973: This change perhaps could go alone.

pkg/sensors/handler.go
```diff
@@ -458,6 +459,7 @@ func (h *handler) listPolicies(domain string) []*tetragon.TracingPolicyStatus {
 		for _, sens := range col.sensors {
 			pol.Sensors = append(pol.Sensors, sens.GetName())
 			pol.KernelMemoryBytes += uint64(sens.TotalMemlock())
+			sens.SetStatus(&pol)
```
FedeDP: I am wondering whether exposing a `Status()` API would fit better with other already exposed methods; then let handler.go do the `append`.
Eg, if you look at other 2 fields in this `for` loop, we are doing that.

FedeDP (commented):
So, i am bit confused: which is the final state for an uprobe sensor with all its hook in `HookState_STATUS_DIGEST_REJECTED` state?
Would it be `TP_STATE_ENABLED`?

My proposal is to add a new state `PARTIALLY_ENABLED` and:
* all hooks loaded -> TP_STATE_ENABLED
* some hooks loaded -> TP_STATE_PARTIALLY_ENABLED
* no hooks loaded -> TP_STATE_LOAD_ERROR

pkg/sensors/tracing/genericuprobe.go
FedeDP: Oh i see, it is because we now can have an uprobe sensor with all `HookState_STATUS_DIGEST_REJECTED` hooks state, therefore  we should skip the sensor creation.

pkg/sensors/handler.go
andrewstrohman: Thanks, yes, that's better. This approach also helped with you other suggestion, TP_STATE_PARTIALLY_ENABLED.

andrewstrohman:
> So, i am bit confused: which is the final state for an uprobe sensor with all its hook in `HookState_STATUS_DIGEST_REJECTED` state? Would it be `TP_STATE_ENABLED`?

Yes, TP_STATE_ENABLED is the final state in this scenario.

> My proposal is to add a new state `PARTIALLY_ENABLED` and:
> 
> * all hooks loaded -> TP_STATE_ENABLED
> * some hooks loaded -> TP_STATE_PARTIALLY_ENABLED
> * no hooks loaded -> TP_STATE_LOAD_ERROR


I added TP_STATE_PARTIALLY_ENABLED for "some hooks loaded". But, I did not do "no hooks loaded -> TP_STATE_LOAD_ERROR"

Since I am headed towards doing an ignore approach, like what is done [here](https://github.com/cilium/tetragon/pull/3825), this situation does not feel like an error to me, because the policy will be rejected unless the user specified that it's OK to ignore all the hooks. Please let me know if you still think that we should set that state of the policy to TP_STATE_LOAD_ERROR when no hook are attached.

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I deleted the previous comment and reverted back to the behavior that you reviewed, and prompted this discussion, because I need to be able to report hook status and whether the policy can be disabled.

FedeDP:
> Since I am headed towards doing an ignore approach, like what is done https://github.com/cilium/tetragon/pull/3825, this situation does not feel like an error to me, because the policy will be rejected unless the user specified that it's OK to ignore all the hooks. Please let me know if you still think that we should set that state of the policy to TP_STATE_LOAD_ERROR when no hook are attached.

In this scenario, i agree with you since it's the user telling us what to do :) 
Thank you so much!

FedeDP:
The only thing i am not sure of is:
* an uprobe with `ignore.digestVerificationFailure: true` would result in a state `TracingPolicyState_TP_STATE_PARTIALLY_ENABLED` when at least an hook gets rejected
* a kprobe with `ignore.callNotFound: true` gets set to `TracingPolicyState_TP_STATE_ENABLED` in any case, even when some of its hooks were not found

Do we want to unify the behavior? (Just a discussion, it can be done in subsequent PR of course since it is tangent to this PR)

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -920,38 +930,65 @@ func createGenericUprobeSensor(
...
+			// The file for that descriptor is closed after sensor load, so we cannot
+			// allow the sensor to be enabled (re-loaded) after being disabled
+			disableNotAllowedReason = disableNotAllowedReasonBinaryDigests
 			entryFile, err = getOrOpenFile(absPath, openedFiles)
 			if err != nil {
 				return nil, err
 			}
 		}
 
+		hook_status := &tetragon.HookStatus{
```
olsajiri: we don't use _ in variable name

olsajiri (approved):
I can't really say the HookState is the best way to carry this info, but the ignore part makes sense to me, left one comment, thanks

FedeDP (approved):
Patch LGTM aside the concern expressed in https://github.com/cilium/tetragon/pull/5301#issuecomment-5031384932 that IMO is not meant to be fixed in this PR.
Thank you very much!

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Thanks for pointing this out. I've fixed it.

andrewstrohman:
> I can't really say the HookState is the best way to carry this info

How can I improve it? Are you saying that you don't like the structure of the schema that was added to the API, or that we should use a separate, internal-only data structure during sensor construction, and then translate that to HookState only for the API?

andrewstrohman:
> The only thing i am not sure of is:
> 
> * an uprobe with `ignore.digestVerificationFailure: true` would result in a state `TracingPolicyState_TP_STATE_PARTIALLY_ENABLED` when at least an hook gets rejected
> * a kprobe with `ignore.callNotFound: true` gets set to `TracingPolicyState_TP_STATE_ENABLED` in any case, even when some of its hooks were not found
> 
> Do we want to unify the behavior? (Just a discussion, it can be done in subsequent PR of course since it is tangent to this PR)

Yes, I want to unify it. I've added some commits to this PR to do that. Unlike uprobes, where a hook either fails to attach in its entirety or not, kprobes can have a partial attach per hook. This can happen when the kprobe's `call` configuration specifies a list, and at least one attach point in the list is not found (and at least one attach point is found).

This PR's proposed status schema is not expressive enough to communicate which call(s) in the list were ignored. I'm hesitant to introduce another layer/level of status to expose this until we find that users need this level of detail. But, I'd like to hear other's opinion on this. uprobe policies can generate multiple probes per hook when multiple addrs/offset/symbols are configured. So, it's possible (but unlikely IMO) that we could have a situation in the future where we partially load an individual uprobe hook.

andrewstrohman:
Here is what kprobes look like after unifying:

The policy:
```
kind: TracingPolicy
metadata:
  name: "pizza-and-bpf"
spec:
  kprobes:
  - call: "pizza_is_the_best"
    syscall: false
    ignore:
      callNotFound: true
  - call: "sys_bpf"
    syscall: true
```

```
andy@builder:~/src/tetragon$ sudo ./tetra tracingpolicy add ~/kprobe-call-not-found.yaml
tracing policy "/home/andy/kprobe-call-not-found.yaml" added
andy@builder:~/src/tetragon$ sudo ./tetra tracingpolicy list
ID   NAME            DOMAIN   STATE               FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE           NPOST   NENFORCE   NMONITOR
2    pizza-and-bpf   grpc     partially_enabled   0          (global)    generic_kprobe   506.32 kB      monitor_only   9       0          0
andy@builder:~/src/tetragon$ sudo ./tetra tracingpolicy list --hook-status
ID   NAME            DOMAIN   SECTION   CFGIDX   DESCRIPTION         STATUS
2    pizza-and-bpf   grpc     kprobes   0        pizza_is_the_best   call_not_found
                              kprobes   1        sys_bpf             loaded
```

FedeDP:
So, i now want a :pizza: :laughing: 
> Here is what kprobes look like after unifying:

I think it looks great, actually; this is what i had in mind. Thanks for doing that.
> This PR's proposed status schema is not expressive enough to communicate which call(s) in the list were ignored. I'm hesitant to introduce another layer/level of status to expose this until we find that users need this level of detail

I agree with you, better not invest too much effort on this until we know there is a real use case for it.

FedeDP (approved):

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -920,38 +930,65 @@ func createGenericUprobeSensor(
...
-			return nil, fmt.Errorf("spec.uprobes[%d]: %w", cfgIdx, err)
+			mismatchErr, ok := errors.AsType[*DigestMismatchError](err)
+			if !ok || !ignoreDigestVerificationFailure(&uprobe) {
+				return nil, fmt.Errorf("spec.uprobes[%d]: %w", cfgIdx, err)
+			}
+
+			hookStatus.State = tetragon.HookState_STATUS_DIGEST_REJECTED
+			logger.GetLogger().Info(
+				fmt.Sprintf("spec.uprobes[%d]: %s", cfgIdx, mismatchErr.Error()),
+			)
```
kkourt, line 962: Should this be a `Debug`? Not sure if it's an error condition or not.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -920,38 +930,65 @@ func createGenericUprobeSensor(
...
 			return nil, err
 		}
 	}
 
-	if useMulti {
-		progs, maps, err = createMultiUprobeSensor(polInfo, name, ids, has)
-	} else {
-		progs, maps, err = createSingleUprobeSensor(polInfo, ids, has)
-	}
+	if len(ids) != 0 {
```
kkourt, line 973: Just curious, is this check in case all hooks were skipped?

kkourt (approved):
LGTM, thanks!

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: I'd like to keep this as Info. When I spoke with users about what they'd like to see when a partial load occurs, they said they want to see that it happened in tetragon's logs. IMHO, I think having this in the log would help us debug as well.

Logging at info level is consistent with how ignored kprobes are handled: https://github.com/cilium/tetragon/blob/8d10f87100a3f8d932625f0ff3270ccc20319852/pkg/sensors/tracing/generickprobe.go#L420

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: Yes, this for when all hooks are skipped. I'm trying to avoid loading programs/maps in this case. Without this condition, this happens:
```
level=warn msg="Server AddTracingPolicy request failed" error="sensor generic_uprobe from collection uprobe-digest-empty failed to load: tetragon, aborting could not load sensor BPF maps: failed to load map 'config_map' for sensor 'generic_uprobe': failed to create map 'config_map': creating map: map create: invalid argument (MaxEntries may be incorrectly set to zero)" metadata.namespace="" metadata.name=uprobe-digest-empty
```

So instead of make that code cope, I opted to prevent it from running, because loading those program/maps is of no use.

This ties into [a discussion](https://github.com/cilium/tetragon/pull/5301#pullrequestreview-4733133823) about whether a policy should be considered TP_STATE_LOAD_ERROR when all of the hooks are ignored. Folks that advocated for TP_STATE_LOAD_ERROR previously, seem to be OK with TP_STATE_PARTIALLY_ENABLED now, because the user is signaling intent with their `ignore` configurations.

I wanted to create the empty sensor because that's the way that I communicate which hooks were attached (none in this case).  But the consequence of this is that I needed to add and check additional state for the sensor in order to prevent spurious error logs. See the last commit of this series for what that entails.  It's also worth noting that I changed kprobes to do this too, for the same reason.

pkg/sensors/tracing/genericuprobe.go
kkourt: OK, let's keep it.

pkg/sensors/collection.go
```diff
@@ -136,8 +139,29 @@ func (c *collection) info() string {
...
+func (c *collection) isEnabled() bool {
+	switch c.state {
+	case EnabledState, PartiallyEnabledState:
+		return true
+	default:
+		return false
+	}
+}
+
+func (c *collection) getEnabledState() TracingPolicyState {
```
mtardy: maybe I would put a name a bit more explicit that states it could return partially enabled in some situations.

cmd/tetra/common/utils.go
```diff
@@ -29,10 +29,13 @@ func HumanizeByteCount(b uint64) string {
...
 }
 
-func PrintTracingPolicies(output io.Writer, policies []*tetragon.TracingPolicyStatus, skipPolicy func(pol *tetragon.TracingPolicyStatus) bool) {
+func PrintTracingPolicies(output io.Writer, policies []*tetragon.TracingPolicyStatus, hookStatus bool, skipPolicy func(pol *tetragon.TracingPolicyStatus) bool) {
 	// tabwriter config imitates kubectl default output, i.e. 3 spaces padding
 	w := tabwriter.NewWriter(output, 0, 0, 3, ' ', 0)
 	header := "ID\tNAME\tDOMAIN\tSTATE\tFILTERID\tNAMESPACE\tSENSORS\tKERNELMEMORY\tMODE\tNPOST\tNENFORCE\tNMONITOR"
+	if hookStatus {
+		header = "ID\tNAME\tDOMAIN\tSECTION\tCFGIDX\tDESCRIPTION\tSTATUS"
+	}
```
mtardy: Could we do better and instead of having the manual `--hook-status` print the hook status when you hit  a partially enabled TP?

Like for a partially enabled TP you maybe inline this new tab print at indentation +1? I wonder if it's too complicated or just doable with tabwriter.

```
andy@builder:~/src/tetragon$ sudo ./tetra tracingpolicy list
ID   NAME                    DOMAIN   STATE               FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE           NPOST   NENFORCE   NMONITOR
2    uprobe-digest-example   grpc     partially_enabled   0          (global)    generic_uprobe   380.79 kB      monitor_only   0       0          0
     SECTION   CFGIDX   DESCRIPTION                 STATUS
     uprobes   0        /home/andy/src/sleep        digest_rejected
     uprobes   1        /home/andy/src/sleep        loaded
     uprobes   2        /home/andy/src/my_program   loade
3    uprobe-othero-example   grpc     enabled             1          (global)    generic_uprobe   380.79 kB      monitor_only   0       0          0

```

api/v1/tetragon/sensors.proto
```diff
@@ -111,6 +111,10 @@ enum HookState {
   STATUS_LOADED = 1;
   // probe was not loaded due to digest mismatch
   STATUS_DIGEST_REJECTED = 2;
+  // probe was not loaded due to call was not found
+  STATUS_CALL_NOT_FOUND = 3;
+  // probe was partially loaded due to call not found
+  STATUS_PARTIAL_CALL_NOT_FOUND = 4;
```
mtardy, line 117: I don't think this enum entry belongs here but on the upper TP status?

pkg/sensors/tracing/generickprobe.go
```diff
@@ -617,6 +631,11 @@ func createGenericKprobeSensor(
 			}
 			ids = append(ids, id)
 		}
+		if valInfo[i].ignoredCount == 0 {
+			hookStatus.State = tetragon.HookState_STATUS_LOADED
+		} else {
+			hookStatus.State = tetragon.HookState_STATUS_PARTIAL_CALL_NOT_FOUND
+		}
```
mtardy, line 638: I'm not following why it's not applied on the TP status not hookStatus

mtardy (commented):
Docs lgtm, just a few questionns on the ` STATUS_PARTIAL_CALL_NOT_FOUND = 4;` that I might misunderstand.

pkg/sensors/tracing/generickprobe.go
andrewstrohman: There is some disussion/explanation about this in [this comment](https://github.com/cilium/tetragon/pull/5301#issuecomment-5039470269).

When a kprobe hook specifies a list for `call:`, then multiple kprobes are generated, one for each function in the list. Some of those functions may not be found while others are. When this happens, we can have a partial load at the hook level too.

api/v1/tetragon/sensors.proto
andrewstrohman: This is intentional. When a kprobe's hook's `call` config specifies a list, the hook itself can be partially loaded. This is related to your other comment as well: https://github.com/cilium/tetragon/pull/5301#discussion_r3659842708

pkg/sensors/collection.go
andrewstrohman: I've change this to `getEnabledStateType()`.  If you can think of a better name, please suggest it, and I'll change it.

cmd/tetra/common/utils.go
andrewstrohman: Ok, I was able to do this by using multiple tabwriters, and manually interleaving the hook status:

```
andy@builder:~/src/tetragon$ sudo ./tetra tracingpolicy list
ID   NAME                    DOMAIN   STATE               FILTERID   NAMESPACE   SENSORS          KERNELMEMORY   MODE           NPOST   NENFORCE   NMONITOR
2    uprobe-digest-example   grpc     partially_enabled   0          (global)    generic_uprobe   380.79 kB      monitor_only   0       0          0
     SECTION   CFGIDX   DESCRIPTION                 STATUS
     uprobes   0        /home/andy/src/sleep        digest_rejected
     uprobes   1        /home/andy/src/sleep        loaded
     uprobes   2        /home/andy/src/my_program   loaded
3    pizza-and-bpf           grpc     partially_enabled   0          (global)    generic_kprobe   0 B            unknown        0       0          0
     SECTION   CFGIDX   DESCRIPTION         STATUS
     kprobes   0        pizza_is_the_best   call_not_found
4    uprobe-selectors        grpc     enabled             0          (global)    generic_uprobe   380.79 kB      monitor_only   0       0          0
```

While I don't think this is a problem, note how the pizza-and-bpf policy has different hook status column widths compared to the uprobe-digest-example hook status.

mtardy: wow nice

mtardy (approved):
thanks!!

PR #5303, opened 2026-07-18
fix(elf): handle zero-argument USDT probes

immanuwell:
### Description

`tetra tracingpolicy generate usdts` treats a zero argument USDT probe as one argument with an empty type. 
Generated policy then fails validation

Repro on `main`:

```sh
tmp=$(mktemp -d)
gcc -Wall contrib/tester-progs/usdt.c -o "$tmp/usdt"
go run ./cmd/tetra tracingpolicy generate usdts -b "$tmp/usdt" | sed -n '/name: usdt0/,+8p'
```

`usdt0` entry gets `type: ""`. This uses `strings.Fields` so no bogus arg is emitted

Tested with `go test ./pkg/elf`

### Changelog

```release-note
Fix generated tracing policies for zero argument USDT probes
```

### DIFF
--- a/pkg/elf/usdtargs_linux.go
+++ b/pkg/elf/usdtargs_linux.go
@@ -33,7 +33,7 @@ func compactDeref(ArgsStr string) string {
 
 // Arg parsing entrypoint
 func parseArgs(spec *UsdtSpec) error {
-	for idx, str := range strings.Split(compactDeref(spec.ArgsStr), " ") {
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
Great fix thanks!
I am not sure whether adding a test is worth considering we are basically testing string.Fields(), but it's not a big deal.

kkourt (approved):

PR #5304, opened 2026-07-20
docs: add section for CelExpr operator

kkourt:


docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -182,6 +183,49 @@ selectors:
...
+    selectors:
+      - matchArgs:
+        - operator: "CelExpr"
+          values:
+          - "arg1 == int32(-1) && arg0 >= int32(1024)"
+```
+
+Arguments, as specified in the `args:` list, are made available in the CEL expression as `argX`
+where `X` is the zero-based index in the list. In the above example, the argument labeled `whence`
+is available as `arg0`, while the argument labeled `fd` is available as `arg1`.
```
mtardy, line 221: Mmh then I just realized we are not respected the explicits `index:` but the ordering of appearance. I guess this was discussed when implementing this? I feel like it would have been more natural to reuse those.

mtardy (approved):
Docs lgtm! Thanks! Just a small remark on the feature itself.

docs/content/en/docs/concepts/tracing-policy/selectors.md
kkourt: The problem with using the `index:` is that you can have two args using the same index (with potentially different resolve sections). See: https://github.com/cilium/tetragon/issues/3711.

An improvement here would be to use the label names (e.g., `arg_whence`), but that's not yet implemented.

mtardy: aah okay didn't remember that detail thanks

PR #5312, opened 2026-07-21
more policytest updates

kkourt:


FedeDP (approved):

tests/policytests/kprobes.go
```diff
@@ -28,12 +28,20 @@ import (
...
 
 var _ = policytest.NewBuilder("kprobe-lseek").WithLabels("kprobes").
-	WithParameter(policytest.Parameter{
-		Name:    "Hook",
-		Default: "kprobes",
-		Values:  []any{"kprobes", "fentries"},
-		Help:    "type of hook to use in the policy",
-	}).WithPolicyTemplate(`
+	WithSkip(func(si *policytest.SkipInfo) string {
+		if si.ParamValues != nil && si.ParamValues["Hook"] == "fentries" {
```
andrewstrohman: Why are we checking `si.ParamValues != nil`? This will always evaluate to `true`, right? Without the parameter, the policy template cannot be rendered.

andrewstrohman (approved):

tests/policytests/kprobes.go
kkourt: makes sense. Fixed.

PR #5321, opened 2026-07-22
fix(tetra): read large stdin event lines

immanuwell:
### Description

`tetra getevents` stops at 64 KiB in stdin mode.
This lets it read lines up to the configured receive limit and covers a real `char_buf.maxData` kprobe event

Repro:

```sh
( printf '%s' '{"process_kprobe":{"args":[{"bytes_arg":"'; head -c 70000 /dev/zero | base64 -w 0; printf '%s\n' '"}]}}' ) | tetra getevents
```

Before this stops with `bufio.Scanner: token too long`. After this the event is read

### Changelog

```release-note
Fix tetra getevents stdin mode for large event lines.
```

### DIFF
--- a/cmd/tetra/getevents/io_reader_client.go
+++ b/cmd/tetra/getevents/io_reader_client.go
@@ -14,6 +14,7 @@ import (
 	"google.golang.org/protobuf/encoding/protojson"
 
 	"github.com/cilium/tetragon/api/v1/tetragon"
+	"github.com/cilium/tetragon/cmd/tetra/common"
 	"github.com/cilium/tetragon/pkg/event"
 	"github.com/cilium/tetragon/pkg/fieldfilters"
 	"github.com/cilium/tetragon/pkg/filters"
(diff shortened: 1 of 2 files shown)


FedeDP (approved):
LGTM thanks for the fix!

tpapagian (approved):

cmd/tetra/getevents/io_reader_client.go
```diff
@@ -31,8 +32,12 @@ type ioReaderClient struct {
 }
 
 func newIOReaderClient(reader io.Reader, debug bool) *ioReaderClient {
+	scanner := bufio.NewScanner(reader)
+	maxEventSize := max(bufio.MaxScanTokenSize, common.MaxRecvMsgSize)
+	scanner.Buffer(make([]byte, bufio.MaxScanTokenSize), maxEventSize)
```
mtardy: nit: Just from `Buffer` doc, it seems it's already the default anyway
> // By default, [Scanner.Scan] uses an internal buffer and sets the
> // maximum token size to [MaxScanTokenSize].
```suggestion
	scanner.Buffer(nil, max(bufio.MaxScanTokenSize, common.MaxRecvMsgSize))
```

mtardy (approved):

immanuwell:
@mtardy applied your suggestion

mtardy:
Try also to include the reason of the changes (basically what's in the PR description) in the commit log also next time! :) thanks

PR #5322, opened 2026-07-22
new(nok8s): json schema validation for tracingpolicies

FedeDP:
### Description

Since in nok8s builds we can't rely upon kubernetes validation, we were lacking any sort of validation for the tracing policy yaml object.
This patch introduces a validation based upon the json schema of the crds that were previously generated and used only in the `kubeconform` make target.

New dependency: github.com/santhosh-tekuri/jsonschema/v6; only used in nok8s build.
Size change:
* main: 49532
* HEAD: 50752

~800K.

Most of the line changes come from:
* the new `pkg/tracingpolicy/schemas/` json schemas
* the new vendored dependency

### Changelog

```release-note
new(nok8s): json schema + CEL validation for tracingpolicies
```


FedeDP:
> [Run static checks / build-nok8s (pull_request)](https://github.com/cilium/tetragon/actions/runs/29928267752/job/88950933665?pr=5322)Failing after 1m

So, this is failing because in the `tetragon-nok8s` target i clean up the `pkg/tracingpolicy/schemas/` folder, but then `tetra-nok8s` fails to build because `go:embed` can't find the file.
2 possibilities:
* we keep the `install/kubernetes tetragon/schemas/` folder alive (and we re-generate it in `generate` target); then both `tetragon-nok8s` and `tetra-nok8s` will rely upon a `pkg/tracingpolicy/schemas/` target that just `cp -r install/kubernetes/tetragon/schemas/ pkg/tracingpolicy/schemas/`; finally, we gitignore `pkg/tracingpolicy/schemas/`.
* both `tetra-nok8s` and `tetragon-nok8s` will depend upon `pkg/tracingpolicy/schemas/` target that generates json schemas from crds, and both targets clean `pkg/tracingpolicy/schemas/` folder up once they are done

For now, i implement 1, but am open to switch to 2 if requested.

EDIT: in the end, i chose to follow what we already do for crds: `make generate` will re-generate the files directly in the desired folder (`pkg/tracingpolicy/schemas/`) and they are then versioned. This allows `go test -tags nok8s` or just `go build -tags nok8s` to work like a charm since the required files are correctly there.

FedeDP:
Example output from feeding a wrong syntax policy:
```
level=error msg="Failed to execute tetragon" error="failed to validate TracingPolicy: jsonschema validation failed with 'file:///tracingpolicy#'\n- at '/spec': additional properties 'lsmhoooks' not allowed"
```
In main, same policy:
```
level=info msg="Added TracingPolicy with success" TracingPolicy=/home/fdipierr/pol-lsm-wrong.yaml metadata.namespace="" metadata.name=protection-policy-cve-2024-21626
```

Policy file (note `lsmhoooks` :D )
```yaml
kind: TracingPolicy
metadata:
  name: "protection-policy-cve-2024-21626"
spec:
  lsmhoooks:
  - hook: "bprm_check_security"
    args:
    - index: 0 # struct linux_binprm *bprm
      type: "path"
      resolve: "mm.owner.fs.pwd"
    selectors:
    - matchArgs:
      - index: 0
        operator: "Prefix"
        values:
        - "/sys/fs/cgroup"
      matchActions:
      - action: Post
    message: cve-2024-21626
```

vendor/github.com/santhosh-tekuri/jsonschema/v6/.swp
andrewstrohman, line 1: I confirmed that this came from the vendor's repo, and it's not something that we created by accident. (I guess our CI would catch it, if it came from us).

pkg/tracingpolicy/nok8s.go
```diff
@@ -22,8 +50,15 @@ func FromYAML(data string) (TracingPolicy, error) {
 
 	switch kind {
 	case v1alpha1.TPKindDefinition:
+		inst, err := jsonschema.UnmarshalJSON(bytes.NewReader(jsonBytes))
+		if err != nil {
+			return nil, fmt.Errorf("failed to unmarshal JSON: %v", err)
```
andrewstrohman: Would `%w` be better than `%v` here?

pkg/tracingpolicy/nok8s.go
```diff
@@ -6,13 +6,41 @@
...
+	tpSchema                *jsonschema.Schema
+)
+
+var _ jsonschema.URLLoader = &MemLoader{}
+
+type MemLoader struct{}
+
+func (m *MemLoader) Load(url string) (any, error) {
+	switch url {
+	case "file:///tracingpolicy":
```
andrewstrohman: If "file:///tracingpolicy" was stored as a const variable, you could use it in this case statement and also below in `c.MustCompile("file:///tracingpolicy")`.  This way, we cannot have drift.

andrewstrohman:
It might be worth doing an explicit validation for drift between the JSON schema and the CRDs in CI.

andrewstrohman (approved):

pkg/tracingpolicy/nok8s.go
FedeDP: Right, thanks!

pkg/tracingpolicy/nok8s.go
FedeDP: Makes sense, thank you!

FedeDP:
> It might be worth doing an explicit validation for drift between the JSON schema and the CRDs in CI.

This should never happen since `make generate` takes care of rebuilding both.

pkg/tracingpolicy/nok8s.go
```diff
@@ -6,13 +6,43 @@
...
+	case tpURL:
+		return jsonschema.UnmarshalJSON(bytes.NewReader(tracingpolicyJsonSchema))
+	}
+	return nil, fmt.Errorf("wrong loader url: %s", url)
+}
+
+func init() {
+	c := jsonschema.NewCompiler()
+	c.UseLoader(&MemLoader{})
+	tpSchema = c.MustCompile(tpURL)
```
kkourt, line 55: How much memory does adding this add to the agent?

kkourt (commented):
> Note: x-kubernetes-validations cel expressions are not supported; we would need to extend the json schema validation and manually support them; it's not super easy to do, but it would allow us to cleanup some code.

Is this something that we plan to do?

Because if we can't do it, then I think the best option would be to do just duplicate the validation logic in go since we are using CEL expressions extensively in our spec.

pkg/tracingpolicy/nok8s.go
FedeDP: Main: `70252`
HEAD: `73168`

FedeDP:
> Is this something that we plan to do?

I think we might want to commit to do that in the future; unfortunately, it is a bit complex and would take quite a bit of time.

> Because if we can't do it, then I think the best option would be to do just duplicate the validation logic in go since we are using CEL expressions extensively in our spec.

I am not sure, it quickly becomes risky to rely upon the double validation (sooner or later we would forget to add it) and also why pay the perf penalty (for validation) in k8s builds when k8s already validates for us?

pkg/tracingpolicy/nok8s.go
```diff
@@ -6,13 +6,43 @@
...
+
 	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
 	"github.com/cilium/tetragon/pkg/nok8s"
 )
 
+const tpURL = "file:///tracingpolicy"
+
+var (
+	//go:embed schemas/tracingpolicy-cilium.io.json
+	tracingpolicyJsonSchema []byte
```
mtardy: nit: should be JSON and not Json (like UnmarshalJSON, etc.)

go.mod
```diff
@@ -31,6 +31,7 @@ require (
 	github.com/prometheus/client_model v0.6.2
 	github.com/prometheus/common v0.70.0
 	github.com/prometheus/procfs v0.21.1
+	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2
```
mtardy, line 34: how much do we trust this new dep?

install/kubernetes/Makefile
```diff
@@ -113,11 +116,7 @@ check-images-tags: ## Verify all image:tag references in values.yaml exist in th
 # CRD. Skip validating the Tetragon CRDs themselves (circular dependency).
 # Skip the cert-manager Certificate too: its schema belongs to cert-manager and
 # is not available here.
-kubeconform: ## Validate Helm chart using kubeconform
-kubeconform:
-	mkdir -p $(JSON_SCHEMAS)/ $(SCHEMAS_CACHE)/
-	$(PYTHON) /bin/bash -c "pip install pyyaml && python /code/install/kubernetes/openapi2jsonschema.py /code/$(CRDS_RELATIVE_DIR)/*"
-	mv $(ROOT_DIR)/install/kubernetes/*-cilium.io.json $(JSON_SCHEMAS)/
+kubeconform: $(TETRAGON_CHART)/schemas ## Validate Helm chart using kubeconform
```
mtardy, line 122: could you explain a little bit in the git commit log the why of these changes? I'm not super familiar on why we were doing all this before

pkg/tracingpolicy/nok8s.go
mtardy: Is this KB? so 3 MB for this?

FedeDP: Exactly

go.mod
FedeDP: Seems widely used and updated. But can't really say :/

install/kubernetes/Makefile
FedeDP: > I'm not super familiar on why we were doing all this before

I'm not either :) i just split the schemas generation code from being part of the `kubeconform` target to its own.
(I just restored the `SCHEMAS_CACHE` thing because it's still useful).
I expanded the commit message to be more precise :) 

Also, note that i tested `make kubeconform` and it works fine.

kkourt:
> > Is this something that we plan to do?
> 
> I think we might want to commit to do that in the future; unfortunately, it is a bit complex and would take quite a bit of time.
> 
> > Because if we can't do it, then I think the best option would be to do just duplicate the validation logic in go since we are using CEL expressions extensively in our spec.
> 
> I am not sure, it quickly becomes risky to rely upon the double validation (sooner or later we would forget to add it)

What's the alternative? Note that my statement is qualified by "if we can't do it" (i.e., translate the CEL validation to a json schema).

> and also why pay the perf penalty (for validation) in k8s builds when k8s already validates for us?

I would, also, expect that performance overhead of this is  small. Indeed, I would expect that it's negligible compared to how long it takes to load a policy.

pkg/tracingpolicy/nok8s.go
kkourt: Thanks! Can you add this information in a commit message?
3MB is not a huge amount, but it's not small either. We might want to consider in the future unload the data from the cache after a given timeout (no need to do it in this PR).

FedeDP:
> Note that my statement is qualified by "if we can't do it" (i.e., translate the CEL validation to a json schema).

I think we should be able to do it, but it's a bit complex; i can give it a shot and keep this PR on hold until i find the right way to achieve that if you prefer.

pkg/tracingpolicy/nok8s.go
FedeDP: Already added :)
> We might want to consider in the future unload the data from the cache after a given timeout (no need to do it in this PR).

Might be good, yes!

kkourt:
> > Note that my statement is qualified by "if we can't do it" (i.e., translate the CEL validation to a json schema).
> 
> I think we should be able to do it, but it's a bit complex

OK, good enough for me :)

> i can give it a shot and keep this PR on hold until i find the right way to achieve that if you prefer.


IMO, no need to put the PR on hold. Let's just go over the review process as normal and merge it (assuming there are on blockers). Once merged, please create an issue to track the followup work.

FedeDP:
@kkourt ooops i think i am really near to make CEL validation work without many changes :rocket: ; pushed the changes in their own commit to keep them separated with their own commit message.

It was easier than i thought, i just needed to dig a bit in the jsonschema library.

FedeDP:
Moving to draft to fix CEL stuff.

FedeDP:
Fixed CEL stuff.

install/kubernetes/Makefile
```diff
@@ -47,6 +47,7 @@ SCHEMAS_CACHE := $(ROOT_DIR)/install/kubernetes/schemas-cache
 PYTHON := docker run --rm \
 		-v $(ROOT_DIR):/code \
 		--workdir=/code/install/kubernetes \
+		--network host \
```
andrewstrohman: should this be in a separate commit?

andrewstrohman (approved):

install/kubernetes/Makefile
FedeDP: Oh gosh, it's just me having issues with my docker install :/ let me drop that thing.
Great catch, thank you so much!

pkg/tracingpolicy/nok8s.go
```diff
@@ -36,11 +41,94 @@ func (m *memLoader) Load(url string) (any, error) {
...
+				return nil, nil
+			}
+			var celRules []celRule
+			for _, r := range rules {
+				m, ok := r.(map[string]any)
+				if !ok {
+					continue
+				}
+				ruleStr, _ := m["rule"].(string)
+				msg, _ := m["message"].(string)
```
mtardy: q, checking: so here we are explicitly ignoring the error in case the type is incorrect, is it expected?

`ruleStr := m["rule"].(string)` would panic, but here we'll just have an empty string in that case, should you want to return an error or just ignore it?

pkg/tracingpolicy/k8s_xvalidation_test.go
```diff
@@ -23,6 +24,7 @@ spec:
 `
 	_, err := FromYAML(crd)
 	require.Error(t, err)
+	require.Contains(t, err.Error(), "ReturnArg not specified with Return=true.")
```
mtardy, line 27: maybe those ones belong in a separate patch?

mtardy (approved):
thanks I think overall lgtm! 

Never a big fan of adding new deps to the projects, especially ones that belong to one individual author but I if the feature is worth it!

pkg/tracingpolicy/nok8s.go
FedeDP: I think that, given how `x-kubernetes-validations` are added to the json schema, we'll never end up with a non-string `rule` or `message`:
```
"x-kubernetes-validations": [
                          {
                            "message": "override action needs at most one of argNewSymbol, argNewAddr or argNewOffset defined",
                            "rule": "!(has(self.argNewSymbol) && has(self.argNewAddr)) && !(has(self.argNewSymbol) && has(self.argNewOffset)) && !(has(self.argNewAddr) && has(self.argNewOffset))"
                          }
                        ],
```
Anyway, i could add a warning in case that happens; better safe than sorry i guess :D

FedeDP:
@mtardy fixed your suggestions, PTAL :pray:

go.mod
kkourt: I'm also concerned about that. It seems to be widely used, but most of the projects that use it in the dependency graph seem to have 0 stars/forks: https://github.com/santhosh-tekuri/jsonschema/network/dependents

Some notable exceptions:
* https://github.com/apache/doris
* https://github.com/redpanda-data/console
* https://github.com/elastic/go-elasticsearch

Would https://github.com/google/jsonschema-go fit our purposes?

kkourt: One thought that came to mind is to do the validation after dropping privileges/capabilities. This is not easy, and AFAICT the best way to achieve it would be by executing a binary (it can even be `tetragon`).

FedeDP: > Would https://github.com/google/jsonschema-go fit our purposes?

After a quick look, i don't think it supports registering custom vocabulary and validate in a single step like we do with `jsonschema` (very elegantly).
And i'd really love to avoid the huge "take all `x-kubernetes-validations` keyword values in the json schema, then for each build the context (ie: which strucutre they are attached), and finally manually execute all cel programs with their own context after the json schema validation".
It becomes a real mess :) And it was the thing i thought we had to do (and that's why i said "it could be done but it's really hard) before discovering about the `custom vocabulary` feat of https://github.com/santhosh-tekuri/jsonschema.

IMO the 3 "notable exceptions" you provided are pretty big.

FedeDP: > This is not easy, and AFAICT the best way to achieve it would be by executing a binary (it can even be tetragon).

:scream:

kkourt: Thanks Federico! I think it's fine. We can always revisit in the future.

pkg/tracingpolicy/nok8s.go
```diff
@@ -36,11 +41,100 @@ func (m *memLoader) Load(url string) (any, error) {
...
+			}
+			var celRules []celRule
+			for _, r := range rules {
+				m, ok := r.(map[string]any)
+				if !ok {
+					continue
+				}
+				ruleStr, isStr := m["rule"].(string)
+				if !isStr {
+					logger.GetLogger().Warn("invalid x-kubernetes-validations rule field type", "rule", r)
```
kkourt, line 108: Should we just `continue` here? Maybe this means that we  can avoid the `if ruleStr != ""{` below?

pkg/tracingpolicy/nok8s.go
```diff
@@ -36,11 +41,100 @@ func (m *memLoader) Load(url string) (any, error) {
...
+				}
+				msg, isStr := m["message"].(string)
+				if !isStr {
+					logger.GetLogger().Warn("invalid x-kubernetes-validations message field type", "rule", r)
```
kkourt, line 112: Ditto: should we `continue` here?

kkourt (approved):
Thanks! LGTM, just a small nit.

pkg/tracingpolicy/nok8s.go
FedeDP: Yeah so the idea was to avoid "spaghetti code" here, since we are already guarded below, as you noticed.

pkg/tracingpolicy/nok8s.go
FedeDP: Same comment as above applies; moreover, if message is empty but the rule applies, we might still want to enforce the rule, with an emtpy message.
What i think we might do here is set `msg` to a fallback, eg: `CEL rule %s failed to apply"` or something similar.

PR #5326, opened 2026-07-23
policytests: fix policytest run

andrewstrohman:
Prior to this change, when we ran "tetra policytest run kprobe-btf-type-module-af-alg-bind", it resulted in:

panic: please call TestSensorsRun() to initialize GetTestSensorsConf

### DIFF
--- a/tests/policytests/kprobes.go
+++ b/tests/policytests/kprobes.go
@@ -22,7 +22,6 @@ import (
 	sm "github.com/cilium/tetragon/pkg/matchers/stringmatcher"
 	tpath "github.com/cilium/tetragon/pkg/reader/path"
 	"github.com/cilium/tetragon/pkg/testutils/policytest"
-	tus "github.com/cilium/tetragon/pkg/testutils/sensors"
 )
 
 // This file contains tests on kernel functions.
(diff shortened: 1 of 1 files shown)


tests/policytests/kprobes.go
```diff
@@ -273,7 +272,7 @@ spec:
 		checker := ec.NewProcessKprobeChecker("").
 			WithFunctionName(sm.Full("security_socket_bind")).
 			WithProcess(ec.NewProcessChecker().
-				WithBinary(sm.Suffix(tus.Conf().SelfBinary))).
+				WithBinary(sm.Suffix(filepath.Base(os.Args[0])))).
```
kkourt, line 275: As an aside, that's abusing policytests somewhat, because the idea was to use a binary to trigger the poilcy.

kkourt (approved):

tests/policytests/kprobes.go
andrewstrohman: Thanks, I want to make sure I understand what is meant by this.

Currently the trigger is run locally, and in this case, by the `tetra` process itself. But this method won't work in the future when we run the trigger on the same machine as the remote agent under test. When we make the trigger run on the remote, we will have to use separate binary that is installed on the remote, and change this to reflect that binary's name?

PR #5329, opened 2026-07-23
pkg/bugtool: avoid pprof logs without gops

### LINKED ISSUE #5308: bugtool: messages regarding gops
Running bugtool on a tetragon agent using default flags results in:

```
level=info msg="Skipping gops dump info as daemon is running without gops, use --gops-address to enable gops"
level=info msg="Skipping gops dump info as daemon is running without gops, use --gops-address to enable gops"
level=info msg="Successfully dumped gops pprof" profile=cpu
level=info msg="Skipping gops dump info as daemon is running without gops, use --gops-address to enable gops"
level=info msg="Successfully dumped gops pprof" profile=heap
```

There are two potential improvements here:
* Only output the "Skipping gops" message once
* The "Successfully dumped ..." messages are inaccurate. See:
     * https://github.com/cilium/tetragon/blob/02adb23b1cc883867cbcfa08654154b7605ddf55/pkg/bugtool/bugtool.go#L598-L601
     * https://github.com/cilium/tetragon/blob/02adb23b1cc883867cbcfa08654154b7605ddf55/pkg/bugtool/bugtool.go#L633-L642

Ehtasham-Yasin:
Fixes #5308

### Description

Skip pprof collection when Tetragon has no gops address. This leaves the
single skip message from the existing gops collection path and prevents false
success messages for CPU and heap profiles.

Add a focused regression test for both reported log problems.

Checks:

- `go test ./pkg/bugtool -run '^TestGopsDisabledLogsOnce$' -count=1`
- Non-BPF bugtool unit tests
- `golangci-lint run ./pkg/bugtool/...`
- `gofmt -d pkg/bugtool/bugtool.go pkg/bugtool/bugtool_test.go`
- `git diff --check`

The full `go test ./pkg/bugtool` command could not complete in the fresh
Linux container because the unrelated `TestFindMaps/BaseSensorMemlock` test
requires the generated `bpf/objs/bpf_exit_v511.o` object, which was not
present.

AI assistance: Codex was used for code-path analysis, the guard and regression
test, running checks, and PR preparation. The final diff and test results were
shown to the contributor for review before submission.

### Changelog

```release-note
Fix misleading pprof logs when bugtool runs without gops.
```

### DIFF
--- a/pkg/bugtool/bugtool.go
+++ b/pkg/bugtool/bugtool.go
@@ -623,6 +623,10 @@ func (s *bugtoolInfo) getPProf(file string, gopsSignal byte) error {
 }
 
 func (s *bugtoolInfo) addPProfInfo() {
+	if s.info.GopsAddr == "" {
+		return
+	}
+
 	profiles := map[string]byte{
 		"cpu":  gopssignal.CPUProfile,
 		"heap": gopssignal.HeapProfile,
--- a/pkg/bugtool/bugtool.go
+++ b/pkg/bugtool/bugtool.go
@@ -319,8 +319,12 @@ func doBugtool(info *InitInfo, outFname string, commandActions []CommandAction, 
 	si.ExecCmd("dmesg.out", "dmesg")
 	si.addTcInfo()
 	si.addBpftoolInfo()
-	si.addGopsInfo()
-	si.addPProfInfo()
+	if si.info.GopsAddr == "" {
+		si.multiLog.Info("Skipping gops dump info as daemon is running without gops, use --gops-address to enable gops")
+	} else {
+		si.addGopsInfo()
+		si.addPProfInfo()
+	}
 	si.dumpPolicyFilterMap()
 	si.addProcessCache()
 	si.addExecveMap()
@@ -595,11 +599,6 @@ func (s *bugtoolInfo) addBpftoolInfo() {
 }
 
 func (s *bugtoolInfo) getPProf(file string, gopsSignal byte) error {
-	if s.info.GopsAddr == "" {
-		s.multiLog.Info("Skipping gops dump info as daemon is running without gops, use --gops-address to enable gops")
-		return nil
-	}
-
 	s.multiLog.WithField("gops-address", s.info.GopsAddr).Info("Contacting gops server for pprof dump")
 
 	conn, err := net.Dial("tcp", s.info.GopsAddr)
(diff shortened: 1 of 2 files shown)


pkg/bugtool/bugtool.go
bar234: This isn't needed anymore since this is checked in the caller

bar234: What do you think about adding check for `si.info.GopsAddr == ""` here before calling `si.addGopsInfo()` and `si.addPProfInfo()`? This is the only place those functions are called and we can remove any extra checks for `si.info.GopsAddr == ""` within them (such as in `addGopsInfo` and `getPProf`)

You could also print the error message here too if the gops addr wasn't provided

bar234 (changes requested):
Few comments

pkg/bugtool/bugtool.go
Ehtasham-Yasin: Done — removed the redundant check since it’s now handled by the caller. Thanks!

Ehtasham-Yasin: Yes, that makes sense. I moved the GopsAddr == "" check to the caller before addGopsInfo() and addPProfInfo(), log the skip message there, and removed the redundant checks from the callees. Thanks for the suggestion.

will-isovalent (changes requested):
Hey, thanks for the PR. I really don't think we need a regression test for this. That seems like overkill. Could you drop the test? Then I think it should be good to merge.

bar234 (approved):
Happy with the changes for my comments. I agree with @will-isovalent, we don't need the extra test case

Ehtasham-Yasin:
Thanks @will-isovalent and @bar234 — I removed TestGopsDisabledLogsOnce and its unused imports in 437111686. The production code is unchanged.

FedeDP (approved):
LGTM

will-isovalent (approved):

mtardy (changes requested):
thanks for the PR, please re-organize the commits so that you don't introduce code that you remove in the same commits, we rebase and merge: https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/#commit-changes

you can probably keep two separated commits from what I see? 

Also please add a git commit log on "why" on all commits.

Thanks a lot!

Ehtasham-Yasin:
Thanks @mtardy — I reorganized the branch into two signed-off commits with why-focused bodies and force-pushed it. The second commit only removes the pre-existing redundant guard, so no commit introduces code that the next one removes. The final diff is unchanged.

kkourt:
Kind note: once the PR is ready for review, you can notify the reviewers by clicking on the github button "Re-request review".

mtardy (approved):
Thanks, lgtm. 

Just as a note I think we should use Go errors there btw instead of moving logging around in general but that'll fix the current issue :) thanks!

PR #5330, opened 2026-07-23
new: support multiple preloaded (string) resolved arguments

FedeDP:
### Description

Both uprobe and usdt support preloading a single `string` resolve argument:
* https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/genericuprobe.go#L1037
* https://github.com/cilium/tetragon/blob/main/pkg/sensors/tracing/genericusdt.go#L474

With this patch, we aim to support `MAX_POSSIBLE_ARGS` arguments (5).

### Changelog

```release-note
new: support multiple preloaded (string) resolved arguments
```

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -539,13 +539,16 @@ FUNC_INLINE long get_pt_regs_arg(struct pt_regs *ctx, struct event_config *confi
 #endif /* __TARGET_ARCH_x86 && (GENERIC_KPROBE || GENERIC_UPROBE) */
 
 #if defined(GENERIC_UPROBE) || defined(GENERIC_USDT)
-FUNC_INLINE unsigned long get_preload_arg(struct pt_regs *ctx, long ty, arg_status_t *status)
+FUNC_INLINE unsigned long get_preload_arg(struct pt_regs *ctx, long ty, int index, arg_status_t *status)
 {
 	unsigned long arg = 0;
-	__u64 id = get_current_pid_tgid();
+	struct preload_data *data;
+	struct preload_key key = {
+		.tgid = get_current_pid_tgid(),
+		.index = index,
+	};
 
-	struct preload_data *data = map_lookup_elem(&sleepable_preload, &id);
-
+	data = map_lookup_elem(&sleepable_preload, &key);
 	if (data)
 		*status = data->status;
 	else
@@ -564,7 +567,7 @@ FUNC_INLINE unsigned long get_preload_arg(struct pt_regs *ctx, long ty, arg_stat
 	return arg;
 }
 #else
-FUNC_INLINE long get_preload_arg(struct pt_regs *ctx, long ty, arg_status_t *status)
+FUNC_INLINE long get_preload_arg(struct pt_regs *ctx, long ty, int index, arg_status_t *status)
 {
 	return 0;
 }
@@ -612,7 +615,7 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 
 	a = (&e->a0)[index];
 	if (am & ARGM_PRELOAD)
-		a = get_preload_arg(ctx, ty, &e->arg_status[index & MAX_POSSIBLE_ARGS_MASK]);
+		a = get_preload_arg(ctx, ty, index, &e->arg_status[index & MAX_POSSIBLE_ARGS_MASK]);
 	else
 		extract_arg(config, index, &a, false, &e->arg_status[index & MAX_POSSIBLE_ARGS_MASK]);
 #else
@@ -626,7 +629,7 @@ FUNC_INLINE long generic_read_arg(void *ctx, int index, long off, struct bpf_map
 	 *   - real argument value
 	 */
 	if (am & ARGM_PRELOAD) {
-		a = get_preload_arg(ctx, ty, &e->arg_status[index & MAX_POSSIBLE_ARGS_MASK]);
+		a = get_preload_arg(ctx, ty, index, &e->arg_status[index & MAX_POSSIBLE_ARGS_MASK]);
 	} else {
 		asm volatile("%[index] &= %1 ;\n"
 			     : [index] "+r"(index)
--- a/bpf/process/user_preload.h
+++ b/bpf/process/user_preload.h
@@ -9,6 +9,12 @@
 #include "errmetrics.h"
 #include "usdt_arg.h"
 
+struct preload_key {
+	__u64 tgid;
+	__u8 index;
+	__u8 pad[7];
+};
+
 struct preload_data {
 	arg_status_t status;
 	unsigned char data[4096];
@@ -17,7 +23,7 @@ struct preload_data {
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, 1); // will be resized by agent when needed
-	__type(key, __u64);
+	__type(key, struct preload_key);
 	__type(value, struct preload_data);
 	__uint(map_flags, BPF_F_NO_PREALLOC);
 } sleepable_preload SEC(".maps");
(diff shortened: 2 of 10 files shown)


bpf/process/user_preload.h
```diff
@@ -186,12 +197,38 @@ user_preload(struct pt_regs *ctx)
...
 
+FUNC_LOCAL int
+cleanup_preload(int idx, struct preload_key *key)
+{
+	asm volatile("%[idx] &= %1 ;\n"
+		     : [idx] "+r"(idx)
+		     : "i"(MAX_POSSIBLE_ARGS_MASK));
+
+	key->index = idx;
+	map_delete_elem(&sleepable_preload, key);
```
FedeDP, line 211: We will cleanup `0` to `MAX_POSSIBLE_ARGS_MASK` (5) keys even if non-existent.
Improving this loop for only existent keys is a bit hard since we would've to do exactly what we do in `try_preload_arg`, ie:
* take the event config
* check which param has `ARGM_PRELOAD` flag
* take their type
* if type == string -> remove the index

bpf/process/user_preload.h
```diff
@@ -17,28 +25,31 @@ struct preload_data {
 struct {
 	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, 1); // will be resized by agent when needed
-	__type(key, __u64);
+	__type(key, struct preload_key);
```
kkourt, line 26: This doubles the size of a fairly large map, but I think with Jiri's recent change https://github.com/cilium/tetragon/pull/5171/changes/5cf94319d3005d3b1f3deee85380683ff662c944 it shouldn't be an issue since most of these maps are not fully populated.

bpf/process/user_preload.h
```diff
@@ -9,6 +9,14 @@
 #include "errmetrics.h"
 #include "usdt_arg.h"
 
+struct preload_key {
+	__u64 tgid;
+	__u8 index;
+	__u8 pad1;
+	__u16 pad2;
+	__u32 pad3;
```
kkourt: nit: `__u8 pad[7]`

bpf/process/user_preload.h
```diff
@@ -186,12 +197,38 @@ user_preload(struct pt_regs *ctx)
...
+FUNC_LOCAL int
+cleanup_preload(int idx, struct preload_key *key)
+{
+	asm volatile("%[idx] &= %1 ;\n"
+		     : [idx] "+r"(idx)
+		     : "i"(MAX_POSSIBLE_ARGS_MASK));
+
+	key->index = idx;
+	map_delete_elem(&sleepable_preload, key);
+	return 0;
```
kkourt, line 212: I think there is no need to return something here. It's always 0 and we don't check it in the caller.

kkourt (commented):
overall lgtm, but let's see what other folk say :)

bpf/process/user_preload.h
FedeDP: Exacty :)

bpf/process/user_preload.h
FedeDP: Oh, right :D

bpf/process/user_preload.h
FedeDP: Mmh this function is being called from within a `bpf_loop`; i think this is the mandatory signature!

pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -550,13 +550,15 @@ func TestUprobeResolvePageFault(t *testing.T) {
...
-		kpArgs    []*ec.KprobeArgumentChecker
+		specTy     string
+		filterVals []string
+		field      string
+		kpArgs     []*ec.KprobeArgumentChecker
 	}{
-		{"string", "hello world!", "subp.buff", []*ec.KprobeArgumentChecker{
-			ec.NewKprobeArgumentChecker().WithStringArg(sm.Full("hello world!")),
+		{"string", []string{"hello", "world!"}, "subp.buff", []*ec.KprobeArgumentChecker{
+			// twice because it is resolved as first and second param, and they are both equal.
```
andrewstrohman: I don't understand this part:
> and they are both equal

I understand why the code is the way it is, I just don't understand the second part of this comment.

andrewstrohman (approved):

pkg/sensors/tracing/uprobe_reg_test.go
FedeDP: Oh right, it came from an old code revision; let me fix that.
Great catch!

bpf/process/user_preload.h
kkourt: How about we add the same semantics in the unrolled, and bpf_for versions? (i.e., check the return argument and break if it's not zero).

FedeDP: We don't have to break in case of `map_delete_elem` error, since:
* the loop goes between 0 and MAX_POSSIBLE_ARGS
* we may have non-present keys in the `sleepable_preload` map (not all arguments must be resolved && are strings
* therefore we should not check `map_delete_elem` return code otherwise we would stop iterating at first non-present key 

Note that https://github.com/cilium/tetragon/pull/5330/changes/6aab277b57321d00e95296494ae304dfa0fd97e1 commit message tries to highlight this; let me know if i should expand.

FedeDP: After a private discussion, i now get what you meant! I am going to push the requested changes asap.
Basically, we want to have the caller code ready in case in the future we start to check for `map_delete_elem` error code.
I'll do the same for `user_preload()` too since it does no error checking today.

kkourt (approved):
LGTM.

Would be nice to get an ACK from @olsajiri, since he introduced the `TestUprobeValidationMultiplePreloadArguments` test: https://github.com/cilium/tetragon/commit/3d6a0e32713d3fd2ed76ffaed3157e82ca96e4ef

FedeDP:
Rebased on top of main to fix conflict.

pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -550,13 +550,14 @@ func TestUprobeResolvePageFault(t *testing.T) {
...
+		specTy     string
+		filterVals []string
+		field      string
+		kpArgs     []*ec.KprobeArgumentChecker
 	}{
-		{"string", "hello world!", "subp.buff", []*ec.KprobeArgumentChecker{
-			ec.NewKprobeArgumentChecker().WithStringArg(sm.Full("hello world!")),
+		{"string", []string{"hello", "world!"}, "subp.buff", []*ec.KprobeArgumentChecker{
+			ec.NewKprobeArgumentChecker().WithStringArg(sm.Full("hello")),
+			ec.NewKprobeArgumentChecker().WithStringArg(sm.Full("world!")),
```
olsajiri: please add specific test for having multiple preload arguments

bpf/process/user_preload.h
```diff
@@ -171,12 +180,17 @@ user_preload(struct pt_regs *ctx)
 	};
 	if (CONFIG(ITER_NUM)) {
 		bpf_for(i, 0, MAX_POSSIBLE_ARGS)
-			try_preload_arg(i, &preload_data);
+		{
+			if (try_preload_arg(i, &preload_data))
```
olsajiri, line 184: try_preload_arg returns always 0, we need to change it as well

bpf/process/generic_calls.h
```diff
@@ -539,13 +539,15 @@ FUNC_INLINE long get_pt_regs_arg(struct pt_regs *ctx, struct event_config *confi
...
 	unsigned long arg = 0;
-	__u64 id = get_current_pid_tgid();
-
-	struct preload_data *data = map_lookup_elem(&sleepable_preload, &id);
+	struct preload_key key = {
+		.tgid = get_current_pid_tgid(),
+		.index = index,
+	};
 
+	struct preload_data *data = map_lookup_elem(&sleepable_preload, &key);
```
olsajiri: nit, let's keep declaration and code separated

olsajiri (commented):
looks good, left some comments

I wonder we should should be more proactive in here wrt `option.Config.SleepablePreloadSize` and when we see multiple preload arguments then either increase the preload map size or at least warn user (I prefer this one)

pkg/sensors/tracing/uprobe_reg_test.go
FedeDP: Do you mean keeping the old test by itself and adding the new one on top?

FedeDP:
> I wonder we should should be more proactive in here wrt option.Config.SleepablePreloadSize and when we see multiple preload arguments then either increase the preload map size or at least warn user (I prefer this one)

I like it, let me add the warning!

pkg/sensors/tracing/uprobe_reg_test.go
FedeDP: Done!

FedeDP:
Re-pushed first commit to move the new test to a policytest :)

PR #5338, opened 2026-07-24
new: make regs_map selector idx aware

FedeDP:
### Description

Prior to this change, if multiple selectors had an `Override` match action for uprobe sensor, we would fail with:
> error: only single instance of regs action is allowed.

This is because we were not being selector id aware, therefore every selector for the same hook shared the same `regs` instance (and thus `regs_maps` entry).
This PR makes `KernelSelectorState.regs` , `MatchActions` and `populateUprobeRegs` selector idx aware.
Also, `TestUprobeOverrideMultipleSelectors` was created to show the bug and avoid future regressions.

### Changelog

```release-note
fix: make regs_map selector idx aware
```

### DIFF
--- a/bpf/process/uprobe_offload.h
+++ b/bpf/process/uprobe_offload.h
@@ -33,7 +33,7 @@ struct uprobe_regs {
 };
 
 struct {
-	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(type, BPF_MAP_TYPE_HASH);
 	__uint(max_entries, 1);
 	__type(key, __u32);
 	__type(value, struct uprobe_regs);
--- a/contrib/tester-progs/uprobe-test.c
+++ b/contrib/tester-progs/uprobe-test.c
@@ -23,8 +23,12 @@ int uprobe_test_lib_string_arg2(int one, int two, char *str, int four, int five)
 int uprobe_test_lib_string_arg3(int one, int two, int three, char *str, int five);
 int uprobe_test_lib_string_arg4(int one, int two, int three, int four, char *str);
 
-int main(void)
+int main(int argc, char *argv[])
 {
+	if (argc > 1) {
+		return uprobe_test_lib_string_arg1(atoi(argv[1]), "two", 3, 4, 5);
+	}
+
 	char *str_arg = "hello world!";
 
 	uprobe_test_lib();
@@ -46,4 +50,5 @@ int main(void)
 	uprobe_test_lib_string_arg2(1, 2, "three", 4, 5);
 	uprobe_test_lib_string_arg3(1, 2, 3, "four", 5);
 	uprobe_test_lib_string_arg4(1, 2, 3, 4, "five");
+	return 0;
 }
--- a/pkg/selectors/kernel_regs_amd64.go
+++ b/pkg/selectors/kernel_regs_amd64.go
@@ -13,9 +13,9 @@ import (
 	"github.com/cilium/tetragon/pkg/asm"
 )
 
-func parseOverrideRegs(k *KernelSelectorState, values []string, errValue uint64, newOffset int64) (uint32, error) {
-	if len(k.regs) > 0 {
-		return uint32(0xffffffff), errors.New("only single instance of regs action is allowed")
+func parseOverrideRegs(k *KernelSelectorState, selIdx int, values []string, errValue uint64, newOffset int64) error {
+	if _, exists := k.regs[selIdx]; exists {
+		return errors.New("only single instance of regs action is allowed")
 	}
 
 	regs := []processapi.RegAssignment{}
@@ -37,7 +37,7 @@ func parseOverrideRegs(k *KernelSelectorState, values []string, errValue uint64,
 	for _, val := range values {
 		ass, err := asm.ParseAssignment(val)
 		if err != nil {
-			return uint32(0xffffffff), err
+			return err
 		}
 
 		regs = append(regs, processapi.RegAssignment{
@@ -50,6 +50,6 @@ func parseOverrideRegs(k *KernelSelectorState, values []string, errValue uint64,
 		})
 	}
 
-	k.regs = regs
-	return uint32(k.uprobeID), nil
+	k.regs[selIdx] = regs
+	return nil
 }
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -115,7 +115,7 @@ type genericUprobe struct {
 	pendingEvents *lru.Cache[pendingEventKey, pendingEvent[*tracing.MsgGenericUprobeUnix]]
 }
 
-func populateUprobeRegs(m *ebpf.Map, id int, regs []processapi.RegAssignment) error {
+func populateUprobeRegs(m *ebpf.Map, id uint32, regs []processapi.RegAssignment) error {
 	uprobeRegs := processapi.UprobeRegs{}
 
 	n := copy(uprobeRegs.Ass[:], regs)
@@ -123,7 +123,17 @@ func populateUprobeRegs(m *ebpf.Map, id int, regs []processapi.RegAssignment) er
 		logger.GetLogger().Warn("register assignments count mismatch", "#regs", len(regs))
 	}
 	uprobeRegs.Cnt = uint32(n)
-	return m.Update(uint32(id), uprobeRegs, ebpf.UpdateAny)
+	return m.Update(id, uprobeRegs, ebpf.UpdateAny)
+}
+
+func populateSelectorRegsMap(m *ebpf.Map, selector *selectors.KernelSelectorState) error {
+	for selIdx, regs := range selector.Regs() {
+		err := populateUprobeRegs(m, selector.UprobeRegsMapID(selIdx), regs)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
 }
 
 func (g *genericUprobe) SetID(id idtable.EntryID) {
@@ -272,7 +282,7 @@ func loadSingleUprobeSensor(uprobeEntry *genericUprobe, args sensors.LoadProbeAr
 				&program.MapLoad{
 					Name: "regs_map",
 					Load: func(m *ebpf.Map, _ string) error {
-						return populateUprobeRegs(m, 0, selector.Regs())
+						return populateSelectorRegsMap(m, selector)
 					},
 				},
 			)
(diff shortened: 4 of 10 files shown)


pkg/sensors/tracing/uprobe_reg_test.go
```diff
@@ -107,6 +107,123 @@ spec:
 	require.NoError(t, err)
 }
 
+func TestUprobeOverrideMultipleSelectors(t *testing.T) {
```
andrewstrohman: We are trying to make all new tests policytests. Can this test be converted to use the policytest framework?

pkg/selectors/kernel_regs_amd64.go
```diff
@@ -13,9 +13,9 @@ import (
 	"github.com/cilium/tetragon/pkg/asm"
 )
 
-func parseOverrideRegs(k *KernelSelectorState, values []string, errValue uint64, newOffset int64) (uint32, error) {
-	if len(k.regs) > 0 {
-		return uint32(0xffffffff), errors.New("only single instance of regs action is allowed")
+func parseOverrideRegs(k *KernelSelectorState, selIdx int, values []string, errValue uint64, newOffset int64) error {
+	if len(k.regs[selIdx]) > 0 {
```
andrewstrohman: IMHO, it would be more clear just to test if the key exists.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -1426,7 +1445,7 @@ func createMultiUprobeSensor(polInfo *policyInfo, sensorPath string, multiIDs []
 
 	if has.sleepableOffload {
 		regsMap := program.MapBuilderProgram("regs_map", load)
-		regsMap.SetMaxEntries(len(multiIDs))
+		regsMap.SetMaxEntries(regsMapEntries)
```
andrewstrohman: I think we need to set the max number of entries of the regsMap in `createUprobeSensorFromEntry` now, too (for the case where that single entry has multiple selectors).

andrewstrohman (commented):
Other than not setting the max number of entries in the regsMap for createUprobeSensorFromEntry, LGTM.

```
diff --git a/pkg/sensors/tracing/genericuprobe.go b/pkg/sensors/tracing/genericuprobe.go
index b03bf1a95..0ec35a5e4 100644
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -1555,7 +1555,15 @@ func createUprobeSensorFromEntry(polInfo *policyInfo, uprobeEntry *genericUprobe
        }
 
        if has.sleepableOffload {
+               regsMapEntries := len(uprobeEntry.loadArgs.selectors.entry.Regs())
+               if regsMapEntries < 1 {
+                       // keep the map valid even when this particular entry has no
+                       // register assignments of its own (e.g. only a sibling uprobe
+                       // in the same policy needs the override action)
+                       regsMapEntries = 1
+               }
                regsMap := program.MapBuilderProgram("regs_map", load)
+               regsMap.SetMaxEntries(regsMapEntries)
                sleepableOffloadMap := program.MapBuilderProgram("sleepable_offload", load)
                sleepableOffloadMap.SetMaxEntries(sleepableOffloadMaxEntries)
                maps = append(maps, regsMap, sleepableOffloadMap)
```

would resolve this concern for me.

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -1377,19 +1389,26 @@ func createMultiUprobeSensor(polInfo *policyInfo, sensorPath string, multiIDs []
...
 			multiRetIDs = append(multiRetIDs, id)
+			selector = gu.loadArgs.selectors.retrn
 		}
 
 		if has.substring && substringMapEntries == 0 {
 			substringMapEntries = len(gu.loadArgs.selectors.entry.SubStrings())
 		}
+
+		if selector != nil && selector.Regs() != nil {
+			regsMapEntries += len(selector.Regs())
```
andrewstrohman, line 1408: We could save a little memory with this:

```
andy@builder:~/src/tetragon$ git --no-pager diff
diff --git a/pkg/sensors/tracing/genericuprobe.go b/pkg/sensors/tracing/genericuprobe.go
index b03bf1a95..becc68c09 100644
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -1390,6 +1390,11 @@ func createMultiUprobeSensor(polInfo *policyInfo, sensorPath string, multiIDs []
        var maps []*program.Map
        var substringMapEntries int
        var regsMapEntries int
+       // entries in loadArgs.selectors are shared (by pointer) across all
+       // genericUprobe entries derived from the same UProbeSpec (e.g. multiple
+       // Symbols/Offsets/Addrs), so track the ones already counted to avoid
+       // summing the same register assignments multiple times.
+       countedSelectors := make(map[*selectors.KernelSelectorState]struct{})
 
        for _, id := range multiIDs {
                gu, err := genericUprobeTableGet(id)
@@ -1407,7 +1412,10 @@ func createMultiUprobeSensor(polInfo *policyInfo, sensorPath string, multiIDs []
                }
 
                if selector != nil && selector.Regs() != nil {
-                       regsMapEntries += len(selector.Regs())
+                       if _, ok := countedSelectors[selector]; !ok {
+                               countedSelectors[selector] = struct{}{}
+                               regsMapEntries += len(selector.Regs())
+                       }
                }
        }
```

Like how we discussed in the previous PR, the small amount of memory savings may not be worth this additional complexity.

pkg/sensors/tracing/uprobe_reg_test.go
FedeDP: Right, let me try. Thanks!

pkg/sensors/tracing/genericuprobe.go
FedeDP: Thanks, nice catch!

FedeDP: I am wondering whether we could run eg: all policytests both normally and with the `disable-multi` flag to catch these kind of issues. I am not sure whether that would work out of the box though.
cc @kkourt

pkg/sensors/tracing/genericuprobe.go
FedeDP: Yeah probably not worth it; the code looks already complex enough :) Thanks anyway for the suggestion, we might apply it in the future if needed!

pkg/testutils/policytest/trigger.go
```diff
@@ -32,60 +32,74 @@ func (c *CmdTrigger) Trigger(ctx context.Context) error {
 
 func (c *CmdTrigger) ExpectExitCode(val int) *ExecTester {
 	return &ExecTester{
-		CmdTrigger: CmdTrigger{
-			Bin:  c.Bin,
-			Args: c.Args,
-		},
-		ExpectedExitCode: &val,
+		MultiCmdTrigger:  MultiCmdTrigger{triggers: []CmdTrigger{*c}},
```
FedeDP, line 35: cc @kkourt for the policytest trigger-related changes.

andrewstrohman (approved):

pkg/sensors/tracing/genericuprobe.go
andrewstrohman: > I am wondering whether we could run eg: all policytests both normally and with the disable-multi flag to catch these kind of issues.

This seems like a good idea to me. But if we cannot for some reason, then maybe we could leverage `WithParameter`, in order to do it at the policy level instead. Parametrizing won't currently work for this use case, because what we want to substitute is multi-line or too complex to pass with `--set-param`. So if we want to pursue this approach, we would need to create an alias/name for what is actually substituted.

bpf/process/uprobe_offload.h
```diff
@@ -33,7 +33,7 @@ struct uprobe_regs {
 };
 
 struct {
-	__uint(type, BPF_MAP_TYPE_ARRAY);
+	__uint(type, BPF_MAP_TYPE_HASH);
```
olsajiri, line 36: lets put this change to previous commit?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -272,7 +272,13 @@ func loadSingleUprobeSensor(uprobeEntry *genericUprobe, args sensors.LoadProbeAr
 				&program.MapLoad{
 					Name: "regs_map",
 					Load: func(m *ebpf.Map, _ string) error {
-						return populateUprobeRegs(m, 0, selector.Regs())
+						for selIdx, regs := range selector.Regs() {
```
olsajiri: looks like we could move regs_map populating to a function and use it for both single/multi load

pkg/selectors/kernel_regs_amd64.go
```diff
@@ -50,6 +50,6 @@ func parseOverrideRegs(k *KernelSelectorState, values []string, errValue uint64,
 		})
 	}
 
-	k.regs = regs
-	return uint32(k.uprobeID), nil
+	k.regs[selIdx] = regs
```
olsajiri, line 53: should we check if the value already exists first for sanity?

contrib/tester-progs/uprobe-test.c
```diff
@@ -23,8 +23,12 @@ int uprobe_test_lib_string_arg2(int one, int two, char *str, int four, int five)
 int uprobe_test_lib_string_arg3(int one, int two, int three, char *str, int five);
 int uprobe_test_lib_string_arg4(int one, int two, int three, int four, char *str);
 
-int main(void)
+int main(int argc, char *argv[])
 {
+    if (argc > 1) {
```
olsajiri: in terminal I see mix of indentation in here, should be tabs

contrib/tester-progs/uprobe-test.c
```diff
@@ -46,4 +50,5 @@ int main(void)
 	uprobe_test_lib_string_arg2(1, 2, "three", 4, 5);
 	uprobe_test_lib_string_arg3(1, 2, 3, "four", 5);
 	uprobe_test_lib_string_arg4(1, 2, 3, 4, "five");
+    return 0;
```
olsajiri: ditto in here

olsajiri (approved):
left few comments, but looks good, thanks

contrib/tester-progs/uprobe-test.c
FedeDP: Weird, i ran `make format` :/

pkg/selectors/kernel_regs_amd64.go
FedeDP: We already check: https://github.com/cilium/tetragon/pull/5338/changes#diff-5d03e98af63a6ecd5abca290ada3772429454abfefae1e8f3ab0504ed6227b78R17

pkg/sensors/tracing/genericuprobe.go
kkourt: > I am wondering whether we could run eg: all policytests both normally and with the disable-multi flag to catch these kind of issues. I am not sure whether that would work out of the box though.

 I believe we can, but I'm not sure if we want to do on every PR, just because the configuration space is large. I think we should think about  some other approach (e.g., weekly or having specific tests for specific configurations, maybe even running the suite over a random version every once in a while).

PR #5346, opened 2026-07-25
make execve sensor use raw tracepoint

olsajiri:
raw tracepoints are faster than standard perf based tracepoints,
switching execve sensor to use it

### DIFF
--- a/bpf/process/bpf_execve_event.c
+++ b/bpf/process/bpf_execve_event.c
@@ -17,15 +17,9 @@
 
 char _license[] __attribute__((section("license"), used)) = "Dual BSD/GPL";
 
-#ifdef __RHEL7_BPF_PROG
-#define exec_ctx_struct ftrace_raw_sched_process_exec
-#else
-#define exec_ctx_struct trace_event_raw_sched_process_exec
-#endif
-
 #ifndef OVERRIDE_TAILCALL
 int execve_rate(void *ctx);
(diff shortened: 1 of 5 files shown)


bpf/process/bpf_execve_event.c
```diff
@@ -249,14 +243,15 @@ read_execve_shared_info(void *ctx, struct msg_process *p, __u64 pid)
 	execve_joined_info_map_clear(pid);
 }
 
-__attribute__((section("tracepoint/sys_execve"), used)) int
-event_execve(struct exec_ctx_struct *ctx)
+__attribute__((section("raw_tracepoint/sys_execve"), used)) int
+event_execve(struct bpf_raw_tracepoint_args *ctx)
 {
 	struct task_struct *task = (struct task_struct *)get_current_task();
```
FedeDP, line 249: I am wondering whether using `ctx->args[0]` is quicker instead since we skip the call.

FedeDP (commented):
LGTM, left a possible suggestion.

bpf/process/bpf_execve_event.c
olsajiri: `bpf_get_current_task` is inlined by verifier, so I think there's no gain, we also call it on other places

FedeDP (approved):

kkourt (approved):

PR #5353, opened 2026-07-27
Do not use tail calls in execve sensor

olsajiri:
The execve sensor fits into single program (on large kernel forward) which enable
us to use ring-buffer reserve api and get rid of per-cpu buffers (coming next).

Keeping tail-calls in execve program for 4.19 kernels.


pkg/testutils/sensors/load.go
```diff
@@ -137,55 +137,93 @@ func mergeSensorMaps(_ *testing.T, maps1, maps2 []SensorMap, progs1, progs2 []Se
...
 	} else {
 		// all programs except for execve_map_update, execve_rate
 		baseMaps = append(baseMaps, SensorMap{Name: "execve_map", Progs: []uint{0, 1, 2, 3, 4}})
 	}
 
 	if cfg.EnableV511Progs() {
-		baseMaps = append(baseMaps, SensorMap{Name: "tg_rb_events", Progs: []uint{0, 1, 2, 3, 5}})
-		baseMaps = append(baseMaps, SensorMap{Name: "tg_conf_map", Progs: []uint{0, 1, 2, 3, 5}})
+		progs := []uint{0, 1, 2, 3, 5}
+		if cfg.EnableLargeProgs() {
```
kkourt: I'm seeing that we are under a `cfg.EnableV511Progs()` check. Doesn't this imply `cfg.EnableLargeProgs()` already?

kkourt (approved):
LGTM, thanks.

pkg/testutils/sensors/load.go
olsajiri: true, will remove, thanks

FedeDP (approved):

PR #5354, opened 2026-07-29
disable process cache tests

andrewstrohman:
[This related PR](https://github.com/cilium/tetragon/pull/5248) introduces a new flag which allows disabling the process cache. This mode of tetragon is limited in its ability to enrich events with process information. As such, I don't expect policytests to pass when the checker utilizes `WithProcess` or `WithParent` while running in this mode.

This PR reverts the effect of `WithProcess()` and `WithParent()` on a checker when a policytest is run against an agent with its process cache disabled.

I added/converted tracepoints tests in order to get policytest coverage for the tracepoint hook type. This way, I am able to test that events for each hook type can still be delivered with the process cache disabled.

### DIFF
--- a/api/v1/tetragon/codegen/eventchecker/eventchecker.pb.go
+++ b/api/v1/tetragon/codegen/eventchecker/eventchecker.pb.go
@@ -434,9 +434,21 @@ func (checker *ProcessExecChecker) WithProcess(check *ProcessChecker) *ProcessEx
 	return checker
 }
 
+// UnsetProcess unsets the Process check to the ProcessExecChecker
+func (checker *ProcessExecChecker) UnsetProcess() *ProcessExecChecker {
+	checker.Process = nil
+	return checker
+}
+
 // WithParent adds a Parent check to the ProcessExecChecker
 func (checker *ProcessExecChecker) WithParent(check *ProcessChecker) *ProcessExecChecker {
 	checker.Parent = check
+	return checker
+}
+
+// UnsetParent unsets the Parent check to the ProcessExecChecker
+func (checker *ProcessExecChecker) UnsetParent() *ProcessExecChecker {
+	checker.Parent = nil
 	return checker
 }
 
@@ -667,9 +679,21 @@ func (checker *ProcessExitChecker) WithProcess(check *ProcessChecker) *ProcessEx
 	return checker
 }
 
+// UnsetProcess unsets the Process check to the ProcessExitChecker
+func (checker *ProcessExitChecker) UnsetProcess() *ProcessExitChecker {
+	checker.Process = nil
+	return checker
+}
+
 // WithParent adds a Parent check to the ProcessExitChecker
 func (checker *ProcessExitChecker) WithParent(check *ProcessChecker) *ProcessExitChecker {
 	checker.Parent = check
+	return checker
+}
+
+// UnsetParent unsets the Parent check to the ProcessExitChecker
+func (checker *ProcessExitChecker) UnsetParent() *ProcessExitChecker {
+	checker.Parent = nil
 	return checker
 }
 
@@ -873,9 +897,21 @@ func (checker *ProcessKprobeChecker) WithProcess(check *ProcessChecker) *Process
 	return checker
 }
 
+// UnsetProcess unsets the Process check to the ProcessKprobeChecker
+func (checker *ProcessKprobeChecker) UnsetProcess() *ProcessKprobeChecker {
+	checker.Process = nil
+	return checker
+}
+
 // WithParent adds a Parent check to the ProcessKprobeChecker
 func (checker *ProcessKprobeChecker) WithParent(check *ProcessChecker) *ProcessKprobeChecker {
 	checker.Parent = check
+	return checker
+}
+
+// UnsetParent unsets the Parent check to the ProcessKprobeChecker
+func (checker *ProcessKprobeChecker) UnsetParent() *ProcessKprobeChecker {
+	checker.Parent = nil
 	return checker
 }
 
@@ -1469,9 +1505,21 @@ func (checker *ProcessTracepointChecker) WithProcess(check *ProcessChecker) *Pro
 	return checker
 }
 
+// UnsetProcess unsets the Process check to the ProcessTracepointChecker
+func (checker *ProcessTracepointChecker) UnsetProcess() *ProcessTracepointChecker {
+	checker.Process = nil
+	return checker
+}
+
 // WithParent adds a Parent check to the ProcessTracepointChecker
 func (checker *ProcessTracepointChecker) WithParent(check *ProcessChecker) *ProcessTracepointChecker {
 	checker.Parent = check
+	return checker
+}
+
+// UnsetParent unsets the Parent check to the ProcessTracepointChecker
+func (checker *ProcessTracepointChecker) UnsetParent() *ProcessTracepointChecker {
+	checker.Parent = nil
 	return checker
 }
 
@@ -1716,9 +1764,21 @@ func (checker *ProcessUprobeChecker) WithProcess(check *ProcessChecker) *Process
 	return checker
 }
 
+// UnsetProcess unsets the Process check to the ProcessUprobeChecker
+func (checker *ProcessUprobeChecker) UnsetProcess() *ProcessUprobeChecker {
+	checker.Process = nil
+	return checker
+}
+
 // WithParent adds a Parent check to the ProcessUprobeChecker
 func (checker *ProcessUprobeChecker) WithParent(check *ProcessChecker) *ProcessUprobeChecker {
 	checker.Parent = check
+	return checker
+}
+
+// UnsetParent unsets the Parent check to the ProcessUprobeChecker
+func (checker *ProcessUprobeChecker) UnsetParent() *ProcessUprobeChecker {
+	checker.Parent = nil
 	return checker
 }
 
@@ -1996,9 +2056,21 @@ func (checker *ProcessUsdtChecker) WithProcess(check *ProcessChecker) *ProcessUs
 	return checker
 }
 
+// UnsetProcess unsets the Process check to the ProcessUsdtChecker
+func (checker *ProcessUsdtChecker) UnsetProcess() *ProcessUsdtChecker {
+	checker.Process = nil
+	return checker
+}
+
 // WithParent adds a Parent check to the ProcessUsdtChecker
 func (checker *ProcessUsdtChecker) WithParent(check *ProcessChecker) *ProcessUsdtChecker {
 	checker.Parent = check
+	return checker
+}
+
+// UnsetParent unsets the Parent check to the ProcessUsdtChecker
+func (checker *ProcessUsdtChecker) UnsetParent() *ProcessUsdtChecker {
+	checker.Parent = nil
 	return checker
 }
 
@@ -2239,9 +2311,21 @@ func (checker *ProcessLsmChecker) WithProcess(check *ProcessChecker) *ProcessLsm
 	return checker
 }
 
+// UnsetProcess unsets the Process check to the ProcessLsmChecker
+func (checker *ProcessLsmChecker) UnsetProcess() *ProcessLsmChecker {
+	checker.Process = nil
+	return checker
+}
+
 // WithParent adds a Parent check to the ProcessLsmChecker
 func (checker *ProcessLsmChecker) WithParent(check *ProcessChecker) *ProcessLsmChecker {
 	checker.Parent = check
+	return checker
+}
+
+// UnsetParent unsets the Parent check to the ProcessLsmChecker
+func (checker *ProcessLsmChecker) UnsetParent() *ProcessLsmChecker {
+	checker.Parent = nil
 	return checker
 }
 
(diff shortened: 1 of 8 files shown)


cmd/tetra/policytest/policytest.go
```diff
@@ -180,6 +182,34 @@ func runCmd() *cobra.Command {
...
+					if !baseFilterFn(t) {
+						return false
+					}
+					for _, l := range t.Labels {
+						if _, ok := labelSet[l]; ok {
+							return true
+						}
+					}
+					return false
+				}
```
FedeDP: Can't we move this inside the case label above?

andrewstrohman: The idea here is that the switch statement is used to do a first pass filter to find candidate tests, and then the labels flag can filter on a second pass of what remains.

If I move this inside `case len(labels) > 0:`, it would mean that if the user specifies the allTests flag, or they specify a list of tests, they would not be able to filter further by label. This might be a fine design choice. But I don't see a way to keep the same behavior when moving this code into the switch statement.

FedeDP: Oh right didn't notice `baseFilterFn` , sorry! It makes sense.

FedeDP:
> This allows me to start tetragon with the `--disable-process-cache` flag and then run a subset of tests against it. I accomplished this with: `tetra policytest run --label disable-process-cache`

I am wondering whether it would be better to design a way for `tetra policytest` to directly inherit running `tetragon` instance flags, and filter on those.
For example:
* add a GRPC call to fetch the flags set in the running agent: `GetRunningFlags()` -> `map[string]string]{"bpf-lib": "bpf/objs", "disable-process-cache": ""}`
* `tetra --filter-agent-flags` -> calls the GRPC method and use the flags to filter (or just do some other stuff, for example use 

The idea is that we might not really bother to control the agent, but just want to run a set of tests against the running one.
Just throwing an idea :)
cc @kkourt wdyt?

kkourt:
I think it's an interesting idea :)

Here are my thoughts:

> > This allows me to start tetragon with the `--disable-process-cache` flag and then run a subset of tests against it. I accomplished this with: `tetra policytest run --label disable-process-cache`
> 
> I am wondering whether it would be better to design a way for `tetra policytest` to directly inherit running `tetragon` instance flags, and filter on those. For example:
> 
>   add a GRPC call to fetch the flags set in the running agent: `GetRunningFlags()` -> `map[string]string]{"bpf-lib": "bpf/objs", "disable-process-cache": ""}`

The agent flags (configuration) are already available via gRPC.
Indeed, they can be used to skip tests via `SkipInfo.AgentInfo.Conf`.

<details>

https://github.com/cilium/tetragon/blob/a5b96a65ffff546b2a068b5039d0c0b76afe96a0/pkg/testutils/policytest/policytest.go#L39-L42

https://github.com/cilium/tetragon/blob/a5b96a65ffff546b2a068b5039d0c0b76afe96a0/pkg/tetragoninfo/info.go#L20-L26

</details>

However, skipping does not quite achieve what Andy wants, unless we  add logic to  all other tests that they should be skipped if the process cache is disabled which is a bad idea (™️).

> `tetra --filter-agent-flags` -> calls the GRPC method and use the flags to filter (or just do some other stuff, for example use

How would that work? Specifically, what would be the metadata in the tests to enable this. I can't think of a better approach than labels to do that. On top of that, it would be possible to have some logic that selects tests using their labels and `tetragoninfo.Info`, but this seems like overengineering to me at this point.

FedeDP:
> I can't think of a better approach than labels to do that.

I agree, my idea was that we automate the `labels` filter based on the config flags the agent is running with; similar to what Andy did, but without the need for the user to actually pass the labels to `tetra policytest` because they get inherited by the running agent instance.

Let's say we have 2 tests:
```go
var _ = policytest.NewBuilder("kprobe-lseek").WithLabels("disable-process-cache")....
var _ = policytest.NewBuilder("nomulti-kprobe-lseek").WithLabels("disable-krpobe-multi")....
```
The former would run only if the agent is running with the `--disable-process-cache` flag, while the latter would only run if the agent is running with `--disable-kprobe-multi` flag.

andrewstrohman:
> I agree, my idea was that we automate the `labels` filter based on the config flags the agent is running with; similar to what Andy did, but without the need for the user to actually pass the labels to `tetra policytest` because they get inherited by the running agent instance.
> 
> Let's say we have 2 tests:
> 
> ```go
> var _ = policytest.NewBuilder("kprobe-lseek").WithLabels("disable-process-cache")....
> var _ = policytest.NewBuilder("nomulti-kprobe-lseek").WithLabels("disable-krpobe-multi")....
> ```
> 
> The former would run only if the agent is running with the `--disable-process-cache` flag, while the latter would only run if the agent is running with `--disable-kprobe-multi` flag.

I like this idea, but I would also like to keep the ability to explicitly filter by label as well. All the pre-existing labels represent the hook type being tested. If a developer has made a change specific to one hook type, they may only want to run tests related to that hook in early development. Since the hook type is not a command line arg, they wouldn't be able to do this without keeping the ability to explicitly filter by label. I think we may want to group tests by things other than just tetragon flags.

With regard to the quoted example, what happens when the user runs tetragon with both --disable-process-cache and --disable-kprobe-multi? Does it mean that both tests run or neither? Also, what about the command line flags that don't correspond to a label, ie 'bpf-lib'? We obviously don't want to label every test with label `bpf-lib`. So, it seems we would need a way to specify which command line args should cause filtering by label. Maybe you're thinking that there would be a value for flag `--filter-agent-flags` which would somehow help with this.

FedeDP:
> With regard to the quoted example, what happens when the user runs tetragon with both --disable-process-cache and --disable-kprobe-multi?

I'd say the union of the tests would run, so both of them.

> So, it seems we would need a way to specify which command line args should cause filtering by label.

Perhaps `tetra` could collect all labels used by the declared tests and only filter on those (and skip the unused ones).
Anyway, this was just an idea that popped my mind while reading this PR :) if both of you think it's useful i could work on that.

> but I would also like to keep the ability to explicitly filter by label as well

I agree, let's go on with this one!

EDIT: forgot to mention
> Since the hook type is not a command line arg, they wouldn't be able to do this without keeping the ability to explicitly filter by label. I think we may want to group tests by things other than just tetragon flags.

You are 100% right

FedeDP (approved):

kkourt (commented):
Thanks!

This makes sense to me, but one problem I see is that we will end up duplicating the tests for cases where the process cache is enabled or not.

If we can't avoid it, that's fine, but I would like to at least explore alternatives.

One approach would be to wrap the process checker. I haven't tried this, but my expectation is that the only things that are affected are the `WithProcess` and `WithParent` checks. So we can have something like:


```go

//go:build !windows

package policytest

import (
	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
)

// policytest wrapper for ProcessKprobeChecker
type ProcessKprobeChecker struct {
	ec.ProcessKprobeChecker
	disabledPC bool
}

func NewProcessKprobeChecker(name string, cnf *Conf) *ProcessKprobeChecker {
	return &ProcessKprobeChecker{
		*ec.NewProcessKprobeChecker(name),
		false, /* will be determined based on cnf, by adding agentInfo */
	}
}

func (checker *ProcessKprobeChecker) WithProcess(check *ec.ProcessChecker) *ProcessKprobeChecker {
	if !checker.disabledPC {
		checker.ProcessKprobeChecker.WithProcess(check)
	}
	return checker
}

func (checker *ProcessKprobeChecker) WithParant(check *ec.ProcessChecker) *ProcessKprobeChecker {
	if !checker.disabledPC {
		checker.ProcessKprobeChecker.WithParent(check)
	}
	return checker
}
```

In `pkg/testutils/policytest/checkers.go` (or similar).

Then we can do something like:

```diff
-var _ = policytest.NewBuilder("kprobe-null-string").WithLabels("kprobes").WithPolicyTemplate(`
+var _ = policytest.NewBuilder("kprobe-null-string").WithLabels("kprobes", "disabled-process-cache").WithPolicyTemplate(`
 apiVersion: cilium.io/v1alpha1
 kind: TracingPolicy
 metadata:
@@ -86,7 +86,7 @@ spec:
       type: "int"
 `).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
        myBin := c.TestBinary("null-mount")
-       argChecker := ec.NewProcessKprobeChecker("kprobe-null-string").
+       argChecker := policytest.NewProcessKprobeChecker("kprobe-null-string", c).
                WithProcess(ec.NewProcessChecker().
                        WithBinary(sm.Full(myBin))).WithArgs(ec.NewKprobeArgumentListMatcher().
                WithOperator(lc.Ordered).
```

One missing piece is that the `AgentInfo` is not, currently, available in `Conf`, but I think we can probably thread it there.

WDYT?

andrewstrohman:
> One approach would be to wrap the process checker. I haven't tried this, but my expectation is that the only things that are affected are the `WithProcess` and `WithParent` checks.

Yes, I think WithProcess and WithParent would be the only checkers affected. I tried this out and it worked when the WithProcess() was the first call, but it doesn't work when it's chained later. 

[Here](https://github.com/cilium/tetragon/blob/66bd081f3e1acc8dd29b82e31f44d9d7480b8c45/tests/policytests/lsm.go#L83-L85) is an example of where this doesn't work due to ordering of the chained calls. The problem is that the first chained call, `With FunctionName` returns the embedded type (`ec.ProcessKprobeChecker`) instead of the wrapped type (`ProcessKprobeChecker`). Because of this, the subsequent `WithProcess` call will be on type   `ec.ProcessKprobeChecker`, instead of `ProcessKprobeChecker`. We could fix this by overriding all function calls (ie `WithArgs`, etc..) but when a new `WithX()` function is introduced, we would have to ensure that we override it, which makes that solution feel fragile.

I've experimented with a different approach:

```

//go:build !windows

package policytest

import (
	ec "github.com/cilium/tetragon/api/v1/tetragon/codegen/eventchecker"
)

func processCacheDisabled(cnf *Conf) bool {
	if cnf != nil && cnf.AgentInfo != nil {
		disabledPC, _ := cnf.AgentInfo.Conf["disable-process-cache"].(bool)
		return disabledPC
	}
	return false
}

// GateProcess returns check unless the process cache is disabled, in which case it
// returns nil. Pass its result to WithProcess/WithParent so the check has no effect
// when the process cache is disabled, regardless of where the call appears in a chain
// (unlike gating the method call itself, this can't be bypassed by Go's promoted-method
// return-type fallthrough on embedded checker types).
func GateProcess(cnf *Conf, check *ec.ProcessChecker) *ec.ProcessChecker {
	if processCacheDisabled(cnf) {
		return nil
	}
	return check
}
```

Then I do this:

```
diff --git a/tests/policytests/kprobes.go b/tests/policytests/kprobes.go
index fe0b97b06..ec2ab47a1 100644
--- a/tests/policytests/kprobes.go
+++ b/tests/policytests/kprobes.go
@@ -87,8 +87,8 @@ spec:
 `).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
        myBin := c.TestBinary("null-mount")
        argChecker := ec.NewProcessKprobeChecker("kprobe-null-string").
-               WithProcess(ec.NewProcessChecker().
-                       WithBinary(sm.Full(myBin))).WithArgs(ec.NewKprobeArgumentLis
tMatcher().
+               WithProcess(policytest.GateProcess(c, ec.NewProcessChecker().
+                       WithBinary(sm.Full(myBin)))).WithArgs(ec.NewKprobeArgumentLi
stMatcher().
                WithOperator(lc.Ordered).
                WithValues(
                        ec.NewKprobeArgumentChecker().WithErrorArg(ec.NewKprobeError
Checker().WithMessage(sm.Full("Bad address for basic type"))),
```

It requires that all the tests call `policytest.GateProcess` for the argument of `WithProcess` and `WithParent`, which is not great, but we don't have to worry about the ordering of the chained functions or keeping `ProcessChecker`'s function implementations up-to-date with `ec.ProcessChecker`. What do you think?

tests/policytests/kprobes.go
```diff
@@ -87,8 +87,8 @@ spec:
 `).AddScenario(func(c *policytest.Conf) *policytest.Scenario {
 	myBin := c.TestBinary("null-mount")
 	argChecker := ec.NewProcessKprobeChecker("kprobe-null-string").
-		WithProcess(ec.NewProcessChecker().
-			WithBinary(sm.Full(myBin))).WithArgs(ec.NewKprobeArgumentListMatcher().
+		WithProcess(policytest.GateProcess(c, ec.NewProcessChecker().
```
FedeDP: I am probably a bit late; but is there a way to avoid this call, and directly act on the `Scenario.EventChecker` instead?
This would make the change invisible to the single tests.
Pseudo code (again, not sure if possible but might be worth investigating):
```go
func (r *LocalRunner) RunScenario(
	l *slog.Logger, scenario *Scenario, polHandler *PolicyHandler, testConf *TestConf,
) ScenarioRes {
	....
    // Pseudo code for the check
    for i, checker := range scenario.EventChecker {
         if checker.type() == "process" {
             checker = gateProcess(r.Conf, checker)
             scenario.EventChecker.update(i, checker)
         }
    }
```

I think we miss some API to do so, but perhaps we can add them.
What i fear is that in the future we end up with multiple `GateFoo()` and each test checker declaration becomes a real mess.

andrewstrohman: Good point. I updated the PR to do this.

FedeDP (approved):
LGTM, thank you!

kkourt (approved):
Thanks Andy!

PR #5356, opened 2026-07-30
ci: skip vmtests artifact delete on fork PRs

### LINKED ISSUE #2010: vmtests post-test delete-artifact fails on fork

amarkdotdev:
Fixes #2010

### Description

`geekyeggo/delete-artifact` in the vmtests / vmpolicytests `post-test` jobs fails on pull requests from forks with `Resource not accessible by integration`, because the restricted `GITHUB_TOKEN` cannot delete artifacts.

Skip the cleanup job for fork PRs. Same-repo PRs and pushes to `main` still delete the build artifact on overall success.

### Changelog

```release-note
ci: skip vmtests artifact deletion on fork PRs
```

### DIFF
--- a/.github/workflows/vmpolicytests.yaml
+++ b/.github/workflows/vmpolicytests.yaml
@@ -169,7 +169,8 @@ jobs:
   post-test:
     runs-on: ubuntu-latest
     needs: [test]
-    if: success()
+    # Fork PRs get a read-only GITHUB_TOKEN that cannot delete artifacts.
+    if: success() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository)
     steps:
       # delete the built binaries from the artifacts in case of overall success
       - uses: geekyeggo/delete-artifact@176a747ab7e287e3ff4787bf8a148716375ca118 # v6.0.0
--- a/.github/workflows/vmtests.yml
+++ b/.github/workflows/vmtests.yml
@@ -177,7 +177,8 @@ jobs:
   post-test:
     runs-on: ubuntu-latest
     needs: [test]
-    if: success()
+    # Fork PRs get a read-only GITHUB_TOKEN that cannot delete artifacts.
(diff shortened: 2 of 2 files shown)


FedeDP (approved):
LGTM, thank you!

amarkdotdev:
@FedeDP Thanks for the approval!

Two CI jobs are still red and look infrastructure-related rather than caused by this change:

1. **validate labels** — the job printed `please add a release-note/* label` with an empty `LABELS` env, but the PR already has `release-note/ci` (plus `kind/enhancement`, `area/ci`). Might be a fork-PR label visibility issue?
2. **Build Windows ebpf programs** — failed while installing LLVM / on a dirty checkout (`local changes would be overwritten by checkout`), which seems unrelated to skipping the vmtests artifact delete on forks.

Happy to push an empty commit to retrigger if that helps, or take any other guidance.

FedeDP:
Ehy! Thanks to you for the PR!

>  Might be a fork-PR label visibility issue?

Had to manually approve the re-run of the workflow :)

> which seems unrelated to skipping the vmtests artifact delete on forks.

Yes, a rebase on top of main would fix this!

Can you please drop the "trigger CI" commit? Thanks!

amarkdotdev:
@FedeDP Rebased onto latest `main` and dropped the empty `ci: retrigger checks after approval` commit. The PR now contains only the real fix commit.

mtardy (approved):
thanks, this was an old issue.

PR #5359, opened 2026-07-31
fix(ci): use actions/checkout to clone microsoft/ntosebpfext in windows CI

FedeDP:
Fixes CI issue: https://github.com/cilium/tetragon/actions/runs/30614449254/job/91104305782?pr=5330, https://github.com/cilium/tetragon/actions/runs/30593881974/job/91041887036?pr=5357

Also, bumps microsoft/ntosebpfext to v0.7.0 (latest release).

### DIFF
--- a/.github/workflows/windows-build-smoke-test.yml
+++ b/.github/workflows/windows-build-smoke-test.yml
@@ -96,24 +96,23 @@ jobs:
           Write-Output "  $(& 'C:\Program Files\LLVM\bin\clang.exe' --version)"
 
       - name: Clone msft/ntosebpfext
-        id: download-ntosebpfet
-        shell: pwsh
-        working-directory: ${{ runner.temp }}
-        run: |
-          git clone --recursive https://github.com/microsoft/ntosebpfext.git
-          cd ${{ runner.temp }}\ntosebpfext
-          git checkout tags/v0.6.0
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          repository: 'microsoft/ntosebpfext'
+          ref: 'v0.7.0'
+          submodules: true
+          path: 'ntosebpfext'
 
       - name: Copy Process_monitor.c file
         shell: pwsh
         run: |
           Copy-Item -Path "${{ github.workspace }}/bpf/windows/process_monitor.c" `
-                    -Destination "${{ runner.temp }}/ntosebpfext/tools/process_monitor_bpf/process_monitor.c" `
+                    -Destination "${{ github.workspace }}/ntosebpfext/tools/process_monitor_bpf/process_monitor.c" `
                     -Force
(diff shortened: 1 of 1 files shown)


.github/workflows/windows-build-smoke-test.yml
```diff
@@ -96,24 +96,23 @@ jobs:
...
+          submodules: true
+          path: 'ntosebpfext'
 
       - name: Copy Process_monitor.c file
         shell: pwsh
         run: |
           Copy-Item -Path "${{ github.workspace }}/bpf/windows/process_monitor.c" `
-                    -Destination "${{ runner.temp }}/ntosebpfext/tools/process_monitor_bpf/process_monitor.c" `
+                    -Destination "${{ github.workspace }}/ntosebpfext/tools/process_monitor_bpf/process_monitor.c" `
                     -Force
```
kkourt, line 111: Out of curiosity, why this change?

kkourt (approved):

.github/workflows/windows-build-smoke-test.yml
FedeDP: Oh that's because `actions/checkout.path` is relative to `${{ github.workspace }}`, therefore we are cloning the repo in the `${{ github.workspace }}/ntosebpfext` folder now.

.github/workflows/windows-build-smoke-test.yml
```diff
@@ -96,24 +96,23 @@ jobs:
...
-        shell: pwsh
-        working-directory: ${{ runner.temp }}
-        run: |
-          git clone --recursive https://github.com/microsoft/ntosebpfext.git
-          cd ${{ runner.temp }}\ntosebpfext
-          git checkout tags/v0.6.0
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          repository: 'microsoft/ntosebpfext'
+          ref: 'v0.7.0'
```
ygvalent, line 102: I am not confident that we can upgrade this, @ExceptionalHandler  would be able to say more here.

ExceptionalHandler:
@FedeDP Why do we think the two issues started happening without anything changing in recent past ? 

I am surprised the CI didn't break when you upgraded ntosebpfext to 7.0 keeping the efw version the same. But since it is green, we should be okay to upgrade.

ExceptionalHandler (approved):

FedeDP:
@ExceptionalHandler the issue is withing the ntosebpfext repo; this comes from a fresh clone of it on my laptop:
```
$ LC_ALL=C git checkout v0.6.0
error: Your local changes to the following files would be overwritten by checkout:
	scripts/setup_build/setup_build.vcxproj
Please commit your changes or stash them before you switch branches.
Aborting
```
Same issue we have in CI.

mtardy (approved):
can we merge?

PR #5367, opened 2026-07-31
ci tarball creation reusable action for package-e2e and vmpolicy tests

dangome3:
### Description
There is some duplication for building the tarball step in `.github/workflows/packages-e2e-tests.yaml` and `.github/workflows/vmpolicytests.yaml`. This PR contains a reusable action so we avoid duplicated code


dangome3:
I'm thinking about using this action in `build-images-release.yml` too, by removing the tarball-release make target and moving the `mv+sha256sum` commands into the workflow. Is there any hard reason not to make this change and keep it as is?
cc: @mtardy @kkourt

FedeDP (approved):
I like it!

.github/workflows/vmpolicytests.yaml
```diff
@@ -35,24 +35,24 @@ jobs:
...
 
-      - name: Getting version tag
-        id: tag
-        run: echo "tag=$(make version)" >> $GITHUB_OUTPUT
-        working-directory: go/src/github.com/cilium/tetragon
+      - name: Build tarball
+        id: tarball
+        uses: ./go/src/github.com/cilium/tetragon/.github/actions/build-tarball
+        with:
+          working-directory: go/src/github.com/cilium/tetragon
```
kkourt: Does this mean that we are building tarball multiple times? If so, would it make sense to only build it once and then reuse it across all workflows that needed?

dangome3: Sure! that makes more sense. Let me re-organize the workflows so we let this happen

dangome3: For context: I'm running now policy tests in the package e2e workflow, this way we only build once. Since the policy tests had the `on: push: main` trigger I leveraged the reusable action to create a that same check without duplicating so much code for not losing this check.

kkourt: Cool! Can you please rename the `package-e2e-tests.yaml` to `tarball-tests.yaml`? I feel it's more appropriate.

kkourt: Also, would it make sense to have ~`package-e2e-tests.yaml`~ `tarball-tests.yaml` run on `main` instead of introducing a new action?

dangome3: Good point! My initial thinking was that running the full `package-e2e-tests` suite on every push to main might be too heavy. But you're right that a dedicated new workflow isn't necessary here and I already fixed it

kkourt (approved):

PR #5377, opened 2026-08-05
new: uprobe set action

FedeDP:
### Description

In a similar vein to #4990 , add support to set function call **integer arguments** in uprobes, via `Set` action (that was previously `USDTs` only).
Again, just like #4990 , the feat is mostly a proxy on top of `Override.argRegs` with an improved UX.

If, in the future, we'd like to expand it to also support string arguments, we should:
* leverage `bpf_probe_write_user`
* add a secondary map to hold `dst_reg, char new_val[256]`
  for each configured `Set` action
* find a way to use a "scratch" address to hold our `new_val` string

The latter is the hardest one, since there is no clear way on how to find the `scratch` are; a full example would be:
```C
char new_str[] = "world!\n";

__u64 scratch_addr = ?????;

long ret = bpf_probe_write_user((void *)scratch_addr, new_str, sizeof(new_str));
if (ret != 0)
    return 0;

ctx->rdi = scratch_addr;
```

and, as you can see, the `scratch_addr` must be there and writable from the uprobe context.

### Changelog

```release-note
new: uprobe set action
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -860,7 +860,7 @@ matches. They are defined under `matchActions` and currently, the following
 - [TrackSock action](#tracksock-action)
 - [UntrackSock action](#untracksock-action)
 - [Notify Enforcer action](#notify-enforcer-action)
-- [USDT Set action](#usdt-set-action)
+- [Set action](#set-action)
 
 {{< note >}}
 `Sigkill`, `Override`, `Post`,
@@ -1606,11 +1606,42 @@ spec:
       - action: "Sigkill"
 ```
 
-### USDT Set action
+### Set action
 
-The `Set` action is specific for USDT probes and allows tetragon to write
-value to configured USDT probe argument. This argument needs to meet few
-conditions:
+The `Set` action is specific for USDT and uprobe probes.
+It uses the following arguments:
+
+- The `argIndex` defines position of the argument.
+- The `argValue` defined the actual value to write.
+
+#### Uprobe
+
+For uprobes, the `Set` action allows setting the value of the argument at the given index.
+The argument needs to meet a few conditions:
+
+- It must be an integer parameter (`argValue` holds an `uint32`)
+- `argIndex` must be between 0 and 5 (the [tetragon maximum number of arguments](https://github.com/cilium/tetragon/blob/v1.7.1/pkg/api/tracingapi/client_kprobe.go#L684))
+- `argIndex` refers to the position of the argument in the traced function, starting from 0 for the first argument
+
+Example policy follows:
+```yaml
+spec:
+  uprobes:
+  - path: my-binary
+    symbols:
+    - "pizza"
+    selectors:
+    - matchActions:
+      - action: Set
+        argIndex: 0
+        argValue: 42
+```
+This policy will set the value of the first argument of the `pizza()` function to `42`.
+
+#### USDT
+
+For USDT probes, the `Set` action allows writing a value to a configured probe argument. 
+The argument needs to meet a few conditions:
 
 - It's stored in memory as `USDT deref` argument
 - It has size of 4 bytes
@@ -1678,11 +1709,6 @@ spec:
           argIndex: 0
           argValue: 1
 ```
-
-The `Set` action uses following arguments:
-
-- The `argIndex` defines position of the return argument.
-- The `argValue` defined the actual value to write.
 
 ## Selector Semantics
 
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -859,7 +859,7 @@ matches. They are defined under `matchActions` and currently, the following
 - [TrackSock action](#tracksock-action)
 - [UntrackSock action](#untracksock-action)
 - [Notify Enforcer action](#notify-enforcer-action)
-- [USDT Set action](#usdt-set-action)
+- [Set action](#set-action)
 
 {{< note >}}
 `Sigkill`, `Override`, `Post`,
@@ -1605,11 +1605,41 @@ spec:
       - action: "Sigkill"
 ```
 
-### USDT Set action
+### Set action
 
-The `Set` action is specific for USDT probes and allows tetragon to write
-value to configured USDT probe argument. This argument needs to meet few
-conditions:
+The `Set` action is specific for USDT and uprobe probes.
+It uses the following arguments:
+
+- The `argIndex` defines position of the argument.
+- The `argValue` defined the actual value to write.
+
+#### Uprobe
+
+For uprobes, the `Set` action allows setting the value of the argument at the given index.
+The argument needs to meet a few conditions:
+
+- It must be an integer parameter (`argValue` holds an `uint32`)
+- `argIndex` must be between 0 and the tetragon maximum number of arguments (5)
+
+Example policy follows:
+```yaml
+spec:
+  uprobes:
+  - path: my-binary
+    symbols:
+    - "pizza"
+    selectors:
+    - matchActions:
+      - action: Set
+        argIndex: 0
+        argValue: 42
+```
+This policy will set the value of the first argument of the `pizza()` function to `42`.
+
+#### USDT
+
+For USDT probes, the `Set` action allows writing a value to a configured probe argument. 
+The argument needs to meet a few conditions:
 
 - It's stored in memory as `USDT deref` argument
 - It has size of 4 bytes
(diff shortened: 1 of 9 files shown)


pkg/selectors/kernel_regs_amd64.go
```diff
@@ -53,3 +53,38 @@ func parseOverrideRegs(k *KernelSelectorState, selIdx int, values []string, errV
 	k.regs[selIdx] = regs
 	return nil
 }
+
+func parseSetRegs(k *KernelSelectorState, selIdx int, argIndex, argValue uint32) error {
```
olsajiri, line 57: should we allow uint64 for argValue?

pkg/sensors/tracing/genericuprobe.go
```diff
@@ -590,6 +590,19 @@ func validateUprobeSpec(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) err
 		return fmt.Errorf("uprobe Override action needs exactly one of either argNewSymbol, argNewOffset or argNewAddr defined; %d found", numArgNewArgs)
 	}
 	state.overrideSymbol = numArgNewArgs == 1
+
+	if selectors.HasSet(spec.Selectors) {
```
kkourt: Calling `HasSet` here seems like a waste of CPU cycles compared to just iterating over the selectors (i.e., omitting `if selectors.HasSet()`).

pkg/selectors/kernel_regs_amd64.go
```diff
@@ -53,3 +53,38 @@ func parseOverrideRegs(k *KernelSelectorState, selIdx int, values []string, errV
...
+		val = fmt.Sprintf("rsi=%d", argValue)
+	case 2:
+		val = fmt.Sprintf("rdx=%d", argValue)
+	case 3:
+		val = fmt.Sprintf("rcx=%d", argValue)
+	case 4:
+		val = fmt.Sprintf("r8=%d", argValue)
+	case 5:
+		val = fmt.Sprintf("r9=%d", argValue)
+	}
```
kkourt, line 72: I'm guessing after `r9` we start passing arguments in the stack? Maybe it's worth having an explicit error for that case?

pkg/selectors/kernel_regs_arm64.go
```diff
@@ -52,3 +52,24 @@ func parseOverrideRegs(k *KernelSelectorState, selIdx int, values []string, errV
 	k.regs[selIdx] = regs
 	return nil
 }
+
+func parseSetRegs(k *KernelSelectorState, selIdx int, argIndex, argValue uint32) error {
+	val := fmt.Sprintf("x%d=%d", argIndex, argValue)
```
kkourt, line 57: Similar to x86, can we add an explicit error for when we ran out of registers and need to use the stack?

kkourt (commented):
Just a couple of nits, but, otherwise, LGTM.

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1605,11 +1605,41 @@ spec:
...
+- The `argIndex` defines position of the argument.
+- The `argValue` defined the actual value to write.
+
+#### Uprobe
+
+For uprobes, the `Set` action allows setting the value of the argument at the given index.
+The argument needs to meet a few conditions:
+
+- It must be an integer parameter (`argValue` holds an `uint32`)
+- `argIndex` must be between 0 and the tetragon maximum number of arguments (5)
```
mtardy: nit: maybe link to the place where we have the limitation to avoid deduplication of these constants (that will be potentially bumped one day)

mtardy: please disambiguate so in this case argIndex is the index of the argument (as in `args[0].index`) or the index of entry in the array defining the args in the CRD?

I imagine here it's `args[i].index` but we have the opposite in CEL so maybe it's nice to have a clear doc (cc @kkourt)

mtardy (commented):
lgtm, just a few comments on the docs.

Aside I think the `Set` name on the action was maybe a bit too broad, from someone without context I don't exactly get the difference between Override and Set at a first glance. But it was merged a long time ago now I see from Jiri's PR...

pkg/selectors/kernel_regs_amd64.go
FedeDP: I didn't want to touch our CRD:
```
// An arg value for the set action
ArgValue uint32 `json:"argValue"`
```
Am not sure whether this really is a breaking change though. We can do it later IMO.

pkg/selectors/kernel_regs_amd64.go
FedeDP: We already have it in genericuprobe:
```go
			if action.ArgIndex >= api.EventConfigMaxArgs {
				return fmt.Errorf("uprobe Set action argIndex %d out of range", action.ArgIndex)
			}
```

pkg/selectors/kernel_regs_arm64.go
FedeDP: Same as above!

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: A couple of lines prior to that, when describing `Uprobe` `Set` action, i wrote:
> For uprobes, the `Set` action allows setting the value of the argument at the given index.

I think it is clear that we refer to the index of the argument in the function signature; do you think we need to expand a bit?

mtardy: It's just that we have other features that make use of the index of the array instead of the index field value, and I think it's becoming confusing so I would add a layer of confirmation that we are talking about index the field or index the number defining your place in the array. I don't think "index" is enough as it can mean two things now in our semantic.

See https://tetragon.io/docs/concepts/tracing-policy/selectors/#matchcel.

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1606,11 +1606,41 @@ spec:
...
+- The `argIndex` defines position of the argument.
+- The `argValue` defined the actual value to write.
+
+#### Uprobe
+
+For uprobes, the `Set` action allows setting the value of the argument at the given index.
+The argument needs to meet a few conditions:
+
+- It must be an integer parameter (`argValue` holds an `uint32`)
+- `argIndex` must be between 0 and the [tetragon maximum number of arguments](../../../../../../pkg/api/tracingapi/client_kprobe.go#L684)
```
mtardy: ah this wouldn't work, btw you can check that what you wrote is working by [looking at the preview](https://deploy-preview-5377--tetragon.netlify.app/docs/concepts/tracing-policy/selectors/#uprobe) or building the docs locally. I don't think we have a system right now to reference stuff from the code in the docs.

I was thinking of places in the docs like https://deploy-preview-5377--tetragon.netlify.app/docs/concepts/tracing-policy/selectors/#limitations, but maybe it's written nowhere that we can just take 5 args already

FedeDP: Oh right i assume i need a proper link then, thanks for catching it!

mtardy: sorry it's never ending but pointing to a line using main will drift from what you are pointing. I'd point either to a release tag version. Again I assumed we had some docs page where it was referenced already but it's not so if you prefer just to put back the const in the text it's fine, sorry for all that noise.

mtardy (commented):
comments in thread

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: No problem really, appreciate the effort to make it clearer :) 
Pointing to a tag is the same than putting the const, but requires one more level of indirection. Let's put the const with a link to its place in the latest tag!

olsajiri (approved):
lgtm, thanks

mtardy (approved):
I see you haven't touched https://github.com/cilium/tetragon/pull/5377#discussion_r3855374862 really but it's  just an enhancement/clarification, lgtm overall.

FedeDP:
> I see you haven't touched https://github.com/cilium/tetragon/pull/5377#discussion_r3855374862 really but it's just an enhancement/clarification, lgtm overall.

Indeed i added:
> - `argIndex` refers to the position of the argument in the traced function, starting from 0 for the first argument

mtardy:
> > I see you haven't touched [#5377 (comment)](https://github.com/cilium/tetragon/pull/5377#discussion_r3855374862) really but it's just an enhancement/clarification, lgtm overall.
> 
> Indeed i added:
> 
> > * `argIndex` refers to the position of the argument in the traced function, starting from 0 for the first argument

ah sorry I overlooked this thanks a lot! I have been trying to use the compare thing from github and when there is a rebase it's hard to follow what changed without manually looking.

PR #5379, opened 2026-08-06
fix(tetra): honor retry settings for all gRPC clients

immanuwell:
Fixes

### Description

The retry flag was half wired for EventLog. Also, retries set to zero made gRPC reject the client before any RPC.

The fix disables the policy cleanly at zero and applies the shared policy to EventLog. Added tests for both paths.

Repro:

    tetra status -retries=0 -timeout=100ms -server-address=127.0.0.1:54329

Before: client creation failed with invalid retry policy, MaxAttempts 1.

    tetra eventlog get -retries=1 -timeout=3s -server-address=127.0.0.1:54329

Before: UNAVAILABLE returned after one attempt. Now the client retries once.

### Changelog

```release-note
Fix tetra retry handling for zero retries and EventLog RPCs
```

### DIFF
--- a/cmd/tetra/common/client.go
+++ b/cmd/tetra/common/client.go
@@ -37,8 +37,8 @@ import (
 // between 0 and the final duration that was computed via to the params:
 // https://github.com/grpc/grpc-go/blob/v1.65.0/stream.go#L702
 func RetryPolicy(retries int) string {
-	if retries < 0 {
-		// gRPC should ignore the invalid retry policy but will issue a warning,
+	if retries <= 0 {
+		// An empty service config disables application-level retries.
 		return "{}"
 	}
 	// maxAttempt includes the first call
--- a/cmd/tetra/common/client.go
+++ b/cmd/tetra/common/client.go
@@ -47,7 +47,10 @@ func RetryPolicy(retries int) string {
 	// since we need to provide a value >0
 	return fmt.Sprintf(`{
 	"methodConfig": [{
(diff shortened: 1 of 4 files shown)


cmd/tetra/common/client.go
```diff
@@ -47,7 +47,10 @@ func RetryPolicy(retries int) string {
 	// since we need to provide a value >0
 	return fmt.Sprintf(`{
 	"methodConfig": [{
-	  "name": [{"service": "tetragon.FineGuidanceSensors"}],
+	  "name": [
+		{"service": "tetragon.FineGuidanceSensors"},
+		{"service": "tetragon.EventLogService"}
+	  ],
```
mtardy, line 53: yeah indeed, now I wonder why EventLogService was created as a separate service and not in FineGuidanceSensors (cc @FedeDP)

mtardy (commented):
thanks, I think you might want to split this into two patchces: one for the EventLogService and one for the <= change?

cmd/tetra/common/client.go
FedeDP: perhaps i didn't want it to be part of the fgs service, but i really can't recall :(

immanuwell:
@mtardy done, split this into 2 separate patches

mtardy (commented):
thanks this looks good, could you include the details of your PR description in the commit changelog?

immanuwell:
@mtardy done, expanded both commit changelogs

mtardy (approved):
thanks, please wrap the commit log also next time! thanks for the patches

PR #5380, opened 2026-08-06
uprobe/usdt: Reject GetUrl and DnsLookup Actions

kevsecurity:
Fixes: https://github.com/cilium/security-tetragon/issues/10

### Description
The GetUrl and DnsLookup actions are not supported for uprobes or usdt. Using them can cause crashes as required state isn't initialised. This pull request checks for GetUrl and DnsLookup actions in uprobe and usdt and returns an error if any are found.
    
Reported-by: Artem Cherezov <cherez0ff@proton.me>

### Changelog
Fix crash when unsupported GetUrl or DnsLookup actions are included in uprobe or usdt tracing policies.

```release-note
Fix crash when unsupported GetUrl or DnsLookup actions are included in uprobe or usdt tracing policies.
```

### DIFF
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -1814,6 +1814,18 @@ func HasOverride(selectors []v1alpha1.KProbeSelector) bool {
 	return false
 }
 
+func HasGetUrlOrDnsLookup(selectors []v1alpha1.KProbeSelector) bool {
+	for _, s := range selectors {
+		for _, action := range s.MatchActions {
+			act := actionTypeTable[strings.ToLower(action.Action)]
+			if act == ActionTypeGetUrl || act == ActionTypeDnsLookup {
+				return true
+			}
+		}
+	}
+	return false
+}
+
 func CountOverrideArgNewSymbolAddrOffset(selectors []v1alpha1.KProbeSelector) int {
 	for _, s := range selectors {
 		for _, action := range s.MatchActions {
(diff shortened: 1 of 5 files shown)


mtardy (commented):
just checking on this, you can also probably write a CEL expression for X-Validation in the CRD directly for even better UX.

kevsecurity:
> just checking on this, you can also probably write a CEL expression for X-Validation in the CRD directly for even better UX.

Do you have an example?

mtardy:
> > just checking on this, you can also probably write a CEL expression for X-Validation in the CRD directly for even better UX.
> 
> Do you have an example?

yeah https://github.com/cilium/tetragon/pull/5071

FedeDP:
Also, once https://github.com/cilium/tetragon/pull/5322 is merged, the CEL expression X-validation will also apply to nok8s builds ;)

kevsecurity:
> just checking on this, you can also probably write a CEL expression for X-Validation in the CRD directly for even better UX.

I've taken a look and I think it will be complex. We only want the rule to check for GetUrl or DnsLookup on uprobe and usdt, so the validation needs to be at the uprobe/usdt level (doing it at the selector level doesn't expose the type of probe I think). And doing this at the uprobe/usdt level means validation that traverses the list of selectors to then check the lists of actions. Having never written CEL, this feels a bit challenging!

mtardy:
> > just checking on this, you can also probably write a CEL expression for X-Validation in the CRD directly for even better UX.
> 
> I've taken a look and I think it will be complex. We only want the rule to check for GetUrl or DnsLookup on uprobe and usdt, so the validation needs to be at the uprobe/usdt level (doing it at the selector level doesn't expose the type of probe I think). And doing this at the uprobe/usdt level means validation that traverses the list of selectors to then check the lists of actions. Having never written CEL, this feels a bit challenging!

it's not a hard req no pressure, we can always merge this.

mtardy (approved):

PR #5385, opened 2026-08-07
fix(selectors): cap matchPIDs.values at 4 and bounds-check selector buffer writes

### LINKED ISSUE #5384: matchPIDs.values with >4 entries panics and crashes the Tetragon agent (unbounded write into fixed 4096-byte selector buffer)

`ParseMatchPid` encodes every PID in `matchPIDs.values` into a fixed `[4096]byte` selector buffer (`pkg/selectors/selectors.go`) with no cap on the count and no bounds check in the writer (`WriteSelectorByteArray`, `pkg/selectors/selectors.go:355`). Every other list-valued selector (`matchArgs`, `matchNamespaces`) caps at 4 values to match `MAX_SELECTOR_VALUES` in `bpf/process/pfilter.h:309` `matchPIDs` was missing this cap. A policy with >~1024 PID values overflows the buffer, and Go's bounds check turns it into an unrecovered panic that kills the whole agent process no `recover()` exists anywhere in the call chain.

Reachable via gRPC `AddTracingPolicy` (default `localhost:54321`, unauthenticated) or a `--tracing-policy` startup file. No RBAC/auth required for the local gRPC path.

**Reproduced on:** Ubuntu, kernel 7.0.0
```yaml
kind: TracingPolicy
metadata: { name: crash }
spec:
  kprobes:
  - call: "sys_read"
    syscall: true
    selectors:
    - matchPIDs:
      - operator: "In"
        values: [0, 1, 2, ..., 1999]   # 2000 values
```

```
$ tetra tracingpolicy add crash.yaml
...
panic: runtime error: index out of range [4096] with length 4096

goroutine 89 [running]:
github.com/cilium/tetragon/pkg/selectors.WriteSelectorByteArray(...)
	pkg/selectors/selectors.go:357
github.com/cilium/tetragon/pkg/selectors.ParseMatchPid(...)
	pkg/selectors/kernel.go:378
github.com/cilium/tetragon/pkg/selectors.ParseMatchPids(...)
	pkg/selectors/kernel.go:385
...
github.com/cilium/tetragon/pkg/server.(*Server).AddTracingPolicy(...)
	pkg/server/server.go:274
```

Agent process exits. Any local unprivileged process on the same host can trigger this against the default TCP gRPC listener, repeatedly denial of service against the security monitor (stops detection + enforcement).

**Impact:** DoS (CWE-400 / CWE-248), memory-safe (Go panic, not OOB write / CWE-787).

### Tetragon Version

v1.8.0-pre.0-559-g26f6496dd

### Kernel Version

7.0.0-1006-aws

### Kubernetes Version

_No response_

### Bugtool

_No response_

### Relevant log output

```shell

```

### Anything else?

_No response_

hanshal101:
Fixes ##5384
Closes #5384 

Fixes agent crash from unbounded `matchPIDs.values` overflowing the fixed 4096-byte selector encoding buffer (panic: `index out of range [4096] with length 4096`).

**Root cause fix:** `pidSelectorValue` now caps `matchPIDs.values` at 4, matching the existing cap on `matchArgs`/`matchNamespaces` and the kernel's own `MAX_SELECTOR_VALUES=4` in `bpf/process/pfilter.h` the kernel matcher already ignores anything past the 4th value, so this closes both the overflow and a pre-existing dead-data gap.

**Defense-in-depth:** `KernelSelectorData` now tracks a sticky error; every `WriteSelector*`/`AdvanceSelectorLength` helper checks bounds before writing and no-ops on overflow instead of panicking. `createKernelSelectorState` surfaces that error once, as a normal returned error. Protects against any future selector type that forgets a cap.

**Schema hardening:** added `maxItems: 4` to `matchPIDs.values` in both CRD YAMLs (regenerated via `controller-gen`) and the embedded gRPC-path JSON schema, so malformed policies are rejected before reaching the parser.

**Verification (Ubuntu, kernel 7.0.0):**

Before fix the agent crashes:
```
panic: runtime error: index out of range [4096] with length 4096
```

After fix: dynamic add via gRPC:
```
$ sudo ./tetra tracingpolicy add crash.yaml
Error: failed to add tracing policy: rpc error: code = Unknown desc = policy handler 'tracing' failed loading policy 'crash': parseMatchPids error: matchpid error: matchPIDs supports up to 4 values per filter (current number of values is 2000)
```
Agent stays up, no panic.

After fix startup-file load (`--tracing-policy crash.yaml`):
```
level=warn msg="Server AddTracingPolicy request failed" error="policy handler 'tracing' failed loading policy 'crash': parseMatchPids error: matchpid error: matchPIDs supports up to 4 values per filter (current number of values is 2000)" metadata.name=crash
```
Clean warning log, agent continues running normally.

**Tests:** added `TestPidSelectorValueTooManyValues`; updated `TestPidSelectorValue` for the new error return. `go build`/`go vet` clean on `pkg/selectors`, `pkg/tracingpolicy`, `pkg/k8s`.

### DIFF
--- a/pkg/selectors/kernel_test.go
+++ b/pkg/selectors/kernel_test.go
@@ -202,9 +202,15 @@ func TestPidSelectorFlags(t *testing.T) {
 func TestPidSelectorValue(t *testing.T) {
 	pid := &v1alpha1.PIDSelector{Operator: "In", Values: []uint32{1, 2, 3}, IsNamespacePID: true, FollowForks: true}
 	expected := []byte{0x1, 0x0, 0x0, 0x0, 0x2, 0x0, 0x0, 0x0, 0x3, 0x0, 0x0, 0x0}
-	if b, l := pidSelectorValue(pid); bytes.Equal(b, expected) == false || l != 12 {
-		t.Errorf("pidSelectorValue: expected %v actual %v\n", expected, b)
+	if b, l, err := pidSelectorValue(pid); err != nil || bytes.Equal(b, expected) == false || l != 12 {
+		t.Errorf("pidSelectorValue: expected %v actual %v (err: %v)", expected, b, err)
 	}
+}
+
+func TestPidSelectorValueTooManyValues(t *testing.T) {
+	pid := &v1alpha1.PIDSelector{Operator: "In", Values: []uint32{1, 2, 3, 4, 5}, IsNamespacePID: true, FollowForks: true}
+	_, _, err := pidSelectorValue(pid)
+	require.Error(t, err)
 }
 
 func TestNamespaceValue(t *testing.T) {
--- a/pkg/selectors/kernel_test.go
+++ b/pkg/selectors/kernel_test.go
@@ -31,7 +31,7 @@ import (
 )
 
 func TestWriteSelectorUint32(t *testing.T) {
-	k := &KernelSelectorState{data: KernelSelectorData{off: 0}}
+	k := &KernelSelectorState{data: KernelSelectorData{}}
 	d := &k.data
 	v := uint32(0x1234abcd)
 
@@ -41,7 +41,8 @@ func TestWriteSelectorUint32(t *testing.T) {
 			d.e[0], d.e[1], d.e[2], d.e[3])
 	}
 
-	d.off = 1024
+	// grow by 1020 from offset 4 to reach offset 1024, then write uint32 at [1024:1028]
+	d.grow(1020)
 	WriteSelectorUint32(d, v)
 	if d.e[1027] != 0x12 || d.e[1026] != 0x34 || d.e[1025] != 0xab || d.e[1024] != 0xcd {
 		t.Errorf("SelectorStateWrite offset(1024) failed: %x %x %x %x\n",
@@ -50,7 +51,7 @@ func TestWriteSelectorUint32(t *testing.T) {
 }
 
 func TestWriteSelectorLength(t *testing.T) {
-	k := &KernelSelectorState{data: KernelSelectorData{off: 0}}
+	k := &KernelSelectorState{data: KernelSelectorData{}}
 	d := &k.data
 	v := uint32(0x1234abcd)
 
@@ -76,7 +77,7 @@ func TestWriteSelectorLength(t *testing.T) {
 }
 
 func TestWriteSelectorByteArray(t *testing.T) {
-	k := &KernelSelectorState{data: KernelSelectorData{off: 0}}
+	k := &KernelSelectorState{data: KernelSelectorData{}}
 	d := &k.data
 	v := []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 0xa, 0xb, 0xc, 0xd, 0xe, 0xf}
 
@@ -252,7 +253,7 @@ func TestParseMatchCmdArgs(t *testing.T) {
 		// full header.
 		ks := NewKernelSelectorState(nil, nil, false, 0, 0, nil)
 		require.NoError(t, ParseMatchCmdArgs(ks, nil))
-		data := ks.data.e[:ks.data.off]
+		data := ks.data.e
 		require.Len(t, data, 24)
 		assert.Equal(t, uint32(len(data)), binary.LittleEndian.Uint32(data[:4]))
 		for off := 4; off < 24; off += 4 {
@@ -282,7 +283,7 @@ func TestParseMatchCmdArgs(t *testing.T) {
 		ks := NewKernelSelectorState(nil, nil, false, 0, 0, nil)
 		require.NoError(t, ParseMatchCmdArgs(ks, cmdArgs))
 
-		data := ks.data.e[:ks.data.off]
+		data := ks.data.e
 		require.GreaterOrEqual(t, len(data), 24)
 		sectionLength := binary.LittleEndian.Uint32(data[0:4])
 		assert.Equal(t, uint32(len(data)), sectionLength)
@@ -368,7 +369,7 @@ func TestInitKernelSelectorsMatchCmdArgsLayout(t *testing.T) {
 	})
 	require.NoError(t, err)
 
-	data := state.data.e[:state.data.off]
+	data := state.data.e
 	readUint32 := func(offset int) uint32 {
 		return binary.LittleEndian.Uint32(data[offset : offset+4])
 	}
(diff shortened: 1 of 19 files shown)


olsajiri (commented):
thanks for the fix, please split your change to separate commits with:
- yaml updates
- pid matching fix
- that k.fits stuff you introduced

hanshal101:
> thanks for the fix, please split your change to separate commits with:
> 
> * yaml updates
> * pid matching fix
> * that k.fits stuff you introduced

Updated!

olsajiri (commented):
first 2 commits lgtm

I don't like the approach of the sticky error so I tried some other ways..
having each WriteSelector* function to return error seems too much,
but growing selector data as we write to it seems ok-ish.. not sure, please check:  https://github.com/cilium/tetragon/pull/5404/changes/1b457222c3d8bce114754d39c3f26bc337dea1a7 @kkourt

hanshal101:
> first 2 commits lgtm
> 
> I don't like the approach of the sticky error so I tried some other ways.. having each WriteSelector* function to return error seems too much, but growing selector data as we write to it seems ok-ish.. not sure, please check: [1b45722](https://github.com/cilium/tetragon/commit/1b457222c3d8bce114754d39c3f26bc337dea1a7) @kkourt

Thanks for the review! I think the growable-buffer approach is much better than the sticky error one it much cleaner, single check at the end instead of threading bounds checks through every writer. I'll rework in that way and push shortly

hanshal101:
also @olsajiri can you please help me in fixing the failed tests!

pkg/selectors/kernel_test.go
```diff
@@ -201,8 +201,15 @@ func TestPidSelectorFlags(t *testing.T) {
 func TestPidSelectorValue(t *testing.T) {
 	pid := &v1alpha1.PIDSelector{Operator: "In", Values: []uint32{1, 2, 3}, IsNamespacePID: true, FollowForks: true}
 	expected := []byte{0x1, 0x0, 0x0, 0x0, 0x2, 0x0, 0x0, 0x0, 0x3, 0x0, 0x0, 0x0}
-	if b, l := pidSelectorValue(pid); bytes.Equal(b, expected) == false || l != 12 {
-		t.Errorf("pidSelectorValue: expected %v actual %v\n", expected, b)
+	if b, l, err := pidSelectorValue(pid); err != nil || bytes.Equal(b, expected) == false || l != 12 {
+		t.Errorf("pidSelectorValue: expected %v actual %v (err: %v)\n", expected, b, err)
```
mtardy: nit: since it was already there, but not needed
```suggestion
		t.Errorf("pidSelectorValue: expected %v actual %v (err: %v)", expected, b, err)
```

pkg/selectors/kernel_test.go
```diff
@@ -201,8 +201,15 @@ func TestPidSelectorFlags(t *testing.T) {
...
+	}
+}
+
+func TestPidSelectorValueTooManyValues(t *testing.T) {
+	pid := &v1alpha1.PIDSelector{Operator: "In", Values: []uint32{1, 2, 3, 4, 5}, IsNamespacePID: true, FollowForks: true}
+	if _, _, err := pidSelectorValue(pid); err == nil {
+		t.Errorf("pidSelectorValue: expected error for more than 4 values, got nil\n")
```
mtardy:
```suggestion
		t.Errorf("pidSelectorValue: expected error for more than 4 values, got nil")
```

mtardy: if you want you can also use `require.Error` that encapsulates the idea also, not required to do but could be helpful

pkg/selectors/kernel_test.go
```diff
@@ -40,7 +40,7 @@ func TestWriteSelectorUint32(t *testing.T) {
 			d.e[0], d.e[1], d.e[2], d.e[3])
 	}
 
-	d.off = 1024
+	d.grow(1020) // pad out to offset 1024; grow() is what off actually tracks
```
mtardy: maybe change the comment to something like
```
// Grow to offset 1020, then write uint32 at [1020:1024]
```
this was unclear to me

pkg/selectors/selectors.go
mtardy, line 167: here initialize the slice to so that we don't spend our time growing the slice unless it's really necessary and we go beyond the expected size
```go
        data: KernelSelectorData{
            e: make([]byte, 0, KernelBufferSize),
        },
```

pkg/selectors/selectors.go
```diff
@@ -108,9 +108,11 @@ type MatchBinariesSelectorOptions struct {
 	MBSetID uint32
 }
 
+const KernelBufferSize = 4096
+
 type KernelSelectorData struct {
-	off uint32     // offset into encoding
-	e   [4096]byte // kernel encoding of selectors
+	off uint32 // offset into encoding (== len(e))
```
mtardy: I would remove this field as it's redundant to `len(e)` maybe we can even inline KernelSelectorData to be the byte slice (or not if it has method I imagine but that's okay)

pkg/selectors/selectors.go
```diff
@@ -319,26 +323,32 @@ func (k *KernelSelectorState) StringPostfixMapsMaxEntries() int {
 	return maxEntries
 }
 
+func (k *KernelSelectorData) grow(size uint32) []byte {
+	off := k.off
+	k.e = append(k.e, make([]byte, size)...)
+	k.off += size
+	return k.e[off:k.off]
```
mtardy: Then if you remove off you can simplify those
```suggestion
	oldLen := len(k.e)
	k.e = append(k.e, make([]byte, size)...)
	return k.e[oldLen:]
```

pkg/selectors/selectors.go
```diff
@@ -217,8 +219,10 @@ func (k *KernelSelectorState) MatchBinariesPathsMaxEntries() int {
 	return maxEntries
 }
 
-func (k *KernelSelectorState) Buffer() [4096]byte {
-	return k.data.e
+func (k *KernelSelectorState) Buffer() [KernelBufferSize]byte {
```
mtardy: nit: I would make the name more explicit
```suggestion
func (k *KernelSelectorState) CopyToFixedBuffer() [KernelBufferSize]byte {
```

mtardy (changes requested):
Thanks!! Also 2 first commits lgtm to me just a few comments on last commit written by Jiri.

Requesting changes mostly for the slice pre-allocation missing.

mtardy (approved):
Thanks, this lgtm, will let Jiri review it :)

kkourt (commented):
Thanks, LGTM! Can you please rebase your PR?

kkourt (approved):

mtardy (approved):
thanks!

olsajiri (approved):
sorry for delay, thanks

PR #5388, opened 2026-08-09
bpf: clear stale LSM output state before policy filtering

### LINKED ISSUE #5237: BPF-LSM: generic_lsm_output re-emits a stale per-CPU event on every unrelated syscall, producing phantom events for pods that did nothing

With a single socket_connect LSM TracingPolicyNamespaced scoped to one pod (observe/Post only), Tetragon emits a continuous stream of process_lsm events for that pod even when the pod makes no connections at all. Every emitted phantom event is an exact duplicate of the pod's last real connect: same destination, same (immutable) creation timestamp, same content. The stream never stops, and its rate tracks node-wide connect activity, not the target pod's.

Root cause: for an LSM hook, generic_lsm_output is attached as a second, independent program on the hook and runs unconditionally, but the per-CPU event's post flag is only reset after the policy_filter_check early-return in generic_start_process_filter. So any syscall from a process that fails the pod filter skips the reset, and generic_lsm_output re-emits whatever stale event is still in the per-CPU heap.

To reproduce:
1. Apply an observe-only LSM policy scoped to one workload:

kind: TracingPolicyNamespaced
metadata:
  name: lsm-phantom-repro
  namespace: demo
spec:
  podSelector:
    matchLabels: { app: victim }
  lsmhooks:
  - hook: socket_connect
    args:
    - { index: 0, type: socket }
    - { index: 1, type: sockaddr }
    - { index: 2, type: int }
    selectors:
    - matchArgs:
      - { index: 1, operator: Family, values: [AF_INET, AF_INET6] }
      matchActions:
      - { action: Post }
2. Make the victim pod perform one outbound TCP connect (e.g. to 1.1.1.1:443). Observe the expected event.
3. Now leave victim completely idle (no further connects).
4. Watch events: tetra getevents (or the pod's tetragon_events_total{type="PROCESS_LSM"} metric) keeps producing events attributed to victim, all with the 1.1.1.1 destination and the original timestamp from step 2. The rate scales with unrelated connect activity elsewhere on the node.

### Tetragon Version

Tetragon v1.7.0

### Kernel Version

6.8.0-124-generic

### Kubernetes Version

Client Version: v1.35.5+k3s1
Kustomize Version: v5.7.1
Server Version: v1.35.5+k3s1

### Bugtool

_No response_

### Relevant log output

```shell

```

### Anything else?

Root cause (source)
generic_start_process_filter() (bpf/process/generic_calls.h) resets the per-CPU event after the pod-filter early-return:

        return 0;            // <-- non-matching process bails here
...
msg->lsm.post = false;       // <-- reset never reached for filtered-out syscalls
generic_lsm_actions (bpf/process/bpf_generic_lsm_core.c) sets e->lsm.post = postit only for events that pass the filter/args. generic_lsm_output (bpf/process/bpf_generic_lsm_output.c) is attached to the hook via its own section "lsm/generic_lsm_output" (a separate lsm_mac link — it is not a member of the lsm_calls tail-call map) and emits unconditionally when post is set:

if (e->lsm.post)
        generic_output(ctx, MSG_OP_GENERIC_LSM);
Because the per-CPU process_call_heap is shared and generic_lsm_output runs on every hook invocation, a syscall from any process that fails policy_filter_check leaves post=1 (and the full stale event) from the last matching event, and output re-emits it.

Why LSM-only: kprobe uses the identical generic_start_process_filter, but generic_kprobe_output is a tail-call target (kprobe_calls[TAIL_CALL_SEND]) reached only from the filtered chain, so a rejected event never reaches output. LSM is the only sensor that attaches output directly to the hook, so it runs even when the entry bailed at the pod filter.

iodn:
Fixes #5237

### Description
The generic LSM output program can read a stale per CPU message before the core program runs.

In `generic_start_process_filter()`, `lsm.post` and `common.flags` were only cleared after the configuration and policy-filter checks. If an invocation was rejected by one of those checks the function returned early and left the previous state intact. The output program could then emit the last matching event again for an unrelated hook invocation.

This change clears the LSM output state immediately after retrieving the per-CPU message, before any early return.
A same-CPU namespaced regression test was also added to verify that only the expected `file_open` event is delivered.

### Changelog

```release-note
bpf: Prevent namespaced LSM policies from emitting stale duplicate events
```

### DIFF
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -50,6 +50,10 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 	if (!msg)
 		return 0;
 
+	/* Clear state consumed by LSM output before any early return. */
+	msg->lsm.post = false;
+	msg->common.flags = 0;
+
 	/* setup index, check policy filter, and setup function id */
 	msg->idx = get_index(ctx);
 	config = map_lookup_elem(&config_map, &msg->idx);
@@ -81,9 +85,6 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 #ifdef __CAP_CHANGES_FILTER
 	msg->sel.match_cap = 0;
 #endif
-
-	msg->lsm.post = false;
-	msg->common.flags = 0;
 
 	/* Tail call into filters. */
 	tail_call(ctx, calls, TAIL_CALL_FILTER);
--- a/pkg/sensors/tracing/policyfilter_test.go
+++ b/pkg/sensors/tracing/policyfilter_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/google/uuid"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/cilium/tetragon/pkg/api/tracingapi"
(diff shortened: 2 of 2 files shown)


kkourt (changes requested):
Can you please add additional information and context in the commit message?

Specifically:
 1. Why is this fix needed? Does it fix a bug? If so, under what conditions can the bug happen
2. Does the test trigger the bug? if so, how?

Please, also, consider splitting your PR into two patches: one for the fix and one for the test.

alitonia:
Just checking on this one, I see that the PR is closed, and it is the right fix for the bug in main.

Currently `generic_start_process_filter()` clears `lsm.post` only after the filter checks, so unrelated syscalls keep re-emitting stale events.

For your questions:
1. It is a bug. The output program runs on every hook invocation, but the state reset only runs for invocations that pass the filter. Rejected ones leave the previous event's post flag behind, and it gets re-emitted. This hits whenever a scoped LSM Post policy is active and anything else on the node touches the same hook.
To trigger:
You need 1 `matchingProcess` and 1 `nonMatchingProcess` process pin to same CPU,  a namespaced LSM policy in `ns1` on the file_open hook, selector "argument 0 equals <targetFile>", action `Post`, matching  open 1 target file, non matching open `/dev/null`, with no bug you get 1, with the bug you get 6.

iodn:
Thanks for the detailed feedback. I have brought the change back in #5460 and incorporated the suggestions there.

The first commit now explains the bug: a filtered-out LSM hook invocation could leave the per-CPU Post state from an earlier matching invocation intact, so the output program could re-emit that earlier event. This requires a namespaced LSM policy using the Post action and another process hitting the same hook on the same CPU but being rejected by policy filtering.

The second commit is the regression test. It pins matching and non-matching processes to one CPU, opens the target file from the matching namespace, then opens /dev/null five times from the non-matching namespace. Before the fix those rejected invocations would re-emit the stale event, producing six events instead of one.

I also rebased the series onto the current main branch and resolved the conflict in policyfilter_test.go. I would appreciate a review on #5460.

PR #5390, opened 2026-08-09
lsm: support matchData selectors

### LINKED ISSUE #5373: lsmhooks do not support matchData despite being documented in the TracingPolicy API reference

The Tetragon TracingPolicy API reference documents `lsmhooks[index].selectors[index].matchData`, but `matchData` is not actually supported for LSM hooks. Any TracingPolicy that uses `matchData` inside an `lsmhooks` selector fails to load with `matchData index out of bonds (0,0)`. The docs are confusing because the field is documented but unimplemented for this hook type.

### Environment

- Tetragon version: v1.7.0 (binary release `tetragon-v1.7.0-amd64`)
- Kernel: 6.12.38+deb13-amd64 (Debian 13)
- Deployment: standalone binary as a systemd service, policies loaded from `/etc/tetragon/tetragon.tp.d/`

### Expected behavior

Since the API reference lists `lsmhooks[index].selectors[index].matchData`, it should be possible to use `matchData` in LSM hook selectors, e.g. to filter on data resolved from the current task (like `cred.euid.val`) when enforcing with the `Override` action.

### Actual behavior

The policy fails to load. Tetragon exits with:

```
level=error msg="Failed to execute tetragon" error="policy handler 'tracing' failed loading policy '<name>': parseMatchArgs  error: matchData index out of bonds (0,0)"
```

### Minimal reproducer

```yaml
kind: TracingPolicy
metadata:
  name: "test-matchdata-lsm"
spec:
  lsmhooks:
  - hook: "path_rename"
    args:
    - index: 0
      type: "path"
      label: "old_dir"
    selectors:
    - matchArgs:
      - args: [0]
        operator: "Prefix"
        values:
        - "/home"
      matchData:
      - index: 0
        operator: "GT"
        values:
        - "1000"
```

### Request

Add a `data` field to `LsmHookSpec` (mirroring `KProbeSpec.Data`)  which is especially useful for policy enforcement (`Override` action) filtering on resolved task fields such as `cred.euid.val`.

If this is intentionally unsupported for LSM hooks, the API reference should be updated to not list `lsmhooks[index].selectors[index].matchData`, and ideally the schema validation should reject `matchData` in LSM selectors with a clear error instead of the cryptic `matchData index out of bonds (0,0)`.

_No response_

### Code of Conduct

iodn:
Fixes #5373

### Description
LSM hook selectors exposed matchData but LsmHookSpec had no corresponding data field and the generic LSM sensor always passed an empty data list to the selector parser. Means policies using matchData failed to load with an out of bounds index error.

I added data configuration support to LSM hooks using the same argument handling as the other generic sensors. Current task fields can now be resolved and filtered through matchData and also included in the ProcessLsm event data.

Regression coverage verifies policy validation, selector initialization and LSM event output.

### Changelog

```release-note
tracing: Support matchData selectors for LSM hooks
```


kkourt (changes requested):
Can you please take a look at https://tetragon.io/docs/contribution-guide/submitting-a-pull-request/, and update your PR accordingly?

Specifically, please split your PR into multiple commits to ease reviewing.

iodn:
The head fork for this PR was deleted, so GitHub cannot reopen it. I recreated the fork and opened #5461 as its replacement, preserving the original commit unchanged. Please continue review there.

PR #5399, opened 2026-08-10
metricsconfig: return bind errors and graceful shutdown for metrics

### LINKED ISSUE #5368: Metrics server silently fails to start when the configured port is already in use

What happened:

If Tetragon is started with `--metrics-server` set to an address/port that's already occupied by another process, it logs `"Starting metrics server"` but never reports the bind error and never actually serves metrics. The main process continues on as if everything is fine.

This makes the failure hard to diagnose.

How to reproduce:
1. Start another listener on port 2112:
`python3 -m http.server 2112 --bind 0.0.0.0`
2. Start Tetragon with metrics enabled on the same `tcp/2112`:
`sudo tetragon --metrics-server=:2112`
3. Observe:
- Tetragon logs: `level=info msg="Starting metrics server" addr=:2112`
- Tetragon continues and reaches `"Listening for events..."`
- `ss -tlnp | grep 2112` shows only the Python process listening
- `curl http://127.0.0.1:2112/metrics` returns the Python HTTP server's response, not Prometheus metrics
- No error or warning is logged by Tetragon

Code reference:
`pkg/metricsconfig/root.go`:
```
func EnableMetrics(address string) {
    reg := GetRegistry()
    logger.GetLogger().Info("Starting metrics server", "addr", address)
    http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
    http.ListenAndServe(address, nil)
}
```

http.ListenAndServe runs in a goroutine (called via go `metricsconfig.EnableMetrics()` in `cmd/tetragon/main.go`) with its return value ignored entirely.

Environment:
- Tetragon version: v1.7.0
- Platform: bare metal / systemd (also affects any non-Kubernetes deployment that enables --metrics-server)
- Kernel: 6.17.13

Impact:
In production deployments managed by configuration management tools (e.g. Chef) that may restart Tetragon rapidly, a port conflict can leave the service running but without working metrics, leading to silent observability gaps.

Proposed fix:
I would like to propose the following, roughly in order of increasing scope, and would appreciate maintainer input on how far to take it:
1. Log the bind error. Capture `http.ListenAndServe`'s return value and log it instead of discarding it. This alone makes the failure observable and requires no API changes.
2. Accept a `context.Context`. Thread `ctx` into `EnableMetrics` and switch from the package-level `http.ListenAndServe` to a dedicated `*http.Server`, so the metrics server can be gracefully stopped via `Shutdown(ctx)` on `ctx.Done()` (currently there's no shutdown path for this server at all).
3. Optional: bind synchronously. If we want startup to fail loudly on a bad/occupied address (rather than only logging after the fact), move `net.Listen` out of the goroutine and into the synchronous startup path, returning the bind error to the caller so main() can abort startup.

Personally I would think 1 and 2 above is sufficent, 3 seems extreme for a secondary metrics system.

### Tetragon Version

v1.7.0

### Kernel Version

6.17.13

### Kubernetes Version

n/a we are not running in kubernetes, however this probably also affects kube deployments

### Bugtool

_No response_

### Relevant log output

```shell

```

### Anything else?

Let me know if I'm understanding the problem correctly, happy to put up a PR!

stefanamaerz:
Fixes #5368

### Description

`metricsconfig.EnableMetrics()` discards `http.ListenAndServe()` errors. As noted in #5368, if metrics port was already occupied, Tetragon logged `"Starting metrics server"` and continued without working metrics. Its worth noting that other error types fail in similar ways, such as not having permissions to bind to a priv port, using an invalid tcp port, or invalid IP address, ect.

This PR proposes a few changes:
- `EnableMetrics` binds synchronously via `net.Listen` and returns bind errors. This allows `main.go` to handle the error in whatever way it sees fit.
- It uses a dedicated `*http.Server` with a private `http.ServeMux` so we can shut down gracefully and avoid accidental route pollution from the global mux.
- It returns a stop function that gracefully shuts the server down (`http.Server.Shutdown`) and blocks until the serving goroutine returns (`sync.WaitGroup.Wait`). Callers defer stop(), which covers both the signal and error paths in main without touching cleanupWg, sidestepping the potential deadlock described at cmd/tetragon/main.go:536.

The daemon caller logs the bind error and continues running. This was suggested in #5368, we could easily swap to terminate on startup easily from `main.go` if desired, however the easiest/non breaking change would be to log and continue running so the behavior is observable.

### Manual verification:

Port already in use:
```
python3 -m http.server 2112 --bind 0.0.0.0&
sudo ./tetragon --metrics-server=':2112' --bpf-dir=$(pwd)/bpf/objs --bpf-lib=$(pwd)/bpf/objs
snip...
level=error msg="Failed to start metrics server" addr=:2112 error="listen tcp :2112: bind: address already in use"
```

nonsensical tcp port config:
```
~/tetragon$ sudo ./tetragon --metrics-server=':::::2112' --bpf-dir=$(pwd)/bpf/objs --bpf-lib=$(pwd)/bpf/objs
...snip
level=error msg="Failed to start metrics server" addr=:::::2112 error="listen tcp: address :::::2112: too many colons in address"
```

Non-local IP:
```
sudo ./tetragon --metrics-server=203.0.113.99:2112 --bpf-dir=$(pwd)/bpf/objs --bpf-lib=$(pwd)/bpf/objs
...snip
level=error msg="Failed to start metrics server" addr=203.0.113.99:2112 error="listen tcp 203.0.113.99:2112: bind: cannot assign requested address"
```

Graceful shutdown on SIGTERM:
```
sudo ./tetragon --metrics-server=':2112' --bpf-dir=$(pwd)/bpf/objs --bpf-lib=$(pwd)/bpf/objs
...snip...
level=info msg="Starting metrics server" addr=:::2112
...snip...
level=info msg="Received signal terminated, shutting down..."
level=info msg="Listening for events completed." error="context canceled"
level=info msg="Unloading sensor base"
level=info msg="Sensor unloaded" sensor=base maps-error=[]
level=info msg="Metrics server stopped" addr=:::2112
level=info msg="BPF events statistics: 16 received, 0% events loss"
level=info msg="Observer events statistics" received=16 lost=0 errors=0 filterPass=0 filterDrop=0
```

### Changelog
Surface metrics server bind errors and add graceful shutdown via a caller-deferred stop function.

### DIFF
--- a/pkg/metricsconfig/root.go
+++ b/pkg/metricsconfig/root.go
@@ -4,6 +4,9 @@
 package metricsconfig
 
 import (
+	"context"
+	"errors"
+	"net"
 	"net/http"
 	"sync"
 
@@ -11,6 +14,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
 	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/logger/logfields"
 )
 
 var (
@@ -25,10 +29,61 @@ func GetRegistry() *prometheus.Registry {
 	return registry
 }
 
-func EnableMetrics(address string) {
+// newMetricsServer builds the metrics server and binds its listener. Binding is
+// separate from serving so that callers observe bind errors synchronously, and
+// so tests can learn the address that was actually bound (e.g. when asking for
+// port 0).
+func newMetricsServer(address string) (*http.Server, net.Listener, error) {
 	reg := GetRegistry()
 
-	logger.GetLogger().Info("Starting metrics server", "addr", address)
-	http.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
-	http.ListenAndServe(address, nil)
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+
+	listener, err := net.Listen("tcp", address)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return &http.Server{Handler: mux}, listener, nil
+}
+
+// serve starts serving on the already-bound listener in a background
+// goroutine and returns a stop function. stop gracefully shuts the server down
+// and blocks until the serving goroutine has returned, so callers can rely on
+// the server being fully stopped once stop returns. stop is safe to call
+// multiple times; typically it is deferred right after a successful
+// EnableMetrics.
+func serve(server *http.Server, listener net.Listener) (stop func()) {
+	var wg sync.WaitGroup
+	wg.Go(func() {
+		if err := server.Serve(listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.GetLogger().Error("Metrics server exited unexpectedly", logfields.Error, err)
+		}
+	})
+
+	var stopOnce sync.Once
+	return func() {
+		stopOnce.Do(func() {
+			if err := server.Shutdown(context.Background()); err != nil {
+				logger.GetLogger().Error("Failed to shutdown metrics server", logfields.Error, err)
+			}
+			wg.Wait()
+			logger.GetLogger().Info("Metrics server stopped", "addr", listener.Addr())
+		})
+	}
+}
+
+// EnableMetrics starts a Prometheus metrics HTTP server on the given address.
+// It binds synchronously and returns any bind error to the caller. On success
+// it returns a stop function that gracefully shuts the server down and waits
+// for it to finish; callers should defer stop() so both the signal and the
+// error paths stop the server.
+func EnableMetrics(address string) (stop func(), err error) {
+	server, listener, err := newMetricsServer(address)
+	if err != nil {
+		return nil, err
+	}
+
+	logger.GetLogger().Info("Starting metrics server", "addr", listener.Addr())
+	return serve(server, listener), nil
 }
--- a/pkg/metricsconfig/root_test.go
+++ b/pkg/metricsconfig/root_test.go
@@ -1,0 +1,95 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+package metricsconfig
+
+import (
+	"io"
+	"net"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+)
+
+// newTestClient returns a client that never reuses connections, so that a
+// request made after the server has shut down actually attempts to dial rather
+// than picking up a pooled connection.
+func newTestClient() *http.Client {
+	return &http.Client{
+		Timeout:   10 * time.Second,
+		Transport: &http.Transport{DisableKeepAlives: true},
+	}
+}
+
+// TestEnableMetricsBindError checks that a failure to bind is reported to the
+// caller instead of being silently swallowed, which was the bug behind the
+// metrics server appearing to start while never serving anything.
+func TestEnableMetricsBindError(t *testing.T) {
+	t.Run("address already in use", func(t *testing.T) {
+		occupied, err := net.Listen("tcp", "127.0.0.1:0")
+		require.NoError(t, err)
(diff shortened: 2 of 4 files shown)


stefanamaerz:
I think this is ready to review: I'm not able to add labels, so I cannot fix that failed CI check.

pkg/metricsconfig/root.go
```diff
@@ -25,10 +29,58 @@ func GetRegistry() *prometheus.Registry {
...
+func serveMetrics(ctx context.Context, server *http.Server, listener net.Listener) {
+	go func() {
+		if err := server.Serve(listener); err != nil && !errors.Is(err, http.ErrServerClosed) {
+			logger.GetLogger().Error("Metrics server exited unexpectedly", logfields.Error, err)
+		}
+	}()
+
+	go func() {
+		<-ctx.Done()
+		if err := server.Shutdown(context.Background()); err != nil {
```
olsajiri: should we wait for this go routine somewhere? could tetragon finish before we run this?

pkg/metricsconfig/root_test.go
```diff
@@ -0,0 +1,100 @@
+// SPDX-License-Identifier: Apache-2.0
```
olsajiri, line 1: please split tests to separate commit

pkg/metricsconfig/root.go
stefanamaerz: Good catch! I believe you're right that nothing waits for the goroutines to return. Signal → cancel() → Shutdown() → Serve() returning unwinds them, but it races with process exit rather than being a guarantee. The error path looks worse: defer cancel() in tetragonExecute() doesn't run until tetragonExecuteCtx has returned, right before logger.Fatal calls os.Exit, so shutdown may not even start there. Let me know if I'm misreading that.

I can think of two ways to address it:

1. Have EnableMetrics return a func() that does Shutdown + wg.Wait(), and defer stop() in main. This covers both signal and error paths and avoids touching cleanupWg, sidestepping the deadlock issue at main.go:536.
2. Drop the context argument, return *http.Server, and let main defer srv.Shutdown(...). Simpler API, but EnableMetrics still starts the Serve goroutine internally, so the caller doesn't wait for it directly.

I'd lean toward option 1 for the explicit wait. Happy to go another way if you prefer.

mtardy: yeah let's do:

> Have EnableMetrics return a func() that does Shutdown + wg.Wait(), and defer stop() in main. This covers both signal and error paths and avoids touching cleanupWg, sidestepping the deadlock issue at main.go:536.

mtardy (commented):
yeah thanks, please split the commits, update git logs and make sure the lifecycle is correct and we're good :)

pkg/metricsconfig/root.go
stefanamaerz: Done. Implemented option 1 in the latest push. `EnableMetrics` now returns a `func()` that runs `server.Shutdown` and then blocks on `wg.Wait()` for the serving goroutine to exit. The ctx parameter is gone entirely; main just does `defer stopMetrics()` on success, which covers both the signal path (via cancel() → normal return → deferred stop) and any error-return path, without touching `cleanupWg`.

The test now exercises this directly: stop() is called, and because it blocks until Serve returns, the follow-up "server no longer answers" assertion is deterministic. No polling needed.

pkg/metricsconfig/root_test.go
stefanamaerz: Done! Tests are in their own commit now: 9d4d8c3 `metricsconfig: add tests for bind errors and graceful shutdown.` The code changes are in the other commit fdb6b67.

stefanamaerz:
> yeah thanks, please split the commits, update git logs and make sure the lifecycle is correct and we're good :)

Implemented in fdb6b67! EnableMetrics now returns func(), main defers it, and the stop function does Shutdown + wg.Wait behind a sync.Once. Thanks for the review!

mtardy (approved):
thanks lgtm, indeed it kinda simplifies everything by bundling this into the natural defer of the function (and not needing to pass the ctx), I think it looks good

mtardy:
let's see what Jiri thinks

olsajiri (approved):
lgtm thanks

PR #5402, opened 2026-08-10
tetragon: Remove load tests

olsajiri:
Over time, it has proven to be useless and a nuisance.


kkourt (approved):
> Over time, it has proven to be useless and a nuisance.

I agree.

I think what ends up happening is that every time we make a change we just change the ids to match whatever the new state of things is.

olsajiri:
> > Over time, it has proven to be useless and a nuisance.
> 
> I agree.
> 
> I think what ends up happening is that every time we make a change we just change the ids to match whatever the new state of things is.

yep, also I don't think it ever caught any bug

PR #5410, opened 2026-08-11
protoc-gen: auto-generate more unset code

andrewstrohman:
The eventchecker codegen already emits per-checker UnsetProcess() and
UnsetParent() helpers. Also generate the UnsetProcessChecks() and
UnsetParentChecks() helpers, so that hand-written code doesn't have to
be updated whenever a new event type is added.


mtardy (approved):
Generated code looks perfectly equivalent to what was manually maintained! So lgtm! Changes look good!

Maybe it would be worth waiting for a review from William but no hard req I think since this looks pretty correct.

PR #5411, opened 2026-08-12
tetragon: fixing debug maps output

bar234:
### Description
Fixes some output issues we were seeing with `tetra debug maps`. It uses full map name where available, fixing how multiple maps are aggregated together (previous used only truncated name, now it keys on 5 parts of the map). It also disables preallocating of the `sleepable_offload` map and adds a flag to configure its max size

### Changelog
```release-note
bugtool: Fixing bug in debug map output aggregation
tetragon: Make sleepable_offload not preload and max size configurable
```

### DIFF
--- a/bpf/process/uprobe_offload.h
+++ b/bpf/process/uprobe_offload.h
@@ -44,6 +44,7 @@ struct {
 	__uint(max_entries, 1); // will be resized by agent when needed
 	__type(key, __u64);
 	__type(value, __u32);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
 } sleepable_offload SEC(".maps");
 
 FUNC_INLINE void do_uprobe_override(void *ctx, __u32 idx)
--- a/pkg/bugtool/maps.go
+++ b/pkg/bugtool/maps.go
@@ -327,38 +327,54 @@ func RunMapsChecks(path string) (*MapsChecksOutput, error) {
 		})
 	}
 
-	// aggregates maps total memory use
-	aggregatedMapsSet := map[string]struct {
-		bpf.ExtendedMapInfo
-		count uint64
+	// aggregates maps total memory use. key on these five fields
+	type aggregatedKey struct {
+		Name       string
+		Type       ebpf.MapType
+		KeySize    uint32
+		ValueSize  uint32
+		MaxEntries uint32
+	}
+
+	aggregatedMapsSet := map[aggregatedKey]struct {
+		Count        uint64
+		TotalMemlock uint64
 	}{}
 	var total uint64
 	for _, m := range union {
 		total += m.Memlock
-		if e, exist := aggregatedMapsSet[m.Name]; exist {
-			e.Memlock += m.Memlock
-			e.count++
-			aggregatedMapsSet[m.Name] = e
-		} else {
-			aggregatedMapsSet[m.Name] = struct {
-				bpf.ExtendedMapInfo
-				count uint64
-			}{m, 1}
+		key := aggregatedKey{
+			Name:       m.Name,
+			Type:       m.Type,
+			KeySize:    m.KeySize,
+			ValueSize:  m.ValueSize,
+			MaxEntries: m.MaxEntries,
 		}
+		if e, exists := aggregatedMapsSet[key]; exists {
+			e.Count++
+			e.TotalMemlock += m.Memlock
+			aggregatedMapsSet[key] = e
+		} else {
+			aggregatedMapsSet[key] = struct {
+				Count        uint64
+				TotalMemlock uint64
+			}{1, m.Memlock}
+		}
+
 	}
 
 	out.AggregatedMaps = make([]AggregatedMap, 0, len(aggregatedMapsSet))
-	for m := range maps.Values(aggregatedMapsSet) {
+	for k, m := range aggregatedMapsSet {
 		out.AggregatedMaps = append(out.AggregatedMaps, AggregatedMap{
-			Name:              m.Name,
-			Type:              m.Type.String(),
-			KeySize:           m.KeySize,
-			ValueSize:         m.ValueSize,
-			MaxEntries:        m.MaxEntries,
-			Count:             m.count,
-			MemlockBytes:      m.Memlock,
-			TotalMemlockBytes: m.Memlock * m.count,
-			PercentOfTotal:    float64(m.Memlock*m.count) / float64(total) * 100,
+			Name:              k.Name,
+			Type:              k.Type.String(),
+			KeySize:           k.KeySize,
+			ValueSize:         k.ValueSize,
+			MaxEntries:        k.MaxEntries,
+			Count:             m.Count,
+			MemlockBytes:      m.TotalMemlock / m.Count,
+			TotalMemlockBytes: m.TotalMemlock,
+			PercentOfTotal:    float64(m.TotalMemlock) / float64(total) * 100,
 		})
 	}
 
--- a/pkg/bugtool/maps.go
+++ b/pkg/bugtool/maps.go
@@ -9,6 +9,7 @@ import (
 	"iter"
 	"maps"
 	"os"
+	"path/filepath"
 	"sort"
 	"syscall"
 
@@ -103,6 +104,10 @@ func FindPinnedMaps(path string) ([]bpf.ExtendedMapInfo, error) {
 		if err != nil {
 			return nil, fmt.Errorf("failed to retrieve extended info from map %v: %w", m, err)
 		}
+
+		// The map name returned from ExtendedInfoFromMap is truncated to 15
+		// characters. It's more helpful to the user to have the full name
+		xInfo.Name = filepath.Base(pin.Path)
 		infos = append(infos, xInfo)
 	}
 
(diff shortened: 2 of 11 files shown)


bpf/process/uprobe_offload.h
```diff
@@ -44,6 +44,7 @@ struct {
 	__uint(max_entries, 1); // will be resized by agent when needed
 	__type(key, __u64);
 	__type(value, __u32);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
```
olsajiri, line 47: we did the same for preload in 082e52d9aa1b00411339f0a68ac5a8e3264871c7 and we added a way to configure it as shared map or program map 70b87d58529a93c7286bec51fab7d2d10d680c2a ... perhaps we could do the same for offload map?

bar234: No problem! I'll implement it in the same fashion as `sleepable_preload`

andrewstrohman (approved):
Thanks, LGTM.

For maps that are not pinned, their names still show up truncated in the final (aggregated) table in the output. I'm not sure if that inconsistency is less confusing than improperly aggregating differently-named maps because they share a 15 character prefix. Did you consider trying to keep the conenction between the maps and their full names, so that the unpinned map names are displayed in full? The value of doing that may not be worth the complexity to do it.

I do agree that it's helpful to display the full map name, and I also think that the truncation of the unpinned map names is not too problematic, because if they are unpinned, they are probably less interesting. So, I think overall, this is an improvement.

pkg/bugtool/maps.go
```diff
@@ -292,7 +296,7 @@ func RunMapsChecks(path string) (*MapsChecksOutput, error) {
 	}
 
 	diff := diff(pinnedMapsSet, pinnedProgsMapsSet)
-	union := union(pinnedMapsSet, pinnedProgsMapsSet)
+	union := union(pinnedProgsMapsSet, pinnedMapsSet)
```
mtardy, line 303: why the switch?

mtardy (commented):
Thanks, changes make sense to me.

> This change does two things to fix/improve this: where available it uses the full map name (only for pinned maps) and it keys on a combination of name, type, key size, value size and max entries

I think this shows that this patch could have been cut in two commits? Maybe one with just 

```go
// Update the map name to the full name in the path
xInfo.Name = filepath.Base(pin.Path)
```

and btw I think it's more important to add a comment about why we do that than what (since the code already tell us what)

> because if they are unpinned, they are probably less interesting

Indeed, the idea was for this tooling to be able to catch those missing ones, but they should be pinned eventually.

pkg/bugtool/maps.go
bar234: `union` starts with the first set and overwrites keys that match with values from the second map. In the previous parameter order, maps from `pinnedProgsMapsSet` would overwrite map values from `pinnedMapsSet` when creating the union, returning values that have the truncated names instead of the expanded names.

I'll admit this was a bit of a simple change and this behavior isn't obvious to the caller. I'm open to a more holistic adjustment to this function. Maybe the values from the first map take priority, or a third parameter to indicate which map values take priority?

mtardy: I think a comment to explain why the order matters here would be enough as you explained :)

bar234:
> For maps that are not pinned, their names still show up truncated in the final (aggregated) table in the output. I'm not sure if that inconsistency is less confusing than improperly aggregating differently-named maps because they share a 15 character prefix. Did you consider trying to keep the conenction between the maps and their full names, so that the unpinned map names are displayed in full? The value of doing that may not be worth the complexity to do it.

@andrewstrohman Thanks for the feedback! I agree, the goal would be to have full names for all the maps. Honestly getting the full names of the unpinned maps looked a level of difficulty above where the rest of these changes were. If this is something we want I'd be happy to spend some time on it, but I feel like compared to the complexity of the rest of this work and what we get from it, it wasn't worth doing at this moment. This is also an area I need to do some digging to understand better. So all that to say I'd love to have the full names for all of the maps but it was going to be a lot more work and I didn't want it blocking the rest of these changes :-)

bar234:
> > This change does two things to fix/improve this: where available it uses the full map name (only for pinned maps) and it keys on a combination of name, type, key size, value size and max entries
> 
> I think this shows that this patch could have been cut in two commits? Maybe one with just
> 
> ```go
> // Update the map name to the full name in the path
> xInfo.Name = filepath.Base(pin.Path)
> ```
> 
> and btw I think it's more important to add a comment about why we do that than what (since the code already tell us what)
> 
> > because if they are unpinned, they are probably less interesting
> 
> Indeed, the idea was for this tooling to be able to catch those missing ones, but they should be pinned eventually.

@mtardy Gotcha, you're right that is a clean division of work that should be in its own commit. I can make that change. As well as update the comment :-)

bar234:
Sorry for the early push, still wrapping up commit for the `--sleepable-offload-size` argument and making it a shared map

bar234:
Pushed tests borrowing heavily from the `sleepable_preload` tests. If it's better to do this as policytests I can do that, it'll just take a bit more time

pkg/sensors/tracing/genericusdt.go
```diff
@@ -252,8 +252,8 @@ func createMultiUsdtSensor(
 	maps = append(maps, createSelectorMaps(load, nil)...)
 
 	if has.sleepableOffload {
-		sleepableOffloadMap := program.MapBuilderProgram("write_offload", load)
-		sleepableOffloadMap.SetMaxEntries(sleepableOffloadMaxEntries)
+		sleepableOffloadMap := program.MapShared("write_offload", load)
```
andrewstrohman, line 255: Why is write_offload treated differently than sleepable_offload?

I see they both use the same option.Config.SleepableOffloadSize. But "write_offload" is unconditionally a shared map, and users cannot make it a per-program by specifying a spec opt.

pkg/defaults/defaults_windows.go
```diff
@@ -62,6 +62,9 @@ const (
 	// DefaultSleepablePreloadSize is the default maximum number of entries in the sleepable preload map.
 	DefaultSleepablePreloadSize = 32768
 
+	// DefaultSleepableOffloadSize is the default maximum number of entries in the sleepable offload map.
+	DefaultSleepableOffloadSize = 32768
```
andrewstrohman, line 66: should this change be in the previous commit?

pkg/sensors/tracing/uprobe_test.go
```diff
@@ -1343,10 +1365,11 @@ spec:` + opts + `
 		sensors.UnloadSensors(sensi)
 	}
 
-	findPreloadMap := func(sens []*sensors.Sensor) *program.Map {
+	findMap := func(sens []*sensors.Sensor) *program.Map {
 		for _, s := range sens {
 			for _, m := range s.Maps {
-				if m.Name == "sleepable_preload" {
+				logger.GetLogger().Info("Sensor: " + m.Name)
```
andrewstrohman: Did you intend to keep this logging?

mtardy:
btw @bar234 if you want another review when you think it's ready, please click this:
image

bar234:
@andrewstrohman Sorry about the condition of this last change! This is what I get for trying to push changes with a fever! I'll clean this up and flag for a re-review when it's ready

pkg/defaults/defaults_windows.go
bar234: Yup, it really should

pkg/sensors/tracing/uprobe_test.go
bar234: Nope, that was accidentally left in

andrewstrohman:
> @andrewstrohman Sorry about the condition of this last change!

No worries!

> This is what I get for trying to push changes with a fever!

I hope you feel better soon.

pkg/sensors/tracing/genericusdt.go
bar234: So I'm trying to understand the `write_offload` case and figure out how it should be changed as part of this PR. It looks functionally equivalent to `sleepable_offload` but it's for usdt. I was following the changes made to `sleepable_preload`, where it is always a shared map with a configurable size in `option.Config.SleepablePreloadSize`, even though this isn't the case with `sleepable_preload` in uprobes.

Are you suggesting the behavior should change here so the user can make it per-program with a spec out? Sorry if this is a naive question, just want to understand your perspective of this issue and figure out the right change to implement

andrewstrohman: When I wrote that comment, I missed the fact that `sleepable_preload` is different for uprobe vs usdt. So, please disregard. I understand that this is consistent now.

andrewstrohman (approved):

pkg/sensors/tracing/options.go
```diff
@@ -108,6 +109,19 @@ var opts = map[string]opt{
...
+			if err != nil {
+				return err
+			}
+			if size <= 0 {
+				return fmt.Errorf("sleepable-offload-size must be positive, got %d", size)
+			}
+			options.SleepableOffloadSize = size
+			return nil
+		},
+	},
```
mtardy, line 124: so this option is both in tracing policy and as a global flag?

mtardy (approved):
Thanks!! 

I would have maybe cut this PR in two, the part on the debug maps and the part on the sleepable map but lgtm! I can especially judge the first part.

pkg/sensors/tracing/options.go
bar234: Correct.

```
      --sleepable-offload-size int                Set the maximum number of entries in the sleepable offload map (default 32768)
```
and
```
    options:
    - name: "sleepable-offload-size"
      value: "1024"
```

bpf/process/uprobe_offload.h
bar234: @olsajiri these changes should be made. When you have a moment, a quick review would be appreciated! Thank you!

bar234:
@ExceptionalHandler You got tagged because this change added a few constants to [defaults_windows.go](https://github.com/cilium/tetragon/pull/5411/changes/d229dec719a8e358c85792793178e4e9a6b104e0#diff-d9f441a1d4c37684e5c80b73a724f3a72d0d415cc70ce0f74e6919e49fbc5307). There shouldn't be any other Windows impacting changes here

PR #5422, opened 2026-08-12
Simplify making grpc clients to tetragon

mstemm:
### Description
Make it easier/remove redundant code from the multiple grpc service clients of tetragon. Currently, each client must create/manage their own grpc connection, contexts, timeouts, and cancellation functions.

Modify tetra common to add a NewConnWithContext(ctx, addr, timeout, service) function and supporting struct that handles creating the grpc connection and setting up the various contexts. It returns a ConnWithContext that can be embedded in a ClientWithContext struct via:

```
type ClientWithContext struct {
	ConnWithContext
	Client tetragon.FineGuidanceSensorsClient
}
```

replacing tetragon.FindGuidanceSensorsClient with whatever grpc interface the caller needs.

### Changelog

```
Make it easier to create grpc clients to tetragon via a new function NewConnWithContext().
```

### DIFF
--- a/cmd/tetra/common/client.go
+++ b/cmd/tetra/common/client.go
@@ -36,7 +36,7 @@ import (
 // Also note that the final backoff duration is completely random and chosen
 // between 0 and the final duration that was computed via to the params:
 // https://github.com/grpc/grpc-go/blob/v1.65.0/stream.go#L702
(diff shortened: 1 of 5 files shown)


mtardy (approved):
This is great thanks!

PR #5438, opened 2026-08-14
celbpf: introduce bitwise operators

kkourt:
See commits.

```release-note
celbpf: introduce bitwise operators for CEL expressions (and, or, xor, not)
```


mtardy (approved):
okay looks good to me, first CEL PR I actually review :)

PR #5439, opened 2026-08-14
celbpf: add support for data in CEL expressions

kkourt:
See commits.

```release-note
celbpf: introduce `MatchCEL` selector that allows mixing arguments and data in CEL expressions
```


pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -184,6 +184,11 @@ type KProbeSelector struct {
 	// A list of macros names, defined in spec.selectorsMacros.
 	// Filters specified in macros will be appended to corresponding filters of the selector.
 	Macros []string `json:"macros,omitempty"`
+	// +kubebuilder:validation:Optional
+	// Match CEL expression. The CEL expression may include:
+	//  argX (e.g., arg0) where X is the index of the argument in Args.
+	//  dataX where X is the index of the data in Data.
```
mtardy: could you be disambiguate which index here for the user without context: the `index` field in the args/data struct or the index of the array containing the args or data.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -264,8 +275,10 @@ type ArgSelector struct {
 	// +kubebuilder:validation:items:Minimum=0
 	// Position of the operator arguments (in spec file) to apply the filter to.
 	Args []uint32 `json:"args,omitempty"`
-	// +kubebuilder:validation:Enum=Equal;NotEqual;Prefix;NotPrefix;Postfix;NotPostfix;GreaterThan;LessThan;GT;LT;Mask;SPort;NotSPort;SPortPriv;NotSportPriv;DPort;NotDPort;DPortPriv;NotDPortPriv;SAddr;NotSAddr;DAddr;NotDAddr;Protocol;Family;State;InMap;NotInMap;CapabilitiesGained;InRange;NotInRange;SubString;SubStringIgnCase;CelExpr;FileType;NotFileType
 	// Filter operation.
+	//
+	// Note: The CelExpr operator is deprecated and will be removed in Tetragon OSS v1.9.0. Use the MatchCEL selector instead.
```
mtardy, line 280: maybe we could add some user facing warning on the parse code so that it's more visible? unsure

Also kubebuilder has
```
// +kubebuilder:deprecatedversion:warning=&#34;v1alpha1 is deprecated; use v1 instead&#34;
```
maybe this could help here? not sure also

pkg/celbpf/celbpf.go
```diff
@@ -39,7 +39,7 @@ func Supported() bool {
 	return bpf.DetectMixBpfAndTailCalls()
 }
 
-func Compile(celExpr string, sig []v1alpha1.KProbeArg, labelPrefix string) (asm.Instructions, []uint16, error) {
+func Compile(celExpr string, sig, data []v1alpha1.KProbeArg, labelPrefix string) (asm.Instructions, []uint16, error) {
```
mtardy, line 42: sorry I haven't follow this codebase too much but why the `sig` name for the args?

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -184,6 +184,11 @@ type KProbeSelector struct {
 	// A list of macros names, defined in spec.selectorsMacros.
 	// Filters specified in macros will be appended to corresponding filters of the selector.
 	Macros []string `json:"macros,omitempty"`
+	// +kubebuilder:validation:Optional
+	// Match CEL expression. The CEL expression may include:
+	//  argX (e.g., arg0) where X is the index of the argument in Args.
+	//  dataX where X is the index of the data in Data.
+	MatchCEL *CELExprSelector `json:"matchCEL,omitempty"`
```
mtardy, line 191: so we have only one cel expression per selector?

pkg/celbpf/compiler.go
```diff
@@ -212,12 +214,15 @@ func (c *compiler) compileCall(expr cgAst.Expr) error {
 }
 
 func (c *compiler) compileArg(argIdx uint16) error {
-	if int(argIdx) >= len(c.args) {
-		return fmt.Errorf("invalid argument (arg%d): undefined", argIdx)
+	var arg v1alpha1.KProbeArg
+	if int(argIdx) < len(c.args) {
+		arg = c.args[argIdx]
+	} else if int(argIdx) < len(c.args)+len(c.data) {
+		arg = c.data[int(argIdx)-len(c.args)]
```
mtardy, line 221: sorry but I'm not sure to get why c.data is organized like that

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -845,6 +806,59 @@ matchWorkloads:
...
+      type: "int"
+      label: "fd"
+    selectors:
+    - matchActions:
+      - action: Post
+      MatchCel:
+        expr: "arg1 == int32(-1) && data0 == arg0 + int32(42)"
+```
+
+Arguments, as specified in the `args:` list, are made available in the CEL expression as `argX` where `X` is the zero-based index in the list. Similary for data items, as specified in the `data:` list.
```
mtardy: please wrap lines

mtardy: also maybe again insist on array `index` vs `index` the field

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -845,6 +806,59 @@ matchWorkloads:
       - "kube-system"
 ```
 
+## matchCEL
+
+
```
mtardy: nit: useless whiteline

mtardy (commented):
This is cool, thanks for the clean PR, I have a few comments.

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -222,6 +227,12 @@ type CapabilitiesSelector struct {
 	Values []string `json:"values"`
 }
 
+type CELExprSelector struct {
```
mtardy, line 230: This should be
```suggestion
type CelExprSelector struct {
```

mtardy: Just kidding 🤡

pkg/k8s/apis/cilium.io/v1alpha1/types.go
kkourt: I tried using "Deprecated: " but it ended up creating warnings in our build, which is why I added the "Note:". I could not find of another way. I've also added it in the release notes.

pkg/celbpf/celbpf.go
kkourt: Not sure, I just re-used the same argument name when I did the initial CEL implementation.

kkourt: FWIW, seems to be there in the first commit in cilium/tetragon: https://github.com/cilium/tetragon/blob/931b70f2c9878ba985ba6b589827bea17da6ec33/pkg/selectors/kernel.go#L308

pkg/k8s/apis/cilium.io/v1alpha1/types.go
kkourt: Yes. Why would we need more?

<details>

CEL allows arbitrary expressions (incudung `&&` and `||`). Instead of having two selectors with expressions `expr1` and `expr2`, you can just have one selector with: `expr1 || expr2`.

</details>

pkg/celbpf/compiler.go
kkourt: In bpf side, there little difference between args and data arguments. They are both pushed to a single array.
See: https://github.com/cilium/tetragon/blob/350c2fce18d36b3a3727989988a4b6eca86cd27a/pkg/selectors/kernel.go#L1113-L1116

So argIdx is the index to this array (that contains both args and data). An `argIdx` that is `>len(c.args)` refers to data and it's information can be found in `c.data[int(argIdx)-len(c.args)]`

mtardy: ah okay thanks makes sense!

mtardy (approved):
:shipit:

PR #5441, opened 2026-08-14
Disable automatic Windows GH runner upgrades.

ygvalent:
see: https://github.com/cilium/tetragon/pull/5298


mtardy (approved):
lgtm! we could have been more precise on the version blablabla but not sure it would be useful tbh

ygvalent:
> we could have been more precise on the version blablabla but not sure it would be useful tbh

I actually wrote it like this on purpose, first the docs say that versioned runners will not match, and second, we don't want any auto-updates for Win builds

PR #5442, opened 2026-08-14
Introduce matchCmdArgs selector

mtardy:
See individual commits.

This is essentially quite similar to `matchArgs` but simpler, we already have some foundations for this.

```release-note
Added matchCmdArgs selectors to filter tracing events based on individual command-line arguments of the invoking process.
```


ygvalent:
@mtardy I see the failed win-ci run here again ..
I think it needs a rebase on main to pickup the changed win-ci workflow

pkg/selectors/kernel.go
```diff
@@ -944,6 +944,72 @@ func ParseMatchData(k *KernelSelectorState, arg *v1alpha1.ArgSelector, data []v1
 	return parseMatchArg(k, arg, data, ty)
 }
 
+// Keep in sync with CMD_ARGS_MAX in bpf/process/types/basic.h.
```
tpapagian: `CMD_ARGS_MAX` is defined in a later patch. Possibly add this comment on that patch?

bpf/process/types/basic.h
```diff
@@ -2407,13 +2413,121 @@ match_args(void *ctx, struct bpf_map_def *tailcalls, __u8 *f,
 	return 1;
 }
 
+#ifdef __LARGE_BPF_PROG
+
+#define CMD_ARGS_MAX	    32
```
tpapagian: Question: Although the values 32 seems reasonable to me, have you also tested larger values? Or is there a fundamental reason for that?

tpapagian (approved):
LGTM, thanks! Just 2 minor comments.

pkg/selectors/kernel.go
mtardy: yeah I could move the comment indeed

bpf/process/types/basic.h
mtardy: so initially I did 256 which made not much sense since the max length of the string is also 256, and it exploded complexity on older kernel pre-6.1. So I just tried to think about a reasonable number of arguments in 256 chars and thought 32 was enough but this is pretty arbitrary. I think pushing it typically to 64 doesn't make much of a difference. I was thinking that if we have users of this they can tell us their need and we can fine tune.

bpf/process/types/basic.h
```diff
@@ -2427,6 +2410,165 @@ selector_arg_offset(void *ctx, struct bpf_map_def *tailcalls,
...
+	/* Reusing the string_maps_heap instead of adding another heap map, using
+	 * the second half as filter_char_buf_value() will use the first half.
+	 */
+	string_heap = map_lookup_elem(&string_maps_heap, &zero);
+	if (!string_heap)
+		return 0;
+	string_heap += CMD_ARG_HEAP_OFFSET;
+
+	/* Walk the arguments and find the one corresponding to the filter->index */
+	for (i = 0; i < CMD_ARGS_MAX; i++) {
```
olsajiri: let's save some verifier instructions and add the possibility to do loop via ITER_NUM ?

bpf/process/types/basic.h
```diff
@@ -2427,6 +2410,165 @@ selector_arg_offset(void *ctx, struct bpf_map_def *tailcalls,
...
+		if (i > filter->index || argsoff >= cached_args->len)
+			break;
+
+		remaining = cached_args->len - argsoff;
+		argsoff &= MAXARGLENGTH - 1;
+		read = probe_read_str(string_heap, MAXARGLENGTH, &cached_args->buf[argsoff]);
```
olsajiri: could you read just `remaining` bytes?

bpf/process/bpf_execve_event.h
```diff
@@ -365,22 +396,9 @@ execve_send_event(struct bpf_raw_tracepoint_args *ctx,
...
-		if (event->exe.arg_len > sizeof(curr->bin.args) - 2)
-			len = sizeof(curr->bin.args) - 2;
-		else
-			len = event->exe.arg_len;
-		with_errmetrics(probe_read, curr->bin.args, len, (char *)&event->process + off);
-
-		// there's a null byte between each argv element, so we terminate with
-		// two of them to make it possible to identify the end of the buffer
-		curr->bin.args[len] = 0x00;
-		curr->bin.args[len + 1] = 0x00;
```
olsajiri, line 383: hum, I wonder what was the reason behind this, maybe is it used in ee?

bpf/lib/process.h
```diff
@@ -319,8 +332,6 @@ struct binary {
 	char end[STRING_POSTFIX_MAX_LENGTH];
 	// STRING_POSTFIX_MAX_LENGTH reversed last bytes of the path
 	char end_r[STRING_POSTFIX_MAX_LENGTH];
-	// args for the binary
-	char args[MAXARGLENGTH];
```
olsajiri, line 323: could this move break ee code?

bpf/process/bpf_execve_event.h
```diff
@@ -365,22 +396,9 @@ execve_send_event(struct bpf_raw_tracepoint_args *ctx,
...
+		copy_args_to_cache(&event->args_source, &curr->args);
```
olsajiri: nit the `_cache` suffix is confusing for me, it suggest different traetment then earlier `copy_exe_to_bin`

bpf/process/types/basic.h
```diff
@@ -2427,6 +2410,165 @@ selector_arg_offset(void *ctx, struct bpf_map_def *tailcalls,
...
+	if (filter->index >= CMD_ARGS_MAX)
+		return 0;
+
+	if (!cached_args->len || cached_args->len > MAXARGLENGTH)
+		return 0;
+
+	/* Reusing the string_maps_heap instead of adding another heap map, using
+	 * the second half as filter_char_buf_value() will use the first half.
+	 */
+	string_heap = map_lookup_elem(&string_maps_heap, &zero);
```
olsajiri: this smells like future problems ;-) the data is already in the map.. could we instead loop through `cached_args->buf` using bpf_strlen (not sure when it got introduced) or just scan for 0? instead of using `probe_read_str` ?

olsajiri (commented):
left some comments,
is there a reason this filter is after arguments? it feels like it would fit better in  `generic_process_filter`

mtardy:
> is there a reason this filter is after arguments? it feels like it would fit better in `generic_process_filter`

It felt natural as it's reusing the same functions and code as the args processing with all the strings helpers that already belong here. May I ask why it would fit better there?

bpf/process/types/basic.h
mtardy: initially I wanted to do it like that but it doesn't work well when calling a local function, that's why I copied it into a map for the processing below it doesn't matter.

Do you think a new heap map would be simpler?

bpf/process/types/basic.h
mtardy: this is 6.4, you mean adding a variant for later kernels with supports (since this thing works from LARGE_PROG)

bpf/process/bpf_execve_event.h
mtardy: so you would prefer `copy_args_to_args`? Or should I rename the other one?

bpf/process/types/basic.h
mtardy: I could try

olsajiri:
> > is there a reason this filter is after arguments? it feels like it would fit better in `generic_process_filter`
> 
> It felt natural as it's reusing the same functions and code as the args processing with all the strings helpers that already belong here. May I ask why it would fit better there?

so the argument filter gets argument data and does the filter on it, here the data are already stored in execve_map so we do not need to dig them out, we can just do the filter like all the other filters in `generic_process_filter` .. plus there's also the related `match_binaries` filter

I'm not sure it will fit in there better, just wondered if we checked that

bpf/process/types/basic.h
olsajiri: maybe, or extend heap map, or somehow ensure we don't use the 2nd half of `string_maps_heap` in future?

bpf/process/bpf_execve_event.h
olsajiri: it's a nit, I don't care ;-)

bpf/process/types/basic.h
olsajiri: I meant the loop arrangement like in `prepend_path` `(if  CONFIG(ITER_NUM))` to make sure later kernels do not suffer from this

bpf/process/types/basic.h
mtardy: so this second half is never used iirc, I think it's only there for the verifier to be happy

mtardy: I agree it's sketchy

bpf/process/bpf_execve_event.h
mtardy: I changed it to `copy_args`

bpf/process/types/basic.h
mtardy: so I checked, and if we have the if else construct, the verifier has the same amount of instructions it seems.

I have `163258` instructions with just the for loop and `163762` if doing it this way

```c
	if (CONFIG(ITER_NUM)) {
		bpf_for(i, 0, CMD_ARGS_MAX)
		{
			if (i > filter->index || argsoff >= cached_args->len)
				break;

			arglen = read_argv(cached_args, &argsoff, string_heap);
			if (arglen && filter->index == i) {
				found = true;
				break;
			}
		}
	} else {
		for (i = 0; i < CMD_ARGS_MAX; i++) {
			if (i > filter->index || argsoff >= cached_args->len)
				break;

			arglen = read_argv(cached_args, &argsoff, string_heap);
			if (arglen && filter->index == i) {
				found = true;
				break;
			}
		}
	}
```

Am I doing something wrong?

~~Dead code elimination should simplify this and the bpf loop variant should be smaller nop? Maybe I'll check in other places with the if config check if I see the same behavior or if it's my code that somehow blocks this.~~ we guard this until 6.9

mtardy:
> > > is there a reason this filter is after arguments? it feels like it would fit better in `generic_process_filter`
> > 
> > 
> > It felt natural as it's reusing the same functions and code as the args processing with all the strings helpers that already belong here. May I ask why it would fit better there?
> 
> so the argument filter gets argument data and does the filter on it, here the data are already stored in execve_map so we do not need to dig them out, we can just do the filter like all the other filters in `generic_process_filter` .. plus there's also the related `match_binaries` filter
> 
> I'm not sure it will fit in there better, just wondered if we checked that

Mmh I think it actually makes more sense to put it there, I kept the string heap reuse however as it's a bit tricky to replace it with something that reads `\0`, every time I hit the 1M+ limit. However if the concern is the reuse of some unused space, we can create a new heap map.

mtardy:
>     verify_test.go:176: bpf_generic_rawtp_v511.o ✗: failed to load object on kernel 6.17.13: program generic_rawtp_process_filter: load program: argument list too long: BPF program is too large. Processed 1000001 insn (24529 line(s) omitted)

knowing that v511 fails on a much later kernel is not super useful, let's see if the actual kernels can load these progs or not and I'll adjust them if needed, see https://github.com/cilium/tetragon/pull/5510

bpf/process/pfilter.h
```diff
@@ -414,6 +415,150 @@ struct nc_filter {
...
+	/* Check for malformed filters */
+	if (filters->arglen < sizeof(struct selector_arg_filters))
+		return 0;
+
+	/* An empty section always matches even without a cached entry */
+	if (!filters->argoff[0])
+		return 1;
+
+	/* The matcher exists but cannot be evaluated because of missing cache */
+	if (!cached_args)
```
olsajiri: this will be always != NULL, right?

bpf/lib/process.h
olsajiri, line 323: there's memcpy in event_wake_up_new_task, which won't cover the args now
`memcpy(&curr->bin, &parent->bin, sizeof(curr->bin));
`
I think we need to add memcpy for args now

olsajiri: also I wonder if we actualy read args in procfs reader? now that we use it, we might want to

bpf/process/bpf_execve_event.h
```diff
@@ -365,22 +396,9 @@ execve_send_event(struct bpf_raw_tracepoint_args *ctx,
...
-			len = sizeof(curr->bin.args) - 2;
-		else
-			len = event->exe.arg_len;
-		with_errmetrics(probe_read, curr->bin.args, len, (char *)&event->process + off);
-
-		// there's a null byte between each argv element, so we terminate with
-		// two of them to make it possible to identify the end of the buffer
-		curr->bin.args[len] = 0x00;
-		curr->bin.args[len + 1] = 0x00;
+		copy_args(&event->args_source, &curr->args);
```
olsajiri, line 401: should we do this as well in `event_find_curr_probe` ? like try to fetch args when we fail to get execve_map record, like we do for binary

bpf/process/pfilter.h
```diff
@@ -414,6 +415,150 @@ struct nc_filter {
...
+	seloff += *(__u32 *)((__u64)f + (seloff & INDEX_MASK));
+
+	return seloff & INDEX_MASK;
+}
+
+FUNC_INLINE int read_argv(struct args *cached_args, __u32 *offset, char *string_heap)
+{
+	__u32 off = *offset & (MAXARGLENGTH - 1);
+	__u32 remaining = (cached_args->len - *offset) & (2 * MAXARGLENGTH - 1);
+	int read = probe_read_str(string_heap, remaining, &cached_args->buf[off]);
```
olsajiri: note, I wonder if we do this the other way around like iterate first args and for each arg we check the filters, we'd get less `probe_read_str` calls.. but I guess it's fine

olsajiri (commented):
looks better in the generic filter, thanks, left few comments

bpf/process/pfilter.h
mtardy: This gives me the idea to propose a completely different version where I first parse the 32 args into an argv style PERCPU_ARRAY and then process the filters. It consumes a bit more memory as it needs one more map but we'll have less nested loops.

bpf/lib/process.h
mtardy: Ah I need to take a look, don't know also

bpf/process/pfilter.h
```diff
@@ -414,6 +415,180 @@ struct nc_filter {
...
+	*dst = heap;
+	return read - 1;
+}
+
+struct {
+	__uint(type, BPF_MAP_TYPE_PERCPU_ARRAY);
+	__uint(max_entries, CMD_ARGS_MAX);
+	__type(key, __u32);
+	__type(value, char[MAXARGLENGTH]);
+} argv SEC(".maps");
```
olsajiri: hum, so this will add like 8KB on each cpu, it's also unconditional, so also sensors without that selector will have that

bpf/process/pfilter.h
```diff
@@ -414,6 +415,180 @@ struct nc_filter {
...
+		return 0;
+
+	/* An empty section always matches even without a cached entry */
+	if (!filters->argoff[0])
+		return 1;
+
+	remaining = cached_args->len;
+
+	/* Parse the arguments once into argv PERCPU_ARRAY */
+	for (i = 0; i < CMD_ARGS_MAX; i++) {
```
olsajiri: so we parse all (CMD_ARGS_MAX) args even if we want just the first one, that's not optimal

I wonder we could do lazy arguments parsing based on what argument index we need for the filter and gather just offsets instead of whole strings in argv map? this way we could lower argv size and reduce the cpu usage a bit

not sure about that big buffer limitation in `filter_char_buf_value` but hopefully we could bypass that somehow?

mtardy: > not sure about that big buffer limitation in filter_char_buf_value but hopefully we could bypass that somehow?

It needs a 4096 buffer size at least because of some of the matching.

> so we parse all (CMD_ARGS_MAX) args even if we want just the first one, that's not optimal
>
> I wonder we could do lazy arguments parsing based on what argument index we need for the filter and gather just offsets instead of whole strings in argv map? this way we could lower argv size and reduce the cpu usage a bit

mmh we can do that but we will need to loop the filters two times, one to get until which index we need and one to do the matching given the operation nop?

mtardy: I think it would optimize the runtime in most case but the worse case (verified path) will be bigger then?

bpf/process/bpf_execve_event.h
mtardy: good point

mtardy:
Things that are missing:
- read args in `event_find_curr_probe`
- cover the args in `event_wake_up_new_task`
- check for what we do currently in the procfs reader

I think that those can be addressed as a follow-up if we are already happy with this (already big) PR.

bpf/process/pfilter.h
```diff
@@ -414,6 +415,196 @@ struct nc_filter {
...
+		filter = (struct selector_arg_filter *)&f[filter_off];
+
+		if (filter->type != string_type)
+			return 0;
+
+		key = filter->index;
+		if (key >= CMD_ARGS_MAX || key >= argc)
+			return 0;
+
+		argsoff = arg_offsets[key];
```
olsajiri, line 586: should arg_offsets be unsigned char * ? reading `__u32 argsoff` out of `char arg_offsets`  could extend the sign bit (ending up with 0xFFFFFFFF instead of 255)

olsajiri (approved):
thanks, left one small comment

bpf/process/pfilter.h
mtardy: yeah let me fix this

PR #5446, opened 2026-08-17
pkg/process: avoid calling GetPodInfo in non k8s deployments

### LINKED ISSUE #219: tetragon: GetPodInfo() is called even a non k8s deployments
GetPodInfo() where we get the pod info for the events is being called even if Tetragon is run with `enable-k8s-api:false` , this is not an urgent or real issue unless it starts pulling some other bugs...

The ideal case would be to disable K8s code completely for such deployments, but before the question: what would be the side effects on k8s deployments then ?  needs to be answered.

t0x01:
### Description

`GetPodInfo()` gets called whether the `enable-k8s-api` option is true or not. In an environment with plain docker containers, this leads to numerous `tetragon_watcher_errors_total{error="failed_to_get_pod",watcher="k8s"}` errors in metrics.

Add a check for `option.Config.EnableK8s` before calling `GetPodInfo()`.

Fixes: #219

### Changelog

```release-note
process: Add a missing guard for GetPodInfo call to avoid attempts to collect k8s pod metadata in non-k8s deployments.
```

### DIFF
--- a/pkg/process/process.go
+++ b/pkg/process/process.go
@@ -330,7 +330,10 @@ func initProcessInternalExec(
 	}
 	creds := &event.Msg.Creds
 	execID := GetExecID(&process)
-	protoPod := GetPodInfo(event.Kube.Docker, process.Filename, args, process.NSPID)
+	var protoPod *tetragon.Pod
+	if option.Config.EnableK8s && event.Kube.Docker != "" {
+		protoPod = GetPodInfo(event.Kube.Docker, process.Filename, args, process.NSPID)
+	}
 	apiCaps := caps.GetMsgCapabilities(event.Msg.Creds.Cap)
 	binary := path.GetBinaryAbsolutePath(process.Filename, cwd)
 	apiNs, err := namespace.GetMsgNamespaces(event.Msg.Namespaces)


mtardy:
I see that I authored this commit apparently? do you know why?

image

mtardy (changes requested):
just trying to understand before reviewing

t0x01:
I basically took the code you proposed [here](https://github.com/cilium/tetragon/issues/219#issuecomment-3077912043), so i thought it would be fair to indicate you as an author of the commit. I was just trying to preserve code authorship. If that's improper, i apologize. If that's the case, what should i change it to?

mtardy:
> I basically took the code you proposed [here](https://github.com/cilium/tetragon/issues/219#issuecomment-3077912043), so i thought it would be fair to indicate you as an author of the commit. I was just trying to preserve code authorship. If that's improper, i apologize. If that's the case, what should i change it to?

ah okay thanks for the precision, sorry I completely forgot about that. I think it would be okay to keep you as author and put me as `Co-authored-by` if you want to credit me :) thanks! 

I think the author and signed-off-by need to match anyway!

pkg/process/process.go
```diff
@@ -320,17 +320,19 @@ func initProcessInternalExec(
 	event *tetragonAPI.MsgExecveEventUnix,
 	parent tetragonAPI.MsgExecveKey,
 ) *ProcessInternal {
+	var parentExecID string
+	var protoPod *tetragon.Pod
+	var user *tetragon.UserRecord
+
```
mtardy: why declare everything at the top? Our convention in Go is just to declare along the way when needed, with `:=` or before when needed on multiple sub scopes as done here.

pkg/process/process.go
```diff
@@ -320,17 +320,19 @@ func initProcessInternalExec(
...
 	args, cwd := ArgsDecoder(process.Args, process.Flags)
-	var parentExecID string
 	if parent.Pid != 0 {
 		parentExecID = GetExecIDFromKey(&parent)
 	} else {
 		parentExecID = GetProcessID(0, 1)
 	}
 	creds := &event.Msg.Creds
 	execID := GetExecID(&process)
-	protoPod := GetPodInfo(event.Kube.Docker, process.Filename, args, process.NSPID)
```
mtardy, line 341: maybe add your replaced lines in place of this existing line?

mtardy (changes requested):
I think the patch is fine just please rework a bit to reduce the churn and just replace line 333 with your 3 patched lines :) thanks!

mtardy (approved):
lgtm

mtardy:
also please can you just put the explanation in git commit log and wrap the lines? thanks!

t0x01:
Sorry for the poor commit quality on my part. Lots of different attention points needed to be addressed lately. Should've given it more thought.

mtardy:
> Sorry for the poor commit quality on my part. Lots of different attention points needed to be addressed lately. Should've given it more thought.

ah it's all good don't worry, thanks for the fix!

PR #5460, opened 2026-08-20
bpf: clear stale LSM output state before policy filtering

### LINKED ISSUE #5237: BPF-LSM: generic_lsm_output re-emits a stale per-CPU event on every unrelated syscall, producing phantom events for pods that did nothing

iodn:
Supersedes #5388, whose head fork was deleted and therefore cannot be reopened. This replacement keeps the same fix and splits it into two commits: the BPF fix followed by its regression test.

Fixes #5237

### Description

The generic LSM output program can read a stale per-CPU message before the core program runs.

In `generic_start_process_filter()`, `lsm.post` and `common.flags` were only cleared after the configuration and policy-filter checks. If an invocation was rejected by one of those checks, the function returned early and left the previous state intact. The output program could then emit the last matching event again for an unrelated hook invocation.

This change clears the LSM output state immediately after retrieving the per-CPU message, before any early return. The follow-up commit adds a same-CPU namespaced regression test that verifies only the expected `file_open` event is delivered.

### Changelog

```release-note
bpf: Prevent namespaced LSM policies from emitting stale duplicate events
```

### DIFF
--- a/bpf/process/bpf_generic_tracepoint.c
+++ b/bpf/process/bpf_generic_tracepoint.c
@@ -262,7 +262,6 @@ generic_tracepoint_event(struct generic_tracepoint_event_arg *ctx)
 #ifdef __CAP_CHANGES_FILTER
 	msg->sel.match_cap = 0;
 #endif
-	msg->common.flags = 0;
 	tail_call(ctx, &tp_calls, TAIL_CALL_FILTER);
 	return 0;
 }
--- a/bpf/process/bpf_generic_tracepoint.c
+++ b/bpf/process/bpf_generic_tracepoint.c
@@ -262,6 +262,7 @@ generic_tracepoint_event(struct generic_tracepoint_event_arg *ctx)
 #ifdef __CAP_CHANGES_FILTER
 	msg->sel.match_cap = 0;
 #endif
+	msg->common.flags = 0;
 	tail_call(ctx, &tp_calls, TAIL_CALL_FILTER);
 	return 0;
 }
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -50,6 +50,12 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 	if (!msg)
 		return 0;
 
+#ifdef GENERIC_LSM
+	/* The LSM output program runs even when this program returns early. */
+	msg->lsm.post = false;
+	msg->common.flags = 0;
+#endif
+
 	/* setup index, check policy filter, and setup function id */
 	msg->idx = get_index(ctx);
 	config = map_lookup_elem(&config_map, &msg->idx);
@@ -81,9 +87,6 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 #ifdef __CAP_CHANGES_FILTER
 	msg->sel.match_cap = 0;
 #endif
-
-	msg->lsm.post = false;
-	msg->common.flags = 0;
 
 	/* Tail call into filters. */
 	tail_call(ctx, calls, TAIL_CALL_FILTER);
@@ -698,6 +701,7 @@ FUNC_INLINE void
 generic_process_init(struct msg_generic_kprobe *e, u8 op)
 {
 	e->common.op = op;
+	e->common.flags = 0;
 
 	e->common.pad[0] = 0;
 	e->common.pad[1] = 0;
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -53,8 +53,9 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 #ifdef GENERIC_LSM
 	/* The LSM output program runs even when this program returns early. */
 	msg->lsm.post = false;
-	msg->common.flags = 0;
 #endif
+	/* Filtering can set MSG_COMMON_FLAG_PROCESS_NOT_FOUND. */
+	msg->common.flags = 0;
 
 	/* setup index, check policy filter, and setup function id */
 	msg->idx = get_index(ctx);
@@ -701,7 +702,6 @@ FUNC_INLINE void
 generic_process_init(struct msg_generic_kprobe *e, u8 op)
 {
 	e->common.op = op;
-	e->common.flags = 0;
 
 	e->common.pad[0] = 0;
 	e->common.pad[1] = 0;
--- a/pkg/sensors/tracing/policyfilter_test.go
+++ b/pkg/sensors/tracing/policyfilter_test.go
@@ -6,8 +6,10 @@
 package tracing
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -22,12 +24,14 @@ import (
 	"github.com/google/uuid"
 	"github.com/opencontainers/runtime-spec/specs-go"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/cilium/tetragon/pkg/api/tracingapi"
 	"github.com/cilium/tetragon/pkg/bpf"
 	"github.com/cilium/tetragon/pkg/build"
 	tgcgroups "github.com/cilium/tetragon/pkg/cgroups"
+	"github.com/cilium/tetragon/pkg/config"
 	grpcexec "github.com/cilium/tetragon/pkg/grpc/exec"
 	"github.com/cilium/tetragon/pkg/grpc/tracing"
 	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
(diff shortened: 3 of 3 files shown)


iodn:
Thanks for the review (https://github.com/cilium/tetragon/pull/5388#pullrequestreview-4895727926). I updated the series to address these points.

The first commit explains the bug and its trigger: with a namespaced LSM policy using the Post action, a filtered-out invocation on the same CPU could retain the previous matching invocation’s post state and re-emit that stale event.

The second commit is the regression test. It pins matching and non-matching processes to one CPU, opens the matching target file first, and then opens /dev/null five times from the non-matching namespace. Before the fix, those rejected calls re-emitted the stale event, so the test would see six events instead of one.

I also split the series into the fix and test commits, and rebased it onto the current main branch.

bpf/process/generic_calls.h
```diff
@@ -50,6 +50,10 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 	if (!msg)
 		return 0;
 
+	/* Clear state consumed by LSM output before any early return. */
+	msg->lsm.post = false;
```
olsajiri, line 55: we have `GENERIC_LSM` section in `generic_process_event_and_setup`, would it be ok to init this in there?

bpf/process/generic_calls.h
```diff
@@ -50,6 +50,10 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
...
+	msg->common.flags = 0;
```
olsajiri, line 56: I think this should go to `generic_process_init`, we could remove that from `generic_tracepoint_event` then

pkg/sensors/tracing/policyfilter_test.go
```diff
@@ -344,6 +348,150 @@ func TestUprobeNamespacedPolicy(t *testing.T) {
 	require.ErrorContains(t, err, "uprobe")
 }
 
+func TestLsmNamespacedPolicyClearsPostState(t *testing.T) {
```
olsajiri, line 351: could this be added as policytest instead?

olsajiri (commented):
thanks, left few comments

bpf/process/generic_calls.h
iodn: I don't think we can initialize lsm.post there. `generic_process_event_and_setup()` is only reached after the configuration and policy filters accept the invocation. The stale event occurs when `generic_start_process_filter()` returns before that point, while the separate LSM output program still runs. So I kept the reset before those early returns and scoped it to `GENERIC_LSM`.

bpf/process/generic_calls.h
iodn: Done. I moved the normal `common.flags` initialization into `generic_process_init()` and removed the duplicate initialization from `generic_tracepoint_event()`. The LSM path still resets it before policy filtering because the separate output program reads the flags even when the core program returns early

pkg/sensors/tracing/policyfilter_test.go
iodn: Hmm, I checked the policytest framework (including the version on current main) but it does not currently support namespaced policies and has no way to register synthetic pod or cgroup membership in the agent policy filter. This regression also needs both processes pinned to the same CPU. Moving it to policytest would then stop reproducing the original failure, so I kept it in policyfilter_test.go.

bpf/process/bpf_generic_tracepoint.c
```diff
@@ -262,7 +262,6 @@ generic_tracepoint_event(struct generic_tracepoint_event_arg *ctx)
 #ifdef __CAP_CHANGES_FILTER
 	msg->sel.match_cap = 0;
 #endif
-	msg->common.flags = 0;
```
kkourt, line 265: Is it safe to remove this?

AFAICT, we do not call `generic_process_event_and_setup` from `bpf_generic_tracepoint.c`.

kkourt (changes requested):

bpf/process/bpf_generic_tracepoint.c
iodn: Yes, tracepoints call `generic_process_init()` directly even though they don't use `generic_process_event_and_setup()`. This PR moves the flags reset into that initializer, so they still get cleared and the removed assignment is redundant

pkg/sensors/tracing/policyfilter_test.go
```diff
@@ -344,6 +348,150 @@ func TestUprobeNamespacedPolicy(t *testing.T) {
...
+			cpu = i
+			break
+		}
+	}
+	require.NotEqual(t, -1, cpu, "no available CPU found")
+
+	var affinity unix.CPUSet
+	affinity.Set(cpu)
+	require.NoError(t, unix.SchedSetaffinity(matchingProcess.pid(), &affinity))
+	require.NoError(t, unix.SchedSetaffinity(nonMatchingProcess.pid(), &affinity))
```
kkourt, line 398: I guess we need affinity to ensure that the two tests run on the same CPU?
If that's the case, it might be worth to add a comment to that effect.

kkourt (approved):
LGTM, thanks!

Edit: assuming we get CI to a green state :)

kkourt:
Can't see what the CI failures are about. Will restart.

kkourt:
@iodn can you maybe have a look at the golangci failures? You can reproduce them locally using `make check`

iodn:
> @iodn can you maybe have a look at the golangci failures? You can reproduce them locally using `make check`

Thanks, I added the affinity comment and fixed the integer range lint warnings.

I also found the cause of the sigkill test failur. My earlier change cleared PROCESS_NOT_FOUND in `generic_process_init()`, after filtering had set it. I moved the reset back before filtering and restored the tracepoint specific reset

CodeWithHusain:
I took a closer look at this while investigating the shared `process_call_heap` state across separately attached LSM programs. I think the underlying issue is that the current per-CPU storage does not provide enough ownership/correlation when multiple tasks can execute the same hook.

### Root cause

`process_call_heap` is currently a `BPF_MAP_TYPE_PERCPU_ARRAY` with a single entry. This effectively makes the heap ownership:

```text
CPU → heap
```

rather than:

```text
task → heap
```

If execution of two different tasks/program instances is interleaved on the same CPU before the heap contents are consumed, the later lookup can observe state produced by the other task. Since the per-CPU entry contains no task identity, there is no way to verify that the data belongs to the current task.

This can explain the phantom/misattributed process or pod events reported in this issue.

### Existing precedent

There are already two relevant fixes in the codebase:

* **#5506** changed `process_call_heap` from a per-CPU map to a hash map for the uprobe/usdt paths, since those paths cannot rely on preemption being disabled.
* **#5460** fixed a related LSM stale-state issue by clearing the per-CPU state immediately after retrieving the heap, before returning from the hook.

For #5237, I think we need the stronger form of the fix: **remove CPU affinity from the heap ownership and explicitly associate the entry with the current task.**

### Proposed solution

Change the map to a bounded LRU hash keyed by `pid_tgid`:

```c
struct {
    __uint(type, BPF_MAP_TYPE_LRU_HASH);
    __uint(max_entries, 32768);
    __uint(map_flags, BPF_F_NO_PREALLOC);
    __type(key, __u64);  // pid_tgid
    __type(value, struct msg_generic_kprobe);
} process_call_heap SEC(".maps");
```

Then make heap access task-specific:

```c
static __always_inline struct msg_generic_kprobe *
get_heap(void)
{
    __u64 pid_tgid = bpf_get_current_pid_tgid();
    struct msg_generic_kprobe *heap;

    heap = bpf_map_lookup_elem(&process_call_heap, &pid_tgid);
    if (heap)
        return heap;

    struct msg_generic_kprobe zero_val = {};

    if (bpf_map_update_elem(&process_call_heap, &pid_tgid,
                             &zero_val, BPF_NOEXIST))
        return NULL;

    return bpf_map_lookup_elem(&process_call_heap, &pid_tgid);
}
```

This changes the ownership model from:

```text
CPU 0 ──→ shared heap
CPU 1 ──→ shared heap
```

to:

```text
task A ──→ heap A
task B ──→ heap B
task C ──→ heap C
```

so a task cannot accidentally consume another task's heap entry merely because both executions happened on the same CPU.

Using an LRU hash also keeps the map bounded and avoids requiring a separate task-exit cleanup path.

### Trade-offs

The main cost is the additional hash-map lookup/update and memory overhead compared with a per-CPU array. However, this is already an accepted pattern in **#5506**, and correctness here seems more important than preserving the cheaper per-CPU representation.

The `BPF_NOEXIST` update also makes initialization safe if two executions attempt to create the same task entry concurrently: one succeeds and the other can simply re-read the existing entry.

If this direction looks reasonable, I can put together a draft PR with the map change and update the relevant heap access paths/tests.

olsajiri:
> so a task cannot accidentally consume another task's heap entry merely because both executions happened on the same CPU.
> 
> Using an LRU hash also keeps the map bounded and avoids requiring a separate task-exit cleanup path.

I'm not sure how to size the LRU hash for this case not to waste the and in the same time to be sure we won't get the entry removed before the last lsm program uses it.. so I'd vote for standard hash with explicit removal.. but I might be missing some details about LRU hash, will check

> ### Trade-offs
> 
> The main cost is the additional hash-map lookup/update and memory overhead compared with a per-CPU array. However, this is already an accepted pattern in **#5506**, and correctness here seems more important than preserving the cheaper per-CPU representation.
> 
> The `BPF_NOEXIST` update also makes initialization safe if two executions attempt to create the same task entry concurrently: one succeeds and the other can simply re-read the existing entry.
> 
> If this direction looks reasonable, I can put together a draft PR with the map change and update the relevant heap access paths/tests.

yea, I think we need that fix for lsm, please submit the PR

olsajiri:
> yea, I think we need that fix for lsm, please submit the PR

fyi on top of https://github.com/cilium/tetragon/pull/5506 seems like we just need to add `GENERIC_LSM` for `USE_HASH_HEAP` ... and the go plumbing for attaching the map

PR #5461, opened 2026-08-20
lsm: support matchData selectors

### LINKED ISSUE #5373: lsmhooks do not support matchData despite being documented in the TracingPolicy API reference

iodn:
Supersedes #5390, whose head fork was deleted and therefore cannot be reopened. This replacement preserves the original commit unchanged.

Fixes #5373

### Description

LSM hook selectors exposed `matchData`, but `LsmHookSpec` had no corresponding data field and the generic LSM sensor always passed an empty data list to the selector parser. Consequently, policies using `matchData` failed to load with an out-of-bounds index error.

This adds data-configuration support to LSM hooks using the same argument handling as the other generic sensors. Current task fields can now be resolved and filtered through `matchData` and included in the `ProcessLsm` event data.

Regression coverage verifies policy validation, selector initialization, and LSM event output.

### Changelog

```release-note
tracing: Support matchData selectors for LSM hooks
```


iodn:
I rebased the change on the current main branch and split it into three logical commits:

1. api: add data fields to LSM hooks and events policy and gRPC API fields, plus generated CRD, schema, vendor, and reference artifacts.
2. pkg/sensors: support matchData selectors for LSM hooks, selector initialization and event handling.
3. tests: cover LSM matchData selectors, validation, setup, and event-output coverage.

Each commit now has a descriptive body explaining why it is needed. I also ran the focused gRPC and LSM validation checks. I would appreciate another look.

pkg/sensors/tracing/lsm_validation_test.go
```diff
@@ -122,6 +123,70 @@ spec:
...
+	require.NoError(t, err)
+}
+
+func TestAddLsmMatchData(t *testing.T) {
+	if !bpf.HasProgramLargeSize() {
+		t.Skip("large BPF programs not supported")
+	}
+	forceLargeProgs(t)
+
+	data := v1alpha1.KProbeArg{
```
olsajiri: I think it's better to have spec in string form like we do for the other tests

api/v1/tetragon/tetragon.proto
```diff
@@ -657,6 +657,8 @@ message ProcessLsm {
   repeated Process ancestors = 10;
   // IMA file hash. Format algorithm:value.
   string ima_hash = 11;
+  // Data definition of the observed LSM hook.
+  repeated KprobeArgument data = 12;
```
olsajiri, line 661: please split tetragon.proto change and the generated changes.. to ease up the review

pkg/grpc/tracing/tracing.go
```diff
@@ -1098,6 +1098,7 @@ type MsgGenericLsmUnix struct {
 	Msg        *tracingapi.MsgGenericKprobe
 	Hook       string
 	Args       []tracingapi.MsgGenericKprobeArg
+	Data       []tracingapi.MsgGenericKprobeArg
```
olsajiri, line 1101: please split the grpc code out of the first commit

olsajiri (commented):
lgtm, left few comments, thanks

pkg/sensors/tracing/lsm_validation_test.go
iodn: Yep, I changed the setup test to parse the policy from a YAML string and reused the same specification for the validation test

api/v1/tetragon/tetragon.proto
iodn: Done, I split the handwritten API definitions

pkg/grpc/tracing/tracing.go
iodn: Done, I moved the gRPC event conversion into a separate commit

kkourt (commented):
Thanks! From a quick look, PR LGTM.
It would be nice if we could avoid the duplication between the different implementations (e.g., kprobe, LSM) but I'm not sure how easy that would be.

Could you please rebase your PR?

Also, as Jiri suggested it might be helpful to split the proto changes in two commits: i) one for the modification of the actual .proto file, and another with the generated and vendored code.

kkourt (approved):
LGTM, thanks.

olsajiri (approved):

PR #5463, opened 2026-08-20
add checks for policy tests validation on nok8s

dangome3:
Changing from `BTFPath` and `BTFType` to `btfPath` and `btfType` to match schema defined for JSON in the CRDs.

This was the only occurence of a wrong definition in a yml file for these params.

Also added tests for preventing these kind of bugs in the future

### DIFF
--- a/.github/workflows/gotests.yml
+++ b/.github/workflows/gotests.yml
@@ -77,8 +77,8 @@ jobs:
     - name: Run nok8s go tests
       run: |
         cd go/src/github.com/cilium/tetragon/
-        go vet -tags nok8s ./pkg/sensors ./pkg/policyfilter ./pkg/tracingpolicy
-        go test -tags nok8s ./pkg/tracingpolicy
+        go vet -tags nok8s ./pkg/sensors ./pkg/policyfilter ./pkg/tracingpolicy ./pkg/testutils/policytest
+        go test -tags nok8s ./pkg/tracingpolicy ./pkg/testutils/policytest
 
     - name: Upload Tetragon logs
       if: failure()
--- a/pkg/testutils/policytest/render_test.go
+++ b/pkg/testutils/policytest/render_test.go
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright Authors of Tetragon
+
+//go:build !windows
+
+package policytest_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/cilium/tetragon/pkg/testutils/policytest"
+	"github.com/cilium/tetragon/pkg/tracingpolicy"
+	_ "github.com/cilium/tetragon/tests/policytests" // so that the policies are registered
+)
+
+// TestRenderPolicyTests renders the policy of every registered policytest, for
+// every combination of its parameters, and checks that it parses and validates.
+func TestRenderPolicyTests(t *testing.T) {
+	all := policytest.AllPolicyTests
+	require.NotZero(t, all.Len(), "no policy tests registered")
+
+	for i := range all.Len() {
+		pt := all.Get(i)
+		t.Run(pt.Name, func(t *testing.T) {
+			if pt.Policy == nil {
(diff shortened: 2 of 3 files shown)


FedeDP (approved):

mtardy (commented):
wait a minute why didn't it fail then before?

FedeDP:
I don't think we are testing policytests in nok8s mode.
Eventually, might be worth to run them!

EDIT: @dangome3 can you double check if this is true? And in case, can you add a commit on top to enable policytests to run in nok8s mode, if that's easy enough?

Oh well, but you are right @mtardy i am not sure why we do not trigger the same error in k8s builds though :/ sorry needed a :coffee: to catch up :laughing:

dangome3:
> @dangome3 can you double check if this is true? And in case, can you add a commit on top to enable policytests to run in nok8s mode, if that's easy enough?

Sure! let me get into it!

dangome3:
I added a test to parse and validate the policy tests. Here the results

Without fixing the typo:
```
        	            	failed to validate TracingPolicy: jsonschema validation failed with 'file:///tracingpolicy#'
        	            	- at '/spec/uprobes/0': additional properties 'BTFPath' not allowed
        	Test:       	TestRenderPolicies/resolve-nested-anon-struct
        	Messages:   	failed to parse policy:

        	            	kind: TracingPolicy
        	            	metadata:
        	            	  name: "uprobe"
        	            	spec:
        	            	  uprobes:
        	            	  - path: /tetragon-policytests-bins/resolve-nested-anon-struct
        	            	    BTFPath: /tetragon-policytests-bins/resolve-nested-anon-struct.btf
        	            	    symbols:
        	            	    - "passit"
        	            	    args:
        	            	    - index: 0
        	            	      type: "int"
        	            	      resolve: "nested.ans.second"
        	            	      btfType: "mystruct"
        	            	    - index: 0
        	            	      type: "int"
        	            	      resolve: "pnested.pans.second"
        	            	      btfType: "mystruct"

FAIL
FAIL	github.com/cilium/tetragon/pkg/testutils/policytest	0.034s
FAIL
```

with typo fixed:
```
ok  	github.com/cilium/tetragon/pkg/testutils/policytest
```

cc: @FedeDP

FedeDP:
Nice!
@mtardy for k8s builds, the problem is going to be fixed in https://github.com/cilium/tetragon/pull/5505; basically k8s `FromYAML` uses internally  `err = yaml.UnmarshalStrict(jsonData, cr)` (from `sigs.k8s.io/yaml`) that calls `err = jsonUnmarshal(bytes.NewReader(jsonBytes), obj, opts...)`; this function internally creates a `*json.Decoder` that, for golang json v1 is  by default case insensitive; therefore we end up being case insensitive.
My #5505  fixes that by leveraging the new json/v2 from go 1.27 that is by default case sensitive.

mtardy:
> Nice! @mtardy for k8s builds, the problem is going to be fixed in #5505; basically k8s `FromYAML` uses internally `err = yaml.UnmarshalStrict(jsonData, cr)` (from `sigs.k8s.io/yaml`) that calls `err = jsonUnmarshal(bytes.NewReader(jsonBytes), obj, opts...)`; this function internally creates a `*json.Decoder` that, for golang json v1 is by default case insensitive; therefore we end up being case insensitive. My #5505 fixes that by leveraging the new json/v2 from go 1.27 that is by default case sensitive.

Was it an issue? Was it a discrepancy between the k8s API server and us or does the API server is also case insensitive?

What I'm saying is that we want to do whatever the k8s API server does, so if they are case insensitive and ignore unknown fields, I wouldn't say it's a bug, if they don't then we want to fix it.

FedeDP:
> Was it a discrepancy between the k8s API server and us or does the API server is also case insensitive?

That's it, k8s would reject it but we wouldn't, since it relies upon the OpenAPI spec.
See https://github.com/OAI/OpenAPI-Specification/blob/main/versions/3.2.0.md#case-sensitivity

mtardy (approved):
ok thanks for the details! lgtm

PR #5467, opened 2026-08-21
contrib/verify: improve the verifier tests

mtardy:
I have been bumping into this while building https://github.com/cilium/tetragon/pull/5442 and the output wasn't super useful, let's fix a few issues (we were not skipping correctly the programs built for a kernel version newer than the current) and make the output more useful.

### DIFF
--- a/contrib/verify/verify_test.go
+++ b/contrib/verify/verify_test.go
@@ -27,6 +27,44 @@ const (
 	tetragondir = "/var/lib/tetragon"
 )
 
+func extractKernelVersion(t *testing.T, fileName string) string {
+	if idx := strings.LastIndex(fileName, "_v"); idx != -1 {
+		versionPart := strings.TrimSuffix(fileName[idx+2:], ".o")
+		if len(versionPart) >= 2 {
+			if versionPart[:1] == "1" {
+				t.Fatalf("version 10.x not supported (we will need to disambiguate notation): %s", versionPart)
+			}
+			return versionPart[:1] + "." + versionPart[1:]
+		}
+	}
+
+	return ""
+}
+
+func TestExtractKernelVersion(t *testing.T) {
+	tests := []struct {
+		fileName string
+		expected string
+	}{
+		{"bpf_generic_kprobe_v612.o", "6.12"},
+		{"bpf_generic_kprobe_v61.o", "6.1"},
+		{"bpf_generic_kprobe_v511.o", "5.11"},
+		{"bpf_generic_kprobe_v53.o", "5.3"},
+		{"bpf_generic_kprobe.o", ""},
+		{"bpf_alignchecker.o", ""},
+		{"bpf_loader_v511.o", "5.11"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.fileName, func(t *testing.T) {
+			result := extractKernelVersion(t, tt.fileName)
+			if result != tt.expected {
+				t.Errorf("extractKernelVersion(%q) = %q, want %q", tt.fileName, result, tt.expected)
+			}
+		})
+	}
+}
+
 func TestVerifyTetragonPrograms(t *testing.T) {
 
 	tetragonDir := os.Getenv("TETRAGONDIR")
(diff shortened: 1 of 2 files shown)


contrib/verify/verify_test.go
```diff
@@ -27,6 +27,44 @@ const (
 	tetragondir = "/var/lib/tetragon"
 )
 
+func extractKernelVersion(t *testing.T, fileName string) string {
+	if idx := strings.LastIndex(fileName, "_v"); idx != -1 {
```
3u13r, line 33: I'd prefer a early return style here, e.g. 
```go
idx := strings.LastIndex(fileName, "_v");
if idx == -1 {
    return ""
}
// ...
```

But it's a style discussion in a test helper, so no need to do anything.

contrib/verify/verify_test.go
```diff
@@ -177,15 +170,16 @@ func TestVerifyTetragonPrograms(t *testing.T) {
...
-				t.Logf("%+v", ve)
+				if *debug {
+					t.Logf("%+v", ve)
+				}
 
 				_, kver, _ := kernels.GetKernelVersion("", "/proc")
-				t.Logf("failed object %s, kernel %s", fileName, kver)
+				t.Errorf("%s ✗: failed to load object on kernel %s: %v", fileName, kver, err)
 			}
+			continue
```
3u13r, line 177: What if we have an error but it's not a verifier error? We should still fail the test and print it, right?

3u13r (changes requested):

contrib/verify/verify_test.go
mtardy: good point! thanks

contrib/verify/verify_test.go
mtardy: yeah it even makes the function a bit more complex to read imo
```go
func extractKernelVersion(t *testing.T, fileName string) string {
	idx := strings.LastIndex(fileName, "_v")
	if idx == -1 {
		return ""
	}

	versionPart := strings.TrimSuffix(fileName[idx+2:], ".o")
	if len(versionPart) >= 2 {
		if versionPart[:1] == "1" {
			t.Fatalf("version 10.x not supported (we will need to disambiguate notation): %s", versionPart)
		}
		return versionPart[:1] + "." + versionPart[1:]
	}
	return ""
}
```

contrib/verify/verify_test.go
```diff
@@ -143,21 +169,15 @@ func TestVerifyTetragonPrograms(t *testing.T) {
...
-			if errors.As(err, &ve) {
-				fmt.Printf("%+v\n", ve)
-
-				_, kver, _ := kernels.GetKernelVersion("", "/proc")
-				fmt.Printf("failed object %s, kernel %s\n", fileName, kver)
+			if errors.As(err, &ve) && *debug {
+				t.Logf("%+v", ve)
 			}
+			_, kver, _ := kernels.GetKernelVersion("", "/proc")
+			t.Errorf("%s ✗: failed to load object on kernel %s: %v", fileName, kver, err)
```
3u13r, line 176: We print `err` independent if it is a verifier error or if debug is enabled. So the 
```go
if errors.As(err, &ve) && *debug {
```
looks superfluous or rather prints the error twice. Or am I mistaken?



Maybe something like:
```go
if errors.As(err, &ve) && !*debug {
	err = errors.New("verifier error, run with -debug to print the full verifier output")
}
_, kver, _ := kernels.GetKernelVersion("", "/proc")
t.Errorf("%s ✗: failed to load object on kernel %s: %v", fileName, kver, err)
```


Also we can omit the `continue` since `t.Errorf()` terminates the test, right? ([ref](https://pkg.go.dev/testing#T.Errorf))

3u13r (changes requested):

contrib/verify/verify_test.go
mtardy: ah no `t.Logf("%+v", ve)` will print the full verifier log while `%v` for err will be the summary oneliner.

t.Error doesn't terminate the tests, that's one of the changes of this, we keep trying the other progs

3u13r: Re: `t.Errorf()`: The docs state `Errorf is equivalent to Logf followed by Fail.`. Is there an override of `t.Errorf` that I'm missing?

mtardy: yeah Fail is not stopping execution, FailNow is

mtardy: In go source code
```go
// Fail marks the function as having failed but continues execution.
func (c *common) Fail() {
	if c.parent != nil {
		c.parent.Fail()
	}
	c.mu.Lock()
	defer c.mu.Unlock()
	// c.done needs to be locked to synchronize checks to c.done in parent tests.
	if c.done {
		panic("Fail in goroutine after " + c.name + " has completed")
	}
	c.failed = true
}
```

3u13r: Ah my bad. I also saw that we have custom formatters for `+v`, TIL.

3u13r (approved):

contrib/verify/verify_test.go
mtardy: no prob!!

will-isovalent (approved):

PR #5473, opened 2026-08-21
ci: improve formatting and code generation checks

3u13r:
### Description

This is somewhat opinionated in that we used `goimports` as a formatter in the Makefile but `go fmt` in the CI. Since `goimports` is a superset of `go fmt`, I've decided to align both places to it.

### Changelog

```release-note
ci: improve formatting and code generation checks
```


tpapagian (approved):

go.mod
```diff
@@ -174,3 +175,5 @@ replace (
 	github.com/cilium/tetragon/api => ./api
 	github.com/cilium/tetragon/pkg/k8s => ./pkg/k8s
 )
+
+tool golang.org/x/tools/cmd/goimports
```
mtardy: we might want to add all this into a separate go tool mod files as some project do it to not mix the tooling deps and the project deps. k8s does this, I did this recently in sig-security/srctl.

It might be a good idea to add golangci-lint like this in the future.

mtardy: typically golangci-lint imports so many things https://github.com/kubernetes/sig-security/blob/main/sig-security-tooling/srctl/tools/go.mod we might want to put it under contrib

3u13r: Sounds good. We already depend on the module of goimports, but I agree that it's way cleaner in a separate go.mod.

3u13r: Thanks a lot for the srctl pointer. I had to invoke goimports slightly differently via `-C gotools` instead of `-modfile=gotools/go.mod` as the second option formatted the go source files using `tools/go.mod` as it's modfile.

Also decided on `gotools` since we already have a `tools` folder which both includes tools the "old" import way as well as go source uses the project root's go.mod file. I'll leave the untangling of that to a later date if that's fine with you.

mtardy: would it be possible to ship it within tools ~~or within contrib~~ to not create a new top level directory? Also we would need to add that new go file in the renovate config.

mtardy: maybe also it's okay to not vendor the go tool go.mod?

I think it would be cool to update our tooling like the rest of the deps (like it would simplify many things with golangci-lint, I could made that PR once you land that one).

mtardy: not sure to follow why putting in `tools` was a bit too complicated sorry. If you want, split that part (the commit using the new go tool directive) of the PR and we can do it in a separate thing.

mtardy (commented):
maybe merge this PR with the first 4 commits and we do the rest in another PR as it's a bit more work actually to create a new go.mod

go.mod
3u13r: > maybe also it's okay to not vendor the go tool go.mod?

What is the original reason why all other go.mod files vendor their dependencies? I don't mind either way. 

> Also we would need to add that new go file in the renovate config.

Done

> not sure to follow why putting in tools was a bit too complicated sorry.

Changed it to have the go.mod in `tools`.

mtardy: > What is the original reason why all other go.mod files vendor their dependencies? I don't mind either way.

I think it's so that the project is self sufficient and can be built from an archive. For the development tooling I guess it's less important.

mtardy: okay cool let's present that idea to the next meeting maybe before merging since it adds one more go.mod but should be okay

vendor/modules.txt
mtardy, line 1: so we are vendoring the go tool thing? I wonder why it appears in the main vendor folder, is it a leftover?

3u13r: Dang it, I pushed from the wrong checkout (maybe time to look into worktrees). 

Now it should match more what you (and I) expect i.e., adding the tool via a non-vendored `go.mod` in `tools/`.

mtardy: will review this one last time tomorrow and merge if everything's good thanks :)

Makefile
```diff
@@ -322,6 +322,11 @@ test-compile: ## Compile unit tests.
 		echo -c ./$$localpkg -o go-tests/$$localtestfile; \
 	done | GOMAXPROCS=1 xargs -P $(JOBS) -L 1 $(GO) test -gcflags=$(GO_BUILD_GCFLAGS)
 
+.PHONY: split-tetragon-gotests
+split-tetragon-gotests:
+	$(GO) -C tools build -o ../bin/split-tetragon-gotests ./split-tetragon-gotests
+	./bin/split-tetragon-gotests -ci-run 1
```
mtardy, line 328: should we use `go run` here then?

Makefile
```diff
@@ -322,6 +322,11 @@ test-compile: ## Compile unit tests.
 		echo -c ./$$localpkg -o go-tests/$$localtestfile; \
 	done | GOMAXPROCS=1 xargs -P $(JOBS) -L 1 $(GO) test -gcflags=$(GO_BUILD_GCFLAGS)
 
+.PHONY: split-tetragon-gotests
+split-tetragon-gotests:
```
mtardy, line 326: btw that's for another PR but I think we could remove this now as we haven't been using it for a long time

mtardy (approved):
Thanks! I think we can merge like this, just a nit on the build/run, and btw maybe we can remove some of these tools now. But that's for another time. The go.mod will be super useful for the tooling anyway! :)

Makefile
3u13r: I mean we use it in the CI, and I think it's nice to have most/all of the inline bash script not in the yaml but as make targets so I can see and reproduce locally what is happening.

Makefile
3u13r: We could right now, but this uses the project root's `go.mod` to compile the tool. If we use any dependency in the tool that's only in `tools/go.mod` and not in `go.mod` compiling/`go run` will fail. Today we don't have such dependencies, but it doesn't sound correct to rely on the fact that we never will.

Compiling with `tools/go.mod` via `go run -C tools split-tetragon-gotests/main.go` executes the binary in the context of the sub-module and we don't get any tests.

The working alternative I see is to add flags to configure the path where split-tetragon-gotests looks for test. That may be something for another PR.

3u13r:
This PR quickly gets out-of-sync. Let me rebase it now. If it diverges again, let's sync on when we want to merge this.

mtardy:
> This PR quickly gets out-of-sync. Let me rebase it now. If it diverges again, let's sync on when we want to merge this.

oops sorry I forgot to merge it after approval

PR #5474, opened 2026-08-21
[v1.5] fix(deps): update module github.com/cilium/ebpf to v0.22.0 [security]

dangome3:
This commit also removes ebpf FlushKernelSpec calls as updating github.com/cilium/ebpf to v0.22.0 fails to build with the following error:

    pkg/bpf/detect_linux.go:576:13: undefined: ebtf.FlushKernelSpec

v0.22.0 drops the process-global BTF cache from btf/kernel.go, and with it FlushKernelSpec(), so the explicit flushes done after feature detection and after each sensor load are no longer needed.

### Changelog

```release-note
update module github.com/cilium/ebpf to v0.22.0
```


dangome3:
golangci-lint (windows) looks like a flaky to me, but I can't trigger again the workflow :)

mtardy (approved):
lgtm, why the cgroup update? it was needed by ebpf?

dangome3:
> lgtm, why the cgroup update? it was needed by ebpf?

that's correct. It needed it because of `asm.Instruction.Sym()`, which ebpf removed in v0.21.00

PR #5485, opened 2026-08-24
new: bpf DEBUG() macro to use rodata config

FedeDP:
### Description

This PR introduces the possibility to enable certain `bpf_trace_printk` messages at tetragon startup via 2 new CLI options:
* `--bpf-debug-area`
* `--bpf-debug-log`

The former allows the user to specify which area of the BPF probe to be traced. It leverages our `SliceEnum` to offer a static list of strings to choose from (the supported bpf areas); currently supported values: `all`, `generic` and `process`.
Debug `bpf_printk` messages in our bpf probes were moved from comp time defines to bpf constant/rodata; finally,
Bpf `DEBUG()` macro was expanded to allow for "areas" tags.

The latter allow to spawn a goroutine that continuously fetches lines from `/sys/kernel/debug/tracing/trace_pipe` and forwards them to tetragon log as INFO level messages.

The idea is that after this, we can start adding more debug messages that can not only be useful while developing, but might also be good to debug users' issues (eg: ask them to start tetragon with `--bpf-debug-area "process" --bpf-debug-log` and attach the log).

How it looks:
```
level=info msg="tetragon@BPF_AREA_PROCESS | set_in_init_tree: parent in init tree"
level=info msg="tetragon@BPF_AREA_PROCESS | set_in_init_tree: parent in init tree"
```

### Changelog

```release-note
new: bpf area-specific DEBUG() rodata config
```

### DIFF
--- a/cmd/tetragon/main.go
+++ b/cmd/tetragon/main.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"log/slog"
 	"net"
@@ -673,6 +674,13 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 		go logStatus(ctx, obs)
 	}
 
+	// Start even if BPFDebugAreas is empty;
+	// BPF probe might've been compiled with TETRAGON_BPF_DEBUG
+	// thus all bpf_trace_printk() are forcefully enabled.
+	if option.Config.BPFDebugLog {
+		go logBPFDebug(ctx)
+	}
+
 	return obs.StartReady(ctx, ready)
 }
 
@@ -769,6 +777,45 @@ func logStatus(ctx context.Context, obs *observer.Observer) {
 				obs.PrintStats()
 				prevLost = lost
 				prevErrors = errors
+			}
+		}
+	}
+}
+
+func logBPFDebug(ctx context.Context) {
+	f, err := os.Open("/sys/kernel/debug/tracing/trace_pipe")
+	if err != nil {
+		log.Warn("failed to open /sys/kernel/debug/tracing/trace_pipe", "err", err)
+		return
+	}
+	defer f.Close()
+
+	buf := make([]byte, 4096)
+	for {
+		if ctx.Err() != nil {
+			return
+		}
+		n, err := f.Read(buf)
+		if n > 0 {
+			lines := strings.SplitSeq(string(buf[:n]), "\n")
+			for line := range lines {
+				// Only print tetragon messages
+				i := strings.Index(line, "tetragon")
+				if i != -1 {
+					// Drop all stuff before tetragon prefix
+					logger.GetLogger().Info(line[i:])
+				} else if strings.Contains(line, "LOST") {
+					// Print LOST messages, eg:
+					// CPU:12 [LOST 711 EVENTS]
+					// CPU:3 [LOST 3698 EVENTS]
+					// CPU:1 [LOST 8509 EVENTS]
+					logger.GetLogger().Info(line)
+				}
+			}
+		}
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				return
 			}
 		}
 	}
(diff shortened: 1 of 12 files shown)


pkg/option/enum.go
```diff
@@ -81,7 +81,9 @@ func (e *SliceEnum) Set(p string) error {
 	if !slices.Contains(e.allowed, p) {
 		return fmt.Errorf("invalid argument %s, please provide one of %s", p, e.Allowed())
 	}
-	e.Values = append(e.Values, p)
+	if !slices.Contains(e.Values, p) {
```
FedeDP, line 84: Small change to properly mimic pflag `StringSlice`.

FedeDP:
Converting to draft since i'd like to make this smarter.

cmd/tetragon/main.go
```diff
@@ -774,6 +779,39 @@ func logStatus(ctx context.Context, obs *observer.Observer) {
...
+	for {
+		if ctx.Err() != nil {
+			return
+		}
+		n, err := f.Read(buf)
+		if n > 0 {
+			lines := strings.SplitSeq(string(buf[:n]), "\n")
+			for line := range lines {
+				// Only print tetragon messages
+				i := strings.Index(line, "tetragon")
```
olsajiri, line 803: should we detect LOST data message and display it ?

cmd/tetragon/main.go
```diff
@@ -673,6 +674,10 @@ func tetragonExecuteCtx(ctx context.Context, cancel context.CancelFunc, ready fu
 		go logStatus(ctx, obs)
 	}
 
+	if len(option.Config.BPFDebug.Values) > 0 {
+		go logBPFDebug(ctx)
```
olsajiri, line 681: could this be optional as well? sometimes it's needed to check bpf printk messages with kernel tracing output

FedeDP: Right, so 2 CLI options:
* `--bpf-debug-area`
* `--bpf-debug-log`

I like it!

cmd/tetragon/main.go
FedeDP: Mmmh what is that sorry?

cmd/tetragon/main.go
FedeDP: Done!

cmd/tetragon/main.go
olsajiri: if the trace buffer can't keep up the event/message can be lost, it will emit messages like:

```
root@krava:/sys/kernel/debug/tracing# cat trace_pipe  | grep -i lost
CPU:12 [LOST 711 EVENTS]
CPU:3 [LOST 3698 EVENTS]
CPU:1 [LOST 8509 EVENTS]
CPU:13 [LOST 3698 EVENTS]
CPU:7 [LOST 24660 EVENTS]
CPU:11 [LOST 17640 EVENTS]
CPU:5 [LOST 24180 EVENTS]
CPU:10 [LOST 21960 EVENTS]

```

FedeDP: Oh nice, i think we should actually; thanks for pointing it out

olsajiri (approved):
lgtm

andrewstrohman (approved):
Nice!

andrewstrohman:
One that that I noticed from experimenting with `--bpf-debug-log` and `--bpf-debug-area` independently, is that a trace message from a previous tetragon instance can be logged in the current tetragon instance, because it remains in the trace buffer, unconsumed from the previous run.

Initially, this was a little confusing for me.

We could drain the pre-existing messages before starting log in `logBPFDebug`, but then we might miss some messages that happen at init. So, I don't have a suggestion about how to make this better, but thought I would mention it anyway.

FedeDP:
You are right @andrewstrohman , didn't think of that!
> We could drain the pre-existing messages before starting log in logBPFDebug, but then we might miss some messages that happen at init.

Yes, exactly, unless we drain the messages before loading the bpf probes, ie: as soon as we enter main, in a blocking way.
Don't know whether it is worth it though, but that would be an easy solution.
For now, i'd ship it as-is and eventually fix that later.

PR #5498, opened 2026-08-24
renovate: retain 5-day cooldown for official Docker Hub images

peoyekunle:


mtardy (approved):
I'm a bit unconvinced that it's making our life easier but let's try it fully before reverting it if needed.

PR #5501, opened 2026-08-24
crdutils: reject unknown fields during validation

andrewstrohman:
Set additionalProperties to false recursively on object schemas used by the standalone validator. This makes misspelled and unknown CRD fields produce validation errors instead of being silently accepted.


FedeDP:
Nice, my https://github.com/cilium/tetragon/pull/5505 found same issues as yours :D

andrewstrohman:
closing in favor of https://github.com/cilium/tetragon/pull/5505

PR #5503, opened 2026-08-25
fix(selectors): cap range span in writeRangeInMap to prevent hang

### LINKED ISSUE #5502: Unbounded enumeration in writeRangeInMap causes CPU-pegging hang during policy parsing

`writeRangeInMap` in `pkg/selectors/kernel.go` (lines 401-488) parses a "min:max"
range string for matchArgs and, instead of storing just the two bounds for a
kernel-side range comparison, enumerates every integer in the range as a
separate Go map entry. There is no upper bound on the range span.

**Reproduction:**

1. Apply a TracingPolicy with a matchArgs `InMap` operator and a wide range
   value on a 64-bit typed arg, e.g.:
   ```yaml
   matchArgs:
   - index: 0
     operator: "InMap"
     values:
     - "0:18446744073709551615"

### Tetragon Version

latest

### Kernel Version

N/A

### Kubernetes Version

_No response_

### Bugtool

_No response_

### Relevant log output

```shell

```

### Anything else?

val | entries | elapsed | heap |
|---|---|---|---|
| 50M | 50M | 4.5s | 752MB |
| 150M | 150M | 16s | 3.3GB |
| 250M | 250M | 38s | 6.1GB |
| 450M | 450M | 90s | 7.3GB |

Growth superlinear (map rehash/GC pressure), not linear. Confirms writeup: this is **hang + memory exhaustion**, not a classic nil-deref/index panic.

hanshal101:
Fixes #5502 
Closes #5502 

## Summary
`writeRangeInMap` (pkg/selectors/kernel.go) parses a "min:max" string for matchArgs InMap/NotInMap and socket operators (Sport/Dport/Protocol/Family/State), and enumerates every integer in the range as a separate map entry instead of storing just the two bounds. There is no bound on the range span, so a policy with a wide range (e.g. "0:18446744073709551615") causes an effectively unbounded loop: single CPU core pegged, memory growing without limit, all inside the synchronous AddTracingPolicy gRPC handler it never returns.

This is separate from the InRange/NotInRange operator path (writeMatchValuesRange), which stores min/max directly for a real kernel-side range comparison and is already capped at 4 values that path is unaffected.

## Fix
Add a `maxRangeSpan` (65536) check before either enumeration loop (signed and unsigned), rejecting oversized ranges with a clear error at policy-parse time instead of hanging.

## Test plan
- [x] Added regression test in `pkg/selectors/kernel_test.go` using an oversized InMap range, asserting `ParseMatchArg` returns an error.
- [x] Verified the test correctly hangs/times out against the pre-fix code (`git stash` the fix) and passes instantly with the fix applied.
- [x] `go vet` / `go build` / `go test ./pkg/selectors/...` all pass.

### DIFF
--- a/pkg/selectors/kernel.go
+++ b/pkg/selectors/kernel.go
@@ -457,10 +457,19 @@ func writeRangeInMap(v string, ty uint32, op uint32, m *ValueMap) error {
 			return fmt.Errorf("unknown type: %d", ty)
 		}
 	}
+	// maxRangeSpan bounds how many entries writeRangeInMap will enumerate into
+	// the map for a single "min:max" value. Without this, a range like
+	// "0:18446744073709551615" would try to insert billions of entries,
+	// pegging a CPU core and exhausting memory while parsing the policy.
+	const maxRangeSpan = 65536
+
 	switch ty {
 	case gt.GenericIntType, gt.GenericS64Type, gt.GenericS32Type, gt.GenericS16Type, gt.GenericS8Type, gt.GenericSyscall64, gt.GenericSizeType:
 		if sRangeVal[0] > sRangeVal[1] {
 			sRangeVal[0], sRangeVal[1] = sRangeVal[1], sRangeVal[0]
+		}
+		if uint64(sRangeVal[1])-uint64(sRangeVal[0]) > maxRangeSpan {
+			return fmt.Errorf("MatchArgs value %s invalid: range span exceeds %d", v, maxRangeSpan)
 		}
 		for val := sRangeVal[0]; ; val++ {
 			var valByte [8]byte
@@ -474,6 +483,9 @@ func writeRangeInMap(v string, ty uint32, op uint32, m *ValueMap) error {
 	case gt.GenericU64Type, gt.GenericU32Type, gt.GenericU16Type, gt.GenericU8Type:
 		if uRangeVal[0] > uRangeVal[1] {
 			uRangeVal[0], uRangeVal[1] = uRangeVal[1], uRangeVal[0]
+		}
+		if uRangeVal[1]-uRangeVal[0] > maxRangeSpan {
+			return fmt.Errorf("MatchArgs value %s invalid: range span exceeds %d", v, maxRangeSpan)
 		}
 		for val := uRangeVal[0]; ; val++ {
 			var valByte [8]byte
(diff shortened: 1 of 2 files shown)


FedeDP:
cc @3u13r if i remember correctly you worked on something similar months ago (https://github.com/cilium/tetragon/commit/1f4623fadf9deb0406a96c0363c7d9f3a519652a)

> This is separate from the InRange/NotInRange operator path (writeMatchValuesRange), which stores min/max directly for a real kernel-side range comparison and is already capped at 4 values that path is unaffected.

Wondering whether we can do the same for these other operators (that would be: `SelectorInMap, SelectorNotInMap` and `SelectorOpSport, SelectorOpDport, SelectorOpNotSport, SelectorOpNotDport, SelectorOpProtocol, SelectorOpFamily, SelectorOpState`) ?

3u13r:
Yes, there wasn't clear consensus either way on the issue if we should enforce a limit ([ref](https://github.com/cilium/tetragon/pull/4848#issuecomment-4256065312)).

My thoughts here are:
1. The effects of people trying to do this could OOM kill other processes, which is bad.
2. Any limit we set is arbitrary.
3. I would suggest to users to use `InRange` for any range over ~8 anyway.

So I'm leaning positive on implementing a restriction in `writeMatchRangesInMap`, which would also cover the the Selectors you mentioned.

pkg/selectors/kernel.go
```diff
@@ -457,11 +457,20 @@ func writeRangeInMap(v string, ty uint32, op uint32, m *ValueMap) error {
...
+	// "0:18446744073709551615" would try to insert billions of entries,
+	// pegging a CPU core and exhausting memory while parsing the policy.
+	const maxRangeSpan = 65536
+
 	switch ty {
 	case gt.GenericIntType, gt.GenericS64Type, gt.GenericS32Type, gt.GenericS16Type, gt.GenericS8Type, gt.GenericSyscall64, gt.GenericSizeType:
 		if sRangeVal[0] > sRangeVal[1] {
 			sRangeVal[0], sRangeVal[1] = sRangeVal[1], sRangeVal[0]
 		}
+		if uint64(sRangeVal[1])-uint64(sRangeVal[0]) > maxRangeSpan {
```
olsajiri, line 471: should we use `>=` in here? given the insert loop below includes the range on both ends?

olsajiri (commented):
I forgot what we decided to do with this, but I'm ok to add the limit

also I wonder we could add some `--expert` mode that would run without such restrictions.. just an idea, no need to implement it in here ;-)

also please add some comment to the commit changelog

PR #5505, opened 2026-08-25
fix(k8s): case sensitive FromYAML()

FedeDP:
### Description

https://github.com/cilium/tetragon/pull/5463 describes a bug that we should be getting in normal k8s builds too; that was not the case since the `FromYAML` function in k8s builds is case insensitive.

### Changelog

```release-note
fix: case sensitive FromYAML() in k8s builds
```

### DIFF
--- a/pkg/crdutils/crdutils.go
+++ b/pkg/crdutils/crdutils.go
@@ -10,6 +10,7 @@ package crdutils
 
 import (
 	"context"
+	"encoding/json/v2"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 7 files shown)


andrewstrohman (approved):

mtardy (approved):
thanks!

mtardy:
I'd be a little be more descriptive on the last commit title as it doesn't really read by itself but otherwise let's ship this. 

Also as a side note, commit titles do not require punctuation as they are not sentences, mostly taking on kernel habits.

FedeDP:
> I'd be a little be more descriptive on the last commit title as it doesn't really read by itself but otherwise let's ship this.

Done!

> Also as a side note, commit titles do not require punctuation as they are not sentences, mostly taking on kernel habits.

Done! Didn't know about this one, thanks!

PR #5506, opened 2026-08-25
tetragon: Switch per-cpu heap to hash for uprobe/usdt sensors

olsajiri:
The uprobe/usdt probes path in kernel do not disable preemption, we need to use hash instead of per-cpu heap.


bpf/process/event_config.h
```diff
@@ -84,7 +84,7 @@ FUNC_INLINE int arg_idx(int index)
 {
 	struct msg_generic_kprobe *e;
 	struct event_config *config;
-	int zero = 0;
+	heap_key_t zero = heap_key();
```
mtardy: all the heap keys named zero are maybe a bit misleading now, maybe change the name?

bpf/process/generic_maps.h
```diff
@@ -5,6 +5,60 @@
...
+
+typedef __u64 heap_key_t;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+	__uint(max_entries, 1); // will be resized by agent
+	__type(key, __u64);
+	__type(value, struct msg_generic_kprobe);
+} process_call_heap SEC(".maps");
```
mtardy, line 25: nit: this is 17 chars, would have been nice to fit in 16

pkg/defaults/defaults_windows.go
```diff
@@ -65,6 +65,10 @@ const (
 	// DefaultSleepableOffloadSize is the default maximum number of entries in the sleepable offload map.
 	DefaultSleepableOffloadSize = 32768
 
+	// DefaultUprobeHeapSize is the default maximum number of entries in the
+	// process call heap map used by uprobe/usdt sensors.
+	DefaultUprobeHeapSize = 32768
+
```
mtardy, line 71: this is aside but we should have a shared defaults and one for linux and one for windows instead of having duplication of stuff there when it needs to be for both.

mtardy (approved):
lgtm, only nits that I noticed

olsajiri:
@andrewstrohman @kkourt plz take a look

bpf/process/generic_maps.h
```diff
@@ -5,6 +5,60 @@
...
+ */
+#if defined(GENERIC_UPROBE) || defined(GENERIC_URETPROBE) || defined(GENERIC_USDT)
+
+typedef __u64 heap_key_t;
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(map_flags, BPF_F_NO_PREALLOC);
+	__uint(max_entries, 1); // will be resized by agent
+	__type(key, __u64);
```
kkourt: nit: s/heap_key_t/__u64/

bpf/process/bpf_generic_retuprobe.c
```diff
@@ -50,41 +50,41 @@ struct {
 __attribute__((section((MAIN)), used)) int
 BPF_KRETPROBE(generic_retuprobe_event, unsigned long ret)
 {
-	return generic_retprobe(ctx, (struct bpf_map_def *)&retuprobe_calls, ret);
+	return heap_dtor(generic_retprobe(ctx, (struct bpf_map_def *)&retuprobe_calls, ret));
```
kkourt, line 53: AFIACT, the `heap_dtor` should be paired with every `heap_update`, but `heap_dtor` seems to be in the caller and `heap_update` in the callee. Any reason for not having them both in the same function (either the caller or the callee)? For what it's worth I think having them both in the callee is cleaner.

olsajiri: the `heap_update` is called on the entry `generic_uprobe_event` -> `generic_start_process_filter` to add the entry in the `process_call_heap`, then in the generic code we use the heap as needed, but we need to be sure we remove the entry when the program is done, and the program can exit on many places, so I put the dotr calls on each program main function 

as an alternative we could use the preload_cleanup way and add extra uprobe with cleanup like we did for offload.. I haven't tried that but I'm not sure it'd be easier

kkourt: How about doing both in the caller?

For example, we can have a first patch that patches the msg in the callee

```diff
diff --git a/bpf/process/bpf_generic_retuprobe.c b/bpf/process/bpf_generic_retuprobe.c
index 7b6c19d55..ae8806d96 100644
--- a/bpf/process/bpf_generic_retuprobe.c
+++ b/bpf/process/bpf_generic_retuprobe.c
@@ -50,7 +50,14 @@ struct {
 __attribute__((section((MAIN)), used)) int
 BPF_KRETPROBE(generic_retuprobe_event, unsigned long ret)
 {
-       return generic_retprobe(ctx, (struct bpf_map_def *)&retuprobe_calls, ret);
+       int zero = 0;
+       struct msg_generic_kprobe *msg;
+
+       msg = map_lookup_elem(&process_call_heap, &zero);
+       if (!msg)
+               return 0;
+
+       return generic_retprobe(ctx, msg, (struct bpf_map_def *)&retuprobe_calls, ret);
 }

 #ifdef __LARGE_BPF_PROG
diff --git a/bpf/process/generic_calls.h b/bpf/process/generic_calls.h
index bd1e82a35..b78d35742 100644
--- a/bpf/process/generic_calls.h
+++ b/bpf/process/generic_calls.h
@@ -1210,23 +1210,17 @@ generic_output(void *ctx, u8 op)
        return 0;
 }

-FUNC_INLINE int generic_retprobe(void *ctx, struct bpf_map_def *calls, unsigned long ret)
+FUNC_INLINE int generic_retprobe(void *ctx, struct msg_generic_kprobe *e, struct bpf_map_def *calls, unsigned long ret)
 {
        struct execve_map_value *enter;
-       struct msg_generic_kprobe *e;
        struct retprobe_info info;
        struct event_config *config;
        bool walker = false;
-       int zero = 0;
        __u32 ppid;
        long size = 0;
        long ty_arg, do_copy;
        __u64 pid_tgid;

-       e = map_lookup_elem(&process_call_heap, &zero);
-       if (!e)
-               return 0;
-
        e->idx = get_index(ctx);

        config = map_lookup_elem(&config_map, &e->idx);
```

And then:

```diff
diff --git a/bpf/process/bpf_generic_retuprobe.c b/bpf/process/bpf_generic_retuprobe.c
index ae8806d96..761e50623 100644
--- a/bpf/process/bpf_generic_retuprobe.c
+++ b/bpf/process/bpf_generic_retuprobe.c
@@ -50,14 +50,20 @@ struct {
 __attribute__((section((MAIN)), used)) int
 BPF_KRETPROBE(generic_retuprobe_event, unsigned long ret)
 {
-	int zero = 0;
+	int ret;
+	heap_key_t = heap_key();
 	struct msg_generic_kprobe *msg;
 
-	msg = map_lookup_elem(&process_call_heap, &zero);
+	if (!heap_update(key))
+		return 0;
+
+	msg = map_lookup_elem(&process_call_heap, &key);
 	if (!msg)
 		return 0;
 
-	return generic_retprobe(ctx, msg, (struct bpf_map_def *)&retuprobe_calls, ret);
+	ret = generic_retprobe(ctx, msg, (struct bpf_map_def *)&retuprobe_calls, ret);
+	heap_dtor(key);
+	return ret;
 }
 
 #ifdef __LARGE_BPF_PROG

```

kkourt: Having the `heap_update()` and the `heap_dtor()` in the same functions makes the code easier to read IMO.

olsajiri: hum.. but there still need to be other instances of `heap_dtor` in `generic_retuprobe_filter_arg` `generic_retuprobe_actions` `generic_retuprobe_output` ... without paired `heap_update` .. I might be missing something

kkourt: I'm probably the one who's missing something :)

OK, I think I understand now: we need to have a dtor in all the functions in the "tail call table" because any of them can either return (where the dtor is executed) or tail call into something else.

bpf/process/generic_path.h
```diff
@@ -196,7 +197,7 @@ FUNC_INLINE long generic_path_offload(void *ctx, long ty, unsigned long arg,
 	/* initialize for next argument */
 	generic_path_init(e);
 
-	buffer = map_lookup_elem(&buffer_heap_map, &zero);
+	buffer = map_lookup_elem(&buffer_heap_map, &bufidx);
```
andrewstrohman: I think that a uprobe policy with:
```
data:
- index: 0
  source: current_task
  type: file
  resolve: mm.exe_file
```
can reach here. Do we need this heap to be a hash too?

bpf/process/bpf_generic_retuprobe.c
andrewstrohman: > as an alternative we could use the preload_cleanup way and add extra uprobe with cleanup like we did for offload.. 

IMHO this would be better because we could eliminate all the boilerplate calls to heap_dtor(). But, your current approach works, and I don't think this is that big of deal.

pkg/sensors/tracing/uprobe_test.go
```diff
@@ -1392,6 +1392,126 @@ spec:` + opts + `
...
+func TestUprobeProcessCallHeapMapConfigWithMax(t *testing.T) {
+	testUprobeProcessCallHeapMapConfig(t, true)
+}
+
+func testUprobeProcessCallHeapMapConfig(t *testing.T, withMax bool) {
+	if runtime.GOARCH == "arm64" {
+		t.Skip("skipping, x86_64 only test")
+	}
+	if !bpf.HasUprobeMulti() {
+		t.Skip("skipping, no uprobe multi support in kernel; process_call_heap is only wired up for the multi-attach path")
```
andrewstrohman, line 1408: Is there a technical reason behind why single probes do not have a pinned/shared map? Or are you just deferring doing that until someone asks for it?

andrewstrohman (approved):

andrewstrohman:
Do we need to do a similar conversion for the following per-cpu maps?
- buffer_heap_map
- string_maps_heap
- string_prefix_maps_heap
- string_postfix_maps_heap
- ratelimit_heap

olsajiri:
> * string_prefix_maps_heap

yea, good point, I think these are affected as well, we should treat them the same way

also having more cache maps like that it might be better to have the cleanup in separate uprobe like we do for offload, I'll check

thanks

bpf/process/ratelimit_maps.h
```diff
@@ -26,6 +28,9 @@ struct ratelimit_key {
 	__u8 data[MAX_POSSIBLE_ARGS * KEY_BYTES_PER_ARG];
 };
 
+_Static_assert(sizeof(struct ratelimit_key) <= HEAP_RO_SIZE,
```
mtardy: nit: it seems this notation is deprecated and we should use `static_assert` now https://en.cppreference.com/c/language/_Static_assert.

bpf/process/heap.h
```diff
@@ -4,13 +4,14 @@
...
 #define __HEAP_H__
 
-#include "ratelimit_maps.h"
+#define HEAP_RO_SIZE 16384
 
 struct heap_ro_value {
-	union {
-		char string_maps_heap[STRING_MAPS_HEAP_SIZE];
-		char ratelimit_heap[sizeof(struct ratelimit_key) + 128];
-	};
```
mtardy, line 13: I think it was quite elegant but I understand the struggle

mtardy (approved):
lgtm.

It's so useful to have maps tied to process that I wonder if the kernel should provide one like it does for PERCPU. I guess nowadays the idea is no since we can construct it like that but that would be nice to have the lifecycle done for you like the socketmap and stuff.

bpf/process/ratelimit_maps.h
andrewstrohman: Should this check include the "+ 128"?

bpf/process/generic_calls.h
```diff
@@ -43,10 +43,13 @@ generic_start_process_filter(void *ctx, struct bpf_map_def *calls)
 {
 	struct msg_generic_kprobe *msg;
 	struct event_config *config;
+	heap_key_t key = heap_key();
 	struct task_struct *task;
-	int i, zero = 0;
+	int i;
 
-	msg = map_lookup_elem(&process_call_heap, &zero);
+	if (!heap_update(key))
```
andrewstrohman, line 50: This is slightly different behavior compared to the other heap types. Here we unconditionally zero out the heap, regardless of whether the the key exists.

The equivalent for string_maps is `string_maps_heap_get`. But, these _get functions introduced in this PR will only zero out the heap if the key doesn't already exist.

Is this difference in behavior intentional and required ?

It seems like this difference would only matter in conjunction with a bug that prevents the heap's key from being deleted.

andrewstrohman (approved):
LGTM!

bpf/process/ratelimit_maps.h
olsajiri: ok, seems it gets introduced in clang 23, we're now in 22

bpf/process/generic_calls.h
olsajiri: yes, I think the idea was that we almost always will have to update the hash and in the case we already have the entry in, let's zero it... but perhaps we could  check the existing entry and call `errmetrics(EEXIST)` to signal the cleanup bug

bpf/process/ratelimit_maps.h
mtardy: ah this is C23 standard not clang 23, it should be in clang now but anyway that's just a nit

olsajiri: hum strange I got compilation fail for that

mtardy: ah then you are right I guess, I followed this thing but it's unclear with their "yes" thing https://clang.llvm.org/c_status.html. nevermind let's close this!

olsajiri:
fyi after last features there's not much space in filter tailcall, I need to split it first in https://github.com/cilium/tetragon/pull/5647 and base on that, moving to draft so far

mtardy:
> It's so useful to have maps tied to process that I wonder if the kernel should provide one like it does for PERCPU. I guess nowadays the idea is no since we can construct it like that but that would be nice to have the lifecycle done for you like the socketmap and stuff.

btw I realized this exist with `BPF_MAP_TYPE_TASK_STORAGE`, I guess you already thought about it, wouldn't it work for this use case for task related heap?

olsajiri:
> > It's so useful to have maps tied to process that I wonder if the kernel should provide one like it does for PERCPU. I guess nowadays the idea is no since we can construct it like that but that would be nice to have the lifecycle done for you like the socketmap and stuff.
> 
> btw I realized this exist with `BPF_MAP_TYPE_TASK_STORAGE`, I guess you already thought about it, wouldn't it work for this use case for task related heap?

yea, but last time I checked there was size limit of 2 pages for that, I thought we could refactor the event cache to use that, but seemed like bigger project

PR #5508, opened 2026-08-26
fix(strutils): handle empty size values

immanuwell:
### Description

An empty `rb-size` drop-in is a supported way to clear an override. 
Right now it panics in `ParseSize`

Treat empty sizes as zero, which already means use the default

Repro:

```sh
config_dir=$(mktemp -d)
touch "$config_dir/rb-size"
go run ./cmd/tetragon --config-dir "$config_dir"
```

Before this change startup hits a slice bounds panic. 
After it gets past config parsing

Test: `go test ./pkg/strutils`

### Changelog

```release-note
Fix Tetragon startup panic when a ring buffer setting is cleared with an empty value.
```

### DIFF
--- a/pkg/strutils/strutls.go
+++ b/pkg/strutils/strutls.go
@@ -28,6 +28,10 @@ func UTF8FromBPFBytes(b []byte) string {
 }
 
 func ParseSize(str string) (int, error) {
+	if str == "" {
+		return 0, nil
+	}
+
 	suffix := str[len(str)-1:]
 
 	if !strings.Contains("KMG", suffix) {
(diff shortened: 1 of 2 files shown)


immanuwell:
Could a maintainer add the `release-note/bug` label please? The label check requires it, but contributors cannot apply labels here.

mtardy (approved):
sure thanks

FedeDP (approved):

FedeDP:
Thank you so much! Nice fix 😃

olsajiri (approved):

PR #5509, opened 2026-08-27
docs: fix argIndex range

andrewstrohman:
We can reach the first five arguments of function. The 5th arg's index is 4, because we start counting at 0.

Also, we shouldn't use EventConfigMaxArgs to explain the maximum reachable argument in a function signature. These concepts are treated separately in the bpf code. See [here](https://github.com/cilium/tetragon/blob/76f209584d7ece0abdec2dd99f9730122693c549/bpf/lib/generic.h#L28-L34).

### DIFF
--- a/pkg/sensors/tracing/genericuprobe.go
+++ b/pkg/sensors/tracing/genericuprobe.go
@@ -601,7 +601,7 @@ func validateUprobeSpec(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) err
 			if action.Action != "Set" {
 				continue
 			}
-			if action.ArgIndex >= api.EventConfigMaxArgs {
+			if action.ArgIndex >= api.MaxAccessibleArgs {
 				return fmt.Errorf("uprobe Set action argIndex %d out of range", action.ArgIndex)
 			}
 		}
@@ -1175,7 +1175,7 @@ func getUprobeArgConfig(spec *v1alpha1.UProbeSpec, has *uprobeHas) (uprobeArgCon
 			cfg.argRetprobe = &spec.Args[i]
 			cfg.argRetprobeIdx = i
 		}
-		if a.Index > 4 {
+		if a.Index >= api.MaxAccessibleArgs {
 			return fmt.Errorf("error add arg: ArgType %s Index %d out of bounds",
 				a.Type, int(a.Index))
 		}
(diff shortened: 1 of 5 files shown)


pkg/sensors/tracing/genericuprobe.go
```diff
@@ -601,7 +601,7 @@ func validateUprobeSpec(spec *v1alpha1.UProbeSpec, state *uprobeConfigState) err
 			if action.Action != "Set" {
 				continue
 			}
-			if action.ArgIndex >= api.EventConfigMaxArgs {
+			if action.ArgIndex > 4 {
```
FedeDP: Thank you very much Andy!
Question: isn't there a proper constant to be used here?

FedeDP (approved):

pkg/sensors/tracing/genericuprobe.go
mtardy: yeah maybe it's worth adding another const in the API

andrewstrohman: I didn't see one, but I added one.

andrewstrohman:
This made me find some args bounds checking issues. I'll mark [this PR ](https://github.com/cilium/tetragon/pull/5514) ready for review once this one is merged.

FedeDP (commented):
Nice thanks!
While at it, can you also fix https://github.com/cilium/tetragon/blob/main/pkg/selectors/kernel_regs_amd64.go#L57 (and same for arm64 version)?
The `case 5` should not be there :)

mtardy (approved):

andrewstrohman:
> Nice thanks!
> While at it, can you also fix https://github.com/cilium/tetragon/blob/main/pkg/selectors/kernel_regs_amd64.go#L57 (and same for arm64 version)?
> The `case 5` should not be there :)

Mahe merged before I could get around to doing this.

When I looked into this, I found that my assessment here was wrong (It's embarrassing). While we are limited to only reading the first five arguments of the hook's signature (because we only have a0 - a5 in `struct msg_generic_kprobe`), this does not limit what we can set.

The limitation for what we can set is about the calling convention, specifically which args are passed via register vs stack. So, the limit for x86_64 is 6 and the limit for arm64 is 8.

Further, I found that golang has a different calling convention, so we cannot set arguments for that language.

I'm trying to rectify this [here](https://github.com/cilium/tetragon/pull/5514). I tested that I could overwrite the 6th arg on amd64 with that change. I am concerned that if we allow setting args that cannot be read, this could be confusing to users. So, I'm not sure if the benefit of allowing this justifies the possible confusion. Please let me know what you think in that PR.

FedeDP:
> The limitation for what we can set is about the calling convention, specifically which args are passed via register vs stack. So, the limit for x86_64 is 6 and the limit for arm64 is 8.

That makes sense.

PR #5514, opened 2026-08-27
Fix args bounds checks

andrewstrohman:
Fix args bound checks for tracepoint, USDT and urprobe `Set` action. See individual commit messages for more details.

The second to last commit allows us to set uprobe args beyond the args that we can read. I'm not sure if the usefulness of this outweighs the potential for user confusion.


FedeDP (approved):
LGTM, thank you so much!

mtardy (approved):
lgtm, thanks for the fixes

PR #5517, opened 2026-08-28
flags: add `--verifier-log-level` for Go tests and replace previous name

mtardy:
When debugging an issue that arise in a Go test, it's useful to have a way of seeing verifier logs directly in the test failure, wire our existing flag into TestSensorRun().

The previous name was confusing: I always mixed it between --debug and --verbose and verbose was missing the idea that it was only for the verifier log and not for tetragon level logs.

### DIFF
--- a/pkg/option/flags.go
+++ b/pkg/option/flags.go
@@ -31,6 +31,7 @@ const (
 	KeyBTF                    = "btf"
 	KeyProcFS                 = "procfs"
 	KeyKernelVersion          = "kernel"
+	KeyVerifierLogLevel       = "verifier-log-level"
 	KeyVerbosity              = "verbose"
 	KeyProcessCacheSize       = "process-cache-size"
 	KeyDisableProcessCache    = "disable-process-cache"
@@ -182,7 +183,10 @@ func ReadAndSetFlags() error {
 	Config.BTF = viper.GetString(KeyBTF)
 	Config.ProcFS = viper.GetString(KeyProcFS)
 	Config.KernelVersion = viper.GetString(KeyKernelVersion)
-	Config.Verbosity = viper.GetInt(KeyVerbosity)
+	Config.VerifierLogLevel = viper.GetInt(KeyVerifierLogLevel)
+	if !viper.IsSet(KeyVerifierLogLevel) && viper.IsSet(KeyVerbosity) {
+		Config.VerifierLogLevel = viper.GetInt(KeyVerbosity)
+	}
 	Config.ForceSmallProgs = viper.GetBool(KeyForceSmallProgs)
 	Config.ForceLargeProgs = viper.GetBool(KeyForceLargeProgs)
 	Config.Debug = viper.GetBool(KeyDebug)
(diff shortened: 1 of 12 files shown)


pkg/option/flags.go
```diff
@@ -182,7 +183,10 @@ func ReadAndSetFlags() error {
 	Config.BTF = viper.GetString(KeyBTF)
 	Config.ProcFS = viper.GetString(KeyProcFS)
 	Config.KernelVersion = viper.GetString(KeyKernelVersion)
-	Config.Verbosity = viper.GetInt(KeyVerbosity)
+	Config.Verbosity = viper.GetInt(KeyVerifierLogLevel)
+	if !viper.IsSet(KeyVerifierLogLevel) && viper.IsSet(KeyVerbosity) {
+		Config.Verbosity = viper.GetInt(KeyVerbosity)
+	}
 	Config.ForceSmallProgs = viper.GetBool(KeyForceSmallProgs)
 	Config.ForceLargeProgs = viper.GetBool(KeyForceLargeProgs)
```
FedeDP, line 195: Can this even happen considering the 
```go
flags.Func(KeyVerbosity, "deprecated alias for --"+KeyVerifierLogLevel, func(value string) error {
		return flags.Set(KeyVerifierLogLevel, value)
	})
```
code below?
I mean, if `KeyVerbosity` gets set, we force-set the new `KeyVerifierLogLevel` key.

pkg/option/flags.go
```diff
@@ -182,7 +183,10 @@ func ReadAndSetFlags() error {
 	Config.BTF = viper.GetString(KeyBTF)
 	Config.ProcFS = viper.GetString(KeyProcFS)
 	Config.KernelVersion = viper.GetString(KeyKernelVersion)
-	Config.Verbosity = viper.GetInt(KeyVerbosity)
+	Config.Verbosity = viper.GetInt(KeyVerifierLogLevel)
```
FedeDP: Do we want to also rename `Config.Verbosity` to something more meaningful? Eg: `Config.VerifierVerbosity`

mtardy: yeah I can rename the config to VerifierLogLevel also

pkg/option/flags.go
mtardy: yeah let's remove one but not keep both, I didn't know the usage of flags.Func for deprecated field so far!

andrewstrohman: My vote is to handle it with:

```
diff --git a/pkg/option/flags.go b/pkg/option/flags.go
index 230f8b729..78d762cb1 100644
--- a/pkg/option/flags.go
+++ b/pkg/option/flags.go
@@ -514,9 +514,7 @@ func AddFlags(flags *pflag.FlagSet) {
        flags.String(KeyProcFS, "/proc/", "Location of procfs to consume existing PIDs")
        flags.String(KeyKernelVersion, "", "Kernel version")
        flags.Int(KeyVerifierLogLevel, 0, VerifierLogLevelHelp)
-       flags.Func(KeyVerbosity, "deprecated alias for --"+KeyVerifierLogLevel, func(value string) error {
-               return flags.Set(KeyVerifierLogLevel, value)
-       })
+       flags.Int(KeyVerbosity, 0, "deprecated alias for --"+KeyVerifierLogLevel)
        flags.MarkDeprecated(KeyVerbosity, "use --"+KeyVerifierLogLevel+" instead")
        flags.Int(KeyProcessCacheSize, 65536, "Size of the process cache")
        flags.Bool(KeyDisableProcessCache, false, "Disable process cache")
```
So that the outcome is not dependent on the ordering of supplied arguments.

andrewstrohman (commented):
LGTM, except for the flag overwrite issue that @FedeDP spotted.

FedeDP (approved):
Thanks!

PR #5520, opened 2026-08-28
perf(pkg/observer): reuse buffers when reading events

rafaelroquetto:
### Description

The existing `Read()` call allocates a buffer on each event read. This event is passed to a downstream sink (receiveEvent()) which never retains the event or parts thereof, but rather uses it to materialise higher-level objects. This means that, for each event received, we have:
```
  event alloc -> event is read -> event is freed
```
which causes significant allocation churn.

Instead, use a combination of `ReadInto()` and `sync.Pool`, so that the event buffers can be reused, i.e.:
```
  event buffer acquire -> event is read -> event buffer returns to the pool
```
Validated with a soak test hammering tetragon with millions of uniquely-marked `execve` events on a controlled cgroup, showing that allocations attributed to 'readRecord()' are almost entirely eliminated.

For example, in the longest run (~73 mins, 6.3M events):

|  | Before | After |
|---|---:|---:|
| total allocated space | 79.10 GB | 73.94 GB |
| total allocation objects | 1.15B | 1.11B |
| **`ringReader.readRecord` allocated space** | **4.86 GB** | **~54 MB** |

This corresponds to a ~98.9% reduction in allocation volume at `ringReader.readRecord`. The same profile showed approximately 19M fewer allocations directly attributed to this function.

The reduction was reproduced across independent runs at different event volumes.

No measurable cpu regression was observed (cpu profiles varied in both directions around the baseline)

The allocated-space figures above are cumulative allocations, rather than resident memory. The purpose of this change is to reduce short-lived allocation churn in the event reading path, not Tetragon's steady-state memory footprint. RSS measurements showed no evidence of increased memory retention from the use of `sync.Pool`.

### Changelog

```release-note
perf: reduce allocation overhead in the bpf event read path
```

### DIFF
--- a/pkg/observer/observer_linux.go
+++ b/pkg/observer/observer_linux.go
@@ -26,6 +26,13 @@ const (
 	perCPUBufferBytes = 65535
 )
 
+var rawSampleBufPool = sync.Pool{
+	New: func() any {
+		buf := make([]byte, 0)
+		return &buf
+	},
+}
+
 func (k *Observer) getRBSize(cpus int) int {
 	var size int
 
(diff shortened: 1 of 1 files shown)


andrewstrohman (approved):
Nice!

tpapagian (approved):

Issue #4661, opened 2026-02-16
celbpf: evalute CEL expressions in BPF


kkourt:
PRs:
 - https://github.com/cilium/tetragon/pull/4504

TODO:
- [ ]  docs
- [ ] support CEL expressions in tracepoints, uprobes, and USDTs (and multi-kprobe)
- [ ] more arithmetic operations: mult, div, shift
- [ ] fuzz testing over expressions
- [ ] helper functions such as ntoh
- [ ]  support .data (in addition to .args)
- [ ] optimize code generation (either by using register-based code generation, or via peephole
- [ ] optimizations over the stack generated code)
- [ ] support member access (e.g., arg0.foo)

PR #5523, opened 2026-08-28
cel: add 16 and 8 bit (u)int support

3u13r:
Part of https://github.com/cilium/tetragon/issues/4661#issue-3947169252

### Description

Add support for 16 and 8 bit (u)int values and arithmetic operations. This is a bit more complicated, since for 64 and 32 bit we can just choose the correct ALU/ALU64 and eBPF takes care of the rest.

So we need to emulate certain operations by adding some conversion eBPF instructions.

### Changelog

```release-note
cel: add 16 and 8 bit (u)int support
```

### DIFF
--- a/docs/content/en/docs/concepts/tracing-policy/selectors.md
+++ b/docs/content/en/docs/concepts/tracing-policy/selectors.md
@@ -913,6 +913,14 @@ Currently, `MatchCEL` supports:
 * Integer casting to 32-bits (`int32()`, `uint32()`)
 * Bitwise operations (`and()`, `or()`, `xor()`, `not()`, `lsh()`, `rsh()`)
 
+Shift operations (`lsh()`, `rsh()`) mask the shift amount to the bit width of
+the operand. In other words, the shift amount is reduced modulo the bit width
+of the operand.
+
+For example, `lsh(uint8(16u), uint8(9u))` is equal to `lsh(uint8(16u),
+uint8(1u))` because the shift amount is masked by 7 (the bit width of a `uint8`
+minus 1) thus `9 & 7 = 1` or reduced modulo the bit width, thus `9 % 8 = 1`.
+
 ## Actions filter
 
 Actions filters are a list of actions that execute when an appropriate selector
--- a/pkg/celbpf/cel_eval_test.go
+++ b/pkg/celbpf/cel_eval_test.go
@@ -21,7 +21,11 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-type celNumber[T int32 | uint32] struct {
+type number interface {
+	int8 | uint8 | int16 | uint16 | int32 | uint32
+}
+
+type celNumber[T number] struct {
 	value    T
 	typeInfo *celTypes.Type
 }
@@ -60,8 +64,28 @@ func (value celNumber[T]) Value() any {
 	return value.value
 }
 
+type celS8 = celNumber[int8]
+type celU8 = celNumber[uint8]
+type celS16 = celNumber[int16]
+type celU16 = celNumber[uint16]
 type celS32 = celNumber[int32]
 type celU32 = celNumber[uint32]
+
+func newCelS8(value int8) celS8 {
+	return celS8{value: value, typeInfo: s8Ty}
+}
+
+func newCelU8(value uint8) celU8 {
+	return celU8{value: value, typeInfo: u8Ty}
+}
+
+func newCelS16(value int16) celS16 {
+	return celS16{value: value, typeInfo: s16Ty}
+}
+
+func newCelU16(value uint16) celU16 {
+	return celU16{value: value, typeInfo: u16Ty}
+}
 
 func newCelS32(value int32) celS32 {
 	return celS32{value: value, typeInfo: s32Ty}
@@ -77,6 +101,14 @@ type celTestTypeAdapter struct {
 
 func (adapter celTestTypeAdapter) NativeToValue(value any) ref.Val {
 	switch value := value.(type) {
+	case celS8:
+		return value
+	case celU8:
+		return value
+	case celS16:
+		return value
+	case celU16:
+		return value
 	case celS32:
 		return value
 	case celU32:
@@ -101,6 +133,10 @@ func celBinArithOp(
 	uintOp func(celTypes.Uint, celTypes.Uint) celTypes.Uint,
 	s32Op func(int32, int32) int32,
 	u32Op func(uint32, uint32) uint32,
+	s16Op func(int16, int16) int16,
+	u16Op func(uint16, uint16) uint16,
+	s8Op func(int8, int8) int8,
+	u8Op func(uint8, uint8) uint8,
 ) func(left, right ref.Val) ref.Val {
 	return func(left, right ref.Val) ref.Val {
 		switch left := left.(type) {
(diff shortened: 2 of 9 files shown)


pkg/celbpf/compiler.go
```diff
@@ -197,14 +194,8 @@ func (c *compiler) compileCall(expr cgAst.Expr) error {
 		i := len(callArgs) - j - 1
 		ty := argTypes[i]
 		switch ty.TypeName() {
-		case "int", "uint":
+		case "bool", "int", "uint", "s32", "u32", "u16", "s16", "u8", "s8":
```
FedeDP, line 213: I guess `16` and `8` should appear in the next commit, not this one, right?

pkg/celbpf/cel_eval_test.go
```diff
@@ -21,7 +21,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-type celNumber[T int32 | uint32] struct {
+type celNumber[T int8 | uint8 | int16 | uint16 | int32 | uint32] struct {
```
FedeDP: Can we declare an interface on top of `T int8 | uint8 | int16 | uint16 | int32 | uint32` so that we can use that instead of specifying it once again in `compareIntegers`?

FedeDP (commented):
LGTM, left a couple of comments.

pkg/celbpf/compiler.go
3u13r: Yes, split the change in 2.

pkg/celbpf/cel_eval_test.go
3u13r: Sounds good, especially since that new interface doesn't need to be exported. Done.

FedeDP (approved):
Changes LGTM, but i'd wait upon @kkourt review!

kkourt (changes requested):
Thanks Leonard!

The first patch LGTM.

Can you split the second patch to ease reviewing?
From a quick look, it seems that it can be split into:
1.  Core changes
2. Tests
3. Changes in the CEL user-space evaluation for testing for randomized testing

kkourt (approved):

pkg/celbpf/compiler.go
```diff
@@ -78,8 +78,7 @@ func (c *compiler) compileCall(expr cgAst.Expr) error {
 		}
 	case uint32Fn:
 		emitCall = func() error {
-			c.cg.emitU32(scratchRegs[0], argTypes[0])
-			return nil
+			return c.cg.emitU32(scratchRegs[0], argTypes[0])
```
mtardy, line 81: is this an unrelated change in this commit?

pkg/celbpf/codegen.go
```diff
@@ -59,11 +59,9 @@ func (g *codeGenerator) emitPushBool(val bool, tmp asm.Register) {
...
 }
 
-func (g *codeGenerator) emitPopBool(reg asm.Register) {
-	g.emitRaw(asm.LoadMem(reg, asm.R10, g.stackTop, asm.DWord))
-	g.stackTop += 8
-}
-
+// emitPushInt64 pushes the 64-bit register onto the stack.
+// This is used for all values, even ones with smaller widths,
+// to keep the stack aligned to 8 bytes for simplicity.
```
mtardy, line 64: In the commit logs I read
> But since we want to keep the stack 8 byte aligned

Can you explain why? I miss the context here

pkg/celbpf/codegen.go
```diff
@@ -107,42 +107,132 @@ func (g *codeGenerator) emitU32(reg asm.Register, regTy *cgTypes.Type) error {
...
+	g.emitRaw(
+		asm.LSh.Imm(reg, 48),
+		asm.ArSh.Imm(reg, 48),
+		asm.StoreMem(asm.R10, g.stackTop, reg, asm.DWord),
+	)
+	return nil
+}
+
+func (g *codeGenerator) emitU16(reg asm.Register, regTy *cgTypes.Type) error {
+
```
mtardy: nit: whiteline

also in emitU8 (but not in the other oens)

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -913,6 +913,8 @@ Currently, `MatchCEL` supports:
 * Integer casting to 32-bits (`int32()`, `uint32()`)
 * Bitwise operations (`and()`, `or()`, `xor()`, `not()`, `lsh()`, `rsh()`)
 
+Shift operations (`lsh()`, `rsh()`) mask the shift amount by the bit width of the operand.
```
mtardy: I think you might want to be a bit more explicit for user reading this without this PR's context. Maybe an example?

mtardy (approved):
Thanks, a few comments (especially on docs since I was pulled for that I imagine), nothing blocking!

pkg/celbpf/codegen.go
3u13r: Added this to the commit message:
```
To quote commit a52c705f08f0:
"To keep things simple, the 32-bit types are saved as 64-bit values in
the stack."
```

If I understand eBPF's alignment constraints from [this kernel selftest](https://github.com/torvalds/linux/blob/08df884136f1c1197bab2a27814404fd329d9aac/tools/testing/selftests/bpf/progs/verifier_xadd.c#L16), then we cannot store a single 8 byte value next to a single 4 byte value on the stack.

So not aligning everything on 8 bytes would mean that we need to do alignment calculations which would be more complex.

pkg/celbpf/codegen.go
3u13r: Fixed it in all lines that I touched. Leaving the newline in `emitU32` for the new linter.

3u13r:
Re-requested mtardy. Not sure if that's the correct flow, but I made changes according to your suggestions. Let me know if you prefer not re-reviewing in such cases.

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -913,6 +913,11 @@ Currently, `MatchCEL` supports:
 * Integer casting to 32-bits (`int32()`, `uint32()`)
 * Bitwise operations (`and()`, `or()`, `xor()`, `not()`, `lsh()`, `rsh()`)
 
+Shift operations (`lsh()`, `rsh()`) mask the shift amount to the bit width of the operand.
+So `lsh(uint8(16u), uint8(9u))` is equal to `lsh(uint8(16u), uint8(1u))`
+because the shift amount is masked by 7 (the bit width of a
+`uint8` minus 1).
```
mtardy: wdyt?
```suggestion
Shift operations (`lsh()`, `rsh()`) mask the shift amount to the bit width of
the operand. In other words, the shift amount is reduced modulo the bit width
of the operand.

For example, `lsh(uint8(16u), uint8(9u))` is equal to `lsh(uint8(16u),
uint8(1u))` because the shift amount is masked by 7 (the bit width of a `uint8`
minus 1) thus `9 & 7 = 1` or reduced modulo the bit width, thus `9 % 8 = 1`.
```

mtardy (approved):
thanks I think the changes helped! :) 

Just one last suggestion for the docs, tell me what you think.

docs/content/en/docs/concepts/tracing-policy/selectors.md
3u13r: Neat, when I tried would couldn't come up with explaining it both in terms of masking and modulo. I like this.

mtardy: nice!! let's merge this then when it's green, ping me if I forget

PR #5524, opened 2026-08-29
metrics: standardize common labels

### LINKED ISSUE #2786: Standardize common metrics labels
There are labels shared by multiple metrics, but using different names. Using the same name across metrics would make metrics more intuitive and easier to correlate.

Such shared labels should be defined in a shared library (`pkg/metrics` most likely), so that they can be easily reused.

Any changes made to the existing metrics must be documented in upgrade notes.

- [ ] Use `opcode` for [ops.OpCode](https://github.com/cilium/tetragon/blob/2fc399fb69a51759a60be1dceaa382febf08da73/pkg/api/ops/ops.go#L68) (rather than `msg_op` or `op`)
- [ ] Add `op` label in addition to `opcode` with a human-readable string
- [ ] Use `event_type` for [EventType](https://github.com/cilium/tetragon/blob/2fc399fb69a51759a60be1dceaa382febf08da73/api/v1/tetragon/events.proto#L19) (rather than `event` or `type`)

AritraDey-Dev:
### Description

This PR standardizes metrics labels on `opcode` for the numeric ops.OpCode, add `opstr` for its human-readable name, and use `event_type` for EventType(see commits). Refs #2786.

**1. opcode for the numeric ops.OpCode (was msg_op in two metrics, op in another)**

**Before:**
```
tetragon_msg_op_total{msg_op="5"} 1
tetragon_bpf_missed_events_total{error="EBUSY",msg_op="5"} 3
tetragon_handling_latency_sum{op="5"} 87
```
**After:**
```
tetragon_msg_op_total{opcode="5",opstr="Execve"} 1
tetragon_bpf_missed_events_total{error="EBUSY",opcode="5",opstr="Execve"} 3
tetragon_handling_latency_sum{opcode="5",opstr="Execve"} 87
```

**2. opstr added with the human-readable name**

**Before:**
```
tetragon_msg_op_total{msg_op="13"} 1
tetragon_handler_errors_total{error="event_handler_failed",opcode="5"} 1
tetragon_data_event_size_sum{op="ok"} 1500
```
**After:**
```
tetragon_msg_op_total{opcode="13",opstr="GenericKprobe"} 1
tetragon_handler_errors_total{error="event_handler_failed",opcode="5",opstr="Execve"} 1
tetragon_data_event_size_sum{status="ok"} 1500
```
data_event_size: op -> status. Its value is ok/bad, so op was a misleading name for it.

**3. event_type for EventType**

**Before:**

`tetragon_events_total{binary="",namespace="",node_name="",pod="",type="PROCESS_EXEC",workload=""} 1`
**After:**
`tetragon_events_total{binary="",event_type="PROCESS_EXEC",namespace="",node_name="",pod="",workload=""} 1`

### Changelog

```release-note
standardize metrics labels: `opcode` for the numeric ops.OpCode, new `opstr` for its readable name, `event_type` for EventType.
```

### DIFF
--- a/contrib/upgrade-notes/latest.md
+++ b/contrib/upgrade-notes/latest.md
@@ -37,4 +37,26 @@ Depending on your setup, changes listed here might require a manual intervention
 
 ### Metrics
 
-* TBD
+Labels holding a message opcode are now consistently named `opcode` (the numeric
+`ops.OpCode`), and every metric using one also exposes a human-readable `opstr`
+label. Update any dashboards, alerts or recording rules that select on the old
+label names. For example:
+
+```
+- tetragon_msg_op_total{msg_op="5"} 1
++ tetragon_msg_op_total{opcode="5",opstr="Execve"} 1
+```
+
+* `tetragon_msg_op_total`: the `msg_op` label is renamed to `opcode`, and a new
+  `opstr` label with the human-readable opcode name is added.
+* `tetragon_bpf_missed_events_total`: the `msg_op` label is renamed to `opcode`,
+  and a new `opstr` label with the human-readable opcode name is added.
+* `tetragon_handling_latency`: the numeric opcode is now reported in a new
+  `opcode` label, and the human-readable name in a new `opstr` label. The old
+  `op` label is removed, so queries matching on it need updating.
+* `tetragon_handler_errors_total`: a new `opstr` label with the human-readable
+  opcode name is added, alongside the existing `opcode` label.
+* `tetragon_data_event_size`: the `op` label is renamed to `status`. It reports
+  whether handling a data event succeeded (`ok`/`bad`) and is not an opcode.
+* `tetragon_events_total`: the `type` label is renamed to `event_type`, matching
+  the label already used for the event type by other metrics.
--- a/pkg/metrics/errormetrics/errormetrics.go
+++ b/pkg/metrics/errormetrics/errormetrics.go
@@ -65,20 +65,6 @@ var (
 		Name:   "error",
 		Values: slices.Collect(maps.Values(errorTypeLabelValues)),
 	}
-	// Constrained label for opcode (numeric strings)
-	opcodeLabel = metrics.ConstrainedLabel{
-		Name: "opcode",
-		Values: func() []string {
-			res := make([]string, 0, len(ops.OpCodeStrings))
-			for opcode := range ops.OpCodeStrings {
-				if opcode != ops.MSG_OP_TEST {
-					// include UNDEF (0) to represent unknown opcodes in docs/metrics
-					res = append(res, strconv.Itoa(int(int32(opcode))))
-				}
-			}
-			return res
-		}(),
-	}
 	// Constrained label for handler error type
 	handlerErrTypeLabel = metrics.ConstrainedLabel{
 		Name:   "error",
@@ -98,7 +84,11 @@ var (
 		metrics.NewOpts(
 			consts.MetricsNamespace, "", "handler_errors_total",
 			"The total number of event handler errors. For internal use only.",
-			nil, []metrics.ConstrainedLabel{opcodeLabel, handlerErrTypeLabel}, nil,
+			nil, []metrics.ConstrainedLabel{
+				metrics.OpCodeLabelWithUndef,
+				metrics.OpCodeNameLabelWithUndef,
+				handlerErrTypeLabel,
+			}, nil,
 		),
 		nil,
 	)
@@ -142,7 +132,7 @@ func ErrorTotalInc(er ErrorType) {
 
 // Get a new handle on the HandlerErrors metric
 func GetHandlerErrors(opcode ops.OpCode, er EventHandlerError) prometheus.Counter {
-	return HandlerErrors.WithLabelValues(strconv.Itoa(int(int32(opcode))), er.String())
+	return HandlerErrors.WithLabelValues(strconv.Itoa(int(opcode)), opcode.String(), er.String())
 }
 
 // Increment the HandlerErrors metric
--- a/pkg/metrics/eventmetrics/bpfcollector.go
+++ b/pkg/metrics/eventmetrics/bpfcollector.go
@@ -10,6 +10,7 @@ import (
 	"github.com/cilium/ebpf"
 	"github.com/prometheus/client_golang/prometheus"
 
+	"github.com/cilium/tetragon/pkg/api/ops"
 	"github.com/cilium/tetragon/pkg/api/processapi"
 	"github.com/cilium/tetragon/pkg/metrics"
 	"github.com/cilium/tetragon/pkg/option"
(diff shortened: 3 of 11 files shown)


pkg/metrics/labels.go
```diff
@@ -52,27 +53,72 @@ func promContainsLabel(labels prometheus.ConstrainedLabels, label string) bool {
...
 	OpCodeLabel = ConstrainedLabel{
-		Name: "msg_op",
+		Name: "opcode",
 		// These are numbers, not human-readable names.
-		Values: getOpcodes(),
+		Values: getOpcodes(false),
+	}
+	// OpCodeNameLabel is the human-readable counterpart of OpCodeLabel.
+	OpCodeNameLabel = ConstrainedLabel{
+		Name:   "op",
```
FedeDP: I'd argue `opstr` is a better name? Ie: something that let user knows this is a string representation.
So we'll have `opcode` and `opstr`.

FedeDP (commented):
I like the changes, thank you so much!
Only thing i'd change is a better name for `op`; i propose `opstr` so that we end up with:
`opcode` and `opstr`.
WDYT?

pkg/metrics/labels.go
AritraDey-Dev: Makes sense! I don't have a better suggestion either, and opstr is much more understandable here. Will update.

FedeDP (approved):
lgtm, thanks!

pkg/metrics/labels.go
```diff
@@ -52,27 +53,72 @@ func promContainsLabel(labels prometheus.ConstrainedLabels, label string) bool {
...
+	}
+	slices.Sort(result)
+	return result
+}
+
+func getOpcodes(withUndef bool) []string {
+	opcodes := sortedOpCodes(withUndef)
+	result := make([]string, len(opcodes))
+	for i, opcode := range opcodes {
+		result[i] = strconv.Itoa(int(int32(opcode)))
```
mtardy, line 112: why the double casting?

pkg/metrics/labels.go
```diff
@@ -52,27 +53,72 @@ func promContainsLabel(labels prometheus.ConstrainedLabels, label string) bool {
...
+	OpCodeLabelWithUndef = ConstrainedLabel{
+		Name:   "opcode",
+		Values: getOpcodes(true),
+	}
+	// OpCodeNameLabelWithUndef is the human-readable counterpart of
+	// OpCodeLabelWithUndef.
+	OpCodeNameLabelWithUndef = ConstrainedLabel{
+		Name:   "opstr",
+		Values: getOpcodeNames(true),
 	}
```
mtardy, line 79: can you explain why we can't just use the version with undef for everything?

pkg/metrics/opcodemetrics/opcodemetrics.go
```diff
@@ -64,14 +46,25 @@ func InitMetrics() {
 	for opcode := range ops.OpCodeStrings {
 		if opcode != ops.MSG_OP_UNDEF && opcode != ops.MSG_OP_TEST {
 			GetOpTotal(opcode).Add(0)
-			LatencyStats.WithLabelValues(strconv.Itoa(int(int32(opcode))))
```
mtardy, line 67: I see it was copied from somewhere, I don't think it's necessary, these opcode are quite limited anyway nop?

pkg/metrics/errormetrics/errormetrics.go
```diff
@@ -65,20 +65,6 @@ var (
...
-			res := make([]string, 0, len(ops.OpCodeStrings))
-			for opcode := range ops.OpCodeStrings {
-				if opcode != ops.MSG_OP_TEST {
-					// include UNDEF (0) to represent unknown opcodes in docs/metrics
-					res = append(res, strconv.Itoa(int(int32(opcode))))
-				}
-			}
-			return res
-		}(),
-	}
```
mtardy, line 81: nice

pkg/metrics/eventmetrics/bpfcollector.go
```diff
@@ -50,16 +51,18 @@ func collect(ch chan<- prometheus.Metric) {
 	for opcode, errors := range sum.SentFailed {
 		for er, count := range errors {
 			if count > 0 {
-				ch <- MissedEvents.MustMetric(float64(count), strconv.Itoa(opcode), perfEventErrors[er])
+				ch <- MissedEvents.MustMetric(float64(count),
+					strconv.Itoa(opcode), ops.OpCode(opcode).String(), perfEventErrors[er])
```
mtardy, line 55: nit: wrapping is a bit strange, is it gofmt?

pkg/observer/data_stats.go
```diff
@@ -30,13 +30,15 @@ func (e DataEventOp) String() string {
...
-	// Constrained labels for event and op
+	// Constrained labels for event and status
 	eventLabel = metrics.ConstrainedLabel{
 		Name:   "event",
 		Values: slices.Collect(maps.Values(DataEventTypeStrings)),
 	}
-	opLabel = metrics.ConstrainedLabel{
-		Name:   "op",
+	// NB: this reports whether handling a data event succeeded, it is not an
+	// operation, so "op" was a misleading name for it.
```
mtardy, line 39: I think it's ok to remove that comment since we have the git history

contrib/upgrade-notes/latest.md
```diff
@@ -37,4 +37,21 @@ Depending on your setup, changes listed here might require a manual intervention
 
 ### Metrics
 
-* TBD
+Labels holding a message opcode are now consistently named `opcode` (the numeric
+`ops.OpCode`), and every metric using one also exposes a human-readable `opstr`
+label. Update any dashboards, alerts or recording rules that select on the old
+label names.
```
mtardy, line 43: nit: could be nice to put one line of example that shows the before after so that it's easy to understand without reading the details.

mtardy (commented):
Thanks!! Looks good overall, just a few small comments.

pkg/metrics/labels.go
AritraDey-Dev: Leftover from the old getOpcodes() - OpCode is an int, so the int32 hop does nothing. dropping it

pkg/metrics/labels.go
AritraDey-Dev: Only handler_errors_total can emit `MSG_OP_UNDEF` (its `unknown_opcode` case), so using the undef variant everywhere would advertise 0/Undef in the docs for metrics that can't produce it. That split already existed, errormetrics had its own label including UNDEF, metrics.OpCodeLabel excluded it. I just kept it when consolidating.

Do you think that doc precision is worth two extra vars and the withUndef param, or would you rather I unify them?

pkg/metrics/opcodemetrics/opcodemetrics.go
AritraDey-Dev: That line is gone now, InitMetrics calls GetLatencyStats() instead. Dropped the same cast from opcodeLabelValues and GetHandlerErrors too.

pkg/metrics/eventmetrics/bpfcollector.go
AritraDey-Dev: Not gofmt, that was me, putting it back on one line.

pkg/metrics/labels.go
mtardy: no no that's perfectly fine, I was mostly wondering because I'm not super familiar with this, thanks for the explanations.

mtardy (approved):
cool thanks

PR #5525, opened 2026-08-30
WinCi refactor and  fix negative WaitGroup counter panic on Windows

ygvalent:
1. Refactored WinCI for stability.

Architecture: smoke-test.ps1 drives a run, harness.ps1 holds everything independent of which features are built, and one module per feature under modules/ contributes the Tetragon flags it needs, the log markers to wait for, its probes and its checks. Modules are discovered by listing that directory. smoke-test.ps1, harness.ps1 and modules/10-process.ps1 are identical to their Tetragon  counterparts.

3. Fixed race in Tetragon shutdown flow on Windows

RunEvents called wg.Add(1) once, but two goroutines deferred wg.Done(): the ring-buffer reader and an empty goroutine above it. The counter hit zero immediately, making defer wg.Wait() a no-op, and the reader's Done() then panicked with "sync: negative WaitGroup counter" when its loop exited. The loop exited only when Read() returned, and nothing closed ringBufReader, so this depended on the deferred ringBufMap.Close() invalidating the map before the process exited. On panic Tetragon died mid-unload with status 2, leaving BPF programs pinned and a stale tetragon.pid; the next start then refused until the eBPF services were cycled.

Fix: use wg.Go for the reader, and close the reader after the stop context fires. The close interrupts the blocking Read, so the deferred wg.Wait returns. Linux closes its readers the same way. Verified on Windows with the OSS binary, where the race window is narrow: widened by a 3s sleep after RunEvents returns, the old code panicked 3/3 with status 2 and one pinned program each time, the new code exited 0 and left nothing 3/3. Without the sleep, six old runs and three new ones all looked clean.

### DIFF
--- a/.github/actions/download-artifact/action.yml
+++ b/.github/actions/download-artifact/action.yml
@@ -50,7 +50,7 @@ runs:
 
     - name: "Download artifact"
       id: download
-      uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6.0.0
+      uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
       with:
         name: ${{ inputs.artifact-name }}
         path: ${{ steps.setup.outputs.ARTIFACT_PATH }}
--- a/.github/actions/download-file/action.yml
+++ b/.github/actions/download-file/action.yml
@@ -49,7 +49,7 @@ runs:
 
     - name: "Restore cache"
       id: cache
-      uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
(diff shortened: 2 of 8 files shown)


FedeDP (approved):
LGTM thanks

FedeDP:
I am going to retry windows CI multiple times to check for the stability of the new implementation ;)

ygvalent:
> I am going to retry windows CI multiple times to check for the stability of the new implementation ;)

By all means, I can't merge this yet though, as it requires approval from somebody on "the list"

FedeDP:
Retried CI many times, always :green_circle:  :D https://github.com/cilium/tetragon/actions/runs/33335644885?pr=5525

mtardy (approved):
I didn't read the change tbh since I don't know windows script but if it's better than the previous I'm all for it!

ExceptionalHandler (approved):

PR #5546, opened 2026-09-01
Update github.com/containerd/containerd in rthooks to v2

tpapagian:


mtardy (commented):
can you then also remove this https://github.com/cilium/tetragon/blob/8f44991e710c831c642f08413ab08091009c0e8e/.github/renovate.json5#L447?

tpapagian:
> can you then also remove this
> 
> https://github.com/cilium/tetragon/blob/8f44991e710c831c642f08413ab08091009c0e8e/.github/renovate.json5#L447
> 
> ?

My understanding is that this is related to https://github.com/cilium/tetragon/blob/589a5ed947180ca3674f8d552b6900cf91bbef1e/go.mod#L13 and not https://github.com/cilium/tetragon/blob/589a5ed947180ca3674f8d552b6900cf91bbef1e/contrib/tetragon-rthooks/go.mod#L10 (i.e. github.com/containerd/**cgroups** vs github.com/containerd/**containerd**) unless I am missing something. 

UPDATE: You are possibly refer to https://github.com/cilium/tetragon/blob/8f44991e710c831c642f08413ab08091009c0e8e/.github/renovate.json5#L458. Let me remove that.

mtardy:
> > can you then also remove this
> > https://github.com/cilium/tetragon/blob/8f44991e710c831c642f08413ab08091009c0e8e/.github/renovate.json5#L447
> > 
> > ?
> 
> My understanding is that this is related to
> 
> https://github.com/cilium/tetragon/blob/589a5ed947180ca3674f8d552b6900cf91bbef1e/go.mod#L13
> 
> and not
> https://github.com/cilium/tetragon/blob/589a5ed947180ca3674f8d552b6900cf91bbef1e/contrib/tetragon-rthooks/go.mod#L10
> 
> (i.e. github.com/containerd/**cgroups** vs github.com/containerd/**containerd**) unless I am missing something.
> UPDATE: You are possibly refer to
> 
> https://github.com/cilium/tetragon/blob/8f44991e710c831c642f08413ab08091009c0e8e/.github/renovate.json5#L458
> 
> . Let me remove that.

ah thanks I pointed to the wrong place 🤦

contrib/tetragon-rthooks/cmd/setup/patch-containerd-conf.go
```diff
@@ -241,23 +249,30 @@ func enableNRI(log *slog.Logger, cnf *enableNRICmd) ([]addLine, error) {
...
-			lastLine := -1
-			elemCol := pPos.Col + 3 // by default, indent by 3
-			// find the last line for this plugin by iterating all of its elements
-			for _, k := range v.Keys() {
-				kPos := v.GetPosition(k)
-				if kPos.Line > lastLine {
-					lastLine = kPos.Line
-					elemCol = kPos.Col
+	plugins, _ := file.Get("plugins").(*toml.Tree)
+	if plugins != nil {
```
mtardy: just to confirm, is it okay to silently ignore this in case `!ok`? I see this function can return an error just in case it's not (it might be).

mtardy (approved):
LGTM, just one small comment, otherwise this is a nochange.

contrib/tetragon-rthooks/cmd/setup/patch-containerd-conf.go
tpapagian: Based on my understanding `file.Get("plugins")` can be null in the case where this section is missing. But `!ok` here would be that it is wrong type (i.e. wrong syntax) which should not happen. I have added another check here just to be on the safe side.

mtardy: ok let's merge then! :) thanks

PR #5548, opened 2026-09-01
ariane: add ariane-config

3u13r:
### Description

Add a ariane config and a mock CI workflow to test the integration. `common-post-jobs.yml` is adapted from the same file in cilium/cilium.

[This](https://github.com/3u13r/tetragon-workflows/pull/1#issuecomment-4987594310) is an example how the invocation might look like in the future.
Merging this PR doesn't change anything about the rest of the CI. It just adds one test workflow to test the Ariane integration. All existing workflows are triggered as before. Once the PR is merged, members of the cilium org can trigger the test workflow using a `/test` comment in a PR.

### Changelog

```release-note
ariane: add ariane-config and test CI workflow
```

### DIFF
--- a/.github/workflows/common-post-jobs.yml
+++ b/.github/workflows/common-post-jobs.yml
@@ -1,0 +1,46 @@
+name: Common Post Jobs
+
+on:
+  workflow_call:
+    inputs:
+      sha:
+        description: "SHA under test on which to publish the final commit status."
+        required: true
+        type: string
+      success:
+        description: "Whether all jobs in the calling workflow completed successfully."
+        required: true
+        type: boolean
+
+permissions:
+  contents: read
+  actions: read
+  statuses: write
+
+jobs:
+  commit-status-final:
+    if: ${{ always() }}
+    name: Commit Status Final
+    runs-on: ubuntu-24.04
+    permissions:
+      statuses: write
+    steps:
+      - name: Set final commit status
+        uses: cilium/actions/set-commit-status@fe0702f5df0d8e44d48f1baa226c023b73ff6b5e # main
+        with:
+          sha: ${{ inputs.sha }}
+          status: ${{ inputs.success && 'success' || 'failure' }}
+
+  fail-workflow:
+    if: ${{ inputs.success == false }}
+    name: Fail job if result is failure
+    runs-on: ubuntu-24.04
+    steps:
+      - name: Fail job if result is failure
+        # Besides reporting in the commit SHA if the test was or not successful,
+        # we should also report it in the workflow itself. This will allow to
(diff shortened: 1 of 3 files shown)


kkourt:
Addded @mtardy and @will-isovalent in case they are interested in reviewing this change.

.github/workflows/common-post-jobs.yml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  commit-status-final:
+    if: ${{ always() }}
+    name: Commit Status Final
+    runs-on: ubuntu-24.04
+    permissions:
+      statuses: write
+    steps:
+      - name: Set final commit status
+        uses: cilium/actions/set-commit-status@fe0702f5df0d8e44d48f1baa226c023b73ff6b5e # main
```
mtardy, line 29: will renovate constantly bump on any commits? could we just tag a version, that would be better and not too hard to do, agree it's a side quest however.

mtardy (approved):
Thanks, let's try.

to be fully transparent this annoys me that we need to do all this and disrupt our current workflow for just this simple feature of not running the CI on strangers, but let's see at least how it behaves on this separate workflow.

3u13r:
> to be fully transparent this annoys me that we need to do all this and disrupt our current workflow for just this simple feature of not running the CI on strangers

Just to check, the main point is that we need to open the PR in github.com and add a comment? That sounds like a problem that can be solved by adding a feature to ariane that watches PR creations itself. Let me try to figure this out and possible implement it.

The other downside this has vs. a pure github.com solution (which apparently does not exists) is that we introduce additional things that can break into the CI pipeline.

.github/workflows/common-post-jobs.yml
3u13r: Saw your issue in cilium/actions, thanks!
Yeah, this is also what cilium/cilium currently does, so I think it's fine until we have the release tag.

3u13r:
Update: I just came across https://github.com/cilium/ariane/commit/2b6737bee4789650451d122aad5290b96b7a74dd. 
There are all things in place to have a `/default` trigger that automatically executes on PR creation but only if the the author is an org member.

Let me rework the ariane-config.yaml in this PR.

Edit: Done.

[This](https://github.com/3u13r/tetragon-workflows/pull/3) is how a **PR of an org member** would look like. Notice the ariane reaction on the PR itself.
[This](https://github.com/3u13r/tetragon-workflows/pull/2) is how a **PR for an external contributor** would look like and how to trigger the tests as an org member.

kkourt (approved):
Let's try!

PR #5549, opened 2026-09-01
add celexpr bit shift operators

bar234:
Add support for bitwise shifting (part of #4661) to CelExpr.

### Description
This adds support for bit shifting (`lsh()` and `rsh()`) to CelExpr, which can be used as part of the `MatchCEL` selector. Note that it also includes logic to use `ArSh` to arithmetically shift right if the values are signed in order to sign extend.

It also expands the oracle in `cel_eval_test` to include bit shifting and was tested with `FuzzCelExpr`. Testing was first done by intentially incorrectly implementing the operators to ensure it was caught, and then corrected.

### Changelog

```release-note
celbpf: add bit shifting operators for CEL expressions (lsh, rsh)
```

### DIFF
--- a/pkg/celbpf/cel_eval_test.go
+++ b/pkg/celbpf/cel_eval_test.go
@@ -160,6 +160,23 @@ var celBitwiseXOR = celBinArithOp(
 	func(a, b uint32) uint32 { return a ^ b },
 )
 
+// Go shifts panic on a negative shift count, ensure they're positive.
+// Also, the shift count operand for ebpf shift instructions is limited to 5/6
+// bits which is why there's a value of 31/63
+var celBitwiseLSH = celBinArithOp(
+	func(a, b celTypes.Int) celTypes.Int { return a << (uint64(b) & 63) },
+	func(a, b celTypes.Uint) celTypes.Uint { return a << (uint64(b) & 63) },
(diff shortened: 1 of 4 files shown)


pkg/celbpf/cel_eval_test.go
```diff
@@ -160,6 +160,22 @@ var celBitwiseXOR = celBinArithOp(
 	func(a, b uint32) uint32 { return a ^ b },
 )
 
+// Go shifts panic on a negative shift count, ensure they're positive.
+// Also, the shift count operand for ebpf shift instructions is limited to 63/31 bits
```
mtardy: I was a bit confused about that comment: t's more like it's limited to 6 bits for 64 bits and 5 bits for 32 bits operands thus the &63 and &31 masks.

mtardy (approved):
lgtm, just one nit on a comment

pkg/celbpf/cel_eval_test.go
bar234: Oh good point, if I'm talking about bits I can see how there's a discrepancy between the number of bits available to specify the count vs the value in the count operand. I'll update this comment

FedeDP (approved):

kkourt (approved):

PR #5550, opened 2026-09-01
resuable action to remove azure apt mirrors

dangome3:
### Description
Add reusable action for removing Azure apt mirrors so we remove duplicated code in the workflows


FedeDP (approved):

mtardy (approved):
thanks, btw don't we have a cilium wide org repo of actions like that? 

edit: here https://github.com/cilium/actions

dangome3:
> don't we have a cilium wide org repo of actions like that?

right! I'll move the first commit there. Thanks!

EDIT: Will update when this PR is merged https://github.com/cilium/actions/pull/3

FedeDP (approved):

PR #5554, opened 2026-09-02
ci: add go fuzz testing

3u13r:
Related to https://github.com/cilium/tetragon/issues/4661.

### Description
Add go fuzz testing to CI.
cc: @bar234 

### Changelog

```release-note
ci: add go fuzz testing
```

### DIFF
--- a/.github/workflows/gotests.yml
+++ b/.github/workflows/gotests.yml
@@ -104,6 +104,58 @@ jobs:
         path: /sys/kernel/btf/vmlinux
         retention-days: 5
 
+  fuzz:
+    name: Go fuzz tests (${{ matrix.os }})
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu-24.04, ubuntu-24.04-arm64 ]
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      with:
+        path: go/src/github.com/cilium/tetragon/
+
+    - name: Install Go
+      uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+      with:
+        # renovate: datasource=golang-version depName=go
+        go-version: '1.27.0'
+
+    - name: Run go fuzz tests
+      env:
+        SUDO: sudo -E
+      run: |
+        cd go/src/github.com/cilium/tetragon/
+        make test-fuzz
+
+    - name: Collect fuzz failures
+      if: ${{ failure() }}
+      id: collect-fuzz-failures
+      shell: bash
+      run: |
+        cd go/src/github.com/cilium/tetragon/
+        destination="${RUNNER_TEMP}/go-fuzz-failures"
+        found=false
+
+        while IFS= read -r -d '' file; do
+          mkdir -p "$destination"
+          cp --parents -- "$file" "$destination"
+          found=true
+        done < <(git ls-files -z --others -- '**/testdata/fuzz/**')
+
+        echo "found=$found" >> "$GITHUB_OUTPUT"
+
+    - name: Upload fuzz failures
+      if: ${{ failure() && steps.collect-fuzz-failures.outputs.found == 'true' }}
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: go-fuzz-failures-${{ matrix.os }}
+        path: ${{ runner.temp }}/go-fuzz-failures
+        retention-days: 5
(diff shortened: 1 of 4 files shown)


.github/workflows/gotests.yml
```diff
@@ -73,6 +73,13 @@ jobs:
         export TETRAGON_LIB=$(realpath "bpf/objs/")
         make test GO_TEST_TIMEOUT=60m
 
+    - name: Run go fuzz tests
+      env:
+        SUDO: sudo -E
+      run: |
+        cd go/src/github.com/cilium/tetragon/
+        make test-fuzz
```
mtardy: should we run this as a parallel job?

.github/workflows/gotests.yml
```diff
@@ -104,6 +104,58 @@ jobs:
...
+
+        echo "found=$found" >> "$GITHUB_OUTPUT"
+
+    - name: Upload fuzz failures
+      if: ${{ failure() && steps.collect-fuzz-failures.outputs.found == 'true' }}
+      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      with:
+        name: go-fuzz-failures-${{ matrix.os }}
+        path: ${{ runner.temp }}/go-fuzz-failures
+        retention-days: 5
```
mtardy, line 157: have you tried and it works? (you can add an obvious issue in some fuzzed function and see). I know GHA can be surprising...

mtardy (commented):
I think why not, in some ways, running the seed corpus in the CI is already a good thing, I wonder how useful running those fuzzer for 30s might be. We don't store the generated corpus every time with this workflows correct? Maybe that would be more interesting so that these 30s combine into way more fuzzing time eventually. (maybe with https://github.com/actions/cache but this could be wiped eventually accidentally  🤔)

Does fuzzing on both platform makes sense btw? do we have fuzzed code that behaves differently on the two platforms or is platform agnostic?

3u13r:
>  I wonder how useful running those fuzzer for 30s might be. We don't store the generated corpus every time with this workflows correct?

Correct, we don't save the added corpus, it's added to the runner's Go build cache. I also think that running the fuzzer for 30s on each PR is a bit arbitrary and not really a long term fuzzing strategy. On the other hand, it's 30s for each run and the fuzzer uses a new seed every time so the time adds up. Of course, we start with 0 coverage and it needs to re-discover interesting paths on each run.

> (maybe with https://github.com/actions/cache but this could be wiped eventually accidentally 🤔)

That is probably a worthwhile follow-up.

Another option is to have a scheduled workflow running on main that fuzzes for at least some minutes. And if the fuzzer finds something on main it creates an PR with the failing input as addition to the corpus. 

Long-term we should look into a oss-fuzz setup. After all, we are building a CEL-to-eBPF compiler that is growing in complexity.



> Does fuzzing on both platform makes sense btw? do we have fuzzed code that behaves differently on the two platforms or is platform agnostic?

It makes a difference for at least [this arm64](https://github.com/cilium/tetragon/blob/6212c2615d92a16a1a1dcd6cef1bcec98be208fd/pkg/asm/assignment_linux_arm64_test.go#L13) and the [amd64](https://github.com/cilium/tetragon/blob/6212c2615d92a16a1a1dcd6cef1bcec98be208fd/pkg/asm/assignment_linux_amd64_test.go#L14) test.
I think for CEL it also makes sense, since we are loading loading the (same) eBPF program to different architectures. I guess that if the find something then we bug is actually in the kernel's eBPF translation or the cpu itself and not in our CEL to eBPF translation.

Let me know what you think. We could also restrict the arm fuzzing to `assignment_linux_arm64_test.go`.

.github/workflows/gotests.yml
3u13r: I only tried it locally, but it seems to work: https://github.com/cilium/tetragon/actions/runs/33702940201?pr=5560

```sh
$ unzip go-fuzz-failures-ubuntu-24.04.zip
Archive:  go-fuzz-failures-ubuntu-24.04.zip
  inflating: pkg/celbpf/testdata/fuzz/FuzzCelExpr/c4d90408859a5b98

$ cat pkg/celbpf/testdata/fuzz/FuzzCelExpr/c4d90408859a5b98
go test fuzz v1
uint64(0)
uint64(1)
```

mtardy: looks like the answer is yes from the test!

mtardy:
> it's added to the runner's Go build cache

ah indeed, it might be already nice!

mtardy (approved):

mtardy:
> Another option is to have a scheduled workflow running on main that fuzzes for at least some minutes. And if the fuzzer finds something on main it creates an PR with the failing input as addition to the corpus.

yeah, maybe for more serious fuzzing that would be appropriate but it sounds reasonable that we have a short fuzzing time in case someone introduces an obvious failure in a PR (since we run on each PR). Let's merge like that first :)

PR #5561, opened 2026-09-03
perf: optimise memory allocation / churn in the execve events codepath

rafaelroquetto:
### Description

This PR introduces a series of commits that reduce allocation churn in the _exec event_ delivery path. This hotpath was uncovered by the profiler, both in the heap analysis and also alloc churn:
image

The **TLDR** is that each execve event triggered lots of transient allocations: the event structs themselves, indirect structs, and the most expensive of all: dealing with the arguments list.

The _simplified_ flow is:

- `execParse()` reads the arguments
  - entails `binary.Read()` on a byte slice
  - entails a couple of `strings.Split()`
  - it finally does a `bytes.Join()` to recreate the now post-processed args string
  - it passes that to `strutils.UTF8FromBPFBytes` and assigns the results to `MsgProces.Args`
- eventually `initProcessInternalExec` receives the `MsgProcess` data described above, tries to process the args to convert them to their final sanitised `string values`
  - that is done inside `ArgsDecoder` which
    - splits the args array _again_
    - sanitises each individual argument
    - allocates several intermediate strings by the means of `args = args + string(a)` (**!!**)
    - returns the final args string

There's way more going on, but this should give the reader a notion of what's going on here, _per event_. The arguments field is constantly being split, joined, re-split, concatenated, and rejoined.

This PR tackles the issue from several fronts:
 - instead of recalculating `args` throughout different levels, it establishes the semantics that `args` and `cwd` are calculated only *once* _upstream_ at the source/time the event is first materialised - everything downstream can then simply read the `Args` and `Cwd` `string`s as is.
   - this is the biggest improvement
 - it mitigates lots of intermediate reallocs by leveraging `bytes.SplitSeq`
 - it replaces `binary.Read()` with `reader.Read()`
   - `binary.Read()` is very expensive due to reflection
   - it relied on `io.ReadFull()` in the hot path, but because the reader is passed as an interface, escape analysis can't prove `Read()` doesn't retain the buffer, so both the buffer and the reader are forced onto the heap.
 - it embeds inner structs as members of outer structs rather than keeping pointers (allowing a single allocation per event rather than 2 or 3 and doing without ptr indirections).
 - removal of unused/dead code

See individual commits/commit messages for more.
Comparative profiling (ran in my local machine/isolated group, thousands of events per second) shows a significant reduction in allocation churn, with ~12.6% lower alloc_space and ~9.1% fewer allocated objects:

```
alloc_space:     11.75 GB -> 10.28 GB    -12.6%
alloc_objects:   174.3 M -> 158.4 M      -9.1%
```

This amounts to roughly 1.5 GB less allocation traffic and 15.9 million fewer allocated objects over the profiling run.

### Changelog

```release-note
Reduce allocation churn in the exec event delivery path
```


pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -113,7 +165,6 @@ func pushExecveEvents(p procs, inInitTreeMap map[uint32]struct{}) {
 		m.Unix.Msg.Creds.Cap = processapi.MsgCapabilities{}
 		m.Unix.Msg.Namespaces = processapi.MsgNamespaces{}
 
-		m.Unix.Process.Size = p.size
```
mtardy, line 155: why the size fields related changes? should this be part of a separate commits or is this related?

pkg/sensors/exec/procevents/proc_reader_linux.go
```diff
@@ -70,13 +71,9 @@ func getCWD(pid uint32) (string, uint32) {
 	cwd, err := os.Readlink(filepath.Join(option.Config.ProcFS, pidstr, "cwd"))
 	if err != nil {
 		flags |= api.EventRootCWD | api.EventErrorCWD
-		return " ", flags
+		return "", flags
```
mtardy, line 71: do you know why we had this space string thing?

pkg/sensors/exec/exec_linux.go
```diff
@@ -73,10 +75,159 @@ func msgToExecveKubeUnix(m *processapi.MsgExecveEvent, execID string, filename s
...
+		nread, err := reader.Read(cwd)
+
+		if err != nil {
+			return "", err
+		}
+
+		if nread != int(exec.SizeCwd) {
+			return "", errors.New("cwd size mismatch")
+		}
+	}
```
mtardy: I know this is a style thing but this code feels very "empty" and not easy to read, could you maybe remove a bunch of empty lines here and there? Not all of them but maybe make the code look like the rest of the codebase.

especially when doing something with the result of a function before:
```diff
 	data, err := resolveArgsData(reader, exec)
-
 	if err != nil {
 		return "", err
 	}
```

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -81,24 +84,73 @@ func (p procs) pargs() []byte {
...
-	}
+func procFilename(p *procs) string {
+	filename := string(p.exe)
 
 	// If this is a kernel thread, we use its filename as process name
 	// similarly to what ps reports.
 	if p.kernelThread {
-		filename = fmt.Sprintf("[%s]", filename)
-		args = ""
+		return fmt.Sprintf("[%s]", filename)
```
mtardy: nit: since we are doing perf here maybe return `return "[" + filename + "]"`, I'm surprised the linter is not proposing this?

pkg/sensors/exec/exec_linux.go
```diff
@@ -73,10 +75,159 @@ func msgToExecveKubeUnix(m *processapi.MsgExecveEvent, execID string, filename s
...
+
+	if nread != int(exec.SizeArgs) {
+		return nil, errors.New("args size mismatch")
+	}
+
+	return data, nil
+}
+
+func resolveArgs(reader *bytes.Reader, exec *processapi.MsgExec) (string, error) {
```
mtardy, line 118: I haven't read the parsing of these funcs in detail and assume it's mostly the same stuff as before but this looks like this could benefit from some unit testing on `resolveArgs`, `resolveCwd` now that they are proper function? what do you think, would it be annoying to write or not?

pkg/sensors/exec/procevents/proc_reader_linux.go
```diff
@@ -36,7 +36,7 @@ const (
 )
 
 func procKernel() procs {
-	kernelArgs := []byte("<kernel>\u0000")
+	kernelArgs := []byte("<kernel>")
```
mtardy, line 39: I wonder if there was an actual reason for this 🤔

pkg/sensors/exec/procevents/proc_reader.go
mtardy: ah this might be part of `procevents, proc: drop vestigial args-buffer trimming and PrependPath`

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -41,14 +41,10 @@ func stringToUTF8(s []byte) []byte {
 }
 
 type procs struct {
-	psize             uint32
 	ppid              uint32
 	pnspid            uint32
 	pflags            uint32
 	pktime            uint64
-	pcmdline          []byte
```
mtardy, line 46: just as a question I was a bit surprised that we don't use the parent command line elsewhere since we use the cmdline itself.

pkg/sensors/exec/procevents/proc_reader_linux.go
mtardy: btw you revert your change in `procevents, proc: drop vestigial args-buffer trimming and PrependPath`

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -41,14 +41,10 @@ func stringToUTF8(s []byte) []byte {
...
-	pexe              []byte
```
mtardy, line 47: q: similarly it feels counterintuitive that we don't use the parent exe, how can we replace this with the cwd and the args, is it that at this stage we still have the exe at args[0]?

pkg/sensors/exec/exec_linux.go
```diff
@@ -223,6 +223,45 @@ func resolveCwd(reader *bytes.Reader, exec *processapi.MsgExec) (string, error)
...
+	if exec.SizeEnvs == 0 {
+		return nil, nil
+	}
+
+	data, err := readRawBytes(reader, exec, exec.SizeEnvs, api.EventDataEnvs)
+
+	if err != nil {
+		return nil, err
+	}
+
```
mtardy, line 232: style: similarly in this commit could you condense a bit the code?

mtardy (commented):
Thanks I feel this is really cool work, appreciate the cover letter details!

Particularly appreciate the benchmark results in the commit logs (is it a benchmark we already have in the code? if not could you add it if it's not too much work)

Just a few notes it feels like some things have slipped in the wrong commits or been reverted by mistakes, also a few comments on the style if you can remove some empty lines, see comment.

pkg/sensors/exec/procevents/proc_reader.go
rafaelroquetto: yeah I was surprised too, but this is only being referenced by the `args()` function that has also been removed - this was used ultimately to derive the _size_ of the parent args command and use that as a variable in truncating the buffer (whose code was removed on line 87) - nothing else.

pkg/sensors/exec/procevents/proc_reader.go
rafaelroquetto: same as previous comment - this was just being used to compute the size of the buffer, and then to truncate it.

Let me try to elaborate it:

These functions were being used to derive to variables down below: `rawArgs` and `rawPargs`.

`rawPargs` was just being used to compute `p.flags` and `p.size`:

```
			if len(rawPargs) > len(rawArgs) {
				p.pflags |= api.EventTruncArgs
				rawPargs = rawPargs[:len(rawPargs)-1]
				p.psize--
            }
```

which is gone.

`rawArgs` was also to compute flags and sizes:

```
			} else {
				p.flags |= api.EventTruncArgs
				rawArgs = rawArgs[:len(rawArgs)-1]
				p.size--
			}
```

and then later it was passed to procsFilename() to compute `filename` and `args`:

```
	args, filename := procsFilename(rawArgs)
```

that is all gone - `args` now comes from `procArgs` and `filename` comes from `procFilename()`.

pkg/sensors/exec/procevents/proc_reader.go
rafaelroquetto: ahh good catch, fixed!

pkg/sensors/exec/procevents/proc_reader.go
rafaelroquetto: that's it

pkg/sensors/exec/procevents/proc_reader_linux.go
rafaelroquetto: honestly I believe it was a mistake, the original code had:

```
	if (flags & api.EventRootCWD) == 0 {
		args = args + " " + cwd
	}
```

which in certain cases would simply append double spaces. This would then be further split again downstream by the now removed `ArgsDecoder`, which would drop it and return `/` because EventRootCWD is set:

```
	if (flags & api.EventNoCWDSupport) != 0 {
		cwd = ""
	} else if (flags & api.EventErrorCWD) != 0 {
		cwd = ""
	} else if (flags & api.EventRootCWD) != 0 {
		cwd = "/"
	} else if (flags & api.EventProcFS) != 0 {
		cwd = strings.TrimSpace(string(argTokens[len(argTokens)-1]))
	} else {
		cwd = string(argTokens[len(argTokens)-1])
	}
```

pkg/sensors/exec/exec_linux.go
rafaelroquetto: `pkg/sensors/exec/exec_test.go` covers these _indirectly_ - I asked Claude to detail what is being exercised, and it tells me:

```
- Empty args — filename + cwd as inline strings, no args
- Empty args and cwd — filename only, no args, no cwd
- Filename as data event — filename via data-event path, cwd inline, no args
- Args as data event — args via data-event path (arg1\0arg2\0), filename + cwd inline
- Filename and args as data event — both filename and args via data-event path, cwd inline
- Filename and args as non-utf8 — invalid UTF-8 bytes in filename, args, and cwd
- Filename with api.EventErrorFilename — <enomem> fallback filename, cwd inline, no args
- Filename, args, cwd and envs — all four fields populated, envs present
- Filename, args, cwd and zero envs — all four fields populated, envs empty/absent
```

having said that, I am happy to write the unit tests - what do you think (non-rhetorical/genuine question)?

pkg/sensors/exec/exec_linux.go
rafaelroquetto: yeah, for sure! Will amend it

pkg/sensors/exec/procevents/proc_reader_linux.go
rafaelroquetto: it's because we no longer treat `p.exe` as a null terminated string that is part of the args, but rather as a standalone value now (so `\00` is no longer required). In other words, before we had:

```
func (p procs) args() []byte {
	// exe and cmdline are already in UTF8
	return proc.PrependPath(string(p.exe), p.cmdline)
}
```

that is entirely gone, and the now code simply does:

```
func procFilename(p *procs) string {
	filename := string(p.exe)

	// If this is a kernel thread, we use its filename as process name
	// similarly to what ps reports.
	if p.kernelThread {
		return "[" + filename + "]"
	}

	return filename
}
```

--

as for the reintroduction on the next comment, I vaguely remember something failing, but I can't recall it - it's nonetheless wrong, so I will just revert the reintroduction, push an update, and take it from there if CI fails.

pkg/sensors/exec/exec_linux.go
rafaelroquetto: you bet

rafaelroquetto:
> Thanks I feel this is really cool work, appreciate the cover letter details!

Thanks! :)

> Particularly appreciate the benchmark results in the commit logs (is it a benchmark we already have in the code? if not could you add it if it's not too much work)

I don't think that is currently feasible - the bench marks are not per commit (apart from the first one), but end-to-end pprof diffs of A/B tests that ran on my machine for several minutes (or sometimes hours) and crafted for the execve event case - these are not currently upstreamable.

pkg/sensors/exec/procevents/proc_reader_linux.go
rafaelroquetto: update: it really was a leftover, fixed - thanks for spotting that

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -72,75 +71,76 @@ type procs struct {
...
+			args.WriteByte(' ')
+		}
+
+		if len(arg) == 0 {
+			args.WriteString(`""`)
+			continue
+		}
+
+		hasWhiteSpace := bytes.Contains(arg, []byte{' '})
+
```
mtardy:
```suggestion
```

pkg/sensors/exec/exec_linux.go
```diff
@@ -80,10 +79,190 @@ func msgToExecveKubeUnix(m *processapi.MsgExecveEvent, execID string, filename s
...
+}
+
+func readRawBytes(reader *bytes.Reader, exec *processapi.MsgExec, size uint16, flags uint32) ([]byte, error) {
+	if size == 0 {
+		return nil, nil
+	}
+
+	if exec.Flags&flags != 0 {
+		data, err := readData(reader, size)
+
```
mtardy:
```suggestion
```

pkg/sensors/exec/exec_linux.go
```diff
@@ -80,10 +79,190 @@ func msgToExecveKubeUnix(m *processapi.MsgExecveEvent, execID string, filename s
...
+		if err != nil {
+			return nil, err
+		}
+
+		return data, nil
+	}
+
+	data := make([]byte, size)
+
```
mtardy:
```suggestion
```

pkg/sensors/exec/exec_linux.go
```diff
@@ -80,10 +79,190 @@ func msgToExecveKubeUnix(m *processapi.MsgExecveEvent, execID string, filename s
...
+		if args.Len() > 0 {
+			args.WriteByte(' ')
+		}
+		if len(arg) == 0 {
+			args.WriteString(`""`)
+			continue
+		}
+
+		hasWhiteSpace := bytes.Contains(arg, []byte{' '})
+
```
mtardy:
```suggestion
```

pkg/sensors/exec/exec_linux.go
```diff
@@ -80,10 +79,190 @@ func msgToExecveKubeUnix(m *processapi.MsgExecveEvent, execID string, filename s
...
+	return strutils.UTF8FromBPFBytes(data), nil
+}
+
+func resolveProcEnvs(reader *bytes.Reader, exec *processapi.MsgExec) ([]string, error) {
+	if exec.SizeEnvs == 0 {
+		return nil, nil
+	}
+
+	data, err := readRawBytes(reader, exec, exec.SizeEnvs, api.EventDataEnvs)
+
```
mtardy:
```suggestion
```

pkg/process/process.go
```diff
@@ -329,7 +329,10 @@ func initProcessInternalExec(
 	parent tetragonAPI.MsgExecveKey,
 ) *ProcessInternal {
 	process := event.Process
-	args, cwd := ArgsDecoder(process.Args, process.Flags)
+
+	args := process.Args
+	cwd := process.Cwd
+
```
mtardy:
```suggestion
	args := process.Args
	cwd := process.Cwd
```

pkg/sensors/exec/exec_linux.go
```diff
@@ -80,10 +79,190 @@ func msgToExecveKubeUnix(m *processapi.MsgExecveEvent, execID string, filename s
...
+	if len(data) > 0 && data[len(data)-1] == '\x00' {
+		data = data[:len(data)-1]
+	}
+	if len(data) == 0 {
+		return "", nil
+	}
+
+	size := 0
+	numArgs := 0
+
```
mtardy:
```suggestion
```

pkg/sensors/exec/exec_linux.go
```diff
@@ -80,10 +79,190 @@ func msgToExecveKubeUnix(m *processapi.MsgExecveEvent, execID string, filename s
...
+		}
+	}
+
+	if numArgs > 1 {
+		size += numArgs - 1
+	}
+
+	var args strings.Builder
+	args.Grow(size)
+
```
mtardy:
```suggestion
```

pkg/sensors/exec/exec_linux.go
```diff
@@ -216,6 +216,44 @@ func resolveCwd(reader *bytes.Reader, exec *processapi.MsgExec) (string, error)
...
+		return nil, err
+	}
+
+	// cut the zero byte
+	if data[len(data)-1] == '\x00' {
+		data = data[:len(data)-1]
+	}
+
+	var ret []string
+
```
mtardy: another whitespace

pkg/sensors/exec/exec_linux.go
mtardy: okay if you think this is already covered by testing no need to double it!

mtardy (approved):
Okay thanks for all the details in the responses. Overall lgtm and if it's already tested then we can have more confidence on the parsing changes.

Aside: I'll just post a few suggestions on the style nit sorry, nice improvement but, I think in general you can remove whitelines just after retrieving function value and treating them (like with smt, err := func()) and also when declaring a variable and using it just below. I think it would make your code blend in the existing code base better. Not a blocker, we can merge, but would be nice. Maybe you can wait on other people's review if you have to make other change before sending another version with this.

rafaelroquetto:
@mtardy I appreciate the review and no need to be sorry - it makes absolutely sense to me to enforce staying consistent with the codebase - I just happen to massively prefer the more "empty" style as to _me_ (and I know a lot of people would disagree) they read better, and old habits die hard - but I will adjust, no biggie.

I do wonder if it would make sense to start enforcing those with a linter like https://github.com/bombsimon/wsl or https://github.com/mvdan/gofumpt - happy to look into the impact of enabling those if you think it would make sense.

rafaelroquetto:
Info: fixed whitespace nits and rebased. No substantial changes otherwise

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -81,24 +84,70 @@ func (p procs) pargs() []byte {
...
+		if hasWhiteSpace {
+			args.WriteByte('"')
+		}
+
+		if utf8.Valid(arg) {
+			args.Write(arg)
+		} else {
+			args.WriteString(strutils.UTF8FromBPFBytes(arg))
+		}
+
```
FedeDP, line 118: Do we still need the `if utf8.Valid()` here since now `UTF8FromBPFBytes` already provides it for us?

pkg/sensors/exec/procevents/proc_reader_linux.go
```diff
@@ -70,13 +71,9 @@ func getCWD(pid uint32) (string, uint32) {
...
 	if err != nil {
 		flags |= api.EventRootCWD | api.EventErrorCWD
-		return " ", flags
+		return "", flags
 	}
 
-	if cwd == "/" {
-		cwd = " "
-		flags |= api.EventRootCWD
-	}
```
FedeDP, line 79: Q: why was this code removed? Also, do we ever have any `EventRootCWD` bit set now without `EventErrorCWD` too?

FedeDP: I mean, i'd expect at least the flag to be still set:
```go
if cwd == "/" {
	flags |= api.EventRootCWD
} 
```

FedeDP (approved):
LGTM, left some questions/comments.
Great job!

pkg/sensors/exec/procevents/proc_reader.go
rafaelroquetto: In this case yes, it's an optimisation: `args.Write()` takes a `[]byte` directly, so there's no need for a temporary `string()` materialisation as is the case in the `else` branch.

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -81,24 +84,70 @@ func (p procs) pargs() []byte {
...
+		hasWhiteSpace := bytes.Contains(arg, []byte{' '})
+		if hasWhiteSpace {
+			args.WriteByte('"')
+		}
+
+		if utf8.Valid(arg) {
+			args.Write(arg)
+		} else {
+			args.WriteString(strutils.UTF8FromBPFBytes(arg))
+		}
```
kkourt: nit: Given that your first patch introduces a fast path for UTF8FromBPFBytes, I wonder if the above branch is intentional so that you use args.Write rather than WriteString.

Note sure if there is a difference (https://cs.opensource.google/go/go/+/refs/tags/go1.27.1:src/strings/builder.go;l=112, https://cs.opensource.google/go/go/+/refs/tags/go1.27.1:src/strings/builder.go;drc=b4771dc03289ec1d038f550063acb6d6a3e54b49;l=87), but if there is how about we add a `strituls.WriteUTFFromBPFBytes(b *strings.Builder, ...)` function?

pkg/sensors/exec/procevents/proc_reader.go
```diff
@@ -81,24 +84,70 @@ func (p procs) pargs() []byte {
...
 	if p.kernelThread {
-		filename = fmt.Sprintf("[%s]", filename)
-		args = ""
+		return "[" + filename + "]"
 	}
 
+	return filename
+}
+
+func procArgs(p *procs) string {
```
kkourt, line 86: As an aside, part of the complexity/ugliness comes from having:

```
  // Arguments passed to the binary at execution.
  string arguments = 6;
```

In our proto descriptions. I think this was a mistake :)

Is there a reasonable way to get away from this?

One approach would be to have something like:
```
  // Arguments passed to the binary at execution.
 repeated string args = 22;
```

Along with a command line switch to enable using it.

I'm not sure if we  we can fully deprecate it the arguments being a string, but having a command line switch for the sane option would be useful, I think.

pkg/sensors/exec/procevents/proc_reader_linux.go
rafaelroquetto: because of the introduction of `MsgProcess.Cwd` - prior to that, the cwd was baked into `Args` and that flag was used to signal to downstream code how to interpret the last arg. So for the `proc_reader` case, it's not being used.

_However_ ebpf can still push an event with `struct msg_process` (which maps to `processapi.MsgExec`, _not_ `processapi.MsgProcess` - the latter is go only). That codepath reads the flag in `resolveCwd()` (in `exec_linux.go`), and if it's `EventRootCWD`, it sets `MsgProcess.Cwd = "/"`.

So in a nutshell, there are 2 parallel codepaths:

1. ebpf:
```
MsgProcess --- reads flag --- > materialises MsgExec.Cwd via resolveCwd()
```

2. proc_reader
```
reads /proc --- reads Cwd directly -> stores Cwd in MsgExec.Cwd directly
```

These are the upstream event sources. Once it hits downstream, the CWD flags don't matter anymore: `Msg.Cwd` is already the final string (which could be "/") that can be consumed.

kkourt (approved):
LGTM, thanks! 🚀

pkg/sensors/exec/procevents/proc_reader.go
rafaelroquetto: > nit: Given that your first patch introduces a fast path for UTF8FromBPFBytes, I wonder if the above branch is intentional so that you use args.Write rather than WriteString.

that is exactly the intent!

And yeah, I like the idea of WriteUFFromBytes, I will introduce it.

pkg/sensors/exec/procevents/proc_reader.go
rafaelroquetto: I think it makes sense (and actually, during the development of this PR, I had an `Arguments` struct field alongside the "old" `Args` as well - I just cleaned it up because at that level it made no longer sense to have both).

I will open an issue, assign it to myself, and investigate the effort to tackle this.

rafaelroquetto: https://github.com/cilium/tetragon/issues/5611

pkg/sensors/exec/procevents/proc_reader_linux.go
FedeDP: Makes sense, thanks.

rafaelroquetto:
Rebased and introduced 2 new commits implementing and using `WriteUTF8FromBPFFBytes()` as proposed by @kkourt.

PR #5562, opened 2026-09-03
fix: close GetUrl HTTP response bodies

immanuwell:
### Description

`GetUrl` leaves successful HTTP response bodies open. 
Repeated matching events can keep connections around until Tetragon hits a socket or file descriptor limit

Close the body after every successful request. 
Requests still fire like before, just no leaked body

Repro:
1. Point the checked-in `open_geturl.yaml` policy at an HTTP server
2. Repeatedly open `/etc/passwd`
3. Watch Tetragon's open connections keep growing

The new unit test reproduces the missed close directly

### Changelog

```release-note
Fix GetUrl actions leaking HTTP response bodies.
```

### DIFF
--- a/pkg/sensors/tracing/generickprobe.go
+++ b/pkg/sensors/tracing/generickprobe.go
@@ -1298,7 +1298,11 @@ func loadGenericKprobeSensor(bpfDir string, load *program.Program, maps []*progr
 
 func getUrl(url string) {
 	// We fire and forget URLs, and we don't care if they hit or not.
-	http.Get(url)
+	resp, err := http.Get(url)
+	if err != nil {
+		return
+	}
+	defer resp.Body.Close()
 }
 
 func dnsLookup(fqdn string) {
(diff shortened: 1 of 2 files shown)


pkg/sensors/tracing/generickprobe.go
```diff
@@ -1298,7 +1298,11 @@ func loadGenericKprobeSensor(bpfDir string, load *program.Program, maps []*progr
 
 func getUrl(url string) {
 	// We fire and forget URLs, and we don't care if they hit or not.
-	http.Get(url)
+	resp, err := http.Get(url)
+	if err != nil {
+		return
+	}
+	defer resp.Body.Close()
```
kevsecurity, line 1305: nit: Why defer and not just close?

kevsecurity (approved):
Thanks!

mtardy (approved):
thanks, agree we can just close immediately in this case but this doesn't make any real difference given the current function.

pkg/sensors/tracing/generickprobe.go
immanuwell: mostly to keep the cleanup next to the successful request path
agreed that either works here

FedeDP (approved):

PR #5563, opened 2026-09-03
new: load uprobe `override:argNewSymbol` from `override:sopath`

FedeDP:
### Changelog

```release-note
new: load uprobe `override:argNewSymbol` from `override:sopath`
```

Policytest passing on arm64 vm:
```
root@kind-6:/host# uname -a
Linux kind-6.18 6.18.48 #1 SMP PREEMPT Mon Aug 31 01:41:24 UTC 2026 aarch64 GNU/Linux

root@kind-6:/host# ./tetra policytest run uprobe-override-symbol-so
level=info msg="Loading sensor" name=generic_uprobe
level=info msg="Loading kernel version 6.18.48"
level=info msg="Loaded generic uprobe program: bpf/objs/bpf_generic_uprobe_v61.o -> /host/contrib/tester-progs/uprobe-simple [pizza]"
level=info msg="Loaded sensor successfully" sensor=generic_uprobe
level=info msg="Unloading sensor generic_uprobe"
level=info msg="Sensor unloaded" sensor=generic_uprobe maps-error=[]
P: uprobe-override-symbol-so ()                          ✅   1/1 scenario(s) succeeded (36.96s)
└S: check call was overridden with symbol loaded by SO   🟢   
```

## Notes

This works even if the process is already running, just like normal uprobes.

### BPF maps
| Name           | Key                                             | Usage                                                                                                                                                                        |
|----------------|-------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| regs_map       | UprobeRegsMapID()                               | Stores reg_assignment AND sopath + symbol address within sopath lib + {mmap,dlopen} relative (to libc base for ASLR) addresses                                                                     |
| pending_calls  | pid_tgid                                        | Stores current state-machine state and original registers to be restored on final jump. Thread-scoped so that each thread in a process has its own pending call lookup entry |
| libc_addrs_map | tgid                                            | Stores final addresses for {mmap,dlopen} for the process, computed after finding the libc base address                                                                 |
| resolved_cache | cache_key {     __u32 tgid;     __u32 sym_id; } | Stores final loaded address for sym_id to speed up subsequent calls redirection. Cached address is unique for each <process, symbol> tuple                                   |

### BPF code
* flow is triggered by `uprobe_offload()` when `regs->sopath_len` > 0
* the flow starts by checking the `resolved_cache` map, to check whether we already cached the symbol address
* we load the process libc base address from its VMA
* we store the libc-base-shifted {mmap,dlopen} addresses in `libc_addrs_map` for the process
* we store original registers of the call that will be later restored prior to the final jump to the override-symbol
* call mmap to alloc space for `sopath`
* mmap retprobe triggers; store in the newly allocated space `sopath`, then call `dlopen(sopath)`
* dlopen retprobe triggers; look for the loaded library base address and add to it the symbol (local to library) address; cache the value
* finally, restore original call registers and jump to it

### Userspace code
* most of the logic lives in `pkg/sensors/program/dynamic_so.go`
* at first policy that enables the `sopath` behavior, we look for host libc and compute {mmap,dlopen} offsets via ELF
* we also load symbol address in `sopath` library via ELF
* then we follow same flow that is used to attach `sleepable_offload` program, ie: in `uprobeAttach` we attach the extra programs (see `dynamic_so.go::uprobeExtraAttach()`)


FedeDP:
Rebased on top of https://github.com/cilium/tetragon/pull/5589; now the tarball-tests CI runs policy test (on arm64 too) so we get a basic coverage of the feat on arm64.

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,382 @@
...
+	__u64 orig_regs[8]; // saved rdi,rsi,rdx,rcx,r8,r9 (or x0-x7 on arm)
+	__u64 scratch_addr; // page returned by our mmap hijack
+	__u64 dlopen_handle; // returned by dlopen, needed for dlsym call
+	__u64 true_return_addr;
+	__u64 orig_addr;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 1024); // TODO: make it customizable from options?
```
FedeDP: Do we want to make it configurable? Or lower it? I am not sure.

This is the number of threads that can concurrently run the whole dynamic SO flow.

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,382 @@
...
+	__u64 mmap_addr;
+	__u64 dlopen_addr;
+	__u64 dlsym_addr;
+};
+
+// Each traced process has its own libc base address (because of ASLR)
+// and thus its own mmap, dlopen and dlsym addresses.
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 1024); // TODO: make it customizable from options?
```
FedeDP, line 60: Again, do we want to make this configurable?
This is tied to the number of processes that run the traced binary.

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,382 @@
...
+
+// Resolved symbol cache, scoped per-process (not per-thread!)
+struct cache_key {
+	__u32 tgid;
+	__u32 sym_id;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 1024); // TODO: make it customizable from options?
```
FedeDP, line 73: Of course, do we want to make this configurable too?
This is the number of `<process,sym_id>` tuples.

bpf/process/uprobe_offload.h
```diff
@@ -5,40 +5,16 @@
 #define __UPROBE_OFFLOAD_H__
 
 #include "regs.h"
-
-struct reg_assignment {
```
FedeDP, line 9: Since `regs_map` has to be used from uprobe_dyn too, had to move it there.
Another solution is to include `uprobe_dyn.h` AFTER these declarations; didn't like that solution in this case because `regs_map` had to be also updated to reflect new fields for dynamic SO feat thus i found it more coherent to just move it there.

pkg/sensors/program/dynamic_so.go
```diff
@@ -0,0 +1,247 @@
...
+	"github.com/cilium/tetragon/pkg/k8s/apis/cilium.io/v1alpha1"
+	"github.com/cilium/tetragon/pkg/logger"
+	"github.com/cilium/tetragon/pkg/sensors/unloader"
+)
+
+const (
+	mmapName   = "mmap"
+	dlopenName = "dlopen"
+	dlsymName  = "dlsym"
+	hostLibcRe = "/usr/lib/*/libc.so.6"
```
FedeDP, line 25: Do we want to make this configurable somehow? Perhaps `libc.so.6` is elsewhere.

pkg/api/processapi/processapi.go
```diff
@@ -294,10 +294,21 @@ type RegAssignment struct {
...
 	Pad uint32
+
+	// Related to uprobe dynamic SO support
+	Sopath     [SOPATH_MAX]byte
+	Symbol     [SYMBOL_MAX]byte
+	SopathLen  uint32
+	SymbolLen  uint32
+	MmapAddr   uint64
+	DlopenAddr uint64
+	DlsymAddr  uint64
```
FedeDP: We pay the additional memory cost even if we are not using the feature.
I found this much cleaner than introducing a new map with same key than `regs_map` for that.

pkg/sensors/program/loader_linux.go
```diff
@@ -348,7 +351,7 @@ func MultiUprobeAttach(load *Program, bpfDir string) AttachFunc {
 	}
 }
 
-func attachMultiUpobeLink(load *Program, prog *ebpf.Program, path string,
+func attachMultiUprobeLink(load *Program, prog *ebpf.Program, path string,
```
FedeDP, line 354: name fix :)

pkg/sensors/program/dynamic_so.go
FedeDP: Users can maybe provide the libcso_path for us? Perhaps even in the policy spec? 
This way, when we'll support uprobes for processes inside containers, users can specify the relative-to-container-root-filesystem libc path.
Also, i think this is the only thing that would need to be adapted when https://github.com/cilium/tetragon/pull/5190 lands, because we need to look for the correct libc to attach our {mmap,dlopen,dlsym} hooks.

olsajiri:
my first impression was this is fragile, but could not really find any big issue, perhaps:
- if signal comes and does dlopen/mmap/dlsym.. unlikely I guess, but that will screw all up, right?
- same if the plugin library calls dlopen/.. in constructor, probably not an issue
- it's uprobes on hot code, any process loading will be slower
- we don't save/restore xmm regs, but not sure dlopen/mmap are using them
- I guess it's meant to be used to patch function entry only right? executing this from the middle of the function  we could overwrite stack redzone
- I think this won't work with intel shadow stack feature, probably not a big issue

what's the use case for this? I wonder if we go that far that we have plugin library for app, we could just restart the app with LD_PRELOAD=plugin.so and have bpf code to redirect execution to new function .. but if we can't do that, there's probably not too many good options left ;-)

FedeDP:
Ehy jiri, thanks for the feedback!
> if signal comes and does dlopen/mmap/dlsym.. unlikely I guess, but that will screw all up, right?

100%; i am actually working to get rid of dlsym; one less jump to think about (see PR body NOTES); in any case, this is a problem

> same if the plugin library calls dlopen/.. in constructor, probably not an issue

I don't understand this one, care to expand? Thanks!

> it's uprobes on hot code, any process loading will be slower

Agree, even if the libc uprobes immediately return when not called from a process in pending_call. 
Again, dropping `dlsym` program is nice; i played a bit during development with dropping `mmap` program too, eg:
```
// Get current user stack pointer
__u64 sp = PT_REGS_SP(ctx);

// Pick a safe scratch address below SP.
// Leave headroom for the x86-64 "red zone" (128 bytes)
// plus extra margin so we don't clobber data the callee needs.
__u64 scratch_addr = sp - 512;
```
This should work fine (at least on x86); i don't know how reliable it is though; it was suggested by an AI and i tested that it worked fine, but i preferred the explicit `mmap` since it felt less tricky.
BUT if we are concerned about jumps, with this trick we would only be left with a single `dlopen` jump, thus a single `libc` uprobe.
Also, most of the eventual perf hit by hooking into libc comes from `mmap` of couse, since it is the most widely used.

> we don't save/restore xmm regs, but not sure dlopen/mmap are using them

Doesn't seem an issue in my tests.

> I guess it's meant to be used to patch function entry only right? executing this from the middle of the function we could overwrite stack redzone

100%

> I think this won't work with intel shadow stack feature, probably not a big issue

Uh probably not.

> what's the use case for this? I wonder if we go that far that we have plugin library for app, we could just restart the app with LD_PRELOAD=plugin.so and have bpf code to redirect execution to new function .. but if we can't do that, there's probably not too many good options left ;-)

Well this is basically live patching via tetragon uprobes; note that we can attach the policy at runtime to an already-running process, thus no need to restart it.
I think it's a nice feature that could be marked as **experimental**, but still a very interesting novel approach.
Don't get me wrong, i know it's fragile, but i don't think there is any better solution; by the way, that's why i tried to make the code of this PR as self-embedded as possible; basically the whole logic is:
* bpf: uprobe_dyn*.h
* userspace: dynamic_so.go

olsajiri:
>     same if the plugin library calls dlopen/.. in constructor, probably not an issue
> 
> I don't understand this one, care to expand? Thanks!

I meant the plugin that bpf program loads could have _init/constructor code defined that could call dlopen/... functions

FedeDP:
Oh right, understood, thanks!

FedeDP:
Force pushed changes:
* marked feature as **experimental** in docs
* dropped `dlsym` jump/bpf program; instead, symbol address gets statically computed via ELF (the same way we do for the uprobe'd symbol) and then we look for the `dlopen`ed library base address in `dlopen` retprobe

FedeDP:
Improved the test to show that we are able to override multiple uprobes hook with the same function.

bpf/process/uprobe_dyn_arm64.h
```diff
@@ -0,0 +1,90 @@
...
+FUNC_INLINE void jump_to_dlopen(struct pt_regs *ctx, __u64 scratch, __u64 dlopen_addr)
+{
+	ctx->regs[0] = scratch;
+	ctx->regs[1] = RTLD_NOW;
+	jump_to(ctx, dlopen_addr);
+}
+
+#endif
+#endif
+#endif
```
mtardy: could you add the respective comments? same on the x86 implem?

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 1);
+	__type(key, __u32); // sym_id, based upon UprobeID and selector idx (see UprobeRegsMapID())
+	__type(value, struct uprobe_regs);
+} regs_map SEC(".maps");
+
+// --- Per-thread in-flight resolution state (the actual "state machine") ---
```
mtardy: any reason for the `---` prefix and suffix in the comment?

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+	__u64 orig_regs[8]; // saved rdi,rsi,rdx,rcx,r8,r9 (or x0-x7 on arm)
+	__u64 scratch_addr; // page returned by our mmap hijack
+	__u64 true_return_addr;
+	__u64 orig_addr;
+};
+
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 1024); // TODO: make it customizable from options?
+	__type(key, __u64); // pid_tgid (thread-scoped so that each thread in a process has its own pending call lookup entry)
```
mtardy: would BPF_MAP_TYPE_TASK_STORAGE be relevant here since the content is fairly small?

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+		}
+	}
+
+	return found ? min_start : 0;
+}
+
+FUNC_INLINE int
+uprobe_dyn_state_machine(struct pt_regs *ctx, struct uprobe_regs *regs, __u32 sym_id)
+{
+	static const char libc[SONAME_MAX] = "libc.so.6\0";
```
mtardy: nit: I don't think you need the `\0`:
```c
static const char libc[] = "libc.so.6";
```
would generate a null terminated string for bpf https://godbolt.org/z/n1oqEe55P

and here also with a fixed length it will be filled with zero anyway for the rest https://godbolt.org/z/obsKe6dPv.

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+{
+	struct vm_area_struct *vma;
+	__u64 min_start = 0;
+	bool found = false;
+	struct task_struct *task = (struct task_struct *)get_current_task_btf();
+
+	if (!task)
+		return 0;
+
+	bpf_for_each(task_vma, vma, task, 0)
```
mtardy, line 102: this was added in 6.4, I imagine it's okay for this use case?

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+		char name[SONAME_MAX] = {};
+		long n;
+		__u64 vm_start;
+
+		dentry = BPF_CORE_READ(vma, vm_file, f_path.dentry);
+		if (!dentry)
+			continue;
+
+		n = probe_read_kernel_str(name, sizeof(name),
+					  BPF_CORE_READ(dentry, d_name.name));
```
mtardy: why do this in two times and not `nameptr = BPF_CORE_READ(vma, vm_file, f_path.dentry, d_name.name);`?

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+			continue;
+
+		vm_start = BPF_CORE_READ(vma, vm_start);
+		if (!found || vm_start < min_start) {
+			min_start = vm_start;
+			found = true;
+		}
+	}
+
+	return found ? min_start : 0;
```
mtardy: might be okay to return early on the first find since vma are sorted by address and [bpf_for_each will just call vma_next](https://github.com/torvalds/linux/blob/5dd1818b15d98d4a20806cd00b1b40320b06004f/kernel/bpf/task_iter.c#L957).

bpf/process/uprobe_dyn.h
mtardy, line 1: I can't comment on the commit log itself but
```
new(bpf): bpf implementation for the feature
```
is to vague as self contained commit message, it is understandable as part of the PR but needs to be indepednant

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+uprobe_dyn_state_machine(struct pt_regs *ctx, struct uprobe_regs *regs, __u32 sym_id)
+{
+	static const char libc[SONAME_MAX] = "libc.so.6\0";
+	__u64 pid_tgid = get_current_pid_tgid();
+	__u32 tgid = pid_tgid >> 32;
+	struct cache_key ckey = { .tgid = tgid, .sym_id = sym_id };
+	__u64 *cached;
+	__u64 base;
+	struct libc_addrs addrs = {};
+	struct pending_call pc = {};
```
mtardy: nit: this acronym is maybe a bit misleading because of pc for program counter considering we are writing uprobe code

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+		// find_library_base() loop failed to find a libc for the binary;
+		// Skip flow for the binary.
+		skip_flow(sym_id, pid_tgid);
+		return 0;
+	}
+	// Update mmap and dlopen addresses by adding libc_base
+	DEBUG_SO("libc base addr for %d: 0x%lx", tgid, base);
+	addrs.mmap_addr = regs->mmap_addr + base;
+	addrs.dlopen_addr = regs->dlopen_addr + base;
+	map_update_elem(&libc_addrs_map, &tgid, &addrs, BPF_ANY);
```
mtardy: why do use tgid here and not the pid_tgid? in general do you need the tgid?

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+SEC(OFFLOAD)
+int handle_mmap_ret(struct pt_regs *ctx)
+{
+	__u64 pid_tgid = get_current_pid_tgid();
+	__u32 tgid = pid_tgid >> 32;
+	struct pending_call *pc;
+	__u64 scratch;
+	struct uprobe_regs *regs;
+	struct libc_addrs *addrs;
+	__u32 sopath_len;
```
mtardy, line 196: style nit: kernel style try to use reverse x-mas tree declaration style

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+
+	pc->stage = STAGE_DLOPEN_PENDING;
+
+	push_fake_frame(ctx, pc);
+
+	jump_to_dlopen(ctx, scratch, addrs->dlopen_addr);
+
+	DEBUG_SO("JUMPING to dlopen '%s'!", regs->sopath);
+
+	return 0;
```
mtardy, line 237: nit style: whitespace

pkg/k8s/apis/cilium.io/v1alpha1/types.go
```diff
@@ -391,6 +391,10 @@ type ActionSelector struct {
 	// Beware that the symbol MUST be binary compatible with the traced uprobe symbol.
 	ArgNewOffset int64 `json:"argNewOffset,omitempty"`
 	// +kubebuilder:validation:Optional
+	// For ArgNewSymbol, indicates to load the new symbol
+	// from a shared object located at sopath.
+	SoPath string `json:"sopath,omitempty"`
```
mtardy, line 398: can you maybe reflect on some length validation the runtime check?

```
			// -1 to account for null terminator
			if len(action.SoPath) > processapi.SOPATH_MAX-1 {
				return fmt.Errorf("sopath %q exceeds maximum length of %d", action.SoPath, processapi.SOPATH_MAX-1)
			}
```

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,344 @@
...
+};
+
+// Each traced process has its own libc base address (because of ASLR)
+// and thus its own mmap and dlopen addresses.
+struct {
+	__uint(type, BPF_MAP_TYPE_HASH);
+	__uint(max_entries, 1024); // TODO: make it customizable from options?
+	__type(key, __u32); // tgid (processs-scoped: each traced process has its own cache for mmap and dlopen addresses)
+	__type(value, struct libc_addrs);
+} libc_addrs_map SEC(".maps");
```
mtardy, line 63: now we try to prefix tetragon maps with `tg_` now and try to stick within 16 chars for user display

pkg/sensors/program/loader_linux.go
mtardy, line 1: same for the commit name `new(pkg): userspace implementation for the feature`

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1228,6 +1231,41 @@ Since tetragon enforces no verification for the last point, this can lead to
...
+  symbols:
+  - "malloc"
+  selectors:
+  - matchActions:
+    - action: Override
+      argNewSymbol: "malloc_patched"
+      sopath: "/home/test/mymalloc.so"
+```
+
+Note that it also supports directly providing `argNewAddr` or `argNewOffset` for the symbol within the shared object.
```
mtardy: nit: please wrap

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1228,6 +1231,41 @@ Since tetragon enforces no verification for the last point, this can lead to
...
+
+In this case, following restrictions apply:
+- uprobe is attached to the beginning of the user space function;
+- user space function is called via `call` instruction;
+- kernel support to manipulate raw registers is available (6.18+);
```
mtardy: - kernel support to manipulate raw registers is only available on 6.18+;

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1228,6 +1231,41 @@ Since tetragon enforces no verification for the last point, this can lead to
...
+- the executable must dynamically link libc.so.6;
+- libc.so.6 must offer `{dlopen}` function, incorporating `libdl` ([libc.so.6 >= 2.34](https://lists.gnu.org/archive/html/info-gnu/2021-08/msg00001.html));
```
mtardy: does the `{}` have a meaning here?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1228,6 +1231,41 @@ Since tetragon enforces no verification for the last point, this can lead to
...
+- libc.so.6 must be somewhere under `/usr/lib/`;
+- new symbol must have the same signature as the original symbol.
```
mtardy, line 1264: nit: maybe move this point up in the list, it seems more important than libc details?

docs/content/en/docs/concepts/tracing-policy/selectors.md
```diff
@@ -1228,6 +1231,41 @@ Since tetragon enforces no verification for the last point, this can lead to
...
+
+{{< warning >}}
+Since tetragon enforces no verification for the last point, this can lead to
+crashing the traced application.
+{{< /warning >}}
+
+{{< warning >}}
+`sopath` is to be considered experimental.
+{{< /warning >}}
```
mtardy: not warning, maybe more note or tip that we should put at the very top of the documentation of this feature.

I don't remember the exact difference between tip and note but I think we used to use tip for this kind of disclaimers. worth trying

mtardy (commented):
Here are a few comments

One thing, some commits need to be renamed, they are "something for the feature" which makes sense in this PR context but not by themselves.

bpf/process/uprobe_dyn.h
FedeDP: Oh right, didn't know this! Thanks!

bpf/process/uprobe_dyn.h
FedeDP: Just old revision of the comment leftover :) let me fix it! Thanks!

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: It was previously `{dlopen, dlsym}` ; leftover.

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: Oh it's the last because it is followed by a warning about it.

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: Nice thanks, let me fix it!

docs/content/en/docs/concepts/tracing-policy/selectors.md
FedeDP: I used the same sentence that was used above for `argError`. Let me fix both of them.

bpf/process/uprobe_dyn.h
FedeDP: Was also wondering whether i should enforce the `_map` suffix on all of them.

bpf/process/uprobe_dyn.h
FedeDP: Nice tip thanks!

bpf/process/uprobe_dyn.h
FedeDP:
```
__type(key, __u32); // tgid (processs-scoped: each traced process has its own cache for mmap and dlopen addresses)
```
The map is process-scoped and not thread-scoped; the whole process shares the same libc base thus `mmap` and `dlopen` symbols.
What i was missing is to look for already-cached `addrs` in `uprobe_dyn_state_machine`; let me fix it!

bpf/process/uprobe_dyn.h
FedeDP: Right, let me update it :D

bpf/process/uprobe_dyn.h
FedeDP: Oh cool, let me try!

bpf/process/uprobe_dyn.h
FedeDP: Considering we require 6.18+ to manipulate raw registers, it should be fine.

bpf/process/uprobe_dyn.h
mtardy: so should we use pid not tgid?

bpf/process/uprobe_dyn.h
mtardy: no I think the suffix is not necessary, I think we mostly use this in bpffs

mtardy: otherwise you don't have enough chars lol

mtardy: it's a "nice to have" btw not hard req, it's just we tend to do on newly added stuff

bpf/process/uprobe_dyn.h
FedeDP: So, the problem is that only `pending_calls` map is thread-scoped, thus it's `task_struct` scoped.
For other maps i could:
```C
struct task_struct *task = (struct task_struct *)get_current_task_btf();
struct task_struct *leader = BPF_CORE_READ(task, group_leader);
struct libc_addrs *addrs = bpf_task_storage_get(&libc_base_addrs, leader, NULL,
                                                 BPF_LOCAL_STORAGE_GET_F_CREATE);
```
but then what happens if the `group_leader` exits? The storage area is automatically cleaned up.
For now i only applied `BPF_MAP_TYPE_TASK_STORAGE` to `pending_calls` map.

bpf/process/uprobe_dyn.h
FedeDP: Isn't `pid` in userspace the same than `tgid` in kernel context?

bpf/process/uprobe_dyn.h
FedeDP: Thanks, let me fix the map names then :D

bpf/process/uprobe_dyn.h
FedeDP: Nothing comes free :D since the code is compiled on pre 5.11 probes too, it would error out since `BPF_MAP_TYPE_TASK_STORAGE` is undefined there.
I had to `#ifdef __V511_BPF_PROG` the whole dynamic-related code.

kkourt (commented):
Had a first look at the PR, and felt like sharing my thoughts :)

This is awesome! And I mean this both in the good way (as in, the technical achievement of making this work is amazing) and that it really scares me (i.e., inspiring awe)

I feel like there are many things that could go wrong here, but was also not able to come up with specific problems.

For example:
* mmap() / dlopen() calls failing seems to be accounted for.
* As long as we respecting calling conventions, I don't see how we can break the caller
* There seems to be a race in that multiple threads on a process may end up doing the `mmap()/dlopen()` dance and we might end up with multiple instances of the loaded symbol and different ones may be called (as the cache gets updates), but I'm not sure if that's any issue (other than wasted memory). Maybe there is a problem if functions have static variables that are loaded multiple times. If we are going ahead with this, maybe we can have other threads just call the old function once one threads has initiated the replacement.

Overall, this seems like a huge footgun but many things in Tetragon are.

As I said, I was not able to come up with problems, but I've also hadn't thought about it for more than 30'. I think it would be useful to have an analysis of all the things that could go wrong and how they are accounted for.

FedeDP:
AFAIK, about specifically:
> There seems to be a race in that multiple threads on a process may end up doing the mmap()/dlopen() dance and we might end up with multiple instances of the loaded symbol and different ones may be called (as the cache gets updates), but I'm not sure if that's any issue (other than wasted memory).

`dlopen` should be concurrent safe, ie it has a locking mechanism + reference counting should avoid any issue (we'll end up with a single library with 2 references).
`mmap` instead can be a problem of course since we are allocating a 4k page, thus if it gets called multiple times because of a race, we'll end up wasting a bit of memory.

> If we are going ahead with this, maybe we can have other threads just call the old function once one threads has initiated the replacement.

This might be a good idea and a simplification over my code; it would make `pending_calls` too a per-process map and would just skip any concurrent `symbol` lookup in the process (following `resolved_cache`, the key of the map could be `struct cache_key {__u32 tgid;__u32 sym_id;};`).
Would love to hear what @olsajiri thinks about it.

FedeDP:
> This is awesome! And I mean this both in the good way (as in, the technical achievement of making this work is amazing) and that it really scares me (i.e., inspiring awe)

Btw, thanks for the kind and realistic words; this is what i feel too; it's between "oh it works :O" and "oh f*ck, it works" :rofl:

bpf/process/uprobe_offload.h
FedeDP: I did just that, ie: moved back `regs` related stuff to `uprobe_offload.h`; this lowers the diff lines.

bpf/process/uprobe_offload.h
```diff
@@ -18,7 +18,9 @@ struct reg_assignment {
 
 #if defined(GENERIC_UPROBE)
 
-#define REGS_MAX 18
+#define REGS_MAX   18
+#define SOPATH_MAX 128
```
FedeDP, line 22: Considering we are just allocating 128B in user context (via `mmap`) i am wondering whether leveraging `malloc` instead would be better.
Right now, we are requesting `mmap` to alloc a 4K memory page; this is a waste considering we only use 128B; i bet `malloc` would be more optimized (ie: internally calling `brk` or returning an already available heap chunkl) for little allocations.

FedeDP: Oh right, forgot: i used `MAP_POPULATE` flag for mmap  to force the kernel to fault in the returned page, so that `probe_write_user()` won't hit a not-yet-resident page.

FedeDP:
@olsajiri re:

> if signal comes and does dlopen/mmap/dlsym.. unlikely I guess, but that will screw all up, right?

Was thinking about a mitigation among the lines of:
```C
pending_c->expected_sp = PT_REGS_SP(ctx);
jump_to_mmap(ctx, cached_addrs->mmap_addr);
```
then in mmap retprobe:
```C
pending_c = task_storage_get(&pending_calls, task, NULL, 0);
if (!pending_c || pending_c->stage != STAGE_MMAP_PENDING)
    return 0;

if (PT_REGS_SP(ctx) != pending_c->expected_sp)
    return 0;
```
After all we are managing the whole call stack during these calls!
It should be a nice defense; while it can happen that `SP` happens to be the same, it is a fair mitigation IMO.
WDYT?

EDIT: i got the patch ready in case you are ok with it ;)

olsajiri:
> @olsajiri re:
> 
> > if signal comes and does dlopen/mmap/dlsym.. unlikely I guess, but that will screw all up, right?
> 
> Was thinking about a mitigation among the lines of:
> 
> ```c
> pending_c->expected_sp = PT_REGS_SP(ctx);
> jump_to_mmap(ctx, cached_addrs->mmap_addr);
> ```
> 
> then in mmap retprobe:
> 
> ```c
> pending_c = task_storage_get(&pending_calls, task, NULL, 0);
> if (!pending_c || pending_c->stage != STAGE_MMAP_PENDING)
>     return 0;
> 
> if (PT_REGS_SP(ctx) != pending_c->expected_sp)
>     return 0;
> ```

nit, don't you store original context already?

> 
> After all we are managing the whole call stack during these calls! It should be a nice defense; while it can happen that `SP` happens to be the same, it is a fair mitigation IMO. WDYT?
> 
> EDIT: i got the patch ready in case you are ok with it ;)

I think that could work, the stack pointer should be safe to check

FedeDP:
> nit, don't you store original context already?

I don't store `sp` though, we don't need that.

FedeDP:
I'll push the changes then post a PR comment with a whole summary of potential issues and if/how they are managed by the code.

bpf/process/uprobe_dyn.h
```diff
@@ -0,0 +1,319 @@
...
+		return 0;
+	}
+
+	DEBUG_SO("symbol %d non found in cache, try to load it", sym_id);
+
+	pending_c = task_storage_get(&pending_calls, task, NULL, BPF_LOCAL_STORAGE_GET_F_CREATE);
+	if (!pending_c || pending_c->stage != STAGE_NONE)
+		return 0; // either failed to get task-storage or already resolving on this thread
+
+	cached_addrs = map_lookup_elem(&libc_addrs_map, &tgid);
```
olsajiri, line 158: looks like we never remove from libc_addrs_map, right? I wonder it'd be easier to have it as another task storage data, but that's per thread, so maybe not the best choice.. we could maybe remove that in exit sensor in `event_exit_send`

bpf/process/uprobe_offload.h
```diff
@@ -90,6 +118,12 @@ uprobe_offload(struct pt_regs *ctx)
 	if (!regs)
 		return 0;
 
+	// We expect to fully override the call to
+	// a dynamically loaded one
+	if (regs->sopath_len > 0) {
```
olsajiri, line 124: so this is very targeted code, we know from action spec and we execute it from this place.. could we have rodata variable for this so it's not loaded when it's not used?

bpf/process/uprobe_dyn.h
olsajiri: then I guess we could add it just to the latests objects and use `__V61_BPF_PROG` instead of 5.11

pkg/sensors/program/dynamic_so.go
olsajiri: I guess you could parse ld.so cache, like bcc and bpftrace (through bcc) do

olsajiri (commented):
curious, did you try to get any performance numbers?  how much slower is calling the function through all this compared to just normal call

not sure we could find good workload to meassure the performance hit with (ret)uprobes on mmap/dlopen

bpf/process/uprobe_offload.h
FedeDP: Isn't rodata config global and bpf probe startup time constant?
I mean, this is a per-spec value that can also change at runtime if an uprobe policy gets pushed with `sopath`

pkg/sensors/program/dynamic_so.go
FedeDP: Interesting, let me give it a look! Thanks for the hint!

bpf/process/uprobe_offload.h
olsajiri: ah crap, not by spec sry.. but there's the vma iter dependency

bpf/process/uprobe_dyn.h
FedeDP: > I wonder it'd be easier to have it as another task storage data, but that's per thread,

Yes, that's the problem :( 

> we could maybe remove that in exit sensor in event_exit_send

I thought about this too, but didn't want to pollute other code areas with code specific for this feature, ie: i wanted to make this as self-contained as possible.
Let me think a bit if we have any more possibility or that's just it.

pkg/sensors/program/dynamic_so.go
FedeDP: Mmmh the problem with that is it won't work for systems that use `musl`, right? Since `musl` does not use/generate `/etc/ld.so.cache`.

FedeDP: I think as a first implementation, relying upon the libc regex for the host is quite ok; either way, we will need to revisit this code in the future when we support uprobes in containers.
(Actually this is the only part of the PR that should need adaption to that; basically instead of loading `hostLibc` we would look for libc inside the container rootfs).

FedeDP:
> curious, did you try to get any performance numbers? how much slower is calling the function through all this compared to just normal call
> 
> not sure we could find good workload to meassure the performance hit with (ret)uprobes on mmap/dlopen

Unfortunately i haven't got any number; but, after all, aside from the very first call that needs the triple jump (hook -> mmap -> dlopen -> symbol), since the symbol address gets cached afterwards, the following calls will get the same performance as `Override.argNewSymbol` action.

bpf/process/uprobe_offload.h
FedeDP: > but there's the vma iter dependency

Sorry i could not quite follow you :D what do you mean ?

olsajiri: I thought we should limit that code for using the vma iterator, but I think the limit is already with register change support, which covers that, so no problem

FedeDP: Oh i see, thanks for clarifying it.

PR #5566, opened 2026-09-04
policytest enhancements

andrewstrohman:
Allow policy tests to declare required agent CLI flag values. Tests are skipped with a descriptive reason when the running agent’s reported configuration does not satisfy those requirements.

Extend agent configuration encoding and decoding to support the additional scalar and list value types needed for reliable comparison, including floating-point values, durations, and integer slices.

### DIFF
--- a/cmd/tetragon-vmtests-run/policytest.go
+++ b/cmd/tetragon-vmtests-run/policytest.go
@@ -20,6 +20,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 	"time"
 
@@ -45,6 +46,7 @@ type PolicyTestConf struct {
 	testerProgsTarball  string
 	resultsDir          string
 	policytestExtraArgs map[string]string
+	environment         map[string]string
 }
 
 func (rc PolicyTestConf) testImageFilename() string {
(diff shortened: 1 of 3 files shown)


pkg/testutils/policytest/policytest.go
```diff
@@ -50,6 +52,81 @@ type Parameter struct {
...
+}
+
+// formatCLIFlagValue formats a value using pflag-compatible syntax.
+// For example, []int{42, 43} is formatted as "42,43" instead of "[42 43]".
+func formatCLIFlagValue(value any) string {
+	if values, ok := value.([]int); ok {
+		parts := make([]string, len(values))
+		for i, value := range values {
+			parts[i] = strconv.Itoa(value)
+		}
```
will-isovalent: Nit: I don't like shadowing `value` here. Can we pick another name?

pkg/testutils/policytest/policytest.go
```diff
@@ -50,6 +52,81 @@ type Parameter struct {
...
+	case []int:
+		actual, ok := actual.([]any)
+		if !ok || len(actual) != len(expected) {
+			return false
+		}
+		for i, value := range actual {
+			number, ok := value.(float64)
+			if !ok || number != float64(expected[i]) {
+				return false
+			}
```
will-isovalent, line 130: We're missing a case to handle `Duration` here.

will-isovalent (commented):
Thanks! Left a couple comments.

pkg/testutils/policytest/policytest.go
andrewstrohman: Make sense. I changed value to val to prevent the shadowing.

pkg/testutils/policytest/policytest.go
andrewstrohman: Good catch. Thanks.

will-isovalent (approved):

